PR #7029, opened 2024-04-05
feat: add version graphql call to get engine version

jedevc:
Before this, there was no way to get the version of an engine - this is *really* useful if the engine is running remotely, and you want to make a query for it to get it programatically (it's always been possible by analyzing logs, etc).

One of my main use cases is for working out what version of dagger the playground is using: https://play.dagger.cloud/playground - there's actually no way to do that right now (this PR would let you).

What feels frustrating is that `checkVersionCompatibility` is *almost* the right method call, but doesn't actually give you the raw data.


helderco (approved):
Needs a change log

PR #7013, opened 2024-04-04
hack: add legacy label conversion for old clients

jedevc:
This allows old clients like dagger v0.10.3 to still connect to and use dagger engine v0.11.0. While this *isn't* a strict requirement, it's nice to not break this if we don't have to, and based on a couple tests, it seems like this was the only blocker for that.

The other option for this was to introduce an explicit failure for when clients mismatch, but... this is potentially a bit more destructive, this change is pretty minimal.

PR #7001, opened 2024-04-03
chore: add v0.11.0 release notes

jedevc:

PR #5892, opened 2023-10-13
feat: fail when connecting to vanilla buildkit

jedevc:
This adds a sanity check to check that when we connect to a remote buildkit instance, that it's running the dagger engine, and not a vanilla instance of buildkit.

See https://github.com/dagger/dagger/issues/5852#issuecomment-1756568885:

> If you are trying to connect to a host that was provisioned via buildx then it's not expected to work at this time. Dagger has a bit of divergence from the buildkitd that buildx uses.

We should hard fail on this kind of scenario, since it's possible to detect this by inspecting the `Package` value from buildkit's `Info` call. However, to do this, I've had to change the meaning of `Package` to contain the go package name, and smuggle the engine name through the `Revision` field. This would make this a breaking change for users with mismatched cli and engine versions - I'm not sure if this is important, if it is, we could do a weaker form of the check and just fail on the case that `info.BuildkitVersion.Package == "github.com/moby/buildkit"`.

It might be useful to upstream the "engine name" functionality into buildkit (e.g. https://github.com/jedevc/buildkit/commit/f0fe58a1a991f915131c1f6e89871c0f11f03ec5), to avoid the need to smuggle through another field, wdyt @sipsma?

Issue #5583, opened 2023-08-04
Customize engine with `dagger run`


shykes:
## Overview

There is increasing demand for customizing the dagger engine, including:

* What version (this carries specific problems because of the version compat matrix, but I will ignore those for now)
* *How* to run the engine container (`docker`, `podman`, `nerdctl`, custom script, etc)
* Runtime customization of the engine container (eg. TLS certificates)
* Custom hooks, for example for fine-grained control of cache synchronization to Dagger Cloud (credits to @kpenfound for deploying those hooks in production)

Right now the dominant method is to *wrap*: either wrap `dagger run`, or wrap native execution of the custom tool; by disabling the engine auto-deployment feature.

Instead I propose a *hook* approach: always use `dagger run`; always rely on auto-deployment; but make it as configurable as needed, with CLI flags.

This is aligned with the newly clarified architectural direction, of making the engine increasingly stateless. See #5484 , cc @sipsma

shykes:
FYI @gerhard

gerhard:
## Thoughts on the problem

I definitely see the problem.

As soon as the CLI stops managing the Engine, the burden falls on the "ops person" to ensure that the CLI/SDK run against the correct Engine version. This brings further complications, e.g. how do we know what Engine version the CLI needs? how do we prevent incompatible versions running together? Ultimately, this results in poor DX, which I know is super important to us.

## Thoughts on the solution

Allowing users to select from various built-in Engine provisioning mechanisms is essential for good DX. `docker` seems like an important built-in provisioning mechanism - not sure about the default one, but let's put a pin in this for now. A built-in `k8s` provisioner is something that I would be excited to contribute. As for a built-in `fly` provisioner, this could be a good start: https://github.com/thechangelog/changelog.com/pull/471/files#diff-1fd071d093832bcb6c4a3a4c9d5b3740a0a71a68a42a8a52b422e43e7c84f12c

## Thoughts on the design / implementation

I think of this as client-side Dagger pipelines since there is no Engine to run these provisioners in. I can also imagine a world where we have an Engine built into the CLI by default, but I remember us looking at this in the past, and it not being straightforward enough. FTR:
- https://github.com/dagger/dagger/issues/3058
- https://github.com/dagger/dagger/pull/3187

The reason why I strongly believe that we should approach the design as a DAG problem is because:
- resilient setups will require a fall-back mechanism
[...]

PR #7031, opened 2024-04-05
Introduce better versioning compatibility guarantees

jedevc:
See the rationale in https://github.com/dagger/dagger/pull/7013, https://github.com/dagger/dagger/pull/7001#discussion_r1549487821. TL;DR, during the release of v0.11.0, we broke compatibility with older clients, but didn't really have the tooling to make sure that this *errored* cleanly out.

This PR ensures that this doesn't happen again - both the client and the server are able to specify a minimum version, defined in `engine/version.go`. If a client/server doesn't satisfy these constraints, an appropriate error is printed out. These checks are introduced into the client/server code *directly*, instead of into each SDK, which don't trigger in dagger module code (see commit messages for more detailed description of the exact logic used).

For example, during the release of v0.11.0, if we'd had this, we would've bumped both the version compats to v0.11.0:
- Clients <=0.10.3 connecting to >v0.11.0 would be rejected by the server
- Servers <=0.10.3 being connected to by >=v0.11.0 would be rejected by the client

Even though we can't add this for v0.11.0, we *can* for v0.11.1 (or later) - it's designed to work with client/servers of as far back as possible (it *should* even work for v0.9, when all compat was deliberately broken in #5892).

If we take this, I wonder if we should think about deprecating/removing `checkVersionCompatibility` call? This new approach enforces constraints on both the client/server instead of just the server, and allows a more find-grained approach (so we can choose specific compatibility with each release on an ad-hoc basis, instead of using [hardcoded rules](https://github.com/dagger/dagger/blob/95b4ce550a2dcd1fa73f8b55199c7e9ed0ef1fc3/core/schema/query.go#L94-L115)).

I have *no* idea how to test this relatively fragile logic - since the version numbers come from hardcoded numbers in the binaries, it might be relatively fiddly. Will do some brainstorming and try and come up with something.

Note: at some point, the compute drivers to be added in #5583 should resolve this issue entirely, by making sure that the client/engine *always* match - but that may be some way off, and this is a relatively simple fix to prevent user issues until then. 

Related:
- #5315 
- #7029 (kept getting frustrated this didn't exist while working on this PR)

### DIFF
--- a/engine/version.go
+++ b/engine/version.go
@@ -8,36 +8,57 @@ import (
 )
 
 var (
-	EngineImageRepo = "registry.dagger.io/engine"
+	// Version holds the complete version number.
+	//
+	// Note: this is filled at link-time.
+	//
+	// - For official tagged releases, this is simple semver like x.y.z
+	// - For builds off our repo's main branch, this is the git commit sha
+	// - For local dev builds, this is a content hash of the source directory
+	Version string
 
-	Package = "github.com/dagger/dagger"
+	// MinimumEngineVersion is used by the client to determine the minimum
+	// allowed engine version that can be used by that client.
+	MinimumEngineVersion = "v0.11.7"
 
-	GPUSupportEnvName = "_EXPERIMENTAL_DAGGER_GPU_SUPPORT"
+	// MinimumClientVersion is used by the engine to determine the minimum
+	// allowed client version that can connect to that engine.
+	MinimumClientVersion = "v0.11.7"
 )
 
-// Version holds the complete version number. Filled in at linking time.
-// For official tagged releases, this is simple semver like x.y.z
-// For builds off our repo's main branch, this is the git commit sha
-// For local dev builds, this is a content hash of dagger repo source from "git write-tree"
-var Version string
-
 func init() {
-	// normalize Version to semver
-	if semver.IsValid("v" + Version) {
-		Version = "v" + Version
+	// hack: dynamically populate version env vars
+	// we use these during tests, but not really for anything else
+	if v, ok := os.LookupEnv(DaggerVersionEnv); ok {
+		Version = v
 	}
+	if v, ok := os.LookupEnv(DaggerMinimumVersionEnv); ok {
+		MinimumClientVersion = v
+		MinimumEngineVersion = v
+	}
+
+	// normalize version strings
+	Version = normalizeVersion(Version)
+	MinimumClientVersion = normalizeVersion(MinimumClientVersion)
+	MinimumEngineVersion = normalizeVersion(MinimumEngineVersion)
 }
 
-func RunnerHost() (string, error) {
-	if v, ok := os.LookupEnv("_EXPERIMENTAL_DAGGER_RUNNER_HOST"); ok {
-		return v, nil
+func normalizeVersion(v string) string {
+	if semver.IsValid("v" + v) {
+		return "v" + v
 	}
+	return v
+}
 
-	gpuSupportEnabled := os.Getenv(GPUSupportEnvName) != ""
-
-	tag := Version
-	if gpuSupportEnabled {
-		tag += "-gpu"
+func CheckVersionCompatibility(version string, minVersion string) error {
+	if !semver.IsValid(version) {
+		return nil // probably a dev version
 	}
-	return fmt.Sprintf("docker-image://%s:%s", EngineImageRepo, tag), nil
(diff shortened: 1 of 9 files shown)


sipsma (commented):
Looks great, but agree that testing is both important and very unclear here.

The best thing I can think of is to change our version strings such that non released versions always include the most recent released tag as a prefix (using something like this from the sermver spec https://semver.org/#spec-item-9). That would *at least* allow us to test this once this has been released. But also much easier said than done.

After that other options I can see:
* Add dumb env vars like `_INTERNAL_DAGGER_TEST_OVERRIDE_VERSION` that let you override the version values of CLI/engine.
   * This is "dumb" since it *technically* exposes it as something available to users, but in the grand scheme of things it's obscure enough that no one would be doing it unintentionally and if they do it intentionally then the warranty is pretty clearly broken I suppose.
   * I'd be totally fine with this approach personally since getting test coverage of this is very high value IMO
* Maybe it's somehow unit testable, though then we'd start leaking those concerns the structure of our code which I never like, but still worth it if nothing else possible

jedevc:
Right :tada: Finally managed to get this one together again, and refresh my memory.

As a note:
- The first commit ensures that we can extract version numbers from old pre-v0.11 clients - otherwise, we have no idea what version they're using. (this is tested manually, since it's a *bit* of a pain to write CI tests with old engine/client versions).
- The second commit ensures that when the engine rejects an old client that the client exits (and doesn't hang). This is a "bug" in older client versions, which could also have happened for other reasons as well. (note, assuming this gets into v0.11.7, if a pre-v0.11.7 client attempts to connect to a v0.11.7 server, then it *will* be rejected, but it *won't* exit, it'll just hang - this isn't really fixable sadly - but it will hang with a "useful" error).
- The third commit is the magic - it ensures that the client checks the server and the server checks the client.
- The final commit is a test - it uses a magical env override to test the possible mismatch scenarios.

> The best thing I can think of is to change our version strings such that non released versions always include the most recent released tag as a prefix (using something like this from the sermver spec [semver.org/#spec-item-9](https://semver.org/#spec-item-9)). That would at least allow us to test this once this has been released. But also much easier said than done.

@sipsma thinking about this, I think we *eventually* want to add this. Something like how go pseudoversions work. For example, on main *right now*, the versions we'd be tagging could look something like `v0.11.7-{timestamp}.{commit-sha}`. That way, builds cut directly off of main would also work with these version compat checks.

However, currently, those aren't really widely used - and also doing this properly is kinda annoying and out of scope.

So for the time being, these versioning checks will only happen on *tagged* releases - any versions that don't conform to semantic versioning are treated as dev releases, and don't have these checks.

engine/version.go
```diff
@@ -8,36 +8,57 @@ import (
...
 
-	Package = "github.com/dagger/dagger"
+	// MinimumEngineVersion is used by the client to determine the minimum
+	// allowed engine version that can be used by that client.
+	MinimumEngineVersion = "v0.11.7"
 
-	GPUSupportEnvName = "_EXPERIMENTAL_DAGGER_GPU_SUPPORT"
+	// MinimumClientVersion is used by the engine to determine the minimum
+	// allowed client version that can connect to that engine.
+	MinimumClientVersion = "v0.11.7"
```
jedevc, line 26: @vito this is the version check I think we want?

From the next release (v0.11.7):
- The CLI needs engine v0.11.7
- The engine needs CLI v0.11.7

vito: Yep sounds good!

sipsma: I guess we should add a step to `RELEASING.md` for updating this when needed

engine/version.go
```diff
@@ -8,36 +8,57 @@ import (
...
 
 var (
-	EngineImageRepo = "registry.dagger.io/engine"
+	// Version holds the complete version number.
+	//
+	// Note: this is filled at link-time.
+	//
+	// - For official tagged releases, this is simple semver like x.y.z
+	// - For builds off our repo's main branch, this is the git commit sha
+	// - For local dev builds, this is a pseudo-random dagger ID of the source directory
```
sipsma, line 17: nit: it's a content hash of the source dir right? as opposed to random per-se

sipsma (approved):
minor comments but LGTM!

engine/version.go
jedevc: @sipsma added to `RELEASING.md`, good shout!

Issue #6933, opened 2024-03-26
Rust arg types for `loadXFromId` are incorrect


jedevc:
For example:

https://github.com/dagger/dagger/blob/3bc4bd1263570d67ced697d0b453c5d418ceb4ce/sdk/rust/crates/dagger-sdk/src/gen.rs#L5156

This type is wrong - it should instead be:

```diff
-     pub fn load_container_from_id(&self, id: Container) -> Container {
+     pub fn load_container_from_id(&self, id: ContainerID) -> Container {
```

The top-level `Query.Container` has [been deprecated for some time](https://github.com/dagger/dagger/blob/3bc4bd1263570d67ced697d0b453c5d418ceb4ce/docs/docs-graphql/schema.graphqls#L2111-L2117), and I'd like to remove it for v0.11.0 - but since `load_container_from_id` takes the wrong type, existing instances can't be replaced.

cc @dagger/sdk-rust

kjuulh:
@jedevc this should now be fixed in #7038 We are still calling `.id()` on the struct I don't know if that is an issue, but getting out of that will require some more trickery. Let me know if that is the case.

jedevc:
Should be fixed in #7038, closing! :tada:

PR #7038, opened 2024-04-06
feat: use trait for Struct -> StructID instead of renaming objects

kjuulh:
This feature should allow both `id` and regular structs/objects/scalars to be passed in to a function. This was initially brought on by: #6933.

Which now allows removing deprecated input


sdk/rust/crates/dagger-codegen/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![deny(warnings)]
+//#![deny(warnings)]
```
jedevc, line 1: :eyes: is this intentional?

jedevc (approved):
Ooh I really like the idea of being able to pass both IDs / the objects from the SDK-side - that's really neat!

That definitely resolves the issue in #6933.

sdk/rust/crates/dagger-codegen/src/lib.rs
kjuulh: Woops nope. I've got another pr with some changes to this and I didn't manage to get that in. I'll get it fixed =D

jedevc (changes requested):
Rebased, and also fixed the warnings issue.

I also tried updating the use of the old `Container` and couldn't:

```diff
-        .container_opts(dagger_sdk::QueryContainerOpts {
-            id: Some(id),
-            platform: None,
-        })
+        .load_container_from_id(id)
```

```
    --> crates/dagger-sdk/tests/mod.rs:78:33
     |
78   |         .load_container_from_id(id)
     |          ---------------------- ^^ the trait `IntoID<ContainerId>` is not implemented for `ContainerId`
     |          |
     |          required by a bound introduced by this call
     |
```

And there we hit the boundaries of my rust knowledge :laughing:

kjuulh:
> Rebased, and also fixed the warnings issue.
> 
> I also tried updating the use of the old `Container` and couldn't:
> 
> ```diff
> -        .container_opts(dagger_sdk::QueryContainerOpts {
> -            id: Some(id),
> -            platform: None,
> -        })
> +        .load_container_from_id(id)
> ```
> 
> ```
>     --> crates/dagger-sdk/tests/mod.rs:78:33
>      |
> 78   |         .load_container_from_id(id)
>      |          ---------------------- ^^ the trait `IntoID<ContainerId>` is not implemented for `ContainerId`
>      |          |
>      |          required by a bound introduced by this call
>      |
> ```
> 
> And there we hit the boundaries of my rust knowledge 😆 It feels like `ContainerID` should definitely implement `IntoID<ContainerID>`.

Yes that was the whole Idea. I'll have a look =D

kjuulh:
It should work now:

```rust
#[tokio::test]
async fn test_can_handle_old_and_new_load() -> eyre::Result<()> {
    let client = dagger_sdk::connect().await?;

    let base_container = client.container().from("alpine:latest");

    client
        .load_container_from_id(base_container.id().await?)
        .with_exec(vec!["echo", "from-id"])
        .sync()
        .await?;

    client
        .load_container_from_id(base_container.clone())
        .with_exec(vec!["echo", "from-id"])
        .sync()
        .await?;

    client
        .container_opts(QueryContainerOpts {
            id: Some(base_container.id().await?),
            platform: None,
        })
        .with_exec(vec!["echo", "from-id-original"])
        .sync()
        .await?;

    Ok(())
}

```

jedevc (approved):

Issue #6310, opened 2023-12-21
🐞 Elixir: strange experimentalWithAllGPUs function name


wingyplus:
### What is the issue?

See https://hexdocs.pm/dagger/Dagger.Container.html#experimental_with_all_gp_us/1. The function name should be `experimental_with_all_gpus`, `experimental_with_all_gp_us`

### Dagger version

dagger v0.9.4

jedevc:
cc @vito since you've been in the graphql schema space :smile: 

I think this is some weirdness with how the string conversion library we use works, `GPUs` gets split into `GP`+`Us` instead of keeping it as is.

vito:
My guess is we need to use [`strcase.ConfigureAcronym`](https://pkg.go.dev/github.com/iancoleman/strcase#ConfigureAcronym)

wingyplus:
I workaround it by replacing it with `Gpus` (https://github.com/wingyplus/dagger/blob/elixir-new-codegen/sdk/elixir/dagger_codegen/lib/dagger/codegen/elixir_generator/formatter.ex#L21). 😂

I found this issue on the Rust SDK  also (https://github.com/dagger/dagger/blob/9bf9f7f303b6768efbba540376a0bd6eae6979a3/sdk/rust/crates/dagger-sdk/src/gen.rs#L889).

jedevc:
Eek we also have this issue in python (cc @helderco): https://github.com/dagger/dagger/blob/2999573c5a34e85b8baac8c0150881d9c08a86b8/sdk/python/src/dagger/client/gen.py#L501

Poked into this - it looks like this is left up to the SDK codegen to convert, the input at the point of generation is `experimentalWithAllGPUs`.

In the rust case (which I'm more familiar with), this function performs the incorrect conversion: https://github.com/dagger/dagger/blob/2999573c5a34e85b8baac8c0150881d9c08a86b8/sdk/rust/crates/dagger-codegen/src/rust/functions.rs#L17-L19

Super weird though, because, if I change the code slightly:

```rust
pub fn format_struct_name(s: &str) -> String {
    let conv = Converter::new()
        .remove_boundary(Boundary::Acronym)
        .to_case(Case::Snake);
    conv.convert(s)
}
```

Then I get:

```diff
diff --git a/sdk/rust/crates/dagger-sdk/src/gen.rs b/sdk/rust/crates/dagger-sdk/src/gen.rs
index 7aa880b6e..5f9803f3c 100644
[...]

PR #7039, opened 2024-04-06
feat(sdk/elixir): improve acronym word conversion

wingyplus:
Use regular expression instead of dictionary of words to enable any module that use acronym word to generate correct name in Elixir.

Updates #6310


wingyplus:
This approach is still not good through. It can solve on `experimentalWithAllGPUs` case, but not for `VCSIgnoredPath`

Issue #6901, opened 2024-03-18
Windows installation instructions do not work


jedevc:
It doesn't look like we can specify a version to download in dagger using the instructions here: https://docs.dagger.io/quickstart/729237/cli:

```
$script = Invoke-WebRequest -UseBasicParsing -Uri https://dl.dagger.io/dagger/install.ps1
$params = "-DaggerVersion 0.9.7"
"$script $params" | Invoke-Expression
```

If I look at the contents at that script:

```
# param (
#     [Parameter(Mandatory)] $PersonalToken
# )
Clear-Host
@"

---------------------------------------------------------------------------------
Author: Alessandro Festa
Dagger Installation Utility  for Windows users
---------------------------------------------------------------------------------

"@

# Since we are already authenticated we may directly download latest version.
$name="dagger"
$base="https://dl.dagger.io"
function http_download {
    $version=Get_Version
    $version=$version -replace '[""]'
    $version=$version -replace '\n'
    $fileName="dagger_v" + $version + "_windows_amd64"
    Clear-Host
    $url = $base + "/" + $name + "/releases/" + $version + "/" + $fileName + ".zip"
    write-host $url
    Pause


    Invoke-WebRequest -Uri $url -OutFile $env:temp/$fileName.zip -ErrorAction Stop
    Expand-Archive -Path $env:temp/$fileName.zip -DestinationPath $env:HOMEPATH/dagger -Force -ErrorVariable ProcessError;
    If ($ProcessError)
    {
    Clear-Host
@"
Whoops apparently we had an issue in unzipping the file, please check
you have the right permission to do so and try to unzip manually the file.
Currently we saved Dagger at your temp folder.
"@
exit
    } else {
        Clear-Host

@"

Thank You for downloading Dagger!

-----------------------------------------------------
Dagger has been saved at <YOUR HOME FOLDER>/dagger/
Please add dagger.exe to your PATH in order to use it
----------------------------------------------------

"@
    }

}

function Get_Version {
    $response = Invoke-RestMethod 'http://releases.dagger.io/dagger/latest_version' -Method 'GET'  -Body $body -ErrorAction SilentlyContinue -ErrorVariable DownloadError
    If ($DownloadError)
    {
    Clear-Host
@"

---------------------------------------------------------------------------
Houston we have a problem!

Apparently we had an issue in downloading the file, please try again
run the script and if it still fail please open an issue on the Dagger repo.
----------------------------------------------------------------------------

"@
exit
    }
    return $response

}

[...]

PR #6843, opened 2024-03-06
Modularize CI

jedevc:
This PR introduces a *fully-operational* `ci/` directory, replacing the old non-functional one.

This new CI allows for all of the following:
- Running our tests
- Utilities like linting, code-gen, bumping
- Building the engine/cli/sdks

> [!WARNING]
>
> This isn't fully ready yet, but opening a work-in-progress PR to make it easier to track.
>
> The aim isn't to get everything perfect on a first-pass, but to do a lift-and-shift of the existing `internal/mage` CI into dagger functions. Then we can follow-up with improvements.

A few things I've realized while doing this that I think are technically out-of-scope for this particular piece of work, but we should make sure to tackle:
- Syncing the source directory is *slooow* - running commands over-and-over takes a lot of time, and it feels it (we need #6155)
- Working with multi-platform images with functions feels very manual (see #6675)
- I keep finding myself doing `--source=. ... export --path=.` to generate code. I get the technical reasons for this, but the verbosity of this makes me want to reach for a wrapper, which I know we're looking to avoid.
- I keep wanting a nice way to load an image into the docker engine, without needing to do the intermediate export/import. This one is tricky, since it would be a CLI-facing change, instead of an engine side change - it would also need to integrate into various other compute drivers (e.g. podman) #5583.
- The build changes from #6714 create a `Sync` in the middle of the chain, meaning that large parts are done in serial - this slows the build down, but struggling to work out how we can make those operations parallel again (cc @sipsma)
	- This was present before, but with the new TUI, it's much easier to visualize what's actually happening at each point, so it's easy to see that the chain gets broken.
- Some operations just seem... slow. I can't quite put a finger on what this is, but even when there are no real changes, it seems to take much longer than it should to get a cached result.
	- Again, I don't think this is introduced by this change, but the TUI shows that there's lots of bottlenecks at different points - it would be interesting to try and track where we're spending all this time just to serve a purely cached result.

TODO:
- [x] Port engine+cli building and tests to new modules structure
	- [x] GPU images need some additional work here
- [ ] Port the SDKs
	- [x] Port the Go SDK
	- [x] Port Java/PHP/Elixir/Rust SDKs
	- [x] Port Python/Typscript SDKs
	- [ ] Port Python/Typscript SDKs (using dev modules)
- [x] Port publishing
	- [x] Port engine publishing
	- [x] Port cli publishing
- [x] Port docs
- [x] Remove the old mage targets, and alias them to the new functions
[...]

PR #5829, opened 2023-09-27
install.sh uses CloudFront CDN for CLI installs

gerhard:
We (+@jpadams) have noticed slow-downs when installing via `curl -L https://dl.dagger.io/dagger/install.sh | sh`

This PR changes the base URL in `install.sh` from `dagger-io.s3.amazonaws.com` to `dl.dagger.io`

Since `install.sh` file was deleted part of https://github.com/dagger/dagger/commit/5a9c772020363668c18e73472534f01a5b03b8d5, this PR also adds it back. cc @aluzzardi

### Next steps
- [x] One of the reviewers approves this PR
- [x] I update https://dl.dagger.io/dagger/install.sh
  - FTR https://github.com/dagger/dagger/pull/1819#issuecomment-1078511779
- [ ] PR gets merged

PR #7050, opened 2024-04-08
chore: restore install.ps1

jedevc:
Fix https://github.com/dagger/dagger/issues/6901.

This script was removed in https://github.com/dagger/dagger/commit/5a9c772020363668c18e73472534f01a5b03b8d5, and we still have windows install instructions, so we should probably restore this!

I've also updated the script with a few additions, like making the function names map to the same as the unix script.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -1,0 +1,95 @@
+#Requires -Version 7.0
+
+param (
+    [Parameter(Mandatory = $false)] [System.Management.Automation.SemanticVersion]$DaggerVersion,
+    [Parameter(Mandatory = $false)] [string]$InstallPath = $env:HOMEPATH + '\dagger',
+
+    [Parameter(Mandatory = $false)] [System.Boolean]$InteractiveInstall = $false
+)
+
+@"
+---------------------------------------------------------------------------------
+Author: Alessandro Festa
+Co Author: Brittan DeYoung
+Dagger Installation Utility for Windows users
+---------------------------------------------------------------------------------
+"@
+
+$name="dagger"
+$base="https://dl.dagger.io"
+
+function execute {
+    $url = base_url
+    $filename = tarball
+    $url = $url + "/" + $filename
+    write-host $url
+    if ($InteractiveInstall) {
+        Pause
+    }
+
+    Invoke-WebRequest -Uri $url -OutFile $env:temp/$filename -ErrorAction Stop
+    Expand-Archive -Path $env:temp/$filename -DestinationPath $InstallPath -Force -ErrorVariable ProcessError;
+    If ($ProcessError) {
+@"
+---------------------------------------------------------------------------
+Whoops apparently we had an issue in unzipping the file, please check
+you have the right permission to do so and try to unzip manually the file.
+Currently we saved Dagger at your temp folder.
+---------------------------------------------------------------------------
+"@
+        exit
+    } else {
+@"
+
+Thank You for downloading Dagger!
+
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -54,6 +54,22 @@ blobs:
     folder: "dagger/releases/{{ .Version }}"
 
 publishers:
+  - name: publish-install-sh
+    cmd: sh -c "aws s3 cp install.sh s3://{{ .Env.AWS_BUCKET }}/dagger/install.sh"
```
jedevc, line 58: From what I can tell, there's currently no automation for this right now? This is what prevented `install.ps1` from being updated in the docs in the first place.

If so, I think this is the right way to do it, we should have goreleaser release the install script.

install.ps1
```diff
@@ -0,0 +1,95 @@
...
+    [Parameter(Mandatory = $false)] [System.Boolean]$InteractiveInstall = $false
+)
+
+@"
+---------------------------------------------------------------------------------
+Author: Alessandro Festa
+Co Author: Brittan DeYoung
+Dagger Installation Utility for Windows users
+---------------------------------------------------------------------------------
+"@
```
jedevc, line 16: I'm not sure about keeping this, none of our other install scripts or tools have author names like this.

jpadams: I added a commit to convert this author block to a comment instead of it being output on every execution.

.goreleaser.yml
jpadams: makes sense to me if it works and has thumbs up from @matipan or @gerhard 👍

install.ps1
```diff
@@ -0,0 +1,93 @@
...
+    [Parameter(Mandatory = $false)] [string]$InstallPath = $env:HOMEPATH + '\dagger',
+
+    [Parameter(Mandatory = $false)] [System.Boolean]$InteractiveInstall = $false
+)
+
+# ---------------------------------------------------------------------------------
+# Author: Alessandro Festa
+# Co Author: Brittan DeYoung
+# Dagger Installation Utility for Windows users
+# ---------------------------------------------------------------------------------
```
jpadams, line 14: This script works fine for me. Now just need the `dagger.exe` that it downloads to give some output. That's the other PR :)

jpadams (approved):
If others can confirm infra side, let's ship it!

.goreleaser.yml
gerhard: We do this manually because we need to invalidate CloudFront too. FTR:
- https://github.com/dagger/dagger/pull/5829

If we want to do this properly, we should add it to a function as a follow-up to https://github.com/dagger/dagger/pull/6843

If we want to do this quick & dirty, we need to also invalidate CloudFront. Note-worthy past comments:
- https://github.com/dagger/dagger/pull/1819#issuecomment-1075508169
- https://github.com/dagger/dagger/pull/1819#issuecomment-1077651647
- https://github.com/dagger/dagger/pull/1819#issuecomment-1078511779

install.ps1
gerhard, line 1: This looks good to me. I haven't tested, but the shape is correct. FTR:

image

image

.goreleaser.yml
jedevc: Sounds good! Hadn't realized that, I've updated the RELEASING guide to include info about doing this.

jedevc:
Merging this since the script looks good - even if this ends up wrong, we can also update this on S3 async to our release process.

PR #7063, opened 2024-04-09
ci: port helm jobs to dagger function

jedevc:
Also, don't bother adding a shim for this one - it's not worth it, we can directly use dagger-for-github.

Again, testing this is a pain.


ci/helm.go
```diff
@@ -0,0 +1,66 @@
...
+	if dryRun {
+		helm = helm.With(util.ShellCmd(pkgCmd))
+	} else {
+		helm = helm.
+			With(util.ShellCmds(
+				"helm registry login ghcr.io/dagger --username dagger --password $GITHUB_TOKEN",
+				pkgCmd,
+				"helm push dagger-helm-"+version+".tgz oci://ghcr.io/dagger",
+				"helm registry logout ghcr.io/dagger",
+			))
```
jedevc, line 61: The previous impl of this was unsafe - we need to have the login+logout paired in the same exec, or the creds written to disk will end up in the cache.

gerhard (approved):
LGTM!

Looking at why some jobs were not picked up. Once all is green, we can merge!

jedevc:
Woop, looks like this worked, first time! https://github.com/dagger/dagger/pkgs/container/dagger-helm/207953084?tag=0.1.2

Issue #7033, opened 2024-04-05
Python: regenerate lock file


helderco:
Python Modules now (since [v0.11.0](https://github.com/dagger/dagger/releases/tag/sdk%2Fpython%2Fv0.11.0)) create a `requirements.lock` file by default, but does this only on `dagger init` (or `develop` if there’s no `pyproject.toml`), so it’s up to the user to update manually from then on: 
```console
uv compile --generate-hashes -o requirements.lock sdk/pyproject pyproject
```

This can create a catch 22 situation where the SDK’s dependencies have changed, you need to codegen to get the new requirements, but you can’t codegen because you don’t have the new requirements (assuming missing these requirements fails on importing the dagger library).

Rather than doing it only on new modules, it should be done automatically during codegen. 

I only didn’t implement it before because I was under the impression that it would upgrade the existing pinned packages every time (defeating the purpose of a lock file), thus leaving it to the user, but only caught the catch 22 situation later.

PR #7064, opened 2024-04-09
Python: fix lock file updates

helderco:
Fixes https://github.com/dagger/dagger/issues/7033

To simplify not needing to install the SDK's dependencies for codegen, the Python codegen script has been moved into an independent package with a minimal dependency.

Added a `codegen introspect` subcommand to avoid the need for the Python codegen to have HTTP capabilities in order to introspect from the API directly.

This helped simplifiy the runtime module. Now, if a module has a `requirements.lock` file, it'll be re-compiled automatically (without upgrading packages). Without this, changes to the SDK's dependencies in a way that affects `import dagger`, would fail during `dagger develop`, leading to a difficult manual update of `requirements.lock`.


cmd/codegen/main.go
```diff
@@ -71,6 +83,35 @@ func ClientGen(cmd *cobra.Command, args []string) error {
 	return Generate(ctx, cfg, dag)
 }
 
+func Introspect(cmd *cobra.Command, args []string) error {
```
wingyplus, line 86: That's nice!!

helderco: Yup 🙂 At some point it can come from the codegen module, but this was a quicker fix for now and can be reused in other SDKs.

wingyplus: I have plan to use the introspection from that module. But after see this PR, I will wait until this PR merge then. 😂

helderco:
Getting lots of failures due to:
```
429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit",
```

jedevc:
> Getting lots of failures due to:
> 
> ```
> 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit",
> ```

Should be fixed by https://github.com/dagger/dagger/pull/7072, sorry for the inconvenience!

ci/sdk_python.go
```diff
@@ -19,7 +19,7 @@ const (
 )
 
 var (
-	pythonVersions = []string{"3.10", "3.11", "3.12"}
+	pythonVersions = []string{"3.10", "3.11"}
```
jedevc, line 22: :eyes:

jedevc (approved):

ci/sdk_python.go
helderco: I added this in a first commit here, but then split in another PR. After rebasing looks like the git thinks I want to remove it!

Issue #6599, opened 2024-02-06
🐞 Dagger modules break in corporate environments


Vermillionaire:
### What is the issue?

For security reasons many companies man-in-the-middle all SSL traffic to the internet with their own certificate. This causes a `x509: certificate signed by unknown authority` issue in two spots for dagger:

1. Dagger engine. This is easy to solve by mounting the certificates in the container.
2. Dagger modules. The modules appear to be loading a tar container `/usr/local/share/dagger/go-module-sdk-image.tar` and building a go program when running any module command. The commands all fail when the `go build` does a `go get` for dependencies and hits the same x509 issue.

Steps I have tried to work around this issue:

1. Building a `ModuleRuntime` as recommended in the discord thread [here](https://discord.com/channels/707636530424053791/1186965742172782664). I quickly found that all the existing module runtimes point back to go as a dependency, which was the issue.
2. Putting my certificates in the engine and the tar container. This resolves the certificate issues but breaks modules for some reason. On a fresh project (or any project) made with `dagger mod init` it would always fail with "main object not found".

Posting this as recommended in discord since there appears to be no workarounds and modules are completely broken in this environment.

### Dagger version

v0.9.6

nipuna-perera:
Thank you for posting this issue I am in a similar env and have the same issues

Some context/related work: 

https://github.com/dagger/dagger/issues/6077
https://github.com/dagger/dagger/issues/6112#issuecomment-1814700418

They are more aligned to corp proxies but same deal with things like certificates.

CC @sipsma

nipuna-perera:
Additional context based on recent findings - any version mismatch (go toolchain) in `go.mod` or `go.work` files within existing go modules may cause the engine pull additional dependencies which may break within corporate environments which are unable to pull and validate modules from the internet without the right proxies or certificates.

https://discord.com/channels/707636530424053791/1209532081697325147

sipsma:
Fully tackling every variation on this general issue and making everything totally seamless will be a significant effort, but as a first step to improve the current situation I think we could perhaps mount the engine's `/etc/ssl/certs` into SDK containers. I think that would remove the currently impassable hard blocker.

It would still require users setup a custom engine (as described [here](https://github.com/dagger/dagger/blob/main/core/docs/d7yxc-operator_manual.md#custom-ca-certs)). But it should at least stop the worst of the bleeding and set us up for more iterative improvements. 

(adding to `v0.10.x` since I *think* this would be a relatively small scope of effort)

[...]

Issue #4989, opened 2023-04-20
Remove need to run shim inside Exec container


sipsma:
We currently run the shim both outside the container (to reconfigure the runc bundle) and inside exec containers, which makes it straightforward to redirect stdout/stderr, serve nested sessions, etc. It's the pid 1 and spawns the actual user process as a subchild.

However, this can cause problems for programs that insist on being pid 1 such as systemd.

We can instead run the shim completely outside the container. This will require at least:
1. Still handling stdout/stderr file redirects correctly, including when they are to submounts. 
   * Most promising path here may be to use `/proc/<pid>/root` to get access to the full filesystem the container sees.
2. Serving nested sessions
   * This will likely require nsenter'ing into the network namespace of the container
   * It also requires that the session reads and writes files from under `/proc/<pid>/root` for all relevant APIs, including file sync. File sync specifically can likely be addressed via some tweaks to [our callback here](https://github.com/dagger/dagger/blob/a4a13e2227b915b20d41b000b8219e56ec3d3ecd/engine/engine.go#L537).

There might be more than that, but that's what immediately comes to mind. 

One annoying part of this is that we currently use `runc run`, which combines create/start/stop into a single operation. But to get the container pid and setup file access in a non-racey way we'll probably need to split out into separate create/start/stop operations.

---

Not a requirement, but we could also consider moving the shim to a different approach as part of this if it makes any of this easier. Two ideas:
1. Switch to a containerd backend and convert our shim to be an actual containerd shim
2. Implement the shim functionality as a custom buildkit executor, which is possible now that we are using buildkit as a library with in our own custom engine binary.

`2` in particular is intriguing in that it may open more avenues such as adding custom session attachables, but it has the major potential downside of us needing to copy a lot of code from buildkit, which makes maintaining it harder in the long run.

We can also consider upstreaming at least the stdout/stderr handling, but I think the nested session might be too dagger specific to upstream unless we can think of a generic way of handling it.

Issue #6111, opened 2023-11-15
Intermittent `failed to get state for index` errors


sipsma:
These sorts of errors have been happening occasionally (~1 / day) in our CI approximately since we switched over everything to our shared runners:
```
Error: input:1: pipeline.pipeline.container.export failed to solve for container publish: failed to compute cache key: failed to get state for index 0 on copy /runc /usr/local/bin/runc
```
([run](https://github.com/dagger/dagger/actions/runs/6870983081/job/18686952381?pr=6102#step:5:980))

The exact message varies but seems to always happen on FileOps and include `failed to compute cache key: failed to get state for index 0`.

There is an upstream issue for this here: https://github.com/moby/buildkit/issues/3635

If we can repro the error and get useful data will move to the upstream issue (or just send PR w/ fix). Creating this to track in case anyone hits it and searches for the error message.

---

The fact that it seems to have only started once we switched to the shared runner is notable. My immediate gut reaction is that it probably has something to do with concurrent solves that overlap in vertexes since that's proven to be a fairly tricky + error prone codepath in buildkit and is something that would have not happened previously before switching to the shared runners.

matiasinsaurralde:
Have been observing it in the past days too: https://github.com/dagger/dagger/actions/runs/6937495840/job/18871658791?pr=6136

gerhard:
I know that we've been seeing these for a while, want to resume tracking them so that we have a better idea of how often they happen:
- https://github.com/dagger/dagger/actions/runs/7729432480/job/21072590483#step:6:720

gerhard:
https://github.com/dagger/dagger/actions/runs/7734943604/job/21103510867#step:6:882

![image](https://github.com/dagger/dagger/assets/3342/f95b7227-3f0a-4db3-b9a3-61bebe6ec86b)

gerhard:
Just hit this again while releasing `0.10.2`: https://github.com/dagger/dagger/actions/runs/8253625814/job/22575935633?pr=6871#step:6:772

<img width="1440" alt="image" src="https://github.com/dagger/dagger/assets/3342/add0d775-6116-45a2-b87a-137c8c2ea21f">

mckinnsb:
I also get this error, but I get a different "source":

```
[...]

PR #7067, opened 2024-04-09
core: support automatic installation of custom CA certs.

sipsma:
Still some cleanup left, but good enough for feedback. TODOs:
- [x] finish description
- [x] dedupe some of the distro ca installation code
- [x] make ca installation/uninstallation non-fatal and cleaned up in case of failure partial way through
- [x] make final call on whether to only selectively enable or just apply to all containers when custom CAs are installed in the engine (at `/usr/local/share/ca-certificates`)
   * I'm personally good with this now. These codepaths only get hit when users have a custom engine w/ CA certs installed and I made the install best-effort with cleanup in case of errors. Still open to feedback from others on this though.
- [ ] expand test coverage a bit more (more debian/rhel derivatives, cover cleanup in case of install failure)
- [ ] test ubuntu gpu image (manually for now since we don't have any automated tests for that yet)
- [x] double check umask clearing mentioned here: https://github.com/dagger/dagger/pull/7067#discussion_r1558135886
- [x] add more comments to extra gnarly/subtle parts of code

---

This adds support for automatic installation of custom CA certs to containers. Originally this was gonna just be the first step outlined [here](https://github.com/dagger/dagger/issues/6599) but as I started implementing the CA certs part it required enough complication that it seemed best to just implement the full support for automatic CA installation for all containers.

## Helpful docs/prior art
* https://wiki.archlinux.org/title/Transport_Layer_Security
* https://go.dev/src/crypto/x509/root_linux.go
* @vito's bass CA autoinstall https://github.com/vito/bass/blob/bf53c90d467e0eaf7125cfbfd005c65349125481/pkg/runtimes/shim/tls.go
* https://github.com/containers/podman/issues/2317
* https://github.com/flatpak/flatpak/pull/1757
* https://github.com/flatpak/flatpak/issues/2721
* https://p11-glue.github.io/p11-glue/p11-kit/manual/remoting.html
   * we don't do this in our implementation, but an interesting possibility for systems/libs using p11 kit
* https://minikube.sigs.k8s.io/docs/handbook/untrusted_certs/
   * https://github.com/kubernetes/minikube/blob/master/pkg/minikube/kubeconfig/settings.go#L85

## CA Certs
I learned a lot about the various nuances surrounding CAs as part of this (and I'm sure I've only scratched the surface), braindump below.

CAs are not standardized in any way; the way they are configured depends on a few (sometimes orthogonal) factors:
1. Consuming software/TLS library
   * "Generic" - includes openssl and many/most language stdlibs
     * The closest thing to a generic default is the use of a "bundle" file, which is a single file containing each CA cert that should be trusted along with a dir containing individual files for each of those CAs and symlinks of the hash of those files to the file (created with the `c_rehash` command)
     * Typically there will also be a command that can be used to automatically update the bundle file and CA dir when custom CAs are installed in another directory.
     * The location of the bundle file and other dirs and the name of the command to update CAs depends on the distro. Typically the distro will compile packages like `curl` with options that point to the right bundle file and/or CA dir to use.
   * `PKCS #11` and `p11-kit`
     * Have not gone too deep into this yet, but my limited understanding is PKCS 11 is an attempt at a standard around trust/key-management and includes CA certs. Not all distros/libraries use the standard, but those that do can load CA certs that are configured via the `trust` or `update-ca-trust` command.
     * RHEL-based distros typically use this for example
     * The `trust` command has an `extract` subcommand that creates a layout compatible with the above "generic" setup, for better compatibility with libs that don't use `libp11-kit`
   * Other TLS implementations like GnuTLS and LibreSSL do their own thing
   * The JVM has it's own CA cert store
   * Browsers typically also have their own CA store
   * Different versions of a given TLS lib may also have their own CA trust store (i.e. the alpine base image has both `/etc/ssl/certs` and `/etc/ssl1.1/certs`
2. Distro
   * As mentioned above, all the paths used for each TLS implementation (bundle file, CA dir, etc.) depend on compile-time options set by each distro when building packages
   * CA functionality is typically implemented as optional packages (rather than as pseudo-required packages like `libc`) and whether said package is installed in base images varies from distro to distro:
      * the rhel based base images seem to usually come with `ca-certificates` fully installed
      * `debian` comes without it installed
      * `alpine` does not have it installed, but nonetheless does have a base bundle file installed (though no separate cert files)

## Implementation Notes
* Automatic install
   * If the user puts certs at `/usr/local/share/ca-certificates` in their custom engine image, the engine will attempt to install that in all containers (but only if it can identify the distro of the container, fallback is a nop right now)
   * would like to leave this in, but requires that any failure during install/uninstall can be made non-fatal and cleanup-able. Otherwise will change this to be configurable somehow
* Uninstall + cleanup
   * It's important that we follow the "leave no trace" rule here; any CAs that we autoinstall in a container should be auto-*un*installed so they don't pollute the final container result unexpectedly
   * This becomes extra complicated in cases where the user's exec actually modifies the state of bundle files/CA dirs, in which case we also want to be very careful to not uninstall anything that the exec actually purposefully modified.
     * This isn't an obscure case; e.g. if you `apk install curl` off the base alpine image the `ca-certificates` package will be installed as a dep and modify the bundle file and CA dir.
* Execution location
   * At first, I implemented all this in the shim process that we currently run as the init of each container. However, this didn't end up working/was not ideal:
      1. If the container user was not root, we would not have permissions to modify certain files/dirs and couldn't run update command like `update-ca-certificates`
      2. We have a long-term goal of removing the need to inject our own init to each container (https://github.com/dagger/dagger/issues/4989), but this approach buries us even deeper into that dependency
   * Instead, we had the option to either A) run it exclusively in the part of the shim that executes outside the container or B) implement our own custom buildkit `Executor` (which runs in the `dagger-engine` process). I ended up going with the `Executor` approach for a few reasons:
     1. In the long-term, we can migrate *all* of the current shim functionality to this Executor. This should result in a much simpler overall architecture since we will be able to avoid all the messy hacks around passing special env vars to the shim process and trying to sneak uncached data into the container via `ftp_proxy`
     2. As mentioned below, ca cert install/uninstall may need to run actual commands as root inside the container, which is much simpler to implement from the Executor relative to the shim (which would need to have new logic for creating a separate bundle for each command, whereas the Executor already has all that logic)
   * The biggest annoyance here is that we need to read/write certain files/dirs from the container before/after the container has actually run, which meant either A) creating all the container mounts temporarily (and unmounting them before the container starts to avoid various corner cases around multiple mounts of overlay) or B) a "vfs" approach that maps read/write operations on files/dirs to the correct underlying source mount even though those haven't been mounted yet.
     * I went with B) because A) is a bit too frought with weird cases and felt like more performance overhead (mount/unmount can be expensive in some cases). This is implement in `cmd/engine/cacerts/fs.go` as `containerFS`. The main complication comes from needing to *very* carefully handle symlinks so that any symlinks in parent dirs of a given path are always resolved correctly to the final mount where they actually point to.
* Commands 
   * Have to support running command like `update-ca-certificates` and similar by actually starting separate container that reuses the same rootfs+mounts as the "actual" user exec container.
   * This is done from the Executor level so there's no extra layers and overhead is relatively minimal, but still does impose some latency unfortunately. This currently only happens when the engine actually has any custom certs installed, so it's relatively safe and won't slow down all users immediately.
   * I think we can optimize this in time to skip running these commands in cases where we are sure we can manually implement the same functionality via file/dir operations directly. Went with this approach for now to be extra safe to start.
* Distro detection
   * We need to positively identify which distro is in use because some don't come with a base image containing any bundle file/ca dir/ etc. (e.g. `debian`), but other packages may still try to check those (we need to manually set those up in these cases).
   * This requires checking various `/etc/` files. Used https://github.com/dekobon/distro-detect/blob/master/linux/distrochecks.go as a reference
   * Currently only implemented support for alpine/debian-derivatives/rhel-derivatives

### DIFF
--- a/cmd/engine/executor.go
+++ b/cmd/engine/executor.go
@@ -1,0 +1,1045 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/containerd/console"
+	"github.com/containerd/containerd/content/local"
+	"github.com/containerd/containerd/diff/apply"
+	"github.com/containerd/containerd/diff/walking"
+	ctdmetadata "github.com/containerd/containerd/metadata"
+	"github.com/containerd/containerd/mount"
+	containerdoci "github.com/containerd/containerd/oci"
+	"github.com/containerd/containerd/platforms"
+	ctdsnapshot "github.com/containerd/containerd/snapshots"
+	"github.com/containerd/continuity/fs"
+	runc "github.com/containerd/go-runc"
+	"github.com/dagger/dagger/cmd/engine/cacerts"
+	"github.com/docker/docker/pkg/idtools"
+	"github.com/moby/buildkit/cache/metadata"
+	"github.com/moby/buildkit/executor"
+	"github.com/moby/buildkit/executor/oci"
+	resourcestypes "github.com/moby/buildkit/executor/resources/types"
+	gatewayapi "github.com/moby/buildkit/frontend/gateway/pb"
+	randid "github.com/moby/buildkit/identity"
+	containerdsnapshot "github.com/moby/buildkit/snapshot/containerd"
+	"github.com/moby/buildkit/solver/pb"
+	"github.com/moby/buildkit/util/bklog"
+	"github.com/moby/buildkit/util/leaseutil"
+	"github.com/moby/buildkit/util/network"
+	"github.com/moby/buildkit/util/network/netproviders"
+	"github.com/moby/buildkit/util/stack"
+	"github.com/moby/buildkit/util/winlayers"
+	"github.com/moby/buildkit/worker/base"
+	wlabel "github.com/moby/buildkit/worker/label"
+	workerrunc "github.com/moby/buildkit/worker/runc"
+	"github.com/moby/sys/signal"
+	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+	bolt "go.etcd.io/bbolt"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/semaphore"
+)
+
+func NewWorkerOpt(
+	root string,
+	snFactory workerrunc.SnapshotterFactory,
+	processMode oci.ProcessMode,
+	labels map[string]string,
+	idmap *idtools.IdentityMapping,
+	nopt netproviders.Opt,
+	dns *oci.DNSConfig,
+	apparmorProfile string,
+	selinux bool,
+	parallelismSem *semaphore.Weighted,
+	traceSocket,
+	defaultCgroupParent string,
+) (base.WorkerOpt, error) {
+	var opt base.WorkerOpt
+	name := "runc-" + snFactory.Name
+	root = filepath.Join(root, name)
+	if err := os.MkdirAll(root, 0700); err != nil {
+		return opt, err
+	}
+
+	np, npResolvedMode, err := netproviders.Providers(nopt)
+	if err != nil {
+		return opt, err
+	}
+
+	exe, err := New(Opt{
+		Root:                filepath.Join(root, "executor"),
+		Cmd:                 "/usr/local/bin/dagger-shim",
+		ProcessMode:         processMode,
+		IdentityMapping:     idmap,
+		DNS:                 dns,
+		ApparmorProfile:     apparmorProfile,
+		SELinux:             selinux,
+		TracingSocket:       traceSocket,
+		DefaultCgroupParent: defaultCgroupParent,
+	}, np)
+	if err != nil {
+		return opt, err
+	}
+	s, err := snFactory.New(filepath.Join(root, "snapshots"))
+	if err != nil {
+		return opt, err
+	}
+
+	localstore, err := local.NewStore(filepath.Join(root, "content"))
+	if err != nil {
+		return opt, err
+	}
+
+	db, err := bolt.Open(filepath.Join(root, "containerdmeta.db"), 0644, nil)
+	if err != nil {
+		return opt, err
+	}
+
+	mdb := ctdmetadata.NewDB(db, localstore, map[string]ctdsnapshot.Snapshotter{
+		snFactory.Name: s,
+	})
+	if err := mdb.Init(context.TODO()); err != nil {
+		return opt, err
+	}
+
+	c := containerdsnapshot.NewContentStore(mdb.ContentStore(), "buildkit")
+
+	id, err := base.ID(root)
+	if err != nil {
+		return opt, err
+	}
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown"
+	}
+	xlabels := map[string]string{
+		wlabel.Executor:       "oci",
+		wlabel.Snapshotter:    snFactory.Name,
+		wlabel.Hostname:       hostname,
+		wlabel.Network:        npResolvedMode,
+		wlabel.OCIProcessMode: processMode.String(),
+		wlabel.SELinuxEnabled: strconv.FormatBool(selinux),
+	}
+	if apparmorProfile != "" {
+		xlabels[wlabel.ApparmorProfile] = apparmorProfile
+	}
+
+	for k, v := range labels {
+		xlabels[k] = v
+	}
+	lm := leaseutil.WithNamespace(ctdmetadata.NewLeaseManager(mdb), "buildkit")
+	snap := containerdsnapshot.NewSnapshotter(snFactory.Name, mdb.Snapshotter(snFactory.Name), "buildkit", idmap)
+	md, err := metadata.NewStore(filepath.Join(root, "metadata_v2.db"))
+	if err != nil {
+		return opt, err
+	}
+
+	opt = base.WorkerOpt{
+		ID:               id,
+		Labels:           xlabels,
+		MetadataStore:    md,
+		NetworkProviders: np,
+		Executor:         exe,
+		Snapshotter:      snap,
+		ContentStore:     c,
+		Applier:          winlayers.NewFileSystemApplierWithWindows(c, apply.NewFileSystemApplier(c)),
+		Differ:           winlayers.NewWalkingDiffWithWindows(c, walking.NewWalkingDiff(c)),
+		ImageStore:       nil, // explicitly
+		Platforms:        []ocispecs.Platform{platforms.Normalize(platforms.DefaultSpec())},
+		IdentityMapping:  idmap,
+		LeaseManager:     lm,
+		GarbageCollect:   mdb.GarbageCollect,
+		ParallelismSem:   parallelismSem,
+		MountPoolRoot:    filepath.Join(root, "cachemounts"),
+	}
+	return opt, nil
+}
+
+type Opt struct {
+	// root directory
+	Root string
+	Cmd  string
+	// DefaultCgroupParent is the cgroup-parent name for executor
+	DefaultCgroupParent string
+	// ProcessMode
+	ProcessMode     oci.ProcessMode
+	IdentityMapping *idtools.IdentityMapping
+	// runc run --no-pivot (unrecommended)
+	NoPivot         bool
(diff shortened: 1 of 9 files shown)


cmd/shim/main.go
```diff
@@ -41,6 +41,11 @@ import (
 	"github.com/dagger/dagger/telemetry"
 )
 
+func init() {
+	// unset umask so permissions are created as specified
+	unix.Umask(0)
```
sipsma, line 46: This is needed because previously we had a umask of `0o22`, which resulted in any files created here never giving `w` perms to group+other, even if created as `0o666`. This became a problem now because if we need to run ca certs update commands as root initially but the final container used a non-root user, then the shim failed at trying to open the `stdout/stderr` files for write.

There may be other ways of addressing this but afaict there shoudn't be any harm in unsetting the umask here. The only thing to double check still is whether the umask here actually propagates to the final container init process, in which case this would be a breaking change (*technically*) and is probably better avoid.

sipsma: Double-checked this, `umask` is actually in oci runtime spec so it's important we honor it. So will need a different approach; I think we could just set the owner of those stdout/stderr files to be that of the final container uid instead, which should be relatively simple still.

sipsma: Fixed here https://github.com/dagger/dagger/pull/7067/commits/0591a1ff832fd8dd385ff6197c0379363284fe4d

vito:
> Uninstall + cleanup
> 
>  [...]

Oof, yeah that sounds rough. Is there any way we can isolate "our" changes to a separate layer and then rebase it out somehow? (Haven't looked how it works yet, maybe it's not so bad, just thinking out loud)

> Commands
> 
>  [...]

Is the `update-ca-certificates` operation cached? i.e. if I repeatedly use the same base image, will it only have to have run once?

Or does it install-then-uninstall on each `withExec` or something like that? (Guessing that's the case based on the description?)

Would it be too weird to install certs as part of `.from` and then somehow "drop" the layer at export/publish time? (Building on the previous question)

sipsma:
> > Uninstall + cleanup
> > [...]
> 
> Oof, yeah that sounds rough. Is there any way we can isolate "our" changes to a separate layer and then rebase it out somehow? (Haven't looked how it works yet, maybe it's not so bad, just thinking out loud)

Not that I am aware of. At the beginning I was trying to use overlay mounts to do that but quickly realized that:
1. overlay is not always available
2. we'd need to do some annoying stuff with making sure that we always use the right set of overlay options (which can vary host to host and on the setup, doable but complicated)
3. it doesn't work for bundle files since those can be re-written with new contents by the user's exec but overlay's granularity is on the whole-file level (whereas we need to look at the "diff" in terms of contents in this case)

> > Commands
> > [...]
> 
> Is the `update-ca-certificates` operation cached? i.e. if I repeatedly use the same base image, will it only have to have run once?
> 
> Or does it install-then-uninstall on each `withExec` or something like that? (Guessing that's the case based on the description?)

The update commands have to re-run on each `withExec`, yeah; but:
1. they will be cached in the same way the `withExec` is cached since buildkit's solver doesn't know about any of this
2. changing the CA certs won't invalidate the cache (mainly matters for remote cache cases)
3. there's only one layer for the exec overall; we aren't adding more layers each time these update commands run
4. this will all only get triggered if the user has a custom engine with custom CAs installed, so won't leak out to every user yet

> Would it be too weird to install certs as part of `.from` and then somehow "drop" the layer at export/publish time? (Building on the previous question)

I considered that too but ruled it out since I want to avoid leaving those files around even in the intermediate layers. Leaving them in will still leak them to the layers of the published image (even if we append an extra layer on top cleaning them up). 
* There's also not really any ways to hack around that since the snapshotter correctly prevents you from modifying layers after they are committed.

I think the best we can do is optimize in follow-ups to not always run these update commands and instead selectively just directly fixup files/dirs in the same way the update command does. We can obviously only do this when we are really _**really**_ sure we know what the update command is (i.e. probably have to match on distro versions too), so still need to support running these commands in each `withExec` for other cases.

cmd/engine/executor.go
```diff
@@ -0,0 +1,1053 @@
...
+
+	if err := json.NewEncoder(f).Encode(spec); err != nil {
+		return err
+	}
+	f.Close()
+
+	bklog.G(ctx).Debugf("> creating %s %v", id, process.Meta.Args)
+	defer bklog.G(ctx).Debugf("> container done %s %v", id, process.Meta.Args)
+
+	if installCACerts {
```
sipsma, line 449: Just a note for reviewers, this code is mostly the same as upstream's runc executor except for this section that handles CA cert installation and some other minor re-arrangement/cleanup of irrelevant codepaths.

core/integration/cacert_test.go
```diff
@@ -0,0 +1,491 @@
...
+			WithWorkdir("/src").
+			WithMountedFile("/ca.crt", caRootCert).
+			WithMountedFile("/server.crt", serverCert).
+			WithMountedFile("/server.key", serverKey).
+			WithMountedFile("/dhparam.pem", dhParam).
+			WithServiceBinding("engine", devEngine.AsService()).
+			WithMountedFile("/bin/dagger", daggerCliFile(t, c)).
+			WithEnvVariable("_EXPERIMENTAL_DAGGER_CLI_BIN", "/bin/dagger").
+			WithEnvVariable("_EXPERIMENTAL_DAGGER_RUNNER_HOST", "tcp://engine:1234").
+			WithEnvVariable(executeTestEnvName, "ya").
```
vito, line 129: :grin:

cmd/engine/executor.go
```diff
@@ -0,0 +1,1053 @@
...
+				// leaving the container in some weird half state
+				return fmt.Errorf("failed to install cacerts: %w", err)
+			}
+			// if install failed but we were able to cleanup, then we should log it but don't need to fail the exec
+			if err != nil {
+				bklog.G(ctx).Errorf("failed to install cacerts but successfully cleaned up, continuing without CA certs: %v", err)
+			} else {
+				defer func() {
+					if err := caInstaller.Uninstall(ctx); err != nil {
+						bklog.G(ctx).Errorf("failed to uninstall cacerts: %v", err)
```
vito, line 491: Paranoia brain: if uninstall fails, should we fail the whole thing for the same reason as the cleanup failure?

Worst case I guess is that we "leak" a CA cert into the container data? But that's not really a sensitive value, just a little odd to leave it behind

vito (approved):
Verified tests fail if I comment out an important line (the one that installs the CA into the engine. Obviously there's a lot going on here but it's fully justified afaict. :laughcry:

cmd/shim/main.go
```diff
@@ -396,6 +404,39 @@ func setupBundle() int {
...
+			} {
+				path := filepath.Join(mnt.Source, strings.TrimPrefix(metaPath, metaMountPath))
+				if err := containerTouchPath(path, 0o666, &spec); err != nil {
+					panic(fmt.Errorf("error touching path %s: %w", path, err))
+				}
+
+				// for stderr specifically, also update errWriter to that file so that any
+				// errors here actually show up in the final error message passed to clients
+				if metaPath == stderrPath {
+					stderrFile, err := os.OpenFile(path, os.O_WRONLY|os.O_APPEND, 0o666)
```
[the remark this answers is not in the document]
sipsma: Sure probably doesn't hurt. I'm waiting to take off on a plane rn, 6 hour flight so should have plenty of time to push a fix provided wifi is usable 😊

cmd/engine/executor.go
sipsma: Yeah I wasn't totally sure here, but in retrospect it's safer to start with an error and make non fatal than vice versa

sipsma:
Just FYI that I was adding more tests last night and did find some cases where we were incorrectly leaving around some parent dirs. This is pretty minor, so should be fine to address in follow up.

I am getting close to finishing the fix there but am mentioning this since apparently my planes wifi is gonna cut off once we are too far into the pacific ocean... so we'll see if I have enough time to fix it before my vacation actually starts. Probably should work out.

jedevc:
I'm gonna take this out of the milestone - given @sipsma is now on holiday, I'd rather we wait to merge this.

jedevc:
I'm gonna take this out of the milestone - given @sipsma is now on holiday, I'd rather we wait to merge this.

sipsma:
Fixed the aforementioned issue of leaving parent dirs and not resetting mtime on files (which results in the diff being non-empty when it should be empty) and added test coverage for all that. Will merge this once green. There's going to be a few more immediate follow ups to add `GOPROXY` and http proxy support too.

sipsma:
The test failure is https://github.com/dagger/dagger/issues/6111 which is being debugged separately here: https://github.com/dagger/dagger/pull/7128 So gonna merge as is rather than keep rerolling forever

PR #6916, opened 2024-03-21
[WIP] Finish isolating sessions to each client

sipsma:
This is the last part of the saga around https://github.com/dagger/dagger/issues/6747 that fully isolates sessions to each client and thus enables support for passing host unix sockets to modules, plugging some existing holes in our secret isolation, etc.

This PR doesn't have all that yet, been in the process of rebuilding it cleanly after an initial messy prototype. Sending it out anyways because what I have here seems to fix the problem in https://github.com/dagger/dagger/pull/6914, so may want to merge a minimum safe changeset here early.
* The only extra thing strictly needed to safely merge is isolating the secret store to each client; in the current state of things nested execs end up with a shared secret store from their parent caller, which is *probably* a breaking change (though it's complicated, not 100% sure)
* Brief description of changes so far:
   * Consolidate module caller digest to just be the client ID. This is a pretty nice simplification from before since we less IDs to think about and pass around, but especially simplifies the rest of the pending changes
   * Share the server with nested execs, whereas previously they got their own. This is again just a generally nice simplification made possible by the fact that we can now isolate their session attachables.

@vito unfortunately these are the changes I was referring to that may cause rebase headaches with your OTEL PR (and there's more coming on top of the commits already here), so we'll have to sort that out if we merge this soon.

PR #6914, opened 2024-03-21
namespace services by server, not by client

vito:
Previously it was possible to start a dependent service in one module API call, and then use it again in a later call, only to have it fail because it cannot resolve the service address, even though it's still running. This happened because each invocation has its own client ID, and client IDs were used to build service addresses. ([Repro](https://github.com/camptocamp/daggerverse/blob/95e15b1c7931c4072e0255791902fe293b0e06df/presentation/main.go))

This change brings service addresses into alignment with the recent change to uniq them by service ID instead of client ID. The overall effect is that services are deduped within a Dagger invocation, even across module calls. So with this change, the service will just stay running and be re-used by a later call, thanks to the grace period.

PR #6505, opened 2024-01-26
fix: dedupe uncached metadata across a session

jedevc:
Fixes #6501.

Thank goodness it's Friday, lol.

The root cause of #6501 is that every single ExecOp gets a unique digest. This means that the buildkit solver is not able to dedupe them, and instead, to ensure it only runs once, we rely on the edge merging behavior. This was kiiiinda hacky, and not super great - edge merging is supposed to be for *different* vertexes that happen to produce similar output (detected by their cache keys). We were using it for pretty much everything.

Since https://github.com/moby/buildkit/commit/e1da8b7f82f09d3e622a98961f1483f613852390 (introduced in #6481), the progress behavior changes - edge merges now duplicate the progress. This makes sense! Because edge merges are for "coincidentally" similar edges, the progress should occur for both.

The trick here is to stop relying on this edge merging behavior, by making more ExecOps have a similar digest. The cause of the different digests is the server ids and the parent client ids, so we make an explicit effort to *not modify these* these for ExecOps that were previously the same, by using a map to cache previous metadata. Because the cache map is on the `Client`, then this effectively dedupes across only the session - different sessions don't get deduped with each other, which is fine, because they would have different server ids anyways! (and in this case, we actually want the edge merge to happen, so we get progress to both! see https://github.com/moby/buildkit/issues/4330) 

> [!NOTE]
> This does seem like a bizarre trick to pull. Is it actually safe to dedupe these like this?
> However, we were doing it similar via edge-merges, this just forces the merge to happen at the buildkit solver level, which is much much easier.

PR #7068, opened 2024-04-09
telemetry: record an API call's LLB op digests

vito:
This data allows the UI to track where lazy operations came from so we can tie them back to the call site that installed them (i.e. `withExec`). This visualization angle is super useful when you're trying to gauge the "cost" of API calls.

Something like this (note the "shadow bars" - UI subject to change):

image

**NOTE:** As a prerequisite, I moved the `ftp_proxy` hackery back to `Container.WithExec` instead of doing it "just-in-time" in the Buildkit client. Otherwise the span ends up with a different digest than what we return, so they can't be correlated at all. The switch from Progrock => OpenTelemetry allows us to do this, I believe; we couldn't before because we had to pass the Progrock parent vertex ID, but the OpenTelemetry equivalent is handled for us by Buildkit itself.

(edit - side note: it seems like we could potentially remove the `ftp_proxy` hack _entirely_ if we replace ServerID with the OpenTelemetry trace + span ID.)

### DIFF
--- a/engine/buildkit/client.go
+++ b/engine/buildkit/client.go
@@ -285,57 +285,6 @@ func (c *Client) Solve(ctx context.Context, req bkgw.SolveRequest) (_ *Result, r
 	var llbRes *bkfrontend.Result
 	switch {
 	case req.Definition != nil && req.Definition.Def != nil:
-		clientMetadata, err := engine.ClientMetadataFromContext(ctx)
-		if err != nil {
-			return nil, err
-		}
-
-		dag, err := DefToDAG(req.Definition)
-		if err != nil {
-			return nil, err
-		}
-		if err := dag.Walk(func(dag *OpDAG) error {
-			execOp, ok := dag.AsExec()
-			if !ok {
-				return nil
-			}
-			if execOp.Meta == nil {
-				execOp.Meta = &bksolverpb.Meta{}
-			}
-			if execOp.Meta.ProxyEnv == nil {
-				execOp.Meta.ProxyEnv = &bksolverpb.ProxyEnv{}
-			}
-
-			// If we already solved for this execop in this dagger session, use
-			// it's uncached metadata to preserve the same vertex digest.
-			// This results in buildkit's solver de-duping the op, instead of
-			// the scheduler's edge merge.
-			c.execMetadataMu.Lock()
-			execMeta, ok := c.execMetadata[*execOp.OpDigest]
-			if !ok {
-				execMeta = ContainerExecUncachedMetadata{
-					ParentClientIDs: clientMetadata.ClientIDs(),
-					ServerID:        clientMetadata.ServerID,
-				}
-				c.execMetadata[*execOp.OpDigest] = execMeta
-			}
-			c.execMetadataMu.Unlock()
-
-			var err error
-			execOp.Meta.ProxyEnv.FtpProxy, err = execMeta.ToPBFtpProxyVal()
-			if err != nil {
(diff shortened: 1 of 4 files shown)


engine/buildkit/client.go
```diff
@@ -285,57 +285,6 @@ func (c *Client) Solve(ctx context.Context, req bkgw.SolveRequest) (_ *Result, r
...
-				execOp.Meta = &bksolverpb.Meta{}
-			}
-			if execOp.Meta.ProxyEnv == nil {
-				execOp.Meta.ProxyEnv = &bksolverpb.ProxyEnv{}
-			}
-
-			// If we already solved for this execop in this dagger session, use
-			// it's uncached metadata to preserve the same vertex digest.
-			// This results in buildkit's solver de-duping the op, instead of
-			// the scheduler's edge merge.
```
jedevc, line 312: So this was originally added because of https://github.com/dagger/dagger/pull/6505.

Because we're not using buildkit's progress stream this shouldn't be an issue anymore right?

We still would be doing more edge merging again because of this - since these are likely what cause inconsistent graph state issues, we might see more of those (though that would definitely help debugging it lol).

If there's an easy way to keep this hack in (I like the explicitness of it, instead of hiding the details in the solver), I'd prefer that, but not a blocker.

> (edit - side note: it seems like we could potentially remove the ftp_proxy hack entirely if we replace ServerID with the OpenTelemetry trace + span ID.)

I like this - this is the best of all worlds - we get this feature, and we get rid of our nasty hack.

jedevc (approved):

engine/buildkit/client.go
vito: > I like this - this is the best of all worlds - we get this feature, and we get rid of our nasty hack.

Yeah, I really want to do this, but it probably makes the most sense to do it after #6914 and #6916 (cc @sipsma) since at that point we for sure don't need the `ParentClientIDs` value, and solely have the `ServerID` to figure out. At that point theoretically we just stop using a random ID for it and use something analogous to `$traceparent` instead. (Carefully pointing out we need trace ID + span ID, not just trace ID, to handle nesting.)

We'd be adding a hard dependency on OpenTelemetry at this layer, but that seems worth the trade-off to me.

sipsma: Have invested 0 time looking myself because on mobile atm, but how is the otel trace id propagated? Not necessarily opposed to using it as the server id but if we can just propagate the server id the same way we propagate the trace id that might be even better (just in the interest of not creating too much entanglement).

Also, the approach I'm adding in the CA cert PR of using a custom Executor rather than the shim should in the long term let us migrate everything in the shim and also be another route of avoiding the ftp proxy hack. Just noting for completeness.

vito: @sipsma We can almost certainly use the new custom Executor if we want to yeah. Right now Buildkit propagates it for us with its OpenTelemetry integration - somewhere around [here](https://github.com/moby/buildkit/blob/647a997b389757068760410053873745acabfc80/executor/oci/spec.go#L125) I suppose (I never had to look).

FWIW, it seems worth considering making ServerID == TraceID+SpanID regardless. I briefly considered just doing it along the way, but had enough change going on and didn't bother. It doesn't seem like we gain a whole lot by making up a separate random string, vs. if we make it match, we can automatically correlate a ServerID to a trace for "free" which could come in handy when going through logs and such. Low conviction on that though, wanted to ask and now's my chance. :grin:

PR #7069, opened 2024-04-10
feat: added support for streaming stdout when --progress=plain

### LINKED ISSUE #7057: 🐞 `--progress=plain` stopped streaming logs with Dagger v0.11.0
### What is the issue?

While developing locally, I often use --progress=plain for long-running tasks. This allows me to better understand what's happening and to detect early warnings. With prior releases, this worked well, as the logs were streamed to my console and I could follow along with the build. However, it appears that there's a regression with v0.11.0. Now, when I run any dagger --progress=plain call ... command, I only see Connected to engine 14def8fb732a (version v0.11.0) for an extended period. The remainder of the log only displays after the job is completed.

### Dagger version

dagger v0.11.0 (registry.dagger.io/engine) darwin/arm64

### Steps to reproduce

I've set up https://github.com/skycaptain/dagger-issue-7057 as minimal reproducible example. Run with `dagger call container-echo` and `dagger --progress=plain call container-echo` and watch that with the first command the logs are streamed, while their not with the second.

### Log output

_No response_

morlay:
fix #7057

### DIFF
--- a/dagql/idtui/streaming_output.go
+++ b/dagql/idtui/streaming_output.go
@@ -1,0 +1,184 @@
+package idtui
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"github.com/vito/progrock/ui"
+
+	"os"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/dagger/dagger/telemetry/sdklog"
+	"github.com/muesli/termenv"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type streamingExporter struct {
+	spanNames sync.Map
+
+	// incr idx for spanID
+	idx                 int64
+	batchPrinterGetters sync.Map
+	batchPrinterWg      sync.WaitGroup
+
+	output   *termenv.Output
+	outputMu sync.RWMutex
+
+	frameTicker *time.Ticker
+
+	done     chan struct{}
+	doneOnce sync.Once
+}
+
+func newStreamingExporter() *streamingExporter {
+	return &streamingExporter{
+		output:      termenv.NewOutput(os.Stdout, termenv.WithProfile(ui.ColorProfile()), termenv.WithUnsafe()),
+		done:        make(chan struct{}),
+		frameTicker: time.NewTicker(50 * time.Millisecond),
+	}
+}
+
+func (t *streamingExporter) Shutdown(ctx context.Context) error {
+	t.doneOnce.Do(func() {
+		t.frameTicker.Stop()
+		close(t.done)
+	})
+
+	// wait all printers outputs
+	t.batchPrinterWg.Wait()
+	return nil
+}
+
+func (t *streamingExporter) ExportSpans(ctx context.Context, spans []sdktrace.ReadOnlySpan) error {
+	for _, span := range spans {
+		t.spanNames.Store(span.SpanContext().SpanID(), span.Name())
+	}
+	return nil
+}
+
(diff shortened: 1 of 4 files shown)


morlay:
image

logs in real CI. (no idea the color lost issue)

dagql/idtui/streaming_output.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func (t *streamingExporter) ExportLogs(ctx context.Context, logs []*sdklog.LogData) error {
+	for _, log := range logs {
+		t.export(log)
+	}
+	return nil
+}
+
+func (t *streamingExporter) getSpanName(spanID trace.SpanID) string {
```
morlay, line 70: the span name is not always get cause otel emit the spans until span.End() called.
may consider set span name in log attrs too.

jedevc:
I think this makes sense :tada: I'm really not familiar with how `frontend.go` works though.

To me, it feels confusing that the logic for adding new views/etc is really complex - ideally, we'd have an `interface` that a new view would be able to implement. Then the fancy new progress view and the plain one would all just be implementations of that interface.

jedevc:
There were still some color issues (particularly in the summary that's still printed at the end), so pushed a fix for that (the extra `WithTTY` calls can be removed once https://github.com/vito/progrock/pull/14 is merged and updated).

jedevc:
I'd like to do some further refactoring of this later, but for now, this fixes the issue, and it would be awesome to include this in v0.11.1, so going to approve now, and I'll happily iterate on this after the release.

dagql/idtui/streaming_output.go
```diff
@@ -0,0 +1,184 @@
...
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	w.lines = append(w.lines, line)
+}
+
+func (w *batchPrinter) render(out *termenv.Output, final bool) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
```
morlay, line 157: @jedevc 

it could ignore logs without spanName for now.
```
spanName := w.getSpanName(w.spanID)
if spanName == "" {
    return
}
```

morlay:
@jedevc spanName may empty.

image

it could safe fixed by  https://github.com/dagger/dagger/pull/7069/files#r1559461303

jedevc:
@morlay can you share an example of where span is empty?

morlay:
> @morlay can you share an example of where span is empty?

this first `6:` in that image.

It logs before span emitted.

jedevc:
Right, yeah, I meant, do you have a publicly-available example I can run to get this?

I'm trying to understand why this can happen.

vito:
> To me, it feels confusing that the logic for adding new views/etc is really complex - ideally, we'd have an `interface` that a new view would be able to implement. Then the fancy new progress view and the plain one would all just be implementations of that interface.

This view is fundamentally different so it makes sense to me. Normally we're rendering into a viewport and dynamically redrawing parts of the screen, TUI style. This one just needs to print output directly without any funny business and let it go into the scrollback infinitely, skipping past all those moving parts and disregarding screen size and all that.

In the end they're both implementing an interface, just perhaps a lower level one than you might have expected (OpenTelemetry span/log exporter).

morlay:
> Already is - if you pass `-vvv` (verbosity level 3) it'll show everything. Probably need that in our docs. :)

but with it, the final reporter is too noisy. 😂
May some option could define the should show.

vito:
> but with it, the final reporter is too noisy. 😂
> May some option could define the should show.

I'd prefer to avoid a bunch of flags, but feel free to make it a param or var instead of a const, if you're using this code directly.

PR #7070, opened 2024-04-10
fix: avoid lost log colors when CI=true

### LINKED ISSUE #7047: [v0.11.0] dagger logs lost colors in CI env, without setting NO_COLOR
### What is the issue?

logs lost colors

### Dagger version

dagger v0.11.0 

### Steps to reproduce

_No response_

### Log output

image

morlay:
fix #7047

### DIFF
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -87,7 +87,7 @@ func New() *Frontend {
 		view:         new(strings.Builder),
 		messagesView: logsView,
 		messagesBuf:  logsOut,
-		messagesW:    ui.NewOutput(io.MultiWriter(logsView, logsOut), termenv.WithProfile(profile)),
+		messagesW:    ui.NewOutput(io.MultiWriter(logsView, logsOut), termenv.WithProfile(profile), termenv.WithUnsafe()),
 	}
 }
 


dagql/idtui/frontend.go
```diff
@@ -87,7 +87,7 @@ func New() *Frontend {
 		view:         new(strings.Builder),
 		messagesView: logsView,
 		messagesBuf:  logsOut,
-		messagesW:    ui.NewOutput(io.MultiWriter(logsView, logsOut), termenv.WithProfile(profile)),
+		messagesW:    ui.NewOutput(io.MultiWriter(logsView, logsOut), termenv.WithProfile(profile), termenv.WithUnsafe()),
```
jedevc, line 90: Can we do `termenv.WithTTY(true)`? Looks like `WithUnsafe` has other implications that aren't just color-related.

---

Also, how did this work before? `ui.NewOutput` states:

```
// NewOutput returns a termenv.Output that will always use color, regardless of
// whether w is a TTY, unless NO_COLOR is explicitly set.
```

We should update this implementation here, but also, I can't work out why this would have stopped working before? It's not like previously the writer (returned from `NewVterm`) was a valid TTY (it doesn't implement `termenv.File`).

jedevc: Oh wait, this never worked - we just never used this codepath, we used the progrock output directly.

jedevc:
Even with this PR I still don't get colors?

```
❯ dagger init --sdk=go --source=. &> out.txt </dev/null
❯ cat out.txt
Connected to engine 612a4a50bcc2 (version 4b825dc642cb6eb9a060e54bf8d69288fbee4904)
✔ connect 0.2s

Error: module already exists                                                                                                                                                                                                                                                                                

Error: module already exists
```

We'd need to update this: https://github.com/morlay/dagger/blob/d199aa438e39afe1188a0fc547bf7dec7aabacff/dagql/idtui/frontend.go#L210

dagql/idtui/frontend.go
vito: Yeah `ui.NewOutput` and `ui.ColorProfile` are _supposed_ to handle this on their own - it's the whole reason they exist. :thinking:

`ui.ColorProfile` decides whether to use color solely based on the `NO_COLOR` env var, and `ui.NewOutput` uses it internally. So we shouldn't need to pass any more settings to get colors without a TTY, but if that's not working, they must have broken somehow.

Also this `messagesW` is just be for log messages, not for the whole UI - the `ui.ColorProfile()` above should be what determines everything.

jedevc: Yeah, so I think what we should do is replace this PR with https://github.com/vito/progrock/pull/14.

Then, I just updated #7069 to use that, and we get color back everywhere (trying to get everything in place across two different PRs was getting fiddly).

vito: @jedevc Thanks for the fix! At this point though we can probably just copy-paste your fixed version into Dagger; this is one of the last few things forcing a dependency on Progrock, would be nice to just get rid of it entirely.

jedevc: Will do!

PR #7071, opened 2024-04-10
Add s390x and ppc64le support

lysliu:
Add s390x and ppc64le support
ignore the release that not exist

### DIFF
--- a/.goreleaser.common.yml
+++ b/.goreleaser.common.yml
@@ -20,8 +20,21 @@ builds:
       - amd64
       - arm64
       - arm
+      - s390x
+      - ppc64le
     goarm:
       - "7"
+    ignore:
+      - goos: darwin
+        goarch: arm
+      - goos: darwin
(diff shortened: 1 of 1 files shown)


jedevc:
cc @gerhard curious on your take here.

Personally, I'm not familiar with these architectures - and I know from past experience that getting infra to test these on is a bit of a pain. Additionally, if there are bugs specific to these architectures, they'll be very difficult to track down for a very small number of users. I'd like to be able to provide the same level of support for all our binary artifacts, but I'm not sure if we'd able to for these (we've already seen issues with our windows releases, and those aren't even that niche!)

I don't think it would be that much of an ask to get users to custom-compile for these architectures? Maybe our docs need some more information on that?

jpadams:
@lysliu I'd love to know more to see if there is a way for us to support this. Please reach out to me jeremy@dagger.io.

PR #7072, opened 2024-04-10
ci: forward host docker dir into tests

jedevc:
See https://github.com/dagger/dagger/pull/6843#discussion_r1552464269 (this behavior was originally removed, this PR just restores it to the prior state)

This ensures that we use the host creds in the test container, so we don't end up maxing out our anonymous usage of docker hub.

We may be able to get rid of this once #6916 lands? But in the meantime, we're seeing a ton of ephemeral failures which are quite annoying :tada:

### DIFF
--- a/ci/mage/util/docker.go
+++ b/ci/mage/util/docker.go
@@ -1,0 +1,24 @@
+package util
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+func HostDockerDir() (string, error) {
+	if runtime.GOOS != "linux" {
+		// doesn't work on darwin, untested on windows
+		return "", fmt.Errorf("cannot get docker dir on %s", runtime.GOOS)
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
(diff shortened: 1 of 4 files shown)


ci/mage/util/docker.go
```diff
@@ -0,0 +1,24 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+func HostDockerDir() (string, error) {
+	if runtime.GOOS != "linux" {
+		// doesn't work on darwin, untested on windows
+		return "", fmt.Errorf("cannot get docker dir on %s", runtime.GOOS)
+	}
```
helderco, line 14: I've made it work in macOS for a long time when this was first added, I don't remember if it was the new multilanguage engine or modules. One thing I had to do though was disable the osxkeychain credential helper (`credsStore`) and put my token directly in the `~/.docker/config.json` file (under `auths`) so it get's accepted by both the macOS host and dagger on the linux vm.

So in this case, mounting my `~/.docker` dir may work, but not with this check here.

helderco (approved):

helderco:
Just going to merge for the benefit of open PRs.

PR #3785, opened 2022-11-10
Include kernel logs in gha output.

sipsma:
Signed-off-by: Erik Sipsma <erik@sipsma.dev>

There's a small chance this might be helpful for debugging this issue: https://github.com/dagger/dagger/issues/3784

And in general it doesn't hurt to have this in case we run into other weird issues in the future.

Issue #3784, opened 2022-11-10
Flaky Test Timeout on TestContainerMultiPlatformExport


sipsma:
There have been two timeouts of this test recently:
* https://github.com/dagger/dagger/actions/runs/3416253271/jobs/5686204381
* https://github.com/dagger/dagger/actions/runs/3439384759/jobs/5736692454

It's not clear whether this is unique to this test case yet or not, but those are the examples we have of it happening.

Both occurences are getting blocked on the same line, waiting for an export to happen.

More interestingly (but possibly a red herring) is that they both may be blocked not on the export itself but building the shim to use in an exec that's needed to build the container being exported. Specifically, in both progress outputs, this is one of the last outputs and in both cases it never completes:
```
[build 6/6] RUN CGO_ENABLED=0 go build -o /_shim -ldflags '-s -d -w' .
```
Also, in the newer occurrence we have access to the buildkitd logs which shows that exec'ing that container is the last thing to happen before radio silence (besides health checks):
```
creating tuiqcoecgb7wx1y5fzt384rc4 [/dev/.buildkit_qemu_emulator /bin/sh -c CGO_ENABLED=0 go build -o /_shim -ldflags '-s -d -w' .]" span="[build 6/6] RUN CGO_ENABLED=0 go build -o /_shim -ldflags '-s -d -w' ."
```

---

Not totally sure what this could be. Use of qemu emulation certainly has been a source of [weird bugs in the past w/ buildkit](https://github.com/moby/buildkit/search?q=qemu&type=issues) but it's probably wrong to rush to blaming it here.

In the very short term, I'll see if I can add `dmesg` output to the end of our GHA runs, just to double check there's nothing from the kernel that could tell us more (the kernel is involved because of the use of `binfmt_misc` and also cause hung tasks can often get printed to the kernel logs).

It's a pretty low chance that'll tell us more, but worth a shot since it's simple and plausible.

If that doesn't help and this continues to happen we may just need to set up a local reproducer (run test in a loop) and maybe attache strace to the build of the shim or something like that to see where it's getting hung up.

sipsma:
Have not encountered this in several months. Hard to say it's conclusively solved but closing if/when it becomes reproducible.

Issue #7024, opened 2024-04-05
`dagger init --sdk python` error on Debian+ZFS


sipsma:
Related discord thread: https://discord.com/channels/707636530424053791/1225451746868985967

It's unknown if ZFS is actually a culprit somehow, but this hasn't been repro'd on other Debian versions with other filesystems yet, so it stands out as an outlier right now.

Error:
```
dagger init --sdk=python test2
✘ ModuleSource.asModule: Module! 0.9s
! failed to create module: select: failed to update codegen and runtime: failed to generate code: failed to call sdk module codegen: select: call function "Codegen": process "/runtime" did not complete successfully: exit code: 2
  ✘ Module.withSource(
      source: ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.0s
    ): Module! 0.9s
  ! failed to update codegen and runtime: failed to generate code: failed to call sdk module codegen: select: call function "Codegen": process "/runtime" did not complete successfully: exit code: 2
  ┃ marshal: json: error calling MarshalJSON for type *dagger.GeneratedCode: input: builtinContainer.rootfs.withoutDire
  ┃ y.withoutFile.file resolve: unlinkat /var/lib/dagger/runc-overlayfs/cachemounts/buildkit3864661092/runtime/template
  ┃ valid argument                                                                                                     
  ┃                                                                                                                    
    ✘ Directory.directory(path: "dagger"): Directory! 0.0s
    ! lstat /dagger: no such file or directory
    ✘ Directory.directory(path: "dagger"): Directory! 0.0s
    ! lstat /dagger: no such file or directory
    ✘ Directory.file(path: "pyproject.toml"): File! 0.0s
    ! unlinkat /var/lib/dagger/runc-overlayfs/cachemounts/buildkit3864661092/runtime/template: invalid argument
      ✘ rm /runtime 0.0s
      ! unlinkat /var/lib/dagger/runc-overlayfs/cachemounts/buildkit3864661092/runtime/template: invalid argument

Error: failed to generate code: input: moduleSource.withContextDirectory.withName.withSDK.withSourceSubpath.asModule resolve: failed to create module: select: failed to update codegen and runtime: failed to generate code: failed to call sdk module codegen: select: call function "Codegen": process "/runtime" did not complete successfully: exit code: 2

Stdout:
[...]

PR #7073, opened 2024-04-10
ci: remove kernel logs from github logs

jedevc:
Came across this while working on continued modularization - ideally, we want to end up in a place where *all* of our CI is just a `dagger call` away! Then we can use https://github.com/dagger/dagger-for-github, and dogfood that more.

This step is more fiddly to modularize (which we want to do so it would show up in dagger cloud) - I think it can be done, but I think we might also just be able to remove it. It was added quite a while ago in https://github.com/dagger/dagger/pull/3785, to help debug https://github.com/dagger/dagger/issues/3784 - which now doesn't occur anymore.

I think it makes sense to just remove these now, I personally have never used them, and they just provide a little bit of friction when trying to scan for failures! If anyone *has* used them recently, I can port them to a dagger function, but just proposing the simpler removal for now.

### DIFF
--- a/.github/workflows/_hack_make.yml
+++ b/.github/workflows/_hack_make.yml
@@ -52,9 +52,6 @@ jobs:
       - name: ${{ inputs.mage-targets }}
         run: |
           ./hack/make ${{ inputs.mage-targets }}
-      - name: "ALWAYS print kernel logs - especialy useful on failure"
-        if: always()
-        run: sudo dmesg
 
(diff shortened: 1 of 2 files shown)


sipsma (approved):
SGTM, it's quite rare they are helpful (one recent occurrence outside of tests https://github.com/dagger/dagger/issues/7024#issuecomment-2040365546) so if they are causing problems I have no problem rming and just adding a step for as needed in the future.

PR #7081, opened 2024-04-11
fix: minor fixes in ts sdk

jedevc:
- Bun cache volume should include the version number
- tsx installation should be part of the base image (so cannot be cache invalidated)
	- @TomChv I think there's some more low-hanging fruit here - the pipeline here doesn't seem to cache in the expected way, we should ensure that wherever possible we can install the dependencies with as few cachebusters around as possible
- Don't mount over /opt (this breaks yarn if you try and use it)

### DIFF
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -78,7 +78,7 @@ func (t *TypeScriptSdk) ModuleRuntime(ctx context.Context, modSource *ModuleSour
 	entrypointPath := path.Join(ModSourceDirPath, subPath, EntrypointExecutablePath)
 	tsConfigPath := path.Join(ModSourceDirPath, subPath, "tsconfig.json")
 
-	ctr = ctr.WithMountedFile(entrypointPath, ctr.Directory("/opt/bin").File(EntrypointExecutableFile))
+	ctr = ctr.WithMountedFile(entrypointPath, ctr.Directory("/opt/module/bin").File(EntrypointExecutableFile))
 
 	switch detectedRuntime {
 	case Bun:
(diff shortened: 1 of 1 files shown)


helderco (approved):

TomChv (approved):

TomChv:
>  I think there's some more low-hanging fruit here - the pipeline here doesn't seem to cache in the expected way, we should ensure that wherever possible we can install the dependencies with as few cachebusters around as possible

Yes I also think there are some issues related to the cache with the TS SDK, but is it also related to the module itself?
I remember we do not cache module's execution, maybe that's related? If we could simply skip the whole setup because the function `CodeGen` or `Runtime` is cached, that would really increase the performances?

jedevc:
> Yes I also think there are some issues related to the cache with the TS SDK, but is it also related to the module itself? I remember we do not cache module's execution, maybe that's related? If we could simply skip the whole setup because the function `CodeGen` or `Runtime` is cached, that would really increase the performances?

The module execution isn't cached, but the steps inside it are - e.g. if I do a `WithExec(["echo", calculateExpensiveString()])` inside a module, then `calculateExpensiveString` doesn't get cached between runs, but the `WithExec` does.

jedevc:
Just dumping here for potential avenues for exploration (duplicating from discord):
- `npm install --package-lock-only` step seems to take a lot of time - is there a faster package manager we could use here? I think this is likely only generating the `package-lock.json`, so I'm not sure what makes this so expensive.
	- Or is this approach even right? It feels like an issue that we wouldn't respect the existing `package-lock.json`, cc @helderco, I know you looked at the python equivalent in #7064.
	- We should actually split out these commands to be separate if possible! There's a lot of stuff being done in groups of commands, so traces only show for that chunk, making it hard to dive deeper.
- Caching doesn't seem to always cache between running `init` and `functions`. Not quite sure why - this should be pretty instantaneous.
- I suspect something in TSX is introducing some latency - is there a way to cache TSX at all? Or maybe we could consider not using TSX at all, and instead compile the typescript into javascript? That feels like it would cache much better potentially, and could move some of the costs upfront.

TomChv:
I'm going to dig into the TS SDK runtime and see if we can somehow optimize it

PR #7095, opened 2024-04-15
fix: preserve windows output

### LINKED ISSUE #7045: 🐞 `dagger.exe` has no output. v0.11.0. Windows.
### What is the issue?

I installed `dagger.exe` and ran `dagger.exe` and got no output.

image

### Dagger version

no ouput, but dagger v0.11.0 amd64 Windows

### Steps to reproduce

Go to user's home directory.

run
`Invoke-WebRequest -UseBasicParsing -Uri https://dl.dagger.io/dagger/install.ps1 | Invoke-Expression`

execute
`./dagger/dagger.exe`

### Log output

<nothing>

jedevc:
Fixes #7045 (nightmare issue yay)

Essentially, with the new TUI, we were failing to set the TUI to raw mode. The reason for this was that we were passing a non-input buffer to containerd/console (through os.Stdin). Unlike on *nix, on windows, the raw mode we were trying to set is *not* interchangeable between stdin/stdout.

Something is insanely weird about what bubbletea is doing here - it's making *loads* of funky assumptions about how terminals can be constructed, and honestly I'm not sure how it works. However, it seems like we can get the right behavior as long as we preserve the differences between input and output buffers - instead of creating one single tty to pass around, we keep two file objects around.

### DIFF
--- a/.changes/unreleased/Fixed-20240415-151604.yaml
+++ b/.changes/unreleased/Fixed-20240415-151604.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Fixed regression that caused crash on windows when TUI output was enabled
+time: 2024-04-15T15:16:04.645921317+01:00
+custom:
+  Author: jedevc
+  PR: "7095"
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -107,8 +107,8 @@ func (fe *Frontend) Run(ctx context.Context, run func(context.Context) error) er
 
 	// find a TTY anywhere in stdio. stdout might be redirected, in which case we
 	// can show the TUI on stderr.
-	tty, isTTY := findTTY()
-	if !isTTY {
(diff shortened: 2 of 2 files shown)


vito (approved):
Well that's fun - nice catch!

jedevc:
Tests failing in `TestModuleDaggerUp/random`, not quite sure why, digging into it now.

jedevc:
Woops, I made a little mistake - we need to gracefully handle the case where stdin is *not* a tty (and therefore `ttyIn` would be `nil`). In this case, we need to gracefully error out (and handle the typed nil issue).

jedevc:
Just for future reference, there is a *slight* behavior change here. Before, if we redirected stdout/stderr to files (but kept stdin to the parent tty), then we would actually end up printing TTY progress still - essentially by printing data to stdin. After this change, directing both stdout/stderr to files will default to plain, and put data to those files.

I think this behavior is definitely "more" expected than the previous, but just thought I'd call it out.

Issue #7093, opened 2024-04-15
TypeScript SDK performance improvement


TomChv:
This issue aims to regroup all my tests and dig into the SDK performances improvement.

## Init

- Env: compiled a binary from main: (bdce95d0abfa9014a35f5e3962a64afce03d5fca)
- Use dagger cli `v0.11.0` 
- Use latest dagger cloud

```
time dagger init --name=perf --sdk=typescript                                              
Initialized module perf in .
dagger init --name=perf --sdk=typescript  0.98s user 0.73s system 11% cpu 14.829 total
➜  perfs git:(main) ✗ time dagger functions                        
Name             Description
container-echo   Returns a container that echoes whatever string argument is provided
grep-dir         Returns lines that match a pattern in the files of the provided Directory
dagger functions  1.32s user 0.78s system 10% cpu 20.765 total
➜  perfs git:(main) ✗ time dagger call container-echo --string-arg "dig into perf" stdout
dig into perf
dagger call container-echo --string-arg "dig into perf" stdout  0.92s user 0.69s system 18% cpu 8.599 total
```

<table>
<tr>
 <td> Operation
 <td> Time
 <td> Cloud URL (by sha)
 <td> Go time (as ref)
<tr>
 <td> Init
 <td> 14.8
 <td> <a href="https://dagger.cloud/Quartz/traces/1951a90ea8679fa505b67f640c4a7ce7">1951a90ea8679fa505b67f640c4a7ce7</a>
 <td> 2.6s
<tr>
 <td> Functions
 <td> 20s
 <td> <a href="https://dagger.cloud/Quartz/traces/7cf74155070bf75952fa5164e3e9667c">7cf74155070bf75952fa5164e3e9667c</a>
 <td> 3.12s
<tr>
 <td> Call container-echo
 <td> 8.6s
 <td> <a href="https://dagger.cloud/Quartz/traces/da94c06b08ba15704f815667a440e44d">da94c06b08ba15704f815667a440e44d</a> 
 <td> 2.89s
</table>

It's obvious that the Go SDK is much faster than TypeScript. Now that we have the context, let's understand why.

## Compares traces

If we compare the [TypeScript init trace](https://dagger.cloud/Quartz/traces/1951a90ea8679fa505b67f640c4a7ce7) and the [Go init trace](https://dagger.cloud/Quartz/traces/15a3e12d0a95affef58dc038b1c60e24).

[...]

PR #7096, opened 2024-04-15
feat: improve TS SDK runtime performances

TomChv:
Following https://github.com/dagger/dagger/issues/7093, I found that most of the time is spent downloading dependencies, it seems I can drastically reduce the download time using `yarn` and the `--production` flag to only install production dependencies.

This is a first step toward improving TS performances, we might switch to `pnpm` later, this require some tests though because I'm hitting an issue (see https://github.com/dagger/dagger/issues/7093#issuecomment-2057038646)


sdk/typescript/runtime/main.go
```diff
@@ -60,6 +60,9 @@ const (
 
 // ModuleRuntime returns a container with the node or bun entrypoint ready to be called.
 func (t *TypeScriptSdk) ModuleRuntime(ctx context.Context, modSource *ModuleSource, introspectionJson string) (*Container, error) {
+	ctx, span := Tracer().Start(ctx, "ModuleRuntime")
+	defer span.End()
```
vito, line 64: I don't think we should add these per-function spans since Dagger will automatically do that for you at the function level. I wouldn't want folks to see this pattern and think they need to do it themselves. Manually added spans should only be for custom parallelism or grouping that the user wants to add on top.

sdk/typescript/runtime/main.go
```diff
@@ -177,38 +186,70 @@ func (t *TypeScriptSdk) CodegenBase(ctx context.Context, modSource *ModuleSource
...
+		// Check if the project has existing source:
+		// if it does: add sdk as dev dependency
+		// if not: copy the template and replace QuickStart with the module name
+		if packageJsonExist {
+			base = base.
+				WithExec([]string{"bun", "install", genDir}).
+				WithExec([]string{"bun", "/opt/module/bin/__tsconfig.updator.ts"})
+		} else {
+			base = base.
+				WithExec([]string{"sh", "-c", "cp -r /opt/module/template/*.json ."})
```
jedevc, line 206: Can this be done using a `WithDirectory` and an `Include` filter? (also same below)

sdk/typescript/runtime/main.go
```diff
@@ -177,38 +186,70 @@ func (t *TypeScriptSdk) CodegenBase(ctx context.Context, modSource *ModuleSource
...
+			WithExec([]string{"cp", "/opt/module/template/src/index.ts", "src/index.ts"}).
+			WithExec([]string{"sed", "-i", "-e", fmt.Sprintf("s/QuickStart/%s/g", strcase.ToCamel(name)), "src/index.ts"})
+	}
+
+	return base, nil
 }
 
-// Base returns a Node or Bun container with cache setup for yarn or bun
-func (t *TypeScriptSdk) Base(runtime SupportedTSRuntime) (*Container, error) {
+// Base returns a Node or Bun container with cache setup for pnpm or bun
```
jedevc, line 248: :eyes: this comment doesn't seem right - we're using yarn right? We should also add a yarn cache here.

Could we replace the `npx` usage with yarn?

sdk/typescript/runtime/main.go
```diff
@@ -89,7 +92,7 @@ func (t *TypeScriptSdk) ModuleRuntime(ctx context.Context, modSource *ModuleSour
 	case Node:
 		return ctr.
 			// Install dependencies
-			WithExec([]string{"npm", "install"}).
+			WithExec([]string{"yarn"}).
```
jedevc, line 95: Could we use `yarn --production` here as well?

TomChv: > Could we use yarn --production here as well?

Unfortunately no we cannot, otherwise it fails with a weird node error that says a package is missing

jedevc: That's because the generated `package.json` doesn't look right:

```json
{
  "dependencies": {
    "typescript": "^5.3.2"
  },
  "devDependencies": {
    "@dagger.io/dagger": "./sdk"
  }
}
```

Is `@dagger.io/dagger` a dev dependency if it's *required* to actually run the module? IMO, it should be moved to a normal dependency, then we can switch this to `--production`.

sdk/typescript/runtime/main.go
TomChv: Done

sdk/typescript/runtime/main.go
TomChv: Will it merge the current `.` with what I provide? If yes then I can

vito: Yep it'll merge (I forget that all the time)

jedevc:
@TomChv how's this one going? Anything you need a hand with?

sdk/typescript/runtime/main.go
TomChv: I found out after multiples tests that it makes the pipeline slower to use yarn to install global packages. so I'll stick to npm

sdk/typescript/runtime/main.go
TomChv: Yes but there will be no improvement in the performance by doing that, since we'll only have 2 main dependencies and no dev one

TomChv:
It seems theses changes in the setup introduce issues in the tests, sometime the dependencies cannot be installed properly and it's a weird typescript error.
I'm digging into it

TomChv:
I can reproduce the error using `dagger call --source=. test custom --run="TestModuleTypescriptInit"`

I'm going to try to reproduce it locally, it seems something is going wrong when using `dagger init` with the new setup, I'll update this comment with my researches.

sdk/typescript/runtime/main.go
```diff
@@ -223,18 +177,117 @@ func (t *TypescriptSdk) Base(runtime SupportedTSRuntime) (*Container, error) {
...
+	// If there's a package.json, run the tsconfig updator script.
+	// else, copy the template config files.
+	if packageJSONExist {
+		executor := "tsx"
+
+		if runtime == Bun {
+			executor = "bun"
+		}
+
+		ctr = ctr.WithExec([]string{executor, "/opt/module/bin/__tsconfig.updator.ts"})
```
helderco, line 213: Why not put this logic in Go since it's faster?

helderco:
> I saw that I copy the `runtime` directory into base container, it may be possible to improve performances by ignoring part of it. I think I could apply the same logic in the `sdk` directory to ignore tests files that are useless during the execution.

You may have forgotten but putting the whole `sdk` directory was just a stop gap so we could work on more important things first. But the intent was to eventually cycle back and refactor the SDK to decouple the parts that aren't needed in modules. Python needs the same.

sdk/typescript/runtime/main.go
TomChv: That's a good point, however I think I should take care of that in a follow up

gerhard:
I would very much like to see a version of this land in the next milestone, `v0.11.5`:
- https://github.com/dagger/dagger/milestone/47 

Even if we don't finish everything, any improvement that we can make will go a long way.

gerhard:
I tried this with a simple TypeScript pipeline for a Next.js app:
```typescript
/**
 * Build automation powered by Dagger
 */
import { dag, Directory, Secret, object, func } from "@dagger.io/dagger"

@object()
class App {
  /**
   * Publish image: dagger call image --source=.:app --token=env:GH_TOKEN
   */
  @func()
  async image(source: Directory, token: Secret): Promise<string> {
    return dag
      .container()
      // https://hub.docker.com/_/node/tags?page=&page_size=&ordering=&name=lts
      .from("node:lts-iron")
      // https://github.com/krallin/tini
      .withExec(["apt", "update"])
      .withExec(["apt", "install", "--yes", "tini"])
      .withDirectory("/app", source)
      .withWorkdir("/app")
      .withExec(["npm", "install"])
      .withExec(["npm", "run", "build"])
      .withEntrypoint(["tini", "--", "npm", "run", "start"])
      .withRegistryAuth("ghcr.io", "gerhard", token)
      .publish("ghcr.io/gerhard/nextjs")
  }
}
```

I am connecting to a remote Docker instance via SSH:
```console
Client: Docker Engine - Community
 Version:           26.0.0
 OS/Arch:           darwin/amd64

Server: Docker Engine - Community
 Engine:
  Version:          26.1.3
  OS/Arch:          linux/amd64
 containerd:
  Version:          1.6.31
 runc:
  Version:          1.1.12
```

Linux 6.8 x86_64 host where Docker / Dagger is running has:
- 16 CPUs
- 32GB DDR5
- NVMe disk

These are the results for various runs:

| COMMAND                         | `v0.11.4` - BEFORE THIS CHANGE | `dev` - AFTER THIS CHANGE |
| -                               | -                              | -                         |
| 1st run - no cache              | `95s`                          | `108s`                    |
| 2nd run - cached                | `11s`                          | `9s`                      |
| 3rd run - `touch foo`           | `50s`                          | `49s`                     |
| 4th run - module comment change | `32s`                          | `19s`                     |
| 5th run - module comment change | `22s`                          | `20s`                     |

Based on the above measurements, I observed one improvement, **4th run**, when I just kept updating a comment in the module, which was `40%` quicker: `19s` vs `32s`

Before this change, that was inconsistent: **5th** run was only `10%` quicker: `20s` vs `22s`

I also observed one deterioration: **1st run - no cache** was `13%` slower: `108s` vs `95s`

Here are Dagger Cloud traces for the interesting runs above:
- `1st run - no cache` [before this change](https://dagger.cloud/dagger/traces/0547c97176228e750c767ed52763aef6) & [after this change](https://dagger.cloud/dagger/traces/db5c8ee47a3539fadc63e8b1d61e06a3)
- `4th run` [before this change](https://dagger.cloud/dagger/traces/f2756499c07b581dd654952459c55e40) & [after this change](https://dagger.cloud/dagger/traces/143b8817ea96ca6abb89fcb8135b9460)
- `5th run` [before this change](https://dagger.cloud/dagger/traces/39e3f02038b8866eb2acc9ab4d1cee66) & [after this change](https://dagger.cloud/dagger/traces/806b48cbbad8676a3c429f1ffe2486e4)

Based on the above measurements, this does not seem a net performance improvement to me. What do you think @TomChv:
1. Do my measurements look correct?
2. Was I measuring the right thing?
3. Do you still think that it's worth continuing in this direction?

TomChv:
Thanks for your amazing feedbacks! It really helps me understand what's going on with a real world use case.

> Do my measurements look correct?

Yep absolutely, I could hit the same kind of behaviour and flakiness on my tests! Specially between 4 and 5.

The improvement of 40% to 10% is also something I got which confirm my work! I could hit a 2x improvement with an old version but it seams that it wasn't supporting every cases (to be more specfic: the `hasPkgJson` & `hasTsConfig`) was failing.

>Was I measuring the right thing?

Yes! Perfect, I needed the run and the traces, to see how much time take every runs and which steps is longer.

>Do you still think that it's worth continuing in this direction?

I think what I've done in this PR is a good start for further improvement, there's still some mystery like this example below where I cannot understand how is that possible.

Screenshot 2024-05-21 at 21 48 54

I'm super confused about the fact that yarn was waaay slower than npm when the `withExec` has multiple instruction, and it's even weirder because npm is only supposed to update the `package.json` (that's why `--package-lock-only` is done) but then the real [`npm install` took only 2.3s](https://dagger.cloud/dagger/traces/0547c97176228e750c767ed52763aef6#e3eb0339e5509fb3) which is super confusing because it should have install all the package.

I think I'm missing something about yarn, which keep us from using the cache. I'm thinking that at some point, we could hit higher performance if we were running the TS SDK in runtime language but that seams quite complex to do.
The overall installation of dependencies takes time, the cache isn't hitting every time (which doesn't make sense since nothing should have before adding user's source in `CodegenBase`.

I think it's still worth merging the PR, because it makes the module much clearer but there's still a long way to go before getting a super fast TS SDK

TomChv:
I'm also having a concern about the CI and the flakiness this PR introduce, on multiple runs, I could see different tests passing & failing, with always this kind of error:

```
couldn't find a package.json file in "/usr/local/share/.cache/yarn/v6/npm-@dagger-io-dagger-0.0.0-1cdff291-b199-40fc-9ef6-3ff1ac6d6d1d-1716293036796/node_modules/@dagger.io/dagger" 
```

Meaning that there's an issue with yarn cache and I cannot understand why is that so, this is something that isn't happening with npm at all.

TomChv:
By doing so, I could successfully get a green success on initialization, which is a good news

Screenshot 2024-05-21 at 23 46 44

TomChv:
@gerhard It seems I could fix the CI, could you do one more benchmark with your example just to see if there's any changes? I think it might get slower, I'll need to find why the yarn cache fails with collision.
A really simple check would be to add a specific cache-key for each module but I'm not sure we want to do that.

TomChv:
**Conclusion of this PR**

This PR changed several internal things in the TypeScript SDK runtime.

I'm not sure I could succeed in my path to reduce time by 2 but I still want to give a summary of what will happens with this PR:
- I  switched from `npm` to `yarn v4` which in theory should make things faster.
- I reorganized the SDK initialization to install the SDK before adding the user's source which should in theory improving the cache.
- I improved the TS SDK code to be more readable and more clear.
- I improve the user's `package.json` update using `npm pkg set` instead of a simple `npm install` which definitelly give at least a second of improvement.
- I moved `@dagger.io/dagger` from `devDependencies` to `dependencies`. That allows to use `--production` flag which in theory is faster than installing all dependencies.
- I deduplicate the TS SDK `yarn.lock` to remove any no-essential dependencies from the lockfile, it's a way to reduce the overall download time.
- I gained valuable knowledge around the runtime and how package manager works.

**Few things to do as next steps**
- Use a faster language to update `tsconfig.json`, this step takes 2s every time.
- Add back cache key on yarn module. I still need to understand what's going on with that but we can save it for later, I think it will take some time and current resources on the web are not that helpful in our specific context.
- Dig more into cache system, the SDK installation is always trigger but it shouldn't, my theory is that its because of the list files operations, I need to find a way to do it in a different way or reorder steps.

sdk/typescript/runtime/main.go
```diff
@@ -218,23 +177,125 @@ func (t *TypescriptSdk) Base(runtime SupportedTSRuntime) (*Container, error) {
 		return dag.Container().
 			From(nodeImageRef).
 			WithoutEntrypoint().
+			WithExec([]string{"corepack", "enable"}).
 			// Install default CA certificates and configure node to use them instead of its compiled in CA bundle.
 			// This enables use of custom CA certificates if configured in the dagger engine.
 			WithExec([]string{"apk", "add", "ca-certificates"}).
 			WithEnvVariable("NODE_OPTIONS", "--use-openssl-ca").
 			WithMountedCache("/root/.npm", dag.CacheVolume(fmt.Sprintf("mod-npm-cache-%s", nodeVersion))).
+			// WithMountedCache("/usr/local/share/.cache", dag.CacheVolume(fmt.Sprintf("mod-yarn-cache-%s", nodeVersion))).
```
jedevc, line 186: Why is this commented out? Either it should be enabled or removed.

Also, this doesn't look like the right path.

sdk/typescript/runtime/main.go
```diff
@@ -218,23 +177,125 @@ func (t *TypescriptSdk) Base(runtime SupportedTSRuntime) (*Container, error) {
 		return dag.Container().
 			From(nodeImageRef).
 			WithoutEntrypoint().
+			WithExec([]string{"corepack", "enable"}).
```
jedevc, line 180: Would be good to get a comment for what this does.

sdk/typescript/runtime/main.go
```diff
@@ -131,79 +137,32 @@ func (t *TypescriptSdk) CodegenBase(ctx context.Context, modSource *ModuleSource
...
-			fmt.Sprintf("mkdir -p src && if ls src/*.ts >/dev/null 2>&1; then true; else cp /opt/module/template/src/index.ts src/index.ts && sed -i -e 's/QuickStart/%s/g' ./src/index.ts; fi", strcase.ToCamel(name))},
-		), nil
+		// Mount users' module with SDK sources and generated client in it.
+		WithMountedDirectory(ModSourceDirPath, modSource.
+			ContextDirectory().
+			WithDirectory(filepath.Join(subPath, GenDir), sdk),
+		).
+		WithWorkdir(filepath.Join(ModSourceDirPath, subPath))
+
+	return t.setupModule(ctx, base, detectedRuntime, name)
```
jedevc, line 165: Good refactor here :tada:

This was *very* hard to understand before, super glad to see this cleaned up! :heart:

jedevc (commented):
I like the refactor here, very nice :tada:

sdk/typescript/runtime/main.go
TomChv: > Add back cache key on yarn module. I still need to understand what's going on with that but we can save it for later, I think it will take some time and current resources on the web are not that helpful in our specific context.

I don't want to remove it, it's a workaround to fix an issue with yarn cache, but yeah I think I miss wrote the cache path too

sdk/typescript/runtime/main.go
TomChv: Done in [387eb49](https://github.com/dagger/dagger/pull/7096/commits/387eb49a0da4916c410830d8cbc6a72cd086dde2)

sdk/typescript/runtime/main.go
TomChv: Howerver, I added a comment to explain why this line is commented and fixed the path ([387eb49](https://github.com/dagger/dagger/pull/7096/commits/387eb49a0da4916c410830d8cbc6a72cd086dde2)

jedevc (approved):
Tentative approval here, I think this definitely lays the foundations to do some more cache analysis/perf improvements here :tada:

Nice start!

TomChv:
> Tentative approval here, I think this definitely lays the foundations to do some more cache analysis/perf improvements here 🎉

> Nice start!

Thanks for the approval, I'm gonna merge the PR once CI is green.

gerhard:
There is a lot of helpful detail in your recent comments @TomChv, thank you for going the extra mile with this.

I am doing one more test now and will share the results - will update this comment.

TomChv:
> 👆 Links to Dagger Cloud Traces inline.

💪 LGTM 🚀

Thanks for these amazing benchmark! I'll add the changie and we can merge it

TomChv:
Release note added in [c423bf9](https://github.com/dagger/dagger/pull/7096/commits/c423bf9e57722a6f4fb75475f967267dd34289e1)

gerhard:
This introduced a bug in provisioning, we will not be able to release without a fix @TomChv: https://github.com/dagger/dagger/actions/runs/9193090706/job/25283718931#step:6:11

image

Issue #5697, opened 2023-08-25
✨ More idiomatic Rust examples


roboteng:
### What are you trying to do?

This isn't quite a feature request, but a request for better examples.

I'm using the Rust sdk, as I want to use dagger to run my Rust project through CI. The examples do a good job of showing how to build go and js projects, but I still found it difficult to get a useable pipeline, written in Rust, for a Rust project.

I mad e Rust project, seperated into workspaces, one of which is `ci`. This is the `main.rs` file that I used to get Dagger to run with a reasonable cache, but it took a lot of effort and research. I think it would be good if we could provide this (or more realistically, something like this) as a good starting point in the examples folder.

```rust
use dagger_sdk::HostDirectoryOpts;

const PROJECT: &str = "/project";
const PROJECT_TARGET: &str = "/project/target";

#[tokio::main]
async fn main() -> eyre::Result<()> {
    let client = dagger_sdk::connect().await?;

    let host_source_dir = client.host().directory_opts(
        ".",
        HostDirectoryOpts {
            exclude: Some(vec!["target"]),
            include: None,
        },
    );

    let build_cache = client.cache_volume("target").id().await?;
    let cargo_registry = client.cache_volume("cargo").id().await?;

    let container = client
        .container()
        .from("rust:1.71.1")
        .with_mounted_directory(PROJECT, host_source_dir.id().await?)
        .with_mounted_cache(PROJECT_TARGET, build_cache);

    let cargo_home = container.env_variable("CARGO_HOME").await?;

    let container = container.with_mounted_cache(&format!("{cargo_home}/registry"), cargo_registry);

    let container = container
        .with_workdir(PROJECT)
        .with_exec(vec!["cargo", "check"]);

    let container = container
        .with_workdir(PROJECT)
        .with_exec(vec!["cargo", "test"]);

    let container =
        container
            .with_workdir(PROJECT)
            .with_exec(vec!["rustup", "component", "add", "clippy"]);

    let container = container
        .with_workdir(PROJECT)
        .with_exec(vec!["cargo", "clippy"]);

    container
        .with_workdir(PROJECT)
        .with_exec(vec!["cargo", "build", "-p", "sample"])
        .stdout()
        .await?;

    Ok(())
}
```

I'm aware that it seems like all the sdk seem to use the same examples. Having a Rust specific example in the Rust sdk would break that pattern, but I think it would be helpful to have, even if not in the examples.

### Why is this important to you?

Having this would have saved me a lot of toil

[...]

PR #7097, opened 2024-04-15
chore(rust-sdk): add examples to the Rust sdk

FlrnFrmm:
As suggested in #5697 & coordinated with @kjuulh. 
Adds 3 examples of how to use the dagger Rust sdk to `sdk/rust/examples`. 
The 3 examples are:
- Building a simple CLI application
- Building & packaging a Axum based backend
- Building & packagang a Leptos based frontend


sdk/rust/examples/backend/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+[package]
+name = "backend"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+clap = { version = "4.5.4", features = ["derive"] }
+dagger-sdk = "0.9.8"
```
kjuulh, line 11: We should depend on the version from the workspace. That way we can actually verify the examples work using the current version. Can be done with 

```toml
dagger-sdk = {path = "../../crates/dagger-sdk"} # may not be the correct path fyi :)
```

Or we can set the dagger-sdk in the workspace and use that here

```toml
# Root Cargo.toml
[workspace.dependencies]
dagger-sdk = {path = "crates/dagger-sdk"}

# ---
# file: examples/backend/Cargo.toml
[dependencies]
dagger-sdk.workspace = true
```

kjuulh: We in general if we can use workspace versions while still keeping the examples clear and easy to use that would be a boon, as we then don't have to maintain many versions of the same packages.

sdk/rust/examples/backend/src/configuration.rs
```diff
@@ -0,0 +1,6 @@
+#[derive(clap::Parser)]
+#[command(version, about)]
+pub struct Configuration {
```
kjuulh, line 3: This doesn't look to be used? 

if you want env here as well. You can enable the env flag for `clap`, then it can take an env = "PORT" as a backup if the flag isn't set on the cli.

sdk/rust/examples/backend/src/main.rs
```diff
@@ -0,0 +1,46 @@
...
+    let image_reference = push_image(image).await?;
+    println!("Image published at: {}", image_reference);
+    Ok(())
+}
+
+async fn build_backend(client: &Query) -> File {
+    let backend_directory = client.host().directory("axum-backend");
+    client
+        .container()
+        .from("rust:alpine")
```
kjuulh, line 23: consider pinning the version of the image, otherwise we may silently fail builds

sdk/rust/examples/backend/src/main.rs
```diff
@@ -0,0 +1,46 @@
...
+        .file("./target/release/axum-backend")
+}
+
+async fn build_prod_image(client: &Query, build: File) -> Container {
+    let Configuration { port } = Configuration::parse();
+    client
+        .container()
+        .from("gcr.io/distroless/static-debian12")
+        .with_file(".", build)
+        .with_env_variable("PORT", port.to_string())
```
kjuulh, line 37: we likely shouldn't set this here unless you of course want a default for the app. Otherwise this will only be set for the build and not when it is actually running.

sdk/rust/examples/cli/src/main.rs
```diff
@@ -0,0 +1,18 @@
...
+
+    let build_file = client
+        .container()
+        .from("rust:slim-buster")
+        .with_directory("./app", app_directory)
+        .with_workdir("/app")
+        .with_exec(vec!["cargo", "build", "--release"])
+        .file("./target/release/app");
+
+    build_file.export("./build/cli").await?;
```
kjuulh, line 15: not entirely sure export works if the folder doesn't already exists. I.e. `build`

sdk/rust/examples/frontend/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+    let image_reference = push_image(image).await?;
+    println!("Image published at: {}", image_reference);
+    Ok(())
+}
+
+async fn build_frontend(client: &Query) -> Directory {
+    let backend_directory = client.host().directory("leptos-frontend");
+    client
+        .container()
+        .from("rust:latest")
```
kjuulh, line 19: again lets pin the version such that we don't silently fail

sdk/rust/examples/frontend/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+        .with_directory("./frontend", backend_directory)
+        .with_workdir("/frontend")
+        .with_exec(vec!["rustup", "target", "add", "wasm32-unknown-unknown"])
```
kjuulh, line 22: we should add dependencies and such before importing the dir. Otherwise we will cache bust everything on a file change

kjuulh (commented):
A few comments. It looks good!

sdk/rust/examples/cli/src/main.rs
FlrnFrmm: I tested the example on Ubuntu and MacOS.
The 'build' folder was on both OS created, if not present.

gerhard:
Looking good @FlrnFrmm, almost there!

Do you want to rebase & get the checks green so that we can get this merged?

jedevc (approved):

Issue #7030, opened 2024-04-05
🐞 Dagger version will throw a warning message for git worktree


slumbering:
### What is the issue?

If you type any dagger command in a git worktree, you'll get the following Warning message:
`WRN failed to get repo HEAD err="reference not found"`

The error could potentially come from here:
https://github.com/dagger/dagger/blob/95b4ce550a2dcd1fa73f8b55199c7e9ed0ef1fc3/telemetry/labels.go#L109-L113

### Dagger version

dagger v0.11.0 (registry.dagger.io/engine) darwin/arm64

### Steps to reproduce

1 - Clone a repository with the flag `--bare`:
`git clone --bare https://github.com/dagger/dagger.git dagger`
`cd dagger`

2 - Add the main branch as a worktree:
`git worktree add main`
`cd main`

3 - Type any Dagger command:
`dagger version`

### Log output

dagger v0.11.0 (registry.dagger.io/engine) darwin/arm64
16:05:41 WRN failed to get repo HEAD err="reference not found"

PR #7098, opened 2024-04-16
fix: ensure git labels can use worktrees

jedevc:
Fixes https://github.com/dagger/dagger/issues/7030.

Git worktrees use a `.git/commondir` to ensure that refs and similar are shared. We need to explicitly enable this with go-git.

Technically, this change is only required in the `telemetry` package to fix the above issue, but we can also just add it everywhere that we call this function.


gerhard (approved):
LGTM!

image

PR #7100, opened 2024-04-16
sdk: bump engine dependency to v0.11.1

jedevc:
This PR was *not* auto-generated. Looks like I accidentally deleted the step that did this as part of `engine:publish` while doing #6843.

Thankfully, we can add this back in later, and this is non-critical :tada:


jpadams (approved):
onwards! 🚀

PR #7104, opened 2024-04-16
ci: fix elixir publish version

jedevc:
Ugh, *again*. The elixir path wasn't right here, this would mean we were instead publishing v0.0.0.

image


marcosnils (approved):

jedevc:
Once more! With feeling!

wingyplus:
Thank you. ❤️

PR #6755, opened 2024-02-27
chore: disable debug=true for published/tested image

jedevc:
There's not a huge reason to include debug=true in tests - these logs add huge amounts of extra logging, that is often not particularly helpful during debugging.

For local development, with `./hack/dev`, these were already enabled with the `--debug` flag - it's just with the engine for tests.

Note - I also removed it for the images we push to `registry.dagger.io` - it can be added back, but I'm actually not convinced adding `debug` logging by default is the right call.

PR #7106, opened 2024-04-17
engine: log buildkit progress to stderr with debug enabled

vito:
Right now if something goes wrong internal to the shim before all the OpenTelemetry plumbing can be wired up, you just get an `exit status 125` with no other output. (The shim currently has to manually send the command's stdout/stderr to OpenTelemetry logging, so there are extra moving parts.)

I added a "break glass to debug telemetry" bit that prints Buildkit logs to stderr, but didn't make it configurable, figuring it would only be useful to me while I bootstrap the OpenTelemetry stack.

But now, as is tradition, a user has a hit this case in production.

So with this PR, now when the engine is configured with `debug = true` it'll also print Buildkit's logs in console format to stderr like this:

```
time="2024-04-17T03:31:29Z" level=debug msg=Unauthorized header="Bearer realm=\"https://cgr.dev/token\",service=\"cgr.dev\",scope=\"repository:chainguard/apko:pull\"" host=cgr.dev
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] #47 ...
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] 
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] #48 resolve image config for docker-image://cgr.dev/chainguard/apko:latest
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] #48 DONE 0.6s
time="2024-04-17T03:31:29Z" level=debug msg="load cache for mkfile /config.yml with 8e1cq0osjisihh8if42eb91d6::nl306rmhh3uqw60xhfd4x4dly"
time="2024-04-17T03:31:29Z" level=debug msg="checked for cached auth handler namespace" cached=true key="cgr.dev/chainguard/apko::pull" name=cgr.dev/chainguard/apko scope=pull
time="2024-04-17T03:31:29Z" level=debug msg="checked for cached auth handler namespace" cached=true key="cgr.dev/chainguard/apko::pull" name=cgr.dev/chainguard/apko scope=pull
time="2024-04-17T03:31:29Z" level=debug msg="load cache for exec apko build --cache-dir /apkache/ ./twLXXUiV_kQ=/config.yml latest ./layout.tar with 8e1cq0osjisihh8if42eb91d6::oavny64osuh8e5c7fqwjoj58t"
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] 
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] #49 mkdir /
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] #49 DONE 0.0s
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] 
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] #50 mkfile /config.yml
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] #50 CACHED
```

I thought about putting this at trace level, but I actually think this is pretty useful in all contexts where `debug = true`, so there's an extra commit for making trace level configurable like `env TRACE=1 ./hack/dev`, which I kept even though it's not needed for this change.

### DIFF
--- a/ci/engine.go
+++ b/ci/engine.go
@@ -23,7 +23,6 @@ type Engine struct {
 	Args   []string // +private
 	Config []string // +private
 
-	Debug bool // +private
 	Trace bool // +private
 
 	GPUSupport bool // +private
@@ -44,11 +43,6 @@ func (e *Engine) WithGPUSupport() *Engine {
 	return e
 }
 
-func (e *Engine) WithDebug() *Engine {
-	e.Debug = true
-	return e
-}
-
 func (e *Engine) WithTrace() *Engine {
 	e.Trace = true
 	return e
@@ -61,7 +55,7 @@ func (e *Engine) Container(
 	// +optional
 	platform dagger.Platform,
 ) (*Container, error) {
-	cfg, err := generateConfig(e.Debug, e.Trace, e.Config)
+	cfg, err := generateConfig(e.Trace, e.Config)
 	if err != nil {
 		return nil, err
 	}
--- a/ci/engine.go
+++ b/ci/engine.go
@@ -23,6 +23,9 @@ type Engine struct {
 	Args   []string // +private
 	Config []string // +private
 
+	Debug bool // +private
+	Trace bool // +private
+
 	GPUSupport bool // +private
 }
 
@@ -41,6 +44,16 @@ func (e *Engine) WithGPUSupport() *Engine {
 	return e
 }
 
+func (e *Engine) WithDebug() *Engine {
+	e.Debug = true
+	return e
+}
+
+func (e *Engine) WithTrace() *Engine {
+	e.Trace = true
+	return e
+}
+
 // Build the engine container
 func (e *Engine) Container(
 	ctx context.Context,
@@ -48,7 +61,7 @@ func (e *Engine) Container(
 	// +optional
 	platform dagger.Platform,
 ) (*Container, error) {
-	cfg, err := generateConfig(e.Config)
+	cfg, err := generateConfig(e.Debug, e.Trace, e.Config)
 	if err != nil {
 		return nil, err
 	}
--- a/ci/config.go
+++ b/ci/config.go
@@ -40,7 +40,8 @@ exec {{.EngineBin}} --config {{.EngineConfig}} {{ range $key := .EntrypointArgKe
 `
 
 const engineConfigTmpl = `
-debug = true
+debug = {{.Debug}}
+trace = {{.Trace}}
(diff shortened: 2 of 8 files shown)


jedevc:
```
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] 
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] #49 mkdir /
[buildkit] [11bgeg5ytjd3h8dy06z9j18re] #49 DONE 0.0s
```

Is there any way to easily correlate this with our dagger logs?

ci/engine.go
```diff
@@ -41,14 +44,24 @@ func (e *Engine) WithGPUSupport() *Engine {
 	return e
 }
 
+func (e *Engine) WithDebug() *Engine {
+	e.Debug = true
+	return e
+}
```
jedevc, line 50: I think this makes `debug = false` the default? So our tests are now effectively running without debug logging? I remember we discussed this in #6755, but not sure if this is intentional here?

vito: Ah, yep looks like I kept it for `./hack/dev` but not the tests, will adjust!

vito: Actually this brings up another topic: our published engine _also_ as `debug = true` by default. Should we keep doing that?

vito:
@jedevc Thanks for taking a look!

I just brought back `debug = true` for now, both for tests and prod. If you think it's OK to remove it for prod I can push another change to enable it for tests but not prod, just wanted to disconnect the PR from any broader decisions for now so I can get a dev image built that the user can try out.

> Is there any way to easily correlate this with our dagger logs?

Depends on what you mean by Dagger logs - user-level telemetry, or the engine logs?

I just added the trace ID to the prefix which should help debug user level logs, and I kept the client ID in case that helps debugging from that direction. But for engine logs it's a bit of a wash - some lines above don't include any IDs at all.

sipsma (approved):
This will result in a ton of extra output on the default log level, but not opposed for now since it's more important to debug the user issue.

One alternative would be to do this on the `ExtraDebug` level I added recently instead. That can be enabled by providing `--extra-debug` as a cmdline arg to the engine. Won't block on that though.

vito:
> One alternative would be to do this on the `ExtraDebug` level I added recently instead. That can be enabled by providing `--extra-debug` as a cmdline arg to the engine. Won't block on that though.

Will keep that in mind if it proves too noisy. I was originally trying to stick to Debug vs. Trace since that's all the config file supports, but a flag would probably be easy enough for them to set too (I think?)

PR #6549, opened 2024-02-01
CLI: Improve display of core API in `--help`

helderco:
Fixes #6358

Putting individual improvements in separate commits.

Notice in the example below that if we follow #6516 and keep the first line of API descriptions up to 50 chars, that'll be short of the end of the line below "EXPERIMENTAL". Meaning that it won't wrap on this screen size.

## Todo

Not all of these are focused on the core API, most impactful is the wrapping on flags and commands. I’m submitting what I have now to be cut in next release but will continue in a separate PR for general `--help` improvements.

- [x] Group commands in `dagger` root (modules, execution, cloud, additional)
- [x] Group function commands in `dagger call`
- [x] Wrapping on flag usages
- [x] Wrapping on available command descriptions
- [x] Only use first line of a function’s description for cmd.Short 
- [x] Move flags before sub commands
- [ ] https://github.com/dagger/dagger/issues/6516
- [ ] Conventionalize on `<command>` syntax instead of `COMMAND` in `cmd.Use`
- [ ] Global flags only on root command and top level commands
    - [ ] Traverse children
    - [ ] Remove persistent flags from child commands after being parsed
        - [ ] TUI (`--silent`, `--focus`)
        - [ ] Modules (`-m`)
- [ ] Add line break between error and usage
- [ ] Style headings in **BOLD UPPERCASE** to help break sections visually
- [ ] Help topics for `dagger call` to keep `--help` lean while still providing more documentation

## Example

Taking an example of a function returning a `dagger.Container` which has a lot of functions.

- **Before:** https://app.warp.dev/block/1W5uEInSvPgnwJ4dN9iC9f
- **After:** https://app.warp.dev/block/KwLaq1XtV6T19lA8M2UQle

PR #7107, opened 2024-04-17
cli: de-clutter CLI

### LINKED ISSUE #6943: De-clutter `dagger call --help`
## Problem

`dagger call [ARGS] --help` plays a crucial role, and must be kept as lean as possible. At the moment there is clutter that could be removed.

## Solution

De-clutter the output of `dagger call [ARGS] --help`.

- [x] Talk about "functions" and "arguments", rather than "function commands" and "flags"
  - https://github.com/dagger/dagger/pull/7286
- [x] Don't show inherited flags when a function is selected
  - https://github.com/dagger/dagger/pull/7419
- [x] Don't show functions that the CLI can't run
  - https://github.com/dagger/dagger/pull/7418
- [x] Keep long description short by linking to the web for more information
  - https://github.com/dagger/dagger/pull/7277

### Additional

Improvements to make `--help` easier to read:

- [x] Style headings in **BOLD UPPERCASE** to help break sections visually
  - https://github.com/dagger/dagger/pull/7126
- [x] Put flags between commands and global flags
  - https://github.com/dagger/dagger/pull/7134
- [x] Visual cue for required flags
  - https://github.com/dagger/dagger/pull/7262
- [x] Examples aren't indented
  - https://github.com/dagger/dagger/pull/7287

Improve consistency:

- [x] "Options" instead of "Flags"
  - https://github.com/dagger/dagger/pull/7170
- [x] Conventionalize usage syntax for consistency
  - https://github.com/dagger/dagger/pull/7143
- [x] `dagger query` has a different help template
  - https://github.com/dagger/dagger/pull/7124
- [x] `dagger functions <function> --help` doesn't work
  - https://github.com/dagger/dagger/pull/7187

grouville:
### Details of current draft implementation

@helderco we had a sync, but I might have forgotten some parts ahah (sorry 👼). Feel free to tell if I'm completely off on some interpretation / implementation:

- [x] **Add line break between error and usage**
This has been added in `cmd/dagger/functions.go`

- [x] **Style headings in BOLD UPPERCASE to help break sections visually**
A new helper function has been added and the corresponding template has been updated

- [x] **Remove discrepancy between `dagger query` and the rest of the commands**
This sub-command hasn't been updated in months and having a separate template is more coordination work. It might be time to make them 1:1 ?

- [x] **Conventionalize on either `<command>` or `COMMAND` syntax in cmd.Use**
I updated all COMMAND as uppercase. I did not remember if you were also referencing on the usage examples inside square brackets like `[arguments]` -> `[ARGUMENTS]`  ? I followed the CLI examples we discussed and some did it that way, in doubt I applied it everywhere but totally open for change

- [x] **Figure out if arguments are better above or below functions/commands ??**
I experimented on this one, and I haven't found an easy way: on a given function call, I do agree that having the commands at the bottom does make sense. However, changing the order on the template directly also changes the order for commands such as: `dagger --help`. _What do you think is the best implementation on this one ? Shall we tweak the template to make it specific for just the `dagger call` section ?_ I feared adding complexity to the template logic

- [x] **Visual cue for required flags**
As you proposed last time, the idea would be to have a required flags section for the `dagger call function` subcommands only ? Will take a look tomorrow. Same fear on the added complexity on the main template, but this one seems more manageable: it is just a section wrapped in an if to show it on `dagger call` subcommand context 

- [x] **"Options" instead of "Flags"?[1](https://github.com/dagger/dagger/issues/6943#user-content-fn-1-6303e1ecaa2e73e24edcc8788f13dcde)**
This is done, with our own implementation of `useLine` to print options instead of args by default.
Solomon requested `arguments` on above issue though. I followed your terminology, but shall we make a specific case for arguments in `dagger call` / `dagger functions` scope ?

- [ ] **Help topics for dagger call to keep --help lean while still providing more documentation**
What topics do you think we should cover first ? More than the implementation, the most important might be the documentation added to those topics. Which could be the first 2 topics to create, so that i draft the implementation tomorrow ?
 
- [ ] **Global flags only on root command and top level commands**
    * Remove persistent flags from child commands after being parsed? Or just hide?
        * TUI (--silent, --focus)
        * Modules (-m)

This is the one I am currently trying to implement in a clean way (using the hiding method at first), and might need some guidance again. I have been trying to extend the `callCmd` implementation by implementing a new `persistentPreRunE`:
```go
PersistentPreRunE: func(_ *FuncCommand, cmd *cobra.Command, _ []string) error {
    rootCmd := cmd
    for rootCmd.Parent() != nil {
        rootCmd = rootCmd.Parent()
    }
    hideGlobalFlags(rootCmd)
    return nil
},
```
to control the behavior and potentially use the funcCommand infos or the cmd infos to mark as hidden, as you suggested, when the call command has subcommands 
However, I'm still having a hard time on this part. Will keep iterating tomorrow again

- [x] **Talk about Functions and arguments instead of Function commands**
This is partially done as I used your `options` terminology globally instead of `arguments`. It is true that `arguments` is pretty accurate in a `dagger call functions` context, but then, do we talk about `arguments` inside `dagger functions` usage examples too ?
 
cc @helderco

### DIFF
--- a/cmd/dagger/call.go
+++ b/cmd/dagger/call.go
@@ -16,7 +16,7 @@ var outputPath string
 var jsonOutput bool
 
 var callCmd = &FuncCommand{
-	Name:  "call [ARGUMENTS] [FUNCTION]...",
+	Name:  "call [OPTIONS] [FUNCTION]...",
 	Short: "Call a module function",
 	Long: strings.ReplaceAll(`Call a module function and print the result.
 
--- a/cmd/dagger/call.go
+++ b/cmd/dagger/call.go
@@ -16,7 +16,7 @@ var outputPath string
 var jsonOutput bool
 
 var callCmd = &FuncCommand{
-	Name:  "call [flags] [FUNCTION]...",
+	Name:  "call [ARGUMENTS] [FUNCTION]...",
 	Short: "Call a module function",
 	Long: strings.ReplaceAll(`Call a module function and print the result.
 
--- a/cmd/dagger/config.go
+++ b/cmd/dagger/config.go
@@ -82,7 +82,7 @@ dagger config -m github.com/dagger/hello-dagger
 }
 
 var configViewsCmd = configSubcmd{
-	Use:   "views [name]",
+	Use:   "views [NAME]",
 	Short: "Get or set the views of a Dagger module",
 	Long:  "Get or set the views of a Dagger module. By default, print the views of the specified module.",
 	PersistentFlags: func(fs *pflag.FlagSet) {
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -43,7 +43,7 @@ var funcCmds = FuncCommands{
 }
 
 var funcListCmd = &FuncCommand{
-	Name:  "functions [ARGUMENTS] [FUNCTION]...",
+	Name:  "functions [OPTIONS] [FUNCTION]...",
 	Short: `List available functions`,
 	Long: strings.ReplaceAll(`List available functions in a module.
 
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -34,7 +34,7 @@ const (
 
 var funcGroup = &cobra.Group{
 	ID:    "functions",
-	Title: "Function Commands",
+	Title: "Functions",
 }
 
 var funcCmds = FuncCommands{
@@ -43,7 +43,7 @@ var funcCmds = FuncCommands{
 }
 
 var funcListCmd = &FuncCommand{
-	Name:  "functions [flags] [FUNCTION]...",
+	Name:  "functions [ARGUMENTS] [FUNCTION]...",
 	Short: `List available functions`,
 	Long: strings.ReplaceAll(`List available functions in a module.
 
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -287,6 +287,9 @@ func (fc *FuncCommand) execute(c *cobra.Command, a []string) (rerr error) {
 			cmd.PrintErrln("Error:", rerr.Error())
 
 			if fc.showHelp {
+				// Intentionally add newline to separate error from usage
+				cmd.PrintErrln()
+
 				// Explicitly show the help here while still returning the error.
 				// This handles the case of `dagger call --help` run on a broken module; in that case
 				// we want to error out since we can't actually load the module and show all subcommands
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -334,7 +334,7 @@ func toUpperBold(s string) string {
 	return termenv.String(upperCase).Bold().String()
 }
 
-const usageTemplate = `{{ "Usage" | toUpperBold }}:
+const usageTemplate = `{{ "Usage" | toUpperBold }}
 
 {{- if .Runnable}}
   {{.UseLine}}
@@ -345,28 +345,28 @@ const usageTemplate = `{{ "Usage" | toUpperBold }}:
 
 {{- if gt (len .Aliases) 0}}
 
-{{ "Aliases" | toUpperBold }}:
+{{ "Aliases" | toUpperBold }}
   {{.NameAndAliases}}
 
 {{- end}}
 
 {{- if isExperimental .}}
 
-{{ "EXPERIMENTAL" | toUpperBold }}:
+{{ "EXPERIMENTAL" | toUpperBold }}
   {{.CommandPath}} is currently under development and may change in the future.
 
 {{- end}}
 
 {{- if .HasExample}}
 
-{{ "Examples" | toUpperBold }}:
+{{ "Examples" | toUpperBold }}
 {{ .Example }}
 
 {{- end}}
 
 {{- if .HasAvailableLocalFlags}}
 
-{{ "Flags" | toUpperBold }}:
+{{ "Flags" | toUpperBold }}
 {{ flagUsagesWrapped .LocalFlags | trimTrailingWhitespaces}}
 
 {{- end}}
@@ -374,7 +374,7 @@ const usageTemplate = `{{ "Usage" | toUpperBold }}:
 {{- if .HasAvailableSubCommands}}{{$cmds := .Commands}}
 {{- if eq (len .Groups) 0}}
 
-{{ "Available Commands" | toUpperBold }}:
+{{ "Available Commands" | toUpperBold }}
 {{- range $cmds }}
 {{- if (or .IsAvailableCommand (eq .Name "help"))}}
 {{cmdShortWrapped .}}
@@ -384,7 +384,7 @@ const usageTemplate = `{{ "Usage" | toUpperBold }}:
 {{- else}}
 {{- range $group := .Groups}}
 
-{{.Title | toUpperBold}}:
+{{.Title | toUpperBold}}
 {{- range $cmds }}
 {{- if (and (eq .GroupID $group.ID) (or .IsAvailableCommand (eq .Name "help")))}}
 {{cmdShortWrapped .}}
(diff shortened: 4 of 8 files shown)


helderco:
> Fixes #6943 (partially for now, but foundation)

A separate PR for each item would actually be preferred, especially if it can stand on its own or it may warrant some discussion or design decision.

---

> * [x]  **Remove discrepancy between `dagger query` and the rest of the commands**
>   This sub-command hasn't been updated in months and having a separate template is more coordination work. It might be time to make them 1:1 ?

👍

---

> * [x] **Conventionalize on either `<command>` or `COMMAND` syntax in cmd.Use**
>   I updated all COMMAND as uppercase. I did not remember if you were also referencing on the usage examples inside square brackets like `[arguments]` -> `[ARGUMENTS]`  ? I followed the CLI examples we discussed and some did it that way, in doubt I applied it everywhere but totally open for change

Yes, it's applied everywhere. The idea is to choose a syntax for representing usage and stick to it. Last time I changed to `[OPTIONS]` but than reverted back to `[flags]` because while I prefer the former for avoiding confusion with static argument names and user provided values in some cases, Cobra uses the latter syntax. So it can seem like swimming against the current because you need to override some default behavior. The Docker CLI does it, but I’m not sure it’s worth it for us.

The `gh` tool uses Cobra’s default, maybe copy that as it’s more common anyway, I believe. I suggest looking at `git` too since there’s lots of examples there.

---

> * [ ]  **Figure out if arguments are better above or below functions/commands ??**
>   I experimented on this one, and I haven't found an easy way: on a given function call, I do agree that having the commands at the bottom does make sense. However, changing the order on the template directly also changes the order for commands such as: `dagger --help`. _What do you think is the best implementation on this one ? Shall we tweak the template to make it specific for just the `dagger call` section ?_ I feared adding complexity to the template logic

Cobra’s default is to put flags after subcommands. In some cases it seemed like I was looking for the flags next to the usage line, so I experimented with moving them there. That’s that `docker` does, but you’ll find a hard time finding a `docker` command with flags **and** sub-commands (example: `docker compose --help`).

However, when you have a **lot** of sub-commands, as when you do  `dagger call` on a function that returns a container, then it’s much harder to find the flags because that long list will occupy the whole screen. Doubly confusing to have “Global flags” at the bottom since that “hints” that non-global flags would be next to them.

Note that with `docker`, in this case, you only see “Global options” in the root  command (`docker --help`), not in any children. That inspired me to do the same.

---

> * [ ]  **Visual cue for required flags**
>   As you proposed last time, the idea would be to have a required flags section for the `dagger call function` subcommands only ? Will take a look tomorrow. Same fear on the added complexity on the main template, but this one seems more manageable: it is just a section wrapped in an if to show it on `dagger call` subcommand context

I wouldn’t put an `if` in the template looking for the `call` command specifically. In this case it can be more general. Rather than breaking in two sections because of required, just append `[required]` to flags marked as required, same as what’s done for default values, and make sure they’re sorted before the others (better in its own PR).

It’s what [Typer](https://typer.tiangolo.com/tutorial/options/required/) does, for example:

```
Usage: app [OPTIONS] NAME  

Options:  
  --lastname TEXT       Last name of the person [required]
  --help                Show this message and exit.
```

The idea of a “required option” can be strange though, but at least right now that only happens in `dagger call`, which brings us to the next point.

---

> * [x]  **"Options" instead of "Flags"?[1](https://github.com/dagger/dagger/issues/6943#user-content-fn-1-6303e1ecaa2e73e24edcc8788f13dcde)**
>   This is done, with our own implementation of `useLine` to print options instead of args by default.
>   Solomon requested `arguments` on above issue though. I followed your terminology, but shall we make a specific case for arguments in `dagger call` / `dagger functions` scope ?

`dagger functions` doesn’t show any function arguments. It doesn’t even create sub-commands so they don’t show in `--help` as well. It works by normal stdout so no customization necessary here.

Also, remember that `dagger call` (top level command) is the only command that can have two kinds of flags. An actual flag like `--output` and flags from **constructor** arguments. But for any `dagger call <function>` sub-command, it’s just function arguments.

If it was just the latter, it would be simply a matter of customizing the name of the section via an annotation, for example. But I think there’s a value in separating the flags in `call` and the constructor arguments.

How to do that? If you see how marking a flag as required works, Cobra is adding an [annotation through the flagset](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.SetAnnotation). So you can just create an annotation for marking a flag as a “function argument”, and in the template, similarly to how you handle required flags, you split argument flags, and if there’s any, show a section below “Options” for “Arguments”.

Both sections will only show at the same time in the top-level `dagger call` and flags marked as “required" will only show under `Arguments` so there’s no “required option” incongruity in practice.

I suggest three separate PRs then:
1. Replace “Flags” with “Options”
2. Append `[required]` to a flag marked as required, and sort the list of flags with the required ones first
3. Adopt “Functions” and “Arguments” terminology in `dagger call`

---

> * [ ]  **Help topics for dagger call to keep --help lean while still providing more documentation**
>   What topics do you think we should cover first ? More than the implementation, the most important might be the documentation added to those topics. Which could be the first 2 topics to create, so that i draft the implementation tomorrow ?

Not a priority, better create a separate issue for this. The intent is to provide more information to the user in the CLI (i.e., without requiring a visit to the docs), while reducing the clutter in normal `--help`. A good example for `dagger call` is how it works through chaining, what happens with a few core types (`Contaienr`, `Directory`…), how to use `--output` and `--json`, etc. 

`gh` uses help topics for this, but `docker` relies on web documentation, doing its utmost to keep `--help` output to a minimum. I didn’t find any *help topic* being used (as a feature) in that CLI.

Since Dagger has its nuances, I’ve been wondering which approach is better here. We could just “link” to the web for more information. \cc @vikram-dagger

---

> * [ ]  **Global flags only on root command and top level commands**
>   
>   * Remove persistent flags from child commands after being parsed? Or just hide?
>     
>     * TUI (--silent, --focus)
>     * Modules (-m)
> 
> This is the one I am currently trying to implement in a clean way (using the hiding method at first), and might need some guidance again. I have been trying to extend the `callCmd` implementation by implementing a new `persistentPreRunE`:
> 
> ```go
> PersistentPreRunE: func(_ *FuncCommand, cmd *cobra.Command, _ []string) error {
>     rootCmd := cmd
>     for rootCmd.Parent() != nil {
>         rootCmd = rootCmd.Parent()
>     }
>     hideGlobalFlags(rootCmd)
>     return nil
> },
> ```
> 
> to control the behavior and potentially use the funcCommand infos or the cmd infos to mark as hidden, as you suggested, when the call command has subcommands However, I'm still having a hard time on this part. Will keep iterating tomorrow again

The way I’ve envisioned this is through `main.go`, for all commands. Not just `call`. `docker` hides them for any command that’s not the root command. So you’d only see `--debug` and `--silent` for example, in `dagger --help` but not `dagger call`. 

I think some of those can be useful at “top level”  though (direct child of root), especially `-m,--mod`, but has been painful only in `dagger call <function>` sub-commands actually. Additionally, some of the global flags have already been either hidden or removed since I first added this to the item list.

So I’d just not add these flagsets to “function” sub-commands in `call` (not simply hide), which also helps prevent collisions with a user’s argument name choices. Including `-m,--mod`

---

> * [x]  **Talk about Functions and arguments instead of Function commands**
>   This is partially done as I used your `options` terminology globally instead of `arguments`. It is true that `arguments` is pretty accurate in a `dagger call functions` context, but then, do we talk about `arguments` inside `dagger functions` usage examples too ?

Already talked about this in a previous point. To reiterate, "Options" is also closer to *function arguments* than “Flags”, but it’s better to have constructor arguments in `dagger call` be separate from that command’s own flags like `--output`, and it also avoids having “required options”, since that only happens for function arguments.

None of this applies to `dagger functions` since it works differently. It’s not through `--help` like `call`. There’s no child subcommands and not flags from function arguments, just normal positional arguments to the `functions` command.

helderco:
@grouville, with "Figure out if arguments are better above or below functions/commands", I realize I wasn't clear in addressing your question. The answer to doing it only for `call` is no. Do it for every command, i.e., basically revert this commit: https://github.com/dagger/dagger/pull/6549/commits/a8ae912c52437e5be755bb38ee48c29a449a9ebd

Note that only `dagger call` may have a long list of sub-commands. But if there's a command (like `dagger config`) that can also have sub-commands, it won't be that big, so I don't think you'll notice that much if the flags are below or above the sub-commands as they should be easy to find. So reposiitioning for all commands is the best option.

Issue #6706, opened 2024-02-21
Go SDK issues with newer go.mod than packaged into engine


jedevc:
Writing up a brain dump, so might be a bit messy.

Currently the go version used by the Go SDK is `1.21.3`. However, a user module can set a custom go version (using `go.mod`/`go.work`) to a higher version, like `1.22.0`.

Now, this can cause fun issues it seems? I'm not quite sure what happens, but essentially in https://github.com/dagger/dagger/blob/b2450fc1c61dfff5f04fe1679bc09b412c33a7e0/cmd/codegen/generator/go/generator.go#L292-L302

We can get a package returned that:
```json
{
  "ID": "foobar",
  "Name": "main",
  "PkgPath": "foobar",
}
```
for a `go.mod` that looks like this:
```go
module foobar

go 1.22.0
```

This silent failure is caused by https://github.com/golang/tools/blob/a821e617f30094a078bfdc1d63623b93365b3130/go/packages/packages.go#L1039 - `Checker.Files` *can` return an error if we try and type-check for a go version that is greater than the currently installed version.

This behavior makes *sense* (though the silent failure really doesn't). Note: `GOTOOLCHAIN=auto` (as added in #6652 doesn't help for this, it only helps for compiling, not parsing+processing the package).

However, this leaves us with the problem of how to manage this from the point-of-view of the Go SDK (cc @vito @sipsma). A few options:
- We simply explicitly ban doing this - if the go version bundled into the Go SDK tarball is X, then we don't allow >X with a nice error message.
	- Note, we could potentially decouple the codegen go version from the engine go version to be higher with very little risk.
	- However, since the Go SDK is built into the engine, we'd have to re-release every time we want to bump go version, which is a pain (but not unavoidable at our pace of releases).
- We rework package loading significantly, instead somehow custom compiling a binary at runtime to do this (with the right go-version), then serializing + deserializing this into the codegen binary.
	- This would go poorly if the types structure changes, not sure if the go compatibility requirement allows this.
	- Also, complexity :sob:
- Just attempt to parse it as best we can, by explicitly overriding the go version for typechecking - if we fail, append a warning about the upgrade, but otherwise continue on!
	- This could just work, there are really minor type differences between versions anyways.
	- Not quite sure how we implement this though, hopefully we don't need to mess around with any `x/tools` internals.

Open to other suggestions :tada:

jedevc:
[...]

PR #7109, opened 2024-04-17
chore: update images to go 1.22

jedevc:
This patch updates all the go base images to versions that have go 1.22.

Note! We can't update the version used to build runc, due to an upstream incompatability - however, because the build is separate, we can keep building it using an older version of go.

Also note, we can't *yet* bump to actually use 1.22 in our code by updating the go.mod. This is because the `ci` package must continue using 1.21 until the next release (insert elaborate codegen reasons here) - and then the `ci` package depends on the top-level package, so that must also stay at 1.21. There *is* a potential workaround that would allow us to split out `distconsts` into it's own go.mod, and then we could do this, but this can be a follow-up if we really want this.

### DIFF
--- a/.github/workflows/_hack_make.yml
+++ b/.github/workflows/_hack_make.yml
@@ -30,7 +30,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.21"
+          go-version: "1.22"
           cache-dependency-path: "ci/go.sum"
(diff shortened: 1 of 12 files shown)


jedevc:
The main rationale for wanting to do this is that without this, using a `go.mod` with go 1.22 can't work :cry: See https://github.com/dagger/dagger/issues/6706 for the entire logic.

vito (approved):

Issue #6452, opened 2024-01-18
Potential issues introduced by `fix: skip head merge commits on pull requests`


jedevc:
A small issue potentially introduced by https://github.com/dagger/dagger/pull/6253 (cc @grouville @marcosnils)

So, prior to this PR, we didn't perform any network operations when starting a dagger session. However, after this, we start doing git operations. My worry is that this has introduced subtle bugs with how this works.

I've seen this weird message appear fairly consistently throughout our logs (failure https://github.com/dagger/dagger/actions/runs/7571487756/job/20619120120?pr=6451, success https://github.com/dagger/dagger/actions/runs/7653266578/job/20854758251?pr=6496) - not quite sure how/why we see this though: 
```
1: > in init
1: starting engine 
Error: timed out waiting for session params: Error fetching branch: error fetching branch from origin: exit status 128Error fetching branch: error getting commit: object not found1: connect
```
```
: > in init
1: starting engine 
1: starting engine [0.00s]
1: starting session 
Error fetching branch: error fetching branch from origin: exit status 128Error fetching branch: error adding fork as remote: exit status 31: [0.40s] OK!
1: starting session [0.40s]
1: connect DONE
OK!
```

This is pretty consistent - the error *doesn't* appear to be fatal (the fatal case is caused by an internal infra issue and is being investigated separately)

grouville:
Nice one 🙏

marcosnils:
strange that we're seeing that now since that has been merged  a while back 🤔
once thing that I've noticed is that we never added a timeout to the git fetch operation. It'd be nice to have a timeout there and probably just log the error and continue executing the pipeline?

jedevc:
My initial assessment was wrong :tada:

There are two issues at play here:
- We see `Error fetching branch: error fetching branch from origin: exit status 128Error fetching branch: error getting commit: object not found1` on seemingly *every* run: https://github.com/dagger/dagger/actions/runs/7653266578/job/20854758251?pr=6496
	- These messages should not be printed - though the issue appears to be non-fatal.
- The `Error: timed out waiting for session params` is related to an internal infrastructure issue with our cloud caching service.

I'll update the issue description to focus instead on the first issue.

marcosnils:
> * The `Error: timed out waiting for session params` is related to an internal infrastructure issue with our cloud caching service.

[...]

PR #6253, opened 2023-12-12
fix: skip `head merge commits` on pull requests

grouville:
Fixes https://linear.app/dagger/issue/DEV-3086/pull-user-commit-message-in-a-context-pr

When collecting labels on the engine in a VCS context, sometimes, the commit head is an appended commit (from that VCS) [PR / MR]. This fix skips the merge request commit when it is the head.

Repro/test: https://gist.github.com/grouville/6cf442d86c41c0746bbef1812160e7c0

### 12/13 Advancements

> Realization: only GitHub puts the PR in a detached HEAD with only the merge commit. GitLab keeps the git history of the branch being tested

After reviews + some testing, a few changes have been implemented:
1. Make sure that the main branch doesn't skip the merge commits. Only `pull_requests` shall be be impacted
2. Due to the realization, we need to fetch the branch and get the commits from there. However, this is an operation requiring authorization. To circumvent this, we rely on the fact that GitHub runners automatically add the `GITHUB_TOKEN` secrets to git CLI commands

### 12/15 Advancements
Previous implementation was introducing a bug by printing errors to stdout, which polluted the `dagger listen` connection string.

The error came from the fact that on Dagger, the PRs come from a fork and not the origin. However, after spending some time reverse engineering Github Actions' runner context, I did not find a 100% solution to retrieve the origin of a fork. This is a best try approach, where we build the fork's remote of a repo from available information. If we do not manage to do it, then we fallback on the current logic

PR #7111, opened 2024-04-17
fix: ensure we get correct data for prs

jedevc:
Fixes https://github.com/dagger/dagger/issues/6452.

Instead of messing around with a bunch of remotes, etc, we can apply this simple little trick.
    
`GITHUB_REF` contains the "the fully-formed ref of the branch or tag that triggered the workflow run" (see https://docs.github.com/en/actions/learn-github-actions/contexts#github-context).
    
Now, if the ref starts with "refs/pull/", it's a pull request. These usually end in "/merge", to indicate that this is the result of merging with the base branch. However, we can just swap this "/merge" for a "/head" to get the actual head - then we can grab all the interesting commit details from there.
    
What's really cool is that these should just all be available on the "origin" - GitHub automagically populates them for us.

### DIFF
--- a/telemetry/labels.go
+++ b/telemetry/labels.go
@@ -1,6 +1,7 @@
 package telemetry
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"encoding/base64"
 	"errors"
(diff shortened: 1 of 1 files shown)


jedevc:
cc @grouville who wrote the original implementation of this in https://github.com/dagger/dagger/pull/6253.

telemetry/labels.go
```diff
@@ -395,75 +388,31 @@ func (labels Labels) WithAnonymousGitLabels(workdir string) Labels {
...
 	err := cmd.Run()
 	if err != nil {
-		return nil, fmt.Errorf("error adding fork as remote: %w", err)
+		return nil, fmt.Errorf("error fetching branch from origin: %w\n%s", err, output.String())
 	}
+	output.Reset()
 
-	cmd = exec.Command("git", "fetch", "--depth", "1", "fork", branch)
+	// Get the commit hash of the fetched remote
+	cmd = exec.Command("git", "rev-parse", "FETCH_HEAD")
```
vito, line 404: Would this work?:

```suggestion
	cmd = exec.Command("git", "rev-parse", target)
```

Since this runs on the host, just trying to avoid any funny business with parallel fetches.

vito (approved):
:tada:

grouville:
Impressed, how did you find out this trick ? 🤯 I love it 👏

PR #7101, opened 2024-04-16
Improve releasing during v0.11.1

jedevc:

PR #7117, opened 2024-04-18
chore: Add ci module `dagger develop` info to README

gerhard:
Required after a fresh clone.

Also:
- Bump ci module Engine version to v0.11.1
- Remove `$` from ci README code snippets

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -20,6 +20,17 @@ Available functionality:
 > engine) as defined in [`.github/workflows/_hack_make.yml`](../.github/workflows/_hack_make.yml).
 > Without this, you may hit unexpected errors or other weird behavior.
 
+## Developing after a fresh clone
+
+If you want to develop this module following a repository clone, remember to
+run the following one-off command:
+
+    cd ..
+    dagger develop --sdk go --source ci
+
+This will re-create all the files required by your code editor - see
+`.gitignore` for a list of what they are.
+
 ## Tests
 
 Run all tests:
--- a/ci/README.md
+++ b/ci/README.md
@@ -35,17 +35,17 @@ This will re-create all the files required by your code editor - see
 
 Run all tests:
 
-    $ dagger call --source=. test all
+    dagger call --source=. test all
 
 Run a specific test (e.g. `TestModuleNamespacing`):
 
-    $ dagger call --source=. test custom --run="^TestModuleNamespacing" --pkg="./core/integration"
+    dagger call --source=. test custom --run="^TestModuleNamespacing" --pkg="./core/integration"
 
 ## Dev environment
 
 Start a little dev shell with dagger-in-dagger:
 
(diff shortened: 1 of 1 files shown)


ci/go.mod
```diff
@@ -1,12 +1,12 @@
 module github.com/dagger/dagger/ci
 
-go 1.21.7
+go 1.22
```
jedevc, line 3: This can't be bumped yet. See https://github.com/dagger/dagger/pull/7109#issue-2248420383.

ci/go.mod
```diff
@@ -1,12 +1,12 @@
...
 
 replace github.com/dagger/dagger => ../
 
 require (
-	dagger.io/dagger v0.11.0
-	github.com/dagger/dagger v0.11.0
+	dagger.io/dagger v0.11.1
+	github.com/dagger/dagger v0.11.1
```
jedevc, line 9: Also relevant here: #7101

ci/go.mod
gerhard: Yup, just pushed a change to `1.21`. `1.21.7` is failing for me:
image

How did this work for you?

ci/README.md
```diff
@@ -20,63 +20,74 @@ Available functionality:
...
 
-    $ dagger call --source=. engine service --name=dagger-engine up --ports=1234:1234
+    dagger call --source=. engine service --name=dagger-engine up --ports=1234:1234
 
 Connect to it from a dagger cli:
 
-    $ export _EXPERIMENTAL_DAGGER_RUNNER_HOST=tcp://0.0.0.0:1234
-    $ dagger call -m github.com/shykes/daggerverse/hello@main hello
+    export _EXPERIMENTAL_DAGGER_RUNNER_HOST=tcp://0.0.0.0:1234
+    dagger call -m github.com/shykes/daggerverse/hello@main hello
```
jedevc, line 73: If we remove the `$` from here, there's no way to distinguish between the prompt and the output from it.

I know github has a way for copy-pasting code blocks, so if we want to do that, we need to remove the output from a command (or move it into a separate code block afterwards).

ci/go.mod
gerhard: Merged! Rebasing now.

ci/go.mod
jedevc: :thinking: that's weird. The format did change at some point, is your version of the go language server up to date?

That said, I think it's fine to have it set to 1.21.

gerhard: I found the issue. It was incorrect ordering in `$PATH`, which was picking up the wrong `gopls`. All fixed now!

ci/README.md
gerhard: I see what you mean. Adding a comment will fix it.

jedevc (approved):

Issue #7114, opened 2024-04-17
🐞 On Windows, 'dagger init'  fails if it does not find or create certain expected files


scotthain:
### What is the issue?

When running `dagger init` in an emply directory, it will fail with an error message if the directory does not have a .git directory or a dagger.json.

There is another issue that is related but not the same, I'll file another issue for that one and link it back. (edit: #7115 )

I traced it back in the dagger codebase to the buildit module's `filesync.go`. Based on something I ran into years ago, my suspicion is that the CreateFile needs either a UNC path or double slashes on Windows. I wasn't able to get a windows build working to debug further though!

### Dagger version

dagger v0.11.1 ([registry.dagger.io/engine](http://registry.dagger.io/engine)) windows/amd64

### Steps to reproduce

```
mkdir dagger-test
cd .\dagger-test\
dagger init
```

### Log output

```
[scott][~\code\dagger-test]> dagger init
✔ connect 11.3s
✘ ModuleSource.resolveContextPathFromCaller: String! 0.0s
! failed to find up root: failed to lstat .git: failed to receive file bytes message: rpc error: code = Unknown desc = stat path: CreateFile C:/Users/scott/code/dagger-test/.git: The system cannot find the file specified.

Error: failed to get configured module: failed to get local root path: input: moduleSource.resolveContextPathFromCaller resolve: failed to find up root: failed to lstat .git: failed to receive file bytes message: rpc error: code = Unknown desc = stat path: CreateFile C:/Users/scott/code/dagger-test/.git: The system cannot find the file specified.
```

jedevc:
Argh, yeah, this will be the result of https://github.com/dagger/dagger/pull/7003 - it's apparently still an issue. I'm not quite sure what happened here? I could have *sworn* I tested this, though maybe in a slightly different environment.

Regardless, can confirm this is an issue, the trick here is that for some reason, while we're encoding paths to be `filepath.ToSlash` over-the-wire, we're not un-encoding them using `filepath.FromSlash` in any consistent way. What a pain, but at least the fix should be pretty simple - will take a look at this tomorrow :tada:

Thanks for the quick report! :tada: 🎉

scotthain:
Awesome! I was going to try and hop on it myself but I couldn't get the build to... build with dagger on Windows 😄 and ran out of time to play with it today.

jedevc:
Oops, #7118 should have fixed this :tada: Closing now!

[...]

PR #7003, opened 2024-04-03
fix: windows dagger init command

jedevc:
Fixes https://github.com/dagger/dagger/issues/6923 :tada:

> [!WARNING]
> Depends on https://github.com/tonistiigi/fsutil/pull/196

Still needs some tidy-ups, but I managed to actually get this working on windows again!

Love filepaths. So fun.

image

PR #7118, opened 2024-04-18
fix: all paths should be encoded over-the-wire

### LINKED ISSUE #6923: 🐞 Dagger init can't run on windows
### What is the issue?

Cannot run `dagger init` on windows

seems it has problem with `filepath.IsAbs` for windows

### Dagger version

dagger v0.10.2 windows/amd64

### Steps to reproduce

1. install dagger CLI for windows
2. run command `dagger init`

### Log output

from cli

```
Error: failed to get configured module: failed to get local root path: input: moduleSource.resolveContextPathFromCaller resolve: failed to find up root: path is not absolute: C:\projects\test
```

from dagger engine

```
2024-03-25 16:24:26 time="2024-03-25T09:24:26Z" level=warning msg="error resolving moduleSource(refString: \".\").resolveContextPathFromCaller: String!" error="map[error:failed to find up root: path is not absolute: C:\\projects\\test kind:*fmt.wrapError stack:<nil>]" source="&{github.com/dagger/dagger/tracing.AroundFunc.ProgrockAroundFunc.func1 /app/tracing/graphql.go 107}"
```

jedevc:
Fixes #6923 #7114 (second time's the charm hopefully), follow-up to #7003.

*Right*. Descriptions are split up into commits to make it a bit easier to understand, but TL;DR, there were two main issues:
- Paths weren't being correctly decoded after being encoded on the wire - I should have caught this and didn't (my bad). The correct way to do this is to handle the encoding/decoding as close as possible in the code to sending/receiving, then it doesn't sneak everywhere through the code.
- Windows has different file-not-found message strings for `os.Stat` and `os.ReadFile` (for reasons, I have no idea). We were missing a check for the message you get from `os.ReadFile` ("The system cannot find the file specified"). Instead of adding *yet another check*, I move the check to the client with `errors.Is(err, os.ErrNotExist)`, and then transforming this into a nice gRPC status code (but need to preserve the other one for backwards-compat, yay).
	- Also, as part of this, I discovered that we weren't doing error-wrapping properly in loads of places (so the status codes didn't get propagated nicely). To solve this, I added a linter check to *always* use the `%w` formatting directive where possible, and then fixed *all* the instances in the codebase.

Windows is hard :cry:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,6 +5,7 @@ linters:
     - dogsled
     - dupl
     - exportloopref
+    - errorlint
     - gocritic
     - gocyclo
     - gofmt
@@ -46,4 +47,9 @@ linters-settings:
 
   govet:
     enable:
-      - nilness
+
+  errorlint:
+    errorf: true
+    errorf-multi: false
+    asserts: false
+    comparison: false
--- a/ci/engine.go
+++ b/ci/engine.go
@@ -119,17 +119,26 @@ func (e *Engine) Service(
 }
 
 // Lint the engine
-func (e *Engine) Lint(ctx context.Context) error {
+func (e *Engine) Lint(
+	ctx context.Context,
+	// +optional
+	all bool,
+) error {
 	pkgs := []string{""}
 	// pkgs := []string{"", "ci"}
 
 	ctr := dag.Container().
 		From(consts.GolangLintImage).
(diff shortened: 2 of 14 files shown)


scotthain:
Nice catch on the unwrap! Thanks for the fix - code looks good to me from my 100-foot view 🚀 🍰

PR #7121, opened 2024-04-18
fix: parse go pragmas with CRLF line endings

jedevc:
Pragmas were not being correctly parsed when we had windows line endings - this fix ensures that we correctly handle these.

This was resulting in some *very* weird to debug issues where optional arguments could be detected as required instead.


vito (approved):
lgtm!\r\n

sipsma (approved):

PR #7123, opened 2024-04-18
docs: Fixes Dagger version in curl examples and links

vikram-dagger:


jedevc (approved):

jedevc:
@vikram-dagger I've also updated our `RELEASING` guide, so this should now be kept up to date as part of our release workflow :tada:

vikram-dagger:
> @vikram-dagger I've also updated our `RELEASING` guide, so this should now be kept up to date as part of our release workflow 🎉

@jedevc see similar issue in https://github.com/dagger/dagger/pull/7164, which I corrected in a different way. I'm not sure which is better /cc @gerhard for his opinion

PR #7124, opened 2024-04-18
cli: remove special template on query subcommand

grouville:

`dagger query` is not the main entrypoint of the dagger CLI anymore. Even though the special template made some sense, it is a code duplication to maintain. With the upcoming changes (titles in bold, etc ...) and the remaining presence of the example on the command's help, it does make sense to remove it and keep a single source of truth

#### New output
```shell
./dagger query --help
Send API queries to a dagger engine.

When no document file is provided, reads query from standard input.

Can optionally provide the GraphQL operation name if there are multiple
queries in the document.

Usage:
  dagger query [flags] [OPERATION]

Aliases:
  query, q

Examples:
dagger query <<EOF
{
  container {
    from(address:"hello-world") {
      withExec(args:["/hello"]) {
        stdout
      }
    }
  }
}
EOF


Flags:
      --doc string        Read query from file (defaults to reading from stdin)
  -m, --mod string        Path to dagger.json config file for the module or a directory containing that file. Either local
                          path (e.g. "/path/to/some/dir") or a github repo (e.g. "github.com/dagger/dagger/path/to/some/subdir")
      --var strings       List of query variables, in key=value format
      --var-json string   Query variables in JSON format (overrides --var)

Global Flags:
  ```

cc @marcosnils, as he was the original author of this

### DIFF
--- a/cmd/dagger/query.go
+++ b/cmd/dagger/query.go
@@ -115,37 +115,6 @@ func getKVInput(kvs []string) map[string]interface{} {
 }
 
 func init() {
-	// changing usage template so examples are displayed below flags
-	queryCmd.SetUsageTemplate(`Usage:{{if .Runnable}}
-  {{.UseLine}}{{end}}{{if .HasAvailableSubCommands}}
-  {{.CommandPath}} [command]{{end}}{{if gt (len .Aliases) 0}}
-
-Aliases:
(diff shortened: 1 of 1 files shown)


helderco (approved):
🚢

PR #7126, opened 2024-04-18
cli: style headings in BOLD UPPERCASE to help break sections visually

grouville:
Part of the de-clutter CLI implementation: https://github.com/dagger/dagger/issues/6943#issuecomment-2021310987.

Suggested by @helderco and looks (in my sense) better. Takes inspiration from the `gh` CLI to visually break the heading sections. It relies on a helper function directly used on the template.

New style:
image


helderco (approved):
That's it! 🎉

PR #7127, opened 2024-04-18
cli: line break between error and usage

grouville:
Part of the de-clutter CLI implementation: https://github.com/dagger/dagger/issues/6943#issuecomment-2021310987.

When querying `dagger call --help` and `dagger functions --help` without being in a module or for an inexistant function, an error is being raised.

This commit adds a newline between the usage and the error, as it was hard to distinct between them

Example:
```shell
./dagger -m . functions toto --help
✔ initialize 41.2s

Error: no function 'toto' in object type 'Dagger'
```

### DIFF
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -287,6 +287,9 @@ func (fc *FuncCommand) execute(c *cobra.Command, a []string) (rerr error) {
 			cmd.PrintErrln("Error:", rerr.Error())
 
 			if fc.showHelp {
+				// Intentionally add newline to separate error from usage
+				cmd.PrintErrln()
+
 				// Explicitly show the help here while still returning the error.
 				// This handles the case of `dagger call --help` run on a broken module; in that case
 				// we want to error out since we can't actually load the module and show all subcommands


helderco:
Err... seems like this has been fixed in v0.11 actually, with the change in TUI and the move from progrock to open telemetry! 

```console
❯  ./dagger-0.10.3 call toto --help
✔ initialize 6.3s
  ✔ Module.initialize: Module! 0.6s
    ✔ exec /runtime 0.2s
Error: unknown command "toto" for "dagger call"
```

```console
❯ ./dagger-0.11.0 -vv call toto --help
✔ connect 1.0s
  ✔ starting session 0.2s
  ┃ OK!
✔ initialize 0.5s
  ✔ Module.initialize: Module! 0.1s

Error: unknown command "toto" for "dagger call"
```

I increased verbosity to more closely match the previous version. No difference with the change in this PR though:
```console
❯ dagger-dev call toto --help
✔ initialize 2.0s

Error: unknown command "toto" for "dagger call"
```
If you agree, this can be closed 🙂

grouville:
Correct ✅ 🙏

grouville:
@helderco I found one case where this is still not true: on all commands where `ExactArgs` is set.
_E.g._
```
Args:    cobra.ExactArgs(1),
```

Example: 
```shell
./dagger install         
Error: accepts 1 arg(s), received 0
USAGE
  dagger install [flags] MODULE

ALIASES
  install, use

EXAMPLES
dagger install github.com/shykes/daggerverse/ttlsh@16e40ec244966e55e36a13cb6e1ff8023e1e1473

FLAGS
  -m, --mod string    Path to dagger.json config file for the module or a directory
                      containing that file. Either local path (e.g. "/path/to/some/dir") or
                      a github repo (e.g. "github.com/dagger/dagger/path/to/some/subdir")
  -n, --name string   Name to use for the dependency in the module. Defaults to the name of
                      the module being installed.

GLOBAL FLAGS
```

Is it ok as is and not worth unifying ?

helderco:
That's actually [Cobra's default](https://cobra.dev/#usage-message). Compare with:

```
$ cobra --invalid
Error: unknown flag: --invalid
Usage:
  cobra [command]
```

So our change in the headings makes it stand out a bit more, maybe:

<pre>
$ dagger install
Error: accepts 1 arg(s), received 0
<b>USAGE</b>
  dagger install [flags] MODULE
</pre>

It's still worthwhile to add that new line, but it'll take a larger change.

## Technical details

The reason you're seeing it with `ExactArgs` is because argument validation happens before running `cmd.Run`. More specifically, it’s before `PersistentPreRunE`, which disables showing the usage on an error:

https://github.com/dagger/dagger/blob/90de6b62df7727da3a401826a20bfa80b8a53ac9/cmd/dagger/main.go#L103-L105

It’s the same with flag parsing:

```
dagger install --toto
```

When the usage is not silenced on errors, Cobra prints the error to  `stderr` and then the usage to `stdout`.

```go
// If root command has SilenceErrors flagged,
// all subcommands should respect it
if !cmd.SilenceErrors && !c.SilenceErrors {
    c.PrintErrln(cmd.ErrPrefix(), err.Error())
}

// If root command has SilenceUsage flagged,
// all subcommands should respect it
if !cmd.SilenceUsage && !c.SilenceUsage {
    c.Println(cmd.UsageString())
}
```

You can’t get in the middle of that without a global `SilenceUsage: true` and printing the error ourselves, which is what we need to do.

But I’m already thinking on making changes that’ll affect this code anyway. I think we should show a more slimmed down version of the usage when there’s an error, versus when you’re asking for `--help`. So fixing this newline could be done at that time as well. 

Also, as soon as we've hid `PersistentPreRunE`, there's no more usage, so it's not very consistent. We need to figure that out as well, but can be a separate issue.

In any case, it’s not the solution presented in this PR, so let’s just open a new one rather than re-open.

Issue #6574, opened 2024-02-01
`Container.From` should ignore image entrypoint by default


shykes:
## Problem

As discussed on Discord:

> I find myself wishing that we did not honor OCI image entrypoints by default...
> I would much rather know that, by default, From("foo").WithExec("bar") will always actually execute the binary "bar" in the image "foo", unless I specifically tell Dagger that I care what the image author put in the entrypoint (most of the time I don't).

> Kind of like the way Kubernetes manifest ignores a bunch of fields in the docker image, and just makes you re-write them in the kub config. At the time it irked me, but now I get it. Clean slate. And more control over the overall experience for the wrapping tool (then: kube. now: dagger) 

@vito added:

> I struggled with this for a while in Bass, since there the syntax is ($ go build --foo) and typing ($ build --foo) because you know the entrypoint is go just felt nonsensical. I eventually had to add support for ENTRYPOINT/CMD to support publishing images, but had to square their semantics with everything else.
> https://github.com/vito/bass/releases/tag/v0.12.0 Ctrl+F "Entrypoint" if you're curious to see the direction I went

## Solution

Change the behavior of `Container.From` to *not* honor the image entrypoint by default. Make this configurable with a flag, to explicitly ask for it.

Or, alternatively, deprecate the support for entrypoint completely, except by raw inspection of the OCI image data (which we should support, by the way).

vito:
Just a note: we have `withExec(skipEntrypoint: true)` already, so this maybe could just be tantamount to flipping the default and adding a `useEntrypoint: true` for anything that needs the old behavior.

...and then dealing with the fallout. :boom:

helderco:
Yeah, I think doing this in `withExec` is better. This way the entrypoint is still available if you need it, but not in the way if you don't.

helderco:
Draft submitted, but it's an alternative solution to the one suggested in this issue, based on the above comments:

* [https://github.com/dagger/dagger/pull/7136](https://github.com/dagger/dagger/pull/7136)

Just something easy to take a breather 🙂 Doing bits on the side, but if there's no consensus on the solution it's not worth it to continue. So requesting feedback on that.

vito:
@helderco Nice - now I guess the question is when we pull the trigger. Maybe we should just get it over with? Doesn't seem like we gain much by waiting. :grimacing: 

Maybe we need to signal boost this decision to get more community input + spread awareness?

PR #7136, opened 2024-04-19
💥 api: skip entrypoints by default in `withExec`

helderco:
Fixes https://github.com/dagger/dagger/issues/6574 (although with an alternative solution)

> [!WARNING]
> It's a breaking change for anyone currently depending on the entrypoint in `withExec`, unless using `skipEntrypoint` (except in Go since you can't use zero values in options).

Flipped the default on `withExec` instead. Currently you need to explicitly skip the entrypoint, but with this change, you'll have to explicitly use it instead. Otherwise the entrypoint is still in the `Container` object, and preserved on `publish`.

It's **recommended** to avoid using the entrypoint in `withExec`. It's much more readable to be explicit here, especially when it's not clear what the entrypoint is from the line where `withExec` is defined.

## Deprecation path

Consider if the entrypoint is really necessary and add the command explicitly to `withExec`. This makes the code compatible before and after this change.

Otherwise, users can set `skipEntrypoint: false` today, which will preserve that same behavior after this change[^1], although this won't work in Go.

[^1]: Albeit deprecated until a few releases down the line, so users should be encouraged to change to `useEntrypoint: true` when able, or not rely on the entrypoint anymore.


shykes (approved):
I support this change, even though it is a breaking change.

Rationale:

- The whole point of Dagger is to express the logic of your containerized pipeline in code
- Using "entrypoint" by default makes less of your logic visible in the code. The answer to "what does this line do?" is now "well it depends on the container entrypoint". Better to be more explicit.
- The intent behind "entrypoint" was to make Docker containers more "programmable". It was part of a broader design roadmap that has since been abandoned. In other words, it's a relic. Dagger is the spiritual successor of Docker's abandoned roadmap to make the container not just a dumb artifact, but a programmable component. This change makes it easier for Dagger to get the job done.

shykes:
Sorry for the flip-flop, I am not necessarily against this change, but had wrongly assumed it was an implementation of #6574 . I know realize it's an alternative to it, so I need to think about the implications. (Currently discussing with @helderco on Discord)

shykes:
OK after having discussed with @helderco I am now on board with this implementation :)

To summarize our relationship to the concept of entrypoint in container images:

- Entrypoint is enshrined in the OCI spec. So whether we like it or not, we are stuck with it. The Dagger API should acknowledge its existence
- Dagger considers entrypoint to be "dumb" metadata on a container image. You can read it, and change it.
- But Dagger does not use it to configure its own execution of containers, because it doesn't need it: the same thing can be achieved with code.
- I am on the fence on the `useEntrypoint` optional argument. Technically this argument should not exist, but since querying the entrypoint is a little annoying in mainstream SDKs because of error checking, it's reasonable to allow its existence as a small convenience.

PR #7143, opened 2024-04-19
cli: conventionalize on `<command>` in cmd.Use

grouville:

Unifies the DX around usage across all commands. The convention follows the standards (gh / git CLIs) as well as the defaults from cobra. More context from @helderco in above comment

### Proposal

The grammar is as follow:
- []: optional
- <>: user input, to differenciate with static args
- ...: one or more

Another point: [flags] shall always be positioned directly after a command. Or, at least to have a consistent positionning across all commands.

### Context

This PR applies the grammar everywhere for consistency. And, fixes the `dagger gen` and `dagger config views` command where flags were not positioned in the same order as all the other commands.

#### 1. Open for debate / proposition

Both for `functions` and `call`, I took the liberty to modify the usage a bit to be more descriptive.
```shell
USAGE
  dagger functions [flags] [<function> [<child-function>...]]
```
```shell
USAGE
  dagger call [flags] [<function> [<child-function>...]]
  dagger call [flags] [<command>]
```

The origin of this change is that I got suprised when testing `dagger functions`: I tested to chain two functions at the same level, which the previous usage _could_ suggest:
```shell
dagger functions [flags] [<function>]...
```
I am totally fine on keeping that format too, especially if you feel that it is easier to read and less confusing overall. You are the DX master 👼 

> The confusing part can be showed with below example. With the previous usage, this could be understood as a valid command, where base and debug are two top-level functions from the mariadb module:
> ```shell
> dagger -m github.com/levlaz/daggerverse/mariadb@v0.2.1 functions base debug
> ```
> The proposed usage _might_ be more accurate / precise but less easy to understand at first

The terminology could also be: 
* `<child-function>`
* `<derived-function>`
* `<sub-function>`

#### 2. dagger config
It seems that the `dagger config` subcommand, which is still experimental and with hidden subcommands might need an issue on its own as the behavior of presenting sub-commands and the flags ordering seems a bit different.

## Update 04/23/2024
I added a new commit following Cobra's [usage](https://github.com/spf13/cobra/blob/main/command.go#L52-L60)
>  Recommended syntax is as follows:
   [ ] identifies an optional argument. Arguments that are not enclosed in brackets are required.
   ... indicates that you can specify multiple values for the previous argument.
   |   indicates mutually exclusive information. You can use the argument to the left of the separator or the
       argument to the right of the separator. You cannot use both arguments in a single use of the command.
   { } delimits a set of mutually exclusive arguments when one of the arguments is required. If the arguments are
       optional, they are enclosed in brackets ([ ]).
 Example: add [-F file | -D dir]... [-f format] profile
 
But without differentiating mandatory and optional arguments, as proposed in your comment below: https://github.com/dagger/dagger/pull/7143#issuecomment-2067519490. I also reverted the functions and the call usage as they were before

## Update 05/02/2024
Following the discussion on the issue, proposed syntax:
- [] optionality
- <> non optional
- every flag / positional argument has to have either [] or <> to differenciate with commands

Also, another inherent rule that you proposed:
- Do not show [options] on commands that just have inherited options


helderco:
`call` and `functions` are unique in that they have a dynamic and recursive amount of arguments/commands. Even if in `functions` they're positional arguments, unlike `call` where they're actual sub-commands, what trumps here is that they both represent the same function chaining that may take a new user a bit of getting used to or learn about.

For other cases like `dagger config`, just follow what `docker` does. Just `dagger config [flags] [command]`. That is:
- if it has flags add `[flags]`
- if it has sub-commands add `[command]`
- If you do `--help` on a sub-command, it's the same, just the path of the current command changes

That's it. It should actually be the default in the template.

Git for example, shows all the options in the usage. In our case I say only if there's a clear benefit in helping the user learn how to use the command, otherwise just leave the default.

grouville:
> ction chaining that may take a new user a bit of getting used to or learn about.
> 
> For other cases like `dagger config`, just follow what `docker` does. Just `dagger config [flags] [command]`. That is:
> 
> * if it has flags add `[flags]`
> * if it has sub-commands add `[command]`

Ok. Just to confirm, `docker --help` shows:
```shell
Run 'docker COMMAND --help' for more information on a command.
```

We do represent it as such:
`dagger [command] --help`

cmd/dagger/call.go
```diff
@@ -16,7 +16,7 @@ var outputPath string
 var jsonOutput bool
 
 var callCmd = &FuncCommand{
-	Name:  "call [flags] [FUNCTION]...",
+	Name:  "call [flags] [function]...",
```
helderco, line 19: This will produce:
```
dagger call [flags] [function]...
dagger call [flags] [command]
```

Which isn't right because here, "function" and "command" are the same thing. As per the conventions:

> When multiple synopsis lines are given for a utility, it is an indication that the utility has mutually-exclusive arguments.

This happens because you're describing _function_ as a positional argument, when it's actually a subcommand. In the template, cobra sees that it has subcommands so it adds the second line.

Docker, for example, doesn't print two lines. It always prints the same if it has subcommands:

```
{{- if not .HasSubCommands}}  {{.UseLine}}{{end}}
{{- if .HasSubCommands}}  {{ .CommandPath}}{{- if .HasAvailableFlags}} [OPTIONS]{{end}} COMMAND{{end}}
```

In our case, we'd have to check if the command is required or not. That's a plus in docker for consistency that we don't have.

In this case the subcommand is indeed optional, but you mark that by making the usage about the `call` command itself and not include children. If you change the usage to just `call` and remove the square brackets in `[command]` in the template, it produces this which is more correct:

```
dagger call [flags]
dagger call <command> [flags]
```

But Cobra's default is actually `[command]` and in the cases where the command is actually required, then the first line doesn't make sense.

Since we have a mix (sometims required, sometimes optional), we may have to use a helper function to make a single usage line instead, that detects which applies. But I suggest keeping it for another PR and simply keep the usage here as:

```suggestion
	Name:  "call",
```

cmd/dagger/module.go
```diff
@@ -173,7 +173,7 @@ The "--source" flag allows controlling the directory in which the actual module
 }
 
 var moduleInstallCmd = &cobra.Command{
-	Use:     "install [flags] MODULE",
+	Use:     "install [flags] [module]",
```
helderco, line 176: Not optional.

cmd/dagger/run.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 var runCmd = &cobra.Command{
-	Use:     "run [flags] COMMAND",
+	Use:     "run [flags] [command]",
```
helderco, line 23: Maybe add `...` since it can have multiple arguments:
```suggestion
	Use:     "run [flags] [command]...",
```

cmd/dagger/watch.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 var watchCmd = &cobra.Command{
-	Use:    "watch [flags] COMMAND",
+	Use:    "watch [flags] [command]",
```
helderco, line 13: This doesn't actually have subcommands nor accept any arguments:
```suggestion
	Use:    "watch",
```

docs/current_docs/cli/979595-reference.mdx
helderco, line 1: Is this file still being auto-generated? It was removed from mage, wasn't it?

cmd/shim/main.go
```diff
@@ -89,7 +89,7 @@ func main() {
 
 func internalCommand() int {
 	if len(os.Args) < 2 {
-		fmt.Fprintf(os.Stderr, "usage: %s <command> [<args>]\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "usage: %s <command> [args]\n", os.Args[0])
```
helderco, line 92: This is the shim, better leave it as it is. Only `cmd/dagger` in scope here.

helderco (commented):
> ## Update 04/23/2024
> I added a new commit following Cobra's [usage](https://github.com/spf13/cobra/blob/main/command.go#L52-L60)
> 
> > Recommended syntax is as follows:
> > [ ] identifies an optional argument. Arguments that are not enclosed in brackets are required.

I’m not fully convinced on the second statement yet. It’s more common to use `<>` as placeholder, see [(POSIX) Utility Conventions](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html), or at least put it in caps.

In `dagger install module` how do you know if `module` is something you have to type or replace? `dagger install <module>` looks much clearer to me. If the angle brackets are just ugly, that’s why I actually prefer `dagger install MODULE`, but as I’ve pointed out before, that may require more changes to go against Cobra’s default.

How about this, can you tell which is a command and which is an argument (assume only global flags exist here)?
```
dagger config views name set value
```

How about now?
```
dagger config views <name> set <value>
```

In the first one you rely on names that helps your brain fill in the blank, while in the second it’s more immediately obvious.

Lastly, I want to point out that adding `[flags]` may help understand where the split is:

```
dagger run command
# vs
dagger run [flags] command
```

But then it may not apply in some case so I try to picture this without it:

```
dagger run command
# vs
dagger run <command>
```

---

> But without differentiating mandatory and optional arguments, as proposed in your comment below: [#7143 (comment)](https://github.com/dagger/dagger/pull/7143#issuecomment-2067519490).

That’s not what I suggested.

> > For other cases like `dagger config`, just follow what `docker` does. Just `dagger config [flags] [command]`. That is:
> > 
> > * if it has flags add `[flags]`
> > * if it has sub-commands add `[command]`
> 
> Ok 👌 👼 . Just to confirm, `docker --help` shows:
> 
> ```shell
> Run 'docker COMMAND --help' for more information on a command.
> ```
> 
> We keep representing it as such: `dagger [command] --help`

This is correct, but find me a docker command that has subcommands and does something on its own. Most docker commands that have subcommands seem to be used only for grouping those subcommands, so not something you’d use directly. In that case it makes sense to mark the subcommand as required.

But `dagger call` and `dagger config` have valid usages without subcommands, thus the `[]` in `[command]`.

---

> But, for the commands with a mandatory input such as `dagger install [flags] MODULE`, how would you represent it ? ==> Is it `dagger install [flags] module` ? Or do we not make any distinction: `dagger install [flags] [module]`

Yes, we make the distinction.

To be clear, when I gave docker’s example it wasn’t to replicate the syntax or to forcibly adopt the “required” subcommand explicitly, it was to replicate the simplicity of delegating details to subcommands’ `--help`, rather than be more verbose on the parent command’s usage.

You always need to take into account if you’re comparing apples to apples in regards to the syntax of different tools/clis.

helderco:
@grouville, let me know when you're ready for another review 🙏

helderco (approved):
I pushed an update to get this merged quickly :shipit:

cmd/dagger/call.go
helderco: Now that `[flags]` was renamed to `[options]`, we need to add it explicitly. As for the two usage lines in the template, I don't think we can generalize because we have a command that requires a subcommand, and another where it's optional. Better be explicit in `cmd.Use`. I pushed an update with these changes so I can merge this quickly.

docs/current_docs/cli/979595-reference.mdx
helderco: I removed this file. Don't know how it got here. 🙂

PR #7149, opened 2024-04-21
Fix miss makezero bug

alingse:
when I want to add feature for linter  [`makezero`](https://github.com/ashanbrown/makezero) in PR https://github.com/ashanbrown/makezero/pull/15 , I run a github action for real world repos,  and github action report this bug.

I not going to add this `makezero` to your `golangci.yaml` , but I suggest to set to `enable-all: true` and add some `disable` linters.

### DIFF
--- a/telemetry/servers.go
+++ b/telemetry/servers.go
@@ -213,13 +213,13 @@ func logValue(v *otlpcommonv1.AnyValue) log.Value {
 	case *otlpcommonv1.AnyValue_BoolValue:
 		return log.BoolValue(v.GetBoolValue())
 	case *otlpcommonv1.AnyValue_KvlistValue:
-		kvs := make([]log.KeyValue, len(x.KvlistValue.GetValues()))
+		kvs := make([]log.KeyValue, 0, len(x.KvlistValue.GetValues()))
 		for _, kv := range x.KvlistValue.GetValues() {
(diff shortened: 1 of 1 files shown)


jedevc (approved):

jedevc:
@alingse thanks! Could you please sign-off your commit? See [here](https://github.com/dagger/dagger/pull/7149/checks?check_run_id=24072112359) for instructions.

Potentially we could also enable `makezero` in our golangci-lint config? https://golangci-lint.run/usage/linters/#makezero. I don't see an issue with enabling that to make sure we don't regress.

> but I suggest to set to enable-all: true and add some disable linters.

I'm generally not a huge fan of this approach - this means that when upgrading golangci, any new linters would automatically get enabled, and we have to fix the issues. Lots of linters produce a lot of noise (but don't actually detect bad behavior) - so I'd argue it's better to be selective about it.

That said, improving the state of linting in dagger is definitely worth doing, so any improvement PRs would definitely be appreciated!

alingse:
@jedevc done for DCO.

btw, in my work, I will lock the version of golangci-lint to avoid some  unnecessary noise. When it comes to upgrading the tool, we will choose carefully to disable  certain linters temporarily or add some noLintRules. The noise can only disturb the developer once.

but `makezero` has some false-positive case see the PR I mentioned in first description, it's not ready to open it.

PR #7153, opened 2024-04-22
chore: fix ts docs linter

TomChv:

PR #7154, opened 2024-04-22
ci: avoid panic in sdk building

jedevc:
Using functions that can error in `Container.With` is a bad idea. So, what we can do is to put the actual computation into an error group (which has the convenient little side-effect that they all get computed in parallel, hopefully improving build times).

This lets us remove the panic, which was crashing the binary. Because of (I think) telemetry draining, this meant that the client job hung forever on this kind of failure.

cc @helderco I know we discussed previously :tada:


jedevc:
Linting failure is from https://github.com/dagger/dagger/pull/7153.

TomChv (approved):

Issue #6466, opened 2024-01-19
🐞 Platform type is converted to string by the code generator


sagikazarmark:
### What is the issue?

The core Platform type is converted to string when using a module that references that type.

### Dagger version

dagger v0.9.6 (registry.dagger.io/engine) darwin/arm64

helderco:
Can you provide more details?

sagikazarmark:
I use `Platform` in [this](https://github.com/sagikazarmark/daggerverse/blob/main/go/main.go#L162) module.

When using that module as a dependency, the code generator turns `Platform` to `string`:

```go
type GoWithSourceBuildOpts struct {
	// Target platform in "[os]/[platform]/[version]" format (e.g., "darwin/arm64/v7", "windows/amd64", "linux/arm64").
	//
	Platform string
}
```

helderco:
So the issue must extend to all scalars then.

@sipsma, do we need a KindScalar?

PR #6739, opened 2024-02-26
feat: custom flag parser for `Platform` args

jedevc:
This is tricky, since the scalar aliases were getting lost in our `TypeDef` code. We can preserve these, and then provide a custom parser.

With this, we can add validation that the target platform is valid, as well as allow a special `current` platform, which allows this use case:

```go
package main

import (
	"fmt"
)

type Playground struct{}

func (m *Playground) Hello(platform Platform) string {
	return fmt.Sprintf("hello i am platform %q", platform)
}
```

```
❯ dagger call hello --platform=current
hello i am platform "linux/amd64"
```

PR #6967, opened 2024-03-29
Elixir Module support

wingyplus:
Let's try:

```
$ ./hack/dev
$ export PATH=$PWD/hack:$PWD/bin:$PATH
$ with-dev dagger init --sdk=github.com/dagger/dagger/sdk/elixir/runtime potato
Initialized module potato in potato
$ cd potato
$ with-dev dagger functions
Name             Description
container-echo   -
grep-dir         -
$ with-dev dagger call container-echo --string-arg 'Hello, Dagger' stdout
Hello, Dagger
```

🎉 

## Technical design

CAUTION: This is the long road to make it done. The design can be change depends on the discussion and feedback.

Let's take a look at module that SDK generate for you.

```elixir
defmodule Potato do
  @moduledoc """
  A generated module for Potato functions.
  """

  use Dagger.Mod, name: "Potato"

  defstruct [:dag]

  @doc """
  Returns a container that echos whatever string argument is provided.
  """
  @function [
    args: [
      string_arg: [type: :string]
    ],
    return: Dagger.Container
  ]
  def container_echo(self, args) do
    self.dag
    |> Dagger.Client.container()
    |> Dagger.Container.from("alpine:latest")
    |> Dagger.Container.with_exec(~w"echo #{args.string_arg}")
  end

  @function [
    args: [
      directory_arg: [type: Dagger.Directory],
      pattern: [type: :string]
    ],
    return: :string
  ]
  def grep_dir(self, %{directory_arg: directory, pattern: pattern}) do
    self.dag
    |> Dagger.Client.container()
    |> Dagger.Container.from("alpine:latest")
    |> Dagger.Container.with_mounted_directory("/mnt", directory)
    |> Dagger.Container.with_workdir("/mnt")
    |> Dagger.Container.with_exec(["grep", "-R", pattern, "."])
    |> Dagger.Container.stdout()
  end
end
```

Any module that want to be a Dagger module needs to call `use Dagger.Mod` and need to declare the name of the module using `:name` option.

Each function has a module attribute called `function`. The value of that attribute is a keyword (for non-Elixir developer, think like a ordered key/value in another language) that contain the information of the function, currently has only argument and the return type. The module runtime will use this information to generate module declaration using `Module` API and do encode/decode value between Dagger and Elixir world.

The reason I declare this is because in Elixir, we have no other ways to obtain type from function argument because it is a dynamic programming language. But luckily that we can use data-structure to describe the function structure. :)

[...]

Issue #6780, opened 2024-02-29
✨ functions: argument values


rsteube:
### What are you trying to do?

Provide a way to define possible [argument](https://docs.dagger.io/developer-guide/go/264203/functions/#arguments) values of functions.
Expose functions as `json` in `dagger functions`.

```json
[
  {
    "Name": "containerEcho",
    "Description": "example usage: \"dagger call container-echo --string-arg yo stdout\"",
    "ReturnType": {
      "Kind": "OBJECT_KIND",
      "Optional": false,
      "AsObject": {
        "Name": "Container",
        "Functions": null,
        "Fields": null,
        "Constructor": null,
        "SourceModuleName": ""
      },
      "AsInterface": null,
      "AsInput": null,
      "AsList": null
    },
    "Args": [
      {
        "Name": "stringArg",
        "Description": "",
        "TypeDef": {
          "Kind": "STRING_KIND",
          "Optional": false,
          "AsObject": null,
          "AsInterface": null,
          "AsInput": null,
          "AsList": null
        },
        "DefaultValue": ""
      }
    ]
  },
  {
    "Name": "grepDir",
    "Description": "example usage: \"dagger call grep-dir --directory-arg . --pattern GrepDir\"",
    "ReturnType": {
      "Kind": "STRING_KIND",
      "Optional": false,
      "AsObject": null,
      "AsInterface": null,
      "AsInput": null,
      "AsList": null
    },
    "Args": [
      {
        "Name": "directoryArg",
        "Description": "",
        "TypeDef": {
          "Kind": "OBJECT_KIND",
          "Optional": false,
          "AsObject": {
            "Name": "Directory",
            "Functions": null,
            "Fields": null,
            "Constructor": null,
            "SourceModuleName": ""
          },
          "AsInterface": null,
          "AsInput": null,
          "AsList": null
        },
        "DefaultValue": ""
      },
      {
        "Name": "pattern",
        "Description": "",
        "TypeDef": {
          "Kind": "STRING_KIND",
          "Optional": false,
          "AsObject": null,
          "AsInterface": null,
          "AsInput": null,
          "AsList": null
        },
        "DefaultValue": ""
      }
    ]
  }
]
```

### Why is this important to you?

shell completion

helderco:
@rsteube, these are two separate things. Can you split the `json` output into its own issue?

helderco:
High level design for the schema required for this:
- https://github.com/dagger/dagger/pull/7158#issuecomment-2098168444

[...]

PR #7158, opened 2024-04-22
feat: add scalar kind

jedevc:
[DEV-3403](https://linear.app/dagger/issue/DEV-3403)

Fixes https://github.com/dagger/dagger/issues/6466

This is the "proper" implementation of #6739. This allows us to preserve the original scalar name.

Implemented functionality:
- [x] `TypeDef`s have an `AsScalar` field, so that they can be connected to their original value. 
	- For example, this allows the CLI to know that a flag is a `dagger.Platform` flag. Eventually, this could allow us to have a special value, like `"current"` as a shorthand to represent the client platform.
- [x] Modules can call `WithScalar` to define their own non-core scalar values - these are represented as scalars in `TypeDef`s, as well as in the GraphQL schema.
- [x] Other modules can import another module that have a custom scalar, and use it.

TODO:
- [x] Implement all the above functionality
- [x] Add a magical CLI handler for `dagger.Platform` (this was my motivation for doing all this lol)
- [x] Add some tests
- [x] Fixup all TODOs


jedevc:
A fun side-effect I just discovered - because of how dagql does enums, we get enum support included alongside this!

jedevc:
Note: this is only fully enabled for the Go SDK right now - there will need to be minor updates to both Typescript and Python (cc @TomChv @helderco), so that functions that take a scalar argument mark this using `WithScalar` to the engine (instead of using `WithString`).

helderco:
I'll get on that asap and coordinate with @TomChv. Thanks for pushing this through! 🙏

core/integration/module_test.go
```diff
@@ -3395,6 +3395,70 @@ func (t *Toplevel) SayHello(ctx context.Context) (string, error) {
...
+		WithNewFile("main.go", dagger.ContainerWithNewFileOpts{
+			Contents: `package main
+			
+import "fmt"
+
+type Foo struct {}
+
+type CustomString string
+type CustomInt int
+type CustomBool bool
```
helderco, line 3419: You're allowing users to set their own custom scalars, do you see any issues with that @sipsma and @vito?

core/integration/module_test.go
```diff
@@ -3395,6 +3395,70 @@ func (t *Toplevel) SayHello(ctx context.Context) (string, error) {
 	require.JSONEq(t, `{"toplevel":{"sayHello":"hello!"}}`, out)
 }
 
+func TestModuleGoCustomScalar(t *testing.T) {
```
helderco, line 3398: If you can, try to design tests that can be shared between the sdks, like some do. Already setting up the testcase loop and thinking of the test in an agnostic manner really helps. Scalar and enum support are especially well suited for that. If there's some sdk specific idiosyncracies, we can still have a common test and other sdk specific subtests or tests. There's several cases of this.

To be clear, I'm not saying these test steps aren't transferrable, I'm speaking in more general terms and only specifically about preparing the `[]testcase`. It helps to bring in the required SDK consistency on board. 😄

cmd/dagger/flags.go
```diff
@@ -592,6 +597,36 @@ func (v *moduleSourceValue) Get(ctx context.Context, dag *dagger.Client, _ *dagg
 	return modConf.Source, nil
 }
 
+type platformValue struct {
+	platform string
+}
+
+func (v *platformValue) Type() string {
+	return Module
```
helderco, line 605: This should be `Platform`.

core/integration/module_test.go
jedevc: As discussed in discord, I've removed the custom scalars part of this patch - I'm not super fussed on this, so just going to remove it for now. If there's demand, I can always bring it back in https://github.com/dagger/dagger/compare/main...jedevc:dagger:add-custom-scalar-kind.

core/integration/module_test.go
jedevc: This specific test is now removed since I've removed the custom scalars, but I've now added a table test for custom scalar args from core.

jedevc:
> Asking for https://github.com/dagger/dagger/pull/6967, is `WithString` still support?.

@wingyplus, yes, `WithString` is still here - it's just that now, by using `WithScalar` for a type like `Platform`, it indicates it as specifically a `Platform` input, instead of just a normal string - this allows the CLI to have a significantly better interface for it/etc.

helderco:
It's more important to be in the same release than in the same PR. I've already made the change for Python locally but in holiday today and the tests need improvement. I asked Tom to do the same but it may not be as easy there because of how it's written. Hold on at least for Python and I'll check up with Tom.

core/integration/module_test.go
```diff
@@ -2888,6 +2888,56 @@ class Foo {
...
+		{
+			sdk: "python",
+			source: `import dagger
+from dagger import function, object_type
+
+@object_type
+class Test:
+    @function
+    def say_hello(self, platform: dagger.Platform) -> str:
+        return f"hello {platform}"
```
helderco, line 2920: FTR, I've tested the Python change manually and it fixes the issue for both scalars and enums. Should add tests for enum as well.

But... I'm not sure this is the right test here. In Python, type checking doesn't do anything at runtime so `platform` here could be a string and everything runs without error. In fact, if I remove the error in Python when a scalar is used (reverting back to a normal string), this function passes without error.

However the generated client does have runtime type checks and the issue in https://github.com/dagger/dagger/issues/6466 is specifically about how codegen generates this function for the client. Maybe good to have a test create a dependent module to use this function?

Or just check if it creates the expected type in the API schema?

helderco: In Go, this also passes on `main`, so it's only failing on the invalid platform error.

core/integration/module_test.go
```diff
@@ -2888,6 +2888,49 @@ class Foo {
...
+				WithWorkdir("/work").
+				With(daggerExec("init", "--name=foo", "--sdk="+tc.sdk)).
+				With(sdkSource(tc.sdk, tc.source))
+
+			out, err := modGen.With(daggerQuery(`{foo{sayHello(platform: "linux/amd64")}}`)).Stdout(ctx)
+			require.NoError(t, err)
+			require.Equal(t, "hello linux/amd64", gjson.Get(out, "foo.sayHello").String())
+
+			_, err = modGen.With(daggerQuery(`{foo{sayHello(platform: "invalid")}}`)).Stdout(ctx)
+			require.ErrorContains(t, err, "unknown operating system or architecture")
```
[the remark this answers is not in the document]
jedevc: The error comes from the conversion in https://github.com/dagger/dagger/pull/7158/files#diff-1f747bfcbf3f060e72bd8621bc9daa68a4566fedc17f60fc837bf32d068c270bR222-R241.

Essentially, the failure comes from attempting to create a `Platform` scalar from a `string` input, which should fail, using the validation on the `Platform` scalar itself.

core/integration/module_test.go
jedevc: That's intended - the first check is just a sanity check, the second one is the validation thing as described above.

helderco: But what purpose does the sanity check have if it also passes without the changes in this PR?

jedevc: Because the PR has changed logic related to the parsing of scalars, and there were no tests prior to this that explicitly tested passing a valid scalar into a function that requires a scalar.

If this test isn't present, we are only testing that scalar validation can fail, not that scalar validation can succeed.

helderco: Yeah, now that I see where the error comes from it does seem like a good enough test. 👍

core/integration/module_test.go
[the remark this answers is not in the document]
helderco: Ok, so the test is basically validating between the GraphQL query and the API server, before it tries to send it as input to the SDK. But the SDK needs to register the argument correctly in order for the `Platform` scalar parsing to be picked up. 👍

What's strange now is that in Python, this query correctly fails with that error if I use `.Sync(ctx)` instead of `.Stdout(ctx)`, where the latter returns "hello invalid" if I look at the output. I would expect the same error regardless of `Stdout` or `Sync`. 🤔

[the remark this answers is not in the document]
jedevc: Uh oh, that's... not great. Out of curiosity, if you clean your cache, does this keep happening? I remember a specific cloud cache related issue when this could potentially happen, but otherwise not sure.

I'm assuming this is only for python? Or does go do it too?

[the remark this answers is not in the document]
helderco: Only happens in Python it seems 😕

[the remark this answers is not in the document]
helderco: I wiped everything and rebuilt. Now Python passes with `Stdout` too! Going to stress test this a bit more.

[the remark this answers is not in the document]
helderco: Yeah, now it passes everytime no matter what I do. 🤷‍♂️

core/integration/module_call_test.go
```diff
@@ -499,6 +500,83 @@ func (m *Test) Cacher(ctx context.Context, cache *CacheVolume, val string) (stri
...
+		out, err = modGen.With(daggerCall("from-platform", "--platform", "current")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, platforms.DefaultString(), out)
+		_, err = modGen.With(daggerCall("from-platform", "--platform", "invalid")).Stdout(ctx)
+		require.ErrorContains(t, err, "unknown operating system or architecture")
+
+		out, err = modGen.With(daggerCall("to-platform", "--platform", "linux/amd64")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "linux/amd64", out)
+		_, err = modGen.With(daggerCall("from-platform", "--platform", "invalid")).Stdout(ctx)
```
helderco, line 539: Did you mean `to-platform` here?

jedevc: Oops. Yes!

helderco (approved):
LGTM 👍

I mirrored the tests from `dagger call` to test the return values and enum types.

Let's wait on merging for TypeScript a bit more. Tom said he's working on it.

jedevc:
Rebased, since main is a little less flaky right now :tada:

TomChv:
@jedevc @helderco 

I pushed the support for TypeScript on your PR directly: [a0a9d61](https://github.com/dagger/dagger/pull/7158/commits/a0a9d61ea0acff4dad7b6d7146a68dd099b96c67)

jedevc:
@helderco will do - looks like `TestModulePythonScalarKind` and `TestModulePythonEnumKind` are failing in CI though.

helderco:
> @helderco will do - looks like `TestModulePythonScalarKind` and `TestModulePythonEnumKind` are failing in CI though.

Ah, forgot about those. Will fix in a minute 🙏

helderco:
@jedevc, user created enums are actually very useful to support (instead of custom scalars):
- https://github.com/dagger/dagger/issues/6780

jedevc:
Yeah this is actually a bit more of a pain in go since native enum types aren't *really* a thing. I'm happy for someone else to lead on the custom enums work and I can work out how it makes sense in go, but probably makes sense to lead with a language that actually supports them nicely :laughing: 

See https://github.com/dagger/dagger/compare/main...jedevc:dagger:add-custom-scalar-kind for a base for how to handle the custom scalar plumbing (enums are pretty similar to enums for this).

helderco:
I think Go's implementation of enums is fine. Custom scalars are more problematic because their implementation has to define:

- How the scalar's value is represented in the module's language (aka, back-end representation)
- How the value's back-end representation is serialized to a JSON-compatible type
- How the JSON-compatible representation is deserialized to the back-end representation

That can be simplified if we just assume/limit to a normal string everywhere, but it still opens a can of worms when we don't have enough feedback that it's actually needed.

However with enums, it's **just a fixed list of supported string values**. No serialization/deserialization required. To support defining new enums via the API it’s similar to `ObjectTypeDef`, where the fields are just a list of values and their descriptions:

```graphql
type EnumTypeDef {
  name: String!
  description: String!
  values: [EnumValueTypeDef!]!
}

type EnumValueTypeDef {
  name: String!
  description: String!
}

type TypeDef {
  """
  If kind is ENUM_KIND, the enum-specific type definition.
  If kind is not ENUM_KIND, this will be null.
  """
  asEnum: EnumTypeDef
    
  """
  Returns a TypeDef of kind Enum with the provided name.
  
  Note that an enum's values may be omitted if the intent is only to refer to an enum.
  This is how functions are able to return their own, or any other circular reference.
  """
  withEnum(name: String!, description: String = ""): TypeDef!

  """
  Adds a static value for an Enum TypeDef, failing if the type is not an enum.
  """
  withValue(name: String!, description: String = ""): TypeDef!
}

enum TypeDefKind {
  """
  A GraphQL enum type and its values.
  
  Always paired with an EnumTypeDef.
  """
  ENUM_KIND
}
```

PR #7164, opened 2024-04-23
docs: Uses version/arch placeholder in Argo Workflow snippet

vikram-dagger:
This commit removes the hardcoded version in the snippet in favour of a placeholder. There were already other placeholders in the snippet, so this should not cause a major disruption. More information in https://discord.com/channels/707636530424053791/1231554152815919114/1231554152815919114

### DIFF
--- a/docs/current_docs/integrations/snippets/argo-workflow.yaml
+++ b/docs/current_docs/integrations/snippets/argo-workflow.yaml
@@ -26,7 +26,7 @@ spec:
             mode: 0755
             http:
               # replace with the latest available version of Dagger for your platform
-              url: https://github.com/dagger/dagger/releases/download/vX.Y.Z/dagger_vX.Y.Z_PLATFORM_ARCH.tar.gz
+              url: https://github.com/dagger/dagger/releases/download/vX.Y.Z/dagger_vX.Y.Z_OS_ARCH.tar.gz
       container:
         image: alpine:latest
         command: ["sh", "-c"]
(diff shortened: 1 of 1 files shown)


docs/current_docs/integrations/snippets/argo-workflow.yaml
```diff
@@ -18,13 +18,15 @@ spec:
...
             path: /work
             git:
+              # replace with your repository URL
               repo: YOUR_REPOSITORY_URL_HERE
               revision: "main"
           - name: dagger-cli
             path: /usr/local/bin
             mode: 0755
             http:
-              url: https://github.com/dagger/dagger/releases/download/v0.10.2/dagger_v0.10.2_linux_arm64.tar.gz
```
jedevc, line 27: :eyes: I don't think we should point people to github to download the CLI - we should point people to `dl.dagger.io`, right?

marcosnils: +1 we should follow the same pattern as here: https://github.com/dagger/dagger/blob/main/docs/current_docs/integrations/snippets/circle.yml#L16-L17

I believe @kpenfound has an example about this maybe?

vikram-dagger: @kpenfound can you suggest how to change this?

kpenfound: Looks like input artifacts only support git, HTTP, GCS, and S3, but not scripts. So the alternative would be to run a script as another step, which isn't great considering this really should be an input artifact IMO.
I'm good with the current state of this PR

kpenfound (approved):

Issue #7139, opened 2024-04-19
🐞 Python Telemetry: detect when to configure from environment


helderco:
More context in:
- https://github.com/dagger/dagger/discussions/7116
- [🧵 Disabling opentelemetry logs](https://discord.com/channels/707636530424053791/1230840665215336511) (Discord)

I didn't realize `dagger run` doesn't set the `OTEL_` env vars (just `TRACEPARENT`), so the SDK is trying to export unconditionally and failing. Worst part is there's an exponential back-off retry mechanism that's fixed to around 1 minute, so if you don't shut it down it can delay when the script is able to finish. Besides, it pollutes the logs. It's not serious, just annoying.

This affects Python users **not** using modules in v0.11.1 (`dagger run python script.py` or `python script.py`). In the meantime, been suggesting that they turn telemetry off:

```python
import logging
import opentelemetry

# TODO: Remove when https://github.com/dagger/dagger/issues/7139 is released
logging.getLogger("opentelemetry").propagate = False
logging.getLogger("opentelemetry").addHandler(logging.NullHandler())
opentelemetry.trace.get_tracer_provider().shutdown()
```

The logger statement is to avoid showing the _"Already shutdown, dropping span"_ message.

PR #7167, opened 2024-04-23
Refactor Python telemetry initialization

helderco:
Fixes https://github.com/dagger/dagger/issues/7139

This refactor was part of adding _live progress_ support to Python telemetry, but it fixes the above and we need it in next release as I won't have time to finish otherwise. It'll make the _live progress_ PR easier to review anyway.

## Notable changes
- Moved telemetry code to `dagger.telemetry` to establish a convention, [after Go](https://github.com/dagger/dagger/tree/f94d62b7ddd08120370623c926a8935f683bff1f/sdk/go/telemetry).
- `dagger.telemetry.get_tracer` ensures initialization and a consistent instrumentation name as per [`TracerProvider.get_tracer`](https://github.com/open-telemetry/opentelemetry-python/blob/7c67637716261fd2c62c2350b731af8b02e06d5a/opentelemetry-api/src/opentelemetry/trace/__init__.py#L202-L207), for use by the SDK.
- Configuration is now a singleton, that ensures it’s only configured once, so it can be added when it's needed in multiple places.
- `OTEL_` based configuration (providers and exporters) only happens if any of those env vars exist or not explicitly disabled, while propagation always happens. We want to at least pass the trace parent on, even if telemetry is not enabled, and do it without hanging.
- Replaced manual instrumentation for automatic instrumentation, provided by default by the [OpenTelemetry Python SDK](https://opentelemetry-python.readthedocs.io/en/latest/sdk/index.html). Added defaults to environment in order to affect said configuration, without requiring code (manual) as much as possible.
- Because of the previous point, also added support for metrics and logs.
- Removed span over client connection. Since tracing is not enabled in `dagger run`, just using a module’s *registration* and *execution* phases as parent spans, masking (replacing) their more generic `exec /runtime` parent. For scripts, since there's an explicit call to `dagger.connection`, users can put it under their own span if desired.

\cc @TomChv


jedevc (approved):
Approving, since this does fix a fairly critical issue :tada: 

That said, a lot of trust here, I really don't understand the python SDK internals (or OTEL yet, to be fair).

sdk/python/src/dagger/telemetry/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+    def _prepare_env(self):
+        """Prepare environment variables for auto-configuring the SDK."""
+        # When a Resource is created, it defaults to the following env var
+        # for the service name.
+        os.environ.setdefault(OTEL_SERVICE_NAME, SERVICE_NAME)
+
+        # The default is a NoOpProvider.
+        os.environ.setdefault(OTEL_PYTHON_TRACER_PROVIDER, "sdk_tracer_provider")
+
+        # TODO: The INSECURE env vars should be set by the shim next to ENDPOINT
```
vito, line 149: Agreed, ran into this with custom Go integration for tests, don't remember if I made this change yet (on mobile)

sdk/python/src/dagger/telemetry/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+    _DaggerPropagationConfigurator().configure()
+    _DaggerOtelConfigurator().configure()
+
+
+def get_tracer() -> trace.Tracer:
+    """Returns a tracer to use with Dagger."""
+    initialize()
+    return trace.get_tracer(
+        "dagger.io/sdk.python",
+        schema_url=SpanAttributes.SCHEMA_URL,
```
vito, line 58: Any idea what semconv version we're on here? I'm guessing we'll need to keep it in sync across all SDKs, currently it's 1.24.0. Not completely sure. I've seen cryptic warnings when there's a mismatch, so kind of dreading the day it comes up again.

vito (approved):

PR #7170, opened 2024-04-23
cli: rename `flags` to  `options`

grouville:
One of the action items discussed in https://github.com/dagger/dagger/pull/7107#issuecomment-2061525961. It is the first one out of 3, discussed with Helder

Branched from https://github.com/dagger/dagger/pull/7143, which needs to be merged prior this one.

1. Rename all the flag mentions to "options", as proposed by Helder in https://github.com/dagger/dagger/pull/7107#issuecomment-2061525961
2. Remove the automatic [flags] append on all commands: https://github.com/dagger/dagger/pull/7107#discussion_r1569024798

## Update 04/24/2024

- Removed the branch out from #7143, as Helder suggested
- Implemented comments discussed below: 
  - Added `<pattern>` to `add` subcommand
  - Removed `[options]` from commands with just global flags.
  - Fix usage of watch command also (as it touched the "removed `[options]` from commands with just global flags)

### DIFF
--- a/cmd/dagger/config.go
+++ b/cmd/dagger/config.go
@@ -359,7 +359,7 @@ func (c configSubcmd) Command() *cobra.Command {
 	if c.SetCmd != nil {
 		setCmd := &cobra.Command{
 			// TODO: make these specific based on the parent
-			Use:     "set [options] <pattern>",
+			Use:     "set [options] <pattern>...",
 			Short:   "Set the configuration value",
 			Long:    "Set the configuration value",
 			Example: c.SetExample,
@@ -378,7 +378,7 @@ func (c configSubcmd) Command() *cobra.Command {
 	if c.AddCmd != nil {
 		addCmd := &cobra.Command{
 			// TODO: make these specific based on the parent
-			Use:     "add [options]",
+			Use:     "add [options] <pattern>...",
 			Short:   "Add a value to the configuration",
 			Long:    "Add a value to the configuration",
 			Example: c.AddExample,
@@ -397,7 +397,7 @@ func (c configSubcmd) Command() *cobra.Command {
 	if c.RemoveCmd != nil {
 		removeCmd := &cobra.Command{
 			// TODO: make these specific based on the parent
-			Use:     "remove [options]",
+			Use:     "remove [options] <pattern>...",
 			Short:   "Remove a value from the configuration",
 			Long:    "Remove a value from the configuration",
 			Example: c.RemoveExample,
--- a/cmd/dagger/config.go
+++ b/cmd/dagger/config.go
@@ -28,7 +28,7 @@ func init() {
 }
 
 var configCmd = &cobra.Command{
-	Use:   "config",
+	Use:   "config [options]",
 	Short: "Get or set the configuration of a Dagger module",
 	Long:  "Get or set the configuration of a Dagger module. By default, print the configuration of the specified module.",
 	Example: strings.TrimSpace(`
@@ -82,7 +82,7 @@ dagger config -m github.com/dagger/hello-dagger
 }
 
 var configViewsCmd = configSubcmd{
-	Use:   "views [name]",
+	Use:   "views [options] [name]",
 	Short: "Get or set the views of a Dagger module",
 	Long:  "Get or set the views of a Dagger module. By default, print the views of the specified module.",
 	PersistentFlags: func(fs *pflag.FlagSet) {
@@ -359,7 +359,7 @@ func (c configSubcmd) Command() *cobra.Command {
 	if c.SetCmd != nil {
 		setCmd := &cobra.Command{
 			// TODO: make these specific based on the parent
-			Use:     "set",
+			Use:     "set [options] <pattern>",
 			Short:   "Set the configuration value",
 			Long:    "Set the configuration value",
 			Example: c.SetExample,
@@ -378,7 +378,7 @@ func (c configSubcmd) Command() *cobra.Command {
 	if c.AddCmd != nil {
 		addCmd := &cobra.Command{
 			// TODO: make these specific based on the parent
-			Use:     "add",
+			Use:     "add [options]",
 			Short:   "Add a value to the configuration",
 			Long:    "Add a value to the configuration",
 			Example: c.AddExample,
@@ -397,7 +397,7 @@ func (c configSubcmd) Command() *cobra.Command {
 	if c.RemoveCmd != nil {
 		removeCmd := &cobra.Command{
 			// TODO: make these specific based on the parent
-			Use:     "remove",
+			Use:     "remove [options]",
 			Short:   "Remove a value from the configuration",
 			Long:    "Remove a value from the configuration",
 			Example: c.RemoveExample,
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -43,7 +43,7 @@ var funcCmds = FuncCommands{
 }
 
 var funcListCmd = &FuncCommand{
-	Name:  "functions [flags] [FUNCTION]...",
+	Name:  "functions [options] [FUNCTION]...",
 	Short: `List available functions`,
 	Long: strings.ReplaceAll(`List available functions in a module.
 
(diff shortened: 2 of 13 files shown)


helderco:
> Branched from #7143, which needs to be merged prior this one.

I would think you'd flip this, i.e., base the other one off of this one. That's because this one is much simpler to review and approve. The other one has room for interpretation and requires more discussion.

cmd/dagger/config.go
```diff
@@ -359,7 +359,7 @@ func (c configSubcmd) Command() *cobra.Command {
 	if c.SetCmd != nil {
 		setCmd := &cobra.Command{
 			// TODO: make these specific based on the parent
-			Use:     "set",
+			Use:     "set [options]",
```
helderco, line 362:
```suggestion
			Use:     "set [options] <pattern>...",
```

cmd/dagger/main.go
```diff
@@ -172,6 +174,15 @@ func installGlobalFlags(flags *pflag.FlagSet) {
 	}
 }
 
+// disableFlagsInUseLine disables the automatic addition of [flags]
+// when calling UseLine.
+func disableFlagsInUseLine(cmd *cobra.Command) {
+	for _, c := range cmd.Commands() {
+		c.DisableFlagsInUseLine = true
+		disableFlagsInUseLine(c)
```
helderco, line 182: I see you didn't use [my sugestion](https://github.com/dagger/dagger/pull/7107#discussion_r1569024798) of the custom useLine:
```go
c.Use = useLine(c)
```

```go
// custom useLine function that prints options instead of flags.
func useLine(c *cobra.Command) string {
	return strings.Replace(c.UseLine(), "[flags]", "[options]")
}
```
It would replace the need to manually add `[options]` to a lot of commands, although I'll also accept with them manually added due to explicitness.

cmd/dagger/main.go
```diff
@@ -399,14 +410,14 @@ const usageTemplate = `{{ "Usage" | toUpperBold }}
...
-{{ "Flags" | toUpperBold }}
+{{ "Options" | toUpperBold }}
 {{ flagUsagesWrapped .LocalFlags | trimTrailingWhitespaces}}
 
 {{- end}}
 
 {{- if .HasAvailableInheritedFlags}}
 
-{{ "Global Flags" | toUpperBold }}
+{{ "Global Options" | toUpperBold }}
```
helderco, line 420: This is more accurate and is what `gh` uses:
```suggestion
{{ "Inherited Options" | toUpperBold }}
```

cmd/dagger/functions.go
```diff
@@ -441,10 +441,11 @@ func (fc *FuncCommand) addSubCommands(cmd *cobra.Command, dag *dagger.Client, fn
 
 func (fc *FuncCommand) makeSubCmd(dag *dagger.Client, fn *modFunction) *cobra.Command {
 	newCmd := &cobra.Command{
-		Use:     cliName(fn.Name),
-		Short:   strings.SplitN(fn.Description, "\n", 2)[0],
-		Long:    fn.Description,
-		GroupID: funcGroup.ID,
+		Use:                   cliName(fn.Name),
```
helderco, line 444: Don't you need to add `[options]` here?

cmd/dagger/main.go
grouville: I tried, but there were two reasons for that:
1. the useLine as suggested broke the parsing of the commands, as the c.UseLine string starts with the program name and the command, whereas the 'Use' field of the command is used for the command parsing.

With the useLine(c) function:
```shell
./dagger listen
Error: unknown command "listen" for "dagger"
Run 'dagger --help' for usage.
```

We could strip the "dagger " part of the `UseLine()` string, but is it really worth it as the explicitness could be more readable ?

Working implementation:
```go
// useLine replaces "[flags]" with "[options]" on the usage
// it keeps the same c.Use format (without CLI name), as it is being used for parsing
func useLine(c *cobra.Command) string {
	usageLine := c.UseLine()
	modifiedUsageLine := strings.Replace(usageLine, "[flags]", "[options]", -1)
	
	// Remove the "dagger " prefix by slicing the string from the 7th index onwards.
	// breaks command parsing otherwise
	if strings.HasPrefix(modifiedUsageLine, "dagger ") {
		return modifiedUsageLine[7:]
	}
	return modifiedUsageLine
}
```

2.  Explicitly specifying the `[options]` ensures the position of the appended `[flags]`, where for example, in our current CLI, the position of `[flags]` is not coherent across commands:
```shell
dagger install --help
Add a Dagger module as a dependency of a local module.

Usage:
  dagger install [flags] MODULE
```

```shell
dagger gen --help
Generate CLI reference documentation in the given file path.

Usage:
  dagger gen FILE [flags]
```
But, this could also potentially conflict with https://github.com/dagger/dagger/pull/7143#discussion_r1577045429, so no strong opinion on that

cmd/dagger/functions.go
grouville: You might be right, but given that we currently cannot interpret the `--help` after a module's function in the `functions` command, I did not see the need to implement it right away (I might be wrong and might be testing improperly).

```shell
./dagger -m github.com/levlaz/daggerverse/mariadb@v0.2.1 functions base --help 
✔ initialize 1.6s

```

and if putting the help before, I have the same usage as with `dagger functions --help`:
```shell
 ./dagger -m github.com/levlaz/daggerverse/mariadb@v0.2.1 functions --help base

available functions.

USAGE
  dagger functions [options] [function]...

OPTIONS
  -m, --mod string   Path to dagger.json config file for the module or a directory
                     containing that file. Either local path (e.g. "/path/to/some/dir") or a
                     github repo (e.g. "github.com/dagger/dagger/path/to/some/subdir")

INHERITED OPTIONS
```

cmd/dagger/cloud.go
```diff
@@ -23,15 +23,15 @@ func init() {
 	rootCmd.AddGroup(group)
 
 	loginCmd := &cobra.Command{
-		Use:     "login",
+		Use:     "login [options]",
```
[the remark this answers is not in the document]
grouville: I also removed the COMMAND from `dagger watch`, and removed the added `[options]`, to be consistent with this syntax: https://github.com/dagger/dagger/issues/6943#issuecomment-2067514623

cmd/dagger/main.go
helderco: This isn't making any sense. Which implementation of UseLine are you referring to? I put it on the first comment here.

helderco: I linked to the original suggestion, have you taken it into consideration? What you're describing doesn't reflect what I said.

cmd/dagger/cloud.go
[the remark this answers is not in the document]
helderco: I find it easier to confirm resolutions on subsequent reviews if comments are marked resolved by the reviewer. It's like crossing off a check list 🙂

cmd/dagger/main.go
helderco: Oh wait, yeah I see it now. I'll get back to you on that (I'm on phone), in the meantime disregard my feedback here.

cmd/dagger/cloud.go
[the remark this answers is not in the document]
grouville: > I find it easier to confirm resolutions on subsequent reviews if comments are marked resolved by the reviewer. It's like crossing off a check list 🙂

unresolved, sorry 🙏

cmd/dagger/functions.go
helderco: But this code doesn't apply to `functions` right? It's for `call` subcommands. In `functions` it's positional arguments.

cmd/dagger/config.go
helderco: The `...` is missing, and on `add` and `remove` too. Can also revert here and add on the other PR since you're making a general correction on the usages there.

helderco (approved):
Approving the rename, let's continue on usage fixes in #7143

cmd/dagger/functions.go
grouville: > But this code doesn't apply to `functions` right? It's for `call` subcommands. In `functions` it's positional arguments.

You are right, my bad. It is indeed required for `dagger call` with two functions:
```shell
./dagger --progress=plain -m github.com/levlaz/daggerverse/mariadb@v0.2.1 call base workdir --help
Retrieves the working directory for all commands.

USAGE
  dagger call base workdir [options]
```

cmd/dagger/config.go
grouville: My bad, reverting the `<pattern>...` right now to fix on the general correction PR

PR #7173, opened 2024-04-24
feat: expand env variable by default

### LINKED ISSUE #7171: Allow env variable interpolation in Container.withEnvVariable
## Problem

Dagger has no direct equivalent to this simple Dockerfile line: `ENV PATH=foo:${PATH}`.

## Solution

Add an optional argument `Container.withEnvVariable` to allow interpolating the current container's env variables into string arguments. And possibly, add it to other `Container` functions as well.

The default for this new argument should be `true`, I think. In other words, interpolation should happen by default.

Then this would work out of the box:

```golang
func addToPath(c *Container) *Container {
 return c.WithEnvVariable("PATH", "$HOME/bin:$PATH")
```

wingyplus:
Fixes https://github.com/dagger/dagger/issues/7171

The `withEnvVariable` API will expand shell env variable by default. And
add `noExpand` option to do not expand when it's set to `true`. The
`expand` option is now deprecated but still compatible with the previous
version.

### DIFF
--- a/core/integration/container_test.go
+++ b/core/integration/container_test.go
@@ -1090,7 +1090,16 @@ func TestContainerWithEnvVariableExpand(t *testing.T) {
 	t.Run("add env var without expansion", func(t *testing.T) {
 		out, err := c.Container().
 			From(alpineImage).
-			WithEnvVariable("FOO", "foo:$PATH").
+			WithEnvVariable("FOO", "foo:$PATH", dagger.ContainerWithEnvVariableOpts{Expand: false, NoExpand: true}).
+			WithExec([]string{"printenv", "FOO"}).
+			Stdout(ctx)
+
+		require.NoError(t, err)
+		require.Equal(t, "foo:$PATH\n", out)
+
+		out, err = c.Container().
+			From(alpineImage).
+			WithEnvVariable("FOO", "foo:$PATH", dagger.ContainerWithEnvVariableOpts{NoExpand: true}).
 			WithExec([]string{"printenv", "FOO"}).
 			Stdout(ctx)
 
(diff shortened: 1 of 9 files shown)


wingyplus:
Just open a PR to see what you thought. If you're all agree, I can put this work forward. :)

wingyplus:
I have hit a issue on Go SDK now. When I change `Expand` default to `true`. The no expand variable test get failure:


```
=== NAME  TestContainerWithEnvVariableExpand/add_env_var_without_expansion
    container_test.go:1080: 
                Error Trace:    /home/wingyplus/src/github.com/dagger/dagger/core/integration/contain
                Error:          Not equal: 
                                expected: "foo:$PATH\n"
                                actual  : "foo:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbi
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,2 +1,2 @@
                                -foo:$PATH
                                +foo:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
                                 
                Test:           TestContainerWithEnvVariableExpand/add_env_var_without_expansion
```

After digging into an issue. The `WithEnvVarible` Go SDK api detect optional argument by `IsZero`, cause boolean value not set to query when `Expand` is `false`. 

If I understand it correctly, we need to modify the codegen to always pass the boolean value. Or we have another way to fix this?

helderco:
> If I understand it correctly, we need to modify the codegen to always pass the boolean value. Or we have another way to fix this?

Yeah, I guess we've always avoided this. Even in https://github.com/dagger/dagger/pull/7136, the way to flip the default on `skipEntrypoint` is to deprecate it in favor of a new `useEntrypoint`.

To do the same here we'd need a new option like `noExpand` that defaults to `false`. It also helps with the deprecation because you can detect if someone is using the older `expand` or not.

wingyplus:
@helderco I just add `noExpand` option as you suggested. I'm not sure how to do with `expand` so I make it compatible for now. 🙏

helderco:
Waiting on status of:
- https://github.com/dagger/dagger/issues/7171

helderco:
Thanks @wingyplus, but closing for now since parent issue was closed. If people are interested, comment on the issue first so we reopen.

PR #7174, opened 2024-04-24
chore: enable unparam linter

jedevc:
Partially inspired by https://github.com/moby/buildkit/pull/4828 upstream.

Unused parameters can sometimes indicate weird issues, at the very least that are a bit of a code smell (additionally, gopls in vscode loves to flag these, which is the main reason I wanted to clear them up lol).

Thankfully, none of these are significant, there are no hiding bugs, it's just a little bit of a tidy up.

`unparam` definitely seems to be optimized for less "noise", and doesn't pick up on *every* unused parameter - it seems to try and avoid finding cases where this is used in an interface/higher-order function.


dagql/objects.go
jedevc, line 1: While I was in the area, I also fixed up some of the incorrect doc comments here :tada:

sipsma (approved):

PR #7183, opened 2024-04-25
cli: fix `--help` for `dagger functions`

grouville:
`dagger functions func1 --help` is currently breaking, as `--help` is interpreted / manually parsed as a function and not as a flag
 
 This was discussed here: https://github.com/dagger/dagger/issues/6943#issuecomment-2067514623. 
Branched from https://github.com/dagger/dagger/pull/7170. To be rebased once merged

#### Before
```shell
dagger -m github.com/levlaz/daggerverse/mariadb@v0.2.1 functions base --help
✔ initialize 2.1s

```

#### After
```shell
./dagger -m github.com/levlaz/daggerverse/mariadb@v0.2.1 functions base --help

available functions.

USAGE
  dagger functions [options] [FUNCTION]...

OPTIONS
  -m, --mod string   Path to dagger.json config file for the module or a directory
                     containing that file. Either local path (e.g.
                     "/path/to/some/dir") or a github repo (e.g.
                     "github.com/dagger/dagger/path/to/some/subdir")

INHERITED OPTIONS
```

### DIFF
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -54,6 +54,13 @@ available functions.
 		"`",
 	),
 	Execute: func(fc *FuncCommand, cmd *cobra.Command) error {
+		// Manually check for --help in the arguments
+		for _, arg := range cmd.Flags().Args() {
+			if arg == "--help" || arg == "-h" {
+				return cmd.Help()
+			}
+		}
+
 		tw := tabwriter.NewWriter(cmd.OutOrStdout(), 0, 0, 3, ' ', tabwriter.DiscardEmptyColumns)
 		var o functionProvider = fc.mod.GetMainObject()
 		fmt.Fprintf(tw, "%s\t%s\n",
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -43,7 +43,7 @@ var funcCmds = FuncCommands{
 }
 
 var funcListCmd = &FuncCommand{
-	Name:  "functions [flags] [FUNCTION]...",
+	Name:  "functions [options] [FUNCTION]...",
 	Short: `List available functions`,
 	Long: strings.ReplaceAll(`List available functions in a module.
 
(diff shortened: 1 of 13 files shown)


helderco:
> Branched from #7170. To be rebased once merged

Since this change has no dependency or overlap with #7170, it would have been better to create the PR against main instead. 🙂

cmd/dagger/functions.go
```diff
@@ -54,6 +54,13 @@ available functions.
 		"`",
 	),
 	Execute: func(fc *FuncCommand, cmd *cobra.Command) error {
+		// Manually check for --help in the arguments
+		for _, arg := range cmd.Flags().Args() {
+			if arg == "--help" || arg == "-h" {
+				return cmd.Help()
+			}
+		}
```
helderco, line 62: If you look at what `cmd.Flags().Args()` does, it should return the command's arguments without flags. The reason it's appearing here is that `FuncCommand` disables interspersed flag parsing. The better solution is to handle this in `FuncCommand` so that flags are parsed before getting to this point. 

## Technical explanation

<details><summary>Why a <code>FuncCommand</code> abstration?</summary>

Here’s a bit of context. `FuncCommand` was created to support several commands that work like `dagger call`:
- `dagger download`
- `dagger up`
- `dagger shell`
- `dagger print` (eventually)

`dagger functions` was special in that it requires most of the same boilerplate, but without the dynamic generation of Cobra commands based on function chaining. So today we’re in a weird place where we have a `FuncCommand` abstraction for just 1 case and 50% of another case (without the subcommands).

We could simplify to just have a common bootstrapping function but the time to refactor this isn’t worth it when there’s other priorities. Besides, it actually keeps all the chaining logic away from the specific `call` implementation so it’s easier to see what `call` does specifically. Same for `functions`. 
</details>

In `dagger call`, we need to disable interspersed flag parsing because it would try to parse a flag to a command that isn’t created yet. These commands are created dynamically and recursively as long as the next positional argument matches an available function from the object that the current function returns.

During this process, each subcommand will have its own set of flags, which is going to be parsed during this creation process. That’s what *traverse children* does.

For `functions` though, it doesn’t go forward with creating subcommands, so by the time `Execute`  runs, any flags after a positional argument haven’t been parsed. But that [can be fixed](https://github.com/dagger/dagger/pull/7187).

helderco:
I submitted a different implementation here:
- https://github.com/dagger/dagger/pull/7187

Thanks for taking the initiative to fix it! 🙏

cmd/dagger/functions.go
grouville: Thanks for the context. I was confused at first with the FuncCommand being halfway executed when `Execute` is present, but the context indeed makes sense 🙏 

> This is best handled by native flag parsing which knows which flags are defined, rather than hardcode these values. But it should be done in FuncCommand for better separation of concerns.
Yes, agreed. I tried but failed. Your implementation is insightful, I would have never found out 🙏

PR #7186, opened 2024-04-25
feat: propagate DAGGER_ENGINE_* environ variables on engine to pipelines

haoqixu:
This PR updates the shim to propagate `DAGGER_ENGINE_*` environ on the engine to pipeline containers. Part of #6599


sipsma:
@haoqixu thanks for the PR! We need to go with a slightly different approach (we don't want all of these env vars to be set in every container, need to be more selective) and add integ test coverage for all this. I missed this and started working on all of it already here: https://github.com/dagger/dagger/pull/7255

So while I appreciate this I'm gonna close this for now; the support is coming very soon though!

PR #7187, opened 2024-04-25
cli: Fix flag parsing when interspersed with arguments in `dagger functions`

helderco:
Alternative solution to #7183

Discussed in https://github.com/dagger/dagger/issues/6943#issuecomment-2067514623.

## Example
```console
dagger -m github.com/levlaz/daggerverse/mariadb@v0.2.1 functions base --help
```

**Before**
```
```

**After**
```

available functions.

USAGE
  dagger functions [flags] [FUNCTION]...

FLAGS
  -m, --mod string   Path to dagger.json config file for the module or a directory containing that file.
                     Either local path (e.g. "/path/to/some/dir") or a github repo (e.g.
                     "github.com/dagger/dagger/path/to/some/subdir")

GLOBAL FLAGS
```

### DIFF
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -329,12 +329,7 @@ func (fc *FuncCommand) execute(c *cobra.Command, a []string) (rerr error) {
 		return cmd.Help()
 	}
 
-	err = cmd.RunE(cmd, flags)
-	if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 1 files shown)


jedevc (approved):

grouville (approved):
Works like a charm. Thanks for the explanation 🙏

PR #7193, opened 2024-04-25
move buildkit logs to extra-debug level

vito:
follow-up to #7106

### DIFF
--- a/cmd/engine/main.go
+++ b/cmd/engine/main.go
@@ -783,7 +783,7 @@ func newController(ctx context.Context, c *cli.Context, cfg *config.Config, pubs
 	}
 
 	bkLogsW := io.Discard
-	if cfg.Debug {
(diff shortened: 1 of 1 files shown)


cmd/engine/main.go
```diff
@@ -783,7 +783,7 @@ func newController(ctx context.Context, c *cli.Context, cfg *config.Config, pubs
 	}
 
 	bkLogsW := io.Discard
-	if cfg.Debug {
+	if c.IsSet("extra-debug") {
```
jedevc, line 786: We could also do `if slog.Default().Enabled(ctx, slog.LevelExtraDebug)` - this means it gets automatically enabled if `trace` is turned on.

vito: nice, will do that instead

jedevc (approved):

PR #7194, opened 2024-04-25
fix: avoid overwriting go mod/sum files on module generation

### LINKED ISSUE #7145: 🐞 Go SDK: go.mod in root directory is overwritten by dagger init when there’s also a package
### What is the issue?

Investigation after a report by @sagikazarmark in:

* [https://github.com/dagger/dagger/issues/6963#issuecomment-2066935830](https://github.com/dagger/dagger/issues/6963#issuecomment-2066935830)

It appears that `dagger init --sdk go` overwrites an existing `go.mod` when run in a *context directory* with a go package. The *context directory* is usually the directory that has `.git`, but if no parent directory has one, it defaults to the directory where `dagger.json` is.

Also tested on v0.10.3 and got the same result.

### Dagger version

dagger v0.11.1 ([registry.dagger.io/engine](http://registry.dagger.io/engine)) darwin/arm64

### Steps to reproduce

From a directory that's not inside a git repo:

```console
mkdir testgo
cd testgo
go mod init example.com
cat > main.go <
package main
 
import "fmt"
 
func main() {
    fmt.Println("Hello, World!")
}
EOF
dagger init --sdk go
```

You'll see that there's no `go.mod` in `dagger/` and the root `go.mod` has the SDK's dependencies now.

### Log output

```
find .
.
./go.mod
./LICENSE
./dagger
./dagger/dagger.gen.go
./dagger/internal
./dagger/internal/querybuilder
./dagger/internal/querybuilder/marshal.go
./dagger/internal/querybuilder/querybuilder.go
./dagger/internal/dagger
./dagger/internal/dagger/dagger.gen.go
./dagger/internal/telemetry
./dagger/internal/telemetry/batch_processor.go
./dagger/internal/telemetry/processor.go
./dagger/internal/telemetry/proxy.go
./dagger/internal/telemetry/span.go
./dagger/internal/telemetry/init.go
./dagger/internal/telemetry/attrs.go
./dagger/.gitignore
./dagger/.gitattributes
./dagger/main.go
./go.sum
./dagger.json
./main.go
```

```
cat go.mod
module example.com
 
go 1.21.6
 
require (
 github.com/99designs/gqlgen v0.17.45
 github.com/Khan/genqlient v0.7.0
 github.com/vektah/gqlparser/v2 v2.5.11
 go.opentelemetry.io/otel v1.25.0
 go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.25.0
 go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.25.0
 go.opentelemetry.io/otel/sdk v1.25.0
 go.opentelemetry.io/otel/trace v1.25.0
 golang.org/x/exp v0.0.0-20240416160154-fe59bbe5cc7f
 golang.org/x/sync v0.7.0
 google.golang.org/grpc v1.63.2
)
 
require (
 github.com/cenkalti/backoff/v4 v4.3.0 // indirect
 github.com/go-logr/logr v1.4.1 // indirect
 github.com/go-logr/stdr v1.2.2 // indirect
 github.com/google/uuid v1.6.0 // indirect
 github.com/grpc-ecosystem/grpc-gateway/v2 v2.19.0 // indirect
 github.com/sosodev/duration v1.2.0 // indirect
 go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.25.0 // indirect
 go.opentelemetry.io/otel/metric v1.25.0 // indirect
 go.opentelemetry.io/proto/otlp v1.1.0 // indirect
 golang.org/x/net v0.23.0 // indirect
 golang.org/x/sys v0.18.0 // indirect
 golang.org/x/text v0.14.0 // indirect
 google.golang.org/genproto/googleapis/api v0.0.0-20240227224415-6ceb2ff114de // indirect
 google.golang.org/genproto/googleapis/rpc v0.0.0-20240401170217-c3f982113cda // indirect
 google.golang.org/protobuf v1.33.0 // indirect
)
```

jedevc:
Fixes #7145 - essentially, we need to a more careful merge of `go.mod` and `go.sum` when adding dependencies, instead of replacing them.

Essentially, when doing a `dagger init --sdk=go` in an existing go repo, we were following a code path that meant we weren't using the builtin `go.mod`/`go.sum`s at all, and were instead relying on `go mod tidy` to do something reasonable (which is also slower!).

Additionally, when using `dagger develop`, we *never* preserved `go.sum`, always completely overwriting it (instead of merging it with the previous contents). This only worked because `go mod tidy` does something reasonable (again, which is slower).

Doing this was a bit tricky - the previous logic here was quite difficult to follow along with, so I've also updated it to be a bit clearer.


cmd/codegen/generator/generator.go
```diff
@@ -24,17 +24,18 @@ const (
...
+	// OutputDir is the path to place generated code.
 	OutputDir string
 
-	// Name of the module to generate code for
-	ModuleName        string
-	ModuleContextPath string
+	// ModuleName is the module name to generate code for.
+	ModuleName string
+	// ModulePath is the subpath where a module can be found.
+	ModulePath string
```
sipsma, line 36: I'd prefer to keep the name more specific and include `Context` since that's what it is. Otherwise there's 3 dirs this could be referring to (context is the whole git repo the module lives in, source is where the dagger.json lives, source subpath is where the actual source code for the module is)

cmd/codegen/main.go
```diff
@@ -66,10 +68,17 @@ func ClientGen(cmd *cobra.Command, args []string) error {
...
-			return fmt.Errorf("--module-name requires --module-context")
+		if modulePath == "" {
+			return fmt.Errorf("--module-name requires --module-path")
 		}
-		cfg.ModuleContextPath = moduleContextPath
+		modulePath, err = relativeTo(outputDir, modulePath)
+		if err != nil {
+			return err
+		}
+		if strings.HasPrefix(modulePath, "..") || strings.HasPrefix(modulePath, "../") {
```
sipsma, line 78: nit, just informational: `filepath.IsLocal` is convenient in these situations

sipsma (approved):
One naming comment, but otherwise LGTM, thanks for the extra cleanup too!

cmd/codegen/generator/generator.go
jedevc: Fixed, and restored the name to `ModuleContextPath`.

cmd/codegen/main.go
jedevc: Ah, we can't use that here since we only want to check if it's a parent - however, I did simplify this to using `strings.Cut` to avoid the repetition here.

PR #7196, opened 2024-04-25
DRAFT docs: Revises quickstart with CLI-first approach

vikram-dagger:
This commit revises the quickstart to use an example application and build a pipeline for it primarily by using the CLI, with additional examples of using custom Functions.

Also see #7275


vikram-dagger:
This is a scratch PR to come up with a new approach and structure for the quickstart based on feedback to date. 

The goals are:
- fewer or no "toy" or "hello world" examples - feedback is that users struggle to connect them with "real" CI pipeline operations
- explain reasons and benefits to create a custom module vs using one off Daggerverse - feedback is that we explain the "how" but not "why" or "when"
- quickstart steps should connect to each other - feedback is that while the examples on an individual page connect up, going from one page to another does not connect eg. there's no connection seen between the step of building a binary with golang and the step of building a container with wolfi (this connection appears later, near the end of the quickstart)
- less theory/concepts - this feedback is mainly related to the initial "hello world" examples and arguments, some users feel like these sections spend too much time on concepts
- actionable knowledge - feedback is that users are expecting the quickstart to teach them how to build a CI pipeline which they can apply to their own project, and it's not doing that

The outlined structure is fairly wide ranging at the moment and is being actively iterated on, would love to get specific feedback from others on:
- the approach (now starts with an example repo, explains concepts in the context of building a pipeline for that repo)
- the content (what can we cut/add? I added a lot to begin with just for feedback purposes, but the eventual goal is to make this a 10 min max experience)

vikram-dagger:
Closing in favour of #7275 per decision to try a code-first approach to the quickstart.

PR #7082, opened 2024-04-11
Migrate to new ci runners

jedevc:
This switches all our GitHub workflows to use the second generation of Dagger Runners. They are based on [Run Dagger on Amazon EKS with GitHub Actions Runner and Karpenter](https://archive.docs.dagger.io/0.9/934191/eks-github-karpenter), currently running:
- AWS EKS `v1.28.7`
- Karpenter `v0.32.0`

As of today, we expose the following runner options:
- `dagger-v0-10-1`
- `dagger-v0-10-2`
- `dagger-v0-10-3`
- `dagger-v0-10-3-4c`
- `dagger-v0-10-3-8c`
- `dagger-v0-10-3-16c`
- `dagger-v0-11-0`
- `dagger-v0-11-0-4c`
- `dagger-v0-11-0-8c`
- `dagger-v0-11-0-16c`
- `dagger-v0-11-1`
- `dagger-v0-11-1-4c`
- `dagger-v0-11-1-4c-nvme`
- `dagger-v0-11-1-8c`
- `dagger-v0-11-1-8c-nvme`
- `dagger-v0-11-1-16c`
- `dagger-v0-11-1-16c-nvme`
- `dagger-v0-11-2`
- `dagger-v0-11-2-4c`
- `dagger-v0-11-2-4c-nvme`
- `dagger-v0-11-2-8c`
- `dagger-v0-11-2-8c-nvme`
- `dagger-v0-11-2-16c`
- `dagger-v0-11-2-16c-nvme`

As soon as a new version of Dagger gets released - e.g. `v0.11.3` - a bunch of variants with that version will appear automatically.

This change is great because:
1. Dagger Engine version can be bumped by changing the GitHub Workflow definition - no more coordination required!
2. We can test different configurations (and versions!) in PRs and see which one performs best
3. We are not forced to bump the Engine version for all workflows at the same time - it was an all-or-nothing proposition before

To learn more, see the following (private) [infra repo](https://github.com/dagger/dagger.io/tree/47aba9e81162bf4790baa648c9ef258e0668272d/infra/ci/eks-2023-12-15). @matipan @gerhard that have all the context.

---

Something worth mentioning separately is that this also introduces asynchronous triggers.

We are running a specific job - `engine:testrace` - in a production bare-metal K8s cluster. This is done **for comparison reasons only** since we know that it's going to fail (we only have 12CPU threads available). See https://github.com/dagger/dagger/pull/7223#issuecomment-2091059448 for more details.

[...]

PR #7197, opened 2024-04-25
CI: Try running Engine & CLI workflow in larger GitHub runners

gerhard:
Is this an intermediary step that we can take towards migrating to new CI runners? Related to:
- https://github.com/dagger/dagger/pull/7082

If the Engine & CLI workflows passes, I would be tempted to merge this before the other PR. The focus is to migrate off the legacy CI as soon as possible.

Trade-offs:
- If we keep the current `DAGGER_CLOUD_TOKEN`, we will not be able to use the Cloud Cache in PRs coming from forks
- If we add the `DAGGER_CLOUD_TOKEN` secret, this will not be available to PRs coming from forks, meaning that we will not have Dagger Cloud Traces in those PRs

### DIFF
--- a/.github/workflows/_hack_make.yml
+++ b/.github/workflows/_hack_make.yml
@@ -21,7 +21,7 @@ on:
         default: 60  
       size:
         type: string
-        default: "dagger-runner-2c-8g"
+        default: "ubuntu-latest"
         required: false
       dagger-version:
         type: string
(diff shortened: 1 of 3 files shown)


gerhard:
Closing, we are now fully committed to pushing this one through:
- https://github.com/dagger/dagger/pull/7082

PR #7201, opened 2024-04-26
add missing backtick

sagikazarmark:


jedevc (approved):

jedevc:
Nice catch, just realized this was also in the python equivalent of this page, so fixed that too!

PR #7206, opened 2024-04-26
ci: pin macos version to 12

jedevc:
An automatic upgrade here seems to have caused an issue where colima was not available.


gerhard (approved):
Good to merge as soon as the last check passes 👍

Issue #6749, opened 2024-02-27
Zenith: Confusion over what "optional" means


helderco:
## Summary

While documenting optional and default values in Python functions I came upon some inconsistencies in implementations due to different assumptions. The term “Optional” that’s used in modules is misleading.

## Terminology

In Python, an argument that is optional is one with a default value because **the true meaning of “optional” is that it can be omitted**.

```python
def foo(bar: str = ""): ...
```

In GraphQL it’s mostly the same, with the exception that a **nullable field without a default can also be omitted**, in which case it automatically defaults to `null`:

```graphql
foo: String  # null if omitted
```

The same doesn’t happen in Python:

```python
def foo(bar: str | None): ...  # this is a required argument!
```

But in the Dagger API, we use “optional” to mean nullable. I think people usually find the nullable/non-null terminology a bit strange and tend to prefer thinking about the “!” at the end of a type as meaning “required” (me included):

```graphql
foo: String!  # does not mean required! it means null is not allowed
```

So some may find it surprising that you can actually have a type that is non-null and has a default:

```graphql
foo: String! = "bar"  # this is actually optional due to the default!
```

## Core schema generation

Our core API has only nullable fields when a default is used:

```graphql
foo: String = "bar"
```

There’s a use case for that, especially in terms of being able to express the difference between “omitted” and “unset”:
- **Not provided**: Use the default “bar” string.
- **Provided as `null`**: Explicitly remove or clear foo.
- **Provided as a non-null string**: Update foo to the provided value.

But I think, at least in most cases, they should be *non-null* instead. 

In Go, where the schema is expressed in, to be nullable, the type should be a pointer, that allows for *nil*. Descriptions are optional, but being a non-null empty string is enough. There is no advantage in also allowing a `null` to clear this value.

For example, this is how core generates the following argument:

```go
Description string `default:""`  // description: String = ""
```

But I think it should be this:

```go
Description string `default:""`  // description: String! = ""
Description *string `default:""`  // description: String = ""
```

\cc @vito 

## SDK inconsistency

Let’s see how the SDKs can express the following GraphQL schema:

```graphql
foo(
    a: String!,
    b: String,
    c: String! = "foo",
    d: String = null,
    e: String = "bar"
): String!
```

### Go

Go modules can currently express all of them, it’s just the terminology of having “optional” independent from “default” that can be confusing:

```go
type Foo struct{ }

func (m *Foo) Bar(
	a string,
	// +optional
	b *string,
	// +default="foo"
[...]

PR #7207, opened 2024-04-26
fix: remove automatic go optional primitives

jedevc:
Previously, we would mark a parameter as optional in two cases:
- With a `// +optional` pragma comment,
- Any pointer to a primitive (string, integer, boolean)

This second one is a leftover from when optionals in the Go SDK was purely indicated using pointer-iness. When working on this before, I think I must have missed this one, or I would have tried to remove it then :D


vito (approved):

helderco:
There are tests depending on this:

https://github.com/dagger/dagger/blob/379e812dee7c36713c30040324b99559bd74cf36/core/integration/module_test.go#L1749-L1760

Maybe bundle with https://github.com/dagger/dagger/issues/6749#issuecomment-2078334724 which replace that behavior?

Issue #6951, opened 2024-03-26
Using Services in multiple containers results in network issues


kpenfound:
This appears to be a regression introduced in 0.10.2, things work as expected in 0.10.1

From @nipuna-perera on Discord:

Basically run  (container A + service). 

file = get file from (container A + service)

Try to run (container B + file) fails with:

```
host alias: lookup t6sjv1ru3serq on 10.87.0.1:53: no such host
  ┃ lookup t6sjv1ru3serq.c1mh4mk35essm.dagger.local on 10.87.0.1:53: no such host
```

kpenfound:
Repro:
```go
func (r *Repro) Run(ctx context.Context) *Container {
	redis := dag.Container().
		From("redis").
		WithExposedPort(6379).
		AsService()
	cli := dag.Container().
		From("redis").
		WithoutEntrypoint().
		WithServiceBinding("redis", redis)

	ctrA := cli.WithExec([]string{"sh", "-c", "redis-cli -h redis info >> /tmp/out.txt"})

	file := ctrA.Directory("/tmp").File("/out.txt")

	ctrB := dag.Container().
		From("alpine").
		WithFile("/out.txt", file)

	return ctrB.WithExec([]string{"cat", "/out.txt"})
}
```

`dagger call run`

on v0.10.1:
```
Container evaluated. Use "dagger call run --help" to see available sub-commands.
```

on v0.10.2:
```
✔ Repro.run: Container! 5.1s
  ✔ Container.directory(path: "/tmp"): Directory! 3.2s
    ● start docker-entrypoint.sh redis-server 9.3s
    ┃ 17:C 27 Mar 2024 13:08:02.780 # WARNING Memory overcommit must be enabled! Without it, a background save or replication may fail under low memory condition. Being disabled
    ┃ an also cause failures without low memory condition, see https://github.com/jemalloc/jemalloc/issues/1328. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.
    ┃ nd then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
    ┃ 17:C 27 Mar 2024 13:08:02.781 * oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
    ┃ 17:C 27 Mar 2024 13:08:02.781 * Redis version=7.2.4, bits=64, commit=00000000, modified=0, pid=17, just started
[...]

PR #7213, opened 2024-04-28
engine: nested exec simplifications and service fixes

sipsma:
This is a combo of:
* A few parts extracted out from: https://github.com/dagger/dagger/pull/6916
* A cherry-pick of @vito's fix here: https://github.com/dagger/dagger/pull/6914

Basically, I finally got back to #6916 and found the bare minimum changes needed to get #6914 working fully, which is the first commit here. But then I also realized that I actually needed #6914 in order for services to work and existing tests to pass; basically, the changes ended up being codependent and I had to cherry pick that one.

Either way, each commit has the details in the message. It amounts to a lot of simplifications though besides the service bug fix:
* ModuleCallerDigest is gone (it's just client ID now)
* *All* nested execs share a server with the main client caller (whereas previously module functions shared the server but plain execs that were nested did not)
* (from @vito's commit) ParentClientIDs is all gone too

Also worth noting this is hopefully just the beginning of simplifications, more that are made possible by these changes and some other recent ones:
* ServerID and entire concept of `DaggerServer` can be rm'd (will most likely end up doing that as part of #6916 now after revisiting it)
* ftp proxy hack can go away (helped enormously by removing all these IDs, but the new custom executor added in CA support should seal the deal here)

Added an integ test for the service bug fixed and a backfilled one for https://github.com/dagger/dagger/issues/6951

### DIFF
--- a/core/git.go
+++ b/core/git.go
@@ -60,14 +60,20 @@ func (*GitRef) TypeDescription() string {
 
 func (ref *GitRef) Tree(ctx context.Context) (*Directory, error) {
 	bk := ref.Query.Buildkit
-	st := ref.getState(ctx, bk)
-	return NewDirectorySt(ctx, ref.Query, *st, "", ref.Repo.Platform, ref.Repo.Services)
+	st, err := ref.getState(ctx, bk)
+	if err != nil {
+		return nil, err
+	}
+	return NewDirectorySt(ctx, ref.Query, st, "", ref.Repo.Platform, ref.Repo.Services)
 }
 
 func (ref *GitRef) Commit(ctx context.Context) (string, error) {
 	bk := ref.Query.Buildkit
-	st := ref.getState(ctx, bk)
-	p, err := resolveProvenance(ctx, bk, *st)
+	st, err := ref.getState(ctx, bk)
+	if err != nil {
+		return "", err
+	}
+	p, err := resolveProvenance(ctx, bk, st)
 	if err != nil {
 		return "", err
 	}
@@ -77,7 +83,7 @@ func (ref *GitRef) Commit(ctx context.Context) (string, error) {
 	return p.Sources.Git[0].Commit, nil
 }
 
-func (ref *GitRef) getState(ctx context.Context, bk *buildkit.Client) *llb.State {
+func (ref *GitRef) getState(ctx context.Context, bk *buildkit.Client) (llb.State, error) {
 	opts := []llb.GitOption{}
 
 	if ref.Repo.KeepGitDir {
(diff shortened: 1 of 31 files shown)


sipsma:
Only test failures are `compute cache` which has upstream buildkit fixes pending: https://github.com/dagger/dagger/issues/6111

Not even bothering with reruns since it's so flaky, passes locally and a known separate issue

engine/buildkit/socket.go
```diff
@@ -29,10 +33,28 @@ func (p *socketProxy) ForwardAgent(stream sshforward.SSH_ForwardAgentServer) err
...
+	if v, ok := opts[sshforward.KeySSHID]; ok && len(v) > 0 && v[0] != "" {
+		var err error
+		connURL, err = url.Parse(v[0])
+		if err != nil {
+			return status.Errorf(codes.InvalidArgument, "invalid id: %s", err)
+		}
+	}
+
+	caller := p.c.MainClientCaller
+	if sessionID := connURL.Fragment; sessionID != "" {
```
sipsma, line 49: Was re-reading (smelling own farts) and noticed this needs a check for `connURL != nil`, will fix

core/git.go
```diff
@@ -96,26 +102,10 @@ func (ref *GitRef) getState(ctx context.Context, bk *buildkit.Client) *llb.State
...
-	var st llb.State
-	if useDNS {
-		// NB: only configure search domains if we're directly using a service, or
-		// if we're nested beneath another search domain.
-		//
-		// we have to be a bit selective here to avoid breaking Dockerfile builds
-		// that use a Buildkit frontend (# syntax = ...) that doesn't have the
-		// networks API cap.
-		//
-		// TODO: add API cap
```
vito, line 115: :+1: - I think this comment has been outdated for a while, should be fine to simplify. I might have only even noticed it while testing Bass's Dockerfile `# syntax` support, which has always been janky because of weird Docker <-> Buildkit <-> buildx versioning weirdness anyway (`oci-layout://` source doesn't work for example, so no Container.import). Either way, the c2c networking code doesn't use a networks API cap anymore.

edit: oh wait, this is my own change lol. well I APPROVE

core/integration/module_test.go
```diff
@@ -5628,6 +5628,118 @@ func TestModuleUnicodePath(t *testing.T) {
 	require.JSONEq(t, `{"test":{"hello":"hello"}}`, out)
 }
 
+func TestModuleStartServices(t *testing.T) {
```
vito, line 5631: :heart_eyes:

vito (approved):
So happy to see this I might even attempt to explain it to my parents on vacation.

PR #7216, opened 2024-04-29
chore: keep pointer receiver consistent in generated go code

jedevc:
Some linters don't like this, which is a small issue, but it's super easy to fix :D


sipsma (approved):
This change seems fine, but sort of confused that a linter would complain about this since it's completely legit+expected to sometimes need a mix of pointer and non-pointer receivers. Hopefully it's intelligent enough to identify those cases. Just mentioning since I don't think we can necessarily guarantee we will always follow this rule indefinitely.

jedevc:
Ugh, can't re-open this PR since I force-pushed the branch after closing it? Opened #7219 instead.

PR #7219, opened 2024-04-30
chore: keep pointer receiver consistent in generated go code

jedevc:
Alternative to #7216 - turns out, #7216 *is* an issue but for the opposite reason.

`MarshalJSON` should always be defined on the base, not on a pointer receiver - we had mismatched between the object + interface in our codegen, this PR updates our interface to be the pointer receiver.

Sorry @aweris :cry:


sipsma (commented):
I think this will change the behavior such that previously we'd return `""` for nil structs but now we will return `null`.

No tests are failing, which is good, but it seems like it *might* at least trigger this log (though just in theory, worth confirmation): https://github.com/sipsma/dagger/blob/df31953deab1f0deb8f1b9ed8872f595b0de67f2/core/interface.go#L28-L28

Maybe it's worth removing that log if it's now expected to get hit? Just worth a double check cc @vito

PR #7224, opened 2024-04-30
fixed generated default values for typescript SDK

0xbochi:


helderco:
Thanks @0xbochi! Please regenerate the client by running `./hack/make sdk:typescript:generate`.

TomChv:
We worked together on this fix reported 2 weeks ago!

I tested this manually and it fixed the issue :rocket:

**Repro**

- Create 2 modules (e.g., `foo` and `target`)
- Add the following code:

```typescript
// foo/index.ts

import { dag, Container, object, func } from "@dagger.io/dagger"

@object()
class foo {
  @func()
  containerEcho(stringArg: string = "Hello, world"): Container {
  }
}
```

```typescript
// target/index.ts

import { dag, Container, Directory, object, func } from "@dagger.io/dagger"

@object()
class Target {
  @func()
  async coucou(): Promise<Container> {
    return dag.foo().containerEcho() // We no longer need the parameter, but we can set it in `{ stringArg }`, otherwise the default value is used.
  }
}
```

- Use foo in module target.
- The TS generator now generate arguments with default value as optional.

Thank you @0xbochi for the collaboration!

TomChv (approved):

.changes/unreleased/Fixed-20240430-164552.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixed
+body: Fixed default values typescript generator
```
helderco, line 2:
```suggestion
body: Fixed default values generation for TypeScript function arguments
```

helderco (approved):
Thanks again @0xbochi! This could use some test coverage, but will approve for now.

PR #7228, opened 2024-05-01
engine: remove ftp_proxy hack

sipsma:
This removes the ftp_proxy hack we've had for a while for passing uncached metadata to containers (previously a lot, but recently trimmed down to just ServerID).

Instead, we take advantage of the new custom executor added recently and use that to set the _DAGGER_SERVER_ID env var in containers. The plumbing passes a custom worker around via job values, which get read by the solver's ResolveOp callback and used to tell the Executor for ExecOps which ServerID to set in the container.

There's quite a bit more cleanup coming to everything involved here, but wanted to get this first step out since that hack, besides being annoying, was kind of dubious in nature and was a suspect in some strange behavior reported sometimes.

---

FTR, this is another spin-off of work on https://github.com/dagger/dagger/pull/6916. It was self-contained and beneficial enough that it seemed worth getting out for the next release.


sipsma:
Oddly, a few SDK example tests failed on `Container.Build` due to `serverID` not being set in the executor, but locally the `ContainerBuild*` integ tests pass against my dev engine. Not sure why, getting late so may not fix up til tomorrow morning but worst case can just be a little less zealous in requiring it's always set (it's theoretically fine if not for Dockerfile builds, it just wasn't expected).

engine/buildkit/executor.go
```diff
@@ -1,4 +1,4 @@
-package main
+package buildkit
```
jedevc, line 1: Could we add a note to this file from where it's pulled from in buildkit? Just thinking that like the git source/etc, we'll need to make sure that we port changes/fixes from upstream here.

Out of curiosity, do you see a path here to making some updates upstream so that we could re-use more parts of the runc executor specifically? It feels like we want some generic "hooks" before+after the runc execution, and that might let us get the required cacerts/ftp_proxying stuff.

Not a thing to necessarily work on right now, but I have some future work planned where I do want to remove our custom git source by allowing it to interface neatly with upstream, so I might also include this - if you think it's potentially at all possible.

engine/server/buildkitcontroller.go
```diff
@@ -94,7 +99,41 @@ func NewBuildkitController(opts BuildkitControllerOpts) (*BuildkitController, er
 
 	genericSolver := solver.NewSolver(solver.SolverOpt{
 		ResolveOpFunc: func(vtx solver.Vertex, builder solver.Builder) (solver.Op, error) {
-			return w.ResolveOp(vtx, llbSolver.Bridge(builder), opts.SessionManager)
+			// if this is an ExecOp, pass in an executor configured with the server ID as found
+			// in the job keys
```
jedevc, line 103: This is a neat trick :tada: I hadn't realized we could plug in `ResolveOp` like this, I like it :laughing:

jedevc (approved):
LGTM, once various SDK things are resolved :tada: Nice to see our metadata hack go away!

engine/server/buildkitcontroller.go
sipsma: Haha yeah, you actually *have* to provide this callback, so we're even just using things as intended now!

engine/buildkit/executor.go
jedevc: Yeah, fair enough, all very legitimate :laughing: Agreed with all the above points!

sipsma: Added comment with that link

sipsma:
Okay fixed that problem with Dockerfiles that use syntax pragmas in the not-fastest-but-best-long-term way. The high-level idea is the same as before but with some adjustments described in commit message here: https://github.com/dagger/dagger/pull/7228/commits/7baea5b291fb346e5ed6e84b67ef206559ac0adf
* cc @jedevc worth another look if interested

Side note: I noticed during those changes that `opTrackingGateway` seemed to amount to a no-op but introduced some convolution in trying to follow code, so I deleted it in this commit: https://github.com/dagger/dagger/pull/7228/commits/d0a00e5f4ee9a91b7871853942be588d03c025c5.
* Would appreciate a quick double check I'm not missing something obvious
* cc @vito (just if you are around, not sure of status, please ignore if vacationing 😄)

sipsma:
Also worth noting: the tests have been **way** less flaky so far (knock on wood, etc.). My educated guess is that ftp_proxy settings were resulting in different vtx digests that ended up edge merged, but now vtx digests end up the same more often and don't need to go through the (currently buggy, pending upstream fixes) edge merge logic?

I did still see the `compute cache` error in a publish workflow here once, but tests have not flaked out yet. Hopefully I'm not just getting lucky 🍀

sipsma:
I just pivoted to start on proxy support and immediately realized that I very much want the changes here because otherwise I'll just be recreating the same problems we had with `ftp_proxy` (even though we would be using those settings for their intended purpose now 😅), so I'm actually gonna go ahead and merge this but if you have any comments about the modifications since your last approval @jedevc please leave them and I'll make any follow-ups if needed!

PR #6934, opened 2024-03-26
💥 Remove deprecated schema fields

jedevc:
Removes previously deprecated fields - we've had these fields deprecated for several months at this point, so hopefully this should cause minimal impact.

PR #7233, opened 2024-05-01
chore: use `load*FromID` in tests

jedevc:
Use these newer methods instead of using `container` and `file` to load IDs, which is deprecated and *going to be removed soon*.

This is now possible because of #7038, and is a step towards #6934.


jedevc:
Ah, I started with just updating the rust tests, but realized I might as well grab all of them while in the area.

kjuulh (approved):

PR #7234, opened 2024-05-01
ci: exclude .git dir in ci build

jedevc:
Follow-up to #7231.

As a fun coincidence, this should fix the continued weirdness when doing this build in a git-worktree.

Previously, we were using this to generate a work-tree hash, so that new builds would have a different version tag, which helped to solve some weird caching issues with nested execs.

However! A problem with this was that we needed to upload *the entire* git directory every time we built - which is an expensive operation, my .git directory for dagger/dagger is 147MB.

Instead of relying on git for this, we can use the ID of the input directory! Because of our use of blob sources, a different ID represents a different directory. Additionally, we prefix these new builds with a special "dev-" string to specifically indicate that these are dev builds, with no specific version passed in - these aren't commit hashes, and cannot be treated as such.


jedevc:
With this change and with #7231, we've massively cut down the size of the context transferred here:

```
$ du -h --max-depth=0 .
385M	.
$ du -sh -- * .* | sort -h
...
147M	.git
182M	bin
```

Without uploading these expensive directories, we now only need to transfer 56MB.

sipsma (approved):
Nice! Yeah I had wondered when doing this if we could have added a new API for getting a content hash, but using the hash already in the ID is even easier 🎉

PR #7255, opened 2024-05-02
engine: support for system proxy settings

sipsma:
(related issue: https://github.com/dagger/dagger/issues/6599)

This adds support for:
* Automatically passing standard proxy settings (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, `ALL_PROXY`, `FTP_PROXY`) from the engine to all containers.
* Automatically passing custom `GOPROXY` settings from the engine container to the Go SDK.
   * Specifically, the engine container must have `_DAGGER_ENGINE_SYSTEMENV_GOPROXY` set for that value to be passed as `GOPROXY` to the Go SDK.
   * More details in commit message here: https://github.com/dagger/dagger/pull/7255/commits/c20f761c9f5dd7c95c461abea73c8cdff642d961

There's integ test coverage for everything except:
* `FTP_PROXY` - too obscure to be worth it right now
* `ALL_PROXY` - seems obscure and interpreted differently by different programs (or not interpreted at all, i.e. by Go stdlib). The other tests cover passing of these env vars well enough anyways.

Commit messages have more thorough descriptions of plumbing+reasoning as needed.


core/schema/container.go
```diff
@@ -122,6 +122,13 @@ func (s *containerSchema) Install() {
 					`environment variables defined in the container (e.g.,
 				"/opt/bin:$PATH").`),
 
+		// NOTE: this is internal-only for now (hidden from codegen via the __ prefix) as we
+		// currently only want to use it for allowing the Go SDK to inherit custom GOPROXY
+		// settings from the engine container. It may be made public in the future with more
+		// refined design.
+		dagql.Func("__withSystemEnvVariable", s.withSystemEnvVariable).
+			Doc(`(Internal-only) Inherit this environment variable from the engine container if set there with a special prefix.`),
```
jedevc, line 130: This is more just something to note - the number of "internal-only APIs" has kind of started to grow.

I'd really like to try splitting some of these out, ideally by having a concept of a "privileged" session, which serves those APIs - other calls wouldn't see it. Then codegen wouldn't generate any of these for those.

jedevc (approved):

core/schema/container.go
sipsma: It would be nice to have something more fleshed out here I agree, but this `__` prefix does [result in them being hidden from codegen](https://github.com/sipsma/dagger/blob/1bc2cffabc80e614afea9dc0155a0e77af1aa4c4/dagql/introspection/types.go#L519-L519) (just to be clear, not sure if that's what you were getting at with "Then codegen wouldn't generate any of these for those").

PR #7259, opened 2024-05-02
telemetry: handle more attr type conversions

vito:
We use a StringSlice for process args, so we actually hit this branch pretty often. Here's a quick and dirty attempt at handling the cases that we can handle.

### DIFF
--- a/telemetry/servers.go
+++ b/telemetry/servers.go
@@ -195,9 +195,59 @@ func attrValue(v *otlpcommonv1.AnyValue) attribute.Value {
 		return attribute.Int64Value(v.GetIntValue())
 	case *otlpcommonv1.AnyValue_BoolValue:
 		return attribute.BoolValue(v.GetBoolValue())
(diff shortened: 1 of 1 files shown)


telemetry/servers.go
```diff
@@ -195,9 +195,59 @@ func attrValue(v *otlpcommonv1.AnyValue) attribute.Value {
...
+					for _, v := range vals {
+						bools = append(bools, v.AsBool())
+					}
+					return attribute.BoolSliceValue(bools)
+				default:
+					slog.Error("otlpcommonv1.AnyValue -> attribute.Value: unhandled array value type conversion", "type", fmt.Sprintf("%T", x))
+					return attribute.StringValue(fmt.Sprintf("UNHANDLED ARRAY ELEM TYPE: %+v (%s)", vals, t))
+				}
+			}
+			panic("unreachable")
```
jedevc, line 242: Bleh, of course go can't realize this is impossible :cry:

jedevc (approved):

PR #7262, opened 2024-05-03
cli: visual cue for required flags

grouville:

Append [required] to flags that are marked as required. As proposed by Helder, this follows the behavior of Typer: https://typer.tiangolo.com/tutorial/options/required/

Required flags are shown first, then option arguments are shown. Apart from this, they are sorted in the order of the first flagset: https://discord.com/channels/707636530424053791/1232477387724292209/1232711077415092295

**Example**
```shell
./dagger -m github.com/jpadams/chainloop@8374b22aee238452467bd712e874937b8e4254c9 call  init --help

OPTIONS
      --token Secret               Chainloop API token [required]
      --contract-revision string   Workflow Contract revision, default is the latest
      --repository Directory       Path to the source repository to be attested

```

### DIFF
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -356,7 +356,8 @@ func sortRequiredFlags(originalFlags *pflag.FlagSet) *pflag.FlagSet {
 	// separate optional flags from required flags
 	originalFlags.VisitAll(func(flag *pflag.Flag) {
 		// Append [required] and show required flags first
-		if flag.Annotations["cobra_annotation_bash_completion_one_required_flag"] != nil {
+		requiredAnnotation, found := flag.Annotations[cobra.BashCompOneRequiredFlag]
+		if found && requiredAnnotation[0] == "true" {
 			flag.Usage += " [required]"
 			mergedFlags.AddFlag(flag)
 		} else {
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -367,9 +367,7 @@ func sortRequiredFlags(originalFlags *pflag.FlagSet) *pflag.FlagSet {
 	})
 
 	// Add optional flags back, after all required flags
-	optionalFlags.VisitAll(func(flag *pflag.Flag) {
-		mergedFlags.AddFlag(flag)
-	})
+	mergedFlags.AddFlagSet(optionalFlags)
 
 	return mergedFlags
 }
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -88,7 +88,7 @@ func init() {
 	cobra.AddTemplateFunc("flagUsagesWrapped", flagUsagesWrapped)
 	cobra.AddTemplateFunc("cmdShortWrapped", cmdShortWrapped)
 	cobra.AddTemplateFunc("toUpperBold", toUpperBold)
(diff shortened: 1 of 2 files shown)


cmd/dagger/main.go
```diff
@@ -345,6 +346,34 @@ func toUpperBold(s string) string {
 	return termenv.String(upperCase).Bold().String()
 }
 
+// separateAndModifyFlags separates optional flags from required flags
+func separateAndModifyFlags(originalFlags *pflag.FlagSet) *pflag.FlagSet {
```
helderco, line 350: Name isn't descriptive of what it does. I suggest: `sortRequiredFlags`.

cmd/dagger/main.go
```diff
@@ -345,6 +346,34 @@ func toUpperBold(s string) string {
...
+	mergedFlags := pflag.NewFlagSet("merged", pflag.ContinueOnError)
+	mergedFlags.SortFlags = false
+
+	optionalFlags := pflag.NewFlagSet("optional", pflag.ContinueOnError)
+
+	// separate optional flags from required flags
+	originalFlags.VisitAll(func(flag *pflag.Flag) {
+		clone := *flag
```
helderco, line 358: This isn’t enough. It’ll make a shallow clone of the `Flag` struct but will leave out the annotations map and you need it for validation. Even if you then also manually clone the annotations, it’s possible in a later update a slice field is added to `Flag`. I think it’s ok to change the flag here without cloning.

Another alternative would be to append `[required]` only to function arguments, while creating the flags for them as I don’t expect we’ll use required flags anywhere else.

cmd/dagger/main.go
```diff
@@ -345,6 +346,34 @@ func toUpperBold(s string) string {
...
+			mergedFlags.AddFlag(&clone)
+		} else {
+			optionalFlags.AddFlag(&clone)
+		}
+	})
+
+	// Add optional flags back, after all required flags
+	optionalFlags.VisitAll(func(flag *pflag.Flag) {
+		mergedFlags.AddFlag(flag)
+	})
```
helderco, line 372: This can be simplified:
```suggestion
	mergedFlags.AddFlagSet(optionalFlags)
```

cmd/dagger/main.go
```diff
@@ -345,6 +346,34 @@ func toUpperBold(s string) string {
...
+	mergedFlags.SortFlags = false
+
+	optionalFlags := pflag.NewFlagSet("optional", pflag.ContinueOnError)
+
+	// separate optional flags from required flags
+	originalFlags.VisitAll(func(flag *pflag.Flag) {
+		clone := *flag
+
+		// Append [required] and show required flags first
+		if flag.Annotations["cobra_annotation_bash_completion_one_required_flag"] != nil {
```
helderco, line 361: Better to use the const:
```suggestion
		if flag.Annotations[cobra.BashCompOneRequiredFlag] != nil {
```

helderco (commented):
Outstanding! 🎉

Main change is the shallow clone drops the annotations.

helderco (approved):
Pushed some quick changes based on feedback so I can approve this quickly 🙂

Now we need to change "Options" to "Function Arguments" in `dagger call` so we don't have "required options" anymore. I can take that.

Issue #7220, opened 2024-04-30
OpenSSL cannot read SSH keys mounted as secrets


sagikazarmark:
### What is the issue?

While working on #7202 I noticed that mounting an SSH key as a secret in a container always leads to `Load key "/ssh-key": error in libcrypto` errors.

### Dagger version

dagger v0.11.2 ([registry.dagger.io/engine](http://registry.dagger.io/engine)) darwin/arm64

### Steps to reproduce

The code reproducing the issue can be found in my daggerverse:

```shell
git clone https://github.com/sagikazarmark/daggerverse.git
cd daggerverse
git checkout ssh-key-repro
cd ssh-key-repro
chmod 600 id_ed25519
 
# Test that mounts the ssh key as a file
dagger call test-ok stdout
 
# Test that reads the ssh key as a file and creates a secret
dagger call test-ok-too stdout
 
# Test that reads the ssh key as a file and creates a secret
dagger call test-fail --ssh-key file:id_ed25519 stdout
```

### Log output

```
❯ dagger -- call test-ok stdout
ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIF0PEuo6tWBJ3kuKjBAx/MxxKnzO6AwL/rRo2YjxqA7i mark@Mark-M2MBP.localdomain
 
❯ dagger -- call test-ok-too stdout
ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIF0PEuo6tWBJ3kuKjBAx/MxxKnzO6AwL/rRo2YjxqA7i mark@Mark-M2MBP.localdomain
 
❯ dagger -- call test-fail --ssh-key file:id_ed25519 stdout
✔ SshKeyRepro.testFail(
    sshKey: ✔ setSecret(name: "60a21fa96beb12b2d78c933778c32887cf5ddfbffc6d2755e4ad8d889c693c5f"): Secret! 0.0s
  ): Container! 1.4s
  ✔ Wolfi.container(packages: ["git", "openssh"]): Container! 1.2s
    ✔ Apko.wolfi(packages: ["git", "openssh"]): Container! 1.0s
      ✔ Container.from(address: "cgr.dev/chainguard/apko"): Container! 0.7s
        ✔ HTTP GET 0.4s
        ✔ remotes.docker.resolver.HTTPRequest 0.2s
          ✔ HTTP HEAD 0.2s
✘ Container.stdout: String! 0.1s
[...]

PR #6845, opened 2024-03-06
fix file and cmd secret source inputs

kpenfound:
Reported here: https://github.com/dagger/dagger/pull/6829#discussion_r1513464630

Based on my testing, the following did not work:

```
dagger call foo --token env:TOKEN
echo $TOKEN > token.txt
dagger call foo --token file:./token.txt
```

Breaking out the file parsing code, the `token.txt` would be parsed with a newline character at the end.

This adds `TrimSpace()` to file and cmd secret inputs. `TrimSpace()` trims leading and trailing unicode whitespace characters. I can't think of any cases where this would have unintended consequences for secrets.

PR #7271, opened 2024-05-03
Revert secret whitespace stripping

marcosnils:
Fixes https://github.com/dagger/dagger/issues/7220

Reverts dagger/dagger#6845 given that seems like it triggers some edge cases that we haven't spotted and we don't seem to have a better option for now


jedevc:
Can we have a changelog?

marcosnils:
> Can we have a changelog?

we sure do!

jedevc (approved):

sagikazarmark:
If it's already merged, isn't it going to be in 0.11.3?

PR #7223, opened 2024-04-30
ci: use nested dagger for testdev

jedevc:
Specifically! This doesn't spin up a docker container with a dev version of dagger, it spins up a dagger service with a dev version of dagger - dagger in dagger!

This simplifies deployment to the new CI runners (since docker is no longer needed).

---

At some point, it would be nice to have the `./hack/dev` script use this as well - however, I'm not 100% sure how this would work with other core devs' workflows. Personally, I periodically run `./hack/dev` to build and restart the docker engine, and use `./hack/with-dev` to run commands in that context. This doesn't work *as* well with this setup, where it's more difficult to observe the logs in a running service, etc.

So maybe we need to keep `./hack/dev` at least for now - we can discuss and maybe do this in the future though.

PR #7272, opened 2024-05-03
feat: improve plain progress

### LINKED ISSUE #7137: ✨ Output in automatic provisioning like 0.10

Dagger 0.11 made changes to the console output which I believe to be a large downgrade.

### What I used to do:

I would run my pipeline without the dagger binary, simply running `python main.py`. This would print the output from the container as well as my program output (i.e., calls to `logger.info()`) to the console in real time. Once the process was complete, all the container logs + my own would still be clearly visible for reference.

I found this approach to be the best for 2 reasons:
* I did not have to install the dagger binary, I just needed to make sure that `dagger-io` was installed in my Python virtual environment.
* It produced output that was by far the most understandable, readable, and as a result, useful.

### The Downgrade

With dagger 0.11, simply running `python main.py` produces only the program output and none from the containerised process itself. 

The best alternative is to run `dagger run -vvv python main.py`, but I find that to be significantly worse than running `python main.py` with dagger 0.10 due to the following points:
* It requires the dagger binary to be installed (extra work)
* The output is split into 2 sections, "container output" first and "program output" second, which is harder to make sense of.
* The container output is significantly summarised and not nearly as verbose as that generated in earlier versions.

Seeing the output of a program is incredibly helpful for pretty much every part of the development process.

I am not upgrading to dagger 0.11, and using 0.10 instead.

jedevc:
Fixes #7137.

With the changes as part of #6835 (in v0.11.0), all old TUIs were removed, including the old `--progress=plain` output. With #7069 (in v0.11.1), a limited version of the plain TUI was restored. However, as noted in https://github.com/dagger/dagger/issues/7137#issuecomment-2090450408 and https://github.com/dagger/dagger/pull/7223#issuecomment-2090499541, this new plain progress has issues - it only reports spans that have logs, and loses a lot of additional information.

This PR attempts to strike a balance between the v0.10.X plain output and the v0.11.X plain output - unfortunately, the old behavior cannot be perfectly restored due to the architectural changes in the move to OpenTelemetry.

Specifically, some guiding ideas I'm trying to stick to:
- We should share as much code as possible between plain + tty output. We want to ensure that the visibility of spans is the same between both outputs, and that the verbosity options create roughly the same effects for both.
- We should attempt to organize the logs by function call as much as possible - this is the direction of the tty output, as well as dagger cloud - being aligned keeps the outputs somewhat similar.
- We should keep a familiar style to the old plain logs, with numerical prefixes, similar colors, etc.
- We should attempt to default to *linearly* readable logs in the case of *linear* progress (e.g. a linear series of calls in the CLI should display linearly in the progress). We should aim to show the *exact* amount of interesting information. More complex parallel pipelines are out of scope, there is no way to do this neatly in a log format.


jedevc:
As a sneak preview of what I'm building towards at the moment:

image

gerhard:
Looking good @jedevc 💪

jedevc:
Okay, continuing to make progress here, did some refactoring, but also started reworking the output a bit.

One of the main problems with plain progress is that we lose a *lot* of context for when anything appears. This was always kind of a problem, but it's now really quite annoying. So ideally, what we should do is try and provide some amount of context for each vertex.

One idea I've been playing around with is to show each vertex with its parent call (if there is one):

image

The merging between parts isn't quite right, but the grouping does seem to help improve readability. Unlike the pretty progress, it's not really doable to nest these indefinitely, since this will really clog up the output - but showing the most recent parent call does really help to understand where something comes from.

jedevc:
Okay, so a lot of work from this got spun out into separate PRs: https://github.com/dagger/dagger/pull/7347 https://github.com/dagger/dagger/pull/7371 https://github.com/dagger/dagger/pull/7385

I also got some more time to focus on this one for a bit yesterday and today - I refactored most of the progress, and so now we have a lot of nice shared code between the plain and the pretty output. Here's the current state:

image

There's a few more steps to take early next week before this is done:
- Introduce a small delay buffer for context switching (similar to what `streaming_output.go` was doing) - this will help for grouping, as well as avoiding out-of-order issues caused by logs and spans received in interesting orders.
- Add some vertical spacing for visual separation
- Tidy up any specifically nasty hacks
- Get some user feedback

jpadams:
@jedevc I like the way this renders a lot. I don't think it's part of this PR likely, but I'd still like to see less/none of the `HTTP HEAD/POST/PUT` calls.
image

jedevc:
Ah thanks @jpadams - I accidentally managed to refactor away the logic that hid these spans, some checks were in the wrong order.

You can see the new logic for hiding spans is here now: https://github.com/jedevc/dagger/blob/aadc3512762b39d52858c9ea4ddc6561e56dd1ed/dagql/idtui/frontend.go#L324-L356

jedevc:
Also realized `--silent` had stopped working on `--progress=plain`, so have fixed that too.

**No** idea why the python sdk tests have started hanging though.

dagql/idtui/frontend.go
```diff
@@ -55,6 +57,257 @@ type Frontend interface {
...
+						return err
+					}
+				} else {
+					r.renderLiteral(out, arg.GetValue())
+					fmt.Fprintln(out)
+				}
+			}
+			depth--
+			indent(out, depth)
+			depth-- //nolint:ineffassign
```
sipsma, line 165: wtf, silly linter... 😄

sipsma (approved):
Code LGTM! Obviously need to solve the test failure mystery, but otherwise good to merge

cmd/dump-id/main.go
```diff
@@ -10,6 +10,8 @@ import (
 )
 
 func main() {
+	// XXX: check that this still works!
```
vito, line 13: Not a big deal if it doesn't, it's already a bit nerfed from TUI refactoring. It used to give you a deep dump of the ID, but now it abbreviates into things like `Container.withExec` which isn't as useful.

dagql/idtui/frontend.go
```diff
@@ -17,631 +16,75 @@ import (
...
-		// doesn't have any garbage before/after
-		return ""
-	}
-	if fe.done && fe.eof {
-		// print nothing; make way for the pristine output in the final render
-		return ""
-	}
-	return view
+	// ConnectedToEngine is called when the CLI connects to an engine.
+	// TODO: remove this
```
vito, line 53: Mostly curious, any particular reason to remove it? The thinking was that a plain UI would just print this, but a pretty TUI might show it somewhere onscreen

dagql/idtui/frontend.go
```diff
@@ -17,631 +16,75 @@ import (
...
-			sigquit()
-			return fe, nil
-		default:
-			return fe, nil
-		}
-
-	case tea.WindowSizeMsg:
-		fe.SetWindowSize(msg)
-		return fe, nil
+type Frontend interface {
```
vito, line 38: :+1:

vito (approved):
First off, this looks great! I love that it keeps all the info and coloring from the TUI, and it doesn't feel too insanely verbose. The refactor is very sensible too.

Purely UI feedback, but we can iterate, it's already better than what we have:

> I mentioned spacing previously, and I think keeping the top-level items separate but keeping the nested items flush is probably the best we can do. Besides maybe double-spacing at the top and single-spacing at the inner levels. Maybe worth trying, but dunno if it'd be better.
>
> That being said it seems like sometimes calls "break out" into separate sections instead of being nested properly. Maybe a big? (See the extra "Apko.wolfi" header below)

image


> Could use a blank line here above "Container evaluated." to make it feel more final.

image

> Since we're already omitting args on the "DONE" line maybe we should omit the return type too? Seems slightly noisy.

image

> The spacing on the numbers feels a little awkward but I get that it keeps things aligned. Maybe the number could be right-aligned so it's flush with the `:`? Maybe 0-padded?

image

> MAYBE the top-level inline args are indented one level too far? Not sure if intentional, but it means the children align a little differently than when it's not a multiline code block.

image

That's all I got for now! Again nothing blocking.

dagql/idtui/frontend.go
jedevc: It gets refactored away in https://github.com/dagger/dagger/pull/7385/files#diff-375c41f6a94e797edee687b8fa151865e2192d8aa9088acaffa174d0de2c3578R284

We can just use `slog` here (at least for now).

jedevc:
Rebased onto main after #7385 was merged.

jedevc:
Right. I think I have a handle on what's going on. Compare the [failing check](https://github.com/dagger/dagger/actions/runs/9211758328/job/25341857993?pr=7272) on this PR here and [the passing check](https://github.com/dagger/dagger/actions/runs/9211781489/job/25341928942?pr=7450) on my [work-in-progress PR here](https://github.com/dagger/dagger/actions/runs/9211781489/job/25341928942?pr=7450).

There's one key difference here: [`100aaf2` (#7450)](https://github.com/dagger/dagger/pull/7450/commits/100aaf2f0030544d17f53f80b0bbb506c9bcd718)

Here's my best guess what's going on - somewhere, there's an unexpected layer of output buffering, which is preventing flushing data from the output stream - this seems to manifest as blocking something critical in a python test. If we just become more chatty on our logs... we flush more data. 

Alongside the *hard* evidence of the different PRs:
- This explains why this PR is a "regression" - the new plain progress is a *lot* less chatty, so we might flush less.
- This would explain why enabling `--debug` "fixes" the issue - we re-enable a ton of chatty output, so we would flush more.

From the python subprocess side, we do set `bufsize=0` explicitly - this should disable buffering. I wonder if there's still maybe something to pull on here though, because we *don't* see this in any other language SDK.

helderco:
> From the python subprocess side, we do set `bufsize=0` explicitly - this should disable buffering. I wonder if there's still maybe something to pull on here though, because we _don't_ see this in any other language SDK.

Interesting! Yeah, the runtime also disables it for modules:

https://github.com/dagger/dagger/blob/643aa617eda6aabb16550dda553824421f06d2c4/sdk/python/runtime/main.go#L160

But CI doesn't:

https://github.com/dagger/dagger/blob/643aa617eda6aabb16550dda553824421f06d2c4/ci/sdk_python.go#L181-L183

It's an easy change to put there.

jedevc:
Hm, that doesn't seem to have worked.

Something weird here is also that the extra chattiness I added was for stderr - while I think we do a lot of our *essential* session communication over stdout. Still not quite sure what this could be though.

jedevc:
Finally - I've determined what this is. The *default* here is to hang - see a reproducer here: https://github.com/jedevc/dagger/tree/fix-weird-stall/examples/sdk/go/repro. Something about the weirdly-constructed nested sessions causes a hang - however, it has for a while, this PR doesn't change it.

So why hasn't this been hanging in this test forever? Well, all is revealed if we add a `print(proc.returncode)` here: https://github.com/dagger/dagger/blob/6efd184405408359352c8ca9e1a0f150592e1fe4/sdk/python/src/dagger/_engine/session.py#L34-L41

This prints `-13` - which indicates the process terminated by signal 13 - which is `SIGPIPE` (attempting to write to a pipe with no readers). This happens because of how `__exit__` works in python - it closes `proc.stdout`+`proc.stderr` before `wait`ing. Somehow, the previous progress printing would attempt to keep writing progress (which is fairly reasonable), but would attempt to print to a closed `stderr` - which causes `SIGPIPE`, so the `dagger session` process crashes, without getting to hang.

This plain progress PR adjusts the pace/timing of progress output - and clearly means that we don't violate this issue most of the time (note, this also explains why this issue seems to be racy, and not occur every time - tiny changes in timing from different python versions might cause `proc.stdout.close` to occur *before* the progress finishes printing, to avoid the hang.

However this doesn't happen in the go provisioning tests. It hangs when run manually - `cd sdk/go && with-dev dagger run go test -run=TestProvision -v ./...`. It *also* hangs when running `with-dev dagger call --source=.:default -v sdk go test` (but not `dagger call --source=.:default -v sdk go test`) - weirdly, this isn't at *all* related to this branch, or even `main` - this can be reproduced on a checkout of `v0.11.4`. I wonder if there's something weird with telemetry draining these kind of nested calls when the calling cli and the nested cli both use the same dagger engine:

- This passes: `../../bin/dagger run --silent sh -c "_EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-container://dagger-engine.dev _EXPERIMENTAL_DAGGER_CLI_BIN=../../bin/dagger go test -run=TestProvision -test.v -count=1 ./..."`
- This hangs: `_EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-container://dagger-engine.dev ../../bin/dagger run --silent sh -c "_EXPERIMENTAL_DAGGER_CLI_BIN=../../bin/dagger go test -run=TestProvision -test.v -count=1 ./..."`

There is something *incredibly* cursed about the telemetry in this kind of setup - the python code has always been hanging, at least as long as OTEL (but passing because of the `SIGPIPE`). Because of the rarity of these kind of nested sessions, I think it makes sense to disable the provision tests from running inside of dagger, opening another issue to track those issues, and merging this PR as-is.

Issue #6515, opened 2024-01-29
CLI: print module object fields with `dagger call`


helderco:
## Problem

When `dagger call` ends on an object (as opposed to a scalar value), it returns an error with a message saying "please query a field of this object".

This is a problem because:

1. At a glance, it gives the impression that the user did something wrong, when in fact they did not. So an error is sending the wrong message.
2. It makes a valid task (produce an object and look at it) harder than it should be. Just show me the object!

## Solution

1. Never show an error when a function has successfully be called, and has returned an object
2. Print the contents of the object in a useful format

There are two options for *what to print*:

### Option 1: call all scalar functions

In this option, we call all functions in the object that return a scalar value, and print the result, either in JSON, YAML or pretty-print format (configurable by a flag). Possibly excluding the ID by default (also configurable).

*On the question of "expensive scalar functions*: for example, `Container.stdout()` is "expensive" because it forces actually executing a container, in order to retrieve its standard output. Therefore, it is argued, the CLI should not blindly call all scalar functions of an object, because it may cause expensive computation which the user does not want. *This is premature optimization*. Let's ship the feature in a naive implementation; then see how important it actually is to avoid expensive computation. My guess is that, in most cases, it won't matter. For example, if I query a container, I don't care that it's actually executed.  Or, if it turns out it's a problem, then we can implement a special case in the core type. I certainly won't care if my own custom types don't benefit from such a special case.

#### Pre-requisites

As a prerequisite, we need:
- https://github.com/dagger/dagger/issues/6514

### Option 2: print list of all functions

In this option, we don't call any functions. Instead, we print a list of them, possibly with a description. Again, this could be either pretty-printed, or exported to json or yaml.

In this scenario, `dagger call` becomes the simplest way to discover available functions. This in turn would allow simplifying `dagger call --help`, [which is quite noisy](#6943). `--help` would no longer need to print all functions of the return types: it could focus on printing available arguments. This solves a common problem, especially for functions that return core types like `Container` or `Directory`, which have many functions - these tend to drown out the flags.

shykes:
@helderco are you thinking we should print the *values* of the fields, or just their names?

I initially thought the former (show value of all scalar fields). But now I'm thinking that simply printing the name of all fields, would be easier and actually more useful.

shykes:
FYI @grouville

helderco:
> @helderco are you thinking we should print the _values_ of the fields, or just their names?

I was thinking both name and values.

[...]

PR #7277, opened 2024-05-03
CLI: simplify and clarify usage text

shykes:


grouville:
related to https://github.com/dagger/dagger/pull/7143

cmd/dagger/functions.go
```diff
@@ -34,7 +34,7 @@ const (
 
 var funcGroup = &cobra.Group{
 	ID:    "functions",
-	Title: "Function Commands",
+	Title: "Functions",
```
helderco, line 37: You can remove this as it's already integrated in:
- https://github.com/dagger/dagger/pull/7286

cmd/dagger/call.go
```diff
@@ -55,6 +44,7 @@ dagger call lint stdout
 		case Terminal:
 			c.Select("websocketEndpoint")
 		default:
+			// FIXME: don't make this an error, it's confusing to users
 			return fmt.Errorf("return type %q requires a sub-command", name)
```
helderco, line 48: If not for this you'll get an error from GraphQL anyway. It was added to provide a better error message. 

It will be shown less often when we print scalars from objects by default but there's some technical pre-requisites to allow for that.

This error still needs to exist though, since it's possible you have an object without scalars to print, if it only acts as a constructor to other objects to chain, for example. And if you return a `CacheVolume` for some reason, it only has an `id` field and I don't think we'll print ids as part of the feature mentioned above.

If the error is confusing to users, do you have a better alternative besides improving the error message?

cmd/dagger/call.go
```diff
@@ -17,28 +17,17 @@ var jsonOutput bool
...
-If the last argument is either a Container, Directory, or File, the pipeline
-will be evaluated (the result of calling ´sync´) without presenting any output.
-Providing the ´--output´ option (shorthand: ´-o´) is equivalent to calling
-´export´ instead. To print a property of these core objects, continue chaining
-by appending it to the end of the command (for example, ´stdout´, ´entries´, or
-´contents´).
+	Short: "Call a function, or a pipeline of functions",
+	Long: strings.ReplaceAll(`Call a function, or a pipeline of functions, and print the result.
+
+More information about pipelines: https://docs.dagger.io/manuals/user/539201/chaining
```
helderco, line 23: I agree with this approach, just giving a bit of context as I’ve talked about this with multiple people before.

I’ve had this thing about creating help topics:

- https://github.com/dagger/dagger/pull/6549
- https://github.com/dagger/dagger/issues/6943#issuecomment-2021310987
- https://github.com/dagger/dagger/pull/7107#issuecomment-2061525961

Quoting from that last comment:

> Not a priority, better create a separate issue for this. The intent is to provide more information to the user in the CLI (i.e., without requiring a visit to the docs), while reducing the clutter in normal `--help`. A good example for `dagger call` is how it works through chaining, what happens with a few core types (`Contaienr`, `Directory`…), how to use `--output` and `--json`, etc.
> 
> `gh` uses help topics for this, but `docker` relies on web documentation, doing its utmost to keep `--help` output to a minimum. I didn’t find any _help topic_ being used (as a feature) in that CLI.
> 
> Since Dagger has its nuances, I’ve been wondering which approach is better here. We could just “link” to the web for more information. \cc [@vikram-dagger](https://github.com/vikram-dagger)

So, with help topics, instead of showing lots of info in the long description, you’d have it on special sub-commands that don’t run, just display help (it’s a Cobra feature), and show in a section like the following example:

Screenshot 2024-05-06 at 00 32 17


But lately I’ve been leaning more strongly towards linking to the web:
- It has it’s own release cycle so we can fix something more quickly
- There’s more flexibility in how to present information
- Better medium to consume the information
- Keeps the CLI leaner and simpler

The downsides are:
- Information is no longer tied to the CLI version you’re using, only to the latest version (acceptable for several reasons, I think).
- Stronger need to keep redirects to old URLs depending on how often we change the address or re-structure. Easier if we generalize like the following example, at the end of `--help` output:
Screenshot 2024-05-06 at 00 52 08

helderco (approved):

helderco:
@shykes, I rebased. Only not merging because waiting for reply to https://github.com/dagger/dagger/pull/7277#discussion_r1590453771.

gerhard:
@helderco you may want to rebase this again.

These are the most important change that this PR has not picked up yet:
- https://github.com/dagger/dagger/pull/7082
- https://github.com/dagger/dagger/pull/7296

cmd/dagger/call.go
shykes: It's just a general note of a problem (calling a function that returns an object without error, should not return an error to the user). I don't have a specific solution at the moment (tried initially in this PR, but gave up). Hence the FIXME.

cmd/dagger/call.go
shykes: I will remove the URL from this PR, so we can fast-track it. I like the web-focused approach, but we probably need to be very careful of what URLs we bake in the CLI - once they're in, we're stuck with that infra forever.

My urgent priority in this PR was 1) to make the output shorter since it's in the critical path of my devloop, and 2) clean up terminology.

shykes: also now I have merge conflicts

cmd/dagger/call.go
helderco: To make a query you need to end in a scalar/leaf, so this error needs to exist. We can only reduce the chance of hitting it when we can print an object's fields but that won't replace this error 100% of the time.

shykes: Those are artificial constraints. If we decide we want to replace the error 100% of the time, it's easy to do so. The question is do we agree on doing it.

Example: if the object has no scalars, print nothing. Or, print the ID. Or, print the same message as before, but without prefixing it with "error". All these are plausible solutions to the problem.

shykes:
> @shykes, I rebased. Only not merging because waiting for reply to [#7277 (comment)](https://github.com/dagger/dagger/pull/7277#discussion_r1590453771).

Thanks, sorry I had missed your rebase earlier.

cmd/dagger/call.go
helderco: Yeah, I guess what I'm saying is that returning nothing or returning an ID doesn't seem better to me but let's discuss at another time.

cmd/dagger/call.go
```diff
@@ -55,6 +41,7 @@ dagger call lint stdout
 		case Terminal:
 			c.Select("websocketEndpoint")
 		default:
+			// FIXME: don't make this an error, it's confusing to users
```
helderco, line 44: This error has been replaced 100% with Option 2 from (PR soon):
- https://github.com/dagger/dagger/issues/6515

Basically any chain that ends in an object that has itself functions, lists them under "Available Functions".

shykes:
This was meant to be a quick win... Does it even make sense for me to update this, with other improvements in flight?

PR #7279, opened 2024-05-04
install.sh - fix bash completion instructions

iximiuz:
The XDG_DATA_HOME env var can be empty (expectedly), and it breaks the printed bash completion setup instructions. This little PR makes the install script use the default $HOME/.local/share location when the XDG_DATA_HOME is empty. See  https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -295,8 +295,8 @@ ${binexe} has built-in shell completion. This is how you can install it for:
 
     2. Add dagger completion to your personal bash completions dir
 
-      mkdir -p $XDG_DATA_HOME/bash-completion/completions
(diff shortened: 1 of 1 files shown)


marcosnils (approved):

jedevc (approved):

jedevc:
Nice catch, nice catch! Going to merge this now, but for this to appear on the dl site, we need to do some manual AWS hackery :tada: (cc @gerhard @matipan)

iximiuz:
Thanks for the quick approval!

gerhard:
Reminder to self + @matipan: add a Dagger function that uploads this to S3 & invalidates the cache.

Did this manually:
image

Thanks @iximiuz 💪

PR #7297, opened 2024-05-06
ca certs: include output of failed update command

sipsma:
We collected this output but didn't actually put it in the error message, which left us with only an exit code to debug with.

Now the actual output is there, which is much more useful.


sipsma:
Failure is just due to the known infra issue being worked on: https://github.com/dagger/dagger/actions/runs/8977335506/job/24655889101?pr=7297

All tests actually passed

jedevc (approved):

Issue #7192, opened 2024-04-25
🐞 `github.com/dagger/dagger/dagql/idtui.CollectSpan` invalid memory address or nil pointer dereference


gerhard:
### What is the issue?

We've hit this error a few times in https://github.com/dagger/dagger/actions/runs/8821414426/job/24217223788?pr=7082, part of:
- https://github.com/dagger/dagger/pull/7082

```
2024-04-24T18:36:16.3979153Z [95m117: [0mDONE 1325 tests, 3 skipped, 2 failures in 605.184s
2024-04-24T18:36:16.3979462Z 
2024-04-24T18:36:16.3979728Z panic: runtime error: invalid memory address or nil pointer dereference
2024-04-24T18:36:16.3980302Z [signal SIGSEGV: segmentation violation code=0x1 addr=0x4b pc=0xf4363d]
2024-04-24T18:36:16.3980649Z 
2024-04-24T18:36:16.3980743Z goroutine 1 [running]:
2024-04-24T18:36:16.3981331Z github.com/dagger/dagger/dagql/idtui.CollectSpans(0xc0003e0120, {0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...})
2024-04-24T18:36:16.3981982Z 	/app/dagql/idtui/types.go:52 +0x41d
2024-04-24T18:36:16.3982474Z github.com/dagger/dagger/dagql/idtui.(*Frontend).recalculateView(0xc0003e20d0)
2024-04-24T18:36:16.3983016Z 	/app/dagql/idtui/frontend.go:395 +0x33
2024-04-24T18:36:16.3983493Z github.com/dagger/dagger/dagql/idtui.(*Frontend).finalRender(0xc0003e20d0)
2024-04-24T18:36:16.3983984Z 	/app/dagql/idtui/frontend.go:217 +0xa5
2024-04-24T18:36:16.3984563Z github.com/dagger/dagger/dagql/idtui.(*Frontend).Run(0xc0003e20d0, {0x1419d10, 0x1d6d6c0}, 0x12e4450)
2024-04-24T18:36:16.3985182Z 	/app/dagql/idtui/frontend.go:131 +0xe5
2024-04-24T18:36:16.3985471Z main.main()
2024-04-24T18:36:16.3985706Z 	/app/cmd/dagger/main.go:212 +0xac
2024-04-24T18:36:16.4731230Z Error: exit status 2
[...]

PR #7301, opened 2024-05-07
docs: Adds cookbook

vikram-dagger:
This commit adds a cookbook with sample code for common use cases. It only contains a few examples right now, to set up the initial structure, formatting etc. For the same reason, the navbar entry is hidden until we have more content.

Preview URL: https://deploy-preview-7301--devel-docs-dagger-io.netlify.app/manuals/developer/cookbook
Preview URL for local testing: /manuals/developer/cookbook


docs/sidebars.ts
```diff
@@ -567,6 +567,11 @@ module.exports = {
             }
           ]
         },
+//        {
```
vikram-dagger, line 570: I recommend we keep this hidden until we have more content in the cookbook.

gerhard (approved):
LGTM! OK to merge as soon as all checks pass. 

f you want to +1 existing issues @vikram-dagger, that might help surface how many times we are hitting known issues in our CI. E.g.:
- https://github.com/dagger/dagger/issues/7303
- https://github.com/dagger/dagger/issues/7192

docs/sidebars.ts
kpenfound: Do you want to merge this PR and we can contribute sections in separate PRs and then uncomment this in a follow up PR after?
cc @levlaz

vikram-dagger: > Do you want to merge this PR and we can contribute sections in separate PRs and then uncomment this in a follow up PR after? cc @levlaz

Yes, exactly

docs/current_docs/manuals/developer/cookbook.mdx
```diff
@@ -0,0 +1,172 @@
...
+</TabItem>
+<TabItem value="TypeScript">
+
+```typescript file=./cookbook/snippets/write-dir/typescript/index.ts
+```
+
+</TabItem>
+</Tabs>
+
+Example usage:
```
vikram-dagger, line 37: This is a change from the previous cookbook. WDYT of it?

kpenfound: I don't think it's necessary, but I'm not strongly against it either. It may not make sense for certain examples (which is fine) and may lead us to trying to show cli-oriented examples in cookbook (which is avoidable)

vikram-dagger: Your thoughts @helderco @levlaz @shykes ?

levlaz: I kind of like it, but I think we should think about adding two parts here. 

1. Example SDK usage (maybe even show how to use in other SDKs) 
2. Example CLI usage (or make it clear that is what is happening) 

I think this is especially important because there are some things that are different between CLI and SDK. For example, in go, optional args are not actually optional if you are using them via the SDK. 

Risk here is that it is possible this becomes too verbose.

vikram-dagger: AFAIK we are de-emphasizing the SDK use case, so IMHO we should not show that part for the moment.

kpenfound: I vote ship it and we can start the migration and make a decision as we go

Issue #7284, opened 2024-05-05
🐞 Latest TypeScript SDK throws `TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts"`


harleylang:
### What is the issue?

Since the latest changes were merged in on Friday I have found that TypeScript Dagger Modules and Functions will fail to run, with the following error thrown in the logs: `TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts"`

The best hypothesis I could drum up is that this error was introduced after this PR merged in on Friday:

- https://github.com/dagger/dagger/pull/7194

Here is the commit where I *think* the bug was introduced:

[https://github.com/dagger/dagger/commit/56a039a3ee96b4ab826b711d3e5843da8aadd8cb#diff-a8b5f7f08501c3951b9991d2450b8b0a68ac764af369185bbf7044d9654ac58c](https://github.com/dagger/dagger/commit/56a039a3ee96b4ab826b711d3e5843da8aadd8cb#diff-a8b5f7f08501c3951b9991d2450b8b0a68ac764af369185bbf7044d9654ac58c)

In the above PR the TypeScript SDK was changed. I *think* the error occurs due to how the SDK now uses paths to target module. But I am not 100% sure. Here is the change that is leading me to think this might be the case:

[https://github.com/dagger/dagger/commit/56a039a3ee96b4ab826b711d3e5843da8aadd8cb#diff-a8b5f7f08501c3951b9991d2450b8b0a68ac764af369185bbf7044d9654ac58cL78-R79](https://github.com/dagger/dagger/commit/56a039a3ee96b4ab826b711d3e5843da8aadd8cb#diff-a8b5f7f08501c3951b9991d2450b8b0a68ac764af369185bbf7044d9654ac58cL78-R79)

CC @jedevc

### Dagger version

v0.11.1 - but somehow my setup is using the latest upstream code and I'm not sure how.

### Steps to reproduce

I have provided an example repository that includes a way to simulate the bug both locally and in a GitHub action. The repository can be found here: [https://github.com/harleylang/dagger-11.2-bug](https://github.com/harleylang/dagger-11.2-bug)

Notice that in my repository I have targeted Dagger Engine v0.11.1 and v0.11.0 for the module itself and that worked fine Friday Morning. Though I note in the logged output that it appears the engine is using \~v0.11.2\~ the latest Dagger TypeScript SDK somehow.

### Log output

GitHub Action output:

[...]

PR #7302, opened 2024-05-07
ci: lint runtime modules

jedevc:
In addition to making sure that all the module runtimes are being properly linted, this also ensures that the `go.mod`s are kept up to date!

This is important, so that we don't see these `go.mod`s not reflecting the state of reality (see the report in https://github.com/dagger/dagger/issues/7284, which mis-identifies the source of the issue - the dependency in question was *always* being installed, but the `go.mod` wasn't accurately reflecting that).


gerhard (approved):
LGTM 👍

jedevc:
Test failure is a known panic, current discussion is in discord (private): https://discord.com/channels/707636530424053791/1234820699365969981/1237374164684247100

helderco:
This is one step closer to replace having to commit all the `dagger.gen.go` in docs with a function that finds all those directories and lints their modules concurrently.

PR #7306, opened 2024-05-07
docs: Add multi-language chaining page

vikram-dagger:


levlaz:
I know this is just an example of one, but do we plan to keep the existing order? 
Screenshot 2024-05-07 at 11 48 29 AM

vikram-dagger:
> I know this is just an example of one, but do we plan to keep the existing order? 

I think so, unless we decide to drop some items eg. if they are no longer valid, duplicated elsewhere, etc. 

Was there anything specific you wanted to change in the sequence?

levlaz (approved):
LGTM 🚢

levlaz:
No suggestions on what to change, just wanted to clarify the plan for the order. 

I think this may need to be a different issue but I think the example we show here may be a bit too simple and I don't really walk away feeling the power of chaining. I don't have a great suggestion at the moment but feels like we should come up with a more real world example of where this might be really powerful.

Issue #6947, opened 2024-03-26
Allow calling core functions from the CLI


shykes:
## Problem

There is no way to call core functions from the CLI

## Solution

Allow calling core functions from the CLI.

See https://daggerverse.dev/mod/github.com/shykes/core for a userland implementation.

shykes:
FYI @sipsma, this may be a duplicate, after a quick search I couldn't find a direct equivalent, but may not have found it.

helderco:
Would you use `-m core` to select it? If so, we need to reserve that name:
- Forbid creating a module with name "core"
- Forbid installing a dependency with name "core"

shykes:
> Would you use `-m core` to select it? If so, we need to reserve that name:
> 
> * Forbid creating a module with name "core"
> * Forbid installing a dependency with name "core"

Good question... any thoughts @sipsma @vito @aluzzardi @jedevc ?

jedevc:
Good point - I like the idea of needing to explicitly select it, so yes, we'd probably need to reserve this ASAP.

helderco:
I have a working POC:

<img width="878" alt="Screenshot 2024-05-07 at 17 08 30" src="https://github.com/dagger/dagger/assets/174525/11b9d4a6-2cdc-409a-ae27-d7cf73efe28e">

Just needs the name reservation and a bit of polish to reduce what's available:

```console
❯ dagger-dev -m core functions
Name                                   Description
blob                                   Retrieves a content-addressed blob.
builtin-container                      Retrieves a container builtin to the engine.
cache-volume                           Constructs a cache volume for a given cache key.
check-version-compatibility            Checks if the current Dagger Engine is compatible with an SDK's required version.
container                              Creates a scratch container.
current-function-call                  The FunctionCall context that the SDK caller is currently executing in.
current-module                         The module currently being served in the session, if any.
current-type-defs                      The TypeDef representations of the objects currently being served in the session.
default-platform                       The default platform of the engine.
directory                              Creates an empty directory.
file                                   -
function                               Creates a function.
generated-code                         Create a code generation result, given a directory containing the generated code.
git                                    Queries a Git repository.
host                                   Queries the host environment.
http                                   Returns a file containing an http remote url content.
load-cache-volume-from-id              Load a CacheVolume from its ID.
load-container-from-id                 Load a Container from its ID.
load-current-module-from-id            Load a CurrentModule from its ID.
load-directory-from-id                 Load a Directory from its ID.
[...]

PR #7310, opened 2024-05-07
cli: allow calling core functions directly

helderco:
Fixes https://github.com/dagger/dagger/issues/6947

Current implementation uses a top-level command for core functions:
- `dagger core <function>`  → execute a function from the core API
- `dagger call -m <module> <function>` → execute a function from a module

## How

Can be run from anywhere as it doesn't require a module to load:

```console
❯ dagger core version
[dev-fdc07a5bfb46ca6a37ae379d767210d54f141c4f](version: v0.12.5-010101000000-dev-8e3b1f29b72b)
```

Allows experimentation in the CLI without the need for a module

```console
❯ dagger core container from --address=alpine file --path=/etc/os-release contents
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.20.2
PRETTY_NAME="Alpine Linux v3.20"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```

Can be a nice alternative to `docker run`:

```console
❯ dagger core container from --address=cgr.dev/chainguard/wolfi-base terminal
dagger / $ 
```

Without arguments:

```console
❯  dagger core
_type: Client
defaultPlatform: linux/arm64
version: v0.12.5-010101000000-dev-8e3b1f29b72b
```

Listing:

```console
❯  dagger core --help
Call a core function

USAGE
  dagger core [options] <function>

FUNCTIONS
  blob                Retrieves a content-addressed blob.
  container           Creates a scratch container.
  current-type-defs   The TypeDef representations of the objects currently being served in the session.
  dagger-engine       The Dagger engine container configuration and state
  default-platform    The default platform of the engine.
  directory           Creates an empty directory.
  git                 Queries a Git repository.
  host                Queries the host environment.
  http                Returns a file containing an http remote url content.
  module              Create a new module.
  module-dependency   Create a new module dependency configuration from a module source and name
  module-source       Create a new module source instance from a source ref string.
  version             Get the current Dagger Engine version.

OPTIONS
  -j, --json            Present result as JSON
  -o, --output string   Save the result to a local file or directory

INHERITED OPTIONS
  -d, --debug             Show debug logs and full verbosity
  -i, --interactive       Spawn a terminal on container exec failure
      --progress string   Progress output format (auto, plain, tty) (default "auto")
  -q, --quiet count       Reduce verbosity (show progress, but clean up at the end)
  -s, --silent            Do not show progress at all
  -v, --verbose count     Increase verbosity (use -vv or -vvv for more)
  -w, --web               Open trace URL in a web browser

Use "dagger core [command] --help" for more information about a command.
```

## Next

The `core` top-level command is possibly a stepping stone to select a few core functions as their own top-level commands. That'll open up creating more commands by implementing them via the API rather than client side. \cc @sipsma

### DIFF
--- a/.changes/unreleased/Added-20240629-000342.yaml
+++ b/.changes/unreleased/Added-20240629-000342.yaml
@@ -5,9 +5,13 @@ body: |
   Usage: `dagger core <function>`
   Example: `dagger core container from --address=alpine terminal`
 
-  Works the same as `dagger call`, but instead of loading a user module, it only uses functions in the core API.
+  Works the same as `dagger call`, but instead of loading a user module, 
+  it only uses functions from the core API.
   
   Run `dagger core --help` for available functions.
+
+  _Note that this command is experimental and the DX for calling core functions 
+  in the CLI may change in the future._
 
 time: 2024-06-29T00:03:42.175445Z
 custom:
--- a/.changes/unreleased/Added-20240629-000342.yaml
+++ b/.changes/unreleased/Added-20240629-000342.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: 'cli: allow calling core functions directly'
+time: 2024-06-29T00:03:42.175445Z
+custom:
+    Author: helderco
+    PR: "7310"
--- a/.changes/unreleased/Added-20240629-000342.yaml
+++ b/.changes/unreleased/Added-20240629-000342.yaml
@@ -1,5 +1,14 @@
 kind: Added
-body: 'cli: allow calling core functions directly'
+body: |
+  cli: allow calling core functions directly
+
+  Usage: `dagger core <function>`
+  Example: `dagger core container from --address=alpine terminal`
+
+  Works the same as `dagger call`, but instead of loading a user module, it only uses functions in the core API.
+  
+  Run `dagger core --help` for available functions.
+
 time: 2024-06-29T00:03:42.175445Z
 custom:
     Author: helderco
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -36,6 +36,9 @@ var (
 	moduleURL   string
 	moduleFlags = pflag.NewFlagSet("module", pflag.ContinueOnError)
 
+	loadCoreFuncs bool
+	functionFlags = pflag.NewFlagSet("functions", pflag.ContinueOnError)
+
 	sdk           string
 	licenseID     string
 	compatVersion string
@@ -60,9 +63,16 @@ const (
 func init() {
 	moduleFlags.StringVarP(&moduleURL, "mod", "m", "", "Path to the module directory. Either local path or a remote git repo")
 
+	functionFlags.BoolVarP(&loadCoreFuncs, "core", "c", false, "Load core functions instead of a module (mutually exclusive with -m)")
+
+	for _, c := range funcCmds.All() {
+		c.PersistentFlags().AddFlagSet(moduleFlags)
+		c.PersistentFlags().AddFlagSet(functionFlags)
+		c.MarkFlagsMutuallyExclusive("core", "mod")
+	}
+
 	listenCmd.PersistentFlags().AddFlagSet(moduleFlags)
 	queryCmd.PersistentFlags().AddFlagSet(moduleFlags)
-	funcCmds.AddFlagSet(moduleFlags)
 	configCmd.PersistentFlags().AddFlagSet(moduleFlags)
 
 	moduleInitCmd.Flags().StringVar(&sdk, "sdk", "", "Optionally install a Dagger SDK")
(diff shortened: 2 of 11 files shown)


helderco:
Update: changed implementation from `-m` flag to new `-c,--core` flag. PR this was based on was merged so taking out of draft.

helderco:
Reimplemented as `dagger core`.

cmd/dagger/module.go
```diff
@@ -1093,6 +1108,30 @@ func (o *modObject) IsCore() bool {
...
+			"generatedCode",
+			"currentFunctionCall",
+			"currentModule",
+			"typeDef",
+			// for tests only
+			"secret",
+			// deprecated
+			"pipeline",
+		},
+	}
```
helderco, line 1128: These are the hardcoded exclusions when you do `dagger core --help`. May want to add more to the list. To be clear, this means these commands won't be created. It's not just for listing in `--help`.

Example of the output with current list in this PR's description.

helderco: I think these should be excluded as well:
- set-secret
- cache-volume

I can see a use case for the remaining ones even if just to debug, introspect, or experiment/prototype.

shykes (approved):
We've been iterating for a while, I think we should merge and make adjustments as needed.

I also like `call --core`, but after discussing with @helderco, `dagger core` has a few advantages.

sdk/go/querybuilder/querybuilder.go
jedevc, line 1: Trying to understand the querybuilder changes - is this just a refactor/interface change, or is there a behavior change? If the latter, do we need a new test?

docs/static/api/reference/index.html
jedevc, line 1: We need https://github.com/dagger/dagger/pull/8118 to avoid these random changes :cry:

.changes/unreleased/Added-20240629-000342.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Added
+body: 'cli: allow calling core functions directly'
```
jedevc, line 2: :pray: could we get another line here? A small example, or even just a redirection to run `dagger core --help`.

jedevc (approved):
There's a *lot* of cli changes, so hopefully okay to not review in lots of details, but I like the changes that simplify the `FuncCommand` fields, those were tricky before.

Left some small nits, but LGTM.

sdk/go/querybuilder/querybuilder.go
helderco: It's a refactor for what was introduced in the printing object fields PR. Only place where it's used (when dagger call returns an object in the CLI).

.changes/unreleased/Added-20240629-000342.yaml
helderco: Added in https://github.com/dagger/dagger/pull/7310/commits/1d36e77d177a6c7d4614feb490dddd7c2f859ea2.

helderco:
> I also like `call --core`, but after discussing with @helderco, `dagger core` has a few advantages.

We can keep the refactoring that allows the stepping stone to `dagger container`. It's a very small change to turn this back into the `call --core` flag. The inverse wasn't true but that work is done and not lost. It would take around 5mins to switch to the flag.

Basically add the mutually exclusive flag (code already in a previous commit in this PR), and set `fc.DisableModuleLoad = true` dynamically if `--core` is set, instead of statically here:

https://github.com/dagger/dagger/blob/1d36e77d177a6c7d4614feb490dddd7c2f859ea2/cmd/dagger/call.go#L20-L29

Perhaps at least mark `dagger core` as _experimental_ which shows a message in `dagger core --help` (the feature is still available, even though we're not using it in any command atm). That would signal to users that it could later change to something else, whether it's `dagger call -c container` or `dagger container`.

jedevc:
> Perhaps at least mark `dagger core` as _experimental_ which shows a message in `dagger core --help` (the feature is still available, even though we're not using it in any command atm)

I think this is worth doing regardless.

I'd also be happier with a `--core` flag - but at this point, we've been discussing for so longer, I really really would just rather ship something in the next release. As you mention, if it's experimental, we can adjust with feedback later :tada:

I've added this to the next milestone, so we'll definitely get this for the next release - at least then we've got an upper limit on how long we're allowed to keep discussing this for :laughing:

sdk/go/querybuilder/querybuilder.go
jedevc: Thanks for the explanation, much appreciated :pray:

helderco:
I added the "experimental" annotation. This way it's justifiable if we decide to change the DX. Just waiting for the tests to pass before merge.

PR #7312, opened 2024-05-07
docs: Add multi-language secrets page

kpenfound:


levlaz:
This looks great to me, my only question is around the order that it shows up in the sidebar. Are we going to preserve what was going on inside each language section?

vikram-dagger (approved):

PR #7315, opened 2024-05-08
Session Cleanup+Simplifications

sipsma:
Another part of the effort to support https://github.com/dagger/dagger/pull/6916 while also doing general cruft cleanup and setup for various upcoming efforts.

This changeset focuses on making sessions + associated state management simpler:
1. More comprehensible+centralized state management
   * Rather than spread all over the place and tied together in random places, all of the state associated with a given session is in a `daggerSession` object and all of the state associated with a given client in a session is a `daggerClient` object
   * The code is also a lot more structured and "boring" in terms of locking/mutating state/etc. Not a rube goldberg machine anymore
   * The whole "pre-register a nested client's state before it calls", which was a fountain of confusion and bugs, is gone.
      * e.g. a bug was reported recently with use of terminal and nested sessions that was caused by this registration, but this PR had already accidentally fixed it, so there's just [a commit with test coverage](https://github.com/dagger/dagger/pull/7315/commits/3808d39d4d7fd4164c0173e76f65a2d505a3bf4e) here
2. No more insane gRPC tunneling, the engine API is just an HTTP server now
   * graphQL http requests are just that, don't have to tunnel them through gRPC streams
   * session attachables are still gRPC based, but over a hijacked http conn (as opposed to a gRPC stream embedded in another gRPC stream)
   * This allowed us to move off the `Session` method from buildkit's upstream controller interface
   * That in turn let us delete huge chunks of complicated code around handing conns (i.e. `engine/server/conn.go`) and no longer need to be paranoid about gRPC max message limits in as many places
   * This also allowed us to enable connection re-use for requests from the engine client to the engine server
3. The overall engine-wide state (mostly various buildkit+containerd entities) is also centralized now rather than spread confusingly amongst many files, which is slightly tangential but supported the above efforts.

---

### Objects + state + naming
* [`Server`](https://github.com/dagger/dagger/blob/22deba615a453987f70f0f90c9e97f19d0cf18a2/engine/server/server.go#L89) - formerly known as `BuildkitController`
   * This is not to be confused with the thing previously called `Server`, which was really more like the session state (and was thus confusing)
   * All the "global" state for various buildkit+containerd entities like snapshotters, various cache dbs, the solver, worker+executor, etc. Also top level state for which sessions currently exist
      * There's _a lot_ in there, but I personally much prefer it in one place rather than spread all over.
   * Serves an HTTP API for gql queries, session attachables and shutdown, with requests scoped to the client based on `clientMetadata` (which is now sent in an http header). [Code](https://github.com/dagger/dagger/blob/22deba615a453987f70f0f90c9e97f19d0cf18a2/engine/server/session.go#L629-L632).
   * Still implements the `BuildkitController` API since we do have some reliance on `ListWorker` at least, though we are free to change any+all of those to core API (i.e. gql) calls at any point (just got a bit too out of scope here)
* [`daggerSession`](https://github.com/dagger/dagger/blob/22deba615a453987f70f0f90c9e97f19d0cf18a2/engine/server/session.go#L47) and [`daggerClient`](https://github.com/dagger/dagger/blob/22deba615a453987f70f0f90c9e97f19d0cf18a2/engine/server/session.go#L88)
   * Basically what it says on the tin: the session-wide state for each session and the client-specific state for each client in a session
   * Does state tracking with an enum of possible states like `uninitialized`, `initialized`, `deleted` (not complicated enough to go full-on state machine, but this still makes it all more obvious and easy to follow I think, especially when it comes to locking the state for mutations)
   * I moved all the state that used to co-exist in `core.Query` and `buildkit.Client` to be in these structs too so there's less places to look+think about
   * One notable thing gone is `ClientCallContext` - instead of trying to register all of that we're "stateless" in that the module+function-call metadata for a client is just plumbed through `ExecutionMetadata`+`ClientMetadata`, following the request path rather than both that and a pre-registration side-channel
      * e.g. [here's](https://github.com/dagger/dagger/blob/22deba615a453987f70f0f90c9e97f19d0cf18a2/engine/buildkit/executor_spec.go#L980) where the executor supplies the `ClientMetadata` it was plumbed in the requests made by the nested client
   * The logic for deciding when to end a session is now just "when the main client caller has no more active connections", at which point the session state is torn down and released. This is done by just incrementing and decrementing that count at the beginning/end of each http request

### HTTP/2 Usage
* I updated all the HTTP servers we create to explicitly support HTTP/2 (with h2c, aka no TLS), while also supporting HTTP/1 clients too
* The main motivation here was:
   * We wanted to get rid of the gRPC tunneling (for simplicity, performance, detaching from the `BuildkitController.Session` API and it's associated complications, etc.)
   * But that meant that every time the http client needed to add to the connection pool it would have to invoke the connhelper (rather than open a new gRPC stream) which is an expensive operation for e.g. `docker-container`, `kube-pod`, etc. (spawns a subprocess)
   * HTTP/2 solves that problem though via stream multiplexing; go's http/2 client by default only needs a pool of 2 conns (one for reqs, one for resps) and can just multiplex everything from there
     * I briefly looked at HTTP/3 since just sending udp packets back and forth would be even simpler conceptually, but it's still too immature+low-level in the go ecosystem
* This seems to have worked pretty seamlessly, other than one gotcha I hit where typescript tests *using node 18 only* were erroring out ([fix with details here](https://github.com/dagger/dagger/pull/7315/commits/22deba615a453987f70f0f90c9e97f19d0cf18a2))
* There is also still a need to serve some gRPC APIs for the few remaining buildkit controller APIs we use and OTel, which is done via [gRPC http handlers](https://github.com/sipsma/dagger/blob/22deba615a453987f70f0f90c9e97f19d0cf18a2/cmd/engine/main.go#L407-L407)
   * The docs on that suggest there are some missing advanced gRPC features (e.g. BFD, big frame detection, which is a performance optimization) but none of them have been obviously relevant to our use case. Utterly worst case scenario, there is a fallback option of serving http + grpc on separate listeners, but avoiding that complication unless proven 100% necessary
   * These can also be migrated to pure graphql/plain-http APIs as desired

### Session Attachables
* As mentioned above, session attachables no longer require 2 layers of gRPC tunnels; instead there's just a `/sessionAttachables` http endpoint, which the server hijacks and uses as a raw conn for establishing the gRPC streams
* That "hijack and invert client/server relationship" process involves a small dance in order to be robust against accidentally mixing/overlapping http+gRPC traffic, which can, unsurprisingly, confuse the computer
   * [Client-side](https://github.com/dagger/dagger/blob/22deba615a453987f70f0f90c9e97f19d0cf18a2/engine/client/client.go#L414-L448) and [server-side](https://github.com/dagger/dagger/blob/22deba615a453987f70f0f90c9e97f19d0cf18a2/engine/server/session.go#L662-L697) implementation, with comments explaining. Basically just http req/resp + a 1 byte ack to synchronize the switch to gRPC
   * I wanted to use upstream buildkit's builtin `SessionManager.HandleHTTP` method, which is somewhat similar, but it didn't handle the switch from http->grpc synchronously and was resulting in data getting mixed sometimes
* A nice side effect of this in combination with the session state simplifications is that we no longer need to do the whole "retry making a request to verify the session is working"
   * Instead, if we successfully connect these session attachables, we can know the session as a whole was successfully initialized and we can unblock client connect, returning to the caller
   * That had more nice side effects of reducing possibilities of race conditions when requesting the caller session in various server-side APIs


sipsma:
Ended up spinning out the support for serving nested execs from the executor [here](https://github.com/dagger/dagger/pull/7367) (ended up just becoming removal of the shim entirely). Coming back here now to finish up the rest of this refactor on top of that.

cmd/engine/main.go
```diff
@@ -303,36 +348,32 @@ func main() { //nolint:gocyclo
...
-		if cfg.Trace {
-			go logTraceMetrics(context.Background())
+		var traceSocket string
+		if pubsub != nil {
+			bklog.G(ctx).Debug("starting otel controller")
+			traceSocket = filepath.Join(cfg.Root, "otel-grpc.sock")
+			if err := runOtelController(traceSocket, pubsub); err != nil {
+				logrus.Warnf("failed set up otel-grpc controller: %v", err)
+				traceSocket = ""
+			}
```
sipsma, line 376: reminder to self: double check this is rm'd after rebase on https://github.com/dagger/dagger/pull/7452#pullrequestreview-2100372653

analytics/analytics.go
```diff
@@ -30,7 +30,8 @@ type Event struct {
 	Properties map[string]string `json:"properties,omitempty"`
 
 	DeviceID string `json:"device_id,omitempty"`
-	ServerID string `json:"server_id,omitempty"`
+	// TODO: is it safe to update the json name or do we need cloud coordination?
+	SessionID string `json:"server_id,omitempty"`
```
sipsma, line 34: This is a question for @vito @aluzzardi, not sure if we can just change the name of this json field and cloud will be fine or if it's more involved.

It doesn't *actually* matter that much, mostly aesthetics, but I guess discrepancies like this will add up in confusion in the long term.

sipsma:
Hit a test failure flake I haven't seen previously:
```
services_test.go:478: 
        	Error Trace:	/app/core/integration/services_test.go:478
        	Error:      	"input: container.from.withServiceBinding.withExec.sync resolve: start d8dj7hiquom5i (aliased as www): health check errored: checking for port 8080/tcp: namespace for kjvnbk4mnlt2v514715uk730l not found in running state\n" does not contain "start d8dj7hiquom5i (aliased as www): exited:"
```

Will presume it's this PR's fault until proven otherwise, may be missing some synchronization of services somewhere? Or could be the service exiting prematurely for unrelated reasons?

sipsma:
@vito FYI I think I may have hit the theoretical flake you described in the OTEL PR:
```
1534

    client_test.go:104: 

1535

        	Error Trace:	/app/core/integration/client_test.go:104

1536

        	Error:      	Not equal: 

1537

        	            	expected: 1

1538

        	            	actual  : 0

1539

1540

        	Test:       	TestClientMultiSameTrace
```
(tangential feature request - easier to copy logs from the cloud traces output 😄)

Does that look like it may be the flake you were imagining? This PR obviously changes all kinds of timing of almost everything, so not sure if it's jsu tthat or a legit issue.

vito:
> Does that look like it may be the flake you were imagining? This PR obviously changes all kinds of timing of almost everything, so not sure if it's jsu tthat or a legit issue.

Yep, that's the one. Sorry about that! I have an idea of how to fix it but it's a bit tricky. The problem is that trace and log data arrives independently, and we can't know whether a span has logs until we see logs for it for the first time, after which point we wait until EOF. But the test can still flake if we don't see the start of the logs before calling `Close()`. There's a `echo hey; sleep 0.5` to try to counteract that, I suppose we could bump that `sleep`, but it might still flake under load.

In terms of an actual fix, I'm thinking we would need to set an attribute on the span to indicate that logs or at least an EOF should be consumed for it, and update the draining logic accordingly. But the problem is we don't control the span creation. We can man-in-the-middle it, and look for spans starting with `exec ` I suppose, simillar to how we already man-in-the-middle the `[internal] ` prefix into an attribute.

sipsma:
> Hit a test failure flake I haven't seen previously:
> 
> ```
> services_test.go:478: 
>         	Error Trace:	/app/core/integration/services_test.go:478
>         	Error:      	"input: container.from.withServiceBinding.withExec.sync resolve: start d8dj7hiquom5i (aliased as www): health check errored: checking for port 8080/tcp: namespace for kjvnbk4mnlt2v514715uk730l not found in running state\n" does not contain "start d8dj7hiquom5i (aliased as www): exited:"
> ```
> 
> Will presume it's this PR's fault until proven otherwise, may be missing some synchronization of services somewhere? Or could be the service exiting prematurely for unrelated reasons?

Can repro this one locally, think I know what's happening:
* Service `startContainer` runs health checks and container start in parallel
* The container exits
* But the select on healthcheck+wait-for-exit happens to hit a health check error first https://github.com/sipsma/dagger/blob/4b96edabf78e0ae10d5a7b4425e5bd0afa207baf/core/service.go#L447-L463
* It thus returns the health check error instead of the the container exiting error

cc @vito to double check this adds up to you too

If that's correct, I *think* this is technically independent of this PR and more likely just getting triggered now due to timing differences introduced here. But need to double check and will attempt a fix in the interest of not increasing probability of flakes here.

sipsma:
On a positive note, with the extra telemetry draining fix commits appended to this PR, I'm seeing full engine tests take as low as 9 minutes, which is the fastest I've seen in a very long time 🎉

sipsma:
> > Hit a test failure flake I haven't seen previously:
> > ```
> > services_test.go:478: 
> >         	Error Trace:	/app/core/integration/services_test.go:478
> >         	Error:      	"input: container.from.withServiceBinding.withExec.sync resolve: start d8dj7hiquom5i (aliased as www): health check errored: checking for port 8080/tcp: namespace for kjvnbk4mnlt2v514715uk730l not found in running state\n" does not contain "start d8dj7hiquom5i (aliased as www): exited:"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Will presume it's this PR's fault until proven otherwise, may be missing some synchronization of services somewhere? Or could be the service exiting prematurely for unrelated reasons?
> 
> Can repro this one locally, think I know what's happening:
> 
> * Service `startContainer` runs health checks and container start in parallel
> * The container exits
> * But the select on healthcheck+wait-for-exit happens to hit a health check error first https://github.com/sipsma/dagger/blob/4b96edabf78e0ae10d5a7b4425e5bd0afa207baf/core/service.go#L447-L463
> * It thus returns the health check error instead of the the container exiting error
> 
> If that's correct, I _think_ this is technically independent of this PR and more likely just getting triggered now due to timing differences introduced here. But need to double check and will attempt a fix in the interest of not increasing probability of flakes here.

Found and fixed that flake here https://github.com/dagger/dagger/pull/7610

analytics/analytics.go
vito: Looking at the server-side analytics code, I think you're good to rename this. There's a struct with this field defined, but nothing references it after unmarshaling. I can put up a quick PR to change it.

engine/client/client.go
```diff
@@ -952,25 +997,70 @@ func allCacheConfigsFromEnv() (cacheImportConfigs []*controlapi.CacheOptionsEntr
...
+		UpstreamCacheExportConfig: c.upstreamCacheExportOptions,
+		Labels:                    c.labels,
+		CloudToken:                os.Getenv("DAGGER_CLOUD_TOKEN"),
+		DoNotTrack:                analytics.DoNotTrack(),
+	}
+}
+
+func (c *Client) AppendHTTPRequestHeaders(headers http.Header) http.Header {
+	headers = c.clientMetadata().AppendToHTTPHeaders(headers)
+	propagation.TraceContext{}.Inject(c.internalCtx, propagation.HeaderCarrier(headers))
```
vito, line 1017: This should probably be `otel.GetTextMapPropagator()` so it'll propagate baggage as well (which is configured globally). Just recently started using baggage in #7605 for propagating things like `--debug` and `NO_COLOR=1` through to server-side `slog` usage, might build on it in the future for analytics.

sipsma: Fixed here https://github.com/dagger/dagger/pull/7315/commits/b8ab108ccc61b2ebb97b3558a15228ea0c03c861 thanks again!

engine/buildkit/filesync.go
```diff
@@ -111,7 +111,7 @@ func (c *Client) diffcopy(ctx context.Context, opts engine.LocalImportOpts, msg
 
 	ctx = opts.AppendToOutgoingContext(ctx)
 
-	clientCaller, err := c.GetSessionCaller(ctx, true)
+	clientCaller, err := c.GetSessionCaller(ctx, false)
```
vito, line 114: Curious about this `true` becoming `false` - is it somehow guaranteed that it'll be available immediately?

engine/client/client.go
```diff
@@ -325,112 +327,171 @@ func (c *Client) startEngine(ctx context.Context) (rerr error) {
...
-	c.httpClient = &http.Client{Transport: &http.Transport{
-		DialContext: c.DialContext,
-		// connection re-use in combination with the underlying grpc stream makes
-		// managing the lifetime of connections very confusing, so disabling for now
-		// TODO: For performance, it would be better to figure out a way to re-enable this
-		DisableKeepAlives: true,
-	}}
-
-	// there are fast retries server-side so we can start out with a large interval here
-	if err := retry(ctx, 3*time.Second, func(elapsed time.Duration, ctx context.Context) error {
```
vito, line 417: Nice seeing this retry logic going away. Guessing it's just not needed anymore, for reasons similar to the other comment?

engine/client/client.go
```diff
@@ -325,112 +327,171 @@ func (c *Client) startEngine(ctx context.Context) (rerr error) {
...
+		}()
+		c.sessionSrv.Run(ctx)
+		return nil
+	})
+
+	c.httpClient = c.newHTTPClient()
+	return nil
+}
+
+func ConnectBuildkitSession(
```
vito, line 390: neat!

engine/server/bk_session.go
vito, line 1: I'm always gonna read this as "burger king session" and I'm OK with that

engine/server/session.go
```diff
@@ -0,0 +1,953 @@
...
+	containers   map[bkgw.Container]struct{}
+	containersMu sync.Mutex
+
+	dagqlCache dagql.Cache
+}
+
+type daggerSessionState string
+
+const (
+	sessionStateUninitialized daggerSessionState = ""
```
vito, line 84: Any reason not to make this (and `clientStateUninitialized`) more explicitly `"uninitialized"` or something? Might be nice to have clarity on it vs. a value being dropped somewhere (assuming this ends up in logs)

engine/server/session.go
```diff
@@ -0,0 +1,953 @@
...
+		http.Error(w, fmt.Sprintf("incompatible client version: %s", err), http.StatusInternalServerError)
+		return
+	}
+
+	httpHandlerFunc(srv.serveHTTPToClient, &ClientInitOpts{
+		ClientMetadata: clientMetadata,
+	}).ServeHTTP(w, r)
+}
+
+// TODO: doc
```
vito, line 619: (just noting in case you wanted to get this in, nonblocking ofc)

engine/buildkit/cleanup.go
```diff
@@ -2,40 +2,66 @@ package buildkit
...
 	"errors"
+	"sync"
+	"time"
 
 	"github.com/dagger/dagger/engine/slog"
 )
 
-type cleanups struct {
-	funcs []cleanupFunc
+// TODO: move elsewhere
```
vito, line 11: (here too)

engine/buildkit/cleanup.go
```diff
@@ -2,40 +2,66 @@ package buildkit
...
+	c.funcs = append(c.funcs, fOnce)
+	return CleanupFunc{fOnce}
 }
 
-func (c *cleanups) addNoErr(msg string, f func()) {
-	c.add(msg, func() error {
-		f()
-		return nil
-	})
+// TODO: doc, needed in some cases
```
vito, line 36: (and here, spoiler I just Ctrl+F'd TODO:)

core/container_exec.go
```diff
@@ -94,44 +91,27 @@ func (container *Container) WithExec(ctx context.Context, opts ContainerExecOpts
...
 		// include the engine version so that these execs get invalidated if the engine/API change
 		runOpts = append(runOpts, llb.AddEnv(buildkit.DaggerEngineVersionEnv, engine.Version))
 
 		// include a digest of the current call so that we scope of the cache of the ExecOp to this call
 		runOpts = append(runOpts, llb.AddEnv(buildkit.DaggerCallDigestEnv, string(dagql.CurrentID(ctx).Digest())))
 
-		if !callerOpts.Cache {
-			// include the ServerID here so that we bust cache once-per-session
+		if execMD.CachePerSession {
+			// include the SessioID here so that we bust cache once-per-session
```
vito, line 109:
```suggestion
			// include the SessionID here so that we bust cache once-per-session
```

vito (approved):
Just nits/questions!

engine/buildkit/filesync.go
sipsma: Yep, previously the buildkit session attachables got connected in parallel with gql requests, but now the whole session is initialized by a synchronous request to connect the buildkit session attachables, so it's not possible to have a race here anymore.

I went ahead and changed these values because at one point I had a bug that caused the attachables to fail, but it manifested as these lines deadlocking. Fixed the bug obviously but now if something goes wrong in the future it will be an error rather than deadlock 🙂

engine/client/client.go
sipsma: Yep, once we've successfully made any request from a given client in a given session, we know that the client+session state are fully initialized. So because we have already made a request successfully at this point (to hook up the attachables) we know we're good to go and don't need this "retry a request" stuff anymore.

engine/server/bk_session.go
sipsma: 😆 I had the same thought while writing it, also OK with it

engine/server/session.go
sipsma: Good point, the fact that the zero value was uninitialized was mildly convenient but having it be more explicit is better I agree. Fortunately there is exactly one place session+client get initialized, so easy to just explicitly set there.

engine/server/session.go
sipsma: Doh on all accounts, fixed them all up

analytics/analytics.go
sipsma: Almost missed this, but awesome, just changed it to be session_id

PR #7328, opened 2024-05-08
core: make Directory.diff logic less fragile

sipsma:
The TS SDK was very subtly broken when used from a git ref due to the fact that the `Directory.diff` logic was way too fragile (has existed since the beginning of time). This fixes that logic and updates the TS SDK to be slightly simpler now that it doesn't need to try to avoid that fragility. 

Individual commits have more detailed messages.

---

We also need to add integ tests that use SDKs other than go sourced from git refs. Starting the effort now, will send out a separate PR if this one is merged by the time I'm done.

### DIFF
--- a/core/directory.go
+++ b/core/directory.go
@@ -6,7 +6,6 @@ import (
 	"io/fs"
 	"path"
 	"path/filepath"
-	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


core/directory.go
```diff
@@ -647,14 +646,17 @@ func (dir *Directory) WithNewDirectory(ctx context.Context, dest string, permiss
 func (dir *Directory) Diff(ctx context.Context, other *Directory) (*Directory, error) {
 	dir = dir.Clone()
 
-	if dir.Dir != other.Dir {
-		// TODO(vito): work around with llb.Copy shenanigans?
-		return nil, fmt.Errorf("TODO: cannot diff with different relative paths: %q != %q", dir.Dir, other.Dir)
+	thisDirPath := dir.Dir
+	if thisDirPath == "" {
+		thisDirPath = "/"
```
jedevc, line 651: Potentially we could use `path.Clean`?

jedevc (approved):

core/directory.go
sipsma: https://go.dev/play/p/MZpwcipFm03 I think that would result in comparing "." and "/".

Honestly it would be nice to just make all of this consistent from the start so that `Directory.Dir` is always a consistent value once set, but I'm trying to be minimal here since we want to release this ASAP.

vito (approved):
Thanks for tracking this down!

gerhard:
What a legend @sipsma 🦄

PR #7330, opened 2024-05-08
Migrate Publish Engine & CLI to the new CI

matipan:
We probably want Publish Engine & CLI to re-use _hack_make. The main benefit of doing that is that we control the dagger version being used in a single place and trigger all jobs with the same logic. However, we can to unblock jobs from main quickly since we deleted the legacy CI and now no jobs can be execute.


sipsma (approved):
🙏

PR #7332, opened 2024-05-08
docs: make go snippet compilable

sipsma:
This was causing errors on main and in PRs: https://github.com/dagger/dagger/actions/runs/9008181567/job/24749649865?pr=7330#step:5:502

Couldn't have caught it because the docs PRs (correctly) don't trigger engine tests, but I think the way the engine tests run result in an error if there's a compilation problem on any Go packages in the whole repo, which includes the docs since we have all those go examples.
* There's definitely some room for improvement here, but want to fix this to unblock everything right now.


kpenfound (approved):
🚢

Issue #7334, opened 2024-05-08
🐞 Setting the same secret in parallel causes Dagger to hang


sagikazarmark:
### What is the issue?

```
Failed to connect; retrying... make request: Post "http://dagger/query": rpc error: code = Unknown desc = failed to verify client: client ID "***" registered with different secret token
```

https://github.com/sagikazarmark/daggerverse/pull/96/commits/1090eee44f94be6aa09e20d28301af0e8d582261

### Dagger version

dagger v0.11.3 (registry.dagger.io/engine) darwin/arm64

PR #7335, opened 2024-05-09
engine: make client IDs fully random again

sipsma:
Fixes https://github.com/dagger/dagger/issues/7334

A few releases ago, client IDs were changed to be digests of the dagql.CurrentID in order to remove the number fields in our ftp_proxy hack and to consolidate client ID w/ module caller digest.

This is not entirely working as expected though (detailed below). Fortunately the new custom worker gives us a non-hacky path to set whatever IDs we want in containers without worrying about cache busting or the various problems caused by the ftp_proxy hack, which has been gone entirely as of last release.

This updates the engine to make client IDs random strings again.

The problem we're hitting is that when there are inputs to a Function of type Secret, the Functions seem to always run without any caching no matter what, even in the same dagger server. And even though before this client IDs would be the same, each client still generated it's own random secret token and thus conflicted with each other. Making client IDs random again fixes it because the secret tokens no longer mismatch.

However, the fact that we are not getting cache hits when invoking the same function repeatedly when there is an input of type Secret is still a mystery at this time. I confirmed that we do get the expected caching for other input types (string, Container, etc.), it's somehow only happening for Secret inputs.

I'm going to open a separate issue with more details for figuring out what's happening there.

### DIFF
--- a/engine/server/server.go
+++ b/engine/server/server.go
@@ -407,13 +407,6 @@ func (s *DaggerServer) VerifyClient(clientID, secretToken string) error {
 	return nil
 }
 
-func (s *DaggerServer) UnregisterClient(clientID string) error {
-	s.clientIDMu.Lock()
-	defer s.clientIDMu.Unlock()
-	delete(s.clientIDToSecretToken, clientID)
-	return nil
-}
-
 func (s *DaggerServer) ClientCallContext(clientID string) (*core.ClientCallContext, bool) {
 	s.clientCallMu.RLock()
 	defer s.clientCallMu.RUnlock()
--- a/.changes/unreleased/Fixed-20240509-114653.yaml
+++ b/.changes/unreleased/Fixed-20240509-114653.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Avoid hang caused by client id conflicts
+time: 2024-05-09T11:46:53.847848236+01:00
+custom:
+  Author: sipsma
+  PR: "7335"
--- a/core/container.go
+++ b/core/container.go
@@ -1052,6 +1052,9 @@ func (container *Container) WithExec(ctx context.Context, opts ContainerExecOpts
 		// include the engine version so that these execs get invalidated if the engine/API change
 		runOpts = append(runOpts, llb.AddEnv("_DAGGER_ENGINE_VERSION", engine.Version))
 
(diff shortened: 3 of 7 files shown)


sipsma:
Figured out that separate cache bug mentioned in the commit message: https://github.com/dagger/dagger/pull/7336

sipsma:
Looking into the test failures, they all appear to be like:
```
2024-05-09T00:49:01.8445964Z [95m23: [0m        	            	Error: failed to generate code: input: moduleSource.withContextDirectory.withName.withSDK.withSourceSubpath.asModule resolve: failed to create module: select: failed to update codegen and runtime: failed to get module runtime: failed to call sdk module moduleRuntime: select: IDFor: GeneratedCode has[0m
2024-05-09T00:49:01.8448008Z [95m23: [0m        	            	 no such field: "withWorkdir"
```

Which is a bit odd to say the least...

sipsma:
Fixed that weird issue here: https://github.com/dagger/dagger/pull/7335/commits/3c53f14c7ce3696a2636c5ced00178cdc64e74f4

engine/server/server.go
```diff
@@ -407,13 +407,6 @@ func (s *DaggerServer) VerifyClient(clientID, secretToken string) error {
 	return nil
 }
 
-func (s *DaggerServer) UnregisterClient(clientID string) error {
-	s.clientIDMu.Lock()
-	defer s.clientIDMu.Unlock()
-	delete(s.clientIDToSecretToken, clientID)
-	return nil
-}
-
```
jedevc, line 416: :eyes: what's the logic for the removal of this? Doesn't this cause the map to grow forever?

I'm *guessing* this was used when debugging https://github.com/dagger/dagger/issues/7334? Disabling still allows the tests to pass.

gerhard (approved):
Excited to see this fix come together! Love the description & the clean implementation.

Leaving the release notes & merge to you @jedevc 💪

jedevc (approved):

engine/server/server.go
sipsma: Oh sorry I should have mentioned this in the commit message. There's context for why this `UnregisterClient` method was added recently here: https://github.com/dagger/dagger/pull/7213/commits/df0753150527377c2fb9f8f0f996e08ce1fd2b61

Basically, going back to random IDs invalidates all of those considerations so we can just remove this once again.

PR #7336, opened 2024-05-09
core: don't bust cache with random OTEL values in function calls

sipsma:
When OTEL support was added, module functions were updated to set env vars with the OTEL traceparent values. These values appear to be random every time a function is invoked, even if coming from the same caller and making the call with the same inputs, etc.

That resulted in them busting the Buildkit cache every time.

Fortunately, most of the time the dagql cache within the context of a session prevented this from being noticeable, but in the case where dagql IDs were impure (which prevented dagql caching from being hit), then we'd run the Function every call without any cache hits at all.

The fix here just uses our new and improved mechanism for passing random IDs to containers without invalidating the cache, which was added for proxy env support but has been repeatedly paying dividends since.

I would not be suprised if this is a noticeable performance improvement, hopefully restoring what it was pre-v0.11 since this cache invalidation also prevented SDK codegen calls (which were supposed to be cached across all sessions) from being cached.

### DIFF
--- a/.changes/unreleased/Fixed-20240509-114745.yaml
+++ b/.changes/unreleased/Fixed-20240509-114745.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Avoid unneccessary module cache invalidation from internal plumbing values
+time: 2024-05-09T11:47:45.1739708+01:00
+custom:
+  Author: sipsma
+  PR: "7336"
--- a/core/container.go
+++ b/core/container.go
@@ -1036,6 +1036,7 @@ func (container *Container) WithExec(ctx context.Context, opts ContainerExecOpts
 
 	// this allows executed containers to communicate back to this API
 	var clientID string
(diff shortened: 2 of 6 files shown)


jedevc (approved):
Ah such a good find :heart: 

What's *so* annoying, is when we started getting reports of worse caching performance, I dove through this and noticed the trace parent stuff - but the thought process was "oh function calls currently aren't cached like this anyways, so it's fine" :cry: Of course - functions can be called multiple times in a session, not sure why I didn't follow through!

Nice catch :tada:

gerhard (commented):
This is such a big and important fix. I would call it fix++

Trying it out now locally.

jedevc:
:warning: There are conflicts with this and https://github.com/dagger/dagger/pull/7336, we should probably merge that one first.

gerhard:
Still running, will update when finished once, and then how the second run behaves:
image

PR #7340, opened 2024-05-09
docs: Create multi-language custom types page

vikram-dagger:


docs/current_docs/manuals/developer/custom-types.mdx
```diff
@@ -0,0 +1,66 @@
+---
+slug: /manuals/developer/custom-types
```
vikram-dagger, line 2: I think the code snippets may not be 1:1 equivalent here, so would like additional eyes. @helderco @kpenfound could you please check and suggest changes as needed to make them equivalent?

helderco: Only the Go snippet needs to match up.

helderco: The main difference is due to the way the URL fields are created:
- Python generates the URL in the objects' constructors (`Account` and `Organization`) which is the proper thing to do from an OOP POV.
- TypeScript does it only in `Account`, but for `Organization` and `GitRepository`  it's set in `daggerOrganization` (technically a _user_ of `Organization`).
- Go would need a separate function for both of those so it's creating `Organization.URL` and `GitRepository` within `DaggerOrganization` (same as TS); and Account is generating at runtime via a function, instead of setting the field.

For context, Python and TypeScript have natural object constructors. So it's natural to move the setting of URLs there. But Go doesn't have it. You use a function as constructor, typically named `New<struct name>()`. Other than the func name, Go is closer to GraphQL in this regard.

In this specific case, I don't think it's worth it to have different implementations due to what's natural in each language. We could find another example that doesn't rely on constructors as much (i.e., not having a field that depends on another field's value).

But I think it's ok to just mirror Go here, to make the simplest common implementations. In Dagger, one object's function is another object's constructor. That means users won't create these objects directly from their types, only the module's code does it. It's also simpler to create a function that returns the URL instead of setting it in the constructor, even though it's technically less performant.

helderco: I pushed a change reflecting the above.

docs/current_docs/manuals/developer/snippets/custom-types/typescript/src/index.ts
helderco, line 1: @TomChv, there's a problem running this:

```console
$ dagger call dagger-organization members username
Error: could not find result property username
```

Same thing with any field:
```console
$ dagger call dagger-organization url
Error: could not find result property username
```

It's coming from here:

https://github.com/dagger/dagger/blob/b5f0488771e0cf86f480128010a5cc40a4731af6/sdk/typescript/entrypoint/load.ts#L227-L229

helderco:
I moved the snippets into paths that match the default location on `dagger init`. This makes it easier to bootstrap a module to test. For example:
```console
cd snippets/custom-types/typescript
dagger init --sdk=typescript --source=. --name=github
dagger call dagger-organization members url
```

vikram-dagger:
> I moved the snippets into paths that match the default location on `dagger init`. This makes it easier to bootstrap a module to test. For example:
> 
> ```
> cd snippets/custom-types/typescript
> dagger init --sdk=typescript --source=. --name=github
> dagger call dagger-organization members url
> ```

I don't think we should make this change in the current PR, as this structure is not currently followed for other snippets and so it will create an inconsistency. I agree that we should do this in general, but I would prefer to do it globally for all snippets in a separate PR.

docs/current_docs/manuals/developer/snippets/custom-types/typescript/src/index.ts
TomChv: This will be fixed by https://github.com/dagger/dagger/pull/7381

helderco (approved):
Just needs a rebase

PR #7347, opened 2024-05-09
fix: attach cli args to initialize parent

jedevc:
Discovered during #7272, when trying to make sense of some of the output.

Before, the context was not being propagated, so when loaded, all CLI args would be associated with the root, instead of under the initialize span (where it makes a lot more sense).

Before:

	❯ dagger call -vv --source=. version string
	✔ connect 0.9s
	  ✔ starting session 0.2s
	  ┃ OK!
	✔ initialize 7.9s
	  ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.1s
	  ✔ ModuleSource.asModule: Module! 0.1s
	    ✔ Module.withSource(
		source: ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.0s
	      ): Module! 0.1s
	  ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.1s
	✔ ModuleSource.resolveFromCaller: ModuleSource! 0.1s
	  ✔ upload /home/jedevc/Documents/Projects/dagger from dragon (client id: rmi44eawc8e5b1mgzkqurv8rf) (exclude: **/.git) (include: **/vendor/, **/*.go, dagger.json, ci/**/*, **/go.mod, **/go.sum, **/go.work, **/go.work.sum) 0.1s
	✔ ModuleSource.resolveDirectoryFromCaller(path: "."): Directory! 7.1s
	  ✔ upload /home/jedevc/Documents/Projects/dagger from dragon (client id: rmi44eawc8e5b1mgzkqurv8rf) 1.5s
	  ✔ [internal] copy upload /home/jedevc/Documents/Projects/dagger from dragon (client id: rmi44eawc8e5b1mgzkqurv8rf) 1.1s
	✔ ModuleSource.resolveFromCaller: ModuleSource! 0.1s
	✔ dagger(
	    source: ✔ blob(digest: "sha256:d3c51e1d6dfbb342d5fb5e61198ae92b1b114b56baf1bbe2afc8111a5a10de68", mediaType: "application/vnd.oci.image.layer.v1.tar+zstd", size: 286554564, uncompressed: "sha256:d968cd1b0e2ce34160657f12a6d89beee17d0f236e7907e4fff81b9cac4fe917"): Directory! 0.0s
	  ): Dagger! 4.4s
	  ✔ connect 0.2s
	    ✔ starting session 0.2s
	    ┃ OK!
	✔ DaggerVersionInfo.string: String! 0.6s
	  ✔ connect 0.2s
	    ✔ starting session 0.2s
	    ┃ OK!

	dev-13f98b90bc7ef203bbee37cb88660567cb60df09

After:

	❯ dagger call -vv --source=. version string
	✔ connect 0.3s
	  ✔ starting session 0.2s
	  ┃ OK!
	✔ initialize 3.3s
	  ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.1s
	    ✔ upload /home/jedevc/Documents/Projects/dagger from dragon (client id: krj53xrrct4xjiiqejoxd7299) (exclude: **/.git) (include: **/*.go, dagger.json, ci/**/*, **/go.mod, **/go.sum, **/go.work, **/go.work.sum, **/vendor/) 0.1s
	  ✔ ModuleSource.asModule: Module! 0.1s
	    ✔ Module.withSource(
		source: ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.0s
	      ): Module! 0.1s
	  ✔ Module.initialize: Module! 0.7s
	    ✔ connect 0.1s
	      ✔ starting session 0.1s
	      ┃ OK!
	  ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.1s
	    ✔ upload /home/jedevc/Documents/Projects/dagger from dragon (client id: krj53xrrct4xjiiqejoxd7299) (exclude: **/.git) (include: **/vendor/, **/*.go, dagger.json, ci/**/*, **/go.mod, **/go.sum, **/go.work, **/go.work.sum) 0.3s
	  ✔ ModuleSource.resolveDirectoryFromCaller(path: "."): Directory! 1.6s
	    ✔ upload /home/jedevc/Documents/Projects/dagger from dragon (client id: krj53xrrct4xjiiqejoxd7299) 1.6s
	✔ dagger(
	    source: ✔ blob(digest: "sha256:890e28b5c1be8182fe65e1dad5adace9cdb0d2d74c152872e5cfb5027171a07e", mediaType: "application/vnd.oci.image.layer.v1.tar+zstd", size: 286554406, uncompressed: "sha256:2bea48213663a365ceab033bff95e7956ed0c19af65e29db732cbbc0c6fff519"): Directory! 0.0s
	  ): Dagger! 3.6s
	✔ DaggerVersionInfo.string: String! 0.7s
	  ✔ connect 0.2s
	    ✔ starting session 0.2s
	    ┃ OK!

	dev-d145a908790e319749ba1be917d028bde79b9043

This should also improve the view in dagger cloud :tada:

### DIFF
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -388,13 +388,13 @@ func (fc *FuncCommand) load(c *cobra.Command, a []string) (cmd *cobra.Command, _
 		if err := fc.addArgsForFunction(c, a, obj.Constructor); err != nil {
 			return nil, nil, err
 		}
-		fc.selectFunc(obj.Name, obj.Constructor, c, dag)
+		fc.selectFunc(ctx, obj.Name, obj.Constructor, c, dag)
 	} else {
 		fc.Select(obj.Name)
 	}
 
 	// Add main object's functions as subcommands
-	fc.addSubCommands(c, dag, obj)
+	fc.addSubCommands(ctx, c, dag, obj)
 
 	if fc.showHelp {
 		return c, nil, nil
@@ -437,11 +437,11 @@ func (fc *FuncCommand) traverse(c *cobra.Command) (*cobra.Command, []string, err
 	return fc.traverse(cmd)
 }
 
-func (fc *FuncCommand) addSubCommands(cmd *cobra.Command, dag *dagger.Client, fnProvider functionProvider) {
+func (fc *FuncCommand) addSubCommands(ctx context.Context, cmd *cobra.Command, dag *dagger.Client, fnProvider functionProvider) {
 	if fnProvider != nil {
 		cmd.AddGroup(funcGroup)
 		for _, fn := range fnProvider.GetFunctions() {
-			subCmd := fc.makeSubCmd(dag, fn)
+			subCmd := fc.makeSubCmd(ctx, dag, fn)
 			cmd.AddCommand(subCmd)
 		}
 
@@ -451,7 +451,7 @@ func (fc *FuncCommand) addSubCommands(cmd *cobra.Command, dag *dagger.Client, fn
 	}
 }
 
-func (fc *FuncCommand) makeSubCmd(dag *dagger.Client, fn *modFunction) *cobra.Command {
+func (fc *FuncCommand) makeSubCmd(ctx context.Context, dag *dagger.Client, fn *modFunction) *cobra.Command {
 	newCmd := &cobra.Command{
 		Use:                   cliName(fn.Name),
 		Short:                 strings.SplitN(fn.Description, "\n", 2)[0],
@@ -467,7 +467,7 @@ func (fc *FuncCommand) makeSubCmd(dag *dagger.Client, fn *modFunction) *cobra.Co
 			if fnProvider == nil && fn.ReturnType.AsList != nil {
 				fnProvider = fn.ReturnType.AsList.ElementTypeDef.AsFunctionProvider()
 			}
-			fc.addSubCommands(cmd, dag, fnProvider)
+			fc.addSubCommands(ctx, cmd, dag, fnProvider)
 
 			// Show help for first command that has the --help flag.
 			help, _ := cmd.Flags().GetBool("help")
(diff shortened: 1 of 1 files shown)


cmd/dagger/functions.go
helderco, line 1: You only need one line of code to solve this. Just set the command's context right after getting the new one from the span:

```diff
	ctx := c.Context()
	dag := fc.c.Dagger()

	ctx, span := Tracer().Start(ctx, "initialize", telemetry.Encapsulate())
	defer telemetry.End(span, func() error { return rerr })
+
+	c.SetContext(ctx)
```

jedevc: Doesn't this change the context for the entire cobra command though? I think this results in the context being used outside of `load`, which means more stuff will get attached to this context necessary.

(Will check in more detail later though)

helderco: You can SetContext to the previous one again on the defer at the end of the span to be sure.

jedevc: Hm, ok verifying - if we just have the `SetContext`, then this attaches *everything* to `initialize`. Essentially, the `ctx` set here in `load` gets propagated down to `traverse`, where it gets copied to every child command.

The `defer` approach doesn't work because `load` calls `traverse`. We can reset it right back before calling `traverse` (but the error handling in this case is a bit more tricky, since we'd want to reset it on error).

---

Additionally - I've also added another change that makes this change even harder - see `makeSubCmd` - we need access to *both* contexts. We need to build the query using the `initialize` context (otherwise calling `ID` and such without this can cause these to leak into the top-level *sometimes*), but the query should be executed using the command context.

helderco: Yeah, that's quite alright. 👍 Only if it were simpler. 🙂 

I think the `cmd.SetContext` was added to support users needing to pass a context from one Cobra function to another, like setting in `PreRunE` and using in `RunE` for example.

helderco (approved):
LGTM!

PR #7349, opened 2024-05-10
Start removing mage :tada:

jedevc:
This PR takes the first step to removing mage from our CI - unfortunately, to do the *whole* thing is still a little tricky right now.

There are several parts:
- The CI uses `dagger call` instead of `hack/make`.
- `hack/make` is **removed**! All CI jobs that still use it, now directly call `go run`.
- All unnecessary mage shim functions are removed. We keep:
	- `engine:dev`, used for `./hack/dev`. Eventually, this should be replaced by https://github.com/dagger/dagger/pull/7223.
	- `engine:publish`, used during github publishing. This connects the cli+engine publishing together, and ensures that the cli parts are all bumped. This is trickier to remove - we'd want to make our `_hack_make.yml` script an *action*, instead of a *workflow* - then, we can connect it through to other actions.

Note! This PR changes some workflows on main, which don't run in a PR - so there is a chance that these workflows might fail when merged.


gerhard:
Picking this up now since I have some related changes coming up and it would be best to have this merged before then.

jedevc:
Rebased, and also realized that we had missed bumping every runner in the last "improving release" PR - specifically the release job was still using v0.11.2.

gerhard (approved):
This now looks good to me!

If you agree with the changes @jedevc, it's OK to merge as soon as all checks pass. We can follow-up with improvements 💪

PR #7350, opened 2024-05-10
ci: add linting check for changie

jedevc:
Cause a linting failure if we generated a `.changes` file that isn't reflected in the `CHANGELOG.md`.

This happened because I missed a step during #7342. We can automate it, so this doesn't need to happen again :tada:


gerhard (approved):
👍 🚀

PR #7366, opened 2024-05-13
ci: update goreleaser to v1.26.0

jedevc:
https://github.com/dagger/dagger/pull/7264 updated to explicitly use go 1.21.7 everywhere.

However - the go releaser version we were using apparently only supports up to 1.21.4 - so we should update it, so that the failing publish job on `main` can pass: https://github.com/dagger/dagger/actions/runs/9064274083/job/24903799826

### DIFF
--- a/ci/cli.go
+++ b/ci/cli.go
@@ -34,7 +34,7 @@ func (cli *CLI) File(
 
 const (
 	// https://github.com/goreleaser/goreleaser/releases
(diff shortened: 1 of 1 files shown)


gerhard (approved):
👍 🚀

jedevc:
Passed on `main` :tada:

https://github.com/dagger/dagger/actions/runs/9066326504/job/24908981264

Issue #7369, opened 2024-05-14
✨ Hide <nil> return value when a function returns an error


sagikazarmark:
### What are you trying to do?

Get rid of `<nil>` when a function returns a single error.

### Why is this important to you?

Generally when you run a CLI tool, no output along with exit 0 means success. That extra `<nil>` is just unnecessary. (One might say annoying, but one might risk being called obsessive 😄)

### How are you currently working around this?

I look away. 😄

helderco:
Agreed!
- 👉 https://github.com/dagger/dagger/pull/7370

sagikazarmark:
![The-A-Team jpg](https://github.com/dagger/dagger/assets/1226384/170c52ce-619f-4512-916b-694272279a89)

PR #7370, opened 2024-05-14
cli: don't print `<nil>`

helderco:
Fixes https://github.com/dagger/dagger/issues/7369


jedevc:
Hm, why does the check at the bottom for `ReturnType.Kind != dagger.VoidKind` not trigger here?

helderco:
> Hm, why does the check below this for `ReturnType.Kind != dagger.VoidKind` not trigger here?

Good call, let me turn this around... one sec.

jedevc (approved):

helderco:
Been failing due to flakiness I guess. It's always different tests that fail, and they pass locally.

Without any more info:
```
process "..." did not complete successfully: exit code: 125
```

Latest:
```
testing.go:1398: race detected during execution of test
```

PR #7371, opened 2024-05-14
feat: mark `[internal]` buildkit prefixes as internal

jedevc:
Follow-up to #6835 (found during hackery on #7272)

We used to use the magic `[internal]` prefix in buildkit vertex names to mark these as internal to the UI. However, we've since updated our TUI and cloud logic to instead pull this information from an OTEL span attribute (which makes much more sense). However, we didn't have a way of connecting buildkit vertices to this attribute.

This wasn't immediately obvious - because of how [our TUI logic doesn't show spans that last longer than 100 milliseconds](https://github.com/dagger/dagger/blob/e25ff72f7111d09800ffa208492f00e0d1b6c309/dagql/idtui/frontend.go#L433-L436), and because lots of these were very quick `[internal] merge`s, then we didn't see these most of the time. However, some of the longer ones, like our internal blob copying, could show up if the context directory was too large. Additionally, if we remove this logic for `--progress=plain`, it shows up every time.

This patch provides this hiding functionality again, by making a custom `TraceProvider` implementation that wraps another one - with this provider, we can intercept calls to `Tracer.Start`, using this to strip the `[internal]` prefix from the name, and set the correct attribute on the span.

### DIFF
--- a/core/container.go
+++ b/core/container.go
@@ -1022,16 +1022,9 @@ func (container *Container) WithExec(ctx context.Context, opts ContainerExecOpts
 		return nil, err
 	}
 
-	var namef string
-	if container.Focused {
-		namef = buildkit.FocusPrefix + "exec %s"
-	} else {
-		namef = "exec %s"
-	}
-
 	runOpts := []llb.RunOption{
 		llb.Args(args),
-		llb.WithCustomNamef(namef, strings.Join(args, " ")),
+		llb.WithCustomNamef("exec %s", strings.Join(args, " ")),
 	}
 
 	// this allows executed containers to communicate back to this API
--- a/engine/buildkit/telemetry.go
+++ b/engine/buildkit/telemetry.go
@@ -1,0 +1,71 @@
+package buildkit
+
+import (
+	"context"
+	"strings"
+
+	"github.com/dagger/dagger/telemetry"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/embedded"
+)
+
+// buildkitTelemetryContext returns a context with a wrapped span that has a
+// TracerProvider that can process spans produced by buildkit. This works,
+// because of how buildkit heavily relies on trace.SpanFromContext.
+func buildkitTelemetryContext(ctx context.Context) context.Context {
+	if ctx == nil {
+		return nil
+	}
+	sp := trace.SpanFromContext(ctx)
+	sp = buildkitSpan{Span: sp, provider: &buildkitTraceProvider{tp: sp.TracerProvider()}}
+	return trace.ContextWithSpan(ctx, sp)
+}
+
+type buildkitTraceProvider struct {
+	embedded.TracerProvider
+	tp trace.TracerProvider
+}
+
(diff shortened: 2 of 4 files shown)


telemetry/inflight/proxy.go
```diff
@@ -47,7 +47,8 @@ type ProxyTracer struct {
 
 func (t ProxyTracer) Start(ctx context.Context, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
 	ctx, span := t.tracer.Start(ctx, spanName, opts...)
-	return ctx, proxySpan{sp: span, onUpdate: t.onUpdate}
+	span = proxySpan{sp: span, onUpdate: t.onUpdate}
+	return trace.ContextWithSpan(ctx, span), span
```
jedevc, line 51: @vito this was just something I found while in the area... I found I needed this for the `buildkitTracer` I made, and thought it *probably* makes sense here as well?

That said, I'm not 100% sure what kind of effect this would have - my guess is that previously we might have ended up with child spans that wouldn't have appeared as "live" updated spans?

engine/buildkit/telemetry.go
```diff
@@ -0,0 +1,75 @@
...
+	provider *buildkitTraceProvider
+}
+
+func (t *buildkitTracer) Start(ctx context.Context, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
+	internal := false
+
+	if strings.HasPrefix(spanName, "load cache: ") {
+		internal = true
+	}
+
```
vito, line 49: Not sure about this one - I think these are still useful to show in the UI, if anything I'd rather they just didn't have the `load cache: ` prefix and instead had a `cached: true` attribute or something so we could add a proper UI cue.

This is one area I was thinking we might want to just make upstream PRs to Buildkit, since I'm not sure how opinionated they are about it or if this is just their initial rough draft of Otel integration. Or maybe we just do all this at the SpanProcessor layer; there's probably a limit to how tightly we want couple Buildkit to our UI.

I'd say for now we should just leave it alone, since we'd want to coordinate any new `cached: true` attribute with making the UI/TUI actually respect it.

core/container.go
```diff
@@ -1022,16 +1022,9 @@ func (container *Container) WithExec(ctx context.Context, opts ContainerExecOpts
...
-	if container.Focused {
-		namef = buildkit.FocusPrefix + "exec %s"
-	} else {
-		namef = "exec %s"
-	}
-
 	runOpts := []llb.RunOption{
 		llb.Args(args),
-		llb.WithCustomNamef(namef, strings.Join(args, " ")),
+		llb.WithCustomNamef("exec %s", strings.Join(args, " ")),
```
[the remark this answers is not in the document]
jedevc: Looks like it's missing the `exec` prefix in buildkit: https://github.com/moby/buildkit/blob/6acf12ff9d488f8812482a6cc4845e395a82ae41/solver/llbsolver/vertex.go#L337-L338

I think keeping it makes sense, so will leave this as is for now.

engine/buildkit/telemetry.go
jedevc: Yup, sounds good to me, have removed this case.

vito (approved):

PR #7376, opened 2024-05-14
Setup dagger engines with buildkit gc disabled

matipan:
This is a follow up from an internal discussion where @samalba has been investigating the performance of our CI and found that disabling bulidkit's GC has the potential of improving stability of the most heavy jobs. We want to test this in production for a few days and have side by side data to compare the performance and stability of the jobs. I've been running a few isolated tests here: https://github.com/dagger/dagger/actions/runs/9084221778, but we need more data in order to drive a conclusion regarding the GC.

In this PR we are adding two jobs for the engine-and-cli workflow:
- test-nogc
- testdev-nogc

This jobs will run in parallel and in a synchronous manner to baseline jobs. To implement it there were two minor changes introduced to `_dagger_call`:
- size-dind variable: this variable allows us to customize the `runs-on` for testdev which we couldn't so far.
- concurrency: allows the workflow dispatching dagger_call to control the concurrency group. We need this, otherwise nogc and baseline will step over each other's toes and cancel each other out.

### DIFF
--- a/.github/workflows/engine-and-cli.yml
+++ b/.github/workflows/engine-and-cli.yml
@@ -28,6 +28,14 @@ jobs:
       function: test all --race=true
       size: "dagger-v0-11-4-16c-nvme"
 
+  test-nogc:
+    uses: ./.github/workflows/_dagger_call.yml
+    secrets: inherit
+    with:
+      function: test all --race=true
+      size: "dagger-v0-11-4-16c-nogc"
+      concurrency: "-nogc"
+
   # Run Engine tests in dev Engine so that we can spot integration failures early
   # Only run a subset of important test cases since we just need to verify basic
   # functionality rather than repeat every test already run in the other targets.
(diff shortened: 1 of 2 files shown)


.github/workflows/_hack_make.yml
```diff
@@ -21,7 +21,11 @@ on:
         default: 60  
       size:
         type: string
-        default: "dagger-v0-11-4-4c-nvme"
+        default: "dagger-v0-11-4-4c-nogc"
+        required: false
+      dev-engine-size:
+        type: string
+        default: "dagger-v0-11-4-dind-nogc"
```
gerhard, line 28: Rather than introduce a new variable, could we append `-dind` for testdev?

.github/workflows/engine-and-cli.yml
```diff
@@ -22,6 +22,13 @@ permissions:
 
 jobs:
   test:
+    uses: ./.github/workflows/_hack_make.yml
+    secrets: inherit
+    with:
+      mage-targets: engine:testrace
+      size: "dagger-v0-11-4-16c-nogc"
```
gerhard, line 29: I would leave the default name as it was, and instead introduce a new job with `-nogc` appended. This way, it will be clearer that we are introducing a new variant. It's easier to explain & understand, rather than what we had now becomes `baseline`, and the default changed to not having GC.

gerhard (commented):
Left a few tiny comments. Also, this will need rebasing after these changes went it:
- https://github.com/dagger/dagger/pull/7349

PR #7378, opened 2024-05-14
docs: Add multi-lang IDE integration page

vikram-dagger:


helderco:
Some thoughts:
- I think we can make a common explanation for this doc, before the tabs.
- The user needs to learn about this much sooner, right after creating the first function as it's important for DX to get that LSP experience.
- If there's info on multiple tools (like in Python), put them in a nested tabbed group.

vikram-dagger:
@helderco please see if my edits address your feedback. Thanks!

helderco:
This goes without saying that the IDE integrations page needs to go after the user creates their first function, but I assume it's placed the way it is because the nav is in flux due to the pending PRs. The problem with the multiple PRs here is that it puts the docs in an unpublished state in main. Would  have made sense to merge PRs against a larger PR for the broader effort, while still keeping the individual reviewable changesets. 

There's a few things to correct in Python as only pip needs to manually create and activate the venv for example, but I'll need a bit more time to suggest specific changes.

Few more notes:

- I think it's useful to have a common paragraph on `dagger develop`, before the tabs.
- Just installing the dependencies locally isn't enough. The focus should be on the IDEs. If you're using VSCode for example, it may not matter if you've installed and activated the venv in your shell if the editor handles those in it's own way. I think it's still usefull to show those instructions, but as an addition to making sure the editor knows about the venv.
- I suppose this page could be simplified if you think about a person using the default template. In that case, the user won't be using Poetry for example, but it might be using Hatch though. These tool specific details could go into the language-specific details page though, and keep this one as simple as possible for those that are starting out.

PR #7384, opened 2024-05-15
ci: fix typo in publish go sdk

jedevc:
Follow-up to #7349

This should be "go", not "elixir". Looks like a bad case of copy-pasta.


helderco (approved):

helderco:
Are the errors in Elixir and Java flukes?

jedevc:
@helderco I'm assuming so, but clearly something is a bit weird here - I don't quite know why these exit with 125, and *no* apparent error message?

Issue #7375, opened 2024-05-14
✨ Dagger CLI should time out more when there is an issue with docker 


levlaz:
### What are you trying to do?

Right now if there is some issue with docker the dagger CLI will continue to try to `connect` indefinitely. We had a very patient person who waited 56 minutes before giving up. :) 

Ideally we either check to make sure docker is going to work, or time out if we are stuck connecting for more than say 5 minutes. 

### Why is this important to you?

Especially for people new to Dagger this can be a very bad first impression and a very poor DX. 

### How are you currently working around this?

There is no real workaround, you have to know to check for docker issues and that is not always obvious to newcomers.

Issue #6643, opened 2024-02-09
🐞 0.9.9 Regression, TUI is not showing the engine init status


nipuna-perera:
### What is the issue?

When dagger engine runs initially with a new version the TUI used to show that the new engine image is being downloaded. Now it looks like it's frozen.

### Dagger version

v0.9.9

### Steps to reproduce

remove the existing dagger engine (if already at 0.9.9) and run `dagger call`

PR #7385, opened 2024-05-15
Improved connection progress

jedevc:
Continued progress exploration! I just kinda kept pulling on #7375, and kept finding more things to tidy up :tada:

Fixes:
- #7375
- #6643

This changes a few things about the initial connection:
- Changes timeouts, to provide a sane upper limit on the amount of time allowed to provision an engine, e.g. pulling+starting a docker container (10 minutes) 
- Splits up the `connect` step into separate parts - this allows for improved debugability if this step is hanging - we can see exactly which part is being slow.
	- Removes noisy "OK!" message from session connection - this isn't really needed anymore, success is indicated by the completion of the span.
- Makes sure that `Connected to engine` messages correctly appear with our pretty progress (they weren't being displayed)
- Restores support for `Connected to cloud` messages.
- Ensures that all logging messages respect the detected terminal color profile.

Before:

```
$ dagger call --source=.:default helm test -v
✔ connect 0.8s
✔ initialize 1.1s
✔ ModuleSource.resolveFromCaller: ModuleSource! 0.1s
  ✔ upload /home/jedevc/Documents/Projects/dagger from dragon (client id: kh8jcn3x5ql8ba7z8dhurxnzi) (exclude: **/.git) (include: **/go.mod, **/go.sum, **/go.work, **/go.work.sum, **/vendor/, **/*.go, dagger.json, ci/**/*) 0.1s
✔ ModuleSource.resolveDirectoryFromCaller(path: ".", viewName: "default"): Directory! 0.3s
  ✔ upload /home/jedevc/Documents/Projects/dagger from dragon (client id: kh8jcn3x5ql8ba7z8dhurxnzi) (exclude: bin, .git, **/node_modules, **/.venv, **/__pycache__) 0.3s
✔ dagger(
    source: ✔ blob(digest: "sha256:cbe8a7976450916e3ed0ebadd6eb143df283cc4f4610601bbbb0a32a48662059", mediaType: "application/vnd.oci.image.layer.v1.tar+zstd", size: 5153392, uncompressed: "sha256:a6464e8f195d11a5f4f4588cd8ab8362b85ed8870953b8d66863c637ab9a3c80"): Directory! 0.0s
  ): Dagger! 0.8s
✔ Dagger.helm: DaggerHelm! 0.6s
✔ DaggerHelm.test: Void 1.5s
  ✔ Container.from(address: "alpine/helm:3.12.3"): Container! 0.8s
    ✔ HTTP GET 0.3s
    ✔ remotes.docker.resolver.HTTPRequest 0.4s
      ✔ HTTP HEAD 0.4s

<nil>
```

After:

```
$ dagger call --source=.:default helm test -v
12:11:45 INF Connected to cloud url=https://dagger.cloud/traces/8d82a45c63f06f1cdd68607d76263869
12:11:46 INF Connected to engine name=cda0040f8104 version=v0.11.4

✔ connect 0.8s
  ✔ starting engine 0.6s
  ✔ connecting to engine 0.1s
  ✔ starting session 0.2s
✔ initialize 0.9s
✔ dagger(
    source: ✔ blob(digest: "sha256:cbe8a7976450916e3ed0ebadd6eb143df283cc4f4610601bbbb0a32a48662059", mediaType: "application/vnd.oci.image.layer.v1.tar+zstd", size: 5153392, uncompressed: "sha256:a6464e8f195d11a5f4f4588cd8ab8362b85ed8870953b8d66863c637ab9a3c80"): Directory! 0.0s
  ): Dagger! 0.6s
✔ Dagger.helm: DaggerHelm! 0.6s
✔ DaggerHelm.test: Void 1.1s
  ✔ Container.from(address: "alpine/helm:3.12.3"): Container! 0.4s
    ✔ remotes.docker.resolver.HTTPRequest 0.4s
      ✔ HTTP HEAD 0.4s

<nil>
```

### DIFF
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -31,6 +31,7 @@ import (
 	"github.com/dagger/dagger/analytics"
 	"github.com/dagger/dagger/dagql/idtui"
 	"github.com/dagger/dagger/engine"
+	"github.com/dagger/dagger/engine/slog"
 	"github.com/dagger/dagger/telemetry"
 	"github.com/dagger/dagger/telemetry/sdklog"
 )
@@ -271,6 +272,10 @@ func main() {
 		ctx, stdout, stderr := telemetry.WithStdioToOtel(ctx, "dagger")
 		rootCmd.SetOut(stdout)
 		rootCmd.SetErr(stderr)
+
+		if url, ok := telemetry.URLForTrace(ctx); ok {
+			slog.Info("Connected to cloud", "url", url)
+		}
 
 		return rootCmd.ExecuteContext(ctx)
 	}); err != nil {
--- a/engine/client/client.go
+++ b/engine/client/client.go
@@ -168,10 +168,7 @@ func Connect(ctx context.Context, params Params) (_ *Client, _ context.Context, 
 		return nil, nil, fmt.Errorf("cache config from env: %w", err)
 	}
 
-	connectSpanOpts := []trace.SpanStartOption{
-		telemetry.Encapsulate(),
-	}
-
+	connectSpanOpts := []trace.SpanStartOption{}
 	if configuredServerID != "" {
 		// infer that this is not a main client caller, server ID is never set for those currently
 		connectSpanOpts = append(connectSpanOpts, telemetry.Internal())
@@ -226,14 +223,19 @@ func (c *Client) startEngine(ctx context.Context) (rerr error) {
 		cloudToken = v
 	}
 
-	connector, err := driver.Provision(ctx, remote, &drivers.DriverOpts{
+	provisionCtx, provisionSpan := Tracer().Start(ctx, "starting engine")
+	connector, err := driver.Provision(provisionCtx, remote, &drivers.DriverOpts{
 		UserAgent:        c.UserAgent,
 		DaggerCloudToken: cloudToken,
 		GPUSupport:       os.Getenv(drivers.EnvGPUSupport),
 	})
+	telemetry.End(provisionSpan, func() error { return err })
 	if err != nil {
 		return err
 	}
+
+	ctx, span := Tracer().Start(ctx, "connecting to engine")
+	defer telemetry.End(span, func() error { return rerr })
 
 	if err := retry(ctx, 10*time.Millisecond, func(elapsed time.Duration, ctx context.Context) error {
 		// Open a separate connection for telemetry.
--- a/engine/client/client.go
+++ b/engine/client/client.go
@@ -224,11 +224,13 @@ func (c *Client) startEngine(ctx context.Context) (rerr error) {
 	}
 
 	provisionCtx, provisionSpan := Tracer().Start(ctx, "starting engine")
+	provisionCtx, provisionCancel := context.WithTimeout(provisionCtx, 10*time.Minute)
 	connector, err := driver.Provision(provisionCtx, remote, &drivers.DriverOpts{
 		UserAgent:        c.UserAgent,
 		DaggerCloudToken: cloudToken,
 		GPUSupport:       os.Getenv(drivers.EnvGPUSupport),
 	})
+	provisionCancel()
 	telemetry.End(provisionSpan, func() error { return err })
 	if err != nil {
 		return err
--- a/engine/client/client.go
+++ b/engine/client/client.go
@@ -295,7 +295,7 @@ func (c *Client) startSession(ctx context.Context) (rerr error) {
 	ctx, sessionSpan := Tracer().Start(ctx, "starting session")
 	defer telemetry.End(sessionSpan, func() error { return rerr })
 
-	ctx, stdout, stderr := telemetry.WithStdioToOtel(ctx, InstrumentationLibrary)
+	ctx, _, stderr := telemetry.WithStdioToOtel(ctx, InstrumentationLibrary)
 
 	hostname, err := os.Hostname()
 	if err != nil {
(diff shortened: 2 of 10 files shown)


cmd/dagger/main.go
```diff
@@ -253,6 +254,10 @@ func main() {
 		rootCmd.SetOut(stdout)
 		rootCmd.SetErr(stderr)
 
+		if url, ok := telemetry.URLForTrace(ctx); ok {
+			slog.Info("Connected to cloud", "url", url)
+		}
```
jedevc, line 259: So this works :tada: At this point we have a root span, so we have a trace ID so can get the URL for the telemetry to be viewed at - but I'm not sure, at this point is this page guaranteed to exist? At what point can we guarantee this (and defer printing till then ideally)?

telemetry/url.go
```diff
@@ -0,0 +1,15 @@
+package telemetry
+
+import (
+	"context"
+
+	"go.opentelemetry.io/otel/trace"
+)
+
+func URLForTrace(ctx context.Context) (string, bool) {
```
jedevc, line 9: Fun little idea about this - this function could imply the existence of a `URLForSpan` function :tada:

On a failure, we could print out the URL of a failed span (or even use [OSC 8](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) to display it inline!

engine/client/client.go
```diff
@@ -288,10 +286,7 @@ func (c *Client) startEngine(ctx context.Context) (rerr error) {
 
 	c.bkClient = bkClient
 
-	if c.EngineNameCallback != nil {
-		engineName := fmt.Sprintf("%s (version %s)", bkInfo.BuildkitVersion.Revision, bkInfo.BuildkitVersion.Version)
-		c.EngineNameCallback(engineName)
-	}
+	slog.Info("Connected to engine", "name", bkInfo.BuildkitVersion.Revision, "version", bkInfo.BuildkitVersion.Version)
```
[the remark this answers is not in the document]
vito: Worth a shot yea! We don't normally show logs so this was probably just flying under the radar.

There's a general idea that the bottom of the screen in a CLI is actually the "prime real estate" unlike with a web UI, since that's where the user will be looking - but I think you could easily make the argument that the active things belong in the prime real estate, instead of the logs, which are more in the background.

telemetry/url.go
```diff
@@ -0,0 +1,15 @@
...
+	if !configuredCloudTelemetry {
+		return "", false
+	}
+	url := "https://dagger.cloud/traces/" + trace.SpanContextFromContext(ctx).TraceID().String()
```
aluzzardi, line 13: The Cloud URL requires the org name: `https://dagger.cloud/<ORG>/traces/<TRACE ID>`

Unfortunately, we don't have that readily available, but that's being fixed:

- When using `dagger login` https://github.com/dagger/dagger/pull/7125, there will be an `org.Name` available
- When using a `DAGGER_CLOUD_TOKEN`, there's a new token format that allows deriving the org name from the token (`dag_<org name>_secret`). Coming today
  - If using an old token format, we should tell the user a generic "Visit https://dagger.cloud -- blah blah -- rotate your token if you want to see the trace URL from the CLI"

jedevc: Mm, I figured this would be required - that said, it does still "seem" to land on the right page interestingly.

aluzzardi: > Mm, I figured this would be required - that said, it does still "seem" to land on the right page interestingly.

Yeah, AFAIK if no org is explicitly passed, then it just redirects to "current" org (current being the last one you used). So it does work if the conditions are alright, but it's not super dependable

engine/client/client.go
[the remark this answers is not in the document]
jedevc: Updated to do this now, it's thankfully a super simple change - it also improves consistency, since this is what we do during `finalRender` already :tada:

gerhard (approved):
Looks great! The Engine info & Dagger Cloud Trace URL are particular nice touches. The structure of the output is also much neater. Great job 👏

First `1m 30s` - top is this change, bottom is `v0.11.4`
https://github.com/dagger/dagger/assets/3342/24390312-ee09-4f2c-8933-3ca00d61001b

Last `1m` - top is this change, bottom is `v0.11.4`
https://github.com/dagger/dagger/assets/3342/0887f181-473b-4914-9083-6650611cacc5

helderco:
Sorry for being late here, but I think the "Connected to engine" line should be at debug level.

jedevc:
@helderco I think that's fair - the only issue is that then this will only display with the `--debug` flag, and turning that on produces far too much internal logging. Ideally, some of the internal logging would be set to our `ExtraDebug` level/trace.

Previously, we always displayed the engine connection information in plain progress, it was just missing in the TUI progress.

helderco:
The thing is, we used to have both the version and the cloud url in a single, more compact line, but it was constantly being flagged as noise, especially by Solomon. Maybe having it at the top is better, but at least the version is not necessary to see it every time. Doesn't have to be a debug record, can you put it at `-v`?

jedevc:
Yeah, that's possible requires a little bit of undoing some of the refactors of this PR - I've put something somewhat reasonable in [`c2f2cec` (#7272)](https://github.com/dagger/dagger/pull/7272/commits/c2f2cec7b6d0d24d7f36410471706e3c138c2904).

aluzzardi:
@jedevc This is awesome, thank you!

One post merge nit: If we're running with a "legacy" token, instead of not printing, any chance we could say something like "Blah blah trace sent to https://dagger.cloud/. Please rotate your Cloud token to get the trace URL from the CLI"

aluzzardi:
@vito @jedevc Some minor UX nitpicks: When using `tty` progress, the URL gets "wiped out" eventually. And with plain progress it's at the very beginning, which works well in CI logs starting from the top but it's harder to find in a terminal.

Thoughts on how we could make it more prominent?

PR #7386, opened 2024-05-15
fix: hide remotes.docker.resolver spans

jedevc:
Before this, we would display all the containerd/buildkit spans that relate to pulling an image. These are *super* unhelpful by default, so we want to hide them in the TUI (unless the parent command errors).

We already have this concept today with `telemetry.Encapsulate`. However, this is set on the parent (in our case, `Container.from`), which isn't quite what we want. Firstly, this would silence all output from this function (even if later we would want to add some more useful output of our own). Secondly, in some builds, these requests might not necessarily even occur as part of `Container.From` - we might end up lazily pulling the image later.

So, we add a new `telemetry.Encapsulated` property - this is exactly the same as the above, but is instead set on a child span. We can set this using our new buildkit tracing interceptor, and only set it on the buildkit spans that we care about.

Before:
```
$ dagger call --source=.:default engine lint -v
✔ connect 1.0s
✔ initialize 7.3s
...
● DaggerEngine.lint: Void 8.1s
  ✔ Container.from(address: "golangci/golangci-lint:v1.57-alpine"): Container! 0.7s
    ✘ remotes.docker.resolver.HTTPRequest 0.3s
      ✘ HTTP HEAD 0.3s
    ✔ HTTP GET 0.2s
    ✔ remotes.docker.resolver.HTTPRequest 0.2s
      ✔ HTTP HEAD 0.2s
  ✔ Container.from(address: "golang:1.22.3-alpine3.18"): Container! 0.9s
    ✔ HTTP GET 0.4s
    ✔ remotes.docker.resolver.HTTPRequest 0.1s
      ✔ HTTP HEAD 0.1s
    ✔ remotes.docker.resolver.HTTPRequest 0.3s
      ✔ HTTP GET 0.3s
```

After:
```
$ dagger call --source=.:default engine lint -v
...
● DaggerEngine.lint: Void 8.9s
  ✔ Container.from(address: "golangci/golangci-lint:v1.57-alpine"): Container! 0.9s
```

### DIFF
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -592,8 +592,10 @@ func (fe *Frontend) renderRow(out *termenv.Output, row *TraceRow, depth int) err
 	}
 	if !row.Span.Encapsulate || row.Span.Status().Code == codes.Error || fe.Verbosity >= 2 {
 		for _, child := range row.Children {
-			if err := fe.renderRow(out, child, depth); err != nil {
-				return err
+			if !child.Span.Encapsulated || row.Span.Status().Code == codes.Error || fe.Verbosity >= 2 {
+				if err := fe.renderRow(out, child, depth); err != nil {
+					return err
+				}
 			}
 		}
 	}
--- a/engine/buildkit/telemetry.go
+++ b/engine/buildkit/telemetry.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 
 	"github.com/dagger/dagger/telemetry"
-	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/embedded"
 )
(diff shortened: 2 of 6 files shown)


telemetry/attrs.go
```diff
@@ -33,8 +33,16 @@ const (
...
 	// Hide child spans by default.
+	//
+	// Encapsulated child spans may typically be revealed if the parent span errors.
 	UIEncapsulateAttr = "dagger.io/ui.encapsulate"
 
+	// Hide span by default.
+	//
+	// This is functionally the same as UIEncapsulateAttr, but is instead set
+	// on a child instead of a parent.
+	UIEncapsulatedAttr = "dagger.io/ui.encapsulated"
```
vito, line 44: Cool idea, but is this significantly different from `Internal()`? :thinking:

jedevc: Yes! So with Internal, if it fails, it always shows.

These requests will sometimes fail, e.g. with a 401 if they need authentication (this happens pretty often). Encapsulated lets us hide this unless the failure propagates to the parent, and the whole pull fails.

dagql/idtui/frontend.go
```diff
@@ -592,8 +592,10 @@ func (fe *Frontend) renderRow(out *termenv.Output, row *TraceRow, depth int) err
 	}
 	if !row.Span.Encapsulate || row.Span.Status().Code == codes.Error || fe.Verbosity >= 2 {
 		for _, child := range row.Children {
-			if err := fe.renderRow(out, child, depth); err != nil {
-				return err
+			if !child.Span.Encapsulated || row.Span.Status().Code == codes.Error || fe.Verbosity >= 2 {
```
vito, line 595: Ah ha, got it - so we'll reveal inner encapsulated spans if the parent fails. Nice.

engine/buildkit/telemetry.go
```diff
@@ -41,19 +40,23 @@ type buildkitTracer struct {
 }
 
 func (t *buildkitTracer) Start(ctx context.Context, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
```
vito, line 42: I wonder if there's a clean way to do this as part of our SpanProcessor pipeline instead - it might be more robust than relying on Buildkit using `ctx`. Not a blocker, just thinking out loud. Vaguely thinking something like identifying a span's instrumentation library ("this came from buildkit") and then directly modifying it in `OnStart` where we have write access. But I'm not sure if there's enough identifying information.

vito (approved):

engine/buildkit/telemetry.go
jedevc: > "this came from buildkit"

I think this is difficult right now:

https://github.com/moby/buildkit/blob/75bbf7b2071ef8b1d8dff70d4d3d878a53c6d991/util/tracing/tracing.go#L25-L27

This is where you're supposed to put a nice little uri right? We put `dagger.io/...` here, maybe some more upstream PRs are called for :cry:

vito: Haha yep, just ran into that myself. I think it'd be worth it. Otherwise it just defaults to the Go Otel SDK instrumentation library value I think.

PR #7390, opened 2024-05-16
docs: Update docs site header to match dagger.io

vikram-dagger:


vikram-dagger:
Notes:

- The font and style is different. This will need a change to the docs site theme/CSS. Maybe @crjm can help with this?
- Should `Get involved` -> `Get Involved`, on both sites?
- I can't find a way to move the search bar down, I believe this will need some customization to the theme/CSS so needs @crjm or someone else familiar with front-end UX

mircubed:
@shykes This is the preview: https://deploy-preview-7390--devel-docs-dagger-io.netlify.app/


This is the closest we can get to the universal heading for now without additional engineering resources dedicated to it. Are you ok with moving forward with this?

docs/docusaurus.config.ts
```diff
@@ -127,21 +127,68 @@ const config: Config = {
...
+          className: "navbar-blog-link",
+        },
+        {
+          position: "left",
+          type: "dropdown",
+          label: "Community",
+          className: "navbar-blog-link",
+          items: [
+            {
+              label: "Get involved",
```
mircubed, line 172: Change to "Get Involved". I updated dagger.io after this PR went out.

crjm:
> I can't find a way to move the search bar down, I believe this will need some customization to the theme/CSS so needs @crjm or someone else familiar with front-end UX

@vikram-dagger @mircubed sorry, missing some context here, where should I position the searchbar?

Neelajacques:
+1 Feels like a step in the right direction, I look forward to seeing us getting closer and closer alignment across the two sites.

mircubed:
It looks great @crjm! However, the "daggerverse", "resources" and "blog" tabs are still opening as new tabs. All the clickable options under "platform" are working great. Can you fix that? Once done, let me know and we can merge

mircubed (approved):
LGTM!

PR #7394, opened 2024-05-16
Update community link in README

mircubed:
updated link to community activities since it was broken


jedevc (approved):

jedevc:
Not as sure about the process for this, but we should also get a redirect from `/events` to `/community`.

gerhard:
That is a change that needs to be done in Framer.io. cc @mircubed

Issue #6894, opened 2024-03-16
🐞 Setting `--oci-max-parallelism 2` can cause deadlocks


gerhard:
### What is the issue?

Setting `--oci-max-parallelism 2` on the Engine causes deadlocks, both with & without modules.

This is what that looks like from the CLI / user perspective:

https://github.com/dagger/dagger/assets/3342/826ddb73-c36e-469e-9560-f21b3fd13388

Attaching Dagger Engine v0.10.2 logs + `SIGQUIT`. Thank you @vito for https://vito.github.io/swirly/ and the tips!
- [dagger-engine-010x.sigquit.txt](https://github.com/dagger/dagger/files/14624908/dagger-engine-010x.sigquit.txt)
- [dagger-engine-010x.log.txt](https://github.com/dagger/dagger/files/14624909/dagger-engine-010x.log.txt)

FWIW:

https://github.com/dagger/dagger/assets/3342/43657975-5e84-41f4-b199-14bb7773a5a9

cc @sipsma @vito @jedevc @aluzzardi

### Dagger version

Confirmed on 0.9.11 & 0.10.2

Scalahansolo:
The team over here at Motion have seen this happen to us as well on 0.10.3

jedevc:
I think this is expected - suppose the following module structure:

- Module A starts (running a container)
- Module A calls a function in Module B
- Module B start (running a container)
- Module B calls `dag.Container()` (running a container)
	- This hangs forever! There are already 2 containers running in parallel, so no more can be scheduled.

I think with modules, this effectively means that `max-parallelism` is pretty much just completely broken - it just fits very badly with the idea of nested containers.

Not sure what the right fix is here:
- Don't use `max-parallelism` (generally, my preference, you should enforce max usage using different mechanisms instead, since a single job-step could use anywhere from 0% to 100% of available resources - they're not just replaceable)
- When using a nested step, don't have this "consume a resource" (this prevents deadlocks, but isn't great, since there's now more containers running than the user asked for)

I don't think there's a clear way forwards here, but to me this kind of highlights that we need a better way of constraining resources - `max-parallelism` is just really not fit for the shape that pipelines tend to take these days. Maybe we should provide some cgroup config options instead (if the end desire is to constrain CPU usage/etc).

jedevc:
Adding this to the v0.12.0 milestone since by that release we should probably either:
- Remove this flag
- Ensure that nested execs don't count towards max-parallelism
[...]

PR #7395, opened 2024-05-16
Remove oci-max-parallelism entirely

gerhard:
Fixes https://github.com/dagger/dagger/issues/6894

This is not an option that we should support. Justin puts it best:

    I think with (Dagger) Modules, this effectively means that
    `max-parallelism` is pretty much just completely broken - it just fits
    very badly with the idea of nested containers.

    ... but to me this kind of highlights that we need a better way of
    constraining resources - max-parallelism is just really not fit for
    the shape that pipelines tend to take these days. Maybe we should
    provide some cgroup config options instead (if the end desire is to
    constrain CPU usage/etc).

While this can still be configured if a custom Engine toml config is provided, it means that users went out of their way to have this. It's hard on purpose and we don't recommend using this setting anymore.

WDYT @sipsma @jedevc @aluzzardi @vito ?


gerhard:
Speaking with @aluzzardi, these checks may have failed due to a ClickHouse migration which killed some in-flight traces. Rebasing & force pushing so that we force re-run all failed checks.

jedevc:
If possible, I'd like to hold off on merging this. If I have the time, I think keeping this option would be best for people using it, but we'd need the improved semantics previously described. But we can keep this open at least as a tracking issue.

> @matipan if everyone is in agreement with this change, we should remove this from all our infra. We can do that before the release goes out.

IMO, we should do this as soon as possible - it's definitely a potential candidate for some of the weird ephemeral CI failures we've seen. With this option it's possible to get every running test is a stuck state (if the right combination of tests all start at the right time) - which prevents any progress being made (which sounds like a symptom we've seen before).

sipsma:
Agree with @jedevc, this option is used already by multiple users and while it's totally buggy once modules are used with too low a value, that feels more like a bug than enough justification to remove this entirely.
* BTW, you asked about the `num-cpu` option; that was added because it was requested by a user a while back. It makes it easier for users to scale this setting across many different hosts with many different CPU counts without having to jump through hoops to update the engine config on each host.

I'd totally be in favor of adding a scary warning message to the help for this though and tracking a long term fix to the problem.
* Actually now that I think about it, all the session related refactoring work should make fixing this a lot more tenable (what would have been a large project to medium at most).

gerhard:
Thank you for the extra context @sipsma & @jedevc.

I will close this & re-open it as a non-breaking change that is restricted to the Helm Chart only.

PR #7396, opened 2024-05-16
Build Dagger docs with Dagger

shykes:
Examples:

- Build the docs site: `dagger call --source=. site export --path=./docs-site`
- Build the docs server: `dagger call --source=. server`
- Run docs locally: `dagger call --source=. server as-service up`
- Run docs for a PR: `dagger call --source=https://github.com/dagger/dagger#pull/7382/head server as-service up`

### DIFF
--- a/ci/docs.go
+++ b/ci/docs.go
@@ -27,6 +27,29 @@ pagination_prev: null
 # CLI Reference
 `
 
+// Build the docs website
+func (d Docs) Site() *Directory {
+	return dag.
+		Docusaurus(
+			d.Dagger.Source,
+			DocusaurusOpts{Dir: "/src/docs", DisableCache: true},
+		).
+		Build()
+}
+
+// Build the docs server
+func (d Docs) Server() *Container {
+	nginxConfig := dag.CurrentModule().Source().File("docs-nginx.conf")
+	return dag.
+		Container().
+		From("nginx").
+		WithoutEntrypoint().
+		WithFile("/etc/nginx/conf.d/default.conf", nginxConfig).
+		WithDefaultArgs([]string{"nginx", "-g", "daemon off;"}).
+		WithDirectory("/var/www", d.Site()).
+		WithExposedPort(8000)
+}
+
 // Lint documentation files
 func (d Docs) Lint(ctx context.Context) error {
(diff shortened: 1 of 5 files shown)


ci/docs.go
```diff
@@ -27,6 +27,29 @@ pagination_prev: null
...
+		Docusaurus(
+			d.Dagger.Source,
+			DocusaurusOpts{Dir: "/src/docs", DisableCache: true},
+		).
+		Build()
+}
+
+// Build the docs server
+func (d Docs) Server() *Container {
+	nginxConfig := dag.CurrentModule().Source().File("docs-nginx.conf")
```
jedevc, line 42: :eyes: this file isn't part of this PR:

```
marshal: json: error calling MarshalJSON for type *dagger.Container: input: blob
.withDirectory.withNewFile.withNewFile.withNewFile.directory.file resolve: lstat
 /ci/docs-nginx.conf: no such file or directory
```

shykes: > 👀 this file isn't part of this PR:

Thanks, fixed!

> The docusaurus module also looks to have a Serve field directly as well.

That function runs what I consider to be a dev version `npm run serve`. I chose to serve the static files with nginx. This more closely matches the current production environment. Next step will be to apply the same redirect configuration (currently expressed in a proprietary netlify.tom file).

shykes:
To test that this PR works:

```console
dagger -m github.com/dagger/dagger@pull/7396/head call --source=github.com/dagger/dagger#pull/7396/head \
 docs \
 server \
 as-service \
 up
```

vito:
> I reference the same remote git branch twice, but have to use two different syntaxes (`#` vs `@`) (is there an issue for this? cc @helderco @jedevc @vito )

This was part of the refs v2 idea, unifying module ref syntax with git ref syntax. Not sure if there's a path to that from @grouville's current work on VCS-independent refs. Maybe?

> How do I hook up Dagger Cloud so that I can run it on my laptop, and the result is added as a check on this PR? cc @aluzzardi @vito @crjm

This was disabled for now because it seemed too unusual, wanted to give it more thought. Now the checks will be only added if it's CI=true. I wonder if you could fool it with `env CI=true dagger call ...`?

vito (approved):

PR #7407, opened 2024-05-18
docs: Convert debugging page

vikram-dagger:


vikram-dagger:
Noting here that the Python debugging page contains additional information, which has been transferred into the language-specific page. See https://github.com/dagger/dagger/pull/7408/

vikram-dagger:
https://dagger.cloud/dagger/traces/8b7aa1a600a20884ffd6f895633babbc - TS lint shows a failure, but https://github.com/dagger/dagger/actions/runs/9140565827/job/25134037910?pr=7407 - TS lint (which is presumably doing the same thing) shows success. Not quite sure what this means or how to turn it green. Please lmk if any suggestions @gerhard @matiasinsaurralde

gerhard:
I am assuming that the trace failed because the span never terminated:
image

The only way that I was able to make this type of check green is to run it again, even if the job succeeded. CC-ing @vito & @crjm @aluzzardi  in case they have a different suggestion. Triggering the re-run to check if that still works as I remember.

docs/current_docs/manuals/developer/debugging.mdx
```diff
@@ -12,3 +23,7 @@ The Dagger Engine runs in a dedicated Docker container and emits log messages as
 DAGGER_ENGINE_DOCKER_CONTAINER="$(docker container list --all --filter 'name=^dagger-engine-*' --format '{{.Names}}')"
 docker logs $DAGGER_ENGINE_DOCKER_CONTAINER
 ```
+
+Also see:
+
+- Additional debugging suggestions (Python)
```
gerhard, line 29: Was this meant to be a link?

gerhard (approved):
Apart from my comment, LGTM!

docs/current_docs/manuals/developer/debugging.mdx
vikram-dagger: Yes, thanks for the catch. Fixed now.

crjm:
> I am assuming that the trace failed because the span never terminated

Yes, this is the reason why this is happening. 
The error comes from the API, and not from the engine, as the span signaling the end of the execution never reached the API.
This is also why the engine doesn't exit with an error code in CI.

@vito it seems like we are setting the state in the FE for timeouts, but we are not showing any UI for that state.
Should we add an element to represent it?

docs/current_docs/manuals/developer/debugging.mdx
```diff
@@ -40,19 +54,19 @@ Using the command `dagger call --debug echo --msg="hello"`, should print
 something like:
 
 ```
-✔ MyModule.echo(msg: "hello"): String! 0.5s
+✔ Mymodule.echo(msg: "hello"): String! 0.5s
```
vikram-dagger, line 57: @helderco this is not the current output. The current output contains a very large number of spans, and I wasn't able to find anything similar to what is shown here. Please suggest what should be done for this output.

levlaz:
@vikram-dagger when I run this I see the following final output: 

```
✔ initialize 5.3s
✔ debug: Debug! 0.9s
┃ [DEBUG] dagger.mod._resolver: func => <Signature () -> None>
┃ [DEBUG] dagger.mod._resolver: input args => {}
┃ [DEBUG] dagger.mod._resolver: structured args => {}
┃ [DEBUG] dagger.mod._module: result => Debug()
┃ [DEBUG] dagger.mod._module: output => '{}'
┃ [DEBUG] dagger.client._session: Closing client session to GraphQL server
  ✔ connect 0.1s
✔ Debug.echo(msg: "hello"): String! 0.9s
┃ [DEBUG] dagger.mod._resolver: func => <Signature (msg: str) -> str>
┃ [DEBUG] dagger.mod._resolver: input args => {'msg': 'hello'}
┃ [DEBUG] dagger.mod._resolver: structured args => {'msg': 'hello'}
┃ [DEBUG] dagger.mod._module: result => 'hello'
┃ [DEBUG] dagger.mod._module: output => '"hello"'
┃ [DEBUG] dagger.client._session: Closing client session to GraphQL server
  ✔ connect 0.1s

hello
```

However, while its executing I see a ton of DBG messages like this: 

```
15:41:17 DBG recording span span=moby.buildkit.v1.Control/Session id=cd9772c3f2d5b7a0
15:41:17 DBG recording span child span=moby.buildkit.v1.Control/Session parent=3c1b9516c5eb911a child=cd9772c3f2d5b7a0
15:41:17 DBG new end old="2024-05-20 15:41:16.972974556 -0700 PDT" new="2024-05-20 15:41:17.077128834 -0700 PDT"
15:41:17 DBG recording span span="dagger call --debug echo --msg=hello" id=2998d3b4f5ce8c55
15:41:17 DBG got EOF
15:41:17 DBG got EOF
15:41:17 DBG run finished err=<nil>
```

I think it might be fine to explain that the user will see a lot of DBG
info in the output and then end with the nicely formatted thing that
gets printed out at the end. 

Are you seeing something completely different?

vikram-dagger:
> that the user will see a lot of DBG info in the output and then end with the nicely formatted thing that gets printed out at the end.

@levlaz Yes, I see the DBG output but at the end I only see this output, which only partially matches what you have:

```
✔ connect 1.9s
✔ initialize 58.2s
✔ myModule: MyModule! 2.4s
┃ [DEBUG] dagger.mod._module: output => '{}'                                                                      
┃ [DEBUG] dagger.client._session: Closing client session to GraphQL server                                        
  ✔ connect 0.2s
✔ MyModule.echo(msg: "hello"): String! 2.3s
┃ [DEBUG] dagger.mod._module: output => '"hello"'                                                                 
┃ [DEBUG] dagger.client._session: Closing client session to GraphQL server                                        
  ✔ connect 0.1s

hello
```

Which of the two outputs is correct. @helderco ?

docs/current_docs/manuals/developer/debugging.mdx
```diff
@@ -63,13 +77,4 @@ The above gives a lot of useful information:
 - The user inputs after being converted to more complex types (structuring)
 - The function's result before and after serialization
 
-Additionally, with `--debug` the GraphQL query that the CLI produced is also shown.
```
vikram-dagger, line 66: I also don't see the GraphQL query any longer in the output, so I'm unsure if this statement is still valid. WDYT @levlaz @helderco ?

helderco: The query still shows, but it's no longer at the end, you'll have to search for it.

helderco:
Fyi, to show Python logs you don't need to have `--debug` in the CLI. They're independent.

PR #7408, opened 2024-05-18
docs: Create language-specific pages

vikram-dagger:


docs/current_docs/manuals/developer/python.mdx
```diff
@@ -0,0 +1,98 @@
...
+:::note
+With the TUI, you need to use a progress output that doesn't collapse on success
+like `--progress=plain` or `--debug`, otherwise it won't show in the terminal.
+:::
+
+Using the command `dagger call --debug echo --msg="hello"`, should print
+something like:
+
+```
+✔ MyModule.echo(msg: "hello"): String! 0.5s
```
vikram-dagger, line 75: @helderco the output now contains a very large number of spans, and I wasn't able to find anything similar to what is shown here. Please suggest what should be done for this output.

vikram-dagger:
I'm marking this as draft because it should be merged last, as it removes the Go, Python, TypeScript categories from the sidebar. However it can be reviewed and approved.

vikram-dagger:
Noting that after this merge, the sidebar will be slightly weird because it will have these single pages, plus "Develop with Go/Python/TypeScript" categories. So this PR should be immediately followed by another one to remove those categories, assuming they are fully converted. Alternatively this could be the last PR merged after all other pages are converted, and in this case the sidebar could be updated (after rebase) to remove those categories from the tree.

vikram-dagger:
We decided not to have any language-specific pages. This information will be transferred to common pages in separate PRs.

Issue #7387, opened 2024-05-15
🐞 Dagger CLI v0.11.4 deadlocks when running `dagger call --source=.:default test all`


gerhard:
This was discovered part of:
- https://github.com/dagger/dagger/issues/7339

Somewhat similar to this, but different since in this case host has 16CPUs & no oci max parallelism:
- https://github.com/dagger/dagger/issues/6894

## What is the issue?

When running `dagger call --source=.:default test all` locally the CLI will remain stuck and it will need to be manually killed, as mentioned by @jpadams in [a private thread](https://discord.com/channels/707636530424053791/1239649353015103648). This happens approximately 1 in every 10 runs.

When a timeout is used - [defaults to `30m` in our case](https://github.com/dagger/dagger/blob/2516b3a446804907eb42f3c7e61592b192c6277d/ci/test.go#L73) - this is what the system looks like for that entire period (I have this set to `60m` in my local runs):
<img width="1375" alt="image" src="https://github.com/dagger/dagger/assets/3342/f6dbdef6-c16d-45ae-92a1-be55b72a975f">

This was observed with the standard `v0.11.4` Engine running in Docker:
```console
/ # ps aux | grep '[e]ngine'
    1 root      9:20 /usr/local/bin/dagger-engine --config /etc/dagger/engine.toml --debug
751327 root      4:07 /usr/local/bin/dagger-engine --config /etc/dagger/engine.toml --network-cidr=10.88.0.0/16 --network-name=dagger-dev

/ # cat /etc/dagger/engine.toml

debug = true
trace = false
insecure-entitlements = ["security.insecure"]
```

Notice that there is:
- no `--oci-max-parallelism` set
- no GC custom config

When we hit this deadlocks, sending a `SIGQUIT` to the CLI returns the following: [dagger-cli-v0.11.4-sigquit.txt](https://github.com/dagger/dagger/files/15323848/dagger-cli-v0.11.4-sigquit.txt). To visualise it, https://vito.github.io/swirly/ is most useful - thanks @vito 💪

Highlighting a couple of goroutines that caught my attention:
<img width="1410" alt="image" src="https://github.com/dagger/dagger/assets/3342/d993159c-4277-4faa-8fc3-e0947b560cda">
<img width="1407" alt="image" src="https://github.com/dagger/dagger/assets/3342/b2841745-d771-45cc-a584-a5921e93bffd">
[...]

PR #7416, opened 2024-05-20
ci: Add extra flags for Engine & CLI tests

gerhard:
`--parallel` now defaults to the number of CPUs. Before, this was hard-coded to `16` which was set the number of CPUs in the larger GitHub Runners, the `ubuntu-22.04-16c-64g-600gb` instance size specifically. If this new default value does not work well on the larger CI runners (they have 48CPUs & we run multiple pipelines on them at the same time), we can specify this value explicitly in the workflows.

`--timeout` defaults to `30m`. This was changed not that long ago - the value was set to 1h when using race. The bottom line is that none of our test suites should take more than 10 minutes. This is something that we are actively working towards. For now, until the pipeline becomes more reliable, we will stick with the current 30m default. When running locally with fewer CPUs, or when hunting deadlocks, experience recommends a value of 180m.

`--failfast` is a new one. It sets `--max-fails=1` in `gotestsum`. If a test failed, there is no sense in continuing since the CI will fail and we need to run everything all over again. FTR:
https://github.com/gotestyourself/gotestsum/blob/e9677fb405a5c0dc0b9d7217a7c73c3a20c5cb66/cmd/testdata/gotestsum-help-text#L20

---

Part of:
- https://github.com/dagger/dagger/issues/7387#issuecomment-2120385298

FTR:

```sh
dagger-test-all() {
        local parallel="${1:-16}"
        local timeout="${2:-180m}"
        time dagger call --debug --source=".:default" test all --race=true --failfast=true --parallel=$parallel --timeout=$timeout 2>&1 | ts | tee test.all.race.failfast.parallel${parallel}.timeout${timeout}.$(date "+%Y-%m-%d.%H-%M-%S").txt
}
```


jedevc:
I'm not sure we should set failfast by default - this is very helpful when making large refactorings, and trying to observe all the genuine failures. It's annoying in the case of flakes, or going for a fully green run, but getting all the failures is genuinely useful a lot of the time in my experience.

ci/test.go
```diff
@@ -49,35 +63,62 @@ func (t *Test) Custom(
...
+		parallel = runtime.NumCPU()
+	}
+	args = append(args, fmt.Sprintf("-parallel=%d", parallel))
+
+	// Default timeout to 30m
+	// No test suite should take more than 30 minutes to run
+	if timeout == "" {
+		timeout = "30m"
+	}
+	args = append(args, fmt.Sprintf("-timeout=%s", timeout))
```
jedevc, line 112: We should use `// +default` for this one.

gerhard: If we did this, we would have the default in 3 places:
1. `All()`
2. `Important()`
3. `Custom()`

Since the `race` flag uses the same approach - default in code rather than `// +default` - I think that we should leave this as is. WDYT?

gerhard:
> @jedevc I'm not sure we should set failfast by default - this is very helpful when making large refactorings, and trying to observe all the genuine failures.

Makes sense. Changing the default to `false`.

ci/test.go
jedevc: Hm, alternatively, we should refactor all of these into one function - then which tests to run should probably be flags on in.

Not a blocker though, happy as is.

.github/workflows/engine-and-cli.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     uses: ./.github/workflows/_dagger_call.yml
     secrets: inherit
     with:
-      function: test all --race=true
+      function: test all --race=true --parallel=16
```
jedevc, line 28: Is this different from `NumCPU`s? I think we can leave this unset.

jedevc (approved):

.github/workflows/engine-and-cli.yml
gerhard: Yes, we are explicitly setting this to `16` in CI.

The new CI runners usually have 48 and even 64 CPUs which makes tests fail. Before setting this, checks failed.

gerhard:
Going to merge this since the failing check is due to:
- https://github.com/dagger/dagger/issues/7423

The check is currently showing as failing:
image

But the last run - attempt `#3` - succeeded 🤷‍♂️
image

Issue #7398, opened 2024-05-17
🐞 Don't require any params when asking for help 


levlaz:
### What are you trying to do?

Right now if I want to call --help on a module or function that has some required parameters I need to add those parameters before getting the help output. This is especially painful when using a constructor. 

I expect this to work closer to daggerverse.dev where I can inspect everything a module or function can do without needing to think about the inputs. 

### Why is this important to you?

I think its a much cleaner DX if we decouple --help from execution

helderco:
That's actually a regression, reported in DEV-3785, and actively being worked on.

Issue #7007, opened 2024-04-03
CLI: don't show stderr twice on error


shykes:
## Problem

When `dagger call` returns an error, the contents of the error is printed twice to the screen:

* A first time within the call trace
* A second time at the end of the output

When the call trace is large, or the error message is long, or both, this makes the output very hard to read. The user will often get lost trying to isolate a useful snippet of text to read.

## Solution

Only print the error message once, outside the call trace.

|  | Before | After |
| -- | -- | -- |
| Show error message within call trace? | Yes | No |
| Show error message after call trace? | Yes | Yes |

Here's an example before the change:

```shell
$ dagger call -m github.com/dagger/dagger/sdk/python/runtime base-image
✔ initialize 6.1s
  ✔ ModuleSource.asModule: Module! 0.6s
  ✔ Module.initialize: Module! 0.4s
    ✔ exec go build -o /runtime . 0.3s
    ✔ exec /runtime 0.1s
✘ pythonSdk: PythonSdk! 0.1s
  ✘ exec /runtime 0.1s
  ┃ /src/sdk/python/runtime/internal/dagger/dagger.gen.go:2235
  ┃ main.New(0x0)
  ┃ /src/sdk/python/runtime/main.go:63 +0x10c
  ┃ main.invoke({0x3fe3d0, 0x638e00}, {0x40001824b8, 0x4, 0x8}, {0x40001823f0?, 0x4000104c68?}, {0x0?, 0x40000100e0?}, 0x10?)
  ┃ /src/sdk/python/runtime/dagger.gen.go:800 +0x174
  ┃ main.main()
  ┃ /src/sdk/python/runtime/dagger.gen.go:591 +0x4b0
 
Error: response from query: input: pythonSdk resolve: call constructor: process "/runtime" did not complete successfully: exit code: 2
 
Stderr:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2c023c]
 
goroutine 1 [running]:
python-sdk/internal/dagger.(*Directory).WithoutDirectory(...)
 /src/sdk/python/runtime/internal/dagger/dagger.gen.go:2235
main.New(0x0)
 /src/sdk/python/runtime/main.go:63 +0x10c
[...]

PR #7417, opened 2024-05-20
fix(cli): don't validate flags when requesting `--help`

helderco:
Fixes https://github.com/dagger/dagger/issues/7398

The `--help` flag is now global (interspersed). So the help will no longer be shown next to the command where it’s defined if there’s more to the right. It’s position in the chain no longer matters, it will be the same as adding it to the end. This was needed to look ahead and skip some validations while that flag wasn’t parsed yet, and while it was possible to preserve the previous behavior, making it global seems more consistent with CLI conventions, besides being simpler.

**Before**

```console
❯ dagger call cli --help file
Develop the Dagger CLI

USAGE
  dagger call cli <function>
```

**After**

```console
❯ dagger call cli --help file
Build the CLI binary

USAGE
  dagger call cli file [arguments] <function>
```

## Refactor

Did a bit of refactoring to make the logic easier to follow and encapsulate some responsibilities better in functions.

There’s also some groundwork in preparation for:
- https://github.com/dagger/dagger/issues/6947

### Silence errors

Error showing twice after:
- https://github.com/dagger/dagger/pull/7125

```
Error: required flag(s) "source" not set
Error: required flag(s) "source" not set
exit status 1
```

### Changed telemetry span

The "initialize" telemetry span was being applied to processing constructor flags and building out the next commands, which showed the module loading as failed, and context upload step adding noise to a missing required flag error, for example. So moved only the module initialization to it’s own function to limit the "initialize" span scope.

**Before**

```console
❯ dagger call cli
✘ initialize 12.1s
! required flag(s) "source" not set
  ✔ Module.initialize: Module! 1.4s
  ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.6s
    ✔ upload /Users/.../dagger from xxx (client id: oq8yf7p509lm3o9qqhden90vx) (exclude: **/.git) (include: ci/**/*, **/go.mod, **/go.sum, **/go.work, **/go.work.sum, **/vendor/, **/*.go, dagger.json) 0.6s

Error: required flag(s) "source" not set
```

**After**

```console
❯ dagger call version string

✔ initialize 6.7s
✘ prepare 0.0s
! required flag(s) "source" not set

Error: required flag(s) "source" not set
Run 'dagger call --help' for usage.
exit status 1
```

### Removed aliases

When they were added, the API returned the original name used by the SDK. Somewhere along the way, to avoid the need for SDKs to handle naming conventions, the engine started to store the original name under a private field, and converted name to the normalized GraphQL convention for fields. So, for a while now, the aliases haven’t been the name that users see in code, which was why we added them in the first place.

### DIFF
--- a/.changes/unreleased/Fixed-20240520-144206.yaml
+++ b/.changes/unreleased/Fixed-20240520-144206.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: "cli: don't validate flags when requesting `--help`"
+time: 2024-05-20T14:42:06.270495Z
+custom:
+  Author: helderco
+  PR: "7417"
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -185,8 +185,9 @@ type FuncCommand struct {
 	// showUsage flags whether to show a one-line usage message after error.
 	showUsage bool
 
-	q *querybuilder.Selection
-	c *client.Client
+	q   *querybuilder.Selection
+	c   *client.Client
+	ctx context.Context
 }
 
 func (fc *FuncCommand) Command() *cobra.Command {
@@ -382,17 +383,20 @@ func (fc *FuncCommand) initializeModule(ctx context.Context) (rerr error) {
 }
 
 // loadCommand finds the leaf command to run.
-func (fc *FuncCommand) loadCommand(c *cobra.Command, a []string) (*cobra.Command, []string, error) {
+func (fc *FuncCommand) loadCommand(c *cobra.Command, a []string) (rcmd *cobra.Command, rargs []string, rerr error) {
 	// If a command implements Execute, it doesn't need to build and
 	// traverse the command tree.
 	if fc.Execute != nil {
 		return c, nil, nil
 	}
 
-	builder := fc.cobraBuilder(
-		c.Context(),
-		fc.mod.MainObject.AsObject.Constructor,
-	)
+	ctx := c.Context()
+
+	spanCtx, span := Tracer().Start(ctx, "prepare", telemetry.Encapsulate())
+	defer telemetry.End(span, func() error { return rerr })
+	fc.ctx = spanCtx
+
+	builder := fc.cobraBuilder(ctx, fc.mod.MainObject.AsObject.Constructor)
 
 	cmd, args, err := fc.traverse(c, a, builder)
 	if err != nil {
@@ -459,7 +463,7 @@ func (fc *FuncCommand) cobraBuilder(ctx context.Context, fn *modFunction) func(*
 		}
 
 		// Easier to add query builder selections as we traverse the command tree.
-		return fc.selectFunc(ctx, fn, c, dag)
+		return fc.selectFunc(fn, c, dag)
 	}
 }
 
@@ -599,7 +603,7 @@ func (fc *FuncCommand) makeSubCmd(ctx context.Context, dag *dagger.Client, fn *m
 }
 
 // selectFunc adds the function selection to the query.
-func (fc *FuncCommand) selectFunc(ctx context.Context, fn *modFunction, cmd *cobra.Command, dag *dagger.Client) error {
+func (fc *FuncCommand) selectFunc(fn *modFunction, cmd *cobra.Command, dag *dagger.Client) error {
 	fc.Select(fn.Name)
 
 	for _, arg := range fn.SupportedArgs() {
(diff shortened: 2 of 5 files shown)


nipuna-perera:
Hey @helderco I have a request. It's not necessary to be done as part of this PR but I wanted to bring it to the team's attention. I am trying to run some dagger functions in parallel within my CI system (Jenkins). Since the function calls are the first thing dagger related that happens in the pipeline, I have a separate stage to run `dagger call --help` to "initialize" the engine (basically pull the image and get it ready). 

I see this same pattern in the dagger CI too. There is a "Wait for engine to be ready" step in the actions.

Now the issue is, my modules are in a `daggerverse` folder and I when I do `dagger call --help` I have to pass in a module like `dagger call -m ./mymodule --help`. Which isn't that ergonomic. Is there a better way to get the engine to "initialize" without calling `--help`? If not, can you make the `dagger call --help` not return a nonzero when it's outside of a module folder?

helderco:
Hey @nipuna-perera, you'll be able to do that with:
- https://github.com/dagger/dagger/pull/7310

Then I'd suggest you call the `version` core function:

```console
❯ dagger -m- call version
v0.11.5
```

That has much less output but still informative and it can be run from anywhere. Doesn't need to load any module.

nipuna-perera:
> Hey @nipuna-perera, you'll be able to do that with:
> 
> * [cli: allow calling core functions directly #7310](https://github.com/dagger/dagger/pull/7310)
> 
> Then I'd suggest you call the `version` core function:
> 
> ```
> ❯ dagger -m- call version
> v0.11.5
> ```
> 
> That has much less output but still informative and it can be run from anywhere. Doesn't need to load any module.

That is perfect! Thank you! Looking forward to these DX cleanup/improve PRs getting merged :)

jedevc:
> The "initialize" telemetry span was being applied to processing constructor flags and building out the next commands, which showed the module loading as failed, and context upload step adding noise to a missing required flag error, for example. So moved only the module initialization to it’s own function to limit the "initialize" span scope.

Something in this change seems to effectively revert https://github.com/dagger/dagger/pull/7347

I think this is because all cli args were moved to be part of that span - including user args. I'm *not* sure what the answer is here, ideally we should keep the spans for loading these attached to the intitialize context, but I agree that we shouldn't unfold this on an error.

*Potentially*, we should only error the entire span if there's a loading error, but on a required flag error, the span shouldn't error - initialization has succeeded, and we can propagate the error back up, but not error the span.

helderco:
> > The "initialize" telemetry span was being applied to processing constructor flags and building out the next commands, which showed the module loading as failed, and context upload step adding noise to a missing required flag error, for example. So moved only the module initialization to it’s own function to limit the "initialize" span scope.
> 
> Something in this change seems to effectively revert #7347

Thanks for catching that, I'll see what I can do.

helderco:
If I put in another span it's cleaner. 

**Before**

```console
❯ dagger call version string

✘ initialize 6.9s
! required flag(s) "source" not set
  ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.7s
    ✔ upload /Users/helder/Projects/dagger from Sirius (client id: mr82245cdk074u11yv694vty4) (exclude: **/.git) (include: **/*.go, dagger.json, ci/**/*, **/go.mod, **/go.sum, **/go.work, **/go.work.sum, **/vendor/) 0.7s

Error: required flag(s) "source" not set
```


**After**

```console
❯ dagger call version string

✔ initialize 6.7s
✘ prepare 0.0s
! required flag(s) "source" not set

Error: required flag(s) "source" not set
Run 'dagger call --help' for usage.
```

Don't love the repetition that I wasn't getting before:

```console
❯ dagger call version string

✔ initialize 4.8s

Error: required flag(s) "source" not set
Run 'dagger call --help' for usage.
exit status 1
```

But we already have an issue on that:
- https://github.com/dagger/dagger/issues/7007

helderco:
One thing to note is that most (probably, haven't counted) cli args will show at the root anyway, rather than this span. Even for a `Directory` arg, if it's from git, the flag returns a lazy value. Some flags return a lazy value, others don't. So at this point, I'm not sure of the value of putting them in their own span. Whatever you see above the main pipeline is just the inputs to that pipeline. That's concistent with the general feel in the Cloud UI.

jedevc:
As discussed (note to self):
- We'll use with `prepare` that's marked as encapsulate	
	- Eventually, we could have a top-level span for *each* flag
- We might unlazy git arguments to get them to show at the right top-level span, instead of later on

jedevc (approved):

helderco:
@wingyplus, any idea why the Elixir test is failing?

Issue #6608, opened 2024-02-07
✨ CLI: Don't show functions that can't be called


marcosnils:
### What are you trying to do?

I'm writing a module that exposes a function with complex arguments which are not currently supported via the `dagger call` command. I would really like that function not to be callable or displayed in the `dagger functions` given that it's only intended to use via an SDK and not the CLI.

### Why is this important to you?

To provide a clearer API to my module users.

sipsma:
I'll tentatively add to v0.10 milestone since I think it's likely pretty trivial implement and imo makes perfect sense to do

PR #7418, opened 2024-05-20
cli: don’t show functions that can’t be called

helderco:
Fixes https://github.com/dagger/dagger/issues/6608

Most common reason for not being able to call a function is unsupported flags. A function will be skipped if there's any required arguments with unsupported flags, but if the unsupported flags are only for optional arguments, they just won't be available to the CLI.

### DIFF
--- a/cmd/dagger/flags.go
+++ b/cmd/dagger/flags.go
@@ -24,6 +24,19 @@ import (
 	"dagger.io/dagger"
 )
 
+type UnsupportedFlagError struct {
+	Name string
+	Type string
+}
+
+func (e *UnsupportedFlagError) Error() string {
+	msg := fmt.Sprintf("unsupported type for flag --%s", e.Name)
+	if e.Type != "" {
+		msg = fmt.Sprintf("%s: %s", msg, e.Type)
+	}
+	return msg
+}
+
 // GetCustomFlagValue returns a pflag.Value instance for a dagger.ObjectTypeDef name.
 func GetCustomFlagValue(name string) DaggerValue {
 	switch name {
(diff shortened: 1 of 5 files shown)


helderco:
In draft because I think it's best to not exclude functions that only have optional arguments as unsupported flags.

jedevc (approved):

cmd/dagger/flags.go
```diff
@@ -629,115 +642,138 @@ func (v *platformValue) Get(ctx context.Context, dag *dagger.Client, _ *dagger.M
...
-			return val, nil
+			return nil
 		}
 
 		// TODO: default to JSON?
-		return nil, fmt.Errorf("unsupported object type %q for flag: %s", objName, name)
+		return &UnsupportedFlagError{
+			Name: name,
+			Type: fmt.Sprintf("%q object", objName),
+		}
```
gerhard, line 693: Does the new return type address the `TODO: default to JSON?`

cmd/dagger/flags.go
```diff
@@ -629,115 +642,138 @@ func (v *platformValue) Get(ctx context.Context, dag *dagger.Client, _ *dagger.M
...
-			return val, nil
+			return nil
 		}
 
 		// TODO: default to JSON?
-		return nil, fmt.Errorf("unsupported input type %q for flag: %s", inputName, name)
+		return &UnsupportedFlagError{
+			Name: name,
+			Type: fmt.Sprintf("%q input", inputName),
+		}
```
gerhard, line 707: Idem https://github.com/dagger/dagger/pull/7418/files#r1609990441

cmd/dagger/functions.go
```diff
@@ -556,8 +556,11 @@ func (fc *FuncCommand) addArgsForFunction(cmd *cobra.Command, cmdArgs []string,
 	}
 
 	for _, arg := range fn.Args {
-		_, err := arg.AddFlag(cmd.Flags())
-		if err != nil {
+		if err := arg.AddFlag(cmd.Flags()); err != nil {
+			var e *UnsupportedFlagError
+			if errors.As(err, &e) {
```
marcosnils, line 561: shall we print something in the CLI output to warn the user why that function is not being displayed? I'm thinking that some might be confused if they're defining functions with unsupported flags.

marcosnils (approved):
LGTM overall, left a small nit.

cmd/dagger/functions.go
helderco: That was my plan, yes, as commented in DEV-3613:

> It may be confusing for some users to create a function that doesn't show. But maybe we can add a summary at the end like "Skipped 13 unsupported functions", even if only on debug or something.

However, when making the changes to `module.go` I got concerned about having to "leak" that information from those functions somehow in order to present later at `--help`, in `FuncCommand`. With missing arguments it's doable since the function itself isn't skipped.

Also concerned about polluting `--help` as another positive for this feature is limiting the amount of core functions that get listed in `--help`, although it would add one line, instead of one for each skipped. 

Neither of those are blockers if it's the right thing to do but I'm wondering if using debug logs could be enough (still a challenge for skipped functions though). Didn't come back to revisit this, wdyt?

cmd/dagger/flags.go
helderco: No, feature wise it's the same. I just replaced a generic error with a custom one in order to tell the difference later.

PR #7421, opened 2024-05-20
cli: list available functions when chain ends in an object

helderco:
Fixes https://github.com/dagger/dagger/issues/6515

This allows for a much **nicer feedback loop** while you navigate the command tree. It also makes `dagger function` redundant.[^1]

[^1]: If we deprecate `dagger function` it also solves an issue with chains that end in a leaf, which return an error that there are no more functions available. With `dagger call` it just executes it.

### Before

```console
❯ dagger call cli file
File evaluated. Use "dagger call cli file --help" to see available sub-commands.
```

**Waited a long time** while the `File` was evaluated, to end in a usage message.

### After

**Fast response** because no query is executed (and more informative):

```console
❯ dagger call cli file
AVAILABLE FUNCTIONS
  contents          Retrieves the contents of the file.
  export            Writes the file to a file path on the host.
  name              Retrieves the name of the file.
  size              Retrieves the size of the file, in bytes.
  sync              Force evaluation in the engine.
  with-timestamps   Retrieves this file with its created/modified timestamps set to the given time.

Use 'dagger call cli file <function> --help' for more information about a function.
```

> [!NOTE]
> This removes `sync` as a default selection for commands ending in a `Container`, `Directory` or `File`.   \cc @vikram-dagger for impact on docs.

### DIFF
--- a/.changes/unreleased/Changed-20240520-155221.yaml
+++ b/.changes/unreleased/Changed-20240520-155221.yaml
@@ -1,0 +1,6 @@
+kind: Changed
+body: 'cli: list available functions when chain ends in an object'
+time: 2024-05-20T15:52:21.507104Z
+custom:
+  Author: helderco
+  PR: "7421"
--- a/cmd/dagger/call.go
+++ b/cmd/dagger/call.go
@@ -49,13 +49,9 @@ dagger call lint stdout
 				if name == File {
 					c.Arg("allowParentDirPath", true)
 				}
-				return nil
 			}
-			c.Select("sync")
 		case Terminal:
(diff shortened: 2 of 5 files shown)


shykes:
Sorry but I still prefer the other option discussed in #6515: actually printing the contents. I understand the desire for a better discovery flow, `--help` is not ideal. But when I type `dagger call FUNC`, I expect `FUNC` to be... called.

I like the feature in this PR but think it needs another home than `dagger call`.

helderco:
Closing this in favor of:
- https://github.com/dagger/dagger/pull/7479

PR #7425, opened 2024-05-20
docs: Add common runtimes page

vikram-dagger:


docs/current_docs/manuals/developer/runtimes.mdx
levlaz, line 1: This is a bit confusing because first we say it supports node, deno and bun and then we say it only supports node and bun, is this out of date?

docs/current_docs/manuals/developer/runtimes.mdx
```diff
@@ -17,7 +21,7 @@ For now, the TypeScript SDK only supports Node.js (stable) and Bun (experimental
 
 To change the TypeScript SDK runtime, set the field `dagger.runtime` in the `package.json` file.
```
levlaz, line 22: Can we specify which package.json we mean, are we talking about the project level one or the one that we generate inside of `dagger/package.json` when running `init --sdk=typescript`

levlaz (approved):
Overall this is good, I really like this format over a language specific page. 

I made two comments but they do not necessarily need to block this PR since the structure is perfect IMO.

docs/current_docs/manuals/developer/runtimes.mdx
TomChv: It's the one at the root of the repo, not in `sdk/` since it's ignored by vsc

docs/current_docs/manuals/developer/runtimes.mdx
TomChv: We don't say that we support node, deno and bun, we say that "_TypeScript_" (the language) supports multiples runtimes, but we only support 2 of them. Maybe we can change the sentence to be more clear

> TypeScript is supported by multiple runtimes: [Node.js](https://nodejs.org/en) (via `tsx`), [Deno](https://deno.com/) (natively) and [Bun](https://bun.sh/) (natively).

docs/current_docs/manuals/developer/runtimes.mdx
vikram-dagger: Added, please check

TomChv (approved):

PR #7434, opened 2024-05-21
fix: throw error on invalid module registration

### LINKED ISSUE #6624: 🐞 Typescript module dependency not included in dagger.gen.go
This is happening on main, but not on version 0.9.8

Repro: 

```sh
git clone -b daggerize https://github.com/kpenfound/example-voting-app
cd example-voting-app
dagger functions
```

The module has 4 local module dependencies and 1 remote. Of the local dependencies, there are 2 Go modules (Seed, Worker), 1 Python module (Vote), and 1 Typescript module (Result).

The codegen is not including Result for some reason, and any call results in:
`dag.Result undefined (type *Client has no field or method Result)`

TomChv:
Fixes #6624 

Improve error message clarity.
Remove non-required `await`


gerhard (approved):
💪 🚀

PR #7437, opened 2024-05-22
feat: support`AllowWildCard` and `AllowNotFound`

### LINKED ISSUE #3338: `withoutDirectory` and `withoutFile` should support `AllowWildcard` and `AllowNotFound`
Context: https://github.com/dagger/dagger/pull/3330#discussion_r993827447

TomChv:
Fixes #3338

⚠️ I'm hitting an issue locally when I try to test this feature, it seems my new params are not taken into account.
I'm doing a push to see if I hit the same behaviour in the CI (maybe it's due to some reloading or cache in the API)


core/schema/directory.go
```diff
@@ -55,7 +55,9 @@ func (s *directorySchema) Install() {
 			ArgDoc("permissions", `Permission given to the copied file (e.g., 0600).`),
 		dagql.Func("withoutFile", s.withoutFile).
 			Doc(`Retrieves this directory with the file at the given path removed.`).
-			ArgDoc("path", `Location of the file to remove (e.g., "/file.txt").`),
+			ArgDoc("path", `Location of the file to remove (e.g., "/file.txt").`).
+			ArgDoc("allowWildCard", `Allow wildcards in the path (e.g., "*.txt").`).
+			ArgDoc("allowNotFound", `Allow the operation to not fail if the directory does not exist.`),
```
jedevc, line 60: FYI - we would also want this on `container`. Side note (and not sure how to resolve this), it feels weird that we have so much repetition with the 4 methods there.

TomChv:
Based in this discussion: https://discord.com/channels/707636530424053791/1240714480275689574/1242807626723491921 we're going to hold this PR for the next release and find a better API design.

There's still that strange issue where my parameter are not considered, even in the issue, I'm not sure what's the root cause.

/cc @vito @sipsma @helderco

core/directory.go
```diff
@@ -677,16 +677,21 @@ func (dir *Directory) Diff(ctx context.Context, other *Directory) (*Directory, e
...
 		return nil, err
 	}
 
+	fmt.Println("path", path)
+	fmt.Println("allowNotfound", allowNotFound)
+	fmt.Println("allowWildCard", allowWildCard)
+
+
 	path = filepath.Join(dir.Dir, path)
-	err = dir.SetState(ctx, st.File(llb.Rm(path, llb.WithAllowWildcard(true), llb.WithAllowNotFound(true))))
```
jedevc, line 689: Huh. I didn't know we defaulted to this.

This feels *slightly* inconsistent with other stuff - see another discussion on magic processing in https://github.com/dagger/dagger/pull/7173. A bit strange that some paths can be globs/wildcards, and some aren't.

TomChv: Yes definitely, I think that's why we wanted to add the parameter.

core/integration/directory_test.go
```diff
@@ -521,6 +521,13 @@ func TestDirectoryWithoutDirectoryWithoutFile(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, []string{"some-dir", "some-file"}, entries)
 
+	// Test without dir with allow not found = false
+	_, err = dir1.
+		WithoutDirectory("non-existant", dagger.DirectoryWithoutDirectoryOpts{AllowNotFound: false}).
```
jedevc, line 526: @TomChv this is your issue.

In go, the empty struct defaults to `false` - therefore, we have the API constraint today that all boolean options must default to false.

What you need is to invert the option, to something like `MustExist: true`, so that the default is `false`.

PR #7438, opened 2024-05-22
feat: allow sending arbitrary signals to running service

### LINKED ISSUE #7391: `Service.Stop` should allow passing an arbitrary linux signal
While debugging https://github.com/dagger/dagger/issues/7387, it would be useful if on a test timeout failure we could send the test dagger engine an explicit SIGQUIT signal to dump a stack trace.

This would make the engine logs a little more helpful, and would allow for a full example to test.

jedevc:
Fixes #7391

This started as just adding a new signals enum, and allowing it to be attached to `stop`. However, I realized - not every signal should necessarily cause a process to terminate, so we should also probably have a way of signalling a process, without waiting for that to happen. So I also added a new `signal` method.

Bit of a jumble of commits, see each one for more details.


cmd/codegen/generator/typescript/templates/src/method_solve.ts.gtpl
```diff
@@ -77,8 +77,9 @@
 
       		{{- with $optionals }}
       			{{- if $required }}, {{ end }}
-				{{- "" }}...opts{{- if gt (len $enums) 0 -}}, __metadata: metadata{{- end -}}
+				{{- "" }}...opts
 			{{- end }}
+      {{- if gt (len $enums) 0 -}}, __metadata: metadata{{- end -}}
```
jedevc, line 82: cc @TomChv - I *think* this is right.

Previously, we never had any required enum arguments. When this patch added them, then the defined `metadata` would never be attached (which caused a linting/codegen failure). I think this is the right fix, but just wanted to make sure.

Side note: this file is *weirdly* formatted, and uses an inconsistent mixture of tabs and spaces, which makes it fairly challenging to view/edit. Didn't want to fix this right in this PR to avoid a massive diff, but we should fix this.

TomChv: > Side note: this file is weirdly formatted, and uses an inconsistent mixture of tabs and spaces, which makes it fairly challenging to view/edit. Didn't want to fix this right in this PR to avoid a massive diff, but we should fix this.

Yeah this part needs a refactor but is not critical enough to be a priority

TomChv (approved):

gerhard:
WDYT @jedevc, does this feel like patch change or do we want to push it into `v0.12.0` ?

PR #7440, opened 2024-05-22
wip: windows provision job

jedevc:


jedevc:
Sigh.

- minikube's hyperv driver just hangs forever waiting for an IP - this seems like an often reported issue on minikube's github, with no answer, it just gets autoclosed.
	- The VM is *running* - but no IP yay.
- the QEMU driver isn't really feasible here either, no idea why it's suggested with no instructions (I guess "experimental")
- installing docker desktop in CI doesn't seem to be at all trivial.
- WSL is a no for completely non-apparent reasons.

Does no one do this? How are people testing this kind of flow?

sipsma:
@jedevc Yeah this is shockingly complicated; I run into the same sort of problems because I run windows in a VM. For that I setup an insane rube-goldberg machine of tailscale+socat proxies to my linux machine's docker socket (which is not a good idea for CI lol)

Given how painful our lack of testing on Windows is becoming, maybe we should "just" setup some pet windows machines and throw a GHA runner on them (though I'm sure "just" is doing a lot of work there). Can just run on main for now while we sort it out, but I'd take a setup like that over not having any tests on windows I think.

PR #7445, opened 2024-05-22
Fix powershell install script to work when on a different drive

angrybat:
Changed the default install path from $env:HOMEPATH to $env:USERPROFILE as this includes the drive. Currently when calling the install script on the D: drive the transfer from the temp directory to the install path fails. This is not reported as an error and the install script appears to have worked correctly. This is because $env:HOMEPATH does not include the drive in the path. Changing this to $env:USERPROFILE has the same path but the path includes the drive, this fixes the issue.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -3,7 +3,7 @@
 param (
     [Parameter(Mandatory = $false)] [System.Management.Automation.SemanticVersion]$DaggerVersion,
     [Parameter(Mandatory = $false)] [string]$DaggerCommit,
-    [Parameter(Mandatory = $false)] [string]$InstallPath = $env:HOMEPATH + '\dagger',
+    [Parameter(Mandatory = $false)] [string]$InstallPath = $env:USERPROFILE + '\dagger',
 
     [Parameter(Mandatory = $false)] [System.Boolean]$InteractiveInstall = $false
 )


jedevc:
LGTM to me once DCO is added, looks like `USERDRIVE` is the generally preferred variable to use instead of `HOMEPATH` these days as well.

Thanks! :tada:

angrybat:
@jedevc Hopefully I have signed the commit now. Sorry for the delay but I am happy to help where I can 😄

jedevc:
@angrybat it looks like it's still not done, sorry! Essentially, you should just need to run `git commit --amend -s` on this branch.

Otherwise, I can manually sign it off for you, if you agree with that.

Sorry for the delay, thanks again for this PR, we just need to have the sign-offs so we can ensure that we actually have permission to *use* all the contributed code in this project.

jedevc:
Rebased.

I've also signed off the commit so we can bring this in, since @angrybat attempted to here: https://github.com/dagger/dagger/pull/7445#issuecomment-2132104128:

> @jedevc Hopefully I have signed the commit now. Sorry for the delay but I am happy to help where I can 😄

jedevc (approved):

PR #7451, opened 2024-05-23
fix(core): check error on `normalizeCallerLoadedSource`

grouville:
`normalizeCallerLoadedSource` function can segfault if `withContextDirectory` returns an error, as we do not check for the returned error on the `withContextDirectory` call.

The following dagql calls can be executed: on the `dagger init`, for example, the `withName` gets called on my tests, and trigger the segfault.

It is currently (almost) never triggered, as there are two potential edge cases in which we return an error on `withContextDirectory`:
1. when the modulesourcekind is not local
2. when we fail to load the dir

### DIFF
--- a/core/schema/modulesource.go
+++ b/core/schema/modulesource.go
@@ -805,6 +805,9 @@ func (s *moduleSchema) normalizeCallerLoadedSource(
 			},
 		},
 	)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


marcosnils (approved):

vito (approved):

sipsma (commented):
Good catch, thanks!

PR #7452, opened 2024-05-23
OTel cleanups + better draining

### LINKED ISSUE #7048: Dagger call hangs on telemetry draining
To reproduce on `main` (can also repro against `v0.11.0` using the new modularized ci on main):

```
$ dagger call --source=. sdk go generate entries
sdk
(hangs here, even though the task is complete)
```

Dumping trace logs on the engine:

```
time="2024-04-08T11:58:34Z" level=trace msg="server removed" client_call_digest= client_hostname=daggerdoer client_id=jui8v1z5en9a42r1l9sncx1zp server_id=cl0so4twc5tpx6ggse1v4rpxa
time="2024-04-08T11:58:34Z" level=debug msg="session call done" client_call_digest= client_hostname=daggerdoer client_id=jui8v1z5en9a42r1l9sncx1zp register_client=true server_id=cl0so4twc5tpx6ggse1v4rpxa
time="2024-04-08T11:58:34Z" level=debug msg="exporting spans to pubsub" call=uvo90qb6cll202ijuzbgxa7ww spans=1
time="2024-04-08T11:58:34Z" level=debug msg="pubsub exporting span" call=uvo90qb6cll202ijuzbgxa7ww endTime="2024-04-08 11:58:34.317715668 +0000 UTC" id=6ce54437520d9481 span=moby.buildkit.v1.Control/Session status=Error trace=ad1f63c1d21669763c9995a867fa80ce
time="2024-04-08T11:58:34Z" level=debug msg="exporting spans to subscriber" spans=1 trace=ad1f63c1d21669763c9995a867fa80ce
time="2024-04-08T11:58:34Z" level=debug msg="waiting for spans" activeSpans=1 draining=true immediate=false trace=ad1f63c1d21669763c9995a867fa80ce
time="2024-04-08T11:58:34Z" level=debug msg="waiting for span" activeSpans=1 draining=true id=63b55c6f67dff984 immediate=false span=moby.buildkit.v1.Control/Session trace=ad1f63c1d21669763c9995a867fa80ce
```

The `moby.buildkit.v1.Control/Session` never seems to complete, so the client never seems to finish telemetry: https://github.com/dagger/dagger/blob/b583addd9d9588054dc22334363107aee2e753e2/engine/client/client.go#L471-L473

Not quite sure what's special about the `sdk go generate` job specifically, but it seems to consistently cause this hang on v0.11.0+. cc @vito, any ideas?

### LINKED ISSUE #7510: 🐞 New `--plain=progress` implementation swallows prints and logs from modules
### What is the issue?

The new `--plain=progress` implementation shiped with v0.11.5 seems to swallow print statements and logs from modules among a lot of helpful debugging information (i.e. the command called).

### Dagger version

dagger v0.11.5 (registry.dagger.io/engine) darwin/arm64

### Steps to reproduce

Take the following example:

```python
import logging

from dagger import function, object_type
from dagger.log import configure_logging

configure_logging(logging.DEBUG)
logger = logging.getLogger("dagger")


@object_type
class Test:
    @function
    def test(self):
        logger.warning("This is a warning using the logger")
        print("This is a print statement")
```

Running the above with v0.11.4 resulted in the following on stderr:

```plain
1: in exec docker start dagger-engine-835668aa455cbb8f
1: dagger-engine-835668aa455cbb8f

Connected to engine 08d250bad272 (version v0.11.4)
2: in starting session
2: OK!

3: in starting session
3: OK!

4: in starting session
4: OK!

5: in Query.test
5: [DEBUG] dagger.client._session: Configuring shared connection to GraphQL server
5: [DEBUG] dagger.client._session: Establishing client session to GraphQL server
5: [DEBUG] dagger.telemetry: Found TRACEPARENT
5: [DEBUG] dagger.telemetry: Initializing telemetry
5: [DEBUG] dagger.telemetry: Initializing traces telemetry with exporters: otlp_proto_grpc
5: [DEBUG] dagger.telemetry: Initializing metrics telemetry with exporters: none
5: [DEBUG] dagger.telemetry: Initializing logs telemetry with exporters: none
5: [DEBUG] dagger.telemetry: Telemetry initialized
5: [DEBUG] dagger.mod._module: invoke => {'parent_name': 'Test', 'parent_json': 'null', 'name': '', 'input_args': '{}'}
5: [DEBUG] dagger.mod._module: resolver => Test()
5: [DEBUG] dagger.mod._resolver: func => <Signature () -> None>
5: [DEBUG] dagger.mod._resolver: input args => {}
5: [DEBUG] dagger.mod._resolver: structured args => {}
5: [DEBUG] dagger.mod._module: result => Test()
5: [DEBUG] dagger.mod._module: output => '{}'
5: [DEBUG] dagger.client._session: Closing client session to GraphQL server

6: in starting session
6: OK!

7: in Test.test
7: [DEBUG] dagger.client._session: Configuring shared connection to GraphQL server
7: [DEBUG] dagger.client._session: Establishing client session to GraphQL server
7: [DEBUG] dagger.telemetry: Found TRACEPARENT
7: [DEBUG] dagger.telemetry: Initializing telemetry
7: [DEBUG] dagger.telemetry: Initializing traces telemetry with exporters: otlp_proto_grpc
7: [DEBUG] dagger.telemetry: Initializing metrics telemetry with exporters: none
7: [DEBUG] dagger.telemetry: Initializing logs telemetry with exporters: none
7: [DEBUG] dagger.telemetry: Telemetry initialized
7: [DEBUG] dagger.mod._module: invoke => {'parent_name': 'Test', 'parent_json': '{}', 'name': 'test', 'input_args': '{}'}
7: [DEBUG] dagger.mod._module: resolver => Test.test
7: [DEBUG] dagger.mod._resolver: func => <Signature ()>
7: [DEBUG] dagger.mod._resolver: input args => {}
7: [DEBUG] dagger.mod._resolver: structured args => {}
7: [WARNING] dagger: This is a warning using the logger
7: This is a print statement
7: [DEBUG] dagger.mod._module: result => None
7: [DEBUG] dagger.mod._module: output => 'null'
7: [DEBUG] dagger.client._session: Closing client session to GraphQL server

8: in /Users/[redacted]/Downloads/dagger_v0.11.4_darwin_arm64/dagger --progress=plain call test
8: <nil>

✔ connect 1.1s
✔ initialize 0.5s
✔ test: Test! 0.5s
┃ [DEBUG] dagger.client._session: Configuring shared connection to GraphQL server                                                                                                                                                                                                                           
┃ [DEBUG] dagger.client._session: Establishing client session to GraphQL server                                                                                                                                                                                                                             
┃ [DEBUG] dagger.telemetry: Found TRACEPARENT                                                                                                                                                                                                                                                               
┃ [DEBUG] dagger.telemetry: Initializing telemetry                                                                                                                                                                                                                                                          
┃ [DEBUG] dagger.telemetry: Initializing traces telemetry with exporters: otlp_proto_grpc                                                                                                                                                                                                                   
┃ [DEBUG] dagger.telemetry: Initializing metrics telemetry with exporters: none                                                                                                                                                                                                                             
┃ [DEBUG] dagger.telemetry: Initializing logs telemetry with exporters: none                                                                                                                                                                                                                                
┃ [DEBUG] dagger.telemetry: Telemetry initialized                                                                                                                                                                                                                                                           
┃ [DEBUG] dagger.mod._module: invoke => {'parent_name': 'Test', 'parent_json': 'null', 'name': '', 'input_args': '{}'}                                                                                                                                                                                      
┃ [DEBUG] dagger.mod._module: resolver => Test()                                                                                                                                                                                                                                                            
┃ [DEBUG] dagger.mod._resolver: func => <Signature () -> None>                                                                                                                                                                                                                                              
┃ [DEBUG] dagger.mod._resolver: input args => {}                                                                                                                                                                                                                                                            
┃ [DEBUG] dagger.mod._resolver: structured args => {}                                                                                                                                                                                                                                                       
┃ [DEBUG] dagger.mod._module: result => Test()                                                                                                                                                                                                                                                              
┃ [DEBUG] dagger.mod._module: output => '{}'                                                                                                                                                                                                                                                                
┃ [DEBUG] dagger.client._session: Closing client session to GraphQL server                                                                                                                                                                                                                                  
✔ Test.test: Void 0.6s
┃ [DEBUG] dagger.client._session: Configuring shared connection to GraphQL server                                                                                                                                                                                                                           
┃ [DEBUG] dagger.client._session: Establishing client session to GraphQL server                                                                                                                                                                                                                             
┃ [DEBUG] dagger.telemetry: Found TRACEPARENT                                                                                                                                                                                                                                                               
┃ [DEBUG] dagger.telemetry: Initializing telemetry                                                                                                                                                                                                                                                          
┃ [DEBUG] dagger.telemetry: Initializing traces telemetry with exporters: otlp_proto_grpc                                                                                                                                                                                                                   
┃ [DEBUG] dagger.telemetry: Initializing metrics telemetry with exporters: none                                                                                                                                                                                                                             
┃ [DEBUG] dagger.telemetry: Initializing logs telemetry with exporters: none                                                                                                                                                                                                                                
┃ [DEBUG] dagger.telemetry: Telemetry initialized                                                                                                                                                                                                                                                           
┃ [DEBUG] dagger.mod._module: invoke => {'parent_name': 'Test', 'parent_json': '{}', 'name': 'test', 'input_args': '{}'}                                                                                                                                                                                    
┃ [DEBUG] dagger.mod._module: resolver => Test.test                                                                                                                                                                                                                                                         
┃ [DEBUG] dagger.mod._resolver: func => <Signature ()>                                                                                                                                                                                                                                                      
┃ [DEBUG] dagger.mod._resolver: input args => {}                                                                                                                                                                                                                                                            
┃ [DEBUG] dagger.mod._resolver: structured args => {}                                                                                                                                                                                                                                                       
┃ [WARNING] dagger: This is a warning using the logger                                                                                                                                                                                                                                                      
┃ This is a print statement                                                                                                                                                                                                                                                                                 
┃ [DEBUG] dagger.mod._module: result => None                                                                                                                                                                                                                                                                
┃ [DEBUG] dagger.mod._module: output => 'null'                                                                                                                                                                                                                                                              
┃ [DEBUG] dagger.client._session: Closing client session to GraphQL server                                                                                                                                                                                                                                  

<nil>   
```

Please note that both, the print statement as well as the log message are printed to stderr.

The same code with v0.11.5 however results in:

```plain
1   : connect
2   :   starting engine
2   :   starting engine DONE [0.8s]
3   :   connecting to engine
18:51:09 INF Connected to engine name=d00ca9e72672 version=v0.11.5
3   :   connecting to engine DONE [0.1s]
4   :   starting session
4   :   starting session DONE [0.2s]
1   : connect DONE [1.1s]

5   : initialize
5   : initialize DONE [0.3s]

6   : test: Test!
7   :   exec /runtime
7   :   [0.3s] | [DEBUG] dagger.client._session: Configuring shared connection to GraphQL server
7   :   [0.3s] | [DEBUG] dagger.client._session: Establishing client session to GraphQL server
6   : test: Test! DONE [0.5s]

8   : Test.test: Void
9   :   exec /runtime
9   :   [0.3s] | [DEBUG] dagger.client._session: Configuring shared connection to GraphQL server
9   :   [0.3s] | [DEBUG] dagger.client._session: Establishing client session to GraphQL server
8   : Test.test: Void DONE [0.5s]
```

Neither the print statement nor the log message are printed.

### Log output

_No response_

vito:
fixes #7510 
fixes #7048 

* Adopt upstream logging SDK (still WIP - `otelloggrpc` is a noop for example)
* De-duplicate `github.com/dagger/dagger/telemetry` and `sdk/go/telemetry` (no longer `internal/`)
  * `dagger.io/dagger/telemetry` is now the place where we centralize OTel setup across the Dagger CLI, engine, and SDKs. Aligns incentives with the Go SDK (minimize dependency sprawl).
* ~~Fix integration test suite OTel integration - now we'll have one span per `connect(t)`, and one span will show up beneath them~~
  * ~~Unfortunately `testutil.Query` call sites won't. Maybe we could refactor those to use `connect(t)` internally, actually...~~
  * Cutting scope from this PR, will address later, involves widespread changes across the test suite to switch to an instrumentable test runner
* Standardize "OTel" abbreviation styling
* Get rid of OTel proxy socket - have our TCP listener send straight to PubSub
* Drain telemetry based on client ID instead of by trace
  * Previously if you had multiple connections in the same trace, you'd get a 5 minute hang on `.Close()` because they'd wait on each other
  * I noticed this in this PR because `TestContainerLabel` was hanging, but for some reason it doesn't hang on `main`. Probably because they all actually ended up being different traces.
  * Engine installs a SpanProcessor that grabs `engine.ClientMetadataFromContext` and adds the client ID as a span attribute
  * PubSub now tracks active spans by client ID attribute, rather than by trace
  * Client shutdown now passes its ID to draining
* Get rid of the `Span` and `TracerProvider` wrapping
  * Instead, you just write a `SpanProcessor` wrapper whose `OnStart` simply calls `OnEnd`.
  * No more `OnUpdate` - not worth the wrapping complexity. For realtime updates, use SDK logging instead. (maybe :hot_pepper: take)
  * [x] Move span heartbeating to generic `Exporter` interface for Cloud.
* [x] Test coverage for draining with multiple clients
* [x] Test coverage for multiple clients not having cross-talk
* [x] Implement draining for logs, too
  * Logging code helpers have been refactored to support closing the streams when you're done, which sends EOF.
  * User code doesn't have to care about this, but core code does. All code paths that send logs should `defer Close()` at some point.
* [x] Fix #7510 by directing logs + telemetry to the function call's SpanContext
  * This also removes the need for `dagger.io/ui.mask` - now we can just hide the `exec /runtime` vertex. Simplified the SDKs accordingly, but could use a look from @helderco @TomChv - https://github.com/dagger/dagger/pull/7452/commits/8e0e9469a833be635857ea3fd1409fe68e64dbd5
* [x] Consider a span to be drained if its parent span has completed, to avoid indefinite hangs in non-clean-shutdown scenarios


cmd/dagger/engine.go
```diff
@@ -3,18 +3,26 @@ package main
...
-	"github.com/dagger/dagger/telemetry"
+	"github.com/moby/buildkit/identity"
 )
 
 type runClientCallback func(context.Context, *client.Client) error
 
+// ClientID is the client ID for the CLI's connection to the engine.
+//
+// It must be established as soon as possible for telemetry purposes.
+var ClientID = identity.NewID()
```
vito, line 17: The new draining logic is based on client IDs, so now we determine the client ID even earlier than connecting to the engine, so we can immediately annotate telemetry coming out of the CLI

cmd/dagger/licenses.go
```diff
@@ -62,19 +61,19 @@ var licenseFiles = []string{
 }
 
 func findOrCreateLicense(ctx context.Context, dir string) error {
-	log := telemetry.ContextLogger(ctx, slog.LevelWarn)
```
vito, line 65: `ContextLogger` and `ioctx` are gone now

cmd/dagger/main.go
```diff
@@ -31,8 +32,8 @@ import (
 	"github.com/dagger/dagger/analytics"
 	"github.com/dagger/dagger/dagql/idtui"
 	"github.com/dagger/dagger/engine"
-	"github.com/dagger/dagger/telemetry"
-	"github.com/dagger/dagger/telemetry/sdklog"
+	enginetel "github.com/dagger/dagger/engine/telemetry"
```
vito, line 35: Could maybe rename this package, since it's used by both the CLI and the engine, but don't really know a better name. :thinking: This code is split out to minimize the `dagger.io/dagger/telemetry` dependency graph.

cmd/dagger/session.go
```diff
@@ -39,6 +40,10 @@ type connectParams struct {
 }
 
 func EngineSession(cmd *cobra.Command, args []string) error {
+	// discard SIGPIPE, which can happen when stdout or stderr are closed
+	// (possibly from the spawning process going away)
+	signal.Notify(make(chan os.Signal, 1), syscall.SIGPIPE)
```
vito, line 45: cc @jedevc for continuity - ended up having to add an even harder fix for this, since it was still happening, only this time for stderr. Seems fine to just drop all SIGPIPEs.

cmd/engine/main.go
```diff
@@ -888,25 +876,6 @@ func parseBoolOrAuto(s string) (*bool, error) {
 	return &b, err
 }
 
-// Run a separate gRPC serving _only_ the trace/log exporter services.
-func runOtelController(p string, pubsub *telemetry.PubSub) error {
```
vito, line 892: This isn't needed anymore :tada: - now we just send OTLP over HTTP and the requests get sent straight to the PubSub, instead of mounting in a unix socket and setting up a TCP proxy.

core/container_exec.go
```diff
@@ -101,7 +101,6 @@ func (container *Container) WithExec(ctx context.Context, opts ContainerExecOpts
 		if err != nil {
 			return nil, fmt.Errorf("register caller: %w", err)
 		}
-		execMD.OTELEnvs = callerOpts.OTELEnvs
```
vito, line 104: Nothing was actually setting/using `callerOpts.OTELEnvs` anymore so I removed it - but we should bring back its original purpose - see #7510. The regression is already live so this can be fixed in a followup.

core/integration/client_test.go
```diff
@@ -13,3 +20,87 @@ func TestClientClose(t *testing.T) {
 	err := c.Close()
 	require.NoError(t, err)
 }
+
+func TestClientMultiSameTrace(t *testing.T) {
```
vito, line 24: This test covers quite a bit, and I confirmed that it fails before this PR's changes, but it's pretty sensitive. I've had it pass 100 times in a row locally but theoretically there's still a timing issue lurking with logs (see FIXME below). The timing issue is real, but I suspect it mostly affects nesting situations like our test suite.

core/modfunc.go
```diff
@@ -197,18 +195,6 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Typ
...
-	tc := propagation.TraceContext{}
-	carrier := propagation.MapCarrier{}
-	tc.Inject(ctx, carrier)
-	for _, f := range tc.Fields() {
-		name := "DAGGER_FUNCTION_" + strcase.ToScreamingSnake(f)
-		if val, ok := carrier[f]; ok {
-			callMeta.OTELEnvs = append(callMeta.OTELEnvs, name+"="+val)
-		}
-	}
-
```
vito, line 211: see https://github.com/dagger/dagger/pull/7452/files#r1628294096 - correction: we were setting it, but nothing reads this env var anymore

dagql/idtui/frontend_plain.go
```diff
@@ -259,17 +278,15 @@ func (fe *frontendPlain) render() {
 }
 
 func (fe *frontendPlain) finalRender() {
-	if fe.Debug || fe.Verbosity > 0 {
-		fe.render()
-	}
```
vito, line 264: @jedevc - pretty sure this conditional didn't quite make sense for pretty output. I'm assuming it took this logic from the TUI, but the TUI only had it because in that case it was printing the _full_ progress output. In this case the final `render()` call is needed in order to not miss the last bit of progress. Without this fix, the `too bad` test was flaky/failing.

engine/buildkit/executor_spec.go
```diff
@@ -346,7 +346,7 @@ func (w *Worker) generateBaseSpec(ctx context.Context, state *execState) error {
 		w.idmap,
 		w.apparmorProfile,
 		w.selinux,
-		w.tracingSocket,
+		"",
```
vito, line 349: By passing in `""` we don't need to deal with Buildkit's configured `OTEL_` env vars, and can just set our own.

engine/buildkit/executor_spec.go
```diff
@@ -626,114 +616,136 @@ func (w *Worker) setupStdio(_ context.Context, state *execState) error {
...
-		OTELExporterProtocolEnv+"="+state.otelProto,
-		OTELExporterEndpointEnv+"="+state.otelEndpoint,
-		OTELTracesProtocolEnv+"="+state.otelProto,
-		OTELTracesEndpointEnv+"="+state.otelEndpoint,
+		OTelExporterProtocolEnv+"="+otelProto,
+		OTelExporterEndpointEnv+"="+otelEndpoint,
+		OTelTracesProtocolEnv+"="+otelProto,
+		OTelTracesEndpointEnv+"="+otelEndpoint+"/v1/traces",
+		// Indicate that the /v1/trace endpoint accepts live telemetry.
+		OTelTracesLiveEnv+"=1",
```
vito, line 700: Previously nested `dagger` CLIs couldn't tell that they should emit live telemetry because it's going to the PubSub, so things would look a bit wonky with child spans appearing before their parents. Now we just set a separate env var.

engine/buildkit/executor_spec.go
```diff
@@ -626,114 +616,136 @@ func (w *Worker) setupStdio(_ context.Context, state *execState) error {
...
+	// Configure our OpenTelemetry proxy. A lot.
 	state.spec.Process.Env = append(state.spec.Process.Env,
-		OTELExporterProtocolEnv+"="+state.otelProto,
-		OTELExporterEndpointEnv+"="+state.otelEndpoint,
-		OTELTracesProtocolEnv+"="+state.otelProto,
-		OTELTracesEndpointEnv+"="+state.otelEndpoint,
+		OTelExporterProtocolEnv+"="+otelProto,
+		OTelExporterEndpointEnv+"="+otelEndpoint,
+		OTelTracesProtocolEnv+"="+otelProto,
+		OTelTracesEndpointEnv+"="+otelEndpoint+"/v1/traces",
```
vito, line 698: Pretty sure this var along with the log + metrics equivalent were just wrong before - I was getting `404` errors because telemetry was being sent to the base URL, without `/v1/(traces|logs|metrics)`.

engine/client/buildkit.go
```diff
@@ -21,12 +21,11 @@ const (
 
 func newBuildkitClient(ctx context.Context, remote *url.URL, connector drivers.Connector) (_ *bkclient.Client, _ *bkclient.Info, rerr error) {
 	opts := []bkclient.ClientOpt{
-		// TODO verify?
-		bkclient.WithTracerProvider(otel.GetTracerProvider()),
+		bkclient.WithTracerProvider(otel.GetTracerProvider()), // TODO verify?
+		bkclient.WithContextDialer(func(ctx context.Context, _ string) (net.Conn, error) {
+			return connector.Connect(ctx)
+		}),
```
vito, line 27: Sneaky local var capture was happening here (note the now-named `ctx context.Context` arg). Detected with `-race`.

engine/telemetry/heartbeat.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/dagger/dagger/engine/slog"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+const HeartbeatInterval = 30 * time.Second
+
+// SpanHeartbeater is a SpanExporter that keeps track of live spans and
+// re-exports them periodically to the underlying SpanExporter to indicate that
+// they are indeed still live.
+type SpanHeartbeater struct {
```
vito, line 17: Previously heartbeating was a responsibility of each SDK's "live" OTel integration. Now it's just a layer we wrap around our Cloud exporter, and each SDK's minimal "live" implementation is just a SpanProcessor that calls OnEnd for every OnStart.

sdk/go/telemetry/init.go
```diff
@@ -103,21 +104,104 @@ func ConfiguredSpanExporter(ctx context.Context) (sdktrace.SpanExporter, bool) {
...
+					name, value, _ := strings.Cut(header, "=")
+					headers[name] = value
+				}
+			}
+			configuredLogExporter, err = otlploghttp.New(ctx,
+				otlploghttp.WithEndpointURL(endpoint),
+				otlploghttp.WithHeaders(headers))
+
+		case "grpc":
+			// FIXME: bring back when it's actually implemented
```
vito, line 152: NOTE: currently-shipped Dagger uses gRPC for nested OTel logging, so this means a dev Dagger in shipped Dagger won't be able to emit logs.

cmd/dagger/engine.go
sipsma: In the case of a nested dagger CLI the client ID ends up being the one set by the executor, so this would not be used. Guessing that's fine cause we [short-circuit that case](https://github.com/sipsma/dagger/blob/0b36cdd745611b75b43f13ef6158de5a45d4bd00/engine/client/client.go#L145-L145) in the client anyways, just double checking my overall understanding

sdk/go/telemetry/proxy.go
```diff
@@ -1,97 +1,6 @@
...
-func (s proxySpan) SetName(name string) {
-	s.sp.SetName(name)
-	s.onUpdate(s.sp)
-}
-
-func (s proxySpan) TracerProvider() trace.TracerProvider { return s.sp.TracerProvider() }
+// FIXME: this file exists to plant a "tombstone" over the previously generated
+// proxy.go file.
+//
+// We should maybe just withoutDirectory('./internal') or something instead.
```
vito, line 6: This is an awkward little scenario I ran into when dealing with codegen. :confused:

engine/buildkit/executor_spec.go
```diff
@@ -626,114 +616,136 @@ func (w *Worker) setupStdio(_ context.Context, state *execState) error {
...
-			return net.Listen("tcp", "127.0.0.1:0")
-		})
-		if err != nil {
-			return fmt.Errorf("otel tcp proxy listen: %w", err)
-		}
-		state.otelEndpoint = "http://" + listener.Addr().String()
+	otelProto := "http/protobuf"
+	otelEndpoint := "http://" + listener.Addr().String()
+	otelSrv := &http.Server{
+		Handler: w.telemetryPubSub,
```
sipsma, line 652: Neat, https://github.com/dagger/dagger/pull/7315 among other things makes the whole engine API just a single HTTP server, but needed to do some very light hijinks to still serve the gRPC otel endpoints: https://github.com/sipsma/dagger/blob/55a8e049ca22e7b8d76e86079e9b9e152d8b4465/cmd/engine/main.go#L424-L424

Would it make sense to now just install this pubsub http handler instead?

It's really not that big of a deal either way, checking the content type for grpc is ultimately fine, but would take that simplification if it makes sense.

This may require way too much context about the other PR to answer now in which case plz ignore; can figure it out later.

engine/buildkit/executor_spec.go
```diff
@@ -626,114 +616,136 @@ func (w *Worker) setupStdio(_ context.Context, state *execState) error {
...
-				proxyConnPool.Go(func() {
-					defer remote.Close()
-					io.Copy(remote, conn)
-				})
-
-				proxyConnPool.Go(func() {
-					defer conn.Close()
-					io.Copy(conn, remote)
-				})
+			l := &singleConnListener{
```
sipsma, line 675: Do we still need to do this or can we just simplify to do `otelSrv.Serve(listener)`? I think manually calling `Accept` was needed for the proxying but since that's gone it seems like we only need the one "real" listener? May be missing something though

cmd/dagger/session.go
sipsma: This makes sense and sgtm; I was reading the linux docs about `SIGPIPE` and got kind of confused about what the end effect would be here with the go runtime, but the go docs answered it clearly: https://pkg.go.dev/os/signal#hdr-SIGPIPE

Maybe worth including that link in the comment for any future confused individuals

cmd/engine/main_oci_worker.go
```diff
@@ -377,7 +377,7 @@ func snapshotterFactory(commonRoot string, cfg config.OCIConfig, sm *session.Man
 				grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(defaults.DefaultMaxRecvMsgSize)),
 				grpc.WithDefaultCallOptions(grpc.MaxCallSendMsgSize(defaults.DefaultMaxSendMsgSize)),
 			}
-			conn, err := grpc.Dial(dialer.DialAddress(address), gopts...)
+			conn, err := grpc.NewClient(dialer.DialAddress(address), gopts...)
```
sipsma, line 380: no problem but curious what the reason for this change is

cmd/dagger/session.go
```diff
@@ -92,7 +97,7 @@ func EngineSession(cmd *cobra.Command, args []string) error {
 		paramBytes = append(paramBytes, '\n')
 		go func() {
 			if _, err := os.Stdout.Write(paramBytes); err != nil {
-				panic(err)
+				fmt.Fprintln(cmd.ErrOrStderr(), err)
```
sipsma, line 100: We don't need to panic here, but I think we should exit the process if this fails; the client will never be able to find the port so may as well die 💀

engine/client/client.go
```diff
@@ -126,6 +131,7 @@ func Connect(ctx context.Context, params Params) (_ *Client, _ context.Context,
 	c.internalCtx, c.internalCancel = context.WithCancel(context.WithoutCancel(ctx))
 	c.closeCtx, c.closeRequests = context.WithCancel(context.WithoutCancel(ctx))
 
+	// FIXME: is this still needed? (not a problem, just want to be sure)
 	c.eg, c.internalCtx = errgroup.WithContext(c.internalCtx)
```
sipsma, line 135: The errgroup is still useful for running the session attachables server async, so it's fine for now I think. The ctxs in this file are still confusing though, but would leave as is for now.

engine/telemetry/heartbeat.go
```diff
@@ -0,0 +1,87 @@
...
+		case <-p.heartbeatCtx.Done():
+			return
+		case <-ticker.C:
+			p.activeSpansL.Lock()
+			var stayinAlive []sdktrace.ReadOnlySpan
+			for _, span := range p.activeSpans {
+				stayinAlive = append(stayinAlive, span)
+			}
+			if len(stayinAlive) > 0 {
+				if err := p.ExportSpans(p.heartbeatCtx, stayinAlive); err != nil {
```
sipsma, line 80: We hold `p.activeSpansL.Lock()` here and then call `p.ExportSpans` which also starts with a call to `p.activeSpansL.Lock()`. Is that deadlock or am I misreading?

engine/telemetry/pubsub.go
```diff
@@ -0,0 +1,826 @@
...
+func (ps *PubSub) clientsFor(traceID trace.TraceID, spanID trace.SpanID) []string {
+	ps.spansL.Lock()
+	defer ps.spansL.Unlock()
+	key := spanKey{
+		TraceID: traceID,
+		SpanID:  spanID,
+	}
+	// slog := slog.With("traceID", traceID, "spanID", spanID)
+	clients := map[string]struct{}{}
+	for {
```
sipsma, line 160: nit: in the interest of paranoia, is it worth some sanity check for the conditions that would result in this loop accidentally spinning forever? e.g. checking `parent` is not the same value as the existing `SpanID`. That would require a bug elsewhere but would rather error if that happens than get stuck in a loop.

Haven't fully loaded all context yet so may be a dumb suggestion, not a blocker.

engine/telemetry/pubsub.go
```diff
@@ -0,0 +1,826 @@
...
+type ClientMetadata struct {
+	// ClientID is unique to each client. The main client's ID is the empty string,
+	// any module and/or nested exec client's ID is a unique digest.
+	ClientID string `json:"client_id"`
+}
+
+const (
+	// NB: a bit odd to move this here, but feels less risky than duplicating
+	ClientMetadataMetaKey = "x-dagger-client-metadata"
+)
```
sipsma, line 251: nit: if we changed the field `ClientMetadata.Labels` in `engine/opts.go` to just be `map[string]string` could we just import that `ClientMetadata` type into this package and use it? Or do we not want to use that type here for other reasons?

I don't care about where the const is defined or anything like that really, just seeing if we can de-dupe some of the code like `decodeMeta` and such.

engine/telemetry/pubsub.go
```diff
@@ -0,0 +1,826 @@
...
+	var req coltracepb.ExportTraceServiceRequest
+	if err := proto.Unmarshal(body, &req); err != nil {
+		slog.Error("error unmarshalling request", "err", err)
+		rw.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	if err := ps.Spans().ExportSpans(r.Context(), telemetry.SpansFromPB(req.ResourceSpans)); err != nil {
+		slog.Error("error exporting spans", "err", err)
+		rw.WriteHeader(http.StatusInternalServerError)
```
sipsma, line 304: nit: not sure if useful but if you do `http.Error(rw, err, http.StatusInternalServerError)` the error message gets sent back to the client to as the response body, which can make debugging easier provided it's printed. May be irrelevant here though, not sure.

engine/telemetry/heartbeat.go
vito: Whoops, totally, thanks! This is supposed to call through to the underlying exporter, not to its sibling method.

engine/telemetry/pubsub.go
vito: Yeah that unties the knot - done, thanks!

cmd/engine/main_oci_worker.go
vito: nothing big - `grpc.Dial` is just deprecated, my linter/editor got mad about it in one file so I changed them all

engine/buildkit/executor_spec.go
vito: Haha yes, this can be way simplified - just pushed a commit which reduces it down to basically just `http.Serve` and `srv.Shutdown`. No idea how it ended up so complicated before, was probably flailing late at night. :joy:

cmd/dagger/session.go
jedevc: SGTM - writing to stdout failing with an error seems like the expected behavior for this.

@vito if you were seeing this, then that might have been a symptom of something else - why was stderr being closed in the first place? It's likely fine, but the only reason I noticed this in the first place was because the `SIGPIPE` behavior was the only thing preventing an infinite telemetry drain.

dagql/idtui/frontend_plain.go
jedevc: Hm, you're right, not quite sure what I was thinking.

However, (just noticed this) we should ideally track the equivalent of `fe.err` - otherwise, we don't show any progress output on an error with `--progress=plain --silent`. Hopefully that makes sense.

We'd also want to set `contextHold = 0` here, if that's the case :thinking:

cmd/dagger/licenses.go
jedevc: :pray: thank goodness :smile:

Genuinely happy to see these gone, they were quite confusing :)

core/container_exec.go
jedevc: Should we also remove the `execMD.OTELEnvs` field now?

dagql/idtui/frontend.go
```diff
@@ -101,10 +103,13 @@ func (r renderer) renderIDBase(out *termenv.Output, call *callpbv1.Call) error {
 		parent = parent.Foreground(moduleColor)
 	}
 	fmt.Fprint(out, parent.String())
+	if r.Verbosity > 0 && call.ReceiverDigest != "" {
+		fmt.Fprint(out, out.String(fmt.Sprintf("@%s", call.ReceiverDigest)).Foreground(faintColor))
+	}
```
jedevc, line 108: Hm this adds a bit of noise for me.

I use `-v` by itself just to get the pretty logs to stick around (I use terminal scrollback a lot if you can't tell :laughing:) - I'm actually starting to wonder if we should split persistence out from `-v`.

engine/client/client.go
```diff
@@ -237,21 +243,33 @@ func (c *Client) startEngine(ctx context.Context) (rerr error) {
...
+	slog.Info("connecting", "runner", c.RunnerHost, "client", c.ID)
+
 	bkClient, bkInfo, err := newBuildkitClient(ctx, remote, connector)
 	if err != nil {
 		return fmt.Errorf("new client: %w", err)
 	}
 	c.bkClient = bkClient
 
 	if err := retry(ctx, 10*time.Millisecond, func(elapsed time.Duration, ctx context.Context) error {
+		slog.Info("dialing telemetry", "remote", c.RunnerHost)
```
jedevc, line 263: Oooh I *really* like this - using the `SpanLogger` gets *significantly* better results IMO, getting all the log messages to the right place :pray: 

Some of these are a bit noisy though (this is with just `-v`):

```
✔ connect 0.6s
  ✔ connecting to engine 0.1s
  ┃ 15:18:26 INF connecting runner=docker-container://dagger-engine.dev client=p                                                                  
  ┃ 166djtditnzvem0omeqm                                                                                                                          
  ┃ 15:18:26 INF dialing telemetry remote=docker-container://dagger-engine.dev                                                                    
  ┃ 15:18:26 INF dialed telemetry elapsed=8.44µs                                                                                                  
```

engine/client/client.go
```diff
@@ -237,21 +243,33 @@ func (c *Client) startEngine(ctx context.Context) (rerr error) {
 		return err
 	}
 
+	c.connector = connector
+
 	ctx, span := Tracer().Start(ctx, "connecting to engine")
 	defer telemetry.End(span, func() error { return rerr })
 
+	logs, slog := slog.SpanLogger(ctx, InstrumentationLibrary, slog.LevelDebug)
```
jedevc, line 251: I like this, but - one issue with this is the user (through the frontend) can't configure the logging level.

I hit this before (it's not new) - but ideally it would be really nice if we could somehow defer this decision into the frontend?

dagql/idtui/frontend.go
vito: Ah this was mostly me adding it because I needed it in the moment, I can bump the verbosity level to `-vv`. I think `-v` being mostly just persistence is ideal.

There are a bunch of individual flags we could probably add but I'd rather try to just stick with the verbosity level mindset as long as possible and try to curate it, since any other flag I can think of tends to be long, and I'm not gonna want to type it, and if we give these flags short names that's potentially claiming a lot of short-flag-name real estate.

engine/client/client.go
jedevc: Also something seems wrong in the plain progress now:

```
1   : connect
2   :   connecting to engine
2   :   [0.0s] | 15:26:13 INF connecting runner=docker-container://dagger-engine.dev client=trurchzatdscg1loohyjplb48
2   :   [0.1s] | 15:26:13 INF dialing telemetry remote=docker-container://dagger-engine.dev
2   :   [0.1s] | 15:26:13 INF dialed telemetry elapsed=6.18µs
2   :   [0.1s] | 
2   :   [0.1s] | 
2   :   connecting to engine DONE [0.1s]
3   :   starting session
3   :   [0.2s] | 
3   :   [0.2s] | 
3   :   starting session DONE [0.2s]
1   : connect DONE [0.3s]
```

Not sure where the excess newlines are coming from. Also, the double timestamps are a *little* annoying (but I can suffer).

jedevc:
Hm, just a note - it appears that accessing this engine with an old client causes some hangs. I guess with such major changes to telemetry that's to be expected.

If we *don't* fix that though, IMO we should make sure to get https://github.com/dagger/dagger/pull/7031 in the same release (to avoid the same user complaints when users hit issues with this before - sadly, not everyone seems to read the release notes).

core/container_exec.go
vito: #7510 is fixed in https://github.com/dagger/dagger/pull/7452/commits/8e0e9469a833be635857ea3fd1409fe68e64dbd5 now

dagql/idtui/types.go
```diff
@@ -49,19 +48,6 @@ func CollectSpans(db *DB, traceID trace.TraceID) []*Span {
 		if traceID.IsValid() && span.SpanContext().TraceID() != traceID {
 			continue
 		}
-		if span.Mask && span.Parent().IsValid() {
-			masked := db.Spans[span.Parent().SpanID()]
-			if masked != nil {
-				masked.Passthrough = true
-			} else {
-				// FIXME(vito): still investigating why this happens, but in
-				// the mean time, better not to panic
```
jedevc, line 58: Oh! Does this get fixed now?

How are masked spans handled now without this block?

vito: tl;dr: maybe?

https://github.com/dagger/dagger/pull/7452/commits/8e0e9469a833be635857ea3fd1409fe68e64dbd5 refactors it away, so this error is gone, but there were many possible ways to trigger it.

Previously we would run the function entrypoint and require the SDK to create a span that combines `dagger.io/ui.mask` and `dagger.io/ui.passthrough` to hide the `exec /runtime` span and the "dummy" span that carries the attributes. Now we just redirect logs and traces to the function span, and mark the `exec /runtime` span internal in the core implementation (I actually still need TODO this last bit, just by adding the `[internal] ` prefix). cc @helderco @TomChv (sorry for all the pings, just connecting dots :P)

There was one occurrence of this error ages ago just after shipping OTel (back when it was still a nil panic) where it turned out a ton of data was completely missing as if it was never sent out from the engine. But I've never been able to repro (after many many attempts), and users don't seem to be reporting that, so it's possible it was just a client/engine mismatch or something weird with our infra.

Another possible cause for this is receiving child spans before parent spans, which could easily happen because that's _normally_ just how OTel works. In practice I saw this happen in our test suite because our nested `dagger` calls weren't instructed to send live telemetry. That would also manifest itself by having spans show up awkwardly at the bottom of the TUI and then "pop" in to place when their parent span arrived. That's fixed now too; we set an unambiguous env var that the `dagger` CLI detects and configures live telemetry.

So, this error log indeed doesn't exist anymore, and now I can only hope I've addressed all the issues that led to it. At least by that point we won't be possibly conflating 10 different root causes, and maybe as a result we'll get deeper analysis of what went wrong by looking at a more important symptom (i.e. missing data) rather than a cryptic log message.

ci/mage/engine.go
```diff
@@ -163,7 +167,7 @@ func (t Engine) Dev(ctx context.Context) error {
 		"--privileged",
 	}...)
 
-	runArgs = append(runArgs, imageName, "--debug", "--debugaddr=0.0.0.0:6060")
+	runArgs = append(runArgs, imageName, "--extra-debug", "--debugaddr=0.0.0.0:6060")
```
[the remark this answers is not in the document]
vito: Yeah, I think there's a lot of upside to enabling this for `./hack/dev` and basically no downside, since these logs don't end up in CI. Without enabling this I'm basically helpless when debugging OTel issues.

Not sure what'll happen to this when we retire Mage, but I can burn that bridge when I get to it.

engine/client/client.go
vito: Totally - just added a `LogLevel` param which gets set by `withEngine`, so now it'll respect `--debug`

engine/client/client.go
vito: Moved these down to debug level, now that it respects `--debug` (https://github.com/dagger/dagger/pull/7452#discussion_r1629658902)

core/integration/module_test.go
```diff
@@ -4277,7 +4277,7 @@ class Test {
 				require.NoError(t, c.Close())
 
 				t.Log(logs.String())
-				require.Contains(t, logs.String(), "too bad")
+				require.Regexp(t, "too bad: so sad", logs.String())
```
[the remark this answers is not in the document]
sipsma: Yikes at the false positive, that's extremely subtle but makes sense...

Were the times this test failed due to one of the OTEL flushing issues and fixed now? Just confirming because my other refactoring PR seems to hit failures on it even more often than before, but not sure if it's likely to be just a timing difference and fixed once I rebase on this PR.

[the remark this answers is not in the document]
vito: I think it's actually the switch to the new plain output that introduced the flakiness - see https://github.com/dagger/dagger/pull/7452#discussion_r1628304802

Definitely timing related, should be fixed after rebasing on this since it applies the aforementioned fix to the plain frontend.

engine/client/client.go
vito: @jedevc Ah just saw the second comment - those blank lines are probably the same bug that happens in Cloud now. The new logging infra send a log record on `Close()` whose body is an empty string, with `stdio.eof=true`, to indicate that you're received the last log record. The plain format must be splitting on `\n` and not handling the case where there's no linebreak.

core/integration/module_test.go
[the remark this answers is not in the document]
sipsma: Thanks, just did a rebase on this PR and can confirm it passes consistently in the other PR now 🎉

engine/telemetry/pubsub.go
vito: Yeah sure, might as well. https://github.com/dagger/dagger/pull/7452/commits/1515f71e2b695264d27407cf28732d633a709473 makes it keep track of which keys it's seen as it goes and bail if it sees one again. That should also detect cycles.

cmd/dagger/session.go
vito: > @vito if you were seeing this, then that might have been a symptom of something else - why was stderr being closed in the first place? It's likely fine, but the only reason I noticed this in the first place was because the `SIGPIPE` behavior was the only thing preventing an infinite telemetry drain.

The OS closed the pipe because the spawning process does `cmd.StdoutPipe` and then exits. So nothing actually called `.Close()`, the pipe just broke on its own, and then the `dagger session` process got `SIGPIPE` when it tried to write to it.

@sipsma Yeah good point, I ended up on that same doc, and it relates to the above question, so worth a link indeed. Added!

cmd/dagger/engine.go
vito: Good call, ended up removing this, it wasn't load-bearing and I couldn't really come up with a good reason to keep it besides "having a client set everywhere felt good." It doesn't end up being set everywhere anyway, and it doesn't need to, since the full span ancestry chain is walked to collect client IDs.

Would rather just not have it than have misleading values in telemetry for the nesting case. :+1:

dagql/idtui/frontend_plain.go
vito: I just pushed a fix for `--silent` - removing these conditionals actually made it print progress on exit with `--silent` which seems like very much not what you want. :smile: 

Not sure about `contextHold = 0` and `fe.err`, but I tried running things that errored with `--silent` and it seems OK. :thinking: Lemme know if this is blocking, or feel free to push changes to this PR :pray: I think you have access

engine/buildkit/executor_spec.go
vito: Ha sweet, yeah that sounds plausible! I guess the only concern would be whether we want to also expose the full engine API, or just these OTLP endpoints, since this will be passed to everything regardless of whether they enabled nesting.

cmd/dagger/session.go
vito: Good point - added a `l.Close()` like the others, which should unblock the `http.Serve` below - seems to be what we do elsewhere, too.

vito:
> Hm, just a note - it appears that accessing this engine with an old client causes some hangs. I guess with such major changes to telemetry that's to be expected.
> 
> If we _don't_ fix that though, IMO we should make sure to get #7031 in the same release (to avoid the same user complaints when users hit issues with this before - sadly, not everyone seems to read the release notes).

I just went to see how difficult backwards-compatibility will be, and it is pretty annoying to deal with, so yeah, I'd be good with prioritizing #7031 - will review it now!

sipsma (approved):
🎉

dagql/idtui/frontend_plain.go
jedevc: Aha, we don't need `fe.err`, but I pushed a change to `contextHold` - that's just to disable the default delay in how we do "switching" between different contexts.

jedevc:
Yeahhh let's go for it :tada: the sooner we can get this in and start testing it, the better IMO :tada:

FYI, if we can get #7544 in, then we can provide a command for users to install a version of dagger off of `main`, which provides us a way to try and get some more testing of this *before* we release this next week.

PR #7459, opened 2024-05-24
docs snippets: don't commit generated code, fix test build

vito:
* PR tests fail after #7454 because the new matrix-build snippet doesn't bring its own `go.mod`, so the first commit fixes that.
  * [Example failure](https://dagger.cloud/dagger/traces/8d01f0fd5ec61ec161174c33424452fb?span=5f8f35bae4045a16&logs)
* The second commit removes _all_ generated code, which I'm pretty sure we want, but I'm not sure if anything breaks from it. Putting up a quick PR to see if it's OK.


vikram-dagger:
I see the `dagger.json` is present in some cases and not in others. For consistency should we remove that everywhere as well?

vikram-dagger:
Also as a follow up: what is the minimal set of files needed for a Go snippet?

vito:
@vikram-dagger I think `dagger.json` should be included, otherwise you can't run the module locally with the `dagger` CLI, which seems useful while working on the snippets. I can add it anywhere that's missing if you've got a list handy - kind of hard for me to tell. :pray: 

The minimal set of files are:

* `dagger.json` - so we can invoke it as a module
* `main.go` - the snippet itself
* `go.mod` and `go.sum` - so it doesn't get included with `go test ./...` from the root
* `.gitignore` - so we don't commit all the generated code
* `.gitattributes` - not super necessary, but it comes with `dagger develop`, so I think we should just include it as we would with any other module

vito:
FWIW it's understandable that we missed `go.mod` with #7454 because the CLI will by default look upwards until it finds the `go.mod` in the root of the repo.

When we're initializing a new snippet we should probably `go mod init dagger/my-module` to explicitly make a new one first.

vikram-dagger (approved):

vikram-dagger:
> @vikram-dagger I think `dagger.json` should be included, otherwise you can't run the module locally with the `dagger` CLI, which seems useful while working on the snippets. I can add it anywhere that's missing if you've got a list handy - kind of hard for me to tell. 🙏

I think `dagger.json` and `.gitignore` are missing in almost all the snippets. The ones which do have them will be the exception rather than the rule. These two files were not included to date because they were not needed to pass CI tests.

vito:
> I think `dagger.json` and `.gitignore` are missing in almost all the snippets. The ones which do have them will be the exception rather than the rule. These two files were not included to date because they were not needed to pass CI tests.

OK to fix these gradually then? I see you approved already, would like to merge just to unbreak PRs.

For the record, here are the commands I ran to quickly fix up anything that had a `dagger.json`:

```fish
# copy .gitignore to all modules
for js in (find docs/ -name dagger.json); cp ci/.gitignore (dirname $js); end

# remove .gitignored files (by removing in git only and re-adding)
for js in (find docs/ -name dagger.json); pushd (dirname $js); git rm -r --cached .; git add .; popd; end
```

vito:
@vikram-dagger Did another bulk cleanup:

```fish
# put a dagger.json and .gitignore anywhere that we commit dagger.gen.go
for js in (find docs/ -name dagger.gen.go); cp ./docs/current_docs/manuals/developer/cookbook/snippets/builds/matrix-build/go/dagger.json (dirname $js); cp ci/.gitignore (dirname $js); end

# same as before, to remove the dagger.gen.go files from above
for js in (find docs/ -name dagger.json); pushd (dirname $js); git rm -r --cached .; git add .; popd; end
```

vikram-dagger:
> OK to fix these gradually then? I see you approved already, would like to merge just to unbreak PRs.

Yes, of course!

vikram-dagger:
@vito I came across a reason why we probably also need the generated code, but lmk what you think: if the `dagger.gen.go` file is missing and you then go in to edit an existing snippet, VSCode automatically adds `"dagger.io/dagger/dag"` as an import. This then becomes part of the code snippet and visible in the docs, although it isn't strictly necessary.

vito:
@vikram-dagger That's true, but it's also true for `ci/` or any other module. It's pretty annoying, not sure if we have an issue tracking that particular papercut. I would recommend running `dagger develop` before editing in VSCode, since most modules still won't compile properly even after that auto-import is added.

vikram-dagger:
> @vikram-dagger That's true, but it's also true for `ci/` or any other module. It's pretty annoying, not sure if we have an issue tracking that particular papercut. I would recommend running `dagger develop` before editing in VSCode, since most modules still won't compile properly even after that auto-import is added (because they'd still be missing `Container` and all the other core types).

This can be done, but if I understood correctly, it adds significant friction to the docs editing process. Instead of a simple edit to a file, the process changes to `dagger develop` (which very often takes time), edit file, delete `dagger.gen.go`, commit and push. If the CI fails and the file needs a further edit, repeat the above process (which will usually take longer than the time needed for the edit because of the `dagger develop` generation). IMHO it's actually better to retain/commit the `dagger.gen.go` file as including it in the minimal list eliminates this friction.

docs/current_docs/quickstart/snippets/publish/go/dagger.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "my-module",
```
vikram-dagger, line 2: This and other quickstart `dagger.json` module names should be `hello-dagger`

vito:
> This can be done, but if I understood correctly, it adds significant friction to the docs editing process. Instead of a simple edit to a file, the process changes to `dagger develop` (which very often takes time), edit file, delete `dagger.gen.go`, commit and push. If the CI fails and the file needs a further edit, repeat the above process (which will usually take longer than the time needed for the edit because of the `dagger develop` generation). IMHO it's actually better to retain/commit the `dagger.gen.go` file as including it in the minimal list eliminates this friction.

You don't need to delete `dagger.gen.go` since it's `.gitignore`d. You should only need to `dagger develop` "every now and then" to make your editor happy, so even if it's a bit slow, it shouldn't be in the critical path of that feedback loop.

docs/current_docs/quickstart/snippets/publish/go/dagger.json
vito: Nice catch - putting up a PR soon! :pray:

PR #7461, opened 2024-05-24
CI: start spinning out our Go and GraphQL toolchain

shykes:
This is a first incremental change in a larger set of planned changes. I am simply picking low-hanging fruits in the existing CI codebase, and transposing them into a standalone module, with as few changes as possible. When we have done a first pass at this (which will require more follow-up PRs), we should assess the overall codebase, and decide what to refactor, if anything.

- `std/go`: Go toolchain in a Dagger module
- `std/graphql`: GraphQL toolchain in a Dagger module
- `dirdiff`: a small utility to compare two directories (extracted from an internal utility)


shykes:
I can't find the root cause of the CI failures... 😭 

To reproduce in the CLI:

```console
$ dagger call \
 -m github.com/dagger/dagger@pull/7461/head --source=https://github.com/dagger/dagger#pull/7461/head \
  engine \
  container \
  terminal
```

Output:

```console
go: finding module for package github.com/dagger/dagger/ci/internal/dagger
go: github.com/dagger/dagger/ci imports
        github.com/dagger/dagger/ci/internal/dagger: module github.com/dagger/dagger@latest found (v0.11.4), but does not contain package github.com/dagger/dagger/ci/internal/dagger
```

:sob:

shykes:
I think it's related to the newly added `.gitignore`. Perhaps the current CI module relies on an internal package *not* being git-ignored, and my new sub-module incorrectly messes with that fragile balance?

shykes:
> I think it's related to the newly added `.gitignore`. Perhaps the current CI module relies on an internal package _not_ being git-ignored, and my new sub-module incorrectly messes with that fragile balance?

Fixed

shykes:
I have one failing check left, and have no idea why. I'm not confident it's related to my change. Any guidance @matipan @gerhard ?

ci/std/go/dagger.json
```diff
@@ -0,0 +1,16 @@
...
+  "dependencies": [
+    {
+      "name": "daggy",
+      "source": "github.com/shykes/daggerverse/daggy@dfb1f91fa463b779021d65011f0060f7decda0ba"
+    },
+    {
+      "name": "wolfi",
+      "source": "github.com/shykes/daggerverse/wolfi@dfb1f91fa463b779021d65011f0060f7decda0ba"
+    }
+  ],
```
jedevc, line 13: I guess this is more of a philosophical question - but should our standard library here consume other modules in the daggerverse?

My personal opinion is we should avoid this, and try and be a "root" that others could pull from - but maybe you disagree.

ci/go.work
jedevc, line 1: Is this file required? Ideally, we should allow devs to setup their own `go.work` if it's not required to allow CI to build.

Also, `go.work` is a pain (citation, many PRs I opened in dagger/dagger to try and handle it) - I'd like to avoid increasing the complexity of the build setup.

ci/sdk_go.go
```diff
@@ -138,6 +139,7 @@ func gitPublish(ctx context.Context, opts gitPublishOpts) error {
 		base = dag.Container().From(consts.AlpineImage)
 	}
 
+	// FIXME: move this into std modules
```
jedevc, line 142: There's a few `FIXME`s in here - is the aim to do these before/after PR merge?

ci/sdk_go.go
```diff
@@ -85,8 +85,9 @@ func (t GoSDK) Publish(
 	githubToken *Secret,
 ) error {
 	return gitPublish(ctx, gitPublishOpts{
-		source:       "https://github.com/dagger/dagger.git",
-		alpineBase:   util.GoBase(t.Dagger.Source),
+		source: "https://github.com/dagger/dagger.git",
+		// FIXME: the go env is not alpine-based anymore, is that a problem?
+		alpineBase:   t.Dagger.Go().Env(),
```
jedevc, line 90: Ah, this is confusing. This should be renamed, it's the base container that the `sourceFilter` is run in.

So we could change this to `sourceContainer` or similar. `alpine` has nothing to do with it.

ci/go.work
shykes: IDE auto-complete breaks without it. Is the idea that Go developers might have their own `go.work` outside of the repo?

ci/std/go/dagger.json
shykes: I think that's a good goal. I was planning on pulling in these modules as a follow-up PR.

Also, I don't know why we're importing `daggy`.

ci/std/go/main.go
jedevc, line 1: So, struggling to put my finger on this, but. We seem to be trending towards each module is a file that defines really only a couple of functions.

Is there any reason we don't have *one* `std` module, that defines *multiple* types, like `graphql` and `go`, etc?

There's a few reasons why:
- Complexity - if we decide to have some go helpers shared between all the modules, it's a little bit trickier to handle this.
- Performance - loading a module has a cost. Even if we do the work (which we should) to speed this up, it's still not going to be 0.
- Lots of extra files - we have a separate go module for every dagger module, this adds some clutter. Additionally, to lint each of these, we're going to need to extend our engine linting config in `engine.go` to include each and every one of these.

I guess none of those are a hard-enough reason to *not* do this pattern, the big upside of the current method is that "go" can be imported by itself, with only those deps.

But also curious to hear your thoughts.

ci/go.work
jedevc: Yeah, so the pattern I've been using is to allow individual users to manage their own `go.work`s. That's also kinda weird, my vscode install very happily manages autocomplete without needing any go.works.

If we do decide to add a `go.work`, I guess this is a monorepo after all so is pretty reasonable, we should add it at the top-level, instead of in the `ci` subdirectory.

ci/sdk_go.go
shykes: After. There is a lot to do in this code base, and honestly it's quite painful to gain momentum. I would really prefer getting a first change in, and keep iterating from there. I'm already preparing my next patch.

ci/std/go/main.go
shykes: I think that's a good topic for discussion, as long as it doesn't hold up merging

shykes: The main reason for following the pattern of multiple modules, is that we tell our users to do it. So if it's painful for them, we should incur that pain also, to understand it.

I can confirm that adding these 2 modules was very painful for me, in a "death by a thousand cuts" kind of way.

shykes: > Lots of extra files - we have a separate go module for every dagger module, this adds some clutter. Additionally, to lint each of these, we're going to need to extend our engine linting config in engine.go to include each and every one of these.

Can you explain this one more?

ci/std/go/dagger.json
shykes: > Also, I don't know why we're importing daggy.

Removed

ci/go.work
shykes: Removed `go.work` entirely for now

ci/std/go/main.go
shykes: This is where @sagikazarmark would say: "create a separate `tests` submodule, to lint the linter :)"

shykes: > I think for now, we should add ci/std/go and such to there.

Done

ci/sdk_go.go
shykes: If you're OK with that, I will leave the FIXME in place, and will address it in a follow-up PR (which I look forward to)

jedevc:
Ahh, I thought this was ready to go? I'm kinda confused by some of the changes.

I rebased and pushed a few tiny little fixups, but.

We split our `dirdiff` - but it seems that we get weird errors with this when working with this? `failed to resolve dependency: select: module dep source root path "../../dirdiff" escapes root`. (I think this is probably a genuine bug in dagger). Also why isn't this in `std`?

Seems like python/typescript linters are failing as well?

---

What's the plan for this PR? I can pick up some of the failures/etc if you want me to, but we need to get all checks passing before we can merge.

jedevc:
Should have fixed all the issues above. I think this is ready to merge. Also noted a few genuine linting failures being picked up.

Will hold until you get a chance to glance over @shykes, and confirm.

ci/sdk_go.go
jedevc: Fixed this in a patch I pushed here.

jedevc (approved):

PR #7021, opened 2024-04-04
docs: Port Spring guide to use functions

vikram-dagger:
Fixes https://linear.app/dagger/issue/DOCS-191/spring

This commit ports the Spring guide at https://archive.docs.dagger.io/0.9/882813/build-test-publish-java-spring to use Dagger Functions.

PR #7463, opened 2024-05-25
docs: Remove extra go files that are not snippet

jpadams:
During the course of development/testing of the
go snippet, files like the go.mod, go.sum, etc
were committed, but aren't needed as snippets.
Introduced in a commit as part of
https://github.com/dagger/dagger/pull/7021


vikram-dagger (changes requested):
My understanding is that only the dagger.gen.go should be removed. See the discussion in https://github.com/dagger/dagger/pull/7459#issuecomment-2130317932. Also if required the `.gitignore` should be added back.

jpadams:
I'll wait for a bit more clarity on the approach from @vikram-dagger and @vito

jpadams:
Seems uneeded.

PR #7465, opened 2024-05-27
Object Prototype exploration

Xiot:
Exploring how object prototypes should play into the typescript sdk

### DIFF
--- a/sdk/typescript/introspector/test/testdata/objectParam/index.ts
+++ b/sdk/typescript/introspector/test/testdata/objectParam/index.ts
@@ -8,6 +8,10 @@ class Message {
   constructor(content: string) {
     this.content = content
   }
+
+  toUpperCase() {
+    return this.content.toUpperCase()
+  }
 }
 
 @object()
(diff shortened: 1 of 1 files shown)


sdk/typescript/introspector/test/testdata/objectParam/index.ts
```diff
@@ -19,7 +23,7 @@ class ObjectParam {
 
   @func()
   upper(msg: Message): Message {
-    msg.content = msg.content.toUpperCase()
+    msg.content = msg.toUpperCase()
```
Xiot, line 26: Because the prototype chains aren't respected, this call fails because `toUpperCase` is undefined.

Xiot: It still fails, since the prototype isn't attached.  

You just start with a simple object and just attach the properties onto it
https://github.com/Xiot/dagger/blob/87f82c38ad7aa91bce2d7974dfc098cf363f5f33/sdk/typescript/entrypoint/load.ts#L221

It also looks like having aliases on the input objects messes things up as well, as it does the mapping in the wrong order.  The code that handles the aliasing assumes that its only output types that are aliased.

PR #7473, opened 2024-05-27
chore(python): throw error if session exits unexpectedly

jedevc:


jedevc:
Note, the `SIGPIPE` failure is known, and is roughly similar to https://github.com/dagger/dagger/pull/7272.

jedevc:
Before merging, this probably needs an update to correct the python process stopping logic, so that we **don't** close stdout/stderr.

helderco (approved):
Fyi, I made a few adjustments to cleanup properly. See if it makes sense to you.

jedevc:
Aha, yes that's significantly easier to read and manage :tada: Thanks @helderco!

jedevc:
Hm it looks like `pytest_subprocess` mock object doesn't have a `stdin` property :thinking:

helderco:
Fixed it. I could try fixing the test (and muck around with FakeProcess), but it was simpler to be more flexible in prod code.

jedevc:
Okaaay this looks good now :tada:

jedevc:
DCO is failing because of https://github.com/dcoapp/app/issues/211.

PR #7474, opened 2024-05-27
Plain progress improvement follow ups

jedevc:
Follow up to #7272, specifically going to aim to correct some of @vito's comments in https://github.com/dagger/dagger/pull/7272#pullrequestreview-2071794451 and @aluzzardi's comments in https://github.com/dagger/dagger/pull/7385#issuecomment-2138365717.

Improvements:
- Don't skip displaying fast top-level progress rows, this was resulting in quick steps like `File.export` at the end from sometimes not displaying.
- Always simplify confusing `blob(...)` to their creators (we should always simplify `Internal` nodes if possible).
- Add an empty line before printing the final result.
- Omit the return type when printing `DONE` (we already omit the args for brevity, so can do the same thing here)
- Tidy up cloud link printing when logged in

Also some little refactorings that made sense as I worked on this.


jedevc:
For the new cloud links, this is what they look like. For plain progress:

image

For pretty progress:

image
image

---

I think this is *better* than before - it's still not perfect, but getting there.

The plain progress is definitely the tricky one here - I've chosen to attach the cloud link twice, once as a "virtual" log under the connecting span, and once at the very end. This makes it possible to easily find it on a finished job, as well as on an in-progress job.

jedevc:
Oh, and when a legacy dagger token is used:

image

gerhard:
This seems to be the only open PR before `v0.11.7` ships.

Would you like a review @jedevc, or are you OK to leave it for `v0.11.8` ?

jedevc:
I'd like a review for v0.11.7 if possible - the only major change since my last comment is to still default to not showing the cloud link in the TUI progress, except if `-v` is set.

gerhard (approved):
All looks good to me!

dagger.cloud link is the most obvious changed. Everything else remained largely unchanged, which is a great thing (it's a small improvement, fit for a patch).

Before: 

https://github.com/dagger/dagger/assets/3342/5000225b-a352-4cf6-999e-af1b4ef6ee0a

After:

https://github.com/dagger/dagger/assets/3342/22e08f87-8ec0-4aea-a219-515674a7667a

Issue #7212, opened 2024-04-27
Can't load local module at absolute path


shykes:
## Problem

When loading a local module, absolute paths are not supported.

Example:

```shell
dagger -m /foo/bar functions
```

Output:

```shell
  ! local module source root path is absolute: /foo/bar
```

## Solution

Allow loading local modules from a local path.

grouville:
@sipsma, is there any security / behavior concerns around allowing absolute paths ?
It seems that the restriction allows for just related path inside a module's context to be run

sipsma:
> @sipsma, is there any security / behavior concerns around allowing absolute paths ? It seems that the restriction allows for just related path inside a module's context to be run. I am currently checking #7476, where the tests seem to pass (after a rebase), just wanted to double check 🙏

Got back here https://github.com/dagger/dagger/pull/7476#issuecomment-2174413739

shykes:
I know this is already in progress. Just noting for the record that I ran into this issue again. Wanted to add a custom `DAGGER_MODULE=$(pwd)/...` in an `.envrc` file but not possible

shykes:
Thank you!

PR #7476, opened 2024-05-27
fix: allow absolute paths for local module paths

michaeljohnalbers:
* Closes https://github.com/dagger/dagger/issues/7212

### DIFF
--- a/core/integration/module_call_test.go
+++ b/core/integration/module_call_test.go
@@ -2,7 +2,9 @@ package core
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/containerd/containerd/platforms"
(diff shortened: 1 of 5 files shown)


shykes:
Thank you for the contribution!

FYI @gerhard @sipsma @jedevc @grouville @TomChv this would be a quick win, I just encountered #7212 and it's annoying 🙏

grouville:
Hi @michaeljohnalbers, thanks for the contribution 🙏 . I'll rebase it on your fork if it's ok with you🙏 There seems to be a few failures introduced, but I'll take a look today

image

grouville:
Waiting for @sipsma's  feedback on https://github.com/dagger/dagger/issues/7212#issuecomment-2174255325 prior approving, just to make sure this doesn't introduce unexpected behavior

sipsma:
@grouville  AFAICT this change should probably work in the current state of things, but generally speaking we want to ensure that values of paths we set are either consistently absolute or relative; so in this case the value is `RootSubpath` https://github.com/sipsma/dagger/blob/b213c4f88de0f4bfa164e047276d717024b78cb2/core/schema/modulesource.go#L49-L49

This is based on experience where trying to handle both absolute+relative paths in every place a given path is used ends up being a lot more work+thinking+bug-prone than just setting it to one or the other consistently in the first place.

IIRC this is why that restriction ended up there is the first place (plus being leftover from earlier iterations of the code where being non-absolute was a more strict requirement).

We can do this pretty easily by still checking whether the path is absolute but rather than return an error make the path relative to the caller's cwd in that case:
```go
if filepath.IsAbs(parsed.modPath) {
   cwdStat, err := query.Buildkit.StatCallerHostPath(ctx, ".", true)
   // if err != nil boilerplate

   parsed.modPath, err = filepath.Rel(cwdStat.Path, parsed.modPath)
   // if err != nil 
}
```

The other crucial thing to add are integ tests, probably in `TestModuleDaggerInit`, `TestModuleDaggerDevelop`, `TestModuleDaggerInstall` and new test in `module_call_test.go`, all of which verify that setting an abs path for the `-m` value works. Otherwise I'm sure some future tweak will accidentally break this again.

gerhard:
Hey @michaeljohnalbers 👋

Do you want to make the change & add the integration tests mentioned by @sipsma? We could get this in fairly quickly afterwards.

grouville:
> Hey @michaeljohnalbers 👋
> 
> Do you want to make the change & add the integration tests mentioned by @sipsma? We could get this in fairly quickly afterwards.

@michaeljohnalbers I have taken the liberty to implement those changes, and add some integration tests

grouville:
Failures are unrelated to implementation, due to flakiness on the Python side + signal test

core/schema/modulesource.go
```diff
@@ -42,7 +42,16 @@ func (s *moduleSchema) moduleSource(ctx context.Context, query *core.Query, args
...
+			cwdStat, err := query.Buildkit.StatCallerHostPath(ctx, ".", true)
+			if err != nil {
+				return nil, fmt.Errorf("failed to stat caller's current working directory: %w", err)
+			}
+
+			relPath, err := filepath.Rel(cwdStat.Path, parsed.modPath)
+			if err != nil {
+				return nil, fmt.Errorf("failed to make path relative: %w", err)
+			}
+			parsed.modPath = relPath
```
jedevc, line 54: Hm, just a note, I think this will probably cause issues with windows paths.

core/integration/module_call_test.go
```diff
@@ -1175,6 +1175,62 @@ func (ModuleSuite) TestCallByName(ctx context.Context, t *testctx.T) {
...
+
+			func (m *ModB) Fn(ctx context.Context) string {
+				return "hi from mod-b"
+			}
+			`,
+			}).
+			WithWorkdir("/work").
+			With(daggerExec("init")).
+			With(daggerExec("install", "--name", "foo", "/work/mod-a")).
+			With(daggerExec("install", "--name", "bar", "/work/mod-b"))
```
jedevc, line 1214: Couple notes:
- we shouldn't have absolute paths in the `dagger.json`, the test should check for this - if absolute paths are allowed here, then the module isn't really reusable by anyone else on a different fs layout
- we still shouldn't load modules from above the root - I *think* that's what's happening here? we'd want an equivalent of the [dep points out of root](https://github.com/dagger/dagger/blob/a9155f4b4e68c264e673bbb2a0271701d76c5eb3/core/integration/module_config_test.go#L116-L161) test

core/schema/modulesource.go
grouville: Thanks, you're totally right. I updated the code, with a `crossPlatformRelativePath` function  corresponding tests. I figured that we couldn't rely on the `filepath.ToSlash()` as the core/schema runs inside a linux container, or, the toslash implementation relies on the os to guess the separator to use

core/integration/module_call_test.go
grouville: > we shouldn't have absolute paths in the dagger.json, the test should check for this - if absolute paths are allowed here, then the module isn't really reusable by anyone else on a different fs layout

Thanks, I updated the test to include that case 🙏 

> we still shouldn't load modules from above the root - I think that's what's happening here? we'd want an equivalent of the [dep points out of root](https://github.com/dagger/dagger/blob/a9155f4b4e68c264e673bbb2a0271701d76c5eb3/core/integration/module_config_test.go#L116-L161) test
I think the test is actually correct, as dep is a relative child inside the /work top level module

I however update the deps points out of root to handle those cases. The error message are not the best, but, we error correctly on the cases I anticipated

grouville:
Race is unrelated from PR. Ready for review again

sipsma (approved):

PR #7477, opened 2024-05-27
feat(ts): keep function prototype during loading

### LINKED ISSUE #7446: 🐞 [Typescript] Object prototypes aren't preserved for fields
### What is the issue?

Custom values that are stored in a `@field` do not have their prototype chain preserved.

This leads to unexpected behaviour, especially since objects need to be created as a class with `@object` in order to be serialized at all.

### Dagger version

dagger v0.11.4 (registry.dagger.io/engine) darwin/arm64

### Steps to reproduce

I created a very simple module, that just passes a custom object between two modules.

```typescript
@object()
class RootModule {
  @func() test() {
    return new PModule(new PCheck(4))
  }
}

@object()
class PCheck {
  @field() value: number

  constructor(value: number) {
    this.value = value
  }
  get doubled() {
    return this.value * 2
  }
}

@object()
class PModule {
  @field() value: PCheck
  constructor(value: PCheck) {
    this.value = value
  }
  @func() print() {
    console.log('value', this.value, Object.getPrototypeOf(this.value))
    return this.value.doubled ?? 0
  }
}
```

Running the above with
```
dagger call test print
```

will return 0 and will print

```typescript
value { value: 4 } [Object: null prototype] {} 
```

The fact that the `value` field has a `null` prototype seems very unexpected.

If I change the implementation of `print` to attach the prototype, then things work as expected.
```typescript
  @func() print() {
    this.value = Object.setPrototypeOf(this.value, PCheck.prototype)
    console.log('value', this.value, Object.getPrototypeOf(this.value))
    return this.value.doubled ?? 0
  }
```

### Log output

_No response_

TomChv:
Fixes #7446 by loading the object as class if it exist inside the module.

If an object exists inside the module, it loads the class by calling his constructor so now public methods can be called inside the module.

### DIFF
--- a/core/integration/module_typescript_test.go
+++ b/core/integration/module_typescript_test.go
@@ -945,7 +945,7 @@ import { field, func, object } from "@dagger.io/dagger"
 @object()
 class Test {
   @func() 
-	test() {
+  test() {
     return new PModule(new PCheck(4))
   }
 }
@@ -953,7 +953,7 @@ class Test {
 @object()
 class PCheck {
   @field() 
-	value: number
+  value: number
 
   constructor(value: number) {
     this.value = value
@@ -967,13 +967,13 @@ class PCheck {
 @object()
 class PModule {
   @field() 
-	value: PCheck
+  value: PCheck
   
-	constructor(value: PCheck) {
+  constructor(value: PCheck) {
     this.value = value
   }
   @func() 
-	print() {
+  print() {
     return this.value.doubled ?? 0
   }
 }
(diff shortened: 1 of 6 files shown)


core/integration/module_typescript_test.go
```diff
@@ -926,3 +926,61 @@ class Alias {
...
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work").
+			With(daggerExec("init", "--name=test", "--sdk=typescript")).
+			With(sdkSource("typescript", `
+import { field, func, object } from "@dagger.io/dagger"
+
+@object()
+class Test {
+  @func() 
+	test() {
```
helderco, line 948: There's several lines in this code that aren't indented correctly. Can you fix the formatting to make it easier to read?

helderco (approved):

core/integration/module_typescript_test.go
TomChv: sure! Thanks for the catch!

Issue #6514, opened 2024-01-29
query builder (go): support selecting siblings


helderco:
This is different from https://github.com/dagger/dagger/issues/3577, which is about figuring out a DX for doing this in the SDKs.

Python’s query builder is from a third-party library, and supports everything from GraphQL. The one in Go is a custom solution, built entirely for our chainable API. Because of this and the issue above, it doesn’t support sibling selection. You can only select one field from an object.

However, dropping down to the query builder level can unblock certain use cases:
- https://github.com/dagger/dagger/issues/6515

PR #7479, opened 2024-05-28
cli: print module object fields with `dagger call`

helderco:
Fixes https://github.com/dagger/dagger/issues/6514
Fixes https://github.com/dagger/dagger/issues/6515

Previously, when a `dagger call` chain ended in an object, the result was inconsistent and not very useful:
- Without any functions, it would show `--help`
- `Container`, `Directory` and `File` core types evaluated the pipeline and returned a "XXX evaluated." message
- For custom user types, an error was returned: "return type XXX requires a sub-command"

With this change, in all of those cases, **the CLI finds all functions without required arguments that return a scalar/primitive value, and prints them**. This includes a module's main object, when calling `dagger call` without any more arguments.

> [!NOTE]
> At minimum, a `_type` field is added to the response so there's always something to show. 
> When not a TTY, `--json` format is used instead of the more human-readable default format.

See following examples.

### Before

```console
❯ dagger call
[output of --help]
```

### After

```console
❯ dagger call
_type: Playground
baseImage: python:3.12-alpine
cacheKey: playground-cache
```

### Before

```console
❯ dagger call bar
Error: invalid selection for command "bar": return type "PlaygroundBar" requires a sub-command
Run 'dagger call bar --help' for usage.
exit status 1
```

### After

```console
❯ dagger call bar
_type: PlaygroundBar
```

### Before

```console
❯ dagger call container
Container evaluated. Use "dagger call container --help" to see available sub-commands.
```

### After

```console
❯ dagger call container
_type: Container
defaultArgs:
    - /bin/sh
entrypoint: []
mounts: []
platform: linux/arm64
stderr: ""
stdout: |
    NAME="Alpine Linux"
    ID=alpine
    VERSION_ID=3.20.0
    PRETTY_NAME="Alpine Linux v3.20"
    HOME_URL="https://alpinelinux.org/"
    BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
user: ""
workdir: ""
```

### Before

```console
❯ dagger call container directory --path=/etc/ssl
Directory evaluated. Use "dagger call echo directory --help" to see available sub-commands.
```

### After

```console
❯ dagger call container directory --path=/etc/ssl
_type: Directory
entries:
    - cert.pem
    - certs
    - ct_log_list.cnf
    - ct_log_list.cnf.dist
    - openssl.cnf
    - openssl.cnf.dist
    - private
```

### Before

```console
❯ dagger call container file --path=/etc/os-release
File evaluated. Use "dagger call echo file --help" to see available sub-commands.
```

### After

```console
❯ dagger call container file --path=/etc/os-release
_type: File
name: os-release
size: 188
```

### Before

```console
❯ dagger call containers
Error: "containers" requires a sub-command
Run 'dagger call foo containers --help' for usage.
exit status 1
```

### After

```console
❯ dagger call containers
- _type: Container
  defaultArgs:
    - /bin/sh
  entrypoint: []
  mounts: []
  platform: linux/arm64
  stderr: ""
  stdout: |
    NAME="Alpine Linux"
    ID=alpine
    VERSION_ID=3.20.0
    PRETTY_NAME="Alpine Linux v3.20"
    HOME_URL="https://alpinelinux.org/"
    BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
  user: ""
  workdir: ""
- _type: Container
  defaultArgs:
    - /bin/sh
  entrypoint: []
  mounts: []
  platform: linux/amd64
  stderr: ""
  stdout: |
    NAME="Alpine Linux"
    ID=alpine
    VERSION_ID=3.20.0
    PRETTY_NAME="Alpine Linux v3.20"
    HOME_URL="https://alpinelinux.org/"
    BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
  user: ""
  workdir: ""
- _type: Container
  defaultArgs:
    - /bin/sh
  entrypoint: []
  mounts: []
  platform: linux/386
  stderr: ""
  stdout: |
    NAME="Alpine Linux"
    ID=alpine
    VERSION_ID=3.20.0
    PRETTY_NAME="Alpine Linux v3.20"
    HOME_URL="https://alpinelinux.org/"
    BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
  user: ""
  workdir: ""
```

### DIFF
--- a/.changes/unreleased/Changed-20240628-141105.yaml
+++ b/.changes/unreleased/Changed-20240628-141105.yaml
@@ -1,0 +1,6 @@
+kind: Changed
+body: 'cli: print module object fields with `dagger call`'
+time: 2024-06-28T14:11:05.367066Z
+custom:
+    Author: helderco
+    PR: "7479"
--- a/cmd/dagger/call.go
+++ b/cmd/dagger/call.go
@@ -7,120 +7,91 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
 )
 
+var outputFormat string
 var outputPath string
 var jsonOutput bool
 
 var callCmd = &FuncCommand{
 	Name:  "call [options]",
 	Short: "Call a module function",
-	Long: strings.ReplaceAll(`Call a module function and print the result.
-
-If the last argument is either a Container, Directory, or File, the pipeline
-will be evaluated (the result of calling ´sync´) without presenting any output.
-Providing the ´--output´ option (shorthand: ´-o´) is equivalent to calling
-´export´ instead. To print a property of these core objects, continue chaining
-by appending it to the end of the command (for example, ´stdout´, ´entries´, or
-´contents´).
-`,
-		"´",
-		"`",
-	),
-	Example: strings.TrimSpace(`
-dagger call test
-dagger call build -o ./bin/myapp
-dagger call lint stdout
-`,
-	),
 	Init: func(cmd *cobra.Command) {
+		cmd.PersistentFlags().StringVarP(&outputPath, "output", "o", "", "Save the result to a local file or directory")
+
 		cmd.PersistentFlags().BoolVar(&jsonOutput, "json", false, "Present result as JSON")
-		cmd.PersistentFlags().StringVarP(&outputPath, "output", "o", "", "Path in the host to save the result to")
 	},
-	OnSelectObjectLeaf: func(c *FuncCommand, name string) error {
-		switch name {
-		case Container, Directory, File:
-			if outputPath != "" {
-				c.Select("export")
-				c.Arg("path", outputPath)
-				if name == File {
-					c.Arg("allowParentDirPath", true)
-				}
-				return nil
-			}
-			c.Select("sync")
-		default:
-			return fmt.Errorf("return type %q requires a sub-command", name)
+	OnSelectObjectLeaf: func(fc *FuncCommand, obj functionProvider) error {
+		names := make([]string, 0)
+		for _, f := range GetLeafFunctions(obj) {
+			names = append(names, f.Name)
 		}
-		return nil
-	},
-	BeforeRequest: func(_ *FuncCommand, _ *cobra.Command, modType *modTypeDef) error {
+		if len(names) > 0 {
+			// FIXME: Consider adding a method to the query builder specifically
+			// for multiple selections to avoid this workaround. Even if there's
+			// just one field it helps to show the field's name, not just the
+			// value, and multiple selection allows it because it binds to the
+			// parent selection.
+			if len(names) == 1 {
+				names = append(names, "")
+			}
+			fc.Select(names...)
+			outputFormat = "yaml"
+		}
 		return nil
 	},
 	AfterResponse: func(c *FuncCommand, cmd *cobra.Command, modType *modTypeDef, response any) error {
-		switch modType.Name() {
-		case Container, Directory, File:
-			if outputPath != "" {
-				logOutputSuccess(cmd, outputPath)
-				return nil
-			}
-
-			// Just `sync`, don't print the result (id), but let user know.
-
-			// TODO: This is only "needed" when there's no output because
-			// you're left wondering if the command did anything. Otherwise,
-			// the output is sent only to progrock (TUI), so we'd need to check
-			// there if possible. Decide whether this message is ok in all cases,
-			// better to not print it, or to conditionally check.
-			cmd.PrintErrf("%s evaluated. Use \"%s --help\" to see available sub-commands.\n", modType.Name(), cmd.CommandPath())
-			return nil
-		default:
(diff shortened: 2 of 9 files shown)


helderco:
Added a `_type` field to the response with the object's name so there's always something to show that's useful, instead of returning an error.

sourishkrout:
Heads up, when the final leaf happens to be a binary `File` the `contents` basically dumps the file into the terminal. Bit of usability issue.

shykes:
Could we change the default to pretty-printed json, with (perhaps) a global flag to change the format to yaml? I think it would be more convenient for scripting, piping into `jq` etc

helderco:
> Could we change the default to pretty-printed json, with (perhaps) a global flag to change the format to yaml? I think it would be more convenient for scripting, piping into `jq` etc

It's already supported via the `-f json` flag (replaces `--json`). The reason yaml is the default it's because it was made to be human readable, while json is targeted for machines.

shykes:
> The reason yaml is the default it's because it was made to be human readable, while json is targeted for machines

It's true that it makes sense to pretty-print. Then I have 2 requests:

1. Can we keep `--json | -j` as a convenience for `--format json`? It deserves special treatment as json output is the gold standard when scripting, and we are trying to improve the scripting experience. It would certainly be less keystrokes for me..

2. If the default output is "human friendly pretty-print", we must protect the freedom to iterate on that output. If our first implementation of a pretty-print is simply yaml, as a convenience, then we should tweak the output to just enough that it is *not* compliant yaml - just yaml-ish. If never make the promise that the output is compatible yaml, we won't break that promise later. This keeps the door wide open for innovating on the best output in the future, TUI-style. (thanks @aluzzardi for the suggestion)

shykes:
>  If never make the promise that the output is compatible yaml, we won't break that promise later. This keeps the door wide open for innovating on the best output in the future, TUI-style

Any thoughts on this part @vito ? :)

sourishkrout:
> 1. Can we keep `--json | -j` as a convenience for `--format json`?

I'd be fantastic if there was also a way to set this via an ENV var too, please.

vito:
> Any thoughts on this part @vito ? :)

Agree - ideally the TUI can render objects however it wants without worrying about compatibility. But it seems like compatibility should only matter when the output is going to a machine, so maybe we just infer that from `stdout`?

* If stdout is a pipe/file, emit JSON.
* If stdout is a TTY, emit some human-compatible format.

At that point we could probably still do YAML as an implementation detail, since you if you try to capture it you'll get JSON instead (...which is also valid YAML! :nerd_face:) - but if we're still concerned (or just Team No-YAML) we could do some small tweak like you suggest.

helderco:
> If the default output is "human friendly pretty-print", we must protect the freedom to iterate on that output. If our first implementation of a pretty-print is simply yaml, as a convenience, then we should tweak the output to just enough that it is _not_ compliant yaml - just yaml-ish. If never make the promise that the output is compatible yaml, we won't break that promise later. This keeps the door wide open for innovating on the best output in the future, TUI-style. (thanks @aluzzardi for the suggestion)

How about I revert the `--format` flag and keep `--json`, using `yaml` as an implementation detail? Switching to `--json` automatically based on tty or not too.

helderco:
> How about I revert the `--format` flag and keep only `--json`, using `yaml` as an implementation detail? Switching to `--json` automatically based on tty or not too.

I made this change. I don't think it's necessary to make sure the default format is invalid yaml. "yaml" isn't mentioned anywhere nor is given as an option so it's an implementation detail that can't be relied on.

> Heads up, when the final leaf happens to be a binary `File` the `contents` basically dumps the file into the terminal. Bit of usability issue.

Not sure what to do with this. Is it a blocker or can it be improved in a follow up?

shykes:
> > How about I revert the `--format` flag and keep only `--json`, using `yaml` as an implementation detail? Switching to `--json` automatically based on tty or not too.
> 
> I made this change. 

👍🙏

> I don't think it's necessary to make sure the default format is invalid yaml. "yaml" isn't mentioned anywhere nor is given as an option so it's an implementation detail that can't be relied on.

👍🙏

> > Heads up, when the final leaf happens to be a binary `File` the `contents` basically dumps the file into the terminal. Bit of usability issue.
> 
> Not sure what to do with this. Is it a blocker or can it be improved in a follow up?

Ideally `contents` would return a type like `ByteStream` or `Bytes`. If we can figure out the correct behavior for a leaf of that future type, then we can pretend `contents` has that type, and adopt that behavior now.

shykes:
> > > Heads up, when the final leaf happens to be a binary `File` the `contents` basically dumps the file into the terminal. Bit of usability issue.
> Ideally `contents` would return a type like `ByteStream` or `Bytes`. If we can figure out the correct behavior for a leaf of that future type, then we can pretend `contents` has that type, and adopt that behavior now.

@helderco I would skip `File.contents` for now. That way we can always revisit later without breaking anyone. That way we can merge this :)

shykes (approved):

helderco:
Updated and rebased. Waiting for checks to pass.

PR #7480, opened 2024-05-28
docs: Add recipe for directory copy with exclusions to cookbook 

vikram-dagger:


docs/current_docs/manuals/developer/cookbook/snippets/copy-filter-dir/typescript/index.ts
```diff
@@ -0,0 +1,34 @@
...
+     * Directory exclusion pattern
+     */
+    excludeDirectory?: string,
+    /**
+     * File exclusion pattern
+     */
+    excludeFile?: string,
+  ): Container {
+    let filteredSource = source
+    if (excludeDirectory != null) {
```
TomChv, line 24: You should use `!` instead, to also test `undefined`

```suggestion
    if (!excludeDirectory) {
```

docs/current_docs/manuals/developer/cookbook/snippets/copy-filter-dir/typescript/index.ts
```diff
@@ -0,0 +1,34 @@
...
+      filteredSource = filteredSource.withoutDirectory(excludeDirectory)
+    }
+    if (excludeFile != null) {
```
TomChv, line 27: You should usse `!` to also test `undefined`

```suggestion
    if (!excludeFile) {
```

docs/current_docs/manuals/developer/cookbook/snippets/copy-filter-dir/typescript/index.ts
TomChv, line 1: To improve code clarity, could you add some blanklines between statements?

Like

```ts
if (xxx) {
  // blabla
}

if (xxx) {
   blabla
}

return dag.xxx
```

TomChv (commented):
Thanks for this guide! I left few comments related to the TS implementation.

TomChv:
@vikram-dagger Thanks for the changes, however I cannot find it in the docs preview, could I get the link?

TomChv:
@jpadams Thanks

TomChv (approved):
LGTM

vikram-dagger:
> Agreed that we need views docs. Do we have anything written @vikram-dagger ? Agreed that we should put views above excludes so folks don't upload a big directory and then exclude it (e.g. node_modules/)

No, we don't have any docs on views. I'll open an issue to add this. I'll also update this recipe to use views instead of the current approach.

vikram-dagger:
@jpadams @kpenfound @marcosnils before I remove this recipe, can you confirm that there are no use cases where this might be useful ie. that views are the better approach for all scenarios?

levlaz:
Hey team, 

On Wed, May 29, 2024 at 12:29:42AM -0700, vikram-dagger wrote:
> > Agreed that we need views docs. Do we have anything written @vikram-dagger ? Agreed that we should put views above excludes so folks don't upload a big directory and then exclude it (e.g. node_modules/)
> 
> No, we don't have any docs on views. I'll open an issue to add this. I'll also update this recipe to use views instead of the current approach.

FYI, I opened this already yesterday: https://github.com/dagger/dagger/issues/7489

Issue #7433, opened 2024-05-21
✨ Consider Pinning Node to LTS release 


levlaz:
### What are you trying to do?

A user reported that they were surprised to see us using a non-LTS version of Node in our Typescript SDK. 

We should consider pinning our TypeScript SDK to the LTS version of node as we approach "dagger 1.0" 

### Why is this important to you?

Since we expect people to adopt this SDK in their production pipelines, I think we will have a lot less issues long term if we stick to a stable version of node. Node's own documentation recommends using only LTS releases for production applications since they will come with 30 months of support instead of just 6 (for odd numbered releases)

> Major Node.js versions enter Current release status for six months, which gives library authors time to add support for them. After six months, odd-numbered releases (9, 11, etc.) become unsupported, and even-numbered releases (10, 12, etc.) move to Active LTS status and are ready for general use. LTS release status is "long-term support", which typically guarantees that critical bugs will be fixed for a total of 30 months. Production applications should only use Active LTS or Maintenance LTS releases.

### How are you currently working around this?

N/A

TomChv:
Thanks for reporting, I can take care of the changes and see if it break anything this week.

TomChv:
As explained in #7481, instead of pinning `Node` to LTS release, we'll support dynamic version so the user can configure which version he wants to use by setting the field `engine.node` in its package.json.

By default, we'll use the LTS version, probably 20.14 until v22 become the active LTS release.

Here's a copy of the discussion in #7481 

> Node upgrades / downgrades are always tricky since its not just the user code that you have to be worried about. Internally we use 20.11, and one of our devs was setting up a new machine and he had 20.13 installed. Things broke.

> I like the ability to configure which version of node is used, while having a minimum supported version. Something like this will probably be needed for the bun runtime since that is adding new features on a quicker cadence

> Alright, I'm closing this PR for now, as you said, it's better to allow node configuration which is possible by checking the engine.node field of the package.json.

> I'll take care of that later, I'll put a comment in the issue as a reminder

noah-troncoso:
Hi. So, the use of version 21 is actually breaking me right now. Is there a workaround I can use to get Dagger to use version 20 until this is released?

PR #7481, opened 2024-05-28
feat(ts-sdk): use lts version in runtime

TomChv:
This PR is an attempt to fixe #7433, if the CI is green, we can merge it, otherwise I'll dig into what breaking change happens with this new version.

### DIFF
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -13,10 +13,10 @@ import (
 
 const (
 	bunVersion  = "1.0.27"
-	nodeVersion = "21.3"
+	nodeVersion = "22.2.0-alpine"
 
 	nodeImageDigest = "sha256:3dab5cc219983a5f1904d285081cceffc9d181e64bed2a4a18855d2d62c64ccb"
-	bunImageDigest  = "sha256:82d3d3b8ad96c4eea45c88167ce46e7e24afc726897d48e48cc6d6bf230c061c"
+	bunImageDigest  = "sha256:94567107148ac59f1eb2ad9b7c1db03f1a1a12d28717b29eda0535aa3bd2f71e"
 
 	nodeImageRef = "node:" + nodeVersion + "-alpine@" + nodeImageDigest
 	bunImageRef  = "oven/bun:" + bunVersion + "-alpine@" + bunImageDigest
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -13,9 +13,10 @@ import (
 
 const (
 	bunVersion  = "1.0.27"
-	nodeVersion = "22.2.0-alpine"
+	// 20.14.0 is the latest LTS version of node as of 2023-10-24, it ends on 2026-04-30
+	nodeVersion = "iron-alpine3.19"
 
-	nodeImageDigest = "sha256:94567107148ac59f1eb2ad9b7c1db03f1a1a12d28717b29eda0535aa3bd2f71e"
+	nodeImageDigest = "sha256:2d0ce60a5c7bfa2bcdcfa7d463d8f8b7b16cab55051562606a9dc32c8a8169a9"
 	bunImageDigest  = "sha256:82d3d3b8ad96c4eea45c88167ce46e7e24afc726897d48e48cc6d6bf230c061c"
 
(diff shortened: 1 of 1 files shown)


TomChv:
> Since this is an upgrade of versions, I don't think we need to do a breaking change for this - unless there's something specifically very different in this new version.

Yes it's not supposed to be a breaking change but I wanna verify that it doesn't break the TS SDK

TomChv:
@jedevc There's no breaking change with this PR, that's nice to know!

Xiot:
From the node docs:
> Production applications should only use Active LTS or Maintenance LTS releases.

https://nodejs.org/en/about/previous-releases

Node 22 doesn't become active LTS until the end of October
https://github.com/nodejs/release#release-schedule

sdk/typescript/runtime/main.go
```diff
@@ -13,9 +13,10 @@ import (
 
 const (
 	bunVersion  = "1.0.27"
-	nodeVersion = "21.3"
+	// 20.14.0 is the latest LTS version of node as of 2023-10-24, it ends on 2026-04-30
+	nodeVersion = "iron-alpine3.19"
```
jedevc, line 17: Can we use the `20.14.0-alpine3.19` tag please instead.

jedevc:
Hm, this is now a version *downgrade*. Looking through https://nodejs.org/en/blog/announcements/v21-release-announce, it looks like 21 doesn't bring *that* many user-facing changes.

That said, I'm not a js programmer - but the downgrade would definitely be unexpected. One option would be to delay something like this to v0.12.0 if we're concerned about breaking changes (just because our CI doesn't hit a breaking change doesn't mean it isn't there) cc @helderco

Potentially, we should also allow configuring the node version used for modules by having a field in `dagger.json` - I know we've also discussed this for python/go (though go makes this a *right* pain).

TomChv:
> That said, I'm not a js programmer - but the downgrade would definitely be unexpected. One option would be to delay something like this to v0.12.0 if we're concerned about breaking changes (just because our CI doesn't hit a breaking change doesn't mean it isn't there) cc @helderco

Yeah that's right, maybe the best thing to do is to close this PR and wait for october until we can put v22 when it's the active production release?

jedevc:
I think the downgrade is probably alright, but we shouldn't do it until v0.12.0 if we do.

Or we can upgrade to v22 now (upgrades are find whenever IMO), and just temporarily suffer while it's not an LTS version (I'm not sure what kind of issue this would give us).

TomChv:
> Or we can upgrade to v22 now (upgrades are find whenever IMO), and just temporarily suffer while it's not an LTS version (I'm not sure what kind of issue this would give us).

In the CI everything looked fine but yeah I don't know either, normally node tries to be backward compatible so it should be okay.

I'll keep the Pr on hold for now and take care of that later in the week

Xiot:
Node upgrades / downgrades are always tricky since its not just the user code that you have to be worried about.  Internally we use `20.11`, and one of our devs was setting up a new machine and he had `20.13` installed.  Things broke.

I like the ability to configure which version of node is used, while having a minimum supported version.  Something like this will probably be needed for the `bun` runtime since that is adding new features on a quicker cadence

TomChv:
> Node upgrades / downgrades are always tricky since its not just the user code that you have to be worried about. Internally we use 20.11, and one of our devs was setting up a new machine and he had 20.13 installed. Things broke.

Alright, I'm closing this PR for now, as you said, it's better to allow node configuration which is possible by checking the `engine.node` field of the package.json.

I'll take care of that later, I'll put a comment in the issue as a reminder

helderco:
> Potentially, we should also allow configuring the node version used for modules by having a field in `dagger.json` - I know we've also discussed this for python/go (though go makes this a _right_ pain).

Python supports this via `pyproject.toml`. TypeScript can do the same, just as it's able to select a different runtime (bun), with `package.json`.

Xiot:
You should probably allow the user to configure the package manager as well.  Having that switch out from beneath you (like it recently did) can cause issues as well.

TomChv:
> Python supports this via pyproject.toml. TypeScript can do the same, just as it's able to select a different runtime (bun), with package.json.

Yes but do Python has no breaking change between package manager version? Because that's the case for TypeScript (see https://github.com/dagger/dagger/pull/7518)

PR #7491, opened 2024-05-29
feat: add `withName` function to File

TomChv:
Fixes #6474 

~~⚠️ This PR also includes @jedevc commit from [f7b3a9c](https://github.com/dagger/dagger/pull/7438/commits/f7b3a9cbb9d6ca5cc53bc280d9352cebdba97258)[f7b3a9c](https://github.com/dagger/dagger/pull/7438/commits/f7b3a9cbb9d6ca5cc53bc280d9352cebdba97258) to handle global client generation.~~ Merged by #7492 

Fixes OSS-62

PR #7492, opened 2024-05-29
ci: add sdk all group

jedevc:
When we removed mage, we removed the easy helpers that allowed generating/linting/etc on all SDKs. That meant that making API changes became much more fiddly.

This patch restores the "all" group, but this time directly in the dagger module. One day, we might remove this with a script, but we don't have that right now.


jedevc:
I split this out from https://github.com/dagger/dagger/pull/7438, since it's useful for other PRs, like https://github.com/dagger/dagger/pull/7491.

TomChv (approved):

PR #7493, opened 2024-05-29
Add php runtime

carnage:
Very minimal attempt at a PHP runtime for module support. 

To the best of my knowledge, the go part of this PR is complete & working. 

Outstanding is to write the PHP code to 

- Create a module entry point
- Scan the user module src directory for dagger functions
- Retrieve the call information from the dagger engine
- Dispatch the call to the detected function
- Return the result the the dagger engine


carnage:
@dragoonis

sdk/php/src/Command/CodegenCommand.php
```diff
@@ -30,12 +32,31 @@ public function __construct()
...
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $io = new SymfonyStyle($input, $output);
-        $client = $this->daggerConnection->connect();
 
-        $schema = (new SchemaGenerator($client))->getSchema();
+
+        if ($input->getOption('schema-file') !== null && file_exists($input->getOption('schema-file'))) {
+            $fileContents = file_get_contents($input->getOption('schema-file'));
+            $schemaArray = json_decode($fileContents, true);
```
dragoonis, line 53: Change this to
```
json_validate()
https://www.php.net/json_validate
We need PHP 8.3 on this, now
```

sdk/php/template/composer.json
```diff
@@ -0,0 +1,28 @@
...
+  "authors": [
+  ],
+  "repositories": [
+    {
+      "type": "path",
+      "url": "./sdk"
+    }
+  ],
+  "require": {
+    "php": "^8.1",
```
dragoonis, line 15: bump this to PHP 8.3 because of introducing `json_validate()`

sdk/php/src/Command/CodegenCommand.php
charjr: From [php.net/json_validate](https://www.php.net/json_validate)

> Caution
> 
> Calling json_validate() immediately before [json_decode()](https://www.php.net/manual/en/function.json-decode.php) will unnecessarily parse the string twice, as [json_decode()](https://www.php.net/manual/en/function.json-decode.php) implicitly performs validation during decoding.
> 
> json_validate() should therefore only be used if the decode JSON payload is not immediately used and knowing whether the string contains valid JSON is needed.

https://github.com/dagger/dagger/blob/c5bcedfab3a0f6bd32fd4bb80b8e6b5612037a8a/sdk/php/src/Command/CodegenCommand.php#L53-L54

I might be misunderstanding the intent, but the decoded JSON payload is immediately used. Is it still worth using `json_validate` in this instance?

sdk/php/src/Command/EntrypointCommand.php
```diff
@@ -0,0 +1,239 @@
...
+        string $parentName
+    ): int {
+        $errorOutput = $output instanceof ConsoleOutputInterface ?
+            $output->getErrorOutput() :
+            $output;
+
+        $className = "DaggerModule\\$parentName";
+        $functionName = $functionCall->name();
+        $class = new $className();
+        $class->client = $this->daggerClient;
```
helderco, line 119: Just a note for later.

By convention, our SDKs provide a **global** client connection under the name `dag`. See https://dagger.io/blog/simpler-reusability-with-the-global-client.

## Python

https://github.com/dagger/dagger/blob/aef8596314ad777085f11580c6911346f57c806c/sdk/python/src/dagger/client/gen.py#L7100-L7101

## TypeScript

https://github.com/dagger/dagger/blob/aef8596314ad777085f11580c6911346f57c806c/sdk/typescript/api/client.gen.ts#L9205

## Go

https://github.com/dagger/dagger/blob/aef8596314ad777085f11580c6911346f57c806c/cmd/codegen/generator/go/templates/src/_dagger.gen.go/defs.go.tmpl#L143-L151

sdk/php/src/Command/EntrypointCommand.php
```diff
@@ -0,0 +1,239 @@
...
+        if ($result instanceof Dagger\Client\IdAble) {
+            $result = (string) $result->id();
+        }
+
+        $functionCall->returnValue(new Dagger\Json(json_encode($result)));
+
+        return Command::SUCCESS;
+    }
+
+    private function getTypeDef(Type $type): TypeDef
```
helderco, line 151: If a type accepts null values, or has a default but you can't serialize it or report it to Dagger, you should set `$typeDef->withOptional(true)`.

Enums and custom scalars are easy to support too:

https://github.com/dagger/dagger/blob/aef8596314ad777085f11580c6911346f57c806c/sdk/python/src/dagger/mod/_converter.py#L111-L115

Examples you can try from core:
- Scalar: `Platform`
- Enum: `NetworkProtocol`

sdk/php/src/ValueObject/DaggerArgument.php
```diff
@@ -0,0 +1,59 @@
...
+                );
+            $default = $betterReflection->getDefaultValue();
+            return new Json(json_encode(
+                $default instanceof IdAble ? (string) $default->id() : $default
+            ));
+        }
+
+        if ($parameter->allowsNull()) {
+            return new Json(json_encode(null));
+        }
```
helderco, line 55: If an argument allows nulls make sure to register it with `$typeDef->withOptional(true)`.

sdk/php/src/ValueObject/Type.php
```diff
@@ -0,0 +1,102 @@
...
+namespace Dagger\ValueObject;
+
+use Dagger\Client\IdAble;
+use Dagger\TypeDefKind;
+use ReflectionClass;
+use ReflectionNamedType;
+use ReflectionType;
+use RuntimeException;
+
+//@TODO support union/intersection types
```
helderco, line 14: Dagger doesn't support [union types](https://graphql.org/learn/schema/#union-types). If this is referring to unions in PHP only, I'd be curious to know how that translates. In Python and TypeScript, the only union is with nulls (`<type> | None`). Could also be unions with similar types that map to the same thing when serialized (e.g., str and bytes, both serialized as `String` in the API).

helderco (commented):
Looking good! 🚀

I personally don't like to repeat in a prefix the namespace. So I'd rather see `#[Dagger\Object]` than `#[DaggerObject]` because that's really `Dagger\DaggerObject`. Not a blocker, just an opinion. 🙂

I know `DaggerObject` is technically in `Dagger\Attribute` but we try to put all user facing types under the root `Dagger` namespace so I'm commenting as if you'd done that. 😁

charjr:
> I personally don't like to repeat in a prefix the namespace. So I'd rather see `#[Dagger\Object]` than `#[DaggerObject]` because that's really `Dagger\DaggerObject`. Not a blocker, just an opinion. 🙂


I completely agree with that opinion but the repetition is a result of me working around PHP keywords. `object` and `function` are keywords in PHP so they cannot be used as class names. 

Another option is to add a character such as an underscore to avoid keyword conflicts; that way we'd get `Dagger\_Object` but honestly, I'm not a fan of that either.

sdk/php/dagger.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "php-sdk",
+  "sdk": "go",
+  "include": [
+    "."
+  ],
```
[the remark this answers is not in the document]
charjr: Added what we believe is safe to exclude, we haven't adjusted the "include" though.

sdk/php/src/Command/EntrypointCommand.php
charjr: For consistency I've now renamed the `daggerClient` variable in the `EntryPointCommand` to `dag`.

The current solution of `$class->client = $this->daggerClient;` is more of a quick hack to get it working than a final implementation. 

Global variables don't tend to be used often in PHP so may be confusing for PHP developers. The plan is to have the SDK automatically inject the `Client` dependency into `DaggerObjects` requiring it. We plan to implement this shortly after sorting out custom objects.

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,155 @@
...
+
+	ctr = ctr.
+		WithExec([]string{
+			"./install-composer.sh",
+		}).
+		WithExec([]string{
+			"php", "composer.phar", "install",
+		})
+
+	filepath.Join(ModSourceDirPath, subPath, RuntimeExecutablePath)
```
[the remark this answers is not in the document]
charjr: Fixed:

https://github.com/dagger/dagger/blob/5113a2271318beca8639370a908d46f822266ccd/sdk/php/runtime/main.go#L144-L148

I combined it with the last line which was originally just hardcoding "dagger" which is what the RuntimeExecutablePath was.

helderco:
> I completely agree with that opinion but the repetition is a result of me working around PHP keywords. `object` and `function` are keywords in PHP so they cannot be used as class names.
> 
> Another option is to add a character such as an underscore to avoid keyword conflicts; that way we'd get `Dagger\_Object` but honestly, I'm not a fan of that either.

In Python, `object` and `enum` are built-ins so i named the decorators `@object_type` and `@enum_type` instead, so `#[ObjectType]` is an option. If `Function` is reserved in PHP, it could be `#[Func]` like in TS.

charjr:
> `#[ObjectType]` is an option. If `Function` is reserved in PHP, it could be `#[Func]` like in TS.

@carnage and I just reviewed the naming conventions:

We changed `#[DaggerArgument]` to `#[Argument]`.

We have not changed`#[DaggerFunction]`; we discussed it and no other name matches as closely to Dagger's Domain Specific Language. We always refer to them as "Dagger Functions" whether it is in documentation or simply informal discussions.

We have not currently changed `#[DaggerObject]`; there is no "Dagger Object" in the DSL, so this one is more open to change. We currently left this alone because it ties in nicely with `#[DaggerFunction]`.

sdk/php/dagger.json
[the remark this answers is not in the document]
helderco: Can you flip it and figure out what specifically you need to include, rather than rely on excludes?

sdk/php/src/Command/EntrypointCommand.php
charjr: We have now also defined `DAG` as a constant, making it globally accessible.

It has been named in uppercase as that is the standard PHP naming convention for constants.

https://github.com/dagger/dagger/blob/cb6b89da641d2d883682ba037951f32ea3b3f678/sdk/php/template/dagger#L20-L27

helderco: > Global variables don't tend to be used often in PHP so may be confusing for PHP developers.

I know I said "global" but it doesn't have to be a literal `global` variable. It can (and should) be imported. It just needs to setup it's own connection automatically and return a (shared) client instance.

sdk/php/src/Command/EntrypointCommand.php
```diff
@@ -0,0 +1,224 @@
...
+                ));
+            case TypeDefKind::OBJECT_KIND:
+                if ($type->isIdable()) {
+                    return $typeDef->withObject($type->getShortName());
+                }
+
+                throw new RuntimeException(sprintf(
+                    'Currently cannot handle custom classes: %s',
+                    $type->name
+                ));
```
helderco, line 165: This currently only works for objects that are already in the API but it's easy to support other objects in the same module. At this point, it's just handling references like in the return type of a method that returns the object, so the `withObject` here just needs to know the name of the class, and to check if it's a `[#DaggerFunction]`.

sdk/php/src/ValueObject/Type.php
```diff
@@ -0,0 +1,111 @@
...
+use Dagger\Exception\MissingSubtype;
+use Dagger\Exception\UnsupportedType;
+use Dagger\TypeDefKind;
+use DomainException;
+use ReflectionClass;
+use ReflectionNamedType;
+use ReflectionType;
+use RuntimeException;
+
+//@TODO support intersection types
```
helderco, line 18: You mean "union" types? Those don't exist in Dagger.

sdk/php/src/ValueObject/Type.php
```diff
@@ -0,0 +1,111 @@
...
+            $parents = class_parents($nameOfType);
+            assert(is_array($parents));
+
+            if (in_array(AbstractScalar::class, $parents, true)) {
+                return TypeDefKind::SCALAR_KIND;
+            }
+
+            if (enum_exists($nameOfType)) {
+                return TypeDefKind::SCALAR_KIND;
+            }
```
helderco, line 96: Enums have their own kind: `TypeDefKind::ENUM_KIND`.

sdk/php/template/README.md
```diff
@@ -0,0 +1,194 @@
...
+- grep-dir
+
+These functions are written within `src/Example.php`.  
+This aptly named file demonstrates how DaggerFunctions should be written.
+
+
+### Call Module Functions ###
+
+```text
+dagger call echo  --value="hello world" -mod=<path-to-your-module>
```
helderco, line 52: Nit: The `-m` flag is best at the beginning of the command. I usually put it right after `dagger`, but at the very least it should be on the `dagger <top-level>` command:
```suggestion
dagger call -m <path-to-your-module> echo  --value="hello world" 
```

sdk/php/template/dagger
```diff
@@ -0,0 +1,27 @@
...
+    // Useful when doing development on this package
+    require __DIR__.'/vendor/autoload.php';
+}
+
+$console = new Application();
+
+$console->add(new EntrypointCommand());
+$console->setDefaultCommand('dagger:entrypoint');
+
+define('DAG', \Dagger\Dagger::connect());
```
helderco, line 25: I'm not sure about this. Doesn't feel right as a constant. How about just `\Dagger\Dagger::dag()`? Unless you make it a function and it becomes `\Dagger\dag()`?

The point is that multiple calls to it should return the same instance (i.e., not establish a new connection).

sdk/php/tests/Unit/ValueObject/TypeTest.php
```diff
@@ -0,0 +1,272 @@
...
+        foreach (self::provideIdAbleClasses() as $idAble) {
+            yield $idAble => [
+                (new ReflectionClass($idAble))->getShortName(),
+                $idAble,
+            ];
+        }
+    }
+
+    /** @return class-string[] */
+    private static function provideIdAbleClasses(): array
```
helderco, line 235: Will this function need to be updated if new core types are added?

charjr: Not if new types are added. Only if existing implementation of `Dagger\Client\Idable` is changed to stop implementing `Dagger\Client\Idable`.

It is a Data Provider for the following test:
https://github.com/dagger/dagger/blob/b11ab3ca47a1e8a1ecf95218f6ec5a09b299468d/sdk/php/tests/Unit/ValueObject/TypeTest.php#L63-L68

As such, it does not need to be an exhaustive list of all implementations, it would probably be fine to only use 3-5 cases.

If we want a dynamically updating list, we could use this:
```php
foreach (get_declared_classes() as $className) {
    if (in_array(Idable::class, class_implements($className))) {
        yield => $className;
    }
}
```

However, looping through every declared class may be a bit excessive.

sdk/php/src/ValueObject/Type.php
charjr: Thank you, fixed.

sdk/php/template/README.md
charjr: That is clearer; I've adjusted it to match your suggestion.

sdk/php/tests/Unit/ValueObject/TypeTest.php
helderco: > As such, it does not need to be an exhaustive list of all implementations, it would probably be fine to only use 3-5 cases.

Got it, thanks 👍

sdk/php/src/ValueObject/Type.php
helderco: I meant that "union" types don't exist in Dagger 🙂

sdk/php/template/dagger
charjr: The function itself is not assigned to the constant, its return value is.

So `DAG` will always be the same instance.

We picked a constant because it is the closest match to how it is used in other SDKs which just call `dag` as a global variable.

sdk/php/src/ValueObject/Type.php
charjr: > Enums have their own kind: `TypeDefKind::ENUM_KIND`.

Ahah! Thank you, I hadn't rebased recently, so I didn't realise 0.12.0 added enum support. I'll get on that.

TomChv (commented):
Looks cool overall! Just left a comment related to enum

sdk/php/src/Command/EntrypointCommand.php
```diff
@@ -0,0 +1,229 @@
...
+    {
+        $typeDef = DAG->typeDef()->withOptional($type->nullable);
+
+        switch ($type->typeDefKind) {
+            case TypeDefKind::BOOLEAN_KIND:
+            case TypeDefKind::INTEGER_KIND:
+            case TypeDefKind::STRING_KIND:
+            case TypeDefKind::VOID_KIND:
+                return $typeDef->withKind($type->typeDefKind);
+            case TypeDefKind::SCALAR_KIND:
```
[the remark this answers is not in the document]
charjr: Yep, just solved enum. I was having a bit of trouble upgrading to 0.12.0.

Fortunately @jedevc and @helderco were able to guide me through it.

sdk/php/src/Command/EntrypointCommand.php
charjr: Both should now be supported within this PR, thank you for the assistance!

dragoonis:
> > `#[ObjectType]` is an option. If `Function` is reserved in PHP, it could be `#[Func]` like in TS.
> 
> @carnage and I just reviewed the naming conventions:
> 
> We changed `#[DaggerArgument]` to `#[Argument]`.
> 
> We have not changed`#[DaggerFunction]`; we discussed it and no other name matches as closely to Dagger's Domain Specific Language. We always refer to them as "Dagger Functions" whether it is in documentation or simply informal discussions.
> 
> We have not currently changed `#[DaggerObject]`; there is no "Dagger Object" in the DSL, so this one is more open to change. We currently left this alone because it ties in nicely with `#[DaggerFunction]`.

@helderco @charjr 

We can alias our import statements at the top, such as

``` php
use Dagger\Dagger as Dag
```

This way we can do

``` php
Dag::connect()
```

As well as

```
use Dagger\DaggerFunction as Func
```

This way we can do

```
#[Func]
```

Keeping it consistent with other SDKs is important.

sdk/php/dagger.json
charjr, line 1: @TomChv We have changed to using `include` with only a single `exclude` for vendor as @helderco pointed out that the Go SDK includes it by default. 

We've included all the files we believe are required for the SDK to work, however, I seem to be able to run my dagger modules completely fine with nothing but:
```json
    ...
    "include": [
        "dagger.json"
    ],
    ...
   
```
Which seems wrong to me, what should be breaking if this file isn't correct?

helderco: The `dagger.json` is already included by default by the engine. So that's the same as being empty there. Is that "include" for a user module or the runtime module?

charjr: This is the `sdk/php/dagger.json`.

You are correct, it seems "dagger.json" is included by default.

```
{
    "name": "php-sdk",
    "sdk": "go",
    "exclude": [
        "vendor/"
    ],
    "source": "runtime",
    "engineVersion": "v0.12.0"
}
```

This still works, as long as I don't explicitly exclude `dagger.json`.

I can call dagger functions from the `sdk/php` directory and they work fine.

I can also create a PHP module, using that version of the sdk, which should have nothing except `dagger.json` then call the following function through dagger:

```php
 #[DaggerFunction('Echo the value to standard output')]
     public function echo(string $value): Container
     {
         return dag()
             ->container()
             ->from('alpine:latest')
             ->withExec(['echo', $value]);
     }
```

My expectation would be, this would blow up in my face because without including `codegen/` how does dagger know what a PHP SDK's `Container` is?

charjr: We've set the `sdk/php/dagger.json` to this currently:

```json
{
    "name": "php-sdk",
    "sdk": "go",
    "include": [
        "codegen",
        "dagger.json",
        "install-composer.sh",
        "init-template.sh",
        "src/",
        "template/"
    ],
    "exclude": [
        "vendor/"
    ],
    "source": "runtime",
    "engineVersion": "v0.12.0"
}

```

This is everything we believe is required in order for it to work. The fact it works **without requiring them** is the confusing part.

helderco: > [...] The fact it works **without including them** is the confusing part.

Ok, I'll need to investigate.

sdk/php/src/Command/EntrypointCommand.php
charjr: We now have `sdk/php/src/functions.php` which lets us import the client like so:

```php
//...

use function Dagger\dag;

#[DaggerObject]
class Example
{
     #[DaggerFunction('Echo the value to standard output')]
     public function echo(string $value): Container
     {
         return dag()
             ->container()
             ->from('alpine:latest')
             ->withExec(['echo', $value]);
     }

// ...
```

sdk/php/src/Command/EntrypointCommand.php
```diff
@@ -0,0 +1,224 @@
...
+
+    private function getTypeDef(ListOfType|Type $type): TypeDef
+    {
+        $typeDef = DAG->typeDef()->withOptional($type->nullable);
+
+        switch ($type->typeDefKind) {
+            case TypeDefKind::BOOLEAN_KIND:
+            case TypeDefKind::INTEGER_KIND:
+            case TypeDefKind::STRING_KIND:
+            case TypeDefKind::SCALAR_KIND:
```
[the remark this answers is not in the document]
charjr: This is fixed in latest version

sdk/php/src/ValueObject/Type.php
charjr: Ah! I apologise, I misunderstood.

I assume Dagger does not support "union" or "intersection" types?

I have removed this TODO comment entirely.

sdk/php/src/ValueObject/Type.php
charjr: Enums are now supported

sdk/php/src/ValueObject/Type.php
helderco: Can you remove this comment to avoid confusion please?

sdk/php/template/dagger
charjr: We now have `sdk/php/src/functions.php` which lets us import the client like so:

```php
//...

use function Dagger\dag;

#[DaggerObject]
class Example
{
     #[DaggerFunction('Echo the value to standard output')]
     public function echo(string $value): Container
     {
         return dag()
             ->container()
             ->from('alpine:latest')
             ->withExec(['echo', $value]);
     }

// ...
```

sdk/php/composer.lock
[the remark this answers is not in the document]
charjr: The composer.lock has been added back in

sdk/php/src/ValueObject/Type.php
charjr: The comment is now removed. :+1:

sdk/php/src/Command/EntrypointCommand.php
helderco: I'll mark it as resolved since it appears there's already a new PR for this.

sdk/php/template/composer.json
carnage: We're not using json_validate as the implementation moved on from there. 8.2 should be fine; we can bump it in future when 8.2 is not under active support.

helderco (commented):
Submitting a first batch now while I take a deeper look at the include/exclude patterns. Not all of this is necessarily blocking, but best if you can resolve as much as possible so it's not forgotten in a follow-up.

Do what you can and we'll talk after.

sdk/php/template/composer.json
```diff
@@ -0,0 +1,28 @@
...
+    "dagger/dagger": "*@dev"
```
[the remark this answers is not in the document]
charjr: I believe we intended to swap that to a stable version once we'd got a working SDK but had become forgotten tech-debt.

Resolved and based on "^0.12.0", I will see about basing it off of "^0.12.2" after solving other issues.

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,171 @@
...
+	// Directory with the PHP SDK source code.
+	// +optional
+	sdkSourceDir *dagger.Directory,
+) *PhpSdk {
+	if sdkSourceDir == nil {
+		// TODO: change back to git or something remote when the runtime is ready.
+		// Go back to ../ to get the root of the PHP SDK
+		// This is a bit of a hack, you need to call it from a parent directory but it works.
+		// Example from ../../ (the directory above the dagger git repo)
+		// dagger init --sdk=dagger/sdk/php --source=test
```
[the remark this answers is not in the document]
charjr: Comment is now removed.

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,171 @@
...
+
+	/**
+	 * Mounts the PHP SDK code,
+	 * Installs composer into the container
+	 * Runs composer install in the codegen directory
+	 * Runs codegen using the schema json provided by the dagger engine
+	 */
+	ctr = ctr.
+		WithMountedDirectory("/codegen", sdk.SourceDir).
+		WithoutEntrypoint().
```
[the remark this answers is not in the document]
charjr: I have now removed the call to `WithoutEntryPoint()`

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,171 @@
...
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"php-sdk/internal/dagger"
+)
+
+const (
+	DefaultImage          = "php:8.3-cli-alpine"
```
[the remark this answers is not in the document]
charjr: Thank you very much for the tip as well! I appreciate learning something new :100: 

The digest has now been included in the name.

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,171 @@
...
+		WithoutEntrypoint().
+		WithWorkdir("/codegen").
+		WithExec([]string{"apk", "add", "git", "openssh", "curl"}).
+		WithExec([]string{
+			"git",
+			"config",
+			"--global",
+			"url.https://github.com/.insteadOf",
+			"git@github.com:",
+		}).
```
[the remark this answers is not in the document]
charjr: When we were first working on the SDK, we had issues accessing git repos over SSH and had to resort to HTTPS.

I forgot exactly what the issue was now, but I've tested the branch without it and it still works, so I assume the issue is no longer present.

This call to `WithExec` has now been completely removed.

sdk/php/template/composer.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "daggermodule/example",
+  "description": "An example module you can use as a starting point for building your own dagger modules",
```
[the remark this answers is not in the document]
helderco: Doesn't have to be the Module's name btw, it can be something general like "main", but it's been planned for a long time in Python to be able to use the Module's name, so in code you can do `import trivy` instead of `import main`, for example.

If not trivial or if changing the name here has consequences in code, you can choose something static for now like "main", just not "example".

sdk/php/template/dagger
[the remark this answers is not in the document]
charjr: Completely agree.

The `dagger` file has now been renamed `entrypoint.php`

sdk/php/template/composer.json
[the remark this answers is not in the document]
charjr: This has to be `*@dev` because it's installing the package from a directory inside the Container. Because it's installing it from a directory, it doesn't consider it a "tagged" version, so we can't use a tag.

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,171 @@
...
+		WithExec([]string{"./install-composer.sh"}).
```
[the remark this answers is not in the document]
charjr: I am now getting composer from a docker image:

https://github.com/dagger/dagger/blob/d879d2b459f6ae12e3744db3aaef80907137f750/sdk/php/runtime/main.go#L91-L95

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,171 @@
...
+		WithMountedDirectory(ModSourceDirPath, modSource.ContextDirectory()).
+		WithWorkdir(filepath.Join(ModSourceDirPath, subPath)).
+		WithDirectory(
+			GenDir,
+			ctr.Directory("/codegen"),
+			dagger.ContainerWithDirectoryOpts{Exclude: []string{
+				"codegen",
+				"runtime",
+				"vendor",
+				"install-composer.sh",
```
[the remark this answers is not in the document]
charjr: This has been solved in tandem with removing the script altogether in favour of getting it from a docker image:

https://github.com/dagger/dagger/pull/7493#discussion_r1693159463

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,171 @@
...
+		).
+		WithExec([]string{
+			"/codegen/init-template.sh",
+			filepath.Join(ModSourceDirPath, subPath),
+			name,
+		}).
+		WithFile(
+			"./install-composer.sh",
+			ctr.File("/codegen/install-composer.sh"),
+		)
```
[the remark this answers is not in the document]
charjr: I think this has also been solved in tandem with changing to a composer docker image: https://github.com/dagger/dagger/pull/7493#discussion_r1693159463

I've checked the 2 remaining calls, one is necessary when generating the module. The other is necessary when running the module. I may have the terminology slightly wrong, but point being, removing either call breaks _something_. So both calls remaining are required.

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	/**
+	 * Mounts the PHP SDK code,
+	 * Installs composer into the container
+	 * Runs composer install in the codegen directory
+	 * Runs codegen using the schema json provided by the dagger engine
+	 */
+	ctr = ctr.
+		WithMountedDirectory("/codegen", sdk.SourceDir).
```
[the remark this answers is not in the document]
charjr: We've now renamed codegen to `entrypoint.php`.

This ties the naming convention together:

`sdk/php/entrypoint.php` is the entry point when generating the module.
`<your-module>/dagger/entrypoint.php` is the entry point when running `<your-module>`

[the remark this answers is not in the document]
charjr: We've now renamed codegen to `entrypoint.php`.

This ties the naming convention together:

`sdk/php/entrypoint.php` is the entry point when a generating the module.
`<your-module>/dagger/entrypoint.php` is the entry point when running `<your-module>`

The "/codegen/" directory has kept its name currently, it only exists while generating the code, what ends up in `<your-module>` does end up in an "sdk/" directory.

sdk/php/runtime/main.go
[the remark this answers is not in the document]
helderco: Cool! Now it's better if pulling that image comes much earlier in the pipeline, before mounting the sources to avoid cache invalidation.

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,156 @@
...
+	// Directory with the PHP SDK source code.
+	// +optional
+	sdkSourceDir *dagger.Directory,
+) *PhpSdk {
+	if sdkSourceDir == nil {
+		sdkSourceDir = dag.
+			CurrentModule().
+			Source().
+			Directory("../").
+			WithoutDirectory("runtime")
```
helderco, line 37: Because of the difference between local and git remotes I think for now we need to duplicate the include/exclude patterns in a `WithDirectory` here. Hopefully when _context directories_ are available, the patterns will be able to be unified under a `// +ignore` pragma in this constructor's `sdkSourceDir` argument.

sdk/php/runtime/main.go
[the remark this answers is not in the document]
helderco: Why not name it `/sdk` though? That would be clearer.

Note that in the beginning that's how Python installed the SDK. It mounted under `/sdk`, and manually installed that directory before installing the module itself.

Now, I recommend adding the SDK to the `SourceDir` instead, codegen on top of that before mounting it, and install the sdk as a normal dependency of the module with a simple `composer install`, which should also work locally.

sdk/php/dagger.json
helderco: So, the reason for this is that _include/exclude_ patterns apply to local directories, but not git remote sources. When it's from a git remote, partial clones aren't supported yet, so the entire repo is cloned, subdir selected, but files aren't excluded because the purpose of that would be to avoid a long upload/pull but the entire repo has already been downloaded.

Continue in https://github.com/dagger/dagger/pull/7493#discussion_r1693687752.

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,171 @@
...
+		WithDirectory(
+			GenDir,
+			ctr.Directory("/codegen"),
+			dagger.ContainerWithDirectoryOpts{Exclude: []string{
+				"codegen",
+				"runtime",
+				"vendor",
+				"install-composer.sh",
+				"composer.phar",
+				"composer.lock",
```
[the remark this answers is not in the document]
helderco: It's ok to require the user to do a `dagger develop` first before running `composer install` locally. This happens for all SDKs.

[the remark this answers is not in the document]
helderco: > [...] which won't have a fixed reference in the lock.

Wait, what does this mean? How does it show up in the lock?

sdk/php/runtime/main.go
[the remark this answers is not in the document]
helderco: I'm not sure what 2 calls you're referring to.

sdk/php/runtime/main.go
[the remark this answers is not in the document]
charjr: We've noticed that @carnage and I had different references to dagger in the composer.lock.

We're not entirely sure how composer is forming the reference for dagger since it installs dagger from generated code from a file path instead of a git repository.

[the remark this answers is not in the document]
helderco: Doesn't matter if the path is generated. It should be the same relative path for both of you. Can you experiment and confirm this difference in the lock?

sdk/php/runtime/main.go
charjr: https://github.com/dagger/dagger/blob/26ca9200d15dea7c43632c1c8b5ceb3e070db1c4/sdk/php/runtime/main.go#L38-L50

sdk/php/template/composer.json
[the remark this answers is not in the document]
charjr: It now dynamically generates the package name and has a blank description:

dagger init --sdk=github.com/carnage/dagger/sdk/php@add-php-runtime --name="Potato Pie"

```json
{
  "name": "daggermodule/potato-pie",
  "description": "",
```

sdk/php/runtime/main.go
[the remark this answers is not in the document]
helderco: Sorry, today is a new day and looking at this comment the lock file in question looks like it's from the SDK. In that case it's correct to exclude. What shouldn't be excluded is when you create a new module, a new lock file should be created and added to the source dir for the module. I suppose `compose install` will do that already.

sdk/php/runtime/main.go
helderco: What's in `sdkSourceDir` is what needs to go in the generated `sdk/` directory, generated for the module.

This is what I expect:

```go
// TODO: Replace with a *default path from context* when 
// https://github.com/dagger/dagger/pull/7744 becomes availble.
 sdkSourceDir = dag. 
 	Directory(). 
 	// NB: these patterns should match those in `dagger.json`. When `--sdk` points to a git remote
 	// the files aren't filtered using `dagger.json` include/exclude patterns since the whole repo is cloned.
 	// It's still useful to have the same patterns in `dagger.json` though, to avoid the unnecessary uploads
 	// when loading the SDK from a local path.
 	WithDirectory("", sdkRoot, dagger.DirectoryWithDirectoryOpts{
 	 	Include: []string{
			"composer.json",
			"src",
			"LICENSE",
			"README.md",
		},
 	 	Exclude: []string{
			"vendor",
		},
 	)
```

As I explained in other comments, I think you should move some files into `runtime/`, which is why the following files aren't in the list above:

- `template/` → `runtime/template/` (it's only used by the runtime to bootstrap a new module)
- `entrypoint.php`  → `runtime/scripts/entrypoint.php` (it's only as the runtime's entrypoint command)
- `init-template.sh` → `runtime/scripts/init-template.sh` (it's only by the runtime)
- ~`dagger.json`~  (not necesary as part of the SDK sources)
- ~`generated/`~ (codegen should be able to run with this being empty so it generates it from a clean slate)

For any of these, instead of getting them through `sdkSourceDir`, you should get them through the module's source directory. Remember that in `dagger.json` there's `"source": "runtime"`:
```go
tmplDir := dag.CurrentModule().Source().Directory("template")
entrypointFile := dag.CurrentModule().Source().File("scripts/entrypoint.php")
initTemplate := dag.CurrentModule().Source().File("scripts/init-template.sh")
```

sdk/php/runtime/main.go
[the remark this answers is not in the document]
helderco: I've placed the `SourceDir` under `/sdk` and the "entrypoint" is now under `/sdk/scripts/codegen.php`. I think this makes it clearer.

sdk/php/runtime/main.go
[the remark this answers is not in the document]
helderco: I moved composer installation to the _base_ part of the pipeline as suggested.

sdk/php/runtime/main.go
```diff
@@ -0,0 +1,171 @@
...
+	 * Copies the generated code and rest of the sdk into the module directory under the sdk path
+	 * Runs the init template script for initialising a new module (this is a no-op if a composer.json already exists)
+	 */
+	ctr = ctr.
+		WithMountedDirectory(ModSourceDirPath, modSource.ContextDirectory()).
+		WithWorkdir(filepath.Join(ModSourceDirPath, subPath)).
+		WithDirectory(
+			GenDir,
+			ctr.Directory("/codegen"),
+			dagger.ContainerWithDirectoryOpts{Exclude: []string{
```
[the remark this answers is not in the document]
helderco: I've addressed this in my changes.

sdk/php/runtime/main.go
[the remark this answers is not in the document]
helderco: I've addressed this in my changes.

sdk/php/runtime/main.go
helderco: I've addressed this in my changes.

helderco (approved):
lgtm! 🎉

I made some changes to the runtime module to address feedback more quickly and get this merged.

## Follow-ups
- It would be great to decouple codegen from the SDK so you don't need to install the the whole SDK just for that. I did something similar in Python.
- Rather then an `entrypoint.php` with custom requires for autoload, isn't there an easier way to support scripts via `composer.json`?
- PHP tests still aren't integrated in our CI. Now that you have modules support I suggest you **dogfood** it and create a `sdk/php/dev` module (like in Python) to replace the docker composer environment for testing, with Dagger. Then we can simply hook that easily in our CI, to call the PHP function to run the tests (and linter).

helderco:
I didn't notice the failing DCO before but this PR is long running now so I've set the DCO to pass because there's multiple contributors we'd need to go after and change the whole history of commits. @carnage is now aware of the need to sign-off. 👍

PR #7494, opened 2024-05-29
chore(sdk/elixir): improve elixir codegen part 2

wingyplus:
In this PR, I do:

* Move code template from EEx to iodata. I tried the EEx way for a while but it's quite hard to modify the template when the logic get complicated. The iodata is quite a bit easier to read and easier to change the logic.
* Add more unit tests, I use a library called `mneme` to do snapshot testing the code instead of adding assertion by hand. The tests aren't cover all the case but the important cases (object renderer) are most cover. We can add it more later. 
* Reduce dependencies in `mix.lock` file, including `nestru`.
* Run generate and format code concurrently.


sipsma (approved):
Not an expert on elixir by any means but LGTM from what I can follow

PR #7420, opened 2024-05-20
core: multi SCM support

grouville:
First working draft of multi SCM support.

Currently only works on public VCS repo, as well as public vanity URL relying on the `go-get` vanity URL logic from Go.

**TODO**
- [x] More detailed history of https://github.com/grouville/go-vcs when merging
- [x] Improve error messages: on init, install, call, functions -> add ++++ integration tests
- [x] Add vanity URL test by changing `dagger.io/dagger` redirect (side PR) -> https://github.com/dagger/dagger.io/pull/3730
- [ ] Add `vcsToBuildKitRef` logic in core/schema/git` to handle same ref format for dirs (follow-up PR), and remove test from this one
- [ ] Extend `moduleSource` type to handle `SSH_AUTH_SOCK` (ongoing) for private module handling (potentially follow-up PR)

PR #7496, opened 2024-05-29
chore: import `go-vcs` library with clean history

grouville:
### Context
First PR of a serie, splitting the implementation of https://github.com/dagger/dagger/pull/7420, in a cleaner way.

This one imports the codebase from https://github.com/grouville/go-vcs directly into our codebase for ease/maintainability purposes. I took the liberty of making it part of the `engine/ci` subdir, but open to discussion

#### First commit
The first commit is Google's library https://pkg.go.dev/golang.org/x/tools/go/vcs#RepoRootForImportPath filtered-history squashed into one, with all the context:
```shell
This commit adds a forked / improved version of Google's RepoRootForImportPath logic. Most of this code is originally part of the go/internal section of Google's codebase: https://github.com/golang/go/blob/16d3040a84be821d801b75bd1a3d8ab4cc89ee36/src/cmd/go/internal/vcs/vcs.go#L1281.

However, Justin found a version of this library exposed by Google without all the internal dependencies. It is now deprecated as unmaintained, but the core logic: the dynamic resolving coupled with the regexes are still relevant and enough for our use-case: https://pkg.go.dev/golang.org/x/tools/go/vcs#RepoRootForImportPath.

We had to fork the library in order to upgrade the go version used, bump the dependencies and more importantly fix a few unported changes:
1. We remove unsupported VCS, as BuildKit only support Git based VCS
2. We remove the deprecated mercurial check from bitbucket vcs options (https://github.com/golang/go/commit/5b1b80b) that led to an unability to parse the root of a repo from a bitbucket private ref
3. Update the regexes and the tests for jazz, apache, git.sr.ht SCM (following current official package's logic)
4. Remove the deprecation notices, as this package does not aim to be 100% compliant with upstream but just handle the split between the root of a repo and the subpath from its URL + rely on io.ReadAll instead of ioutil
```

#### Follow-up commits
All the follow-up commits are mostly maintenance work implemented to sync the deprecated library in a simi-official state with Google's internal version: https://github.com/golang/go/blob/master/src/cmd/go/internal/vcs/vcs.go
- [chore: apply latest SCM regexes + remove mercurial from bitbucket vcs options](https://github.com/grouville/go-vcs/commit/d414b863a7b89c8a66d059fa690e210712740138) mainly

Then, it is followed by a few Dagger specific changes:
- [remove unsupported VCS as BuildKit only supports Git](https://github.com/grouville/go-vcs/commit/2e8fc1bb70efdc4400afe4e3b963a94c5daf2485)
- [chore: comply with golintci convention](https://github.com/grouville/go-vcs/commit/be2c60c91275f4762da41b58f34f865670b1a5bd)
- [extend regex to allow subpath traversal](https://github.com/grouville/go-vcs/commit/b60debcb91ca5d2eaaf75565001456704f68d764)


marcosnils:
> `engine/ci` subdir, but open to discussion

how about using `vcs` as the package name which is the same the original library is using?

grouville:
> how about using `engine/vcs` as the package name which is the same the original library is using?

Given that the `labels.go` is refactoring all the vcs terminology to ci, I don't have a strong opinion around that: https://discord.com/channels/707636530424053791/1233176406112993350/1243237579676061767

marcosnils:
> Given that the `labels.go` is refactoring all the vcs terminology to ci, I don't have a strong opinion around that: https://discord.com/channels/707636530424053791/1233176406112993350/1243237579676061767

but `labels.go` is for telemetry labels, right? In this case we're talking about an engine package to handle `vcs` information for different URL resolvers. I think having an `engine/vcs` package actually makes a bit more sense. 

Actually, the package is already called `vcs` (https://github.com/dagger/dagger/pull/7496/files#diff-7e6077f012d1b44204a4d56bf23fc79ac711db5517d15e7c87fd65ebb095cd49R5) but it's in a `ci` directory which is not generally very go idiomatic.

vito:
Agreed - vcs makes sense here. I wouldn't try to line this up with the labeling terminology; this part is more low level and has separate concerns. Nothing is set in stone for the new label plans anyhow

grouville:
> Actually, the package is already called `vcs` (https://github.com/dagger/dagger/pull/7496/files#diff-7e6077f012d1b44204a4d56bf23fc79ac711db5517d15e7c87fd65ebb095cd49R5) but it's in a `ci` directory which is not generally very go idiomatic.

You're right, I made a mistake and pushed a wrong version at first. I changed and moved to `engine/vcs` as per your recommendations 🙏

marcosnils (approved):

PR #7497, opened 2024-05-29
ci: build dumb-init for target platform and add verification

sipsma:
dumb-init was accidentally always being built for the host platform, which meant if it was built on an x86 host the published arm64 image would incorrectly be x86.

This fixes that and also adds general purpose verification of all the binaries we build checking that they are of the expected target platform, to reduce the chances of this happening again.

---

Tested the verification logic by locally going back to building the dumb-init binary for the host platform and getting an error when I built for x86 on my arm host.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,7 +18,6 @@ linters:
     - misspell
     - nakedret
     - nolintlint
-    - prealloc
     - revive
     - rowserrcheck
     - staticcheck
--- a/ci/build/builder.go
+++ b/ci/build/builder.go
@@ -191,15 +191,43 @@ func (build *Builder) Engine(ctx context.Context) (*dagger.Container, error) {
 		return nil, fmt.Errorf("unsupported engine base %q", build.base)
 	}
 
-	ctr := base.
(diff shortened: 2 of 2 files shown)


sipsma:
@jedevc @vito after hitting the "consider pre-allocating" this slice of 4 elements linter complaint again (think of the nanoseconds you could be saving!!), so  I just went ahead and disabled it. It's by far the most common lint I hit and has been useful 0% of the time. If you have any objections let me know and I'll be less salty and just pre-allocate my 4 elements 😂

helderco:
> @jedevc @vito after hitting the "consider pre-allocating" this slice of 4 elements linter complaint again (think of the nanoseconds you could be saving!!), I just went ahead and disabled it. It's by far the most common lint I hit and has been useful 0% of the time. If you have any objections let me know and I'll be less salty and just pre-allocate my 4 elements 😂

I felt the same this week when making changes to GetFunctions in the CLI. 😅

vito (approved):
Nice - also, I'm more than fine with disabling that linter. That refactor has led to brittle code in some places, so I won't miss it.

Issue #7401, opened 2024-05-17
✨ [Typescript] Support string literal types


Xiot:
### What are you trying to do?

I'm trying to use a literal type in order to have better typing in my dagger code.

I have a field like this on an object.
```typescript
@field() type: 'head' | 'merge'
```

However it looks like the code-gen doesn't seem to support this as when I try to run the dagger module, I get
```
 Error: query module objects: returned error 400 Bad Request: failed to get schema introspection JSON: introspection query failed: input: __schema.types[75].fields[2].type.ofType panic while resolving __Type.ofType: unknown type: headmerge
```

### Why is this important to you?

Typing is important

### How are you currently working around this?

Currently, I am just using a string for the `field` and am casting it later.

Xiot:
Related to this, I just got this since my module was returning a fixed set of static string.  
You most likely just need to change the representation to a string if you detect a string literal while traversing the AST

```
query module objects: returned error 400 Bad Request: failed to get schema for module "motion-dagger": failed to create function "on": failed to find mod type for function "on" return type: "TaskNotNeededOnDraftTaskNotNeededOnChoreNoChangeNeededOk!"
```

TomChv:
@Xiot Thanks for reporting this issue, we could solve this issue by allow user to implement their own Scalar or enum.

We made a [PR](https://github.com/dagger/dagger/pull/7158) few weeks ago to add support for Scalar, however it's only internal values for now.

Here's also a related issue https://github.com/dagger/dagger/issues/6780

@helderco Should I start to work on an implementation to support custom scalar registration?

Xiot:
This is great
From a Typescript perspective I prefer unions of string literal over enums. Its generally closer to what I tend to use, as enums in typescript are weird

TomChv:
> From a Typescript perspective I prefer unions of string literal over enums. Its generally closer to what I tend to use, as enums in typescript are weird

I totally agree, and right now we're closer than even to support custom unions (called `Scalar` in our case)

helderco:
> @helderco Should I start to work on an implementation to support custom scalar registration?

Please don't. What OP is asking is clearly an enumeration, not a custom scalar:
- https://github.com/dagger/dagger/issues/6780

That is useful 👍 and I also want it. 🙂

> Yes! And the idea is to have a better readability by supporting Scalar or Enum, I'm not sure we want to support inline union type since it's specific to TypeScript and not Go nor Python can do that kind of things.

[...]

PR #7498, opened 2024-05-29
feat: add custom enum support in Engine API & TypeScript SDK

### LINKED ISSUE #6780: ✨ functions: argument values

TomChv:
This PR implements custom enum support from a user module and the TypeScript support for it.

### Current state

- [x] `Enum` TypeDef GraphQL schema extension
- [x] `Enum` registration using the GQL API
- [x] Internal `EnumTypeDef` support
- [x] Add `ENUM_KIND` to `TypeDef`
- [x] Complete TypeScript SDK registration system
- [x] Handle enum field query
- [x] Handle enum return type
- [x] Handle enum argument 

💡 There's maybe something wrong in my implementation but it works!

### Changes

To make that possible, I needed to add a `DynamicEnumValue` to register dynamic enum in the GraphQL server since the current implementation is static (because of usage of generic)

Also extends the **GraphQL** schema based on @helderco comments

```graphql
type EnumTypeDef {
  name: String!
  description: String!
  values: [EnumValueTypeDef!]!
}

type EnumValueTypeDef {
  name: String!
  description: String!
}

type TypeDef {
  """
  If kind is ENUM_KIND, the enum-specific type definition.
  If kind is not ENUM_KIND, this will be null.
  """
  asEnum: EnumTypeDef
    
  """
  Returns a TypeDef of kind Enum with the provided name.
  
  Note that an enum's values may be omitted if the intent is only to refer to an enum.
  This is how functions are able to return their own, or any other circular reference.
  """
  withEnum(name: String!, description: String = ""): TypeDef!

  """
  Adds a static value for an Enum TypeDef, failing if the type is not an enum.
  """
  withValue(name: String!, description: String = ""): TypeDef!
}

enum TypeDefKind {
  """
  A GraphQL enum type and its values.
  
  Always paired with an EnumTypeDef.
  """
  ENUM_KIND
}
```

**Note that only `string` value is supported for now**

### Example for TypeScript

```typescript
import { func, object, field, daggerEnum } from "@dagger.io/dagger"

/**
 * Enum for Status
 */
@daggerEnum()
class Status {
  /**
   * Active status
   */
  static readonly ACTIVE: string = "ACTIVE"

  /**
   * Inactive status
   */
  static readonly INACTIVE: string = "INACTIVE"
}

@object()
export class Enums {
  @field()
  status: Status = Status.ACTIVE

  @func()
  setStatus(status: Status): Enums {
    this.status = status

    return this
  }

  @func()
  getStatus(): Status {
    return this.status
  }
}
```

💡 Due to TypeScript flexibility, it's pretty tricky to implement it in another way, so I tried to create a syntax close to what we could do in Golang.
The fact that we cannot declare decorators on top of `enum` is also a limitation. 
I'm thinking about a workaround but I think this type of implementation is the cleanest we could get in Typescript.

Examples of commands:

```shell
➜  enums git:(main) ✗ dagger call status
ACTIVE

➜  enums git:(main) ✗ dagger call set-status --status FOO status
✔ enums: Enums! 2.4s
✘ Enums.setStatus(status: "FOO"): Enums! 0.0s
! failed to convert arg "status": ModuleEnumType.ConvertToSDKInput: invalid enum value "FOO" for "Status"

Error: response from query: input: enums.setStatus resolve: failed to convert arg "status": ModuleEnumType.ConvertToSDKInput: invalid enum value "FOO" for "Status"

➜  enums git:(main) ✗ dagger call set-status --status INACTIVE get-status
INACTIVE
```

I think the input part should be handle in the CLI, with possible values displayed, so we can catch the error sooner (/cc @helderco)

### Additional changes

**This PR also include some improvement on the TypeScript SDK.**

- Deletion of method `typedef` in abstract module definition in favor of directly calling the method.
   This avoid duplicated function call and simplify the overall code, there's no more 2 concepts of typedef inside the TS 
   internals.
- Load codes representation inside the constructor so results is in memory and not recomputed on every call. This will improve the performances as well as simplify the code.
- Move `typeToTypedef` into `abstraction` directory since it was the only place it was used. 
- Move `toPascalCase` into  `module` since it was only used there.
- Delete `utils.ts` (I've never been a fan of this file tbh).
- Avoid decorators naming duplication by reusing the decorator name inside registry, this avoid typos.
- Refactor `module.modTypeFor` to be more readable and flexible.

Fixes #6780 
Fixes OSS-141


TomChv:
It seems I'm not able to regenerate the Rust client, other SDKs are well generated.

I'm starting to work on the TS registration

sdk/typescript/entrypoint/load.ts
```diff
@@ -162,6 +163,20 @@ export async function loadValue(
...
+      const enumType = (type as TypeDef<TypeDefKind.EnumKind>).name
+
+      // Workaround to call to get the enum that has an id
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      if (dag[`load${enumType}FromID`]) {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        return dag[`load${enumType}FromID`](value)
+      }
```
TomChv, line 176: Since Enum are not loadable, I should remove this part, and potentielly just return the value, since it should have been validated by the engine first!

TomChv:
I need to write some tests but the PRs is ready for a review

core/schema/coremod.go
```diff
@@ -350,8 +417,11 @@ func introspectionRefToTypeDef(introspectionType *introspection.TypeRef, nonNull
 	case introspection.TypeKindEnum:
 		return &core.TypeDef{
 			// just call it a string for now
```
jedevc, line 419: this comment isn't true anymore

core/schema/module.go
```diff
@@ -280,6 +283,18 @@ func (s *moduleSchema) Install() {
...
+			Doc(`Returns a TypeDef of kind Enum with the provided name.`,
+				`Note that an enum's values may be omitted if the intent is only to refer to an enum.
+				This is how functions are able to return their own, or any other circular reference.`).
+			ArgDoc("name", `The name of the enum`).
+			ArgDoc("description", `A doc string for the enum, if any`),
+
+		dagql.Func("withEnumValue", s.typeDefWithEnumValue).
+			Doc(`Adds a static value for an Enum TypeDef, failing if the type is not an enum.`).
+			ArgDoc("value", `The name of the value in the enum`).
+			ArgDoc("description", `A doc string for the value, if any`),
```
jedevc, line 297: Hm, why do we have a `withEnumValue`? Couldn't `withEnum` instead accept an array? At point of instantiation we should know all the possible values right?

Not a deal breaker at all, I think I have a preference for having just one method.

core/schema/module.go
```diff
@@ -303,20 +320,23 @@ func (s *moduleSchema) typeDef(ctx context.Context, _ *core.Query, args struct{}
 
 func (s *moduleSchema) typeDefWithOptional(ctx context.Context, def *core.TypeDef, args struct {
 	Optional bool
-}) (*core.TypeDef, error) {
+},
+) (*core.TypeDef, error) {
```
jedevc, line 324: nit: this formatting shouldn't change - there's a few instances of this in this file.

core/typedef.go
```diff
@@ -369,6 +375,8 @@ func (typeDef *TypeDef) ToInput() dagql.Input {
 		typed = DynamicID{typeName: typeDef.AsInterface.Value.Name}
 	case TypeDefKindVoid:
 		typed = Void{}
+	case TypeDefKindEnum:
+		typed = dagql.NewScalar[dagql.String](typeDef.AsEnum.Value.Name, dagql.String(""))
```
jedevc, line 379: Hm, why isn't this an `EnumObject` or similar? I think this might accidentally mean that we lose input validation (but I may be wrong).

When you write the tests, make sure to include tests with invalid enum values (see how we do that for scalars already). If they fail, it's likely this line at fault.

sdk/java/dagger-codegen-maven-plugin/src/main/resources/schemas/schema-dev-1720da5a9ffe2d9ca485e0d005265f25d974fda1.json
jedevc, line 1: We don't generate the Java SDK automatically.

core/schema/module.go
vito: Each value also has a description, so I can see the reasoning here. To pass them all in as an array you'd need to pass a structure that has both, right?

jedevc: Mm, you're right. Yup, I guess in that case, we should keep this as is.

TomChv: I just followed the API @helderco designed, I'm not sure which design is the best though

core/typedef.go
TomChv: I still got the input validation but it's done by the `ConvertToSDKInput` & `ConvertFromSDKResult` in `enum`

To use `EnumObject`, I must implement the `Input` interface but I already did it for `DynamicEnumValue` so I don't know.
I've been quite confused on this part honestly, about the  `Input` interface..

sdk/java/dagger-codegen-maven-plugin/src/main/resources/schemas/schema-dev-1720da5a9ffe2d9ca485e0d005265f25d974fda1.json
TomChv: Oups, is it bad that I updated it?

core/schema/coremod.go
TomChv: Done

core/schema/module.go
TomChv: normally fixed

core/typedef.go
jedevc: Okay, so here's what I would try (may not actually be like this, but this would hugely simplify the structure).
- Remove the `DynamicEnumValue` interface - we should have a counterpart for the `EnumValues` struct called `DynamicEnumValues`. Both of these should implement `EnumValue` (the interface).
- Collapse all the functionality of `EnumObject` into `DynamicEnumValues` - I don't think we need both structs.

But I think this line is probably correct as-is - I think this is fine, but I still think we should try the above. I think that's what's causing a lot of the confusion.

core/integration/module_test.go
```diff
@@ -3104,6 +3104,73 @@ class Test {
...
+	for _, tc := range []testCase{
+		{
+			sdk: "typescript",
+			source: `import { func, object, field, daggerEnum } from "@dagger.io/dagger"
+
+/**
+ * Enum for Status
+ */
+@daggerEnum()
+class Status {
```
jedevc, line 3123: Question: why this syntax instead of typescript's builtin enum type? https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

I think we should stick to the language semantics if possible.

jedevc: Do we need a decorator here? Can we not take the same approach as in go, where we create typedefs for every referenced type - e.g. so if a function accepts/returns an enum, then we automatically codegen for it.

Maybe that's less inline with the rest of the typescript SDK, and makes for more complications.

TomChv: I mean I wish it worked like that, but I need to refactor a lot of the Typescript internal introspection logic to get that working.

Should I work on that?

core/typedef.go
TomChv: Okay I'm going to try that! Thanks for the hint

TomChv: Done! Thanks a lot

TomChv:
My test passes, the other fails du to flakiness or is it something else?

Screenshot 2024-06-10 at 22 28 18

The fail come from there, it might be from my implementation?

Screenshot 2024-06-10 at 22 29 18

TomChv:
Quick update! I finally found a way to fix the issue with interfaces but it's pretty hacky so I'm trying to find a better way.

It's basically an issue in `module.go/ModTypeFor` function, that doesn't correctly handle the interfaces kind for some reason.
This bug has been introduced after my refactor of this function.

A quick fix is to rollback the change for `KindInterface` but it's quite ugly...
I want to make it work, I'm digging into a better fix

TomChv:
Enum tests is passing! @helderco @jedevc You can work on the other languages integrations

dagql/types.go
```diff
@@ -864,6 +865,132 @@ type EnumValue interface {
...
+func (e *DynamicEnumValues) ToLiteral() call.Literal {
+	return call.NewLiteralEnum("")
+}
+
+func (e *DynamicEnumValues) Lookup(val string) (Input, error) {
+	// Don't verify the enum if values are not loaded yet, we cannot compare it to nil since
+	// we call `make` in the constructor.
+	// Instead we log a warning for debug purposes
+	if len(e.values) == 0 {
+		slog.Warn("DynamicEnumValues.Lookup: values not loaded yet inside enum", "val", val)
```
jedevc, line 976: Does this warning occur during normal execution? If so, we shouldn't have it, warnings in the logs should be useful to the user IMO.

docs/docs-graphql/schema.graphqls
```diff
@@ -1157,6 +1157,48 @@ The `DirectoryID` scalar type represents an identifier for an object of type Dir
...
+  id: EnumValueTypeDefID!
+
+  """The name of the enum value."""
+  name: String!
+}
+
+"""
+The `EnumValueTypeDefID` scalar type represents an identifier for an object of type EnumValueTypeDef.
+"""
+scalar EnumValueTypeDefID
```
jedevc, line 1200: Hm something here doesn't look right. `EnumValues` shouldn't have IDs should they?

Should this even be exposed over the API?

dagql/types.go
```diff
@@ -864,6 +865,132 @@ type EnumValue interface {
...
+func NewDynamicEnum(t Enum) *DynamicEnumValues {
+	enumValues := t.ListValues()
+
+	values := make([]string, len(enumValues))
+	for _, v := range enumValues {
+		values = append(values, v.Name)
+	}
+
+	return &DynamicEnumValues{t: t, values: values}
+}
```
jedevc, line 920: Something about this still doesn't smell quite right - we shouldn't be taking an `ast` value as input in any way here - `dagql` should inform what `ast` is output, not used to inform what is input.

I think the `ListValues` in the `Enum` interface is kinda the source of this. Why do we need to have the `Enum` interface at all?

IMO, the way we should have this is that `NewDynamicEnum` should take a list of `[]DynamicEnumValue` - so it's essentially the "at-runtime" version of `NewEnum` that doesn't use generics.

dagql/types.go
```diff
@@ -864,6 +865,132 @@ type EnumValue interface {
...
+	// Instead we log a warning for debug purposes
+	if len(e.values) == 0 {
+		slog.Warn("DynamicEnumValues.Lookup: values not loaded yet inside enum", "val", val)
+		return NewDynamicEnumValue(e, val), nil
+	}
+
+	for _, possible := range e.values {
+		if val == possible {
+			// Convert it to a string so it can be serialized in JSON
+			return NewString(val), nil
```
jedevc, line 983: Shouldn't this be `NewDynamicEnumValue`? If it can't be serialized to JSON, then you can implement `MarshalJSON` on it to get it to do that. 

Otherwise, we're just stripping type data away.

jedevc (commented):
Left some comments on the engine implementation, will also start looking at how to do enums in go.

dagql/types.go
TomChv: Yeah it does, it's when the argument is loaded but the enums values are not because it requires the actual typescript , you remember the issue we got our pair programming?

docs/docs-graphql/schema.graphqls
TomChv: I don't know, I just followed @helderco API

dagql/types.go
TomChv: I agree with you, however this would require to loop on all possible values to construct them, which make it quite ugly in the code

dagql/types.go
TomChv: Oh yeah in that case I need to implement the interface

dagql/types.go
jedevc: It'll surely just add a few lines no? Where specifically will it make it ugly?

For example, in `core/typedef.go` it's still very simple - we don't give any possible values, since those aren't known at that point.

Also, in general, go prefers a more explicit style - it's okay to be a bit long-winded, especially in a case like this, where the "neater" approach is more difficult to understand and has unclear dependencies.

core/schema/coremod.go
```diff
@@ -349,9 +417,11 @@ func introspectionRefToTypeDef(introspectionType *introspection.TypeRef, nonNull
 
 	case introspection.TypeKindEnum:
 		return &core.TypeDef{
-			// just call it a string for now
-			Kind:     core.TypeDefKindString,
+			Kind:     core.TypeDefKindEnum,
 			Optional: !nonNull,
+			AsEnum: dagql.NonNull(&core.EnumTypeDef{
+				Name: introspectionType.Name,
+			}),
```
jedevc, line 424: Hm, wait, we're deliberately leaving `Values` empty here?

docs/docs-graphql/schema.graphqls
jedevc: `EnumValueTypeDefID` isn't in @helderco's suggested API (at least the one in the top-level of your PR)

core/schema/coremod.go
TomChv: Oh that's a good catch

dagql/types.go
TomChv: > Also, in general, go prefers a more explicit style - it's okay to be a bit long-winded, especially in a case like this, where the "neater" approach is more difficult to understand and has unclear dependencies.

Alright, I'll do the changes

core/integration/module_call_test.go
```diff
@@ -627,13 +627,71 @@ func (m *Test) ToProto(proto string) NetworkProtocol {
...
+
+func (m *Test) ToStatus(status string) Status {
+	return Status(status)
+}
+`,
+			})
+
+		out, err := modGen.With(daggerCall("from-status", "--status", "ACTIVE")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "ACTIVE", out)
```
jedevc, line 679: Note to self: this test is currently failing - that's because the typedef presented to the CLI is `Active` (fine, very good). But inside the engine, there's no conversion back to the `OriginalName` `ACTIVE` (not fine, bad).

dagql/types.go
jedevc: Removed this, and essentially moved `DynamicEnum` into `core` - it's not a base "dagql" type like everything else, it's specifically much more like an object, which is defined in `core` specifically for modules to expose.

dagql/types.go
jedevc: Resolved this, there's a lot less type dancing now.

core/integration/module_call_test.go
jedevc: Fixed this, by refactoring to avoid type dancing, a lot of this magic thankfully goes away.

helderco:
I see that this:

```ts
/**
 * Enum for Status
 */
@daggerEnum()
class Status {
  /**
   * Active status
   */
  static readonly ACTIVE: string = "ACTIVE"

  /**
   * Inactive status
   */
  static readonly INACTIVE: string = "INACTIVE"
}
```

Is producing this:

```graphql
"""Enum for Status"""
enum TestStatus {
  """Active status"""
  Active

  """Inactive status"""
  Inactive
}
```

Values should be `ACTIVE` and `INACTIVE` instead.

core/integration/module_test.go
helderco: I think not being able to use a decorator in an enum shouldn't be an argument against using `enum` if that's the most natural thing. That said, I remember a user saying they didn't like TypeScript's enums, but I don't know why.

helderco: Also, part of me thinks `@function` is the only one we could need. Both for class methods and properties. In Python however, `@object` does change the class, unlike TypeScript. If not that, users would need `@dataclass` anyway.

jedevc:
Ah will look at the go docs extraction, something's not right there.

> Values should be `ACTIVE` and `INACTIVE` instead.

Should we not be doing value normalization? We track `OriginalName`, so we can convert it to the expected value for the module, but IMO the user should see it normalized in the same we normalize other value types.

helderco:
> > Values should be `ACTIVE` and `INACTIVE` instead.
> 
> Should we not be doing value normalization? We track `OriginalName`, so we can convert it to the expected value for the module, but IMO the user should see it normalized in the same we normalize type names.

Well, GraphQL enums don't have both name and value, just **value**. And that's what's after the `=` in the examples above. I don't think you can normalize that. 

Anyway, some users can get confused if they choose names that don't match the values, in the same language:

```go
type Status string

const (
	Foo Status = "ACTIVE"
	Bar Status = "INACTIVE"
)
```

If you install this module as a dependency this should produce:

```go
type TestStatus string

func (TestStatus) IsEnum() {}

const (
	Active Status = "ACTIVE"
	Inactive Status = "INACTIVE"
)
```

I can see them raising an issue that `Status.Foo` from the module becomes `TestStatus.Active` in another.

GraphQL's convention for enum values is UPPER_SNAKE. Go uses that as the value and normalizes the name based on the value. When registering a custom enum with `withEnumValue(val)`, the name needs to be ignored and only use the value.

But we may want to raise a warning or something when the names don't normalize to the same thing in the same language. Basically in Go, see if the name for the value `ACTIVE` corresponds to the normalized `Active` that codegen will produce. While in Python, just check if they're the same.

TomChv:
In TypeScript, I actually use the `Name` of the property as value because as you said, there's no such mapping in GraphQL so I choose to use the name because it's easier to parse and I expect people to use the same value and name.

> But we may want to raise a warning or something when the names don't normalize to the same thing in the same language. Basically in Go, see if the name for the value ACTIVE corresponds to the normalized Active that codegen will produce. While in Python, just check if they're the same.

Yeah I think the name and the value should be the same and we should emit a warning or an error

helderco:
> In TypeScript, I actually use the `Name` of the property as value because as you said, there's no such mapping in GraphQL so I choose to use the name because it's easier to parse and I expect people to use the same value and name.

To be technically correct, you should use the values. In GraphQL, even though there's no "mapping", it's an enumeration of values. It's the languages that have a corresponding names in addittion to the values, but they're mostly variable names to hold the value in a type safe way. It's the value that should be serialized and deserialized.

While it's convention in Python to use `Status.PENDING_RECEIPT`, it seems in TS/JS, the convention is `Status.PendingReceipt`, like in Go. That's what the TS codegen generates actually:

https://github.com/dagger/dagger/blob/84b170c9410e169c2ae71bad7fd8946dff83c702/sdk/typescript/api/client.gen.ts#L1030-L1034

I expect that custom user enums would match as closely as possible to what the codegen generates. Even though the client binding is an actual enum, at least the usage would be consistent if you fix the naming convention for the properties.

sdk/typescript/introspector/decorators/decorators.ts
```diff
@@ -6,3 +6,4 @@ import { registry } from "../registry/registry.js"
 export const object = registry.object
 export const func = registry.func
 export const field = registry.field
+export const daggerEnum = registry.daggerEnum
```
helderco, line 9: I don't really like repeating the package name in a var like this. For example (hypothetically), if the naming clash with `enum` was just shadowing a built-in instead of a syntax error, you could resolve the conflict with:

```diff
- import { enum } from '@dagger.io/dagger';
+ import { enum as enumType } from '@dagger.io/dagger';
```

I've resolved the conflict in Python with `@object_type` and `@enum_type` because there's a built-in `object` and `enum` in the std lib. Adding "type" matches GraphQL since they both create different types. Would be the same if we decided to support custom _input types_ (there's also a built-in `input` in Python).

So how about `enumType` instead?

TomChv: I like that, I can make the switch

core/typedef.go
```diff
@@ -472,6 +480,23 @@ func (typeDef *TypeDef) WithObjectConstructor(fn *Function) (*TypeDef, error) {
...
+	return typeDef
+}
+
+func (typeDef *TypeDef) WithEnumValue(name, desc string) (*TypeDef, error) {
+	if !typeDef.AsEnum.Valid {
+		return nil, fmt.Errorf("cannot add value to non-enum type: %s", typeDef.Kind)
+	}
+
+	typeDef = typeDef.Clone()
+	typeDef.AsEnum.Value.Values = append(typeDef.AsEnum.Value.Values, NewEnumValueTypeDef(name, desc))
```
helderco, line 495: Need to error if it's a duplicate. Otherwise it overwrites a previous one, which can be confusing. Doesn't have to be here, just that it's validated by the API at some point.

helderco, line 495: Need to error if it's a duplicate. Otherwise it overwrites a previous one, which can be confusing.

core/typedef.go
```diff
@@ -836,6 +863,101 @@ func (typeDef *InputTypeDef) ToInputObjectSpec() dagql.InputObjectSpec {
...
+
+func (*EnumValueTypeDef) TypeDescription() string {
+	return "A definition of a value in a custom enum defined in a Module."
+}
+
+func NewEnumValueTypeDef(name, description string) *EnumValueTypeDef {
+	return &EnumValueTypeDef{
+		Name:         strcase.ToCamel(name),
+		OriginalName: name,
+		Description:  description,
```
helderco, line 951: An enum value has just an optional description attached, so there's no "name" in a GraphQL "enum value" here.

If normalizing, the convention would be to use `SCREAMING_SNAKE_CASE` instead of `PascalCase`[^1], but it would be confusing when you're using a module that defines values with a different convention. I think it's best to just use the original value as is, but I don't have a clear example why atm though.

[^1]: The strcase lib incorrectly uses the "camel" vs "lower camel" designation btw, instead of the more common "pascal" vs "camel".

helderco (commented):
Mostly comments around the use of "name", but othwerwise needs support in Go for value descriptions and to test for those description. Might as well test that the schema is created with proper naming convention for the values.

jedevc:
Pushed a fix for the go enum value doc strings, and added tests for them (that apply to all SDKs).

TomChv:
I also implemented @helderco comments in my last commit

core/integration/module_test.go
Xiot: > I think we should stick to the language semantics if possible.  

I strongly agree with this.  That being said, enums in typescript can act fairly weirdly.  The general way, that at least I use, is to just use a type literal as its more natural in Typescript.  But that ends up with the same issue of not being able to add a decorator.

>  That said, I remember a user saying they didn't like TypeScript's enums, but I don't know why.

I left it on a comment in a PR, I'll try to find it.

> Also with Helder we aggreed that explicit declaration is better than implicit, I could add a special case for enum but I don't know, I'm not sure.

The explicit declaration is a little cumbersome in Typescript, as it generally adds a bunch of boiler plate, especially for things like this enum

Xiot: Here was my original comment on enums.  
https://github.com/dagger/dagger/issues/7401#issuecomment-2125999323

core/typedef.go
helderco: I'm relaxing the use of "name" in `EnumValueTypeDef` after taking another look at the spec. 

To be clear, this is different from the use of "name" in the SDKs. In GraphQL, this is **not the name of the value**, it's that **the value is a GraphQL "Name"**, which means that it needs to follow the usual rules for variable names: only letters, digits and underscores, and has to start with a letter or a single underscore.

Screenshot 2024-06-20 at 13 22 00

Screenshot 2024-06-20 at 13 22 13

Screenshot 2024-06-20 at 13 23 38

helderco: I'm relaxing the use of "name" in `EnumValueTypeDef` after taking another look at the spec. 

To be clear, this is different from the use of "name" in the SDKs. In GraphQL, this is **not the name of the value**, it's that **the value is a GraphQL "Name"**, which means that it needs to follow the usual rules for variable names: only letters, digits and underscores, and has to start with a letter or a single underscore.

Screenshot 2024-06-20 at 13 22 00

Screenshot 2024-06-20 at 13 22 13

Screenshot 2024-06-20 at 13 23 38

helderco: So we need to validate if the values are valid GraphQL names. I just tested with core enums, and you're able to register invalid values, but it'll produce an invalid schema. I think it's ok to only validate in custom enums though since we are in control of core, and it's probably easier to give a better error message if it's in the type definition.

sdk/typescript/introspector/scanner/abtractions/enumValue.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  toJSON() {
+    return {
+      name: this.value,
+      description: this.description,
+    }
+  }
+
+  // LoadValue load the value of the enum value from the property initializer.
```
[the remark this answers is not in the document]
TomChv: Yes it's hard to lose that habit, sorry haha

core/typedef.go
```diff
@@ -491,6 +491,13 @@ func (typeDef *TypeDef) WithEnumValue(name, desc string) (*TypeDef, error) {
 		return nil, fmt.Errorf("cannot add value to non-enum type: %s", typeDef.Kind)
 	}
 
+	// Verify if the enum value is duplicated.
+	for _, v := range typeDef.AsEnum.Value.Values{
+		if v.OriginalName == name {
+			return nil, fmt.Errorf("enum value %s is already defined", name)
+		}
+	}
```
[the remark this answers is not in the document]
TomChv: It seems you have a custom error handling in Python for enum, which is not returning that error

```
 duplicate values found in <enum 'Status'>: DUPLICATE -> ACTIVE
```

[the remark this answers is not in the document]
TomChv: I added the test in [d49c885](https://github.com/dagger/dagger/pull/7498/commits/d49c88556f98294ecfc0d8942bab8552e87be161)

sdk/typescript/introspector/scanner/abtractions/enumValue.ts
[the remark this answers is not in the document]
TomChv: Fixed in [d49c885](https://github.com/dagger/dagger/pull/7498/commits/d49c88556f98294ecfc0d8942bab8552e87be161)

sdk/typescript/introspector/scanner/abtractions/enumValue.ts
```diff
@@ -0,0 +1,66 @@
...
+  // If the initializer is not set, it will throw an error.
+  // We do not use the actual property name because it may varies between SDKs languages conventions.
```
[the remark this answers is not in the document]
TomChv: Fixed in [d49c885](https://github.com/dagger/dagger/pull/7498/commits/d49c88556f98294ecfc0d8942bab8552e87be161)

helderco:
Need to validate the values, see https://github.com/dagger/dagger/pull/7498#discussion_r1647751607.

core/integration/module_test.go
helderco: As I said, not being able to use a decorator shouldn't be an argument against using something else that is more natural. The problem with a type of string literals really is how do you add descriptions to the values?

helderco (approved):
lgtm :shipit: just needs a rebase and green checks

TomChv:
Hitting an issue with Python, nothing related to the feature, can I merge the PR @helderco ?

https://dagger.cloud/dagger/traces/319f0dabaaadec36dd30a3ff00570aed?span=23de8a3cbca388c2

core/schema/coremod.go
TomChv: Ok so I looked into this one and we're doing exactly the same for Object, we don't need to load the values there since they are loaded from the dagql server itself.

TomChv: Ok so I looked into this one and we're doing exactly the same for Object, we don't need to load the values there since they are loaded from the dagql server itself.

gerhard:
Checks need to be green before we merge. Do we know what is failing? Is it failing locally?

TomChv:
> Checks need to be green before we merge. Do we know what is failing? Is it failing locally?

Everything alright locally on the enum tests, the fails is also on main with:  https://discord.com/channels/707636530424053791/1253675236843454556/1253675239188070430

Issue #7131, opened 2024-04-18
CLI: calling 'export' just prints 'true'


shykes:
## Problem

When calling `Directory.export()` or `File.export()` from the CLI, the output is confusing: it just prints `true` without context or explanation.

Example:

```console
dagger call \
 -m github.com/shykes/core@770692357aeac6bf898f28177614763ae5cc620c \
 directory \
 export --path=./out
```

Output:

```console
true
```

## Solution

Change the CLI output to something less confusing. I propose *no output*, which is the expected convention for a command that completes successfully.

kpenfound:
Alternatively, we could have it show the same output as the `--output <path>` option:
```
 ~  dagger -m github.com/shykes/daggerverse/wolfi@v0.1.3 call container --output ./wolfi.tar                                                     Fri 19 Apr 2024 11:20:34 AM EDT
Saved output to "/home/kpenfound/wolfi.tar".
```

helderco:
`dagger call` consistently prints the results of functions. When you use `--output` it actually uses `export` underneath, and reports the absolute path back to you. 

Remember this?
- https://github.com/dagger/dagger/issues/6321

There's also an issue called "Update Bool -> Void" (DEV-2774) since it's no longer necessary that some functions like `Export` return a boolean. You can just return `Void` and check for error. In this case though, if `export: Void!`, rather than seeing `true` you'll end up seeing nothing actually.

But https://github.com/dagger/dagger/issues/6357 was supposed to replace using export in the CLI, and I hoped the difference in output would help nudge people towards `-o` as the goto for saving any output from functions to the host. However we're showing more of `export` in the docs rather than `-o`. 

If we make it so `dagger call func export --path PATH` in the CLI does the same as `dagger call func -o PATH`, it's going to be even more ambivalent when to use one vs the other. My preference is to adjust our docs and help support to favor `-o` because of consistency.

I do understand how using `export` can also be more recognizeable since it's in the core API and users have been using it before modules. We may also end up deciding to do something about these side-effect functions, like https://github.com/dagger/dagger/issues/6321. In the meantime, if there is consensus to make `export` the same as `-o`, there's a bit of pre-requisite work to support it which is the same for another issue related to detecting "sync".

shykes:
> If we make it so dagger call func export --path PATH in the CLI does the same as dagger call func -o PATH, it's going to be even more ambivalent when to use one vs the other. My preference is to adjust our docs and help support to favor -o because of consistency.

[...]

PR #7500, opened 2024-05-29
💥 api: return absolute path on export instead of boolean

helderco:
Fixes https://github.com/dagger/dagger/issues/7131

It's much more useful for `export` to return the absolute path on the host where the asset (Container, Directory, File) was exported to, rather than a boolean. The boolean wasn't useful at all because it's never false. It's either true or an error.

It's possibly a breaking change in Go if any user is assigning the return value (boolean), which is now a different type. However, since that boolean value is useless this is considered very low impact. Affected users (if any), aren't probably aware that it's never `false`.

### DIFF
--- a/.changes/unreleased/Breaking-20240529-181906.yaml
+++ b/.changes/unreleased/Breaking-20240529-181906.yaml
@@ -1,0 +1,6 @@
+kind: Breaking
+body: 'api: return absolute path on export instead of boolean'
+time: 2024-05-29T18:19:06.709015Z
+custom:
+  Author: helderco
+  PR: "7500"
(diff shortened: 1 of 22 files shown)


TomChv (approved):
LGTM

shykes:
@helderco is this mergeable?

jpadams:
Just rebased to keep it fresh. Looks like this is for the v0.12.0 release because of the breaking change.

helderco:
> Just rebased to keep it fresh. We're holding off merge until ready for v0.12.0 release because of the breaking change possibility for Go SDK users.

Just to be clear, any Go user affected by this is most likely relying on code that doesn't work because the boolean is useless (never false). I expect most users are using `_, err := ctr.Export(ctxt)` correctly already (not breaking), and even if someone's affected, it'll at least shed some light that they're probably relying on code that doesn't work (like checking for condition of boolean being false).

shykes (approved):

PR #7501, opened 2024-05-29
executor: always include search setting in resolv.conf

sipsma:
Draft pending:
- [x] test, need to fully grok the situations in which this bug was occurring since it seems to be host dependent
   - Added now, FTR the host dependency was whether or not the host's /etc/resolv.conf had a `search` setting or not. That ends up propagated to the engine container and to execs, but this bug was triggered when there was no `search` setting in the base resolv.conf, so it ended up being dependent on the resolv.conf of the host.

### DIFF
--- a/core/integration/services_test.go
+++ b/core/integration/services_test.go
@@ -1676,6 +1676,76 @@ func TestServiceContainerToHost(t *testing.T) {
 	})
 }
 
+func TestServiceSearchDomainAlwaysSet(t *testing.T) {
+	// verify that even if the engine doesn't have any search domains to propagate to execs, we still
+	// set search domains in those execs
+
+	// NOT parallel since this requires setting env vars unfortunately
+	// t.Parallel()
+
+	// FIXME: This test seems particularly prone to telemetry flushing hanging, so use
+	// a timeout for now until that is fixed.
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	t.Cleanup(cancel)
+
+	c, err := dagger.Connect(ctx, dagger.WithLogOutput(newTWriter(t)))
+	require.NoError(t, err)
+	t.Cleanup(func() { c.Close() })
+
+	resolvContents, err := c.Container().From("alpine").
+		WithExec([]string{"cat", "/etc/resolv.conf"}).
+		Stdout(ctx)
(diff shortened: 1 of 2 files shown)


core/integration/services_test.go
```diff
@@ -1676,6 +1676,68 @@ func TestServiceContainerToHost(t *testing.T) {
...
+			Backend:  1234,
+			Frontend: 32132,
+		}},
+	}).Start(ctx)
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		hostSvc.Stop(ctx, dagger.ServiceStopOpts{Kill: true})
+	})
+
+	t.Setenv("_EXPERIMENTAL_DAGGER_RUNNER_HOST", "tcp://127.0.0.1:32132")
```
vito, line 1723: Non-blocking obviously but would be nice find another way to do this; I've been hacking on an instrumentable test suite system for better OTel integration, and I made it `t.Parallel()` by default. Could go back on that, but in general we pay a high cost for non-parallelizable tests since they add a constant overhead to the test run-time. (Preaching to the choir I'm sure, the fix is most important for now, and it's awesome that we have a test for it.)

vito (approved):

core/integration/services_test.go
sipsma: Yeah, I've personally written at least 3 different variations on this general theme of "integ test that requires a custom nested engine". This was a new one this time 😅 It fortunately only took 4.5 seconds to run on my host, which is why I went with it despite the lack of parallel.

The downside of the other approaches is the complexity+overhead of connecting back a new client to the nested engine (re-exec tests as a container in new engine, etc.), which didn't feel worth it in this case since the test is otherwise so simple. Will go with this for now but it should be adjustable to one of the different approaches if needed.

In general, we probably should figure out some more clever+consistent way of approaching all these sort of tests (or maybe adjust the Go SDK to support this without the need to set env vars?).

sipsma: Also FYI @vito, I had to rebase and when re-running the tests locally I hit a hang on telemetry flushing a few times (but inconsistent still), so I updated the test with an explicit timeout context (went with a highly generous 60s), but something we can undo once your fixes for all that are in place.

PR #7502, opened 2024-05-29
docs: Add service optimization and export recipes to cookbook

vikram-dagger:


docs/current_docs/manuals/developer/cookbook.mdx
```diff
@@ -668,3 +746,37 @@ Set environment variables in a container:
...
+
+</TabItem>
+</Tabs>
+
+#### Example
+
+Create a Redis service with a cache volume to persist its data:
+
+```shell
+dagger call redis-service
```
vikram-dagger, line 781: @vito @kpenfound I'm not sure if I'm testing this wrong. I ran it once, then commented the `set/save` commands in the function and ran it again, expecting to see the output `abc` since the value should be stored in the cache from the first run. However no output was returned on the second run. Am I missing something or is this code broken?

marcosnils: @vikram-dagger seems like the issue is coming from here: https://github.com/dagger/dagger/blob/main/docs/current_docs/manuals/developer/snippets/services/persist-service-state/go/main.go#L15. If you remove the `WithWorkdir` step, that works

docs/current_docs/manuals/developer/cookbook.mdx
```diff
@@ -589,18 +589,96 @@ The following Dagger Function tags a container image multiple times and publishe
...
+
+- Export the file returned by the Dagger Function to the `/home/admin/myfile` path on the host:
+
+    ```shell
+    dagger call \
+      get-file \
+      export --path=/home/admin/myfile
+    ```
+
+### Export a container image to the host
```
marcosnils, line 647: In this example it might be useful to show how you can also `docker load` it and run it in your docker engine

docs/current_docs/manuals/developer/cookbook.mdx
vikram-dagger: @marcosnils I tried it after removing that line but it still doesn't work when I comment out the `save` on the second run. The value `abc` doesn't appear in the output.

mircubed:
Next steps as of 5/31 - @vito please review this issue and let us know if you need any additional support https://github.com/dagger/dagger/pull/7502#discussion_r1622210718

docs/current_docs/manuals/developer/cookbook.mdx
vikram-dagger: Added

mircubed:
As of 6/3, @marcosnils will take the next steps to review this PR.

docs/current_docs/manuals/developer/cookbook.mdx
kpenfound: This line probably doesn't need to block this PR anyway since the related snippet isn't part of this PR unless I'm misunderstanding @vikram-dagger @marcosnils

kpenfound: Reviewing the snippet that Marcos linked, it looks like that isn't intended to show caching with consecutive executions, it shows it all in one run. The `return redisCLI. {get foo}` shows the cache volume since redisCLI wasn't reassigned to itself in the `set` above it

marcosnils:
I'm taking this one.

marcosnils:
@vikram-dagger  latest changes LGTM! only thing missing here seems to be the `go` example for the `persist-services`  use-case. LMK if I can help with anything!

marcosnils (approved):

Issue #7505, opened 2024-05-29
🐞 dagger CLI: no such file or directory


pjmagee:
### What is the issue?

I've been trying to follow the documentation on passing a file via the CLI.
It appears that the file is uploaded, but somehow there is an issue with the file being detected.

Using windows and following the default tutorial for host file access, it does not work on Windows

The weird part with -d is that this file seems to be "uploaded to the engine" fine..

```
    ✔ upload . from PATRICK-PC (client id: jhyf9iajxjh4gcitjjjlelp1i) (include: C:\Projects\example\README.md) 0.0s
    ✔ load cache: copy upload . from PATRICK-PC (client id: jhyf9iajxjh4gcitjjjlelp1i) (include: C:\Projects\example\README.md) 0.0s
```

![image](https://uploads.linear.app/0b2a9228-3827-45ff-bb20-ccd34f57f694/070fdecd-8979-4c96-89f0-502520700c5b/47f7a4dc-7b86-45c5-b7e1-80bedb7d8b9f?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzBiMmE5MjI4LTM4MjctNDVmZi1iYjIwLWNjZDM0ZjU3ZjY5NC8wNzBmZGVjZC04OTc5LTRjOTYtODlmMC01MDI1MjA3MDBjNWIvNDdmN2E0ZGMtN2I4Ni00NWM1LWI3ZTEtODBiZWRiN2Q4YjlmIiwiaWF0IjoxNzE3MDI3NjE3LCJleHAiOjMzMjg3NTg3NjE3fQ.t5VARnTkRfzgWsLJIOI8SYDiWXZOXlZd18SQO3O5EZI)

[...]

PR #7506, opened 2024-05-29
cli: convert Windows paths to / slash format

sipsma:
Before this, file+dir arg values to dagger call on a Windows client were not converted from `\` format to `/` format, which resulted in calls like `filepath.Split` in the Linux engine container to not work as expected.

Now the CLI calls `filepath.ToSlash` before sending paths, which *helps* handle the paths better in the engine container; specifically enough to avoid at least one bug that resulted in the engine trying to find a path at e.g. `/C:\Foo\Bar`.

To be clear, this is probably not the end of our need for improved Windows path handling, but may be enough to at least squash this one bug.

---

Fixes https://github.com/dagger/dagger/issues/7505

TODO:
- [x] Test on Windows manually to verify (mostly theoretical at this point still)
   * Done, could successfully call both of these now
      * `dagger.exe -m github.com/shykes/daggerverse/wolfi@v0.1.2 call container with-file --path=/README.md --source=C:\\Users\me\repo\README.md with-exec --args="cat","/README.md" stdout`
      * `dagger.exe -m github.com/shykes/daggerverse/wolfi@v0.1.2 call container with-file --path=/README.md --source=.\README.md with-exec --args="cat","/README.md" stdout`

### DIFF
--- a/cmd/dagger/flags.go
+++ b/cmd/dagger/flags.go
@@ -244,6 +244,7 @@ func (v *directoryValue) Get(ctx context.Context, dag *dagger.Client, modSrc *da
 	// POSIX "portable filename character set":
 	// https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282
 	path, viewName, _ := strings.Cut(path, ":")
+	path = filepath.ToSlash(path) // make windows paths usable in the Linux engine container
 	return modSrc.ResolveDirectoryFromCaller(path, dagger.ModuleSourceResolveDirectoryFromCallerOpts{
 		ViewName: viewName,
 	}).Sync(ctx)
@@ -319,6 +320,7 @@ func (v *fileValue) Get(_ context.Context, dag *dagger.Client, _ *dagger.ModuleS
 			return nil, fmt.Errorf("failed to resolve absolute path: %w", err)
 		}
 	}
+	vStr = filepath.ToSlash(vStr) // make windows paths usable in the Linux engine container
 	return dag.Host().File(vStr), nil
 }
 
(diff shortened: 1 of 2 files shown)


vito (approved):
I guess this is effectively saying our API layer expects `/` everywhere?

I'm fine with that but don't feel too strongly. Could see the argument for the API normalizing it, but the downside is `/` vs `\` would bust the query cache while remaining semantically equivalent. (But that's also true of e.g. including a trailing slash.)

sipsma:
> I guess this is effectively saying our API layer expects `/` everywhere?
> 
> I'm fine with that but don't feel too strongly. Could see the argument for the API normalizing it, but the downside is `/` vs `\` would bust the query cache while remaining semantically equivalent. (But that's also true of e.g. including a trailing slash.)

@vito Yeah, actually the main reason I didn't do it API-side is that Go's stdlib makes it kind of hard... Admittedly our use case of "Linux binary that needs to handle both Linux and Windows paths" is not *too* common, but this is still mildly frustrating at times.

Namely, if you call `filepath.ToSlash("C:\Foo\Bar")` on Linux, it just returns `C:\Foo\Bar`: https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/path/filepath/path.go;l=218

But if you call it on Windows it does what you actually want (`C:/Foo/Bar`), so it's easier at the moment to just do "pre-processing" client-side.

For this exact issue in particular, `filepath.ToSlash` is simple enough we could just implement it ourselves probably, but in other places it's more important we do the pre-processing client-side so just trying to stay consistent-ish at the moment.

My dream is that A) there would be a `path`-like library that lets you choose at runtime between implementations for different platform paths (excluding those that require syscalls of course) and B) we wouldn't have to write it ourselves 😅 (some of the implementations like `IsAbs` are a lot more complicated and probably annoying to maintain). But if it comes down to it we may just need to do that ourselves; my searching for something like this has been unfruitful so far.

PR #7511, opened 2024-05-30
feat(core): allow other CIs than GitHub

grouville:
Branched from https://github.com/dagger/dagger/pull/7496

This PR allows the use of other CIs than GitHub as a ref

In order to do it without changing the current logic, we do 3 things:
- Stat if the dir exists locally
- then extend parseRefString to actually try to isolate the root of the remote
- then fallback as a local source if no root of the remote is found

At the same time, we introduce a function to split the root and the subdir. 
All this logic relies on Google's `repoRootForImportPath` function that returns a repo Root from a ref, either from regexes (https://github.com/dagger/dagger/pull/7496), or from querying the remote with `go-get=1`

A follow-up PR will be done tomorrow to unify the ref format to the dirs, and another one (a bit later) will handle private repo refs

This PR is currently reviewable / testable.


core/schema/modulesource.go
```diff
@@ -145,25 +163,130 @@ func (s *moduleSchema) moduleSource(ctx context.Context, query *core.Query, args
...
 	modVersion string
 	hasVersion bool
-	isGitHub   bool
 	kind       core.ModuleSourceKind
+	repoRoot   *vcs.RepoRoot
 }
 
-func parseRefString(refString string) parsedRefString {
+// interface used for host interaction mocking
+type BuildkitClient interface {
```
marcosnils, line 190: is it possible to make this interface private? specially because we probably don't want to expose it to other packages.

core/schema/modulesource.go
```diff
@@ -145,25 +163,130 @@ func (s *moduleSchema) moduleSource(ctx context.Context, query *core.Query, args
...
+func parseRefString(ctx context.Context, bk BuildkitClient, refString string) (parsedRefString, error) {
 	var parsed parsedRefString
 	parsed.modPath, parsed.modVersion, parsed.hasVersion = strings.Cut(refString, "@")
-	parsed.isGitHub = strings.HasPrefix(parsed.modPath, "github.com/")
 
-	if !parsed.hasVersion && !parsed.isGitHub {
-		parsed.kind = core.ModuleSourceKindLocal
-		return parsed
+	// We do a stat in case the mod path github.com/username is a local directory
+	stat, err := bk.StatCallerHostPath(ctx, parsed.modPath, false)
```
marcosnils, line 204: we're not checking for this `err !=` nil anywhere?

core/schema/modulesource.go
```diff
@@ -27,10 +31,10 @@ type moduleSourceArgs struct {
 }
 
 func (s *moduleSchema) moduleSource(ctx context.Context, query *core.Query, args moduleSourceArgs) (*core.ModuleSource, error) {
-	parsed := parseRefString(args.RefString)
-	modPath, modVersion, hasVersion, isGitHub := parsed.modPath, parsed.modVersion, parsed.hasVersion, parsed.isGitHub
+	parsed, err := parseRefString(ctx, query.Buildkit, args.RefString)
```
marcosnils, line 34: we're not checking of error after this? seems like we need to revisit the error handling of this `parseRefString` function as it seems like we might want to a specific error in some cases?

core/schema/modulesource.go
grouville: We are, it is being returned in https://github.com/dagger/dagger/blob/a257857d89659ee6f1fd86c6252021d4361f19ab/core/schema/modulesource.go#L221, in the edge case where we fallback to a local source kind.

This is currently not being exploited to its fullest potential I agree, but it is being checked here: https://github.com/dagger/dagger/blob/a257857d89659ee6f1fd86c6252021d4361f19ab/core/schema/modulesource.go#L52, currently only as a log to specify the user that a fallback just happened. It could be refined a bit more afterwards to not log every case ?

core/schema/modulesource.go
```diff
@@ -145,25 +163,130 @@ func (s *moduleSchema) moduleSource(ctx context.Context, query *core.Query, args
 	return src, nil
 }
 
+// shared logic with originToPath, copy-pasted
+// to avoid dependency on Buildkit
```
jedevc, line 167: Not sure I understand this comment - this function uses `gitutil` which already has a dep on buildkit.

I'm happy with the impl though.

core/schema/modulesource.go
```diff
@@ -145,25 +163,130 @@ func (s *moduleSchema) moduleSource(ctx context.Context, query *core.Query, args
...
+	// we try to isolate the root of the git repo
+	repoRoot, err := vcs.RepoRootForImportPath(parsed.modPath, false)
+	if err == nil && repoRoot != nil && repoRoot.VCS != nil && repoRoot.VCS.Name == "Git" {
+		parsed.kind = core.ModuleSourceKindGit
+		parsed.repoRoot = repoRoot
+		return parsed, nil
+	}
+
+	parsed.kind = core.ModuleSourceKindLocal
+	return parsed, err
```
jedevc, line 221: Hm, is this an error or not?

I don't think we should have this half-errored state, either this should be a "invalid module" error, or we should silently fallback to a local source.

core/schema/modulesource.go
```diff
@@ -41,27 +45,41 @@ func (s *moduleSchema) moduleSource(ctx context.Context, query *core.Query, args
...
-		src.AsGitSource.Value.URLParent = segments[0] + "/" + segments[1] + "/" + segments[2]
+		// resolve the extracted repo root to a git URL
+		computedRepoRoot, err := resolveGitURL(parsed.repoRoot.Repo)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve git ref %s: %w", parsed.repoRoot.Repo, err)
+		}
+
+		// splits the root and subdir from the repo root path and the ref without version path
+		root, subdir := splitRootAndSubdir(refWithoutVersion, computedRepoRoot.Path, userRefWithoutVersion.Path)
+		root = computedRepoRoot.Host + "/" + root
```
jedevc, line 80: This logic seems *very* complicated, it's split up into a bunch of functions, I'm struggling to understand what's actually going on.

Why isn't this as simple as getting `computedRepoRoot.Host + "/" + computedRepoRoot.Path`? Why do we need to reference `refWithoutVersion` again?

Also, isn't `parsed.repoRoot.Root` equivalent to `subdir` at this point? If it's not, what's the difference, and why isn't this logic part of the `parseRefString` logic earlier?

core/integration/ref_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"dagger.io/dagger"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRefIntegration(t *testing.T) {
```
jedevc, line 11: We should also have some more complete integration tests - see https://github.com/dagger/dagger/blob/9d8f5cb804f8d647191b3f803f846a23a1b336ad/core/integration/module_config_test.go#L1169-L1185

We can mirror https://github.com/dagger/dagger-test-modules to some places in gitlab, and parameterize the tests to run against that repo hosted in different places (like a subgroup in gitlab). That way, we're testing against these other VCSs in the same way that we're testing against github, so they're both first-class citizens.

core/schema/modulesource.go
grouville: Done 🙏

core/schema/modulesource.go
grouville: I changed the logic to log directly inside parseRefString, to be able to debug when wrongful fallback happen

core/schema/modulesource.go
grouville: Reverted this change, as said above 🙏

core/schema/modulesource.go
grouville: Corrected the logic, with the function now logging when these is an error with vcs.RepoRootForImportPath, to be able to hint the user / debug why tge fallback to the local kind is wrong in this case

core/schema/modulesource.go
grouville: I refactored the code to move it to its own package (to be shared with cmd/flags package to unify the directory logic)

core/integration/ref_test.go
grouville: I discussed it with @marcosnils and I am not sure that integration tests are necessary here. Most of the logic here is to basically split the ref into root of repo and subdir. With the tests on vcs. repoRootForImportPath and the new `vcs.ExtractRootAndSubdirFromRef`

grouville:
> do we want to exclusively rely on go-get=1? Could we potentially attach two query parameters, e.g. ?go-get=1&dagger-module=1. And also support dagger-specific meta tag names? 

That's an amazing idea and a great follow-up for sure cc @marcosnils

core/integration/ref_test.go
jedevc: Will re-review in a bit - but. I don't necessarily think we *need* to test this for an initial merge, but we absolutely need some sanity integration tests at *some* point soon.

Previously, changes to some logic in git resolution/etc changes this in *incredibly* subtle and unexpected ways. If we aren't testing this, this becomes much more likely. This *has* happened before, multiple times - the module source logic gets touched a lot, and from experience it seems to be very fragile.

grouville: Ok, this makes sense 👌. I was planning on adding integration tests for the private git support, it seems a great complementary add at this time, wdyt?

jedevc: SGTM :rocket:

core/schema/modulesource.go
jedevc: Still seems to be public.

core/schema/modulesource.go
```diff
@@ -149,20 +148,47 @@ type parsedRefString struct {
...
+		parsed.kind = core.ModuleSourceKindGit
+		parsed.repoRoot = repoRoot
 		return parsed
 	}
-	parsed.kind = core.ModuleSourceKindGit
+
+	// log warning to hint that the remote ref fallbacked as a local source kind
+	if err != nil {
+		slog.Warn("ref %s has not been parsed as a git remote: %v", refString, err)
+	}
```
jedevc, line 189: Hm, so I think we'll only hit this when later this will fail right? We're requesting a module source that can't be resolved to a git repo, and wasn't a local directory.

So this will fail later, I think it's fine to silently fail here IMO.

core/schema/modulesource.go
marcosnils: yep, still public :D

marcosnils (approved):

marcosnils (changes requested):
approved accidentally. Most of the PR is there, only one nit is following regarding the interface visiblity.

jedevc:
Shared this with @grouville in DMs: https://github.com/dagger/dagger/commit/741b2b23d3f356604ea2b890a8915530a614891f

I worked out how we might be able to hugely simplify some of this code, by using some more of the things that the VCS lib returns.

core/schema/modulesource.go
grouville: my bad, shall be done this time 😇

core/schema/modulesource.go
grouville: Removed the slog accordingly, as it will fail later you are right

jedevc:
Going to pull this one in now!

PR #7517, opened 2024-05-30
improve releasing during v0.11.6

sipsma:
TODO:
- [ ] get rid of references to `./hack/make` 
- [ ] clarify `# Update .github/workflows dagger versions $ENGINE_VERSION` and `# Update docs/current_docs files to point to new dagger version` steps
   - I had to do some careful grepping around (can't prefix with `v` to find the docs file, yamls use v0-11-5, etc.)
- [ ] add note that homebrew-core repo may not get an automated update PR for a few hours (based on timestamps from previous release)

### DIFF
--- a/.github/workflows/_dagger_call.yml
+++ b/.github/workflows/_dagger_call.yml
@@ -24,7 +24,7 @@ on:
 
       size:
         type: string
-        default: "dagger-v0-11-5-4c-nvme"
+        default: "dagger-v0-11-6-4c-nvme"
         required: false
 
       dagger-version:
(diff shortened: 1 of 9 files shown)


jedevc:
@sipsma thanks for doing the release :tada:

I noticed a couple missed versions in `.github` so I bumped those too.

I think I did your TODOs too:

- I removed `./hack/make` references in the `RELEASING.md` file.
- I also clarified the section you mentioned about updating versions in `.github` and `docs` with the exact files/formats to search for. Ideally, we need a more automated way to do this, but out-of-scope for here I think.
- Added a note about homebrew-core.

jedevc (approved):

jedevc:
Oh *no* - we got `inconsistent graph state`: https://github.com/dagger/dagger/actions/runs/9365341205/job/25780355587?pr=7517.

I thought this was fixed by the upstream buildkit work :cry:

sipsma:
> Oh _no_ - we got `inconsistent graph state`: https://github.com/dagger/dagger/actions/runs/9365341205/job/25780355587?pr=7517.
> 
> I thought this was fixed by the upstream buildkit work 😢

 😢 indeed, though I technically only ever repro'd and verified a fix for the other error (`failed to get state for edge`), the connection to `inconsistent graph state` was purely theoretical. So I'll allow myself the cope that at least the other error seems to be squashed (was way more common than `inconsistent graph state`)...

PR #7518, opened 2024-05-30
fix(ts-sdk) set corepack inside of SDK install instead of global

TomChv:
#7096 introduced a regression by using `yarn install --production`. If the field `packagaManager` is set, it might lead to failure because yarn post v1.22.11 doesn't support the `--production` flag anymore, it has been removed in favor of `yarn workspaces focus --production`.

However, depending on the yarn major version (v2, v3, v4), the options are different but also the results and the way they handle subdependencies.

### Current state (that work)

~~It seems `npm` is the only package manager that actually force install all the sub dependencies, which ensure the TS SDK runtime properly works.~~

By initialiazing corepack inside the SDK installation, we ignore the field `packageManager` when installing dependencies with yarn, which will be compatible with any package manager version.
It's a nice way to keep yarn because it's faster than npm but be generic!

**Also based ont the discussion below, I removed the `--production` flag since TS module are not transpiled to javascript, and dependencies are isolated into their own sub `package.json`, it is less likely that unusefull dependencies will be installed from `devDependencies`, and if that's the case, it should impact performances that much**

### Things I tried that didn't worked

**Adapt yarn install command based on the yarn version**
- For some reasons, every major version has a special way to handle production dependencies.
  * v1: `yarn install --production`
  * v2: `yarn install`, there's no concept of `--production` (https://stackoverflow.com/questions/70992443/yarn-2-0-zero-installs-production-vs-development-deployment)
  * v3: `yarn workspaces focus --production` but it requires to enable the worspace plugin (https://stackoverflow.com/questions/74104685/how-to-install-only-production-dependencies-in-modern-yarn-when-using-node-modul)
  * v4: `yarn workspace focus --production` (no pre setting required)

However, it seems that the v3 and v2 doesn't correctly save `node_modules` subDependencies which leads to fails in the runtime (because of missing sub dep, example: `@opentelemetry/core`)

**Use pnpm**
- Create issue if you have the field `packageManager` set to `yarn`, you need to force use `pnpm` using `corepack use pnpm@<tag>`. This is not optimal at all.
- Same issues as yarn, pnpm has no such concept as not module, you can use `--shameless-hoist` to fill `node_modules` but it creates the same issue.

**Use `yarn install`**
- Same issue, I got some missing dependencies, it seems each package manager and each version of this package has a specific way of doing things.

**Force set yarn pkgManager field inside runtime**
- I'm testing right now, this is the last solution I have in mind, it's still failing :/

**Use `npm install --production`**
This is the worst solution possible because it might impact performances, but this ensure generic support no matter the package manager used
Npm works, but will reduce TS performances sadly.

### Another possible solution

Another workaround would be for you to set `packageManager`to v4.x.x which supports `yarn workspace focus --production` natively.
So we check the yarn version and then:
- if yarn version < 2 -> `yarn install --production`
- if > 4 `yarn workspace focus --production`

### DIFF
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -90,7 +90,7 @@ func (t *TypescriptSdk) ModuleRuntime(ctx context.Context, modSource *ModuleSour
 	case Node:
 		return ctr.
 			// Install dependencies
-			WithExec([]string{"npm", "install", "--omit=dev"}).
+			WithExec([]string{"yarn", "install", "--production"}).
 			// need to specify --tsconfig because final runtime container will change working directory to a separate scratch
 			// dir, without this the paths mapped in the tsconfig.json will not be used and js module loading will fail
 			// need to specify --no-deprecation because the default package.json has no main field which triggers a warning
@@ -177,8 +177,6 @@ func (t *TypescriptSdk) Base(runtime SupportedTSRuntime) (*Container, error) {
 		return dag.Container().
 			From(nodeImageRef).
 			WithoutEntrypoint().
-			// Enable corepack so we can use yarn v4 which is supposed to be faster than npm or yarn v1.
-			WithExec([]string{"corepack", "enable"}).
 			// Install default CA certificates and configure node to use them instead of its compiled in CA bundle.
 			// This enables use of custom CA certificates if configured in the dagger engine.
 			WithExec([]string{"apk", "add", "ca-certificates"}).
@@ -263,6 +261,8 @@ func (t *TypescriptSdk) installedSDK(ctr *Container, runtime SupportedTSRuntime)
 		return ctr.WithExec([]string{"bun", "install", "--no-verify", "--no-progress", "--summary"}).Directory(ModSourceDirPath)
 	case Node:
 		return ctr.
+			// Enable corepack so we can use yarn v4 which is supposed to be faster than npm or yarn v1.
+			WithExec([]string{"corepack", "enable"}).
 			WithExec([]string{"yarn", "set", "version", "stable"}).
 			WithExec([]string{"yarn", "workspaces", "focus", "--production"}).Directory(ModSourceDirPath)
 	default:
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -90,7 +90,7 @@ func (t *TypescriptSdk) ModuleRuntime(ctx context.Context, modSource *ModuleSour
 	case Node:
 		return ctr.
 			// Install dependencies
-			WithExec([]string{"yarn", "install", "--production"}).
+			WithExec([]string{"yarn", "install"}).
 			// need to specify --tsconfig because final runtime container will change working directory to a separate scratch
 			// dir, without this the paths mapped in the tsconfig.json will not be used and js module loading will fail
 			// need to specify --no-deprecation because the default package.json has no main field which triggers a warning
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -90,7 +90,8 @@ func (t *TypescriptSdk) ModuleRuntime(ctx context.Context, modSource *ModuleSour
 	case Node:
 		return ctr.
 			// Install dependencies
-			WithExec([]string{"yarn", "install", "--production"}).
+			//WithExec([]string{"yarn", "set", "version", "stable"}).
+			WithExec([]string{"npm", "install", "--omit=dev"}).
 			// need to specify --tsconfig because final runtime container will change working directory to a separate scratch
 			// dir, without this the paths mapped in the tsconfig.json will not be used and js module loading will fail
 			// need to specify --no-deprecation because the default package.json has no main field which triggers a warning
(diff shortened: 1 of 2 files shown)


Xiot:
> However, it seems that the v3 and v2 doesn't correctly save node_modules subDependencies which leads to fails in the runtime (because of missing sub dep, example: @opentelemetry/core)

Both `yarn` and `pnpm` can be configured to use the `node-linker` in their respective configuration files.

Do you actually depend on the files being in `node_modules` ?  Are you explicitly going into that folder?

Xiot:
I'm still really surprised that `yarn classic` is faster than the latest `npm` in your testing

Xiot:
Also, be careful since `corepack` is still marked as `experimental`, even in Node 22

TomChv:
> Both yarn and pnpm can be configured to use the node-linker in their respective configuration files.

How do you do that? :O

> Do you actually depend on the files being in node_modules ? Are you explicitly going into that folder?

Yes I do, because when I do the SDK installation, I want to keep the node modules so when I install the user's dependencies, it doesn't have to reinstall all the modules but only the user's.
It seems that if I don't do that, everything actually break because some sub sub dependencies will be missing because 
the user may not have a lockfile in the project, to also include *dependencies of local dependencies*.

This has 3 effects:
1. Improve performances by using yarn v4
2. Avoid potential sub dependencies issues, for example a missing module (which is commonly what happens)
3. Benefit from cache

gerhard:
I can see that the need for `--production` comes from:

```
* I moved @dagger.io/dagger from devDependencies to dependencies. That allows to use --production flag which in theory is faster than installing all dependencies.
```

Is this the list of dependencies?

https://github.com/dagger/dagger/blob/0be1329ea7cb9fcf01c4c1916e96025c72a0500d/sdk/typescript/runtime/template/package.json

TomChv:
> I can see that the need for --production comes from:

> * I moved @dagger.io/dagger from devDependencies to dependencies. That allows to use --production flag which in theory is faster than installing all dependencies.

>Is this the list of dependencies?

Yes it's the default list of dependencies, but the easier could also add its own to use them in the module.
So a common pattern in TS is to add the required dependencies in `dependencies` and the dev one's (optional like linters..) in `devDependencies`

gerhard:
I see.

Does it make sense for a TypeScript Dagger module to have development dependencies?

I am also wondering how common this is among the TypeScript modules on https://daggerverse.dev/

The reason why I'm digging into this direction is because maybe we don't need to do `--production` at all.

Xiot:
We probably don't need to worry too much about only installing the non dev dependencies. 
Even if people do have some dev tools in their list of dependencies, it's not going to make much of a difference with installs. 

Usually the deciding factor for whether something goes into `dependencies| or `devDependencies` is "do I need this at runtime?"

In a non-dagger package, `typescript` doesn't meet this criteria. You almost never publish the raw typescript files, and instead just publish the compiled JavaScript files and the `.d.ts` files. 

Dagger is a little different since it is the one doing the compilation

TomChv:
So in your opinion, I should just stick to `yarn install` and that's all?

Xiot:
Ya I would just avoid the `--production` flag.  
One of the main reasons why I put my dagger module into its own repo was so that `yarn install` only installed what I needed for dagger.  If I had this in my main mono-repo, then it would have installed everything.

Also, by setting `yarn set version stable` without a `.yarnrc` file, it will use `pnp` which probably isn't what you want since it doesn't use a `node_modules`

TomChv:
Okay, I'm going to update the PR

TomChv:
Ok PR's updated, I removed the `--production` flag

sdk/typescript/runtime/main.go
```diff
@@ -263,6 +261,8 @@ func (t *TypescriptSdk) installedSDK(ctr *Container, runtime SupportedTSRuntime)
 		return ctr.WithExec([]string{"bun", "install", "--no-verify", "--no-progress", "--summary"}).Directory(ModSourceDirPath)
 	case Node:
 		return ctr.
+			// Enable corepack so we can use yarn v4 which is supposed to be faster than npm or yarn v1.
+			WithExec([]string{"corepack", "enable"}).
 			WithExec([]string{"yarn", "set", "version", "stable"}).
 			WithExec([]string{"yarn", "workspaces", "focus", "--production"}).Directory(ModSourceDirPath)
```
Xiot, line 267: The `yarn set version stable` is what is actually enabling `yarn v4`.  
That may cause issues since it uses the `pnp` linker by default.

You also probably want to change the `yarn workspaces focus --production` to just use `yarn install`

TomChv: Hmm I'm only use it for the TS SDK and it worked well so far :O

TomChv:
CI's failing due to flacky tests, I'm going to trigger it until it's green

TomChv:
@gerhard PR's ready to be merged, I applied your suggestions and removed the prod flag

sdk/typescript/runtime/main.go
Xiot: It is using `pnp` though.

gerhard:
Taking a look at this now.

gerhard (commented):
I just tried this in the same context as:
- https://github.com/dagger/dagger/pull/7096

First run with no cache took `58s` (down from `108s`). I am going to run it again to double-check before approving & merging. Initial result is most promising 💪

gerhard (approved):
Before this change: https://dagger.cloud/dagger/traces/a0119070c442ed1c81f7eff37d6e972a

image

---
After this change: https://dagger.cloud/dagger/traces/3223f970d8972076bdea2c6e2fa98fde

image

PR #7519, opened 2024-05-31
engine: fix engine panic during dagger call ... terminal with nesting

sipsma:
TODO:
- [ ] Actually fix, just reproing in integ test so far and sending out so I don't forget about this
   * Problem is the engine hits a nil panic due to `dagql.CurrentID()` being nil during `Query.RegisterClient`, which is due to the fact that the terminal is being connected to direct over a websocket at that point, not via a gql query
   * This will also be fixed by https://github.com/dagger/dagger/pull/7315, so if that's merged first then I'll just throw this test in there so we have coverage of this flag on terminal now.
- [ ] Perhaps separate issue, but the fact that the CLI hangs when this panic happens rather than exiting is also a bug

### DIFF
--- a/core/integration/module_terminal_test.go
+++ b/core/integration/module_terminal_test.go
@@ -232,6 +232,86 @@ type Test struct {
 		err = cmd.Wait()
 		require.NoError(t, err)
 	})
+
+	t.Run("nested client", func(t *testing.T) {
+		modDir := t.TempDir()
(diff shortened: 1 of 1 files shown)


sipsma:
Okay yeah this test passes on top of https://github.com/dagger/dagger/pull/7315 due to the fact that the bug here involves client registration and that PR completely removes the entire notion of that registration (good example of why it's a worthwhile simplification...), so I just migrated the test to that PR and will let us pick up the fix from it. Closing this one out.

PR #7532, opened 2024-06-03
add `testctx` for instrumenting integration tests with OTel

vito:
Adds `testsctx`, a little `*testing.T` compatible testing library that adds `ctx` propagation and middleware. Ideally we'd extract it for general use once it's fully baked so Go folks can use it and get pretty OTel integration in Cloud.

Adds the following `testctx` middleware to our integration test suite:

* `t.Parallel` everywhere. Non-parallel tests have been adapted to work in parallel. (I think.)
  * If we need to adjust this, the middleware can be made conditional on `t.Name()` or something (i.e. put "sync" in the test name), or we can just add a non-parallel suite that doesn't use the middleware and treat it like tracking tech debt.
* OTel tracing integration: each test and sub-test becomes a span.
* OTel logging integration: `t.Log` and friends log to the OTel span.

The net result is a fully instrumented integration test suite, where you don't have to read `go test` or `gotestsum` output; you can just see everything visualized in the TUI or Cloud.

TODO (in future PRs):

* [ ] Finish `testctx` design
  * The `FooSuite` type is mostly a convenient way to corral around a set of methods, but currently the value itself serves no purpose. Worse, it's tempting to put state there, but `t.Run` sub-tests will all capture the same var, causing race conditions.
* [ ] Better parallelism internals than `t.Parallel`?
   * Probably still need use `t.Parallel` as a primitive for compatibility with `go test` - there's no way to access the `-parallel` flag value, so we have to just respect it. Don't want a custom CLI.
   * This mostly comes down to "how to initialize state without races," so maybe related to previous point

Leaving these alone for now because they're hard and I'm tired of rebasing this PR. :sweat_smile:

### DIFF
--- a/ci/test.go
+++ b/ci/test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
-	"runtime"
 
 	"github.com/moby/buildkit/identity"
 
@@ -88,24 +87,19 @@ func (t *Test) test(
 ) error {
 	cgoEnabledEnv := "0"
 	args := []string{
-		"gotestsum",
-		"--format", "testname",
-		"--no-color=false",
-		"--jsonfile=./tests.log",
-	}
-
-	if failfast {
-		args = append(args, "--max-fails=1")
+		"go",
+		"test",
 	}
 
 	// All following are go test flags
-	args = append(args, "--")
-
-	// Default parallel to number of CPUs
-	if parallel == 0 {
-		parallel = runtime.NumCPU()
+	if failfast {
+		args = append(args, "-failfast")
 	}
-	args = append(args, fmt.Sprintf("-parallel=%d", parallel))
+
+	// Go will default parallel to number of CPUs, so only pass if set
+	if parallel != 0 {
+		args = append(args, fmt.Sprintf("-parallel=%d", parallel))
+	}
 
 	// Default timeout to 30m
 	// No test suite should take more than 30 minutes to run
@@ -119,7 +113,7 @@ func (t *Test) test(
 		cgoEnabledEnv = "1"
 	}
 
-	// Disable test caching
+	// Disable test caching, since these are integration tests
 	args = append(args, fmt.Sprintf("-count=%d", count))
 
 	if testRegex != "" {
@@ -136,7 +130,6 @@ func (t *Test) test(
 	_, err = cmd.
 		WithEnvVariable("CGO_ENABLED", cgoEnabledEnv).
 		WithExec(args).
-		WithExec([]string{"gotestsum", "tool", "slowest", "--jsonfile=./tests.log", "--threshold=1s"}).
 		Sync(ctx)
 	return err
 }
@@ -192,7 +185,6 @@ func (t *Test) testCmd(ctx context.Context) (*Container, error) {
 
 	utilDirPath := "/dagger-dev"
 	tests := t.Dagger.Go().Env().
-		WithExec([]string{"go", "install", "gotest.tools/gotestsum@v1.10.0"}).
 		WithMountedDirectory(utilDirPath, testEngineUtils).
 		WithEnvVariable("_DAGGER_TESTS_ENGINE_TAR", filepath.Join(utilDirPath, "engine.tar")).
 		WithServiceBinding("dagger-engine", devEngineSvc).
--- a/.github/actions/call/action.yml
+++ b/.github/actions/call/action.yml
@@ -100,12 +100,12 @@ runs:
         if [[ "${{ inputs.module }}" == "." ]]; then
           # set some sane defaults for the current module
           if [[ -f $HOME/.docker/config.json ]]; then
-            dagger -m . call --source=".:default" --host-docker-config=file:"$HOME/.docker/config.json" ${{ inputs.function }}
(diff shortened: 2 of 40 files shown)


vito:
I think this is good enough to merge and stop battling with rebases, but no pressure obviously if you have any objections! The rebases aren't _too_ bad in truth.

With the recent TUI optimizations (#7671), cause/effect tracking (#7686), and OTel fixes (#7735) running our integration tests should finally be a nice speedy experience with full insight into what's going on, even in crazy nesting scenarios. :crossed_fingers:

Aside from the addition of `testctx` itself, the changes in this PR are largely mechanical:

```go
func TestFooBar(t *testing.T) {
  t.Parallel()
  t.Run("foo", func(t *testing.T) {
      t.Parallel()
```

=>

```go
func (FooSuite) TestBar(ctx context.Context, t *testctx.T) {
  t.Run("foo", func(ctx context.Context, t *testctx.T) {
```

Notably, `t.Parallel` is now the default thanks to the `WithParallel` middleware that we install for every suite. Most of the follow-up commits deal with the fallout from this change, since we had a few tests that weren't ready for parallelism. I think I found them all but given the nature of the change, there may be more lurking - you can ping me if you see a failure that seems suspicious.

Unfortunately this change might break some `go test -run` muscle memory since now the main test name is actually a subtest. For example using `-run ClientMulti` to match `TestClientMultiSameTrace` becomes `-run Client/Multi` since it lives beneath the `Client` test suite.

There are a couple of TODOs noted in the description, but I think it'll take a bit more thinking to figure out how to address them, and until then I don't see them causing us any harm.

jedevc:
Ah I hate to be the bearer of bad news - but it looks like the `test` and `testdev` jobs in this PR are running at least x2 as slow as currently on `main`.

ci/test.go
```diff
@@ -192,7 +185,6 @@ func (t *Test) testCmd(ctx context.Context) (*Container, error) {
 
 	utilDirPath := "/dagger-dev"
 	tests := t.Dagger.Go().Env().
-		WithExec([]string{"go", "install", "gotest.tools/gotestsum@v1.10.0"}).
```
jedevc, line 195: Phew :tada: Installing deps here was kinda bad.

vito:
@jedevc Yeah I'm not sure what's up with that. I saw the GHA jobs were slower yesterday for some reason, but if you look at the actual traces they complete faster than `main`. :thinking: 

This PR:

* [`test all`](https://julian-dev-4195-allow-users-to-select-and-focus.preview.dagger.cloud/dagger/traces/8778502b0915c9ef839e4aca6bbb80df): 9m20s
* [`test important`](https://julian-dev-4195-allow-users-to-select-and-focus.preview.dagger.cloud/dagger/traces/de09ee1578a7b74198d80ce89fdb31e4): 7m38s

`main`:

* [`test all`](https://dagger.cloud/dagger/traces/b46ad752bedea60addf53338c9b79240): 9m35s
* [`test important`](https://dagger.cloud/dagger/traces/dc4c35b86b59ff6e1f444231d0c139cc): 8m57s

There are conflicts now so I'll rebase and see if the weird GHA slowness happens again.

vito:
Happened again.

* [`main all` in Cloud](https://dagger.cloud/dagger/traces/06c037dce036f81088f29baeb8ba8e61): 11m15s (same ballpark as other PRs atm)
* [`main all` in GHA](https://github.com/dagger/dagger/actions/runs/9665431655/job/26662487796?pr=7532): 22m6s

wat?

Comparing span count:

* PR: 145k
* main: 138k

A bit of a difference there but that's not surprising considering what the PR does. There doesn't seem to be much slowdown on the Cloud side; I was able to view the completed trace there while the GHA job was ongoing.

Comparing the raw logs ([PR](https://github.com/dagger/dagger/actions/runs/9665431655/job/26662487796?pr=7532) vs. [main](https://github.com/dagger/dagger/actions/runs/9664624181/job/26659714191), there is a pretty big difference:

```
❯ wc -l /tmp/*-all.log
   178060 /tmp/main-all.log
   425990 /tmp/pr-all.log
```

Guess I'll look into that next.

vito:
@jedevc Figured it out.

Previously we would send test logs (i.e. `dagger.WithLogOutput`) to `*testing.T`, so those logs would only ever get printed in the `gotestsum` / `go test` output and only when the test failed.

But this PR changed that to send test logs to the test's OpenTelemetry span, making all of that log volume unconditional.

The thing is, those logs are completely redundant with the data already being sent over telemetry. I was already tempted to remove it because it made the output pretty verbose, although it was kind of cool to see the plain logs format side-by-side with the TUI.

So, the latest commit changes the client `NewTWriter` usage to send to `t.T` instead of `t`. It's subtle, but the net effect should be what we want: we'll collect logs and show them in `go test` output on failure (so the testing DX isn't *reliant* on OTel integration), without sending a ton of redundant data over OTel.

Seems like it did the trick; now the tests finish in approximately the same amount of time as `main`. :tada: 

(Except the latest `test` run flaked out on ProjectLayout again so I reran it.)

jedevc (approved):

jedevc:
Woo thanks for diving into all this :tada:

Merged this now, just to avoid accidentally merging something before this that would have required *more* rebase hell for you :cry:

PR #7538, opened 2024-06-03
docs: add oci annotations example

levlaz:


docs/current_docs/manuals/developer/cookbook/snippets/builds/oci/go/main.go
```diff
@@ -0,0 +1,28 @@
...
+func (m *MyModule) Build(
+	ctx context.Context,
+) (string, error) {
+	ref, err := dag.Container().
+		From("alpine").
+		WithLabel("org.opencontainers.image.title", "my-alpine").
+		WithLabel("org.opencontainers.image.version", "1.0").
+		WithLabel("org.opencontainers.image.created", time.Now().String()).
+		WithLabel("org.opencontainers.image.source", "https://github.com/alpinelinux/docker-alpine").
+		WithLabel("org.opencontainers.image.licenses", "MIT").
```
jedevc, line 20: Note: labels are *not* the same as OCI annotations. Even though the same key/value pairs are allowed.

Labels are attached to the [config object](https://github.com/opencontainers/image-spec/blob/main/config.md#properties), while annotations are attached to a [manifest](https://github.com/opencontainers/image-spec/blob/main/manifest.md)(or descriptor). This is fine (and we should aim to support annotations as well), but we should take care to call these labels.

levlaz: Thanks @jedevc - this was a straight port from this entry, so it seems like it was wrong there too? https://archive.docs.dagger.io/0.9/cookbook/#add-oci-annotations-to-image 

Ill update the description in the text to make this part clear.

jedevc: Yup, looks like the old cookbook is wrong too :scream:

jedevc (approved):

PR #7544, opened 2024-06-04
chore: add DAGGER_COMMIT option to install.sh

jedevc:
This allows installing an arbitrary commit from main, to make it easier to install versions from main for testing.

TODO:
- [x] Test `install.ps1` changes on windows

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -2,7 +2,7 @@
 
 param (
     [Parameter(Mandatory = $false)] [System.Management.Automation.SemanticVersion]$DaggerVersion,
-    [Parameter(Mandatory = $false)] [System.Management.Automation.SemanticVersion]$DaggerCommit,
+    [Parameter(Mandatory = $false)] [string]$DaggerCommit,
     [Parameter(Mandatory = $false)] [string]$InstallPath = $env:HOMEPATH + '\dagger',
 
     [Parameter(Mandatory = $false)] [System.Boolean]$InteractiveInstall = $false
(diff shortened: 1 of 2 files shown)


jedevc:
fyi @shykes

install.ps1
```diff
@@ -71,21 +72,25 @@ run the script and if it still fail please open an issue on the Dagger repo.
...
-        $version = $DaggerVersion
+        $version = "v" + $DaggerVersion
+    } elseif ($DaggerCommit) {
+        $version = $DaggerCommit
     } else {
-        $version = latest_version
+        $version = "v" + (latest_version)
     }
-    $fileName="dagger_v" + $version + "_windows_amd64"
+    $fileName="dagger_" + $version + "_windows_amd64"
```
pjmagee, line 93: I thought this PR was only for .sh but can you check my PR maybe for some improvements i.e:

- The amd64 is fixed here but Windows runs ARM too WinRT
- The $InteractiveInstall could be a `[switch]` for a cleaner looking `.\script -Install` (no bool)

jedevc: Thanks @pjmagee for #7569, they look pretty good to me!

I think both PRs do something different, this is just adding a new feature into both scripts to make it easier to consume builds from `main`. I think my idea is to try and merge this one first, and then we can rebase #7569 on top of that with changes?

pjmagee: yeah, Lets do that

jedevc: The keeping it inline was mostly just me attempting to tidy up the windows script when I reimported it - there's no issue if you want to diverge IMO, if that's more familiar to windows users :tada:

gerhard (approved):
Code changes look right to me. I also checked that:
- [x] `./install.sh` installs latest released version
- [x] `DAGGER_VERSION=v0.11.5 ./install.sh` installs specific version
- [x] `DAGGER_COMMIT=2cbda9c7547809a35eb93ff7677e5bab4afd00c1 ./install.sh` installs dev version
- [x] `DAGGER_COMMIT=2cbda9c ./install.sh` fails as expected (full git sha is required)

LGTM 👍 🚀

After this gets merged, DM me @jedevc so that I can do the CloudFront dance 🕺

PR #7549, opened 2024-06-05
core: reduce module ID size by not including full schemas

sipsma:
Before this, we were passing full introspection schema JSONs as strings as args to non-Go SDKs, which resulted in them being included in the full IDs for non-Go modules.

Those strings got huge and could appear multiple times in a single ID, which inflated their size, and also ended up in progress output (I think the source of `ETOOBIG`?) + telemetry.

This has been a known issue for a while but not harmful enough to be prioritized for a fix. However, my [other PR](https://github.com/dagger/dagger/pull/7315) that refactors a bunch of stuff around sessions ended up hitting this problem hard since it required including Module+FunctionCall IDs as HTTP headers, but then hitting limits on the max size of HTTP headers (go defaults to 1MB, which is pretty reasonable+generous).
* Some headers in that PR that previously ended up summing to 1.7MB now max out at 70kb (still too high but I think there's other issues to fix separately).

The fix here changes the introspection json to be passed as type File rather than string. The File is created using the existing blob-functionality so it's all nice and content hashed.

As I updated the SDK implementations to account for this (technically a small breaking change), I also noticed that all of them turned the string into a File anyways internally, so this seems like an all around better approach for everyone involved.

---

cc @helderco @TomChv @wingyplus this is a very small but technically breaking change to the SDKs-as-modules interface, as noted above, so just FYI. Let me know if there's anyone else I should `@`

### DIFF
--- a/core/schema/sdk.go
+++ b/core/schema/sdk.go
@@ -144,7 +144,7 @@ func (s *moduleSchema) newModuleSDK(
 
 // Codegen calls the Codegen function on the SDK Module
 func (sdk *moduleSDK) Codegen(ctx context.Context, deps *core.ModDeps, source dagql.Instance[*core.ModuleSource]) (*core.GeneratedCode, error) {
-	introspectionJSON, err := deps.SchemaIntrospectionJSON(ctx, true)
+	schemaJSONFile, err := deps.SchemaIntrospectionJSONFile(ctx, true)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get schema introspection json during %s module sdk codegen: %w", sdk.mod.Self.Name(), err)
 	}
@@ -159,7 +159,7 @@ func (sdk *moduleSDK) Codegen(ctx context.Context, deps *core.ModDeps, source da
 			},
 			{
 				Name:  "introspectionJson",
-				Value: dagql.String(introspectionJSON),
+				Value: dagql.NewID[*core.File](schemaJSONFile.ID()),
 			},
 		},
 	})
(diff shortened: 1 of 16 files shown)


wingyplus:
I think we need to add a note to Elixir SDK to pin the SDK version for any users who uses dagger released version because it can be broken if user uses the runtime on latest main branch. 

I can help update a note after this PR merged. So you might not need to worry about it.

jedevc (approved):
Nice :tada: @wingyplus if you can add a breaking change note to this PR, I think that works :tada:

We should probably note this for the engine notes (not just elixir) since it is an engine interface change. Even though it's a breaking change, I think in practice this is fine to pull into the next v0.11.X release.

wingyplus:
> Nice 🎉 @wingyplus if you can add a breaking change note to this PR, I think that works 🎉
> 
> We should probably note this for the engine notes (not just elixir) since it is an engine interface change. Even though it's a breaking change, I think in practice this is fine to pull into the next v0.11.X release.

For Elixir, add a note to the `sdk/elixir/README.md` would be enough. But for the engine, we should add a changie log right?

jedevc:
> > Nice 🎉 @wingyplus if you can add a breaking change note to this PR, I think that works 🎉
> > We should probably note this for the engine notes (not just elixir) since it is an engine interface change. Even though it's a breaking change, I think in practice this is fine to pull into the next v0.11.X release.
> 
> For Elixir, add a note to the `sdk/elixir/README.md` would be enough. But for the engine, we should add a changie log right?

Yup, exactly that SGTM :pray:

core/schema/sdk.go
```diff
@@ -469,19 +469,15 @@ func (sdk *goSDK) baseWithCodegen(
 	}
 
 	if err := sdk.dag.Select(ctx, ctr, &ctr, dagql.Selector{
-		Field: "withNewFile",
+		Field: "withMountedFile",
```
vito, line 472: Just curious: is there any reason not to use `withFile` here? I recall discussions around possibly getting rid of `withMountedXXX` since the normal versions already do the efficient MergeOp technique under the hood.

vito (approved):
:heart_eyes:

core/schema/sdk.go
sipsma: MergeOp can only be used directly when copying from `/` to `/`, otherwise there needs to be an intermediate copy step to align everything to be at `/`. There's ways of fixing that upstream (basically, use the same hardlink approach of mergeop but for all copy-ops too), but until then `WithMountedX` is still gonna be more efficient in the general case (though this only really becomes noticeable once you get to like thousands of files and/or files greater than 100MB).

sipsma:
@wingyplus good call on handling the elixir SDK breaking change more carefully! I added just a changie log for now; wasn't sure the format/way to include the breaking change note in `sdk/elixir/README.md` you'd prefer so will just leave that to you. Thank you!

sipsma:
Test failures are the ones fixed by https://github.com/dagger/dagger/pull/7452 so just gonna merge as is rather than keep re-rolling

PR #7551, opened 2024-06-05
fix(ts-sdk): fix subpath issue on codegen

TomChv:
@marcosnils raised an issue happening when loading a module as
a subpath of a root directory.

This leaded to an error happening during the diff

```
failed to diff generated code: select:
cannot diff with different relative paths
```

This PR fixes that issue by not creating a directory
as generated result code but instead reusing the
codegen dir.

**Example now**

```graphql
{
  git(url:"github.com/marcosnils/daggerverse") {
    branch(name:"main") {
     tree { 
      directory(path:"gptools") {
        asModule (sourceRootPath:"examples/typescript") {
          id
        }
      }
     }  
    }
  }
}
```

Result:

```
 dagger query --doc query.gql
{
    "git": {
        "branch": {
            "tree": {
                "directory": {
                    "asModule": {
                        "id": "ChV4eGgzOjZlNzllZjllZDVjNjlmZjUSawoVeHhoMzo0MGQ0OTRkOTFiYjcxYTA5ElIKFXh4aDM6ZWUwZjU0YzY0NmI2MmU3ORIKCgZHaXRSZWYYARoGYnJhbmNoIg4KBG5hbWUSBjoEbWFpbkoVeHhoMzo0MGQ0OTRkOTFiYjcxYTA5EoYBChV4eGgzOjZlNzllZjllZDVjNjlmZjUSbQoVeHhoMzo4NzQwODY2ZTE3MzJhMDcxEgoKBk1vZHVsZRgBGghhc01vZHVsZSInCg5zb3VyY2VSb290UGF0aBIVOhNleGFtcGxlcy90eXBlc2NyaXB0ShV4eGgzOjZlNzllZjllZDVjNjlmZjUSdAoVeHhoMzo4NzQwODY2ZTE3MzJhMDcxElsKFXh4aDM6ZDhhMTgxZTE5ODM5NTFkMBINCglEaXJlY3RvcnkYARoJZGlyZWN0b3J5IhEKBHBhdGgSCToHZ3B0b29sc0oVeHhoMzo4NzQwODY2ZTE3MzJhMDcxElwKFXh4aDM6ZDhhMTgxZTE5ODM5NTFkMBJDChV4eGgzOjQwZDQ5NGQ5MWJiNzFhMDkSDQoJRGlyZWN0b3J5GAEaBHRyZWVKFXh4aDM6ZDhhMTgxZTE5ODM5NTFkMBJ0ChV4eGgzOmVlMGY1NGM2NDZiNjJlNzkSWxIRCg1HaXRSZXBvc2l0b3J5GAEaA2dpdCIqCgN1cmwSIzohZ2l0aHViLmNvbS9tYXJjb3NuaWxzL2RhZ2dlcnZlcnNlShV4eGgzOmVlMGY1NGM2NDZiNjJlNzk="
                    }
                }
            }
        }
  
```

Signed-off-by: Tom Chauveau <tom@epitech.eu>


jedevc:
Hm @marcosnils do you have the full error message? Which two paths didn't match?

cc @sipsma, I know you fixed something like this in https://github.com/dagger/dagger/pull/7328, just wanted to check this with you.

I wonder if it might just be worth biting the bullet and removing this limitation with Diff.

TomChv:
> Hm @marcosnils do you have the full error message? Which two paths didn't match?

Here's an example

```
! failed to create module from directory: select: failed to create module: select: failed to update codegen and runtime: failed to diff generated code: select: cannot diff with different relative paths: "gptools" != "/"

Error: make request: input: git.branch.tree.directory.asModule resolve: failed to create module from directory: select: failed to create module: select: failed to update codegen and runtime: failed to diff generated code: select: cannot diff with different relative paths: "gptools" != "/"
```

jedevc:
Hm, fair enough, looks like a genuine case, and not a typo-y issue like before.

Could we have a test for this? Based on past work, this does seem to be fragile logic, so I'm a bit concerned that we don't seem to have a test that covers this edge case. Do we not have a test for module subpaths in a root directory for typescript? I think this is covered for go.

TomChv:
> Could we have a test for this? Based on past work, this does seem to be fragile logic, so I'm a bit concerned that we don't seem to have a test that covers this edge case. Do we not have a test for module subpaths in a root directory for typescript? I think this is covered for go.

We got some tests:
- With different root
https://github.com/dagger/dagger/blob/aa9fc534b0c750f71f26ae0f288c6369fa030d06/core/integration/module_typescript_test.go#L34

- Related to dependencies integrations
https://github.com/dagger/dagger/blob/aa9fc534b0c750f71f26ae0f288c6369fa030d06/core/integration/module_typescript_test.go#L623

Do you have a precise example in mind of a test that I could write??

jedevc:
Yeah I can see have tests for typescript module loading - but clearly a case is missing, for this bug to not have not been caught by the test - we should add that case.

TomChv:
> Yeah I can see have tests for typescript module loading - but clearly a case is missing, for this bug to not have not been caught by the test - we should add that case.

Okay I made a test to repro that case, I'm checking if it works locally

jedevc (approved):

marcosnils (approved):
LGTM! thx for taking care about this one Tom!

Only thing that it kinds of unsettles me a bit is that this has to be tested per-sdk runtime. I wish there was a way to either test this in isolation or test all the runtimes so we avoid it happens again.

PR #7422, opened 2024-05-20
docs: add cookbook: secrets

kpenfound:

PR #7552, opened 2024-06-05
docs: Publish cookbook

vikram-dagger:


vikram-dagger:
This does not include the secrets snippets from https://github.com/dagger/dagger/pull/7422

PR #7554, opened 2024-06-05
ci: avoid installing git on wolfi

jedevc:
It's already installed.

And also, we should move all our dependency installation *out* of here, and into previous steps, this is generally a bad practice in the first place.

### DIFF
--- a/.github/workflows/sdk-go.yml
+++ b/.github/workflows/sdk-go.yml
@@ -30,3 +30,10 @@ jobs:
     with:
       function: sdk go test
       timeout: 10
+
+  test-publish:
+    uses: ./.github/workflows/_dagger_call.yml
+    secrets: inherit
+    with:
+      function: sdk go publish --dry-run=true --tag=$GITHUB_REF
+      timeout: 10
--- a/.github/workflows/sdk-python.yml
+++ b/.github/workflows/sdk-python.yml
@@ -30,3 +30,10 @@ jobs:
     with:
       function: sdk python test
       timeout: 10
+
+  test-publish:
+    uses: ./.github/workflows/_dagger_call.yml
+    secrets: inherit
+    with:
+      function: sdk python publish --dry-run=true --tag=$GITHUB_REF
+      timeout: 10
--- a/.github/workflows/sdk-typescript.yml
+++ b/.github/workflows/sdk-typescript.yml
@@ -30,3 +30,10 @@ jobs:
     with:
       function: sdk typescript test
       timeout: 10
+
+  test-publish:
+    uses: ./.github/workflows/_dagger_call.yml
+    secrets: inherit
+    with:
+      function: sdk typescript publish --dry-run=true --tag=$GITHUB_REF
+      timeout: 10
--- a/ci/sdk_go.go
+++ b/ci/sdk_go.go
@@ -134,12 +134,13 @@ type gitPublishOpts struct {
 func gitPublish(ctx context.Context, opts gitPublishOpts) error {
 	base := opts.sourceEnv
 	if base == nil {
-		base = dag.Container().From(consts.AlpineImage)
+		base = dag.Container().
+			From(consts.AlpineImage).
+			WithExec([]string{"apk", "add", "-U", "--no-cache", "git"})
 	}
 
 	// FIXME: move this into std modules
 	git := base.
-		WithExec([]string{"apk", "add", "-U", "--no-cache", "git"}).
 		WithExec([]string{"git", "config", "--global", "user.name", opts.username}).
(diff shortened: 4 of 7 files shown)


gerhard (commented):
Good catch!

I think that we should either:
- Move `test-publish` to the publish workflows
- Move publishing into the `sdk-*` workflows

As it currently stands, the publishing logic is currently split between the two workflows.

Also, we should do the same for all other SDKs.

jpadams:
Not sure how similar the wolfi module is vs just using
```
docker run -it --rm cgr.dev/chainguard/wolfi-base sh
```
```
/ # apk add go=1.22.3
fetch https://packages.wolfi.dev/os/aarch64/APKINDEX.tar.gz
ERROR: unable to select packages:
  go-1.22-1.22.4-r0:
    breaks: world[go=1.22.3]
```

if instead I do this and get `go 1.22.4` and don't require `1.22.3`, it completes.
```
/ # apk add go git
fetch https://packages.wolfi.dev/os/aarch64/APKINDEX.tar.gz
(1/36) Installing libbrotlicommon1 (1.1.0-r2)
(2/36) Installing libbrotlidec1 (1.1.0-r2)
(3/36) Installing libnghttp2-14 (1.62.1-r0)
(4/36) Installing libunistring (1.2-r2)
(5/36) Installing libidn2 (2.3.7-r2)
(6/36) Installing libpsl (0.21.5-r2)
(7/36) Installing libcurl-openssl4 (8.8.0-r0)
(8/36) Installing libexpat1 (2.6.2-r1)
(9/36) Installing libpcre2-8-0 (10.43-r1)
(10/36) Installing git (2.45.2-r0)
(11/36) Installing ncurses-terminfo-base (6.4_p20231125-r3)
(12/36) Installing ncurses (6.4_p20231125-r3)
(13/36) Installing bash (5.2.21-r4)
(14/36) Installing libgcc (13.2.0-r7)
(15/36) Installing libstdc++ (13.2.0-r7)
(16/36) Installing binutils-gold (2.42-r2)
(17/36) Installing binutils (2.42-r2)
(18/36) Installing libstdc++-dev (13.2.0-r7)
(19/36) Installing posix-cc-wrappers (1-r3)
(20/36) Installing libatomic (13.2.0-r7)
(21/36) Installing gmp (6.3.0-r2)
(22/36) Installing libgo (13.2.0-r7)
(23/36) Installing libgomp (13.2.0-r7)
(24/36) Installing isl (0.26-r3)
(25/36) Installing mpfr (4.2.1-r3)
(26/36) Installing mpc (1.3.1-r3)
(27/36) Installing gcc (13.2.0-r7)
(28/36) Installing libxcrypt-dev (4.4.36-r7)
(29/36) Installing linux-headers (6.6.29-r1)
(30/36) Installing nss-db (2.39-r6)
(31/36) Installing nss-hesiod (2.39-r6)
(32/36) Installing glibc-dev (2.39-r6)
(33/36) Installing make (4.4.1-r3)
(34/36) Installing pkgconf (2.2.0-r2)
(35/36) Installing build-base (1-r8)
(36/36) Installing go-1.22 (1.22.4-r0)
OK: 673 MiB in 50 packages
```

jedevc:
A couple questions about wolfi:
- Does this just happen from time to time? Do dependencies just go missing? If so, that's a bit annoying.
- I don't quite understand why installing git would cause conflicts for go.

shykes:
> A couple questions about wolfi:
> 
> * Does this just happen from time to time? Do dependencies just go missing? If so, that's a bit annoying - that has the potential to just spontaneously break our CI even with no changes, I'd like to not do that if at all possible.
> * I don't quite understand why installing git would cause conflicts for go.

cc @amouat, we're trying to use Wolfi more in our CI, but running into some issues which this PR tries to fix.

jedevc:
> I think that we should either:
> 
>     * Move `test-publish` to the publish workflows
> 
>     * Move publishing into the `sdk-*` workflows

The reason to have these separate at the moment is so that the test-publish can actually occur on PR runs - atm, the first sign that publishing could have gone wrong is during release (either on main, or from tags).

Some initial refactoring of the workflow/job structure is in https://github.com/dagger/dagger/pull/7483, I think this makes a lot of sense in that context.

I think ideally we should move publish into the SDK workflows. Then we use a conditional - if it was triggered by a PR, we do the dry-run, otherwise, we do the actual publish.

amouat:
@shykes let me take a look and get back to you

jedevc:
Aha thanks @amouat :tada: 

What's weird to me is that this issue only triggers when we try and call `apk` manually - when using an apko config file (created by https://github.com/vito/daggerverse/blob/main/apko/apko.bass, cc @vito), this seems to "work".

vito (approved):

PR #6971, opened 2024-04-01
Fix Python dev module

helderco:
The Python dev module has been broken since the removal of `dag.Host`. This is a new implementation that fixes it. 

## Notes

- Adopts uv for a simpler project management
- Removes the upper bound on ruff and pytest
- Replaces black with ruff formatter
- Delegates most of the logic from `dev/sdk_python.go` to  `sdk/python/dev`

PR #7557, opened 2024-06-05
Chore: start fixing sdk/python/dev

shykes:
- Remove deprecated 'root' field
- Convert filters on deprecated root, to a view
- Update broken dependency

Note: the module itself still doesn't build


sdk/python/dagger.json
jedevc, line 1: Why is this a separate file?

jedevc: Also, huh, do views allow you to split out onto separate files like this? If so, nice 🎉 I didn't know that

shykes: > Why is this a separate file?

- `sdk/python/dev/dagger.json`: an actual dagger module, to help develop the Python SDK
- `sdk/python/dagger.json`: configure views for `sdk/python/`, for consumption by other modules

helderco:
Please don’t waste your time here.

I have a draft pr for this but it’s very outdated. I have more local changes that I haven’t submitted. It’s not just a fix for the removed root, I’m changing more things. That took me on a curve to explore some new tools in the ecosystem like rye, uv, and changes in hatch, for a std module, including exploring multiple ways to structure said dagger module. In the end I only moved forward implementing some of that in the runtime module but then had to prioritize other things like the CLI. 

Just wanted to say that this is not forgotten and there’s more changes coming for the Python dev module that will replace this and the Python part of the ci module.

shykes:
Thanks for the info. Not a waste of time so far.

shykes:
Since @helderco is already working on a python implementation that's further ahead than #6971, I'm going to focus on the interface between top-level module and each per-SDK module. I'll make my skeleton implementation in Go, to move faster, and emphasize that it's a throwaway, to be replaced by the more complete, and more native, Python implementation.

PR #7562, opened 2024-06-06
chore: update alpine base to 3.20

jedevc:
Revival of https://github.com/dagger/dagger/pull/7323.

This is motivated slightly by some of the issues in https://github.com/dagger/dagger/pull/7554 - running up-to-date software is pretty important.

### DIFF
--- a/ci/consts/versions.go
+++ b/ci/consts/versions.go
@@ -9,19 +9,21 @@ const (
 )
 
 const (
-	GolangVersion = "1.22.3"
+	GolangVersion = distconsts.GolangVersion
+	GolangImage   = distconsts.GolangImage
 	// GolangVersionRuncHack needs to be 1.21, since 1.22 is not yet
 	// supported, and can cause crashes: opencontainers/runc#4233
-	GolangVersionRuncHack = "1.21.7"
+	GolangVersionRuncHackVersion = "1.21.7"
+	GolangVersionRuncHackImage   = "golang:" + GolangVersionRuncHackVersion + "-alpine"
+
+	AlpineVersion = distconsts.AlpineVersion
+	AlpineImage   = distconsts.AlpineImage
+
+	WolfiImage   = "cgr.dev/chainguard/wolfi-base"
+	WolfiVersion = "latest" // Wolfi is a rolling distro; no release to pin to
 
 	GolangLintVersion = "v1.57"
-
-	AlpineVersion = "3.18"
-	AlpineImage   = "alpine:" + AlpineVersion
-	WolfiImage    = "cgr.dev/chainguard/wolfi-base"
-	WolfiVersion  = "latest" // Wolfi is a rolling distro; no release to pin to
-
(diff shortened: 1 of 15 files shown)


core/integration/container_test.go
```diff
@@ -78,7 +79,9 @@ func TestContainerFrom(t *testing.T) {
 			}
 		}`, &res, nil)
 	require.NoError(t, err)
-	require.Equal(t, res.Container.From.File.Contents, "3.18.2\n")
+
+	releaseStr := res.Container.From.File.Contents
+	require.True(t, strings.HasPrefix(releaseStr, distconsts.AlpineVersion), releaseStr)
```
vito, line 84: super nitpick: this will give a failure like `false is not true` which isn't very helpful, might be better to assert against `distconsts.AlpineVersion+"\n"`

not blocking, since these tests _probably_ won't start failing anyway

vito (approved):

core/integration/container_test.go
gerhard: `+1`

engine/distconsts/consts.go
```diff
@@ -16,3 +16,11 @@ const (
 	PythonSDKManifestDigestEnvName     = "DAGGER_PYTHON_SDK_MANIFEST_DIGEST"
 	TypescriptSDKManifestDigestEnvName = "DAGGER_TYPESCRIPT_SDK_MANIFEST_DIGEST"
 )
+
+const (
+	AlpineVersion = "3.20.0"
+	AlpineImage   = "alpine:" + AlpineVersion
+
+	GolangVersion = "1.22.3"
```
gerhard, line 24: Should we bump this to `1.22.4` so that we benefit from the [latest security fixes](https://github.com/golang/go/issues?q=milestone%3AGo1.22.4+label%3ACherryPickApproved)?

ci/consts/versions.go
```diff
@@ -9,19 +9,21 @@ const (
 )
 
 const (
-	GolangVersion = "1.22.3"
+	GolangVersion = distconsts.GolangVersion
+	GolangImage   = distconsts.GolangImage
 	// GolangVersionRuncHack needs to be 1.21, since 1.22 is not yet
 	// supported, and can cause crashes: opencontainers/runc#4233
-	GolangVersionRuncHack = "1.21.7"
+	GolangVersionRuncHackVersion = "1.21.7"
```
gerhard, line 16: Should we bump this to `1.21.11` so that we benefit from the latest security fixes?

gerhard (approved):
LGTM 💪 🚀

Left a few comments which can be addressed as follow-ups, no need to hold this.

engine/distconsts/consts.go
jedevc: Done!

core/integration/container_test.go
jedevc: :tada: Fixed this up now!

Issue #7392, opened 2024-05-16
✨ Improved garbage collection defaults


jedevc:
Spinning this out of a private discord discussion [here](https://discord.com/channels/707636530424053791/1239760726801776680/1240332775819575296).

We inherit [buildkit's default gc policies](https://github.com/moby/buildkit/blob/75bbf7b2071ef8b1d8dff70d4d3d878a53c6d991/cmd/buildkitd/config/gcpolicy.go#L71-L97) - which is to try and keep cache usage under 10% of available disk.

This is not actually great for a lot of cases. Here's a few dagger scenarios, and what kind of cache policies we might want for these:
- Using dagger on a personal linux machine
	- I want dagger to use quite a lot of space - I use dagger for most of my dev work, so maybe it could use 20% of my disk (on the scale of several hundred gigs of data).
- Using dagger in docker-desktop/lima/etc on a personal mac/windows/linux machine
	- Because this is a VM, there's a disk provisioned for it - that's usually quite small (the docker desktop seems unreasonably so sometimes).
	- I want dagger to use most of the available disk storage (but not all), maybe something like 50% (this works out to *mostly* the same as the above option - but now there are two dials to turn, the docker desktop disk size dial, and the dagger disk size dial).
- Using dagger in a dedicated CI worker
	- I want dagger to use as much storage as it requires - I want the limit to be something like 80-90%.

At the moment, we just have a flat "10%" for everything - which is pretty bad, *especially* for the last scenario.

Part of this is related to documentation + usability of configuring cache policy in the first place:
- https://github.com/dagger/dagger/issues/7022
- https://github.com/dagger/dagger/issues/6260

However, this is still different - we should have a way of making sure that the *defaults* are good for 90% of users. There's unfortunately no easy way to detect what kind of case we're in.

I think this makes sense more sense in the context of #5583 though. When we start a dagger engine with a driver, we can allow specifying the cache policies for that engine, and potentially automatically pick some reasonable defaults (e.g. based on the presence of `GITHUB_` env vars, etc).

Scalahansolo:
One thing to call out here.... If running Dagger in K8s, Dagger will try to keep it's memory cache around the **requested** memory, but has the potential to spike well above that. Our memory limit is set really high to account for spikes, but GC can be psuedo tuned by setting that value.

jedevc:
Yeah, that feels like a reasonable option for the short-term honestly.

[...]

PR #7563, opened 2024-06-06
engine: bump default oci worker keep to 75%

marcosnils:
stopgap fix for #7392. Increasing the GC keep size to more reasonable defaults
so users don't get cache evicted so aggresively.


jedevc:
nit: can we change this in the codebase itself, instead of adding a config setting?

https://github.com/dagger/dagger/blob/8c5f0c2d08048aa9d9a9f1b3b986368af4b788e5/cmd/engine/main_oci_worker.go#L160

jedevc (approved):
Nice! :tada:

Probably worth release-noting, but otherwise LGTM!

marcosnils:
> Nice! 🎉
> 
> Probably worth release-noting, but otherwise LGTM!

lol. Always forget about my Changie friend. I'll add a quick not there.

PR #7566, opened 2024-06-06
doc(sdk/elixir): warning user about runtime breaking changes

wingyplus:
The runtime are not compatible with older dagger. It's need to specify a version in the `--sdk` option to make works.

### DIFF
--- a/sdk/elixir/README.md
+++ b/sdk/elixir/README.md
@@ -60,6 +60,10 @@ The SDK support the Dagger Function by initiate it with:
 $ dagger init --sdk=github.com/dagger/dagger/sdk/elixir/runtime <name>
 ```
 
+**CAUTIONS**: Please note that `dagger` version 0.11.6 and earlier are not
+compatible with the runtime on `main` branch, please pin the sdk to `github.com/dagger/dagger/sdk/elixir/runtime@sdk/elixir/v0.11.6`
+instead.
+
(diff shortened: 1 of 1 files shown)


sdk/elixir/README.md
```diff
@@ -60,6 +60,10 @@ The SDK support the Dagger Function by initiate it with:
 $ dagger init --sdk=github.com/dagger/dagger/sdk/elixir/runtime <name>
 ```
 
+**CAUTIONS**: Please note that `dagger` version 0.11.6 and earlier are not
+compatible with the runtime on `main` branch, please pin the sdk to `github.com/dagger/dagger/sdk/elixir/runtime@sdk/elixir/v0.11.6`
```
sipsma, line 64:
```suggestion
compatible with the runtime on `main` branch. If you are using `dagger` v0.11.6, please pin the sdk to `github.com/dagger/dagger/sdk/elixir/runtime@sdk/elixir/v0.11.6`
```

nit: slight tweak to make it extra clear you only need to pin if you are on that dagger version, just IMO, non-blocking

sipsma (approved):
one nit if you agree with it, otherwise LGTM!

sdk/elixir/README.md
wingyplus: Thank you. 🙏

sipsma:
Netlify deploy preview failure shouldn't be relevant here, so merging

PR #7571, opened 2024-06-07
Various TS + general test fixes + improvements

sipsma:
While debugging something nodejs-related in https://github.com/dagger/dagger/pull/7315 I noticed a bunch of problems with the TS tests and some related areas:
1. Fix missing awaits in TS tests that caused accidental parallelism + tidy up env restoration: https://github.com/dagger/dagger/commit/15c43aef428a74dbe9c18d3c08ae2feba1e4a2d6 https://github.com/dagger/dagger/commit/32638a4809cfca968ebacddfb4e2428d9d769fed
2. Fix bun image to be multiplatform and upgrade bun version to avoid SIGSEGV/SIGABRT crashes: https://github.com/dagger/dagger/commit/22653f1c951abec324440582b6c8eba31e12e3a7 https://github.com/dagger/dagger/commit/f884fe9dcd6010a1fbedb3c684760d4ce51d2db5
3. Fill in a bunch of module integ tests that were missing `t.Parallel()` calls https://github.com/dagger/dagger/pull/7571/commits/b4e860f2738e361bd446b347822128cfa1fcf3f4

Sort of a grab-bag but all small + related enough to just group together in this PR.

I can't be 100% sure, but I was seeing a lot of flakes locally that looked similar to the ones in our CI that went away after https://github.com/dagger/dagger/commit/15c43aef428a74dbe9c18d3c08ae2feba1e4a2d6 specifically, so 🤞those are gone. cc @kpenfound

### DIFF
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -12,11 +12,11 @@ import (
 )
 
 const (
-	bunVersion  = "1.0.27"
+	bunVersion  = "1.1.12"
 	nodeVersion = "21.3"
 
 	nodeImageDigest = "sha256:3dab5cc219983a5f1904d285081cceffc9d181e64bed2a4a18855d2d62c64ccb"
-	bunImageDigest  = "sha256:a87fa7c4c7fbd2e98cf9c5616e1d47aed65f96805c7a7830f734b27ef8bd75de"
+	bunImageDigest  = "sha256:6568a679b87107d3d7d46b829f614c443e73bbe3bf7d6ea5c9ceb8f845869c96"
 
 	nodeImageRef = "node:" + nodeVersion + "-alpine@" + nodeImageDigest
 	bunImageRef  = "oven/bun:" + bunVersion + "-alpine@" + bunImageDigest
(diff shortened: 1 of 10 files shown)


sdk/typescript/runtime/main.go
```diff
@@ -12,11 +12,11 @@ import (
 )
 
 const (
-	bunVersion  = "1.0.27"
+	bunVersion  = "1.1.12"
 	nodeVersion = "21.3"
 
 	nodeImageDigest = "sha256:3dab5cc219983a5f1904d285081cceffc9d181e64bed2a4a18855d2d62c64ccb"
-	bunImageDigest  = "sha256:82d3d3b8ad96c4eea45c88167ce46e7e24afc726897d48e48cc6d6bf230c061c"
+	bunImageDigest  = "sha256:6568a679b87107d3d7d46b829f614c443e73bbe3bf7d6ea5c9ceb8f845869c96"
```
jedevc, line 19: Hm, this has happened before as well.

I wonder if we need some more tests where we *don't* have emulation enabled to be able to catch these.

jedevc (approved):

TomChv (approved):

sdk/typescript/runtime/main.go
sipsma: The thing is, I did have emulation enabled (`docker run --privileged --rm tonistiigi/binfmt --install all`) and this still failed. Mildly confused as to why... but I think the only way to catch this robustly is to just have ARM runners in addition to x86.

PR #7584, opened 2024-06-07
ts: add temporary standalone sdk importable from git

sipsma:
This will allow users to workaround the lack of pinning of tsx in previous engine releases until our next release.


sdk/typescript/standalone/.gitignore
```diff
@@ -0,0 +1,4 @@
+/dagger.gen.go
+/internal/dagger
+/internal/querybuilder
+/internal/telemetry
```
TomChv, line 4: Why the files in `internal` are pushed if they are ignored by this files?

TomChv (commented):
Thanks for the changes! How the user can use this runtime though?

Can you show me an example?

sdk/typescript/standalone/.gitignore
sipsma: I probably added them before adding the .gitignore. I'm gonna leave them in so we don't have weird linter complaints.

sipsma:
> Thanks for the changes! How the user can use this runtime though?
> 
> Can you show me an example?

Going to include instructions in https://github.com/dagger/dagger/pull/7584 once this is merged

The tl;dr is the example in discord earlier: https://discord.com/channels/707636530424053791/1003718839739105300/1248698818212532316

sdk/typescript/standalone/.gitignore
TomChv: Okay!

TomChv (approved):

PR #7586, opened 2024-06-07
attachable terminal sessions

aluzzardi:
- Use session attachables to pop open terminals from the engine to the
  CLI
- Reworked the CLI logic to accomodate attachable terminals
- Updated `Container.Terminal()` signature to return `Container`
- Added `Directory.Terminal()`


cmd/dagger/terminal.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+var terminalMu sync.Mutex
+
+func withTerminal(fn func(stdin io.Reader, stdout, stderr io.Writer) error) error {
+	// only allow one terminal session at a time
+	terminalMu.Lock()
+	defer terminalMu.Unlock()
+
+	if silent || !(progress == "auto" && hasTTY || progress == "tty") {
```
aluzzardi, line 18: @vito I copied this logic over -- any chance we could make terminals work in plain mode?

cmd/dagger/call.go
```diff
@@ -60,29 +60,10 @@ dagger call lint stdout
...
-			return fmt.Errorf("running shell without the TUI is not supported")
-		}
-		if outputPath != "" {
-			return fmt.Errorf("running shell with --output is not supported")
-		}
 		return nil
 	},
 	AfterResponse: func(c *FuncCommand, cmd *cobra.Command, modType *modTypeDef, response any) error {
 		switch modType.Name() {
-		case Terminal:
```
sipsma, line 80: nit: can delete `Terminal` from here too now I think https://github.com/sipsma/dagger/blob/22deba615a453987f70f0f90c9e97f19d0cf18a2/cmd/dagger/functions.go#L28-L28

cmd/dagger/terminal.go
```diff
@@ -0,0 +1,54 @@
...
+	"io"
+	"sync"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+var terminalMu sync.Mutex
+
+func withTerminal(fn func(stdin io.Reader, stdout, stderr io.Writer) error) error {
+	// only allow one terminal session at a time
```
sipsma, line 14: If more than one come in at the same time does it just drop you into them consecutively? No problem with that, just confirming it's that vs. only one will be successful

core/terminal.go
```diff
@@ -52,41 +29,73 @@ type TerminalArgs struct {
...
+	if ctr == nil {
+		ctr, err = NewContainer(dir.Query, dir.Platform)
 		if err != nil {
-			bklog.G(r.Context()).WithError(err).Error("terminal handler failed to upgrade")
-			w.WriteHeader(http.StatusInternalServerError)
-			return
+			return err
 		}
-		defer ws.Close()
+		ctr, err = ctr.From(ctx, "alpine:3.20")
```
sipsma, line 65: We have a const for the current version of alpine images we use here: https://github.com/sipsma/dagger/blob/22deba615a453987f70f0f90c9e97f19d0cf18a2/ci/consts/versions.go#L20-L20

Could move that const to `engine/distconsts/consts.go` so then it can be shared here and in our CI module. Just in the interest of keeping everything up to date easily.

core/terminal.go
```diff
@@ -1,44 +1,21 @@
...
-}
-
-func (*Terminal) TypeDescription() string {
-	return "An interactive terminal that clients can connect to."
-}
-
-func (term *Terminal) WebsocketURL() string {
-	return fmt.Sprintf("ws://dagger/%s", term.Endpoint)
-}
+const terminalPrompt = `PS1=\\033[38;5;11mdagger:\w$ \\033[0m`
```
sipsma, line 18: nit: a comment on what this actually looks like would help for non-native tty code speakers 😄

core/terminal.go
```diff
@@ -52,41 +29,73 @@ type TerminalArgs struct {
...
+) error {
+	var err error
+
+	if ctr == nil {
+		ctr, err = NewContainer(dir.Query, dir.Platform)
 		if err != nil {
-			bklog.G(r.Context()).WithError(err).Error("terminal handler failed to upgrade")
-			w.WriteHeader(http.StatusInternalServerError)
-			return
+			return err
```
sipsma, line 63: nit: here and elsewhere, I've been trying to be more consistent about always wrapping/annotating error messages, just because it helps trace down exactly where they are coming from when they get hit rather than having to guess

core/terminal.go
```diff
@@ -102,79 +111,80 @@ func (container *Container) runTerminal(
...
+	eg, egctx := errgroup.WithContext(ctx)
 	runningSvc, err := svc.Start(
 		ctx,
 		svcID,
 		true,
-		func(w io.Writer, svcProc bkgw.ContainerProcess) {
+		func(stdin io.Writer, svcProc bkgw.ContainerProcess) {
+			eg.Go(func() error {
+				_, err := io.Copy(stdin, term.Stdin)
+				if err != nil {
```
sipsma, line 131: should this get the same "ignore `io.ErrClosedPipe`" check as the other copies below?

engine/session/terminal.go
```diff
@@ -0,0 +1,157 @@
...
+			return nil
+		}
+	}
+}
+
+func (s TerminalAttachable) sendSize(srv Terminal_SessionServer, stdout io.Writer) error {
+	f, ok := stdout.(*os.File)
+	if !ok || !isatty.IsTerminal(f.Fd()) {
+		slog.Debug("stdin is not a terminal; cannot get terminal size")
+		return nil
```
sipsma, line 102: seems like it should just return an err in this case I think?

engine/session/terminal.go
```diff
@@ -0,0 +1,157 @@
...
+			return fmt.Errorf("error reading terminal: %w", err)
+		}
+		switch msg := req.GetMsg().(type) {
+		case *SessionRequest_Stdout:
+			stdout.Write(msg.Stdout)
+		case *SessionRequest_Stderr:
+			stderr.Write(msg.Stderr)
+		case *SessionRequest_Exit:
+			fmt.Fprintf(stderr, "exit %d\n", msg.Exit)
+			return nil
```
sipsma, line 93: these errors feel worth checking/returning

cmd/dagger/terminal.go
aluzzardi: Yes, one after the other

core/terminal.go
aluzzardi: @vito Are there TUI utilities for coloring I could reuse instead of manual escapes?

core/terminal.go
```diff
@@ -102,79 +112,84 @@ func (container *Container) runTerminal(
...
-			})
-		}
+	bk := container.Query.Buildkit
+	term, err := bk.OpenTerminal(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to open terminal: %w", err)
 	}
 
+	fmt.Fprintf(term.Stderr, "Attaching interactive terminal: ")
+	idtui.DumpID(idtui.NewOutput(term.Stderr, termenv.WithTTY(true)), svcID)
```
aluzzardi, line 122: @vito does this look ok to you? is this the best way there is to represent "where we are"?

Screenshot 2024-06-10 at 11 13 28 AM

core/terminal.go
jedevc: `termenv` is pretty good for this - we use a lot of it in `frontend.go` fyi.

aluzzardi: Thank you @jedevc! Updated

aluzzardi:
@sipsma Comments addressed and tests fixed, PTAL when you get a chance

core/integration/module_terminal_test.go
```diff
@@ -189,8 +202,8 @@ type Test struct {
 		require.NoError(t, err)
 
 		// cache the returned container so there's less to wait for in the shell invocation below
-		_, err = hostDaggerExec(ctx, t, modDir, "--debug", "call", "ctr", "sync")
-		require.NoError(t, err)
+		// _, err = hostDaggerExec(ctx, t, modDir, "--debug", "call", "ctr", "sync")
+		// require.NoError(t, err)
```
sipsma, line 206: is this strictly necessary to comment out, or leftover from testing?

core/schema/container.go
```diff
@@ -475,7 +475,7 @@ func (s *containerSchema) Install() {
 			absolutely necessary and only with trusted commands.`),
 
 		dagql.NodeFunc("terminal", s.terminal).
-			Doc(`Return an interactive terminal for this container using its configured default terminal command if not overridden by args (or sh as a fallback default).`).
+			Doc(`Opens an interactive terminal for this container using its configured default terminal command if not overridden by args (or sh as a fallback default).`).
```
sipsma, line 478: One thing I just remembered, we want to call `.Impure(...)` on these because they are side-effectful now rather than content defined. Otherwise dagql caching will prevent the same call from running more than once in a session (which isn't gonna be a common case, but still realistic).

Same for the API on directory.

sipsma (approved):
The comment about calling `Impure` is important but super quick; one other quick comment about commented out code. Otherwise LGTM!

core/integration/module_terminal_test.go
aluzzardi: Sorry, this was meant to be removed. Since I moved the `Terminal` call into the code itself (rather than CLI invoked), this is causing the terminal to pop open, resulting in test failure.

core/schema/container.go
aluzzardi: Done!

I wasn't aware of `.Impure()` ... does this mean it would be technically possible to return the modified container/directory without affecting cache?

sipsma: No `Impure` only affects dagql-level caching, buildkit caching is completely separate atm.
* (tangential) If we move off LLB then things like this would be possible, just noting since this topic keeps popping it's head up in random places a lot recently...

aluzzardi:
@sipsma Comments addressed. A couple of self comments about the design:

- Does it still make sense to have a `WithDefaultTerminalCmd`? I imagine this made sense when `Terminal` was supposed to be called from the CLI to give some kind of control to the code, but since now it's not the case, wondering if we should get rid of it to reduce the scope of the API
- `Impure` question above -- still wondering if it would make sense to return modified state from `Terminal`. I'm thinking it won't be easy to change later on since it's a somewhat breaking change in behavior, yet it won't fail compilation but yield a different result. The use case is not that crucial, but still pretty cool -- for instance: run in debugging mode, see a command fail (e.g. `command not found: curl`), pop a terminal, fix with apk add, exit, see if the build works, then adjust the pipeline.

core/schema/container.go
aluzzardi: Then disregard the second half of my comment below :)

aluzzardi:
sdk rust test failed twice in a row, re-retrying

sipsma:
@aluzzardi I tried locally quick and realized that because `Terminal` returns a container, the CLI ends up printing `Container evaluated. Use "dagger call ctr terminal --help" to see available sub-commands.`, which is due to the fact that `call` commands that end in non-scalars will just have `Sync` called on them and then print that warning.

I think we want to tidy that up for this case. Doesn't have to be this PR necessarily but should before this is released ideally. 

> Does it still make sense to have a WithDefaultTerminalCmd? I imagine this made sense when Terminal was supposed to be called from the CLI to give some kind of control to the code, but since now it's not the case, wondering if we should get rid of it to reduce the scope of the API

(somewhat related to above) I think it should still be totally valid to call this from the CLI though right? Generally speaking, any of the APIs on core types like Container/Directory/etc. should work from the CLI too, you don't *have* to update your code (even if that's a new possibility now)

engine/buildkit/client.go
```diff
@@ -577,6 +577,112 @@ func (c *Client) ListenHostToContainer(
...
+	Stderr   io.WriteCloser
+	ResizeCh chan bkgw.WinSize
+	ErrCh    chan error
+	Close    func(exitCode int) error
+}
+
+func (c *Client) OpenTerminal(
+	ctx context.Context,
+) (*TerminalClient, error) {
+	terminalClient := session.NewTerminalClient(c.Opts.MainClientCaller.Conn())
```
aluzzardi, line 592: @sipsma What is the equivalent now on top of your commits?

engine/client/client.go
```diff
@@ -341,6 +343,8 @@ func (c *Client) startSession(ctx context.Context) (rerr error) {
 		authprovider.NewDockerAuthProvider(config.LoadDefaultConfigFile(os.Stderr), nil),
 		// host=>container networking
 		session.NewTunnelListenerAttachable(ctx, nil),
+		// terminal
+		session.NewTerminalAttachable(ctx, c.Params.WithTerminal),
```
aluzzardi, line 347: @sipsma rebased, LGTY?

engine/buildkit/client.go
sipsma: Ah yeah, sorry we realized we needed that big PR quickly since it ended up fixing some other pre-existing issues and merged it asap, the equivalent now is `c.getMainClientCaller()`, e.g. how it's used here: https://github.com/sipsma/dagger/blob/546c900580c7558befe2c0a9caa5e9bbfb00f9c5/engine/server/socket.go#L59-L59

engine/client/client.go
sipsma: Yep looks right!

engine/buildkit/client.go
aluzzardi: 🙏

aluzzardi:
Waiting for v0.11.8 to be tagged before merging /cc @sipsma

PR #7587, opened 2024-06-08
Improve standard go linter

shykes:
First pass at improving our standard Go linter.

- Skeleton implementation works, but needs refinement
- Breaks the rest of our CI, which depends on the old `Lint()`. Need to wire in the new `LintAssertPass` for retro-compat
- Ideally start using the new `Lint` directly at least in one place.

### DIFF
--- a/ci/default-golangci.yml
+++ b/ci/default-golangci.yml
@@ -27,6 +27,7 @@ linters:
     - unused
     - unparam
     - whitespace
+    - gomodguard
 
 issues:
   exclude-rules:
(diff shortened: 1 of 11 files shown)


Vad1mo:
@shykes can you attach a few before after samples. 
Adopting Dagger to a wide community with knowledge spread across the spectrum, we currently have a hard time arguing for dagger if the output that people are familiar with becomes a hurdle. 

[See Discord discussion](https://discord.com/channels/707636530424053791/708371226174685314/1249756422086987991)

 
We are genuinely interested in that topic, happy to help here.

Issue #7522, opened 2024-05-31
🐞 Secrets in Dockerfile not available with buildkit directive


Alvise88:
### What is the issue?

Secrets building container in dagger with dockerfile are not available. Lastest working version is v0.9.11

### Dagger version

v0.11.6

### Steps to reproduce

You can find an example in this repository: https://github.com/Alvise88/dagger-dockerfile-secret-issue

marcosnils:
I was able to reproduce this and it seems to be related to the #syntax directive at the very top of the Dockerfile. If you remove that, it should work.  ref: https://discord.com/channels/707636530424053791/1245745373356625990/1246142688504320134

If you are able to validate this, could you please update the issue description? 

Thanks!

Alvise88:
Ok, removing buildkit directive it starts working properly.

PR #7595, opened 2024-06-10
fix: ensure nested frontend builds get secret translation

jedevc:
Fixes https://github.com/dagger/dagger/issues/7522.

When doing a nested frontend build - i.e. when dockerfile.v0 calls gateway.v0 (triggered by a syntax directive), we need to ensure that the second frontend gets the same secret translation as each prior layer.
    
To do this, we heavily refactor the filtering gateway and essentially have it pass itself into the frontend Solve - this propagates the filtering all the way down, so secrets can be correctly accessed.

Once we have the session cleanup and refactoring in https://github.com/dagger/dagger/pull/7315, hopefully this logic can go - but this is a quick fix we can get in for the next release.

### DIFF
--- a/.changes/unreleased/Fixed-20240610-162518.yaml
+++ b/.changes/unreleased/Fixed-20240610-162518.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: 'core: handle secrets in dockerfile builds with syntax directives'
+time: 2024-06-10T16:25:18.328274656+01:00
+custom:
+  Author: jedevc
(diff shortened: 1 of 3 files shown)


marcosnils (approved):
LGTM from the functionality perspective. I'll defer to @sipsma for the buildkit specific parts.

jedevc:
Going to merge this, since we should include this in the release - I'm confident in our tests here, and the critical code path has just been moved, so the behavior should remain the same for the basic cases at least.

PR #7596, opened 2024-06-10
helm: Allow disabling `--oci-max-parallelism num-cpu`

gerhard:
While we are keeping this default so that we don't break existing users, setting this value is something that we want to move away from.

The problem is that this setting limits how many operations can run in parallel. It is still possible for a single operation to max out all available cores. It is also known for a value of `2` to cause deadlocks, i.e. https://github.com/dagger/dagger/issues/6894

For now, we just allow this to be disabled with either `--set engine.args=''` or by explicitly setting this value to an empty list.

This started as https://github.com/dagger/dagger/pull/7395 which turned out to be too big of a change. We since scaled back the initial ambition & are taking a smaller step towards eventually phasing this out.

FWIW, all the Dagger Engines that we run inside the Dagger infra do not use the `--oci-max-parallelism` option.

This also removes the option from tekton-dagger-task docs example.

### DIFF
--- a/helm/dagger/values.yaml
+++ b/helm/dagger/values.yaml
@@ -14,6 +14,12 @@ engine:
   #   [log]
   #     format = "json"
   labels: {}
+  ### Customize Dagger Engine start args
+  # "--oci-max-parallelism num-cpu" is kept as the default behaviour so that we don't surprise users.
+  # You may want to remove it in your deployments. FTR: https://github.com/dagger/dagger/pull/7395
+  args:
+    - "--oci-max-parallelism"
+    - "num-cpu"
   resources:
     limits: {}
     # limits:
--- a/docs/current_docs/integrations/snippets/tekton-dagger-task.yaml
+++ b/docs/current_docs/integrations/snippets/tekton-dagger-task.yaml
@@ -19,9 +19,6 @@ spec:
     - name: dagger-engine
       # modify to use the desired Dagger version
       image: registry.dagger.io/engine:vX.Y.Z 
-      args:
-        - "--oci-max-parallelism"
-        - "num-cpu"
       securityContext:
         privileged: true
         capabilities:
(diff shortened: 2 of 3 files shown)


helm/dagger/values.yaml
```diff
@@ -14,6 +14,12 @@ engine:
   #   [log]
   #     format = "json"
   labels: {}
+  ### Customize Dagger Engine start args
+  # "--oci-max-parallelism num-cpu" is kept as the default behaviour so that we don't surprise users.
+  # You may want to remove it in your deployments. FTR: https://github.com/dagger/dagger/pull/7395
+  args:
+    - "--oci-max-parallelism"
+    - "num-cpu"
```
jedevc, line 22: Only issue, when should we change the default?

Would we want to do a helm minor version bump instead of a patch? IMO, I think that's reasonable to do at any point, as long as we call it out in our release notes.

jedevc (approved):

helm/dagger/values.yaml
jedevc: Yes :tada:

Sounds perfect, let's do that!

jedevc: Could you create a tracking issue with that for the v0.12 milestone? Just so we definitely make sure to get it in.

gerhard:
This failure seems legit to me:
image

The Rust SDK is known be CPU hungry. I would run it `4CPUs` and see if we need more.

jedevc:
@gerhard if you look at the logs, it indicates that the tests are actually complete: https://github.com/dagger/dagger/actions/runs/9449394865/job/26025540600?pr=7596#step:6:1331

But it doesn't exit. I think discussing with @vito, we think it may be a telemetry draining issue.

matipan (approved):
Definitely a good idea to to allow disabling. We already suggest that for many customers running dagger in prod. I agree with Justin, we should remove it as the default in the next release! 🚀

PR #7603, opened 2024-06-10
ci: enable best-effort cache reuse for nested engines

sipsma:
We were previously always putting a random id in the cache volume used for engines created in various CI functions, which meant none of them ever had any local cache. The main reason was that two engines can't use the same local cache at the same time.

We can't lift that limitation, but by using the engine version and private cache mounts we can get *best effort* re-use of these cache volumes, which is a simple enough performance improvement for some cases that we may as well take it.

### DIFF
--- a/ci/engine.go
+++ b/ci/engine.go
@@ -121,8 +121,9 @@ func (e *Engine) Container(
 // Create a test engine service
 func (e *Engine) Service(
 	ctx context.Context,
-	version *VersionInfo,
 	name string,
+	// +optional
+	version *VersionInfo,
 ) (*Service, error) {
 	var cacheVolumeName string
 	if version != nil {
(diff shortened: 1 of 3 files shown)


gerhard:
Very excited to see this change! Trying it out now.

jedevc (approved):

gerhard:
I was not able to run this the second time, aborted after running it for `1h 48m`: https://dagger.cloud/dagger/traces/026e372bfa3b084a409a5d3f80cb4985

image

```sh
dagger-test-all () {
	local parallel="${1:-16}"
	local timeout="${2:-180m}"
	time dagger call --debug --source=".:default" test all --race=true --failfast=true --parallel=$parallel --timeout=$timeout 2>&1 | ts | tee test.all.race.failfast.parallel${parallel}.timeout${timeout}.$(date "+%Y-%m-%d.%H-%M-%S").txt
}
```

Glad that you managed to push it through @jedevc 👍

PR #7605, opened 2024-06-10
OTel log draining follow-ups

vito:
* Standardize on `slog` everywhere that we print output from the engine, instead of "faking" stdout/stderr in places.
  * Goal: `SpanLogger` no longer leaves ambiguity as to whether caller needs to `Close()`. It _never_ does.
  * We will no longer drain to ensure we get every last log, but this was never really an issue, doesn't seem worthwhile atm.
* Rename `telemetry.Logs` to `telemetry.SpanStdio` so its intentions are clearer.
  * Goal: again, remove ambiguity as to whether caller needs to `Close()`. It _always_ does.
* Propagate client-side `NO_COLOR=true` and `--debug` through to the engine so `SpanLogger` can respect it.
  * Goal: now that everything is `slog` we should respect the user's preferences.
  * I needed this to get the `dagger up --random` test to pass without choking on ANSI colors.

### DIFF
--- a/cmd/dagger/licenses.go
+++ b/cmd/dagger/licenses.go
@@ -61,7 +61,7 @@ var licenseFiles = []string{
 }
 
 func findOrCreateLicense(ctx context.Context, dir string) error {
-	_, slog := slog.SpanLogger(ctx, InstrumentationLibrary, slog.LevelWarn)
+	slog := slog.SpanLogger(ctx, InstrumentationLibrary, slog.LevelWarn)
 
 	id := licenseID
 	if id == "" {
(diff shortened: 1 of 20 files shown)


sipsma (approved):
🎉 Tests are a lot faster than current main it seems

I saw the failure in the test workflow:
```
server "8n179cewnjt7l9k0z281pel15" not found
```

I've seen that occasionally recently before this PR, so don't think it's related. I'm tempted to just say that we should just assume https://github.com/dagger/dagger/pull/7315 fixes it since that PR completely reworks how all of that works.

PR #7608, opened 2024-06-11
engine: fix context error wrapping in executor

sipsma:
When an exec fails due to context canceled, we were dropping the cancelation error and just returning the exec error. I believe this got fat-fingered while updating the code from upstream to use fmt.Errorf instead of github.com/pkg/errors.

It's not clear if this is really meaningful, but noticed it changed while attempting to debug occurances of "exit code 137" in our CI, which plausibly could be related to all of this. It's possible there's calling code that tries to handle cancelation with errors.Is, which this would have broken. This is purely hypothetical at this point though.

---

This is the related upstream version of this code, if helpful https://github.com/moby/buildkit/blob/715276d7423f006dab1c96bd6910d4ca55faa230/executor/runcexecutor/executor.go#L374

### DIFF
--- a/engine/buildkit/executor.go
+++ b/engine/buildkit/executor.go
@@ -414,7 +414,7 @@ func exitError(ctx context.Context, exitCodePath string, err error) error {
 
 		select {
 		case <-ctx.Done():
-			exitErr.Err = fmt.Errorf(exitErr.Error())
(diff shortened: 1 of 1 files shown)


vito (approved):

sipsma:
Hit the same thing I hit in my other PR: https://github.com/dagger/dagger/pull/7315#issuecomment-2157837642

So I guess it is a pre-existing issue. I'll see if I can fix it before the release tomorrow

Issue #6957, opened 2024-03-27
CLI: support passing a remote URL as file argument


shykes:
## Problem

When calling a Dagger Function from the CLI, I can pass the path of a local file as an argument of type `File`, but not the URL of a remote file. This would be very useful.

## Solution

Add support for remote URLs when passing arguments of type `File` from the CLI.

helderco:
This has already been done in:
- https://github.com/dagger/dagger/pull/6769

shykes:
No, I think this is different. #6769 allows passing a file from a git repo - which is indeed very useful. But I want to pass a regular http url to a regular file. And that unfortunately does not seem to work.

helderco:
Oh, sorry! I actually read URL as Git! 😅 

That's a good idea, and easy to add 👍

helderco:
Hmm... both `http://` and `https://` are valid git schemes. How would you know if you want `dag.Git` or `dag.Http`?

helderco:
Some tools prefix the scheme with `git+`, like `git+https://github.com/...` ([example](https://hatch.pypa.io/latest/config/dependency/#supported-vcs)).

jedevc:
FYI, this is related to https://github.com/dagger/dagger/pull/7609 as well, since it also has different schemes.

helderco:
@guillaume, can you own this? It's easy to do but requires a decision on the scheme and consistency with the other changes you've been making.

grouville:
> @grouville, can you own this? It's easy to do but requires a decision on the scheme and consistency with the other changes you've been making.

Yes sure 🙏 👌

helderco:
Update: this is blocked on standardizing URI references (git, http, other vcs…) in:

* [https://github.com/dagger/dagger/pull/7708](https://github.com/dagger/dagger/pull/7708)

PR #7609, opened 2024-06-11
feat(core): apply new ref scheme for directories and files

grouville:
Follow-up of https://github.com/dagger/dagger/pull/7511 (rebased on top of it)

This applies the new ref scheme to directories and files (moving out from the buildkit syntax), whilst remaining retrocompatible

#### Implementation

This does not change the cmd/flags logic as we only convert the new ref scheme to buildkit's ref scheme whilst remaining retrocompatible

As part of the implementation, we move the parseRefString implementation to the vcs package, to avoid dependency hell issue between the core and the CLI.

We introduce a ParseRef Func type, aimed to be used by ConvertToBuildKitRef to handle the two specific use-cases:
parseRefStringDir and parseRefStringFile

We rename parseRefString to parseRefStringDir, as it was intended for directory ref parsing, and we duplicate its logic for files. The only difference is the stat type: parseRefStringFile makes sure that the file exists locally as part of the implicit local file interpolation


engine/vcs/convert_test.go
```diff
@@ -0,0 +1,274 @@
...
+		{
+			name:           "Git ref with version",
+			urlStr:         "git@github.com:sipsma/daggerverse@v0.9.1",
+			expectedUrlStr: "ssh://git@github.com:sipsma/daggerverse#v0.9.1",
+		},
+		{
+			name:           "Git ref with .git and version",
+			urlStr:         "git@github.com:sipsma/daggerverse.git@v0.9.1",
+			expectedUrlStr: "ssh://git@github.com:sipsma/daggerverse.git#v0.9.1",
+		},
```
jedevc, line 82: Hm, this shouldn't be a valid syntax.

The supported syntaxes should either be:
- The older buildkit ref style
- The newer module ref style

I don't think this is either of the above?

grouville: Ok 👌 
I wanted to cover all the test cases of the current TestParseGit: https://github.com/dagger/dagger/blob/main/cmd/dagger/module_test.go#L148

jedevc: Oh agreed, it's just that this specific case isn't in there.

grouville:
Failure seems unrelated to implementation

gerhard:
Want to rebase @grouville as soon as we have upgraded our CI to `v0.11.8`? Pretty sure there will be a PR from @jedevc for this later today.

grouville:
> Want to rebase @grouville as soon as we have upgraded our CI to `v0.11.8`? Pretty sure there will be a PR from @jedevc for this later today.

I think we should review / rebase it once #7708 gets merged as it will impose the ref style for private git repo

jedevc:
x-linking https://github.com/dagger/dagger/issues/6957 - not entirely sure how these should interact.

cmd/dagger/flags.go
```diff
@@ -218,25 +221,29 @@ func (v *directoryValue) Get(ctx context.Context, dag *dagger.Client, modSrc *da
...
-			KeepGitDir: true,
-		}
-		if authSock, ok := os.LookupEnv("SSH_AUTH_SOCK"); ok {
-			gitOpts.SSHAuthSocket = dag.Host().UnixSocket(authSock)
-		}
-		gitDir := dag.Git(parsedGit.Remote, gitOpts).Branch(parsedGit.Fragment.Ref).Tree()
-		if subdir := parsedGit.Fragment.Subdir; subdir != "" {
-			gitDir = gitDir.Directory(subdir)
+	bk := &osBuildkitClient{}
+	ref, kind := vcs.ConvertToBuildKitRef(ctx, v.String(), bk, vcs.ParseRefStringDir)
```
jedevc, line 225: We should be able to remove this function. Instead, we have `parseGit` call `vcs.ParseRef`, and then we can extract host, subpath, etc from there :tada:

We should avoid the dance of converting a parsed string into a raw string to parse it again - that's going to more error prone and difficult to follow :scream:

PR #7617, opened 2024-06-11
fix: remove bashisms in install.sh

jedevc:
Reported in discord: https://discord.com/channels/707636530424053791/1250124959406227547/1250124959406227547

Introduced in #7544.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -267,9 +267,9 @@ latest_version() {
 base_url() {
   os="$(uname_os)"
   arch="$(uname_arch)"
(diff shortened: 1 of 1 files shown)


gerhard (approved):
Good save! I should have caught it while reviewing:
- https://github.com/dagger/dagger/pull/7544

The fix is now deployed, even before this PR gets merged, so it should be all good!

The truth is that this will work as intended in all Ubuntu installations - as it did for me when checking - even though it's indeed surprising behaviour:

image

All good 👍

Issue #6441, opened 2024-01-17
✨ Windows install: winget package creation


nunix:
### What are you trying to do?

Windows has a "package manager" called [winget](https://github.com/microsoft/winget-cli) and the community can contribute to add new packages to the [official repository](https://github.com/microsoft/winget-pkgs).

This will be an alternative to the current script installation. I created the initial package and the PR can be found here: https://github.com/microsoft/winget-pkgs/pull/135188

There's an automated workflow that can be created in order to update the repo whenever Dagger as new releases (see: https://github.com/microsoft/winget-create#using-windows-package-manager-manifest-creator-in-a-cicd-pipeline for examples).

Finally, I'll create a docs PR to show how to use it once the winget PR is approved.

PS: please note that winget **does not** rebuild packages, it takes the current Github releases and install them.

### Why is this important to you?

Winget is getting more popular and becoming the "standard" way of installing software packages on Windows.
Having Dagger on the Winget repo might help reach more Windows users and also be part of machines provisioning (i.e. developer environment setup from scratch).

### How are you currently working around this?

The current powershell script works fine, this is an alternative solution to install Dagger on Windows.

marcosnils:
> There's an automated workflow that can be created in order to update the repo whenever Dagger as new releases (see: https://github.com/microsoft/winget-create#using-windows-package-manager-manifest-creator-in-a-cicd-pipeline for examples).

Great idea for a Dagger module? Would you like to give it a try?

nunix:
As quickly discussed offline, yes I'll try to do it and make it the "story" of my next blog on wsl.dev about Dagger.

pjmagee:
This has already been published and is usable. But it's published under the publisher name of Dagger.

pjmagee:
https://github.com/microsoft/winget-pkgs/tree/master/manifests/d/Dagger

pjmagee:
I've submitted 0.11.6 to WinGet https://github.com/microsoft/winget-pkgs/pull/157191

jedevc:
https://github.com/dagger/dagger/pull/10073 added this, so will keep the issue open until I've confirmed that v0.18.3 release works properly with winget :tada:

pjmagee:
> > There's an automated workflow that can be created in order to update the repo whenever Dagger as new releases (see: https://github.com/microsoft/winget-create#using-windows-package-manager-manifest-creator-in-a-cicd-pipeline for examples).
> 
[...]

PR #7620, opened 2024-06-11
feat: Adding winget to goreleaser

pjmagee:
I strongly believe this would be a perfect replacement for official releases of Dagger on Windows, making the `install.ps1` a secondary option (which has the ability for Git Commit installs, or for anyone who is not having WinGet, - but WinGet is the official package manager for windows, and is basically apt-get for Windows

Attempting to solve #6441 

This seems like a more Dagger supported method for installation because:

- Its supported by existing goreleaser
- It handles the checksum
- winget allows the portable zip binary 
- It can also handle adding the install to the users PATH via an alias 
- Supports simple uninstall and upgrades of the Cli

I have used the existing package Id that was already found by someone who published the 0.9.6 to Winget, since that time, I have published later versions but it also is a common pattern with winget, Cli tools are having this `suffix`. I have also added tags which is a supported winget feature

**Note**: These versions in the screenshot did not come from goreleaser, they're previous releases from me and 1 other contributor

image

Not sure how we can test the goreleaser, I'm not too familiar with that tool 

Previous releases look like this, when using the winget tool locally to publish:

`wingetcreate update --submit --urls https://dl.dagger.io/dagger/releases/0.11.7/dagger_v0.11.7_windows_amd64.zip https://dl.dagger.io/dagger/releases/0.11.7/dagger_v0.11.7_windows_arm64.zip https://dl.dagger.io/dagger/releases/0.11.7/dagger_v0.11.7_windows_armv7.zip --version 0.11.7 Dagger.Cli`

https://github.com/microsoft/winget-pkgs/pulls?q=Dagger.Cli

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,3 +1,5 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/goreleaser/goreleaser/v2.2.0/www/docs/static/schema-pro.json
+
 version: 2
 
 includes:
@@ -52,15 +54,21 @@ nix:
 # https://goreleaser.com/customization/winget/
 winget:
   - name: dagger
+    repository:
+      owner: "{{ .Env.GH_ORG_NAME }}"
+      name: winget
+    commit_author:
+      name: dagger-bot
+      email: noreply@dagger.io
+    url_template: "https://{{ .Env.ARTEFACTS_FQDN }}/dagger/releases/{{ .Version }}/{{ .ArtifactName }}"
+    homepage: "https://dagger.io"
     publisher: Dagger
     short_description: "Dagger is an integrated platform to orchestrate the delivery of applications"
     license: "asl20"
-    homepage: "https://dagger.io"
     publisher_url: "https://dagger.io"
     publisher_support_url: https://github.com/dagger/dagger/issues/new/choose
     package_identifier: Dagger.Cli
-    use: archive
-    url_template: "https://{{ .Env.ARTEFACTS_FQDN }}/dagger/releases/{{ .Version }}/{{ .ArtifactName }}"
+    use: archive    
     skip_upload: false
     tags:
       - dagger
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -49,6 +49,24 @@ nix:
         --fish <($out/bin/dagger completion fish) \
         --zsh <($out/bin/dagger completion zsh)
 
+# https://goreleaser.com/customization/winget/
+winget:
+  - name: dagger
+    publisher: Dagger
+    short_description: "Dagger is an integrated platform to orchestrate the delivery of applications"
+    license: "asl20"
+    homepage: "https://dagger.io"
+    publisher_url: "https://dagger.io"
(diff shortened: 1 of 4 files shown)


.goreleaser.yml
```diff
@@ -1,3 +1,5 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/goreleaser/goreleaser/v1.26.0/www/docs/static/schema-pro.json
```
pjmagee, line 1: this is so VSCode shows the correct schema validation, based on the version of go releaser in Dagger, which I found to be 1.26

pjmagee:
@jedevc  It would be awesome if we could update Docs with `WinGet` and the new Powershell in the same release? But im not sure whom can help validate the goreleaser stuff with me?

jedevc:
Hey @pjmagee sorry for the delay in getting to this! I think we're interested in taking this, it's a much easier way to manage for windows users. cc @jpadams 

`winget` currently seems to be publishing to https://github.com/dagger/winget - that seems similar to how we already have nix, and brew tap, etc. However, I'm assuming we want to PR it to `winget-pkgs` (there's an example of how that might be done in https://blog.goreleaser.com/announcing-goreleaser-v1-19-the-big-release-b01565c72658 I think)?

If we don't set up to publish to winget-pkgs, what do the installation steps look like, if we have our own manifest repository?

pjmagee:
> Hey @pjmagee sorry for the delay in getting to this! I think we're interested in taking this, it's a much easier way to manage for windows users. cc @jpadams
> 
> `winget` currently seems to be publishing to https://github.com/dagger/winget - that seems similar to how we already have nix, and brew tap, etc. However, I'm assuming we want to PR it to `winget-pkgs` (there's an example of how that might be done in https://blog.goreleaser.com/announcing-goreleaser-v1-19-the-big-release-b01565c72658 I think)?
> 
> If we don't set up to publish to winget-pkgs, what do the installation steps look like, if we have our own manifest repository?

Ive been continuing to publish to winget myself, following on from a user who published a version to winget a long time ago, but because it wasnt being maintained, it had a very stale version of dagger.

pjmagee:
@jedevc  I rebased since some changes happened due to goreleraser 2.2, I can only think of 1 benefit of dagger having its own source repository, instant or immediate package updates. But it also means it will not be 'discoverable' by default, users will need to add a custom repository (a bit like helm I suppose)

Then adding the winget packages there, instead of waiting for Microsoft/Community members to approve them. So then in this case, the user would do something like:


```
winget source add -n dagger -a https://github.com/dagger/winget-releases/ -t winget
winget search dagger.cli
```

Example output would be like this:

```
winget search dagger.cli
Name       Id         Version Source
-------------------------------------
Dagger CLI Dagger.Cli 0.12.4  winget (old ones ive been publishing)
Dagger CLI Dagger.Cli 0.12.5  dagger (new ones post merge would show up here)
```

I suppose this can speed up releases being pushed to winget. Requires a bit more effort though.

Would also need to update the docs to include adding the winget source. This PR right now just assumes using the winget-pkg repository maintained by Microsoft at https://github.com/microsoft/winget-pkgs/tree/master/manifests/d/Dagger/Cli

Issue #7630, opened 2024-06-12
🐞 dagger.io/dagger v0.11.7 is not available through dagger.io


kjuulh:
### What is the issue?

I think your proxy was left out in the cold when you pushed the latest release to github

### Dagger version

dagger v0.11.7

### Steps to reproduce

go get dagger.io/dagger@v0.11.7 => dagger.io/dagger v0.11.6

### Log output

```
❯ go get dagger.io/dagger@v0.11.7
go: downloading dagger.io/dagger v0.11.7
go: dagger.io/dagger@v0.11.7 requires dagger.io/dagger@v0.11.7: parsing go.mod:
        module declares its path as: github.com/dagger/dagger
                but was required as: dagger.io/dagger
```

_No response_

morlay:
https://proxy.golang.org/dagger.io/dagger/@v/v0.11.7.mod
https://proxy.golang.org/github.com/dagger/dagger/@v/v0.11.7.mod

seems `github.com/dagger/dagger` is released as `dagger.io/dagger`

jedevc:
Investigating now.

Looks like https://github.com/dagger/dagger-go-sdk, the source of truth for this is wrong.

kjuulh:
@jedevc it still seems broken:

```
❯ go get dagger.io/dagger@v0.11.7
go: dagger.io/dagger@v0.11.7 requires dagger.io/dagger@v0.11.7: parsing go.mod:
        module declares its path as: github.com/dagger/dagger
                but was required as: dagger.io/dagger
```

jedevc:
Yeah, sorry we still need to repush the v0.11.7 tag - @gerhard is looking into it.

sagikazarmark:
Maybe tag a new version instead? Go is quite sensitive to changing tags.

gerhard:
The Go SDK has now been re-tagged: https://github.com/dagger/dagger-go-sdk/releases/tag/v0.11.7

This is the expected version & directory structure: https://github.com/dagger/dagger-go-sdk/tree/443305dc487d22254940e3029bd42d1b45de2dcc

<img width="1552" alt="image" src="https://github.com/dagger/dagger/assets/3342/1d631da7-abc0-4ff4-a323-be8d3841772c">

And this is the expected `go.sum`: https://github.com/dagger/dagger-go-sdk/blob/443305dc487d22254940e3029bd42d1b45de2dcc/go.mod

With all that in place, I am still getting the following error - as captured in this issue:

```console
[...]

PR #7631, opened 2024-06-12
ci: fix git publish (for go and php)

jedevc:
Fixes https://github.com/dagger/dagger/issues/7630.

Introduced in [`8421636` (#7554)](https://github.com/dagger/dagger/pull/7554/commits/8421636acaaa7cd312eb3546f6b3b1870090d83c).

This caused the https://github.com/dagger/dagger-go-sdk read-only copy to include the full dagger/dagger repo, instead of the filtered one.

### DIFF
--- a/ci/sdk_go.go
+++ b/ci/sdk_go.go
@@ -174,7 +174,7 @@ func gitPublish(ctx context.Context, opts gitPublishOpts) error {
 			"push",
 			"-f",
 			opts.dest,
-			fmt.Sprintf("%s:%s", opts.sourceTag, opts.destTag),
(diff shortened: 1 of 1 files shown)


jedevc:
Just merging this should fix `main`. Not sure how we should go about fixing the tags - we could retag the go/php v0.11.7 tags onto this once it's merged, no changes to those SDKs have occurred, so it should be safe to do.

gerhard (approved):
I noticed the side-effect of this yesterday - same workflows running in those repositories - but I didn't have time to investigate further. This will most definitely fix that issue.

Good to merge, will force update the tag afterwards 👍

PR #7632, opened 2024-06-12
ci: only enable sdk dev engine checks on main

jedevc:
Based on https://github.com/dagger/dagger/pull/7483

Rework of https://github.com/dagger/dagger/pull/7628, that doesn't entirely disable these checks.

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -25,15 +25,30 @@ jobs:
     uses: ./.github/workflows/_sdk_check.yml
     with:
       sdk: go
+  go-dev:
+    if: github.ref_name == 'main'
+    uses: ./.github/workflows/_sdk_check.yml
+    with:
+      sdk: go
       dev-engine: true
 
   python:
     uses: ./.github/workflows/_sdk_check.yml
     with:
       sdk: python
+  python-dev:
+    if: github.ref_name == 'main'
+    uses: ./.github/workflows/_sdk_check.yml
+    with:
+      sdk: python
       dev-engine: true
 
   typescript:
+    uses: ./.github/workflows/_sdk_check.yml
+    with:
+      sdk: typescript
+  typescript-dev:
+    if: github.ref_name == 'main'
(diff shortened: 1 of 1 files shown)


.github/workflows/sdk.yml
```diff
@@ -22,36 +22,66 @@ concurrency:
 
 jobs:
   go:
+    uses: ./.github/workflows/_sdk_check.yml
+    with:
+      sdk: go
+  go-dev:
```
matipan, line 28: When running on main, do we still want to run both `<sdk>-dev` and `<sdk>` jobs?

gerhard:
I would flip the condition & run `dev` in PRs, and stable on `main`. Since PRs are less stable than `main`, I would think that testing `dev` is more important there. As for stable, that seems to fit more in `main` since it's checking for backwards compatibility.

The more I think about this, the more I believe that we should run against both `dev` and `stable` all the time. The mental model that I have is:
- did any of the changes in this PR break anything (a.k.a. `dev`) ?
- are the changes in this PR compatible with latest release (a.k.a. `stable`) ?

Since new PRs could be merged in `main` while this PR is open, we want to run the same there since we want to be confident that the combination & ordering of PRs still results in green pipelines.

jedevc:
> I would flip the condition & run `dev` in PRs, and stable on `main`. Since PRs are less stable than `main`, I would think that testing `dev` is more important there. As for stable, that seems to fit more in `main` since it's checking for backwards compatibility.

SGTM - do you want to handle rebasing / switching the check? I can also do it.

jedevc:
I would still like to take this as-is - we should be testing against both dev and non-dev for these jobs, which aren't currently doing at all (we're only using dev).

I think we should rework the priority of how we test `dev` at all, moving towards testing everything on `dev` but still keeping around a few more dedicated compatibility checks - but I think that's out of scope for here.

gerhard:
Looking at this now.

gerhard (approved):
👍 🚀

PR #7644, opened 2024-06-12
Use dynamically sized dind runners for SDK and testdev

matipan:
As was discussed [here](https://github.com/dagger/dagger/pull/7483#discussion_r1636854755), since [#7628](https://github.com/dagger/dagger/pull/7628) got merged we are using dind runners for SDK jobs. Default dind runners have 32c and 16Gi, which for SDK jobs is excessive. Since we now have support for dynamically sized dind runners, we migrate SDK jobs to 4c (like we used to run it) and we leave testdev with 32c jobs.

### DIFF
--- a/ci/mage/engine.go
+++ b/ci/mage/engine.go
@@ -16,10 +16,29 @@ import (
 	"github.com/dagger/dagger/engine/distconsts"
 )
 
+var (
+	OutputDir           = ""
+	EngineContainerName = "dagger-engine.dev"
+)
+
+func init() {
+	if v, ok := os.LookupEnv(util.DevContainerEnvName); ok {
+		EngineContainerName = v
+	}
+	if v, ok := os.LookupEnv(util.DevOutputEnvName); ok {
+		OutputDir = v
+	}
+}
+
 type Engine mg.Namespace
 
 // Dev builds and starts an Engine & CLI from local source code
 func (t Engine) Dev(ctx context.Context) error {
+	binDir := OutputDir
+	if binDir == "" {
+		binDir = filepath.Join(os.Getenv("DAGGER_SRC_ROOT"), "bin")
+	}
+
 	gpuSupport := os.Getenv(util.GPUSupportEnvName) != ""
 	trace := os.Getenv(util.TraceEnvName) != ""
 	race := os.Getenv(util.RaceEnvName) != ""
@@ -34,7 +53,7 @@ func (t Engine) Dev(ctx context.Context) error {
 	if race {
 		args = append(args, "with-race")
 	}
-	tarPath := "./bin/engine.tar"
+	tarPath := filepath.Join(binDir, "engine.tar")
 	args = append(args, "container", "export", "--path="+tarPath)
 	args = append(args, "--forced-compression=Gzip") // use gzip to avoid incompatibility w/ older docker versions
 	err := util.DaggerCall(ctx, args...)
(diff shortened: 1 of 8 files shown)


jedevc (approved):

jedevc:
The job failures were due to jobs being scheduled on the same node and attempting to use the same container names/etc with a shared docker socket - not sure why we never observed this before.

However, it's pretty simple to resolve - we just need to have `./hack/dev` allow taking parameters to customize the output location and the container name. Then in github, we parameterize this by the github unique run id + number + attempt, so we always get a different result for each job :tada:

ci/mage/engine.go
```diff
@@ -116,6 +135,27 @@ func (t Engine) Dev(ctx context.Context) error {
...
-	fmt.Println("export _EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-container://" + util.EngineContainerName)
+	fmt.Println("export _EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-container://" + containerName)
+	fmt.Println("export _DAGGER_TESTS_ENGINE_TAR=" + filepath.Join(binDir, "engine.tar"))
+	fmt.Println("export PATH=" + binDir + ":$PATH")
+
 	return nil
 }
+
+// Get environment variable updates for running dagger
+func (t Engine) DevEnv(ctx context.Context) {
```
jedevc, line 146: This now lets us move a ton of the env var logic all into one place in this file, instead of needing to have loads of it in `./hack/dev` and `./hack/with-dev`.

jedevc:
Hm, now there are 137 errors - looks like `dial-stdio` is getting SIGKILLed somehow, not quite sure why that happens.

jedevc:
Also see https://github.com/dagger/dagger/pull/7644 for a related refactor, that aims to run only the sdk dev checks on `main`.

jedevc (changes requested):
See above - I think the instances are too small to support running multiple dagger engines all on one machine.

jedevc:
Another option would be to switch to the sidecar model for doing dind, instead of using a daemon set.

matipan:
@jedevc it seems that bumping up resources to 8c per SDK did the trick. Pipelines succeeded and they took roughly the same time as in main.

The priority is to stop increasing CI cost so we will go ahead and merge this one. However, I would like to understand a bit better your second suggestion: dind sidecar. How would using a dedicated dind container per job improve? And could we do that and use 4c like before?

jedevc (approved):

jedevc:
Yeah, bumping resources for now is good - that's fine to bring costs down in the meantime.

So if we could have a dind sidecar, we wouldn't be sharing a docker daemon between multiple pods being scheduled - instead, each runner pod would have it's own docker daemon. Assuming that cgroup limits were set appropriately, etc, I think we'd be able to go back down to 4c.

I think we already have the sidecar for the dagger daemon? Or is that still shared between all the jobs on a single worker?

matipan:
> So if we could have a dind sidecar, we wouldn't be sharing a docker daemon between multiple pods being scheduled - instead, each runner pod would have it's own docker daemon. Assuming that cgroup limits were set appropriately, etc, I think we'd be able to go back down to 4c.
I can give it a go! The main thing to figure out is with the filesystem mount we have on the host. But it should be too complicated

> I think we already have the sidecar for the dagger daemon? Or is that still shared between all the jobs on a single worker?
We do not. We run a single dagger engine per host and multiple runners within that host

jedevc:
> We do not. We run a single dagger engine per host and multiple runners within that host

Oh in that case, I don't think we should special case the `dind` case either. Ignore my suggestion!

PR #3486, opened 2022-10-24
Prepare to publish dagger-cue artefacts

gerhard:
This PR is:

- part of https://github.com/dagger/dagger/issues/3471
- supersedes https://github.com/dagger/dagger/pull/3483
- partial follow-up to https://github.com/dagger/dagger/pull/3356

It changes the `release-v0.2.x.yml` workflow to publish `dagger-cue` binaries.

Linux `install.sh` & Windows `install.ps1` have been updated too. The Linux variant has been tested.

**Help wanted for the Windows one**.

This was tested via https://github.com/gerhard/dagger/actions/runs/3316263177/jobs/5477850676 & it published:

### 1/2. AWS S3 files

image

Tested with `./install.sh`

image

### 2/2. Homebrew tap

image

Tested with `brew install gerhard/tap/dagger-cue`

image

---

## Next steps

- [x] review & merge https://github.com/dagger/dagger/pull/3506
- [x] review & merge https://github.com/dagger/dagger/pull/3505
- [ ] review & merge this PR
- [ ] publish an official `dagger-cue`
- [ ] @vikram-dagger to update Dagger CUE SDK documentation (it will also ensure that it works as expected)

---

FWIW we ended up upgrading to GoReleaser Pro with @aluzzardi so that we get the monorepo support. It has a bunch of other useful features too that we might want to look into https://beckersoft.gumroad.com/l/goreleaser. The license key has been configured in GitHub Actions. The source of truth is the shared 1Password vault.

PR #7648, opened 2024-06-13
ci: fix git publish tag confusion

jedevc:
Follow up to #7631

This is fairly fiddly to do correctly. Essentially, we need to be able to handle at least these inputs:
- Branch names, like `main`
- Tag names, like `v0.11.7`
- Full refs, like `refs/pull/<pr>/merge`

We can't simply use `git fetch <ref>`, since that doesn't create a local ref copy. We shouldn't use a refspec that mirrors to a single name, since that strips `refs/tag/...` which means tags are lost.

The correct way to do this is using a `git fetch` refspec, to mirror all remote refs that match the target name to a local ref. Then this satisfies all the above inputs correctly.

Additionally, for dry-runs, I've added a `rev-parse` sanity check, so can check the results more easily.

### DIFF
--- a/ci/sdk_go.go
+++ b/ci/sdk_go.go
@@ -155,18 +155,17 @@ func gitPublish(ctx context.Context, opts gitPublishOpts) error {
 			WithSecretVariable("GIT_CONFIG_VALUE_0", dag.SetSecret("GITHUB_HEADER", fmt.Sprintf("AUTHORIZATION: Basic %s", encodedPAT)))
 	}
 
-	tmpBranch := "internal/publish"
 	result := git.
 		WithEnvVariable("CACHEBUSTER", identity.NewID()).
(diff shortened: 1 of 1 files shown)


gerhard (approved):
This looks right to me.

Is there a way to test this before merging? I used to do this in the past to double-check that everything works as intended:
- https://github.com/dagger/dagger/pull/3486

It's been a bumpy few days & I am just more cautious than usual.

jedevc:
It can be tested by providing custom parameters to this dagger function when using `dagger call`: https://github.com/dagger/dagger/blob/6e5396e521c8a861cdd063f99009987138af7648/ci/sdk_go.go#L74-L85

I've also added the `git rev-parse` step so you can manually check the type of tag/branch/etc by getting the full ref name. But that's not a full end-to-end test.

PR #7649, opened 2024-06-13
docs: Add CommonRoom pixel

vikram-dagger:


vikram-dagger:
See comment in internal tracker before merging

PR #7651, opened 2024-06-13
Fix invalid dagger develop command

egladman:
The command is invalid, and I suspect it's missing a space.

```
❯ dagger develop...
Error: unknown command "develop..." for "dagger"
```


jedevc:
Hey @egladman - so I think the `...` is used as a placeholder in these docs. It's not actually intended to be run in the terminal.

However, `dagger develop` works fine as standalone, so no `...` should be required here at all - if you want to update your PR to remove that, then it should be good to merge :tada:

Thanks for the contribution, sorry it took so long to get to reviewing it :tada:

egladman:
@jedevc done, and no worries. I hadn't seen that convention before. I'll typically seen something like: `foo bar [options...]`

jedevc (approved):

PR #7655, opened 2024-06-13
chore: remove progrock

jedevc:
@vito has said we should do this a couple times, so I just yanked it out (look at all those deps gone!).

We just need to pull in a bubbletea helper, and also bring in all the defined symbols that we use for rendering.


vito (approved):
nice, thanks!

dagql/idtui/symbols.go
```diff
@@ -0,0 +1,35 @@
+package idtui
+
+const (
+	Block               = "█"
+	CaretDownEmpty      = "▽"
+	CaretDownFilled     = "▼"
+	CaretLeftFilled     = "◀" //"<"
+	CaretRightEmpty     = "▷" //">"
+	CaretRightFilled    = "▶" //">"
```
gerhard, line 9: There is a tiny `golint` issue here: https://github.com/dagger/dagger/actions/runs/9514121731/job/26225628274#step:3:441

image

jedevc: Already have an open PR: https://github.com/dagger/dagger/pull/7666

PR #7658, opened 2024-06-13
feat: improve license generation

### LINKED ISSUE #6666: Create LICENSE on `dagger develop`, not `dagger init`
## Problem

When I run `dagger init`, only 2 files are created: `dagger.json`, and `LICENSE`. Since there is no code, creating a LICENSE doesn't make sense.

Conversely, when I call `dagger develop`, no LICENSE file is created, even though new code is generated.

## Solution

Move the LICENSE file creation to `dagger develop` instead of `dagger init`.

TomChv:
This changes the  `dagger init` command to  not generate `LICENSE` if no code has been generated (when `--sdk`) is not specified.

It also generates a license if then `dagger develop` is executed with `--sdk` flag.

Fixes #6666
Close DEV-3989

### DIFF
--- a/.changes/unreleased/Added-20240613-221441.yaml
+++ b/.changes/unreleased/Added-20240613-221441.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: Generate license only if code has been generated
+time: 2024-06-13T22:14:41.883445+02:00
+custom:
+  Author: TomChv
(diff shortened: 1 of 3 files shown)


vito (approved):
:tada:

jedevc (approved):

PR #7661, opened 2024-06-14
fix(sdk/elixir): fix runtime image cannot run on arm64

wingyplus:
I found out that Elixir runtime cannot run on my Apple M1 machine and realize that the image in Elixir runtime has only amd64 architecture. :(
 
So I do fix by bump image to a new version that have both amd64 and aarch64 architectures.


sipsma (approved):
LGTM! The CI failures you're hitting should be fixed with a rebase on main

wingyplus:
@sipsma rebased 🙏

PR #7663, opened 2024-06-14
chore/fix(core): add integration tests to other VCS

grouville:
This extends the integration tests to other VCS.

To do so, it relies on a mirror of dagger-module-tests:
- https://gitlab.com/dagger-modules/test/more/dagger-test-modules-public (auto-sync to-be-done)
- https://bitbucket.org/dagger-modules/dagger-test-modules-public/
(more could be added over time, but most use provider behavior is contained in GitHub, GitLab and Bitbucket)

Current implementation extends the test suite, fixes a few inconsistencies around the VCS URL (URL should almost never contain the .git if present in the ref, as providers do not like them).

It also contains a naive change `HTMLURL()` to make it compatible with all VCS. I wonder if we should not add an enum field representing the VCS inside the `gitSourceModule` type, what do you all think ? 

Now, all integration test needed to run against GitHub should be run against vcsTests, with a helper function provided for ease of testing: [testOnMultipleVCS](https://github.com/dagger/dagger/pull/7663/files#diff-2f09d96afa5c2fcd508142d24e636c3d7555c89f8e06bfafaecc0c0f1c94e83cR1250)

⚠️ This increases the integration test suite time linearly (with cache though)

TODO (guillaume), but non blocking:
- [ ] Add automatic syncing (followup PR on both repo, after private repo support is done)

### DIFF
--- a/core/integration/module_config_test.go
+++ b/core/integration/module_config_test.go
@@ -127,9 +127,11 @@ func (ModuleSuite) TestConfigs(ctx context.Context, t *testctx.T) {
 				require.ErrorContains(t, err, `source path "/tmp" contains parent directory components`)
 			})
 
-			t.Run("git", func(ctx context.Context, t *testctx.T) {
-				_, err := base.With(daggerCallAt(testGitModuleRef("invalid/bad-source"), "container-echo", "--string-arg", "plz fail")).Sync(ctx)
-				require.ErrorContains(t, err, `source path "../../../" contains parent directory components`)
+			testOnMultipleVCS(t, func(ctx context.Context, t *testctx.T, tc vcsTestCase) {
+				t.Run("git", func(ctx context.Context, t *testctx.T) {
+					_, err := base.With(daggerCallAt(testGitModuleRef(tc, "invalid/bad-source"), "container-echo", "--string-arg", "plz fail")).Sync(ctx)
+					require.ErrorContains(t, err, `source path "../../../" contains parent directory components`)
+				})
 			})
 		})
 
@@ -214,9 +216,11 @@ func (ModuleSuite) TestConfigs(ctx context.Context, t *testctx.T) {
 				require.ErrorContains(t, err, `module dep source root path "../dep" escapes root`)
 			})
 
-			t.Run("git", func(ctx context.Context, t *testctx.T) {
-				_, err := base.With(daggerCallAt(testGitModuleRef("invalid/bad-dep"), "container-echo", "--string-arg", "plz fail")).Sync(ctx)
-				require.ErrorContains(t, err, `module dep source root path "../../../foo" escapes root`)
+			testOnMultipleVCS(t, func(ctx context.Context, t *testctx.T, tc vcsTestCase) {
+				t.Run("git", func(ctx context.Context, t *testctx.T) {
+					_, err := base.With(daggerCallAt(testGitModuleRef(tc, "invalid/bad-dep"), "container-echo", "--string-arg", "plz fail")).Sync(ctx)
+					require.ErrorContains(t, err, `module dep source root path "../../../foo" escapes root`)
+				})
 			})
 		})
 	})
@@ -670,13 +674,15 @@ func (ModuleSuite) TestDaggerDevelop(ctx context.Context, t *testctx.T) {
 		require.Contains(t, ents, "main.go")
 	})
 
-	t.Run("fails on git", func(ctx context.Context, t *testctx.T) {
-		c := connect(ctx, t)
+	testOnMultipleVCS(t, func(ctx context.Context, t *testctx.T, tc vcsTestCase) {
+		t.Run("fails on git", func(ctx context.Context, t *testctx.T) {
+			c := connect(ctx, t)
 
-		_, err := goGitBase(t, c).
-			With(daggerExec("develop", "-m", testGitModuleRef("top-level"))).
-			Sync(ctx)
-		require.ErrorContains(t, err, `module must be local`)
+			_, err := goGitBase(t, c).
+				With(daggerExec("develop", "-m", testGitModuleRef(tc, "top-level"))).
+				Sync(ctx)
+			require.ErrorContains(t, err, `module must be local`)
+		})
 	})
 
 	t.Run("source is made rel to source root by engine with absolute path", func(ctx context.Context, t *testctx.T) {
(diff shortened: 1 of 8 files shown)


grouville:
> cc @jedevc, I realized at the same time that the current HTMLURL() implementation does not handle bitbucket's URL format.

We currently build it as such: `u := "https://" + src.Root + "/tree/" + src.Commit`, but a real bitbucket example would be: `https://bitbucket.org/test-travail/test/commits/b17a8e871052f7d14dbb198b7071986f997ab7a8`.

> I still wonder what is the best implementation: shall we store the corresponding VCS as part of the gitmodulesource ? As the HTMLURL seems resolved from the gitsource, I don't think we will have access to this information ?

engine/vcs/vcs.go
```diff
@@ -499,8 +499,11 @@ func RepoRootForImportDynamic(importPath string, verbose bool) (*RepoRoot, error
 		return nil, fmt.Errorf("%s: invalid repo root %q: %w", urlStr, metaImport.RepoRoot, err)
 	}
 	rr := &RepoRoot{
-		VCS:  ByCmd(metaImport.VCS),
-		Repo: metaImport.RepoRoot,
+		VCS: ByCmd(metaImport.VCS),
+		// ensure that dynamic discovery does not contain .git
+		// 99% of providers do not work with .git in URL
+		// e.g. Bitbucket, GitLab
+		Repo: strings.TrimSuffix(metaImport.RepoRoot, ".git"),
```
jedevc, line 506: Hm, we shouldn't strip this off right? It's part of the specified URL.

The user shouldn't specify `.git` on a GitLab URL if it's not supported.

core/integration/module_config_test.go
```diff
@@ -1166,134 +1178,211 @@ func (m *Test) Fn() ([]string, error) {
...
+type vcsTestCase struct {
+	name              string
+	gitTestRepoRef    string
+	gitTestRepoCommit string
+	baseHTMLURL       string // expected base HTML URLs, might differs from ref (e.g. not contain .git)
+	urlPathComponent  string // path separator to access `tree` view of src at commit, per provider
+	host              string // host component of repoURL
+}
+
+var vcsTestCases = []vcsTestCase{
```
jedevc, line 1191: I like this helper, but why is it neccessary to test against with and without `.git`? Effectively, this x6 the tests we need to run, x3 already seems like quite a few.

core/integration/module_config_test.go
```diff
@@ -1166,134 +1178,211 @@ func (m *Test) Fn() ([]string, error) {
...
-	gitTestRepoURL    = "github.com/dagger/dagger-test-modules"
-	gitTestRepoCommit = "2cb6cb4b0dba52c1e65b3ff46dd1a4a8f9a02f94"
-)
+// VCS to test behavior against
+type vcsTestCase struct {
+	name              string
+	gitTestRepoRef    string
+	gitTestRepoCommit string
+	baseHTMLURL       string // expected base HTML URLs, might differs from ref (e.g. not contain .git)
+	urlPathComponent  string // path separator to access `tree` view of src at commit, per provider
```
jedevc, line 1187: nit: for readability, if these are part of the "expected" values for the test case, they should be prefixed with `expected`.

core/modulesource.go
```diff
@@ -447,10 +447,16 @@ func (src *GitModuleSource) Symbolic() string {
 }
 
 func (src *GitModuleSource) HTMLURL() string {
-	u := "https://" + src.Root + "/tree/" + src.Commit
+	u := src.CloneURL
+	if strings.HasPrefix(src.CloneURL, "https://github.com") || strings.HasPrefix(src.CloneURL, "https://gitlab.com") {
+		u = u + "/tree/" + src.Commit
+	} else {
+		u = u + "/src/" + src.Commit
+	}
```
jedevc, line 455: Bleh, the fact we need to do this is annoying. Sounds good to me though, I'm sure we'll get bug reports from users, and we can fix these one-by-one as they come in.

nit: we should use the parsed urls and look at the host FYI, I think this would break with a `http://` scheme?

jedevc (commented):
Yay! Nice to have tests for this!

Left a couple nit comments, I'm not 100% sure about the `.git` stripping - if the VCS doesn't provide it, we should let the VCS error out with it, and not be super clever here.

core/integration/module_config_test.go
grouville: You're right, this has been implemented 🙏

core/modulesource.go
grouville: You're totally right, this has been extended ✅

core/integration/module_config_test.go
grouville: It is not necessary per se. But I wanted to have at least a few runs to make sure that all the tests were passing even with it.

This `.git` is going to be tested with private git repos, so I removed them from the matrix 🙏 x3 is already a lot I agree

engine/vcs/vcs.go
grouville: Mmh, I kind of agree but the current behavior of `RepoRootForImportPath` currently surprises me.

I feel more comfortable to have a coherent and expected behavior (detailed below), and be a bit smarter in this part of the logic instead of potentially paying the price later and handling those edge cases inside the `ModuleSource` part of the codebase.

A nice and easy mental model where:
- `Repo` represents a URL that a user can use to access the URL of the module, and all its variation (commit, tree view)
- `Root` represents the name of the module source. It is currently only being used as the name of the module. But being coherent on whether the user input ref contained a `.git` makes it a valid source of truth of the input ref (and would help us debug)

Given this heuristic:
- a repo should never contain a `.git`, whether it was specified or not
- a root should contain the `.git` if it was provided

#### Why is it important not to have the `.git` in the repo
Internally, `Repo` represents the URL to access the repo from a given ref. Upon manual tests, GitHub does support URLs with the `.git` inside ; but, most VCS do not, especially when trying to query the `tree` view of the commit (bitbucket / Gitlab)

#### Why is it important to keep track of the `.git`
If we follow the Go convention for private repo handling, and based on the current implementation of the `RepoRootForImportPath` it is recommended to use `.git` in the ref for private repos (due to GitLab returning fake metadata on `go-get=1` for unauthenticated refs). 

With this postulate, relying on this declarative model empowers the user and facilitates the handling on our side.

_What is the impact_ 
When having a ref with `.git`, the way to separate the root of the repo and potential subpath changes:
- without it, it uses HTTP requests
- with it, it uses the git CLI

I'll dig the implications a bit further on your https://github.com/dagger/dagger/pull/7708#issuecomment-2186297919. But, a nice way of handling root and subpaths is to ask users to actually specify the `.git`. This enables us to have ways to load auth at the `vcs.RepoRootForImportPath`

grouville: > Left a couple nit comments, I'm not 100% sure about the .git stripping - if the VCS doesn't provide it, we should let the VCS error out with it, and not be super clever here.

Plus, I don't think it is just being clever: Google's logic for GitLab is currently a bit weird -> https://github.com/dagger/dagger/pull/7663/files#diff-a0ad18eec985477f121c56ba61b7956832670d35b67db059eac6f429085386b0L159. Here, the user referenced a ref without a `.git`, the metadata parsing from GitLab contained a `.git`, but as stated above, GitLab does not allow URLs with tree view on a commit when the URL contains a `.git`

grouville:
Both failing tests are due to flakiness

engine/vcs/vcs.go
jedevc: Honestly, fair enough point :tada: I'm happy with it as it! There's maybe *some* edge cases where this might not work, e.g. if `dagger.git` and `dagger` were separate repos - but that feels sooooo edge-casey, that we can fix it when we get to it.

Thanks for all the work on this, will be awesome to have tests for this!

jedevc (approved):

PR #7666, opened 2024-06-14
chore: appease go linter after removing progrock

jedevc:
Ah, missed a linting failure in https://github.com/dagger/dagger/pull/7655.


gerhard (approved):
👍 🚀

PR #7667, opened 2024-06-14
feat(sdk/elixir): relocate module runtime to sdk/elixir

wingyplus:
This PR comes from discussion in [dev-engine](https://discord.com/channels/707636530424053791/1250226289919660105/1250226291832000542) and in [php](https://discord.com/channels/707636530424053791/1162025276872609832/1251168485736976445) discord channel.

I do relocate the runtime module from `sdk/elixir/runtime` into `sdk/elixir`. The benefit of this way is the sdk and runtime are run on the same commit, unlike previous implementation that the sdk use git remote to fetch sdk source.

However, this comes with the breaking changes. Previously, user who use sdk will fetch with 

```
dagger init --sdk=github.com/dagger/dagger/sdk/elixir/runtime
```

Will need to change to 

```
dagger init --sdk=github.com/dagger/dagger/sdk/elixir
```


wingyplus:
@helderco what do you prefer between move all files to `sdk/elixir` or still keep main.go and friends inside the runtime except dagger.json?

wingyplus:
@gmile Just FYI. :)

helderco:
> @helderco what do you prefer between move all runtime files to `sdk/elixir` or still keep it inside the runtime directory except dagger.json?

I think keeping source in `runtime` is neater.

gerhard (approved):
LGTM!

Please rebase & ensure all checks are green before merging 💪

wingyplus:
@gerhard Sorry for taking it so long. Rebased. 🙏

jedevc:
> > @helderco what do you prefer between move all runtime files to `sdk/elixir` or still keep it inside the runtime directory except dagger.json?
> 
> I think keeping source in `runtime` is neater.

I think keeping `runtime` consistent with the other SDKs is neater - if we want to move them around, we should do it across all the SDKs. Additionally, this is confusing since each SDK might have a `dev` module for developing the SDK itself, while `runtime` is entirely separate.

wingyplus:
@helderco @jedevc I can finally relocate it to runtime with keeping dagger.json into the root of sdk/elixir. It's make me less happy than having runtime source in `sdk/elixir` because it cannot allows me to run `--sdk=..` but needs to push to git first but it's keep consistent with other SDKs. 

Please kindly review. 🙏

wingyplus:
@helderco @jedevc or i should wait for https://github.com/dagger/dagger/pull/7744 and use that to locate the sdk directory?

sdk/elixir/dagger.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "elixir-sdk",
   "sdk": "go",
-  "source": ".",
-  "engineVersion": "v0.11.9"
+  "include": ["."],
```
helderco, line 4: This looks too broad. Should include only what's necessary for what the runtime module needs to do to generate the SDK dir in the module's sources.

helderco:
> @helderco @jedevc or i should wait for #7744 and use that to locate the sdk directory?

We can merge this sooner and adjust to context directory later as that's not quite imminent yet.

sdk/elixir/dagger.json
wingyplus: I try adding `exclude` but it didn't works for me. Even changing to add `!` to include (like in the commit ca96233a7fd6d49c1a9dce5dbc2b28b2fd8c8e94) is also not working. 😢

wingyplus: Oh, I just realize that `include` option doesn't effect `Directory("..")` on https://github.com/dagger/dagger/pull/7667/files#diff-6665f7b190862df91665d328260eaf53cdfbe4ddaddbbfb52a290f531a8671cdR24. I start confusing now.

helderco: > Oh, I just realize that `include` option doesn't effect `Directory("..")` on https://github.com/dagger/dagger/pull/7667/files#diff-6665f7b190862df91665d328260eaf53cdfbe4ddaddbbfb52a290f531a8671cdR24. I start confusing now.

It works for me:
```json
{
  "name": "elixir-sdk",
  "sdk": "go",
  "include": [
    "dagger_codegen",
    "lib",
    "mix.exs"
  ],
  "source": "runtime",
  "engineVersion": "v0.12.1"
}
```
Note: Remove `Terminal()` from `runtime/main.go` as you can add that via the CLI.

```console
$ dagger call sdksource-dir
_type: Directory
entries:
    - dagger.json
    - dagger_codegen
    - lib
    - mix.exs
    - runtime
```

wingyplus: Yeah, at 771ed2cb2f265849f10fff20d769f7f1cb677547, I'm now at the same state as you run:

```yaml
_type: Directory
entries:
    - .formatter.exs
    - .gitignore
    - LICENSE
    - dagger.json
    - dagger_codegen
    - lib
    - mix.exs
    - mix.lock
    - runtime
```

But when I try `mkdir potato && dagger init --sdk=github.com/wingyplus/dagger/sdk/elixir@771ed2cb2f265849f10fff20d769f7f1cb677547 --source=.`, I still got all files unlike get from `sdksource-dir`:

```
$ ls -la dagger_sdk
total 144
drwxr-xr-x  15 thanabodee  staff    480 Jul 19 23:30 ./
drwxr-xr-x   7 thanabodee  staff    224 Jul 20 01:42 ../
drwxr-xr-x  38 thanabodee  staff   1216 Jul 20 01:36 .changes/
-rw-r--r--   1 thanabodee  staff   1200 Jul 20 01:36 .changie.yaml
-rw-r--r--   1 thanabodee  staff   8866 Jul 20 01:36 .credo.exs
-rw-r--r--   1 thanabodee  staff     97 Jul 20 01:36 .formatter.exs
-rw-r--r--   1 thanabodee  staff    666 Jul 20 01:36 .gitignore
-rw-r--r--   1 thanabodee  staff  19531 Jul 20 01:36 CHANGELOG.md
-rw-r--r--   1 thanabodee  staff  10758 Jul 20 01:36 LICENSE
-rw-r--r--   1 thanabodee  staff   1570 Jul 20 01:36 README.md
-rw-r--r--   1 thanabodee  staff    419 Jul 20 01:36 dagger.json
drwxr-xr-x   4 thanabodee  staff    128 Jul 19 23:30 lib/
-rw-r--r--   1 thanabodee  staff   1159 Jul 20 01:36 mix.exs
-rw-r--r--   1 thanabodee  staff   3170 Jul 20 01:36 mix.lock
drwxr-xr-x   5 thanabodee  staff    160 Jul 20 01:36 test/
```

That's what I get confuse what's going on here. T_T

wingyplus: > Note: Remove Terminal() from runtime/main.go as you can add that via the CLI. 

I using it during debugging during call `dagger init` but yeah, removing it. 🙏

wingyplus: From discussed in https://discord.com/channels/707636530424053791/1133501257051553845/1263913471326818385. I think I start to understand what's happens now. 🙏

sdk/elixir/dagger.json
```diff
@@ -0,0 +1,23 @@
...
+    "**/test",
+    "**/_build",
+    "**/deps",
+    "**/doc",
+    "**/.elixir-tools",
+    "**/.elixir_ls",
+    "**/.lexical",
+    "**/credo.exs",
+    "**/erl_crash.dump"
+  ],
```
helderco, line 20: Are all of these excludes really necessary? As I said before, it's better to rely more on _include_ and only add to _exclude_ when necessary. The reason is that it's quite easy to add new files to `sdk/elixir` and you don't want to have to remember to add it to the list of excludes.

See if you can **avoid starting the patterns with `**/`** as I think that's more difficult to manage.

One example is if you include 'lib/**/*.ex`, then you don't need to exclude `**/test`, etc.

sdk/elixir/runtime/main.go
```diff
@@ -17,13 +18,10 @@ const (
...
 	// +optional
-	sdkSourceDir *Directory,
+	sdkSourceDir *dagger.Directory,
 ) *ElixirSdk {
 	if sdkSourceDir == nil {
-		sdkSourceDir = dag.Git("https://github.com/dagger/dagger.git").
-			Branch("main").
-			Tree().
-			Directory("sdk/elixir")
+		sdkSourceDir = dag.CurrentModule().Source().Directory("..")
```
helderco, line 24: You can add `.WithoutDirectory("runtime")` here.

sdk/elixir/runtime/main.go
```diff
@@ -142,25 +140,34 @@ func (m *ElixirSdk) WithNewElixirPackage(ctx context.Context, modName string) *E
...
+				"changie.yaml",
+				".credo.exs",
 				"dagger_codegen",
+				"test",
 				// We'll do generate code on the next step.
 				"lib/dagger/gen",
+				// Now the runtime is in the root of the SDK. We need
+				// to exclude them.
+				"dagger.json",
 				"runtime",
```
helderco, line 163: This is probably not necessary. The `SDKSourceDir` should already include only what you need for this step.

sdk/elixir/runtime/main.go
wingyplus: Added and also exclude `dagger.json` at this step.

sdk/elixir/dagger.json
wingyplus: I changed to include only necessary files. The exclude here is not necessary anymore. 🎉

sdk/elixir/runtime/main.go
wingyplus: Fixed at https://github.com/dagger/dagger/pull/7667#discussion_r1685321338.

wingyplus:
I do rebased and squash commit to remove noise commits. 

Ready to review again. 🙇

sdk/elixir/runtime/main.go
helderco: > Fixed at [#7667 (comment)](https://github.com/dagger/dagger/pull/7667#discussion_r1685321338).

That was just `dagger.json`, but I actually meant all of these exclude patterns. There are things here that are not in `dagger.json`'s "include", so no need to exclude what's not in the dir.

helderco:
> I do rebased and squash commit to remove noise commits.
> 
> Ready to review again. 🙇

Almost there, just need to address https://github.com/dagger/dagger/pull/7667#discussion_r1688203563.

wingyplus:
> > I do rebased and squash commit to remove noise commits.
> > Ready to review again. 🙇
> 
> Almost there, just need to address [#7667 (comment)](https://github.com/dagger/dagger/pull/7667#discussion_r1688203563).

@helderco Apply the suggestion. Please help review it again.

sdk/elixir/runtime/main.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"dagger/runtime/internal/dagger"
```
wingyplus, line 5: I just realize that I declare a `module` in go.mod as `dagger/runtime`. I think renaming it to `elixir-sdk` for consistent with Python. 

BTW, the TypeScript using `main` as a module name. Do we need to renaming it into `typescript-sdk`? I can open a PR to fix this if you want.

wingyplus: Fixed the Elixir go module name. 👍

helderco (approved):
I made a few simplifications. Good to merge.

helderco:
Failed test is a known flake.

wingyplus:
@helderco thank you 🙏

PR #7670, opened 2024-06-15
ci: use legacy iptables instead of nf_tables

marcosnils:
the v0.11.7 release bumped the base images that the engine uses which
caused a newer version of iptables to be used. This newer version of
iptables which uses nftables is still not widely supported and it's
currently causing issues for some users.

cc @iximiuz 

ref: https://discord.com/channels/707636530424053791/1251537699903639623

### DIFF
--- a/ci/build/builder.go
+++ b/ci/build/builder.go
@@ -163,13 +163,17 @@ func (build *Builder) Engine(ctx context.Context) (*dagger.Container, error) {
 				// for Buildkit
 				"git", "openssh", "pigz", "xz",
 				// for CNI
-				"dnsmasq",
+				"dnsmasq", "iptables", "ip6tables", "iptables-legacy",
 			}).
-			WithExec([]string{
-				"apk", "add",
-				"-X", "https://dl-cdn.alpinelinux.org/alpine/v3.18/main",
-				"iptables=1.8.9-r2", "ip6tables=1.8.9-r2",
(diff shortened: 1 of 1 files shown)


marcosnils:
moved this PR into draft since it seems like something else is happening which doesn't allow the engine to start now. Currently getting this error message now:

```
buildkitd: failed to create engine: failed to create network providers: CNI setup error: plugin type="bridge" failed (add): running [/sbin/iptables -t nat -C CNI-4bf6835a2527ef38ccf8ed61 -d 10.87.0.2/16 -j ACCEPT -m comment --comment name: "dagger" id: "rpe40a773875zfc429qjpqbvd" --wait]: exit status 2: iptables v1.8.10 (legacy): Couldn't load match `comment':No such file or directory
```

marcosnils:
This is ready to review. The only way I could make the alpine version to work is to install the `iptables` packages from 3.18 (legacy by default). I've tried by installing `iptables-legacy` and then creating a symlin to `/sbin/iptables` but for some reason that didn't work.

jedevc:
@marcosnils I worked out how to install legacy iptables - you need to install *both* `iptables` and `iptables-legacy` - otherwise an important shared library is missing, and so it won't work.

I pushed a commit that does this, that should fix the issue you're having.

jedevc (approved):

PR #7671, opened 2024-06-16
tui: navigation, scrolling, optimizations

vito:
[asciicast](https://asciinema.org/a/vdUEqRXX0K6Dw9zqLjkwlCxNc)

* ~~Now shows effects at their origin (`withExec`) rather than their unlazying point (`sync`).~~
  * Moved to separate PR: #7686 
* Now supports navigation, instead of being stuck with the bottom of the visible region
  * `hjkl` or arrow keys to move around
  * `enter` to "zoom in" to the selected span
  * `end` or `space` to follow the bottom of the screen
  * `home` to go to beginning
* Instead of having spans disappear after a second, we just collapse them when they finish.
* Now supports mouse scroll wheel again
  * I initially skipped this because I found it occasionally annoying for the TUI to steal mouse input. But you can just press Shift to bypass it.
* Now only renders the visible region, instead of rendering everything and cutting it off
  * This was a lot of work, but seems obviously worth it for large traces.
  * It's now possible to run our full integration suite without the TUI starting to lag. It used to get _real_ bad before.
* Now only renders logs when the logs need to be painted, rather than when the CLI receives them
* Tons of optimizations in the UI rendering loop to avoid repeated computations
* Avoid pushing cloud URL offscreen (it was also messing up the TUI a bit by printing directly to stdio)
* Now detects inifnite loop when printing simplified spans to avoid RAM usage explosion (cc @jedevc this should buy us some time)
* Now shows a keymap for better discoverability
* Now supports `+` and `-` keybinds increase and decrease verbosity at runtime
  * Want to know what's going on under `initialize`? Just hold `+`!

Here's a profile comparison for `dagger -m . call --source=.:default test important --race=true` after letting it run for ~9 minutes:

Before:

image

After:

image

Still spending a chunk of time in `WalkSpans`, but the total time in CPU is much shorter (23.58s -> 3.04s), and there's much less going on in there now (mostly just map usage).

### DIFF
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -20,6 +20,7 @@ import (
 	"dagger.io/dagger/telemetry"
 	"github.com/dagger/dagger/dagql/call"
 	"github.com/dagger/dagger/dagql/call/callpbv1"
+	"github.com/dagger/dagger/engine/slog"
 	sdklog "go.opentelemetry.io/otel/sdk/log"
 )
 
@@ -70,19 +71,25 @@ func DumpID(out *termenv.Output, id *call.ID) error {
 	for dig, call := range dag.CallsByDigest {
 		db.Calls[dig] = call
 	}
-	r := renderer{
-		db:    db,
-		width: -1,
-	}
+	r := newRenderer(db, -1, FrontendOpts{})
 	return r.renderCall(out, nil, id.Call(), "", 0, false, false)
 }
 
 type renderer struct {
 	FrontendOpts
 
-	db *DB
+	db        *DB
+	width     int
+	rendering map[string]bool
+}
 
-	width int
+func newRenderer(db *DB, width int, fe FrontendOpts) renderer {
+	return renderer{
+		FrontendOpts: fe,
+		db:           db,
+		width:        width,
+		rendering:    map[string]bool{},
+	}
 }
 
 const (
@@ -109,6 +116,13 @@ func (r renderer) renderIDBase(out *termenv.Output, call *callpbv1.Call) error {
 }
 
 func (r renderer) renderCall(out *termenv.Output, span *Span, call *callpbv1.Call, prefix string, depth int, inline bool, internal bool) error {
+	if r.rendering[call.Digest] {
+		slog.Warn("cycle detected while rendering call", "span", span.Name(), "call", call.String())
+		return nil
+	}
+	r.rendering[call.Digest] = true
+	defer func() { delete(r.rendering, call.Digest) }()
+
 	if !inline {
 		fmt.Fprint(out, prefix)
 		r.indent(out, depth)
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -115,7 +115,7 @@ func (r renderer) renderIDBase(out *termenv.Output, call *callpbv1.Call) {
 		parent = parent.Foreground(moduleColor)
 	}
 	fmt.Fprint(out, parent.String())
-	if r.Verbosity > 2 && call.ReceiverDigest != "" {
+	if r.Verbosity > ShowDigestsVerbosity && call.ReceiverDigest != "" {
 		fmt.Fprint(out, out.String(fmt.Sprintf("@%s", call.ReceiverDigest)).Foreground(faintColor))
 	}
 }
@@ -212,7 +212,7 @@ func (r renderer) renderCall(
 		fmt.Fprint(out, typeStr)
 	}
 
-	if r.Verbosity > 2 {
+	if r.Verbosity > ShowDigestsVerbosity {
 		fmt.Fprint(out, out.String(fmt.Sprintf(" = %s", call.Digest)).Foreground(faintColor))
 	}
 
@@ -363,17 +363,27 @@ func (r renderer) renderDuration(out *termenv.Output, span *Span) {
 // 	return nil
 // }
 
+const (
+	ShowCompletedVerbosity    = 1
+	ShowInternalVerbosity     = 2
+	ShowEncapsulatedVerbosity = 2
+	ShowSpammyVerbosity       = 3
+	ShowDigestsVerbosity      = 3
+
+	DefaultVerbosity = ShowCompletedVerbosity
+)
+
 func (opts FrontendOpts) ShouldShow(tree *TraceTree) bool {
 	if opts.Debug {
 		// debug reveals all
 		return true
 	}
 	span := tree.Span
-	if span.IsInternal() && opts.Verbosity < 2 {
+	if span.IsInternal() && opts.Verbosity < ShowInternalVerbosity {
 		// internal steps are hidden by default
 		return false
 	}
-	if tree.Parent != nil && (span.Encapsulated || tree.Parent.Span.Encapsulate) && tree.Parent.Span.Err() == nil && opts.Verbosity < 2 {
+	if tree.Parent != nil && (span.Encapsulated || tree.Parent.Span.Encapsulate) && tree.Parent.Span.Err() == nil && opts.Verbosity < ShowEncapsulatedVerbosity {
 		// encapsulated steps are hidden (even on error) unless their parent errors
 		return false
 	}
@@ -381,7 +391,7 @@ func (opts FrontendOpts) ShouldShow(tree *TraceTree) bool {
 		// show errors
 		return true
 	}
-	if opts.TooFastThreshold > 0 && span.Duration() < opts.TooFastThreshold && opts.Verbosity < 3 {
+	if opts.TooFastThreshold > 0 && span.Duration() < opts.TooFastThreshold && opts.Verbosity < ShowSpammyVerbosity {
 		// TODO(vito): bring this back
 		// if tree.Parent != nil && tree.Parent.Span.SpanContext().SpanID() != opts.PrimarySpan {
 		// 	return false
@@ -393,10 +403,10 @@ func (opts FrontendOpts) ShouldShow(tree *TraceTree) bool {
 		// show running steps
 		return true
 	}
-	// if opts.GCThreshold > 0 && time.Since(span.EndTime()) > opts.GCThreshold && opts.Verbosity < 1 {
-	// 	// stop showing steps that ended after a given threshold
-	// 	return false
-	// }
+	if opts.GCThreshold > 0 && time.Since(span.EndTime()) > opts.GCThreshold && opts.Verbosity < ShowCompletedVerbosity {
+		// stop showing steps that ended after a given threshold
+		return false
+	}
 	return true
 }
 
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -224,7 +224,7 @@ func (r renderer) renderCall(
 	return nil
 }
 
-func (r renderer) renderVertex(
+func (r renderer) renderSpan(
 	out *termenv.Output,
 	span *Span,
 	name string,
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -33,6 +33,12 @@ type FrontendOpts struct {
 
 	// Verbosity is the level of detail to show in the TUI.
 	Verbosity int
+
+	// Don't show things that completed beneath this duration. (default 100ms)
+	TooFastThreshold time.Duration
+
+	// Remove completed things after this duration. (default 1s)
+	GCThreshold time.Duration
 }
 
 type Frontend interface {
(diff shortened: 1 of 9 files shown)


dagql/idtui/frontend_plain.go
```diff
@@ -246,7 +243,9 @@ func (fe plainLogExporter) Export(ctx context.Context, logs []sdklog.Record) err
 	fe.mu.Lock()
 	defer fe.mu.Unlock()
 
-	slog.Debug("frontend exporting logs", "logs", len(logs))
+	// NOTE: if we log in here we'll just go into a loop, so...
+	// let's just hope everything is working by now :D
+	// slog.Debug("frontend exporting logs", "logs", len(logs))
```
jedevc, line 248: Mm, I have not actually found these logs of *particular* value - do we actually need any of these in `ExportSpans`/etc? They're quite noisy when enabled, and I'm not sure if there's a case where they're useful.

If we want to see the logs that the client receives, then it feels like maybe we should have that in the db?

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.22
 require (
 	dagger.io/dagger v0.11.6
 	github.com/dagger/dagger/engine/distconsts v0.11.6
+	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
```
jedevc, line 8: :shakes_fist_at_automated_tooling:

Love it, yeah, why not, makes total sense, this block definitely isn't separated for clarity :cry:

dagql/idtui/frontend_plain.go
vito: As someone who spent the last month debugging OpenTelemetry draining issues I can attest to their value. :sweat_smile:

Could maybe move them to `ExtraDebug` level but every time we do this it only makes debugging OpenTelemetry more difficult for me. I would suggest just leaving these as-is until the coast feels clear.

> If we want to see the logs that the client receives, then it feels like maybe we should have that in the db?

Not sure what you mean here.

dagql/idtui/frontend_pretty.go
```diff
@@ -446,16 +626,58 @@ func (fe *frontendPretty) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+		case "-":
+			fe.FrontendOpts.Verbosity--
+			if fe.FrontendOpts.Verbosity < 0 {
+				fe.FrontendOpts.Verbosity = 0
+			}
+			fe.recalculateViewLocked()
+			return fe, nil
+		case "c":
+			if fe.cloudURL != "" {
+				browser.OpenURL(fe.cloudURL)
```
jedevc, line 664: Can we give some visual feedback when this happens? I was testing this, didn't realize it had opened, hammered the button, and went to my browser later to see 100s of cloud tabs open?

dagql/idtui/frontend_pretty.go
```diff
@@ -430,9 +601,18 @@ func (fe *frontendPretty) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+			fe.goDown()
+		case tea.MouseButtonWheelUp:
+			fe.goUp()
+		}
+		return fe, nil
+
 	case tea.KeyMsg:
 		switch msg.String() {
-		case "q", "esc", "ctrl+c":
+		case "q", "ctrl+c":
```
jedevc, line 615: Hm, something here is weird, it "feels" like CTRL+C now takes longer for the process to exit - there's a noticeable pause.

dagql/idtui/frontend_pretty.go
```diff
@@ -446,16 +626,58 @@ func (fe *frontendPretty) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+			}
+			fe.recalculateViewLocked()
+			return fe, nil
+		case "c":
+			if fe.cloudURL != "" {
+				browser.OpenURL(fe.cloudURL)
+			}
+			return fe, nil
+		case "enter":
+			fe.zoomed = fe.focused
```
jedevc, line 668: Once we're zoomed, can we give some indication of the "context" we're in? Maybe like a breadcrumbs trail?

jedevc (commented):
Left some little things, I like this though :tada: :tada:

go.mod
vito: Yeah that's been annoying. Maybe if we move it below the other block it'll be safe? :thinking:

dagql/idtui/frontend.go
```diff
@@ -338,41 +372,47 @@ func (r renderer) renderDuration(out *termenv.Output, span *Span) {
...
+		// ignore fast steps; signal:noise is too poor
+		return false
 	}
-	if row.IsRunning {
+	if tree.IsRunningOrChildRunning {
 		// show running steps
 		return true
 	}
-	if sf.gcThreshold > 0 && time.Since(span.EndTime()) > sf.gcThreshold && opts.Verbosity < 1 {
+	if opts.GCThreshold > 0 && time.Since(span.EndTime()) > opts.GCThreshold && opts.Verbosity < ShowCompletedVerbosity {
```
jedevc, line 415: :question: this doesn't seem to be working on `tty` progress anymore?

dagql/idtui/frontend_pretty.go
vito: It used to skip straight to an immediate interrupt because with the previous draining implementation it was too easy to end up in a scenario where you're stuck waiting on a "doomed" span (i.e. parent `kill -9`ed).

Now that that's fixed, I changed interrupt to not interrupt immediately, which gives telemetry a chance to drain, but takes a second or two. I can look into why. Maybe we have a 1 second sleep retry somewhere.

dagql/idtui/frontend_pretty.go
jedevc: Oooh, maybe we could also make the `Cloud` text a hyperlink :eyes:

dagql/idtui/frontend_plain.go
jedevc: Yeah fair enough - worth keeping then!

> Not sure what you mean here.

I mean, we could move them into `ExportSpans` in `db.go`, that way they can be shared between the different frontends, instead. Just a suggestion, it may not work for logistical reasons :sweat_smile:

dagql/idtui/frontend.go
vito: The verbosity level defaults to 1, so this only kicks in when you press `-`

dagql/idtui/frontend_pretty.go
vito: Done! I think you might need to hold "Shift" for it to work though, not sure if worth it. Try it out in your terminal? This sort of thing varies a lot between implementations.

dagql/idtui/frontend_pretty.go
vito: Done! Now it'll pin your zoomed span to the top and nest the children, feels a lot more intuitive and avoids the feeling of having another UI element:

image

vito:
@gerhard @jedevc Tossed this in the milestone since it seems ready and just needs approval, but feel free to skip it if you want to keep release scope tidy!

jedevc:
Gonna bring this one out of the milestone - I think it's ready, but:
1. Ideally the next release to run will be v0.12.0 -  that fits nicely into here
2. Anything to improve the chances of giving a stable v0.11.8 :pray: :crossed_fingers: :cry:

dagql/idtui/frontend.go
```diff
@@ -340,41 +375,47 @@ func (r renderer) renderDuration(out *termenv.Output, span *Span) {
...
-		if row.Parent != nil && row.Parent.Span.SpanContext().SpanID() != sf.db.PrimarySpan {
-			return false
-		}
+	if opts.TooFastThreshold > 0 && span.Duration() < opts.TooFastThreshold && opts.Verbosity < ShowSpammyVerbosity {
+		// TODO(vito): bring this back
+		// if tree.Parent != nil && tree.Parent.Span.SpanContext().SpanID() != opts.PrimarySpan {
+		// 	return false
+		// }
+		// ignore fast steps; signal:noise is too poor
+		return false
```
jedevc, line 412: Are we intentionally not bringing this back before merging? Ideally I'd still like it, it made quite a few steps just never appear if they were quick enough.

dagql/idtui/frontend_pretty.go
```diff
@@ -332,50 +298,250 @@ func (fe *frontendPretty) Background(cmd tea.ExecCommand) error {
...
+
+	// Blank line prior to keymap
+	for i, key := range []keyHelp{
+		{quitMsg, []string{"q", "ctrl+c"}, true},
+		{out.Hyperlink(fe.cloudURL, "Cloud"), []string{"c"}, fe.cloudURL != ""},
+		{"move", []string{"←↑↓→", "up", "down", "left", "right", "h", "j", "k", "l"}, true},
+		{"first", []string{"home"}, true},
+		{"last", []string{"end", " "}, true},
+		{"zoom", []string{"enter"}, true},
+		{"reset", []string{"esc"}, true},
```
jedevc, line 328: incredible nit: I got confused between `reset` and `quit` - `reset` goes to the top-level, and `quit` terminates everything.

I would probably expect `quit` to be at the end of the line as the last option, and maybe call `reset` something else? I can't think of a better name though, so take with a pinch of salt :salt:

dagql/idtui/frontend_pretty.go
jedevc: Yay :tada: feels like cloud now :tada:

dagql/idtui/frontend.go
vito: Didn't mean to leave this commented (it just needed access to more info post-refactor), but I do find the heuristic a little confusing and don't know how it should be updated for the new UI. Why do we want to show really fast spans beneath the primary span in particular, as opposed to the usual "really fast = boring/noise" logic? Does this affect the Plain UI more than the regular TUI? Does it affect the primary span in a way that's unique to the primary span and not other calls? Do we want this to apply to zoomed spans, too?

I'm trying to tease apart whether this is really a heuristic that belongs here or if it's working around a gap in other parts of the UI, like maybe when you select a span we can have a key to show its full recipe in a separate pane or something.

dagql/idtui/frontend_pretty.go
vito: Done!

* moved quit to the end
* reset -> unzoom, show only when zoomed

dagql/idtui/frontend.go
vito: Brought this back in the last commit to keep the PR unblocked but yeah, still find it a bit confusing. Found myself wondering if other things should also be revealed in this scenario (like internal steps) and couldn't intuit a rhyme or reason.

dagql/idtui/frontend.go
```diff
@@ -32,6 +33,12 @@ type FrontendOpts struct {
 
 	// Verbosity is the level of detail to show in the TUI.
 	Verbosity int
+
+	// Don't show things that completed beneath this duration. (default 100ms)
+	TooFastThreshold time.Duration
```
TomChv, line 38: I like this, will save a lot of logs, however will this still appears on cloud?

TomChv (commented):
Just left a question, I'm going to try the PR manually then give an approval

TomChv:
I gave it a try, this is absolutely amazing! Love it

I don't understand the `first` and `last` section in the context though, but I guess it's me missing something.

We should have a page on our doc with some records 
/cc @vikram-dagger 

That's absolutely amazing.

Uploading Screen Recording 2024-06-20 at 23.24.32.mov…

I also tried with failing pipelines, it works well, giving a nice output way clearer than before, this is an amazing improvement

TomChv (approved):

dagql/idtui/frontend.go
vito: Yeah this is only for the local TUI. I've experimented with a similar change for Cloud but haven't found a version I'm 100% happy with yet.

vito:
> Is there some ways btw to tests that feature with UT? Like by checking the content of the SpanTree or asserts some values? That would be nice for later (not in this PR indeed)

Possibly yeah. I've also had good results with golden-style testing in Progrock, which let you cover a lot of behavior with fast tests + high confidence (you can visually see if it looks "right," catch regressions with ergonomic errors (you'll see the diff), and update them easily. ([example](https://github.com/vito/progrock/blob/main/tape_test.go) - though there's not much to see)

dagql/idtui/frontend.go
jedevc: Yeah, this is tricky, the heuristic definitely isn't perfect but there is a reason for it.

Suppose I have `dagger call do-thing process-it export --path=thing`. Ideally, my top-level spans should look *something* like:

- `MyModule`
- `MyModule.DoThing`
- `Thing.ProcessIt`
- `Directory.Export`

Each of these is tied directly to a CLI subcommand, so they should *always* show. The reason I originally noticed this was because the `export` was sometimes showing, sometimes not, and was essentially racy depending on the size of the file.

I think we should always show these user subcommand spans at the top level, which translated into the heuristic of always showing the top-level.

PR #7679, opened 2024-06-17
Fix forced simplification infinite progress loop on high verbosity settings

jedevc:
Reported by @vito in an internal discord: https://discord.com/channels/707636530424053791/1250122361131499664/1250193357683163208

> huh, running that command with --debug or -vvv seems to cause the CLI to eat up all of my memory, seems like a bug with the call simplification code
>
> last time this came up it was because of a call simplifying to itself or something and ending up in a loop, all the memory usage is from trying to render the indentation

Turns out, the changes to call simplification to get `blob`s to never show was a little faulty.

For, example, on high verbosity settings, the following could happen:
- `moduleSource.resolveDependency` is an identity function (for git sources), taking an argument and returning it unmodified.
- Arguments in a call to `moduleSource.resolveDependency(dep: ...)` (from an unrelated function) will be forced to be simplified, since it's an internal function.
- There are no reasonable simplifications for the `dep` argument here - so since it must be simplified, it is simplified to `moduleSource.resolveDependency(dep: ...)` (since `resolveDependency` is an identity function.
- This recurses infinitely, with `dep` being expanded to it's parent.

There are two fixes in this PR (they both separately solve the issue, but both of them are logic issues).

1. The first fix ensures that we *always* return a valid simplification, and never a complexification, even when `force`d to `Simplify`. So we'll never expand it to a "parent".
2. The second fix tries to be a little bit less aggressive in using the `force` option, by only enforcing it for the first internal call we see, not for all the internal calls recursively (which is a bit too aggressive, and arguably, not super useful).

### DIFF
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -147,12 +147,12 @@ func (r renderer) renderCall(out *termenv.Output, span *Span, call *callpbv1.Cal
 				val := arg.GetValue()
 				fmt.Fprint(out, " ")
 				if argDig := val.GetCallDigest(); argDig != "" {
-					var argSpan *Span
-					var argInternal bool
-					if span != nil {
-						argSpan = r.db.MostInterestingSpan(argDig)
-						if argSpan != nil {
-							argInternal = argSpan.Internal
+					argSpan := r.db.MostInterestingSpan(argDig)
+					argInternal := false
+					if argSpan != nil {
+						argInternal = argSpan.Internal
+						if span == nil {
+							argSpan = nil
 						}
 					}
 					argCall := r.db.Simplify(r.db.MustCall(argDig), argInternal)
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -74,7 +74,7 @@ func DumpID(out *termenv.Output, id *call.ID) error {
 		db:    db,
 		width: -1,
 	}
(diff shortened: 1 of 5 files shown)


dagql/idtui/frontend.go
```diff
@@ -147,16 +147,18 @@ func (r renderer) renderCall(out *termenv.Output, span *Span, call *callpbv1.Cal
...
-					var argInternal bool
-					if span != nil {
-						argSpan = r.db.MostInterestingSpan(argDig)
-						if argSpan != nil {
-							argInternal = argSpan.Internal
+					forceSimplify := false
+					internal := internal
+					argSpan := r.db.MostInterestingSpan(argDig)
+					if argSpan != nil {
+						forceSimplify = argSpan.Internal && !internal // only for the first internal call (not it's children)
```
vito, line 154: Checking I have this right: if we're rendering a non-internal call, force simplification for internal args. But if we're already rendering an internal call, there's no need to simplify internal args, since we're already internal.

vito (approved):
previously: https://github.com/dagger/dagger/pull/6631 :smile:

dagql/idtui/frontend.go
jedevc: Yes :laughing: sorry, I'm kinda struggling if there's an easier way to write this :cry:

PR #7683, opened 2024-06-17
Interactive Debugging

aluzzardi:
Add a `--interactive` (`-i`) option to `dagger call` which will spawn
a terminal when a container execution fails

### DIFF
--- a/engine/buildkit/executor.go
+++ b/engine/buildkit/executor.go
@@ -46,6 +46,12 @@ type ExecutionMetadata struct {
 	SecretToken string
 	Hostname    string
 
+	// Unique (random) ID for this execution.
+	// This is used to deduplicate the same execution that gets evaluated multiple times.
+	ExecID string
+
+	// Internal execution initiated by dagger and not the user.
+	// Used when executing the module runtime itself.
 	Internal bool
 
 	EncodedModuleID     string
--- a/engine/buildkit/executor.go
+++ b/engine/buildkit/executor.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/containerd/console"
 	runc "github.com/containerd/go-runc"
+	"github.com/dagger/dagger/dagql/call"
 	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/executor"
 	"github.com/moby/buildkit/executor/oci"
@@ -54,6 +55,8 @@ type ExecutionMetadata struct {
 	// Used when executing the module runtime itself.
 	Internal bool
 
+	// TODO: can rm EncodedModuleID now
+	CallID              *call.ID
 	EncodedModuleID     string
 	EncodedFunctionCall json.RawMessage
 
--- a/engine/buildkit/executor.go
+++ b/engine/buildkit/executor.go
@@ -46,6 +46,8 @@ type ExecutionMetadata struct {
 	SecretToken string
 	Hostname    string
 
+	Internal bool
+
 	EncodedModuleID     string
 	EncodedFunctionCall json.RawMessage
 
--- a/engine/buildkit/ref.go
+++ b/engine/buildkit/ref.go
@@ -423,7 +423,11 @@ func debugContainer(ctx context.Context, execOp *bksolverpb.ExecOp, execErr *llb
 
 	output := idtui.NewOutput(term.Stderr)
 	fmt.Fprintf(term.Stderr,
-		output.String(idtui.IconFailure).Foreground(termenv.ANSIRed).String()+" Exec failed, attaching terminal\r\n")
+		output.String(idtui.IconFailure).Foreground(termenv.ANSIRed).String()+" Exec failed, attaching terminal: ")
+	if err := idtui.DumpID(output, execMd.CallID); err != nil {
+		return fmt.Errorf("failed to serialize service ID: %w", err)
+	}
+	fmt.Fprintf(term.Stderr, "\r\n")
 	fmt.Fprintf(term.Stderr,
 		output.String("! %s\r\n\n").Foreground(termenv.ANSIYellow).String(), execErr.Error())
 
--- a/engine/buildkit/ref.go
+++ b/engine/buildkit/ref.go
@@ -422,14 +422,17 @@ func debugContainer(ctx context.Context, execOp *bksolverpb.ExecOp, execErr *llb
 	}
 
 	output := idtui.NewOutput(term.Stderr)
-	fmt.Fprintf(term.Stderr,
+	fmt.Fprint(term.Stderr,
 		output.String(idtui.IconFailure).Foreground(termenv.ANSIRed).String()+" Exec failed, attaching terminal: ")
-	if err := idtui.DumpID(output, execMd.CallID); err != nil {
+	dump := idtui.Dump{Newline: "\r\n", Prefix: "    "}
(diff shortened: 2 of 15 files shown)


engine/buildkit/ref.go
```diff
@@ -348,6 +356,101 @@ func wrapError(ctx context.Context, baseErr error, sessionID string) error {
...
 }
 
+func debugContainer(ctx context.Context, execOp *bksolverpb.ExecOp, execErr *llberror.ExecError, client *Client) error {
+	if !client.Opts.Interactive {
+		return nil
+	}
+
+	// If this is the (internal) exec of the module itself, we don't want to spawn a terminal.
+	for _, env := range execOp.Meta.Env {
+		if strings.HasPrefix(env, "_DAGGER_CALL_DIGEST=") {
```
sipsma, line 366: This part was a hack to unblock proving the concept. I can live with it (just slightly fragile to future code changes) but ideally we'll be able instead get the `ExecutionMetadata` struct associated with the exec here (https://github.com/sipsma/dagger/blob/66e98c39ef4ed0aa17bbb2c8b47e48b10620bdaa/engine/buildkit/executor.go#L43-L43)

Then we can plumb whatever we want here, which should also help with providing more debug information on what failed container they are even dropping into, customizing to something besids /bin/sh.

Unfortunately on a first check it looks like buildkit is not quite giving us the field we need to extract that from. I'm double checking now and if it's indeed missing will send an upstream PR to support it provided it's as simple as I'm imagining right now.

Not a blocker, after that I'll finish reviewing, just an FYI.

sipsma (commented):
One high-level question, if you currently do something like:
1. lazily do a Container.WithExec and get an output directory
2. Provide that lazy directory to another container, synchronously evaluate it
3. The container.withExec in `1.` fails

Will we open a terminal in the original container in `1.` *and* the container in `2`? Or just `1.`?

Basically, if one container depends on another and there's a failure, do we keep dropping into debug shells all along the DAG? 

Not a blocker if so, just could be confusing. Ideally we could find a way to just drop into "leaf" failures so to speak.

engine/buildkit/ref.go
sipsma: Sent out that upstream PR here https://github.com/moby/buildkit/pull/5108

aluzzardi:
As far as I can see, only the leaf container gets terminal'd:

```go
func (m *Interactive) Two(ctx context.Context) (*Container, error) {
	ctr1 := dag.Container().
		From("alpine:latest").
		WithExec([]string{"sh", "-c", "touch /fail && exit 42"})

	ctr2 := dag.Container().
		WithMountedDirectory("/ctr1", ctr1.Rootfs()).
		WithExec([]string{"sh", "-c", "cat /ctr1/fail"})

	return ctr2.Sync(ctx)
}
```

```console
$ dagger-dev call -i two
✘ Exec failed, attaching terminal
! process "sh -c touch /fail && exit 42" did not complete successfully: exit code: 42

/ # exit
exit 0
✔ connect 0.2s
✔ initialize 0.3s
✔ prepare 0.0s
✔ interactive: Interactive! 0.0s
✘ Interactive.two: Container! 3.1s
! call function "Two": process "/runtime" did not complete successfully: exit code: 2
┃ invoke: input: container.withMountedDirectory.withExec.sync resolve: process "sh -c touch /fail && exit 42" did not complete successfully: exit code: 42
  ✔ Container.from(address: "alpine:latest"): Container! 0.3s
  ✘ Container.sync: ContainerID! 2.7s
  ! process "sh -c touch /fail && exit 42" did not complete successfully: exit code: 42
  ┃ / # exit

Error: response from query: input: interactive.two resolve: call function "Two": process "/runtime" did not complete successfully: exit code: 2

Stdout:
invoke: input: container.withMountedDirectory.withExec.sync resolve: process "sh -c touch /fail && exit 42" did not complete successfully: exit code: 42
```

engine/buildkit/ref.go
sipsma: @aluzzardi after a rebase on main you should be able to do this:
```diff
diff --git a/engine/buildkit/ref.go b/engine/buildkit/ref.go
index e42f9c591..6542b604f 100644
--- a/engine/buildkit/ref.go
+++ b/engine/buildkit/ref.go
@@ -297,6 +297,14 @@ func wrapError(ctx context.Context, baseErr error, client *Client) error {
                return baseErr
        }
 
+       execMd, ok, err := ExecutionMetadataFromDescription(opErr.Description)
+       if err != nil {
+               // ...
+       }
+       if !ok {
+               // ... this is a valid case for containers created by buildkit internals like the dockerfile frontend
+       }
+
        // This was an exec error, we will retrieve the exec's output and include
        // it in the error message
 ```

Then we can decide whether to open the debug terminal based on that execution metadata (and use anything else or add more fields if useful for customizing the debug container; the only requirement for ExecutionMetadata is that it's json serializable).

That metadata gets initialized here: https://github.com/sipsma/dagger/blob/66e98c39ef4ed0aa17bbb2c8b47e48b10620bdaa/core/container_exec.go#L72-L72
* With some additional input for module functions specifically here: https://github.com/sipsma/dagger/blob/66e98c39ef4ed0aa17bbb2c8b47e48b10620bdaa/core/modfunc.go#L180-L180

So I guess we can decide whether to drop into the shell by default and have a field there for disabling it, or the reverse. I suppose doing it by default and just explicitly disabling for module function containers sounds better to me at first thought, but either could make sense.

**Comment by sipsma** [comment]
> As far as I can see, only the leaf container gets terminal'd:

Cool, yeah in retrospect that makes sense in the current way things work where there's just a single Solve call and thus a single error (at least as far as I can remember), thanks for testing

File: engine/buildkit/ref.go (hunk 1 above)
Comment (aluzzardi) [reply engine/buildkit/ref.go]: 🙏 

Tried something (last commit) with terminal by default (`Internal` flag on `ExecutionMetadata`), PTAL

This is pretty sweet -- I was thinking we could add `call.ID` or similar in order to be able to print the execution path (what `Terminal()` does using `idtui.DumpID`), but I didn't see an obvious path in getting that

Comment (sipsma) [reply engine/buildkit/ref.go]: Oh I actually am already adding the call ID to ExecutionMetadata in a separate PR (https://github.com/dagger/dagger/pull/7804/files#diff-6e7d0c02c8bd065fb312e4109724ee002cb041cd44c4c6e97abe267ce31cedf4R184), you can get the value from `dagql.CurrentID(ctx)`.

I think you'll also need these lines I added there so the ID json serializes (feel free to just copy-pasta): https://github.com/dagger/dagger/blob/ccdea050df7fbe7fb4237a308bdcc7016ee7337c/dagql/call/id.go#L271-L285

File: engine/buildkit/executor.go [hunk 2]
```diff
@@ -41,11 +41,14 @@ import (
 )
 
 type ExecutionMetadata struct {
+	ExecID      string
 	ClientID    string
 	SessionID   string
 	SecretToken string
 	Hostname    string
 
+	Internal bool
```
sipsma, line 50: nit: ExecID and Internal are a bit less self-evident by their names in terms of what their purpose is, can just add comments for these fields

sipsma (approved):
LGTM! Just a nit about a comment

engine/buildkit/executor.go
aluzzardi: Done!

PR #7686, opened 2024-06-18
tui: associate effects to their origin, not their unlazier

vito:
The place where something becomes un-lazied is not that useful. Better to show effects at their point of origin instead. This is a bit of :sparkles: movie magic :sparkles:  but it's a lot more useful than telling "the truth" (arguably an implementation detail of our telemetry plumbing).

This also effectively works around a Buildkit bug that ends up associating LLB to incorrect spans[^1], which would sometimes put a failed span beneath a succeeding one, making it especially hard to find.

[^1]: possibly from [here](https://github.com/moby/buildkit/blob/master/solver/jobs.go#L696) which the race detector is also unhappy with)

* [x] Bikeshed the attribute name? We overload Buildkit's `vertex` attribute, but we have the power to translate that to something like `dagger.io/effect.id`.
  * There's now a SpanProcessor that translates the attribute as described. New scheme:
    * `vertex` -> `dagger.io/effect.id`
    * `dagger.io/llb.digests` -> `dagger.io/effect.ids`
  * These attributes are set by services now too, so it makes sense to decouple from Buildkit's naming.
  * [x] Update Cloud to handle new attribute names
    * https://github.com/dagger/dagger.io/pull/3806

### DIFF
--- a/engine/buildkit/telemetry.go
+++ b/engine/buildkit/telemetry.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"strings"
 
+	"go.opentelemetry.io/otel/attribute"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/embedded"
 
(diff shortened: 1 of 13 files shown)


jedevc (approved):
Neat :tada:

engine/buildkit/telemetry.go
```diff
@@ -40,36 +42,86 @@ type buildkitTracer struct {
...
+
+	ctx, span := t.tracer.Start(ctx, spanName, opts...)
+	newSpan := buildkitSpan{Span: span, provider: t.provider}
+	return trace.ContextWithSpan(ctx, newSpan), newSpan
+}
+
+type buildkitSpan struct {
+	trace.Span
+	provider *buildkitTraceProvider
+}
```
jedevc, line 64: Aha, this is *so* much neater now that this logic isn't done entirely in `Start` :tada:

Issue #7684, opened 2024-06-17
🐞 GitHub and GitLab urls supplied as arguments for Directory assume `main` even if that's not the default branch


jpadams:
### What is the issue?

When a user provides a git url, for example as a `Directory` argument, without any branch, tag, commit sha, etc, then we assume they want `#main`.
https://github.com/dagger/dagger/blob/main/cmd/dagger/flags.go#L253-L268

But the `default_branch` might be `master` (most common alternative) or maybe something else. So we shouldn't assume.

The `default_branch` can be looked up before pulling, but requires access to the git server over the network. For example:

```
~ ➤ curl -s https://api.github.com/repos/finos/legend-engine | jq -r '.default_branch'
master
~ ➤ curl -s https://gitlab.com/api/v4/projects/jpadams301%2F\fossa | jq -r '.default_branch'
main
```

Here I ran without specifying the branch. `main` was assumed even though default is `master`.
```
dagger call legend-engine --source https://github.com/finos/legend-engine --config ./config.yaml
```
![Screenshot 2024-06-17 at 1 44 47 PM](https://github.com/dagger/dagger/assets/3187222/87da85f8-bc72-4ad8-a465-4b4e9e0c9f4e)

### Dagger version

dagger v0.11.7

PR #7688, opened 2024-06-18
fix: don't make cli default to "main" branch

jedevc:
Fixes https://github.com/dagger/dagger/issues/7684.

Previously, we didn't use to have a way to get the "default" branch - however, now we do, we can just do `.Head()` on a git repository. (note: this is just as lazy as all the other git calls).

So we don't need to assume "main" at any point.


gerhard (approved):
👍

grouville (approved):
Awesome implementation ✅

PR #7691, opened 2024-06-18
ci: use standard worker for scan-engine job

jedevc:
Minor regression from #7483 - we don't such a powerful machine for this job.


gerhard (approved):
👍 🚀

Issue #7640, opened 2024-06-12
Module versioning compatibility


jedevc:
In the upcoming v0.12 release, we plan to make a few backwards incompatible changes to the API. This includes:
- Changing function signatures
- Removal of deprecated fields
- etc.

However, these changes will *likely* require updates to module code, across the ecosystem. We want to make this transition as easy as possible - when users upgrade their cli/engine versions, they shouldn't need to worry about manually ensuring compatibility with their engines.

There are a couple of approaches possible:
- API emulation - serve the correct API to each module depending on their declared `engineVersion`.
  - pro: very smooth, very easy to upgrade
  - con: miserable for our own code, we have to support every API we create indefinitely
- Explicit breakage - have the engine explicitly reject modules that haven't explicitly upgraded to a version with breaking changes. So if a module has `v0.11.6` labelled, and the engine is running `v0.11.8` which has a `MinimumModuleVersion` of `v0.11.7` (because we introduced a backwards incompat there), then we should break loading the module.
  - pro: means we don't need to keep bug-for-bug compatibility or anything like that - if it's broken, we remove it
  - con: harder for users to upgrade potentially

I think the right approach is going to involve a feature case-by-case for each. We are going to need to break backwards compatibility sometimes to do general cleanups, etc. But also, for some things we need to minimize breakage.

I think we should do a couple things here:
- [x] Build internal scaffolding (and tests!) for having a `MinimumModuleVersion` https://github.com/dagger/dagger/pull/7692
- [x] Build internal scaffolding for serving different API versions (or performing different codegen) to different declared `engineVersion`s. https://github.com/dagger/dagger/pull/7759

Once we've done this, it's just a matter of looking through the [v0.12 milestone](https://github.com/dagger/dagger/milestone/45) and working out which one goes into which - which ones should explicitly break, and which ones are gonna have graceful fallbacks.

Prior art:
- Version compatibility between cli and engine: #5315

Previous discussions:
- https://discord.com/channels/707636530424053791/1250340115050664029

cc @gerhard @vito @sipsma @helderco @shykes

netikras:
(2¢ from a users perspective)
As long as you're hosting all the versions (i.e. available for download at any time), I don't see a problem with "Explicit breakage". If a module only works with 0.9.6, ten there's no point in upgrading the engine in the pipeline -- let it run with 0.9.6. Whenever a module developer is ready for the new engine's features, it's time for a dagger SDK's version bump and a rewrite of previous integrations.

[...]

PR #6469, opened 2024-01-22
Module engine version tracking

sipsma:
Goals:
1. Add engine version module was most recently updated with to dagger.json for the purposes of tracking as metadata
   * Currently updates on `dagger mod init/install/sync`
   * This also opens up a path to serving modules compatible schemas based on their version
     * This PR doesn't include all the internal functionality for doing so, but in the future if we make backwards incompatible changes we'll at least have the version available so we could implement this if we felt compelled.
2. Fix problem where module cache is not invalidated when engine version changes
   * This problem impacted dagger devs all the time since a change to the engine implementation often did not result in test execution being invalidated, causing us to typically have to manually clear the dev engine cache before running module tests to be sure they were actually testing our changes.
   * However, this same problem could also manifest for users if they updated their engine but ended up using local/cloud cache from older engines.
3. Publish engine builds off main to an image tag w/ their git commit rather than a global+mutable "main" tag
   * We've wanted to do this for a while since it makes it easier to just download a CLI built off main and use it with the right corresponding engine.
   * This is not *strictly* required to be here, but it felt like a good time to fix while I was in this general area
   * cc @gerhard, let me know if there's any concerns around publishing those images. We'll have a new image tag per engine commit on main, so a lot more tags than previously.
4. Remove CLI from being embedded in the engine image + in nested execs
   * This hasn't been needed for a while and may as well be removed since it's slightly increasing the size of the engine.
5. Fix problem where local imports were always retransferred in full ([details](https://github.com/dagger/dagger/pull/6469#issuecomment-1933575986))

Individual commit messages w/ more details on the above:
* [Support engine versioning by commit+local changes.](https://github.com/dagger/dagger/commit/3385f2557fce7464daf69ab24552c9aa41191db9)
* [Add engine version to ModuleConfig in dagger.json.](https://github.com/dagger/dagger/commit/f183d556550b520276b463e9cfcdd38414f05021)
* [Invalidate nested exec cache if engine version changes.](https://github.com/dagger/dagger/commit/07bf57d217e01bc906156be6bafc3a4f718e4a1b)

---

TODOs:
- [x] Do manual testing of the parts that can't be automated by setting up some release infra on my fork (i.e. test image publish off main, make sure version is correct, same for a tag, etc.)

PR #5315, opened 2023-06-15
feat: check version compatibility

TomChv:
Add checkVersionCompatibility endpoint.
Generate new clients.

Fixes DAG-1801

PR #7692, opened 2024-06-18
feat: add module versioning compat checks

jedevc:
As discussed in https://github.com/dagger/dagger/issues/7640.

This is a follow-up to https://github.com/dagger/dagger/pull/5315, which will allow us to add a backwards compatibility check if/when we make breaking changes to the module API, etc.

This uses the `engineVersion` field introduced in [`1377f13` (#6469)](https://github.com/dagger/dagger/pull/6469/commits/1377f132cfbf280c21f6e29af88451ecf8273c38).

### DIFF
--- a/core/schema/module.go
+++ b/core/schema/module.go
@@ -721,6 +721,17 @@ func (s *moduleSchema) moduleWithSource(ctx context.Context, mod *core.Module, a
 		return nil, fmt.Errorf("failed to decode module source: %w", err)
 	}
 
+	engineVersion, err := src.Self.ModuleEngineVersion(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load module config: %w", err)
+	}
+	if err := engine.CheckVersionCompatibility(engineVersion, engine.MinimumModuleVersion); err != nil {
+		return nil, fmt.Errorf("module requires incompatible engine version: %w", err)
+	}
+	if err := engine.CheckVersionCompatibility(engine.Version, engineVersion); err != nil {
+		return nil, fmt.Errorf("module requires newer engine version: %w", err)
+	}
+
 	mod = mod.Clone()
 	mod.Source = src
 	mod.NameField, err = src.Self.ModuleName(ctx)
--- a/core/integration/engine_test.go
+++ b/core/integration/engine_test.go
@@ -305,6 +305,7 @@ func (EngineSuite) TestVersionCompat(ctx context.Context, t *testctx.T) {
 		Stderr(ctx)
 	require.NoError(t, err)
 	require.Contains(t, stderr, devEngineVersion)
+	require.NotContains(t, stderr, "incompatible")
 
 	// client version is a development version
 	stderr, err = clientCtr.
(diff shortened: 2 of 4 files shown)


jedevc:
Hm, had an idea - so currently, in this PR, we have the ability to enforce a minimum version compatibility for modules that require engine versions that are too *old*.

Should we also enforce a minimum compatibility for engine versions that too *new*? e.g. if I try and run a module that has `engineVersion` v0.12.1 on an engine version v0.12.0, should we explicitly error out?

Obviously, this has the potential to introduce some more friction in upgrading, but on the plus side, we get a more useful error message about what's wrong (instead of getting weird type errors all over the place from each SDK as they encounter things that don't exist). Additionally, this constraint could hopefully motivate users to upgrade to more recent dagger versions - if a daggerverse module uses dagger v0.12.1, then a user has to upgrade to use that.

jedevc:
Right, decided to implement my above comment - if a user tries to run a v0.13.0 module on v0.12.0, this should definitely fail - it will probably not compile.

This is similar in behavior to how `go.mod` works in compilation of it's code - so shouldn't be too unexpected.

core/schema/module.go
```diff
@@ -721,6 +721,17 @@ func (s *moduleSchema) moduleWithSource(ctx context.Context, mod *core.Module, a
...
+	engineVersion, err := src.Self.ModuleEngineVersion(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load module config: %w", err)
+	}
+	if err := engine.CheckVersionCompatibility(engineVersion, engine.MinimumModuleVersion); err != nil {
+		return nil, fmt.Errorf("module requires incompatible engine version: %w", err)
+	}
+	if err := engine.CheckVersionCompatibility(engine.Version, engineVersion); err != nil {
+		return nil, fmt.Errorf("module requires newer engine version: %w", err)
+	}
```
vito, line 733: How strict is this? Feels like it could be frustrating if someone isn't enabled to directly upgrade their Dagger version.

While we're pre-1.0 this is probably closer to the reality of our releases (patch releases introduce new features all the time that the module might depend on), but I'd expect post-1.0 when our versions are more meaningful that this would be way too strict.

I guess we could codify that ahead of time? If version is < 1.0, require strict `>=`, but if >= 1.0, allow wiggle room at the patch level?

jedevc: Ooh that's a good point. Hm, I think I agree - but we can encode that constraint when we get to v1.0 right? Since the released engine will be the one doing the checks.

vito: Yeah - as long as we remember. :smile: 

Maybe we should make a v1.0 milestone for things like that.

jedevc: Yeah I'm happy with that - aside from that, any outstanding issues?

vito: Opened https://github.com/dagger/dagger/issues/7865 (+ milestone) for the v1.0 change.

I think the main question is whether this is still too strict even for today, but I don't have a great read on that.

vito (approved):
Approving in principle, have some concern over being too strict even for today, but it seems worth trying and we can always relax it later

Issue #7687, opened 2024-06-18
If you accidentally use `Boolean` as type in TypeScript SDK, you can get cryptic error


jpadams:
![image](https://github.com/dagger/dagger/assets/3187222/ce65ab19-bdc6-4687-83cc-9a493001104c)

![image](https://github.com/dagger/dagger/assets/3187222/f69b3a80-57b1-466e-9df8-9a564befff75)

remedy is to use `boolean`, but sometimes IDE will autocomplete `Boolean` and it's hard to see next to capitalized Dagger types like `Container` and `Directory`.
cc @TomChv @helderco

TomChv:
Yeah it's because `Boolean`, `String`, `Number` etc... refer to the primitive but I could handle that in the introspector to make it work even with this notation

TomChv:
<img width="1316" alt="Screenshot 2024-06-18 at 16 58 14" src="https://github.com/dagger/dagger/assets/61683879/f67e5166-a991-4a41-ab31-b4bb5aaf281e">

Funny things is that the compiler can actually complain if you use `Number`, I'm not sure what's the right thing to do

TomChv:
What I propose is to raise a specific error for that kind of issues, to suggest the user to use the correct type instead of converting the type.
WDYT @helderco ?

helderco:
> What I propose is to raise a specific error for that kind of issues, to suggest the user to use the correct type instead of converting the type. 

Yeah, that sgtm.

PR #7693, opened 2024-06-18
feat(ts-sdk): throw explicit error on invalid primitive types

TomChv:
Fixes https://github.com/dagger/dagger/issues/7687 by returning an explicit error in case the user uses primitive types in Typescript such as `String`, `Boolean` or `Number`.

The error returned will be of shape (example with `Boolean`):

```
Use of primitive Boolean type detected, did you mean boolean?
```


helderco (approved):
lgtm

PR #7696, opened 2024-06-18
tests: attempt to deflake proxy tests

sipsma:
I couldn't repro this locally but there were occasional CI flakes in the proxy test that pulls a git repo and verifies the proxy logs show a connection to github.com (needed to verify that git ops correctly use proxy settings).

My best guess is that the squid logs were just not always flushed by the time they were read for the test assertion, so this updates the tests to explicitly start/stop the squid service and do some retries on reading the logs.

Will have to see if this actually fixes the problem. If not there's more complicated options like starting a separate custom git service, intercepting traffic and asserting there. Would prefer to avoid complicating these already complicated tests further if possible though.

### DIFF
--- a/core/integration/proxy_test.go
+++ b/core/integration/proxy_test.go
@@ -233,6 +233,7 @@ http_access allow localhost
 					test.proxyLogTest(t, c, func(ctx context.Context) (string, error) {
 						return c.Container().From(alpineImage).
 							WithMountedCache("/var/log/squidaccess", squidLogsVolume).
+							WithEnvVariable("CACHEBUSTER", identity.NewID()).
 							WithExec([]string{"cat", "/var/log/squidaccess/access.log"}).
 							Stdout(ctx)
 					})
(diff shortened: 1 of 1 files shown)


core/integration/proxy_test.go
```diff
@@ -215,17 +224,18 @@ http_access allow localhost
...
 		for _, test := range tests {
 			test := test
 			if test.proxyLogTest != nil {
 				t.Run(test.name+"-proxy-logs", func(t *testing.T) {
 					t.Parallel()
-					test.proxyLogTest(t, c, squidLogs)
+					test.proxyLogTest(t, c, func(ctx context.Context) (string, error) {
+						return c.Container().From(alpineImage).
+							WithMountedCache("/var/log/squidaccess", squidLogsVolume).
+							WithExec([]string{"cat", "/var/log/squidaccess/access.log"}).
```
vito, line 236: Does this need a cache-buster for the `cat`?

sipsma: Ah yeah good point! Wouldn't be surprised if that was part/whole of the flake before, I'll keep these updates but add that in too

PR #6811, opened 2024-03-04
chore: improve dependabot config publish weekly with groups

jedevc:
Follow-up on comments from https://github.com/dagger/dagger/pull/6352#discussion_r1439349557.

Getting kinda annoyed with all the dependabot spam :cry:

- By moving from once-a-month to once-a-week, we stay more on top of changes, which means there should be fewer to address all at once. Additionally, I've set the time to trigger on Monday morning - that way, we can hopefully review and merge before the next release.
- I've also used `group`s to try and ensure that each set of updates is all combined, so we don't get a complete overflow of updates. Unfortunately, we don't seem to be able to combine all the go updates together (see https://github.com/dependabot/dependabot-core/issues/7547) - so we may still get invalid PRs, since the `go.mod`s interact with each other in fun ways.
- Finally, I've removed the `version-update:semver-patch` from Go - at least for the core engine+components, I'd like to try following patch updates as well. Some projects follow semver a little differently (even though ideally they wouldn't :laughing:), and I'd like to make sure we don't fall behind.

PR #7697, opened 2024-06-18
dependabot: Un-ignore patch bumps for SDKs

aluzzardi:
Thanks to #6811, SDK dependencies are now updated in groups (bulk). This removes the need for ignoring patch releases.


gerhard (approved):
👍 🚀

aluzzardi:
set the milestone to v0.12, we don't want to mass-update dependencies right before the v0.11.8 release

aluzzardi:
@gerhard Random CI failure FYI -- can't see the logs though

PR #7701, opened 2024-06-19
feat(ts-sdk): deprecate `@field` in favor of `@func`

TomChv:
Fixes DEV-3701

I kept the backward compatibility but added a `@deprecated` notice for users.
So it's not a breaking change until we remove the code (maybe we can save that for `v1.0.0`)?.

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -1727,7 +1727,7 @@ class Minimal:
 		{
 			sdk: "typescript",
 			source: `
-import { object, func, func } from "@dagger.io/dagger"
+import { object, func } from "@dagger.io/dagger"
 
 @object()
 class Minimal {
@@ -1843,7 +1843,7 @@ class Test:
 		},
 		{
 			sdk: "typescript",
-			source: `import { object, func, func } from "@dagger.io/dagger"
+			source: `import { object, func } from "@dagger.io/dagger"
 
 @object()
 class Test {
@@ -2063,7 +2063,7 @@ def repeater(msg: str, times: int) -> Repeater:
 		{
 			sdk: "typescript",
 			source: `
-import { object, func, func } from "@dagger.io/dagger"
+import { object, func } from "@dagger.io/dagger"
 
 @object()
 class Repeater {
@@ -2150,7 +2150,7 @@ def my_function() -> X:
 		{
 			sdk: "typescript",
 			source: `
-import { object, func, func } from "@dagger.io/dagger"
+import { object, func } from "@dagger.io/dagger"
 
 @object()
 class X {
@@ -2239,7 +2239,7 @@ class Foo:
 		{
 			sdk: "typescript",
 			source: `
-import { object, func, func } from "@dagger.io/dagger"
+import { object, func } from "@dagger.io/dagger"
 
 @object()
 class X {
@@ -2342,7 +2342,7 @@ class Playground:
 		{
 			sdk: "typescript",
 			source: `
-import { object, func, func } from "@dagger.io/dagger"
+import { object, func } from "@dagger.io/dagger"
 
 @object()
 class Bar {
@@ -2447,7 +2447,7 @@ class Playground:
 		{
 			sdk: "typescript",
 			source: `
-import { dag, Container, File, object, func, func } from "@dagger.io/dagger"
+import { dag, Container, File, object, func } from "@dagger.io/dagger"
 
 @object()
 class Foo {
@@ -2576,7 +2576,7 @@ class Playground:
 		{
 			sdk: "typescript",
 			source: `
-import { dag, Container, object, func, func } from "@dagger.io/dagger"
+import { dag, Container, object, func } from "@dagger.io/dagger"
 
 @object()
 class ScanReport {
@@ -2885,7 +2885,7 @@ class Foo:
 		{
 			sdk: "typescript",
 			source: `
-import { object, func, func } from "@dagger.io/dagger"
+import { object, func } from "@dagger.io/dagger"
 
 @object()
 class Message {
@@ -4100,7 +4100,7 @@ class Test:
 			{
 				sdk: "typescript",
 				source: `
-import { Directory, object, func, func } from '@dagger.io/dagger';
+import { Directory, object, func } from '@dagger.io/dagger';
 
 @object()
 class Test {
@@ -4528,7 +4528,7 @@ class Wrapper:
 		{
 			sdk: "typescript",
 			source: `
-import { dag, Container, object, func, func } from "@dagger.io/dagger"
+import { dag, Container, object, func } from "@dagger.io/dagger"
 
 @object()
 class WrappedContainer {
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -1586,14 +1586,14 @@ class Test:
  *
  * Long description, with full sentences.
  */
-import { object, field } from '@dagger.io/dagger'
+import { object, func } from '@dagger.io/dagger'
 
 /**
  * Test object, short description
  */
 @object()
 class Test {
-    @field()
+    @func()
     foo: string = "foo"
 }
 `,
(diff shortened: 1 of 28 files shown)


sdk/typescript/introspector/scanner/abtractions/property.ts
```diff
@@ -5,7 +5,7 @@ import { UnknownDaggerError } from "../../../common/errors/UnknownDaggerError.js
 import { FieldTypeDef, TypeDef } from "../typeDefs.js"
 import { typeToTypedef } from "../utils.js"
 
-const PROPERTY_DECORATOR = "field"
+const PROPERTY_DECORATOR = "func"
```
helderco, line 8: This isn't enough because now `@field` doesn't work.

TomChv: Oh yeah oups, thanks for catching that!

helderco:
It's good that you've updated the docs, but isn't it best to split that into another PR? Otherwise we may need to publish docs before this is released.

core/integration/module_test.go
```diff
@@ -1843,7 +1843,7 @@ class Test:
 		},
 		{
 			sdk: "typescript",
-			source: `import { object, func, field } from "@dagger.io/dagger"
+			source: `import { object, func, func } from "@dagger.io/dagger"
```
helderco, line 1846: `func` is imported twice. Happens in several places.

helderco (commented):
When testing manually, `@func` doesn't seem to work on fields.

After a `dagger init --sdk=typescript --source=deco deco`, and changing the code to:

```ts
import { object, func, field } from "@dagger.io/dagger"

@object()
class Deco {
  /**
   * The greeting to use
   */
  @field()
  greeting: string

  /**
   * Who to greet
   */
  @func()
  name: string

  constructor(
    /**
     * The greeting to use
     */
    greeting: string = "Hello",
    /**
     * Who to greet
     */
    name: string = "World",
  ) {
    this.greeting = greeting
    this.name = name
  }

  /**
   * Return the greeting message
   */
  @func()
  message(): string {
    return `${this.greeting}, ${this.name}!`
  }
}
```

This branch returns this:

```console
❯ ./hack/with-dev dagger -m deco functions
Name       Description
greeting   The greeting to use
message    Return the greeting message
```

Additionally, I'm seeing the following error in the IDE:

Screenshot 2024-06-20 at 11 17 37

TomChv:
> When testing manually, @func doesn't seem to work on fields.

Hmm that's strange because all tests passes, I'm going to have a new look on it

TomChv:
> It's good that you've updated the docs, but isn't it best to split that into another PR? Otherwise we may need to publish docs before this is released.

I can revert this, and put this into another Pr yeah

TomChv:
> When testing manually, @func doesn't seem to work on fields.

Are you sure you're testing with the right version of the branch?

I do not see your IDE error on my tests:

Screenshot 2024-06-20 at 13 54 30

Same when I do dagger functions

```shell
dagger functions
Name            Description
greeting        The greeting to use
message         Return the greeting message
name            Who to greet
```

```shell
dagger call message
Hello, World!
```

core/integration/module_test.go
TomChv: Fixed in [58b4d9d](https://github.com/dagger/dagger/pull/7701/commits/58b4d9d46f56661bbf48e4cf03ae5b4353236f6a)

Xiot:
Just out of curiousity, why are you using `@func` for fields?  
Seems a little confusing since a `string` is not a function

TomChv:
> Just out of curiousity, why are you using @func for fields?
> Seems a little confusing since a string is not a function

It's because a `field` isn't a GraphQL concept, in GQL, everything is a callable function and the Dagger API follows this logic.

When you decorate a property with `@field`, it's actually creating a function in the Dagger API to access the value of the property, so in fact, it's a function.
By deprecating `@field` in favor of `@func`, we are more consistent with what actually happens in the Dagger API.

TomChv:
> When a user writes a module, they shouldn't know (or care) about the underlying transport layer. This just seems to make the usage more confusing

It's confusing because you were used to use `@field` but it actually makes more sens to use `@func` because you're creating a function to access the property. So `@field` is actually the confusing keyword haha

Xiot:
> > When a user writes a module, they shouldn't know (or care) about the underlying transport layer. This just seems to make the usage more confusing
> 
> It's confusing because you were used to use `@field` but it actually makes more sens to use `@func` because you're creating a function to access the property. In fact`@field` is actually the confusing keyword with you think about it haha

It depends on where you are looking at it from.  As a module developer, it is a field.  The fact that its actually a function on the dagger side is almost irrelevant.

helderco:
> > It's because a field isn't a GraphQL concept, in GQL, everything is a callable function and the Dagger API follows this logic.
> > By deprecating @field in favor of @func, we are more consistent with what actually happens in the Dagger API.
> 
> Isn't that just bleeding out an implementation detail? When a user writes a module, they shouldn't know (or care) about the underlying transport layer. This just seems to make the usage more confusing

It's wrong to say that field isn't a GraphQL concept, in fact it's the opposite. @func and @field both produce a GraphQL field underneath, and GraphQL fields are turned into Dagger Functions. That's the implementation detail. When we say "Function", we're talking about a Dagger concept, not GraphQL.

Using a `@field` decorator, is what poses more confusion because it makes it seems to introduce a "Dagger Field" concept when it's all functions. Thus, using `@func` on object properties basically means that it'll expose a function in Dagger, as you can see with  `dagger functions`. It's a convenience so you don't have to create simpler getter functions to access an object's property.

helderco:
> It depends on where you are looking at it from. As a module developer, it is a field. The fact that its actually a function on the dagger side is almost irrelevant.

It's a field from which POV? Both `@field` and `@func` create a GraphQL field underneath. Different languages use different terminology here:
- TypeScript: object properties
- Python: object attributes
- Go: struct fields

helderco:
> That makes more sense. What makes it confusing on my end is that I'm just using internal modules, so I never actually see the dagger representation. 

You mean you're using a TypeScript module as a dependency so you don't see its functions with `dagger functions`, just the client bindings from codegen? Well, if you create a simple method without arguments (`@func`) and expose an object's property (`@field`), you'll notice that there won't be any difference in the client bindings from codegen. You won't be able to tell from that alone if on the "server side" they're implemented as a method or as a property.

> That combined with the fact that I need to use `@field` on anything that I want persisted, even though it is never exposed via the cli.

That doesn't seem right. You only need it to expose a function, it's not necessary for persistence, as in, to preserve values while chaining.

> As an example, in this [thread](https://discord.com/channels/707636530424053791/1249409277760765954/1249409540861067394) I need to add `@object` and `@field` to `ApplicationArgs` in order for code-gen to run successfully.

I'll need to take a loot at that.

helderco:
> As an example, in this [thread](https://discord.com/channels/707636530424053791/1249409277760765954/1249409540861067394) I need to add `@object` and `@field` to `ApplicationArgs` in order for code-gen to run successfully.

I may not quite understand the issue, but you should only need to use `@field` or `@func` if you want to expose a function on the API. It shouldn't be required just to make it run. That could be the sign of a bug or a misunderstanding.

Xiot:
> You mean you're using a TypeScript module as a dependency so you don't see its functions with dagger functions, just the client bindings from codegen? 

I just have a single `dagger.json` module, however I have a bunch of modules inside.

My dagger calls look something like
```
dagger call merge-request --iid=123 turbo build-changed
```

so I have `MotionDagger` (root) -> `MergeRequestModule` -> `TurboModule` -> `buildChanged` (func).

> That doesn't seem right. You only need it to expose a function, it's not necessary for persistence, as in, to preserve values while chaining.

If I don't add `@field` to those properties then they won't be available on the object

> I may not quite understand the issue, but you should only need to use @field or @func if you want to expose a function on the API. It shouldn't be required just to make it run. That could be the sign of a bug or a misunderstanding.

Moved to discord

helderco (approved):
:shipit:

PR #7705, opened 2024-06-19
feat: add automatic sdk cli versioning

jedevc:
:warning: depends on #8061.

This is part of a piece of a more general piece of work to improve our own releasing process.

TL;DR: our releasing process (defined in https://github.com/dagger/dagger/blob/main/RELEASING.md) has steadily become more and more complex over time. Now that we're aiming for faster release cadences (at least once every couple of weeks, and often once a week), this process is starting to incur a lot of overhead. (I estimate I probably spend at least half a day per release). We can take some of this time, and use it to automate, to save time later on.

---

This first piece attempts to resolve the weirdness around tagging. The current process (summarized) is:
- Generate engine release note and merge them
- Tag the engine as `vX.Y.Z`
- Wait for the engine to be released, auto-generating a PR with bumped cli version files ([`da92186` (#7690)](https://github.com/dagger/dagger/pull/7690/commits/da921862f96052eb599905733c61b20c8dae8dd8)).
- Generate SDK release notes, and merge the PR
- Tag the SDKs as `sdk/<sdk>/vX.Y.Z`
- Wait for the SDKs to be released

This is weird, and this dependency introduces a slowdown in our release process. Instead, with this PR, the goal is to:
- Generate engine and SDK release notes and merge them all together
- Tag the engine as `vX.Y.Z`
- Wait for the engine to be released
- Tag the SDKs as `sdk/<sdk>/vX.Y.Z`
- Wait for the SDKs to be released

There should be no extra PR steps at this stage, etc.

---

The main obstacle with this is that we need to bump the cli version for each SDK - this version number is used by each standalone SDK (i.e. not with modules) to perform automatic provisioning when a CLI has not been detected in the environment (i.e. when run with `dagger run`). This manual bump is the main reason for the PR step in the first place.

However! We can remove this step by automatically setting the version number *only on publish* - while in `main`, the version number will be unset, but when published to pypi/crates/etc, we'll set it then.

- This is very easy for the python, elixir, "just upload a tarball kind of release" - we just modify it right before publishing
- This is an incredible pain for go and php, "mirror to a git repository kind of release" - we need to rewrite history to automatically insert the right version numbers (while taking *incredible care* to make sure to avoid needlessly changing commit IDs).

---

TODO:
- [x] Implement the insane git filter script
- [x] Make version file generation simpler (by splitting them out as much as possible)
- [x] Connect the generation script into CI, reworking `Bump` and `Publish`
- [x] Select filler values for `main`
- [x] Test! Do some fake releases on a fork.


jedevc:
Decided to rework this, and removed the `version-update.py` script entirely. The original intent of this was to be able to tag *any* commit and have it release correctly. Unfortunately, this isn't really a possible goal, due to release notes, other prep, etc.

Additionally, it introduces a *significant* amount more complexity, and also since it introduces more complexity to the mirroring, it becomes harder to remove it in the future - this is essentially a one-way street. Finally, it's really really unclear how this would scale nicely into a release-branch model, e.g. tagging v1.2.3 off of a v1.2 branch.

So instead, we simplify the process - instead of doing the SDK version bumps in a separate PR, we'll handle this as part of the release notes PR. This means that we'll have one commit that we can tag with all the tags, but also means we still keep some level of manual control. There is one disadvantage to this: main will temporarily have SDKs that point to versions of the engine that do not yet exist. This should be fine though - we just need to update our provisioning tests to handle this, and we can do this. Additionally, the only SDKs that are directly consumable off of `main` are Go and PHP - this will *temporarily* place the state of `main` for them into an invalid state, however, I think this is probably worth it for the complexity that this removes from our release.

jedevc:
Linting flake is https://github.com/dagger/dagger/pull/8062.

sdk/elixir/lib/dagger/core/engine_conn.ex
```diff
@@ -95,5 +93,7 @@ defmodule Dagger.Core.EngineConn do
     Dagger.Session.stop(pid)
   end
 
-  def engine_version(), do: @dagger_cli_version
+  def engine_version() do
+    Dagger.Core.Version.engine_version()
+  end
```
wingyplus, line 98: I just notice that no one use this function. I can help removing this function after this PR get merged.

sipsma:
> So instead, we simplify the process - instead of doing the SDK version bumps in a separate PR, we'll handle this as part of the release notes PR. This means that we'll have one commit that we can tag with all the tags, but also means we still keep some level of manual control. 

I like this approach a lot. I think it's worth keeping manual checks + control in this area but agree it'd be very nice to consolidate the number of PRs involved in the overall release.

---

Thinking out loud: to avoid the possible issues around SDKs having engine versions that don't exist yet, I'm wondering if there's some way to actually publish engine+CLI versions in the PR itself in a safe way.

In theory, it seems possible to push an engine image and upload a CLI without that triggering the whole world to start automatically using those versions. Some way to "stage" them essentially, such that they are really uploaded but not officially published per-se and not yet considered the "latest" version.

Then the process would look something like:
1. Open PR for release notes + engine version bumps
2. CI publishes provisional engine+CLI for the new version
3. SDK provisioning tests run against that provisional engine+CLI
4. Once merged to main, we manually push the tags for the engine+CLI and all SDKs (same as today)
   * Tag push for engine+CLI triggers the staged publishes to be made "official" and picked up automatically when installing the "latest" version
   * Tag push for each SDK triggers them to be published (same as today)

One potential gotcha here is that by default it would allow anyone to open a PR and trigger publishes of these provisional engine+CLI versions.
* Maybe there's some way to require manual approval from a dagger maintainer
* Or maybe there's some way to structure the staging of provisional engine+CLI publishes such that even if someone did this, it wouldn't cause any actual harm. This is an obscure situation anyways, so even if it requires some manual cleanup or something I'd be okay with that.

I'm leaving the exact way you'd stage provisional engine+CLIs vague obviously, and not saying any of this is a blocker here. Just curious what others think.

jedevc:
> In theory, it seems possible to push an engine image and upload a CLI without that triggering the whole world to start automatically using those versions. Some way to "stage" them essentially, such that they are really uploaded but not officially published per-se and not yet considered the "latest" version.

I *think* we kind of have this today with publishing directly off of `main`? Or do you mean, like a full tagged build, with versions all correctly configured?

> One potential gotcha here is that by default it would allow anyone to open a PR and trigger publishes of these provisional engine+CLI versions

The problem is, I don't actually know if this is possible at all - we'd need to get secrets into a PR context, and however we do that, it's not really secure. One option instead of this would be to use manual workflow dispatch - you could trigger a staging release, since that's locked behind maintainer permissions. I'm not quite sure where the artifacts would end up though.

:thinking: I'm not 100% sure we'd want this to be part of every release process - but I can definitely see the value in doing a "staged" release for testing purposes. That said, yeah, I'm not sure we should block on this, it feels like another big piece of work on top of this.

dev/sdk.go
```diff
@@ -92,3 +96,99 @@ func (dev *DaggerDev) introspection(ctx context.Context, installer func(*dagger.
...
+			"git", "filter-branch", "-f", "--prune-empty",
+			"--subdirectory-filter", opts.sourcePath,
+			"--tree-filter", opts.sourceFilter,
+			"--", opts.sourceTag,
+		})
+	if !opts.dryRun {
+		result = result.WithExec([]string{
+			"git",
+			"push",
+			// "--force", // NOTE: disabled to avoid accidentally rewriting the history
```
gerhard, line 156: Should we remove `--force` entirely?

gerhard (approved):
This is a great improvement @jedevc!

I really like the simplification of the releasing process and skipping an extra step which was easily adding an extra hour to the total releasing time.

Very excited to see how this works in practice 👨‍🔧

sipsma (approved):
> 🤔 I'm not 100% sure we'd want this to be part of every release process - but I can definitely see the value in doing a "staged" release for testing purposes. That said, yeah, I'm not sure we should block on this, it feels like another big piece of work on top of this.

Yeah it needs more thought and like I said definitely not a blocker here.
* Maybe the publish of a staged release could be part of a locally run command (where only dagger devs have access to those). Which would align with a long term goal of modularizing the whole release and thus making it runnable locally + in CI. 🤔

Either way, big step forward in this PR! LGTM

.github/workflows/engine-and-cli-publish.yml
```diff
@@ -264,22 +241,29 @@ jobs:
...
-              RUNNER_HOST="docker-image://${{ secrets.RELEASE_DAGGER_ENGINE_IMAGE }}:${GITHUB_REF_NAME}"
+              RUNNER_HOST="docker-image://${{ vars.RELEASE_DAGGER_ENGINE_IMAGE }}:${GITHUB_REF_NAME}"
             fi
-            echo "_INTERNAL_DAGGER_TEST_CLI_URL=${ARCHIVE_URL}" >> $GITHUB_ENV
-            echo "_INTERNAL_DAGGER_TEST_CLI_CHECKSUMS_URL=${CHECKSUMS_URL}" >> $GITHUB_ENV
-            echo "_EXPERIMENTAL_DAGGER_RUNNER_HOST=${RUNNER_HOST}" >> $GITHUB_ENV
+          else
+            BASE_URL="https://dl.dagger.io/dagger"
+
+            # this is a pr, just default to main artifcats
```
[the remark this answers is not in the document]
jedevc: Oops, typo :scream: SGTM.

dev/sdk.go
jedevc: Going to leave it in - technically, if `main` or a tag is force-pushed, then we need the `--force` flag here. However, we shouldn't really do this in general.

Issue #6113, opened 2023-11-15
Allow installing modules from private Git repositories.


nipuna-perera:
### What are you trying to do?

Currently you can only `dagger install` modules hosted in public Git repositories. While this is great for OSS, it's not ideal for Enterprises that use private Git repositories like Gitlab and Github Enterprise. 

We require a way to somehow pass in/use authentication (GH PAT for example) when adding a module hosted in a private git repository.

Discord thread: https://discord.com/channels/707636530424053791/1171809848057401375

### Why is this important to you?

- Companies in highly regulated industries cannot publish modules in the open. Git repositories hosted within our Github Enterprise instance requires authentication in order to access. This is a blocker for using dagger modules.
- Personal projects that don't want to create public Git repositories aren't currently supported.

### How are you currently working around this?

No known workaround so far which makes this a high priority request for us and any others using private Git repositories to host Dagger modules.

skycaptain:
Bringing attention back to this issue now that modules have become the default. (I suppose, as the docs only mention the modules and previous cookbooks and guides have been removed). For us, this is the only obstacle preventing us from using Dagger modules at present.

Alvise88:
If i want to add a private lib in go.mod the problem arise the same

mehemken:
Just adding to the **Why is this important to you?** section.

I would like to keep my personal projects private.

No need to summon "highly regulated industry". It is just basic privacy.

Transmitt0r:
Would be a very nice feature, I'm interested in using this with modules hosted on our internal Bitbucket instance.

lpcalisi:
You could use git submodules as a temporal workaround.

Simple usage 
`````
git submodule add ssh://git@github.com/your_dagger_module.git

dagger install your_dagger_module
`````

grouville:
This is under active development at this very moment

It is tightly linked to the implementation of https://github.com/dagger/dagger/issues/7218. 

The first part is this PR: https://github.com/dagger/dagger/pull/7511, followed by another one for the directory refs, in which we unify the currently different way to reference directories and files on git remotes

This one has been unblocked recently by the Erik's work on the CA certificates that allows passing SSH sockets securely. A draft PR shall be available this week, on which feedback from the community is welcome 😁

LiHRaM:
@oblogic7
I followed the links posted by @grouville and found two PRs that haven't been merged

- https://github.com/dagger/dagger/pull/7708
- https://github.com/dagger/dagger/pull/7609

[...]

PR #7708, opened 2024-06-19
feat(core): add private git support via SSH auth

grouville:
WIP implementation of private Git support (works, just not with a generic / nice ref yet)

This is rebased on top of #7663, which adds integration tests and fixes some of the VCS logic.

#### Core changes
- Following @sipsma's work to securely isolate sockets per sessions. We can now automatically and securely load the SSH socket when the `SSH_AUTH_SOCK` env variable is set. This is the auth that will be used to access the private repo.
- File / Directory passed as CLI args do not rely on the os.env lookup anymore

Currently this has only been manually tested, but works e2e, just needs polish.

E.g. Currently accepted private repo ref
```shell
dagger_dev -m ssh://gitlab.com/dagger-modules/private/test/more/dagger-test-modules-private.git:top-level@2cb6cb4b0dba52c1e65b3ff46dd1a4a8f9a02f94 functions
```

TODO:
- Fix `HTMLURL` if only ssh ref style is possible
- Handle private ref neatly. I _think it is possible_ to consider all refs with `.git` inside as SSH refs, bind it with the `vcs.RepoRootForImportPath`. This is the current exploration. With this vision, the above ref would be: `gitlab.com/dagger-modules/private/test/more/dagger-test-modules-private.git/top-level@2cb6cb4b0dba52c1e65b3ff46dd1a4a8f9a02f94`
- Implement corresponding test suite


nipuna-perera:
Just have a clarifying question. Are you considering private repo access that doesn't use SSH also? In my company we aren't allowed to use SSH to access our Github Enterprise repos. Instead we have to use a Personal Access Token (PAT). Just want to make sure you are taking that scenario into consideration also.

grouville:
> Hey @grouville I have a clarifying question. Are you considering private repo access that doesn't use SSH also? In my company we aren't allowed to use SSH to access our Github Enterprise repos. Instead we have to use a Personal Access Token (PAT). Just want to make sure you are taking that scenario into consideration also.

Hi @nipuna-perera,
Mmmh, great remark. The first stopgap is the ssh version, the PAT version might also be implementable, but probably as a follow-up. I need to check with Erik to see his vision on the subject 🙏

jedevc:
Nice work so far @grouville :tada:

I'm not sure I think we should accept buildkit style refs here. Here's an alternative idea:

Here's a suggestion for how we could resolve this:
- We redefine module refs as `[proto://]host/path[@version]` or `[.]/path/to/module` (for local refs)
	- `proto` is optional - when empty, it defaults to `https`
- If it's `http://`, we do our cool path splitting using `HTTP`?
- If it's `ssh://`, I guess we can't do path splitting? Maybe there's something else we could do?
	- e.g. we could do this: `ssh git@github.com "git-receive-pack dagger/dagger"` - split the path at each `/`, and repeat until we find the first one that succeeds. You can poke around to see how that works.

Examples:
- `github.com/dagger/dagger` is the same as `https://github.com/dagger/dagger`, which is this repository
- `ssh://git@github.com/dagger/dagger` is this repository (but over ssh)
	- SCP-style URLs shouldn't be supported for now, they're an insane pain.

core/integration/git_test.go
grouville, line 1: Such test is necessary. As we still cannot pass Socket via the CLI in modules, this test relies on the connect to access the socket and mount it. The client implements the new loading of socket when the `SSH_AUTH_SOCK` env variable is set, which we need to ensure is not set during the test execution, as it could interfere with the state of the test

On the other side, this private key adds read only rights to the private test repository

jedevc: @grouville heads up! https://github.com/dagger/dagger/pull/7804

grouville:
### Update
Thanks @jedevc for the proposal on the ref style. This unlocked me getting and helped me realized that the scp style I was trying to implement was a rabbithole. Current state of PR is not finished, but, we made major improvements in a sense that we can already see the new ref design: https://github.com/dagger/dagger/pull/7708/files#diff-8432d1812b1f1b774598e92639e78ab8a25fcb6804617f7587d692149d155e21

Your suggestions are pretty strong:
> We redefine module refs as `[proto://]host/path[@version]` or `[.]/path/to/module` (for local refs)
> proto is optional - when empty, it defaults to git+http/git+https ✅ 

> If it's `git+http://`, we do our cool path splitting using HTTP?
If it's `git+ssh://`, I guess we can't do path splitting? Maybe there's something else we could do?
e.g. we could do this: `ssh git@github.com "git-receive-pack dagger/dagger"` - split the path at each /, and repeat until we find the first one that succeeds. 
If it's `ssh://`, we could fallback to `git+ssh://`.

The only tweak that I personally prefer is to just continue relying on vcs.RepoRootForImportPath to isolate the root of the repo and the subdir. For most providers, the root of a private repo will either be:
    1. found via the regex
    2. the provider returns the root of repo with go-get=1 even for private
       repos
    3. Last resort (GitLab case, which purposefully [obfuscates the root of the repo](https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/lib/gitlab/middleware/go.rb#L210-221)), the solution that Go uses is to be declarative: add the `.git` inside the ref

The implications of that is that we would need, for some refs, have this `.git`, but that's really how most people end up resolving the issue in Go: https://github.com/golang/go/issues/29953

> In this model, all git URLs are prefixed with git+, as suggested by @helderco in https://github.com/dagger/dagger/issues/6957#issuecomment-2183575359 - that way, we can distinguish between https://path/to/file and https://path/to/git/repo easily in the CLI.

It is perfect, and totally makes sense

#### Context
Go ref style

One of the biggest advantage of Go's ref style is its simplicity and decoupling from VCS providers. People just point the ref, and it automagically figures out how to query it.

This works quite well but it is not a 100% perfect solution ([link to Go auth issues](https://github.com/golang/go/issues/29953)).
To know if Dagger shall copy this model, we first need to list all the things to consider:

Let's deep dive on the requirements / possibilities:
1. Splitting repo ref / path: [github.com/dagger/dagger/dev/markdown](https://github.com/dagger/dagger/tree/main/dev/markdown) is a module inside the dagger/dagger monorepo. We need to have a ref format that works for both private AND public repos whilst remaining explicit (especially on the errors) AND that does not fail (GitLab / Go issue for private repos)
2. Multi VCS: currently Dagger only supports Git but there is totally a path towards multi VCS support (Bazaar, Mercurial, Subversion) -> https://github.com/grouville/go-vcs/commit/2e8fc1bb70efdc4400afe4e3b963a94c5daf2485
3. Auth protocol support: What are the protocols handled for auth. Currently, only SSH auth is supported, but we already have requests for HTTP / PAT support ; also, each of these vcs have different ways of authenticating under the hoods -> https://docs.gomods.io/configuration/authentication/#atlassian-bitbucket-and-ssh-secured-git-vcss.
4. Ref -> web view. Dagger provides link to automatically click and see the tree of the current module on the provider. Whilst accepting new formats, we need to make sure to convert the refs to a working URL
5. CLI: support passing a remote URL as file argument #6957. We need to differenciate a call to a git repo and an http call inside the CLI
-> https://github.com/dagger/dagger/issues/6957


Now, as it is a user-facing feature, its implications are big: 
1. retro-compatibility / no breaking changes
2. Good DX
3. Shell command interpretability

Currently, the design decision balances between the explicitness / easy DX and the 


Let's dive a bit deeper each point:
1. we currently rely on the vcs.RepoRootForImportPath function (taken from Google) to split the root of a repo and a path inside. This actually works well for both private and public repos on most of VCS apart for GitLab as GitLab purposefully returns a correct incorrect meta tag if we query a private repo with 

2. As shown in this [commit](https://github.com/grouville/go-vcs/commit/2e8fc1bb70efdc4400afe4e3b963a94c5daf2485), there is an easy path for us to include inside our RepoRootPath helpers to separate the root for the path. Then, the following would be to implement the equivalent of our git implementation and decouple moduleSource from the corresponding VCS.

In terms of ref design, this is also not a problem, as above proposal works too: we could copy Helder's tool that relies on `bzr+`, `svn+` / `hg+` to isolate the vcs provider explicitely. Coupled with our extension of RepoRootForImportPath, we seem to have a solid path towards a unified DX across all those providers and a quite elegant solution

3. The underlying auth protocol should not impact the ref style. What was important in such exploration is to have a style that could fit all use-cases (please add some if you miss some). Here, vcs.RepoRootForImportPath implementation above should not fail with PAT auth for example, as we rely on a declarative model for the GitLab VCS

4. As we keep relying on RepoRootForImportPath, we have "for free" the root URL, which is the same for public and private repos

5. As we rely on this declarative model for module resolution: `git+` would be consistent for CLIs file / repo refs

core/schema/modulesource.go
```diff
@@ -239,32 +260,84 @@ type buildkitClient interface {
...
+	// Fallback to local reference
+	return localParsed
+}
+
+func parseGitEndpoint(refString string) (parsedRefString, error) {
+	scheme, schemelessRef := parseScheme(refString)
+
+	// we append "ssh" to ensure that NewEndPoint properly parses the host / path and user
+	// HTTP refs are valid SSH refs. Bonus: NewEndpoint library handles correctly implicit SSH refs
+	endpoint, err := transport.NewEndpoint("ssh://" + schemelessRef)
```
jedevc, line 297: Hmmm? This is a bit confusing. Why do we try and extract an `sshusername` if the scheme wasn't for `ssh://`?

engine/buildkit/executor.go
```diff
@@ -43,10 +43,11 @@ import (
...
 type ExecutionMetadata struct {
-	ClientID    string
-	SessionID   string
-	SecretToken string
-	Hostname    string
+	ClientID          string
+	SessionID         string
+	SecretToken       string
+	Hostname          string
+	SSHAuthSocketPath string
```
jedevc, line 50: nit: this shouldn't be in this top block - those are grouped together as the most important metadata parts.

It also should have a comment explaining what it's for.

engine/client/client.go
```diff
@@ -1019,6 +1019,7 @@ func (c *Client) clientMetadata() engine.ClientMetadata {
 		CloudToken:                os.Getenv("DAGGER_CLOUD_TOKEN"),
 		DoNotTrack:                analytics.DoNotTrack(),
 		Interactive:               c.Interactive,
+		SSHAuthSocketPath:         os.Getenv("SSH_AUTH_SOCK"),
```
jedevc, line 1022: Do we need to make this an absolute path?

core/schema/modulesource_test.go
```diff
@@ -53,16 +56,133 @@ func TestParsePublicRefString(t *testing.T) {
...
+				modPath:        "github.com/shykes/daggerverse/ci",
+				kind:           core.ModuleSourceKindGit,
+				repoRoot:       &vcs.RepoRoot{Root: "github.com/shykes/daggerverse", Repo: "https://github.com/shykes/daggerverse"},
+				repoRootSubdir: "ci",
+				scheme:         core.SchemeGitSSH,
+				sshusername:    "",
+			},
+		},
+		{
+			urlStr: "ssh://github.com/shykes/daggerverse.git/ci",
```
jedevc, line 108: Is there a different between `git+ssh` and `ssh` or are they equivalent?

If they're the same, IMO we should only allow the more explicit `git+ssh`.

core/schema/modulesource_test.go
```diff
@@ -91,6 +242,37 @@ func TestParsePublicRefString(t *testing.T) {
 				kind:           core.ModuleSourceKindGit,
 				repoRoot:       &vcs.RepoRoot{Root: "bitbucket.org/test-travail/test.git", Repo: "https://bitbucket.org/test-travail/test"},
 				repoRootSubdir: "depth1",
+				scheme:         core.NoScheme,
+				sshusername:    "",
+			},
+		},
+		{
+			urlStr: "git@github.com/shykes/daggerverse.git/ci",
```
jedevc, line 250: As mentioned above - I think we should remove this for now. Implicit git urls are messy.

Also note - these are really likely to get confused with SCP-style URLs, which would be `git@github.com:shykes/daggerverse.git/ci`.

core/schema/git.go
```diff
@@ -193,15 +237,51 @@ func (s *gitSchema) tags(ctx context.Context, parent *core.GitRepository, args t
...
-		return nil, err
+	if parent.SSHAuthSocket != nil {
+		socketStore, err := parent.Query.Sockets(ctx)
+		if err == nil {
+			sockpath, cleanup, err := socketStore.MountSocket(ctx, parent.SSHAuthSocket.IDDigest)
+			if err != nil {
+				return nil, fmt.Errorf("failed to mount SSH socket: %w", err)
+			}
+			defer func() {
+				err := cleanup
```
jedevc, line 250:
```suggestion
				err := cleanup()
```

jedevc (commented):
Hey hey! Nice work on this! :tada:

I think the structure is pretty good, nice tests as well! Left a few nit-like comments, I don't think there's anything major - the only big thing is that I think we should avoid `ssh://` schemes, and not allow implicit `git@` urls, at least for now - we can always add those back later.

cc @sipsma for the socket store usage, I'm not as familiar with that.

grouville:
Implemented all the changes (thank you for the review), removed the http / https schemes as we might want to move to git+https / git+http or no scheme as we have right now

I left the Azure repo commented as an example, but unfortunately we cannot scope the SSH key to a read-only one

It is also rebased on top of https://github.com/dagger/dagger/pull/8063, which should be merge first

grouville:
Some new inputs following a discussion with our CTO: https://github.com/dagger/dagger/issues/6113#issuecomment-2264712473 and some feedback from users

grouville:
This is again ready for review 🙏 cc @jedevc

jedevc (approved):
Yeah, this is neat work :tada: very cool :tada:

I've left a few small nits/comments/etc - only one large piece about not *removing* `cloneURL` but *deprecating* it. That should massively help the daggerverse process, as well as avoid breakages from anyone else who *might* be using this API (though I don't think that's anyone - better to be safe than sorry!).

I've approved though - once you've resolved the comments, and tests are passing, feel free to merge whenever :shipit:

docs/docs-graphql/schema.graphqls
```diff
@@ -1598,6 +1598,11 @@ type GitModuleSource {
   """A unique identifier for this GitModuleSource."""
   id: GitModuleSourceID!
 
+  """
+  The URL to access the web view of the repository (e.g., GitHub, GitLab, Bitbucket)
+  """
+  repositoryUrl: String!
```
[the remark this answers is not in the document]
grouville: You're right, it is a bit confusing, but:
- htmlURL is the base URL to access a file on a Git provider. In order to do so, it constructs opinionated URLs (with `src`, `tree`..)
- repositoryURL is the base URL to that Git provider, raw

Daggerverse currently requires on the user ref, when publishing a module, to transform all relative paths inside the Markdown (functions, object documentations / Readme [in the past]). For that, it needs another format of URL, to access the raw content for each provider: https://raw.githubusercontent.com/dagger/dagger/6249cc93f9467434215d155e6e768431e507feb7/core/integration/git_test.go, which can be accessed originally as https://github.com/dagger/dagger/raw/6249cc93f9467434215d155e6e768431e507feb7/core/integration/git_test.go

I am totally fine in hiding it, and have a function on Daggerverse's side to parse `HTMLURL()` and use that ; having this field as public could lead to some more painful changes in the future. So ok for the small overhead added internally 👌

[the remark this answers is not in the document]
jedevc: Ah, right. I was missing that context, my bad :tada:

Yeah, it's absolutely fine to have this, daggerverse shouldn't have to parse `HTMLUrl`, that's very annoying! But yeah, let's do the suggestions from above: it should be called `htmlRepoURL` instead.

PR #7710, opened 2024-06-20
ci: run trivy on source directory

jedevc:
This should hopefully improve visibility of new vulnerabilities.

cc @gerhard @samalba

### DIFF
--- a/.dagger/engine.go
+++ b/.dagger/engine.go
@@ -366,12 +366,7 @@ func (e *Engine) Publish(
 	return nil
 }
 
-func (e *Engine) Scan(ctx context.Context) (string, error) {
-	target, err := e.Container(ctx, "", nil, false)
-	if err != nil {
-		return "", err
-	}
-
+func (e *Engine) Scan(ctx context.Context) error {
 	ignoreFiles := dag.Directory().WithDirectory("/", e.Dagger.Source(), dagger.DirectoryWithDirectoryOpts{
 		Include: []string{
 			".trivyignore",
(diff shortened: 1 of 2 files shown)


ci/engine.go
```diff
@@ -276,28 +271,75 @@ func (e *Engine) Scan(ctx context.Context) (string, error) {
...
+		}
+		if len(ignoreFileNames) > 0 {
+			args = append(args, "--ignorefile=/mnt/ignores/"+ignoreFileNames[0])
+		}
+		args = append(args, "/mnt/src")
+
+		// HACK: filter out npm examples :)
+		src := e.Dagger.Source
+		src = src.WithoutDirectory("sdk/rust/crates/dagger-sdk/examples")
+		src = src.WithoutDirectory("examples/sdk/nodejs")
```
jedevc, line 303: @TomChv any ideas about this? :cry: 

I genuinely cannot seem to find a way to get this into a good state right now - it seems like the `ws` dependency is vulnerable - and `websocket-stream` (somewhere in our dep tree) uses a horrendously old version that hasn't got a backport (and of course, looks like `websocket-stream` is abandoned).

IMO I think it's totally reasonable to ignore vulnerabilities in the non-dagger example code, but I wonder if we would want to rework our examples so we don't have this problem in the first place.

jedevc: FYI, this same issue is affecting the `docs` directory.

jedevc: I've expanded this selection to exclude `docs` for now.

Not that those aren't issues, but IMO the `engine scan` pipeline should scan for issues in the engine/sdks/etc - not necessarily the examples and docs, and such.

gerhard (approved):
This got lost in my backlog.

Approving optimistically. OK to merge if we rebase & all checks pass 👍

ci/engine.go
TomChv: That's a good point! I agree that we should change our example to not use outdated dependencies
@vikram-dagger, could you work on this if you have some time?

TomChv (approved):
Ok to merge too after rebase

jedevc:
Scan is now succeeding after fixing the java dep, and rebasing, so merging now :tada:

PR #7714, opened 2024-06-20
docs: update ts docs field to `@func`

TomChv:
Follows up #7701 with a doc update

### DIFF
--- a/docs/current_docs/manuals/developer/snippets/state-functions/typescript/index.ts
+++ b/docs/current_docs/manuals/developer/snippets/state-functions/typescript/index.ts
@@ -21,7 +21,7 @@ class MyModule {
     /**
      * Who to greet
      */
-    name: string = "World"
+    name: string = "World",
   ) {
     this.greeting = greeting
     this.name = name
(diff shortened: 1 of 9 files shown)


docs/current_docs/manuals/developer/snippets/state-functions/typescript/index.ts
```diff
@@ -1,11 +1,11 @@
-import { dag, Container, object, func, field } from "@dagger.io/dagger"
+import { object, func } from "@dagger.io/dagger"
```
gerhard, line 1: Was `dag` & `Container` redundant here?

gerhard (approved):
Apart from that one comment, everything else looks good to me. WDYT @helderco ?

docs/current_docs/manuals/developer/snippets/state-functions/typescript/index.ts
TomChv: It was not used.

helderco:
This should only be merged after https://github.com/dagger/dagger/pull/7701 is released. By that time there could be more doc snippets to update. That's why I'm holding off review until then.

jedevc:
@helderco should we take this and re-deploy docs?

PR #7715, opened 2024-06-21
feat: don't generate license if not explicitly asked

TomChv:
Also fix invalid tests.

Fixes #7703

PR #7719, opened 2024-06-21
fix: dagger develop generate license only if --sdk is set

### LINKED ISSUE #7703: 🐞 `dagger develop` creates a LICENSE file
### What is the issue?

`dagger develop` creates a LICENSE file. That's not something I want in private projects. (Another instance of contextual vs reusable modules)

### Dagger version

dagger v0.11.8 (registry.dagger.io/engine) darwin/arm64

### Steps to reproduce

`dagger develop`

### Log output

_No response_

TomChv:
A `LICENSE` will only be generated now if `--sdk` is set on `dagger develop`, like in `dagger init`. It's also possible to explictly disable creating a license by providing an empty `--license=`.

Fixes #7703

It will be followed up by #7715 after discussion

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -163,10 +163,11 @@ The "--source" flag allows controlling the directory in which the actual module 
 				return fmt.Errorf("failed to generate code: %w", err)
 			}
 
-			if sdk != "" && licenseID != "" {
+			if sdk != "" {
 				// If we're generating code by setting a SDK, we should also generate a license
 				// if it doesn't already exists.
-				if err := findOrCreateLicense(ctx, modConf.LocalRootSourcePath); err != nil {
+				searchExisting := !cmd.Flags().Lookup("license").Changed
+				if err := findOrCreateLicense(ctx, modConf.LocalRootSourcePath, searchExisting); err != nil {
 					return err
 				}
 			}
@@ -382,8 +383,9 @@ If not updating source or SDK, this is only required for IDE auto-completion/LSP
 			}
 
 			// If no license has been created yet, and SDK is set, we should create one.
-			if developSDK != "" && licenseID != "" {
-				if err := findOrCreateLicense(ctx, modConf.LocalRootSourcePath); err != nil {
+			if developSDK != "" {
+				searchExisting := !cmd.Flags().Lookup("license").Changed
+				if err := findOrCreateLicense(ctx, modConf.LocalRootSourcePath, searchExisting); err != nil {
 					return err
 				}
 			}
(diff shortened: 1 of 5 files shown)


cmd/dagger/module.go
```diff
@@ -380,9 +380,11 @@ If not updating source or SDK, this is only required for IDE auto-completion/LSP
 				return fmt.Errorf("failed to generate code: %w", err)
 			}
 
-			// If no license has been created yet, we should create one.
-			if err := findOrCreateLicense(ctx, modConf.LocalRootSourcePath); err != nil {
-				return err
+			// If no license has been created yet, and SDK is set, we should create one.
+			if developSDK != "" {
```
helderco, line 384: I suggest setting the default value for the flag to be `defaultLicense` (in `licenses.go`), instead of implicitly using the default on an empty (`""`) value. This way you can check for empty license here as well, and not create it if empty:

```suggestion
			if developSDK != "" && licenseID != "" {
```

This way it's clearer when looking at `--help` what the default is, and it also allows to opt-out of creating the file (`--license ""`) as today it's not possible.

Note that there's another bug that you can fix right now. The `--license` flag is not defined on `dagger develop` like it is in `dagger init`. So in fact users can't choose any license atm.

TomChv: > (--license "") as today it's not possible. Yes but that means if users already has a LICENSE at the top dir, it's going to create a license too if it's not set to `""`.
Should I also update this part of the code to not create a license no matter what if it already exist in the repo?

helderco: Of course, rather than check for `""` you check if it's the default or not. That preserves current behavior in other parts of the code.

core/integration/module_test.go
```diff
@@ -479,6 +479,20 @@ func TestModuleInitLICENSE(t *testing.T) {
...
+
+		c, ctx := connect(t)
+
+		modGen := c.Container().From(golangImage).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work").
+			With(daggerExec("init", "--name=empty-license", "--sdk=go", "--license="))
+
+		_, err := modGen.File("LICENSE").Contents(ctx)
+		require.Error(t, err)
```
helderco, line 493: I suggest looking for the file with perhaps `Entries`. Checking for any error on `Contents` can lead to false positives.

cmd/dagger/module.go
TomChv: But what if the user's want the license `Apache-2.0` ?? and set `license="Apache-2.0"`, it's not going to work as expected and will not create it if I test against the default.

TomChv: I'm going the push as you requested but I'm not sure that's the right solution, instead if would be better to check if the flag has been explicitely set so we know if we generate a license (like in #7715)

helderco: > But what if the user wants the license `Apache-2.0` ?? and set `license="Apache-2.0"`, it's not going to work as expected and will not create it if I test against the default.

You can check if the flag is set or not. You don't have to depend on the value alone.

helderco: Afaict, if you explicitly set `--license=Apache-2.0` with a license in a parent directory, it wouldn't create it. While if the value is different from the default such as `--license=MIT` then it would. So I added a check to search parent dirs only if flag not set, independently of the value.

helderco (approved):
LGTM 🚀

helderco:
Test fail fluke. Re-ran the action in GH, which turned green, but corresponding dagger cloud trace stayed the same. Good to merge.

PR #7720, opened 2024-06-21
Chore: improve how we lint

shykes:
- Define a new linting framework, with common types for producing and querying lint reports: `LintReport` and `LintIssue`. (`lint.go`)
- Integrate our Python linter, Ruff, into the new framework (`lint_python.go`)
- Integrate our Go linter, golangci-lint, into the new framework (`lint_go.go`)
- Integrate our Typescript linter, eslint, into the new framework (`lint_typescript.go`)
- Start using the new implementations in some of our CI pipelines, but not all. Note, backwards compatibility is preserved (`sdk_typescript.go`, `sdk_python.go`, `engine.go`)

### DIFF
--- a/ci/lint.go
+++ b/ci/lint.go
@@ -1,0 +1,106 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"sync"
+)
+
+const (
+	golangCiLintImage = "docker.io/golangci/golangci-lint@sha256:b5f8712114561f1e2fbe74d04ed07ddfd992768705033a6251f3c7b848eac38e"
+)
+
+// A linting report
+type LintReport struct {
+	mu     sync.Mutex
+	Issues []LintIssue
+	// +private
+	LLReport *File `json:"-"`
+}
+
+// An individual linting issue
+type LintIssue struct {
+	// The name of the tool that produced the issue
+	Tool string
+	// True if the issue is an error, false if it's a warning'
+	IsError bool
+	// The text explaining the issue
+	Text string
+	// FIXME add more fields
+}
+
+// Return the linting report as a JSON file
+func (lr *LintReport) JSON(
+	ctx context.Context,
+	// Return the low-level linting tool's report instead of the high-level one
+	// +optional
+	ll bool,
+) (*File, error) {
+	if ll {
+		if lr.LLReport == nil {
+			return nil, fmt.Errorf("no low-level report available")
+		}
+		return lr.LLReport, nil
+	}
+	data, err := json.MarshalIndent(lr, "", "  ")
+	if err != nil {
+		return nil, err
+	}
+	f := dag.
+		Directory().
+		WithNewFile("lint.json", string(data)).
+		File("lint.json")
+	return f, nil
+}
+
+// Return the total number of linting errors
+func (lr *LintReport) ErrorCount() int {
+	var count int
+	for _, issue := range lr.Issues {
+		if issue.IsError {
+			count += 1
+		}
+	}
+	return count
+}
+
+// Return the total number of linting warnings
+func (lr *LintReport) WarningCount() int {
+	var count int
+	for _, issue := range lr.Issues {
+		if !issue.IsError {
+			count += 1
+		}
+	}
+	return count
+}
+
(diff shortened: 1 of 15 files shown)


ci/lint_python.go
```diff
@@ -0,0 +1,101 @@
...
+		//		WithExec([]string{"python", "-m", "venv", "/opt/venv"}).
+		//		WithEnvVariable("VIRTUAL_ENV", "/opt/venv").
+		//		WithEnvVariable(
+		//			"PATH",
+		//			"$VIRTUAL_ENV/bin:$PATH",
+		//			dagger.ContainerWithEnvVariableOpts{Expand: true},
+		//		).
+		//		WithEnvVariable("HATCH_ENV_TYPE_VIRTUAL_PATH", "/opt/venv").
+		//		WithMountedFile("requirements.txt", source.File("requirements.txt")).
+		//		WithExec([]string{"pip", "install", "-r", "requirements.txt"}).
```
jedevc, line 48: nit: commented out code shouldn't be merged (at least without a comment as to why)

ci/lint.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"sync"
+)
+
+const (
+	golangCiLintImage = "docker.io/golangci/golangci-lint@sha256:b5f8712114561f1e2fbe74d04ed07ddfd992768705033a6251f3c7b848eac38e"
```
jedevc, line 11: nit: we should also include the tag this in here so it's clear which version this is: `docker.io/golangci/golangci-lint:<version>@sha256:<hash>`

ci/sdk_typescript.go
```diff
@@ -29,51 +27,112 @@ type TypescriptSDK struct {
...
+	//	docsReport, err := new(TypescriptLint).
+	//		// FIXME: why not use eslint.config.js in the docs directory?
+	//		WithConfig(clientSource.File("eslint-docs.config.js")).
+	//		Lint(ctx, docs)
+	//	if err != nil {
+	//		return err
+	//	}
+	//	return report.merge(docsReport)
+	//})
+	// Check that generated client library is up-to-date
```
jedevc, line 115: :eyes: this disables a check?

ci/sdk_go.go
```diff
@@ -7,26 +7,21 @@ import (
...
 type GoSDK struct {
 	Dagger *Dagger // +private
 }
 
-// Lint the Go SDK
+// Lint the Go SDK source code
 func (t GoSDK) Lint(ctx context.Context) error {
-	eg, ctx := errgroup.WithContext(ctx)
-	eg.Go(func() error {
-		return t.Dagger.Go().Lint(ctx, []string{"sdk/go"}, false)
```
jedevc, line 24: Okay, so we're not using this module for go linting anymore.

What's the end game for this? Is the idea to do all the linting in each module for each SDK? Which is a bit less trivial because of the interfaces? (but should still be possible today).

If that's the case, should we remove the linting functionality from the `ci/std/go` module for the time-being? I'd rather not have duplicate code, one of which isn't used (makes grepping through code, etc, much more confusing).

ci/main.go
```diff
@@ -68,6 +70,36 @@ func (ci *Dagger) Check(ctx context.Context) error {
...
+	// Find all dagger modules in the source
+	src := ci.Source
+	daggerJSONs, err := src.Glob(ctx, "**/dagger.json")
+	if err != nil {
+		return nil, err
+	}
+	// FIXME: parallelize
+	for _, daggerJSON := range daggerJSONs {
+		// Skip test data, which contains test modules
+		if strings.HasPrefix(daggerJSON, "core/integration/testdata") {
```
jedevc, line 84: We should also skip `docs` prefixes.

jedevc (commented):
I like this :tada:

- Left a couple comments of things to tidy up, nothing major for now, let's prioritize iterating in `main` instead on branches :tada:
- The linter is failing in CI, the failure is legitimate it seems.
- As part of this can be move `linters/markdown` into `ci`? I don't think it makes to sense to have it at the top-level anymore.

shykes:
Thanks for the review. I meant for this to be a draft - making that change now.

PR #7724, opened 2024-06-24
docs: Fix uv flag in language dependencies manual.

KGB33:
I ran into the following error trying to regenerate the `requirements.lock` file as described [here](https://docs.dagger.io/manuals/developer/language-dependencies#lock-file) (under the Python tab). I think `--update-all` should be `--upgrade`.

Additionally, it might be worth adding the `--quiet` flag to the docs in the same vein as #7270.

```
❯ uv version
uv 0.2.13

❯ uv pip compile \
    --generate-hashes \
    --update-all \
    -o requirements.lock \
    pyproject.toml sdk/pyproject.toml
    
error: unexpected argument '--update-all' found

  tip: to pass '--update-all' as a value, use '-- --update-all'

Usage: uv pip compile --generate-hashes <SRC_FILE>...

For more information, try '--help'.
```


helderco (approved):
Thanks @KGB33!

PR #7736, opened 2024-06-24
feat: add explicit stop to dagger process

kjuulh:
Add explicit cleanup for the dagger child process if started from the rust dagger sdk itself. This requires a breaking change to the connect, as we're missing a few features from rust that are currently only available on nightly.

BREAKING: change for the dagger sdk:

```rust

let client = dagger_sdk::connect().await?;
// use client

// breaking change to:

dagger_sdk::connect(|client| async move {
  // use client

  Ok(())
}).await?;
```

The reason for the breaking change is that we
didn't properly clean up the dagger session.

I initialy attempted an async drop functionality.

However it didn't work because async drop is nightly.
With a single threaded runtime we'd deadlock the thread.

Unlike sqlx which has the same behavior, I cannot count
on the dagger session being available after the last call.

Sqlx spawns a background thread to cleanup, usually dagger_sdk is the
only thing called and as such it will never get a chance to actually
cleanup

We had 3 options:

1. Require multithreaded runtimes such that a
   homegrown async drop could be implemented.
   This was too buggy, and would lead to unintuitive usage
2. Require nightly for async drop functionality.
   async drop is fairly new so I didn't want to be hooked on that feature
   I don't want the dagger-sdk to only be available on nightly
3. Implement a cleanup closure, this is what is done in many such
   tasks, it does lead to some not so nice ergonomics, but this is
   what I accepted do to it being a known pattern.

### DIFF
--- a/sdk/rust/crates/dagger-sdk/src/core/cli_session.rs
+++ b/sdk/rust/crates/dagger-sdk/src/core/cli_session.rs
@@ -1,5 +1,6 @@
 use std::{fs::canonicalize, path::Path, process::Stdio, sync::Arc};
 
+use eyre::Context;
 use tokio::{io::AsyncBufReadExt, sync::broadcast};
 
 use crate::core::{config::Config, connect_params::ConnectParams};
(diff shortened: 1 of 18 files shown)


sdk/rust/crates/dagger-sdk/src/core/cli_session.rs
```diff
@@ -19,27 +25,54 @@ impl CliSession {
...
+}
+
+impl Drop for DaggerSessionProc {
+    fn drop(&mut self) {
+        // FIXME: This is a bit of a hack, we spawn a neutral runtime, which doesn't automatically prune blocking tasks.
+        // This should be removed when async_drop is made stable
+        futures::executor::block_on(async move {
+            tracing::trace!("waiting for dagger subprocess to shutdown");
+
+            if let Err(e) = self.inner.wait().await {
```
vito, line 45: As explained to me in Discord, this `.wait()` also does the job of closing `stdin`:

https://docs.rs/tokio/latest/tokio/process/struct.Child.html#method.wait

> The stdin handle to the child process, if any, will be closed before waiting. This helps avoid deadlock: it ensures that the child does not block waiting for input from the parent, while the parent waits for the child to exit.

vito (approved):
Approving in principle, will just wait for CI to be :green_apple:

PR #7740, opened 2024-06-24
fix(ts-sdk): correctly resolve non-primitive default value

TomChv:
There was an issue that created an error when a non-primitive default value was set to a function like

```typescript
@object()
class Default {
  @func()
  a(
    dir: Directory = dag.directory()
      .withNewFile("/foo", "hello world")
    ): Container {
    return dag.container().withMountedDirectory("/dir", dir)
  }
}
```

This commit fixes the issue to ignore the default value registration if it's not a primitive type and let the TypeScript runtime resolve it during the execution.

It's a workaround to accomplish the result because resolving the value during the registration requires to run the TS code in a custom runtime that must contains all user's dependency which may be resource consumming + complex to resolve.

### DIFF
--- a/sdk/typescript/entrypoint/register.ts
+++ b/sdk/typescript/entrypoint/register.ts
@@ -114,13 +114,22 @@ function addArg(args: Arguments): (fct: Function_) => Function_ {
         description: arg.description,
       }
 
-      if (arg.defaultValue) {
-        opts.defaultValue = arg.defaultValue as string & { __JSON: never }
-      }
-
       let typeDef = addTypeDef(arg.type)
       if (arg.isOptional) {
         typeDef = typeDef.withOptional(true)
+      }
+
+      // We do not set the default value if it's not a primitive type, we let typescript
+      // resolves the default value during the runtime instead.
+      // If it has a default value but is not primitive, we set the value as optional
+      // to workaround the fact that the API isn't aware of the default value and will
+      // expect it to be set as required input.
+      if (arg.defaultValue) {
+        if (isPrimitiveType(arg.type)) {
+          opts.defaultValue = arg.defaultValue as string & { __JSON: never }
+        } else {
+          typeDef = typeDef.withOptional(true)
+        }
       }
 
       fct = fct.withArg(arg.name, typeDef, opts)
@@ -155,4 +164,13 @@ function addTypeDef(type: ScannerTypeDef<TypeDefKind>): TypeDef {
     default:
       return dag.typeDef().withKind(type.kind)
   }
+}
+
+function isPrimitiveType(type: ScannerTypeDef<TypeDefKind>): boolean {
+  return (
+    type.kind === TypeDefKind.BooleanKind ||
+    type.kind === TypeDefKind.IntegerKind ||
+    type.kind === TypeDefKind.StringKind ||
+    type.kind === TypeDefKind.EnumKind
+  )
 }
--- a/sdk/typescript/entrypoint/register.ts
+++ b/sdk/typescript/entrypoint/register.ts
@@ -119,8 +119,8 @@ function addArg(args: Arguments): (fct: Function_) => Function_ {
         typeDef = typeDef.withOptional(true)
       }
 
-      // We do not set the default value if it's not a primitive type, we let typescript
-      // resolves the default value during the runtime instead.
+      // We do not set the default value if it's not a primitive type, we let TypeScript
+      // resolve the default value during the runtime instead.
       // If it has a default value but is not primitive, we set the value as optional
       // to workaround the fact that the API isn't aware of the default value and will
       // expect it to be set as required input.
--- a/sdk/typescript/introspector/scanner/abtractions/argument.ts
+++ b/sdk/typescript/introspector/scanner/abtractions/argument.ts
@@ -162,6 +162,14 @@ export class Argument {
     return this.formatDefaultValue(this.param.initializer.getText())
   }
 
+  /**
+   * Return true if the parameter is optional.
+   *
+   * A parameter is considered optional if he fits one of the following:
+   * - It has a question token (e.g. `foo?: <type>`).
+   * - It's variadic (e.g. `...foo: <type>[]`).
+   * - It's nullable (e.g. `foo: <type> | null`).
+   */
   private loadIsOptional(): boolean {
     return (
       this.param.questionToken !== undefined ||
--- a/sdk/typescript/introspector/scanner/abtractions/argument.ts
+++ b/sdk/typescript/introspector/scanner/abtractions/argument.ts
@@ -165,7 +165,7 @@ export class Argument {
   /**
    * Return true if the parameter is optional.
    *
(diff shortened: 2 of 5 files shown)


jedevc:
Needs a changelog entry.

helderco:
> It's a workaround to accomplish the result because resolving the value during the registration requires to run the TS code in a custom runtime that must contains all user's dependency which may be resource consumming + complex to resolve.

I didn't get this. Doesn't registration and execution run in the same runtime, with the same access to the API? Why the need for a custom runtime?

sdk/typescript/introspector/scanner/abtractions/argument.ts
```diff
@@ -162,6 +162,14 @@ export class Argument {
     return this.formatDefaultValue(this.param.initializer.getText())
   }
 
+  /**
+   * Return true if the parameter is optional.
+   *
+   * A parameter is considered optional if he fits one of the following:
```
helderco, line 168: Was just correcting the "he", but the rest can be removed:
```suggestion
   * A parameter is considered optional if:
```

sdk/typescript/entrypoint/register.ts
```diff
@@ -114,15 +114,24 @@ function addArg(args: Arguments): (fct: Function_) => Function_ {
...
-        opts.defaultValue = arg.defaultValue as string & { __JSON: never }
-      }
-
       let typeDef = addTypeDef(arg.type)
       if (arg.isOptional) {
         typeDef = typeDef.withOptional(true)
       }
 
+      // We do not set the default value if it's not a primitive type, we let typescript
+      // resolves the default value during the runtime instead.
```
helderco, line 123:
```suggestion
      // We do not set the default value if it's not a primitive type, we let TypeScript
      // resolve the default value during the runtime instead.
```

sdk/typescript/entrypoint/register.ts
```diff
@@ -114,15 +114,24 @@ function addArg(args: Arguments): (fct: Function_) => Function_ {
...
+      // If it has a default value but is not primitive, we set the value as optional
+      // to workaround the fact that the API isn't aware of the default value and will
+      // expect it to be set as required input.
+      if (arg.defaultValue) {
+        if (isPrimitiveType(arg.type)) {
```
helderco, line 128: Another way to do this is to try to serialize to JSON, and run the `else` block on the exception handling.

TomChv:
> I didn't get this. Doesn't registration and execution run in the same runtime, with the same access to the API? Why the need for a custom runtime?

No it doesn't, the registration isn't executing the code, it's only doing a static analysis

sdk/typescript/entrypoint/register.ts
TomChv: I think it's less clear than checking the type

TomChv:
> Needs a changelog entry.

I added the changelog entry

TomChv:
I don't understand this workflow: https://github.com/dagger/dagger/actions/runs/9666769666/job/26667052999?pr=7740

My PR has a changelog :O

sdk/typescript/entrypoint/register.ts
helderco: Yes, but the problem is this is more limiting.

helderco:
> > I didn't get this. Doesn't registration and execution run in the same runtime, with the same access to the API? Why the need for a custom runtime?
> 
> No it doesn't, the registration isn't executing the code, it's only doing a static analysis

By "registration" it's implied to be sending the Module ID with the typedefs to the API. At that point you are executing TS code and have access to the API.

TomChv:
> By "registration" it's implied to be sending the Module ID with the typedefs to the API. At that point you are executing TS code and have access to the API.

Yes but the default value is encoded as a string, so I would need to create an equivalent so I can call the Dagger API, but what if the code is more complex than that and call an external function?
There's something in the compiler API to execute TS code but it doesn't include dependencies, and we cannot know what the default func needs to have in order to be executed, if it was loaded as an actual code instance, that would be great but it's not the case

helderco:
You don't have to have the value during scan, if you can save just enough that can be used later when talking to the API. I also wonder if there's anything else from the introspection stuff that can get you something better than just the string representation. There's a lot of methods in that introspection data model from TypeScript.

TomChv:
> I also wonder if there's anything else from the introspection stuff that can get you something better than just the string "representation". There's a lot of methods in that introspection data model from TypeScript.

Based on Chat GPT

> To get the value of a default argument as its real type using the TypeScript Compiler API, you need to evaluate the initializer expression rather than just convert its type to a string. This involves interpreting the AST node corresponding to the default value.

> ```typescript
>      function evaluateExpression(node: ts.Expression): any {
>        if (ts.isStringLiteral(node)) {
>            return node.text;
>        } else if (ts.isNumericLiteral(node)) {
>            return parseFloat(node.text);
>        } else if (ts.isBooleanLiteral(node)) {
>            return node.kind === ts.SyntaxKind.TrueKeyword;
>        } else if (ts.isArrayLiteralExpression(node)) {
>            return node.elements.map(evaluateExpression);
>        } else if (ts.isObjectLiteralExpression(node)) {
>            const result: { [key: string]: any } = {};
>            node.properties.forEach(prop => {
>                if (ts.isPropertyAssignment(prop) && ts.isIdentifier(prop.name)) {
>                    result[prop.name.text] = evaluateExpression(prop.initializer);
>                }
>            });
>            return result;
>        } else if (ts.isFunctionExpression(node) || ts.isArrowFunction(node)) {
>            // Convert the function expression to JavaScript code
>            const functionCode = node.getText();
>            // Evaluate the function in a safe environment
>            return evaluateFunction(functionCode);
>        } else if (ts.isIdentifier(node)) {
>            // Handle identifiers (e.g., variable names)
>            const symbol = checker.getSymbolAtLocation(node);
>            if (symbol && symbol.valueDeclaration) {
>                const declaration = symbol.valueDeclaration;
>                if (ts.isVariableDeclaration(declaration) && declaration.initializer) {
>                    return evaluateExpression(declaration.initializer);
>                }
>            }
>        }
>        return undefined;
>    }
> ```

> ```typescript
>     function evaluateFunction(functionCode: string): any {
>        // Use a new Function to safely evaluate the function code
>        try {
>            const func = new Function(`return (${functionCode})`)();
>            return func();
>        } catch (error) {
>            console.error('Error evaluating function:', error);
>            return undefined;
>        }
>    }
> ```

I can try to see if I can execute the actual function but I'm not sure it will work if it's using external dependencies.

helderco:
I wouldn't try to evaluate the string based expression outside of the runtime process. What I was wondering is if the TypeScript Compiler API could have enough information to would allow you you piece it together at runtime.

However, I'd explore decorators a bit more first. Moving off of the experimental decorators, and maybe using a third-party library for reflection, you could get the standard value that way, directly within the runtime process. It could even make the compiler sub-process not necessary perhaps.

helderco:
To be clear, I'm not suggesting doing this now. Ok to leave things as is.

PR #7742, opened 2024-06-24
core(git): add git tags API

grouville:
This is the draft implementation to support a new `git.Tags` API, with corresponding tests, which enables us to retrieve the tags of a remote.

It has been mostly adapted from https://github.com/dagger/dagger/pull/6187/commits/260bde5070780b3ce267bde79de3a66c73eca6a6 (from @vito)

**TODO**
- find a way to exec git with the mounted socket / use the gitdns.git implementation ?

#### Questions
cc @sipsma, I am a bit stuck on the auth part for this API. Does it have to live inside the core/schema implementation ? If yes, how to use the mounted auth socket inside the core/schema package ?

You referenced privately, and in the past, this BuildKit call: https://github.com/sipsma/dagger/blob/64f4aaa2f794f4cb92ae0a833b61a953bc63df6f/engine/sources/gitdns/source.go#L286-L286, but there is also a wrapper that exists inside the gitdns package: newGitCLI.

### DIFF
--- a/core/schema/git.go
+++ b/core/schema/git.go
@@ -45,6 +45,9 @@ func (s *gitSchema) Install() {
 		dagql.Func("tag", s.tag).
 			Doc(`Returns details of a tag.`).
 			ArgDoc("name", `Tag's name (e.g., "v0.3.9").`),
+		dagql.Func("tags", s.tags).
+			Doc(`tags that match any of the given glob patterns.`).
+			ArgDoc("patterns", `Glob patterns (e.g., "refs/tags/v*").`),
 		dagql.Func("commit", s.commit).
 			Doc(`Returns details of a commit.`).
 			// TODO: id is normally a reserved word; we should probably rename this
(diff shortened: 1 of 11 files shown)


core/schema/git.go
```diff
@@ -167,6 +170,51 @@ func (s *gitSchema) tag(ctx context.Context, parent *core.GitRepository, args ta
 	}, nil
 }
 
+type tagsArgs struct {
+	Patterns dagql.Optional[dagql.ArrayInput[dagql.String]] `name:"patterns"`
+}
+
+func (s *gitSchema) tags(ctx context.Context, parent *core.GitRepository, args tagsArgs) ([]string, error) {
```
jedevc, line 177: Can we consolidate the implementation of this with https://github.com/dagger/dagger/blob/3dfbb7db81aa82db37a11d92cf4fa41239e173f0/core/schema/git.go#L254-L255

grouville: Not sure to follow, do you mean adding the `--symref` flag ? Otherwise, this seems to be a pretty equivalent implementation apart the fact that we rely on the `git.URL` instead of having it as an argument ; and the fact that we can add some glob patterns ?

jedevc: Yeah, this is the same implementation - so we should avoid having two copies of the same implementation, and instead have one function for all of this.

Otherwise, this can introduces bugs and other confusions if the two implementations diverge over time.

grouville: Sure yes, the purpose is indeed to replace it with this call 🙏 I am more stuck at the moment on how to mount the auth socket to have an authenticated call instead of relying on the underlying os.exec

sipsma:
@grouville RE: the SSH auth sock

There is indeed some overlap with the gitdns LLB source, but IMO we don't need to worry about that right now or try to consolidate. I think that would be a much larger issue than your effort here and expand the scope too much. We already run `git` as a subcommand outside of gitdns in a few places so your code here doesn't make the situation any worse.
* It's also not worth trying to update gitdns to have this functionality; it's too tied into buildkit's LLB sources and very hard to adapt to this use case of just querying tags.

You should however be able to do something similar-ish to what gitdns is doing with mounting the auth sock + ssh known hosts, but the implementation for this API can just be in the `schema` package. Outline:
1. the `parent *GitRepository` arg has a field `SSHAuthSocket *Socket`, you will want to turn that into a mounted unix socket available to the `git` subprocess
2. For now auth socks can only come from the "main client" (i.e. the actual user running a CLI command, as opposed to other nested clients running in modules), so you can get a connection with them via something like `parent.Query.Buildkit.SessionManager.Get(ctx, parent.Query.MainClientCallerID, true)`
   * Another example of doing something similar [here](https://github.com/sipsma/dagger/blob/0290df5df67b2514a0dd045652b0083017a86b14/core/c2h.go#L26-L26)
3. The return value of that is what you will plug in as the `caller` value to `sshforward.MountSSHSocket`
   * e.g. similar to [here](https://github.com/sipsma/dagger/blob/0290df5df67b2514a0dd045652b0083017a86b14/engine/sources/gitdns/source.go#L286-L286)
4. The return value of `MountSSHSocket` is a path to the mounted socket. You should be able to pass that to the `git` subprocess by setting the `SSH_AUTH_SOCK` env var on it, with the path to the mounted socket as the value.

That's pretty high-level still, but the gist of the idea. Let me know if it's making sense so far!

grouville:
Separated the handling of auth with git.Tags from this PR, to manage all the private access tests in the private module support PR, as it makes more sense

This PR now correctly removes the previous gitTags() implementation and relies on the git.Tags internal call inside the moduleSource function. This is useful for the private Git support as it enables us to implement the auth at the git.tags schema implementation level

jedevc:
LGTM, can we get a changelog entry though? (note the job for it is failing because of https://github.com/dagger/dagger/pull/7806, that can be ignored)

Issue #7662, opened 2024-06-14
flaky test: `TestPythonProjectLayout` - `The "main" module could not be found`


sipsma:
Saw this fail here: https://github.com/dagger/dagger/actions/runs/9509605842/job/26212800646?pr=7660

Don't have the cloud trace on hand because it got re-run and couldn't find it in the history. But ugly output extracted from GHA logs is:
```
2024-06-14T02:27:54.5780199Z [95m122 : [0m    [90m[5m41.8s] [0m[90m|[0m     module_python_test.go:306: 
2024-06-14T02:27:54.5781462Z [95m122 : [0m    [90m[5m41.8s] [0m[90m|[0m         	Error Trace:	/app/core/integration/module_python_test.go:306
2024-06-14T02:27:54.5782663Z [95m122 : [0m    [90m[5m41.8s] [0m[90m|[0m         	Error:      	Received unexpected error:
2024-06-14T02:27:54.5785058Z [95m122 : [0m    [90m[5m41.8s] [0m[90m|[0m         	            	input: container.from.withMountedFile.withWorkdir.withNewFile.withNewFile.withExec.withExec.stdout resolve: process "dagger --debug call hello" did not complete successfully: exit code: 1
2024-06-14T02:27:54.5787069Z [95m122 : [0m    [90m[5m41.8s] [0m[90m|[0m         	            	
2024-06-14T02:27:54.5787841Z [95m122 : [0m    [90m[5m41.8s] [0m[90m|[0m         	            	Stderr:
2024-06-14T02:27:54.5788940Z [95m122 : [0m    [90m[5m41.8s] [0m[90m|[0m         	            	[2m02:27:46[0m DBG frontend exporting [2mspans=[0m2
2024-06-14T02:27:54.5791159Z [95m122 : [0m    [90m[5m41.8s] [0m[90m|[0m         	            	[2m02:27:46[0m DBG frontend exporting span [2mtrace=[0ma9d4b5aa2511eee50f6d9555c85e5e64 [2mid=[0mc3b21a11941dfd56 [2mparent=[0m27c50ec63ed30b6a [2mspan=[0m"dagger --debug call hello"
[...]

PR #7744, opened 2024-06-24
Feat/context directory

### LINKED ISSUE #7647: Context directory
*This design is a subset of #7199 and #7432 . It intentionally keeps a narrow focus, to allow for rapid implementation, and easing user pain as soon as possible.*

## Problem

Dagger modules are often *contextual*: they are used in the context of another software projet. Almost always, these modules need to access their *context directory* - typically the project's source code.

- Today a module *cannot directly access its context directory*. Instead it must be passed by the caller as an argument. This is inconvenient.
- To make things worse, modules cannot *filter* the contents of the context directory. This is essential to avoid the prohibitive cost of uploading a huge monorepo on every task.

## Solution

Give modules the ability to access their context directory at runtime.

### Status

- Design consensus: ✅
- Implementation: #7744 

### Spec

To access the context directory, simply set a default value to your directory arguments. When applying the default value, the engine will interpret it as a path in the module's context directory, and pass the corresponding directory.

- The context directory is the git repository containing the module. If there is not git repository, the context is simply the module.
- Relative paths are relative to the current source file, scoped to the context directory for security
- Absolute paths are rooted in the context directory
- There is no limit to the number of directories which can have default values of this sort
- Optionally, directories can be filtered with a gitignore-style annotation

In this example, we use this technique for several files and directories, with a mix of relative and absolute paths.

```golang
// Build the app
func (app *App) Build(
  // Frontend source code
  // +default="./frontend/src"
  frontend *Directory,
  // Backend source code
  // +default="./backend/app/src"
  // +ignore=["*", "!**/*.go", "!go.mod", "!go.sum"]
  backend *Directory,
  // Build tool configuration, shared across the monorepo
  // +default="/tools/builder.conf"
  buildConfig *File,
) *Directory {
  // Execute the build as usual
}
```

### Caching considerations

We favored a design that used function arguments, rather than core function calls (`dag.Foo()`) because it makes it easier to protect caching. In this design, the engine always knows exactly which subset of the context directory, if any, a function needs. So it can invalidate cache if, and only if, that subset changes. This allows for massive performance optimizations in large monorepos especially.

### Relationship to other proposals

This design is a subset of #7199 and #7432 . It intentionally keeps a narrow focus, to allow for rapid implementation, and easing user pain as soon as possible.

TomChv:
This is an implementation of https://github.com/dagger/dagger/issues/7647

### API changes

This PR updates the `Function` and `FunctionArg` from our engine API to support `defaultPathFromContext`

```gql
# Function

  withArg(
    """
    If the argument is a Directory or File type, default to load path from context directory, relative to root directory.
    """
    defaultPath String = ""

   ignore: [String!]
# }

# FunctionArg {

  """
  If the argument is a Directory or File type, default to load path from context directory, relative to root directory.
  """
  defaultPath: String!

  ignore: [String!]
# }
```

### Engine changes

- Update `Call` to load contextual values and insert it as input.
- Add a `loadContextualArg` function that take the typedef and retrieve its contextual default value.
- Add a function `loadContext` to load from `git` or `local` source the contextual path.
  This implements the spec designed in #7647
  For git: we load it from the context directory since it includes the whole repo.
  For local: we load it from `importLocal` with the specific directory specified so we never load the full repo expect for `/`

**Current progress**

- [x] Loading directory
- [x] Get `*Directory` to `ID`: I found a workaround with `Blob` but it's not okay for `File`, I will check if I can use select instead.
- [x] Load File, using `Select` actually works!
- [x] Handle relative path
- [x] Write tests
- [x] Complete all SDK support

**Spec Reminder**
- [x] The context directory is the git repository containing the module. If there is not git repository, the context is simply the module.
- [x] Relative paths are relative to the current source file, scoped to the context directory for security
- [x] Absolute paths are rooted in the context directory
- [x] There is no limit to the number of directories which can have default values of this sort
- [x] Optionally, directories can be filtered with a gitignore-style annotation

### SDKs support

- [x] Go using pragmas directive on `Directory` or `File` type ⌛ 

- [x] Typescript using `@argument({ defaultPath: "path", ignore: ["xx", "!xx"] })` ✅
Example

```typescript
@func()
build(
  @argument({ defaultPath:"./frontend/src" }) frontend: Directory,
  @argument({ defaultPath:"./backend/app/src" }) backend: Directory,
  @argument({ defaultPath: "./tools/builder.conf" }) buildConfig: File,
): Directory {}
```

- [x] Python using annotations 

Fixes #7647

### DIFF
--- a/cmd/codegen/generator/go/templates/module_funcs.go
+++ b/cmd/codegen/generator/go/templates/module_funcs.go
@@ -301,10 +301,6 @@ func (ps *parseState) parseParamSpecVar(field *types.Var, docComment string, lin
 
 	ignore := []string{}
 	if v, ok := pragmas["ignore"]; ok {
-		if strings.HasPrefix(v, `"`) && strings.HasSuffix(v, `"`) {
-			v = v[1 : len(v)-1]
-		}
-
 		if err := json.Unmarshal([]byte(v), &ignore); err != nil {
 			return paramSpec{}, fmt.Errorf("ignore pragma '%s', must be a valid JSON array: %w", v, err)
 		}
--- a/cmd/codegen/generator/go/templates/module_funcs.go
+++ b/cmd/codegen/generator/go/templates/module_funcs.go
@@ -135,6 +135,19 @@ func (spec *funcTypeSpec) TypeDefCode() (*Statement, error) {
 			argOptsCode = append(argOptsCode, Id("DefaultValue").Op(":").Id("dagger").Dot("JSON").Call(Lit(argSpec.defaultValue)))
 		}
 
+		if argSpec.defaultPath != "" {
+			argOptsCode = append(argOptsCode, Id("DefaultPath").Op(":").Lit(argSpec.defaultPath))
+		}
+
+		if len(argSpec.ignore) > 0 {
+			ignores := make([]Code, 0, len(argSpec.ignore))
+			for _, pattern := range argSpec.ignore {
+				ignores = append(ignores, Lit(pattern))
+			}
+
+			argOptsCode = append(argOptsCode, Id("Ignore").Op(":").Index().String().Values(ignores...))
+		}
+
 		// arguments to WithArg (args to arg... ugh, at least the name of the variable is honest?)
 		argTypeDefArgCode := []Code{Lit(argSpec.name), argTypeDefCode}
 		if len(argOptsCode) > 0 {
@@ -276,6 +289,26 @@ func (ps *parseState) parseParamSpecVar(field *types.Var, docComment string, lin
 			optional, _ = strconv.ParseBool(v)
 		}
 	}
+	defaultPath := ""
+	if v, ok := pragmas["defaultPath"]; ok {
+		defaultPath = v
+		if strings.HasPrefix(v, `"`) && strings.HasSuffix(v, `"`) {
+			defaultPath = v[1 : len(v)-1]
+		}
+
+		optional = true // If defaultPath is set, the argument becomes optional
+	}
+
+	ignore := []string{}
+	if v, ok := pragmas["ignore"]; ok {
+		if strings.HasPrefix(v, `"`) && strings.HasSuffix(v, `"`) {
+			v = v[1 : len(v)-1]
+		}
+
+		if err := json.Unmarshal([]byte(v), &ignore); err != nil {
+			return paramSpec{}, fmt.Errorf("ignore pragma '%s', must be a valid JSON array: %w", v, err)
+		}
+	}
 
 	// ignore ctx arg for parsing type reference
 	isContext := paramType.String() == contextTypename
@@ -302,6 +335,8 @@ func (ps *parseState) parseParamSpecVar(field *types.Var, docComment string, lin
 		isContext:    isContext,
 		defaultValue: defaultValue,
 		description:  comment,
+		defaultPath:  defaultPath,
+		ignore:       ignore,
 	}, nil
 }
 
@@ -314,6 +349,7 @@ type paramSpec struct {
 	// isContext is true if the type is context.Context
 	isContext bool
 
+	// Set a default value for the argument. Value must be a json-encoded literal value
 	defaultValue string
 
 	// paramType is the full type declared in the function signature, which may
@@ -326,4 +362,13 @@ type paramSpec struct {
 	// parent is set if this paramSpec is nested inside a parent inline struct,
 	// and is used to create a declaration of the entire inline struct
 	parent *paramSpec
+
+	// Only applies to arguments of type File or Directory.
+	// If the argument is not set, load it from the given path in the context directory
+	defaultPath string
+
+	// Only applies to arguments of type Directory.
+	// The ignore patterns are applied to the input directory, and
+	// matching entries are filtered out, in a cache-efficient manner.
+	ignore []string
 }
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -7511,6 +7511,59 @@ class Test {
 			})
 		}
 	})
+
+	t.Run("deps", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		ctr := goGitBase(t, c).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work/dep").
+			With(daggerExec("init", "--source=.", "--name=dep", "--sdk=go")).
+			WithNewFile("main.go", `package main
+
+import (
+	"dagger/dep/internal/dagger"
+)
+
+type Dep struct{}
+
+func (m *Dep) GetSource(
+	// +defaultPath="/dep"
+	// +ignore=["!yo"]
+	source *dagger.Directory,
+) *dagger.Directory {
+	return source
+}
+`,
+			).
+			WithNewFile("yo", "yo")
+
+		out, err := ctr.With(daggerCall("get-source", "entries")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "yo\n", out)
+
+		ctr = ctr.
+			WithWorkdir("/work").
+			With(daggerExec("init", "--source=.", "--name=test", "--sdk=go")).
+			With(daggerExec("install", "./dep")).
+			WithNewFile("main.go", `package main
+
+import (
+	"dagger/test/internal/dagger"
+)
+
+type Test struct{}
+
+func (m *Test) GetDepSource() *dagger.Directory {
+	return dag.Dep().GetSource()
+}
+`,
+			)
+
+		out, err = ctr.With(daggerCall("get-dep-source", "entries")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "yo\n", out)
+	})
 }
 
 func (ModuleSuite) TestModuleDevelopVersion(ctx context.Context, t *testctx.T) {
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -7053,7 +7053,7 @@ class Test {
 				t.Run("absolute and relative root context dir", func(ctx context.Context, t *testctx.T) {
 					out, err := modGen.With(daggerCallAt("ci", "dirs")).Stdout(ctx)
 					require.NoError(t, err)
-					require.Equal(t, "backend\nci\nfrontend\nLICENSE\ndagger\ndagger.json\n", out)
+					require.Equal(t, ".git\nbackend\nci\nfrontend\nLICENSE\ndagger\ndagger.json\n", out)
 				})
 
 				t.Run("dir ignore", func(ctx context.Context, t *testctx.T) {
@@ -7317,7 +7317,7 @@ class Test {
 				t.Run("absolute and relative root context dir", func(ctx context.Context, t *testctx.T) {
 					out, err := modGen.With(daggerCall("dirs")).Stdout(ctx)
 					require.NoError(t, err)
-					require.Equal(t, "LICENSE\nbackend\ndagger\ndagger.json\nfrontend\nLICENSE\nbackend\ndagger\ndagger.json\nfrontend\n", out)
+					require.Equal(t, ".git\nLICENSE\nbackend\ndagger\ndagger.json\nfrontend\n.git\nLICENSE\nbackend\ndagger\ndagger.json\nfrontend\n", out)
 				})
 
 				t.Run("absolute context dir subpath", func(ctx context.Context, t *testctx.T) {
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -6905,6 +6905,74 @@ func (t *Test) Files(
 `,
 			},
 			{
+				sdk: "python",
+				source: `from typing import Annotated
+
+import dagger
+from dagger import DefaultPath, Ignore, function, object_type
+
+
+@object_type
+class Test:
+    @function
+    async def dirs(
+        self,
+        root: Annotated[dagger.Directory, DefaultPath("/")],
+        relativeRoot: Annotated[dagger.Directory, DefaultPath(".")],
+    ) -> list[str]:
+        return [
+            *(await root.entries()),
+            *(await relativeRoot.entries()),
+       ]
+
+    @function
+    async def dirs_ignore(
+        self,
+        root: Annotated[dagger.Directory, DefaultPath("/"), Ignore(["!backend", "!frontend"])],
+        relativeRoot: Annotated[dagger.Directory, DefaultPath("."), Ignore(["dagger.json", "LICENSE"])],
+    ) -> list[str]:
+        return [
+            *(await root.entries()),
+            *(await relativeRoot.entries()),
+        ]
+
+    @function
+    async def root_dir_path(
+        self,
+        backend: Annotated[dagger.Directory, DefaultPath("/backend")],
+        frontend: Annotated[dagger.Directory, DefaultPath("/frontend")],
+        mod_src_dir: Annotated[dagger.Directory, DefaultPath("/ci/dagger/sub")],
+    ) -> list[str]:
+        return [
+            *(await backend.entries()),
+            *(await frontend.entries()),
+            *(await mod_src_dir.entries()),
+        ]
+
+    @function
+    async def relative_dir_path(
+        self,
+        mod_src_dir: Annotated[dagger.Directory, DefaultPath("./dagger/sub")],
+        backend: Annotated[dagger.Directory, DefaultPath("../backend")],
+    ) -> list[str]:
+        return [
+            *(await mod_src_dir.entries()),
+            *(await backend.entries()),
+        ]
+
+    @function
+    async def files(
+        self,
+        license: Annotated[dagger.File, DefaultPath("/ci/LICENSE")],
+        index: Annotated[dagger.File, DefaultPath("./dagger/sub/sub.txt")],
+    ) -> list[str]:
+        return [
+            await license.name(),
+            await index.name(),
+        ]
+`,
+			},
+			{
 				sdk: "typescript",
 				source: `import { Directory, File, object, func, argument } from "@dagger.io/dagger"
 
@@ -6931,8 +6999,8 @@ class Test {
 
   @func()
   async rootDirPath(
-    @argument({ defaultPath: "/backend" }) backend: Directory, 
-    @argument({ defaultPath: "/frontend" }) frontend: Directory, 
+    @argument({ defaultPath: "/backend" }) backend: Directory,
+    @argument({ defaultPath: "/frontend" }) frontend: Directory,
     @argument({ defaultPath: "/ci/dagger/sub" }) modSrcDir: Directory,
   ): Promise<string[]> {
     const backendFiles = await backend.entries()
@@ -6944,7 +7012,7 @@ class Test {
 
   @func()
   async relativeDirPath(
-    @argument({ defaultPath: "./dagger/sub" }) modSrcDir: Directory, 
+    @argument({ defaultPath: "./dagger/sub" }) modSrcDir: Directory,
     @argument({ defaultPath: "../backend" }) backend: Directory,
   ): Promise<string[]> {
     const modSrcDirFiles = await modSrcDir.entries()
@@ -7120,6 +7188,62 @@ func (t *Test) Files(
 `,
 			},
 			{
+				sdk: "python",
+				source: `from typing import Annotated
+
+import dagger
+from dagger import DefaultPath, function, object_type
+
+@object_type
+class Test:
+    @function
+    async def dirs(
+        self,
+        root: Annotated[dagger.Directory, DefaultPath("/")],
+        relative_root: Annotated[dagger.Directory, DefaultPath(".")],
+    ) -> list[str]:
+        return [
+            *(await root.entries()),
+            *(await relative_root.entries()),
+        ]
+
+    @function
+    async def root_dir_path(
+        self,
+        backend: Annotated[dagger.Directory, DefaultPath("/backend")],
+        frontend: Annotated[dagger.Directory, DefaultPath("/frontend")],
+        mod_src_dir: Annotated[dagger.Directory, DefaultPath("/dagger/sub")],
+    ) -> list[str]:
+        return [
+            *(await backend.entries()),
+            *(await frontend.entries()),
+            *(await mod_src_dir.entries()),
+        ]
+
+    @function
+    async def relative_dir_path(
+        self,
+        mod_src_dir: Annotated[dagger.Directory, DefaultPath("./dagger/sub")],
+        backend: Annotated[dagger.Directory, DefaultPath("./backend")],
+    ) -> list[str]:
+        return [
+            *(await mod_src_dir.entries()),
+            *(await backend.entries()),
+        ]
+
+    @function
+    async def files(
+        self,
+        license: Annotated[dagger.File, DefaultPath("/LICENSE")],
+        index: Annotated[dagger.File, DefaultPath("./dagger.json")],
+    ) -> list[str]:
+        return [
+            await license.name(),
+            await index.name(),
+        ]
+`,
+			},
+			{
 				sdk: "typescript",
 				source: `import { Directory, File, object, func, argument } from "@dagger.io/dagger"
 
@@ -7127,7 +7251,7 @@ func (t *Test) Files(
 class Test {
   @func()
   async dirs(
-    @argument({ defaultPath: "/" }) root: Directory, 
+    @argument({ defaultPath: "/" }) root: Directory,
     @argument({ defaultPath: "." }) relativeRoot: Directory,
   ): Promise<string[]> {
     const res = await root.entries()
@@ -7139,7 +7263,7 @@ class Test {
   @func()
   async rootDirPath(
     @argument({ defaultPath: "/backend" }) backend: Directory,
-    @argument({ defaultPath: "/frontend" }) frontend: Directory, 
+    @argument({ defaultPath: "/frontend" }) frontend: Directory,
     @argument({ defaultPath: "/dagger/sub" }) modSrcDir: Directory,
   ): Promise<string[]> {
     const backendFiles = await backend.entries()
@@ -7266,6 +7390,48 @@ func (t *Test) NonExistingFile(
 `,
 			},
 			{
+				sdk: "python",
+				source: `from typing import Annotated
+
+import dagger
+from dagger import DefaultPath, function, object_type
+
+@object_type
+class Test:
+    @function
+    async def too_high_relative_dir_path(
+        self,
+        backend: Annotated[dagger.Directory, DefaultPath("../../")],
+    ) -> list[str]:
+        # The engine should throw an error
+        return []
+
+    @function
+    async def non_existing_path(
+        self,
+        dir: Annotated[dagger.Directory, DefaultPath("/invalid")],
+    ) -> list[str]:
+        # The engine should throw an error
+        return []
+
+    @function
+    async def too_high_relative_file_path(
+        self,
+        backend: Annotated[dagger.File, DefaultPath("../../file.txt")],
+    ) -> str:
+        # The engine should throw an error
+        return ""
+
+    @function
+    async def non_existing_file(
+        self,
+        file: Annotated[dagger.File, DefaultPath("/invalid")],
+    ) -> str:
+        # The engine should throw an error
+        return ""
+`,
+			},
+			{
 				sdk: "typescript",
 				source: `import { Directory, File,object, func, argument } from "@dagger.io/dagger"
 @object()
@@ -7292,7 +7458,7 @@ class Test {
     // The engine should throw an error
     return ""
   }
-}			
+}
 `,
 			},
 		} {
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -6770,6 +6770,575 @@ func schemaVersion(ctx context.Context) (string, error) {
 	})
 }
 
+func (ModuleSuite) TestContextDirectory(ctx context.Context, t *testctx.T) {
+	type testCase struct {
+		sdk    string
+		source string
+	}
+
+	t.Run("load context inside git repo with module in a sub dir", func(ctx context.Context, t *testctx.T) {
+		for _, tc := range []testCase{
+			{
+				sdk: "go",
+				source: `package main
+
+import (
+  "context"
+)
+
+type Test struct {}
+
+func (t *Test) Dirs(
+  ctx context.Context,
+
+  // +defaultPath="/"
+  root *Directory,
+
+  // +defaultPath="."
+  relativeRoot *Directory,
+) ([]string, error) {
+  res, err := root.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+  relativeRes, err := relativeRoot.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+  return append(res, relativeRes...), nil
+}
+
+
+func (t *Test) DirsIgnore(
+  ctx context.Context,
+
+  // +defaultPath="/"
+  // +ignore="["!backend", "!frontend"]"
+  root *Directory,
+
+  // +defaultPath="."
+  // +ignore="["dagger.json", "LICENSE"]"
+  relativeRoot *Directory,
+) ([]string, error) {
+  res, err := root.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+  relativeRes, err := relativeRoot.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+  return append(res, relativeRes...), nil
+}
+
+func (t *Test) RootDirPath(
+  ctx context.Context,
+
+  // +defaultPath="/backend"
+  backend *Directory,
+
+  // +defaultPath="/frontend"
+  frontend *Directory,
+
+  // +defaultPath="/ci/dagger/sub"
+  modSrcDir *Directory,
+) ([]string, error) {
+  backendFiles, err := backend.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+  frontendFiles, err := frontend.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+  modSrcDirFiles, err := modSrcDir.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+
+	res := append(backendFiles, append(frontendFiles, modSrcDirFiles...)...)
+
+  return res, nil
+}
+
+func (t *Test) RelativeDirPath(
+  ctx context.Context,
+
+  // +defaultPath="./dagger/sub"
+  modSrcDir *Directory,
+
+  // +defaultPath="../backend"
+  backend *Directory,
+) ([]string, error) {
+  modSrcDirFiles, err := modSrcDir.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+  backendFiles, err := backend.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+
+  return append(modSrcDirFiles, backendFiles...), nil
+}
+
+func (t *Test) Files(
+  ctx context.Context,
+
+  // +defaultPath="/ci/LICENSE"
+  license *File,
+
+  // +defaultPath="./dagger/sub/sub.txt"
+  index *File,
+) ([]string, error) {
+  licenseName, err := license.Name(ctx)
+  if err != nil {
+    return nil, err
+  }
+  indexName, err := index.Name(ctx)
+  if err != nil {
+    return nil, err
+  }
+
+  return []string{licenseName, indexName}, nil
+}
+`,
+			},
+			{
+				sdk: "typescript",
+				source: `import { Directory, File, object, func, argument } from "@dagger.io/dagger"
+
+@object()
+class Test {
+  @func()
+  async dirs(@argument({ defaultPath: "/" }) root: Directory, @argument({ defaultPath: "."}) relativeRoot: Directory): Promise<string[]> {
+    const res = await root.entries()
+    const relativeRes = await relativeRoot.entries()
+
+    return [...res, ...relativeRes]
+  }
+
+  @func()
+  async dirsIgnore(
+    @argument({ defaultPath: "/", ignore: ["!backend", "!frontend"] }) root: Directory,
+    @argument({ defaultPath: ".", ignore: ["dagger.json", "LICENSE"] }) relativeRoot: Directory,
+  ): Promise<string[]> {
+    const res = await root.entries();
+    const relativeRes = await relativeRoot.entries();
+
+    return [...res, ...relativeRes];
+  }
+
+  @func()
+  async rootDirPath(
+    @argument({ defaultPath: "/backend" }) backend: Directory, 
+    @argument({ defaultPath: "/frontend" }) frontend: Directory, 
+    @argument({ defaultPath: "/ci/dagger/sub" }) modSrcDir: Directory,
+  ): Promise<string[]> {
+    const backendFiles = await backend.entries()
+    const frontendFiles = await frontend.entries()
+    const modSrcDirFiles = await modSrcDir.entries()
+
+    return [...backendFiles, ...frontendFiles, ...modSrcDirFiles]
+  }
+
+  @func()
+  async relativeDirPath(
+    @argument({ defaultPath: "./dagger/sub" }) modSrcDir: Directory, 
+    @argument({ defaultPath: "../backend" }) backend: Directory,
+  ): Promise<string[]> {
+    const modSrcDirFiles = await modSrcDir.entries()
+    const backendFiles = await backend.entries()
+
+    return [...modSrcDirFiles, ...backendFiles]
+  }
+
+  @func()
+  async files(
+    @argument({ defaultPath: "/ci/LICENSE" }) license: File,
+    @argument({ defaultPath: "./dagger/sub/sub.txt" }) index: File,
+  ): Promise<string[]> {
+    return [await license.name(), await index.name()]
+  }
+}
+`,
+			},
+		} {
+			tc := tc
+
+			t.Run(tc.sdk, func(ctx context.Context, t *testctx.T) {
+				c := connect(ctx, t)
+
+				modGen := goGitBase(t, c).
+					WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+					WithWorkdir("/work").
+					WithDirectory("/work/backend", c.Directory().WithNewFile("foo.txt", "foo")).
+					WithDirectory("/work/frontend", c.Directory().WithNewFile("bar.txt", "bar")).
+					WithWorkdir("/work/ci").
+					With(daggerExec("init", "--name=test", "--sdk="+tc.sdk)).
+					With(sdkSource(tc.sdk, tc.source)).
+					WithDirectory("/work/ci/dagger/sub", c.Directory().WithNewFile("sub.txt", "sub")).
+					WithWorkdir("/work")
+
+				t.Run("absolute and relative root context dir", func(ctx context.Context, t *testctx.T) {
+					out, err := modGen.With(daggerCallAt("ci", "dirs")).Stdout(ctx)
+					require.NoError(t, err)
+					require.Equal(t, "backend\nci\nfrontend\nLICENSE\ndagger\ndagger.json\n", out)
+				})
+
+				t.Run("dir ignore", func(ctx context.Context, t *testctx.T) {
+					out, err := modGen.With(daggerCallAt("ci", "dirs-ignore")).Stdout(ctx)
+					require.NoError(t, err)
+					require.Equal(t, "backend\nfrontend\ndagger\n", out)
+				})
+
+				t.Run("absolute context dir subpath", func(ctx context.Context, t *testctx.T) {
+					out, err := modGen.With(daggerCallAt("ci", "root-dir-path")).Stdout(ctx)
+					require.NoError(t, err)
+					require.Equal(t, "foo.txt\nbar.txt\nsub.txt\n", out)
+				})
+
+				t.Run("relative context dir subpath", func(ctx context.Context, t *testctx.T) {
+					out, err := modGen.With(daggerCallAt("ci", "relative-dir-path")).Stdout(ctx)
+					require.NoError(t, err)
+					require.Equal(t, "sub.txt\nfoo.txt\n", out)
+				})
+
+				t.Run("files", func(ctx context.Context, t *testctx.T) {
+					out, err := modGen.With(daggerCallAt("ci", "files")).Stdout(ctx)
+					require.NoError(t, err)
+					require.Equal(t, "LICENSE\nsub.txt\n", out)
+				})
+			})
+		}
+	})
+
+	t.Run("load context inside git repo with module at the root of the repo", func(ctx context.Context, t *testctx.T) {
+		for _, tc := range []testCase{
+			{
+				sdk: "go",
+				source: `package main
+
+import (
+  "context"
+)
+
+type Test struct {}
+
+func (t *Test) Dirs(
+  ctx context.Context,
+
+  // +defaultPath="/"
+  root *Directory,
+
+  // +defaultPath="."
+  relativeRoot *Directory,
+) ([]string, error) {
+  res, err := root.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+  relativeRes, err := relativeRoot.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+  return append(res, relativeRes...), nil
+}
+
+
+func (t *Test) RootDirPath(
+  ctx context.Context,
+
+  // +defaultPath="/backend"
+  backend *Directory,
+
+  // +defaultPath="/frontend"
+  frontend *Directory,
+
+  // +defaultPath="/dagger/sub"
+  modSrcDir *Directory,
+) ([]string, error) {
+  backendFiles, err := backend.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+  frontendFiles, err := frontend.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+  modSrcDirFiles, err := modSrcDir.Entries(ctx)
+  if err != nil {
+    return nil, err
+  }
+
+	res := append(backendFiles, append(frontendFiles, modSrcDirFiles...)...)
+
+  return res, nil
+}
+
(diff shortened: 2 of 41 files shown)


sipsma:
(continuing a conversation from discord)

For now, I think the way to general way to approach this feature should be to leave the current module context loading as is since it's used by the SDKs to build modules and then add this support for user arg context dirs as separate context loading calls. There's likely ways of unifying and optimizing this but better to start simple for now.

There's different ways of approaching this, I'm just gonna outline the first reasonable possibility that comes to mind. Not set in stone at all though so happy to hear other opinions and happy to help out more if reality of implementing it doesn't align with theory.

1. (discussed before) `ModuleFunction` has access to the `Module` and `ModuleSource` via this field: https://github.com/sipsma/dagger/blob/5231aa7de19266430581da054604ce533cc85898/core/modfunc.go#L25-L25
2. Add a new method to `ModuleSource` called something like `LoadContext(path: string) (dagql.Instance[*Directory], error)` that allows you to dynamically load dirs from the module source context.
   * This would be different than the existing [`ContextDirectory` method](https://github.com/sipsma/dagger/blob/5231aa7de19266430581da054604ce533cc85898/core/modulesource.go#L247-L247) that returns the static one loaded for the SDKs (we can consider renaming it maybe for clarity).
   * The way that you implement loading the context depends on whether it's coming from a remote git repo or a local git repo.
      * git is easy because we haven't implemented fancy dynamic loading, we just have to pull the whole repo, so you can just copy the directory from it's already loaded context dir
      * local is more involved since you will need to actually load it on demand. You can look at [this code](https://github.com/sipsma/dagger/blob/5231aa7de19266430581da054604ce533cc85898/core/schema/modulesource.go#L641-L641) and the methods it calls to get an idea of how to do that (one of the really key parts being [this section](https://github.com/sipsma/dagger/blob/5231aa7de19266430581da054604ce533cc85898/core/schema/modulesource.go#L781-L794) that does the load itself) 
         * Another thing to be aware of is that the `LocalImport` call is sensitive to values passed in `ctx` which determine which client the dir should be loaded from. For now, we can just load from the "main client caller". I already wrote too much here, so just ping if you need pointers on this 🙂 
3. Once that's in place, you can plug the arg in as a directory value same as others are (i.e. in [this section](https://github.com/sipsma/dagger/blob/5231aa7de19266430581da054604ce533cc85898/core/modfunc.go#L138-L138))

TomChv:
> This would be different than the existing [ContextDirectory method](https://github.com/sipsma/dagger/blob/5231aa7de19266430581da054604ce533cc85898/core/modulesource.go#L247-L247) that returns the static one loaded for the SDKs (we can consider renaming it maybe for clarity).

I understand this but this directory will not have the full repo if it's not part included in `include`, how do I workaround that?

sipsma:
> > This would be different than the existing [ContextDirectory method](https://github.com/sipsma/dagger/blob/5231aa7de19266430581da054604ce533cc85898/core/modulesource.go#L247-L247) that returns the static one loaded for the SDKs (we can consider renaming it maybe for clarity).
> 
> I understand this but this directory will not have the full repo if it's not part included in `include`, how do I workaround that?

You would need to add a new method that does the actual load of files from the original context directory (i.e. loads from the user's local dirs or loads from the git repo). That's what I was referring to under "The way that you implement loading the context depends on whether it's coming from a remote git repo or a local git repo." in my previous comment

TomChv:
> Relative paths are relative to the current source file, scoped to the context directory for security

To do such thing, we need to know the file where the module func comes from, I'm not sure we got this information in the current API.
I think the resolution should be done by the SDK, what do you think @helderco @jedevc?

shykes:
> > Relative paths are relative to the current source file, scoped to the context directory for security
> 
> To do such thing, we need to know the file where the module func comes from, I'm not sure we got this information in the current API. I think the resolution should be done by the SDK, what do you think @helderco @jedevc?

Sorry I forgot to update that line after further discussion. It should be relative to the module root, not the file (that would be too difficult to find out).

TomChv:
> Sorry I forgot to update that line after further discussion. It should be relative to the module root, not the file (that would be too difficult to find out).

Ohh then yeah it's much easier haha

TomChv:
Another question @helderco @jedevc @shykes, should the API returns an error during registration if the typedef has both a default value and a contextual one?
This is possible in TypeScript and Python due to the notation so should this results in an error? I think so but I want your opinion.

Example of notation:

```typescript
@func()
 async test(@context("/") dir: Directory = dag.directory()): Promise<string[]>
```

shykes:
> should the API returns an error during registration if the typedef has both a default value and a contextual one? This is possible in TypeScript and Python due to the notation so should this results in an error? I think so but I want your opinion.

I thought we were using default values? I might have missed (or forgotten) the last round of bikeshedding on that topic.

TomChv:
> I thought we were using default values? I might have missed (or forgotten) the last round of bikeshedding on that topic.

No it's a specific field in the API, see the change of spec in the description

shykes:
@TomChv could you add example snippets of what that feature would look like, in the 3 main languages please? It would make it easier to discuss the DX tradeoffs

helderco:
> I thought we were using default values? I might have missed (or forgotten) the last round of bikeshedding on that topic.

I updated my view in https://github.com/dagger/dagger/issues/7647#issuecomment-2174570033:

> [...] warming up to not using the default value because of that as it makes an overall simpler implementation, just sacrificing on DX for SDKs that support default values, but that can be improved later.

---

> * [x]  Typescript using `@context("path")` ⚡ to be tested once the engine changes are completed
> * [ ]  Python using `@context("path")` ? To be confirmed by @helderco  ⌛

We need to bikeshed on the `@context` name.

In Python, this would have to be done using metadata annotation on the type:

```python
def build(
    self,
    backend: Annotated[
        dagger.Directory,
        DefaultFromContext("../backend/app/src"),
        Ignore(...),
    ]
) -> dagger.Directory:
    ...
```

TomChv:
> @TomChv could you add example snippets of what that feature would look like, in the 3 main languages please? It would make it easier to discuss the DX tradeoffs

I think it should look like this:

**Go**

```go
func (app *App) Build(
  // +context="./frontend/src"
  frontend *Directory,
  // +context="./backend/app/src"
  backend *Directory,
  // +context="/tools/builder.conf"
  buildConfig *File,
) *Directory {}
```

**Typescript**

```typescript
@func()
build(
  @context("./frontend/src") frontend: Directory,
  @context("./backend/app/src") backend: Directory,
  @context("./tools/builder.conf") buildConfig: File,
): Directory {}
```

**Python**

```python
@function
def build(
  self,
  @context("./frontend/src") frontend: Directory,
  @context("./backend/app/src") backend: Directory,
  @context("./tools/builder.conf") build_config: File,
) -> Directory:
```

helderco:
In Go, rather than `// +context`, it should be `// +default`.

shykes:
@TomChv could you add an example of how regular default values work in each case?

TomChv:
> @TomChv could you add an example of how regular default values work in each case?

Example on https://docs.dagger.io/manuals/developer/functions#default-values

**Go**

 I'm not sure you can default to a directory value in Go so I don't know??

But for primitive type, you set a pragma like `+default=4`


**Typescript**

```typescript
@func()
build(
  frontend: Directory = dag.directory(),
  backend: Directory = dag.directory(),
  buildConfig: File = dag.directory().WithNewFile("xxx", "xxx").File("xxx"),
): Directory {}
```

**Python**

I'm not sure about the syntax but probably:

```python
def build(
  self,
  backend: Annotated[
     dagger.Directory,
  ] = dag.directory()
) -> dagger.Directory:
  ...
```

TomChv:
Few changes that we'll introduced on this PR based on my conversation with:
- change `@context` to `@argument` with `default` and `ignore` metadata
- support `ignore` as parameter 
- Move test to `module_test.go` because it's common to every SDK, not testing the API
- Raise error if context & default is set, check inside TS SDK if context & default is set then throw error

TomChv:
@helderco 

I've updated the PR with the new API and implement ignore.
All point above has been implemented 🚀

helderco (commented):
If I'm in a module in `/foo` (/ == git root) and the default path is `/bar/man`, the path is not resolved correctly:
```
$ dagger call debug
lstat /Users/.../my-repo/foo/foo: no such file or directory
```

If I'm in `/bat`:
```
$ dagger call -m ../foo debug
lstat /Users/.../my-repo/bat/foo: no such file or directory
```

TomChv:
> If I'm in a module in /foo (/ == git root) and the default path is /bar/man, the path is not resolved correctly:

Could I also see the code snippet and the tree view of the directory please? So I can make a repro locally

TomChv:
Test are failing due to 2 Python errors: https://dagger.cloud/dagger/traces/04e8dd584b027109a12f120f41a0e209?span=f60a702af34c8c5f

Nothing directly related to the PR, I can also see this issue on #7824

helderco:
> Test are failing due to 2 Python errors: https://dagger.cloud/dagger/traces/04e8dd584b027109a12f120f41a0e209?span=f60a702af34c8c5f
> 
> Nothing directly related to the PR, I can also see this issue on #7824

That's a known flake:
- https://github.com/dagger/dagger/issues/7662

docs/docs-graphql/schema.graphqls
```diff
@@ -1435,6 +1440,9 @@ type Function {
     """A doc string for the argument, if any"""
     description: String = ""
 
+    """Patterns to ignore when loading the contextual argument value."""
+    ignore: [String!] = []
```
jedevc, line 1444: Can we call this something like `defaultPathIgnore`? Something to make it clear that this applies *only* to the `defaultPath`?

Alternatively, we could add a new method `withPathArg`, for clarity, but that might be less popular.

core/modulesource.go
```diff
@@ -245,6 +249,150 @@ func (src *ModuleSource) AutomaticGitignore(ctx context.Context) (*bool, error)
...
+// If the module is git, it will load the directory from the git repository
+// using its context directory.
+//
+// The path is relative to the module context directory, even if it's absolute.
+//
+// Improvement for later: use `.gitignore` to excludes paths from the context.
+func (src *ModuleSource) LoadContext(ctx context.Context, dag *dagql.Server, path string, ignore []string) (inst dagql.Instance[*Directory], err error) {
+	// We exclude .git by default because it's not useful and tends to invalidate the cache.
+	excludes := []string{"**/.git"}
+	includes := []string{}
```
jedevc, line 266: I don't think we should apply this by default - there are legitimate cases for needing the `.git` directory, like getting commit ids, etc. There's no way to disable this really, and it's a bit non-obvious.

Ideally, we'd be able to use context directories *in combination* with some `include`/`exclude`s/views in `dagger.json`, so `.git` could be disabled globally there - but I don't think we should apply it as a default for everywhere.

core/modulesource.go
```diff
@@ -245,6 +249,150 @@ func (src *ModuleSource) AutomaticGitignore(ctx context.Context) (*bool, error)
...
+	}
+
+	switch src.Kind {
+	case ModuleSourceKindLocal:
+		bk, err := src.Query.Buildkit(ctx)
+		if err != nil {
+			return inst, fmt.Errorf("failed to get buildkit api: %w", err)
+		}
+
+		slog.Debug("moduleSource.LoadContext: loading contextual directory from local source", "path", path, "kind", src.Kind)
```
jedevc, line 293: Just an FYI, you can print the type using the `%T` format specifier, instead of manually duplicating the information :tada:

core/modulesource.go
```diff
@@ -245,6 +249,150 @@ func (src *ModuleSource) AutomaticGitignore(ctx context.Context) (*bool, error)
...
+
+		dir, err := ctxDir.Directory(ctx, path)
+		if err != nil {
+			return inst, fmt.Errorf("failed to load contextual directory %q: %w", path, err)
+		}
+
+		loadedDir, err := dir.AsBlob(ctx, dag)
+		if err != nil {
+			return inst, fmt.Errorf("failed to get dir instance: %w", err)
+		}
```
jedevc, line 333: This uses one more blob conversion than neccessary - you still need the `Instance`, but you can get that by using `dag.Select`.

core/modulesource.go
```diff
@@ -245,6 +249,150 @@ func (src *ModuleSource) AutomaticGitignore(ctx context.Context) (*bool, error)
...
+
+		return loadedDir, nil
+	default:
+		return inst, fmt.Errorf("unsupported module src kind: %q", src.Kind)
+	}
+}
+
+// resolveContextPaths returns the context path to the .git directory
+// if it exists. Otherwise, it returns the source root directory.
+func (src *ModuleSource) resolveContextPath(ctx context.Context, bk *buildkit.Client) (string, error) {
```
jedevc, line 343: This (and `callerHostFindUpContext`) look like direct copies from `core/schema/modulesource.go`. That makes sense, but we should now remove those implementations, and have the schema ones call these ones.

Duplicates are generally not great when avoidable, since in practice they're fiddly to maintain, and they have a higher potential to drift, which can lead to all sorts of exciting bugs.

core/module.go
```diff
@@ -63,6 +63,9 @@ type Module struct {
 
 	// InstanceID is the ID of the initialized module.
 	InstanceID *call.ID
+
+	// Server is the dagql server
+	Server *dagql.Server
```
jedevc, line 68: This shouldn't go here.

I see they're attached so that `ModuleFunction` can use this - we should probably instead just add it as an arg everywhere it needs to go. It feels like a code smell to attach it in this way to me. I feel like it's tempting some very subtle bugs in the context of https://github.com/dagger/dagger/pull/7759.

cmd/codegen/generator/go/templates/module_funcs.go
```diff
@@ -276,6 +289,26 @@ func (ps *parseState) parseParamSpecVar(field *types.Var, docComment string, lin
...
 		}
 	}
+	defaultPath := ""
+	if v, ok := pragmas["defaultPath"]; ok {
+		defaultPath = v
+		if strings.HasPrefix(v, `"`) && strings.HasSuffix(v, `"`) {
+			defaultPath = v[1 : len(v)-1]
+		}
+
+		optional = true // If defaultPath is set, the argument becomes optional
```
jedevc, line 299: Hm, this doesn't match the behavior of `default`. Definitely, the current behavior of needing `optional` and `default` is absolutely weird, but if we fix this, I think we should fix this consistently.

cmd/codegen/generator/go/templates/module_funcs.go
```diff
@@ -276,6 +289,26 @@ func (ps *parseState) parseParamSpecVar(field *types.Var, docComment string, lin
 			optional, _ = strconv.ParseBool(v)
 		}
 	}
+	defaultPath := ""
+	if v, ok := pragmas["defaultPath"]; ok {
```
jedevc, line 293: I'm not a huge fan of using `defaultPath` as the pragma here (even though it's distinct at the API level). Could we just use `default` here?

It doesn't make sense to use `defaultPath` on a non File/Directory arg, and it doesn't make sense to use `default` on a File/Directory arg: we should condense this down into one `default`.

docs/docs-graphql/schema.graphqls
helderco: The thing is that it's not only for the defaulth path though. It can be used as a replacement for _views_ for example, applying to a local directory referenced from the CLI.

core/modulesource.go
helderco: I believe that was copied from here:

https://github.com/dagger/dagger/blob/540d884cf4583313356ebcb21b97da2b926122d1/core/schema/modulesource.go#L758-L761

I think it's fine, if the function author wants to allow `.git` they can add `!**/.git` to `// +ignore`.

core/modulesource.go
helderco: Agreed.

cmd/codegen/generator/go/templates/module_funcs.go
helderco: We do plan on fixing this, see:
- https://github.com/dagger/dagger/issues/6749#issuecomment-2078334724

Simple enough to add to this release.

cmd/codegen/generator/go/templates/module_funcs.go
helderco: He's following the design in:
- https://github.com/dagger/dagger/issues/7647#issuecomment-2201921231

Best to bikeshed this in the issue.

helderco: FTR, I agree. It's Python and Go that need a distinction since they do support default values in parameters, but in Go it's simpler if it's the same (i.e., `default`) as it doesn't have the same constraints here. The only downside is it's less clear when comparing or porting code from one language to another (consistency).

core/integration/module_test.go
```diff
@@ -6027,6 +6027,741 @@ func (m *Test) Fn() string {
 	require.NoError(t, err)
 }
 
+func (ModuleSuite) TestContextDirectory(ctx context.Context, t *testctx.T) {
```
helderco, line 6030: nit: this may be testing too much for the SDKs. When we test multiple SDKs like this, you want to make sure you're only testing if the SDK is making the right API calls.

Then you make another test case to more fully test the engine's behavior, using `dagger call` for example, but only need the Go SDK for that. That's where you make sure all the relative and absolute paths are working as expected.

Also, forcing the request within the function and concatenating the string slices make it more complicated than it needs to be. You could just return the `Directory` directly and get the `entries` with `dagger call`.

sdk/python/src/dagger/mod/_arguments.py
```diff
@@ -1,41 +1,100 @@
...
 
-    has_default: bool = dataclasses.field(init=False)
-    is_optional: bool = dataclasses.field(init=False)
+    def __str__(self) -> str:
+        return self.name
+
+
+@deprecated("Arg is deprecated, use Alias instead.")
+def Arg(name: str) -> Alias:  # noqa: N802
+    return Alias(name)
```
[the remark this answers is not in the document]
helderco: There's also a planned change that will make `Alias` less necessary.

core/modulesource.go
helderco: Also FTR, we're tying to see if this feature can replace _views_, so consider it might get deprecated.

cmd/codegen/generator/go/templates/module_funcs.go
helderco: Reminder that `optional = true` here, actually means _nullable_, but in this case it's a lazy default, i.e., it doesn't get saved in the schema as the actual default. It makes sense for that to be `null`. It's at runtime that the engine sees that the value is _null_, so it resolves `defaultPath` and feeds that as input.

With a normal `// +default`, it's different. If you always assume that always means it's `optional = true`, you're actually saying any arg with a default value is also nullable.

core/module.go
TomChv: I agree but the function where `dag` is used is `Call`, which implement `Callable` and I cannot access `dag` everywhere, that's why I did this hack.
Do you have an idea to workaround that? Because adding `dag` in opts could also lead to other issues such as a nil dagql server... which can be catch but will lead to weird error no?

jedevc: Yup, adding it to the `Opts` sounds fine to me.

If the dagql server is `nil` we can error out to avoid panicking - yes, that's weird, so we should make sure to always pass it in. Where can't you access `dag`? Looking at all call-sites of `Call`, there's either a `dag` in the current scope, or you can easily propagate it down. For `Module.Initialize`, you can get it from `moduleSchema.dag`. For `objFun`, you need to propagate it through `functions`.

core/modulesource.go
```diff
@@ -261,6 +265,153 @@ func (src *ModuleSource) AutomaticGitignore(ctx context.Context) (*bool, error)
...
+		if err != nil {
+			return inst, fmt.Errorf("failed to get buildkit api: %w", err)
+		}
+
+		slog.Debug("moduleSource.LoadContext: loading contextual directory from local source", "path", path, "kind", src.Kind)
+
+		ctxPath, _, err := src.ResolveContextPathFromCaller(ctx)
+		if err != nil {
+			return inst, fmt.Errorf("failed to resolve context path: %w", err)
+		}
```
jedevc, line 314: There's a *very* subtle issue with using this in this way. If we have a `Directory` and we call `AsModule` on it, then the path defined by this module has *no corresponding context directory on the host*, so we need to instead get it from the `src.ContextDirectory()`.

This is a real concern for the builtin sdk runtimes (which use this). However, there's some also some *supreme* strangeness with how `ResolveContextPathFromCaller` works at all - specifically, it's a surprise it works at all. The following scenario on this PR breaks:

```
$ git init
$ mkdir ctxmod && cd ctxmod
$ dagger init --sdk=go --source=.
$ cat <<EOF > main.go
package main

import (
	"context"
)

type Ctxmod struct{}

func (t *Ctxmod) Dirs(
	ctx context.Context,
	// +defaultPath="/"
	root *Directory,
	// +defaultPath="."
	relativeRoot *Directory,
) ([]string, error) {
	res, err := root.Entries(ctx)
	if err != nil {
		return nil, err
	}
	relativeRes, err := relativeRoot.Entries(ctx)
	if err != nil {
		return nil, err
	}
	return append(res, relativeRes...), nil
}
EOF
$ dagger call -m . dirs
...
13  : Ctxmod.dirs: [String!]!
13  : Ctxmod.dirs ERROR [0.0s]
13  : ! failed to load contextual arg "root": failed to load contextual directory "/": failed to resolve context path: failed to stat source root "./ctxmod": failed to stat path: failed to receive file bytes message: rpc error: code = NotFound desc = rpc error: code = NotFound desc = stat path: lstat /home/jedevc/Documents/Projects/dagger-test-modules/ctxmod/ctxmod: no such file or directory
```

Effectively, we're applying the subpath from the root *twice*. This appears to be an issue from *before* this PR.

I think ResolveContextPathFromCaller's implementation is actually wrong:
- it first gets the root subpath (relative to the context)
- it then tries to look to find that root subpath in the current directory on the client (using StatCallerHostPath, which is relative to the current directory)
- but the current directory on the client is not neccessarily the subpath - the client ends up looking for <abspath-to-context>/subpath/subpath somehow

Again, this is a real issue, and needs solving before we can merge this PR. I'm happy to pick these up, *but* I think these take longer than we have before the v0.12 release - so I'm going to push these into the v0.12.1 milestone - in the interest of shipping *something* this week.

---

A few TODOs:
- A `LocalSource` should track *where* the context directory it's been loaded from is. This removes the complexity of this specific dance.
- We should create a `ScratchSource` type that can be created for `Directory.AsModule` - this type *doesn't* map to a directory on the host, so we should avoid all chances of confusion, which this PR steps into.

Some other notes from investigating:
- Why are there two implementations of `findUp`? One on the cli, one on the engine. This is *super* confusing, and isn't helping simplify file interactions.
- This extends into a "who is responsible for loading a module" problem - it's really unclear how the cli/server interact to load a module, with lots of jumping between cli/filesync/engine
	- @helderco has also noticed subtle bugs around this:
		> - One example is when using an "unbundled" sdk with --sdk and --source. The path to the SDK is calculated in relation to --source, which is wrong
		> - There's another issue where if a module fails to load, it reverts to a parent module (parent as in the filesystem tree)

jedevc (changes requested):

core/modulesource.go
sipsma: > Why are there two implementations of findUp? One on the cli, one on the engine. This is super confusing, and isn't helping simplify file interactions.
This extends into a "who is responsible for loading a module" problem - it's really unclear how the cli/server interact to load a module, with lots of jumping between cli/filesync/engine

The duplication between CLI+engine used to be 100x worse (essentially two parallel implementations of the entire process). 
* e.g. git sourced modules can have "local deps" (which are "local" to the git repo) and it used to be that there were completely separate implementations of loading those "local" deps and loading actually local (on the caller's disk) modules even though the process is 95% the same. Many other similar situations of duplication used to exist.

While we were getting ready for the modules release there were too many bugs that when fixed just spawned more bugs, so I started centralizing the logic in the server-side to help at least keep the complication in mostly one place, which helped at least put out the fires in our hair at that time.

The current state of things where there's still some remaining logic in the CLI is just because on a practical level I had hit the timebox on this whole area and the known fires were out. So no _good_ reason there's duplication remaining, just pragmatic ones from the past.

Point being, yes we should centralize the loading logic even more and get rid of the remaining split. Because the server-side *has* to be the place for a large subset of this logic (e.g. git sourced modules and their "local" deps), it's the only reasonable place to centralize it, so I think we should continue going down that path.

I'll just warn that even though I always imagined all this to be complicated going into it, it still always ended up being several times more complicated than I imagined 😄, so I would just expect this to take a bit to get working in every case. We have a shit-ton of integ tests for all this path/loading stuff, but I'm sure it's still missing quite a bit, so while fixing what remains here we likely need to upgrade to a _metric_ shit-ton of integ tests.

> I think ResolveContextPathFromCaller's implementation is actually wrong:
> 
> * it first gets the root subpath (relative to the context)
> * it then tries to look to find that root subpath in the current directory on the client (using StatCallerHostPath, which is relative to the current directory)
> * but the current directory on the client is not neccessarily the subpath - the client ends up looking for /subpath/subpath somehow

Yeah it does appear to be inconsistent whether `SourceRootSubpath` is made to be relative to the cwd or not
* We do it for abs paths here https://github.com/sipsma/dagger/blob/6d2bd214bcf6c722399ef820c025193eafd9da0e/core/schema/modulesource.go#L49-L49
* We do it during `dagger init` since it takes an explicit path to the new source: https://github.com/sipsma/dagger/blob/6d2bd214bcf6c722399ef820c025193eafd9da0e/cmd/dagger/module.go#L112-L112
* But don't appear to for user-supplied relative paths when just using the `-m` flag https://github.com/sipsma/dagger/blob/6d2bd214bcf6c722399ef820c025193eafd9da0e/cmd/dagger/module.go#L575-L575

So it seems like we could get away with just fixing that last point. _However_, this is a very good example of the remaining split between CLI+engine being confusing and requiring jumping all over the place. Centralizing more will help a lot.

---

I honestly think at this point we should just go all the way and add core APIs that do literally everything so that each relevant CLI subcommand (dagger init/install/develop/call) is just a one-line call to the core API.
* This aligns with the discussion in https://github.com/dagger/dagger/issues/6947#issuecomment-2135899729 around it making sense to turn the whole CLI into a stub for just calling core functions

TomChv:
I continued your research based on your repro @jedevc and that's super weird.

Here's the dump of some logs I get when I try your repro:

During module loading:

```
2024-07-19 15:12:58 time="2024-07-19T13:12:58Z" level=error msg=moduleSource.ResolveContextPathFromCaller.rootSubpath rootSubPath=.
2024-07-19 15:12:58 time="2024-07-19T13:12:58Z" level=error msg=moduleSource.ResolveContextPathFromCaller.sourceRootAbsPath sourceRootAbsPath=/Users/tomchauveau/Documents/github.com/quartz-technology/playground/e2e-tmp/git-ctx-dir/ctxmod
2024-07-19 15:12:58 time="2024-07-19T13:12:58Z" level=error msg=moduleSource.ResolveContextPathFromCaller.sourceRootAbsPath contextAbsPath=/Users/tomchauveau/Documents/github.com/quartz-technology/playground/e2e-tmp/git-ctx-dir contextFound=true sourceRootAbsPath=/Users/tomchauveau/Documents/github.com/quartz-technology/playground/e2e-tmp/git-ctx-dir/ctxmod
```

During context Dir loading:

```
2024-07-19 15:13:00 time="2024-07-19T13:13:00Z" level=error msg=moduleSource.ResolveContextPathFromCaller.rootSubpath rootSubPath=./ctxmod
2024-07-19 15:13:00 time="2024-07-19T13:13:00Z" level=warning msg="error resolving ctxmod.dirs: [String!]!" error="map[error:failed to load contextual arg \"root\": failed to load contextual directory \"/\": failed to resolve context path: failed to stat source root: failed to stat path: failed to receive file bytes message: rpc error: code = NotFound desc = rpc error: code = NotFound desc = stat path: lstat /Users/tomchauveau/Documents/github.com/quartz-technology/playground/e2e-tmp/git-ctx-dir/ctxmod/ctxmod: no such file or directory kind:*fmt.wrapError stack:<nil>]"
```

The `sourceRootPath` has change between the two operations, how is that possible?

core/module.go
TomChv: Ok I made the changes for that, was pretty easy with your hints!

sipsma:
@TomChv I mentioned here what we should need to do to fix the problem: https://github.com/dagger/dagger/pull/7744#discussion_r1671111212

> Yeah it does appear to be inconsistent whether SourceRootSubpath is made to be relative to the cwd or not
> * We do it for abs paths here https://github.com/sipsma/dagger/blob/6d2bd214bcf6c722399ef820c025193eafd9da0e/core/schema/modulesource.go#L49-L49
> * We do it during dagger init since it takes an explicit path to the new source: https://github.com/sipsma/dagger/blob/6d2bd214bcf6c722399ef820c025193eafd9da0e/cmd/dagger/module.go#L112-L112
> * But don't appear to for user-supplied relative paths when just using the -m flag https://github.com/sipsma/dagger/blob/6d2bd214bcf6c722399ef820c025193eafd9da0e/cmd/dagger/module.go#L575-L575
> So it seems like we could get away with just fixing that last point.

I think we can just patch that issue as described for now. The rest of the discussion is about larger changes that could simplify things overall, but not needed in the short term.

jedevc:
@sipsma's point is correct - that should fix the weird directory loading case.

But, also just want to repeat my point from above, I think we also need to understand how to fix the following:

> There's a _very_ subtle issue with using this in this way. If we have a `Directory` and we call `AsModule` on it, then the path defined by this module has _no corresponding context directory on the host_, so we need to instead get it from the `src.ContextDirectory()`.

TomChv:
The function `getExplicitModuleSourceRef` is pretty small, I tried some changes but it didn't fixed the issue.

It's related to where the user is, some buildkit func seems to depends on that so I'm trying to workaround it with a new function but it's not working as expected.
I'm doing a lot of context switch with #7864 which is really slowing my productivity, hopefully I get everything done by tomorrow

TomChv:
Ok so last friday we did a big exploration with @helderco to find the issue and it seems it's the related to the set of the `sourceRootSubPath` when calling `ResolveFromCaller` on the last initialization, the path is set to be relative to the context directory and no longer the user's location in its host (from where he calls the dagger CLI) which creates issues when resolving the path with `ResolvePathFromCaller`.

I'm going to fix this issue by updating this to keep the previous path instead of setting it to the relative.

https://github.com/dagger/dagger/blob/d9a15262229f1dba0aa53dbd0216908438331995/core/schema/modulesource.go#L893

TomChv:
The fails in the CI doesn't seem related to my PR, more flackiness:
- https://dagger.cloud/dagger/traces/c62c35a0fea4e7ee46b82e36677052de?span=2719a2b8918be39f#25daada6376b1058
- https://dagger.cloud/dagger/traces/c62c35a0fea4e7ee46b82e36677052de?span=2719a2b8918be39f#e3dca007344955fe

About the Test important, I'm super confused about it https://dagger.cloud/dagger/traces/4a3b8d288d9af71bcda3004c4c4c8637#76074fdc7901bd68, the `test all` works, but not `test important`??

core/modulesource.go
sipsma: I agree that we should not exclude `.git` by default in this exact place.

The place where `.git` got excluded previously was specifically when loading the module source that was given to SDKs for building the module. For SDKs, they don't (currently) care about `.git` so it was indeed a waste of time and invalidated cache more often than necessary. For that case specifically, it's probably worth keeping the default exclude of `.git`.

But in the general case of loading any contextual directory, I don't think it should be excluded by default. Like @jedevc said there are legit use cases for it.

core/modulesource.go
sipsma: There's an argument for doing blob conversion here in that it means contextual directories with equal contents will always be cached the same no matter their source. This can actually matter in cases like:
1. Local and git have same contents (i.e. local checkout of a git repo)
2. Getting the same directory from different git commits and/or forks of the same repo
3. Commonly reoccurring directories, "empty dir" being the most common I'd guess

But there's also overhead to blob-ification, so hard to say one is better than the other. If forced to choose I'd probably just leave this as is for now though.

core/modulesource.go
```diff
@@ -256,6 +271,203 @@ func (src *ModuleSource) AutomaticGitignore(ctx context.Context) (*bool, error)
...
+			return inst, fmt.Errorf("failed to load local module src: %w", err)
+		}
+
+		return inst, nil
+	case ModuleSourceKindGit:
+		slog.Debug("moduleSource.LoadContext: loading contextual directory from git", "path", path, "kind", src.Kind, "repo", src.AsGitSource.Value.HTMLURL())
+
+		// Use the Git context directory.
+		ctxDir := src.AsGitSource.Value.ContextDirectory.Self
+
```
sipsma, line 348: Shouldn't we have the same behavior as local modules where "absolute path is resolved relative to context" and "relative path is resolved relative to module source root" here for git dirs?

Otherwise you'd get different behavior for a given context dir path depending on whether you invoked the module from git or if you checked out the git repo locally and invoked the module locally, right?

TomChv: I'm not sure about that, if you use a gitsource, the  URL is the context dir of that source and it would also be the URL of the module.

`AsGitSource` doesn't provide any method to get a specific source root or anything, so I'm not sure how I could handle that

TomChv:
> There's an argument for doing blob conversion here in that it means contextual directories with equal contents will always be cached the same no matter their source. This can actually matter in cases like:

> Local and git have same contents (i.e. local checkout of a git repo)
Getting the same directory from different git commits and/or forks of the same repo
Commonly reoccurring directories, "empty dir" being the most common I'd guess
But there's also overhead to blob-ification, so hard to say one is better than the other. If forced to choose I'd probably just leave this as is for now though.

So what should I do, I'm a bit confused

sipsma:
> > There's an argument for doing blob conversion here in that it means contextual directories with equal contents will always be cached the same no matter their source. This can actually matter in cases like:
> 
> > Local and git have same contents (i.e. local checkout of a git repo)
> > Getting the same directory from different git commits and/or forks of the same repo
> > Commonly reoccurring directories, "empty dir" being the most common I'd guess
> > But there's also overhead to blob-ification, so hard to say one is better than the other. If forced to choose I'd probably just leave this as is for now though.
> 
> So what should I do, I'm a bit confused

GIthub's UI is making it hard to see, but to be clear that was a response to this earlier comment: https://github.com/dagger/dagger/pull/7744#discussion_r1668392984

But like I said at the end I'd just leave it as is, so no change.

helderco:
Just rebased.

helderco (commented):
There's several places that need to take `DefaultPath` into consideration now, in order to know if an argument is optional or not.

Most importantly, in the CLI and in the codegen of all SDKs. Also needs test coverage for this.

TomChv:
> There's several places that need to take DefaultPath into consideration now, in order to know if an argument is optional or not.

> Most importantly, in the CLI and in the codegen of all SDKs. Also needs test coverage for this.

Yeah, do you think it's doable in a follow up or it should be on this PR? I'm not sure I can update all the SDKs since the PR is already fairly big.

helderco:
> Yeah, do you think it's doable in a follow up or it should be on this PR? I'm not sure I can update all the SDKs since the PR is already fairly big.

Yeah, you you need to do it in this PR or it isn't ready. I'll add the Python codegen bit, you can handle Go and TypeScript.

TomChv:
Ohhh but it's already done in TypeScript

https://github.com/dagger/dagger/pull/7744/files#diff-4df8e2e3923a9ee4798a1fb594a383c303c92e3610e1f3e03a6a1f523ba244ecR143-R146

And in Go: https://github.com/dagger/dagger/pull/7744/files#diff-7733b13eee92afbeedff5903d1e7879b0454f9dbde3996c54ef6f70d65e11170R299

helderco:
> Ohhh but it's already done in TypeScript
> 
> https://github.com/dagger/dagger/pull/7744/files#diff-4df8e2e3923a9ee4798a1fb594a383c303c92e3610e1f3e03a6a1f523ba244ecR143-R146
> 
> And in Go: https://github.com/dagger/dagger/pull/7744/files#diff-7733b13eee92afbeedff5903d1e7879b0454f9dbde3996c54ef6f70d65e11170R299

You're right. It was late for me when I made that review trying it out locally. Got my wires crossed, sorry about that!

There's quite a few combinations of _has default_, _is nullable_, _has default path_, setting as optional even though it's not in code, etc... that it's better to offload all of this to the API, rather than every SDK having to get it right.

helderco:
The `+ignore` pragma needs to specify all patterns in a single line. Is it possible to make it multiline? Something like:

```go
func New(
	// +defaultPath="/"
	// +ignore=[
	//    "bin",
	//    ".git",
	//    "**/node_modules",
	//    "**/.venv",
	//    "**/dev/sdk",
	//    "**/__pycache__"
	//  ]
	source *dagger.Directory,
) (*DaggerDev, error) {
}
```

jedevc:
> I need to check how I can parse that but that must be possible.

Hm, you want `parsePragmaComment`. This is tricky, currently it's all assumed in one line with a regexp.

Being able to do opening/closing braces means the language is no longer regular, so we won't be able to use regexp for it. I'm not confident there's a simple way to do it (but I'm happy to do it in a follow-up) - I don't think we should delay this PR for that though.

TomChv:
> I'm not confident there's a simple way to do it (but I'm happy to do it in a follow-up) - I don't think we should delay this PR for that though.

Alright, I'm waiting for approvals and let's merge it :D

helderco:
There’s something wrong when depending on a module that uses a context directory.

Example:

```console
dagger-dev init --sdk=go test
dagger-dev init --sdk=go test/dep
```

```go
// test/dep/main.go
package main

import "github.com/dagger/dagger/test/dep/internal/dagger"

type Dep struct{}

func (m *Dep) Test(
	// +defaultPath="/"
	// +ignore=["!test"]
	source *dagger.Directory,
) *dagger.Directory {
	return source
}
```

```go
// test/main.go
package main

import "github.com/dagger/dagger/test/internal/dagger"

type Test struct{}

func (m *Test) Dep() *dagger.Directory {
	return dag.Dep().Test()
}
```

Calling *dep* directly uploads `test` from the absolute path for the context directory in my host:
```console
❯ dagger-dev -v call -m test/dep test

✔ Dep.test: Directory! 0.3s
  ✔ upload /Users/helder/Projects/dagger from itvcgmpmay0maeo073750fp1j (client id: nkb4aqx9f17ve8p2d73tloq65, session id: ke5jlcudbg0uy9zizohjm6e37) (include: test) 0.1s

_type: Directory
entries:
    - test
```

But calling as a dependency, tries to upload from `/scratch`:
```console
❯ dagger-dev -vvv call -m test dep

  ✔ Dep.test: Directory! 0.3s
    ✔ upload /scratch from kbzudk1n18ifj8vezs8o6h2jd (client id: w4cn43mbmxq5tublqyyvul7ae, session id: zv3pwm0sk60jn5fi6fa4apvrn) (include: test) 0.0s

_type: Directory
entries: []
```

TomChv:
Hmmm interesting, okay I need to dig into that! I didn't cover that kind of cases during my tests.

TomChv:
Ok so I've done some research into the engine and the issue make sense but is pretty complex to solve.

It seems that the environment dependencies are solved in different than the host's context.

For example, this is the values I get for a module called directly from host:

```shell
08-16 15:24:56 time="2024-08-16T13:24:56Z" level=error msg="ModuleSource.ResolveContextPathFromModule: get source root subpath" relHostPath=test/dep
2024-08-16 15:24:56 time="2024-08-16T13:24:56Z" level=error msg="ModuleSource.ResolveContextPathFromModule: stat caller host path" moduleRootAbsPath=/Users/tomchauveau/Documents/github.com/quartz-technology/playground/e2e-tmp/context-dep/test/dep
2024-08-16 15:24:56 time="2024-08-16T13:24:56Z" level=error msg="ModuleSource.LoadContext: resolveContextPathFromModule" ctxPath=/Users/tomchauveau/Documents/github.com/quartz-technology/playground/e2e-tmp/context-dep modPath=/Users/tomchauveau/Documents/github.com/quartz-technology/playground/e2e-tmp/context-dep/test/dep path=/
2024-08-16 15:24:56 time="2024-08-16T13:24:56Z" level=error msg="ModuleSource.LoadContext: importing local module source" excludes="[]" includes="[test]" path=/Users/tomchauveau/Documents/github.com/quartz-technology/playground/e2e-tmp/context-dep
```

^^^Here as you see, the path matches with my host tree `/Users/tomchauveau/xxxx`

And this is the result when a module call a dependency that has a contextual argument:

```shell
2024-08-16 15:21:19 time="2024-08-16T13:21:19Z" level=debug msg="moduleFunction.loadContextualArg: loading contextual directory" dir=/ fn=source
2024-08-16 15:21:19 time="2024-08-16T13:21:19Z" level=error msg="ModuleSource.ResolveContextPathFromModule: get source root subpath" relHostPath=
2024-08-16 15:21:19 time="2024-08-16T13:21:19Z" level=error msg="ModuleSource.ResolveContextPathFromModule: stat caller host path" moduleRootAbsPath=/scratch
2024-08-16 15:21:19 time="2024-08-16T13:21:19Z" level=error msg="ModuleSource.LoadContext: resolveContextPathFromModule" ctxPath=/scratch modPath=/scratch path=/
2024-08-16 15:21:19 time="2024-08-16T13:21:19Z" level=error msg="ModuleSource.LoadContext: importing local module source" excludes="[]" includes="[test]" path=/scratch
```

^^^Here, it's `/scratch`, because I think the context module isn't the host, but a container inside the engine that execute the module.

So, to fix this issue, we must find a way to insert the host context to the dep's container, however, I have no idea how to do such thing, I'll need some help to accomplish that.

@sipsma @jedevc @vito, do you have any idea?

sipsma:
@jedevc @TomChv I pulled this PR locally, will try the repro from @helderco and see if I can get a fix working (or at least an outline). 

Disclaimer: There's a reasonable likelihood I'm still missing a detail or just repeating what was stated above, still catching up on everything. 

But FWICT generally speaking the caching concerns around keeping track of client IDs in order to know where to pull module source from for context dirs feels like the same problem as the ones around secrets+sockets.

The general solution we have for this is:
1. Get a "content-defined key" for the resource
   * For secrets, it's the dagql ID digest for the call that creates the secret
   * For sockets, it's the dagql ID digest for the call that creates the socket
2. Put the "not content-defined" or mutable data in a "Store" that maps the above key to the out-of-band data. Stores are part of session+client state.
   * For secrets, we put the actual secret plaintext as the value in the store
   * For sockets, we put the metadata for the socket (path for unix, addr/port for ip) as the value in the store
3. Pass the "content-defined key" around the API calls (caching works because it's content defined) and when an API implementation needs to access the associated metadata, read it from the store using that key.

I think the solution to this problem has the same outline, except it's probably even simpler:
1. Use the module source digest as the "content-defined key" - we already rely on this digest's stability
2. Add a new "Store" (really just a map) to the session state that maps module source digest -> metadata required to load contextual dir
   * For local source modules, that metadata would include the client ID. For git source modules, just the dagql ID of the module source would suffice.
3. When setting up the module function call, ensure we get that module source digest and use it to look up the metadata in the store, loading the directory *synchronously* at that time and providing that dir as the argument.

Like I said though, trying this locally to ensure theory aligns with reality.

sipsma:
Update:

Have a rough implementation of the idea described in the [previous comment](https://github.com/dagger/dagger/pull/7744#issuecomment-2297277098).
* This was all said and done extremely simple, [diff here](https://github.com/dagger/dagger/commit/6ae66ffe0a086296975b65bc2058432aacd98e23)
* It's not beautiful yet, but fine. I have ideas on more aesthetically pleasing approaches but not worth it on top of this already large effort IMO. 
* **There's some fixes needed to ensure that the actual paths we use in each case are always based on the "callees" paths when a dep, but the important part of using the right client is fixed.**

After re-reading previous comments and catching up on discord conversations, I see also the issue where you could have:
1. multiple modules with the exact same digest due to having same name, same source code, etc.
2. *different contextual files/dirs* (i.e. not those loaded to initialize the module itself but that can be loaded by context dir args)
3. all loaded into the same session (via dagger-in-dagger)

In this type of case, using the module source digest would result in multiple modules having the same digest in the same session and thus could result in inconsistent contextual dir loads if there are differences between the modules.
* Let me know if a concrete example would be helpful here, I realize this is all a bit abstract

The thing is, I ***really*** struggle to imagine why we or users should ever care about this. It's such an absurdly obscure situation outside of our own integ tests that I personally can't justify us spending time worrying about it. Unless I'm missing something, I think it would be a huge waste of everyone's time.

Instead, we can just make some trivial adjustments to our integ tests if it is indeed an issue. Something as simple as giving each module a unique name would fix this (and tbh wouldn't be surprised if it fixes other flakes https://github.com/dagger/dagger/issues/8184).

---

To summarize:
1. I think we can go with the [diff](https://github.com/dagger/dagger/commit/6ae66ffe0a086296975b65bc2058432aacd98e23) that fixes deps to use the right client IDs. It avoids the cache invalidation problems.
   * There's separate fixes needed to ensure deps use the right paths, but should be a trivial addition on top of that.
2. There are corner cases that, unless I'm missing something, would be a big waste of time to care about.
   * The only thing I can imagine hitting it is our own integ tests and that can be fixed in trivial ways such as just using module names unique to each test.

cc @TomChv @jedevc @shykes @vito

sipsma:
Okay turns out that approach had problems involving caching that made it not worth salvaging. Basically, dagql caching resulted in clients not always getting registered and old clients from nested execs to be used when they had already exited.

Fortunately, I *think* found something much much simpler and more robust, pushed to this PR here: https://github.com/dagger/dagger/pull/7744/commits/b8220368ff27312986c4adac3df1de977360a639

The idea is that a chain of module calls involving transitive deps always looks something like:
1. Non-module client calling a module (i.e. `dagger call` on the host, or in a nested exec in the case of our integ tests)
2. Module client calling a dep
3. Dep being called (and possibly using contextual args)

If the dep is a local module and thus needs context loaded from local dirs, then the parent module must have also been loaded from the same local context. That parent module in turn must have been loaded from the original non-module client.

Basically, **to find the client to load local context from, just walk client callers backwards until you find a non-module client**.

The existing integ tests pass and I added a new one (just go so far) to cover this dep case, which also passes.

core/integration/module_test.go
```diff
@@ -6770,6 +6770,877 @@ func schemaVersion(ctx context.Context) (string, error) {
...
+import (
+	"dagger/dep/internal/dagger"
+)
+
+type Dep struct{}
+
+func (m *Dep) GetSource(
+	// +defaultPath="/dep"
+	// +ignore=["!yo"]
+	source *dagger.Directory,
```
sipsma, line 7533: nit: I think it's worth adding coverage of relative paths when setting `defaultPath` for deps in addition to this coverage of absolute paths. There's a lot of logic to get correct when it comes to resolving those paths so seems worth another test.

sipsma (approved):
One nit about more test coverage, otherwise LGTM!

core/modulesource.go
```diff
@@ -266,6 +282,208 @@ func (src *ModuleSource) AutomaticGitignore(ctx context.Context) (*bool, error)
...
+	}
+
+	switch src.Kind {
+	case ModuleSourceKindLocal:
+		bk, err := src.Query.Buildkit(ctx)
+		if err != nil {
+			return inst, fmt.Errorf("failed to get buildkit api: %w", err)
+		}
+
+		localSourceClientMetadata, err := src.Query.NonModuleParentClientMetadata(ctx)
```
jedevc, line 314: Note: I think this will do weird things if we load a module manually using `Directory.AsModule`? In that case, a module is loading another module, and so we'd be essentially escaping back to the top-level?

At the very least, we need a test for this. We can test it using `Module.Serve`, and then sending a raw-graphql query.

Not blocking - since this is an edge case, but we should definitely cover it.

core/integration/module_test.go
```diff
@@ -6770,6 +6770,877 @@ func schemaVersion(ctx context.Context) (string, error) {
...
+
+
+	return dag.LoadDirectoryFromID(dagger.DirectoryID(directoryIDRes.Dep.GetSource.ID)), nil
+}
+			`,
+			)
+
+		out, err := ctr.With(daggerCall("get-dep-source", "--src", "./dep", "entries")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "yo\n", out)
```
jedevc, line 7640: I *think* this test is technically passing, but I have a hunch it might be wrong.

So, what if we have a dep module that *doesn't* exist on disk, and we created the directory source inside the `GetDepSource` func? I think it would still read from the disk of the parent.

...I'm not sure that's right, but I'm also not convinced what the right behavior *should* be. Potentially I wonder if we should just not allow a context directory at all in this case (or fallback to `.ContextDirectory`), since there's not really a corresponding client it was loaded from here. cc @sipsma 

No need to block here, since it's a minor point in a relatively unused API (but note, we use it in our CI!).

sipsma: Are you referring to the case where the `Directory.asModule` API is used with a directory that was created with some set of arbitrary API calls?

If so, yeah that wouldn't work atm. The logic for loading context dirs/files is based on the Kind of the module source (e.g. [here](https://github.com/TomChv/dagger/blob/0e3ab2e05c6ef4e28ec6f79b6188bcc12d96ff88/core/modulesource.go#L307)), so I think this just means that modules created from arbitrary directories should be marked as their own separate kind from `Local`/`Git` (need to be marked as a different kind [here](https://github.com/TomChv/dagger/blob/0e3ab2e05c6ef4e28ec6f79b6188bcc12d96ff88/core/schema/module.go#L700)).

Then context can be loaded from them. I think the logic would be almost identical to the logic for loading from git.

Either way, agree this is extremely obscure, doesn't need to be a blocker.

jedevc: > Are you referring to the case where the `Directory.asModule` API is used with a directory that was created with some set of arbitrary API calls?

Exactly this.

> I think this just means that modules created from arbitrary directories should be marked as their own separate kind

I suggested a `ScratchKind` for this back in 
https://github.com/dagger/dagger/pull/7744#discussion_r1670959469, can look at that at some point :tada:

core/integration/module_test.go
TomChv: Done in [79e917c](https://github.com/dagger/dagger/pull/7744/commits/79e917cb5a27c2baaa07b013b17e1688264ac003), with fixes on relative dep module resolution normally.

Transmitt0r:
Would love to have this also supported in go sdk :)

jedevc:
It is supported in the go sdk. See the go examples in https://docs.dagger.io/manuals/developer/functions/#directories-and-files.

Transmitt0r:
Thanks, I was confused by https://docs.dagger.io/manuals/developer/entrypoint-function#default-values-for-complex-types

PR #7746, opened 2024-06-24
sdk: Bump engine dependency to v0.11.9

sipsma:
Have to manually create this for now so it's based on the `release-v0.11.9` branch, see https://github.com/dagger/dagger/pull/7745#issuecomment-2187636760


sipsma:
Hit https://github.com/dagger/dagger/issues/7662 in testdev but other engine tests passed, so just re-rolling since it's a known flake.

The macos provisioning tests passed before I pushed the separate commits that update the sdk release notes etc. but now they are getting stuck on installing docker (before dagger is involved at all)... I'll give them another re-roll for purity's sake but otherwise will just merge as is since they passed before.

Issue #7711, opened 2024-06-20
Dagger cache can fill the disk in some situations


shykes:
## Problem

In some situations, the Dagger cache can fill up a machine's entire disk. There is a garbage collector to prevent this, but under certain circumstances the disk can fill up anyway.

## Solution

Design a more robust system for preventing the cache from filling the disk.

shykes:
FYI @jedevc . I left the wording very vague, because I didn't want to get the details wrong. But this is about the buildkit gc and how we use it. Including 1) the change of gc threshold from 10% to 75% of total disk space, and 2) whether the "percentage of total disk space" heuristic is adequate.

jedevc:
Opened a proposal with a sketch of some example code as a possible way to avoid the "percentage of total disk space" heuristic that we've encountered issues with: https://github.com/moby/buildkit/pull/5079

jedevc:
Looks like part of this was also related to https://github.com/dagger/dagger/pull/7738.

But going to keep moving the upstream PR, I think that's the road to much better defaults over all.

vikram-dagger:
This took 11m and ate a significant percentage of the disk

```
admin@ip-172-31-46-122:~$ df
Filesystem     1K-blocks     Used Available Use% Mounted on
udev             1997996        0   1997996   0% /dev
tmpfs             401904      592    401312   1% /run
/dev/xvda1      24532104 15518220   7913608  67% /
tmpfs            2009508        0   2009508   0% /dev/shm
tmpfs               5120        0      5120   0% /run/lock
/dev/xvda15       126678    11816    114862  10% /boot/efi
tmpfs             401900        0    401900   0% /run/user/1000
admin@ip-172-31-46-122:~$ dagger init --name=multipush --sdk=typescript
Full trace at https://dagger.cloud/dagger/traces/7d9add45978fe10b448fd46c86366778

✔ connect 1.3s
✔ cache request: mkfile /schema.json 0.0s
✔ load cache: mkfile /schema.json 0.0s
✔ cache request: blob://sha256:84f5e5fc2e7aaed03bfe831f1566ddfcd56170a8aff09fc4b820397f3f22ba79 0.0s
✔ load cache: blob://sha256:84f5e5fc2e7aaed03bfe831f1566ddfcd56170a8aff09fc4b820397f3f22ba79 0.0s
✔ moduleSource(refString: "."): ModuleSource! 0.0s
✔ ModuleSource.kind: ModuleSourceKind! 0.0s
[...]

PR #7749, opened 2024-06-25
fix: ensure gc policy percentage is correct

jedevc:
When originally updating this in #7563, we only updated the default CLI option - however, this is confusing, and isn't exactly how that works - my suggested refactor wasn't actually correct.

Because we use `GlobalIsSet`, this default is never actually the one that's used. So, we can refactor this, and set a global const to define the default gc policy ourselves, outside of buildkit.

This also opens us up to modifying the gc policy in more detail in the future.

### DIFF
--- a/cmd/engine/main.go
+++ b/cmd/engine/main.go
@@ -212,7 +212,7 @@ func addFlags(app *cli.App) {
 			Value: func() int64 {
 				keep := defaultConf.Workers.OCI.GCKeepStorage.AsBytes(defaultConf.Root)
 				if keep == 0 {
-					keep = config.DiskSpace{Percentage: 75}.AsBytes(defaultConf.Root)
+					keep = config.DiskSpace{Percentage: server.DiskSpacePercentage}.AsBytes(defaultConf.Root)
 				}
 				return keep / 1e6
 			}(),
--- a/engine/server/gc.go
+++ b/engine/server/gc.go
@@ -140,7 +140,7 @@ func getGCPolicy(cfg config.GCConfig, root string) []bkclient.PruneInfo {
 		return nil
 	}
 	if len(cfg.GCPolicy) == 0 {
-		cfg.GCPolicy = config.DefaultGCPolicy(cfg.GCKeepStorage)
+		cfg.GCPolicy = defaultGCPolicy(cfg.GCKeepStorage)
 	}
 	out := make([]bkclient.PruneInfo, 0, len(cfg.GCPolicy))
 	for _, rule := range cfg.GCPolicy {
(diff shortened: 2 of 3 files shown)


cmd/engine/main.go
```diff
@@ -212,7 +212,7 @@ func addFlags(app *cli.App) {
 			Value: func() int64 {
 				keep := defaultConf.Workers.OCI.GCKeepStorage.AsBytes(defaultConf.Root)
 				if keep == 0 {
```
gerhard, line 214: My understanding is that we change [BuildKit's 10% default]( https://github.com/moby/buildkit/blob/master/cmd/buildkitd/config/gcpolicy_unix.go#L10) here and set it to `75%`, as specified in `gc.go` . How do we know that this is a safe default? Would `50%` be more appropriate here?

marcosnils (approved):

engine/server/gc.go
```diff
@@ -153,3 +153,33 @@ func getGCPolicy(cfg config.GCConfig, root string) []bkclient.PruneInfo {
...
+func defaultGCPolicy(keep config.DiskSpace) []config.GCPolicy {
+	if keep == (config.DiskSpace{}) {
+		keep = config.DiskSpace{Percentage: DiskSpacePercentage}
+	}
+	return []config.GCPolicy{
+		// if build cache uses more than 512MB delete the most easily reproducible data after it has not been used for 2 days
+		{
+			Filters:      []string{"type==source.local,type==exec.cachemount,type==source.git.checkout"},
+			KeepDuration: config.Duration{Duration: time.Duration(48) * time.Hour}, // 48h
+			KeepBytes:    config.DiskSpace{Bytes: 512 * 1e6},                       // 512MB
```
gerhard, line 166: How did we pick `48h` & `512MB` ?

engine/server/gc.go
```diff
@@ -153,3 +153,33 @@ func getGCPolicy(cfg config.GCConfig, root string) []bkclient.PruneInfo {
...
+		},
+		// remove any data not used for 60 days
+		{
+			KeepDuration: config.Duration{Duration: time.Duration(60) * 24 * time.Hour}, // 60d
```
gerhard, line 170: Why `60d` ?

gerhard (commented):
I am trying to better understand the values that we picked:
- `75%` for the default
- `48h` & `512MB`
- `60d`

This feels like an important change that we should capture in a doc somewhere, and also offer a config example for those that want to change this. That brings me to another question: can users change this via config?

jedevc:
@gerhard these are pulled directly from buildkit - these are the values we are already using today.

https://github.com/moby/buildkit/blob/05213dbc2ff1a97360d2f91c6d22a8f1f6201d68/cmd/buildkitd/config/gcpolicy.go#L71-L97

cmd/engine/main.go
jedevc: Assuming the GC correctly kicks in (fixed in https://github.com/dagger/dagger/pull/7738), then any value here should be good. I've had no issues with any value here in all my time running buildkit.

Potentially this will result in a large space usage from dagger - but IMO, this is a better result than aggressive pruning that results in poor performance.

There is ongoing work to improve the defaults by introducing new heuristics upstream, see https://github.com/dagger/dagger/issues/7711#issuecomment-2186790489 for context on our side.

cc @marcosnils

sipsma (approved):
Pulled this locally to my WIP PR for adding integ tests around all of this by adding queries to the API that return the current GC config and it does indeed fix the percentage. Previously was still seeing 25GB on my 250GB disk, now it's 187GB:
```
sipsma@dagger_dev:~/repo/github.com/sipsma/dagger$ echo '{engineState{pruneConfig}}' | ./hack/with-dev ./bin/dagger query
~/repo/github.com/sipsma/dagger/ci/mage ~/repo/github.com/sipsma/dagger
~/repo/github.com/sipsma/dagger
✔ connect 0.2s
✔ ModuleSource.asModule: Module! 8.3s
✔ Module.initialize: Module! 1.0s
✔ Module.serve: Void 0.0s
✔ engineState: EngineState! 0.0s
✔ EngineState.pruneConfig: [PruneConfig!]! 0.0s

{
    "engineState": {
        "pruneConfig": [
            {
                "All": false,
                "Filter": [
                    "type==source.local,type==exec.cachemount,type==source.git.checkout"
                ],
                "KeepBytes": 512000000,
                "KeepDuration": 172800000000000
            },
            {
                "All": false,
                "Filter": null,
                "KeepBytes": 187000000000,
                "KeepDuration": 5184000000000000
            },
            {
                "All": false,
                "Filter": null,
                "KeepBytes": 187000000000,
                "KeepDuration": 0
            },
            {
                "All": true,
                "Filter": null,
                "KeepBytes": 187000000000,
                "KeepDuration": 0
            }
        ]
    }
}
```

---

RE: for how to config this, for now it will still only be possible via either `/etc/dagger/engine.toml` or cmdline args to the engine container. We can consider adding it to this internal doc for now: https://github.com/dagger/dagger/blob/f00d48c89708942fbcdee9a29961d1e433597b8c/core/docs/d7yxc-operator_manual.md

But API-ifying all this will open up the possibility to make this configurable via the CLI (including at runtime). Not pursuing that in immediate term since there's more implications to making it mutable rather than just queryable, but I think that's what we should be aiming for in time as the "official way".

PR #7750, opened 2024-06-25
ci: add actions check for checking changelog

jedevc:
This PR adds a check that if a PR is labelled with `needs/changelog`, then the job will fail and add a comment if no changelog was generated as part of the PR.

This should help us automate some of our changelog writing.

See an example run of this in https://github.com/jedevc/dagger/pull/5.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,46 @@
+name: needs/changelog
+
+on:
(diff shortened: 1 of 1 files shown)


gerhard (approved):
This is a very nice improvement! Our users will definitely appreciate better changelogs for our releases!

PR #7751, opened 2024-06-25
chore: remove `checkVersionCompatibility` field

jedevc:
Part of #7640 and co.

With changes to how we do version compatibility checks, we no longer need this check in the engine. Thankfully - we can just remove it: all of the SDKs were treating this as a warning level, and not causing a fatal error - this means that if the method is removed, it should not cause a build to fail.

The Go SDK also removes the `WithoutVersionCompatibilityCheck` - since now this option is always the case.


helderco (approved):
LGTM, just needs to fix linter.

Issue #7722, opened 2024-06-23
Typescript / Node examples are different than the generated code, thus confusing


sporto:
### What is the issue?

When following the getting started with Dagger using the typescript SDK. You get a file which uses a class with decorators.

I was trying to create a db service. Which needs to mount a host directory.

Then looking at https://github.com/dagger/dagger/tree/main/examples/sdk/nodejs, you find these examples using a very different approach i.e. using `connect`. This is confusing, not very clear how you would translate what you see in those examples to the class approach generated. 

Specifically, `client` in those examples. Where is `client` in the generated class? It would be better if the example resemble the generated code by the cli closely.

PR #7752, opened 2024-06-25
chore: remove examples directory

jedevc:
Fixes https://github.com/dagger/dagger/issues/7722.

The examples here are not particularly discoverable, and they're also mostly out of date.

As @shykes suggested in #7722, if we miss any of the content, we can bring it back in the cookbook.


gerhard (approved):
👍 🚀

PR #7754, opened 2024-06-25
fix: plain progress should never use hyperlinks

jedevc:
Github actions doesn't support them :cry:

See https://github.com/openmeterio/openmeter/actions/runs/9572449099/job/26391770769?pr=1031#step:3:2161.

In the raw logs, we can see:

> 2024-06-18T21:40:49.1751535Z [1mFull trace at [0m]8;;https://dagger.cloud/\https://dagger.cloud/]8;;\ (rotate dagger.cloud token for full url)

We just can't have hyperlinks here, I missed this before.

---

I've also added a sanity test for the `DAGGER_CLOUD_TOKEN` format, since I was suspicious that this might not be doing the right thing.


gerhard (approved):
👍 🚀

vito (approved):

sipsma (approved):

PR #7758, opened 2024-06-26
chore: add changelogs for attachable terminal sessions

jedevc:
#7586 introduces breaking changes, and adds features - so this should be changelogged.


aluzzardi (approved):
🙏

PR #7759, opened 2024-06-26
Dynamically serve module schemas based on engineVersion

jedevc:
As discussed in https://github.com/dagger/dagger/issues/7640.

The essential idea here is to use the `engineVersion` defined in `dagger.json` to determine any backwards compatible schemas to serve. This will allow us to make breaking changes in a more controlled manner.

Without this, any breaking changes we make to the API will be applied everywhere - this will potentially break parts of the daggerverse, introduces module dependencies on specific dagger versions, and will prevent users from easily upgrading (essentially, a python 2/3 situation)

With this, we can introduce breaking changes more gracefully using the `engineVersion`. At schema install time we can adapt the schema we choose based on the value of the defined `engineVersion`. This means that *new* and *upgraded* modules will be served the new API, while old modules will be served an older API for compatibility.

We still will likely need to make breaking changes outside of this - for that, we have https://github.com/dagger/dagger/pull/7692, but ideally we won't need to actually enforce this for some time.

---

TODO:
- [x] Implementation
- [x] Cleanup
- [x] Implement this for #7586
- [x] Implement this for #7500
- [x] Commit cleanup


jedevc:
Decided to take this in a totally different direction - having and managing multiple `dagql.Server`s just wasn't going to work, so I've decided to introduce the lower-level concept of dagql "views", that allow different components to have different perspectives on the same server.

That way, we have one core module (instead of multiple), which prevents us needing to work out how to convert an ID value from one schema into another schema (doesn't actually seem possible). e.g. this would have broken if we chose to remove a field from `Container`, and an old module was using it and returning it to another module with a newer version.

This now works!

Some more TODOs:
- Introspection is probably a little ambitious, it needs scoping to views.
- Testing approach needs rethinking - the `schemaVersion` helper I was using doesn't really work with our more powerful views helper.
	- :bulb: we could probably get quite a lot done with a lower-level dagql test, and having our compat checks will cover a good number of issues
	- But it would still be nice to simulate generic removal/modification/addition of various schema things.
- Huge cleanup needed, I broke a ton of weird things while playing around
- Ideally, we'd also support removing/adding new objects (currently they need to be added to the "global view"), so we wouldn't be able to remove `Terminal` just yet.

jedevc:
I'm getting to a place where I'm much happier with this. I've also backported legacy conversions (with tests!) for the already merged breaking changes https://github.com/dagger/dagger/pull/7586 and https://github.com/dagger/dagger/pull/7500.

Going to spend some time tomorrow cleaning up the commit history / checking test failures / etc.

But this *should* be somewhat ready-to-review (ignoring the above).

sipsma (commented):
Just perused on a high-level so far but looks good!

Double-checking understanding of a few things:
1. When version is not semver (i.e. git commit or content hash of local dev engine), we just treat that as "latest" essentially
2. In the case where you have a local module and you do `dagger develop`:
   * The engine version gets updated first
   * *Then* we compile the module with that latest version (erroring out if it's broken now)

I'll come back for more in-depth review, just getting this out first.

---

Thought for the longer-term: 

The views approach makes sense to me for the reasons you mentioned around "id from one version to another version", but it is a bit complicated in that, in a given session, each client gets its own dagql server but then each dagql server also has support for multiple views.
* Maybe this could be rectified in time though by also using views to serve different schemas to different clients depending on which modules have been installed, their deps, etc.?
* So then the idea would be "one dagql server for the whole session" but different views for different clients (on top of views based on version of the client)
* And then in the even longer term if we get cross-session dagql caching working (mostly just need LRU-type logic I think), the entire engine can just have a single dagql server that runs indefinitely

That could allow these views to actually simplify everything relative to today.

Just planting a potential seed there, seems like way too much of a scope expansion to tackle in this PR.

dagql/call/callpbv1/call.proto
```diff
@@ -80,6 +80,12 @@ message Call {
   // A unique digest of this Call. Note that this must not be set when
   // calculating the Call's digest.
   string digest = 9;
+  
+  // The view that this call was made in. Since a graphql server may present
+  // slightly different views depending on the specified view, to be able to
+  // evaluate calls wherever, we need to track which view the call was actually
+  // called in.
+  string view = 10;
```
sipsma, line 88: One thought I just had, since this ends up in IDs it will impact the caching of function calls (once that's in place). Right? 

If so, while It's a bit complicated to think about, I *think* that's sometimes exactly what we'd want and other times not ideal.

E.g. if you received an ID as an arg from a module on an older engine version that involves back-incompat changes, it probably makes sense the ID would result in a different cache key (I think...).
* *But* if there's modules on say v0.12.0 and v0.12.1 and there were no breaking changes there, then the different versions would result in different cache keys for the function call unnecessarily

Not even 100% sure yet this is true and even if it is, I can't tell how big an issue it would be or the best way to fix it yet. 

So not a blocker, just calling out in case others have more thoughts.

jedevc:
Updated with a neater commit history :tada:

> When version is not semver (i.e. git commit or content hash of local dev engine), we just treat that as "latest" essentially

Yes - hence #7760 to try and fix this for dev builds. Thankfully, I think users using git commit versions is *pretty* rare, but I'm gonna try and see if I can get this one done for v0.12.0 if I *possibly* can :tada:

> In the case where you have a local module and you do `dagger develop`:
>     * The engine version gets updated first
>     * _Then_ we compile the module with that latest version (erroring out if it's broken now)

Ah, not right now - the version is updated last. I need to poke into this logic, it's a bit odd. Will take a look at this, thanks for the reminder!

---

> The views approach makes sense to me for the reasons you mentioned around "id from one version to another version", but it is a bit complicated in that, in a given session, each client gets its own dagql server but then each dagql server also has support for multiple views.

Thankfully, for this all to work, we don't need one server, we need one core "schema" (even if that's loaded into multiple servers). That way, IDs from one server can just be copied to the other server, even if they're serving different views.

In my original pass at this, I actually just attempted to have each session have a server, and each server would have it's own "schema" (so selectively installing components based on a version). This *doesn't work at all* - because now IDs that one module produces and returns have no way of being translated into an ID for another schema - and since IDs are so central to how we do everything, there's really no way to resolve this neatly.

So we get the state this is in right now.

> So then the idea would be "one dagql server for the whole session" but different views for different clients (on top of views based on version of the client)

This is really funky, I like it, it did cross my mind. The problem is now, a "View" is now an intersection of various conditions, the core module uses a version, user modules use set inclusion in a set of deps (or similar). That kinda makes sense, I've got a generic `View` interface for this, you just to implement `Contains(string)` - in the future, we could update this to be `Contains(T)` with a type parameter `T`? Then for a dagger server, `T = Module`, so we could derive the config and the dependencies?

Perhaps this isn't too tricky actually? Definitely seems like it could simplify some code, but probably worth keeping it separate for now, I've got some other pieces to squeeze in before the next release.

dagql/call/callpbv1/call.proto
jedevc: Mm, I hadn't thought about this :thinking: But yes, this would break cache.

Suppose we had two modules build containers, but one is doing so with schema X, and the other does so with schema Y. Even if they do exactly the same thing, they're gonna have different IDs - which would impact module caching (but wouldn't impact any `WithExec` caching, since the underlying LLB should be the same). I do think we want the IDd behavior for safety, it's not safe to just always alias these as you point out.

Aside: frustratingly, if we wanted to avoid this behavior it would be a total pain - there's really not anywhere else we can attach this info. There needs to be someway to link IDs to where they were constructed.

One option that might not suck - if a field is a member of the default global view (no view has been defined for it), then we can set the ID's view to the empty string `""`. Then every view of the server will have the same IDs for those fields, *but* we'd need to undo this logic (which requires a bit more explicitness in schema defs, but not too bad): https://github.com/dagger/dagger/blob/12ca11a58c3d84275b5b8d04cd17e1735a3b4fa5/dagql/objects.go#L95-L97

We'd need to undo that, to actually *ensure* that every view into the schema always gets the same results for items in the global view.

At least this way, only the incompatible APIs would break cache (exactly what we want). Will give this a go to see if this is at all feasible.

jedevc: Added in [`b092e39` (#7759)](https://github.com/dagger/dagger/pull/7759/commits/b092e39d3227e0f2e98196913a8eef6b637049c7).

core/schema/modulesource.go
```diff
@@ -269,17 +269,23 @@ func parseRefString(ctx context.Context, bk buildkitClient, refString string) pa
...
+	args struct {
+		EngineVersion dagql.Optional[dagql.String]
+	},
 ) (inst dagql.Instance[*core.Module], err error) {
+	withSourceInputs := []dagql.NamedInput{
+		{Name: "source", Value: dagql.NewID[*core.ModuleSource](src.ID())},
+	}
+	if args.EngineVersion.Valid {
+		withSourceInputs = append(withSourceInputs, dagql.NamedInput{Name: "engineVersion", Value: args.EngineVersion})
+	}
```
jedevc, line 281: We *should* just be able to unconditionally attach this, but something goes bizarrely wrong if we do, and we just hang during a module call.

I spent ages trying to understand this. Apparently, if I attach the the arg *directly*, modules freeze up on initialization, but with no stack trace in an obviously guilty place and no errors anywhere in the logs. I checked with dagql, and there's nothing obviously incorrect about attaching an invalid optional like this.

No idea *why* or what could possibly be happening here... but this is a quick fix to unblock this. I wonder if maybe there's something potentially about maybe creating a module with a slightly different ID? cc @vito @sipsma maybe something stands out here?

vito: Yeah this might have sent you down a dusty corner of optional/`null` handling. I think this would result in an ID that has `"engineVersion": null` in it. Might be wise to have our ID construction just skip null optionals since it'll be considered different from the arg not being provided, for no good reason.

No idea what's up with the hang though, from a distance.

dagql/objects.go
```diff
@@ -78,22 +78,58 @@ func (class Class[T]) IDType() (IDType, bool) {
...
+}
+
+func (class Class[T]) fieldLocked(name string, views ...string) (Field[T], bool) {
+	fields, ok := class.fields[name]
 	if !ok {
 		return Field[T]{}, false
 	}
-	return *field, ok
+	for i := len(fields) - 1; i >= 0; i-- {
+		field := fields[i] // iterate backwards, since
```
sipsma, line 93: think this comment got cut off

sipsma (approved):
Just some nits, LGTM!

cmd/dagger/module.go
```diff
@@ -68,6 +69,7 @@ func init() {
 	moduleInitCmd.Flags().StringVar(&moduleName, "name", "", "Name of the new module (defaults to parent directory name)")
 	moduleInitCmd.Flags().StringVar(&moduleSourcePath, "source", "", "Directory to store the module implementation source code in (defaults to \"dagger/ if \"--sdk\" is provided)")
 	moduleInitCmd.Flags().StringVar(&licenseID, "license", defaultLicense, "License identifier to generate - see https://spdx.org/licenses/")
+	moduleInitCmd.Flags().StringVar(&engineVersion, "engine-version", "latest", "Engine version to update code to")
```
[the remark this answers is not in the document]
helderco: I think we shouldn't provide the flag. Always use latest for new modules.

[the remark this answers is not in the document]
jedevc: Yeaaah, agreed with @helderco here. I've left the API changes in, we can add this flag back in if we really feel it's missing.

We should be trying to nudge users to upgrading, and having this backwards compat super visible in this way might tempt users to *not* do this.

PR #7761, opened 2024-06-26
DagQL ID optimizations

vito:
* Detect when a query's sole purpose is to get an ID, and just statically return it rather than evaluating each intermediate call
* Look deeply within a query for ID inputs and evaluate them all in parallel


jedevc:
Just a quick note I raised on the call - since some users are using `ID` as a way to force evaluation on arbitrary objects (sort of, it's super hacky), then if we do this, we should probably also make `Sync` automatically installed on every object (another breaking change, yay).

PR #7765, opened 2024-06-26
cli: add quickstart for for `dagger login`

marcosnils:
this commits triggers the quickstart flow when the user which is login
in does not belong to any organization. It'll redirect the user to
Dagger Cloud so the organization creation flow can continue there.

### DIFF
--- a/cmd/dagger/cloud.go
+++ b/cmd/dagger/cloud.go
@@ -2,8 +2,12 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
+	"time"
 
+	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
 
 	"github.com/dagger/dagger/internal/cloud"
@@ -36,8 +40,7 @@ func init() {
 	rootCmd.AddCommand(logoutCmd)
 }
 
-type CloudCLI struct {
-}
+type CloudCLI struct{}
 
 func (cli *CloudCLI) Client(ctx context.Context) (*cloud.Client, error) {
 	return cloud.NewClient(ctx)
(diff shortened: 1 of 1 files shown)


aluzzardi (approved):

cmd/dagger/cloud.go
```diff
@@ -104,6 +115,40 @@ func (cli *CloudCLI) Login(cmd *cobra.Command, args []string) error {
...
+		fmt.Fprintf(w, "Unable to open browser automatically, please visit %s to create an organization.\n", url)
+	}
+
+	timer := time.After(15 * time.Second)
+	t := time.NewTicker(1 * time.Second)
+
+	defer t.Stop()
+
+	for {
+		select {
```
jedevc, line 131: Missing a `case <-ctx.Done`.

cmd/dagger/cloud.go
```diff
@@ -104,6 +115,40 @@ func (cli *CloudCLI) Login(cmd *cobra.Command, args []string) error {
...
 }
 
+func createNewOrg(ctx context.Context, cli *cloud.Client, w io.Writer) (*auth.Org, error) {
+	url := "https://dagger.cloud/signup?quickstart=true"
+	err := browser.OpenURL(url)
+	if err != nil {
+		fmt.Fprintf(w, "Unable to open browser automatically, please visit %s to create an organization.\n", url)
+	}
+
+	timer := time.After(15 * time.Second)
```
jedevc, line 125: Is 15 seconds long enough to go through the whole quickstart? Shouldn't this be a couple of minutes?

marcosnils: this timeout is not for the whole quickstart, but a sensible number to wait until the browser gets opened (or the user opens it if that's not possible) and the organization is created. We call this the "quickstart" flow but the gist of it is that it streamlines account creation basically

jedevc (approved):

PR #7766, opened 2024-06-26
Pipeline cleanup

shykes:
This PR combines cleanup tasks to make our development pipelines more pleasant to use and improve:

- Rename `./ci` to `./dev` (to match SDK dev)
- Move `./dagger.json` to `./dev/dagger.json` (to match SDK dev)
- Breaking change: repo root is no longer our CI module. '-m dev' is required.
- Remove intermediary `std/` directory. `std/go` becomes `go`, and so on.
- `dirdiff/` is no longer a weird special case
- move `linters/markdown` to `dev/markdown`
- Shorten argument lengths to make checks more readable
- Scrub "CI" from variable and type names


shykes:
I seem to be hitting a failure because of custom logic in our GHA workflows.

shykes:
```
  set -x
  if [[ "dev" == "." ]]; then
    # set some sane defaults for the current module
    if [[ -f $HOME/.docker/config.json ]]; then
      dagger -m . call --source=".:default" --host-docker-config=file:"$HOME/.docker/config.json" docs lint
    else
      dagger -m . call --source=".:default" docs lint
    fi
  else
    dagger -m "dev" call docs lint
  fi
```

image

shykes:
#7647 can't come soon enough...

dev/go/lint.json
jedevc, line 1: :eyes: What's this file?

jedevc:
> Breaking change: repo root is no longer our CI module. '-m dev' is required.

Hm, not the biggest fan of moving `dagger.json` to dev, but I can live with it - I know you feel strongly about this one.

Is anyone *currently* bundling dagger using our dagger function? If so, we will need to update there - cc @jpadams.

> Remove intermediary std/ directory. std/go becomes go, and so on.

Hm. The problem is, now we have some subdirectories of `dev` that are just util go packages like `mage`, `build` and `const`, while others are dagger modules, which makes it trickier to understand while reading.

I'd prefer to keep these in separate places. Maybe we could have a top-level `std` and a top-level `dev`? That also then starts to make this really part of an "official" standard library I guess.

Another alternative is to do more refactoring to get rid of the go subpackages, but that is definitely trickier for `build` (and we might want to bring them back one day anyways).

Or we could just leave it as-is in `main` for now.

---

Just a note for follow-ups - I'd like to make sure we're using our "std" modules, like dirdiff consistently. ATM, we're not good examples of how to do this, since we have mixed use of dirdiff *and* `util.DiffDirectoryF`.

jedevc:
Pushed a commit to resolve the go package issues, were you able to run this locally @shykes? It wasn't just failing in CI for me, I was having this issue locally.

shykes:
> > Breaking change: repo root is no longer our CI module. '-m dev' is required.
> 
> Hm, not the biggest fan of moving `dagger.json` to dev, but I can live with it - I know you feel strongly about this one.

My reasoning was:

- We should be consistent across our repo
- Our root module and SDK modules are not consistent (`ci` + source dir for one; `dev` + submodule for the others). We should pick one
- The SDKs cannot adopt the "ci/ + source dir" pattern, because it doesn't work when you're developing a dagger module - in that case you end up with 2 modules mixed into a single dagger.json
- However the root module *can* adopt the "dev + submodule" pattern

> > Remove intermediary std/ directory. std/go becomes go, and so on.
> 
> Hm. The problem is, now we have some subdirectories of `dev` that are just util go packages like `mage`, `build` and `const`, while others are dagger modules, which makes it trickier to understand while reading.

I don't think that will be a problem in practice. The way I see it, this is a Dagger module first, and a Go codebase second. Dagger submodules shouldn't be "parked" in a special subdirectory to make space for Go sprawl. If anything I would prefer the other way around: when I navigate the directory tree of my Dagger module, I'm looking for Dagger submodules first and foremost. I find that `std/` is a distraction from that.

> I'd prefer to keep these in separate places. Maybe we could have a top-level `std` and a top-level `dev`? That also then starts to make this really part of an "official" standard library I guess.
> 
> Another alternative is to do more refactoring to get rid of the go subpackages, but that is definitely trickier for `build` (and we might want to bring them back one day anyways).
> 
> Or we could just leave it as-is in `main` for now.

I would like us to try what I propose in this PR for a little bit, and see how it feels, if that's OK.

> Just a note for direction of follow-ups in this area - I'd like to make sure we're using our "std" modules, like dirdiff consistently. ATM, we're not good examples of how to do this, since we have mixed use of dirdiff _and_ `util.DiffDirectoryF`.

Yes, my other linting module removes `util` altogether, it's just bundled with a bigger change. Coming soon :)

dev/go/lint.json
shykes: Debug file caught in the commit... I'll remove

shykes:
> Pushed some commits to resolve the go package issues, were you able to run this locally @shykes? It wasn't just failing in CI for me, I was having this issue locally.

Yeah I was still getting errors... But ran out of time yesterday.

jedevc:
> I would like us to try what I propose in this PR for a little bit, and see how it feels, if that's OK.

Could we do the opposite then, and move the go code sub-packages all into the `dev/internal` package? I understand the logic for what you suggest, but I still think it's important to have some separation here.

shykes:
@jedevc I moved `consts`, `util` and `build` to `dev/internal/`. I didn't touch `mage` though, as I'm not sure how that is used, and I'm afraid to poke the YAML monster...

shykes:
@jedevc does this look good to you? Failed checks look like a flake.

jedevc (approved):
Good to merge - but make sure to keep an eye on main after merge, just to make sure we don't any unexpected failures there.

Also, would be good to announce in #engine-dev, so no one is surprised by the changes.

dev/README.md
jedevc, line 1: Oop, the `dagger call` commands in here need a `-m dev` now right?

shykes: I almost made that changed, but realized that it depends whether their workdir is `dev` (where that README lives) or the repo root. Sadly the same is true for the `--source` argument...

I decided to leave it alone for now, and focus on fixing those UX papercuts ASAP.

jedevc: Up to you, but I'd assume the working directory is probably the top-level of the repo (we can state the assumption at the top).

Not a blocker though!

shykes: OK makes sense, I'll make the change in a follow-up PR to get this through before the window closes :)

shykes:
@sipsma @grouville @matipan  I'm sorry to bother you, but it's night time for Justin, and I'm stuck on this failing engine test... I don't know if it's a flake or not. I'm worried that my window for merging this will close, and I won't have time to keep up with merge conflicts before going on vacation... Any help would be appreciated 🙏

PR #7767, opened 2024-06-27
core: add API for local engine cache and integ tests

sipsma:
Main commit:

This adds a new graphql API under Query called DaggerEngine, which is
intended to be used for querying and controlling engine-wide state, e.g.
cache, global configuration currently only in the toml config file, etc.

To start out here, there are just a few APIs for:
1. Querying which cache entries exist in the local cache
2. Triggering a manual prune of the entire cache (no filtering/selection
   yet)
3. Querying the current keepBytes config of the engine's automatic GC

The initial motivation here is to support integration testing of the
local cache pruning logic, which is badly needed after [a few recent bugs](https://github.com/dagger/dagger/pull/7738)
in that area.

Secondary motivation is to:
1. Setup support for CLI commands that query+prune local cache
2. Get the plumbing in place for fleshing this API out further with more
   cache and configuration control. The in progress work on support for
   [storing cache volumes on the host](https://github.com/dagger/dagger/pull/7646) in particular will be adding more
   to it.

The API exposed here is purposely pretty minimal to start, hiding
details that would be exposing low-level buildkit details (i.e. strings
used to filter cache entries) that we may not want to be attached to.

---
The updated graphql schema (see integ tests added for usage examples):
```graphql
"""The Dagger engine configuration and state"""
 type DaggerEngine {
   """A unique identifier for this DaggerEngine."""
   id: DaggerEngineID!

   """The local (on-disk) cache for the Dagger engine"""
   localCache: DaggerEngineCache!
 }

 """A cache storage for the Dagger engine"""
 type DaggerEngineCache {
   """The current set of entries in the cache"""
   entrySet: DaggerEngineCacheEntrySet!

   """A unique identifier for this DaggerEngineCache."""
   id: DaggerEngineCacheID!

   """
   The maximum bytes to keep in the cache without pruning, after which automatic pruning may kick in.
   """
   keepBytes: Int!

   """Prune the cache of releaseable entries"""
   prune: Void
 }

 """An individual cache entry in a cache entry set"""
 type DaggerEngineCacheEntry {
   """Whether the cache entry is actively being used."""
   activelyUsed: Boolean!

   """The time the cache entry was created, in Unix nanoseconds."""
   createdTimeUnixNano: Int!

   """The description of the cache entry."""
   description: String!

   """The disk space used by the cache entry."""
   diskSpaceBytes: Int!

   """A unique identifier for this DaggerEngineCacheEntry."""
   id: DaggerEngineCacheEntryID!

   """The most recent time the cache entry was used, in Unix nanoseconds."""
   mostRecentUseTimeUnixNano: Int!
 }

 """A set of cache entries returned by a query to a cache"""
 type DaggerEngineCacheEntrySet {
   """The total disk space used by the cache entries in this set."""
   diskSpaceBytes: Int!

   """The list of individual cache entries in the set"""
   entries: [DaggerEngineCacheEntry!]!

   """The number of cache entries in this set."""
   entryCount: Int!

   """A unique identifier for this DaggerEngineCacheEntrySet."""
   id: DaggerEngineCacheEntrySetID!
 }
```

---

Other commits are interrelated cleanups+fixes:
* [tests: cleanup nested engine network config](https://github.com/dagger/dagger/pull/7767/commits/addf193d3a224f779f53ee68b4b1d56da57fcc75)
* [engine: speed up session attachable disconnect](https://github.com/dagger/dagger/pull/7767/commits/ce01f52a0b782bcc56d43b6c6617066f587e0f82)


jedevc:
:thinking: Do we want this to be part of the core API? IMO, this would make more sense as part of a meta "debug" API. Specifically, these are not really consistent at all across different engines, and really designed as introspection entry points.

I guess for now, we don't really have this distinction - but I think we should definitely make sure we're not code-generating this, they definitely remain as internal, ideally only top-level clients like the CLI should see it.

sipsma:
> 🤔 Do we want this to be part of the core API? IMO, this would make more sense as part of a meta "debug" API. Specifically, these are not really consistent at all across different engines, and really designed as introspection entry points.
> 
> I guess for now, we don't really have this distinction - but I think we should definitely make sure we're not code-generating this, they definitely remain as internal, ideally only top-level clients like the CLI should see it.



> 🤔 Do we want this to be part of the core API? IMO, this would make more sense as part of a meta "debug" API. Specifically, these are not really consistent at all across different engines, and really designed as introspection entry points.
> 
> I guess for now, we don't really have this distinction - but I think we should definitely make sure we're not code-generating this, they definitely remain as internal, ideally only top-level clients like the CLI should see it.

@jedevc I will for sure hide it from module codegen for now and enforce the caller is the main client.

Generally speaking though, what I feel strongly about is that this is a graphql API (these cache querying APIs actually fit into the "standard gql" model really well) and at that point I can't see why it shouldn't just be grouped in with our other core APIs.
* One thing to clarify is that while these are mostly just wrappers around the worker's disk usage+prune APIs right now, I'm envisioning this growing much more in the short-to-long term and becoming more and more specialized to our use case
* e.g. I want to enable pruning of cache volumes stored on the host via this API (which diverges wildly from how existing cache volumes are stored and thus can't use the existing buildkit pruning code). That implies a connection point between this API and the existing core `CacheVolume` API, so splitting this out entirely may make less and less sense
* in the longer term, I could even imagine doing something like connecting cache entries in this new API with the ID that generated them, so you can map back and forth between all our various entities in the core API and all these cache entries, which may enable all sorts of interesting things

The only small question mark left for me is the API call that enables pruning, which feels *mildly* privileged. It's not a security concern, so given how useful a manual prune API call is I'm currently going forward with it. But ideally in the long term we may want to add some sort of auth so only certain clients can make these sorts of calls (big project though of course).

jedevc:
Yes, think we're on the same page - this should definitely be a graphql api, my only concern is about modules accessing it.

Which... maybe there are legitimate use cases for? But feels like we need to have a bit more of a think about that, it feels potentially "here-be-dragons", so maybe we should see how the API involves over time before exposing it into modules (since deprecating/removing those is hard :scream:)

sipsma:
> Yes, think we're on the same page - this should definitely be a graphql api, my only concern is about modules accessing it.
> 
> Which... maybe there are legitimate use cases for? But feels like we need to have a bit more of a think about that, it feels potentially "here-be-dragons", so maybe we should see how the API involves over time before exposing it into modules (since deprecating/removing those is hard 😱)

Agreed that there could be interesting things here with modules like custom cache control via module... but that would most definitely call for some form a trust model or another, so now is not the right time to open that door.

sipsma:
Been working on the integ tests here and finding all sorts of fun pre-existing issues all over the place:
1. Various problems with SDKs/dagql API correctness
   * https://github.com/dagger/dagger/issues/7788
   * https://github.com/dagger/dagger/issues/7789
3. An issue where sometimes session attachables wait an extra 30s before shutting down due to grpc healthchecks not failing until then. Not the end of the world but keeps state around longer than needed and results in my tests currently needing a dumb `time.Sleep`
   * This one seems straightforward to fix so will likely spin out to separate PR rather than opening issue

However, the automatic GC is indeed working based on the test I added, so at least there's that.

sipsma:
Hm just saw that some of the SDKs are not playing nice with the updated code that hides the EngineGCConfig type so it's internal only...

I doubt it's really the SDKs fault since that hiding is probably slightly outside the bounds of the gql spec in retrospect (there's an API that exists but is hidden from introspection).

I guess for now I'll just simplify further to get rid of that. We won't be able to have integ test coverage of fully custom GC configs but I can at least leave in an API for querying the `keepBytes` setting in isolation, which avoids these problems since it can just return an int rather than a full custom type.

sipsma:
The changelog check is failing with 
```
git diff --name-only origin/main...origin/refs/pull/7767/merge -- .changes/unreleased sdk/elixir/.changes/unreleased sdk/go/.changes/unreleased sdk/php/.changes/unreleased sdk/python/.changes/unreleased sdk/rust/.changes/unreleased sdk/typescript/.changes/unreleased
fatal: bad revision 'origin/main...origin/refs/pull/7767/merge'
```

Which seems like a bug in the check itself. Seeing if it's a quick fix.

internal/testutil/engine.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+var (
+	nestedEngineCount   uint8 = 1
+	nestedEngineCountMu sync.Mutex
+)
+
+// returns a device name and cidr to use; enables us to have unique devices+ip ranges for nested
+// engine services to prevent conflicts
+func GetUniqueNestedEngineNetwork() (deviceName string, cidr string) {
```
jedevc, line 15: :tada:

255 tests ought to be enough for everyone :slightly_smiling_face:

jedevc (commented):
I think this looks good? There's still a few suspicious looking TODOs though, like the policy ones?

Also, the note above about including these in SDK codegen.

internal/testutil/engine.go
sipsma: Haha yeah I considered doing some bit fiddling to make it a uint16 and use the other part of the IP range, but figured this should last us for a good while and will be happy to know once we've run 255+ tests with nested engines 😄

sipsma:
> I think this looks good 🎉 There's still a few suspicious looking TODOs though, like the policy ones?

Ah yeah I did miss some TODOs for comments on the new methods in engine/gc.go, which I fixed now.

> Also, the note above about including these in SDK codegen.

They are hidden from the module SDK codegen (there's an integ test too), but I'm purposely leaving them in for non-module SDKs. We are on the same page there right? I feel comfortable with that, just wanted to hide it from modules for now (there's also checks that only the main client caller can execute these APIs, so no holes with raw gql queries either there).

sipsma:
The engine tests passed: https://dagger.cloud/dagger/traces/d4fb2d8ccb62cbfc7daf58529fe5352f

But the GHA job failed: https://github.com/dagger/dagger/actions/runs/9768812699/job/26967051776?pr=7767 (canceled after 24m)

Think it's another instance of https://github.com/dagger/dagger/issues/7786

jedevc:
> They are hidden from the module SDK codegen (there's an integ test too), but I'm purposely leaving them in for non-module SDKs. We are on the same page there right?

I am still a *little* bit worried about external SDKs using this API - it feels a bit out of scope for CI code, and it might lead to mixing CI + engine management code. I'm also not sure about users being able to develop their own engine management tools, that hook into the engine in this way.

That said - modules are really the future of where we're going, so I'm not actually too worried. Aside from the small concerns above, yeah, as long as it's hidden from modules, all good :tada:

jedevc (approved):

PR #7771, opened 2024-06-27
ci: use nested dagger for testdev (take 2)

jedevc:
Attempt to revive #7223, since we've had a lot of perf improvements (especially around otel :tada:)

Hopefully this workflow should run a bit faster now.


jedevc:
Hm, so - it looks like this is the same order of magnitude of speed as on `main`, and we remove the external `docker` dependency, and tidy up our custom `call` action.

@sipsma this should be a fairly reasonable first step to tidying up our `test` vs `testdev` worflows as we discussed I think? `./hack/dev` is now entirely removed from our CI, and is only for local use.

.github/actions/call/action.yml
```diff
@@ -91,26 +54,43 @@ runs:
...
+          if [[ -f $HOME/.docker/config.json ]]; then
+            # don't evaluate env vars in internal command...
+            flags="$flags "'--docker-cfg=file:${HOME}/.docker/config.json'
+            # ...but do for the external wrapper
+            flagsdev="$flagsdev --docker-cfg=file:${HOME}/.docker/config.json"
+          fi
+        fi
+
+        cmd="dagger -m ${{ inputs.module }} call ${flags} ${{ inputs.function }}"
+        dagger -m dev call ${flagsdev} dev --target=".:default" --cmd "${cmd}" directory --path=. -o .
```
sipsma, line 94: Double-checking this is basically saying "start dev engine as service, run our `dev` module from a container with that service as a dep". I *think* it is but it's inherently a bit mild-melting (not new, using dagger to build+test dagger always has been).

That sounds perfect if so, IIRC the utility of testdev was partially just to ensure our CI still worked on dev engines as opposed to just stable engines, which is what this would result in.

jedevc: Yes :tada: If you have a better way to write this as code, then please shout loudly, it is mind melting as you say :cry:

gerhard (approved):
This looks really good! Doing away with the Docker dependency is such a huge improvement. There is **so** much infra code that we can remove after this gets merged cc @matipan 

This cannot get merged soon enough 🚁

gerhard:
We were talking about this with @matipan in the context of:
- https://github.com/dagger/dagger.io/pull/3760

How close are we to merging this @jedevc ? If we're not far off, we can simplify that PR today.

jedevc:
Rebased and resolved conflicts, *should* be good to go after that.

jedevc:
Hm, yeah, something is messed up, failures are legitimate.

Issue #7763, opened 2024-06-26
🐞 Jenkins does not work with Dagger Cloud following integration instructions


levlaz:
### What is the issue?

If you set up a Jenkins pipeline following the [instructions in our docs](https://docs.dagger.io/integrations/jenkins) then the trace does not show up in the Dagger Cloud UI home page. 

### Dagger version

dagger v0.11.9 (registry.dagger.io/engine) linux/amd64

levlaz:
This works now, thanks @crjm

PR #7776, opened 2024-06-27
Update Jenkins Instructions to work Cloud

levlaz:
This commit updates the Jenkins instructions to work with Dagger Cloud. It fixes the issues reported in #7763

* refactor Jenkinsfile to not require a go project for it to work, this uses a simpler example so that people can get this working right away without needing to make any changes.
* add instructions to show how to properly checkout git branches for workign with Dagger Cloud on different types of Jenkins projects
* add new snippet showing "the long way" to get this to work by adding a big blob of config directly into your Jenkinsfile
* remove part of doc that said that we assume you have dagger CLI installed because the Jenkinsfile installs the Dagger CLI. I also updated the Dagger CLI install script to put it in bin and call dagger with bin/dagger because I was having trouble updating the PATH in my own Jenins install. I think this approach will work more universally for people - but ideally they would pre-install the CLI


nipuna-perera:
Hey @levlaz to get the PATH working you can specify it also in the environment block

```
environment {
    DAGGER_VERSION = "0.11.9"
    PATH = "/tmp/dagger/bin:$PATH"
}
```

Then you have dagger CLI available in your PATH for the entire pipeline. To install the CLI I do something like this. I define a function outside of the `pipeline` def which then I call.

```
def installDaggerCli() {
    sh("""
        mkdir -p /tmp/dagger
        cd /tmp/dagger
        curl -L https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=$DAGGER_VERSION sh
        """)
}
```

You can also just point it to `BIN_DIR` instead of `cd` like you've done in the PR.

Finally, please test this. I don't think you need to do a implicit checkout to checkout the branch. Have you tried the following in your pipeline? The implicit checkout should still work here

```
git fetch -a
git checkout {branch}
```

crjm:
>  I don't think you need to do a explicit clone/checkout to checkout the branch.

Hey @nipuna-perera!
We've found that the default process Jenkins uses to checkout a specific point in history is not compatible with the way we currently collect metadata to display relevant information in Dagger Cloud.

If you are not using Dagger Cloud, this extra configuration can be omitted, as it's is mostly for the engine to be able to collect all the necessary Git metadata to correctly.

This is not a fix, but a patch, as we are looking into other ways of improving metadata collection, mostly to avoid unnecessary config like this to be added to runner configuration.

**[This is an explanation on how the Jenkins Git plugin default setting work apparently](https://stackoverflow.com/a/72752979)**

Of course, if you are aware of a simpler way to fetch this information without adding extra config to the `Jenkins` file, it would be awesome, as I personally don't know much of Jenkins.

In case you are interested, [we also tried this other approach, which also worked](https://www.youtube.com/watch?v=ai1kf4ihZUo).

nipuna-perera:
> This is not a fix, but a patch, as we are looking into other ways of improving metadata collection, mostly to avoid unnecessary config like this to be added to runner configurations/provisioning.

Sounds good! Unfortunately with Jenkins single branch pipelines the options are a bit limited. Have you tried multibranch pipelines? They do a branch based checkout as the jobs are by nature branch based.

docs/current_docs/integrations/snippets/Jenkinsfile
```diff
@@ -7,14 +7,12 @@ pipeline {
...
-  // assumes a Go project
-  // modify to use different function(s) as needed
   stages {
     stage("dagger") {
       steps {
         sh '''
-            curl -L https://dl.dagger.io/dagger/install.sh | BIN_DIR=$HOME/.local/bin sh
-            dagger -m github.com/kpenfound/dagger-modules/golang@v0.1.5 call build --project=. --args=.
+            curl -L https://dl.dagger.io/dagger/install.sh | BIN_DIR=bin sh
+            bin/dagger call -m github.com/shykes/daggerverse/hello@v0.1.2 hello --greeting "bonjour" --name "from jenkins"
```
shykes, line 15:
```suggestion
            bin/dagger call -m github.com/shykes/hello hello --greeting "bonjour" --name "from jenkins"
```

This will be slighter faster

levlaz:
Thanks so much for the tip about `PATH`  and the overall feedback 

> Sounds good! Unfortunately with Jenkins single branch pipelines the options are a bit limited. Have you tried multibranch pipelines? They do a branch based checkout as the jobs are by nature branch based.

Yes, it seems that by default Jenkins does a checkout in a detached HEAD state. The instructions cover both multibranch pipelines and regular pipelines using a simple setting in the Jekins UI.

I'm going to update the instructions to make it more clear that the verbose in-line checkout is only necessary if the UI-based options don't work for some reason.

levlaz:
Hey Marcos,

I don't think there is any way for us to solve this because how can we group by branch if we don't have any way to know what the branch is?

I don't think this PR is too bad, especially because you have to remember that the huge blob of config is not strictly necessary. It boils down to this:

* Dagger Cloud Needs Branch Info

* Here's an easy way to do it in Jenkins (add a single setting in the Jenkins UI)

* Here's a harder way to do it in Jenkins (add this big blob of config to your pipeline)

We're likely going to need to configure many integrations in a similar way so I don't see why this is any different.

Lev Lazinskiy
Head of Solutions @ dagger.io ( http://dagger.io )
e: ***@***.***
p: 415.470.2142

levlaz:
Hey @crjm - I think this is ready to ship. I shipped an updated patch that includes the enhancements we discussed, can you let me know what you think?

crjm:
> Hey @crjm - I think this is ready to ship. I shipped an updated patch that includes the enhancements we discussed, can you let me know what you think?

IMO, this is a good temporary workaround while we discuss a long term fix for all providers in #7802, but want to make sure we are all aligned before merging cc @marcosnils

marcosnils:
> I don't think there is any way for us to solve this because how can we group by branch if we don't have any way to know what the branch is?

@levlaz I spoke with @crjm about this yesterday and one of the ideas that came up was to check if the `git` jenkins plugin was effectively setting this variables (https://plugins.jenkins.io/git/#plugin-content-environment-variables) with the values we expect using the default configuration. If this is the case, then I stand on my original comment about making the changes in the engine so it picks up and uses those variables automatically instead of adding this to our docs and indicate users to change their pipelines configuration.

levlaz:
On Tue, Jul 02, 2024 at 08:08:57PM -0700, Marcos Nils wrote:
> > I don't think there is any way for us to solve this because how can we group by branch if we don't have any way to know what the branch is?
> 
> @levlaz I spoke with @crjm about this yesterday and one of the ideas that came up was to check if the `git` jenkins plugin was effectively setting this variables (https://plugins.jenkins.io/git/#plugin-content-environment-variables) with the values we expect using the default configuration. If this is the case, then I stand on my original comment about making the changes in the engine so it picks up and uses those variables automatically instead of adding this to our docs and indicate users to change their pipelines configuration.

Thanks for the additional context and the link, I agree it would be
great to get this information automatically. I'm glad to hear its
possible. 

What is the best way to test if the engine can see these variables? I
have a jenkins instance up and running if you need access to it.

crjm:
> What is the best way to test if the engine can see these variables? I
have a jenkins instance up and running if you need access to it.

I'll create a PR that collects those variables (if they are available) and sends the proper labels to Cloud.

This is why #7802 is important. We currently don't know which providers are not supported out of the box in Dagger Cloud. In this specific case we realized it was Jenkins, but there could be others that might be other contexts sending partial git telemetry, and it might be hard for a user to realize that and provide feedback on it.

levlaz:
Thanks Julian, 

> I'll create a PR that collects those variables (if they are available) and sends the proper labels to Cloud.

Is there anything I can do to help you with this?

> This is why #7802 is important. We currently don't know which providers are not supported out of the box in Dagger Cloud. In this specific case we realized it was Jenkins, but there could be others that might be other contexts sending partial git telemetry, and it might be hard for a user to realize that and provide feedback on it.

This is a great point, I am glad we're thinking about this more
defensively and not just in the context of Jenkins.

crjm:
> Is there anything I can do to help you with this?

Thanks for following up!
Once #7835 is approved/merged, we'll have Jenkins properly integrated in Cloud.
As for the docs, we could be a bit more specific, based on what I came across while trying out Jenkins.

---
Apparently, there are quite a few ways to create projects, being [pipelines](https://www.jenkins.io/doc/book/pipeline/) the latest/most recommended way. 
Thing is, when using pipelines, you **must** create a `Jenkinsfile`, and that can either be by pulling it from a repo, or by writing it in the Jenkins config tab.
When pulling it from a repo, Jenkins creates a local branch and fetches the branch, but never clones the repo, thus it ends in a weird Git context.

I would guess users tend to pull the Jenkins file from a standalone repo OR the `Jenkinsfile` is stored alongside the source code. 
In the first case, this fix won't really do much, because users will definitely checkout the source code separately after checking out the Jenkinsfile. In fact, [using the checkout step is the recommended way to use Git per the Jenkins docs](https://www.jenkins.io/doc/pipeline/steps/git/#git-git).
In the second case, this fix comes in: when no checkout is done, but the `Jenkinsfile` was pulled from a repo.
There, we parse the Git-related env variables set in the context of the Jenkins runner to complement the missing Git info.

levlaz:
On Fri, Jul 05, 2024 at 01:26:49PM -0700, Julián Cruciani wrote:
> > Is there anything I can do to help you with this?
> 
> Thanks for following up!
> Once #7835 is approved/merged, we'll have Jenkins properly integrated in Cloud.
> As for the docs, we could be a bit more specific, based on what I came across while trying out Jenkins.
> 
> ---
> Apparently, there are quite a few ways to create projects, being [pipelines](https://www.jenkins.io/doc/book/pipeline/) the latest/most recommended way. 
> Thing is, when using pipelines, you **must** create a `Jenkinsfile`, and that can either be by pulling it from a repo, or by writing it in the Jenkins config tab.
> When pulling it from a repo, Jenkins creates a local branch and fetches the branch, but never clones the repo, thus it ends in a weird Git context.

Hm, I'm seeing something different. Check out this build - it does not have
any additinal checkout step in the pipeline, but it pulls the Jenkinsfile from a repo and and it clones the repo as I expect it to.

https://jenkins.levlaz.org/job/jenkins-multi-demo/job/main/

https://github.com/levlaz/jenkins-demo/blob/main/Jenkinsfile

This project is a multi-branch pipeline, it has Build Configuration set
to "by Jenkinsfile" pointing to the Jenkinsfile in the repo, and it has
the additional setting to "Check out to matching local branch"
configured in the UI. 

https://jenkins.levlaz.org/job/jenkins-multi-demo/configure

In either case, its still not clear to me what the best way to test this
out is, how could I run a version of cloud and/or engine in my jenkins
install to see what the behavior looks like for the various forms of
config?

crjm:
> Check out to matching local branch

This is why it will have the necessary metadata, even if you are not explicitly setting the checkout step.
I do see the checkout step being printed to the console because of that setting, even though the step is not set in the `Jenkinsfile` itself: 
```
[Pipeline] stage
[Pipeline] { (Declarative: Checkout SCM)
[Pipeline] checkout
``` 

If, for example, you were to write the `Jenkinsfile` directly in the UI _without_ pulling it, no Git metadata would be present, and it would be shown as an orphaned trace in Cloud.

This is also why I would **encourage** users to be sure the git context is set, instead of running around possible variables and such from the engine. There are just too many ways to configure Jenkins, and too many ways that configuration resolves SCM operations.

crjm:
>  How could I run a version of cloud and/or engine in my jenkins install to see what the behavior looks like for the various forms of config?

I built a dev CLI [following this](https://github.com/dagger/dagger/tree/main/dev#build-the-cli)
Then I [made that CLI available to Jenkins by placing it in the userContent directory on my local Jenkins.](https://www.jenkins.io/doc/book/managing/user-content/)
And then I downloaded it through curl inside the `Jenkinsfile`:
```groovy
  stages {
    stage("dagger") {
      steps {
        sh 'printenv'
        sh 'curl -v -u admin:f095ce071d12486d92762ec2a156a90c http://localhost:8080/userContent/dagger --output dagger'
        sh 'chmod +x dagger'
        sh './dagger version'
      }
    }
  }
```

jpadams:
https://plugins.jenkins.io/git/
Super late to this party, but indeed, I think most people probably install the `git` plugin:
image

levlaz:
On Fri, Jul 05, 2024 at 02:13:55PM -0700, Jeremy Adams wrote:
> https://plugins.jenkins.io/git/
> Super late to this party, but indeed, I think most people probably install the `git` plugin:

I think that is a fair assessment, I installed Jenkins with the
recommended defaults which included 8 different git related plugins
including git.

levlaz:
> > Check out to matching local branch
> 
> This is why it will have the necessary metadata, even if you are not explicitly setting the checkout step.
> I do see the checkout step being printed to the console because of that setting, even though the step is not set in the `Jenkinsfile` itself: 
> ```
> [Pipeline] stage
> [Pipeline] { (Declarative: Checkout SCM)
> [Pipeline] checkout
> ``` 
> 
> If, for example, you were to write the `Jenkinsfile` directly in the UI _without_ pulling it, no Git metadata would be present, and it would be shown as an orphaned trace in Cloud.
> 
> This is also why I would **encourage** users to be sure the git context is set, instead of running around possible variables and such from the engine. There are just too many ways to configure Jenkins, and too many ways that configuration resolves SCM operations.

Thank you, make sense. 

I'm back to thinking we should just merge this docs PR as written, if
Cloud + Jenkins works for people using the engine vars then great. If
not, the steps described here should be enough to get most installations
working.

crjm:
> If not, the steps described here should be enough to get most installations working.

Agree with this, but, it's something that happens for all providers (GitHub, GitLab, etc): if you don't set a git context, you won't get CI groups in Dagger Cloud. We would need to update all docs of all providers in that case, or add it as a general warning in the Cloud docs, instead of only the Jenkins example.
Given that writing a pipeline directly in the Jenkins editor (no SCM config set) with no checkout step is an edge-case, I don't think it's worth it.

crjm:
Wait, just found something. Will come back with a proper update after triple-checking something.

levlaz:
On Mon, Jul 08, 2024 at 08:15:27AM -0700, Julián Cruciani wrote:
> > If not, the steps described here should be enough to get most installations working.
> 
> Agree with this, but, it's something that happens for all providers (GitHub, GitLab, etc): if you don't set a git context, you won't get CI groups in Dagger Cloud. We would need to update all docs of all providers in that case, or add it as a general warning in the Cloud docs, instead of only the Jenkins example.
> Given that writing a pipeline directly in the Jenkins editor (no SCM config set) with no checkout step is an edge-case, I don't think it's worth it.

Yeah you are right, I was working with GitLab the other day and I
noticed that they have a similar config option in their UI that allows
you to configure Git checkouts in various ways. It seems like if people
happen to do the checkout in a detached HEAD state then we would run
into this same issue on any provider. 

So it seems that the real answer is to either instruct people to
configure their CI provider in a certain way, or rethink how we approach
this problem to handle any possible scenario.

crjm:
> So it seems that the real answer is to either instruct people to configure their CI provider in a certain way, or rethink how we approach this problem to handle any possible scenario.

Yes, handling any possible scenario is the goal of solving #7802 

In this case, in order to make telemetry work properly in Jenkins, users will have to define the branch they want to build in the [`checkout()` step](https://plugins.jenkins.io/git/#plugin-content-checkout-with-defaults) _**only if**_ they have their `Jenkinsfile` and the source code in different repositories. 
But this is basically a must, as Jenkins will fail if no branch is specified in the `checkout()` step (it won't know what to build).

levlaz:
I'm going to close this PR for now and reopen it in the future if the fixes pending in the other PRs still make it tough to use Jeknins with cloud

Thanks for all the help @crjm

PR #7778, opened 2024-06-28
fix: ctr / dir terminal race condition

grouville:
Experiment to try fixing race condition that seems to have been introduced by https://github.com/dagger/dagger/commit/82b67892fa47038e85f7bf226e25dfedce0aa320.

cf. corresponding trace: https://dagger.cloud/dagger/traces/c71cecd3d42758687c7437e4b7ec98d1?span=1302c9920a784c73

Not the best implementation but fixes the go races


grouville:
SDK errors are unrelated, local runs do work on my side

grouville:
Race got triggered again, possible that this is not the root cause

grouville:
closing as this is probably unrelated

Issue #5668, opened 2023-08-18
Standardize on a single base OS for engine release


shykes:
## Problem

We currently maintain 3 different flavors of the Dagger Engine OCI image, based on 3 different base OS:

1. The default image, based on Alpine Linux
2. An alternative image based on Ubuntu, which is required for experimental GPU support (Alpine doesn't have the required nvidia tooling)
3. A new Wolfi-based image, with the potential of combining the best of the other two

Having several flavors of the engine image is fine as a temporary stopgap, but it has been several months, and the stopgap is starting to look temporary. This is a problem.

## Solution

Get to a single unified image as soon as possible. Let's pick a winner, make sure it works great, and discontinue the others. If we need to do this in a cautious staged way, that's fine: but let's please get it done. @gerhard I already know you agree - can you own this?

gerhard:
Makes sense!

In terms of naming, are you thinking `registry.dagger.io/engine-ubuntu`? I can imagine `registry.dagger.io/engine-gpu` too.

shykes:
FYI @jedevc @dlorenc @amouat FYI this issue tracks the (unfinished) work of standardizing engine images on a single distro (currently GPU uses Ubuntu, and non-GPU uses Alpine).

I know youy started discussing in #6887, just wanted this issue to be on your radar.

gerhard:
`doesn't close #5668` closed it 🤦‍♂️

Issue #7760, opened 2024-06-26
Introduce go-style pseudoversions for cli + engine releases


jedevc:
Currently, all releases are tagged using semantic versioning - e.g. `v0.11.9`, the latest release. This tag is builtin to the resulting CLI and Engine binaries, and is used (among other things) for determining version compatibility (see https://github.com/dagger/dagger/issues/7640 for more information).

However, all builds cut from `main` are instead built using a commit SHA version, and all builds made in a local dev environment are built using a `dev-<id>` version.

This is workable, but isn't great. The issue is that because these other cases aren't semver, we can't easily compare them. Currently, we just assume that any semver version is the newest, and should be excluded from various version compat checks. Users installing these versions though (including core devs for testing, or users living on the bleeding edge) have a bad experience, where we would still want to perform version compat checking, or serve the right collection of APIs.

To do this, we should *always* generate a semver tag, in similar style to go.
- If the last release was `v0.11.9`, a main build should have a tag of the form `v0.11.10-<timestamp>-<commit>`
- If the last release was `v0.11.9`, a dev build should have a tag of the form `v0.11.10-<timestamp>-dev-<id>`

With this, it's now possible to compare versions - both of these above are prereleases, so will be correctly sorted to be *later* than v0.11.9, but before v0.11.10. The timestamps help to ensure some level of sorting when building off of `main` - this isn't perfect, but it's probably about the best we can do.

sipsma:
Agree we should do this, only nit is it might be nice to stay in semver spec, which I think allows for all of this via `+` metadata tags: https://semver.org/#semantic-versioning-specification-semver

Only slight advantage being that we can continue using existing semver libs/tools to parse it. Not a big deal either way though.

jedevc:
Ahh, so I tried this - the issue is that by using build metadata, "vX+<meta>" counts as a release of "vX" (with metadata). I don't think we want that.

Instead, we're doing go's trick where we use the pre-release - since *technically*, this is a prerelase.

PR #7781, opened 2024-06-28
ci: Publish main Engine image when publishing from the main branch

gerhard:
> [!WARNING]  
> Builds on top of https://github.com/dagger/dagger/pull/7843. Rebase onto `main` & merge **after** that PR is merged.

We want this so that we can reference `registry.dagger.io/engine:main` in our CI, and have our runners pick up the latest Engine Image (a.k.a. dev). I didn't want to call this tag `dev` since:
- it assumes there is a single dev branch (unlikely to be true for long based on the latest releasing discussions)
- it requires us to make an extra mental jump from dev to main to understand it's true origin
- it required me to change existing code (we are already using the dev concept when building from local)

We now provide all tags explicitly, as a comma-separated string, to the publish function. This keeps things simple. This new --tags flag is not optional so that we don't forget to provide it.

Related to:
- #5668 

---

image


ci/build/engine_image_ref.go
```diff
@@ -0,0 +1,28 @@
+package build
+
+type EngineImageRef struct {
+	Wolfi   bool
+	GPU     bool
+	Image   string
+	Version string
+}
+
+func (e *EngineImageRef) String(explicitVersion string) string {
```
jedevc, line 10: `String` is usually indicated to return the string representation of a struct.

Not sure I understand why we have an `explicitVersion` here, it's quite confusing to follow.

ci/engine.go
```diff
@@ -201,25 +201,23 @@ func (e *Engine) Publish(
...
 		platform = []Platform{Platform(platforms.DefaultString())}
 	}
 
-	ref := fmt.Sprintf("%s:%s", image, e.Dagger.Version)
-	if strings.Contains(e.ImageBase, "wolfi") {
-		ref += "-wolfi"
-	}
-
-	if e.GPUSupport {
-		ref += "-gpu"
```
jedevc, line 215: Could we potentially simplify this by moving the logic of what tag names to use into github actions? And then just pass a `tags` parameter that we push to all of.

jedevc (approved):

ci/build/engine_image_ref.go
gerhard: Yes, `EngineImageRef` is a struct and this is the string representation for it.

When `explicitVersion` is set, it overrides struct's `Version` property.

ci/engine.go
gerhard: I really like this idea! Let me play with it and see how I can simplify the current approach.

gerhard:
Re-requesting reviews since I changed the initial implementation based on review feedback.

gerhard:
Rebased on top of `main` to load the changes from:
- https://github.com/dagger/dagger/pull/7766

sipsma (changes requested):
> It no longer tries to compute a digest from the directory contents, version must be provided explicitly. This makes it easier to use it in cases where a directory is not needed.

I don't understand why this is being removed, can't we just move the place where that digest is calculated so it's provided as an arg to `newVersion`? Or even just do a check if the provided dir is nil?

That logic is extremely useful because it results in local dev engines having cache that is version dependent getting invalidated after you rebuild. Before it we constantly needed to manually prune the local cache of our dev engines in order for tests to work as expected.

gerhard:
> That logic is extremely useful because it results in local dev engines having cache that is version dependent getting invalidated after you rebuild. Before it we constantly needed to manually prune the local cache of our dev engines in order for tests to work as expected.

I was missing this context. It makes sense to keep it. Making the change now.

gerhard:
The change is now back @sipsma. How does this look now?

sipsma (approved):
> The change is now back @sipsma. How does this look now?

LGTM, thank you!

dev/main.go
```diff
@@ -25,16 +27,22 @@ func New(
...
 	version string,
 	// +optional
+	gitBranch string,
+	// +optional
+	tagSuffix string,
+	// +optional
 	dockerCfg *Secret,
 ) (*DaggerDev, error) {
-	versionInfo, err := newVersion(ctx, source, version)
+	versionInfo, err := newVersion(ctx, source, version, tagSuffix)
```
jedevc, line 36: The tag suffix shouldn't go in here - I think it only makes sense to pass it to `Publish`.

---

Note that by passing it into `DaggerDev.Version`, this means that binaries built inside this engine, like the dagger binary will have `-gpu` in the version name, so we'll get versions of the form `v0.12.0-gpu` which will be evaluated as a prerelease of `v0.12.0`, which isn't the right intention.

We shouldn't build the tag suffix into the binaries, so it probably shouldn't go here. If we *absolute really* want to do this, we could put `+gpu` so it's a build variant instead (but I don't actually think it should go there at all).

jedevc (changes requested):

dev/main.go
gerhard: That's the behaviour that we have today. In the last release, we published:
- `registry.dagger.io/engine:v0.11.9`
- `registry.dagger.io/engine:v0.11.9-gpu`

While changing to `registry.dagger.io/engine:v0.12.0+gpu` makes sense, more changes are coming in the PR that builds on top of this: https://github.com/dagger/dagger/pull/7782

If all version info is collected here, why would we move the tag suffix into `Publish`?

Does this change build the tag suffix into the CLI binary? My understanding is that this only affects the Engine, not the CLI. If you are sure that this will affect the CLI binary, I will set everything up in my fork to test publishing from `main` before we merge this.

jedevc: This isn't quite the behavior we have today - the tags are the same, yes, but the contents of the published image is now different.

To see this:

```
❯ dagger -m dev call --source=.:default --git-branch=main --tag-suffix=bad --version=v0.12.0 dev terminal

dagger /mnt $ dagger version
dagger v0.12.0-bad (registry.dagger.io/engine) linux/amd64
dagger /mnt $ echo "{version}" | dagger query
{
    "version": "v0.12.0-bad"
}
```

The suffix *must not* appear in the version info here - there's some discussion in #7760 around populating these, but it's still a bit unclear. If we don't change this, then this build, despite being of v0.12.0 will look like a pre-release of v0.12.0 - which will break a lot of the work we're version compat (e.g. dagger v0.12.0 clients will entirely reject all connections to v0.12.0-gpu, since v0.12.0-gpu < v0.12.0 :scream:). Unfortunately, since we don't have gpu/wolfi variant provisioning tests, it's unlikely we would have caught this in tests once merged.

These versions are inserted at this point in the builder:

https://github.com/dagger/dagger/blob/2f9e3b39504b7a2798ca24ede94974d9b1e8e437/dev/internal/build/builder.go#L313-L315

The confusion is coming from the changes that now treat the `Version` struct for two separate purposes - determining all the refs/tags to tag with, and determining the version info in the builder, these are different, and need to be separate.

Splitting out the `taxSuffix` to just be on publish would hopefully make this issue much less likely to happen - implementing this comment here would completely resolve this https://github.com/dagger/dagger/pull/7781#discussion_r1658921778, since it's much more explicit.

gerhard: That was a good point. I started from scratch and reworked the implementation. WDYT?

jedevc (approved):

gerhard:
Now that https://github.com/dagger/dagger/pull/7843 got merged, rebasing on top of `main` so that this can get merged straight after.

PR #7782, opened 2024-06-28
Standardize Engine image on a single base OS

gerhard:
This change is meant to inform us of the following:
 - How does the CI behave after a few days and at least 5 runs using this new image?
 - What changes are required to make this work?
 - What are the unknowns?

 Some of the obstacles that we encountered so far:
 1. If we pin the Wolfi image using a SHA256, what guarantees do we have that this will remain available?
 2. How do we consume a specific version of Go which is based on the Wolfi base image? The only option that I could find is "Contact us" meaning reach out to the Chainguard team.

 Other note-worthy changes:
- The GPU image variant is based on Wolfi (standardizing is nice)
- We are still publishing the Ubuntu GPU variant as `*-ubuntu-gpu`
- The Alpine variant is now suffixed with `*-alpine`. We can get the behaviour before this change by using that image variant.
- We have minimized the glibc difference between Ubuntu & Debian by bumping to Debian 12 (Bookworm). More details in the comment.
- We are now pinning the Alpine & Golang images using SHA256 digests (tags can change, we don't want to leave this to chance)
- Bumped Dagger to v0.11.9 in `ci`

This adds on top of:
- https://github.com/dagger/dagger/pull/7781

It closes:
- https://github.com/dagger/dagger/issues/5668

---

**After** https://github.com/dagger/dagger/pull/7781 is merged, the second part of this change should start working, namely the CI change. My current though is to have this commit removed **after** we gather enough data about how this behaves. We don't want to merge this **unless** we are happy to always test against an unreleased version of the Engine. While I really like that idea, I'm not sure if this PR is the right one to introduce this change in. After all, this is still **an experiment**.


amouat:
> If we pin the Wolfi image using a SHA256, what guarantees do we have that this will remain available?

That should be fine -- we've never deleted old images.

PR #7803, opened 2024-07-01
GolangCI

shykes:
Spin out our Go Linter (golangci-lint) into a low-level module. The emphasis is on incremental change and small scope.


dev/golangci/dagger.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "golangci",
+  "sdk": "go",
+  "source": ".",
+  "engineVersion": "dev-bb9c713a37c73cd4176da27c1e0c4654682f4de7"
```
gerhard, line 5: Should this a release version?

dev/golangci/lint-config.yml
jedevc, line 1: Should we remove `.golangci.yml` then? I'm actually not convinced about this file being part of the module, since it reduces it's re-usability - anyone outside of dagger, will probably want to customize this.

dev/golangci/dagger.json
jedevc, line 5: These versions are now significant, the API this wants is v0.11.9.

dev/golangci/lint.go
```diff
@@ -0,0 +1,173 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"path"
+	"strings"
+)
+
+const (
+	lintImageRepo   = "docker.io/golangci/golangci-lint"
+	lintImageTag    = "v1.59-alpine"
```
jedevc, line 13: We should have a `WithVersion` to be able to override this.

dev/golangci/dagger.json
shykes: My bad, I will correct this.

- Question: how are dev versions interpret by compat mode?
- Observation: this makes it even more annoying that switching back and forth between dev and stable engines on the same machine causes the engine container to be wiped

dev/golangci/lint.go
shykes: I have some concerns about that (no guarantee that every version of the tool works with every version of the module + not sure anyone actually needs to control the version), but I can understand the frustration in case someone does need this urgently and there is no flag for it.

For now can we agree to make it a follow-up improvement, like config file customization?

gerhard (approved):
Apart from the Engine version, this looks good to me.

Tried it out locally & it behaves as expected.

Approving optimistically 👍

dev/golangci/lint-config.yml
jedevc: Sounds fine about the follow-up task - but we should still remove the top level golangci yaml file to avoid confusion of having two files, one of which isn't currently used.

dev/golangci/dagger.json
jedevc: There's some improvements to this in https://github.com/dagger/dagger/pull/7858

However, compat mode and dev versions are very fiddly. The current behavior (and in that pr) is that a dev version is *always* considered to consume the latest API and schema.

I'm actively thinking about how to get this to work in a reasonable way - it's very much a work in progress.

dev/golangci/lint-config.yml
shykes: Yes you're right! Sorry I misunderstood what you were saying.

shykes: done!

dev/golangci/dagger.json
shykes: > However, compat mode and dev versions are very fiddly. The current behavior (and in that pr) is that a dev version is always considered to consume the latest API and schema.

That seems reasonable! See [this discord thread](https://discord.com/channels/707636530424053791/1260743246917795840) about the need for some sort of flag to control compat mode explicitly.

shykes:
`test-publish` has failed, is it a flake?

shykes:
@gerhard @matipan @jedevc are you able to tell if my remaining check failures are flakes? I don't see anything to my change in there, but could be wrong (CI logs are hard to read)

shykes:
This worked locally:

```console
dagger call -m github.com/dagger/dagger/dev@pull/7803/merge --source https://github.com/dagger/dagger#pull/7803/merge engine test-publish
```

shykes:
I'm going to merge since I'm positive this module doesn't affect e2e tests

Issue #6747, opened 2024-02-27
Support `Socket` args from the CLI


sipsma:
Right now you can't invoke Functions that accept args of type `Socket` from the CLI. In theory, this should be as simple as supporting e.g. `dagger call fn --sock unix:///var/run/docker/docker.sock` or `dagger fn --sock tcp://localhost:1234`, etc.

However, we need to be very careful when implementing this to not accidentally create a backdoor where any (malicious) module could gain access to any host socket. I think this will require explicit tracking of which modules have access to a given socket based on being passed it as an arg (or being passed an arg like a `Container` that has a socket embedded somewhere in its DAG definition).
* *Also* need to make sure this is done in such a way that we don't invalidate cache with any random IDs

Related: https://github.com/dagger/dagger/issues/6726#issuecomment-1963586621
Also related (plugs previous hole that made it possible to access these sockets from a module): https://github.com/dagger/dagger/pull/6748

jedevc:
> I think this will require explicit tracking of which modules have access to a given socket based on being passed it as an arg

I *think* this sounds similar to what I was suggesting for secrets as well - https://github.com/dagger/dagger/pull/6601#issuecomment-1964629954. Like I mentioned there, I think the tricky part is needing to handle the case where you pass objects around that contain secret fields (in a potentially nested way).

But I do really like this as the way that we perform isolation between these sensitive types, `Socket`s, `Secret`s, potentially even `File`s/`Container`s (not sure if these can be accessed via raw graphql today).

Regardless, kinda neat to have both thought up the same way of performing better module isolation, it definitely has my vote.

sipsma:
Just to update, been slowly chipping away at the pre-reqs needed to support this (https://github.com/dagger/dagger/pull/6806 and https://github.com/dagger/dagger/pull/6836). There's one last PR needed to add support; it involves some internal refactoring that I need to coordinate with other refactorings going on now, but should make it to a release in the near future.

helderco:
@sipsma, can you give an update on the pre-reqs here?

sipsma:
@helderco Yeah sorry there's been a ton of work going on that I've been mentioning in other various PRs but I will summarize the status here.

Basically, in order to support this while also *not creating a backdoor that would allow any malicious module to access to any socket on the host*, we need fine grain isolation of session resources. But before I started this effort, the lack of isolation was a very deeply embedded assumption over almost the entire codebase.
[...]

PR #7804, opened 2024-07-02
Isolate sockets+secrets per-client + support socket args

sipsma:
***FINALLY*** fixes https://github.com/dagger/dagger/issues/6747

This builds on top of all the recent refactoring and:
1. Fills existing holes in secret isolation (unskips test that was previously skipped that covers this isolation)
2. Isolates sockets per-client
3. Adds support for passing sockets as args using `dagger call` (e.g. `dagger call --docker-sock /var/run/docker.sock`)

## High-level idea
These session resources (just unix-sockets+secrets for now) are now isolated per-client such that they can only access a given one if:
1. They explicitly created them
2. (module function) They were explicitly passed them as an arg (directly or transitively embedded in the DAG definition of a given arg)
3. (module function) They were explicitly returned them in a call (directly or transitively embedded in the DAG defintion of a return value)

## Implementation Details
Some extra details to help make reviewing easier and to help any future efforts to isolate resources like this beyond secrets+sockets (goal here was to find a re-usable pattern)
* In addition to the [SecretStore](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/core/secret.go#L57-L57), there's now a similar [SocketStore](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/core/socket.go#L66-L66).
   * Where-as previously the SecretStore was part of session-wide state, now each client in a session gets its own. Same for the newly added SocketStore. [Relevant code](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/engine/server/session.go#L111-L112).
* Secrets+Sockets in those stores are now keyed by their dagql ID digest
   * This property was important because it allows you to lookup a given resource in a store purely from walking a `call.ID`, without needing to necessarily actually load the ID first
   * Not needing to load the ID means we don't need to necessarily have the dagql server of the client that created the ID, which hugely simplifies dealing with cases where we are getting a cached result and thus client execution may have been skipped entirely. [Relevant code](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/engine/server/client_resources.go#L31-L31).
* The above means that it's important the dagql ID digest is a unique identifier for the resource
   * Previously for secrets we used the "accessor" (hmac that included the secret name + module source ID digest) for this (though it was itself the key to the store previously)
   * [We still use that same idea](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/core/client_resource.go#L12-L12), but now the accessor is just an arg to the gql fields that return [secrets](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/core/schema/secret.go#L47-L47) and [sockets](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/core/schema/host.go#L409-L409) and is otherwise not used.
     * The fact that we provide a dagql arg to a field just to influence it's resulting ID digest feels a little _meh_, but was the least worst option I hit so far and is actually conceptually comparable to how the `blob` API works (which is how we isolated local dir resources in the past)
     * I think there's some grand idea slowly emerging here around patterns for handling "content hashed" and/or side-effectful dagql objects where the digest is not purely from the recipe but also mixes in something from the value. But still too hazy+early to do much with at this exact moment. cc @vito, seems relevant to various conversations 😄
* Once we have all of the above, we can just consistently use the ID digest of the secret/socket to
   * [Find it in the IDs of input args](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/engine/server/session.go#L338-L338)
   * [Find it in the IDs of return values](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/core/modfunc.go#L287-L287)
   * Use as the LLB ID when we need to allow buildkit APIs to access it ([secrets](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/core/secret.go#L38-L38), [sockets](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/core/socket.go#L42-L42))

## Future work in this area
There are more resources that are worth isolating, but ended up involving far out-of-scope changes:
1. Host ***IP*** sockets (as opposed to unix sockets)
   *[ I actually did some groundwork for this](https://github.com/sipsma/dagger/blob/34eabec0c3c709a8980bb7cc956cfb13e59a71b2/core/schema/host.go#L342-L342) by having the `Host.service` follow the same sort of pattern as `Host.unixSocket` and include `Socket` types in it's args (so they would be hit when walking its ID)
   * However, I soon realized that this would all be purely theatrical since host services are proxied through a CNI netns and we don't have any sort of isolation in our current networking setup (you could just port scan essentially)
   * So I stopped the updates there, leaving in the initial groundwork since it was a minor cleanup anyways. 
   * I see two options for finishing this in the future:
     1. Add CNI network isolation so a given container can only talk to others it explicitly has a binding with
     2. Reimplement host service proxying to not create a single netns but instead nsenter into each container that has a binding and setup proxying sockets in those individually (though there's some conundrums around the Endpoint api with this)
2. Registry Auth sourced from the client's `.docker` auth config
   * This in particular has always been more "ambient" without any explicit API object backing it. So at the moment it doesn't fit into this model well.
   * One potentially nice way to address this would be to make this registry auth an explicit object using the new context arg approach, so then access is modeled more explicitly. That will be a breaking change though of course.

---

TODOs:
- [x] Doc a bit more thoroughly the overall idea here. Current description is a start but probably worth more details
- [x] If time, cleanup commit history so I can avoid a large-ish Squash+Merge
- [x] Address TODO comments in the code, little cleanups needed in various places
- [x] Add more test coverage
   - Update current socket arg test to not rely on docker being present on the host
   - Add coverage for sockets being embedded transitively in the DAG def of an arg
- [x] (maybe) split out the `engine: disentangle client+session state from Query` commit to separate PR, which is a very tedious/boilerplately/uninteresting update but bloats the size of this PR a lot
   - done here https://github.com/dagger/dagger/pull/7812


sipsma:
Was seeing weird inconsistent test failures in a few SDKs, turned out to be a pretty interesting case to think about + handle here.

The problem happened when a function call was cached, which is *currently* just some SDK building calls but will in the near future be expanding to any user function calls.

The issue was that the new logic for "add secrets+sockets when returned them by a function call" error'd out if the function call never actually ran, since in that case the called client never ran, never created a secret/socket store to find those resources in, etc.

The handling logic I went with was:
1. If there were no secrets/sockets in the returned IDs, don't even try to find the called client
2. If there were secrets/sockets *but* they are all ones we already know about (e.g. we passed them to the called client in args), skip them.
   * This in particular required updates to store secrets+sockets in stores keyed by their dagql ID digests, which is a sort of neat property to have anyways
3. If there were secrets/sockets in the returned IDs *and* the client ID wasn't found due to caching, error out clearly (error message says along the lines of "secrets/sockets returned in cached call")

When we do add generalized function caching, users will start hitting this error is existing cases where a module function creates a new secret and returns it (directly or in DAG), unless they mark their function as never cached.
* There's nothing we can do about this, that change is just breaking, but was interesting to see it come to a head already.

sipsma:
The TODO I left for handling interfaces here is proving to be a real conundrum due to some pre-existing awkward integration between dagql+interfaces.

Problem is:
1. We need to walk input args + output returns to find all the IDs for secrets/sockets/etc. so that we can give the client receiving args/receiving return vals access to those secrets/sockets/etc.
2. But, the `id` field on an interface is currently implemented in the exact same way as `id` for everything else, which is arguably not exactly right since it means we drop the id of the underlying object implementing the interface in some cases.
   * This is specifically a problem when you do something like [this](https://github.com/sipsma/dagger/blob/6d2bd214bcf6c722399ef820c025193eafd9da0e/core/integration/testdata/modules/go/ifaces/test/main.go#L79) where you provide an interface as an arg, the function does some stuff to it and then returns it.
3. I looked at changing `id` on an interface to return the ID of the underlying object. It seems relatively straightforward, but I don't think it will work entirely as is since dagql still needs to know when loading the ID it is supposed to be wrapped in the interface.

I'm gonna try something (what I mentioned in `3.` above but see if it's possible to return the ID of the underlying object *plus* a selection of the `as*` field that converts it back to the interface so dagql knows how to handle it).

But if that doesn't pan out, I'm gonna timebox this to today and otherwise just get this PR merged and tackle this problem in follow ups since it might require more intricate dagql changes. Would rather get this merged and handle the corner cases separately in that scenario.

cc @vito in case you have any ideas (if this is buried far too deep in the weeds and making no sense, no worries though 😄)

sipsma:
> The TODO I left for handling interfaces here is proving to be a real conundrum due to some pre-existing awkward integration between dagql+interfaces.

Okay after expending some more brain calories I think I found the shortest path to this problem. It involves some more type assertions to get the right dagql.Instance of the right class, but is no worse than the existing code. https://github.com/dagger/dagger/pull/7804/commits/ec15ad38313ac8e37551af31e11b9794d071a012

That being said, I did realize that the idea I outlined in the previous comment would be pretty beneficial for other reasons:
1. Right now calling a field on an interface goes "direct" to the ModuleObject, which means that dagql caches it entirely separately from the equivalent call on the ModuleObject with no interface wrapping. That would be fixed by tracking the underlying ID/Instance of the implementing object and doing a dagql.Select on it.
2. I *think* this might allow us to get rid of the old TODO comment [here](https://github.com/sipsma/dagger/blob/6d2bd214bcf6c722399ef820c025193eafd9da0e/core/interface.go#L32) about loading returned interface values being expensive (which I agree with) and as a side effect get rid of the kind of complicated/confusing `Query.IDDeps` method (which has been a minor thorn in supporting in some other efforts recently).

It's just way way out of scope here, so have to defer to another time for now.

sipsma:
This is good to review now, I'll come back to update the PR description with some more details on the machinations a bit more so it can be a reference for any similar "isolate session resource" efforts in the future.

If I have time I'll also see if I can cleanup the commit history to avoid a squash and merge since it ended up being somewhat large in terms of diff.

engine/server/session.go
```diff
@@ -322,7 +327,21 @@ func (srv *Server) initializeDaggerClient(
 	failureCleanups *buildkit.Cleanups,
 	opts *ClientInitOpts,
 ) error {
-	// initialize all the buildkit state for the client
+	// TODO: this is wrong right? Check with Alex
```
sipsma, line 330: (apologies for the numerous pings but) cc @vito, I noticed that we were still using this spanCtx in some places but I can't see that there's anything in the context to give a span. Maybe there is some setup somewhere that I'm missing?

TBH I haven't really done my homework here, I just left this TODO since there was too much other crap to do, so if you aren't immediately sure just let me know and I'll dig around more.

sipsma:
Hitting known flakes in CI (https://github.com/dagger/dagger/issues/7900), will save the re-rolls to before merge.

engine/server/session.go
vito: All good on the pings, sorry I haven't given this all due diligence yet!

From a quick skim on `main` it seems like `spanCtx` is never actually initialized. I always see `traceID=000000000000000000` in the Buildkit logs too. So, any change you make here is probably better than the current state.

I think this was purely in service of getting trace propagation to work for socket and registry auth session attachables, which might change fundamentally (for the better) with this PR, so it seems like the appropriate time to fix them up.

The best way to test that is to send telemetry to Honeycomb and see if the trace ends up with spans that have missing parents. (It'll put a warning up top.)

core/interface.go
```diff
@@ -18,6 +20,14 @@ type InterfaceType struct {
 
 	// the type def metadata, with namespacing already applied
 	typeDef *InterfaceTypeDef
+
+	// cache of loaded underlying implementation types, id digest -> loadedIfaceImpl
+	loadImplCache sync.Map
```
vito, line 25: Is this cache load-bearing or more of a speculative improvement?

core/schema/host.go
```diff
@@ -207,20 +211,43 @@ type hostSocketArgs struct {
...
-	clientMetadata, err := engine.ClientMetadataFromContext(ctx)
+func (s *hostSchema) socket(ctx context.Context, host *core.Host, args hostSocketArgs) (inst dagql.Instance[*core.Socket], err error) {
+	socketStore, err := host.Query.Sockets(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get client metadata: %w", err)
+		return inst, fmt.Errorf("failed to get socket store: %w", err)
 	}
-	mainClientCallerID, err := host.Query.MainClientCallerID(ctx)
+
+	accessor, err := core.GetClientResourceAccessor(ctx, host.Query, args.Path)
```
vito, line 220: Nonblocking: I guess it's possible for two different types of objects to have the same accessor, considering this is basically `currentModule.Source.ID:/foo` - but it's not likely for that to be an issue, since those would be interacting with different subsystems (e.g. socket LLB ID, secret LLB ID), and these keys are only relevant within their subsystem.

core/secret.go
```diff
@@ -2,46 +2,21 @@ package core
...
-}
-
-func NewSecretAccessor(name string, scope string) string {
-	// Use an HMAC, which allows us to keep the scope secret
-	// This also protects from length-extension attacks (where if we had
-	// access to secret FOO in scope X, we could derive access to FOOBAR).
-	h := hmac.New(digest.SHA256.Hash, []byte(scope))
-	dt := h.Sum([]byte(name))
-	return hex.EncodeToString(dt)
+	// The digest of the DagQL ID that created this secret, used as its identifier
```
vito, line 17: nit: "Created" to me reads as the call that set the secret, but this is the ID of `secret(name: ...)`, not `setSecret`, which to me 'creates' the secret.

Maybe this?

```suggestion
	// The digest of the DagQL ID that accessed this secret, used as its identifier
```

vito (approved):
LGTM! The tests are fun acrobatics for sure, not in a bad way. Only had questions/comments to make sure I'm following along.

---

(Pardon the massive comment, wish GitHub had top-level threading...)

> Problem is:
> 
> 1. We need to walk input args + output returns to find all the IDs for secrets/sockets/etc. so that we can give the client receiving args/receiving return vals access to those secrets/sockets/etc.
> 
> 2. But, the `id` field on an interface is currently implemented in the exact same way as `id` for everything else, which is arguably not exactly right since it means we drop the id of the underlying object implementing the interface in some cases.
>        
>    * This is specifically a problem when you do something like [this](https://github.com/sipsma/dagger/blob/6d2bd214bcf6c722399ef820c025193eafd9da0e/core/integration/testdata/modules/go/ifaces/test/main.go#L79) where you provide an interface as an arg, the function does some stuff to it and then returns it.
> 
> 3. I looked at changing `id` on an interface to return the ID of the underlying object. It seems relatively straightforward, but I don't think it will work entirely as is since dagql still needs to know when loading the ID it is supposed to be wrapped in the interface.

Making sure I understand: we want to dissect IDs returned by functions to find any embedded Secret/Socket IDs to grant this client access to. But the problem is that sometimes (e.g. with interfaces) the IDs act like symlinks/wrappers, and might not directly embed the socket/secret.

On the one hand, this feels like maybe overloading IDs a bit too much since now they're a significant data structure on their own rather than a lazily evaluated black-box object, but maybe the fact that we have flexibility in doing things like this is a blessing and not a curse. Not a blocker, just something to maybe keep an eye on.

Also - this problem feels pretty similar `PBDefinitions`, only that approach felt less invasive since it's just a Go interface implemented by a value at runtime, so the behavior is determined by the object and not its ID.

> I'm gonna try something (what I mentioned in `3.` above but see if it's possible to return the ID of the underlying object _plus_ a selection of the `as*` field that converts it back to the interface so dagql knows how to handle it).

If we want to stick to this approach, that sounds reasonable. :+1:  But it seems like this same problem could manifest in other scenarios too. :thinking: 

> Okay after expending some more brain calories I think I found the shortest path to this problem. It involves some more type assertions to get the right dagql.Instance of the right class, but is no worse than the existing code. [ec15ad3](https://github.com/dagger/dagger/commit/ec15ad38313ac8e37551af31e11b9794d071a012)
> 
> That being said, I did realize that the idea I outlined in the previous comment would be pretty beneficial for other reasons:
> 
>     1. Right now calling a field on an interface goes "direct" to the ModuleObject, which means that dagql caches it entirely separately from the equivalent call on the ModuleObject with no interface wrapping. That would be fixed by tracking the underlying ID/Instance of the implementing object and doing a dagql.Select on it.

Yeah this is something to monitor for sure, and could be an argument in favor of some sort of ID canonicalization process.

>     2. I _think_ this might allow us to get rid of the old TODO comment [here](https://github.com/sipsma/dagger/blob/6d2bd214bcf6c722399ef820c025193eafd9da0e/core/interface.go#L32) about loading returned interface values being expensive (which I agree with) and as a side effect get rid of the kind of complicated/confusing `Query.IDDeps` method (which has been a minor thorn in supporting in some other efforts recently).

:smile: I wrote that comment in the thick of the DagQL conversion because it definitely felt like it was doing a whole lot of work, but I never went back to actually measure it. I've been hoping it's just something that feels like conceptually "a lot" (creating a whole one-off `dagql.Server`) but is actually reasonably performant in practice / never actually a bottleneck.

But, still, feels heavy.

engine/server/session.go
```diff
@@ -306,6 +307,10 @@ func (srv *Server) removeDaggerSession(ctx context.Context, sess *daggerSession)
 type ClientInitOpts struct {
 	*engine.ClientMetadata
 
+	// TODO: doc
```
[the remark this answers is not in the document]
sipsma: I swear I search for `TODO` on my git diffs and still somehow miss at least one everytime 😞 Will fix!

core/interface.go
sipsma: Speculative, worth questioning. I'll do a quick check and rm it if I don't see any obvious performance difference running the tests.

core/schema/host.go
sipsma: True, that would be possible but yeah I think that should be okay in the current state of things

sipsma:
> Making sure I understand: we want to dissect IDs returned by functions to find any embedded Secret/Socket IDs to grant this client access to. But the problem is that sometimes (e.g. with interfaces) the IDs act like symlinks/wrappers, and might not directly embed the socket/secret.

I'd say it's more that it felt sensible to use the dagql ID digest as the identifier for resources in general; it ends up simplifying some things and is relatively low performance overhead. 

For totally "recipe-defined" resources it obviously makes sense, but we can do the same thing for these more "content-defined"-ish resources like secrets/sockets via the accessor pattern (or with `blob` for filesync, which in retrospect is essentially the exact same idea, just swap out SecretStore or SocketStore for the aptly named containerd ContentStore).

So the end result feels nice and unified.

Originally the most immediate motivation to switch to keying everything by ID digest was when I hit bugs around cached function calls.
* Before that, after invoking a function call, I was getting the dagql server of that client and loading every Secret/Socket ID found, then using the accessor key to obtain the secret/socket and put it in the caller's store.
* But that broke with caching (just sdk codegen calls for now) because the invoked function call's client never even ran, so there wasn't even a dagql server from them to use to load those 
* So one option was to just use a different dagql server to load these
* But I realized that if we just keyed by ID digest, then this wouldn't be an issue in the first place since you don't even need to try doing a `dagql.Load` of the IDs to find the accessor. You can find them based purely ID digests
* All those digests are pre-calculated, so it's super low overhead.

Sorry for the rant, not sure how much sense it's making 😅

core/interface.go
sipsma: Yeah no obvious difference so got rid of it for now

PR #7805, opened 2024-07-02
chore: restore default view to top-level dagger module

shykes:
In #7766 we moved `dagger.json` from `./` to `./dev/`. But we forgot to leave the `views` behind - those apply specifically to the root of the repo.

### DIFF
--- a/dagger.json
+++ b/dagger.json
@@ -1,0 +1,16 @@
+{
+ "name": "dagger",
+ "views": [
+  {
+    "name": "default",
+    "patterns": [
+      "!bin",
+      "!.git",
+      "!**/node_modules",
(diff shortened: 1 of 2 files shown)


jedevc:
I'm *reasonably* sure that the views are defined by the module that's being called. I don't think they're defined by the directory being loaded.

If the view wasn't being fetched correctly, we'd see an error message for it `failed to get view: view "default" not found` (which we're not seeing).

shykes:
I think you're right... That's not how I intended views to work, it seems there was a big misunderstanding at implementation.

Anyway views will be effectively deprecated by context directory, so this should take care of itself.

PR #7807, opened 2024-07-02
cli: add quickstart flow for new users

marcosnils:
This commit changes the CLI so it prints a custom Dagger Cloud trace URL
when the user is not logged in or no `DAGGER_CLOUD_TOKEN` has been set.

Additionally, a custom message gets printed at the end of `dagger init`
and fro any commands that the user supplied the `-v` or `--debug` flags.

Fixes [DEV-4269](https://linear.app/dagger/issue/DEV-4269) [DEV-4268](https://linear.app/dagger/issue/DEV-4268)

Here's some screenshots about how the feature looks:

Log-in message when `init` is called and CLI is not authenticated
image

Full logged out trace URL: 
image

Log-in message when `-v` is supplied: 
image

The custom `/traces/$id` URL will take the user to the quickstart flow which will create a Dagger Cloud organization after logging in.

### DIFF
--- a/cmd/dagger/call.go
+++ b/cmd/dagger/call.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log/slog"
 	"os"
 	"path/filepath"
 
@@ -13,10 +12,12 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-var outputFormat string
-var outputPath string
-var jsonOutput bool
-var responsePayload map[string]any
+var (
+	outputFormat    string
+	outputPath      string
+	jsonOutput      bool
+	responsePayload map[string]any
+)
 
 var callCmd = &FuncCommand{
 	Name:  "call [options]",
@@ -68,11 +69,7 @@ var callCmd = &FuncCommand{
 		switch modType.Name() {
 		case Container, Directory, File:
 			if outputPath != "" {
-				respPath, ok := response.(string)
-				if !ok {
-					return fmt.Errorf("unexpected response %T: %+v", response, response)
-				}
-				cmd.PrintErrf("Saved to %q.\n", respPath)
+				logOutputSuccess(cmd, outputPath)
 				return nil
 			}
 		}
@@ -128,13 +125,7 @@ var callCmd = &FuncCommand{
 			if err := writeOutputFile(outputPath, buf); err != nil {
 				return fmt.Errorf("couldn't write output to file: %w", err)
 			}
-			path, err := filepath.Abs(outputPath)
-			if err != nil {
-				// don't fail because at this point the output has been saved successfully
-				slog.Warn("Failed to get absolute path", "error", err)
-				path = outputPath
-			}
-			cmd.PrintErrf("Saved output to %q.\n", path)
+			logOutputSuccess(cmd, outputPath)
 		}
 
 		writer := cmd.OutOrStdout()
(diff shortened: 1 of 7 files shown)


engine/telemetry/cloud.go
```diff
@@ -17,13 +17,15 @@ import (
 	"golang.org/x/oauth2"
 )
 
-var configuredCloudSpanExporter sdktrace.SpanExporter
-var configuredCloudLogsExporter sdklog.Exporter
-var configuredCloudTelemetry bool
-var ocnfiguredCloudExportersOnce sync.Once
```
jedevc, line 23: lol :laughing:

engine/client/client.go
```diff
@@ -323,6 +323,9 @@ func (c *Client) startEngine(ctx context.Context) (rerr error) {
 	if c.CloudCallback != nil {
 		if url, msg, ok := enginetel.URLForTrace(ctx); ok {
 			c.CloudCallback(ctx, url, msg)
+		} else {
+			// fallback to quickstart flow if the user is not logged in.
+			c.CloudCallback(ctx, "https://dagger.cloud/traces/"+uuid.NewString(), "logged out")
```
jedevc, line 328: This is really weird from a usability perspective - my initial read of this PR was that we were uploading *all* traces to dagger cloud, and then allowing users to create an organization with them in here. For this case, we shouldn't be printing a URL that looks *anything* like a traces view URL, it definitely confused me - it's not a real trace URL, so we shouldn't pretend it is.

jedevc: Also note, it's slightly weird that we show the "logged out" message, at the end of a trace in `tui` mode if you're not logged in, but that you don't get the full trace URL if you are logged in?

We should balance that.

marcosnils: > it's not a real trace URL, so we shouldn't pretend it is.

 cc @shykes

shykes: As discussed in the issue - the logged in experience did show the Trace URL when we started working on this. But it has since been removed.

marcosnils: resolving this as I've removed this code since it requires further discussion

marcosnils:
@jedevc just reverted the `call` prints and changed the URL to be more clear about signup. Mind taking one more quick :eyes: ?

jedevc (approved):

jedevc:
LGTM once lint/tests are passing :tada:

marcosnils:
mergint as-is since the test failure is a flake.

cmd/dagger/call.go
```diff
@@ -191,6 +182,17 @@ func writeOutputFile(path string, buf *bytes.Buffer) error {
...
 
+// logOutputSuccess prints to stderr the output path to the user.
+func logOutputSuccess(cmd *cobra.Command, path string) {
+	path, err := filepath.Abs(path)
+	if err != nil {
+		// don't fail because at this point the output has been saved successfully
+		cmd.PrintErrf("WARNING: failed to get absolute path: %s\n", err)
+		path = outputPath
+	}
+	cmd.PrintErrf("Saved output to %q.\n", path)
```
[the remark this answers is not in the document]
marcosnils: this was my bad since I didn't rebase with main before merging and this is something helder fixed recently. Will revert in a follow-up PR. Thx for bringing this up alex.

cmd/dagger/main.go
```diff
@@ -166,6 +167,23 @@ var rootCmd = &cobra.Command{
...
+		switch {
+		case !ok &&
+			cmd.Flags().Changed("verbose") ||
+			cmd.Flags().Changed("debug") ||
+			cmd == moduleInitCmd:
+			fmt.Fprintf(cmd.ErrOrStderr(), `
+Log-in or create an account to visualize %s in Dagger Cloud:
+https://dagger.cloud/signup?quickstart=true
+
+`, highlight.Render(`pipeline traces`))
```
jedevc, line 184: Just noticed this now - this should be `!ok && (...)` - since `&&` has a higher precedence, it means that even when logged in, the message is still shown for `debug` and `moduleInitCmd`.

marcosnils: noted, thx for reporting this one Justin <3

Issue #7673, opened 2024-06-16
🐞 Regression of #7057 with v0.11.7


skycaptain:
### What is the issue?

It appears issue #7057 has recurred in v0.11.7 (also tested with the latest b213c4f88de0f4bfa164e047276d717024b78cb2). Now, it occurs with all progress modes. As before, the issue arises in CI where only `--progress=plain` is available. Long-running tasks may be canceled if nothing is printed for a certain period. It would be beneficial to have the ability to read the log while the build is running.

### Dagger version

dagger dev-071bba814c2411bb8ab5c016897210711f0f26a4 (registry.dagger.io/engine) darwin/arm64

### Steps to reproduce

https://github.com/skycaptain/dagger-issue-7057

jedevc:
Hm, I'm not seeing anything unexpected. Do you have a gif or something similar, or a specific point that you're noticing a lack of progress?

One of the changes we made recently was a decision to hide a lot of the progress under the `initialize` step (which can often be slow). The reason for that is that it's entirely internal to the creation of a module, and isn't code that the user wrote - you can expand that though by passing more verbosity flags - `-vv` should expand more.

If you're not seeing any progress at *all*, exactly as in #7057, this could be due to mismatched client and engine versions (the automagic compatibility check to handle this is disabled in dev builds or versions built directly off of `main`).

skycaptain:
Thanks for your hint using `-vv`. I guess it seems that the changes introduced with v0.11.7 not only hide the progress under the initialise step, but also user-written code.

I've recorded the behaviour of v0.11.6 and v0.11.7 each with and without `--progress=plain` and v0.11.7 with `-vv`. I've check that both versions are using their corresponding engine version.

#### v0.11.6

![2024-06-17 13 26 16](https://github.com/dagger/dagger/assets/1119641/b5fcb9ea-0cb9-45b7-8d08-a174d7e176d4)
![2024-06-17 13 25 03](https://github.com/dagger/dagger/assets/1119641/1bc1bfd5-bcea-49d7-aaae-aa284f6883de)

####  v0.11.7

![2024-06-17 13 08 40](https://github.com/dagger/dagger/assets/1119641/fd9167c9-c8d8-4606-87fa-9f7ae8b37a77)
![2024-06-17 13 08 10](https://github.com/dagger/dagger/assets/1119641/72352ee0-d586-4cfd-9d05-d8bbe20c94b3)

####  v0.11.7 with `-vv`

[...]

PR #7811, opened 2024-07-02
fix(sdk/python): restore telemetry initialization in modules

helderco:
Fixes https://github.com/dagger/dagger/issues/7673

### DIFF
--- a/sdk/python/src/dagger/client/_session.py
+++ b/sdk/python/src/dagger/client/_session.py
@@ -17,7 +17,7 @@ from gql.transport.httpx import HTTPXAsyncTransport
 from opentelemetry import propagate
 from typing_extensions import Self
 
-from dagger import ClientConnectionError
+from dagger import ClientConnectionError, telemetry
 from dagger._config import ConnectConfig, Retry
 from dagger._managers import ResourceManager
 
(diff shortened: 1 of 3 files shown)


sdk/python/src/dagger/client/_session.py
```diff
@@ -172,6 +172,7 @@ def __await__(self):
         return self.connect().__await__()
 
     async def __aenter__(self) -> Self:
+        telemetry.initialize()
```
helderco, line 175: The initialization was setup initially while the SDK establishes a connection to the API, but it was added only to the connection that has automatic provisioning support. So added to this one too, which is the one used in modules.

sdk/python/src/dagger/mod/_module.py
```diff
@@ -189,39 +189,31 @@ def __call__(self) -> None:
...
+            msg = f"Failed to serialize result: {e}"
+            raise InternalError(msg) from e
+
+        logger.debug(
+            "output => %s",
+            textwrap.shorten(repr(output), 144),
+        )
+        await dag.current_function_call().return_value(dagger.JSON(output))
 
-    @serve()
```
helderco, line 224: The decorator was created because of the spans, so removing again.

helderco (commented):
The SDK used to create an OTel span to replace `exec /runtime`. It was removed in  https://github.com/dagger/dagger/pull/7452, which inadvertently also disconnected the telemetry initialization.

skycaptain:
I can confirm that the changes fix issue #7673.

vito (approved):
Whoops, thanks for fixing! :pray:

PR #7812, opened 2024-07-02
engine: prevent dagql cache cross-talk of Query object

sipsma:
(Spinning this commit out of https://github.com/dagger/dagger/pull/7804 because it's super tedious+boring and bloating that other PR too much on top the actual interesting parts.)

To fully isolate client-specific resources within a session, we need to move off of storing things like the SecretStore as fields on Query. This is because the dagql cache is shared across a session, meaning it can return objects from other clients that should not be used.

I just went ahead and got rid of all the fields on Query and turned them into methods. *Technically* it wasn't necessary for all of them, but it's more consistent this way and sets us up for cross-session dagql caching if ever needed.

### DIFF
--- a/core/container.go
+++ b/core/container.go
@@ -281,7 +281,10 @@ func (mnts ContainerMounts) With(newMnt ContainerMount) ContainerMounts {
 }
 
 func (container *Container) From(ctx context.Context, addr string) (*Container, error) {
-	bk := container.Query.Buildkit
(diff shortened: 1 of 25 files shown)


sipsma:
Oops, I cherry-picked this but ended up with some diffs from earlier commits in the other PR, fixing

vito (approved):

Issue #7788, opened 2024-06-29
sdks: `Sync` doesn't always work as expected with certain APIs


sipsma:
A tangential issue I noticed during some adventures with https://github.com/dagger/dagger/pull/7767

At least in the Go SDK, the generated `Sync` method will just return the query it has built so far, e.g. https://github.com/dagger/dagger/blob/adca4e6cb4cd28090bfe5a87c497c93561c9161f/sdk/go/dagger.gen.go#L2102-L2106

However, this does not play nicely with our core APIs that rely on returning a `dagql.Instance` of a different API call, e.g. https://github.com/dagger/dagger/blob/adca4e6cb4cd28090bfe5a87c497c93561c9161f/core/host.go#L63-L69

While `Sync` is correctly called, the SDK doesn't use the returned ID and instead uses the query it already built up.

This is especially wrong in the case of impure APIs, since it will result in the API being repeatedly called incorrectly with the "original" ID.
* e.g. if the query is to load a local dir and sync is called, then any subsequent calls on the returned Directory will re-run the load of the local dir.

In terms of the practical impact today, not 100% sure, it did break new APIs I was adding in #7767 but:
1. I'm not sure if it causes significant incorrect/unperformant behavior for existing core APIs. It seems possible but would require more tracing of code and testing
2. I only looked at the Go SDK so far, not sure what others are impacted if any

cc @vito @helderco @jedevc

helderco:
It was done this way because at the time we didn't have a good way to convert any ID to its corresponding object. That changed with the `loadXFromID` fields. 

I left a comment in codegen, but was removed in https://github.com/dagger/dagger/commit/7393e19bbbb2ae2ed07c93af09183f079d0a83b5#diff-de9a6011ccf4884745982bf7658dae4ec2e8d2564dd5a5b1063ef7dbe5678de7L145-L153:

<blockquote><pre>
// FIXME: <strong>We don't have a simple way to convert any ID to its
// corresponding object (in codegen) so for now just return the
// current instance.</strong> Currently, `sync` is the only field where
// the error is what we care about but more could be added later.
// To avoid wasting a result, we return the ID which is a leaf value
// and triggers execution, but then convert to object in the SDK to
// allow continued chaining. For this, <strong>we're assuming the returned
// ID represents the exact same object but if that changes, we'll
// need to adjust.</strong>
</pre></blockquote>

helderco:
[...]

PR #7817, opened 2024-07-03
fix(sdk/elixir): sync functions should return struct not an id type

wingyplus:
In every `sync` functions, load the new state using load from id function to return a type itself. And also fix test file missing in the `dagger_codegen` to make it pass. I just realize that `dagger_codegen` didn't get test in the CI. 🙈 I'll open a follow-up PR to improve this. 

This updates #7788 since it's a fix only Elixir SDK.


helderco (approved):

wingyplus:
Re-running the CLI tests

PR #7818, opened 2024-07-03
feat: expand tilde in secret, directory and file argument

wingyplus:
In this PR, add expand tilde (`~`) to home directory on directory, file and secret file source argument. The home directory rely on `os.UserHomeDir()` that using `$HOME` for unix and `$env:USERPROFILE` on Windows. 

Ref for discussion: https://discord.com/channels/707636530424053791/1003718839739105300/1258097710851162145

### DIFF
--- a/cmd/dagger/flags.go
+++ b/cmd/dagger/flags.go
@@ -6,6 +6,7 @@ import (
 	"crypto/sha256"
 	"encoding/csv"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -285,6 +286,10 @@ func (v *directoryValue) Get(ctx context.Context, dag *dagger.Client, modSrc *da
 	// POSIX "portable filename character set":
 	// https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282
 	path, viewName, _ := strings.Cut(path, ":")
+	path, err = expandHomeDir(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to expand home directory: %w", err)
+	}
 	path = filepath.ToSlash(path) // make windows paths usable in the Linux engine container
 
 	return modSrc.ResolveDirectoryFromCaller(path, dagger.ModuleSourceResolveDirectoryFromCallerOpts{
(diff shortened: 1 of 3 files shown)


jedevc:
Should we also do this for file/directory values?

wingyplus:
@jedevc Yeah, support both of it would be great!

cmd/dagger/flags.go
```diff
@@ -423,7 +425,16 @@ func (v *secretValue) Get(ctx context.Context, c *dagger.Client, _ *dagger.Modul
...
 	case fileSecretSource:
-		filePlaintext, err := os.ReadFile(v.sourceVal)
+		sourceVal := v.sourceVal
+		// For unix, automatic expand tilde (~) to $HOME
+		if strings.HasPrefix(sourceVal, "~") {
+			usr, err := user.Current()
+			if err != nil {
+				return nil, err
+			}
+			sourceVal = strings.Replace(sourceVal, "~", usr.HomeDir, 1)
```
wingyplus, line 435: What do you think of using `homedir.Expand` like we did in Go SDK?

helderco: We used to use https://github.com/mitchellh/go-homedir instead of `os/user` for this.

helderco: Maybe it's no longer an issue though:
- https://github.com/mitchellh/go-homedir/issues/34

wingyplus: I see. Will take a look at it tomorrow. 🙏

wingyplus: I ended up by using `os.UserHomeDir` to get home directory from shell environment but copy path detection from go-homedir instead. 🙇

wingyplus:
@helderco @jedevc The PR is ready for review. 🙏

cmd/dagger/flags.go
```diff
@@ -285,7 +286,12 @@ func (v *directoryValue) Get(ctx context.Context, dag *dagger.Client, modSrc *da
...
 	path, viewName, _ := strings.Cut(path, ":")
+	path, err = expandHomeDir(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to expand home directory: %w", err)
+	}
 	path = filepath.ToSlash(path) // make windows paths usable in the Linux engine container
+
 	return modSrc.ResolveDirectoryFromCaller(path, dagger.ModuleSourceResolveDirectoryFromCallerOpts{
 		ViewName: viewName,
 	}).Sync(ctx)
```
wingyplus, line 297: May I ask why directory value need sync while file value is not?

jedevc (approved):

cmd/dagger/flags.go
jedevc: Ahh, this is so that if the path doesn't exist we can error here - instead of later, after we've done a bunch of work.

But yes, this is weird, we'd like to get rid of it at some point I think.

wingyplus:
Rebased and add changie log.

wingyplus:
I found the `Engine & CLI / test` failure due to `TestIntrospection` in `dagql` package. I submitted a fix in https://github.com/dagger/dagger/pull/7846 to make it pass but not sure if it is the right fix.

Issue #7786, opened 2024-06-28
tests: 5-10m block possibly during `TestExecError` in GHA engine tests


sipsma:
I am not entirely sure what's precisely going on here yet, so just going to list the details I've gathered so far. The end effect is that we seem to be getting multi-minute hangs in GHA when printing certain output. The tests still pass thankfully but it appears to possibly be increasing CI test job time a lot.

---

One example:
1. Cloud traces: https://dagger.cloud/dagger/traces/dda1b716e38a06eee8f434b500b6c16e#87f6ab6bc7234db9
2. GHA job): https://github.com/dagger/dagger/actions/runs/9711496785/job/26804333495

For some reason, the GHA job output seems to more clearly show this hang.

In there, the test output moves quickly for a while until here:
```
2024-06-28T10:49:44.3199842Z [95m2603: [0m    Container.[1msync[0m[2m: ContainerID![0m
2024-06-28T10:49:44.3200401Z [95m2604: [0m    exec sh -c base64 -d encout >&1; base64 -d encerr >&2; exit 1
2024-06-28T10:55:21.7314561Z [95m2604: [0m    [90m[0.1s] [0m[90m|[0m <truncated really really long string from TestExecError>
```

Note the 6m gap before the output of the exec is printed ([exec in question here](https://github.com/sipsma/dagger/blob/0290df5df67b2514a0dd045652b0083017a86b14/core/integration/container_test.go#L3306)).

After the long string, there's another 5m time gap before the next line, after which everything proceeds fast again:
```
2024-06-28T11:00:02.2189162Z [95m2604: [0m    exec sh -c base64 -d encout >&1; base64 -d encerr >&2; exit 1[33m ERROR[0m[90m [0.1s][0m
2024-06-28T11:00:02.2190162Z [95m2604: [0m    [33m! process "sh -c base64 -d encout >&1; base64 -d encerr >&2; exit 1" did not complete successfully: exit code: 1[0m
2024-06-28T11:00:02.2190948Z [95m2603: [0m    Container.[1msync[0m[33m ERROR[0m[90m [0.2s][0m
2024-06-28T11:00:02.2191772Z [95m2603: [0m    [33m! process "sh -c base64 -d encout >&1; base64 -d encerr >&2; exit 1" did not complete successfully: exit code: 1[0m
[...]

PR #7819, opened 2024-07-03
tests: avoid GHA slowdown from huge single line strings

sipsma:
*Possibly* fixes, or at least helps, https://github.com/dagger/dagger/issues/7786 by avoiding https://github.com/actions/runner/issues/1031

### DIFF
--- a/dagql/idtui/frontend_plain.go
+++ b/dagql/idtui/frontend_plain.go
@@ -20,6 +20,8 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
+const plainMaxLiteralLen = 256 // same value as cloud currently
+
 type frontendPlain struct {
 	FrontendOpts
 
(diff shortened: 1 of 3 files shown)


dagql/idtui/frontend_plain.go
```diff
@@ -509,7 +511,7 @@ func (fe *frontendPlain) renderLogs(row *TraceTree, depth int) {
 	span := row.Span
 	spanDt := fe.data[span.ID]
 
-	r := newRenderer(fe.db, -1, fe.FrontendOpts)
+	r := newRenderer(fe.db, 1024, fe.FrontendOpts)
```
vito, line 514: missed a spot?

```suggestion
	r := newRenderer(fe.db, plainMaxLiteralLen, fe.FrontendOpts)
```

vito (approved):

dagql/idtui/frontend_plain.go
sipsma: Doh, fixed

sipsma:
Hit a separate known flake (python main module missing) but the times in the test runs here seem to align much more closely. I think some difference is expected since there's a little overhead in starting everything (especially testdev), but so far looks good. Will see if that continues to be the case in more runs after merging.

PR #7820, opened 2024-07-03
fix infinite logging loop (and CPUPROFILE)

vito:
Currently if you run with `--debug` the CLI will go into a loop constantly logging the same log line repeatedly until it's eventually just a bunch of `\\\\\\\\\\`s from escaping strings in strings.

This was introduced by the recent TUI changes which directed logs to go to the primary span rather than to a separate TUI-only pane. Unfortunately, as part of the log handling (TUI consumes logs), we were logging at debug level, hence `--debug` introducing an infinite loop for any log line printed.

I think I can live without these logs now that draining seems fixed (knock on wood) so we'll just remove them. This was already done for the plain TUI, just not the pretty one.

(Also, fixes `$CPUPROFILE` along the way, since it's been broken for a while.)

### DIFF
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -120,9 +120,8 @@ var rootCmd = &cobra.Command{
 			if err != nil {
 				return fmt.Errorf("create profile: %w", err)
 			}
-			defer profF.Close()
+
 			pprof.StartCPUProfile(profF)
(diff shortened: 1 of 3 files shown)


crjm (approved):
`--debug` flag works!

:shipit:

Issue #7801, opened 2024-07-01
TUI regression?


shykes:
## Problem

In `main`, output from the TUI stays on the terminal after completion. This is different from the behavior on the current stable release, 0.11.8.

- On 0.11.8:

```console
$ dagger call -m github.com/shykes/hello hello
hello, world!
```

- On main (3767c8f0b3076a41c218c545c96641080b0e819b):

```console
$ dagger call -m github.com/shykes/hello hello
✔ connect 0.1s
✔ initialize 3.1s
✔ prepare 0.0s
✔ hello: Hello! 0.0s
✔ Hello.hello: String! 0.2s

hello, world!
```

## Solution

Revert the behavior to be the same as 0.11.8.

vito:
This was an intentional change. We would frequently get questions that came down to users not knowing about `-v`, so printing nothing at the end didn't really feel like a sane default to me. I found myself passing `-v` nearly all the time because running something, tabbing away, and coming back to having no visibility into what actually ran, and having nothing useful in my scrollback after repeated runs, felt pretty frustrating and not in line with the TUI experience from other tools.

shykes:
OK, it's definitely a fair point to revisit. A heads up would be nice though, this is a known bike shedding hotspot, since nobody brought it up I assumed it was unintentional.

shykes:
>  I found myself passing -v nearly all the time because running something, tabbing away, and coming back to having no visibility into what actually ran, and having nothing useful in my scrollback after repeated runs, felt pretty frustrating and not in line with the TUI experience from other tools.

But was that in the context of developing or debugging the engine or TUI itself? As an end user of Dagger, I can't remember ever being in that situation *unless the pipeline fails*, in which case I do get TUI output. When my pipeline doesn't fail, I find it distracting to have residue from its progress in my terminal. What use case am I missing that lead to people asking for more TUI residue when their pipeline succeeds?

vito:
My main concern is that the quiet behavior is a sane default for toy examples like the "hello world" above but is detrimental to the experience of using Dagger "in anger" because it eliminates the return on investment for a slow command (build/test/lint). I've got 10,000 lines of scrollback and only so many hours in a day; I'd rather spend the scrollback than spend time chasing ghosts because ambient information was taken from me to keep the output clean (concrete example for this later).

> But was that in the context of developing or debugging the engine or TUI itself?

[...]

PR #7822, opened 2024-07-04
tui: add `-q`/`$DAGGER_QUIET`, better durations + errors

vito:
CLI verbosity scheme is now:

* `-s`/`--silent`: total silence, no progress displayed ever
* `-q`/`--quiet`/`$DAGGER_QUIET`: show progress, clean up completed progress
* default: show progress, keep completed items
* `-v`: show internal and encapsulated progress
* `-vv`: show spammy progress, reveal digests
* `-d`/`--debug`: show everything, set debug log level, reveal span IDs and other misc things

WIth `$DAGGER_QUIET` set, use `-v` to achieve the original default, which also matches the v0.11.x behavior.

Some fixes along the way:

* Final render now respects `-s`

### DIFF
--- a/.changes/unreleased/Added-20240704-183832.yaml
+++ b/.changes/unreleased/Added-20240704-183832.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: "cli: add -q flag and DAGGER_QUIET=1 to restore previous verbosity default"
+time: 2024-07-04T18:38:32.603222928-04:00
+custom:
+  Author: vito
+  PR: "7822"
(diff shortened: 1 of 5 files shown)


helderco:
I think this strikes the right balance. Just my own 2c is that I felt the lingering progress annoying so love the option to silence that as I was frustrated that I couldn't disable, but I don't feel that strongly about making it the default, after knowing that it brings value to others.

jedevc (approved):

vito:
Started adding more on top but don't want to go beyond what's approved, so splitting it out. Added a changelog, so merging!

Also: there's _still a chance_ that we go back on this, and in fact I'm slightly leaning towards undoing this later since you can adjust the level at runtime anyway if you forget `-v`. If you have a preference, vote here:

* https://github.com/dagger/dagger/issues/7801#issuecomment-2207798136

PR #7824, opened 2024-07-04
fix: init in current directory instead of `dagger`

### LINKED ISSUE #6963: SDKs should generate files in current working directory by default, not ./dagger
## Problem

When initializing an SDK with `dagger develop` or `dagger init --sdk`, source files are generated in a subdirectory called `./dagger` by default. It seems that most users are either confused by that default, or actively dislike it. I personally always use `--source .`. I would prefer for `.` to be the default, and I think most users will agree.

## Solution

Change the default for `dagger develop --source` and `dagger init --source` to `.` instead of `./dagger`.

## Context

There is actually a reason for the current behavior:

- We encourage users to create a Dagger module at the root of their application repo.
- We also encourage users to (eventually) develop their CI logic as functions in that top-level repo. For this, they use `dagger develop` or `dagger init --sdk`.
- In this situation, it is often preferable to tuck away the generated source code in a subdirectory, instead of polluting the repo root. Thus, the choice of `./dagger`
- However, in all other cases, `.` makes more sense

Nnote: I initially advocated for `./dagger` as the default, so I'll take the L 😅

TomChv:
Fixes #6963

### DIFF
--- a/.changes/unreleased/Changed-20240719-120907.yaml
+++ b/.changes/unreleased/Changed-20240719-120907.yaml
@@ -1,0 +1,7 @@
+kind: Changed
+body: SDK now generate files in the current directory by default and no longer in
+  `./dagger`
+time: 2024-07-19T12:09:07.3248+02:00
+custom:
+  Author: TomChv
+  PR: "7824"
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -53,7 +53,7 @@ var (
 const (
 	moduleURLDefault = "."
 
-	defaultModuleSourceDirName = "dagger"
+	defaultModuleSourceDirName = "."
 )
 
 func init() {
@@ -66,7 +66,7 @@ func init() {
 
 	moduleInitCmd.Flags().StringVar(&sdk, "sdk", "", "Optionally initialize module for development in the given SDK")
 	moduleInitCmd.Flags().StringVar(&moduleName, "name", "", "Name of the new module (defaults to parent directory name)")
-	moduleInitCmd.Flags().StringVar(&moduleSourcePath, "source", "", "Directory to store the module implementation source code in (defaults to \"dagger/ if \"--sdk\" is provided)")
+	moduleInitCmd.Flags().StringVar(&moduleSourcePath, "source", "", "Directory to store the module implementation source code in (defaults to \".\"/ if \"--sdk\" is provided)")
 	moduleInitCmd.Flags().StringVar(&licenseID, "license", defaultLicense, "License identifier to generate - see https://spdx.org/licenses/")
 
 	modulePublishCmd.Flags().BoolVarP(&force, "force", "f", false, "Force publish even if the git repository is not clean")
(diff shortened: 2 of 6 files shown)


TomChv:
I identified few tests that fails become of this changes, I'm currently fixing each of them

TomChv:
Still having issues fixing the tests, specially:  https://github.com/dagger/dagger/blob/ab7c1ae9dbe35a7c78674e892e06faa564d5d69e/core/integration/module_config_test.go#L1144

I don't understand why it's not finding the `fn` functions...

```
          ✘ go 1m37.7s
          ! test failed
          ┃                         2   : 24280b98c853dc7b: prepare ERROR [0.0s]                                                                       
          ┃                         2   : 24280b98c853dc7b: ! unknown command "fn" for "dagger call"                                                   
          ┃                                                                                                                                            
          ┃                         Error: unknown command "fn" for "dagger call"                                                                      
          ┃                         Run 'dagger call --help' for usage.                                                                                
          ┃         Test:           TestModule/TestIncludeExclude/go                                                                                   
          ✘ python 2m42.2s
          ! test failed
          ┃                         2   : 5f9eebda17b70c15: prepare ERROR [0.0s]                                                                       
          ┃                         2   : 5f9eebda17b70c15: ! unknown command "fn" for "dagger call"                                                   
          ┃                                                                                                                                            
          ┃                         Error: unknown command "fn" for "dagger call"                                                                      
          ┃                         Run 'dagger call --help' for usage.                                                                                
          ┃         Test:           TestModule/TestIncludeExclude/python                                                                               
          ✘ typescript 2m32.6s
          ! test failed
          ┃                         2   : d8153cf6f0eda008: prepare ERROR [0.0s]                                                                       
          ┃                         2   : d8153cf6f0eda008: ! unknown command "fn" for "dagger call"  
```

TomChv:
Update: got some tests to fix but not much and should be good, it's the first item on my list for tomorrow

TomChv:
I fixed couple of Go tests, I'm waiting for the CI if there's more to do

TomChv:
GoExtendCore failed on the CI but is good locally, it's a flack, all other tests have been fixed!

This Pr is ready for a review @helderco

helderco (commented):
lgtm, but not approving because we shouldn't merge this yet. There are some discussions that need to be resolved first, namely this ["dev module" pattern](https://discord.com/channels/707636530424053791/1260894336883626025) thread in Discord.

It's ok not to keep the tests rebased and updated until we've decided to merge.

shykes (approved):
After discussing with @helderco and @jedevc today, I think we should go ahead and do this.

- It will immediately make life easier for module authors. When creating a standalone module, `.` is clearly better than `dagger` as a default source dir
- On its own, this does not solve the problem of "how to embed a dagger module in an existing project". So we will need some follow-up design work. But the most promising design (sdk-specific native sub-packages) seems compatible with this change.
- Meanwhile, this does not make the experience significantly worse for embedded modules.

So it's a leap of faith, but one worth taking!

sipsma (approved):
LGTM, I added the needs-changelog label since we definitely want that before merging

.changes/unreleased/Changed-20240719-120907.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Changed
+body: SDK now generate files in the current directory by default and no longer in
+  `./dagger`
```
helderco, line 3: Isn't this a syntax error?

PR #7825, opened 2024-07-04
fix: ensure terminal dump id is readable

jedevc:
Previously, `DumpID` was producing bad output, after lots of refactors to the frontend. Specifically, we were no longer sending newlines after `renderCall` which meant that everything appeared on the same line (additionally, `DumpID` wasn't taking into account that the client terminal is in raw mode, so needs `\r` at each newline).

To resolve this, we create a new little helper struct called `Dump`, which allows configuration of basic options, and can be configured by the dagger engine to get some more readable output.

Before:

	❯ dagger call container-echo --string-arg=x terminal
	● Attaching terminal: container: Container!Container.from(address: "alpine:latest"): Container!Container.withExec(args: ["echo", "x"]): Container!

	dagger / $

After:

	❯ dagger call container-echo --string-arg=x terminal
	● Attaching terminal:
	    container: Container!
	    Container.from(address: "alpine:latest"): Container!
	    Container.withExec(args: ["echo", "x"]): Container!

	dagger / $

There's still future possible improvements of course, but this is a *lot* more readable than just squashing everything into one horrible to read line.

### DIFF
--- a/cmd/dump-id/main.go
+++ b/cmd/dump-id/main.go
@@ -19,7 +19,7 @@ func main() {
 		panic(err)
 	}
 
-	if err := idtui.DumpID(idtui.NewOutput(os.Stdout), &idp); err != nil {
+	if err := new(idtui.Dump).DumpID(idtui.NewOutput(os.Stdout), &idp); err != nil {
 		panic(err)
 	}
 }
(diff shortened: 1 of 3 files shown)


vito (approved):
Much better!

Could probably do with more pixel pushing, like repeating the `Container` type so often is probably not necessary, we could make it look like one call chain instead.

But yeah, that's more work, this is a clear improvement.

PR #7826, opened 2024-07-04
docs: move k8s integrations resources to the bottom

marcosnils:
mostly doing this since the results currently returned by the query
are not properly organized and curated.


gerhard (approved):
👍 🚀

Issue #7821, opened 2024-07-03
CLI: don't run a container's default args when printing it


shykes:
## Context

Since #6515, the CLI prints objects returned by a pipeline. To do this, it calls certain functions in the returned object, specifically functions which 1) return a printable value, and 2) have no required arguments.

## Problem

When a pipeline returns a `Container` object, the CLI calls `Container.stdout()` and `Container.stderr()`. In some cases<sup>1</sup>, this will unexpectedly execute a long-running service specified in the container's `defaultArgs`.

*[1] normally, `stdout` and `stderr` return the result of the last `withExec` operation. But if there is no `withExec` in the container state, `defaultArgs` are executed as a fallback. That is what causes the unexpected behavior.*

For example:

```console
dagger call -m github.com/dagger/dagger/dev@3767c8f0b3076a41c218c545c96641080b0e819b \
 docs server
```

- Expected behavior: print container metadata, and return control to the user
- Actual behavior: execute the `defaultArgs` (in this case a long-running server), and hang forever

## Solution

~~Manually exclude `Container.stdout` and `Container.stderr` from the list of functions to call when printing an object. If any other functions in `Container` cause it to be executed, exclude that too.~~

Change the behavior of `Container.stdout` and `Container.stderr` so that it doesn't fallback to executing `defaultArgs`. Also make sure its behavior no longer depends on how many `withExec`s are in the container spec.

shykes:
FYI @helderco

helderco:
I guess that answers this, right? 🙂 

> _On the question of "expensive scalar functions_: for example, `Container.stdout()` is "expensive" because it forces actually executing a container, in order to retrieve its standard output. Therefore, it is argued, the CLI should not blindly call all scalar functions of an object, because it may cause expensive computation which the user does not want. _This is premature optimization_. Let's ship the feature in a naive implementation; then see how important it actually is to avoid expensive computation. My guess is that, in most cases, it won't matter. For example, if I query a container, I don't care that it's actually executed. Or, if it turns out it's a problem, then we can implement a special case in the core type. I certainly won't care if my own custom types don't benefit from such a special case.

\- _Originally posted by @shykes in https://github.com/dagger/dagger/issues/6515_

shykes:
@helderco yeah I didn't think about the case where executing the container runs a server and never returns...

So it looks like I got this part wrong:

> For example, if I query a container, I don't care that it's actually executed.

And it's time to do this part:

[...]

PR #7829, opened 2024-07-04
tui: don't print Container.stdout/stderr

vito:
This felt particularly spammy:

[asciicast](https://asciinema.org/a/PU8bCnemL5aVmkuLnHmFRpmlb)

So, applying the same logic from `File.contents` here.

Alternatively I suppose we could truncate?


helderco:
Fixes https://github.com/dagger/dagger/issues/7821

Are there any more scalars that trigger execution?

helderco (approved):

vito:
> Are there any more scalars that trigger execution?

Oof, actually this runs deeper than I thought - this is a breaking change, because previously we'd call `Sync`, and now we were relying on collecting `Stdout` and `Stderr` to force evaluation as before, but without those it just prints container metadata and doesn't run the container.

Is this still what we want? Was there a lot of bikeshedding over this that I missed? Seems like it'll break a bunch of folks' CI entrypoints, if they were following the 'just return a Container' practice (by far the easiest thing to do imo). Are we no longer encouraging that? Is there a suggested alternative?

shykes:
Fixes #7853

helderco:
Pushed the `sync` fix.

helderco:
Failures are unrelated.

cmd/dagger/call.go
```diff
@@ -40,6 +42,18 @@ var callCmd = &FuncCommand{
...
+		// There's no fields in `Container` that trigger container execution so
+		// we use `sync` first to evaluate, and then load the new `Container`
+		// from that response before continuing.
+		if typeName == Container {
+			var cid dagger.ContainerID
+			fc.Select("sync")
+			if err := fc.Request(ctx, &cid); err != nil {
+				return err
+			}
+			fc.q = fc.q.Root().Select("loadContainerFromID").Arg("id", cid)
```
vito, line 54: nonblocking, more forward-looking: I think we don't actually need to do this since `sync` will always just return the original ID with DagQL. Not sure if it mattered before, but we also added this before `Void` iirc. I wonder if `sync` should just return `Void` now?

helderco: Not according to:
- https://github.com/dagger/dagger/issues/7788

helderco: >  I wonder if `sync` should just return `Void` now?

The original reason for returning `ContainerID` was actually to make it easier to continue chaining and avoid the need for a `Void` at the time:

```diff
- ctr := test()
- _, err := ctr.Sync(ctx)
+ ctr, err := test().Sync(ctx)
if err != nil {
    return "", err
}
return ctr.Stdout(ctx)
```

Doesn't seem that different but see Python:

```diff
- ctr = test()
- await ctr.sync()
- return await ctr.stdout()
+ return await (await test().sync()).stdout()
```

But now that we have `Void`, and if we really don't need to load from ID, considering also that there's https://github.com/dagger/dagger/pull/7773, then the exception here stands out a bit. `sync` is the only one that works like this so it needs special handling in codegen.

PR #6680, opened 2024-02-15
feat: split out go module defs into package

jedevc:
Finally! This patch splits up the codegen for go modules to produce the main external struct defs into a separate sub-package. There are a couple of reasons for doing this:
- Tighter control over exposed structs and their internal private methods
- Ability to import the generated structs from a sub-package

Before:

```
❯ tree .
.
├── api.go
├── dagger.gen.go
├── dagger.json
├── go.mod
├── go.sum
├── main.go
├── querybuilder
│   ├── marshal.go
│   └── querybuilder.go
└── README.md

2 directories, 9 files
```

After:

```
❯ tree .
.
├── api.go
├── dagger
│   └── dagger.gen.go
├── dagger.gen.go
├── dagger.json
├── go.mod
├── go.sum
├── main.go
├── querybuilder
│   ├── marshal.go
│   └── querybuilder.go
└── README.md

3 directories, 10 files
```

Essentially, we split up `dagger.gen.go` into a top-level `dagger.gen.go`, and a sub-package `dagger/dagger.gen.go`.

- Top-level `dagger.gen.go` contains type-aliases to all the public sub-package types, any required interfacey/struct method implementations, and the implementation of `main` and `invoke`.
- Packaged `dagger/dagger.gen.go` contains all the go-generated templates for the served schema types.
	- This package can now also be imported by sibling packages, see the new test `TestModuleGoUtilsPkg`, which tests for #6661.

---

Initially this might sound simple - and prior to the addition of interfaces, it was, however, these add a lot of confusion. Specifically, interface codegen is added alongside the main code, and includes adding new methods on user-defined structs as well as creating an implementation of user-defined interfaces.

So where should these new definitions belong? They actually have to go at the same level as the definition of the struct/interface, so that 1. we can actually attach the methods, and 2. so we can actually access the interface in arg/return types (we can't just copy the interface into the sub-package, since it might also refer to user-defined concrete types, which we can't copy).

But this results in a couple of complications:
- We defined a LoadFromID method for our own interfaces - we can no longer attach these to the client, since the client is now part of the sub-package. This is easy to resolve, we just make this a top-level method, it doesn't actually *need* to be on the client object, that was just a convention.
[...]

PR #7831, opened 2024-07-05
feat: remove go aliases

jedevc:
Completes the work from https://github.com/dagger/dagger/pull/6680 by removing the aliases, which were really tricky to work with from a code completion perspective.

By using #7759, we can avoid breaking the daggerverse, and just apply this for v0.12 and after.

With this change, all dagger types will need to be referenced directly from the `internal/dagger` package. Before:

```go
package main

import (
	"context"
)

type MyModule struct{}

func (m *MyModule) ContainerEcho(stringArg string) *Container {
}
```

After (see the change from `Container` to `dagger.Container`):

```go
package main

import (
	"context"
	"dagger/my-module/internal/dagger"
)

type MyModule struct{}

func (m *MyModule) ContainerEcho(stringArg string) *dagger.Container {
}
```


helderco:
I support this! It makes it more consistent with the other SDKs, generates less code, and makes the IDE happier because of hidden docstrings in the aliases.

jedevc:
AH! An issue I wasn't expecting to encounter - there are deps in our dependency tree (see https://github.com/shykes/daggerverse/pull/10), which have random hashes as their `engineVersion`, which defaults to the latest.

This is fiddly, and not great. For now, we can probably *make do* by ensuring that our dependencies have good hashes anyways.

But long term, we *really* need to do #7760 - I'm going to add that to our milestone as well.

jedevc:
Tests are failing due to https://github.com/dagger/dagger-test-modules/pull/3.

jedevc:
Hm, one more test is failing, `TestModule/TestGoExtendCore` - not quite sure why, will dig into this asap tomorrow. Otherwise, is ready for review.

jedevc:
I pointed https://github.com/vito/daggerverse` to my fork at https://github.com/jedevc/daggerverse-vito, so I could test with https://github.com/dagger/dagger/pull/7831.

sdk/go/.changes/unreleased/Breaking-20240709-143029.yaml
```diff
@@ -0,0 +1,9 @@
+kind: Breaking
+body: |-
+  Remove top level type alises for modules
```
helderco, line 3:
```suggestion
  Remove top level type aliases for modules
```

helderco (approved):
LGTM!

cmd/codegen/generator/go/templates/module_objects.go
```diff
@@ -419,6 +419,14 @@ func (spec *parsedObjectType) concreteFieldTypeCode(typeSpec ParsedType) (*State
 		if typeSpec.alias != "" {
 			s.Id(typeSpec.alias)
 		} else {
+			tp := typeSpec.GoType()
+			if basic, ok := tp.(*types.Basic); ok {
+				if basic.Kind() == types.Invalid {
+					s.Id("any")
+					break
+				}
```
vito, line 427: This bit seems interesting - is it to prevent codegen from blowing up on an unknown type? Maybe because it gets figured out on another pass?

cmd/codegen/generator/go/generator.go
```diff
@@ -385,23 +389,28 @@ package main
 
 import (
 	"context"
+	"%[2]s/internal/dagger"
```
vito, line 392: I suppose now if I never run `dagger develop` and the generated module code isn't committed, instead of my editor giving me a bunch of errors for undefined types, there will just be this one unknown import, which is better

vito (approved):
Will be nice to skip the extra hop for go-to-definition, although I guess for that what I'd really want is some sort of magical go-to-backing-implementation feature (instead of the codegen'd client).

Also, I guess you could `.`-import if you preferred the syntax sugar from before?

Anyway, approving, didn't find anything wonky

cmd/codegen/generator/go/templates/module_objects.go
jedevc: Yup, that's the behavior in [`c50cb10` (#7831)](https://github.com/dagger/dagger/pull/7831/commits/c50cb109a0a71c8d92d3e97da6c2e340c2be16ce). It produces invalid code, but the code wouldn't have compiled in the first place - this means `dagger develop` can at least succeed, so you can actually fix the issues, instead of having to just "guess".

cmd/codegen/generator/go/generator.go
jedevc: That too I guess! The main reason for this in here is that if you *don't* do it, go cleverly attempts to "find it" - and often seems to select `dagger.io/dagger` which is reasonably unhelpful.

vito: I think that issue still exists but for `dag` right? I remember that auto-importing from somewhere. :thinking: Or is that fixed by now too?

jedevc: Ah, that issue exists for users if they don't run `dagger develop` first - since it's in `dagger.gen.go`. Sadly, I don't know about how to fix that one :thinking:

PR #7843, opened 2024-07-08
fix: various core dynamic schema follow-ups

jedevc:
Collection of various follow ups from #7759:

- Cloning issues - without doing cloning in the right way, we can get all sorts of fun issues, where deps are shared in confusing ways.
- Update golden testdata for dagql
- Elegantly handle a dagger.json without an `engineVersion` (assume it's from an older dagger version, before we had `engineVersion`)

### DIFF
--- a/core/moddeps.go
+++ b/core/moddeps.go
@@ -46,8 +46,12 @@ type ModDeps struct {
 func NewModDeps(root *Query, mods []Mod) *ModDeps {
 	return &ModDeps{
 		root: root,
-		Mods: mods,
+		Mods: append([]Mod{}, mods...),
 	}
+}
+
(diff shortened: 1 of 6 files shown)


TomChv (approved):

core/moddeps.go
```diff
@@ -46,10 +46,14 @@ type ModDeps struct {
 func NewModDeps(root *Query, mods []Mod) *ModDeps {
 	return &ModDeps{
 		root: root,
-		Mods: mods,
+		Mods: append([]Mod{}, mods...),
```
jedevc, line 49: Lol, this was fun :cry: What was happening was that the `defaultDeps` were being modified - then this affected all the other module loading. To avoid this, we just need to be *very* careful, and always copy the input.

PR #7854, opened 2024-07-09
helm: Align App (Engine) version to Chart version

### LINKED ISSUE #7597: Align Helm Chart version with the Dagger version + start tracking a changelog
We did this with all SDKs in summer 2023:
- https://github.com/dagger/dagger/discussions/5374

We should do the same for the Helm chart (it's the only odd one out) and also start a changelog for it. As discussed here:
- https://github.com/dagger/dagger/pull/7596#discussion_r1633290716

gerhard:
So that there is no extra mental hop between the Chart version and the Dagger Engine version. We do the same for all other artefacts that we release.

This also adds changie for the Helm chart & automates the version bump.

Resolves #7597


gerhard:
I ran this locally & committed the version bump to `0.12.0`. Rolling back now so that CI will do this when we release `v0.12.0`

I still want to add a few changes before we release `v0.12.0`, but we shouldn't hold the release over those changes.

jedevc (approved):

jedevc:
Nice!

gerhard:
Confirmed that `needs/changelog` works as expected:

image

PR #7857, opened 2024-07-09
core: don't fallback to the default command on `stdout` and `stderr`

helderco:
Fixes https://github.com/dagger/dagger/issues/7821


helderco:
@shykes should this go on v0.12.0?

Note that this is covered by the compat layer, so modules with an `engineVersion` to an older stable release should work without change, unless `dagger develop` is used.

core/container_exec.go
```diff
@@ -319,13 +321,7 @@ func (container *Container) WithExec(ctx context.Context, opts ContainerExecOpts
...
 func (container *Container) MetaFileContents(ctx context.Context, filePath string) (string, error) {
 	if container.Meta == nil {
-		ctr, err := container.WithExec(ctx, ContainerExecOpts{
-			UseEntrypoint: true,
-		})
-		if err != nil {
-			return "", err
-		}
-		return ctr.MetaFileContents(ctx, filePath)
+		return "", fmt.Errorf("%w: include an exec step with empty arguments to use the default command", ErrNoCommand)
```
vito, line 324: nit: terminology might be worth bikeshedding, don't know if 'include an exec step' ("step" in particular) will get the idea across.

Strawman:

```suggestion
		return "", fmt.Errorf("%w: call withExec with empty arguments to use the default command", ErrNoCommand)
```

Which has the downside of hardcoding GraphQL casing convention, but feels more explicit. Super low conviction though.

vito (approved):
One nit, lgtm either way!

core/container_exec.go
helderco: Yeah, agree that bikeshedding is needed, forgot to flag it 🙂

helderco: > has the downside of hardcoding GraphQL casing convention

Yeah, exactly what I was trying to avoid. The reasoning was that `withExec` adds an `exec`, so maybe users would understand.

vito: I'm OK with it as-is, "step" feels like it's in roughly the same family as "pipeline" which we also use colloquially without a firm meaning, so maybe we just keep using it in situations like this.

helderco: @shykes, how would you phrase this error?

```go
out, err := src.DockerBuild().Stdout(ctx)
```
```console
error: no command has been set: include an exec step with empty arguments to use the default command
```

shykes: I wouldn't mention the empty arguments or default command. I would simply say 'stdout: nothing to execute'. Or perhaps the slightly more detailed: `stdout: nothing to execute. Call withExec() first`

helderco: Ok, I'll get back to the error I had initially:
```
error: no command has been set: stdout requires an exec
```

Really trying not to reference an API field due to different naming conventions in SDKs, and I believe that `exec` is clarifying enough. Let me know if there's strong feelings against it.

Kept "no command has been set" part because it's the error that's returned in a `withExec` where there's no arguments to use, for example, no entrypoint or default args and `withExec(args: [])`.

So, if a user uses `stdout` and **has no default command**, they'll first get:
> error: no command has been set: stdout requires an exec

And if they add an empty `withExec` after that, they'll get:
> error: no command has been set

shykes: I would just keep the second part: `stdout requires an exec`. The first part confuses me. my 2c

helderco: Problem is I'm depending on the wrapping of that error to catch it on the legacy compat layer, but I'll move that around.

shykes: don't worry about it if it's too much work - not worth it

PR #7858, opened 2024-07-09
chore: tag all releases with semver

jedevc:
Fixes https://github.com/dagger/dagger/issues/7760

Instead of just using commit hashes and directory IDs, which produces nearly incomprehensible versions (which is more of a pain, now that `engineVersion` is actually important). This makes it a *bit* clearer when and where a non-tagged build came from by using changie to get the last release we built.

Frustratingly, it seems non-trivial to do version comparison on these, so we keep skipping these. In the future, we should work out on compromise. For an example of why this is hard: suppose that we add a MinVersion of v0.12.0 - the next pre-release will end up being something like "v0.11.10-dev-240709191414-1ccb8772510a". So this version will fail these checks.

We could add a little piece of code to note what the *next* release will be - which would make these comparison a lot simpler (essentially, we'd treat pre-releases of vX as equivalent to X). But this would require too much extra maintainance for now - but maybe we could dupe this with MinVersion numbers in `version.go` today.

image

You get a semver tag!

### DIFF
--- a/.github/workflows/engine-and-cli-publish.yml
+++ b/.github/workflows/engine-and-cli-publish.yml
@@ -27,11 +27,24 @@ jobs:
         with:
           go-version: "1.22"
           cache-dependency-path: "dev/go.sum"
+      - name: "Determine tag"
+        run: |
+          mkdir -p /tmp/publish
+          echo "${{ github.ref_name == 'main' && github.sha || github.ref_name }}" > /tmp/publish/tag
+      - name: "Determine version"
+        uses: ./.github/actions/call
+        with:
+          # determine the version before releasing any components - this keeps
+          # it consistent across all components (since pre-releases might have
+          # timestamps)
+          function: --version="$(cat /tmp/publish/tag)" version string
+          redirect: "/tmp/publish/version"
       - name: "Publish Engine"
         uses: ./.github/actions/call
         with:
           function: |
-            --version="${{ github.ref_name == 'main' && github.sha || github.ref_name }}" \
+            --version="$(cat /tmp/publish/version)" \
+            --tag="$(cat /tmp/publish/tag)" \
             engine \
             publish \
             --platform="linux/amd64" --platform="linux/arm64" \
@@ -49,7 +62,8 @@ jobs:
         uses: ./.github/actions/call
         with:
           function: |
-            --version="${{ github.ref_name == 'main' && github.sha || github.ref_name }}" \
+            --version="$(cat /tmp/publish/version)" \
+            --tag="$(cat /tmp/publish/tag)" \
             engine with-base --image=ubuntu --gpu-support=true \
             publish \
             --platform="linux/amd64" \
@@ -67,7 +81,8 @@ jobs:
         uses: ./.github/actions/call
         with:
           function: |
-            --version="${{ github.ref_name == 'main' && github.sha || github.ref_name }}" \
+            --version="$(cat /tmp/publish/version)" \
+            --tag="$(cat /tmp/publish/tag)" \
             engine with-base --image=wolfi \
             publish \
             --platform="linux/amd64" \
@@ -85,7 +100,8 @@ jobs:
         uses: ./.github/actions/call
         with:
           function: |
-            --version="${{ github.ref_name == 'main' && github.sha || github.ref_name }}" \
+            --version="$(cat /tmp/publish/version)" \
+            --tag="$(cat /tmp/publish/tag)" \
             engine with-base --image=wolfi --gpu-support=true \
             publish \
             --platform="linux/amd64" \
@@ -103,7 +119,8 @@ jobs:
         uses: ./.github/actions/call
         with:
           function: |
-            --version="${{ github.ref_name == 'main' && github.sha || github.ref_name }}" \
+            --version="$(cat /tmp/publish/version)" \
+            --tag="$(cat /tmp/publish/tag)" \
             cli \
             publish \
             --git-dir=./.git \
--- a/.goreleaser.common.yml
+++ b/.goreleaser.common.yml
@@ -12,6 +12,7 @@ builds:
     ldflags:
       - -s -w
       - -X github.com/dagger/dagger/engine.Version={{.Env.ENGINE_VERSION}}
+      - -X github.com/dagger/dagger/engine.Tag={{.Env.ENGINE_TAG}}
     goos:
       - linux
       - windows
--- a/core/integration/engine_test.go
+++ b/core/integration/engine_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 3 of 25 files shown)


jedevc:
This follows the idea in https://go.dev/doc/modules/version-numbers.

sipsma:
One thing I just noticed, we construct the default runner host to connect to based on `Version`: https://github.com/sipsma/dagger/blob/b06d45a7e0e026bbe557994314afb1f09282cf8e/engine/consts.go#L26-L26

This will continue working for tagged releases, but I think the change here would create a disconnect for CLIs published off main in that now we publish the engine image using just the git commit SHA and `Version` is more involved.

Getting CLIs built off main to automatically provision the correct engine image built off main by default was a pretty nice property to have, so it would be nice to maintain. It's not an utterly crucial requirement though

dev/version.go
```diff
@@ -44,11 +54,64 @@ func (info *VersionInfo) String() string {
...
+		return "", err
+	}
+	h := sha1.New() //nolint:gosec
+	h.Write([]byte(id))
+	dgst := hex.EncodeToString(h.Sum(nil))
+	return dgst, nil
+}
+
+func pseudoversionTimestamp(t time.Time) string {
+	return t.Format("060102150405")
```
sipsma, line 153: comment here would help 😄

dev/version.go
```diff
@@ -44,11 +54,64 @@ func (info *VersionInfo) String() string {
...
+	if nextVersion == "" {
+		nextVersion = "v0.0.0"
+	}
 	if info.Commit != "" {
-		return info.Commit
+		return fmt.Sprintf("%s-%s-%s", nextVersion, pseudoversionTimestamp(time.Now()), info.Commit[:12])
 	}
 	if info.Dev != "" {
-		return "dev-" + info.Dev
+		return fmt.Sprintf("%s-%s-dev-%s", nextVersion, pseudoversionTimestamp(time.Now()), info.Dev[:12])
```
sipsma, line 68: This timestamp will end up being used as the cache key for nested execs, so I think it will result in more invalidations than are necessary.

Not the utter end of the world, just a local dev testing thing, but I do like the property of this being purely content hashed from the dir. Can we just use all zeros for the timestamp?

sipsma (commented):
Makes sense on a high level! I do greatly prefer the explicit file with the next version that you added in the last commit over inferring from changie (since whether the next version is a patch or minor bump is a human decision)

dev/version.go
jedevc: Yeahhhh, bleh. Good call, fixed this to use all zeroes. All `-dev` versions are now stripped neatly :tada: So `vX.Y.Z-dev-123` is counted as compatible with `vX.Y.Z-dev-456`.

jedevc:
> Getting CLIs built off main to automatically provision the correct engine image built off main by default was a pretty nice property to have, so it would be nice to maintain. It's not an utterly crucial requirement though

Oof, just missed this till now - I've split this behavior out to populate `Tag` separately, so we can preserve this.

.goreleaser.common.yml
```diff
@@ -12,6 +12,7 @@ builds:
     ldflags:
       - -s -w
       - -X github.com/dagger/dagger/engine.Version={{.Env.ENGINE_VERSION}}
+      - -X github.com/dagger/dagger/engine.Tag={{.Env.ENGINE_VERSION}}
```
sipsma, line 15: Don't we want to set this to a different value, i.e. `WithEnvVariable("ENGINE_TAG", cli.Dagger.Version.TagString()` here https://github.com/sipsma/dagger/blob/2a9bb343e013194b9153739fe17a06c55282a841/dev/cli.go#L85

jedevc: Eek, yes.

jedevc: Hm, actually, the timestamps are an issue here :thinking: Because we publish the cli and engine in separate steps, the timestamps will be different, which is weird.

I guess we can get the timestamp in github actions, and pass it everywhere, but that's really annoying :cry:

jedevc:
Marking as draft while I rework this a bit.

jedevc:
So I think this is getting much closer to ready - this changes some release things, so I'll keep an eye on main once it's merged. However, the major changes are:

- Every build gets semver tags
- All builds cut directly off of `main` pretty much keep working as they do today
- All dev builds will happily work with each other (as long as they share the same major/minor/patch) part of their build - we all roughly know what we're doing.
- The compat mode for a module with engineVersion `vX.Y.Z-<anything>` is always `vX.Y.Z`
	- Note: we should still *strongly* advise against doing this - imagine the following scenario:
	- We add breaking change 1 to dagger
	- You generate a module with a version with this engineVersion
	- We add breaking change 2 to dagger
	- Your module is still requesting the same api version - and so your module breaks.
	- This is expected! There's really no way to avoid this without having every single breaking change in dagger introduced with a timestamp (I thought about this for a good while, I don't really like it, but we could always return to it if it's a sticking point)
	- The *only* other option that I thought of is that `vX.Y.Z-dev` could serve the schema `vX.Y.(Z - 1)` - but then this means that builds cut off of `main` won't work with new features (making it harder to pick up), you'd need to manually specify `--compat=vX.Y.Z` - that feels fiddly - and would probably some weird test hackery.

Future work:
- Because we add a new link-time variable, we need to update *everywhere* we do builds of dagger:
	- The downstream homebrew formula: https://github.com/Homebrew/homebrew-core/blob/master/Formula/d/dagger.rb#L28-L33
	- @jpadams are there others I need to be aware of? It would be nice to not break folks.
- Daggerverse should have some way to ensure that modules with pre-release engine versions are carefully rejected - or at least highlighted.
	- As mentioned above, we really really want to make sure that users aren't really using these for general use.

.goreleaser.common.yml
jedevc: Fixed this in the last commit, by adding a new `ENGINE_TAG` env that's populated from the new `--tag` parameter. `dev` builds won't have this set, so we should error out - which makes sense, the current behavior is to try and pull a non-existent image.

Additionally, the timestamps in version numbers are an issue for the github multi-step release - this is resolved by just having a version pre-calculation step, which is messy, but eventually we'll shift all of this logic into dagger, so it'll hopefully all go away...

core/integration/engine_test.go
sipsma, line 1: Just saw in the tests that we only have cases where the engine sets Version and MinVersion to the same values. Would it be valuable to have additional tests where Version>MinVersion? That's the situation we'd find ourselves in most often so IIUC it seems like good to have too. Might be misunderstanding something though.

Also as a nit: these tests are **extremely** valuable for understanding what the intended behavior is in the first place (almost documentation of it), which makes me wonder if it would be worth the effort to refactor them as table-driven tests.

Then we can just easily look through every permutation of (engine version, engine min version, client version, client min version) in a centralized place (as opposed to a bunch of separate tests with boilerplate overhead) and know what the expected behavior is. It would also probably make it easier to see if we are missing some particular permutation.
* I realize this might be a tedious change from the current state, so not a blocker for me at all, just a suggestion if you agree

sipsma (commented):
LGTM so far, couple significant comments and some nits.

I'd generally say that we ***really*** need to lean on the tests here. There are just so many dimensions (`(version, min version) x (engine, client) x (dev, not dev)`) that it gets really hard to think about all of them purely by reading the code. That's just inherent though, so whatever we can do to make sure we have as much test coverage as feasible in a readable way will pay dividends I think.

core/integration/engine_test.go
jedevc: Mm, agreed - I started doing some of this in `version_test.go` with pretty much exactly this kind of test (a lot more lightweight as well, though they aren't integration tests).

Does that kind of work? I can add more inline doc comments to explain those if they're currently unclear. I think we could keep the heavier integration tests with the current scenarios, and then use `version_test.go` to cover all the more detailed cases.

The reason I didn't go for table tests in the first place was that spinning up a bunch of engines with different versions is relatively heavy. We could definitely do it. The other problem is that the `_EXPERIMENTAL_DAGGER_VERSION` isn't *perfect* - it's the best way to test it right now, but really we'd want to have dagger dev-module tests I think to make sure they're properly propagated from the build process.

Still worth doing?

.github/workflows/engine-and-cli-publish.yml
```diff
@@ -27,11 +27,24 @@ jobs:
...
+          mkdir -p /tmp/publish
+          echo "${{ github.ref_name == 'main' && github.sha || github.ref_name }}" > /tmp/publish/tag
+      - name: "Determine version"
+        uses: ./.github/actions/call
+        with:
+          # determine the version before releasing any components - this keeps
+          # it consistent across all components (since pre-releases might have
+          # timestamps)
+          function: --version="$(cat /tmp/publish/tag)" version string
+          output: "/tmp/publish/version"
```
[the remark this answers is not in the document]
jedevc: Oops, yes.

engine/version.go
```diff
@@ -12,18 +14,20 @@ var (
...
+	// - For builds off our repo's main branch, this is a pre-release of the
+	//   form vX.Y.Z-<timestamp>-<commit>
+	// - For builds off our repo's main branch, this is a pre-release of the
+	//   form vX.Y.Z-<timestamp>-dev-<dirhash>
 	Version string
 
 	// MinimumEngineVersion is used by the client to determine the minimum
 	// allowed engine version that can be used by that client.
-	MinimumEngineVersion = "v0.12.0"
+	MinimumEngineVersion = "v0.12.1"
```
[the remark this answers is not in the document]
jedevc: Hmmm, I did this for a reason - and I now cannot remember it. We can always re-bump it later if it turns out to have been for a good reason.

engine/consts.go
```diff
@@ -23,7 +34,11 @@ func RunnerHost() (string, error) {
 		return v, nil
 	}
 
-	tag := Version
+	tag := Tag
+	if tag == "" {
+		// can happen during dev builds
+		return "", fmt.Errorf("no builtin image tag available")
```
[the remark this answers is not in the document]
jedevc: Yeaaaa, this is already weird uncharted territory, so this makes sense to me.

core/integration/engine_test.go
jedevc: > That might make the code overall easier to keep track of too. WDYT?

Yeah, this is honestly the most important thing for this change, so I've just sucked it up and converted it - it's an absolute pain, but it's now done :tada:

Now it should be super easy to add new tests there, I don't *think* there's anything major that's not covered?

sipsma (approved):

PR #7860, opened 2024-07-09
engine: prevent dnsmasq from shutting down prematurely

sipsma:
@matipan noticed that in some integration tests where the engine is run as a service, when SIGTERM is sent to the engine it was also being sent to dnsmasq, which caused it to shutdown quickly and interfere with the engine's attempt to gracefully shutdown (e.g. flush cache, which can rely on dnsmasq still existing).
* related discord thread: https://discord.com/channels/707636530424053791/1260319690002141286/1260319691906220082

This turned out to be due to the fact that dumb-init, by default, will forward signals to the whole process group, so when it sends SIGTERM to the engine, that signal also went to dnsmasq.

After looking around, it *seems* like forwarding signals to the whole process group is probably a better default behavior and closer to expectations (though there's certainly no clear cut answer). A commonly cited example of this across the docs for different init implementations is that it's the same behavior you get when starting processes with a shell.
* e.g. the fact that pressing ctrl-C on `sh -c 'sleep 10'` results in an immediate exit relies on the fact that SIGINT gets sent to the whole process group, including `sleep 10` and not just `sh`.

In general, there will certainly be use cases that arise where users want some different behavior here. However, I think the best way to support that is to add an option to Container.withExec that allows users to disable use of an automatic init if they want.
* To be clear, it's important that we default to an automatic init since init-less containers behave in highly surprising+confusing ways. So this would just be an option for advanced users who really know what they are doing and need some specialized behavior that calls for no init or swapping in their own init process with custom config.

So given that, keeping the current behavior of signaling the whole process group feels like the least surprising default, with the option to add a knob in the future for advanced users.

The fix here thus just updates the dnsmasq subprocess to handle this situation better by:
1. Being spawned in a separate process group, thus preventing signals forwarded to the engine's process group from being forwarded to dnsmasq too
2. Setting PDEATHSIGNAL to SIGKILL, which basically means that dnsmasq will die when the engine dies, which is what we want in our particular case I think.

---

TODOs:
- [x] Haven't tested yet, @matipan since you had the repro pulled up would it be possible to give this change a quick try there and verify it works now?
   - If there's more changes needed I'll set everything up locally so you don't have to be an intermediary, just figured this would be a fast path to start 😄

### DIFF
--- a/network/netinst/dnsmasq.go
+++ b/network/netinst/dnsmasq.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
+	"syscall"
 )
 
 func InstallDnsmasq(ctx context.Context, name string) error {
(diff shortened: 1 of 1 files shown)


sipsma:
Oh I totally just had a flashback to how PEATHSIGNAL is not safe in Go (due to the runtime's intrinsic, not user-controlled threading) unless you use runtime.LockOSThread. Will update

matipan:
This is amazing!! Thanks for the quick fix. Will run the tests locally with this engine and report back.

Regarding the default behavior: it makes sense to follow what init processes usually do. In the tests I've done both docker and containerd behave by just sending sigterm to the main process, but they also do not add an init process that handles this. Having a knob for it is the best of both worlds. Non advanced users do not have surprising behaviors and advanced users get to still control the shutdown process as they see fit.

matipan:
Mm, magicache's tests are getting stuck. It seems we reach the point of stopping the engine service but that never finishes, in the engine logs that is controlling that service I can see the detach logs:
```
time="2024-07-09T22:24:19Z" level=debug msg="detach: service still has binders" bindings=3 service=v2fmagseenie6.s53mi37a4asbu.dagger.local
```

Need to leave now, will debug tomorrow morning!

matipan:
> Mm, magicache's tests are getting stuck. It seems we reach the point of stopping the engine service but that never finishes, in the engine logs that is controlling that service I can see the detach logs:
> 
> ```
> time="2024-07-09T22:24:19Z" level=debug msg="detach: service still has binders" bindings=3 service=v2fmagseenie6.s53mi37a4asbu.dagger.local
> ```
> 
> Need to leave now, will debug tomorrow morning!

Fixed the getting stuck, it was an issue with our integration tests. Now getting an error for dagger version mismatch between client and engine server. I'll do the upgrade and report back!

matipan (approved):

PR #7861, opened 2024-07-10
fix(ts-sdk): test detect node-version

TomChv:


jedevc (approved):

core/integration/module_typescript_test.go
```diff
@@ -601,12 +601,10 @@ func (ModuleSuite) TestTypescriptRuntimeDetection(ctx context.Context, t *testct
 		require.Error(t, err)
 	})
 
-	// This test will need to be updated when a new lts version is released
-	// https://nodejs.org/en/about/previous-releases
-	t.Run("should detect pinned lts node version", func(ctx context.Context, t *testctx.T) {
+	t.Run("should detect specificpinned node version 20.15.0", func(ctx context.Context, t *testctx.T) {
```
helderco, line 604: Isn't this already being tested in the next subtest named "should detect a specific pinned node version 22.4.0"?

You can use `node@tls` but I suggest that rather than asserting for the exact version you change the test to **confirm that node gets installed correctly**.

You can actually check if a node runtime is LTS or not, but I don't think it's necessary:
```console
❯ docker run -it --rm node:lts-alpine -e "console.log(process.release.lts)"
Iron
❯ docker run -it --rm node:21-alpine -e "console.log(process.release.lts)"
undefined
```

However, if people use `node@lts` they could go from one LTS to another, and that may be a breaking change. It's better if they're explicit and upgrade manually.

PR #7862, opened 2024-07-10
fix(codegen): load object when output needs to be converted from ID in Go and TypeScript

helderco:
Fixes https://github.com/dagger/dagger/issues/7788

Didn't affect just `sync`. Not sure if the affected fields in `Service` had issues before this.


helderco:
Failures are unrelated.

vito (approved):
Makes sense!

PR #7863, opened 2024-07-10
ci: Allow skipping specific tests - e.g. TestModule - or only run specific ones

gerhard:
> [!NOTE]
> While this PR started by splitting our Engine & CLI tests in `TestModule` and everything else, it has since been changed. We are keeping the existing test suite **as is**, while adding two new test suites:
> 1. `test-module`
> 2. `test-everything-else`
>
> This is meant to inform the next steps for:
> - https://github.com/dagger/dagger/issues/8184

---

This PR replaces the `test custom` subcommand with `test specific`, e.g.:

```console
dagger call --source=.:default test specific --skip=TestModule
```

In the example above, we run all tests **except** the `TestModule` ones. To **only** run `TestModule` tests, the command is:

```console
dagger call --source=.:default test specific --run=TestModule
```

Also, `test important` is now:

```console
dagger call --source=.:default test specific --run='TestModule|TestContainer'
```

Here are a few more examples:

```console
# Only run TestContainer, TestDirectory and TestFile tests
dagger call --source=.:default test specific --run="TestContainer|TestDirectory|TestFile"

# Skip all tests starting with TestLoad
dagger call --source=.:default test specific --skip="TestLoad"

# Skip all tests starting with TestC, TestD, TestE, ..., TestZ
dagger call --source=.:default test specific --skip="Test[C-Z]"
```

All tests are now running in verbose `-v` mode so that we can see the `go test` output in addition to the Dagger Cloud Traces. We can't fully trust those yet **in this context**. [FTR](https://github.com/dagger/dagger/issues/8184).

The other useful feature about the `--skip` flag is that now we can add additional metadata to these runs. In the following example, I am running all tests and capturing the specific instance properties in the `--skip` flag (somewhat hacky, but essential for distinguishing runs in Dagger Cloud):

```console
dagger call --source=.:default test specific --skip="32C-6548N|64GB-DDR5|VIRTIO-NVME"
```

This also adds a new test subcommand which lists all top-level tests in alphabetical order:

```console
dagger call --source=.:default test list

TestBaseVersion
TestBasic
TestBuiltins
TestCache
TestCacheMapConcurrent
TestCacheMapErrors
TestCacheMapRecursiveCall
TestCheckResponse
TestClient
TestContainer
TestCoreModTypeDefs
TestDefToDAG
TestDefaults
TestDirectory
TestEmptyID
TestEngine
TestEngineNameLabel
TestEnums
TestFile
TestFromDir
TestGPU
TestGit
TestHTTP
TestHost
TestIDFormat
TestIDsDoNotContainSensitiveValues
TestIDsReflectQuery
TestImpureIDsReEvaluate
TestInputObjects
TestIntrospection
TestIsSemver
TestLegacy
TestLexicalRelativePath
TestListResults
TestLoadCircleCILabels
TestLoadClientLabels
TestLoadGitHubLabels
TestLoadGitLabLabels
TestLoadGitLabels
TestLoadGitRefEnvLabels
TestLoadJenkinsLabels
TestLoadSecretsToScrubFromEnv
TestLoadSecretsToScrubFromFiles
TestLoadServerLabels
TestLoadingFromID
TestMatchGoImport
TestMatchVersion
TestModule
TestNamespaceObjects
TestNormalizeVersion
TestNullableResults
TestObject
TestObjects
TestOriginToPath
TestParallelism
TestParallelismFlag
TestParseAuthAddress
TestParseDaggerToken
TestParseGit
TestParseMetaGoImports
TestParsePragmaComment
TestParsePublicRefString
TestPassingObjectsAround
TestPlatform
TestPureIDsDoNotReEvaluate
TestRegistryAuthProvider
TestRemoteCache
TestRepoRootForImportPath
TestScrubSecretLogLatency
TestScrubSecretWrite
TestSecret
TestSecretScrubWriterWrite
TestSecretStore
TestSecretStoreNotFound
TestService
TestServicesStartConcurrentHappy
TestServicesStartConcurrentSad
TestServicesStartConcurrentSadThenHappy
TestServicesStartHappy
TestServicesStartHappyDifferentServers
TestServicesStartSad
TestSortInputFields
TestSpanName
TestSplitRequiredOptionalArgs
TestTelemetry
TestTrie
TestTrieExtend
TestTrieReinsert
TestType
TestTypeDefConversions
TestValidateRepoRoot
TestVersionCompatibility
TestViews
TestViewsCaching
TestViewsIntrospection
```

---

> [!NOTE]
> While this is how this PR started, I am only keeping it here so that the first several comments make sense. Only the content above this line is relevant for the merged changes.

This PR splits our Engine & CLI tests into:

We now run them as separate CI jobs which splits the test data and makes it easier to understand where the slowness is coming from. It should also make it easier to focus our test refactoring efforts on module tests since the rest of our test suite doesn't seem to be a problem.

Furthermore, before this change, running all these tests in CI was putting pressure on a bunch of components as they generate **a lot** (>200MB) of output. Even though https://dagger.cloud processes all this data, we then keep stalling the browser tab since it cannot handle a DOM that large. Yup, hundreds of megabytes.

When I run module tests locally, they take over 1h which makes them impractical. All other tests complete within 10 minutes (regardless whether we use `-race` or not). This is what I've learned:

### TestModule
- `3642s` (timed out) @ Wed 10 Jul 08:42:25 BST 2024
- `3632s` (timed out) @ Wed 10 Jul 10:12:09 BST 2024

### TestContainer
- `442s` @ Wed 10 Jul 08:49:47 BST 2024
- `280s` @ Wed 10 Jul 12:06:55 BST 2024

### TestEngine
- `141s` @ Wed 10 Jul 08:52:08 BST 2024
- `120s` @ Wed 10 Jul 12:12:53 BST 2024
- `115s` @ Wed 10 Jul 12:15:21 BST 2024

### All other test suites combined
- `166s` @ Wed 10 Jul 07:38:34 BST 2024
- `146s` @ Wed 10 Jul 07:41:43 BST 2024
- `145s` @ Wed 10 Jul 10:43:38 BST 2024
- `136s` @ Wed 10 Jul 11:55:14 BST 2024

This PR also changes:
- adds `dagger call ... test list`
- `dagger call ... test custom` is now `dagger call ... test specific`
- removes `dagger call ... test important`
- `important` tests now means everything else apart from modules, which we know are the heavy part of our test suite

---

Another interesting observation when running `TestModule` **only** locally: the NVMe disk gets fully saturated (22k writes @ 26MB/s gives us an average of 1KB / write operation). All other resources (CPU / memory / network) are plenty:

Screenshot 2024-07-10 at 16 46 15

Screenshot 2024-07-10 at 16 49 29

---

I am keen to see how this behaves in our CI. I am also curious to hear what others think about this change, especially @sipsma & @jedevc that have been doing a lot of work in this area recently.

### DIFF
--- a/.dagger/test.go
+++ b/.dagger/test.go
@@ -36,36 +36,40 @@ func (t *Test) All(
 	// +optional
 	race bool,
 ) error {
-	return t.test(ctx, "", "./...", failfast, parallel, timeout, race, 1)
+	return t.test(ctx, "", "", "./...", failfast, parallel, timeout, race, 1)
 }
 
-// Run "important" engine tests
-func (t *Test) Important(
-	ctx context.Context,
-	// +optional
-	failfast bool,
-	// +optional
-	parallel int,
-	// +optional
-	timeout string,
-	// +optional
-	race bool,
-) error {
-	// These tests give good basic coverage of functionality w/out having to run everything
-	return t.test(ctx, `^(TestModule|TestContainer)`, "./...", failfast, parallel, timeout, race, 1)
+// List all tests
+func (t *Test) List(ctx context.Context) (string, error) {
+	cmd, err := t.testCmd(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	return cmd.
+		WithExec([]string{"sh", "-c", "go test -list=. ./... | grep ^Test | sort"}).
+		Stdout(ctx)
 }
 
-// Run custom engine tests
-func (t *Test) Custom(
+// Run specific tests
+func (t *Test) Specific(
 	ctx context.Context,
+	// Only run these tests
+	// +optional
 	run string,
+	// Skip these tests
+	// +optional
+	skip string,
 	// +optional
 	// +default="./..."
 	pkg string,
+	// Abort test run on first failure
 	// +optional
 	failfast bool,
+	// How many tests to run in parallel - defaults to the number of CPUs
 	// +optional
 	parallel int,
+	// How long before timing out the test run
 	// +optional
 	timeout string,
 	// +optional
@@ -74,12 +78,13 @@ func (t *Test) Custom(
 	// +optional
 	count int,
 ) error {
-	return t.test(ctx, run, pkg, failfast, parallel, timeout, race, count)
+	return t.test(ctx, run, skip, pkg, failfast, parallel, timeout, race, count)
 }
 
 func (t *Test) test(
 	ctx context.Context,
-	testRegex string,
+	runTestRegex string,
+	skipTestRegex string,
 	pkg string,
 	failfast bool,
 	parallel int,
@@ -91,6 +96,7 @@ func (t *Test) test(
 	args := []string{
 		"go",
 		"test",
+		"-v",
 	}
 
 	// Add ldflags
@@ -125,8 +131,12 @@ func (t *Test) test(
 	// Disable test caching, since these are integration tests
 	args = append(args, fmt.Sprintf("-count=%d", count))
 
-	if testRegex != "" {
-		args = append(args, "-run", testRegex)
+	if runTestRegex != "" {
+		args = append(args, "-run", runTestRegex)
+	}
+
+	if skipTestRegex != "" {
+		args = append(args, "-skip", skipTestRegex)
 	}
 
 	args = append(args, pkg)
--- a/.github/actions/call/action.yml
+++ b/.github/actions/call/action.yml
@@ -134,7 +134,7 @@ runs:
       if: always() && inputs.upload-logs == 'true'
       uses: actions/upload-artifact@v4
       with:
-        name: call-logs-${{ runner.name }}-${{ inputs.function }}
+        name: call-logs-${{ runner.name }}-${{ github.job }}
         path: /tmp/actions/call/call.log
         overwrite: true
 
--- a/.github/workflows/engine-and-cli.yml
+++ b/.github/workflows/engine-and-cli.yml
@@ -46,6 +46,28 @@ jobs:
           function: "test all --race=true --parallel=16"
           upload-logs: true
 
+  test-modules:
+    runs-on: "${{ github.repository == 'dagger/dagger' && 'dagger-g2-v0-12-5-16c-st-od' || 'ubuntu-latest' }}"
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v4
+      - name: "test"
+        uses: ./.github/actions/call
+        with:
+          function: "test specific --run=TestModule --race=true --parallel=16"
+          upload-logs: true
+
+  test-everything-else:
+    runs-on: "${{ github.repository == 'dagger/dagger' && 'dagger-g2-v0-12-5-16c-st-od' || 'ubuntu-latest' }}"
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v4
+      - name: "test"
+        uses: ./.github/actions/call
+        with:
+          function: "test specific --skip=TestModule --race=true --parallel=16"
+          upload-logs: true
+
   # Run Engine tests in dev Engine so that we can spot integration failures early
   # Only run a subset of important test cases since we just need to verify basic
   # functionality rather than repeat every test already run in the other targets.
@@ -57,7 +79,7 @@ jobs:
       - name: "testdev"
         uses: ./.github/actions/call
         with:
-          function: "test important --race=true --parallel=16"
+          function: "test specific --run='TestModule|TestContainer' --race=true --parallel=16"
           dev-engine: true
           upload-logs: true
 


jedevc:
Hm, interested to hear @sipsma's thoughts as well, but I'm not entirely convinced on the value of splitting these out. That said, I don't really see any *huge* harm in it. I do think though that we should feel the pain of having big jobs be slow, etc - I know we're trying to get test runs under 10 minutes, but I think we should do that by fixing actual performance issues, instead of splitting up the runs. There's also a small risk that by *reducing* the amount of parallelism we have in our tests we are less likely to encounter some of the tricky parallel bugs we find like https://github.com/dagger/dagger/issues/7900 and https://github.com/dagger/dagger/issues/7662 (which while bad for regular users, is useful for us).

We know that the module tests are the heaviest - a huge number of our new tests are added as module tests, since we're very much module focused at the moment. In practice - we could probably even do to split our `ModuleSuite` into more specific suites, like ones that test `call`, querying, etc. cc @vito

There's definitely an issue with high disk usage (which needs some dedicated investigation time) - but does splitting them actually reduce disk usage? I guess the jobs are more split out, so we reduce contention for a single machine.

vito:
Shower thought tier analysis:

It seems like our module test suite spends a ton of time calling `dagger init` in each test, downloading the same dependencies and building the same empty initial module over and over again. I wonder if we could share that across tests so they all share a common starting point, one per SDK?

Either that or we need to somehow magically speed up `dagger init` - something has to give.

Maybe our tests could even share the same client/session?

There's always a risk of getting further from reality as we optimize the tests, but the other side of that coin is whether our tests are realistic as-is, and if we're bending over backwards to support unrealistic scenarios. Not making a call there one way or another, just raising the point, since our test suite growth strikes me as more ad hoc than planned/designed. In production for example it seems more common to have a single global client. 

I do agree that I'd rather try optimizations like above than divvy up the suite.

sipsma:
First off, the metrics are super useful @gerhard, thank you! We've definitely had vague awareness that disk was something of a bottleneck previously, but I'm still surprised the extent to which it is based on this. Extremely useful to know.

That being said, I agree with the sentiment that we should understand the performance problems more deeply before refactoring the tests. If we drill down the metrics to a finer grain and use that to fix the performance problems then we are benefitting both users and ourselves rather than just updating our own tests.

Specifically, I think we need to at least know the answer to "what processes are consuming the disk bandwidth" based on actual measurements (though the finer grain the better, e.g. what a specific process is doing that is consuming bandwidth). 

100% agree that the SDKs and reading/writing deps is a top suspect, but IME it's very much worth confirming with data since you will 99.9% of the time discover at least one surprising performance bottleneck once you measure it.
* And if that is the case, we should see what we can do to cache these better. If there's nothing that can be done (whether fundamentally or just on a practical level), then yeah we can see if just refactoring the tests will avoid it for now.

@gerhard I know you've done trials with parca and such, do you think those or anything similar would be able to help get deeper metrics on the source of the disk bandwidth hogs?

gerhard:
I agree with the "keep investigating" part, and I think that we should release our `main` branch from the grip of these flakes. This is the current state:

image

The context is more nuanced, and while [one new genuine issue](https://github.com/dagger/dagger/pull/8155) is responsible for nearly half of all the failures mentioned above, the other half is the following tests in `TestModule`:
1. `TestPythonProjectLayout`
2. `TestGoInit`
3. `TestGit`

If we want to keep the test suite as is, I think that we should do the following:
1. Delete all tests that flake. Keep deleting until there are no more flakes left.
2. Keep all tests as they are in a separate branch, and block at least 1h sync time every day until we get to the bottom of this. I would go as far as pulling the [Andon cord](https://en.wikipedia.org/wiki/Andon_(manufacturing)) and not doing anything else until this is fixed. This can mean rewriting the tests, deleting them, or something else. We should give ourselves a timebox (1 week?) and force a conclusion at the end of that period.

In the spirit of moving this forward, I intend to:
- [ ] Create a special tag in our repository
- [ ] Submit the first PR which deletes the 3 flakes in `TestModule`
- [ ] Book 1h sync time in our calendars as of next week to dig into this as small group

How does the above sound?

jedevc:
These are also potentially legitimate issues:

- `TestPythonProjectLayout` - https://github.com/dagger/dagger/issues/7662
- `TestGoInit` - https://github.com/dagger/dagger/issues/8031
- `TestGit` - not seen this one, do you have a link?

The first two seem potentially related - files are getting cached incorrectly, it was expected that this was fixed as part of https://github.com/dagger/dagger/issues/7900, but it seems like it still continues to happen.

The problem with removing these test is - there's nothing in these tests that we can identify that *specifically* causes the flake. Potentially, if we remove these tests today, maybe some new tests we add tomorrow will trigger the same behavior? To clarify - the tests *are correct* - something in the engine is wrong. I don't think removing the tests is the right call. `TestGoInit` specifically, we've relied on to make quite a few recent changes, for example #8095. Without those tests, we could have caused huge breakage.

Potentially, we could have a retry loop for these known flaky tests? That way we can continue to run them, but we give them an amount of tolerance to pass? That's still hacky and annoying, but the tests do test for valuable behavior.

> block at least 1h sync time every day

I don't think there's a time that works consistently for everyone - without asking folks to wake up early or stay late consistently, which I'm definitely not personally able to commit to. I'm okay with timeboxing time to try investigating, but I don't think having everyone in a call together is feasible (or even necessarily productive). Additionally, we currently have two of the core team currently on leave - I don't think we can do anything this extreme until they are back.

---

My suggestion is to loop any known flaking tests, with a delay in between attempts. This is a quick fix that should hopefully remove the flakes without removing any test coverage.

samalba (approved):

aluzzardi (approved):
LGTM as this allows to quickly experiment to figure out what’s up

/cc @sipsma @jedevc

.dagger/test.go
```diff
@@ -36,36 +36,40 @@ func (t *Test) All(
...
+		WithExec([]string{"sh", "-c", "go test -list=. ./... | grep ^Test | sort"}).
+		Stdout(ctx)
 }
 
-// Run custom engine tests
-func (t *Test) Custom(
+// Run specific tests
+func (t *Test) Specific(
 	ctx context.Context,
+	// Only run test tests
```
aluzzardi, line 57: Typo?

sipsma (changes requested):
This is breaking testdev atm: https://github.com/dagger/dagger/actions/runs/10477444049/job/29018560013?pr=7863#step:3:1125

Otherwise the support for specifying tests to skip LGTM!

sipsma:
Also, on the topic of what to do to get all these flakes under control:
1. I 100% agree with @jedevc that those test failures may easily be legit bugs rather than test-only problems. Generally speaking, I think we should assume a flaky test indicates a genuine bug until proven otherwise; it quite often has turned out to be the case in the past and it would be sad to miss a bug because we just refactored a test to avoid it.
2. I also totally agree with @gerhard that we need to do something here. I think 2 weeks ago we were on a slow upward trend but we keep regressing every time we make progress. It's starting to feel like a rut, which is not good.
   * I don't have any specific suggestions yet other than just getting more hands on deck to
     1. deep dive flakes to root cause them
     2. work on the general tooling around debugging + performance analysis

.dagger/test.go
gerhard: Good spot!

gerhard:
@sipsma fixing that job failure was the first thing on my list today.

What makes the test flakes interesting is that I did not hit a single one in the 30-50 runs over the last few days. I am however able to get the `test all` test suite to fail consistently in this scenario: https://github.com/dagger/dagger/issues/8184#issuecomment-2299484907

I am still trying to narrow down the specific combination of factors which makes tests more likely to fail. While I have some clues, I will be continuing with capturing context in https://github.com/dagger/dagger/issues/8184 so that anyone is able to reproduce the same state, and then have it "frozen" so that we can continue digging deeper.

gerhard:
This now looks good to me, including the final log artefacts that we generate:

image

Worth noting:
1. `test-modules` succeeded in `13m 10s` (`12m
2. `test-everything-else` succeeded in `6m 47s`
3. `test` succeeded in `16m 22s` (it includes both `test-modules` & `test-everything-else`)
4. `testdev` succeeded in `15m 5s`

This is what the CI runner resources looked like while the above were running:

`test`
image

`test-modules`
image

`test-everything-else`
image

`testdev`
image

I am curious to see how the various test suites behave after a few runs.

---
FWIW @vito @crjm, after `5m`, I gave up on waiting for https://dagger.cloud/dagger/traces/9befdc22e1c85d1ebbbb9003f2ba8abd to load. I am using a public WiFi which maxes out at ~500KB/s. This is what that looked like:

https://github.com/user-attachments/assets/b3e01997-b124-4935-ba5d-604844077b8c

As a suggestion on how to make this experience better, it would have helped to have the "outer layer" load within a minute or two the most. If I clicked around and needed more data, grabbing that in follow-up requests would have been ideal. The thing which I missed the most was a loading progress indicator which would have told me how much data was downloaded, and how much is left to download. OK, it finished loading by the time I typed all of this up.

Issue #7851, opened 2024-07-09
💡 Suggesting using yarn as package manager with TS SDK


TomChv:
As the TS runtime is using yarn as package manager to download SDK & user's dependency, it's best to use `yarn` as package manager until we ship a way to configure the package manager inside the TS SDK runtime.

We could also benefit from this system to allow configure the node runtime.

What we could do is setting a field in the same way we can configure the runtime:

```json
{
  "dagger": {
    "packageManager": "npm@xxx | yarn@xxx | pnpm@xxx",
    "runtime": "bun@xxx | node@xxx", // Right now it's only `node | bun`, by support pinning a version we allow customizing the runtime version
  }
}
```

TomChv:
Should be closed by https://github.com/dagger/dagger/pull/7864

PR #7852, opened 2024-07-09
feat(ts-sdk): add runtime version configuration

TomChv:

PR #7864, opened 2024-07-10
feat(ts-sdk): detect & configure package manager

TomChv:
Following up on https://github.com/dagger/dagger/pull/7852, this PR uses the `packageManager` field of the `package.json` to use the correct package manager.
If not, it detect the used lockfile.

This PR also includes a change to generate a lockfile (yarn v1 by default) but it can also generate a lockfile for pnpm or npm with the right version.

⚠️ It seems yarn v4, and specially corepack  with yarn v4 isn't working well with Dagger, I'm writing tests for everything except yarn v4 and may handle this later.

**Example**

```json
{
  "packageManager": "yarn@1.22.22" // or "pnpm@x.x.x" or "npm@x.x.x"
}
```

Closes https://github.com/dagger/dagger/issues/7851


Xiot:
> ⚠️ It seems yarn v4, and specially corepack with yarn v4 isn't working well with Dagger, I'm writing tests for everything except yarn v4 and may handle this later.

What issue are you running into with yarn v4?
A potential issue that you will probably run into is that since the users module has a `@dagger.io/dagger` in `dependencies`, it will install them as well when `yarn install` (for example) is run.

TomChv:
> What issue are you running into with yarn v4?
A potential issue that you will probably run into is that since the users module has a @dagger.io/dagger in dependencies, it will install them as well when yarn install (for example) is run.

It doesn't resolve dependencies correctly and output that `there's no type Container` in the basic quickstart example.
I'm super confused about that and think it might be because of the worspace logic? Since it's not included in the package.json?

Xiot:
> It doesn't resolve dependencies correctly and output that `there's no type Container` in the basic quickstart example.
> I'm super confused about that and think it might be because of the worspace logic? Since it's not included in the package.json?

One of the issues is due to the fact that that you aren't using the `node_modules` linker.  To enable that you would need a `.yarnrc` file with

```yaml
nodeLinker: node-modules
```

Since the users module isn't using 'pnp', it will just check in the `node_modules` folder.

The 'workspace logic' should only really affect the case where you try to do a `yarn install` inside of the `sdk` directory from the users module.

(By 'users module', I'm referring to the 'dagger' folder and not the 'sdk' folder)

TomChv:
> One of the issues is due to the fact that that you aren't using the node_modules linker. To enable that you would need a .yarnrc file with

Okay, I'll take care of that in a follow up PR then, supporting yarn v4 is something different thant the package manager configuration

TomChv:
Update: on test failing left: `TestTypescriptWithOtherModuleTypes`

My guess is that by resolving the modules differently, it fails to provide the corresponding symbol of the dependencies, I'm trying to find a way to make it work.

TomChv:
Ok the bug should be fixed! There was an issue with the alias resolution.

Xiot:
> Okay, I'll take care of that in a follow up PR then, supporting yarn v4 is something different thant the package manager configuration

It's not really a `yarn-v4` thing.  When `install` (for any package manager) is run, it will install the dependencies for the `sdk` as well since its listed as a dependency.  Since the 'user' module doesn't have access to the lock file, these intermediate dependencies will also change.

The fact that the sdk is using the `pnp` linker also means that the dependencies will never match, regardless of which package manager is used

TomChv:
> It's not really a yarn-v4 thing. When install (for any package manager) is run, it will install the dependencies for the sdk as well since its listed as a dependency. Since the 'user' module doesn't have access to the lock file, these intermediate dependencies will also change.

> The fact that the sdk is using the pnp linker also means that the dependencies will never match, regardless of which package manager is used

That's strange it's only failing with yarn v4 then? Other package manager works fine

Xiot:
> That's strange it's only failing with yarn v4 then? Other package manager works fine

So `pnpm` isolates its dependencies, so that may not be an issue, but I'm pretty sure that `npm` does.  It may still work, but its not going to pick up the `yarn.lock` and apply them, so there is still the potential (depending on your version ranges in the sdk) that it will get run with different versions.

TomChv:
> So pnpm isolates its dependencies, so that may not be an issue, but I'm pretty sure that npm does. It may still work, but its not going to pick up the yarn.lock and apply them, so there is still the potential (depending on your version ranges in the sdk) that it will get run with different versions.

If you set up `npm`, it will generate & pick up the `package-lock.json` file, for yarn the yarn.lock etc...

However, it seems the fix I made on this PR also fixes the issue with yarn v4, it seems the compiler API does not resolve the symbols the same way depending of the package manager, that's pretty funny.

Xiot:
I know it will pick up the other lock file if it's at the root. Wasn't sure if it was going to pick it up from another folder. 

Yarn patches typescript I believe. Although I never looked into what it actually changed

TomChv:
> Yarn patches typescript I believe. Although I never looked into what it actually changed

Yeah, that's quite strange... but at least this PR will improve the DX :D

sdk/typescript/runtime/main.go
```diff
@@ -75,7 +86,12 @@ func (t *TypescriptSdk) ModuleRuntime(ctx context.Context, modSource *ModuleSour
...
 	detectedRuntime, _, err := t.detectRuntime(ctx, modSource, subPath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create module runtime: %w", err)
+		return nil, fmt.Errorf("failed to detect module runtime: %w", err)
+	}
+
+	packageManager, _, err := t.detectPackageManager(ctx, modSource, subPath, detectedRuntime)
+	if err != nil {
+		return nil, fmt.Errorf("failed to detect package manager: %w", err)
 	}
```
helderco, line 95: Looks like this is running twice because it's already in `CodegenBase`.

sdk/typescript/runtime/main.go
```diff
@@ -31,6 +31,17 @@ const (
...
+type SupportedPackageManager string
+
+const (
+	Yarn       SupportedPackageManager = "yarn"
+	Pnpm       SupportedPackageManager = "pnpm"
+	Npm        SupportedPackageManager = "npm"
+	BunManager SupportedPackageManager = "bun"
+)
+
+const YarnDefaultVersion = "1.22.22"
```
helderco, line 43: Why default to v1 when there's other more [modern versions](https://yarnpkg.com/migration/overview)?

sdk/typescript/runtime/main.go
```diff
@@ -340,17 +375,138 @@ func (t *TypescriptSdk) detectRuntime(ctx context.Context, modSource *ModuleSour
...
+
+	return Yarn, YarnDefaultVersion, nil
+}
+
+// generateLockFile generate a lock file for the matching package manager.
+func generateLockFile(ctx context.Context, ctr *Container, packageManager SupportedPackageManager, version string) (*Container, error) {
+	switch packageManager {
+	case Yarn:
+		// Sadly, yarn < v3 doesn't support generating a lockfile without installing the dependencies.
+		// So we use npm to generate the lockfile and then import it into yarn.
```
helderco, line 464: What if yarn `version` is >1?

sdk/typescript/runtime/main.go
```diff
@@ -195,12 +231,13 @@ func (t *TypescriptSdk) Base(runtime SupportedTSRuntime, version string) (*Conta
 			WithoutEntrypoint().
 			// Install default CA certificates and configure node to use them instead of its compiled in CA bundle.
 			// This enables use of custom CA certificates if configured in the dagger engine.
-			WithExec([]string{"apk", "add", "ca-certificates"}).
+			WithExec([]string{"apk", "add", "ca-certificates", "jq"}).
```
helderco, line 234: Where is `jq` needed?

sdk/typescript/runtime/main.go
```diff
@@ -340,17 +375,138 @@ func (t *TypescriptSdk) detectRuntime(ctx context.Context, modSource *ModuleSour
...
+			WithExec([]string{"corepack", "enable"}).
+			WithExec([]string{"corepack", "use", fmt.Sprintf("yarn@%s", version)}).
+			WithExec([]string{"yarn", "install", "--mode", "update-lockfile"}), nil
+	case Pnpm:
+		return ctr.
+			WithExec([]string{"npm", "install", "-g", fmt.Sprintf("pnpm@%s", version)}).
+			WithExec([]string{"pnpm", "install", "--lockfile-only"}), nil
+	case Npm:
+		return ctr.
+			WithExec([]string{"npm", "install", "-g", fmt.Sprintf("npm@%s", version)}).
```
helderco, line 475: npm installs npm? Does the newly installed version come first in the `$PATH`?

sdk/typescript/runtime/main.go
```diff
@@ -90,9 +106,13 @@ func (t *TypescriptSdk) ModuleRuntime(ctx context.Context, modSource *ModuleSour
 			WithExec([]string{"bun", "install", "--no-verify", "--no-progress"}).
 			WithEntrypoint([]string{"bun", entrypointPath}), nil
 	case Node:
+		// Install dependencies
+		ctr, err = installNodeDependencies(ctx, ctr, packageManager)
+		if err != nil {
+			return nil, fmt.Errorf("failed to install dependencies: %w", err)
+		}
```
helderco, line 113: Looks like `installNodeDependencies` supports `BunManager` so the dependencies install can be moved out of the switch.

sdk/typescript/runtime/main.go
```diff
@@ -340,17 +375,138 @@ func (t *TypescriptSdk) detectRuntime(ctx context.Context, modSource *ModuleSour
...
+// installDependencies installs the dependencies using the detected package manager.
+//
+// If the package.json file has a field "packageManager", it will use that to
+// determine the package manager to use. Otherwise, it will use the default
+// package manager based on the lock files present in the module.
+//
+// If none of the above works, it will use yarn.
+//
+// Except if the package.json has an invalid value in field "packageManager", this
+// // function should never return an error.
```
helderco, line 494: Looks like this is duplicated from `detectPackageManager`. Doesn't make sense here.

sdk/typescript/runtime/main.go
```diff
@@ -278,9 +315,7 @@ func (t *TypescriptSdk) installedSDK(ctr *Container, runtime SupportedTSRuntime)
 	case Node:
 		return ctr.
 			// Enable corepack so we can use yarn v4 which is supposed to be faster than npm or yarn v1.
-			WithExec([]string{"corepack", "enable"}).
-			WithExec([]string{"yarn", "set", "version", "stable"}).
-			WithExec([]string{"yarn", "workspaces", "focus", "--production"}).Directory(ModSourceDirPath)
+			WithExec([]string{"yarn", "install", "--production"}).Directory(ModSourceDirPath)
```
helderco, line 318: Shouldn't `installedSDK` use the same package manager used for the module's? Why is this needed anyway? Isn't the `./sdk` included in `package.json`? Meaning that installing the module's dependencies would also install the SDK. That goes for the lock as well. The lock in the module should have the transitive dependencies pinned, including the ones from the SDK. This is probably using the SDK's own lock file and `node_modules` which is wasteful.

sdk/typescript/runtime/main.go
helderco: Just noticed the comment in the description about Yarn v4. What about v3, have you tried it?

sdk/typescript/runtime/main.go
Xiot: I agree that the sdk shouldn't have its own lockfile.  Since it is listed as a dependency to the users module, it would ultimately be the one responsible for installing and pinning the versioned dependencies

helderco: To be clear, the SDK does need its own lock file but it's for developing the SDK and running CI tasks on it. Not to be used when installed as dependency like in modules. That lock can be excluded when bundling the SDK into the engine.

core/integration/module_typescript_test.go
```diff
@@ -641,6 +641,140 @@ func (ModuleSuite) TestTypescriptRuntimeDetection(ctx context.Context, t *testct
...
+			WithNewFile("package.json", `
+{
+  "dependencies": {
+    "typescript": "^5.3.2",
+    "@dagger.io/dagger": "./sdk"
+  },
+	"packageManager": "pnpm@8.15.4"
+}`).
+			With(daggerExec("develop"))
+
```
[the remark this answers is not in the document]
TomChv: I do that: https://github.com/dagger/dagger/blob/0aefd1c51bda1bbfd78fac36c56c9e2dc8fc284b/sdk/typescript/runtime/main.go#L266

core/integration/module_typescript_test.go
```diff
@@ -641,6 +641,140 @@ func (ModuleSuite) TestTypescriptRuntimeDetection(ctx context.Context, t *testct
...
+			With(daggerExec("init", "--name=Package-Detection", "--sdk=typescript", "--source=."))
+
+		files, err := modGen.Directory(".").Entries(ctx)
+		require.NoError(t, err)
+		require.Contains(t, files, "yarn.lock")
+
+		// Check that the package manager is set to yarn.
+		packageManager, err := modGen.File("package.json").Contents(ctx)
+		require.NoError(t, err)
+		require.Contains(t, packageManager, `"packageManager": "yarn@1.22.22`)
```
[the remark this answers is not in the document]
TomChv: By using corepack for yarn (because upper version of yarn [needs corepack](https://yarnpkg.com/corepack)), it automatically set the field so I'm not sure it's a good idea to remove it.
But the user is free to override it after that

sdk/typescript/runtime/main.go
```diff
@@ -31,6 +31,17 @@ const (
...
+	Pnpm       SupportedPackageManager = "pnpm"
+	Npm        SupportedPackageManager = "npm"
+	BunManager SupportedPackageManager = "bun"
+)
+
+const YarnDefaultVersion = "1.22.22"
+
 func New(
 	// +optional
 	sdkSourceDir *Directory,
```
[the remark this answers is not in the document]
TomChv: I'm not sure I understood this one

sdk/typescript/runtime/main.go
TomChv: v3 has the same issue yep

sdk/typescript/runtime/main.go
TomChv: then it's possible to generate the lockfile, but it would have the same result that this method so I prefer keeping it that way,it cover more cases

sdk/typescript/runtime/main.go
TomChv: I forget to remove it, was for a test

sdk/typescript/runtime/main.go
TomChv: Yes it does, I tried running a node image locally and installing npm

sdk/typescript/runtime/main.go
TomChv: True

sdk/typescript/runtime/main.go
TomChv: I also think it's possible to remove these lines but I saved that for a follow up, it's more an optimization than a DX change.

TomChv:
@helderco I did some improvement in the PR to improve the code readability and steps.

I now store module related config in `moduleConfig` so it's easier to reuse everywhere.
For the installSDK, I would prefer doing that in another follow up PR as I explained in my comment.

sdk/typescript/introspector/scanner/abtractions/typeToTypedef.ts
```diff
@@ -95,6 +95,16 @@ export function typeToTypedef(
         }
       }
 
+      // If we cannot resolve the symbol, we check for the alias symbol.
+      // This should mostly lead to a failure since external types are not supported by
+      // dagger yet.
+      if (type.aliasSymbol && type.aliasSymbol.flags & ts.TypeFlags.Object) {
```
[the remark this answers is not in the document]
helderco: You're assuming I know what those "symbols" are. 🙂 Can you give an example?

core/integration/module_typescript_test.go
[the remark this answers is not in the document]
helderco: What I mean is that the runtime module shouldn't generate a default config in `package.json`. Basically if it's not set then use the default, rather then writing it to the user's `package.json`.

sdk/typescript/runtime/main.go
helderco: You're assuming the result is the same, but ok.

sdk/typescript/runtime/main.go
helderco: > I also think it's possible to remove these lines but I saved that for a follow up, it's more an optimization than a DX change.

That doesn't make sense to me. Your PR is refactoring how dependencies are installed. The SDK seems to be installed separately, unnecessarily. Now is a good time to remove that.

core/integration/module_typescript_test.go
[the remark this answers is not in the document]
helderco: Ok, then what I'm saying is that instead of:
1. dagger init 
2. change package.json
3. dagger develop
4. dagger call

You do:
1. Add custom package.json
2. dagger init
3. dagger call

TomChv:
@helderco I need to ignore `node_modules`before merging the PR, otherwise it makes the TS SDK much slower

TomChv:
@helderco I finally found a way, it makes the SDK a bit faster, I like it a lot!

I'll also try to remove the duplication of install, we might not need it anymore.

drmikecrowe:
I just want to level-set on some issues with the current dagger 0.12.2 and this proposed PR:

- You must assume there's a `package.json` at the top-level directory (above the `dagger/` folder)
- This `package.json` _should_/_may_ have a `packageManager` field -- if this is provided, you must use this package manager or the install  breaks
- You have to ensure that the lockfile is NOT updated as part of this process.  That means you will likely need to run one of the following (from [antifu ni](https://github.com/antfu-collective/ni)):

```sh
# npm ci
# yarn install --frozen-lockfile (Yarn 1)
# yarn install --immutable (Yarn Berry)
# pnpm install --frozen-lockfile
# bun install --frozen-lockfile
``` 
 
For example, I removed the `packageManager` field and executed the following:

```ts
    return (
      dag
        .container()
        .from('node:18-slim')
        .withExec(['corepack', 'enable'])
        .withExec(['corepack', 'prepare', 'yarn@latest', '--activate'])
        .withDirectory('/src', source)
        .withWorkdir('/src')
        .withExec(['yarn', 'install', '--immutable'])
        .stdout()
    )
```

This failed with:

```
    ┃ ➤ YN0028: │ The lockfile would have been modified by this install, which is explicitly forbidden.
    ┃ ➤ YN0000: └ Completed in 0s 264ms
    ┃ ➤ YN0000: · Failed with errors in 1s 589ms
  ✘ Container.stdout: String! 2m47.3s
  ! process "yarn install --immutable" did not complete successfully: exit code: 1

Error: response from query: input: internalApps.build resolve: call function "build": process "tsx --no-deprecation --tsconfig /src/dagger/tsconfig.json /src/dagger/src/__dagger.entrypoint.ts" did not complete successfully: exit code: 1

Stderr:
Error: resolve: process "yarn install --immutable" did not complete successfully: exit code: 1
```

TomChv:
Thanks for your feedbacks!

> You must assume there's a package.json at the top-level directory (above the dagger/ folder)

The module init creates a `package.json` and it's pretty much required to get everything working, if there's no package.json, the SDK will provide one during the installation.

> This package.json should/may have a packageManager field -- if this is provided, you must use this package manager or the install breaks

Yes, if there's not packageManager field, it's going to use `yarn@1.22.22`

> You have to ensure that the lockfile is NOT updated as part of this process. That means you will likely need to run one of the following (from [antifu ni](https://github.com/antfu-collective/ni)):

Hmmm why is that required? I'm not sure this will be possible because it depends of which version of your package manager you use.
In your example your using yarn latest, so probably v4.xxx but if you made your setup with v1, the lockfile is completely different.

I can add this `--immutable field` but I'm not sure it will fix the issue thought, it might create more actually

sdk/typescript/introspector/scanner/abtractions/typeToTypedef.ts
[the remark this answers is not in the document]
TomChv: The symbol is the reference of the node in the AST, it refers to the metadata of the code defined. It's like the abstract representation of the code

core/integration/module_typescript_test.go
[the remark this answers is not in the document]
TomChv: corepack automatically do it so I'm not sure how I can do that.

sdk/typescript/runtime/main.go
TomChv: Yeah I'll do that!

TomChv: Done!

sdk/typescript/runtime/main.go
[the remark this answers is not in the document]
TomChv: Done

drmikecrowe:
> Hmmm why is that required? I'm not sure this will be possible because it depends of which version of your package manager you use.

Because we're going to be using this in our CI/CD pipeline, and that's the best practice.  What I'm including above (`.withExec(['yarn', 'install', '--immutable'])` will be how many projects will want to use dagger.

> Yes, if there's not packageManager field, it's going to use yarn@1.22.22

My 2c -- use `npm` and avoid the yarn version nonsense.  I don't quite understand how this works yet, but if you run yarn v1 in the top level folder, this will probably definitely conflict with the v4 lock file.  

It might be worth considering using `nodenv` and a separate node process for the SDK.  That way you fix the node version/package manager the SDK requires independently of the target node requirements

TomChv:
> Because we're going to be using this in our CI/CD pipeline, and that's the best practice. What I'm including above (.withExec(['yarn', 'install', '--immutable']) will be how many projects will want to use dagger.

But that means you'll  need to use the same yarn version as the runtime, so you need to configure your packageManager to be consistent. Which is not what you're doing in your snippet.

Could you try the same code but with the same yarn version?

> My 2c -- use npm and avoid the yarn version nonsense. I don't quite understand how this works yet, but if you run yarn v1 in the top level folder, this will probably definitely conflict with the v4 lock file.

Yes but npm is much slower, I could use pnpm but it's not adopted by everyone so I'm not sure what's the right decision.

@helderco any thought on this one?

drmikecrowe:
> But that means you'll need to use the same yarn version as the runtime, so you need to configure your packageManager to be consistent. Which is not what you're doing in your snippet.

Sorry, my snippet was illustrating the problem.  Remember, I have this in my `package.json`:

```json
  "packageManager": "yarn@4.1.1",
```

But, even using `4.1.1`, I get:

```
Container.withExec(args: ["corepack", "prepare", "yarn@4.1.1", "--activate"]): Container! 1.7s
...
    ┃ ➤ YN0028: │ The lockfile would have been modified by this install, which is explicitly forbidden.
    ┃ ➤ YN0000: └ Completed in 0s 278ms
    ┃ ➤ YN0000: · Failed with errors in 1s 552ms
  ✘ Container.stdout: String! 2m44.7s
  ! process "yarn install --immutable" did not complete successfully: exit code: 1
```
so the error still occurs.  I think this is due to the SDK's `yarn install` (v1 will modify the lock file), and then my `yarn install --immutable` will fail due to this

> Yes but npm is much slower, I could use pnpm but it's not adopted by everyone so I'm not sure what's the right decision.

Yeah, I know -- I'm a huge `pnpm` fan, but our company uses `yarn`.  but, `npm` _is the default_, so it might be the best choice  -- speed means nothing with the other issues it introduces

TomChv:
Found a new issue by not installing the SDK node_modules in the runtime, it seems the subpackage that sdk depends on are not added to the lockfile, I don't understand why thought but they are required to make it work.

I'm trying to find a way.

Xiot:
Does the sdk get installed separately from the module?  That could be causing these issues since the dependencies would potentially change when the module is added. 

Also since there isn't a `.yarnrc.yml` file in the sdk, it will use `PnP` instead of the node modules linker which would cause other issues.

TomChv:
> Also since there isn't a .yarnrc.yml file in the sdk, it will use PnP instead of the node modules linker which would cause other issues.

Yes, I'm currently trying to change the node linker but it's the same problem with yarn v4

> The sdk needs to be treated like any other npm dependency

Yes but the same issue is that it's a local dependencies, so the dependencies of this deps needs to be installed at the root node_modules.
I found a fix with pnpm, it works with yarn v1 & npm but not with yarn v4, I'm trying to see if I can fix that with a workspaces or something

TomChv:
I added a commit with theses changes, it works locally, I'm currently checking if it passes on the CI

TomChv:
Only 1 test failing, I can reproduce locally with dagger call but not with a manual test, this is weird

https://dagger.cloud/dagger/traces/9144e633456d21a1f0ab7ce11bb0a7cd?span=a29638ceb3992471#a49f16c55743f48b

TomChv:
Ok I normally fixed the tests, it seems it was related to the typescript version installed with that specific bun version, I fixed it by adding the deps to the package.json

TomChv:
Ci is green, ready for a last review before we merge :D

drmikecrowe:
Hey, can somebody show me how to test this locally before it's merged?  I have pulled his branch down, I'm confused about how I can reference it

TomChv:
> Hey, can somebody show me how to test this locally before it's merged? I have pulled his branch down, I'm confused about how I can reference it

Hey sure, you can go to the branch then follow these steps:

```
# From the root of the dagger repo
./hack/dev

# Go to your project and put this in your .envrc
DAGGER_SRC_ROOT=<path to the dagger repro>/dagger/dagger
export _EXPERIMENTAL_DAGGER_CLI_BIN=$DAGGER_SRC_ROOT/bin/dagger
export _EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-container://dagger-engine.dev
export PATH=$DAGGER_SRC_ROOT/bin:$PATH

# Verify your CLI with dagger version
dagger version
dagger dev-66331ea9a8772a0d0fced5cbceb6e4f26f454b2c (registry.dagger.io/engine) darwin/arm64

# And verify your dev engine is up
docker ps
e7faf74a16df   localhost/dagger-engine.dev:latest   "dagger-entrypoint.s…"   18 hours ago   Up 18 hours             dagger-engine.dev

# Now you can use dagger with the branch version.
```

drmikecrowe:
OK, I think this is looking better, but I need help identifying if this is a local test issue or something more:

```log
✔ connect 0.2s
✔ ModuleSource.withName(name: "internal-apps"): ModuleSource! 0.0s
✔ ModuleSource.withSDK(sdk: "typescript"): ModuleSource! 0.0s
✔ ModuleSource.withSourceSubpath(path: "."): ModuleSource! 0.0s
✔ ModuleSource.resolveFromCaller: ModuleSource! 1m7.5s
✘ ModuleSource.asModule(engineVersion: "latest"): Module! 2m43.6s
! failed to create module: select: failed to update codegen and runtime: failed to generate code: failed to call sdk module codegen: select: call function "Codegen": process "/runtime" did not complete successfully: exit code: 2
Error: failed to generate code: input: moduleSource.withContextDirectory.withName.withSDK.withSourceSubpath.asModule resolve: failed to create module: select: failed to update codegen and runtime: failed to generate code: failed to call sdk module codegen: select: call function "Codegen": process "/runtime" did not complete successfully: exit code: 2

Stdout:
marshal: json: error calling MarshalJSON for type *dagger.GeneratedCode: input: container.from.withoutEntrypoint.withExec.withEnvVariable.withMountedCache.withMountedCache.withMountedCache.withExec.withMountedDirectory.withMountedDirectory.withWorkdir.withExec.withExec.withDirectory.withDirectory.withExec.withExec.withExec.withExec.file resolve: process "corepack use yarn@4.1.1" did not complete successfully: exit code: 1

Stdout:
Installing yarn@4.1.1 in the project...

➤ YN0000: · Yarn 4.1.1
➤ YN0000: ┌ Resolution step
➤ YN0085: │ + @dagger.io/dagger@file:./sdk#./sdk::hash=06fb56&locator=internal-apps%40workspace%3A.
➤ YN0085: │ - @dagger.io/dagger@file:./sdk#./sdk::hash=e3430a&locator=dagger-bc5392%40workspace%3Adagger, dagger-bc5392@workspace:dagger
➤ YN0000: └ Completed in 0s 765ms
➤ YN0000: ┌ Post-resolution validation
➤ YN0002: │ @repo/pinnacle-layers@workspace:packages/pinnacle-layer doesn't provide eslint (p5fa97), requested by @typescript-eslint/parser.
➤ YN0002: │ skills@workspace:apps/skills doesn't provide esbuild (p15d2f), requested by serverless-esbuild.
➤ YN0086: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code.
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0066: │ typescript@patch:typescript@npm%3A5.5.2#optional!builtin<compat/typescript>::version=5.5.2&hash=5adc0c: Cannot apply hunk #1
➤ YN0066: │ typescript@patch:typescript@npm%3A5.5.3#optional!builtin<compat/typescript>::version=5.5.3&hash=5adc0c: Cannot apply hunk #1
➤ YN0013: │ 2605 packages were added to the project (+ 1.02 GiB).
➤ YN0000: └ Completed in 24s 233ms
➤ YN0000: ┌ Link step
➤ YN0005: │ vue-material@npm:1.0.0 lists build scripts, but its build has been explicitly disabled through configuration.
➤ YN0007: │ protobufjs@npm:7.3.2 must be built because it never has been before or the last one failed
➤ YN0007: │ svelte-preprocess@npm:5.1.4 [2e35e] must be built because it never has been before or the last one failed
➤ YN0007: │ esbuild@npm:0.21.5 must be built because it never has been before or the last one failed
➤ YN0007: │ aws-sdk@npm:2.1646.0 must be built because it never has been before or the last one failed
➤ YN0007: │ sqlite3@npm:5.1.7 [20c11] must be built because it never has been before or the last one failed
➤ YN0007: │ esbuild@npm:0.19.2 must be built because it never has been before or the last one failed
➤ YN0007: │ es5-ext@npm:0.10.64 must be built because it never has been before or the last one failed
➤ YN0007: │ @sentry/cli@npm:2.32.1 must be built because it never has been before or the last one failed
➤ YN0007: │ better-sqlite3@npm:8.7.0 must be built because it never has been before or the last one failed
➤ YN0007: │ core-js@npm:3.37.1 must be built because it never has been before or the last one failed
➤ YN0007: │ vue-demi@npm:0.14.5 [d85ca] must be built because it never has been before or the last one failed
➤ YN0007: │ esbuild@npm:0.17.19 must be built because it never has been before or the last one failed
➤ YN0007: │ yorkie@npm:2.0.0 must be built because it never has been before or the last one failed
➤ YN0007: │ @fortawesome/fontawesome-common-types@npm:6.5.2 must be built because it never has been before or the last one failed
➤ YN0007: │ cypress@npm:13.12.0 must be built because it never has been before or the last one failed
➤ YN0007: │ vue-demi@npm:0.14.5 [499a1] must be built because it never has been before or the last one failed
➤ YN0007: │ vue-demi@npm:0.14.5 [90d6b] must be built because it never has been before or the last one failed
➤ YN0007: │ @sveltejs/kit@npm:2.5.17 [13444] must be built because it never has been before or the last one failed
➤ YN0007: │ serverless@npm:3.39.0 must be built because it never has been before or the last one failed
➤ YN0007: │ @fortawesome/fontawesome-svg-core@npm:6.5.2 must be built because it never has been before or the last one failed
➤ YN0007: │ @fortawesome/free-solid-svg-icons@npm:6.5.2 must be built because it never has been before or the last one failed
➤ YN0009: │ @sveltejs/kit@npm:2.5.17 [13444] couldn't be built successfully (exit code 1, logs can be found here: /tmp/xfs-16527659/build.log)
➤ YN0000: └ Completed in 40s 119ms
➤ YN0000: · Failed with errors in 1m 6s

internal-apps  dagger @ 4c5430a ⩮✘∽+ 3m52s ❄️ devenv-shell-env❯ dagger version
dagger dev-fb9b6428d0f4be165359994e87fc0e8b15326811 (registry.dagger.io/engine) linux/amd64

internal-apps  dagger @ 4c5430a ⩮✘∽+ ❄️ devenv-shell-env❯ cat .yarnrc.yml
enableMirror: true

nodeLinker: pnpm

pnpMode: loose

globalFolder: ./.yarn/berry

```

TomChv:
> OK, I think this is looking better, but I need help identifying if this is a local test issue or something more:

Hey, I think that's strange `corepack use` failed, doesn't seem related ot the runtime itself but maybe I do something wrong during the setup, could you try in a simple node image to see if it works?

Try with  `node:lts` if you can please 

Also I think the `nodeLinker` will need to be set to `node_modules` to work with the SDK, but let's first identify the corepack issue.

drmikecrowe:
> Try with `node:lts` if you can please

Not sure how to do that, given Im' doing `dagger init --sdk=typescript`.  How would I specify this?

> Also I think the `nodeLinker` will need to be set to `node_modules` to work with the SDK, but let's first identify the corepack issue.

I hope not -- I've migrated away from that for cache reasons.  By including `globalFolder: **./.yarn/berry**`, that brings the cache inside the container to allow the linking to work (I hope)

TomChv:
> Not sure how to do that, given Im' doing dagger init --sdk=typescript. How would I specify this?

I meant, add your project to a docker container and try corepack use, just to test

```
docker run -it -v $PWD:. node:lts /bin/sh

corepack use yarn@v4.1.1

Does it fails?
```

drmikecrowe:
jeez, sorry, I should have thought of that.  However, works fine:

```log
internal-apps  dagger @ 4c5430a ⩮✘∽+ ❄️ devenv-shell-env❯ docker run -it -v $PWD:/src node:lts /bin/bash
root@7d451bb044f6:/# cd /src
root@7d451bb044f6:/src# corepack enable
root@7d451bb044f6:/src# corepack use yarn@4.1.1
Installing yarn@4.1.1 in the project...

➤ YN0065: Yarn will periodically gather anonymous telemetry: https://yarnpkg.com/advanced/telemetry
➤ YN0065: Run yarn config set --home enableTelemetry 0 to disable

➤ YN0000: · Yarn 4.1.1
➤ YN0000: ┌ Resolution step
➤ YN0085: │ - @dagger.io/dagger@file:./sdk#./sdk::hash=e3430a&locator=dagger-bc5392%40workspace%3Adagger, @graphql-typed-document-node/core@npm:3.2.0, @grpc/grpc-js@npm:1.11.1, @grpc/proto-loader@npm:0.7.13, @isaacs/fs-minipass@npm:4.0.1, @js-sdsl/ordered-map@npm:4.4.2, @lifeomic/axios-fetch@npm:3.1.0, @opentelemetry/api-logs@npm:0.52.1, @opentelemetry/api@npm:1.9.0, and 61 more.
➤ YN0000: └ Completed in 0s 817ms
➤ YN0000: ┌ Post-resolution validation
➤ YN0002: │ @repo/pinnacle-layers@workspace:packages/pinnacle-layer doesn't provide eslint (p5fa97), requested by @typescript-eslint/parser.
➤ YN0002: │ skills@workspace:apps/skills doesn't provide esbuild (p15d2f), requested by serverless-esbuild.
➤ YN0086: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code.
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0066: │ typescript@patch:typescript@npm%3A5.5.2#optional!builtin<compat/typescript>::version=5.5.2&hash=5adc0c: Cannot apply hunk #1
➤ YN0013: │ 2536 packages were added to the project (+ 974.91 MiB).
➤ YN0000: └ Completed in 18s 587ms
➤ YN0000: ┌ Link step
➤ YN0005: │ vue-material@npm:1.0.0 lists build scripts, but its build has been explicitly disabled through configuration.
➤ YN0007: │ svelte-preprocess@npm:5.1.4 [2e35e] must be built because it never has been before or the last one failed
➤ YN0007: │ esbuild@npm:0.21.5 must be built because it never has been before or the last one failed
➤ YN0007: │ aws-sdk@npm:2.1646.0 must be built because it never has been before or the last one failed
➤ YN0007: │ sqlite3@npm:5.1.7 [20c11] must be built because it never has been before or the last one failed
➤ YN0007: │ esbuild@npm:0.19.2 must be built because it never has been before or the last one failed
➤ YN0007: │ es5-ext@npm:0.10.64 must be built because it never has been before or the last one failed
➤ YN0007: │ @sentry/cli@npm:2.32.1 must be built because it never has been before or the last one failed
➤ YN0007: │ better-sqlite3@npm:8.7.0 must be built because it never has been before or the last one failed
➤ YN0007: │ core-js@npm:3.37.1 must be built because it never has been before or the last one failed
➤ YN0007: │ vue-demi@npm:0.14.5 [d85ca] must be built because it never has been before or the last one failed
➤ YN0007: │ esbuild@npm:0.17.19 must be built because it never has been before or the last one failed
➤ YN0007: │ yorkie@npm:2.0.0 must be built because it never has been before or the last one failed
➤ YN0007: │ @fortawesome/fontawesome-common-types@npm:6.5.2 must be built because it never has been before or the last one failed
➤ YN0007: │ cypress@npm:13.12.0 must be built because it never has been before or the last one failed
➤ YN0007: │ vue-demi@npm:0.14.5 [499a1] must be built because it never has been before or the last one failed
➤ YN0007: │ vue-demi@npm:0.14.5 [90d6b] must be built because it never has been before or the last one failed
➤ YN0007: │ @sveltejs/kit@npm:2.5.17 [13444] must be built because it never has been before or the last one failed
➤ YN0007: │ serverless@npm:3.39.0 must be built because it never has been before or the last one failed
➤ YN0007: │ @fortawesome/fontawesome-svg-core@npm:6.5.2 must be built because it never has been before or the last one failed
➤ YN0007: │ @fortawesome/free-solid-svg-icons@npm:6.5.2 must be built because it never has been before or the last one failed
➤ YN0000: └ Completed in 37s 281ms
➤ YN0000: · Done with warnings in 57s 192ms
root@7d451bb044f6:/src# echo $?
0
root@7d451bb044f6:/src#


```

TomChv:
> jeez, sorry, I should have thought of that. However, works fine:

Uhhhg then it's quite strange, the steps are basically the same in the runtime, I'm quite confused :/

helderco (approved):
Preemptively approving, but expecting the comments to be resolved before merging.

core/integration/module_typescript_test.go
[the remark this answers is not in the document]
helderco: I thought it was a dagger specific field, under `"dagger"`, like the runtime setting.

drmikecrowe:
> The `nodeLinker`, `globalFolder`, and `cacheFolder` are slightly different things, so you should be able to still pull in the `globalFolder` inside of the repo path and still change what `nodeLinker` you are using

True, but here's what I found:
* When you use `nodeLinker: pnpm`, it uses symlinks and saves a ton of space (ideal for dagger caching IMO)
* However, the cache folder needs to be in the repo by default (hence the need for `globalFolder: **./.yarn/berry**`)

Without this, I got some symlink outside of filesystem or something in my code when trying to compile

drmikecrowe:
Stepping back, here's a question:  _Should_ we recommend that `dagger.json` and `dagger/` be **above** the target folder? Most of my comments/critiques/objections to this implementation have been because I didn't want to move all my folders down a level in my monorepo to support dagger.  However, I wonder if it might not be a case of `caveat emptor` if you choose to use dagger directly inside your target folder like I did.

In theory, we would need to ensure these permutations worked:

| Pkg Mgr | Version | Source | Permutation | 
| --- | --- | --- | --- |
| all | all | below | n/a |
| npm | all | same level |   |
| pnpm | all | same level |   |
| pnpm | all | same level | packageManager specified |
| yarn | 1 | same level |   |
| yarn | 1 | same level | packageManager specified |
| yarn | 2+ | same level |   |
| yarn | 2+ | same level | packageManager specified |
| yarn | 2+ | same level | nodeLinker: pnp |
| yarn | 2+ | same level | nodeLinker: node_modules |
| yarn | 2+ | same level | nodeLinker: pnpm |

**NOTE**: Not sure how `globalLinker` might affect the yarn testing

TomChv:
> Stepping back, here's a question: Should we recommend that dagger.json and dagger/ be above the target folder? Most of my comments/critiques/objections to this implementation have been because I didn't want to move all my folders down a level in my monorepo to support dagger. However, I wonder if it might not be a case of caveat emptor if you choose to use dagger directly inside your target folder like I did.

In my your dagger.json should be at the root of your repo but the source in a sub directory so you do not mixup your dependencies.

For example

```shell
.
.git
dagger.json // -> source ./ci
ci/
--- package.json
--- package-lock.json
--- tsconfig
--- src/index.ts
package.json // your project package.json
xx... rest of your repo
```

drmikecrowe:
> In my your dagger.json should be at the root of your repo but the source in a sub directory so you do not mixup your dependencies.

LOL, why would a new user (i.e. me) assume that?  All of my gitlab configuration is at the top of the repo:

```
.gitlab_ci.yaml
apps/
  reviews/
  skills/
  stocks/
  tnb/
docker/
node_modules/
packages/
docker-compose.yml
Dockerfile
README.md
package.json
tsconfig.json
turbo.json
yarn.lock
```

If I'm switching to dagger, nothing explicit in the docs says that.  
`dagger init` happily installs alongside my app (and did, leading me to be involved with this PR due to all the issues I was having)
I just spent a few minutes in the docs trying to find this as a requirement, but it isn't specified. 
The only thing I can see in the docs is this: The quickstart _repo_ is formatted correctly, but nowhere do you say "Install dagger a layer above your TypeScript project"

For reference, I did this (and I've been doing ci/cd for a while):
* Came to dagger.io and perused the docs (guilty for assuming)
* Went to the quickstart and followed the quickstart *in MY repo*
* Churned

I have multiple thoughts:

* Maybe `dagger init` should throw an error if `package.json` is in the current folder?
* Will this limitation apply to other platforms?  What happens with a `dagger init` at the top level of a go project or python project?  Are there conflicts there?

@jpadams -- Food for thought based on our call and new user experiences

sdk/typescript/runtime/main.go
```diff
@@ -339,18 +363,211 @@ func (t *TypescriptSdk) detectRuntime(ctx context.Context, modSource *ModuleSour
...
+func (t *TypescriptSdk) generateLockFile(ctr *dagger.Container) (*dagger.Container, error) {
+	packageManager := t.moduleConfig.packageManager
+	version := t.moduleConfig.packageManagerVersion
+
+	switch packageManager {
+	case Yarn:
+		// Enable corepack
+		ctr = ctr.
+			WithExec([]string{"corepack", "enable"}).
+			WithExec([]string{"corepack", "use", fmt.Sprintf("yarn@%s", version)})
```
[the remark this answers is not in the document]
TomChv: The [standard](https://nodejs.org/api/packages.html#packagemanager) said that `packageManager` should have `<packageManager>@<version>`, so version isn't supposed to be empty.

In fact, I should catch that and throw an error in that case, because if you say `packageManager: yarn`, depending on the version you can have so many different result (between v1, 2, 3 and 4)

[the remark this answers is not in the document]
TomChv: I think I can remove them, it's really different than go.mod, there's not such workspace logic for node in the dagger repo, so these are "useless" configs.

sdk/typescript/runtime/main.go
```diff
@@ -257,38 +294,24 @@ func (t *TypescriptSdk) setupModule(ctx context.Context, ctr *Container, runtime
...
-	ctr = ctr.
-		WithWorkdir(ModSourceDirPath).
-		WithDirectory(".", t.SDKSourceDir, ContainerWithDirectoryOpts{
+// addSDK returns a directory with the SDK sources.
+func (t *TypescriptSdk) addSDK() *dagger.Directory {
+	return dag.
+		Directory().
+		WithDirectory("/", t.SDKSourceDir, dagger.DirectoryWithDirectoryOpts{
 			Exclude: []string{"codegen", "runtime"},
 		})
```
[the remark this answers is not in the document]
TomChv: True, nice catch, I'll try that one

Xiot:
My case is slightly different than @drmikecrowe , since I have all of my dagger code in a separate repository than my application code.  My concerns are centered around the dependencies of the sdk and the rest of my dagger module.  (I'm using the node_modules linker for yarn)

> In my your dagger.json should be at the root of your repo but the source in a sub directory so you do not mixup your dependencies.

How would this fix 'mixing up your dependencies'?  Since dagger is referenced as a dependency, calling `yarn install` will install the dependencies of dagger itself and use the lockfile of the root of the repo (unless the dagger module is not part of the workspace and has its own yarn.lock. In that case it will use the local lock file)

> Maybe dagger init should throw an error if package.json is in the current folder?

I think what we need is a clear explanation of what you think the folder structure should be for both when dagger is in a monorepo and when its in its own repo.

drmikecrowe:
> In my your dagger.json should be at the root of your repo but the source in a sub directory so you do not mixup your dependencies.

I now believe this must be our requirement:

```
.
.git
dagger.json // -> source ./ci
ci/
--- package.json
--- package-lock.json
--- tsconfig
--- src/index.ts
project/ (or source)
--- package.json // your project package.json
--- xx... rest of your repo
```

In short, I think dagger's `package.json` and the project's `package.json` must be in parallel folders.  I don't see how they can be parent/children with any monorepo setup

PR #7870, opened 2024-07-10
cli: improve UX to show cloud URL

marcosnils:
this commit introduces the following changes:

- Keystroke for web visualization is now `w` for both logged-in and
  logged-out clients.
- New `--web` global flag visualize trace in web browser.
- Trace URL display:
	- logged in: URL gets printed at the top honoring -q and -s
	  flags
	- logged out: Same as logged in with the ability to disable the
	  message by setting an environment variable.

Screenshots below:

<details>

<summary>Logged out</summary>

- dagger command with TUI shows `w` action
image

-  dagger command with TUI message lingers at the top
image

- dagger command `-q` with TUI  message is not printed
image

- dagger command `--progress plain` message is printed at the top
image

- setting env var skips logged out message
image

</details>

<details>

<summary>Logged in</summary>

-  dagger command with TUI message lingers at the top
image

- dagger command `-q` with TUI  message is not printed
image

- dagger command `--progress plain` message is printed at the top
image

</details>

### DIFF
--- a/cmd/dagger/cloud.go
+++ b/cmd/dagger/cloud.go
@@ -116,7 +116,7 @@ func (cli *CloudCLI) Login(cmd *cobra.Command, args []string) error {
 }
 
 func createNewOrg(ctx context.Context, cli *cloud.Client, w io.Writer) (*auth.Org, error) {
-	url := "https://dagger.cloud/signup?quickstart=true"
+	url := "https://dagger.cloud/traces/setup"
 	err := browser.OpenURL(url)
 	if err != nil {
 		fmt.Fprintf(w, "Unable to open browser automatically, please visit %s to create an organization.\n", url)
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"math/rand"
 	"os"
 	"strings"
 	"time"
@@ -24,6 +25,12 @@ import (
 	sdklog "go.opentelemetry.io/otel/sdk/log"
 )
 
+// having a bit of fun with these. cc @vito @jedevc
+var skipLoggedOutTraceMsgEnvs = []string{"NOTHANKS", "SHUTUP", "GOAWAY", "STOPIT"}
+
+var loggedOutTraceMsg = fmt.Sprintf("Setup full trace at %%s.\nTo turn this off, export %s=1",
+	skipLoggedOutTraceMsgEnvs[rand.Intn(len(skipLoggedOutTraceMsgEnvs))])
+
 type FrontendOpts struct {
 	// Debug tells the frontend to show everything and do one big final render.
 	Debug bool
@@ -39,6 +46,9 @@ type FrontendOpts struct {
 
 	// Remove completed things after this duration. (default 1s)
 	GCThreshold time.Duration
+
+	// Open web browser with the trace URL as soon as pipeline starts.
+	OpenWeb bool
 }
 
 type Frontend interface {
@@ -57,8 +67,8 @@ type Frontend interface {
 
 	// ConnectedToEngine is called when the CLI connects to an engine.
 	ConnectedToEngine(ctx context.Context, name string, version string, clientID string)
-	// ConnectedToCloud is called when the CLI has started emitting events to The Cloud.
-	ConnectedToCloud(ctx context.Context, url string, msg string)
+	// SetCloudURL is called after the CLI checks auth and sets the cloud URL.
+	SetCloudURL(ctx context.Context, url string, msg string, logged bool)
 }
 
 type Dump struct {
@@ -473,4 +483,13 @@ func renderPrimaryOutput(db *DB) error {
 		fmt.Fprintln(os.Stdout)
 	}
 	return nil
+}
+
+func skipLoggedOutTraceMsg() bool {
+	for _, env := range skipLoggedOutTraceMsgEnvs {
+		if os.Getenv(env) != "" {
+			return true
+		}
+	}
+	return false
 }
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -29,7 +29,8 @@ import (
 var skipLoggedOutTraceMsgEnvs = []string{"NOTHANKS", "SHUTUP", "GOAWAY", "STOPIT"}
 
 //nolint:gosec
-var loggedOutTraceMsg = fmt.Sprintf("Setup full trace at %%s.\nTo turn this off, export %s=1",
+// Keep this to one line, and 80 characters max (longest env var name is NOTHANKS)
+var loggedOutTraceMsg = fmt.Sprintf("Setup tracing at %%s. To hide: export %s=1",
 	skipLoggedOutTraceMsgEnvs[rand.Intn(len(skipLoggedOutTraceMsgEnvs))])
 
 type FrontendOpts struct {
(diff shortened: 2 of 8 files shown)


engine/client/client.go
```diff
@@ -322,9 +322,11 @@ func (c *Client) startEngine(ctx context.Context) (rerr error) {
 	if c.EngineCallback != nil {
 		c.EngineCallback(ctx, bkInfo.BuildkitVersion.Revision, bkInfo.BuildkitVersion.Version, c.ID)
 	}
-	if c.CloudCallback != nil {
+	if c.CloudURLCallback != nil {
 		if url, msg, ok := enginetel.URLForTrace(ctx); ok {
-			c.CloudCallback(ctx, url, msg)
+			c.CloudURLCallback(ctx, url, msg, ok)
+		} else {
+			c.CloudURLCallback(ctx, "https://dagger.cloud/traces/setup", "", ok)
```
vito, line 329: nit: I'm 50/50 on this, would have expected that we don't call the callback and just handle it being unset when you press the hotkey. Seems like a possible footgun later to have a URL that can mean two different things, e.g. phrasing around it might be like "View your trace in Cloud: <signup url>"

But on the other hand maybe it's nice to distinguish between "I don't know if I'm logged in" vs. "I know I am, here's the trace URL" vs. "I know I'm not, here's the signup URL."

:man_shrugging:

dagql/idtui/frontend_pretty.go
```diff
@@ -53,15 +53,19 @@ type frontendPretty struct {
...
+	window     tea.WindowSizeMsg // set by BubbleTea
+	view       *strings.Builder  // rendered async
+	viewOut    *termenv.Output
+	browserBuf *strings.Builder // logs if browser fails
 
 	// held to synchronize tea.Model with updates
 	mu sync.Mutex
+
+	// messages to print before the final render
+	msgPreFinalRender strings.Builder
```
vito, line 68: nit: Seems simpler to just save a `loggedIn bool` alongside `cloudURL` and use that to conditionally print a message in `finalRender`, rather than keep an extra buffer

cmd/dagger/cloud.go
```diff
@@ -116,7 +116,7 @@ func (cli *CloudCLI) Login(cmd *cobra.Command, args []string) error {
 }
 
 func createNewOrg(ctx context.Context, cli *cloud.Client, w io.Writer) (*auth.Org, error) {
-	url := "https://dagger.cloud/signup?quickstart=true"
+	url := "https://dagger.cloud/traces/setup"
```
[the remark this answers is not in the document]
marcosnils: Not yet, but it will exist before tomorrow's release

dagql/idtui/frontend_pretty.go
marcosnils: > nit: Seems simpler to just save a `loggedIn bool` alongside `cloudURL` and use that to conditionally print a message in `finalRender`, rather than keep an extra buffer

I thought about it but I liked the fact that I could set the message in the `SetCloudURL` function and keep everything scoped there instead of having the logic split between two functions. Happy to add the isLogged flag is that seems like a better approach :pray:

engine/client/client.go
marcosnils: > But on the other hand maybe it's nice to distinguish between "I don't know if I'm logged in" vs. "I know I am, here's the trace URL" vs. "I know I'm not, here's the signup URL.

That's exactly the same feeling I had while doing it. We'll always have a URL now regardless is the user is logged in or not, that's why I also renamed the function to `SetCloudURL` instead of the former `ConnectedToCloud` name which was only used in "logged in" cases.

dagql/idtui/frontend_pretty.go
vito: Ah I see the idea - sort of abstracting a 'pre' region for the final paint so that other places can write to it and keep all their concerns local. The cost of entry is still a code change in two spots, so the benefit comes in when multiple things need to populate it, but the issue I see there is we might actually care about the order that those separate pieces are rendered, which you have less control over it's all decoupled.

So in the end I think it's best to just push state into the model and let the render phase handle everything. TUI rendering is super fiddly so IME it's easier to treat it as one big monolith and pray it doesn't get too big. :sweat_smile: 

But yeah, really splitting hairs, if you agree feel free to make the change but no biggie either way

vito (approved):

dagql/idtui/frontend.go
```diff
@@ -24,6 +25,13 @@ import (
 	sdklog "go.opentelemetry.io/otel/sdk/log"
 )
 
+// having a bit of fun with these. cc @vito @jedevc
+var skipLoggedOutTraceMsgEnvs = []string{"NOTHANKS", "SHUTUP", "GOAWAY", "STOPIT"}
```
shykes, line 29: Love that you actually did this 😁❤️

Just making sure that we honor all of them all the time right? Otherwise our joke won't be nearly as funny...

dagql/idtui/frontend.go
```diff
@@ -24,6 +25,13 @@ import (
...
+
+//nolint:gosec
+var loggedOutTraceMsg = fmt.Sprintf("Setup full trace at %%s.\nTo turn this off, export %s=1",
```
shykes, line 32:
```suggestion
var loggedOutTraceMsg = fmt.Sprintf("Setup full trace at %%s. To hide this: export %s=1",
```

Shorten + keep on one line

shykes (changes requested):
Looks awesome! One small nit to keep the logged out nag on a single line.

dagql/idtui/frontend.go
jedevc: Yup, yup. We should definitely keep this to one line :tada:

However, even with this suggestion, it's *still* 87 characters (with the longest `NOTHANKS`), so can overflow a single standard terminal line.

We could trim it further, and remove ` this` (gets us to 82), and potentially change `Setup` to `See` (which brings us down to 80 perfectly).

```go
var loggedOutTraceMsg = fmt.Sprintf("See full trace at %%s. To hide: export %s=1"
```

dagql/idtui/frontend_pretty.go
```diff
@@ -205,6 +225,11 @@ func (fe *frontendPretty) finalRender() error {
 	fe.recalculateViewLocked()
 
 	if fe.Debug || fe.Verbosity >= ShowCompletedVerbosity || fe.err != nil {
+		var msg string
+		if fe.msgPreFinalRender.Len() > 0 {
+			fmt.Fprintln(os.Stderr, fe.msgPreFinalRender.String())
+		}
+		fmt.Fprintln(os.Stderr, msg)
```
jedevc, line 232: nit: this results in an extra line at the start of output if `NOTHANKS=1`:

image

I think we can move this into the condition before?

dagql/idtui/frontend.go
shykes: @jedevc wouldn't "see full trace" feel slightly inaccurate since it sends you to a setup, and then technically you won't see *that* particular trace?

- As an alternative: "Setup tracing at https://dagger.cloud/traces/setup. To hide, export NOTHANKS=1" (78)

shykes: (I updated my proposal based on this)

dagql/idtui/frontend.go
marcosnils: yep, we do

jedevc (approved):

shykes (approved):

dagql/idtui/frontend_pretty.go
marcosnils: hey @vito did not have the time to refactor this for the release. Will change in a follow-up PR :pray:

PR #7872, opened 2024-07-10
pipelines: streamline codegen

shykes:
## Problem

- The source code of Dagger includes some Dagger pipelines
- Therefore our CI/CD pipelines sometimes process the source code of Dagger Modules - for example to lint them
- This processing sometimes requires invoking Dagger codegen against those modules - the equivalent of `dagger develop`
- So far our pipelines have done this on a case-by-case basis, but that makes the source code harder to read

## Solution

- all access to the Dagger source code from any part of the pipeline, is gated by `DaggerDev.source()` which ~~*always*~~ optionally applies codegen on all dagger modules.
- The list of modules is manually maintained for now, but should be made dynamic in a follow-up improvement
- ~~This can probably be optimized (not all pipelines that access the dagger source code actually need codegen, but they all trigger it). But for now I propose we pay the slight performance tax (which gets absorbed by caching), for the sake of much-needed simplification. Then we can gradually optimize.~~

### DIFF
--- a/dev/cli.go
+++ b/dev/cli.go
@@ -21,7 +21,7 @@ func (cli *CLI) File(
 	// +optional
 	platform dagger.Platform,
 ) (*dagger.File, error) {
-	builder, err := build.NewBuilder(ctx, cli.Dagger.Source)
+	builder, err := build.NewBuilder(ctx, cli.Dagger.Source())
 	if err != nil {
 		return nil, err
 	}
@@ -72,7 +72,7 @@ func (cli *CLI) Publish(
 	}
 	_, err = ctr.
 		WithWorkdir("/app").
-		WithMountedDirectory("/app", cli.Dagger.Source).
+		WithMountedDirectory("/app", cli.Dagger.Source()).
 		WithDirectory("/app/.git", gitDir).
 		WithEnvVariable("GH_ORG_NAME", githubOrgName).
 		WithSecretVariable("GITHUB_TOKEN", githubToken).
@@ -106,7 +106,7 @@ func (cli *CLI) TestPublish(ctx context.Context) error {
 	oses := []string{"linux", "windows", "darwin"}
 	arches := []string{"amd64", "arm64", "arm"}
 
-	builder, err := build.NewBuilder(ctx, cli.Dagger.Source)
+	builder, err := build.NewBuilder(ctx, cli.Dagger.Source())
 	if err != nil {
 		return err
 	}
--- a/dev/main.go
+++ b/dev/main.go
@@ -6,12 +6,15 @@ import (
 	"context"
 
 	"github.com/dagger/dagger/dev/internal/dagger"
+	"golang.org/x/sync/errgroup"
 )
 
 // A dev environment for the DaggerDev Engine
 type DaggerDev struct {
-	Source  *dagger.Directory // +private
+	Src     *dagger.Directory // +private
 	Version *VersionInfo
+	// When set, module codegen is automatically applied when retrieving the Dagger source code
+	ModCodegen bool
 
 	// Can be used by nested clients to forward docker credentials to avoid
 	// rate limits
@@ -33,10 +36,16 @@ func New(
 	}
 
 	return &DaggerDev{
-		Source:    source,
+		Src:       source,
 		Version:   versionInfo,
 		DockerCfg: dockerCfg,
 	}, nil
+}
+
+// Enable module auto-codegen when retrieving the dagger source code
+func (dev *DaggerDev) WithModCodegen() *DaggerDev {
+	dev.ModCodegen = true
+	return dev
 }
 
 // Check that everything works. Use this as CI entrypoint.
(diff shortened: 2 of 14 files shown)


jedevc:
Sorry, my bad, should have reviewed this one first - I left a couple comments in https://github.com/dagger/dagger/pull/7877.

https://github.com/dagger/dagger/pull/7877#discussion_r1678100351

> Hm, I'm not sure about switching everywhere to use this - that feels like we're adding extra cost to places that don't actually need it.
>
> Could we instead rename this method to GeneratedSource/SourceGenerated, and then keep Source as before?

shykes:
> I get that we can gradually optimize, but I would rather not add extra performance cost where we don't need it - I would rather we start with a faster pipeline, and then simplify from there.
> 
> Mixing up where the generated code is required from I think also _adds_ additional complexity - just by looking at a function, it's not clear if the generated code is actually used or not. It's fine to be longer and more explicit here I think.

I'll add an optional `codegen` argument to `Source()`

shykes:
> I get that we can gradually optimize, but I would rather not add extra performance cost where we don't need it - I would rather we start with a faster pipeline, and then simplify from there.
> 
> Mixing up where the generated code is required from I think also _adds_ additional complexity - just by looking at a function, it's not clear if the generated code is actually used or not. It's fine to be longer and more explicit here I think.

I'll add an optional `codegen` argument to `Source()`

jedevc:
> > I'll add an optional `codegen` argument to `Source()`
> 
> Done @jedevc

Now we just have `false`/`true` propagated everywhere. Can we either have a private method `.source` that takes an opts struct for it, so we don't need to explicitly write `true`/`false`, or just have a different method for `Source`/`GeneratedSource`?

shykes:
> Now we just have `false`/`true` propagated everywhere.

Only because  we don't have self calls yet, and Go doesn't have optional arguments. From a Dagger API standpoint this is transparent.

> Can we either have a private method `.source` that takes an opts struct for it, so we don't need to explicitly write `true`/`false`, or just have a different method for `Source`/`GeneratedSource`?

No, this is a Dagger module so we should design with the Dagger API in mind. When we ship self calls we will be rewarded with code simplification.

jedevc:
I would significantly prefer to have something more readable and then do more refactoring after we actually ship self calls, instead of refactoring in anticipation of a feature that hasn't even really been started.

Users who are looking at dagger to see a good example of how to use it will get confused by the extra booleans hanging around - I understand we will remove them - but in the meantime, I think it just adds verbosity that makes our own dagger module harder to understand and read.

I don't really want to keep debating this though, so approving.

dev/cli.go
```diff
@@ -21,13 +21,12 @@ func (cli *CLI) File(
...
 	platform dagger.Platform,
 ) (*dagger.File, error) {
-	builder, err := build.NewBuilder(ctx, cli.Dagger.Source)
+	builder, err := build.NewBuilder(ctx, cli.Dagger.Source(false))
 	if err != nil {
 		return nil, err
 	}
 	builder = builder.WithVersion(cli.Dagger.Version.String())
 	if platform != "" {
-		builder = builder.WithPlatform(platform)
```
jedevc, line 30: This shouldn't be removed.

shykes:
> I would significantly prefer to have something more readable and then do more refactoring after we actually ship self calls, instead of refactoring in anticipation of a feature that hasn't even really been started.

OK let me find a way

dev/cli.go
shykes: Thanks, fixed

shykes:
@jedevc I changed the design, to avoid propagating true/false everywhere. Also fixed the other issue you reported, and rebased on the latest lint fix. This should be all good to go!

dev/main.go
```diff
@@ -73,30 +82,62 @@ func (dev *DaggerDev) CLI() *CLI {
...
+		"sdk/elixir/runtime",
+		// FIXME: broken
+		// "sdk/python/dev",
+		"sdk/python/runtime",
+		"sdk/typescript/dev",
+		"sdk/typescript/dev/node",
+		"sdk/typescript/runtime",
+	}
+	eg, _ := errgroup.WithContext(context.TODO())
+	layers := make(chan *dagger.Directory, len(modules))
```
jedevc, line 108: This code doesn't actually need to be parallelized. All of the operations in this errorgroup are actually already lazy.

Sorry, I missed this earlier.

dev/main.go
```diff
@@ -33,12 +36,18 @@ func New(
...
 	return &DaggerDev{
-		Source:    source,
+		Src:       source,
 		Version:   versionInfo,
 		DockerCfg: dockerCfg,
 	}, nil
 }
 
+// Enable module auto-codegen when retrieving the dagger source code
+func (dev *DaggerDev) WithModCodegen() *DaggerDev {
```
jedevc, line 46: I think this is alright, but because with can be applied to the entire dev module, it's possible for the user to enable it, which could potentially interfere with the operation of the module. Ideally it would be private to avoid this.

I still don't quite understand why `Source`/`GeneratedSource` isn't alright here, but, I suppose this is fine as is.

dev/main.go
shykes: They may be lazy but I think there's enough overhead to slow me down. I added this after seeing a stair-shaped trace that seemed to support that.

jedevc (approved):

PR #7876, opened 2024-07-11
pipelines: implement a Ruff module, for linting Python code

shykes:
This PR only implements the Ruff module, but doesn't use it yet. A separate PR will wire it in. This is to avoid merge conflicts, since the Python SDK linting logic calls both golangci (go linting) and ruff (python linting) in the same function.

Since we separated ruff and golangci in different PRs (smaller PRs with smaller scope to survive the firehose of merge conflicts), it makes sense to add one more PR to the queue.

## Example usage:

Print a linting summary:

```console
dagger call -m github.com/dagger/dagger/dev/ruff@pull/7876/merge \
  lint --source https://github.com/dagger/dagger#main:sdk/python \
  summary
```

Dump the ruff report file:

```console
dagger call -m github.com/dagger/dagger/dev/ruff@pull/7876/merge \
  lint --source https://github.com/dagger/dagger#main:sdk/python \
  report \
  contents
```

Return an error if the report contains error (note that other commands don't fail, even if the report contains errors).

```console
dagger call -m github.com/dagger/dagger/dev/ruff@pull/7876/merge \
  lint --source https://github.com/dagger/dagger#main:sdk/python \
  assert
```


shykes:
FYI @charliermarsh :)

shykes:
I'm very confident all test fails are flakes, since this module isn't even wired in?

shykes:
Failing check is a flake

dev/ruff/main.go
```diff
@@ -0,0 +1,198 @@
...
+
+	"github.com/dagger/dagger/dev/ruff/internal/dagger"
+)
+
+var (
+	pythonVersion     = "3.11"
+	pythonImageRepo   = "docker.io/library/python"
+	pythonImageTag    = fmt.Sprintf("%s-slim", pythonVersion)
+	pythonImageDigest = "sha256:fc39d2e68b554c3f0a5cb8a776280c0b3d73b4c04b83dbade835e2a171ca27ef"
+	pythonImage       = pythonImageRepo + ":" + pythonImageTag + "@" + pythonImageDigest
```
helderco, line 18: I don't think you need Python to use Ruff. A better installation method is to use their OCI image. See end of [Installing Ruff](https://docs.astral.sh/ruff/installation/):

> On **Docker**, it is published as `ghcr.io/astral-sh/ruff`, tagged for each release and `latest` for the latest release.
> ```console
> docker run -v .:/io --rm ghcr.io/astral-sh/ruff check
> docker run -v .:/io --rm ghcr.io/astral-sh/ruff:0.3.0 check
> ```

dev/ruff/main.go
```diff
@@ -0,0 +1,198 @@
...
+	// +private
+	Source *dagger.Directory
+}
+
+// Return a JSON report file for this run
+func (run LintRun) Report() *dagger.File {
+	cmd := []string{
+		"ruff", "check",
+		"--exit-zero",
+		"--output-format", "json",
```
helderco, line 46: Ruff supports outputting as junit, same as popular test suites, including Pytest which we use.

Isn't it better to have a common util that can parse that format and reuse with multiple tools?

helderco (approved):
Approving for now. Can improve later when I integrate the changes in dev module (soon).

dev/ruff/main.go
shykes: This is directly derived from our current CI code. (`dev/sdk_python.go`). I am 100% on board with the improvement, but in the spirit of the "spin out then improve" approach, OK to make it a follow-up PR?

dev/ruff/main.go
shykes: I can add an argument, or additional function, to produce a junit report in a follow-up PR. For internal use the format doesn't matter since we wrap it in a GraphQL API anyway.

dev/ruff/main.go
helderco: Yes, of course. I already said as much in the summary 🙂

dev/ruff/main.go
helderco: I was thinking more in the lines of reducing the custom work on multiple linting modules (for different tools) that parse different report formats. Since `junit` is a common format, it can reduce the work to create the Dagger "linting report entries" (and even tests) from that output accross different tools in a common dependent module.

helderco: To be clear, it's just a suggestion for another time since the work to read Ruff's report is already done here.

shykes: Ah I see. Yeah that makes sense. Sadly not *all* our tools support junit (I forget which one doesn't, but I remember researching exactly that). We could still do it for the others, though!

PR #7877, opened 2024-07-11
pipelines: use the dirdiff module everywhere

shykes:
> [!WARNING]
>
> *This depends on #7872, please review that first*

This removes all references to the old util function DiffDirectory, and uses the newer dirdiff module instead.

As promised to @jedevc


dev/main.go
```diff
@@ -73,30 +74,58 @@ func (dev *DaggerDev) CLI() *CLI {
 	return &CLI{Dagger: dev}
 }
 
+// Return the Dagger source code, after applying codegen
+func (dev *DaggerDev) Source() *Directory {
```
jedevc, line 78: Hm, I'm not sure about switching *everywhere* to use this - that feels like we're adding extra cost to places that don't actually need it.

Could we instead rename this method to `GeneratedSource`/`SourceGenerated`, and then keep `Source` as before?

dev/cli.go
```diff
@@ -21,13 +21,12 @@ func (cli *CLI) File(
...
 	platform dagger.Platform,
 ) (*dagger.File, error) {
-	builder, err := build.NewBuilder(ctx, cli.Dagger.Source)
+	builder, err := build.NewBuilder(ctx, cli.Dagger.Source())
 	if err != nil {
 		return nil, err
 	}
 	builder = builder.WithVersion(cli.Dagger.Version.String())
 	if platform != "" {
-		builder = builder.WithPlatform(platform)
```
jedevc, line 30: :eyes: this shouldn't be removed.

dev/docs.go
```diff
@@ -75,18 +74,25 @@ func (d Docs) Lint(ctx context.Context) error {
...
-				Directory("/src"), nil
-		}, "CHANGELOG.md")
+		before := d.Dagger.Source()
+		// FIXME: spin out a changie module
+		after := dag.
+			Container().
+			From("ghcr.io/miniscruff/changie").
+			WithMountedDirectory("/src", d.Dagger.Source()).
+			WithWorkdir("/src").
+			WithExec([]string{"merge"}).
```
sipsma, line 93: This should be `{"/changie", "merge"}` I think.

sipsma (approved):
Typo in the changie command I think, but LGTM after that fix

jedevc (approved):

dev/docs.go
jedevc: Fixed, and rebased.

matipan (approved):

shykes:
Thank you for your help guys!

PR #7885, opened 2024-07-11
docs: adapt kubernetes integration page to new template

marcosnils:


levlaz:
Reviewing this PR caused me to find this https://github.com/dagger/dagger/issues/7895 - it should not block this review, but wanted to make note of the issue so we can polish it later.

levlaz:
Clicking on the zoom icon in the diagram does not seem to work for me, its possible its my machine though, could you confirm if this works for you @marcosnils?

[Uploading Screencast from 2024-07-11 16-10-26.webm…]()

levlaz (changes requested):
I propose that we either cut optimization 2 for now, or make sure that we say cloud cache is experimental right now. The way this reads right now implies to me that it's ready for production usage at scale.

marcosnils:
> Clicking on the zoom icon in the diagram does not seem to work for me, it just causes the screen to turn blank, its possible its my machine though, could you confirm if this works for you @marcosnils?

weird.. this works for me indeed.

> Reviewing this PR caused me to find this #7895 - it should not block this review, but wanted to make note of the issue so we can polish it later.

thx for reporting. Found the reason why it's happening and just added a comment in that issue :pray: 

> I propose that we either cut optimization 2 for now, or make sure that we say cloud cache is experimental right now. The way this reads right now implies to me that it's ready for production usage at scale.

I think we can mention cloud cache being experimental for now given that once we launch storage drivers, this approach will be very similar.

marcosnils:
just pushed some fixes @levlaz. Mind giving another :eyes: ?

docs/current_docs/integrations/kubernetes.mdx
```diff
@@ -4,73 +4,17 @@ slug: /integrations/kubernetes
 
 # Kubernetes
 
-## Deployment with Helm
+This section covers different strategies for deploying Dagger on a Kubernetes cluster.
+Its primary focus are teams that looking forward to self-host their CI infrastructure and
+use Dagger at its core to run their pipelines.
```
irabinovitch, line 9: Could we add more about why one would want to do this?

irabinovitch: Thanks Marcos. Much clearer why you'd wnat to use this now. However, could we move it to the top of the page? It feels a bit buried at the bottom right now.  I think we want someone coming to the docs for an integration to know immediately why they should bother setting this up and what the goals are.  

Could also do something like this:
Deploy Dagger Engine on your Kubernetes cluster to allow self-hosted CI runners run Dagger based pipelines. Enjoy the benefits of a shared global cache across all your Kubernetes hosted runners for faster builds.

levlaz (approved):
The updated version looks great, lets ship it!

PR #7887, opened 2024-07-11
chore(ts): use private cache for bun

jedevc:
We keep seeing this error over-and-over: https://dagger.cloud/dagger/traces/fac54672078ee95878b8684af05a814b?span=4c9874e76c2371de#ca88400300bc5d96

Lots of things of the form:

```
	            	error: moving "binary-extensions" to cache dir failed: PathAlreadyExists

	            	  From: .544670b85583f3c2binary-extensions

	            	    To: binary-extensions@2.3.0
```

Something doesn't seem very right there, but given it's flakiness, I suspect it's something to do with concurrent cache access - so we can just give bun access to Private cache.

### DIFF
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"main/internal/dagger"
 	"path"
 	"path/filepath"
 	"slices"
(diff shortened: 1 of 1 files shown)


helderco (approved):
Makes sense!

jedevc:
:scream: :scream: :scream: 

image

my favorite, yay.

PR #7889, opened 2024-07-11
Some very small version compat follow-ups

jedevc:
Some cherry-picks and pieces from #7858, since that one has some issues, and is risky!

Specifically:
- Fix the process of updating a `dagger.json` during `dagger develop` in certain edge cases with low engine versions (this was broken before, and now has some tests as well)
- The engine should serve the CLI the API that it's `clientVersion` indicates it has - this isn't super useful right now, but is quite nice in the context of #7858.

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -333,7 +333,7 @@ If not updating source or SDK, this is only required for IDE auto-completion/LSP
 			// ResolveFromCaller call first
 			modConf.Source = modConf.Source.ResolveFromCaller()
 
-			modSDK, err := modConf.Source.AsModule().SDK(ctx)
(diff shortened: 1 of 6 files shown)


sipsma (approved):

jedevc:
Bleh, if it's legitimately failing at this point, I'll pick it up again after the release.

sipsma:
> Bleh, if it's legitimately failing at this point, I'll pick it up again after the release.

We do need the fix to dagger develop though right?

One is a known timeout flake (the exit zero test), the other appears to be that flake where where the main module is not found.
* It's hard to tell for sure, but since it is so similar to that flake that has plagued us for a while in the python test, I'm okay with merging this and continuing forward as is, with whatever the problem is there being pre-existing.

Merging

Issue #7762, opened 2024-06-26
🐞 Module name cannot have `@` in path name


cbochs:
### What is the issue?

When using a local directory which has an `@` in the path name (e.g `/Users/cbochs@company.com`) as the source for a dagger module, dagger seems to assume it has special meaning and fails to resolve the module. This appears to be the case with `dagger init` as well:

```
dagger init --name=test --sdk=go test@test
```

Results in

```
✔ connect 1.1s
  ✔ starting engine 1.0s
    ✔ create 1.0s
✔ moduleSource(refString: "test@test"): ModuleSource! 0.0s
✔ ModuleSource.kind: ModuleSourceKind! 0.0s
✘ ModuleSource.resolveContextPathFromCaller: String! 0.0s
! failed to stat source root: failed to stat path: failed to receive file bytes message: rpc error: code = NotFound desc = rpc error: code = NotFound desc = stat path: lstat /Users/cbochs@company.com/git/daggerverse/test: no such file or directory

Error: failed to get configured module: failed to get local root path: input: moduleSource.resolveContextPathFromCaller resolve: failed to stat source root: failed to stat path: failed to receive file bytes message: rpc error: code = NotFound desc = rpc error: code = NotFound desc = stat path: lstat /Users/cbochs@company.com/git/daggerverse/test: no such file or directory
```

### Dagger version

dagger v0.11.8 (registry.dagger.io/engine) darwin/arm64

### Steps to reproduce

Run `dagger init --name=test --sdk=go test@test`

### Log output

```
$ dagger functions -m "/Users/cbochs@company.com/git/daggerverse/passenger"
✔ connect 1.3s
  ✔ starting engine 1.2s
    ✔ create 1.2s
✘ initialize 0.1s
! failed to get configured module: failed to get module ref kind: input: moduleSource resolve: local module source root path is absolute: /Users/cbochs

  ✘ moduleSource(refString: "/Users/cbochs@company.com/git/daggerverse/passenger"): ModuleSource! 0.0s
  ! local module source root path is absolute: /Users/cbochs

Error: failed to get configured module: failed to get module ref kind: input: moduleSource resolve: local module source root path is absolute: /Users/cbochs
```

helderco:
You can use `--name` to override. The directory name is just the default.

jedevc:
cc @grouville since you've been in this area.

The `@` is being parsed as the module version - however, local module paths can't really have versions, so we probably shouldn't do this transformation here.

[...]

PR #7891, opened 2024-07-11
fix(core): allow local module ref with @

grouville:
Fixes https://github.com/dagger/dagger/issues/7762

Current implementation of parseRefString does not allow local refs to have @, as it considers this character as the module version separator

However, this is not supposed to be true for local refs, to which the notion of version is not relevant

### DIFF
--- a/core/schema/modulesource.go
+++ b/core/schema/modulesource.go
@@ -244,11 +244,9 @@ func parseRefString(ctx context.Context, bk buildkitClient, refString string) pa
 
 	// We do a stat in case the mod path github.com/username is a local directory
 	stat, err := bk.StatCallerHostPath(ctx, parsed.modPath, false)
-	if err == nil {
-		if !parsed.hasVersion && stat.IsDir() {
-			parsed.kind = core.ModuleSourceKindLocal
-			return parsed
-		}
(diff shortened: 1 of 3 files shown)


core/schema/modulesource.go
```diff
@@ -240,7 +240,7 @@ type buildkitClient interface {
 // - if nothing worked, fallback as local ref, as before
 func parseRefString(ctx context.Context, bk buildkitClient, refString string) parsedRefString {
 	var parsed parsedRefString
-	parsed.modPath, parsed.modVersion, parsed.hasVersion = strings.Cut(refString, "@")
+	parsed.modPath = refString
```
sipsma, line 243: nit: the check right below here for `!parsed.hasVersion` will always be `true` now, so it's a bit redundant + confusing. I think with the new logic we can just remove that check now?

sipsma (approved):
One tiny nit but otherwise LGTM! Can just fix before merging if you agree

core/schema/modulesource.go
grouville: You're right, corrected 🙏

grouville:
Failure is unrelated to PR, merging

PR #7896, opened 2024-07-11
bring back redundant errors

vito:
since there seem to be cases that the progress report don't cover

will bring this back along with more robust TUI error handling


sipsma (approved):
Tested locally and the errors from incompatibilities show up now, thanks!

PR #7897, opened 2024-07-12
chore: add release notes for v0.12.0

sipsma:
Did some hand editing in this case:
1. Added a few missing breaking API changes (don't return void + use `@func` instead of `@field`)
2. Wrote an extra section under `Breaking Changes` with some more details on the new compat mode and how to upgrade modules via `dagger develop`
3. Reordered some of the entries to match the blog post + small misc formatting fixes

Would like to link to blog post here, but catch-22, so we can edit the release notes after the blog post is out with that link.

In the mean time, I do mention here that the breaking API change PRs have descriptions with more details. I think that's mostly true but will double check and do some light editing of the PR descriptions where it makes sense if needed.


sipsma:
Just realized this doesn't have notes about the alias removal in the Go SDK, will add

sipsma:
🤦‍♂️I had run a command to change my git remote so that I would avoid permission errors (https://github.com/dagger/dagger/blob/05b530dbbbcc06519101ca2d471a5e05997f97f3/RELEASING.md?plain=1#L202) and in the process accidentally set the remote I used for my fork to the main repo, so this PR is accidentally sourced from a branch in the dagger repo...

Closing and re-opening here: https://github.com/dagger/dagger/pull/7899

PR #7899, opened 2024-07-12
chore: add release notes for v0.12.0

sipsma:
Did some hand editing in this case:
1. Added a few missing breaking API changes (don't return void + use `@func` instead of `@field`)
2. Wrote an extra section under `Breaking Changes` with some more details on the new compat mode and how to upgrade modules via `dagger develop`
3. Reordered some of the entries to match the blog post + small misc formatting fixes
4. Added a breaking changelog about the Go SDK alias removal
   * In our current system, this only showed up in the Go SDK release notes. However, I don't think that makes sense for module-specific changes since it's a change that impacts users of the *engine* after upgrading, not users of the "go sdk as a library".

Would like to link to blog post here, but catch-22, so we can edit the release notes after the blog post is out with that link.

In the mean time, I do mention here that the breaking API change PRs have descriptions with more details.


CHANGELOG.md
```diff
@@ -7,6 +7,61 @@ and is generated by [Changie](https://github.com/miniscruff/changie).
...
 
+## v0.12.0 - 2024-07-11
+
+### 🔥 Breaking Changes
+
+- api: Don't fallback to the default command on `stdout` and `stderr` by @helderco in https://github.com/dagger/dagger/pull/7857
+- api: Return absolute path on export instead of boolean by @helderco in https://github.com/dagger/dagger/pull/7500
+- api: Align `Container.withNewFile` signature with `Directory.withNewFile` by @helderco in https://github.com/dagger/dagger/pull/7293
+- api: Skip entrypoint by default in `withExec` by @helderco in https://github.com/dagger/dagger/pull/7136
+- sdks: Don't return Void by @helderco in @helderco in https://github.com/dagger/dagger/pull/7773
```
sagikazarmark, line 18:
```suggestion
- sdks: Don't return Void by @helderco in https://github.com/dagger/dagger/pull/7773
```

???

jedevc: Nope, seems totally reasonable to me :laughing: 

Will tidy this up, thanks @sagikazarmark :heart:

.changes/v0.12.0.md
```diff
@@ -0,0 +1,52 @@
+## v0.12.0 - 2024-07-11
```
sagikazarmark, line 1: 2024-07-12

jedevc (approved):

.changes/v0.12.0.md
```diff
@@ -0,0 +1,74 @@
+## v0.12.0 - 2024-07-12
+
+### 🔥 Breaking Changes
+
+- sdks: Various breaking changes to SDKs \
+  See SDK-specific release notes for more information
```
helderco, line 9: The breaking changes in Python and TS are just about the return `Void` thing which isn't really breaking in these SDKs. In Go, in addition to that, it's the aliases removal thing but that should only affect Modules. And even on the `Void` return, it only affects these core APIs related to modules:
- `Module.serve` (very advanced use case)
- `Service.up` (for CLI)
- `FunctionCall.returnValue` (for SDKs)

Doesn't seem worth mentioning other than Go for those two things. I guess I’m being mostly picky with "Various" because it seems to be more than it is.

.changes/v0.12.0.md
```diff
@@ -0,0 +1,74 @@
...
+- api: Align `Container.withNewFile` signature with `Directory.withNewFile` by @helderco in https://github.com/dagger/dagger/pull/7293 \
+  Callers of `Container.withNewFile` will need to be updated depending on SDK.
```
helderco, line 11:
```suggestion
  Callers of `Container.withNewFile` will need to change `contents` from optional to required argument.
```

.changes/v0.12.0.md
```diff
@@ -0,0 +1,77 @@
...
+- api: Skip entrypoint by default in `withExec` by @helderco in https://github.com/dagger/dagger/pull/7136 \
+  Callers relying on a `Container`s default entrypoint will need to be updated
+  to opt-in with `UseEntrypoint`.
```
helderco, line 14:
```suggestion
  Callers relying on a `Container`'s entrypoint will need to be updated
  to opt-in with `useEntrypoint`.
```

gerhard (approved):
👍 🚀

.changes/v0.12.0.md
```diff
@@ -0,0 +1,77 @@
...
+- api: `Container.terminal` now returns a `Container` by @aluzzardi in https://github.com/dagger/dagger/pull/7586 \
+  Callers of `terminal` will need be updated to handle the updated type.
+- api: Return absolute path on export instead of boolean by @helderco in https://github.com/dagger/dagger/pull/7500 \
+  Callers expecting a boolean return will need to be updated to instead
+  handle a string path.
+- api: Don't fallback to the default command on `Container.stdout` and `Container.stderr` by @helderco in https://github.com/dagger/dagger/pull/7857 \
+  Callers of `stdout` and `stderr` without a previous `withExec` will need to
+  insert an explicit empty `withExec`.
```
helderco, line 22: Move this next to the entrypoing breaking change. If you also relied on the entrypoint in this case, you'll need both: empty args and `useEntrypoint`:
```go
build.WithExec([]string{}, dagger.ContainerWithExecOpts{UseEntrypoint: true})
```

.changes/v0.12.0.md
```diff
@@ -0,0 +1,77 @@
...
+will work again and your module can be updated in Daggerverse if desired.
+
+More detailed instructions on addressing the individual API incompatible
+changes can be found in the PR descriptions linked above.
+
+### Added
+
+- api: `terminal` can be inserted into the middle of `Container` and `Directory` pipelines to pop an interactive shell by @aluzzardi in https://github.com/dagger/dagger/pull/7586
+- api: Introduced module versioning compatibility by @jedevc in https://github.com/dagger/dagger/pull/7759
+- cli: Add `-q` flag and `DAGGER_QUIET=1` to restore previous verbosity default by @vito in https://github.com/dagger/dagger/pull/7822
```
helderco, line 56: Was this default verbosity changed in v0.11.9? I thought it was in this release.

jedevc: Good point - it changed after v0.11.8 https://github.com/dagger/dagger/issues/7801 (v0.11.9 is right after v0.11.8)

.changes/v0.12.0.md
jedevc: Scoped this down, and just mentioned go.

.changes/v0.12.0.md
jedevc: Moved!

.changes/v0.12.0.md
jedevc: I've added another release note here to mention https://github.com/dagger/dagger/pull/7671, where this change comes from.

helderco (approved):
:shipit:

lemeurherve:
Is it expected that the blog post link https://dagger.io/blog/dagger-0-12 currently returns a 404?

jedevc:
Should be fixed now!

lemeurherve:
Thanks!

PR #7901, opened 2024-07-12
fix: ensure we serve nested clients the correct schema version

jedevc:
Follow up from #7889

We should also serve generalized nested clients the right client API - and then have tests for this.


jedevc:
Test flake is the annoying bun one, see https://github.com/dagger/dagger/pull/7887 for some context.

gerhard (approved):
👍 🚀

PR #7906, opened 2024-07-12
chore: fix helm versions to keep semver in version field

jedevc:
Without this, the following steps from `RELEASING.md` create the wrong tag name:

```
export HELM_CHART_VERSION="$(awk '/^version: / { print $2 }' helm/dagger/Chart.yaml)"
git tag "helm/chart/v${HELM_CHART_VERSION:?must be set}" "${SDK_GIT_SHA:?must be set}"
```

This creates `helm/chart/vv0.12.0` (with two `v`s). According to https://helm.sh/docs/topics/charts/, the `version` field must be semver, no `v` prefix is allowed.

### DIFF
--- a/dev/helm.go
+++ b/dev/helm.go
@@ -41,7 +41,7 @@ func (h *Helm) chart() *Container {
 func (h *Helm) SetVersion(
 	ctx context.Context,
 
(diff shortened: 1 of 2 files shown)


gerhard (approved):
Thank you @jedevc 💪

PR #7917, opened 2024-07-14
fix(helm): ensure appVersion is always semver

### LINKED ISSUE #7916: 🐞Helm Chart fails with v0.12
### What is the issue?

It looks like [this change](https://github.com/dagger/dagger/commit/d7d52886952a6db83dba16aae51a685b87363ac5#diff-7f29c29598c7eecdb3332043e4de48253d5124ac50303c94a226735044ea2d98) introduced a `v` in front of the the fallback default `appVersion`.
As the appVersion is already prefix with a `v` that deamonset tries to use `registry.dagger.io/engine:vv0.12.0`

### Dagger version

v0.12.0

### Steps to reproduce

https://docs.dagger.io/integrations/kubernetes

### Log output

kubectl describe ....

```
Events:
  Type     Reason     Age                   From               Message
  ----     ------     ----                  ----               -------
  Normal   Scheduled  13m                   default-scheduler  Successfully assigned dagger/dagger-dagger-helm-engine-dfz9k to docker-desktop
  Normal   Pulling    12m (x4 over 13m)     kubelet            Pulling image "registry.dagger.io/engine:vv0.12.0"
  Warning  Failed     12m (x4 over 13m)     kubelet            Failed to pull image "registry.dagger.io/engine:vv0.12.0": Error response from daemon: manifest unknown
```

jedevc:
Fixes #7916 (which is a regression from #7854), follow-up to #7906

We should ensure that both fields are semver, and have no "v" prefix - then we can add the v prefix ourselves in the templated code.

### DIFF
--- a/dev/helm.go
+++ b/dev/helm.go
@@ -57,7 +57,7 @@ func (h *Helm) SetVersion(
 
 	version = strings.TrimPrefix(version, "v")
 	meta.Version = version
-	meta.AppVersion = "v" + version
+	meta.AppVersion = version
 
 	err = meta.Validate()
 	if err != nil {
--- a/helm/dagger/Chart.yaml
+++ b/helm/dagger/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: v0.12.0
+appVersion: 0.12.0
 description: Dagger Helm chart
 name: dagger-helm
 type: application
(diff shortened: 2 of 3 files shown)


gerhard (approved):
LGTM!

We discussed this sync, let's force push `helm/chart/v0.12.0` & have this fixed.

Short-term: add an integration test in Dagger that installs the chart. cc @sagikazarmark @marcosnils @matipan 

Long-term: consider pre-releases (we've multiple cases where a pre-release would have caught mistakes like this one). Also, releasing should be fully automated and take 1h at most. That's the ideal that we are working towards.

PR #7927, opened 2024-07-15
shorter span name for dagger call

vito:
This hack should buy us a bit more time for browse-by-function.

We initialize the Frontend/TUI/root span very very early in the CLI lifecycle - before even parsing `os.Args`, which requires loading modules, which needs to be beneath the root span.

So, we have to resort to hacks/heuristics to get a shorter name.

For `dagger call` specifically, statically parse `os.Args` with the following rules:

* Drop all flags passed before `call` (e.g. `-m`, `--debug`)
* Drop all flags passed after `call` and before its first function (i.e. module entrypoint flags)
* Keep everything thereaftor (including flags in the chain)

For anything else, we just keep the full `os.Args`.


cmd/dagger/span_name.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	if !isCall {
+		// if we're not a call, just use the original command
+		keep = args
+	} else if len(keep) == 0 {
+		// we're a call, but failed to parse the chain, probably confused by a
+		// boolean flag, so just show the full call
+		keep = fullCall
+	}
+	return strings.Join(keep, " ")
```
jedevc, line 101: This is just a display representation, so probably not an issue, but if an item in here contains a space, then we should have quoted it. But also, quoting a string (or even detecting if it properly needs quoting) is not very fun - I can't think of a library function/shlex library that'll do this neatly.

cmd/dagger/span_name_test.go
```diff
@@ -0,0 +1,40 @@
...
+		{args: []string{"dagger", "install", "foo"}, want: "dagger install foo"},
+		{args: []string{"dagger", "call", "foo"}, want: "foo"},
+		{args: []string{"dagger", "-m", "dev", "call", "foo"}, want: "foo"},
+		{args: []string{"dagger", "call", "-m", "dev", "foo"}, want: "foo"},
+		{
+			args: []string{"dagger", "call", "test", "important", "--race=true", "--parallel=16"},
+			want: "test important --race=true --parallel=16",
+		},
+		{args: []string{"dagger", "call", "--source", ".:default", "foo"}, want: "foo"},
+		{args: []string{"dagger", "call", "--source", ".:default", "foo", "--fizz"}, want: "foo --fizz"},
```
jedevc, line 27: suggestion: a testcase for `--source=.:default` combined, I think this is handled by the code above.

jedevc (approved):
Neat hack :tada: Left a couple of points, but nothing blocking.

jedevc:
FYI @shykes @marcosnils since this interacts with cloud things.

cmd/dagger/span_name.go
vito: Briefly checked if `shlex` has a helper for this and didn't see one. Going to punt, since at worst it's the same as before, don't want to polish the turd too much

vito:
Gonna merge as-is and then switch over to #7928 which might be able to undo some of the hackery now that it happens _after_ CLI parsing.

Issue #7856, opened 2024-07-09
✨ Do not send `dagger version` to cloud 


levlaz:
### What are you trying to do?

Right now when you call `dagger version` we send a trace to the cloud. This seems a bit silly and for larger teams contributes to the noise. We should supress these traces from being sent. 

### Why is this important to you?

Reducing signal vs noise in cloud is a good thing, this feels like a small but easy win.

jpadams:
I'd suggest that most commands NOT send traces except `dagger call`, but you could force a send with `-d` or other flag which would be great for collaborative troubleshooting.
```
DAGGER CLOUD COMMANDS                                                 TODAY            NEXT
  login         Log in to Dagger Cloud                                 ❌               ❌
  logout        Log out from Dagger Cloud                              ❌               ❌

DAGGER MODULE COMMANDS
  call          Call a module function                                 ✅               ✅
  config        Get or set the configuration of a Dagger module        ✅               ❌ (-d)
  develop       Setup or update all the resources needed               ✅               ❌ (-d)
                    to develop on a module locally
  functions     List available functions                               ✅               ❌ (-d)
  init          Initialize a new Dagger module                         ✅               ❌ (-d)
  install       Add a new dependency to a Dagger module                ✅               ❌ (-d)
  ```

levlaz:
> I'd suggest that most commands NOT send traces except `dagger call`, but you could force a send with `-d` or other flag which would be great for collaborative troubleshooting.
> ```
> DAGGER CLOUD COMMANDS                                                 TODAY            NEXT
>   login         Log in to Dagger Cloud                                 ❌               ❌
>   logout        Log out from Dagger Cloud                              ❌               ❌
> 
> DAGGER MODULE COMMANDS
>   call          Call a module function                                 ✅               ✅
>   config        Get or set the configuration of a Dagger module        ✅               ❌ (-d)
>   develop       Setup or update all the resources needed               ✅               ❌ (-d)
>                     to develop on a module locally
>   functions     List available functions                               ✅               ❌ (-d)
>   init          Initialize a new Dagger module                         ✅               ❌ (-d)
>   install       Add a new dependency to a Dagger module                ✅               ❌ (-d)
>   ```

This is a great suggestion, I agree. 

P.S. how did you make this awesome table? :)

jpadams:
@levlaz the power of Vim! Fixed-width FTW!

marcosnils:
> @levlaz the power of Vim! Fixed-width FTW!

[...]

PR #7928, opened 2024-07-15
fix: only initialize telemetry when connecting to engine

jedevc:
This goes someway towards fixing https://github.com/dagger/dagger/issues/7856.

Only commands that connect to a running engine upload telemetry, while commands that don't, like `version`, `login`/`logout`, and the top-level dagger command (as well as non-existent commands, like running `dagger nope-this-doesnt-exist`) won't initialize telemetry to send to cloud, or the frontend to display it.

### DIFF
--- a/cmd/dagger/engine.go
+++ b/cmd/dagger/engine.go
@@ -2,11 +2,15 @@ package main
 
 import (
 	"context"
-	"log/slog"
+	"os"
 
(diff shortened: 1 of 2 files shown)


vito (approved):
Tried out a bunch of commands including error states and it seems fine 👍

PR #7930, opened 2024-07-15
Do not use `quote` in labels to make it a valid string

matipan:
In https://github.com/dagger/dagger/commit/3bedee341e04a8d92a8c9161e6e72a8ad19e1170 we introduced a change that breaks the chart when installing it since it tries to create a label of `v"0.12.0"` which is not a valid value:

We don't need to apply the quote in this case, it will be treated as a string. With the fix I can install it successfully:


jedevc (approved):

jedevc:
As discussed, happy to take this one instead of the other.

Issue #7920, opened 2024-07-15
'dagger login' broken in 0.12


shykes:
## Problem

- In Dagger 0.12, `dagger login` prints an auth URL, but doesn't let me click on it. I also can't select any text anywhere in that terminal window.
- After I exit the TUI with `q` or `ctrl-C`, the link becomes clickable *BUT* at that point it's too late to click it. The auth process must take place while the TUI is running. **The only solution is to manually enter the URL in the browser while the TUI is running, and complete the auth process**
- If I exit the TUI and follow the printed link, the web auth process will succeed (my account is created), but it doesn't send me to my Cloud dashboard - it just prints "Congratulations! You're all set", which 1) is incorrect since the CLI failed to authenticate, and 2) gives me no further action to take.
- At the end of this process, the CLI is not visibly broken. Everything will appear to work. I just won't be logged in in the CLI. The `w` keystroke in TUI will send me to `https://github.com/traces/setup` which, since my browser is logged in, will redirect me to my list of traces. Then I will look for my trace, not find it, and be confused.

I tried this on 2 different terminal emulators (Zed and Ghostty on MacOS).

## Solution

1. **Make the login link clickable from the TUI**
2. default `dagger login` to open a browser directly, when the environment allows it. This creates an additional fallback
3. Make it easier to verify whether a CLI is logged in or not
4. Make the CLI auth flow more clear on the web side (don't tell me "I'm all set" if the CLI never actually authenticated). Or at least, redirect me to a useful place like my dagger.cloud dashboard

## Workarounds

- `dagger login`, keep the TUI running, manually copy the URL
- On MacOS: `dagger login`, take a screenshot of the terminal, click the link from the screenshot

shykes:
This is pretty bad.

shykes:
I think we should consider bypassing the TUI altogether for `dagger login`, we don't use its features at all, and it seems to be part of the problem in this particular case. Instead of going down a rabbit hole trying to fix clickable links in TUI for all commands, might be faster to just re-implement `dagger login` with good old `printf`?

marcosnils:
until this gets addressed, another workaround should be using `--progress plain` @shykes

shykes:
> until this gets addressed, another workaround should be using `--progress plain` @shykes 

I tried but it hung & printed nothing for me. Does it work for you?

jpadams:
As I mentioned in Discord, I unconciously did a hover+CMD on my mac to turn the link clickable
[...]

PR #7932, opened 2024-07-15
login: open browser automatically

vito:
Implements part 2 of @shykes' solution in https://github.com/dagger/dagger/issues/7920

We already integrate with the browser in other places, so why not extend it to `login`. This way we don't have to worry about whether the URL is clickable/selectable.

### DIFF
--- a/internal/cloud/auth/auth.go
+++ b/internal/cloud/auth/auth.go
@@ -7,8 +7,10 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/adrg/xdg"
+	"github.com/pkg/browser"
 	"golang.org/x/oauth2"
 )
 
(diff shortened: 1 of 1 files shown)


internal/cloud/auth/auth.go
```diff
@@ -47,7 +49,16 @@ func Login(ctx context.Context, out io.Writer) error {
...
+	authURL := deviceAuth.VerificationURIComplete
+
+	browserBuf := new(strings.Builder)
+	browser.Stdout = browserBuf
+	browser.Stderr = browserBuf
+	if err := browser.OpenURL(authURL); err != nil {
+		fmt.Fprintf(out, "\nFailed to open browser: %s\n\n%s", err, browserBuf.String())
+	}
+
+	fmt.Fprintf(out, "\nTo authenticate, visit:\n\t%s\n\n", authURL)
```
jedevc, line 61: nit: should we print a message that indicates that the browser window has been opened?

jedevc (approved):

internal/cloud/auth/auth.go
marcosnils: we're not printing the message in other places  so maybe we can just avoid it for consistency?

marcosnils (approved):

internal/cloud/auth/auth.go
vito: Tweaked the wording:

```
⬢ [fedora-toolbox:38] ❯ dagger-dev login dagger
Browser opened to: https://auth.dagger.cloud/activate?user_code=XLVW-QSJR
Confirmation code: XLVW-QSJR

Success.
```

Here's how it looks when we failed to open the browser:

```
⬢ [fedora-toolbox:38] ❯ dagger-dev login dagger
Failed to open browser: oh no

Authenticate here: https://auth.dagger.cloud/activate?user_code=GFZT-DZXD
Confirmation code: GFZT-DZXD
```

Also the confirmation code is in bold:

image

We only showed it in the URL before which seemed a bit subtle considering the landing page tells you to confirm it.

marcosnils (approved):

internal/cloud/auth/auth.go
marcosnils: SGTM :heavy_check_mark:

PR #7933, opened 2024-07-15
github: upload dev engine logs

vito:
Seeing some runs that seem to hang for a while after they supposedly complete. Ideally we could do this for all jobs, not just `testdev`, but it's easy to start with it.

note: thought about making this only run on failure, but it seems like we can just do it always; the `.zip` is ~7MB for a successful run and afaik doesn't cost us anything.


grouville (approved):
✅

PR #7936, opened 2024-07-16
pipeline: bump docusaurus module

levlaz:
I recently updated the docusaurus module to be compatible with v0.12, this commit updates it for this this project as well. I tested this out locally to make sure it is working as expected.

`dagger call -m dev --source . docs server as-service up`

### DIFF
--- a/dev/dagger.json
+++ b/dev/dagger.json
@@ -8,7 +8,7 @@
     },
     {
       "name": "docusaurus",
-      "source": "github.com/levlaz/daggerverse/docusaurus@72ec19206da7dbe2815da94c62e491b54935b633"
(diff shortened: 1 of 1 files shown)


jedevc:
Our CI isn't yet running v0.12 yet on main, so we should merge https://github.com/dagger/dagger/pull/7904 first.

levlaz:
Thanks there is no urgency on my end, but just to be clear this does not depend on 0.12, it just supports it, the only thing that really  changed was renaming the aliases (i.e. Container → dagger.Container)

Lev Lazinskiy
Head of Solutions @ dagger.io ( http://dagger.io )
e: ***@***.***

Sent via Superhuman ( ***@***.*** )

jedevc (approved):

PR #7937, opened 2024-07-16
ci: update API reference on new CI

grouville:
Since Daggerization of the CI, the API reference is not part of the docs update. It is however still exposed on our docs, and should follow the current API state: https://docs.dagger.io/api/reference/

This adds to the docs module a new `GenerateAPIReference` function, also added to the `Generate` function.

It also updates the index.html to the latest API reference

### DIFF
--- a/dev/docs.go
+++ b/dev/docs.go
@@ -14,8 +14,9 @@ type Docs struct {
 }
 
 const (
-	generatedSchemaPath = "docs/docs-graphql/schema.graphqls"
-	generatedCliZenPath = "docs/current_docs/reference/cli.mdx"
+	generatedSchemaPath       = "docs/docs-graphql/schema.graphqls"
+	generatedCliZenPath       = "docs/current_docs/reference/cli.mdx"
+	generatedApiReferencePath = "docs/static/api/reference/index.html"
 )
 
 const cliZenFrontmatter = `---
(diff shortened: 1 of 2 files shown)


jedevc:
:thinking: Huh, nice catch - however, weird thing - the page you link to doesn't seem to be out-of-date to me. It includes the new `DaggerEngine` API added in #7767.

So what's actually generating that page? Since the HTML file in our source tree is definitely out of date.

grouville:
> 🤔 Huh, nice catch - however, weird thing - the page you link to doesn't seem to be out-of-date to me. It includes the new `DaggerEngine` API added in #7767.
> 
> So what's actually generating that page? Since the HTML file in our source tree is definitely out of date: https://github.com/dagger/dagger/commits/main/docs/static/api/reference/index.html

Mmmh, I presume it is netlify then: https://github.com/dagger/dagger/blob/main/docs/netlify.toml#L2C1-L3C1

jedevc (approved):

jedevc:
> Might be nice to daggerize it nonetheless ?

Absolutely, yup, thanks for going down the rabbit hole and finding that :tada: Yeah, I think with this change potentially we can remove the `graphql-docs` build step - but let's leave that until Vikram is back from PTO.

crjm:
Yeah, it's not the best implementation.
The docs generation is done with a [`yarn run graphql-docs`](https://github.com/dagger/dagger/blob/main/docs/package.json#L15). Currently, the build command in Netlify is customized and set to `yarn build && yarn run graphql-docs` IIRC.
We should daggerize it for sure.

PR #7938, opened 2024-07-16
sdk(typescript): handle required enum options

jedevc:
:cherries: Cherry-picked out of #7438, since that one is kinda stalled, not really sure what to do with it.

In the process of the above PR, I'd attempted to add a required enum arg, which the typescript codegen hadn't previously covered - so I added this case.

While this isn't used by any API at the moment, it's useful to cover it anyway, so when we inevitably do, we don't spend any time on hunting down why it's not working.

### DIFF
--- a/cmd/codegen/generator/typescript/templates/src/method_solve.ts.gtpl
+++ b/cmd/codegen/generator/typescript/templates/src/method_solve.ts.gtpl
@@ -81,8 +81,9 @@
 
       		{{- with $optionals }}
       			{{- if $required }}, {{ end }}
-				{{- "" }}...opts{{- if gt (len $enums) 0 -}}, __metadata: metadata{{- end -}}
+				{{- "" }}...opts
 			{{- end }}
+      {{- if gt (len $enums) 0 -}}, __metadata: metadata{{- end -}}
 {{- "" }} },
 		{{- end }}
         },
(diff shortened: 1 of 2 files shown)


cmd/codegen/generator/typescript/templates/src/method_solve.ts.gtpl
```diff
@@ -81,8 +81,9 @@
 
       		{{- with $optionals }}
       			{{- if $required }}, {{ end }}
-				{{- "" }}...opts{{- if gt (len $enums) 0 -}}, __metadata: metadata{{- end -}}
+				{{- "" }}...opts
 			{{- end }}
+      {{- if gt (len $enums) 0 -}}, __metadata: metadata{{- end -}}
```
TomChv, line 86: I think you'll need to regenate the TS client if you updated the generator

jedevc: > While this isn't used by any API at the moment

This isn't currently used by any core API at the moment, so the codegen doesn't need to be run. The linter detects if it's not been run, and it's not currently failing, so it's up to date.

gerhard:
@TomChv 👆

cmd/codegen/generator/typescript/templates/src/method_solve.ts.gtpl
TomChv: Okay, LGTM then

TomChv (approved):

PR #7940, opened 2024-07-16
pipelines: use alpine and wolfi modules

shykes:
Make our pipelines more efficient and easier to read, by using the centralized `wolfi` and `alpine` modules.


shykes:
I have a remaining issue with the `apko` module and arch configuration

engine/distconsts/consts.go
```diff
@@ -19,7 +19,6 @@ const (
 
 const (
 	AlpineVersion = "3.20.1"
-	AlpineImage   = "alpine:" + AlpineVersion
```
sipsma, line 22: What's the reasoning for getting rid of this? Can we at minimum maintain usage of the `AlpineVersion` const elsewhere like in `core/integration/images.go` + `core/terminal.go`? It makes it much easier to update everywhere and we haven't had any needs for different alpine versions in different places.

shykes: The idea is that the alpine module should be the place where alpine version is upgraded.

- `core/integration/images.go` should ideally use the alpine module also, but I don't dare touch it for now
- `core/terminal.go` is a special case, there's no reason for that feature to use the same alpine version as our CI pipelines - or even to couple its use of alpine to our CI pipeline's use of alpine. For example if we switch our CI build to wolfi, does that mean we should also move everyone's default terminal to wolfi? It seems weird.

shykes: Generally relying on `const` and `distconsts` is at odds with making modules the source of truth for composing neatly encapsulated devops tasks.

sipsma (approved):
Question about the alpine image const, otherwise LGTM!

engine/distconsts/consts.go
sipsma: The reasoning is more that if there is a CVE or something in alpine it's easier to only have to remember there's one place to update it. Generally the container used for interactive terminals shouldn't be something users rely on toe be stable, at least in the current state of things where it's just a default one we choose (IMO).

It's not a huge deal either way, this is splitting hairs in obscure situations.

shykes: Yeah in the case of terminal, in a perfect world I would advocate for depending on a module also (either the alpine module or another), to centralize updates as you say. But since it's in the core schema, that's not an option of course.

Maybe we change the core terminal image to target alpine latest?

I hear you re: splitting hairs :) Thanks for being open minded

jedevc:
Rebased this to handle all the changes we've made in the meantime, and also updated to pin the alpine version in places, since this *has* broken things in the past, and we shouldn't rely on `edge` everywhere.

I'm gonna merge this now (in the spirit of moving quickly), if I've made any changes that aren't desirable, we can revert those in a separate PR - I'd like to get the alpine/wolfi changes in now.

jedevc (approved):

Issue #7900, opened 2024-07-12
Test flake: TestLegacy `package dagger/test/internal/dagger is not in std`


sipsma:
Started hitting a lot of flakes in the new `TestLegacy` suite, errors look like:
```
input: module.withSource.initialize resolve: failed to initialize module: failed to call module "test" to get functions: call constructor: process "go build -o /runtime ." did not complete successfully: exit code: 1
                            
Stderr:
main.go:3:8: package dagger/test/internal/dagger is not in std (/usr/local/go/src/dagger/test/internal/dagger)
```

Could be unrelated, but it smells awfully similar to https://github.com/dagger/dagger/issues/7662 in that both could be explain by dirs missing files or otherwise having unexpected contents.

---

Thankfully can repro locally, so made some progress in debugging:

On one occurrence of the error in `TestLegacyNoExec`, the contents are showing that the go.mod of the module starts with:
```
module dagger/bare

go 1.22.4
```

This is extremely wrong.... That go.mod is the one that should be used for `TestLegacyExportAbsolutePath`.... This test should have `module dagger/test` (which explains the error)

The other files in there like `main.go` and `dagger.gen.go` all look like the right ones for `TestLegacyNoExec`. At least in this case it's just `go.mod` which is wrong.

Haven't root caused it yet, but could run from Go SDK problems all the way down to buildkit solver problems.
* All I can say is that the test case itself looks totally fine so far (and I can't imagine how this would ever be a test case problem anyways)

---

FTR, so far I've been debugging by running with this diff:
```diff
diff --git a/core/schema/sdk.go b/core/schema/sdk.go
index 8bbaed3e4..c031b15e1 100644
--- a/core/schema/sdk.go
+++ b/core/schema/sdk.go
@@ -364,9 +364,13 @@ func (sdk *goSDK) Runtime(
                                {
                                        Name: "args",
                                        Value: dagql.ArrayInput[dagql.String]{
+                                               "sh", "-c",
+                                               `go build -o /runtime . || (echo WTF && sleep 100000 && false)`,
+                                               /* TODO:
                                                "go", "build",
                                                "-o", goSDKRuntimePath,
                                                ".",
+                                               */
                                        },
                                },
                        },
```

Then running `dagger -m dev call --source=.:default test custom --timeout=100h --run=TestLegacy --pkg=./core/integration --parallel=8 --count=100`.

[...]

PR #7943, opened 2024-07-16
Fix nested client filesync incorrect deduplication

sipsma:
As detailed in #7900, certain obscure corner cases can result in
separate clients on separate hosts (or separate nested exec containers)
from re-using each others' filesync cache refs unexpectedly and, if
files happen to match metadata precisely, end up with incorrect
file contents being used in the sync.

The requirements for this to happen are as obscure as it gets, but it is
impacting our CI and causing flakes, so needs a fix. It is *technically*
possible for users to hit this too, so may as well solve it for them.

Previously part of the cause here was that we used hostname as a
semi-stable host identifier, but nested clients mostly have the same
default container hostname.
* There was much more to it than this, but this is the part we can
  address.

The fix here replaces the hostname with an ID that is persisted in the
client's XDG state dir (i.e. `~/.local/state/dagger/`). It's random when
newly initialized but re-used after that.

That solves this problem for non-nested clients. For nested clients, we
just set this to a random ID in each exec for now.
* It would also be possible to try to persist this in the nested exec,
  but it's not clear yet if that's worth the complication in that case,
  so keeping it simple for now.

### DIFF
--- a/engine/client/stableid.go
+++ b/engine/client/stableid.go
@@ -1,0 +1,65 @@
+package client
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/adrg/xdg"
+	"github.com/dagger/dagger/engine/slog"
+	"github.com/moby/buildkit/identity"
+)
+
+const StableIDFileName = "stable_client_id"
+
+// GetHostStableID returns a random ID that's persisted in the caller's XDG state directory.
+// It's currently used to identify clients that are executing on the same host in order to
+// tell buildkit which filesync cache ref to re-use when syncing dirs+files to the engine.
+func GetHostStableID(lg *slog.Logger) string {
+	id, err := internalGetStableID(filepath.Join(xdg.StateHome, "dagger"))
+	if err != nil {
+		lg.Warn("failed to get stable ID, defaulting to random value", "error", err)
+		return identity.NewID()
+	}
+	return id
+}
+
+func internalGetStableID(parentDirPath string) (string, error) {
+	if parentDirPath == "" {
+		return "", errors.New("parentDirPath is not set")
+	}
+
+	stableIDPath := filepath.Join(parentDirPath, StableIDFileName)
+	if stableID, err := os.ReadFile(stableIDPath); err == nil {
+		// already exists
+		return string(stableID), nil
+	} else if !os.IsNotExist(err) {
+		return "", fmt.Errorf("failed to read stable ID: %w", err)
+	}
+
+	if err := os.MkdirAll(parentDirPath, 0o700); err != nil {
+		return "", fmt.Errorf("failed to create parent directory: %w", err)
+	}
+
+	// there's a race in the (obscure) case of clients concurrently running here,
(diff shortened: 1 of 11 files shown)


sipsma:
Ah, I actually just remembered that there's a really big tradeoff with relying on hostnames (separate from ones I originally had in mind), which is that it results in CNI network namespaces not being re-usable, which adds likely meaningful performance overhead...

I guess I will just go with the persistent host identifier approach after all.

Also, on my first try here I did get a flake in the `TestPythonProjectLayout` tests still 😢 Was hoping that one might have been the same as #7900 (which is about `TestLegacy`), but either it isn't or there was something wrong about the "unique hostname" approach. Will see what happens after I switch approaches I guess.

sipsma:
Currently letting `TestLegacy` run in a loop for a good while to verify flake is gone. Will do the same for the module tests as a whole. GHA is having an incident right now, but once that's recovered will do re-runs in CI too to check.

sipsma:
Locally:
* `TestLegacy` passed in a loop for an hour
* `TestModule` has passed twice in a row (takes ~30m locally, so gonna keep it looping for a while longer)

In CI:
* Still hitting separate flakes (https://github.com/dagger/dagger/issues/7979 and https://github.com/dagger/dagger/issues/7662)
* But no occurrences of `TestLegacy` flake https://github.com/dagger/dagger/issues/7900

So at least seems to have fixed the expected flake, though sad the others are still going to require their own investigation.

Gonna keep looping in CI a bit more to be extra sure.

sipsma:
Had to do lots of re-runs after other unrelated flakes, but never saw `TestLegacy` flake in CI either, so feel good it's gone.

engine/client/stableid.go
```diff
@@ -0,0 +1,76 @@
...
+	// but the worst case is some of the clients using different IDs, which is
+	// just a temporary mild performance deficiency, so not worth more complication
+	tmpFile, err := os.CreateTemp(parentDirPath, StableIDFileName)
+	if err != nil {
+		return "", fmt.Errorf("failed to create stable ID temp file: %w", err)
+	}
+	defer os.Remove(tmpFile.Name())
+	defer tmpFile.Close()
+
+	if err := tmpFile.Chown(uid, gid); err != nil {
```
jedevc, line 62: Hello, welcome to windows hell again :tada: `File.Chown` always fails on Windows: https://pkg.go.dev/os#File.Chown :scream:

jedevc (approved):
Aside from the note that this likely won't work on windows right now (haven't tested to be sure), this looks good to me :tada:

*insane* deep dive, nice job :partying_face:

engine/client/stableid.go
sipsma: Oi, I had not encountered that "fun" fact previously, thanks for the pointer! I'll patch this up quick; I actually can drop the chown for now. It was added in an earlier iteration where I was trying to run this code in nested execs too (and thus needed to deal with different uid/gid), but I ended up backing out of that for simplicity.

sipsma: Fixed in force push

PR #7944, opened 2024-07-16
feat(ts-sdk) upgrade decorator to non-experimental

### LINKED ISSUE #7499: ✨ [TypeScript SDK] Consider using ECMAScript Decorators

Currently, the TypeScript SDK uses experimental decorators which are considered deprecated. TypeScript has provided support for standard ECMAScript decorators since [version 5.0](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#decorators). I would suggest considering the use of standard decorators over the now deprecated experimental decorator implementation which will never be usable from JavaScript and might even get removed from TypeScript in the (admittedly very) long-term.

The main benefit comes from better integration into the larger JS ecosystem and future proofing the current work. The decorator's proposal has been at [stage 3](https://github.com/tc39/proposal-decorators) in TC39 for a while now. While It will land in all the JS engines in the foreseeable future (browser vendors are actively working on it), it's usable already today in TypeScript.

*No respon*

TomChv:
This commit is a breaking change on decorators.
It removes the `experimentalDecorators` setting in the `tsconfig.json` to use the official released.

This changes the prototypes of decorators to now take the target and the context as argument.

This commit also updates the generated template and has a script to update existing tsconfig.

Fixes #7499

### DIFF
--- a/core/integration/module_typescript_test.go
+++ b/core/integration/module_typescript_test.go
@@ -1104,3 +1104,44 @@ class Test {
 		require.JSONEq(t, `{"test": {"foo": "bar"}}`, out)
 	})
 }
+
+func (ModuleSuite) TestTypeScriptExperimentalDecoratorsCompatibility(ctx context.Context, t *testctx.T) {
+	t.Run("should work with experimental decorators", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		modGen := c.Container().From(golangImage).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work").
+			With(daggerExec("init", "--name=test", "--sdk=typescript")).
+			With(sdkSource("typescript", `
+import { func, object } from "@dagger.io/dagger"
+
+@object()
+class Test {
+  @func()
+  foo(): string {
+    return "bar"
+  }
+}
+`,
+			)).
+			WithoutFile("/work/dagger/tsconfig.json").
+			WithNewFile("/work/dagger/tsconfig.json", `
+{
+  "compilerOptions": {
+    "target": "ES2022",
+    "moduleResolution": "Node",
+		"experimentalDecorators": true,
+		"emitDecoratorMetadata": true,
+    "paths": {
+      "@dagger.io/dagger": ["./sdk"],
+      "@dagger.io/dagger/telemetry": ["./sdk/telemetry"]
+    }
+  }
+}`)
+
+		out, err := modGen.With(daggerQuery(`{test{foo}}`)).Stdout(ctx)
+		require.NoError(t, err)
+		require.JSONEq(t, `{"test": {"foo": "bar"}}`, out)
+	})
+}
--- a/core/integration/module_typescript_test.go
+++ b/core/integration/module_typescript_test.go
@@ -1125,14 +1125,13 @@ class Test {
 }
 `,
 			)).
-			WithoutFile("/work/dagger/tsconfig.json").
 			WithNewFile("/work/dagger/tsconfig.json", `
 {
   "compilerOptions": {
     "target": "ES2022",
     "moduleResolution": "Node",
-		"experimentalDecorators": true,
-		"emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
     "paths": {
       "@dagger.io/dagger": ["./sdk"],
       "@dagger.io/dagger/telemetry": ["./sdk/telemetry"]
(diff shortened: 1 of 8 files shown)


helderco:
Can you clarify how this is a breaking change?

TomChv:
> Can you clarify how this is a breaking change?

It drops the support for experimental decorator which have a different prototype than the no experimental.
So the code will run if it's executed with `experimentalDecorators: true`, this is why I added the change in the script.

So normally the script should ensure a smooth transition for current & new user but it still a breaking change even if it should not affect users.

helderco:
If the runtime is updating experimentalDecorator setting and it doesn't affect users then I don't see how this is a breaking change. Which users need to updating their code manually when updating Dagger because of this?

TomChv:
> If the runtime is updating experimentalDecorator setting and it doesn't affect users then I don't see how this is a breaking change. Which users need to updating their code manually when updating Dagger because of this?

Normally nobody if the update from the script is well done

TomChv:
> If users are affected, we need to make sure that no module in daggerverse breaks by providing backwards compat with old modules. See all of the work done in the last release to provide this level of compat.

The user should at least need to execute `dagger develop`

jedevc:
@TomChv sorry, I think I'm still confused - what's the actual impact of this change? I get that the decorator type is changed - but what impact does this have?

Can you give an example of a module that can potentially work on v0.12, but would break with this change?

helderco:
I think the main question is: do some users need `experimentalDecorators` outside of Dagger in a module?

If we want to support that use case then it's a breaking change for them but I think it's reasonable that we don't support it because the non-experimental decorators have been stable for a while. It's probably more likely that users have been limited by Dagger's experimental decorators.

If Dagger doesn't work with experimental decorators anymore, I think it's reasonable to udpate the tsconfig not only on develop but also on dagger call. Same reason as we do it for client bindings. If we do this, then it's not a breaking change.

TomChv:
> Can you give an example of a module that can potentially work on v0.12, but would break with this change?

There's no difference in the code, it's only in the configuration, the `tsconfig.json` must not have `experimentalDecorators` field set to true, that's all.
And it should use at least typescript >5 but we already use it.

TomChv:
> If Dagger doesn't work with experimental decorators anymore, I think it's reasonable to udpate the tsconfig not only on develop but also on dagger call. Same reason as we do it for client bindings. If we do this, then it's not a breaking change.

Yeah I updated it in both case, so I think it's not a breaking change for the user

TomChv:
I'll update the release note to `Added` instead of breaking

helderco:
> I'll update the release note to `Added` instead of breaking

Should be "Changed".

TomChv:
> Should be "Changed".

Changes done

helderco:
Missing integration test for an existing module that has experimental setting, and then runs `dagger call` and `dagger develop`.

helderco (commented):
Missing integration test for an existing module that has experimental setting, and then runs `dagger call` and `dagger develop`.

core/integration/module_typescript_test.go
```diff
@@ -1104,3 +1104,44 @@ class Test {
...
+    "target": "ES2022",
+    "moduleResolution": "Node",
+		"experimentalDecorators": true,
+		"emitDecoratorMetadata": true,
+    "paths": {
+      "@dagger.io/dagger": ["./sdk"],
+      "@dagger.io/dagger/telemetry": ["./sdk/telemetry"]
+    }
+  }
+}`)
```
helderco, line 1141: nit: can you fix the formatting here?

core/integration/module_typescript_test.go
```diff
@@ -1104,3 +1104,44 @@ class Test {
...
+@object()
+class Test {
+  @func()
+  foo(): string {
+    return "bar"
+  }
+}
+`,
+			)).
+			WithoutFile("/work/dagger/tsconfig.json").
```
helderco, line 1128: You don't need to remove because the next line replaces it.

core/integration/module_typescript_test.go
```diff
@@ -1104,3 +1104,44 @@ class Test {
...
+    "paths": {
+      "@dagger.io/dagger": ["./sdk"],
+      "@dagger.io/dagger/telemetry": ["./sdk/telemetry"]
+    }
+  }
+}`)
+
+		out, err := modGen.With(daggerQuery(`{test{foo}}`)).Stdout(ctx)
+		require.NoError(t, err)
+		require.JSONEq(t, `{"test": {"foo": "bar"}}`, out)
```
helderco, line 1145: Can you replace with `dagger call` (just because it's what people use and we want to confirm that follows expected behavior) and add `dagger develop` to show that the "experimental" is removed? Followed by another `dagger call` to confirm it runs without the setting as well.

This follows the sequence of someone updating a module:
- dagger call (with experimental)
- dagger develop
- dagger call (without experimental)

Xiot:
Here are the [docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#differences-with-experimental-legacy-decorators) from TS 5.0

> This new decorators proposal is not compatible with --emitDecoratorMetadata, and it does not allow decorating parameters. Future ECMAScript proposals may be able to help bridge that gap.

Typescript 5.2 did add support for metadata though [blog](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#decorator-metadata)

TomChv:
> ie, ecmascript decorators don't support using them on method arguments [proposal](https://github.com/tc39/proposal-class-method-parameter-decorators)

I'm confused, there are examples in the doc: ie, ecmascript decorators don't support using them on method arguments [proposal](https://github.com/tc39/proposal-class-method-parameter-decorators)

However, I might be able to workaround that with an hybrid version? https://blog.logrocket.com/practical-guide-typescript-decorators/#parameter-decorators

Xiot:
> I'm confused, there are examples in the doc: ie, ecmascript decorators don't support using them on method arguments [proposal](https://github.com/tc39/proposal-class-method-parameter-decorators)

That proposal is only Stage 1

> However, I might be able to workaround that with an hybrid version? https://blog.logrocket.com/practical-guide-typescript-decorators/#parameter-decorators

I don't think that TS does any tranpilation with ES decorators, so if it isn't supported in node, then it may not work.  Although I haven't tried working with ES decorators yet, so check it and see.

TomChv:
Hmm okay so I should hold this PR until ES decorators supports parameters

TomChv:
Alright, I'm closing this PR since it doesn't support function's argument decorator and we need it for #7744.

This will have to wait for https://github.com/tc39/proposal-class-method-parameter-decorators to be completed to be done.

PR #7945, opened 2024-07-17
docs: fixup linters examples with ruff

jpadams:
These example had issues due to a change in the Dagger CI modules layout and an older dependency. Changed to using `ruff` module instead. Otherwise, very similar.

### DIFF
--- a/docs/current_docs/manuals/user/artifacts/consumption/export.mdx
+++ b/docs/current_docs/manuals/user/artifacts/consumption/export.mdx
@@ -26,10 +26,10 @@ Instead of exporting an entire directory, you can also export a file. Here is an
 dagger -m github.com/kpenfound/dagger-modules/golang@v0.2.0 call build --source=https://github.com/dagger/dagger --args=./cmd/dagger file --path=dagger export --path=./my-binary-file
 ```
 
-Here is another example, this time exporting the rules file used by a linter Dagger Function as `/tmp/rules` on the host:
+Here is another example, this time exporting the results of a `ruff` linter Dagger Function as `/tmp/report.json` on the host:
 
 ```shell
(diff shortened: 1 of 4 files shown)


docs/current_docs/manuals/user/artifacts/consumption/export.mdx
```diff
@@ -26,10 +26,10 @@ Instead of exporting an entire directory, you can also export a file. Here is an
 dagger -m github.com/kpenfound/dagger-modules/golang@v0.2.0 call build --source=https://github.com/dagger/dagger --args=./cmd/dagger file --path=dagger export --path=./my-binary-file
 ```
 
-Here is another example, this time exporting the rules file used by a linter Dagger Function as `/tmp/rules` on the host:
+Here is another example, this time exporting the results of a `ruff` linter Dagger Function as `/tmp/report.json` on the host:
 
 ```shell
-dagger -m github.com/dagger/dagger/linters/markdown@88d89e8d15ab6ad9ca4043a920d3cd735a6405fd call rules export --path=/tmp/rules
+dagger call -m github.com/dagger/dagger/dev/ruff@a29dadbb5d9968784847a15fccc5629daf2985ae lint --source https://github.com/dagger/dagger report export --path=/tmp/report.json
```
marcosnils, line 32: nit: it'd be nice to have the `ruff` module somwhere else so we can tag it with a proper semver version.

marcosnils (approved):
LGTM. Only left a nit comment.

docs/current_docs/manuals/user/artifacts/consumption/export.mdx
jpadams: Agreed. It does illustrate the options, though, which is kind of nice. We have `golang@v0.2.0` and `wolfi@0.1.4` as examples in some of the same places.

PR #7946, opened 2024-07-17
pipelines: daggerDev().cli().binary()

shykes:
`binary` is more intuitive than `file` (I kept both for backwards compat).

### DIFF
--- a/dev/cli.go
+++ b/dev/cli.go
@@ -14,8 +14,17 @@ type CLI struct {
 	Dagger *DaggerDev // +private
 }
 
-// Build the CLI binary
+// Build the CLI binary (deprecated)
 func (cli *CLI) File(
+	ctx context.Context,
+	// +optional
+	platform dagger.Platform,
+) (*dagger.File, error) {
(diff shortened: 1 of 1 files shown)


dev/cli.go
```diff
@@ -15,7 +15,13 @@ type CLI struct {
 }
 
 // Build the CLI binary
-func (cli *CLI) File(
+// +deprecated
+func (cli *CLI) File(ctx context.Context, platform dagger.Platform) (*dagger.File, error) {
+	return cli.Binary(ctx, platform)
+}
```
jedevc, line 21: Couple points:
- This removes the optionalness of `platform`, so callers will still likely break
- `// +deprecated` isn't a pragma with a special value, I don't think we should use it like this (if we want it to be a thing, we can add it)

shykes: > This removes the optionalness of platform, so callers will still likely break

That's an oversight, thanks for catching.

jedevc (approved):

PR #7948, opened 2024-07-17
feat: add --compat flag to dagger to specify api version

jedevc:
We removed this last minute in https://github.com/dagger/dagger/pull/7759/#discussion_r1668419158.

However, it does seem like we need *something* here - so proposing this back in. Let the bikeshedding of naming/etc begin, I don't have particularly strong opinions here :tada:

### DIFF
--- a/.changes/unreleased/Added-20240717-101724.yaml
+++ b/.changes/unreleased/Added-20240717-101724.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: 'cli: new --compat flag for develop to target a specific api version'
+time: 2024-07-17T10:17:24.1005004+01:00
+custom:
+  Author: jedevc
+  PR: "7948"
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -36,8 +36,9 @@ var (
 	moduleURL   string
 	moduleFlags = pflag.NewFlagSet("module", pflag.ContinueOnError)
 
-	sdk       string
-	licenseID string
+	sdk           string
+	licenseID     string
+	compatVersion string
 
 	moduleName       string
 	moduleSourcePath string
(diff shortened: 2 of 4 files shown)


helderco:
`--no-upgrade` could be an option too, to just not update that field. But good with `--compat <version>` as well since it's perhaps clearer what it's for.

Maybe allow `--compat [version]` (optional version) so that a standalone `--compat` means "don't update engineVersion", while specifying the version overrides the value for it.

I also suggest keeping it for `develop` only, instead of new modules.

jedevc:
I'd kinda like to have it also for `init` since that can do codegen as well. I think it's definitely a *less* common usecase, but I definitely have it while playing around.

I like the sound of an optional version, to keep the current version.

I'll also add some tests.

helderco:
> I'd kinda like to have it also for `init` since that can do codegen as well. I think it's definitely a _less_ common usecase, but I definitely have it while playing around.

How about this workaround then?
```
dagger init
dagger develop --compat --sdk=...
```

jedevc:
@helderco done all the above.

Weird note - it needs to specified either as `--compat` or `--compat=<version>` - you can't use `--compat <version>`, see https://github.com/spf13/cobra/issues/866.

shykes (approved):

docs/static/api/reference/index.html
helderco, line 1: Is this now always producing a diff on generate due to random ids? Might make sense to discard changes if so.

helderco (approved):
lgtm 👍

docs/static/api/reference/index.html
jedevc: Hm, this was added in https://github.com/dagger/dagger/pull/7937. Yeah, these look random...

But what's super annoying is that even today these random examples will be re-updated on every single netlify deploy? We should definitely fix this.

cc @grouville (also, looks like we're missing this in docs linting, which would have caught this)

PR #7939, opened 2024-07-16
pipelines: upgrade to Dagger 0.12

shykes:

PR #7949, opened 2024-07-17
ci: fix broken cli publishing from upgrade to v0.12

jedevc:
Follow-up to https://github.com/dagger/dagger/pull/7939

We need to use the entrypoint explicitly here, see https://github.com/dagger/dagger/actions/runs/9971938618/job/27554089721#step:8:453

### DIFF
--- a/dev/cli.go
+++ b/dev/cli.go
@@ -91,7 +91,9 @@ func (cli *CLI) Publish(
 			return ctr
 		}).
 		WithEntrypoint([]string{"/sbin/tini", "--", "/entrypoint.sh"}).
-		WithExec(args).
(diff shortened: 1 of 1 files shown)


dev/cli.go
```diff
@@ -91,7 +91,9 @@ func (cli *CLI) Publish(
 			return ctr
 		}).
 		WithEntrypoint([]string{"/sbin/tini", "--", "/entrypoint.sh"}).
-		WithExec(args).
+		WithExec(args, dagger.ContainerWithExecOpts{
+			UseEntrypoint: true,
```
shykes, line 95:
```suggestion
			// FIXME: execute the full command to make this compat flag unnecessary
			UseEntrypoint: true,
```

PR #7950, opened 2024-07-17
ci: fix go linter

jedevc:
Noticed in https://github.com/dagger/dagger/pull/7872#discussion_r1680759963

There are two issues:
- Some errors have an empty "Severity" - not entirely sure why, but some of the code was handling this correctly, but Assert wasn't.
- The `GOLANGCI_LINT_CONFIG` doesn't seem to have been being respected, I can't find any reference to it. So we can just use the suggested use of from the docs.

### DIFF
--- a/dev/golangci/lint.go
+++ b/dev/golangci/lint.go
@@ -103,6 +103,7 @@ func (run LintRun) WarningCount(ctx context.Context) (int, error) {
 
 // Return a JSON report file for this run
 func (run LintRun) Report() *dagger.File {
+	home := "/root"
 	cmd := []string{
 		"golangci-lint", "run",
 		"-v",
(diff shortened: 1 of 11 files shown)


dev/golangci/lint.go
```diff
@@ -54,7 +54,7 @@ func (run LintRun) Assert(ctx context.Context) error {
 		summaries []string
 	)
 	for _, iss := range issues {
-		if iss.Severity != "error" {
+		if !iss.IsError() {
```
shykes, line 57: Does this actually fix the issue?

jedevc: Reworked this entirely, now the report is normalized right after parsing.

shykes:
I am trying to test this, but golangci-lint just runs and runs, and eventually times out. Can anyone else reproduce?

```console
dagger call \
 -m github.com/dagger/dagger/dev@pull/7950/merge \
 --source https://github.com/dagger/dagger#pull/7950/merge \
 engine \
 lint
```

shykes (approved):

Issue #7954, opened 2024-07-17
Jobs hang strangely after upgrading CI to v0.12.0


jedevc:
After applying #7904, we've seen jobs hang weirdly.

See an example:
- https://github.com/dagger/dagger/actions/runs/9970573229/job/27549768418?pr=7936#step:3:89
- https://dagger.cloud/dagger/traces/6573dc5045a64c4d4e20b7e8b521b76a

Side note: somehow, I did reproduce locally, not sure how, I can't get it to happen again :scream: 

Something is *bizarrely* wrong here - the hanging at the end feels reminiscent of a telemetry draining issue from where we saw them before.

The root cause might be that we never see the end of `installing module` and `initialize` - even though cloud is convinced it saw the span end. Potentially, this isn't the most accurate though, the plain progress could potentially be deceiving us here (though it's hard to tell how).

A couple notes:
- Something is generally a little weird with the new spans in https://github.com/dagger/dagger/pull/7881.
	- These seem odd, but don't seem like the cause of the issue at least.
- During the execution of `asModule`, we do get an error: `! process "codegen --output /src --module-context-path /src/bass --module-name bass-sdk --introspection-json-path /schema.json" did not complete successfully: otel tcp proxy listen: context canceled: context canceled`
	- This is strange - because the parent span seems to succeed fine.
	- This span also appears in dagger cloud: https://dagger.cloud/dagger/traces/dbc42634c596df5e871f481e1012c698?span=d8f2bfca58033ca2#4e608f9ebee0fb4d
	- This fails? How is the pipeline able to continue?

cc @vito @sipsma

jedevc:
Hangs should be mitigated by https://github.com/dagger/dagger/pull/7966, but the source of the `otel tcp proxy listen: context canceled: context canceled` is still unknown.

jedevc:
We're not seein gthese hangs,

PR #7955, opened 2024-07-17
chore: ensure spans during module initialization make sense

jedevc:
These were previously able to overlap / end multiple times, which is confusing. This was introduced in #7881 in v0.12.0.

Maybe this isn't the cause of *all* the weird failures in https://github.com/dagger/dagger/issues/7954, but this isn't quite right. 

The `resolveSpan` could be ended multiple times.

### DIFF
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -394,20 +394,19 @@ func (fc *FuncCommand) initializeModule(ctx context.Context) (rerr error) {
 	defer telemetry.End(span, func() error { return rerr })
 
 	resolveCtx, resolveSpan := Tracer().Start(ctx, "resolving module ref", telemetry.Encapsulate())
-	defer telemetry.End(resolveSpan, func() error { return rerr })
 	modConf, err := getDefaultModuleConfiguration(resolveCtx, dag, true, true)
+	telemetry.EndNow(resolveSpan, err)
 	if err != nil {
 		return fmt.Errorf("failed to get configured module: %w", err)
 	}
 	if !modConf.FullyInitialized() {
 		return fmt.Errorf("module at source dir %q doesn't exist or is invalid", modConf.LocalRootSourcePath)
 	}
-	resolveSpan.End()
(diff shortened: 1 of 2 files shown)


cmd/dagger/functions.go
```diff
@@ -394,20 +394,19 @@ func (fc *FuncCommand) initializeModule(ctx context.Context) (rerr error) {
 	defer telemetry.End(span, func() error { return rerr })
 
 	resolveCtx, resolveSpan := Tracer().Start(ctx, "resolving module ref", telemetry.Encapsulate())
-	defer telemetry.End(resolveSpan, func() error { return rerr })
```
vito, line 397: Hmm I think this was fine before, but I can see how it's a little confusing.

Calling `End` multiple times on a span is fine - the later calls will be no-ops. The intent was to complete the span with either of the errors following the `getDefaultModuleConfiguration` too. So this pattern is a bit like having a `defer foo.Close()` as a safety net along with a `foo.Close()` in the happy path.

sdk/go/telemetry/span.go
```diff
@@ -43,3 +43,12 @@ func End(span trace.Span, fn func() error) {
 	}
 	span.End()
 }
+
+// EndNow is equivalent to End, but doesn't take a function.
+func EndNow(span trace.Span, err error) {
```
vito, line 48: I do like this though. Also thought about changing `End` to take a `*error` as its second arg, instead of a whole closure, which should be enough for the same basic use case (capturing `rerr`). But both are a little crufty.

cmd/dagger/functions.go
jedevc: Yeah, okay - I think that makes sense! I'll close this one then, I think it's unlikely that this one is actually causing any issues then.

sdk/go/telemetry/span.go
jedevc: Well I'll keep this bit :)

PR #7956, opened 2024-07-17
fix: allow context lock renewal in plain progress

jedevc:
Context locks were introduced (in #7272) as a way to prevent needless switching between two spans: e.g. suppose that span A and span B were running at the same time producing logs. If they both produce high volumes of logs, we want to ensure that we switch between them at a "reasonable" rate, and not switch back and forth immediately as they become available (which is significantly harder to read).

However, this initial implementation was slightly flawed: once a span acquired a lock, it would only be able to hold it for a maximum of a second - after that, another span would be immediately able to acquire the lock, even if it was actively being used.

This caused a very concrete issue for us:

	86  :   analyzing module
	86  :   analyzing module DONE [0.6s]
	6   : initialize DONE [22.8s]

	87  : prepare
	87  : prepare DONE [1.8s]

	88  : daggerDev(
	88  :     source: ModuleSource.resolveDirectoryFromCaller(path: ".", viewName: "default"): Directory!
	88  :   ): DaggerDev!
	88  : daggerDev DONE [1.2s]

	89  : DaggerDev.sdk: DaggerDevSdk!
	89  : DaggerDev.sdk DONE [0.6s]

	90  : DaggerDevSdk.go: DaggerDevGoSdk!
	90  : DaggerDevSdk.go DONE [0.0s]

	91  : DaggerDevGoSdk.lint: Void
	91  :   ...
	91  :   ...
	91  :   ...

	6   : initialize
	38  :   Module.withSource(
	38  :       source: ModuleSource.withName(name: "golangci"): ModuleSource!
	38  :     ): Module!
	202 :     Directory.diff(
	202 :         other: Container.directory(path: "/src"): Directory!
	202 :       ): Directory!
	203 :     diff (blob://sha256:de48f4cbe2e8e802e715f21f0b242caf2ecfe1b19ce7cb0d3d59d539b8a09e61) -> (exec codegen --output /src --module-context-path /src/dev/golangci --module-name golangci --introspection-json-path /schema.json)
	203 :     diff (blob://sha256:de48f4cbe2e8e802e715f21f0b242caf2ecfe1b19ce7cb0d3d59d539b8a09e61) -> (exec codegen --output /src --module-context-path /src/dev/golangci --module-name golangci --introspection-json-path /schema.json) DONE [0.7s]
	202 :     Directory.diff DONE [0.0s]
	204 :     Directory.withDirectory(
	204 :         directory: Directory.diff(
	204 :           other: Container.directory(path: "/src"): Directory!
	204 :         ): Directory!
	204 :         path: "/"
	204 :       ): Directory!
	204 :     Directory.withDirectory DONE [0.0s]

	91  : DaggerDevGoSdk.lint: Void
	91  :   ...
	91  :   ...
	91  :   ...

(see https://github.com/dagger/dagger/actions/runs/9973149598/job/27557889642#step:3:258)

This example occurs because the `initialize` span finishes rendering - but before it is able to print it's final message, the next span appears! This next span is able to acquire the lock, and we start with that. But eventually, `initialize` gets the lock back, and finishes printing, interrupting the currently ongoing span.

This problem is fixed by adding the concept of lock re-acquiry. If a span that holds the lock attempts to render it's context, it re-acquires the lock, refreshing the lock's timestamp, so it can continue to hold it. This prevents the above case, since `initialize` is still attempting to complete, and we can time from the *last* usage instead of the *first* usage.

There is one issue with the above though - a very noisy span could easily stifle the results of a span beneath it, which is annoying. To attempt to prevent this, we impose a max possible lock time (10 seconds) - this should be sufficient to avoid the issue we see in `initialize` and similar, but should be low enough to avoid the noisy span problem.


jedevc:
I found this and tried to fix this as part of https://github.com/dagger/dagger/issues/7954, but I don't actually think the issue is related.

helderco (approved):

PR #7958, opened 2024-07-17
feat: add vim gg/G movement operators

jedevc:
As mentioned :tada: :tada:

@vito I'm sure you'll love this, but also now, I want the vim fold operators :laughing: `zo`, `zc`, all that.

### DIFF
--- a/dagql/idtui/frontend_pretty.go
+++ b/dagql/idtui/frontend_pretty.go
@@ -655,6 +655,7 @@ func (fe *frontendPretty) update(msg tea.Msg) (*frontendPretty, tea.Cmd) { //nol
 		return fe, nil
 
 	case tea.KeyMsg:
+		lastKey := fe.pressedKey
 		fe.pressedKey = msg.String()
 		fe.pressedKeyAt = time.Now()
 		switch msg.String() {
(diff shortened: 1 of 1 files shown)


dagql/idtui/frontend_pretty.go
```diff
@@ -687,8 +688,10 @@ func (fe *frontendPretty) update(msg tea.Msg) (*frontendPretty, tea.Cmd) { //nol
 		case "home":
 			fe.goStart()
 			return fe, nil
-		case "end", " ":
+		case "end", "G", " ":
 			fe.goEnd()
+			fe.pressedKey = "end"
```
vito, line 693: The alternative to faking these out is to list `G` in the keys in the keymap, but it's not like that's much better since it's duplicative.

Someday maybe we should probably refactor into a data structure that drives both the UI and the key interpretation.

dagql/idtui/frontend_pretty.go
```diff
@@ -726,10 +729,20 @@ func (fe *frontendPretty) update(msg tea.Msg) (*frontendPretty, tea.Cmd) { //nol
 			fe.zoomed = fe.focused
 			fe.recalculateViewLocked()
 			return fe, nil
-		default:
-			return fe, nil
 		}
 
+		switch lastKey { //nolint:gocritic
+		case "g":
+			switch msg.String() {
```
vito, line 736:
```suggestion
			switch msg.String() { //nolint:gocritic
```

vito (approved):
Nice! Approving in principle but needs another `nolint`.

PR #7960, opened 2024-07-17
fix: print current directory without parameter expansion

vmaffet:
When getting a terminal in Debian based containers, the working directory does not show up correctly:
image

This is especially visible when using non alpine golang or debian images.

To reproduce simply do `dag.Container().From("golang").Terminal()`

This change in PS1 does not use parameter expansion but `pwd` and `sed` to replicate the same behavior.


jedevc (approved):
Thanks :tada:

I don't particularly like that this has to execute `sed` in the environment (what if it's not present for some bizarre reason?), but I can't think of a better way to solve this.

If we had *replacement* in pattern expansion in dash, then we could, but dash doesn't support that (since it's not part of the unix spec).

I don't think it's tooooo wild to assume `sed` is around if we have a unix shell that's reading `PS1`. That said if you have any clever ideas for how we could avoid calling it out to it, I'm all ears :cry:

vito (approved):

PR #7962, opened 2024-07-17
ci: update DAGGER_CLOUD_TOKEN

jedevc:
Updates our special telemetry-only token to use the new format, so we don't get *constantly* prompted to rotate our creds.

Thanks @marcosnils for the token :tada:


jedevc:
*excellent* 
image

PR #7970, opened 2024-07-18
docs: doc support for unix socket args

sipsma:
Support for unix sockets as function args was merged here: https://github.com/dagger/dagger/pull/7804 so it will be in the upcoming release.

I created a tiny demo module that I reference in the docs; it currently just lets you pass a host docker unix sock as an arg and then calls "docker version" on it: https://github.com/sipsma/daggerverse/blob/c19ee68f84b228942bac8047a9d8337fc22a3479/docker-client/main.go

**Should not be deployed until next release (0.12.1)**

### DIFF
--- a/docs/current_docs/manuals/user/host/host-services.mdx
+++ b/docs/current_docs/manuals/user/host/host-services.mdx
@@ -2,11 +2,13 @@
 slug: /manuals/user/host-services
 ---
 
-# Host Services Access
+# Connecting to Host Services
 
 Dagger Functions do not have access to the network of the host you invoke the Dagger Function from (i.e. the host you execute a CLI command like `dagger call` from). Instead, host network services need to be explicitly passed when executing `dagger call`.
 
-To pass host network services as arguments when invoking a Dagger Function, specify them in the form `tcp://<host>:<port>`.
+## TCP and UDP Services
+
+To pass host TCP or UDP network services as arguments when invoking a Dagger Function, specify them in the form `tcp://<host>:<port>` or `udp://<host>:<port>`.
 
 Assume that you have a PostgresQL database running locally on port 5432, as with:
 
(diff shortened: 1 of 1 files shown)


docs/current_docs/manuals/user/host/host-services.mdx
```diff
@@ -2,7 +2,7 @@
 slug: /manuals/user/host-services
 ---
 
-# Host Services Access
+# Host IP Services Access
```
shykes, line 5:
```suggestion
# Connecting to Host Services
```

docs/current_docs/manuals/user/host/host-services.mdx
```diff
@@ -2,7 +2,7 @@
...
 
 Dagger Functions do not have access to the network of the host you invoke the Dagger Function from (i.e. the host you execute a CLI command like `dagger call` from). Instead, host network services need to be explicitly passed when executing `dagger call`.
```
shykes, line 7:
```suggestion
Dagger Functions do not have access to the network of the host you invoke the Dagger Function from (i.e. the host you execute a CLI command like `dagger call` from). Instead, host network services need to be explicitly passed when executing `dagger call`.

## TCP and UDP services
```

docs/current_docs/manuals/user/host/host-services.mdx
```diff
@@ -19,3 +19,15 @@ Here is an example of passing this host service as argument to a PostgreSQL clie
 ```shell
 dagger -m github.com/kpenfound/dagger-modules/postgres@v0.1.0 call client --db=postgres --user=postgres --password=postgres --server=tcp://localhost:5432
 ```
+
+# Host Unix Socket Access
```
shykes, line 23:
```suggestion
# Unix Sockets
```

shykes (approved):
Thanks for adding the docs page! A few nits on structure, typically you want a single H1 title, then H2 titles below that. I also tweaked the wording of titles.

sipsma:
@shykes thanks! Suggestions all make sense to me, updated with those

PR #7973, opened 2024-07-18
chore: encourage multiline changie bodies

jedevc:
This is part of an effort to start writing better changelogs! Just added some people to review who I've chatted to about this before :tada:

Changie supports multiline bodies, that looks something like this:

image

The idea is to have more extensive changelogs as we go forward, essentially *trying* to follow something like a "git commit" format:
- The first line should be a summary of the change
- Any other details afterward can provide more details about the contents of the change, e.g. for breaking changes, information on how to migrate

After allowing these multiline bodies, we also need to update the output format so that the author and PR appear on the first line, while the details appear below - essentially the format we used for v0.12.0 for the breaking changes (which I quite liked): https://github.com/dagger/dagger/blob/1a28a7a467b25b863a4d62b6d18c18e24fd01b1e/.changes/v0.12.0.md#L10-L11

### DIFF
--- a/.changie.yaml
+++ b/.changie.yaml
@@ -5,7 +5,15 @@ changelogPath: CHANGELOG.md
 versionExt: md
 versionFormat: '## {{.Version}} - {{.Time.Format "2006-01-02"}}'
 kindFormat: '### {{.Kind}}'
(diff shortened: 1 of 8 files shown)


vito (approved):
Nice! Always thought one line was way too short. Time to write some novels.

jedevc:
Woop, got at least one person who likes this :tada: Going to merge.

PR #7975, opened 2024-07-18
Fix broken tests that got lost in the noise

jedevc:
While merging a lot of the things for v0.12.1, a few things got lost in the noise:

- `TestContainer/TestInsecureRootCapabilitesWithService` does docker network access, and so should be given the host's docker config if it can be found.
- `TestModuleDevelopVersion` has some duplication with `TestModuleSchemaVersion` - we can remove those, since all the permutations are tricky to handle in the develop test.


jedevc:
Tests are failing due to the known flake #7900.

sipsma (approved):

PR #7981, opened 2024-07-18
Change default module source directory from ./dagger to .

### LINKED ISSUE #6963: SDKs should generate files in current working directory by default, not ./dagger

shykes:
Fixes #6963

When installing an SDK in a module, either with `dagger init --sdk` or first call to `dagger develop --sdk`, the default source directory will be `.` instead of `dagger/`.

Note that the `source` field in `dagger.json` is *not* created by default. If we want `"source': "."` to be explicitly added, I need help modifying this PR to achieve that behavior. cc @sipsma as you know that explicit behavior is usually your preference.

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -54,7 +54,7 @@ var (
 const (
 	moduleURLDefault = "."
 
-	defaultModuleSourceDirName = "dagger"
+	defaultModuleSourceDirName = "."
 )
 
 func init() {
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -87,20 +87,15 @@ func init() {
 }
 
 var moduleInitCmd = &cobra.Command{
-	Use:   "init [options] [path]",
+	Use:   "init [options] [PATH]",
 	Short: "Initialize a new Dagger module",
-	Long: `Initialize a new Dagger module in a local directory.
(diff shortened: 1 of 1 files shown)


helderco:
@TomChv already worked on this in:
- https://github.com/dagger/dagger/pull/7824

TomChv:
Yup, I already made a PR with the change, just took some time to fixe all the tests depending on the location of the generated file

shykes:
Duplicate of #7824

PR #7986, opened 2024-07-19
pipelines: lint: more readable traces

shykes:
Use custom spans to make traces more readable

### DIFF
--- a/dev/docs.go
+++ b/dev/docs.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 
+	"go.opentelemetry.io/otel/codes"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/dagger/dagger/dev/internal/dagger"
@@ -51,11 +52,18 @@ func (d Docs) Server() *dagger.Container {
 }
 
 // Lint documentation files
-func (d Docs) Lint(ctx context.Context) error {
+func (d Docs) Lint(ctx context.Context) (rerr error) {
 	eg, ctx := errgroup.WithContext(ctx)
 
 	// Markdown
-	eg.Go(func() error {
+	eg.Go(func() (rerr error) {
+		ctx, span := Tracer().Start(ctx, "lint markdown files")
+		defer func() {
+			if rerr != nil {
+				span.SetStatus(codes.Error, rerr.Error())
+			}
+			span.End()
+		}()
 		_, err := dag.Container().
 			From("tmknom/markdownlint:0.31.1").
 			WithMountedDirectory("/src", d.Dagger.Source()).
@@ -73,7 +81,14 @@ func (d Docs) Lint(ctx context.Context) error {
 		return err
 	})
 
-	eg.Go(func() error {
+	eg.Go(func() (rerr error) {
+		ctx, span := Tracer().Start(ctx, "check that generated GraphQL and CLI references are up-to-date")
+		defer func() {
+			if rerr != nil {
+				span.SetStatus(codes.Error, rerr.Error())
+			}
+			span.End()
+		}()
 		before := d.Dagger.Source()
 		after, err := d.Generate(ctx)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


grouville (approved):
LGTM ✅

dev/docs.go
```diff
@@ -52,11 +53,25 @@ func (d Docs) Server() *dagger.Container {
...
 // Lint documentation files
-func (d Docs) Lint(ctx context.Context) error {
+func (d Docs) Lint(ctx context.Context) (rerr error) {
+	ctx, span := Tracer().Start(ctx, "lint Dagger documentation")
+	defer func() {
+		if rerr != nil {
+			span.SetStatus(codes.Error, rerr.Error())
+		}
+		span.End()
+	}()
```
vito, line 63: I understand if you want to skip it to stay familiar with the raw OTel SDK, but there's a helper available for this provided by the Dagger Go SDK:

```go
import (
	"github.com/dagger/dagger/dev/internal/telemetry"
)
```

```go
```

We're not skipping `Tracer()` so this would be consistent with using the helpers, but yeah no big deal if you prefer not to lean on it.

dev/docs.go
```diff
@@ -74,11 +89,40 @@ func (d Docs) Lint(ctx context.Context) error {
...
-		return util.DiffDirectoryF(ctx, d.Dagger.Source(), d.Generate, generatedSchemaPath, generatedCliZenPath)
+	eg.Go(func() (rerr error) {
+		ctx, span := Tracer().Start(ctx, "check that generated GraphQL reference documentation is up-to-date")
+		defer func() {
+			if rerr != nil {
+				span.SetStatus(codes.Error, rerr.Error())
+			}
+			span.End()
+		}()
+		return util.DiffDirectoryF(ctx, d.Dagger.Source(), d.Generate, generatedSchemaPath)
```
vito, line 100: Could be worth pushing the span setup into `util.DiffDirectoryF` by giving it a required arg for the name of the thing being diffed, like `"GraphQL reference documentation"`, and using that to construct the span name + a descriptive error message.

dev/sdk_go.go
```diff
@@ -19,21 +20,49 @@ type GoSDK struct {
...
 // Test the Go SDK
-func (t GoSDK) Test(ctx context.Context) error {
+func (t GoSDK) Test(ctx context.Context) (rerr error) {
+	ctx, span := Tracer().Start(ctx, "test the Go SDK")
+	defer func() {
+		if rerr != nil {
+			span.SetStatus(codes.Error, rerr.Error())
+		}
+		span.End()
+	}()
```
vito, line 65: Personally think this is a step too far, since this span is 1:1 with the original function call, and will add an extra step to expand while jumping into its contents[^1]. I wouldn't expect every user to do with this with every function, so I don't think we should either.

Agree that we should be using function descriptions somehow in Cloud, but don't think they're a total substitue for the function call either, since the args may be significant. Maybe we could always show the description once expanded and/or in a tooltip?

[^1]: Granted you'd probably be just jumping straight to the error in practice, so maybe that doesn't matter.

dev/docs.go
shykes: I have a separate PR that removes `DiffDirectoryF` altogether in favor of `dirdiff` module. #7877 

I will look into this idea for the successor. But the span would be less expressive.

dev/sdk_go.go
shykes: Yeah I agree

shykes: > Agree that we should be using function descriptions somehow in Cloud, but don't think they're a total substitue for the function call either, since the args may be significant.

I think a toggle would be amazing. Toggle between "text view" (only show text descriptions, either from custom spans or scraped from comments) and "function view" (function calls only, no custom spans). Implemented in both TUI and web. I think it could be magical.

shykes:
I am back online, it's a little depressing that this hasn't been picked up, and now I have to parse through massive merge conflicts.

dev/docs.go
shykes: @vito any chance the helper could receive the error value, so I can just pass it through? Would save me 1 line in each diff.

Or even better, just pass it through for me: `defer telemetry.End(span)`

vito: @shykes The trick is the helper needs the `error` value at the time the function returns, so we need to capture a reference to it so the `defer` doesn't just call it with `nil` (`rerr` value at the time).

Right now it's a closure, but I regret that now - I think a pointer would also work:

```go
defer telemetry.End(span, &rerr)
```

I don't think there's any way to magically grab the error value in Go.

PR #7987, opened 2024-07-19
cli: improve usage messages

shykes:


grouville (approved):
Love it ! 👼  Needs an update on the references though I think 👀

cmd/dagger/module.go
```diff
@@ -303,19 +297,16 @@ var moduleInstallCmd = &cobra.Command{
...
 	Use:   "develop [options]",
-	Short: "Setup or update all the resources needed to develop on a module locally",
-	Long: `Setup or update all the resources needed to develop on a module locally.
-
-This command regenerates the module's code based on dependencies
-and the current state of the module's source code.
-
-If --sdk is set, the config file and generated code will be updated with those values reflected. It currently can only be used to set the SDK of a module that does not have one already.
+	Short: "Prepare a local module for development",
+	Long: `This command ensures that a module's SDK is installed, configured, and all its files re-generated. You should run it before every interaction between local tools and a module's source code: IDE, linter, source control, etc.
```
levlaz, line 301: > You should run it before every interaction between local tools and a module's source code: IDE, linter, source control, etc.

This part feels confusing, I rarely use `dagger develop` and don't seem to run into issues, because I feel like it does some stuff automatically.

levlaz (approved):
All great improvements, I have one non blocking question.

grouville:
Pushed a patch for the failing test + docs. Current failures seem unrelated to PR

cmd/dagger/module.go
```diff
@@ -78,29 +78,24 @@ func init() {
...
+	moduleDevelopCmd.Flags().StringVar(&developSourcePath, "source", "", "Source directory used by the installed SDK. Defaults to module root")
+	moduleDevelopCmd.Flags().StringVar(&licenseID, "license", defaultLicense, "License identifier to generate. See https://spdx.org/licenses/")
 	moduleDevelopCmd.Flags().StringVar(&compatVersion, "compat", modules.EngineVersionLatest, "Engine API version to target")
 	moduleDevelopCmd.Flags().Lookup("compat").NoOptDefVal = "skip"
 	moduleDevelopCmd.PersistentFlags().AddFlagSet(moduleFlags)
 }
 
 var moduleInitCmd = &cobra.Command{
-	Use:   "init [options] [path]",
+	Use:   "init [options] [PATH]",
```
helderco, line 90: We've adopted the lowercase convention. See:
- https://github.com/dagger/dagger/pull/7143

```suggestion
	Use:   "init [options] [path]",
```

cmd/dagger/module.go
shykes: > I rarely use `dagger develop` and don't seem to run into issues, because I feel like it does some stuff automatically when doing `dagger call` or `dagger install`

Yes `dagger install` will re-generate bindings. `dagger call` doesn't do anything though.

> I used it recently to do the 0.12 port, I don't remember the last time I used it.

Noted, I will try to downgrade the FUD :)

> > 3. Update the target engine version if needed
> 
> Is this because this behavior is new and we are now recommending that people use this command on a regular basis?

Not sure what you mean. What do you mean by "this" and "this behavior"?

cmd/dagger/module.go
shykes: Good to know, thanks

helderco (approved):

Issue #7979, opened 2024-07-18
Flaky test: `TestGoExtendCore` - `cannot define methods on objects from outside this module`


sipsma:
Started seeing this one a bunch recently:
* Cloud trace: https://dagger.cloud/dagger/traces/ce4f383ede89e9bf39d00fe89aacd444?span=087b98784bf05440#909ebe7a106b977f
* GHA job: https://github.com/dagger/dagger/actions/runs/9998434018/job/27637312803?pr=7978

From what I can tell looking in the engine logs above it, it *does* appear the expected error is likely getting hit, it's just not getting relayed to the clients as expected.

cc @vito @jedevc no initial clues what changes would have resulted in this (if any)

sipsma:
@vito mentioned this should be fixed by otel draining re-work: https://github.com/dagger/dagger/pull/7989#issuecomment-2239244084

jedevc:
Going through old CI issues, I can't recall having seen this one at all recently.

Guess the OTEL rework has been successful :tada:

cwlbraa:
[pretty sure this is still happening](https://dagger.cloud/dagger/traces/7b25518867c0f2d2531570db190bc37e#L580-a86f3bb556663896)

PR #7989, opened 2024-07-19
tests: add retries to waiting for log output

sipsma:
Possibly fixes https://github.com/dagger/dagger/issues/7979 (to be confirmed in CI since that's the only place I can get it to happen).

I also am not sure yet if this is handling expected behavior more robustly vs. just avoiding a bug. If it's just avoiding a bug we shouldn't go with it.

Basically, I noticed that these flakes appear to just be cutting off the output while it's still being received. If you look in the logs you can see most of the expected output there but not all of it, which results in the error message expected not showing up.

cc @vito @jedevc on confirming whether it's expected that if you close a client while output is still being received then it is expected you might be cutting off some of the output, since you've worked on that whole are more than I

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -1915,9 +1915,9 @@ func (c *Container) Echo(ctx context.Context, msg string) (string, error) {
 			With(daggerQuery(`{container{from(address:"` + alpineImage + `"){echo(msg:"echo!"){stdout}}}}`)).
 			Sync(ctx)
 		require.Error(t, err)
-		require.NoError(t, c.Close())
-		t.Log(logs.String())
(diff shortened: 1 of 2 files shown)


vito:
Yeah I think this would just be working around a bug - I'd leave it and see if my draining rework fixes it. The expectation is that if all work is done prior to `client.Close` - which it is in this case (the `Sync` above it) - that calling `Close` synchronously flushes all output. So you shouldn't need to poll after closing it.

sipsma:
> Yeah I think this would just be working around a bug - I'd leave it and see if my draining rework fixes it. The expectation is that if all work is done prior to `client.Close` - which it is in this case (the `Sync` above it) - that calling `Close` synchronously flushes all output. So you shouldn't need to poll after closing it.

Okay good to know, thanks!

PR #7991, opened 2024-07-19
DNM: debug PythonProjectLayout test

helderco:
See:
- https://github.com/dagger/dagger/issues/7662#issuecomment-2238844398

### DIFF
--- a/core/integration/module_python_test.go
+++ b/core/integration/module_python_test.go
@@ -275,21 +275,15 @@ build-backend = "poetry.core.masonry.api"
 		t.Run(fmt.Sprintf("%s/%s", tc.name, tc.path), func(ctx context.Context, t *testctx.T) {
 			c := connect(ctx, t)
 
-			conf := tc.conf + fmt.Sprintf(`
-# Test: %s / %s
-`, tc.name, tc.path)
-
 			modGen := daggerCliBase(t, c).
-				With(fileContents("pyproject.toml", conf)).
-				With(fileContents(tc.path, fmt.Sprintf(`
+				With(fileContents("pyproject.toml", tc.conf)).
+				With(fileContents(tc.path, `
 from dagger import function
-
-# Test: %s / %s
 
 @function
 def hello() -> str:
     return "Hello, world!"
-`, tc.name, tc.path),
+`,
 				)).
 				With(daggerInitPython())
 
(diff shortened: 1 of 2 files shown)


core/integration/module_python_test.go
```diff
@@ -275,17 +275,33 @@ build-backend = "poetry.core.masonry.api"
...
+			conf := tc.conf + fmt.Sprintf(`
+# Test: %s / %s
+`, tc.name, tc.path)
+
+			modGen := daggerCliBase(t, c).
+				With(fileContents("pyproject.toml", conf)).
+				With(fileContents(tc.path, fmt.Sprintf(`
 from dagger import function
 
+# Test: %s / %s
```
helderco, line 287: @sipsma, maybe this is causing the test to pass by making the contents different for every subtest?

Some tests have the same `pyproject.toml`:
- setuptools
  - `src/main/__init__.py`
  - `src/main.py`
- poetry
  - `src/main/__init__.py`
  - `src/main.py`
  - `main/__init__.py`
  - `main.py`

Would be useful to know if the flake happened in other subtests or just these.

I'm going to remove the addition to the contents to see if it triggers the flake.

helderco: Hatch failed in this run: https://dagger.cloud/dagger/traces/dfaba0c8512cff8ef05e7bee37018c82?span=819126d8189f7b28

But I wonder if the reason is because it's getting a `pyproject.toml` from one of the above or something. Just submitted a PR to merge into main. It's the same as this one, but without the `--verbose` flag in `uv`.

helderco: Actually, `setuptools` have different contents in those subtests:

```diff
[tool.setuptools]
- packages = ["main"]
+ py-packages = ["main"]
```
So it's just `poetry` that has the same contents.

helderco:
@sipsma, finally got a hit here: https://dagger.cloud/dagger/traces/d4bda71e6b5513a7c396f60c2bfd6919?span=a581973f3f6cdda6

This is the failing test case:

https://github.com/dagger/dagger/blob/d9a15262229f1dba0aa53dbd0216908438331995/core/integration/module_python_test.go#L114-L123

Inspecting these files show the correct contents:

https://github.com/dagger/dagger/blob/695ec7efb60ca34deb72f4bb2e72c1817fff0fc9/core/integration/module_python_test.go#L290-L300

Shows the correct contents:

```
==== pyproject.toml ====
[project]
name = "main"
version = "0.0.0"

==== src/main.py ====:
from dagger import function
@function
def hello() -> str:
    return "Hello, world!"
```

PR #7996, opened 2024-07-20
redo OTel engine => client pipeline

vito:
Replaces the live one-shot pub/sub approach with saving telemetry to a per-client database.

Previously:

* The engine configured `PubSub` as the global tracer provider for all telemetry in the engine.
* A SpanProcessor would carefully annotate each span with the originating client ID.
* Clients would subscribe to their own OTLP traffic via a side channel gRPC service.
* Everything would export to `PubSub`, which would send relevant spans to all affected clients, including keeping track of parents
* `PubSub` would drain when all spans for a client have completed and all stdio logs saw EOF.

This model forced every decision to be made in "one shot" - during exporting. There was no persistence, so the only way to tell we got all of a client's data is to watch for its activity to stop, which is fraught with peril.

Now:

* The engine's globally configured OTel stack only captures engine-wide telemetry, i.e. everything but client traffic.
* Each cilent has its own `TracerProvider` and `LoggerProvider` installed for each request, which writes to the client's own SQLite database.
* Each client DB stores OTLP payloads with autoincrementing IDs so it's easy to tell the ordering of events (important for subscribing later).
* Child clients also write to their parent client's databases (they're just additional SpanProcessors in the client's singular TracerProvider).
* Child clients now subscribe over HTTP (no more gRPC) using SSE and poll for new data every <nearly-immediate-interval=100ms> using the autoincrementing event ID as a cursor.
* When a client shuts down we stop polling for new data.

Now that there's persistence we can decouple the event writing from the event reading and just read all events until conceptual EOF when a client shuts down. No more draining on spans completing. No more hanging waiting for new events. Just a pretty straightforward state machine.

TODO:

* [x] Garbage collect client DBs
* [x] Figure out what's causing orphaned gRPC service spans

---

Worth mentioning:

* Could maybe use this per-client database for more things in the future. It's kinda neat.
* Client databases migrate on initialization, so we don't need to worry about migrating old data.
  * Curiously, clients will be initialized multiple times in some scenarios, so the migration just does `CREATE TABLE IF NOT EXISTS`.
* The engine will hold more files open now, since there's a DB per-client, and subscribers hold a separate connection from writers. I've bumped the open files limit in `dagger-entrypoint.sh` to match the value we set in our production CI runners (`1048576`). Otherwise it was picking up the super low default of `1024` on my system which made tests fail.
* The Engine `test` and `testdev` jobs now redirect their `dagger call` output to a file and upload it as an artifact, to work around https://github.com/actions/runner/issues/1031
* The SSE OTLP endpoint could likely be a GraphQL subcription in the future, but didn't want to boil the ocean in this PR.


vito:
Currently investigating test instability, will keep a rolling log in this comment.

I'm not sure if these test flakes are even related to these changes, but don't want to risk letting more slip by.

I'll be focusing on `testdev` since it uploads both the `dagger call` and the engine logs as GHA artifacts.

### `testdev` timing out in GHA despite passing in Cloud

* Currently assuming this is mostly related to https://github.com/actions/runner/issues/1031 because it would reliably hang before but now that I've redirected output to a file it succeeds like normal (modulo other flakes).
  * `TestLargeObjectFieldVal` definitely logs a 32MB string, but I'm not sure why this would only affect this PR, and/or why it would only affect `testdev`.

The only hang I've seen after adding the output redirect is this run:

* https://github.com/dagger/dagger/actions/runs/10208915003/job/28246099907?pr=7996
  * But I can't tell easily why it got stuck. The same test (`TestDaggerListen`) passes 100 times in a row for me. Currently trying to catch it in the act so I can send a SIGQUIT.
  * This is also different from the long-lines theory since in this case the run itself took 30m, whereas previously the Cloud trace would be ~10-15m and GHA would time out after 30m.

### `TestDaggerDevelop`

#### `cannot find default codegen DaggerObject interface`

* https://github.com/dagger/dagger/actions/runs/10208915003/job/28269984693
  * https://dagger.cloud/dagger/traces/dd543dcaf22203e61eabcf74c7453442?span=3b9dd82b03c86330#3b9dd82b03c86330:L23

Saw this happen in another run too, not sure what GHA job this correlates to, sorry:

* https://dagger.cloud/dagger/traces/1efe90a49e5643c2afdc3b7205bdf482?span=7ba1f90d34396539#96851fa0f5d1206e

#### `failed to get requester session: session for "jac9cuigsokznjs9z4c0hnfpv" not found`

* https://dagger.cloud/dagger/traces/7907224c54774ad663a885a347f1badb?span=2c7507d92a263361

vito:
Follow-up: the only unique flake was `TestDaggerListen` which seemed to be caused by a race condition with subscribing to telemetry while submitting other client requests. I originally wanted to make subscribing synchronous (sync attach, async stream) but that caused this test to hang, even though SSE chunked responses should be able to immediately return the headers. As a band-aid I made it fully async, but that was causing a lot of unpredictability (and I suspect the hang), so I wanted to try making it sync again.

I found the root cause for the hang and fixed it in ce1fb22c93a1bb4cd792ecf986ba684bdce72072 ("synchronously subscribe, fix client buffering") - now the `io.Copy` in the engine client `ServeHTTP` that proxies the response will additionally `(http.Flusher).Flush()` after every write. With that change, synchronously subscribing works - we get the headers back, and the initial `subscribed` event, and block like normal.

After making that change I haven't seen the test hang; every `testdev` has been green. Knock on wood.

I can spend some more time updating the PR description, but I'm marking this "ready for review" as I'm running out of TODOs.

vito:
Noticed there's a huge difference between `test all` and `testdev` log volume, looking into that now. If it turns out there's a bug there I can undo the GHA logs=>artifacts change.

.dagger/config.go
```diff
@@ -36,6 +36,10 @@ if [ -f /sys/fs/cgroup/cgroup.controllers ]; then
 		> /sys/fs/cgroup/cgroup.subtree_control
 fi
 
+# expect more open files due to per-client SQLite databases
+# many systems default to 1024 which is far too low
+ulimit -n 1048576
```
jedevc, line 41: This is 1024 * 1024 - curious, is this pretty much arbitrary?

engine/clientdb/span.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	return telemetry.InstrumentationScopeFromPB(is)
+}
+
+// Ended returns whether the span has ended
+func (ros *readOnlySpan) Ended() bool {
+	return ros.DB.EndTime.Valid
+}
+
+// type ReadOnlySpan interface {
```
jedevc, line 213: :eyes:

engine/clientdb/dbs.go
```diff
@@ -0,0 +1,115 @@
...
+	_ "modernc.org/sqlite"
+)
+
+type DBs struct {
+	Root string
+}
+
+// CollectGarbageAfter is the time after which a database is considered
+// garbage and can be deleted.
+const CollectGarbageAfter = time.Hour
```
jedevc, line 24: Did we change this? I seem to remember in our core sync we discussed that this was quite a bit longer, something like 48 hours? I can't quite remember, but the logic was that we want to avoid the case where a job might accidentally get cleaned up?

Or is that just not possible, and I'm totally misremembering?

jedevc (commented):
A few questions and nits, but overall this *feels* better, and a more resilient to weird failures :tada: That said, I think we should ideally let this soak in `main` for at least a day before cutting a release with it, so I'd probably suggest pushing the release back to Wednesday at least.

As discussed in discord, `dagger watch` likely breaks, but that's fine, we "Can fix it later, I think no one knows it exists" (@vito).

engine/buildkit/executor_spec.go
```diff
@@ -631,17 +629,24 @@ func (w *Worker) setupOTel(ctx context.Context, state *execState) error {
...
+	var destSession string
+	var destClientID string
+	if w.execMD != nil { // NB: this seems to be _always_ set
+		destSession = w.execMD.SessionID
+
+		// Send telemetry to the caller client, *not* the nested client (ClientID).
+		//
+		// If you set ClientID here, nested dagger CLI calls made against an engine running
+		// as a service in Dagger will end up in a loop sending logs to themselves.
+		destClientID = w.execMD.CallerClientID
```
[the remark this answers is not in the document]
vito: Unfortunately it just goes to one of them, almost certainly the first one. In fact we don't see any spans _at all_ from Buildkit in the second one.

I don't think Buildkit's `MultiSpan` does anything for us at the moment - it just picks the first:

https://github.com/moby/buildkit/blob/add7328b642f6cd7a0e1287420de505132a3d0c4/util/tracing/multispan.go#L8

Prior to switching to OTel Buildkit handled this very nicely for us - its progress stream happily multiplexed to anyone attaching to the same workloads.

Maybe we need to make MultiSpan actually be a MultiSpan?

vito (commented):
@jedevc Thanks for the review - think I addressed everything now!

re: shipping, I'm reasonably confident in this change but yeah seems sensible to give it another day to bake.

edit: ack, GitHub made this really hard to follow since I had a prior pending review 🤦‍♂️ - sorry

engine/buildkit/executor_spec.go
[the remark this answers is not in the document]
jedevc: Ahh, okay, that's kinda expected.

I started hacking while working on the pending/cached spans on https://github.com/moby/buildkit/compare/master...jedevc:buildkit:multi-span, maybe that's a place to start? Haven't actually tested that it really works, but in principle the code makes sense.

jedevc (approved):
:shipit: 

I think this probably deserves a small "fix" changelog?

vito:
@jedevc for sure - added! Will merge on green.

PR #8002, opened 2024-07-22
docs: revise for new --source=. default

vito:
follow-up to #7824 

Went for minimal changes here to maintain correctness. I _think_ each of these makes sense in context but could use a double check.


sipsma (approved):
LGTM, I do wonder if we should instead just have a single example of setting `--source` to a non-default value and otherwise use the default as much as possible, but agree with going for the minimal change atm and figuring that out in follow-ups.

PR #8012, opened 2024-07-23
ci: Publish nightly as `main/head`

gerhard:
So that we can reference it in our `runs-on` CI runner configurations.

Setting `DAGGER_COMMIT=head` will pull the latest CLI built from `main`. We are not templating the branch name on purpose. When we introduce release branches, we should revisit this. For now, we are keeping this as simple as possible for as long as possible.

After this gets merged:
- [ ] Update `https://dl.dagger.io/dagger/install.sh`

---
Tested after running it locally against the release destination replica:
image


gerhard:
Want to test this locally before marking **Ready for review**.

Early feedback appreciated!

vito (approved):

PR #8019, opened 2024-07-23
cli: propagate exit code from function

helderco:
A while ago in Discord, a user expected that `dagger call` exited with the same exit code of the test suite tool they were using in the function. I thought why not, and it's easy to add.


jedevc (approved):
I think this makes sense - the one thing to be aware of though is that it does potentially prevent us from using other exit codes ourselves in the CLI to signify types of failures.

However, I can't think of a reason right now why we'd really want to do that - we already have progress output for printing that level of detail.

PR #8022, opened 2024-07-23
fix dev module custom test running documentation

cwlbraa:
The existing dev module documentation instructs developers to run a test which doesn't exist. This PR updates the example to point at the renamed test, with the added side benefit of making it clear that many core/integration tests are subtests.


jedevc (approved):
Thanks! Good catch - we restructured the tests, I'd forgotten there was a reference to a test in here.

PR #8023, opened 2024-07-23
tests: restructure TestExitsZeroOnSignal to avoid timeout flake

sipsma:
This test flakes occasionally due to timeout errors (just enough to be annoying). [Looking at the traces](https://dagger.cloud/dagger/traces/5a3c0a61daf4b164dff6b580605eace2?span=53daff10b1926f68#e7124c9fee8125f1), it seems like when under heavy load the container import step just takes a while and brings it close enough to the 1 minute timeout to hit it during the test.

This updates the test case to first do a synchronous import of the engine container and only after that apply the timeout. If my understanding is correct that should ensure the existing highly generous 1 minute timeout is only being applied to the part of the test where we care about it.


jedevc (approved):
Very reasonable :tada:

PR #8027, opened 2024-07-24
fix: ensure dagger dev source gen works

jedevc:
The recent changes to dagger codegen (#7872) meant that locally generated `dagger.gen.go` files were overwriting the generated files. This was because `GeneratedContextDirectory` takes the entire context directory (including any local `dagger.gen.go` files), and applies it's changes on top. So if `dev/go/dagger.gen.go` was generated on an old version of dagger, then it would be applied as part of a later codegen step, e.g. from `sdk/typescript/runtime` :scream: 

The reason this happened is because we changed the dag construction to be more parallelized, instead of stacked - to do this properly though, we need to select each module directly, and compose them all together afterwards. Making it stacked is the right call - it speeds up execution - but in this form it wasn't quite right.

I needed access to the `module` variable while iterating and applying `WithDirectory`, so I additionally removed the complex parallelism that wasn't actually being used (since all of the operations are actually lazy) - as I suggested in https://github.com/dagger/dagger/pull/7872/files#r1682612665.

### DIFF
--- a/dev/main.go
+++ b/dev/main.go
@@ -6,7 +6,6 @@ import (
 	"context"
 
 	"github.com/dagger/dagger/dev/internal/dagger"
-	"golang.org/x/sync/errgroup"
 )
 
 // A dev environment for the DaggerDev Engine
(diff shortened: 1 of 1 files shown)


shykes:
@jedevc this is probably what I encountered here: https://discord.com/channels/707636530424053791/1003718839739105300/1263440723018252341

> @jed sorry to add to your worries, my "streamline codegen" PR is done, I believe it's implemented properly, but it seems to be surfacing some weird interference between codegen and compat mode, I think?

PR #8028, opened 2024-07-24
chore(dev): remove python modules from list of modules to source

helderco:
Python's dev module is no longer broken.

### DIFF
--- a/dev/main.go
+++ b/dev/main.go
@@ -96,8 +96,7 @@ func (dev *DaggerDev) Source() *dagger.Directory {
 		"dev/markdown",
 		"dev/shellcheck",
 		"sdk/elixir/runtime",
-		// FIXME: broken
-		// "sdk/python/dev",
+		"sdk/python/dev",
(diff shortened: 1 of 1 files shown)


jedevc:
Hm, `python lint` looks broken :confused:

helderco:
> Hm, `python lint` looks broken 😕

That's because the generated files from generating the module are now being included in the source that gets passed to the lint function.

What's the purpose of generating these modules into Source anyway? Maybe that's only for Go modules?

jedevc:
cc @shykes, this was added in #7872

The go linter needs this codegen so that the go code can be linted - otherwise, there's no `dagger.gen.go` and a bunch of things will be missing.

helderco:
Ok, so no need to include Python _dev_ then.

Same for TypeScript though, I don't think these are needed too:
- `sdk/typescript/dev`
- `sdk/typescript/dev/node`

\cc @TomChv

TomChv (approved):

Issue #6902, opened 2024-03-19
Module include/exclude unexpectedly impacts other local dependencies


sipsma:
Noticed this while working on https://github.com/dagger/dagger/pull/6857

Right now if you have two modules in a local repo checkout and there is a dependency between them, if
* Module A does an include of `*.foo`
* Module B does an exclude of `*.foo`

Then the exclude of Module B will override the include from Module A. The direction of the dependency between A and B doesn't matter, the end effect will be the same.

The problem is that module loading currently combines all the includes+excludes of local modules to one big list of includes+excludes, which results in unexpected interactions like this.

A related issue is that we currently dedupe these include/exclude paths using a map, which results in the end list of each being in a random order, which isn't correct as the order of a given include/exclude list actually matters.

---

Instead, while include+exclude should have interactions within the context of a single module, they should not impact others.

Unfortunately, I'm not seeing any super quick fixes here. If we load all the modules separately and then merge them together, we will end up reloading a lot of duplicate data potentially.

The only viable option I can think of so far that wouldn't annihilate performance would be to re-use the same underlying cache ref for the local module loading, but load each one one-at-a-time. After each load, we can look at the files that were loaded, compare against the include/exclude of the *next* module that we're going to load and update those includes/excludes to ensure that we don't delete any files that were loaded for the previous modules 😵‍💫 Hopefully there's some elegant/clever way to model this, but it seems like it may be a fountain of corner cases, so lots of tests will be needed.
* In the very long term, a new implementation of local dir loading that dedupes files across distinct cache refs (hardlink farm idea) would make this all much simpler

PR #8042, opened 2024-07-25
pipelines: merge dev module back into the root

shykes:
## Context

`-m dev` is just too cumbersome. Meanwhile in the latest design discussions, we are moving away from the concept of "dev module". See [this discord thread](https://discord.com/channels/707636530424053791/1260894336883626025).

Since the prospect of a "quick win" where a dev module pattern solves all our problems is less and less certain, there's no point in incurring the pain of the extra `-m dev` at this point.

## Summary

- Main Dagger module is now `github.com/dagger/dagger`
- No more `github.com/dagger/dagger/dev` (this breaks external references to these modules, which we intentionally discouraged because of anticipated changes like this one)
- Sub-modules moved to the root of the repo (we didn't have sub-modules before)
- The rest is in `.dagger` (better than `ci`)

## TODO before merging

Remaining tasks:

- [ ] fix CI *help wanted*
- [ ] remove references to `dev` in Github Actions config *in progress*
- [x] ~~Remove `-m dev` from SDK docs *help wanted*~~ this is a different *dev* module, *wontfix*
- [x] Reach consensus on where to move sub-modules

## TODO after merging

- Change docs references. See [this comment](https://github.com/dagger/dagger/pull/8042#issuecomment-2252589771)

## Where should sub-modules go?

This ~~is~~ was a contentious point, but we have consensus on *modules/*.

The modules in question:

| Before | After |
| -- | -- |
| `github.com/dagger/dagger/dev/dirdiff` | `github.com/dagger/dagger/modules/dirdiff` |
| `github.com/dagger/dagger/dev/go` | `github.com/dagger/dagger/modules/go` |
| `github.com/dagger/dagger/dev/golangci` | `github.com/dagger/dagger/modules/golangci` |
| `github.com/dagger/dagger/dev/graphql` | `github.com/dagger/dagger/modules/graphql` |
| `github.com/dagger/dagger/dev/markdown` | `github.com/dagger/dagger/modules/markdown` |
| `github.com/dagger/dagger/dev/ruff` | `github.com/dagger/dagger/modules/ruff` |
| `github.com/dagger/dagger/dev/shellcheck` | `github.com/dagger/dagger/modules/shellcheck` |
| `github.com/dagger/dagger/dev/wolfi` | `github.com/dagger/dagger/modules/wolfi` |
| `github.com/dagger/dagger/dev/ps-analyzer` | `github.com/dagger/dagger/modules/ps-analyzer` |

Options we discussed:

| Option | Example |
| -- | -- |
| top-level | `github.com/dagger/dagger/ruff` |
| **under `modules`** ⭐  | `github.com/dagger/dagger/modules/ruff` |
| under `.dagger` | `github.com/dagger/dagger/.dagger/ruff` |
| spin out stdlib | `github.com/dagger/std/ruff` |
| spin out CI module + submodules | `github.com/dagger/ci/ruff` |
| spin out individual submodules | `github.com/dagger/ruff` |
| "your (Go) house your (Go) rules" | `github.com/dagger/dagger/internal/ruff` |
| "your (Python) house your (Python) rules" | `github.com/dagger/dagger/sdk/python/ruff` |
| give up and revert everything | `github.com/dagger/dagger/dev` |

## FAQ

> Will this break external imports of `github.com/dagger/dagger/dev/foo`?

No, because `dagger.json` is pinned.

It *will* break `dagger install` commands that reference the module without a pinned tag or commit. It *may* also break a future `dagger update` command.

Given the limited exposure of the `dev` module to external imports (on purpose, as we anticipated future changes), I think there is no actual breakage to worry about.

> Is `.dagger` really the right source directory?

It's better than `ci`... Let's try, and see how it feels. See also [this discussion](https://github.com/dagger/dagger/pull/8042#issuecomment-2251564093).

> Isn't weird to have dagger sub-modules at the root of the repo? Like `github.com/dagger/dagger/ruff`...

This is a contentious point. See "where should sub-modules go"?


shykes:
Here is why I had to disable `sdk().python()` from the main module. Can anyone reproduce this from a fresh checkout?

```console
$ dagger install --name python-sdk-dev ./sdk/python/dev
Full trace at https://dagger.cloud/dagger/traces/1da44b623400f23f1489e9d42c7afe6e

✔ connect 0.8s
✔ moduleSource(refString: "./sdk/python/dev"): ModuleSource! 0.0s
✔ moduleSource(refString: "sdk/python/dev"): ModuleSource! 0.0s
✔ moduleDependency(
    name: "python-sdk-dev"
    source: ✔ moduleSource(refString: "sdk/python/dev"): ModuleSource! 0.0s
  ): ModuleDependency! 0.0s
✔ ModuleSource.withDependencies(dependencies: [xxh3:2950ff247d28f4b3]): ModuleSource! 0.0s
✘ ModuleSource.asModule: Module! 0.0s
! failed to create module: select: failed to update module dependencies: module "dagger-dev" dependency "python-sdk-dev" with source root path "sdk/python/dev" does not exist or does not have a configuration file
Error: failed to generate code: input: moduleSource.withContextDirectory.withDependencies.asModule resolve: failed to create module: select: failed to update module dependencies: module "dagger-dev" dependency "python-sdk-dev" with source root path "sdk/python/dev" does not exist or does not have a configuration file
```

Note that `./sdk/python/dev` exists, and is a valid module:

```console
$ dagger -m ./sdk/python/dev functions
Full trace at https://dagger.cloud/dagger/traces/8624ba7bcbb62c339935cfbdf027430f

✔ connect 0.9s
✔ initialize 2.8s

Name                 Description
build                Build Python SDK package for distribution.
container            Container to run commands in
docs                 Preview the reference documentation.
format               Format source files.
generate             Generate the client bindings for the API.
lint                 Check for linting errors.
supported-versions   Supported Python versions.
test                 Run the test suite.
test-versions        Run the test suite for all supported versions.
typecheck            Run the type checker (mypy).
with-container       Replace container.
with-directory       Mount a directory on the base container.
```

grouville:
Same behavior, checking 👀

jedevc:
I think the other 3 points make sense, those SGTM 🎉 Just a little caught up on moving all the helper modules to the top level.

I would rather we keep those in `.dagger`:
- All the dagger code can live in one place, no need to jump around
- It provides a logical distinction between the code I'm building and the code that builds
- It's easier for new contributors to find all the dagger code - just look in the named `.dagger` folder, and it's all in there
- It keeps the top-level from becoming cluttered - I often use the directory view to navigate the repo (I suspect I'm not entirely alone) and so it's nice to keep each level as small as possible.

I suspect there's a reason for moving them up - I'm guessing it's nicer for non dagger callers to access it using `github.com/dagger/dagger/shellcheck` instead of `github.com/dagger/dagger/ci/shellcheck` or similar. But if that's the goal, I'd argue that those modules aren't actually ready to be consumed by external callers - and if they are, we should be using them as external callers ourselves, and move them into `github.com/dagger/std` or similar.

Is there another reason we should move them up? I think that's probably the most controversial thing I can see in this PR, just off a quick glance, but if the aim is to merge quickly and iterate, potentially we could split that out and discuss that one separately - the other parts SGTM 🎉 

(I'm not 100% sure about `.dagger`, but I'm willing to try it for a bit, it does really clearly label where to look for dagger code, and I'm a big fan of that kind of explicit signposting)

shykes:
> Is there another reason we should move them up?

Nope, having them importable at a sane address is indeed the main reason.

> But if that's the goal, I'd argue that those modules aren't actually ready to be consumed by external callers - and if they are, we should be using them as external callers ourselves, and move them into github.com/dagger/std or similar - that way we can dogfood the experience ourselves.

Or, let's dogfood the experience of having importable submodules in the same repo, like I'm proposing here. Our users will encounter this situation so we should make sure it works well. Spinning out into a separate repo can happen later. To do it now would be rushing it.

> if the aim is to merge quickly and iterate, potentially we could split that out and discuss that one separately

Not ideal in this case, I really don't want to leave a trail of modules imported at `github.com/dagger/dagger/.dagger/foo`

> I'm not 100% sure about .dagger, but I'm willing to try it for a bit, it does really clearly label where to look for dagger code, and I'm a big fan of that kind of explicit signposting

I also tried an alternative: `internal/dagger/main` but it got confusing. Still viable IMO but requires merging go modules (to avoid 2 levels of internal modules aka `github.com/dagger/dagger/internal/dagger/main/internal/dagger`..) and that would be extra work. May try again later.

grouville:
> Isolating the root cause

Found the culprit: it is this line -> https://github.com/dagger/dagger/blob/main/sdk/python/dev/dagger.json#L9

As the module's `dagger.json` has an exclude with `**/sdk` AND the module lies in this path, it seems to exclude itself. Sounds like a bug, and I am not even sure on why this exclude is necessary. Any idea cc @helderco ?

shykes:
> As the module's `dagger.json` has an exclude with `**/sdk` AND the module lies in this path, it seems to exclude itself. Sounds like a bug, and I am not even sure on why this exclude is necessary.

Thank you @grouville ! This was originally `**/dev/sdk`, it probably got search-replaced by mistake. I don't understand the original purpose of the exclude. Probably a question for @helderco. Maybe it's a generalization of `sdk/python/dev/sdk`?

grouville:
> > As the module's `dagger.json` has an exclude with `**/sdk` AND the module lies in this path, it seems to exclude itself. Sounds like a bug, and I am not even sure on why this exclude is necessary.
> 
> Thank you @grouville ! ~This was originally `**/dev/sdk`, it probably got search-replaced by mistake. I don't understand the original purpose of the exclude. Probably a question for @helderco. Maybe it's a generalization of `sdk/python/dev/sdk`?~ ignore what I wrote here, I thought the issue came from the root `dagger.json` I had edited. I just realized your link was to another dagger.json.

That's the fix: https://github.com/dagger/dagger/pull/8045. It does not fix the issue per se, probably a global replace mistake I agree, but at least it adds some nice debugging to help the user

grouville:
> I just realized your link was to another dagger.json. 

I think the issue is a bit more subtle:
When loading a module and its submodules, basically we do not scope the globals to the submodules, parent module's context seems to inherit it (which is understandable with the point of view: an imported module works best when ignoring those files, so parent module should follow the same rules. However, it could lead to below issue):

*Example with Solomon's PR, with `dagger install sdk/python/dev`*
```shell
Includes: [golangci/dagger.json wolfi/dagger.json **/go.work **/*.go graphql/**/* sdk/python/dev/**/* **/go.mod dirdiff/dagger.json dirdiff/**/* wolfi/**/* go/**/* **/go.sum golangci/**/* shellcheck/dagger.json dagger.json **/go.work.sum **/vendor/ .dagger/**/* go/dagger.json graphql/dagger.json shellcheck/**/* sdk/python/dev/dagger.json]"

Excludes: [**/.git **/.direnv **/.venv **/__pycache__ **/.*cache **/sdk]
```
Here, the `**/sdk` exclusion interferes with the inclusion `sdk/python/dev/**/*`, with `**/sdk` coming from the `dagger.json` in `sdk/python/dev/**/*` itself.

3 strategies: find out the conflict and error OR remove one of the two OR scope exclusions / inclusions globals to the context of their dagger.json ?

I do understand the logic behind inheriting the patterns from the parent module, but this can easily create some funky states

My preference is on erroring, which linked PR does

shykes:
> > I just realized your link was to another dagger.json.
> 
> I think the issue is a bit more subtle: When loading a module and its submodules, basically we do not scope the globals to the submodules, parent module's context seems to inherit it (which is understandable with the point of view: an imported module works best when ignoring those files, so parent module should follow the same rules. However, it could lead to below issue):
> 
> _Example with your PR, with `dagger install sdk/python/dev`_
> 
> ```shell
> Includes: [golangci/dagger.json wolfi/dagger.json **/go.work **/*.go graphql/**/* sdk/python/dev/**/* **/go.mod dirdiff/dagger.json dirdiff/**/* wolfi/**/* go/**/* **/go.sum golangci/**/* shellcheck/dagger.json dagger.json **/go.work.sum **/vendor/ .dagger/**/* go/dagger.json graphql/dagger.json shellcheck/**/* sdk/python/dev/dagger.json]"
> 
> Excludes: [**/.git **/.direnv **/.venv **/__pycache__ **/.*cache **/sdk]
> ```
> 
> Here, the `**/sdk` exclusion interferes with the inclusion `sdk/python/dev/**/*`, with `**/sdk` coming from the `dagger.json` in `sdk/python/dev/**/*` itself.
> 
> 3 strategies: find out the conflict and error OR remove one of the two silently (having an opinion in case of conflicts) OR scope exclusions / inclusions globals to the context of their dagger.json ?
> 
> I do understand the logic behind inheriting the patterns from the parent module, but this can easily create some funky states
> 
> My preference is on erroring, which linked PR does

I don't think sub-modules should inherit excludes. A module's exclude/include should only apply to itself IMO.

shykes:
- Fixed the broken sdk/python/dev import (simply removed the conflicting exclude for now)
- Updated the README
- Added a FAQ in the PR description

I would love to get this merged ASAP 🙏

jedevc:
Hm, CI looks to be legitimately failing :weary: 

We should also make sure to update (just found by grepping for `dev` through the codebase):
- `docs` references to `github.com/dagger/dagger/dev`
- `.github` references to the `dev` module (it's a bug that CI manages to run at all - since `-m dev` should fail out)
- References to `-m dev` in various internal SDK docs

> Or, let's dogfood the experience of having importable submodules in the same repo, like I'm proposing here.

I mentioned above that I think this hurts the navigability of the code of the repo - I would like to work out how to address that before merging this. We want this repo to be an example of how to use dagger best - but I personally don't think mixing dagger-code and code-code is a best practice, I would rather people didn't try and copy that.

This is really confusing to me:
```
❯ ls
analytics  auth  bin  CHANGELOG.md  cmd  CODE_OF_CONDUCT.md  CONTRIBUTING.md  core  dagger.json  dagql  dirdiff  docs  engine  go  golangci  go.mod  go.sum  graphql  hack  helm  install.ps1  install.sh  internal  LICENSE  markdown  network  NOTICE  README.md  RELEASING.md  ruff  sdk  shellcheck  testctx  wolfi
```

Which of the above directories are dagger modules vs go packages? Is `sdk` an installable module? What about `go`? The only reason I know is because I work on dagger - a new contributor looking through will likely get confused looking for engine code in a module, or a module user looking for modules will get confused by engine code.

---

I might just be getting hung up on this - curious to hear what other core folks think on this, cc @sipsma @vito @helderco. It's a change to the structure of the repo, and I don't want to speak for anyone else :smile: 

A few things we could maybe consider doing if we want neater importable paths but without mix-matching the structure:
- We could move all of the go engine code into a neat `engine` package. This means that at least it's all in one place, but now all of the import paths for the go packages is different - this is a bit of a pain.
- We could look at setting up vanity URLs - this is now supported because of @grouville's work for module refs. We could potentially have the implementation of our sub-modules in `.dagger/shellcheck`, but referenced by doing `dagger.io/shellcheck`. I kinda like this idea, it means that we could move our directories to different places, use different git repos, whatever, and users don't need to update their calling code.

shykes:
> docs references to github.com/dagger/dagger/dev

Those are currently commit-pinned. So if we want the replacement to be commit-pinned also (seems reasonable) then it will need to be in a follow-up PR. Note that the command will not break before that, thanks to the commit-pin.

shykes:
> I mentioned above that I think this hurts the navigability of the code of the repo - I would like to work out how to address that before merging this

I get it, and open to discussing alternatives. Spinning out to a different repo IMO is not a good one (we should make that decision on its own merits).

I see you are proposing other alternatives, thank you, now turning my attention to those.

helderco:
### Confusion around `python/dev` issue

The `**/sdk` exclude was meant to exclude what's on `.gitignore`, basically the generated files from the SDK (i.e., `sdk/python/dev/sdk`), which shouldn't be uploaded back. The starting `**/` isn't really needed now, but it was added because the `sdk/python/dev` module **used to have sub-modules**. 

Additionally, I assumed those exclude patterns were being prefixed with the subdir to the module, relative to the context directory, like include patterns do. Basically I expected it was being expanded to this:
`sdk/python/dev/**/sdk`, which would exclude any `sdk` generated dir from those submodules. Haven't confirmed this is not the reality personally, just going off from @grouville's findings.

That pattern can be adjusted, I just want to exclude the `sdk` inside `sdk/python/dev`.

Moving `/dev` to the root changes things because it’s now a parent module to all the monorepo’s modules and that has an impact. @sipsma wrote about this in:
- https://github.com/dagger/dagger/issues/6902


### Modules in root

I’m with @jedevc on grouping those modules. Doesn’t have to be in `.dagger`, which can be reserved for the “dev” module, but at least a `github.com/dagger/dagger/modules/shellcheck`  seems reasonable to me, at least for now. Maybe restructure later with the vanity URLs and bringing them back to `.dagger` (or keep them in `modules`, just providing a cooler import URL).

shykes:
> Which of the above directories are dagger modules vs go packages? Is `sdk` an installable module? What about `go`? The only reason I know is because I work on dagger - a new contributor looking through will likely get confused looking for engine code in a module, or a module user looking for modules will get confused by engine code.

There is merit to that concern for sure. But just to keep things in perspective: the structure of the repo was already not crystal clear unless you had inside maintainer knowledge.

> I might just be getting hung up on this - curious to hear what other core folks think on this, cc @sipsma @vito @helderco. It's a change to the structure of the repo, and I don't want to speak for anyone else 😄
> 
> A few things we could maybe consider doing if we want neater importable paths but without mix-matching the structure:
> 
> * We could move all of the go engine code into a neat `engine` package. This means that at least it's all in one place, but now all of the import paths for the go packages is different - this is a bit of a pain.

If this fixes a repo readability issue that was already there, but is now in the spotlight because of my PR, then yes we should do it.

If it doesn't improve the readability of the repo regardless of this PR, then we should not do it IMO.

> * We could look at setting up vanity URLs - this is now supported because of @grouville's work for module refs. We could potentially have the implementation of our sub-modules in `.dagger/shellcheck`, but referenced by doing `dagger.io/shellcheck`. I kinda like this idea, it means that we could move our directories to different places, use different git repos, whatever, and users don't need to update their calling code.

I think we should avoid adding vanity URLs into the mix at this point. It wouldn't address the core issues of repository layout and how to cleanly embed one project within another. It would only paper over those issues while adding an extra layer of complexity.

One example of core issue to address:

- Yes on the one hand, making dagger sub-modules also sub-directories of the "host" project may be confusing as you noted
- But on the other hand, nesting a submodule under `.dagger` is also problematic because it obsfucates the relationship between module hierarchy and directory hierarchy. It also requires understanding the concept of "module source dir" and how it related to "module root dir" in order to understand how everything is organized. You were talking about requiring arcane knowledge...

shykes:
My suggestion:

- If moving engine code under `engine/` makes the repo more readable, then we should do that, but in a followup PR.

- Moving submodules to a subdirectory makes sense. But the key IMO is to think of those subdirectories as actual components of the dagger project, and name them accordingly. It's a Go project so we should follow Go conventions. I see two good candidates: `pkg/foo` or `internal/foo`

- Option 1: `pkg/foo`. A common convention for utility sub-packages of a Go module. For example `pkg/ruff` would be a package related to integrating with Ruff. Sure for technical reasons (ie. limitations in our DX) it's difficult to have a Go Dagger module also be a proper reusable Go package. But that can be fixed later. The point is that code exists, it's in that directory, it's legitimately a Go package, so it should live in the right location within the host project 

- Option 2: `internal/foo`. Same as `pkg`, but explicitly not to be imported. "This is go code for the private use of this project".

Anticipating the obvious objection to both options:

>  but the host project doesn't actually use this ruff integration code! It may arguably be a valid sub-package, but it's not relevant to the host project

That is a valid concern. If 1) a submodule of the "guest project" (our module) is not relevant enough to the host principle to justify squatting its namespace, but 2) it's a valid submodule of the guest, and should be importable at a good and stable address relative to the guest... Then that contradiction needs to be resolved somehow.

One way is to *spin out the guest*, eg. spin out our CI to `github.com/dagger/ci` or something equivalent.

Another way is to spin out the sub-modules as you suggested.

I have concerns with both, mostly around the timing (don't want to rush important supply chain decisions based on tactical blockers in a PR).

If the contradiction exists but can't be resolved by spinning out the offending modules for timing reasons, then a clear convention for *hiding them* would help in the meantime. Perhaps an argument for `internal/foo`?

helderco:
@shykes, you haven't addressed my suggestion of grouping the modules under something like `modules`. It's not `.dagger`, `dev`, or `ci`. I think `modules` is reasonable and trivial to apply in this PR. Better than spread on the root of the repo. 

Also, the organization of the modules as just normal Go packages don't account for modules in other languages. Ruff could be written in a Python module. A good reason for that is because Ruff users will most likely be Python users and could benefit from learning from its source code in how to make Dagger Modules.

shykes:
@helderco `modules` doesn't solve @jedevc 's concern about confusing what is part of the host project, and what is not. In the context of a git checkout by a new contributor, "modules" at the root looks like core components of the host.

You make a good point re: python modules in a go project. I think taking the "your house your rules" approach, you could argue that `ruff` and `uv` are not part of the dagger engine and CLI, but of the Python SDK project. Therefore they should live under `sdk/python/` following python sub-package conventions.

In that context it's weird that `ruff` is written in Go, but that actually surfaces a contradiction that already exists and would force us to address it.

Another outcome of "your house your rules": we conclude that Dagger is actually a multi-language project, not Go-only. And we reorganize our repo accordingly.

jedevc:
> @helderco `modules` doesn't solve @jedevc 's concern about confusing what is part of the host project, and what is not. In the context of a git checkout by a new contributor, "modules" at the root looks like core components of the host.

It's pretty close to solving the problem tbh - having modules live in `modules` is clear *enough*, and would be kinda fine at least until we split them out. Not perfect, but also not the worst.

---

Complete tangent - because I'm a little worried about moving these modules around, when we do it next, we should add support for a `dagger.json` that has a `redirect` field - that way you can keep referring to a `dagger.json` in an old place, even once it's been removed. That way this makes these transitions in the future a bit easier.

helderco:
> @helderco `modules` doesn't solve @jedevc 's concern about confusing what is part of the host project, and what is not. In the context of a git checkout by a new contributor, "modules" at the root looks like core components of the host.

We already had those modules under `dev`, I think the main issue is them being spread on the root, mixed with other stuff. I'm not sure he's suggesting the PR is blocked on going the extra mile in making that distinction clear now.

shykes:
OK, I addressed `modules` so please address my other suggestions 🙏 . Then let's regroup.

For me this is an important discussion of best practices, with consequences beyond just our repo. So worth discussing the different options in depth.

shykes:
I added a list of remaining tasks in the description, thanks @jedevc for contributing most of them. I just pushed a tentative fix to CI config.

shykes:
Added a top-level section "where should the sub-modules go", dedicated to our little bikeshedding session 😛

shykes:
Relevant: https://github.com/dagger/dagger/issues/8048

matipan:
I like the idea of having a `modules` folder where all modules live, even if not all are relevant to the host project. Talking as a newbie to this codebase, I don't think this proposed structure would make understanding how the project is built any harder.

My major concern with using a `dot` folder is that they are often referred to as hidden folders/files. Going into a terminal and doing an `ls` on the project would, by default, not show all the dagger specific codebase. Many fuzzy finders also ignore dot folders and files by default (like fzf, ctrl-p and others). Our point with Dagger is that your CI is an important part of your project that you should treat as regular software. Putting the contents of the top level module that builds, packages and releases our software project into a hidden folder makes it feel less relevant. I understand that this approach is already present, such as `.github`, but I honestly never liked it and I believe dagger code has more importance than the config files often written into `.` folders.

wingyplus:
Needs to add dev/ps-analyzer module to the module in questions list?

shykes:
I'm back online, picking this back up. Any merge conflicts to look out for @jedevc @helderco ? Is this a lost cause?

shykes:
@jedevc @helderco @matipan it looks like we have consensus on moving sub-modules to `modules/`, so I pushed that.

shykes:
> Needs to add dev/ps-analyzer module to the module in questions list?

Done! Thank you

jedevc (approved):

jedevc:
Think CI should be back and healthy again - approving now, if everything looks good, LGTM :tada:

sdk/python/README.md
helderco, line 1: This doesn't apply to this module. The `-m dev` in this file is for running `sdk/python/dev` when inside `sdk/python`.

sdk/python/dev/dagger.json
```diff
@@ -1,12 +1,7 @@
 {
   "name": "python-sdk-dev",
   "sdk": "python",
-  "exclude": [
-    "**/.venv",
-    "**/__pycache__",
-    "**/.*cache",
-    "sdk"
-  ],
+  "exclude": ["**/.venv", "**/__pycache__", "**/.*cache"],
```
helderco, line 4: It's a shame that the sdk exclude was removed here. There's a bug where include/exclude patterns from dependencies aren't prefixed with the relative paths to them.

sdk/typescript/dev/src/index.ts
```diff
@@ -89,7 +89,7 @@ class TypescriptSdkDev {
   /**
    * Bump the TypeScript SDK's Engine dependency.
    *
-   * Example usage: `dagger -m dev call --source=. bump --version v0.10.2 -o .`
+   * Example usage: `dagger call --source=. bump --version v0.10.2 -o .`
```
helderco, line 92: Same as Python, this is not about the root `/dev` module. It's about `sdk/typescript/dev`.

PR #8043, opened 2024-07-25
[WIP] daggerize sdk changelog generation

sipsma:
Spinning this out from https://github.com/dagger/dagger/pull/8040

In the effort of slowly automating our release process towards being one-click (or at least "few-click"), this automates the SDK changelog generation, which is currently a manual step in RELEASING.md (among many others we should gradually squash over time).

I hit my timebox on this today, so a few TODOs left.
- [ ] Figure out how to **_safely_** export deleted files (e.g. `sdk/<name>/.changes/unreleased/*`) back to the caller
   - Emphasis on "safely" because while trying this I used the `--wipe` option on export and accidentally deleted my `.git` dir, which was mildly devastating 🥲
 - [ ] Add to CI. Only tricky part is these changelogs need the PR number of the bump engine PR, so we need to first create the PR, generate these changelogs, and then update the PR with the extra diff
 - [ ] Update RELEASING.md (delete a big chunk of instructions)

cc @shykes @jedevc

### DIFF
--- a/dev/sdk.go
+++ b/dev/sdk.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"context"
+	"fmt"
+	"path"
 
 	"github.com/dagger/dagger/dev/internal/build"
 	"github.com/dagger/dagger/dev/internal/consts"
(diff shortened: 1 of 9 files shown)


jedevc:
@sipsma is this still relevant after #7705? I think a changie module would definitely still be valuable, but the sdk changelog generation process has been hugely simplified.

PR #8045, opened 2024-07-26
core(fix): error on modules include / exclude patterns conflict

grouville:
Current implementation of module inclusion / exclusion can easily lead to silent conflicts between glob patterns, during a `dagger install`: https://github.com/dagger/dagger/pull/8042#issuecomment-2252289788.

> For example, one submodule can exclude: `**/sdk`, whilst we try to install a module located at `depth1/sdk/*`

For BuildKit, this would lead to a priority for the exclusion, leading to a weird failure on `dagger install`

This commit ensures that Dagger fails with a readable explanation in case such occurence of conflict.

TODO:
- [ ] add an integration test

### DIFF
--- a/core/schema/modulesource.go
+++ b/core/schema/modulesource.go
@@ -18,6 +18,7 @@ import (
 	"github.com/dagger/dagger/dagql"
 	"github.com/dagger/dagger/engine/buildkit"
 	"github.com/dagger/dagger/engine/vcs"
+	"github.com/moby/patternmatcher"
 	"github.com/tonistiigi/fsutil/types"
 )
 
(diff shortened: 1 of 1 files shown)


helderco:
I'm not sure we should prevent an exclude from overriding an include or vice versa. It can be used by the user to override a default pattern.

grouville:
This is a potential stopgap for https://github.com/dagger/dagger/issues/6902

> I'm not sure we should prevent an exclude from overriding an include or vice versa. It can be used by the user to override a default pattern.

I am not sure either, thanks for that example🙏

Maybe shall we just provide intelligible warning to the user, showing that some of them might conflict (with the extension of the provenance: which module have which inclusion / exclusion ?) cc @sipsma @jedevc WDYT ?

helderco:
This is not just about:
- https://github.com/dagger/dagger/issues/6902

The problem you saw in:
- https://github.com/dagger/dagger/pull/8042#issuecomment-2252289788

Actually comes from a bug:
- https://github.com/dagger/dagger/issues/8136

So that `**/sdk` pattern in `sdk/python/dev/dagger.json` should be excluding `sdk/python/dev/**/sdk` and not `**/sdk`. Without that bug, there would be no conflict here.

Issue #8035, opened 2024-07-25
🐞 Docker build secrets seems broken in Python + Example in documentation is wrong for build secrets in Python


lepouletsuisse:
### What is the issue?

Injecting a secret in the `Directory.docker_build()` function seems either broken or extremely confusing.
First, the example provided in the Cookbook (See https://docs.dagger.io/cookbook#use-secret-in-dockerfile-build in Python tab) is wrong, no `dagger.DockerBuildArgs` exists, only `dagger.BuildArgs` exists, which I suppose is the one required here as it correctly injects the build args in the Dockerfile. 
Also, the example seems wrong, the `id=gh-secret` in the Dockerfile cannot exist as the id should be the name of the Dagger secrets, which should be the SHA256 of the secret (If I refer to [this comment](https://github.com/dagger/dagger/issues/5242#issuecomment-2051557730)).

Second, even while following the instruction from [this contributor](https://github.com/dagger/dagger/issues/5242#issuecomment-2051557730), I can't make these build secrets work, no matter what. 
Maybe I didn't understand at all how injecting the Docker secrets work but I can't figure out how to do it in Python. I would like to inject it as a Secret to avoid exposing it in Dagger Cloud as we plan to use it when our PoC of Dagger is done.
I tried multiple ways but nothing seems to work, just like the `secrets=[...]` part of the `Directory.docker_build()` function doesn't get correctly translated to `docker build --secret id=<secret_name>,src=<location of the secret for Dagger> ...` and doesn't include the secret at all during the build process.

### Dagger version

dagger v0.12.0 (registry.dagger.io/engine) linux/amd64

### Steps to reproduce

**Dagger main.py code**
```
# I have a Secret injected from the cmd set in self.gh_secret. I checked that the value is correctly set in the secret using 
# the plaintext function
secret_name = await self.gh_secret.name()
test_container = self.source.docker_build(
    dockerfile="Dockerfile",
    build_args=[BuildArg(name="SECRET_ID", value=secret_name)],
    secrets=[self.gh_secret]
)
```

**Dockerfile**
```
...
ARG SECRET_ID
RUN --mount=type=secret,id=$SECRET_ID,target=/run/secrets/secret.txt cat /run/secrets/secret.txt
...
```

### Log output

```
[...]

PR #8049, opened 2024-07-26
core: fix dockerfile secrets in modules by making secret pure

sipsma:
The use case of:
1. Building Dockerfiles via the `Container.build` or `Directory.dockerBuild` APIs
2. That have secrets attached
3. That were created by modules and returned as `Container` values

Appears to have been broken for a while.

Before the secret isolation in v0.12.1, it was broken because these returned Dockerfiles were re-evaulated when the core ID was loaded after the function returned but the accessor was calculated again based on the *caller*, which resulted in a different secret value and thus missing secrets when the Dockerfile evaluation got unlazied.

After v0.12.1, the error changed to be a bit earlier, but still happened as the core ID for the returned Dockerfile build was being loaded. It now was happening because the `Container.build` implementation synchronously looked up secrets before they had been identified and passed from the function call client to the caller's secret store.

The fix here is deceptively simple: just don't mark `secret` as impure anymore. It's no longer necessary to do this because of the changes made in v0.12.1. Whereas previously `secret` actually looked up and calculated the accessor in some codepaths, it now just returns a static value based purely on the inputs.

The reason this fixes the issue is a bit subtle and honestly might be a bit shaky; worth more thought. It's because when secret is pure, the whole `dockerBuild` API doesn't get cache invalidated and thus allows loading the return value to rely on dagql caching and skip the part where we were synchronously trying to find secrets that weren't loaded yet.

Instead what happens now is the value gets loaded but returned from the cache, which allows the passing of secrets from the function call client to the caller to happen. Then when the Dockerfile-built Container is evaluated, the secrets will indeed be there.

Some more details in this comment: https://github.com/dagger/dagger/issues/8035#issuecomment-2252014406

---

Fixes https://github.com/dagger/dagger/issues/8035

This does appear to fix the issue based on the integ test added and the other secret tests I ran so far seem unimpacted.

However this is such a subtle one-line change that I want to digest it a little more and at minimum probably add even more integ tests to ensure every possible corner case is exercised. Marking as a draft pending those extra tests.

EDIT: after thinking some more, it still makes sense to me. Expanded the test coverage a bit more too to test evaluation from both within the module and on the cli client side

### DIFF
--- a/.changes/unreleased/Fixed-20240729-234852.yaml
+++ b/.changes/unreleased/Fixed-20240729-234852.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Fix errors when using Dockerfile builds from module functions that have secrets.
+time: 2024-07-29T23:48:52.402969146-07:00
+custom:
+  Author: sipsma
+  PR: "8049"
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -5451,6 +5451,57 @@ func (t *Toplevel) TryArg(ctx context.Context) error {
 		})
 	})
 
+	t.Run("dockerfiles in modules", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+		ctr := c.Container().From(golangImage).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work").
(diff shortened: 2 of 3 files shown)


helderco:
This fixes the use case in the docs (pending docs fix that I'll submit). 👍

jedevc (approved):

helderco (approved):

PR #8051, opened 2024-07-27
fix PythonProjectLayout test flake

helderco:
Fixes https://github.com/dagger/dagger/issues/7662

### DIFF
--- a/core/integration/module_python_test.go
+++ b/core/integration/module_python_test.go
@@ -276,6 +276,9 @@ build-backend = "poetry.core.masonry.api"
 			c := connect(ctx, t)
 
 			out, err := daggerCliBase(t, c).
+				// uv caches a project's metadata by its absolute path so we can't change
+				// build backends and package location on the same `pyproject.toml` path
+				// concurrently because uv may return the cached metadata from another subtest.
 				WithWorkdir(fmt.Sprintf("/work/%s-%d", tc.name, i)).
 				With(fileContents("pyproject.toml", tc.conf)).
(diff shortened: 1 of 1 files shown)


core/integration/module_python_test.go
```diff
@@ -269,13 +269,14 @@ build-backend = "poetry.core.masonry.api"
...
 
-	for _, tc := range testCases {
+	for i, tc := range testCases {
 		tc := tc
 
 		t.Run(fmt.Sprintf("%s/%s", tc.name, tc.path), func(ctx context.Context, t *testctx.T) {
 			c := connect(ctx, t)
 
 			out, err := daggerCliBase(t, c).
+				WithWorkdir(fmt.Sprintf("/work/%s-%d", tc.name, i)).
```
jedevc, line 279: Hmm, these definitely shouldn't be conflicting right? It's the weird cache behavior thing (again?).

Can we add a quick note about this inline, just so it doesn't accidentally get removed? (and we can remove it if we think we have a longer term fix).

jedevc (approved):

core/integration/module_python_test.go
helderco: Yeah, sure. This fix is for uv's cache however, not ours.

sipsma:
Was merged before I was online, but want to say thanks for tracking this down @helderco! Realize we can't be 100% sure it's gone yet but the comment you left about the cause makes sense and was obviously a nightmare to figure out, so all greatly appreciated.

PR #8058, opened 2024-07-29
docs: fix docker build with secrets cookbook

helderco:
This still depends on the following fix to work:
- https://github.com/dagger/dagger/pull/8049

But can be merged before that as the snippets weren't quite right.

It's part of a fix for:
- https://github.com/dagger/dagger/issues/8035

Using a build arg is one approach, but you need to adjust the `Dockerfile` to use that as the ID. I think the approach in this PR is simpler and doesn't require changing the `Dockerfile`.

## Suggestion for the future

Better fix would be a breaking change so that the `secrets` argument to the docker build would be similar to `dagger.BuildArg`:
```graphql
input BuildSecret {
    value: Secret!
    id: String! = ""  // default to secret's name or not worth it?
}
```
Then usage would be simpler:
```go
return source.
	DockerBuild(dagger.DirectoryDockerBuildOpts{
		Secrets: []dagger.BuildSecret{
			{ID: "gh-secret", Value: secret},
		},
	}), nil
```


marcosnils (approved):

marcosnils:
//reminder: have a way to track which `dagger.json` are set to `dev-*` and eventually upgrade them when the release gets rolled out. cc @vikram-dagger

vikram-dagger (approved):

PR #8061, opened 2024-07-30
ci: update release secrets to release vars

jedevc:
These values are commonly known, so don't need to be secrets.

This will prevent them from being censored in the logs.

### DIFF
--- a/.github/workflows/engine-and-cli-publish.yml
+++ b/.github/workflows/engine-and-cli-publish.yml
@@ -41,7 +41,7 @@ jobs:
             --registry-username="$DAGGER_ENGINE_IMAGE_USERNAME" \
             --registry-password=env:DAGGER_ENGINE_IMAGE_PASSWORD
         env:
-          DAGGER_ENGINE_IMAGE: ${{ secrets.RELEASE_DAGGER_ENGINE_IMAGE }}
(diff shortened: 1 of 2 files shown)


.github/workflows/engine-and-cli-publish.yml
```diff
@@ -114,15 +114,15 @@ jobs:
             --aws-secret-access-key=env:AWS_SECRET_ACCESS_KEY \
             --aws-region=env:AWS_REGION \
             --aws-bucket=env:AWS_BUCKET \
-            --artefacts-fqdn=env:ARTEFACTS_FQDN
+            --artefacts-fqdn="$ARTEFACTS_FQDN"
```
gerhard, line 117: Don't we encourage the `env:ARTEFACTS_FQDN` format?

gerhard (approved):
LGTM apart from one small nit.

We will know for sure if this works once we merge. Everything looks good, let's see it run 🏃

.github/workflows/engine-and-cli-publish.yml
jedevc: I've updated the `artefactsFQDN` in the CLI to be a string arg, not a secret anymore. This means that the `env:` form isn't supported there.

PR #8063, opened 2024-07-31
fix: extend multi git server support to Azure devOps

grouville:
When working on private SSH support, I realized that our support for Azure refs is broken on monorepos.

**Why ?**
 Azure's handling of `go-get=1` is not foolproof:
1. `repoRootForImportPath` does work well on refs leading to a repo, infering properly the vcs provider used
2. But,  `go-get=1` returns a 500 when a subpath is included
3. Azure ref's style is explicit with the vcs being used: git is mentioned for a git based repo

If we are ok with this explicitness, we should not try to infer the vcs based on the go-get for Azure and rely on regexes

### DIFF
--- a/.changes/unreleased/Fixed-20240731-061239.yaml
+++ b/.changes/unreleased/Fixed-20240731-061239.yaml
@@ -2,7 +2,7 @@ kind: Fixed
 body: |
     feat: extend multi git server support to Azure devOps
 
-    Our support for Azure refs is broken on monorepos, as Azure returns a 500 when we query a ref containing a subpath with `go-get=1`. Instead, we rely on regexes to infer the root of the repo and the subpath from the ref
+    Our support for Azure refs was broken on monorepos. We special-case Azure DevOps since it doesn't work with the go standard convention of discovering the root of a git repository
 time: 2024-07-31T06:12:39.205696046Z
 custom:
     Author: grouville
--- a/.changes/unreleased/Fixed-20240731-061239.yaml
+++ b/.changes/unreleased/Fixed-20240731-061239.yaml
@@ -1,0 +1,9 @@
+kind: Fixed
+body: |
+    feat: extend multi git server support to Azure devOps
+
+    Our support for Azure refs is broken on monorepos, as Azure returns a 500 when we query a ref containing a subpath with `go-get=1`. Instead, we rely on regexes to infer the root of the repo and the subpath from the ref
+time: 2024-07-31T06:12:39.205696046Z
+custom:
+    Author: grouville
+    PR: "8063"
--- a/core/integration/module_config_test.go
+++ b/core/integration/module_config_test.go
@@ -1390,6 +1390,14 @@ var vcsTestCases = []vcsTestCase{
 		expectedBaseHTMLURL:      "bitbucket.org/dagger-modules/dagger-test-modules-public",
 		expectedURLPathComponent: "src",
 	},
+	{
+		name:                     "Azure DevOps without .git",
+		gitTestRepoRef:           "dev.azure.com/daggere2e/public/_git/dagger-test-modules",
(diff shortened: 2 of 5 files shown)


core/integration/module_config_test.go
```diff
@@ -1421,7 +1429,13 @@ func (ModuleSuite) TestDaggerGitRefs(ctx context.Context, t *testctx.T) {
...
 			htmlURL, err := rootModSrc.AsGitSource().HTMLURL(ctx)
 			require.NoError(t, err)
-			require.Equal(t, fmt.Sprintf("https://%s/%s/%s", tc.expectedBaseHTMLURL, tc.expectedURLPathComponent, tc.gitTestRepoCommit), htmlURL)
+			var expectedURL string
+			if tc.expectedHost == "dev.azure.com" {
+				expectedURL = fmt.Sprintf("https://%s/%s/%s", tc.expectedBaseHTMLURL, tc.expectedURLPathComponent, tc.gitTestRepoCommit)
+			} else {
+				expectedURL = fmt.Sprintf("https://%s/%s/%s", tc.expectedBaseHTMLURL, tc.expectedURLPathComponent, tc.gitTestRepoCommit)
+			}
+			require.Equal(t, expectedURL, htmlURL)
```
jedevc, line 1438: These are the same in both branches?

Instead of hardcoding `dev.azure.com` here, we could potentially have a `expectedPathPrefix` property or something in the test case, and the have that set to `?path=` for azure (and left empty for the others)?

engine/vcs/vcs.go
```diff
@@ -652,6 +652,30 @@ var vcsPaths = []*vcsPath{
 		repo:   "https://{root}",
 	},
 
+	// Azure DevOps
+	// HTTPS ref
+	{
+		prefix: "dev.azure.com/",
+		re:     `^(?P<root>dev\.azure\.com/(?P<account>[A-Za-z0-9_.\-]+)(/(?P<project>[A-Za-z0-9_.\-]+))?/_git/(?P<repo>[A-Za-z0-9_.\-]+)(\.git)?)(/[\p{L}0-9_.\-]+)*(/.*)?$`,
```
jedevc, line 659: I shall trust you on this :laughing:

.changes/unreleased/Fixed-20240731-061239.yaml
```diff
@@ -0,0 +1,9 @@
+kind: Fixed
+body: |
+    feat: extend multi git server support to Azure devOps
+
+    Our support for Azure refs is broken on monorepos, as Azure returns a 500 when we query a ref containing a subpath with `go-get=1`. Instead, we rely on regexes to infer the root of the repo and the subpath from the ref
```
jedevc, line 5:
```suggestion
    Our support for Azure refs was broken on monorepos, as Azure returns a 500 when we query a ref containing a subpath with `go-get=1`. Instead, we rely on regexes to infer the root of the repo and the subpath from the ref
```

I don't know if we need the detail about 500s, we can just say we special-case azure dev-ops since it doesn't work with the go standard convention of discovering the root of a git repository.

jedevc (commented):
LGTM, a couple tiny nits :tada:

Good catch!

core/integration/module_config_test.go
grouville: Done 🙏

.changes/unreleased/Fixed-20240731-061239.yaml
grouville: Changed, you're right, it is better 🙏

grouville:
Failures are unrelated 🙏

jedevc (approved):

jedevc:
Awesome awesome, let's get this for the release today :tada:

Issue #7585, opened 2024-06-07
Make all SDKs usable from git refs


sipsma:
Background: the root cause of https://github.com/dagger/dagger/issues/7583 was lack of pinning deps in the typescript sdk (one of the unpinned ones updated out of band and broke every ts module in all previous engine releases) but it was made much more painful by the fact that we needed to do an engine release in order to fix it rather than just patching the TS SDK.

If we could instead allow users to easily import all of the builtin SDKs as git refs if they want, that would improve this situation greatly:
* TS - not possible atm due to it's dep on the codegen bin, which doesn't get built as part of the module itself
* Go - not possible atm because it is the special root SDK that isn't actually a module
* Python - might actually be possible? But haven't tried yet to confirm

---

For the short term, there's a couple viable paths to improve the situation:
1. For the TS dep on the codegen bin, we can either pass the codegen bin as an optional constructor arg or update the TS sdk module to build it itself (adds overhead, but should be highly cacheable at least)
   * This feels like a relatively low-risk approach
3. For the Go SDK, we could create a wrapper SDK that does the same thing but is an actual module and is importable from git
   * Less sure how I feel about this, very one-way-door-y for a short term solution

---

For the longer term, a vague outline of what I'm imagining:
1. We solve the Go problem by making it an actual module.
   * One way to model this would be to still have a special root SDK, but just don't make it Go. It can be "root-codegen" or something similar. Then go uses that, python+ts can still use go, etc.
1. We update dagger.json to automatically transform existing unpinned sdks like "go", "typescript", "python" to the equivalent git refs
   * we can still support the shorthand names in cli commands; this would just change dagger.json
1. We still package the SDKs in the engine container, but as a cache of their git refs rather than as the actual primary definition of the SDK

harleylang:
This would be an amazing feature to ship so teams can quickly respond to upstream breakages. @sipsma is this now partly shipped in https://github.com/dagger/dagger/pull/7584?

sipsma:
> This would be an amazing feature to ship so teams can quickly respond to upstream breakages. @sipsma is this now partly shipped in #7584?

That PR was just a temporary workaround; the idea here is to make the *official* SDKs all usable inherently from git refs (whereas that workaround needed to create a wrapper SDK, which technically worked but is not super maintainable in the long run).

PR #8067, opened 2024-07-31
feat: add elixir/php sdk aliases

jedevc:
This allows users to use `elixir` to use the elixir SDK, and `php` to use the php SDK, without needing the full `github.com/dagger/dagger/sdk/<sdk>/runtime@<version>` URL which is very unieldy.

~~:warning: requires https://github.com/dagger/dagger/pull/7858 so that we can access `engine.Tag` which is the version of the release, so the v0.12.4 release will correctly point this to `github.com/dagger/dagger/sdk/<sdk>/runtime@v0.12.4`.~~

### DIFF
--- a/core/schema/sdk.go
+++ b/core/schema/sdk.go
@@ -6,11 +6,13 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/opencontainers/go-digest"
 
 	"github.com/dagger/dagger/core"
 	"github.com/dagger/dagger/dagql"
+	"github.com/dagger/dagger/engine"
 	"github.com/dagger/dagger/engine/distconsts"
 )
 
@@ -92,8 +94,24 @@ func (s *moduleSchema) builtinSDK(ctx context.Context, root *core.Query, sdkName
 	case "typescript":
 		return s.loadBuiltinSDK(ctx, root, sdkName, digest.Digest(os.Getenv(distconsts.TypescriptSDKManifestDigestEnvName)))
 	default:
-		return nil, fmt.Errorf("%s: %w", sdkName, errUnknownBuiltinSDK)
+		sdkName, sdkVersion, hasVersion := strings.Cut(sdkName, "@")
+		if !hasVersion {
+			sdkVersion = engine.Tag
+		}
+		sdkSuffix := ""
+		if sdkVersion != "" {
+			sdkSuffix = "@" + sdkSuffix
+		}
+
+		switch sdkName {
+		case "elixir":
+			return s.sdkForModule(ctx, root, "github.com/dagger/dagger/sdk/elixir"+sdkSuffix, dagql.Instance[*core.ModuleSource]{})
+		case "php":
+			return s.sdkForModule(ctx, root, "github.com/dagger/dagger/sdk/php"+sdkSuffix, dagql.Instance[*core.ModuleSource]{})
+		}
 	}
+
+	return nil, fmt.Errorf("%s: %w", sdkName, errUnknownBuiltinSDK)
 }
 
 // moduleSDK is an SDK implemented as module; i.e. every module besides the special case go sdk.
--- a/.changes/unreleased/Added-20240731-134929.yaml
+++ b/.changes/unreleased/Added-20240731-134929.yaml
@@ -1,0 +1,9 @@
+kind: Added
+body: |-
+  New SDK aliases for `elixir` and `php`
+
+  SDKs with experimental module support (elixir and php) can now be accessed using `--sdk=<sdk>` (such as `--sdk=elixir` and `--sdk=php` respectively) instead of the full form `--sdk=github.com/dagger/dagger/sdk/<sdk>@<version>`.
+time: 2024-07-31T13:49:29.069938099+01:00
+custom:
+  Author: jedevc
+  PR: "8067"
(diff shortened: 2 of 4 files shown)


core/schema/sdk.go
```diff
@@ -91,11 +92,22 @@ func (s *moduleSchema) builtinSDK(ctx context.Context, root *core.Query, sdkName
 		return s.loadBuiltinSDK(ctx, root, sdkName, digest.Digest(os.Getenv(distconsts.PythonSDKManifestDigestEnvName)))
 	case "typescript":
 		return s.loadBuiltinSDK(ctx, root, sdkName, digest.Digest(os.Getenv(distconsts.TypescriptSDKManifestDigestEnvName)))
+	case "community/elixir":
+		return s.sdkForModule(ctx, root, pinUpstreamModule("github.com/dagger/dagger/sdk/elixir/runtime"), dagql.Instance[*core.ModuleSource]{})
+	case "community/php":
+		return s.sdkForModule(ctx, root, pinUpstreamModule("github.com/dagger/dagger/sdk/php/runtime"), dagql.Instance[*core.ModuleSource]{})
```
carnage, line 98: The dagger.json for the php sdk is in the sdk root, not the runtime dir so this should be: `github.com/dagger/dagger/sdk/php`

jedevc: Good point. @helderco are we okay with this being inconsistent with go/typescript/python, or should we update those/these?

jedevc:
> For the Elixir, if the PR #7667 get merged, the sdk would be `community/elixir@<version>` instead?

Left a note inline that this will need to be changed once #7667 is merged.

I've also updated the PR to allow using `@` in these - you can specify `community/elixir@v0.12.0` for example, to explicitly get an older version of the elixir SDK.

helderco:
> For the Elixir, if the PR #7667 get merged, the sdk would be `community/elixir@<version>` instead?

Yes, we need to get your PR merged as soon as possible.

core/schema/sdk.go
helderco: I think it's ok for now. It's consistent with community SDKs anyway. We'll eventually probably want to do the same for Py and TS too, even if just for:
- https://github.com/dagger/dagger/issues/7585

jedevc:
One thing that did cross my mind... we could potentially also have git-reachable versions for our core SDKs (cc @sipsma who did this before in https://github.com/dagger/dagger/pull/7584).

We could do this by doing this only when we detect a version in the name, something like `"sdk": "typescript@v0.12.0"` (note, this still isn't trivial for go, grr).

---

As an aside - by doing this we're making a commitment to not move `dagger.json`s for the SDKs! If these move, then either:
- We have to explicitly note the version where they moved in `sdk.go`
- Or we have to redirect them somehow using a dummy `dagger.json` (which I've also suggested somewhere else, probably worth doing anyways)

helderco:
> One thing that did cross my mind... we could potentially also have git-reachable versions for our core SDKs (cc @sipsma who did this before in #7584).

Yeah, I think that's the idea behind:
- https://github.com/dagger/dagger/issues/7585

> [...] we're making a commitment to not move `dagger.json`s for the SDKs! [...]

Hmmm... the reason we needed to put `dagger.json` in the root of those SDKs is because there's no way to get the sources for the SDK if it points to the `runtime` subdir. Python and TypeScript also need this for now if we want to be able to install from a git remote ref.

However, contextual directories solve this. So we could have the `dagger.json` files back in their `runtime` dir when that comes out.

I also find it nicer to point `--sdk=...` to the path where the SDK is rather than its runtime dir, but this PR's aliases hide that from users anyway.

gerhard (commented):
This is coming together nicely @jedevc. So far, so good. Will check out https://github.com/dagger/dagger/pull/7858 in the next timebox.

core/schema/sdk.go
```diff
@@ -92,8 +94,24 @@ func (s *moduleSchema) builtinSDK(ctx context.Context, root *core.Query, sdkName
...
+		if sdkVersion != "" {
+			sdkSuffix = "@" + sdkSuffix
+		}
+
+		switch sdkName {
+		case "community/elixir":
+			return s.sdkForModule(ctx, root, "github.com/dagger/dagger/sdk/elixir"+sdkSuffix, dagql.Instance[*core.ModuleSource]{})
+		case "community/php":
+			return s.sdkForModule(ctx, root, "github.com/dagger/dagger/sdk/php"+sdkSuffix, dagql.Instance[*core.ModuleSource]{})
+		}
```
helderco, line 111: Hmmm... at this point we could even do away with the `community/` prefix, it just becomes an implementation detail that it's being fetched from git rather than bundled in the engine. The only advantage of the prefix is as a stability signal for users, but not sure if that's worth it.

TomChv: I'm not sure about this one, by prefixing `community`, we are explicit about who take care of the support, it may  happen that community SDKs implements features later than dagger's so it creates a simple way to know what's "official" and what's "community".

@shykes I think your opinion would be useful on this one

shykes:
There is no concept of community SDK. There are stable and experimental SDKs.

core/schema/sdk.go
jedevc: Removed the `community` prefix.

carnage (commented):
Happy with the final version; can't approve officially though.

PR #8068, opened 2024-07-31
feat: Support custom objects in PHP SDK runtime

charjr:
Add support for custom objects in the PHP SDK runtime.

Tested this manually with the following three classes in a module called Potato:

```php
#[DaggerObject]
class Potato
{
    #[DaggerFunction('Get Chips')]
    public function getChips(): Chips
    {
        return new Chips(
            dag()->container()->from('alpine:latest'),
            dag()->defaultPlatform(),
            new Sauce('mayo'),
        );
    }
}
```

```php
#[DaggerObject]
class Chips
{
    public function __construct(
        private readonly Container $container,
        private readonly Platform $platform,
        private readonly Sauce $sauce,
    ) {
    }

    #[DaggerFunction('Serve the chips')]
     public function serve(): string
     {
         return $this->container
             ->withExec(['echo', sprintf(
                 'Chips with a side of %s, covered in %s',
                 $this->platform->getValue(),
                 $this->sauce->type,
             )])
             ->stdout();
     }
}
```

```php
final class Sauce
{
    public function __construct(
        public readonly string $type
    ) {}
}
```

I was able to call `dagger call get-chips serve` and receive the following output:
```
Chips with a side of linux/amd64, covered in mayo
```

### DIFF
--- a/sdk/php/composer.json
+++ b/sdk/php/composer.json
@@ -7,6 +7,7 @@
     "minimum-stability": "stable",
     "require": {
         "php": ">=8.2",
+        "jms/serializer": "^3.30",
         "roave/better-reflection": "^6.25",
         "symfony/process": "^6.3|^7.0",
         "symfony/console": "^6.3|^7.0",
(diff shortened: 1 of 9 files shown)


helderco (approved):

helderco:
Failed tests are from flakes.

Issue #4855, opened 2023-03-29
✨ Log warning when using deprecated APIs


helderco:
### What are you trying to do?

API fields that are marked as deprecated only get codegen'ed into the docblock. That shows in reference docs, and maybe the IDE helps make it more visible as well (depending on language), but we should log a warning when these calls are being used.

### Why is this important to you?

Logging a warning makes it more visible whenever you're actually running a pipeline, no matter if you use a smart IDE or consult the reference docs during development. Removing the warning acts as encouragement to migrate the new behavior.

TomChv:
@helderco Is it a priority? I can work on this ASAP

helderco:
Not a priority, but may be easy to do. We no longer have any deprecated apis for now though. Python has this, only missing in Go and Node.js.

TomChv:
@helderco I think instead of doing the job on the SDK side, we should log warning in our core API using progrock, this would be a generic and durable solution :)

WDYT @vito ?

TomChv:
@vito @helderco Any thought on this so we can make progress?

> I think instead of doing the job on the SDK side, we should log warning in our core API using progrock, this would be a generic and durable solution :)

helderco:
Yeah, been thinking about this because we have pending deprecations. For example:
- https://github.com/dagger/dagger/pull/7136

I added this warning:

https://github.com/dagger/dagger/blob/fc6a74297956c41801d36081d841c7410ac9651e/core/schema/container.go#L582

The intent was for the warning to show for the user when this API is used, both in the TUI and hopefully in cloud traces as well. Nothing was shown last time I ran this.

I defer to @vito to help point out what needs to be done to make this happen.

vito:
@helderco Yeah, I think it's time to start integrating with OpenTelemetry logging properly, which I take to mean integrating `slog` with `ctx` and setting OpenTelemetry's severity levels etc. instead of mimicking stdout/stderr.

I had an idea for a "global logger" that we could use sort of like Progrock's previous logging system:

https://github.com/dagger/dagger/blob/4ec5818fab06e4ac42e4b0a74dd913e0d2c02a85/telemetry/logging.go#L40-L47

[...]

Issue #8065, opened 2024-07-31
Remove deprecated APIs in v0.13.0


jedevc:
v0.13.0 allows for breaking API changes (as long as we use the versioning compat views to keep old modules working).

So we can remove the following APIs:
- `pipeline()` APIs, deprecated in https://github.com/dagger/dagger/pull/8064

helderco:
The Python deprecations haven't been removed yet, but we can push it to 0.14.

PR #8074, opened 2024-08-01
Deprecate top-level functions in Python modules

helderco:
This isn't visibile to the user until we solve:
- https://github.com/dagger/dagger/issues/4855

This is an historical feature from early days if _Modules_ development. We hid it before the Modules launch, so it was never documented. The important thing here is just the announcement and the change log, to give a heads up that it will be removed.


jedevc (approved):
SGTM - can you also add this as a list item to #8065? That way we remember to remove it.

PR #8079, opened 2024-08-01
Improve releasing during v0.12.4

jedevc:
This is a bit more of a restructure than usual, taking into account the changes introduced in https://github.com/dagger/dagger/pull/7705.

The big parts are:
- No more `bump-engine` PR
	- Steps that were performed there are moved forwards to the new "prep" step.
	- Now there's no tricky wait for CI to pass/etc, we can shave off a huge amount of time here now.
- Go SDK release is sped up a lot
	- I realized we don't need to wait for our tests to pass in this specific step - the reason for this is, now we ourselves consume the modularized go SDK, not the external one. So technically this wasn't adding any coverage.
	- The tests would already have fun on `main` before releasing - we're okay to skip it here.
- Added a tooling section to aggregate all the tooling that should be installed prior to starting the release
- Split out dedicated sub-sections for improving releasing as we go, which makes it clearer what's going on.

This was actually a really fast release! I tagged v0.12.4 at 11:44 BST, and finished releasing all the SDKs and helm chart at 12:43 BST. Really happy with the new format, it felt much smoother and less stressful.

There are some next steps I'd like to take:
- Github release notes uploads should be automatic, I'd like to not do these manually anymore. Each SDK should just be tag+push (and we could even parallelize these).
- Add commands for *every* step - it should be possible to copy-paste the entire release process for standard releases off of `main`.
- Automate PRs for post-release:
	- Our github action: https://github.com/dagger/dagger-for-github
	- Our internal infra: Playground, Daggerverse, Cloud API + Magicache (cc @matipan)

### DIFF
--- a/.github/actions/call/action.yml
+++ b/.github/actions/call/action.yml
@@ -13,7 +13,7 @@ inputs:
 
   version:
     description: "Dagger version to run against"
-    default: "v0.12.3"
+    default: "v0.12.4"
     required: false
 
   dev-engine:
--- a/.github/workflows/_sdk_check.yml
+++ b/.github/workflows/_sdk_check.yml
@@ -15,7 +15,7 @@ on:
 
       version:
         type: string
-        default: "v0.12.3"
+        default: "v0.12.4"
         required: false
 
       dev-engine:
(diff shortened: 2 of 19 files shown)


gerhard:
Starting to look at this now.

[`TestPythonProjectLayout` is known test flake](https://dagger.cloud/dagger/traces/f2c7ac55c6558be4fa848e26102f47c7?span=9eec98df2a0712f7#9eec98df2a0712f7:L33), re-running it manually.

image
image

gerhard (approved):
I can definitely see the simplification, great job @jedevc 💪

I really like that we kept the initial intent - documentation first - while adding more automation and making everything quicker & less error-prone.

🚀

PR #8085, opened 2024-08-01
feat(rust-sdk): release 0.11.10

kjuulh:
It has been a while, releasing 0.11.10, with included breaking changes to the rust sdk api


TomChv (approved):

jedevc (changes requested):
See discord thread - ideally we should not release SDK versions for which there are not corresponding engine releases!

kjuulh:
We'll handle this using the actual release process Closing here for now

PR #8094, opened 2024-08-02
💥feat(sdk/elixir): rework on api for declaring a function

wingyplus:
I would like to introduce 2 breaking changes:

First, change the `use Dagger.Mod` into `use Dagger.Mod.Object` to specific this module implement the object type. This breaking changes bring the api to looks similar with other SDKs (Python, TypeScript, etc.).  Here's the example:

before:

```elixir
defmodule Potato do
  use Dagger.Mod, name: "Potato"

  #...
end
```

after:

```elixir
defmodule Potato do
  use Dagger.Mod.Object, name: "Potato"

  # ...
end
```

The second break change is change the api on how to declare a function to use macro instead of module attribute. 

before:

```elixir
@function [
  args: [string_arg: :string],
  return: :string
]
def container_echo(self, args) ...
```

after:

```elixir
function [string_arg: String.t()], String.t()
def container_echo(self, args) ...
```

The macro `function` accepts 2 parameters, first, the function argument, and the second, the return type. Both parameters are use Elixir [typespec](https://hexdocs.pm/elixir/typespecs.html) to declare the type, but limit support and it will convert to data to uses for register the module. Currently the type allows only required value (not allowed nil) but will support optional and default later on.


TomChv:
> The macro function accepts 2 parameters, first, the function argument, and the second, the return type. Both parameters are use Elixir [typespec](https://hexdocs.pm/elixir/typespecs.html) to declare the type, but limit support and it will convert to data to uses for register the module. Currently the type allows only required value (not allowed nil) but will support optional and default later on.

Why this change? Is it better for elixir? I'm not too familiar with the language and it seems it improves the capabilities but I would like more information about that.

wingyplus:
> > The macro function accepts 2 parameters, first, the function argument, and the second, the return type. Both parameters are use Elixir [typespec](https://hexdocs.pm/elixir/typespecs.html) to declare the type, but limit support and it will convert to data to uses for register the module. Currently the type allows only required value (not allowed nil) but will support optional and default later on.
> 
> Why this change? Is it better for elixir? I'm not too familiar with the language and it seems it improves the capabilities but I would like more information about that.
> 
> It might be better too if you support backward compatibility, do you think it's possible?

With the macro I can enable multiple things:

* Shorten syntax compare to previous version as you see in the example. Even me is still hard to remember on how to define type to the function.
* I can shorten on how to define argument. The previous version need to annotate the type via datastructure like `{:nullable, type}` for optional argument. But with macro, we can do something like `String.t() | nil` like Python or TypeScript. :)
* I choose the typespec because it's the standard thing that Elixir ecosystem do to define a function type. So it should be hard when they read the syntax.

wingyplus:
i think I discovered the way on define a function better than using this way. I write a proposal on discord to ask for opinion at https://discord.com/channels/707636530424053791/1133501257051553845/1270802649817677897. If the prototype work it well, we can replace this PR with that new syntax.

jedevc:
FYI @wingyplus just wanted to mention something around modules - up to you whether you want to follow it I guess.

We have the versioning compat work stuff, with the idea that "no published module will break" - so the go/python/typescript SDKs and trying to do this, so when we introduce breaking changes, we keep the old behavior for old `engineVersion`s.

I guess it's up to you whether you feel you want to support that at the moment - especially since elixir module support is significantly newer, I think it's fine for it to be a bit more unstable at the moment.

wingyplus:
> FYI @wingyplus just wanted to mention something around modules - up to you whether you want to follow it I guess.
> 
> We have the versioning compat work stuff, with the idea that "no published module will break" - so the go/python/typescript SDKs and trying to do this, so when we introduce breaking changes, we keep the old behavior for old `engineVersion`s.
> 
> I guess it's up to you whether you feel you want to support that at the moment - especially since elixir module support is significantly newer, I think it's fine for it to be a bit more unstable at the moment.

Thanks for your information. As you mentioned, the Elixir SDK is experimental and the module is still very new and a lot of works I made is not suitable with the module concept. 🙈 

By the way, since Elixir are using git based installation, unlike 3 main SDKs that bundle in the engine, user can pin the version to older version before migrate to a new one and it should work because module code generation and sdk code generation are run separately.

jedevc:
Yup, that sounds perfect to me as is - just wanted to raise in case that was something you were interested in!

wingyplus:
I'll close this PR now in favor of #8172. 🙏

PR #8096, opened 2024-08-02
Fix enum list in CLI

helderco:
Reported by user in [Discord](https://discord.com/channels/707636530424053791/1266410249565372498).

> something changed, and ive been out of loop for some time as i was not paying full attention, i was defining a go type, to simulate an enum in some ways, i.e type lang = string and then lang = Go, lang = "Csharp" etc etc, I was then using this custom type (which still is a string) as part of the dagger function API, since upgrading it no longer works and ive just rid of that concept and used pure strings instead. I'm more of a c# person, but im still interested to know if it was expected?
>
> ```go
> type Language string
> 
> const (
>     CSharp Language = "CSharp"
>     Go     Language = "Go"
>     Java   Language = "Java"
>     Python Language = "Python"
> )
> 
> ...
> 
> GenerateKiotaClients(
>     ctx context.Context,
>     languages []Language, ... )
> ```
>
> this used to work, now i replaced with []string instead

### DIFF
--- a/.changes/unreleased/Fixed-20240808-180607.yaml
+++ b/.changes/unreleased/Fixed-20240808-180607.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Fixed enum lists in modules
+time: 2024-08-08T18:06:07.0133Z
+custom:
+    Author: helderco
+    PR: "8096"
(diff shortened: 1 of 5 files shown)


helderco:
Added the tests.

jedevc (approved):

PR #8099, opened 2024-08-03
feat(sdk/elixir): make dag instance be global

wingyplus:
~NOTE: This requires #8094 to be merge before review this PR.~

Instead of setting `dag` to the struct of the object type, use GenServer
to store it during invoke the function. 

This changeset still passing support using dag instance via a struct but the code from codegen changes to use `dag()` function and ignore the struct.

~This is another breaking changes, since user don't need to declare a struct 
that containing dag instance and function parameters is reduce to only 1 parameter.~


wingyplus:
I think I found the way to make it happens without breaking changes. Will fix it tonight.

jedevc (commented):
This LGTM once the linting/test jobs can pass :tada:

(adding `needs/changelog`, since it would be good to get this in release notes)

wingyplus:
@jedevc Sorry to keep you waiting. I fix the lint failure and add the changelog. 🙏

jedevc (approved):

PR #8103, opened 2024-08-05
test(sdk/elixir): port integration tests from core/integration to the sdk

wingyplus:
I try port the integration tests from `core/integration` to the Elixir SDK by using Python and TypeScript as a role model. 

The hard part is that I cannot use `--sdk=..` is not working since we merged the relocated runtime Pull Request. I'm not sure why it is not working though. So I decided to set the SDK path and commit though the environment variable and let the test suite run against that version.


wingyplus:
Closing this for now.

Issue #6964, opened 2024-03-28
🐞 Optional fields from Python Module are mandatory in Go


skycaptain:
### What is the issue?

I have two modules, `ModuleA` written in Python using string fields with default values. I depend on this module in my second module `ModuleB` written in Golang. It seems that the generated code in `dagger.gen.go` in `ModuleB` does not respect the default values of the fields from `ModuleA`, and requires to set values for these fields again.

### Dagger version

dagger v0.10.2 (registry.dagger.io/engine) linux/amd64

### Steps to reproduce

I've setup https://github.com/skycaptain/dagger-issue-6964 to illustrate the issue, where this line fails: https://github.com/skycaptain/dagger-issue-6964/blob/c53a88810c6a179392731555ad81a016ba4fe602/main.go#L6

helderco:
Yeah, I need to rethink this:
- https://github.com/dagger/dagger/issues/6749

Right now the SDKs are consistent in translating to and from GraphQL, but I need to think a bit more on cross-language DX.

What's happening is that while an argument with a default makes it optional in Python, the internal APIs that the SDKs use equate "optional" to mean "nullable". In fact, in GraphQL, a nullable field is in indeed optional, but it's also possible to have a non-null type with a default value which also makes it optional.

So when you have this:

```python
@object_type
class ModuleA:
    greeting: str = "Hello"
    name: str = "World"
```

It creates the following constructor in the schema:

```graphql
type Query {
    moduleA(greeting: String! = "Hello", name: String! = "World"): ModuleA!
}
```

And so Go is interpreting those non-null types as being required, even though they have defaults. It's equivalent to the following Go constructor:

```go
func New(
    // +default="Hello"
    greeting string,
    // +default="World"
    name string,
) *ModuleA {
}
```

To make those arguments optional, you have to add `// +optional`:

```go
func New(
    // +default="Hello"
    // +optional
    greeting string,
    // +default="World"
    // +optional
    name string,
) *ModuleA {
}
```

Which, cycling back to Python, the equivalent being:

```python
@object_type
class ModuleA:
    greeting: str | None = "Hello"
    name: str | None = "World"
```

The Go SDK’s codegen (for the client) probably needs to consider arguments with defaults to be optional, not just nullables.

[...]

PR #8106, opened 2024-08-05
fix(sdk/go): consider default values when grouping optional arguments in codegen

helderco:
- Part of https://github.com/dagger/dagger/issues/6749
- Fixes https://github.com/dagger/dagger/issues/6964

> [!warning]
> This is a breaking change for **Go SDK** users when depending on a module that has a non-null function argument with a default value. With this change, the argument will change from required to optional in the client bindings (`dagger.gen.go`).

## Example

If a Go Module depends on another Module that has a non-null Function argument with a default value, the generated client bindings would consider that argument required because it didn’t consider having a default, just that it was nullable or not:

```go
// ❌ Before v0.13.0 it's a required argument
dag.Greeter("Dagger").Greeting(ctx)

// ✅ After v0.13.0 it's an optional argument
dag.Greeter().Greeting(ctx)                             // output: Hello, World!
dag.Greeter(GreeterOpts{Name: "Dagger"}).Greeting(ctx)  // output: Hello, Dagger!
```

Example Python code for the *Greeter* dependency:
```python
import dagger

@dagger.object_type
class Greeter:
    name: str = "World"

    @dagger.function
    def greeting(self) -> str:
        return f"Hello, {self.name}!"                
```

And corresponding generated API schema (for the constructor):
```graphql
type Query {
    greeter(name: String! = "World")
}
```

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -403,6 +403,13 @@ class Minimal {
 }
 
 func (ModuleSuite) TestOptionalDefaults(ctx context.Context, t *testctx.T) {
+	// Test expressiveness for following schema:
+	//   a: String!
+	//   b: String
+	//   c: String! = "foo"
+	//   d: String = null
+	//   e: String = "bar"
+
 	for _, tc := range []struct {
 		sdk      string
 		source   string
@@ -422,10 +429,8 @@ func (m *Test) Foo(
 	b *string,
 	// +default="foo"
 	c string,
-	// +optional
 	// +default=null
 	d *string,
-	// +optional
 	// +default="bar"
 	e *string,
 ) string {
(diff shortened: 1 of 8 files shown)


core/integration/module_test.go
```diff
@@ -1890,6 +1895,92 @@ class Test {
...
+import "fmt"
+
+type Dep struct {}
+
+func (m *Dep) Ctl(
+	a string,
+	// +optional
+	b *string,
+	// +default="foo"
+	c string,
```
jedevc, line 1912: So to check my understanding:
- With just `optional`, this would be `String` instead of `String!` (which marks it as required in graphql)
- With just `default`, this would be `String! = "foo"`
- With both, we get `String = "foo"`

So with this change, *all* of the above are interpreted as "optional" for go? That makes sense to me, it brings things more inline with the other SDKs, and go doesn't realllly have either concept, so it makes sense to blur them a bit imo.

jedevc (commented):
When are we interested in landing this change? I *think* this makes sense in a patch release, like v0.12.5, as long as we have the version compat changes as well.

If we need to land it in v0.13, then we should actually be able to do that - we can add future versions in calls to `CheckVersionCompatibility`, and then our tests just need to manually set the `engineVersion` to test against that future incoming behavior.

core/integration/module_test.go
helderco: > So with this change, _all_ of the above are interpreted as "optional" for go? 

Yes, that's correct. 

`!` doesn't mean "required", it just means "doesn't allow `null`". An optional value in GraphQL is one that either has a default value or is nullable. That's because GraphQL implicitly adds `null` as the default value for any nullable without a default.

helderco:
> When are we interested in landing this change? I _think_ this makes sense in a patch release, like v0.12.5, as long as we have the version compat changes as well.

I would think so too, considering also that this is actually a bug. But my concern is that even after the version compat in place, it'll require changing code if you want to work on an existing module, and some people may feel like a breaking change because of it. So the line is a little blurred for me there, but it does make sense that a breaking change is defined by requiring changes to the module in order to run/compile, simply from updating a dependency (in this case Dagger).

> If we need to land it in v0.13, then we should actually be able to do that - we can add future versions in calls to `CheckVersionCompatibility`, and then our tests just need to manually set the `engineVersion` to test against that future incoming behavior.

Yeah, that makes sense. Forgot about this, which actually helps not having to hold on to PRs until the last minute.

gerhard:
If you still want us to get this in `v0.13.0` @helderco, will need to get it in a mergeable state today.

If you don't get to it, that's OK, there is always the next minor 👍

helderco:
This is ready for merge now. Rebased and waiting for checks to pass.

jedevc (approved):

Issue #7169, opened 2024-04-23
🐞 Directory.glob doesn't seem to be working as expected


marcosnils:
### Discussed in https://github.com/dagger/dagger/discussions/7156

<div type='discussions-op-text'>

<sup>Originally posted by **skycaptain** April 22, 2024</sup>
I may be missing something simple, but I'm currently struggling with `Directory.glob` returning an empty list instead of the files. Consider this example where I want to match all the files in the `deploy/sdk` directory. Note the different `tmp-*` directories:

```python
mydir = (
    dag.directory()
    .with_new_file("build/tmp-build-a/deploy/sdk/test.txt", "FOO")
    .with_new_file("build/tmp-build-a/deploy/sdk/test.manifest", "BAR")
    .with_new_file("build/tmp-build-b/deploy/ipk/test.ipk", "BAZ")
)

await mydir.glob('build/tmp-*/deploy/sdk/*')
```

This example returns `[]`. Running the same command with regular `ls` works fine. I tried different combinations of the pattern, but none have worked as expected. What could I be missing?</div>

marcosnils:
cc @TomChv

cwlbraa:
I did a little digging here, there may be 2 overlapping bugs in dir.Glob. 

first,
```
entries, err := ref.ReadDir(ctx, bkgw.ReadDirRequest{
		Path:           path.Join(dir.Dir, src),
		IncludePattern: pattern,
	})
```
doesn't appear to handle single stars in directory specifications.

second, even using an input like `build/**/deploy/sdk/*`,
```
if entry.IsDir() {
			subEntries, err := dir.Glob(ctx, entryPath, pattern)
			if err != nil {
				return nil, err
			}
			paths = append(paths, subEntries...)
		}
```
may not recurse correctly. after matching `tmp-build-*`, it then tries to match the same pattern, `build/**/deploy/sdk/*` against the subentry (which now doesn't include the `/build/` part)

PR #5824, opened 2023-09-25
feat: improve entries with include and recursive

TomChv:
Support include pattern.
Support recursive option.
Add tests

Replace and complete #5686 

Related to: #5685

PR #8108, opened 2024-08-05
core: fix Directory.glob doesn't seem to be working as expected

khrisrichardson:
`IncludePattern` was added to the `ReadDirRequest` options as per [this discussion](https://github.com/dagger/dagger/pull/5824#pullrequestreview-1644134442) during the initial implementation of `Directory.Glob`, and set to the same value as the `Directory.Glob` `pattern` argument.

While this works as intended for globstar patterns like `**/*.md`, it has the unintended side-effect of excluding sub-directories from being recursed for less well behaved patterns. Take a simple example: if the pattern is `core/**/*.go`, as soon as the recursive call to `Glob` encounters the `core/` sub-directory, reads of that sub-directory with the same pattern will fail unless it also has a sub-directory named `core`.

As confirmed by running an [engine service](https://github.com/dagger/dagger/tree/main/dev#run-the-engine-service) with this commit, updating `IncludePattern` to only consider the rightmost part of the pattern fixes the behavior identified in https://github.com/dagger/dagger/issues/7169.

It seems the `patternmatcher.MatchesOrParentMatches` is more geared to exclusion, so to overcome the cases where the parent matches the pattern (as the name of the function alludes) but the base does not, filtering `ReadDir` with the `IncludePattern` manipulated as described ensures the base is also considered when evaluating the glob.

It's probably worthwhile considering whether negated patterns are intended to work, because if so none of the integration tests cover that scenario and this workaround will probably not have the intended effect.

### DIFF
--- a/.changes/unreleased/Fixed-20240805-164529.yaml
+++ b/.changes/unreleased/Fixed-20240805-164529.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: 'engine: fix `Directory.Glob` doesn''t seem to be working as expected'
+time: 2024-08-05T16:45:29.636669914-07:00
+custom:
+    Author: khrisrichardson
+    PR: "8108"
--- a/core/directory.go
+++ b/core/directory.go
@@ -311,7 +311,7 @@ func (dir *Directory) Glob(ctx context.Context, src string, pattern string) ([]s
 
 	entries, err := ref.ReadDir(ctx, bkgw.ReadDirRequest{
 		Path:           path.Join(dir.Dir, src),
-		IncludePattern: pattern,
+		IncludePattern: filepath.Join("**", filepath.Base(pattern)),
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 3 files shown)


jedevc:
Heya @khrisrichardson thanks thanks thanks for picking this up :pray: :heart: 

Unfortunately, I think this isn't quite the right solution - I continued pulling on the thread in #8110, hope that's alright!

> Take a simple example: if the pattern is `core/**/*.go`, as soon as the recursive call to `Glob` encounters the `core/` sub-directory, reads of that sub-directory with the same pattern will fail unless it also has a sub-directory named `core`.

Yeah you're entirely right here, yup. I somehow missed this on suggestion/review, good catch.

I think this is right, but the proposed fix isn't *quite* correct - I think `filepath.Join("**", filepath.Base(pattern)),` would now start to dive into every directory, which is potentially quite costly. The reason for the really complicated `IncludePatterns` handling is so that we can avoid doing that traversal.

Thankfully, I think we can rework the method! We can avoid the root cause - which is that we continually manipulate `Path`, without changing `IncludePatterns`.

khrisrichardson:
@jedevc no offense taken whatsoever. I was also inclined to add a walk function, but wasn't quite sure how to add one in the context of *Directory. Nice work

PR #8110, opened 2024-08-06
fix: ensure glob works with subdir recursive entries

jedevc:
> [!NOTE]
> 
> This is an alternative to https://github.com/dagger/dagger/pull/8108 - thanks @khrisrichardson for doing most of the hard investigation, but I think it was worth doing the restructuring of glob entirely, it needed doing here anyways. I've added you as a co-author to this PR :tada:

Fixes https://github.com/dagger/dagger/issues/7169

Globs weren't correctly working - the root issue here was that `IncludePatterns` is interpreted as relative to the `Path` passed into `ReadDir`.

This is kinda not great, since we keep the same `IncludePatterns`, but continually adjust `Path` to go deeper.

To fix the issue, we entirely restructure globbing - now the entire filesystem is only mounted once (hopefully a performance improvement in its own right), and we apply a custom walking function, instead of needing to call `ReadDir` multiple times.

One thing that also seems to have been broken was using `MatchesOrParentMatches` - this potentially included things in the glob that we didn't want. To fix this, we instead switch to *only* matching on the full pathname - which is the behavior we wanted in the first place.

### DIFF
--- a/core/directory.go
+++ b/core/directory.go
@@ -269,12 +269,7 @@ func (dir *Directory) Entries(ctx context.Context, src string) ([]string, error)
 }
 
 // Glob returns a list of files that matches the given pattern.
-//
-// Note(TomChv): Instead of handling the recursive manually, we could update cacheutil.ReadDir
-// so it will only mount and unmount the filesystem one time.
-// However, this requires to maintain buildkit code and is not mandatory for now until
-// we hit performances issues.
-func (dir *Directory) Glob(ctx context.Context, src string, pattern string) ([]string, error) {
+func (dir *Directory) Glob(ctx context.Context, pattern string) ([]string, error) {
 	svcs, err := dir.Query.Services(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get services: %w", err)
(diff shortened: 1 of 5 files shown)


TomChv (approved):
Add change note and good to merge it

core/directory.go
```diff
@@ -303,43 +298,38 @@ func (dir *Directory) Glob(ctx context.Context, src string, pattern string) ([]s
...
+	var paths []string
+	err = ref.WalkDir(ctx, buildkit.WalkDirRequest{
+		IncludePattern: pattern,
+		Path:           dir.Dir,
+		Callback: func(path string, info *fstypes.Stat) error {
+			// HACK: ideally, we'd have something like MatchesExact, which
+			// would skip the parent behavior that we don't really want here -
+			// oh well, let's just fake it with false
+			//nolint:staticcheck
+			match, err := pm.MatchesUsingParentResult(filepath.Clean(path), false)
```
jedevc, line 320: We might be able to replace this with https://github.com/moby/patternmatcher/pull/3, if that gets merged :tada:

PR #8112, opened 2024-08-06
PHP: Use phpcs for PhpSdkDev Linting

charjr:
Add PHP Code Sniffer
  - Configured to lint against standards: PSR1 and PSR12

Can be called through the PhpSdkDev.Lint

Added to to Dev module in the same way as Python's Linting works.


sdk/php/composer.json
```diff
@@ -29,7 +29,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^11.0",
-        "symfony/console": "^6.3|^7.0",
```
charjr, line 32: `symfony/console` has been removed from `require-dev` because it is already in the `require` section, making this line unnecessary as the package is already required.

sdk/php/src/Connection/version.php
```diff
@@ -1 +1,3 @@
-<?php /* Code generated by dagger. DO NOT EDIT. */ return '0.12.4';
+<?php
+
+/* Code generated by dagger. DO NOT EDIT. */ return '0.12.4';
```
helderco, line 3: This might produce a linting error when the version is bumped via the release process. Should update the code that generates this in `/dev/sdk_php.go` so that it doesn't produce a linting error.

helderco (commented):
Looks good! 

If you make a follow up to delegate codegen to `/sdk/php/dev`, then we don't need the duplicated base in `/dev/sdk_php.go` anymore. 🙂

dev/sdk_php.go
```diff
@@ -25,12 +27,28 @@ type PHPSDK struct {
...
+		_, err := dag.PhpSDKDev().Lint(src).Sync(ctx)
 		return err
-	}
-	return dag.Dirdiff().AssertEqual(ctx, before, after, []string{filepath.Join(phpSDKPath, phpSDKGeneratedDir)})
+	})
+
+	eg.Go(func() error {
+		before := t.Dagger.Source()
+		after, err := t.Generate(ctx)
+
```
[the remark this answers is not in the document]
charjr: I've now made this change. 

For future reference, may you please explain why this change was suggested?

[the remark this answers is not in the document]
helderco: Not important. It's just more conventional to put the usual `if err != nil { return nil }` error handling just below the statement that returns the error, especially if it's a one liner.

sdk/php/src/Connection/version.php
helderco: FTR, it just means you need to adjust the following to look like that (add two `\n` after `<?php`):

https://github.com/dagger/dagger/blob/0f8b9d58535fcf65d560448ce4e31dfebecdfdac/dev/sdk_php.go#L102-L103

dev/sdk_php.go
```diff
@@ -99,8 +116,10 @@ func (t PHPSDK) Publish(
 // Bump the PHP SDK's Engine dependency
 func (t PHPSDK) Bump(ctx context.Context, version string) (*dagger.Directory, error) {
 	version = strings.TrimPrefix(version, "v")
-	content := fmt.Sprintf(`<?php /* Code generated by dagger. DO NOT EDIT. */ return '%s';
-`, version)
+	content := fmt.Sprintf(
+		"<?php\n\n/* Code generated by dagger. DO NOT EDIT. */return '%s';",
```
helderco, line 120: You removed a space there. And should also probably add a newline, according to the linter:
```
FILE: /src/sdk/php/src/Connection/version.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 3 | ERROR | [x] Expected 1 newline at end of file; 0 found
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```
```suggestion
		"<?php\n\n/* Code generated by dagger. DO NOT EDIT. */\nreturn '%s';\n",
```

sdk/php/dev/src/PhpSdkDev.php
```diff
@@ -28,6 +28,11 @@ public function test(
         );
     }
 
+    #[DaggerFunction('Run linter in source directory')]
+    public function lint(Directory $source): Container {
+        return $this->base($source)->withExec(['phpcs']);
+    }
```
helderco, line 34: You need a "format" function as well. It'll make it easier for other maintainers that get PHP linting errors, to run the formatter in order to solve them.

You can return a Directory with the diff of the changes, like this:

https://github.com/dagger/dagger/blob/b0d27e15843341643954a4e9ef0841cd785b78b3/sdk/python/dev/src/main/__init__.py#L186-L187

helderco (approved):
It'd be great if you can include the formatter, but don't want to block on it so you can do it in a follow up.

sdk/php/dev/src/PhpSdkDev.php
charjr: I've added #8124 to ensure I get that done in a follow up.

If we can get this merged separately the `format()` should be a fairly small PR.

charjr:
@helderco I've created two issues for the suggested follow ups: #8124 and #8125.

I will do those separately so we can keep the PRs small and easier for review.

PR #8113, opened 2024-08-06
chore: update TypeScript dependencies

TomChv:
execa will have to wait for bun@v1.1.22 since https://github.com/oven-sh/bun/issues/11716 has been merge.


sdk/typescript/runtime/main.go
```diff
@@ -93,6 +93,7 @@ func (t *TypescriptSdk) ModuleRuntime(ctx context.Context, modSource *ModuleSour
 		return ctr.
 			// Install dependencies
 			WithExec([]string{"yarn", "install"}).
+			WithExec([]string{"npm", "pkg", "set", "type=module"}).
```
gerhard, line 96: Why is this necessary?

sdk/typescript/runtime/template/package.json
```diff
@@ -1,6 +1,7 @@
 {
+  "type": "module",
```
gerhard, line 2: I am assuming that this change is related to `npm pkg set type=module`. Wondering the same thing here.

sdk/typescript/.changes/unreleased/Changed-20240806-223717.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Changed
+body: |
+  Drop CommonJS support in favor of Module. May be a breaking change if you use very old modules.
```
gerhard, line 3: I am wondering:
1. What is a "very old module"?
2. Is there any way that can make this change without breaking "very old modules"?
3. How many "very old modules" are available on https://daggerverse.dev? We may need @grouville @jpadams @vito @marcosnils help to answer this.

gerhard (commented):
This looks decent **apart** from the potential breaking change.

It's worth capturing a few answers before we push forward with this.

Anything else to add @helderco @jedevc ?

sdk/typescript/.changes/unreleased/Changed-20240806-223717.yaml
jedevc: I think we should very much try and do 2. If we do go forward with this specific change, we need to only apply it based on the contents of the `engineVersion` - if it's larger than "vX.Y.Z", we can set this, otherwise, we don't.

Can you give an example of a module that *would* break?

TomChv:
https://github.com/oven-sh/bun/releases/tag/bun-v1.1.22 has been release, I'll update this PR with execa changes soon

sdk/typescript/runtime/main.go
TomChv: Because of `graphql-request`, it dropped the support for CommonJS, in order to be as smooth as possible with the transition, I set the package.json type to module in the runtime so an user who didn't add this entry to its package will see it working.

sdk/typescript/runtime/template/package.json
TomChv: Same reason, for new init :D

sdk/typescript/.changes/unreleased/Changed-20240806-223717.yaml
TomChv: > Can we make this change without breaking "very old modules"?

Sadly not, except if we use different `graphql-request` version depending on the module's type, but that seems hacky.

I don't have an example in mind, there's not much packages that use commonJS anymore.

According to this [issue](https://github.com/nodejs/node/issues/33222), it doesn't look backward compatible but again, I don't know

gerhard:
Those replies were helpful @TomChv, thank you!

If we can't make this change without breaking "very old modules", I think that we need to:
1. Find out how many modules will be affected in the https://daggerverse.dev
2. Split this change into a non-breaking one, and a breaking one
3. Merge the non-breaking change & ship in a patch
4. Keep the breaking change for the next minor, `v0.13.0`

TomChv:
@gerhard I updated the PR to remove the breaking change, this PR will be ready soon :D

gerhard (approved):
LGTM 👍 🚀

PR #8114, opened 2024-08-06
feat: compute file digest

### LINKED ISSUE #7709: Get a file's digest
## Problem

The Dagger API doesn't provide an efficient way to compute a file's digest. Which, in turn, makes it impossible to efficiently compare two files, or check if a file has changed for example. There are workarounds, but they either add complexity, hurt performance, or both.

## Solution

Add a function to compute a file's digest. Ideally reusing the underlying buildkit digest.

```graphql
extend type File {
  """
  Return the file's digest.
  The format of the digest is not guaranteed to be stable between releases of Dagger. It is guaranteed to be stable between invocations of the same Dagger engine.
  """
  digest: String!
}
```

TomChv:
Fixes #7709

### DIFF
--- a/.changes/unreleased/Added-20240806-231920.yaml
+++ b/.changes/unreleased/Added-20240806-231920.yaml
@@ -1,5 +1,11 @@
 kind: Added
-body: Add `Digest` method to `File`
+body: |
+  Add `Digest` method to `File`.
+
+  This method provides an efficient way to compute a file's digest, which unlock optimized file comparison
+  or check for file changes.
+  This function benefits from Buildkit's cache system and include file's metadata in the digest computation.
+  There's also an option `excludeMetadata` to compute the file's digest from its content only.
 time: 2024-08-06T23:19:20.37563+02:00
 custom:
   Author: TomChv
--- a/.changes/unreleased/Added-20240806-231920.yaml
+++ b/.changes/unreleased/Added-20240806-231920.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: Add `Digest` method to `File`
+time: 2024-08-06T23:19:20.37563+02:00
+custom:
+  Author: TomChv
+  PR: "8114"
--- a/core/file.go
+++ b/core/file.go
@@ -11,6 +11,7 @@ import (
 	"github.com/moby/buildkit/client/llb"
 	bkgw "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/solver/pb"
+	"github.com/opencontainers/go-digest"
 	fstypes "github.com/tonistiigi/fsutil/types"
 	"github.com/vektah/gqlparser/v2/ast"
 
@@ -190,6 +191,15 @@ func (file *File) Contents(ctx context.Context) ([]byte, error) {
 		offset += len(chunk)
 	}
 	return contents, nil
+}
+
+func (file *File) Digest(ctx context.Context) (string, error) {
+	content, err := file.Contents(ctx)
+	if err != nil {
+		return "", fmt.Errorf("failed to get file content: %w", err)
+	}
+
+	return digest.FromBytes(content).String(), nil
 }
 
 func (file *File) Stat(ctx context.Context) (*fstypes.Stat, error) {
--- a/core/file.go
+++ b/core/file.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"io"
 	"io/fs"
@@ -11,6 +12,7 @@ import (
 	"github.com/moby/buildkit/client/llb"
 	bkgw "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/solver/pb"
+	"github.com/opencontainers/go-digest"
 	fstypes "github.com/tonistiigi/fsutil/types"
 	"github.com/vektah/gqlparser/v2/ast"
 
@@ -192,18 +194,36 @@ func (file *File) Contents(ctx context.Context) ([]byte, error) {
 	return contents, nil
 }
 
-func (file *File) Digest(ctx context.Context) (string, error) {
-	result, err := file.Evaluate(ctx)
-	if err != nil {
-		return "", fmt.Errorf("failed to evaluate file: %w", err)
+func (file *File) Digest(ctx context.Context, excludeMetadata bool) (string, error) {
+	// If metadata are included, directly compute the digest of the file
+	if !excludeMetadata {
+		result, err := file.Evaluate(ctx)
+		if err != nil {
+			return "", fmt.Errorf("failed to evaluate file: %w", err)
+		}
+	
+		digest, err := result.Ref.Digest(ctx, file.File)
+		if err != nil {
+			return "", fmt.Errorf("failed to compute digest: %w", err)
+		}
+
+		return digest.String(), nil
 	}
 
-	digest, err := result.Ref.Digest(ctx, file.File)
+	// If metadata are excluded, compute the digest of the file from its content.
+	reader, err := file.Open(ctx)
 	if err != nil {
-		return "", fmt.Errorf("failed to compute digest: %w", err)
+		return "", fmt.Errorf("failed to open file to compute digest: %w", err)
 	}
 
-	return digest.String(), nil
+	defer reader.Close()
+
+	h := sha256.New()
+	if _, err := io.Copy(h, reader); err != nil {
+		return "", fmt.Errorf("failed to copy file content into hasher: %w", err)
+	}
+
+	return digest.FromBytes(h.Sum(nil)).String(), nil
 }
 
 func (file *File) Stat(ctx context.Context) (*fstypes.Stat, error) {
--- a/core/file.go
+++ b/core/file.go
@@ -201,7 +201,7 @@ func (file *File) Digest(ctx context.Context, excludeMetadata bool) (string, err
 		if err != nil {
 			return "", fmt.Errorf("failed to evaluate file: %w", err)
 		}
-	
+
 		digest, err := result.Ref.Digest(ctx, file.File)
 		if err != nil {
 			return "", fmt.Errorf("failed to compute digest: %w", err)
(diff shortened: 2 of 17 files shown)


core/file.go
```diff
@@ -192,6 +193,15 @@ func (file *File) Contents(ctx context.Context) ([]byte, error) {
 	return contents, nil
 }
 
+func (file *File) Digest(ctx context.Context) (string, error) {
+	content, err := file.Contents(ctx)
```
sipsma, line 197: This will require loading the whole file into memory first, which is bad if it's a large file (100s of MB or greater).

We can actually just stream the file contents though which should avoid this. The existing `file.Open` method returns a `io.ReadCloser` which we can stream right into a `sha256` hasher, e.g. https://pkg.go.dev/crypto/sha256#example-New-File

core/schema/file.go
```diff
@@ -25,6 +25,12 @@ func (s *fileSchema) Install() {
 			Doc(`Retrieves the size of the file, in bytes.`),
 		dagql.Func("name", s.name).
 			Doc(`Retrieves the name of the file.`),
+		dagql.Func("digest", s.digest).
```
sipsma, line 28: One thing worth consideration is if the digest should include the file metadata in addition to the file contents (i.e. owner, permissions, timestamps, etc.). The file metadata is part of its definition, impacts caching, etc. so depending on the use case it may or may not be what the user wants.

We can just add an optional argument for whether or not metadata is included of course, so then the question is just what the default behavior is (metadata included or not).

Personally, I have a slight preference for including metadata in the digest by default, but curious if others have thoughts.

Not a blocker for me, just raising the question.

core/file.go
helderco: Doesn't BuildKit already have a digest that it uses for caching purposes, that we can reuse here?

sipsma: It doesn't have anything as part of its official API that exposes it.

We *could* use it's internals directly I suppose, specifically this `Checksum` func: https://github.com/moby/buildkit/blob/78f24da143968ed6fa4237c25e766b14a96df159/cache/contenthash/checksum.go#L54

We'd need to be careful to provide the right `path` arg (`file.File`) and such, but yeah I think that'd be another option.

Overall it would probably be slightly more convoluted than what I suggested above, but also slightly more reusable for other use cases (especially if it's added as a method on `Result` in [this package](https://github.com/dagger/dagger/blob/main/engine/buildkit/ref.go)), so I don't mind either approach.

core/schema/file.go
jedevc: I think I'd expect the default to just be a content digest - I can imagine a lot of the cases of wanting a digest of the file to just get the contents - especially as that's what lots of users would be use to using `sha256sum` or similar command line tools.

Having this as an option definitely makes sense though (but - there's no authoritative way to hash file metadata in a consistent way - it feels a bit weird to have something non-standard like this, but I don't have any better suggestions).

core/file.go
TomChv: I understand the logic, I see that I can add a method `Digest` in the package to get the digest directly from the cache keys.
However I'm confused about the cache key.

Here's a snippet of my attempt:

```go
func (r *ref) Digest(ctx context.Context) (string, error) {
	cacheResult, err := r.resultProxy.Result(ctx)
	if err != nil {
		return "", err
	}

    // This is an array of `[]ExportableCacheKey` that have a method `Digest` in it, however
    // What should I do with this array? Should I sum all of them in a hasher?? I'm confused 
	keys := cacheResult.CacheKeys()
```

As explained in the code's comment, what should I do with the array? Should I only keep the last entry of the array or should I compute a sum of all digest?

The `ref.Digest` method can then be called from the file proxy service, like for evaluate so that super cool but I'm stuck on the part above.
Let me know what I should do with it.

TomChv: It seems I found a workaround with 

```go
func (r *ref) Digest(ctx context.Context, path string) (digest.Digest, error) {
	cacheRef, err := r.CacheRef(ctx)
	if err != nil {
		return "", err
	}

	sessionGroup := bksession.NewGroup(r.c.ID())

	return contenthash.Checksum(ctx, cacheRef, path, contenthash.ChecksumOpts{}, sessionGroup)
}
```

core/schema/file.go
TomChv: Is it possible with the buildkit cache digest implementation? I didn't see the options to do so

TomChv:
@sipsma I made the changes you asked, hope it improve the implementation :D

TomChv:
PHP linter works locally so I'm quite confused by the fail in the CI

```
dagger call -m dev --source=.:default sdk php lint
Full trace at https://dagger.cloud/Quartz/traces/e045e23d3885374f42d894078e3fce12

✔ connect 1.2s
✔ initialize 7.7s
✔ prepare 1.1s
✔ daggerDev(
    source: ✔ ModuleSource.resolveDirectoryFromCaller(path: ".", viewName: "default"): Directory! 0.0s
  ): DaggerDev! 0.3s
✔ DaggerDev.sdk: DaggerDevSdk! 0.3s
✔ DaggerDevSdk.php: DaggerDevPhpsdk! 0.0s
✔ DaggerDevPhpsdk.lint: Void 24.3s
```

core/schema/file.go
sipsma: > Having this as an option definitely makes sense though (although, there's no authoritative way to hash file metadata in a consistent way - it feels a bit weird to have something non-standard like this, but I don't have any better suggestions).

The closest thing to a standard is the tar format, which includes all file metadata. That's how container layer digests are computed, so I suppose it's part of the OCI standard in that respect.

That's also what the `contenthash` package ends up doing (without creating an actual tar file on disk): https://github.com/sipsma/buildkit/blob/699623575173eedf09b43b797bd74439c86a3071/cache/contenthash/filehash.go#L47-L47

> Is it possible with the buildkit cache digest implementation? I didn't see the options to do so

The `Checksum` implementation you switched to does include the file metadata in addition to the file contents. You're right it doesn't currently have an option to disable inclusion of metadata.

Like I said before, I'm personally fine with including the metadata by default and adding an option down the line to enable digesting only based off the contents. There's a few approaches for doing that (not needed now IMO, just noting for the future):
* Upstream changes to the buildkit package to support that
* When that "contents-only" option is enabled, reset all the metadata on the file so it's consistent and then get the hash from `Checksum`
* Just have a separate codepath when "contents-only" is enabled where we just read the contents and hash those directly rather than use the `Checksum` function

sipsma: BTW, I do see your point Justin about contents-only being the expectation from using `sha256sum` and similar.

I feel torn overall still because in the rest of Dagger the file metadata does impact the "checksum" of a file (and caching etc.) so including it by default feels aligned with the rest of our behavior. Really splitting hairs here though.

So still fine with including it by default, but not strongly opinionated if there's any more thoughts on this topic.

sipsma (approved):
Code LGTM, but would like to ensure others are okay with the behavior of including metadata by default before merging: https://github.com/dagger/dagger/pull/8114/files#r1706167239

core/schema/file.go
TomChv: I added the `excludeMetadata` options with the separate codepath :D

.changes/unreleased/Added-20240806-231920.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Added
+body: Add `Digest` method to `File`
```
jedevc, line 2: :eyes: can we get another line in the changelog body here explaining roughly what this method does and what it's for?

I'd really like to have users read release notes and get excited about the features/fixes we're shipping, and having a bit more of an explanation for each point goes a long way to getting that :tada: :tada:

TomChv: Sure

core/integration/file_test.go
```diff
@@ -393,6 +393,26 @@ func (FileSuite) TestContents(ctx context.Context, t *testctx.T) {
 	}
 }
 
+func (FileSuite) TestDigest(ctx context.Context, t *testctx.T) {
```
jedevc, line 396: Can we expand these tests? Ideally each test should have have two files with the same contents, but different metadata.

Then in "without metadata", the files will have the same digest, but in "with metadata" they'll be different.

jedevc (approved):
Left a new nits, but good to merge IMO once those are handled :tada:

I *think* it's fine to default to included metadata (I don't have a good enough reason to block), but let's just keep an eye on feedback, we can always change this in the future if it turns out to be super confusing for users.

core/integration/file_test.go
TomChv: Done :D

TomChv:
I asked for help in the Discord php channel for the lint failure, we need to fix that before merging.

TomChv:
Test failed because of flacky Python, not related to the PR

https://dagger.cloud/dagger/traces/1ca0b885c9ff83203ba1f4faf1c25ea7?span=96c559bed31fd793

TomChv:
PR still failing because of flacky TestGit & TestGoInit, should I merge that PR even if it fails these? My changes are not impacting this part.

core/schema/file.go
vito: If I had to pick a side I think I'd argue for alignment with `sha56sum` being less surprising, and if we want a digest that includes the metadata and anything else that influences caching, maybe that should be something like `cacheKey`.

Going back to the original request (#7709) it sounds like one goal is to see if two files are "the same" - which I think most people would take to mean same contents, metadata aside, the same way `sha256sum` etc. works.

But, yeah, low conviction and I'm coming in late so I'm good with shipping in either form

sipsma:
> PR still failing because of flacky TestGit & TestGoInit, should I merge that PR even if it fails these? My changes are not impacting this part.

Yeah I think you are good to merge if it's just those ones

core/schema/file.go
sipsma: Since @TomChv added support for both codepaths, it's trivial to switch the default behavior to not including metadata. Seems like more are leaning towards switching to that default, so we have time before the release to switch it.

PR #8115, opened 2024-08-06
fix using enum in constructor

helderco:
Reported by @sagikazarmark in [Discord](https://discord.com/channels/707636530424053791/1123687185837740053/1270438030284689539).

Enum doesn't work in constructors:
```
Error: make request: input: failed to get schema: failed to get schema for module "test": failed to install constructor: failed to create function: failed to find mod type for function "" arg "status" type
```

See failing tests. Affects all SDKs, the problem is in core (modfunc).

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -3128,7 +3128,16 @@ const (
 	Inactive Status = "INACTIVE"
 )
 
-type Test struct{}
+func New(
+	// +default="INACTIVE"
+	status Status,
+) *Test {
+	return &Test{Status: status}
+}
+
+type Test struct {
+	Status Status
+}
 
 func (m *Test) FromStatus(status Status) string {
 	return string(status)
@@ -3153,6 +3162,8 @@ class Status(dagger.Enum):
 
 @dagger.object_type
 class Test:
+    status: Status = dagger.field(default=Status.INACTIVE)
+
     @dagger.function
     def from_status(self, status: Status) -> str:
         return str(status)
@@ -3191,6 +3202,14 @@ class Status {
 @object()
 export class Test {
   @func()
+  status: Status
+
+  // FIXME: this should be Status.Inactive instead of "INACTIVE"
+  constructor(status: Status = "INACTIVE") {
+    this.status = status
+  }
+
+  @func()
   fromStatus(status: Status): string {
     return status as string
   }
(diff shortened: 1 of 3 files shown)


helderco:
Looks like interfaces in the constructor return a similar error.

jedevc:
Hm, the main issue seems resolved, but the typescript test is still failing:

```
make request: input: failed to get schema: failed to get schema for module "test": failed to install constructor: failed to get field spec: failed to decode default value for arg "status": invalid character 'S' looking for beginning of value
```

I suspect this is because of https://github.com/dagger/dagger/blob/main/sdk/typescript/introspector/scanner/abtractions/argument.ts#L162 - I think `getText` isn't correct here, I guess that's setting the default directly to `Status.Inactive`, which can't be JSON unmarshaled in the core.

TomChv:
> I suspect this is because of [main/sdk/typescript/introspector/scanner/abtractions/argument.ts#L162](https://github.com/dagger/dagger/blob/main/sdk/typescript/introspector/scanner/abtractions/argument.ts?rgh-link-date=2024-08-13T16%3A06%3A14Z#L162) - I think getText isn't correct here, I guess that's setting the default directly to Status.Inactive, which can't be JSON unmarshaled in the core.

That's right, there's no such evalutation by reference in the TS SDK, I'm not sure how to fix it, I'll need to add a new capabilities to the SDK.
We might need to delay this for the next release since I'm stuck with this package manager issue.

core/integration/module_test.go
```diff
@@ -3190,6 +3201,14 @@ class Status {
 
 @object()
 export class Test {
+  @func()
+  status: Status
+
+  // FIXME: this should be Status.Inactive instead of "INACTIVE"
+  constructor(status: Status = "INACTIVE") {
```
jedevc, line 3208: @TomChv this is kinda a nasty hack, but this works to make sure that this actually works :tada:

If it looks good to you, I'll open a follow-up issue, and can try tackling that when I have a bit more time.

TomChv: LGTM, resolving reference by their accessor it's a complex issue, related to TS SDK only, I'll fix that later!

TomChv (approved):

core/integration/module_test.go
jedevc: Spun out to https://github.com/dagger/dagger/issues/8152!

PR #8116, opened 2024-08-07
ci: add helm install tests

jedevc:
This will help avoid issues like https://github.com/dagger/dagger/issues/7916 in the future - this took a lot of messing around to fix, so it would be good to *never* hit this again.

### DIFF
--- a/.dagger/helm.go
+++ b/.dagger/helm.go
@@ -2,37 +2,93 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/dagger/dagger/.dagger/internal/dagger"
 	"github.com/dagger/dagger/.dagger/util"
+	"github.com/moby/buildkit/identity"
 	"helm.sh/helm/v3/pkg/chart"
 	"sigs.k8s.io/yaml"
 )
 
-const (
-	// https://hub.docker.com/r/alpine/helm/tags
-	// Pin image to ref so that it caches better & protects from tag overrides
-	helmImage = "alpine/helm:3.15.2@sha256:761b0f39033ade8ce9e52e03d9d1608d6ca9cad1c7e68dc3e005f9e4e244410e"
-)
-
 type Helm struct {
+	Dagger *DaggerDev        // +private
 	Source *dagger.Directory // +private
 }
 
-func (h *Helm) Test(ctx context.Context) error {
+func (h *Helm) Lint(ctx context.Context) error {
 	_, err := h.chart().
 		WithExec([]string{"helm", "lint"}).
-		WithExec([]string{"helm", "lint", "--debug", "--namespace", "dagger", "--set", "magicache.token=hello-world", "--set", "magicache.enabled=true"}).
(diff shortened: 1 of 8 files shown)


jedevc:
*Hm* - there appears to be something in our CI infra here:
- Using our dagger runners (fail): https://github.com/dagger/dagger/actions/runs/10285227388/job/28462983902?pr=8116
  `RunContainerError: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error setting cgroup config for procHooks process: failed to write "a *:* rwm": write /sys/fs/cgroup/devices/kubepods/besteffort/podac92bc34-526c-43c7-af5a-a80e9fedc5c0/8ed7faec5bc845ae4599a250fdd3d1e6b982d221e018155101ff1de0ac1be9b4/devices.allow: operation not permitted: unknown`
- Using `ubuntu-latest` (success): https://github.com/dagger/dagger/actions/runs/10286345815/job/28466773136?pr=8116

Not 100% sure what would be causing this, cc @matipan @gerhard, any ideas?

gerhard:
I suspect that we need `cgroup v2` for nesting support for: https://github.com/marcosnils/daggerverse/blob/e0bd6b9f5519c49db4b6eb0689927214720976f9/k3s/main.go#L13-L36

To get cgroup v2, we would need to switch from AL2 to either:
- `AL2023` https://github.com/aws/containers-roadmap/issues/1953
- or `Bottlerocket`

We have been talking about this with @matipan a few times in the past. I propose that we create a new experimental runner group which uses this new OS. WDYT @matipan ?

jedevc:
Thanks @gerhard :tada: If that's going to take a bit of time, is there any reason we can't switch these jobs to use `ubuntu-latest`, and later change them back?

At least that way we can get the helm coverage asap!

gerhard:
Yes, `ubuntu-latest` sounds like a good short-term solution 👍

jedevc:
Rebased, and updated to use `ubuntu-latest` - I think this should be alright to merge now? cc @gerhard @marcosnils

.dagger/helm.go
```diff
@@ -2,37 +2,94 @@ package main
...
+
+	test, err := h.chart().
+		WithMountedFile("/usr/bin/dagger", cli).
+		WithServiceBinding("helm-test", k3ssvc).
+		WithFile("/.kube/config", k3s.Config()).
+		WithEnvVariable("KUBECONFIG", "/.kube/config").
+		WithEnvVariable("CACHEBUSTER", identity.NewID()).
+		WithExec([]string{"kubectl", "get", "nodes"}).
+		WithExec([]string{"helm", "install", "--create-namespace", "--namespace=dagger", "dagger", "."}).
+		WithExec([]string{"kubectl", "wait", "--for=condition=Ready", "--timeout=120s", "pod", "--selector=name=dagger-dagger-helm-engine", "--namespace=dagger"}).
```
marcosnils, line 52: nit: this can be removed by adding `--wait` to the `helm install` command.

marcosnils (approved):

PR #8122, opened 2024-08-08
fix: replace strcase library to fix module name transformation related issues

### LINKED ISSUE #7941: 🖐️ Quickstart issues with Go (`dagger init` + `dagger call`)
### What happened? What did you expect to happen?

Hey! I'm testing out Dagger for the first time and followed the [quickstart guide](https://docs.dagger.io/quickstart/daggerize).

When I ran `dagger init --sdk=go` in the root of my existing repository (not `hello-dagger`), it generated a dagger Go module:

```sh
$ tree dagger
dagger
├── .gitattributes
├── .gitignore
├── dagger.gen.go
├── go.mod
├── go.sum
├── internal
│   ├── dagger
│   │   └── dagger.gen.go
│   ├── querybuilder
│   │   ├── marshal.go
│   │   └── querybuilder.go
│   └── telemetry
│       ├── attrs.go
│       ├── env.go
│       ├── exporters.go
│       ├── init.go
│       ├── live.go
│       ├── logging.go
│       ├── proxy.go
│       ├── span.go
│       └── transform.go
└── main.go

4 directories, 18 files
```

So, following the spirit of the quickstart, I tried to call the `Echo` example target with: `dagger call echo --stringArg=ping` (instead of replacing the contents immediately with quickstart-specific functions) and get the following output:

```sh
$ dagger call echo --stringArg=ping
✔ connect 3.6s
✘ initialize 0.5s
! main object not found
  ✔ resolving module ref 0.2s
  ✔ installing module 0.2s
  ✘ analyzing module 0.2s
  ! main object not found

Error: main object not found
```

Same output repeats with `dagger call echo ping` and `dagger call echo`.

What am I missing here? :)

My `./dagger/main.go` looks like such: 

```go
package main

import (
        "context"
        "dagger/repo-name/internal/dagger"
)

type RepoName struct{}

func (m *RepoName) Echo(stringArg string) string {
        return stringArg
}

// Other example targets
```

---

Dagger info:

```sh
$ dagger version
dagger v0.12.0 (registry.dagger.io/engine) linux/amd64

$ cat dagger.json
{
  "name": "repo-name",
  "sdk": "go",
  "source": "dagger",
  "engineVersion": "v0.12.0"
}
```

Running on WSL2 Ubuntu 22.04 with Docker Desktop.

rajatjindal:
This PR attempts to replace the existing strcase library with a different implementation to fix #7941 

In addition to replacing the library, we are moving this implementation to a local dagger package (currently under `dagql` folder, more on this later), so that:

1. We can customize the behavior as per our requirement
2. If needed, we can easily replace the backing library in future.
3. Add unit test coverage as this transformation is critical for dagger codegen and graph functionality.

Why is this under `dagql` folder? 

I considered a few options e.g. `core`, `internal` or a new pkg under root and `dagql` seemed like the best option among those. Also `dagql` was a common dependency for most of the places where strcase is currently used.

Although, for sdk-runtime, it seems like there is no dependency on dagger's core/dagql package, so right now we are duplicating the code from strcase pkg into those files. But this mean, they can go out of sync quickly. 

how do folks feel about adding a dependency from sdk-runtime onto dagger/dagql package? (one option could be to have a separate repo under dagger, but that seemed like an overkill). 

i am open to suggestions here if folks think we can move the strcase package elsewhere.


sdk/elixir/runtime/go.mod
```diff
@@ -2,40 +2,4 @@ module elixir-sdk
 
 go 1.22
 
-require (
```
rajatjindal, line 5: the go.mod changes were triggered by running `go mod tidy` in the runtime folder. I am trying to identify if there is a way to update the file without triggering these unwanted changes.

sdk/typescript/runtime/main.go
```diff
@@ -9,11 +9,30 @@ import (
...
+	var splitFn = strcase.NewSplitFn(
+		[]rune{'*', '.', ',', '-', '_'},
+		strcase.SplitCase,
+		strcase.SplitAcronym,
+		strcase.PreserveNumberFormatting,
+		strcase.SplitBeforeNumber,
+		strcase.SplitAfterNumber,
+	)
+
+	return strcase.NewCaser(false, nil, splitFn)
```
rajatjindal, line 33: This is done this way to avoid dependency on dagql in runtime. But can we add this to `internal` package and use it from there? what do you think @jedevc @marcosnils?

rajatjindal:
The CI is green (finally). and I think we can review the changes now. There is one question around configuring strcase in runtime (vs dependency on dagql or adding the configuration part to internal pkg).

core/integration/module_test.go
```diff
@@ -88,6 +88,25 @@ func (ModuleSuite) TestGoInit(ctx context.Context, t *testctx.T) {
 		require.Contains(t, generated, "module dagger/my-module")
 	})
 
+	t.Run("issue-7941, single character in kebab case module name", func(ctx context.Context, t *testctx.T) {
```
jedevc, line 91: nit: the issue number can be a comment here, something like `dagger/dagger#7941` - but shouldn't be in the test name.

dagql/strcase/strcase_test.go
jedevc, line 1: Nice :tada: these are *super* valuable, thanks for adding tests :tada:

dagql/strcase/strcase.go
jedevc, line 1: I don't think this should be under `dagql/strcase`, though I get the logic! `dagql` is more our custom graphql server (and other *stuff* like progress visualization, that could probably be moved out :thinking:)

`engine/strcase` should be good I think. We also *probably* want to have it be it's own go module (with it's own `go.mod`/`go.sum`), so that we don't end up pulling in all the deps of the main package into each runtime module if we remove the duplication of the `strcase` initialization and such from there as I suggest above.

There's some precedent for this in `engine/distconsts` you might want to use as inspiration.

sdk/typescript/runtime/main.go
jedevc: > Although, for sdk-runtime, it seems like there is no dependency on dagger's core/dagql package, so right now we are duplicating the code from strcase pkg into those files. But this mean, they can go out of sync quickly.

Yeah, the duplication is definitely not good :cry: I like the idea of adding a dependency onto the root module, that sounds fine to me - we can then have a `replace` directive in `elixir`/`typescript`/`python` to point it back to the top (though *maybe* this might require a bit of a chance to how the runtime gets built).

Another *potential* option would be to add this into the auto-generated `internal` package - but I'm not a huge fan of this - since it's kind of a work-around that only we're able to do - we should try and dogfood the other approach if we can.

jedevc (commented):
Nice :tada: left some comments, but I think the shape is mostly right! I like having our own package in front of this, keeping all this centralized is good (and means it's importable by anyone else, like an external sdk author).

One very sad question - is this a breaking change? Does this change the casing of any args/function names/modules? If so... that's annoying.

We would have to decide to either:
- Just *break* :boom: It's likely these cases were kinda broken already, and maybe they're uncommon enough.
- Add version compatibility :tada: This is trickier, but means that we can have older modules not break, but expose the newer behavior correctly.
	- This would probably work by somehow passing `engineVersion` to `strcase`, and then we would do a semver comparison - versions before v0.12.X get the old behavior, and versions after get the new behavior.
	- This seems like a *total* pain to plumb it all through correctly, propagating `dagger.json`s `engineVersion` all the way through to all the places we call this function is not very fun.

jedevc:
> Having said that, maybe we can write an automation script to list all modules from our daggerverse list, and try calling their functions with new engine. Maybe we can identify some edge cases using that. what do you think

Perfect, yes, that's a good idea - cc @marcosnils @matipan who have a script for this already I think :thinking:

rajatjindal:
> Yeah, the duplication is definitely not good 😢 I like the idea of adding a dependency onto the root module, that sounds fine to me - we can then have a replace directive in elixir/typescript/python to point it back to the top (though maybe this might require a bit of a chance to how the runtime gets built).

> Another potential option would be to add this into the auto-generated internal package - but I'm not a huge fan of this - since it's kind of a work-around that only we're able to do - we should try and dogfood the other approach if we can.

how do you feel about importing `github.com/dagger/dagger/engine/strcase` as a dependency in runtime? given that its a separate module (after the suggestion you made), maybe we could import it. 

(if we do that, we may have to do the runtime change in next PR because right now `go mod tidy` fails to find that new dependency in dagger github repo).

jedevc:
> (if we do that, we may have to do the runtime change in next PR because right now go mod tidy fails to find that new dependency in dagger github repo).

This shouldn't be an issue if we set the `replace` directive in `go.mod`. `sdk/elixir/runtime/go.mod` can have a `replace github.com/dagger/dagger/engine/strcase => ../../../engine/strcase` in them.

dagql/strcase/strcase.go
rajatjindal: 👍, moved it to engine/strcase as suggested. thank you for the suggestion, it makes total sense.

rajatjindal:
> This shouldn't be an issue if we set the replace directive in go.mod. sdk/elixir/runtime/go.mod can have a replace github.com/dagger/dagger/engine/strcase => ../../../engine/strcase in them.

ah ok. i was not sure about that (I don't completely understand the concept of `runtime` in sdk yet). I will make that change 👍

rajatjindal:
> > (if we do that, we may have to do the runtime change in next PR because right now go mod tidy fails to find that new dependency in dagger github repo).
> 
> This shouldn't be an issue if we set the `replace` directive in `go.mod`. `sdk/elixir/runtime/go.mod` can have a `replace github.com/dagger/dagger/engine/strcase => ../../../engine/strcase` in them.

it seems like this may not work? when I run a test with this, i get following error:

```
github.com/dagger/dagger/engine/strcase@v0.0.0-00010101000000-000000000000 (replaced by ../../../engine/strcase): reading /engine/strcase/go.mod: open /engine/strcase/go.mod: no such file or directory
```

which means during the dagger init execution, engine code may not be copied over to the container directly?

jedevc:
The SDK runtimes are the modules that are run by the engine to manage modules in a language - the typescript SDK manages modules written in typescript, generating code for them, running them, etc.

This is where the SDK runtimes get built:

https://github.com/dagger/dagger/blob/ac342afe551bd7ca0d198862029ab30112c606d1/.dagger/build/sdk.go#L78-L100

To make sure that `strcase` is available, we'd need to modify `rootfs` so that it instead it follow the structure of the main repo. So `/sdk/typescript` is going to have `sdk/typescript` at it, and `/engine/strcase` is going to have `engine/strcase` at it. That way, all the paths will all make sense. Additionally, you'll need to set `WithWorkdir`.

I still don't really know if that would fix it, it might take some fiddling around to get it working. You might also need to put a `dagger.json` at the root of the filesystem there to get the module resolution working.

> what do you think about adding dagger/dagger/engine as external dependency for runtime?

We definitely need to make sure to use a `replace` directive still - it can't be external, otherwise, we can't really make any changes without first updating `main` which is really not great.

It should also still be in a sub-package.

---

Regardless. This all feels horribly out of scope.

I'd ignore this issue for now and go back to what you were doing before (without the `go.mod` and everything, etc) - we can just split each of the initialization off to `strcase.go` in each `runtime` directory, adding a comment for where the "original" declaration of this is. And also, make a note at the original declaration, so we know where we need to update if we update it there.

rajatjindal:
> I'd ignore this issue for now and go back to what you were doing before (without the go.mod and everything, etc)

to make sure I understand it correctly, are you suggesting we still keep it under engine, but just don't create a separate go.mod file for it. OR are you suggesting to move it back to dagql subfolder?

jedevc:
I'm suggesting leave it under `engine`, but go back to duplicating the information in `runtime` folders (with comments though).

go.mod
```diff
@@ -168,6 +167,7 @@ require (
 	github.com/dustin/go-humanize v1.0.1 // indirect
 	github.com/emirpasic/gods v1.18.1 // indirect
 	github.com/erikgeiser/coninput v0.0.0-20211004153227-1c3628e74d0f // indirect
+	github.com/ettle/strcase v0.2.1-0.20230114185658-e5db6a6becf3
```
rajatjindal, line 170: Hi @jedevc, one thing I want to bring up is that this version is being pulled from a branch for `ettle/strcase` lib here.  https://github.com/ettle/strcase/pull/13

I know not ideal, but out of the other libraries that i tried, only this combination works as we expected. 

do you think we should fork this library until that PR is merged?

core/integration/module_test.go
```diff
@@ -41,6 +41,57 @@ func TestModule(t *testing.T) {
...
+func (ModuleSuite) TestEngineVersionCompatibilityForModule(ctx context.Context, t *testctx.T) {
+	aEngineVersion := "v0.12.5"
+	bEngineVersion := "dev" //dev for devEngine
+
+	testcases := []string{}
+	err := json.Unmarshal(compatibilityTestModuleListRaw, &testcases)
+	require.NoError(t, err, "reading the module list to do compatibility check")
+
+	for _, module := range testcases {
+		if module != "github.com/sagikazarmark/daggerverse/go@5b826062b6bc1bfbd619aa5d0fba117190c85aba" {
```
rajatjindal, line 78: this is just temporary while I am testing through these changes

core/integration/engine_test.go
```diff
@@ -81,6 +105,18 @@ func engineWithConfig(ctx context.Context, t *testctx.T, cfgFns ...func(context.
...
 }
 
+// engineClientContainerWithVersion returns a container with specific version of dagger cli
+// and connected to the given devEngine service
+func engineClientContainerWithVersion(ctx context.Context, t *testctx.T, c *dagger.Client, devEngine *dagger.Service, version string) (*dagger.Container, error) {
+	endpoint, err := devEngine.Endpoint(ctx, dagger.ServiceEndpointOpts{Port: 1234, Scheme: "tcp"})
+	if err != nil {
+		return nil, err
+	}
+	return c.Container().From(fmt.Sprintf("ghcr.io/dagger/engine:%s", version)).
```
rajatjindal, line 115: version >= v0.12.5 have dagger binary as part of the container image. If we want this test to support comparing version < v0.12.5, we may have to add a function to pull that from GitHub release artifacts.

sdk/typescript/runtime/main.go
rajatjindal: I actually like this idea. this way we may be able to omit the manual copy step. but I am not entirely sure if there are any other downsides to it.

go.mod
rajatjindal: on thinking more about it, I think we should use a fork of this library until the PR is merged.

rajatjindal:
the engine compatibility test has been moved to being a separate module in #8197

core/schema/module.go
```diff
@@ -1174,3 +1176,7 @@ func (s *moduleSchema) writeDaggerConfig(
 
 	return nil
 }
+
+func (s *moduleSchema) addCompatToCtx(ctx context.Context) context.Context {
+	return compat.AddCompatToContext(ctx, s.dag.View)
+}
```
jedevc, line 1182: nit: dead code.

core/integration/module_test.go
```diff
@@ -41,6 +41,110 @@ func TestModule(t *testing.T) {
...
+			out, err := versionAModGen.
+				With(daggerQuery(`{minimal{withSecondFunction(skipTparse:"hello"){stdout}}}`)).
+				Stdout(ctx)
+			require.NoError(t, err)
+			require.JSONEq(t, `{"minimal":{"withSecondFunction":{"stdout":"hello\n"}}}`, out)
+		}
+	}
+}
+
+func (ModuleSuite) TestModuleNamingCompatFuncName(ctx context.Context, t *testctx.T) {
```
jedevc, line 96: Hm, I suspect there's still something missing here :cry: 

Suppose:
- We have a dep using "new dagger" that was one of these weird cases like `WithDaggerCLIAlpine`
- We have a main module that's using an "old dagger" that install `WithDaggerCLIAlpine`
- What does the codegen look like for the main module? I think it probably needs to generate with the "old dagger" form.

Maybe this "just works" as is, but I'd like a test to confirm that (and ideally for go/python/typescript consumers).

If this is an issue, then the fix will be to propagate the `strcase.Caser` code into `cmd/codegen` and each of the runtime modules. Thankfully we already propagate the `engineVersion` into there, so *hopefully* wouldn't be too bad with all the context plumbing you've already done :tada:

core/module.go
```diff
@@ -122,11 +123,25 @@ func (mod *Module) IDModule() *call.Module {
 	return call.NewModule(mod.InstanceID, mod.Name(), ref)
 }
 
+func (mod *Module) addCompatToCtx(ctx context.Context) (context.Context, error) {
```
jedevc, line 126: I *think* this isn't needed? By calling this in the server, doesn't this propagate down everywhere to where it needs to go?

Might be just leftover, but can't quite tell.

core/typedef.go
```diff
@@ -64,6 +66,10 @@ func (fn Function) Clone() *Function {
 	if fn.ReturnType != nil {
 		cp.ReturnType = fn.ReturnType.Clone()
 	}
+	//this is ugly, but lets try it
+	if fn.ctx != nil {
+		cp.ctx = fn.ctx
+	}
```
jedevc, line 72: nit: we can remove this now, `WithArg` should take a `context` arg, as with all the other typedef methods.

core/compat/go.mod
jedevc, line 1: I don't think needs to be a separate module. `strcase` doesn't either now, since we're going with the copy-paste approach for now.

jedevc (commented):
Some notes, I *wonder* if we need to do this same thing in the SDKs as well, but I suggested writing a few tests to see if we need to - we might not.

Quick note - I *really* like the new `core/compat` package, I think this is a good approach to move forward with. I like the idea of being to attach more things in the future :smile:

TomChv:
What's the status of this PR? I see a lot of conflict with main, I think this one should be rebase to have a change to be merged :D

rajatjindal:
apologies for losing track of this PR updates here. but I do plan to pick this one up again sometime soon.

Issue #7836, opened 2024-07-06
CLI: return object on sync


helderco:
This supersedes:
- https://github.com/dagger/dagger/issues/6517

IDs are hidden from the CLI for a reason (see https://github.com/dagger/dagger/issues/6710). However you can get the ID by calling `sync`. A better alternative is to **follow the SDKs on how they handle `sync` by returning the `Container`** corresponding to the resulting `ContainerID` using `loadContainerFromID()`.

After https://github.com/dagger/dagger/issues/6515 it doesn’t seem necessary since a chain ending in a `Container` calls `stdout` and thus forces evaluation. The problem is that it also forces execution, and that may be undesirable:
- https://github.com/dagger/dagger/issues/7821

This could be a solution to that by making sure we don’t pre-select any fields in `Container` that trigger execution, but reverting to implicitly calling `sync` again, before selecting the fields for the image data.

In any case, making `sync` chainable like in SDKs is more useful than returning an ID that you can’t use. Users could then add `sync` to the middle of a command chain, and everything before it would be evaluated before even parsing the rest of the command.

PR #8127, opened 2024-08-08
feat: automatically generate sync methods for user objects

jedevc:
We've chatted about something like this before - the idea is that we should automagically generate `Sync` methods for user defined types.

Today, users can kind of fake this by calling `id` on an object. However, this is really hacky, and additionally with the kind of changes suggested in https://github.com/dagger/dagger/pull/7761 this would no longer be possible.

Instead, we should have a real way of forcing evaluation for a specific user object, by automatically generating `Sync` methods for all user objects. Additionally, `Sync` should also ensure that all other `Sync`able fields of a user object are also `Sync`ed - I think this is mostly expected?

### DIFF
--- a/core/schema/coremod.go
+++ b/core/schema/coremod.go
@@ -143,6 +143,9 @@ func (m *CoreMod) TypeDefs(ctx context.Context) ([]*core.TypeDef, error) {
 					isIdable = true
 					continue
 				}
+				if introspectionField.Name == "sync" {
+					continue
+				}
 
 				fn := &core.Function{
 					Name:        introspectionField.Name,
--- a/core/container.go
+++ b/core/container.go
@@ -1035,7 +1035,13 @@ func (container *Container) WithGPU(ctx context.Context, gpuOpts ContainerGPUOpt
 	return container, nil
 }
 
-func (container Container) Evaluate(ctx context.Context) (*buildkit.Result, error) {
+func (container Container) Evaluate(ctx context.Context) error {
+	_, err := container.evaluate(ctx)
+	return err
+}
+
+//noline:unparam
+func (container Container) evaluate(ctx context.Context) (*buildkit.Result, error) {
 	if container.FS == nil {
 		return nil, nil
 	}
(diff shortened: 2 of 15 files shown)


core/schema/coremod.go
```diff
@@ -142,6 +142,9 @@ func (m *CoreMod) TypeDefs(ctx context.Context) ([]*core.TypeDef, error) {
 					isIdable = true
 					continue
 				}
+				if introspectionField.Name == "sync" {
+					continue
+				}
```
jedevc, line 147: I was thinking about this - but I don't think that the sync method should appear in the `TypeDef`s. `ID` doesn't, and I think with this change, `Sync` becomes more standardized, and so is more like `ID`.

Removing it from here means the CLI doesn't see `sync` as a field property... which kind of actually makes sense to me - but this would break users who are calling `sync` on the CLI. Since we now don't default to `sync`ing... could this be an issue?

@helderco curious about your thoughts here. I'm almost thinking that if we want to do explicit syncing in the CLI, we should "raise" the syntax up, and have a special flag to do so (like `-o` today), instead of requiring `sync` to be somewhere around. This would also then allow us to implement entirely custom behavior around https://github.com/dagger/dagger/issues/7836 (and maybe do some other interesting things as well).

helderco: I believe this is relevant:
- https://github.com/dagger/dagger/issues/7836

jedevc: As discussed in discord, we want to keep `Sync` in the typedefs for now. For consistency, we should also add it in the typedefs for user modules as well.

However, we could also potentially have `Syncable` as a property of `ObjectTypeDef`, this might also make sense.

jedevc: Argh, so I came back to this, and it's *so* weird to have a type def for user objects for this when it's not been explicitly declared by the sdk - so weird, the codebase is not structured at all to allow it, and the thought of changing it to allow it doesn't appeal really. I think maybe `sync` needs to be supported in `dagql` directly.

I'm also really reconsidering whether it should evaluate all items on the object. That also doesn't feel right to me, and additionally so because we can't do it for private fields.

Issue #8030, opened 2024-07-24
Module Self Invocation


aluzzardi:
Currently, a module invokes itself through the native language (e.g. self.otherFunction), which completely bypasses dagger.

Instead, dagger could provide a proper mechanism for a module to invoke one of its own functions via the engine in order to get:

- **Caching**: Calling your own function should get the same caching treatment as invoking an external function
- **Observability**: Invoking your own function should create a new span, visible from TUI/Cloud

This could potentially also be useful for testing, TBD.

/cc @sipsma @vito @helderco

aluzzardi:
Thinking about the implementation, I think this would be a pretty big change for module authors. I guess that rather than calling their code directly, they would do something like:

- `dag.Self`?
- Or maybe a special extra argument (`self`, `this`) -- but this will feel "alien" to the native language
- Or through some sorcery in codegen (language specific) to inject some kind of wrappers?

Also, there's the question of state. If the user sets a variable in their own struct/object, is it going to be available to the invoked function?

jedevc:
I'm quite interested in this as well :tada:

One idea I really liked is that we could integrate this through the module dependencies - you can install the module as a dependency as itself. So you could do `dagger install .`.

One reason this is neat, is because then you can name your module of yourself as you like - `Self`, or `This`, or even `FullNameOfModule`. We could pick a semi-reasonable default, but then users can edit their `dagger.json` config.

R.e. state - I think without a huge rework of the call mechanism, it'll have entirely different state, since it'll be in a different `ExecOp`. I *think* that's what we want - it should just be like any other call.

kpenfound:
This is super impactful for building agents! Primarily because we want to be able to pass types from our current module to `dag.Env`.

Talking with @TomChv live, it sounded like there could be some complexity with syncing module state between the current module and handling it somehow through `dag.Self`, however, I do think a simpler implementation where we just get the current module's types generated in the API would be enough to unlock a ton of potential.

PR #8128, opened 2024-08-08
WIP: fix using interface in constructor

khrisrichardson:
As reported [here](https://github.com/dagger/dagger/pull/8115#issuecomment-2276115241), in addition to there being an issue with enums in constructors, there's similarly a couple issues with interfaces in constructors.

* installing module with an interface in the constructor results in following error

```
Stderr:
Error: failed to generate code: input: moduleSource.withContextDirectory.withDependencies.asModule resolve:
failed to create module: select: failed to update codegen and runtime: failed to generate code: failed to get
schema introspection json during module sdk codegen: failed to get schema for module "mallard": failed to
install constructor: failed to create function: failed to find mod type for function "" arg "d" type
```

* developing module with an interface embedded in the struct results in the following errors

```
Stderr:
Error: failed to get module SDK: input: moduleSource.withContextDirectory.asModule resolve: failed to create
module: select: failed to update codegen and runtime: failed to generate code: failed to get modified source
directory for go module sdk codegen: select: process "codegen --output /src --module-context-path /src
--module-name mallard --introspection-json-path /schema.json" did not complete successfully: exit code: 1

Error: generate code: template: module.go.tmpl:79:3: executing "_dagger.gen.go/module.go.tmpl" at
<ModuleMainSrc>: error calling ModuleMainSrc: failed to parse method XXX_GraphQLType: failed to find decl
for method XXX_GraphQLType: no decl for XXX_GraphQLType
```

When the enum bug is addressed, I can test whether it also fixes the interface issue or issues. Chances are that it might fix the module install issue but not the struct embedding issue.

### DIFF
--- a/core/integration/module_iface_test.go
+++ b/core/integration/module_iface_test.go
@@ -128,3 +128,106 @@ func (m *Test) GetDuck() Duck {
 	require.NoError(t, err)
 	require.Equal(t, "mallard quack", strings.TrimSpace(out))
 }
+
+func (ModuleSuite) TestIfaceGoConstructor(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	out, err := c.Container().From(golangImage).
+		WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+		WithWorkdir("/work/mallard").
+		With(daggerExec("init", "--source=.", "--sdk=go")).
+		WithNewFile("main.go", `package main
+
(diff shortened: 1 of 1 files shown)


khrisrichardson:
I will revisit what I was attempting to do here when https://github.com/dagger/dagger/issues/8030 lands. The issue that spurred this investigation is fixed, but the usage of interfaces is still somewhat hampered.

PR #8129, opened 2024-08-09
docs: skip generating spectaql generation in netlify deploy

jedevc:
We now do this as part of the `docs generate` dagger pipeline - since it's already there, we might as well just use it - see #7937 and #8118. 
    
Additionally, we do some shenanigans there so that we generate stable results, that don't have random results inserted in there - this requires a custom build of spectaql.

I also switched the `docs` directory to use yarn instead of npm. We were already using this in our netlify deploy step, but:
1. Because we only had a `package-lock.json` file, and not a `yarn.lock` file, none of our dependencies were actually getting pinned!
2. This differed from the recommended dev setup.

I've just switched over from npm to yarn, hopefully that's alright!


docs/ARCHIVING.md
```diff
@@ -60,7 +60,7 @@ NOTE: At the time of writing, this is a completely manual process. This is becau
 
 - In `docs/` sub-directory:
   - Replace `/img` paths with `/0.1/img` paths
-- Run `npm run build:withoutAuth` and store the `build/` directory as `site/0.1`
+- Run `yarn run build:withoutAuth` and store the `build/` directory as `site/0.1`
```
vikram-dagger, line 63: @jedevc I don't think the changes in this file are correct. These archives are built from specific past releases, presumably `yarn` will not exist at those time points?

vikram-dagger (changes requested):

vikram-dagger:
I would like @crjm to comment if he sees any issue with this specific change:

> I also switched the docs directory to use yarn instead of npm. We were already using this in our netlify deploy step...

Apart from the above and one requested change, this seems fine to me.

docs/ARCHIVING.md
jedevc: Woops, fixed :tada:

vikram-dagger (approved):

PR #8132, opened 2024-08-09
Fix `failed to collect IDs from parent fields`

jedevc:
When returning an object from a module function that contains a nil field, we can get the following error:

```
Error: response from query: input: test.hello resolve: failed to collect IDs from parent fields: failed to collect IDs for field "<name>": expected *ModuleObject, got <nil>
```

cc @sipsma


sipsma:
Thanks for adding the integ test @jedevc! I pushed a separate commit with the fix, just needs a nil check

sipsma (approved):

jedevc:
Lgtm 🎉

PR #8135, opened 2024-08-09
[DISCUSSION]: move `dagger core` to `dagger call -c`

helderco:
This changes the way you call core functions in the CLI (currently `dagger core <function>`):

- `dagger call -c <function>` → Core function
- `dagger call -m <module> <function>` → Module function

## Why?

TL;DR: Consistency.

This relates to:
- https://github.com/dagger/dagger/issues/6947

In that issue there was a lot of discussion on the DX to call core functions from the CLI. In the end, we were left with three main options:

- `dagger core <function>` → Merged in https://github.com/dagger/dagger/pull/7310
- `dagger call -c <function>` → This PR
- `dagger <function>` → There's interest in it, but not for now

The first option had a few advantages:
- Slightly shorter to type
- Stepping stone to the third option (future possibility)
- Was ready to merge (after going through a refactor to support multiple top-level commands that behave like `dagger call`)

However, it has the disadvantage that it can feel [inconsistent](https://github.com/dagger/dagger/issues/6947#issuecomment-2264107770) with `dagger call`. As I mentioned in https://github.com/dagger/dagger/pull/7310#issuecomment-2276640860, it’s very easy to preserve the refactor that simplified the code, while changing back to using a flag (`--core`).

So we still get the stepping stone (implementation wise) to `dagger container`, thus **the only advantage at this time of keeping `dagger core` is being slightly shorter to type, at the cost of feeling inconsistent with `dagger call`**.

## Next

Since I’m going away on PTO, I leave this alternative implementation **so the team has a little more time to consider the DX before the next release**, without blocking the feature (which is already merged).

Feel free to close or merge this change.

\cc @samalba @shykes @aluzzardi @jedevc

### DIFF
--- a/cmd/dagger/call.go
+++ b/cmd/dagger/call.go
@@ -12,21 +12,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var funcCmds = []*FuncCommand{
-	callModCmd,
-	callCoreCmd,
-}
-
-var callCoreCmd = &FuncCommand{
-	Name:              "core [options]",
-	Short:             "Call a core function",
-	DisableModuleLoad: true,
-	Annotations: map[string]string{
-		"experimental": "true",
-	},
-}
-
-var callModCmd = &FuncCommand{
+var callCmd = &FuncCommand{
 	Name:  "call [options]",
 	Short: "Call one or more functions, interconnected into a pipeline",
 }
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -157,6 +157,9 @@ func (fc *FuncCommand) Command() *cobra.Command {
 				}
 				c.FParseErrWhitelist.UnknownFlags = false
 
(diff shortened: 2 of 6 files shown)


jedevc:
Personally, I prefer this option over `dagger core`, so has my LGTM :+1:

FYI, given we're aiming to release early this week, it would good to try and get some consensus as soon as feasible :tada: Ofc, if there's more things to debate/discuss, we can delay the release.

gerhard:
From a consistency perspective, I also like the idea of `dagger call -c` when it comes to calling core functions.

With the upcoming `0.13.0` release, I think that we should either:
- Close this
- Merge this and aim to release in `0.13.0`

Any thoughts @dagger/maintainers ?

shykes:
we already had a whole debate and got to (weak) consensus on dagger core. Let's give that a chance and close this.

PR #8143, opened 2024-08-12
chore(php): update php module example to match others

jedevc:
The other SDKs are all consistent in their naming of their autogenerated example methods - PHP should be the same.


carnage (commented):
All good. (I can't give an explicit approval)

charjr (approved):
Looks fine, thank you for amending the descriptions as well :+1:

Issue #7733, opened 2024-06-24
Allow easy healthcheck for the dagger engine


jedevc:
Spun out from this discord discussion: https://discord.com/channels/707636530424053791/1253636253358755840/1253636253358755840

It's currently quite difficult to write a healthcheck for the dagger engine itself - how do you know if it's running, and ready to accept connections?

Previously, you used to be able to use the `buildctl` command to connect manually (but that was removed), or even use a dummy `dagger query` command (but that was also removed).

We should make it easy to run a command in the engine container to see if the engine is running and healthy. Ideally, this would mean shipping `dagger` inside the engine container, and maybe even providing a simple `dagger ping` or `dagger health` command to check communication (we could use this in our own CI as well).

jedevc:
Oops. That's not right :scream: 

https://github.com/dagger/dagger/blob/8aac1a62a204529ded0d9aebafdcdcb5df2397be/ci/build/builder.go#L257-L259

`buildctl` needs to exist, but it's a symlink to `dial-stdio` - this was changed *ages* ago, in https://github.com/dagger/dagger/pull/6100#issuecomment-1809271407.

matipan:
Good to know!

What does `dial-stdio` right now do for us in the context of health check?

jedevc:
Uh, it hangs until `stdin` is closed :thinking:

shykes:
I think the solution is:

1) Bundle the CLI and the engine together into a single image (#6887)
2) Add a simple `ping() String!` function in the dagger api

Then the healtcheck can simply be `dagger query << '{ping}'` or, in the future, `dagger core ping`

sipsma:
I'm planning on adding a whole `engine` API under query in our graphql API (to support cache query and control as part of https://github.com/dagger/dagger/pull/7646), at which point we can move `version` and/or add `ping` under there. So then this would be `dagger core engine version`/`dagger core engine ping`.

> Bundle the CLI and the engine together into a single image

Still agree with doing this generally speaking, but I don't think it would be a pre-req for implementing this functionality. At least that I can see it's orthogonal.

shykes:
> I'm planning on adding a whole `engine` API under query in our graphql API (to support cache query and control as part of #7646), at which point we can move `version` and/or add `ping` under there. So then this would be `dagger core engine version`/`dagger core engine ping`.

Isn't the entire API the engine API already? What's the rule for what goes under `engine` and what doesn't?

> > Bundle the CLI and the engine together into a single image

[...]

Issue #6887, opened 2024-03-15
✨ Official Dockage image for the CLI?


Scalahansolo:
### What are you trying to do?

The dagger docs for gitlab / github / etc have each job installing curl followed by installing the dagger CLI to use for a given CI run. It would be much nicer DX, and probably faster to just have an official docker image that can be used that has the CLI already installed?

gerhard:
Agreed! WDYT @jedevc @sipsma @shykes?

jedevc:
No objections, I think this could definitely be an improvement for running dagger functions in k8s/etc :tada: Previously, in the standalone SDK model, I don't think this was as important, since you could just use a docker runtime for your SDK language, but with the module model, we only need a single CLI entrypoint, and the engine provisions the language component.

I'd suggest we put this in `registry.dagger.io/cli` to contrast with `registry.dagger.io/engine` - that was it's clear which one is which, and users don't end up with the wrong one! 

Ideally, this can be a very lightweight image: it just needs the dagger CLI, but may also need to include `git`/etc and any other referenced binaries.

The steps to do this would be:
- Add a new mage target to build/publish a CLI image (or wait for #6843)
- Add some basic provisioning tests to check that this works as expected!

shykes:
1. I agree with a CLI image.
2. We should briefly talk about the relationship between the `cli` and `engine` images. I think in the future they will be more tightly coupled: one will be "CLI only" and the other will be "CLI + buildkit runtime". This is a topic for production readiness.
3. We should talk about choice of distro. Right now we have Alpine for one version of the engine image, and Ubuntu for the other. Sooner or later we need to standardize. This issue is a natural forcing function to discuss that standardization.

@dlorenc suggested that Wolfi might be a good candidate for this, thoughts?

jedevc:
@shykes could you elaborate a bit on point 2? I think the engine image probably won't have the CLI embedded, it'll just have the engine as we build it today - I don't necessarily see that changing dramatically, but I guess this also touches on #5583 (like everything else I guess lol).

For 3, for the CLI, I actually think we might be able to do something like distroless? I think we probably just need a little collection of static binaries. For the engine, I would like to unify the GPU/standard builds to use a common base - ideally, I'd also like to avoid ubuntu, due to size constraints. If we could get alpine/wolfi to install all the right CUDA stuff (agh, it would be nice to have some tests for this in CI) - then yup, we should go all in on that :tada:

dlorenc:
[...]

PR #8147, opened 2024-08-13
feat: Bundle CLI in the Engine image (a.k.a. One True Artefact)

gerhard:
So that the Engine image already ships with the matching CLI, all configured and ready to go.
- It addresses the primary issues raised in https://github.com/dagger/dagger/issues/6887

This change adds an extra `27MB` to the final container image:
- `466MB` before this change
- `493MB` after this change (`~6%` larger)

## FOLLOW-UPs
- Use a `dagger query` to test the healthiness of the Engine in the Helm chart
  - What would make a good default query?
  - Tie this to https://github.com/dagger/dagger/issues/7733
- **Maybe** publish a CLI image with git preinstalled **DISCUSS**
- Close the initial issue

### DIFF
--- a/.changes/unreleased/Added-20240813-161514.yaml
+++ b/.changes/unreleased/Added-20240813-161514.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: Bundle CLI in the Engine image so that both parts of Dagger (CLI+Engine) ship as a single artefact
+time: 2024-08-13T16:15:14.821633+01:00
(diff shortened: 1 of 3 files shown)


jedevc (approved):

gerhard:
Thanks @jedevc 💪

Will merge as soon as the last check passes.

shykes:
Thank you for getting this done @gerhard !

PR #8148, opened 2024-08-13
fix: allow optional enum arguments in user functions

jedevc:
This was broken from the beginning when user-defined enums were introduced in #7498. `ModTypeFor` was refactored, and essentially always returned early before the handling of `typeDef.Optional`. This was incorrect, and meant that we ended up with `DynamicOptional`s during enum conversion - see https://github.com/dagger/dagger/pull/7498/files#diff-5eb7740241ddbf3efbef98d43f59b2081a4b6f7bfd00a3daba69d36f107d53bbL262-R287.

To resolve this, we can only return early when we've loaded the mod type directly from the deps - when we've executed the load ourselves, we need to complete the rest of the function.

### DIFF
--- a/core/enum.go
+++ b/core/enum.go
@@ -59,22 +59,11 @@ func (m *ModuleEnumType) ConvertToSDKInput(ctx context.Context, value dagql.Type
 	if value == nil {
 		return nil, nil
 	}
-
-	var val string
-	switch x := value.(type) {
-	case *ModuleEnum:
-		val = x.Value
-	case dagql.Scalar[dagql.String]:
-		val = string(x.Value)
-	default:
-		return nil, fmt.Errorf("%T.ConvertToSDKInput cannot handle type %T", m, x)
-	}
-
 	decoder, err := m.getDecoder(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("%T.ConvertToSDKInput: failed to get decoder: %w", m, err)
 	}
-	return decoder.DecodeInput(val)
+	return decoder.DecodeInput(value)
 }
 
 func (m *ModuleEnumType) CollectCoreIDs(ctx context.Context, value dagql.Typed, ids map[digest.Digest]*call.ID) error {
@@ -163,6 +152,8 @@ func (e *ModuleEnum) DecodeInput(val any) (dagql.Input, error) {
 		return e.Lookup(x)
 	case dagql.Scalar[dagql.String]:
 		return e.Lookup(string(x.Value))
+	case *ModuleEnum:
+		return e.Lookup(x.Value)
 	default:
 		return nil, fmt.Errorf("cannot create dynamic Enum from %T", x)
 	}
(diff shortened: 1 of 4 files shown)


core/module.go
```diff
@@ -282,39 +282,34 @@ func (mod *Module) ModTypeFor(ctx context.Context, typeDef *TypeDef, checkDirect
...
+		slog.ExtraDebug("module did not find scalar", "mod", mod.Name(), "scalar", typeDef.AsScalar.Value.Name)
 	case TypeDefKindEnum:
-		modType, ok, err = mod.modTypeForEnum(ctx, typeDef, checkDirectDeps)
-		if checkDirectDeps && ok {
-			return modType, true, nil
+		modType, ok, err = mod.modTypeFromDeps(ctx, typeDef, checkDirectDeps)
+		if ok || err != nil {
+			return modType, ok, err
 		}
+		modType, ok = mod.modTypeForEnum(typeDef)
```
vito, line 312: Nothingburger, preserved for posterity:

Does this mean we'll find a dependency's definition before our own?

Actually, it looks like we did that before, too.

Is that weird? 🤔 Maybe a pre-existing bug that just hasn't come up yet? Or maybe it just doesn't matter, because TypeDefs are namespaced. That must be it. There's no "shadowing", your TypeDef is `Foo` and their TypeDef is `SomeModule_Foo`, so we'd never find it in both.

I guess we could even do something silly like "if it has a `_`, only look in deps, and if it doesn't, don't look in deps."

core/module.go
```diff
@@ -397,46 +392,18 @@ func (mod *Module) modTypeForInterface(_ context.Context, typeDef *TypeDef, _ bo
...
-		// check to see if this is from a *direct* dependency
-		depType, ok, err := mod.Deps.ModTypeFor(ctx, typeDef)
-		if err != nil {
-			return nil, false, fmt.Errorf("failed to get scalar type from dependency: %w", err)
-		}
-		if ok {
-			return depType, true, nil
-		}
-	}
-
```
vito, line 411: lol, this was repeated 4x - worth the refactor

core/enum.go
```diff
@@ -163,6 +152,8 @@ func (e *ModuleEnum) DecodeInput(val any) (dagql.Input, error) {
 		return e.Lookup(x)
 	case dagql.Scalar[dagql.String]:
 		return e.Lookup(string(x.Value))
+	case *ModuleEnum:
+		return e.Lookup(x.Value)
```
vito, line 156: This is only slightly odd because no other scalar types support decoding from their own type, but I'm not gonna pretend it's wrong. IME there's not much downside to just letting these conversion functions grow organically in a system where we have to anticipate multiple representations of values. Definitely worth cleaning up that `switch` statement.

(Or: How I Learned to Stop Worrying and Love the Switch)

vito (approved):
Nice fix + refactors!

core/module.go
jedevc: > ? Or maybe it just doesn't matter, because TypeDefs are namespaced. That must be it.

Yeah, this is it - all items are namespaced, including the "self" module. I actually expect that if we had a name clash we'd have more pressing issues other than "dependencies get resolved first", I think there'd be some fun graphql funkiness in there, we expect unique names all through there.

PR #8149, opened 2024-08-14
fix: various cross-module secret sharing issues

jedevc:
Thanks @sagikazarmark for the original bug report :tada: (I *think* this is the same issue? it does have a slightly different error message).

```
invoke: input: dep.secretMount.mount resolve: failed to add client resources from ID: failed to add secret: secret xxh3:bc7dcced26cc08c7 not found in other store
```

The issue appears to be that the `SecretMount.Secret` is marked as `// +private` - if this is removed, the test appears to pass with no issues. I'm *guessing* that what's happening is that because the field is marked as private, it's not appearing as a field in the typedef.

Then, when we iterate to `CollectCoreIDs`, we don't manage to see this field:

https://github.com/dagger/dagger/blob/3c52de5cd5daf202f57d4879f914dac3d2bc3ccf/core/object.go#L103

I think that makes rough sense - note `PBDefinitions` has this same issue - potentially this could cause bugs in `linkDependencyBlobs`? Maybe, but I'm not sure what the symptoms of that would even look like.

There are a couple things that I think probably need fixing:
- Typedefs should be extended to have private fields - the engine should know about these private fields, SDKs shouldn't be hiding them. I think this is the only way we can reasonable fix the `linkDependencyBlobs` thing from above as well.
- It seems odd that the returned container with the mounted secret isn't detected as a secret? Is there an issue in our ID walking code? I think the secret ID *should* appear in the returned container right?

cc @sipsma

### DIFF
--- a/.changes/unreleased/Fixed-20240911-105222.yaml
+++ b/.changes/unreleased/Fixed-20240911-105222.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Fix setting secrets on module object in constructor
+time: 2024-09-11T10:52:22.201566784+01:00
+custom:
+  Author: sipsma
+  PR: "8149"
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -2756,6 +2756,48 @@ func (t *Test) FnB(ctx context.Context) (string, error) {
 		require.Equal(t, "omg\n", string(decoded))
 	})
 
+	t.Run("parent field set in constructor", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+		ctr := c.Container().From(golangImage).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c))
+
+		ctr = ctr.
+			WithWorkdir("/work").
+			With(daggerExec("init", "--name=test", "--sdk=go", "--source=.")).
+			WithNewFile("main.go", `package main
+
+import (
+	"context"
+	"dagger/test/internal/dagger"
+)
+
+type Test struct {
+	Ctr *dagger.Container
+}
+
+func New() *Test {
+	t := &Test{}
+	secret := dag.SetSecret("FOO", "omfg")
+	t.Ctr = dag.Container().From("`+alpineImage+`").
+		WithSecretVariable("SECRET", secret)
+	return t
+}
+
+func (t *Test) GetEncoded(ctx context.Context) (string, error) {
+	return t.Ctr.
+		WithExec([]string{"sh", "-c", "echo $SECRET | base64"}).
+		Stdout(ctx)
+}
+`,
+			)
+
+		encodedOut, err := ctr.With(daggerCall("get-encoded")).Stdout(ctx)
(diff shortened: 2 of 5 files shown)


sagikazarmark:
I may have found another repro: https://github.com/sagikazarmark/daggerverse/pull/155/files#diff-2598c9fcbe43a9002d7558bf5ca98f0336b6eddbc7da5f699e0f59780e35c8fdR74-R82

When getting secrets from a different module through an interface, it fails with the following error:

```
invoke: input: svix.service resolve: failed to add client resources from ID: failed to add secret: secret xxh3:2396ae2da06a92ab not found in other store
```

Reproduce: dagger -m github.com/sagikazarmark/daggerverse/svix/examples/go@pull/155/merge call svix-postgres

Unfortunately, I can't verify if it's working on 0.12.0, because interfaces in a constructor were broken on that version. But I haven't seen this issue (secret not found) on 0.12.0.

sipsma:
(Sorry about the long radio-silence, unexpected IRL events followed by other fires to put out once back)

The problem around `// +private` makes perfect sense and agree with the fix being to actually track those in typedefs (just don't put in the external schema).

For the repro here: https://github.com/dagger/dagger/pull/8149#issuecomment-2302181149 I *think* it may be due to the same thing rather than a separate interface bug? Specifically because of this line: https://github.com/sagikazarmark/daggerverse/pull/155/files#diff-2598c9fcbe43a9002d7558bf5ca98f0336b6eddbc7da5f699e0f59780e35c8fdR17-R18
* That would align with the error specifically occurring at `svix.service`

sagikazarmark:
Thanks @sipsma! I won't pretend to understand the problem or the solution, but I look forward to the latter. Currently, I'm stuck on 0.12.0 where this problem doesn't occur.

I've tried to work around the problem, but since some of the private fields are types from other modules (that I can't make non-private), I can't.

sagikazarmark:
Hm, it seems to me there are multiple errors regarding the svix module:

```
Stdout:
invoke: input: svix resolve: call constructor: process "/runtime" did not complete successfully: exit code: 2

Stdout:
invoke: input: secret.plaintext resolve: secret not found: xxh3:37aa0d668210df47
input: svix.service resolve: call function "Service": process "/runtime" did not complete successfully: exit code: 2
```

Removing `// +private` from the container didn't solve the problem. And the first error indicates there is something wrong inside the constructor. The reason behind could still be a `// +private` field in the downstream module.

jedevc:
```
module.go.tmpl:85:3: executing "_dagger.gen.go/module.go.tmpl" at <ModuleMainSrc>: error calling ModuleMainSrc: failed to parse field type: unsupported type for named type reference *types.Map
```

Previously, users could put *any* type in a `private` field. This error occurs because `map[string]any` is in a `private` field, and we don't support map types. I guess we need to support `map` types? Or at the very least, allow defining this as an `Undefined` type in the typedefs API.

jedevc:
LGTM with just @sipsma's commits in here, but can't approve and merge since I opened the original PR.

PR #8150, opened 2024-08-14
chore: bump go to 1.23

jedevc:

PR #6637, opened 2024-02-09
Upgrade to go 1.22

jedevc:
See https://go.dev/doc/go1.22. Worth reading, but in particular:

> - Previously, the variables declared by a "for" loop were created once and updated by each iteration. In Go 1.22, each iteration of the loop creates new variables, to avoid accidental sharing bugs. The [transition support tooling](https://go.dev/wiki/LoopvarExperiment#my-test-fails-with-the-change-how-can-i-debug-it) described in the proposal continues to work in the same way it did in Go 1.21.
> - "For" loops may now range over integers.
>   ...
>   See the spec for [details](https://go.dev/ref/spec#For_range).

Also need to bump to 1.56 of golangci-lint for these changes.

PR #8151, opened 2024-08-14
chore: use features introduced recently in go 1.22

jedevc:
We've been using go 1.22 for a while now: #6637. However, this introduced some really interesting features, and we've not really converted any code to use it.

This PR starts using the two main changes in the language from https://go.dev/doc/go1.22:
- Use `range <int>` syntax wherever possible
- Avoid explicit loop capture, this is now the default

Hopefully, after #8150 lands, we can also start using the new func iterators :tada:


gerhard (approved):
👍 🚀

Issue #6355, opened 2024-01-03
`.Export` partially transfers files and then gets stuck


jedevc:
This seems to be the same issue as upstream in buildkit: https://github.com/moby/buildkit/issues/2950 (cc @sipsma)

We've seen from a customer, as well as reported on discord: https://discord.com/channels/707636530424053791/1191805052378161192/1191805052378161192:

> I've not seen the issue happening without a large amount of files being exported, so this might be related indeed

We should try and commit some debugging resources to trying to track down the deadlock upstream, so that we can fix it once and for all :tada:

jedevc:
The suggested workaround for users currently hitting this is to try and reduce the number of files being exported if possible - you can remove files from the target directory / container until the export reliably succeeds.

johnstonematt:
I have had the same issue, trying to export a large directory containing many files. To hack around, I zipped up the directory to a `.zip` file and then exported the single file (instead of the directory), then unzipped on the host machine, and it worked.

johnstonematt:
I made a script to recreate the issue and explore it further, you can see it [here](https://gist.github.com/johnstonematt/33a6d28ceb0528d0688291fede44a8ff).

It's clearly related to not just file size, but file count, as you can successfully export a 10GB directory with only 5 files in it, but fail to export a 750MB directory with 60k files in it.

PR #8155, opened 2024-08-14
workaround hanging `TestTypeScriptPackageManagerDetection` test

vito:
This test creates a single client and uses it for 4 tests in parallel. Each test runs `dagger develop` at some point, and we're seeing this part hang in the export step:

image

Giving each test its own seems to work around the issue (0/2 runs hang with the fix, 2/2 hang without).

It doesn't seem to be a regression related to the test or the feature being tested, and it's better for each test to open its own client anyway so we don't leave one hanging around until they run.

### Deadlock analysis

tl;dr: it appears we've found a repro for #6355 🎉 - so after merging this PR we can just revert locally to investigate it.

Here are some runs that demonstrate the hang:

* https://dagger.cloud/dagger/traces/03b9d7e773d03d5323334f5d3da2177b?span=35610e5c887dc49e
* https://dagger.cloud/dagger/traces/10f31bb9e437e889577e7b61939a3df5?span=d01eb9cf70822eae
* https://dagger.cloud/dagger/traces/419926ad0eb9749dd0c0c354db14f204?span=dc5a8b811ba5a8b9
* https://dagger.cloud/dagger/traces/5629f35bfe59bc3f4489dc9ca190e3e4?span=0b8319c0a87c416c
* https://dagger.cloud/dagger/traces/ee425d99d4354819ee699656727365ad?span=69ff424763e0e8a8

For the last one I managed to capture a stack dump of the engine, with the help of @gerhard and @matipan. 🙏

Which is blocked on these calls:

image

=>

image

=>

image

=> this is holding a lock which these other goroutines want real bad:

image

### DIFF
--- a/.dagger/test.go
+++ b/.dagger/test.go
@@ -150,7 +150,8 @@ func (t *Test) testCmd(ctx context.Context) (*dagger.Container, error) {
 		WithConfig(`registry."docker.io"`, `mirrors = ["mirror.gcr.io"]`).
 		WithConfig(`grpc`, `address=["unix:///var/run/buildkit/buildkitd.sock", "tcp://0.0.0.0:1234"]`).
 		WithArg(`network-name`, `dagger-dev`).
-		WithArg(`network-cidr`, `10.88.0.0/16`)
+		WithArg(`network-cidr`, `10.88.0.0/16`).
+		WithArg(`debugaddr`, `0.0.0.0:6060`)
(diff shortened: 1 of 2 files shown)


rajatjindal:
I believe this is what I am running into when trying out CI for https://github.com/dagger/dagger/pull/8122. I tried running it locally, and this is where it is stuck.

jedevc (approved):

PR #8157, opened 2024-08-15
chore(ts): don't export pnpm store

jedevc:
This is the node_modules equivalent for pnpm - we should never export it, just in our tests these can end up at 100MB, which causes our exports to sometimes hang.


TomChv (approved):
LGTM

rajatjindal:
this fixed the CI problem I was having with my PR ❤️

PR #8168, opened 2024-08-16
tests: big deflake attempt

jedevc:
Includes:
- Attempt to kill misbehaving subprocesses in `TestDaggerUp` and similar *harder*, see https://dagger.cloud/dagger/traces/3e12c1a3aa6b40119ab74e136dad9e50?span=1e20b5449908ff06#0f8f4e636ab28886:L198
- `testctx.T` should have a way to allow tests to repeat

Getting test repetition working is not very fun, unsurprisingly, go's wonderous test framework does not want such a convenience.

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -13,6 +13,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


jedevc:
Grr, this way of doing things *really* does not work. The problem is `t.Parallel`, and generally the following constraint on `Run`:

```go
// Run may be called simultaneously from multiple goroutines, but all such calls
// must return before the outer test function for t returns.
```

Because we don't know ahead of time how many attempts we might be doing, we can't actually satisfy the above constraint - when the tests are started in parallel, we can't really solve the above at all.

`t.Parallel` is horrendous and such a painful construct to try and do things with. `gotestsum` might be the thing to do.

PR #8171, opened 2024-08-16
feat: implements --interactive-command for overriding the default command used in interactive mode

### LINKED ISSUE #7974: 'dagger --interactive': customize shell command
## Problem

`dagger --interactive` is awesome, but sometimes it doesn't work because the container being inspected doesn't have `/bin/sh` and as a result, the terminal fails.

## Solution

Let me customize the command to execute in the interactive container, or, perhaps, fall back to executing the default container with the target container's workdir and/or rootfs mounted? That way, I can at least look at the files.

samalba:
Adds the ability to override the default command (`/bin/sh`) when entering in interactive mode when something fails.

Fixes #7974

Note: I added shlex on the command so we can support fancy stuff like: `dagger call --interactive --interactive-command='sh -c "apk update && apk add bash && bash"'` (for example to setup the debug container with an external script).

Remaining:

- [x] add an integration test
- [x] add a changie entry (feature)

### DIFF
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"unicode"
 
+	"github.com/google/shlex"
 	"github.com/mattn/go-isatty"
 	"github.com/muesli/reflow/indent"
 	"github.com/muesli/reflow/wordwrap"
@@ -47,13 +48,15 @@ var (
 
 	workdir string
 
-	silent      bool
-	verbose     int
-	quiet, _    = strconv.Atoi(os.Getenv("DAGGER_QUIET"))
-	debug       bool
-	progress    string
-	interactive bool
-	web         bool
+	silent                   bool
+	verbose                  int
+	quiet, _                 = strconv.Atoi(os.Getenv("DAGGER_QUIET"))
+	debug                    bool
+	progress                 string
+	interactive              bool
+	interactiveCommand       string
+	interactiveCommandParsed []string
+	web                      bool
 
 	stdoutIsTTY = isatty.IsTerminal(os.Stdout.Fd())
 	stderrIsTTY = isatty.IsTerminal(os.Stderr.Fd())
@@ -169,6 +172,7 @@ var rootCmd = &cobra.Command{
 				"name": cmdName,
 			})
 		}
+
 		return nil
 	},
 }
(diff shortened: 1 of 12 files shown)


aluzzardi (approved):
LGTM

One comment:

buildkit has a built-in shlex, which would behave exactly like Dockerfiles thus less surprising. I don't know if we can use that directly since we're on top of a few dagger abstractions but worth looking into.

aluzzardi:
Also: CI is failing because you have to re-generate the reference docs when changing CLI flags. There's a dagger pipeline to do that (docs generate --export . or something)

samalba:
> buildkit has a built-in shlex, which would behave exactly like Dockerfiles thus less surprising. I don't know if we can use that directly since we're on top of a few dagger abstractions but worth looking into.

`llb.Shlex` is a tiny wrapper on top of the normal shlex. That's why we already add the dependency. I think we want to the normal shlex call here given we're lexing a simple shell command (I would not import the `llb` pkg in the cli just for that).

samalba:
Added the docs generate + one integration test. Ready to merge when the CI is green ✅

cmd/dagger/main.go
```diff
@@ -181,6 +185,7 @@ func installGlobalFlags(flags *pflag.FlagSet) {
 	flags.BoolVarP(&debug, "debug", "d", debug, "Show debug logs and full verbosity")
 	flags.StringVar(&progress, "progress", "auto", "Progress output format (auto, plain, tty)")
 	flags.BoolVarP(&interactive, "interactive", "i", false, "Spawn a terminal on container exec failure")
+	flags.StringVar(&interactiveCommand, "interactive-command", "/bin/sh", "Change the default command for interactive mode")
```
jedevc, line 188: Sorry for getting to the party late!

Potentially, how would you feel about having this merged into the existing `interactive` command? (we already do this with `--compat` for example). That way, we could have `--interactive` go to the default, and `--interactive="foo"` execute command `foo`. The one downside is that you *can't* do `--interactive "foo"`, you need to do it all in one arg with `=`.

Up to you, just offering another suggestion!

jedevc (approved):

cmd/dagger/main.go
samalba: I wasn't aware of the `--compat` one, interesting. I'd like to keep this way because 90% of the time, users will only use `--interactive`, and it's more intuitive from the help to just have a boolean command. With your suggestion, I would instead default to --interactive="", meaning the help would show `--interactive string[=""]`.

sipsma:
test failures may be fixed by https://github.com/dagger/dagger/pull/8196

jedevc:
Yup the test failures here look to be the same incomprehensible data races I saw prior to https://github.com/dagger/dagger/pull/8196

sipsma (approved):

Issue #8031, opened 2024-07-24
(Another different) `TestLegacy` flake


sipsma:
I saw [this flake](https://github.com/dagger/dagger/actions/runs/10082096372/job/27875587261) on main ([Traces](https://dagger.cloud/dagger/traces/cfafdb893c9e695f4ad9e6fac435e0ff)) and got very sad since it was in `TestLegacy` which should [have been fixed](https://github.com/dagger/dagger/issues/7900).

But this flake appears to be distinct from the old one, even though part of the error message is the same:
1. The mismatch in files occurs in `TestLegacyExportAbsolutePath` and crucially is in `internal/dagger/dagger.gen.go` rather than in `go.mod`
2. Looking through the engine logs from that run, I can see that filesync cache refs are not being re-used between different client sessions, which was the important part of the root cause for the other flake
3. I could repro the previous flake locally by running `TestLegacy` in a loop for at most an hour (but usually closer to 20m). But I have been running that same loop on main for going on 3 hours now locally with it passing every time (takes about 45s per run).

---

The error this time looks like
```
	Error Trace:	/app/core/integration/legacy_test.go:62
	Error:      	Received unexpected error:
	            	input: container.from.withMountedFile.withWorkdir.withExec.withNewFile.withNewFile.withExec.stdout resolve: process "dagger --debug query" did not complete successfully: exit code: 1
	            	
	            	Stderr:
	            	Error: failed to serve module: input: module.withSource.initialize resolve: failed to initialize module: failed to call module "bare" to get functions: call constructor: process "go build -o /runtime ." did not complete successfully: exit code: 1
	            	
	            	Stderr:
	            	internal/dagger/dagger.gen.go:23:2: package dagger/test/internal/querybuilder is not in std (/usr/local/go/src/dagger/test/internal/querybuilder)
	            	internal/dagger/dagger.gen.go:24:2: package dagger/test/internal/telemetry is not in std (/usr/local/go/src/dagger/test/internal/telemetry)
	Test:       	TestLegacy/TestLegacyExportAbsolutePath
```

I spent a while pouring through the engine logs to find any clues, but everything checks out and looks expected from what I can see based on what we include there.

I don't have any strong hunches yet, but some a couple things to note/investigate:
1. Not clear yet how often this happens, just saw it this one time. Can't repro locally so will just have to see what happens in CI
[...]

PR #8181, opened 2024-08-19
chore: increase verbosity of codegen error

jedevc:
This error seems to appear as a weird little flake (see https://github.com/dagger/dagger/issues/8031#issuecomment-2260352601) - and never seems to appear legitimately.

Because of this, we should crank the verbosity up to try and understand what the go parser is *actually* seeing when this happens, which might give us a hint as to what is really going on.

### DIFF
--- a/cmd/codegen/generator/go/templates/modules.go
+++ b/cmd/codegen/generator/go/templates/modules.go
@@ -117,7 +117,7 @@ func (funcs goTemplateFuncs) moduleMainSrc() (string, error) { //nolint: gocyclo
 	}
 
 	if ps.daggerObjectIfaceType == nil {
-		return "", fmt.Errorf("cannot find default codegen %s interface", daggerObjectIfaceName)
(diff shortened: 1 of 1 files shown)


jedevc:
Well clearly something is off here:

```
	            	Error: generate code: template: module.go.tmpl:79:3: executing "_dagger.gen.go/module.go.tmpl" at <ModuleMainSrc>: error calling ModuleMainSrc: cannot find default codegen DaggerObject interface in:

	            	[Bare, DaggerObject, ExecError, Tracer, dag, marshalCtx, setMarshalContext]
```

:thinking: **it's there**

jedevc:
@vito pointed out that this occurred very consistently in https://github.com/dagger/dagger/pull/8150. The cause of this was the new `types.Alias` types that are by default generated in go 1.23.

It feels *unlikely* that this is the cause here :thinking: But, obviously, not impossible - it could be enabled in old versions of go using `GODEBUG`, but we don't set that explicitly anywhere...

gerhard (approved):
👍 🚀

PR #8185, opened 2024-08-19
tests: move TypeScript test to its own suite

TomChv:
As discussed with @gerhard and it order to improve current CI, I'm making an experimentation by moving the TypeScript tests to its own suite so we can run it independently from other's suites.
The long term idea is to move group of tests into their suites so they can be run into different engine and hopefully remove flakes.

Local CI run: https://dagger.cloud/Quartz/traces/a2c3a4a24346f31efac7dc52f23b6137

I'll try to paralize some tests and see if it can also make it faster.

@gerhard I let you create a GA job to run this suite 🚀 

Local CI took 7minutes to exec tests (https://dagger.cloud/Quartz/traces/a2c3a4a24346f31efac7dc52f23b6137#7c9a8264d0620b45) (4 min setup on my local machine)

### DIFF
--- a/core/integration/module_typescript_test.go
+++ b/core/integration/module_typescript_test.go
@@ -4,13 +4,20 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"testing"
 
 	"github.com/dagger/dagger/testctx"
 	"github.com/stretchr/testify/require"
 	"github.com/tidwall/gjson"
 )
 
-func (ModuleSuite) TestTypescriptInit(ctx context.Context, t *testctx.T) {
+type TypescriptSuite struct {}
+
+func TestTypescript(t *testing.T) {
+	testctx.Run(testCtx, t, TypescriptSuite{}, Middleware()...)
+}
+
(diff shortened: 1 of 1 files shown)


TomChv:
It seems I cannot add `t.Parallel` because everything is already parallized.

```
panic: testing: t.Parallel called multiple times [recovered]
        panic: testing: t.Parallel called multiple times
```

I'm not sure I can also mutualize `connect` since it might break traces... I'm looking for some optimization

jedevc:
> It seems I cannot add `t.Parallel` because everything is already parallized.

See the test middleware `WithParallel` introduced in https://github.com/dagger/dagger/pull/7532. If the tests are run with that middleware, they're already in parallel and can't be parallelized again.

TomChv:
> See the test middleware WithParallel introduced in https://github.com/dagger/dagger/pull/7532. If the tests are run with that middleware, they're already in parallel and can't be parallelized again.

Yep exactly! I'm not sure I have much area to optimize the TS test generally, it's still going decently but some tests are still long and traces are not as helpful.

jedevc (approved):

Issue #8184, opened 2024-08-19
Engine & CLI tests are reliable & fast


gerhard:
![badgen.net](https://badgen.net/static/Last%20updated%20on%202025-02-10/ONGOING/red)

Reliable = pass `99%` of the time.
Fast = `< 10mins` to pass.

| Dagger version                                                                                                                                                              | Engine & CLI main branch reliability                                                                                                               | Pass duration (P95)                                                                                                                                | Fail duration (P95) |
| --                                                                                                                                                                          | --                                                                                                                                                 | --                                                                                                                                                 | --                  |
| `v0.15.3` 5 days, 18 runs                                                                                                                                                   | [`17%`](https://ui.honeycomb.io/dagger/environments/experimental/datasets/dagger-dagger-github-actions/board-query/JzDnrpu3Jvg/result/8KdmaRuCDpn) | [`20m`](https://ui.honeycomb.io/dagger/environments/experimental/datasets/dagger-dagger-github-actions/board-query/jX8xZEStYP2/result/cZALvSh7rEF) | `23m`               |
| `v0.15.2` 19 days, 64 runs                                                                                                                                                  | [`50%`](https://ui.honeycomb.io/dagger/environments/experimental/datasets/dagger-dagger-github-actions/board-query/JzDnrpu3Jvg/result/HTRRTdZBTR7) | [`22m`](https://ui.honeycomb.io/dagger/environments/experimental/datasets/dagger-dagger-github-actions/board-query/jX8xZEStYP2/result/3imbTtc1i2q) | `32m`               |
| `v0.15.[0-1]` 35 days, 140 runs                                                                                                                                             | [`56%`](https://ui.honeycomb.io/dagger/environments/experimental/datasets/dagger-dagger-github-actions/board-query/JzDnrpu3Jvg/result/BkaD4CFUB18) | [`22m`](https://ui.honeycomb.io/dagger/environments/experimental/datasets/dagger-dagger-github-actions/board-query/jX8xZEStYP2/result/FbEXCdaBFUU) | `32m`               |
[...]

PR #8191, opened 2024-08-20
tests: try killing subprocesses harder if they misbehave

jedevc:
Part of https://github.com/dagger/dagger/issues/8184, and split out of https://github.com/dagger/dagger/pull/8168, since my general approach there turns out to not be possible :cry: 

Somehow, we see weird timeouts because calls to the dagger cli aren't properly exiting out: https://dagger.cloud/dagger/traces/da4d9c2d21aef7f3eb8ff0856946c0c1#7e1bbffd12a2fbe1

This is it's own problem, but we shouldn't hang the test while waiting for this, we should actually exit and fail after a timeout.

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -13,6 +13,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


gerhard (approved):
👍 🚀

PR #8199, opened 2024-08-21
docs: fix sudo sh install example

jedevc:
The install example has been broken for some time! `sudo` doesn't forward environment variables by default for security purposes, and requires the `-E` flag to do this correctly.

### DIFF
--- a/docs/current_docs/partials/_install-cli.mdx
+++ b/docs/current_docs/partials/_install-cli.mdx
@@ -41,7 +41,7 @@ If your user account doesn't have sufficient privileges to install in `/usr/loca
 
 ```shell
 cd /usr/local
-curl -L https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.12.5 sudo sh
+curl -L https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.12.5 sudo -E sh
 ```
 
 </TabItem>
--- a/docs/current_docs/partials/_install-cli.mdx
+++ b/docs/current_docs/partials/_install-cli.mdx
@@ -30,7 +30,7 @@ If you do not have Homebrew installed, or you want to install a specific version
 
 ```shell
 cd /usr/local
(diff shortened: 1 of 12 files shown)


docs/current_docs/partials/_install-cli.mdx
```diff
@@ -41,7 +41,7 @@ If your user account doesn't have sufficient privileges to install in `/usr/loca
 
 ```shell
 cd /usr/local
-curl -L https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.12.5 sudo sh
+curl -L https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.12.5 sudo -E sh
```
thaJeztah, line 44: Be aware that, depending on the configuration (look for `secure_path`), the `PATH` variable may still be reset with a default path (IIRC even with `-E` set); https://superuser.com/a/927599

thaJeztah: Oh! Also the `curl` should probably use `-fsSL` to force TLS, follow redirects, and to make sure it doesn't output garbage when something fails 😅

jedevc: Hallo! :wave: fancy seeing you here :smile: 

Yeah, the `PATH` isn't actually the issue, it's just about preserving the `DAGGER_VERSION`/`BIN_PATH` envs.

Also updated to use `-fsSL` consistently, we did something similar in https://github.com/dagger/dagger-for-github/commit/e267a26198f567f715e45aa3b23036347c45b147, but looks like it never made it over here!

Thanks :heart:

thaJeztah: 👋 heh, yeah, I'm subscribed to the repo; mostly to peek over the fence once in a while to see if we introduced regressions or otherwise, and I saw `sudo` and knew about some of the fun little (and often overlooked) corners, so it triggered my curiosity 😄

thaJeztah (commented):
changes look sane to me

LGTM 😄

PR #8203, opened 2024-08-21
[DNM] deflaking CI

sipsma:


sipsma:
Re-opening to see if I can catch https://github.com/dagger/dagger/issues/8554 in the act

PR #8212, opened 2024-08-22
ci: split `TestModule` into multiple suites

TomChv:
Run a specific suite:

```
dagger -m .dagger call --source=.:default test specific --pkg="./core/integration" --run="^MySuite$"
```

New suite after the split
- `TypescriptSuite` (Done in https://github.com/dagger/dagger/pull/8185): Execute all module TypeScript tests (run in 8m)
- `PythonSuite`: Execute all module Python tests (run in 8m30s)
- `GoSuite`: Execute all module Go tests (run in 5m15s)
- `ElixirSuite`: Execute all module Elixir tests (run in 2m15s
- `PHPSuite`: Execute all module PhP tests (run in 1m40s)
- `DaggerCallSuite`: Execute all tests from `module_call_test.go` (run in 10m)
- `DaggerConfigSuite`: Run all tests from `module_config_test.go` (run in 13m)
- `DaggerCLISuite`: Run all tests from `module_cli_test.go` (run in 4m30)
This is the result of a rename of `module_function` and the move of some CLI related tests from `module_config` to `module_cli`
- `TestInterface`: Run all test from `module_iface_test.go` (run in 2m)
- `TestDaggerType`: Run all tests related to module’s type (run  in 8m)

`TestModule` still exists but is much smaller.

:bulb: I think there are better way to split the tests, but this is a first draft to gather teams's opinion.

Now I think the ideal would be to run some suite together in the CI, for example having a function `dagger call test sdks` that run all SDK related suite etc...

Screenshot 2024-08-22 at 12 03 56

Some suites are still super expensive in compute resources but at least they are all runnable locally now :)

@gerhard let me know what I can do around the CI itself to improve it too!


TomChv:
@gerhard The PR isn't mergeable in this state, but several changes need to happen in our GA to make it work.

I'll need your advice on this seems it might be counter-productive to spawn multiple engine as you explained

> Each test run incurs a 3-5 minutes penalty since:
A new EC2 instance needs to be provisioned & join the EKS cluster
Dagger CLI + Engine need to be installed
Our dagger/dagger/.dagger module needs to be installed
A dev Engine needs to be built and spun up as a service

> If the tests take up to 5 minutes to run, then we would be paying an extra 3-5 minutes penalty by splitting them out.

gerhard (commented):
I went through the code changes, looking good @TomChv 👍

I need to run this locally, and compare the test suites before and after to make sure that everything behaves correctly before approving & merging. Will do it right after the next meeting.

Super excited to see that big `TestModule` suite getting split into smaller suites. At least things are manageable now! The fact that you are able to run subsets of interest locally is most promising.

TomChv:
> Yes, indeed, that is the second part which should be done part of this PR too, for sure.

Yes because right now I'm not sure all tests are actually run

gerhard:
@TomChv all tests are still running part of `Engine & CLI / test`.

I am now going to run the test suites locally, and see how they compare before & after.

gerhard:
I ran `dagger call --source=.:default test list | tee test-list.txt` and then compared the two results:

```console
diff dagger/test-list.txt dagger.TomChv/test-list.txt

11a12,15
> TestDaggerCLI
> TestDaggerCall
> TestDaggerConfig
> TestDaggerType
14a19
> TestElixir
23a29
> TestGo
30a37
> TestInterface
55a63
> TestPHP
66a75
> TestPython
```

This is exactly as expected. New test suites are showing in this branch, but otherwise all other tests are the same.

gerhard:
Given the following host `gerhard-2024-08-23-c-32-intel-lon1-01`:
- `32vCPUs` [Intel Xeon Gold 6548N Processor](https://www.intel.com/content/www/us/en/products/sku/237567/intel-xeon-gold-6548n-processor-60m-cache-2-80-ghz/specifications.html)
- `64GB` DDR5
- `400GB` VIRTIO NVMe
- Dagger `0.12.5` on Docker `27.1.2`
- Linux `6.8.0` via Ubuntu `24.04` LTS

## One-by-one

| TEST SUITE | DURATION |
| - | - |
| `TestDaggerCLI` | [✅ `1m 42s`](https://dagger.cloud/dagger/traces/76e8a283fbbd5fbd7e9bd55aab7754bb) |
| `TestDaggerCall ` | [✅ `2m 3s`](https://dagger.cloud/dagger/traces/7179624424bb483ed55da1c3c8fabde8) |
| `TestDaggerConfig` | [✅ `2m 18s`](https://dagger.cloud/dagger/traces/590624711ccc941c4ee707f209f17f20) |
| `TestDaggerType` | [✅ `1m 55s`](https://dagger.cloud/dagger/traces/5b8fb21fe778226806bc344b4839e3eb) |
| `TestElixir ` | [✅ `1m 8s`](https://dagger.cloud/dagger/traces/273292ff69dd5eaab5177890d7945ab3) |
| `TestGo` | [✅ `1m 38s`](https://dagger.cloud/dagger/traces/beb5201ad8507b2f43a6df7b11e89925) |
| `TestInterface` | [✅ `58s`](https://dagger.cloud/dagger/traces/49430b54ad9b6c8141e4266680cdfcc5) |
| `TestPHP ` | [✅ `56s`](https://dagger.cloud/dagger/traces/6e24e267247630a41c93c2ae67df596f) |
| `TestPython ` | [✅ `2m 14s`](https://dagger.cloud/dagger/traces/62439bd07c3e9e4d8c6469a8cd6c32fe) |
| `TestModule ` | [✅ `4m 10s`](https://dagger.cloud/dagger/traces/d7558c325f67aada7f64c7f88f0aeb56) |

I'm noticing that all tests passed the first time. No flakes in sight.

I am now going to try and run them grouped, e.g. `--run="TestGo|TestPython|TestTypescript|TestElixir|TestPHP"` etc.

TomChv:
@gerhard Thanks for all these updates! Looks like it has a positive impact on the CI 🚀

jedevc:
I think we should probably keep the same tests in `testdev` that were there before.

i.e. everything that `TestModule` has been split into as part of this PR can be included.

---

Something we should do btw - we should make it easier to group suites together. We could do this with the idea of "embedded suites" - we just need to extend `testctx` to look into embedded structs and traverse those as well.

That way we can define the logic of "groups of tests" *in* the integration test code, instead of having the definition in GitHub YAML - which we're trying to move away from.

gerhard:
If we keep the same tests in `testdev` as before `TestModule` was split, we can't expect anything to change in terms of its stability.

In the interest of keeping the same test coverage, I will split `testdev` into:
1. `testdev-module-runtimes`
2. `testdev-modules`
3. `testdev-container`

I am less certain about keeping the same size for the CI runners, but I would like to keep changes to a minimum, so that will need to happen after we see a few runs on `main` after this split.

sipsma:
I generally think that we need to reconsider whether we even need separate `test` and `testdev` in the first place anymore. That split happened a long time ago and I think in the new world we probably can just get away with only `testdev`.

But given that, in the meantime keeping the coverage the same but splitting up `testdev` sounds fine.

gerhard:
I really like that idea @sipsma!

Since `test` already builds and tests a dev Engine inside Dagger, could we instead remove `testdev`? That would remove our dependency on Docker, which has a bunch of complications that span the entire stack (including how we provision those CI runners).

sipsma:
> I really like that idea @sipsma!
> 
> 
> 
> Since `test` already builds and tests a dev Engine inside Dagger, could we instead remove `testdev`? That would remove our dependency on Docker, which has a bunch of complications that span the entire stack (including how we provision those CI runners).

The point of testdev is to run our CI code itself against the dev engine, which is important to catch any issues pre-release. Otherwise our CI gets broken only once we do the release. So if we only have one we need to have testdev.

gerhard:
> The point of testdev is to run our CI code itself against the dev engine, which is important to catch any issues pre-release. Otherwise our CI gets broken only once we do the release. So if we only have one we need to have testdev.

Got it. Instead of building & running the dev engine in Docker, could we instead use `dagger-g2-main-...`? This is the last version built from `main`. WDYT @jedevc @sipsma ?

jedevc:
Potentially we can use this, but now we have an implicit dependency on `main` - `testdev` (or it's equivalent) would then need to run only after the publish step. This also means it can't be run in PR environments, so we wouldn't be able to make `testdev` (or equivalent) the default.

IMO, the way to do this properly would be to:
- Build the engine and CLI, and upload them as a workflow artifact
- Then each next test/lint/etc step should download the artifact and start the "dev" engine in docker based on this
- Also, the publish step just becomes: download artifacts, publish artifacts.

This is a pretty substantial rework, but since we can't publish an image/cli from a PR, the default testing environment can't involve consuming the `dagger-g2-main...` env.

sipsma:
@jedevc @gerhard Something like https://github.com/dagger/dagger/pull/7771 should avoid the dep on docker in the infra

Or if that isn't salvageable for whatever reason, we can try variations on it like:
* build dev engine, proxy service to localhost and run CI against that
* just straight-up run docker as a service in the engine, proxy to localhost (feels silly somehow, but also avoids the hard dep on infra and creates the same end effect as today, so 🤷‍♂️)

gerhard:
I understand the importance of `testdev` over `test`, which means starting with an outer dev engine built from the PR source, rather than the latest released version.

I also understand why we can't use `:main` for this - we need an Engine built from the source in the PR.

Let me think about this some more, and see where we land with `testdev`.

The current change looks good enough for now, I am keen on merging this so that we can start answering:
- How often do these new pipelines fail?
- Which of these pipelines fails the most?
- How long do these pipelines take to run?

Just to re-emphasize, `test` & `testdev` remained as they were before this split, we can compare the new, smaller pipelines against the original ones. Today's snapshot which shows the state of flakes **after** the first split:
image

As soon as this gets merged, we will start capturing the behaviour of these new pipelines.

gerhard (approved):
LGTM 👍 🚀

Leaving the actual merge to today's sync, in case there are any last-minute objections.

PR #8215, opened 2024-08-23
Literally run Dagger docs

sourishkrout:
I'm curious to hear feedback on this DevX concept. It's a throwback to the Magefile (or Taskfile/Makefile, etc.).

However, instead of maintaining some separate task definitions, the "tasks" are lifted out of the documentation (as long as they are cells "exported" with a name). The description is a cell's preceding paragraph in the markdown AST. Made some minimal edits to the existing `.dagger/README.md`.

https://github.com/user-attachments/assets/df858c80-6bb3-4508-b9d9-7151319ab911

It's trivial to run individual tasks, e.g. `runme run dev-dagger-cli` (from `runme list`) and use them as test targets in CI/CD (via https://github.com/stateful/runme-action or first-class Dagger function) effectively testing your DevX and documentation.

PS: I'm working on bringing `export SRC="$(git rev-parse --show-toplevel):default"` back. However, I wanted to share the concept first.

### DIFF
--- a/.dagger/README.md
+++ b/.dagger/README.md
@@ -1,3 +1,8 @@
+---
+cwd: ..
+terminalRows: 16
+---
+
 # Developing Dagger with Dagger
 
 This Dagger module defines pipelines to develop the Dagger itself, including building and releasing the CLI, engine and SDKs.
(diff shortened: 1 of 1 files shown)


.dagger/README.md
```diff
@@ -12,117 +17,144 @@ A module is a collection of functions (and types) which can be used by `dagger c
 
 To discover available functions in a given module: `dagger functions`.
 
-With that in mind: this document includes examples of *typical pipelines* that are useful while developing Dagger. But remember, you are free to compose your own pipelines, either by modifying the examples, of starting from scratch. This flexibility is one of the killer features of Dagger.
+```sh {"id":"01J5Y82842KAER30RGCN96HH6M"}
```
jedevc, line 20: :eyes: what are these IDs?

Does editing this doc then mean that adding new codeblocks *requires* the use of the vscode extension to edit it? If so, that feels quite overly restrictive, and limits who can edit this.

---

Aside from that, not really any strong opinions on the change, I'm not sure what value this provides (it's really just a cheatsheet for devs), but also it's not overly invasive, so :woman_shrugging:

shykes:
I love Runme and I'm all in favor of hosting something like this somewhere in the project, but not in the main README. That's too invasive. Sorry

.dagger/README.md
sourishkrout: > 👀 what are these IDs?
> 
> Does editing this doc then mean that adding new codeblocks _requires_ the use of the vscode extension to edit it? If so, that feels quite overly restrictive, and limits who can edit this.

Nobody's required to use Runme CLI/extension for editing or viewing - it's just markdown. The IDs are optional, and I should have turned them off. They easily evaded my selective perception because I'm so used to seeing them.

> could potentially be interesting for `RELEASING.md` in it's current state

I'm happy to help if you ever decide to take a stab at it.

sourishkrout:
> I love Runme and I'm all in favor of hosting something like this somewhere in the project, but not in the main README. That's too invasive. Sorry

That's fair. Appreciate the feedback, @shykes & @jedevc 👍.

Would you guys be open to me closing this and sending another PR that converts any command containing a command to use fenced code blocks? Plain old markdown... no attributes or additional words.

That way, I can run the docs myself, and nobody has to be the wiser about Runme or not, which is always optional.

Issue #6475, opened 2024-01-22
✨ Add WithFiles to Directory


sagikazarmark:
### What are you trying to do?

I find myself writing code like this lately:

```go
dir := dag.Directory()

for _, file := range files {
    dir = dir.WithFile("", file)
}
```

I'd like to see a shorthand where I can easily add a list of files to a directory:

```go
dir := dag.Directory().WithFiles(files...) // OR: WithFiles(files)
```

### Why is this important to you?

When using CLI tools (or even in Dockerfiles) files and directories are usually described by a path without having to make a distinction between them.

It's not the case in Dagger, but in many cases a list of files (result of some build step) may be better represented as a directory (archiving).

### How are you currently working around this?

See code snippet above.

tomasmota:
If the maintainers are ok with this suggestion I'd like to try implementing it

helderco:
+1 from me. In the meantime you can [make a util](https://docs.dagger.io/592101/custom-callbacks#example):

```go
dir := dag.Directory().With(util.Files(files...))
```

jedevc:
@sagikazarmark @helderco @tomasmota WDYT about the idea of a `withoutFiles` property?

Currently, there's not really a way of removing multiple files, might be nice to have something like that as well.

helderco:
Yes, sgtm 👍

PR #8216, opened 2024-08-23
feat: add withoutFiles on Container and Directory

jedevc:
Continuing on from #6475, we should have `withoutFiles`, just as we have `withFiles`.

Ended up wanting this for working on an internal test, so decided to just add it :tada:


helderco (approved):
LGTM

PR #8217, opened 2024-08-23
fix: handle comment parsing with windows line endings

jedevc:
Our comment parsing code was previously very inflexible, and didn't handle carriage returns in the comment. I *originally* thought I fixed this in https://github.com/dagger/dagger/pull/7121, but I guess not :cry: 

This is because of this helpful little note in go's comment ast struct declaration:

	// The Text field contains the comment text without carriage returns (\r) that
	// may have been present in the source. Because a comment's end position is
	// computed using len(Text), the position reported by [Comment.End] does not match the
	// true source end position for comments containing carriage returns.

So calling `.End()` on a comment essentially returns absolutely nonsense when there are carriage returns - there is no good way of recovering these.

Instead, we switch to another approach of detecting comment doc+line comments - instead of attempting to look for the last element of the line in the comment, we perform an entirely line-based approach, by checking if the comment starts/ends on the correct line.

This is relatively fiddly, and is a bit fragile, so added some more tests as well. Also, handling block comments `/* */` becomes a bit trickier, so extended tests to handle those as well.

### DIFF
--- a/cmd/codegen/generator/go/templates/modules.go
+++ b/cmd/codegen/generator/go/templates/modules.go
@@ -998,24 +998,29 @@ func (ps *parseState) commentForFuncField(fnDecl *ast.FuncDecl, unpackedParams [
 		}
 
 		if allowDocComment {
-			// take the last position in the last line, and try and find a
-			// comment that contains it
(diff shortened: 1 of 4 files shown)


gerhard (approved):
LGTM 👍 🚀

PR #6100, opened 2023-11-13
Bump otelgrpc to fix CVE issue

wingyplus:
Found in the Trivy scan in #6099. Bump `otelgrpc` library to 0.46.0 and cleanup dependencies with `go mod tidy`.

PR #8219, opened 2024-08-23
Use a `dagger core` command as readiness probe

matipan:
Since https://github.com/dagger/dagger/pull/6100#issuecomment-1809271407 that using buildctl as the healthcheck for the engine is essentially a no-op. There is an on-going discussion [here](https://github.com/dagger/dagger/issues/7733) of what the ideal healthcheck for the engine would be, we haven't yet reached a conclusion but now that we have the `dagger` CLI in the engine image and the `core` subcommand we can use something more suitable on our helm chart. I don't think this should be considered as "the way" to check for engine readiness, but its the best I could come up with and what we were doing was already broken.

Internally we were doing something similar to this already, but taking care of downloading the CLI and using `dagger query`. Now with the CLI in the engine image and the `core` subcommand we can simplify it and officially release it here.


TomChv (approved):

helm/dagger/templates/engine-daemonset.yaml
```diff
@@ -70,7 +70,11 @@ spec:
           resources: {{- toYaml .Values.engine.resources | nindent 12 }}
           readinessProbe:
             exec:
-              command: ["buildctl", "debug", "workers"]
+              command:
+                - sh
+                - -exc
+                - |-
+                  dagger core container with-new-file --contents "We are ready" --path "/tmp/ready" file --path "/tmp/ready" contents
```
gerhard, line 77: This is a great, much needed change!

Instead of using `dagger core` which is unstable and likely to change - see https://github.com/dagger/dagger/pull/8135 - I propose that we use this command instead: `echo "{ directory { id } }" | dagger query`

FTR:
```console
❯ echo "{ directory { id } }" | time dagger query
Setup tracing at https://dagger.cloud/traces/setup. To hide: export SHUTUP=1

✔ connect 0.7s
✔ cache request: mkfile /schema.json 0.0s
✔ load cache: mkfile /schema.json 0.0s
✔ cache request: blob://sha256:b618d3d179857032b7a416ad037342be687116289c0fd1cc8b4a39a9f2d65d4e 0.0s
✔ load cache: blob://sha256:b618d3d179857032b7a416ad037342be687116289c0fd1cc8b4a39a9f2d65d4e 0.0s
✔ directory: Directory! 0.0s

{
    "directory": {
        "id": "ChV4eGgzOmMwMmVlMmZiODlhYjRmMDQSSgoVeHhoMzpjMDJlZTJmYjg5YWI0ZjA0EjESDQoJRGlyZWN0b3J5GAEaCWRpcmVjdG9yeUoVeHhoMzpjMDJlZTJmYjg5YWI0ZjA0"
    }
}
dagger query  0.16s user 0.14s system 20% cpu 1.492 total
```

**After** `dagger core` stabilises, we can easily switch to it.

jedevc: I'd add to this - we don't even need to do a full operation at all, we could just do `{ version }` as a query. We're really just interested if the dagger server is running at all, I don't think we need to simulate an operation here?

gerhard (approved):
👍 🚀

jedevc:
Potentially worth a changelog entry? Up to ya :tada:

PR #8221, opened 2024-08-24
fix(core): drop leading slash from paths in withVCSGeneratedPaths API

wingyplus:
This fix comes from accidentally add a leading slash in the generated paths during fix the Elixir issue like this:

```go
modName := "potato" // assume the name is this.
dag.GeneratedCode(directory).
    WithVCSGeneratedPaths([]string{"/src/"+modName+"/lib/mix/tasks/dagger.invoke.ex"})
```

The `.gitattributes` produces as:

```
//src/potato/...
```

It's happens because the API prepending it without detecting the path that user given. This fixes by dropping it if the path already contained.

### DIFF
--- a/core/schema/module.go
+++ b/core/schema/module.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"dagger.io/dagger/telemetry"
 	"golang.org/x/sync/errgroup"
@@ -1125,6 +1126,7 @@ func (s *moduleSchema) updateCodegenAndRuntime(
 				// already has some config for the file
 				continue
 			}
+			fileName := strings.TrimPrefix(fileName, "/")
 			gitAttrsContents = append(gitAttrsContents,
 				[]byte(fmt.Sprintf("/%s linguist-generated\n", fileName))...,
 			)


wingyplus:
I'm not sure if it's make sense to fix. If it's ok, I will find the way to add it into tests. 🙏

TomChv (commented):
I think the fix make sense for consistency, add tests and I can approve it :D

You could add some in our `TestModule` suite, not that it will be part of a smaller suite once https://github.com/dagger/dagger/pull/8212 is merged so maybe it's worth waiting for it ;)

wingyplus:
@TomChv I’ll waiting for that PR to be merged then. :)

core/schema/module.go
```diff
@@ -993,6 +993,9 @@ func (s *moduleSchema) updateCodegenAndRuntime(
 				// already has some config for the file
 				continue
 			}
+			if fileName[0] == '/' {
+				fileName = fileName[1:]
+			}
```
jedevc, line 998: `strings.TrimPrefix("/")`

SGTM though, don't think this needs tests.

wingyplus: How do I forgot the trim prefix function. I will do fix it in 1-2 hours.

wingyplus: Fixed. And re-run all failed jobs.

wingyplus:
Rebased the change.

Would you like me to add a test?

jedevc (approved):

PR #8222, opened 2024-08-25
fix(sdk/elixir): fix return enum value on generated functions

### LINKED ISSUE #8173: 🐞 elixir: a function that return scalar type return incorrect result
### What is the issue?

Let's see the `TypeDef.kind` function:

https://github.com/dagger/dagger/blob/a453b63780d5ffb572966beda4d79c3ae3c59ca5/sdk/elixir/lib/dagger/gen/type_def.ex#L96-L102

It have 2 issues:

1. The Typespec returns incorrect result, it should be `{:ok, Dagger.TypeDefKind.t()}` instead of `Dagger.TypeDefKind.t()`.
2. The result return `{:ok, scalar_value}`, the `scalar_value` is string not an atom defined in `Dagger.TypeDefKind.t()`.

### Dagger version

0.12.5

wingyplus:
Fix any functions that return enum value to convert string into atom after request successfully. And fix Typespec to be `{:ok, <enum_type>}` instead of just `<enum_type>`.

Fixes #8173


TomChv (approved):

TomChv:
Ok to merge, engine tests fails are not related to elixir.

PR #8232, opened 2024-08-26
initialize: trace docker ps command

aluzzardi:
@sipsma @helderco Unsure how to test this since using the dev engine bypasses this code

Context: Over 1 second is lost on every dagger command (at least on macOS with d4m). Trying to fo figure out what's up. Example: https://dagger.cloud/dagger/traces/6642af8fb7e4308be318136ff67c83b5?span=e5599dec41cad509

Most of the time is spent in unaccounted parts of `starting engine` (there's only ~200ms out of ~900ms that are "spanned").

### DIFF
--- a/engine/client/drivers/docker.go
+++ b/engine/client/drivers/docker.go
@@ -81,7 +81,11 @@ func (d *dockerDriver) create(ctx context.Context, imageRef string, opts *Driver
 		// We only have a tag in the image ref, so resolve it to a digest. The default
 		// auth keychain parses the same docker credentials as used by the buildkit
 		// session attachable.
-		if img, err := remote.Get(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain), remote.WithUserAgent(opts.UserAgent)); err != nil {
+		_, span := otel.Tracer("").Start(ctx, "resolve image")
+		img, err := remote.Get(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain), remote.WithUserAgent(opts.UserAgent))
+		telemetry.End(span, func() error { return err })
+
+		if err != nil {
 			slog.Warn("failed to resolve image; falling back to leftover engine", "error", err)
 			if strings.Contains(err.Error(), "DENIED") {
 				slog.Warn("check your docker registry auth; it might be incorrect or expired")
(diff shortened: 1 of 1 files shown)


nipuna-perera:
I was about to report this! Glad it's been looked at. It hangs there for literally every call. Even when it's a subsequent identical call. FWIW, I am not using cloud.

marcosnils (approved):

jedevc (approved):
LGTM, but I have another idea of where this slow down might be coming from :tada:

---

We should also add tracing for: https://github.com/dagger/dagger/blob/421b94aec0e9eeb386ddde4b3102e1d7763b5b16/engine/client/drivers/docker.go#L81-L89

This is a remote API call to the registry, it's definitely possible that this call takes a little bit to complete.

The image ref in the CLI *isn't* digested, so every dagger invocation is going down this path, so we're constantly hitting the registry.

---

Ideally, we *should* be storing the digest. This is a bit of a release mangling issue though, since now the CLI needs to be published *after* the engine, so we know the digest of the image (and not sure how we handle the CLI in the engine, I guess we just use the tag there).

Another easy thing to do here would be to work off the assumption that all our registry tags are immutable - if we've downloaded `:v0.12.5`, then we shouldn't need to keep hammering the registry to find if the digest has changed.

aluzzardi:
@jedevc 🙏 

Done, PTAL

engine/client/drivers/docker.go
```diff
@@ -81,7 +81,11 @@ func (d *dockerDriver) create(ctx context.Context, imageRef string, opts *Driver
 		// We only have a tag in the image ref, so resolve it to a digest. The default
 		// auth keychain parses the same docker credentials as used by the buildkit
 		// session attachable.
-		if img, err := remote.Get(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain), remote.WithUserAgent(opts.UserAgent)); err != nil {
+		_, span := otel.Tracer("").Start(ctx, "resolve image")
+		img, err := remote.Get(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain), remote.WithUserAgent(opts.UserAgent))
+		telemetry.End(span, func() error { return rerr })
```
jedevc, line 86: nit: should be `return err` here, `rerr` is the return value from the function which isn't set at this point.

jedevc (approved):
One nit, otherwise LGTM :tada:

engine/client/drivers/docker.go
aluzzardi: 🙏

Issue #8204, opened 2024-08-21
Test race condition in `setupPrivateRepoSSHAgent`


sipsma:
First seen here: https://github.com/dagger/dagger/pull/8196#issuecomment-2301840224
* We reverted that PR because it for *some* reason seems to make this more likely to get hit
* However afaict the race condition that was detected exists independent of the changes that were reverted

```
WARNING: DATA RACE
Read at 0x00c002eef703 by goroutine 57880:
  testing.(*common).logDepth()
      /usr/lib/go/src/testing/testing.go:1023 +0x504
  testing.(*common).log()
      /usr/lib/go/src/testing/testing.go:1010 +0xa4
  testing.(*common).Logf()
      /usr/lib/go/src/testing/testing.go:1061 +0x6a
  github.com/dagger/dagger/testctx.(*T).Logf()
      /app/testctx/testctx.go:182 +0x90
  github.com/dagger/dagger/core/integration.setupPrivateRepoSSHAgent.func1()
      /app/core/integration/ssh_helper.go:55 +0xe4
  github.com/dagger/dagger/core/integration.setupPrivateRepoSSHAgent.func2()
      /app/core/integration/ssh_helper.go:62 +0x202

Previous write at 0x00c002eef703 by goroutine 40731:
  testing.tRunner.func1()
      /usr/lib/go/src/testing/testing.go:1676 +0x8fa
  runtime.deferreturn()
      /usr/lib/go/src/runtime/panic.go:602 +0x5d
  testing.(*T).Run.gowrap1()
      /usr/lib/go/src/testing/testing.go:1742 +0x44

Goroutine 57880 (running) created at:
  github.com/dagger/dagger/core/integration.setupPrivateRepoSSHAgent()
      /app/core/integration/ssh_helper.go:58 +0x684
  github.com/dagger/dagger/core/integration.mountedPrivateRepoSocket()
      /app/core/integration/module_test.go:6927 +0x2f
  github.com/dagger/dagger/core/integration.ModuleSuite.TestConfigs.func2.2.3.1()
      /app/core/integration/module_config_test.go:220 +0x79
  github.com/dagger/dagger/core/integration.ModuleSuite.TestConfigs.func2.2.3.(*T).Run.2()
      /app/testctx/testctx.go:163 +0xfe
  testing.tRunner()
      /usr/lib/go/src/testing/testing.go:1689 +0x21e
  testing.(*T).Run.gowrap1()
[...]

PR #8196, opened 2024-08-21
Revert "tests: try killing subprocesses harder if they misbehave"

sipsma:
#8191 may be causing a lot of failures on main and PRs based off main. Not 100% sure yet but they seem to have started with that PR being merged. Confirming by trying a revert here. Example failures:
1. https://dagger.cloud/dagger/traces/3c464d3dda1f22988df596b9595e9952
2. https://dagger.cloud/dagger/traces/05d1adca2d582b9fc1683daeefbf2f48

PR #8233, opened 2024-08-26
chore(test): fix race on setupPrivateSSHKey

grouville:
Fixes https://github.com/dagger/dagger/issues/8204 and reintroduces #8191

This PR addresses a data race where goroutines were potentially accessing testing.T after test completion. The solution implements a controlled shutdown mechanism using context cancellation and a WaitGroup.

Key changes:
- Introduce context for cancellation signaling
- Use sync.WaitGroup to track goroutines
- Implement graceful shutdown in cleanup function

This resolves the race condition by:
1. Ensuring all goroutines complete before test finalization
2. Providing immediate shutdown signaling
3. Guaranteeing proper resource release order

The result is a thread-safe setup process that aligns with the test lifecycle, preventing any out-of-bounds operations.

### DIFF
--- a/core/integration/module_config_test.go
+++ b/core/integration/module_config_test.go
@@ -126,7 +126,7 @@ func (ModuleSuite) TestConfigs(ctx context.Context, t *testctx.T) {
 			testOnMultipleVCS(t, func(ctx context.Context, t *testctx.T, tc vcsTestCase) {
 				t.Run("git", func(ctx context.Context, t *testctx.T) {
 					mountedSocket, cleanup := mountedPrivateRepoSocket(c, t)
-					defer cleanup()
+					t.Cleanup(cleanup)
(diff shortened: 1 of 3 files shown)


grouville:
Unable to understand the root cause of the failure, SGTM on my end

jedevc (approved):
Yay nice! Fingers crossed no more weird race behavior comes back :tada:

gerhard:
@jedevc @grouville I noticed that we changed this revert when this got merged:
- https://github.com/dagger/dagger/pull/8196

I don't think that was intentional. I just put it back part of:
- https://github.com/dagger/dagger/pull/8212

Once that gets merged, the revert will be reapplied. That's what I **think** we still want, but sharing here in case something changed since. cc @sipsma

Issue #8206, opened 2024-08-22
`telemetry.Close` now requires a `context.Context`


jedevc:
See discord conversation: https://discord.com/channels/707636530424053791/1003718839739105300/1275579407528099894

> Hmm the error looks like we changed the telemetry API in a non backwards compat way in the otel refactoring recently
> I guess we should have had module compat for this? But we've not had this case before, I hadn't really thought of that telemetry API as "stable" in that way before so missed this

In #7996, the signature of `telemetry.Close` was changed - however, this broke the bass SDK which was calling this library. Previously, we've not thought about `querybuilder` and `telemetry` as being particularly stable interfaces, but we should probably change that.

In the long-term, we should try and have a proper compat layer here (either using templates, or a collection of separately versioned components). But! In the meantime, we should restore `telemetry.Close` to it's old `Context`less form, and either remove the need for it, or have a `CloseWithContext` equivalent.

PR #8234, opened 2024-08-26
fix OTel `sdk/log` and `telemetry.Close` breaking changes

vito:
* Pin `sdk/go` + module's `go.opentelemetry.io/otel/sdk/log` dependency to v0.3.0 (not a bump) to avoid `OnEmit` breaking API change (`log.Record` -> `*log.Record`).
  * Alternatively we could bump the version but that's more work and doesn't fix the root problem - the logging SDK is pre-1.0, so we can expect turbulence like this. Despite the version in `go.mod` it seems that it's too easy to stray ahead to v0.5.0 which breaks everything. Pinning with a `replace` rule seems to be the solution until they ship v1.0.
* Undoes the `telemetry.Close` breaking change from #7996 
  * Not actually necessary, so not worth the breakage.
  * Fixes https://github.com/dagger/dagger/issues/8206

### DIFF
--- a/sdk/go/go.mod
+++ b/sdk/go/go.mod
@@ -30,7 +30,12 @@ require (
 	google.golang.org/grpc v1.64.0
 )
 
-replace go.opentelemetry.io/otel/sdk/log => go.opentelemetry.io/otel/sdk/log v0.3.0
+replace (
+	go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc => go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc v0.0.0-20240518090000-14441aefdf88
+	go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp => go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp v0.3.0
+	go.opentelemetry.io/otel/log => go.opentelemetry.io/otel/log v0.3.0
+	go.opentelemetry.io/otel/sdk/log => go.opentelemetry.io/otel/sdk/log v0.3.0
+)
 
 require (
 	github.com/cenkalti/backoff/v4 v4.3.0 // indirect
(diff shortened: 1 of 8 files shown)


sdk/go/go.mod
```diff
@@ -30,6 +30,8 @@ require (
 	google.golang.org/grpc v1.64.0
 )
 
+replace go.opentelemetry.io/otel/sdk/log => go.opentelemetry.io/otel/sdk/log v0.3.0
```
jedevc, line 33: :cry:

While we're doing this, should we also pin any *other* OTEL deps? Or do you think it's fine to keep this as is?

jedevc (approved):

sdk/go/go.mod
vito: This one is needed because it's pre-1.0 and they're still working out the API.

Theoretically this is a problem for any Go SDK + modules dependency that makes a breaking change without a major version bump. But if we expand the scope to include all pre-1.0 modules, suddenly that includes quite a few things (`Khan/genqlient`, `99designs/gqlgen`, `golang.org/x/sync`, `golang.org/x/exp`).

Arguably this is another crufty symptom of carrying around the source code for `querybuilder` and `telemetry` - pretty sure if they were external packages they could just have normal dependencies that would always be respected.

Or maybe we could just give each package its own `go.mod`? 🤔 

In the short term we should at least pin the rest of the OTel SDK logging related packages since they're most likely to break again, so I'll do that.

jedevc (approved):

sdk/go/go.mod
vito: That's true, but I think we don't need to worry so much about that, because `dagger.io/dagger` will have its own set of dependencies in its own `go.mod`, and those should still be respected - we don't have to worry about the `dagger.io/dagger` dependencies getting bumped out-of-band.

At least, that's my understanding of how Go module dependencies work. The Dagger module case is unique because it vendors the relevant telemetry code directly, so it's subject to the module's own `go.mod`. The `replace` rules are just in `sdk/go/go.mod` so we can have it all in one place and carry them over to each module's `go.mod` and overrule any accidental bumps.

PR #8236, opened 2024-08-26
feat: optimize TS SDK runtime

TomChv:
Reorder steps to not list all entries when configuring modules. Move install dependencies step & corepack init before mounting sources.
Add template after initializing dependencies.

It seems that it really improve the cache hit, putting the TS SDK near Go speed when cache hits.

I updated the sources between the 2 runs and we can see that initialize kept being quick! We need more tests but it looks really promising.

Screenshot 2024-08-27 at 00 58 24

### DIFF
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -101,7 +101,7 @@ func (t *TypescriptSdk) ModuleRuntime(ctx context.Context, modSource *dagger.Mod
 		return nil, fmt.Errorf("failed to analyze module config: %w", err)
 	}
 
-	ctr, err := t.CodegenBase(ctx, modSource, introspectionJSON)
+	ctr, err := t.CodegenBase(ctx, modSource, introspectionJSON, true)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create codegen base: %w", err)
 	}
@@ -111,11 +111,6 @@ func (t *TypescriptSdk) ModuleRuntime(ctx context.Context, modSource *dagger.Mod
 		t.moduleConfig.entrypointPath(),
 		ctr.Directory("/opt/module/bin").File(EntrypointExecutableFile),
 	)
-
-	ctr, err = t.installDependencies(ctr)
-	if err != nil {
-		return nil, fmt.Errorf("failed to install dependencies: %w", err)
-	}
 
 	switch t.moduleConfig.runtime {
 	case Bun:
@@ -140,8 +135,8 @@ func (t *TypescriptSdk) Codegen(ctx context.Context, modSource *dagger.ModuleSou
 		return nil, fmt.Errorf("failed to analyze module config: %w", err)
 	}
 
-	// Get base container
-	ctr, err := t.CodegenBase(ctx, modSource, introspectionJSON)
+	// Get base container without dependencies installed.
+	ctr, err := t.CodegenBase(ctx, modSource, introspectionJSON, false)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create codegen base: %w", err)
 	}
@@ -170,7 +165,7 @@ func (t *TypescriptSdk) Codegen(ctx context.Context, modSource *dagger.ModuleSou
 
 // CodegenBase returns a Container containing the SDK from the engine container
 // and the user's code with a generated API based on what he did.
-func (t *TypescriptSdk) CodegenBase(ctx context.Context, modSource *dagger.ModuleSource, introspectionJSON *dagger.File) (*dagger.Container, error) {
+func (t *TypescriptSdk) CodegenBase(ctx context.Context, modSource *dagger.ModuleSource, introspectionJSON *dagger.File, installDep bool) (*dagger.Container, error) {
 	base, err := t.Base()
 	if err != nil {
 		return nil, fmt.Errorf("failed to create codegen base: %w", err)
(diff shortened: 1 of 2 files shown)


TomChv:
After more small tests on cloud, it looks much faster and I see that the dep install step is cached:

Screenshot 2024-08-27 at 01 05 14

Some run example:

- [`container-echo --string-arg foo`](https://dagger.cloud/Quartz/traces/354c40f0cf12823a460cf2c5457eddee): 6.2s (first run)
- [`container-echo --string-arg bar`](https://dagger.cloud/Quartz/traces/9719b0ce1500bc03ef0315e465cb1a93): 4.9s  (no edit, just arg changed)
- [`foo`](https://dagger.cloud/Quartz/traces/02d4b16df01b468e60774bdaa6714b16): new function created (14.3s), I see that no dependencies is downloaded, most time is spent to copy files
- [`foo`](https://dagger.cloud/Quartz/traces/16f6f7983a08d9f13e2df911cf61d372): 12.4s, re executed, no edit, again most step is about coping files

I wonder if I can speedup the process by using `WithMountedDirectory` instead of `WithDirectory` but when I try to use that, files are missing so I don't know what I'm missing :/

gerhard:
Keen to have https://github.com/dagger/dagger/pull/8250 merged, and this rebased on top of it, so that:
- We get a baseline reading
- We are able to see how this improves on that baseline

TomChv:
It seems i'm hitting an issue with some tests, I'm digging to understand what's going on:
- https://dagger.cloud/dagger/traces/4004298681fa29da6a7abc811aa9bb65
- https://dagger.cloud/dagger/traces/51d6cb0c6498df91a5dc567406942bfc

TomChv:
Ok issue fixed, the `sdk` dir was overwritten when adding source file because I forget to exclude it.
Tests should pass now

sdk/typescript/runtime/main.go
```diff
@@ -184,29 +179,57 @@ func (t *TypescriptSdk) CodegenBase(ctx context.Context, modSource *dagger.Modul
...
 		WithMountedDirectory("/opt/module", dag.CurrentModule().Source().Directory(".")).
-		// Mount users' module with SDK sources and generated client in it.
-		WithMountedDirectory(ModSourceDirPath,
+		// Mount user's module configuration (without sources) and the generated client in it.
+		WithDirectory(ModSourceDirPath,
 			dag.Directory().WithDirectory("/", modSource.ContextDirectory(), dagger.DirectoryWithDirectoryOpts{
 				Include: []string{
-					fmt.Sprintf("%s/**", t.moduleConfig.subPath),
+					fmt.Sprintf("%s/package.json", t.moduleConfig.subPath),
+					fmt.Sprintf("%s/*lock*", t.moduleConfig.subPath),
```
aluzzardi, line 187: Isn't this too broad? e.g. this will upload things like `clock.ts`, `block.js`, `Flock`.

sdk/typescript/runtime/main.go
```diff
@@ -184,29 +179,57 @@ func (t *TypescriptSdk) CodegenBase(ctx context.Context, modSource *dagger.Modul
...
+		}
+	}
+
+	// Add user's source files
+	base = base.WithDirectory(ModSourceDirPath,
+		dag.Directory().WithDirectory("/", modSource.ContextDirectory(), dagger.DirectoryWithDirectoryOpts{
+			// Include the rest of the user's module except config files to not override previous steps & SDKs.
+			Exclude: []string{
+				fmt.Sprintf("%s/package.json", t.moduleConfig.subPath),
+				fmt.Sprintf("%s/*lock*", t.moduleConfig.subPath),
```
aluzzardi, line 218: same comment.

Also, is this `exclude` really needed? Those files are very small and should 100% match still at this point right?

sdk/typescript/runtime/main.go
```diff
@@ -253,7 +276,41 @@ func (t *TypescriptSdk) Base() (*dagger.Container, error) {
 	}
 }
 
-// setupModule initialiaze the user's module.
+// addTemplate adds the template files to the user's module if there's no
+// source files in the src directory.
+func (t *TypescriptSdk) addTemplate(ctx context.Context, ctr *dagger.Container) (*dagger.Container, error) {
```
aluzzardi, line 281: I don't understand why we're doing this?

Shouldn't `dagger init` have done this -- and if not, just fail at this point?

sdk/typescript/runtime/main.go
TomChv: True, I can make it more precise

sdk/typescript/runtime/main.go
TomChv: Yes, to not overwritten potential update on package.json and lockfiles

sdk/typescript/runtime/main.go
TomChv: Dagger init is calling `Codegen`

sdk/typescript/runtime/main.go
TomChv: Done in [52fd637](https://github.com/dagger/dagger/pull/8236/commits/52fd6376526ba1e9fdec6e1d0c307a48cf370754)

sdk/typescript/runtime/main.go
TomChv: Done in [52fd637](https://github.com/dagger/dagger/pull/8236/commits/52fd6376526ba1e9fdec6e1d0c307a48cf370754)

sdk/typescript/runtime/main.go
jedevc: See #8201 for how I want to handle this in the future - but yes, atm, there's no explicit init step, it needs to all get done in `Codegen` (which leads to all sorts of bugs, hence #8201).

sdk/typescript/runtime/main.go
```diff
@@ -535,10 +581,13 @@ func (t *TypescriptSdk) analyzeModuleConfig(ctx context.Context, modSource *dagg
 	}
 
 	// If a first init, there will be no directory, so we ignore the error here.
+	// We also only include package.json & lockfiles to benefit from caching.
 	t.moduleConfig.source = modSource.ContextDirectory().Directory(t.moduleConfig.subPath)
-	entries, err := t.moduleConfig.source.Entries(ctx)
+	configEntries, err := dag.Directory().WithDirectory(".", t.moduleConfig.source, dagger.DirectoryWithDirectoryOpts{
+		Include: []string{"package.json", "*lock*", "tsconfig.json", "pnpm-workspace.yaml", ".yarnrc.yml"},
```
jedevc, line 587: Isn't this also too broad? Same as in https://github.com/dagger/dagger/pull/8236#discussion_r1735072725

jedevc (approved):
One comment, otherwise LGTM, we should get this in soon.

sdk/typescript/runtime/main.go
TomChv: You're right! I'll do the update

TomChv: Fixed in [4236a3f](https://github.com/dagger/dagger/pull/8236/commits/4236a3fa44a24681968da6b4530726cd1aac19ab)

PR #8241, opened 2024-08-27
chore: generate CI config & use context dir

shykes:
This is a small first step towards eventually generating 100% of our Github Actions config.

- Docs
- SDK checks: test, lint, test-publish (on stable and dev engine)


.github/workflows/elixir-sdk-lint-dev-engine.yml
jedevc, line 1: We could add all these files into a `.gitattributes` in this folder, that way changes to them won't show up as massive diffs, etc :tada:

jedevc, line 1: Would it possible to *group* multiple dagger commands into a single file? e.g. I think it would be really nice to have all the `elixir` checks in one single pipeline, but potentially all different jobs, or maybe even steps in a single job?

Not 100% sure, I don't think we need to be optimizing to have a pretty view in github. But mainly, I think it would help avoid spinning up too many new CI workers (which can be pricey).

.github/main.go
```diff
@@ -0,0 +1,167 @@
...
+		OnPullRequestReopened:       true,
+		OnPullRequestSynchronize:    true,
+		OnPullRequestReadyForReview: true,
+		PullRequestConcurrency:      "preempt",
+	})
+	command = fmt.Sprintf("--source=.:default --docker-cfg=file:$HOME/.docker/config.json %s", command)
+	return gh.withPipeline(name, command, opts...)
+}
+
+func (gh *Github) withSdkPipelines(sdk string) *Github {
```
jedevc, line 52: This is *so* neat. I really really really like this.

.github/main.go
```diff
@@ -0,0 +1,167 @@
...
+		withSdkPipelines("java").
+		withSdkPipelines("elixir").
+		withSdkPipelines("rust").
+		withSdkPipelines("php")
+}
+
+func (gh *Github) withSimplePipeline(name, command string, opts ...dagger.GhaWithPipelineOpts) *Github {
+	opts = append(opts, dagger.GhaWithPipelineOpts{
+		OnPushBranches:              []string{"main"},
+		OnPullRequestOpened:         true,
```
jedevc, line 42: Potential interface idea - instead of needing to have a bunch of booleans here, we could maybe have a custom enum of event types in the `gha` module? Then we just pass a list of `Events []GhaEvent` in (which also lets you more easily programmatically control these).

.github/main.go
```diff
@@ -0,0 +1,167 @@
...
+	// +optional
+	dind bool,
+) string {
+	return gh.Runner(2, daggerVersion, 32, true, true, dind)
+}
+
+// Generate Github Actions pipelines to call our Dagger pipelines
+func (gh *Github) Generate() *dagger.Directory {
+	return gh.CI.Config().Directory(".github")
+}
```
jedevc, line 167: A little ask - could we also use the `dirdiff` module to implement a `lint` action here as well (and *maybe* have a pre-defined workflow to make sure this in sync?).

I wish there was a way to avoid needing to have these generated files in the repo at all, I might have some fun ideas for that (hackathon project?), but they involve way too much hackery at the github actions level for now.

jedevc (commented):
My only major complaint with this PR is that I didn't get to do it :laughing: Seriously, it's such a neat approach, I'm a fan :folding_hand_fan: 

This isn't a real review, just was reading through and left some thoughts about some future features/fixes - none of them are *really* blockers. The only more major point is the one around the fact that I *think* this increases our CI cost by spinning up more workers (but I left a suggestion as to what we might do about that).

I think we should try and get this in... relatively soon? Maybe sometime next week works? I think starting with just the `docs` and `sdk` steps is good, since those are "simpler", but also, means that if we turn out to not like the approach (very doubtful in my mind), we can more easily backtrack.

.github/workflows/docs.yml
```diff
@@ -1,29 +1,246 @@
...
+
+                # Detect if a dev engine is available, if so: use that
+                # We don't rely on PATH because the GHA runner messes with that
+                if [[ -n "$_EXPERIMENTAL_DAGGER_CLI_BIN" ]]; then
+                    export PATH=$(dirname "$_EXPERIMENTAL_DAGGER_CLI_BIN"):$PATH
+                fi
+
+                GITHUB_OUTPUT="${GITHUB_OUTPUT:=github-output.txt}"
+                GITHUB_STEP_SUMMARY="${GITHUB_STEP_SUMMARY:=github-summary.md}"
+                export NO_COLOR="${NO_COLOR:=1}" # Disable colors in dagger logs
```
jedevc, line 83: Color? :pleading_face:

.github/main.go
shykes: I actually implemented this already 😁 kudos to @kpenfound for helping figure it out. Unfortunately it's not that great, the best gha can do is generate the files and commit/push them for you. Unlike other CIs it doesn't support continuation.

jedevc: Ahhhh, yeah that's a neat approach. That's annoying it's not great, I think we can just stick with manual invocation for now :cry: (definitely not worth blocking on)

.github/workflows/elixir-sdk-lint-dev-engine.yml
shykes: Good idea. I guess we need to match the generated files on something, any suggestion for the pattern?

shykes: It's possible but requires some extra implementation work, to "collapse" pipelines that can be collapsed. Either automatically or with a "hint" from the caller. I would love a PR on the gha module with that feature :)

.github/main.go
shykes: Yes that's a great idea. My only concern is: what do we do about the `onXYZ` arguments that are not boolean? Like `onPullRequestBranch` etc. How do we avoid making the split feel weird to callers?

.github/main.go
shykes: I think this gap is a possible insertion point for bridging to pocketCI :) But I don't know exactly how. cc @kpenfound @matipan

.github/workflows/docs.yml
shykes: I needed to disable color to avoid polluting my markdown report - did you see that feature?

I realize now it's a tradeoff between 1) neat logs in markdown report, and 2) colors in GHA logs

.github/main.go
shykes: @jedevc I added a `Check()` function that uses `dirdiff`. But I wasn't sure where to integrate it in the main module, so I didn't wire it in. I could use your guidance on that.

gerhard (approved):
I like the idea of using Dagger to generate GitHub Actions YAML. I am not excited about removing `180` lines of YAML, and adding `11k` lines of generated YAML.

Some of the new scripting is significantly more complex than what we had before. While `action.yml` would not win any scripting competitions, it was simpler.

The only real blocker that I see to merging this is losing the `timeout` setting. Restricting `GITHUB_TOKEN` permissions would be nice to keep, but less of a blocker.

Overall, I am curious to see how well this works in practice. Specifically:
- When a job fails, is it easier to understand which step failed & why?
- How well does the workflow dispatch work after the test & lint split?
- What impact does this split have on scheduling jobs? We will need 2x as many runners.
- What's it like to live with 2 modules, `.dagger` and `.github` ?

image

One thing which I know will hurt is when we hit GitHub's `100` checks limit. After this change, we are at `91`. GitHub's UI makes it impossible to make sense of more than 100 checks. I don't think the number goes higher than 100 (this limitation goes back to 2021), and the small scrolling area makes it very difficult to find anything. Also, the `Checks` tab is getting more difficult to use:

image

I am going to approve this optimistically so that we can merge this (after resolving the conflicts) and try living with this change for a few weeks. YOLO!

.github/workflows/elixir-sdk-lint-dev-engine.yml
shykes: @jedevc thinking about this more:

> I think it would be really nice to have all the elixir checks in one single pipeline, but potentially all different jobs

This is possible (with some implementation work) BUT it will NOT change the number of machines started, because each job gets its own machine, regardless of workflow grouping.

> or maybe even steps in a single job?

There is no benefit to this. It should be done at the Dagger level.

.github/main.go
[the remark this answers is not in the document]
shykes: Good catch. I actually implemented `timeout` already, I just forgot to use it in the code :) Will fix now

.github/workflows/docs.yml
```diff
@@ -1,29 +1,246 @@
...
-    types:
-      - opened
-      - synchronize
-      - reopened
-      - ready_for_review
-  # Enable manual trigger for easy debugging
-  workflow_dispatch:
-
-permissions:
-  contents: read
```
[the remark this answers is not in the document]
shykes: Makes sense. Will do.

.github/main.go
```diff
@@ -0,0 +1,198 @@
...
+
+// Platinum runner: Single-tenant on-demand instance, 32 cpu
+func (ci *CI) PlatinumRunner(
+	// Enable docker-in-docker
+	// +optional
+	dind bool,
+) string {
+	return ci.Runner(2, daggerVersion, 32, true, true, dind)
+}
```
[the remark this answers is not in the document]
shykes: I considered that too, but that implies a quantitative increase (like memory, cpus) and some of the dimensions are more subtle - multi-tenant/single-tenant, spot/nospot etc. This felt safer and more flexible for you.

.github/main.go
```diff
@@ -0,0 +1,198 @@
...
+	return ci.WithPipeline(name, strings.Join(command, " "), runner, devEngine)
+}
+
+// Assemble a runner name for a pipeline
+func (ci *CI) Runner(
+	generation int,
+	daggerVersion string,
+	cpus int,
+	singleTenant bool,
+	noSpot bool,
```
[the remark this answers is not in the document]
shykes: It's because "spot instance" became a generic term for a well-known pattern beyond AWS (instances that can be preempted depending on demand and the price you set). Whereas "on demand" is AWS-specific jargon that doesn't carry the same obvious meaning industry-wide.

So `noSpot` will be more obvious to more people.

shykes:
Thanks Gerhard! A few notes below.

> I like the idea of using Dagger to generate GitHub Actions YAML. I am not excited about removing `180` lines of YAML, and adding `11k` lines of generated YAML.

A more fair comparison is that I replaced 180 lines of YAML with 196 lines of Go, which can collapse to a `13` line overview:

```golang
ci.Gha = dag.Gha(dagger.GhaOpts{
		DaggerVersion: daggerVersion,
		PublicToken:   "dag_dagger_sBIv6DsjNerWvTqt2bSFeigBUqWxp9bhh3ONSSgeFnw",
		Runner:        ci.BronzeRunner(false),
	})
	return ci.
		WithPipeline("Docs", "docs lint", "", false).
		WithSdkPipelines("python").
		WithSdkPipelines("typescript").
		WithSdkPipelines("go").
		WithSdkPipelines("java").
		WithSdkPipelines("elixir").
		WithSdkPipelines("rust").
		WithSdkPipelines("php")
```

I expect the ratio will only get better as switch more pipelines to be generated.

This would be nicer if Github Actions supported continuations - then the generation step could be hidden better. We'll start with proper `.gitattributes`. But certainly, to appreciate this change, you need to adopt a mindset where the generated YAML is not source code: it's like a compiled binary.

> Some of the new scripting is significantly more complex than what we had before. While `action.yml` would not win any scripting competitions, it was simpler.

It was not actually simpler. The perceived increase in complexity comes from several things:

1. Additional features. In particular the markdown report requires a larger shell script. The `action.yaml` doesn't have that feature, so it looks simpler.
2. Inlining. Most of the generated shell script is actually *identical* to the manually edited one, or only slightly tweaked. But it's all shown in one place, instead of spread out across an action YAML, a workflow YAML, and in some cases one or more intermediary workflow YAML.
3. Simpler YAML. I transferred some of the complexity from proprietary YAML to Bash. This makes the bash script look slightly more complicated (not that much really). But in return we get less context switching between YAML and bash, and more local debugability. For example you can debug the markdown generation feature with `COMMAND='echo hello world' ./scripts/exec.sh; less github-summary.md`. Same for all the other steps.

Some of the above ☝️ are not specific to generating the YAML. For example, I *could* generate a workflow that references a shared static `action.yaml`, or intermediary workflow. I just chose not to, because I like this tradeoff. Operationally one single self-contained YAML, with everything inlined and no fancy runtime abstractions, is ultimately more reliable. But, even if that turns out to be wrong, we can keep the generator model and just change the layout that we generate to. We could even support several layouts, and make it configurable.

> The only real blocker that I see to merging this is losing the `timeout` setting. Restricting `GITHUB_TOKEN` permissions would be nice to keep, but less of a blocker.

Thanks for catching that. I'm on it.

> Overall, I am curious to see how well this works in practice. Specifically:
> 
> * When a job fails, is it easier to understand which step failed & why?

So far it seems the be either the same or better, depending on the situation.

- The SDK split is nice
- More standardized layout across workflows is a subtle advantage. I know what to expect across all checks, it's the same structure, so less cognitive load when debugging
- The markdown summary is very useful, I often never go to the logs at all, I go straight to the summary (sadly github UI makes that 2 clicks from the PR)

> * How well does the workflow dispatch work after the test & lint split?

Not sure what you mean by that.

> * What impact does this split have on scheduling jobs? We will need 2x as many runners.

I don't think so. My understanding is that machines are allocated at the job level, not the workflow level. This is one reason I went for one pipeline per workflow file. It didn't seem to impact the runtime behavior that much.

> * What's it like to live with 2 modules, `.dagger` and `.github` ?

I am not attached to 2 modules particularly, but it seems most aligned with the current consensus on how to tackle code generation pipelines with Dagger. That is totally something we can change.

> One thing which I know will hurt is when we hit GitHub's `100` checks limit. After this change, we are at `91`. GitHub's UI makes it impossible to make sense of more than 100 checks. I don't think the number goes higher than 100 (this limitation goes back to 2021), and the small scrolling area makes it very difficult to find anything. Also, the `Checks` tab is getting more difficult to use:

I actually think we can bring the number of checks down easily. I didn't mind the increased number of checks (if anything I started to look at the Dagger Cloud checks less, because I get the same granularity from GHA checks now). But I can prioritize bringing the number the checks down.

The easiest way to do this, is to collapse pipelines in Dagger. Lint, test and test-publish for each SDK can be collapsed into a single dagger call. Splitting that across 3 VMs per SDK is overkill anyway.

Separately, we should talk about what this model means for Dagger Cloud checks. Do we still need that granularity in a world where generated CI config is the norm, giving you the granularity for free? Maybe we can switch to a single Dagger Cloud check that links to the entire trace group? cc @vito @aluzzardi @kpenfound @crjm 

> I am going to approve this optimistically so that we can merge this (after resolving the conflicts) and try living with this change for a few weeks. YOLO!

Thank you, I appreciate the detailed review!

.github/workflows/elixir-sdk-lint-dev-engine.yml
shykes: Done

.github/main.go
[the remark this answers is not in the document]
shykes: Fixed!

.github/workflows/docs.yml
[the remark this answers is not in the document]
shykes: @gerhard Done :) https://github.com/dagger/dagger/pull/8241/commits/c78537f1404c6b6f3460f29977c2caad95e22f4e

.github/main.go
jedevc: Hm, I'd add a new github workflow for right now for this - something like `github lint`? I suppose wiring it into the main module makes sense, but we could also just have a single workflow that directly calls the github module, I'm not hugely biased towards either.

.github/main.go
jedevc: So what I can tell - the event block here is made up of two things - the thing that can happen to the branch, and the branch target.

So *actually* it's a bit trickier than an enum - it's almost like we want an event to be:

```go
type GithubEvent struct {
	PullRequest *GithubPullRequestEvent
	Issue *IssueEvent
	... // more events
}

type GithubPullRequestEvent struct {
	Branches []string
	Triggers []GithubPullRequestTrigger
}

type GithubPullRequestTrigger string
const (
	GithubPullRequestOpened = "opened"
	// ...
)
```

Then, you just pass in a list of `GithubEvent` to your `WithPipeline` - maybe constructing them something like:

```
event := dag.OnPullRequest().WithBranch("main").OnOpen().OnClose()
```

How does something like that sound? Feels very literate-style, but then each event is it's own dagger object to manipulate (which would be *really* cool, now you could build little functions to encapsulate the logic for your project).

.github/workflows/docs.yml
jedevc: Mm, I see. Yeah, that's alright for now I think.

Here's an idea: we could use something like `ansi2txt` from https://github.com/kilobyte/colorized-logs to take the colored output and remove it for the markdown report.

(I can't see where that feature is btw, might be missing something).

.github/workflows/elixir-sdk-lint-dev-engine.yml
jedevc: > This is possible (with some implementation work) BUT it will NOT change the number of machines started, because each job gets its own machine, regardless of workflow grouping.

Yeah, fair enough - I guess it's more just for sorting at the github level. I don't *currently* have the time to PR this in (reviews + flakes + prep for v0.13 are most of my attention this week) - but can try and find a moment after that's landed if no one else can pick it up.

> There is no benefit to this. It should be done at the Dagger level.

Fair. I think I agree - this does mean that if we want lots of tasks done, you need to use your language primitives for it. This is good.

But, I think if we're doing this more, the performance of the traces view in dagger cloud might need some help - when you have a *lot* of spans together, the performance can really suffer, and the page takes way more time to load than the basic github actions logs view. But hopefully we can feel the pain of this and do something here.

.github/dagger.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "ci",
+  "sdk": "go",
+  "dependencies": [
+    {
+      "name": "gha",
+      "source": "github.com/shykes/gha@b67ffd088b9411c8a01f5d6ca7b83365e3d38484"
```
jedevc, line 7: What's the plan for where this lives long-term btw? Could we move it to the `dagger` org or under `modules/`? Not a blocker at all, just curious, personally, I prefer having it "semi-official", with permissions/reviews managed by the same system we use would be good.

.github/main.go
```diff
@@ -0,0 +1,196 @@
...
+	ci := new(CI)
+	ci.Gha = dag.Gha(dagger.GhaOpts{
+		DaggerVersion: daggerVersion,
+		PublicToken:   "dag_dagger_sBIv6DsjNerWvTqt2bSFeigBUqWxp9bhh3ONSSgeFnw",
+		Runner:        ci.BronzeRunner(false),
+		Repository:    repository,
+	})
+	return ci.
+		WithPipeline("Docs", "docs lint", "", false).
+		WithSdkPipelines("python").
```
matipan, line 38: Thinking a bit how this can be translated to the idea of self contained functions in the context of pocketci.

Each of the pipelines described here can be treated as its own self contained function that defines when it should be triggered and where it should run. At the moment we are separating this even further (although I know you already think its overkill) since each step of SDK jobs (lint/test/publish) is its own pipeline with a separate dagger call. If we could bring everything into one we could take an approach that would yield numerous dagger functions (this is still experimental):
```go
// Function only gets called when there is a pull request (opened,reopened,synchronize,ready_for_review) or a push event (to branch main).
func (dev *DaggerDev) SdkGoOnGithub(ctx context.Context,
	src *dagger.Directory,
	// +default="pull_request,push"
	event string,
	// +default="opened,reopened,synchronize,ready_for_review,main"
	filter string,
	eventTrigger *dagger.File) error {
	sdk := dev.SDK().Go

	if err := sdk.Lint(ctx); err != nil {
		return err
	}

	return sdk.Test(ctx)
}
```

Each SDK would have its own function with virtually with the same trigger. This is definitely more verbose than what we currently have (and we still have not solved the compute problem). I'm also not liking so much how the filtering of events and actions within an event is being built, there is some more thinking I need to do there. Its worth doing a bit more thinking and playing with dagger/dagger more to see how quickly the idea of functions defining its own trigger logic breaks.

.github/workflows/docs.yml
jedevc: Ah, I found it. FYI, there's still some control color codes in there sometimes: https://github.com/dagger/dagger/actions/runs/10648614702?pr=8241

.github/dagger.json
shykes: Yes, we should move it in. There are a few options, I will open a follow-up proposal.

.github/main.go
shykes: This won't work IMO. You need to keep the CI-specific logic (event routing etc) decoupled, at a higher layer.

.github/workflows/docs.yml
shykes: Could it be because of nested dagger engines?

shykes:
@matipan @gerhard @jedevc I have `elixir test` and `engine test specific` failing, are those flakes?

jedevc:
Hm, no the failing elixir job looks genuine, never seen this failure before. cc @wingyplus who might have some ideas.

It looks like:
- We see a bunch of `ERR failed to emit telemetry error="traces export: failed to send to http://127.0.0.1:39707/v1/traces: 500 Internal Server Error`
- And also get a test failure?

`engine test specific` failing looks like a known flake to me.

---

One unrelated note: the new checks names have duplication in them :thinking: 

image

Ideally the job grouping would remedy some of that, but in the meantime, could we have something like `elixir sdk: test / check` with `check` (or similar) being the job name?

wingyplus:
> Hm, no the failing elixir job looks genuine, never seen this failure before. cc @wingyplus who might have some ideas.
> 
> It looks like:
> 
> * We see a bunch of `ERR failed to emit telemetry error="traces export: failed to send to http://127.0.0.1:39707/v1/traces: 500 Internal Server Error`
> * And also get a test failure?
> 
> `engine test specific` failing looks like a known flake to me.
> 
> One unrelated (non-blocker) note: the new checks names have duplication in them 🤔
> 
> image
> 
> Ideally the job grouping would remedy some of that, but in the meantime, could we have something like `elixir sdk: test / check` with `check` (or similar) being the job name?

Will looking into that problem in around 3-4 hours. ✌️

shykes:
> Ideally the job grouping would remedy some of that, but in the meantime, could we have something like `elixir sdk: test / check` with `check` (or similar) being the job name?

Sadly we cannot. Initially the naming was exactly what you suggested. But the job name is used to match "required checks". The Cubzh team use this feature and after much debugging, we landed on this.

We *could* make that behavior configurable with a flag, although it would be an obscure one.

Can we make this a followup? 🙏

wingyplus:
@jedevc From what I can tell. The test failed on [mounted cache test]( https://github.com/dagger/dagger/blob/main/sdk/elixir/test/dagger/client_test.exs#L125). And from this [failure](https://github.com/dagger/dagger/actions/runs/10648614702/job/29517873407#step:5:3141), it looks like cache volume is stored from the previous run somehow cause it gets double result. I open a PR to fix this at https://github.com/dagger/dagger/pull/8317 by using LOCKED mode but I'm also not sure if it's the right fix.

shykes:
> @jedevc From what I can tell. The test failed on [mounted cache test](https://github.com/dagger/dagger/blob/main/sdk/elixir/test/dagger/client_test.exs#L125). And from this [failure](https://github.com/dagger/dagger/actions/runs/10648614702/job/29517873407#step:5:3141), it looks like cache volume is stored from the previous run somehow cause it gets double result. I open a PR to fix this at #8317 by using LOCKED mode but I'm also not sure if it's the right fix.

Is it possible that my PR caused the issue? Maybe a subtle change in the YAML config?

.github/workflows/elixir-sdk-lint-dev-engine.yml
shykes: > But, I think if we're doing this more, the performance of the traces view in dagger cloud might need some help - when you have a lot of spans together, the performance can really suffer, and the page takes way more time to load than the basic github actions logs view. But hopefully we can feel the pain of this and do something here.

I believe @vito has been focusing specifically on Traces UI performance :)

.github/main.go
shykes: @jedevc this is what I initially want to do, it most closely models the reality of the underlying data structure. But, it wouldn't be usable from the CLI. I am setting a constraint for myself to make the API great in both code and the CLI.

dagger.json
```diff
@@ -48,7 +48,7 @@
     }
   ],
   "source": ".dagger",
-  "engineVersion": "v0.12.7",
+  "engineVersion": "v0.12.6",
```
wingyplus, line 51: Why this version got revert back to 0.12.6?

.github/main.go
jedevc: tangent: is this then an argument to make the CLI more powerful, so it can handle construction of more complex objects, i.e. ones that take other objects as parameters?

cc @helderco

.dagger/main.go
```diff
@@ -74,33 +76,82 @@ func (dev *DaggerDev) WithModCodegen() *DaggerDev {
...
+		base = &ElixirSDK{Dagger: dev}
+	}
+	return func(ctx context.Context) error {
+		if err := base.Lint(ctx); err != nil {
+			return err
+		}
+		if err := base.Test(ctx); err != nil {
+			return err
+		}
+		// FIXME: PublishTest
```
jedevc, line 106: :eyes: just wanted to check, is this being brought back before/after we merge this? This does catch real issues, so don't want to leave it too long.

shykes: Sorry yes, that is WIP, will bring back today

dagger.json
shykes: (no idea how this happened btw)

.github/main.go
shykes: > tangent: is this then an argument to make the CLI more powerful, so it can handle construction of more complex objects, i.e. ones that take other objects as parameters?

Yes :)

.dagger/main.go
shykes: Done.

shykes:
@gerhard in an effort to make you happy, we are now down to 3k lines of YAML instead of 11k :)

shykes:
Force-pushing a small change to re-run all failed workflows...

shykes:
FYI I am combining #8284 into this, it's too much work to maintain both in parallel.

.github/workflows/elixir-dev.gen.yml
```diff
@@ -0,0 +1,248 @@
...
+            - main
+    pull_request:
+        types:
+            - opened
+            - reopened
+            - synchronize
+            - ready_for_review
+    workflow_dispatch: {}
+concurrency:
+    group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
wingyplus, line 15: The `github.workflow` is a workflow name. So then how is it important to keep this because generated workflow has only single job?

shykes: It's to control concurrency of multiple instances of the workflow. Typically if you push 10 times to the same PR, that triggers 10 different instances of the worfklow. With the concurrency group, only 1 instance will run at a time.

wingyplus: Ahh, I see. Thanks for the info.

PR #8245, opened 2024-08-27
fix(python): use unique module paths to avoid uv cache conflict

jedevc:
Attempts to fix https://github.com/dagger/dagger/issues/7662, follow-up to https://github.com/dagger/dagger/pull/8051.

I *think* https://github.com/dagger/dagger/pull/8051 was the right approach - however, the tests weren't the issue. The paths in the tests aren't actually part of the uv cache key, so the underlying issue remains.

The real fix for this needs to be in the python sdk runtime, where we can correctly set these paths - based on the module id (or some other "unique enough" id, that can guarantee that we're not going to get weird horrible conflicts).

cc @sipsma @helderco

### DIFF
--- a/sdk/python/runtime/main.go
+++ b/sdk/python/runtime/main.go
@@ -4,7 +4,9 @@ package main
 
 import (
 	"context"
+	"crypto/sha256"
 	_ "embed"
+	"encoding/hex"
 	"fmt"
 	"path"
 	"python-sdk/internal/dagger"
@@ -84,6 +86,14 @@ type PythonSdk struct {
 	// Resulting container after each composing step
 	Container *dagger.Container
 
+	// SourcePath is a unique host path for the module being loaded
+	//
+	// HACK: this property is computed as a unique value for a ModuleSource to
+	// provide a unique path on the filesystem. This is because the uv cache
+	// uses hashes of source paths - so we need to have something unique, or we
+	// can get very real conflicts in the uv cache.
+	SourcePath string
+
 	// Discovery holds the logic for getting more information from the target module
 	// +private
 	Discovery *Discovery
@@ -95,11 +105,11 @@ func (m *PythonSdk) Codegen(
 	modSource *dagger.ModuleSource,
 	introspectionJSON *dagger.File,
 ) (*dagger.GeneratedCode, error) {
-	self, err := m.Common(ctx, modSource, introspectionJSON)
+	m, err := m.Common(ctx, modSource, introspectionJSON)
 	if err != nil {
 		return nil, err
 	}
-	return dag.GeneratedCode(self.Container.Directory(ModSourceDirPath)).
+	return dag.GeneratedCode(m.Container.Directory(m.SourcePath)).
 		WithVCSGeneratedPaths(
 			[]string{GenDir + "/**"},
 		).
@@ -114,11 +124,11 @@ func (m *PythonSdk) ModuleRuntime(
 	modSource *dagger.ModuleSource,
 	introspectionJSON *dagger.File,
 ) (*dagger.Container, error) {
-	self, err := m.Common(ctx, modSource, introspectionJSON)
+	m, err := m.Common(ctx, modSource, introspectionJSON)
 	if err != nil {
 		return nil, err
 	}
-	return self.
+	return m.
 		WithInstall().
 		Container.
 		WithEntrypoint([]string{RuntimeExecutablePath}), nil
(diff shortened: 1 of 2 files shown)


jedevc:
My theory for how this affected CI more than local environments is related to disk throughput - I *suspect* with heavily busy enough engines, the disk could be maxed out, and we might end up having some read/write operations interleaved. This means that potentially with all these parallel python tests, we *might* end up with some of the calls to `uv` and `python` interleaved weirdly, which may not have happened locally.

sdk/python/runtime/main.go
```diff
@@ -157,6 +169,16 @@ func (m *PythonSdk) Load(ctx context.Context, modSource *dagger.ModuleSource) (*
...
 	}
+
+	id, err := m.Discovery.ModSource.ID(ctx)
+	if err != nil {
+		return nil, err
+	}
+	h := sha256.New()
+	fmt.Fprint(h, id)
+	idHex := hex.EncodeToString(h.Sum(nil))
+	m.SourcePath = path.Join(ModSourceDirPath, idHex)
```
jedevc, line 180: Hm, this isn't the right stable ID to use - it *should* be related to the context directory I think?

For git sources, maybe we could use the commit ID? And for local sources, the `Digest`? Maybe we want a `ModuleSource.digest` property more generally for things like this.

jedevc:
Oh hello, it's you again: `TestModule/TestPythonAltRuntime`:

```
cannot diff with different relative paths: "/" != "/src"
```

sdk/python/runtime/main.go
helderco: > Maybe we want a `ModuleSource.digest` property more generally for things like this.

I've found the need for a unique ID per module to use as a key for a cache volume, for `node_modules` and `.venv`, but was able to avoid it (just not creating a cache volume for it). So I agree that it can be useful in general.

We'll soon have isolated/namespaced cache volumes per module though. Even though that'll apply to the runtime module itself and not the target module, maybe the same approach that @sipsma is going for can be reused here, even if just the part that uniquely identifies a modules. Not sure if it's a simple namespace based on the module's name (like type namespaces) though.

jedevc: Isolated cache volumes for each SDK instance would I think solve this issue, but the problem then is that we wouldn't be able to cache any *legitimately* shared dependencies if they were entirely separated :cry:

I guess the uv cache is just not designed to be shared across multiple hosts, so, we need some little trick just for those paths :cry:

helderco: Yeah, I know, I was saying maybe you can reuse some part of it, in order to get the unique value for the mount. Unless the idea for the namespaced cache valumes is to just use the module name.

jedevc: *Oh* right I see :tada: :tada: (coffeeee time)

sipsma (approved):

sipsma:
Gonna merge this now, let's see if this chunk of flakes is officially squashed! 🤞

jedevc:
Looks like this wasn't merged - I'm gonna hold it till after the release later today.

sdk/python/runtime/main.go
jedevc: I ended up implementing `Directory.digest` and `ModuleSource.digest` so I could have a better way of doing this.

For some reason, the `ID` kept changing, I'm guessing `normalizeCallerLoadedSource` isn't helping here? Having a `Digest` helps keep things a bit more consistent and content-based.

sipsma:
> Looks like this wasn't merged - I'm gonna hold it till after the release later today.

I'm so confused, I swear I pressed merge... 😵‍💫

jedevc:
:scream: okay, all the test failures are either flakes, or... the issue that the linter gets very sad when trying to codegen a feature in `sdk/python/runtime` that hasn't actually been released.

This is annoying, two options:
- Just go back to doing `ID` hashing, but I didn't love that very much
- Exclude `sdk/python/runtime` from the linter

helderco:
> 😱 okay, all the test failures are either flakes, or... the issue that the linter gets very sad when trying to codegen a feature in `sdk/python/runtime` that hasn't actually been released.

Probably would need to release the new digests in the API first, and then use it in the runtime right?

However, I'm good with either options you provided until next release. Excluding from the linter is just easier though, but temporary anyway.

jedevc:
Decided to revert to the older approach, and just deferred adding the `.Digest` API for later.

sipsma (approved):

PR #8250, opened 2024-08-27
Benchmark SDKs

aluzzardi:
Using a freshly built dagger engine, for each of the Go, Python and Typescript SDK, benchmark the following operations:

- `dagger init`
- `dagger functions`
- `dagger functions` (**cached**)
- `dagger functions` (after modifying the source code)

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -32,7 +32,7 @@ jobs:
           mkdir -p /tmp/benchmark
           cd /tmp/benchmark
 
-          dagger --debug init --source=. --name=test --sdk=${{ matrix.sdk }}
+          dagger init --name=benchmark --sdk=${{ matrix.sdk }}
         env:
           DAGGER_CLOUD_TOKEN: "dag_dagger_sBIv6DsjNerWvTqt2bSFeigBUqWxp9bhh3ONSSgeFnw"
 
@@ -42,7 +42,7 @@ jobs:
           source /tmp/local-envs
           cd /tmp/benchmark
 
-          dagger --debug functions
+          dagger functions
         env:
           DAGGER_CLOUD_TOKEN: "dag_dagger_sBIv6DsjNerWvTqt2bSFeigBUqWxp9bhh3ONSSgeFnw"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,131 @@
...
+        run: |
+          echo "::group::Starting dev engine"
+          if ! [[ -x "$(command -v docker)" ]]; then
+            echo "docker is not installed"
+            exit 1
+          fi
+
+          # put env variables in /tmp/local-envs instead of
+          # $GITHUB_ENV to avoid leaking into parent workflow
+          (cd .dagger/mage; go run main.go -w ../.. engine:dev) | tee /tmp/local-envs
```
aluzzardi, line 67: @gerhard Any better of way of doing all of this? I had to copy some bits of the call action

Also -- are those envs path safe? e.g. Is the code below doing a `source` and then `dagger <command>` using the right stuff?

gerhard: I don't think that we need any of this.

We expect this to only run on the `dagger/dagger` repository, which means that we will **always** have access to our CI runners. We can then simplify all of this to use:
- `runs-on: dagger-g2-v0-12-5-4c`
- `runs-on: dagger-g2-main-4c`

That will basically benchmark against the latest released version, and the latest unreleased version. `main` is wired through our entire stack, meaning that the CLI will be automatically installed and made available to the workflow, and the corresponding Engine will be ready to go by the time the workflow starts.

.github/workflows/benchmark.yml
```diff
@@ -17,78 +17,19 @@ concurrency:
...
-          echo "::group::Dagger server version"
-          echo "{version}" | dagger query
-          echo "::endgroup::"
-        env:
-          DAGGER_CLOUD_TOKEN: "dag_dagger_sBIv6DsjNerWvTqt2bSFeigBUqWxp9bhh3ONSSgeFnw"
-
       - name: dagger init (${{ matrix.sdk }})
         shell: bash
         run: |
           source /tmp/local-envs
```
gerhard, line 31: I don't think that we need to `source /tmp/local-envs` - the file no longer exists.

.github/workflows/benchmark.yml
```diff
@@ -17,78 +17,19 @@ concurrency:
 
 jobs:
   benchmark:
```
gerhard, line 19: I think that we should add `if: ${{ github.repository == 'dagger/dagger' }}` so that this will only run in the context of our repository.

gerhard (approved):
OK to merge after the two tiny changes are addressed.

gerhard:
LGTM! 👍 🚀

jedevc (approved):

PR #8256, opened 2024-08-28
chore: upgrade goreleaser

jedevc:
To get go 1.23 in goreleaser, we need to bump goreleaser up to the latest release - there are also breaking changes and deprecations we need to resolve.

This should fix the broken cli and engine publish on main: https://github.com/dagger/dagger/actions/runs/10594823172/job/29359279613


gerhard (approved):
LGTM 👍 🚀

PR #8261, opened 2024-08-28
chore: fix linting

jedevc:
Fixes these issues on `main`: https://github.com/dagger/dagger/actions/runs/10596511181/job/29364633106, which was introduced in #8151.

Some of these lints had stopped working, since the paths had gotten mixed up - this was because we were taking loop references (which is safe in go 1.22 - but requires that all the modules actually *use* go 1.22 in their go.mods).

This fixes that issue by updating go.mods, disables the loop capture checker (since this isn't required anymore in go 1.22), and also fixes the linting issues that we hadn't been catching anymore!


gerhard (approved):
LGTM 👍 🚀

PR #8262, opened 2024-08-28
ci: benchmark: fix runners, change schedule

aluzzardi:
- **ci: benchmark: use different runner**
- **ci: benchmark: change schedule**


gerhard (approved):
👍 🚀

PR #8268, opened 2024-08-29
chore: prep for v0.12.6 release

jedevc:


helm/dagger/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
-appVersion: 0.12.5
```
gerhard, line 2: Do we expect the Helm test to fail when we bump this?

gerhard (approved):
LGTM 👍

Left a Helm test comment, non-blocking.

helm/dagger/Chart.yaml
jedevc: Yeah - I can take a look at this after the release process is done. Essentially what's happening is it's trying to install v0.12.6 into the k3s cluster - but that's not available yet (so it'll keep failing on main until v0.12.6 is actually tagged and released).

The solution here is going to be to install the `:main` tag instead of the versioned tag when testing on `main`, similar to how we already do for the provisioning tests, which is kind of what this is already like.

PR #8269, opened 2024-08-29
ci: build all engine variants before publishing

jedevc:
As discussed during the v0.12.6 release process:

> imo, we can take this opportunity to update this process and daggerize more - ideally we should build all the images, and then push all the images

This consolidates a ton of the publishing logic together into a loop over a table (familiar to anyone familiar with go table tests). Because all of the logic in go, it's now super easy to force evaluation of all the containers before we start pushing them.

Finally, I removed the `TestPublish` function, and replaced it with a `dryRun` parameter in `Publish` (to avoid duplicating a ton of logic) which matches with how we already do this for the SDKs.

Here's an example dry-run: https://dagger.cloud/jedevc/traces/7c283e20d59719ff45c1038d7e74174d

image


gerhard (approved):
This is a great refactoring!

LGTM 👍 🚀

PR #8270, opened 2024-08-29
docs: Consolidate pages on deps and package managers

vikram-dagger:


docs/sidebars.ts
```diff
@@ -386,18 +386,10 @@ module.exports = {
           "type": "doc",
           "id": "manuals/developer/runtimes"
         },
-        {
-          "type": "doc",
-          "id": "manuals/developer/package-managers"
-        },
```
jedevc, line 392: Do we need redirects?

docs/current_docs/manuals/developer/dependencies.mdx
```diff
@@ -0,0 +1,242 @@
...
+
+Here is an example of accessing the installed `helloWorld` module from your own module's code:
+
+<Tabs groupId="language">
+<TabItem value="Go">
+
+```go
+func (m *MyModule) Greeting(ctx context.Context) (string, error) {
+  return dag.HelloWorld().Message(ctx)
+}
```
jedevc, line 49: This doesn't look right :eyes: The module name installed above is `hello`, so it would be accessible at `dag.Hello()`. Same for the other languages.

The correct usage is `dag.Hello().Hello(ctx)` (trying this locally). Confused, this has clearly been wrong for a while.

docs/sidebars.ts
vikram-dagger: I considered it but did not implement because
- this page was added just a couple of days ago so less likely to have been indexed
- once the merged manual is live all the current URLs and existing redirects will become invalid anyway

docs/current_docs/manuals/developer/dependencies.mdx
vikram-dagger: Good catch, thanks! Fixed now.

jedevc (commented):
LGTM to me :tada:

jedevc (approved):

TomChv (approved):
LGTM

Issue #8207, opened 2024-08-22
Cannot find default codegen `DaggerObject` interface


jedevc:
Splitting out this one from https://github.com/dagger/dagger/issues/8031#issuecomment-2260352601 to debug - it's quite clearly a different issue.

The first step towards understanding this was attempted in https://github.com/dagger/dagger/pull/8181, and [now we have a few failure with this](https://dagger.cloud/dagger/traces/a5124f9ebf3ec80cbbc12b6918aea96c?span=162d324564a7d657#162d324564a7d657:L7) in `TestModule/TestGoInit/from_scratch`:

```
generating go module: bare
writing dagger.gen.go
writing go.mod
writing go.sum
running post-command: go mod tidy
needs another pass...

Error: generate code: template: module.go.tmpl:79:3: executing "_dagger.gen.go/module.go.tmpl" at <ModuleMainSrc>: error calling ModuleMainSrc: exected DaggerObject to be *types.Interface, but got *types.Basic
```

This is bizarre, `DaggerObject` is not a basic type, it's an interface - need to investigate more - I suspect that it's an unknown type (which is a basic type for some reason).

Opened #8214 to try and narrow down the possibilities:
- Is is actually an unknown type?
- Do `internal/` packages actually exist at the time of the bad `dagger call`? And do they have the expected contents?
- Are `internal/` files getting weirdly regenerated by go formatting of imports?

Potentially, we're hitting something really weird, and causing some weird race condition in the go build cache? Otherwise, really stretching to imagine what could possibly be going on. 

Note - this doesn't look like something in the dagger engine going wrong, like confusing the caching on different sets of files, we're actually running all of this within a single `codegen` pass.

jedevc:
Hm, just noticed that the places that this flakes in corresponds to where we use the go sdk with `--name=bare` pretty clearly - `TestGit` and `TestGoInit`.

sipsma:
Made some minor progress with this:
1. It took absolutely forever to repro in my debugging PR when I was just re-running every engine test workflow over and over (i.e. re-running all engine tests suites in parallel)
   * **BUT** it seems to repro much easier if I instead push an empty commit, which triggers *every* workflow (including non-engine-test ones)...
   * Aligns with this possibly being some sort of race condition that's more likely to happen when the engine's host is under extra load? Otherwise I don't see how other workflows could be related
[...]

PR #8272, opened 2024-08-29
chore: update x/tools dep

sipsma:
Want us on the latest version of this since it's in the realm of plausibility that a bug in golang.org/x/tools/packages could explain [some of the flakes](https://github.com/dagger/dagger/issues/8207) we've been having.

---

There's a few commits in the diff that have some potential of being related (based on a cursory look and somewhat limited understanding of the go internals/deps):
* https://cs.opensource.google/go/x/tools/+/7cc3be7d11326f7aa5fc7a51166590788049d072
* https://cs.opensource.google/go/x/tools/+/dddd55df4919cb0fab2ab5c30445f358a39da8fb
* https://cs.opensource.google/go/x/tools/+/db513b0915042d9b1ac8cff2c21e191f46578c44

Got a couple green runs in a row in my separate deflaking PR with this dep upgraded, so figured it doesn't hurt to merge to main either and get even more testing.


shykes (approved):

sipsma:
Helm tests are failing due to looking for v0.12.6 engine (I think side effect of today's attempt at a release that we backed out on?), same is on main and everything else green so getting this in

Issue #7699, opened 2024-06-18
ephemeral `context canceled` errors


sipsma:
This has been reported by users and seen occasionally in CI. Most recent occurrence I saw was here https://github.com/dagger/dagger/actions/runs/9551005799/job/26324343803

That passed on many re-runs and also tried running SDK tests again in commits around it without it happening again.

I've been trying to repro what happens in that elixir SDK check failure locally since yesterday but have not replicated it.
* To be clear, I don't think this has anything to do with the elixir SDK, the error in that occurrence was just on trying to load the dagger ci module (happens implicitly when `echo {version} | dagger -s query` runs in the dir with our CI module).

Interestingly, whenever I've seen this happening in our CI it has been during module initialization calls. Things that stick out about that are:
* It's cached, unlike other function calls.
* It potentially gets run in parallel fairly often
* Both of those may impact edge merge behavior, which is an eternal suspect in generally weird issues like this

Since I've had no luck in repro'ing so far, I may try locally enabling caching of all function calls and see if I can hit this, just on the hunch this may have something to do with the caching/edge-merging.

sipsma:
A more recent occurrence: https://dagger.cloud/dagger/traces/d3d78f332e319053276fdfb8c361e847?span=dc9053fa403f4550#dc9053fa403f4550:L8

sipsma:
Fixed a problem with the debug logs around this issue and made a change that should result in user-facing errors having more information required to debug this: https://github.com/dagger/dagger/pull/8553

Still quite rare and have not hit it locally or in my deflaking PR, so continuing to rely on those logs+error messages to make progress atm

sipsma:
Caught this with the updated logs [here](https://dagger.cloud/dagger/traces/eaf8be8f50c029ffe37def2ce8a1b2cd?span=cd508bf447957d24#995e7b387b088431):
```
failed to compute cache key: context canceled: github.com/moby/buildkit/solver.(*edge).createInputRequests.(*edge).createInputRequests.func1.func2
```

Need to follow up with investigation into that code in buildkit, but finally have a new breadcrumb to work with here

sipsma:
[...]

PR #8274, opened 2024-08-30
chore: use ContextCause everywhere

sipsma:
There's a chance this could help us diagnose [ephemeral context canceled flakes](https://github.com/dagger/dagger/issues/7699). 

I have a[ similar change upstream in buildkit](https://github.com/moby/buildkit/pull/5278) that I think is much more likely to be useful, but this is a best practice anyways for providing more context to errors than "canceled" so may as well make this change in our code too.

### DIFF
--- a/cmd/engine/main.go
+++ b/cmd/engine/main.go
@@ -245,10 +245,10 @@ func main() { //nolint:gocyclo
 
 	addFlags(app)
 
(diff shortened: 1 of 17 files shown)


gerhard (approved):
Excited to get more insight into some of the test flakes 💪

jedevc (approved):

PR #8264, opened 2024-08-28
chore: auto determine lint targets

jedevc:
...and also fix the found issues.

PR #8276, opened 2024-08-30
chore: bump golangci-lint timeout

jedevc:
This is currently failing out on `main` - by doing #8264, we have massively increased the number of packages we're linting, so resources on the machine are now more contested. I think it's definitely reasonable to bump this up now.

### DIFF
--- a/modules/golangci/lint.go
+++ b/modules/golangci/lint.go
@@ -107,7 +107,6 @@ func (run LintRun) Report() *dagger.File {
 	cmd := []string{
 		"golangci-lint", "run",
 		"-v",
-		"--timeout", "5m",
 		// Disable limits, we can filter the report instead
 		"--max-issues-per-linter", "0",
 		"--max-same-issues", "0",


gerhard (commented):
`5m` for linting seems long enough. Instead of changing this value, could we:
- Increase the runner size so that this still completes within `5m`
- Declare it as a function argument so that we can change it locally

jedevc:
Counter-proposal - I think we should just remove the timeout from the dagger linter code directly - we're already controlling it at a higher level in github actions:

https://github.com/dagger/dagger/blob/eefbef0aa3b7c50d9ca82abb764f1b972b6a15f0/.github/workflows/engine-and-cli.yml#L24-L36

We can't anticipate the hardware that the caller has, so I think it makes more sense to have no timeout. Though I think, if we want it, we could add it as a parameter to the linter function and propagate it around - but I still think the default should be "unlimited".

PR #8281, opened 2024-08-30
chore: remove deprecated properties for v0.13.0

jedevc:
Makes the API changes specified in #8065.


core/modfunc.go
```diff
@@ -376,7 +376,7 @@ func moduleAnalyticsProps(mod *Module, prefix string, props map[string]string) {
 		git := source.AsGitSource.Value
 		props[prefix+"source_kind"] = "git"
 		props[prefix+"git_symbolic"] = git.Symbolic()
-		props[prefix+"git_clone_url"] = git.CloneURL // todo(guillaume): remove as deprecated
+		props[prefix+"git_clone_url"] = git.CloneRef // todo(guillaume): remove as deprecated
 		props[prefix+"git_clone_ref"] = git.CloneRef
```
helderco, line 380: This is used for analytics? When can we delete it (according to the node, @grouville)?

core/integration/legacy_test.go
```diff
@@ -297,6 +298,45 @@ func (m *Test) Skip() *dagger.Container {
 	require.Equal(t, "hello\n", out)
 }
 
+func (LegacySuite) TestExecWithSkipEntrypointCompat(ctx context.Context, t *testctx.T) {
+	// Changed in ...
```
helderco, line 302: Easy to forget 🙂 
```suggestion
	// Changed in dagger/dagger#8281
```

core/schema/container.go
```diff
@@ -670,12 +701,16 @@ func (s *containerSchema) rootfs(ctx context.Context, parent *core.Container, ar
...
 
 func (s *containerSchema) withExec(ctx context.Context, parent *core.Container, args containerExecArgs) (*core.Container, error) {
-	if args.ContainerExecOpts.SkipEntrypoint != nil {
+	if args.SkipEntrypoint != nil {
 		slog.Warn("The 'skipEntrypoint' argument is deprecated. Use 'useEntrypoint' instead.")
-		if !args.ContainerExecOpts.UseEntrypoint && !*args.ContainerExecOpts.SkipEntrypoint {
+		if !args.ContainerExecOpts.UseEntrypoint && !*args.SkipEntrypoint {
 			args.ContainerExecOpts.UseEntrypoint = true
 		}
 	}
```
helderco, line 716: This is interesting. So this function serves both v0.12 and v0.13, it's just that in v0.13 `SkipEntrypoint` is always `nil`, right?

core/schema/module.go
```diff
@@ -183,6 +183,10 @@ func (s *moduleSchema) Install() {
 	dagql.Fields[*core.GitModuleSource]{
 		dagql.Func("htmlURL", s.gitModuleSourceHTMLURL).
 			Doc(`The URL to the source's git repo in a web browser`),
+		dagql.Func("cloneURL", s.gitModuleSourceCloneURL).
+			View(BeforeVersion("v0.13.0")).
+			Doc(`The URL to clone the root of the git repo from`).
+			Deprecated(`Use CloneRef instead. CloneRef supports both URL-style and SCP-like SSH references`),
```
helderco, line 189: The backticks in the reason message are understood by codegen to format the field name to the language's convention:
```suggestion
			Deprecated("Use `cloneRef` instead. `cloneRef` supports both URL-style and SCP-like SSH references"),
```

helderco (approved):
LGTM, just some minor comments.

helderco:
There's a PHP test using pipeline. Could either be replaced by another API call or removed, depending on what it's really testing. Haven't looked at it.

core/schema/container.go
```diff
@@ -312,6 +313,36 @@ func (s *containerSchema) Install() {
 		dagql.Func("withExec", s.withExec).
 			View(AllVersion).
 			Doc(`Retrieves this container after executing the specified command inside it.`).
+			ArgDoc("args",
+				`Command to run instead of the container's default command (e.g., ["run", "main.go"]).`,
+				`If empty, the container's default command is used.`).
+			ArgDoc("useEntrypoint",
+				`If the container has an entrypoint, prepend it to the args.`).
+			ArgRemove("skipEntrypoint").
```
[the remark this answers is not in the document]
jedevc: This is a shorthand for removing this as a parameter - without this, you'd have to redefine the entire `args` struct, and have a new fun to take this.

Ideally, we'd have even more shorthands for common operations - since redeclaring all the common fields over and over is quite annoying.

core/schema/container.go
jedevc: Yup, exactly :tada: I'd like to do some more cleanup around here, I think some of the changes to multiple schema versions are hard to track, but it works for now :tada:

core/modfunc.go
jedevc: FYI @aluzzardi

jedevc:
Bringing this change in, since the suggested changes have been made now :heart:

PR #8282, opened 2024-08-30
feat: add `Directory.digest` and `ModuleSource.digest`

jedevc:
Builds on #8114, adding `Digest` to `Directory` and `ModuleSource`.

The original need for these is internal, see the discussion in https://github.com/dagger/dagger/pull/8245#discussion_r1733172781. However, there's still a use for these in the public API as well, and it's just generally good consistency :tada:

I guess this actually validates our idea of keeping the digest with the metadata as the default - since it doesn't really make sense to compute a `Directory.digest` *without* metadata, so we can just leave off that parameter.


helderco:
Yeah, you're right! Giving a looksy :)

helderco (approved):
LGTM!

PR #8289, opened 2024-08-31
sdk(go): simplify DaggerObject aliasing slightly

sipsma:
This gets rid of the multiple levels of type aliasing used to be just one level.

This is a very mild simplification that worst case seems harmless, but the motivation right now is a shot in the dark attempt at seeing if it avoids [some of the flakes](https://github.com/dagger/dagger/issues/8207) we have been getting in the Go SDK with this object.
* Obviously if this *does* indeed get rid of the flakes then it's not great in that we've probably just temporarily avoided a bug, but at this point that seems worth it.

I have been running with this change in my deflaking PR for a while now and haven't seen the flake since, so figured it didn't hurt to spin out at this point.
* 90% of time I've said something like this in the past the flake suddenly reappears, so 🤞

### DIFF
--- a/cmd/codegen/generator/go/templates/modules.go
+++ b/cmd/codegen/generator/go/templates/modules.go
@@ -19,10 +19,11 @@ import (
 )
 
 const (
-	daggerGenFilename     = "dagger.gen.go"
-	contextTypename       = "context.Context"
-	constructorFuncName   = "New"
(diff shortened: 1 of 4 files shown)


jedevc (approved):
Yeah let's try it - fingers crossed this does *something*.

Hopefully with the improved debugging logs and such, this should give a bit more context as well as to what could be really happening.

Issue #6655, opened 2024-02-13
Dagger module error message with a mistyped --sdk flag


slumbering:
When I initialize a new module with `dagger init` and I mistyped the `--sdk`, I'm getting the following error.

```
$ dagger init --name=my-module --sdk=typescrip
Error: failed to generate code: input: resolve: moduleSource: withName: withSDK: withSourceSubpath: resolveFromCaller: failed to collect local module source deps: failed to collect local sdk: missing config file /Users/slumbering/workdir/dagger-zenith/trivy/typescrip/dagger.json
```

I'm wondering if we could improve this message. Do you think there's anything we could do to make it clearer or more effective?

e.g:

```
The typescrip SDK does not exist.

Available SDKs:
- go
- typescript
- python
- ...
```

TomChv:
@helderco @jedevc @vito Is it still a relevant issue?

jedevc:
Yes, the confusing error still appears.

slumbering:
I like the way AWS CDK makes the error clear:

```shell
$ cdk init app --language g
Invalid values:
  Argument: language, Given: "g", Choices: "csharp", "fsharp", "go", "java", "javascript", "python", "typescript"
 ```

**Comment by TomChv** [comment]
@helderco I'm pretty sure it's doable from the CLI, but the `sdk` field could also be a URL so we would need to be generic on the helpers.

Like

```
The typescrip SDK does not exist.

Available SDKs:
- go
- typescript
- python

You can also use a no-bundled SDK from its github URL (e.g., https://github.com/dagger/dagger/sdk/rust@main)
```

rajatjindal:
I am taking a look at it today. I am planning to add:

- a local var in cmd/dagger/module.go with list of sdk supported.
- a validation function that verifies that if the sdk is provided, it is either one of the configured list or an `https` url.
- add a `PreRunE` step to cobra commands `init` and `develop`.

I think we should catch this error even before our call reaches engine, and therefore i am adding validation in `PreRunE` step, but also open to suggestions here.

If this approach looks good, I can open a PR for this.

jedevc:
> * add a `PreRunE` step to cobra commands `init` and `develop`.

Hm, this is a bit problematic. If we do this, then it means if the engine has a new SDK that the CLI doesn't know about it might not work. We have *some* tolerance for different versions of CLI and engines to connect, so we need to take that into account.

> `https` url.

[...]

PR #8297, opened 2024-09-02
fix: use sdkVersion when loading aliased sdk

rajatjindal:
while working on a different [issue](https://github.com/dagger/dagger/issues/6655), I found what seems like we are not using `sdkVersion` when loading aliased sdk.


rajatjindal:
the `.next` currently refers to v0.12.7? should I change that to v0.13.0 (as you added milestone as v0.13.0)

jedevc:
See the instructions in `CONTRIBUTING.md` to see how to add a changelog - you need `changie` installed, you don't need to manually edit `.next`.

Issue #8291, opened 2024-09-02
🖐️ dagger install fails


wvell:
### What happened? What did you expect to happen?

I'm probably doing something wrong but i cant seem to figure out what is going wrong.
When i try to install the example from dagger install --help i get the following output:

    dagger install github.com/shykes/hello@v0.1.0

```
Output:
✘ moduleSource(refString: "github.com/shykes/hello@v0.1.0"): ModuleSource! 0.0s
! failed to resolve git tags: select: git command failed: exit status 128
Error: failed to get module ref kind: input: moduleSource resolve: failed to resolve git tags: select: git command failed: exit status 128
stdout:
stderr: fatal: 'github.com/shykes/hello' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

```

```
Version git:
git version 2.43.0

Version dagger:
dagger v0.12.6 (registry.dagger.io/engine:v0.12.6) linux/amd64
```

jochumdev:
for me dagger install works when i prepend the url with the https scheme:

```

**Comment by wvell** [comment]
Thanks a lot. That fixes it for me :)

**Comment by jedevc** [comment]
Hm, this definitely used to work. This is likely a regression introduced by #7708 - cc @grouville.

Leaving off the version tag also seems to cause this to work, but we should fix this asap.

**Comment by jedevc** [comment]
Prepping a patch release today for this: https://github.com/dagger/dagger/pull/8299 - sorry for the inconvenience y'all :heart: 

As a workaround, yup, you can manually use `https://`.

**Comment by jedevc** [comment]
This is now fixed in v0.12.7. However, note, the command `dagger install github.com/shykes/hello@v0.1.0` will not work.

This module version number is out of date (and the docs are as well for including it). As an example hello world module, I recommend using `dagger install github.com/shykes/daggerverse/hello@v0.3.0` (and I've also updated the docs to use this as well) :tada:

**Comment by pjmagee** [comment]
`dagger install github.com/shkyes/daggerverse/hello@v0.3.0` doesn't work for me. :( 

```
mageep@MacBook-Air dagger-private-hello-world % dagger install github.com/shkyes/daggerverse/hello@v0.3.0
Setup tracing at https://dagger.cloud/traces/setup. To hide: export STOPIT=1

✔ connect 0.7s
✔ cache request: mkfile /schema.json 0.0s
✔ load cache: mkfile /schema.json 0.0s
[...]

PR #8298, opened 2024-09-02
fix: make Git.tags url parsing consistent with Git.tree

jedevc:
Fixes https://github.com/dagger/dagger/issues/8291, regression from https://github.com/dagger/dagger/pull/7708.

Previously, extracting the git tags from a URL skipped a level of url pre-processing that was in buildkit (essentially, one that allowed no protocols in URLs). We need to add this level of pre-processing manually.

Without this, parsing module refs of the form `github.com/<org>/<repo>@<version>` would fail with:

	! failed to resolve git tags: select: git command failed: exit status 128
	Error: failed to get module ref kind: input: moduleSource resolve: failed to resolve git tags: select: git command failed: exit status 128
	stdout:
	stderr: fatal: 'github.com/<org>/<repo>' does not appear to be a git repository
	fatal: Could not read from remote repository.

This is because it's not a repo - it needs that same processing to be consistent with `Git.Tree`.

> [!WARNING]
>
> We *need* to add a module source test for this, but I could not see a super easy way to do this without tagging in every `dagger-test-modules` repo. As a follow-up, we should do this, and additionally, make sure that everyone has access to those repos, sharing those creds in 1password.

### DIFF
--- a/.changes/unreleased/Fixed-20240902-130323.yaml
+++ b/.changes/unreleased/Fixed-20240902-130323.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: |
+  make Git.tags url parsing consistent with Git.tree
+time: 2024-09-02T13:03:23.920914271+01:00
+custom:
+  Author: jedevc
+  PR: "8298"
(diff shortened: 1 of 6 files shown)


gerhard (approved):
LGTM!

WDYT @helderco ?

helderco (approved):
LGTM, but there's two change logs?

PR #8299, opened 2024-09-02
Prep v0.12.7 release

jedevc:
Changes since last release: https://github.com/dagger/dagger/compare/v0.12.6...main


jedevc:
note: helm test is expected to fail since it's trying to pull the unreleased v0.12.7 tag. I need to fix this, but haven't had time since the last release.

.changes/v0.12.7.md
```diff
@@ -0,0 +1,12 @@
+## v0.12.7 - 2024-09-02
+
+
+### Fixed
+- Fixed regression in module ref parsing for no-protocol refs that had version suffixes by @jedevc in https://github.com/dagger/dagger/pull/8298
+- Make `Git.tags` URL parsing consistent with `Git.tree` by @jedevc in https://github.com/dagger/dagger/pull/8298
```
gerhard, line 6: This reads as 2 fixed, when there's just one. Want to merge & maybe reword @jedevc?

helm/dagger/CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and is generated by [Changie](https://github.com/miniscruff/changie).
 
 
 
+## v0.12.7 - 2024-09-02
+
+
```
gerhard, line 12: We should not have releases with no changes. The change worth mentioning here is that the Engine version now defaults to `v0.m.p`, same as we do for SDKs that may not have any changes.

gerhard (approved):
2 changes worth making before merging, but otherwise LGTM.

Approving so that this gets expedited 💨

.changes/v0.12.7.md
gerhard: OK!

helm/dagger/CHANGELOG.md
jedevc: Resolved, will update the release guide for next time.

Issue #8301, opened 2024-09-02
Helm: Do not allow specifying Dagger version


gerhard:
## Initial report by @smolinari 

<img width="996" alt="image" src="https://github.com/user-attachments/assets/6db84b4d-3af5-4e06-8347-75bf8fb09c34">

## The actual problem + solution

<img width="1098" alt="image" src="https://github.com/user-attachments/assets/63fdb613-814d-47a9-b0a1-af68c1d19685">

## The better solution
Now that we publish a new Helm version whenever there is a new Dagger version - in other words the chart version and the Dagger version are moving in lock-step - we should not allow specifying different versions for Dagger.

We should still support setting custom images in case someone wants to use a specific image variant. We should add a comment warning which explains why the two versions need to be kept in lock-step.

smolinari:
So, just for my understanding for the future, let's say I have a bunch of CI processes running on engine 0.X.X., If I upgrade the Helm chart, it could break the processes. This is because Dagger is still considered pre-production (per SemVer) and the API can have breaking changes (as you noted will happen in 0.13.0 in our Discord chat @gerhard), so upgrading the Helm chart must be done with caution and understanding which Engine versions will have breaking changes is very important. 

Once Dagger is 1.0.0, only major version changes (i.e. 2.x.x, 3.x.x, 4.x.x etc.) will need external work done to upgrade the Helm chart. 

Is that a correct take?
 
Scott

gerhard:
Yes, that is correct.

Before `1.0`, minor versions can be considered major ones. We tend to introduce the most breaking changes in minors, e.g.:
- `v0.12.0` https://github.com/dagger/dagger/blob/main/CHANGELOG.md#v0120---2024-07-12
- `v0.11.0` https://github.com/dagger/dagger/blob/main/CHANGELOG.md#v0110---2024-04-03

There are also cases when we don't, e.g.:
- `v0.10.0` https://github.com/dagger/dagger/blob/main/CHANGELOG.md#v0100---2024-02-27

The truth is that we also introduce smaller breaking changes in patch versions (search for breaking): https://github.com/dagger/dagger/blob/main/CHANGELOG.md

In conclusion: always look at the release notes before upgrading the Helm chart version. And always pin!

cc @jedevc

marcosnils:
> ## The better solution
> Now that we publish a new Helm version whenever there is a new Dagger version - in other words the chart version and the Dagger version are moving in lock-step - we should not allow specifying different versions for Dagger.

[...]

PR #8305, opened 2024-09-02
ci: use main image tag for helm deployment test

jedevc:
This should prevent the weird helm test failure described in https://github.com/dagger/dagger/pull/8268#discussion_r1736117495 and https://github.com/dagger/dagger/pull/8299#issuecomment-2324783582, which happens when we bump the `appVersion` in the `Chart.yaml` to a version of the dagger engine that hasn't yet been released.  

Potentially conflicts with https://github.com/dagger/dagger/issues/8301, but I'm not quite sure of another good approach here :thinking:


.dagger/helm.go
```diff
@@ -48,7 +48,7 @@ func (h *Helm) Test(ctx context.Context) error {
 		WithEnvVariable("KUBECONFIG", "/.kube/config").
 		WithEnvVariable("CACHEBUSTER", identity.NewID()).
 		WithExec([]string{"kubectl", "get", "nodes"}).
-		WithExec([]string{"helm", "install", "--wait", "--create-namespace", "--namespace=dagger", "dagger", "."}).
+		WithExec([]string{"helm", "install", "--wait", "--create-namespace", "--namespace=dagger", "--set=engine.image.tag=main", "dagger", "."}).
```
gerhard, line 51: Instead of the `engine.image.tag`, we should set the entire image.

Let me whip up that PR - that will be a breaking change, just in time for `v0.13.0` - and then this can be rebased on top of that one.

gerhard: https://github.com/dagger/dagger/pull/8348

jedevc: Rebased to use `engine.image.ref`

gerhard (approved):
👍 🚀

PR #8306, opened 2024-09-03
fix: fix the parseBuiltinSDK function when engine.Tag is not empty value

rajatjindal:
I realized that the sdkVersion defaults to `engine.Tag` and in a released version it would be a semver release version. The unit tests didn't consider that in the original implementation.

The unit tests have now been modified to set an `engine.Tag` to simulate real usage, and validation logic has been fixed to account for that correctly. I have also added several additional tests to cover these scenarios.

This PR also modifies the function `parseSDKName` to include basic validation, which was done earlier in the `builtinSDK` function. Therefore, it also renames the function to `parseAndValidateBuiltinSDKName`.

### DIFF
--- a/core/schema/sdk.go
+++ b/core/schema/sdk.go
@@ -20,16 +20,21 @@ import (
 
 const (
 	runtimeWorkdirPath = "/scratch"
-	SDKGo              = "go"
-	SDKPython          = "python"
-	SDKTypescript      = "typescript"
-	SDKPHP             = "php"
-	SDKElixir          = "elixir"
+)
+
+type SDK string
+
+const (
+	SDKGo         SDK = "go"
+	SDKPython     SDK = "python"
+	SDKTypescript SDK = "typescript"
+	SDKPHP        SDK = "php"
+	SDKElixir     SDK = "elixir"
 )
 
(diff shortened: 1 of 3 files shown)


rajatjindal:
cc @jedevc

core/schema/sdk.go
```diff
@@ -99,25 +99,42 @@ func (s *moduleSchema) sdkForModule(
...
-// for sdkName with format <sdk-name>, it returns <sdk-name> with
-// empty suffix.
-func parseSDKName(sdkName string) (string, string) {
+// If sdk is one of go/python/typescript and <version>
+// is specified, we return an error as those sdk don't support
+// specific version
+//
+// if sdk is one of php/elixir and version is not specified,
+// we defaults the version to [engine.Tag]
+func parseAndValidateBuiltinSDKName(sdkName string) (string, string, error) {
```
jedevc, line 113: We can keep this called `parseSDKName` - I'm not a fan of overly verbose names, they make the code much harder to read :cry: 

One other suggestion that might improve readability in this area if you feel like it - we could make `SDKGo` and similar into an enum with a new `type SDK string`.

jedevc (approved):

PR #8308, opened 2024-09-03
ci: fix schema reference generation

jedevc:
The schema reference generation takes the `schema.graphqls` file as input, so each generate step effectively required two passes. This is kinda messy, so we should just feed one generation step into the next.

This is kinda neat, since the dagger engine will deduplicate the schema generation, so we don't end up computing it and generating it twice.


grouville (approved):
Love it ! Lint failed after rebase, timeout and succeeded prior rebase ✅

PR #8314, opened 2024-09-03
docs: Update Dagger version numbers in docs

vikram-dagger:


vikram-dagger:
I know we are doing this already as part of the release process, but it looks like it's only making the change in the installation page. Should we also be updating in the integrations and other pages?

kpenfound (approved):

jedevc:
@vikram-dagger ah, we definitely should, sorry, I've been missing these.

Feel free to add any other places that need updating in https://github.com/dagger/dagger/blob/91614fbc11ef9785088ca9692c8e6a4ed548499e/RELEASING.md#L332-L333.

*At some point* we should script this :cry:

Issue #7777, opened 2024-06-27
✨ `KeepGitDir` should be set to `true` everywhere by default as it's often needed and can cause cryptic errors


jpadams:
**NOTE: Likely this would be a good candidate for `v0.12.0` as it's a possible 💥  breaking change for some users.**
 
## The problem
If `KeepGitDir` is `true`, the `.git` directory is preserved when a git repo is used as a Dagger `Directory`. If it's `false`, then the `.git` directory is omitted.

Today, there is an inconsistency in the default for `KeepGitDir`:

- In the Dagger API, `KeepGitDir` is set to `false` by default.
- When a git repo ref is provided as an argument of type Directory in the dagger CLI, then it behaves as ifKeepGitDir is set to true by default.

## The proposal
- I'd like to see the default to be `true` everywhere since I find myself needing the `.git` dir often, haven't had any issues with it being there by default, and have had cryptic errors when it was missing but required subsequent steps. In any case, the split between CLI and API should be made consistent.

### In the Dagger API, `KeepGitDir` is set to `false` by default.
```typescript
dag.git("https://github.com/dagger/dagger").branch("main").tree() // keepGitDir false by default
// dag.git("https://github.com/dagger/dagger", {keepGitDir: true}).branch("main").tree()
```
[Used in this function](https://github.com/jpadams/testgit/blob/main/dagger/src/index.ts#L26C40-L26C77):
```
dagger call -m github.com/jpadams/testgit \
git-ls --repoStr https://github.com/dagger/dagger stdout

total 336K
drwxr-xr-x   19 root     root        4.0K Jun 27 19:08 .
drwxr-xr-x    1 root     root        4.0K Jun 27 19:16 ..
drwxr-xr-x    3 root     root        4.0K Jun 27 19:08 .changes
-rw-r--r--    2 root     root         962 Jun 27 19:08 .changie.yaml
drwxr-xr-x    4 root     root        4.0K Jun 27 19:08 .github
-rw-r--r--    2 root     root         561 Jun 27 19:08 .gitignore
-rw-r--r--    2 root     root         157 Jun 27 19:08 .gitmodules
-rw-r--r--    2 root     root        1.5K Jun 27 19:08 .golangci.yml
...
```

### When a git repo ref is provided as an argument of type `Directory` in the `dagger` CLI, then it behaves as if`KeepGitDir` is set to `true` by default.

[...]

Issue #3050, opened 2022-08-22
`core` `git` should have `keepGitDir` option and perhaps `depth`. Should it be `true` by default?


jpadams:
I've found it confusing that when I see demos of doing a git pull of the `dagger/dagger` repo and a `dockerbuild` that the `docker version` command run afterwards doesn't show a short git hash in the version number like it does when I `git clone` and build `dagger` locally on my machine.

For example:

```
dagger devel (708210f74) darwin/arm64
```

```
dagger devel (708210f74) linux/arm64
```

```graphql
{
    core {
        git(remote: "https://github.com/dagger/dagger") {
            dockerbuild(dockerfile: "Dockerfile") {
                exec(input:{args:["dagger", "version"]}) {
                    stdout
                }
            }
        }
    }
}

{
  "data": {
    "core": {
      "git": {
        "dockerbuild": {
          "exec": {
            "stdout": "dagger devel () linux/arm64\n"
          }
        }
      }
    }
  }
}
```
I expected a `git clone` type functionality. Not sure if other folks are expecting something else. If the size of the `.git` directory is a concern, we could use something like `git clone --depth=1`.

It seems that perhaps the default BuildKit behavior is to not preserve the `.git` directory (not sure), but I see that this is an option in BuildKit LLB land (depending on things like dockerfile frontends and such I think), and so we should have that option in `cloak` to keep the `.git` directory and likely limit the clone depth and other bells and whistles.

https://github.com/dagger/dagger/blob/main/plan/task/gitpull.go#L40-L43
https://github.com/dagger/dagger/blob/main/version/version.go#L16-L30
https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#example-keep-git-dir

sipsma:
We have support for including the git dir now. We don't have an option for git depth though

vito:
I'm leaning towards keeping the `.git` dir by default at this point. It seems like we've needed it more than half the time we've used the Git API.

helderco:
Closing since this is done. If reopening because of making `keepGitDir` default to `true`, please update issue or create new one.

PR #8318, opened 2024-09-03
feat: replace `keepGitDir` with `discardGitDir`

jedevc:
Implements https://github.com/dagger/dagger/issues/7777


aluzzardi:
@jedevc 👋 Late to the party -- won't this affect caching? I thought the reason buildkit defaults to NOT doing it is because of how it resets the cache pretty often.

jedevc:
@aluzzardi some discussion in https://github.com/dagger/dagger/issues/7777#issuecomment-2214944610, does that help?

gerhard:
Do we still want this in `v0.13.0`? The ETA is this week.

If this does not look ready, we can push it to a patch. WDYT @aluzzardi @helderco ?

jedevc:
Because it's a deprecation it doesn't need to be in this release - happy to push.

jedevc:
Updated this - if we're changing the property, we should also move it to `GitRef.tree` instead of being at the top-level.

core/integration/git_test.go
```diff
@@ -116,6 +116,38 @@ func (GitSuite) TestGit(ctx context.Context, t *testctx.T) {
 	require.Contains(t, res.Git.Tag.Tree.File.Contents, "Dagger")
 }
 
+func (GitSuite) TestDiscardGitDir(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	t.Run("git dir is present", func(ctx context.Context, t *testctx.T) {
+		dir := c.Git("https://github.com/dagger/dagger").Branch("main").Tree()
+		ent, _ := dir.Entries(ctx)
```
vito, line 124: nit: always worth including the `require.NoError` check for these imo - having wasted time before chasing confounding second-order failures

core/integration/git_test.go
```diff
@@ -116,6 +116,38 @@ func (GitSuite) TestGit(ctx context.Context, t *testctx.T) {
...
+		require.Contains(t, ent, ".git")
+	})
+
+	t.Run("git dir is not present", func(ctx context.Context, t *testctx.T) {
+		dir := c.Git("https://github.com/dagger/dagger").Branch("main").Tree(dagger.GitRefTreeOpts{DiscardGitDir: true})
+		ent, _ := dir.Entries(ctx)
+		require.NotContains(t, ent, ".git")
```
vito, line 131: (continuing previous nit) - this would false-pass if `Entries` failed

core/integration/legacy_test.go
```diff
@@ -535,3 +535,60 @@ func (LegacySuite) TestModuleSourceCloneURL(ctx context.Context, t *testctx.T) {
 	require.Equal(t, "https://github.com/dagger/dagger.git", res.ModuleSource.AsGitSource.CloneRef)
 	require.Equal(t, res.ModuleSource.AsGitSource.CloneRef, res.ModuleSource.AsGitSource.CloneURL)
 }
+
+func (LegacySuite) TestGitWithKeepDir(ctx context.Context, t *testctx.T) {
+	// Changed in dagger/dagger#...
```
vito, line 540: i gotchu

```suggestion
	// Changed in dagger/dagger#8318
```

core/integration/services_test.go
```diff
@@ -953,7 +953,7 @@ func (ServiceSuite) TestDirectoryEntries(ctx context.Context, t *testctx.T) {
 
 	entries, err := c.Git(repoURL, dagger.GitOpts{ExperimentalServiceHost: gitDaemon}).
 		Branch("main").
-		Tree().
+		Tree(dagger.GitRefTreeOpts{DiscardGitDir: true}).
```
vito, line 956: nit: I would be cool with updating the assertions instead to just include `.git`, since setting this here makes it feel more significant than it really is (I think?)

docs/docs-graphql/schema.graphqls
```diff
@@ -1630,7 +1630,10 @@ type GitRef {
   id: GitRefID!
 
   """The filesystem tree at this ref."""
-  tree: Directory!
+  tree(
+    """Set to true to discard .git directory."""
+    discardGitDir: Boolean = false
```
vito, line 1635: Mini-bikeshed: I wonder if we even need this? With #8436 you could just exclude `.git` at the argument site, which would then be applied no matter how the `Directory` was created. But - I guess you don't want every call site to have to anticipate and exclude a `.git` directory. So it kind of makes sense to do it before passing it in, too. But you could also do `.withoutDirectory(".git")` which might be even less typing depending on your SDK. 🤷‍♂️

vito (approved):
Approving in principle - glad to see this change (https://github.com/dagger/dagger/issues/3050#issuecomment-1426197069). Just some nits, and looks like it needs a rebase.

docs/docs-graphql/schema.graphqls
TomChv: There's an issue with Python uv that hold #8436 so for now we cannot merge it, but yeah I think it's not a good DX to force `// +ignore=[".git/"]` on every call.

jedevc: Good point. I'm happy with the idea of removing it, but also not opposed to keeping it.

sipsma (approved):

docs/docs-graphql/schema.graphqls
sipsma: I think the point that this would prevent you from needing to always try to ignore/remove the dir out of caution is enough justification for keeping this option in place.

PR #8321, opened 2024-09-03
docs: Improve PHP integration page

vikram-dagger:


levlaz:
Couple of thoughts here. 

The guide feels bit fragmented because were trying to show three things at once but not all of the sections are relevant to them. 

Daggerize a PHP project 
Daggerize a pipeline using the PHP SDK 
Write a dagger module using the PHP SDK 

We mention #2, but show #1 in the example. 

We don't really show an example of #3. 

It also feels a bit odd that we leave pre-req empty, that might be a good place to discuss any local PHP requirements for IDE integration for example. 

I can see similar fragmentation in all the language/framework guides like this, so may be good to come up with a unified approach here before getting too far.

vikram-dagger:
> The guide feels bit fragmented because were trying to show three things at once but not all of the sections are relevant to them.
> 
> 1. Daggerize a PHP project
> 2. Daggerize a pipeline using the PHP SDK
> 3. Write a dagger module using the PHP SDK
> 
> We mention #2, but show #1 in the example.
> 
> We don't really show an example of #3.

I think this is a fair point. I have restructured the content to focus on one approach, adding information about the experimental SDK at the end as a simple note. I think this offers clearer guidance.

> It also feels a bit odd that we leave pre-req empty, that might be a good place to discuss any local PHP requirements for IDE integration for example.

I don't think there are any such requirements but lmk in case you have more info on this.

mircubed:
@vikram-dagger , wait to merge. Community members are working on an update page. See Discord chat here - https://discord.com/channels/707636530424053791/1162025276872609832/1283537090948108298

PR #8331, opened 2024-09-04
fix: allow field access with default constructor

jedevc:
Previously, if a field access was made immediately after the default constructor was called, then the access would fail (because the field had not been set).

So, in this case, we try and convert a nil value, to get the default value for this.

---

Genuinely not quite sure how we never managed to hit this before, it's very easy - I just found it while playing around with some tests in #8149.


gerhard (approved):
Approving optimistically.

OK to merge after `changie` & **maybe** rebase, but otherwise LGTM.

TomChv (approved):
LGTM

Issue #8326, opened 2024-09-03
 `dagger core dagger-engine local-cache` prunes the cache


shykes:
## Dagger version

0.12.6

## What I did

`dagger core dagger-engine local-cache`

## What I expected

Some output on the terminal, possibly useful, with no side effects.

## What happened

The local cache is pruned

## Why?

`prune()` is a function which returns a scalar, and has no required arguments, so it gets selected by the "auto-print filter", and is called.

helderco:
Should be skipped because it returns a `Void`:
```graphql
type DaggerEngineCache {
  prune: Void
}
```
 However, it seems nullable voids are being converted to an optional `SCALAR_KIND`, instead of an optional `VOID_KIND`, when create TypeDefs out of core types.

PR #8336, opened 2024-09-04
fix void types from core incorrectly being seen as named scalar types

helderco:
Fixes https://github.com/dagger/dagger/issues/8326


jedevc (approved):
Eek. Nice catch :tada:

PR #8340, opened 2024-09-05
docker driver: Avoid registry lookups if the image was already pulled

aluzzardi:
See https://github.com/dagger/dagger/pull/8232 for context

Alternative of #8341

PR #8341, opened 2024-09-05
docker driver: Use tags instead of digests to avoid registry lookups

aluzzardi:
See https://github.com/dagger/dagger/pull/8232 for context

Alternative of #8340

This is a much simpler alternative that works as follows:

- If the `docker-image://` contains a digest, it uses that as the container ID (same behavior as before)
- Otherwise, it will use the image tag as container ID (as opposed to reaching out to a registry to convert tag to digest)

### DIFF
--- a/engine/client/drivers/docker.go
+++ b/engine/client/drivers/docker.go
@@ -12,9 +12,7 @@ import (
 	"strings"
 
 	"dagger.io/dagger/telemetry"
-	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
-	"github.com/google/go-containerregistry/pkg/v1/remote"
 	connh "github.com/moby/buildkit/client/connhelper"
 	connhDocker "github.com/moby/buildkit/client/connhelper/dockercontainer"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


engine/client/drivers/docker.go
```diff
@@ -194,6 +141,27 @@ func (d *dockerDriver) create(ctx context.Context, imageRef string, opts *Driver
...
+			return "", errors.Errorf("invalid image reference %q", imageRef)
+		}
+		return id[:hashLen], nil
+	}
+	if tag, ok := ref.(name.Tag); ok {
+		// Otherwise, fallback to the image tag
+		return tag.TagStr(), nil
+	}
+
+	return "", errors.Errorf("invalid image reference %q", imageRef)
```
jedevc, line 162: :thinking: this shouldn't error out - theoretically, you *can* do `docker-image://registry.dagger.io/engine` without a tag. In this case, we should default the `Tag` to `latest`?

tangent: Also what, `name.ParseReference` is weird. Can it not parse mixed tags+digests :thinking: which is totally valid as a thing.

aluzzardi: Done for fallback to `latest` -- PTAL

re: `name.ParseReference` it seems to be working in all cases:

```
# Tag only
$ _EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-image://registry.dagger.io/engine:v0.12.5 ./bin/dagger functions
      ✔ exec docker run --name dagger-engine-v0.12.5 -d --restart always -v /var/lib/dagger --privileged registry.dagger.io/engine:v0.12.5 --debug 0.2s

# Tag+digest
$  _EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-image://registry.dagger.io/engine:v0.12.5@sha256:7d4613636ea6099e3e99c1063d9bbebece1a3465fffac18b98040aa371a00c74 ./bin/dagger functions
      ✔ exec docker run --name dagger-engine-7d4613636ea6099e -d --restart always -v /var/lib/dagger --privileged registry.dagger.io/engine:v0.12.5@sha256:7d4613636ea6099e3e99c1063d9bbebece1a3465fffac18b98040aa371a00c74 --debug 0.2s

# Digest-only
$  _EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-image://registry.dagger.io/engine@sha256:7d4613636ea6099e3e99c1063d9bbebece1a3465fffac18b98040aa371a00c74 ./bin/dagger functions
      ✔ exec docker run --name dagger-engine-7d4613636ea6099e -d --restart always -v /var/lib/dagger --privileged registry.dagger.io/engine@sha256:7d4613636ea6099e3e99c1063d9bbebece1a3465fffac18b98040aa371a00c74 --debug 0.2s
```

sipsma:
Was catching up on discord msgs during PTO and came across all this, for a little more context on the approach before this PR: the reasoning behind the lookup of the digest for a semver tagged image was to handle the case where we needed to re-push the image.

In the (relatively speaking) distant past we'd had situations where we hit one or another issue and pushed fixed/updated engine images of an already published tag.

We haven't done that in a long time and are at the point where I think we can definitively say "we never should and never will do that" (treat the engine image tags as immutable, essentially), so the simplification here SGTM! If we ever need to fix something in an image we should just do another release with an incremented patch version.

sipsma (approved):

gerhard (approved):
LGTM 👍 🚀

Issue #8259, opened 2024-08-28
🐞 Newline log spam with `--progress=plain`


skycaptain:
### What is the issue?

When running the command below, I get a lot of newlines, which spam the output, when using `--progress=plain`. I'm uncertain whether this is an issue with Dagger or something else. I've only observed this behaviour with `apt` so far, but I cannot recall to have seen this with previous Dagger releases. If I build the equivalent with Docker I do not see the newlines.

### Dagger version

dagger v0.12.5 (registry.dagger.io/engine:v0.12.5) darwin/arm64

### Steps to reproduce

Build with dagger:

```shell
dagger --progress=plain core container from --address debian with-exec --args 'apt,update' with-exec --args 'apt,install,-y,python3'
```

Build with Docker

```dockerfile
FROM debian
ENV DEBIAN_FRONTEND="noninteractive"
RUN apt-get update && apt-get install -y python3
```

```shell
docker build --progress=plain --no-cache .
```

### Log output

Dagger output:

```plain
1   : connect
2   :   starting engine
3   :     create
4   :       exec docker pull registry.dagger.io/engine:v0.12.5
4   :       [1.2s] | v0.12.5: Pulling from engine
4   :       [1.5s] | a258b2a6b59a: Pulling fs layer
4   :       [1.5s] | 387969522d8d: Pulling fs layer
4   :       [1.5s] | 5f0af5112188: Pulling fs layer
4   :       [1.5s] | 6d60be01f395: Pulling fs layer
4   :       [1.5s] | 06bb48973f05: Pulling fs layer
4   :       [1.5s] | b4821e571c3f: Pulling fs layer
4   :       [1.5s] | 661cd339ce91: Pulling fs layer
4   :       [1.5s] | 9a5d66e8e99b: Pulling fs layer
4   :       [1.5s] | df33b8fa60c8: Pulling fs layer
4   :       [1.5s] | ef4bee3e3526: Pulling fs layer
4   :       [1.5s] | cc0f2f9d4961: Pulling fs layer
4   :       [1.5s] | b8a82878608c: Pulling fs layer
4   :       [1.5s] | 98c75ef084f6: Pulling fs layer
4   :       [1.5s] | aa753023df44: Pulling fs layer
4   :       [1.5s] | e77de883480e: Pulling fs layer
4   :       [1.5s] | 8a4361f30d7a: Pulling fs layer
4   :       [1.5s] | 9d7f0519fee0: Pulling fs layer
[...]

PR #8347, opened 2024-09-05
fix: correctly handle carriage returns for plain progress

jedevc:
Fixes https://github.com/dagger/dagger/issues/8259

This was incorrectly rendering some carriage return output (outside of certain test cases). To handle this, we need to track a proper cursor, which is a bit annoying, but unavoidable.

The approach in https://github.com/dagger/dagger/issues/7653 wasn't *quite* right (since `\r\n` would essentially wipe the entire contents of the line). But also, what silly unix tool would produce windows-style newlines like that? (apt, that's who)

### DIFF
--- a/.changes/unreleased/Fixed-20240911-102046.yaml
+++ b/.changes/unreleased/Fixed-20240911-102046.yaml
@@ -1,0 +1,9 @@
+kind: Fixed
+body: |-
+  Plain progress correctly displays carriage returns
+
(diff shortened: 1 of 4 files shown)


vito (approved):
Nice - this sort of approach might even make sense for regular log output tbh. A full blown vterm is overkill for most situations - just keeping track of formatting and handling `\r` would at least cover common cases like progress bars. (Just, maybe not multi-line progress bars...)

PR #8348, opened 2024-09-05
💥 Breaking(helm): Remove appVersion - this is now the chart version

gerhard:
Now that the chart version moves in lock-step with the app version, we want to remove the app version so that we minimize the change for incompatibilities, as was the case in https://github.com/dagger/dagger/issues/8301.

The options are:
A. You want a custom app image. You need to ensure that the chart is compatible with this image.
B. You use the default image. We are responsible for ensuring compatibility with the chart.

The third option is to use this YAML as inspiration, and manage your own raw manifests, which give you full control, and also responsibility.

Fixes https://github.com/dagger/dagger/issues/8301

### DIFF
--- a/helm/dagger/templates/engine-daemonset.yaml
+++ b/helm/dagger/templates/engine-daemonset.yaml
@@ -44,7 +44,7 @@ spec:
       {{- end }}
       containers:
         - name: dagger-engine
-          image: {{ .Values.engine.image.repository }}:{{ if .Values.engine.image.tag }}{{ .Values.engine.image.tag }}{{ else }}v{{ .Chart.AppVersion }}{{ end }}
+          image: {{ if .Values.engine.image.ref }}{{ .Values.engine.image.ref }}{{ else }}registry.dagger.io/engine:v{{ .Chart.Version }}{{ end }}
           imagePullPolicy: {{ .Values.engine.image.pullPolicy }}
           {{- if .Values.engine.args }}
           args:
@@ -58,9 +58,9 @@ spec:
           - secretRef:
               {{- if .Values.magicache.secretName }}
               name: {{ .Values.magicache.secretName }}
-              {{- else}}
+              {{- else }}
               name: {{ include "dagger.fullname" . }}-magicache-token
-              {{- end}}
+              {{- end }}
           {{- end }}
           securityContext:
             privileged: true
--- a/helm/dagger/templates/engine-daemonset.yaml
+++ b/helm/dagger/templates/engine-daemonset.yaml
@@ -74,7 +74,7 @@ spec:
                 - sh
                 - -exc
                 - |-
-                  echo '{ version }' | dagger query
+                  dagger core version
             {{- if .Values.engine.readinessProbeSettings }}
             {{- toYaml .Values.engine.readinessProbeSettings | nindent 12 }}
             {{- end }}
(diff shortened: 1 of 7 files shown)


jedevc:
Pushed a commit to remove `appVersion` from the ci `setVersion` task as well.

LGTM :tada:

jedevc (approved):

helm/dagger/templates/engine-daemonset.yaml
```diff
@@ -44,7 +44,7 @@ spec:
       {{- end }}
       containers:
         - name: dagger-engine
-          image: {{ .Values.engine.image.repository }}:{{ if .Values.engine.image.tag }}{{ .Values.engine.image.tag }}{{ else }}v{{ .Chart.AppVersion }}{{ end }}
+          image: {{ if .Values.engine.image.ref }}{{ .Values.engine.image.ref }}{{ else }}registry.dagger.io/engine:v{{ .Chart.Version }}{{ end }}
```
jedevc, line 47: May also be worth release-noting this change as well, from `tag` to `ref`?

helm/dagger/values.yaml
```diff
@@ -47,25 +54,37 @@ engine:
...
   #         - key: actions-runner
   #           operator: Exists
+
   ### Set priorityClassName to avoid eviction
   priorityClassName: ""
+
   readinessProbeSettings: 
     initialDelaySeconds: 5
-    timeoutSeconds: 30
+    timeoutSeconds: 14
```
jedevc, line 63: :eyes: 14 seconds?

gerhard: image

gerhard:
Can I get your 👀 on this @matipan @marcosnils ?

helm/dagger/values.yaml
```diff
@@ -14,12 +15,14 @@ engine:
   #   [log]
   #     format = "json"
   labels: {}
+
   ### Customize Dagger Engine start args
-  # "--oci-max-parallelism num-cpu" is kept as the default behaviour so that we don't surprise users.
-  # You may want to remove it in your deployments. FTR: https://github.com/dagger/dagger/pull/7395
+  #   "--oci-max-parallelism num-cpu" is kept as the default behaviour so that we don't surprise users.
+  #   You may want to remove it in your deployments. FTR: https://github.com/dagger/dagger/pull/7395
   args:
```
matipan, line 22: Now that we are doing a breaking change, should we bite the bullet and remove the default `--oci-max-parallelism` setting we have here? We have been running without it for a while now 

cc @gerhard @jedevc

helm/dagger/templates/engine-daemonset.yaml
```diff
@@ -74,7 +74,7 @@ spec:
                 - sh
                 - -exc
                 - |-
-                  echo '{ version }' | dagger query
+                  dagger core version
```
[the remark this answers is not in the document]
gerhard: Yes: https://github.com/dagger/dagger/pull/8135#issuecomment-2321526180

image

helm/dagger/values.yaml
gerhard: We most definitely should. Would like to keep it as a follow-up PR, so that we don't change too much at once.

PR #8353, opened 2024-09-05
chore: fix .git introspection in remote module

shykes:
When loading the module from a git remote, .git is not available. This caused our default 'dagger call -m github.com/dagger/dagger` to fail for every function call.


levlaz:
LGTM :shipit:

grouville (approved):
Lgtm

PR #8355, opened 2024-09-06
Use improved monorepo support for our own CI

shykes:
As of 0.13, Dagger has much better support for monorepos. It's now realistic to split up a large repo into several Dagger modules, one for each logical component in the repo. We should apply this pattern to our own repo, for several reasons:

- It will make our pipeline code simpler. This is even more true for us, since we adopted Dagger earlier than most, and have accumulated some cruft.
- It will make it easier to contribute, by making our pipelines easier to discover and call
- It will make it faster, especially with the benefits of pre-call filtering
- It will allow us to experiment with best practices - what's the best way to leverage these new features? Let's find out!

## Status

This is still a draft, not fully functional. BUT some of the spun-out submodules already work.

## Modules

- `cmd/dagger`: the Dagger CLI
- `docs`: the Dagger docs
- `cmd/*`: other command-line tools
- `sdk/*`, `sdk/*/dev`: Dagger SDKs.
- `helm`: the Dagger helm chart
- `engine`: the Dagger engine

- `modules/go`: a reusable Go module
- `modules/version-info`: a reusable module for managing versions in a repo
- `modules/supermod`: a reusable module for managing dagger modules (and in particular, their codegen)
- `modules/types`: a stopgap module for defining shared interfaces (workaround for limitations in our interface support)

## Examples

`dagger call -m github.com/dagger/dagger/cmd/dagger@pull/8355/head binary --platform=current`: build the Dagger CLI for your current platform

`dagger call -m github.com/dagger/dagger/cmd/dagger@pull/8355/head reference contents`: generate the CLI reference in markdown, and print it

`dagger call -m github.com/dagger/dagger/sdk/go@pull/8355/head test`: test the Go SDK


version/main.go
jedevc, line 1: I'm not a huge fan of this module :cry: 

I think instead of having this as a module that everything pulls in, the `version` should be computed at the *very* top level, in the constructor, and then passed down to everything that needs it.

Having it split out and take different arguments, and allowing the caller to potentially send in different `.changes` directories/etc, tags, commits, means that potentially we'll get different components who generate different versions (which could cause *miserable* bugs, this was really difficult to get right in the first place).

If this *is* a standalone module, it shouldn't be possible to pass it any args - it should return the exact same value every single time.

version/main.go
```diff
@@ -12,65 +17,102 @@ import (
...
+func New(ctx context.Context,
+	// A directory containing all the inputs of the artifact to be versioned: the Dagger engine + CLI
+	// An input is any file that changes the artifact if it changes.
+	// This directory is used to compute a digest. If any input changes, the digest changes.
+	// - To avoid false positives, only include actual inputs
+	// - To avoid false negatives, include *all* inputs
+	// +optional
+	// +defaultPath="/"
+	// +ignore=[".git", "bin", "**/node_modules", "**/testdata", "**.changes", "docs", "helm", "release", "version", "modules", "*.md", "LICENSE", "NOTICE", "hack"]
+	inputs *dagger.Directory,
```
jedevc, line 32: Fun thought - what's weird is that we almost don't *want* a user to be able to override this contextual directory.

I wonder if we should have another helper here, something like `// +contextual` - any contextual arg *must* come from the context, and can't be overridden (this potentially helps solve my issue above).

jedevc (commented):
Huge huge huge thanks for getting this started, and kicking off the discussion: I'm really optimistic about getting some changes around this in soon :tada: :tada:

Left some higher level thoughts - I think I like some of the smaller splits, but I'm a bit conflicted, sorry. I think the module split out of `sdk/go` is important and we need to do that one, making go like the other sdks needs doing.

> [!NOTE]
>
> Bit of a brain dump below, so a quick TL;DR:
> - Some module splits go a bit too far - `runc`/`dumb-init`/etc. These aren't dagger, they're implementation details, I think we should keep these in the `engine` build (since they're only going to be used there).
> - There are splits of modules across two axes - functionality (e.g. docs/release) and components (e.g. engine/cli). I think we should decide on one.
> - Adding more modules slightly degrades the development/understanding experience right now - this is because of flaws in *dagger*, not in this PR. I think we should work out what those all are, and make issues/plans to fix those and schedule that work (and discuss if we need to block on that, which might be controversial, sorry!).
> - I think we could simplify some of this work by passing more `interface`s around (like we do with `Sidecar`), instead of relying on separate module - this could help us split this work out more, so we can see smaller, more incremental improvements.

But, because of the knock-on effect that this has, and all the modules needing to be split, I think this makes things more complicated than it was before:

- There's now so many modules - jumping between modules in an editor/IDE to understand how everything works is a very bad experience at the moment :cry: (but has been "alright" with only a few modules). Users who arrive to try and understand our CI as a golden example are going to struggle to understand how these pieces fit together :thinking: 
	- I think we should work to solve this "jumping" problem before we split out this dramatically, otherwise, our golden example will be really hard to understand, and might scare users away (tbf, today that's also the case, but we should aim to improve the current state)
	- But even if we do make it easier to jump around, it's still split out in a way that makes it harder to see the *full* build in one place.
- The build of the engine container (which contains a lot of components) is now very split out - previously, it was pretty much all in one file `builder.go` (which agreed, did need refactoring), but it was still the build of one component. With it all split out, changing how the build works means looking all over for how runc/dumb-init/codegen/individual go binaries, etc are built.
	- As much as possible, we should try and map modules to *dagger* components - `runc` isn't a dagger component, it's an implementation detail, and so shouldn't have a module to itself.
- There's a mix of module splitting - some modules are split as functionality (like `docs` and `release`), while other modules are split as components (like `engine` and `cli`). I think these are mutually exclusive, and we should split by functionality - like `build`/`docs`/`release`/etc, or by component - like `engine`/`cli`/`sdk`s/etc (my preference).
	- I think these are exclusive, because suppose I have a component like the `CLI`, that has the cli reference docs associated with it - where does the functionality for this go? Does it go with the `docs` (so all the docs can be understood as a group), or with the `cli` (so the entire CLI can be understood at a glance).

Some other thoughts, not really related to complexity:

- There's *so* many `go.mod`/`go.sum`/`.gitattributes` now :broom: This is really weird - I know we've previously rejected the idea of reusing the top-level `go.mod` (https://github.com/dagger/dagger/pull/8095), but duplicating this everywhere feels wrong as well.
	- Again, we should solve this problem separately, it feels very important for this work.

One thing I noted while playing around with this is that we have two "mechanisms" for passing things around (e.g. how does the `sdk` component get access to the `engine`/`cli` component so it can generate the docs for it):
- Interfaces
	- Pros: keeps control in the caller, can be dynamically switched out
	- Cons: not well supported in every language, lots of usability issues, docs problems
- Splitting out the thing that needs passing into it's own module, and installing it as a dependency
	- Pros: separated out as a standalone component
	- Cons: creates complex dependency graphs, complicated to jump around

I think we could make this improved modularization a bit easier if we rely more on interfaces, like we currently do with `Sidecar` (I like this one a lot). If we did this, then we don't need to fully split out the `build` package right away, we can return a `Build` object or similar, that has a `CLI` and `Engine` method on it - then, we can pass this `Build` into SDKs/docs/helm/etc modules, and allow splitting these out first.

modules/dirdiff/dirdiff.go
```diff
@@ -39,3 +41,48 @@ func (dd *Dirdiff) AssertEqual(
...
 }
+
+// Search for a filename in a directory and return paths of matching parent directories
+func (dd *Dirdiff) Find(
+	ctx context.Context,
+	dir *dagger.Directory,
+	filename string,
+	// +optional
+	exclude []string,
+) ([]string, error) {
```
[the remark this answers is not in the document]
shykes: Yes I plan on renaming dirdiff to "superdir" or something equivalent. It's really about an improved directory type.

version/main.go
shykes: I agree it shouldn't receive args (except for the purposes of default values eg. inputs directory). But to achieve that, it needs a way to introspect git metadata. Remember that `gitDir` argument in main that you don't like? :) That's what it's for.

version/main.go
shykes: cc @vito @sipsma we bent over backwards in the context directory design to protect the argument system - make sure everything is an argument no matter what. I'm flagging this suggestion because if many users ask for the same thing, it may be a sign that we got it wrong?

vito: I'm not seeing why we wouldn't want this as an argument. 🤔 Especially after #8436 which would mean the `+ignore` rule is respected for non-contextual args too. Maybe that's the difference-maker?

shykes: @vito the fear in this particular scenario, is that different callers might pass different input directories (because they can), resulting in different version strings within the same build, which might break things in horrible ways - which @jedevc would then have to debug :)

gerhard:
Notes to self (September 23, 2024 sync):
- Do our pipelines behave the same after this change?
- Can we split this into smaller PRs?
- Can we introduce this **alongside** the existing setup, same as we do for all infra changes?

PR #8356, opened 2024-09-06
fix: allow engine entrypoint to fail to set ulimit fd limit

### LINKED ISSUE #8210: 🐞 Engine cannot start: `sh: error setting limit: Operation not permitted`
### What is the issue?

Since v0.12.5, the engine fails to start with this error in the logs:
```
#!/bin/sh
set -e

cat $0

# cgroup v2: enable nesting
# see https://github.com/moby/moby/blob/38805f20f9bcc5e87869d6c79d432b166e1c88b4/hack/dind#L28
if [ -f /sys/fs/cgroup/cgroup.controllers ]; then
	# move the processes from the root group to the /init group,
	# otherwise writing subtree_control fails with EBUSY.
	# An error during moving non-existent process (i.e., "cat") is ignored.
	mkdir -p /sys/fs/cgroup/init
	xargs -rn1 < /sys/fs/cgroup/cgroup.procs > /sys/fs/cgroup/init/cgroup.procs || :
	# enable controllers
	sed -e 's/ / +/g' -e 's/^/+/' < /sys/fs/cgroup/cgroup.controllers \
		> /sys/fs/cgroup/cgroup.subtree_control
fi

# expect more open files due to per-client SQLite databases
# many systems default to 1024 which is far too low
ulimit -n 1048576

exec /usr/local/bin/dagger-engine --config /etc/dagger/engine.toml "$@"
sh: error setting limit: Operation not permitted
```

I’m using Podman and I guess the issue only happens with Podman.

### Dagger version

dagger v0.12.5 (registry.dagger.io/engine:v0.12.5) linux/amd64

### Steps to reproduce

```
mkdir test
cd test
dagger init
```

### Log output

_No response_

samalba:
Dagger v0.12.5 added a ulimit call in the engine entrypoint, breaking podman running in rootless.

This allows the call to fail with a warning (the ulimit can be done manually by the user if needed).

It's tricky to test because the latest version of dagger (v0.12.7) does not run on podman rootless (both root mode VMs do not share any data), so you cannot build the engine in rootless to test this change. I had to build the engine in rootful, export the `dagger-engine.dev` container, export it from a rootless VM and check the logs that the ulimit gracefully failed.

Fixes #8210

### DIFF
--- a/.dagger/config.go
+++ b/.dagger/config.go
@@ -40,7 +40,7 @@ fi
 
 # expect more open files due to per-client SQLite databases
 # many systems default to 1024 which is far too low
-ulimit -n 1048576
+ulimit -n 1048576 || echo "cannot increase open FDs with ulimit, ignoring"
 
 exec {{.EngineBin}} --config {{.EngineConfig}} {{ range $key := .EntrypointArgKeys -}}--{{ $key }}="{{ index $.EntrypointArgs $key }}" {{ end -}} "$@"
 `


jedevc (approved):
LGTM :tada:

(I put this in the milestone, since it's a simple change we can merge whenever, and it will be good to get it into whatever the next release is)

vito (approved):
Thanks! Works well enough to put the fire out 👍 - we'll see if we get issues with "too many open files"

Issue #7428, opened 2024-05-20
Cache function calls by default


shykes:
## Problem

Dagger has great caching, but Dagger Functions don't fully benefit from it, because their runtime containers are not cached.

This has several consequences:

1. Functions that perform compute-intensive work themselves, via native code, are most affected: they are re-run each time, adding a major performance penalty.
2. Functions that are not compute-intensive (the majority) are less affected, but still pay a small performance tax at each execution, because of the overhead of executing the runtime container each time. For complex pipelines this can add up.
3. Module initialization is also slower, because SDKs themselves are modules and calling the initialization functions incurs the same performance tax.

The result is that most `dagger call` commands are noticeably slower than they would be, because of insufficient caching.

## Solution

Cache function calls by default!

That would be trivial to do, there's just one problem...

### Cache invalidation

Sometimes we want a function call to *not* be cached, for a variety of reasons:

- Perhaps the function is not "pure" and will not work properly if cached
- Perhaps the function *can* be cached, but the operator wants to control when to cache it and for how long]
- Perhaps the function requires custom caching behavior - for example a dependency lookup might be useful to persist in a lockfile accessible to the user

In order to cache function calls by default, we need a minimal design to allow disabling that caching. That design should 1) be as simple as possible so we can ship fast - but 2) be future-proof so that we don't need to break it later.

May the bikeshedding begin :)

shykes:
FYI @sipsma @vito here is a penciled in proposal based on the discord discussion. I kept the word "pure" out of it, because I think it might more loaded than we originally thought (ie. from a functional programming purist's perspective, none of these cacheable functions qualify as "pure". Perhaps best not to kick the hornet nest...)

sipsma:
In the interest of being able to ship something quickly and then iterate, I think it's worth considering starting with TTL settings as being defined as properties of the function being invoked (i.e. by the author, rather than the caller).

Starting there is relatively straightforward implementation-wise. You could do it without having to go all the way into lock files or anything like that; could just re-use existing cache mechanisms (just need to scope cache to a time interval in the same way you already can do that in "user-space", just automatically now).

Supporting the caller setting the TTL opens up a pandora's box:
1. For the caching to work as expected for each caller (who may now all call the same function with different TTLs), I think we'd need to support per-client lock files (and some way of specifying which one to use)
   * on top of the previously discussed trickiness of lock files like pruning
2. Figure out how to implement cases like "multiple calls in the same session that call the same function with different TTLs"
[...]

PR #8358, opened 2024-09-06
Fix secret transfer issues between modules

jedevc:
This PR contains a few fixes, extracted out of #8149, and others.

Added two changelogs, since there are two distinct fixes here, but not worth splitting them out into separate PRs, since they're both part of the same effort here :laughing: 

Both issues were introduced in https://github.com/dagger/dagger/pull/7804.

---

There's a rather nasty bug with `// +private` fields (and the equivalent in python/typescript). If an ID is stored in there, we have no way of knowing it's there.

As discussed [in discord](https://discord.com/channels/707636530424053791/1281246101487353917), it's not feasible to expose each private field to the API - there may be data in there that cannot be exposed to the API (e.g. today maps are a big case), but there may definitely be others. For backwards-compat reasons, it's probably way too much effort to try and expose the structure of those private fields - we've made our bed, now we have to lie in it.

This PR is a draft idea to do a "best-effort" detection - we look through the returned JSON objects, and scan for things that look enough like IDs - then we try and grab secrets and sockets from there. This resolves the issue, but I'm also not entirely convinced it's the right approach - it feels very hacky, and I'm somewhat worried we might misinterpret something as an ID when it shouldn't be.

Some alternative ideas I've considered (ranging from worse->better, easier->harder):
- Update all IDs to have a prefix - something like `id:` - this hopefully reduces false positives, but also strings could legitimately have this prefix.
- Don't encode IDs passed to modules as strings, encode them as a special object, something like `{ "_dagger.id": "<id>" }`. Again, we're just reducing false positives, modules could definitely still produce these.
- Dump all the IDs that we know about to a side-channel (any time we get an ID, record it, then when we're traversing fields, we know exactly what we're looking for)
- Roll our own module interchange format (or find one) that allows for typed data. This is very tricky, since it needs to be supported in each language natively.

---

Also, we have a caching bug, described in more detail [here](https://github.com/dagger/dagger/pull/8358#issuecomment-2346006330), which we also fix here as well.

### DIFF
--- a/.changes/unreleased/Fixed-20240913-134016.yaml
+++ b/.changes/unreleased/Fixed-20240913-134016.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: |
+  Allow private secrets to pass between different modules
+time: 2024-09-13T13:40:16.517262613+01:00
+custom:
+  Author: jedevc
+  PR: "8358"
--- a/.changes/unreleased/Fixed-20240913-134354.yaml
+++ b/.changes/unreleased/Fixed-20240913-134354.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: |
+  Handle session-wide cached functions that return secrets
+time: 2024-09-13T13:43:54.343379515+01:00
+custom:
+  Author: jedevc
+  PR: "8358"
--- a/core/container_exec.go
+++ b/core/container_exec.go
@@ -335,6 +335,10 @@ func (container *Container) Stderr(ctx context.Context) (string, error) {
 	return container.metaFileContents(ctx, buildkit.MetaMountStderrPath)
 }
 
+func (container *Container) usedClientID(ctx context.Context) (string, error) {
+	return container.metaFileContents(ctx, buildkit.MetaMountClientIDPath)
+}
+
 func (container *Container) metaFileContents(ctx context.Context, filePath string) (string, error) {
 	if container.Meta == nil {
 		return "", fmt.Errorf("%w: %s requires an exec", ErrNoCommand, filePath)
--- a/core/enum.go
+++ b/core/enum.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/dagger/dagger/dagql"
 	"github.com/dagger/dagger/dagql/call"
+	"github.com/dagger/dagger/engine/server/resource"
 	"github.com/dagger/dagger/engine/slog"
 	"github.com/opencontainers/go-digest"
 	"github.com/vektah/gqlparser/v2/ast"
@@ -66,7 +67,7 @@ func (m *ModuleEnumType) ConvertToSDKInput(ctx context.Context, value dagql.Type
 	return decoder.DecodeInput(value)
 }
 
-func (m *ModuleEnumType) CollectCoreIDs(ctx context.Context, value dagql.Typed, ids map[digest.Digest]*call.ID) error {
+func (m *ModuleEnumType) CollectCoreIDs(ctx context.Context, value dagql.Typed, ids map[digest.Digest]*resource.ID) error {
 	return nil
 }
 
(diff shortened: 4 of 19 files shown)


jedevc:
Hm, even the `// +private` removal doesn't solve everything.

Turns out, we can very easily hit this case: https://github.com/dagger/dagger/pull/7804#issuecomment-2216493990 (and I've pushed a test for it in https://github.com/dagger/dagger/pull/8358/commits/1b8c1758faea4171b0d44c02b8313d6d1dfcb88b). We incorrectly assume that if a call is cached (and so `setSecret` never ran), then it must already be in our secret store.

This isn't the case in the following scenario (which I do in the test):
- Client 1 calls `Foo` and `Bar` functions in module `Test` (in parallel, in serial, it doesn't matter)
- Client 2 (during evaluation of `Foo`) calls function `Mount` in module `Dep`, which in turn creates Client 3, calls `setSecret` and returns the `Secret`. After evaluation, we inspect the ID, and load the secret from Client 3 into Client 2.
- Client 4 (during evaluation of `Bar`) calls function `Mount` again, but this time it's cached - no `setSecret` happens. We still get the same ID though, but when we try and load the secret, it doesn't exist in `Mount`s client (Client 5, which never gets used).

This is very fiddly, and I'm not 100% sure what the fix to this looks like.

*Even* if we actually perform the ID load as suggested in https://github.com/dagger/dagger/pull/7804#issuecomment-2234090051 (using a different dagql server, and correctly mock a fake client so that we can actually correctly load the secret into Client 5, we hit the issue that `setSecret` isn't even *in* the ID, so we'd never actually insert it into Client 5 (we'd just be trying to access it, which would now fail, since it hadn't been created in this secret store). I think as long as we keep this separation where we return a `secret` instance from `setSecret`, this isn't fixable using that technique.

Another option (this one might work :tada:), could be to have a `Call` additionally return the client ID that it actually used. This is neat, because now when we cached, we would ignore the randomly generated ID for Client 5, and could instead get the ID of Client 4 and load secrets from there. This still doesn't solve the generalized caching of function calls #7428, but gets us over this hurdle temporarily.

However, it feels to me that if we want to solve the general caching for #7428, we're going to need to make `setSecret` "super-impure" (name tbd :laughing:) - if a function calls it, then we cannot cache that result. This is a bit trickier than impure currently is, because it passes on it's impurity, and we don't know if a function did call this until *after* it ran. This makes some amount of sense to me - e.g. if a `Secret` gets cached without a corresponding `setSecret` anywhere, we can't re-evaluate anything that depends on it - you then get into this weird state where re-builds may work without a valid secret, or may not, entirely dependent.

Gonna have a bit more of a think about this, I'm leaning towards the final option, since I think we're gonna need to do it anyways one day.

jedevc:
> Another option (this one might work 🎉), could be to have a `Call` additionally return the client ID that it actually used. This is neat, because now when we cached, we would ignore the randomly generated ID for Client 5, and could instead get the ID of Client 4 and load secrets from there.

This turns out to be very easy to implement, with our new worker setup - we just write a special meta file to record this.

We also need to disable dagql-level caching of function calls, which is sad, but we need more complex logic to handle the secret transfers, and if we keep this enabled, then we have no-where to put it.

> This is a bit trickier than impure currently is, because it passes on it's impurity, and we don't know if a function did call this until after it ran.

This is fiddly. I don't really have any good ideas for how we would even do this - so deferred this to some generic point in the future.

sipsma (approved):
LGTM (just some nit pick comments), thanks for tracking all this down while I was away!

> Another option (this one might work 🎉), could be to have a Call additionally return the client ID that it actually used. This is neat, because now when we cached, we would ignore the randomly generated ID for Client 5, and could instead get the ID of Client 4 and load secrets from there. This still doesn't solve the generalized caching of function calls https://github.com/dagger/dagger/issues/7428, but gets us over this hurdle temporarily.

Nice, I like that idea a lot! As for the longer term problem with function caching, `SetSecret` indeed is a known thorn that will pop up with all of that. Users will just have to mark functions that make `SetSecret` calls as uncached.
* In the somewhat longer-term I think we'll just end up implementing secret providers (there's an issue for this somewhere, I can't find it rn though), which will allow us to deprecate the nasty side-effectful `SetSecret` and replace it with something more interoperable with caching and just generally more comprehensible to work with.

PR #8362, opened 2024-09-06
fix: context directory ignore discrepency

### LINKED ISSUE #8287: Discrepancy between our +ignore and gitignore
## Problem

Our new `+ignore` pragma (for directory arguments) should be compatible with the gitignore format (for simplicity to the user, and easier interop), but currently it's not.

A few examples:

| Pragma | Expected behavior | Actual behavior |
| -- | -- | -- |
| `+ignore=["**", "!.github/workflows"]` | everything is ignored except `.github/workflows` | everything is ignored |
| `+ignore=["!.github/workflows"]` | nothing is ignored | everything is ignored except  `.github/workflows` |

## Solution

Change the behavior of our `+ignore` to be fully compatible with `.gitignore`.

## Design considerations

As explained by @TomChv , the difficulty is that buildkit APIs expose an `include`/`exclude` model that doesn't map directly to Git ignore.

I would like to hear more about how much difficulty we're talking about exactly. This is an important aspect of the spec, which is why I addressed the possible difficulty early:

[On Discord, June 27](https://discord.com/channels/707636530424053791/1003718839739105300/1255782856899301457)
>  Don't underestimate the ignore though, it doesn't map directly to llb.local include/exclude. I guess it's the exact opposite of patterns from views though, maybe that will help?
>  I mean you will need to convert gitignore format to buildkit include/exclude format

TomChv:
Remove `Include` usage in context directory to only use excludes negation since it seems it works.
Add a couple of tests to verify that behaviour.

Fixes #8287

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -4504,140 +4504,140 @@ import (
 type Test struct{}
 
 func (t *Test) IgnoreAll(
-	// +ignore=["**"]
-	// +defaultPath="."
-	dir *dagger.Directory,
+  // +ignore=["**"]
+  // +defaultPath="."
+  dir *dagger.Directory,
 ) *dagger.Directory {
-	return dir
+  return dir
 }
 
 func (t *Test) IgnoreThenReverseIgnore(
-	// +ignore=["**", "!**"]
-	// +defaultPath="."
-	dir *dagger.Directory,
+  // +ignore=["**", "!**"]
+  // +defaultPath="."
+  dir *dagger.Directory,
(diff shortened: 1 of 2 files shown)


shykes (approved):
Thank you! Didn't review the implementation, but assuming it works as advertised: SHIPT

TomChv:
Only [scan failing](https://dagger.cloud/dagger/traces/9051eef4235f6b1aa3d33e66b1ef646c) but it's related to alpine and should be address in a seperate PR

PR #8365, opened 2024-09-07
core: make `SSH_AUTH_SOCK` absolute + handle shell expansion in `SSH_AUTH_SOCK` env var

### LINKED ISSUE #8337: core(private-modules): `SSH_AUTH_SOCK` env loading seems to break with shell expansion characters
Bug found: https://discord.com/channels/707636530424053791/1280527928861655112/1280958016144932885

I believe it is linked to the lexical relative path function: https://github.com/dagger/dagger/blob/main/engine/client/pathutil.go#L9-L12

grouville:
Fixes #8337 AND fixes some relative path issues

This PR addresses issues with SSH_AUTH_SOCK handling and improves test coverage:
1. Expands shell variables in SSH_AUTH_SOCK path
2. Make SSH_AUTH_SOCK path absolute
3. Adds integration tests for shell expansion handling and relative path calls

https://discord.com/channels/707636530424053791/1281294189463732244/1281294189463732244

### Detailed explanation for not making SSH_AUTH_SOCK path relative:

The decision to avoid making the SSH_AUTH_SOCK path relative is based on several factors:

Users have been complaining of some parsing issues regarding the SSH_AUTH_SOCK. After some digging in the codebase, it seems better not to handle it relatively:

#### 1. Client - Host interaction

On the engine side, we [URL encode](https://github.com/dagger/dagger/blob/main/core/socket.go#L174-L189) the socket Path in order to forward it from the client to the engine
    
On the client side, we decode it with [url.Parse()](https://github.com/dagger/dagger/blob/main/engine/client/socket.go#L40) and retrieve the [u.Path](https://github.com/dagger/dagger/blob/main/engine/client/socket.go#L46)
    
This operation is destructive of all relative Path: ../.test/socket becomes /.test/socket

This leads to some bugs such as the one the user got in the thread
    
#### 2. Nested execs

In order to handle nested execs, we rely on [UnixPathMapper](https://github.com/dagger/dagger/blob/main/engine/client/socket.go#L48) to retrieve the socket at the path inside the rootfs.

The current implementation [does not seem to expect relative paths](https://github.com/dagger/dagger/blob/main/engine/buildkit/executor_spec.go#L948-L951)
 
 
cc @jedevc I am not sure about the UnixPathMapper implementation. This was the main reason I preferred to rollback to an absolute path. The first implementation relied on encoding / decoding the path with `url.PathEscape()`: https://go.dev/play/p/hj8PeEwtUfL. I also added a test suite to test those relative path handling, in case UnixPathMapper is not necessarily expecting an absolute path

### DIFF
--- a/engine/buildkit/executor_spec.go
+++ b/engine/buildkit/executor_spec.go
@@ -901,8 +901,14 @@ func (w *Worker) setupNestedClient(ctx context.Context, state *execState) (rerr 
 	w.execMD.ClientStableID = randid.NewID()
 
 	// include SSH_AUTH_SOCK if it's set in the exec's env vars
-	if v, ok := state.origEnvMap["SSH_AUTH_SOCK"]; ok {
-		w.execMD.SSHAuthSocketPath = v
+	if sockPath, ok := state.origEnvMap["SSH_AUTH_SOCK"]; ok {
+		if homeDir, ok := state.origEnvMap["HOME"]; ok {
+			expandedPath, err := client.ExpandHomeDir(homeDir, sockPath)
+			if err != nil {
+				return fmt.Errorf("failed to expand homedir: %w", err)
+			}
+			w.execMD.SSHAuthSocketPath = expandedPath
+		}
 	}
 
 	filesyncer, err := client.NewFilesyncer(
--- a/engine/buildkit/executor_spec.go
+++ b/engine/buildkit/executor_spec.go
@@ -902,12 +902,18 @@ func (w *Worker) setupNestedClient(ctx context.Context, state *execState) (rerr 
 
 	// include SSH_AUTH_SOCK if it's set in the exec's env vars
 	if sockPath, ok := state.origEnvMap["SSH_AUTH_SOCK"]; ok {
-		if homeDir, ok := state.origEnvMap["HOME"]; ok {
-			expandedPath, err := client.ExpandHomeDir(homeDir, sockPath)
-			if err != nil {
-				return fmt.Errorf("failed to expand homedir: %w", err)
+		if strings.HasPrefix(sockPath, "~") {
+			if homeDir, ok := state.origEnvMap["HOME"]; ok {
+				expandedPath, err := client.ExpandHomeDir(homeDir, sockPath)
+				if err != nil {
+					return fmt.Errorf("failed to expand homedir: %w", err)
+				}
+				w.execMD.SSHAuthSocketPath = expandedPath
+			} else {
+				return fmt.Errorf("HOME not set, cannot expand SSH_AUTH_SOCK path: %s", sockPath)
 			}
-			w.execMD.SSHAuthSocketPath = expandedPath
+		} else {
+			w.execMD.SSHAuthSocketPath = sockPath
 		}
 	}
 
--- a/engine/client/client.go
+++ b/engine/client/client.go
@@ -1034,10 +1034,15 @@ func allCacheConfigsFromEnv() (cacheImportConfigs []*controlapi.CacheOptionsEntr
 
 func (c *Client) clientMetadata() engine.ClientMetadata {
 	sshAuthSock := os.Getenv("SSH_AUTH_SOCK")
-
-	expandedPath, err := expandPath(sshAuthSock)
+	// expand ~ into absolute path for consistent behavior with CLI
+	// ⚠️ When updating clientMetadata's logic, please also update setupNestedClient
+	// for consistent behavior of CLI inside nested execution
+	homeDir, err := os.UserHomeDir()
 	if err == nil {
-		sshAuthSock = expandedPath
+		expandedPath, err := ExpandHomeDir(homeDir, sshAuthSock)
+		if err == nil {
+			sshAuthSock = expandedPath
+		}
 	}
 
 	clientVersion := c.Version
--- a/engine/client/client.go
+++ b/engine/client/client.go
@@ -1033,18 +1033,11 @@ func allCacheConfigsFromEnv() (cacheImportConfigs []*controlapi.CacheOptionsEntr
 }
 
 func (c *Client) clientMetadata() engine.ClientMetadata {
-	// retrieve SSH_AUTH_SOCK path and make it relative
 	sshAuthSock := os.Getenv("SSH_AUTH_SOCK")
-	if sshAuthSock != "" {
-		expandedPath, err := expandPath(sshAuthSock)
-		if err == nil {
-			sshAuthSock = expandedPath
-			if cwd, err := os.Getwd(); err == nil {
-				if relPath, err := LexicalRelativePath(cwd, sshAuthSock); err == nil {
-					sshAuthSock = relPath
-				}
-			}
-		}
+
+	expandedPath, err := expandPath(sshAuthSock)
+	if err == nil {
+		sshAuthSock = expandedPath
 	}
 
 	clientVersion := c.Version
--- a/engine/client/client.go
+++ b/engine/client/client.go
@@ -1036,9 +1036,13 @@ func (c *Client) clientMetadata() engine.ClientMetadata {
 	// retrieve SSH_AUTH_SOCK path and make it relative
 	sshAuthSock := os.Getenv("SSH_AUTH_SOCK")
 	if sshAuthSock != "" {
-		if cwd, err := os.Getwd(); err == nil {
-			if relPath, err := LexicalRelativePath(cwd, sshAuthSock); err == nil {
-				sshAuthSock = relPath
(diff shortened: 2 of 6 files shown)


grouville:
I am actually super confused about this failure. Unable to repro locally. 

It seems like a race, but happening on the nested exec, where the path is given by UnixPathMapper

```
Error: failed to get configured module: failed to get module ref kind: input: moduleSource resolve: failed to resolve git src to commit: failed to load cache key: NotFound: NotFound: rpc error: code = NotFound desc = socket /.ssh-sock not found: stat /tmp/rootfs1116108047/.ssh-sock: no such file or directory
```

grouville:
So, the PR correctly fixes several bugs encountered regarding the socket path (shell expansion AND relative path issues). It just needs ONE change to be all green on the test-suite, but the logic is already here.

I need some help on finding the right primitive to unlock last point though, which only touches nested executions

Here’s the full context dump (other issues + the current one that the issue faces):


1. New inputs from users + further investigations do confirm that the socket path cannot be stored relatively as it is right now. The relative path is not handled by UnixPathMapper at all (and confirms why the isAbs check is present when entring this function)

Given these informations, relative paths never worked, but I never saw it in my tests, nor on the first use. Some context:

When running a dagger command, the client metadata is collected and sent to the engine
When we collect the client metadata, currently, the LexicalRelativePath function is used to generate a relative path from the current workdir (which is the workdir of dagger CLI when calling a dagger command) to the value of the SSH_AUTH_SOCK. Upon tests, I confirm that the paths returned by LexicalRelativePath are totally correct.
We store this computed relative path inside the clientMetadata into the clientMetadata.sshAuthSockPath field. One important thing to note is that this clientMetadata is passed via the context and have helper functions (setters and getters). It is used a bit everywhere in the code. This is our source of truth to locate the path of the socket when the engine will want to request it to the client, usually in a sshforward context.
But, for a reason explained right below, it occurs that relative path never worked, so, how did it bypass my / our tests ?

Basically, the way I tested the setup was to rely on the path generated by my ssh-agent, which is generally located in the /var/run/sockXXXX.agent. Meaning that when I was testing, I was testing in the HOME or /tmp or any OTHER path than /var/run. This led to the generation of a path such as ../../../var/run/sockXXXX.agent


Due to another serialization bug explained in the PR context above (or in the second point of the list of issues), when the server requests to the client, all the `../../..` are stripped, generating a valid absolute path: /var/run/sockXXXX.agent.

This is why only a handful of users are complaining at the moment: it is mostly users that use 1password to authenticate the socket or all the users that setup symlinks who generally have a location of the socket which is inside the HOME dir or in any subdirectories of HOME, leading to incomplete relative paths such as: ./subdir/sockXXXX.agent which becomes /subdir/sockXXXX.agent which is not valid when checking if the agent is present

2. The serialization bug is not fixed in the current state of the PR, but will be in a follow-up as it is more of an edge case.

Following a conversation / debug with a user today, it has been confirmed that we need to url.EscapePath the unix Path to avoid URL encoding spaces and special chars as URL chars (in the unix context). (Today, a user had its path with a space inside being converted to %20). Links are also present in the readme / first comment of the PR, to see where it is located in the codebase.

This is a bug that was present since the creation of the core.Socket type, I am surprised that no one encountered it before.
Logic:
1. When the engine needs to have access to a forwarded socket, it sends a request to the client to do a checkAgent at a given path

2. When doing this request, we rely on a url.URL type to encode and transmit this information. Why ? Because the path could represent 2 types of sockets: unix and tcp

However, when url.URL encodes it and when we decode it client-side with url.Parse, the unix path is cleaned, meaning that we lose all those ../.. AND we transform all the special chars as URL encoded chars, leading to those bugs in the UNIX context. This will be part of a second PR, as a follow-up, as it is a bit less critical.

3. This bug is subtle and is probably the reason why we have some flaky SSH path issues in the CI (very high odds). This is the bug that currently makes this PR red, and is where I need some external help

Some context:
This PR adds a test suite of all the encountered bugs by users as of now (apart the space encoding). When reproducing them locally, it works BUT, inside the test suite it bugs. Why ?

Logic:

There is a setupNestedClient function that sets everything for the nested execution. Inside, we forward the value of the SSH_AUTH_SOCK to the nested engine
This setupNestedClient also sets up the unixPathMapper function, whose role is to map the path of the socket to a valid path in a nested context

I just said above that the setupNestedClient « forwards » the value of the SSH_AUTH_SOCK , let’s focus a bit on it as this is where the bug occurs:

In an ideal world, what we want is for the expanded value of SSH_AUTH_SOCK to be passed down to the nested engine. Basically, what I would expect is for the value of clientMetadata.SSHAuthSockPath to be passed down as the SSH_AUTH_SOCK of the nested engine)
Now, the current implementation (which is in setupNestedClient) works a bit differently:

It does not collect this SSHAuthSockPath from the clientMetadata but retrieves the value of SSH_AUTH_SOCK present inside the execState
It passes it to the nested engine with w.execMD.SSHAuthSockPath (which is the correct way to forward it to the nested engine)

However, it turns out that this SSH_AUTH_SOCK value, when retrieved from the execState, equals to the value of the SSH_AUTH_SOCK when the client did the os.GetEnv(« SSH_AUTH_SOCK »), so even before being transformed or made relative by LexicalRelativePath.

I am not a BuildKit expert, but some testing confirmed that the seems to be only collected once by the client (at this os.getenv call). What is weirder is that I send the clientMetadata with the expanded value to the engine, so I don’t really know where this execState does collect it from (apart a cache from the os.getenv call from the client ???)

Nonetheless, the fix is theoretically simple: we need to not rely on the execState to fix it but we need to forward to the nested engine the value of the clientMetadata.SSHAuthSocketPath value to the nested engine


My problem is that I haven’t been able to retrieve it in this context; Any idea ?

gerhard (approved):
LGTM. WDYT @jedevc ?

---
The only thing missing is the changelog fragment, but we could add this part of preparing the final release notes, no need to force another re-run just because of those.

grouville:
The nested buildkit path expansion is not fixed in current state of pr, just trying to fit in the release. Will be a follow-up

engine/client/pathutil.go
```diff
@@ -56,3 +57,23 @@ func getDrive(path string) string {
...
 }
+
+// expandPath expands a given path to its absolute form, handling home directory
+// expansion (~ or ~user) and environment variable expansion. It aims to be cross-platform
+func expandPath(path string) (string, error) {
+	if path == "" {
+		return "", nil
+	}
+
+	// the ~ expansion is a unix convention
```
jedevc, line 68: `expandHomeDir` already exists - we should make that a public function and use that instead.

engine/client/client.go
```diff
@@ -1040,14 +1040,11 @@ func allCacheConfigsFromEnv() (cacheImportConfigs []*controlapi.CacheOptionsEntr
...
-	// retrieve SSH_AUTH_SOCK path and make it relative
 	sshAuthSock := os.Getenv("SSH_AUTH_SOCK")
-	if sshAuthSock != "" {
-		if cwd, err := os.Getwd(); err == nil {
-			if relPath, err := LexicalRelativePath(cwd, sshAuthSock); err == nil {
-				sshAuthSock = relPath
-			}
-		}
+
+	expandedPath, err := expandPath(sshAuthSock)
```
jedevc, line 1045: Should we now not do the opposite, and force this to be absolute?

engine/client/pathutil.go
grouville: Sure 👍

engine/buildkit/executor_spec.go
```diff
@@ -896,8 +896,14 @@ func (w *Worker) setupNestedClient(ctx context.Context, state *execState) (rerr
 	w.execMD.ClientStableID = randid.NewID()
 
 	// include SSH_AUTH_SOCK if it's set in the exec's env vars
-	if v, ok := state.origEnvMap["SSH_AUTH_SOCK"]; ok {
-		w.execMD.SSHAuthSocketPath = v
+	if sockPath, ok := state.origEnvMap["SSH_AUTH_SOCK"]; ok {
+		if homeDir, ok := state.origEnvMap["HOME"]; ok {
```
sipsma, line 900: This will result in `SSH_AUTH_SOCK` only working if `HOME` is set, even if the path doesn't have a `~` in it, right? It seems like `HOME` should only matter if there's a `~` in the `SSH_AUTH_SOCK` path.

sipsma (commented):
Noticed one thing that should be an easy fix, otherwise LGTM!

engine/buildkit/executor_spec.go
grouville: Corrected, thanks 🙏

sipsma (approved):

PR #8366, opened 2024-09-07
chore: print the cloud trace url after progress output

### LINKED ISSUE #8162: CLI: print trace URL at the bottom
## Problem

`dagger call` prints the URL to a full trace (or to traces setup, if logged out). It does so at the top of the output, which is cumbersome in practice. When I want to investigate a problem, usually it's after a fairly large and complex pipeline, so I need to scroll up past a large and complex output. After running a few of these in a row (very common), it's easy to scroll too far, and get lost: "is this the last trace, or the one before"? It adds to the confusion and gets in the way of resolving my issue.

## Solution

Print the trace at the URL instead. If I need it, it's right there.

rajatjindal:
fixes #8162

### DIFF
--- a/.changes/unreleased/Changed-20240910-191026.yaml
+++ b/.changes/unreleased/Changed-20240910-191026.yaml
@@ -1,0 +1,6 @@
+kind: Changed
+body: The trace url is printed just before the final output to make it easy to find. Also, the url will be printed only for subset of dagger commands to reduce noise.
+time: 2024-09-10T19:10:26.083774+05:30
+custom:
+    Author: rajatjindal
+    PR: "8366"
--- a/cmd/dagger/call.go
+++ b/cmd/dagger/call.go
@@ -12,6 +12,18 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	printTraceLinkKey = "printTraceLink"
+)
+
+func isPrintTraceLinkEnabled(annotations map[string]string) bool {
+	if val, ok := annotations[printTraceLinkKey]; ok && val == "true" {
+		return true
+	}
+
+	return false
+}
+
 var funcCmds = []*FuncCommand{
 	callModCmd,
 	callCoreCmd,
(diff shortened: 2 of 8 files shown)


rajatjindal:
Hi @shykes, I believe this PR is ready for review (the engine scan is failing due to a CVE caught in engine for `runc` and it does not seem related to these changes).

following are the changes as per this PR:

- Trace link will only be printed for `call`, `query`, `core` and `run` commands. (irrespective of if user is logged-in or not)
-  The trace link will be printed after the `progress` and before the final `primary` output.

Thank you

shykes (approved):
Thanks @rajatjindal! It looks good, one request (but can be done in a follow-up PR): can you make sure it honors "quiet mode" aka `-q`?

```console
$ dagger call -q github.com/shykes/hello hello
```

With `-q`, the "lingering" TUI output is wiped from the terminal, leaving only the actual output of the pipeline. I think that should apply to the trace URL also.

Thank you!

marcosnils (approved):

rajatjindal:
> can you make sure it honors "quiet mode" aka -q?

oh, that is a regression. I have now updated it to not print trace url when -q flag is provided.

gerhard (commented):
Looking at this now @rajatjindal 👍

Can you add a release note fragment please? FTR: https://github.com/dagger/dagger/blob/main/CONTRIBUTING.md#3-add-release-notes-fragment

rajatjindal:
> Can you add a release note fragment please?

@gerhard, thank you. I have added the release note. (could you please double check if that note looks ok)

many thanks

gerhard (approved):
✅ Does not show on `functions`
image

✅ Shows on `call`
image

✅ Shows on `--progress=plain`
image

✅ Does not show on `--quiet`
image

✅ Shows on `dagger core version`
image
🤔 While not technically wrong, this is our primary mechanism for checking Engine healthiness (in the context of the Helm chart). Not worth blocking the merge over this, but I can see how this could become annoying when logged into Dagger Cloud. cc @shykes 

Good to merge as soon as all checks pass 👍

PR #8369, opened 2024-09-07
feat(sdk/php): use context directory to get sdk source dir

wingyplus:


sdk/php/runtime/go.mod
```diff
@@ -38,3 +38,11 @@ require (
...
 	google.golang.org/protobuf v1.34.1 // indirect
 )
+
+replace go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc => go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc v0.0.0-20240518090000-14441aefdf88
+
+replace go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp => go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp v0.3.0
+
+replace go.opentelemetry.io/otel/log => go.opentelemetry.io/otel/log v0.3.0
+
+replace go.opentelemetry.io/otel/sdk/log => go.opentelemetry.io/otel/sdk/log v0.3.0
```
charjr, line 48: What does this do? As I noticed they are the exact versions already specified above:

https://github.com/dagger/dagger/blob/8dabfd657d299904fab8be718e6b7d06409f31a6/sdk/php/runtime/go.mod#L10-L16

charjr (commented):
I'm still quite new to the inner workings of context directories so I cannot tell if this is all that is required. 

Overall LGTM, but I think @helderco is better suited to approve these changes.

sdk/php/runtime/go.mod
gerhard: Not sure about this change. WDYT @vito?

sdk/php/runtime/main.go
gerhard, line 1: These changes look good to me! @TomChv?

gerhard (commented):
Not sure about the `go.mod` change, but otherwise LGTM.

sdk/php/runtime/main.go
```diff
@@ -25,11 +25,10 @@ type PhpSdk struct {
 func New(
 	// Directory with the PHP SDK source code.
 	// +optional
+	// +defaultPath="/sdk/php"
```
helderco, line 28: Needs the ignore patterns.

sdk/php/runtime/main.go
```diff
@@ -25,11 +25,10 @@ type PhpSdk struct {
...
 	sdkSourceDir *dagger.Directory,
 ) *PhpSdk {
 	if sdkSourceDir == nil {
-		// TODO: Replace with a *default path from context* when
-		// https://github.com/dagger/dagger/pull/7744 becomes available.
```
helderco, line 32: `if sdkSourceDir == nil` should actually return an error now, since when used as `--sdk=php` it should be provided by `defaultPath`.

helderco (commented):
With contextual directory, I think `sdk/php/dagger.json` should move back to `sdk/php/runtime`. This allows making `sdk/php/dagger.json` point to the dev module so you don't have to specify `-m` every time while developing.

Use includes (instead of excludes)  in `// +ignore`, prefix each pattern with `!`.

sdk/php/runtime/main.go
wingyplus: I think we can eliminate `sdkSourceDir == nil` now since it is already available via the defaultPath.

sdk/php/runtime/main.go
```diff
@@ -25,35 +25,10 @@ type PhpSdk struct {
 func New(
 	// Directory with the PHP SDK source code.
 	// +optional
+	// +defaultPath="/sdk/php"
+	// +ignore=["**/*", "!generated/", "!src/", "!scripts/", "!composer.json", "!composer.lock", "!LICENSE", "!README.md"]
```
wingyplus, line 29: It seems that it doesn't ignore the directory:

```
wingyplus@WINGYMOMO ~/s/g/d/dagger (php-context-dir)> with-dev dagger -m github.com/wingyplus/dagger/sdk/php/runtime@c5f72bbef56cbc5f6f70d28f29bb54f2fe5f2015 call source-dir
Setup tracing at https://dagger.cloud/traces/setup. To hide: export STOPIT=1

✔ connect 0.2s
✔ initialize 0.6s
✔ prepare 0.0s
✔ phpSdk: PhpSdk! 0.3s
✔ PhpSdk.sourceDir: Directory! 0.0s
✔ Directory.entries: [String!]! 0.0s

_type: Directory
entries:
    - .changes
    - .changie.yaml
    - .gitattributes
    - .gitignore
    - .php-cs-fixer.dist.php
    - CHANGELOG.md
    - LICENSE
    - README.md
    - composer.json
    - composer.lock
    - dev
    - docker
    - docker-compose.yml
    - generated
    - phpcs.xml
    - phpunit.xml.dist
    - psalm.xml
    - runtime
    - scripts
    - src
    - tests
 ```

What am I wrong?

File: sdk/php/runtime/main.go (hunk 3 above)
Comment (helderco) [reply sdk/php/runtime/main.go:32]: I kept it in Python but just return an error. That's because the argument is a pointer and I don't to hit any situation where it can be `nil` which would raise a panic.

File: sdk/php/runtime/go.mod (hunk 1 above)
Comment (vito) [reply sdk/php/runtime/go.mod:48]: This is from https://github.com/dagger/dagger/pull/8234

tl;dr: We still depend on some pre-1.0 versions of OTel SDK packages, and it has happened before that even though our `go.mod` initially pins them appropriately, a user running `go get -u` or similar will have them bump ahead to a version that breaks everything. So we explicitly add `replace` rules for them now.

File: sdk/php/runtime/main.go (hunk 3 above)
Comment (wingyplus) [reply sdk/php/runtime/main.go:32]: I see. Fixed on https://github.com/dagger/dagger/pull/8369/commits/ada61c18667d24329ace94bc94be148ba2054201

File: sdk/php/runtime/main.go (hunk 4 above)
Comment (wingyplus) [reply sdk/php/runtime/main.go]: After rebase main branch, the issue is now disappear.

**Comment by helderco** [comment]
> With contextual directory, I think `sdk/php/dagger.json` should move back to `sdk/php/runtime`. This allows making `sdk/php/dagger.json` point to the dev module so you don't have to specify `-m dev` every time while developing.

@wingyplus, sorry, I know I asked for this, but I'm now having second thoughts. I'm having internal discussions related to this. I'll explain more later, but please hold on for a decision in that regard.

File: sdk/php/dagger.json [hunk 5]
```diff
@@ -1,18 +1,8 @@
...
-    "LICENSE",
-    "README.md"
-  ],
-  "exclude": [
-    "vendor/"
-  ],
-  "source": "runtime",
-  "engineVersion": "v0.12.3"
+  "include": [],
+  "exclude": [],
```
wingyplus, line 5:
```suggestion
```

sdk/php/dagger.json
```diff
@@ -1,18 +1,6 @@
...
-    "src/",
-    "scripts/",
-    "composer.json",
-    "composer.lock",
-    "LICENSE",
-    "README.md"
-  ],
-  "exclude": [
-    "vendor/"
-  ],
```
helderco: Don't remove this exclude. The Go SDK adds `**/vendor` to includes, but it's meant for Go modules. However, PHP has a heavy `vendor` and will be included by default without this.

helderco (approved):

PR #8390, opened 2024-09-09
clientdb: preserve more OTLP data

vito:
Previously OTLP log data recorded into the client's SQLite database was dropping somewhat important information such as the OTel resource that generated the logs. I noticed this while investigating duplicate logs in Cloud, since half of them had a blank resource, and half of them didn't - the root cause of that ended up being https://github.com/open-telemetry/opentelemetry-go/issues/5782 but that blank resource value seemed worth fixing anyway, and led to a few other refactors which I'll explain inline.

(No behavior change is expected here - just some housekeeping)

### DIFF
--- a/engine/client/client.go
+++ b/engine/client/client.go
@@ -681,13 +681,8 @@ func (c *Client) exportLogs(ctx context.Context, httpClient *httpClient) error {
 		if err := protojson.Unmarshal(data, &req); err != nil {
 			return fmt.Errorf("unmarshal spans: %w", err)
 		}
-
-		logs := telemetry.LogsFromPB(req.GetResourceLogs())
-
-		slog.ExtraDebug("received logs from engine", "len", len(logs))
-
-		if err := c.EngineLogs.Export(ctx, logs); err != nil {
-			return fmt.Errorf("export %d logs: %w", len(logs), err)
+		if err := enginetel.ReexportLogsFromPB(ctx, c.EngineLogs, &req); err != nil {
+			return fmt.Errorf("re-export logs: %w", err)
 		}
 		return nil
 	})
--- a/engine/clientdb/log.go
+++ b/engine/clientdb/log.go
@@ -90,8 +90,8 @@ func LogsToPB(dbLog []Log) []*otlplogsv1.ResourceLogs {
 			SeverityText:   sd.SeverityText,
 			Body:           &bodyPb,
 			Attributes:     attrs,
-			TraceId:        []byte(tid[:]),
-			SpanId:         []byte(sid[:]),
+			TraceId:        tid[:],
+			SpanId:         sid[:],
 		})
 		ssm[k] = scopeLog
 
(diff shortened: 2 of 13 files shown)


engine/client/client.go
```diff
@@ -681,13 +681,8 @@ func (c *Client) exportLogs(ctx context.Context, httpClient *httpClient) error {
...
 			return fmt.Errorf("unmarshal spans: %w", err)
 		}
-
-		logs := telemetry.LogsFromPB(req.GetResourceLogs())
-
-		slog.ExtraDebug("received logs from engine", "len", len(logs))
-
-		if err := c.EngineLogs.Export(ctx, logs); err != nil {
-			return fmt.Errorf("export %d logs: %w", len(logs), err)
+		if err := enginetel.ReexportLogsFromPB(ctx, c.EngineLogs, &req); err != nil {
```
vito, line 684: Here our goal is to relay an OTLP logs payload from the receiving endpoint to another exporter, which expects a `[]sdklog.Record`. Previously we tried to construct that from the payload, but the only way to actually set the `Resource` on a `sdklog.Record` is to go through a `sdklog.LoggerProvider`. Fortunately in all cases that we want to do that we're about to send them off to a `log.Exporter` which is exactly what a `LoggerProvider` points to, so this `ReexportLogsFromPB` helper method does the conversion and exporting in one shot, and is used in both places.

engine/clientdb/log.go
```diff
@@ -2,50 +2,126 @@ package clientdb
...
-	var rec sdklog.Record
-
-	rec.SetTimestamp(time.Unix(0, dbLog.Timestamp))
-
-	rec.SetSeverity(log.Severity(dbLog.Severity))
-
-	if dbLog.TraceID.Valid {
-		tid, _ := trace.TraceIDFromHex(dbLog.TraceID.String)
-		rec.SetTraceID(tid)
+func LogsToPB(dbLog []Log) []*otlplogsv1.ResourceLogs {
```
vito, line 18: NOTE: this converts from our SQLite database `Log` model, as opposed to the `[]sdklog.Record` input you might expect from normal OTLP transformations.

Previously we went from DB model -> `sdklog.Record` -> OTLP, now we just skip the middle step.

engine/clientdb/queries.sql
vito, line 1: (slightly annoying diff, see corresponding schema)

sdk/go/telemetry/transform.go
```diff
@@ -588,13 +597,7 @@ func (s *readOnlySpan) InstrumentationLibrary() instrumentation.Library {
 
 // Resource returns information about the entity that produced the span.
 func (s *readOnlySpan) Resource() *resource.Resource {
-	if s.resource == nil {
-		return nil
-	}
```
sipsma, line 593: this nil check seems like it got dropped, not sure if important or not, just noting in case

sipsma (approved):

sdk/go/telemetry/transform.go
vito: Pretty sure the nil check is redundant - every span and log must have a resource in OTel. Will keep an eye out for panics in case I'm wrong.

PR #8396, opened 2024-09-10
update checks for the CLI

aluzzardi:
Run an update check in the background and notify if updates are available.

This is a best effort check: if not completed by the time the command completes, then the check is cancelled.

TODO:
- ~~Replace the http check with a registry check [/cc @marcosnils @gerhard]~~
- ~~Better TUI integration [/cc @vito]~~

### DIFF
--- a/.dagger/build/builder.go
+++ b/.dagger/build/builder.go
@@ -150,12 +150,7 @@ func (build *Builder) Engine(ctx context.Context) (*dagger.Container, error) {
 		})
 	}
 
-	base := dag.
-		Container(dagger.ContainerOpts{Platform: build.platform})
-	if build.version != "" {
-		base = base.WithLabel("io.dagger.version", build.version)
-	}
-
+	var base *dagger.Container
 	switch build.base {
 	case "alpine", "":
 		base = dag.
@@ -180,7 +175,7 @@ func (build *Builder) Engine(ctx context.Context) (*dagger.Container, error) {
 				ln -s /sbin/ip6tables-legacy-restore /usr/sbin/ip6tables-restore
 			`})
 	case "ubuntu":
-		base = base.
+		base = dag.Container(dagger.ContainerOpts{Platform: build.platform}).
 			From("ubuntu:"+consts.UbuntuVersion).
 			WithEnvVariable("DEBIAN_FRONTEND", "noninteractive").
 			WithEnvVariable("DAGGER_APT_CACHE_BUSTER", fmt.Sprintf("%d", time.Now().Truncate(24*time.Hour).Unix())).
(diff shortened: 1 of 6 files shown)


cmd/dagger/version.go
```diff
@@ -27,3 +61,124 @@ func short() string {
...
+		}
+
+	default:
+		return nil, fmt.Errorf("unsupported media type: %s\n", desc.MediaType)
+	}
+
+	return annotations, nil
+}
+
+func versionNag(latest string) {
```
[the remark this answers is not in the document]
aluzzardi: 🙏 

That's it? Nothing more complicated than that?

.dagger/build/builder.go
```diff
@@ -210,6 +210,10 @@ func (build *Builder) Engine(ctx context.Context) (*dagger.Container, error) {
 		return nil, fmt.Errorf("unsupported engine base %q", build.base)
 	}
 
+	if build.version != "" {
+		base = base.WithAnnotation(distconsts.OCIVersionAnnotation, build.version)
```
aluzzardi, line 214: This is what's causing the CI failure:

```
build/builder.go:214:15: base.WithAnnotation undefined (type *dagger.Container has no field or method WithAnnotation)
```

It's expected since we just merged `WithAnnotation` and CI is still running the latest released version

cmd/dagger/version.go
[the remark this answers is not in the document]
vito: Yep - it'll always be appropriate to use `cmd.OutOrStdout()` and/or `cmd.ErrOrStderr()` since it can be dynamically redirected as needed externally:

https://github.com/dagger/dagger/blob/cf4cf5214036efea8c31003fa318b30ad02f6014/cmd/dagger/engine.go#L93-L94

For commands like `dagger version` it'll (I think) just go straight to stderr, and for commands that wire up the TUI, it'll be configured properly via the above two lines. So just always remember to trust the `cmd` out/err.

aluzzardi:
@gerhard Can you confirm we're seeing registry hits from this?

cmd/dagger/version.go
```diff
@@ -1,23 +1,59 @@
...
+	cmd := &cobra.Command{
+		Use:   "version",
+		Short: "Print dagger version",
+		// Disable version hook here to avoid double version check
+		PersistentPreRun: func(*cobra.Command, []string) {},
+		Args:             cobra.NoArgs,
+		Run: func(cmd *cobra.Command, args []string) {
+			fmt.Fprintln(cmd.OutOrStdout(), long())
+			if forceVersionCheck {
+				updateAvailable, err := updateAvailable(cmd.Context())
```
jedevc, line 43: I *think* this ends up getting the version twice potentially? Since it's called directly here, and indirectly from `checkForUpdates`?

jedevc: Ah woops my bad :cry:

jedevc (approved):

marcosnils (approved):

gerhard (approved):
👍 We confirmed sync that this behaves as expected. FTR:
image

PR #8406, opened 2024-09-11
💥 Breaking(helm): Remove default oci-max-parallelism num-cpu

gerhard:
This may result in a surprising behaviour, as captured in:
- https://github.com/dagger/dagger/pull/7395

You can still set this if needed, we are just removing it as the default option.

---

Follow-up to:
- https://github.com/dagger/dagger/pull/8348#discussion_r1750138175

---

> [!NOTE]
> The more I think about this, it's not **technically** a breaking change. We change the default behaviour of the Engine provisioned with the Helm chart. We are emphasising the importance of this change, but it's not breaking. OK to leave as is, just clarifying for anyone that might stumble across this PR and wonder what exactly is breaking.


jedevc (approved):

gerhard:
Thanks @jedevc 💪

Will merge as soon as the `Helm / test` passes.

Issue #6999, opened 2024-04-03
[Container.Export] Support custom annotations for exporting oci tar.


morlay:
### What are you trying to do?

I wan't to make the exported tar for `ctr i import`, but now I could't added any custom annotation into `index.json`,
which is important for https://github.com/containerd/containerd/blob/main/core/images/annotations.go#L22`

expect 

```json
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:1e87421ef24417b823d41cacbc9d87bd9133919123479e5c8ad02c5fa14f003c",
      "size": 481,
      "annotations": {
        "io.containerd.image.name":          "quay.io/k0sproject/calico-node:v3.26.1-1",
        "org.opencontainers.image.created": "2024-04-03T03:09:18Z"
      },
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ]
}
```
but now got
```
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:1e87421ef24417b823d41cacbc9d87bd9133919123479e5c8ad02c5fa14f003c",
      "size": 481,
      "annotations": {
        "org.opencontainers.image.created": "2024-04-03T03:09:18Z"
      },
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ]
}
```

### Why is this important to you?

exported ocr tar can `ctr i import` directly without manual tag

mihaelTBTL:
Since @stobias123 closed his issue (#7368), I am bumping this one, but my use case alligns more closely to his.
Im unsure if there are any plans to support this feature, but it would be great if we could specify repository and tag when creating a tarball/export. 

I manage to create a tarball and get it onto my machine, but after loading it into my Docker Engine with `docker load < my_repo:my_tag.tar`, I only see `<none>:<none>` when listing the images. 

It appears to me that Dagger's [Container.as_tarball()](https://dagger-io.readthedocs.io/en/sdk-python-v0.11.9/client.html#dagger.Container.as_tarball) and [Container.import_()](https://dagger-io.readthedocs.io/en/sdk-python-v0.11.9/client.html#dagger.Container.import_) are equivalent to `docker container import` and `docker container export` commands, whereas we're looking for an equivalent of `docker image save` and `docker image load` which do preserve the metadata.

[...]

PR #8409, opened 2024-09-11
container: support OCI annotations

aluzzardi:
Fixes https://github.com/dagger/dagger/issues/6999

### DIFF
--- a/core/integration/container_test.go
+++ b/core/integration/container_test.go
@@ -2806,31 +2806,76 @@ func (ContainerSuite) TestPublish(ctx context.Context, t *testctx.T) {
 }
 
 func (ContainerSuite) TestAnnotations(ctx context.Context, t *testctx.T) {
-	c := connect(ctx, t)
+	build := func(c *dagger.Client, platform dagger.Platform) *dagger.Container {
+		return c.Container(dagger.ContainerOpts{Platform: platform}).
+			From(alpineImage).
+			WithAnnotation("org.opencontainers.image.version", "v0.1.2")
+	}
 
-	testRef := registryRef("container-annotations")
+	t.Run("single-platform", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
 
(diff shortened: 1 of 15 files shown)


core/schema/container.go
```diff
@@ -419,6 +419,11 @@ func (s *containerSchema) Install() {
 			Doc(`The error stream of the last executed command.`,
 				`Will execute default command if none is set, or error if there's no default.`),
 
+		dagql.Func("withAnnotation", s.withAnnotation).
```
helderco, line 422: Can you add a `withoutAnnotation` as well?

core/integration/container_test.go
```diff
@@ -2805,6 +2805,34 @@ func (ContainerSuite) TestPublish(ctx context.Context, t *testctx.T) {
 	require.Equal(t, "im-a-default-arg\n", output)
 }
 
+func (ContainerSuite) TestAnnotations(ctx context.Context, t *testctx.T) {
```
aluzzardi, line 2808: Test is failing because, for some reason, the annotations only end up in multi platform images, not single platform

/cc @jedevc

sdk/php/generated/Client.php
```diff
@@ -146,7 +146,8 @@ public function git(
         ServiceId|Service|null $experimentalServiceHost = null,
         ?string $sshKnownHosts = '',
         SocketId|Socket|null $sshAuthSocket = null,
-    ): GitRepository {
+    ): GitRepository
```
aluzzardi, line 149: No idea why the re-generation changed the formatting? CI is failing because of this

Was generated using: `dagger --source . call sdk php generate export --path .`

/cc @dagger/sdk-php

jedevc (approved):
LGTM :tada: 

Once tests/checks are passing, should be good to merge.

core/schema/container.go
aluzzardi: Done! PTAL

marcosnils (approved):

PR #8391, opened 2024-09-10
fix: relative path in context dir now point to root source dir

TomChv:
Changes relative path root to root source dir instead of source dir based on https://discord.com/channels/707636530424053791/1280818249633890365/1282809411353575424

Adapt tests based on that.

PR #8411, opened 2024-09-11
feat: Support DefaultPath and Ignore in PHP SDK

charjr:
Add support for `default path` and `ignore` in modules created using the PHP SDK.


sdk/php/runtime/template/README.md
```diff
@@ -311,3 +311,48 @@ class Example
      // ...
 }
 ```
+
+#### Directories and Files
+
+Directories and Files have some additional metadata that can be specified
+
+##### Defaults
```
TomChv, line 319: You could also link to the official doc page for more info

https://docs.dagger.io/manuals/developer/functions/#directories-and-files

sdk/php/runtime/template/README.md
```diff
@@ -311,3 +311,48 @@ class Example
...
+
+Specifying a default value for a `Directory` or `File` isn't easily possible. 
+Instead you can specify a `DefaultPath` attribute with an absolute or relative path.
+If relative then the path is relative to the `source` specified in the `dagger.json` file.
```
TomChv, line 323: Not anymore after https://github.com/dagger/dagger/pull/8391

sdk/php/runtime/template/README.md
```diff
@@ -311,3 +311,48 @@ class Example
...
+
+This tells Dagger; if no argument is specified, fetch the `Directory` or `File` from the `DefaultPath`.
+
+The following example shows how you would get the parent directory of your `source`:
+
+```php
+#[DaggerFunction]
+public function myDaggerFunction(
+    #[DefaultPath('../')]
```
TomChv, line 332: With https://github.com/dagger/dagger/pull/8391, it's now `.`

sdk/php/runtime/template/README.md
```diff
@@ -311,3 +311,48 @@ class Example
...
+    // ...
+}
+```
+
+##### Ignore
+
+For Directories only, an `Ignore` Attribute to specify what files|directories should be ignored.
+This is done by providing an array of strings.
+Each string follows the [`.gitignore` syntax](https://git-scm.com/docs/gitignore/en): 
+prefixing a path with `!` will actually include it instead of ignoring it.
```
TomChv, line 344: It's a bit more subtle, when prefix a path with `!`, you actually negate a precedent ignore.

For example: `["go/*", "!go/main.go"]` will ignore everything in the `go` dir except `go/main.go`.
So if you just do `["!main.go"]`, nothing is actually ignored at all.

Also keep in mind that the order of the ignore is important.
If you do `["!main.go", "**"]`, everything will be ignored since the main.go is included then exclude.

sdk/php/runtime/template/README.md
```diff
@@ -311,3 +311,48 @@ class Example
...
+
+The following example would ignore your `vendor/` and `tests/` directories:
+
+```php
+#[DaggerFunction]
+public function myDaggerFunction(
+    #[DefaultPath('../')]
```
TomChv, line 351: Should be `.` with https://github.com/dagger/dagger/pull/8391

sdk/php/dev/dagger.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "php-sdk-dev",
-  "sdk": "github.com/dagger/dagger/sdk/php",
+  "sdk": "php",
   "source": ".",
-  "engineVersion": "v0.12.3"
+  "engineVersion": "v0.12.7"
```
TomChv, line 5: You should update it to `0.13` to benifit from https://github.com/dagger/dagger/pull/8391

TomChv (commented):
Amazing PR, left some comments to make it even better 🚀

sdk/php/runtime/template/README.md
charjr: This is my fault for not making it more clear.

What I mean is we can't do a _default value_ in the way we normally can in PHP.

For instance, I cannot do this:

```php
public function myDaggerFunction(
    Directory $dir = '.'
)
```

Because PHP will complain that the string `'.'` is not an instance of the `Directory` object.

Thus we have to use a `DefaultPath` attribute instead.

Thank you for pointing this out, it's very useful to see where I haven't been clear. I'll sort it out shortly!

sdk/php/runtime/template/README.md
charjr: From what I understood `.` is the source dir.

My example is to get the parent of the source dir.

Do you think it would be helpful for me to demonstrate both?

sdk/php/runtime/template/README.md
charjr: I pinched and reworded this bit, so this may want a bit more detail too:

https://github.com/dagger/dagger/blob/a2d4b1f0cd11a0affde61836b980f4462174e92e/docs/current_docs/manuals/developer/functions.mdx?plain=1#L435

I can update this at the same time to match what you've just taught me if you'd like?

sdk/php/runtime/template/README.md
charjr: I've aimed at the parent directory again, maybe I should have one example as current dir and the other as parent dir... avoid lengthening the docs but still demonstrate both?

sdk/php/runtime/template/README.md
charjr: https://github.com/dagger/dagger/blob/d08ed252db2e8bf2fd340450c8cacfbab4a1e1b3/sdk/php/runtime/template/README.md?plain=1#L322-L326

Is this better?

sdk/php/runtime/template/README.md
charjr: I now demonstrate both and then link to the docs for further detail.

https://github.com/dagger/dagger/blob/d08ed252db2e8bf2fd340450c8cacfbab4a1e1b3/sdk/php/runtime/template/README.md?plain=1#L330-L354

sdk/php/dev/dagger.json
TomChv: I do not see it updated, is it done in another PR?

sdk/php/runtime/template/README.md
```diff
@@ -311,3 +311,63 @@ class Example
...
+
+##### Default Values
+
+It would be tricky to instantiate a `Directory` or `File` object from scratch.
+So the standard way of specifying a default value in PHP is hard to apply.
+
+Instead specify a `DefaultPath` attribute with a `string` path.
+Dagger will do the heavy lifting of creating the default object for you.
+
+Paths are relative to the `source` provided in the `dagger.json` file.
```
TomChv, line 328: I still do not understand this part.

Here are the 3 cases:
```
If an absolute path is specified:
- in a Git repository (defined by the presence of a .git sub-directory), the default context is the root of the Git repository.
- in a non-repository location (defined by the absence of a .git sub-directory), the default context is the directory containing a dagger.json file.

If a relative path is specified the default context is always the directory containing a dagger.json file.
```

TomChv (commented):
Left more, comment, that's better but still room for improvement

sdk/php/runtime/template/README.md
TomChv: Yeah I think that's better to focus on usage in this doc, and check for gitignore page is some1 need more conceptual info

sdk/php/runtime/template/README.md
TomChv: > From what I understood . is the source dir.

Nope, `.` is the location of the `dagger.json` dir, not the source dir

> My example is to get the parent of the source dir.

I think you should also add an example for absolute path.

sdk/php/runtime/template/README.md
TomChv: It's better yep, I still left a comment to clarify what's the root depending of its relative or absolute path

helderco: @charjr, the point of the defaultPath isn't to provide a default value for a complex type, in this case `dagger.Directory`. Its purpose is to allow getting files from outside the _source directory_. Without it, the caller of the module needs to explicitly pass in the directory from outside what the source dir is.

helderco: Note that it only works for directories inside the same git repo (context directory), or within the same module (root directory, i.e., where `dagger.json` is). Before it, modules could only get files from inside of what's in `"source"` (in `dagger.json`).

sdk/php/src/Attribute/DefaultPath.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Dagger\Attribute;
+
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+final readonly class DefaultPath
```
helderco, line 8: You already have a `Dagger\Attribute\Argument`. `DefaultPath` and `Ignore` should be arguments/properties of it. Just like a Function argument's description, they are also metadata for the argument.

sdk/php/runtime/template/README.md
charjr: Apologies, @helderco solved my confusion.

I was testing default paths on an outdated version of Dagger where it was relative to `source`. I've fixed this now and literally copy-pasted your description.

Thank you, for patiently reviewing this :)

charjr: I've updated it once more

https://github.com/dagger/dagger/blob/d925956f33d1dfcdbb3b0e163b2bcf3d053eecc8/sdk/php/runtime/template/README.md?plain=1#L328-L359

sdk/php/src/Attribute/DefaultPath.php
charjr: When we discussed the possibility we felt it was neater to have separate attributes for each.

Reasoning as follows:
1. `DefaultPath` and `Ignore` are not applicable to every `Argument`
2. The convention I've witnessed most commonly in PHP is that each piece of meta data tends to be given a separate attribute.

For example, PHPUnit's attributes look like this:

```php
#[CoversClass(Thing::class)]
#[UsesClass(OtherThing::class)]
class ThingTest extends TestCase
{
    #[Test]
    #[TestDox('description of what is being tested')]
    #[DataProvider('provideTestCases')]
    public function testThingPasses()
    // ...
``` 

Which matches closely to our current implementation: 

```php
#[Argument('Your source directory')]
#[DefaultPath('.')]
#[Ignore('vendor/', 'generated/')]
Directory $source,
```

sdk/php/runtime/template/README.md
charjr: Thank you @helderco I think I have addressed this now.

> Its purpose is to **allow getting files from outside the _source directory_**. 

https://github.com/dagger/dagger/blob/eb1facdfbaf2376efcaf75f2f1bc1870577150b0/sdk/php/runtime/template/README.md?plain=1#L322



> Note that it only works for directories inside the same git repo (context directory), or within the same module (root directory, i.e., where `dagger.json` is) if not in a git repo.

https://github.com/dagger/dagger/blob/eb1facdfbaf2376efcaf75f2f1bc1870577150b0/sdk/php/runtime/template/README.md?plain=1#L323-L325

https://github.com/dagger/dagger/blob/eb1facdfbaf2376efcaf75f2f1bc1870577150b0/sdk/php/runtime/template/README.md?plain=1#L353-L361

sdk/php/src/Attribute/DefaultPath.php
helderco: If that's the convention, then you should deprecate `Argument` in favor of `Doc`.

charjr: > If that's the convention, then you should deprecate `Argument` in favor of `Doc`.

Good point, I agree. Shall I do that in this PR, or make an issue to do in a separate small PR?

charjr: @helderco I've made a PR to implement this change: https://github.com/dagger/dagger/pull/8486

charjr: @helderco https://github.com/dagger/dagger/pull/8486 has now been merged.

`Argument` is deprecated (but still functions).

`Doc` is the way to specify a doc-string for both _Dagger Functions_ and _arguments_.

`Argument` is deprecated, but still functional.

If both `Doc` and `Argument` are present: `Doc` is used, `Argument` is ignored.

sdk/php/runtime/template/README.md
charjr: @helderco @TomChv I think this feedback has been resolved now, please let me know if otherwise. :+1:

sdk/php/src/ValueObject/Argument.php
```diff
@@ -18,7 +18,24 @@ public function __construct(
...
+            ));
+        }
+    }
+
+    public function isOptional(): bool
+    {
+        return $this->type->nullable
+            || $this->default !== null
+            || $this->defaultPath !== null;
     }
```
helderco, line 39: The `TypeDef.withOptional` API actually means "does it take null values?" See:

It's best not to propagate that into the SDK, in this case by renaming the function more accurately. Then you'll see that having a default doesn't belong here:

```suggestion
    public function isNullable(): bool
    {
        return $this->type->nullable
            || $this->defaultPath !== null;
    }
```

But now you can also see that having a `defaultPath` doesn't really make it nullable. In fact this is an implementation detail leak from the engine and we should force it there instead of in the SDKs. So now if you remove that too, you're left with:

```php
    public function isNullable(): bool
    {
        return $this->type->nullable;
    }
```

And there's not much point in that 🙂

For now, you can just remove the `default !== null` part, and I'll move the `defaultPath` part to the engine soon.

charjr: @helderco I see, thank you for the detailed feedback!

I think I have addressed this now, please let me know if it looks okay.

Since it is more of a temporary fix and the method would eventually be redundant; I've removed it, in favour of just checking it directly in the entrypoint, with a todo to remove it once handled by the engine:

https://github.com/dagger/dagger/blob/b9041a4d7eabfa2b5802b7f903589324f1b343be/sdk/php/src/Command/EntrypointCommand.php#L75-L82

helderco: Fyi:
- https://github.com/dagger/dagger/pull/8572

helderco (approved):
LGTM

PR #8412, opened 2024-09-11
docs: improve install instructions regarding path on linux

marcosnils:


marcosnils:
> If we make `sudo` an installation requirement - which I don't think that we should - then `/usr/local/bin` is the more appropriate `BIN_DIR` as `dagger` is not managed by a package manager:

I suggested this way as we're already doing it for the `mac` instructions here: https://github.com/dagger/dagger/blob/a2d4b1f0cd11a0affde61836b980f4462174e92e/docs/current_docs/partials/_install-cli.mdx?plain=1#L44

vikram-dagger:
Adding feedback from the docs survey on this topic, may be relevant? 

image

vikram-dagger:
Also worth noting that in AWS Codebuild I needed to manually update the path in an Amazon Linux runner: https://github.com/dagger/dagger/pull/8455/files#diff-dcdb1e157f89a391d4531bf149788ee95ed0c84e23478cd13ab534ae40384a4cR15

levlaz:
> Adding feedback from the docs survey on this topic, may be relevant?
> 
> image

I think this is a slightly different comment because the objection is piping a downloaded script into bash which some people object to. 

Overall, I agree we should not make sudo the default but we should document the `-E` option in the same way that we do for MacOS

levlaz (approved):
This looks good to me, lets get this merged.

PR #8423, opened 2024-09-12
ci: Restore testdev to the same coverage as it had before #8212

gerhard:
During https://github.com/dagger/dagger/pull/8212, `testdev` was refactored and I missed these important test suites. This puts them back. We should expect it to get slower and more flaky, but at least the coverage will be the same as before.

We were still running the same tests in the split `testdev-module-runtimes`, so were not really missing on anything test
coverage-wise, just the comparison between the original `testdev` and the split one was less accurate.

### DIFF
--- a/.github/workflows/engine-and-cli.yml
+++ b/.github/workflows/engine-and-cli.yml
@@ -57,7 +57,7 @@ jobs:
       - name: "testdev"
         uses: ./.github/actions/call
         with:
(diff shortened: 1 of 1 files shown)


gerhard:
While this seemed like a good idea, `test` failed 3 times in a row when `--parallel=16` is removed.

I will re-purpose this PR and just keep the `testdev` fix.

Issue #8395, opened 2024-09-10
🐞 Parent directory is uploaded when working in a module subdir


gerhard:
### What is the issue?

When running a module from a subdirectory, the parent directory gets uploaded to the Engine:

```console
# ...
cache request: upload /var/folders/z5/n5y08g4j073_ld2d9019j3_80000gn/T/tmp.s1fViQInu2/dagger from 45fhofmgv7lscg6ac23c50pz3
upload /var/folders/z5/n5y08g4j073_ld2d9019j3_80000gn/T/tmp.s1fViQInu2/dagger from 45fhofmgv7lscg6ac23c50pz3
copy upload /var/folders/z5/n5y08g4j073_ld2d9019j3_80000gn/T/tmp.s1fViQInu2/dagger from 45fhofmgv7lscg6ac23c50pz3
# ...
```

As discovered with @aluzzardi earlier today.

See the steps on how to reproduce & the Dagger Cloud Trace link. FWIW:

<img width="1552" alt="image" src="https://github.com/user-attachments/assets/0b0cdda9-f539-4933-93a6-5bb584b50b71">

### Dagger version

dagger v0.12.7 (registry.dagger.io/engine:v0.12.7) darwin/amd64

### Steps to reproduce

```console
cd $(mktemp -d)

git clone https://github.com/dagger/dagger
cd dagger

mkdir foo
cd foo
dagger init --sdk=go --name=foo -vvv
```

### Log output

https://dagger.cloud/dagger/traces/5d85d7b1d5483304f37b6c8dfc3db8e2

helderco:
Uhh.. this explains a lot actually!

helderco:
Wait, taking a closer look. It seems that it's loading the context directory, not necessarily the parent directory right? That's expected.

gerhard:
I'm running the command in `dagger/foo`. That makes me think that the context directory is `foo`. There is no reason for it to go up a level - there is no configuration in the default init module that would make it behave this way.

Did you try to follow the steps to reproduce?

helderco:
A module has 3 directories (ignore the new "context directories" feature as that's something different):
- **context directory**: git root (where `.git` is) or the _root directory_ if not in a git repo
- **root directory**: the directory where  `dagger.json` is
- **source directory**: the directory referenced in `"source"` in `dagger.json`

[...]

PR #8425, opened 2024-09-12
.dagger: fix gitDir glob pattern to ignore all but `.git`

vito:
right now this is uploading everything which takes quite a while

### DIFF
--- a/.dagger/main.go
+++ b/.dagger/main.go
@@ -43,7 +43,7 @@ func New(
 	// Git directory, for metadata introspection
 	// +optional
 	// +defaultPath="/"
-	// +ignore=["!.git"]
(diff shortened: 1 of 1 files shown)


jedevc:
Aha :saluting_face: noticed this, thanks!

gerhard:
I **think** this fixes:
- https://github.com/dagger/dagger/issues/8395

Would you agree @vito?

vito:
@gerhard Hmm hard to say - this particular fix should only affect invoking the Dagger module at the root of the repo. Sounds like those issues are more general.

shykes:
This is because of the breaking change in 0.13. Previously you had to omit the `*'

PR #8426, opened 2024-09-12
ci: remove TestDev reference

jedevc:
`TestDev` is not a suite that exists. Not sure what this is doing here, I don't think this has an effect?

Even if it did have an effect, it looks like it would be doing the *opposite*?

Looks like this was introduced here: [`8ec2238` (#8212)](https://github.com/dagger/dagger/pull/8212/commits/8ec22381d85f88654e356bf2fd57591537aaf6ec)

### DIFF
--- a/.github/workflows/engine-and-cli-split.yml
+++ b/.github/workflows/engine-and-cli-split.yml
@@ -73,7 +73,7 @@ jobs:
       - name: "testdev"
         uses: ./.github/actions/call
         with:
-          function: "test specific --run='TestModule' --skip='TestDev' --race=true --parallel=16"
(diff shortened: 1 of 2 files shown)


gerhard (commented):
This was intentional.

Looking at the list of traces / checks, it is impossible to know which one is running against a Dev engine without clicking on it. This makes it easy to spot test suites which run against a dev Engine, e.g.

Screenshot 2024-09-12 at 15 03 31

While introducing a new flag would have been cleaner & the intent clearer, this felt like a quick shortcut. I'm finding it useful in a bunch of other places too:

image

jedevc:
Can we update to use a flag that's a no-op then? The current intention isn't clear, and isn't documented either inline or in #8212.

PR #8431, opened 2024-09-12
Allow the specification of a custom Python package index

ClifHouck:
Adds the ability to specify the Python package index that dagger uses for calls to `uv pip install`.

Related to #6599 and https://github.com/G-Research/gr-oss/issues/780


helderco:
I believe this is already supported in the uv table, no need to add it to dagger's. Doesn't this work for you?

```toml
# in the module's `pyproject.toml`
[tool.uv]
index-url = "https://gresearch.github.io/simple"
```

ClifHouck:
> I believe this is already supported in the uv table, no need to add it to dagger's. Doesn't this work for you?
> 
> ```toml
> # in the module's `pyproject.toml`
> [tool.uv]
> index-url = "https://gresearch.github.io/simple"
> ```

When trying that method dagger still pulls from PyPI to bootstrap itself/creating the module, ie `dagger init` fails because PyPI is not available and it doesn't respect the `tool.uv` setting.

helderco:
I'm not sure I follow. Which part does it fail in particular? What `dagger version` are you using?

helderco:
Oh, I suppose it may fail when running codegen because that's not using the module's pyproject.toml, it's using the one from the `dagger-io` library.

I need to think about this a bit more. I'm very much keen on finding another solution rather than adding more configs to pass to uv.

helderco:
In the meantime, there's a workaround you can do which is to customize the base image with that env var, but you'll need to publish that image in a registry.

ClifHouck:
Seems like another option is to have dagger respect the project's `[tool.uv] index-url` setting during `init`.

helderco:
> Seems like another option is to have dagger respect the project's `[tool.uv] index-url` setting during `init`.

Yes, that's a short term quick solution. You can just read that setting in `Discovery` and add the env var to the base image automatically if it's set so it gets picked up even by codegen, and as early as `dagger init`. Do you want to make that change? Basically read from `[tool.uv]` rather than `[tool.dagger]`.

The longer term fix I want to do is to avoid running codegen in it's own environment.

ClifHouck:
> > Seems like another option is to have dagger respect the project's `[tool.uv] index-url` setting during `init`.
> 
> Yes, that's a short term quick solution. You can just read that setting in `Discovery` and add the env var to the base image automatically if it's set so it gets picked up even by codegen, and as early as `dagger init`. Do you want to make that change? Basically read from `[tool.uv]` rather than `[tool.dagger]`.
> 
> The longer term fix I want to do is to avoid running codegen in it's own environment. Then it won't matter.



> > Seems like another option is to have dagger respect the project's `[tool.uv] index-url` setting during `init`.
> 
> Yes, that's a short term quick solution. You can just read that setting in `Discovery` and add the env var to the base image automatically if it's set so it gets picked up even by codegen, and as early as `dagger init`. Do you want to make that change? Basically read from `[tool.uv]` rather than `[tool.dagger]`.
> 
> The longer term fix I want to do is to avoid running codegen in it's own environment. Then it won't matter.



> > Seems like another option is to have dagger respect the project's `[tool.uv] index-url` setting during `init`.
> 
> Yes, that's a short term quick solution. You can just read that setting in `Discovery` and add the env var to the base image automatically if it's set so it gets picked up even by codegen, and as early as `dagger init`. Do you want to make that change? Basically read from `[tool.uv]` rather than `[tool.dagger]`.
> 
> The longer term fix I want to do is to avoid running codegen in it's own environment. Then it won't matter.

I think I've accomplished this, let me know what you think. Thanks!

.changes/unreleased/Added-20240916-122324.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Added
+body: A Python project's tool.uv index-url setting will be respected and used as the index-url for dagger's codegen as well as project package retrieval via uv.
+time: 2024-09-16T12:23:24.696966201-05:00
+custom:
+    Author: Clif Houck
```
helderco, line 5: Should be your GH username, not name. 
```suggestion
    Author: ClifHouck
```

.changes/unreleased/Added-20240916-122324.yaml
helderco, line 1: Should be in `sdk/python` instead of repo root.

helderco (commented):
Thanks for your contribution! Now we're on the right path. 🎉 Needs a test case.

sdk/python/runtime/main.go
```diff
@@ -226,6 +227,7 @@ func (m *PythonSdk) WithBase() (*PythonSdk, error) {
 		WithEnvVariable("UV_SYSTEM_PYTHON", "1").
 		WithEnvVariable("UV_LINK_MODE", "copy").
 		WithEnvVariable("UV_NATIVE_TLS", "1").
+		WithEnvVariable("UV_INDEX_URL", m.IndexUrl()).
```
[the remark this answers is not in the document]
ClifHouck: Addressed in https://github.com/dagger/dagger/pull/8431/commits/f290f71d11f6cf7fc6fdc883e57602bb53a695e5

sdk/python/runtime/main.go
```diff
@@ -383,7 +385,7 @@ func (m *PythonSdk) WithInstall() *PythonSdk {
 			// If there's a lock file, we assume that all the dependencies are
 			// included in it so we can avoid resolving for them to get a faster
 			// install.
-			install = append(install, "--no-deps", "-r", PipCompileLock)
+			install = append(install, "--no-deps", "-r", PipCompileLock, "--index-url", m.IndexUrl())
```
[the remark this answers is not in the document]
ClifHouck: Addressed in https://github.com/dagger/dagger/pull/8431/commits/f290f71d11f6cf7fc6fdc883e57602bb53a695e5

sdk/python/runtime/extension.go
```diff
@@ -82,6 +82,13 @@ func (m *PythonSdk) UvVersion() string {
 	return m.Discovery.UserConfig().UvVersion
 }
 
+func (m *PythonSdk) IndexUrl() string {
```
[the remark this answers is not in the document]
ClifHouck: Addressed in https://github.com/dagger/dagger/pull/8431/commits/f290f71d11f6cf7fc6fdc883e57602bb53a695e5

sdk/python/runtime/discovery.go
```diff
@@ -24,6 +24,11 @@ var DirExcludes = []string{".venv", "sdk"}
 // files concurrently rather than making blocking calls later.
 var FileContents = []string{"pyproject.toml", ".python-version"}
 
+// Uv config bits we'd like to consume.
+type UvConfig struct {
+	IndexUrl string `toml:"index-url"`
```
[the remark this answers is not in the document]
ClifHouck: Addressed in https://github.com/dagger/dagger/pull/8431/commits/f290f71d11f6cf7fc6fdc883e57602bb53a695e5

.changes/unreleased/Added-20240916-122324.yaml
ClifHouck: Addressed in https://github.com/dagger/dagger/pull/8431/commits/f290f71d11f6cf7fc6fdc883e57602bb53a695e5

.changes/unreleased/Added-20240916-122324.yaml
ClifHouck: Addressed in https://github.com/dagger/dagger/pull/8431/commits/f290f71d11f6cf7fc6fdc883e57602bb53a695e5

ClifHouck:
> Thanks for your contribution! Now we're on the right path. 🎉 Needs a test case.

I'm kind of struggling where to put the test case. There don't seem to be any unit tests in "sdk/python/runtime" and that seems like that's where a test addressing this PR should go. Do have any guidance on where and how to add a test for this change?

helderco:
> Do have any guidance on where and how to add a test for this change?

Yes, they should be in this file:

https://github.com/dagger/dagger/blob/7cfdf1302d7ad735765cc19bdbed2d64b21a82ee/core/integration/module_python_test.go

.changes/unreleased/Added-20240916-122324.yaml
helderco: Not commited? 🙂

sdk/python/runtime/discovery.go
```diff
@@ -111,6 +118,10 @@ func (d *Discovery) UserConfig() *UserConfig {
 	return &d.Config.Tool.Dagger
 }
 
+func (d *Discovery) UvConfig() *UvConfig {
```
helderco, line 121: Missing description:
```suggestion
// UvConfig is the configuration the user can set in pyproject.toml, under
// the "tool.uv" table.
func (d *Discovery) UvConfig() *UvConfig {
```

sdk/python/runtime/main.go
```diff
@@ -228,6 +229,13 @@ func (m *PythonSdk) WithBase() (*PythonSdk, error) {
 		WithEnvVariable("UV_NATIVE_TLS", "1").
 		WithWorkdir(path.Join(m.SourcePath, m.Discovery.SubPath))
 
+	if m.IsUvIndexUrlSpecified() {
+		m.Container = m.Container.WithEnvVariable("UV_INDEX_URL", m.IndexUrl())
+	}
+	if m.IsUvExtraIndexUrlSpecified() {
+		m.Container = m.Container.WithEnvVariable("UV_EXTRA_INDEX_URL", m.ExtraIndexUrl())
+	}
```
helderco, line 237: The extra "...Specified" functions aren't necessary, just check for an empty string:
```suggestion
	if m.IndexUrl() != "" {
		m.Container = m.Container.WithEnvVariable("UV_INDEX_URL", m.IndexUrl())
	}
	if m.ExtraIndexUrl() != "" {
		m.Container = m.Container.WithEnvVariable("UV_EXTRA_INDEX_URL", m.ExtraIndexUrl())
	}
```

sdk/python/runtime/extension.go
```diff
@@ -82,6 +82,32 @@ func (m *PythonSdk) UvVersion() string {
...
+
+// IndexUrl specified by [tool.uv] index-url from project's
+// pyproject.toml configuration if specified, otherwise
+// defaults to DefaultPackageIndexUrl.
+func (m *PythonSdk) IndexUrl() string {
+	if m.IsUvIndexUrlSpecified() {
+		return m.Discovery.UvConfig().IndexUrl
+	}
+	return DefaultPackageIndexUrl
+}
```
helderco, line 98: This function is exposed to the API so it follows the API's convention for a description, instead of Go's.

```suggestion
// Uv's "index-url" setting
func (m *PythonSdk) IndexUrl() string {
	return m.Discovery.UvConfig().IndexUrl
}
```

sdk/python/runtime/extension.go
```diff
@@ -82,6 +82,32 @@ func (m *PythonSdk) UvVersion() string {
...
+
+// Lets us determine if [tool.uv] extra-index-url is set or not.
+func (m *PythonSdk) IsUvExtraIndexUrlSpecified() bool {
+	return m.Discovery.UvConfig().ExtraIndexUrl != ""
+}
+
+// ExtraIndexUrl specified by [tool.uv] extra-index-url from project's
+// pyproject.toml configuration.
```
helderco, line 106:
```suggestion
// Uv's "extra-index-url" setting
```

.changes/unreleased/Added-20240916-122324.yaml
ClifHouck: Moved.

.changes/unreleased/Added-20240916-122324.yaml
ClifHouck: Whoops, fixed this time.

ClifHouck:
@helderco Do you have any tips for running integration tests locally? I'm trying to add a test to `TestPython/TestUv` but it fails on trying to fetch the golang container during test execution. It shouldn't have any problem reaching the internet as far as I know.

helderco:
> @helderco Do you have any tips for running integration tests locally? I'm trying to add a test to `TestPython/TestUv` but it fails on trying to fetch the golang container during test execution. It shouldn't have any problem reaching the internet as far as I know.

Just run this from the root of the repo:
```
dagger call test specific --pkg=./core/integration --run="Python/Uv"
```

That'll run the `TestUv` test in the `PythonSuite` suite.

ClifHouck:
> Just run this from the root of the repo:
> 
> ```
> dagger call test specific --pkg=./core/integration --run="Python/Uv"
> ```
> 
> That'll run the `TestUv` test in the `PythonSuite` suite.

Right, I'm already running a similar command to run just my specific test, but it is failing due to not being able to retrieve the go docker image, "golang:1.23.1-alpine" inside the executed test. Am I missing some setup step that would allow it to retrieve this image? I don't have a problem resolving "https://mirror.gcr.io/v2/library/golang/manifests/1.23.1-alpine?ns=docker.io" outside of the test execution.

helderco:
Possibly because that command builds a dev engine to run the test suite against and it may not pick up on your current engine's mirror config. 

Do you know @sipsma, @marcosnils, or @jedevc?

helderco:
@ClifHouck, in the meantime you can just push what you have and I'll fix it if necessary. It's a quick test to make.

ClifHouck:
> @ClifHouck, in the meantime you can just push what you have and I'll fix it if necessary. It's a quick test to make.

Yeah I tried to figure out what the network issue is but I've given up. I pushed a commit with a test that might work? Hard to tell without being able to run the test.

helderco (approved):
Thanks for your first contribution, @ClifHouck! 🎉

Just waiting for checks to pass now.

ClifHouck:
@helderco Thanks for shepherding this into the repo! Next time I'll try to get my tests working locally...

PR #8435, opened 2024-09-12
feat(sdk/elixir): apply `+ignore` to sdkSourceDir

wingyplus:
Make Elixir SDK consistent with PHP (in #8369), the include patterns is now moved to `+ignore` directive.


gerhard:
👋 @wingyplus 

We know that the job cancellations have been super disruptive. We are planning on changing all jobs to use on-demand instances, so that they don't get cancelled. @jedevc is looking into it.

helderco (approved):

Issue #7832, opened 2024-07-05
✨ Apply ignore patterns from Directory argument in `dagger call`


helderco:
With https://github.com/dagger/dagger/pull/7744 we can define ignore patterns to apply when loading a default directory from the context. See:
- https://github.com/dagger/dagger/issues/7647

However, if you specify a directory explicitly with `dagger call`, those patterns won’t be applied. Module authors need to continue setting views and tell end users to specify it next to the path.

**If there are ignore patterns for a Directory type**, the CLI can use them. If a view is passed, use it. But if not, use the ignore patterns from the argument.

I’m assuming that when a module author specifies ignore patterns in a directory argument it’s always desirable to apply those patterns to any directory. An example of that is a function that lints Python code. There’s only very specific file extensions that you’d want to filter here, as any others would unnecessarily invalidate the cache or upload more than it needs to.

If not, there’s the *views* escape hatch, however I’m not sure if views should continue to exist since they were created as a stopgap to enable filtering local directories, provided as input to the CLI. If we do remove views, then we need to think a bit more if there’s a need for an escape hatch or not.

jedevc:
Can we split out a separate issue for passing directories from a module to another module? Just so we have a place to track this?

PR #8436, opened 2024-09-12
feat: apply ignore metadata on directory flags from the CLI

TomChv:
Closes https://github.com/dagger/dagger/issues/7832

Update `ModuleFunction.setCallInputs` to apply ignore patterns on `Directory` that have `ignore` as metadatas.
Update `ResolveDirectoryFromCaller` arguments to supports optional ignore argument.
Update CLI `Directory` handling to also apply ignore on Directory if `ignore` is set.

~~Still need to add tests, I only manually tested now~~: Test added

Example with Typescript:

```ts
import { Directory, object, func, argument } from "@dagger.io/dagger"

@object()
class Test {
  @func()
  ignoreDir(
    @argument({ ignore: ["**", "!src/"]})
    dir: Directory
  ): Directory {
    return dir
  }
}
```

```
$ dagger call ignore-dir --dir .

_type: Directory
digest: sha256:c19b982b8d2eaab4a9fae8b20da1e906b77f8d3579a6506ea2f597afc0a17f30
entries:
    - src
```

### DIFF
--- a/.changes/unreleased/Added-20240920-122701.yaml
+++ b/.changes/unreleased/Added-20240920-122701.yaml
@@ -1,8 +1,11 @@
 kind: Added
 body: |-
-  Apply `ignore` function parameter metadata on any Directory passed from the CLI.
+  Apply `ignore` function parameter metadata on Directory passed from the CLI.
 
-  This extends the usage of `ignore` to not only dir loaded with `defaultPath` but any directory passed as function's param.
+  This extends the usage of `ignore` to not only dir loaded with `defaultPath` but directory passed as function's param
+  from the CLI.
+
+  Note: ignore patterns are not applied for module to module calls.
 time: 2024-09-20T12:27:01.295001+02:00
 custom:
   Author: TomChv
--- a/.changes/unreleased/Added-20240920-122701.yaml
+++ b/.changes/unreleased/Added-20240920-122701.yaml
@@ -1,0 +1,9 @@
+kind: Added
+body: |-
+  Apply `ignore` function parameter metadata on any Directory passed from the CLI.
+
+  This extends the usage of `ignore` to not only dir loaded with `defaultPath` but any directory passed as function's param.
+time: 2024-09-20T12:27:01.295001+02:00
+custom:
+  Author: TomChv
+  PR: "8436"
--- a/core/modfunc.go
+++ b/core/modfunc.go
@@ -137,14 +137,6 @@ func (fn *ModuleFunction) setCallInputs(ctx context.Context, opts *CallOpts) ([]
 			return nil, fmt.Errorf("failed to convert arg %q: %w", input.Name, err)
 		}
 
-		// Disable for now until we figure out how to not "double-ignore" arguments from CLI
-		// if len(arg.metadata.Ignore) > 0 {
-		// 	converted, err = fn.applyIgnoreOnDir(ctx, opts.Server, arg.metadata, converted)
-		// 	if err != nil {
-		// 		return nil, fmt.Errorf("failed to apply ignore pattern on arg %q: %w", input.Name, err)
-		// 	}
-		// }
-
 		encoded, err := json.Marshal(converted)
 		if err != nil {
 			return nil, fmt.Errorf("failed to marshal arg %q: %w", input.Name, err)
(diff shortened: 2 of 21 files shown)


TomChv:
I still need to write tests for them, will do tomorrow!

TomChv:
Test added, however the CI look super bad, I'm trying to make it green

TomChv:
After fixing my issue with type handling, I'm seeing such a weird error on that PR:

All the `sdk check` steps are failing for the same reason:

```
error: No `pyproject.toml` found in current directory or any parent directory
```

-> `check --targets=sdk/elixir`: https://dagger.cloud/dagger/traces/80fbfe075e520d2284f01154b0787ce9
-> `check --targets=sdk/typescript`: https://dagger.cloud/dagger/traces/cab80ff071727e3005615ddf2bbe86b9
-> `check --targets=sdk/python`: https://dagger.cloud/dagger/traces/4907620f08913488f055d319944446af

And basically all the SDK check related job.

@helderco do you have any idea why I would get such error?

I also can see all the tests python failing in `TestModules`: https://dagger.cloud/dagger/traces/8cfb5ad4249ebe191b9f84010cc73fd3

helderco:
I don't think it's `sdk check` particularly. Should happen on just `dagger functions`. It's because `/.dagger` depends on `/sdk/python/dev` and there must be an issue loading that module's files.

helderco:
The PR title isn't accurate. This is only adding support for applying ignore on directory arguments provided via CLI.

core/modfunc.go
```diff
@@ -137,6 +137,13 @@ func (fn *ModuleFunction) setCallInputs(ctx context.Context, opts *CallOpts) ([]
 			return nil, fmt.Errorf("failed to convert arg %q: %w", input.Name, err)
 		}
 
+		if len(arg.metadata.Ignore) > 0 {
+			converted, err = fn.applyIgnoreOnDir(ctx, opts.Server, arg.metadata, converted)
+			if err != nil {
+				return nil, fmt.Errorf("failed to apply ignore pattern on arg %q: %w", input.Name, err)
+			}
+		}
```
helderco, line 145: For a directory provided via CLI, and uploaded via `bk.LocalImport`, this will filter the resulting directory another time, but with this, right?
```graphql
{
    directory { 
        withDirectory(path: "/", directory: "<ID>", exclude: ["<same ignore pattterns>"]) 
    }
}
```

I'd rather find a fix so it doesn't filter twice, before enabling this for all directories. Simply having support for the CLI is enough to deprecate _views_.

This is what's causing the issue with `sdk/python/dev` by the way.

core/schema/modulesource.go
```diff
@@ -1265,6 +1266,10 @@ func (s *moduleSchema) moduleSourceResolveDirectoryFromCaller(
 		}
 	}
 
+	if len(args.Ignore) > 0 {
+		excludes = append(excludes, args.Ignore...)
+	}
+
```
helderco, line 1272: I don't think this is a good idea. When specifying a view:
```
dagger call --source=.:default source
```
This code is adding the patterns from the view, with different semantics from `// +ignore` because of the split into exclude/include based on `!`, and adding the `// +ignore` patterns on top.

I think it's less surprising if you apply one or the other. Basically let a view override the patterns from code, instead of acting as a default.

core/modfunc.go
TomChv: > For a directory provided via CLI, and uploaded via bk.LocalImport, this will filter the resulting directory another time, but with this, right?

You're right, so you suggest that for now, we only applied //+ignore when the module is called from the CLI?

core/schema/modulesource.go
TomChv: > This code is adding the patterns from the view, with different semantics from // +ignore because of the split into exclude/include based on !, and adding the // +ignore patterns on top.

And also because orders matter, it might totally break the expect result.

TomChv:
I'm trying your suggestion, see if it fixes my issues

core/modfunc.go
helderco: Yes.

core/schema/modulesource.go
helderco: Exactly.

core/modfunc.go
TomChv: Looks like it fixed the issue! I'm trying to get the CI green and then we can do the ignore process in a follow up

core/modfunc.go
```diff
@@ -137,6 +137,14 @@ func (fn *ModuleFunction) setCallInputs(ctx context.Context, opts *CallOpts) ([]
 			return nil, fmt.Errorf("failed to convert arg %q: %w", input.Name, err)
 		}
 
+		// Disable for now until we figure out how to not "double-ignore" arguments from CLI
+		// if len(arg.metadata.Ignore) > 0 {
+		// 	converted, err = fn.applyIgnoreOnDir(ctx, opts.Server, arg.metadata, converted)
+		// 	if err != nil {
+		// 		return nil, fmt.Errorf("failed to apply ignore pattern on arg %q: %w", input.Name, err)
+		// 	}
+		// }
```
jedevc, line 146: Why are we disabling this?

What's the issue with the double-ignore? My expectation is that this would work right? It's just that we would not actually end-up excluding things, so it's slightly worse performance, but it would still do the expected thing right?

If it's just that, we should enable it, leave a note, and come back and optimize later by fixing the "double-ignore" issue.

core/modfunc.go
```diff
@@ -435,6 +443,73 @@ func (fn *ModuleFunction) linkDependencyBlobs(ctx context.Context, cacheResult *
 	return nil
 }
 
+// func (fn *ModuleFunction) applyIgnoreOnDir(ctx context.Context, dag *dagql.Server, arg *FunctionArg, value any) (any, error) {
```
jedevc, line 446: We should avoid committing commented code - it simply ends up getting out of date with the rest of the code.

core/modfunc.go
TomChv: Because of this: https://github.com/dagger/dagger/pull/8436#discussion_r1763662136, but I think I'll just remove that part of the code since I talked with Helder and he said it wasn't very important to support ignore in module to module

core/modfunc.go
TomChv: Yeah, I'll remove it

TomChv: I removed the commented code in [d974dcb](https://github.com/dagger/dagger/pull/8436/commits/d974dcbd73cbb7d23e947bc087577d451a75216e)

core/modfunc.go
TomChv: Removed in [d974dcb](https://github.com/dagger/dagger/pull/8436/commits/d974dcbd73cbb7d23e947bc087577d451a75216e), supporting ignore from module to module will be done in a follow up later.

core/modfunc.go
jedevc: Hm, I disagree here - I think the filtering twice is frustrating from a performance point of view, but I'd rather that than have entirely different behavior when invoking from the CLI vs from another module.

Is your concern just performance @helderco? Or is there a correctness concern here as well?

---

As a suggestion, the way to track this would be add a field to `Directory`s that indicate which `ignore` patterns were used during `resolveDirectoryFromCaller`. Then we just check if those were equal to the patterns specified here, and if they are, we can skip the `applyIgnoreOnDir`.

This should be pretty simple, it just needs to be plumbed through?

.changes/unreleased/Added-20240920-122701.yaml
```diff
@@ -0,0 +1,9 @@
+kind: Added
+body: |-
+  Apply `ignore` function parameter metadata on any Directory passed from the CLI.
+
+  This extends the usage of `ignore` to not only dir loaded with `defaultPath` but any directory passed as function's param.
```
jedevc, line 5: See https://github.com/dagger/dagger/pull/8436#discussion_r1768373644 - I think we should ship both of these together, it's a weird functionality gap to only do one of them (especially given it doesn't seem tricky to do them at the same time?)

But if we decide to ship them separately, then this sentence should be rephrased for clarity - it's not *any* directory passed as a function param, it's *only* for directories passed on the CLI.

TomChv: > (especially given it doesn't seem tricky to do them at the same time?)

Yes the function I did was quite simple but it broke all the tests for a weird reason where python had missing file, Helder suggested to disable the double ignore and it fixed the issue.

I'll update the release note

TomChv:
The lint error doesn't seem related to my changes, I'm trying locally to see if I can repro

https://dagger.cloud/dagger/traces/d36cdff1f83c06434901bb028379210b?span=341b0c4c8564184f#341b0c4c8564184f

core/modfunc.go
helderco: > Why are we disabling this?

The implementation didn't seem to correctly handle the ignore patterns in `/sdk/python/dev` when used as a dependency in `/.dagger` (missing `pyproject.toml` for example). I suggested releasing the CLI use case first, depending on how long it would take to do this properly for all directories, since users are reporting slowness there.

jedevc (approved):
SGTM, won't block on a full module implementation - but we should follow-up with the asap.

TomChv:
Screenshot 2024-09-23 at 18 28 45
Lint is good locally, so guess it was a flack, gonna run the CI until it's green so I can merge that one

Issue #8421, opened 2024-09-11
✨ Enable ability for Dagger functions to return an error and a response


marcosnils:
### What are you trying to do?

# Problem

Dagger functions can currently return an error OR any other value but they can't do both at the same time. This makes it impossible to implement pipelines steps that may fail (and ultimately should fail the pipeline), but always yields some report. Typical example: test or lint results.

ref: https://discord.com/channels/707636530424053791/1275216033862647890

### Why is this important to you?

Similarly to how other CI pipeline steps currently work today, it should be possible to achieve the same behavior in Dagger as doing: `./my_command 2>my_output_report`  to get the execution status as well as any artifacts that might be generated.

### How are you currently working around this?

The only way to go around this limitation is by actually fiddling with the pipeline code and perform this in two `dagger calls`. One to fetch whatever artifact we need from our pipeline, and the second to actually make the execution fail.

vito:
Half baked thought: we actually don't even have a nice way for functions to return errors at the moment.

If you return an error in the Go SDK that just bubbles up into a `fmt.Println(os.Stderr, err.Error())` and `os.Exit(1)` in the runtime entrypoint. which just turns into an ugly error message in the trace, even if you returned a nicely formatted message.

If we changed that to something like `function.ReturnError(…)` similar to how we register function return values, that might be a building block here.

vito:
A bikeshedding opportunity emerges:

* https://github.com/dagger/dagger/pull/8442#issuecomment-2353402147

MVP API:

```graphql
type Query {
  # Construct an Error value to be returned by a function.
  error(message: String!): Error!
}

type Error {
  id: ErrorID!
  # A brief description of the error.
  message: String!
}

type FunctionCall {
  # Return an error from the function.
  returnError(error: ErrorID!): Void
}
```

Quoting for convenience:

> Broader question: can we envision building on this to satisfy the "error reports" use case? For example, by supporting multiple errors, or hierarchical errors, or other extensions to the `Error` type? Or, do we think this is fundamentally off-base and we actually need to support 'arbitrary type + error' returns?
> 
> Putting my cards on the table: sticking with "return a value OR an error" seems familiar and conventional to me, so it'd be pretty convenient if every use case that motivated #8421 (lints, tests, ???) neatly fit into a rich `Error` API. But, that's just my current thinking - I see this more like the start of a beautiful bikeshed. 😄
> 
[...]

PR #7833, opened 2024-07-05
tui/go sdk: clean up redundant error wrapping

vito:
fixes #7007

* remove a couple of not really helpful error wrapping spots (`invoke:`, `response from query:`)
* specifically for modules, stop printing ExecErrors and GraphQL errors,  so that we don't have redundant output with the trace visualization
  * cc @helderco @TomChv - we should make sure the equivalent is done by Python and TypeScript SDKs too. Basically, the module entrypoint should just skip logging things that look like "typical" errors (exec error or GraphQL API error). Not sure what they currently do.
* hoist error logs up to the top level and print them beneath the progress tree on the final render

PR #8442, opened 2024-09-13
Better errors, cached states, pending states, duration accounting, fewer spans

vito:
This PR is a revamp of a bunch of status-related things. I'll give each one its own section.

## New Telemetry test suite

First off - everything in this PR (I think) is now covered by a nice high-level test suite that will show us when there are regressions in the TUI. There are about a million moving parts involved with getting it to look right, so I'm super happy to finally have a safety net for all that.

**Reviewer advice:** there's a lot of change piled together here and it's all pretty inter-dependent. Would personally recommend mostly relying on the test coverage / golden output examples here and otherwise trying out the CLI by hand, though I certainly wouldn't mind reviewing the code too!

The tests work via an additional TUI progress format, `report`, which tells it to just print the final TUI render. We then apply a million regexes to stabilize the output for future comparison (so things like `1.3s` become `X.Xs`).

To update the expected test output, run:

```sh
dagger call test telemetry --update -o .
```

## General TUI polish

* Calls will now be rendered as chained pipelines when possible, with gaps in between for legibility.
* Indent guides are now displayed.

(No Before/After since it's visible in all subsequent screenshots.)

## Better errors

Functions can now return nice error messages, instead of just printing to stderr and exiting. No more `process /runtime did not complete successfully: exit status 2`. Instead we show the actual error returned by the function.

* @helderco and @TomChv - the Python and TypeScript SDKs will need to be updated for this - you just need to call `ReturnError` with an `Error` value containing the desired message. Until that's done they'll just keep the same behavior as today.
* More info: https://github.com/dagger/dagger/pull/8442#issuecomment-2353402147

Also, the TUI (and upcoming web UI) will now bubble up error logs to the top-level so you don't need to deep-dive anymore. I decided to _only_ bubble up spans that have logs, since it seems to have exceptionally good signal:noise. This is compared to bubbling up all errors, which the web UI currently does, and often just leads to a sea of red without any idea of what to click on.

Before:

[asciicast](https://asciinema.org/a/HhmNYQ3YI8LhV5Ga0PHV3TXHW)

After:

[asciicast](https://asciinema.org/a/2gdGjfqWFcGvlMJXsz5R86b5f)

## Cached state

The `CACHED` state is back! Thanks to @jedevc for getting started on this in #8016. 🙏 

This involved quite a few tricks:

* Translating `load cache:` spans to a plain span with `dagger.io/dag.cached: true`.
* Keeping track of all Buildkit ops that we have seen so we can deeply inspect them in the telemetry pipeline.
* Using that info, for any completed Buildkit span, emit an additional attribute containing all deep vertex digests that were an input to the span.
* When we see a cache hit for vertex A, we also infer that input vertices B, C, and D were cached.

The last part is important because Buildkit doesn't emit a span for each and every vertex involved - only as far as it needs to look. But the UI still needs to know the fate of vertices B, C, and D, so we have other spans "vouch" for them.

Before:

[asciicast](https://asciinema.org/a/qzjc2hhnwMcU9DuGZKS5g10yX)

After:

[asciicast](https://asciinema.org/a/UUK0Mv5gOnPJiIZbuS5pzjb8Z)

## Pending state

We now pay closer attention to which 'effects' a span like `Container.withExec` claims it installed, and will mark the span "pending" (empty circle) until all of its effects have completed.

This also builds on the work needed to implement `CACHED` state (the "vouching" bit).

Before:

[asciicast](https://asciinema.org/a/ajXMRYdSGkRZu0FUn2xZMq5fT)

After:

[asciicast](https://asciinema.org/a/SzFV8sHiJqNHvM9sc5Kup0gFC)

## More accurate duration accounting (via span links)

Previously `withExec` would just say "I installed effect A" and then the Buildkit span (`exec foo`) would arrive saying "I am effect A" - but this broke down in some cases where it was possible for an effect to run multiple times (e.g. our integration tests). Now each Buildkit span directly [links](https://opentelemetry.io/docs/concepts/signals/traces/#span-links) to its specific installer span, so we don't have to guess.

## Fewer spans

* Most Buildkit spans are now treated as an implementation detail of the span that they link to. They contribute to the linked span's status and duration, and just hide out of view by setting `dagger.io/ui.passthrough`.
* The `sync` span is now treated as if it's `dagger.io/ui.passthrough` in the UI - it's never interesting on its own. (Similar to `id` calls.)

Exceptions to this rule:

* `resolving docker.io/library/alpine:latest` - this is a synchronous action that is useful to show
* Dockerfile spans (`[1/5] FROM alpine`, etc.) - these spans have a "one-to-many" relationship to the install site (`Container.build`) and are obviously valuable to show

## More accurate failure propagation

Even though we have span links now, there are still scenarios where we want to use effect IDs to determine a span's status in the TUI.

Consider this example:

```go
func (b *Bass) Integration() *dagger.Container
```

This function returns a `*dagger.Container`, and if that container fails to run, we actually want to treat `Integration()` as if it failed - even though it "successfully" returned the failed container, that's not very meaningful.

So, now in addition to span links we'll bubble up errors to any sites that installed a failed effect. As part of this, spans only report _newly installed_ effects by comparing to effects installed by the parent.

Before:

[asciicast](https://asciinema.org/a/0SjSbM47ihcQ17djhkLNohMHB)

After:

[asciicast](https://asciinema.org/a/bcwI6pCRmMgTOPT8kw5fH2R21)

## Service logs

Service logs are now always directed to the `asService` span, instead of a separate `start ...` span.

In the current model this is the only thing that really makes sense. You might think it makes sense to send logs to the service's `withExec`, but the issue is there's not always a `withExec`.

As an implementation detail, there is still a separate `exec ...` span, but it links to the `asService` span and correlates to the `withExec` span.

## Canceled state (...kind of)

Sorry, don't get too excited. I added a freebie here - the API will detect a canceled request and mark the request span canceled - but in practice there's more work to be done to reflect this in some scenarios. For example, canceling a `sync` won't propagate the canceled state to the things it runs, and since the `sync` call itself is hidden, you don't actually see a canceled state anywhere.

## Shorter exec errors for modules

Exec errors in modules no longer collect and return their stdout/stderr. With the TUI and Cloud UI these giant error messages are noisy and redundant. They are kept for the non-module code path (`dagger session`) because it's still a reasonable thing to do when you can't make as many assumptions about the UI.


vito:
@shykes @marcosnils I implemented a minimal API for error returns from functions. It was either that or keep doing janky workarounds so I figured I'd take a swing at it - which I was originally going to do in https://github.com/dagger/dagger/pull/7833 before getting stuck on other things.

The basic idea is to have functions return a proper `Error` type as opposed to just a string. By using a proper type we can support interesting things like adding stack traces, file/line information, maybe even a `Directory` containing screenshots, etc.

```graphql
type Query {
  # Construct an Error value to be returned by a function.
  error(message: String!): Error!
}

type Error {
  id: ErrorID!
  # A brief description of the error.
  message: String!
}

type FunctionCall {
  # Return an error from the function.
  returnError(error: ErrorID!): Void
}
```

Internally this works by having `returnError` write the `ErrorID` to a file in the caller's filesystem, so the same way that `returnValue` works. I briefly got stuck on the whole "Buildkit can't fail _and_ return something" - but then remembered we already dealt with this when collecting the `Stdout`/`Stderr` for `ExecError`, so it just uses the same trick: when we fail to evaluate the function, we rescue that value from the mount by directly loading it from the snapshot.

End result:

image

v0.13.0 for comparison:

image

The main question (for this PR): is this API agreeable enough to ship as an MVP?

---

Broader question: can we envision building on this to satisfy the "error reports" use case? For example, by supporting multiple errors, or hierarchical errors, or other extensions to the `Error` type? Or, do we think this is fundamentally off-base and we actually need to support 'arbitrary type + error' returns?

Putting my cards on the table: sticking with "return a value OR an error" seems familiar and conventional to me, so it'd be pretty convenient if every use case that motivated #8421 (lints, tests, ???) neatly fit into a rich `Error` API. But, that's just my current thinking - I see this more like the start of a beautiful bikeshed. 😄

Connecting dots:

* #8421

dagql/idtui/golden_test.go
```diff
@@ -0,0 +1,360 @@
...
+		err := warmup.Run()
+		if ex.Fail {
+			require.Error(t, err)
+		} else {
+			require.NoError(t, err)
+		}
+		// FIXME: there appears to be some delay before a cache is able to be hit
+		// by a separate session. no clue where this comes from (sorry) but a 10
+		// second wait passed 25 times in a row.
+		time.Sleep(10 * time.Second)
```
vito, line 139: @sipsma Any idea what might be causing this? tl;dr if you run two `dagger call` commands back to back quickly enough it's very possible for the second call to have cache misses, but if you wait ~10+ seconds you'll get cache hits.

sipsma: No immediate clue, that's very surprising and a great find. I don't see us doing anything asynchronously in our session shutdown: https://github.com/sipsma/dagger/blob/22e6902807be2bcb6f1503d9e1643494d9f578c5/engine/server/session.go#L283-L283

Highly possibly buildkit is throwing some cleanup into an async goroutine, but even then I didn't think there was any part of the caching specifically that would only kick in async like that... 🤔

Can you easily repro this with some `dagger calls` on your host?

vito: > Can you easily repro this with some `dagger calls` on your host?

Strangely enough, I can't. 🤔 And if I run these tests outside of Dagger against `./hack/dev` they pass just fine without the 10 second sleep. It _only_ fails when running in Dagger, against an engine running in Dagger.

I'm gonna keep poking at this since it's causing the tests to fail in CI. Will let you know if I find anything.

vito: Seems to be related to concurrency, somehow, even with seemingly non-overlapping pipelines.

Fails immediately without the 10 second wait:

```sh
# runs use-exec-service and use-no-exec-service in parallel
#
dagger-dev call test telemetry --run '//service'
```

Passes repeatedly:

```fish
# using a loop instead of --count to ensure a fresh engine each time
while dagger-dev call test telemetry --run '//use-exec-service'; and sleep 1; echo ok: (date); end
```

These two pipelines do not overlap at all - one descends from `redis` and the other descends from `alpine` and `busybox`:


```go
func (*Viztest) NoExecService() *dagger.Service {
	return dag.Container().
		From("redis").
		WithExposedPort(6379). // TODO: would be great to infer this
		AsService()
}

func (v *Viztest) UseNoExecService(ctx context.Context) (string, error) {
	return dag.Container().
		From("redis").
		WithServiceBinding("redis", v.NoExecService()).
		WithEnvVariable("NOW", time.Now().String()).
		WithExec([]string{"redis-cli", "-h", "redis", "ping"}).
		Stdout(ctx)
}
```

```go
func (*Viztest) ExecService() *dagger.Service {
	return dag.Container().
		From("busybox").
		WithExec([]string{"echo", "im cached for good"}).
		WithExec([]string{"echo", "im also cached for good"}).
		WithExec([]string{"echo", "im a buster", time.Now().String()}).
		WithExec([]string{"sleep", "1"}).
		WithExec([]string{"echo", "im busted by that buster"}).
		WithNewFile("/srv/index.html", "<h1>hello, world!</h1>").
		WithExec([]string{"httpd", "-v", "-h", "/srv", "-f"}).
		WithExposedPort(80).
		AsService()
}

func (v *Viztest) UseExecService(ctx context.Context) error {
	_, err := dag.Container().
		From("alpine").
		WithServiceBinding("exec-service", v.ExecService()).
		WithEnvVariable("NOW", time.Now().String()).
		WithExec([]string{"wget", "-q", "-O-", "http://exec-service"}).
		Sync(ctx)
	return err
}
```

Despite that, the issue only happens when they both run. And it fails reliably.

Overall there should be four sessions involved:

* A: use-exec-service warmup
* B: use-no-exec-service warmup
* C: use-exec-service test
* D: use-no-exec-service test

Normally A and B warm the cache so that when C and D run they have a predictable view of things for test assertions. Timing wise the two are within the same ballpark, one doesn't seem decidedly faster or slower than the other.

Still no idea what would cause this to happen, seems like parallel sessions is the only clue. CI has been failing even with the 10 second sleep, so it doesn't seem like I'm out of the woods yet.

vito: Seems interesting that I'm able to run these two `pending` examples in parallel just fine, but not in parallel with either of the service examples. Adding a service test to the mix makes both of them see cache misses.

* [✅ pending, ✅ pending-duped](https://dagger.cloud/dagger/traces/3452437bb958e46b07f5fe00b66fda38?span=dd5b48bf55c8e1ff)
* [❌ pending, ❌ pending-duped, ✅ use-no-exec-service](https://dagger.cloud/dagger/traces/57885b5e3bef087ad460f110798730fe?span=c8001d5688e967ba)
* [❌ pending, ❌ pending-duped, ❌ use-exec-service](https://dagger.cloud/dagger/traces/48b05cea960cb42524f49f7957c3a8a0?span=1db347d92155b690)

Maybe something gateway-related?

```go
func (*Viztest) Pending(ctx context.Context) error {
	_, err := dag.Container().
		From("alpine").
		WithExec([]string{"echo", "pending cached for good"}).
		WithExec([]string{"echo", "im also cached for good"}).
		WithEnvVariable("NOW", time.Now().String()).
		WithExec([]string{"sleep", "1"}). // wait a bit to help eyeballing
		WithExec([]string{"false"}).      // fail the pipeline
		WithExec([]string{"sleep", "1"}). // will be pending at the end
		Sync(ctx)
	return err
}

func (*Viztest) PendingDuped(ctx context.Context) error {
	_, err := dag.Container().
		From("alpine").
		WithExec([]string{"echo", "pending duped cached for good"}).
		WithExec([]string{"echo", "im also cached for good"}).
		WithEnvVariable("NOW", time.Now().String()).
		WithExec([]string{"sleep", "1"}). // wait a bit to help eyeballing
		WithExec([]string{"false"}).      // fail the pipeline
		WithExec([]string{"sleep", "1"}). // will be pending at the end
		Sync(ctx)
	return err
}
```

vito:
Have to run out, leaving a quick TODO for tomorrow:

* [ ] If no error returned by function, just return original error (currently returns a composite error)
* [ ] If test suite detects more than one trace, have it fail

dagql/idtui/golden_test.go
vito: For now, I've refactored the `CACHED` state tests out into separate examples (prob the right thing to do anyway) and configured them to just skip if a failure is detected (`Flaky: true`).

dagql/idtui/symbols.go
```diff
@@ -32,4 +32,5 @@ const (
 	IconSkipped         = "∅"
 	IconSuccess         = "✔"
 	IconFailure         = "✘"
+	IconCached          = "$" // cache money
```
jedevc, line 35: :money_with_wings: :money_with_wings:

dagql/idtui/testdata/TestTelemetry/TestGolden/cached-execs
```diff
@@ -0,0 +1,30 @@
...
+
+✔ prepare X.Xs
+
+✔ viztest: Viztest! X.Xs
+✔ .cachedExecs: Void X.Xs
+│ ✔ container: Container! X.Xs
+│ $ .from(address: "alpine"): Container! X.Xs CACHED
+│ │ ✔ resolving docker.io/library/alpine:latest X.Xs
+│ ✔ .withExec(args: ["echo", "cached-execs cached for good"]): Container! X.Xs
+│ ┃ cached-execs cached for good
```
jedevc, line 21: Why doesn't this get cached? :eyes: Is that the flake?

jedevc:
Some UI thoughts below:

- I like it :tada: Honestly, nothing huge worth changing, I think it's pretty much just straightforwardly better in every way.

- Not 100% sure about the newlines in the output, it does make the output a little trickier to track. I'm also struggling to understand the logic as to when they get spaced separately vs not? I think it's *this*: https://github.com/vito/dagger/blob/713a82cf6957cbdda18e4a1357c906f1f272ee95/dagql/idtui/frontend_pretty.go#L931-L939, but I'm not sure what the effect of that is? (comment maybe?)

  This also seems to result in this newline appearing/disappearing as children are added/removed? I think this contributes to a feeling that the new output is "jumpier"?

  That said, I do think it generally feels *nicer* spaced out (though `./hack/dev` is still pretty hard to read).

- Span rows seem to be removed at the beginning of a chain (as the spans complete and hit the gc threshold) before the end of a chain is completed?

  Again, this feels kinda jumpy, I guess it's not worse than before, but it means that as steps complete, you lose the cool chain context for why that step was even running in the first place. I think there's an argument for keeping the whole chain if any part of the chain is still pending/running?

- With the newlines in the output, doing a console output looks a bit odd now (there's lots of newlines between each `Container.*` call).:

  ```
  $ with-dev dagger call container-echo --string-arg=123
  ✔ connect 0.2s
  ✔ initialize 0.6s
  ✔ prepare 0.0s
  
  ✔ foo: Foo! 0.0s
  ✔ .containerEcho(stringArg: "123"): Container! 2.5s
  ✔ .defaultArgs: [String!]! 0.0s
  
  ✔ Container.mounts: [String!]! 0.0s
  
  ✔ Container.workdir: String! 0.0s
  
  ✔ Container.entrypoint: [String!]! 0.0s
  
  ✔ Container.platform: Platform! 0.0s
  
  ✔ Container.user: String! 0.0s
  
  _type: Container
  defaultArgs:
      - /bin/sh
  entrypoint: []
  mounts: []
  platform: linux/amd64
  user: ""
  workdir: ""
  ```

  Ideally the `Container.*` parts at the end could be hidden? Or somehow grouped? cc @helderco 

- Somehow HTTP HEAD errors have started flashing in very quickly:
  
  image

  Not quite sure why :thinking: 

- Could we have a newline before the menu line?

  image

  Spacing it out might feel a bit better?

core/error.go
```diff
@@ -0,0 +1,29 @@
...
+
+type Error struct {
+	Query *Query `json:"-"`
+
+	Message string `field:"true" doc:"A description of the error."`
+
+	// Stack    string `field:"true" doc:"The stack trace of the error."`
+	// Source   *File  `field:"true" doc:"The source file where the error occurred."`
+	// Line     int    `field:"true" doc:"The line in the source file where the error occurred."`
+	// Original *Error `field:"true" doc:"The original error that caused this error."`
```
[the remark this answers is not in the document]
vito: Yeah this is probably the most interesting design point to dissect. It seems like your design is more geared towards source locations within modules, whereas I went for `*File` with things like linters in mind, which might run against an arbitrary source `*Directory`.

I went for `*File` since it felt the most powerful; any system interpreting the error can just read the file contents and display the errored line. For example, the TUI could show you the relevant code region.

But it might be better to have a context `*Directory` and a relative path beneath that directory.

```go
SourceDir *Directory
SourcePath string
```

That way you are able to see the full path, which is often interesting. And if there's no applicable `Directory` (go stdlib) maybe it can be left blank. (Though ideally even in the Go stdlib case we'd have a `Directory` somehow.)

Also, as a north star when thinking about this, it'd be nice to support things like [GitHub actions annotations](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message) which require a path relative from repo root, which is only possible with a path - so that's another point against `File` I suppose.

> It would be cool if `Stack` was a `[]SourceMap`, then we can have a consistent stack trace between different languages, and can even combine/interleave them as needed 🎉

💯

core/service.go
```diff
@@ -409,6 +423,8 @@ func (svc *Service) startContainer(
 		forwardStderr(stderrClient)
 	}
 
+	var stopped int32
```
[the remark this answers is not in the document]
vito: TIL! Looked for that sort of thing via `atomic.Load<tab>` completion, didn't see the type and that it has its own methods.

Will change

core/util.go
```diff
@@ -28,6 +28,10 @@ type HasPBDefinitions interface {
 	PBDefinitions(context.Context) ([]*pb.Definition, error)
 }
 
+type HasPBOutput interface {
+	PBOutput(context.Context) (*pb.Definition, error)
+}
```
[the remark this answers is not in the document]
vito: Nope - dead code from a failed experiment, removing

core/telemetry.go
```diff
@@ -57,13 +83,19 @@ func AroundFunc(ctx context.Context, self dagql.Object, id *call.ID) (context.Co
...
-			// TODO maybe this should be an event?
+			// NOTE: this is never actually called on cache hits, but might be in the
+			// future.
 			span.SetAttributes(attribute.Bool(telemetry.CachedAttr, true))
 		}
 
+		if ctx.Err() != nil {
+			// If the request was canceled, reflect it on the span.
+			span.SetAttributes(attribute.Bool(telemetry.CanceledAttr, true))
+		}
```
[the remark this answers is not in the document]
vito: Don't think we need to; the only possible `ctx.Err()` is rooted in either `context.Canceled` or `context.DeadlineExceeded`, and we don't make a distinction for this status at the moment.

I suppose we could add a `TimeoutAttr` sometime but I'm not sure if it would even realistically be detected in all cases; from the this code's POV a timeout enforced client-side will always just look like a canceled request.

dagql/idtui/testdata/TestTelemetry/TestGolden/cached-execs
vito: Yep 😭 - see this thread: https://github.com/dagger/dagger/pull/8442#discussion_r1765037152

I could commit the non-flaked output but it'd be a Sisyphean task to keep it from just getting re-set to the broken output once someone runs `dagger call test telemetry --update -o .` and commits the result.

core/error.go
```diff
@@ -0,0 +1,29 @@
...
+
+	// what about screenshots and stuff?
```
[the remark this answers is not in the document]
vito: Hmm we could actually do this in a somewhat strongly typed way using IDs, since they embed their expected type. You could technically take any ID, context-free, parse it to discover the type, and then call `loadFooTypeFromID(id) { ... }` to interact with the object. 

But, I guess that only helps with storing objects, as opposed to scalars.

Really depends on what we need from this IMO in terms of real world use cases - I'm hesitant to jump straight to a generic map for example, since it's hard to do something high-level + useful for that (e.g. display screenshots for browser test failures in Cloud) without defining some sort of schema/semantic conventions. Might be better to just have hyper specific things like `ImagesDir *Directory`. Super low conviction though.

Tangent: in "pure GraphQL" the closest thing is just having a generic `ID` scalar type and doing something like this (see [Global Object Identification](https://graphql.org/learn/global-object-identification/)):

```graphql
{
  node(id: "DSFKjhsckdjhsdkjv") {
    ... on Directory {
      export(path: ".")
    }
  }
}
```

But `Node` interface is `id: ID!`, and we prefer `id: DirectoryID!` for codegen purposes, so we have all these `loadFooFromID` APIs instead.

cmd/codegen/generator/go/templates/modules.go
```diff
@@ -306,6 +311,14 @@ func dispatch(ctx context.Context) error {
 	setMarshalContext(ctx)
 
 	fnCall := dag.CurrentFunctionCall()
+	defer func() {
+		if rerr != nil {
+			if ` + voidRet + ` := fnCall.ReturnError(ctx, dag.Error(rerr.Error())); err != nil {
+				fmt.Println("failed to return error:", err)
+			}
+		}
+	}()
```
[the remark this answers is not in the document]
jedevc: Yeah, worth leaving till later!

core/error.go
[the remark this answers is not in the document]
jedevc: > But it might be better to have a context `*Directory` and a relative path beneath that directory.

That feels better to me, seems like it gets the best of both worlds? Not 100% how we'll get a `SourceDir` into something that's a `TypeDef` though. Maybe we could have like `SourceMap` know the `Module`/`ModuleSource` it was declared from? Then we could have a method to grab the `File`. Though now there's a circular dependency (is this an issue?)

I think we might need *something* like this anyways for supporting stack traces - the current module "name" I've implemented actually isn't good enough, since it's "local" to the module that consumes it - so stack traces that pass through multiple modules will not be super great.

I guess I'm suggesting altering #8498 from `Module string` to `Module *Module` - then we can have a `File` method easily.

---

Is the plan to do this in this PR, or to leave this commented for now and come back to it later?

core/error.go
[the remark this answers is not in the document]
jedevc: > Really depends on what we need from this IMO in terms of real world use cases

Yeah, agreed, happy to defer on this!

> Might be better to just have hyper specific things like `ImagesDir *Directory`. Super low conviction though.

Ooh, I do not like the idea of hyper-specific stuff in the API like that. Not quite sure what a alternative is though, hm, without punting all the hard problems into "cloud".

If we could though (:laughing:), maybe cloud could have a series of "matchers" to decide how to display generic dagger api objects?

- a `Container` gets a list of layers
- a `File` with a `.png`/`.jpeg`/etc filename extension gets displayed in an `` tag
- a `File` with a `.md` filename is rendered as a markdown file
- a `Directory` is just displayed as the sum of it's parts
- a future magical `Report` type gets rendered as a table

This would be *pretty* neat, and could also help solve how any return type could get rendered in cloud (which I feel like we're gonna need going into pipeline results around future planned ci stuff?).

Just brainstorming really, not really related to this issue anymore.

core/error.go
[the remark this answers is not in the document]
vito: The plan is to leave it commented for now - the scope here is already way too broad. Some of my plans around the `Error` type are also starting to change with side discussions (e.g. error hierarchies might be redundant with returning a magic `Report` type derived from OTel spans instead), but the minimal implementation here is still good, I think.

core/modfunc.go
```diff
@@ -361,6 +382,69 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Typ
 	return returnValueTyped, nil
 }
 
+func extractError(ctx context.Context, client *buildkit.Client, baseErr error) (dagql.ID[*Error], bool, error) {
```
sipsma, line 385: Not a blocker because this current approach works, but I do feel like we could skip all the shenanigans of writing the error back to the module container and then reading it from there. The reason we currently do that for the (non-error) return value was to be setup for caching function calls (since the output would be cached in the filesystem of the module container and thus integrated w/ buildkit's caching), but for error cases none of that should matter.

I think `ReturnError` could just store the error in the client state, where it can be retrieved by anything that needs it.

Again, not a blocker, just seems like it would probably be simpler and delete a few chunks of code.

TBH if we also stored the actual return value in the client state too so we could skip reading from the filesystem I wouldn't be shocked if we saved a just-barely-noticeable amount of overhead to function calls, though that's getting way out of scope, just musing...

core/service.go
```diff
@@ -359,6 +349,7 @@ func (svc *Service) startContainer(
 
 			res, err := bk.Solve(ctx, bkgw.SolveRequest{
 				Definition: def,
+				Evaluate:   true,
```
sipsma, line 352: I think this means that we will be solving each of these mount inputs one-by-one in serial, which would invalidate the later resolving of them in parallel: https://github.com/sipsma/dagger/blob/dc83928c3a13d6b315bf0953befde001ec359238/engine/buildkit/client.go#L223-L223

Somewhat obscure, but solving in parallel feels like an optimization worth keeping here if at all possible. Not sure how load-bearing setting Evaluate to true was.

sipsma (commented):
Reviewed an initial chunk, looks cool 😎

Also obscurely excited to have the telemetry testing you setup here, thank you for that! My otel metrics PR has a test but it's proving to be flaky by nature so I might rebase on this PR and just use the new report functionality here 🎉

vito:
Current status: sitting on this for a bit since I'd like its release to coincide with switching Cloud trace UI over to WASM, and there's a bit more work to be done there. Once we merge this the current React UI won't properly understand the new telemetry, so I'm in no rush to merge it, provided that rebasing remains pretty painless.

Leaving TODOs based on feedback:

* [ ] Avoid breaking pipeline chains per @jedevc's suggestion - I think it's a good idea, had a TODO at one point but might as well go for it.
* [ ] Evaluate Service mounts in parallel

core/modfunc.go
vito: That makes sense - OK to fix this up after? Leaning towards merging sooner to give it more time to "cook."

vito:
@sipsma @jedevc This should be good to go. There's more polish I could do but I think this PR has gone long enough at this point 😅

.dagger/test.go
```diff
@@ -41,6 +41,99 @@ func (t *Test) All(
...
+	// +optional
+	timeout string,
+	// +optional
+	race bool,
+	// +default=1
+	count int,
+	// +optional
+	verbose bool,
+) (*dagger.Directory, error) {
+	engine := t.Dagger.Engine().
```
jedevc, line 68: nit: a lot of this logic seems shared with `testCmd`?

engine/buildkit/telemetry.go
```diff
@@ -2,58 +2,83 @@ package buildkit
...
+	"go.opentelemetry.io/otel/trace/noop"
 
 	"dagger.io/dagger/telemetry"
 )
 
+func WithTracePropagation(ctx context.Context) llb.ConstraintsOpt {
+	mc := propagation.MapCarrier{}
+	telemetry.Propagator.Inject(ctx, mc)
+	return llb.WithDescription(mc)
+}
```
jedevc, line 26: Sorry, struggling to understand what the effect of calling this actually is :sweat_smile:

jedevc (approved):

dagql/dagui/opts.go
```diff
@@ -51,33 +56,46 @@ const (
...
+		return true
 	}
-	if opts.GCThreshold > 0 && time.Since(span.EndTime()) > opts.GCThreshold && opts.Verbosity < ShowCompletedVerbosity {
+	// TODO: avoid breaking chains
+	// if opts.TooFastThreshold > 0 &&
+	// 	span.ActiveDuration(time.Now()) < opts.TooFastThreshold &&
+	// 	opts.Verbosity < ShowSpammyVerbosity {
+	// 	// ignore fast steps; signal:noise is too poor
+	// 	return false
+	// }
```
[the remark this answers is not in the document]
vito: Yeah - now that we have a "pending" state we actually _want_ to show most spammy spans, otherwise you wouldn't see the pending state.

I'm going to experiment more with this - I think we probably want to hide entire pipeline chains that took less than some threshold, instead of individual spans.

.dagger/test.go
vito: Yeah could probably be refactored or done differently - I didn't want _all_ of `testCmd` because I wanted a faster feedback loop and that does some other stuff I don't need, like building an engine tarball. I spent time trying to figure out how to dedupe but think I ended up wasting too much time on silly things and decided to just duplicate for now.

jedevc:
FYI - I think some of https://github.com/dagger/dagger/pull/8442#issuecomment-2400324702 might have gotten buried, specifically noticing 

> Somehow HTTP HEAD errors have started flashing in very quickly:

(I was showing this to @levlaz who pointed this out as well)

.dagger/test.go
jedevc: No worries, I'll probably take a look at it at "some point in the future" if that's okay with you :saluting_face:

Issue #8441, opened 2024-09-12
🐞 `Directory.digest` on a scratch Directory causes a panic


pjmagee:
### What is the issue?

I don't think the CLI is supposed to return with a go panic. 

`panic while resolving Directory.digest: runtime error: invalid memory address or nil pointer dereference` followed by a large stacktrace

### Dagger version

dagger v0.13.0 (registry.dagger.io/engine:v0.13.0) windows/amd64

### Steps to reproduce

`dagger call directory` - happens on windows/linux afaik

I checked if a more 'legimate' usage was also broken, such as `dagger core directory with-new-directory --path /new` and that worked **fine**:

```
mageep@PATRICK-PC:~$ dagger core directory with-new-directory --path /new
✔ connect 1.0s
✔ initialize 0.1s
✔ prepare 0.0s
✔ directory: Directory! 0.0s
✔ Directory.withNewDirectory(path: "/new"): Directory! 0.0s
✔ Directory.digest: String! 0.0s
✔ Directory.entries: [String!]! 0.0s

Setup tracing at https://dagger.cloud/traces/setup. To hide: export NOTHANKS=1

_type: Directory
digest: sha256:62c8b61632016892fef530f9c1a5fa2cff01be2b8d3a895c43e92767f5ae46e1
entries:
    - new
```

### Log output

```
 patri   desktop-linux    Projects  dagger core directory                                          PS  23:02:06 
✔ connect 1.1s
✔ initialize 0.2s
✔ prepare 0.0s
✔ directory: Directory! 0.0s
✔ Directory.digest: String! 0.0s
✔ Directory.entries: [String!]! 0.0s                                                                                                                                                                                                                    Setup tracing at https://dagger.cloud/traces/setup. To hide: export STOPIT=1                                                
Error: response from query: input: directory.digest panic while resolving Directory.digest: runtime error: invalid memory address or nil pointer dereference

goroutine 730185 [running]:
runtime/debug.Stack()
        /usr/lib/go/src/runtime/debug/stack.go:26 +0x5e
github.com/dagger/dagger/dagql.(*Server).resolvePath.func1()
        /app/dagql/server.go:642 +0x77
panic({0x2064ee0?, 0x39baf60?})
        /usr/lib/go/src/runtime/panic.go:785 +0x132
github.com/dagger/dagger/core.(*Directory).Digest(0xc0045ac630, {0x27f07a8, 0xc0006592c0})
        /app/core/directory.go:187 +0x6f
[...]

PR #8445, opened 2024-09-13
fix: `Directory.digest` should not panic on scratch directory

jedevc:
Fixes https://github.com/dagger/dagger/issues/8441


jedevc:
Hm, test failures are legit, digging in.

sipsma (approved):

Issue #8329, opened 2024-09-04
`buildctl dial-stdio` has exited with exit status 137


jedevc:
Occasionally, this error appears:

```
Error: response from query: Post "http://:mem/query": command [docker exec -i dagger-engine-d522a778c2389390 buildctl dial-stdio] has exited with exit status 137, make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=
```

Seen in:
- https://discord.com/channels/707636530424053791/1003718839739105300/1280688927652450399 from @shykes 
- https://discord.com/channels/707636530424053791/1120503349599543376/1272596156848934973 by @sagikazarmark 
- https://discord.com/channels/707636530424053791/1263881471026925688/1268233582111887450

The `http://:mem/query` is very odd, but it seems to come from: https://github.com/dagger/dagger/blob/7dd8635b56a24b60a66b89f44c6aeedd48853ba8/engine/client/client.go#L1144

Not entirely sure why it's set to that (blame points to the git blame hole that is https://github.com/dagger/dagger/pull/5757).

idlsoft:
I'm getting this error consistently when I run on a Mac, but not on Linux.
Dagger 0.14.0.
Running:
dagger shell -q s1.dag

s1.dag:
```
#!/usr/bin/env dagger shell -q
ctr1=$(.container | from alpine | with-exec echo,hello,world)
..echo $ctr1 | stdout
```

Linux (works):
```
docker version
Client: Docker Engine - Community
 Version:           27.3.1
 API version:       1.47
 Go version:        go1.22.7
 Git commit:        ce12230
 Built:             Fri Sep 20 11:41:03 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          27.3.1
  API version:      1.47 (minimum version 1.24)
  Go version:       go1.22.7
  Git commit:       41ca978
  Built:            Fri Sep 20 11:41:03 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.23
  GitCommit:        57f17b0a6295a39009d861b89e3b3b87b005ca27
 runc:
  Version:          1.1.14
  GitCommit:        v1.1.14-0-g2c9f560
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

Mac (doesn't work, using Colima):
```
docker version
Client: Docker Engine - Community
 Version:           27.3.1
 API version:       1.47
[...]

PR #8463, opened 2024-09-16
chore: change DirectConn host to "dagger"

jedevc:
*Partially* fix for https://github.com/dagger/dagger/issues/8329.

The `:mem:` host here is *really* weird, I don't know the logic for it, but it makes for some very confusing error messages, so we should fix it to something more reasonable and recognizable.

### DIFF
--- a/engine/client/client.go
+++ b/engine/client/client.go
@@ -1148,7 +1148,7 @@ func (f DirectConn) Do(r *http.Request) (*http.Response, error) {
 }
 
 func (f DirectConn) Host() string {
-	return ":mem:"
+	return "dagger"
 }
 
 func (f DirectConn) Close() error {


sipsma:
I didn't come up with the name `:mem:` myself, I definitely saw it being used somewhere else (I thought buildkit somewhere) and just cargo-culted, but can't find it anymore.

Either way, it's not load bearing in and of itself. The only real-world-ish consideration here is that we ideally don't want a name that would ever accidentally trigger a DNS resolve of a real host (which could "upgrade" a plain old bug to become a security vulnerability).

This is very obscure, so if there's distaste for `:mem:` showing up then I have no problem changing to `dagger`. It would be nice if we could just leave the host blank and not care about this at all (since it's irrelevant), but last I checked the go http client got upset when you did that.

---

This PR is currently linked to #8329 such that the issue will be closed by it, which doesn't seem right to me, so unlinking it.

sipsma (approved):

Issue #8141, opened 2024-08-10
PHP: configure which exit codes prevent subsequent calls.


charjr:
## Real Example of Current Behaviour

When my source directory has formatting errors, if I call:

`dagger/sdk/php $ dagger call -m dev format --source=. export --path=.`:

The `format` command begins:
1. The formatter, PHPCBF, **executes successfully**. It formats the errors **but returns exit code 1**
    - This is a quirk of the formatter's exit codes, [it may be fixed in the next major version](https://github.com/PHPCSStandards/PHP_CodeSniffer/issues/184).
2. Execution stops, because it received an exit code 1.

## Desired Behaviour

1. Because I know what exit codes the formatter will return, I configure which exit codes will not halt further calls to dagger.
2. The formatter runs successfully, formatting the error and returning exit code 1.
3. The diff is returned and used in the call to `export`.

jedevc:
@charjr would https://github.com/dagger/dagger/issues/5981 resolve this? Or this entirely related to the internal behavior of the PHP SDK?

charjr:
> @charjr would #5981 resolve this? Or this entirely related to the internal behavior of the PHP SDK?

@jedevc resolving #5981 would also resolve this. 

More explicitly your PR #8466 would resolve this. 

(When I made the issue, I wasn't sure whether it was specific to the SDK or not)

Issue #5981, opened 2023-10-27
✨ AllowToFail in ContainerWithExecOpts


schlapzz:
### What are you trying to do?

In CI there are many jobs where you need to decide based on there return code  of a command (ok / not ok) how to proceed.

Simple example, upload the artefact if it is not already there If it exists, skip upload. 

The current workaround is as described in your docs
https://docs.dagger.io/cookbook#continue-using-container-after-command-execution-fails

Of course, the solution is working. But I have often the case that I want decide the further based on the exit code, and the code gets polluted by writing/reading the exit code into/from files. In view of a better DX it would be great if we can move this into Dagger and hide it from the developer.

For example something like that

```go
_, err := c.WithExec([]string{"curl -f https://example.com/file/v222"}, ContainerWithExecOpts{
		AllowToFail: true,
	})

if err != nil && !errors.Is(err, dagger.ExecNonZero){
  return err
}
``` 

### Why is this important to you?

* Improve the DX
* Avoid boilerplate code

### How are you currently working around this?

The current workaround is as described in your docs
https://docs.dagger.io/cookbook#continue-using-container-after-command-execution-fails

marcosnils:
> Of course, the solution is working. But I have often the case that I want decide the further based on the exit code, and the code gets polluted by writing/reading the exit code into/from files. 

this is already the case if you want to handle the exit code yourself. This guide (https://docs.dagger.io/cookbook#continue-using-container-after-command-execution-fails) is mostly for pipelines where some command "always" return a status code != 0 and you still need your pipeline to continue. 

Here's an example on how to handle status codes: 

```go
func main() {
	ctx := context.Background()
	client, err := dagger.Connect(ctx, dagger.WithLogOutput(os.Stderr))
	if err != nil {
		panic(err)
	}
	defer client.Close()

	_, err = client.Pipeline("test").
		Container().
		From("alpine").
		WithExec([]string{"sh", "-c", "exit 20"}).Sync(ctx)

	var execErr *dagger.ExecError
	if errors.As(err, &execErr) && execErr.ExitCode == 20 {
		fmt.Println("failed with 20")
	}
}

```

helderco:
> ref: #5539

That's also in the cookbook: [Handle exit code and unexpected errors](https://docs.dagger.io/cookbook/#handle-exit-code-and-unexpected-errors).

schlapzz:
This is what I need. Close the issue

jedevc:
[...]

PR #8466, opened 2024-09-16
feat: add `expect` to `Container.withExec` for custom return statuses

jedevc:
Fixes https://github.com/dagger/dagger/issues/5981
Fixes https://github.com/dagger/dagger/issues/8141

Pulls in a buildkit update as well, so we can get https://github.com/moby/buildkit/pull/5339


marcosnils:
thx for carrying this one over Justin! code LGTM. 

While reviewing this I was thinking that I'd love if there's a way I could just make ALL exit codes not to fail. I'd assume that there's going to be CLI's not very well documented that could potentially return different exit codes depending on the error which could potentially frustrate users having to reverse engineer it or dig into code to correctly parametrize here.

jedevc:
> While reviewing this I was thinking that I'd love if there's a way I could just make ALL exit codes not to fail

I did think about this - I'm not convinced we should do it though. Well-behaved programs *should* use exit codes to convey information on how the process failed (see https://tldp.org/LDP/abs/html/exitcodes.html). Specifically, `128 + <signal>` is the exit code if the process is killed by a signal. So if the process is `SIGKILL`ed, the exit status is 137. If this happens (e.g. an OOM), then this type of failure is *quite* different from "the command failed" - we don't want to allow just continuing on, this is probably very likely not the desired behavior.

So, maybe we could just catch ranges of codes, like `0-128` (excluding the sigals). But sadly, while the above is a convention that lots of programs will follow, there's no *guarantee* that 137 means a `SIGKILL` - any process can just `exit(137)`.

I think we should probably require that the selection of exit codes is explicitly listed out - if a user *really really* wants to have all the codes, they can pass an array of `0-255` in - but the fact that's quite difficult to do should imply that that's probably not what they should be doing.

jedevc:
As pointed out by @charjr and @vito, we probably should also add a `Container.exitCode` API alongside this, since now an exec has multiple possible exit codes, and it can be useful to determine what it was.

jedevc:
@helderco any ideas as to why python is failing here?

```
         try:
            result = await resolver.get_result(self._converter, root, inputs)
        except Exception as e:
>           raise FunctionError(e) from e
E           dagger.mod._exceptions.FunctionError: Function execution error: 'Argument validExitCodes does not exist in Field: Container!.'
```

jedevc:
In discord, I've had separate conversations with both @vito ([here](https://discord.com/channels/707636530424053791/1275216033862647890/1285283637599797268)) and @sipsma ([here](https://discord.com/channels/707636530424053791/911305510882513037/1286004869127213068)) about whether operations with an `exitCode` of non zero should cache or not.

We should probably bikeshed a bit, but here's my two cents:

- There are good reasons for both behaviors - because of this, in the future, ideally we should be able to mark *any* operation as do-not-cache after it's completion
  - For caching: some operations can return non-zero exit codes and succeed (e.g. tests that check for error behavior, which is a pass :heavy_check_mark:)
  - Against caching: it's weird to see flaky tests cache
  	- @shykes *did* point out though that if you've marked your tests to allow failing with exit code 1, usually because you might want to assemble and process a test report, you now actually have a very powerful primitive to handle your own flake logic - you can choose to retry certain tests, etc - this maybe makes sense to cache then.
- I think it makes *more* sense to cache as the default - doing otherwise means that now operations can continue on in the flow, but now won't cache - I think this feels like a trap for users? We shouldn't hide this behavior in `validExitCodes`, but expose it at a higher+clearer level for users.
- Implementing the do-not-cache after completion is *necessary* (for things like my idea for "super-impure" in https://github.com/dagger/dagger/pull/8358#issuecomment-2346006330) one day. But it's also quite tricky, we probably need to dive quite deep into buildkit, and I'm not sure what implications this has.

This does feel weirdly linked to https://github.com/dagger/dagger/issues/8421 as well - e.g. if you have a function that returns both a value and error (either bundled into the error object, or kept separate, however we decide), is this module call cacheable now? That said, I'm not quite it's the same - this is kind of more around providing a primitive in a core api call, while creating rich error types is more of a modules-y question.

helderco:
> @helderco any ideas as to why python is failing here?
> 
> ```
>          try:
>             result = await resolver.get_result(self._converter, root, inputs)
>         except Exception as e:
> >           raise FunctionError(e) from e
> E           dagger.mod._exceptions.FunctionError: Function execution error: 'Argument validExitCodes does not exist in Field: Container!.'
> ```

That just means there's a pipeline **inside a function** that returning that error from the API. What's the contents of the function that's returning this error?

helderco:
It seems to me that the failure happens when the Python test suite is run without a dev build of the engine, which is understandable. But does that make sense to you?

helderco:
It's actually not failing on an API request. It's failing because the Python query builder makes an introspection query to the API right after establishing a connection to the server:

https://github.com/dagger/dagger/blob/5e3299c08d62c3fc91d1ce02b79466ea20c02562/sdk/python/src/dagger/client/_session.py#L82

Then, when `with_exec` is used in a pipeline, I noticed the following line in the client binding has a bug that always adds mutable default arguments to the query:

```python
Arg("validExitCodes", [] if valid_exit_codes is None else valid_exit_codes),
```

So I fixed that bug and will be pushing it shortly.

shykes:
My issue with an array of exit codes, is that it makes it difficult to say "catch all errors". I think it's cool to have that level of control, but what I actually want is 1) an argument to say "always return control to me regardless of exit code" plus 2) a way to check the error code. If I have that, I can handle everything myself from there. The array feels unnecessary.

shykes:
Yeah but at that point, why have the array at all? If you let me catch all error codes, and know what error code I caught, I can implement anything I need on top of that.

sipsma:
> Yeah but at that point, why have the array at all? If you let me catch all error codes, and know what error code I caught, I can implement anything I need on top of that.

I think the biggest gap right now is that this all impacts caching too: if an exit code is ignored the exec op is still cached. So there would still be utility in only ignoring some exit codes vs ignoring all of them.

Obviously it would be ideal if we could also give cache control over all this but that requires rewiring the inner guts of caching in buildkit, so easier said than done.

shykes:
Ah I see. Not sure it makes a big difference since devs will just provide the array that corresponds to "catch all the app-level errors". So either way, all my app-level errors will be cached.

Re: deep in the buildkit guts. That's something I meant to ask. Don't we already have the ability to do this in the shim, instead of buildkit itself? Not sure it would change anything in the ability to cache control, but maybe?

sipsma:
> Re: deep in the buildkit guts. That's something I meant to ask. Don't we already have the ability to do this in the shim, instead of buildkit itself? Not sure it would change anything in the ability to cache control, but maybe?

No, we have control over the container execution but that's way below the stack in terms of caching. Components of buildkit that control caching call our worker (technically the correct term now since the "shim" is gone), but we can't control the caching from there.

We sort of have a sandwich going on where dagger is a top-level frontend and a low-level executor, but buildkit caching lies in the meaty center and we don't really have direct control over its mechanisms yet.
* I wouldn't rule out some hack being possible but also would not advise that generally

jedevc:
> The array feels unnecessary.

I previously put some points in https://github.com/dagger/dagger/pull/8466#issuecomment-2355244959 about this.

Here's some reasons why it makes sense to me:

- I have a step that will always fail, and return exit status 1. We have a few of these in our integration tests, and users who write tests for their modules might use this to confirm that certain things definitely fail. Having an array lets me easily express this - without breaking the chain. If I return all the exit codes, and have to do manual work after to parse them, then I have to have a `context`, I break the chain, and have to insert a lot of extra logic :cry: 
- I don't actually usually want to catch *all* the exit codes - some exit codes (like `SIGKILL` 137) are truly exceptional - I don't think the user should be manually handling issues that could be caused by the underlying infrastructure (e.g. I run out of memory, OOMKiller steps in and `SIGKILL`s a `withExec`). Having the user manually work with these feels like now the user is exposed to details of where and how dagger runs. Most of my well-behaved tools are going to return `0` on success and `1` on failure, so I expect most of these usages will use `validExitCodes: [0, 1]`.
	- Do you have a concrete use case for when you do actually want to catch *every* possible reason that the process exited?
- Caching is a real problem here (as pointed out by @sipsma) - we really shouldn't be caching environmental failures (like the above `SIGKILL` from the OOMKiller) - the problem is, it's very hard to determine what an environmental failure is, and how it presents.
	- It's not just signals, e.g. a process that follows [sysexits.h](https://github.com/openbsd/src/blob/master/include/sysexits.h) might exit with code `69` to indicate a "service unavailable" error - this is transient, and so we wouldn't want to catch that.
- `ExecError`s are important - catching all the signals means we never throw an `ExecError` for that step (even for really exceptional cases), which means that `-i` interactive debugging won't work, we won't be able to display the pretty errors from #8442, and at some point it might hurt observability in cloud if we add better `ExecError` displays there.

shykes:
> > Having to figure out what exit codes to catch in the array, by a painful process of trial and error, won't make the caching situation any better, will it?
> 
> I suspect that _most_ tools will be somewhat well behaved, and will either return an exit code of 0 or 1.

That is not an assumption we should make. The Unix convention is that non-zero exit code means an error. The bare minimum feature is for Dagger to make it easy to catch non-zero exit codes. Sure, sometimes tools explicitly document which exit code they will return; sometimes they even use it to communicate information about the *type* of error. But just as frequently, the tool is a black box and the only fact you can rely on for sure, is that non-zero = error.

So to say "catch all non-zero exit code" is not a niche special case: it's the primary case. Knowing exactly which exit codes your tool will return, and crafting an array based on that knowledge - that's the special case.

> > catch all errors, and let me process them
> 
> I guess I'm just a little confused as to why we would want to do this - _all_ errors could include a bunch of very strange low-level system errors.

I don't mean all Dagger errors. I mean all non-zero exit codes. See above for why that's not strange.

> * We keep `validExitCodes` array as currently

May I propose `catchExitCodes` instead. "Valid" is confusing to me.

> * We add a new `allExitCodes` boolean, which catches 0-127 exit codes (not higher, those are magical signal exit codes as mentioned above, catching these by default is really not a good idea)

Similarly: `catchErrors`? I want to make that nice and simple since it will be what everyone uses :-P

> > I have a step that will always fail, and return exit status 1. We have a few of these in our integration tests, and users who write tests for their modules might use this to confirm that certain things definitely fail. Having an array lets me easily express this - without breaking the chain. If I return all the exit codes, and have to do manual work after to parse them, then I have to have a context, I break the chain, and have to insert a lot of extra logic 😢

I think we need a reference test case. How will you implement a `go test` which catches failed tests, to report them in a "report" object rather than a dagger error?

Here's what I'm picturing:

```golang
func (m *MyModule) Test(src *dagger.Directory, ctx context.Context) (*Report, error) {
  ctr := dag.Container().
    From("golang").
    WithMountedDirectory(".", src).
    WithExec([]string{"go", "test", "./..."}, dagger.ContainerWithExecOpts{
      CatchErrors: true,
    })
  code, err := ctr.ExitCode(ctx)
  if err != nil {
    return nil, err
  }
  report := new(Report)
  if code != 0 {
    report.Success = false
  }
  return report, nil
}
```

In this scenario, checking for exit code is kind of the point. Yes it breaks the chain - just like checking for stdout - but I'm at the end of the chain anyway.

This is the use case I have in mind when I make the arguments above, about catching non-zero exit codes as the primary use case. I don't care what specific exit code `go test` returns. I know it's non-zero, because that's the convention. There's no benefit for me in making it more specific. It's more work to find out, and if they change the convention of their tool, it breaks my module.

In some cases, like `curl` there's a legitimate reason to do it. But in that case, I can just replace `if (code != 0)` with `if (code == 22)` or whatever the specific code is. You may have another use case in mind - but in mine, I'm pretty sure that array is useless.

jedevc:
> I think we need a reference test case. How will you implement a go test which catches failed tests, to report them in a "report" object rather than a dagger error?

For this case, `catchErrors` makes more sense.

But for a program that I expect to *always* fail, then I want the array to specify the code that I want it to fail with.

With the array (producing a valuable `ExecError` for readability in case of failure):

```go
ctr.
	WithExec([]string{"command", "that", "fails"}, dagger.ContainerWithExecOpts{CatchExitCodes: []int{1}}).
```

Without the array (which also loses the readability):

```go
code, err := ctr.
	WithExec([]string{"command", "that", "fails"}, dagger.ContainerWithExecOpts{CatchAllExitCodes: true}).
	ExitCode(ctx)
if code != 1 {
	return err
}
```

---

Maybe instead of an array we could have an enum? `Expect` - which takes a value `Success`, `Failure`, `Any`? That's zero, non-zero, and all. Handles all the use cases, but means that you can't distinguish between individual exit codes without getting `.ExitCode`, but maybe that's okay.

shykes:
If we do `catchErrors: bool` and `catchExitCodes: <array>` now, and add `Container.exitCode` as a rapid follow-up, is that ok with you? I don't mind the array being there, as long as the "go test" case has a really nice and simple solution.

jedevc:
`Container.exitCode` is actually in here already - no need for a follow-up.

---

Having had the weekend to think about this a bit more, I think I like your original proposal of one `catchErrors`, I'm less sure about having a big array - but I would really like a way to *just* catch failures (and fail on `0`).

Could we have an `Expect` value as I mentioned above? So:

```go
# Success (default) - dagger expects the command to success with exit code 0
ctr.
	WithExec([]string{"command"}, dagger.ContainerWithExecOpts{Expect: dagger.Success)

# Failure - dagger expects the command to fail with any exit code 1-127
ctr.
	WithExec([]string{"command"}, dagger.ContainerWithExecOpts{Expect: dagger.Failure)


# Any - dagger doesn't expect anything, the exit code is discarded
ctr.
	WithExec([]string{"command"}, dagger.ContainerWithExecOpts{Expect: dagger.Any)
```

With *all* of the above, you can chain this into `ExitCode`:

```go
exitCode, err := ctr.
	WithExec([]string{"command"}, dagger.ContainerWithExecOpts{Expect: dagger.Any).
	ExitCode(ctx)
```

This covers the main use cases:
- Standard "exit 0" behavior that everything has today
- Failure mode, that covers inverting a status code (to ensure a command fails)
- Anything mode, that covers building your own report

If the user requires more precise covering of exit codes, I think it's fine to force the user to use `exitCode` (we don't need the array). Does this model work?

If it does, I can update the PR in a few minutes, we can ship it whenever.

shykes:
Yeah that looks great @jedevc ! I say ship it

docs/docs-graphql/schema.graphqls
```diff
@@ -2791,6 +2801,13 @@ type Query {
   version: String!
 }
 
+"""Expected return type of an execution"""
+enum ReturnType {
+  Success
+  Failure
+  Any
```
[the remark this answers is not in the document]
jedevc: Yikes, yup. Got caught out by copy-pasting from above: https://github.com/dagger/dagger/pull/8669#discussion_r1820892962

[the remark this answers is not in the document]
shykes: Should we prefix it with `Expect` to avoid having a super-generic type lying around? ie `EXPECT_ANY`, `EXPECT_SUCCESS` etc? Or do we plan on using these types for different purposes?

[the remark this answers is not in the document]
helderco: They're scoped by the enum and #8669 is the fix that scopes the enum values to their enum types.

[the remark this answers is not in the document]
jedevc: We need to solve this in https://github.com/dagger/dagger/pull/8669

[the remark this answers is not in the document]
jedevc: A note on choosing `ReturnType` as a name instead of `Expect` in the name: I think there's potential applications in other places.

For example: we could have a `restartOn=ReturnType.Success` for a dagger service that would automatically restart a service if it stopped for some reason. Not saying we should do *that* feature, but `ReturnType` is more generic in other places than `ExpectType` or similar.

core/schema/container.go
```diff
@@ -489,6 +492,10 @@ func (s *containerSchema) Install() {
 			Doc(`The error stream of the last executed command.`,
 				`Will execute default command if none is set, or error if there's no default.`),
 
+		dagql.Func("exitCode", s.exitCode).
+			Doc(`The exit code of the last executed command.`,
+				`Will execute default command if none is set, or error if there's no default.`),
```
helderco, line 497: The "execute default command" part was only true before v0.12.0. Since this is a new field, it won't even show in that version.

helderco: Hmmm, I notice `stdout` and `stderr` also have it for `AllVersion`. 🤔

helderco: Seems like an oversight in https://github.com/dagger/dagger/pull/7857/files#diff-89b0b6c38f29bb6615327b6126c2479d29b032ef974017a9358c0c0721fa17f9

helderco: Should say:
```diff
-				`Will execute default command if none is set, or error if there's no default.`),
+				`Returns an error if no command was set.`),
```

jedevc: Ah, sorry, github didn't refresh the last message :cry: looks like we realized the same thing.

Will update all the messages.

docs/docs-graphql/schema.graphqls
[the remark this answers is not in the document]
helderco: To be fair, from an API POV the enum values are global it seems. So you won't be able to have another enum with the `ANY` value for example.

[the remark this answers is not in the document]
jedevc: Wait they're global? :scream: If so, that feels like a big issue.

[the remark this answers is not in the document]
jedevc: Phew, okay. I was operating under the assumption that you could have duplicate enum values at the graphql layer - you always know what type you need to unmarshal into, so it should be fine.

docs/docs-graphql/schema.graphqls
```diff
@@ -2837,6 +2847,13 @@ type Query {
   version: String!
 }
 
+"""Expected return type of an execution"""
+enum ReturnType {
+  SUCCESS
+  FAILURE
+  ANY
```
helderco, line 2854: Isn't it useful to document these values? We don't seem to do it for the core types, but we do in `TypeDefKind` for example. Would show in user's IDE.

cmd/dagger/module.go
```diff
@@ -1116,6 +1116,7 @@ var skipLeaves = map[string][]string{
 		// stdout and stderr may be arbitrarily large and jarring to see (e.g. test suites)
 		"stdout",
 		"stderr",
+		"exitCode",
```
helderco, line 1119: I think you can leave `exitCode` out since it doesn't trigger execution anymore. We only left `stdout` and `stderr` anyway due to potentially large output.

sdk/python/src/dagger/client/gen.py
```diff
@@ -436,7 +446,8 @@ def as_tarball(
         _args = [
             Arg(
                 "platformVariants",
-                [] if platform_variants is None else platform_variants,
+                () if platform_variants is None else platform_variants,
+                (),
```
helderco, line 450: Got confused because I remembered fixing this! Took me a while to realize I fixed this **bug** in this PR 😂

helderco (approved):

docs/docs-graphql/schema.graphqls
jedevc: Oop, yup!

cmd/dagger/module.go
jedevc: :tada: added a comment.

charjr:
@jedevc Bit late to the party but `expect` is brilliant. Thank you very much! Looking forward to making use of this :rocket:

PR #8467, opened 2024-09-16
fix: TS regression introduce by grapql-request dep bump

TomChv:
Rawcode spotted an issue https://discord.com/channels/707636530424053791/1284180822135406656 where traces where no longer published to traces.
The issue is introduced by a change of the type from a plain JSON object to `Headers` which break the current opentelemetry propagator. This PR fixes that issue by using a custom injector to set the `traceparent` header


vito (approved):
👍 - sneaky. I'll be sure to add coverage for this in the new telemetry suite.

Issue #8464, opened 2024-09-16
🐞 dagger call Secret reveals plaintext


wanderingkaiju:
### What is the issue?

I have a function that returns `Secret[]` and when I call it from the CLI it prints all the secret material to the terminal.

```
dagger call secrets

- _type: Secret
  name: cloudflare.apiToken
  plaintext: XXX
- _type: Secret
  name: restate.host
  plaintext: XXX
- _type: Secret
  name: restate.apiToken
  plaintext: XXX
- _type: Secret
  name: restate.identityToken
  plaintext: XXX
- _type: Secret
  name: restate.adminApiToken
  plaintext: XXX
- _type: Secret
  name: turso.platform.apiToken
  plaintext: XXX
- _type: Secret
  name: turso.groups.platform.apiToken
  plaintext: XXX
```

### Dagger version

v0.13.0

jedevc:
Mm, we need a way of marking these fields that they shouldn't be printed out.

I think we should have a `dagql.Sensitive` or similar for items in our server - then we can put this into the typedefs api, and fields that have this should be hidden on the CLI.

This is probably better than a generic "don't show", since `Sensitive` is more specific for this case, and it would be nice if we could show `plaintext: ***`.

cc @helderco does that make sense to you? Am happy to work on this for the release.

jedevc:
Or, I wonder if we should just hook up the `SecretCensor` to the output here :thinking: that would also be cool.

helderco:
Hey, I made a quickfix in:
- https://github.com/dagger/dagger/pull/8468

> [...] it would be nice if we could show `plaintext: ***`.
> 
> cc @helderco does that make sense to you?

I don't see the point since it's not useful information.

sipsma:
I think Helder's fix makes the most sense, it's just not something that should be printed automatically by object field auto-selection.

I would say there are probably use cases for a user very explicitly asking for the plaintext of a secret, but that should only be when they explicitly select the `plaintext` field in their dagger call, never automatic. So whatever we do we probably want to preserve that as a feature (I believe Helder's PR does preserve that).

PR #8468, opened 2024-09-16
fix(cli): fix secret plaintext being printed on screen

helderco:
Fixes https://github.com/dagger/dagger/issues/8464


sipsma (approved):
LGTM, agree test is worth adding though too

jedevc:
Approach SGTM, but I think there's still an underlying issue - if the user explicitly calls `plaintext` on the secret, the secret still ends up in the otel trace.

I think we might also want to have the OTEL trace logs scanned and filtered.

helderco:
> Approach SGTM, but I think there's still an underlying issue - if the user explicitly calls `plaintext` on the secret, the secret still ends up in the otel trace.
> 
> I think we might also want to have the OTEL trace logs scanned and filtered.

True, but that's a separate PR.

jedevc (approved):

helderco:
Added tests.

PR #8470, opened 2024-09-16
engine: change go sdk cache mounts to be locked mode

sipsma:
Trying out a possible mitigation to https://github.com/dagger/dagger/issues/8031 (which is pending on [an upstream go issue](https://github.com/golang/go/issues/69179)).

This changes the cache mounts for the Go SDK to be mode Locked, rather than Shared. This means that only one invocation of the Go SDK can run at a time (though the cache mount will still be re-used across different invocations).

The Go bug is not root caused, so it remains to be seen if this even actually avoids the problem. Will require a bunch of re-runs of CI in this PR.

*If* it does seem to disappear, then we have may a somewhat tough decision to make since it's highly likely that this change will hurt performance:
* On one hand, obviously we want to this bug to be mitigated
* On the other hand, we have only seen this in our own CI, so it's mostly likely highly rare for our real-world users to hit it. So it's unclear if the performance hit is really worth it.
   * One possibility would be to make this configurable somehow on the engine and only enable locked mode in our CI.


sipsma:
Hmm... all the CI test workflows timed out with changing the mode to Locked, so I'm gonna rule that out as non-viable in terms of the performance impact for now.

I will try mode Private instead

sipsma:
There does seem to be meaningful slowdown with Private cache volume mode, though not as devastating as with Locked (some workflows seem like a minute slower, though noisy so hard to really tell yet).

@rajatjindal had the good suggestion to try disabling the module index as go devs did in the upstream issue (https://github.com/golang/go/issues/69179#issuecomment-2323300381) so will give that a shot too to see what the practical impact is

sipsma:
Hm disabling the module index doesn't appear to have much effect on the runtime across a few runs so far 🤷‍♂️🎉

[I asked the go devs](https://github.com/golang/go/issues/69179#issuecomment-2359644134) what the implications are of `GODEBUG=goindex=0` in the upstream issue; it appears to be an undocumented setting (unless my googling is failing me).

I'm a little wary of setting undocumented feature-flags in our production code, but we *could* make this a system env var (similar to our existing support for `GOPROXY`) and just set it in our CI for now?

I'll setup the plumbing for that here and do some more re-runs while waiting to hear back from them.

rajatjindal:
:shipit:

jedevc:
Yeah, glancing through the source code for the mod index (which is wayyyy more complex than I have time to dive into right now), it's just a cache for https://pkg.go.dev/go/build#Context.Import - if the cache is disabled, we just fetch this data from the filesystem each time.

If we don't see a meaningful performance drop in the test runtime, no objection here, since the flakes are reasonably blocking - though obviously let's follow the upstream issue and help out where we can.

---

One thing that did stand out weirdly to me though: https://github.com/golang/go/blob/64a5d1d7de14ad74b1a77614de0e17c659ef12b6/src/cmd/go/internal/modindex/read.go#L73-L77

This feels very very similar to the case we saw for uv in https://github.com/dagger/dagger/issues/7662#issuecomment-2312903772. Because everything is mounted at `/src` (the value of `goSDKUserModContextDirPath`), then I guess it's possible that we produce a *ton* of similar hashes? And then potentially something gets into a bad state from there (the docs for `protect`/`unprotect` do state that it might not be used for trivial accesses?).

@sipsma WDYT about *trying* our python fix for go, and seeing if that improves the flakes? That feels potentially a bit less invasive, and also if it *is* that, would help narrow down the go investigation as well?

Approving regardless though, will let you make the call here!

jedevc (approved):

sipsma:
@jedevc good find, that's definitely suspicious! I think what I'll do is merge this one as is and open a follow-up PR where I try replacing this approach with that one. The main reasons being that A) I'd like to get a potential fix out ASAP and B) I think the other approach actually is (debatably) more invasive in that it's a meaningful change to the production code (whereas this one does it in such a way that only our test suite will have the module index disabled for now).

Issue #8254, opened 2024-08-28
AsModule & Initialize not Properly Cached within the same session


aluzzardi:
We are calling two functions after `AsModule().Initialize()`, and the second one is not hitting cache:

```go
mod := modConf.Source.AsModule().Initialize()

err = mod.Serve(serveCtx)
name, err := mod.Name(ctx)
```

https://dagger.cloud/dagger/traces/fefb70456e44a55fad89198adf9ef3ac?span=845b4c6062bbfaba

- **Serve**
  - **AsModule**: *6.4s*
  - **Initialize**: *2.1s*
- **Name**
  - **AsModule**: *501ms*
  - **Initialize**: *1.6s*

Additional context:

https://discord.com/channels/707636530424053791/1275839141929680957/1278063586786148353

/cc @sipsma @vito

sipsma:
I didn't get time to dive into before PTO, so just leaving a few breadcrumbs if anyone picks it up before I'm back:
1. I'm confused why the impure API runs twice despite https://github.com/dagger/dagger/issues/7788 being fixed
   * The second run is only adding like 30ms overhead atm, but if this wasn't happening we wouldn't be reloading any context and would be side-stepping this problem. So worth understanding at some point
2. The main problem seems to be some non-determinism in loading the local source blob, things to look at:
   * There's some maps being used in the implementation ([here](https://github.com/sipsma/dagger/blob/134a8aa4cb518c3cacc74b0ba1c01bb679f92ddf/core/schema/modulesource.go#L779-L784) and [here](https://github.com/sipsma/dagger/blob/134a8aa4cb518c3cacc74b0ba1c01bb679f92ddf/core/schema/modulesource.go#L1066-L1066), maybe elsewhere too). Worth checking if they are being iterated over in a way that results in some non-determinism
     * Tangentially related issue: https://github.com/dagger/dagger/issues/6902
   * We can just do a diff of the blobs to debug what's inconsistent. The `Diff` API obviously would help; if something fancier is called for I've wanted to try https://diffoscope.org for a while now to get fine-grain info on precisely what is different between two dirs

marcosnils:
@aluzzardi have you checked if this improved after v0.13.1? As Erik mentioned, we think #8471 should have fixed some of these aspects :pray:

helderco:
@sipsma, this is no longer a thing since there's no more "Module.initialize" function, right?

sipsma:
[...]

PR #8471, opened 2024-09-16
engine: set include and exclude globs deterministically

marcosnils:
change includeSet and excludeSet to use a slice without duplicates given
that the order of the globs is important when calling buildkit.


sipsma (approved):
LGTM, this looks like it addresses part of what I mentioned in this comment: https://github.com/dagger/dagger/issues/8254#issuecomment-2322753408

Not sure if that entire issue is solved yet, but probably at minimum improved 🎉

---

It would be nice to have a test here, but it's really quite difficult to test in a deterministic way, so I'm fine merging as is for now.

Issue #8349, opened 2024-09-05
Occasional flake in `TestPlatform/TestCacheMounts`


jedevc:
Mostly opening this one for tracking, I've seen this a few times now:

```
	Error Trace:	/app/core/integration/platform_test.go:186
	Error:      	Received unexpected error:
	            	input: container.from.withMountedCache.withExec.sync resolve: process "sh -x -c cat /cache/s2r819a2l1q07r1x64xvzwajylinux/amd64/uname | grep 'x86_64' && cat /cache/s2r819a2l1q07r1x64xvzwajylinux/arm64/uname | grep 'aarch64' && cat /cache/s2r819a2l1q07r1x64xvzwajylinux/s390x/uname | grep 's390x'" did not complete successfully: exit code: 1

	            	Stderr:
	            	+ cat /cache/s2r819a2l1q07r1x64xvzwajylinux/amd64/uname
	            	+ grep x86_64
	            	cat: can't open '/cache/s2r819a2l1q07r1x64xvzwajylinux/amd64/uname': No such file or directory
	Test:       	TestPlatform/TestCacheMounts
```

A failed trace: https://dagger.cloud/dagger/traces/db059c4ff25c8b8b5bfe683ecc1e22ea?span=9adaaf6bcd75cc09#9adaaf6bcd75cc09

Not 100% sure how this can happen - it looks like content is going missing from a cache mount while the test is running?

sipsma:
Another recent occurrence: https://dagger.cloud/dagger/traces/7126092f67cc23b68c661cc42ec1643b?span=c8ff2f8b862d0f1c#250fccb7b7b646b3

Will take a look at this as part of https://github.com/dagger/dagger/issues/8184

sipsma:
Might be fixed by https://github.com/dagger/dagger/pull/8474 but will wait until we haven't seen it for a while 🤞

sipsma:
I haven't seen this one since, so closing for now

sipsma:
This one happened once again 🤔 https://dagger.cloud/dagger/traces/e43deb6cdb405124a6f2a4b1bf391a43

It's the only time I've seen it in months, so not going to start actively looking into again yet, but will reopen for tracking in case it's problematic again.

[...]

PR #8474, opened 2024-09-17
tests: deflake TestPlatform/TestCacheMounts

sipsma:
This is all theoretical still, but the flakes we see in the test would be explained by actual expected behavior: cache mounts are best effort and free to be pruned at any time. If the cache mount used in this test is pruned in the middle of it (between execs that use it), then we'd expect to see the errors we're getting of `No such file or directory`.

This attempts to deflake the test by running an exec in the background that has the cache volume mounted throughout the test.

Other possible mitigations were even more complicated (i.e. run a separate engine with configuration that disabled cache volume pruning, etc.), so while this feels a little weird I can't think of anything better at this time other than removing the test entirely.

---

Fixes (theoretically) https://github.com/dagger/dagger/issues/8349, but phrasing this sentence in such a way that GH won't auto close that issue yet since I want to keep it open until we've run for a while without seeing this 😄

### DIFF
--- a/core/integration/platform_test.go
+++ b/core/integration/platform_test.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


core/integration/platform_test.go
```diff
@@ -184,6 +203,9 @@ func (PlatformSuite) TestCacheMounts(ctx context.Context, t *testctx.T) {
 		WithExec([]string{"sh", "-x", "-c", strings.Join(cmds, " && ")}).
 		Sync(ctx)
 	require.NoError(t, err)
+
+	cancel(cancelErr)
```
jedevc, line 207: do we also need to call `cancel` on `t.Cleanup`?

jedevc (approved):
Nice :tada:

core/integration/platform_test.go
sipsma: Yep good call, fixed

PR #8476, opened 2024-09-17
feat: package manager benchmark job

TomChv:


aluzzardi:
The benchmarks are automatically uploaded in OTLP (https://github.com/dagger/dagger/blob/main/.github/workflows/trace-workflows.yml) and charted in Honeycomb.

Please make sure your changes are correctly named in that context as the naming is crucial for charting.

The current job looks like this:

Screenshot 2024-09-17 at 5 04 52 PM

I would suggest (but up to you to verify things work correctly):

- Remove `typescript` from the SDK matrix as your job is already covering that. Otherwise we'll have data point duplicates
- Make sure your matrixed jobs have properly distinguishable names (e.g. `
dagger init (typescript::pnpm)` perhaps?)
- Since typescript is a special case and now we're only left with Go and Python, **maybe** it's not worth keeping the matrix above just for that (e.g. split Go and Python into their own jobs)?
  - If so, make sure the job names make sense (and maybe are the same as before, e.g. `Benchmark (Python)`)

TomChv:
@aluzzardi How can I verify that it works correctly? Since the job only run on main

TomChv:
@aluzzardi I refactored the job based on your comment, how can I test it?

aluzzardi:
I think you can (temporarily) change the job so it runs on PRs (this PR) and see what happens

TomChv:
To apply that, it needs to be published on main, so I'm not sure I can do that

TomChv:
```
git push -f --set-upstream origin feat/package-manager-benchmark
Enumerating objects: 214, done.
Counting objects: 100% (214/214), done.
Delta compression using up to 10 threads
Compressing objects: 100% (105/105), done.
Writing objects: 100% (137/137), 34.06 KiB | 8.51 MiB/s, done.
Total 137 (delta 103), reused 54 (delta 29), pack-reused 0
remote: Resolving deltas: 100% (103/103), completed with 67 local objects.
remote: error: GH013: Repository rule violations found for refs/heads/feat/package-manager-benchmark.
remote: Review all repository rules at http://github.com/dagger/dagger/rules?ref=refs%2Fheads%2Ffeat%2Fpackage-manager-benchmark
remote: 
remote: - Cannot create ref due to create name restrictions.
remote: 
To github.com:dagger/dagger.git
 ! [remote rejected]     feat/package-manager-benchmark -> feat/package-manager-benchmark (push declined due to repository rule violations)
error: failed to push some refs to 'github.com:dagger/dagger.git'
```

TomChv:
@helderco helped me found a way to test this Pr, will do some tests before merging it

TomChv:
I duplicated my dashboard to show my changes: https://ui.honeycomb.io/dagger/environments/experimental/board/jFXtYLadoUi

Screenshot 2024-09-19 at 17 29 13

gerhard (approved):
LGTM 👍

Merging now sync so that we can fix it in case something breaks.

PR #8479, opened 2024-09-17
feat(engine): add 'container.up' api

### LINKED ISSUE #7399: Container.up()
## Problem

When I call a function returning a `Container`, to run that container as a service, I have to run:

```console
dagger call <MYFUNC> as-service up
```

That is 2 extra functions in the chain, which is cumbersome. It's common to write a wrapper function that returns a `Service`, for the sole purpose of typing:

```console
dagger call <MY_OTHER_FUNCTION> up
```

But now I have to maintain one more function, with another name. It's also cumbersome and feels unnecessary.

## Solution

Implement `Container.up()` as an alias for `Container.asService.up()`.

This is similar to `Container.directory()` for example, which is a thin wrapper for `Container.rootfs().directory()`.

rajatjindal:
fixes #7399


rajatjindal:
requesting review from @shykes @jedevc @marcosnils

rajatjindal:
woohoo !!, all the tests are green.

rajatjindal:
I think I have fixed all the flakes now in the `DaggerUp` test case and this should be ready for review/merge now. thanks

core/schema/service.go
```diff
@@ -21,6 +21,10 @@ func (s *serviceSchema) Install() {
 		dagql.Func("asService", s.containerAsService).
 			Doc(`Turn the container into a Service.`,
 				`Be sure to set any exposed ports before this conversion.`),
+		dagql.NodeFunc("up", s.containerUp).
+			ArgDoc("random", `Bind each tunnel port to a random port on the host.`).
+			ArgDoc("ports", `List of frontend/backend port mappings to forward.`,
+				`Frontend is the port accepting traffic on the host, backend is the service port.`),
```
jedevc, line 27: Needs a `Doc(...)` call here.

.changes/unreleased/Added-20240917-214812.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Added
+body: Added "container.up" api
```
jedevc, line 2:
```suggestion
body: Added `Container.up` api
```

I'd also recommend adding a second line that explains that it's an alias to `.AsService().Up()`.

core/integration/module_up_test.go
```diff
@@ -13,17 +14,26 @@ import (
...
 func (ModuleSuite) TestDaggerUp(ctx context.Context, t *testctx.T) {
-	modDir := t.TempDir()
-	err := os.WriteFile(filepath.Join(modDir, "main.go"), []byte(`package main
+	// test both `container up` and `container as-service up`
+	for _, asObject := range []string{"container", "service"} {
+		defaultPort := "23400"
+
+		if asObject == "container" {
+			defaultPort = "23100"
+		}
```
jedevc, line 25: suggestions: use a testcase matrix pattern here so we don't need `if asObject == nil` throughout the tests. So essentially each testcase becomes:

```go
struct {
	name string
	port int
	cmd []string  // e.g. ["as-service", "up"]
}
```

rajatjindal: done. had to add some reusable function to make it work, but it looks nice now.

rajatjindal:
I have added commit from https://github.com/dagger/dagger/pull/8470/ to try out and check if there are any genuine failures in this branch.

jedevc (approved):

PR #8485, opened 2024-09-18
fix(python): use ModuleSource.digest for stable unique path

jedevc:
See https://github.com/dagger/dagger/pull/8245#discussion_r1733172781 for context.

The previous module ID hashing just really isn't very stable - we can hash the contents for *way* better results.

### DIFF
--- a/sdk/python/runtime/main.go
+++ b/sdk/python/runtime/main.go
@@ -4,9 +4,7 @@ package main
 
 import (
 	"context"
-	"crypto/sha256"
 	_ "embed"
-	"encoding/hex"
 	"fmt"
 	"path"
 	"python-sdk/internal/dagger"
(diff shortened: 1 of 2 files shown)


helderco (approved):
:shipit:

sdk/python/runtime/main.go
```diff
@@ -167,14 +165,10 @@ func (m *PythonSdk) Load(ctx context.Context, modSource *dagger.ModuleSource) (*
 		return nil, fmt.Errorf("runtime module load: %w", err)
 	}
 
-	// FIXME: ModuleSource.id is not stable
-	modID, err := m.Discovery.ModSource.ID(ctx)
+	modDigest, err := m.Discovery.ModSource.Digest(ctx)
```
helderco, line 168: What are the inputs to the ModuleSource digest? Does a change in source code create a different digest?

I think we're still not done here, just want to get it out since it's an improvement, but the only thing we need to worry about here is the building of the module itself. Contents don't matter (except for `pyproject.toml`), only file paths, for example. The `./sdk` dir should be excluded, etc...

I think uv now has the ability to specify cache keys for a Python package, but haven't had the time to look into it. Then the best solution would be to put these variables that matter in there and remove this digest in the file tree.

helderco: Context: https://docs.astral.sh/uv/concepts/cache/#dependency-caching

jedevc: > What are the inputs to the ModuleSource digest? Does a change in source code create a different digest?

For git, it's the git commit SHA - while this might not be the full content, it is *very* efficient to calculate (since we already have it).

For local, it's the content hash of the source directory.

I definitely think there's room for improvement, but I think it needs to potentially be calculated directly by the Python SDK then?

Issue #8401, opened 2024-09-10
🐞  core(SSH_AUTH_SOCK): encoding issue on socket path


sheldonhull:
### What is the issue?

The path for agent.socket is being evaluated as a URL, escaping spaces for example on darwin, which gave values like `%20` for space etc.

### Dagger version

v0.12.7

sheldonhull:
cc @grouville

grouville:
Adding some context: 
The way the engine requests the socketPath to load to the client is being serialized by `url.Parse`, which encodes the spaces into `%20`: https://github.com/dagger/dagger/blob/main/core/socket.go#L174-L189.

As part of https://github.com/dagger/dagger/pull/8365, it seems the relying on `url.PathEscape()` is mandatory to handle properly such cases. It seems that this issue has been here since implementation of `core.Socket` type. Quite surprised that this has not been found out earlier :pray:

grouville:
@sheldonhull could you please test the new version v0.13.1, and see if you still have the issue, without using the symlink workaround ?

sheldonhull:
Ok, I upgraded and ran again and don't have to set the SSH_AUTH_SOCK override any further.
If I run into the encoding issue or related I'll be sure to let you know with an update. Thanks for the follow-up and your time!

PR #8493, opened 2024-09-19
fix(engine): properly encode / decode socket path

grouville:
Current engine / client communication relies on the url.URL type to encode / decode different kinds of socket path:
- unix
- tcp/udp

However, this leads to some encoding issues for the unix path: spaces are converted to %20, special chars are escaped.

This PR fixes it and adds a corresponding test

### DIFF
--- a/engine/client/socket.go
+++ b/engine/client/socket.go
@@ -43,7 +43,15 @@ func (p SocketProvider) CheckAgent(ctx context.Context, req *sshforward.CheckAge
 	}
 	switch u.Scheme {
 	case "unix":
-		path := u.Path
+		var path string
+		if u.RawPath != "" {
+			path, err = url.PathUnescape(u.RawPath)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "Error unescaping path: %s", err)
+			}
+		} else {
+			path = u.Path
+		}
 		if p.UnixPathMapper != nil {
 			path, err = p.UnixPathMapper(path)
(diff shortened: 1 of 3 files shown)


engine/client/socket.go
```diff
@@ -43,7 +43,15 @@ func (p SocketProvider) CheckAgent(ctx context.Context, req *sshforward.CheckAge
 	}
 	switch u.Scheme {
 	case "unix":
-		path := u.Path
+		var path string
+		if u.RawPath != "" {
+			path, err = url.PathUnescape(u.RawPath)
```
jedevc, line 48: I don't get understand this :thinking: 

Why is this behavior different from getting `u.Path`? Shouldn't it be the same? From `url.go`'s `setPath`, called during `url.Parse`:

```
func (u *URL) setPath(p string) error {
	path, err := unescape(p, encodePath)
	if err != nil {
		return err
	}
	u.Path = path
```

grouville: Mmmh, you are right, sorry 🙏 I extended my tests and I realized that I was basing my implementation on some issue a user had, that led to some encoding problems: https://github.com/dagger/dagger/issues/8401.

However, I can't repro, and following this comment, you're basically right.

As we moved from relative path to absolute path, this does not seem necessary anymore: just path traversal are trimmed in current form "../..". It was problematic in the relative path implementation but the Clean is correct in this case

grouville:
closing as not as relevant

Issue #4368, opened 2023-01-12
DenoJS SDK


shykes:
## Overview

Dagger has a [NodeJS SDK](https://docs.dagger.io/sdk/nodejs). What if it also had a [DenoJS](https://deno.land) SDK?

Deno is a solid alternative to NodeJS, by the creator of NodeJS, and has great traction. It could help gain additional exposure with the kind of forward-thinking developers that tend to enjoy Dagger.

slumbering:
That's a great idea !! 

However NodeJS is still a huge ecosystem that will live for many years. 
Deno has many advantage and has great traction indeed but the community is still pretty small compare to NodeJS.
I'm just wondering how much effort do we want to put on create a "new" SDK. I'm not familiar with Deno usage, so maybe a migration from Node to Deno could be not that painfull (but I'm also always sceptical about painless migration...).

helderco:
Isn't Deno just another JavaScript/Typescript runtime? It should be compatible with the current SDK, at least in code. Probably no need to create a new SDK.

The "NodeJS" name is unfortunate. It supports both Typescript and JavaScript so we avoided both names and chose NodeJS. Now there's other runtimes like Deno so again it feels restrictive.

The **common denominator** here is JavaScript. We should probably rename the NodeJS SDK to just the JS SDK. Even though we favor Typescript, it's a superset of JavaScript and it's all a part of the JS ecosystem anyway.

Even if we need to split, I'd rather have a JS SDK and a TS SDK (language based instead of runtime based), where the JS SDK could be auto generated entirely from the TS one.

TomChv:
> Dagger has a [NodeJS SDK](https://docs.dagger.io/sdk/nodejs). What if it also had a [DenoJS](https://deno.land/) SDK?

I love Deno and I think it fit well to Dagger to have a Deno SDK. Especially since its usage is closer to Go (we can create binary from Deno code etc). The community is still small, but building tools in Deno can only help too!  
However it's true that Typescript code shall works on Deno, the only difference may be dependencies installed.

I think we can [use Node package](https://deno.land/manual@v1.12.2/npm_nodejs) in Deno, but I think using Deno modules would be better.

There will not much difference between those SDK though, the DX and code will be almost the same except for external library call.

I think it does not require much work, compared to creating a RustSDK for instance but it will still need time.

As a Deno user, I'm down to work on it :D

TomChv:
[...]

PR #8497, opened 2024-09-19
feat: deno runtime

TomChv:
:warning: Based on #8494!

This PR is the bases work for Deno runtime integration so anyone can add its contribution.

## TODO:

- [ ] Create a `deno.json` to map all deps import
- [ ] Add `yarn test:deno` command and making sure it works
- [ ] Add Deno to the test list in `ci/sdk/typescript`
- [ ] Update the runtime to use Deno if `deno.json` is detected
- [ ] See how it reacts to deps install (Do we actually needs `node_modules` to run `deno run`?)
- [ ] Add some tests in `TestTypeScript suite`

### DIFF
--- a/sdk/typescript/.mocharc.json
+++ b/sdk/typescript/.mocharc.json
@@ -1,5 +1,4 @@
 {
   "extension": ["ts"],
-  "spec": "**/*.spec.ts",
-  "loader": "ts-node/esm"
+  "spec": "**/*.spec.ts"
 }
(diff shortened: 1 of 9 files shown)


TomChv:
Pushed some updates, I got weird issue, I left couple of info there, I'm still looking for a solution to make these tests running.

https://github.com/dagger/dagger/issues/4368#issuecomment-2368054398

PR #8500, opened 2024-09-19
re-bump midterm to fix double linebreaks

vito:
Dependabot erroneously bumped this to a _lower_ version - what a jerk

This sort of breakage will be prevented in the future via the test suite introduced in #8442


jedevc (approved):
:pray: :pray: :pray:

PR #8512, opened 2024-09-20
fix: remove required path in ts sdk

TomChv:
Follow-up to https://github.com/dagger/dagger/pull/7864#discussion_r1712142479


helderco:
For context, these patterns were supposed to be removed but were instead changed because of a misunderstanding. They're not relative to the runtime module, they're relative to the target module's context directory.

I don't expect any user to be affected by this as they'd need a `sdk/typescript` subdir in their own repos, with those files, and for a module to interact with them in some way.

helderco (approved):
Just waiting for checks to pass before merge.

PR #8522, opened 2024-09-20
allow '=' to also increase verbosity

joeshaw:
On a US keyboard, `-` does not require shift to be pressed, but `+` does (on the `=` key).  This makes it annoying to adjust the verbosity quickly.  This adds `=` as an alias for `+` so either can be used to increase the verbosity, and you can avoid using the shift key altogether.

This has been a paper cut for me as I try to turn up the verbosity to learn what dagger is doing and then trying to dial it back down to the default.


jedevc (approved):

vito (approved):
Had a bit of hesitation here just being conscious of our keymap economy, but I think it's fine:

* Vim doesn't have a normal mode binding for <kbd>=</kbd>, so no potential conflict there
* Browsers let you zoom in with <kbd>Ctrl</kbd>+<kbd>=</kbd>, so there's prior art

so, lgtm - thanks!

PR #8525, opened 2024-09-21
Fix Python telemetry

helderco:
Custom spans haven't been working in Python since the move to the http OTLP exporter:
- https://github.com/dagger/dagger/pull/7315

Simplest solution is to replace the package dependency from the grpc exporter to http, but apparently these env vars are no longer [set by the engine](https://github.com/dagger/dagger/issues/8526) too:
- `OTEL_LOGS_EXPORTER=otlp`
- `OTEL_METRICS_EXPORTER=otlp`
- `OTEL_TRACES_EXPORTER=otlp`

\cc @TomChv for TypeScript


sdk/python/dev/pyproject.toml
```diff
@@ -2,7 +2,11 @@
 name = "main"
 version = "0.0.0"
 requires-python = ">=3.12"
-dependencies = ["dagger-io"]
+dependencies = [
+    "dagger-io",
+    # Temporary, until in SDK's next release.
+    "opentelemetry-exporter-otlp-proto-http>=1.27.0",
```
helderco, line 8: We need to keep our modules compatible with both release and dev versions for CI. 😮‍💨

sdk/python/pyproject.toml
```diff
@@ -34,7 +34,7 @@ dependencies = [
     "typing_extensions>=4.8.0",
     "rich>=10.11.0",
     "opentelemetry-sdk>=1.23.0",
-    "opentelemetry-exporter-otlp-proto-grpc>=1.23.0",
+    "opentelemetry-exporter-otlp-proto-http>=1.27.0",
```
helderco, line 37: This should have been enough to fix the issue.

sdk/python/src/dagger/mod/_module.py
```diff
@@ -197,6 +197,7 @@ def get_resolver(
     def __call__(self) -> None:
         if self._log_level is not None:
             configure_logging(self._log_level)
+        telemetry.initialize()
```
helderco, line 200: Just ensuring that telemetry is initialized early instead of when using a custom span. There used to be custom root spans in this file, which took care of this, but those were removed in https://github.com/dagger/dagger/pull/7315.

sdk/python/src/dagger/telemetry.py
```diff
@@ -64,19 +66,9 @@ def otel_configured() -> bool:
...
 
-def otel_enabled(name: str = "") -> bool:
-    """Checks if a specific OpenTelemetry instrumentation is disabled.
-
-    Based on environment variables:
-    - If name is not provided, checks for global OTEL_SDK_DISABLED=true.
-    - Otherwise, checks for OTEL_PYTHON_{name}_INSTRUMENTATION_DISABLED=true.
-    """
-    name = (
-        f"OTEL_PYTHON_{name.upper()}_INSTRUMENTATION_DISABLED"
```
helderco, line 75: There's a `OTEL_PYTHON_LOGGING_INSTRUMENTATION_DISABLED` in the `opentelemetry.sdk` package because logs support is experimental, but this looks for:
```env
OTEL_PYTHON_TRACES_INSTRUMENTATION_DISABLED=true
OTEL_PYTHON_LOGS_INSTRUMENTATION_DISABLED=true
OTEL_PYTHON_METRICS_INSTRUMENTATION_DISABLED=true
```
The `_LOGGING_` (instead of `_LOGS_`) variation is still picked up by the `opentelemetry.sdk` so it's not a problem that they don't match. The idea was to allow the user to disable only one kind of instrumentation rather than the whole thing but I've wanted to simplify this for a while now.

sdk/python/src/dagger/telemetry.py
```diff
@@ -146,8 +138,15 @@ def _prepare_env(self):
...
-        # TODO: The INSECURE env vars should be set by the shim next to ENDPOINT
-        # rather than in the SDK.
+        # TODO: The following env vars should be set by the shim rather than in the SDK.
+
+        for exporter in (
+            OTEL_TRACES_EXPORTER,
+            OTEL_LOGS_EXPORTER,
+            OTEL_METRICS_EXPORTER,
+        ):
+            os.environ.setdefault(exporter, "otlp")
```
helderco, line 148: These env vars used to be added by the engine. Took me a beat to debug and understand what was wrong.

sdk/python/src/dagger/telemetry/attributes.py
```diff
@@ -1,4 +0,0 @@
-from typing import Final
-
-UI_ENCAPSULATE: Final = "dagger.io/ui.encapsulate"
```
helderco, line 3: This used to be used by the two root spans in the SDK, at the beginning of the _registration_ and _invocation_ phases for a function call, but those were removed, so just removing this before someone depends on it.

helderco (commented):
Leaving some notes for reviewer.

vito (commented):
**EDIT: nevermind** - I needed to `dagger develop`. I'm surprised by that since normally we re-codegen before running module anyway, but I guess dependency changes are required externally before running?

Trying this out locally and adding a test to the new telemetry test suite (#8442), but I get this error:

```
RuntimeError: Requested component 'otlp_proto_http' not found in entry point                                                                                                                                                                                                                 
'opentelemetry_traces_exporter'                                                                                                                                                                                                                                                              
```

With this module:

```python
from opentelemetry import trace
import datetime

from dagger import dag, function, object_type

tracer = trace.get_tracer(__name__)


@object_type
class Python:
    @function
    async def echo(self, msg: str) -> str:
        return await (
            dag.container()
            .from_("alpine:latest")
            .with_exec(["echo", msg])
            .stdout()
        )

    @function
    async def custom_span(self) -> str:
        with tracer.start_as_current_span("custom span"):
            return await self.echo("hey dude!")

    @function
    async def pending(self):
        return await (
            dag.container()
            .from_("alpine:latest")
            .with_env_variable("NOW", str(datetime.datetime.now()))
            .with_exec(["sleep", "1"])
            .with_exec(["false"])
            .with_exec(["sleep", "1"])
            .sync()
        )
```

vito (approved):
Approving - was able to verify the fix locally, and once this is merged I can add test coverage for it in #8442 🎉

TomChv (approved):

TomChv:
I'm currently working on the TS SDK to also use the HTTP client instead of GRPC. will open a PR soon

PR #8528, opened 2024-09-21
fix readthedocs command

helderco:
The `sdk/python/.readthedocs.yaml` isn't being used yet. It's meant to replace the [one at the root](https://github.com/dagger/dagger/blob/5ec03a604d50b721aaa65ba7c4db8c875696ac72/.readthedocs.yaml), but needs this fix before switching.


helderco:
This file isn't used anywhere right now, and when it will be it's going to be in an external system, so any checks failing are flukes.

TomChv (approved):

TomChv:
LGTM, let's merge it

Issue #8524, opened 2024-09-20
🐞 Typescript SDK - Intermittent Issues "Encountered an unknown error while requesting data via graphql"


levlaz:
### What is the issue?

When running a build with the Tyepscript SDK this error causes the entire build to fail. 

We have seen this happen in several different Typescript projects since 0.13 came out. 

### Dagger version

dagger v0.13.2 (registry.dagger.io/engine:v0.13.2) darwin/arm64

### Steps to reproduce

I run into this intermittently on this project: https://github.com/levlaz/medplum on the daggerize branch 

`dagger call build-matrix entries`

I am not able to reproduce this consistently 

### Log output

Error: Encountered an unknown error while requesting data via graphql

jedevc:
Okay, so it *does* look like the pipeline is failing - the `npm ci` step fails: https://dagger.cloud/jedevc/traces/714d4614707c58ab16f8088da5ad1b48?span=cb510c70337818de#cb303a9ea768c3d2

![image](https://github.com/user-attachments/assets/4efa6f3b-520b-4af7-8315-3881e0160285)

However, what I don't understand is why this error message throws "Encountered an unknown error while requesting data via graphql" - this should be an `EXEC_ERROR` from the condition in https://github.com/dagger/dagger/blob/5ec03a604d50b721aaa65ba7c4db8c875696ac72/sdk/typescript/api/utils.ts#L206-L239

cc @TomChv - could the graphql library changes we made recently be causing this?

jedevc:
I modified the exception to the following:

```diff
    // Just throw the unknown error
+   throw e;
    throw new UnknownDaggerError(
      "Encountered an unknown error while requesting data via graphql",
      {
        cause: e,
      },
    )
```

And get this error:

```
Error: response from query: input: medplum.buildMatrix resolve: call function "buildMatrix": process "tsx --no-deprecation --tsconfig /src/.dagger/tsconfig.json /src/.dagger/src/__dagger.entrypoint.ts" did not complete successfully: exit code: 1

Stderr:
Error: fetch failed
```

`fetch failed` doesn't seem like the right error message to be presenting here, not quite sure why we're not getting a graphql error?

TomChv:
Because we catch the traces in the execution to not display it because it was too verbose for the user. But I think I'll rollback on that decision since it creates confusion.

`npm ci` isn't supposed to fail, and isn't related to that `graphql-request` error.
[...]

PR #8544, opened 2024-09-23
feat: ts-sdk now print error cause in error case

TomChv:
To have a better understanding of https://github.com/dagger/dagger/issues/8524 this PR display the `cause` field from the error object.

Test function:

```typescript
  @func()
  error(): string {
    throw new Error("foo", { cause: new Error("bar") })

    return 'a'
  }
```

Before

```
$ dagger call error

Stderr:
Error: foo
```

Now: 

```
$ dagger call error

Stderr:
Error: foo, Error: bar
```

### DIFF
--- a/sdk/typescript/entrypoint/entrypoint.ts
+++ b/sdk/typescript/entrypoint/entrypoint.ts
@@ -59,7 +59,7 @@ export async function entrypoint() {
           })
         } catch (e) {
           if (e instanceof Error) {
-            console.error(`Error: ${e.message}`)
+            console.error(`Error: ${e.message}, ${e.cause}`)
           } else {
             console.error(e)
           }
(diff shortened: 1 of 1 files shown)


jedevc (approved):

sdk/typescript/entrypoint/entrypoint.ts
```diff
@@ -59,7 +59,7 @@ export async function entrypoint() {
           })
         } catch (e) {
           if (e instanceof Error) {
-            console.error(`Error: ${e.message}`)
+            console.error(`Error: ${e.message}, ${e.cause}`)
```
jedevc, line 62: Is it ever possible for `e.cause` to be unset?

helderco (approved):

jedevc (approved):

PR #8549, opened 2024-09-23
feat: extend typescript client timeout

TomChv:


sdk/typescript/graphql/client.ts
```diff
@@ -14,6 +35,7 @@ class CustomSetter {
 
 export function createGQLClient(port: number, token: string): GraphQLClient {
   const client = new GraphQLClient(`http://127.0.0.1:${port}/query`, {
+    fetch: createFetchWithTimeout(1000 * 60 * 30), // 30minutes timeout
```
sipsma, line 38: There's realistically no expected timeout anywhere; if the user is performing an operation that takes several hours we don't want to timeout.

If we are forced to set this to some value or else it will default to a low value, then I'd just set it as high as it can be (or like 1 week, something very very long).

cc @helderco @jedevc @vito in case I'm forgetting something where we have an expected operation timeout in our stack somewhere

TomChv: I extended it to a week, because it's quite obscur what's the timeout config in the `graphql-request` client.

TomChv: It seems node-fetch has a 5minute timeout by default now (based on https://github.com/nodejs/undici/issues/1373) which is not great for our case.

sdk/typescript/graphql/client.ts
```diff
@@ -1,6 +1,27 @@
 import * as opentelemetry from "@opentelemetry/api"
 import { GraphQLClient } from "graphql-request"
 
+const createFetchWithTimeout =
+  (timeout: number) => async (input: RequestInfo | URL, init?: RequestInit) => {
+    if (init?.signal) {
+      throw new Error(
+        "it looks like graphql-request started using AbortSignal on its own. Please check graphql-request's recent updates",
```
jedevc, line 8: :eyes: This is a *very* strange error to send to the user - this is an internal error, we should mark it as such, I don't think the user can do anything about it.

The suggestion for what to do can be a code comment IMO.

sdk/typescript/graphql/client.ts
```diff
@@ -14,6 +35,7 @@ class CustomSetter {
 
 export function createGQLClient(port: number, token: string): GraphQLClient {
   const client = new GraphQLClient(`http://127.0.0.1:${port}/query`, {
+    fetch: createFetchWithTimeout(1000 * 60 * 60 * 24 * 7), // 1 week timeout so we should never hit that one.
```
jedevc, line 38: nit: probably worth commenting the logic for why we have such a high timeout.

sdk/typescript/graphql/client.ts
jpadams: @TomChv does this change override that fetch in undici timeout, or do you need to re-import it and override it, etc?

TomChv: Yeah this override it

sdk/typescript/graphql/client.ts
TomChv: Done

sdk/typescript/graphql/client.ts
jedevc: It's good as is, but the error message that was being thrown previously would have been better suited as a code comment imo.

jedevc (approved):

PR #8551, opened 2024-09-23
engine: optimize parseRefString by checking for local paths first

### LINKED ISSUE #8398: 🐞 `dagger functions` can be up to `12.5x` slower on certain hosts
### What is the issue?

This is a follow-up to:
- https://github.com/dagger/dagger/issues/8395

Given the following host:

<details><summary><code>noefetch</code></summary><pre>
             /////////////                gerhard@h22
         /////////////////////            -----------
      ///////*767////////////////         OS: Pop!_OS 22.04 LTS x86_64
    //////7676767676*//////////////       Kernel: 6.9.3-76060903-generic
   /////76767//7676767//////////////      Uptime: 31 days, 7 hours, 31 mins
  /////767676///*76767///////////////     Packages: 2346 (dpkg), 13 (brew), 26 (flatpak)
 ///////767676///76767.///7676*///////    Shell: bash 5.1.16
/////////767676//76767///767676////////   Resolution: 1024x768
//////////76767676767////76767/////////   Terminal: /dev/pts/2
///////////76767676//////7676//////////   CPU: AMD Ryzen 7 5800X (16) @ 3.800GHz
////////////,7676,///////767///////////   GPU: NVIDIA 3f:00.0 NVIDIA Corporation Device 2704
/////////////*7676///////76////////////   Memory: 15210MiB / 128719MiB
///////////////7676////////////////////
 ///////////////7676///767////////////
  //////////////////////'////////////
   //////.7676767676767676767,//////
    /////767676767676767676767/////
      ///////////////////////////
         /////////////////////
             /////////////
</pre></details>

<details><summary><code>docker info</code></summary><pre>
Client: Docker Engine - Community
 Version:    27.1.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose
Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 11
 Server Version: 27.1.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2bf793ef6dc9a18e00cb12efb64355c2c9d5eb41
 runc version: v1.1.13-0-g58aa920
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.9.3-76060903-generic
 Operating System: Pop!_OS 22.04 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 125.7GiB
 Name: h22
 ID: 1c93b978-5459-4719-a1aa-92422d716091
 Docker Root Dir: /data-2024-04-10/docker
 Debug Mode: false
 Username: infradagger
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
</pre></details>

`dagger init --sdk=go --name=foo` takes `17s` in an empty directory https://dagger.cloud/dagger/traces/16339e06ad4f1364f6c5ae164a7cb18b

`dagger functions` takes `4s` the first time https://dagger.cloud/dagger/traces/621b9e3901b585cd260f62a9c79c949e

`dagger functions` takes `1.5s` the second time https://dagger.cloud/dagger/traces/83f2d9a9459f9b62d22397ed384bd031

So far, so good.

Same host, running the same command **inside** a subdir of a fresh clone of the `dagger/dagger` repository:

```console
cd $(mktemp -d)
git clone https://github.com/dagger/dagger
cd dagger
mkdir foo
cd foo
```

`dagger init --sdk=go --name=foo` takes `25s` in this `dagger/dagger/foo` directory https://dagger.cloud/dagger/traces/65ff21e193c7b7681d862aaf6253b047

`dagger functions` takes `25s` the first time https://dagger.cloud/dagger/traces/dbf4420a92e4874e38fd8fd67b73c96f

`dagger functions` takes `22s` the second time https://dagger.cloud/dagger/traces/7f2f5efad11460ee07d0728a2cbf460c

This is completely unexpected. All that time is spent in [`moduleSource(refString:"./foo", relHostPath:".")`](https://dagger.cloud/dagger/traces/7f2f5efad11460ee07d0728a2cbf460c#99926847be9c0c78)
image

What's even more peculiar is that this behaviour is **inconsistent** between different hosts.

Given the following host:

<details><summary><code>noefetch</code></summary><pre>
            .-/+oossssoo+/-.               gerhard@h24
        `:+ssssssssssssssssss+:`           -----------
      -+ssssssssssssssssssyyssss+-         OS: Ubuntu 24.04 LTS x86_64
    .ossssssssssssssssssdMMMNysssso.       Host: Intel Z790 Ddr5
   /ssssssssssshdmmNNmmyNMMMMhssssss/      Kernel: 6.8.0-40-generic
  +ssssssssshmydMMMMMMMNddddyssssssss+     Uptime: 7 hours, 7 mins
 /sssssssshNMMMyhhyyyyhmNMMMNhssssssss/    Packages: 1724 (dpkg), 18 (brew), 13 (snap)
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Shell: bash 5.2.21
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   Resolution: 2560x1440
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   Terminal: /dev/pts/0
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   CPU: 13th Gen Intel i9-13900KS (32) @ 3.200GHz
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   GPU: NVIDIA GeForce RTX 4090
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Memory: 2250MiB / 192932MiB
 /sssssssshNMMMyhhyyyyhdNMMMNhssssssss/
  +sssssssssdmydMMMMMMMMddddyssssssss+
   /ssssssssssshdmNNNNmyNMMMMhssssss/
    .ossssssssssssssssssdMMMNysssso.
      -+sssssssssssssssssyyyssss+-
        `:+ssssssssssssssssss+:`
            .-/+oossssoo+/-.
</pre></details>

<details><summary><code>docker info</code></summary><pre>
Client: Docker Engine - Community
 Version:    27.1.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 4
 Server Version: 27.1.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 8fc6bcff51318944179630522a095cc9dbf9f353
 runc version: v1.1.13-0-g58aa920
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-40-generic
 Operating System: Ubuntu 24.04 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 32
 Total Memory: 188.4GiB
 Name: h24
 ID: 2637752e-9a21-47b5-adb0-844c1c544e90
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
</pre></details>

```console
cd $(mktemp -d)
git clone https://github.com/dagger/dagger
cd dagger
mkdir foo
cd foo
```

`dagger init --sdk=go --name=foo` takes `15s` in this `dagger/dagger/foo` directory https://dagger.cloud/dagger/traces/e2d70104b8fb18faddc976d9d18229bc

`dagger functions` takes `4s` the first time https://dagger.cloud/dagger/traces/9a53db92c4f4e86ad633951b450ae64e

`dagger functions` takes `2s` the second time https://dagger.cloud/dagger/traces/29e336eb13ae7db3c55f9a6a9a144709

How can this second host be `12x` slower than the first host? There is no disk or CPU contention. Both hosts are running on the same network, with access to the same WAN. Here is a side-by-side comparison (left is the slow host, right is the fast one):

https://github.com/user-attachments/assets/0dfbf8af-afe9-43fe-aa97-d47bbf5f5dd3

### Dagger version

dagger v0.12.7 (registry.dagger.io/engine:v0.12.7) linux/amd64

### Steps to reproduce

```console
cd $(mktemp -d)
git clone https://github.com/dagger/dagger
cd dagger
mkdir foo
cd foo

# the following command takes 25s to run:
dagger init --sdk=go --name=foo

# the following command takes 25s on the first run:
dagger functions

# the following command takes 22s on all subsequent runs, even if nothing changes
dagger functions
```

### Log output

_No response_

marcosnils:
Given how parseRefString works, it could happen that the ref parsed
yields to a `StatCallerHostPath` failure and also be a local ref as
well. This commit shorts circuits the function by exiting early and
avoiding an extra client RPC call

fixes #8398

### DIFF
--- a/core/schema/modulesource.go
+++ b/core/schema/modulesource.go
@@ -9,6 +9,7 @@ import (
 	"sort"
 	"strings"
 
+	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -17,6 +18,7 @@ import (
 	"github.com/dagger/dagger/core/modules"
 	"github.com/dagger/dagger/dagql"
 	"github.com/dagger/dagger/engine/client"
+	"github.com/dagger/dagger/engine/slog"
 	"github.com/dagger/dagger/engine/vcs"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/tonistiigi/fsutil/types"
@@ -213,16 +215,27 @@ type buildkitClient interface {
 // - if not, try to isolate root of git repo from the ref
 // - if nothing worked, fallback as local ref, as before
 func parseRefString(ctx context.Context, bk buildkitClient, refString string) parsedRefString {
+	ctx, span := core.Tracer(ctx).Start(ctx, fmt.Sprintf("parseRefString: %s", refString), trace.WithSpanKind(trace.SpanKindInternal))
+	defer span.End()
 	localParsed := parsedRefString{
 		modPath: refString,
 		kind:    core.ModuleSourceKindLocal,
 		scheme:  core.NoScheme,
 	}
 
+	// if the refString is a relative path, we can short-circuit here as we
+	// don't really about the `stat` return as the refString will always
+	// be local in this case.
+	if strings.HasPrefix(refString, ".") {
+		return localParsed
+	}
+
 	// First, we stat ref in case the mod path github.com/username is a local directory
 	stat, err := bk.StatCallerHostPath(ctx, refString, false)
 	if err == nil && stat.IsDir() {
 		return localParsed
+	} else if err != nil {
+		slog.Debug("parseRefString stat error", "error", err)
 	}
 
 	// Parse scheme and attempt to parse as git endpoint


grouville (approved):
✅  LGTM

grouville:
Runners seem to hang. I'd like to have all of these green prior merging 🙏

sipsma (approved):
The change itself LGTM but the implication of this fixing #8398 is that `bk.StatCallerHostPath` takes 20+ seconds *only on certain hosts*, correct @marcosnils?

Not a blocker on merging this, just want to confirm that because that would indicate something deeper wrong happening. If so, worth separately digging into that because fixing whatever deeper issue is happening might reap benefits elsewhere.

grouville:
> The change itself LGTM but the implication of this fixing https://github.com/dagger/dagger/issues/8398 is that bk.StatCallerHostPath takes 20+ seconds only on certain hosts, correct @marcosnils?

Not really, it's more that sometimes, given some particular conditions explained below, the bk.Stat returns false, leading us to try resolving the ref as a git module (which takes more time + timeouts).
As parseRefString always fallbacks as a local ModuleSource when we can't resolve the git dir, it is a no-op, just longer

Cause: depending on the context in which we run the dagger command (git repo or not), and where the CLI command is being run, the ref is being resolved differently: in the case of #8398, in the slow example, the moduleSource ref is `./foo` (as the module is at this given path relative to the root of the git dir)

But, the CLI is in the subdirectory, making the bk.StatDir appending the ref to the current workdir: which leads to cwd+./foo, which does not exist in this context, leading to use trying to resolve it as a git module, then timeout, then fallback as local module

core/schema/modulesource.go
```diff
@@ -213,16 +214,25 @@ type buildkitClient interface {
 // - if not, try to isolate root of git repo from the ref
 // - if nothing worked, fallback as local ref, as before
 func parseRefString(ctx context.Context, bk buildkitClient, refString string) parsedRefString {
+	ctx, span := core.Tracer(ctx).Start(ctx, fmt.Sprintf("parseRefString: %s", refString))
+
```
jedevc, line 218: nit: empty line

core/schema/modulesource.go
```diff
@@ -213,16 +214,25 @@ type buildkitClient interface {
...
+	if strings.HasPrefix(refString, ".") {
+		return localParsed
+	}
+
 	// First, we stat ref in case the mod path github.com/username is a local directory
 	stat, err := bk.StatCallerHostPath(ctx, refString, false)
 	if err == nil && stat.IsDir() {
 		return localParsed
+	} else if err != nil {
+		bklog.GetLogger(ctx).Errorf("parseRefString stat:%v err:%v", stat, err)
```
jedevc, line 235: I think we want to use `slog` here?

Also, I think since this is "expected", this should be lowered to a `Debug` level log.

core/schema/modulesource.go
```diff
@@ -213,16 +214,25 @@ type buildkitClient interface {
...
+	defer span.End()
 	localParsed := parsedRefString{
 		modPath: refString,
 		kind:    core.ModuleSourceKindLocal,
 		scheme:  core.NoScheme,
 	}
 
+	if strings.HasPrefix(refString, ".") {
+		return localParsed
+	}
```
jedevc, line 228: suggestion: *might* be worth a comment on explaining why this optimization works.

jedevc (commented):
> However, if the CLI is run from a subdirectory, bk.StatDir appends the reference to the current working directory. This results in checking for "cwd+./foo", which doesn't exist.

Something seems off there. I think we should be able to handle this case, and correctly navigate up to the top-level?

marcosnils:
> Something seems off there. I think we should be able to handle this case, and correctly navigate up to the top-level?
> 
> This fix seems good to me as a hotfix, but I think the underlying cause is that we're looking for the wrong path?

 Yes, but we don't really care if the path exists or not in this case because the function will ultimately end up fallbacking to `localParsed` regardless.

jedevc:
> > Something seems off there. I think we should be able to handle this case, and correctly navigate up to the top-level?
> > This fix seems good to me as a hotfix, but I think the underlying cause is that we're looking for the wrong path?
> 
> Yes, but we don't really care if the path exists or not in this case because the function will ultimately end up fallbacking to `localParsed` regardless. LMK if the comment I've added in the code makes sense @jedevc 🙏

Mm, yes, this makes sense, and this isn't a blocker!

I think though, the path we're looking in doesn't make sense, something seems wrong there.

---

Potential follow-up as well - should we try and have all module refs with relative paths canonicalize to have a `./` or `../`? Currently it looks like we don't: https://github.com/dagger/dagger/blob/7cba95531a426dbf9dc6f24b41fc6357e8b3f9b7/dagger.json#L51

core/schema/modulesource.go
```diff
@@ -213,16 +214,27 @@ type buildkitClient interface {
 // - if not, try to isolate root of git repo from the ref
 // - if nothing worked, fallback as local ref, as before
 func parseRefString(ctx context.Context, bk buildkitClient, refString string) parsedRefString {
+	ctx, span := core.Tracer(ctx).Start(ctx, fmt.Sprintf("parseRefString: %s", refString))
```
jedevc, line 217: *might* be worth annotating this as `internal`? not-a-blocker

jedevc (approved):

PR #8555, opened 2024-09-24
chore: limit file upload size in generated CI jobs

shykes:
This fixes an issue encountered by @jedevc , where the Java SDK logs break the "markdown summary" feature of the generated CI configs, because they are too large and hit the 1024k markdown file upload limit. This limits to the last 1000 last lines of logs in the markdown.


.github/go.mod
```diff
@@ -17,26 +17,26 @@ require (
...
-	google.golang.org/genproto/googleapis/api v0.0.0-20240520151616-dc85e6b867a5 // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20240515191416-fc5f0ca64291 // indirect
-	google.golang.org/protobuf v1.34.1 // indirect
+	golang.org/x/net v0.29.0 // indirect
+	golang.org/x/sys v0.25.0 // indirect
+	golang.org/x/text v0.18.0 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20240903143218-8af14fe29dc1 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20240903143218-8af14fe29dc1 // indirect
+	google.golang.org/protobuf v1.34.2 // indirect
 )
```
shykes, line 40: I don't know why this gets automatically added by `dagger develop`, but 🤷

jedevc: Yeah, this is correct. We update the dependencies in `go.mod` to a minimum version, since otherwise we end up targetting slightly wrong versions (which has led to some very very difficult bugs previously).

PR #8557, opened 2024-09-24
use '.dagger' as source dir during 'init' and 'develop' when current dir is not empty

### LINKED ISSUE #8123: Daggerizing without littering
## Problem

By default, daggerizing an existing project (either with `dagger init --sdk=NAME`, or `dagger develop --sdk=NAME`) will cause littering, by mixing Dagger-managed files with pre-existing files. This can cause frustration and confusion just as a user is familiarizing themselves with Dagger.

Experienced Daggernauts can avoid this littering by explicitly setting `--source` to a subdirectory. But 1) this is cumbersome 2) it prevents a shared convention on default source directory 3) it doesn't help beginners.

## Solution

Change how the default source directory is selected, so that littering never happens by default.

| | Before | After |
| -- | -- | -- |
| Source directory, if module root is empty | `.` ✅  | `.` ✅ |
| Source directory, if module root is **not** empty | `.` ❌ | `.dagger`✅ |

### Custom boilerplate

@helderco asks:

> What if I want to customize my language-specific boilerplate, for example with my own Python config files? This requires creating files before `dagger init`, which would trigger the `.dagger` behavior, but I don't want that.

In this case, you actually want to override the anti-littering protection: you're OK with the Dagger SDK mixing files into your directory, because you know what you're doing. It's reasonable to express that with an explicit `--source .`.

### Weird to have code in `.dagger`

One concern with `.dagger` is that it feels weird to have source code in a dot directory, and may possibly confuse some IDEs.

I acknowledge the concern, but I think it's the best tradeoff:

- In spite of thorough bikeshedding, we have not found an alternative that got stronger consensus
- IDEs don't seem to be actually confused. Github actions using `.github/workflows` works to our advantage
- Using a dot directory as "annotation" is a familiar pattern. Again, `.github` works in our favor.
- The "weird" feeling is proportional to the size of the module source code. If it grows too large, we actually want you to spin it out into standalone modules. So the gradual "weirdness" factor works in our favor, because it acts a nudge towards desired best practices.

rajatjindal:
fixes #8123

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -55,9 +55,36 @@ var (
 
 const (
 	moduleURLDefault = "."
-
-	defaultModuleSourceDirName = "."
 )
+
+// if the source root path already has some files
+// then use `srcRootPath/.dagger` for source
+func inferSourcePathDir(srcRootPath string) (string, error) {
+	list, err := os.ReadDir(srcRootPath)
+	if err != nil {
+		return "", err
+	}
+
+	for _, l := range list {
+		if l.Name() == "dagger.json" {
+			continue
+		}
+
+		// .dagger already exist, return that
+		if l.Name() == ".dagger" {
+			return ".dagger", nil
+		}
+
+		// ignore hidden files
+		if strings.HasPrefix(l.Name(), ".") {
+			continue
+		}
+
+		return ".dagger", nil
+	}
+
+	return ".", nil
+}
 
 func init() {
 	moduleFlags.StringVarP(&moduleURL, "mod", "m", "", "Path to the module directory. Either local path or a remote git repo")
@@ -149,17 +176,29 @@ If --sdk is specified, the given SDK is installed in the module. You can do this
 
 			// only bother setting source path if there's an sdk at this time
 			if sdk != "" {
+				// if user didn't specified moduleSourcePath explicitly,
+				// check if current dir is non-empty and infer the source
+				// path accordingly.
 				if moduleSourcePath == "" {
-					moduleSourcePath = filepath.Join(modConf.LocalRootSourcePath, defaultModuleSourceDirName)
+					inferredSourcePath, err := inferSourcePathDir(modConf.LocalRootSourcePath)
+					if err != nil {
+						return err
+					}
+
+					moduleSourcePath = filepath.Join(modConf.LocalRootSourcePath, inferredSourcePath)
 				}
-				// ensure source path is relative to the source root
-				sourceAbsPath, err := filepath.Abs(moduleSourcePath)
-				if err != nil {
-					return fmt.Errorf("failed to get absolute source path for %s: %w", moduleSourcePath, err)
-				}
-				moduleSourcePath, err = filepath.Rel(modConf.LocalRootSourcePath, sourceAbsPath)
-				if err != nil {
-					return fmt.Errorf("failed to get relative source path: %w", err)
+
+				if moduleSourcePath != "" {
+					// ensure source path is relative to the source root
+					sourceAbsPath, err := filepath.Abs(moduleSourcePath)
+					if err != nil {
+						return fmt.Errorf("failed to get absolute source path for %s: %w", moduleSourcePath, err)
+					}
+
+					moduleSourcePath, err = filepath.Rel(modConf.LocalRootSourcePath, sourceAbsPath)
+					if err != nil {
+						return fmt.Errorf("failed to get relative source path: %w", err)
+					}
 				}
 			}
 
@@ -368,8 +407,14 @@ This command is idempotent: you can run it at any time, any number of times. It 
 			}
 			// if SDK is set but source path isn't and the user didn't provide --source, we'll use the default source path
 			if modSDK != "" && modSourcePath == "" && developSourcePath == "" {
-				developSourcePath = filepath.Join(modConf.LocalRootSourcePath, defaultModuleSourceDirName)
+				inferredSourcePath, err := inferSourcePathDir(modConf.LocalRootSourcePath)
+				if err != nil {
+					return err
+				}
+
+				developSourcePath = filepath.Join(modConf.LocalRootSourcePath, inferredSourcePath)
 			}
+
 			// if there's no SDK and the user isn't changing the source path, there's nothing to do.
 			// error out rather than silently doing nothing.
 			if modSDK == "" && developSourcePath == "" {
--- a/core/integration/module_call_test.go
+++ b/core/integration/module_call_test.go
@@ -1920,7 +1920,7 @@ func (CallSuite) TestByName(ctx context.Context, t *testctx.T) {
 			`,
 			).
 			WithWorkdir("/work").
-			With(daggerExec("init")).
+			With(daggerExec("init", "--source=.")).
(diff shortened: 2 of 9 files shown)


cmd/dagger/module.go
```diff
@@ -59,6 +59,26 @@ const (
...
+func getDefaultSourcePathDir(srcRootPath string) (string, error) {
+	list, err := os.ReadDir(srcRootPath)
+	if err != nil {
+		return "", err
+	}
+
+	for _, l := range list {
+		// ignore hidden files
+		if strings.HasPrefix(l.Name(), ".") {
+			continue
```
rajatjindal, line 73: are there any specific dirs that we would like to ignore when checking if the current folder has files/folders. Right now it ignores only the hidden folders.

shykes:
I can't think of anything we'd want to ignore. Worst case we add it later?

cmd/dagger/module.go
sipsma: Agree with Solomon that we don't need to check anything extra for now.

However, we probably *should* add an exception for `.dagger` here. Obscure, but if the only thing we find is `.dagger` it probably still makes sense to use it? I could see this mattering in cases where someone is trying to "re-init" by deleting everything in `.dagger` but leaves that dir itself around.

rajatjindal: Hi @sipsma, there are a few scenarios here:

1. `.dagger` exists and no other file exists - currently it will use current dir. I've pushed a change to use .dagger in this scenario.
2. `.dagger` exists and other files exists in the current folder - will already use .dagger
3. `.dagger` does not exist and other files exists - will already use .dagger
4. blank dir - will use current dir

thanks

cmd/dagger/module.go
```diff
@@ -59,6 +59,29 @@ const (
...
+
+		// ignore hidden files
+		if strings.HasPrefix(l.Name(), ".") {
+			continue
+		}
+
+		return ".dagger", nil
+	}
+
+	return defaultModuleSourceDirName, nil
```
jedevc, line 82: nit: we can delete this `const`.

cmd/dagger/module.go
```diff
@@ -147,11 +170,19 @@ If --sdk is specified, the given SDK is installed in the module. You can do this
 				moduleName = filepath.Base(modConf.LocalRootSourcePath)
 			}
 
+			// if the source root path already has some files
+			// then use `srcRootPath/.dagger` for source
```
jedevc, line 174: nit: this doc comment should be on the `getDefaultSourcePathDir` function.

core/integration/module_call_test.go
```diff
@@ -1920,7 +1920,7 @@ func (CallSuite) TestByName(ctx context.Context, t *testctx.T) {
 			`,
 			).
 			WithWorkdir("/work").
-			With(daggerExec("init")).
+			With(daggerExec("init", "--source=.")).
```
jedevc, line 1923: suggestion: could we switch to using `With(withModInit)` everywhere we do `init`?

I've wanted to do this for a while, and it'll make changes where we have to update `init`s everywhere a lot easier. That said, maybe we should merge this one quickly, and maybe follow-up with some test cleanups? I can make a separate issue (cc @marcosnils).

jedevc (approved):
LGTM :tada: Only a couple nits.

Can you also add a changelog with `changie new` as well? It's technically a breaking change, so we should explain that.

core/integration/module_call_test.go
rajatjindal: Yeah, IMO we should do a follow up for this change. But also not strongly opposed to that if you insist :)

core/integration/module_go_test.go
```diff
@@ -430,6 +430,57 @@ func (m *HasNotMainGo) Hello() string { return "Hello, world!" }
 		require.NoError(t, err)
 		require.Contains(t, sourceSubdirEnts, "go.mod", "go.sum")
 	})
+
+	t.Run("init module in .dagger if files present in current dir", func(ctx context.Context, t *testctx.T) {
```
jedevc, line 434: Ah sorry, should have caught this before - I think we should add a test for a `init`+`develop`.

How does it handle:

```
$ dagger init --name=foo
$ ls -a
.  ..  dagger.json
$ dagger develop --sdk=go
```

`dagger.json` should be ignored, and we should put it in `.` instead of `.dagger`.

core/integration/module_go_test.go
```diff
@@ -430,6 +430,72 @@ func (m *HasNotMainGo) Hello() string { return "Hello, world!" }
...
+	})
+
+	t.Run("develop command does not use .dagger when init was in current folder", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		entries, err := c.Container().From(golangImage).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work").
+			With(daggerExec("init", "--name=bare", "--sdk=go")).
+			With(daggerExec("develop", "--sdk=go")).
```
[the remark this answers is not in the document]
rajatjindal: FYI, I am looking at adding these exact steps as a testcase. Kindly do not merge until I can confirm that they are working.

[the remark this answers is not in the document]
rajatjindal: added additional testcases and adjusted code as well accordingly.

cmd/dagger/module.go
```diff
@@ -368,7 +405,16 @@ This command is idempotent: you can run it at any time, any number of times. It
...
-				developSourcePath = filepath.Join(modConf.LocalRootSourcePath, defaultModuleSourceDirName)
+				inferredSourcePath, err := inferSourcePathDir(modConf.LocalRootSourcePath)
+				if err != nil {
+					return err
+				}
+
+				if inferredSourcePath != "" {
+					developSourcePath = filepath.Join(modConf.LocalRootSourcePath, inferredSourcePath)
+				} else {
+					developSourcePath = filepath.Join(modConf.LocalRootSourcePath, defaultModuleSourceDirName)
```
jedevc, line 416: This `const` is back? https://github.com/dagger/dagger/pull/8557#discussion_r1808451125

cmd/dagger/module.go
```diff
@@ -59,6 +59,31 @@ const (
...
+
+		// ignore hidden files
+		if strings.HasPrefix(l.Name(), ".") {
+			continue
+		}
+
+		return ".dagger", nil
+	}
+
+	return "", nil
```
jedevc, line 84:
```suggestion
	return ".", nil
```

Right? The extra handling from having to handle the empty string is a bit confusing.

Why can't we just ignore `dagger.json` in this loop to get the desired behavior?

rajatjindal: I restructured the code a bit more to make this work. While doing so I also found and fixed one more scenario that involved initializing in a sub-dir.

I have also added more testcases to cover the sub-dir scenario and I hope this makes sense now.

thank you for your review comments.

cmd/dagger/module.go
rajatjindal: yep, I made the changes from scratch and missed it. thanks. its removed now.

jedevc (approved):

PR #8563, opened 2024-09-24
docs: Update MacOS sudo instructions

levlaz:
The instructions we have for MacOS will not work because it tells you to `cd /usr/local` and then run the install script, but users do not have permission to install things in `/usr/local` by default.

### DIFF
--- a/docs/current_docs/partials/_install-cli.mdx
+++ b/docs/current_docs/partials/_install-cli.mdx
@@ -30,7 +30,7 @@ If you do not have Homebrew installed, or you want to install a specific version
 
 ```shell
 cd /usr/local
-curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.13.3 sh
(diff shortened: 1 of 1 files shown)


levlaz:
This is related to https://github.com/dagger/dagger/pull/8412

docs/current_docs/partials/_install-cli.mdx
```diff
@@ -30,7 +30,7 @@ If you do not have Homebrew installed, or you want to install a specific version
 
 ```shell
 cd /usr/local
-curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.13.3 sh
+curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.13.3 sudo -E sh
```
marcosnils, line 33: The command below does the exact same thing so this is a bit redundant. Maybe here we can just remove the `cd /usr/local` instead of adding the `-E`? This way the binary will be downloaded into `$workdir/bin/dagger` which is useful for some people that don't like adding `sudo` to their curl | bash installs.

Issue #8520, opened 2024-09-20
API to query current git context


shykes:
## Problem

Sometimes I need information about the current state of the git repository for my context. Things like:

- What's the current commit?
- What's the current branch?
- What tags point to the current commit?
- Is this a clean checkout? If not, what are the uncommitted changes?
- What's the remote?

I can get *some* of this information indirectly with hacks, but it's not 100% reliable.

## Solution

Add a core API to introspect git information about the context.

jedevc:
Is this essentially "contextual git repo" arguments?

Here's my suggestion:
- We extend `// +defaultPath=<path>` to allow annotating `dagger.Git` args. The `path` in this case becomes a path to a `.git` directory (or maybe to a directory containing a `.git` directory).
- We extend `// +defaultPath=<path>` to allow annotating `dagger.GitRef` args. The `path` in this case becomes a path to a checkout of a directory.
- Then, we modify the core Git type - the default implementation is still created from the `dag.Git` query, but we also allow sourcing it from the client automatically using a session attachable.
	- While we're here, we could also allow creating a git directory using `Directory.AsGit`.
- We extend the Git API to allow getting better information about the current commit/branch/tags, working out dirtiness, remotes, etc.

We need to be careful though, and make sure that this behaves the same as the contextual directory - i.e. loading from the callee. When I get a contextual git repo, I want the repo for the callee. But need to think through this - I think it's a much less common use case, the common case is going to be things like our suggestion `version` module.

shykes:
@jedevc did I understand correctly that you're working on this? If so, I'll assign so you can take it over

jedevc:
x-linking https://github.com/dagger/dagger/issues/8720

jedevc:
> @jedevc git diff command will be supported?

Design still up in the air, but essentially you'll get back a `Git` type in the API - whatever you can do with that, you can do with the current git context.

I don't think `diff` will *directly* be supported - but you'll still be able to call `Git.tree` and then mount it into a container with git installed to do the operation yourself. We'll see :tada:

PR #8564, opened 2024-09-25
chore: move version string logic to its own dagger module

shykes:
This is part of wider refactor of our monolithic CI module. See #8355 .


version/main.go
```diff
@@ -0,0 +1,477 @@
...
+	if !v.gitDirExists(ctx) {
+		return fmt.Sprintf("%s-%s", next, digest), nil
+	}
+	commit, err := v.Commit(ctx)
+	if err != nil {
+		return "", err
+	}
+	// FIXME we don't differentiate between clean and dirty checkout:
+	// instead we always add the commit + digest
+	return fmt.Sprintf("%s-%s-%s", next, commit, digest), nil
```
jedevc, line 104: We need to make sure that "dev" builds get a proper "dev" entry in their tag - this is load-bearing: https://github.com/dagger/dagger/blob/42e8daf846dffe91476679d5cf0d3955630c9171/engine/version.go#L133-L138

Also, note on terminology here - there are 3 cases:
- Release builds (have an associated tag)
- Main builds (have an associated commit, but no tag)
	- These are still technically released and uploaded to the registry and s3
- Dev builds (made locally, have no associated commit or tag)
	- This isn't the same as a `main` build
I think we need to keep this terminology, see the above load-bearing code snippets.

That said, I'm not 100% sure how we do this. Ideally dev should be set only on dirty checkouts, but I'm not quite sure how we compute that. The only thing I can think of is to import *everything*, so we can run a `git status`, but that's *so* expensive. We really need contextual git #8520, but not sure what a temporary workaround looks like.

*Maybe* we could have the main dagger module keep track of a `--dev` flag? Then when building locally, you have to make sure to set a `--dev` flag (we can set this automagically in `./hack/dev` that most of us use).

version/main.go
```diff
@@ -0,0 +1,477 @@
...
+		// minor version
+		nextVersion = definedNextVersion
+	}
+	return nextVersion, nil
+}
+
+func pseudoversionTimestamp(t time.Time) string {
+	// go time formatting is bizarre - this translates to "yymmddhhmmss"
+	return t.Format("060102150405")
+}
```
jedevc, line 470: Also, this logic is actually important - this is what allows us to say "this commit on main is older that this other commit from main" - without this, there's no way to compare, since commits aren't sortable (so the timestamp allows us to do this).

I'd like to keep the format we currently have.

version/main.go
```diff
@@ -0,0 +1,477 @@
...
+	// FIXME: contribute GitRepo.head() uptsream
+	commit, err := v.gitRepo().Command(cmd).Stdout(ctx)
+	if err != nil {
+		return "", err
+	}
+	return strings.Trim(commit, "\n"), nil
+}
+
+// Compute a stable digest of the input files, to differentiate dev builds on the same commit
+// FIXME: the digest doesn't seem very stable...
```
jedevc, line 246: We can use `Directory.Digest` now :tada:

version/main.go
```diff
@@ -0,0 +1,477 @@
...
+func (v Version) InputsDigest(ctx context.Context) (string, error) {
```
jedevc, line 247: This and some of the other functions in here should be private - since they're internal implementation details.

jedevc:
Did some tidying up - ended up removing `supergit` in preference of some more opinionated git methods in `git.go`, which allows for easily grabbing lots of tag information along with their dates / etc (we can move this out in the future).

Also fixed the `dev` versions, so that now we also handle newly added files.

Finally, refactored to get the old behavior of version numbers back - if we change the format, we should do so as a separate change.

version/git.go
```diff
@@ -0,0 +1,231 @@
...
+		// module sources)
+		remote := "https://github.com/dagger/dagger.git"
+		ctr = ctr.
+			// we need the unshallowed HEAD, so we can determine which tags are
+			// in HEAD's history later
+			WithExec([]string{"git", "fetch", "--unshallow", remote}).
+			// we need all the tags to determine the latest tags later
+			WithExec([]string{"git", "fetch", "--no-tags", remote, "+refs/tags/*:refs/tags/*"}).
+			// we need main, so we can determine the merge base for it later
+			WithExec([]string{"git", "fetch", "--no-tags", "-u", remote, "refs/heads/main:refs/heads/main"})
```
jedevc, line 33: It *really* sucks that we need all this extra `fetch`ing here, but we can do some really really cool things with it (e.g. we work out the last common `main` commit, so we are sure that that tag will actually exist).

However - contextual git should be able to handle this one day, we should be able to allow "unshallowed" clones.

jedevc: Note to self: maybe we could do these steps conditionally, by detecting if the repo is a shallow clone first (that way we don't need it for the local case) - that way we can avoid doing network access if we don't need to.

shykes: As discussed in discord: it's totally fine for now. We can optimize later.

shykes:
Thank you @kpenfound @jedevc ! Great teamwork

PR #8568, opened 2024-09-25
chore: rename dagger-engine to engine GraphQL query

marcosnils:
** DO NOT MERGE BEFORE v0.14 **

This way we can use `dagger core engine` instead of `dagger core dagger-engine` which feels weird.

```[tasklist]
### Follow-ups
- [ ] https://github.com/dagger/dagger/issues/8878
- [ ] https://github.com/dagger/dagger/issues/8879
```


core/schema/engine.go
```diff
@@ -17,7 +17,7 @@ var _ SchemaResolvers = &engineSchema{}
 
 func (s *engineSchema) Install() {
 	dagql.Fields[*core.Query]{
-		dagql.Func("daggerEngine", s.engine).
+		dagql.Func("engine", s.engine).
```
jedevc, line 20: Note: this is a breaking change.

Can we create a new method and deprecate the old one? I don't think anyone *is* depending on it yet, but it's good to avoid breaking things in patch releases wherever possible.

marcosnils: sure, SGTM. Adding a new method seems ok.

helderco: Make sure to deprecate the old one 😄

marcosnils: yep, will mark the old one as deprecated :+1:

jedevc: As mentioned below - this *isn't* a breaking change for modules - which is what we care about.

The API still changes, so `dagger core dagger-engine` becomes `dagger core engine`, and external SDKs are still breaking, but modules don't have access to this type *in the first place*. So I think we can take this whenever.

jedevc (approved):
LGTM :tada: (no deprecation needed, as discussed)

Just needs a breaking changelog, and should be good to go :shipit:

marcosnils:
> LGTM 🎉 (no deprecation needed, as discussed)
> 
> Just needs a breaking changelog, and should be good to go :shipit:

Done @jedevc :pray: 

CI will complain as I had to rename our `Engine` type to `DaggerEngine` to avoid collisions as commented during our call.

jedevc:
I actually don't know if we need to rename the type - because that type isn't generated, there's no risk of a conflict.

Our testdev workflows would have caught that it that was the case.

sipsma:
@marcosnils that SGTM, I will hold off on merging this until right before v0.14.0 to limit the churn time of the temporary name

sipsma:
I noticed that while the "missing lazy blobs" error in logs is *often* gone, it still shows up sometimes. After poking a bit more I really can't see how this is anything but a buildkit bug since it's also inconsistent which blobs are missing.

The change still seems to improve the situation by making it happen less often and the error is non-fatal (just results in duplicate cache volumes being created inside buildkit), so I will go with it for now but open a follow-up issue to figure out wtf is happening there.

marcosnils:
> @marcosnils that SGTM, I will hold off on merging this until right before v0.14.0 to limit the churn time of the temporary name

:+1:  I'm adding a task-list at the top comment with corresponding issues for 0.14.1 so we don't forget about this

marcosnils:
> @marcosnils does this change impact the examples in https://docs.dagger.io/configuration/cache?

 yes, forgot about the docs :scream: . Can you help me with that Vikram?

sipsma (approved):

PR #8569, opened 2024-09-25
feat: use context directory and ignore in TypeScript runtime

TomChv:
⚠️ Currently broken, I need to find a way to fetch the go codegen binary from Git

```
invoke: failed to create codegen base: failed to add template: could not list module source entries: input: directory.withDirectory.withoutDirectory.file resolve: lstat /codegen: no such file or directory
```


TomChv:
converting that one to draft since it's not doable yet.

PR #8572, opened 2024-09-25
chore(refactor): make engine set optional on defaultPath so SDKs don't have to

helderco:
This refactor makes the engine enforce `withOptional(true)` on function arguments that use `defaultPath`. Otherwise it's a footgun for the SDKs to ensure, especially since `withOptional` is incorrectly named.

Context:
- https://github.com/dagger/dagger/pull/8411#discussion_r1775675521


jedevc (approved):

charjr (approved):
For the PHP SDK: LGTM.

I believe this is the only change required.

Thank you for fixing this :100:

Issue #8385, opened 2024-09-09
ci(flake): socket default not found


grouville:
There seems to be a flake / race on the socket resolving, when running the test suite in parallel: `socket default not found`.

A bit confused by this error: https://dagger.cloud/dagger/traces/6d3e1fd82f7bb9ee5ff855155a6f9e97?span=c24cdc1442769090#c24cdc1442769090:L5

From current understanding, I think this flake is the result of a deeper issue on the socket / git resolving, not at the test suite level, but at the engine itself

It seems to originate from https://github.com/dagger/dagger/blob/main/core/schema/modulesource.go#L164-L167, which is a bit surprising. Basically, once we have the proper ref, we try to resolve the commit with https://github.com/dagger/dagger/blob/main/core/git.go#L69-L87. What it says here is that: I received an SSH key, but failed to find / retrieve the socket:

```go
func (ref *GitRef) Commit(ctx context.Context) (string, error) {
	bk, err := ref.Query.Buildkit(ctx)
	if err != nil {
		return "", fmt.Errorf("failed to get buildkit client: %w", err)
	}
	st, err := ref.getState(ctx)
	if err != nil {
		return "", err
	}
	p, err := resolveProvenance(ctx, bk, st)
	if err != nil {
		return "", err
	}
	if len(p.Sources.Git) == 0 {
		return "", errors.Errorf("no git commit was resolved")
	}
	return p.Sources.Git[0].Commit, nil
}
```

Error is triggered probably by either `getState()` or `gitdns.Git()`. Have to dig a bit deeper, but tricky as hard to replicate

jedevc:
As suggested by @sipsma in https://github.com/dagger/dagger/pull/8574#issuecomment-2375505502, I think this is potentially caused by the caching socket issue, very similar to what I described around secrets in https://github.com/dagger/dagger/pull/8358#issuecomment-2346006330 (though the solution we used there doesn't really apply).

Marking `Git.tree` and `Git.tags` as impure feels like it would work - but at least for `Git.tags`, that means that the result would *never* cache (`tree` does because it goes through buildkit), which feels bad.

Potentially we could just remove this flake with just `Git.tree` impurity, and we *should* add it to `Git.tags` but it feels so over the top. I wonder if there's a "lighter" form of dagql cache invalidation we could use here.

sipsma:
I don't think I've seen this one since https://github.com/dagger/dagger/pull/8574 was merged, so may be fixed, but will give it another week or two before closing out

[...]

PR #8574, opened 2024-09-25
fix: flake on TestSSHAuthSockPathHandling test

### LINKED ISSUE #8571: Flake?: `TestSSHAuthSockPathHandling`
Seeing this test fail in #8442 but it doesn't seem related.

Example:

* https://dagger.cloud/dagger/traces/592be290e8f8cba6b3aca5baf50ae88b?span=005d3cee7b94ac88

```
Error: module at source dir "" doesn't exist or is invalid
```

Looks like this test was added in https://github.com/dagger/dagger/pull/8365

cc @grouville

grouville:
Closes #8571

When testing the TestSSHAuthSockPathHandling, Alex encountered rightfully a flake on this test, leading to this error:

```
Error: module at source dir "" doesn't exist or is invalid
```

It turns out that the error comes from the tests being run in parallel whilst relying on the same client

More context [here](https://github.com/dagger/dagger/issues/8571#issuecomment-2375333486). After the change, no flake after 30 runs on my side

> Still a bit surprised that in a nested environment, when interacting with the host, the same client cannot be used in parallel. Is it my test / setup that is wrong or is it by-design / expected ? 

The very reason that it sometimes work would be that one of the test sets the nested SESSION's env vars prior the second one's erasing the HOME var, or before ?

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -5527,11 +5527,10 @@ func (ModuleSuite) TestSSHAgentConnection(ctx context.Context, t *testctx.T) {
 }
 
 func (ModuleSuite) TestSSHAuthSockPathHandling(ctx context.Context, t *testctx.T) {
-	c := connect(ctx, t)
-
 	repoURL := "git@gitlab.com:dagger-modules/private/test/more/dagger-test-modules-private.git"
 
 	t.Run("SSH auth with home expansion and symlink", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
 		mountedSocket, cleanup := mountedPrivateRepoSocket(c, t)
 		defer cleanup()
 
@@ -5555,6 +5554,7 @@ func (ModuleSuite) TestSSHAuthSockPathHandling(ctx context.Context, t *testctx.T
 	})
 
 	t.Run("SSH auth from different relative paths", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
 		mountedSocket, cleanup := mountedPrivateRepoSocket(c, t)
 		defer cleanup()
 


sipsma:
The error message you were getting here is very weird, poked around and realized that if you look in the engine logs you actually can see:
```
[buildkit] [client=2zpsohynpqyijy61a4upbhlt6] #3 git://gitlab.com/dagger-modules/private/test/more/dagger-test-modules-private.git
[buildkit] [client=2zpsohynpqyijy61a4upbhlt6] #3 ERROR: NotFound: rpc error: code = NotFound desc = socket xxh3:f9b3115dbb65e64e not found
```

So I think this *is* actually related to https://github.com/dagger/dagger/issues/8385

The reason the error message ends up being `module at source dir "" doesn't exist or is invalid` is not 100% clear to me yet, didn't dig in that deeply yet, but it has something to do with the fact that [this line](https://github.com/sipsma/dagger/blob/583b8437d5f2004008562bb24de6b1aaa38d0b45/cmd/dagger/module.go#L619-L619) is saying the Git module source config file doesn't exist yet.
* I'm guessing that it doesn't exist because of the error that showed up in the logs above? But would have liked this to fail more cleanly
* Also the error message itself is extra confusing because it is assuming the module was local, not git (hence the `source dir ""` part of the message). This should probably be improved separately.

I'll keep poking at this for a few in parallel with you since it turns out to be one the good flakes that are actually reproducible locally 🎉

---

If it comes down to it, I'm totally fine merging this PR as is to avoid the flakes. These nested exec setups are not anything close to what real world users would do, so it's not a huge loss of test coverage if we just avoid whatever this problem is by giving each test case its own client. Worth a little bit of effort to dig into though in case it reveals any problems that are user-facing after all though.

sipsma:
After some println debugging, the problem appears to be that when concurrent requests are made by the clients, *sometimes* only one of the `Query.git` calls actually executes between them? At least, when I add a log to [this line](https://github.com/sipsma/dagger/blob/583b8437d5f2004008562bb24de6b1aaa38d0b45/core/schema/git.go#L151) and the tests fail, only one of the clients actually prints anything from it. I can however see both clients are hitting the `git` source implementation in buildkit, which is where the socket is requested and fails in *one* of them (the one with the missing `Query.git` call).

So that all adds up: a client never executes `Query.git`, which is where the socket is registered in its store, and then when buildkit goes to the store to ask for the socket, it's not there.

You can also see this by running the tests with `-parallel=1` and without the change in this PR: that also avoids the flake for me.

I think the reason only one of the clients actually executes `Query.git` is dagql caching: that API is not marked impure and so it gets deduped based on its dagql ID, which is identical for the two clients.

The last thing I don't understand is the fact that when I marked `Query.git` as impure (which *technically* would be more correct since it can end up implicitly accessing+registering these host dependent sockets), I still get the same behavior as before.
* I'll poke at this for a little bit longer, but consider it separate from all this.

---

Overall, I don't think it's really productive for us to care about any of this. These requirements for making some API calls impure because there could *technically* be multiple nested clients in the same session using the same paths for different resources is something that exists only inside our test framework, not the real world. So we are honestly better off ignoring things like this for now IMO rather than going down unnecessary rabbit holes (we have plenty of necessary rabbit holes first).

@grouville tl;dr lets go with just giving each test case its own client, as you did in this PR. It might be worth a minute of your time to check if any of the other SSH socket related tests are sharing a dagger client and make the same change to this ones.

cc @vito (related to impurity, caching and all that fun stuff if you're in the mood)

sipsma (approved):
LGTM, with added suggestion mentioned in other comment that it might be worth making the same change to the other SSH socket related test cases if needed.

jedevc (approved):

jedevc:
Ohhh @sipsma could it be because the socket access only actually occurs in `tree`? We wouldn't want to mark `Query.git` as impure, we'd want to mark `Git.tree` and `Git.tags`.

Still seems good to me though, but also... confused. I feel like this only started appearing yesterday, about when I merged https://github.com/dagger/dagger/pull/8318. But really really unsure as to why this would have any effect at all.

Issue #8136, opened 2024-08-09
🐞 Include/exclude from module dependencies not prefixed with subdir to module


helderco:
When a module is used as a dependency inside the same git repo, the dependency’s include/exclude patterns aren’t prefixed with the path to that module, relative to the repo root (context directory). When the module is called directly, the prefix is added correctly, which means that the patterns from dependencies 

For example, suppose the `sdk/python/dev` module has the following exclude:

```json
{
  "name": "python-sdk-dev",
  "sdk": "python",
  "exclude": [
    ".venv"
  ]
}
```

If you run the module directly, the final list of excludes when loading the module, include the subdir to the module:

<img width="772" alt="Screenshot 2024-08-09 at 17 02 42" src="https://github.com/user-attachments/assets/15c84ec4-e1c6-448a-9096-c6028162cd59">

However, the global `dev` module, which depends on `sdk/python/dev`, doesn’t have the right final excludes, because the subdir to `sdk/python/dev` isn’t being prefixed to those patterns:

<img width="762" alt="Screenshot 2024-08-09 at 17 10 15" src="https://github.com/user-attachments/assets/d389af04-8231-406d-939f-4c60d55a65a0">

This can cause a lot of issues, see for example:
- https://github.com/dagger/dagger/pull/8045

\cc @TomChv

PR #8575, opened 2024-09-26
fix include/exclude from dependencies not prefixed correctly

helderco:
Fixes https://github.com/dagger/dagger/issues/8136

Note: I switched the order in the UI to show includes before excludes because that's the order that buildkit applies them.


helderco:
Added the test case.

sipsma (approved):

PR #8576, opened 2024-09-26
feat: ts-sdk replace `fetch` with `node-fetch`

TomChv:
Following up with #8549, it seems the issue is caused by the native `fetch`. I'm trying to replace the client with `node-fetch` which is a more optimized version of `fetch` for server-side request.


jedevc (approved):
Sure, why not :smile: 

If we could get user feedback before we release tomorrow, that would be perfect - just to confirm if this actually fixes the issue.

Issue #7898, opened 2024-07-12
✨ Allow me to keep tags in dagger.json 


levlaz:
### What are you trying to do?

Right now if I add a tag to dagger.json, it will install the correct version of my module, but it will overwrite the dagger.json to show the whole git sha. It would be nice to keep the tag visible because it's easier for me to look at a module with many installed modules and look at the human readable version to see what version I am on. 

### Why is this important to you?

Keeping the human readable version around is a smoother DX because it reduces cognitive load. Without it I need to go poke around in various places to see what version I am on.

jedevc:
Agreed - what we should really have here is a way of attaching *both* a tag and a git sha in a module ref. When we resolve a module ref's tag, we should rewrite it to include both - that way, you get the pinning and the readability all in one.

For example, you can do this with docker images: `<image>:<tag>@sha256:<hash>`. Not *quite* sure what that would look like, maybe `github.com/dagger/dagger/dev@<tag>:<hash>`?

Alternatively, the `dagger.json` could have two fields, `source`, which has the name it was installed by, and `resolved`, which has the full hashed name it was resolved to.

One cool thing about tracking the tag we used at the beginning, is that if a tag is updated (e.g. `v1.2` updated from `v1.2.1` to `v1.2.2`), then this could be updated to the latest version here - but don't know if this is something we're interesting in actually supporting as a common use case.

jedevc:
Since this is kind of related to module ref formats, cc @grouville who is owning lots of stuff around that :tada:

nadirollo:
Our use case could benefit from something related to this.

Context: 
Our team owns plenty of dagger modules provided to the rest of the engineering team at our company. 

Once an engineer writes its own dagger module/pipeline and installs one of our maintained modules as a dependency, we lose all ability to push updates and bugfixes to the dependency module (it requires action from the engineer to do `dagger install` again).

We are coming from Jenkins world where we provide the same type of service through a shared library where we have "rolling" tags. Engineers will write their pipelines using the `stable` tag, and they will have fixes and features pushed to the library without having to take any action. 

I understand how pinning to a version that will be changing is not the current model, but there's a case for that (at least in our company). 
Is this something that could be considered? Or any ideas on how this could be implemented in the current model?

shykes:
I also need this! https://github.com/dagger/dagger/issues/8271

PR #8587, opened 2024-09-27
feat: add module source pins

jedevc:
Fixes https://github.com/dagger/dagger/issues/7898.

This PR modifies the module config `dependencies` field to ensure that we preserve the original tags that the user specified during `dagger install`.

So with a command like:

```
$ dagger install github.com/dagger/dagger-test-modules
```

We get the following (see the diff between before and after):

```diff
diff --git a/dagger.json.old b/dagger.json
index 4ec8db5..533fc4a 100644
--- a/dagger.json.before
+++ b/dagger.json.after
@@ -4,7 +4,8 @@
   "dependencies": [
     {
       "name": "root-mod",
-      "source": "github.com/dagger/dagger-test-modules@323d56c9ece3492d13f58b8b603d31a7c511cd41"
+      "source": "github.com/dagger/dagger-test-modules@main",
+      "pin": "323d56c9ece3492d13f58b8b603d31a7c511cd41"
     }
   ],
   "source": ".",
```

This essentially allows us to keep the original source that we installed a dependency via, while also keeping the pins.

> [!NOTE]
>
> Naming note - I chose `pin` because that matches up with terminology in other tooling, and it's also generic. If we ever support non-git remote source (e.g. OCI images), then `pin` is more generic and correct than `commit`.

To gracefully upgrade old `dagger.json`s, we leave `source` unaltered, and copy the `pin` from the commit in the `source` (which is always guaranteed to be there).


core/integration/module_config_test.go
```diff
@@ -1193,3 +1193,110 @@ func (m *Test) Fn(dir *dagger.Directory) *dagger.Directory {
 	require.NoError(t, err)
 	require.Equal(t, "", strings.TrimSpace(out))
 }
+
+func (ConfigSuite) TestDepPins(ctx context.Context, t *testctx.T) {
```
sipsma, line 1197: It would be very nice to have a test where the pinned commit is behind HEAD on the branch and we verify that the pinned commit is used correctly

jedevc: Added a test for this using https://github.com/dagger/dagger-test-modules/pull/5, and another one for testing that `dagger develop` doesn't modify the pins.

jpadams:
cc @vmaffet-fast

core/integration/module_config_test.go
sipsma: Approved that one, will let you press the merge button in case anything else needed

sipsma (approved):

PR #8589, opened 2024-09-27
ci: fix version module when no git metadata is available

jedevc:
If an empty directory was passed to the version module, then `.git` wouldn't be available, and so all the git operations would fail. The fix is to detect it before running any of the commands.

But to do that, we need a context and an error - so we just move all this operation straight into the constructor, which also avoids a ton of extra calls to the `Git` function anyways.

### DIFF
--- a/version/git.go
+++ b/version/git.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"dagger/version/internal/dagger"
+	"errors"
 	"slices"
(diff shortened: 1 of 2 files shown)


shykes:
Merging, @jedevc has confirmed he will be on-call 24/7 to fix this if it goes wrong, thank you Justin! 😛

shykes (approved):

PR #8593, opened 2024-09-29
chore(ci): dont send interrupt signal for foreground process

rajatjindal:


rajatjindal:
> Can you explain the logic for this?

> The original reasoning for sending interrupt signals for processes is that we want to ensure that the process has terminated for the test to complete - if it hangs forever, the test should fail (no dagger command should continue on being sent a CTRL+C). We also don't want dangling processes hanging around.

The "interrupt" signal msg in the trace https://dagger.cloud/dagger/traces/b3d53b83e726ac25892517310b0ee9d2#2e580acd7b4b57ea:L3 you noticed in https://github.com/dagger/dagger/pull/8586 is coming from sending that signal to "dagger init" and "dagger functions" command runs. These commands are running in foreground, and run to completion, so I thought it should be ok to not have a cleanup function for it. But I guess what you are saying is that these commands can get hung as well, and that cleanup function should be useful in those cases?

Also, the tests itself seems to work fine and I was not able to reproduce a failure or hang in those tests. I also did some negative tests to ensure tests fail when the server returns unexpected msg, which worked as well. 

In the original trace the test shows "Cancelled", but I could not reproduce that error. I will keep an eye on random CI failures to check if I see this happening again.

jedevc:
> These commands are running in foreground, and run to completion, so I thought it should be ok to not have a cleanup function for it.

They *should* run to completion. But clearly they weren't - see https://github.com/dagger/dagger/pull/8191.

Issue #5774, opened 2023-09-11
[Zenith] `dagger script`: invoke any function from any module, in your favorite language, from the CLI


shykes:
## Overview

In #5757, we take the Zenith in a bold new direction:
* everything is a function
* functions are packaged together in modules
* each module has an API
* A module's API can be queried either 1) from other functions, or 2) **directly by the end user**

But for this design to work, we need **a great UI for querying a module's API**. More specifically, we need **a great UI for running a module's functions** (using its API).

We think Dagger will support many ways to do this, including:

* Crafting raw graphql queries with `curl`, `postman` or other HTTP/graphql clients
* Crafting a query from CLI arguments and flags, using `dagger do` or an equivalent subcommand of the Dagger CLI
* Using a custom web UI, either running locally, in Dagger Cloud, or in an IDE plugin
* Crafting a query in a familiar programming language, with the help of a specially packaged language interpreter

Of all these options, I propose that we focus primarily on the last (familiar language interpreter), and call it `dagger script`

## `dagger script`: call Dagger functions from the command-line, in a familiar language

This section is work in progress. Bikeshedding is expected and encouraged :)

### Overall architecture

FIXME

### The versioning problem

One unresolved question is how to manage versioning of the various pieces (CLI, engine, SDK), taking into consideration ongoing changes (#5583 , #5484), in a way that 1) makes `dagger script` great to use, 2) doesn't interfere with ongoing design work, 3) ongoing design work doesn't later break `dagger script`.

--> FIXME, add details here. For now, see [this discord thread](https://discord.com/channels/707636530424053791/1148666309324845167/1148710601338654760)

helderco:
This sounds a lot like `dagger run` today, rather than the "one-liner" we discussed a few times, based on the direct access to the user's system. Unless you mean through an API?

sipsma:
Yeah agree reading this sounds more and more like https://github.com/dagger/dagger/issues/5993; if you want to write a script that can call out to your functions, you will need codegen bindings to those functions.

I think if you take that and combine it with the ability for `dagger run` to compile your code for you rather than require host dependencies (i.e. `dagger run main.go` rather than `dagger run go run main.go`) you'd end up with a pretty simple and easy way to write complex logic that executes direct on your host and calls out to functions.

I've been thinking through other variations too in that issue, but something along these lines feels right. I'm hoping to pick up #5993 in the near future, so should be able to start playing around with ideas soon 🤞

shykes:
[...]

Issue #8025, opened 2024-07-24
✨ Import image to local container runtime


jedevc:
The CLI should support loading images into the local container image store - whether that's docker, containerd through `ctr`/`nerdctl`, podman, etc.

Today, this might look something like:

```
$ dagger call container-echo --string-arg=foo -o ./image.tar
$ docker load -i ./image.tar
```

However, we should really have a *native* way to do this - something like:

```
$ dagger call container-echo --string-arg=foo --load <image>:<tag>
```

Why?
- Makes it easier for dagger to exist as a full replacement for docker build, as in @verdverm's use case illustrated in discord https://discord.com/channels/707636530424053791/1265384148856471622/1265384148856471622
- We want to avoid users feeling the need to wrap dagger using another cli tool as much as possible - not having an integrated way to do this common task is sad.
- The current approach is slow - even if we have a containerd content API which avoids uploading layers that weren't changed, we can't really take advantage of it (currently, not exposed in dockerd, but somewhere on the roadmap see https://github.com/moby/moby/issues/44369)

So how would this work?
- The CLI should have support for a generic "loader" interface - these take a core container object, and somehow "load" them. We can implement two loaders to start:
	- `docker load` (the default)
	- Containerd content store streaming, implemented using the session attachables that buildkit itself has - we can allow forwarding a local containerd socket api, or opening a local content store (as the `oci` export in buildkit uses).
- The syntax could then look something like:
	- `dagger call container-echo --string-arg=foo --load <image>:<tag>` to load into the default `docker` store
	- `dagger call container-echo --string-arg=foo --load ./path/to/oci/layout/<image>:<tag>` to export to a local [oci layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md)
	- `dagger call container-echo --string-arg=foo --load ./path/to/containerd/socket` to load to the containerd store accessible via that API 

Related:
- https://github.com/dagger/dagger/issues/6271

mihaelTBTL:
Is this also related to #6999? I had a similar [use case](https://github.com/dagger/dagger/issues/6999#issuecomment-2203203921) to #7368, but ultimately didn't manage to get it working as expected.

Even with the described current practice:
```
$ dagger call container-echo --string-arg=foo -o ./image.tar
$ docker load -i ./image.tar
```
I'd run into this issue where the image name and tag would not be preserved when loading the tarball.
The export method of container didn't seem to behave like `docker image save`.

mihaelTBTL:
[...]

PR #8611, opened 2024-10-01
dsh

shykes:
## Context

- https://github.com/dagger/dagger/issues/5774

## Todo

> [!note]
> This is a first iteration. Remaining features could be implemented in separate PRs.

- [x] Run with REPL (`dagger shell`)
- [x] Feature parity with `dagger call` (object handling and printing)
- [x] Run with inline code (`dagger shell -c container`)
- [x] Run with stdin (`echo 'container' | dagger shell`)
- [x] Run with external script (`dagger shell script.sh`)
- [x] Fix REPL interference with TUI
- [x] Fix builtins in non-interactive modes
- [x] History
- [x] `.functions` to render like `dagger functions`
- [x] Support module constructor arguments
- [x] Allow targeting core types
- [x] Allow stitching with subshells
- [x] Circumvent shell builtins (e.g., "echo") to handoff to our handler
- [ ] Call core functions without having to load a module
- [ ] Allow targeting module dependencies
- [ ] Completions


shykes:
🍿❤️🙏

jedevc:
Hm, we might want to use an alternative library at some point. :thinking: 

The vim emulation in `chzyer/readline` is not great :cry:

jedevc:
Pushed a quick commit for a little nit I had - otherwise, I don't think there's anything blocking shipping this as experimental :tada: :tada:

A couple notes for later (absolutely non-blocking):
- nit: `.functions` doesn't list all the functions - no core functions.
- Variables don't seem to work - can't see this listed on the TODOs. Is it in scope (eventually)?

And a couple design things that I think are worth discussing before we finalize the design (and make the feature stable):
- Not really sure about mixing in shell commands at the top-level
  ```
  $ dagger shell
  Dagger interactive shell. Type ".help" for more information.
  > whoami
  jedevc
  ```

  How do we handle conflicts - e.g. if I have a function `whoami` in my module. Is there any way of specifying I definitely want the external command version of it? Maybe we could put external commands behind a special prefix?

  If it's allowed, is it something that's "best practice"? It makes a dagger script less portable between different platforms - I might not be able to run a script someone else has written. It also might be a security issue if we're potentially interested in integrating this into cloud/native ci?

  Potential proposal: could we have it available with a special prefix *only* in the interactive mode? It makes perfect sense to have it available there, I'm just a bit less sure about having it in written scripts.

- I'm still not 100% convinced on the design of `.config` - it feels a bit clunky, and I'm not sure I'd expect required constructor arguments to work. It works fine for a "singleton" module, but if a module is designed to be easily instantiated several times, then I don't think it's currently possible to work with.
  
  That said, my best alternative suggestion is to require the module name as a prefix for *all* module commands (with a magic `.` builtin or something that gets the current module). Not a huge fan of that either though, so :shrug:

jedevc (approved):

helderco:
> Pushed a quick commit for a little nit I had - otherwise, I don't think there's anything blocking shipping this as experimental 🎉 🎉

That passed by me but I agree with the change, thanks!

> A couple notes for later (absolutely non-blocking):
> 
> * nit: `.functions` doesn't list all the functions - no core functions.

Yeah, for now, to list core functions just do `.core` (without arguments). We're missing the `--help` on a function to see available arguments and next functions, but leaving it for another PR, probably via a builtin `.doc` command.

I'm planning on updating this PR's description to document how to use it. Not sure if there's a better place.

> * Variables don't seem to work - can't see this listed on the TODOs. Is it in scope (eventually)?

It's in scope eventually, just haven't even tested it for now.
  
>   How do we handle conflicts - e.g. if I have a function `whoami` in my module. Is there any way of specifying I definitely want the external command version of it? Maybe we could put external commands behind a special prefix?

To give a bit of context, the handlers run in the following order:

1. Interpreter built-ins (e.g., `shell`, `exec`, `true`, `false`, etc)
2. Defined shell functions
3. **The dagger shell custom handler** (Dagger functions)
4. The default handler, which executes simple commands on the host (e.g., `cat`)

So if there’s a naming conflict with an interpreter built-in, it wouldn’t be possible to have a Dagger function with the same name, so we favor the function but allow prefixing the command with `..` to use the builtin instead. For example, if you have a function called `echo`:

```shell
container | from $(..echo alpine) | file /etc/os-release
```

However, the default handler only executes when there’s no module or core function with the same name. I haven’t created an explicit prefix for executing those directly, but you should be able to use the `exec` builtin:

```shell
..exec whoami
```

>   If it's allowed, is it something that's "best practice"? It makes a dagger script less portable between different platforms - I might not be able to run a script someone else has written. It also might be a security issue if we're potentially interested in integrating this into cloud/native ci?

Yeah, you’re right. I added it because I felt being able to use `cat`, `ls`, etc, could be useful. Let’s talk about it.

>   Potential proposal: could we have it available with a special prefix _only_ in the interactive mode? It makes perfect sense to have it available there, I'm just a bit less sure about having it in written scripts.

Yes, that’s possible.

> * I'm still not 100% convinced on the design of `.config` - it feels a bit clunky, and I'm not sure I'd expect required constructor arguments to work. It works fine for a "singleton" module, but if a module is designed to be easily instantiated several times, then I don't think it's currently possible to work with.
>   That said, my best alternative suggestion is to require the module name as a prefix for _all_ module commands (with a magic `.` builtin or something that gets the current module). Not a huge fan of that either though, so 🤷

Me neither, let’s discuss. The single `.` is an interpreter builtin by the way (used to source a script). Would rather avoid that one if possible. For clarity, if current module has a single required argument for a `Directory`, this is how you call it now:

```shell
.config .; my-func
```

But, there’s actually an advantage to having it be a separe command. Because you can avoid passing it in repeatedly when calling functions:

```shell
> .config .
> my-func1
> my-func2
```

If it had to be a part of the command chain, you’d have to be more repetitive, however the same verbosity could happen if you usually repeat half of the chain, and change functions further down.

So one solution could be to save partial chains in a variable:

```sh
PY=$(.config sdk/python; sdk | python)
$PY | test
$PY | lint
```

💡 Tangentially, I’m thinking about allowing `&` at some point to define functions that should run concurrently:
```
sdk | python | test & lint
```

helderco:
> [...] I haven’t created an explicit prefix for executing those directly, but you should be able to use the `exec` builtin:
> 
> ```shell
> ..exec whoami
> ```

Doesn't work, we'll need to figure something out.

jedevc:
> Yeah, you’re right. I added it because I felt being able to use `cat`, `ls`, etc, could be useful. Let’s talk about it.

Agreed, I think this is very useful, especially for developing a module. e.g. it would be *so* nice if I could start my day with `dagger shell`, and keep reloading the module into it, and do all of my manual shell stuff through `git` and `ls` and `etc`, without needing to keep dropping out.

I'm kinda wondering what it would look like in a script though. I'm struggling to come up with use cases that can't be handled by series of dagger API calls (or what we should have).

One interesting case I'd like is to handle `docker import` so we could entirely replace `hack/dev` - but ideally we'd have something like https://github.com/dagger/dagger/issues/8025 and have this directly in our API. That's really the only one I can think of off the bat though.

I think having just the `.exec` builtin available in interactive mode is fine or similar. We could even have a special `!` line prefix, which takes the entire rest of the line and directly passes it to the shell (so we wouldn't have to handle shell parsing, redirects, etc, in this case). I know GDB (and some other debuggers) do this?

> If it had to be a part of the command chain, you’d have to be more repetitive, however the same verbosity could happen if you usually repeat half of the chain, and change functions further down.
>
> So one solution could be to save partial chains in a variable:

This is what I was imagining, yeah. I think without variables, the DX would be so bad and repetitive, it's probably not worth considering something that's not `.config`. But if we do have variables, then I think it becomes a viable option?

Maybe we could have this?

```
# no constructor args
> dagger-dev | engine | container

# constructor args
> dagger-dev --docker-cfg=... | engine | container

# save constructor, use later
> DEV=$(dagger-dev --docker-cfg=...)
> $DEV | engine | container
```

The last one isn't quite as short as:

```
> .config --docker-cfg=...
> engine | container
```

But it also:
- Allows multiple `dagger-dev` instances all at once easily, without global state of constructors
- Name scoping is nicely handled, the top-level functions aren't all the top
	- If we ever decided to allow *multiple* modules as part of a single shell, then this could help scope which functions belong to which.
- Gracefully handles required constructor args with the same DX as for any required args.
	- e.g. imagine using a "go build" module that takes a required directory - having to start every shell session with `.config path/to/directory` is a bit annoying.

Maybe we could get the best of both worlds, with a user-modifiable PATH:
- When looking up a top level command, we look it up in the PATH
- The default path is `root` (or maybe it's `root` + `user module` + `core`)
- The user can add more things to the PATH - e.g. if you're working on a specific variable a lot, you could `.path-add $VAR`, and everything from that scope would be added to the top-level.

PR #8615, opened 2024-10-02
chore: upgrade go to 1.23.2

sipsma:
I was getting ephemeral panics from the Go runtime on my OTEL metrics PR that (seemingly?) disappeared when I upgraded to the new version of Go. Not sure precisely which change fixed it (they were "invalid pointer found in heap" errors) but there were a few fixes to the runtime and open issues that seemed possibly related, so worth an upgrade.

### DIFF
--- a/.dagger/cli.go
+++ b/.dagger/cli.go
@@ -34,9 +34,7 @@ func (cli *CLI) Binary(
 
 const (
 	// https://github.com/goreleaser/goreleaser/releases
-	goReleaserVersion = "v2.2.0"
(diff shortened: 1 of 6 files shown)


jedevc (approved):

vito (approved):

jedevc:
Bleh, test fails are legit, need to bump something else somewhere.

jedevc:
> frontend grpc server closed unexpectedly

in logs looks incredibly suspicious.

PR #8641, opened 2024-10-07
add Service.WithHostname

vito:
Allows you to manually configure a hostname for a service, which will be automatically namespaced within the module. This ability allows you to express circular service dependencies, which was previously impossible with content-addressed hostnames.

Implementation notes:

* New API: `Service { withHostname(hostname: String!): Service! }`
  * Sets a custom hostname to be used instead of the regular content-addressed (query ID -derived) hostname
* When a service has a custom hostname its FQDN will be `<host>.<module id>.<session id>.dagger.local`
  * The `<module id>` is a new component to ensure two modules setting the same hostname don't conflict
* A `withExec` always installs its calling `<module id>.<session id>.dagger.local` to its list of search domains
  * In other words, it will always be able to reach service hostnames run by the module that _installed_ the withExec
* A container service always installs its _starting_ `<module id>.<session id>.dagger.local` to its list of search domains
  * This allows a service returned by a module to reach services with custom hostnames running via the _using_ module
  * This is in addition to the search domain installed by the original `withExec`

### DIFF
--- a/.changes/unreleased/Added-20241008-125457.yaml
+++ b/.changes/unreleased/Added-20241008-125457.yaml
@@ -1,0 +1,13 @@
+kind: Added
+body: |
+  Services can now be given an explicit hostname via `Service.withHostname`.
+
+  Previously, you could only express a DAG of services, because each service hostname was derived from the service's configuration, and changing a service to point to another service would inherently change that service's configuration.
+
+  Now you can set your own hostnames ahead of time, tell your services about each other's hostnames, and start they manually.
+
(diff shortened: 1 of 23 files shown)


jedevc (approved):
Looks good to me :tada: (from my knowledge of the network stack here). We should probably follow up with some of these more complex patterns in our docs?

(failures look like just docs/sdk generation? engine/cli tests look good)

j-martin:
Great! This was the main blocker for us in adopting Dagger.

PR #8642, opened 2024-10-07
cli: internal tooling for graphviz output

sipsma:
This is a slightly cleaned up version of what I did at Dagger's recent team hackathon, differences being:
1. Code is *slightly* improved from hackathon-quality (just so it can be easier to build off of and not complete throwaway)
2. No exec metrics yet (like disk usage) since that PR is still pending merge
3. It's controlled via hidden top-level dagger CLI flags now (instead of inlined consts):
   * `--dot-output=<path>` will trigger writing a `.dot` formatted file to the given path
   * `--dot-focus-field=<gql field name>` gives some very crude filtering ability by resulting in only calls for the given field name and their descendents being shown
   * `--dot-show-internal` results in calls under internal spans showing up too (they are hidden by default).

E.g. to look at the DAG for running a set of integ tests you could run:
```
dagger --dot-output /home/sipsma/dag.dot call test specific --pkg=./core/integration --run=TestCall/TestSocketArg

dot -Tsvg -x -o dag.svg dag.dot # can open dag.svg in most web browsers
```

The above command requires `dot` to be installed, which is typically available in system packages named something like `graphviz`.

It should work on any command that has progress output, not just `dagger call`.

This is still very barebones, but just useful enough to be worth merging especially since it's controlled with hidden flags and otherwise unintrusive to our code (just reuses existing telemetry the cli already collects).

The idea is to merge the basic thing now and then use+improve it while debugging/bottleneck-hunting.

Some low-hanging fruit improvements possible as we iterate:
* Make the "highlight bottlenecks in red" more configurable (just colors a vertex red if it takes longer than a second right now). Can also make degree of highlighting proportional rather than off/on
* Make it work off OTEL data persisted somewhere (i.e. cloud, a json dump of all OTEL data received by the CLI, etc.) so that you can change filters/configuration without having to actually run `dagger call` again
* Make use of clustering in graphviz (by span?)
* Make use of other graphviz layout engines that might have nicer output for complex cases
* Add a dagger module for going from `.dot`->`.svg` (and more)

### DIFF
--- a/dagql/dagui/dot.go
+++ b/dagql/dagui/dot.go
@@ -1,0 +1,328 @@
+package dagui
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"time"
+
+	"github.com/dagger/dagger/dagql/call/callpbv1"
+)
+
+func (db *DB) WriteDot(
+	outputFilePath string,
+	focusField string,
+	showInternal bool,
+) {
+	if outputFilePath == "" {
+		return
+	}
+	out, err := os.Create(outputFilePath)
+	if err != nil {
+		panic(err)
+	}
+
+	dag := db.getDotDag(focusField, showInternal)
+	dag.writeTo(out)
+}
+
+type dotDag struct {
+	vtxByCallDgst map[string]*dotVtx
+}
+
+type dotVtx struct {
+	dgst string
+	call *callpbv1.Call
+	span *Span
+
+	children map[string]*dotEdge
+	parents  map[string]*dotEdge
+
+	// whether to include in dot output
+	show bool
+}
+
+type dotEdge struct {
+	kind edgeKind
+	// only for kind arg
+	argName string
+}
+
+func (db *DB) getDotDag(focusField string, showInternal bool) *dotDag {
+	dag := &dotDag{
+		vtxByCallDgst: make(map[string]*dotVtx),
+	}
+
+	for _, span := range db.Spans {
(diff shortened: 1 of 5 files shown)


dagql/dagui/dot.go
```diff
@@ -0,0 +1,329 @@
...
+		return fmt.Sprintf("%d", val.Int)
+	case *callpbv1.Literal_Float:
+		return fmt.Sprintf("%f", val.Float)
+	case *callpbv1.Literal_String_:
+		if len(val.String_) > 256 {
+			return "ETOOBIG"
+		}
+		return fmt.Sprintf("%q", val.String_)
+	case *callpbv1.Literal_CallDigest:
+		return "<input>"
```
vito, line 301: You could almost use the existing `Display() string` code from `dagql/call.Literal` but I guess you might want to avoid the value getting too big with nesting, hence `<input>`. Not a big deal to just roll your own here anyway, since the number of options shouldn't really ever change.

vito (approved):
lgtm!

dagql/dagui/dot.go
sipsma: Yeah, `<input>` also makes sense since there's an edge drawn for arg inputs that are from a different call, so no need to repeat. Agree copy-pasta is probably better here; also keeps this from having to touch any code that impacts user-facing stuff (would be sad to have a bug related to changes meant only for internal debug tooling)

Issue #8644, opened 2024-10-08
Hitting `TOOMANYREQUESTS` for `scan-engine` job


jedevc:
See https://github.com/dagger/dagger/actions/runs/11233423649/job/31227113107

```
Stdout:
invoke: input: container.from.withMountedDirectory.withMountedCache.withMountedDirectory.withExec.sync resolve: process "trivy fs --format=json --exit-code=1 --scanners=vuln --vuln-type=library --severity=CRITICAL,HIGH --show-suppressed /mnt/src" did not complete successfully: exit code: 1

Stderr:
2024-10-08T10:20:52Z	INFO	Need to update DB
2024-10-08T10:20:52Z	INFO	Downloading DB...	repository="ghcr.io/aquasecurity/trivy-db:2"
2024-10-08T10:20:52Z	FATAL	Fatal error	init error: DB error: failed to download vulnerability DB: database download error: oci download error: failed to fetch the layer: GET https://ghcr.io/v2/aquasecurity/trivy-db/blobs/sha256:95d05be1e5cf9431ddaa617d84af4d0526511a39984d7a709f4043339931da6a: TOOMANYREQUESTS: retry-after: 454.813µs, allowed: 44000/minute
Error: Process completed with exit code 2.
```

This is annoying, there's maybe a token somewhere we need to be passing into the trivy container? Filing this issue and moving on, but will circle back unless someone picks this up.

jedevc:
Still seeing this, even after #8645.

jedevc:
Ohhh. I think https://github.com/dagger/dagger/pull/8645#issuecomment-2399543690 is the right fix, but I think the `docker.json` we pass only has creds for docker hub - not for `ghcr.io`.

We need to do a `docker login ghcr.io` somewhere.

gerhard:
Could this be a per image limit? `44000/minute` is `733/s` which I seriously doubt that we are generating this ourselves. Checking if I can find something about this upstream.

gerhard:
These seem related:
- https://github.com/aquasecurity/trivy/issues/7598
- https://github.com/aquasecurity/trivy/issues/7659

gerhard:
`--db-repository public.ecr.aws/aquasecurity/trivy-db` seems to be the short-term fix that we want: https://github.com/aquasecurity/trivy/issues/7598#issuecomment-2401784140

PR #8645, opened 2024-10-08
ci: Bump & sha256 pin aquasec/trivy image

gerhard:
Fixes https://github.com/dagger/dagger/issues/8644

Expecting it to fix the following error:

    Fatal errorinit error: DB error: failed to download vulnerability DB: database download error: OCI repository error: 1 error
    GET https://ghcr.io/v2/aquasecurity/trivy-db/manifests/2: TOOMANYREQUESTS: retry-after: 566.363µs, allowed: 44000/minute

---

Follow-up to:
- https://github.com/dagger/dagger/pull/8643

### DIFF
--- a/.dagger/engine.go
+++ b/.dagger/engine.go
@@ -393,7 +393,7 @@ func (e *Engine) Scan(ctx context.Context) error {
 	}
 
 	ctr := dag.Container().
-		From("aquasec/trivy:0.53.0").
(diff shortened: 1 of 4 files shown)


jedevc:
Just pushed an additional commit to forward the `docker/config.json` into the trivy container (I think this is the root issue - though obv we should also be pinning the image).

LGTM :tada:

jedevc (approved):

PR #8647, opened 2024-10-08
chore: remove `withFocus`/`withoutFocus`

jedevc:
> [!WARNING]
>
> Only merge for v0.14.0!

I have no idea why we didn't remove these before - they've been deprecated since https://github.com/dagger/dagger/pull/6835. They should've gone when we got rid of `pipeline` as well, but looks like it got missed.

This is *technically* a breaking change, so maybe needs to wait till v0.14.0.


sipsma:
(rebasing on main since Justin is out, reviewing now)

sipsma (approved):
LGTM but waiting to press merge until 100% confirmation we are doing v0.14.0 next

PR #8438, opened 2024-09-12
ci: Stop using spot instances

gerhard:
> [!NOTE]
> Already includes changes from https://github.com/dagger/dagger/pull/8643

They are getting disrupted often, and that becomes annoying after a few weeks of turbulence.

Before we can merge this, need to introduce `dind-od` instances.

PR #8652, opened 2024-10-08
feat: support tmpfs size in withMountedTemp

### LINKED ISSUE #8573: Support `tmpfs` attributes in `withMountedTemp`
Right now `withMountedTemp` will get you a `tmpfs` mounted at the specified location, but there's no control over important attributes like the size.

We should expose at minimum an option for the size of the tmpfs mount, which should be easy since that's supported in LLB.

There are [more options available on tmpfs mounts](https://man7.org/linux/man-pages/man5/tmpfs.5.html) that we could consider, but size is the highest priority right now.

cwlbraa:
fixes #8573, at least partially

this pr adds a size parameter to withMountedTemp and passes that down to buildkit's mount initialization.

I had many questions that @sipsma answered in the comments.
1. I don't particularly like the structure of the test, adding a whole additional test that's 99% repeated code, but it seems hard to parameterize around optional params; putting 2 WithMountedTemps in parallel creates a marshaling-into-struct ambiguity that's unresolvable.... should I be chaining or something? 
2. Is the "size" parameter in the appropriate place in the graphql schema? it can only be supplied for tmpfses, but it lives on the same level. do we want errors if folks try to supply sizes for other mounts? where would one add that logic?
3. corollary question, should we be adding a more open-to-extension TmpfsOptions sort of struct here?
4. In schema/container.go, do I need to expand env vars for size the same way it's done for paths?
5. there's a TODO regarding handling of optionals in dagql -- can I just blindly traverse these structs without worrying about nil deferences? it kinda seems like it from the test results.
6. do the SDKs just work themselves out from API level changes? or am I missing some steps here?
7. I'm sure I have unknown-unknowns, plz look for them :)

### DIFF
--- a/core/schema/container.go
+++ b/core/schema/container.go
@@ -221,6 +221,7 @@ func (s *containerSchema) Install() {
 		dagql.Func("withMountedTemp", s.withMountedTemp).
 			Doc(`Retrieves this container plus a temporary directory mounted at the given path. Any writes will be ephemeral to a single withExec call; they will not be persisted to subsequent withExecs.`).
 			ArgDoc("path", `Location of the temporary directory (e.g., "/tmp/temp_dir").`).
+			ArgDoc("size", `Size of the temporary directory in bytes.`).
 			ArgDoc("expand",
 				"Replace ${VAR} or $VAR in the value of path according to the current "+
 					`environment variables defined in the container (e.g. "/$VAR/foo").`),
@@ -1308,6 +1309,7 @@ func (s *containerSchema) withMountedCache(ctx context.Context, parent *core.Con
 
 type containerWithMountedTempArgs struct {
 	Path   string
+	Size   dagql.Optional[dagql.Int]
 	Expand bool `default:"false"`
 }
 
@@ -1317,7 +1319,8 @@ func (s *containerSchema) withMountedTemp(ctx context.Context, parent *core.Cont
 		return nil, err
 	}
 
-	return parent.WithMountedTemp(ctx, path)
+	// TODO: is the optional safe to dereference?
+	return parent.WithMountedTemp(ctx, path, args.Size.Value.Int())
 }
 
 type containerWithoutMountArgs struct {
--- a/core/schema/container.go
+++ b/core/schema/container.go
@@ -1319,7 +1319,6 @@ func (s *containerSchema) withMountedTemp(ctx context.Context, parent *core.Cont
 		return nil, err
 	}
 
-	// TODO: is the optional safe to dereference?
 	return parent.WithMountedTemp(ctx, path, args.Size.Value.Int())
 }
 
--- a/.changes/unreleased/Added-20241008-154901.yaml
+++ b/.changes/unreleased/Added-20241008-154901.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: enabled withMountedTemp size configuration
+time: 2024-10-08T15:49:01.009581-07:00
+custom:
+    Author: cwlbraa
+    PR: "8652"
--- a/core/container.go
+++ b/core/container.go
@@ -245,6 +245,9 @@ type ContainerMount struct {
 	// Configure the mount as a tmpfs.
 	Tmpfs bool `json:"tmpfs,omitempty"`
 
+	// Configure the size of the mounted tmpfs in bytes
+	Size int `json:"size,omitempty"`
+
 	// Configure the mount as read-only.
 	Readonly bool `json:"readonly,omitempty"`
 }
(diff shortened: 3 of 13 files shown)


core/schema/container.go
```diff
@@ -1317,7 +1319,8 @@ func (s *containerSchema) withMountedTemp(ctx context.Context, parent *core.Cont
 		return nil, err
 	}
 
-	return parent.WithMountedTemp(ctx, path)
+	// TODO: is the optional safe to dereference?
+	return parent.WithMountedTemp(ctx, path, args.Size.Value.Int())
```
sipsma, line 1323: I believe not checking it means that when it wasn't set it will default to the zero value, which in the case of an int is literally zero; so provided that `WithMountedTemp` understands `0` as "unset" I think this should be okay. It looks like LLB also treats 0 as unset so think this works out 👍

Just for future reference, if you do want to check if the optional was set you can use `.Valid` before calling `.Value`, e.g. as done here https://github.com/sipsma/dagger/blob/9cf2e8bfdb443223bae84329cd157b3ea7b7e061/core/schema/container.go#L639-L644

sipsma (commented):
> I don't particularly like the structure of the test, adding a whole additional test that's 99% repeated code, but it seems hard to parameterize around optional params; putting 2 WithMountedTemps in parallel creates a marshaling-into-struct ambiguity that's unresolvable.... should I be chaining or something?

The test LGTM as is, but if you want to de-dupe it a bit more you could:
* Switch away from the "raw gql query" approach, which only exists because a lot of these tests were added in ancient times before the Go SDK was fully formed. If you look later in `container_test.go` you'll see them switch to using the Go SDK (e.g. https://github.com/sipsma/dagger/blob/9cf2e8bfdb443223bae84329cd157b3ea7b7e061/core/integration/container_test.go#L3740-L3740)
* If you do that, you can split out into subtests but have them share some common code, or you could do a table driven test approach (e.g. [here](https://github.com/sipsma/dagger/blob/9cf2e8bfdb443223bae84329cd157b3ea7b7e061/core/integration/container_test.go#L3980-L3980)), etc.

> Is the "size" parameter in the appropriate place in the graphql schema? it can only be supplied for tmpfses, but it lives on the same level. do we want errors if folks try to supply sizes for other mounts? where would one add that logic?

The change here looks like it only adds `size` to the `withMountedTemp` API, so it should be scoped just to tmpfs mounts, which is what we want I think.

If you were referring to the new `Size` field in the `ContainerMount` struct in `core/container.go`, that's actually just internal only and doesn't get into the user-facing gql schema. I don't have a problem with Size existing there even if it's only used for when `Tmpfs==true` at the moment.

> corollary question, should we be adding a more open-to-extension TmpfsOptions sort of struct here?

You *could* but it just doesn't make a big difference right now IMO; if I was to leave that comment it would be an uber-nit 🙂

> In schema/container.go, do I need to expand env vars for size the same way it's done for paths?

Wouldn't consider that a blocker at all, I think the use case of configuring the size of a tmpfs mount from an env var in a container is extremely obscure. No objections if you do add it, but I'd personally just leave that out for simplicity for now.

> there's a TODO regarding handling of optionals in dagql -- can I just blindly traverse these structs without worrying about nil deferences? it kinda seems like it from the test results.

Left a comment about this one

> do the SDKs just work themselves out from API level changes? or am I missing some steps here?

Ah no, we should add this to `CONTRIBUTING.md`. You need to generate them all by running `dagger call sdk all generate export --path .` locally from your repo and committing the resulting changes.

> I'm sure I have unknown-unknowns, plz look for them :)

This is worth a changelog entry (I added the needs changelog label so our CI guards that one is added). Based on parallel discord conversations I think you saw that but let me know if more pointers needed

core/schema/container.go
cwlbraa: I had noticed the Valid field in my autocomplete, it's what made me worry I was being unsafe by not checking it :)

sipsma (approved):
LGTM! I forgot that in addition to generating the SDKs you also need to regenerate the docs (specifically some generate graphql schema files that get consumed by our docs website), which is currently a separate command unfortunately:
```
dagger call docs generate export --path .
```
(we should probably just make a function for generating everything, sdks+docs included, but separate issue)

That's the cause of the Docs failure in CI.

The rest of the SDK failures are from an annoying infra issue that should be fixed by https://github.com/dagger/dagger/pull/8438, they pass on a re-run

jedevc:
Regenerated the docs, in the interest of getting this in for the release :tada:

LGTM!

jedevc (approved):

Issue #8513, opened 2024-09-20
✨ Dagger should use Entrypoint without any additional init process


conorpp:
### What are you trying to do?

We ship containers based on the [s6 init overlay](https://github.com/just-containers/s6-overlay), not the common default `dumb-init`.  We want to use dagger to test this container, but it requires the containers that dagger spawns to be allowed to use a different init system.

Dagger does currently allow jobs to specify a custom entrypoint, so I thought this would work already.  But dagger is still prepending it's own `/.init` process to it, which I find unexpected as that's not how entrypoint in docker works.

```go
func (m *Treasury) InitTest(
    ctx context.Context,
) (*dagger.Container, error) {
    container := dag.Container().
        From("alpine:latest").
        WithEntrypoint([]string{"/bin/ps", "axu"}).
        WithExec([]string{}, dagger.ContainerWithExecOpts{UseEntrypoint: true})

    return container, nil
}
```

https://dagger.cloud/conor/traces/fea359d73ad18d9f38c85703789f297e
https://dagger.cloud/conor/traces/fea359d73ad18d9f38c85703789f297e?span=28e5281641b0c959

I would expect `ps` to have PID 1, but instead `/.init` is still run and it gets PID 1. 

Whereas running with docker, we see that `ps` would get PID 1.

```bash
docker run --entrypoint "ps" alpine:latest aux
```
```
PID   USER     TIME  COMMAND
    1 root      0:00 ps aux
```

### Why is this important to you?

This is important to use as our application relies on s6 init, and running without it would greatly reduce the amount of functionality we could test using dagger.

sipsma:
We configure this automatic init because it results in more expected expected behavior in the common case (there's various weird quirks of pid 1 to the point I've seen e.g. several docker maintainers express they wished they had defaulted to using an init process).

However there's plenty of valid use cases for user control of pid 1 even if they aren't the most common case, so we can pretty easily add an option to `withExec` to disable our automatic init and hand over control to the user, exec'ing their process directly.

PR #8656, opened 2024-10-09
core: support disabling automatic init in withExec

sipsma:
Dagger automatically injects an init process in withExecs in order to avoid surprising behavior around processes being PID 1. I have generally gathered that this is the right default behavior, to the point that I've seen some docker maintainers say that they wish docker had defaulted to this same behavior.

However, there are less common use cases that do call for user processes being PID 1, such as when they want very custom init behavior or are testing init systems (i.e. testing systemd in a container).

For a long time this was impossible because of the shim but after we moved to the executor, which operates entirely in the engine process, it became technically optional to have an init process injected into the container (even if the right default).

This change adds an option to `withExec` for disabling the automatic init process, `noInit` so we can support those use cases. The default behavior is left alone.

---

Fixes https://github.com/dagger/dagger/issues/8513

### DIFF
--- a/.changes/unreleased/Added-20241008-171908.yaml
+++ b/.changes/unreleased/Added-20241008-171908.yaml
@@ -1,0 +1,9 @@
+kind: Added
+body: |-
+  Add `noInit` option to `withExec` to support disabling automatic init process.
+
+  Use cases that strictly require the user exec is PID 1 of the container are now supported by setting `noInit` to true.
+time: 2024-10-08T17:19:08.364837213-07:00
+custom:
+  Author: sipsma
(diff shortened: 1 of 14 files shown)


sipsma:
php checks failing for me because when I run codegen locally I get a bunch of weird formatting changes that don't happen in CI (????).

Even when I run php codegen on main locally I get the same formatting changes: https://github.com/sipsma/dagger/commit/7f15a0020fc89ff912a56af3ed4b01580492f02b

Gonna resort to hand-fixing with `git add -i` locally for now 🤷‍♂️

jedevc (approved):

PR #8665, opened 2024-10-09
dev: add blurb for necessary steps following GraphQL schema changes

cwlbraa:
steps I missed in #8652, bundled all together.

added a top level dev function "generate" intended to all code generation in one go. 

as an aside: writing this with the heading "GraphQL schema changes" immediately gave me the urge to write the typical warnings regarding backwards-incompatible changes to graphql schemas that you've got to drumbeat in a more typical SaaS graphql deployment (nullables can't be made non-nullable, renames must be duplicated and deprecated, etc) but it occurs to me that maybe our release structure is less sensitive to backwards incompatibility? how concerned are we with breaking clients via engine api changes?

### DIFF
--- a/.dagger/main.go
+++ b/.dagger/main.go
@@ -164,6 +164,20 @@ func (dev *DaggerDev) Test() *Test {
 	return &Test{Dagger: dev}
 }
 
+// Run all code generation - SDKs, docs, etc
+func (dev *DaggerDev) Generate(ctx context.Context) (*dagger.Directory, error) {
+	docs, error := dev.Docs().Generate(ctx)
+	if error != nil {
+		return nil, error
+	}
+	sdks, error := dev.SDK().All().Generate(ctx)
+	if error != nil {
+		return nil, error
+	}
+
+	return docs.WithDirectory(".", sdks), nil
+}
+
 // Develop Dagger SDKs
 func (dev *DaggerDev) SDK() *SDK {
 	return &SDK{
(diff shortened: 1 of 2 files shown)


.dagger/main.go
```diff
@@ -201,6 +201,19 @@ func (dev *DaggerDev) Test() *Test {
 	return &Test{Dagger: dev}
 }
 
+func (dev *DaggerDev) Generate(ctx context.Context) (*dagger.Directory, error) {
```
cwlbraa, line 204: open question: do we want this to do `dagger call engine generate` as well? under what circumstances does one need to go generate in this codebase?

jedevc: Yeaaah, probably - essentially `engine generate` is just doing `go generate` for all the engine files, which is a lot of `protoc` and `sqlc` things.

Up to you whether to add it here - it doesn't need updating on every api change.

.dagger/main.go
```diff
@@ -201,6 +201,20 @@ func (dev *DaggerDev) Test() *Test {
...
+// Run all code generation - SDKs, docs, etc
+func (dev *DaggerDev) Generate(ctx context.Context) (*dagger.Directory, error) {
+	docs, error := dev.Docs().Generate(ctx)
+	if error != nil {
+		return nil, error
+	}
+	sdks, error := dev.SDK().All().Generate(ctx)
+	if error != nil {
+		return nil, error
+	}
```
jedevc, line 213: We can probably parallelize these in an `errgroup`.

.dagger/main.go
cwlbraa: added

sipsma:
This LGTM but need to do a rebase

jedevc (approved):

jedevc (approved):
Feel free to merge @cwlbraa :tada:

PR #8669, opened 2024-10-10
chore(go): deprecated unscoped enum values

jedevc:
In go, enum values are top-level to the package (unlike in typescript/python/etc where they are members of the type itself). Because of this, it's very easily possible to clash - two different enum types might have the same value.

To prevent this clash, we should scope the enum value with the name of the enum it's from - this is pretty standard go practice, and should be familiar.

We keep the old consts around, but alias them to the new consts, and deprecate them (to remove in v0.14.0).

### DIFF
--- a/cmd/codegen/generator/go/templates/module_funcs.go
+++ b/cmd/codegen/generator/go/templates/module_funcs.go
@@ -15,7 +15,7 @@ import (
 const errorTypeName = "error"
 
 var voidDef = Qual("dag", "TypeDef").Call().
-	Dot("WithKind").Call(Id("dagger").Dot("VoidKind")).
+	Dot("WithKind").Call(Id("dagger").Dot("TypeDefKindVoidKind")).
 	Dot("WithOptional").Call(Lit(true))
 
 func (ps *parseState) parseGoFunc(parentType *types.Named, fn *types.Func) (*funcTypeSpec, error) {
--- a/cmd/codegen/generator/go/templates/src/_types/enum.go.tmpl
+++ b/cmd/codegen/generator/go/templates/src/_types/enum.go.tmpl
@@ -9,8 +9,8 @@ func ({{ $enumName }}) IsEnum() {}
 {{- $needsScopedEnums := CheckVersionCompatibility "v0.14.0" | not }}
 const (
 	{{- range $index, $field := .EnumValues | SortEnumFields }}
-	{{- $fieldName := ($field.Name | FormatEnum) }}
-	{{- $fullFieldName := print $enumName $fieldName }}
+	{{- $fieldName := ($field.Name | FormatEnum "") }}
+	{{- $fullFieldName := print ($field.Name | FormatEnum $enumName) }}
 	{{ $field.Description | Comment }}
 	{{ $fullFieldName }} {{ $enumName }} = "{{ $field.Name }}"
 	{{- with .Directives.SourceMap -}} // {{ .Module }} ({{ .Filelink | ModuleRelPath }}) {{- end }}
--- a/cmd/codegen/generator/go/templates/src/_types/enum.go.tmpl
+++ b/cmd/codegen/generator/go/templates/src/_types/enum.go.tmpl
@@ -6,11 +6,20 @@ type {{ $enumName }} string
 
 func ({{ $enumName }}) IsEnum() {}
 
+{{- $needsScopedEnums := CheckVersionCompatibility "v0.14.0" | not }}
 const (
 	{{- range $index, $field := .EnumValues | SortEnumFields }}
+	{{- $fieldName := ($field.Name | FormatEnum) }}
+	{{- $fullFieldName := print $enumName $fieldName }}
 	{{ $field.Description | Comment }}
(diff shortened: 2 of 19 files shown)


cmd/codegen/generator/go/templates/module_funcs.go
```diff
@@ -14,7 +14,7 @@ import (
 const errorTypeName = "error"
 
 var voidDef = Qual("dag", "TypeDef").Call().
-	Dot("WithKind").Call(Id("dagger").Dot("VoidKind")).
+	Dot("WithKind").Call(Id("dagger").Dot("TypeDefKindVoidKind")).
```
jedevc, line 17: So this is annoying - the name now stutters for some of our core types (but I guess this has always been the case in typescript/python).

Ideally we should also have the value here just be `VOID` instead of `VOID_KIND`? But this would be a pretty big backwards incompat change.

We could do this with the rest of the enum changes we're planning?

sdk/go/dagger.gen.go
```diff
@@ -8115,57 +8115,99 @@ func (CacheSharingMode) IsEnum() {}
...
 type ImageMediaTypes string
 
 func (ImageMediaTypes) IsEnum() {}
 
 const (
-	Dockermediatypes ImageMediaTypes = "DockerMediaTypes"
+	ImageMediaTypesDockermediatypes ImageMediaTypes = "DockerMediaTypes"
+
+	// Deprecated: use ImageMediaTypesDockermediatypes instead
+	Dockermediatypes ImageMediaTypes = ImageMediaTypesDockermediatypes
```
helderco, line 8173: Why isn't this `DockerMediaTypes` and `ImageMediaTypesDockerMediaTypes`? Can you take the opportunity to fix the casing in the new scoped one? I've seen other instances of this. `ImageMediaTypesOcimediatypes` could be `ImageMediaTypesOCIMediaTypes`.

sdk/go/dagger.gen.go
```diff
@@ -8115,57 +8115,99 @@ func (CacheSharingMode) IsEnum() {}
...
+
+	ImageMediaTypesOcimediatypes ImageMediaTypes = "OCIMediaTypes"
```
helderco, line 8175: Oh, I just noticed these values aren't in the right casing in the API:

```diff
"""Compression algorithm to use for image layers."""
enum ImageLayerCompression {
-  Gzip
-  Zstd
-  EStarGZ
-  Uncompressed
+  GZIP
+  ZSTD
+  E_STAR_GZ
+  UNCOMPRESSED
}

"""Mediatypes to use in published or exported image metadata."""
enum ImageMediaTypes {
-  OCIMediaTypes
-  DockerMediaTypes
+  OCI_MEDIA_TYPES
+  DOCKER_MEDIA_TYPES
}
```

I wonder if that was to preserve the capitalization in <code>EStar<strong>GZ</strong></code> and <code><strong>OCI</strong>MediaTypes</code>, but if it's producing `Estargz` and `Ocimediatypes` it's not working out. It would at least be `EStarGz` and `OciMediaTypes`, unless we pass it through the acronym thing, which is needed in a lot of places already.

helderco (approved):

cmd/codegen/generator/go/templates/src/_types/enum.go.tmpl
```diff
@@ -4,10 +4,18 @@ type {{ $enumName }} string
 
 func ({{ $enumName }}) IsEnum() {}
 
+{{- $needsScopedEnums := CheckVersionCompatibility "v0.14.0" | not }}
 const (
 	{{- range $index, $field := .EnumValues | SortEnumFields }}
+	{{- $fieldName := ($field.Name | FormatEnum) }}
+	{{- $fullFieldName := print $enumName $fieldName }}
```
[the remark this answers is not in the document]
helderco: You mean for user modules that may already be scoping right?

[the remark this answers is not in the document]
jedevc: Yup, exactly.

sdk/go/dagger.gen.go
helderco: Yeah, but I think the opposite :) The SCREAMING_SNAKE_CASE is a convention in GraphQL, so I would expect this:

```graphql
"""Mediatypes to use in published or exported image metadata."""
enum ImageMediaTypes {
  OCI_MEDIA_TYPES @value("OCIMediaTypes")
  DOCKER_MEDIA_TYPES @value("DockerMediaTypes")
}
```

sdk/go/dagger.gen.go
jedevc: Pushed a commit to resolve this - I've completely changed this implementation for the new enum names, now we actually try and avoid explicitly removing all the casing.

All the remaining case issues are an artifact of #8122.

cmd/codegen/generator/go/templates/src/_types/enum.go.tmpl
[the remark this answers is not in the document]
helderco: Agreed 👍

Issue #8146, opened 2024-08-13
✨ Typescript sdk should support overriding base image (like the python sdk does)


cchanley2003:
### What are you trying to do?

Of the 3 main sdks the typescript one doesn't allow for overriding the base image. Go doesn't need it and python has the [discovery](https://github.com/dagger/dagger/blob/62e369bc8ba0a29061eeb474fee01055f2babe44/sdk/python/runtime/discovery.go#L320). Typescript should support similar functionality to the python sdk.

Similar to the python we could put the  image override in the config section of the package.json

### Why is this important to you?

1. I want to be able to provide my own base image
2. For corporate envs may need to pull from a private registry

### How are you currently working around this?

Only work around atm is to fork the typescript sdk

TomChv:
It's possible to override the current node version, do you want to actually change the image itself?

https://docs.dagger.io/manuals/developer/typescript/356352/runtime#nodejs

If so, I can add that capabilities

cchanley2003:
Yes, change the image not just the version or package manager.

TomChv:
> Yes, change the image not just the version or package manager.

Okay I can make that happen, I'll do a change on the runtime :D

PR #8675, opened 2024-10-10
feat: ts sdk custom base image

TomChv:
Fixes https://github.com/dagger/dagger/issues/8146


core/integration/module_typescript_test.go
```diff
@@ -704,6 +704,42 @@ func (TypescriptSuite) TestRuntimeDetection(ctx context.Context, t *testctx.T) {
...
+func (TypescriptSuite) TestCustomBaseImage(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	t.Run("should use custom base image if different than node or bun", func(ctx context.Context, t *testctx.T) {
+		modGen := c.Container().From(golangImage).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work").
+			WithNewFile("package.json", `{
+			"dagger": {
+				"runtime": "vasektechnology/node-alpine:test@sha256:ce757edbc0509ede4ef9d2d9e1b329dc43762fa55ccf55be00b51fe2da0fa473"
```
helderco, line 716: I think it's better to use `baseImage` instead of `runtime` here.

Can you use a different `node-alpine` version that's not the default one to check if the setting is applied, rather than using an entirely custom published image?

That's how it's tested in Python, and it's enough to check whether the setting is applied or not:

https://github.com/dagger/dagger/blob/8321d35cd20b1c5cb998074a19b6396a76c62259/core/integration/module_python_test.go#L492

core/integration/module_typescript_test.go
```diff
@@ -704,6 +704,42 @@ func (TypescriptSuite) TestRuntimeDetection(ctx context.Context, t *testctx.T) {
 	})
 }
 
+func (TypescriptSuite) TestCustomBaseImage(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
```
helderco, line 708: Move connect into each sub test to avoid issues with parallelism and flakes.

core/integration/module_typescript_test.go
TomChv: > Can you use a different node-alpine version that's not the default one to check if the setting is applied, rather than using an entirely custom published image?

We already support different node-alpine version using runtime: [docs.dagger.io/manuals/developer/typescript/356352/runtime#nodejs](https://docs.dagger.io/manuals/developer/typescript/356352/runtime#nodejs)

What was not supported was using a completely different base image as runtime, which was something asked by the user on this issue: https://github.com/dagger/dagger/issues/8146

> I think it's better to use baseImage instead of runtime here.

Wouldn't it be confusing because users could think that can fill both when actually it's one or the other?

helderco: > What was not supported was using a completely different base image as runtime, which was something asked by the user on this issue: #8146

It just needs to be different from what gets produced as the default, like a slightly older patch version, variant, or just digest. No reason to rely on `vasektechnology/`.

See this comment: 

https://github.com/dagger/dagger/blob/8321d35cd20b1c5cb998074a19b6396a76c62259/core/integration/module_python_test.go#L387-L388

sdk/typescript/runtime/main.go
```diff
@@ -400,26 +406,34 @@ func (t *TypescriptSdk) detectRuntime(ctx context.Context) (SupportedTSRuntime,
...
 		}
 	}
 
 	// Try to detect runtime from lock files
 	if t.moduleConfig.hasFile("bun.lockb") {
-		return Bun, "", nil
+		t.moduleConfig.runtime = Bun
+
+		return nil
 	}
```
helderco, line 426: What I'm seeing is that if the user provides an image ref with a digest to `dagger.runtime` but has a `bun.lockb`, the final image will be:

```
oven/bun:sha256:xxxxxxxxxxxxxxxxx-alpine
```

core/integration/module_typescript_test.go
helderco: > > I think it's better to use baseImage instead of runtime here.
> 
> Wouldn't it be confusing because users could think that can fill both when actually it's one or the other?

Why not allow both? If an image is used in `runtime`, you resort to auto-detecting for lockfiles to figure out the runtime. Isn't there a use case for explicitly specifying what the runtime is? It's not just that the image has the user's expected executable, it's that the runtime module makes decisions based on knowing what the runtime is.

I would treat them as two separate things where the default for `baseImage` is extracted from `runtime`. But if both are provided, that just means don't depend on auto-detection for runtime, just use what's given. 

It's simpler to reason with that way and because of that, it also makes the logic cleaner and less prone to bugs. Right now this only works for Node, the version happens to match the `@` split, but is still used in Bun incorrectly.

TomChv: > It just needs to be different from what gets produced as the default, like a slightly older patch version, variant, or just digest. No reason to rely on vasektechnology/.

Then this is already supported.

> I would treat them as two separate things where the default for baseImage is extracted from runtime. But if both are provided, that just means don't depend on auto-detection for runtime, just use what's given.

I see, that make sense

So I shall just extends the usage of `runtime` to support any string and not just `bun/node`

sdk/typescript/runtime/main.go
```diff
@@ -400,26 +406,34 @@ func (t *TypescriptSdk) detectRuntime(ctx context.Context) (SupportedTSRuntime,
...
 			case Bun, Node:
-				return runtime, version, nil
+				t.moduleConfig.runtime = runtime
+				t.moduleConfig.runtimeVersion = version
+
+				return nil
+			// Custom base image different from node or bun
 			default:
-				return "", "", fmt.Errorf("detected unknown runtime: %s", runtime)
+				t.moduleConfig.runtime = Node
```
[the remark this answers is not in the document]
TomChv: It supported for `node` & `bun`

sdk/typescript/runtime/main.go
TomChv: No, because I return before in case there's an explicit config from the user

https://github.com/TomChv/dagger/blob/feat/custom-ts-sdk-runtime-base-image/sdk/typescript/runtime/main.go#L412

sdk/typescript/runtime/main.go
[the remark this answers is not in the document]
helderco: No, because only Node has this:

```go
if image != "" {
	ctr = ctr.From(image)
}
```

Bun doesn't:

https://github.com/TomChv/dagger/blob/e56c3629bd9e716f77e1981f86d336e54f59064c/sdk/typescript/runtime/main.go#L253-L258

core/integration/module_typescript_test.go
helderco: > > It just needs to be different from what gets produced as the default, like a slightly older patch version, variant, or just digest. No reason to rely on vasektechnology/.
> 
> Then this is already supported.

I'm talking about the test case, not the feature. What I mean is that the test case just needs to use a different image from the default so we can see that the setting is being applied correctly. It doesn't have to be a custom image from `vasektechnology/`. 

> So I shall just extends the usage of `runtime` to support any string and not just `bun/node`

That's not what I'm saying, that's what you're actually doing the PR though. I'm saying the opposite. Keep `runtime` to just `bun/node`, and use another `dagger.baseImage` config for the base image. If that's empty, create it from either `dagger.runtime` (if set) or use the default image, like on main.

helderco (approved):
lgtm

PR #8678, opened 2024-10-10
ci: use on-demand instances for lint+test-publish

sipsma:
The spot-instance version of these nodes are having trouble existing right now, we want to move off spot instances anyways so just updating these right now to unblock these jobs hopefully.


grouville (approved):
✅

PR #8673, opened 2024-10-10
ci: automatically publish release notes for sdks

jedevc:
We couldn't do this before because we needed https://github.com/sagikazarmark/daggerverse/pull/183.

Now we can mark SDK release notes as *never* the latest, so we can just automatically publish these without going through draft.

PR #8682, opened 2024-10-11
fix: allow decoding enums that overlap with bool types

jedevc:
Fixes https://github.com/dagger/dagger/pull/8673#issuecomment-2407223417

`false` and `true` values are always decoded by gqlparser as booleans - however, this might not always be the case: a standalone value *might* potentially be decodable as an enum instead.
    
The solution for this is to just add another case for decoding booleans in enums.
    
> [!NOTE]
>
> There's really no way to avoid this hack from what I can tell - you can put a `false`/`true` value as an enum value, but when you come to parse it, the parse tree you *should* construct is dependent on the types that you declare inside the parse tree :cry: Not actually sure if there exists a general class of parser that can correctly parse these :laughing:

### DIFF
--- a/.changes/unreleased/Fixed-20241014-161755.yaml
+++ b/.changes/unreleased/Fixed-20241014-161755.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: |
+    Allow custom enums that include ambiguous names (such as "true"/"false")
+time: 2024-10-14T16:17:55.998067541+01:00
+custom:
+    Author: jedevc
+    PR: "8682"
--- a/core/enum.go
+++ b/core/enum.go
@@ -47,7 +47,7 @@ func (m *ModuleEnumType) ConvertFromSDKResult(ctx context.Context, value any) (d
 
 		val, err := decoder.DecodeInput(value)
 		if err != nil {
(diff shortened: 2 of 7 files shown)


sipsma (approved):
The diff here seems fine to me and makes general sense. The convolution of the different layering of types+value parsing is something I always have to re-grok everytime I look through it (despite having written large chunks of code in this area), so tbh I can't be 100% certain every case is handled, but good enough to approve. Worst case we just keep iterating and adding tests 🙂

One general comment: I guess I would have expected that enum values are all sent as strings and thus an enum value would be sent as `"true"` rather than `true` (i.e. string instead of json literal) and would have saved us here, but reality must be disconnected with that somewhere. Just curious if you know where and/or have any feelings on if changing the code to follow that pattern would be a viable route to fix this (feels more robust to me overall).

jedevc:
Refactored this patch slightly, and now instead of trying to force all of the magical parsing logic everywhere to consider this, we can just handle it directly in dagql's `ParseField` and perform a clever conversion there.

> One general comment: I guess I would have expected that enum values are all sent as strings and thus an enum value would be sent as `"true"` rather than `true` (i.e. string instead of json literal)

I would have expected this too - this is sadly not how graphql enums work though - they're unquoted.

However, we could switch our clients to sending string values? Dagql just does the right thing in these cases - but I think this might require some changes in all our querybuilders.

PR #8687, opened 2024-10-11
Fix macos docker install in CI

sipsma:
Still getting problems starting colima: https://github.com/dagger/dagger/actions/runs/11298598079/job/31428172736#step:3:192

Based on various issues around github, getting this to work is an utter shitshow, so we're not alone.

Found this relevant issue: https://github.com/lima-vm/lima/issues/1742

Which lead to a linked issue that showed a successful setup of installing docker, trying it out here.

### DIFF
--- a/.github/workflows/engine-and-cli-publish.yml
+++ b/.github/workflows/engine-and-cli-publish.yml
@@ -95,7 +95,7 @@ jobs:
       always() &&
       github.repository == 'dagger/dagger' &&
       (needs.publish.result == 'success' || needs.publish.result == 'skipped')
-    runs-on: macos-15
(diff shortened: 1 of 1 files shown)


sipsma:
Apparently github is having issues atm w/ actions, so probably not a good time to work on this. Will pick back up again later

(cc @jedevc, seems like we need some more iteration here)

sipsma:
MacOS doesn't run in PRs but ran successfully in a temp workflow here, so think this should work on main now 🤞: https://github.com/dagger/dagger/actions/runs/11352050252/job/31573962144?pr=8687

gerhard (approved):
👍 🚀

PR #8690, opened 2024-10-11
fix: SSH flake on integration test suite

grouville:
Follow-up of https://github.com/dagger/dagger/issues/8385#issuecomment-2406128224

Attempting the same fix to all the tests: a few were missing + unifying how we cleanup the cancel()


jedevc (dismissed):

grouville:
Finally got to the root of the flake: 
1. the `base` helper was shared across several tests and was used inside the `testOnMultipleVCS`, which was used on the next test too, leading to an error every single time
2. the cleanup mechanism of the socket should be relying on `defer cleanup()` instead of relying on the test Cleanup, state conflict too with parallel executions

jedevc (approved):

PR #8691, opened 2024-10-12
fix: use latest go release version 1.23.2 in .dagger

grouville:
During the CI / release improvements, go version in .dagger/go.mod was bumped to 1.23 which has no release for ARM, leading to this error:

.dagger git:(fix-go-version-daggerci) go env GOPROXY go: downloading go1.23 (linux/arm64)
go: download go1.23 for linux/arm64: toolchain not available

As per https://go.dev/dl/, the latest official version is 1.23.2

### DIFF
--- a/.dagger/go.mod
+++ b/.dagger/go.mod
@@ -1,6 +1,6 @@
 module github.com/dagger/dagger/.dagger
 
-go 1.23
+go 1.23.2
 
 require github.com/dagger/dagger/engine/distconsts v0.13.5
 


sipsma:
What command were you running that hit the `toolchain not available` error? I've been running today on top of that change without any issues, so just curious how this gets triggered

jedevc (approved):

jedevc:
Ahhh yeah, sorry.

This is possible if the version of go used is too low I think? That triggers the download attempt of the correct toolchain, which then fails because you need the patch release here.

I think this might be possible on older versions of dagger? Or an older local version of go?

Can't quite remember, but I have seen this before - the syntax for this file continually trips me up.

grouville:
> 
What command were you running that hit the toolchain not available error? I've been running today on top of that change without any issues, so just curious how this gets triggered

any go command from inside the `.dagger` directory

Issue #8535, opened 2024-09-22
Python: Allow module package names other than `main`


helderco:
## Summary

Generate  a `src/my_module` package for new modules instead of the hardcoded `src/main` package.

## Motivation

The entrypoint for Python’s runtime container imports a module’s source code via an hardcoded `import main`:

https://github.com/dagger/dagger/blob/5ec03a604d50b721aaa65ba7c4db8c875696ac72/sdk/python/src/dagger/mod/cli.py#L44-L48

This forces the module’s layout to always be under `src/main` (unless otherwise configured for the build backend), so every module looks the same and forces extra configuration to customize something that should be as natural as what `uv init` or `poetry init` produces.

It’s also been a source of confusion for some users in how to properly break a bigger module into multiple files because they may not understand whether `main` refers to the package or the module, in a `src/main/main.py` file.

For example, in order to move the *main object* from `__init__.py` to a `main.py` file, some users try to import in `__init__.py` with:

```python
import main
```

However, in a *src layout*, imports start from `src`, which means that the above file is actually importing itself! 

It would have to be either:
```python
import main.main  # absolute import: src/main/main.py
```

Or:
```python
import .main  # relative import: ./main.py
```

The confusion over this double `main` has prevented me from changing the default template to put the source code under a `src/main/main.py` instead of `src/main/__init__.py`:
- https://github.com/dagger/dagger/issues/7430#issuecomment-2123157460

Note that the file `main.py` is aptly named because every Dagger module has a *main object*, so that’s what `main` refers to, in a multi-file package.

## Specification

For a new Dagger module, generate a Python project that follows conventions more closely, perhaps actually delegating the project creation to `uv init --lib`:
- Python project name following the Dagger module’s name, in *kebab-case*
- Python package name for the source code following the same name, in *snake_case*

Additionally:
- Example *main object* code under a `main.py` module
- Dagger module description and Python import in `__init__.py`

And finally, for extra flexibility:
- Introduce a Python [entry point](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins) to enable customization over how the module’s code is imported

### Example

For a new module:

```console
❯ dagger init --sdk=python my-module
```

Generate the name `my-module` for the project:

```diff
[project]
-name = "main"
+name = "my-module"
```

And put sources under `src/my_module`:

```diff
.
 ├── sdk
 ├── src
-│  └── main
+│  └── my_module
 │     ├── __init__.py
+│     └── main.py
[...]

PR #8709, opened 2024-10-14
feat(python): allow module package names other than main

helderco:
Fixes https://github.com/dagger/dagger/issues/8535

> [!note]
> This is a backwards compatible change. Only new modules get the new file structure. For existing modules, there's more freedom to change, but not enforced.

## What?

This changes the layout for **new modules**. For example:

```console
❯ dagger init --sdk=python my-module
```

Generates the name `my-module` for the project:

```diff
[project]
-name = "main"
+name = "my-module"
```

And puts sources under `src/my_module`:

```diff
.
 ├── sdk
 ├── src
-│  └── main
+│  └── my_module
 │     ├── __init__.py
+│     └── main.py
 ├── dagger.json
 ├── pyproject.toml
 └── uv.lock
```

With `__init__.py` using this import:

```python
from .main import MyModule as MyModule
```

But users have now complete control to specify where the main object needs to be imported from. For example:

```toml
# pyproject.toml

[project.entry-points."dagger.mod"]
main_object = "my_module.main:MyModule"
```

With the above, the `__init__.py` could be empty as it no longer needs to import the main object there because the entry point is specifically importing from `my_module.main`, not `my_module`.

## Refactors

<details><summary><h3>Simpler loading of <code>dagger.mod.Module</code></h3></summary>

Previously, user code was imported by an hardcoded:

```python
import main
```

The SDK would look for instances of `dagger.mod.Module`, which is what’s actually needed to execute. So you could use a custom subclass to customize some aspect of module handling. One early example of what I had in mind for this is to allow registering more hooks to teach dagger how to convert some custom types:

```python
import dagger.mod

class CustomModule(dagger.mod.Module)
    def __call__(self):
        # customize initialization
        ...

mod = CustomModule()

@mod.object_type
class Foo:
    ...
```

To do this, I added the `dagger.mod.Module` to the `@object_type` decorated class. This way, the SDK only needs to import the main object itself, since it’s now able to get the corresponding `dagger.mod.Module` instance from there. So even if you have different instances of `dagger.mod.Module` for some reason, it doesn’t matter. 

This is more correct, and avoids the need for the extra “smartness” and validation.

</details>

<details><summary><h3>Added customization on where to import the <em>main object</em></h3></summary>

I could just replace the hardcoded `main:<MainObject>` by the normalized module name coming from an env variable (e.g., `my_module:<MainObject>`), but I wanted to take advantage of the new file structure and module loading simplification, to give users more control by being able to specify explicitly where the main object should be imported from.

This way users have even less restrictions on how to organize their files. For example, it’s now possible to even remove the import in `src/my_module/__init__.py`:

```diff
-from .main import MyModule as MyModule
```

Since you can now tell the SDK where the main object is located, with a Python entry point:

```toml
# pyproject.toml

[project.entry-points."dagger.mod"]
main_object = "my_module.main:MyModule"
```

If this setting doesn’t exist, the SDK will try to load from `<package>:<MainObject>`:

```toml
main_object = "my_module:MyModule"
```

And if that import fails, it will fallback to trying out `main` (e.g., for currently existing modules):

```toml
main_object = "main:MyModule"
```

This ensures backwards compatibility, and less limitations on code organization. The only one remaining is that it still needs to be a distribution package (installable).

</details>

## Todo

- [x] https://github.com/dagger/dagger/pull/8874


core/integration/module_python_test.go
```diff
@@ -1599,7 +1731,7 @@ func pythonSource(contents string) dagger.WithContainerFunc {
 }
 
 func pythonSourceAt(modPath, contents string) dagger.WithContainerFunc {
-	return fileContents(path.Join(modPath, pythonSourcePath), contents)
+	return fileContents(path.Join(modPath, "src/test/__init__.py"), contents)
```
sipsma, line 1734: nit: could the name of the module become a parameter rather than hardcoding `test`? Just seems like a gotcha to have to name the module `test` in order for this to work as expected (IIUC).

Realize this might be a super tedious update, not a blocker

sipsma (approved):
One nit and looks like docs are pending; just want to make sure those are in by the release on Friday if we want this for v0.14.0 🙂

core/integration/module_python_test.go
helderco: I think this is fine considering the alternative. If I make it a parameter then it becomes pretty tedious whenever this is used in any language. So I figured a convention for a module named "test" is simpler. There's test cases that use a different module name, they just don't use this helper.

helderco:
> One nit and looks like docs are pending; just want to make sure those are in by the release on Friday if we want this for v0.14.0 🙂

Yeah, those are comming tomorrow, I hope. I just need to prioritize `dagger shell` to get it in time too.

TomChv (approved):
Great! Love that change

helderco:
Docs update in its own PR:
- https://github.com/dagger/dagger/pull/8874

Issue #8585, opened 2024-09-27
🐞 Iptables rules are persisting between pipelines


conorpp:
### What is the issue?

I create iptables rules in my container, and they persist and accumulate between pipeline runs.  I would expect them to not affect containers in future pipelines as they should be isolated.

### Dagger version

dagger v0.13.3 (registry.dagger.io/engine:v0.13.3) darwin/arm64

### Steps to reproduce

Run this multiple times and you'll see that the rules listed each time accumulates.

```go
func (m *Treasury) TestIptables(
	ctx context.Context,
) (string, error) {
	container := dag.Container().
		From("alpine:latest").
		WithExec([]string{"apk", "add", "iputils", "iptables"}).
		WithEnvVariable("CACHE_BUST", uuid.NewString())
	opts := dagger.ContainerWithExecOpts{InsecureRootCapabilities: true}

	container = container.WithExec([]string{
		// List the nat rules, should be none
		"sh", "-c", "echo 'pre-existing rules:' && iptables -t nat -L -v -n",
	}, opts)

	out, err := container.Stdout(ctx)
	fmt.Println(out)

	container = container.WithExec([]string{
		// Add rule 1
		"sh", "-c", "iptables -t nat -A PREROUTING -p tcp -j DNAT --to-destination 127.0.0.1",
	}, opts).WithExec([]string{
		// Add rule 2
		"sh", "-c", "iptables -t nat -A POSTROUTING -p tcp -j MASQUERADE",
	}, opts)

	out, err = container.Stdout(ctx)
	fmt.Println(out)

	return out, err
}
```

### Log output

Initial run:

```

pre-existing rules:
Chain PREROUTING (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination         
   10   600 DNAT       6    --  *      *       0.0.0.0/0            0.0.0.0/0            to:10.87.0.64
    0     0 DNAT       6    --  *      *       0.0.0.0/0            0.0.0.0/0            to:127.0.0.1
```

After running a few times:
```
pre-existing rules:
Chain PREROUTING (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination         
   10   600 DNAT       6    --  *      *       0.0.0.0/0            0.0.0.0/0            to:10.87.0.64
    0     0 DNAT       6    --  *      *       0.0.0.0/0            0.0.0.0/0            to:127.0.0.1
[...]

PR #8712, opened 2024-10-15
fix(engine): isolate netnses so insecure-root-added iptables rules dont leak through the bk cni pool [DEV-4584]

cwlbraa:
attempt at fixing #8585. 

This PR sets a host name for every netns to avoid buildkits netns pooling, which leaks changes to netnses across the pool. This is only relevant to insecure privileged containers, but can lead to very unpredictable network behavior when users modify the network namespace. 

This fix does come with a perf hit, at least on macos. See below for benchmarks. The perf hit has been mitigated by only avoiding pooling for privileged containers.

### DIFF
--- a/core/integration/sequential_test.go
+++ b/core/integration/sequential_test.go
@@ -1,0 +1,62 @@
+package core
+
+import (
+	"context"
+	"testing"
+
+	"dagger.io/dagger"
+	"github.com/dagger/dagger/testctx"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+)
+
+// This suite contains tests that must run on their own. This comes with a heavy time penalty.
+// Add additional tests here only if strictly necessary, typical if verifying some type of isolation.
+type SequentialSuite struct{}
+
+func TestSequential(t *testing.T) {
+	testctx.Run(
+		testCtx,
+		t,
+		SequentialSuite{},
+		// omitting testctx.WithParallel middleware to get the desired sequential behavior
+		testctx.WithOTelLogging(Logger()),
+		testctx.WithOTelTracing(Tracer()),
+	)
+}
+
+func (SequentialSuite) TestInsecureRootNetNSIsolation(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	opts := dagger.ContainerWithExecOpts{InsecureRootCapabilities: true}
+	baseContainer := c.Container().
+		From("alpine:latest").
+		WithExec([]string{"apk", "add", "iputils", "iptables"}).
+		WithEnvVariable("CACHE_BUST", uuid.NewString())
+
+	listNATRules := func(ctr *dagger.Container) (string, error) {
+		return ctr.
+			WithExec([]string{"sh", "-c", "iptables -t nat -L -v -n"}, opts).
+			Stdout(ctx)
+	}
+
+	withoutRules, err := listNATRules(baseContainer)
+	require.NoError(t, err)
+	require.NotContains(t, withoutRules, "DNAT")
+	require.NotContains(t, withoutRules, "to:127.0.0.1")
+
+	// iptables rules should not persist to "child" contains - ideally we could clone+CoW, but buildkit makes this difficult.
+	withoutClonedRules, err := listNATRules(baseContainer.WithExec([]string{
+		"sh", "-c", "iptables -t nat -A PREROUTING -p tcp -j DNAT --to-destination 127.0.0.1",
+	}, opts))
+	require.NoError(t, err)
+	require.NotContains(t, withoutClonedRules, "DNAT")
+	require.NotContains(t, withoutClonedRules, "to:127.0.0.1")
+
+	withRules, err := baseContainer.WithExec([]string{
+		"sh", "-c", "iptables -t nat -A PREROUTING -p tcp -j DNAT --to-destination 127.0.0.1 > /dev/null && iptables -t nat -L -v -n",
+	}, opts).Stdout(ctx)
+	require.NoError(t, err)
+	require.Contains(t, withRules, "DNAT")
+	require.Contains(t, withRules, "to:127.0.0.1")
+}
--- a/core/integration/services_test.go
+++ b/core/integration/services_test.go
@@ -400,7 +400,7 @@ func (m *Hoster) Run(ctx context.Context) error {
 		WithExec([]string{"httpd", "-v", "-f"}).
 		WithExposedPort(80).
 		AsService().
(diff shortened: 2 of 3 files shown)


sipsma:
> I suspect this means every WithExec will get a completely fresh netns, when really we want them inheriting their parent, but making them unable to modify it (cloning it, basically?)

I get the desire for that inheritance behavior but for now I'd actually say we should just ensure each exec gets a fresh netns; if a user needs certain settings to be applied it's up to them to do that in each exec.
* Reason being that getting COW-like behavior for netns settings is most likely extremely difficult (if possible at all)

> I also tried setting the CNI pool size to 0. That appears to have no effect- the test fails the same way as it does on main.

Ah, looking a little deeper it appears that when it's 0 they will still be reused but released after some timeout:
https://github.com/moby/buildkit/blob/4b4a9aab1e21a8e3754e30e0b16f3bd35bf52f8f/util/network/cniprovider/cni.go#L217

So I think the quickest way to try out disabling the pool would probably be to set the hostname to a random value if not already set.

Then once we know whether there's any noticeable difference we can go from there.

cwlbraa:
@sipsma sgtm, I'll rejigger the test and try to poke at performance. any tips on how or what you'd benchmark?

cwlbraa:
for posterity, here are some benchmark results using TestModule

Main, pooled netns:
`dagger call test specific --run="TestModule" --parallel=12  262.70s user 116.87s system 26% cpu 23:57.99 total`

https://dagger.cloud/dagger/traces/f4a24989481466a34e8301f25b07f8db

With isolated netns, timeout
`dagger call test specific --run="TestModule" --parallel=12  298.42s user 149.93s system 23% cpu 31:52.20 total`

https://dagger.cloud/dagger/traces/769b8a04c4babaa55716a60c7e8a2b2d

With isolated netns, no timeout:
`dagger call test specific --run="TestModule" --parallel=12  291.09s user 139.21s system 26% cpu 26:43.30 total`

https://dagger.cloud/dagger/traces/54370f5c8ac54498e961af0b64c20437

cwlbraa:
Benchmarking without all of our test weirdness, just spawning many alpine processes:

```
type HelloDagger struct{}

func (m *HelloDagger) ContainerEcho(stringArg string) *dagger.Container {
}

func (m *HelloDagger) Benchmark(ctx context.Context) error {
	g, ctx := errgroup.WithContext(ctx)

	for i := 0; i < 256; i++ {
		i := i // https://golang.org/doc/faq#closures_and_goroutines
		g.Go(func() error {
			guid := uuid.New().String()
			echoMessage := fmt.Sprintf("Echo call #%d - GUID: %s", i+1, guid)
			_, err := m.ContainerEcho(echoMessage).Stdout(ctx)
			return err
		})
	}

	return g.Wait()
}
```

Using released version:

```hyperfine --warmup 3 "dagger call benchmark"
Benchmark 1: dagger call benchmark
  Time (mean ± σ):     10.767 s ±  1.374 s    [User: 0.342 s, System: 0.127 s]
  Range (min … max):    9.055 s … 13.179 s    10 runs```


With main:
```hyperfine --warmup 3 "~/src/dagger/hack/with-dev dagger call benchmark"
Benchmark 1: ~/src/dagger/hack/with-dev dagger call benchmark
  Time (mean ± σ):     11.581 s ±  2.329 s    [User: 0.344 s, System: 0.129 s]
  Range (min … max):    9.071 s … 15.622 s    10 runs```

With netns isolation:
```hyperfine --warmup 3 "~/src/dagger/hack/with-dev dagger call benchmark"
Benchmark 1: ~/src/dagger/hack/with-dev dagger call benchmark
  Time (mean ± σ):     15.038 s ±  3.349 s    [User: 0.400 s, System: 0.147 s]
  Range (min … max):   10.796 s … 20.001 s    10 runs```


There’s clearly a significant perf penalty with this approach. Running more authoritative benchmarks right now.

**Comment by cwlbraa** [comment]
with 1 warmup run and 64 runs to clear out noise:

with main:

```sh
Benchmark 1: with-dev dagger call benchmark
  Time (mean ± σ):     11.540 s ±  1.743 s    [User: 0.368 s, System: 0.138 s]
  Range (min … max):    8.807 s … 15.886 s    64 runs
```

with netns isolation:
```sh
Benchmark 1: with-dev dagger call benchmark
  Time (mean ± σ):     13.848 s ±  2.160 s    [User: 0.431 s, System: 0.157 s]
  Range (min … max):   10.767 s … 18.507 s    64 runs
```

sipsma:
@cwlbraa awesome, thanks for gathering all that, super useful!

In absolute terms, that's not *terrible* but yeah feels like it's worth avoiding.

I'd be okay with only throwing the netns away for privileged execs, which AFAICT would be the only ones capable of making these sort of persistent networking changes anyways. Then the only impacted use cases would be users spinning up *tons* of privileged execs very rapidly. Most known use cases for privileged execs are relatively long-running services, so the tradeoff feels worth it here.

Rough idea to implement:
1. You can check whether it's a privileged exec in `executor_spec.go` w/ `if state.procInfo.Meta.SecurityMode == pb.SecurityMode_INSECURE`
2. For now, probably just ensure the hostname is not `""` if it's privileged (set it to a random value if not already set). I think this should work totally fine
  * The alternative is to upstream something that lets you force the netns to not return to the pool, but that would mostly just look slightly nicer while having the same effect

cwlbraa:
easy enough, I can't believe I hadn't already considered that mitigation.

cwlbraa:
due diligence reruns, 10 runs (so we see the noise here, but all else should be equal)

on main
```
Benchmark 1: with-dev dagger call benchmark
  Time (mean ± σ):     14.387 s ±  2.685 s    [User: 0.777 s, System: 0.231 s]
  Range (min … max):   11.071 s … 18.956 s    10 runs
```

with isolation:
```
Benchmark 1: with-dev dagger call benchmark
  Time (mean ± σ):     12.395 s ±  2.294 s    [User: 0.730 s, System: 0.212 s]
  Range (min … max):   10.100 s … 16.482 s    10 runs
```

cwlbraa:
this appears to [break](https://dagger.cloud/dagger/traces/554123412bcb0bd583a122fc03aa4cec) "TestWithHostnameModuleScoping/works within module" ... I can't tell if that's by design or what, gonna have to do some research on the intent of that test.

vito:
@sipsma Yeah seems most likely to just be a latent race.

cwlbraa:
@vito tight, will fix

vito (approved):
lgtm!

sipsma (approved):
🚀

PR #8713, opened 2024-10-16
docs: improve description of export properties

### LINKED ISSUE #8706: Description of `Directory.export` is misleading
Reported by user in [Discord](https://discord.com/channels/707636530424053791/1295320684658036767/1295351071732138015).

The API has the following description on [`Directory.export`](https://docs.dagger.io/api/reference/#Directory-export):

> Writes the contents of the directory to a path on the host.

This is misleading because when using a Dagger module, the directory is exported to the runtime container, not necessarily what you'd interpret as the "host". We should clarify this, but also for [`Container.export`](https://docs.dagger.io/api/reference/#Container-export) and [`File.export`](https://docs.dagger.io/api/reference/#File-export).

mcruzdev:
## What?

This pull request aims to improve `Directory.export`, `Container.export` and `File.export` description.

## Why?

Closes #8706

### DIFF
--- a/core/schema/container.go
+++ b/core/schema/container.go
@@ -524,8 +524,9 @@ func (s *containerSchema) Install() {
 
 		dagql.Func("export", s.export).
 			View(AllVersion).
-			Impure("Writes to the local host.").
-			Doc(`Writes the container as an OCI tarball to the destination file path on the host.`,
+			Impure("Writes to the filesystem where the API call is made from.").
+			Doc(`Writes the container as an OCI tarball to the destination file path on the current runtime container spawned by Dagger engine. 
+					When running in a module, it's written to the runtime container setup by the SDK.`,
 				`It can also export platform variants.`).
 			ArgDoc("path",
 				`Host's destination path (e.g., "./tarball").`,
--- a/docs/docs-graphql/schema.graphqls
+++ b/docs/docs-graphql/schema.graphqls
@@ -168,7 +168,8 @@ type Container {
   ): Container!
 
   """
-  Writes the container as an OCI tarball to the destination file path on the host.
+  Writes the container as an OCI tarball to the destination file path on the
+  current runtime container spawned by Dagger engine.
   
   It can also export platform variants.
   """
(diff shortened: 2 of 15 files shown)


vikram-dagger:
Ping @marcosnils for review

docs/docs-graphql/schema.graphqls
```diff
@@ -168,7 +168,7 @@ type Container {
   ): Container!
 
   """
-  Writes the container as an OCI tarball to the destination file path on the host.
+  Writes the container as an OCI tarball to the destination file path on the current runtime container spawned by Dagger engine.
```
vikram-dagger, line 171: I'm not sure it's necessary to update this file. AFAIK it is generated from the core schema. This might also be why the tests are failing. I'd like other reviewers to confirm this though.

core/schema/container.go
```diff
@@ -524,8 +524,8 @@ func (s *containerSchema) Install() {
 
 		dagql.Func("export", s.export).
 			View(AllVersion).
-			Impure("Writes to the local host.").
-			Doc(`Writes the container as an OCI tarball to the destination file path on the host.`,
+			Impure("Writes to the current runtime container.").
```
helderco, line 527: This isn't entirely correct. This is only true with modules, but you can still run dagger pipelines without modules, in which case it does write to your host.

docs/docs-graphql/schema.graphqls
helderco: When changes are made to the schema, you need a `dagger call` to regenerate docs and sdks.

https://github.com/dagger/dagger/blob/447c8a416f554e55aa2c9f0925f83d381b90b8cc/.dagger/README.md?plain=1#L81

https://github.com/dagger/dagger/blob/447c8a416f554e55aa2c9f0925f83d381b90b8cc/.dagger/README.md?plain=1#L116-L118

core/schema/container.go
helderco: It writes to the filesystem where the client call is made from. As it turns out, since modules run in a container, it's going to be exported to that container. If you run the pipeline directly from your host (i.e., `dagger run`) then that's where it will be exported to.

I'd prefer to avoid making the distinction here of running in a module vs non-module, as being the two options. On one hand, "host" can easily lead you astray. On the other hand, "runtime container" is too specific to this use case. Additionally "non-module" has implications with terminology that aren't cemented yet.

Best if there's a general way to phrase this, in line with "writes to the filesystem where the API call is made from" (i.e., avoiding the usual connotation to the word "host", but still general enough that it can match a container, virtual machine, etc...), but add a cautionary _note_ in the longer description with something like "when running in a module, it's written to the runtime container setup by the SDK".

PR #8714, opened 2024-10-16
ci: implement Alpine module w/ go apk lib for better caching

sipsma:
The previous module implementation called apko in a withExec, which resulted in every permutation of a given set of alpine packages creating a new tarball on disk (which added up to multiple GBs in our CI) and importing those tarballs to their own, single-layer containers into the engine. All of this wasted disk space and time.

The new implementation uses the apk go bindings inside apko to directly download, unpack and merge packages together all natively in Dagger.

The end result is that we only download+unpack packages once and use merge-op to efficiently combine them into different permutations of package installs, getting rid of the previous duplication and improving caching.

This just updates our in-repo Alpine module for now. We should also do the same for the Wolfi module so we can get rid of all usages of wrapping apko in a withExec.

---

TODO:
- [ ] I'll get some metrics w/ the improvements tomorrow (disk space and runtime), but running locally it is fast uncached and seems noticeably faster cached too


jedevc:
Hm, reverted the changes to wolfi to keep the old impl for now, but my commit is still failing :cry:

jedevc:
CI was failing because of this:

```
# github.com/dagger/dagger/auth.test

/usr/lib/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1

/usr/bin/gcc -m64 -s -o $WORK/b546/auth.test -rdynamic /tmp/go-link-1555625082/go.o /tmp/go-link-1555625082/000000.o /tmp/go-link-1555625082/000001.o /tmp/go-link-1555625082/000002.o /tmp/go-link-1555625082/000003.o /tmp/go-link-1555625082/000004.o /tmp/go-link-1555625082/000005.o /tmp/go-link-1555625082/000006.o /tmp/go-link-1555625082/000007.o /tmp/go-link-1555625082/000008.o /tmp/go-link-1555625082/000009.o /tmp/go-link-1555625082/000010.o /tmp/go-link-1555625082/000011.o /tmp/go-link-1555625082/000012.o /tmp/go-link-1555625082/000013.o /tmp/go-link-1555625082/000014.o /tmp/go-link-1555625082/000015.o /tmp/go-link-1555625082/000016.o /tmp/go-link-1555625082/000017.o /tmp/go-link-1555625082/000018.o /tmp/go-link-1555625082/000019.o /tmp/go-link-1555625082/000020.o /tmp/go-link-1555625082/000021.o /tmp/go-link-1555625082/000022.o /tmp/go-link-1555625082/000023.o /tmp/go-link-1555625082/000024.o /tmp/go-link-1555625082/000025.o -O2 -g -lresolv -O2 -g -lpthread -O2 -g -O2 -g

/usr/lib/gcc/x86_64-pc-linux-gnu/14/../../../../x86_64-pc-linux-gnu/bin/ld: cannot find /lib/ld-linux-x86-64.so.2: No such file or directory
```

The problem was that `ld-linux` writes files to `lib64` but `wolfi-baselayout` creates `lib64` as a symlink - so we need to have some kind of hack for this.

---

Additionally, I had to switch away from wolfi's `openssh` package - see:

```
$ dagger call -m modules/alpine --packages=openssh --distro=distro_wolfi container terminal
✔ connect 0.2s
✔ initialize 0.7s
✔ prepare 0.0s
✔ alpine(distro: "DISTRO_WOLFI", packages: ["openssh"]): Alpine! 0.5s
✔ Alpine.container: Container! 2.5s
✘ Container.sync: ContainerID! 0.2s
! process "/tmp/openssh-server.post-install" did not complete successfully: exit code: 2

Full trace at https://dagger.cloud/jedevc/traces/2737fca571261092e579ae47edae653f

Error: response from query: input: container.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withDirectory.withMountedFile.withExec.withoutMount.withDirectory.withDirectory.withMountedFile.withExec.withoutMount.withMountedFile.withExec.withoutMount.withEnvVariable.sync resolve: process "/tmp/openssh-server.post-install" did not complete successfully: exit code: 2

Stderr:
[dumb-init] /tmp/openssh-server.post-install: No such file or directory
Run 'dagger call container --help' for usage.
```

sipsma:
> EDIT: looks like the [busybox](https://github.com/wolfi-dev/os/blob/main/busybox.yaml) package does the linking as a trigger - so at this point in installation, the trigger hasn't run... so sh doesn't exist. Do we need to run triggers earlier? @sipsma not super familiar with now this logic should work.

Ah good fine, I had been inferring the right behavior based on output of running apk add manually so far, but I'll go to the source and see the expected times these runs

sipsma:
It looks like `apk` itself does have the same behavior we have in our module where:
1. pre-install happens before individual package install https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/master/src/database.c#L3012
2. post-install happens after individual package install https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/master/src/database.c#L3012
3. triggers run after the whole set of new packages have been installed https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/master/src/commit.c#L391

It also seems significant this only appears to happen w/ wolfi and not alpine (afaict), wonder if they are relying on some subtle different behavior in the way they build their images that avoids this issue.

I'd be fine with doing a little special casing here if nothing else. It might make sense to do a full install of the "basic" packages (baselayout, busybox, etc.) and *then* do the install of the rest of the packages. The equivalent of first doing an `apk add baselayout busybox` followed by a separate `apk add <rest of the packages requested>`. If our understanding is correct that might avoid this issue without many downsides. Giving it a shot.

sipsma:
@jedevc pushed a fix for what I described in previous comment. Locally it looks like it fixed that issue w/ openssh, so we should be on more solid ground. I left in your commit that changes to openssh-client anyways because we may as well only install that package if client-side is all we need.

Based on some comparisons:
* The new approach saves 2GB disk space for an uncached engine build
* Is about the same runtime uncached, so this approach is at least as fast as calling to apko in a withExec. Comparing cached runtime is hard because the problems arise when the engine prunes one entry that ends up causing cascading invalidation, but the new approach should handle that close to optimally by only pulling packages once and not invalidating unrelated pulls if one cache entry gets pruned.
   * Was able to run on airplane wifi successfully with local cache from earlier

I noticed that we still have the old apko approach show up in our module dependency dag; I found https://github.com/sagikazarmark/daggerverse/blob/68e9daa611183f5334b4059bac6f4aad62da7a37/gh/dagger.json#L14 specifically. Would be nice to fix that too but out-of-repo so leaving it out of scope.

jedevc (approved):
Whew :sweat_smile: that all took a bit longer than expected between us :laughing: But LGTM now :tada:

I applied one final change to add a note about `/usr/lib64`, and moved the `WithEnv("PATH")` to only be called once. I also saw what you did with the nvidia drivers for `gpuSupport`, so also moved the ubuntu logic to there as well, so it's all just in one place :smile: 

:shipit:

PR #8715, opened 2024-10-16
feat: various gha improvements

jedevc:
Rough summary:

- Imports https://github.com/shykes/gha into `modules/`
- Add `WithJob` to allow combining all SDKs together into one github workflow together
- Remove the `Settings` object - we started out keeping this, but as we split out some parameters into `Job`s, and some parameters in `Pipeline`s, it started making more sense to split them all out.
	- See `.github/main.go` for what this looks like now.
- No more constructor on the `gha` module - we wanted to be able to construct and manage pipelines separately, and then later combine them all later with `Generate`. This also opens up the possibility in the future to easily create other gha types - like `Action`s, etc, which we saw around the codebase and might want to use one day?

Next steps:
- Switch all jobs to use the `gha` module (as possible, ideally we'd split this into a separate PR).

> [!WARNING]
>
> Note to merger - do a rebase-merge, so we keep the original import of `shykes/gha` with no changes.


gerhard:
There is an extra change in `gha` was merged a few days ago:
- https://github.com/shykes/gha/pull/18

I also have an additive change locally, which I am just about to push on a separate branch (using it to test different runners).

Since there are many more changes on this PR than my changes, I am keen to see this merged so that I can bring across those two other sets of changes.

jedevc:
@gerhard in the first commit in this stack, we already included the change from shykes/gha#18 - I think we did it correctly, but could you check?

matipan (approved):

modules/gha/dagger.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "gha",
+  "sdk": "go",
+  "dependencies": [
+    {
+      "name": "wolfi",
+      "source": "github.com/shykes/daggerverse/wolfi@daae822bfa6d67e9e241b92ebe43e029e0c2c043"
```
gerhard, line 7: Should we depend on the local Wolfi module?

It will make iterating on changes quicker, same as for the `gha` module.

modules/gha/scripts/warm-engine.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Make sure not to load any implicit module
+cd $(mktemp -d)
+# Run a simple query to "warm up" the engine
+echo '{directory{id}}' | dagger query
```
gerhard, line 6: I found it useful to add `-vvv` to this query so that we see when pre-provisioned Engines are being used, instead of new ones spun up.

gerhard (approved):
Apart from a few nits which are OK to skip, this looks good to merge. I really like the new version's ability to compress multiple jobs into a single workflow. I was missing this when benchmarking the SDK pipelines across different runner types (I needed a new workflow which triggers multiple workflows, one per SDK variant).

I have a bunch of local changes that I would like to layer on top of this. Will wait for @shykes to confirm this merge before I refactor those changes (they are additive, so it won't be difficult to go either way). For now, I have them working with a local `gha` module which doesn't have these changes.

shykes:
👀

modules/gha/dagger.json
jedevc: Done :tada:

shykes:
I am not done with the review, but will start commenting as I go, to save time on round trips.

I have several issues with this PR, some of which I shared last week:

1. It's a major API change, almost a rewrite. Getting a chance to discuss it before a patch bomb would have been nice. I am feeling pressured to merge without the opportunity to discuss the design tradeoffs beforehand. Put yourselves in my shoes as a maintainer. Just ping me about it next time so I get a chance to weigh in without the pressure.

2. It changes the meaning of "Pipeline". In the original version it means "a Dagger pipeline", in other words: "a dagger call". In this PR it now means (I think) "a Github Actions Workflow". That's a major design change. I'm not sure if it's better or worse (I know we need a way to group several Dagger Pipelines into a single GHA workflow), but IMO if we're exposing GHA workflows, we should just call them workflows. The additional abstraction is no longer useful in this design.

shykes:
> Getting a chance to discuss it before a patch bomb would have been nice. I am feeling pressured to merge without the opportunity to discuss the design tradeoffs beforehand. Put yourselves in my shoes as a maintainer. Just ping me about it next time so I get a chance to weigh in without the pressure.

Additional note: if we "promote" community modules into a core stdlib in the future, like we discussed doing, we should be mindful of this.

shykes:
3. No more constructor. Why?

> No more constructor on the gha module - we wanted to be able to construct and manage pipelines separately, and then later combine them all later with Generate.

The current version of `gha` makes it very easy to set a common configuration for my entire GHA config. For example, my Dagger Cloud token. You removed that ability completely. Looking at our own CI, you moved that code to a "base job" abstraction in the caller module. So now, every caller of the gha module has to do more work. That doesn't seem like a good tradeoff to me.

4. No more config directory management

Same rationale I guess:

> we wanted to be able to construct and manage pipelines separately, and then later combine them all later with Generate.

The plumbing already existed for generating a single config file for a single workflow. It just needed to be exposed in the API. Why blow up the ability to manage an entire config directory, which is also very useful? For example I had implemented rename handling: in case of renaming / refactoring (which happens all the time), old generated files would be automatically removed, while leaving manual files alone. I think that feature just got nuked.

And for what benefit? You're re-implementing the same thing in our own CI module anyway, just less reusable. You say "more control" I just see "more work for everyone".

I'm sure I'm over-simplifying, but this part of the refactor doesn't seem necessary. Is this really needed to solve the problem at hand (grouping jobs into a workflow to solve the "one hundred manual retries" problem)? If not, let's reduce the scope of the change please.

shykes (dismissed):
- Exposing the underlying concepts of GHA workflow and job makes sense. I tried to abstract it away with "pipeline", but as we discovered, that abstraction is leaky. Specifically, the Github Actions UI maps manual retries to workflows, so a single workflow with 10 jobs is more manageable in the UI than 10 single-job workflows.
- We don't need the "pipeline" abstraction anymore. Let's just call these objects what they are: Workflows and Jobs. In practice that means simply renaming `Pipeline` to `Workflow`.
- No strong opinions on `Settings`, it was just an implementation detail anyway, and a janky one (kind of arbitrary what went into it and what didn't). But, it served a purpose... see next point.
- I like the ability to configure and generate a single Workflow file directly. But, I don't want to lose the ability to configure a whole directory also. 1) I want the ability to set common defaults. My cloud token; my default runner; etc. 2) I also want to keep the rename management; that was genuinely useful/ I understand the benefits of a cleaner leaner API; but objectively this PR nukes a useful feature. I don't have strong opinions about the API. For example, if you want to add a `Gha.Repository()` which carries the state for a repository-wide configuration, including defaults, list of pipelines etc, that's fine with me. I used the top-level type that way, de facto - it seemed simpler, but I'm not going to argue over specific API. I just don't want to lose the feature.
- CLI-friendliness: one reason the API is the way it is, is to preserve the ability to use it from the CLI. Since you can't pass object as argument to the CLI, I couldn't have an API that implied creating individual pipeline objects, then passing them into a higher-level repository object. But, with the shell coming soon, that restriction will soon be lifted. So, perhaps that can help us find a good compromise: merge the current "lighter" version, but as a follow-up, add `Gha.Repository().WithPipeline(Pipeline)` ?

In summary:

1. Please rename `Pipeline` to `Workflow` if you don't mind 🙏 to stick to GHA concepts
2. OK to merge as-is, but can we make a follow-up to add repo-wide config as described above?

shykes:
Note: we discussed this live with Justin and are aligned.

jedevc:
As discussed offline, I've dropped *all* of the refactoring changes from this PR so we can merge the import of this module quickly - this should unblock @gerhard from making the required changes he needs to make.

I'll quickly open a separate PR with the following:
- The refactorings done with @matipan
- Rename of `Pipeline` to `Workflow`
- Allow top-level configuration of common settings somehow:
	- @shykes the original logic for removing top-level common settings via constructor parameters was the following:
	- Suppose we have a `Gha.Pipeline` function that creates a pipeline, and then we attach it with `Gha.WithPipeline` - *which* top-level default settings should this pull from? Those two `Gha` objects could conceptually be two different `Gha` pipelines.
	- We settled on removing all top-level config in general, which solves this, but there are some good other alternatives:
		- In the case of multiple different `Gha` objects, we pull from both of them
		- Set all the defaults on a new `Repository` type - then we have `Gha.Pipeline` and `Repository.WithPipeline`, and this confusion is avoided.
	- I think we should drop the cli requirement - `dagger shell` is coming soon. 
- Bring back the `generate` command somehow (either at the top-level, or on the `Repository` type)
	- This should also have "rename support".

jedevc:
I'm going to merge this now, and continue the work in https://github.com/dagger/dagger/pull/8765.

The only changes to the actual content of the imported module are linting checks in [`f6c9405` (#8715)](https://github.com/dagger/dagger/pull/8715/commits/f6c94059cfab3f51fdce85c0760c2e5702455644) (since every go module in this repo is automatically linted, we need to ensure that it conforms).

PR #8724, opened 2024-10-17
feat(engine): ensure cache volumes are namespaced

### LINKED ISSUE #7211: Namespace cache volumes by module
## Problem

Cache volumes exist in the engine's global namespace: if modules `A` and `B` each create a cache volume with id `foo`, they will share read and write access to the same volume, as long as they are run on the same engine. This allows one module to corrupt the data of another module, either accidentally or maliciously.

## Solution

Namespace the name of cache volumes, so that each module can only read and write to its own cache volumes.

## See also

*This issue is an up-to-date reboot of https://github.com/dagger/dagger/issues/3345*

rajatjindal:
fixes #7211 

This PR namespaces the cache volumes so that across the modules they have different keys.

### DIFF
--- a/core/integration/cache_test.go
+++ b/core/integration/cache_test.go
@@ -107,3 +107,226 @@ func (CacheSuite) TestLocalImportCacheReuse(ctx context.Context, t *testctx.T) {
 
 	require.Equal(t, out1, out2)
 }
+
+func (CacheSuite) TestCacheIsNamespaced(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	fooTmpl := `package main
+
+import (
+	"context"
+)
+
+type Foo struct{}
+
+func (f *Foo) GetCacheVolumeId(ctx context.Context) (string, error) {
+	id, err := dag.CacheVolume("volume-name").ID(ctx)
+	return string(id), err
+}
+`
+	barTmpl := `package main
+
+import (
+	"context"
+)
+
+type Bar struct{}
+
+func (b *Bar) GetCacheVolumeId(ctx context.Context) (string, error) {
+	id, err := dag.CacheVolume("volume-name").ID(ctx)
+	return string(id), err
+}
+`
+	ctr := c.Container().
+		From(golangImage).
+		WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+		WithWorkdir("/work/bar").
+		With(daggerExec("init", "--name=bar", "--source=.", "--sdk=go")).
+		WithNewFile("main.go", barTmpl).
+		WithWorkdir("/work/foo").
+		With(daggerExec("init", "--name=foo", "--source=.", "--sdk=go")).
+		WithNewFile("main.go", fooTmpl)
+
+	fooID, err := ctr.
+		WithWorkdir("/work/foo").
+		With(daggerExec("call", "get-cache-volume-id")).
+		Stdout(ctx)
+	require.NoError(t, err)
+
+	barID, err := ctr.
+		WithWorkdir("/work/bar").
+		With(daggerExec("call", "get-cache-volume-id")).
+		Stdout(ctx)
+
+	require.NoError(t, err)
+	require.NotEqual(t, fooID, barID)
+}
+
+func (CacheSuite) TestCacheIdSameAcrossSession(ctx context.Context, t *testctx.T) {
+	session1 := connect(ctx, t)
+
+	fooTmpl := `package main
+	import (
+		"context"
+	)
+
+	type Foo struct {}
+	func (f *Foo) GetCacheVolumeId(ctx context.Context) (string, error) {
+		id, err := dag.CacheVolume("volume-name").ID(ctx)
+		return string(id), err
+	}
+	`
+
+	ctr1 := session1.Container().
+		From(golangImage).
+		WithMountedFile(testCLIBinPath, daggerCliFile(t, session1)).
+		WithWorkdir("/work/foo").
+		With(daggerExec("init", "--name=foo", "--source=.", "--sdk=go")).
+		WithNewFile("main.go", fooTmpl)
+
+	fooID, err := ctr1.
+		WithWorkdir("/work/foo").
+		With(daggerExec("call", "get-cache-volume-id")).
+		Stdout(ctx)
+	require.NoError(t, err)
+
+	session2 := connect(ctx, t)
+	ctr2 := session2.Container().
+		From(golangImage).
+		WithMountedFile(testCLIBinPath, daggerCliFile(t, session2)).
+		WithWorkdir("/work/foo").
+		With(daggerExec("init", "--name=foo", "--source=.", "--sdk=go")).
+		WithNewFile("main.go", fooTmpl)
+
+	fooID2, err := ctr2.
+		WithWorkdir("/work/foo").
+		With(daggerExec("call", "get-cache-volume-id")).
+		Stdout(ctx)
+
+	require.NoError(t, err)
+	require.Equal(t, fooID, fooID2)
+}
+
+func (CacheSuite) TestCacheVolumePassedAcrossModules(ctx context.Context, t *testctx.T) {
+	session := connect(ctx, t)
+
+	fooTmpl := `package main
+
+import (
+	"context"
+	"dagger/foo/internal/dagger"
+	"fmt"
+)
+
+type Foo struct{}
+
+func (f *Foo) Populate(ctx context.Context, input string) (*dagger.Container, error) {
+	return dag.Container().
+		From("alpine").
+		WithMountedCache("/tmp-cache-mount", dag.CacheVolume("cache-name")).
+		WithExec([]string{"sh", "-c", fmt.Sprintf("echo '%s' > /tmp-cache-mount/input.txt", input)}).
+		Sync(ctx)
+}
+
+func (f *Foo) Fetch(ctx context.Context) (string, error) {
+	cache := dag.CacheVolume("cache-name")
+	return dag.Bar().Fetch(ctx, cache)
+}
+`
+
+	barTmpl := `package main
+
+import (
+	"context"
+	"dagger/bar/internal/dagger"
+)
+
+type Bar struct{}
+
+func (f *Bar) Fetch(ctx context.Context, vol *dagger.CacheVolume) (string, error) {
+	return dag.Container().
+		From("alpine").
+		WithMountedCache("/tmp-cache-mount-bar", vol).
+		WithExec([]string{"sh", "-c", "cat /tmp-cache-mount-bar/input.txt"}).
+		Stdout(ctx)
+}
+`
+
+	ctr := session.Container().
+		From(golangImage).
+		WithMountedFile(testCLIBinPath, daggerCliFile(t, session)).
+		WithWorkdir("/work/bar").
+		With(daggerExec("init", "--name=bar", "--source=.", "--sdk=go")).
+		WithNewFile("main.go", barTmpl).
+		WithWorkdir("/work").
+		With(daggerExec("init", "--name=foo", "--source=.", "--sdk=go")).
+		WithNewFile("main.go", fooTmpl).
+		With(daggerExec("use", "./bar"))
+
+	inpContent := "some-foo-bar-content"
+	_, err := ctr.
+		WithWorkdir("/work").
+		With(daggerExec("call", "populate", "--input", inpContent)).
+		Stdout(ctx)
+	require.NoError(t, err)
+
+	fetchedFromCache, err := ctr.
+		WithWorkdir("/work").
+		With(daggerExec("call", "fetch")).
+		Stdout(ctx)
+	require.NoError(t, err)
+	require.Equal(t, inpContent, strings.TrimSpace(fetchedFromCache))
+}
+
+func (CacheSuite) TestCacheNotImpactedByChangeInModuleSource(ctx context.Context, t *testctx.T) {
+	session := connect(ctx, t)
+
+	fooTmpl := `package main
+	import (
+		"context"
+	)
+
+	type Foo struct {}
+	func (f *Foo) UseCacheVolume(ctx context.Context) (string, error) {
+		id, err := dag.CacheVolume("cache-name").ID(ctx)
+		return string(id), err
+	}
+	`
+
+	barTmpl := `package main
+	import (
+		"context"
+	)
+
+	type Foo struct {}
+	func (f *Foo) UseCacheVolume(ctx context.Context) (string, error) {
+		id, err := dag.CacheVolume("cache-name").ID(ctx)
+		return string(id), err
+	}
+
+	func (f *Foo) PassCacheVolume(ctx context.Context) (string, error) {
+		return f.UseCacheVolume(ctx)
+	}
+	`
+
+	ctr := session.Container().
+		From(golangImage).
+		WithMountedFile(testCLIBinPath, daggerCliFile(t, session)).
+		WithWorkdir("/work").
+		With(daggerExec("init", "--name=foo", "--source=.", "--sdk=go"))
+
+	fooID, err := ctr.
+		WithWorkdir("/work").
+		WithNewFile("main.go", fooTmpl).
+		With(daggerExec("call", "use-cache-volume")).
+		Stdout(ctx)
+	require.NoError(t, err)
+
+	fooID2, err := ctr.WithWorkdir("/work").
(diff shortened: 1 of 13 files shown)


rajatjindal:
one additional issue I am running into is how to hide this internal api `cache`. We don't want this internal api to be exposed to SDK's. 

I was looking for patterns in code around this, and noticed some api's with name like `__internal<SOMETHING>`, is that an ok pattern to use?

dagql/server.go
```diff
@@ -592,6 +592,26 @@ func NewInstanceForCurrentID[P, T Typed](
 	}, nil
 }
 
+// InstanceWithNewSelector creates a new Instance that's set to the ID calculated
+// using new selector.
+//
+// This is useful in scenarios where we want to inject something into the selector
+// automatically without exposing it to the user. e.g. namespacing of the selector
+func InstanceWithNewSelector[P, T Typed](
```
rajatjindal, line 600: cc @sipsma - this is the function I added on top of `NewInstanceForCurrentID`.  It seems to have worked, but also I am not 100% sure if there are any side-effects of creating a new id by changing the selector like this.

core/schema/cache.go
```diff
@@ -35,10 +37,26 @@ type cacheArgs struct {
...
-	// TODO(vito): inject some sort of scope/session/project/user derived value
-	// here instead of a static value
-	//
-	// we have to inject something so we can tell it's a valid ID
-	return core.NewCache(args.Key), nil
+func (s *cacheSchema) cacheVolume(ctx context.Context, parent dagql.Instance[*core.Query], args cacheArgs) (dagql.Instance[*core.CacheVolume], error) {
+	var inst dagql.Instance[*core.CacheVolume]
+
+	m, err := parent.Self.Server.CurrentModule(ctx)
+	if err != nil && !errors.Is(err, core.ErrNoCurrentModule) {
```
marcosnils, line 44: will this continue working in non-module pipelines? @rajatjindal is currently testing this now.

rajatjindal: Hi @marcosnils 

I verified that this works with non-module pipelines. I used following code to verify this:

```
package main

import (
	"context"
	"fmt"
	"os"

	"dagger.io/dagger/dag"
)

func main() {
	ctx := context.TODO()

	vol := dag.CacheVolume("volume-name")
	outout, err := dag.Container().From("alpine:latest").
		WithMountedCache("/foo", vol).
		WithExec([]string{"sh", "-c", "touch /foo/bar.txt"}).
		WithExec([]string{"sh", "-c", "echo 'hello'"}).
		Stdout(ctx)

	if err != nil {
		fmt.Printf("ERROR: %v\n", err)
		os.Exit(1)
	}

	fmt.Println(outout)
}

```

core/integration/cache_test.go
```diff
@@ -107,3 +107,98 @@ func (CacheSuite) TestLocalImportCacheReuse(ctx context.Context, t *testctx.T) {
 
 	require.Equal(t, out1, out2)
 }
+
+func (CacheSuite) TestCacheIsNamespaced(ctx context.Context, t *testctx.T) {
```
sipsma, line 111: We'll want an additional test where a function in one module passes a cache volume to a function in a different module and make sure that it works as expected (cache volume should be tied to the one who created it, not used it)

core/schema/cache.go
```diff
@@ -19,8 +20,9 @@ func (s *cacheSchema) Name() string {
 
 func (s *cacheSchema) Install() {
 	dagql.Fields[*core.Query]{
-		dagql.Func("cacheVolume", s.cacheVolume).
+		dagql.NodeFunc("cacheVolume", s.cacheVolume).
 			Doc("Constructs a cache volume for a given cache key.").
+			Impure("cacheVolume should be namespaced").
```
sipsma, line 25: Saw that the test workflows in this PR seems noticeably slower and am wondering if the impurity here is the cause. Marking this as impure and then returning another instance that includes `cacheVolume` might be tainting too many dagql IDs and thus giving us less dagql cache benefit (I am just going off theory here though, worth verification).

If so, one solution would be to go back to having a separate `__internalCacheVolume` API that's not impure; this use case is just sort of different than some of the other recent ones where we managed to avoid that hack...

I'd really like it if we could find something better though since that hack really pollutes telemetry, the API, etc.

I think the fundamental problem is that the current impurity system is too coarse (something that's come up in many discussions, not a new idea). Fully reworking it is too much to block this PR on but trying to think of some relatively quick steps we could take to improve it and fix this problem.

I guess what we want is that impurity can be decided on per-call rather than fixed in stone in the schema spec. I actually think that's doable in terms of the low-level dagql Call format. So the basic idea would be that when we check for impurity [here](https://github.com/rajatjindal/dagger/blob/16be68ae06badbcc65e2a6c14aad3b4752035831/dagql/server.go#L891-L891), the impurity value in the spec is just a default but could be overridden.

*Maybe* this is just a matter of setting up that plumbing between a `dagql.Select` call and that line linked above?

cc @vito in case you have any ideas

@rajatjindal this is a bit of ranty comment, sorry. Let me know if it makes sense, can try to clarify more if needed.

dagql/server.go
sipsma: This comment is related: https://github.com/dagger/dagger/pull/8724#discussion_r1844746862

We might need something sort of different than this if the impurity is indeed wreacking havoc on the dagql caching

core/schema/cache.go
vito: Took a swing at adding support for dynamic purity in DagQL - it's a PR that targets _this_ PR, so if we're good with the change we can either just merge it into this, or I can create a separate PR. It was just easiest to test in the context of the thing that needed it.

* https://github.com/rajatjindal/dagger/pull/1

tl;dr: now you can call `dagql.Taint(ctx)` dynamically anywhere in a resolver code path and it'll prevent the current API call from being cached, and also change the returned ID to be tainted - exactly the same effects of calling `Impure` at the schema level, just determined at runtime instead.

rajatjindal: Hi @vito

thanks for adding the support for dynamic purity. I rebased this branch on top of your changes (the new PR to the dagger repo) and accomodated the namespacing logic in this PR accordingly. It seems to have worked as expected.

thanks again.

vito: @rajatjindal Ah cool, glad it worked! Feel free to toss out some of the later commits, the socket + regen ones in particular - I think I need to do a bit more work to fix that case, but the rest of the PR seems OK.

rajatjindal: Hi @sipsma, one more thing I noticed, while testing these changes more, is that if we change the source of the module, that would mean the digest is changed and with the new `taint` functionality we will be creating a new cache volume. I don't think we want that behavior. 

I added a testcase for it, and it seems like that is indeed the case. if we use just the name for namespacing, that might not be sufficient too (high chance of namespace conflict). 

what do you think?

rajatjindal:
I think this PR is ready for final review/merge

jedevc:
We should probably review + merge https://github.com/dagger/dagger/pull/8971 first? cc @vito is that one ready to go?

vito:
@jedevc I flew too close to the sun with the last few commits and haven't had time to hunt down what went wrong, so I just pushed `HEAD~2:HEAD` and hopefully it'll pass now. I'm good with reviewing + merging it first assuming it passes.

core/schema/cache.go
sipsma: @rajatjindal great point... it's definitely a delicate balance in terms of exactly how stable an ID we want to use for the module here.

One important point is that today, it's "extremely stable" in that it's not even tied to the module at all, so I think it would make sense for us to start out with something pretty stable still rather than suddenly switching to something that's changing all the time.

I think this is what you said you switched to in discord, but to confirm, I'm imagining:
1. For git modules, the module name + the URL (excluding scheme like `https`/`ssh`/etc.) + the module source path
2. For local modules, the module name + the module source path

This is definitely not flawless, but seems like an okay starting place. This is worth revisiting though. Do you mind opening an issue for discussing this point more so we don't lose track of it after merging this PR?

rajatjindal: I've reported https://github.com/dagger/dagger/issues/9017 as a follow up. I was not using module source path explicitly (as it is part of modPath), but I've added it as part of namespace key now. 👍

core/schema/cache.go
```diff
@@ -32,13 +34,68 @@ func (s *cacheSchema) Dependencies() []SchemaResolvers {
...
+		refString, err := m.Source.Self.RefString()
+		if err != nil {
+			return inst, err
+		}
+		parsedRef := parseRefString(ctx, bk, refString)
+
+		// use combination of name, modPath and subpath as namespace key.
+		// if we only use name there is a high chance for namespace key conflict
+		// and high chance of cache-miss if we use module source digest for namespace key
+		namespaceKey = name + parsedRef.modPath + parsedRef.repoRootSubdir
```
jedevc, line 74: We can just use `m.Source.Self.Symbolic` right? That includes the clone-ref + subdir without the version information.

`parseRefString` is actually potentially an expensive call - it potentially makes network requests to try and determine the root/etc. We've already got that information available in the parsed module source, so we should prefer that.

core/schema/cache.go
```diff
@@ -32,13 +34,68 @@ func (s *cacheSchema) Dependencies() []SchemaResolvers {
...
+	}
+
+	// if no namespace key, just return the NewCache based on key
+	if namespaceKey == "" {
+		return dagql.NewInstanceForCurrentID(ctx, s.srv, parent, core.NewCache(args.Key))
```
jedevc, line 79: nit: for the args to `core.NewCache` we should always do something like `namespaceKey + ":" + args.Key`.

We don't want there to be a possibility for a non-module call to get the same cache-key as a module api call - which currently it seems like it can.

core/schema/cache.go
```diff
@@ -19,8 +20,9 @@ func (s *cacheSchema) Name() string {
 
 func (s *cacheSchema) Install() {
 	dagql.Fields[*core.Query]{
-		dagql.Func("cacheVolume", s.cacheVolume).
+		dagql.NodeFunc("cacheVolume", s.cacheVolume).
 			Doc("Constructs a cache volume for a given cache key.").
+			Impure("cache volume should be namespaced").
```
jedevc, line 25: I think this can now be rebased to use the new dynamic purity in https://github.com/dagger/dagger/pull/8971?

core/schema/util.go
```diff
@@ -103,6 +103,12 @@ func (maxVersion BeforeVersion) Contains(version string) bool {
 	return semver.Compare(version, string(maxVersion)) < 0
 }
 
+type InternalOnly string
+
+func (i InternalOnly) Contains(version string) bool {
+	return false
+}
```
jedevc, line 110: Looks like some dead code here :thinking:

core/schema/cache.go
vito: Dynamic purity was swapped out for elective purity - so now it's correct to mark this `Impure`. Instead, the `dagql.Select` call sets `Pure: true` when it re-execs itself.

jedevc:
I know it's annoying, but I'm going to remove this from the milestone - it's frustrating if it won't make it in, but I know there's still issues outstanding that we should take the time to properly understand and fix (cc @rajatjindal do you have an update on those, I remember we discussed them sync last week).

With this as the last feature remaining in the milestone, I don't think we should hold the entire release for this to be ready.

rajatjindal:
> I know it's annoying, but I'm going to remove this from the milestone - it's frustrating if it won't make it in, but I know there's still issues outstanding that we should take the time to properly understand and fix (cc @rajatjindal do you have an update on those, I remember we discussed them sync last week).
> 
> With this as the last feature remaining in the milestone, I don't think we should hold the entire release for this to be ready.

Hi @jedevc

Yeah, I understand. I am still working through the other issue (`update-cmd`, in retrospect I should have prioritized this issue as the other one was moved out of the milestone earlier). I will get back to this issue tomorrow, but I guess that would be too late for the release.

core/integration/cache_test.go
```diff
@@ -107,3 +107,235 @@ func (CacheSuite) TestLocalImportCacheReuse(ctx context.Context, t *testctx.T) {
...
+		WithExec([]string{"sh", "-c", "cat /tmp-cache-mount/input.txt"}).
+		Stdout(ctx)
+}
+
+func (f *Bar) FetchDirect(ctx context.Context) (string, error) {
+vol := dag.CacheVolume("cache-name")
+	return dag.Container().
+		From("alpine").
+		WithMountedCache("/tmp-cache-mount", vol).
+		WithExec([]string{"sh", "-c", "cat /tmp-cache-mount/input.txt"}).
```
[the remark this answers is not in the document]
rajatjindal: Hi @sipsma, I think this scenario should work because we are passing along the cache volume to `Bar` from `Foo`. do you still think that this testcase should have failed? if so, did I misunderstood the testcase you mentioned in other comment:

> We'll want an additional test where a function in one module passes a cache volume to a function in a different module and make sure that it works as expected (cache volume should be tied to the one who created it, not used it)

OR It maybe that this comment is showing up on a different line number now that I've pushed some more changes.

core/schema/cache.go
rajatjindal: yep, this seems to have worked as you suggested 👍

core/schema/cache.go
rajatjindal: For a module call, this should never happen and should always have a namespaceKey populated. we are returning with just Key only if namespaceKey turns out to be empty.

core/integration/cache_test.go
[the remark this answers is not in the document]
sipsma: Ah no, not sure if I misread it when I left the comment or if something changed, but looking again now this seems like it's testing+working as expected 👍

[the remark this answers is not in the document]
rajatjindal: just to avoid any confusion around caching, I changed the volume mount to a diff path.  👍

core/schema/cache.go
jedevc: Yes, this is my point - if a non-module call (that's not using modules at all, by using the standalone SDK or even a standalone `dagger query`), it *still* shouldn't be able to access the same cache volumes as modules. They should be entirely separate.

rajatjindal: if we attach a random key as namespaceKey here for "non-module" scenario, I think that will suffice here?

jedevc: If we do a random key, then it's not reproducible? So non-module usage cache volumes would never be reused.


> for the args to `core.NewCache` we should always do something like `namespaceKey + ":" + args.Key`.

Is there anything wrong with just doing this original suggestion?

rajatjindal: just thinking out loud here:

IIUC, there are 2 different ways this function will be called by the enduser:

- module-sdk making the call
- non-module sdk making the call

further for `module-sdk making the call`, there are two ways this function will be called:

-  if `namespace` is not provided. In this case we need to populate the `namespaceKey` arg and call this function again.  (lets call it scenario 1)
- if `namespace` is provided. Then it will return `core.NewCache(<namespaceKey> + ":" + <key-provided-by-user>)`.  (lets call it scenario 2). This is an internal call that we are making from within the function itself

for `non-module sdk making the call`, the `currentModule` will be nil, so `namespaceKey` will be empty string. so we are currently returning `core.NewCache(<key-provided-by-user>` here.  (lets call it scenario 3)

IIUC, what you are suggesting is that for scenario 3 here, we should always return `core.NewCache(<namespaceKey> + ":" + <key-provided-by-user>)`. is that understanding correct?

If so, what I am suggesting is that for both `module-sdk` and `non-module-sdk` *IF* namespace-key is empty, it will still return same `arg` to core.NewCache function call, so I am unable to understand what difference adding `"" + ":"` will make here. 

I am pretty sure I am missing something here, so apologies for going back and forth in advance :)

jedevc: Just realized that the user can actually just provide the `namespaceKey` themselves. Because of this, I don't think my point really matters.

Don't think it's worth continuing to debate this, it's incredibly minor - but what can happen is that module with key `abc` can define a CacheVolume `foo` to get `abc:foo`. Then a non-module user can specify just the key `abc:foo` and get exactly the same cache volume. This shouldn't be possible - because the callers are *different*, they shouldn't be able to get exactly the same cache volume.

The solution is to always have a `":"` there - the empty string `""` at the beginning is the non-module call - it makes it impossible to construct the same cache volume from the external module.

jedevc (approved):
LGTM now, @sipsma and @vito any last comments before merging?

sipsma (approved):

PR #8725, opened 2024-10-17
feat: use new cache policies

jedevc:
Fixes https://github.com/dagger/dagger/issues/7711.

This PR updates buildkit to use the new complex GC policies introduced in https://github.com/moby/buildkit/pull/5079 https://github.com/moby/buildkit/pull/5359, which allows us to set a more complex default policy that works out-of-the-box in more scenarios.


core/integration/localcache_test.go
jedevc, line 1: @sipsma I've updated these tests to also include examples of the new policies - but would appreciate a close look at these to make sure they're doing what I *think* they're doing.

engine/server/gc.go
```diff
@@ -143,52 +140,65 @@ func getGCPolicy(cfg config.GCConfig, root string) []bkclient.PruneInfo {
...
-	return keep.AsBytes(root)
 }
+
+const (
+	diskSpaceReservePercentage int64 = 10
+	diskSpaceReserveBytes      int64 = 10 * 1e9 // 10GB
+	diskSpaceFreePercentage    int64 = 20
+	diskSpaceMaxPercentage     int64 = 80
+	diskSpaceMaxBytes          int64 = 200 * 1e9 // 200GB
+)
```
jedevc, line 204: The old policy is very simple - use 75% of the disk.

The new policy is more complex:
- Always allow use of 10% or 10GB (whichever is smaller, to handle disks that are very small, like docker-desktop's defaults) - dagger's cache will never be forced to reduce below this amount
- Never consume more than 80% or 200GB (whichever is smaller again) - dagger's cache will also be gc-ed at this amount
- Try and keep 20% of the disk free at any point - but never below the reserved storage of 10%/10GB.

I *think* this makes sense? It does mean than on a huge disk, like a 1TB one, we won't use more than 200GB of storage - I think that's the right call for a local dev experience? For CI, I'm not sure, would appreciate any inputs @marcosnils @sipsma.

go.mod
```diff
@@ -282,3 +282,5 @@ replace (
 	go.opentelemetry.io/otel/log => go.opentelemetry.io/otel/log v0.3.0
 	go.opentelemetry.io/otel/sdk/log => go.opentelemetry.io/otel/sdk/log v0.3.0
 )
+
+replace github.com/moby/buildkit => github.com/jedevc/buildkit v0.8.2-0.20241010144035-f69c0798296a
```
jedevc, line 286: This points to a custom fork right now (so as to avoid pulling in all the protobuf changes before they've actually been battle-tested in a release - also I suspect it'll be a bit of a complex change :cry:)

We should point it to our fork at https://github.com/dagger/buildkit before merging.

sipsma (approved):
LGTM, no blockers but left one comment with thoughts about the defaults

engine/server/gc.go
jedevc: Yeah, makes sense - I've reverted the default to 75% usage, without the cap - if we do that, we can do it separately.

I'm working on a new config file mechanism as well, which should give us more ownership over our config, and will help us more easily configure the defaults (so we don't have to keep pointing to the buildkit docs).

PR #8728, opened 2024-10-17
ci: allow to share engine cache across dev calls

marcosnils:
This allows sharing the engine cache locally across `dagger call dev`
calls which results quite convenient when testing different engine
versions

i.e `dagger -m github.com/dagger/dagger@pull/8728/head call dev --shared-cache terminal`


sipsma (approved):

jedevc (approved):
LGTM, I wonder if we could also change the default here :thinking:

PR #8733, opened 2024-10-18
ci: speed up golangci-lint via cache volumes

sipsma:
Noticed another quick performance fix while working on other stuff.

Linting go code requires downloading deps + compiling code, so there's huge benefit to using cache volumes. Data from local runs:

| Test Scenario | No Cache      | Cached      | Cached + code change |
|---------------|---------------|-------------|----------------------|
| **main**      | 5m50s         | 21.6s       | 4m33s                |
| **w/ fix**     | 4m55s         | 21.6s       | 2m5s                 |


modules/golangci/lint.go
```diff
@@ -34,6 +45,10 @@ type LintRun struct {
 	Source *dagger.Directory
 	// +private
 	Path string
+	// +private
+	GoModCache *dagger.CacheVolume
+	// +private
+	GoBuildCache *dagger.CacheVolume
```
jedevc, line 51: :bulb: feature idea - it would be cool to have "derived" cache-volumes - the fact we need to pass multiple cache volumes around, when it's conceptually for the same thing is a bit weird.

Ideally you'd be able to pass one, and then do `CacheVolume.cacheVolume` or similar, and get another sub-cachevolume (either it would just modify the internal `name`, or it could even just mount a subpath of the cache volume into the container?).

jedevc (approved):

PR #8736, opened 2024-10-18
engine: optimize+stabilize container.From with digests

sipsma:
There's two interrelated improvements here.

---

First, before this change we were always making network requests for image metadata even in the case where the image ref had a digest (and was thus immutable) *and* the image was already cached locally.

That was wasteful since there's no need to query registries for metadata on immutable images we already have locally.

That's fixed by using the "PreferLocal" resolve mode on images with a digest, which tells buildkit to check the local cache before checking a registry.

---

Second, before this change even though we always resolved image ref digests (if they didn't already have one), the actual ID for container.From still included the unresolved image ref the user provided. That had some subtle but significant side effects.

For example, while running the `initialize` step of our CI module repeatedly, you'd often see a few hundred milliseconds spent repeatedly resolving a `uv` image (used by the python sdk module), especially if you waited a little bit between runs.

The reason was that the Python SDK didn't provide a digest for that image ref, so after the cached resolve timed out we'd resend requests for its metadata. This was especially unfortunate+unnecessary since the SDK functions are all cached. Even though the function never ran, the engine still found the unresolved image ref in the returned IDs and thus triggered network requests to resolve the image ref's digest.

The change here modifies container.From to still resolve image ref digests *but* now the returned ID includes that resolved ref rather than the unresolved one the user provided.

This results in the cached results of the Python SDK including that digest in the ID and thus saving us from having to re-resolve it when that SDK call was cached.

More generally, I think this behavior makes us more reproducible. E.g. when we enable general function caching we'll avoid this problem and ensure that cached return values of functions always point to the precise same image as when the value was originally cached.

---

Overall, when testing locally I can see in traces that this trims a few 100ms from spans, but the overall runtime is about the same since those resolves currently happen in parallel with other steps.

However, any future tweaks to our CI module could shuffle steps around such that this resolution actually would have extended the runtime of initialize, so we are now protected against that.

Additionally, I'm on gigabit connection at home. Anyone with more network latency (especially the "airplane wifi" type cases) will likely benefit from this quite a bit more noticeably.

### DIFF
--- a/core/container.go
+++ b/core/container.go
@@ -283,7 +283,7 @@ func (mnts ContainerMounts) With(newMnt ContainerMount) ContainerMounts {
 	return mntsCp
 }
 
-func (container *Container) From(ctx context.Context, addr string) (*Container, error) {
+func (container *Container) FromRefString(ctx context.Context, addr string) (*Container, error) {
 	bk, err := container.Query.Buildkit(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get buildkit client: %w", err)
@@ -299,10 +299,10 @@ func (container *Container) From(ctx context.Context, addr string) (*Container, 
 	refName = reference.TagNameOnly(refName)
 
 	if refName, isCanonical := refName.(reference.Canonical); isCanonical {
-		return container.FromCanonical(ctx, refName)
+		return container.FromCanonicalRef(ctx, refName, nil)
 	}
 
-	_, digest, _, err := bk.ResolveImageConfig(ctx, refName.String(), sourceresolver.Opt{
+	_, digest, cfgBytes, err := bk.ResolveImageConfig(ctx, refName.String(), sourceresolver.Opt{
 		Platform: ptr(platform.Spec()),
 		ImageOpt: &sourceresolver.ResolveImageOpt{
 			ResolveMode: llb.ResolveModeDefault.String(),
@@ -316,10 +316,15 @@ func (container *Container) From(ctx context.Context, addr string) (*Container, 
 		return nil, fmt.Errorf("failed to set digest on image %s: %w", refName.String(), err)
 	}
 
-	return container.FromCanonical(ctx, canonRefName)
+	return container.FromCanonicalRef(ctx, canonRefName, cfgBytes)
 }
 
-func (container *Container) FromCanonical(ctx context.Context, refName reference.Canonical) (*Container, error) {
+func (container *Container) FromCanonicalRef(
+	ctx context.Context,
+	refName reference.Canonical,
+	// cfgBytes is optional, will be retrieved if not provided
+	cfgBytes []byte,
+) (*Container, error) {
 	container = container.Clone()
 
 	bk, err := container.Query.Buildkit(ctx)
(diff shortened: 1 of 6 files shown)


core/schema/container.go
```diff
@@ -715,8 +718,64 @@ type containerFromArgs struct {
...
+			return inst, err
+		}
+
+		return dagql.Instance[*core.Container]{
+			Constructor: dagql.CurrentID(ctx),
+			Self:        ctr,
+			Class:       parent.Class,
+			Module:      parent.Module,
+		}, nil
+	}
```
sipsma, line 747: The updates here were another situation where we want a `NodeFunc` so that we can *conditionally* return a modified ID depending on the input.

In the past I had resigned to adding "internal" apis prefixed with `__` (to hide them from codegen) to do this, but realized the approach used here of:
1. creating a `dagql.Instance` and setting the constructor to `dagql.CurrentID` when you *don't* need any modifications 
2. calling the current field with a `Select` and modified args when you *do* need some modifications
 
is much much cleaner; avoided the need for `__internalFrom` and such (which pollutes telemetry besides being ugly).

cc @rajatjindal @helderco @vito, related to recent conversations on discord.

core/container.go
```diff
@@ -289,30 +289,59 @@ func (container *Container) From(ctx context.Context, addr string) (*Container,
...
+	canonRefName, err := reference.WithDigest(refName, digest)
+	if err != nil {
+		return nil, fmt.Errorf("failed to set digest on image %s: %w", refName.String(), err)
 	}
 
-	digested, err := reference.WithDigest(refName, digest)
+	return container.FromCanonical(ctx, canonRefName)
+}
+
+func (container *Container) FromCanonical(ctx context.Context, refName reference.Canonical) (*Container, error) {
```
jedevc, line 322: nit: should `container.From` take a `reference.Reference` to match this?

core/schema/container.go
jedevc: Neat :tada: we could do this with quite a few things :smile: 

Should we have a helper for this `Instance` incantation?

core/container.go
```diff
@@ -289,30 +289,59 @@ func (container *Container) From(ctx context.Context, addr string) (*Container,
...
-		return nil, fmt.Errorf("failed to resolve image %s: %w", ref, err)
+		return nil, fmt.Errorf("failed to resolve image %s: %w", refName.String(), err)
+	}
+	canonRefName, err := reference.WithDigest(refName, digest)
+	if err != nil {
+		return nil, fmt.Errorf("failed to set digest on image %s: %w", refName.String(), err)
 	}
 
-	digested, err := reference.WithDigest(refName, digest)
+	return container.FromCanonical(ctx, canonRefName)
```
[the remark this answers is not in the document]
sipsma: Doing this for one codepath, but it's kind of tricky for others, namely the one where:
1. `from` is called with a ref w/out digest and ResolveImageConfig is called to get that digest
2. `from` is re-called with the digest set, ResolveImageConfig is called again to get the actual image config

I think to skip the second call we'd have to make the image config bytes and optional arg to `from`, which feels not super clean. I'm not super concerned about this though; the second call to ResolveImageConfig is prefer local, so it most likely will not actually do anything. Even in the extremely obscure case it does because something got cleaned up, it's not harmful to actually call it again.

core/schema/container.go
sipsma: Yeah good idea, adding `dagql.NewInstanceForCurrentID` (mouthful, but it's at least "what it says on the tin" and that's what autocomplete is for anyways). Generalized it slightly so it can handle parent and self being different types too.

core/container.go
sipsma: That just puts the onus on the caller to do more stuff itself and handle more errors, etc. so don't think it's worth it. I renamed the methods to be `FromRefString` and `FromCanonicalRef` to clear it up more though.

core/container.go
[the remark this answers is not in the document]
jedevc: Yeah, sounds good to me :tada:

jedevc (approved):

PR #8745, opened 2024-10-21
feat(cli): add dagger uninstall command

rajatjindal:
add a `dagger uninstall` command. Following test scenarios are covered:

-  ✅  if local dependency is not available in dagger.json. it should error out. works
-  ✅  local dependency is not used - works.
-  ⚠️ local dependency is used - (The code gives compile error as the dependency usage is still there in module).
-  ✅ git dependency is not used - it should error out. works
-  ⚠️ git dependency is used - (The code gives compile error as the dependency usage is still there in module)
-  ✅ git dependency uninstall by git repo url + version - works
-  ✅ git dependency uninstall by git repo url only - works
-  ✅  git dependency uninstall by name - works
-  ✅ dependency source does not exist (git or local) e.g. folder removed or git repo deleted - What should be the expected behavior here?

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -346,6 +346,42 @@ var moduleInstallCmd = &cobra.Command{
 	},
 }
 
+var moduleUnInstallCmd = &cobra.Command{
+	Use:     "uninstall [options] <module>",
+	Short:   "Uninstall a dependency",
+	Long:    "Uninstall module as a dependency from the current module. The target module must be local.",
+	Example: "dagger uninstall hello",
+	GroupID: moduleGroup.ID,
+	Args:    cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, extraArgs []string) (rerr error) {
+		ctx := cmd.Context()
+		return withEngine(ctx, client.Params{}, func(ctx context.Context, engineClient *client.Client) (err error) {
+			dag := engineClient.Dagger()
+			modConf, err := getDefaultModuleConfiguration(ctx, dag, true, false)
+			if err != nil {
+				return fmt.Errorf("failed to get configured module: %w", err)
+			}
+			if modConf.SourceKind != dagger.ModuleSourceKindLocalSource {
+				return fmt.Errorf("module must be local")
+			}
+
+			modSrc := modConf.Source.
+				WithoutDependencies([]string{extraArgs[0]}).
+				ResolveFromCaller()
+
+			_, err = modSrc.
+				AsModule().
+				GeneratedContextDiff().
+				Export(ctx, modConf.LocalContextPath)
+			if err != nil {
+				return fmt.Errorf("failed to generate code: %w", err)
+			}
+
+			return nil
+		})
+	},
+}
+
 var moduleDevelopCmd = &cobra.Command{
 	Use:   "develop [options]",
 	Short: "Prepare a local module for development",
--- a/core/integration/module_cli_test.go
+++ b/core/integration/module_cli_test.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"testing"
 
+	"dagger.io/dagger"
 	"github.com/dagger/dagger/core/modules"
 	"github.com/dagger/dagger/testctx"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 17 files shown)


core/schema/modulesource.go
```diff
@@ -485,6 +485,23 @@ func (s *moduleSchema) moduleSourceDependencies(
...
+				for _, dep := range src.Self.WithoutDependencies {
+					removedDep, err := dep.Self.Source.Self.Symbolic()
+					if err != nil {
+						return err
+					}
+
+					// the currentDep is being uninstalled
+					if currentDep == removedDep {
+						return nil
+					}
```
jedevc, line 502: Should we cover non-exact matches? e.g. if I install `github.com/foo/bar@v1.2.3`, and uninstall `github.com/foo/bar` should it match?

I think it probably should (unless there's another potential match, in which case we should error with a special "ambiguous dependency" - maybe tricky to do based on current code structure).

rajatjindal:
> > * ❓ dependency source does not exist (git or local) e.g. folder removed or git repo deleted - What should be the expected behavior here?
> 
> Depends.
> 
> * If the dependency is in `dagger.json`, it should succeed: we _shouldn't_ be attempting to actually resolve the dependency that we're removing (so no git requests etc should happen here). We should just match against the source/name in `dagger.json`.

do you mean reading `dagger.json` directly and making changes to it? Right now, the logic is to handle this on server side by using `WithoutDependencies` api (which is a new API). but if the dependency source is not available, it errors out even before that. 

> * If the dependency isn't in `dagger.json` it should error out (though I think this is already covered in your tests).

I was thinking we should keep uninstall api idempotent and if we call uninstall and the dependency does not exist, we should let it succeed. (that is the current behavior). what do you think?

core/schema/modulesource.go
rajatjindal: right now it works like that. if we install `github.com/foo/bar@v1.2.3`, we can uninstall using exact url as well as without version.

however currently it does not work if we try to use just the name of the module as specified in `dagger.json`. do we need to support that as well?

jedevc:
> do you mean reading dagger.json directly

No, no, the API we've got at the moment in this PR makes perfect sense. But, we should make sure that the `Source` on the `ModuleDependency`s in `WithoutDependencies` isn't ever actually resolved - so it shouldn't matter if the `git` reference is impossible to resolve, or if the local directory doesn't exist.

> if we call uninstall and the dependency does not exist, we should let it succeed

Not convinced about this - I get the idea of having it be idempotent, but it goes against expectation from pretty much every other tool - `apt` / `npm` / `apk` / `pacman` / etc.

rajatjindal:
> No, no, the API we've got at the moment in this PR makes perfect sense. But, we should make sure that the Source on the ModuleDependencys in WithoutDependencies isn't ever actually resolved - so it shouldn't matter if the git reference is impossible to resolve, or if the local directory doesn't exist.

gotcha, I will try this out. I did started making some conditional changes, need to check more.

> > if we call uninstall and the dependency does not exist, we should let it succeed
> 
> Not convinced about this - I get the idea of having it be idempotent, but it goes against expectation from pretty much every other tool - apt / npm / apk / pacman / etc.

that is a dair point. I will make the change to error out if its not available in `dagger.json` 👍

vito:
> > Not convinced about this - I get the idea of having it be idempotent, but it goes against expectation from pretty much every other tool - apt / npm / apk / pacman / etc.
> 
> that is a fair point. I will make the change to error out if its not available in `dagger.json` 👍

FWIW this behavior isn't consistent across the board. `apt` exits 0 for a package that isn't installed:

```
❯ sudo apt remove htop
[sudo] password for vito:
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package 'htop' is not installed, so not removed
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.

~ took 3s
❯ echo $status
0
```

`pacman` errors out but `npm` doesn't.

Personally, I prefer the idempotency; much easier to script without cluttering with a bunch of checks/conditionals

cmd/dagger/module.go
```diff
@@ -346,6 +346,42 @@ var moduleInstallCmd = &cobra.Command{
 	},
 }
 
+var moduleUnInstallCmd = &cobra.Command{
+	Use:     "uninstall [options] <module>",
+	Short:   "Uninstall a dependency",
+	Long:    "Uninstall module as a dependency from the current module. The target module must be local.",
+	Example: "dagger uninstall github.com/shykes/daggerverse/hello",
```
helderco, line 353: I think you'd most likely want to uninstall by name rather than full ref, so the example should reflect that.

rajatjindal: done

core/integration/module_cli_test.go
```diff
@@ -1079,3 +1079,251 @@ func (m *OtherObj) FnE() *dagger.Container {
...
 	})
 }
+
+func (CLISuite) TestDaggerUnInstall(ctx context.Context, t *testctx.T) {
+	t.Run("local dep", func(ctx context.Context, t *testctx.T) {
+		t.Run("uninstall a dependency currently used in module", func(ctx context.Context, t *testctx.T) {
+			c := connect(ctx, t)
+
+			ctr := c.Container().
+				From("alpine:latest").
```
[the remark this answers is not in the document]
rajatjindal: done

rajatjindal:
> > > Not convinced about this - I get the idea of having it be idempotent, but it goes against expectation from pretty much every other tool - apt / npm / apk / pacman / etc.
> > 
> > 
> > that is a fair point. I will make the change to error out if its not available in `dagger.json` 👍
> 
> FWIW this behavior isn't consistent across the board. `apt` exits 0 for a package that isn't installed:
> 
> ```
> ❯ sudo apt remove htop
> [sudo] password for vito:
> Reading package lists... Done
> Building dependency tree... Done
> Reading state information... Done
> Package 'htop' is not installed, so not removed
> 0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
> 
> ~ took 3s
> ❯ echo $status
> 0
> ```
> 
> `pacman` errors out but `npm` doesn't.
> 
> Personally, I prefer the idempotency; much easier to script without cluttering with a bunch of checks/conditionals

Thank you for your feedback on this @vito. I am going to adjust the behavior to make it itempotent.

core/integration/module_cli_test.go
```diff
@@ -1079,3 +1079,251 @@ func (m *OtherObj) FnE() *dagger.Container {
...
+				WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+				WithWorkdir("/work").
+				With(daggerExec("init", "--sdk=go", "--name=foo", "--source=.")).
+				With(daggerExec("install", "github.com/shykes/daggerverse/hello"))
+
+			daggerjson, err := ctr.File("dagger.json").Contents(ctx)
+			require.NoError(t, err)
+			require.Contains(t, daggerjson, "hello")
+
+			daggerjson, err = ctr.With(daggerExec("uninstall", "github.com/shykes/daggerverse/hello@v0.3.0")).
```
vito, line 1284: Is this version ignored, or does the version have to match? If the latter, won't this test fail if a new version of `hello` ships?

core/integration/module_cli_test.go
```diff
@@ -1079,3 +1079,251 @@ func (m *OtherObj) FnE() *dagger.Container {
...
+				WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+				WithWorkdir("/work").
+				With(daggerExec("init", "--sdk=go", "--name=foo", "--source=."))
+
+			_, err := ctr.With(daggerExec("uninstall", "github.com/shykes/daggerverse/hello@v0.3.0")).
+				File("dagger.json").Contents(ctx)
+			require.NoError(t, err)
+		})
+
+		// this one is currently failing - I think this should be fixed
```
vito, line 1304: Still accurate? CI checks look green to me (aside from whatever's going on with `java-dev` and `typescript-dev`)

vito (approved):
lgtm (i think) - just left two questions!

jedevc (approved):

core/integration/module_cli_test.go
vito: Yeah agreed on both counts - definitely an edge case so I'm not too bothered by the current behavior, but it doesn't seem like it should be ignored if it's specified

rajatjindal: I've adjusted the code and tests to check for specific version if version number is specified. 

As part of this I also refactored the testcases (for `uninstalling git dependency` relates tests) to use test-matrix instead of individual tests that makes them significantly more readable. 

I'll try to get this done for `local dependency` testcases as well. 

in the meantime it would be great to receive any additional feedback that you might have.

core/integration/module_cli_test.go
```diff
@@ -1079,3 +1079,251 @@ func (m *OtherObj) FnE() *dagger.Container {
...
+			c := connect(ctx, t)
+
+			ctr := c.Container().
+				From(alpineImage).
+				WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+				WithWorkdir("/work/bar").
+				With(daggerExec("init", "--sdk=go", "--name=bar", "--source=.")).
+				WithWorkdir("/work").
+				With(daggerExec("init", "--sdk=go", "--name=foo", "--source=.")).
+				With(daggerExec("install", "./bar"))
```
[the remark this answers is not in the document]
rajatjindal: I've restructured the tests to use test-matrix pattern to remove this duplicate code.

rajatjindal:
Hi @jedevc/@vito 

Thank you for your reviews on this PR. I think this is ready for final iteration of review/merge. 

Thanks
Rajat Jindal

jedevc (approved):

Issue #8720, opened 2024-10-16
✨ Need "local" support for git repos that are directories on the filesystem


jpadams:
### What are you trying to do?

Today you can only refer to remote git urls so if you have a known git repo directory (has a `.git` dir), there is no easy way to get the current repo name, commit sha, current branch, current tags, etc.

@vmaffet-fast

### Why is this important to you?

Users are implementing modules for this functionality which may before in core instead.

It's often useful for tagging containers with the git sha or using in a go release process, for example.

### How are you currently working around this?

Writing own module for "local git" or running commands

jedevc:
Oops, this is fixed by #9730, and should have been closed then!

cc @vmaffet-fast

PR #8749, opened 2024-10-21
Use our new version module in more places in CI

jedevc:
While trying to move our https://github.com/dagger/dagger/blob/main/.github/workflows/engine-and-cli-publish.yml workflow more into dagger, I realized there's just a lot of complex versioning logic still in here (as well as in each of our corresponding SDK publishing modules).

This is an expansion of using our special opinionated version git module - ideally one day we can move a lot of it's functionality into https://github.com/dagger/dagger/issues/8720. By using it everywhere, we can simplify:
- `TestPublish`, and the need to pass `GIT_REF` around
- One git dir, and one git dir only - there were several hanging around, for go publishing, for git publishing, etc.
- Some new helpful APIs in `version.go` - these aren't currently used, but should be useful one-day, I added them in an attempt to decouple publishing from github events (not super easy right now).


.dagger/cli.go
```diff
@@ -61,8 +59,8 @@ func (cli *CLI) Publish(
 	}
 	ctr = ctr.
 		WithWorkdir("/app").
-		WithMountedDirectory("/app", cli.Dagger.Source()).
-		WithDirectory("/app/.git", gitDir)
+		WithDirectory("/app", cli.Dagger.Source()).
+		WithDirectory("/app", cli.Dagger.Git.Directory())
```
gerhard, line 63: Is `WithDirectory` additive, or does it change `/app` to `cli.Dagger.Git.Directory()` after the second call?

jedevc: It's additive, it uses `MergeOp` under the hood :tada: So we're merging the directories.

PR #8750, opened 2024-10-22
engine: add cpu stat/pressure metrics

sipsma:
Only displaying CPU pressure time in TUI right now since that's what I'm most interested in, but other CPU usage is plumbed  now too. Need to rework metrics display before adding more since it's starting to get to be too long a line.

---

Fun fact: when initializing our CI module from empty cache and a go codegen step takes a long time (~30s), up to half the time is spent shows up as full CPU pressure (which means every thread in the container is blocked waiting for a CPU to be available)... This on my 8 core m1 macbook.

Working on some fixes there (go build doesn't de-dupe parallel ongoing work across processes) but these metrics were useful to prove the problem was w/ cpu contention, so spinning out in the meantime.

### DIFF
--- a/engine/buildkit/resources/cpustat.go
+++ b/engine/buildkit/resources/cpustat.go
@@ -1,0 +1,145 @@
+package resources
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"dagger.io/dagger/telemetry"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+const (
+	cpuStatFile     = "cpu.stat"
+	cpuPressureFile = "cpu.pressure"
+
+	cpuUsageKey  = "usage_usec"
+	cpuUserKey   = "user_usec"
+	cpuSystemKey = "system_usec"
+)
+
+type cpuStatSampler struct {
+	cpuStatFilePath string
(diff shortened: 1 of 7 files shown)


engine/buildkit/resources/cpustat.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func newCPUPressureSampler(cgroupPath string, meter metric.Meter, commonAttrs attribute.Set) (*cpuPressureSampler, error) {
+	s := &cpuPressureSampler{
+		cpuPressureFilePath: filepath.Join(cgroupPath, cpuPressureFile),
+		commonAttrs:         commonAttrs,
+	}
+	var err error
+
+	s.someTotal, err = meter.Int64Gauge(telemetry.CPUStatPressureSomeTotal, metric.WithUnit(telemetry.MicrosecondUnitName))
```
vito, line 108: Passing thought, not blocking: could be worth configuring [`WithDescription`](https://pkg.go.dev/go.opentelemetry.io/otel/metric#WithDescription) for all of these. I'm guessing that will make it all the way to the front-end, where it could be e.g. a tooltip in the web UI.

vito (approved):
neat!

engine/buildkit/resources/cpustat.go
sipsma: Agreed, added those

PR #8753, opened 2024-10-22
chore(dependabot): avoid bumping certain tricky sets of packages

jedevc:


.github/dependabot.yml
```diff
@@ -16,22 +16,12 @@ updates:
         applies-to: version-updates
         patterns:
           - "*"
-
-  - package-ecosystem: "gomod"
```
TomChv, line 20: Just to confirm, it's because `/dev` no longer exist right?

.github/dependabot.yml
```diff
@@ -80,6 +70,9 @@ updates:
         applies-to: version-updates
         patterns:
           - "*"
+        exclude-patterns:
```
TomChv, line 73: Yes good idea, however do we have any cron or something to still remind us to update deps? Otherwise we'll most likely forget

TomChv (approved):
Approved, but I left some questions :D

.github/dependabot.yml
jedevc: Yeah, I'm a bit unsure of what to do here.

Dependabot will still notify us of security issues - but aside from that, I think we should probably hold updates until we really need functionality in new versions. Updating it has been a pain in the past :cry: 

Let's see how this new config works, and we can see if we want automated upgrades on a cron job later.

.github/dependabot.yml
jedevc: Yup, we missed this one when we moved it to `.dagger`.

PR #8755, opened 2024-10-22
Unify publish steps

jedevc:
This PR unifies engine and SDK publishing together - there's no need to keep these separate anymore.

Essentially instead of manually and creating `sdk/*/vX.Y.Z` + `helm/chart/vX.Y.Z` tags, we instead auto-magically create them using the github release process. This means that we can fan-out the SDK releases *automatically* from the main tag.

This should cut a good amount of the toil from the release process.


RELEASING.md
```diff
@@ -2,80 +2,16 @@
...
-- [🪣 Install scripts ⏱ `2mins`](#-install-scripts--2mins#)
-- [🐙 dagger-for-github ⏱ `10mins`](#-dagger-for-github--10mins#)
-- [🍺 dagger Homebrew ⏱ `2mins`](#-dagger-homebrew--2mins#)
-- [❄️ nix ⏱ `2mins`](#-nix--2mins#)
-- [⚙️ CI ⏱ `2mins`](#-ci--2mins)
-
-This is a high-level diagram of how all the pieces fit together:
-
-```mermaid
-flowchart TB
```
jedevc, line 24: So with such a large refactor of the process, this probably needs updating - but a couple things:

- Is this actually helpful? I find mermaid's rendering slightly unhelpful to actually improve understanding.
- This is *very* hard to update - updating the diagram means lots of weights often need rebalancing to get a sane result.
- What's the actual purpose of this diagram? Is it to understand the process? The resulting artifacts?

I think it's worth removing.

.github/workflows/publish.yml
```diff
@@ -68,16 +68,139 @@ jobs:
...
+      - name: "notify"
+        if: github.ref_name != 'main'
+        uses: ./.github/actions/notify
+        with:
+          message: "🐹 Go SDK: https://github.com/${{ github.repository }}/releases/tag/${{ github.ref_name }}"
+          discord-webhook: ${{ secrets.NEW_RELEASE_DISCORD_WEBHOOK }}
+
+  publish-sdk-php:
+    needs: publish
+    runs-on: "${{ github.repository == 'dagger/dagger' && 'dagger-g2-v0-13-5-4c' || 'ubuntu-latest' }}"
```
sipsma, line 91: here and everywhere below should be v0-13-6 I think?

sipsma (approved):
Just noticed some engine runner versions that probably need updating, otherwise LGTM!

jedevc:
Success! This ran on `main`:

image

https://github.com/dagger/dagger/actions/runs/11596667367

One publish workflow for everything!
- Engine and CLI first
- Go + PHP use git publishing, so run (but the notify step is correctly skipped)
- All other SDKs and helm need to be tagged, so are skipped

PR #8758, opened 2024-10-22
fix(engine): ensure secret variable is expanded in WithExec

### LINKED ISSUE #8752: Using `expand` with a non-existent variable should fail
With https://github.com/dagger/dagger/issues/7951 implemented in https://github.com/dagger/dagger/pull/8427, we have some surprising behavior.

It's possible that the variable isn't set on the container - currently this silently fails - I think we should consider making failure the default.

Even if we don't - we at least explicitly handle the `withSecretVariable` case - we can't/shouldn't expand secret variables, so we should fail with a nice error message in this case. See this discord thread: https://discord.com/channels/707636530424053791/1297911858778341428/1298211444298350634

```python
        versions = await (
            dag.container()
            .from_("alpine:latest")
            .with_secret_variable("GITEA_TOKEN", token)
            .with_env_variable("URL", url)
            .with_env_variable("OWNER", owner)
            .with_exec(["apk", "add", "curl"])
            .with_exec(
                [
                    "sh",
                    "-c",
                    (
                        'curl "$URL/api/v1/packages/$OWNER?access_token=$GITEA_TOKEN"'
                        # ' --header "Authorization: token $(cat /run/secrets/gitea-token)"'
                        # ' --header "Authorization: token $GITEA_TOKEN"'
                        # " -H 'accept: application/json'"
                    ),
                ],
                expand=True
            )
            .stdout()
        )
```

Setting `expand` to `True` here should have explicitly failed with a message about `GITEA_TOKEN` not being set.

rajatjindal:
potentially fixes #8752 

Please do not merge yet. Needs discussion/review from @jedevc @shykes

### DIFF
--- a/core/integration/container_test.go
+++ b/core/integration/container_test.go
@@ -4657,6 +4657,18 @@ func (ContainerSuite) TestEnvExpand(ctx context.Context, t *testctx.T) {
 		require.Equal(t, hashStr, hashStrCmd)
 	})
 
+	t.Run("using secret variable with expand returns error", func(ctx context.Context, t *testctx.T) {
+		secret := c.SetSecret("gitea-token", "password2")
+		_, err := c.Container().
+			From("alpine:latest").
+			WithEnvVariable("CACHEBUST", identity.NewID()).
+			WithSecretVariable("GITEA_TOKEN", secret).
+			WithExec([]string{"sh", "-c", "test ${GITEA_TOKEN} = \"password\""}, dagger.ContainerWithExecOpts{Expand: true}).
+			Sync(ctx)
+
+		require.ErrorContains(t, err, "expand cannot be used with secret env variable \"GITEA_TOKEN\"")
+	})
+
 	t.Run("env variable is expanded in Export", func(ctx context.Context, t *testctx.T) {
 		wd := t.TempDir()
 
--- a/core/schema/container.go
+++ b/core/schema/container.go
@@ -7,6 +7,7 @@ import (
 	"io/fs"
 	"os"
 	"path"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -1479,10 +1480,28 @@ func expandEnvVar(ctx context.Context, parent *core.Container, input string, exp
 		return "", err
 	}
 
-	return os.Expand(input, func(k string) string {
+	secretEnvs := []string{}
+	for _, secret := range parent.Secrets {
+		secretEnvs = append(secretEnvs, secret.EnvName)
+	}
+
+	var secretEnvFoundError error
+	expanded := os.Expand(input, func(k string) string {
+		// set error if its a secret env variable
+		if slices.Contains(secretEnvs, k) {
+			secretEnvFoundError = fmt.Errorf("expand cannot be used with secret env variable %q", k)
+			return ""
+		}
+
 		v, _ := core.LookupEnv(cfg.Env, k)
 		return v
-	}), nil
+	})
+
+	if secretEnvFoundError != nil {
+		return "", secretEnvFoundError
+	}
+
+	return expanded, nil
 }
 
 type containerWithSecretVariableArgs struct {


core/integration/container_test.go
```diff
@@ -4657,6 +4657,22 @@ func (ContainerSuite) TestEnvExpand(ctx context.Context, t *testctx.T) {
...
+		secret := c.Host().SetSecretFile("mysecret", filepath.Join(dir, "some-file"))
+		output, err := c.Container().
+			From("alpine:latest").
+			WithEnvVariable("CACHEBUST", identity.NewID()).
+			WithSecretVariable("GITEA_TOKEN", secret).
+			WithExec([]string{"sh", "-c", "echo -n ${GITEA_TOKEN}"}, dagger.ContainerWithExecOpts{Expand: true}).
+			Stdout(ctx)
+
+		require.NoError(t, err)
+		require.Equal(t, "***", output)
```
rajatjindal, line 4673: is there a way to verify secret is set to the value as expected?

rajatjindal: using a `test` command in `withExec` now to verify this.

core/schema/container.go
```diff
@@ -1479,7 +1480,17 @@ func expandEnvVar(ctx context.Context, parent *core.Container, input string, exp
...
+	secretEnvs := []string{}
+	for _, secret := range parent.Secrets {
+		secretEnvs = append(secretEnvs, secret.EnvName)
+	}
+
 	return os.Expand(input, func(k string) string {
+		// do not expand if its a secret env variable
+		if slices.Contains(secretEnvs, k) {
+			return "${" + k + "}"
+		}
```
jedevc, line 1492: Huh this is a kinda neat idea - if we can't expand a var, leave it there.

That said, I think it's still a bit confusing - the case for `expand` is that there is no shell, and you want to expand it without one. So leaving shell-related expansion around in the result isn't great.

Maybe we should expand secrets as `***`? Censor them here if we really want this to succeed. Although, I still think that raising an error here makes more sense, since we can't really safely/generically expand secrets here at all.

rajatjindal: Hi @jedevc,

one thing I want to make sure is that you have read through this comment where I have tried to summarize the issue: https://github.com/dagger/dagger/issues/8752#issuecomment-2429864720

tl;dr; it seems the secrets expansion happens [here](https://github.com/dagger/dagger/blob/90befd2748d41b6c5112f9e296a3579a773471c9/core/container_exec.go#L212C8-L212C17), which happens after the expand function has been called already. 

so if we leave the environment variables which are suppose to be secrets-env-variables, then they will expand in the downstream function as expected.

does that make sense?

jedevc: > tl;dr; it seems the secrets expansion happens [here](https://github.com/dagger/dagger/blob/90befd2748d41b6c5112f9e296a3579a773471c9/core/container_exec.go#L212C8-L212C17), which happens after the expand function has been called already.

Secrets expansion is not actually done here. The code you link to only specifies to buildkit that those secrets exist, and provides them as environment variables - there is no expansion in dagger or buildkit. The reason that secrets are expanded at all is because the shell `sh` is doing it.

> so if we leave the environment variables which are suppose to be secrets-env-variables, then they will expand in the downstream function as expected.

Right, but only if there is a shell - if there is no shell, for instance you're doing `WithFile("${secret_name}")`, then this would just create a file called `${secret_name}` - which is confusing.

I think the better behavior is to just provide a neat error message to users if they expand a secret variable: skipping it and leaving it empty is confusing, expanding to `***` is possible, but also confusing/

rajatjindal: ah I see what you are saying. 

one of the intermediate revision I had was to implement a new function just for WithExec, where I keep the secret string as is, and for all other scenarios, we can error out if the secret variable is used in expand env.

I think that should cover both exec and other scenario 

OR

we can disable expandEnv in WithExec and rely on shell expansion.

rajatjindal: 👍 , so I have now made change to not allow expanding Secret Variables when using Expand. It will now error out with a msg `expand cannot be used with secret env variable "FOO"`

jedevc (approved):

PR #8762, opened 2024-10-23
feat(sdk/dotnet): implements SDK pipeline

wingyplus:
Add Dotnet to the `SDK` type and implements 3 functions, `lint`,
`test` and `generate`. Those functions are calling `dotnet-sdk-dev`
module that living inside `sdk/dotnet/dev`.

The custom dagger code in dev module and introspection.graphql are also
removed since we're rely on Dagger Engine on the root module.

And also add a dotnet pipeline to the GHA to make it run on CI.


jedevc:
@wingyplus is this okay to bring out of draft?

wingyplus:
> @wingyplus is this okay to bring out of draft?

The code is ready for review. But need to investigate more on CI why it hanged up during running tests. (looks like it’s a DotNet issue with the container).

wingyplus:
Update the progress:

* I do implement `generate` function from our discussion in #9288. So you can call `dagger call sdk dotnet generate export --path=.` from the root dagger repository.
* Since we're rely on introspection.json from root module, the introspection part in `sdk/dotnet/dev` module is now removed. 
* Pass all tests 🎉 
* I found lint is failure due to code format issue. This PR fix it already.

jedevc (approved):

PR #8765, opened 2024-10-23
API changes to gha module

jedevc:
Follow-up to https://github.com/dagger/dagger/pull/8715.

This PR brings in several refactorings, specifically with the intention of allowing multiple jobs per workflow, so we can have groups like https://github.com/dagger/dagger/actions/runs/11501610429/job/32014608115?pr=8765

As part of this, we need to rework the model slightly:
- We now map closer directly to Github `Workflow`s, instead of dagger `Pipeline`s - so we rename the type for this
- We create a top-level `Workflow` type, instead of just having `WithWorkflow(all the args)` - this means we can store `Workflow`s, and add jobs to them using `WithJob`, as well as setting default workflow settings easily.
- The top-level constructor can be used to specify default workflow settings and default job settings - these are applied on `Workflow`/`Job` and `WithWorkflow`.

Some non-user facing refactorings:
- Code is slightly reoganized - API types are moved into `api`, self-checks are added into `checks.go`, and all the steps are moved into `steps.go`.
- The rename support is slightly simplified, our generated workflows all have consistent names for `.gitattributes` with `.gen.yml`, so we can assume the same for clearing up old workflows.


modules/gha/main.go
```diff
@@ -8,228 +8,217 @@
...
-	PullRequestConcurrency string
-	Debug                  bool
-	FileExtension          string
-	Repository             *dagger.Directory
-	TimeoutMinutes         int
-	Permissions            Permissions
+func (gha *Gha) Repository(
+	jobDefaults *Job, // +optional
+	workflowDefaults *Workflow, // +optional
+) *Repository {
```
jedevc, line 35: *Kinda* reconsidering this in light of #8792. Having the defaults set on the `Repository` is a bit odd, I wonder if you should be setting them on the constructor.

That way you could do:

```go
dagger := dag.Gha(dag.GhaOpts{Pipeline: dag.Pipeline(dag.PipelineOpts{
	OnPushBranches: []string{"main"},
	Runner: []string{"dagger-runner-1"},
})})

depot := dag.Gha(dag.GhaOpts{Pipeline: dag.Pipeline(dag.PipelineOpts{
	OnSchedule: []string{"6 0 * * *"},
	Runner: []string{"depot-runner-1"},
})})
```

Then you just pass your `Gha` objects around to create workflows, and assemble your workflows using `Repository.WithPipeline`.

Thoughts @matipan?

shykes: > `dag.Gha(dag.GhaOpts{Workflow:`

- Assuming `Workflow:` is the default workflow, I would prefer calling it `Defaults`.
- Slight preference for pseudo-passing a `WorkflopOpts` argument instead of a `Workflow`. I know that involves a lot of copy-pasting today, but IMO it's still worth it. Easier to change internal implementation than user-facing API.
- I can live with keeping the `Workflow` type as long as we call it `Defaults` (in other words: an argument name that corresponds to the intent).

matipan: I agree that settings the defaults at a constructor level gives users more flexibility without introducing a new thing into the module. Having two `Repository` objects would be a bit confusing. `WorkflowGroups` would be an alternative but I hate it 😆. Constructor-based defaults sounds good!

jedevc: Okay, made these changes - `Repository` is gone :tada: Now it's much clearer and succinct.

Take a look at the new `.github/main.go` for a more complicated example: we have two `Gha` objects - one as a `Workflows`, another as a `DaggerRunner`. This complexity isn't required *right* now, they could just be the same - but it's a baseline for how we should implement https://github.com/dagger/dagger/pull/8792. The runners have defaults for triggers+github runners, and we can easily split out different jobs to different jobs. This is possible because we take into account *both* the `Gha`s: all the defaults always propagate, no matter where they're specified.

jedevc:
Should be ready for a final review - I've made suggested changes, finalized the refactoring, would appreciate any last minute thoughts!

shykes (approved):
A few notes but can be a follow-up 👍

PR #8769, opened 2024-10-24
feat(sdk/elixir): move entrypoint invoke task to dagger_sdk

wingyplus:
Before this changeset, the invoke task living inside a user defined module. This is confuses for someone who read the module source code.

So, this changeset move the invoke task to the `dagger_sdk` and modify the runtime to use this task instead.

And the runtime is no longer generate mix task anymore.


jedevc (commented):
LGTM :tada:

jedevc (approved):

PR #8775, opened 2024-10-24
docs: add link to daggerverse examples on dagger mod init partial

kpenfound:


jpadams (approved):
Tested! Works well. Great way for folks to find examples they want to emulate in their language of choice! 👍

PR #8776, opened 2024-10-24
engine: fix excessive cache invalidation for non-function nested execs

sipsma:
User reported unexpected cache invalidation here: https://discord.com/channels/707636530424053791/1299094006592049262/1299094006592049262

Funnily enough, I had today ran into the exact same problem but in the context of our Go SDK's internal `codegen` exec, so we also need this fix (among others) to improve module initialization speed 😄

---

Before this, when nesting was enabled for an exec we would always scope the cache key with the dagql call digest for that exec.

This is important for the case of a module function call since it results in the cache key being scoped to that function's args and receiver.

However, this was also happening for non-function nested execs (i.e. use of the ExperimentalPrivilegedNesting flag). For those execs, this was unnecessary. It also resulted in excessive cache invalidation since dagql call digests aren't (yet) smart enough to incorporate content-based cache keys.

Only enabling the cache key scoping for function calls specifically fixes the problem.

### DIFF
--- a/.changes/unreleased/Fixed-20241024-152529.yaml
+++ b/.changes/unreleased/Fixed-20241024-152529.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: Fix excessive cache invalidation of `withExec`s using the `ExperimentalPrivilegedNesting`
+  flag
+time: 2024-10-24T15:25:29.532874059-07:00
+custom:
+  Author: sipsma
+  PR: "8776"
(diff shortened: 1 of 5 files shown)


jedevc (approved):
SGTM :tada:

I think we only need the dagger engine version for the generic nested case right - in my head, this is what we were doing the whole time :thinking: :laughing:

sipsma:
> I think we only need the dagger engine version for the generic nested case right - in my head, this is what we were doing the whole time 🤔 😆

@jedevc well we currently include the version in the cache key for all nested execs (generic and function call), this change leaves that alone. The problem was just that we also were scoping every nested exec by the `dagql.CurrentID` so that they get scoped by their args/receiver (buildkit doesn't know about any of that). But that only matters for the function call case, not generic, so this just only includes the dagql id for functions specifically now.

We are gonna hit this problem when we enable function call caching for every function too. At that time I actually don't know what we can do besides enhance dagql w/ content-based caching awareness (like we talked about the other day) 🤷‍♂️🎉

PR #8778, opened 2024-10-25
core(gitdns-fix): ensure that gitdns support token support works with all Git server providers

grouville:
Bitbucket Cloud requires a specific auth format for Git operations using tokens, where the username must be 'x-token-auth'. This fixes token authentication for Bitbucket while maintaining compatibility with other Git providers like GitHub and GitLab, and Azure.

### Context for Bitbucket Token Handling

Our implementation for Bitbucket Cloud authentication is based on the [Repository Access Tokens Documentation](https://support.atlassian.com/bitbucket-cloud/docs/using-access-tokens/), which specifically covers Git operations. While Bitbucket's [HTTP Access Tokens Documentation](https://confluence.atlassian.com/bitbucketserver/personal-access-tokens-939515499.html) suggests using Bearer authentication for some cases, this appears to only be working for to REST API calls (upon personal tests).

For Git operations, the documentation recommends using `x-token-auth` as the username, with the token as password, regardless of the token type. After testing both user tokens and repository tokens, I found that Bearer authentication did not work reliably for Git operations, leading my choice to always rely on that for bitbucket auth

### DIFF
--- a/engine/sources/gitdns/source.go
+++ b/engine/sources/gitdns/source.go
@@ -206,8 +206,9 @@ func (gs *gitSourceHandler) shaToCacheKey(sha string) string {
 }
 
 type authSecret struct {
-	token bool
-	name  string
+	token      bool
+	name       string
+	credHelper bool
 }
 
 func (gs *gitSourceHandler) authSecretNames() (sec []authSecret, _ error) {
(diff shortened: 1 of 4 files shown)


sipsma (approved):
Should add a changelog for this, otherwise LGTM

engine/sources/gitdns/source.go
```diff
@@ -247,10 +248,21 @@ func (gs *gitSourceHandler) getAuthToken(ctx context.Context, g session.Group) e
 				}
 				return err
 			}
-			if s.token {
-				dt = []byte("basic " + base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("x-access-token:%s", dt))))
+
+			if strings.Contains(gs.src.Remote, "bitbucket.org") {
```
jedevc, line 252: suggestion: maybe we should just check the host here? e.g. we don't want to do this for something weird like `git@github.com:jedevc/bitbucket.org`

engine/sources/gitdns/source.go
```diff
@@ -247,10 +248,21 @@ func (gs *gitSourceHandler) getAuthToken(ctx context.Context, g session.Group) e
...
+				// For Bitbucket Cloud Git operations, always use x-token-auth
+				// https://support.atlassian.com/bitbucket-cloud/docs/using-access-tokens/
+				gs.auth = []string{
+					"-c", fmt.Sprintf("credential.helper=!f() { echo \"username=x-token-auth\"; echo \"password=%s\"; }; f", string(dt)),
+					"-c", "credential.https://bitbucket.org.username=x-token-auth",
+					"-c", "credential.useHttpPath=true",
+				}
```
jedevc, line 259: suggestion: I would apply the logic for detecting this in `authSecretNames`, and then extend `authSecret` to have a `credhelper bool` property. This keeps the logic decoupled, but also means that we don't end up trying to put `AuthHeaderSecrets`s as passwords (we should only put the tokens there).

If this is called using an `AuthHeader`, we **shouldn't** be plugging the entire auth header in as a `password` - we should plug it in as a header as requested. It's only for the `AuthToken` that we want to apply the logic in this PR.

engine/sources/gitdns/source.go
grouville: thanks 🙏

engine/sources/gitdns/source.go
grouville: agreed, updated 🙏

Issue #8516, opened 2024-09-20
CLI: Print object state instead of all scalars


helderco:
Follow-up to:
- https://github.com/dagger/dagger/issues/6515
- https://github.com/dagger/dagger/issues/7821#issuecomment-2220684386

## Problem

When a `dagger call` chain ends in an object, the CLI selects every "no-arguments" function that returns a scalar value (primitive).

For example:

```
❯ dagger call
_type: Playground
baseImage: python:3.12-alpine
cacheKey: playground-cache
```

It’s very useful in many cases, but there’s also the potential for **unwanted side-effects**, for example with a `push` or `deploy` function that returns a string with the published address.

We even hit a case early on with core’s own `Container` object because of the `stdout` function. It doesn’t take any arguments and returns a string, but it triggers container execution which is especially bad if it’s a long running process.

There’s also the performance to consider, even in normal cases. While for some, users may appreciate that a single `dagger call` runs the linter, the tests, etc… others may not understand why the CLI is taking a long time to return something when just exploring the module’s functions in the terminal.

## Solution

Limit the functions that get printed to the simple getters that return the [object’s state](https://docs.dagger.io/manuals/developer/state-functions).

These getter functions are very fast to execute because it’s state that the engine returns directly, once it has an instance of the object, which means it doesn't have to execute the module's runtime container to get that value.

Core types don’t have this kind of distinction in their functions but we can continue to print the simple scalars, with a few hardcoded exclusions in the meantime.

shykes:
For the record, I am on board with this :)

PR #8788, opened 2024-10-25
change(cli): limit printing module objects to fields

helderco:
Fixes https://github.com/dagger/dagger/issues/8516


core/integration/module_call_test.go
```diff
@@ -1747,6 +1747,10 @@ func (t *Test) Files() []*dagger.File {
     }
 }
 
+func (*Test) Deploy() string {
+    return "here be dragons!"
+}
```
jedevc, line 1752: I might be missing something, but I can't see where we check that this property isn't included in the output.

jedevc (approved):

core/integration/module_call_test.go
helderco: Yeah, good catch!

jedevc (approved):

PR #8792, opened 2024-10-26
ci: Generate nightly workflows for alternative runners

gerhard:
They were generated in the root of the repo using the following command:

    dagger call -m .github workflows generate export --path .

The new workflows are only configured to run in the context of `dagger/dagger` repository. These runners will not be available in forks & running the same job with the `ubuntu-latest` runner doesn't make sense.

The new workflows start with `sdk-` so that it's easier to group them when looking at metrics. Before this, a special `OR` filter had to created which was making additional filtering impossible.

### DIFF
--- a/.github/dagger_runner.go
+++ b/.github/dagger_runner.go
@@ -1,0 +1,142 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+)
+
+const (
+	generation = 2
+
+	smallRunner    = 2
+	mediumRunner   = 4
+	largeRunner    = 8
+	xlargeRunner   = 16
+	xxlargeRunner  = 32
+	xxxlargeRunner = 64
+)
+
+type Runner interface {
+	Small() Runner
+	Medium() Runner
+	Large() Runner
+	XLarge() Runner
+	XXLarge() Runner
+	XXXLarge() Runner
+
+	SingleTenant() Runner
+	DaggerInDocker() Runner
+	Cached() Runner
+
+	Size() string
+	AddLabel(string) Runner
+	Labels() []string
+
+	Pipeline(string) string
+	RunsOn() []string
+}
+
+type DaggerRunner struct {
+	cores          int
+	daggerVersion  string
+	daggerInDocker bool
+	generation     int
+	labels         []string
+	singleTenant   bool
+	withCaching    bool
+}
+
+func NewDaggerRunner(
+	daggerVersion string,
+) DaggerRunner {
+	return DaggerRunner{
+		generation:    generation,
+		daggerVersion: daggerVersion,
+		labels:        []string{},
+		withCaching:   false,
+	}
+}
+
+func (r DaggerRunner) RunsOn() []string {
+	// We add size last in case the runner was customised
+	return r.AddLabel(r.Size()).Labels()
+}
+
+func (r DaggerRunner) AddLabel(label string) Runner {
+	r.labels = append(r.labels, label)
+
+	return r
+}
+
+func (r DaggerRunner) Labels() []string {
+	return r.labels
+}
+
+func (r DaggerRunner) Size() string {
+	size := fmt.Sprintf("dagger-g%d-%s-%dc",
+		r.generation,
+		strings.ReplaceAll(r.daggerVersion, ".", "-"),
+		r.cores)
+	if r.daggerInDocker {
+		size += "-dind"
+	}
+	if r.singleTenant {
+		size += "-st"
+	}
+
+	return fmt.Sprintf(
+		"${{ github.repository == '%s' && '%s' || '%s' }}",
+		upstreamRepository,
+		size,
+		defaultRunner)
+}
+
+func (r DaggerRunner) Pipeline(name string) string {
+	return name
+}
+
+func (r DaggerRunner) Small() Runner {
+	r.cores = smallRunner
+	return r
+}
+
+func (r DaggerRunner) Medium() Runner {
+	r.cores = mediumRunner
+	return r
+}
+
+func (r DaggerRunner) Large() Runner {
+	r.cores = largeRunner
+	return r
+}
+
+func (r DaggerRunner) XLarge() Runner {
+	r.cores = xlargeRunner
+	return r
+}
+
+func (r DaggerRunner) XXLarge() Runner {
+	r.cores = xxlargeRunner
+	return r
+}
+
+func (r DaggerRunner) XXXLarge() Runner {
+	// Infrastructure constraint - EC2 instance sizing
+	r.cores = xxlargeRunner
+	return r
+}
+
+func (r DaggerRunner) SingleTenant() Runner {
+	r.singleTenant = true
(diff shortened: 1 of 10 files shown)


modules/gha/main.go
```diff
@@ -674,6 +683,16 @@ func (p *Pipeline) warmEngineStep() JobStep {
...
 
+func (p *Pipeline) dockerhubAuthStep() []JobStep {
+	return []JobStep{
+		p.bashStep("dockerhub-auth", map[string]string{
+			"DOCKERHUB_USERNAME": "${{ vars.DOCKERHUB_USERNAME }}",
+			"DOCKERHUB_PASSWORD": "${{ secrets.DOCKERHUB_PASSWORD }}",
+			"DOCKERHUB_SERVER":   "${{ vars.DOCKERHUB_SERVER }}",
+		}),
+	}
+}
```
jedevc, line 694: Could we somehow allow the user to specify these?

We shouldn't assume the secrets layout from the gha module, since it's meant to be a re-usable module.

modules/gha/workflow.go
```diff
@@ -90,6 +90,7 @@ type DispatchInput struct {
 type Job struct {
 	RunsOn         []string          `json:"runs-on" yaml:"runs-on"`
 	Permissions    *JobPermissions   `json:"permissions,omitempty" yaml:"permissions,omitempty"`
+	RunIf          string            `json:"if,omitempty" yaml:"if,omitempty"`
```
jedevc, line 93: Nice :tada:

jedevc (commented):
Couple thoughts, I think we should try and keep the current `WithSDKPipelines` abstraction, some of that has gotten lost in the refactoring.

I also think we should merge https://github.com/dagger/dagger/pull/8765 first, I'll try and get to that in this week.

samalba (approved):

.github/dagger_runner.go
```diff
@@ -0,0 +1,143 @@
...
+	DaggerInDocker() Runner
+	Cached() Runner
+
+	Size() string
+	AddLabel(string) Runner
+	Labels() []string
+
+	Pipeline(string) string
+	RunsOn() []string
+}
```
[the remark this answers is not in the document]
gerhard: > I like this idea of having an interface for this, but I feel like we should split this into two interfaces? `RunnerProvider` and `Runner` maybe?

Yes, good idea for a follow-up. Right now, it's more valuable to have the metrics from the workflows that this code produces, than get this refactored.

> Why the logic for renaming `Bronze` / `Silver` / etc into `Small` / `Medium` / etc - has something changed from https://github.com/dagger/dagger/pull/8241#discussion_r1739268699

I did the simplest thing for ending up with 6 sizes, from `Small` to `XXXLarge` that everyone would understand. I didn't want to spend any time thinking about what metals would come before `Bronze`, after `Platinum` or in between. I honestly don't care what we call these as long as it's obvious that these are different sizes.

> What is `Size`? It seems to be the tag value? It seems wound up with `Labels` + `AddLabels` and `RunsOn`. Do we really need all of these methods, they all seem to just be internal to each implementation.

`Size` is the size & type of the runner (the two are coupled in some runner implementations). From that perspective you are right, it is not descriptive enough. Would `SizeAndType()` make more sense - WDYT @shykes?

.github/main.go
```diff
@@ -27,138 +27,94 @@ func New(
...
-		WithSdkPipelines("python").
-		WithSdkPipelines("typescript").
-		WithSdkPipelines("go").
-		WithSdkPipelines("java").
-		WithSdkPipelines("elixir").
-		WithSdkPipelines("rust").
-		WithSdkPipelines("php")
+		onDaggerRunners().
+		onDepotRunners().
+		onNamespaceRunners()
```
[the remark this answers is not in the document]
gerhard: OK, I'll check this out.

modules/gha/main.go
gerhard: Rather than doing that, I would much rather remove this entirely and add a guard if the file does not exist.

modules/gha/workflow.go
shykes, line 93: If the new focus is to map closely to the low-level GHA API, it's probably better to call this field `If` rather than `RunIf`? Better to avoid unnecessary editorializing in this new design focus IMO.

gerhard: Actually this may not be possible - I now remember hitting this error the first time when I also assumed that `If` would work:

```console
dagger --mod=.github call generate export --path=.
✔ connect 0.4s
✘ initialize 1.8s
! input: moduleSource.withContextDirectory.asModule resolve: failed to create module: select: failed to update codegen and runtime: failed to generate code: failed to get modified source directory for go module sdk codegen: select: process "codegen --output /src --module-context-path /src/.github --module-name ci --introspection-json-path /schema.json" did not complete successfully: exit code: 1
  ✔ resolving module ref 0.2s
  ✘ installing module 1.6s
  ! input: moduleSource.withContextDirectory.asModule resolve: failed to create module: select: failed to update codegen and runtime: failed to generate code: failed to get modified source directory for go module sdk codegen: select: process "codegen --output /src --module-context-path /src/.github --module-name ci --introspection-json-path /schema.json" did not complete successfully: exit code: 1
    ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.2s
    ✘ ModuleSource.asModule: Module! 1.4s
    ! failed to create module: select: failed to update codegen and runtime: failed to generate code: failed to get modified source directory for go module sdk codegen: select: process "codegen --output /src --module-context-path /src/.github --module-name ci --introspection-json-path /schema.json" did not complete successfully: exit code: 1

Error: input: moduleSource.withContextDirectory.asModule resolve: failed to create module: select: failed to update codegen and runtime: failed to generate code: failed to get modified source directory for go module sdk codegen: select: process "codegen --output /src --module-context-path /src/.github --module-name ci --introspection-json-path /schema.json" did not complete successfully: exit code: 1

Stdout:
generating go module: ci
Stderr:
Error: generate code: error formatting generated code: 5306:9: expected '}', found 'if' (and 10 more errors)
error: Recipe `gha` failed on line 16 with exit code 1
```

gerhard:
I am going to rebase this, fix all the issues & make it follow https://github.com/dagger/dagger/pull/8765

The thing which I care about the most right now is have those workflows run and generate metrics. If we cannot get this merged in the next few days, I am inclined to close this PR and try a different approach.

shykes:
@gerhard if you have a time constraint, but also want to find consensus with your co-maintainer :) May I suggest opening a separate PR with only the generated workflows that you need - as if they were manually generated. Then you can finish this PR without rushing it, and also get your workflows running right away.

modules/gha/workflow.go
shykes: Uh oh... I see, you hit a codegen bug! Nevermind then.

gerhard:
The alternative PR worked well in practice, closing this. We can resurrect in the future if needed.

Issue #8405, opened 2024-09-11
✨ Publish JSON Schema for dagger.json


rawkode:

drmikecrowe:
Yes, and document in website

jedevc:
We should auto-generate a schema from this struct definition here: https://github.com/dagger/dagger/blob/5ec03a604d50b721aaa65ba7c4db8c875696ac72/core/modules/config.go#L14-L43

Then we should have this as a pipeline in `.dagger/docs.go`, similar to how we generate the cli reference docs in https://github.com/dagger/dagger/blob/5ec03a604d50b721aaa65ba7c4db8c875696ac72/.dagger/docs.go#L155-L162

wingyplus:
Just poc the solution (in [repo](https://github.com/wingyplus/dagger-json-schema)). I use [json-schema](https://github.com/invopop/jsonschema) library to generate from the type in dagger. The result looks really great:

```
$ go run . 
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://github.com/dagger/dagger/core/modules/module-config",
  "$ref": "#/$defs/ModuleConfig",
  "$defs": {
    "ModuleCodegenConfig": {
      "properties": {
        "automaticGitignore": {
          "type": "boolean",
          "description": "Whether to automatically generate a .gitignore file for this module."
        }
      },
      "additionalProperties": false,
      "type": "object"
    },
    "ModuleConfig": {
      "properties": {
        "name": {
          "type": "string",
          "description": "The name of the module."
        },
        "sdk": {
          "type": "string",
          "description": "The SDK this module uses"
        },
        "include": {
          "items": {
            "type": "string"
          },
          "type": "array",
          "description": "Paths to explicitly include from the module, relative to the configuration file."
        },
        "exclude": {
          "items": {
            "type": "string"
          },
          "type": "array",
          "description": "Paths to explicitly exclude from the module, relative to the configuration file."
        },
        "dependencies": {
          "items": {
            "$ref": "#/$defs/ModuleConfigDependency"
          },
          "type": "array",
          "description": "The modules this module depends on."
        },
        "source": {
          "type": "string",
          "description": "The path, relative to this config file, to the subdir containing the module's implementation source code."
        },
        "engineVersion": {
          "type": "string",
          "description": "The version of the engine this module was last updated with."
        },
        "views": {
          "items": {
            "$ref": "#/$defs/ModuleConfigView"
          },
          "type": "array",
[...]

PR #8801, opened 2024-10-28
chore: autogenerate json schema for dagger.json

jedevc:
Fixes https://github.com/dagger/dagger/issues/8405.

This takes @wingyplus's original work from https://github.com/dagger/dagger/issues/8405#issuecomment-2375861357, and additionally applies my suggested changes from https://github.com/dagger/dagger/issues/8405#issuecomment-2399572899.

I was thinking of this in the context of https://discord.com/channels/707636530424053791/1003718839739105300/1300446424894410836, where it would good to have good-quality docs from the start for a new engine config file format.


jedevc:
@vikram-dagger where would you like to see this integrated into our docs? Should it just be a sidebar link somewhere, or maybe we could also auto-generate a `.mdx` page that contains just the JSON schema? Not really sure here, I'd mostly just like to have this available somewhere in the docs.

cmd/json-schema/main.go
```diff
@@ -0,0 +1,61 @@
...
+		if !slices.Contains(args, target.id) {
+			continue
+		}
+
+		r := new(jsonschema.Reflector)
+		if err := r.AddGoComments("github.com/dagger/dagger", "./core/modules"); err != nil {
+			return err
+		}
+
+		s := r.Reflect(&modules.ModuleConfig{})
```
wingyplus, line 32: It needs to be:

```suggestion
		if err := r.AddGoComments("github.com/dagger/dagger", target.path); err != nil {
			return err
		}

		s := r.Reflect(target.value)
```

??

jedevc: Bleh, had these changes locally, but somehow missed committing them :sweat_smile:

wingyplus (approved):
Thank you for pushing this forward. 

LGTM 👍

shykes (approved):

Issue #7787, opened 2024-06-29
engine: support for user-specified Git URL rewrite rules


sipsma:
[Git has support for rewriting URLs](https://git-scm.com/docs/git-config/#Documentation/git-config.txt-urlltbasegtinsteadOf), which can be useful for use cases like:
* Enabling a portable URL to be used that is switched based on execution environment (i.e. URL may vary from one machine to another in complex corp environments)
* [Enabling the Go toolchain to access private git repos over ssh](https://docs.gomods.io/configuration/authentication/#atlassian-bitbucket-and-ssh-secured-git-vcss) (wouldn't be surprised if there's other langs/tools rely on the same mechanism)

Dagger doesn't support this right now though. The most conceptually similar thing we support are container image registry URL aliases, which can be specified in the engine's config file, but that's very specific to pulling images obviously.

We can add support for this though:
1. The engine can automatically check for these setting's in the user's `.gitconfig` file and use them. We wouldn't even need a new session attachable I think; our filesync plumbing supports just reading individual files as byte streams that never end up in cache or anything.
2. We could add support to the engine config file too if desired, though given that currently requires custom provisioning an engine I *suspect* it's more often not as convenient as loading this from the client caller (just a hunch, could be wrong)

---

Related Discord thread: https://discord.com/channels/707636530424053791/1255946935140089936/1255946935140089936
* Use case around different URLs depending on the client

marcosnils:
I think we'll also have to make some minor changes to our `git` client fork since we're currently disabling any abilities to load configs https://github.com/dagger/dagger/blob/main/engine/sources/gitdns/cli.go#L104-L112

samalba:
Loading the user's \`.gitconfig\` would break the pipeline's portability (it would behave differently based on contextual data). Instead, I'd support it at the API level with the ability to change how the git client behaves for the whole pipeline.

sipsma:
> Loading the user's `.gitconfig` would break the pipeline's portability (it would behave differently based on contextual data). Instead, I'd support it at the API level with the ability to change how the git client behaves for the whole pipeline.

I agree that's a worthy consideration, but would argue that's actually what we want here. There's pre-existing contextual data we use from clients in the form of local dirs, local docker auth config, local services/unix-socks, etc.

[...]

PR #8805, opened 2024-10-28
feat: PAT support -- private git support on `http` / `https`

### LINKED ISSUE #7990: 🖐️ Load Dagger module from private Github repository via PAT auth.
### What happened? What did you expect to happen?

We tried to load Dagger module from our private Github repository but failed. We'd like to know whether this capability supported now? Or is there any plan to support this in the future? By the way, our company only allows PAT auth to do this.

grouville:
Fixes #7990

This is the PR introducing PAT support auth.

## Implementation
As requested on the issue by the community, we rely on the git credential to request for authentication:
- this allows us to maintain the same auth experience as your local `git` CLI
- When needed, the engine requests the host for the `git-credential` 
- Your local gitconfig sets up the git credential auth you want

This gives userland the power to configure auth however they want, with their preferred auth tool

## How does this work
`git` introduces a standard: [git credential](https://git-scm.com/docs/git-credential) which is queryable from the `git` CLI: 

```shell
echo -e "protocol=https\nhost=github.com\npath=dagger/dagger\n" | git credential fill
```

The engine is directly querying this command on the user's host, via the CLI and parses the output of the command ; it then injects it back inside its session, safely, as a secret

## How can you test
Please break this PR as you would, especially try to see if this maps your usual auth pattern (personal and enterprise environment): 
```shell
$ cd ~/dagger
$ ./hack/dev # wait until it finishes
$ ./hack/with-dev bash # jump into the dev version shell
$ dagger version $ shall output dagger v0.13.7-...
$ dagger -m private-module functions
```

### git credential managers

There are a lot of [git credential managers](https://github.com/git-ecosystem/git-credential-manager) out there, all mapping with the `git credential` standard. They can also be chained, and this will allow you to manage the caching side of those tokens client-side

TODO: 
- [x] Finish unit and integration test
- [x] Some user testing to see any edge case I might have missed
- [ ] Document the feature

cc @nipuna-perera @sheldonhull


raghavendramallela:
There are other configs in `.gitconfig` too…. like `http.postBuffer` `https.proxy` etc… can these be included?

grouville:
> There are other configs in .gitconfig too…. like http.postBuffer https.proxy etc… can these be included?

I think this is more related to https://github.com/dagger/dagger/issues/7787 ? What you want is to propagate those .gitconfig inside the engine, from what I understand ? What use do you have of those options ?

raghavendramallela:
hi @grouville , if i query:
```
dagger core git --url privgit.com -vvvvvv
✔ connect 0.4s
  ✔ starting engine 0.0s
  ✔ connecting to engine 0.1s
  ✔ starting session 0.1s
  ✔ subscribing to telemetry 0.1s
    ✔ consuming /v1/traces 0.1s
    ✔ consuming /v1/logs 0.0s
    ✔ consuming /v1/metrics 0.0s
✔ initialize 0.4s
  ✔ cache request: mkfile /schema.json 0.0s
  ✔ load cache: mkfile /schema.json 0.0s
  ✔ cache request: blob://sha256:fa3709c3d62221111c45ca4bbe0d3a942e18fae76e0bfd225694b40480d40d88 0.0s
  ✔ load cache: blob://sha256:fa3709c3d62221111c45ca4bbe0d3a942e18fae76e0bfd225694b40480d40d88 0.0s
  ✔ blob(digest: "sha256:fa3709c3d62221111c45ca4bbe0d3a942e18fae76e0bfd225694b40480d40d88", mediaType: "application/vnd.oci.image.layer.v1.tar", size: 167936, uncompressed: "sha256:fa3709c3d62221111c45ca4bbe0d3a942e18fae76e0bfd225694b40480d40d88"): Directory! = xxh3:71f907b752f0adac 0.0s
  ✔ Directory.file(path: "schema.json"): File! = xxh3:ffef5cf427d85037 0.0s
✔ prepare 0.0s
✔ git(url: "privgit.com"): GitRepository! = xxh3:0321a4028fb013f5 0.0s
  ✔ host: Host! = xxh3:ff73c85214b8f75c 0.0s
  ✔ Host.__internalSocket(accessor: "2f746d702f7373682d383254536f633846494149712f6167656e742e313537363539b613679a0814d9ec772f95d778c35fc5ff1697c493715653c6c712144292c5ad"): Socket! = xxh3:868707ad72e3f5af 0.0s
✘ GitRepository.tags: [String!]! = xxh3:a4bc5136ba5c3a61 10.0s
! git command failed: exit status 128

Error: response from query: input: git.tags resolve: git command failed: exit status 128
stdout: 
stderr: fatal: unable to access 'https://privgit.com/': Could not resolve host: privgit.com
```
but, `privgit.com` is resolvable via `HTTPS_PROXY` env var - git cli, or `https.proxy` in `.gitconfig` .
PS: i have initiated the engine container with the proxy variable `HTTPS_PROXY` already:
```
podman exec -it dengine env | grep HTTPS_PROXY
HTTPS_PROXY=http://proxyserver.com:443
```

sheldonhull:
I normally install the binary via Aqua. 
for testing this PR out directly do I run go install daggerrepo@branchname and or is there a doc with best approach?

grouville:
> I normally install the binary via Aqua. for testing this PR out directly do I run go install daggerrepo@branchname and or is there a doc with best approach?

Hi @sheldonhull,
No, you need to checkout this PR locally and build the dev version of the engine. I explained how to do it above, in the `How can you test` section. But, if you have any issue, I have published the engine to `ghcr.io`. Nipuna cannot build the dev version (https://discord.com/channels/707636530424053791/1300550470783205478/1300550473194930260), so I published it to get his feedback

core/schema/git.go
```diff
@@ -120,62 +125,141 @@ func (s *gitSchema) git(ctx context.Context, parent *core.Query, args gitArgs) (
...
+				if err == nil {
+					// Credentials found, create and set auth token
+					var secretAuthToken dagql.Instance[*core.Secret]
+					if err := s.srv.Select(ctx, s.srv.Root(), &secretAuthToken,
+						dagql.Selector{
+							Field: "setSecret",
+							Args: []dagql.NamedInput{
+								{
+									Name:  "name",
+									Value: dagql.NewString("gitAuthtoken"),
```
vito, line 244: Not sure if this is still the case, but iirc these secret names are session-wide. So if you use two separate private repos in the same session it seems like these might clobber. Maybe just set a name that's a hash of the plaintext, like we do for the CLI?

engine/sources/gitdns/source.go
```diff
@@ -206,26 +206,33 @@ func (gs *gitSourceHandler) shaToCacheKey(sha string) string {
...
+	credHelper bool
 }
 
 func (gs *gitSourceHandler) authSecretNames() (sec []authSecret, _ error) {
 	u, err := url.Parse(gs.src.Remote)
 	if err != nil {
 		return nil, err
 	}
+
+	isBitbucket := u.Host == "bitbucket.org"
```
vito, line 220: Do we need to worry about any other hosts, like we do with GitHub enterprise? 🤔

vito (commented):
Overall approach and implementation seems great 👍 - just a couple of comments, otherwise lgtm!

engine/session/gitcredential.proto
```diff
@@ -0,0 +1,40 @@
...
+message GitCredentialRequest {
+  string protocol = 1;
+  string host = 2;
+  string path = 3; // optional
+}
+
+message GitCredentialResponse {
+  oneof result {
+    CredentialInfo credential = 1;
+    ErrorInfo error = 2;
```
aluzzardi, line 18: nit (can be changed later): I **think** the gRPC way of doing this is to return an actual error from the function, rather returning `nil` as an error with a custom `error` field in the response

engine/sources/gitdns/source.go
```diff
@@ -247,9 +255,23 @@ func (gs *gitSourceHandler) getAuthToken(ctx context.Context, g session.Group) e
 				}
 				return err
 			}
+
+			// Handle token authentication
 			if s.token {
+				if s.credHelper {
```
aluzzardi, line 261: nit: is there a way to get this from the client config instead?

as @vito mentioned above, we might have a problem with GH enterprise as well

engine/sources/gitdns/source.go
sheldonhull: my host name could be different for azure devops based on a combination of the protocol with http or ssh, as well as the way dev.azure.com and visualstudio.com are both valid addresses (due to legacy url).

core/schema/git.go
grouville: @sipsma, are these session-wide ?

core/schema/git.go
```diff
@@ -120,62 +125,141 @@ func (s *gitSchema) git(ctx context.Context, parent *core.Query, args gitArgs) (
...
+		}
+
+		if remote.Scheme == "https" || remote.Scheme == "http" {
+			// Check if repo is public
+			repo := git.NewRemote(memory.NewStorage(), &config.RemoteConfig{
+				Name: "origin",
+				URLs: []string{remote.Remote},
+			})
+
+			_, err := repo.List(&git.ListOptions{Auth: nil})
```
sipsma, line 211: nit: should probably use `ListContext` so we can cancel the request if the call is canceled

engine/sources/gitdns/source.go
grouville: Mmh, I do agree that we could miss some edge cases, but here it was changes that were part of https://github.com/dagger/dagger/pull/8778. Basically, the core git implementation was not passing the PAT properly, as bitbucket has another way of passing the auth tokens.

We could have some edge cases on github enterprise and some azure, but I added some tests https://github.com/grouville/dagger/blob/a924e0e9b054ece63086badc9c84142cd0bb11d0/core/integration/git_test.go. I tested azure, and github enterprise has already being used by users. I do think that this should be ok ; if it breaks we should then extend I agree 👍

core/schema/git.go
sipsma: Secrets are stored per-client so that sort of clobbering shouldn't be an issue.

The only extremely obscure case I can think of that would be impacted is:
1. You're using an SDK directly on the host (i.e. not just `dagger call`)
2. You set your own secret that also happens to be named `gitAuthtoken`

Not worth agonizing over, but yeah we could just set name to a sha256 of the plaintext to squash any possibility of that happening.

core/integration/gitcredential_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+type GitCredentialSuite struct{}
+
+func TestGitCredential(t *testing.T) {
+	testctx.Run(testCtx, t, GitCredentialSuite{}, Middleware()...)
+}
+
+// TestGitCredentialErrors verifies Git authentication for private modules across different providers
+// using host-specific PATs and isolated Git credentials per test.
+func (GitCredentialSuite) TestGitCredentialErrors(ctx context.Context, t *testctx.T) {
```
sipsma, line 24: Could we get one more test case that's similar to the existing ones except that it calls a module and passes a `Directory` as an arg that's from a git repo? e.g. something like:

```
dagger call test --dir https://github.com/grouville/daggerverse-private.git
```

and make sure ops on the git dir work as expected in the module (can list the dir or something like that)?

That's just one tricky case where the resolution with git needs to happen within the CLI+auth-token and after should be usable from modules even if they don't have access to the auth provider directly.

sipsma (commented):
Looking good so far!

sipsma (approved):
🚀

grouville:
> realized

Doing it rn 🙏

core/schema/git.go
```diff
@@ -103,88 +109,183 @@ type gitArgs struct {
 	SSHAuthSocket dagql.Optional[core.SocketID] `name:"sshAuthSocket"`
 }
 
-func (s *gitSchema) git(ctx context.Context, parent *core.Query, args gitArgs) (*core.GitRepository, error) {
+func (s *gitSchema) git(ctx context.Context, parent dagql.Instance[*core.Query], args gitArgs) (inst dagql.Instance[*core.GitRepository], _ error) {
```
jedevc, line 112: @grouville suggestion for a follow-up - is there a way to break down some of this logic into separate functions, maybe even into `core/git.go` itself?

The comments here *really* help, but if we could use function calls to help show the structure that would be even better :tada:

grouville: Yeah sure 👍 I saw a lot of monolithic functions, so wasn't sure ; but 👌

PR #8237, opened 2024-08-26
chore: update TS breaking change deps

TomChv:
Bump `graphql-request` to `7.1.0` which drop `CommonJS` supports. This introduce a breaking change for the TS runtime that also needs to drop `CommonJS` support.

Bump `execa` to `9.3.1` which requires to bump Bun to `v1.1.26` that includes support for underlying engine
call.

⚠️ This PR needs to be merge for `v0.13.0` since it introduces some breaking changes (it might not impact users though)

Follow up of #8113.

PR #8811, opened 2024-10-29
docs: use new `withExec` `expect` arg

jedevc:
Adds docs for features introduced in https://github.com/dagger/dagger/pull/8466.


docs/current_docs/cookbook/snippets/continue-after-errors/python/main.py
```diff
@@ -27,9 +27,9 @@ async def test(self) -> TestResult:
                 dag.container()
                 .from_("alpine")
                 # add script with execution permission to simulate a testing tool.
-                .with_new_file("run-tests", SCRIPT, permissions=0o750)
-                # if the exit code isn't needed: "run-tests; true"
-                .with_exec(["sh", "-c", "/run-tests; echo -n $? > /exit_code"])
+                .with_new_file("/run-tests", SCRIPT, permissions=0o750)
+                # run-tests but allow any return code
+                .with_exec(["/run-tests"], expect=ReturnType.ANY)
```
helderco, line 32: For consistency (remove the import too):
```suggestion
                .with_exec(["/run-tests"], expect=dagger.ReturnType.ANY)
```

vikram-dagger:
Can you clarify why the TS example was moved to a `src` subdir?

Also, AFAIK we shouldn't need the `dagger.json` for the Python and TS snippets, but we do need the `go.*` files for the Go snippet. See https://github.com/dagger/dagger/tree/main/docs/current_docs/cookbook/snippets/copy-filter-dir for a recent example of this structure.

Note: I'm aware that the above might (and should) kick off a separate discussion about the structure of our snippets, which might require some global changes to all snippets. I'm happy to do that in a separate discussion/PR to avoid blocking this PR, approving this one in advance so we can merge and then iterate further.

vikram-dagger (approved):

docs/current_docs/cookbook/snippets/continue-after-errors/typescript/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+   * Handle errors
+   */
+  @func()
+  async test(): Promise<TestResult> {
+    const ctr = await dag
+      .container()
+      .from("alpine")
+      // add script with execution permission to simulate a testing tool.
+      .withNewFile("/run-tests", SCRIPT, { permissions: 0o750 })
+      // if the exit code isn't needed: "run-tests; true
```
sipsma, line 33:
```suggestion
      // run-tests but allow any return code
```

sipsma (approved):
@vikram-dagger I'm going to merge this as is so we can get it for the release on Friday, just FYI for the other follow-up discussions you mentioned.

jedevc:
> Can you clarify why the TS example was moved to a `src` subdir?

Ah, so this is because it's now required to have the code there for the module to build (I *assume* we want that) - see https://github.com/dagger/dagger/pull/8237#issuecomment-2343097603. I seem to remember I needed to move it to get the module example to work.

> Note: I'm aware that the above might (and should) kick off a separate discussion about the structure of our snippets, which might require some global changes to all snippets.

I would really like it if all the `docs` snippets were runnable. Imagine if you could click a snippet and get a copy-pastable command to run it in the cli. Or as a shell. Or even to open in a future dagger-cloud to play around.

But if we want that, then I think some of the snippets need tidying up, and we would need validation to ensure that they *keep* working.

vikram-dagger:
> > Can you clarify why the TS example was moved to a `src` subdir?
> 
> Ah, so this is because it's now required to have the code there for the module to build (I _assume_ we want that) - see [#8237 (comment)](https://github.com/dagger/dagger/pull/8237#issuecomment-2343097603). I seem to remember I needed to move it to get the module example to work.

Was there a specific error you saw? I'm asking because we don't have a `src` subdir for other TS snippets, so this would be a new practice. To clarify, I'm not against doing this, just trying to understand if it's necessary and if we should replicate this globally to all other TS snippets.

> But if we want that, then I think some of the snippets need tidying up, and we would need validation to ensure that they _keep_ working.

I would definitely like to do this, maybe we can discuss this in a separate thread or in Discord? From past discussions, my understanding is that the structure and files needed for each snippet differ according to their language, but we do not yet have a conclusive checklist for each case.

jedevc:
> Was there a specific error you saw?

```
❯ ls
dagger.json  index.ts
❯ cat dagger.json
{
  "name": "my-module",
  "sdk": "typescript",
  "source": ".",
  "engineVersion": "v0.13.7"
}
❯ dagger functions
✔ connect 0.2s
✔ loading module 15.6s

Name             Description
```

To actually get this to work, I had to `mkdir src && mv index.ts src`.

cc @TomChv just to check - is my understanding correct here? I can't find a release note that documents this as a breaking change.

PR #8815, opened 2024-10-29
feat: identify and set Harness metadata

levlaz:
This commit is modeled after what we did in #7835 for Jeknins. It adds support for Harness in Dagger Cloud with the same caveats.

I chose HARNESS_ACCOUNT_ID based on the available default env vars: https://www.notion.so/dagger-io/Default-Harness-Env-12e206f38800807e8b46fd9846a4c18f


jedevc:
:thinking: we also have a bunch of `DRONE_` args set in the notion page.

What's the relationship between the products/companies here? Do we want to detect drone or harness? Both?

jedevc (approved):

levlaz:
> 🤔 we also have a bunch of `DRONE_` args set in the notion page.
> 
> What's the relationship between the products/companies here? Do we want to detect drone or harness? Both?

DRONE = Harness CI, its the same product now. If someone is somehow running drone without harness and wants it to show up dagger cloud, then we should revisit.

PR #8818, opened 2024-10-30
Filesync performance improvements

sipsma:
## Goals
Primary:
1. Improve caching behavior of local filesyncs; re-use cache of already synced files in more cases
2. Use the above to simplify module context/source loading + reduce unneeded cache invalidation of module builds
   * Splitting this out to a separate follow-up PR since this one is big enough on its own, but it's all setup here
3. Take control of the filesync implementation to enable future support for more features in the shell, live filesync, etc.

Secondary:
1. Improve performance of filesync in all cases (memory+CPU usage in the engine)
2. Explicitly identify and error out when concurrent changes on the client filesystem would result in inconsistent loads (previously was just undefined behavior essentially)

## Performance Improvements
I used the [git repo for the Linux kernel](https://github.com/torvalds/linux) as a stress-test case since it contains multiple GB of source files.

The main improvements are around filesyncs of already loaded data and parallel syncs, but there's some meaningful improvement to uncached syncs too for huge filesync cases.

The improvements in cache re-use and parallel syncs are expected to be most applicable to module loading and contextual directory argument usage.

* Uncached load of full repo except `.git` (`.host | directory . --exclude .git | entries`)
   * **v0.14.0** - 44.6s, ~1.3GB RSS ([trace](https://dagger.cloud/dagger/traces/b72f5fb06ff567be33d047cb10b1b77c?span=ff55914850afc762))
   * **This PR** - 29.7s, ~650MB RSS ([trace](https://dagger.cloud/dagger/traces/a56b8d4e1152eccf63295080471ffc72?span=e1b6a25d6d5ce4e4))
   * ~33% faster and 50% less memory

* Subsequent load of `drivers/` subdir (`.host | directory drivers | entries`)
  * **v0.14.0** - 20.2s ([trace](https://dagger.cloud/dagger/traces/b72f5fb06ff567be33d047cb10b1b77c?span=02c4d4c97a0446a7))
   * **This PR** - 4.9s ([trace](https://dagger.cloud/dagger/traces/a56b8d4e1152eccf63295080471ffc72?span=a70f513c9afffa70))
   * ~75% faster
   * The `drivers/` subdir was already loaded in the previous sync so there shouldn't be any loading of new data. The old implementation didn't handle this, but the new one does. You can see in the traces that the majority of the time in the new implementation is spent copying files/dirs (in order to make an immutable cache entry); this is also possible to improve in the future too.

* Cached load of full repo except `.git` using include instead of exclude (`.host | directory . --include **,!.git | entries`)
   * **v0.14.0** - 22.5s ([trace](https://dagger.cloud/dagger/traces/b72f5fb06ff567be33d047cb10b1b77c?span=dfe7df3936137b64))
   * **This PR** - 1.9s ([trace](https://dagger.cloud/dagger/traces/a56b8d4e1152eccf63295080471ffc72?span=6fe78f3345ef82f5))
   * ~91% faster
   * This case is especially silly; we are re-running the same load as before but using include settings instead of exclude settings, so there should be nothing to sync or copy. Previous implementation didn't realize this though.

* Re-run full repo load from separate parallel session (`.host | directory . --exclude .git | entries`)
   * **v0.14.0** - 21.8s ([trace](https://dagger.cloud/dagger/traces/3dc4f1dda86ea4ce8564c47699a145ea?span=64f9adb45de46174))
   * **This PR** - 1.8s ([trace](https://dagger.cloud/dagger/traces/8481833b0125ac78601d5deeac0397da?span=212c1fbbab0948c5))
   * ~91% faster
   * The previous implementation re-ran a lot of equivalent work when it happened from different clients/sessions. Concurrency across clients/sessions is handled now, which allows more work to be re-used.

## Implementation Details
Biggest problems with previous implementation:
1. Buildkit would create a new cache ref for every "name" of a provided local source
   * The name we set was derived from the path being loaded, so each path on the client's filesystem would get its own ref
   * We also couldn't use a single name and sync in different dirs under it because buildkit's cache logic requires local cache refs be locked while in use; if one is locked it just creates a brand new one (which was hard to predict and control and thus would result in lots of mysterious uncached syncs). That would have gotten much worse if we tried to maintain a single cache ref for each sync from a client.
   * Now, there's a single cache ref for each client (as identified by their stable client ID) for their entire rootfs. Each sync of an individual dir just fills the relevant part of the cache ref. Concurrency is handled allowing different syncs to safely operate on the same underlying cache ref.
   * Additionally, if the client changed files in the middle of sync previously, it was undefined what would happen. Now the implementation explicitly notices these situations and returns an error.
2. Making local syncs immutable required exporting files into the content store, which had a lot of overhead
   * We have always needed to include an extra step after a sync to copy data and make it immutable. This became our `blob` API and went through the content store for simplicity. Going through the content store involves a ton of overhead and extra copying though. It also created more situations in which the local cache sync ref was locked and would thus result in more mysterious uncached syncs.
   * Now, the copy step that creates an immutable ref is part of the filesync implementation. It uses the `contenthash` lib from upstream to calculate and store content hash metadata rather than going through the content store. It also re-uses existing equivalent refs whenever possible.

There's a ton of other little optimizations in various places (noted most of them with comments in the code).

## Future improvements and features
* Handle nested clients better?
   * Nested clients all get a brand new random stable client ID, so they don't benefit from cache reuse a ton. This is because the COW semantics of `withExec` layers complicate cache re-use quite a bit (you can end up with different filesystems that have the same stable client ID)
   * Handling this would mostly be an improvement to our integ tests (which is likely the heaviest user of nested clients by far), so may investigate this angle when looking into improving test performance
* The overhead of the copy step could be lessened by maintaining a "hardlink farm" of files stored by their content hash and just linking in the files rather than copying them on disk. 
   * In the example above, copying the 1.1GB `drivers/` subdir added ~5s of overhead, to give an idea of how much of a difference this might be able to make (not enough to be worth it here, but possibly worth it someday)
* We can enable options to force reloads of data rather than caching given ops permanently within the context of the session. This is needed for `dagger shell` use cases
* We're in more control of the filesync low-levels now, which may help "live" filesync use cases being discussed.


sipsma:
Need an upstream fix for some (potentially all?) of the remaining test failures: https://github.com/moby/buildkit/pull/5521

core/directory.go
```diff
@@ -741,7 +740,7 @@ func (dir *Directory) Export(ctx context.Context, destPath string, merge bool) (
 	}
 
 	var defPB *pb.Definition
-	if dir.Dir != "" {
+	if dir.Dir != "" && dir.Dir != "/" {
```
sipsma, line 743: not technically related to filesync, just noticed a bunch of extra `copy / /` steps when running exports and fixed that here

engine/server/session.go
```diff
@@ -1044,6 +1044,11 @@ func (srv *Server) serveSessionAttachables(w http.ResponseWriter, r *http.Reques
 	}
 
 	ctx = client.daggerSession.withShutdownCancel(ctx)
+
+	// Disable collecting otel metrics on these grpc connections for now. We don't use them and
+	// they add noticeable memory allocation overhead, especially for heavy filesync use cases.
+	ctx = trace.ContextWithSpan(ctx, trace.SpanFromContext(nil)) //nolint:staticcheck // we have to provide a nil context...
```
sipsma, line 1050: cc @vito, just want to confirm we are for sure not using any of these spans from buildkit session stuff

engine/sources/local/localfs.go
```diff
@@ -0,0 +1,808 @@
...
+func (local *localFS) toFullPath(path string) string {
+	return filepath.Join(local.rootPath, local.subdir, path)
+}
+
+// the absolute path under local.rootPath
+func (local *localFS) toRootPath(path string) string {
+	return filepath.Join(local.subdir, path)
+}
+
+// GetPreviousChange is called when the differ indentifies that our cache and the client's cache match at this path.
```
sipsma, line 407:
```suggestion
// GetPreviousChange is called when the differ identifies that our cache and the client's filesystem match at this path.
```

engine/sources/local/singleflight.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	"context"
+	"sync"
+)
+
+// SingleflightGroup is similar to sync.Singleflight but:
+// 1. Handles context cancellation (ctx provided to callback is only cancelled once no one is waiting on the result)
+// 2. Caches results until all returned CachedResults are released
+type SingleflightGroup[K comparable, V any] struct {
```
sipsma, line 11: Note for reviewers: this is the heart of all the tricky concurrency handling we needed here. We use paths in the filesystem being synced as keys and run operations on the paths in the callback to `Do`.

Part of the utility is that equivalent operations on the same path from different concurrent syncs are deduped and don't stomp over each other.

The fact that the results are cached helps with deduping and also allows the code to identify the client filesystem is changing during the sync and error out (see `verifyExpectedChange` and its usage). If the returned result was cached from a previous operation and it doesn't match what we expected to apply, we know something changed in the middle of the sync.

The other important part is the behavior where results are cached until all returned references to them are `.Release()`'d. This allows us to essentially "lock" individual paths as they are referenced in a given sync but release them when we're done.

engine/sources/local/diff.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	return len(p1) - len(p2)
+}
+
+func min(x, y int) int {
+	if x < y {
+		return x
+	}
+	return y
+}
```
vito, line 178: nit: I think this is a Go builtin now?

engine/sources/local/localfs.go
```diff
@@ -0,0 +1,808 @@
...
+					if err := cacheCtx.HandleChange(appliedChange.Result().kind, path, appliedChange.Result().stat, nil); err != nil {
+						return fmt.Errorf("failed to handle change in content hasher: %w", err)
+					}
+				}
+
+				return nil
+
+			case upperStat.Mode&uint32(os.ModeDevice) != 0 || upperStat.Mode&uint32(os.ModeNamedPipe) != 0:
+				// NOTE: not handling devices for now since they are extremely non-portable and dubious in terms
+				// of real utility vs. enabling bizarre hacks
```
vito, line 167: Maybe worth a log? 🤔

engine/sources/local/localfs.go
```diff
@@ -0,0 +1,808 @@
...
+		}
+	}
+
+	if forParents {
+		// we created the parent dirs, nothing else to do now
+		return nil, nil
+	}
+
+	ctx, copySpan := newSpan(ctx, "copy")
+	defer copySpan.End()
```
vito, line 275: nit: would recommend using `telemetry.End(copySpan, func() error { return rerr })` pattern so the span fails if an error occurs from this point onward

engine/buildkit/blob.go
```diff
@@ -5,109 +5,90 @@ import (
...
+		md := contenthash.CacheRefMetadata{RefMetadata: ref}
+		dgst, ok := md.GetContentHashKey()
+		if ok {
+			bklog.G(ctx).Debugf("DefToBlob reusing ref %s with digest %s", ref.ID(), dgst)
+			return dgst, nil
+		}
 
-	desc = remote.Descriptors[0]
+		ctx, span := Tracer(ctx).Start(ctx, fmt.Sprintf("checksum def: %s", ref.ID()), trace.WithSpanKind(trace.SpanKindInternal))
+		defer span.End()
```
vito, line 64: nit: would recommend using `telemetry.End` helper here too

engine/buildkit/blob.go
```diff
@@ -5,109 +5,90 @@ import (
...
-	}
+		md := contenthash.CacheRefMetadata{RefMetadata: ref}
+		dgst, ok := md.GetContentHashKey()
+		if ok {
+			bklog.G(ctx).Debugf("DefToBlob reusing ref %s with digest %s", ref.ID(), dgst)
+			return dgst, nil
+		}
 
-	desc = remote.Descriptors[0]
+		ctx, span := Tracer(ctx).Start(ctx, fmt.Sprintf("checksum def: %s", ref.ID()), trace.WithSpanKind(trace.SpanKindInternal))
```
vito, line 63: I'm seeing this show up in [somewhat surprising places](https://v3.dagger.cloud/dagger/traces/6abb5fd15689b362fecd189df9109a27?span=a6f2960eec708828)

That `trace.WithSpanKind(trace.SpanKindInternal)` probably isn't doing what you want - that's some other part of OTel that we don't overload for UI purposes since all spans are actually "KindInternal" by default.

```suggestion
		ctx, span := Tracer(ctx).Start(ctx, fmt.Sprintf("checksum def: %s", ref.ID()), telemetry.Internal())
```

vito (approved):
neat! just a few nits

core/schema/host.go
```diff
@@ -38,29 +37,13 @@ func (s *hostSchema) Install() {
 		}).Doc(`Queries the host environment.`),
 
 		dagql.Func("blob", func(ctx context.Context, parent *core.Query, args struct {
-			Digest       string `doc:"Digest of the blob"`
-			Size         int64  `doc:"Size of the blob"`
-			MediaType    string `doc:"Media type of the blob"`
-			Uncompressed string `doc:"Digest of the uncompressed blob"`
```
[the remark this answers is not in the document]
sipsma: Yeah, I'm hoping `blob` can be rm'd entirely someday by adding content-digest awareness to dagql IDs themselves, but for another time :-)

core/schema/sdk.go
```diff
@@ -641,10 +641,6 @@ func (sdk *goSDK) baseWithCodegen(
 						"codegen",
 					}, codegenArgs...),
 				},
-				{
-					Name:  "experimentalPrivilegedNesting",
-					Value: dagql.NewBoolean(true),
-				},
```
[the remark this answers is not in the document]
sipsma: I *think* it was that it used to be necessary, then became unnecessary at some point but this got left in until now.

dagql/idtui/testdata/TestTelemetry/TestGolden/viztest/broken/broken
```diff
@@ -12,6 +12,8 @@ Expected stderr:
 │ ! input: module.withSource.initialize failed to initialize module: failed to call module "broken" to get functions: call constructor: process "go build -ldflags -s -w -o /runtime ." did not complete successfully: exit code: 1
 │ │ $ ModuleSource.resolveFromCaller: ModuleSource! X.Xs CACHED
 │ │ │ ✔ upload /XXX/XXX/XXX from XXXXXXXXXXX (client id: XXXXXXXXXXX, session id: XXXXXXXXXXX) (include: XXXXXXXXXXX) (exclude: XXXXXXXXXXX) X.Xs
+│ │ │ │ ✔ filesync X.Xs
+│ │ │ │ │ ✔ copy X.Xs
```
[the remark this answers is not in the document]
sipsma: What were you thinking of in terms of further improvements, just more spans? I added these mostly because they are useful for knowing why a filesync is slow, but they seemed useful enough for users potentially that they may as well be non-internal.

> Side note: is the (client id: XXXXXXXXXXX, session id: XXXXXXXXXXX) part still useful?

I'd say yes, though admittedly mostly for debugging as a dagger developer, I don't think they are useful/meaningful to end users.

engine/sources/local/diff.go
sipsma: True, will use that

engine/sources/local/localfs.go
sipsma: Yeah definitely, will add

engine/buildkit/blob.go
sipsma: Oh okay good to know, I just copy-pasta'd this from some other part of the codebase (parseRefString I think) Will fix

engine/server/session.go
sipsma: Oh yeah it's confusing, having a span in the context triggered this code in buildkit: https://github.com/moby/buildkit/blob/c234dd97ba77033c55b844e59b4ab1122e7977f1/session/grpc.go#L51-L57

But it sets up a grpc StatsHandler which ends up collecting otel metrics.

Not sure if there's a reason span+metrics are tied together this way, but we don't want any of it so 🤷‍♂️

dagql/idtui/testdata/TestTelemetry/TestGolden/viztest/broken/broken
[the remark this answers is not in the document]
vito: > What were you thinking of in terms of further improvements, just more spans?

Oh nothing in particular, these seem OK as-is - I was just wondering things like 'should more info be in the name' and 'do we need two spans' - but eyeballing it irl, there's a clear time gap between the `filesync` and the `copy` starting, so it seems worth the breakdown.

helderco:
When doing a `dagger functions` in my local dagger/dagger clone, [I get this](https://v3.dagger.cloud/Correia/traces/265a697f6615635c7203f0dba425b5ad?listen=1b957c9883c2c799&listen=7b5d41467b47f207&listen=d5199c96e75b68cf):
```
│ │ │ ✘ upload /Users/helder/projects/dagger from hal2h27jjpkdly619pkxpnu7a (client id: 5hi0yiyaqsnmgnxfnbv4vwwzi, session id: vtnnzkwszc33xev528mgna510) (include: dagger.json, .dagger/**/*, modules/alpine/dagger.json, modules/alpine/**/*, modules/go/dagger.json, modules/go/**/*, version/dagger.json, version/**/*, modules/ps-analyzer/dagger.json, modules/ps-analyzer/**/*, sdk/python/dev/dagger.json, sdk/python/dev/**/*, cmd/dagger/dagger.json, cmd/dagger/.dagger/**/*, modules/dirdiff/dagger.json, modules/dirdiff/**/*, modules/wolfi/dagger.json, modules/wolfi/**/*, modules/golangci/dagger.json, modules/golangci/**/*, sdk/php/dev/dagger.json, sdk/php/dev/**/*, modules/shellcheck/dagger.json, modules/shellcheck/**/*, sdk/elixir/dev/dagger.json, sdk/elixir/dev/**/*, modules/graphql/dagger.json, modules/graphql/**/*, **/go.mod, **/go.sum, **/go.work, **/go.work.sum, **/vendor/, **/*.go) (exclude: **/.git, sdk/python/dev/**/.venv, sdk/python/dev/**/__pycache__, sdk/python/dev/**/.*cache) 0.0s
│ │ │ ! failed to get snapshot: failed to sync: failed to create local fs: failed to create base fs: resolve : lstat /var/lib/dagger/worker/snapshots/snapshots/83380/fs/Users: no such file or directory
│ │ │ │ ✔ filesync 0.0s
Error: failed to serve module: input: moduleSource.resolveFromCaller failed to import local module source: failed to get snapshot: failed to sync: failed to create local fs: failed to create base fs: resolve : lstat /var/lib/dagger/worker/snapshots/snapshots/83380/fs/Users: no such file or directory
```

core/schema/host.go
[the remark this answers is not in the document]
jedevc: Just noting - this is a backwards incompat change to this API - which is totally fine, it's all internal anyways.

But, while we're here, maybe we could rename it to `__blob`? That way we won't get any codegen of it at all. I think we named it `blob` before we starting using `__` prefixes as convention, which we do in other internal APIs, like `__withSystemEnvVariable`.

core/modfunc.go
```diff
@@ -376,9 +371,10 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Typ
 		}
 	}
 
-	if err := fn.linkDependencyBlobs(ctx, result, returnValueTyped); err != nil {
-		return nil, fmt.Errorf("failed to link dependency blobs: %w", err)
-	}
+	// NOTE: once generalized function caching is enabled we need to ensure that any non-reproducible
```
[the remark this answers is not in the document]
jedevc: Sounds good - note, that also this wasn't working in some cases anyways, because of the weird way that private inputs work: https://github.com/dagger/dagger/pull/8358#issue-2510229992

engine/sources/local/identifier.go
jedevc, line 1: Can we keep using upstream here? This file doesn't seem to have been changed.

engine/sources/local/singleflight.go
jedevc: So, this is part of the `localFSSharedState`. So, multiple mounters of the big shared local state don't run over each other.

But just to confirm - that means all operations for this now need to go through this group to avoid accidentally running over each other? Not really a concern, just wanting to check my understanding, though it does mean that we wouldn't be able to have multiple dagger engines accessing local cache refs from the same store (though I guess there are significantly more complicated blockers for this anyways).

jedevc (approved):
Code changes LGTM, the new local source looks *horribly* complicated, but most of it is pulled from `fsutil`, so it's really just the same complexity as before :tada:

Only thing I'd say is that we should maybe have a couple comments on structs like `localFS` and `localFSSharedState`.

jedevc:
I do wonder how the big shared mirror of the user filesystem works with windows :thinking: Hopefully the paths all make sense.

sipsma:
> When doing a dagger functions in my local dagger/dagger clone, [I get this](https://v3.dagger.cloud/Correia/traces/265a697f6615635c7203f0dba425b5ad?listen=1b957c9883c2c799&listen=7b5d41467b47f207&listen=d5199c96e75b68cf):

@helderco Thanks for trying and catching this!

Can you confirm that somewhere in the path you were loading there's a symlink? IIRC `/Users` is often a symlink on macos, but on my macbook it isn't, so not sure if I'm misremembering or what.

Either way, I created a path w/ a symlink in it, cd'd in and was able to reproduce the same error when loading a module, so pretty sure that's it. Just need to handle resolving all symlinks in the parents dirs before starting the load to avoid inconsistent behavior.

engine/sources/local/singleflight.go
sipsma: Yeah I mean the behavior before this was to only have one mount at a time and create a brand new empty cache ref if all others were in use at the time of the sync, so now it's possible to use the same one in parallel within the same process. Wasn't attempting to make it safe across processes yet, but that'd be doable with the hardlink-farm approach (since it would become one big "content store" of individual files essentially and thus feasible to make concurrent safe w/ flock). This PR paves the way for that approach, but too big a change on top of a big change, so saving for down the line once needed.

helderco:
> Can you confirm that somewhere in the path you were loading there's a symlink? IIRC `/Users` is often a symlink on macos, but on my macbook it isn't, so not sure if I'm misremembering or what.

Hmm, no symlinks on macOS, but for sure in the virtual machine where the docker engine lives (OrbStack).

sipsma:
> Still get the same error. No symlinks on macOS, but `/Users` is mounted in the virtual machine where the docker engine lives (OrbStack). Could that be it?

@helderco Oh interesting, glad I caught the symlink problem but must be something else left to handle...

Can you share the output of `stat /Users`? Also, how did you configure that mount of /Users into orbstack? I have orbstack too so I could try to replicate it

sipsma:
> I do wonder how the big shared mirror of the user filesystem works with windows 🤔 Hopefully the paths all make sense.

Yeah I was gonna save testing Windows for a follow-up but just went ahead and did it now. Dealing with drives was not super clean unless I changed a bunch of client-side stuff in fsutil (which I've managed to avoid so far), so I ended up handling drives by just using a different filesync cache ref for each drive on a client (i.e. for client `foo` there's a separate cache ref for `C:`, `Z:`, etc.). Linux/MacOS logic stays the same and since a single filesync from a windows client is not going to span across multiple drives, I think this approach shouldn't have any downsides for windows clients.

sipsma:
@helderco and I figured out what he's hitting, it has to do with case-insensitive filesystems returning inconsistent paths (thanks MacOS for that brilliant feature 🫠). Will figure out a fix and merge this after since this is definitely going to get hit by users otherwise

PR #8823, opened 2024-10-30
feat: use lts-alpine as the default node tag

dvdksn:
Node.js 18 is in maintenance mode, Node.js 22 is the current LTS. Updating the default version to `lts-alpine` ensures that the current (active) LTS is used by default.

### DIFF
--- a/.dagger/sdk_typescript.go
+++ b/.dagger/sdk_typescript.go
@@ -21,10 +21,10 @@ const (
 	typescriptRuntimeSubdir    = "sdk/typescript/runtime"
 	typescriptGeneratedAPIPath = "sdk/typescript/api/client.gen.ts"
 
-	nodeVersionMaintenance = "18"
-	nodeVersionLTS         = "20"
+	nodeVersionMaintenance = "20.18.1"
+	nodeVersionLTS         = "22.11.0"
 
-	bunVersion = "1.1.26"
+	bunVersion = "1.1.38"
 )
 
 type TypescriptSDK struct {
--- a/.dagger/sdk_typescript.go
+++ b/.dagger/sdk_typescript.go
@@ -21,8 +21,8 @@ const (
 	typescriptRuntimeSubdir    = "sdk/typescript/runtime"
 	typescriptGeneratedAPIPath = "sdk/typescript/api/client.gen.ts"
 
-	nodeVersionMaintenance = "20.18.1"
-	nodeVersionLTS         = "22.11.0"
+	nodePreviousLTS = "20.18.1"
+	nodeCurrentLTS  = "22.11.0"
 
 	bunVersion = "1.1.38"
 )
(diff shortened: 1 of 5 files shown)


sdk/typescript/dev/node/src/index.ts
```diff
@@ -12,7 +12,7 @@ import { Commands } from "./commands"
 @object()
 class Node {
   @func()
-  version = "18-alpine"
+  version = "lts-alpine"
```
helderco, line 15: That's a good point but this only changes the default version in the module for developing the SDK. The runtime is currently on v20:

https://github.com/dagger/dagger/blob/6fdf164c7f62a8e57d6306ba07692feef1c04501/sdk/typescript/runtime/main.go#L19-L21

@TomChv, doesn't it make sense to bump the runtime to the latest LTS?

TomChv:
@dvdksn I'll continue this PR to include the changes on the TS runtime + add a changenote! Thanks a lot for the reminder 🚀

sdk/typescript/dev/node/src/index.ts
gerhard: Good catch! Remember to **always** use actual versions so that `lts` does not get updated to a new versions without us being ready for it.

If possibly, we should also use sha256 pinning so that we:
- do not do extra round-trips to check if the tag changed
- will be resilient to image tag updates

helderco: Yes, for best performance it should be the exact same as what's in the runtime. Version down to the patch (instead of "lts" or just the major), and digest.

.dagger/sdk_typescript.go
```diff
@@ -21,10 +21,10 @@ const (
 	typescriptRuntimeSubdir    = "sdk/typescript/runtime"
 	typescriptGeneratedAPIPath = "sdk/typescript/api/client.gen.ts"
 
-	nodeVersionMaintenance = "18"
-	nodeVersionLTS         = "20"
+	nodeVersionMaintenance = "20.18.1"
+	nodeVersionLTS         = "22.11.0"
```
gerhard, line 25: Having discussed this with @TomChv sync, my understanding is that the reason why we have both versions here is so that we test with the previous LTS as well as the current LTS.

If we are still finding this useful, then we should rename these two variables so that they are self-explanatory:
- `nodePreviousLTS`
- `nodeCurrentLTS`

When v24 comes out in 2024, that will become the current LTS version, and v22 will become the previous one.

gerhard (approved):
OK to merge after:
- the variable name comment is addressed
- all checks pass - apart from `test-cli-engine` https://github.com/dagger/dagger/issues/9091

.dagger/sdk_typescript.go
TomChv: Done in [696ceef](https://github.com/dagger/dagger/pull/8823/commits/696ceefc8dd7193550bec96a4996761df361f34a)

TomChv:
I'm merging this PR, I verified on our CI that it's only the `sad` test that is failing.

See: https://dagger.cloud/dagger/traces/5dc7ddb5c9b0acefaef79c62ddcd2ba4?span=98e7dcf47735f820

Issue #8676, opened 2024-10-10
✨ TypeScript SDK: resolve introspection by reference


TomChv:
The current TypeScript introspector is quite limited and remove a bunch of interesting feature that are natively available and broadly uses by TS engineers such as the complete type system, generics and complex default value.

The global idea is to make the TS SDK DX much more natural to a TS engineer.

For example, this simple snippet isn't working with dagger but should after this improvement:

```typescript
type User = {
  name: string
  age: number
}

const DEFAULT_NAME = "john"

@object()
export class Test {

  // This is broken because `User` will not be registered in the Dagger API.
  @func() 
  greet(user: User): string {
     return `Hey I'm ${user.name}, ${user.age.toString()} years old`
  }

  // This is broken because `DEFAULT_NAME` isn't resolved during the introspection.
  @func()
  newPerson(name = DEFAULT_NAME): User {
     return { name, age: 19 }
  }
}
```

Right now, you need to create a `class User`, exposing field using decorator and creating a constructor if you don't want to assign fields one by one. It's possible to hack it using `Object.assign` but it's doesn't feels right and it's not commonly used in TypeScript project.

**Then why don't we just supports the `type`keyword?**

The answer is simple: how do we know if the type should actually be exposed to the API or not? We managed that with decorators like `@object()` and `@func()` but `type`, `enum` and `interface` keywords doesn't support decorators.
So if we aim to support these keywords, we would need to resolve types & value when they are references.

### How does this work

If we reuse the same example, the flow of introspection would be:

1. Get the module name, convert it to PascalCase with will result in `Test`
2. Search for the `Test` class in the user's module code (verify that it's decorated)
3. Register existing functions and detect if there's a reference to an external declaration
4. Lookup all reference (here it would be `User` and `DEFAULT_NAME`) and register them (for default value it's just resolving what the value is)
5. Continue recursively until there's no more reference.

**What does it enable**

This would also enable constructing complex nested type without a huge boilerplate such as

```typescript
type File = {
   name: string
   size: number
}

type Directory = {
  files: File[]
  path: string
}

Instead of

@object()
class File {
  @func()
  name: string

  @func()
  size: number

  constructor(
    name: string
    size: number
  ) {
     this.name = name
     this.size = size
  }
}

// ... I think it's pretty explicit for the rest
```

And also indeed, resolve default value in a more accurate way.

It would also be an occasion to improve the introspection error handling to give more hint to the user.
[...]

PR #8824, opened 2024-10-30
feat: typescript SDK resolve value by reference.

TomChv:
Fixes https://github.com/dagger/dagger/issues/8676.

Refactor typedef resolution to be resolve by reference. The introspection starts from the entrypoint (which is the class named the same as the module) and recursively resolve every references.

Refactor the TS compiler API to be abstract with a AST that handle lookups, typedef resolutions and default value resolution.

Use `imports` to resolve default value that refers to an exported variable.

Add custom introspection error.

Improve error handling to point to the line that couldn't be introspected.

Add supports for native `enum` keyword and `type` native keyword.

Simplify overall code abstractions too minimal.
Avoid long abstractions functions.

:bulb: This is not a breaking change! I want to be compatible with the previous way, just adding new features.

**Copy of the changelog**

  Add support for non-decorable keywords by resolving declarations by references.
  Support native `enum` keyword to define enumeration without decorators.
  Example
  ```ts
  enum Example {
    A = "A",
    B = "B",
  }
  ```
  Support native `type` keyword to define type without decorators.
  This can be use to define data object or primitive type.
  Example
  ```ts
  type Example = {
    a: string
    b: number
  }
  type Integer = number
  type Foo = string
  ```
  Note: If the type is a data object, all properties will be exposed
  to API & CLI.
  Support resolution of variable and enumerations as default values if they
  are exported by the same module.
  Example
  ```ts
  export enum Example {
    A = "A",
    B = "B",
  }
  export const defaultValue = "foo"
  @object()
  class Test {
    @func()
    defaultEnumVar(value: Example = Example.A): Example {
      return value
    }
    @func
    defaultStringVar(value: string = defaultValue): string {
      return value
    }
  }
  ```


TomChv:
The scan engine error isn't related to this PR so we can ignore that one: https://github.com/dagger/dagger/actions/runs/11723011730/job/32655548946?pr=8824

sdk/typescript/introspector/scanner/typescript_module/explorer.ts
```diff
@@ -0,0 +1,23 @@
+import Module from "node:module"
+
+export function findModuleByExportedName(
```
helderco, line 3: What if two modules have the same exported name?

sdk/typescript/introspector/scanner/typescript_module/ast.ts
```diff
@@ -0,0 +1,368 @@
...
+    if (type.flags & ts.TypeFlags.Void) return { kind: TypeDefKind.VoidKind }
+
+    // If a type has a flag Object, is can basically be anything.
+    // We firstly wants to see if it's a promise or an array so we can unwrap the
+    // actual type.
+    if (type.flags & ts.TypeFlags.Object) {
+      const objectType = type as ts.ObjectType
+
+      // If it's a reference, that means it's a generic type like
+      // `Promise<T>` or `number[]` or `Array<T>`.
```
helderco, line 262: These explanations, especially when showing what you see in code like in the second line, and especially when it's not something obvious like a string, are very useful. Please do more of these 👍

sdk/typescript/introspector/scanner/typescript_module/ast.ts
```diff
@@ -0,0 +1,368 @@
...
+    const type = typeof value
+
+    switch (type) {
+      case "string":
+      case "number":
+      case "bigint":
+      case "boolean":
+      case "object":
+        // Value will be jsonified on registration so it must be a string.
+        return `${value}`
```
helderco, line 311: I don't understand this cast. Why not jsonify it?

helderco (commented):
Overall looks good to me.

All the tests are in a single file. Wouldn't hurt to have a few using multiple files, including with conflicting exported names, which would also see if aliased imports are supported:

```ts
import { Foo as Bar } from './foobar'
```

sdk/typescript/introspector/scanner/typescript_module/ast.ts
TomChv: Because it's `jsonify` during the registration so we don't need to handle it here.

core/integration/module_typescript_test.go
```diff
@@ -1329,7 +1329,330 @@ export class Test {
...
+    /**
+		 * A
+		 */
+    A = "a",
+
+		/**
+		 * B
+		 */
+    B = "b",
+}
```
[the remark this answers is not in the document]
TomChv: It's already different name and value, `a` !== `A`

core/integration/module_typescript_test.go
```diff
@@ -136,17 +136,17 @@ func (TypescriptSuite) TestInit(ctx context.Context, t *testctx.T) {
 			WithWorkdir("/work").
 			WithDirectory("/work/src", c.Directory()).
 			WithNewFile("/work/src/index.ts", `
-				import { dag, Container, object, func } from "@dagger.io/dagger"
+        import { dag, Container, object, func } from "@dagger.io/dagger"
```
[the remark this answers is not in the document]
TomChv: I did a global search for tab and replace them with space but it seems I didn't re updated the code after.

sdk/typescript/introspector/scanner/typescript_module/explorer.ts
helderco: No, I was thinking you can rename on import but Python doesn't handle that as well. It uses the name of the class itself, not the name it's imported with.

sdk/typescript/introspector/scanner/typescript_module/ast.ts
helderco: Yes, but why the cast to string?

core/integration/module_typescript_test.go
```diff
@@ -136,17 +136,17 @@ func (TypescriptSuite) TestInit(ctx context.Context, t *testctx.T) {
...
-				  @func()
-				  helloWorld(stringArg: string): Container {
-					return dag.container().from("alpine:latest").withExec(["echo", stringArg])
-				  }
-				}
+        @object()
+        export class ExistingSource {
+          @func()
+          helloWorld(stringArg: string): Container {
+          return dag.container().from("alpine:latest").withExec(["echo", stringArg])
```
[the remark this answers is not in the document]
TomChv: I cannot find that const in the snippet

```
        import { dag, Container, object, func } from "@dagger.io/dagger"

        @object()
        export class ExistingSource {
          @func()
          helloWorld(stringArg: string): Container {
          }
        }
```

Am I missing something?

sdk/typescript/introspector/scanner/typescript_module/ast.ts
TomChv: Yeah you're right, I could leave it as plain value

sdk/typescript/introspector/scanner/typescript_module/explorer.ts
TomChv: I think the TS SDK doesn't handle it as well.

core/integration/module_typescript_test.go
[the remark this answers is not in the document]
helderco: Our tests use the same image ref for alpine, golang, etc. From here:

https://github.com/dagger/dagger/blob/598860d477ef24c6e039d566c4a414bc02c889bd/core/integration/images.go#L6

For example:

https://github.com/dagger/dagger/blob/598860d477ef24c6e039d566c4a414bc02c889bd/core/integration/module_test.go#L1680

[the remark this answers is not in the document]
TomChv: Oh okay! I didn't knew that, thanks!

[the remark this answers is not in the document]
TomChv: Done

core/integration/module_typescript_test.go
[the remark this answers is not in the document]
helderco: Please don't mark feedback as resolved. It should be the reviewer upon confirming it. Otherwise it's more difficult to keep track of what we need to check for.

In this case the formatting is still all over the place. Doesn't mean I'm going to block because of it, but it's not resolved either.

sdk/typescript/introspector/scanner/typescript_module/explorer.ts
helderco: This isn't resolved because there's an unaswered question:

> What kinds of things goes through this function?

Will not block though, and let users raise any issues and deal with them when they arrive.

sdk/typescript/introspector/scanner/typescript_module/ast.ts
helderco: What do you mean? That doesn't make sense. Those are valid JSON primitives, so if you cast an integer to a string and then convert to JSON, you'll get a JSON string, not a JSON integer.

Screenshot 2024-11-18 at 23 22 56

core/integration/module_typescript_test.go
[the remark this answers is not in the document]
TomChv: Yeah I did a global cleanup of the tab, that's why there's that many changes on the formatting

sdk/typescript/introspector/scanner/typescript_module/ast.ts
TomChv: I can try to not stringify it and see what happens then

TomChv: [f805f57](https://github.com/dagger/dagger/pull/8824/commits/f805f575ead92372c4d457a82980c5259a1265bf) I'll see if the CI does well with that

core/integration/module_typescript_test.go
```diff
@@ -1329,7 +1329,330 @@ export class Test {
...
+	})
+
+	t.Run("native enum type - default value by reference", func(ctx context.Context, t *testctx.T) {
+		out, err := modGen.With(daggerCall("test-enum")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, `a`, out)
+	})
+}
+
+func (TypescriptSuite) TestReferencedDefaultValue(ctx context.Context, t *testctx.T) {
```
helderco, line 1391: Please add a test checking inspection for what's the actual default value (in JSON) that was registered to the API with these primitives. Same way you test for documentation, for example.

TomChv: Okay!

TomChv: Test added

helderco: Thanks, so as you can see, if you add that cast back you get this error:
```
	Error:      	Not equal: 
	            	expected: "4"
	            	actual  : "\"4\""
	            	
	            	Diff:
	            	--- Expected
	            	+++ Actual
	            	@@ -1 +1 @@
	            	-4
	            	+"4"
	Test:       	TestTypescript/TestReferencedDefaultValue/check_default_value_in_doc
```

PR #8839, opened 2024-11-02
feat(cli): add dagger update command

### LINKED ISSUE #8821: `dagger install` should not allow duplicate deps
It's possible to have multiple dependencies with the same `name` in `dagger.json`:

```
$ dagger init --name=test --sdk=go
$ dagger install github.com/shykes/gha
$ dagger install github.com/heuritech/gha
```

The resulting `dagger.json`:

```json
{
  "name": "test",
  "sdk": "go",
  "dependencies": [
    {
      "name": "gha",
      "source": "github.com/heuritech/gha@strategy",
      "pin": "f0c4da95fb3e7e5351d17dfb1d20362d3ae8b2a2"
    },
    {
      "name": "gha",
      "source": "github.com/shykes/gha",
      "pin": "f6c91e490043946d71100645be8253471ba626bd"
    }
  ],
  "source": ".",
  "engineVersion": "v0.13.6"
}
```

This is happening because the existing de-duping logic only handles the "symbolic" name: `github.com/heuritech/gha` and `github.com/shykes/gha` - we're entirely ignoring the `name`s.

https://github.com/dagger/dagger/blob/d4d759140e4964437f9468d6537ff7a9fd313837/core/schema/modulesource.go#L557-L574

rajatjindal:
This PR adds the `dagger update` command. Based on discussion in this ticket, we have decided not to update the version of the dependency (unless explicitly asked) but instead move the pin to the latest commit of the current version.

e.g. 

given current dependency of `<github.com/path/name>@<branch/tag>`:

- if the update command is `dagger update github.com/path/name`, it will update the `pin` to latest commit of the branch/tag. 
- however, if the update command is `dagger update github.com/path/name@<diff-version>`, it will update the `pin` to the latest commit for `<diff-version>` branch/tag.

potentially also fixes #8821

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -347,6 +347,43 @@ var moduleInstallCmd = &cobra.Command{
 	},
 }
 
+var moduleUpdateCmd = &cobra.Command{
+	Use:     "update [options] <module>",
+	Aliases: []string{"use"},
+	Short:   "Update a dependency",
+	Long:    "Update a dependency to the latest version (or the version specified). The target module must be local.",
+	Example: `"dagger update github.com/shykes/daggerverse/hello@v0.3.0" or "dagger update hello"`,
+	GroupID: moduleGroup.ID,
+	RunE: func(cmd *cobra.Command, extraArgs []string) (rerr error) {
+		ctx := cmd.Context()
+		return withEngine(ctx, client.Params{}, func(ctx context.Context, engineClient *client.Client) (err error) {
+			dag := engineClient.Dagger()
+			modConf, err := getDefaultModuleConfiguration(ctx, dag, true, false)
+			if err != nil {
+				return fmt.Errorf("failed to get configured module: %w", err)
+			}
+			if modConf.SourceKind != dagger.ModuleSourceKindLocalSource {
+				return fmt.Errorf("module must be local")
+			}
+			if !modConf.FullyInitialized() {
+				return fmt.Errorf("module must be fully initialized")
+			}
+
+			_, err = modConf.
+				Source.WithUpdateDependencies(extraArgs).
+				ResolveFromCaller().
+				AsModule().
+				GeneratedContextDiff().
+				Export(ctx, modConf.LocalContextPath)
+			if err != nil {
+				return fmt.Errorf("failed to generate code: %w", err)
+			}
+
+			return nil
+		})
+	},
+}
+
 var moduleUnInstallCmd = &cobra.Command{
 	Use:     "uninstall [options] <module>",
 	Short:   "Uninstall a dependency",
--- a/core/integration/module_cli_test.go
+++ b/core/integration/module_cli_test.go
@@ -3,6 +3,7 @@ package core
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"strings"
@@ -676,6 +677,27 @@ func (CLISuite) TestDaggerInstall(ctx context.Context, t *testctx.T) {
 		})
 	})
 
+	t.Run("installing a dependency with duplicate name is not allowed", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		ctr := c.Container().
+			From("alpine:latest").
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work").
+			With(daggerExec("init", "--sdk=go", "--name=foo", "--source=.")).
+			With(daggerExec("install", "github.com/shykes/daggerverse/docker@v0.4.1"))
+
+		daggerjson, err := ctr.File("dagger.json").Contents(ctx)
+		require.NoError(t, err)
+		require.Contains(t, daggerjson, "github.com/shykes/daggerverse/docker@docker/v0.4.1")
+
+		_, err = ctr.
+			With(daggerExec("install", "github.com/shykes/daggerverse/wolfi@v0.1.4", "--name=docker")).
+			Sync(ctx)
+
+		require.ErrorContains(t, err, "two or more dependencies are trying to use the same name")
+	})
+
 	t.Run("install dep from various places", func(ctx context.Context, t *testctx.T) {
 		c := connect(ctx, t)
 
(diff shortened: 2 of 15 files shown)


core/schema/module.go
```diff
@@ -125,6 +125,11 @@ func (s *moduleSchema) Install() {
 		dagql.NodeFunc("dependencies", s.moduleSourceDependencies).
 			Doc(`The effective module source dependencies from the configuration, and calls to withDependencies and withoutDependencies.`),
 
+		dagql.Func("withUpdateDependencies", s.moduleSourceWithUpdateDependencies).
+			Doc(`Update one or more module dependencies.`).
+			ArgDoc("dependencies", `The dependencies to update.`).
+			ArgDoc("all", `Update all dependencies`),
```
helderco, line 131: I suggest empty list of dependencies to mean all, rather than a boolean.

rajatjindal:
based on the discussion today, following would be the end UX for dagger update after this PR:

## tl;dr;

if user specifies a version update to that version, otherwise just re-pin to the latest commit of current branch/tag.

## testcases
| Current version in dagger.json  | mod provided by user for update | After update | comment
| ------------- | ------------- | ------------- | ------------- |
|  (no version specified) | `...@v0.4.2`  | `...@v0.4.2` ||
| `...@v0.4.1`  | `...@v0.4.2`  | `...@v0.4.2` ||
| `...@foo-bar`  | `...@v0.4.2`  | `...@v0.4.2` ||
|  (no version specified) |  (no version specified) | (no version updated) | re-pinned to latest of default branch|
| `...@v0.4.1`  | (no version specified) | `...@v0.4.1` | re-pinned to latest of `v0.4.1` |
| `...@foo-bar`  | (no version specified)| `...@foo-bar` | re-pinned to latest of `foo-bar` |
|  (no version specified) | `<just-name>`  | (no version updated) | re-pinned to latest of default branch|
| `...@v0.4.1`  | `<just-name>`  | `...@v0.4.1` |re-pinned to latest of `v0.4.1`|
| `...@foo-bar`  | `<just-name>`  | `...@foo-bar` |re-pinned to latest of `foo-bar`|
|  (no version specified) | `<just-name>@v0.4.2`  | `...@v0.4.2` | |
| `...@v0.4.1`  | `<just-name>@v0.4.2`  | `...@v0.4.2` ||
| `...@foo-bar`  | `<just-name>@v0.4.2`  | `...@v0.4.2` ||
| `...@foo-bar`  | NO MOD PROVIDED  | re-pin all deps to the latest of their current version ||
| dep does not exist  | with/without version | `error that dep not found in dep list` | |

shykes:
The dependency update UX makes sense to me.

Can we talk about engine and SDK update?

- A module can specify any module as SDK. Presumably that dependency is also pinned. How do I update it?
- A module can also specify an engine version. Updating that engine version is a common and important task for module devs. At the moment the UX for it is opaque and hard to guess (no command to do it explicitly; then we sneak it in when you `develop`). This seems like a good time to discuss how that UX might be improved and how it might fit with this.

jedevc:
> A module can specify any module as SDK. Presumably that dependency is also pinned. How do I update it?

Hm, good point. https://github.com/dagger/dagger/pull/8587 didn't actually include the SDK.

We don't track the version separately from the pin, so we can't actually do it easily.

However, `dagger update` *should* upgrade the SDK similarly to all the modules I think (but we need to track source and pin separately).

Essentially:

https://github.com/dagger/dagger/blob/fa4b0dc671ab82d1aaaf4224c2715b23233730c3/core/modules/config.go#L22-L23

should become

```
	// The SDK this module uses
	SDK ModuleConfigDependency `json:"sdk,omitempty"`
```

Though, maybe the `name` field doesn't make as much sense :thinking: We could make `name` `omitempty`, which should resolve that.

I *think* this can be a follow-up after this, maybe @rajatjindal would be able to help out with this one too?

> A module can also specify an engine version. Updating that engine version is a common and important task for module devs. At the moment the UX for it is opaque and hard to guess (no command to do it explicitly; then we sneak it in when you `develop`).

Mm, `dagger develop` was chosen as a point to do this to encourage it to be more automagic (instead of having lots of modules on very old versions which are difficult to support) - and to keep the existing behavior of the command.

I don't think `dagger update` should update the engine version - I think we should keep that command limited to our dependencies.

I think when originally working on module compat, I was hoping it would be mostly invisible from users - when you work on a module, you run `dagger develop` to get the API bindings - the API might have changed for the engine you're working against, so you might have to apply some backwards incompat changes yourself. The purpose of recording the `engineVersion` is more to keep old modules on daggerverse working, and to prevent running modules that might be incompatible - not really to allow users to target old versions.

> Updating that engine version is a common and important task for module devs.

I think we should maybe unpack this more - why is this a common and important task for devs? It wasn't originally intended to be, so I think I'm missing something. I think we should try and make the experience smoother (and keep it relatively hidden), if that's at all possible, instead of trying to surface it more in the API.

One potential reason for surfacing it more would be to *explicitly* allow users to target old versions of the API - e.g. I want to write a module that works with dagger versions v0.12.0+, but I'm using v0.14.0. This doesn't work today though - we only guarantee to supply a *compatible* API to old versions, we don't guarantee to supply *exactly* the same API. We could change that though, if we were interested in that use case (though personally I'm hesitant to prioritize features that keep users on old dagger versions).

.dagger/main.go
```diff
@@ -16,9 +16,10 @@ import (
 type DaggerDev struct {
 	Src *dagger.Directory // +private
 
-	Version string
-	Tag     string
-	Git     *dagger.VersionGit // +private
+	EngineVersion string
```
jedevc, line 19: This looks accidental? Not sure where this is set/used.

cmd/dagger/module.go
```diff
@@ -346,6 +346,51 @@ var moduleInstallCmd = &cobra.Command{
...
+			}
+			if !modConf.FullyInitialized() {
+				return fmt.Errorf("module must be fully initialized")
+			}
+
+			modSrc := modConf.Source
+			if len(extraArgs) > 0 {
+				modSrc = modSrc.WithUpdateDependencies(extraArgs)
+			} else {
+				//move this logic to api
```
jedevc, line 375: We should make sure that this is done before merging!

core/integration/module_cli_test.go
```diff
@@ -1300,3 +1322,266 @@ func (f *Foo) ContainerEcho(ctx context.Context, input string) (string, error) {
...
+			installCmds: [][]string{
+				{"install", "github.com/shykes/daggerverse/docker@v0.4.1"},
+			},
+			verifySetup: daggerjsonContains([]string{`"github.com/shykes/daggerverse/docker@docker/v0.4.1"`}),
+			updateCmd:   []string{"update", "github.com/shykes/daggerverse/docker@v0.4.2"},
+			verifyFns: []func(t *testctx.T, daggerjson string){
+				daggerjsonContains([]string{`"github.com/shykes/daggerverse/docker@docker/v0.4.2"`}),
+				daggerjsonNotContains([]string{`github.com/shykes/daggerverse/docker@docker/v0.4.1`}),
+			},
+			expectedError: "",
```
jedevc, line 1364: - `initCmd` is always the same - so can be removed.
- We don't need `verifySetup` - testing this behavior should actually be done in `TestDaggerInstall` (we shouldn't duplicate the testing)
- `daggerjsonContains` is kinda neat, but I think we could just have `contains` and `notContains` properties right?
- The default value of `expectedErr` is `""` anyways - so we don't need to manually set it (same with all other zero values)

---

Just a heads-up - if the tests start to feel tricky to get all into one giant table test, that's not an issue - you can split up testing into subtests (e.g. you could one that tests the `pin` updating, another that tests upgrading to a specified branch/tag, etc). Not required, but that's another option if the table tests start to feel too large.

core/integration/module_cli_test.go
```diff
@@ -1300,3 +1322,266 @@ func (f *Foo) ContainerEcho(ctx context.Context, input string) (string, error) {
...
+				{"install", "github.com/shykes/daggerverse/docker@v0.4.2"},
+			},
+			verifySetup: daggerjsonContains([]string{`"github.com/shykes/daggerverse/docker@docker/v0.4.2"`}),
+			updateCmd:   []string{"update", "docker"},
+			verifyFns: []func(t *testctx.T, daggerjson string){
+				daggerjsonContains([]string{`github.com/shykes/daggerverse/docker@docker/v0.4.2`}),
+				daggerjsonNotContains([]string{`"github.com/shykes/daggerverse/docker"`}),
+			},
+			expectedError: "",
+		},
```
jedevc, line 1475: This test (and maybe some others) need to check that the `pin` value is updated. You can do this by manually setting the `pin` in `dagger.json` to an old commit for this repo, and then checking it's correctly modified to the actual v0.4.2 commit.

core/integration/module_cli_test.go
```diff
@@ -676,6 +677,27 @@ func (CLISuite) TestDaggerInstall(ctx context.Context, t *testctx.T) {
 		})
 	})
 
+	t.Run("installing a dependency with duplicate name is not allowed", func(ctx context.Context, t *testctx.T) {
```
jedevc, line 680: Make sure to add a test for two dependencies without explicit `--name`s.

(though, note, this will probably fail until https://github.com/dagger/dagger/pull/9052 is merged - since at the point of checking, the `name` is actually empty :cry:)

core/schema/modulesource.go
```diff
@@ -514,6 +514,106 @@ func (s *moduleSchema) filterUnInstalledDeps(ctx context.Context, bk *buildkit.C
...
 
+func (s *moduleSchema) applyDepUpdates(ctx context.Context, bk *buildkit.Client, currentDeps []*modules.ModuleConfigDependency, updateList []string, updateAll bool) ([]*modules.ModuleConfigDependency, error) {
+	if len(updateList) == 0 && !updateAll {
+		return currentDeps, nil
+	}
+
+	updatesDepsMap := map[string]parsedRefString{}
+	toBeUpdated := map[string]string{}
+	for _, dep := range updateList {
+		depParsed := parseRefString(ctx, bk, dep)
```
jedevc, line 525: Updating a local dependency isn't possible, so we should error out here (also a test would be good).

core/schema/modulesource.go
```diff
@@ -514,6 +514,106 @@ func (s *moduleSchema) filterUnInstalledDeps(ctx context.Context, bk *buildkit.C
...
+
+		// for scenario where user tries to update using name
+		var cleanDepPath = filepath.Clean(depParsed.modPath)
+
+		updatesDepsMap[cleanDepPath] = depParsed
+		toBeUpdated[cleanDepPath] = dep
```
jedevc, line 531: Instead of having two maps here, the `parsedRefString` should instead include the `original` string it was parsed from.

core/schema/modulesource.go
```diff
@@ -514,6 +514,106 @@ func (s *moduleSchema) filterUnInstalledDeps(ctx context.Context, bk *buildkit.C
...
 }
 
+func (s *moduleSchema) applyDepUpdates(ctx context.Context, bk *buildkit.Client, currentDeps []*modules.ModuleConfigDependency, updateList []string, updateAll bool) ([]*modules.ModuleConfigDependency, error) {
+	if len(updateList) == 0 && !updateAll {
+		return currentDeps, nil
+	}
+
+	updatesDepsMap := map[string]parsedRefString{}
+	toBeUpdated := map[string]string{}
+	for _, dep := range updateList {
```
jedevc, line 524: Might be worth doing this in an `errgroup`.

core/schema/modulesource.go
```diff
@@ -514,6 +514,106 @@ func (s *moduleSchema) filterUnInstalledDeps(ctx context.Context, bk *buildkit.C
...
+				}, true
+			}
+		}
+
+		return nil, false
+	}
+
+	updatedDependencies := []*modules.ModuleConfigDependency{}
+	for _, currentDep := range currentDeps {
+		if updatedDep, ok := needsUpdate(currentDep, updatesDepsMap, updateAll); ok {
```
jedevc, line 596: This logic with this inline function is hard to follow - there's also lots of repetition with multiple returns. We should split out an entirely separate function `applyDepUpdate` that just updates one single `ModuleConfigDependency` to an optional specified version.

I think with that, the rest of the logic becomes quite a bit clearer.

.dagger/main.go
rajatjindal: yes it is accidental. I'll remove these.

cmd/dagger/module.go
rajatjindal: ha ha. I noticed it as well while reviewing the changes after last push and made the changes locally. will push the fix soon.

marcosnils:
> I _think_ this can be a follow-up after this, maybe @rajatjindal would be able to help out with this one too?
> 
> >

SGTM! @rajatjindal  let us know how's your backlog coming to tackle this one :pray:

core/integration/module_cli_test.go
rajatjindal: this is a great suggestion. I refactored tests to use old commit to start with and then verify that the commit is moved to the actual commit.

I also made changes to use creating `dagger.json` manually rather than using `dagger install` to avoid manipulating the `dagger.json` file.

jedevc:
> > I _think_ this can be a follow-up after this, maybe @rajatjindal would be able to help out with this one too?
> > >
> 
> SGTM! @rajatjindal let us know how's your backlog coming to tackle this one 🙏

@marcosnils, I created a new issue for this: https://github.com/dagger/dagger/issues/9156

core/schema/modulesource.go
rajatjindal: Thinking more about this. The UX would most likely be user trying to update all the deps (in which case this code is not exercised) or updating one dep at a time. 

so I am thinking if it is worth the complexity to add the `errgroup` here? what do you think?

core/schema/modulesource.go
rajatjindal: done

core/schema/modulesource.go
rajatjindal: I pulled the function out and also refactored it to make it more readable and reduce duplicate code. Please let me know if that looks better and also pls do let me know if you have more suggestions around this.

core/integration/module_cli_test.go
rajatjindal: do we want to move the fix for this change to the other PR to keep similar functionality in same PR?

jedevc (approved):

jedevc:
Oops, realized this didn't have a changelog - @rajatjindal feel free to add one as a follow-up, otherwise I'll add one during the release process.

core/integration/module_cli_test.go
jedevc: added into [`fdb3eb5` (#9052)](https://github.com/dagger/dagger/pull/9052/commits/fdb3eb5f5b53859fcc14fc2d5775b9f26e6e4d94)

PR #8861, opened 2024-11-05
don't bother with transactions for client DB

vito:
Some users are still hitting 'database is locked' errors despite SQLite being tuned properly as far as I can tell. We really don't need transactions for this, so just remove them.

### DIFF
--- a/engine/server/telemetry.go
+++ b/engine/server/telemetry.go
@@ -332,13 +332,7 @@ func spanNames(spans []sdktrace.ReadOnlySpan) []string {
 func (ps SpansPubSub) ExportSpans(ctx context.Context, spans []sdktrace.ReadOnlySpan) error {
 	slog.ExtraDebug("pubsub exporting spans", "client", ps.client.clientID, "count", len(spans))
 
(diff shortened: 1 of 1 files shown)


marcosnils (approved):

sipsma:
@vito is it the `busy_timeout` setting that allows concurrent writes to work with no locking? Just curious

vito:
@sipsma I think `journal_mode=WAL` is closer to that - `busy_timeout` just controls how long it'll wait for a lock. [This post](https://briandouglas.ie/sqlite-defaults/) does a quick rundown of common settings, but we already set all of them. (Even that post doesn't cover all of it btw, there's more in the comments [here](https://lobste.rs/s/0ykkjb/sensible_sqlite_defaults) - which we also already configure.)

It's a bit surprising this happened considering we already tune all the "magic knobs" I know of and even have one-db-per-client, but I'm pretty certain we don't actually need transactions for this, so it seems worth trying and might help perf a bit. Merged this a bit aggressively so that I can hand the CI-pushed image ref to the affected user and see if it fixes things for them. Will keep an eye out for any sketchiness.

PR #8865, opened 2024-11-06
use default cmd args when starting service

### LINKED ISSUE #8140: 💥 API inconsistency with `AsService` using entrypoint
## Summary

I propose that `asService` always executes the entrypoint and default arguments, instead of only as a fallback.

## Background

When we were close to releasing v0.12.0, we rushed to revert the ["Default exec"](https://github.com/dagger/dagger/issues/4833) because it became very easy to trigger the execution of the entrypoint undesirably, just by ending a `dagger call` chain in a `Container` object:
- https://github.com/dagger/dagger/issues/7821

I’s ok if the entrypoint is a CLI like `git`, but not if it’s a long running command. So now you have to be explicit, in case you want the `stdout` of a container and there’s no `withExec` defined, but expect to run the image’s `CMD`:
```go
dag.Container().
    Build(src).
    WithExec([]string{}).
    Stdout(ctx)  // executes CMD
```

And if `ENTRYPOINT` is also required:
```go
dag.Container().
    Build(src).
    WithExec([]string{}, dagger.ContainerWithExecOpts{UseEntrypoint: true}).
    Stdout(ctx)  // executes ENTRYPOINT + CMD
```

That’s necessary because **`Stdout` applies to the last `WithExec`**, and a `withExec` with empty arguments uses the *default arguments* (aka, `CMD`).

However, **if `CMD` is a long running process, would you want to use `Stdout` for it?** No, that’s what services as for.

## Problem

It made sense, while removing the fallback to the “default command” from `stdout`, to leave `asService` alone, because services are long running commands, and that’s more likely to be in the `CMD` of the container image. That includes using the entrypoint in that case too, because a lot of docker images have entrypoint scripts that do a lot of setup before a long running command starts, like databases.

However, this creates a weird inconsistency. It’s not even that `asService` uses the entrypoint while `stdout` and `stderr` no longer do, per se. We can reason with that (long running vs not). It’s that **it only happens if there’s no `withExec`**:

```go
dag.Container().
    From("registry:2").
    AsService()  // executes ENTRYPOINT + CMD
```

But if you add a `withExec`, then that’s what `AsService()` will execute, and without the entrypoint:
```go
dag.Container().
    From("registry:2").
    WithFile("/setup", setupFile).
    WithExec([]string{"/setup"}).
    AsService()  // executes /setup, without the entrypoint, and exits early
```

So if you actually need to execute the entrypoint and default args, you need to do this:
```go
dag.Container().
    From("registry:2").
    WithFile("/setup", setupFile).
    WithExec([]string{"/setup"}).
    WithExec([]string{}, dagger.ContainerWithExecOpts{UseEntrypoint: true}).
    AsService()  // executes ENTRYPOINT + CMD
```

## Solution

I propose that **services always execute the entrypoint and default arguments**, instead of the last `withExec`. So the empty exec would no longer be necessary:

```go
dag.Container().
    From("registry:2").
    WithFile("/setup", setupFile)
    WithExec([]string{"/setup"}).
    AsService()  // executes ENTRYPOINT + CMD
```

But if your intent is for change the command, then just change the entrypoint and/or default arguments:

```go
dag.Container().
    From("registry:2").
    WithFile("/setup", setupFile)
    WithExec([]string{"/setup"}).
    WithEntrypoint([]string{"/my-entrypoint.sh"}).
    WithDefaultArgs([]string{"/init", "stage"}).
    AsService()  // executes new ENTRYPOINT + CMD
```

## Rationale

My first impulse was to follow this idea in https://github.com/dagger/dagger/pull/7136#issuecomment-2135708579:
> Dagger considers entrypoint to be "dumb" metadata on a container image. You can read it, and change it. But Dagger does not use it to configure its own execution of containers, because it doesn't need it: the same thing can be achieved with code.

And remove the fallback from `asService` as well, in order to resolve the inconsistency. However, it does seem to me that for a service, you’ll likely want to run the `ENTRYPOINT` + `CMD` in an image. Otherwise, it’s going to be more verbose in the common case.

I actually felt that when trying to remove the use of entrypoints in our codebase, and there were a few valid uses, like the registry in our CI (long running). But not, when the same module was setting the `codegen` CLI as an entrypoint to be used in `withExec`s in a far away place (in SDKs), and having a hard time pin pointing what said `withExec`s were doing.

It’s simple to explain because these are different types:

| Type        | Long running | Use last `withExec` | Run default command        |
| ----------- | ------------ | ------------------- | -------------------------- |
| `Container` | No           | Yes                 | No, not even as a fallback |
| `Service`   | Yes          | No                  | Yes, always                |

rajatjindal:
Following changes are being done to the `AsService` API:

- Runs the `DefaultCmd` by default. This can be overwritten by passing optional `Args` option to the `AsService` API.
- Entrypoint is disabled by default. This can be enabled using the option `UseEntrypoint` option to the `AsService` API. If this is enabled, then we prepend the "Entrypoint" to the default args.

fixes #8140

### DIFF
--- a/.changes/unreleased/Breaking-20241111-195013.yaml
+++ b/.changes/unreleased/Breaking-20241111-195013.yaml
@@ -1,0 +1,11 @@
+kind: Breaking
+body: |-
+    AsService now runs "DefaultCmd" by default instead of last "Exec" command.
+
+    User can override the args by providing "ContainerAsServiceOpts.Args" option. They can also configure the container to use Entrypoint by using "ContainerAsServiceOpts.UseEntrypoint" option.
+
+    See sdk documentation for additional options.
+time: 2024-12-06T08:20:21.242739+05:30
+custom:
+    Author: rajatjindal
+    PR: "8865"
--- a/.dagger/test.go
+++ b/.dagger/test.go
@@ -358,9 +358,6 @@ func registry() *dagger.Service {
 	return dag.Container().
 		From("registry:2").
 		WithExposedPort(5000, dagger.ContainerWithExposedPortOpts{Protocol: dagger.NetworkProtocolTcp}).
-		WithExec(nil, dagger.ContainerWithExecOpts{
-			UseEntrypoint: true,
-		}).
 		AsService()
 }
 
@@ -373,8 +370,5 @@ func privateRegistry() *dagger.Service {
 		WithEnvVariable("REGISTRY_AUTH_HTPASSWD_REALM", "Registry Realm").
 		WithEnvVariable("REGISTRY_AUTH_HTPASSWD_PATH", "/auth/htpasswd").
 		WithExposedPort(5000, dagger.ContainerWithExposedPortOpts{Protocol: dagger.NetworkProtocolTcp}).
-		WithExec(nil, dagger.ContainerWithExecOpts{
-			UseEntrypoint: true,
-		}).
 		AsService()
 }
--- a/core/container.go
+++ b/core/container.go
@@ -1675,7 +1675,28 @@ func (container *Container) ImageRefOrErr(ctx context.Context) (string, error) {
 	return "", errors.Errorf("Image reference can only be retrieved immediately after the 'Container.From' call. Error in fetching imageRef as the container image is changed")
 }
 
-func (container *Container) AsService(ctx context.Context) (*Service, error) {
+type ContainerAsServiceArgs struct {
+	// Command to run instead of the container's default command
+	Args []string `default:"[]"`
+
+	// If the container has an entrypoint, prepend it to this exec's args
+	UseEntrypoint bool `default:"false"`
+
+	// Provide the executed command access back to the Dagger API
+	ExperimentalPrivilegedNesting bool `default:"false"`
+
+	// Grant the process all root capabilities
+	InsecureRootCapabilities bool `default:"false"`
+
+	// Expand the environment variables in args
+	Expand bool `default:"false"`
+
+	// Skip the init process injected into containers by default so that the
+	// user's process is PID 1
+	NoInit bool `default:"false"`
+}
+
+func (container *Container) AsServiceLegacy(ctx context.Context) (*Service, error) {
 	if container.Meta == nil {
 		var err error
 		container, err = container.WithExec(ctx, ContainerExecOpts{
@@ -1685,6 +1706,27 @@ func (container *Container) AsService(ctx context.Context) (*Service, error) {
 			return nil, err
 		}
 	}
+	return container.Query.NewContainerService(ctx, container), nil
+}
+
+func (container *Container) AsService(ctx context.Context, args ContainerAsServiceArgs) (*Service, error) {
+	var cmdargs = container.Config.Cmd
+	if len(args.Args) > 0 {
+		cmdargs = args.Args
+	}
+
+	container, err := container.WithExec(ctx, ContainerExecOpts{
+		Args:                          cmdargs,
+		UseEntrypoint:                 args.UseEntrypoint,
+		ExperimentalPrivilegedNesting: args.ExperimentalPrivilegedNesting,
+		InsecureRootCapabilities:      args.InsecureRootCapabilities,
+		Expand:                        args.Expand,
+		NoInit:                        args.NoInit,
+	})
+	if err != nil {
+		return nil, err
+	}
+
 	return container.Query.NewContainerService(ctx, container), nil
 }
 
--- a/core/integration/container_test.go
+++ b/core/integration/container_test.go
@@ -4781,3 +4781,74 @@ func (ContainerSuite) TestExecInit(ctx context.Context, t *testctx.T) {
 		require.Contains(t, out, "1 ps")
 	})
 }
+
+func (ContainerSuite) TestContainerAsService(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+	maingo := `package main
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+	"strings"
+)
+
+func main() {
+	http.HandleFunc("/hello", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, "args: %s", strings.Join(os.Args, ","))
+	})
+
+	fmt.Println(http.ListenAndServe(":8080", nil))
+}`
+	buildctr := c.Container().
+		From(golangImage).
+		WithWorkdir("/work").
+		WithNewFile("/work/main.go", maingo).
+		WithExec([]string{"go", "build", "-o=app", "main.go"})
+
+	binctr := c.Container().
+		From(alpineImage).
+		WithFile("/bin/app", buildctr.File("/work/app")).
+		WithEntrypoint([]string{"/bin/app", "via-entrypoint"}).
+		WithDefaultArgs([]string{"/bin/app", "via-default-args"}).
+		WithExposedPort(8080)
+
+	curlctr := c.Container().
+		From(alpineImage).
+		WithExec([]string{"sh", "-c", "apk add curl"})
+
+	t.Run("use default args by default", func(ctx context.Context, t *testctx.T) {
+		output, err := curlctr.
+			WithServiceBinding("myapp", binctr.AsService()).
+			WithExec([]string{"sh", "-c", "curl -vXGET 'http://myapp:8080/hello'"}).
+			Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "args: /bin/app,via-default-args", output)
+	})
+
+	t.Run("can override default args", func(ctx context.Context, t *testctx.T) {
+		withargsOverwritten := binctr.
+			AsService(dagger.ContainerAsServiceOpts{Args: []string{"sh", "-c", "/bin/app via-service-override"}})
+
+		output, err := curlctr.
+			WithServiceBinding("myapp", withargsOverwritten).
+			WithExec([]string{"sh", "-c", "curl -vXGET 'http://myapp:8080/hello'"}).
+			Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "args: /bin/app,via-service-override", output)
+	})
+
+	t.Run("can enable entrypoint", func(ctx context.Context, t *testctx.T) {
+		withargsOverwritten := binctr.
+			AsService(dagger.ContainerAsServiceOpts{
+				UseEntrypoint: true,
+			})
+
+		output, err := curlctr.
+			WithServiceBinding("myapp", withargsOverwritten).
+			WithExec([]string{"sh", "-c", "curl -vXGET 'http://myapp:8080/hello'"}).
+			Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "args: /bin/app,via-entrypoint,/bin/app,via-default-args", output)
+	})
+}
--- a/core/integration/engine_test.go
+++ b/core/integration/engine_test.go
@@ -31,6 +31,13 @@ func TestEngine(t *testing.T) {
 	testctx.Run(testCtx, t, EngineSuite{}, Middleware()...)
 }
 
+func devEngineContainerAsService(ctr *dagger.Container) *dagger.Service {
+	return ctr.AsService(dagger.ContainerAsServiceOpts{
+		UseEntrypoint:            true,
+		InsecureRootCapabilities: true,
+	})
+}
+
 // devEngineContainer returns a nested dev engine.
 func devEngineContainer(c *dagger.Client, withs ...func(*dagger.Container) *dagger.Container) *dagger.Container {
 	// This loads the engine.tar file from the host into the container, that
@@ -53,15 +60,12 @@ func devEngineContainer(c *dagger.Client, withs ...func(*dagger.Container) *dagg
 	return ctr.
 		WithMountedCache("/var/lib/dagger", c.CacheVolume("dagger-dev-engine-state-"+identity.NewID())).
 		WithExposedPort(1234, dagger.ContainerWithExposedPortOpts{Protocol: dagger.NetworkProtocolTcp}).
-		WithExec([]string{
+		WithDefaultArgs([]string{
 			"--addr", "tcp://0.0.0.0:1234",
 			"--addr", "unix:///var/run/buildkit/buildkitd.sock",
 			// avoid network conflicts with other tests
 			"--network-name", deviceName,
 			"--network-cidr", cidr,
-		}, dagger.ContainerWithExecOpts{
-			UseEntrypoint:            true,
-			InsecureRootCapabilities: true,
 		})
 }
 
@@ -185,7 +189,7 @@ func (ClientSuite) TestWaitsForEngine(ctx context.Context, t *testctx.T) {
 			WithEntrypoint([]string{"/usr/local/bin/slow-entrypoint.sh"})
 	})
 
-	clientCtr := engineClientContainer(ctx, t, c, devEngine.AsService())
+	clientCtr := engineClientContainer(ctx, t, c, devEngineContainerAsService(devEngine))
 	_, err := clientCtr.
 		WithNewFile("/query.graphql", `{ version }`). // arbitrary valid query
 		WithExec([]string{"dagger", "query", "--doc", "/query.graphql"}).Sync(ctx)
@@ -198,11 +202,11 @@ func (EngineSuite) TestSetsNameFromEnv(ctx context.Context, t *testctx.T) {
 
 	engineName := "my-special-engine"
 	engineVersion := engine.Version + "-special"
-	devEngineSvc := devEngineContainer(c, func(c *dagger.Container) *dagger.Container {
+	devEngineSvc := devEngineContainerAsService(devEngineContainer(c, func(c *dagger.Container) *dagger.Container {
 		return c.
 			WithEnvVariable("_EXPERIMENTAL_DAGGER_ENGINE_NAME", engineName).
 			WithEnvVariable("_EXPERIMENTAL_DAGGER_VERSION", engineVersion)
-	}).AsService()
+	}))
 
 	clientCtr := engineClientContainer(ctx, t, c, devEngineSvc)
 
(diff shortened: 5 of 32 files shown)


core/container.go
```diff
@@ -1644,6 +1677,34 @@ func (container *Container) Service(ctx context.Context) (*Service, error) {
 	return container.Query.NewContainerService(ctx, container), nil
 }
 
+// TODO(rajatjindal): bikeshedding for name
+func (container *Container) AsService(ctx context.Context, args ContainerAsServiceArgs) (*Service, error) {
```
helderco, line 1681: I think we tend to rename the current one to `LegacyXXX` (`LegacyService` & `Service`) if it's kept around just for compat reasons. Makes clear why there's two versions. Sometimes you can use the same function for old and new.

rajatjindal:
Hi @helderco @shykes @marcosnils,

I think this PR is ready for review now. 

Thanks

core/integration/container_test.go
```diff
@@ -4831,3 +4832,74 @@ func (ContainerSuite) TestExecInit(ctx context.Context, t *testctx.T) {
...
+		From("golang:1.23-alpine").
+		WithWorkdir("/work").
+		WithNewFile("/work/main.go", maingo).
+		WithExec([]string{"go", "build", "-o=app", "main.go"})
+
+	binctr := c.Container().
+		From(alpineImage).
+		WithFile("/bin/app", buildctr.File("/work/app")).
+		WithDefaultArgs([]string{"/bin/app", "via-default-args"}).
+		WithEntrypoint([]string{"/bin/app", "via-entrypoint"}, dagger.ContainerWithEntrypointOpts{KeepDefaultArgs: true}).
```
helderco, line 4864: A simple way to avoid the `KeepDefaultArgs` option is to set the entrypoint before the default args.

core/integration/engine_test.go
```diff
@@ -50,15 +57,12 @@ func devEngineContainer(c *dagger.Client, withs ...func(*dagger.Container) *dagg
 	return ctr.
 		WithMountedCache("/var/lib/dagger", c.CacheVolume("dagger-dev-engine-state-"+identity.NewID())).
 		WithExposedPort(1234, dagger.ContainerWithExposedPortOpts{Protocol: dagger.NetworkProtocolTcp}).
-		WithExec([]string{
+		WithDefaultArgs([]string{
 			"--addr", "tcp://0.0.0.0:1234",
```
helderco, line 61: What's the actual entrypoint? Can it be moved as first argument here?

core/schema/service.go
```diff
@@ -18,9 +18,21 @@ var _ SchemaResolvers = &serviceSchema{}
 
 func (s *serviceSchema) Install() {
 	dagql.Fields[*core.Container]{
-		dagql.Func("asService", s.containerAsService).
+		dagql.Func("asService", s.containerAsServiceLegacy).
+			View(BeforeVersion("v0.13.8")).
```
helderco, line 22: We didn't release a v0.13.8, next version is v0.14.1, but maybe this is supposed to be scheduled for v0.15.0?

core/schema/service.go
```diff
@@ -18,9 +18,21 @@ var _ SchemaResolvers = &serviceSchema{}
...
 				`Be sure to set any exposed ports before this conversion.`),
+
+		dagql.Func("asService", s.containerAsService).
+			Doc(`Turn the container into a Service.`,
+				`Be sure to set any exposed ports before this conversion.`).
+			ArgDoc("args",
+				`Command to run instead of the container's default command (e.g., ["run", "main.go"]).`,
+				`If empty, the container's default command is used.`).
+			ArgDoc("useEntrypoint",
+				`If the container has an entrypoint, prepend it to the args.`).
```
helderco, line 33: Missing docs for the other args.

core/schema/service.go
```diff
@@ -18,9 +18,21 @@ var _ SchemaResolvers = &serviceSchema{}
...
+			View(AfterVersion("v0.13.8")),
```
helderco, line 34: More visible as first in the chain, before `Doc`. Same adjustment for the version.

core/schema/service.go
```diff
@@ -68,8 +80,23 @@ func (s *serviceSchema) Install() {
 	}.Install(s.srv)
 }
 
-func (s *serviceSchema) containerAsService(ctx context.Context, parent *core.Container, args struct{}) (*core.Service, error) {
-	return parent.Service(ctx)
+func (s *serviceSchema) containerAsServiceLegacy(ctx context.Context, parent *core.Container, args struct{}) (*core.Service, error) {
+	return parent.ServiceLegacy(ctx)
+}
```
helderco, line 85: Should add a test under LegacySuite.

docs/docs-graphql/schema.graphqls
```diff
@@ -60,7 +60,25 @@ type Container {
...
+    """
+    args: [String!] = []
+    expand: Boolean = false
+    expect: ReturnType = SUCCESS
+    experimentalPrivilegedNesting: Boolean = false
+    insecureRootCapabilities: Boolean = false
+    noInit: Boolean = false
+    redirectStderr: String = ""
+    redirectStdout: String = ""
+    stdin: String = ""
```
helderco, line 77: Missing docs for these arguments.

core/integration/container_test.go
rajatjindal: ah. that is good to know. thanks

core/schema/service.go
rajatjindal: oh right. thanks for catching this.

core/schema/service.go
rajatjindal: done.

core/schema/service.go
rajatjindal: I think I did it as you suggested, but pls let me know if i misunderstood this suggestion.

core/schema/service.go
rajatjindal: 🤦 , thank you,

core/schema/service.go
rajatjindal: I added with version 0.14.0 for now. If I change this to version v0.15.0, it didn't work for me.

core/integration/engine_test.go
rajatjindal: I think its coming from the OCI import of `engine tar file`. I can add it here explicitly, but I am thinking it may be a. duplicate call here. what do you think?

core/integration/container_test.go
rajatjindal: done

core/integration/container_test.go
```diff
@@ -4831,3 +4832,74 @@ func (ContainerSuite) TestExecInit(ctx context.Context, t *testctx.T) {
...
+
+func main() {
+        http.HandleFunc("/hello", func(w http.ResponseWriter, r *http.Request) {
+                fmt.Fprintf(w, "args: %s", strings.Join(os.Args, ","))
+        })
+
+        fmt.Println(http.ListenAndServe(":8080", nil))
+}`
+	buildctr := c.Container().
+		From("golang:1.23-alpine").
```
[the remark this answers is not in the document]
rajatjindal: done

.dagger/test.go
```diff
@@ -243,9 +243,6 @@ func registry() *dagger.Service {
 	return dag.Container().
 		From("registry:2").
 		WithExposedPort(5000, dagger.ContainerWithExposedPortOpts{Protocol: dagger.NetworkProtocolTcp}).
-		WithExec(nil, dagger.ContainerWithExecOpts{
-			UseEntrypoint: true,
-		}).
```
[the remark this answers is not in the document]
rajatjindal: yeah, it seems like it is only used from `testCmd` function, which binds this registry to the devEngine. but I could be misunderstanding something here.

.changes/unreleased/Breaking-20241111-195013.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Breaking
+body: "AsService now runs \"DefaultCmd\" by default instead of last \"Exec\" command. \n\nUser can override this behavior by providing \"ContainerAsServiceOpts.Args\" option. User can also configure the container to use Entrypoint by using \"ContainerAsServiceOpts.UseEntrypoint\" option. See documentation for more options. "
```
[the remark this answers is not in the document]
rajatjindal: done

core/integration/engine_test.go
helderco: If you can track where that's defined and used, I'd probably remove the entrypoint and explicitly add the argument on the callers.

core/schema/service.go
helderco: > If I change this to version v0.15.0, it didn't work for me.

If you look at other tests in `core/integration/legacy_test.go` you can see that there's a version override for GraphQL queries (`testutil.QueryOptions{Version: "xxx"}`). You can use this to simulate a future version.

So `BeforeVersion` should target `v0.15.0`, and as for `AfterVersion`, it's useful if you want to merge before that release so you don't have to keep rebasing and potentially having to solve merge conflicts over and over.

But if you intend to hold on to the merge until the `v0.15.0` release, then just use `View(AllVersion)` for the new default, which will take effect for any version that doesn't match the `BeforeVersion`.

 See some current examples of this in the schema (like `schema/container.go`).

shykes:
Just checking on the status of this?

core/integration/engine_test.go
rajatjindal: I have moved the entrypoint config to the caller. It seems like that is the only place we are using this from code perspective, but i am also wondering if something implicitly depends on that. e.g. `hack/dev`?

core/schema/service.go
rajatjindal: specifying the engine version in `dagger.json` worked for me this time. I am using `v0.15.0` for both `AfterVersion` and `BeforeVersion` as we do intend to merge this before v0.15.0

@jedevc has targetted this issue for v0.14.1 as well, i am wondering if we need to handle that as well.

core/integration/engine_test.go
rajatjindal: It seems like this triggered failures in other areas. I just removed this commit to verify if that fixes the CI issues.

core/schema/service.go
rajatjindal: Hi @helderco 

I have a question:

- so lets say current release is v0.14.0
- now I changed the AsService api so that for `BeforeVersion("v0.15.0")` it should serve old api, and for `AfterVersion("v0.15.0")` it serves the new version.
- now when we run the tests, it runs with the current version of dagger, which would be `v0.14.1-.....`
- that would mean that the api visible to tests would be the old one. so we don't need any changes in existing tests yet.
- now lets say `v0.15.0` is released. this would mean tests would suddently start seeing `v0.15.0` version of api, and the tests would fail as the new api is no longer backward compatible. 
- now we would need to make changes to the existing tests?

does ^^ makes sense? It almost feels like I am missing something obvious here.

.dagger/test.go
jedevc, line 1: Can you remove the changes from #9011 from this PR? We should review and merge that one separately.

core/container.go
```diff
@@ -1656,7 +1656,40 @@ func (container *Container) ImageRefOrErr(ctx context.Context) (string, error) {
...
+	Stdin string `default:""`
+
+	// Redirect the command's standard output to a file in the container
+	RedirectStdout string `default:""`
+
+	// Redirect the command's standard error to a file in the container
+	RedirectStderr string `default:""`
+
+	// Exit codes this exec is allowed to exit with
+	Expect ReturnTypes `default:"SUCCESS"`
```
jedevc, line 1676: :thinking: so are these args the duplicate of `WithExec`? If so, we should embed the `core.ContainerExecOpts` struct to avoid duplication and prevent drift.

But also, not entirely sure, since some of these aren't *strictly* the same right now. See https://github.com/dagger/dagger/blob/d50fee49415d9802add4d6c6588a842ed855e953/core/service.go#L434-L445

Not all of the ops from the `WithExec` are actually propagated through to a container:
- `Stdin`, `RedirectStdout`, `RedirectStderr` are actually misleading, because they're actually passed into `startContainer`. Even if you *could* redirect output, you can't read files from a running container (yet).
- `Expect` doesn't affect anything, since the service's exit code is really just discarded right?
- `NoInit` doesn't seem to do anything?

I think we should probably remove these options from the new `AsService` method entirely, better to not have them, than to have them there and be confusing.

core/schema/service.go
jedevc: Oop, responded on discord: https://discord.com/channels/707636530424053791/1003718839739105300/1312018293812494357

.dagger/test.go
rajatjindal: yes. will do. currently i am debugging the failures in the CI for this PR.

core/container.go
rajatjindal: > 🤔 so are these args the duplicate of WithExec

One of the difference is that `Args` is optional in `AsService` api.

> NoInit doesn't seem to do anything?

hmm, this seems to be the case. But that means it didn't had any effect with the current `AsService` api either. I'll double check this.

> Expect doesn't affect anything, since the service's exit code is really just discarded right?

yeah, that makes sense. I'll double check once, and then remove this and other suggested fields from AsService api.

jedevc: > hmm, this seems to be the case. But that means it didn't had any effect with the current `AsService` api either. I'll double check this.

Yeah good point! I do think that having this on `AsService` directly though *feels* different than getting a previous `WithExec` to run as a service. If we're moving to a cleaner API, I think it's fair to clear this ambiguity up!

rajatjindal:
depends on https://github.com/dagger/dagger/pull/9073 (which updates the version to v0.15.0).

.dagger/test.go
rajatjindal: changes from 9011 are now removed.

core/integration/legacy_test.go
```diff
@@ -756,3 +756,78 @@ func (LegacySuite) TestContainerWithFocus(ctx context.Context, t *testctx.T) {
 		})
 	require.NoError(t, err)
 }
+
+func (LegacySuite) TestContainerAsService(ctx context.Context, t *testctx.T) {
```
jedevc, line 760: Can you add a comment like in the other tests explaining what this test is specifically looking for, and which PR changed it?

It's quite useful when readying through these later.

core/integration/legacy_test.go
```diff
@@ -756,3 +756,78 @@ func (LegacySuite) TestContainerWithFocus(ctx context.Context, t *testctx.T) {
...
+	c := connect(ctx, t)
+
+	serversource := `package main
+
+import (
+        "fmt"
+        "net/http"
+        "os"
+				"strings"
+)
```
jedevc, line 770: formatting nit :bug:

jedevc (approved):

rajatjindal:
there are a couple of failures after rebasing and resoving the conflict. I am fixing those.

rajatjindal:
I think this PR is ready for merge now. I believe I have addressed all the review comments, and the CI looks happy as well.

.changes/unreleased/Breaking-20241111-195013.yaml
```diff
@@ -0,0 +1,12 @@
+kind: Breaking
+body: |
+AsService now runs "DefaultCmd" by default instead of last "Exec" command.
+
+User can override the args by providing "ContainerAsServiceOpts.Args" option. They can also 
+configure the container to use Entrypoint by using "ContainerAsServiceOpts.UseEntrypoint" option. 
+
+See sdk documentation for additional options.
```
helderco, line 8: This isn't valid yaml is it?
```suggestion
    AsService now runs "DefaultCmd" by default instead of last "Exec" command.

    User can override the args by providing "ContainerAsServiceOpts.Args" option. They can also 
    configure the container to use Entrypoint by using "ContainerAsServiceOpts.UseEntrypoint" option. 

    See sdk documentation for additional options.
```

core/integration/container_test.go
```diff
@@ -4781,3 +4781,74 @@ func (ContainerSuite) TestExecInit(ctx context.Context, t *testctx.T) {
...
+        "os"
+				"strings"
+)
+
+func main() {
+        http.HandleFunc("/hello", func(w http.ResponseWriter, r *http.Request) {
+                fmt.Fprintf(w, "args: %s", strings.Join(os.Args, ","))
+        })
+
+        fmt.Println(http.ListenAndServe(":8080", nil))
```
helderco, line 4801:
```suggestion
import (
	"fmt"
	"net/http"
	"os"
	"strings"
)

func main() {
	http.HandleFunc("/hello", func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprintf(w, "args: %s", strings.Join(os.Args, ","))
	})

	fmt.Println(http.ListenAndServe(":8080", nil))
```

core/integration/legacy_test.go
```diff
@@ -756,3 +756,83 @@ func (LegacySuite) TestContainerWithFocus(ctx context.Context, t *testctx.T) {
...
+		File("/work/app")
+
+	ctr := daggerCliBase(t, c).
+		WithWorkdir("/work/").
+		WithFile("app", app).
+		WithNewFile("main.go", daggermodmaingo).
+		WithNewFile("dagger.json", `{"name": "foo", "sdk": "go", "source": ".", "engineVersion": "v0.13.7"}`)
+
+	// verify that the engine uses the entrypoint when serving the legacy AsService api
+	t.Run("use entrypoint by default", func(ctx context.Context, t *testctx.T) {
```
helderco, line 831: This doesn't cover everything. In legacy, the entrypoint is used only when there's no withExec steps. Can you add a test case confirming that if there's a withExec that's what's used, instead of the entrypoint+default args?

core/schema/service.go
```diff
@@ -18,9 +18,40 @@ var _ SchemaResolvers = &serviceSchema{}
...
+			View(BeforeVersion("v0.15.0")).
 			Doc(`Turn the container into a Service.`,
 				`Be sure to set any exposed ports before this conversion.`),
+
+		dagql.Func("asService", s.containerAsService).
+			View(AfterVersion("v0.15.0")).
+			Doc(`Turn the container into a Service.`,
+				`Be sure to set any exposed ports before this conversion.`).
+			ArgDoc("args",
+				`Command to run instead of the container's default command (e.g., ["run", "main.go"]).`,
```
helderco, line 31: This example assumes there's a `["go"]` entrypoint, right?

rajatjindal: yes and that it is enabled.

helderco: We want to discourage the usage of entrypoint, so our examples, docs, even our own code, should avoid it too.

helderco: E can see `withExec` has the same example, but we should change both.

helderco (approved):

PR #8768, opened 2024-10-23
docs: Add Depot to GitHub integration page

vikram-dagger:

PR #8866, opened 2024-11-06
docs: Add CI sub-section to navbar

vikram-dagger:
Relates to https://github.com/dagger/dagger/pull/8768#issuecomment-2460353863

### DIFF
--- a/docs/sidebars.ts
+++ b/docs/sidebars.ts
@@ -150,6 +150,10 @@ module.exports = {
     {
       "type": "category",
       "label": "Integrating Dagger with CI",
+      "link": {
+        "type": "doc",
+        "id": "integrations/ci"
+      },
       "collapsible": true,
       "collapsed": true,
       "items": [
--- a/docs/sidebars.ts
+++ b/docs/sidebars.ts
@@ -149,11 +149,7 @@ module.exports = {
     },
     {
       "type": "category",
-      "label": "Integrations",
-      "link": {
-        "type": "doc",
-        "id": "integrations/index"
-      },
+      "label": "Integrating Dagger with CI",
       "collapsible": true,
       "collapsed": true,
       "items": [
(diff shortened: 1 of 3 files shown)


shykes:
Note to @dagger/maintainers: when I try to review this PR locally, I get an engine error:

```console
$ dagger call -m github.com/dagger/dagger@pull/8866/head docs server up
Full trace at https://dagger.cloud/dagger/traces/9b7354ca9aa57ada43e8dae8fc411582

Error: get module name: input: failed to get schema: failed to get schema for module "dagger-dev": type "Engine" is already defined by module "daggercore"
```

This is with dagger 0.14.0

docs/sidebars.ts
```diff
@@ -149,10 +149,10 @@ module.exports = {
     },
     {
       "type": "category",
-      "label": "Integrations",
+      "label": "Integrating Dagger with CI",
```
gerhard, line 152: `Integrating with CI` reads better to me.

gerhard (approved):
This is definitely a step in the direction!

Looking at what is left in the `Integrations` category, I would go further and split into:
1. Platforms (GitHub, Kubernetes, etc.)
2. Languages (Java, PHP)

The second category feels forced. I would be inclined to put it in a **guide** or **cookbook recipe** (recipe doesn't seem to have the right shape for this content). Still, an entire category for two languages doesn't make sense unless we want to add Rust & Elixir to this mix.

LGTM 👍

---

WDYT @samalba ?

samalba (approved):
I like it, indeed it clears the confusion. More intuitive IMO.

docs/sidebars.ts
vikram-dagger: Done

vikram-dagger:
> Looking at what is left in the `Integrations` category, I would go further and split into:
> 
> 1. Platforms (GitHub, Kubernetes, etc.)
> 2. Languages (Java, PHP)
> 
> The second category feels forced. I would be inclined to put it in a **guide** or **cookbook recipe** (recipe doesn't seem to have the right shape for this content). Still, an entire category for two languages doesn't make sense unless we want to add Rust & Elixir to this mix.

I'm in favour of doing this but probably not at the moment, because we will likely end up with one overweight sub-category (Platforms) and one underweight, as you already noted (Languages). We also end up with one more level of nesting in the sidebar. So it's con "unbalanced categories+additional nesting level", pro "more precise categorization".  For the moment, I would leave this as is, but I have opened an internal issue to revisit it once we have more integrations page to consider.

PR #8868, opened 2024-11-07
Improve our Dagger module for Go

shykes:
Major improvements to github.com/dagger/dagger/modules/go, with all required features to build and test Dagger.

New configuration settings:

- Download and build caches
- Base container
- Lldflags
- Injected values
- CGO support
- Race detector
- DWARF generation
- Symbol table

New functions:

- Build: build specified packages and return a directory
- Binary: build the specified binary and return a file
- Download: download dependencies to the cache
- Test: run tests for the specified packages (supports test filtering, parallel, timeout, failfast)
- Tests: list available tests


shykes:
FYI @dagger/maintainers this is extracted from #8355 . It does *not* wire in the module into our build, but it should be easy to do so. At the time of the original PR, this version of the Go module had all the features needed to be wired in seamlessly. I will look into doing that as a follow up.

shykes:
Looks like the remaining issues are flakes.

modules/go/main.go
```diff
@@ -2,28 +2,104 @@ package main
...
+		// Cache volumes should be namespaced by module, but they aren't (yet).
+		// For now, we namespace them explicitly here.
+		moduleCache = dag.CacheVolume("github.com/dagger/dagger/modules/go:modules")
+	}
+	if buildCache == nil {
+		// Cache volumes should be namespaced by module, but they aren't (yet).
+		// For now, we namespace them explicitly here.
+		buildCache = dag.CacheVolume("github.com/dagger/dagger/modules/go:build")
+	}
+	if base == nil {
```
jedevc, line 70: I wonder if we should still allow this.

We need the base to be customizable (we had a discussion somewhere before about Ubuntu being built on Ubuntu, alpine on alpine, wolfi on wolfi - otherwise we hit weird libc issues). But now we have the additional constraint that this has to have `protoc` installed.

I'm not convinced about the overlay concept here as well in terms of our need to (eventually) do multi platform images here - how do we ensure it works with Ubuntu/alpine/wolfi?

Maybe we could use a distro enum, like we do at the top-level module?

Not a blocker, but would like to discuss the overlay concept more separately in that pr.

modules/go/main.go
```diff
@@ -34,57 +110,280 @@ type Go struct {
...
-	return dag.
-		Wolfi().
-		Container(dagger.WolfiContainerOpts{Packages: pkgs}).
-		WithEnvVariable("GOLANG_VERSION", p.Version).
-		WithEnvVariable("GOPATH", "/go").
-		WithEnvVariable("PATH", "${GOPATH}/bin:${PATH}", dagger.ContainerWithEnvVariableOpts{Expand: true}).
-		WithDirectory("/usr/local/bin", dag.Directory()).
-		WithMountedCache("/go/pkg/mod", p.goModCache()).
-		WithMountedCache("/root/.cache/go-build", p.goBuildCache())
+// Download dependencies into the module cache
```
jedevc, line 136: Hm, this is odd - we shouldn't have a function that has side effects to populate a cache volume. These two systems interact in weird ways.

It's *totally* possible that this go mod download exec could cache, but the cache volume remains empty - that's confusing, but entirely expected.

Do we really need this function? If we really do, we should remove the cache volumes in this function before running it. I don't think it is needed though, it also feels like it exposes too much of the underlying go command to the user.

modules/go/main.go
```diff
@@ -34,57 +110,280 @@ type Go struct {
...
+		Env(platform)
+	cmd := []string{"go", "build", "-o", output}
+	for _, pkg := range mainPkgs {
+		env = env.WithExec(goCommand(cmd, []string{pkg}, ldflags, p.Values, p.Race))
+	}
+	return dag.Directory().WithDirectory(output, env.Directory(output)), nil
+}
+
+// Build a single main package, and return the compiled binary
+func (p Go) Binary(
```
jedevc, line 246: While we're here - what do you think about `Executable` instead of `Binary`? `Binary` feels too generic still (it's better than `File` which is what it did used to be ages ago).

jedevc:
From a coffee shop on holiday, lgtm ☕ 

Just a quick note on backwards compat here - I'd like to leave the door open for future breaking API changes in this module. Until we get all our CI in a stable state, I think we shouldn't box ourselves in - just raising this now, since it was a big pain point in the gha module changes recently. If we want to freeze that *now* - I think maybe I'd like a bit more time to bike shed this design.

Left some small notes - if any of them can be fixed before merging great, but otherwise, we can get 'em after 🎉

jedevc (approved):

modules/go/main.go
shykes: The `overlay` feature is definitely up for discussion (not a coincidence that it's in a separate PR :)

But `base` I think is a must-have. It is very easy to understand, and opens a very useful escape hatch without much additional complexity. If you bring your own, you're responsible for making sure go is installed. Kind of a no-brainer IMO.

shykes:
@jedevc yes I consider `github.com/dagger/dagger/modules/` to be *unstable*, it is totally OK to keep breaking those APIs (as long as there's a good reason to, and ideally there's good design consensus).

As I mentioned elsewhere, I would be more comfortable if we made this *more clear* to the community. At the moment these modules show up in daggerverse searches, and because they're under the dagger org, they have an official aura that makes them "stdlib-ish". I would like to protect our ability to break these modules without making anyone sad.

modules/go/main.go
shykes: Mm, maybe I'm misunderstanding the best practices. This `Download` is my interpretation of our current best practices, as a clean API.

The best practice in question: "download go dependencies as a separate command from building, to take advantage of cache volumes". You're right about the command caching - perhaps I'm missing a cache-buster?

Did I get the best practice wrong @sipsma @kpenfound ?

modules/go/main.go
shykes: In the context of CI/CD, I don't think "binary" is too generic. Here's what I get from gpt4:

> > In a devops and CI/CD context, what is a "binary"?
> In DevOps and CI/CD, a “binary” refers to an executable file produced by compiling source code. This binary file contains machine code that a computer can run directly without needing an interpreter or further compilation. In a CI/CD pipeline, binaries are often built as artifacts during the build stage and then tested, deployed, or distributed as part of software delivery.

modules/go/main.go
jedevc: Sounds good.

Could we add a sanity check that `base` has the right things installed before we start to use it? It could check the `go version` is the right one, make sure that `protoc` is installed, etc.

Fine as a follow-up, just on a quest to remove footguns!

modules/go/main.go
jedevc: I *think* this makes sense as an internal detail, for sure. Splitting the download and usage is good (although, a bit odd I suppose, since if the cache volume is actually empty and `go mod download` gets cached, then we end up downloading deps in the build step).

I think the issue is more that it shouldn't be a public API method for users to consume - it can just be a private method, since ideally this module should conceal how the sausage gets made.

modules/go/main.go
jedevc: Sure! No objections, was just wondering if we wanted to change this while in the area - sounds good as is :tada:

jedevc:
Pushed a commit to fixup a typo I noticed (and also sorted the `go.mod` file a little bit).

shykes:
@jedevc implemented the fix we discussed

jedevc (approved):

jedevc:
Leeeet's go :rocket:

PR #8869, opened 2024-11-07
modules/go: add overlay support

shykes:
*This depends on #8868 , please review that first*

- In `github.com/dagger/dagger/modules/go`, you can now pass "overlays" which can be installed on the base container to modify it. This can be used to install extra packages, attach external services, or any other custom modification.

- `github.com/dagger/dagger/modules/alpine` now implements the `Overlay` interface used by the `go` module, such that any Alpine (or Wolfi) configuration can be passed as an overlay to the Go container

- `github.com/dagger/dagger` takes advantage of this to install extra packages (`protoc`) as an overlay

Note: this takes advantage of the interfaces feature, special thank you to @kpenfound and @marcosnils for helping me figuring it out, a second time... It's clunky but it's the best we have right now. The potential is huge! Worth looking past the superficial complexity.


.dagger/main.go
```diff
@@ -125,7 +125,19 @@ func (dev *DaggerDev) containing(ctx context.Context, filename string) ([]string
 
 // Dagger's Go toolchain
 func (dev *DaggerDev) Go() *GoToolchain {
-	return &GoToolchain{Go: dag.Go(dev.Source())}
+	// Modify the standard Go container by installing the protoc binary
```
jedevc, line 128: Hm 🤔 as mentioned in the other thread, I'm a touch worried about how this will work for the case of building an Ubuntu container (which is how our gpu images are built). Packages aren't necessarily just overlayable like this always.

That said - I'm not sure the best way to do this. I think having overlays isn't necessarily *wrong* but overlaying wolfi packages into operating systems that aren't wolfi doesn't feel right to me. And potentially could give scary errors later that are *very* hard to debug.

Could we overlay a single file, statically compiled protoc package? That would work in more cases. And also keeps the overlays, which I think is an elegant solution that splits out protoc from the concerns of the go package.

My main concern - I want to avoid the "use at your own risk" comment - it makes sense, and it's a good warning to have - but for our golden example CI that folks might copy-paste code and ideas from, I think we should try and avoid foot guns as much as possible.

Open to being persuaded! But as is, I do think this deserves some more discussion and bikeshedding 🚲🚲 (since it sets a new design and pattern, that we might use in other places, I think it's worth thinking it through together).

PR #8871, opened 2024-11-07
Fix cgroup error logs + cache volume re-use in go SDK

sipsma:
I noticed a few easy fix issues while perusing engine logs from a test run earlier today. They technically aren't related but combining them into one PR for convenience.

--- 

[engine: ignore missing cpu.stat/cpu.pressure files](https://github.com/dagger/dagger/commit/cca2f47d4427108494d391bbaef63c51d8822c64)

We already ignored missing cgroup files for io.stat/io.pressure but
missed that for cpu files. These files can be missing if the engine
doesn't support cgroups for execs and result in a ton of useless log
noise in that case.

---

[engine: fix cache volume re-use when sourced from builtinContainer](https://github.com/dagger/dagger/commit/adc5b6a5c4facc0e480abf1750086eba62ccd07b)

I noticed that there were a bunch of engine logs about not being able to
re-use cache mount refs due to missing descriptors for lazy blobs.

These is buildkit mumbo jumbo that means it left certain cache refs lazy
but when it needed to unlazy them it didn't find the required metadata
to do so hidden in the context.

This was happening specifically because of the recent change to set go
module building containers to use cache volumes sourced from the base
container. The base container was loaded using OCI LLB which leaves
cache refs lazy.

I am pretty sure this is just a case not handled by buildkit's cache
mount code, but it's easiest to fix in our own code by ensuring these
refs aren't lazy in `builtinContainer`, which this update does by
performing the right magical incantation (solve the ref and call
`.Extract` on it).

### DIFF
--- a/core/schema/host.go
+++ b/core/schema/host.go
@@ -10,14 +10,18 @@ import (
 	"github.com/containerd/containerd/content/local"
 	"github.com/containerd/containerd/labels"
 	"github.com/moby/buildkit/client/llb"
+	bkgw "github.com/moby/buildkit/frontend/gateway/client"
+	bkworker "github.com/moby/buildkit/worker"
 	"github.com/opencontainers/go-digest"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/dagger/dagger/core"
 	"github.com/dagger/dagger/dagql"
 	"github.com/dagger/dagger/engine"
 	"github.com/dagger/dagger/engine/buildkit"
 	"github.com/dagger/dagger/engine/distconsts"
+	"github.com/dagger/dagger/engine/slog"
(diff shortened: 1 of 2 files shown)


vito (approved):

sipsma:
I noticed that while the "missing lazy blobs" error in logs is *often* gone, it still shows up sometimes. After poking a bit more I really can't see how this is anything but a buildkit bug since it's also inconsistent which blobs are missing.

The change still seems to improve the situation by making it happen less often and the error is non-fatal (just results in duplicate cache volumes being created inside buildkit), so I will go with it for now but open a follow-up issue to figure out wtf is happening there.

sipsma:
I figured out the upstream bug: https://github.com/dagger/dagger/issues/8884

Using that knowledge I'm going to try a more robust workaround in our code here that should make the fix consistent. Just don't want to block these fixes waiting on upstream since it's degrading our performance when we have less cache volume reuse.

PR #8880, opened 2024-11-07
feat(engine): add memory telemetry

cwlbraa:
this PR adds telemetry for memory usage, peak and current.


sipsma (commented):
> this PR adds telemetry for current memory usage. peak might be relevant here, too, or even preferable? could also do swap if we want?

Yeah I'd take both current + peak. I was wondering if we'd even need peak since we could just show the max "current" we've seen, but since we only sample every 3s peak might still catch spikes that would otherwise be missed.

Swap isn't high on my priority list (until proven otherwise).

cwlbraa:
current and peak working, forgoing swap for now.

cwlbraa:
squashed, rebased, and added release notes

PR #8883, opened 2024-11-07
feat(engine): make metrics in ui more readable

cwlbraa:
this PR makes the tui print human-readable byte and duration values. 

any suggestions on how to manually test the pressure duration printing? I never have these print on `dagger call check`.

bytes, however, are confirmed working and looking good: `exec sh -c git checkout -q $(git rev-parse HEAD) 0.4s | Disk Read Bytes: 45 MB | Disk Write Bytes: 12 kB`

### DIFF
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -11,6 +11,7 @@ import (
 
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
+	"github.com/dustin/go-humanize"
 	"github.com/muesli/termenv"
 	"github.com/opencontainers/go-digest"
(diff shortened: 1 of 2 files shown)


sipsma:
Nice!

> any suggestions on how to manually test the pressure duration printing? I never have these print on dagger call check.

Oh interesting, I think those being present in the cgroup might be dependent on kernel config, so possible you aren't getting them. I'll pull this locally and give it a quick check.

sipsma (approved):
> > any suggestions on how to manually test the pressure duration printing? I never have these print on dagger call check.
> Oh interesting, I think those being present in the cgroup might be dependent on kernel config, so possible you aren't getting them. I'll pull this locally and give it a quick check.

Tried it out locally and looks good!

PR #8885, opened 2024-11-08
tests: deflake cache volume tests from pruning

sipsma:
I saw a few of these tests flake today due to cache volumes not having contents, which would be entirely possible by the cache volume being pruned by buildkit in between usage. There was a recent fix to a platform test to handle this and it hasn't flaked since; just applying that fix to a few more tests now.


vito (approved):
Nice

PR #8888, opened 2024-11-08
Dagger module: github.com/dagger/dagger/helm

shykes:
The official Dagger Helm chart, as a Dagger module.

- `lint`: lint the helm chart
- `test`: test the helm chart against an ephemeral k3s
- `publish`: publish the helm chart


helm/.dagger/main.go
```diff
@@ -128,55 +140,36 @@ func (h *Helm) SetVersion(
...
-			tag:         "helm/chart/" + version,
-			target:      tag,
-			notes:       changeNotes(h.Dagger.Src, "helm/dagger", version),
-			gitRepo:     gitRepoSource,
-			githubToken: githubToken,
-			dryRun:      dryRun,
-		}); err != nil {
-			return err
-		}
-	}
```
jedevc, line 179: Wait, why is this removed? :thinking: This is a critical part of the release process for this component, we need to do this, or the tag+release will not actually be created. This is blocking, sorry!

I'm happy to work on this a bit :tada:, since I'm doing release things right now, so some options:
- We move `githubRelease` into a common shared module
- Or, each component that gets published isn't responsible for doing this bit - instead, this would get done from a top-level `release` module, that would call each individual components `Publish` function.
	- I *think* this would be my preference.
- Or, we just unblock temporarily, and copy-paste the code into here as well.

This is a good component to think about how we want the releasing structure of our modules to look like, since it's relatively much more self contained, so I think worth thinking about it a bit more.

helm/.dagger/main.go
```diff
@@ -28,34 +38,37 @@ func (h *Helm) Lint(ctx context.Context) error {
...
 	test, err := h.chart().
-		WithMountedFile("/usr/bin/dagger", cli).
+		WithMountedFile("/usr/bin/dagger", dag.DaggerCli().Binary()).
 		WithServiceBinding("helm-test", k3ssvc).
 		WithFile("/.kube/config", k3s.Config()).
 		WithEnvVariable("KUBECONFIG", "/.kube/config").
 		WithEnvVariable("CACHEBUSTER", identity.NewID()).
 		WithExec([]string{"kubectl", "get", "nodes"}).
-		WithExec([]string{"helm", "install", "--wait", "--create-namespace", "--namespace=dagger", "--set=engine.image.ref=registry.dagger.io/engine:main", "dagger", "."}).
+		WithExec([]string{"helm", "install", "--wait", "--create-namespace", "--namespace=dagger", "dagger", "."}).
```
jedevc, line 56: Setting this image is deliberate, we should keep it as is:
- https://github.com/dagger/dagger/pull/8305

jedevc (changes requested):
Looks mostly good to me :tada: I did leave a blocking comment on the releasing - this does break our releasing flow :cry: so we should discuss it in there.

Also, pushed a separate commit to split out the helm part of https://github.com/dagger/dagger/pull/8995, since we're in the area.

helm/.dagger/main.go
shykes: Probably a mistake when copying files over. Probably that snippet predates my original PR.

helm/.dagger/main.go
shykes: also done by error, let me clean up real quick

helm/.dagger/main.go
shykes: Damnit, this pulls an additional dependency from the spaghetti, let me see if I can chop it off cleanly.

shykes:
@jedevc with `./releaser` merged, this should be ready to go

jedevc (approved):
LGTM, brought back the `helm.gen.yml` github changes I made that got lost somewhere, and also updated a few subtle things in the releasing that have changed in the meantime.

PR #8895, opened 2024-11-08
docs: TS add `type` & `enum` keyword in examples

TomChv:
Follow up for #8824 (must be merged before ; this PR's branch is based on that one)

### DIFF
--- a/docs/current_docs/api/custom-types.mdx
+++ b/docs/current_docs/api/custom-types.mdx
@@ -40,7 +40,7 @@ that returns a custom `Organization` type, itself containing a collection of
 ```typescript file=./snippets/custom-types/typescript/index.ts
 ```
 
-TypeScript has multiple ways to support complex data types. However, the Dagger TypeScript SDK only supports it through fields (decorated with `@func()`) inside classes because `type` and `interface` keywords do not allows decorators.
+TypeScript has multiple ways to support complex data types. Use a `class` when you need methods and privacy, use `type` for plain data objects with only public fields.
 </TabItem>
 </Tabs>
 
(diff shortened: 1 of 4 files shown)


docs/current_docs/api/custom-types.mdx
```diff
@@ -40,7 +40,7 @@ that returns a custom `Organization` type, itself containing a collection of
 ```typescript file=./snippets/custom-types/typescript/index.ts
 ```
 
-TypeScript has multiple ways to support complex data types. However, the Dagger TypeScript SDK only supports it through fields (decorated with `@func()`) inside classes because `type` and `interface` keywords do not allows decorators.
+TypeScript has multiple ways to support complex data types. Use a `class` when you need methods and privacy, use `type` for plain data objects with only public fields.
```
vikram-dagger, line 43: @TomChv should the Python and Go sections also be updated?

docs/current_docs/api/snippets/custom-types/typescript/index.ts
```diff
@@ -19,30 +19,31 @@ class Account {
   }
 }
 
-@object()
-class Organization {
-  @func()
+/**
```
vikram-dagger, line 22: @TomChv should the Python and Go snippets also be updated?

docs/current_docs/api/custom-types.mdx
helderco: No, this is specific to TypeScript. For example, in Python you use a class for both, and in Go you also use the same struct for both, so no need to mention.

docs/current_docs/api/snippets/custom-types/typescript/index.ts
helderco: No, this change basically allows TypeScript to define simple objects in other ways that feel more natural to TypeScript developers that want to use them. So it's a "multiple ways of doing the same thing" kind of thing.

If a user wants to use a "simple object", then it can't have functions. It's just fields, and they're all exposed to the Dagger API. For this use case that's probably what the user wants, the problem is if they then want to add a function they'd have to change the type to a class, but I think that would be expected for a TypeScript user as it's not Dagger specific.

PR #8898, opened 2024-11-08
feat(engine): make metrics in ui more readable

cwlbraa:
does what it says on the tin, but I'm not sure how to manually test the pressure stat display, they never show up for me. any hints?


sipsma:
We merged this here :-) https://github.com/dagger/dagger/pull/8883

cwlbraa:
oh lul i was confused why I couldn't find it

PR #8902, opened 2024-11-08
feat(engine): telemetry for netns bytes rx and tx, plus packet drops

cwlbraa:
does what it says on the tin, using BK's network namespace "Sample" function.

note this prints % dropped, not % loss. packet loss is harder to instrument, especially on the rx side. if we feel the need, we could probably get stats about tcp retransmissions, but that is out-of-scope here.

### DIFF
--- a/dagql/idtui/golden_test.go
+++ b/dagql/idtui/golden_test.go
@@ -342,6 +342,12 @@ var scrubs = []scrubber{
 		" | Disk Read: 9.3 kB",
 		"",
 	},
+	// Network telemetry
+	{
+		regexp.MustCompile(`\s?\| Network (Tx|Rx): \d+(\.\d+)?\s(B|kB|MB|GB|TB)(\s\(\d+(\.\d+)?\% dropped\))?`),
+		" | Network Tx: 3 kB (0.1% dropped)",
+		"",
+	},
 }
 
 func TestScrubbers(t *testing.T) {
--- a/.changes/unreleased/Added-20241114-123823.yaml
+++ b/.changes/unreleased/Added-20241114-123823.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: 'engine: added network telemetry for execs'
+time: 2024-11-14T12:38:23.340204-08:00
+custom:
+    Author: cwlbraa
+    PR: "8902"
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -401,43 +401,61 @@ func (r renderer) renderMetrics(out *termenv.Output, span *dagui.Span) {
 		return
 	}
 
-	if dataPoints := span.MetricsByName[telemetry.IOStatDiskReadBytes]; len(dataPoints) > 0 {
-		lastPoint := dataPoints[len(dataPoints)-1]
-		fmt.Fprint(out, " | ")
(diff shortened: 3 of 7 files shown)


vito:
> still testing, open to ideas on how to simulate packet loss... maybe just pull something big and far away?

I've had success with `netem` for this: https://stackoverflow.com/questions/614795/simulate-delayed-and-dropped-packets-on-linux

cwlbraa:
netem packet loss doesn't appear to effect BK netns stats, even when it's active. tested the display by hardcoding "dropped" to 1 in the netstat sampler and it printed well, like this: ```
✔ exec bash -c echo '-------------------' && ping -c 10 8.8.8.8 &&for i in {1..10}; do echo "Request $i:" && curl -v -s -w '\nHTTP Status: %{http_code}\nTime: %{time_total}s\n' http://httpstat.us/200 || true; echo '-------------------'; sleep 1; done 20.8s | Disk Read: 1.7 MB | Disk Write: 0 B | Network Rx: 101 kB (0.0697% loss) | Network Tx: 23 MB (0.0625% loss)
```

**Review by sipsma — APPROVED** [review]
LGTM!

**Comment by cwlbraa** [comment]
@sipsma this also refactors the tui code quite a bit, hoping to rebase it on top of the memory stats once that's merged.

**Comment by cwlbraa** [comment]
added release note

**Comment by sipsma** [comment]
@cwlbraa looked at the failing tests here, I think it's legit-ish though not a problem with this PR: https://dagger.cloud/dagger/traces/70db22f796c415b7e5328358aa90a27f?span=35af6147ddc09b8d

@vito I think metrics are going to make the telemetry golden tests sort of hard to maintain since that test is failing due to there being some network usage (and thus including that in the output and creating a diff w/ the golden example). Is there an easy way to make those tests ignore the metrics output, since that is unlikely to ever be perfectly reproducible? Could we just trim metrics from the output before doing the comparison?

**Comment by cwlbraa** [comment]
@vito found the scrubbers, handling this merge conflict myself

**Comment by vito** [comment]
Yep yep - the scrubbers (a delightful bag of about two dozen regexes) are the solution to that. :)

File: dagql/idtui/golden_test.go [hunk 1]
```diff
@@ -330,6 +330,12 @@ var scrubs = []scrubber{
 		"docker.io/library/alpine:latest@sha256:beefdbd8a1da6d2915566fde36db9db0b524eb737fc57cd1367effd16dc0d06d",
 		"sha256:XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
 	},
+	// Bytes
+	{
+		regexp.MustCompile(`\d+(\.\d+)?\s(B|kB|MB|GB|TB)`),
+		"9.3 kB",
+		"X.X B",
+	},
```
cwlbraa, line 338: this works, I'm a little worried it might be flaky in the 0B case where it's not supposed to display.

cwlbraa: indeed, it is flaky, gonna fix it to instead scrub the whole thing

cwlbraa: fixed, including memory and io, in https://github.com/dagger/dagger/pull/8902

PR #8933, opened 2024-11-12
tests: fix race condition in testctx error setting

sipsma:
Saw the race detector go off in [a test run](https://dagger.cloud/dagger/traces/b536e8f790c81fd17ff163926a0e0006?span=f8747f05b83c6abb&logs). Looks like we were just appending to a slice concurrently without any locks.


vito (approved):
whoops - thanks!

Issue #8934, opened 2024-11-12
Ephemeral `context deadline exceeded` errors during Go SDK codegen


sipsma:
I've seen this error popping up randomly on occasion: https://dagger.cloud/dagger/traces/866688214ddc91ea4a13dbbd6aa40db1?span=2928af623c703d97#2928af623c703d97:L11

```
	Error Trace:	/app/core/integration/module_test.go:1172
	            				/app/testctx/testctx.go:163
	Error:      	Received unexpected error:
	            	input: container.from.withExec.withExec.withExec.withMountedFile.withWorkdir.withExec.withMountedFile.withWorkdir.withNewFile.withExec.withWorkdir.withNewFile.withExec.withWorkdir.withExec.withExec.withExec.withNewFile.withEnvVariable.withExec.stdout resolve: process "dagger install ./bar" did not complete successfully: exit code: 1
	            	
	            	Stderr:
	            	1   : connect
	            	1   : connect DONE [0.0s]
	            	
	            	Error: input: moduleSource.withContextDirectory.asModule resolve: failed to create module: select: failed to update codegen and runtime: failed to generate code: failed to get modified source directory for go module sdk codegen: select: process "codegen --output /src --module-context-path /src/bar --module-name bar --introspection-json-path /schema.json" did not complete successfully: context deadline exceeded
```

Haven't noticed any pattern in which test cases.

PR #8936, opened 2024-11-12
engine: make otel proxy shutdown error non-fatal

sipsma:
Fixes https://github.com/dagger/dagger/issues/8934

I haven't tracked down why, but it appears that sometimes in our test
workflows the "shutdown otel proxy" cleanup step takes over 10s, hits
its timeout and then causes the exec to fail.

It shouldn't take that long, but in the meantime there's no reason for
this to be fatal to the exec. The error is now just logged.

---

@vito let me know what you think of this workaround. I'm a bit mystified and concerned why shutting down the otel proxy server is taking >10s sometimes. The only recent change I can think of that's plausibly related is getting rid of the transactions, which I suppose could *somehow* be causing the http handlers take longer? Maybe? Pretty shot in the dark guess.

But for now it at least stops this from breaking user execs.

### DIFF
--- a/engine/buildkit/cleanup.go
+++ b/engine/buildkit/cleanup.go
@@ -2,6 +2,7 @@ package buildkit
 
 import (
 	"errors"
+	"fmt"
 	"sync"
 	"time"
 
@@ -23,6 +24,7 @@ func (c *Cleanups) Add(msg string, f func() error) CleanupFunc {
 		err := f()
 		if err != nil {
 			slog.Error("cleanup failed", "msg", msg, "err", err, "duration", time.Since(start))
+			err = fmt.Errorf("cleanup failed: %q: %w", msg, err)
 		} else {
 			slog.ExtraDebug("cleanup succeeded", "msg", msg, "duration", time.Since(start))
 		}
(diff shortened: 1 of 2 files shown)


sipsma:
Caught it after a few re-runs: https://dagger.cloud/dagger/traces/6d46046448a86fe4dcc83f4e0f9761b6?span=f64e19849bc74e52&logs

Was indeed a cleanup func, specifically `shutdown otel proxy`.

Need to look into why this started happening recently. Not sure if something is just taking longer than it used to (maybe related to more otel data flowing due to metrics?) or if something is actually going wrong.

Either way, that doesn't seem like something that should be fatal to an exec at least at face value.

vito (approved):
> @vito let me know what you think of this workaround. I'm a bit mystified and concerned why shutting down the otel proxy server is taking >10s sometimes. The only recent change I can think of that's plausibly related is getting rid of the transactions, which I suppose could somehow be causing the http handlers take longer? Maybe? Pretty shot in the dark guess.

...Huh, yeah that's concerning. Looking into it now. Do you happen to have a repro, or is it more possibly load-induced, so just "run the whole suite and keep an eye on the logs"?

That [thread](https://discord.com/channels/707636530424053791/1300880185326964797) about `begin tx: database is locked` is also ongoing and possibly demonstrates the same issue after removing the transaction.

Might be time to add some spans to our OTel data pipeline. :yodawg:

I'm good with merging this since telemetry failures shouldn't bubble up to errors in userland imo.

vito:
Follow-up: https://github.com/dagger/dagger/pull/8950

Issue #8932, opened 2024-11-12
✨ shell: dynamic module loading


helderco:
> [!note]
> - Follow-up from: https://github.com/dagger/dagger/pull/8611

## Motivation

Currently, you can start `dagger shell` with `--no-load` to skip loading a module if all you want is core functions. However, if it turns out you want to load the module after all, you have to exit the shell and re-run without the flag.

Would be useful to be able to load the module from within the interactive interpreter. Even better if you can load any module, not just the current one.

## Specification

Introduce two built-ins:
- `.load [module]`
    - load a module and return its introspection
    - if module is omitted, assume current module
- `.use`
    - chained after `.load`, sets the loaded module as the default for the current session

### Examples

Started the shell on a module directory with `dagger shell --no-load`, but now want to load the current module and set it globally:
```console
⋈ .load | .use
```

See documentation for an external module:
```console
⋈ .load github.com/shykes/gha | .doc
```

Run a one-off function from another module:
```console
⋈ .load github.com/shykes/hello | hello
hello world
⋈ .load sdk/python/dev | lint
...
```

Set another module as the global for current session:
```console
⋈ .load github.com/shykes/hello | .use
⋈ hello
hello world
```

## Rationale

The reason for adding `.use` is to make `.load` not create a side effect. This way you can quickly check the documentation for a module, or execute a one-off function, without leaving the current interactive session.

So `.use` gives an opt-in to set the global state to any module, not just the current one.

PR #8938, opened 2024-11-12
feat(shell): load module dynamically

helderco:
Closes https://github.com/dagger/dagger/issues/8932

## Prompt

Since we can now load different modules within the same session, the prompt now shows the default module (global), if loaded:

Screenshot 2024-11-15 at 21 33 00

## Colors

In order to be easier to spot, errors are now shown in red:

Screenshot 2024-11-15 at 21 36 31

Thanks to @cwlbraa for this idea, the prompt symbol (⋈) is also turned red if the last command errored. It's green otherwise.

## Fixes

- A few core functions like `.cache-volume` weren't showing, that's now fixed
- Module description is now shown when using `.doc` on the main object, and constructor output improved
- A non existing function won't fallback to core anymore since core functions are readily available with the `.` prefix

## Todo

- [x] Add `.load` builtin to load module when `--no-load` was used
- [x] Add `.use` builtin and pass introspection through state
- [x] Load any module
- [x] Tests

### DIFF
--- a/cmd/dagger/call.go
+++ b/cmd/dagger/call.go
@@ -70,7 +70,7 @@ available functions.
 			// Walk the hypothetical function pipeline specified by the args
 			for _, field := range cmd.Flags().Args() {
 				// Lookup the next function in the specified pipeline
-				nextFunc, err := mod.GetFunction(o, field)
+				nextFunc, err := GetSupportedFunction(mod, o, field)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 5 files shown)


shykes:
definitely a killer feature.

Question: what's the inverse of `.use`? Exit just exits the whole shell. Would be cool if it were like a subshell? Not sure we can hook into ctrl-d?

helderco:
> Question: what's the inverse of `.use`? 

That's a good question. I thought about it several times, but haven't formed any idea. Maybe `.use` on its own to basically say "use core"?

It's just changing a string (module ref) in a variable (`""` means core).


> Exit just exits the whole shell. Would be cool if it were like a subshell? Not sure we can hook into ctrl-d?

That's interesting, and we could use ctrl+d multiple times to undo the "stack", so it would take twice to exit the shell after undoing a `.use`. However I'm concerned about this possibly being cumbersome to unwind. If you start the shell with a loaded module (the default), wouldn't it take two `Ctrl+D` to exit the shell (or type "exit")? What if you use `.use` after another `.use`?

It's an easy change but I guess I'm wondering what's the value in it?

helderco:
Btw, as an implementation detail, if you do `.load my-module` twice, it won't reload a second time. It stays in a (memory) cache. So if you want to go back and forth it's very performant. 

However, I wonder if there's a need to force the reload, I mean before https://github.com/dagger/dagger/issues/8987 is resolved.

shykes (approved):
I think we should merge this as-is, and continue discussing follow-up improvements. Easier that way.

PR #8945, opened 2024-11-13
cli: Alias `fs` to `functions`

gerhard:
So that pro CLI users can run:

    d fs

Instead of:

    dagger functions

Because **pro** CLI users already alias `d` to `dagger`

---

image

---

In the absence of this:

```console
which dfs
dfs: aliased to dagger functions
```


gerhard:
OK to close, alias works for me.

PR #8948, opened 2024-11-13
fix(engine): secrets created inside module constructor fail on repeated accesses

### LINKED ISSUE #8935: 🐞 secrets created inside module constructor fail on repeated accesses
### What is the issue?

Reported by @@vmaffet-fast [in discord](https://discord.com/channels/707636530424053791/1304065468583379016), with a lovely repro.

> I've been getting weird errors when creating secrets inside a module's constructor and trying to use them later on. 
> The actual use case is exchanging OIDC tokens for a token authenticating to a service, which I then need to use later on.
> Here is a reproduction example:


### Dagger version

unclear, 2024/11/7

### Steps to reproduce

I have a module that generates and stores a secret on initialization:
```go
package main

import (
    "context"

    "dagger/generator/internal/dagger"
)

type Generator struct {
    // +private
    Password *dagger.Secret
}

func New() *Generator {
    return &Generator{
        Password: dag.SetSecret("pass", "admin"),
    }
}

func (m *Generator) Gen(ctx context.Context, name string) error {
    _, err := m.Password.Plaintext(ctx)
    return err
}
```

I have a second module that depends on the first one and wraps calls to it like so:
package main
```go
import (
    "context"
)

type Keychain struct{}

func (m *Keychain) Get(ctx context.Context, name string) error {
    return dag.Generator().Gen(ctx, name)
}
```


Finally, I use the second module from a contextual module:
```go
package main

import (
    "context"
    "fmt"
)

type Mymodule struct{}

func (m *Mymodule) Issue(ctx context.Context) error {
    kc := dag.Keychain()

    err := kc.Get(ctx, "a")
    if err != nil {
        return fmt.Errorf("first get: %w", err)
    }

    err = kc.Get(ctx, "b")
    if err != nil {
        return fmt.Errorf("second get: %w", err)
    }
    return nil
}
```


The first call to kc.Get is succesful. 
However, the second one fails with the log output below.

### Log output

```
dagger call -m mymodule issue
✔ connect 0.3s
✔ initialize 0.5s
✔ prepare 0.0s
✔ mymodule: Mymodule! 0.0s
✘ Mymodule.issue: Void 0.7s
! call function "Issue": process "/runtime" did not complete successfully: exit code: 2
┃ invoke: second get: input: keychain.get resolve: call function "Get": process "/runtime" did not complete successfully: exit code: 2                                                                                                                                                                                      
┃                                                                                                                                                                                                                                                                                                                           
┃ Stdout:                                                                                                                                                                                                                                                                                                                   
┃ invoke: input: generator.gen resolve: call function "Gen": process "/runtime" did not complete successfully: exit code: 2                                                                                                                                                                                                 
┃                                                                                                                                                                                                                                                                                                                           
┃ Stdout:                                                                                                                                                                                                                                                                                                                   
┃ get parent name: input: get or init client: initialize client: failed to add client resources from ID: failed to add secret from source client byn8ttu7ly3h7zul71p7ncibf: secret xxh3:cf46d07f3c9661af not found in other store                                                                                           
  ✔ Keychain.get(name: "a"): Void 0.4s
  ✘ Keychain.get(name: "b"): Void 0.2s
  ! call function "Get": process "/runtime" did not complete successfully: exit code: 2
  ┃ invoke: input: generator.gen resolve: call function "Gen": process "/runtime" did not complete successfully: exit code: 2                                                                                                                                                                                               
  ┃                                                                                                                                                                                                                                                                                                                         
  ┃ Stdout:                                                                                                                                                                                                                                                                                                                 
  ┃ get parent name: input: get or init client: initialize client: failed to add client resources from ID: failed to add secret from source client byn8ttu7ly3h7zul71p7ncibf: secret xxh3:cf46d07f3c9661af not found in other store                                                                                         
    ✘ Generator.gen(name: "b"): Void 0.1s
    ! call function "Gen": process "/runtime" did not complete successfully: exit code: 2
    ┃ get parent name: input: get or init client: initialize client: failed to add client resources from ID: failed to add secret from source client byn8ttu7ly3h7zul71p7ncibf: secret xxh3:cf46d07f3c9661af not found in other store                                                                                       

Setup tracing at https://dagger.cloud/traces/setup. To hide: export STOPIT=1

Error: response from query: input: mymodule.issue resolve: call function "Issue": process "/runtime" did not complete successfully: exit code: 2

Stdout:
invoke: second get: input: keychain.get resolve: call function "Get": process "/runtime" did not complete successfully: exit code: 2

Stdout:
invoke: input: generator.gen resolve: call function "Gen": process "/runtime" did not complete successfully: exit code: 2

Stdout:
get parent name: input: get or init client: initialize client: failed to add client resources from ID: failed to add secret from source client byn8ttu7ly3h7zul71p7ncibf: secret xxh3:cf46d07f3c9661af not found in other store
```

cwlbraa:
intended to fix #8935, turns out this was a problem with constructor purity

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -3007,9 +3007,14 @@ func (m *Mymodule) Issue(ctx context.Context) error {
         return fmt.Errorf("first get: %w", err)
     }
 
+    err = kc.Get(ctx, "a")
+    if err != nil {
+        return fmt.Errorf("second get, same args: %w", err)
+    }
+
     err = kc.Get(ctx, "b")
     if err != nil {
-        return fmt.Errorf("second get: %w", err)
+        return fmt.Errorf("third get: %w", err)
     }
     return nil
 }
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -2934,6 +2934,92 @@ func (m *Test) Test(ctx context.Context) (string, error) {
 			require.Equal(t, "HELLO FROM FOO", out)
 		})
 
+		t.Run("double nested and called repeatedly", func(ctx context.Context, t *testctx.T) {
+			c := connect(ctx, t)
+			ctr := c.Container().From(golangImage).
+				WithMountedFile(testCLIBinPath, daggerCliFile(t, c))
+
+			// Set up the base generator module
+			ctr = ctr.
+				WithWorkdir("/work/keychain/generator").
+				With(daggerExec("init", "--name=generator", "--sdk=go", "--source=.")).
+				WithNewFile("main.go", `package main
+
+import (
+    "context"
+    "dagger/generator/internal/dagger"
+)
+
+type Generator struct {
+    // +private
+    Password *dagger.Secret
+}
+
+func New() *Generator {
+    return &Generator{
+        Password: dag.SetSecret("pass", "admin"),
+    }
+}
+
+func (m *Generator) Gen(ctx context.Context, name string) error {
+    _, err := m.Password.Plaintext(ctx)
+    return err
+}
+`)
+
+			// Set up the keychain module that depends on generator
+			ctr = ctr.
+				WithWorkdir("/work/keychain").
+				With(daggerExec("init", "--name=keychain", "--sdk=go", "--source=.")).
+				With(daggerExec("install", "./generator")).
+				WithNewFile("main.go", `package main
+
+import (
+    "context"
+)
+
+type Keychain struct{}
+
+func (m *Keychain) Get(ctx context.Context, name string) error {
+    return dag.Generator().Gen(ctx, name)
+}
+`)
+
+			// Set up the main module that uses keychain
+			ctr = ctr.
+				WithWorkdir("/work").
+				With(daggerExec("init", "--name=mymodule", "--sdk=go", "--source=.")).
+				With(daggerExec("install", "./keychain")).
+				WithNewFile("main.go", `package main
+
+import (
+    "context"
+    "fmt"
+)
+
+type Mymodule struct{}
+
+func (m *Mymodule) Issue(ctx context.Context) error {
+    kc := dag.Keychain()
+
+    err := kc.Get(ctx, "a")
+    if err != nil {
+        return fmt.Errorf("first get: %w", err)
+    }
+
+    err = kc.Get(ctx, "b")
+    if err != nil {
+        return fmt.Errorf("second get: %w", err)
+    }
+    return nil
+}
+`)
+
+			// Test that repeated calls work correctly
+			_, err := ctr.With(daggerCall("issue")).Sync(ctx)
+			require.NoError(t, err)
+		})
+
 		t.Run("cached", func(ctx context.Context, t *testctx.T) {
 			c := connect(ctx, t)
 			ctr := c.Container().From(golangImage).
--- a/core/object.go
+++ b/core/object.go
@@ -336,13 +336,6 @@ func (obj *ModuleObject) installConstructor(ctx context.Context, dag *dagql.Serv
 
 	spec.Name = gqlFieldName(mod.Name())
 
-	// NB: functions actually _are_ cached per-session, which matches the
-	// lifetime of the server, so we might as well consider them pure.
-	// That way there will be locking around concurrent calls, so the user won't
-	// see multiple in parallel. Reconsider if/when we have a global cache and/or
-	// figure out function caching.
-	spec.ImpurityReason = ""
-
 	spec.Module = obj.Module.IDModule()
 
 	if fn.metadata.SourceMap != nil {
--- a/engine/server/session.go
+++ b/engine/server/session.go
@@ -434,7 +434,7 @@ func (srv *Server) initializeDaggerClient(
 		// of the caller
 		for _, id := range opts.ParentIDs {
 			if err := srv.addClientResourcesFromID(ctx, client, id, opts.CallerClientID, false); err != nil {
-				return fmt.Errorf("failed to add client resources from ID: %w", err)
+				return fmt.Errorf("failed to add parent client resources from ID: %w", err)
 			}
 		}
 	}


cwlbraa:
poking around and adding debug logs, on the second call, the secret store we're trying to fetch from really does not have the secret we're looking for, in fact it has none. sorry for the munged af logs:

first, successful call:
```
  ┃ time="2024-11-15T19:30:51Z" level=debug msg="!!! AddSecretFromOtherStore" client_hostname=buildkitsandbox client_id=vowuw8q8aap10qxlyc9k5khfm session_id=1i79h53783
  ┃ w5xnx1y22m7 spanID=5173a77d32b6f779 traceID=19beae6ed4bf10a3f6da4a2cb8a4749a
  ┃ time="2024-11-15T19:30:51Z" level=debug msg="!!! secret: &core.Secret{Query:(*core.Query)(0x4007252e40), IDDigest:\"xxh3:a304521904080314\"}" client_hostname=build
  ┃ andbox client_id=vowuw8q8aap10qxlyc9k5khfm session_id=1i79h53783ae2ow5xnx1y22m7 spanID=5173a77d32b6f779 traceID=19beae6ed4bf10a3f6da4a2cb8a4749a
  ┃ time="2024-11-15T19:30:51Z" level=debug msg="!!! store: &core.SecretStore{secrets:map[digest.Digest]*core.storedSecret{}, mu:sync.RWMutex{w:sync.Mutex{state:0, sem
  ┃ 0}, writerSem:0x0, readerSem:0x0, readerCount:atomic.Int32{_:atomic.noCopy{}, v:0}, readerWait:atomic.Int32{_:atomic.noCopy{}, v:0}}}" client_hostname=buildkitsand
  ┃ client_id=vowuw8q8aap10qxlyc9k5khfm session_id=1i79h53783ae2ow5xnx1y22m7 spanID=5173a77d32b6f779 traceID=19beae6ed4bf10a3f6da4a2cb8a4749a
  ┃ time="2024-11-15T19:30:51Z" level=debug msg="!!! otherStore: &core.SecretStore{secrets:map[digest.Digest]*core.storedSecret{\"xxh3:a304521904080314\":(*core.stored
  ┃ et)(0x400bb5e4e0)}, mu:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:atomic.Int32{_:atomic.noCopy{}, v:0}, readerWait:ato
  ┃ Int32{_:atomic.noCopy{}, v:0}}}" client_hostname=buildkitsandbox client_id=vowuw8q8aap10qxlyc9k5khfm session_id=1i79h53783ae2ow5xnx1y22m7 spanID=5173a77d32b6f779 t
  ┃ ID=19beae6ed4bf10a3f6da4a2cb8a4749a
```

second call (note the empty map)
```
  ┃ time="2024-11-15T19:30:51Z" level=debug msg="!!! AddSecretFromOtherStore" client_hostname=buildkitsandbox client_id=o53fav0csil31n2ktdntgnta3 session_id=1i79h53783
  ┃ w5xnx1y22m7 spanID=7b13e12b25adcc21 traceID=19beae6ed4bf10a3f6da4a2cb8a4749a
  ┃ time="2024-11-15T19:30:51Z" level=debug msg="!!! secret: &core.Secret{Query:(*core.Query)(0x4007252e40), IDDigest:\"xxh3:a304521904080314\"}" client_hostname=build
  ┃ andbox client_id=o53fav0csil31n2ktdntgnta3 session_id=1i79h53783ae2ow5xnx1y22m7 spanID=7b13e12b25adcc21 traceID=19beae6ed4bf10a3f6da4a2cb8a4749a
  ┃ time="2024-11-15T19:30:51Z" level=debug msg="!!! store: &core.SecretStore{secrets:map[digest.Digest]*core.storedSecret{}, mu:sync.RWMutex{w:sync.Mutex{state:0, sem
  ┃ 0}, writerSem:0x0, readerSem:0x0, readerCount:atomic.Int32{_:atomic.noCopy{}, v:0}, readerWait:atomic.Int32{_:atomic.noCopy{}, v:0}}}" client_hostname=buildkitsand
  ┃ client_id=o53fav0csil31n2ktdntgnta3 session_id=1i79h53783ae2ow5xnx1y22m7 spanID=7b13e12b25adcc21 traceID=19beae6ed4bf10a3f6da4a2cb8a4749a
  ┃ !!! destClient: (*server.daggerClient)<Client o53fav0csil31n2ktdntgnta3: uninitialized>!!! srcClient: (*server.daggerClient)<Client f9qyfciabzvi9c15f5prm8yaa: init
  ┃ zed>time="2024-11-15T19:30:51Z" level=debug msg="!!! otherStore: &core.SecretStore{secrets:map[digest.Digest]*core.storedSecret{}, mu:sync.RWMutex{w:sync.Mutex{sta
  ┃ , sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:atomic.Int32{_:atomic.noCopy{}, v:0}, readerWait:atomic.Int32{_:atomic.noCopy{}, v:0}}}" client_hostname=bui
  ┃ tsandbox client_id=o53fav0csil31n2ktdntgnta3 session_id=1i79h53783ae2ow5xnx1y22m7 spanID=7b13e12b25adcc21 traceID=19beae6ed4bf10a3f6da4a2cb8a4749a
```

I've climbed the stack a bit, but don't really understand why the src client would have an empty secret store like this and they don't carry quite carry enough data to identify where they came from... each call seems to be a new session.

cwlbraa:
I climbed the stack a bit more just now, found that the failure to add the secret from the other store occurs specifically in [client initialization when attempting to add resources from parent calls sourced from the caller client](https://github.com/cwlbraa/dagger/blob/ef6f8d8707bc698ce1971751d60ee1cf6d524a60/engine/server/session.go?plain=1#L432-L440)

Interestingly, I've also modified the test to do 2 gets of "a" followed by one of "b"... the failure only occurs on the get of "b", even though the value is the same for both calls. My understanding is that this means that cache-keying on fn arguments either avoids this error entirely for repeated calls with the same input or possibly causes it somehow, like maybe we expect to find this resource cached bc on one level we've already got it, but it's keyed differently on the "b" calls? tbh that feels like the wrong track simply because when we get further down the stack, we find that we're trying to pull our desired secret from an empty secret store.

cwlbraa:
did some real crazy log processing of the added test and came out with at least a hypothesis. From my log processing, it looks like the "Issue" client never gets the secret from the "Keychain" clients copied to its, but when Issue goes back to call Keychain repeatedly with different args, the engine expects to find that secret in its secret store.

I think the problem here, with nesting, is that the Secret resource is _not_ returned from the test keychain's "Get" method, but the code expects that it's propagated up to the caller client SecretStore, "Issue" in the case the of of the test.

messy notes attached for posterity, came up with this by very carefully extracting client IDs from the logs and then timelining the events. see the added test if you want this to be anything other than gibberish.

cc @jedevc as well, because this feels very entangled with #8358 and the fn.root.AddClientResourcesFromIDs call in ModuleFunction.Call.
```
Hl94 -> actual parent
24wv -> keychain
M1ap -> generator
Tddrb -> second generator(a) call? if so, why does 24wv get to be re-used but this one doesn't?

Lskg35 -> keychain with b? 
Xlkg -> generator with b?

1. From hl94 to 24wv -> nothing to copy
2. 24wv to m1apmms ? nothing to copy
3. m1apmms to 24wv -> copied 
4. From 24wv to tddrb -> copied
5. hl94 to lskg35 -> nothing to copy
6. From lskg35 to xlgx -> nothing to copy, happens twice, by callID then parent
```

jedevc:
Okay, so I *think* I've pushed a fix for this, but I may be wrong.

I think this is actually the exact same issue as described in https://github.com/dagger/dagger/pull/8358#issuecomment-2346006330 - and the fix works - but it wasn't being applied for constructors (which seems like an oversight on my part).

What's happening in your example from https://github.com/dagger/dagger/issues/8935:
- The second call to `dag.Generator` (with "b") is cached
- But then we try and access the secret in `Gen` - but since that client never actually ran, we have no client to grab the secret from
- The solution is the same as in [`c3b1774` (#8358)](https://github.com/dagger/dagger/pull/8358/commits/c3b177462bc1cdf36335f78ef716897a75af143c). We need to do the secret transfer from the client that was originally cached.

There's more info in the commit I pushed.

cwlbraa:
@jedevc's fix works. repeating my log-processing on the working version, iiuc, allowing constructors to maintain field impurity makes it so the client can source the secret from the generator client that originally returned it instead of looking for it in the "issue" client that never receives it. 

From the original log processing, I don't really understand wtf client tddrb is, I'd have liked it to be hl94, but this should be mergeable as-is.

jedevc (approved):
Approving, but since I wrote the fix, probably would also be good to have someone else take a glance too :laughing:

sipsma (approved):

Issue #8773, opened 2024-10-24
`dagger call ... directory terminal` mysteriously fails with `exit 1`


jedevc:
Using dagger v0.13.5, running against `main`.

```
$ dagger call -m modules/wolfi container directory --path="." terminal
...
exit 1
✔ connect 0.3s
✔ initialize 1.1s
✔ prepare 0.0s
✔ wolfi: Wolfi! 0.0s
✔ Wolfi.container: Container! 3.9s
✔ Container.directory(path: "."): Directory! 0.0s
✔ Directory.terminal: Directory! 0.4s
✔ Directory.digest: String! 0.0s
✔ Directory.entries: [String!]! 0.0s

Full trace at https://dagger.cloud/jedevc/traces/7a8b4b3206772b06a7dedf64845cb745
```

Weirdly enough, setting `--path` to something else works as intended:

```
$ dagger call -m modules/wolfi container directory --path="/usr" terminal
...
dagger /src $
```

As does just running the container directly:

```
$ dagger call -m modules/wolfi container terminal
...
dagger /src $
```

I've also seen this with other modules that return directories:

```
$ dagger call -m modules/gha/examples/go/ gha-on-push terminal
...
exit 1
✔ connect 0.2s
✔ initialize 17.8s
✔ prepare 0.0s
✔ examples: Examples! 0.0s
✔ Examples.ghaOnPush: Directory! 3.3s
✔ Directory.terminal: Directory! 1.1s
✔ Directory.digest: String! 0.0s
✔ Directory.entries: [String!]! 0.0s

Full trace at https://dagger.cloud/jedevc/traces/b9310677146686a977eda240d58c52f0
```

---

Inspecting the engine logs for the failure case:

```
time="2024-10-24T14:03:18Z" level=info msg="service exited" err="map[error:stat mount source overlay: stat overlay: no such file or directory kind:*errors.withStack stack:<nil>]" id="directory(path: \".\")" service="xxh3:6552ef689ad84dab"
time="2024-10-24T14:03:18Z" level=debug msg="handling http request" client_hostname=dragon client_id=pk6z1dz7vdrzz64l4zmma3a74 contentType= method=POST path=/shutdown session_id=wa47md9key6s951pkuud9ydkm span=70fd5ac36b78e3b8 spanID=70fd5ac36b78e3b8 trace=b4cd2d7b9b81292925e0d161ef351eee traceID=b4cd2d7b9b81292925e0d161ef351eee upgradeHeader=
[...]

PR #8952, opened 2024-11-14
Fix broken terminal tests and broken Directory.terminal feature

sipsma:
While debugging an unrelated PR I realized that our terminal tests were not working as expected and that the Directory.terminal feature just didn't work at all:

---
[tests: fix terminal tests not asserting as expected](https://github.com/dagger/dagger/commit/a708114f344e0d10e75d600028d6956fa5e7ddf0)

Our terminal tests were not actually working how we thought they were.
Many of them were not waiting for prompts and instead just matching on
strings right away, which ended up including all the progress output
before the terminal was opened.

This broke a few of them since the string they were expecting happened
to also show up in the progress output, which isn't what we actually
wanted to assert on.

This updates the tests to correctly wait for prompts (doing some
terminal escape sequence handling when needed).

This commit leaves the `TestDaggerTerminal/directory` test case failing
because that feature was actually totally broken due to a separate issue
(which went unnoticed until now because the test case was broken).

---
[engine: fix Directory.terminal and overlay mount handling](https://github.com/dagger/dagger/commit/deb49b901b6cc73ce73fe7229265758180029034)

The previous commit fixed assertions in our terminal test cases, which
revealed that the Directory.terminal feature was totally broken.

The problem was in the executor and it's handling of mounts that were of
type overlay. It was looking at the `Source` field of the mount assuming
it was a bind and that the field pointed to the source of the bind.

However, for non-bind mounts the `Source` is just the type of mount.

The reason this only broke the Directory.terminal case so far was that
Directory.terminal is the only place in our code where we create
read-only mounts. This triggers a different codepath in buildkit's cache
logic that results in actual overlay mounts being provided to the
executor; in other read-write mount cases buildkit has some logic that
creates a "shared" overlay mount which is just bind-mounted to other
places (to avoid conflicts around multiple overlay mounts with the same
upperdir).

This commit just fixes handling of that `Source` field to interpret it
correctly based on whether it's a bind mount or not.

### DIFF
--- a/.changes/unreleased/Fixed-20241113-221407.yaml
+++ b/.changes/unreleased/Fixed-20241113-221407.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: |
+  Directory.terminal API works now
+time: 2024-11-13T22:14:07.998787448-08:00
+custom:
+  Author: sipsma
+  PR: "8952"
--- a/core/integration/legacy_test.go
+++ b/core/integration/legacy_test.go
@@ -130,16 +130,27 @@ func (t *Test) Debug() *dagger.Terminal {
 		err = cmd.Start()
 		require.NoError(t, err)
 
+		prompt := fmt.Sprintf("/coolworkdir%s $ ", resetSeq)
+
+		_, err = console.ExpectString(prompt)
+		require.NoError(t, err)
+
 		_, err = console.SendLine("pwd")
 		require.NoError(t, err)
 
-		_, err = console.ExpectString("/coolworkdir")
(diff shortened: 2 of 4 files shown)


jedevc:
@sipsma does this fix https://github.com/dagger/dagger/issues/8773?

jedevc (approved):
*wut*. Good find, weird that `source` isn't empty in this case :thinking:

sipsma:
> @sipsma does this fix #8773?
> 
> Edit: confirmed that it does!

Oh nice, I had missed that issue!

Issue #8499, opened 2024-09-19
Dagger is slow in large monorepos


marcosnils:
# Problem

ref: https://discord.com/channels/707636530424053791/1286344441950375986

When running dagger in a bug monorepo, because of the default SDK `include` and `exclude` patterns in the `initialize` phase, the engine can take quite some time (> 30s) to initialize. 

I think part of this problem is that we're currently being very optimistic in the patterns that we set "just in case" the user decides to use some of his monorepo code into the module. Maybe a different approach here could be so the user has to specifically add  any other paths that they need in case their module requires to import any of the current monorepo packages.

# Solution

Make DX faster in monorepos.

marcosnils:
cc @helderco not sure how relevant this is to SDK coexistence :pray:

helderco:
See:
- https://github.com/dagger/dagger/pull/8501
- https://github.com/dagger/dagger/issues/8496

helderco:
I cover the problem with the default includes from Go in:
- https://github.com/dagger/dagger/issues/8496

But here’s a simple example that’s not related to that. It’s just where you put your `dagger.json` and whether you have the necessary excludes or not.

_(I pruned the cache before each example)_

## Daggerize

First, let’s try daggerizing a repo by leaving `dagger.json` at the repo’s root:

```
❯ cd $(mktemp -d)
❯ git init
❯ mkdir -p wat/.yarn
❯ mkfile 6g wat/.yarn/cache
❯ echo .yarn/ > .gitignore
❯ dagger init --sdk=typescript --source=.dagger --name=foo
❯ lt -a
.
├── .dagger/...
├── .git/...
├── wat
│  └── .yarn
│     └── cache
├── .gitignore
├── dagger.json
└── LICENSE
```

Notice that I put a “monster” file, in a seemingly unrelated directory, and it’s included in the upload from the host because by default, anything inside the module’s directory (aka root directory, i.e., where `dagger.json` is) is included:

<img width="1032" alt="Screenshot 2024-09-20 at 15 46 05" src="https://github.com/user-attachments/assets/3753dd07-32f0-409b-a163-c43383eaf98e">

This causes compounding issues in the pipeline if you have heavy loads, for example when it needs to be copied because of a `withDirectory`:

<img width="1322" alt="Screenshot 2024-09-20 at 15 49 41" src="https://github.com/user-attachments/assets/f762d4db-e171-4cec-9b6e-736904fb788d">

And it even gets re-exported back to the host in the end:

[...]

Issue #8496, opened 2024-09-19
All SDKs upload all go files in the repo when loading a module


helderco:
## Summary

Stop including all Go packages in a repo with every Go module within it by default, in favor of transferring that decision to the module author.

## Motivation

The biggest reason for doing this is **performance**. It’s also **confusing** to folks that this happens in the first place:
- https://github.com/dagger/dagger/issues/8395

The `RequiredPaths` feature was added in v0.9.9, when Dagger Modules were still experimental[^1]. It allows SDKs to specify files from the repo that they may need in order to load a module.

[^1]: Dagger Functions were officially [released in v0.10.0](https://dagger.io/blog/introducing-dagger-functions).

Currently, **only the Go SDK** uses it:

https://github.com/dagger/dagger/blob/fa9d3c7cceb8abfbab8a79210eb99287c3ca4e99/core/schema/sdk.go#L485-L498

But the **problem** is that those glob patterns match against the **whole git repo**, not just parent directories, and they're **always** applied. Additionally, it includes the `vendor/` directory which is also present for example in PHP projects, and can be quite big. If that directory exists in a given repo, every Go module will load it.

Besides, the Go SDK is used to bootstrap every other SDK, which means the PHP SDK module includes all those files from our repo, including test cases, in every user module’s *initialize* phase, for example.

## Solution

Just **remove** support for the `RequiredPaths` field in SDKs, and make a recommendation for users on how best to integrate with code outside the module.

By **giving the control to the user**, we can **default to not load anything extra** from outside the root directory (in all SDKs), while still **allowing the use case of “importing code” from parent directories**, but only on the modules that **need it**, and only the files that are actually needed. 

It’s also more self-documenting and easier to detect if a module needs files outside of its root.

### Historical context

This issue was [known from the start](https://github.com/dagger/dagger/pull/6575), although it was initially about `go.mod` but then had to include more files. It was necessary in order for the Go SDK to edit an existing parent dir’s `go.mod` on `dagger init`, but that no longer happens (by default) since:
- https://github.com/dagger/dagger/issues/7527
- https://github.com/dagger/dagger/pull/8095

In the [original PR](https://github.com/dagger/dagger/pull/6575)’s description it says:

[...]

Issue #6627, opened 2024-02-08
Use `.gitignore` by default when loading modules


helderco:
## Motivation

I have to constantly add `"exclude": ["sdk", ".venv"] ` to `dagger.json` after `dagger init`.

I don't expect these files to be uploaded. They're not completely replaced in codegen. During local dev, some files can be created in there and they will stay after codegen in the container. It affects the Source dir cache and slows down module loading.

## Solution

The engine already loads `dagger.json` first. Load `.gitignore` too. Prepend the `.gitignore` patterns to the `exclude` patterns from `dagger.json` when collecting the patterns for loading the module’s files.

As an escape hatch, if the user needs to include something in `.gitignore` they can add the same pattern manually to `include` in `dagger.json`.

## Rationale

This seems like the easiest solution, with less work for the user but keeping control.

Doesn’t require extra work. Most modules in the Daggerverse have the automatically generated `.gitignore` but no patterns in `"exclude"`. By using `.gitignore`, those modules would benefit automatically. Of course, this is only necessary when `dagger develop`ing modules locally.

**When a module is loaded via a git reference, those patterns are already applied because the git clone won’t have those files. So using** `.gitignore` **is actually more consistent because it avoids differences in dirty working copies.**

It’s still possible to have dirty woking copies because global ignores won’t be picked up, but rather than adding some of those to `exclude`, users might as well just add them to `.gitignore` which also helps other users contributing to a module that don’t have the same global ignores.

It’s also possible for files in a module to be ignored by some other `.gitignore` up the file tree until the git repo root, but we could automatically find those and add their patterns, prefixed with their file path relative to the context directory, to the module’s excludes.

helderco:
@sipsma, I updated the description. Previously, my thinking was just to add the ignore patterns to the `execlude` in `dagger.json` in `dagger { init | develop } --sdk=`, but wdyt about loading from `.gitignore`?

shykes:
This seems reasonable to me but I may not be aware of all the subtleties of the design.

In general I'm in favor of adopting the standard gitignore *format* whenever we need to express filters on directories (see for example the `ignore` part of #7647 ). This is one step further in the same general direction.

How would the default behavior be disabled or customized?

jpadams:
> @sipsma, I updated the description. Previously, my thinking was just to add the ignore patterns to the `execlude` in `dagger.json` in `dagger { init | develop } --sdk=`, but wdyt about loading from `.gitignore`?

[...]

PR #8953, opened 2024-11-14
Use .dagger-root to improve module load performance on large monorepos (#8496, #8499)

idlsoft:
`dagger` looks for the git repo root when it loads a module.
This has a performance impact, especially on large monorepos.
This change will allow to limit the scope of the module by adding  a `.dagger-root` file.

### DIFF
--- a/core/modulesource.go
+++ b/core/modulesource.go
@@ -518,6 +518,10 @@ func callerHostFindUpContext(
 	if err == nil {
 		return curDirPath, true, nil
 	}
+	_, err = bk.StatCallerHostPath(ctx, filepath.Join(curDirPath, ".dagger-root"), false)
+	if err == nil {
+		return curDirPath, true, nil
+	}
 	// TODO: remove the strings.Contains check here (which aren't cross-platform),
 	// since we now set NotFound (since v0.11.2)
 	if status.Code(err) != codes.NotFound && !strings.Contains(err.Error(), "no such file or directory") {


helderco:
Should open an issue proposing this change before opening a PR, but I don't think this is the right approach.

You mention https://github.com/dagger/dagger/issues/8496 (in the title), but I think the better solution there is to remove those includes by default, and have users be explicit if they want to add anything outside their module root. 

Additionally with https://github.com/dagger/dagger/issues/6627, more stuff should be excluded by default, even within your module like a heavy `node_modules`, `.yarn_cache`, or `vendor`. I commented on the other issue that you reference (in the title) how this really affects performance. See https://github.com/dagger/dagger/issues/8499#issuecomment-2364170118. 


We could do both of these soon, now that https://github.com/dagger/dagger/pull/8818 has been merged. It's just a matter of bandwidth and priorities right now.

This way the module's context directory will continue to be limited at the git repo root, which we make assumptions about all over the code base, without affecting performance since if you don't add any files outside the module's root (in `dagger.json`), it'll just find the path of the git root, not load any files from it.

idlsoft:
> Should open an issue proposing this change before opening a PR, but I don't think this is the right approach.

This is definitely a stopgap solution. I was originally thinking about adding settings to `dagger.json` but this was just easier.


> This way the module's context directory will continue to be limited at the git repo root, which we make assumptions about all over the code base, without affecting performance since if you don't add any files outside the module's root (in `dagger.json`), it'll just find the path of the git root, not load any files from it.

Right, I think most of the time is being spent in trying to locate all the '**/...' patterns.
As long as this doesn't happen implicitly the performance should not suffer.

In my example the module had one simple root, but ideally it'd be nice to be able to specify a list of directories to include in the module context, so you could have the main module + a few libraries with common build logic.
In a monorepo these libraries would typically reside somewhere near the top of the git repo, and modules using them closer to the actual code that needs to be built.

helderco:
> ideally it'd be nice to be able to specify a list of directories to include in the module context, so you could have the main module + a few libraries with common build logic.

You can already do this with the `include` field in `dagger.json`. Use a relative path. It's just that go modules implicitly include all go files in the repo, that's what needs to change. Other SDKs don't suffer from this. Then, users relying on this implicitness will have to explicitly add the right patterns to their `dagger.json` files.

Issue #8956, opened 2024-11-14
🐞 `go test` won't build on macOS


cwlbraa:
### What is the issue?

`dagger call test` obviously works, but this can be annoying when you need to be playing around in engine guts. `go test` should build and run on macOS so you can run against a built dev engine.

### Dagger version

main

### Steps to reproduce

`go test ./dagql/idtui`, or any other package with tests, really. 

### Log output

```
# github.com/dagger/dagger/engine/sources/gitdns
engine/sources/gitdns/source_unix.go:41:20: undefined: syscall.Unshare
engine/sources/gitdns/source_unix.go:41:36: undefined: syscall.CLONE_FS
engine/sources/gitdns/source_unix.go:41:55: undefined: syscall.CLONE_NEWNS
engine/sources/gitdns/source_unix.go:54:3: unknown field Pdeathsig in struct literal of type "syscall".SysProcAttr
engine/sources/gitdns/source_unix.go:81:19: undefined: mount.Mount
engine/sources/gitdns/source_unix.go:87:21: undefined: syscall.Mount
engine/sources/gitdns/source_unix.go:87:75: undefined: syscall.MS_BIND
# github.com/dagger/dagger/engine/buildkit [github.com/dagger/dagger/dagql/idtui.test]
engine/buildkit/linux_namespace.go:108:21: undefined: unix.CLONE_NEWNET
engine/buildkit/linux_namespace.go:155:18: undefined: unix.Setns
engine/buildkit/linux_namespace.go:170:18: undefined: unix.Setns
FAIL	github.com/dagger/dagger/dagql/idtui [build failed]
FAIL
```

marcosnils:
It's a known issue as there's some internal packages that are Linux only and can't be ported to darwin

cwlbraa:
@marcosnils yeah, and that's fine, see #8957 for how those packages should be handled– they can still have "unimplemented" builds

PR #8957, opened 2024-11-14
fix(test): make go test work on macOS

cwlbraa:
see https://github.com/dagger/dagger/issues/8956 for error logs. this PR makes it so we can build our test packages (at least tui and core/integration) on macOS, by adding panic(unimplemented) impls hidden behind conditional build directives.

### DIFF
--- a/engine/buildkit/linux_namespace.go
+++ b/engine/buildkit/linux_namespace.go
@@ -1,3 +1,5 @@
+//go:build !darwin && !windows
+
 package buildkit
 
 import (
--- a/engine/buildkit/unimplemented_namespace.go
+++ b/engine/buildkit/unimplemented_namespace.go
@@ -1,0 +1,53 @@
+//go:build darwin || windows
+
+package buildkit
+
+import (
+	"context"
+	"os"
+	"time"
+)
+
+const (
+	idleTimeout    = 1 * time.Second
+	workerPoolSize = 5
+)
+
+func runInNetNS[T any](
+	ctx context.Context,
+	state *execState,
+	fn func() (T, error),
+) (T, error) {
+	panic("implemented only on linux")
+}
+
+func (w *Worker) runNetNSWorkers(ctx context.Context, state *execState) error {
+	panic("implemented only on linux")
+}
+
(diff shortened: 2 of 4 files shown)


jedevc (approved):
:tada: nice!

Do you think it would make sense to add this as a check to CI? I think we've fixed some variant of this problem 2 or 3 times before, so might be nice to avoid accidentally breaking it again on restructures.

cwlbraa:
@jedevc it'd be nice, for sure, though I'm uncertain about the best way to structure that and whether we'd need an actual macOS build runner... x-compilation of tests might be enough here (as opposed to actually running the tests)

cwlbraa:
@jedevc for example:

```
go test -exec /bin/true ./...
package github.com/dagger/dagger/cmd/engine
	imports github.com/dagger/dagger/engine/server
	imports github.com/containerd/containerd/snapshots/overlay: build constraints exclude all Go files in /Users/braa/go/pkg/mod/github.com/containerd/containerd@v1.7.23/snapshots/overlay
package github.com/dagger/dagger/cmd/engine
	imports github.com/dagger/dagger/engine/server
	imports github.com/containerd/containerd/snapshots/overlay/overlayutils: build constraints exclude all Go files in /Users/braa/go/pkg/mod/github.com/containerd/containerd@v1.7.23/snapshots/overlay/overlayutils
package github.com/dagger/dagger/cmd/engine
	imports github.com/dagger/dagger/engine/server
	imports github.com/containerd/fuse-overlayfs-snapshotter: build constraints exclude all Go files in /Users/braa/go/pkg/mod/github.com/containerd/fuse-overlayfs-snapshotter@v1.0.8
```

I can try to make this pass too, ostensibly, with a panic(unimplemented) engine, but there's a slippery slope where big packages will need to have copies of a lot of shit or i'll need to break files down into parts w/ used-outside symbols that build x-platform vs implementations that import platform-specific dependencies

sipsma (approved):

jedevc:
x-compilation of tests would be fine, yup. but yes, totally not a blocker :tada:

PR #8971, opened 2024-11-17
dagql: support elective purity

vito:
Currently impurity is static at the schema level: if a field is marked impure, it will never be cached, and IDs coming out of it will be
tainted.

This quickly improved insufficient: we want some APIs to determine whether they're pure at runtime, based on their inputs.

For example, we want this call:

    cacheVolume(key: "foo")

to return a namespaced volume, expanding implicitly to:

    cacheVolume(key: "foo", namespace: "modID").

With this change, we can just have `cacheVolume` call itself with `dagql.Select` by providing a namespace argument and `Pure: true`.

### DIFF
--- a/dagql/cachemap.go
+++ b/dagql/cachemap.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	"github.com/opencontainers/go-digest"
 )
@@ -81,11 +82,13 @@ func (m *cacheMap[K, T]) GetOrInitializeOnHit(ctx context.Context, key K, fn fun
 	m.calls[key] = c
 	m.l.Unlock()
 
+	purity := purityTrackerFromContext(ctx)
+
 	ctx = context.WithValue(ctx, cacheMapContextKey[K, T]{key: key, m: m}, struct{}{})
 	c.val, c.err = fn(ctx)
 	c.wg.Done()
 
-	if c.err != nil {
+	if c.err != nil || (purity != nil && purity.IsTainted()) {
 		m.l.Lock()
 		delete(m.calls, key)
 		m.l.Unlock()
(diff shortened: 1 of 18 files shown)


dagql/cachemap.go
```diff
@@ -81,11 +82,13 @@ func (m *cacheMap[K, T]) GetOrInitializeOnHit(ctx context.Context, key K, fn fun
...
 	m.l.Unlock()
 
+	purity := purityTrackerFromContext(ctx)
+
 	ctx = context.WithValue(ctx, cacheMapContextKey[K, T]{key: key, m: m}, struct{}{})
 	c.val, c.err = fn(ctx)
 	c.wg.Done()
 
-	if c.err != nil {
+	if c.err != nil || (purity != nil && purity.IsTainted()) {
```
sipsma, line 91: I think there might be a subtle but meaningful problem here (double check my thinking).

For APIs that are marked impure in the schema, they skip the cache map entirely. That means that two clients making the same call to that impure API ***at the same time*** with the same args will execute separately.
* That's what we want for e.g. the cacheVolume use case since different clients making the same call are gonna get different results

However, in this codepath, those two clients making the same call at the same time will have the execution of the impure API deduped between the two. Even if we skip storing the value in the cache, one of those two clients will unexpectedly get the result from the other's execution.

---

If that's correct, I wonder if we should go with a slightly different approach: rather than allowing an API call to mark itself as impure in the middle of execution, we can just allow calls to `dagql.Select` override the default impurity of a call.

E.g. for the `cacheEntrySet` case updated here, we can leave the API marked as Impure by default, but when we make the call to `dagql.Select` (line 74 atm) we override the impurity setting to be pure.
* Could override via context or an explicit arg to `Select`, etc.

AFAICT that should work for the rest of the use cases we have in mind here since they all need to make `dagql.Select` calls to create new pure IDs.

vito: Yeah good catch - definitely breezed by that deduping behavior too quickly. We could work around it, but I think I like your proposal more, since we keep the schema level hint that the API _may_ be impure.

I think we can do something pretty slick: just add a `Pure bool` field to `dagql.Selector` and have its truthiness override the spec's purity in `(dagql.Selector).AppendTo`. Don't need a context arg or even an additional arg, and you can specify which exact field you're making pure. Will push a commit to this effect soon!

vito:
@sipsma Ready for another look!

sipsma (approved):

PR #8982, opened 2024-11-18
ci(gha): remove job outputs

jedevc:
As raised by @rajatjindal [in discord](https://discord.com/channels/707636530424053791/1003718839739105300/1306867294743302168):

> The java-dev and javascript-dev actions are failing on GitHub actions with error: 
>
> ```
> Error: Fail to evaluate job outputs
> Error: The template is not valid. System.InvalidOperationException: Maximum object size exceeded
> ```

An example failure: https://github.com/dagger/dagger/actions/runs/11887650112/job/33120760976?pr=8973

---

The gha module was generating outputs for some reason - these outputs contained *all* of stdout and stderr, which could sometimes exceed the total allowed object size in github actions:

    Error: The template is not valid. System.InvalidOperationException: Maximum object size exceeded

However, it didn't seem like we were ever using these outputs - we still need to capture them for the process of producing the summary report, but we don't actually seem to need the job outputs - so I've just gone and removed them, which should hopefully fix the issue.

I don't think we need them @shykes? I suspect they were leftover from an earlier refactor, but hopefully you can confirm?

### DIFF
--- a/.github/workflows/daggerverse-preview.gen.yml
+++ b/.github/workflows/daggerverse-preview.gen.yml
@@ -77,7 +77,6 @@ jobs:
                     export PATH=$(dirname "$_EXPERIMENTAL_DAGGER_CLI_BIN"):$PATH
                 fi
 
-                GITHUB_OUTPUT="${GITHUB_OUTPUT:=github-output.txt}"
                 GITHUB_STEP_SUMMARY="${GITHUB_STEP_SUMMARY:=github-summary.md}"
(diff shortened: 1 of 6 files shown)


gerhard (approved):
SGTM!

@TomChv was mentioning this issue earlier today, this seems like a good quick fix to me.

jedevc:
Merging this now to unblock some broken CI - if it's discovered later that we actually needed these outputs (for some reason that escapes me), then we can evaluate options from there.

PR #8986, opened 2024-11-18
fix(core): drop leading slash from paths for gitignore

jedevc:
Follow-up to https://github.com/dagger/dagger/pull/8221.

We've got some very similar code below, we should probably handle it for `gitignore` as well :thinking:

### DIFF
--- a/core/schema/module.go
+++ b/core/schema/module.go
@@ -1175,6 +1175,7 @@ func (s *moduleSchema) updateCodegenAndRuntime(
 			if bytes.Contains(gitIgnoreContents, []byte(fileName)) {
 				continue
 			}
+			fileName := strings.TrimPrefix(fileName, "/")
 			gitIgnoreContents = append(gitIgnoreContents,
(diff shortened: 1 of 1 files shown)


cwlbraa (approved):

core/schema/module.go
```diff
@@ -1175,6 +1175,7 @@ func (s *moduleSchema) updateCodegenAndRuntime(
 			if bytes.Contains(gitIgnoreContents, []byte(fileName)) {
 				continue
 			}
+			fileName := strings.TrimPrefix(fileName, "/")
```
helderco, line 1178: I'm not sure about this. If you have `/sdk` it may mean you want to exclude the `sdk` dir next to `.gitignore`, but not `client/sdk`.

jedevc: See the line after - we always unconditionally prefix the `/`. This PR just ensures we don't end up with double `//` if the SDK already included one.

PR #8991, opened 2024-11-18
OTel: bump to 1.32.0/1.26.0

vito:
This is the first time bumping since adding the replace pinning to module go.mods, and it took a bit of a rework to deal with go.work.

In particular, any modules that rely on go.work need to additionally have a replace rule installed to go.work, otherwise you'll get an error as soon as you bump one module's individual replace rules.

So now the go.mod and go.work replacing happens by running `go mod edit` and `go work edit` in `PostCommands`, which also avoids the issue of `modfile.ParseLax` not parsing replace rules.

Also, now we `go mod tidy` before `go work use .` which is important for bumping the Go version that ends up in go.work.


cmd/codegen/generator/go/generator.go
```diff
@@ -445,3 +474,15 @@ func (m *%[1]s) GrepDir(ctx context.Context, directoryArg *dagger.Directory, pat
 }
 `, moduleStructName, pkgInfo.PackageImport)
 }
+
+func goEnv(dir string, env string) (string, error) {
+	buf := new(bytes.Buffer)
+	findGoWork := exec.Command("go", "env", env)
```
jedevc, line 480: nit: this can be generic, e.g. `findGoEnv`.

jedevc (approved):
LGTM :tada: shame about the weird `ParseLax` behavior, but oh well :smile:

vito:
note: holding off on this, since nothing requires it, and it currently breaks the Bass SDK 😭 

```
Error logs:

✘ .withExec(args: ["go", "build", "-o", "/bass", "./entrypoint"]): Container! 0.6s
# dagger.io/dagger/telemetry                                                                                                                                                                                                                                                                 
/go/pkg/mod/dagger.io/dagger@v0.12.2/telemetry/exporters.go:92:23: cannot use log (variable of type "go.opentelemetry.io/otel/sdk/log".Record) as *"go.opentelemetry.io/otel/sdk/log".Record value in argument to e.OnEmit                                                                   
! process "go build -o /bass ./entrypoint" did not complete successfully: exit code: 1
```

I think the fix is trivial - the Bass SDK should just be using the internal codegen'd `telemetry` package instead - but yeah, this would still break modules in the ecosystem.

Safest bet might be to just fix the Bass SDK and give time for peoples' dependency graph to roll forward to it.

jedevc:
Maybe a specific reason to do this - we can't bump grpc until we add this. 

https://github.com/grpc/grpc-go/blob/e8055ea11f96238b0fb14b1062abb598c33fc9c3/go.mod#L13-L18

jedevc:
Rebased + resolved conflicts. @vito is there a blocker on merging this now?

jedevc:
Bleh, looks like we'll need https://github.com/shykes/daggerverse/pull/12, since we now rely on that wolfi module in our tests.

vito:
@jedevc Surprising - any chance we could switch it to the in-repo Wolfi module?

PR #8999, opened 2024-11-19
ci(alt nightly): Disable our heaviest pipelines on alternative runners & run twice daily

gerhard:
They have been failing days in a row since they require at least 2x the CPU resources that are made available by default. We could bump the timeout, but that will only result in higher infra costs and not provide much value - we already know about the limitations & current behaviour.

We are also only keeping pipelines for a subset of SDKs. The goal is to have decent coverage, not be exhaustive.

**Also** run the workflow twice so that we test the caching efficiency.

### DIFF
--- a/.github/workflows/alternative-ci-runners.yml
+++ b/.github/workflows/alternative-ci-runners.yml
@@ -12,21 +12,21 @@ permissions:
   contents: read
 
 jobs:
-  engine-lint-on-depot:
-    uses: ./.github/workflows/_dagger_on_depot_pre-provisioned_with_cache.yml
-    with:
-      function: engine lint
-  engine-test-on-depot:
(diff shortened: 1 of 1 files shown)


aluzzardi (approved):

cwlbraa:
> Also run the workflow twice so that we test the caching efficiency.

if we're going to nightlies, running the tests multiple times with cache invalidation could be used to programmatically detect flakes

gerhard:
Most of our cache gets invalidated when there is **any** change, anywhere in the codebase. We also run these tests **a lot** - `~3000 hours total` in October (that's right, running about 100 hours per day). I think that we have plenty of data already to detect flakes programatically, it's just a matter of wiring it all together.

PR #9011, opened 2024-11-20
chore(tests): archive test results in json format with CI runs

rajatjindal:
this is just a poc to try out gotestsum with CI. The benefits includes a little clearer output (more work needs to be done there) and that we can archive results in a json file, which we can then ingest in a tests dashboard to look at the trends over the executions.

happy to discuss if folks have opinions/concerns regarding this. I am just evaluating how this can improve debugging of CI and finding flakes.

### DIFF
--- a/.dagger/test.go
+++ b/.dagger/test.go
@@ -2,10 +2,13 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"path/filepath"
 	"strings"
+	"time"
 
+	"github.com/google/uuid"
 	"github.com/moby/buildkit/identity"
 
 	"github.com/dagger/dagger/.dagger/internal/dagger"
@@ -190,7 +193,8 @@ func (t *Test) test(
 	if err != nil {
 		return err
 	}
-	_, err = t.goTest(
+
+	ctr, err := t.goTest(
 		cmd,
 		runTestRegex,
 		skipTestRegex,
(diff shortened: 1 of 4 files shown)


rajatjindal:
I want to open it up for discussion. The main objective of this PR is to add test results as artifact to the github actions runs.

gerhard (commented):
This is almost there!

I can definitely see us approving & merging this after this is rebased against `main` & all conflicts are solved.

.dagger/test.go
```diff
@@ -350,7 +353,9 @@ func (t *Test) testCmd(ctx context.Context) (*dagger.Container, error) {
 		WithMountedFile(cliBinPath, devBinary).
 		WithEnvVariable("_EXPERIMENTAL_DAGGER_CLI_BIN", cliBinPath).
 		WithEnvVariable("_EXPERIMENTAL_DAGGER_RUNNER_HOST", endpoint).
-		With(t.Dagger.withDockerCfg) // this avoids rate limiting in our ci tests
+		With(t.Dagger.withDockerCfg). // this avoids rate limiting in our ci tests
+		WithExec([]string{"sh", "-c", "go install gotest.tools/gotestsum@v1.12.0"})
```
jedevc, line 357: We should avoid installs at this point - this will require a fresh download every single time we change any of the things in the pipeline above.

jedevc (commented):
@vito removed gotestsum in https://github.com/dagger/dagger/pull/7532#issue-2329943663.

I can't fully remember why, but probably worth checking first.

vito:
> I can't fully remember why, but probably worth checking first.

No big reason, just seemed redundant with the new `testctx` integration. I suspect the rest of this PR will become unnecessary once we implement analytics in Cloud, but we're not there yet, so seems fine to bring it back

.dagger/test.go
```diff
@@ -204,6 +205,70 @@ func (t *Test) test(
...
 		testVerbose,
 	).Sync(ctx)
+	if err != nil {
+		return err
+	}
+
+	return t.publishTestResults(ctx, ctr)
+}
+
+func (t *Test) publishTestResults(ctx context.Context, ctr *dagger.Container) error {
```
jedevc, line 215: Can we split this publishing up to the dashboard site into a separate PR?

I'm a little worried about exporting our test results to an external service that can go down that we don't have full control over. I think we need more discussion on this. If we *do* do this, I don't think it should be part of every `test` call, which is also used locally - it would need to have logic to only apply in CI.

The rest of the PR around `gotestsum`/etc looks good, and I don't want to block those changes from getting in while we're discussing that.

.dagger/test.go
```diff
@@ -279,7 +346,7 @@ func (t *Test) goTest(
 
 	return cmd.
 		WithEnvVariable("CGO_ENABLED", cgoEnabledEnv).
-		WithExec(args)
+		WithExec(args, dagger.ContainerWithExecOpts{Expect: dagger.ReturnTypeAny})
```
jedevc, line 349: Now all `test` commands will always pass? :thinking: 

We should still fail the job, probably by checking the return code.

modules/go/main.go
```diff
@@ -81,10 +81,12 @@ func New(
 				// The specific version is dictated by Dagger's own requirement
 				// FIXME: make this optional with overlay support
 				"protoc~3.21.12",
+				"curl",
 			}}).
 			WithEnvVariable("GOLANG_VERSION", version).
 			WithEnvVariable("GOPATH", "/go").
 			WithEnvVariable("PATH", "${GOPATH}/bin:${PATH}", dagger.ContainerWithEnvVariableOpts{Expand: true}).
+			WithExec([]string{"sh", "-c", "go install gotest.tools/gotestsum@v1.12.0"}).
```
jedevc, line 89: nit: 

```suggestion
			WithExec([]string{"go", "install", "gotest.tools/gotestsum@v1.12.0"}).
```

We can also move this below the `WithMountedCache` calls, so that this can be part of populating the cache if there are shared dependencies.

jedevc (changes requested):

.github/workflows/engine-and-cli.yml
```diff
@@ -35,6 +35,32 @@ jobs:
         with:
           function: "scripts lint"
 
+  test:
```
jedevc, line 38: These jobs have now been removed as part of #8998.

.dagger/test.go
rajatjindal: apologies for the noise. But this endpoint is temporary here. Me and @gerhard had discussion regarding this and we are going to deploy this service on dagger infrastructure. This is me just testing out stuff and making sure other things are aligned and work.

rajatjindal: I just changed this PR back to being a draft PR until I am done testing other things.

rajatjindal: Hi @jedevc

The URL has been updated to post the test results to dashboard deployed on dagger infra. I hope we can mark this comment as resolved?

rajatjindal:
to make sure we don't miss it among other things: there is a failure happening which seems likely due to these changes:

```
multi.go:85: 88 : Host.directory(path: "/app"): Directory!
multi.go:85: 92 : │ filesync
multi.go:85: 92 : │ filesync ERROR [0.2s]
multi.go:85: 92 : │ ! failed to sync: conflict at "testresults.json": size changed from "8388059" to "8428160" during sync
multi.go:85: 89 : │ upload /app from x0hfej85gfigcnz5mt6nahap7 (client id: if31f540x18253ycvfnwd8mwp, session id: szgx9vf13z1kk2eid885olwqo) ERROR [0.2s]
multi.go:85: 89 : │ ! failed to get snapshot: failed to sync: conflict at "testresults.json": size changed from "8388059" to "8428160" during sync
multi.go:85: 88 : Host.directory ERROR [0.2s]
multi.go:85: 88 : ! host directory /app: failed to get snapshot: failed to sync: conflict at "testresults.json": size changed from "8388059" to "8428160" during sync
```

I am trying to figure out why this is happening.

.dagger/test.go
rajatjindal: Added code to handle the error code. (it was always the plan, but had not implemented it yet. nonetheless, thank you for your detailed code review as always).

rajatjindal:
> to make sure we don't miss it among other things: there is a failure happening which seems likely due to these changes:
> 
> ```
> multi.go:85: 88 : Host.directory(path: "/app"): Directory!
> multi.go:85: 92 : │ filesync
> multi.go:85: 92 : │ filesync ERROR [0.2s]
> multi.go:85: 92 : │ ! failed to sync: conflict at "testresults.json": size changed from "8388059" to "8428160" during sync
> multi.go:85: 89 : │ upload /app from x0hfej85gfigcnz5mt6nahap7 (client id: if31f540x18253ycvfnwd8mwp, session id: szgx9vf13z1kk2eid885olwqo) ERROR [0.2s]
> multi.go:85: 89 : │ ! failed to get snapshot: failed to sync: conflict at "testresults.json": size changed from "8388059" to "8428160" during sync
> multi.go:85: 88 : Host.directory ERROR [0.2s]
> multi.go:85: 88 : ! host directory /app: failed to get snapshot: failed to sync: conflict at "testresults.json": size changed from "8388059" to "8428160" during sync
> ```
> 
> I am trying to figure out why this is happening.

this is now fixed. 

@gerhard @jedevc I believe this PR is ready for review/merge. 

Thanks

Issue #8554, opened 2024-09-23
TestFlake: `TestDaggerTerminal` hangs


sipsma:
I've seen this one a handful of times in CI now, most recent occurrence a few commits ago on main: 
* https://dagger.cloud/dagger/traces/aa0c05d27e1f539f743f501368d4126d#cd9e27691b1db165

The test case appears to be going as expected, the interactive container is opened after the failed exec, runs some commands, successfully sends `exit` and then the test hangs until the timeout on the whole suite of 30m is hit.

The next test case lines don't show up in progress output, so I'd *guess* it's hanging somewhere around this line: https://github.com/sipsma/dagger/blob/5e3299c08d62c3fc91d1ce02b79466ea20c02562/core/integration/module_terminal_test.go#L545-L545

No immediate strong clues, first hunch is to look for any race conditions in the debug container code that could result in a hang after an exit. Given the load the CI runners are under wouldn't be surprising if a rare race condition only gets triggered there.
* Will start by re-opening my deflaking PR and keep re-running there til I hit it and then shell into the CI runner and getting a stack trace from the engine to see what line, if any, it's stuck on.

sipsma:
This one is still appearing ([GHA](https://github.com/dagger/dagger/actions/runs/11259583691/job/31308896759)) ([Traces](https://dagger.cloud/dagger/traces/a800de1a9b253530bdc80a5773484f26))

sipsma:
Another recent occurrence https://dagger.cloud/dagger/traces/15c22afdc0ae043aed5e80b9282135f8

It seems like the `on failure` test in particular is prone to this (might be the only one?) cc @aluzzardi

sipsma:
While working on the filesync stuff I started hitting this flake locally relatively consistently by running:
```
./hack/with-dev go test -v -run='TestModule/TestDaggerTerminal|TestModule/TestDaggerUp' -count=10 -parallel=8 ./core/integration/
```

can't repro locally unless it's the specific combination of tests (most likely due to subtle timing differences).

So took a tangent and figured out what's happening; there's a few levels of issues but the main one is that we aren't closing some stdout/stderr pipes in error cases involving terminal containers, which leads to [this line](https://github.com/sipsma/dagger/blob/e1347f2d7e0d350b6a6c692aab916de09adf4911/engine/buildkit/executor.go#L773-L773) sometimes blocking forever if it happens to be in the middle of a write to stdout (will unblock only if it happened to be in the middle of reading from the pty).

[...]

PR #9013, opened 2024-11-20
engine: fix goroutine leaks + deadlock in terminal

sipsma:
Fixes https://github.com/dagger/dagger/issues/8554

While running some tests locally for the (unrelated) filesync PR I accidentally stumbled on a set of tests to run that coincidentally repro'd the flake in #8554 pretty consistently locally:
```
./hack/with-dev go test -v -run='TestModule/TestDaggerTerminal|TestModule/TestDaggerUp' -count=10 -parallel=8 ./core/integration/
```

I would hit the flake where the `on_failure` test hung after 3-4 counts pretty consistently. The set of tests most likely just happens to tweak timing on my machine in just the right way that the race conditions it was caused by would get triggered more often.

There were multiple problems from the test case down through the engine, described in the commit below.

With this fix in place I can't repro the flake locally anymore. Additionally, after running the terminal tests and sending SIGQUIT to the engine I previously saw a ton of leaked goroutines; now after this change they are all gone.

I also included a small one-line performance fix that I noticed while going through the code here.

---

[engine: fix goroutine leaks + deadlock in terminal](https://github.com/dagger/dagger/commit/2d49491b70f58ff67b8924e2a1efd35483034f1e)

Before this, every interactive debug terminal container would leak
multiple goroutines due to various io pipes not getting closed. In one
particular codepath, the read-side of the stdout io.Pipe would be left
open while the other side was in the middle of a write, which caused the
executor code to block forever and never return. This manifested as an
occasional flake in the TestModule/TestDaggerTerminal/on_failure test
case.

Additionally, the part of that test case which meant to assert on
handling of --interactive-command values which didn't exist was not
checking error values. It was hitting an error due to the fact that it
didn't have a TTY setup rather than the error around the command not
existing. That test case is fixed now to setup a TTY and verify the
error message is the expected one.

---

[engine: slightly optimize terminal reading by reusing buffers](https://github.com/dagger/dagger/commit/6f149d02eb694bb365c025b434ea7dcaa916e502)

Rather than allocating a new byte slice for every read from
stdout/stderr sent to clients, we can re-use the same byte slice.

This reduces the amount of allocations the engine does. Not a
world-changing optimization but I have seen similar fixes in other
contexts make a noticeable difference in CPU usage since the Go runtime
has a lot less work to do.

### DIFF
--- a/.changes/unreleased/Fixed-20241120-160241.yaml
+++ b/.changes/unreleased/Fixed-20241120-160241.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: |
+  Fix resource leaks in the engine that occurred after each debug terminal was opened.
+time: 2024-11-20T16:02:41.381210606-08:00
+custom:
+  Author: sipsma
+  PR: "9013"
--- a/core/integration/module_terminal_test.go
+++ b/core/integration/module_terminal_test.go
@@ -597,13 +597,25 @@ type Test struct {
 		err = cmd.Wait()
 		require.Error(t, err)
 
-		//  We try again with an invalid shell to confirm we replaced the default command
+		// We try again with an invalid shell to confirm we replaced the default command
+		// We have to set a TTY though or else the error will just be that the --interactive flag doesn't work without a terminal
+		console, err = newTUIConsole(t, 60*time.Second)
+		require.NoError(t, err)
+		defer console.Close()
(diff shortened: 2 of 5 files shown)


jedevc (approved):

gerhard (approved):
This must have been a pain to track down & fix. LGTM!

PR #8695, opened 2024-10-12
docs: Add page for caching configuration

vikram-dagger:

PR #9022, opened 2024-11-21
docs: rework custom configuration docs with `engine.json`

jedevc:
With https://github.com/dagger/dagger/pull/8800 about to merge, we can substantially improve our engine configuration docs.

*All* engine configuration for `engine.json` and `engine.toml` are moved into `engine.mdx`. This is no longer split up across multiple pages which are harder to navigate. As part of this:
- `custom-registry.mdx` is removed and merged into this now.
- The `custom-runner.mdx` page is reworked to only include information about starting an engine.
- `custom-ca.mdx` and `proxy.mdx` are linked to from the bottom of `engine.mdx` - their configs cannot *yet* be configured via `engine.json`/`engine.toml`, but we link them here for improved navigation.


docs/current_docs/configuration/cache.mdx
```diff
@@ -60,9 +60,7 @@ The default cache policy attempts to keep the long-term cache storage under 75%
 of the total disk, while also ensuring that at least 20% of the disk remains
 free for other applications and tools.
 
-The cache policy can be manually configured by editing the `engine.toml` file in the engine
-(which uses [BuildKit's syntax config](https://github.com/moby/buildkit/blob/v0.17.0/docs/buildkitd.toml.md?plain=1#L66-L82)).
-See the [Custom Runner](./custom-runner.mdx) docs for more information.
+The cache policy can be manually configured using the [engine config](./engine.mdx).
```
jedevc, line 63: This page (introduced in #8695) is still useful, but the actual configuration information is now all in `engine.mdx`. I think it makes sense to move this page somewhere else, but not entirely sure where, any suggestions @vikram-dagger?

docs/current_docs/configuration/engine.mdx
```diff
@@ -0,0 +1,425 @@
...
+
+</TabItem>
+
+</Tabs>
+
+## Options
+
+### Logging
+
+The dagger engine supports various log verbosity levels - the default level is to enable `debug`.
```
vikram-dagger, line 100:
```suggestion
Dagger supports various log verbosity levels. The default level is to enable `debug`.
```

docs/current_docs/configuration/engine.mdx
```diff
@@ -0,0 +1,425 @@
...
+
+</Tabs>
+
+### Security
+
+The engine by default has an open security policy. This policy allows exec
+operations to access the `insecureRootCapabilities` options, which effectively
+allows all operations root access (similar to [`docker run`'s --privileged` flag](https://docs.docker.com/reference/cli/docker/container/run/#privileged).
+
+This capability can be disabled to lock down a dagger installation.
```
vikram-dagger, line 161:
```suggestion
This capability can be disabled to lock down a Dagger installation.
```

docs/current_docs/configuration/engine.mdx
```diff
@@ -0,0 +1,425 @@
...
+The specified `hello-world` container will now be pulled from the mirror
+instead of from Docker Hub.
+
+</TabItem>
+
+</Tabs>
+
+### Custom proxy
+
+Currently, custom configuration authorities cannot be configured through
```
vikram-dagger, line 418:
```suggestion
Currently, custom proxies cannot be configured through
```

docs/current_docs/configuration/engine.mdx
```diff
@@ -0,0 +1,425 @@
...
+`engine.json` or `engine.toml`, and must be configured [separately](./proxy.mdx).
+
+### Custom Certificate Authorities
```
vikram-dagger, line 421:
```suggestion
### Custom certificate authorities
```

docs/current_docs/configuration/engine.mdx
```diff
@@ -0,0 +1,425 @@
...
+
+Currently, custom configuration authorities cannot be configured through
```
vikram-dagger, line 423:
```suggestion
Currently, custom certificate authorities cannot be configured through
```

docs/netlify.toml
```diff
@@ -1197,6 +1197,12 @@
   to = "/quickstart/428201/custom-function"
   status = 301
 
+[[redirects]]
+  # redirect 
+  from = "/configuration/custom-registry"
+  to = "/configuration/engine#configuring-custom-registries"
+  status = 301
+
```
vikram-dagger, line 1205: If deleting the caching page as suggested, we should add a redirect for that as well.

vikram-dagger (changes requested):

docs/current_docs/configuration/custom-ca.mdx
```diff
@@ -2,29 +2,59 @@
...
+<CodeBlock language="shell">
+{`docker run --rm \\\n`}
+{`    -v /var/lib/dagger \\\n`}
+{`    -v $PWD/ca-certificates:/usr/local/share/ca-certificates/ \\\n`}
+{`    --name dagger-engine-custom \\\n`}
+{`    --privileged \\\n`}
+{`    registry.dagger.io/engine:${daggerVersion}`}
+</CodeBlock>
 
 ## Configuration Applied to User Containers
```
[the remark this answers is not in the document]
jedevc: Sounds good! I just reformatted these files a bit, but we should definitely pick these up while in the area :tada:

jedevc:
@vikram-dagger if it's easier, feel free to just directly make changes and push them to my branch as a separate commit!

docs/current_docs/configuration/cache.mdx
jedevc: Hm, I'm not 100% sure about this.

I think having a separate caching page where we can talk about "how caching works" is important. Long term, I know we want to talk more about it (and it's probably going to get more complex), and we could even add information on how to optimize your pipelines for getting better cache results, with examples in different languages.

The `dagger query` examples to inspect cache and prune it are really useful, but I don't think they're related to configuration, and I wouldn't expect to find them under there.

WDYT about merging into https://docs.dagger.io/features/caching? Although, that section doesn't really have any code examples, so not sure about that. But it does feel like there's a lot of duplication between those.

vikram-dagger: I would not put it in the features page, mainly because these seem like "operational" commands and the features pages are more "big picture"/"concepts" (side note: I'm not very confident that "caching best practices" in the features caching page is correct either, at some point this will move out).

WDYT of adding an entry to the FAQ page with these commands? e.g. `How do I inspect the Dagger cache?`

vikram-dagger:
> @vikram-dagger if it's easier, feel free to just directly make changes and push them to my branch as a separate commit!

Thanks, I might do that as it will be easier for minor copy/style fixes.

docs/current_docs/configuration/cache.mdx
vikram-dagger: JFYI I also thought of including these commands in https://docs.dagger.io/api/cache-volumes, but I think this is probably wrong as that page is specific to cache volumes only, not the layer cache. However let me know if you have a different opinion @jedevc

jedevc: > WDYT of adding an entry to the FAQ page with these commands? e.g. `How do I inspect the Dagger cache?` as a stopgap until we have a "proper" caching page?

I'm happy with that - we can add a redirect to there?

> JFYI I also thought of including these commands in [docs.dagger.io/api/cache-volumes](https://docs.dagger.io/api/cache-volumes), but I think this is probably wrong

Yeah, agreed - it's also that that page is relevant to the *usage* of cache volumes, while this is more about "operationally" using cache.

One *other* option would be to create a new API page on the core `engine` function, and the `Engine` type. Then, we could document the cache info there, as well as any other functionality we put under there. It's more specialized, and more geared towards operating a dagger engine.

vikram-dagger: > One other option would be to create a new API page on the core engine function, and the Engine type. Then, we could document the cache info there, as well as any other functionality we put under there. It's more specialized, and more geared towards operating a dagger engine

I like this idea more than the FAQ, because this is more scalable to future changes. I'll add an atomic commit with a skeleton of this page so we can iterate further (assuming that's OK, lmk if not).

jedevc: Yup, that sounds perfect, thanks!

vikram-dagger: Added page

docs/current_docs/configuration/engine.mdx
```diff
@@ -0,0 +1,394 @@
...
+import TabItem from "@theme/TabItem";
+import { daggerVersion } from "../partials/version";
+
+# Engine
+
+Dagger is designed to be run out-of-the-box with sensible defaults.
+However, even with these defaults, it may sometimes be necessary to modify various
+settings.
+
+TODO: update version when base PR is merged
```
vikram-dagger, line 16: @jedevc this looks ready to merge except for this TODO - do you want to remove it or is there something else required here?

jedevc: Yeah, we need to merge https://github.com/dagger/dagger/pull/8800 first.

vikram-dagger (approved):
Approving in advance so this can be merged at will, after the pending TODO item is fixed.

Issue #9023, opened 2024-11-21
Test suite: `dagger call check --targets=sdk/go` fails to run in CI from an upstream branch


gerhard:
<img width="1068" alt="image" src="https://github.com/user-attachments/assets/733976e8-331f-46b2-9e96-89b6e40d9bbb">

Dagger Cloud trace: https://dagger.cloud/dagger/traces/32a617781fc7ceb70228f10669351cc4

This is not a flake, it happens consistently:
1. https://github.com/dagger/dagger/actions/runs/11955188869/job/33326999788#step:3:17208
2. https://github.com/dagger/dagger/actions/runs/11955734409/job/33328800503#step:3:17539

It was working fine 3 hours ago: https://github.com/dagger/dagger/actions/runs/11953083867/job/33320214719

These are the only changes since the last time this passed:
1. https://github.com/dagger/dagger/commit/e6d47bd549f8a5b6d8c4ba332fdb1a42f72fc945 @helderco 
2. https://github.com/dagger/dagger/commit/c3ed7490b475dfbdb16c71e9ddae7697a5b7313b @helderco 
3. https://github.com/dagger/dagger/commit/ab4ae221c5dd80a16d342b757522c849042206ab @rajatjindal 

This is all the time that I have to investigate this particular issue, but still worth taking the first step towards this new habit:
- https://github.com/dagger/dagger/pull/8998#issuecomment-2491426455

---

cc @jedevc @sipsma in case you know what the issue is by skimming it.

jedevc:
Looks like we have this: https://dagger.cloud/dagger/traces/32a617781fc7ceb70228f10669351cc4?span=4a4eb949733c92c8#1b65c00adbc4fbe9

```
$ git checkout ci-alt-runners-common-terminology --
fatal: invalid reference: ci-alt-runners-common-terminology
```

jedevc:
Right, this is actually caused by an intentional regression in #8442 - this is likely because something is running a dev version of dagger off of main.

Discussing in discord here with @vito https://discord.com/channels/707636530424053791/1003718839739105300/1309198728455917620, to see if we can avoid a breakage, or at least minimize it.

jedevc:
Putting it into the milestone so that we at least don't forget about this! There's a few possible ways forward that don't break backwards-compat, but we would at least need to keep `Stdout`/`Stderr` on `ExecError` if we want to do that, and *maybe* try and keep the format of `Error` (but bleh, that's annoying that we depend on that).

Issue #8440, opened 2024-09-12
Catch missing embedded files at build?


shykes:
## Problem

I can access files "embedded" in my module source, with `dag.CurrentModule().Source()`, which is super practical. But if I make a mistake and the file is not there, I will not get an error when loading the module - only when calling the particular function that uses the file. I just got bit by this today.

## Solution

It would be great if missing embedded files were detected at module load. This would give me more confidence that I am not shipping my module with a hidden flaw that my users will find later.

## Design

There are several ways to do this, but they all involve a breaking change, I think.

For example, we could exclude all files not needed by the SDK, by default. Then require explicitly adding the files needed. Then the required files would be listed in `dagger.json`, and the module loader could check for their existence.

PR #9033, opened 2024-11-22
Preserve ExecError Stdout/Stderr, just don't show it

vito:
Follow-up to #8442 for better backwards compatibility.

We'll still break anyone relying on `.Error()` to contain Stdout/Stderr, but at least now they can check the Stdout/Stderr fields on the error, which is more explicit anyway.

Included some prior attempts at preserving even the `Error()` behavior, but I think it's much too brittle to rely on, so I reverted it and added the alternative (less backwards-compatible) approach on top.

edit: brought back some of that error-unwrapping code, since it was pretty nice that it removed those noisy `input: container.withExec.withExec.withExec...` messages.

### DIFF
--- a/sdk/elixir/lib/dagger/core/error.ex
+++ b/sdk/elixir/lib/dagger/core/error.ex
@@ -20,36 +20,6 @@ defmodule Dagger.Core.ExecError do
 
   @impl true
   def message(exception) do
-    stdout =
-      if String.trim(exception.stdout) != "" do
-        [
-          "Stdout:",
-          ?\n,
-          exception.stdout
-        ]
-      else
-        ""
-      end
-
-    stderr =
-      if String.trim(exception.stderr) != "" do
-        [
-          "Stderr:",
-          ?\n,
-          exception.stderr
-        ]
-      else
-        ""
-      end
-
-    [
-      Exception.message(exception.original_error),
-      ?\n,
-      stdout,
-      ?\n,
-      stderr,
-      ?\n
-    ]
-    |> IO.iodata_to_binary()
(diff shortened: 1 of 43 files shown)


jedevc (approved):
LGTM :tada: Shame about breaking `Error` in old code, but I had a quick browse through daggerverse - I didn't see a *huge* amount of usage like this. Hyrum's law is always gonna be a thing, but it shouldn't stop us making changes.

cc @helderco @TomChv @wingyplus on the relevant SDK changes - it's pretty much just the equivalent of the go ones.

I've also pushed a couple commits, one to ensure that our CI for https://github.com/dagger/dagger/issues/9023 continues to work with this change (confirmed locally), and another to add SDK-specific changelogs for this (but feels free to drop either of these if they should be separate PRs).

sdk/elixir/lib/dagger/core/error.ex
```diff
@@ -20,36 +20,6 @@ defmodule Dagger.Core.ExecError do
...
-    stderr =
-      if String.trim(exception.stderr) != "" do
-        [
-          "Stderr:",
-          ?\n,
-          exception.stderr
-        ]
-      else
-        ""
-      end
```
wingyplus, line 43: So we can remove `cmd`, `stdout`, `stderr` and `exit_code` out of the exception?

vito: I would keep them - that way users can still check for them if they're able to assert against the concrete error type. We went through the same for Go: initially disabled collecting that info in entirely #8440, but with this PR we brought it back, kept the fields, and just removed them from the display method (`Error() string`).

Issue #9010, opened 2024-11-20
`test-publish` timesout sometimes


sipsma:
Example:
* GHA: https://github.com/dagger/dagger/actions/runs/11915718428/job/33206646864
* Traces:  https://dagger.cloud/dagger/traces/f3b2b0dedf3575d00e1917764f17f911
  * Seems like it gets hung on `go build std` when building for the non-host platform (arm64 here since CI hosts are x86)

Best guess is that we are triggering the qemu emulator code paths, which are not super reliable; would not be surprised if it's choking on the compilation of the go std lib since that's probably extra complicated (there's assembly, cgo, etc.)

If so, best fix would be to refactor the CI to use cross-compilation rather than emulation (should be much faster anyways)

PR #9036, opened 2024-11-22
ci: cross-compile pre-built std lib

jedevc:
Fixes https://github.com/dagger/dagger/issues/9010.

@sipsma is there anything we can do to sanity test? It feels a bit fragile, not sure quite how to test this :thinking:

### DIFF
--- a/.dagger/build/sdk.go
+++ b/.dagger/build/sdk.go
@@ -3,6 +3,7 @@ package build
 import (
 	"context"
 	"encoding/json"
+	"runtime"
 
 	"github.com/dagger/dagger/engine/distconsts"
 	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
@@ -49,16 +50,7 @@ func (build *Builder) pythonSDKContent(ctx context.Context) (*sdkContent, error)
 		AsTarball(dagger.ContainerAsTarballOpts{
 			ForcedCompression: dagger.ImageLayerCompressionZstd,
 		})
-
-	sdkDir := dag.
-		Alpine(dagger.AlpineOpts{
-			Branch: consts.AlpineVersion,
-		}).
-		Container().
-		WithMountedDirectory("/out", dag.Directory()).
-		WithMountedFile("/sdk.tar", sdkCtrTarball).
-		WithExec([]string{"tar", "xf", "/sdk.tar", "-C", "/out"}).
-		Directory("/out")
+	sdkDir := unpackTar(sdkCtrTarball)
 
 	var index ocispecs.Index
 	indexContents, err := sdkDir.File("index.json").Contents(ctx)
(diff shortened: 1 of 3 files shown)


.dagger/build/sdk.go
```diff
@@ -128,18 +129,22 @@ func (build *Builder) typescriptSDKContent(ctx context.Context) (*sdkContent, er
...
 func (build *Builder) goSDKContent(ctx context.Context) (*sdkContent, error) {
-	base := dag.Container(dagger.ContainerOpts{Platform: build.platform}).
+	base := dag.Container().
 		From(consts.GolangImage).
-		WithExec([]string{"apk", "add", "git"})
+		With(build.goPlatformEnv).
+		WithEnvVariable("BUILDPLATFORM", "linux/"+runtime.GOARCH).
+		WithEnvVariable("TARGETPLATFORM", string(build.platform)).
+		WithExec([]string{"apk", "add", "git"}).
+		WithDirectory("/", dag.Container().From(consts.XxImage).Rootfs())
```
sipsma, line 138: This might be annoying or nonfeasible, but it's not ideal that we're including a ton of extra stuff in this go base image we include in the engine image. If in the practice the size difference before and after is negligible then doesn't matter. But if there is a meaningful size bump we could try to wipe this extra stuff out.

One possibility would be to run all this and then start with a clean `dag.Container().From(consts.GolangImage)` and copy in the stuff we added (mainly the mod download cache and the go build cache). You might also be able to do some fancy tricks with Diff+Merge (merge-op is triggered with copys that only involve `/` and no include/exclude), but that's a bit trickier since we hide merge-op as an implementation detail.

sipsma (commented):
> is there anything we can do to sanity test? It feels a bit fragile, not sure quite how to test this 🤔

We get a bit of coverage because the image we build is included in the engine and used whenever any module gets built. So I guess I could try to run this from an x86 machine and then use it on my arm64 host as an extra test?

.dagger/build/sdk.go
```diff
@@ -128,18 +129,22 @@ func (build *Builder) typescriptSDKContent(ctx context.Context) (*sdkContent, er
 }
 
 func (build *Builder) goSDKContent(ctx context.Context) (*sdkContent, error) {
-	base := dag.Container(dagger.ContainerOpts{Platform: build.platform}).
+	base := dag.Container().
```
jedevc, line 132: Oops, this also isn't entirely right. We should preserve the platform of the build image.

.dagger/build/sdk.go
jedevc: Did this, since we should also be setting the `Platform` correctly - otherwise, we're building an image for the wrong platform entirely :scream: :scream:

sipsma (approved):

Issue #9018, opened 2024-11-21
shell: persist variables across interactive prompt lines


helderco:
## Motivation

Variables do work in `dagger shell` but when in interactive mode they have to be defined in the same prompt because each line is a separate runner execution of the interpreter.

They can come in handy in a session so it’s worth it to make it work.

## Specification

Not yet clear on the solution, but have I few things I want to explore.

Issue #9035, opened 2024-11-22
Shell: navigation model


shykes:
## Overview

This is a proposal for a complete navigation model in dagger shell. It defines:

- Filesystem navigation
- Module loading
- Command execution

And the relationship between those features.

## Filesystem navigation

- The dagger shell presents a virtual filesystem to the user, which they can navigate
- Navigation is done by changing the shell's current location in the virtual filesystem: the working directory or *workdir*.
- A location in virtual filesystem has 2 components: *context* and *path*.
- A context is a virtual filesystem which acts as a sandbox for convenience and safety. The most common kinds of contexts are git repositories - either local or remote.

### Initial context

- When the shell is started, it gets an initial context
- By default, the initial context is inferred from the working directory in the system, by finding an enclosing git repository or dagger module. The initial context is *not* the entire host filesystem
- Alternatively, the shell can be started with a custom workdir (`dagger shell --workdir PATH|URL`) or with no workdir (`--no-workdir`)

### Navigating within a context

- `.cd PATH` changes the workdir within the current context.
- `.cd /` navigates to the root of the context, *not* of the host filesystem.
- The context root cannot be escaped
- `.`,  `..`  in a path have the usual meaning, eg. `./foo`, `../../foo` etc.

### Navigating across contexts

Dagger can navigate to other contexts beyond the initial context.

- `.cd URL` changes the current context and workdir to a remote filesystem. For example `.cd github.com/dagger/dagger/docs@v0.13.0` changes the context to `github.com/dagger/dagger@v0.13.0` and the workdir to `/docs`.
- `.cd` with no argument navigates back to the initial context

### Other navigation builtins

- `.back [N]` goes back N steps in the navigation (default 1). Like the back button in a browser :)
- `.ls [PATH|URL]` lists contents of the specified directory
- `.pwd` prints the current workdir
- The workdir is shown in the prompt

## Module loading

Modules are automatically loaded from the workdir. For example, navigating to the dagger repo will automatically load the corresponding module:

```console
⋈ .cd github.com/dagger/dagger@v0.13.0
github.com/dagger/dagger@v0.13.0 ⋈ 
```

## Command execution

To execute each command in a pipeline, the dagger shell looks up the corresponding function from the current context, and calls it. Function lookup follows this sequence:

1. If the command is the name of a function in the current object, the function is executed

2. If the command is the name of a dependency in the current module, the dependency is loaded and its constructor executed

3. If the command is the name of a function in the *stdlib*, the function is executed

[...]

PR #9038, opened 2024-11-22
shell: persist variables across interactive prompt lines

helderco:
Fixes https://github.com/dagger/dagger/issues/9018

### DIFF
--- a/cmd/dagger/shell.go
+++ b/cmd/dagger/shell.go
@@ -277,7 +277,6 @@ func (h *shellCallHandler) run(ctx context.Context, reader io.Reader, name strin
 		})
 	}()
 
-	h.runner.Reset()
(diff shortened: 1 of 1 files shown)


jedevc (approved):

jedevc:
LGTM, would also be good to have a test though.

helderco:
> would also be good to have a test though.

Agreed, but I haven't yet spent time on how to test the interactive mode[^1] as we're still prototyping and I need to move fast while things are shifting. I have to change some foundational aspects next, for example (see https://github.com/dagger/dagger/issues/9035). I plan on adding some tests for the REPL soon after that. At least for completions.

[^1]: Planned on looking for `Terminal` tests for that.

PR #9042, opened 2024-11-25
dagui: misc tweaks for frontend optimizing

vito:
A grab bag of engine side changes motivated by v3 - see each commit message for details/context.


vito:
I will likely be on PTO when this merged. I backed out the riskiest change (simplify `Simplify`) and all tests are passing so I think it's OK, but it's also not a hard requirement for anything so if y'all deem it too risky that's OK. The v3 web UI just points to this via a submodule, so there are no blockers on that side.

dagql/dagui/spans.go
```diff
@@ -43,23 +47,22 @@ type Span struct {
 // Snapshot returns a snapshot of the span's current state, incrementing its
 // Version with every call.
```
sipsma, line 48: comment out of date about Version; I was also a little surprised that calling Snapshot results in us assuming it's Remote and still not totally following why

sipsma (approved):

dagql/dagui/spans.go
jedevc: I *think* this makes sense - I think it's just the naming (maybe a better to think of it might be `Final`)? Once we snapshot a span, it's in its final state, there's no need to calculate anything else on it.

jedevc (approved):

dagql/dagui/spans.go
vito: Feel free to rename to your hearts content! Was definitely a hasty choice

jedevc: Excellent :D will rename, and merge :tada:

(now go enjoy PTO! :clinking_glasses: :clinking_glasses:)

jedevc:
@sipsma had to push a commit to fix a broken test - looks like we're now trimming the logs under `upload`. Do we want to keep that?

sipsma:
> had to push a commit to fix a broken test - looks like we're now trimming the logs under upload. Do we want to keep that?

Trying locally it looks like they still show up at sufficient verbosity level, which LGTM

PR #9052, opened 2024-11-25
fix: insert `Module.withDependencies` in correct order

jedevc:
Previously, calling `dagger install` with no name would bizarrely insert it into the beginning of the `dagger.json` `dependencies` field. Then calling `dagger develop` would fix it.

This was an absurd little pet-peeve, so took some time to fix it - it will be nice when we can clean up `ModuleSource.resolveFromCaller`, since trying to understand that for this bug was by far the trickiest bit to actually understand and track.

This is what was happening:
- We call `updateDeps` to update our dependencies.
- We were calling `ModuleSource.dependencies` (which merges `withDependencies` nicely) and getting all ordered dependencies.
- However, no name was specified in the `withDependencies` call, so the name was actually the empty string - so was appearing first in the lexigraphic sort.
- Then, later in `updateDeps`, we fill in the missing name - this is correct, but is now too late for the sort, so we have the wrong order.

The solution is fairly simply - `dependencies` should fill in missing names if possible.

We *could* have added `updateDeps` as an alternative, but this is actually not great, since this means that `dependencies` would still be returning `dependencies` that have no name - which is a potential footgun for the next person coming along.

### DIFF
--- a/core/integration/module_cli_test.go
+++ b/core/integration/module_cli_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/dagger/dagger/core/modules"
 	"github.com/dagger/dagger/testctx"
 	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
 )
 
 type CLISuite struct{}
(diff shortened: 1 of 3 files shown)


sipsma:
Looks like there's some legit test failures here, also this is worth it's own test coverage too

jedevc:
Fixed *most* of the tests and added a new one to check the sorted behavior.

Noticed one of them is still failing, will investigate though.

jedevc:
Think all of these are fixed now!

jedevc:
This isn't an urgent change, moving this out of the milestone.

vito (approved):

PR #9055, opened 2024-11-26
new module: github.com/dagger/dagger/releaser

shykes:
Start breaking out release logic into a standalone module. A few notes:

- We only break out the minimum to unblock breaking out the helm module (see #8888 )
- There is more to break out, that justifies a standalone releaser module
- Initially this was called 'release', but had to rename to avoid a codegen error


cwlbraa (approved):

releaser/.dagger/main.go
```diff
@@ -0,0 +1,117 @@
...
+func (r Releaser) GithubRelease(
+	ctx context.Context,
+	// GitHub repository URL
+	repository string,
+	// Tag for the GitHub release
+	// eg. sdk/typescript/v0.14.0
+	tag string,
+	// The target for the release
+	// eg. 🤷‍♂️
+	// FIXME: what's the difference with 'tag'? Who knows, it wasn't documented - SH Nov 2024
```
jedevc, line 56: This function will create the tag if it doesn't exist.

In this case, `target` is the commit/branch/tag/ref to use, and the `tag` is the new tag to create.

jedevc (approved):
LGTM. We should add more stuff into this module over time, ideally, *all* releases should be triggered from here.

PR #9056, opened 2024-11-26
ci: publish install scripts on nightly builds

jedevc:


gerhard (approved):
Let's see how well this works in practice!

PR #9057, opened 2024-11-26
ci: invalidate install scripts cache after publishing

jedevc:
This invalidates the cloudfront cache for `install.sh` and `install.ps1` after publishing.

Also, we change some secrets to vars, while we're here.


gerhard (approved):
Let's try if this works & as a follow-up we should only invalidate if `install.sh` has changed / been uploaded to s3.

PR #9064, opened 2024-11-27
docs: Update images and code in feature pages

vikram-dagger:
This commit updates the screen recording GIFs for the new TUI, and also fixes some Go code snippets in the feature pages


levlaz:
Overall this looks good, but I tested a few of these examples and it seems the new examples here take 3x longer than the old ones, is this something specific to your system or  are you noticing this trend @vikram-dagger? 

On my end running the same build example takes about ~6 seconds vs the 30 I see here.

shykes:
Same comment, it's important that examples appear snappy.

vikram-dagger:
> Overall this looks good, but I tested a few of these examples and it seems the new examples here take 3x longer than the old ones, is this something specific to your system or are you noticing this trend @vikram-dagger?

Updated now

mircubed:
@shykes @levlaz blocking question from @vikram-dagger for this PR related to 0.15 launch: 

Are these recordings still too slow? 

If yes, should we re-record these now with a more powerful server or should we wait until 0.15 to re-record? We are unsure how much the TUI has changed for these commands between 0.14 and 0.15.

gerhard (commented):
https://github.com/dagger/dagger/issues/9373 was created since a user mentioned incorrect examples.

I'm approving this so that we fix the examples, we can improve the speed as a follow-up. It's better for them to be correct than fast.

gerhard (approved):
LGTM 👍

PR #9066, opened 2024-11-27
silence incredibly verbose call ID from logs

vito:
This is printing strings upwards of 200kb in some scenarios, which causes some log systems to choke, leading to a SIGPIPE and the engine dying.

### DIFF
--- a/engine/buildkit/executor_spec.go
+++ b/engine/buildkit/executor_spec.go
@@ -1195,7 +1195,7 @@ func (w *Worker) runContainer(ctx context.Context, state *execState) (rerr error
 	if w.execMD != nil {
 		lg = lg.WithField("caller_client_id", w.execMD.CallerClientID)
 		if w.execMD.CallID != nil {
-			lg = lg.WithField("call_id", w.execMD.CallID.Display())
+			lg = lg.WithField("call_id", w.execMD.CallID.Digest)
 		}
 		if w.execMD.ClientID != "" {
 			lg = lg.WithField("nested_client_id", w.execMD.ClientID)


gerhard (approved):
We are going to merge this so that we can get a dev build to try it out.

This bug is incredibly specific:
- the log limit for the Engine runtime (e.g. Docker / K8s etc.)
- the amount of data that is getting logged (>128KB)

We doubt that anyone has this this or even would know what to look for. Basically the Engine exits with status code `2`.

Stracing it with the following command `strace -s 80 -p 1380 -f -e '!futex,restart_syscall,epoll_wait,epoll_pwait,select,getdents64,close,sched_yield,epoll_ctl,accept4,setsockopt,getsockname,rt_sigreturn,rt_sigprocmask,rt_sigaction,fcntl,mprotect,mmap,nanosleep'` helped us realize what was happening:

```console

[pid 10999] write(2, "time=\"2024-11-27T18:46:20Z\" level=debug msg=\"exporting metrics to clients\" clien"..., 85) = 85
[pid 10999] write(2, "time=\"2024-11-27T18:46:21Z\" level=debug msg=\"exporting metrics to clients\" clien"..., 85 <unfinished ...>
[pid 10999] write(2, "time=\"2024-11-27T18:46:47Z\" level=debug msg=\"starting container\" args=\"[/.init /"..., 216436 <unfinished ...>
[pid 10999] write(2, "test@sha256:db142d4...10ae479dbf92f3b13d693fd1c91053da9979728cceb1dc68\\\").withMo"..., 85364) = -1 EPIPE (Broken pipe)
```

vito:
Just to give an idea of how large this could get :P

image

sipsma:
Oh there was actually a backlog item for fixing this ([linear only](https://linear.app/dagger/issue/DEV-4665/iddisplay-is-profoundly-inefficient-for-large-dags))

Removing it for now SGTM if it's causing problems but I'll send out a PR sometime to re-add it in such a way that the DAG is de-duped and/or more compact since it's **extremely** useful for debugging since it allows you to follow what containers correspond to what.

sipsma:
Also worth noting this same problem occurs anywhere that `.Display()` is called, which after this change is mostly error cases but *might* still show up in the logs at times @vito

gerhard:
We confirmed that this fixed our issue. The pipeline finished running. Follow-ups sound good @sipsma 👍

Issue #8967, opened 2024-11-15
✨ Allow $schema field in dagger.json


JacobLey:
### What are you trying to do?

I am trying to add a `$schema` field to my `dagger.json` file.

e.g.
```json
{
  "$schema": "https://docs.dagger.io/reference/dagger.schema.json",
  "name": "my-module",
  "engineVersion": "v0.14.0",
  "sdk": "go",
  "source": "."
}
```

Currently my IDE flags the `$schema` field as not allowed, because it is not explicitly declared [in the schema](https://docs.dagger.io/reference/dagger.schema.json).

See how other JSON schemas support this, like [JSON schema's meta schema itself](https://json-schema.org/draft/2020-12/meta/core).

The field is also deleted when I run `dagger develop`. I would expect it to be retained, to the exact same URL as before (i.e. perhaps I want a specific version's schema)

### Why is this important to you?

Whenever I am working with a new piece of software/framework, there is a fantastic amount of discoverability/self-documentation available in config files when an IDE can auto-populate descriptions based on a `$schema` field.

I don't necessarily expect Dagger itself to enforce the schema abides by the specified `$schema` URL (can continue validating however it currently does based on current dagger version, with exception of allowing `$schema`).
I am primarily using the field for my own developer experience.

### How are you currently working around this?

Manually adding in the `$schema` field when I want to make edits, to ensure I am using the right field names/updating the right settings.

JacobLey:
Naively, editing the JSON Schema file to support this field is trivial. Would be happy to open a PR for that if agreed upon.

The trickier part is the "don't delete my `$schema` if it exists" part

JacobLey:
Took a swing at a PR: https://github.com/dagger/dagger/pull/9069

marcosnils:
@JacobLey  thx for taking this one. It seems to me that we need to do a bit more work here as IIUC schemas should be versioned with the engine. Apparently, we're only publishing a single version of the schema currently which I assume will cause issues given that we should set the same version as `engineVersion` has.

JacobLey:
> So this issue is more around improving the DX right?

Right, my original issue was that my IDE had no clue what was "right", and the only way to test the correct format of the file is to run dagger. Adding the `$schema` field improved the feedback loop. Personally it is often easier to edit files locally than using CLI to configure them.

Then dagger was nuking that field when I tried running `dagger develop`, so I had to manually add it back. So I mostly wanted it to persist whatever I put there.

> given that we should set the same version as engineVersion

[...]

PR #9069, opened 2024-11-28
config: Allow $schema in dagger.json

JacobLey:
Closes https://github.com/dagger/dagger/issues/8967

Took a first swing at allowing $schema in the parsed config, and writing it back _if_ it exists.

It does not write a value by default, nor does it actually enforce that the json file abides by the schema (technically the URI could point to a completely random schema, or not resolve in the first place).

Primarily for the benefit of local development, because IDEs can help with descriptions and autocompletion if a $schema property is available

### DIFF
--- a/core/integration/module_config_test.go
+++ b/core/integration/module_config_test.go
@@ -45,7 +45,7 @@ func (ConfigSuite) TestConfigs(ctx context.Context, t *testctx.T) {
 		baseWithNewConfig := baseWithOldConfig.With(daggerExec("develop"))
 		confContents, err := baseWithNewConfig.File("dagger.json").Contents(ctx)
 		require.NoError(t, err)
-		var modCfg modules.ModuleConfig
+		var modCfg modules.ModuleConfigWithUserFields
 		require.NoError(t, json.Unmarshal([]byte(confContents), &modCfg))
 		require.Equal(t, "test", modCfg.Name)
 		require.Equal(t, "go", modCfg.SDK)
@@ -56,6 +56,7 @@ func (ConfigSuite) TestConfigs(ctx context.Context, t *testctx.T) {
 		require.Equal(t, "dep", modCfg.Dependencies[0].Name)
 		require.Equal(t, ".", modCfg.Source)
 		require.NotEmpty(t, modCfg.EngineVersion) // version changes with any engine change
+		require.Empty(t, modCfg.Schema)
 
 		// verify develop didn't overwrite main.go
 		out, err := baseWithNewConfig.With(daggerCall("fn")).Stdout(ctx)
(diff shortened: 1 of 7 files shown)


core/integration/module_config_test.go
```diff
@@ -242,6 +243,36 @@ func (ConfigSuite) TestConfigs(ctx context.Context, t *testctx.T) {
...
+			WithWorkdir("/work").
+			With(daggerExec("init", "--source=.", "--name=test", "--sdk=go")).
+			WithNewFile("/work/main.go", `package main
+			type Test struct {}
+
+			func (m *Test) Fn() string { return "wowzas" }
+			`,
+			).
+			WithNewFile("/work/dagger.json", `{
+				"$schema": "https://docs.dagger.io/reference/dagger.schema.json",
```
pjmagee, line 261: so this would end up like:

`https://docs.dagger.io/reference/schema/v0.14.0/dagger.schema.json`

core/modules/config.go
```diff
@@ -13,6 +13,9 @@ const EngineVersionLatest string = "latest"
 
 // ModuleConfig is the config for a single module as loaded from a dagger.json file.
 type ModuleConfig struct {
+	// The self-describing json $schema of
+	Schema string `json:"$schema"`
```
jedevc, line 17: Hm, I was hoping to avoid adding this as a property on here. The code that interacts with `ModuleConfig` shouldn't be able to access this. Ideally it could be a private field - but then it can't be accessed by the `json` parser either.

Doing anything else feels bizarrely hacky though, so happy to leave as is. Maybe make a note that this field is *only* for user usage, and shouldn't be modified/accessed?

core/integration/module_config_test.go
jedevc: It's not currently versioned, so as written is correct.

jedevc (approved):

core/integration/module_config_test.go
jedevc: It *will* end up being correct though - see https://github.com/dagger/dagger/pull/9070.

pjmagee: Ahh, yeah, makes sense!

JacobLey: FWIW, if:

1. someone were to use 0.14.0
2. dagger releases some future version 0.15.0
3. user does _not_ update their local version to 0.15.0 (still on 0.14.0)
4. Their schema will reflect 0.15.0 which may include fields that are not supported by their local 0.14.0

If we wanted a versioned schema, best place would probably be from github itself: https://raw.githubusercontent.com/dagger/dagger/refs/tags/v0.14.0/docs/current_docs/reference/dagger.schema.json

But that also requires keeping this link up to date with whatever version you are using. _Personally_ that is more work than its worth for the benefit.

core/modules/config.go
JacobLey: Some alternative would be to do most of the parsing without this schema field.

Then right before writing it back, we can parse out the `$schema` and attach it to a combined struct.

But personally fine with tacking on a comment to not pay any mind to `Schema`

jedevc: Yeah, that feels like a lot of effort :scream: Happy with the comment!

jedevc:
@JacobLey this needs a DCO sign-off commit: https://github.com/dagger/dagger/pull/9069/checks?check_run_id=33651982122

JacobLey:
Added DCO

Couple tests are still failing though, I think I might still be missing something.

core/modules/config.go
```diff
@@ -42,6 +43,18 @@ type ModuleConfig struct {
...
 }
 
+type ModuleConfigUserFields struct {
+	// The self-describing json $schema
+	Schema string `json:"$schema,omitempty"`
+}
+
+// ModuleConfigWithUserFields is the config for a single module as loaded from a dagger.json file.
+// Includes additional fields that should only be set by the user.
+type ModuleConfigWithUserFields struct {
```
JacobLey, line 53: Took a swing at separating these "user fields" from the dagger-configurable fields.

Needs to be set up this way based on https://stackoverflow.com/questions/52294027/golang-structure-doesnt-unmarshal-properly-when-using-a-custom-unmarshal-for-a/52305457#52305457

JacobLey:
More tests failing...

Some were lint issues that I fixed, but some like [java-dev](https://github.com/dagger/dagger/actions/runs/12092278846/job/33721817249?pr=9069) appear to be failing on
> ./sdk_elixir.go:52:19: sdkDev.Lint undefined (type *dagger.ElixirSDKDev has no field or method Lint)

I don't _think_ I would have broken that...?

Then things like [test-cli-engine](https://github.com/dagger/dagger/actions/runs/12092278849/job/33721816547?pr=9069) just fail without a good explanation?

Locally I've just been running `dagger -m . call test specific '--run=TestConfig' --race=true --parallel=16`, but that fails as well, at least on slightly more sensical failure.
[This assertion](https://github.com/dagger/dagger/blob/e93be82226824cdd8a281a2d05ecfd7f28abb619/core/integration/module_config_test.go#L57) fails, finding `".dagger"` instead of `"."`

jedevc (approved):
LGTM :tada:

I rebased, and also updated some of the `UnmarshalJSON` changes - they didn't look right, and seemed to be causing some of the failures you were noticing.

PR #9070, opened 2024-11-28
docs: version dagger schema

jedevc:
Resolves @marcosnils' comment here: https://github.com/dagger/dagger/issues/8967#issuecomment-2505105814

> Apparently, we're only publishing a single version of the schema currently which I assume will cause issues given that we should set the same version as `engineVersion` has.

To resolve this, we generate versioned schemas. Note: we treat the "next" version as the generated schema to keep regenerating on `main` - once we release a version, then this increases, freezing that schema, and will generate a next.

Note, there's no protection for accidentally generating schemas for versions that will never actually be released - so this requires some discipline in `.changes/.next`, just worth watching out for.


marcosnils (approved):

docs/netlify.toml
```diff
@@ -1197,6 +1197,12 @@
   to = "/quickstart/428201/custom-function"
   status = 301
 
+[[redirects]]
+  # redirect schema to the versioned one
+  from = "/reference/dagger.schema.json"
+  to = "/reference/schemas/v0.14.0/dagger.schema.json"
```
marcosnils, line 1203: shall we remove this redirection altogether and force that all `/referece/*` URLs have to be versioned? I'd prefer to bite the bullet now instead of having to find a way to automatically update this redirect each time a version is released. 

WDYT?

jedevc: This redirection is just so that anyone who was using the schema address before now actually gets redirected to a versioned one (so we don't have to keep publishing that file).

We don't need to automatically update it - we just keep it pinned to v0.14.0 - the main schema address to use will always have versions in it, it's just to avoid breaking users who currently are using it.

marcosnils: > This redirection is just so that anyone who was using the schema address before now actually gets redirected to a versioned one (so we don't have to keep publishing that file).

Yep, that's what I imagined, thing is that we'll have to remember that this particular URL  points to v0.14.0's schema which is super confusing to me.  I'd expect the URL **without** the version to either redirect you to the `latest` schema, or return 404. That's mostly the reason in my comment above I've suggested removing the redirection altogether and to break it for some users in lieu of better consistency.

vikram-dagger: I agree with @marcosnils. I would either remove this redirect entirely, or adjust it to point to a moving `latest`.

jedevc:
Put this in draft, since I'm not entirely sure anymore about wanting to version this :thinking: (and didn't want to accidentally merge)

jedevc:
Yeah, decided - I don't really want to version this file. Closing.

PR #9072, opened 2024-11-28
chore: avoid direct use of `trace.SpanKindInternal`

jedevc:
Instead, use `telemetry.Internal`.

See @vito's comment here: https://github.com/dagger/dagger/pull/8818#discussion_r1861051922

`parseRefString` is indeed showing up in weird places, and is noisy :cry:


vito (approved):
Nice, thanks

PR #9073, opened 2024-11-29
chore: bump next version to v0.15.0

jedevc:
Next release is going to be v0.15.0, with some exciting changes.

### DIFF
--- a/.changes/.next
+++ b/.changes/.next
@@ -8,4 +8,4 @@
 # automagically determined from release note entries (and then adding to the
 # patch release).
 
-v0.14.1
+v0.15.0
--- a/.dagger/engine.go
+++ b/.dagger/engine.go
@@ -176,26 +176,28 @@ func (e *DaggerEngine) Service(
 // Lint the engine
 func (e *DaggerEngine) Lint(
 	ctx context.Context,
+	pkgs []string, // +optional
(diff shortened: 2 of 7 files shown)


jedevc:
Should also update `future_test.go` and shepherd those tests into the main suite.

jedevc:
Hm, the daggerverse preview has failed @gerhard :thinking:

```
Error: response from query: input: daggerverse.deployPreviewWithDaggerMain resolve: call function "DeployPreviewWithDaggerMain": process "/runtime" did not complete successfully: exit code: 2

Stdout:
invoke: GET https://api.github.com/user: 401 Bad credentials []
```

It definitely shouldn't be triggering on PRs (creds aren't available) - but I'm not quite sure that it *should* trigger on `.changes/.next` even on main - I'm just bumping to the next version here. We want it to trigger on the PR right before the release.

We could trigger it on a new markdown file in `.changes/`? That would be the case we care more about.

marcosnils (approved):

jedevc:
Will fix https://github.com/dagger/dagger/pull/9073.

jedevc:
Failure is known failure from https://github.com/dagger/dagger/pull/9094, merging now to get release pieces started :tada:

PR #9076, opened 2024-11-29
docs: Add Dagger shell docs

vikram-dagger:


vikram-dagger:
This draft is heavily based on https://github.com/dagger/dagger/pull/8611.

Some pending questions:
- How do we position this feature vs. our existing content on shell scripting disadvantages? When should a user reach for this (and when not)?
- Should we add a `Shell` tab equivalent to all existing function code snippets?
- - If yes, is there any functionality that cannot be replicated?
- Should we add a new `Shell` tab for all existing `dagger call` commands, showing the `dagger shell` version?
- - If yes, should this be shown in interactive or non-interactive modes?
- - Note this is different from the previous question as it's invoking functions, not writing them

samalba:
I like it, I knew you were going to see this differently and I trust your taste more than mine for those things.

Issue #9091, opened 2024-12-02
test-cli-engine is failing due to GitLab public sad path not failing as expected


gerhard:
```console
Attempting to create SSH agent socket at: /tmp/ssh-agent4011640326/ssh-agent.sock

        Error Trace:    /app/core/integration/suite_test.go:349
                                                /app/core/integration/module_cli_test.go:893
                                                /app/testctx/testctx.go:168
        Error:          An error is expected but got nil.
        Test:           TestCLI/TestDaggerInstall/GitLab_public/git/sad
Cleaned up SSH agent: /tmp/ssh-agent4011640326/ssh-agent.sock
```
https://dagger.cloud/dagger/traces/73d5dbb25b6c4d5f74ae07a84a57bb75

![image](https://github.com/user-attachments/assets/e971e1f7-a8d3-4784-a705-16b8b03e96ae)

Do you know why that may be the case @grouville ?

cc @jedevc since we talked about it today

PR #9094, opened 2024-12-03
fix(vcs): ensure `RepoRootForImportDynamic` trims trailing `../`

grouville:
Fixes https://github.com/dagger/dagger/issues/9091

CI was recently broken due to a change on GitLab's end, when parsing a ref.

In order to isolate the root of the repo to the subdir, we rely on Go's logic:
1. Statically analyze the ref
2. If no match, query the git provider with `ref?go-get=1`
    
It turns out that GitLab recently introduced a change that normalizes refs with `../`, when querying refs such as `gitlab.com/foo/bar/baz/../..?go-get=1`, it now becomes: `gitlab.com/foo/?go-get=1`
    
I suspect this change: https://gitlab.com/gitlab-org/gitlab-foss/-/blame/master/lib/gitlab/middleware/go.rb#L57, introduced 5 months ago, but that could have just been deployed last week.
    
Several options were considerable, but it seems acceptable to directly remove them from the ref itself, when being the trailing elements


engine/vcs/vcs_test.go
```diff
@@ -184,6 +184,14 @@ func TestRepoRootForImportPath(t *testing.T) {
...
 			},
 		},
+		{
+			"gitlab.com/dagger-modules/test/more/dagger-test-modules-public/../../..",
+			&RepoRoot{
+				VCS:  vcsGit,
+				Repo: "https://gitlab.com/dagger-modules/test/more/dagger-test-modules-public",
+				Root: "gitlab.com/dagger-modules/test/more/dagger-test-modules-public",
+			},
+		},
```
jedevc, line 194: What about something like `gitlab.com/dagger-modules/test/more/dagger-test-modules-public/dagger/weirdpath/..`

This *should* resolve the `Root` to `gitlab.com/dagger-modules/test/more/dagger-test-modules-public/dagger`, but I think by stripping the `../` early, we're going to get the wrong result, and instead get `gitlab.com/dagger-modules/test/more/dagger-test-modules-public/weirdpath` which doesn't exist.

Should we re-append the stripped `../`s at the end to the `Root`? I think that might resolve this.

.changes/unreleased/Fixed-20241203-071352.yaml
```diff
@@ -0,0 +1,9 @@
+kind: Fixed
+body: |-
+    Ensure `RepoRootForImportDynamic` properly trims trailing `../` in ref
+
+    Fixes https://github.com/dagger/dagger/issues/9091
```
jedevc, line 5: `RepoRootForImportDynamic` isn't an exposed function - so we shouldn't include it in the changelog. Changelogs are for end-users, so we need to phrase this as an end-user fix.

Does this even need a changelog? From my understanding, we're only fixing an error case - the only refs this resolves weirdly now are refs that were actually invalid.

jedevc:
> It turns out that GitLab recently introduced a change that normalizes refs with `../`, when querying refs such as `gitlab.com/foo/bar/baz/../..?go-get=1`, it now becomes: `gitlab.com/foo/?go-get=1`

Lol, this behavior is so very strange. I guess it makes sense - anything that would normalize URLs incoming to gitlab would cause this. :cry:

.changes/unreleased/Fixed-20241203-071352.yaml
grouville: Ahah, the only time I actually remember to add one it's not necessary 🤣

engine/vcs/vcs_test.go
grouville: So, not sure to get this one.

We do strip, but we strip in `RepoRootForImportDynamic`, which is right before our call to the ref + `go-get=1`.
That way, we are sure that the endpoint does not normalize it suddenly.

It does actually work, and I _believe_ that this is what we actually want.
_Example_: 

This works: `https://gitlab.com/dagger-modules/test/more/dagger-test-modules-public/weirdpath?go-get=1`, but, this `https://gitlab.com/dagger-modules/test/more/dagger-test-modules-public/weirdpath/../../?go-get=1`  does not.

_Working example:`gitlab.com/dagger-modules/test/more/dagger-test-modules-public/weirdpath?go-get=1` _
image

_Failure: `gitlab.com/dagger-modules/test/more/dagger-test-modules-public/weirdpath/../../?go-get=1`_
image


Why ? Because it shortens (at normalization) the ref, making the ref outside of the real repo:

The root is: 
- `gitlab.com/dagger-modules/test/more/dagger-test-modules-public` ; `go-get=1` returns the proper Root
- `gitlab.com/dagger-modules/test/more/dagger-test-modules-public/weirdpath/depth1/depth2...` : `go-get=1` returns the proper Root (any subpath inside a real public repo)
- `gitlab.com/dagger-modules/test/more/dagger-test-modules-public/weirpath/..`: would work too, as `go-get=1` is still inside the repo scope, even after normlization into `gitlab.com/dagger-modules/test/more/dagger-test-modules-public/?go-get=1`

Now, without trimming:
- `gitlab.com/dagger-modules/test/more/dagger-test-modules-public/..` would lead to `gitlab.com/dagger-modules/test/more?go-get=1`, which does not fail but return the wrong root, as shown on the failure picture

But with trimming: `gitlab.com/dagger-modules/test/more/dagger-test-modules-public/..` becomes `gitlab.com/dagger-modules/test/more/dagger-test-modules-public/` which leads to this dynamic query: `gitlab.com/dagger-modules/test/more/dagger-test-modules-public?go-get=1` which is correct

jedevc: Yup, you're entirely right. This looks good to me!

jedevc (approved):

Issue #9019, opened 2024-11-21
shell: consolidate `.doc` and `.help`


helderco:
## Motivation

`.doc` is used for some things and `.help` for others. We can simplify by expanding the scope of `.help` and replacing `.doc` with `.help`.

## What about `.deps`?

We’ve already merged `.functions` into `.doc` for an object type. We could also merge `.deps` into `.help` for a module.

## Bikeshedding

Let’s take the opportunity to bikeshed improvements to what `.help` looks like for:
- A module (or main object)
- An object type
- Any other type
- A function
- A builtin

Issue #8990, opened 2024-11-18
shell: don't print objects by default


shykes:
## Overview

In the context of the shell, it's no longer necessary to auto-print objects. Instead, there should be a `.print` and perhaps also a `.json` builtin, that you can easily add to the pipeline. By default, a pipeline that succeeds should simply print nothing.

## Examples

No print when returning an object:

```
⋈ .http https://dagger.io
⋈
```

To learn more, `.doc`:

```
⋈ .http https://dagger.io | .doc
File

DESCRIPTION

A file.

REFERENCE

https://docs.dagger.io/api/reference/#definition-File

FUNCTIONS

  contents          Retrieves the contents of the file.
  digest            Return the file's digest. The format of the digest is not guaranteed to be stable between releases of Dagger. It is guaranteed to be stable between invocations of the same
                    Dagger engine.
  export            Writes the file to a file path on the host.
  name              Retrieves the name of the file.
  size              Retrieves the size of the file, in bytes.
  sync              Force evaluation in the engine.
  with-name         Retrieves this file with its name set to the given name.
  with-timestamps   Retrieves this file with its created/modified timestamps set to the given time.

SEE ALSO

".doc <function>" for more information on a function.
```

To know the type, `.type`:

```
⋈ .http https://dagger.io | .type
File
```

To print the contents, `.print`. Note the absence of `_Type` field, another hack we no longer need:

```
⋈ .http https://dagger.io | .print
digest: sha256:e37c20d68d9cf63cd164ef06220b6b7b6ea4bc98aed5cc982e3de09c2622c028
name: 78c79fe99761c116238476f0b5a6a528b621e80398bfbaaa991abe41c4942a0d
size: 838056
```

A special case to encode the object to JSON:

```
⋈ .http https://dagger.io | .json
{ 
    "digest": "sha256:e37c20d68d9cf63cd164ef06220b6b7b6ea4bc98aed5cc982e3de09c2622c028",
    "name": "78c79fe99761c116238476f0b5a6a528b621e80398bfbaaa991abe41c4942a0d",
    "size": 838056
}
⋈
```

helderco:
Is there a need for `.type` when that already shows in  `.doc`? At **minimum** this is what `.doc` shows:
```shell
⋈ .directory | sync | .doc
CUSTOM SCALAR
  DirectoryId
```
 Format:
```
<KIND>
  <type>
```
Another example
```shell
dagger-dev ⋈ version | .doc
SCALAR
  string

[...]

Issue #8969, opened 2024-11-15
shell: better support with-exec


shykes:
## Problem

In the shell, `.container | with-exec` requires comma-separated arguments:

```console
.container |
with-exec echo,hello,world |
stdout
```

But it should be instead:

```console
.container |
with-exec echo hello world |
stdout
```

## Solution

Change the syntax of `with-exec` to support the more natural `with-exec echo hello world`.

This is possible to do cleanly, because `with-exec` has a single required argument, and it's an array. In that case, the shell parser should parse arguments in the more natural way I propose. All functions which match that signature would benefit, not just `with-exec`.

helderco:
This is doable, yes. I can make it work when the last required argument is a list, or any optional arg. For **other required args** or for explicitness, we can use a built-in:
```
.container | with-exec $(.list echo "hello world") | stdout
```

Or some syntax sugar:
- `.container | with-exec $(@ echo "hello world") | stdout`
- `.container | with-exec (echo "hello world") | stdout`
- `.container | with-exec [echo "hello world"] | stdout`

These would produce:
```graphql
withExec(args: ["echo", "hello world"])
```

We can even easily support a list of lists this way.

helderco:
> No syntax sugar please. 

Ok, but are you extending that statement to `with-exec echo hello world`? Meaning you'd have to write `with-exec $(.list echo hello world)`?

shykes:
We discussed this live with Helder. We agreed on the following:

1. First a very simple solution to make `with-exec` cooler. Simply support positional args if there's only one required arg, and it's an array. Flags continue to work the same, without special convenience.

2. Meanwhile, we continue discussing possible builtins to make things easier. For example `.list`, `.json` etc. This requires more discussion and is orthogonal to the first point.

helderco:
There's still a problem with ambiguity though. It's easy enough to parse positional args until you hit a flag (`--`), but the command you want to pass to `withExec` can also have flags. I can reduce that conflict by stopping at the first occurrence of a flag that matches an optional argument (not just any flag), which for `withExec` is currently this list:
- `--use-entrypoint`
- `--stdin`
- `--redirect-stdout`
- `--redirect-stderr`
- `--expect`
- `--experimental-privileged-nesting`
- `--insecure-root-capabilities`
- `--expand`
- `--no-init`

[...]

PR #9097, opened 2024-12-03
shell: reimplement module loading and execution according to new model

helderco:
Implements first part of https://github.com/dagger/dagger/issues/9035
Fixes https://github.com/dagger/dagger/issues/9129

## Notes

- documentation for a module and its constructor/entrypoint:
```diff
-.load <module> | .doc
+.doc <module>
```
- documentation for a module’s main object:
```diff
-.load <module> | .config <required arguments>... | .doc
+<module> <required arguments>... | .doc
```
- set a module as the default for current session:
```diff
-.load <module> | .use
+.use <module>
```
- execute a stdlib function (disambiguated)
```diff
-.version
+.stdlib | version
```
- execute any core function
```diff
-.default-platform
-.core default-platform
+.core | default-platform
```

## Todo

- [x] Command execution lookup
- [x] Dynamic module loading through `.doc`
- [x] Document a module, separately from its main object
- [x] Migrate core API builtins to `.stdlib`
- [x] Command disambiguation with `.deps`, `.stdlib`, and `.core`
- [x] More tests
- [x] Use pinned versions for dependencies

## Bikeshedding

- [ ] Rename `.core` to something else?

## Next

- https://github.com/dagger/dagger/issues/8969
- https://github.com/dagger/dagger/issues/9019
- https://github.com/dagger/dagger/issues/8990
- Filesystem navigation (from [proposal](https://github.com/dagger/dagger/issues/9035))

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -862,7 +862,7 @@ func optionalModCmdWrapper(
 	}
 }
 
-// initializeCore loads the core type definitions.
+// initializeCore loads the core type definitions only
 func initializeCore(ctx context.Context, dag *dagger.Client) (rdef *moduleDef, rerr error) {
 	def := &moduleDef{}
 
@@ -873,7 +873,10 @@ func initializeCore(ctx context.Context, dag *dagger.Client) (rdef *moduleDef, r
 	return def, nil
 }
 
-// initializeDefaultModule loads the module's type definitions.
+// initializeDefaultModule loads the module referenced by the -m,--mod flag
+//
+// By default, looks for a module in the current directory, or above.
+// Returns an error if the module is not found or invalid.
 func initializeDefaultModule(ctx context.Context, dag *dagger.Client) (*moduleDef, error) {
 	modRef, _ := getExplicitModuleSourceRef()
 	if modRef == "" {
@@ -882,6 +885,8 @@ func initializeDefaultModule(ctx context.Context, dag *dagger.Client) (*moduleDe
 	return initializeModule(ctx, dag, modRef, true)
 }
 
+// maybeInitializeDefaultModule optionally loads the module referenced by the -m,--mod flag,
+// falling back to the core definitions
 func maybeInitializeDefaultModule(ctx context.Context, dag *dagger.Client) (*moduleDef, string, error) {
 	modRef, _ := getExplicitModuleSourceRef()
 	if modRef == "" {
@@ -890,6 +895,9 @@ func maybeInitializeDefaultModule(ctx context.Context, dag *dagger.Client) (*mod
 	return maybeInitializeModule(ctx, dag, modRef)
 }
 
+// initializeModule loads the module at the given source ref
+//
+// Returns an error if the module is not found or invalid.
 func initializeModule(
 	ctx context.Context,
 	dag *dagger.Client,
@@ -914,7 +922,8 @@ func initializeModule(
 	return initializeModuleConfig(ctx, dag, conf)
 }
 
-// maybeInitializeModule optionally loads the module's type definitions.
+// maybeInitializeModule optionally loads the module at the given source ref,
+// falling back to the core definitions if the module isn't found
 func maybeInitializeModule(ctx context.Context, dag *dagger.Client, srcRef string) (*moduleDef, string, error) {
 	if def, err := tryInitializeModule(ctx, dag, srcRef); def != nil || err != nil {
 		return def, srcRef, err
@@ -924,6 +933,10 @@ func maybeInitializeModule(ctx context.Context, dag *dagger.Client, srcRef strin
 	return def, "", err
 }
 
+// tryInitializeModule tries to load a module if it exists
+//
+// Returns an error if the module is invalid or couldn't be loaded, but not
+// if the module wasn't found.
 func tryInitializeModule(ctx context.Context, dag *dagger.Client, srcRef string) (rdef *moduleDef, rerr error) {
 	ctx, span := Tracer().Start(ctx, "looking for module")
 	defer telemetry.End(span, func() error { return rerr })
@@ -941,6 +954,7 @@ func tryInitializeModule(ctx context.Context, dag *dagger.Client, srcRef string)
 	return initializeModuleConfig(ctx, dag, conf)
 }
 
+// initializeModuleConfig loads a module using a detected module configuration
 func initializeModuleConfig(ctx context.Context, dag *dagger.Client, conf *configuredModule) (rdef *moduleDef, rerr error) {
 	serveCtx, serveSpan := Tracer().Start(ctx, "initializing module", telemetry.Encapsulate())
 	err := conf.Source.AsModule().Initialize().Serve(serveCtx)
@@ -971,6 +985,7 @@ type moduleDef struct {
 	// applying module-specific configs to the arg value.
 	Source *dagger.ModuleSource
 
+	// ModRef is the human readable module source reference as returned by the API
 	ModRef string
 
 	Dependencies []*moduleDependency
@@ -981,7 +996,10 @@ type moduleDependency struct {
 	Description string
 	Source      *dagger.ModuleSource
 
+	// ModRef is the human readable module source reference as returned by the API
 	ModRef string
+
+	// RefPin is the module source pin for this dependency, if any
 	RefPin string
 }
 
@@ -1218,7 +1236,8 @@ func (m *moduleDef) GetObjectFunction(objectName, functionName string) (*modFunc
 }
 
 func (m *moduleDef) GetFunction(fp functionProvider, functionName string) (*modFunction, error) {
-	// This avoids an issue with module constructors overriding core functions
+	// This avoids an issue with module constructors overriding core functions.
+	// See https://github.com/dagger/dagger/issues/9122
 	if m.HasModule() && fp.ProviderName() == "Query" && m.MainObject.AsObject.Constructor.CmdName() == functionName {
 		return m.MainObject.AsObject.Constructor, nil
 	}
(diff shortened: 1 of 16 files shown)


helderco:
@shykes, how do you feel about doing the filesystem navigation in a separate PR? This way you can try the current implementation for a bit to see if the design still makes sense.

Note that the "workdir" awareness in the proposal isn’t implemented yet. Module refs are resolved the same way as `dagger shell -m <ref>`. Adding that is closely tied to the filesystem navigation part. We also need to bikeshed on that `--workdir` flag. 

In the meantime I could move forward with the following changes, which bring more consistency to this one:
- https://github.com/dagger/dagger/issues/8990
- https://github.com/dagger/dagger/issues/9019

And moving the tab-completions forward too.

shykes:
> @shykes, how do you feel about doing the filesystem navigation in a separate PR? This way you can try the current implementation for a bit to see if the design still makes sense.

Yes agreed 👍

helderco:
This should be usable for a test drive now, to check how the new design feels before making more changes. There's more tests I need to add in the meantime.

shykes:
UX feels great!

A few small notes:

- When no module is loaded, `.doc` lists functions from `.core`, but they aren't actually reachable (nor should they)
- Calling a dep directly causes a glitch:

```
.use github.com/dagger/dagger
.deps | version
```

This ☝️ should return a module object, but returns a giant string that looks like the ID for that module.

But this works:

```
.deps | version | git
```

helderco:
> * When no module is loaded, `.doc` lists functions from `.core`, but they aren't actually reachable (nor should they)

Nice catch! How about this?

Screenshot 2024-12-06 at 12 10 55


> * Calling a dep directly causes a glitch:
> 
> ```
> .use github.com/dagger/dagger
> .deps | version
> ```
> 
> This ☝️ should return a module object, but returns a giant string that looks like the ID for that module.

Woah, you’ve actually stepped into two engine bugs at the same time here:
- https://github.com/dagger/dagger/issues/9122
- https://github.com/dagger/dagger/issues/6561

Basically, the `version` module is overriding the core `Query.version`. The reason why this doesn’t happen with `dagger call` or `dagger core` is because one only targets a module, not functions under  `Query`. And the other targets `Query`, without any module loaded.

Since `dagger shell` works with both, you’re seeing the `version` module overriding the core `Query.version` and returning a `Version` object. This confused `dagger shell` into thinking that it doesn’t return an object, so it proceeded to make the query without any sub-selections.

This in turn triggers the other bug where instead of an error, the engine is leaking internal details about the object. What you’re seeing is only the values in that structure.

You can see the full data in JSON with:
```console
❯ echo '{version}' | dagger query -m version
```

helderco:
@shykes, I fixed those issues. I also renamed `--no-load` to `--no-mod` so it pairs more closely to `--mod`.

helderco:
Finished the tasks for this part of the proposal. Resolving paths relative to the loaded module instead of the host workdir will come in a follow-up.

core/integration/module_shell_test.go
```diff
@@ -54,44 +54,342 @@ func (m *Test) Container() *dagger.Container {
...
+type Test struct{
+	Source *dagger.Directory
+}
+
+// Test version
+func (Test) Version() string {
+	return "test function"
 }
+
+// Encouragement
```
jedevc, line 85: :tada: :tada: I feel encouraged.

cmd/dagger/shell.go
```diff
@@ -1472,201 +1812,176 @@ func (h *shellCallHandler) registerBuiltins() { //nolint:gocyclo
...
-					return err
-				}
-
-				return cmd.Send(st)
+				return cmd.Send(h.newCoreState())
 			},
 		},
+		cobraToShellCommand(loginCmd),
+		cobraToShellCommand(logoutCmd),
+		// TODO: Add uninstall command when available.
```
jedevc, line 1983: `uninstall` is now available on `main` - if you want to add it now. `update` isn't merged yet #8839.

cmd/dagger/module.go
```diff
@@ -858,6 +862,101 @@ func optionalModCmdWrapper(
...
+	}
+	if !conf.FullyInitialized() {
+		return nil, fmt.Errorf("module must be fully initialized")
+	}
+
+	return initializeModuleConfig(ctx, dag, conf)
+}
+
+// maybeInitializeModule optionally loads the module's type definitions.
+func maybeInitializeModule(ctx context.Context, dag *dagger.Client, srcRef string) (*moduleDef, string, error) {
```
jedevc, line 918: Struggling to understand the differences at a glance between `try`/`maybe` + `initialize` + `module`/`moduleConfig`. We also still have `shellCallHandler.modDef`, and I'm not sure which ones I want to use in different scenarios, sorry!

Is there any way we can clarify this somehow?

cmd/dagger/module.go
```diff
@@ -871,13 +970,108 @@ type moduleDef struct {
...
 }
 
+type moduleDependency struct {
+	Name        string
+	Description string
+	Source      *dagger.ModuleSource
+
+	ModRef string
+	RefPin string
+}
```
jedevc, line 986: Not really something to deal with now - but it feels frustrating that we have to duplicate all of the structs + graphql queries for all the typedefs and dependency information. Is the reason we don't use our own codegen here because you can't get multiple fields at once?

If so, it feels like we should probably try and rectify that - even if the solution is to have special "cli-codegen" to generate this code. It feels like it's getting a bit complex - and might be worth the refactoring time.

cmd/dagger/module.go
```diff
@@ -1020,6 +1218,10 @@ func (m *moduleDef) GetObjectFunction(objectName, functionName string) (*modFunc
 }
 
 func (m *moduleDef) GetFunction(fp functionProvider, functionName string) (*modFunction, error) {
+	// This avoids an issue with module constructors overriding core functions
```
jedevc, line 1221: Can we link to https://github.com/dagger/dagger/issues/9122 inline?

cmd/dagger/shell.go
```diff
@@ -1472,201 +1812,176 @@ func (h *shellCallHandler) registerBuiltins() { //nolint:gocyclo
 			},
 		},
 		&ShellCommand{
-			Use:   ".doc [function]",
-			Short: "Show documentation for a type, or a function",
-			Args:  MaximumArgs(1),
+			Use: ".doc [module]\n<function> | .doc [function]",
```
jedevc, line 1815: This feels like a good start of where we could start splitting these out into separate files - the implementations of some of these are now quite long, and having the very deep indentation here is hard to read.

jedevc (approved):
Approving - so that this can make it in for the release tomorrow.

However, I do think that this continues to add to a very complex `shell.go` - I think there's some refactors that could be made.

cmd/dagger/shell.go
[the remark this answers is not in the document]
helderco: Yeah, I've been meaning to do this for a while, but haven't because things are still shifting. I actually want to put it in a sub-package to avoid the need for the shell prefix and avoid collisions.

cmd/dagger/module.go
helderco: I haven't properly documented things because it's still going to change some more, especially with the rest of:
- https://github.com/dagger/dagger/issues/9035

There's also a cleanup coming, I'm just trying to move fast on the last mile for this project.

cmd/dagger/module.go
helderco: Agree on the refactoring, but as I've explained, I'm trying to move fast while things are shifting. This will all be cleaned up.

cmd/dagger/module.go
helderco: Done

cmd/dagger/module.go
helderco: > Struggling to understand the differences at a glance between `try`/`maybe` + `initialize` + `module`/`moduleConfig`.

I added a few comments in code. The reason for this split is that there used to be one initialization function with a flag, whether it was for a command that required a module load (like call or functions), vs a command that only loaded core types (i.e., dagger core).

In shell however, there's dynamic module loading on a running session and fallback behaviors in different places. So we may need to load a module if it exists but fallback to core if not, or error, or ignore to try the next thing in the lookup. That's why the initialization function was split into this granularity, as these are used in different places.

> We also still have shellCallHandler.modDef, and I'm not sure which ones I want to use in different scenarios, sorry!

The `shellCallHandler.modDef` no longer exists. It's now a map (in `shellCallHandler.modDefs`) that works a memory cache for loaded modules, in conjunction with `shellCallHandler.modRef`. However, note that this will still change with the filesystem navigation part of https://github.com/dagger/dagger/issues/9035. Most likely `modRef` will become a `[]string` to implement stacked modules. And the cached map may also change due to the need for a forced reload.

cmd/dagger/shell.go
helderco: Done

shykes:
@helderco there is a regression in this branch as of today... When loading any module, I get:

```
 ! query module metadata: input: query:
│ ! query ModuleConfig($source: ModuleSourceID!) {
│ !   source: loadModuleSourceFromID(id: $source) {
│ !     asString
│ !     module: asModule {
│ !       name
│ !       initialize {
│ !         description
│ !       }
│ !       dependencies {
│ !         name
│ !         description
│ !         source {
│ !           asString
│ !           pin
│ !         }
│ !       }
│ !     }
│ !   }
│ ! }
│ ! error: parse selections: parse field "pin": ModuleSource has no such field: "pin"
```

helderco:
@shykes, how are you running that? The last commit requires the dev engine as well. Before then, changes were only in the CLI so you could use a dev engine from main for example.

shykes:
> @shykes, how are you running that? [1070484](https://github.com/dagger/dagger/commit/1070484052eb561ad28cfde1c8b45d32d26f8396) requires the dev engine from this branch as well. Before then, changes were only in the CLI so you could use a dev engine from main for example.

I installed it on my host system with `dev-export`, `docker load -i engine.tar`,  `docker tag registry.dagger.io/engine:$COMMIT`. But it seems like I'm doing it wrong, the CLI seems to download the engine image from a different commit?

jedevc:
Just pushed a commit to regen the ts client, since #9031 was merged.

PR #9098, opened 2024-12-03
feat: add `Directory.asGit`

### LINKED ISSUE #8720: ✨ Need "local" support for git repos that are directories on the filesystem

jedevc:
> [!WARNING]
>
> Implementation is not complete!

Fixes #8720.

Opening a draft PR to track this!


core/git.go
```diff
@@ -111,3 +274,154 @@ func (ref *GitRef) getState(ctx context.Context, discardGitDir bool) (llb.State,
...
+	}
+
+	bk, err := ref.Query.Buildkit(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get buildkit client: %w", err)
+	}
+	pbDef, _, err := bk.EngineContainerLocalImport(ctx, platforms.DefaultSpec(), tmpDir, nil, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to import container tarball from engine container filesystem: %w", err)
+	}
```
jedevc, line 396: As mentioned by @sipsma on [discord](https://discord.com/channels/707636530424053791/1003718839739105300/1313204645316919317), we should avoid the use of this API entirely - it caches very badly, and is not very fun to maintain.

> Updating/adding an LLB source is a better route. I think we could actually customize or even add new LLB ops (not just sources) now that we have our own worker too (just need to handle calling it in [ResolveOp](https://github.com/sipsma/dagger/blob/064c414e4671bbf797bf748b3782e1096fb9f150/engine/buildkit/worker.go#L119-L119)).
>
> Those options give you access to raw snapshots/mounts so you should be able to do whatever you want

PR #9102, opened 2024-12-04
docs: Move installation topics from FAQ and op manual

vikram-dagger:


docs/current_docs/install.mdx
```diff
@@ -7,3 +7,104 @@ import PartialInstallCli from './partials/_install-cli.mdx';
...
+Homebrew users can use the following commands:
+
+```shell
+brew update
+brew upgrade dagger
+```
+
+If you do not have Homebrew installed, you can use [install.sh](https://github.com/dagger/dagger/blob/main/install.sh):
+
+<CodeBlock language="shell">
```
jedevc, line 25: This needs importing (this is causing the docs build to fail).

docs/current_docs/install.mdx
```diff
@@ -7,3 +7,104 @@ import PartialInstallCli from './partials/_install-cli.mdx';
 # Installation
 
 <PartialInstallCli />
+
+## Update
```
jedevc, line 11: We should make some note that you should use the same method that you installed the CLI with in the first place.

docs/current_docs/install.mdx
```diff
@@ -7,3 +7,104 @@ import PartialInstallCli from './partials/_install-cli.mdx';
...
+  - Go (SDK and CLI): [https://github.com/adrg/xdg](https://github.com/adrg/xdg)
+  - Node.js: [https://github.com/sindresorhus/env-paths](https://github.com/sindresorhus/env-paths)
+  - Python: [https://github.com/platformdirs/platformdirs](https://github.com/platformdirs/platformdirs)
+:::
+
+## Versioning
+
+The Dagger CLI is released in tandem with the Dagger Engine and thus shares a version number with it. Dagger SDKs automatically provision a Dagger Engine at a compatible version.
+
+The Dagger Engine runner is distributed as a container image at `registry.dagger.io/engine`. Tags are made for the version of each release. For example, the `0.14.0` release has a corresponding image at `registry.dagger.io/engine:v0.14.0`.
```
jedevc, line 104: Can we use the `daggerVersion` partial to keep this up-to-date?

docs/current_docs/install.mdx
```diff
@@ -7,3 +7,104 @@ import PartialInstallCli from './partials/_install-cli.mdx';
...
+
+Each release notes the compatible Dagger Engine version in its release notes. If running from the CLI, or providing a CLI for the SDK to use via the `_EXPERIMENTAL_DAGGER_CLI_BIN` variable, check the release notes of the SDK, which indicate the required CLI and Engine versions.
```
jedevc, line 106: I don't think we document `_EXPERIMENTAL_DAGGER_CLI_BIN` *anywhere* in our docs properly (except for here).

I think we either:
- Remove this entirely, it's a weird fringe use case anyways (my preference :laughing:)
- Explain how it works somewhere

vikram-dagger: Removed

docs/current_docs/install.mdx
vikram-dagger: I tried to do this as in the install partial but it doesn't seem to work, probably because the syntax I'm using is wrong? See https://deploy-preview-9102--devel-docs-dagger-io.netlify.app/install

docs/current_docs/install.mdx
vikram-dagger: Done, and then removed the duplicate content.

docs/current_docs/install.mdx
vikram-dagger: Added

docs/current_docs/install.mdx
vikram-dagger: Ignore, fixed now

docs/current_docs/install.mdx
```diff
@@ -2,8 +2,80 @@
...
+  - Go (SDK and CLI): [https://github.com/adrg/xdg](https://github.com/adrg/xdg)
+  - Node.js: [https://github.com/sindresorhus/env-paths](https://github.com/sindresorhus/env-paths)
+  - Python: [https://github.com/platformdirs/platformdirs](https://github.com/platformdirs/platformdirs)
+:::
+
+## Versioning
+
+The Dagger CLI is released in tandem with the Dagger Engine and thus shares a version number with it. Dagger SDKs automatically provision a Dagger Engine at a compatible version.
+
+The Dagger Engine runner is distributed as a container image at `registry.dagger.io/engine`. Tags are made for the version of each release. For example, the {daggerVersion} release has a corresponding image at <code>registry.dagger.io/engine:v{daggerVersion}</code>.
```
jedevc, line 79: Hm, something still seems wrong here :thinking: 
image

I *think* we might need `<code>{`registry.dagger.io/engine:v${daggerVersion}`}</code>

vikram-dagger: Yes, I came to the same solution after various attempts :)

jedevc (approved):

PR #9103, opened 2024-12-04
ci: fix distro to be a real enum

jedevc:
`Distro` wasn't a proper enum type :( I got the syntax wrong when adding it.

The enum members need to be *explicitly* declared to be of the distro type to apply.


gerhard (approved):
👍 🚀

PR #9106, opened 2024-12-04
ci: cross-compile dumb-init

jedevc:
Similar in theme to #9036, but this time to fix this very annoying segfault flake: https://github.com/dagger/dagger/actions/runs/12163689482/job/33923486378

```
cc -std=gnu99 -static -s -Wall -Werror -O3 -o dumb-init dumb-init.c
Stderr:
cc: internal compiler error: Segmentation fault signal terminated program cc1
Please submit a full bug report, with preprocessed source (by using -freport-bug).
See <https://gitlab.alpinelinux.org/alpine/aports/-/issues> for instructions.
make: *** [Makefile:11: build] Error 4
```

If we guess that this is also an annoying QEMU issue, then this can be avoided by just using `xx` here as well. `xx` depends on `clang`, so I install that as well.

I also took size samples before:
``` 
# ls -lh $(which dumb-init)
-rwxr-xr-x    1 root     root       57.7K Dec  2 11:30 /usr/local/bin/dumb-init
```
And after:
```
# ls -lh $(which dumb-init)
-rwxr-xr-x    1 root     root       57.6K Dec  4 16:49 /usr/local/bin/dumb-init
```

### DIFF
--- a/.dagger/build/builder.go
+++ b/.dagger/build/builder.go
@@ -436,19 +436,17 @@ func (build *Builder) cniPlugins() []*dagger.File {
 func (build *Builder) dumbInit() *dagger.File {
 	// dumb init is static, so we can use it on any base image
 	return dag.
-		Alpine(dagger.AlpineOpts{
-			Branch: consts.AlpineVersion,
-			Packages: []string{
-				"bash",
-				"build-base",
-				"clang",
-			},
-		}).
 		Container().
+		From(consts.AlpineImage).
 		WithDirectory("/", dag.Container().From(consts.XxImage).Rootfs()).
+		WithEnvVariable("BUILDPLATFORM", "linux/"+runtime.GOARCH).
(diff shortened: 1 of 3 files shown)


.dagger/build/builder.go
```diff
@@ -436,17 +436,16 @@ func (build *Builder) cniPlugins() []*dagger.File {
...
-		Alpine(dagger.AlpineOpts{
-			Branch: consts.AlpineVersion,
-			Packages: []string{
-				"bash",
-				"build-base",
-			},
-			Arch: build.platformSpec.Architecture,
-		}).
 		Container().
+		From(consts.AlpineImage).
```
jedevc, line 440: Annoyingly, we need to go back to relying on this instead of using our `alpine` module, since the database isn't in a valid state after we finish with it.

Working on that here: https://github.com/dagger/dagger/pull/9117

jedevc:
Superseded by https://github.com/dagger/dagger/pull/9262.

PR #9111, opened 2024-12-04
enable more integration tests to run on raw macOS

cwlbraa:
for profiling and fast dev loops, it's often useful to run tests on your host machine aimed at a `./hack/dev`-built engine. 

```
./hack/dev go test -count=1 -v -run TestModule/TestLots ./core/integration
```

on MacOS, many tests won't run locally because we often upload a local dagger CLI binary as part of the test. 

this PR teaches dev-export to copy the engine container's linux binary to ./bin/dagger-linux, then wires that path out through mage/engine.go so it can be consumed by ./core/integration/suite_test.go as the desired binary to upload to test-created containers.

this _does not_ make all or nearly all the tests pass successfully when running on macOS. many tests still rely on running inside of dagger, as @vito noted in `./hack/most-tests`. re-cataloguing which tests those are seems host-machine-dependent, so I don't bother here.

some of those tests are fairly obvious, like when the tests attempt to push or pull from registry:5000. others are far less obvious.

### DIFF
--- a/.dagger/main.go
+++ b/.dagger/main.go
@@ -300,13 +300,19 @@ func (dev *DaggerDev) DevExport(
 	})
 
 	// FIXME: get path from the cli file (windows is already handled)
-	cliPath := "dagger"
+	hostCliPath := "dagger"
 	if platformSpec.OS == "windows" {
-		cliPath += ".exe"
+		hostCliPath += ".exe"
 	}
 	dir := dag.Directory().
 		WithFile("engine.tar", engineTar).
-		WithFile(cliPath, dag.DaggerCli().Binary(dagger.DaggerCliBinaryOpts{Platform: platform}))
+		WithFile(hostCliPath, dag.DaggerCli().Binary(dagger.DaggerCliBinaryOpts{Platform: platform}))
+
+	// this allows our integration tests to plumb built cli binaries into containers when the host OS doesn't match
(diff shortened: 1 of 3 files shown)


.dagger/main.go
```diff
@@ -305,13 +305,19 @@ func (dev *DaggerDev) DevExport(
...
 	dir := dag.Directory().
 		WithFile("engine.tar", engineTar).
-		WithFile(cliPath, dag.DaggerCli().Binary(dagger.DaggerCliBinaryOpts{Platform: platform}))
+		WithFile(hostCliPath, dag.DaggerCli().Binary(dagger.DaggerCliBinaryOpts{Platform: platform}))
+
+	// this allows our integration tests to plumb built cli binaries into containers when the host OS doesn't match
+	if platformSpec.OS != "linux" {
+		linuxCliPath := "dagger-linux"
+		dir = dir.WithFile(linuxCliPath, engineCtr.File(cliPath))
+	}
```
cwlbraa, line 320: @jedevc will this have any negative side-effects on the release process? it adds a single extra binary, dagger-linux, to dev-export's output dir.

jedevc: Nope, you're all good! `DevExport` is only called by that mage call, which is only called by `./hack/dev`.

jedevc (approved):

PR #9113, opened 2024-12-05
docs: Update code  snippets for services API changes

vikram-dagger:
Based on https://github.com/dagger/dagger/pull/8865

### DIFF
--- a/docs/current_docs/api/snippets/services/bind-services/go/main.go
+++ b/docs/current_docs/api/snippets/services/bind-services/go/main.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"context"
-	"main/internal/dagger"
+	"dagger/my-module/internal/dagger"
 )
 
 type MyModule struct{}
--- a/docs/current_docs/api/snippets/services/bind-services/go/main.go
+++ b/docs/current_docs/api/snippets/services/bind-services/go/main.go
@@ -13,9 +13,8 @@ func (m *MyModule) HttpService() *dagger.Service {
 		From("python").
 		WithWorkdir("/srv").
 		WithNewFile("index.html", "Hello, world!").
-		WithDefaultArgs([]string{"python", "-m", "http.server", "8080"}).
 		WithExposedPort(8080).
-		AsService()
+		AsService(dagger.ContainerAsServiceOpts{Args: []string{"python", "-m", "http.server", "8080"}})
 }
 
 // Send a request to an HTTP service and return the response
--- a/docs/current_docs/features/snippets/services-2/go/main.go
+++ b/docs/current_docs/features/snippets/services-2/go/main.go
@@ -3,8 +3,6 @@ package main
 import (
 	"context"
 	"dagger/my-module/internal/dagger"
-
-	"dagger.io/dagger/dag"
 )
 
 type MyModule struct{}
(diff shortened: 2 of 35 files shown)


vikram-dagger:
@marcosnils @rajatjindal I found that 
- all the examples which used `AsService` with `WithExec` worked fine, they did not need changes. However, I have changed them to `WithDefaultArgs` (and they still work fine).
- all the examples which used `AsService` without `WithExec` worked fine even after the API change. I have not made any changes to set the `useEntrypoint` option.

I am not sure if the above is expected behaviour? Let me know if further changes are needed.

rajatjindal:
> @marcosnils @rajatjindal I found that
> 
> * all the examples which used `AsService` with `WithExec` worked fine, they did not need changes. However, I have changed them to `WithDefaultArgs` (and they still work fine).
> * all the examples which used `AsService` without `WithExec` worked fine even after the API change. I have not made any changes to set the `useEntrypoint` option.
> 
> This is using a fresh build from #8865
> 
> ```
> $ dagger version
> dagger v0.15.0-241205104731-93237db542e9 (registry.dagger.io/engine:6152c829bf7bdd053025c1821d19c439d4ed3a39) linux/amd64
> ```
> 
> I am not sure if the above is expected behaviour? Let me know if further changes are needed. The current examples for services are all in https://docs.dagger.io/api/services

I discussed this with @vikram-dagger over chat and also confirmed on my local setup that the version in dagger.json should be `v0.15.0` to reflect the new api changes. Vikram is going to try and confirm that as well. 

thanks

docs/current_docs/api/snippets/services/bind-services/go/main.go
```diff
@@ -13,7 +13,7 @@ func (m *MyModule) HttpService() *dagger.Service {
 		From("python").
 		WithWorkdir("/srv").
 		WithNewFile("index.html", "Hello, world!").
-		WithExec([]string{"python", "-m", "http.server", "8080"}).
+		WithDefaultArgs([]string{"python", "-m", "http.server", "8080"}).
```
helderco, line 16: If it's just for running the service I'd advise using the optional args in `AsService` rather than setting `WithDefaultArgs` and/or `WithEntrypoint`. These should be mostly used for affecting the metadata of a container image to be published.

vikram-dagger: Done

docs/current_docs/api/snippets/services/bind-services/go/main.go
```diff
@@ -2,7 +2,7 @@ package main
 
 import (
 	"context"
-	"main/internal/dagger"
+	"dagger/my-module/internal/dagger"
```
jedevc, line 5: :thinking: The `go.mod` for this still puts this as `module main` - so the previous change is actually the right one.

But also I guess this comes back to a prior discussion about the runability of these examples as modules (vs just as code snippets). So this isn't blocking.

docs/current_docs/features/snippets/services-2/go/main.go
```diff
@@ -3,8 +3,6 @@ package main
 import (
 	"context"
 	"dagger/my-module/internal/dagger"
-
-	"dagger.io/dagger/dag"
```
jedevc, line 7: Oh woops - nice find. Actually, none of our code snippets should contain this, we should remove it everywhere.

jedevc (approved):

docs/current_docs/features/snippets/services-2/go/main.go
jedevc: Pushed a commit with a few more of these I found.

jedevc (approved):

PR #9114, opened 2024-12-05
docs: Update deps page for dagger uninstall command

vikram-dagger:
Based on https://github.com/dagger/dagger/pull/8745

### DIFF
--- a/docs/current_docs/api/module-dependencies.mdx
+++ b/docs/current_docs/api/module-dependencies.mdx
@@ -6,6 +6,8 @@ import TabItem from '@theme/TabItem';
 
 # Module Dependencies
 
+## Installation
+
 You can call Dagger Functions from any other Dagger module in your own Dagger module simply by adding it as a module dependency with `dagger install`, as in the following example:
 
 ```sh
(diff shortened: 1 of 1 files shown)


vikram-dagger:
This does not (yet) include https://github.com/dagger/dagger/pull/8839 as it seems to be under discussion at the time of writing.

docs/current_docs/api/module-dependencies.mdx
```diff
@@ -107,3 +109,39 @@ dagger install ./path/to/module
 :::note
 Installing a module using a local path (relative or absolute) is only possible if your module is within the repository root (for Git repositories) or the directory containing the `dagger.json` file (for all other cases).
 :::
+
+## Uninstallation
+
+To remove a dependency from your Dagger module, use the `dagger uninstall` command, as in the following example:
+
+```sh
+dagger uninstall github.com/shykes/daggerverse/hello
```
jedevc, line 118: I think *probably* we should suggest `dagger uninstall hello` and suggest uninstalling by name first?

Even though we have compatibility with `source` everywhere, I think looking at dependencies by name is shorter and cleaner.

jedevc (approved):

docs/current_docs/api/module-dependencies.mdx
helderco: Agree!

docs/current_docs/api/module-dependencies.mdx
```diff
@@ -107,3 +109,39 @@ dagger install ./path/to/module
...
+```
+
+The `dagger uninstall` command can be passed either a remote repository reference or a local module name. If a version number is specified, only the specified version of the dependency is uninstalled.
```
helderco, line 121: I don't quite follow the use case here. A dependency only has one version. When would you want to uninstall a specific version? Even if someone needs to uninstall that specific version in a script or something, but otherwise leave it be, seems to me this would require manual intervention. I may not be seeing the right use case, but it seems edge casey enough that I'd suggest simplifying this here by omitting this feature. Just uninstall by name.

vikram-dagger: Updated

docs/current_docs/api/module-dependencies.mdx
vikram-dagger: Updated

PR #9115, opened 2024-12-05
docs: Update API docs for TypeScript SDK improvements

vikram-dagger:


docs/current_docs/configuration/modules.mdx
```diff
@@ -150,3 +150,19 @@ When a package manager is not explicitly defined within the `package.json` file,
 :::warning
 This behavior however should be considered a sensible fallback, and not as an explicit configuration. Since this default can change, we encourage you to configure a package manager explicitly.
 :::
+
+### Alternative base images
```
vikram-dagger, line 154: @TomChv what is the relation between this option and the `dagger.runtime` option documented above in this same page?

docs/current_docs/api/module-structure.mdx
```diff
@@ -267,7 +267,7 @@ This can be useful to add a few requirements to the module's execution environme
 </TabItem>
 <TabItem value="TypeScript">
 
-The runtime container is currently hardcoded to run in Node.js 21.3 (although this may be configurable in future).
+The runtime container is currently hardcoded to run in Node.js 22.11.0 (although this may be configurable in future).
```
helderco, line 270: It's already configurable so this is outdated.

docs/current_docs/configuration/modules.mdx
```diff
@@ -150,3 +150,19 @@ When a package manager is not explicitly defined within the `package.json` file,
...
+
+You can specify a custom base image in your module's `package.json`, tailoring the SDK to your project's needs. To change the base image, set the field `dagger.baseImage` in your Dagger module's `package.json` file.
+
```
helderco, line 157: This should be considered an escape hatch for advanced cases and come with a warning to "use at own risk". It's the only reason we're doing this, especially for users in private settings. Otherwise giving this much power to users could be a nightmare to support.

Most common use case would probably be adding some needed environment variables to the runtime container, adding a few files for authentication, or even needed operating system packages, for example. Users should be adivsed not to deviate from the default image too much, at risk of things breaking.

Ideally in the future, adding some features to the engine, would deprecate this. Python was the first to enable customizing the base image, but that came before the engine was even able to mount custom certificates to any container. Thus for that use case, it's no longer necessary to resort to this escape hatch.

As for adding OS packages, there's currently no other alternative, but the point is, ideally there will be. So again, best treat this as an escape hatch for advanced needs.

vikram-dagger: Added

docs/current_docs/api/module-structure.mdx
vikram-dagger: Fixed

docs/current_docs/configuration/modules.mdx
jedevc: `dagger.runtime` sets the js runtime - like `node` or `bun` or maybe one day `deno`.

The image to use is *derived* from the version in the `dagger.runtime` if it's present - `baseImage` allows specifying the exact image.

docs/current_docs/configuration/modules.mdx
```diff
@@ -150,3 +150,25 @@ When a package manager is not explicitly defined within the `package.json` file,
...
+:::warning
+It is recommended to use this feature only for advanced use cases such as adding bespoke environment variables, authentication files or operating system packages to the runtime container. Ensure that you avoid deviating from the default image too much, as doing so could create unexpected results.
+:::
+
+To change the base image, set the field `dagger.baseImage` in your Dagger module's `package.json` file.
+
+```json
+{
+  "dagger": {
+    "baseImage": "node:23.2.0-alpine@sha256:ecefaffd4706..."
```
jedevc, line 167: We should give the full image sha (so it's copy-pasteable):

```suggestion
    "baseImage": "node:23.2.0-alpine@sha256:ecefaffd4706c5879af52e022fdb8ea30cbd6590e2a30d05347790d690727c6c"
```

jedevc (approved):

docs/current_docs/configuration/modules.mdx
vikram-dagger: Done

docs/current_docs/configuration/modules.mdx
vikram-dagger: Added

PR #9117, opened 2024-12-05
ci: alpine module should create a package database

jedevc:
Follow up from https://github.com/dagger/dagger/pull/8714.

This allows running apk commands in the resulting container.

### DIFF
--- a/modules/alpine/main.go
+++ b/modules/alpine/main.go
@@ -207,29 +207,31 @@ func (m *Alpine) withPkgs(
 				WithMountedFile(mntPath, dag.HTTP(url)).
 				WithMountedDirectory(outDir, dag.Directory()).
 				WithWorkdir(outDir).
-				WithExec([]string{"tar", "-xf", mntPath})
+				WithExec([]string{"tar", "-xvf", mntPath})
+
+			tarOut, err := unpacked.Stdout(ctx)
+			if err != nil {
+				return err
+			}
+			entries := strings.Split(strings.TrimSpace(tarOut), "\n")
 
 			alpinePkg := &apkPkg{
 				name: pkg.PackageName(),
 				dir:  unpacked.Directory(outDir),
 			}
 
-			entries, err := alpinePkg.dir.Entries(ctx)
-			if err != nil {
-				return fmt.Errorf("failed to get alpine package entries: %w", err)
-			}
-
 			// HACK: /lib64 is a link, so don't overwrite it
 			// - wolfi-baselayout links /lib64 -> /lib
 			// - ld-linux installs to /lib64
-			// TODO: this should *probably* apply to /usr/lib64/ and
-			// /usr/local/lib64/ as well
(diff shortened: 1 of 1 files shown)


modules/alpine/main.go
```diff
@@ -136,26 +137,47 @@ func (m *Alpine) Container(ctx context.Context) (*dagger.Container, error) {
...
+		key, err := url.PathUnescape(key)
+		if err != nil {
+			return nil, err
+		}
+		ctr = ctr.WithNewFile(filepath.Join("/etc/apk/keys/", key), string(keydata))
+	}
+
+	// create the package database
+	// TODO: use logic from APK.AddInstalledPackage instead (which handles
+	// installed files as well)
```
jedevc, line 170: So I suspect this is actually very important to include :laughing: 

With this package database as is, this command `SIGSEGV`s (it seems to occur when you use `apk` to install a package that has been installed by the module):

```
$ dagger call -m modules/alpine --branch edge --packages git container with-exec --args apk,add,git
```

jedevc: Ok, so we want to call `AddInstalledPackage`. But, this operates directly on an `fs.FS` - which makes sense.

The problem is, we need to get this into dagger somehow - and we want to do it in a cache efficient way, where if the `.apk` file (and the signing key) hasn't changed, we can reuse the extraction result.

But - I can't quite work out how we can do this. It feels *somewhat* similar to what we need in https://github.com/dagger/dagger/pull/9098#discussion_r1868220331, although, the module equivalent.

*Ideally*, it feels like we would want:
- An `unpackPackage` function in this module - it takes a `*dagger.File` and returns a `*dagger.Directory` 
- It extracts the `*dagger.File` to the module fs, unpacks it using the go apko library, and then uploads the `*dagger.Directory` to the engine (can't remember if we can do that today)
- To unpack the `.apk` file, instead of using `tar` in a `WithExec`, we would do a self-call #8030
- This would then cache, assuming we had #7428

But, without global function caching or self calls quite implemented yet, I'm not sure what this would even look like.
- We could work around no self-calls in a few different ways (we could just use our graphql api directly)
- Without global function caching though, we'll be uploading lots more content over-and-over.

sipsma: Yeah agree that the approach you outlined would make the most sense, but obviously has quite a few pre-reqs...

I think you could do something dumb like compile a custom go binary that does what we need, run it in `WithExec`s and output it through a combination of stdout/mount outputs. But that's essentially implementing a highly specialized version of modules in userspace 😅

We should obviously do all that (function caching + self calls), but in the meantime the most immediate motivation is from trying to cross-compile dumb-init w/ xx-apk, right? We could probably do that differently:
* Use some different cross-compilation tool chain
* Just replace dumb-init with our own go binary. Even in C it's very few lines of code and we also don't even use most of the features there: https://github.com/Yelp/dumb-init/blob/master/dumb-init.c so I wouldn't be surprised if a go implementation was extremely short and low-maintainence

PR #9118, opened 2024-12-05
feat: directories in entries and glob have trailing slash

jedevc:
> [!NOTE]
>
> This is a proposal! Thoughts wanted :thinking: 
>
> We *might* consider this a breaking change, so should probably put it behind our module compat if we choose to take it.

This is fairly useful, and matches the output of GNU's `tar xvf` - it makes it a lot easier to inspect and traverse a filesystem using this.

This came up while working with a filesystem, and wanting to filter out the results of `Glob` to *only* the directories, or to *only* the leaf-level files.

### DIFF
--- a/version/git.go
+++ b/version/git.go
@@ -39,7 +39,7 @@ func git(ctx context.Context, gitDir *dagger.Directory, dir *dagger.Directory) (
 	if err != nil {
 		return nil, err
 	}
-	if slices.Contains(entries, ".git") {
+	if slices.Contains(entries, ".git/") || slices.Contains(entries, ".git") {
 		valid = true
 	}
 
--- a/core/directory.go
+++ b/core/directory.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io/fs"
+	"os"
 	"path"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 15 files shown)


gdevillele:
Thank you!
I too, would like to be able to identify directories while walking a dagger file hierarchy. 👍

vito:
Makes sense to me, too.

[In Bass](https://bass-lang.org/bassics.html#paths), `./foo` is a file path, and `./foo/` is a directory path. 😄

jedevc:
Nice, okay, I'm gonna try and revive this for v0.17.0. Because it's a change with potentially unknown breaking implications, I'm gonna try and get it in for that release.

core/schema/directory.go
```diff
@@ -188,7 +189,15 @@ func (s *directorySchema) withTimestamps(ctx context.Context, parent *core.Direc
 }
 
 func (s *directorySchema) name(ctx context.Context, parent *core.Directory, args struct{}) (dagql.String, error) {
```
helderco, line 191: Doesn't this always refer to a directory? Not sure why add the slash here.

version/git.go
```diff
@@ -39,7 +39,7 @@ func git(ctx context.Context, gitDir *dagger.Directory, dir *dagger.Directory) (
 	if err != nil {
 		return nil, err
 	}
-	if slices.Contains(entries, ".git") {
+	if slices.Contains(entries, ".git/") || slices.Contains(entries, ".git") {
```
helderco, line 42: When does `.git` not have a `/`?

helderco (approved):
Overall lgtm 👍, just have a couple questions.

version/git.go
jedevc: Ah, because this module needs to work on both a dev version of dagger, and a stable version. Because there's version compat, it'll technically keep working on the dev build without the slash until we upgrade, but I wanted to add this in now, so I don't forget it later.

I've added a comment though, and will remove it during the next improve-releasing PR.

PR #9134, opened 2024-12-07
Follow-up from changes to `DAGGER_LEAVE_OLD_ENGINE`

jedevc:
Follow-up to #9130.

There's a few issues in the merged PR:
- ~~Committed a built binary~~ (fixed by rewriting git history)
- Hardcoded string in too many places
- Overly complex logic
- Unnecessary logs
- Manual string comparisons instead of `strconv.ParseBool`
- No tests

I've fixed all these, and simplified the code. I've also renamed the variable to `_EXPERIMENTAL_DAGGER_LEAVE_OLD_ENGINE`, since I think this is more inline with other variables that we use in dagger. But happy to leave it as is.

### DIFF
--- a/.github/workflows/engine-and-cli.yml
+++ b/.github/workflows/engine-and-cli.yml
@@ -113,7 +113,7 @@ jobs:
       - name: "test"
         uses: ./.github/actions/call
         with:
-          function: "test specific --run='TestProvision' --race=true --parallel=16"
+          function: "test specific --run='TestProvision' --race=true --parallel=1"
           upload-logs: true
 
   test-everything-else:
--- a/core/integration/container_test.go
+++ b/core/integration/container_test.go
@@ -3632,8 +3632,8 @@ func (ContainerSuite) TestInsecureRootCapabilitesWithService(ctx context.Context
 	// testing it can startup, create containers and bind mount from its filesystem to
 	// them.
 	randID := identity.NewID()
-	dockerd := dockerService(t, c, "23.0.1", middleware)
-	out, err := dockerClient(ctx, t, c, dockerd, "23.0.1", middleware).
+	dockerc := dockerSetup(ctx, t, t.Name(), c, "23.0.1", middleware)
+	out, err := dockerc.
 		WithExec([]string{"sh", "-e", "-c", strings.Join([]string{
 			fmt.Sprintf("echo %s-from-outside > /tmp/from-outside", randID),
 			"docker run --rm -v /tmp:/tmp alpine cat /tmp/from-outside",
@@ -4213,7 +4213,7 @@ func (ContainerSuite) TestImageLoadCompatibility(ctx context.Context, t *testctx
 	c := connect(ctx, t)
 
 	for _, dockerVersion := range []string{"20.10", "23.0", "24.0"} {
-		dockerd := dockerService(t, c, dockerVersion, nil)
+		dockerc := dockerSetup(ctx, t, t.Name(), c, dockerVersion, nil)
 
 		for _, mediaType := range []dagger.ImageMediaTypes{dagger.ImageMediaTypesOcimediaTypes, dagger.ImageMediaTypesDockerMediaTypes} {
 			mediaType := mediaType
@@ -4231,7 +4231,7 @@ func (ContainerSuite) TestImageLoadCompatibility(ctx context.Context, t *testctx
 						})
 					require.NoError(t, err)
 
-					ctr := dockerClient(ctx, t, c, dockerd, dockerVersion, nil).
+					ctr := dockerc.
 						WithMountedFile(path.Join("/", path.Base(tmpfile)), c.Host().File(tmpfile)).
 						WithExec([]string{"docker", "load", "-i", "/" + path.Base(tmpfile)})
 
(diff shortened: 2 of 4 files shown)


gerhard:
https://github.com/dagger/dagger/pull/9130 should not have been merged for all the reasons captured at the top of this PR.

Even if we merge this follow-up PR which addresses these issues, we will still have a 42MB binary in our git history which will require a history rewrite to clean up. This is a lot - as in hours - of follow-up work that we should not be doing.

My proposal for course-correcting:
1. All AI generated code needs to be reviewed by at least **two distinct humans** before it can be merged
2. AI code must include tests. If the functionality is complex and tests are hard, we should not accept the code changes.

Do you agree with the above @shykes?

shykes:
@gerhard the PR was reviewed by two humans.

shykes:
I think rather than focusing on the AI aspect, we should focus on the "commit a big binary" aspect, which is a mistake a human could have easily made. Maybe we can catch this in a distinct check, similar to DCO?

shykes:
@jedevc let's keep the env variable without `EXPERIMENTAL`, it's pretty safe behavior and doesn't expose an unstable API or anything like that. We don't need to prefix all our env vars with "experimental" necessarily, even though we've used it a lot.

gerhard:
> @gerhard the PR was reviewed by two humans.

To be explicitly clear, we should require **two approvals** from humans. A human review - e.g. a comment - doesn't mean approval.

> I think rather than focusing on the AI aspect, we should focus on the "commit a big binary" aspect, which is a mistake a human could have easily made. Maybe we can catch this in a distinct check, similar to DCO?

This change should have been a few lines of code changes, **plus** tests. Two human software engineers should have been involved in the process.

As the code was created by an AI, two human software engineers should have reviewed and **both** approved this code.

That is the point that I'm making.

gerhard:
> @jedevc let's keep the env variable without EXPERIMENTAL, it's pretty safe behavior and doesn't expose an unstable API or anything like that. We don't need to prefix all our env vars with "experimental" necessarily, even though we've used it a lot.

In order to signal new behaviour to users, we introduce new environment variables prefixed with `_EXPERIMENTAL`.

We are now saying that it's OK to make the `_EXPERIMENTAL` prefix for new environment variables an in-the-moment judgement call. I disagree with this approach.

shykes:
> > @jedevc let's keep the env variable without EXPERIMENTAL, it's pretty safe behavior and doesn't expose an unstable API or anything like that. We don't need to prefix all our env vars with "experimental" necessarily, even though we've used it a lot.
> 
> In order to signal new behaviour to users, we introduce new environment variables prefixed with `_EXPERIMENTAL`.

Not systematically, no.

> We are now saying that it's OK to make the `_EXPERIMENTAL` prefix for new environment variables an in-the-moment judgement call.

I call those "design decisions", but yes. Some settings should be marked experimental, because they configure a feature that is itself experimental, or we're not sure about the best way to design the setting. In other cases (like this one), the feature is mature, and the mechanism for configuring it is straightforward. So it doesn't make sense to add the extra complication of `_EXPERIMENTAL`.

 I personally never said otherwise, and I don't remember us ever discussing it. So I guess we've had an ongoing misunderstanding on how those decisions are made.

> I disagree with this approach.

That's fine, you're generally on the more conservative side of change management discussions, and it's a valuable perspective. In this case it just doesn't make sense to have a blanket rule that every env variable always has to go through a phase of being prefixed by `_EXPERIMENTAL_`, no matter what.

shykes:
> We should generally have a conversation around all of the environment variables. We have a _total_ mess right now and we're almost entirely internally inconsistent as to what goes where. I was hoping to have this be `_EXPERIMENTAL` as well, so we can make one giant change in the future that resorts and reorganizes all of them - moving some into the new `engine.json`, maybe merging some together, etc.
> 
> However, I don't really mind about about this specific one (though, it's frustrating to have this discussion _after_ the change has been merged so close to the v0.15.0 deadline) - but by introducing it as not `_EXPERIMENTAL`, I think we're taking on a requirement to avoid breaking it. This is fine - but potentially will introduce technical debt if we decide to change approaches later. See above - I think this actually makes more sense in the `engine.json`, instead of continuing to add to the complexity of our user's environments.
> 
> @shykes if this isn't an experimental change, can you please help put together some docs on this?

Sure happy to do that. Where do we usually document CLI configuration?

shykes:
Re moving configurations to `engine.json`: note that this is not an engine setting, it's on the CLI side...

jedevc:
> Re moving configurations to `engine.json`: note that this is not an engine setting, it's on the CLI side...

Right sure, into whatever config file we choose. However, this does relate to engine provisioning, so I'd argue that `engine.json` is appropriate.

jedevc:
Cli docs related to provisioning can be found around our `custom-runner.mdx` for now. But this maybe isn't the best place for this. I'm not 100% sure - cc @vikram-dagger.

.github/workflows/engine-and-cli.yml
```diff
@@ -113,7 +113,7 @@ jobs:
       - name: "test"
         uses: ./.github/actions/call
         with:
-          function: "test specific --run='TestProvision' --race=true --parallel=16"
+          function: "test specific --run='TestProvision' --race=true --parallel=1"
```
gerhard, line 116: Another reminder to self: deploy new CI runner with cgroups v2 support.

gerhard (approved):
The fixups look good to me.

Tests passed locally in `4m` (first run), then `2m` and `2m` again.

I also confirmed that setting `DAGGER_LEAVE_OLD_ENGINE=true` behaves correctly.

PR #9161, opened 2024-12-10
Update v0.15.0 notes

jedevc:


modules/gha/steps.go
```diff
@@ -111,17 +111,6 @@ func (j *Job) callDaggerStep() api.JobStep {
...
-		if strings.HasPrefix(key, "GITHUB_") {
-			// Inject Github context keys
-			// github.ref becomes $GITHUB_REF, etc.
-			env[key] = fmt.Sprintf("${{ github.%s }}", strings.ToLower(key))
-		} else if strings.HasPrefix(key, "RUNNER_") {
-			// Inject Runner context keys
-			// runner.ref becomes $RUNNER_REF, etc.
-			env[key] = fmt.Sprintf("${{ runner.%s }}", strings.ToLower(key))
-		}
-	}
```
jedevc, line 124: Not sure why any of this is needed - it's incorrect, since it tries to get `${{ github.github_ref }}` instead of `${{ github.ref }}`.

Also this info is all available without this block https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables. Github sets these itself without any need to do it manually.

modules/gha/workflow.go
```diff
@@ -180,7 +180,7 @@ func (gha *Gha) Workflow(
 		w = w.onPullRequest(nil, onPullRequestBranches, nil)
 	}
 	if onPullRequestPaths != nil {
-		w = w.onPullRequest([]string{"paths"}, nil, onPullRequestPaths)
+		w = w.onPullRequest(nil, nil, onPullRequestPaths)
```
jedevc, line 183: `paths` isn't actually a valid event. See https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request.

gerhard: This is valid: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-files-changed-in-a-pull-request

image

modules/gha/steps.go
khris-zeroeyes: I suspect that the code is correct, but the comments are wrong. While GitHub does set those variables, injecting them into the container is a different story. Having those environment variables accessible in the context of the dagger release may be unnecessary, however, there are a couple examples provided to demonstrate how they could be used in different contexts:

- https://github.com/dagger/dagger/blob/8b9fbf5c643285eba1887292a79050540e399955/modules/gha/examples/go/main.go#L47
- https://github.com/dagger/dagger/blob/8b9fbf5c643285eba1887292a79050540e399955/modules/gha/examples/go/main.go#L58

modules/gha/workflow.go
jedevc: Sorry, it's valid as a filter yes, but not valid as a `types` value.

modules/gha/steps.go
jedevc: :thinking: what? These aren't actually set in any container.

This map is exclusively used as the `env` key for the call dagger step.

---

That said. Those tests are also wrong. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

`github.repository_url` or `github.github_repository_url` isn't a real environment variable here. `github.repositoryUrl` is - but the code here won't ever generate that.

khris-zeroeyes: ACK. I see what you mean.

modules/gha/examples/go/main.go
```diff
@@ -44,7 +44,7 @@ func (m *Examples) GhaGithubContext() *dagger.Directory {
 			dag.Gha().Workflow("lint all branches", dagger.GhaWorkflowOpts{
 				OnPush: true,
 			}).
-				WithJob(dag.Gha().Job("lin", "lint --source=${GITHUB_REPOSITORY_URL}#${GITHUB_REF}")),
+				WithJob(dag.Gha().Job("lin", "lint --source=${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}#${GITHUB_REF}")),
```
gerhard, line 47: `lin` is a typo (same as before this change), not blocking, something to come back.

gerhard (approved):
LGTM! A single check was failing, but that seemed like a transient issue.

FTR: https://dagger.cloud/dagger/traces/ec2a500c9d78fa489c235c1afc9ca3ab#L4-d210b79ce5893fbf

modules/gha/examples/go/main.go
jedevc: Oops, missed this somehow - fixed it while in the area.

PR #9166, opened 2024-12-11
chore: update release date for v0.15.0

jedevc:
This didn't happen yesterday, it'll happen today!


jedevc:
Bleh, daggerverse preview failing - this is known, it needs to be done from an upstream branch - however, it already passed before, so it's not an issue.

helderco (approved):

PR #9167, opened 2024-12-11
Improve releasing during v0.15.0

jedevc:
TODO:
- [x] Fix helm publish - https://github.com/dagger/dagger/actions/runs/12275890055/job/34252281126
- [x] Fix automated daggerverse bump - https://github.com/dagger/dagger/actions/runs/12275890055/job/34252280776
- [ ] Fix error detection which breaks in v0.15.0 - (due to changes to how `ExecError` works) https://github.com/dagger/dagger/blob/ed73cf2927373c88f478e292d477883741605d54/modules/daggerverse/main.go#L102-L107 
  - [x] https://github.com/sagikazarmark/daggerverse/pull/202
- [ ] Correctly assign correct github user in https://github.com/dagger/dagger/blob/ed73cf2927373c88f478e292d477883741605d54/modules/daggerverse/main.go#L207-L208
- [ ] Move s3 uploads + cloudfront invalidation out of goreleaser to a dagger function (since each publisher runs on each artifact, resulting in too many API calls)
- [x] Fix incorrect links for discord release notifications for SDKs 
image


modules/daggerverse/main.go
```diff
@@ -132,6 +132,10 @@ func (h *Daggerverse) BumpDaggerVersion(
 		from = strings.TrimSpace(from)
 	}
 
+	// get just the version, without the v semver prefix
+	from = strings.TrimPrefix(from, "v")
+	to = strings.TrimPrefix(to, "v")
+
```
gerhard, line 138: I did not expect this to be necessary. `awk -F'[v.]'` should have taken care of this. Will be taking a closer look at why this is an issue.

jedevc:
> I have a few more for `modules/daggerverse` that I would like to add part of this

Sounds good! I do want to merge this PR ideally today - I won't get all of my todos done by EOD, so will probably split these out into tasks for tomorrow.

modules/daggerverse/main.go
gerhard: I see! I was always dropping the `v` when specifying this value. It make sense that in CI the ref value has `v`, so this is a legit change.

gerhard:
I just realise that there was one important file which was not updated: https://github.com/dagger/dagger/blob/ba70be4ef8c2517bf9b683ca49e565cfaa771cae/.github/workflows/_dagger_on_depot_remote_engine.yml#L18

Pushing this change now.

How did you update all the other GitHub workflows @jedevc? This is how I do it:

```console
❯ jd bump 0.14.0 0.15.0
find . \( -name '*.yml' -o -name '*.yaml' \) -type f -exec gsed -i 's/0-14-0/0-15-0/g; s/0\.14\.0/0\.15\.0/g' {} +
```

sipsma (approved):
Merging now because dagger.io being updated already is causing failures in sdk dev jobs (due to CLI being at v0.14 but engine at v0.15), e.g. https://github.com/dagger/dagger/actions/runs/12287348931/job/34291486681?pr=9171

jedevc:
@sipsma that feels like a bug to me - we shouldn't be pinning the dagger CLI to latest for building the dev engine there. I think it's time for a refactor there (have added it to my todo list).

Issue #9120, opened 2024-12-06
🐞 shell: variable should expand to string, expands to ID instead


shykes:
## What I did

```console
dagger shell -c 'FUNC=$(.container | from alpine | with-exec echo,foo | stdout); .doc $FUNC'
```

## What I expected

```
Error: no function "foo" in type "Query"
```

## What happens

```
Error: no function "DSH:eyJjYWxscyI6W3sib2JqZWN0IjoiUXVlcnkiLCJuYW1lIjoiY29udGFpbmVyIiwiYXJndW1lbnRzIjp7fSwicmV0dXJuT2JqZWN0IjoiQ29udGFpbmVyIn0seyJvYmplY3QiOiJDb250YWluZXIiLCJuYW1lIjoiZnJvbSIsImFyZ3VtZW50cyI6eyJhZGRyZXNzIjoiYWxwaW5lIn0sInJldHVybk9iamVjdCI6IkNvbnRhaW5lciJ9LHsib2JqZWN0IjoiQ29udGFpbmVyIiwibmFtZSI6IndpdGhFeGVjIiwiYXJndW1lbnRzIjp7ImFyZ3MiOlsiZWNobyIsImZvbyJdfSwicmV0dXJuT2JqZWN0IjoiQ29udGFpbmVyIn0seyJvYmplY3QiOiJDb250YWluZXIiLCJuYW1lIjoic3Rkb3V0IiwiYXJndW1lbnRzIjp7fSwicmV0dXJuT2JqZWN0IjoiIn1dfQo=" in type "Query"
```

helderco:
That's not an ID, it's the encoded shell state, i.e., what we pass around in the pipes. I'll resolve it if passed as an argument to any builtin in general.

PR #9168, opened 2024-12-11
shell: expand state arguments in commands

helderco:
Fixes https://github.com/dagger/dagger/issues/9120

Included a refactor to remove some duplication. Simplified usage of `handleResponse` because:
- The shell doesn't use the `-o,--output` flag convenience
- Adding the fake `_type` field when result is an object is not necessary since you can add `... | .doc` to get the type

### DIFF
--- a/core/integration/module_shell_test.go
+++ b/core/integration/module_shell_test.go
@@ -94,12 +94,16 @@ func (Test) Go() string {
 
 package main
 
-type Dep struct{}
+func New() *Dep {
+	return &Dep{
+		Version: "dep function",  
+	}
+}
 
-// Dep version
-func (Dep) Version() string {
-	return "dep function"
-} 
+type Dep struct{
+	// Dep version
+	Version string
+}
 `,
 		)).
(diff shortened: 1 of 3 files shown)


core/integration/module_shell_test.go
```diff
@@ -620,6 +624,15 @@ func (ShellSuite) TestArgsSpread(ctx context.Context, t *testctx.T) {
 	}
 }
 
+func (ShellSuite) TestCommandStateArgs(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+	script := fmt.Sprintf("FOO=$(container | from %s | with-exec -- echo -n foo | stdout); .doc $FOO", alpineImage)
```
jedevc, line 629: Does something like `FOO=$(container | from %s); $FOO | with-exec -- echo -n foo | stdout` also work now?

jedevc (approved):

core/integration/module_shell_test.go
helderco: Here it is!
- https://github.com/dagger/dagger/pull/9191

PR #9171, opened 2024-12-12
fix metrics display via tracking by call ID digest

sipsma:
Some of the recent changes to telemetry span handling resulted in most metrics not showing up in the TUI (other than those for services).

This fixes the display by tracking+displaying metrics based on their associated Call ID digest, which avoids the relative complication of all the various spans we create in different parts of the engine and associated logic on which are actually rendered in the TUI.

### DIFF
--- a/.changes/unreleased/Fixed-20241211-195735.yaml
+++ b/.changes/unreleased/Fixed-20241211-195735.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: |
+  Metrics display in the TUI is fixed to display for all executed containers, rather than just services.
+time: 2024-12-11T19:57:35.761154448-08:00
+custom:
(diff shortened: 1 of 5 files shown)


cwlbraa (approved):
Do you think maybe it should be verbosity=2? 

Lgtm, I kinda wanna backfill some tests tomorrow.

sipsma:
> Do you think maybe it should be verbosity=2?

It was 3 previously (intrinsically because that's when the spans we were associating metrics to were rendered), so just want to leave as is for now

> Lgtm, I kinda wanna backfill some tests tomorrow.

Awesome, thank you 🙏

sipsma:
Engine scan failure is unrelated and fixed by https://github.com/dagger/dagger/pull/9172

PR #9172, opened 2024-12-12
upgrade golang.org/x/crypto dep to avoid CVE

sipsma:
CVE doesn't seem especially relevant to us, but need this to avoid CVE scan failures.

Example failure: https://github.com/dagger/dagger/actions/runs/12287348922/job/34289228329?pr=9171#step:3:3362


sdk/typescript/runtime/go.mod
jedevc, line 1: :thinking: hm, this is a dagger module, this probably needs a `dagger develop` first, these dependencies shouldn't be wiped.

sdk/php/runtime/go.mod
jedevc, line 1: Likewise with the php module here.

sdk/elixir/runtime/go.mod
jedevc, line 1: And elixir.

jedevc (approved):

sdk/typescript/runtime/go.mod
sipsma: Oops yeah I was overzealous with a `find` somewhere, re-ran `dagger develop` on all the sdk runtimes

jedevc (approved):
Ah, looks like some code was still using ye-olde unscoped enum values (breaking linting).

Otherwise, LGTM, good to merge when we're green :tada:

PR #9175, opened 2024-12-12
More release improvements from v0.15.0

jedevc:
Splitting out some of the additional tasks from https://github.com/dagger/dagger/pull/9167


gerhard:
Are [these lint failures](https://dagger.cloud/dagger/traces/79ce0f4d7134f5235f8945633b45e848) addressed in a separate PR @jedevc ?

image

jedevc:
Nope, they're fixed now :laughing:

modules/daggerverse/main.go
gerhard, line 1: LGTM!

Just to double-check:
- the author of this PR will be `dagger-ci` (including assignee)
- the reviewer will be the release engineer (you in the case of `v0.15.0`)

gerhard (approved):
Left a few nits, no reason to block the merge over them.

Look forward to confirming that this works as expected **after** we merge it.

modules/daggerverse/main.go
jedevc: Yup, that's what should happen now! Only the assignee has been changed, so everything will work exactly as before, except for that bit.

.dagger/cli.go
```diff
@@ -151,14 +202,15 @@ func (cli *CLI) TestPublish(ctx context.Context) error {
...
-		WithExec([]string{"apk", "add", "aws-cli"})
+
+	ctr := dag.Container().From(goReleaserImage)
+
+	// HACK: this can be enabled to force a go update (e.g. when we need it for
+	// a security update)
+	// ctr = ctr.WithDirectory(
+	// 	"/usr/local/go",
+	// 	dag.Container().From("golang:<version>-alpine@sha256:<hash>").Directory("/usr/local/go"),
+	// )
```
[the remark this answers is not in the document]
jedevc: Agreed this is cruft, but it's cruft we'll probably use again - IMO, having this inline for when we inevitably need it next (probably the next go release) is better than having to come back and attempt to reconstruct the logic. I want to try and keep as much tribal knowledge out of the release + maintenance process - the more tribal knowledge we have, the more it rests on just a couple of people.

Actually, the idea is to remove this logic entirely. As part of the release improvements, I want to have goreleaser ingest the completely built binaries, instead of attempting to build them itself - it's frustrating to have the discrepancy we already have (and it's required if we want to persue the one-binary goal). With that, we can entirely remove this snippet.

Issue #8711, opened 2024-10-15
🖐️ Monorepo support


drmikecrowe:
### What happened? What did you expect to happen?

Does anybody have any examples of a monorepo?  For example, I'm trying to learn how to:

- Creating the buildEnv at the top level and leveraging this for all apps
- Caching of `node_modules` in each projects (some node tooling we use forces us to use individual node_modules at the moment)
- Supporting 2 source folders (`apps/` and `packages/` -- both of which are used)

mircubed:
@vikram-dagger This is an issue to keep in mind for the documentation that you are working on. @TomChv can sync with you on the best practices.

TomChv:
I'm removing the `sdk/typescript` label since it's not a specific TS issue and we should write guidelines for every SDKs :)

PR #9179, opened 2024-12-12
docs: Add layout options for modules in monorepos

vikram-dagger:
Also related to https://github.com/dagger/dagger/issues/8711

### DIFF
--- a/docs/current_docs/api/module-structure.mdx
+++ b/docs/current_docs/api/module-structure.mdx
@@ -238,6 +238,14 @@ Dagger modules provide a framework that you can use to break up this complexity 
 
 1. One top-level Dagger module per project, with sub-modules for the various CI workflows in that project. This pattern is suitable when there are significant differences between the projects in the monorepo (e.g. a monorepo with SDKs, CLIs, web applications, docs, all of which have different CI requirements).
 
+```mermaid
+graph TD
+    A[Top-level orchestrator module] --> B[Webapp frontend module]
+    A[Top-level orchestrator module] --> C[Webapp backend module]
+    A[Top-level orchestrator module] --> D[Utilities module]
+    A[Top-level orchestrator module] --> E[Docs module]
+```
+
 Benefits of this pattern include:
 
 - Easier debugging: Sub-modules provide a way to separate, and therefore easily debug, the business logic for different CI tasks.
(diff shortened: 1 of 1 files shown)


docs/current_docs/api/module-structure.mdx
```diff
@@ -9,7 +9,7 @@ import DaggerModuleInit from '../partials/_dagger_module_init.mdx';
 
 <DaggerModuleInit />
 
-## Module layout
+## Physical layout options
```
marcosnils, line 12: Is physical layout a used term? Here's how nx references it for example. https://nx.dev/concepts/decisions/folder-structure

TomChv (commented):
That's a great addition, maybe we could add a schema to make it more understandable?

Like this schema for the point number 1

image

Maybe having a repo as example would be nice too (dagger is actually a great example for the point 1), something simple could help too :)

docs/current_docs/api/module-structure.mdx
```diff
@@ -228,6 +230,44 @@ class MyModule {
...
+    A[Shared module] --> D[Microservice 3]
+```
+
+Benefits of this pattern include:
+
+- Code reuse: This reduces code duplication and ensures a consistent CI environment for all projects. For example, the shared module could create a common build environment and leverage this for multiple projects in the monorepo.
+- Reduced onboarding friction: There is no need to create a new CI module when adding a new project or component. New projects can get started faster with their CI implementation.
+- Best practices: All projects benefit from the best practices implemented in the shared module.
+- Knowledge sharing: By contributing to a shared CI module, project teams can learn from each other's CI strategies.
+
```
vikram-dagger, line 270: Should there be a third approach here - "hybrid" combining the previous two approaches? When is that applicable and does it offer any benefits?

vikram-dagger:
> That's a great addition, maybe we could add a schema to make it more understandable?
 
Thanks for the feedback @TomChv , I added some simple diagrams - let me know what you think?

> Maybe having a repo as example would be nice too (dagger is actually a great example for the point 1), something simple could help too :)

My understanding is that `dagger/dagger` is currently a hybrid of both approaches, so I'm not sure if we can use that as an example. Although one of my questions is whether we should also document such a combination approach - see https://github.com/dagger/dagger/pull/9179/files#r1909007607

docs/current_docs/api/module-structure.mdx
TomChv: I think that hybrid approach happens when you have complex microservices infra that share common fcts but also expose their own one that will be used by a top level module.

That's kind of what we have at dagger like you said.

I don't think it's worth documenting that, since every repo si specific and some new unexpected patterns may appears, but it might be good enough to add a link to dagger as a concrete example of these patterns application :)

TomChv (approved):
Approved, documenting the hybrid pattern isn't worth it imo :)

PR #9180, opened 2024-12-12
ci: call publish-metadata in publish workflow

jedevc:
I *thought* I updated this in https://github.com/dagger/dagger/pull/9175.

Apparently not. This updates the `publish.yml` workflow as well to include this.


gerhard (approved):
👍

PR #9181, opened 2024-12-12
ci: add automated docs deployment

jedevc:
Internal ticket for this: https://linear.app/dagger/issue/DEV-4648/automate-docs-release

TODO:
- [x] Add netlify token to gha


.dagger/docs.go
```diff
@@ -260,6 +264,116 @@ export const daggerVersion = "%s";
...
+	netlifyToken *dagger.Secret,
+	// +optional
+	deployment string,
+) error {
+	api := "https://api.netlify.com/api/v1"
+	site := "docs.dagger.io"
+	branch := "main"
+	client := http.Client{}
+
+	token, err := netlifyToken.Plaintext(ctx)
```
gerhard, line 319: Rather than adding this as a GitHub Secret, what are your thoughts @jedevc on wiring this through with `onepassword-sdk-go`, same as we do in https://github.com/dagger/dagger.io/blob/1408b23aaa0c3744e6a1b3831d97d6f452ef6408/dagviz/.dagger/main.go#L51-L83 ?

cc @aluzzardi

gerhard (approved):
LGTM 👍

.dagger/docs.go
jedevc: Ideally, I'd like to avoid putting the onepassword integration into our CI code itself - it makes sense for our dagger/dagger.io repo, since it's private and so can do things like depend on the structure of the vault. But, we probably shouldn't do that for our public CI IMO.

I'd like to use the secret providers when they land - that feels like the right way to go. Especially as we'll want to do that conversion *anyways* - if we integrate `onepassword-sdk-go`, we'll probably be ripping it out later to use the secret providers anyways, so would rather just hold on this action for now.

PR #9184, opened 2024-12-12
fix: ts SDK ignore private prop type resolution

### LINKED ISSUE #9177: 🐞 v0.15.0 break the pipeline both in local and in github actions
### What is the issue?

Dagger was in `latest` on our CI and I realized today that the CIs no longer pass while they passed 2 days ago. 

The only variable different between these 2 days is the release of `v0.15.0`.

So I immediately installed the latter. And indeed it crashes locally too.

So I fixed in `v0.14.0` on our CI for the moment.

I did not expect that a minor version could break everything and I hope that it is me who did something wrong.

There are no commands related to the module that seem to work.

`dagger functions` - doesn't work
`dagger call <any-function>` - doesn't work 

### Dagger version

dagger v0.15.0 (registry.dagger.io/engine:v0.15.0) darwin/arm64 (did the same errors on github actions)

### Steps to reproduce

To be honest, I don't know which information would be useful to give you since I didn't succeed in reproducing this issue. **Don't hesitate to ask for any further information to help with this.**

We have a `bun` repository that uses Dagger with its default package manager, which seems to be `yarn`. Bun is at version `1.1.36`, and I'm using Node version `22.11.0`.

I saw that there's an experimental `bun` runtime to specify in Dagger's `package.json`, but it didn't solve the issue either.

Dagger is set up with the `typescript` SDK.

Here are our package files that might be useful:
```sh
❯ tree -L1
.
├── README.md
├── asserter
├── biome.json
├── build.ts
├── bun.lockb
├── common
├── dagger
├── dagger.json
├── docs
├── grapher
├── node_modules
├── package.json
├── profiler
├── test-utils
├── tsconfig.bun.json
├── tsconfig.json
└── typedoc.json

❯ cd dagger
❯ tree -L1
.
├── package.json
├── sdk
├── src
├── tsconfig.json
└── yarn.lock

❯ cd sdk/src
❯ tree -L1
.
├── api
├── common
├── connect.ts
├── connectOpts.ts
├── index.ts
├── module
└── telemetry
```

There is only one file `index.ts` under `dagger/src/index.ts`.

Here is dagger's config files:
`dagger.json`:
```json
{
  "name": "dragee-model",
  "engineVersion": "v0.15.0",
  "sdk": "typescript",
  "source": "dagger"
}
```

`dagger/tsconfig.json`
```
{
  "compilerOptions": {
    "target": "ES2022",
    "moduleResolution": "Node",
    "experimentalDecorators": true,
    "paths": {
      "@dagger.io/dagger": [
        "./sdk/src"
      ],
      "@dagger.io/dagger/telemetry": [
        "./sdk/src/telemetry"
      ]
    }
  }
}
```

`dagger/package.json`
```json
{
    "type": "module",
    "dependencies": {
        "@dagger.io/dagger": "./sdk",
        "typescript": "^5.5.4"
    },
    "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
}
```

### Log output

Here is the log that fail on the CI with every informations
[0_Check PR's lint and tests.txt](https://github.com/user-attachments/files/18112746/0_Check.PR.s.lint.and.tests.txt)

TomChv:
When merging #8824, I forget to add a condition to not resolve private property type, this PR adds this check and add a test so it will not happen anymore.

Should fixes #9177


core/integration/module_typescript_test.go
```diff
@@ -1779,3 +1779,37 @@ class Test {
...
+
+	_bar: any
+
+	constructor() {
+		this.foo = 'foo'
+		this._bar = 42
+	}
+
+	@func()
+	bar(): string {
```
jedevc, line 1805: What about if this type is something garbage and unresolvable? Like `thisdoesnotexist`?

core/integration/module_typescript_test.go
```diff
@@ -1779,3 +1779,37 @@ class Test {
...
+	modGen := c.Container().From(golangImage).
+		WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+		WithWorkdir("/work").
+		With(daggerExec("init", "--name=test", "--sdk=typescript")).
+		With(sdkSource("typescript", `import { object, func } from '@dagger.io/dagger'
+
+@object()
+export class Test {
+	@func()
+	foo: string
```
jedevc, line 1795: We don't need this property in the test.

core/integration/module_typescript_test.go
TomChv: The return type (here: `string`) is exposed  so it has to be valid and supported by Dagger.
For `_bar`, it needs to be a valid typescript type (not necessarily supported by Dagger though) otherwise the runtime will just fail to run the code, like in go if you put `func a(b: thisdoesnotexist)`, the code will simply not compile.

core/integration/module_typescript_test.go
TomChv: True, I removed it

core/integration/module_typescript_test.go
jedevc: Fair!

jedevc:
I *think* this is fine, but just want to check with @helderco - in go, all of it has to perfectly typechecked (since it's all compiled). But I think it makes sense to keep this in sync with how python works.

TomChv:
>  in go, all of it has to perfectly typechecked (since it's all compiled). But I think it makes sense to keep this in sync with how python works.

I disagree with that statement regarding Go, if I create a Dagger object with a unsupported type, it's gonna break when calling Dagger.
Example:

```go
type Foo struct {
  Test map[string]interface{}
}
```

But if I make it (not exposed) private, it might work because it's not registered in the Dagger API anymore right?
The code is perfectly compiled but the type itself isn't supported by Dagger so it fails, that's exactly the same here for Typescript.

jedevc:
Oh *right*, sorry my confusion.

jedevc:
Hm, but if I have a private function that returns a `string`, but is declared as returning an `int`, the go code won't compile - I think that's more the case I'm thinking of

TomChv:
> Hm, but if I have a private function that returns a string, but is declared as returning an int, the go code won't compile - I think that's more the case I'm thinking of

Even if the type is `interface{}`? I can change that in the test if that's disturbing, the only thing I want to test is that I can store a private property that has a type not supported by Dagger

TomChv:
## ⚠️ DON'T MERGE YET ⚠️ 

After more investigation, this PR doesn't fix the actual issue and shouldn't since using `any` make the execution extremely complex.

Here's an example of error that I tried with v0.14.0 when I added a private `any` field

```
     Error: Unsupported type any
```

Now the error with v0.15 is:

```
IntrospectionError: could not resolve type reference for any
```

So after double checking, it seems the code has the attended behaviour. We need more information on https://github.com/dagger/dagger/issues/9177 to fix it properly.

TomChv:
I'm closing this PR since #9177 is closed too, the error was due to lack of type annotation for return type which is an expected behaviour so this PR doesn't make sense anymore.

PR #9191, opened 2024-12-13
shell: support state result

helderco:
This allows saving a pipeline in a variable and reuse it later:

```console
⋈ CTR=$(container | from alpine)
⋈ $CTR | with-exec echo foobar | stdout
foobar
```

## Before

Previously you had to echo the output yourself (with the added caveat that interpreter builtins like `echo`  are prefixed with `..` to avoid a conflict with function names):
```console
⋈ CTR=$(container | from alpine)
⋈ ..echo $CTR | with-exec echo foobar | stdout
foobar
```
Or you can define a function for a more natural usage:
```console
⋈ ctr() { container | from alpine; }
⋈ ctr | with-exec echo foobar | stdout
foobar
```

### DIFF
--- a/cmd/dagger/shell.go
+++ b/cmd/dagger/shell.go
@@ -476,6 +476,14 @@ func (h *shellCallHandler) Exec(next interp.ExecHandlerFunc) interp.ExecHandlerF
 			args = args[1:]
 		}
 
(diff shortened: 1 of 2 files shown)


jedevc (approved):
Huh! Neat :tada:

PR #9195, opened 2024-12-13
ci: bump dagger runners generation to g3

matipan:
# Description
`dagger-g2-*` runners are often over-provisioned, primarily due to how we request resources internally and how Karpenter allocates instances (full deep-dive of our setup can be seen [here](https://dagger.io/blog/argo-cd-kubernetes)) . Each instance type has a defined set of resources, however, not all of those resources are considered allocatable by Karpenter. An instance with 8 cores and 32Gi of RAM might only have 7 cores and 28Gi of RAM available for use. When we requested resources, we were requesting the exact number. If a new runner showed up and request 8 cores, instead of allocating an 8 core machine we would double and go for 16 cores. This, multiplied by the amount of PRs are created in dagger/dagger and the size of our runners, lead to a not so cost-effective CI.

In `g3` we introduce one major change: we change resource allocation to be more conservative. Now, if a `dagger-g3-v0-15-1-st-16c` is requested, we will allocate an instance that has 16 cores, instead of 32 cores like we did previously. We introduce this change as a new generation because it can have a performance impact. Jobs that were assuming to have 16 cores were actually operating with 32, so performance might defer.

---

Internal PR: https://github.com/dagger/dagger.io/pull/4128

### DIFF
--- a/.github/main.go
+++ b/.github/main.go
@@ -227,7 +227,7 @@ func BronzeRunner(
 	// +optional
 	dind bool,
 ) string {
-	return Runner(2, daggerVersion, 4, false, dind)
+	return Runner(3, daggerVersion, 4, false, dind)
 }
 
 // Silver runner: Multi-tenant instance, 8 cpu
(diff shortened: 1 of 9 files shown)


gerhard:
After all checks pass, I would:
1. rebase on top of `main`
2. mark as ready for review
3. merge https://github.com/dagger/dagger.io/pull/4128
4. merge once it has been approved & 3. completed

jedevc (approved):
LGTM from the CI yaml changes, etc :tada:

gerhard:
Let's ship this 🚀

PR #9199, opened 2024-12-13
fix "return nil on err" cases and enable related linter

sipsma:
Happened across an instance of incorrectly returning nil when err was set while working on stuff. This fixes that and enables the nilerr linter for catching these things, which revealed another occurence.

As far as I can tell these haven't been actively hurting us most likely (the error cases are extremely unlikely to happen) but worth preventing these spooky typos from happening again going forward.

### DIFF
--- a/core/modulesource.go
+++ b/core/modulesource.go
@@ -572,7 +572,7 @@ func (src *ModuleSource) ModuleConfig(ctx context.Context) (*modules.ModuleConfi
 	configFile, err := contextDir.Self.File(ctx, filepath.Join(rootSubpath, modules.Filename))
 	if err != nil {
 		// no configuration for this module yet, so no name
-		return nil, false, nil
+		return nil, false, nil //nolint:nilerr
 	}
 	configBytes, err := configFile.Contents(ctx)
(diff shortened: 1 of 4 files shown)


cwlbraa (approved):
lgtm

core/modulesource.go
```diff
@@ -572,7 +572,7 @@ func (src *ModuleSource) ModuleConfig(ctx context.Context) (*modules.ModuleConfi
 	configFile, err := contextDir.Self.File(ctx, filepath.Join(rootSubpath, modules.Filename))
 	if err != nil {
 		// no configuration for this module yet, so no name
-		return nil, false, nil
+		return nil, false, nil //nolint:nilerr
```
jedevc, line 575: Is this case not intentional? The `nolint` inline comment, and the comment above feels deliberate.

I feel like this might be possible when initializing a module for the first time in an empty directory maybe?

sipsma: Not following, the behavior is intentional to return nil err. I added the nolint comment so that the linter didn't yell at us for this case

PR #9200, opened 2024-12-13
wip: see if keeping APIs lazier speeds anything up

sipsma:
Noticed some ops on Directory/File were unlazying results when it wasn't really necessary to.

I'm pretty sure there are cases where this will for sure hurt parallelism quite a bit, but want to see a full CI run to get an idea on whether there's much of a practical impact right now for us.

If there is, we should consider doing this. If not, I'll probably just close and leave for now since it's *technically* a breaking change and subtle enough that we'd need to carefully think through the implications + impact on progress in the TUI/cloud.

### DIFF
--- a/core/directory.go
+++ b/core/directory.go
@@ -391,55 +391,15 @@ func (dir *Directory) WithNewFile(ctx context.Context, dest string, content []by
 func (dir *Directory) Directory(ctx context.Context, subdir string) (*Directory, error) {
 	dir = dir.Clone()
 
-	svcs, err := dir.Query.Services(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get services: %w", err)
-	}
-	bk, err := dir.Query.Buildkit(ctx)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


sipsma:
Huh, it's very hard to tell for sure yet because there's quite a bit of noise in our job run times, but there are a couple jobs that were way faster (like 2-4x) in the first run here (test-publish, scan-engine, _some_ of the SDK tests). This is comparing against two other CI runs (latest on main and another recent PR).

A few were a little bit slower, but again can't tell if that's just noise or real. It *is* plausible for this change to hurt performance in some cases too because there's such thing as "too much" parallelism, which is why this might be a bit tricky to work through.

I'll re-run CI again a few times in the background to sort through noise. If I see the same thing, I'll leave this PR open but may need to put it on ice for a bit to avoid `ETOOMANYRABBITHOLES`.

sipsma:
Eh re-runs showed less of a benefit if any, I might have gotten lucky the first run with a warm cache and/or more powerful machine. I'll close this for now, might be worth revisiting someday.

Issue #3923, opened 2022-11-18
Save & load pipelines


shykes:
## Overview

I propose adding a new feature to the Dagger API: saving & loading of pipelines.

* Save: snapshot the state of a pipeline, encode it as text, send it to the client
* Load: restore the state of a pipeline, decoded from text received from the client

This would be a very powerful primitive that we could use in many ways:

* **Stitching**. Allows clients to "stitch" pipelines together into more complex pipelines. This is necessary since some pipelines cannot be expressed as a single GraphQL query. *Note: this feature already exists, see next session*
* **Shareable content**. The encoded pipeline state can be shared and reused across Dagger installations. Because it's text, it's easy for users to send each other a pipeline as a text file, or even copy-paste its contents, like they would a Dockerfile.
* **Callbacks**. Possibly, some API features could be built on top of the concept of a "callback": an executable pipeline, passed by the client in encoded form, to be executed by the engine in the future: for example as a secrets provider, or as an events handler, a middleware or hook, scheduled execution, etc. etc. This may or may not require changes to the existing format and call convention (arguments? inputs/outputs?)
** **Advanced UI**. Image a visual UI where you can not only *visualize* your DAG, but select any part of it and *move* them around, copy-paste their contents like you copy-paste an image, or export them to another DAG..

## Relationship to existing IDs

The foundation for this feature already exists, in the form of content-addressable "IDs" in the Dagger API.

I propose that we keep this feature, but change how we think of it, and how we use it.

* When we query the "id" of a container, directory, secret or file, we're actually saving the state of the pipeline that produces that object
* When we lookup by "id" a container, directory, secret or file, we're actually 1) loading the state of the pipeline, and 2) asserting the type of the pipeline's output (eg. a file "id" can't be loaded into a container).

## Pipeline Executable Format

To implement this feature, we need to specify a standard format to encode/decode the pipeline "code". The starting point of this format should be the current ID format. The process of formalizing the specification will probably require some clarifications, cleanup etc. This is probably a good thing.

shykes:
cc @aluzzardi @sipsma

sipsma:
Many thoughts, but one immediate note is that this has significant overlap with some of the problems and possible solutions to session stickiness that @vito and I talked about previously. Specifically, see possible solution `A) Pin sources in returned IDs` in this issue: https://github.com/dagger/dagger/issues/3613

[...]

PR #9204, opened 2024-12-14
[Draft] dagger-native content hashing support

sipsma:
After banging my head for a while, figured out a good approach to the module loading improvements (that were in turn enabled by the filesync improvements) which should both fix the excessive cache invalidation and clean up a decent chunk of the core API.

One of the requirements is support for content-based digests in *dagql* caching (previously it only existed in buildkit). This will enable us to give ModuleSource types (local, git) customized digests that are independent of details irrelevant to SDK function calls. It also should come in handy in the future as we move more caching logic to be "dagger native".

This turns out to be pretty separable from the rest of the change and beneficial on it's own since it allows us to remove the `blob` API entirely, so sending this change out on its own first.

Needs some cleanup still + sdk regen, hence draft, but good subset of tests are passing locally so sending out to see how everything else in CI is doing async.

Will update with more thorough description, but high level is:
1. For now, content-based caching is supported by just enabling ID digests to be customized by implementations, so each ID still has a single digest. In the longer term we _may_ want support for multiple digests used under different circumstances, but for now that's needed so side-stepping that complication
2. `blob` can be removed entirely because all the previous users of it now just set a custom dagql ID digest on their return values, which is then used by the dagql cache
   * AFAICT buildkit's content caching should still work without the `blob` LLB source type thanks to @jedevc's upstream change to enable that in more cases, so that also is removed here.
3. There's a little bit of dagql internal refactoring, which was needed to ensure it always uses the digest specified in an ID rather than recalculating it (which was a bit wasteful in some codepaths anyways).


dagql/server.go
```diff
@@ -467,26 +467,8 @@ func (s *Server) Load(ctx context.Context, id *call.ID) (Object, error) {
...
-				Raw:  arg.Name(),
-			},
-		})
-	}
-	sel, _, err := base.ObjectType().ParseField(ctx, id.View(), astField, vars)
-	if err != nil {
-		return nil, fmt.Errorf("parse field %q: %w", astField.Name, err)
-	}
-	sel.Nth = int(id.Nth())
-	res, id, err := s.cachedSelect(ctx, base, sel)
```
sipsma, line 489: This chunk of code is why `cachedSelect` changed from accepting a Selector to accepting an ID.

This code in `Load` was taking an ID, converting it into a `Selector`, only for it to go back to an ID in `cachedSelect`. I think this was a bit wasteful from a performance perspective but more importantly it broke custom digests since the Selector didn't know about those and instead just always re-calculated the ID digest from the call chain.

I fixed it by just providing the ID directly since that seems simpler overall. The other callers of `cachedSelect` (now renamed to `cachedCall`) are barely impacted since they start with a Selector but can easily turn that into an ID by `Selector.AppendTo`

dagql/server.go
```diff
@@ -886,12 +936,14 @@ func (sel Selector) AppendTo(id *call.ID, spec FieldSpec) *call.ID {
 			// we don't include null arguments, since they would needlessly bust caches
 			continue
 		}
+		var isSensitive bool
 		if arg, found := spec.Args.Lookup(arg.Name); found && arg.Sensitive {
-			continue
+			isSensitive = true
```
sipsma, line 941: This logic around sensitive args changed because it was previously just not including the sensitive arg in the ID at all, which no longer works in the new logic where we pass around IDs rather than Selectors in more places.

The fix works by migrating the hiding of sensitive values in the `call/` pkg. It results in the same behavior as before in that an ID will never serialize (via `Display`, `Encode`, etc.) sensitive args and won't include them in digests. The difference is that the sensitive arg can still be carried around on the ID in memory.

sipsma:
I still need to update all the integ test git repos to fix `TestCustomSDK` (see [this commit](https://github.com/dagger/dagger/pull/9204/commits/d0c1a3a5a169a22c1e24fd835c804e6358d4ec78) with the fix to the local version of that test), but otherwise I think this is good to go, taking out of draft.

sipsma:
https://github.com/dagger/dagger/pull/9245 has some early code showing how all this stuff is gonna be used to help module source loading improvements (the last commit there specifically), just in case that's helpful.

vito:
Very cool to see this taking shape. Wanted to pick your brain on one thing: mostly based on vibes, I feel a little unsettled overloading the existing call digests field for content digests. In my mind those are the digest of the 'recipe' and it may be useful to correlate all such recipes, and as a separate step correlate them to their resulting content. Also, the fact that we have a pre-computed 'digest' field was originally an optimization/implementation detail. So just keeping an ear to the ground for shifts in usage there.

Way back in my [original sketch for IDs](https://github.com/dagger/dagger/issues/3923#issuecomment-1714090592) I suggested we may want to add a content digest field:

> Future possibilities, need more thought:
> 
>  * Allow IDs to embed the `digest` of the content they create so we can verify that they created the right content, or even detect IDs that are actually impure.

What I originally had in mind was that once you got your hands on a "digested/synced" ID (aka an ID that contains its content digest, maybe returned by `sync`), you could later publish that ID alongside the artifact it built. Users could evaluate it, and check that the digest matches the recipe's digest. Or it could be published to the Daggerverse, which could sync and ensure that the digest matches before vouching for its hermetic-ness. Things like that.

There were gaps in this design, since adding the content digest to the ID/call implies a change to the ID/call's digest, but I suppose we already had to deal with that when we added the explicit digest field.

Anyhow - this isn't blocking, mostly tossing up a lob in case it makes things click for a jam, and wanting to be conscious of semantic changes to existing concepts. Curious to hear if there are any major benefits I missed to just rolling forward as-is and making digests work the way this PR does things instead, since most of what I'm throwing around is conceptual and may be more detached from reality than I realize. 🙃

sipsma:
> Anyhow - this isn't blocking, mostly tossing up a lob in case it makes things click for a jam, and wanting to be conscious of semantic changes to existing concepts. Curious to hear if there are any major benefits I missed to just rolling forward as-is and making digests work the way this PR does things instead, since most of what I'm throwing around is conceptual and may be more detached from reality than I realize. 🙃

@vito Yeah I started this thinking I'd be adding support for multiple digests being associated with a single call (so you could have the recipe digest plus some arbitrary number of other digests, including but not limited to content digest), but it ended up landing a little differently. 

A more accurate description of this PR is probably "add support for customizing digests" (I'll update the PR title and description after leaving this comment, the current ones are from the first iterations of this).

The current cases that it impacts are:
* When we had to go through the `blob` API in order to reset an operation's recipe-based digest to a content-based digest.
* When we had to do a big dance to make an operation pure *for a given client* (i.e. mark as impure but then return something that was pure but had the client mixed into the cache key somehow)

The important point is that for all of those use cases, we essentially were already just customizing the digest, but by jumping through a bunch of hoops and ceremonial dances to get the recipe to have the right digest/purity (even if it's not reproducible practically, such as with the `blob` api). Now, we can just do that much more directly, get rid of `blob`, etc.

So I don't think that any of the use cases impacted here would benefit from multiple digests being associated with a single call yet. And the use cases I have in mind for follow-ups don't need that yet either.

I do think in the longer-run we're gonna need something closer to what you're describing where there could be both a recipe-based digest, content-based digest, and maybe more all associated with a single ID. Namely, to fully migrate all caching logic away from buildkit we will want that for exec mounts (which can be purely identified by recipes but also have the same content across different recipes). It's just not needed yet and wouldn't benefit any of the current use cases, so I'm not gonna tackle it until it's needed.
* If it was utterly trivial I'd probably just do it now, but it's trickyish for the reason you mentioned in that we'd have to change how digests are calc'd to handle whether or not additional non-recipe-digests are included (maybe multiple digests are produced? idk yet). This has performance impacts since we call these digest codepaths *a lot*, so it's not a change we can easily make for free.

---

tl;dr all the current use cases for this are essentially the same as what we were doing before, just way simpler, and I think it makes sense to save support for multiple digests associated with a single call for when we actually need it

dagql/server.go
jedevc: Kinda curious what you think about https://github.com/dagger/dagger/pull/9395/files#diff-af763c875b05c71f4824382b701cb6b8075d3d8922153646f2e3e5d530b23d9cR569 - which is semi-related to the refactoring here.

Does it still seem reasonable to convert a chain of selectors into an ID? The idea being, it would be nice to be able to generate an ID for an object - but *without* actually performing any of the evaluation.

core/integration/module_config_test.go
```diff
@@ -843,7 +843,7 @@ var vcsTestCases = []vcsTestCase{
 	{
 		name:                     "GitHub public",
 		gitTestRepoRef:           "github.com/dagger/dagger-test-modules",
-		gitTestRepoCommit:        "323d56c9ece3492d13f58b8b603d31a7c511cd41",
+		gitTestRepoCommit:        "f1b295cc1bce8eeea33cc3f42f89452c6fb3429e",
```
jedevc, line 846: :eyes: the commit IDs for all these module no longer match. Is that intentional? (any chance we could keep them all aligned, so it's clear at a glance that they're all the same content)

core/schema/host.go
```diff
@@ -36,20 +39,6 @@ func (s *hostSchema) Install() {
 			return parent.NewHost(), nil
 		}).Doc(`Queries the host environment.`),
 
-		dagql.Func("blob", func(ctx context.Context, parent *core.Query, args struct {
```
jedevc, line 39: :tada:

dagql/call/id.go
```diff
@@ -243,32 +263,58 @@ func (id *ID) Append(
...
+		}
 	}
 
 	return newID
 }
 
+// Return a new ID that's the same as before except with the given metadata changed.
+// customDigest, if not empty string, will become the ID's digest.
+// tainted sets the ID's tainted flag.
+func (id *ID) WithMetadata(customDigest digest.Digest, tainted bool) *ID {
```
jedevc, line 294: Curious what the interplay between this and elective purity is - do selectors even need to be able to have the `Pure` field anymore? Since you can just do `WithMetadata` to taint the instance after the fact.

(and does that mean we could refactor the `CacheVolume` usage as well?)

jedevc (approved):
Looks good :tada: (approving now, since no major structural concerns!) I think I'm struggling to *fully* get my head round the dagql refactors, but will probably take a bit of tinkering with it myself :laughing: 

Definitely seems like I have a pretty hard dependency on this for https://github.com/dagger/dagger/pull/9395 - there's some conflict on `AsTarball`, but also some of the custom cache keying will be *super* useful (to get rid of the `Impure` hack I have).

core/integration/module_config_test.go
sipsma: I tried to do that by just cp'ing the same files in each repo and using the same commit msg, but I'm guessing that's not sufficient to get the exact same hash. Admittedly I didn't even google it yet so I'll do that 😄

jedevc: Yeah, sadly, the commit date will be different in each of those, so you'll get different commit hashes :cry:

dagql/server.go
sipsma: Discussed earlier today live, but yeah that sort of existed before and still does in this PR (it's now `Instance.Select`). I'm guessing you might need some variation on the exact types used there, but the idea/code should be reusable

dagql/call/id.go
jedevc: Eyyy :tada:

Can confirm that with that commit, looks like we're back down to "normal" logging again - back down to our regular 50MB (will close https://github.com/dagger/dagger/pull/9428 in preference of this significantly better fix)

core/integration/module_config_test.go
sipsma: Gonna merge now before more rebasing needed, will follow up with this

sipsma: Fix here https://github.com/dagger/dagger/pull/9444

PR #9226, opened 2024-12-16
ci: github actions: use runner.temp instead of hardcoded /tmp

clintonsteiner:
* See scenarios such as https://github.com/actions/setup-elixir/issues/24 where this causes issues


clintonsteiner:
Bumping with resolved merge

.github/actions/call/action.yml
```diff
@@ -135,24 +135,24 @@ runs:
...
         name: engine-logs-${{ runner.name }}
-        path: /tmp/actions/call/engine.log
+        path: ${{ runner.temp }}/actions/call/engine.log
         overwrite: true
 
     - name: Cleanup
       shell: bash
       run: |
-        rm -rf /tmp/actions/call
+        rm -rf ${{ runner.temp }}/actions/call
```
jedevc, line 158: I think we should be able to remove these cleanup steps entirely since `runner.temp` is cleaned up automatically.

clintonsteiner: Fixed

jedevc (approved):

Issue #9190, opened 2024-12-12
🐞 existing services break in 0.15


levlaz:
### What is the issue?

I have a sample go application (https://github.com/levlaz/snippetbox) that installs mariadb (https://daggerverse.dev/mod/github.com/levlaz/daggerverse/mariadb@cea1668da940b45864116049bd20087855c8c787) from the daggerverse and mounts it as a service binding. 

THe mariadb module is targeting 0.11.6 and the snippetbox module is targeting   0.13.16

This is no longer working in 0.15 with the following error: 

```
│ ✔ .withEnvVariable(name: "CACHEBUSTER", value: "2024-12-12 22:56:27.108879878 +0000 UTC m=+0.053134793"): Container! 0.0s
│ ✘ .withExec(args: ["sh", "-c", "mysql -h db -u root < internal/db/init.sql"]): Container! 0.1s
│ ┃ mysql: Deprecated program name. It will be removed in a future release, use '
│ ┃ /bin/mariadb' instead                                                        
│ ┃ ERROR 2026 (HY000): TLS/SSL error: self-signed certificate                   
│ ! process "sh -c mysql -h db -u root < internal/db/init.sql" did not complete successfully: exit code: 1
│ ○ .withExec(args: ["sh", "-c", "mysql -h db -u root snippetbox < internal/db/seed.sql"]): Container! 0.0s
│ ○ .withExec(args: ["go", "run", "./cmd/web", "--dsn", "web:pass@tcp(db)/snippetbox?parseTime=true"]): Container! 0.0s
✘ .terminal: Container! 4.9s
! failed to start service: solve mount /: process "sh -c mysql -h db -u root < internal/db/init.sql" did not complete successfully: exit code: 1
```

I suspect its not able to get the service to work due to the new API change, but I was under the impression that this change should not break existing pipelines like this. 

### Dagger version

dagger v0.15.0 (registry.dagger.io/engine:v0.15.0) darwin/arm64

### Steps to reproduce

dagger -m github.com/levlaz/snippetbox call server

### Log output

Here's a trace: https://dagger.cloud/levs-test-org/traces/f06d326bb79ffd99fb7afdc647b58f21

rajatjindal:
Hi @levlaz 

Thank you for reporting this issue. 

I just tried this with v0.14.0 and got the same error:

```
mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
[...]

PR #9231, opened 2024-12-16
`Container.up` consistency with `Container.asService`

jedevc:
Follow-up from #8865, and handles part of the issue in https://github.com/dagger/dagger/issues/9190.

This PR makes two distinct changes:
- Fix the `Container.up` to always use the user-expected version of `Container.asService`. This makes sense, since the former is just sugar for the latter. This was pretty hard to run into in-practice, since it requires that you call `Container.up` *inside* a module.
- Add all the args from `Container.asService` to `Container.up` - we should be able to set args like `insecureRootCapabilities` on `Container.up`, otherwise, there's really no other place to put them. I *think* this was just an oversight.

### DIFF
--- a/core/integration/legacy_test.go
+++ b/core/integration/legacy_test.go
@@ -786,13 +786,18 @@ func main() {
 
 import (
 	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
 	"dagger/foo/internal/dagger"
 )
 
 type Foo struct{}
 
-func (f *Foo) TestServiceUpEntrypoint(ctx context.Context, app *dagger.File) (string, error) {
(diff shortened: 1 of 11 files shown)


jpadams (approved):
I tested this with the cases for `asService` that were fixed in #9247  and got the same results consistently.

To test I did:
```
gh pr checkout 9247
git remote add jed https://github.com/jedevc/dagger
git fetch jed
git cherry-pick 97f4996392efbbb6947a974576d1f901f4f3ef05
git cherry-pick 3069367c1ae3bc5fc7de22ab8cad98ae2e75eee3
# no conflicts
# ran same commands with up instead of as-service...up
```

PR #9232, opened 2024-12-16
ci: Match any workflows which start with Alternative CI Runners

gerhard:
Not 100% sure whether this will work, but it should based on all the other resources which can be pattern-matched.

The other alternative is to list these explicitly.

Follow-up to:
- https://github.com/dagger/dagger/pull/9182


jedevc (approved):

jedevc:
Also no idea if this'll work :D

matipan (approved):
Looks like it could work, lets see!

PR #9234, opened 2024-12-16
Add experimental Deno support

TomChv:
### ✨ Deno support steps

- [x] Add Deno runtime container inside TS runtime
- [x] Add `deno.json` generation
- [ ] Run unit tests with Deno ⚠️ 
- [ ] Run integration tests with Deno ⚠️ 
- [ ] Run Dagger tests inside Deno container ⚠️ 

### :warning: Merge blockers

There are 2 problems that currently block the merge of that PR:
1. [Deno doesn't support local dependency installation](https://github.com/denoland/deno/issues/18474), meaning that we need to add all the SDK dependencies at the root `deno.json`, it's not ideal but that's the only way I could make it work.
2. I cannot run tests with `deno run -A npm:mocha` because  the `--loader` option cannot be passed, it throws an error. Theses issues https://github.com/mochajs/mocha/issues/5108 and https://github.com/mochajs/mocha-examples/issues/99 tracks progress but not much activities happens.

The PR is usable, but we cannot merge it until these 2 blockers are solved since it creates too much problem for long term supports.

### 🔎 Example of usage

```shell
# Init a Dagger project
$ dagger init --sdk=typescript --name=test-deno

# Set runtime to deno in `package.json`
npm pkg set "dagger.runtime=deno" 

# Rerun codegen to update module and generate a `deno.json`
$ dagger develop

# The module will now run inside a deno runtime
dagger call container-echo --string-arg foo
```

Signed-off-by: Tom Chauveau <tom@epitech.eu>

### DIFF
--- a/sdk/typescript/runtime/template/deno.json
+++ b/sdk/typescript/runtime/template/deno.json
@@ -1,0 +1,35 @@
+{
+  "nodeModulesDir": "auto",
+  "unstable": [
+    "bare-node-builtins",
+    "byonm",
+    "sloppy-imports",
+    "node-globals"
+  ],
+  "compilerOptions": {
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true
+  },
+  "imports": {
+    "@dagger.io/dagger": "./sdk/src",
+    "@dagger.io/dagger/telemetry": "./sdk/src/telemetry",
+    "@std/random": "jsr:@std/random@^0.1.0",
+    "graphql-request": "npm:graphql-request@7.1.0",
+    "graphql-tag": "npm:graphql-tag@^2.12.6",
+    "@grpc/grpc-js": "npm:@grpc/grpc-js@^1.11.1",
+    "@lifeomic/axios-fetch": "npm:@lifeomic/axios-fetch@^3.1.0",
+    "@opentelemetry/api": "npm:@opentelemetry/api@^1.5.0",
+    "@opentelemetry/core": "npm:@opentelemetry/core@^1.25.1",
+    "@opentelemetry/exporter-trace-otlp-http": "npm:@opentelemetry/exporter-trace-otlp-http@^0.53.0",
+    "@opentelemetry/sdk-metrics": "npm:@opentelemetry/sdk-metrics@^1.25.1",
+    "@opentelemetry/sdk-node": "npm:@opentelemetry/sdk-node@^0.52.1",
+    "@opentelemetry/semantic-conventions": "npm:@opentelemetry/semantic-conventions@^1.25.1",
+    "adm-zip": "npm:adm-zip@^0.5.15",
+    "env-paths": "npm:env-paths@^3.0.0",
+    "graphql": "npm:graphql@^16.9.0",
+    "node-color-log": "npm:node-color-log@^12.0.1",
+    "node-fetch": "npm:node-fetch@^3.3.2",
+    "reflect-metadata": "npm:reflect-metadata@^0.2.2",
+    "tar": "npm:tar@^7.4.2"
+  }
+}
--- a/sdk/typescript/runtime/template/deno.json
+++ b/sdk/typescript/runtime/template/deno.json
@@ -30,6 +30,7 @@
     "node-color-log": "npm:node-color-log@^12.0.1",
     "node-fetch": "npm:node-fetch@^3.3.2",
     "reflect-metadata": "npm:reflect-metadata@^0.2.2",
-    "tar": "npm:tar@^7.4.2"
+    "tar": "npm:tar@^7.4.2",
+    "typescript": "npm:typescript@^5.5.4"
   }
 }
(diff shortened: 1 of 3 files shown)


sdk/typescript/runtime/template/deno.json
beingminimal, line 1: This is test comment

beingminimal, line 1: This will work i think, i will appreciate it if you can give one try and let me know, if this will not work then i will give poc repo tomorrow. thanks

This is the change we need for consistent offline behavior.  Remember to run `deno cache` while online after making these changes.

```JSON
{
  "nodeModulesDir": "auto",
  "unstable": [
    "bare-node-builtins",
    "byonm",
    "sloppy-imports",
    "node-globals"
  ],
  "compilerOptions": {
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true
  },
  "imports": {
    "@dagger.io/dagger": "./sdk/src",
    "@dagger.io/dagger/telemetry": "./sdk/src/telemetry",
    "@std/random": "jsr:@std/random@0.1.0",
    "graphql-request": "npm:graphql-request@7.1.0",
    "graphql-tag": "npm:graphql-tag@2.12.6",
    "@grpc/grpc-js": "npm:@grpc/grpc-js@1.11.1",
    "@lifeomic/axios-fetch": "npm:@lifeomic/axios-fetch@3.1.0",
    "@opentelemetry/api": "npm:@opentelemetry/api@1.5.0",
    "@opentelemetry/core": "npm:@opentelemetry/core@1.25.1",
    "@opentelemetry/exporter-trace-otlp-http": "npm:@opentelemetry/exporter-trace-otlp-http@0.53.0",
    "@opentelemetry/sdk-metrics": "npm:@opentelemetry/sdk-metrics@1.25.1",
    "@opentelemetry/sdk-node": "npm:@opentelemetry/sdk-node@0.52.1",
    "@opentelemetry/semantic-conventions": "npm:@opentelemetry/semantic-conventions@1.25.1",
    "adm-zip": "npm:adm-zip@0.5.15",
    "env-paths": "npm:env-paths@3.0.0",
    "graphql": "npm:graphql@16.9.0",
    "node-color-log": "npm:node-color-log@12.0.1",
    "node-fetch": "npm:node-fetch@3.3.2",
    "reflect-metadata": "npm:reflect-metadata@0.2.2",
    "tar": "npm:tar@7.4.2",
    "typescript": "npm:typescript@5.5.4"
  }
}
```

TomChv: This looks already like the file set in the template, and I still see the sdk's dependency in your's so it's not what I expect.

What I expect is to see:

```
"imports": {
    "@dagger.io/dagger": "./sdk/src",
    "@dagger.io/dagger/telemetry": "./sdk/src/telemetry"
}
```

And give a way to the deno runtime to install sdk's dependency without having to list them in the deno.json

beingminimal: @TomChv here is why and what has been changed in proposed json changes.

**Key Changes and Explanations:**

*   **Pinning Package Versions (Highly Recommended):** The most important update is the suggestion to "pin" your package versions.  Instead of using `^` (which allows for minor and patch updates) or `~` (which allows for patch updates), use exact version numbers.  For example, instead of `"graphql-request": "npm:graphql-request@^7.1.0"`, use `"graphql-request": "npm:graphql-request@7.1.0"`.  This ensures that you always use the same versions of your dependencies, preventing unexpected behavior due to updates.  This is *especially* critical for offline work; if a dependency updates while you're online, your offline cache might become inconsistent.

*   **No Other Changes Needed (Generally):**  The rest of your `deno.json` looks correct for offline use.  The `nodeModulesDir: "auto"` and the way you're importing local modules are already appropriate.

**Why Pinning Versions is Crucial for Offline:**

When you use `^` or `~`, Deno (or npm, if you're using it directly) will try to fetch the latest compatible version of the package.  This is great for staying up-to-date, but it can cause problems when you go offline.  If a new version is released while you are online and Deno updates your cached version, your offline environment will now be using a different version than what was originally cached. This can lead to unexpected errors.

By pinning the versions, you ensure that the exact same packages that were downloaded initially are the ones used offline.  This provides consistency and makes your offline workflow much more reliable.

**How to Pin Versions:**

1.  **Check your current versions:** Run your Deno project once while online.  This will download and cache the dependencies.  Look in your `node_modules` directory (if you have `nodeModulesDir: "auto"`) to see the *exact* versions of the packages that were installed.

2.  **Update `deno.json`:**  Use those exact version numbers in your `deno.json` file.

3.  **`deno cache` (Recommended):**  After updating `deno.json`, run `deno cache` while online.  This will ensure that the correct pinned versions are in your cache.

By doing this, you'll have a much more stable and predictable offline development experience.

TomChv: Yes but it doesn't solve the problem I mentioned above

helderco: Pinned package versions are usually what lock files are for. Doesn't deno have one?

PR #9235, opened 2024-12-17
ci: merge several dependabot configs into one

jedevc:
This is pretty useful - I think maybe the `directories` option is new? Otherwise, I feel like I would have caught it before when doing all this. https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directories
 
Anyways, this will let us bump the sdk/go and the top-level directory go.mods together, so we won't need to do horrendous dependency resolution ourselves hopefully (like in https://github.com/dagger/dagger/pull/9187#issuecomment-2548200615).

While in the area, we can bump the docker update together as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,17 +15,17 @@ updates:
       - "area/engine"
       - "area/cli"
       - "area/sdk/go"
+    ignore:
+        # these are fiddly, we want to do them manually
+      - dependency-name: "go.opentelemetry.io/*"
+      - dependency-name: "github.com/moby/buildkit"
+      - dependency-name: "github.com/tonistiigi/fsutil"
+      - dependency-name: "github.com/vito/midterm"
     groups:
       engine:
         applies-to: version-updates
         patterns:
           - "*"
-        exclude-patterns:
-          # these are fiddly, we want to do them manually
-          - "go.opentelemetry.io/*"
-          - "github.com/moby/buildkit"
-          - "github.com/tonistiigi/fsutil"
-          - "github.com/vito/midterm"
 
   - package-ecosystem: "npm"
     directory: "/sdk/typescript"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,6 +1,9 @@
 version: 2
 updates:
   - package-ecosystem: "gomod"
+    directories:
+      - "/"
+      - "/sdk/go"
```
wingyplus, line 6: Do we need to add all runtime modules under SDKs to here?

jedevc: Hm, *perhaps*? When the runtime modules are run, all their dependencies that are shared with dagger itself will be auto-bumped to the right version.

So if there are other third party deps, then maybe we should - no real objection.

jedevc:
Just pushed a commit to actually fully ignore certain go packages in the ecosystem - this should stop PRs like https://github.com/dagger/dagger/pull/9220.

.github/dependabot.yml
wingyplus: Make sense. :)

.github/dependabot.yml
```diff
@@ -91,33 +76,18 @@ updates:
...
+      - "/sdk/python/runtime"
+      - "/modules/ruff/build"
     schedule:
       interval: "weekly"
       day: "monday"
       time: "07:00"
       timezone: "UTC"
     labels:
       - "kind/dependencies"
-      - "area/sdk/python"
```
helderco, line 102: Is this intentional?

jedevc: Ah, it was, but I've changed it back, I think it makes sense to leave.

The idea *was* that if we added more non-python docker images, we wouldn't want that label - but ideally in that case, we would just have different PRs for those, with their own labels.

jedevc: Done.

TomChv (approved):

PR #9239, opened 2024-12-17
docs: allow templating code snippets

jedevc:
Internal ticket: https://linear.app/dagger/issue/DEV-4861/automate-docs-snippets-bumps

For now, we just allow templating "{{ version }}" with the latest dagger version - however, we can extend that in the future.


docs/current_docs/integrations/snippets/argo-workflow.yaml
```diff
@@ -14,15 +14,15 @@ spec:
...
+          image: registry.dagger.io/engine:v{{ version }}
           securityContext:
             privileged: true
             capabilities:
               add:
                 - ALL
           readinessProbe:
             exec:
-              command: ["buildctl", "debug", "workers"]
+              command: ["dagger", "core", "version"]
```
vikram-dagger, line 25: I didn't understand the reason for this change.

jedevc: I put this change in a separate commit: https://github.com/dagger/dagger/pull/9239/commits/a3818e9f4517fcbea74d9e6298752cf03dceb09c

See https://github.com/dagger/dagger/pull/8219 for context - we should have updated the argo workflow as part of that PR, but it looks like it was missed.

PR #9243, opened 2024-12-17
backfill a test for withexec metrics propogation in golden test

cwlbraa:
this doesn't attempt to cover all withexec metrics, but does cover disk write and memory bytes, both of which render consistently.

[this prevents regressions like the one we had in our 0.15.0 release](https://github.com/dagger/dagger/pull/9171)

test-style-wise, I don't love this, so open to suggestions, but I wanted to re-use the UI-test-running machinery in dagql/idtui/golden_test.go without finding a way to make the output of `-vvv` calls more consistent.


dagql/idtui/golden_test.go
```diff
@@ -106,6 +107,28 @@ func (s TelemetrySuite) TestGolden(ctx context.Context, t *testctx.T) {
 			"with-exec", "--args", "echo,hey",
 			"stdout",
 		}, Fail: true},
+
+		// tests intended to trigger consistent tui exec metrics output
+		{Function: "disk-metrics", Verbosity: 3, FuzzyTest: func(t *testctx.T, out string) {
```
jedevc, line 112: Even though `Verbosity` is set to `3` here, why are the results of the `disk-metrics` file not using it?

cwlbraa: that file should not have been committed, it is fully unused. see the comments on the example definition.

dagql/idtui/golden_test.go
```diff
@@ -145,10 +171,16 @@ type Example struct {
...
+	// verbosities 3 and higher do not work well with golden, they're not very deterministic atm
+	Verbosity int
+	Fail      bool
+	// if a reason is given for Flaky, ignore failures, but log the failure and the provided explanation
+	// ineffectual if FuzzyTest is in use.
+	Flaky  string
+	Env    []string
+	DBTest func(*testctx.T, *dagui.DB)
+	// Using fuzzytest will eschew golden assertions and files on disk and allow string assertions instead
+	FuzzyTest func(*testctx.T, string)
```
cwlbraa, line 183: naming is bad here, but I'm drawing blanks

jedevc (approved):

cwlbraa:
so the other problem, looking at failing tests, is that this test requires cgroups v2.

we could ostensibly pull this whole suite out to run on namespace runners, but given that this single test also kinda breaks the pattern for the rest of these golden tests, maybe these 2 factors are enough that we should just break out another suite here, basically a verbose_test.go or similar.

thoughts? @vito

jedevc:
@cwlbraa the tests in `ProvisionSuite` suffer the same issue - they needed cgroupsv2, so we can do the nested cgroup magic we need to start docker in dagger.

For now, I think splitting to a separate suite is fine, the plan is to get our runners onto cgroupsv2 to avoid that in the near future.

jedevc:
@cwlbraa just removed the stale label, I think the tests need to be split out here to handle cgroupsv2?

Alternatively, happy to just skip this test now, with a TODO for when we get cgroupsv2 on our CI runners (cc @gerhard)

cwlbraa:
@jedevc yeah i'll get to it this week.

PR #9244, opened 2024-12-18
fix(cli): fix terminal on windows

### LINKED ISSUE #9096: 🐞 terminal fails on Dagger windows distribution
### What is the issue?

It's been a long while since I have used the `terminal` feature, but I know it did work for sure around the v0.10. There's a chance a bug was introduced a while back and I can't give a good timeline of when something may have changed, sorry.

The `terminal` feature is not working on the windows binary and I think it's easy to replicate with just a simple container from and terminal chained command.

I noticed something odd, after trying to debug some adhoc stuff I was learning with C++ build environments in docker, so i thought i'd use Dagger for this. I came across this error and thought maybe it's the docker image I was using, tried a few things, got it working in WSL but now I realised i think the problem is more general and not specific to an image, example below is the basic ubuntu image and that doesn't seem to work.

### Dagger version

dagger v0.14.0 (registry.dagger.io/engine:v0.14.0) windows/amd64

### Steps to reproduce

```
dagger core container from --address ubuntu terminal
```

Swap the image with anything, alpine etc and I'm confident the same result with occur, you are not able to get into a terminal from the dagger windows client binary

### Log output

```
dagger core container from --address ubuntu terminal                            PS  12:30:12 
✔ connect 0.2s
✔ inspecting core types 0.2s
✔ container: Container! 0.0s
✔ Container.from(address: "ubuntu"): Container! 0.4s
✘ Container.terminal: Container! 0.7s
! failed to forward exit code: failed to close terminal: EOF
┃ dagger / $
  ✔ start sh 0.7s

Full trace at https://dagger.cloud/mageep/traces/4756d309d0c41f6979d25cff3a6f02eb

Error: response from query: input: container.from.terminal resolve: failed to forward exit code: failed to close terminal: EOF               
Run 'dagger core container from terminal --help' for usage.
```

Trying on my WSL, with the exact same command, it worked.

```
mageep@PATRICK-PC:~$ dagger version
dagger v0.14.0 (registry.dagger.io/engine:v0.14.0) linux/amd64
mageep@PATRICK-PC:~$ dagger core container from --address ubuntu terminal
● Attaching terminal:
    container: Container!
    Container.from(address: "docker.io/library/ubuntu:latest@sha256:80dd3c3b9c6cecb9f1667e9290b3bc61b78c2678c02cbdae5f0fea92cc6734ab"): Container!

dagger / $
```

rajatjindal:
Fixes #9096

### DIFF
--- a/dagql/idtui/frontend_pretty.go
+++ b/dagql/idtui/frontend_pretty.go
@@ -65,6 +65,7 @@ type frontendPretty struct {
 	view       *strings.Builder  // rendered async
 	viewOut    *termenv.Output
 	browserBuf *strings.Builder // logs if browser fails
+	stdin      io.Reader        // used by backgroundMsg for running terminal
 
 	// held to synchronize tea.Model with updates
 	mu sync.Mutex
@@ -215,6 +216,9 @@ func (fe *frontendPretty) runWithTUI(ctx context.Context, run func(context.Conte
 		}
 	}
 	opts = append(opts, tea.WithInput(in))
+	// store in fe to use in backgroundMsg processing
+	// which is used for terminal command
+	fe.stdin = in
 
 	if out != nil {
 		opts = append(opts, tea.WithOutput(out))
@@ -738,16 +742,19 @@ func (fe *frontendPretty) update(msg tea.Msg) (*frontendPretty, tea.Cmd) { //nol
 		cmd := msg.cmd
 
 		if msg.raw {
-			var restore func() error
+			var restore = func() error { return nil }
 			cmd = &wrapCommand{
 				ExecCommand: cmd,
 				before: func() error {
-					ttyFd := int(os.Stdout.Fd())
-					oldState, err := term.MakeRaw(ttyFd)
-					if err != nil {
-						return err
+					if stdin, ok := fe.stdin.(*os.File); ok {
+						ttyFd := int(os.Stdout.Fd())
+						oldState, err := term.MakeRaw(int(stdin.Fd()))
+						if err != nil {
+							return err
+						}
+						restore = func() error { return term.Restore(ttyFd, oldState) }
 					}
-					restore = func() error { return term.Restore(ttyFd, oldState) }
+
 					return nil
 				},
 				after: func() error {
--- a/.changes/unreleased/Fixed-20250106-162442.yaml
+++ b/.changes/unreleased/Fixed-20250106-162442.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: |
+    Prevent interactive terminal sessions from crashing in Windows terminal
+time: 2025-01-06T16:24:42.762297531Z
+custom:
+    Author: rajatjindal
+    PR: "9244"


dagql/idtui/frontend_pretty.go
```diff
@@ -742,7 +742,7 @@ func (fe *frontendPretty) update(msg tea.Msg) (*frontendPretty, tea.Cmd) { //nol
 			cmd = &wrapCommand{
 				ExecCommand: cmd,
 				before: func() error {
-					ttyFd := int(os.Stdout.Fd())
+					ttyFd := int(os.Stdin.Fd())
```
jedevc, line 745: Hm, just to throw an edge-case your way - this should *also* work if `os.Stdin` is not really a tty. e.g. on linux this is like `echo "" | dagger core container from --address=alpine terminal`. Note, how this works - since stdin isn't a TTY, we try and open `/dev/tty` instead.

To handle this, we should use the same logic as we do with `findTTY`s https://github.com/dagger/dagger/blob/c91fe3d76dc580d662215f654b371b756f48bd1d/dagql/idtui/frontend_pretty.go#L206-L216 - perhaps we can save the result that we pass to `tea.WithInput` onto `frontendPretty` somewhere, and then use that instead?

rajatjindal:
Hi @jedevc 

do you have a specific scenario in mind where a user would do something like this?

dagql/idtui/frontend_pretty.go
rajatjindal: Hi @jedevc

do you have a specific scenario in mind where a user would do something like this?

jedevc: Yeah, it's more applicable when using `dagger shell`:

```
$ echo ".core | container | from alpine | terminal" | dagger shell
```

Then we have to spawn a `terminal`, but the stdin is actually just a pipe from `echo`.

rajatjindal: thank you Justin. 👍

I have pushed the code as suggested and also tested with Mac/Windows client.

thanks

dagql/idtui/frontend_pretty.go
```diff
@@ -742,7 +742,24 @@ func (fe *frontendPretty) update(msg tea.Msg) (*frontendPretty, tea.Cmd) { //nol
...
 				before: func() error {
-					ttyFd := int(os.Stdout.Fd())
+					var in *os.File
+					inr, _ := findTTYs()
+					if inr != nil {
+						in = inr.(*os.File)
+					}
+
+					if in == nil {
+						tty, err := openInputTTY()
```
jedevc, line 752: We should avoid opening this again - it should already be open right? 

> Perhaps we can save the result that we pass to `tea.WithInput` onto `frontendPretty` somewhere, and then use that instead?

I think we can save it from the `runWithTUI` function - or is there some reason that we need to find all the TTYs again?

rajatjindal: no, this is most likely my bad. I'll take a look again at this.

rajatjindal: I made changes to store the already opened handle in `frontendPretty` struct. I tested with Mac client, it works fine. i will test it with Windows client once as well and report back here.

I also want to look at it with fresh eyes to ensure there is no other optimization we could do for these changes.

rajatjindal:
I made some more changes today to fix CI failures, so I want to test this once again with a windows client machine. I will report back the results by tomorrow.

jedevc (approved):

jedevc:
@rajatjindal is this good to go?

rajatjindal:
> @rajatjindal is this good to go?

give me few mins to verify with my windows laptop. I'll do it right away.

rajatjindal:
> > @rajatjindal is this good to go?
> 
> give me few mins to verify with my windows laptop. I'll do it right away.

I can confirm that this works with windows as well as darwin client now. 👍 🎉

jedevc:
Pushed a changelog - will merge when CI passes :tada:

PR #9245, opened 2024-12-18
[WIP] fix module source cache invalidation + cleanup api

sipsma:
This is only half written, doesn't build yet, has commits from #9204, etc. But has enough to show usage of all the new caching features added in https://github.com/dagger/dagger/pull/9204. Sending out in case it's helpful for groking that PR and what the "grand plan" is.

Goals here are:
* Fix various sources of excessive module building cache invalidation
   * changing completely unrelated source files in context
   * building different git commits that have no changes in the module source
   * etc.
* Use the filesync improvements + new caching features in #9204 to cleanup tons of cruft from the API (ResolveFromCaller, Module.initialize, etc.)

This will have some breaking changes in the current plans:
* dagger.json's `include` field will now be co-opted to only need to be set when a module imports code from *local* non-daggerized packages. It should never need to be set for any other case.
   * I will likely remove `exclude` entirely since it's redundant with `!` include patterns
* `views` are *likely* gonna be removed since they don't serve much of a purpose anymore thanks to contextual args, though I'm not 100% set on this


sipsma:
revived here: https://github.com/dagger/dagger/pull/9505

PR #9247, opened 2024-12-18
engine: return error if services don't have a command set

### LINKED ISSUE #9190: 🐞 existing services break in 0.15

marcosnils:
this addresses the issue introduced in v0.15.0 where services will hang
to start if neither CMD or ENTRYPOINT is used.

fixes #9190

### DIFF
--- a/core/container.go
+++ b/core/container.go
@@ -1727,7 +1727,9 @@ func (container *Container) AsService(ctx context.Context, args ContainerAsServi
 	var cmdargs = container.Config.Cmd
 	if len(args.Args) > 0 {
 		cmdargs = args.Args
-		useEntrypoint = false
+		if !args.UseEntrypoint {
+			useEntrypoint = false
+		}
 	}
 
 	container, err := container.WithExec(ctx, ContainerExecOpts{
--- a/core/container.go
+++ b/core/container.go
@@ -1710,14 +1710,26 @@ func (container *Container) AsServiceLegacy(ctx context.Context) (*Service, erro
 }
 
 func (container *Container) AsService(ctx context.Context, args ContainerAsServiceArgs) (*Service, error) {
+	if len(args.Args) == 0 &&
+		len(container.Config.Cmd) == 0 &&
+		len(container.Config.Entrypoint) == 0 {
+		return nil, ErrNoSvcCommand
+	}
+
+	useEntrypoint := args.UseEntrypoint
+	if len(container.Config.Entrypoint) > 0 {
+		useEntrypoint = true
+	}
+
 	var cmdargs = container.Config.Cmd
 	if len(args.Args) > 0 {
 		cmdargs = args.Args
+		useEntrypoint = false
 	}
 
 	container, err := container.WithExec(ctx, ContainerExecOpts{
 		Args:                          cmdargs,
-		UseEntrypoint:                 args.UseEntrypoint,
+		UseEntrypoint:                 useEntrypoint,
 		ExperimentalPrivilegedNesting: args.ExperimentalPrivilegedNesting,
 		InsecureRootCapabilities:      args.InsecureRootCapabilities,
 		Expand:                        args.Expand,
--- a/core/container.go
+++ b/core/container.go
@@ -100,6 +100,9 @@ type Container struct {
 	// (Internal-only for now) Environment variables from the engine container, prefixed
 	// with a special value, that will be inherited by this container if set.
 	SystemEnvNames []string `json:"system_envs,omitempty"`
+
+	// DefaultArgs have been explicitly set by the user
+	DefaultArgs bool `json:"defaultArgs,omitempty"`
 }
 
 func (*Container) Type() *ast.Type {
@@ -1717,7 +1720,7 @@ func (container *Container) AsService(ctx context.Context, args ContainerAsServi
 	}
 
 	useEntrypoint := args.UseEntrypoint
-	if len(container.Config.Entrypoint) > 0 {
+	if len(container.Config.Entrypoint) > 0 && !container.DefaultArgs {
 		useEntrypoint = true
 	}
 
--- a/core/integration/git_test.go
+++ b/core/integration/git_test.go
@@ -438,6 +438,7 @@ func (GitSuite) TestServiceStableDigest(ctx context.Context, t *testctx.T) {
 			WithMountedDirectory("/repo", c.Git(url, dagger.GitOpts{
 				ExperimentalServiceHost: svc,
 			}).Branch("main").Tree()).
+			WithDefaultArgs([]string{"sleep"}).
 			AsService().
 			Hostname(ctx)
 		require.NoError(t, err)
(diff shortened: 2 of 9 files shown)


marcosnils:
moving to draft per last @jpadams 's comment here: https://discord.com/channels/707636530424053791/1318184687914713148/1319028692676251678

core/schema/container.go
```diff
@@ -1011,7 +1011,9 @@ type containerWithDefaultArgs struct {
 }
 
 func (s *containerSchema) withDefaultArgs(ctx context.Context, parent *core.Container, args containerWithDefaultArgs) (*core.Container, error) {
-	return parent.UpdateImageConfig(ctx, func(cfg specs.ImageConfig) specs.ImageConfig {
+	c := parent.Clone()
+	c.DefaultArgs = args.Args
```
jedevc, line 1015: Ok, I see.

Because `DefaultArgs` is only ever set to the same as `Cmd`, probably doesn't make sense to track it as a slice. Maybe we can just track `DefaultArgs bool` for explicitly tracking this error.

core/integration/git_test.go
```diff
@@ -438,6 +438,7 @@ func (GitSuite) TestServiceStableDigest(ctx context.Context, t *testctx.T) {
 			WithMountedDirectory("/repo", c.Git(url, dagger.GitOpts{
 				ExperimentalServiceHost: svc,
 			}).Branch("main").Tree()).
+			WithDefaultArgs([]string{"/bin/sh"}).
```
jedevc, line 441: nit: this command isn't a service - we should explicitly make it a command like `sleep` or something, instead of one that would just immediately exit if it was started.

(nit because the service isn't really started, so it doesn't matter)

core/container.go
```diff
@@ -1710,14 +1710,26 @@ func (container *Container) AsServiceLegacy(ctx context.Context) (*Service, erro
...
+	if len(args.Args) == 0 &&
+		len(container.Config.Cmd) == 0 &&
+		len(container.Config.Entrypoint) == 0 {
+		return nil, ErrNoSvcCommand
+	}
+
+	useEntrypoint := args.UseEntrypoint
+	if len(container.Config.Entrypoint) > 0 {
+		useEntrypoint = true
+	}
```
jedevc, line 1722: Hm, if `withDefaultArgs` is called before this, this will now apply the default entrypoint?

Seems like a mismatch with the proposal in https://github.com/dagger/dagger/issues/9305#issue-2766581758.

marcosnils: Yes, TBH I'm a bit confused about how this logic should actually work. Imagine the following scenario: 

```
myApiContainer := dag.Container().From("blah").WithEntrypoint("ept").WithDefaultArgs("blah") // this is my app container that the pipeline builds which later I want to publish to my OCI registry

//  as part of myApiContainer tests somewhere else in the code I do: 

svc := myApiContainer.AsService()

dag.Container().WithServiceBinding("api",svc)......
```

^ in this case, if we apply the logic to disable the entrypoint when setting `WithDefaultArgs` the above scenario ends up being super confusing since if you `pull` the container from the OCI registry it works, but if I use the one you've built it doesn't. :grimacing:

jedevc: Sigh - this is annoying, and I don't think is possible to fully resolve perfectly. That said, it's fairly minor? so happy to defer until someone hits it, maybe some of the other refactors we make to the service API in the future will help resolve this.

---

As a *potential* alternative - we could have this exact case be an error case (literally just changing `useEntrypoint = true` to `return fmt.Errorf("cannot execute base image default args without entrypoint")` - trying to execute the base image's `CMD` without it's `ENTRYPOINT` is invalid, so we could error (this would force users to put the `UseEntrypoint` in there).

This minimizes the inconsistent behavior from pushing and re-pulling - it makes the error show again even if `WithDefaultArgs` was used when pushed, but I think this is actually *probably* what we would expect.

But also this doesn't actually accomplish the out-of-the-box compatibility from https://github.com/dagger/dagger/issues/9305 - but it does steer users towards easier upgrades by having *clear explicit errors*.

core/container.go
```diff
@@ -1710,14 +1713,26 @@ func (container *Container) AsServiceLegacy(ctx context.Context) (*Service, erro
...
+	if len(args.Args) == 0 &&
+		len(container.Config.Cmd) == 0 &&
+		len(container.Config.Entrypoint) == 0 {
+		return nil, ErrNoSvcCommand
+	}
+
+	useEntrypoint := args.UseEntrypoint
+	if len(container.Config.Entrypoint) > 0 && !container.DefaultArgs {
+		useEntrypoint = true
+	}
```
jedevc, line 1725: Yup, can confirm this implements the proposal in https://github.com/dagger/dagger/issues/9305 as desired. LGTM :tada:

jedevc (commented):
Yup, can confirm this implements the proposal in https://github.com/dagger/dagger/issues/9305 as desired. LGTM :tada:

Holding off an approval though until we can confirm that we're happy to compromise here: https://github.com/dagger/dagger/pull/9247#discussion_r1907264949

jpadams (approved):
LGTM 🙏 

Tested manully here: https://discord.com/channels/707636530424053791/1318184687914713148/1327416811309760572

PR #9248, opened 2024-12-18
fix: silence trace setup messages using descriptive env name

### LINKED ISSUE #9242: ✨ Support descriptive ENV variables in place of SHUTUP, NOTHANKS, GOAWAY etc.

The purportedly whimsical environment variables `NOTHANKS`, `SHUTUP`, `GOAWAY` and others are both mildly offensive and confusingly named.

As a new-comer to the dagger tooling I am repeatedly put off by the existing supported environment variables to disable messaging about using dagger.cloud for tracing: https://github.com/dagger/dagger/blob/7bd6908820fb82e1cb457d92d2b59ab87d12537e/dagql/idtui/frontend.go#L53

Moreover, by not prefixing the names with `DAGGER_` it's unclear to developers what the environment variables control and it's possible that it may change unrelated application behavior unintentionally. (eg. if `GOAWAY` is referenced somewhere else)

Inclusivity, and a welcoming tone are important as the dagger community grows. Using terms like "Shut up" and "Go away" has a chilling effect on dagger users.

Per the instructions that are printed (for example):

```
Setup tracing at https://dagger.cloud/traces/setup. To hide: export GOAWAY=1
```

I'm forced to use one of the supported environment variables.

jedevc:
Fixes #9242.

We'll now only recommend using the prefixed environment variable `DAGGER_NO_NAG` - no more randomly cycling through. However, we'll keep supporting the old environment variables, in case users are using that to silence this output. We can deprecate those in the future.

Some light rephrasing of the message is required to fit the extra `DAGGER_` characters in - it's important we keep this line under 80 characters, so it fits nicely into standard sized 80x24 terminals.

### DIFF
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -3,7 +3,6 @@ package idtui
 import (
 	"context"
 	"fmt"
-	"math/rand"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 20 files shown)


marcosnils (approved):

thommahoney:
Thanks for being so responsive to my issue and for implementing this so quickly!

jedevc:
@thommahoney merged, so this should be part of v0.15.2 - just fyi, this'll probably land sometime in the new year!

PR #9249, opened 2024-12-18
helm: Enable deploying multiple Engines per node

gerhard:
This adds a new property - `kind` - which defaults to `DaemonSet`. It means that nothing will change for existing users: one Engine per K8s node.

For users that want to deploy **multiple** Engines per K8s node, set `kind: StatefulSet`, and the chart will take care of the rest.

This opens the door to experimenting with runtimes such as firecracker-containerd, gVisor, etc. for putting a sandbox around the Engine - should help address some of the `CAP_SYS_ADMIN` concerns.

Paired with @ioboi @chrira from [puzzle.ch](https://puzzle.ch)

## TODO

- [x] Tests
- [x] Changelog
- [ ] Add to next milestone after merging

### DIFF
--- a/helm/dagger/templates/engine-statefulset.yaml
+++ b/helm/dagger/templates/engine-statefulset.yaml
@@ -8,6 +8,9 @@ metadata:
   labels:
     {{- include "dagger.labels" . | nindent 4 }}
 spec:
+  # DO NOT RUN MORE THAN 1 REPLICA.
+  # Only one (1) Engine is able to hold the lock, other replicas will not be able to start.
+  replicas: 1
   selector:
     matchLabels:
       name: {{ include "dagger.fullname" . }}-engine
@@ -53,15 +56,15 @@ spec:
           {{- end }}
           {{- if .Values.magicache.enabled }}
           env:
-          - name: _EXPERIMENTAL_DAGGER_CACHESERVICE_URL
-            value: {{ required "A magicache url is required" .Values.magicache.url }}
+            - name: _EXPERIMENTAL_DAGGER_CACHESERVICE_URL
+              value: {{ required "A magicache url is required" .Values.magicache.url }}
           envFrom:
-          - secretRef:
-              {{- if .Values.magicache.secretName }}
-              name: {{ .Values.magicache.secretName }}
-              {{- else }}
-              name: {{ include "dagger.fullname" . }}-magicache-token
-              {{- end }}
+            - secretRef:
+                {{- if .Values.magicache.secretName }}
+                name: {{ .Values.magicache.secretName }}
+                {{- else }}
+                name: {{ include "dagger.fullname" . }}-magicache-token
+                {{- end }}
           {{- end }}
           securityContext:
             privileged: true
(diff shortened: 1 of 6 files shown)


helm/dagger/templates/engine-statefulset.yaml
```diff
@@ -0,0 +1,108 @@
...
+              subPath: engine.toml
+            {{- end }}
+      terminationGracePeriodSeconds: {{ .Values.engine.terminationGracePeriodSeconds }}
+      volumes:
+        - name: data
+          hostPath:
+            path: /var/lib/dagger-{{ include "dagger.fullname" . }}
+        - name: run
+          hostPath:
+            path: /var/run/dagger-{{ include "dagger.fullname" . }}
```
cwlbraa, line 99: does the engine play well with 2 processes rw-ing to these concurrently?

gerhard: Not at all.

There should always be a single replica, and each install gets a new name, e.g.:

image

```console
root@d24-10-22:/host/var/lib# ls -lahd dagger*
drwxr-xr-x 4 root root 69 Dec 18 12:08 dagger
drwxr-xr-x 4 root root 69 Dec 18 12:08 dagger-dagger2
drwxr-xr-x 4 root root 69 Dec 18 12:08 dagger-dagger3-dagger-helm
drwxr-xr-x 4 root root 69 Dec 20 17:13 dagger-dagger4-dagger-helm
```

cwlbraa:
broad strokes, lgtm. 

Just wanna make sure I'm seeing the whole picture here, though... StatefulSet is the choice here because it'll keep the host volumes and compute together through evictions? Or are there other additional reasons?

That said, I'm sure if there are negative implications to sharing those volumes.

cwlbraa (approved):

gerhard:
Thanks for the review @cwlbraa !

> StatefulSet is the choice here because it'll keep the host volumes and compute together through evictions? Or are there other additional reasons?

Reasons:
- Stable, unique network identifiers.
- Stable, persistent storage.
- Ordered, graceful deployment and scaling.
- Ordered, automated rolling updates.

> That said, I'm sure if there are negative implications to sharing those volumes.

Volumes must **NOT** be shared among Engines. See https://github.com/dagger/dagger/pull/9249#discussion_r1894348544

I would be curious for you to install this in your K8s of a choice in a few different ways and see if you can break it 👨‍🚒

helm/dagger/templates/engine-statefulset.yaml
cwlbraa: ah gotcha, interesting. so the statefulset is not intended to scale horizontally, but via adding additional statefulsets? if I'm understanding correctly, it's probably worth setting `replicas: 1` w/ a comment not to scale

gerhard:
> lol yeah I can tell you I would've broken it quickly by cranking up replicas

Indeed! While the chart does not support configuring replicas on purpose, we are not explicit about it. Let me see if I can add something about it so that people are not tempted to adjust it beyond `1`.

gerhard:
@cwlbraa I set the number of replicas to `1` explicitly, and while I couldn't find anything sensible to lock this down, scaling it above `1` fails in a decent way:

image
image

gerhard:
Went over this with a fresh pair of eyes. Made a few changes to the Helm tests, intend to merge as soon as all checks pass.

PR #9262, opened 2024-12-20
switch to custom init that also handles nested exec sessions

sipsma:
(Was working on this day before vacation, sending out in a spare moment but won't revisit before I'm back in 2 weeks)

---

This change replaces dumb-init with a custom one written in go that also serves session attachables for nested execs.

One motivation is to avoid a few minor headaches around cross-compiling C code (which dumb-init) is written in. Cross-compiling the new go implementation is trivial.

More than that, this allows transplants session attachables from the executor to this init process, which greatly reduces the hardship involved in adding new session attachables.
* See this thread https://github.com/dagger/dagger/pull/8730#issuecomment-2552079719

Rather than having to figure out how to get each attachable to work both in the CLI and in the executor (which involves complicated entering into the container namespaces), we can just have a single implementation of those attachables that works everywhere.

This doesn't go all the way back to the previous shim implementation (and it's associated headaches), only the attachables are running in the container again. We also avoid needing to do more "pass data via internal env var" by instead giving the init one side of a socketpair, which it uses to connect attachables rather than having to do a big dance to establish the connection itself using tons of client-specific metadata.

### DIFF
--- a/.dagger/build/builder.go
+++ b/.dagger/build/builder.go
@@ -246,7 +246,7 @@ func (build *Builder) Engine(ctx context.Context) (*dagger.Container, error) {
 		{path: "/usr/bin/dial-stdio", file: build.dialstdioBinary()},
 		{path: "/opt/cni/bin/dnsname", file: build.dnsnameBinary()},
 		{path: consts.RuncPath, file: build.runcBin()},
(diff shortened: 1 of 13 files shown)


aluzzardi (approved):
Sweet!

PR #9288, opened 2024-12-27
feat(sdk/dotnet): maintenance update

pjmagee:
- renamed Engine to QueryExecutor due to clash with latest API. This aligns both `QueryBuilder` and `QueryExecutor` to be similar and related 

- update introspection.graphql, Looks to be some changes to the original introspection.graphql, so I updated it

- update nuget packages

- ~removal of netstandard2.0: dotnet 8 as netstandard is being phased out and not recommended: This one needs some discussion but I was able to get it generating perfectly fine with net8! There is no need to target .net framework 4.7, it's windows only and this was a main piece for netstandard to allow backward compatible libraries for old and new ecosystem~

- defaultValue is added to nullable parameters, Arguments now should include missing value, e.g `string? dockerFile = "Dockerfile"` and `int? permissions = 420`

- leaf node suffixed with Async to better clearly define async methods. Now its aligning with convention of async functions suffix with `Async` across many libraries/frameworks including .NET own library

- cancellation token is now passed to request from method, also added unit test for cancellation token

- replaced _some_ hardcoded values with Roslyn Syntax (e.g keywords)


pjmagee:
@wingyplus  Before we merge, I am wondering if we can try and figure out the situation with .netstandard and .net5+ because I have read a lot of articles on the future of netstandard, and its basically a dead end. Because we have no reason for .net framework support (dagger functions with .net must work in containers), there is no reason to produce any library in netstandard2.0 as how I am seeing it. 

I know roslyn analyzers have some WARN (but i think this is in the context of very generic analyzers/source generators) which can still be applied across many peoples projects as independant packages, but in our case, this source gen is **only** for this dotnet sdk.

sdk/dotnet/sdk/Dagger.SDK/GraphQL/Types.cs
```diff
@@ -5,86 +5,114 @@ namespace Dagger.SDK.GraphQL;
...
+        return await new StringValue(id.Value).FormatAsync(cancellationToken);
     }
 }
 
 public class StringValue(string value) : Value
 {
-    public override Task<string> Format()
+    public override Task<string> FormatAsync(CancellationToken cancellationToken = default)
     {
-        var s = value
```
pjmagee, line 24: Try to only have one iteration, because Replace iterate over all many time with immutable strings

sdk/dotnet/sdk/Dagger.SDK.SourceGenerator.Tests/Dagger.SDK.SourceGenerator.Tests.csproj
```diff
@@ -7,14 +7,15 @@
...
+        <LangVersion>default</LangVersion>
     </PropertyGroup>
 
     <ItemGroup>
-        <ProjectReference Include="..\Dagger.SDK.SourceGenerator\Dagger.SDK.SourceGenerator.csproj"/>
+        <ProjectReference Include="..\Dagger.SDK.SourceGenerator\Dagger.SDK.SourceGenerator.csproj" />
     </ItemGroup>
 
     <ItemGroup>
-        <PackageReference Include="Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing.MSTest" Version="1.1.1"/>
```
pjmagee, line 17: This has been deprecated, Microsoft recommend move to more generic version (without .MSTest)

sdk/dotnet/sdk/Dagger.SDK/Dagger.cs
```diff
@@ -8,7 +8,7 @@ public static class Dagger
         new(() => new Query(QueryBuilder.Builder(), new GraphQLClient()));
 
     // <summary>
-    // Connect to the Dagger Engine.
+    // Get a Query instance to start building a dag.
     // </summary>
-    public static Query Connect() => Query.Value;
+    public static Query Dag() => Query.Value;
```
pjmagee, line 13: This doesnt actually connect at all. this only returns a lazy evaluated client, which is ready to make a connection, which only occurs at leaf async function calls. `Connect` feels wrong. Maybe `Query` or `Dag` make more logical sense

sdk/dotnet/sdk/Dagger.SDK.Tests/QueryExecutor.cs
```diff
@@ -3,7 +3,7 @@
 namespace Dagger.SDK.Tests;
 
 [TestClass]
-public class EngineTest
+public class QueryExecutor
```
wingyplus, line 6:
```suggestion
public class QueryExecutorTest
```

sdk/dotnet/sdk/Dagger.SDK.SourceGenerator/Code/CodeRenderer.cs
```diff
@@ -351,15 +381,17 @@ private static string RenderArgumentValue(
...
+                case SyntaxKind.StringKeyword:
                     return $"new StringValue({argName})";
-                case "bool":
+                case SyntaxKind.BoolKeyword:
                     return $"new BooleanValue({argName})";
-                case "int":
+                case SyntaxKind.IntKeyword:
                     return $"new IntValue({argName})";
-                case "float":
+                case SyntaxKind.FloatKeyword:
```
wingyplus, line 394: ❤️

sdk/dotnet/sdk/Dagger.SDK.SourceGenerator/Extensions/EnumerableExtensions.cs
```diff
@@ -3,23 +3,23 @@
...
+//         foreach (TSource element in source)
+//         {
+//             if (keys.Contains(keySelector(element)))
+//             {
+//                 continue;
+//             }
+//             yield return element;
+//         }
+//     }
+// }
```
wingyplus, line 25: We don't need it after we upgraded to .NET8.0 right?

wingyplus:
@pjmagee It seems that the test cancellation token is failure when running with `dagger call -m ./dev test` (needs to run under `./sdk/dotnet`). Could you please take a look?

sdk/dotnet/sdk/Dagger.SDK.SourceGenerator/Extensions/EnumerableExtensions.cs
pjmagee: This is what I am having trouble in understanding 100%. .NET8 introduced this, but we made because of netstandard, but netstandard is considered legacy while the docs still recommend netstandard for source generators. I am stuck in the middle of knowing what we should really do to push forward. Can we really forget about netstandard. (I hope we can)

pjmagee:
> TaskCanceledException

OperationCanceledException is the base class..

https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/TaskCanceledException.cs,23)https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/[TaskCanceledException.cs](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/TaskCanceledException.cs,23),23

hmmm

sdk/dotnet/sdk/Dagger.SDK.SourceGenerator/Extensions/EnumerableExtensions.cs
pjmagee: https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1#when-to-target-netx0-vs-netstandard

There is so much explanation discussed here, I really don't like netstandard lol

pjmagee:
> dagger call -m ./dev test

Okay, i ran locally and using `TaskCanceledException` seems to have fixed it and still pass with `dagger run dotnet test...`

sdk/dotnet/sdk/Dagger.SDK.SourceGenerator/Extensions/EnumerableExtensions.cs
pjmagee: I did some more reading, supposedly, Visual Studio may have issues, because Visual Studio is using .NET Framework. This means Source Generator will only work on IDE of VS Code & JetBrains Rider. 

This is actually quite annoying... I am not sure if they have plans for Visual Studio to use the dotnet build pipeline for source generator like vscode.

pjmagee: I think multi target could be an option. `<TargetFrameworks>net8.0;netstandard2.0</TargetFrameworks>` and 

`#if NETSTANDARD2_0` for the `ExceptBy`

pjmagee:
@wingyplus Okay, so one last thing added back was netstandard, but we can multi-target no problem. I tested removing netstandard2.0, Visual Studio didn't want to compile it. Adding `;netstandard2.0` and Visual Studio now is not complaining.

pjmagee:
I have the strangest behaviour its so weird, `System.Text.Json` only works with `8.0.5` not `9.0.0` despite it having no newer requirements for netstandard project. Really weird behaviour to make anything work properly across all ways we compile and use an IDE for source gen

sdk/dotnet/sdk/Dagger.SDK.SourceGenerator/Extensions/EnumerableExtensions.cs
pjmagee: I have put this back, there are some issues and it basically boils down to different IDEs triggering source generation. A bit annoying.

sdk/dotnet/sdk/global.json
```diff
@@ -1,5 +1,9 @@
 ﻿{
   "msbuild-sdks": {
-    "MSTest.Sdk": "3.4.0"
+    "MSTest.Sdk": "3.7.0"
+  },
+  "sdk": {
+    "version": "8.0.101",
```
wingyplus, line 6: It would be better to use `8.0.404` to reflect the latest version from Microsoft.

wingyplus: Done

wingyplus:
@pjmagee I can't find the dotnet version 8.0.101 but 8.0.404 so I modify the global.json file to use that version instead. 

About instrospection.json, I update it to 0.15.1 to reflect with the latest release as well

sdk/dotnet/sdk/Dagger.SDK.SourceGenerator/SourceGenerator.cs
```diff
@@ -93,7 +92,7 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
                         sources[0]!.ToString()
                     )!;
                     string code = codeGenerator.Generate(introspection);
-                    Console.WriteLine(code);
+                    // Console.WriteLine(code);
```
wingyplus, line 95:
```suggestion
```

wingyplus:
> I have the strangest behaviour its so weird, `System.Text.Json` only works with `8.0.5` not `9.0.0` despite it having no newer requirements for netstandard project. Really weird behaviour to make anything work properly across all ways we compile and use an IDE for source gen

Did you encounter this warning?

```
CSC : warning CS8785: Generator 'SourceGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'FileNotFoundException' wit
h message 'Could not load file or assembly 'System.Text.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.'. [C:\Users\WingyMomo\src\github.com
\dagger\dagger\sdk\dotnet\sdk\Dagger.SDK\Dagger.SDK.csproj]
```

After hitting this warning, the `Dagger.SDK` cannot be compiled because the source generator is not working. Is that possibly related to dotnet8 and netstandard2?

wingyplus (approved):
Test with `dagger -m ./dev test` is working as expected. Here is log:

```
PS C:\Users\WingyMomo\src\github.com\dagger\dagger\sdk\dotnet> ~\dagger\dagger -m .\dev call test
✔ connect 0.5s
✔ load module 1.8s

✔ dotnetSdkDev: DotnetSdkDev! 0.6s
✔ .test: Void 40.1s
```

wingyplus:
@pjmagee The DCO check failed because some commits were not signed off. Could you help fix (fix per commit or squash all of them also fine)?

pjmagee:
> @pjmagee The DCO check failed because some commits were not signed off. Could you help fix (fix per commit or squash all of them also fine)?

Yes, i will try find some time this evening to squash and -s. I didn't realise

wingyplus (approved):
The changes look good!

Since I'm not a maintainer of dotnet SDK, I have no right to merge code to the main branch. So let the dotnet maintainer (@jedevc and @helderco) help review it again. 🙇

sdk/dotnet/sdk/Dagger.SDK/introspection.json
jedevc, line 1: I might be missing something, but does this need to be committed?

jedevc (approved):
This LGTM :tada: (aside from the comment on the large `introspection.json`).

We should try and merge this alongside https://github.com/dagger/dagger/pull/8762.

sdk/dotnet/sdk/Dagger.SDK/introspection.json
pjmagee: the json file is required in order to trigger the code generation. I'm not harmful to include in my opinion, though with the dev module we have it can be produced. I'm indifferent with including it to be honest. @wingyplus any strong opinion?

helderco: I don't think it should be committed. It should be generated as an initial step of codegen. The `cmd/codegen` binary can produce this file so the SDKs don't need to do the introspection themselves. For modules support later on, this files is provided by the engine instead.

wingyplus: I think we can remove that file from now. Once this PR merge, the #8762 can implement the generate function to produce introspection file from that function.

wingyplus:
> This LGTM 🎉 (aside from the comment on the large `introspection.json`).
> 
> We should try and merge this alongside #8762.

I need this PR to continue #8762 because c# class on that PR is conflicted with the API (Engine, etc) and this PR fix this issue. 🙇‍♂️

sdk/dotnet/sdk/Dagger.SDK/introspection.json
helderco: Yep 👍 For reference, Python's dev module gets it via function argument:

https://github.com/dagger/dagger/blob/c6daabd262e0f618b8f440261d95a2bf92589e16/sdk/python/dev/src/main/main.py#L134-L137

Which in turn gets it from the root module:

https://github.com/dagger/dagger/blob/c6daabd262e0f618b8f440261d95a2bf92589e16/.dagger/sdk_python.go#L126-L129

pjmagee: Removed

PR #9307, opened 2025-01-03
docs: Document include/exclude fields in dagger.json

vikram-dagger:


TomChv (commented):
Since we now have context directory, I think we should n't document that

@shykes what's your opinion?

vikram-dagger:
> > Since we now have context directory, I think we should n't document that
> 
> They're different use cases. Context directory + ignore is for a function's argument. include/exclude in dagger.json is still needed for filtering what's actually loaded with the module itself. Especially when we don't have automatic `.gitignore` support.

Yes, I agree.

vikram-dagger:
Can I get your approval on this to merge @TomChv @helderco ?

helderco (commented):
Can be useful to clarify that includes are applied before excludes.

vikram-dagger:
> includes are applied before excludes.

Yes, added - thanks!

TomChv (approved):

PR #9308, opened 2025-01-03
docs: Add repository clone recipe using dag.Git

vikram-dagger:


jedevc:
LGTM, but for something that's really just one API call, it feels a bit odd to have a full recipe (with all the complexity of enums, etc).

jedevc (approved):

vikram-dagger:
> LGTM, but for something that's really just one API call, it feels a bit odd to have a full recipe (with all the complexity of enums, etc).

I don't disagree. However we have had a few users asking for this at different times ([example](https://discord.com/channels/707636530424053791/1120503349599543376/1246424825367494686)). and I wanted to showcase all the available methods (branch, tag, commit). I combined them into a single recipe (being more efficient than a separate cookbook recipe for each possibility).

PR #9320, opened 2025-01-06
ci: fix permissions error on pull request comment

jedevc:
Fixes https://github.com/dagger/dagger/actions/runs/12602450696/job/35125546961?pr=9309

```
Run thollander/actions-comment-pull-request@v2
  with:
    message: This PR has been marked with `needs/changelog`, but no changelog has been created.
  
  Run `changie new` to generate one (see [CONTRIBUTING.md](https://github.com/dagger/dagger/blob/main/CONTRIBUTING.md) for details).
  
    GITHUB_TOKEN: ***
    mode: upsert
    create_if_not_exists: true
Error: Resource not accessible by integration
```

If the PR comes from a fork, then we cannot get a write token for pull-requests, so we can't write a comment to the PR.

To be able to get access to this, we need to instead trigger on `pull_request_target`. This uses the job definition from `main` instead.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,13 +1,15 @@
 name: needs/changelog
 
 on:
-  pull_request:
+  # WARNING! pull_request_target is dangerous and *must not* run user code.
+  # See https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/
+  pull_request_target:
     types:
       - opened
       - synchronize
       - labeled
       - unlabeled
-      
+
 permissions:
   contents: read
   pull-requests: write
@@ -19,6 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
+          ref: ${{ github.head_ref }}
           fetch-depth: 0
 
       - name: Fetch refs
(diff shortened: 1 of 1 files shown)


gerhard (commented):
I am not sure about this. My biggest is concern is `pull_request_target` exposing all secrets to the workflow. If someone modifies a workflow, my understanding is that they could gain access to all GitHub Actions secrets. This happened 2 months ago: https://in.relation.to/2024/11/12/ghsl-2024-268/

I really think that we should find a different way of doing this - maybe via a GitHub app / bot with restricted privileges - rather than directly in a GitHub Actions workflow. If we went searching, I am wondering if something like this already exists: https://github.com/marketplace?category=code-review&type=apps

I am OK to close this PR and leave it as a known bug if we feel that it is too much work right now.

jedevc:
> If someone modifies a workflow, my understanding is that they could gain access to all GitHub Actions secrets.

This is correct yes - however, for the workflow to be modified, the modified PR would have be approved + merged. The target workflow is pulled off of `main`, so you can't modify the workflow in the PR - it would still use the `main` workflow.

The security issue (documented thoroughly in https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) is only an issue if the workflow executes arbitrary code from the PR - because we're just executing a diff in this workflow, it should be entirely safe.

PR #9322, opened 2025-01-06
fix(sdk/java): handle & inside javadoc

eunomie:
Javadoc comments are wrapped between HTML `<p>` tags. In this case `&` is not a valid entity and generates the following error:

    error: bad HTML entity
      * <p><br/>Set chart & app version<br/></p>

This happen when you generate the documentation using `mvnw package -Pjavadoc`

This commit escapes the `&` char and also call the `Helpers.escapeJavadoc` method for type and arg descriptions in addition to the already done field one.

---

The fix is quite basic, just for `&`. Maybe we need something more resilient, like to use some kind of escape html, but at least that solves this one issue.

cc @jcsirot

### DIFF
--- a/sdk/java/dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/Helpers.java
+++ b/sdk/java/dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/Helpers.java
@@ -171,6 +171,6 @@ public class Helpers {
 
   /** Fix using '$' char in javadoc */
   static String escapeJavadoc(String str) {
-    return str.replace("$", "$$");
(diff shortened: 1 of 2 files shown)


eunomie:
🤔 
Ok, so this was based on a quite old commit (my bad) and the javadoc generation issue mentioned above is not there anymore, since commit 95cea04729f3c378fe66e3ae237274cd0d3e753b that moved `.dagger/helm.go` to `helm/.dagger/main.go`.

That said, the underlying issue is still there. If we had a `&` in any comment, the javadoc generation will fail without this fix. I also covered type and arg comments.

jedevc (approved):

jedevc:
Thanks @eunomie :tada: LGTM as a pretty simple fix, merged!

Issue #8766, opened 2024-10-23
✨ Support private go dependencies in module development


nadirollo:
### What are you trying to do?

I am developing a module for internal systems developed in go.

To interact with such systems from my dagger module, I need to import private go dependencies.

There isn't currently a way to provide credentials to the Go runtime container in a way that the codegen will be able to pull such private dependencies (`GOPRIVATE` / modified `gitconfig`)

For a bit more context: https://discord.com/channels/707636530424053791/1298541718374715443

### Why is this important to you?

To be able to develop dagger modules for internal systems that provide libraries to interact with them

gtabrar:
`GOPROXY` support could also be an option for us if it didn't require a custom runner, but as it stands, the UX of getting all our engineers to set that up wouldn't be great. This would definitely be preferable!

sipsma:
> cc @sipsma since you worked on this feature in the first place, maybe you envisioned having an engine-agnostic way of doing it eventually?

Yep, mentioned in the description of the issue here: https://github.com/dagger/dagger/issues/6599

SGTM!

jedevc:
Also x-linking https://github.com/dagger/dagger/issues/7787, which feels like a similar thing in terms of wanting to mount the `.gitconfig` over.

nadirollo:
https://github.com/dagger/dagger/issues/8766#issuecomment-2474813161 > For our use case, the issue https://github.com/dagger/dagger/issues/7787 is likely to implement what we need. 

Being able to pass the GOPROXY would allow us to work around the issue (we could set up a proxy to the private git repositories where our go dependencies are).

The engine able to read the `.gitconfig` with URL rewrites will solve it right away, without having to set up any other component in the middle.

PR #9323, opened 2025-01-06
feat(engine): add support for fetching git config

rajatjindal:
Fixes https://github.com/dagger/dagger/issues/8766 by implementing https://github.com/dagger/dagger/issues/7787.


rajatjindal:
I am working through changing the SDK field from being a string to a struct to support the 2nd part of this PR (to pass along goprivate env variable)

rajatjindal:
I have this working end to end. and currently cleaning up the PR and working through test failures. The complete fix for this also needs https://github.com/dagger/dagger/pull/9454 which adds the ability to set environment variables in codegen/runtime container from dagger.json.

the idea is that with this PR, we will pull the selective gitconfig keys from user's configuration and add them to the codegen/runtime container automatically. we are doing selective gitconfig because we don't want to pull all the git-config key/values from user system as they may or may not work on the engine (e.g. a config pointing to a file on the local file system for gitignore). Also we don't want to be in a position where we allowed everything, ran into a problem with some keys but can't revert back because of backward compatibility problems.

engine/session/git.go
```diff
@@ -0,0 +1,266 @@
...
+	// Check if git is installed
+	if _, err := exec.LookPath("git"); err != nil {
+		return newGitConfigErrorResponse(NO_GIT, "Git is not installed or not in PATH"), nil
+	}
+
+	// Ensure no parallel execution of the git CLI happens
+	gitMutex.Lock()
+	defer gitMutex.Unlock()
+
+	// Prepare the git credential fill command
```
jedevc, line 190: nit: wrong comment

core/schema/sdk.go
```diff
@@ -669,6 +688,39 @@ func (sdk *goSDK) baseWithCodegen(
...
+
+	gitconfig, err := bk.GetGitConfig(ctx)
+	if err != nil {
+		return ctr, err
+	}
+
+	// unfortunately git does not support export/import of git config
+	// so we basically have to translate the fetched git config into
+	// git config command for each fetched config.
+	// TODO(rajatjindal): maybe we should do this in the attachable itself?
```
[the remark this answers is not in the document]
rajatjindal: The first version of this PR was doing that, and I changed it to be withExec, 🤣. I will revert that back.

core/integration/module_private_deps_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (PrivateDepsSuite) TestPrivateDeps(ctx context.Context, t *testctx.T) {
+	t.Run("golang", func(ctx context.Context, t *testctx.T) {
+		privateDepCode := `package main
+
+import (
+        "context"
+        "dagger/foo/internal/dagger"
+
+        "github.com/rajatjindal/bkpapi/backend/pkg/api"
```
rajatjindal, line 25: Hi @jedevc, I am currently using one of my private repo for simulating a private dep, but I think we can add a private go module in our [dagger-tests-modules](https://github.com/dagger/dagger-test-modules) repo and use the `private forked` version of that in this test.

core/schema/sdk.go
```diff
@@ -456,8 +537,13 @@ executing the codegen binary inside it to generate user code and then execute
...
 type goSDK struct {
-	root *core.Query
-	dag  *dagql.Server
+	root      *core.Query
+	dag       *dagql.Server
+	rawConfig map[string]interface{}
+}
+
+type goSDKConfig struct {
+	GoPrivate string `json:"goprivate"`
```
jedevc, line 546:
```suggestion
	GoPrivate string `json:"goprivate,omitempty"`
```

core/schema/sdk.go
```diff
@@ -763,6 +858,46 @@ func (sdk *goSDK) baseWithCodegen(
 		},
 	}
 
+	var config goSDKConfig
+	err = mapstructure.Decode(sdk.rawConfig, &config)
```
jedevc, line 862: note: unknown fields should cause an error like for `moduleSDK`. So setting `foobar` as a `config` field should error out here.

core/schema/sdk.go
```diff
@@ -616,6 +702,15 @@ func (sdk *goSDK) Runtime(
...
 	if err := sdk.dag.Select(ctx, ctr, &ctr,
+		dagql.Selector{
+			Field: "withoutUnixSocket",
+			Args: []dagql.NamedInput{
+				{
+					Name:  "path",
+					Value: dagql.NewString("/tmp/dagger-ssh-sock"),
+				},
+			},
+		},
```
jedevc, line 713: :eyes: not sure I get why we're doing this.

core/schema/sdk.go
```diff
@@ -932,3 +1067,133 @@ func (sdk *goSDK) base(ctx context.Context) (dagql.Instance[*core.Container], er
...
+					Name:  "source",
+					Value: dagql.NewID[*core.Socket](sockInst.ID()),
+				},
+			},
+		},
+	}, nil
+}
+
+// TODO: bikeshed on name
+func getSDKConfigSelectors(_ context.Context, config goSDKConfig) []dagql.Selector {
```
jedevc, line 1180: It's private, no bikeshedding required :tada:

core/integration/module_config_test.go
```diff
@@ -524,6 +524,240 @@ func (ConfigSuite) TestDaggerConfig(ctx context.Context, t *testctx.T) {
...
+  "sdk": {
+    "source": "go",
+    "config": {
+      "goprivate": 1234
+    }
+  }
+}`,
+				expectedError: "'GoPrivate' expected type 'string', got unconvertible type 'float64', value: '1234'",
+			},
+		}
```
jedevc, line 563: Add another test for setting an unknown `foobar` settings, we should error on that as well.

core/integration/module_config_test.go
```diff
@@ -524,6 +524,240 @@ func (ConfigSuite) TestDaggerConfig(ctx context.Context, t *testctx.T) {
...
+	barConfig string,
+) *Coolsdk {
+	m.BarConfig = barConfig
+	return m
+}
+		`
+
+		daggerjson := `{
+	"name": "foo",
+	"engineVersion": "v0.16.3-250304150120-33f537f11a03",
```
jedevc, line 645: :eyes: Should this just be `engineVersion` = `v0.16.2`? Don't think we need the dev versions here.

core/integration/module_private_deps_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type PrivateDepsSuite struct{}
+
+func TestPrivateDeps(t *testing.T) {
+	testctx.New(t, Middleware()...).RunTests(PrivateDepsSuite{})
+}
+
+func (PrivateDepsSuite) TestPrivateDeps(ctx context.Context, t *testctx.T) {
```
jedevc, line 17: This can just be part of `ModuleSuite`.

core/schema/sdk.go
rajatjindal: I need to double check this one. I think I cleanedup all the weird changes I had to do to make ssh auth sockets work, and this was required (and also because Erik suggested we should remove the socket from runtime). I will double check and confirm before tagging Erik for his feedback.

rajatjindal: I revisited my notes, and here is the comment from Erik where he suggested this to fix socket not found issue. His suggestion was both from the fix of the issue as well as from security perspective like commented here: https://github.com/rajatjindal/dagger/blob/6ce731c74c119cd83923b7c221f414f25cd7ef06/core/schema/sdk.go#L538-L538

https://discord.com/channels/707636530424053791/1342188325712429187/1342252739551297798

rajatjindal:
Hi @jedevc, this PR is ready for final set of review/merge. When you get chance, could you please look into this. thank you.

core/integration/private_key_ro_dagger_modules_test_github.pem
jedevc, line 1: Why does this change? :eyes:

core/modulesource.go
```diff
@@ -62,7 +62,8 @@ func (proto ModuleSourceKind) HumanString() string {
 }
 
 type SDKConfig struct {
-	Source string `field:"true" name:"source" doc:"Source of the SDK. Either a name of a builtin SDK or a module source ref string pointing to the SDK's implementation."`
+	Source string                 `field:"true" name:"source" doc:"Source of the SDK. Either a name of a builtin SDK or a module source ref string pointing to the SDK's implementation."`
+	Config map[string]interface{} `name:"config" doc:"sdk specific config"`
```
jedevc, line 66:
```suggestion
	Config map[string]interface{}
```

I don't think this needs these if it's not a `field`.

core/schema/sdk.go
```diff
@@ -932,3 +1097,132 @@ func (sdk *goSDK) base(ctx context.Context) (dagql.Instance[*core.Container], er
...
+			},
+		},
+	}
+
+	gitconfig, err := bk.GetGitConfig(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, entry := range gitconfig {
```
jedevc, line 1128: Is there a way to avoid this being in lots of little `WithExec`s? I think I mentioned this before in https://github.com/dagger/dagger/pull/9323#discussion_r1950646188.

jedevc (approved):
Couple comments, otherwise LGTM :tada:

core/schema/sdk.go
rajatjindal: I did tried it. It seems like not... (unless we take complete git config as is).  once we fetch config using `git config -l`, we cannot translate it back to a working gitconfig file unless we call `git config` on those retrieved values again.

jedevc: :thinking: is it not just as simple as appending to the gitconfig file manually? Or is that just too fundamentally complex?

The other option would be to combine all of these by passing them to a `sh` and joining them with `&&`.

jedevc: Potentially https://github.com/go-git/go-git/blob/main/plumbing/format/config/common.go might also be useful here?

rajatjindal: yep, this seems useful to me.

rajatjindal: i tried this one. it seems it does not support merging git configs (e.g. system level vs user level vs repo level).

although when running `git config -l` through session attachable, it does not consider repo level config either.

another thing to notice is that we have a safe list of git config keys allowed right now, which is just one, so having withExec is not going to be a problem unless we allow a large number of keys (or remove that check) from here.

jedevc: Ah, yes, I'd forgotten we only allow the url transformation.

jedevc:
Okay, @rajatjindal merging this now. Ideally we'd also have some docs for this as well - not sure exactly where these should go, maybe into https://docs.dagger.io/api/remote-modules?

It's related to the concept of remote modules, but it's about private deps. Perhaps @grouville and @vikram-dagger can advise here.

rajatjindal:
This PR cover a few changes:

1. Add support to fetch gitconfig from user's host system.
2. Add support for SDK Config in the engine.
3. Add support for configuring GOPRIVATE env variable (by leveraging the new ^^ sdk config feature) for Go SDK


#### Add support to fetch gitconfig from user's host system.

1. Currently we only fetch use this url instead of this type gitconfig.
2. We decided not to fetch everything from user's host gitconfig as it can have user's system specific config which can break the engine/module and trigger unexpected behavior and inconsistencies.
3. We can easily add more gitconfig keys if a user requests in future.
4. We need  git to be installed on the host system for this to work, but it is optional. We ignore if git is not installed.


#### Add support for SDK Config in the engine.

1. We added a new field sdk.config in dagger.json.
2. This is an object (map[string]interface{})
3. moduleSDK can define a WithConfig function. If the function exists, we verify the arguments against the keys specified in sdk.config object.
4. We error out if there is a key in sdk.config, which is not supported by SDK.
5. We use default value (as defined in WithConfig function) by doing introspection on it IF no override is provided in sdk.config object in dagger.json
6. The support for this is already added in Go SDK (currently only has one config `goprivate`, which sets the `GOPRIVATE` env variable in the codegen container.
7. I have some changes in a branch for typescript SDK, but it needs more changes. I will talk a bit more about it in one of the next section.

#### Add support for private dependencies in golang sdk

1. We leverage the above two features to support private deps in golang based modules.
2. The user can now set sdk.config.goprivate in dagger.json, which gets translated to GOPRIVATE env variable in the codegen container.

#### Add support for custom config in typescript sdk

1. This is a WIP right now as it involves quite a bit of refactoring of typescript SDK to ensure the sdk.config could be used in the SDK.
2. Some of the changes involve the configs that we pass using package.json properties today. e.g. runtime etc.
3. For this to work we need to restructure the sdk so that
4. sdk provides a base image (currently this requires some config from package.json)
5. we add our customizations on top of that base image (e.g. SSH Auth Sock and configs from sdk.config
6. The sdk then runs codegen on that container and use that as runtime.
7. I might be oversimplifying the steps here, but that is basically it.
8. I have a WIP branch that I will rebase and cleanup and open PR for review.

PR #9327, opened 2025-01-07
add statuses API + Go/Python/TypeScript SDK support

vito:
### Goals

* [x] High level API for OTel spans (called "statuses" in our API)
  * [x] Add withFoos for customizing the span
* [x] Ergonomic SDK APIs, without too much effort
  * This is all OTel under the hood so this feels like "progressive enhancement"
  * [x] Go
  * [x] Python
  * [x] TypeScript
* [x] Figure out what other high level APIs we might want
  * [x] Global spans - if any are present, we prioritize showing them over the "lower-level" full trace
    * This turned into the "reveal" mechanism
  * [ ] ~~Global logging? Or do we just do spans?~~ - let's do logging later
  * [ ] ~~Logging to a span?~~ - let's do logging later
* [x] Support rich UI controls
  * [x] Icons/emojis?
  * [x] Existing attributes? ~~(internal, encapsulate, encapsulated)~~ reveal/passthrough
  * [ ] ~~Expanded vs. collapsed by default?~~ haven't needed this, 'reveal' covers it?

### Questions/TODOs

* [x] Do we make a point to _hide_ OTel "span" terminology, or is it OK to leave it exposed as an advanced API alongside simpler ones ("task")?
  * June 27 update: yep - "statuses"
* [x] Audit creating nested spans with the same name - I think this breaks atm, too reliant on DagQL cache
* [x] Go SDK should have a pattern for bubbling up errors to spans

### Examples

See the `viztest/` changes for examples in Go, Python, and TypeScript.


cwlbraa:
looking at the example traces, i can't tell– does this approach lose the ability for these traces to parent to a normal trace if users want to display them "normally" ? or is the idea that the CLI prioritizing showing this class of traces over the big ones and links to both?

beingminimal:
@vito Please reopen this PR if this should be supported. Thanks

nipuna-perera:
I would still love to see this! Explicit outputs for end users is still a pain point with TUI default output and the default otel span API is not very ergonomic. Even so, the display result in the TUI is still very crowded and the custom spans are hard to organize and easy to miss.

codethief:
I agree with @nipuna-perera: The default TUI output is incredibly painful. Using `opentelemetry` spans does help a bit but it feels… tacked on, instead of being an integral aspect of Dagger.

nipuna-perera:
Noooooooooo

Nipuna Perera
Software Developer Engineer
817-739-2137

vito:
Thanks for performing CPR on this PR everyone :smile: - I rebased it yesterday but it looks like there's still some bit-rot to figure out. Will get back to this when there's time.

Also, just a note, I'll probably just rename `Span` to `Status` everywhere in the API, but the behavior would remain the same. cc @shykes

vito:
Think this is ready for review/bikeshedding now.

:heavy_check_mark: Span => Status everywhere
:heavy_check_mark: Add 'display()' method for immediatly starting+stopping the span, using it to just display a message
:heavy_check_mark: Updated viztest as appropriate
:heavy_check_mark: Checks are green

docs/docs-graphql/schema.graphqls
```diff
@@ -3579,6 +3582,9 @@ type Query {
     requireKind: ModuleSourceKind
   ): ModuleSource!
 
+  """Returns a status that reveals its child statuses and hides itself."""
+  reveal: Status!
```
vito: Woops, don't know why that's here - fixing!

shykes:
Bikeshed patrol reporting for duty...

@vito do you have an example code snippet anywhere? To show how it would look when using it?

vito:
@shykes You can find examples under `dagql/idtui/viztest/` for Go, Python, and TypeScript, e.g.:

https://github.com/dagger/dagger/pull/9327/files#diff-cf1e6cc74aade66b528a7d3c9fdaa760bcc1e89ac851e582a46fad7aa3287574

```go
func (v *Viztest) CustomStatus(ctx context.Context) (res string, rerr error) {
	ctx, status := dag.Status("custom status").Context(ctx)
	defer status.End(ctx)
	return v.Echo(ctx, "hello from Go! it is currently "+time.Now().String())
}

dag.Status("custom status").Display(ctx)

dag.Status("failing outer status").Run(ctx, func(ctx context.Context) error {
	// ...
	return errors.New("outer failure")
})
```

```python
    @function
    async def custom_status(self) -> str:
        async with dag.status("custom status"):
            return await self.echo(f"hello from Python! it is currently {now}")
```

```typescript
  @func()
  async customStatus(): Promise<string> {
    return dag.status("custom status").run(async () => {
      return this.echo(`hello from TypeScript! it is currently ${now}`)
    })
  }
```

core/telemetry.go
```diff
@@ -220,7 +220,7 @@ func isIntrospection(id *call.ID) bool {
 // isMeta returns true if any type in the ID is "too meta" to show to the user,
 // for example span and error APIs.
 func isMeta(id *call.ID) bool {
-	if anyReturns(id, "Error") {
+	if anyReturns(id, "Error") || anyReturns(id, "Status") {
```
sipsma:
```suggestion
	if anyReturns(id, "Error", "Status") {
```

core/status.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (s Status) Clone() *Status {
+	cp := &s
+	cp.Query = cp.Query.Clone()
+	return cp
+}
+
+func (s *Status) WithActorEmoji(actor string) *Status {
+	cp := s.Clone()
+	cp.ActorEmoji = actor
```
sipsma, line 40: nit: since the arg is just a string, would anything break in the TUI/WebUI if a user provided something silly like `"weird\nactor\n\nstring\n\n\n\n"`. Basically just wondering if we want any validation here.

If not and this really does support any arbitrary string, I wonder if having this be named w/ `Emoji` makes sense. But this is splitting micro-hairs, I get the *intention* is that the user provides an emoji here.

core/status.go
```diff
@@ -0,0 +1,99 @@
...
+
+type Status struct {
+	Name string `field:"true" doc:"The display name of the status."`
+
+	ActorEmoji  string
+	Message     string
+	Reveal      bool
+	Passthrough bool
+
+	Query *Query
```
sipsma: nit: can avoid this now and grab `Query` from context instead. If you don't do it here I'll end up making the change though for the persistent cache 😄

core/query.go
```diff
@@ -25,6 +26,9 @@ import (
 // dependencies for evaluating queries.
 type Query struct {
 	Server
+
+	spans  map[string]*Status
+	spansL *sync.Mutex
```
sipsma: Can this be state on the `daggerSession` (or `daggerClient`) instead? And then exposed via the `Server` interface.

It's just tough to deal with this type of state on API objects (including `Query`) in the world of persistent caching of said objects. So putting it on the "ephemeral session state" instead that's not tied into all of that makes things a lot easier.

core/status.go
vito: Yeah validation is a good idea actually. If you set something other than an emoji the UI will definitely break lol.

Tricky bit will be testing that it's a single emoji, but it looks like I can use this: https://pkg.go.dev/github.com/rivo/uniseg#GraphemeClusterCount

That'll still accept single letters, like `A`, but maybe that's OK.

docs/docs-graphql/schema.graphqls
```diff
@@ -3808,6 +3819,50 @@ The `SourceMapID` scalar type represents an identifier for an object of type Sou
 """
 scalar SourceMapID
 
+"""A status indicator to show to the user."""
+type Status {
+  """
+  Start and immediately finish the status, so that it just gets displayed to the user.
+  """
+  display: StatusID!
```
shykes, line 4697: This confused me. I *think* it's supposed to be a convenience to avoid `.start().end()` when the status is just a message instead of a task with a beginning and an end?

In that case why not just make it so `end()` implies `start()`? Then you can just call `end()`.

docs/docs-graphql/schema.graphqls
```diff
@@ -3808,6 +3819,50 @@ The `SourceMapID` scalar type represents an identifier for an object of type Sou
...
+  name: String!
+
+  """Start a new instance of the status."""
+  start: StatusID!
+
+  """Set an emoji representing the actor of the status."""
+  withActorEmoji(actor: String!): Status!
+
+  """Hide the status itself, and reveal its children."""
+  withPassthrough: Status!
```
shykes: What is this for?

docs/docs-graphql/schema.graphqls
```diff
@@ -3808,6 +3819,50 @@ The `SourceMapID` scalar type represents an identifier for an object of type Sou
...
+  """A unique identifier for this Status."""
+  id: StatusID!
+
+  """Returns the internal ID of the status."""
+  internalId: String!
+
+  """The display name of the status."""
+  name: String!
+
+  """Start a new instance of the status."""
```
shykes, line 4715: "start a new instance" confused me. Does this mean I can call `start()` several times, and each time it's a new separate status that I have to `end()` individually? *Should* I do this?

docs/docs-graphql/schema.graphqls
```diff
@@ -3808,6 +3819,50 @@ The `SourceMapID` scalar type represents an identifier for an object of type Sou
...
+
+  """Hide the status itself, and reveal its children."""
+  withPassthrough: Status!
+
+  """
+  Indicates that the status represents a received message.
+  
+  The message body must be sent as logs, so that it can be streamed. The name of the status is ignored.
+  """
+  withReceivedMessage: Status!
```
shykes: What does "received" mean in this context? Who is receiving the message from whom?

docs/docs-graphql/schema.graphqls
vito: Yep - this was a late addition for convenience for that scenario. We _could_ make `end` imply `start` but I thought it looked a bit weird:

```go
// is this ending a previously created "foo" status? what is "ending" exactly?
dag.Status("foo").End(ctx)
dag.Status("foo").Display(ctx)
```

It could also potentially hide mistakes, like this:

```go
status := dag.Status("foo")

// ... 10 minute task ...

// this has a duration of 0ms
status.End(ctx)
```

> Caveat: a lot of Go code might habitually the `error` return value there anyway, but still true for other SDKs

I'm also happy to just remove this, was trying to find a middle ground for status reporting but maybe it's better not to overload the timing aspect with displaying messages?

Low conviction here, happy to adjust

docs/docs-graphql/schema.graphqls
vito: tl;dr: this is one of our core primitives for controlling what spans show up where, without requiring widespread changes in your instrumentation stack.

This is the mechanism that allows you to "promote" a span's of child spans up one level, replacing the `passthrough` span inline, without having to modify each of their attributes (which is sometimes out of your control).

So with a span hierarchy like this:

* Parent
  * Child
    * Grandchild 1
    * Grandchild 2
    * Grandchild 3
  * Child 2

If you set `passthrough: true` on `Child`, you'll get this:

* Parent
  * Grandchild 1
  * Grandchild 2
  * Grandchild 3
  * Child 2

In practice this is used to hide noisy intermediate/aggregating spans like `POST /query` without hiding their children:

* `dagger call unit`
  * `POST /query`
    * `MyModule.unit`
      * `POST /query`
        * `Container.stdout`

With `passthrough: true` on the `POST /query` spans, you get:

* `dagger call unit`
	* `MyModule.unit`
	  * `Container.stdout`

It can also be composed with `reveal: true` to not only substitute the `passthrough` span with its child spans, but also hoist those child spans up to the toplevel, again without having to modify each child span:

* Parent (top-level span)
  * Child
    * Grandchild 1
    * Grandchild 2
    * Grandchild 3
  * Child 2

With both `passthrough` and `reveal` applied to `Child`, you get:

* Parent (top-level span)
  * Grandchild 1
  * Grandchild 2
  * Grandchild 3

docs/docs-graphql/schema.graphqls
vito: Yes - because a `Status` follows the builder pattern (`dag.Status("foo").WithPassthrough()`), we can't just start it immediately, so there's an explicit `start` call.

It's also totally legit to start the same status multiple times. You might be running something in a loop for example. With OTel at least you don't put an index/offset in the span name; they're not meant to be unique, they're only meant to identify the task being performed, i.e. a low cardinality name that can be aggregated.

docs/docs-graphql/schema.graphqls
vito: This is used to hint to the UI that the status represents a message "received" by the user, i.e. from an LLM response. Theoretically the UI could use this distinction to render "Sent" messages aligned to the right and "Received" messages aligned to the left, but we didn't end up doing that in our UIs.

image

Under the hood this maps to `dagger.io/ui.message=received`, and 'sent' would map to `dagger.io/ui.message=sent`, but it looks like I didn't add an API for that.

We might not need to surface this up to the API, so I'm happy to shelve it for now. The only catch is it's used in our TUI test suite to test Markdown rendering, but I can just use the OTel SDK directly instead. If we remove it we might want to also nix `WithActorEmoji`, though I think the TUI also respects that for non-message spans, so maybe there's still some utility to it.

core/status.go
vito: done!

core/query.go
vito: moved to `daggerClient`

We could consider moving it to `daggerSession` in the future if there comes a use case for passing Statuses between clients, but that seems sort of unusual from an OTel POV, so didn't go for it.

core/status.go
vito: (This has become moot now that I've disabled these APIs - resolving)

vito:
@shykes Went ahead and narrowed the scope dramatically for now in hopes of getting this in with less uncertainty.

* Marked the API experimental.
* Removed all the `withFoo` modifiers - we can add these in response to user requests and perhaps leave the current attributes as implementation details instead of exposing them directly.

(No idea why the docs check is failing, it passes locally and the diff looks empty in CI.)

vito:
Oh, euch, [this failure](https://dagger.cloud/dagger/traces/628a1f09c8e1e8817ae1991695fc3b22?span=c639fde8e24466a1) is legit - it must have something to do with the test using `Status` as the name for its local type.

I was initially worried about a name conflict, but this _should_ be fine, since the module's status will end up as `FooStatus`.

It's maybe a little odd that this code path checks dependencies before checking the module when looking up a type:

https://github.com/dagger/dagger/blob/1a37a08ab17395b5622da949c1f03eaaec2f2c48/core/module.go#L270-L274

But it's even odder that this only fails for TypeScript. :thinking: The modules look equivalent to me in terms of what types + fields they expose:

https://github.com/dagger/dagger/blob/1a37a08ab17395b5622da949c1f03eaaec2f2c48/core/integration/legacy_test.go#L1037-L1148

@TomChv Any idea what might be different in the TypeScript SDK here?

@jedevc Would it make sense to swap the order of all those branches to check the local module first? That feels more right to me, but I could see it being a moot point if e.g. the type is supposed to be namespaced by then.

jedevc:
> Would it make sense to swap the order of all those branches to check the local module first? That feels more right to me, but I could see it being a moot point if e.g. the type is supposed to be namespaced by then.

Yeah, that does seem right to me actually :thinking: if we can switch them, we should.

nipuna-perera:
@vito any updates on this?

vito:
@TomChv Does this fix seem sensible to you? [`d785be1` (#9327)](https://github.com/dagger/dagger/pull/9327/commits/d785be1ef0b2f1b7c185c8909045e3e07e05bfe1)

The issue was that we had tests that define their own `Status` as an enum type. Go and Python correctly handled this by allowing the module to shadow it, but the TypeScript SDK was getting confused, prioritizing the `Status` core object type over the locally defined enum type. I chucked it into a coding agent and it found a fix, and the tests pass now, but want to make sure it's a reasonable change or if there's a better approach.

[Here's the AI trace](https://dagger.cloud/vito/traces/9759c0231935090228dfb2a9e37ec808) (a bit hard to read - need to catch the web UI up to the latest TUI work...).

TomChv:
> @TomChv Does this fix seem sensible to you? [`d785be1` (#9327)](https://github.com/dagger/dagger/pull/9327/commits/d785be1ef0b2f1b7c185c8909045e3e07e05bfe1)
> 
> The issue was that we had tests that define their own `Status` as an enum type. Go and Python correctly handled this by allowing the module to shadow it, but the TypeScript SDK was getting confused, prioritizing the `Status` core object type over the locally defined enum type. I chucked it into a coding agent and it found a fix, and the tests pass now, but want to make sure it's a reasonable change or if there's a better approach.
> 
> [Here's the AI trace](https://dagger.cloud/vito/traces/9759c0231935090228dfb2a9e37ec808) (a bit hard to read - need to catch the web UI up to the latest TUI work...).

I read the full AI trace and that's actually very good! The changes are correct!

Btw is your `modules/coder` available? It looks very useful :)

vito:
@TomChv Sweet, thanks for confirming.

> Btw is your `modules/coder` available? It looks very useful :)

Just pushed a coding agent entrypoint here: https://github.com/dagger/dagger/pull/11199

It's a bit different from `modules/coder` since it just uses our main `.dagger` module as-is, whereas `modules/coder` had a simpler API for running tests, but it _should_ work albeit with some extra hops (`test > specific` instead of just `test` with args).

PR #9330, opened 2025-01-07
Alternative CI Runners 3 - Magicache benchmark

marcosnils:
- **ci: add magicache benchmark**
- **trigger workflow**
- **ci: add scheduled workflow run for magicache benchmark**

### DIFF
--- a/.github/workflows/alternative-ci-runners-3.yml
+++ b/.github/workflows/alternative-ci-runners-3.yml
@@ -1,0 +1,45 @@
+name: Alternative CI Runners 3
+
+on:
+  # Run the workflow every day TWICE:
+  # 1. 9:06AM UTC (low activity)
+  # 2. 9:26AM UTC (cache test - high chance of no code changes)
+  schedule:
+    - cron: "6,26 9 * * *"
+  # Enable manual trigger for on-demand runs - helps when debugging
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  docs-lint-on-magicache:
+    uses: ./.github/workflows/_dagger_on_magicache.yml
+    with:
+      function: docs lint
+      timeout: 20
+
+  test-cli-engine-on-magicache:
+    needs: docs-lint-on-magicache
+    uses: ./.github/workflows/_dagger_on_magicache.yml
+    with:
+      function: test specific --run='TestCLI|TestEngine' --race=true --parallel=16
+      timeout: 20
+
+  sdk-go-on-magicache:
+    needs: docs-lint-on-magicache
+    uses: ./.github/workflows/_dagger_on_magicache.yml
+    with:
+      function: check --targets=sdk/go
+
+  sdk-python-on-magicache:
+    needs: docs-lint-on-magicache
+    uses: ./.github/workflows/_dagger_on_magicache.yml
+    with:
+      function: check --targets=sdk/python
+
+  sdk-typescript-on-magicache:
+    needs: docs-lint-on-magicache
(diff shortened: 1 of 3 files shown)


marcosnils:
@gerhard this should be merged after the magicache runners are set in our CI infra.

gerhard:
OK, checking both out now.

.github/workflows/alternative-ci-runners-3.yml
```diff
@@ -0,0 +1,45 @@
...
+    with:
+      function: docs lint
+      timeout: 20
+
+  test-cli-engine-on-magicache:
+    needs: docs-lint-on-magicache
+    uses: ./.github/workflows/_dagger_on_magicache.yml
+    with:
+      function: test specific --run='TestCLI|TestEngine' --race=true --parallel=16
+      timeout: 20
```
gerhard, line 27: `test-cli-engine` is most resource intensive. It assumes & uses 16 CPUs. FTR:
image

We should choose one of the following options:
- run on 16CPUs
- remove it

.github/workflows/alternative-ci-runners-3.yml
```diff
@@ -0,0 +1,45 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  docs-lint-on-magicache:
+    uses: ./.github/workflows/_dagger_on_magicache.yml
+    with:
+      function: docs lint
+      timeout: 20
```
gerhard, line 20: While `docs-lint` caches well, it is known to use up to 8CPUs (this instance has 16CPUs):
image

Only provisioning 4CPUs might make it run artificially slower.

.github/workflows/alternative-ci-runners-3.yml
```diff
@@ -0,0 +1,45 @@
...
+    needs: docs-lint-on-magicache
+    uses: ./.github/workflows/_dagger_on_magicache.yml
+    with:
+      function: check --targets=sdk/python
+
+  sdk-typescript-on-magicache:
+    needs: docs-lint-on-magicache
+    uses: ./.github/workflows/_dagger_on_magicache.yml
+    with:
+      function: check --targets=sdk/typescript
```
gerhard, line 45: While these jobs cache well, we should provision them with at least 8CPUs, since they are known to be CPU intensive (same as the other jobs).

.github/workflows/alternative-ci-runners-3.yml
gerhard, line 1: We should add `Alternative CI Runners 3` to `trace-workflows.yml` if we want to track the metrics of this workflow alongside the other two.

gerhard (commented):
This is almost there, excited to finish & merge this.

.github/workflows/alternative-ci-runners-3.yml
gerhard: Fixed by running on 16c runners **only** - same as all other alternative workflows.

.github/workflows/alternative-ci-runners-3.yml
gerhard: Fixed by running on 16c runners **only** - same as all other alternative workflows.

.github/workflows/alternative-ci-runners-3.yml
gerhard: Fixed by running on 16c runners **only** - same as all other alternative workflows.

gerhard (approved):
All looking good now.

Going to run one manual test before merging both PRs (incl. the private one).

gerhard:
🚀
image

PR #9334, opened 2025-01-07
fix: avoid restoring stdout after making stdin raw

jedevc:
This just looks like a typo from #9244 - I don't think we actually intend to do this?

### DIFF
--- a/dagql/idtui/frontend_pretty.go
+++ b/dagql/idtui/frontend_pretty.go
@@ -747,12 +747,11 @@ func (fe *frontendPretty) update(msg tea.Msg) (*frontendPretty, tea.Cmd) { //nol
 				ExecCommand: cmd,
 				before: func() error {
 					if stdin, ok := fe.stdin.(*os.File); ok {
-						ttyFd := int(os.Stdout.Fd())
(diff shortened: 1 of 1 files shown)


rajatjindal:
Hi @jedevc,

I am still trying this out (but my windows laptop is crashing on me when trying to compile dagger-client, which is slowing me down).

I did a test by compiling dagger-client on my macbook for my windows host, and while the `./dagger/dagger-client core container from --address ubuntu terminal` works fine, **_exiting_** from `echo ".core | container | from alpine | terminal" | dagger shell` fails with "terminal session failed. The handle is invalid".

rajatjindal:
> @rajatjindal did you manage to confirm what might be happening here?

not yet, but this is on my to-get-done list today.

jedevc:
Further discussion on discord: https://discord.com/channels/707636530424053791/1327270167960617003/1328311674230542369

rajatjindal:
> Further discussion on discord: https://discord.com/channels/707636530424053791/1327270167960617003/1328311674230542369
> 
> Can confirm, we need this change to make #9244 work fully on windows.

+1

PR #9336, opened 2025-01-07
Update chaining docs

thommahoney:
While technically true, the implication is that the produced tarball is gzip-compressed, which it is not! This updates the docs to clarify that the file produced is a raw `tar` archive.


vikram-dagger (approved):
Thank you for the correction @thommahoney !

PR #9337, opened 2025-01-08
ci: un-pin php sdk to use same version as engine tag

jedevc:
We were pinned to an old version of the php sdk that didn't have the proper OTEL work that allowed us to actually observe errors happening.

By unpinning it, we'll always get the same version as the engine.

### DIFF
--- a/cmd/init/main.go
+++ b/cmd/init/main.go
@@ -4,6 +4,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


charjr (approved):
LGTM

wingyplus (approved):
Nice catch! ❤️

jedevc:
Oops, the php failure is entirely legitimate:

```
panic: exec: "phpcs": cannot run executable found relative to current directory                               
                                                                                                              
goroutine 1 [running]:                                                                                        
main.main()                                                                                                   
        /app/cmd/init/main.go:109 +0x8a9                  
```

This is due to a relative path lookup - looks like we need some extra logic since changing our init system in https://github.com/dagger/dagger/pull/9262 (cc @sipsma)

sipsma (approved):

dragoonis:
Merge it!!!

PR #9310, opened 2025-01-03
docs: Add daggerverse documentation

kpenfound:
Docs for the Daggerverse!

PR #9339, opened 2025-01-08
docs: Add note for Python module examples

vikram-dagger:
Follow up to https://github.com/dagger/dagger/pull/9310#issuecomment-2574593854


kpenfound (approved):
Yeah this is accurate enough. The double underscore is a separator for the module+function the example is for and the example name. For unnamed examples (the default example shown for a function), double underscores wouldn't be used

PR #9340, opened 2025-01-08
ci: use dagger-built binaries in goreleaser

jedevc:
Internal ticket: https://linear.app/dagger/issue/DEV-4839

It's better to use the dagger-built binaries as part of our goreleaser release, instead of having goreleaser build them itself. This has a few advantages:

- Consistency. The fewer process differences between local builds and release builds, the better.
- Extensibility. At some point in the future, we want to bundle the engine *into* the CLI - this will require the CLI release build to use dagger, instead of being a simple `go build` command that goreleaser can do.
- Security. By splitting out the building and releasing steps, we're now no longer beholden to the release cadence of goreleaser before we can upgrade go versions.

### DIFF
--- a/.dagger/build/builder.go
+++ b/.dagger/build/builder.go
@@ -128,10 +128,6 @@ func (build *Builder) WithGPUSupport() *Builder {
 	return &b
 }
 
-func (build *Builder) CLI(ctx context.Context) (*dagger.File, error) {
-	return build.binary("./cmd/dagger", true, build.race), nil
-}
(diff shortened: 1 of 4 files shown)


gerhard (approved):
LGTM!

image

image

jedevc:
Success! https://github.com/dagger/dagger/actions/runs/12694387056/job/35384049550#step:4:7392

The latest CLI is now being published using this!

PR #9347, opened 2025-01-09
feat: typescript SDK interface support

TomChv:
### Summary

Add Interface Support for Typescript! I already did some manual tests and works great!
There's no need to update the codegen since most of the job regarding interface is done by the server.

Doc done at https://github.com/dagger/dagger/pull/9353

### Technical details

Compared to Go, supporting interface is a bit harder because I do not generate self module binding.
When you create an interface in Go, example:

```go
// my-module/main.go
type Duck interface {
	DaggerObject

	Quack(ctx context.Context) (string, error)
	SuperQuack(ctx context.Context) (string, error)
}
```

This will generate bindings on its module codegen to call the given implementation through API call:

```go
// my-module/dagger.gen.go
func (r *duckImpl) Quack(ctx context.Context) (string, error) {
	if r.quack != nil {
		return *r.quack, nil
	}
	q := r.query.Select("quack")
	var response string
	q = q.Bind(&response)
	return response, q.Execute(ctx)
}

func (r *duckImpl) SuperQuack(ctx context.Context) (string, error) {
	if r.superQuack != nil {
		return *r.superQuack, nil
	}
	q := r.query.Select("superQuack")
	var response string
	q = q.Bind(&response)
	return response, q.Execute(ctx)
}
```

**BUT** we cannot do that in Typescript so I did a dynamic wrapper called `InterfaceWrapper` to perform the same job as these module bindings.
So when a dagger function receive an interface as parameter, it will receive the interface identifier from the Dagger API, this will be passed to the InterfaceWrapper to:
1. Load the interface from its ID with a GQL call -> `loadXXXFromID`
2. Dynamically assign all the function available on this interface to the wrapper with the resolve that will perform GQL request (the module binding I talk about before)
3. Return this InterfaceWrapper as parameter to the executed function (thanks to Typescript non-strict type system, it will work without problem)

### TODOS
- [x] Add interface registration support
- [x] Add interface argument loading through a dynamic wrapper.
- [x] Update TypeScript unit test
- [x] Add integrations tests
- [x] Add doc (done at https://github.com/dagger/dagger/pull/9353)

### Examples

**Definition**
```ts
import { func, object } from "@dagger.io/dagger"

export interface Duck {
  quack: () => string
  superQuack: () => Promise<string>
}

@object()
export class Iface {
	@func()
  duckQuack(duck: Duck): string {
    return duck.quack()
  }

	@func()
  async duckSuperQuack(duck: Duck): Promise<string> {
    return await duck.superQuack()
  }
}
```

**Implementation**: it's just a module with method that implements the interface
```ts
import { object, func } from "@dagger.io/dagger"

@object()
export class TsImpl {
  @func()
  async quack(): Promise<string> {
    return "quack"
  }

  @func()
  async superQuack(): Promise<string> {
    return "QUACK"
  }
}

```

**Usage**
```ts
import { dag, object, func } from "@dagger.io/dagger"

@object()
export class Test {
  @func()
  async tsQuack(): Promise<string> {
    // We call the definition module and provide as argument an implementation of this interface
    return await dag.iface().duckQuack(dag.tsImpl().asIfaceDuck())
  }

  @func()
  async tsSuperQuack(): Promise<string> {
    // We call the definition module and provide as argument an implementation of this interface
    return await dag.iface().duckSuperQuack(dag.tsImpl().asIfaceDuck())
  }
}
```

Closes DEV-3687


core/integration/testdata/modules/typescript/ifaces/main_test.go
jedevc, line 1: This should be in typescript right? Just for completeness, and to fully test calling the interfaces.

sdk/typescript/src/module/executor.ts
```diff
@@ -67,3 +92,126 @@ export class Executor {
...
+ *   @func()
+ *   async callFoo(example: Example): Promise<string> {
+ *     // <- This example argument here is actually the Interface Wrapper, and `foo` will,
+ *     // directly execute the API call to reach the given implementation.
+ *     return example.foo()
+ *   }
+ * }
+ * ```
+ */
+class InterfaceWrapper {
```
jedevc, line 118: Lol this is neat. I spent ages trying to work out what was missing from this, since the go version of this *is so complex* - very neat :tada:

TomChv: Thanks :D

jedevc (approved):
Gonna approve this, since it's a fairly simple implementation on the typescript side :tada: (also we have plenty of time to find bugs if there are any)

@TomChv feel free to merge whenever, but can you also update our docs in a follow-up? https://docs.dagger.io/api/interfaces/

TomChv:
> @TomChv feel free to merge whenever, but can you also update our docs in a follow-up? [docs.dagger.io/api/interfaces](https://docs.dagger.io/api/interfaces/)

Already done: https://github.com/dagger/dagger/pull/9353

Issue #9346, opened 2025-01-09
PHP CI flake: Attempt to read property "port" on null


jedevc:
We seem to fairly consistently hit this php test flake.

Examples:
- Recently: https://github.com/dagger/dagger/actions/runs/12690493305/job/35371467890?pr=8730#step:5:14974
- Quite old: https://github.com/dagger/dagger/actions/runs/11478378808/job/31942452004?pr=8763
  - There was a discussion in https://discord.com/channels/707636530424053791/1162025276872609832/1298608832397840507

```
.E.............................................................  63 / 227 ( 27%)
............................................................... 126 / 227 ( 55%)
............................................................... 189 / 227 ( 83%)
......................................                          227 / 227 (100%)

Time: 00:04.782, Memory: 24.00 MB

There was 1 error:

1) Dagger\Tests\Integration\ClientTest::testDirectory
RuntimeException: Cannot fetch informations from process session

/src/sdk/php/src/Connection/ProcessSessionConnection.php:84
/src/sdk/php/src/Client/AbstractClient.php:25
/src/sdk/php/src/Dagger.php:30
/src/sdk/php/tests/Integration/ClientTest.php:17
/src/sdk/php/tests/Integration/ClientTest.php:43

--

1 test triggered 1 PHP warning:

1) /src/sdk/php/src/Connection/ProcessSessionConnection.php:74
Attempt to read property "port" on null

Triggered by:

* Dagger\Tests\Integration\ClientTest::testDirectory
  /src/sdk/php/tests/Integration/ClientTest.php:41

ERRORS!
Tests: 227, Assertions: 231, Errors: 1, Warnings: 1, Deprecations: 1.
```

jedevc:
Going to close this issue, I haven't seen this flake in a while now.

PR #9348, opened 2025-01-09
Attempt to fix PHP flake with more lax session output parsing

jedevc:
This is an attempt to fix the PHP flake from https://github.com/dagger/dagger/issues/9346.

What *appears* to be happening is that despite confirming that the output we read has `session_token` in it, we're somehow decoding a JSON object that doesn't have the `port` property.

Not really sure how this can possibly be happening - but assuming that the PHP library works as intended (not unreasonable), and that the `dagger session` code is good (we don't see any other flakes like this for other SDKs), then something is wrong here. Having mixed progress output into the stdout (which happens when a PTY is used) could *potentially* be the cause here? It would also explain why this seems flaky, since it depends on timing of writing to a buffer.

Anyways, probably worth a go, no other SDK opens a PTY like this.

### DIFF
--- a/sdk/php/src/Connection/CliDownloader.php
+++ b/sdk/php/src/Connection/CliDownloader.php
@@ -16,7 +16,7 @@ class CliDownloader implements LoggerAwareInterface
 {
     public const DAGGER_CLI_BIN_PREFIX = 'dagger-';
 
-    private NullLogger $logger;
+    private LoggerInterface $logger;
 
     public function __construct()
     {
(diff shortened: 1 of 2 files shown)


charjr:
Scanning the symfony docs ([setPty](https://symfony.com/doc/current/components/process.html#using-tty-and-pty-modes), [setInput](https://symfony.com/doc/current/components/process.html#streaming-to-the-standard-input-of-a-process)), it looks like a reasonable change. 

cc @carnage

charjr (approved):
Approving, CI is still passing.

It looks to be a relatively equivalent swap in|out.

If it's still flaking then we'll need to look elsewhere.

PR #9351, opened 2025-01-09
chore: ts sdk simplify codegen ideable loader

TomChv:
💡 This is purely an internal code improvement PR 💡 

Replace the manual ID loading with a call to `client.loadXXXFromID` to remove boilerplate.

Add `copy` method to context.
Simplify codegen and remove big boilerplates specially when loading arrays.

This saves around 400 lines of codegen and even more when dependencies are involved 🚀 

**NOTE**: Originally, we discussed about that idea with @helderco and wanted to move the loading logic inside `Context` but it's not possible since we need to call it from `dag` to do this and it will:
- 1: create a circular dependency issue between codegen & context.
- 2: if used as a library, `dag` is different than the client used by the user. 
So instead, I choose to simplify improve the codegen.


helderco (approved):
LGTM

PR #9356, opened 2025-01-10
Encode release process in `releaser` module

jedevc:
Internal ticket: https://linear.app/dagger/issue/DEV-4841

This starts the work of encoding the high level process of a release entirely in dagger (instead of splitting across `RELEASING.md` and github actions). The idea is to split this out so that the `releaser` module assembles lots of little lego bricks from different components in our CI, and publishes them all, and finally produces a pretty report that we can display in the CI run.

As we start to split more components of our CI out (e.g. into modules for each SDK, etc), then the `releaser` module just needs to hook into those.

See an example report: https://gist.github.com/jedevc/46f05140d35d18d17408f4fc5ec04f91


gerhard (approved):
Changes look great, looking forward to seeing this in action in the next release.

PR #9372, opened 2025-01-13
ci: avoid use of `errgroup.WithContext` in checks

jedevc:
This causes some very hard-to-read CI failures when we use this - this means that instead of waiting for *everything* to finish, we bail quickly on the first error. E.g. when linting, we error out on the first lint failure, instead of waiting for all the checks to pass or fail.

For example, of a hard-to-read failure see https://v3.dagger.cloud/dagger/traces/168ed3748bb12bdb7d858df98a9adb94


gerhard (approved):
LGTM! Reviewed sync.

PR #9379, opened 2025-01-14
fix: Replace custom module in PhpSdkDev

charjr:
Thanks to the `:expect` flag on calls to `withExec`
The custom-module _always-exec_ used in the PhpSdkDev is now obsolete.

This PR removes the depedency on the custom module _always-exec_.

### DIFF
--- a/sdk/php/dev/src/PhpSdkDev.php
+++ b/sdk/php/dev/src/PhpSdkDev.php
@@ -9,6 +9,7 @@ use Dagger\Attribute\DaggerObject;
 use Dagger\Attribute\Doc;
 use Dagger\Container;
 use Dagger\Directory;
+use Dagger\ReturnType;
 use GraphQL\Exception\QueryError;
 
 use function Dagger\dag;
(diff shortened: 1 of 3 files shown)


sdk/php/dev/src/PhpSdkDev.php
```diff
@@ -63,15 +64,19 @@ public function lint(Directory $source): Container
...
     public function format(Directory $source): Directory
     {
-        $result = dag()->alwaysExec()->exec($this->base($source), ['phpcbf']);
+        $result = $this->base($source)->withExec(
+            args: ['phpcbf'],
+            expect: ReturnType::ANY,
+        );
 
-        if (dag()->alwaysExec()->lastExitCode($result) === '3') {
+        if ($result->exitCode() === '3') {
```
jedevc, line 72: `exitCode` returns an `int`: https://github.com/dagger/dagger/blob/504787f4370a4032788161af0defdcb8b64bd17f/sdk/php/generated/Container.php#L156-L156 (unlike `lastExitCode` which returns a `string` https://github.com/charjr/dagger-always-exec/blob/d478c1aaf4c5290926d2d1978d98d9f0054529eb/src/AlwaysExec.php#L61)

jedevc (approved):

sdk/php/dev/src/PhpSdkDev.php
charjr: Good spot! I've fixed this and reversed it to be stricter.

Instead of throwing on `3`. I've changed it to throw on _anything_ that isn't `0, 1, 2`

PR #9382, opened 2025-01-14
Use ClientID for keying started tunnels

jedevc:
Found while working on some improved tests in #9381, when I discovered that some tests were mistakenly missing from `LegacyTest`.

Essentially, what I wanted to test was - `AsService.Up` works, and `Container.Up` works. But in this test, I end up calling these within the *same* session. I think at some point maybe this *used* to work, but now we have this logic: https://github.com/dagger/dagger/blob/504787f4370a4032788161af0defdcb8b64bd17f/core/services.go#L145-L148

The key we use for started services is the session ID - but a started tunnel is local to the client. So the call to `Up` succeeds the second time it's called, but a new tunnel is never actually created.

The solution here is:
- Use the `SessionID` for keying normal Services.
- Use the `ClientID` for keying tunnels (which are client specific).

### DIFF
--- a/core/integration/legacy_test.go
+++ b/core/integration/legacy_test.go
@@ -797,6 +797,23 @@ func (f *Foo) TestServiceUpEntrypoint(ctx context.Context, app *dagger.File) (st
 	if err != nil {
 		return "", err
 	}
+	go ctr.AsService().Up(ctx, dagger.ServiceUpOpts{
+		Ports: []dagger.PortForward{
+			{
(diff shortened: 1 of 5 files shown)


jedevc:
cc @sipsma @vito I think this feels right to me? A bit messy, but can clean that up if this looks *about* right.

TomChv (approved):
LGTM, make sure CI is green and let's go.

PR #9386, opened 2025-01-14
fix(ts-sdk): extend timeout for all tests

TomChv:
Attempt to fixes the flakes around TS test that timeout (example: https://github.com/dagger/dagger/actions/runs/12770326585/job/35594928990?pr=9384)

### DIFF
--- a/sdk/typescript/src/module/introspector/test/scan.spec.ts
+++ b/sdk/typescript/src/module/introspector/test/scan.spec.ts
@@ -123,6 +123,8 @@ ${jsonResult}
 
   describe("Should throw error on invalid module", function () {
     it("Should throw an error when no files are provided", async function () {
+      this.timeout(60000)
+
       try {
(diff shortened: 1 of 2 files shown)


sdk/typescript/src/module/introspector/test/scan.spec.ts
jedevc, line 1: Do we need the `this.timeout` calls if these are now set globally?

TomChv: Not really but for extra safety I would prefer to keep it, I remember having a bunch of weird issue with Mocha and Typescript tools that doesn't apply general config so I would prefer to keep this duplication just in case.

jedevc: Hm, that's very odd - but fair enough.

PR #9395, opened 2025-01-16
feat: add dagop

### LINKED ISSUE #6927: `Container.AsTarball` does not cache as expected
Imagine the below:

```go
	ctr := dag.Container().From("alpine")
	ctr.
		WithMountedFile("/ctr.tar", ctr.AsTarball()).
		WithExec([]string{"ls", "-lh", "/"}).
		Sync(ctx)
```

On the first run, this works as expected:
1. `ctr` is built
2. `AsTarball` triggers an OCI export
3. `WithExec` is run on the result

However, on subsequent runs, this doesn't cache as expected - `AsTarball` is *always* executed, which results in `ls` always being run. I don't observe this change if I comment out the `WithMountedFile` line.

---

Currently, this seems to be one of the main issues that causes slow re-builds of the dagger engine image during development. Because we build the SDKs using this, this means we need to always run the `AsTarball` step, which breaks cache invalidation.

This was surfaced during work on https://github.com/dagger/dagger/pull/6843, where the better progress view means it's actually more obvious what the issue is.

jedevc:
Fixes #6927, enables #9098, and continues the project of rebuilding the Ship of Theseus :tada:

This PR introduces a new way of marrying dagger's dagql and buildkit's LLB. Currently, dagql is used to query the graphql API schema, while LLB is used to query the buildkit dag. dagql is much more flexible and easy to extend, while LLB has significantly better caching behavior.

`DagOp` allows us to combine the best of both worlds. Essentially, it allows us to execute a graphql selection *as a buildkit op* - caching it exactly as a native buildkit operation.

## How it works

- We inject custom "LLB" by piggybacking on the `Description` map of an existing LLB operation. Then during `ResolveOp`, we transform this into a magical `CustomOpWrapper`. This LLB is very hacky, but such is life :shrug: Without entirely ripping out LLB (or forking that package), there's not really a neat way of handling this.
- The `CustomOpWrapper` delegates all of it's operations to `Register`ed `CustomOp`s - at the moment, only `DagOp`.
	- Note: since `CustomOp`s are serialized into a `string`, then the `DagOp` has to be JSON-marshallable. This means we have to do some hackery around sending `ID`s instead of `Selector`s directly.
- The `DagOp` evaluates the query, and gets the result as a root `Directory` - it then evaluates this the `Directory`s ref, and returns it.

This effectively allows the dagql query execution to be cached!

## Performance

To show this off, I've ported `AsTarball` to use this (to fix #6927). Essentially, `AsTarball` attempts to self-call itself *inside* a `DagOp` (using a similar technique to #8724). This now means that if the `Container` receiver hasn't changed, then the `File` result isn't forced to be re-evaluated (which it was before).

Just this simple transformation seems to shave a couple seconds off an engine build with `dagger call engine container`.

Without this patch:

```
$ time dagger call engine container
...
Full trace at https://dagger.cloud/jedevc/traces/85dea4974204db7872fb31f732bbdcb2
dagger call engine container  5.77s user 1.31s system 25% cpu 27.576 total
```

With this patch:

```
$ time dagger call engine container
...
Full trace at https://dagger.cloud/jedevc/traces/b384cb5631d72ebaddab576fae164d24
dagger-with-dev dagger call engine container  6.38s user 1.45s system 30% cpu 25.252 total
```

In the traces, you can see that before, the tar unpacking as part of unbundling the SDK images is uncached, while after, they are:

Before:
image
After:
image

## TODO

- [x] Work out some better serialization for `Selector`s, `ID`s might not be it (#9204 is probably related)
- [x] Ensure we get reasonable purity in `AsTarball`
- [x] Work out how to test this

### DIFF
--- a/core/dagop.go
+++ b/core/dagop.go
@@ -2,61 +2,73 @@ package core
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/dagger/dagger/dagql"
 	"github.com/dagger/dagger/dagql/call"
 	"github.com/dagger/dagger/engine/buildkit"
+	bkcache "github.com/moby/buildkit/cache"
+	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/client/llb"
+	bksession "github.com/moby/buildkit/session"
+	"github.com/moby/buildkit/snapshot"
 	"github.com/moby/buildkit/solver"
+	"github.com/moby/buildkit/worker"
 	"github.com/opencontainers/go-digest"
+	"github.com/pkg/errors"
 )
 
-type DagOp struct {
+func init() {
+	buildkit.RegisterCustomOp(DirectoryDagOp{})
+	buildkit.RegisterCustomOp(RawDagOp{})
+}
+
+// NewDirectoryDagOp takes a target ID for a Directory, and returns a Directory
+// for it, computing the actual dagql query inside a buildkit operation, which
+// allows for efficiently caching the result.
+func NewDirectoryDagOp(ctx context.Context, srv *dagql.Server, id *call.ID, inputs []llb.State) (*Directory, error) {
+	requiredType := (&Directory{}).Type().NamedType
+	if id.Type().NamedType() != requiredType {
+		return nil, fmt.Errorf("expected %s to be selected, instead got %s", requiredType, id.Type().NamedType())
+	}
+	dagOp := DirectoryDagOp{ID: id}
+
+	st, err := buildkit.NewCustomLLB(ctx, dagOp, inputs,
+		llb.WithCustomNamef("%s %s", dagOp.Name(), id.Display()),
+		buildkit.WithPassthrough())
+	if err != nil {
+		return nil, err
+	}
+
+	query, ok := srv.Root().(dagql.Instance[*Query])
+	if !ok {
+		return nil, fmt.Errorf("server root was %T", srv.Root())
+	}
+
+	return NewDirectorySt(ctx, query.Self, st, "", Platform{}, nil)
+}
+
+type DirectoryDagOp struct {
 	ID *call.ID
 }
 
-func DagLLB(ctx context.Context, srv *dagql.Server, sels []dagql.Selector, inputs []llb.State) (llb.State, error) {
-	id, err := srv.SelectID(ctx, srv.Root(), sels...)
-	if err != nil {
-		return llb.State{}, err
-	}
-
-	requiredType := (&Directory{}).Type().NamedType
-	if id.Type().NamedType() != requiredType {
-		return llb.State{}, fmt.Errorf("expected %s to be selected, instead got %s", requiredType, id.Type().NamedType())
-	}
-	op := DagOp{ID: id}
-	return buildkit.NewCustomLLB(ctx, op, inputs, llb.WithCustomName("dagop"), buildkit.WithPassthrough())
+func (op DirectoryDagOp) Name() string {
+	return "dagop.directory"
 }
 
-func (op DagOp) Name() string {
-	return "dagop"
-}
-
-func (op DagOp) Backend() buildkit.CustomOpBackend {
+func (op DirectoryDagOp) Backend() buildkit.CustomOpBackend {
 	return &op
 }
 
-func (op DagOp) CacheKey(ctx context.Context) (key digest.Digest, err error) {
+func (op DirectoryDagOp) CacheKey(ctx context.Context) (key digest.Digest, err error) {
 	return op.ID.Digest(), nil
 }
 
-type dagOpContextKey struct{}
-
-func withDagOpContext(ctx context.Context, op *DagOp) context.Context {
-	return context.WithValue(ctx, dagOpContextKey{}, op)
-}
-
-func DagOpFromContext(ctx context.Context) *DagOp {
-	if val := ctx.Value(dagOpContextKey{}); val != nil {
-		return val.(*DagOp)
-	}
-	return nil
-}
-
-func (op DagOp) Exec(ctx context.Context, inputs []solver.Result, srv *dagql.Server) (outputs []solver.Result, err error) {
-	obj, err := srv.Load(withDagOpContext(ctx, &op), op.ID)
+func (op DirectoryDagOp) Exec(ctx context.Context, g bksession.Group, inputs []solver.Result, opt buildkit.OpOpts) (outputs []solver.Result, err error) {
+	obj, err := opt.Server.Load(withDagOpContext(ctx, op), op.ID)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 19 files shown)


core/schema/container.go
```diff
@@ -1807,6 +1807,44 @@ func (s *containerSchema) asTarball(
...
+				{
+					Field: "withFile",
+					Args: []dagql.NamedInput{
+						{
+							Name:  "path",
+							Value: dagql.String(filename),
+						},
+						{
+							Name:  "source",
+							Value: dagql.NewID[*core.File](dagql.CurrentID(ctx).WithMetadata("", true)),
```
jedevc, line 1831: Ahaha, this is really neat :tada: I can do a self call here, forcing it to be tainted - the context it's re-evaluated in has a `DagOp`, so we then execute the other path.

jedevc:
Oooookay, have done a lot of refactoring, and now - these DagOps are implemented as middlewares :tada: And all tests are passing :smile: There's also a few related dagql fixes+improvements that I found while in the area.

This is *the first step* - it's very easy to mark the `asTarball` method to be cached using buildkit now, with very very few changes to the actual content of the method - we'll be able to reuse this approach later.

However, the end goal is to completely own the cache refs - this is completely possible, it's just not *immediately* apparent how this should look in dagql. One option is something like https://github.com/dagger/dagger/commit/7a36d890cf96d1a3a21abc85fe5579bb121a7557. This is another custom op, and requires a lot more manual intervention, but you can write directly to a cache ref and skip the nasty `llb.Local` hack. Eventually we can switch to this approach, but didn't want to keep blocking this PR from getting in.

core/schema/middlewares.go
```diff
@@ -0,0 +1,137 @@
...
+			dagql.Selector{
+				Field: "withFile",
+				Args: []dagql.NamedInput{
+					{
+						Name:  "path",
+						Value: dagql.String(filename),
+					},
+					{
+						Name:  "source",
+						Value: dagql.NewID[*core.File](dagql.CurrentID(ctx).WithMetadata("", true)),
```
sipsma, line 54: Why do we set the digest to empty string for these? Not sure if that confuses the dagql cache and/or telemetry.

core/schema/middlewares.go
```diff
@@ -0,0 +1,137 @@
...
+	hasPBs, ok := a.(core.HasPBDefinitions)
+	if !ok {
+		return nil, nil
+	}
+
+	depsDefs, err := hasPBs.PBDefinitions(ctx)
+	if err != nil {
+		return nil, err
+	}
+	deps := make([]llb.State, 0, len(depsDefs))
```
sipsma, line 128: nit: it looks like some of our impls of `PBDefinitions` use a map so the order is random, which I *think might* change the buildkit cache key and result in extra invalidation?

I'm sort of doubtful this is anything remotely near a big deal atm, so not a blocker. If there's a really easy fix then maybe worth it, otherwise just worth keeping in mind

sipsma (commented):
As a general comment, probably worth a double check we aren't accidentally leaking dagql/client/session memory via all this plumbing through buildkit. I couldn't see anything obvious but there's too many layers of abstraction to really know just by looking

core/schema/middlewares.go
```diff
@@ -0,0 +1,137 @@
...
+			return inst, err
+		}
+		return core.NewRawDagOp[R](ctx, srv, dagql.CurrentID(ctx).WithMetadata("", true), deps)
+	}
+}
+
+// DagOpFileMiddleware caches a file field as a buildkit operation - this is
+// more specialized than DagOpMiddleware, since that serializes the value to
+// JSON, so we'd just end up with a cached ID instead of the actual content.
+func DagOpFileMiddleware[T dagql.Typed, A any](srv *dagql.Server, fn func(ctx context.Context, self dagql.Instance[T], args A) (dagql.Instance[*core.File], error)) func(ctx context.Context, self dagql.Instance[T], args A) (dagql.Instance[*core.File], error) {
```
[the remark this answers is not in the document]
jedevc: Yeah, okay :tada: we need go 1.24 though so that we can do generic type aliases: https://go.dev/doc/go1.24#language

core/dagop.go
```diff
@@ -0,0 +1,221 @@
...
+// allows for efficiently caching the result.
+func NewDirectoryDagOp(ctx context.Context, srv *dagql.Server, id *call.ID, inputs []llb.State) (*Directory, error) {
+	requiredType := (&Directory{}).Type().NamedType
+	if id.Type().NamedType() != requiredType {
+		return nil, fmt.Errorf("expected %s to be selected, instead got %s", requiredType, id.Type().NamedType())
+	}
+	dagOp := DirectoryDagOp{ID: id}
+
+	st, err := buildkit.NewCustomLLB(ctx, dagOp, inputs,
+		llb.WithCustomNamef("%s %s", dagOp.Name(), id.Display()),
```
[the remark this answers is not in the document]
jedevc: Ah, yes. Because this is really just used internally for naming the op (it really only appears in the engine logs), we can just use the receiver + field, so I've made a new `Name` method for ID (so in our case, this will just be `Container.asTarball`).

core/schema/middlewares.go
jedevc: Hah, the description of `WithMetadata` actually does cover this:

> `customDigest, if not empty string, will become the ID's digest.`

That said, yes, confusing. I've added a clearer shorthand `ID.WithTaint` to clarify the intention in these cases.

engine/server/session.go
```diff
@@ -918,6 +918,18 @@ func (srv *Server) ServeHTTPToNestedClient(w http.ResponseWriter, r *http.Reques
 
 const InstrumentationLibrary = "dagger.io/engine.server"
 
+func (srv *Server) DagqlServer(ctx context.Context) (*dagql.Server, error) {
```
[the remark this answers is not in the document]
jedevc: I'm probably missing something, but on creation, the `CustomOp` only has access to the worker. We don't actually have access to the engine server, or the list of sessions, or even a specific client - so I don't *think* there's a way of getting it out?

We just need some *interface* to pass this through via, we could use `core.Server` honestly, it's just that that's in `core`, and we need this in `buildkit`. So I just selected the one function we need here.

core/schema/middlewares.go
sipsma: Lol, forgetting things I wrote, 100% agree the more clear name is needed here 😄

engine/server/session.go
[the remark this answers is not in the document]
sipsma: I misread this originally, I thought it was a method on session.... This all lgtm

sipsma:
> Anything specific to do here? I've added the benchmark label, so we can get some data from that hopefully? Not really sure if there's anything specific to do for measuring "leakage".

For now, just start the dev engine locally with debug endpoints exposed, run some reasonable set of tests for a while (I often just run `TestModule` for 5 or so minutes), ctrl-c and then see what the RSS looks like. Hit the `/debug/gc` and pprofs endpoints of the dev engine to double check things.

sipsma (approved):
Double checking memory here is worth a check before merge, but overall LGTM!

sipsma:
> Still seeing a lot of log.newRing (we know about) and FileActionMkFile.Unmarshal (we know about), but at least nothing new seems to be leaking.

@jedevc those shouldn't be leaking anymore after the other fixes, if they are then something might be going wrong. I'll try locally too.

PR #9396, opened 2025-01-16
feat: add float support for golang

### LINKED ISSUE #9280: 🐞 Dagger does not support floating point numbers
### What is the issue?

I was chatting with @jedevc and he said that our code implies that you should be able to use floating point numbers as a standard type. 

https://github.com/dagger/dagger/blob/main/dagql/builtins.go

However, when you try to return a float, Dagger complains that float32 is an unsupported basic type. 

### Dagger version

dagger v0.15.1 (registry.dagger.io/engine:v0.15.1) darwin/arm64

### Steps to reproduce

```
func (m *Go) FloatingNumber() float32 {
    return 1.23
}  
```

### Log output

```
Error: generate code: template: module.go.tmpl:85:3: executing "_dagg            
dule.go.tmpl" at <ModuleMainSrc>: error calling ModuleMainSrc: failed            
 type def code for Go: failed to convert method FloatingNumber to fun            
ailed to generate return type code: unsupported basic type: float32  
```

TomChv:
### Summary

Add float support and fixes #9280.

Doc page available on https://github.com/dagger/dagger/pull/9408

### Examples

**Go**
```go
package main

import "context"

type Test struct{}

func (m *Test) Test(n float64) float64 {
	return n
}

func (m *Test) TestFloat32(n float32) float32 {
	return n
}
```

**Typescript**

⚠️ Since there's no `float` in Typescript, I created a custom `float` type in the client gen so user can declare a type as `FLOAT_KIND`.

```ts
import { float, object, func } from "@dagger.io/dagger"

@object()
export class Test {
  @func()
  test(n: float): float {
    return n
  }
}
```

**Python**

```python
import dagger
from dagger import dag


@dagger.object_type
class Test:
    @dagger.function
    def test(self, n: float) -> float:
        return n
```

### TODO

- [x] Add `FLOAT_KIND` to client binding `TypeDef`
- [x] Add float support for marshal/unmarshal
- [x] Add float support on CLI
- [x] Add integrations tests
- [x] Add Go support
- [x] Add Typescript support
- [x] Add Python support 
- [x] Update doc pages https://github.com/dagger/dagger/pull/9408

### DIFF
--- a/cmd/codegen/generator/typescript/templates/src/header.ts.gtpl
+++ b/cmd/codegen/generator/typescript/templates/src/header.ts.gtpl
@@ -9,6 +9,11 @@ inherited by futures objects and common types.
  */
 import { Context } from "../common/context.js"
 
+/**
+ * Declare a number as float in the Dagger API.
+ */
+export type float = number
+
 class BaseClient {
   /**
    * @hidden
--- a/.changes/unreleased/Added-20250117-154447.yaml
+++ b/.changes/unreleased/Added-20250117-154447.yaml
@@ -1,0 +1,10 @@
+kind: Added
+body: |
+  Add float support in the engine.
+
+  Note: the precision of float is limited to float64 inside the engine.
+  And any call will serialize/deserialize the value to float64.
+time: 2025-01-17T15:44:47.089106+01:00
+custom:
+  Author: TomChv
+  PR: "9396"
--- a/cmd/codegen/generator/go/templates/format.go
+++ b/cmd/codegen/generator/go/templates/format.go
@@ -37,7 +37,7 @@ func (f *FormatTypeFunc) FormatKindScalarInt(representation string) string {
 }
 
 func (f *FormatTypeFunc) FormatKindScalarFloat(representation string) string {
-	representation += "float"
+	representation += "float64"
 	return representation
 }
 
--- a/cmd/codegen/generator/go/templates/module_types.go
+++ b/cmd/codegen/generator/go/templates/module_types.go
@@ -182,6 +182,8 @@ func (spec *parsedPrimitiveType) TypeDefCode() (*Statement, error) {
 			kind = Id("dagger").Dot("TypeDefKindIntegerKind")
 		case types.IsBoolean:
 			kind = Id("dagger").Dot("TypeDefKindBooleanKind")
+		case types.IsFloat:
+			kind = Id("dagger").Dot("TypeDefKindFloatKind")
 		default:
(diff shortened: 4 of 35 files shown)


jedevc:
fyi @dagger/sdk-elixir @dagger/sdk-php in case you're interested in adding this as well (likely as follow-ups to this).

wingyplus:
> fyi @dagger/sdk-elixir @dagger/sdk-php in case you're interested in adding this as well (likely as follow-ups to this).

Added to my backlog. 🫡

wingyplus:
@TomChv @jedevc here is the follow up PRs:

- elixir - https://github.com/dagger/dagger/pull/9410
- php - https://github.com/dagger/dagger/pull/9411

jedevc:
Just pushed support for python.

cmd/codegen/generator/typescript/templates/src/header.ts.gtpl
```diff
@@ -9,6 +9,11 @@ inherited by futures objects and common types.
  */
 import { Context } from "../common/context.js"
 
+/**
+ * Declare a number as float in the Dagger API.
+ */
+export type float = number
```
jedevc, line 15: Sigh. This is a real pain that we need this :cry: 

Sadly, I have no better ideas - I think it's the right call to create this distinction in typescript, instead of forcing everything else to deal with a type that is both int+float.

That said, kinda curious what happens if a typescript function returns a `number` and you return a floating point from it - does it just get truncated? Curious what the engine does with it, ideally we should get a nice error message from typescript telling the user to use the custom `float` type instead.

jedevc (approved):
LGTM, one nit about checking that we get a good typescript error if you return a floating point number from a `number` field (which will type-check correctly, but we shouldn't allow).

cmd/codegen/generator/typescript/templates/src/header.ts.gtpl
TomChv: Already tried, the engine throws an error because it expect a integer and the `.` is considered as an invalid syntax :)

jedevc: Yeah, having some sort of custom error handling in the typescript SDK would be great :tada:

TomChv: Added: [ceb238b](https://github.com/dagger/dagger/pull/9396/commits/ceb238b167a636244ab48dd738bfc0f32a8255ca)

jedevc (approved):

PR #9400, opened 2025-01-16
fix: ensure the go sdk uses namespaced volumes

jedevc:
Without this, we can end up accidentally using the wrong module by default! This means we end up with distinct cache volumes for these, when they should be shared *everywhere*.

(split out of #9389)

### DIFF
--- a/core/schema/cache.go
+++ b/core/schema/cache.go
@@ -62,12 +62,12 @@ func (s *cacheSchema) cacheVolume(ctx context.Context, parent dagql.Instance[*co
 			return inst, err
 		}
 
-		namespaceKey = name + symbolic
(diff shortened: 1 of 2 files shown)


core/schema/cache.go
```diff
@@ -62,12 +62,12 @@ func (s *cacheSchema) cacheVolume(ctx context.Context, parent dagql.Instance[*co
 			return inst, err
 		}
 
-		namespaceKey = name + symbolic
+		namespaceKey = "mod(" + name + symbolic + ")"
```
jedevc, line 65: Note: this is changed to include `mod` so that no module can be cleverly named `internal` to collide with the internal cache.

sipsma (approved):

PR #9404, opened 2025-01-17
fix: allow injecting system ca-certs for wolfi

jedevc:
And while in the area, updated the logic for reading the `os-release` file (so as to allow a fallback according to https://www.linux.org/docs/man5/os-release.html).

### DIFF
--- a/core/integration/images.go
+++ b/core/integration/images.go
@@ -4,6 +4,7 @@ import "github.com/dagger/dagger/engine/distconsts"
 
 const (
 	alpineImage  = distconsts.AlpineImage
+	wolfiImage   = "cgr.dev/chainguard/wolfi-base"
 	busyboxImage = distconsts.BusyboxImage
 	golangImage  = distconsts.GolangImage
 	debianImage  = "debian:bookworm"
--- a/engine/buildkit/containerfs/fs.go
+++ b/engine/buildkit/containerfs/fs.go
@@ -282,11 +282,17 @@ func (ctrFS *ContainerFS) PathExists(path string) (bool, error) {
 
 func (ctrFS *ContainerFS) OSReleaseFileContains(ids [][]byte, idLikes [][]byte) (bool, error) {
 	f, err := ctrFS.Open("/etc/os-release")
-	if err != nil {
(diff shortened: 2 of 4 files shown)


engine/buildkit/containerfs/fs.go
```diff
@@ -282,12 +282,18 @@ func (ctrFS *ContainerFS) PathExists(path string) (bool, error) {
 
 func (ctrFS *ContainerFS) OSReleaseFileContains(ids [][]byte, idLikes [][]byte) (bool, error) {
 	f, err := ctrFS.Open("/etc/os-release")
-	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			return false, nil
-		}
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
```
marcosnils, line 285: nit the err != nil can be omitted here and below since errors.Is already checks for that

marcosnils, line 285: nit the err != nil can be omitted here and below since errors.Is already checks for that

marcosnils (approved):
LGTM! left minor nits which are not blockers to merge.

core/integration/images.go
```diff
@@ -4,6 +4,7 @@ import "github.com/dagger/dagger/engine/distconsts"
 
 const (
 	alpineImage  = distconsts.AlpineImage
+	wolfiImage   = "cgr.dev/chainguard/wolfi-base"
```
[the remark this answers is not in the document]
jedevc: `core/integration` is the tests package, I avoided adding it to `distconsts` since it's not used anywhere else. Is there somewhere else it should go?

engine/buildkit/containerfs/fs.go
jedevc: I don't think it can? I'm checking for if it's any error that's *not* `os.ErrNotExist`. Without that, `err = nil` would follow down this path, and return immediately.

marcosnils: right..having a brainfart morning. :heavy_check_mark:

PR #9408, opened 2025-01-17
docs: add float & integer documentation on parameter & return value 

TomChv:
Follow up to https://github.com/dagger/dagger/pull/9396 to add doc content on float.
I also added some doc for integer since it was missing.


docs/current_docs/api/arguments.mdx
```diff
@@ -133,6 +133,84 @@ The result will look like this:
 HELLO, WORLD
 ```
 
+## Integer arguments
+
+To pass an integer argument to a Dagger function, add the corresponding flag to the `dagger call` command, followed by the integer value.
+
+Here is an example of a Dagger function that accepts a integer argument:
```
vikram-dagger, line 140:
```suggestion
Here is an example of a Dagger function that accepts an integer argument:
```

docs/current_docs/api/arguments.mdx
```diff
@@ -133,6 +133,84 @@ The result will look like this:
...
+The result will look like this:
+
+```shell
+3
+```
+
+## Float arguments
+
+:::warning
+Floats are currently only supported in Go and TypeScript SDKs for now.
```
vikram-dagger, line 176: The information in this section is only applicable to the Go and TypeScript SDKs. Floating-point number arguments are not currently supported in the Go SDKs.

docs/current_docs/api/arguments.mdx
```diff
@@ -133,6 +133,84 @@ The result will look like this:
...
+dagger call add-integer --a 1 --b 2
+```
+
+The result will look like this:
+
+```shell
+3
+```
+
+## Float arguments
```
vikram-dagger, line 173:
```suggestion
## Floating-point number arguments
```

docs/current_docs/api/arguments.mdx
```diff
@@ -133,6 +133,84 @@ The result will look like this:
...
+
+:::warning
+Floats are currently only supported in Go and TypeScript SDKs for now.
+:::
+
+To pass a float argument to a Dagger function, add the corresponding flag to the `dagger call` command, followed by the float value.
```
vikram-dagger, line 179:
```suggestion
To pass a floating-point number as argument to a Dagger function, add the corresponding flag to the `dagger call` command, followed by the value.
```

docs/current_docs/api/arguments.mdx
```diff
@@ -133,6 +133,84 @@ The result will look like this:
...
+
+Here is an example of a Dagger function that accepts a float argument:
```
vikram-dagger, line 181:
```suggestion
Here is an example of a Dagger function that accepts a floating-point number as argument:
```

docs/current_docs/api/arguments.mdx
```diff
@@ -133,6 +133,84 @@ The result will look like this:
...
+<TabItem value="Go">
+```go file=./snippets/functions/arguments-float/go/main.go
+```
+</TabItem>
+<TabItem value="TypeScript">
+
+:::note
+There's no `float` type keyword in TypeScript because the type keyword `number` already supports floating point numbers.
+
+To declare a `float` argument on the function signature, you need to import `float` from `@dagger.io/dagger` and use it as argument's type.
```
vikram-dagger, line 193:
```suggestion
To declare a `float` argument on the function signature, import `float` from `@dagger.io/dagger` and use it as an argument's type.
```

docs/current_docs/api/return-values.mdx
```diff
@@ -47,6 +47,79 @@ The result will look like this:
...
+The result will look like this:
+
+```shell
+3
+```
+
+## Float return values
+
+:::warning
+Floats are currently only supported in Go and TypeScript SDKs for now.
```
vikram-dagger, line 87:
```suggestion
The information in this section is only applicable to the Go and TypeScript SDKs. Floating-point number arguments are not currently supported in the Go SDKs.
```

docs/current_docs/api/return-values.mdx
```diff
@@ -47,6 +47,79 @@ The result will look like this:
...
+dagger call add-integer --a 1 --b 2
+```
+
+The result will look like this:
+
+```shell
+3
+```
+
+## Float return values
```
vikram-dagger, line 84:
```suggestion
## Floating-point number return values
```

docs/current_docs/api/return-values.mdx
```diff
@@ -47,6 +47,79 @@ The result will look like this:
...
+
+:::warning
+Floats are currently only supported in Go and TypeScript SDKs for now.
+:::
+
+Here is an example of a Dagger Function that returns the sum of two floats:
```
vikram-dagger, line 90:
```suggestion
Here is an example of a Dagger Function that returns the sum of two floating-point numbers:
```

docs/current_docs/api/return-values.mdx
```diff
@@ -47,6 +47,79 @@ The result will look like this:
...
+<TabItem value="Go">
+```go file=./snippets/functions/return-values-float/go/main.go
+```
+</TabItem>
+<TabItem value="TypeScript">
+
+:::note
+There's no `float` type keyword in TypeScript because the type keyword `number` already supports floating point numbers.
+
+To declare a `float` return type on the function signature, you need to import `float` from `@dagger.io/dagger` and use it as return type.
```
vikram-dagger, line 102:
```suggestion
To declare a `float` return type on the function signature, import `float` from `@dagger.io/dagger` and use it as return type.
```

docs/current_docs/api/snippets/functions/arguments-float/go/main.go
```diff
@@ -0,0 +1,7 @@
+package main
```
vikram-dagger, line 1: It looks like the code snippets are duplicated between the args and retvals examples, for each language. Instead of this, I suggest changing the snippet directory names to reflect dual use e.g. `arguments-return-values-float` and then reusing/importing the same snippet for both the args and retvals examples.

vikram-dagger (commented):
Minor copy changes and one change to example snippets structure needed, all good otherwise.

core/integration/module_test.go
```diff
@@ -5183,6 +5183,103 @@ func (t *Test) IgnoreDirButKeepFileInSubdir(
 	})
 }
 
+func (ModuleSuite) TestFloat(ctx context.Context, t *testctx.T) {
```
jedevc, line 5186: Hm, it feels weird to have a dedicated test for this - I was honestly expecting that we would have a more generic test where we pass a lot of different types in (e.g. int, string, etc) - but it doesn't seem like this exists?

Maybe I've missed it, but if it does, that would be a better place to add it (otherwise, up to you if you want to add it).

docs/current_docs/api/arguments.mdx
TomChv: Actually it's now supported in Python so I'll remove this msg :D

core/integration/module_test.go
TomChv: I think I could add that in a follow up PR but not in that one

docs/current_docs/api/snippets/functions/arguments-float/go/main.go
TomChv: Done in: [f538507](https://github.com/dagger/dagger/pull/9408/commits/f538507862a018c8a3ae541e03ed41030d235ab2)

docs/current_docs/api/return-values.mdx
TomChv: Removed because now we have a Python snippet

vikram-dagger (approved):

PR #9409, opened 2025-01-17
[PHP] Update README.md in template

arendjantetteroo:
Fix the first example to use the Doc Attribute.

The $description in the #[DaggerFunction] attribute is deprecated.


charjr (approved):
Thank you for spotting this.

That is completely correct and Doc is now the preferred way.

PR #9413, opened 2025-01-18
ci: add ca-certificates package to the Go module

marcosnils:
This allows the engine to correctly setup certificates in case there's
any custom certs configured

### DIFF
--- a/modules/go/main.go
+++ b/modules/go/main.go
@@ -81,6 +81,7 @@ func New(
 				// The specific version is dictated by Dagger's own requirement
 				// FIXME: make this optional with overlay support
 				"protoc~3.21.12",
+				"ca-certificates",
 			}}).
 			WithEnvVariable("GOLANG_VERSION", version).
 			WithEnvVariable("GOPATH", "/go").


modules/go/main.go
```diff
@@ -81,6 +81,7 @@ func New(
 				// The specific version is dictated by Dagger's own requirement
 				// FIXME: make this optional with overlay support
 				"protoc~3.21.12",
+				"ca-certificates",
```
jedevc, line 84: :thinking: why does this get used? A bit confused.

Also. are they essential to running any wolfi container? Then we could always install them right?

marcosnils: > 🤔 why does this get used? A bit confused.

that's one of the methods used by the certificate provisioning  flow here: https://github.com/marcosnils/dagger/blob/361050092c9d93d7c9be14e16f22021be49a61fa/engine/buildkit/cacerts/distros.go?plain=1#L287  where if the `update-ca-certificates` binary exists it'll just call that instead of manually handing the certificate bundling. 

I usually find this approach more consistent and easier to follow as it leverages the default tooling for installing certs in the system.

> Also. are they essential to running any wolfi container? Then we could always install them right?

Yeah.. I think we should. If you LGTM this comment, I can go ahead and add it to the base Alpine module.

jedevc: Yup, SGTM :tada:

sipsma (approved):

jedevc:
Merging now, the outstanding comment can be resolved as a follow-up :tada:

Issue #9068, opened 2024-11-27
Local workdir interferes with `-m PATH`


shykes:
When loading a module by local path, the load may succeed or fail depending on the current workdir - specifically, dependending on whether the current workdir is inside our outside the target module's context.

Example:

```console
#!/usr/bin/env dagger shell

EXAMPLE_REPO=https://github.com/dagger/dagger
EXAMPLE_COMMIT=c58e58e70af3b3cab15154878253b707283b2740
EXAMPLE_MODULE=cmd/dagger
EXAMPLE_FUNCTION=binary

.load github.com/dagger/dagger |
dev |
with-directory /tmp/src $(
  .git $EXAMPLE_REPO | commit $EXAMPLE_COMMIT | tree
) |
# First call from inside the context: will succeed
with-workdir /tmp/src |
with-exec dagger,call,-m,./$EXAMPLE_MODULE,$EXAMPLE_FUNCTION |
# First call from outside the context: will fail
with-workdir /tmp |
with-exec dagger,call,-m,./src/$EXAMPLE_MODULE,$EXAMPLE_FUNCTION
```

Output:

```
Error: input: daggerCli lstat /src/.git: no such file or directory
```

jedevc:
Reproduced - you can do this pretty easily without shell.

```
$ git clone https://github.com/dagger/dagger.git
$ dagger call -m ./✔ connect 0.2s
✔ load module 6.0s
✔ parsing command line arguments 0.0s

✘ daggerDev: DaggerDev! 20.9s
! lstat /src/.git: no such file or directory
│ ✔ upload /home/jedevc/Documents/work/dagger/dagger from b32t5hc5ezsulbudioavkyjil (client id: 2yx0sqkc15xcoyg06v4tyr7ju, session id: c44uxv87qikusrfue7cpktc9m) (exclude: bin, .git, **/node_modules, **/.venv, **/__pycache__) 0.6s
│ 
│ ✘ version: Version! 19.7s
│ ! lstat /src/.git: no such file or directory
[...]

PR #9418, opened 2025-01-20
fix: relatively resolve dependency context directories

jedevc:
Fixes https://github.com/dagger/dagger/issues/9068.

We need to take into account that dependencies have a *different* relative context directory. The fact this was working before, was that generally, we were only invoking modules from the same location as the `dagger.json` - however, if invoked from above or below, then we'd fail to apply the correct transformation and we'd get the wrong context directory.

Thankfully, it's fairly simple - all we need to do is mirror the logic that modifies the `refString` and apply it to `relHostPath` as well.

### DIFF
--- a/core/schema/modulesource.go
+++ b/core/schema/modulesource.go
@@ -856,10 +856,15 @@ func (s *moduleSchema) moduleSourceResolveDependency(
 	// depSrc.RootSubpath is ../baz and relative to foo/bar.
 	// depSubpath is the resolved path, i.e. foo/baz.
 	depSubpath := filepath.Join(srcRootSubpath, depRootSubpath)
-
 	if !filepath.IsLocal(depSubpath) {
 		return inst, fmt.Errorf("module dep source root path %q escapes root", depRootSubpath)
 	}
+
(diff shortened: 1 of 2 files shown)


core/schema/modulesource.go
```diff
@@ -856,11 +856,16 @@ func (s *moduleSchema) moduleSourceResolveDependency(
...
-
 	if !filepath.IsLocal(depSubpath) {
 		return inst, fmt.Errorf("module dep source root path %q escapes root", depRootSubpath)
 	}
 
+	srcRelHostPath, err := src.SourceRootRelSubPath()
+	if err != nil {
+		return inst, err
+	}
+	depRelHostPath := filepath.Join(srcRelHostPath, depRootSubpath)
```
jedevc, line 867: We *don't* need to do an escape check here - because we actually do one later on, when we go to `LoadContext`.

TomChv (approved):
LGTM

sipsma (approved):

jedevc:
Hm, somehow python + rust jobs time out? Not quite sure why.

Issue #8424, opened 2024-09-12
Resolve terminology ambiguities


helderco:
> [!note]
> This is my own brainstorming epic, just putting it out there to allow me to reference it in discussions. It also provides an overview on the subject, allows for tracking and contributions to the conversation, but should create separate issues to facilitate discussion and tracking progress on specific items.
## Motivation

We have a problem with some terminology. When Modules were being developed, we used the same names for things that have a more expanded meaning in that context. We need to rename some of those things to avoid confusion when talking about them.

For example:
- `sdk`:
    - client library (not just the client bindings), exported to an `sdk` subdir of a module (in some SDKs)
    - _runtime_ Module, referenced in `dagger init --sdk` and the `"sdk"` field of `dagger.json`
    - directory that has both things (`/sdk/<name>`), but doesn’t actually include everything that pertains to an “SDK”
- `codegen`
    - name of the executable (e.g., `/cmd/codegen`) that generates the client bindings for the client library
    - name of the Function in an SDK's *runtime* Module for generating all the files needed for a Module written in that SDK
- Client
    - The Dagger code, in each SDK, to build and execute queries to the Dagger API
    - Code for running pipelines **without** using Modules (`dagger run <script>`)
        - Example: `dagger run <script>`
        - Could be a custom/external script, a compiled CLI, or even a web app
    - Code for running Module Functions
        - Example: `dagger call <function>`
        - Could be a custom/external script, compiled CLI, or even a web app
- Entrypoint
    - Any Container’s `ENTRYPOINT` metadata
    - A [Module’s “constructor”](https://docs.dagger.io/manuals/developer/entrypoint-function), i.e., the Function that instantiates the “main object” in a Module
    - The code in an SDK that’s used as the entrypoint for the “runtime container” (for registering and invoking Module Functions)
        - For example, TypeScript and PHP have an `entrypoint.ts` or `entrypoint.php` file in the library.

*TODO: Add more?*

## Suggestions

- **Client**
    - “client bindings”: the types that are generated from the API schema
    - “client library”: the code that includes the “client bindings”
    - Need terminology to distinguish between “module vs non-module” support
        - “custom/external” isn’t clear enough since it could apply to both
- **`codegen`:** reserve it for “generating the client bindings” for a client library
    - Rename *runtime* Function `Codegen` → `Generate` for preparing the `Directory` that is exported during `dagger init` and `dagger develop`
- **`sdk`**
    - Rename the generated `sdk` in Modules to the name of the published package:
[...]

Issue #7276, opened 2024-05-03
Customize CLI command names


sagikazarmark:
### What are you trying to do?

Allow customizing how function names are translated into commands.

For example:

```
// +command=e2e
func (m *Ci) E2e() {
    // ...
}
```

### Why is this important to you?

Function names are often translated to weird commands. For example `E2e` becomes `e-2-e`.

### How are you currently working around this?

I call my end-to-end test function `Etoe`. 🙈

helderco:
Rather than customizing the CLI name, we need to enable customizing the name in the API. Python and Typescript do that through the decorators:
- [Python: Name Overrides](https://docs.dagger.io/manuals/developer/python/558732/name-overrides)
- [TypeScript: Aliases](https://docs.dagger.io/manuals/developer/typescript/406314/aliases)

This would add paritity in Go, but probably with `+name` instead of `+command`.

However, if doing the same as Py or TS, does it lead to the same issue of `e2e` translating to `e-2-e`? I'll need to check where that's happening because we use two different Go utils in different places. One of those (I suspect _strcase_) is breaking numbers instead of just changes in casing.

The API uses the same naming convention as TypeScript (_camelCase_), which is similar to Go (_PascalCase_), in that both use casing to split words. Python's much simpler to break words into (_snake_case_), same as the CLI (_kebab-case_). So maybe using the CLI's convention is the more neutral and consistent way to override the names in all SDKs.

Think about acronyms. For example `SSH` translates to `s-s-h` unless you have a list of common acronyms (like we do). But overriding to `+name=ssh` would make it less ambiguous or easier to know how to convert consistently.

PR #9422, opened 2025-01-21
Create modules in Java

eunomie:
This PR allows to create Dagger modules using Java.

```console
$ dagger init --sdk=java my-java-module

$ tree my-java-module
my-java-module
├── dagger.json
├── pom.xml
└── src
    └── main
        └── java
            └── io
                └── dagger
                    └── sample
                        └── module
                            ├── MyJavaModule.java
                            └── package-info.java

8 directories, 4 files
```

This will create the module with a default template. This template module is based on the same than Go and others, with the `container-echo` and `grep-dir` functions.

A default `pom.xml` is also generated.

Once done, it's possible to list the functions or call them:

```console
$ dagger functions -m my-java-module

Name             Description
```

```console
$ dagger call -q -m my-java-module container-echo "hello dagger" stdout

hello dagger

```

---

The java module code is using annotations to know what to expose and call:
- `@Module`
- `@Object`
- `@Function`

The entrypoint code will be generated at build time. In that way it's close to the Go modules (not a generic endpoint at runtime only).

The default `pom.xml` set the main class to the generated entrypoint one automatically using the `maven-shade` plugin.

By default only the `src` directory and the `pom.xml` are necessary. But to keep a nice Developer Experience and help IDEs to complete code, all the Dagger SDK dependencies (the Dagger java sdk plus all the generated files for the different types and the entrypoint) can be generated and made available under the `target/generated-sources` directory:
- `dagger`: the generated files based on Dagger introspection JSON
- `entrypoint`: the generated entrypoint based on user's code
- `sdk`: the `sdk/java/dagger-java-sdk` source code (this is not really generated sources but a way to make it available easily)

<details><summary>generated files</summary>
<pre>
$ tree target
my-java-module/target
└── generated-sources
    ├── dagger
    │   └── io
    │       └── dagger
    │           └── client
    │               ├── BuildArg.java
    │               ├── CacheSharingMode.java
    │               ├── CacheVolume.java
    │               ├── CacheVolumeID.java
    │               ├── Client.java
    │               ├── Container.java
    │               ├── ContainerID.java
    │               ├── CurrentModule.java
    │               ├── CurrentModuleID.java
    │               ├── Directory.java
    │               ├── DirectoryID.java
    │               ├── EnumTypeDef.java
    │               ├── EnumTypeDefID.java
    │               ├── EnumValueTypeDef.java
    │               ├── EnumValueTypeDefID.java
    │               ├── EnvVariable.java
    │               ├── EnvVariableID.java
    │               ├── Error.java
    │               ├── ErrorID.java
    │               ├── FieldTypeDef.java
    │               ├── FieldTypeDefID.java
    │               ├── File.java
    │               ├── FileID.java
    │               ├── Function.java
    │               ├── FunctionArg.java
    │               ├── FunctionArgID.java
    │               ├── FunctionCall.java
    │               ├── FunctionCallArgValue.java
    │               ├── FunctionCallArgValueID.java
    │               ├── FunctionCallID.java
    │               ├── FunctionID.java
    │               ├── GeneratedCode.java
    │               ├── GeneratedCodeID.java
    │               ├── GitModuleSource.java
    │               ├── GitModuleSourceID.java
    │               ├── GitRef.java
    │               ├── GitRefID.java
    │               ├── GitRepository.java
    │               ├── GitRepositoryID.java
    │               ├── ImageLayerCompression.java
    │               ├── ImageMediaTypes.java
    │               ├── InputTypeDef.java
    │               ├── InputTypeDefID.java
    │               ├── InterfaceTypeDef.java
    │               ├── InterfaceTypeDefID.java
    │               ├── JSON.java
    │               ├── Label.java
    │               ├── LabelID.java
    │               ├── ListTypeDef.java
    │               ├── ListTypeDefID.java
    │               ├── LocalModuleSource.java
    │               ├── LocalModuleSourceID.java
    │               ├── Module.java
    │               ├── ModuleDependency.java
    │               ├── ModuleDependencyID.java
    │               ├── ModuleID.java
    │               ├── ModuleSource.java
    │               ├── ModuleSourceID.java
    │               ├── ModuleSourceKind.java
    │               ├── ModuleSourceView.java
    │               ├── ModuleSourceViewID.java
    │               ├── NetworkProtocol.java
    │               ├── ObjectTypeDef.java
    │               ├── ObjectTypeDefID.java
    │               ├── PipelineLabel.java
    │               ├── Platform.java
    │               ├── Port.java
    │               ├── PortForward.java
    │               ├── PortID.java
    │               ├── ReturnType.java
    │               ├── ScalarTypeDef.java
    │               ├── ScalarTypeDefID.java
    │               ├── Secret.java
    │               ├── SecretID.java
    │               ├── Service.java
    │               ├── ServiceID.java
    │               ├── Socket.java
    │               ├── SocketID.java
    │               ├── SourceMap.java
    │               ├── SourceMapID.java
    │               ├── Terminal.java
    │               ├── TerminalID.java
    │               ├── TypeDef.java
    │               ├── TypeDefID.java
    │               ├── TypeDefKind.java
    │               ├── Version.java
    │               └── Void.java
    ├── entrypoint
    │   └── io
    │       └── dagger
    │           └── gen
    │               └── entrypoint
    │                   └── Entrypoint.java
    └── sdk
        └── io
            └── dagger
                ├── client
                │   ├── Arguments.java
                │   ├── Convert.java
                │   ├── Dagger.java
                │   ├── DaggerQueryException.java
                │   ├── IDAble.java
                │   ├── InputValue.java
                │   ├── PrivateVisibilityStrategy.java
                │   ├── QueryBuilder.java
                │   ├── QueryPart.java
                │   ├── Scalar.java
                │   └── engineconn
                │       ├── CLIDownloader.java
                │       ├── CLIRunner.java
                │       ├── ConnectParams.java
                │       ├── Connection.java
                │       ├── FileFetcher.java
                │       └── Provisioning.java
                └── module
                    ├── Base.java
                    ├── annotation
                    │   ├── Function.java
                    │   ├── Module.java
                    │   └── Object.java
                    └── info
                        ├── FunctionInfo.java
                        ├── ModuleInfo.java
                        ├── ObjectInfo.java
                        └── ParameterInfo.java

19 directories, 112 files
</pre>
<details>

Those files are not intended to be added to git. To re-generate them: `dagger develop`

> [!WARNING]
> Due to some issues between Docker Desktop, Java and MacBook Pro m4 (it looks like other machines are not impacted) in case of `arm64` platform a specific flag is added to Java/Maven: `-XX:UseSVE=0`.
> See for instance https://github.com/corretto/corretto-21/issues/85


sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -0,0 +1,324 @@
...
+          for (var fnInfo : objectInfo.functions()) {
+            var fnReturnClass = Class.forName(fnInfo.returnType());
+            if (firstFn) {
+              firstFn = false;
+              im.beginControlFlow("if (fnName.equals($S))", fnInfo.name());
+            } else {
+              im.nextControlFlow("else if (fnName.equals($S))", fnInfo.name());
+            }
+            var fnBlock =
+                CodeBlock.builder().add("$T fn = clazz.getMethod($S", Method.class, fnInfo.name());
```
jcsirot, line 183: it should be `fnInfo.qName()` otherwise it should throw a `NoSuchMethodException` when an alias is defined

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -0,0 +1,324 @@
...
+          } else {
+            im.nextControlFlow("else if (parentName.equals($S))", objectInfo.name());
+          }
+          im.addStatement("$T clazz = Class.forName($S)", Class.class, objectInfo.qualifiedName())
+              .addStatement("var obj = $T.fromJSON(dag, parentJson, clazz)", Convert.class)
+              .addStatement(
+                  "clazz.getMethod(\"setClient\", $T.class).invoke(obj, dag)", Client.class);
+          var firstFn = true;
+          for (var fnInfo : objectInfo.functions()) {
+            var fnReturnClass = Class.forName(fnInfo.returnType());
```
jcsirot, line 175: `Class.forName` will fail if the type is List because `returnType` is of the form `java.util.List<foo.bar.xxx>`
Type parsing should be enhanced to handle List or array

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -0,0 +1,324 @@
...
+        f.writeTo(processingEnv.getFiler());
+      } catch (ClassNotFoundException | IOException e) {
+        throw new RuntimeException(e);
+      }
+    }
+
+    return true;
+  }
+
+  private CodeBlock typeDef(String name) {
```
jcsirot, line 300: the `typedef` method should be able to handle types like list and enums

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
eunomie: Yes, you're right, I forgot that case, I will add it now. Thanks

jcsirot: I think there is also the same issue for the parameter types

sdk/java/dagger-java-sdk/src/main/java/io/dagger/module/Base.java
```diff
@@ -0,0 +1,17 @@
+package io.dagger.module;
+
+import io.dagger.client.Client;
+
+public class Base {
```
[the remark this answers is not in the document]
eunomie: Done 👍

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -0,0 +1,376 @@
...
+          var invokeBlock =
+              CodeBlock.builder().add("$T res = ($T) fn.invoke(obj", fnReturnClass, fnReturnClass);
+          for (var parameterInfo : fnInfo.parameters()) {
+            Class<?> paramClazz = classForName(parameterInfo.type());
+            fnBlock.add(", $T.class", paramClazz);
+
+            invokeBlock.add(", $L", parameterInfo.name());
+
+            im.addStatement(
+                "$T $L = ($T) $T.fromJSON(dag, inputArgs.get($S), Class.forName($S))",
```
jcsirot, line 200: `Class.forName(type)` does not work for primitive types. use `classForName` instead

sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/Convert.java
```diff
@@ -0,0 +1,57 @@
+package io.dagger.client;
+
+import com.google.gson.Gson;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.concurrent.ExecutionException;
+
+public interface Convert {
```
jcsirot, line 8: Maybe replace `interface` by a `final class`. It is a good practice for utility classes and it prevents Convert to be implemented

sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/Convert.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  static <T> T fromJSON(Client dag, String json, Class<T> clazz)
+      throws NoSuchMethodException,
+          InvocationTargetException,
+          InstantiationException,
+          IllegalAccessException,
+          ClassNotFoundException {
+    Gson gson = new Gson();
+    var o = clazz.getDeclaredConstructor().newInstance();
```
jcsirot, line 41: If the type does not have a default constructor this line will throw an exception. it is better to use `isAssignableFrom()` : `Scalar.class.isAssignableFrom(clazz)` and `IDAble.class.isAssignableFrom(clazz)`

eunomie:
I improved the primitive types support, they are both working for returning types and arguments.
I also included your (@jcsirot ) improvements regarding javadoc parsing.

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -0,0 +1,433 @@
...
+  }
+
+  private static Boolean isNotBlank(String str) {
+    return str != null && !str.isBlank();
+  }
+
+  private String parseTypeDescription(Element element) {
+    String javadocString = elementUtils.getDocComment(element);
+    if (javadocString == null) {
+      return element.getAnnotation(Module.class).description();
```
jcsirot, line 406: shouldn't be `element.getAnnotation(Object.class).description();` here? Otherwise it may throw a NPE

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -0,0 +1,444 @@
...
+              CodeBlock.builder().add("$T fn = clazz.getMethod($S", Method.class, fnInfo.qName());
+          var invokeBlock =
+              CodeBlock.builder().add("$T res = ($T) fn.invoke(obj", fnReturnClass, fnReturnClass);
+          for (var parameterInfo : fnInfo.parameters()) {
+            Class<?> paramClazz = classForName(parameterInfo.type());
+            fnBlock.add(", $T.class", paramClazz);
+
+            invokeBlock.add(", $L", parameterInfo.name());
+
+            im.addStatement("$T $L = null", paramClazz, parameterInfo.name());
```
jcsirot, line 238: This statement does not work with primitive types (boolean, int, etc). You cannot set a primitive typed var to null

jcsirot: I suggest
```
            if (paramClazz.isPrimitive()) {
              if (paramClazz == int.class) {
                im.addStatement("$T $L = 0", paramClazz, parameterInfo.name());
              } else if (paramClazz == boolean.class) {
                im.addStatement("$T $L = false", paramClazz, parameterInfo.name());
              }
            } else {
              im.addStatement("$T $L = null", paramClazz, parameterInfo.name());
            }
```

jcsirot: we will probably have to handle other primitive types like byte, long, char (and maybe float and double) but we can do it a future PR

eunomie: Fixed, thanks :-)

jedevc (commented):
I think this looks like a good start to me, just left some general integration points as comments.

*Really* not qualified to comment on any of the Java code - will trust you and @jcsirot on this :tada:

Would also like @helderco to take a look if possible, since he's more familiar with the process of how we've integrated with the other SDKs :tada:

.dagger/build/builder.go
```diff
@@ -131,7 +132,7 @@ func (build *Builder) WithGPUSupport() *Builder {
 func (build *Builder) Engine(ctx context.Context) (*dagger.Container, error) {
 	eg, ctx := errgroup.WithContext(ctx)
 
-	sdks := []sdkContentF{build.goSDKContent, build.pythonSDKContent, build.typescriptSDKContent}
+	sdks := []sdkContentF{build.goSDKContent, build.pythonSDKContent, build.typescriptSDKContent, build.javaSDKContent}
```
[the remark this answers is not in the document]
jedevc: Yeah, it's mostly about the size - we don't want to have everyone needing to pull in large external SDKs when actually they're not going to be using them.

...that said, I wonder if maybe we might want to include these in some dev builds. Pulling them from git doesn't work very nicely if they're not actually *in* git, because it's a local working copy. (just thinking out loud here, not really relevant for this discussion).

sdk/java/runtime/main.go
```diff
@@ -0,0 +1,312 @@
...
+					"pom.xml",
+					"**/*.java",
+				}}).
+			// And rename everything so that they match the dagger module name
+			WithExec([]string{"sed", "-i", "-e", fmt.Sprintf("s/dagger-module/%s/g", snakeName), absPath("pom.xml")}).
+			WithExec([]string{"sed", "-i", "-e", fmt.Sprintf("s/DaggerModule/%s/g", camelName), absPath("src", "main", "java", "io", "dagger", "sample", "module", "DaggerModule.java")}).
+			WithExec([]string{"sed", "-i", "-e", fmt.Sprintf("s/dagger-module-name/%s/g", name), absPath("src", "main", "java", "io", "dagger", "sample", "module", "DaggerModule.java")}).
+			WithExec([]string{"sed", "-i", "-e", fmt.Sprintf("s/DaggerModule/%s/g", camelName), absPath("src", "main", "java", "io", "dagger", "sample", "module", "package-info.java")}).
+			WithExec([]string{"sed", "-i", "-e", fmt.Sprintf("s/dagger-module-name/%s/g", name), absPath("src", "main", "java", "io", "dagger", "sample", "module", "package-info.java")}).
+			WithExec([]string{"mv", absPath("src", "main", "java", "io", "dagger", "sample", "module", "DaggerModule.java"), absPath("src", "main", "java", "io", "dagger", "sample", "module", fmt.Sprintf("%s.java", camelName))})
```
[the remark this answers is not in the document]
eunomie: OK, I'll change that. I guess in that case it might be best to read/write them outside

sdk/java/dagger-java-annotation-processor/src/test/resources/io/dagger/gen/entrypoint/entrypoint.java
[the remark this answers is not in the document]
eunomie: It should :-) This is a test file. Basically it covers the full user module parsing - entrypoint generation. There's an example user module and this is the associated entrypoint.
It might not be the most fine grain test, but it's a good help while working on it and covers a lot of the code.

sdk/java/README.md
```diff
@@ -17,6 +18,60 @@
 
 A [Dagger.io](https://dagger.io) SDK written in Java.
 
+## Modules
```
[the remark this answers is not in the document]
eunomie: Yes, I haven't invested much in the readme _for now_. I was waiting a bit to have something good enough. But I'll have a second pass on it.

sdk/java/dagger-codegen-maven-plugin/src/main/resources/schemas/schema-0.15.1.json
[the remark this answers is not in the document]
eunomie: That's a good question. This comes from the way the SDK (not modules) was generated. Basically this is an independent process: from maven, then calling the dagger binary to get the equivalent of the introspection json file.
The module part only relies on the introspection.json sent by dagger. And this part could follow the same pattern.

It might sounds closer to the java world, to only rely on things like maven, but once the sdk library is published nobody (except dagger devs) should have the need to do that.

I'll see if I can change that. This will also remove the duplication of the introspection query. That's an issue as if the query change on dagger engine side, it will not be reflected here.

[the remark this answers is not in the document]
jcsirot: These files were used to generate the Java API for a given version of Dagger before the introduction of modules without having to install the engine of the targeted version. Frankly I think they are no longer necessary today

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -0,0 +1,454 @@
...
+            boolean hasDefaultValue =
+                parameterInfo.optional() && isNotBlank(parameterInfo.defaultValue());
+            if (hasDescription || parameterInfo.optional()) {
+              rm.addCode(", new $T.WithArgArguments()", io.dagger.client.Function.class);
+              if (hasDescription) {
+                rm.addCode(".withDescription($S)", parameterInfo.description());
+              }
+              if (hasDefaultValue) {
+                rm.addCode(
+                    ".withDefaultValue($T.from($S))", JSON.class, parameterInfo.defaultValue());
```
jcsirot, line 184: when the parameter type is String maybe we could handle a special case in order to accept non-escaped default values in Optional annotation.

having to write `@Optional(defaultValue="\"some string\"")` rather `@Optional(defaultValue="some string")` is cumbersome

WDYT?

sdk/java/dagger-codegen-maven-plugin/src/main/resources/schemas/schema-0.15.1.json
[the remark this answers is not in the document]
eunomie: I'm going to remove them for now, and then we can think about changing the way we build the sdk so that we don't rely on a specific query stored in the java sdk. But that's the part outside of the modules.

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -0,0 +1,454 @@
...
+          "dag.typeDef().withKind($T.$L)", TypeDefKind.class, TypeDefKind.BOOLEAN_KIND.name());
+    } else if (name.startsWith("java.util.List<")) {
+      name = name.substring("java.util.List<".length(), name.length() - 1);
+      return CodeBlock.of("dag.typeDef().withListOf($L)", typeDef(name).toString());
+    } else if (name.endsWith("[]")) {
+      name = name.substring(0, name.length() - 2);
+      return CodeBlock.of("dag.typeDef().withListOf($L)", typeDef(name).toString());
+    }
+
+    var clazz = Class.forName(name);
```
jcsirot, line 381: This statement may throw a `ClassNotFoundException` when having a function returning itself or another module object since these objects have not yet been compiled, and then the classes cannot be loaded into the JVM.

For instance:
```
@Object
class Foo() extends Base {
  private Container ctr;
 
  public Foo() {
    this.ctr = dag.container();
  }
 
  public Foo WithAlpine() {
    this.ctr = this.ctr.from("alpine:latest")
    return this;
}
```

Additional type metadata (for instance if the type is an enum or not) should be parsed from the AST provided by the annotation parser rather than introspection

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
eunomie: Yeah, I don't know how much we should escape from the json string.
It's sad we can't have an `Object` for the value.
But yes, maybe we can check if it's a string and not starting by `"` or `'` and in that case add them to the string automatically.

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
eunomie: Hum, true.
I'll improve that.

Regarding enums, for now I haven't found a way to check if it's an enum from the annotation parser 🤔 Maybe I'm missing something.

But at least I can handle the class not found correctly, especially as this code is only used for enums.

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
eunomie: I've done something on that, that should cover the main cases with strings.

sdk/java/runtime/template/src/main/java/io/dagger/sample/module/DaggerModule.java
```diff
@@ -0,0 +1,38 @@
...
+@Object(value = "dagger-module-name")
+public class DaggerModule extends Base {
+  public DaggerModule() {
+    super();
+  }
+
+  /** Returns a container that echoes whatever string argument is provided */
+  @Function
+  public Container containerEcho(String stringArg) {
+    return dag.container()
```
helderco, line 23: Curious, where does `dag` come from?

sdk/java/runtime/main.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+	mvnCtr := m.codegenBase(ctx, modSource, introspectionJSON)
+
+	return dag.
+		GeneratedCode(dag.Directory().WithDirectory("/", m.generateCode(ctx, mvnCtr, introspectionJSON))).
+		WithVCSGeneratedPaths([]string{
+			"target/generated-sources/dagger/**",
+			"target/generated-sources/entrypoint/**",
+			"target/generated-sources/sdk/**",
+		}).
```
helderco, line 83: Maybe this?
```suggestion
		WithVCSGeneratedPaths([]string{
			"target/generated-sources/**",
		}).
```

sdk/java/runtime/main.go
```diff
@@ -0,0 +1,312 @@
...
+func (m *JavaSdk) codegenBase(
+	ctx context.Context,
+	modSource *dagger.ModuleSource,
+	introspectionJSON *dagger.File,
+) *dagger.Container {
+	return m.
+		// We need maven to build the user module
+		mvnContainer(ctx).
+		// Mount the maven folder where we installed all the generated dependencies
+		WithMountedDirectory("/root/.m2", m.buildJavaDependencies(ctx, introspectionJSON).Directory("/root/.m2")).
```
helderco, line 101: Curious, is `$HOME/.m2` a maven convention?

sdk/java/runtime/template/pom.xml
```diff
@@ -0,0 +1,117 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>io.dagger.sample</groupId>
```
helderco, line 7: Maybe
```suggestion
    <groupId>io.dagger.module</groupId>
```

For example, for `dagger init --sdk=java --name=my-module`:
```xml
<groupId>io.dagger.module</groupId>
   <artifactId>my-module</artifactId>
    <version>1.0-SNAPSHOT</version>
    <name>my-module</name>
```
If that conflicts with SDK plugins or something, then maybe something else. I don't see users coming and editing out the "sample" here. Most people just change the generated functions, but mostly leave the rest from the template.

sdk/java/runtime/main.go
```diff
@@ -0,0 +1,312 @@
...
+		// copy the generated entrypoint under target/generated-sources/annotations
+		WithDirectory(
+			filepath.Join(m.moduleConfig.modulePath(), "target", "generated-sources", "entrypoint"),
+			entrypoint.Directory(filepath.Join(m.moduleConfig.modulePath(), "target", "generated-sources", "annotations"))).
+		// copy the sdk source code under target/generated-sources/sdk
+		// this is not really generated-sources, this is the sdk. But we don't want it as the user source code
+		// and we don't want to install it on the user machine. That way the java classes are made available
+		// to a build system or an IDE without to interfere with the user source code
+		WithDirectory(
+			filepath.Join(m.moduleConfig.modulePath(), "target", "generated-sources", "sdk"),
```
helderco, line 219: This makes sense but can you please find another name instead of "sdk"? The reason is that term is too much overloaded. See:
- https://github.com/dagger/dagger/issues/8424

It's also proven to be harder when filtering out files because it conflicts with other `sdk` dirs, including https://github.com/dagger/dagger/tree/main/sdk. Ideally this would be the name of the package that is published to a public registry. Like `src/dagger` or `vendor/dagger` (generally speaking).

sdk/java/runtime/main.go
jcsirot: Yes it is `$HOME/.m2` contains the maven configuration files for the local user and maven stores all downloaded dependencies in the `$HOME/.m2/repository` directory

sdk/java/runtime/template/src/main/java/io/dagger/sample/module/DaggerModule.java
jcsirot: `dag` is defined in the superclass `Base`. Maybe we could find a better name like `AbstractModule`

helderco: Just got surprised about the implicit reference. 🙂 I expected a `this.dag` if it's an object instance parameter.

core/schema/sdk.go
```diff
@@ -125,7 +127,7 @@ func parseSDKName(sdkName string) (SDK, string, error) {
 	}
 
 	// inbuilt sdk go/python/typescript currently does not support selecting a specific version
-	if slices.Contains([]SDK{SDKGo, SDKPython, SDKTypescript}, SDK(sdkNameParsed)) && hasVersion {
+	if slices.Contains([]SDK{SDKGo, SDKPython, SDKTypescript, SDKJava}, SDK(sdkNameParsed)) && hasVersion {
```
helderco, line 130: Needs to be moved to the condition below, next to php and elixir.

sdk/java/runtime/main.go
```diff
@@ -0,0 +1,312 @@
...
+
+func (c *moduleConfig) sdkPath() string {
+	return filepath.Join(c.modulePath(), GenPath)
+}
+
+func New(
+	// Directory with the Java SDK source code.
+	// dagger-java-samples is not necessary to build, but as it's referenced in the root pom.xml maven
+	// will check if it's there. So we keep it, until a better solution (like to extract the samples somewhere else)
+	// +optional
```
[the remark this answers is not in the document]
eunomie: Is that also true when we have `+default` instead of `+defaultPath`?

sdk/java/runtime/main.go
```diff
@@ -0,0 +1,312 @@
...
+
+const (
+	MavenImage  = "maven:3.9.9-eclipse-temurin-23-alpine"
+	MavenDigest = "sha256:77fe6f79f868484d85679bfaa121b34177e3fedf570086ea6babbd6db2223e89"
+	JavaImage   = "eclipse-temurin:23-jre-alpine"
+	JavaDigest  = "sha256:bd8e2c8c19bcadbaa8c6a128051a22384c6f7cfe5fa520cb663fe21fff96f084"
+
+	ModSourceDirPath = "/src"
+	ModDirPath       = "/opt/module"
+	GenPath          = "sdk"
```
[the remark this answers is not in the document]
eunomie: Answering both this comment and https://github.com/dagger/dagger/pull/9422#discussion_r1934323266

`sdk` by itself is only used temporarily to build the libraries as dependencies before to build the user module. from the runtime module or from the user perspective there's no `sdk` folder. Except under `target/generated-sources` and it's not really "visible" from the user's perspective.
So it's not really the same as in php/ts for instance where there's the `src/` and the `sdk/` directories.

That said, I understand the point, I'll change for something better 👍

sdk/java/runtime/main.go
```diff
@@ -0,0 +1,312 @@
...
+	return filepath.Join(c.modulePath(), GenPath)
+}
+
+func New(
+	// Directory with the Java SDK source code.
+	// dagger-java-samples is not necessary to build, but as it's referenced in the root pom.xml maven
+	// will check if it's there. So we keep it, until a better solution (like to extract the samples somewhere else)
+	// +optional
+	// +defaultPath="/sdk/java"
+	// +ignore=["**", "!dagger-codegen-maven-plugin/", "!dagger-java-annotation-processor/", "!dagger-java-sdk/", "!dagger-java-samples", "!LICENSE", "!README.md", "!pom.xml"]
```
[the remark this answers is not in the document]
eunomie: I'll see to remove some of them 👍

sdk/java/runtime/template/src/main/java/io/dagger/sample/module/DaggerModule.java
eunomie: In java `this` is not mandatory :-)
I can add it in the samples so that's more explicit if we think that's better.

sdk/java/runtime/main.go
eunomie: I think I like the explicit way of describing it, but both are working.

sdk/java/runtime/template/pom.xml
eunomie: I _think_ that's quite common on java world to have people using their own namespace. Like `com.company.project`. And I expected people to change it. But I can find something better than `sample`. Maybe I'll avoid completely `io.dagger` to not have conflicts at all 🤔

core/schema/sdk.go
eunomie: 👍 I missed that when I extracted it from the built in list

sdk/java/runtime/template/src/main/java/io/dagger/sample/module/DaggerModule.java
```diff
@@ -0,0 +1,38 @@
...
+import io.dagger.client.Directory;
+import io.dagger.module.Base;
+import io.dagger.module.annotation.Function;
+import io.dagger.module.annotation.Object;
+
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+/** DaggerModule main object */
+@Object(value = "dagger-module-name")
```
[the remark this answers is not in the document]
jcsirot: the annotation parameter is used when you want to alias your exposed Module class. When I started thinking about how to expose Module objects I wasn't sure if creating aliases for the names would be useful or necessary.
Note also that `@Object(value = "dagger-module-name")` and `@Object("dagger-module-name")` are semantically identical

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
jcsirot: Unfortunately the same issue exists also in the `classForName(TypeInfo ti)` method

sdk/java/runtime/main.go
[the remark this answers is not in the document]
helderco: Yes!

Technically, the "optional" in the TypeDef maps to "nullable". So you can have both _optional_ and a _default value_. If the default value is not null, that means it allows a null, even if not the default.

Currently this isn't supported by the engine due to the way the server interprets null values as omissions (i.e., the client can't send a `null` value through the API). But at some point we may fix that so it's better to future proof.

The wording "optional" here is a problem, because it doesn't entirely reflect what it is, and is confusing when used at the same time as a default value.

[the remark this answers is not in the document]
helderco: In Java this could be an issue. The mere existence of `@Optional` makes it nullable, so you can't currently represent the following GraphQL schema:
```graphql
name: String! = "Foo"
```
As I said, it doesn't really matter today, but in the future it's a breaking change to fix. Any ideas?

sdk/java/runtime/template/src/main/java/io/dagger/sample/module/DaggerModule.java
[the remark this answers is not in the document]
helderco: I suggest removing the alias, or at least hiding it as much as possible so people don't learn to depend on it (i.e., consider the need for deprecating it in the future).

Python has it since the beginning in some places (not object names). TypeScript too. The initial intent was to allow circumventing reserved keywords for function names or arguments, for example. It was documented in the docs but later removed for simplifcation sake, and consistency (Go doesn't have it).

However, it's a problem if people use it with names that actually diverge because it can be confusing to users. So to solve the reserved keywords problem, I added support for the SDK to automatically remove the last `_` of a name (which is a Python convention for resolving naming conflicts), when reporting the name to the API. The custom alias support is also still there though, just not documented to discourage abuse.

No one has actually asked for it AFAIK, except in Go (which doesn't have it), because of that language's convention of capitalizing acronyms, making some type names being converted in an unexpected way in the CLI. That's not a naming conflict issue, it's a casing convention conversion issue. See:
- https://github.com/dagger/dagger/issues/7276

Can you evaluate how the Java SDK would currently report something like an `E2e` class? Is it `e2e` or `e-2-e`? What about a `SSH` class? Does it report `ssh` or `s-s-h` (due to camel casing not recognizing the acronym)? Maybe Java users wouldn't think to name a class `SSH` instead of `Ssh`.

sdk/java/runtime/template/src/main/java/io/dagger/sample/module/DaggerModule.java
helderco: I like without it because it's more consistent with the other SDKs. 🙂

sdk/java/runtime/main.go
helderco: It's up to you, but I don't see when you'd want to have something in `generated-sources` that's not marked as "generated" 🙂 So I don't see the value in the explicitness here, and it makes it extra work to change or rename those dirs in the future, for older modules.

To be fair, even if at one point we had to add code manually to clean up these files (`.gitattributes` and `.gitignore`), the point of this API was to centralize that logic so it's probably a non issue today.

sdk/java/runtime/template/pom.xml
helderco: In other SDKs, when there wasn't support for custom names, we conventionalized on "main". So there's lots of modules out there where the language's package name is "main", even when the dagger module's name is something else.

In this case, whatever is more familiar to Java users takes priority. But I suspect lots of people won't be changing that value. Or does it become super visible in the IDE?

eunomie: This is now `io.dagger.modules.usermodule` with `usermodule` based on the user input (the name of the folder, lowercase, no `-`, no `_`. That way it fits the java convention about package naming and doesn't conflict with `io.dagger.module` package.

sdk/java/runtime/main.go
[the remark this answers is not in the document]
eunomie: Just to be sure to understand:
- right now, to have a `default` (or `defaultPath`) will imply `optional`
- but we want to decouple them so that we can express a default value but avoid any null value

Basically what we are supporting is:

```graphql
name: String!        # the default
name: String         # +optional
name: String = "Foo" # +optional +default="Foo"
                     # but +default="Foo" produces the same behavior
```

And there's the 4th case you mentioned

```GraphQL
name: String! = "Foo" # +default="Foo"
```

If that's right, then I "just" have to decouple them, so one annotation `@Optional` and one annotation `@Default`. _For now_ `@Default` will imply `@Optional` behavior but that can then be changed as soon as the engine supports it.

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
eunomie: I just improved that part too.
And I expanded the test case to cover to reference the module object itself, and more regarding primitive types and lists.

sdk/java/runtime/main.go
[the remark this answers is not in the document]
eunomie: Test files and samples are now removed

```
LICENSE
README.md
dagger-codegen-maven-plugin
dagger-codegen-maven-plugin/pom.xml
dagger-codegen-maven-plugin/src
dagger-codegen-maven-plugin/src/main
dagger-codegen-maven-plugin/src/main/java
dagger-codegen-maven-plugin/src/main/java/io
dagger-codegen-maven-plugin/src/main/java/io/dagger
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/DaggerCLIUtils.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/DaggerCodegenMojo.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/DaggerSchemaGeneratorMojo.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/AbstractMultiTypesVisitor.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/AbstractVisitor.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/CodeWriter.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/CodegenVisitor.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/EnumValue.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/EnumVisitor.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/Field.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/Helpers.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/IDAbleVisitor.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/InputObject.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/InputVisitor.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/ObjectVisitor.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/QueryType.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/Scalar.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/ScalarVisitor.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/Schema.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/SchemaVisitor.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/Type.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/TypeKind.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/TypeRef.java
dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/VersionVisitor.java
dagger-codegen-maven-plugin/src/main/resources
dagger-codegen-maven-plugin/src/main/resources/introspection
dagger-codegen-maven-plugin/src/main/resources/introspection/introspection.graphql
dagger-java-annotation-processor
dagger-java-annotation-processor/pom.xml
dagger-java-annotation-processor/src
dagger-java-annotation-processor/src/main
dagger-java-annotation-processor/src/main/java
dagger-java-annotation-processor/src/main/java/io
dagger-java-annotation-processor/src/main/java/io/dagger
dagger-java-annotation-processor/src/main/java/io/dagger/annotation
dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor
dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
dagger-java-samples
dagger-java-samples/pom.xml
dagger-java-sdk
dagger-java-sdk/pom.xml
dagger-java-sdk/src
dagger-java-sdk/src/main
dagger-java-sdk/src/main/java
dagger-java-sdk/src/main/java/io
dagger-java-sdk/src/main/java/io/dagger
dagger-java-sdk/src/main/java/io/dagger/client
dagger-java-sdk/src/main/java/io/dagger/client/Arguments.java
dagger-java-sdk/src/main/java/io/dagger/client/Dagger.java
dagger-java-sdk/src/main/java/io/dagger/client/DaggerQueryException.java
dagger-java-sdk/src/main/java/io/dagger/client/IDAble.java
dagger-java-sdk/src/main/java/io/dagger/client/IDAbleSerializer.java
dagger-java-sdk/src/main/java/io/dagger/client/InputValue.java
dagger-java-sdk/src/main/java/io/dagger/client/PrivateVisibilityStrategy.java
dagger-java-sdk/src/main/java/io/dagger/client/QueryBuilder.java
dagger-java-sdk/src/main/java/io/dagger/client/QueryPart.java
dagger-java-sdk/src/main/java/io/dagger/client/Scalar.java
dagger-java-sdk/src/main/java/io/dagger/client/ScalarSerializer.java
dagger-java-sdk/src/main/java/io/dagger/client/ScalarStringDeserializer.java
dagger-java-sdk/src/main/java/io/dagger/client/engineconn
dagger-java-sdk/src/main/java/io/dagger/client/engineconn/CLIDownloader.java
dagger-java-sdk/src/main/java/io/dagger/client/engineconn/CLIRunner.java
dagger-java-sdk/src/main/java/io/dagger/client/engineconn/ConnectParams.java
dagger-java-sdk/src/main/java/io/dagger/client/engineconn/Connection.java
dagger-java-sdk/src/main/java/io/dagger/client/engineconn/FileFetcher.java
dagger-java-sdk/src/main/java/io/dagger/client/engineconn/Provisioning.java
dagger-java-sdk/src/main/java/io/dagger/module
dagger-java-sdk/src/main/java/io/dagger/module/Base.java
dagger-java-sdk/src/main/java/io/dagger/module/annotation
dagger-java-sdk/src/main/java/io/dagger/module/annotation/Function.java
dagger-java-sdk/src/main/java/io/dagger/module/annotation/Module.java
dagger-java-sdk/src/main/java/io/dagger/module/annotation/Object.java
dagger-java-sdk/src/main/java/io/dagger/module/annotation/Optional.java
dagger-java-sdk/src/main/java/io/dagger/module/info
dagger-java-sdk/src/main/java/io/dagger/module/info/FunctionInfo.java
dagger-java-sdk/src/main/java/io/dagger/module/info/ModuleInfo.java
dagger-java-sdk/src/main/java/io/dagger/module/info/ObjectInfo.java
dagger-java-sdk/src/main/java/io/dagger/module/info/ParameterInfo.java
dagger-java-sdk/src/main/java/io/dagger/module/info/TypeInfo.java
pom.xml
```

sdk/java/runtime/main.go
eunomie: I think I covered all places where there was `sdk`, should be better now

sdk/java/runtime/template/src/main/java/io/dagger/sample/module/DaggerModule.java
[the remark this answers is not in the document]
eunomie: If we create a java module `e2e`, the registered object name will be `E2E`.
If we want, by setting the `value` we can force it to `e2e`. But I'm not seeing any other consequence.

Calling a `e2e` module from java might sounds a bit weird, as it will be a `dag.e2E()` (where might want a `dag.e2e()`) but that's minor.
From Go it will be a `dag.E2E()`.

If it's `ssh` or `SSH` it will be registered as `Ssh`. Calling them from java or go works fine.
From java it will be a `dag.ssh()` and from Go it will be a `dag.SSH()`.

All that to say, I haven't experienced issues with this kind of object names, especially between Go and Java in all possible combinations.

sdk/java/runtime/main.go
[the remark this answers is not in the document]
helderco: > `# but +default="Foo" produces the same behavior`

Today it does, but it produces a different schema:
```graphql
name: String! = "Foo"  # +default="Foo"
name: String = "Foo"   # +default="Foo" +optional
```
The first only allows a string, while the second one allows a null too. That's why I'd rather have "optional" renamed to "nullable" in this case, for clarity.

Ideally I'd like to avoid having both `@Optional` and `@Default` because of the confusion it brings. Even though it would be consistent with Go (and the API the SDK uses), that implementation still needs "fixing".

Is it possible to add another parameter to `@Optional`? For example:
```graphql
name: String! = "Foo"  # @Optional(default="Foo")
name: String = "Foo"   # @Optional(default="Foo", nullable=true)
name: String           # @Optional()
```
Considering:
- In GraphQL, `name: String` is (implicitly) the same as `name: String = null`
- An empty `@Optional()` forces `nullable=true` or `default=null`
- If `default=null` then `nullable=true` is implied
  - _although the engine may enforce this already, because `name: String! = null` should be invalid_

This way the `nullable=true` can just be there for future proofing, but otherwise you wouldn't need to care about it in most cases.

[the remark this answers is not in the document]
helderco: If it's more natural in Java to assume that `@Optional(default="Foo")` also allows _null_, then you just flip that default. For comparison, this is how you can express this in Python:
```graphql
name: String! = "Foo"  # name: str = "Foo"
name: String = "Foo"   # name: str | None = "Foo"
name: String           # name: str | None = None
```
If you can express _nullable_ or not via Java's type system, then there's no need for the extra parameter.

Go doesn't use pointer types as _nullable_ because structs like `dagger.Directory` are always pointers, even when required.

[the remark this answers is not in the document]
eunomie: Well, the typing system will not really help on that. For instance a `String` parameter of a function is always nullable.
But it's possible to add extra instructions to the entrypoint so it ensures (if not declared as nullable) the value sent to the function is not null. It will be done at runtime.

A quite common pattern I guess is to have a `@NonNull` annotation. The problem with that is, even if we add it to the templates we create during a `dagger init`, it means by default the parameter is nullable/optional.

```java
containerEcho(String stringArg)                          // stringArg: String
containerEcho(@NotNull String stringArg)                 // stringArg: String!
containerEcho(@Default("Foo") String stringArg)          // stringArg: String = "Foo"
containerEcho(@NotNull @Default("Foo") String stringArg) // stringArg: String! = "Foo"
```

But we can also do the other way around with a `@Nullable`:

```java
containerEcho(@Nullable String stringArg)                 // stringArg: String
containerEcho(String stringArg)                           // stringArg: String!
containerEcho(@Nullable @Default("Foo") String stringArg) // stringArg: String = "Foo"
containerEcho(@Default("Foo") String stringArg)           // stringArg: String! = "Foo"
```

I need to check to be sure, but it should be possible to have an annotation processor that will inject the right code.
So if `@Nullable` is not there, it will adds a `Objects.requireNonNull(stringArg, "stringArg must not be null");` at the beginning of the method.

I'm not entirely sure which option is best. I guess we don't want `+optional` by default, so the second might be better.

Happy to read your feedback on that @jcsirot

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -0,0 +1,519 @@
...
+                              .addStatement("$T result", JSON.class)
+                              .beginControlFlow("if (parentName.isEmpty())")
+                              .addStatement("$T modID = register()", ModuleID.class)
+                              .addStatement("result = $T.toJSON(modID)", JsonConverter.class)
+                              .nextControlFlow("else")
+                              .addStatement(
+                                  "result = invoke(parentJson, parentName, fnName, inputArgs)")
+                              .endControlFlow()
+                              .addStatement("fnCall.returnValue(result)")
+                              .nextControlFlow("catch ($T e)", Exception.class)
```
jcsirot, line 355: This catch bloc does not correctly handle exceptions thrown by modules and may lead to an error in the engine when the exception message is null. `InvocationTargetException` should also be handled to return the actual exception thrown by the module function

```
  .nextControlFlow("catch ($T e)", InvocationTargetException.class)
  .addStatement("fnCall.returnError(dag.error(e.getTargetException().getMessage()))")
  .addStatement("throw e")
  .endControlFlow()
```

sdk/java/runtime/main.go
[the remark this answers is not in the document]
eunomie: I've pushed a new version that should be able to support all cases.
I've replaced the `@Optional` annotation by two:
- `@Nullable`
- `@Default`

With that, we can express the different cases:

```java
  @Function
  public String nonNullableNoDefault(String stringArg) {
    if (stringArg == null) {
      throw new RuntimeException("can not be null");
    }
    return stringArg;
  }

  @Function
  public String nullableNoDefault(@Nullable String stringArg) {
    if (stringArg == null) {
      stringArg = "was a null value";
    }
    return stringArg;
  }

  @Function
  public String nonNullableDefault(@Default("default value") String stringArg) {
    if (stringArg == null) {
      throw new RuntimeException("can not be null");
    }
    return stringArg;
  }

  @Function
  public String nullableDefault(@Nullable @Default("null") String stringArg) {
    if (stringArg == null) {
      stringArg = "was a null value by default";
    }
    return stringArg;
  }
```

The current version implies `@Nullable` if `@Default` is present.

If it's a string and there's an unquoted default value it will be quoted (just to make it nice) _except_ if the default value is `null` as we can't set a null value by itself. In that case the `null` json is used. It means if we want the string `"null"` it has to be manually quoted `"\"null\""`. But I think that's fine.

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
eunomie: Added, thanks

eunomie:
I think I covered the different feedbacks.
I also included the latest dependabot update from #9499 (to avoid conflicts).

How can we move forward with this PR that become very big and hard to maintain? Is there anything else to do before to merge it?
As said, this PR is just an intermediate step, that will allow to have a nice state to build on top of it, and will help for contributions (that are kind of blocked by this PR).

cc @jedevc @jcsirot @helderco 

---

CI is failing, but I'm not sure to understand why, that looks outside of the PR but I might have missed something.

jcsirot:
Even if we haven't covered 100% of the use cases for the modules, I think this PR provides a sufficiently stable starting point to iterate from. I agree with Yves; it will be much simpler and faster to contribute to Java support once this PR is merged.

sdk/java/runtime/main.go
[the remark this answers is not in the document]
helderco: So you don't really need `@Nullable` if `@Default("null")` right? If the default is `null` it should be set to nullable automatically.

helderco:
> I think I covered the different feedbacks. I also included the latest dependabot update from #9499 (to avoid conflicts).
> 
> How can we move forward with this PR that become very big and hard to maintain? Is there anything else to do before to merge it? 

If you're done addressing the feedback I'll take another look later today. I don't foresee any blockers for merge.

eunomie:
> If you're done addressing the feedback I'll take another look later today. I don't foresee any blockers for merge.

Thanks.
I think I'm done with the latest changes regarding the feedbacks 👍 

@jcsirot I just added one last change, renaming `io.dagger.module.Base` to `io.dagger.module.AbstractModule`.

sdk/java/dagger-java-sdk/src/main/java/io/dagger/module/annotation/Default.java
```diff
@@ -0,0 +1,33 @@
...
+ * and the default value is not between quotes, quotes will be added and all double quotes inside
+ * the value will be escaped
+ *
+ * <p>For instance <code>value="this \" is a quote"</code> will be automatically converted to <code>
+ * value="\"this is \\\" a quote \""</code>
+ *
+ * <p>One exception exists: if <code>value="null"</code> no quotes will be added to allow to set the
+ * json <code>null</code> value.
+ *
+ * <p>To set a default value will set the parameter as optional (nullable).
```
helderco, line 27: This isn't correct as I explained in the previous comment. Having a default value does make the parameter optional, but not necessarily "nullable".

To further explain, the confusion comes from a particularity of GraphQL, which doesn't follow what most languages do. Then our own API is named incorrectly which further confuses SDK authors. We need to insulate our API from GraphQL particularities like this.

That particularity is that in GraphQL, a nullable field without a default value is **implicitly attributed the default value of `null`**:
```graphql
"you set this"
foo: String

"graphql turns into this"
foo: String = null
```

So, for example, when writing docs for users, we need to be careful not to equate "nullable" with "optional", which won't make sense for people that aren't familiar with GraphQL. It's clearer to refer to "parameter with a default value of null" instead of just "nullable parameter" when it's a nullable without a default.

Having a default value is what makes it optional (because it allows it to be omitted) and most people should be able to understand that easily.

helderco (commented):
Almost there, it's an easy fix. 🙂

sdk/java/dagger-java-sdk/src/main/java/io/dagger/module/annotation/Default.java
helderco: For example, in Python, the following parameter is nullable:
```python
name: str | None
```
But it's **not optional**, unless it has a default:
```python
name: str | None = None
```
In the beginning of Python's typing system, when unions were't represented by a pipe char, `name: str | None` was represented as:
```python
# not actually optional!
name: Optional[str]
```
But note that `name` isn't actually optional there, unless you have a default value:
```python
name: Optional[str] = None
# or
name: Optional[str] = "foo"

# this is also optional!
name: str = "foo"
```
You can imagine that this originated a lot of confusion in developers. That's why a lot of people recommended using `Union[str, None]` instead of `Optional`, but didn't gain much traction because it's more verbose. The annotation ended up being deprecated and discouraged, although mostly because a simpler way to represent unions were added.

This has nothing to do with GraphQL, but it's the same kind of story 😄

wingyplus: > For example, in Python, the following parameter is nullable:
> 
> ```python
> name: str | None
> ```
> 
> But it's **not optional**, unless it has a default:
> 
> ```python
> name: str | None = None
> ```

This makes me rethink on Elixir. In Elixir when I declare a type like `name: String.t() | nil`, I convert this type system into optional without declaring a default value. 😂

eunomie:
@helderco thanks for the comments.

I'm trying to summarize here one more time and see if I'm right before to fix it in the code.

So, in term of Graphql we want to represent 4 cases:

```graphql
  arg: String!         // non nullable field, no specific default value, value has to be provided
  arg: String! = "Foo" // non nullable field, with a default value.
  arg: String          // nullable field, no specific default value, but GraphQL automatically sets
                       // a default value to null
                       // this is eq to arg: String = null
  arg: String = "Foo"  // nullable field, but a non null value set. Arg can be omitted but there's
                       // no current way the API can send a null value here. So it's theory possible
                       // but in practical not yet.
```

1. `arg: String!`

    In Java it should be `arg` (no annotation) and on the registering API `.withOptional(false)` (or omitted)

2. `arg: String! = "Foo"

    In Java it should be `@Default("Foo") arg` and on the registering API `.withDefaultValue("\"Foo\"")`.
    It's an optional parameter, the user doesn't have to provide one, but we don't set `.withOptional(true)` because that would  means to accept `null` and we don't want here

3. `arg: String` / `arg: String = null`

    In Java it could be a `@Nullable arg` but it's best to use `@Default(null)` to make it visible `null` will be the default value.
    On the registering API side it will be `.withOptional(true)` (meaning nullable) and `.withDefaultValue("null")` (that can be omitted as it's the default behavior on GraphQL side)

4. `arg: String = "Foo"`

    In Java that should be `@Nullable @Default("Foo") arg` and in the API `.withOptional(true)` and `.withDefaultValue("\"Foo\"")`.
    While it's true there's no way currently to send a null here.

Am I right on the different use cases?
So one way is to only present in the docs the first 3 cases and not the last one. Especially in Java where we can express with types the nullable/non nullable part.

If we want we can completely remove the `@Nullable` annotation so that the 4th case is impossible but we know we will be able to add id when/if doable on the API side. That way from a user's perspective there's only the cases:
- no default value, user must provide the argument
- a default value, user can omit the argument and the API will send the default value
- a default null value, user can omit the argument and the function will receive a null as the value

Does that sounds good?

---

One last thing I'm not sure to understand then:
https://github.com/dagger/dagger/blob/a99139ba894f240808fa6c1c9fd995ffba58cf41/modules/go/main.go#L24-L27
So that means `version` is nullable and has a default value `version: String = "1.23.2"`. But we could have avoid `+optional` the effective result would be the same as we can't send `null`, right?

helderco:
@eunomie, that's all correct 🙂 

> 3. `arg: String` / `arg: String = null`
>    In Java it could be a `@Nullable arg` but it's best to use `@Default(null)` to make it visible `null` will be the default value.
>    On the registering API side it will be `.withOptional(true)` (meaning nullable) and `.withDefaultValue("null")` (that can be omitted as it's the default behavior on GraphQL side)

You could equate `@Nullable` to `withOptional(true)` and otherwise not use that call (i.e., no need for `withOptional(false)`) because I think that if you send `withDefaultValue("null")`, the engine should set `optional = true` automatically. It should be an error to be able to define:
```graphql
arg: String! = null
```

Best confirm. Similar to this, but maybe it's being done elsewhere, like lower level:
- https://github.com/dagger/dagger/pull/8572

https://github.com/dagger/dagger/blob/a99139ba894f240808fa6c1c9fd995ffba58cf41/core/schema/module.go#L591-L596


> Does that sounds good?

👍

> One last thing I'm not sure to understand then:
> 
> https://github.com/dagger/dagger/blob/a99139ba894f240808fa6c1c9fd995ffba58cf41/modules/go/main.go#L24-L27
> 
> 
> So that means `version` is nullable and has a default value `version: String = "1.23.2"`. But we could have avoid `+optional` the effective result would be the same as we can't send `null`, right?

Yes, that's correct. I think that's a product of the confusion because of the word "optional". As in, if you have a default but don't mark it "optional" it's going to be a required argument. But there's also an historical need for this because until https://github.com/dagger/dagger/pull/8572, if this module were used as a dependency of another Go module, the arg would be generated as required. The Go codegen only considered optional values to be nullable, not ones that had a default too. Now that's not an issue.

In fact, in this case it's actually wrong to have `// +optional` because that arg is a `version string`, not `version *string`. As you know that doesn't make a difference today though, but it does incorrrectly create a nullable argument in the schema.

At lot of the core API have fields like this:
```go
Description string `default:""`
```
Producing:
```graphql
description: String = ""
```
When that field doesn't accept `nil`.

See "Core schema generation" in:

jcsirot:
Another point to consider, even though I'm not yet sure if it has any consequences, is the handling of primitive types.

**[Reminder for those not very familiar with Java]** In Java, you have primitive types (boolean, int, double, char, etc.) whose names start with a lowercase letter, and object types (String, Boolean, Integer, List, etc) whose names start with an uppercase letter. Unlike object types, primitive types are not nullable; however, they do have a default value (false for boolean and 0 for int).

This means (given the way the runtime is currently implemented), for example, that if we call the following function:
```java
@Function
public String testDefault(@Nullable boolean bool, @Nullable int integer) {
    return "bool=%s, integer=%s".formatted(bool, integer);
}
```
while passing null for the parameters bool and integer, then at execution, bool will be false and integer will be 0.

helderco:
> while passing null for the parameters bool and integer, then at execution, bool will be false and integer will be 0.

As long as passing that null doesn't thrown an exception I think it's ok, but doesn't make sense to set a type as nullable if it doesn't actually accept null values. It's better in this case to define as:
```java
@Function
public String testDefault(@Default("false") boolean bool, @Default("0") int integer) {
    return "bool=%s, integer=%s".formatted(bool, integer);
}
```

helderco (approved):
LGTM! 🎉

Nice work!

helderco:
Thanks, @eunomie and @jcsirot! Good base to iterate on improvements 🚀

PR #9424, opened 2025-01-21
chore(sdk/php): use DefaultPath to SDK source directory in dev module

wingyplus:
Instead of passing PHP SDK source directory to every functions in `dev` module, set it at constructor and use `DefaultPath` to load parent directory by default.

In the Dagger module, modify the `SDKPHP` to set PHP SDK source directory to module initialization instead of function call.

### DIFF
--- a/.dagger/sdk_php.go
+++ b/.dagger/sdk_php.go
@@ -39,7 +39,7 @@ func (t PHPSDK) Lint(ctx context.Context) error {
 			span.End()
 		}()
 		src := t.Dagger.Source().Directory(phpSDKPath)
-		_, err := dag.PhpSDKDev().Lint(src).Sync(ctx)
+		_, err := dag.PhpSDKDev(dagger.PhpSDKDevOpts{Source: src}).Lint().Sync(ctx)
 		return err
 	})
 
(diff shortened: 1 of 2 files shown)


charjr (commented):
This has one weird edge case right now:

https://github.com/dagger/dagger/blob/f186b26e37f0da492767a0756ef65eaedab2191e/sdk/php/dev/src/PhpSdkDev.php#L28-L29

If we pass in `$source` and `$container`, what is the expected outcome?

https://github.com/dagger/dagger/blob/f186b26e37f0da492767a0756ef65eaedab2191e/sdk/php/dev/src/PhpSdkDev.php#L100-L117

Right now, with both arguments like this. `$source` will be completely ignored if `$container` has been specified.

This isn't a new issue, so it doesn't block this PR as it could still happen otherwise. But it's a weird issue that I'd like to address.

jedevc:
LGTM once rebased @wingyplus :tada:

jedevc:
@charjr good point - @wingyplus, while you're here, could you update to *always* mount the source into the container (even if it's user specified)? Similar to the python dev module here:

https://github.com/dagger/dagger/blob/0d4d90cf0b16fd17f6a5616c1149ffcfd7d57099/sdk/python/dev/src/python_sdk_dev/main.py#L70-L76

wingyplus:
@charjr @jedevc Sorry for leaving it for so long. I refactoring it like Python does. Please kindly review it again. 🙏

jedevc (approved):

PR #9434, opened 2025-01-23
chore: hide typedef registration in progress output (and in CLI)

jedevc:
Hide these from the supported functions available on the command line (since they're for SDKs only).

Additionally, move the helper for this into the `dagui` package, so that we can use it to hide output in the progress view - this has the largest impact when building complex modules in CI (which uses plain progress output), like in our own CI (for example: https://github.com/dagger/dagger/actions/runs/12928659633/job/36056419424#step:5:209)

This was prompted by noticing our plain progress output has once again become kind of difficult to read - this is a fairly low hanging piece of fruit to tidy this up a bit.

### DIFF
--- a/dagql/dagui/opts.go
+++ b/dagql/dagui/opts.go
@@ -1,6 +1,7 @@
 package dagui
 
 import (
+	"slices"
 	"time"
 )
 
@@ -57,7 +58,7 @@ const (
 	ShowMetricsVerbosity      = 3
 )
 
-func (opts FrontendOpts) ShouldShow(span *Span) bool {
+func (opts FrontendOpts) ShouldShow(db *DB, span *Span) bool {
 	if opts.Debug {
 		// debug reveals all
 		return true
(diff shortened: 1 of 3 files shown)


dagql/dagui/opts.go
```diff
@@ -102,3 +116,45 @@ func (opts FrontendOpts) ShouldShow(span *Span) bool {
 	}
 	return true
 }
+
+func ShouldSkipFunction(obj, field string) bool {
+	// TODO: make this configurable in the API but may not be easy to
```
vito, line 121: I have some code in https://github.com/dagger/dagger/pull/9327 for filtering that would be helpful here:

https://github.com/dagger/dagger/pull/9327/files#diff-103e59fe01bf9664b34164f42a0052ad7c2432a4e3472900e99fe611b915b3d9

Feel free to carry on with this though - I can re-integrate + de-couple later.

I should try to land that PR though - maybe I'll just mark the API experimental for now.

jedevc: Ahh, that definitely looks useful :eyes: Being able to filter it better would definitely help.

Any objection to landing this in the meantime though, I'm happy to refactor it to match the newer API once it lands :tada:

vito (approved):
Note: this will hide these spans in the web UI too once I bump its dependency, which is probably fine in the 99.9% case, though we might miss them in the 0.1% case where you're trying to troubleshoot SDK development or something. Can reconsider later.

jedevc:
Ahh, yeah, I think that's *probably* okay.

Would it be worth having a similar verbosity feature in the web view, like we have in the TUI? Would be very useful, as we do cleanups around #9476 (which'll probably result in more stuff getting hidden away).

Issue #9257, opened 2024-12-19
Upgrade gqlgen dependency


jedevc:
Leaving this as a reminder - we need to upgrade `github.com/99designs/gqlgen`, but this is going to enable a bunch of linting rules for our graphql queries. This is good... but for some reason these are failing for very non-obvious reasons.

```
TransportQueryError: Error while fetching schema: {'message': 'Fragment 
"TypeRef" cannot be spread here as objects of type "__Type" can never be of type
"__Type".', 'locations': [{'line': 63, 'column': 8}], 'extensions': {'code': 
'GRAPHQL_VALIDATION_FAILED'}}
```

See https://github.com/dagger/dagger/pull/9187#issuecomment-2548555634.

Issue #6561, opened 2024-02-01
Strange error when querying core type without sub-selection


jedevc:
Given a very simple go module:

```go
package main

type Playground struct{}

func (m *Playground) Alpine() *Container {
	return dag.Container().From("alpine:latest")
}
```

Using `dagger call` without a sub-selection works as expected:

```
❯ dagger call alpine
✔ dagger call alpine [2.19s]
┃ Container evaluated. Use "dagger call alpine --help" to see available sub-commands.                                                                   
• Engine: e254c39a3421 (version devel ())
⧗ 15.50s ✔ 74
```

But using `dagger query` produces a very bizarre error:
```
❯ echo "{playground{alpine}}" | dagger query
✘ dagger query ERROR [2.30s]
├ [0.98s] loading module
✘ query ERROR [1.32s]
• Engine: e254c39a3421 (version devel ())
⧗ 2.69s ✔ 55 ∅ 9 ✘ 2
Error: make request: input: marshal: json: unsupported type: docker.RegistryHosts
input: json: unsupported type: docker.RegistryHosts
```

I remember this to give a more helpful graphql error (something like "must have sub-selection") - but it seems like we've lost this now? Maybe in the conversion to DagQL?

helderco:
> I remember this to give a more helpful graphql error (something like "must have sub-selection") - but it seems like we've lost this now? Maybe in the conversion to DagQL?

Not DagQL. That still happens but not on `Container`, `Directory`, `File` and `Terminal` which have a default sub-selection if none is given (sync for the first 3, websocketEndpoint for Terminal). 

So `dagger call alpine` produces this query: `query { playground { alpine { sync } } }`.

helderco:
The `unsupported type: docker.RegistryHosts` error though, yeah that is definitely a core issue, most likely related to DagQL. GraphQL should say it requires a sub-selection. 

It's not just on `Container`:

```
$ echo '{container}' | dagger-dev query
Error: make request: input: marshal: json: unsupported type: docker.RegistryHosts
```

helderco:
Tested in v0.12, no longer returns an error:

```console
echo "{playground{alpine}}" | dagger -m playground query
```
```json
{
    "playground": {
        "alpine": {
            "Class": {},
[...]

PR #9436, opened 2025-01-23
chore: update graphql dependencies

jedevc:
Fixes https://github.com/dagger/dagger/issues/9257 :tada:

This bumps all our relevant graphql dependencies. This is a bit trickier than just bumping stuff up, since previously we weren't definitely validating our queries! This wasn't really any issue, but we should have really been doing it all along. Also, the next version of gqlgen imports `rules`, so all validation gets enabled anyways, so we should get it working.

We also need to track `schema.PossibleTypes` for some reason - this is what actually allows the fragment expansion rule check to pass. Without this, it'll always fail.


jedevc:
Hm, marking as draft since it looks like enabling validation has actually caught a number of reasonably legitimate-looking failures.

jedevc:
Huh, this might fix https://github.com/dagger/dagger/issues/6561 as well :thinking:

image

jedevc:
@wingyplus looks like we have the same issue in the elixir SDK - all enum values get quoted: https://github.com/dagger/dagger/blob/dd4d69d11fb39cf9247fd3a6add99231133e88b6/sdk/elixir/lib/dagger/core/query_builder.ex#L57-L58

Is there a way to add to the guard here so that if the type is an enum (e.g. like `TypeDefKind`) we don't quote it?

dagql/server.go
```diff
@@ -317,19 +320,27 @@ func (s *Server) Schema() *ast.Schema { // TODO: change this to be updated whene
...
+		Types:         make(map[string]*ast.Definition),
+		PossibleTypes: make(map[string][]*ast.Definition),
+	}
 	for _, t := range s.objects { // TODO stable order
 		def := definition(ast.Object, t, s.View)
 		if def.Name == queryType {
 			schema.Query = def
 		}
 		schema.AddTypes(def)
+		schema.AddPossibleType(def.Name, def)
```
vito, line 333: Is this right? I think "possible types" is meant to be a mapping from unions and interfaces to their concrete implementations, whereas this seems to imply every type has itself as a possible type

dagql/server.go
```diff
@@ -398,6 +409,14 @@ func (s *Server) ExecOp(ctx context.Context, gqlOp *graphql.OperationContext) (m
 		if err != nil {
 			return nil, gqlErrs(err)
 		}
+
+		listErr := validator.Validate(s.Schema(), gqlOp.Doc)
```
vito, line 413: Yay!

vito (commented):
Thanks for catching this!

dagql/server.go
jedevc: Yeah, I don't know actually :thinking: 

But this validation rule implies that it is meant to be set like this? https://github.com/vektah/gqlparser/blob/e21b122b4e0ebb42ad98fe4ecf738bf380e487a3/validator/rules/possible_fragment_spreads.go#L36-L36

jedevc:
Going to disable the linting rule that checks that we've correctly set all the types - since doing is *technically* backwards incompatible. The PHP + elixir SDKs use strings instead of enums, while dagger shell sends pretty much all the types as strings (and fixing this neatly is a pain).

dagql/server.go
vito: Does it fail without it?

The linked is regarding fragment spreads which should only be relevant to type switching on interfaces/enums, but it's for sure weird that it seems to pass the spreaded (concrete) type to `GetPossibleTypes` considering its own godoc says it's only for interfaces/enums:

```go
// GetPossibleTypes will enumerate all the definitions for a given interface or union
func (s *Schema) GetPossibleTypes(def *Definition) []*Definition {
	return s.PossibleTypes[def.Name]
}
```

Either way if it doesn't fail I would just skip it - we don't use interfaces anyway (and I don't think we use unions either)

jedevc: It does fail, which was the reason I actually caught that linting wasn't ever enabled in the first place:

```
TransportQueryError: Error while fetching schema: {'message': 'Fragment 
"TypeRef" cannot be spread here as objects of type "__Type" can never be of type
"__Type".', 'locations': [{'line': 63, 'column': 8}], 'extensions': {'code': 
'GRAPHQL_VALIDATION_FAILED'}}
```

From https://github.com/vektah/gqlparser/blob/e21b122b4e0ebb42ad98fe4ecf738bf380e487a3/validator/rules/possible_fragment_spreads.go#L58-L68

I think potentially maybe this is just... wrong? Might submit a patch upstream.

jedevc: Ah, this is why it works - as part of loading the schema, it adds all `Type`s to `PossibleType`s :thinking: https://github.com/vektah/gqlparser/blob/28ef20ae1564c8a8ac8a0305711e0122fe937e9a/validator/schema.go#L74-L74

So we should do the same thing here.

jedevc:
Moving this to the next milestone, there's no rush for this one.

dagql/server.go
vito: Alright - seems weird, but I'm convinced it's at least consistent 😛

dagql/server.go
```diff
@@ -10,17 +10,35 @@ import (
...
+	//   - PHP + Elixir SDKs send enums quoted
+	//   - The shell sends enums quoted, and ints/floats as strings
+	//   - etc
+	validator.RemoveRule(rules.ValuesOfCorrectTypeRule.Name)
+	validator.RemoveRule(rules.ValuesOfCorrectTypeRuleWithoutSuggestions.Name)
+
+	// HACK: this rule is disabled because PHP modules <=v0.15.2 query
+	// inputArgs incorrectly.
+	validator.RemoveRule(rules.ScalarLeafsRule.Name)
+}
```
vito, line 40: Can these be cleaned up someday? 🤔

vito (approved):

Issue #9156, opened 2024-12-10
Make `sdk` field in `ModuleConfig` a `ModuleConfigDependency`


jedevc:
See original discussion in https://github.com/dagger/dagger/pull/8839#issuecomment-2519944200.

The work for pinging dependencies in #8587 didn't also apply to external SDKs. We should make the single `sdk` field into a `ModuleConfigDependency`, so that it can have both a `source` and a `pin`.

However, it *probably* doesn't make sense to also have the `name` field - we could just have it be a dangling field, or could maybe update `name` to be `omitempty` in this case if we like.

rajatjindal:
instead of pursuing the SDK field as `ModuleConfigDependency`, we decided to have a custom struct for it as SDK and dependencies are initialized differently in the code. 

```
type SDK struct {
	Source string            `json:"source"`
	Pin    string            `json:"pin,omitempty"`
	Env    map[string]string `json:"env,omitempty"`
}
```

- The source will store what is stored in "sdk" field today. so e.g. `go` or `github.com/dagger/dagger/php` etc.
- The pin will (when it is supported) store the pinned commit
- The env variable will contain a map (may change to a different structure) of key/value which will be injected automatically during code-generation and runtime. This will help us with configuring env variables like "GOPRIVATE" as an example.

The existing modules will work as usual, but when user will do "dagger develop", we will change the sdk from a string to the new struct automatically.

PR #9454, opened 2025-01-27
chore(mod): update sdk field in dagger.json from string to struct

rajatjindal:
This (is a WIP) stab at updating the dagger.json's sdk field from a string to a struct. 

fixes https://github.com/dagger/dagger/issues/9156

I am opening this PR to collect initial feedback while I am trying to figure out how we can extend this to include environment variables in the runtime container. (I am guessing we will need those env variable at `Codegen` time instead of `Runtime` (to ensure those work for codegen for the private module's dependency, but that is wht I am currently evaluating).

other things that we need to cover (possibly different PRs?):

- Pin the sdk dependency
- Add ability to update sdk
- Verify we can update to a different SDK?


core/modules/config.go
```diff
@@ -11,6 +11,43 @@ const Filename = "dagger.json"
...
+// can be a reference to a git path e.g. github.com/username/reponame/sdk-name
+type SDK struct {
+	Source string            `json:"source"`
+	Pin    string            `json:"pin,omitempty"`
+	Env    map[string]string `json:"env,omitempty"`
+}
+
+func (sdk *SDK) UnmarshalJSON(data []byte) error {
+	if sdk == nil {
+		return fmt.Errorf("cannot unmarshal into nil ModuleConfigDependency")
```
jedevc, line 25: nit:
```suggestion
		return fmt.Errorf("cannot unmarshal into nil SDK")
```

core/schema/module.go
```diff
@@ -1046,14 +1048,15 @@ func (s *moduleSchema) updateCodegenAndRuntime(
...
-	if mod.NameField == "" || mod.SDKConfig == "" {
+	if mod.NameField == "" || mod.SDKConfig == nil {
 		// can't codegen yet
 		return nil
 	}
 
 	if src.Self.WithInitConfig != nil &&
 		src.Self.WithInitConfig.Merge &&
-		mod.SDKConfig != string(SDKGo) {
+		// THIS WOULD PROBABLY CHANGE TO comparing struct?
```
jedevc, line 1058: Happy to leave it just comparing the `Source` for now.

core/module.go
```diff
@@ -15,6 +15,29 @@ import (
 	"github.com/dagger/dagger/engine/slog"
 )
 
+type SDKConfig struct {
+	Source string            `json:"source"`
+	Pin    string            `json:"pin"`
+	Env    map[string]string `json:"env"`
```
jedevc, line 21: To make these types queryable and appear in the schema, you'll need to add `field` tags, e.g. look at `ModuleSource`:

https://github.com/dagger/dagger/blob/539584e2515c32c2fb5fa053004c4c31e08ff0fa/core/modulesource.go#L58-L77

core/schema/modulesource.go
```diff
@@ -1378,7 +1378,9 @@ func (s *moduleSchema) collectCallerLocalDeps(
 				modCfg.Name = src.WithName
 			}
 			if src.WithSDK != "" {
-				modCfg.SDK = src.WithSDK
+				modCfg.SDK = &modules.SDK{
+					Source: src.WithSDK,
+				}
```
jedevc, line 1383: `WithSDK` would need changing as well - the method should be updated to additionally take `pin` (similar to how `moduleSource` takes it as well), and maybe even `env`? Happy to defer that for now, but IMO the `WithSDK` field should change to be the new `SDKConfig` type instead of a `string`.

jedevc (commented):
Hey, nice! Just left a few small comments, the direction looks right to me.

Happy to defer adding and making the `pin` field work (and working with updates, etc).

core/schema/module.go
```diff
@@ -1195,11 +1198,27 @@ func (s *moduleSchema) updateCodegenAndRuntime(
...
 
-	mod.Runtime, err = sdk.Runtime(ctx, mod.Deps, src)
+	runtime, err := sdk.Runtime(ctx, mod.Deps, src)
 	if err != nil {
 		return fmt.Errorf("failed to get module runtime: %w", err)
 	}
 
+	var runtime2 dagql.Instance[*core.Container]
+	err = s.dag.Select(ctx, runtime.Query.NewModule().Source, &runtime2,
+		dagql.Selector{
```
rajatjindal, line 1208: this part is not working currently.

core/module.go
rajatjindal: with map[string]string we get `cannot convert map[string]string to a Typed value` error. how do you feel about making the env either `[]string` (meaning user will have to type in KEY=VALUE format) OR making it `[]Env`, where `Env` is a struct with a key/value fields.

rajatjindal: well apparently []Env also fails, so I have to keep it as []string. @jedevc what do you think from UX perspective here?

jedevc: Hm, having `[]Env` *should* work - we already have `BuildArg`, see: https://github.com/dagger/dagger/blob/12f298aad280ca957f3c69e9a589c48f7118e8b0/core/schema/query.go#L61-L61

That should allow you to have it in the API.

However, the `ModuleConfig` struct should ideally still represent it as a `map[string]string` (should be fine, since it's serialized as JSON, and I don't *think* it's exposed in the API directly).

core/schema/sdk.go
```diff
@@ -629,6 +629,31 @@ func (sdk *goSDK) baseWithCodegen(
 				},
 			},
 		},
+	}
+
+	// inject sdk specific env variables before withExec
+	// TODO(rajatjindal): do we want to have a safelist here?
+	// e.g. disallow system env variables?
```
jedevc, line 636: Potentially I would filter to anything with a `GO` prefix? Is there any reason we'd want to inject non-go prefixes?

We could always expand the list later, but I'm struggling to think of a legit reason we'd want anything that's not in that list.

core/module.go
```diff
@@ -66,6 +66,32 @@ type Module struct {
 	InstanceID *call.ID
 }
 
+type SDKConfig struct {
+	Source string            `field:"true" name:"source" doc:"Source of the SDK. Either a name of a builtin SDK or a module source ref string pointing to the SDK's implementation."`
+	Pin    string            `field:"true" name:"pin" doc:"Commit sha pinned for the SDK"`
```
jedevc, line 71: I don't see the pin being used in this PR - so, I'm happy to remove it for now (so we don't block the important work), and we can add it back in later. Better to not have the field, than to have it partially implemented and confusing users.

core/module.go
```diff
@@ -66,6 +66,32 @@ type Module struct {
 	InstanceID *call.ID
 }
 
+type SDKConfig struct {
+	Source string            `field:"true" name:"source" doc:"Source of the SDK. Either a name of a builtin SDK or a module source ref string pointing to the SDK's implementation."`
```
jedevc, line 70: :thinking: One day, I wonder if we would want to split this up into a `source` and `builtin` field. But that's kind of out-of-scope for here, just wondering out loud.

jedevc (commented):
Left a couple comments - aside from those, I think this is ready to go. LGTM :tada:

core/modulesource.go
```diff
@@ -70,7 +70,7 @@ type ModuleSource struct {
 	WithoutDependencies       []string
 	WithUpdateDependencies    []string
 	WithUpdateAllDependencies bool
-	WithSDK                   string
+	WithSDK                   SDKConfig
```
jedevc, line 73: FYI https://github.com/dagger/dagger/pull/9505

cmd/dagger/config.go
```diff
@@ -54,7 +54,7 @@ dagger config -m github.com/dagger/hello-dagger
 		if err != nil {
 			return fmt.Errorf("failed to get module name: %w", err)
 		}
-		sdk, err := mod.SDK(ctx)
+		sdk, err := mod.SDK().Source(ctx)
```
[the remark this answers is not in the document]
rajatjindal: it does. I tried running dev engine and using client dagger v0.15.3, and I was able to run dagger calls there.

core/modules/config.go
```diff
@@ -43,6 +43,43 @@ type ModuleConfig struct {
 	Codegen *ModuleCodegenConfig `json:"codegen,omitempty"`
 }
 
+// SDK represents the sdk field in dagger.json
+// The source can be reference to a built-in sdk e.g. go, php, elixir or
+// can be a reference to a git path e.g. github.com/username/reponame/sdk-name
+type SDK struct {
+	Source string            `json:"source"`
+	Pin    string            `json:"pin,omitempty"`
```
jedevc, line 51: nit: `pin` is still here.

rajatjindal: thanks for catching this.

core/schema/sdk.go
```diff
@@ -633,6 +633,34 @@ func (sdk *goSDK) baseWithCodegen(
...
+
+			selectors = append(selectors, dagql.Selector{
+				Field: "withEnvVariable",
+				Args: []dagql.NamedInput{
+					{
+						Name:  "name",
+						Value: dagql.NewString(k),
+					},
+					{
+						Name:  "value",
```
rajatjindal, line 655: @jedevc, @marcosnils  I have been thinking a bit more about this specific code here. I think the runtime here will use the environment variable specified in the `top-level` module's SDK, but what do you think should happen if a dependency exposes an environment variable in SDK config?

jedevc: Bit confused by this - if a dependency exposes an env variable in it's own config, then *it's* runtime should get that environment variable. But the top-level module shouldn't be affected at all.

jedevc (approved):

core/schema/sdk.go
rajatjindal: ah. I keep forgetting deps get their own runtime/server etc. thanks.

core/schema/sdk.go
```diff
@@ -633,6 +633,34 @@ func (sdk *goSDK) baseWithCodegen(
 				},
 			},
 		},
+	}
+
+	// inject sdk specific env variables before withExec
+	// for now allow only Env with prefix GO
+	cfg, ok, _ := src.Self.ModuleConfig(ctx)
```
marcosnils, line 640: @rajatjindal @jedevc  IIUC this env thing will only work for the Go SDK right? did we create an issue somewhere so the other SDKs could potentially implement this as well?

core/schema/sdk.go
```diff
@@ -633,6 +633,34 @@ func (sdk *goSDK) baseWithCodegen(
...
+	if ok && cfg.SDK != nil {
+		for k, v := range cfg.SDK.Env {
+			if !strings.HasPrefix(k, "GO") {
```
marcosnils, line 643: Do we want to add this condition?  I think this will be confusing to some users. I don't see why we would like this to be this way.  For example, there are some Go related variables that don't start with `GO` like `CGO_ENABLED=1`.

rajatjindal: that is a fair point. are you running into a limitation already because of that? The idea was we can extend it on case by case basis to ensure we don't end up allowing something and then having to revert that later (e.g. for security considerations).

core/schema/sdk.go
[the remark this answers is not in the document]
rajatjindal: we have not created the issue for supporting other env variable as we thought we will allow them on case by case basis. on 2nd thoughts, we should also update all docs to reflect this new format of dagger.json -> sdk field? what do you think @jedevc ?

core/schema/sdk.go
marcosnils: > that is a fair point. are you running into a limitation already because of that? The idea was we can extend it on case by case basis to ensure we don't end up allowing something and then having to revert that later (e.g. for security considerations).

oh, I see what you mean.. yeah.. I don't have any particular requirement now. Given the security constraints, it seems ok to me to leave it like this for now :pray:

core/schema/sdk.go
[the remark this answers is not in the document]
jedevc: Mm, I think that makes sense. Currently we only use it for `GODEBUG` and `GOPROXY`. Given we don't document it, I'm happy to remove it.

However - I do actually think that configuring proxy settings are an `engine.json` level setting, *not* a module level setting. Proxies + mirrors *shouldn't* affect the reproducibility of a module right? So configuring them should probably be an infra concern (for whoever hosts the engine) - not a module level decision.

So, I think what this implies is that we should be able to configure these env vars in `engine.json` *as well* as in `dagger.json`. Does that make sense?

core/schema/sdk.go
jedevc: Split this to https://github.com/dagger/dagger/pull/9558.

PR #9459, opened 2025-01-27
fix(ts-sdk): object name conversion with numbers

### LINKED ISSUE #9451: 🐞 TypeScript SDK - Small bug in class naming
### What is the issue?

Reporting as requested from [this short exchange in Discord](https://discord.com/channels/707636530424053791/1157515934597136404/1330512719794671636). When the class name in a module has a number in it, it seems the SDK doesn't process it correctly. This is an example class with the issue:

```ts
import { dag, Container, object, func, Directory } from "@dagger.io/dagger"

@object()
export class M8A {
   /**
   * Returns a container that echoes whatever string argument is provided
   * @param stringArg A text to echo in the container.
   */
  @func()
  containerEcho(stringArg: string): Container {
  }
}
``` 

dagger.json
```json
{
  "name": "m8a",
  "engineVersion": "v0.15.2",
  "sdk": "typescript",
  "source": "dagger"
}
```

### Dagger version

v0.15.2

### Steps to reproduce

- Create a folder with the name like m8a. Make sure a number is in the name.

- Run `dagger init --sdk=typescript --source=./dagger` in this folder. 

- Notice the newly created class has the name "M8A". 

- Run `dagger call container-echo --stringArg "hello dolly"`

### Log output

You'll get this error:

```
✘ .initialize: Module! 4.4s
could not find module entrypoint: class M8a from import. Class should be exported to benefit from all features.
/src/dagger/sdk/src/module/introspector/dagger_module/module.ts:98
      throw new IntrospectionError(
            ^


IntrospectionError: could not find main object M8a in module  located at /src/dagger/src/index.ts,/src/dagger/src/
er.entrypoint.ts,/src/dagger/src/../sdk/src/api/client.gen.ts
    at new DaggerModule (/src/dagger/sdk/src/module/introspector/dagger_module/module.ts:98:13)
    at scan (/src/dagger/sdk/src/module/introspector/index.ts:18:18)
    at async connection.LogOutput (/src/dagger/sdk/src/module/entrypoint/entrypoint.ts:17:26)
    at async <anonymous> (/src/dagger/sdk/src/connect.ts:44:11)
    at async withGQLClient (/src/dagger/sdk/src/common/graphql/connect.ts:24:12)
    at async <anonymous> (/src/dagger/sdk/src/connect.ts:40:9)
    at async connection (/src/dagger/sdk/src/connect.ts:38:5)
    at async entrypoint (/src/dagger/sdk/src/module/entrypoint/entrypoint.ts:11:3) {
  cause: undefined,
  code: 'D110'
}

Node.js v22.11.0
! failed to initialize module: failed to call module "m8a" to get functions: process "tsx --no-deprecation --tsconfig /src/dagger/tsconfig.json /src/dagger/src/__dagger.entrypoint.ts" did not complete successfully: exit code: 1
```

If you replace the class name with "M8a", the function will work. 

(NOTE: this is missing in your issue description) What outcome is expected?

What I'd expect is either the init process to correctly create the class as "M8a", or the TS SDK to correctly find the module with the "M8A" name. 

Scott

TomChv:
### Summary

I fixed a small problem linked to module name that contains a number. The generated class name wasn't consistent with the name search by the introspection.
Basically, a module named `m8a`  generated as `M8A` but converted to `M8a` by the introspector instead of `M8A`.
I also added a bunch of tests cases to verify other format

Result of the #9451 reported issue with this branch version:

Screenshot 2025-01-27 at 18 29 36

Fixes #9451

### DIFF
--- a/sdk/typescript/src/module/introspector/test/case_convertor.spec.ts
+++ b/sdk/typescript/src/module/introspector/test/case_convertor.spec.ts
@@ -33,5 +33,30 @@ describe("case convertor", function () {
       const result = convertToPascalCase("HelloWorld")
       assert.equal(result, "HelloWorld")
     })
+
+    it("should correctly handle numbers as spacers", function () {
+      const result = convertToPascalCase("hello123world")
+      assert.equal(result, "Hello123World")
+    })
+
+    it("should correctly handle number as first word", function () {
+      const result = convertToPascalCase("123helloworld")
+      assert.equal(result, "123Helloworld")
+    })
+
+    it("should correctly handle number as last word", function () {
+      const result = convertToPascalCase("helloWorld123")
+      assert.equal(result, "HelloWorld123")
+    })
+
+    it("should correctly handle special characters as spacers", function () {
+      const result = convertToPascalCase("hello world")
+      assert.equal(result, "HelloWorld")
+    })
+
+    it("should correctly handle mix cases", function () {
+      const result = convertToPascalCase("123hello-world_here")
+      assert.equal(result, "123HelloWorldHere")
+    })
   })
 })
--- a/sdk/typescript/src/module/introspector/case_convertor.ts
+++ b/sdk/typescript/src/module/introspector/case_convertor.ts
@@ -1,18 +1,15 @@
 export function convertToPascalCase(input: string): string {
+  // Handle empty string case
+  if (!input) {
+    return ""
+  }
+
+  // Split on word boundaries before uppercase letters, numbers, and special characters
   const words = input
-    .replace(/[^a-zA-Z0-9]/g, " ") // Replace non-alphanumeric characters with spaces
-    .split(/\s+/)
+    .split(/(?=[A-Z0-9])|[^a-zA-Z0-9]|(?<=[a-zA-Z])(?=\d)|(?<=\d)(?=[a-zA-Z])/g)
     .filter((word) => word.length > 0)
 
-  if (words.length === 0) {
-    return "" // No valid words found
-  }
-
-  // It's an edge case when moduleName is already in PascalCase or camelCase
-  if (words.length === 1) {
-    return words[0].charAt(0).toUpperCase() + words[0].slice(1)
-  }
-
+  // Convert each word to proper case
   const pascalCase = words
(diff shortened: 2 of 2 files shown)


sdk/typescript/src/module/introspector/test/case_convertor.spec.ts
```diff
@@ -33,5 +33,30 @@ describe("case convertor", function () {
...
+      const result = convertToPascalCase("helloWorld123")
+      assert.equal(result, "HelloWorld123")
+    })
+
+    it("should correctly handle special characters as spacers", function () {
+      const result = convertToPascalCase("hello world")
+      assert.equal(result, "HelloWorld")
+    })
+
+    it("should correctly handle mix cases", function () {
```
jedevc, line 57: Slightly curious how it handles something like `introspectionJSON` (where the second word is an acronym). Might be worth checking that that case also works (since those style of acronyms have caused other problems in other SDKs before).

jedevc (approved):

sdk/typescript/src/module/introspector/test/case_convertor.spec.ts
TomChv: I'll add a case, nice catch

jedevc: Ah, this feels very similar to the corresponding go sdk/engine issue as https://github.com/dagger/dagger/issues/7941

We need to redo a lot of string-casing logic and make it entirely consistent everywhere. However, this is such a pain, and so completely breaking, I'm still not sure how we do it properly (which is why the work on https://github.com/dagger/dagger/pull/8122 stalled).

TomChv: Tests cases added, however, like we discussed on Discord, it will still fails on user's end because the engine generates `introspectionJson` instead of `introspectionJSON`

jedevc: Yup, SGTM, good to merge whenever. Let's just fix the immediate pain, we need to circle back round to this eventually one day anyways.

PR #9209, opened 2024-12-15
Docs: automatically record gifs

shykes:
## Overview

This PR does two things:

1) add the necessary tooling to automatically record gifs for our docs
2) demonstrate the tooling with 3 gifs: `build.gif`, `build-export.gif` and `build-publish.gif` from the [programmable pipelines section](http://docs.dagger.io/features/programmable-pipelines).

Note: the generation pipeline is not hooked into CI, so at the moment it must be called manually, like so:

```
dagger shell --no-mod <<EOF
./docs/static/img | generate | export ./docs/static/img
EOF
```

To generate the gifs without a local checkout (read only):

```
dagger shell <<EOF
github.com/shykes/dagger/docs/static/img@docs-recorder |
generate |
export ./dagger-gifs
EOF
```

## Example gifs

| | Before | After |
| -- | -- | -- |
| `build.gif` |  |  |
| `build-publish.gif` |  |  |
| `build-export.gif` |   |  |

- [build.gif](https://github.com/dagger/dagger/pull/9209/files#diff-fbe35f82d9c89b6e929db848ef11164020e293b49509dc295c4ec2361f2888a1)
- [build-publish.gif](https://github.com/dagger/dagger/pull/9209/files#diff-d47b8bf6dd4bee92c32e35267a67715a0c78f0690c164978334fe27e0c7b93c0)
- [build-export.gif](https://github.com/dagger/dagger/pull/9209/files#diff-4e5601582075bf91c360406ebb6233cd38318a625dad37e0ada9cb23d35d21f3)

## Recorder module

The recorder module is pretty neat. You can use it to quickly record your own gifs using the current dagger CLI:

```
dagger shell  <<EOF

github.com/shykes/dagger/docs/recorder@docs-recorder |
exec "uname -a" |
exec "dagger core version" |
exec "dagger core container from --address=alpine with-exec --args=ls,-l stdout" |
gif |
export ./demo.gif

EOF
```

## Making it feel fast

All commands are run twice: a first time to warm up the dagger cache; a second time in the recording. This allows calls to the dagger CLI to feel nice and fast in the recording.

PR #9462, opened 2025-01-27
docs: recorder tooling (#9209)

jpadams:


jpadams:
Todo after this commit would be to complete any features needed for https://github.com/dagger/dagger/pull/9209#issuecomment-2567119640

and to tie it into CI

kpenfound (approved):
Looks good! Was there any discussion about putting this under `/modules`? I don't know if that's better or not

jpadams:
> Looks good! Was there any discussion about putting this under `/modules`? I don't know if that's better or not

@shykes any reason  why it's better under `/docs` vs under `/modules`?

shykes:
Would be better to keep my commits separate, so we can go back in the history to see what I committed and what was changed/fixed after the fact.

shykes:
The idea of putting it under `docs` is that it's not a generic recorder module, it's the tooling specifically to record examples for our docs. If it does become generalizable, we can always spin it out later (but there's already termcast for that).

Also in my PR there were other more specialized modules, one for each snippet I believe.

jpadams:
> Would be better to keep my commits separate, so we can go back in the history to see what I committed and what was changed/fixed after the fact.

Ok, I can try to do that

jpadams:
> Also in my PR there were other more specialized modules, one for each snippet I believe.

There was a bunch of snippet merge conflict, so I tried to avoid it, will look back

jpadams:
will follow up with the snippets separately

PR #9465, opened 2025-01-28
Check for TF_BUILD on Azure Pipelines

jasonmccallister:
Checks if the `TF_BUILD` environment variable is set and marks the run as CI with the vendor as Azure.

@vito asking for your review as you most recently edited 😄


vito (approved):
Don't know a whole lot about this var or Azure, but a quick Google makes it seem legit at least!

jedevc (approved):

PR #9466, opened 2025-01-28
Adds a schema reference page

jasonmccallister:
Adds a schema reference page with descriptions of each fields.

- Added a module to automate the generation of markdown from JSON schema (`dagger call docs generate-reference-pages export --path docs/current_docs/api/testing.mdx`)


vikram-dagger:
Is there a way to generate this page programmatically? How do we keep it updated if the schema changes?

jasonmccallister:
@vikram-dagger that's a really good idea, I could make a function under `docs` module for that... What do you think?

jedevc:
Big :+1: to automation here!

`docs` doesn't have a dedicated module yet - a lot of the logic is in the main module at `.dagger/docs.go`. I'd say it should go in there (or you can honestly just try splitting it all out to a separate `docs` module as well, that's the direction we want long-term anyways).

There's a few auto-generated functions in there already that you should be able to use as inspiration - potentially it's possible to just take the `.schema.json` as input, and generate a markdown file using a go template?

jasonmccallister:
@jedevc @vikram-dagger added the function to the docs module (in opening comment). It generates the following markdown. Also, happy to discuss the best way we want to (or currently) automate this process - although I doubt it changes that much.

Also, thoughts on updating this to include the dagger engine configuration reference as well?

Output:
```markdown
---
slug: /api/dagger-schema-reference
---

# dagger.json Schema Reference

The `dagger.json` file defines the configuration for a Dagger module. Below is a detailed description of each field, including its type, description, and whether it is required.

**Note**: The JSON schema file for configuring your IDE is available at [this URL](https://docs.dagger.io/reference/dagger.schema.json).

## Main Configuration

- **`codegen`** (Optional) — Codegen configuration for this module.

  **Type:** 

- **`$schema`** (Optional) — The self-describing json $schema

  **Type:** string

- **`include`** (Optional) — Paths to explicitly include from the module, relative to the configuration file.

  **Type:** array

- **`exclude`** (Optional) — Paths to explicitly exclude from the module, relative to the configuration file.

  **Type:** array

- **`dependencies`** (Optional) — The modules this module depends on.

  **Type:** array

- **`source`** (Optional) — The path, relative to this config file, to the subdir containing the module's implementation source code.

  **Type:** string

- **`views`** (Optional) — Named views defined for this module, which are sets of directory filters that can be applied to directory arguments provided to functions.

  **Type:** array

- **`name`** (Required) — The name of the module.

  **Type:** string

- **`engineVersion`** (Required) — The version of the engine this module was last updated with.

  **Type:** string

- **`sdk`** (Optional) — The SDK this module uses

  **Type:** string

## Nested Blocks

### `ModuleCodegenConfig`

- **`automaticGitignore`** (Optional) — Whether to automatically generate a .gitignore file for this module.

  **Type:** boolean

### `ModuleConfigDependency`

- **`pin`** (Optional) — The pinned version of the module dependency.

  **Type:** string

- **`name`** (Required) — The name to use for this dependency. By default, the same as the dependency module's name, but can also be overridden to use a different name.

  **Type:** string

- **`source`** (Optional) — The source ref of the module dependency.

  **Type:** string

### `ModuleConfigView`

- **`name`** (Required) — 

  **Type:** string

- **`patterns`** (Optional) — 

  **Type:** array
```

docs/current_docs/api/dagger-schema-ref.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+- **`patterns`** (Optional) — A list of patterns for directory filters.
+
+  **Type:** list(string)
+
+### `codegen`
+
+A `codegen` block supports the following:
+
+- **`automaticGitignore`** (Optional) — Whether to automatically generate a `.gitignore` file for this module.
```
jedevc, line 83: Huh, I forgot this existed :laughing: 

One tiny nit for this one, I wonder if it's possible to explain what the default value is - for most of the optionals, I don't think we need to (it's self explanatory), but for this one it might make sense.

docs/sidebars.ts
jedevc, line 1: nit: this looks like it got reformatted.

.dagger/docs.go
```diff
@@ -198,6 +199,89 @@ func (d Docs) GenerateSchemaReference() *dagger.Directory {
...
+	}
+
+	var s strings.Builder
+
+	s.WriteString("---\n")
+	s.WriteString("slug: /api/dagger-schema-reference\n")
+	s.WriteString("---\n\n")
+	s.WriteString("# dagger.json Schema Reference\n\n")
+	s.WriteString("The `dagger.json` file defines the configuration for a Dagger module. Below is a detailed description of each field, including its type, description, and whether it is required.\n\n")
+	s.WriteString("**Note**: The JSON schema file for configuring your IDE is available at [this URL](https://docs.dagger.io/reference/dagger.schema.json).\n\n")
```
jedevc, line 239: nit: could use a go multi-line string here.

.dagger/docs.go
```diff
@@ -198,6 +199,89 @@ func (d Docs) GenerateSchemaReference() *dagger.Directory {
 	return dag.Directory().WithFile(generatedAPIReferencePath, generatedHTML)
 }
 
+func (d Docs) GenerateReferencePages(ctx context.Context) (*dagger.File, error) {
```
jedevc, line 202: If we call this function in `Docs.Generate` then it'll automatically be updated when we make changes (and it'll also be caught in linting if it's out of date).

.dagger/docs.go
```diff
@@ -198,6 +199,89 @@ func (d Docs) GenerateSchemaReference() *dagger.Directory {
...
+	dir := d.GenerateConfigSchemas()
+
+	contents, err := dir.File(generatedDaggerJSONSchemaPath).Contents(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("error reading contents: %v", err)
+	}
+
+	var schema struct {
```
jedevc, line 210: Could instead use `"github.com/invopop/jsonschema"` here? No objection to pulling it in as a dependency.

jedevc (commented):
Eyyy nice, this looks really good! Couple points, otherwise LGTM :tada:

jedevc:
FYI, on doing the dagger engine config like this - would be really nice as well, but we already have a page for that (that has a lot of explanation for it, contrasting how to do it in buildkit config vs in dagger config). Maybe if have a page like introduced into this PR, we could link to it from the existing page?

docs/sidebars.ts
jasonmccallister: Yeah, zed decided it knows what's best for typescript 😄

docs/current_docs/api/dagger-schema-ref.mdx
jasonmccallister: >I wonder if it's possible to explain what the default value is - for most of the optionals

That's not set by default on a new module, does this refer to the `dagger develop` command that adds the SDK for the module?

.dagger/docs.go
jasonmccallister: Added to the function and updated the return type to `*dagger.Directory` to match.

.dagger/docs.go
jasonmccallister: I did give that a go and it was a little more tricky. I decided the just make a custom struct with the fields we need. Want me to update the schema to use that package?

jedevc: As-is is fine, just a suggestion in case you wanted to!

.dagger/docs.go
```diff
@@ -159,7 +160,8 @@ func (d Docs) Generate() *dagger.Directory {
 		WithDirectory("", d.GenerateSchema()).
 		WithDirectory("", d.GenerateCli()).
 		WithDirectory("", d.GenerateSchemaReference()).
-		WithDirectory("", d.GenerateConfigSchemas())
+		WithDirectory("", d.GenerateConfigSchemas()).
+		WithDirectory("", d.GenerateSchemaReference())
```
jedevc, line 164: :thinking: `GenerateSchemaReference` is repeated twice in this list, should this be `GenerateReferencePage` instead?

PR #9467, opened 2025-01-28
Fix format stdout

charjr:
The custom module `always-exec` had been removed from the dev module but I mistakenly left it's usage in one of the less frequently used methods.

This removes it and fixes the method.

Also upon using 0.15.2 it automatically updated a few Go dependencies, which I assume are okay to commit.


jedevc (approved):
Nice catch, thanks! Should've realized that got missed oops :tada:

Issue #9062, opened 2024-11-27
🐞 CLI continues running when the Engine stops running


gerhard:
### What is the issue?

The CLI continues to run **after** the Engine stops running. This is particularly problematic when the Engine exits / gets killed and the CLI continues to run.

Furthermore, pressing `q` **once** does not quit the CLI. In other words, the CLI is not able to terminate gracefully when it's not connected to the Engine.

### Dagger version

dagger v0.14.0 (registry.dagger.io/engine:v0.14.0) darwin/amd64

### Steps to reproduce

1. Run `dagger call docs lint`
2. Wait for the module to load
3. Now stop the Engine
4. Observe a variation of the following:

<img width="1552" alt="image" src="https://github.com/user-attachments/assets/ab046737-9fd7-4783-9594-310ac6c98c12">

Notice that the CLI does not exit

### Log output

Note: this particular Engine instance was stopped & started a few times.

[dagger-engine-0-14-0-stopped.txt](https://github.com/user-attachments/files/17938258/dagger-engine-0-14-0-stopped.txt)

PR #8995, opened 2024-11-19
ci: move engine and cli tests to gha management

jedevc:
This PR helps move *even* more things into our GHA actions module management, and out of manual management.

This contains all of the engine and cli jobs :tada:

PR #9474, opened 2025-01-29
Improve releasing during v0.15.3

jedevc:
Bit of a bumpy ride:
- The `githubPublish` function was wrong, so those tags weren't created - and were manually done using `gh release create`
- The pretty markdown report wasn't actually being displayed :scream: Also errors were escaping from there, so added a quick little hack while #8421 is unresolved.


gerhard (commented):
All changes look good, apart from the failing `engine lint` which is a genuine issue:

```
✘ .withExec(args: ["diff", "-r", "a/go.mod", "b/go.mod"]): Container! 6.3s
9c9
<       go.opentelemetry.io/otel v1.27.0
---
>       go.opentelemetry.io/otel v1.34.0
17c17
<       go.opentelemetry.io/otel/trace v1.27.0
---
>       go.opentelemetry.io/otel/trace v1.34.0
23a24,25
> require go.opentelemetry.io/auto/sdk v1.1.0 // indirect
>
32d33
<       github.com/stretchr/testify v1.10.0 // indirect
36c37
<       go.opentelemetry.io/otel/metric v1.27.0
---
>       go.opentelemetry.io/otel/metric v1.34.0
```

I am trying to figure out how to find which `go.mod` is failing since we have 123 in our repository and all that I see in this output is `a/go.mod` & `b/go.mod`.

gerhard:
I've narrowed it down to:

```
[".dagger/", ".github/", "cmd/dagger/.dagger/", "dagql/idtui/viztest/", "engine/distconsts/", ".", "helm/.dagger/", "modules/alpine/", "modules/compatcheck/", "modules/daggerverse/", "modules/dirdiff/", "modules/gha/examples/go/", "modules/gha/", "modules/go/", "modules/golangci/", "modules/graphql/", "modules/markdown/", "modules/ps-analyzer/", "modules/ruff/", "modules/shellcheck/", "modules/wolfi/", "releaser/", "sdk/dotnet/dev/", "sdk/elixir/runtime/", "sdk/go/", "sdk/php/runtime/", "sdk/python/runtime/", "sdk/typescript/runtime/", "version/"]
```

gerhard:
I thought https://v3.dagger.cloud/dagger/traces/1f7b89c35ed20bb5f778df58df26b9f6?span=8e279fb009710261 meant `cmd/dagger/.dagger` but there are other lint failures locally. Still trying to get to the bottom of it. FTR:

image

releaser/util.go
```diff
@@ -49,7 +49,7 @@ func (r Releaser) githubRelease(
 		return err
 	}
 
-	commit, err := dag.Version().Git().Commit(target).Commit(ctx)
+	commit, err := dag.Version().Git().Commit(src).Commit(ctx)
```
gerhard, line 52: Are you sure that we want `src` here @jedevc ?

If we do, then we should remove the `target` argument because it is unused and `engine lint` fails.

gerhard (changes requested):
Found the issue! Left a comment & pushed my local fixes. I was wondering about something in my last commit message @jedevc 🤔

gerhard:
I have one more question @jedevc: where does bumping the Dagger version used in the alternative CI runners fit in the new release automation?

This is what I used to run locally:

```console
❯ jd bump 0.15.2 0.15.3
find . \( -name '*.yml' -o -name '*.yaml' \) -type f -exec gsed -i 's/0-15-2/0-15-3/g; s/0\.15\.2/0\.15\.3/g' {} +

❯ which jd
jd: aliased to just --justfile=/Users/gerhard/github.com/dagger/justfile
```

jedevc:
It's still a manual process, I was hoping to unify all the bumping into the github module eventually so that no github yaml files ever need to be updated - but with the increased number of alternative runners, I'm not sure if that's realistic anymore.

Ideally we would have a function to do these bumps so that it's less manual - not quite sure why those are getting missed, I'm just using the find-replace functionality in my editor.

releaser/util.go
jedevc: Fixed, it should have been used later. Clarified further to rename these to `src` and `dest`.

gerhard:
Python does not look happy @helderco. Running it locally to see if I can reproduce.

gerhard:
It worked locally the first time, must be a remote flake, re-running.

image

gerhard:
I am stumped as to why `test-everything-else` started failing all of a sudden. Letting it run locally in the background.

gerhard:
This does not complete within 30 mins locally, it failed with the usual timeout. Will try running it elsewhere to see if I can reproduce the issue.

gerhard:
I have been re-running this a few times, in multiple instances, and the mystery continues.

The job keeps failing consistently in our CI, even though the hosts are definitely recycled, and there is no cache between runs 🤔 https://v3.dagger.cloud/dagger/ci?branch=improve-releasing-during-v0.15.3&ref=ea92b3ce1388ac60bcdf8497272576afa577e232&repo=github.com%2Fdagger%2Fdagger

image

Sometimes, the GitHub jobs gets cancelled 🤷‍♂️ https://github.com/dagger/dagger/actions/runs/13070101851/job/36597645873?pr=9474#step:5:135
image

Other times, the job times out 😨 https://github.com/dagger/dagger/actions/runs/13070101851/job/36477794204?pr=9474#step:5:135
image

And yet, running the same job in our experimental runner fails **consistently** when backed by an NVMe disk with different errors:
- https://dagger.cloud/gerhard/traces/218ef5764501160e4e7799748dd0c94d#b7b43803446b1acf
- https://v3.dagger.cloud/gerhard/traces/2fd3ff2201c3fa9c469ee39f9fdacf3f#1e0dec790842d82d

And it **succeeds consistently** when backed by TMPFS:
- https://v3.dagger.cloud/gerhard/traces/fe1476f9f6814d1ebd26d64073c4fe4f#a61b4f5b301318de

jedevc:
Hm, I wonder if the test engine is getting OOM-killed it looks like the timing out tests are trying desperately to connect to an engine that died after existed after being SIGKILLed (exit code 137), without anything in the logs to indicate what it was.

That's really odd though, I'm not sure why specifically dagger v0.15.3 would make the *internal* engine more memory hungry.

gerhard:
I am going to schedule another job now and pay special attention to the memory behaviour on this host.

gerhard:
Having taken a close look at the resource usage, this workload is CPU bound:

image

image

That sharp drop in memory is indeed an OOM kill:
image

My instinct tells me that we should:
- 2x the runner size in `runs-on:` as a quick fix - can you help @jedevc ?
- investigate the delta between `test-everything-else` and see where this new behaviour comes from - can you help @cwlbraa ?

The reason why the job eventually times out is because we are hitting this other bug:
- https://github.com/dagger/dagger/issues/9062

jedevc:
Ah, looks like I missed this: https://github.com/dagger/dagger/pull/8995/files#diff-a030e946fa7797995c9a82afa63d40933e52e0c89ac89f5afdcd2f874715730bL122-L123

Previously, this was using 32 cores, but I accidentally downgraded it to 16 cores.

gerhard:
Yes indeed! I'm pretty sure that I missed that when reviewing those thousands of lines of YAML.

Let me check one more thing before we merge this.

PR #9478, opened 2025-01-29
fix: netstat: compensate for netns recycling

cwlbraa:
previously the withexec  network tx and rx numbers would run up over the course of a call or even multiple calls. this is because buildkit likes to re-use network namespaces when it can, so the stats for the namespace persist between individual calls. 

this PR samples a baseline for the namespace on construction.

it's totally possible there are other ways to bug out these numbers, knowing what we know. like if one of these recyclable netnses can be concurrently re-used by multiple execs, the numbers will not reflect what we want them to reflect. without forcing all containers to have their own netnses, that wackiness is unavoidable, but I have yet to see evidence that that can occur.

### DIFF
--- a/engine/buildkit/resources/netstat.go
+++ b/engine/buildkit/resources/netstat.go
@@ -15,15 +15,16 @@ type BKNetworkSampler interface {
 }
 
 type netNSSampler struct {
-	netNS       BKNetworkSampler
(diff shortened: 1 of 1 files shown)


sipsma (approved):
🚀

PR #9483, opened 2025-01-30
Cache function calls per-client to avoid duped telemetry 

sipsma:
We previously needed to make function calls impure to handle various
corner cases around Secrets (covered by TestModule/TestSecretNested) in #8358.

However, recently sometime this resulted in us getting duped telemetry
output for calls. It makes sense in that a call may often get
re-evaluated (i.e. it's a base image for multiple vtxs in a DAG). It's
not clear why this only started happening recently though.

Either way, we can fix it by having function calls be cached per-client
via a custom cache key func.

This fixes duped output as repro'd by running:
```
dagger -m github.com/vito/bass call integration --runtime Buildkit
```
* Previously you'd see multiple apko->wolfi outputs appear, now it only
  shows up once as expected.

But keeping the caching per-client rather than fully cached allows us to
still handle the corner cases around Secrets.
* In the long run we'll need to re-arrange how secret transfer works
  (secret providers help a lot here) but this fix should suffice for
  now to at least avoid duped output.

---

Snuck in an extra quick fix as a separate commit too: [hide __schemaJSONFile from telemetry](https://github.com/dagger/dagger/commit/2212ae30a7640d66a907809103865c662d9f3f80)

### DIFF
--- a/core/cachekey.go
+++ b/core/cachekey.go
@@ -1,4 +1,4 @@
-package schema
+package core
 
 import (
 	"context"
@@ -18,8 +18,12 @@ const (
 // CachePerClient is a CacheKeyFunc that scopes the cache key to the client by mixing in the client ID to the original digest of the operation.
 // It should be used when the operation should be run for each client, but not more than once for a given client.
 // Canonical examples include loading client filesystem data or referencing client-side sockets/ports.
-func CachePerClient[P dagql.Typed, A any](ctx context.Context, _ dagql.Instance[P], _ A, origDgst digest.Digest) (digest.Digest, error) {
-	// scope the cache key to the client by mixing in the client ID to the original digest
+func CachePerClient[P dagql.Typed, A any](ctx context.Context, inst dagql.Instance[P], args A, origDgst digest.Digest) (digest.Digest, error) {
+	return CachePerClientObject(ctx, inst, args, origDgst)
+}
+
+// CachePerClientObject is the same as CachePerClient but when you have a dagql.Object instead of a dagql.Instance.
+func CachePerClientObject[A any](ctx context.Context, _ dagql.Object, _ A, origDgst digest.Digest) (digest.Digest, error) {
 	clientMD, err := engine.ClientMetadataFromContext(ctx)
 	if err != nil {
 		return "", fmt.Errorf("failed to get client metadata: %w", err)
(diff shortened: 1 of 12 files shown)


core/cachekey.go
```diff
@@ -20,17 +20,21 @@ const (
 // Canonical examples include loading client filesystem data or referencing client-side sockets/ports.
 func CachePerClient[P dagql.Typed, A any](ctx context.Context, _ dagql.Instance[P], _ A, origDgst digest.Digest) (digest.Digest, error) {
 	// scope the cache key to the client by mixing in the client ID to the original digest
+	return mixDgstWithClientID(ctx, origDgst)
+}
+
+func mixDgstWithClientID(ctx context.Context, origDgst digest.Digest) (digest.Digest, error) {
```
cwlbraa, line 26: meganit: just say digest lol

cwlbraa (approved):
lgtm code-wise. 

iirc correctly, we can treat secrets as pure within a given session because they persist across the session? whereas started in a new session we'd need to re-evaluate them?

core/object.go
```diff
@@ -365,6 +366,10 @@ func (obj *ModuleObject) installConstructor(ctx context.Context, dag *dagql.Serv
 				Server:       dag,
 			})
 		},
+		// cache constructor calls per client; a given client will hit cache when making the same call repeatedly
+		func(ctx context.Context, _ dagql.Object, _ map[string]dagql.Input, origDgst digest.Digest) (digest.Digest, error) {
+			return mixDgstWithClientID(ctx, origDgst)
+		},
```
jedevc, line 372: nit: it would be really nice to be able to use `CachePerClient` here as well - I guess it's the same as the difference between `NodeFunc` and `Func`.

Even if we can't abstract them nicely (I tried lol, for the middlewares in #9395, it's not easy), it might be nice to have this as a predefined function.

core/typedef.go
```diff
@@ -77,15 +77,6 @@ func (fn *Function) FieldSpec() (dagql.FieldSpec, error) {
 		Name:        fn.Name,
 		Description: formatGqlDescription(fn.Description),
 		Type:        fn.ReturnType.ToTyped(),
-
-		// NB: functions actually _are_ cached per-session, which matches the
```
jedevc, line 81: :eyes: can we keep this comment (in a new place) - just thinking about accidental removal, etc (though tests *should* catch it).

jedevc:
> iiuc, we can treat secrets as pure within a given session because they persist across the session? whereas started in a new session we'd need to re-evaluate them?

@cwlbraa yup, that's pretty much it. There's a *lot* more context in #8358, but yes, within a single session it's safe, the only reason it wasn't done this way originally is that we couldn't mark a field as "sometimes" pure/impure, the cache keys let us be a lot more flexible now.

jedevc (approved):

core/object.go
sipsma: Yeah I couldn't find a way to convince the go compiler with generics to have a single one, happy to add a separate pre-defined func for this since

PR #9500, opened 2025-02-03
shell: fix repl output

helderco:
This fixes a shell regression in v0.15.3, but only in interactive mode.

I don't fully understand why but `Frontend.Background` can swallow output if you don't add a newline (`fmt.Fprint` vs `fmt.Fprintln`).

Current solution isn't perfect. You may see extra newlines than intended. For example:

```
⋈ _echo hello
hello

⋈ _echo -n hello
hello
⋈
```

That happens because the `_echo` builtin, for example, sends two separate writes with `_echo hello`:
```
"hello"
"\n"
```
So on the first `"hello" we're adding a `\n`, and then the second one comes but it's still printed.

I tried another approach but didn't work out and ran out of time. Still need to find something better, maybe on bubbletea's side.

### DIFF
--- a/cmd/dagger/shell.go
+++ b/cmd/dagger/shell.go
@@ -99,6 +99,9 @@ type terminalWriter struct {
 }
 
 func (o *terminalWriter) Write(p []byte) (n int, err error) {
+	if len(p) == 0 {
+		return 0, nil
+	}
 	if o.processFn != nil {
 		r, err := o.processFn(p)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


helderco:
@jedevc, also need to add test coverage for the repl so we catch this sort of stuff. Any idea how best to test it? Would the kind of tests for `Terminal` work here?

jedevc (approved):

jedevc:
> Any idea how best to test it? Would the kind of tests for Terminal work here?

Yeah, we'd be looking at the `Terminal` tests for inspiration. But, IMO, we should work to build some test utilities around that, since those tests are kind of sprawling (and used to be quite delicate and flaky - so having some neater wrapped logic would be good).

PR #9501, opened 2025-02-03
shell: allow "exit" in the repl

helderco:
Noticed some people expect to be able to type "exit" to leave the shell REPL.

There's an `exit` builtin and at first I just delegated to it, but it requires the interp runner to run, only to exit right after. So this feels simpler and snappier. Builtin is still available at `_exit`, but it won't exit the REPL. I tried that but got tangled with knowing the difference between exiting a prompt vs exiting the REPL.


jedevc (approved):

jedevc:
Failure in `test-everything-else` is probably the same as being discussed in https://github.com/dagger/dagger/pull/9474#issuecomment-2636966914

PR #9505, opened 2025-02-04
fix module source cache invalidation + cleanup api

sipsma:
# 🔥 Breaking Change
  This improvement comes with a few breaking changes that are not expected to impact most users. Impacted users are those who:
  - have function code that imports local libraries outside of the dagger module source directory
  - rely on a package manager configuration file (e.g. go.mod) that exists in a parent directory of the dagger module source directory

  They may need to update their dagger.json configuration file to add an explicit `"include"` setting to ensure those extra files or directories are loaded with the module. This is because numerous files previously implicitly loaded are now skipped by default, which contributes to the performance improvements.

  For example, if your module code is in `.dagger` and relies on the `go.mod` file in the parent directory, you would add the following to your `dagger.json` file:

  ```json
  {
    "include": ["../go.mod", "../go.sum"]
  }
  ```

  The values in the `"include"` array are relative to the directory containing `dagger.json`. They can be also be glob patterns (e.g. `"**/*.txt"`). To explicitly exclude some files matched by a previous include pattern, you can prepend the include pattern with a `!` character (e.g. `"!**/foo.txt"`).
  - Previously, the `exclude` setting existed but has now been superseded by prefixing an include pattern with a `!` character. However, this is backwards-compatible, so existing `exclude` settings will be automatically turned into equivalent `!`-prefixed include settings when running `dagger develop`. No user intervention is needed.

  The core API's `ModuleSource` and `Module` types have also undergone some cleanup that includes breaking changes. Most users do not interact with these APIs, but any that do may need some minor adjustments to their code. In particular:
  - `Module.initialize` no longer needs to be called; the module is initialized on creation
  - `ResolveFromCaller` no longer needs to be called for local module sources; context is automatically loaded and resolved efficiently as needed by other API calls

Finally, an undocumented feature called `views` has been removed as it has been superseded by [default path arguments](https://docs.dagger.io/api/default-paths/#for-all-other-cases).

---

Goals here are:
* Fix various sources of excessive module building cache invalidation
   * changing completely unrelated source files in context
   * building different git commits that have no changes in the module source
   * etc.
* Use the [filesync improvements](https://github.com/dagger/dagger/pull/8818) + [new caching features](https://github.com/dagger/dagger/pull/9204) to cleanup tons of cruft from the API (ResolveFromCaller, Module.initialize, etc.)
* Setup for more performance improvements like better module remote caching, general function call caching, etc.
   * This step is a pre-req in that it gives us precise control over the cache keys of module source and modules themselves, which is needed first in order to have any more improvements to their cacheability

So far the results are looking good.

  For some concrete numbers, here are `load module` times for the `dagger-dev`'s module in Dagger's repository under different scenarios:
  - `dagger call --help` on new engines with an empty cache
    - v0.15.4: 1m20s
    - v0.16.0: 1m1s
    - ~23% faster
  - re-running `dagger call --help` with no file changes in the repo:
    - v0.15.4: 10.9s
    - v0.16.0: 2.8s
    - ~74% faster
  - re-running `dagger call --help` after making a change to a random source code file in the repo:
    - v0.15.4: 32.1s
    - v0.16.0: 2.8s
    - ~91% faster

The difference in `2.` above of 10x speedup should make our day to day development quite a bit less painful; *a lot* less waiting around an extra 30s every time any change is made.

In terms of the simplifications, the net deletion of over 4k lines of code is indicative of success so far there 😄 

--- 

TODOs:
- [x] get all integ tests passing, anticipating a lot of corner case squashing from stuff missed during refactoring
- [x] dedupe a bunch of copy-pasta chunks of code
- [x] cleanup various TODO comments (fix dagger install telemetry, etc.)
- [x] fill in missing api doc strings
- [x] appease linter
- [x] write up breaking change 
- [x] remove more old cruft like RequiredPaths from sdks (now unused but still laying around)
- [x] update daggerverse code, coordinate release of it
   - PR out + tested in other repo


cmd/dagger/module.go
jedevc, line 1: With all these cli <-> engine changes, I wonder if we should go ahead and bump here? https://github.com/dagger/dagger/blob/a99139ba894f240808fa6c1c9fd995ffba58cf41/engine/version.go#L27-L29

IMO, no real reason to keep thinking about clients that old.

core/schema/modulesource.go
```diff
@@ -1,1648 +1,2031 @@
...
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 type moduleSourceArgs struct {
 	// avoiding name "ref" due to that being a reserved word in some SDKs (e.g. Rust)
 	RefString string
 	RefPin    string `default:""`
+	Stable    bool   `default:"false"`
```
jedevc, line 38: I *think* while we're here we can remove this. https://github.com/dagger/dagger/pull/8587 added pins, and the logic of pins essentially means that everything is pretty much stable now.

core/modulesource.go
```diff
@@ -425,284 +279,82 @@ func (src *ModuleSource) LoadContext(ctx context.Context, dag *dagql.Server, pat
...
-			return inst, fmt.Errorf("local src not set")
-		}
-		if !src.AsLocalSource.Value.ContextDirectory.Valid {
-			return inst, fmt.Errorf("local src context directory not set")
-		}
-		return src.AsLocalSource.Value.ContextDirectory.Value, nil
-	case ModuleSourceKindGit:
-		if !src.AsGitSource.Valid {
-			return inst, fmt.Errorf("git src not set")
+	case ModuleSourceKindDir:
```
jedevc, line 282: Yes!! :tada: this cleans up a very nice long-standing itch for me :tada:

core/schema/module.go
```diff
@@ -95,140 +92,84 @@ func (s *moduleSchema) Install() {
...
 
-		dagql.NodeFunc("dependencies", s.moduleSourceDependencies).
-			Doc(`The effective module source dependencies from the configuration, and calls to withDependencies and withoutDependencies.`),
+		dagql.Func("withSDK", s.moduleSourceWithSDK).
+			Doc(`Update the module source with a new SDK.`).
+			ArgDoc("sdk", `The SDK to set.`),
+
+		dagql.Func("withInit", s.moduleSourceWithInit).
+			Doc(`Sets module init arguments`).
+			ArgDoc("merge", `Merge module dependencies into the current project's`),
```
jedevc, line 112: Hm, another while we're here... @marcosnils any objection to removing this now? I remember we changed the default and left this in here as a way to get the old behavior.

But we never documented it, and I don't know that anyone relies on it? Since we're doing breaking things anyways here, I figure it probably wouldn't hurt to just remove it and simplify the logic changed in https://github.com/dagger/dagger/pull/8095.

core/schema/modulesource.go
```diff
@@ -1,1648 +1,2031 @@
...
+				// NOTE: important that we use stat.Path here rather than curDirPath since the stat also
+				// does some normalization of paths when the client is using case-insensitive filesystems
+				found[soughtName] = filepath.Dir(stat.Path)
+				continue
+			}
+			// TODO: remove the strings.Contains check here (which aren't cross-platform),
+			// since we now set NotFound (since v0.11.2)
+			if status.Code(err) != codes.NotFound && !strings.Contains(err.Error(), "no such file or directory") {
+				return nil, fmt.Errorf("failed to lstat %s: %w", soughtName, err)
+			}
```
jedevc, line 391: If we bump the client <-> engine versions, we can remove this extra sadness as well :tada:

core/schema/module.go
sipsma: I removed it from the CLI. There's still a need for a flag on `codegen` because I think the go sdk (non module) build actually requires the behavior, but it's 100% internal now.

core/modulesource.go
sipsma: Yeah pretending those were local was the epitome of hack :guyfierichefkiss:, glad it's gone now too

cmd/dagger/module.go
sipsma: Yeah I agree. It is a little painful but we'd need to jump through like 10 hoops to make this backwards compatible (not 100% sure if it's even possible really), so might as well bump this time around. Pushed that change

sipsma: Wait, I started getting `start session: connect buildkit session: unexpected status 500: incompatible client version v0.15.3` in testdev jobs after this.... 🤔

Don't have time this exact moment, will look again tomorrow but gonna revert this temporarily so the tests get a clean run again.

cmd/dagger/module_inspect.go
```diff
@@ -97,28 +97,31 @@ func tryInitializeModule(ctx context.Context, dag *dagger.Client, srcRef string)
 	defer telemetry.End(span, func() error { return rerr })
 
 	findCtx, findSpan := Tracer().Start(ctx, "finding module configuration", telemetry.Encapsulate())
-	conf, _ := getModuleConfigurationForSourceRef(findCtx, dag, srcRef, true, true)
+	modSrc := dag.ModuleSource(srcRef, dagger.ModuleSourceOpts{
+		DisableFindUp: true,
+	})
```
helderco, line 102: The previous call to `getModuleConfigurationForSourceRef` had "find up" turned on, but this disables it.

helderco:
Disabling find up isn't working for local sources:
```
⋈ .core | module-source sdk/cue --disable-find-up | as-string
sdk/cue
⋈ .core | module-source github.com/dagger/dagger/sdk/cue --disable-find-up | as-string
Error: input: moduleSource git module source "github.com/dagger/dagger/sdk/cue" does not contain a dagger config file
```

helderco:
Oh, git sources never do findUp:
```
⋈ .core | module-source github.com/dagger/dagger/sdk/cue | as-string
Error: input: moduleSource git module source "github.com/dagger/dagger/sdk/cue" does not contain a dagger config file
```
Can we make it do it?

helderco:
I expected `sdk/php` here:
```
⋈ .core | module-source sdk/php/src | as-string
sdk/php/src
```

cmd/dagger/module_inspect.go
sipsma: I added this while fixing some failing tests, but I dumbly did not add a comment here on *why* it seemed necessary... I will try rm'ing it and see what, if any, tests fail and figure it out from there. I agree it probably makes more logical sense to enable it here for consistency

sipsma: All the tests passed w/ it enabled so yeah I guess this wasn't necessary in the end, re-enabled it

sipsma:
> Oh, git sources never do findUp:

Yeah for sure, added that

> I expected sdk/php here:

Changed this a bit, let me know what you think. I changed it so that the return value for local sources is the full absolute path to the source root. That makes sense in that it's consistent for a given host and not tied to the caller's cwd

However, this did break something in shell help output in that what might have previously been e.g. `Usage: .` now be `Usage: /the/full/abs/path`. So I just did some special case handling in the shell client specifically to handle that and make it a relative path for nicer looking output there (back to `.` in the example before): https://github.com/dagger/dagger/pull/9505/commits/78cc5590ddcfa10e630fc4224146a6f237441dbf

helderco:
> > Oh, git sources never do findUp:
> 
> Yeah for sure, added that

Nice, that makes things simpler. 👍

> > I expected sdk/php here:
> 
> Changed this a bit, let me know what you think. I changed it so that the return value for local sources is the full absolute path to the source root. That makes sense in that it's consistent for a given host and not tied to the caller's cwd

Yeah, that's what I need, both local and git return a full ref. It's easy from the CLI to make it relative to either the context or `os.Getwd()` if I need to.

> However, this did break something in shell help output in that what might have previously been e.g. `Usage: .` now be `Usage: /the/full/abs/path`. So I just did some special case handling in the shell client specifically to handle that and make it a relative path for nicer looking output there (back to `.` in the example before): [78cc559](https://github.com/dagger/dagger/commit/78cc5590ddcfa10e630fc4224146a6f237441dbf)

Yeah, that's going to change anyway because of:
- https://github.com/dagger/dagger/pull/9511

Basically that path will be handled like in other places, which will be relative to where you've navigated to inside the context directory, rather than relative to `os.Getwd()`.

But thanks for the fix so we're not blocked on that PR.

helderco:
@sipsma, I know I said I didn't need `AllowNotExists` on git sources, but would make it more consistent with local and not hard to do, right? Not a blocker, just a suggestion.

cmd/dagger/module_inspect.go
```diff
@@ -57,24 +59,24 @@ func initializeModule(
...
-	doFindUp bool,
 	srcOpts ...dagger.ModuleSourceOpts,
 ) (rdef *moduleDef, rerr error) {
 	ctx, span := Tracer().Start(ctx, "load module")
 	defer telemetry.End(span, func() error { return rerr })
 
 	findCtx, findSpan := Tracer().Start(ctx, "finding module configuration", telemetry.Encapsulate())
-	conf, err := getModuleConfigurationForSourceRef(findCtx, dag, srcRef, doFindUp, true, srcOpts...)
+	modSrc := dag.ModuleSource(srcRef, srcOpts...)
+	configExists, err := modSrc.ConfigExists(findCtx)
```
helderco, line 69: Doesn't `ConfigExists` only make sense when you have `AllowNotExists` now? So the `!configExists` below won't ever be hit because if `srcRef` isn't pointing to a module then this will return an error instead.

I think we can rename `initializeModuleConfig` to `initializeModule` and replace the other one with:
```diff
- initializeModule(ctx, dag, modRef, opts)
+ initializeModule(ctx, dag, dag.ModuleSource(modRef, opts))
```

cmd/dagger/module_inspect.go
```diff
@@ -97,28 +99,29 @@ func tryInitializeModule(ctx context.Context, dag *dagger.Client, srcRef string)
 	defer telemetry.End(span, func() error { return rerr })
 
 	findCtx, findSpan := Tracer().Start(ctx, "finding module configuration", telemetry.Encapsulate())
-	conf, _ := getModuleConfigurationForSourceRef(findCtx, dag, srcRef, true, true)
+	modSrc := dag.ModuleSource(srcRef)
```
helderco, line 102: We need `AllowNotExists` here:
```suggestion
	modSrc := dag.ModuleSource(srcRef, dagger.ModuleSourceOpts{AllowNotExists: true})
```

cmd/dagger/shell_exec.go
```diff
@@ -678,11 +678,13 @@ func (h *shellCallHandler) GetDependency(ctx context.Context, name string) (*She
...
-		var opts []dagger.ModuleSourceOpts
+		opt := dagger.ModuleSourceOpts{
+			DisableFindUp: true,
+		}
 		if dep.RefPin != "" {
-			opts = append(opts, dagger.ModuleSourceOpts{RefPin: dep.RefPin})
+			opt.RefPin = dep.RefPin
 		}
-		return initializeModule(ctx, h.dag, dep.ModRef, false, opts...)
+		return initializeModule(ctx, h.dag, dep.ModRef, opt)
```
helderco, line 687: It was silly of me to add the `if dep.RefPin != ""` condition 😅 because empty value options are stripped from the request anyway. Assuming the rename of `initializeModuleConfig` to `initializeModule`, all of this becomes:
```go
	return initializeModule(ctx, h.dag, h.dag.ModuleSource(dep.ModRef, dagger.ModuleSourceOpts{
		DisableFindUp: true,
		RefPin: dep.RefPin,
	})
```
Or add a method to `moduleDependency`:
```go
	return initializeModule(ctx, h.dag, dep.Source())
```

core/schema/modulesource.go
```diff
@@ -1,790 +1,998 @@
...
+	// and localPath is a named dependency in that dagger.json, the returned source will be for that dependency.
+	doFindUp bool,
 
-		subPath := "/"
-		if parsed.repoRootSubdir != "" {
-			subPath = parsed.repoRootSubdir
+	// if true, tolerate the localPath not existing on the filesystem (for dagger init on directories that don't exist yet)
+	allowNotExists bool,
+) (inst dagql.Instance[*core.ModuleSource], err error) {
+	if doFindUp {
```
helderco, line 102: Should check if `localPath` exists first before doing `findUp` as a fallback:
```console
❯ dagger init --sdk=go graphql
Initialized module graphql in graphql
❯ dagger-dev shell --no-mod -c '.core | module-source graphql | as-string'
/Users/helder/Projects/dagger/modules/graphql
```
Expected: `/Users/helder/Projects/dagger/graphql`

cmd/dagger/module_inspect.go
sipsma: Why would we want AllowNotExists? The intention for allow not exists was really just to support `dagger init` when it's possible the source root dir doesn't exist yet locally. When would `dagger shell` or `dagger call` want to load a module that doesn't exist?

There's complication with enabling it because it in combination with find-up means you can specify any subdir that doesn't exist and just "find-up" to a module that does exist. So like `dagger -m this/is/not/real ...` in the dagger repo results in the `dagger-dev` module being loaded

cmd/dagger/shell_exec.go
sipsma: Simplified this

sipsma:
@helderco 

> I still need for a way to get the original subpath the user used. What do you suggest?

We can just add another api for that (the original user specified path before findup, whether it was local or git).

Are you okay with adding that in your PR that's based on this one instead of including here? The rebasing and churn in this one is getting a but brutal and seems like it would be easier to tackle in yours where it's actually going to be used.

helderco:
Yeah, already did that yesterday. It's on the PR.

cmd/dagger/module_inspect.go
helderco: > Why would we want AllowNotExists? The intention for allow not exists was really just to support `dagger init` when it's possible the source root dir doesn't exist yet locally. 

I get that, but there’s two other use cases that I can think of. 

> When would `dagger shell` or `dagger call` want to load a module that doesn't exist?

It's not that the shell needs to load a module that doesn't exist, it just needs to not fail if the given ref isn't a module. The reason for that is that the shell supports looking into multiple places when you call a function.

For example:
```
⋈ alpine | .doc
⋈ github.com/dagger/dagger/modules/alpine | .doc
```
1. Is it a function in currently loaded module?
2. Is it a dependency of the currently loaded module?
3. Is it a stdlib function?
4. Is it a reference to a module? (this is where I use `tryInitializeModule`)
5. If none of the above, return simple error

This is the same kind of behavior on `dagger query` and `dagger listen`, which load a module optionally if it exists in the current dir. `dagger call` doesn’t need this because it has to be started with a valid module. `dagger shell` doesn’t.

The motivation was mainly to depend on `ConfigExists` rather than ignoring an error:
```
⋈ .core | module-source github.com/dagger/dagger-go-sdk/telemetry | config-exists
Error: input: moduleSource git module source "github.com/dagger/dagger-go-sdk/telemetry" does not contain a dagger config file
```

Because ignoring the error could hide other things that went wrong, not necessarily that there's no module. Besides, it prevents a failing span to show up in telemetry.

But things are a bit different now, and since the _trying_ to load a module is the last step in that fallback strategy, I can just let it return what `ModuleSource` returns, failure or not. But I still need to ignore the error when the shell is started, just like in `optionalModCmdWrapper`.

That being said, even if we end up wanting to make check without an error later on, it can be triggered by a separate option in `ModuleSource` or a separate field altogether, which don't break backwards compat since it would be an addition.

cmd/dagger/module_inspect.go
helderco: I meant to use replace the `modRef` and `opts` arguments with a single `ModuleSource` argument (see first comment).

Then I also meant to keep `tryInitializeModule` as a convenience for adding the `AllowNotExists` option and using `ConfigExists`, but maybe now it's not necessary (see https://github.com/dagger/dagger/pull/9505#discussion_r1959634827).

cmd/dagger/shell_exec.go
helderco: Missing the `RefPin`, but I pushed it to your branch.

core/module.go
```diff
@@ -99,110 +67,46 @@ func (*Module) TypeDescription() string {
...
+	default:
+		panic(fmt.Sprintf("unexpected module source kind %q", mod.Source.Self.Kind))
 	}
-	newMod.InstanceID = newID
 
-	return newMod, nil
+	return call.NewModule(mod.InstanceID, mod.Name(), ref, pin)
+}
+
+func (mod *Module) Evaluate(context.Context) (*buildkit.Result, error) {
```
jedevc, line 108: I might be missing something, but where is `Sync` on a `Module`/`ModuleSource` used?

Edit: I see it used in https://github.com/dagger/dagger.io/pull/4242, might be worth making a note that it's used in daggerverse so we don't accidentally remove it.

(hopefully most of those cases are now gone, but that's one of things that used to make this hard to work with, since you had to check daggerverse to see if a field was being used or not)

docs/static/api/reference/index.html
jedevc, line 1: We recently merged https://github.com/dagger/dagger/pull/9592, so think that the PHP docs will also need regen (should be part of the top-level generate).

cmd/dagger/module.go
```diff
@@ -144,71 +143,98 @@ If --sdk is specified, the given SDK is installed in the module. You can do this
...
+				// create them when exporting the generated context directory.
+				AllowNotExists: true,
+			})
+
+			kind, err := modSrc.Kind(ctx)
 			if err != nil {
-				return fmt.Errorf("failed to get configured module: %w", err)
+				return fmt.Errorf("failed to get module ref kind: %w", err)
+			}
+			if kind != dagger.ModuleSourceKindLocalSource {
```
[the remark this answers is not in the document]
sipsma: Good catch, `requireKind` was being used in daggerverse but agree it makes perfect sense here too

core/schema/module.go
```diff
@@ -34,17 +26,13 @@ func (s *moduleSchema) Install() {
 		dagql.Func("generatedCode", s.generatedCode).
 			Doc(`Create a code generation result, given a directory containing the generated code.`),
 
-		dagql.Func("moduleSource", s.moduleSource).
```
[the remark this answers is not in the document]
sipsma: Yeah it's used by SDKs, when we call a module to ask what types/functions it has it does e.g. `dag.Module().WithDescription(...).WithObject(...)` and then returns that `Module`

I don't really love that approach anymore, sort of vestigial from the first iteration but it's not really hurting anything and don't want to expand the scope of this change more so will leave as is for now

core/schema/module.go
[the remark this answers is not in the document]
sipsma: Yeah there's no good reason for that, just happened at some point and cargo-culted since. I'll split them up

core/modfunc.go
```diff
@@ -92,6 +92,9 @@ type CallOpts struct {
 	Cache          bool
 	SkipSelfSchema bool
 	Server         *dagql.Server
+
+	// TODO: doc
+	SkipCallDigestCacheKey bool
```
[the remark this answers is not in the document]
sipsma: Oops, will fix!

core/module.go
sipsma: Sure yeah I'll add a comment saying they are important

Issue #8671, opened 2024-10-10
Improved custom enum support


jedevc:
Opening an issue from https://discord.com/channels/707636530424053791/1123687185837740053/1290092449183698965 - me and @helderco had a chat about this.

Current enum support has a few issues, which we can fix!

- Naming clashes are very possible in go
	- Enum values aren't namespaced at all.
	- Fixed by https://github.com/dagger/dagger/pull/8669

- Some core enum types have stuttering names
	- e.g. `TypeDefKind.STRING_KIND` should be `TypeDefKind.STRING`, `ModuleSourceKind.GIT_SOURCE` should be `ModuleSourceKind.GIT`

- We confuse enum names and values, e.g. in the following go declaration: `const MyEnumValue MyEnum = "foobar"`, the graphql `enumValues.name` is populated as `foobar` - this is weird, because *everything* else is parsed by identifier (functions, args, structs, etc).
	- This is caused by the fact that graphql doesn't have separate names/values for enums, so we picked values.
	- However, with #8498 incoming, we can add custom directives :smile: 
	- So, we can switch the `enumValues` `name` to be derived from the name - and use a custom `@value` directive for the value.
	- This does a few things:
		- The name generation is now internally consistent
		- Names are correctly shared between different modules
		- Values don't need to conform to the graphql naming format (which only allows *very* specific formats)
		- Enum values could be of other types, not just strings!

sagikazarmark:
Enums in Go is a giant PITA. They aren't just badly supported by the language, there are also competing practices out there with varying UX.

Here is how I define enums these days in Dagger modules:

```
// module: trivy
type ReportFormat string

const (
	ReportFormatTable      ReportFormat = "table"
	ReportFormatJSON       ReportFormat = "json"
	ReportFormatTemplate   ReportFormat = "template"
	ReportFormatSARIF      ReportFormat = "sarif"
	ReportFormatCycloneDX  ReportFormat = "cyclonedx"
	ReportFormatSPDX       ReportFormat = "spdx"
	ReportFormatSPDXJSON   ReportFormat = "spdx_json"
	ReportFormatGitHub     ReportFormat = "github"
	ReportFormatCosignVuln ReportFormat = "cosign_vuln"
)
```

I would expect to be able to use a _similar_ version of this on the other side: probably `dagger.TrivyReportFormatMarkdown`?

I chose to prefix the enum values manually because there might be an overlap between different kinds of enums.

jedevc:
[...]

PR #9518, opened 2025-02-06
feat: separate out enum names and values

jedevc:
Fixes https://github.com/dagger/dagger/issues/8671 - the aim is to fix all the remaining little nits with enums that make them tricky to use (also wanting to unblock #8608).

TODO:
- [x] Tidy up core enum names
- [x] Improved `withEnumMember` API for enum members
- [ ] Kick off migration from `withEnumValue` to `withEnumMember` in all SDKs
	- [x] Go
	- [x] Python
	- [ ] Typescript
- [ ] Backwards compat (this is gonna be bit tricky)
	- [ ] Handle renamed types, `ImageMediaType`->`ImageMediaTypes`, `EnumValueTypeDef`->`EnumMemberTypeDef`
	- [ ] Ensure that codegen for enums in dependencies stays the same
	- [x] Keep old TypeDef field `EnumTypeDef.values`
	- [x] Keep old core enum values (using member aliases)
	- [ ] Add new legacy tests
- [ ] Fix inevitably broken tests
- [ ] Clean up

### DIFF
--- a/core/enum.go
+++ b/core/enum.go
@@ -156,6 +156,10 @@ func (e *ModuleEnum) Decoder() dagql.InputDecoder {
 }
 
 func (e *ModuleEnum) DecodeInput(val any) (dagql.Input, error) {
+	if enum, ok := val.(*ModuleEnum); ok {
+		val = enum.Name
+	}
+
 	v, err := (&dagql.EnumValueName{Enum: e.TypeName()}).DecodeInput(val)
 	if err != nil {
 		return nil, err
--- a/core/enum.go
+++ b/core/enum.go
@@ -40,17 +40,25 @@ func (m *ModuleEnumType) ConvertFromSDKResult(ctx context.Context, value any) (d
 
 	switch value := value.(type) {
 	case string:
-		decoder, err := m.getDecoder(ctx)
+		enum, local, err := m.getEnum(ctx)
 		if err != nil {
-			return nil, fmt.Errorf("%T.ConvertFromSDKResult: failed to get decoder: %w", m, err)
+			return nil, fmt.Errorf("%T.ConvertFromSDKResult: failed to get enum: %w", m, err)
 		}
-
-		val, err := decoder.DecodeInput(value)
-		if err != nil {
-			return nil, fmt.Errorf("%T.ConvertFromSDKResult: invalid enum value %q for %q: %w", m, value, m.typeDef.Name, err)
+		// XXX:
+		// targetName := ...
+		for _, member := range enum.TypeDef.Members {
+			if local && member.OriginalName == value {
+				return enum.DecodeInput(member.Name)
+			} else if !local && member.Name == value {
+				return enum.DecodeInput(member.Name)
+			}
 		}
-
-		return val, nil
+		for _, member := range enum.TypeDef.Members {
+			if member.Value == value {
+				return enum.DecodeInput(member.Name)
+			}
+		}
+		return nil, fmt.Errorf("%T.ConvertFromSDKResult: invalid enum value %q for %q", m, value, m.typeDef.Name)
 	default:
 		return nil, fmt.Errorf("unexpected result value type %T for enum %q", value, m.typeDef.Name)
 	}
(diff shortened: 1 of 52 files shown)


jedevc:
I split some work out to #10103, so will rebase on that once merged.

wingyplus:
Will looking for Elixir within today✌️

sdk/elixir/lib/dagger/gen/type_def_kind.ex
```diff
@@ -95,13 +106,89 @@ defmodule Dagger.TypeDefKind do
   @spec enum_kind() :: :ENUM_KIND
   def enum_kind(), do: :ENUM_KIND
 
+  @doc """
+  A string value.
+  """
+  @spec string() :: :STRING
+  def string(), do: :STRING
```
wingyplus, line 113: The function name is correct, but the value should be `STRING_KIND` from the `enumValue` directive, right?

jedevc: So :stuck_out_tongue: 

It's a bit tricky. Ideally, yes, this value should be `STRING_KIND`. However! It's not required for it all to be the same, the engine *will* allow decoding both.

To get `STRING_KIND` here, yes, you need to use the `emumValue` directive introspection.

core/enum.go
```diff
@@ -95,7 +112,9 @@ func (m *ModuleEnumType) getDecoder(ctx context.Context) (dagql.InputDecoder, er
 
 type ModuleEnum struct {
 	TypeDef *EnumTypeDef
-	Value   string
+	Name    string
+
+	Local bool
```
helderco, line 121: Can you add a descriptive comment for what `Local` is?

core/integration/legacy_test.go
```diff
@@ -1021,3 +1022,229 @@ func (m *Bare) dir() *dagger.Directory {
...
+class Test:
+    status: Status = dagger.field(default=Status.ACTIVE)
+
+    @dagger.function
+    def from_status(self, status: Status) -> str:
+        return str(status)
+
+    @dagger.function
+    def to_status(self, status: str) -> Status:
+        # Doing "Status(proto)" will fail in Python, so mock
```
helderco:
```suggestion
        # Doing "Status(status)" will fail in Python, so mock
```

core/schema/module.go
```diff
@@ -358,7 +373,36 @@ func (s *moduleSchema) typeDefWithEnumValue(ctx context.Context, def *core.TypeD
...
+
+func (s *moduleSchema) typeDefWithEnumMember(ctx context.Context, def *core.TypeDef, args struct {
+	Name        string
+	Value       string
+	Description string `default:""`
+	SourceMap   dagql.Optional[core.SourceMapID]
+}) (*core.TypeDef, error) {
+	if args.Value == "" {
+		return nil, fmt.Errorf("enum value must not be empty")
+	}
```
helderco: I think ideally the value and name should be the same, we're just giving flexibility for users to change that, so I'd make `Value` optional:
```suggestion
	if args.Value == "" {
		args.Value = args.Name
	}
```
In that case the directive for the value in the GraphQL schema is only really needed if they're different.

core/typedef.go
```diff
@@ -1004,32 +1040,49 @@ func (enum EnumTypeDef) Clone() *EnumTypeDef {
 	return &cp
 }
 
-type EnumValueTypeDef struct {
+type EnumMemberTypeDef struct {
 	Name        string     `field:"true" doc:"The name of the enum value."`
+	Value       string     `field:"true" doc:"The value of the enum value"`
```
helderco: Maybe rename "value" to "member" in all the descriptions?
```suggestion
	Value       string     `field:"true" doc:"The value of the enum member"`
```

core/typedef.go
```diff
@@ -1004,32 +1040,49 @@ func (enum EnumTypeDef) Clone() *EnumTypeDef {
...
 
-func NewEnumValueTypeDef(name, description string, sourceMap *SourceMap) *EnumValueTypeDef {
-	return &EnumValueTypeDef{
-		Name:        name,
-		Description: description,
-		SourceMap:   sourceMap,
+func NewEnumMemberTypeDef(name, value, description string, sourceMap *SourceMap) *EnumMemberTypeDef {
+	return &EnumMemberTypeDef{
+		OriginalName: name,
+		Name:         strcase.ToScreamingSnake(name),
```
helderco, line 1068: 👍

core/schema/module.go
```diff
@@ -206,11 +206,22 @@ func (s *moduleSchema) Install() {
 
 		dagql.Func("withEnumValue", s.typeDefWithEnumValue).
 			Doc(`Adds a static value for an Enum TypeDef, failing if the type is not an enum.`).
+			Deprecated("Use withEnumMember instead").
```
helderco: Wrapping in backticks makes SDKs convert the casing appropriately:
```suggestion
			Deprecated("Use `withEnumMember` instead").
```

sdk/python/src/dagger/mod/_converter.py
```diff
@@ -46,12 +48,26 @@ def make_converter():
...
+        return cls[name]
+
+    conv.register_unstructure_hook(enum.Enum, to_enum_name)
+    conv.register_structure_hook(enum.Enum, from_enum_name)
+
+    conv.register_unstructure_hook(base.Enum, to_enum_name)
+    conv.register_structure_hook(base.Enum, from_enum_name)
+
+    conv.register_unstructure_hook(Enum, to_enum_name)
+    conv.register_structure_hook(Enum, from_enum_name)
```
helderco: You can avoid duplication here by using your own predicate function:
```suggestion
def is_enum(t) -> bool:
    return issubclass(t, enum.Enum)

conv.register_unstructure_hook_func(is_enum, to_enum_name)
conv.register_structure_hook_func(is_enum, from_enum_name)
```

sdk/python/src/dagger/mod/_arguments.py
```diff
@@ -148,3 +152,26 @@ def _validate(self):
...
+    conv.register_unstructure_hook(enum.Enum, to_enum_name)
+    conv.register_structure_hook(enum.Enum, from_enum_name)
+
+    conv.register_unstructure_hook(base.Enum, to_enum_name)
+    conv.register_structure_hook(base.Enum, from_enum_name)
+
+    conv.register_unstructure_hook(Enum, to_enum_name)
+    conv.register_structure_hook(Enum, from_enum_name)
+
+    return conv
```
helderco: You can reuse the converter in `_converter.py` to avoid the duplication.

sdk/python/src/dagger/client/_core.py
```diff
@@ -293,4 +295,16 @@ def _struct(d: dict[str, Any], cls: type) -> Any:
...
+
+    def from_enum_name(name: str, cls: type[enum.Enum]) -> enum.Enum:
+        return cls[name]
+
+    conv.register_unstructure_hook(enum.Enum, to_enum_name)
+    conv.register_structure_hook(enum.Enum, from_enum_name)
+
+    conv.register_unstructure_hook(base.Enum, to_enum_name)
+    conv.register_structure_hook(base.Enum, from_enum_name)
+
```
helderco: Same as in `_converter.py`. Maybe best to extract this into its own function that takes a converter as an argument, to make this registration.

helderco (commented):
This took effort, thanks for sticking with it! 😄 Looking great!

core/schema/module.go
jedevc: Made the value optional! And then only inject the directive when it's different.

sdk/python/src/dagger/mod/_converter.py
jedevc: :thinking: convinced I must have done something wrong, when I do this I get

```
TypeError: Invalid first argument to `register()`. <function is_enum_type at 0x7fa5e8b585e0> is not a class or union type            
! failed to call module "python-sdk-dev" to get functions: process "/runtime" did not complete successfully: exit code: 1
```

jedevc: Hm. So this *doesn't* work. https://github.com/dagger/dagger/commit/a2fd403e57a0f85f094b7d97d186fe89446dc6de

https://dagger.cloud/jedevc/traces/26e7a81edf9304da7090245ef15893fe

I could be missing something obvious.

But I seem to remember having this conversation in discord here: https://discord.com/channels/707636530424053791/1337048414751559680/1341125495189798962

helderco: I refactored a bit to allow for it, pushed to your branch.

Basically the `dagger.client.base.Enum` class extended `str` making members of the class basically strings. The problem is that cattrs would dispatch the hook for "str" when encountering a member, instead of the custom added. We could override the hook for "str" too but I want to simplify.

I want to deprecate `dagger.Enum` in favor of `enum.Enum`:
- https://github.com/dagger/dagger/pull/10600

Still added a type hook (instead of function hook) if an enum decorated with `@dagger.enum_type` extends a primitive type like `str` or `int` for backwards compat or if users end up using it on their own directly.

So this should cover all enums in a simple way without having to register 3 different types and still missing cases (e.g., user extending `str` and  `enum.Enum` without one of our base classes).

helderco (approved):
🎉

PR #9519, opened 2025-02-06
docs: Delete PHP integrations page

vikram-dagger:


jpadams:
Redirects here: https://docs.dagger.io/api/sdk/

Makes sense to lose this page where we use the other SDKs to show how to handle a PHP project, since we'll have a PHP SDK...unless we keep this page, add PHP to the languages (making PHP the default one)

jpadams (approved):
👍 I'm okay not to block this.

jasonmccallister:
@vikram-dagger @jpadams should it maybe redirect to the page in https://github.com/dagger/dagger/pull/9592?

PR #9520, opened 2025-02-06
sdk/java: implement fields

eunomie:
Allow to define fields from a class or record and register them.

Fields **must be public** to be exposed.


sdk/java/dagger-java-annotation-processor/src/test/resources/io/dagger/gen/entrypoint/entrypoint.java
```diff
@@ -121,7 +121,9 @@ private ModuleID register() throws ExecutionException, DaggerQueryException,
                     dag.function("nullableDefault",
                         dag.typeDef().withKind(TypeDefKind.STRING_KIND))
                         .withDescription("Set a default value in case the user doesn't provide a value and allow for null value.")
-                        .withArg("stringArg", dag.typeDef().withKind(TypeDefKind.STRING_KIND).withOptional(true), new Function.WithArgArguments().withDefaultValue(JSON.from("\"Foo\"")))));
+                        .withArg("stringArg", dag.typeDef().withKind(TypeDefKind.STRING_KIND).withOptional(true), new Function.WithArgArguments().withDefaultValue(JSON.from("\"Foo\""))))
```
helderco, line 124: I didn't notice this before, but isn't this coming from `containerEcho(@Default("Hello Dagger") String stringArg)`? So there shouldn't be a `withOptional(true)`.

sdk/java/dagger-java-annotation-processor/src/test/resources/io/dagger/java/module/DaggerJava.java
```diff
@@ -14,6 +14,12 @@
 /** Dagger Java Module main object */
 @Object
 public class DaggerJava extends AbstractModule {
+  /** Project source directory */
+  private Directory source;
+
+  /** Go version */
+  private String version;
```
helderco, line 21: I think if it's private it shouldn't be exposed to Dagger.

See https://docs.dagger.io/api/state

If you compare with Go, notice that sometimes `// +private` is used. That's because of the serialization/deserialization, which only works on public fields of a struct so that pragma is needed when a user needs the serialization but not to expose it to Dagger. Otherwise, just making it private (i.e., camelCase instead of PascalCase) is enough to not expose it.

helderco (commented):
Not sure if you're familiar with how fields work so let me clarify.

Fields are actually special in the engine. They're state that the engine returns directly. So if you expose a property like `version`, you'll see a new function show up in `dagger functions` (dagger creates a getter for it in the API), but if you `dagger call` it, it won't actually execute the user's module to get its value. The engine will return the value it has for it immediately.

So even though you're only registering in this PR it should actually be usable. **Even without constructor support**, you can set these fields with `withVersion()` kind of functions that modify the state. You just need to make sure they are serialized when returning the main object (or any other object with state).

You're almost there then, so I suggest you don't leave these for later:
- Figure out visibility of fields
- Make sure they're serialized when the object is returned
- Make sure they're deserialized when you get the `parentJson`

I believe the last two are already handled by `JsonConverter.fromJSON` and `JsonConverter.toJSON`, just not sure if the `private` factors into it like in Go.

Basically, add a test case for chained function calls that modify the state from one function and uses that state from another one, like this:
```console
> dagger shell -c 'my-module | with-version 1.23 | check'
You have version v1.23.
```

sdk/java/dagger-java-annotation-processor/src/test/resources/io/dagger/gen/entrypoint/entrypoint.java
eunomie: This one is coming from `public String nullableDefault(@Nullable @Default("Foo") String stringArg)`

The one you mentioned is for instance `public String nonNullableDefault(@Default("default value") String stringArg)` that generates `.withArg("stringArg", dag.typeDef().withKind(TypeDefKind.STRING_KIND), new Function.WithArgArguments().withDefaultValue(JSON.from("\"default value\""))))` with the default value but no optional.

eunomie:
Thanks @helderco for the details.
Yeah I figured out that while building the constructor side (basically the serialization wasn't working due to private fields).

I'll make them public for now. I don't know if that's really the way people are doing Java nowadays, but at least it will be consistent with the other SDKs and more than that it will work :-)

eunomie:
```console
> dagger shell -c 'my-module | with-version 1.23 | check'
You have version v1.23.
```

I've tried but I'm not able to make it run.
I also tried with a very simple Go module without more success.

```go
type Goc struct {
	Version string
}

func (m *Goc) EchoVersion() string {
	return m.Version
}
```

Then

```console
> dagger shell -c 'goc | with-version 1.2.3 | echo-version'
✔ connect 0.2s
✔ load module . 2.0s
✔ looking for module 0.0s

Full trace at https://dagger.cloud/eunomie/traces/781b5af5181c37aff740df96da7afd1b
```

(with an exit code `1`)

So I guess I'm missing something here.

Otherwise this is working with constructors from #9523

helderco:
Like this:
```go
package main

type Goc struct {
	Version string
}

func (m *Goc) WithVersion(version string) *Goc {
	m.Version = version
	return m
}

func (m *Goc) Check() string {
	if m.Version == "" {
		return "You don't have a version"
	}
	return "You have version " + m.Version
}
```
There's a reason for using the `Check` function. It's to check if it's getting deserialized.

Then:
```console
> dagger shell -c 'with-version 1.23 | check'
You have version 1.23
```

sdk/java/dagger-java-annotation-processor/src/test/resources/io/dagger/java/module/DaggerJava.java
helderco: I'm talking about the one the SDK client does when getting `parentJson` and returning object results.

eunomie:
ha ok 🤦  I misunderstood, I thought you were saying the with* functions where created automatically

So yes, it's working as expected:

```java
@Object
public class FieldModule extends AbstractModule {
  public String version;

  public FieldModule() { super(); }

  @Function
  public FieldModule withVersion(String version) {
    this.version = version;
    return this;
  }

  @Function
  public String echoVersion() {
    return version;
  }
}
```

And with that

```console
> dagger shell -c 'with-version 1.2 | echo-version'
✔ connect 0.3s
✔ load module . 14.9s

✔ fieldModule: FieldModule! 0.0s
✔ .withVersion(version: "1.2"): FieldModule! 1.2s
✔ .echoVersion: String! 1.2s

1.2
```

eunomie:
Is there a good place I can put a test module and run the above `dagger shell -c 'with-version 1.2 | echo-version'` so I ensure this behavior is covered by high level tests?

helderco:
We should add a suite for Java, like this community maintained one:

https://github.com/dagger/dagger/blob/736eabb66f8cbe32ecac21cd49d2696e41110084/core/integration/module_php_test.go

It's how we do acceptance tests for SDKs. Here's another with more coverage, maintained by me:

https://github.com/dagger/dagger/blob/736eabb66f8cbe32ecac21cd49d2696e41110084/core/integration/module_python_test.go

\cc @jedevc

eunomie:
I'll have a look at that then, thanks 👍

helderco (approved):
LGTM 👍

helderco:
Need to regenerate:
```
dagger call -m .github generate -o .
```

jedevc:
Regen command is `dagger call -m .github generate directory --path=.github/workflows export --path=.github/workflows --wipe`

core/integration/module_java_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+		out, err := modGen.
+			With(daggerQuery(`{bare{containerEcho(stringArg:"hello"){stdout}}}`)).
+			Stdout(ctx)
+		require.NoError(t, err)
+		require.JSONEq(t, `{"bare":{"containerEcho":{"stdout":"hello\n"}}}`, out)
+	})
+}
+
+func (JavaSuite) TestFields(_ context.Context, t *testctx.T) {
```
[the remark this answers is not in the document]
eunomie: I'd say 🤷 
At least it works on my machine, the test passes and if I change the test (to break it and be sure the test is run) it doesn't pass anymore.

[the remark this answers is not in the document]
helderco: That's `TestInit`. `TestFields` is using a different method, but pulling in the test module in `testdata/modules/java/fields`.

PR #9523, opened 2025-02-06
sdk/java: constructors

eunomie:
> [!NOTE]
> ~~To be merged after #9605~~ (merged)

---

Allow to define a constructor on classes for an module object.

- a default, empty, constructor calling `super` is required
- only the first non empty constructor will be registered

This constructor can have as the first parameter a `io.dagger.client.Client` arg.
In this case the Dagger client will be injected. Please be sure to then call super with the client:

```java
@object
public class MyModule extends AbstrctModule {
  // to allow creation of empty object
  // mandatory
  public MyModule() {
    super();
  }

  // this constructor will be registered with
  // a string parameter and the dagger client
  // will automatically be injected at creation
  // so it can be used in the constructor body
  public MyModule(Client dag, String stringArg) {
    super(dag);
      // ...
  }
}
```

You do _not_ require to have the client as first parameter.
If your constructor doesn't requires it, it can be omitted.

You can use annotations on constructors' arguments as in functions. For instance:

```java
public String version;

public MyModule(@default("1.2.3") String version) {
  this.version = version;
}
```


helderco:
I haven't seen the code yet, but just for clarity, the "constructor" feature is just about a single constructor. It's the one that creates an instance of the main object, sometimes called the module's "entrypoint" like in the Daggerverse . Other objects are constructed through another object's method.

eunomie:
I put it in review, but this needs to be merged after #9605

eunomie:
> I haven't seen the code yet, but just for clarity, the "constructor" feature is just about a single constructor. It's the one that creates an instance of the main object, sometimes called the module's "entrypoint" like in the Daggerverse. Other objects are constructed through some object's method.

I added checks to ensure only one object (the main one) can expose constructors, and only one constructor can be exposed.

eunomie:
The java failure is because I'm using environment variable, I'll fix that

vikram-dagger:
> I put it in review, but this needs to be merged after #9605

I suggest making this a draft to avoid an accidental merge, until it's ready

eunomie:
I just rebased the branch, nothing else has change

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -117,6 +126,34 @@ ModuleInfo generateModuleInfo(Set<? extends TypeElement> annotations, RoundEnvir
                     .formatted(qName));
           }
 
+          Optional<ConstructorInfo> constructorInfo = Optional.empty();
+          if (mainObject) {
+            Optional<? extends Element> constructorDef =
+                typeElement.getEnclosedElements().stream()
+                    .filter(elt -> elt.getKind() == ElementKind.CONSTRUCTOR)
+                    .filter(elt -> !((ExecutableElement) elt).getParameters().isEmpty())
+                    .findFirst();
```
jcsirot, line 135: Maybe we should throw an exception when there is more than one constructor

eunomie: Yeah, I didn't really know. I was wondering if we should add a specific annotation to specify which one is the right one.

eunomie: I've changed my mind and went on your direction. I know throw an exception if there's more than 1 non empty constructor. That way we ensure the behavior is the one we expect.

sdk/java/dagger-java-annotation-processor/src/test/resources/io/dagger/gen/entrypoint/entrypoint.java
```diff
@@ -133,18 +135,24 @@ private ModuleID register() throws ExecutionException, DaggerQueryException,
                         .withArg("a", dag.typeDef().withKind(TypeDefKind.FLOAT_KIND))
                         .withArg("b", dag.typeDef().withKind(TypeDefKind.FLOAT_KIND)))
                 .withField("source", dag.typeDef().withObject("Directory"), new TypeDef.WithFieldArguments().withDescription("Project source directory"))
-                .withField("version", dag.typeDef().withKind(TypeDefKind.STRING_KIND)));
+                .withField("version", dag.typeDef().withKind(TypeDefKind.STRING_KIND))
+                .withConstructor(
+                    dag.function("New",
```
helderco, line 140: I don't think it makes a difference in practice but can be confusing when other people try to understand from this. Constructor name should be empty:
```suggestion
                    dag.function("",
```

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -513,11 +610,15 @@ private static CodeBlock functionInvoke(ObjectInfo objectInfo, FunctionInfo fnIn
 
   public static CodeBlock withFunction(ObjectInfo objectInfo, FunctionInfo fnInfo)
       throws ClassNotFoundException {
+    boolean isConstructor = fnInfo.name().equals("<init>");
     CodeBlock.Builder code =
         CodeBlock.builder()
-            .add("\n                dag.function($S,", fnInfo.name())
+            .add("\n                dag.function($S,", isConstructor ? "New" : fnInfo.name())
```
helderco, line 616:
```suggestion
            .add("\n                dag.function($S,", isConstructor ? "" : fnInfo.name())
```
See:

https://github.com/dagger/dagger/blob/469639e6be403eea3e1a2cc7044a99a2b2523005/sdk/python/src/dagger/mod/_module.py#L170-L171

https://github.com/dagger/dagger/blob/469639e6be403eea3e1a2cc7044a99a2b2523005/sdk/python/src/dagger/mod/_resolver.py#L219-L220

helderco (approved):

sdk/java/dagger-java-annotation-processor/src/test/resources/io/dagger/gen/entrypoint/entrypoint.java
eunomie: ok, I'll change that, to have it right from start.

But I took it from the generated Go code. Is there a place where those rules are defined?

For instance if you `dagger develop` in `modules/go` you'll get:

```go
WithConstructor(
    dag.Function("New",
        dag.TypeDef().WithObject("Go")).
```

helderco: > For instance if you `dagger develop` in `modules/go` you'll get:
> 
> ```go
> WithConstructor(
>     dag.Function("New",
>         dag.TypeDef().WithObject("Go")).
> ```

Oh, I wasn't actually aware of that. At one point a constructor was registered by a function with an empty name. Just like the way you know to call it is via a `FunctionCall.name` with empty value. 

Now with the `withConstructor` call it doesn't really matter, and is wiped server side:

https://github.com/dagger/dagger/blob/1d7fa24fb5d02ffefaf9760fa92df70c2ab28c6a/core/schema/module.go#L298-L301

I suppose in Go it was just easier to leave the function name in to reuse the general function to typedef conversion (that's why the name is "New", because in Go the constructor is always called "New"), but I still prefer not to register with a name so it doesn't confuse others that it's something that may be needed. Especially someone looking into adding modules support to an SDK and is looking to other SDKs for comparison 🙂

PR #9526, opened 2025-02-07
bundle tsx in the engine image to speed up the TS sdk's node-runtime module init

cwlbraa:
with a cold cache, this shaves ~7-10 seconds off TS module init with a hello world-ish module. prior to this change, it'd take ~31s on my machine and after it's somewhere in the 24s range. it makes my dev image marginally larger, which is a bit of a tradeoff, but the dev image is already 449Mb and tsx ends up being like 5mb additional. 

~sidenote: mucking with module-sdk "runtime" go.mod dependencies is pretty error-prone in its current state... it matters quite a bit for module init performance that the go-SDK-based SDKs share exact dependencies with the go SDK, and blindly adding `require`s or naively `go mod tidy`-ing can mess that up... i think the intent is that we use `dagger develop` to make things match? that seemed to work decently well, although things may still be off by patch-versions, especially indirect dependencies.~ turns out module code generation already [does this slick thing where it bumps outdated go.mods to the SDK's go.mod version](https://github.com/dagger/dagger/blob/115a4552ed99de13db32aeee1de875490995eca6/cmd/codegen/generator/go/generator.go#L271-L280).

really most the speedup here comes from bundling tsx ~the node_modules end up being pretty marginal because package locks don't tend to align exactly between our container build and individual module builds. we could probably improve that using an approach similar to the go SDK.~ tore the node_modules bundling out of this PR, will try again later on main.

~we prolly need to more holistically document how to manage the go mod dependencies in all these SDK modules. started a thread to do that [here](https://discord.com/channels/707636530424053791/1338997699214839809).~ the go.mods are pretty well-managed by dependabot, as it turns out.

big ups to @TomChv for dealing with my incessant questions :)

### DIFF
--- a/.dagger/build/sdk.go
+++ b/.dagger/build/sdk.go
@@ -3,6 +3,7 @@ package build
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"runtime"
 
 	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
@@ -11,6 +12,7 @@ import (
 
 	"github.com/dagger/dagger/.dagger/consts"
 	"github.com/dagger/dagger/.dagger/internal/dagger"
+	"github.com/dagger/dagger/sdk/typescript/runtime/tsdistconsts"
 )
 
 type sdkContent struct {
@@ -69,7 +71,14 @@ func (build *Builder) pythonSDKContent(ctx context.Context) (*sdkContent, error)
 	}, nil
 }
 
+const TypescriptSDKTSXVersion = "4.15.6"
+
 func (build *Builder) typescriptSDKContent(ctx context.Context) (*sdkContent, error) {
+	tsxNodeModule := dag.Container().
+		From(tsdistconsts.DefaultNodeImageRef).
+		WithExec([]string{"npm", "install", "-g", fmt.Sprintf("tsx@%s", TypescriptSDKTSXVersion)}).
+		Directory("/usr/local/lib/node_modules/tsx")
+
 	rootfs := dag.Directory().WithDirectory("/", build.source.Directory("sdk/typescript"), dagger.DirectoryWithDirectoryOpts{
 		Include: []string{
 			"src/**/*.ts",
@@ -86,9 +95,22 @@ func (build *Builder) typescriptSDKContent(ctx context.Context) (*sdkContent, er
 		},
 	})
 
+	sdkNodeModules := dag.Container().
+		From(tsdistconsts.DefaultNodeImageRef).
+		WithWorkdir("/work").
+		WithDirectory("/work/sdk", rootfs).
+		WithoutEntrypoint().
+		WithMountedCache("/root/.npm", dag.CacheVolume(fmt.Sprintf("npm-cache-node-%s", tsdistconsts.DefaultNodeVersion))).
+		WithFile("/work/package.json", rootfs.File("./runtime/template/package.json")).
+		WithExec([]string{"npm", "install", "--package-lock-only"}).
+		WithExec([]string{"npm", "ci"}).
+		Directory("/work/node_modules")
+
 	sdkCtrTarball := dag.Container().
 		WithRootfs(rootfs).
 		WithFile("/codegen", build.CodegenBinary()).
+		WithDirectory("/tsx_module", tsxNodeModule).
+		WithDirectory("/sdk_node_modules", sdkNodeModules).
 		AsTarball(dagger.ContainerAsTarballOpts{
 			ForcedCompression: dagger.ImageLayerCompressionZstd,
 		})
--- a/.dagger/build/sdk.go
+++ b/.dagger/build/sdk.go
@@ -95,22 +95,10 @@ func (build *Builder) typescriptSDKContent(ctx context.Context) (*sdkContent, er
 		},
 	})
 
-	sdkNodeModules := dag.Container().
-		From(tsdistconsts.DefaultNodeImageRef).
-		WithWorkdir("/work").
-		WithDirectory("/work/sdk", rootfs).
-		WithoutEntrypoint().
-		WithMountedCache("/root/.npm", dag.CacheVolume(fmt.Sprintf("npm-cache-node-%s", tsdistconsts.DefaultNodeVersion))).
-		WithFile("/work/package.json", rootfs.File("./runtime/template/package.json")).
-		WithExec([]string{"npm", "install", "--package-lock-only"}).
-		WithExec([]string{"npm", "ci"}).
-		Directory("/work/node_modules")
-
 	sdkCtrTarball := dag.Container().
 		WithRootfs(rootfs).
 		WithFile("/codegen", build.CodegenBinary()).
 		WithDirectory("/tsx_module", tsxNodeModule).
-		WithDirectory("/sdk_node_modules", sdkNodeModules).
 		AsTarball(dagger.ContainerAsTarballOpts{
 			ForcedCompression: dagger.ImageLayerCompressionZstd,
 		})
--- a/dagger.json
+++ b/dagger.json
@@ -1,72 +1,75 @@
 {
-  "name": "dagger-dev",
-  "engineVersion": "v0.15.4",
-  "sdk": {
-    "source": "go"
-  },
-  "dependencies": [
-    {
-      "name": "alpine",
-      "source": "modules/alpine"
+    "name": "dagger-dev",
+    "engineVersion": "v0.15.4",
+    "sdk": {
+        "source": "go"
     },
-    {
-      "name": "dagger-cli",
-      "source": "cmd/dagger"
-    },
-    {
-      "name": "dirdiff",
-      "source": "modules/dirdiff"
-    },
-    {
-      "name": "docusaurus",
-      "source": "github.com/levlaz/daggerverse/docusaurus@main",
-      "pin": "47f5206067011dad0f581d4e7db7afddda32acd0"
-    },
-    {
-      "name": "dotnet-sdk-dev",
-      "source": "sdk/dotnet/dev"
-    },
-    {
-      "name": "elixir-sdk-dev",
-      "source": "sdk/elixir/dev"
-    },
-    {
-      "name": "go",
-      "source": "modules/go"
-    },
-    {
-      "name": "graphql",
-      "source": "modules/graphql"
-    },
-    {
-      "name": "helm",
-      "source": "helm"
-    },
-    {
-      "name": "php-sdk-dev",
-      "source": "sdk/php/dev"
-    },
-    {
-      "name": "ps-analyzer",
-      "source": "modules/ps-analyzer"
-    },
-    {
-      "name": "python-sdk-dev",
-      "source": "sdk/python/dev"
-    },
-    {
-      "name": "shellcheck",
-      "source": "modules/shellcheck"
-    },
-    {
-      "name": "version",
-      "source": "version"
-    },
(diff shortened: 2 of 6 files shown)


.dagger/consts/versions.go
```diff
@@ -9,8 +9,9 @@ const (
 )
 
 const (
-	GolangVersion = distconsts.GolangVersion
-	GolangImage   = distconsts.GolangImage
+	GolangVersion   = distconsts.GolangVersion
+	GolangImage     = distconsts.GolangImage
+	TypescriptImage = "node:22.11.0-alpine"
```
cwlbraa, line 14: this really shouldn't be re-declared here, its actual source of truth is currently in `sdks/typescript/runtime/main.go`. is it inadvisable to link that here? should these be functions in `typescript-sdk-dev`? I think some of the other module sdks do it that way.

sdk/typescript/runtime/main.go
```diff
@@ -275,8 +275,10 @@ func (t *TypescriptSdk) Base() (*dagger.Container, error) {
 			WithMountedCache("/root/.npm", dag.CacheVolume(fmt.Sprintf("npm-cache-%s-%s", runtime, version))).
 			WithMountedCache("/root/.cache/yarn", dag.CacheVolume(fmt.Sprintf("yarn-cache-%s-%s", runtime, version))).
 			WithMountedCache("/root/.pnpm-store", dag.CacheVolume(fmt.Sprintf("pnpm-cache-%s-%s", runtime, version))).
-			// Install tsx
-			WithExec([]string{"npm", "install", "-g", "tsx@4.15.6"}), nil
+			// install tsx from its bundled location in the engine image
+			WithDirectory("/usr/local/lib/node_modules/tsx", t.SDKSourceDir.Directory("/tsx_module")).
```
cwlbraa, line 279: this also works with WithMountedDirectory, and might be a little faster, but i presume mounting like that puts us at risk of different modules overwriting?

sipsma: > but i presume mounting like that puts us at risk of different modules overwriting

by "different modules" do you mean "node modules" or "dagger modules"?

The only risk of WithMountedDirectory I can see is that rather than overlay onto anything existing in `/usr/local/lib/node_modules/tsx`, it will replace that directory entirely.

So if it's a safe assumption that that dir is always empty to start with then I think WithMountedDirectory should be good.

(sorry if that's just begging the question you were asking, the overlapping terminology might be confusing me 😅)

cwlbraa: I mean dagger modules. Mounts are read-only? Or does each container get its own view of the dir?

.dagger/consts/versions.go
TomChv: Interesting question! The source of trust should indeed be from `runtime/main.go` but, because you only install TSX prior to calling the TS runtime, you need a default base image at some point.
I think it shouldn't be defined here, because it creates some assumptions that the TS runtime image static while it's not the case.
Maybe just inline it in `build/sdk.go` as a base image to install utility tools to the runtime.

sdk/typescript/runtime/main.go
TomChv: I think it's fine to use mounted dir, the tsx directory should never change during the execution, it's just a binary binding.
Additional bin or node_modules installation will not impact the `tsx` dir.

TomChv (commented):
Very neat improvement, I like how you did your implemention since it doesn't impact custom user image!
Very nice.

I think we can merge if all tests are green and then you can keep going with node_modules :)

.dagger/build/sdk.go
```diff
@@ -70,6 +70,10 @@ func (build *Builder) pythonSDKContent(ctx context.Context) (*sdkContent, error)
 }
 
 func (build *Builder) typescriptSDKContent(ctx context.Context) (*sdkContent, error) {
+	sdkCache := dag.Container().
+		From(consts.TypescriptImage).
+		WithExec([]string{"npm", "install", "-g", "tsx@4.15.6"})
```
TomChv, line 75: That's the same for  node_modules installation btw, you don't even have to care about the package manager you use in this sdkCache.
As long as you can get an optimized artifact to mount to the tarball, it's perfect :)

.dagger/consts/versions.go
cwlbraa: > I think it shouldn't be defined here, because it creates some assumptions that the TS runtime image static while it's not the case.

yeah we're saying the same thing, but `build/sdk.go` is only marginally better... is there something wrong with putting it in typescript-sdk-dev and having both runtime/main.go and build/sdk.go import that? or importing `sdks/typescript/runtime/` here in `build/sdk.go`? (fwiw the second one _feels_ wrong to me)

sdk/typescript/runtime/main.go
cwlbraa: two ground truths i'm missing, but i trust y'all saying mounts work here and will swap over to that.

@sipsma implicitly, `WithMountedDirectory` gives us a copy-on-write/overlay mount here? fwiw this is what i'd assume, and i feel like i should be able to validate that assumption myself, but beneath [builtinContainer](https://github.com/cwlbraa/dagger/blob/6aed55a063b42a19d78d9c3f564002bfe7e9f8eb/core/schema/host.go?plain=1#L42) there's like a 8 layers of Buildkit abstraction.

@TomChv can users set a different (`-g` global?) version of tsx in their package.json? this is a very hard question to get answers to across 2 runtimes and 4 package managers lol

sipsma: > implicitly, WithMountedDirectory gives us a copy-on-write/overlay mount here? fwiw this is what i'd assume, and i feel like i should be able to validate that assumption myself, but beneath [builtinContainer](https://github.com/cwlbraa/dagger/blob/6aed55a063b42a19d78d9c3f564002bfe7e9f8eb/core/schema/host.go?plain=1#L42) there's like a 8 layers of Buildkit abstraction.

Yes exactly, the directory you provide to withMountedDirectory serves as the base layers and you get a new overlay layer on top, so no possibility of concurrent usage interfering with each other.

Generally that's how everything works except for `withMountedCache`, which is the escape hatch for when you actually do want concurrent users to share the same writable filesystem.

sdk/typescript/runtime/go.mod
```diff
@@ -1,11 +1,17 @@
...
-toolchain go1.23.2
+toolchain go1.23.6
 
 require github.com/iancoleman/strcase v0.3.0
 
+require github.com/dagger/dagger/sdk/typescript/runtime/tsdistconsts v0.15.3
+
+require go.opentelemetry.io/auto/sdk v1.1.0 // indirect
+
+replace github.com/dagger/dagger/sdk/typescript/runtime/tsdistconsts => ./tsdistconsts/
```
cwlbraa, line 13: @helderco

.dagger/build/sdk.go
```diff
@@ -86,9 +95,22 @@ func (build *Builder) typescriptSDKContent(ctx context.Context) (*sdkContent, er
 		},
 	})
 
+	sdkNodeModules := dag.Container().
+		From(tsdistconsts.DefaultNodeImageRef).
+		WithWorkdir("/work").
+		WithDirectory("/work/sdk", rootfs).
+		WithoutEntrypoint().
+		WithMountedCache("/root/.npm", dag.CacheVolume(fmt.Sprintf("npm-cache-node-%s", tsdistconsts.DefaultNodeVersion))).
```
sipsma, line 103: nit: do we use an npm cache volume anywhere else in our modules like this? If so because of module-based cache volume namespacing it would be efficient-er to define a function in one module that returns a consistent cache volume to use, so then they are all shared

This is not a big deal at all, just an idea if quick to implement

sipsma (approved):

.dagger/build/sdk.go
cwlbraa: does the typescript SDK count as a module in this case? otherwise we do not, but the SDK not only uses this for its own purposes, it also mounts this for user modules.

sdk/typescript/runtime/tsdistconsts/consts.go
```diff
@@ -0,0 +1,11 @@
...
+
+const (
+	DefaultNodeVersion  = "22.11.0" // LTS version, JOD (https://nodejs.org/en/about/previous-releases)
+	nodeImageDigest     = "sha256:b64ced2e7cd0a4816699fe308ce6e8a08ccba463c757c00c14cd372e3d2c763e"
+	DefaultNodeImageRef = "node:" + DefaultNodeVersion + "-alpine@" + nodeImageDigest
+
+	DefaultBunVersion  = "1.1.38"
+	bunImageDigest     = "sha256:5148f6742ac31fac28e6eab391ab1f11f6dfc0c8512c7a3679b374ec470f5982"
+	DefaultBunImageRef = "oven/bun:" + DefaultBunVersion + "-alpine@" + bunImageDigest
+)
```
TomChv, line 11: Can't you just put them in a children package of the runtime? Does it make sense to create a go package just for that?

.dagger/build/sdk.go
TomChv: How fast do we get thanks to this bundling? I saw 7s faster but it should be much more normally since installing deps always take around +10seconds.

sdk/typescript/runtime/tsdistconsts/consts.go
cwlbraa: i think so. i just started doing it this way before i got answers from you and helder, i'll see about doing it that way.

.dagger/build/sdk.go
cwlbraa: there are 2 bundlings: tsx saves 7 seconds pretty consistently. node modules saves 0-2s on `corepack use yarn` because irl some package versions don't end up matching and it takes a second to re-compute integrity across package managers.

cwlbraa: if im revisiting this to fix the extraneous go.mod i'll double check these numbers.

sdk/typescript/runtime/tsdistconsts/consts.go
cwlbraa: @TomChv im pretty sure this doesn't work bc runtime/ is `package main`?

cwlbraa: think i got this working here https://github.com/dagger/dagger/pull/9526/commits/4fff7a66617c3551100d8f96f00cfdb8dd3b7717

.dagger/build/sdk.go
cwlbraa: yeah, revisited, the node_modules caching adds no speedup here, so theoretically slower because of the increased image size. I'm tearing it out of this PR, and will take another shot at it in an isolated setup.

exporting from DMs, too, I may've found that one of the cache volumes doesn't hit, but fixing it will require x-package-manager research:

ts sdk:
```
WithMountedCache("/root/.cache/yarn", dag.CacheVolume(fmt.Sprintf("yarn-cache-%s-%s", runtime, version))).
````
shelling in there:
```
dagger /src/.dagger $ yarn cache dir
/usr/local/share/.cache/yarn/v6
```

dagger.json
cwlbraa, line 1: will this autoindent thrash against `dagger develop` or other folks' autoformat? do i need to modify my vim config to be less aggro here? cc @jedevc bc i know we're both pretty close to lazyvim defaults. 

(fwiw this happens async, no way to avoid with the usual `:noa w`)

PR #9527, opened 2025-02-07
use extracted testctx package

vito:
Finally getting around to this TODO since I've needed it in 3 separate repos now.

Made some changes to the API along the way - I think this new model feels a bit more natural.


dagql/idtui/golden_test.go
```diff
@@ -35,33 +33,19 @@ import (
 	"github.com/dagger/dagger/dagql/idtui"
 	"github.com/dagger/dagger/engine/slog"
 	"github.com/dagger/dagger/internal/testutil"
-	"github.com/dagger/dagger/testctx"
+	"github.com/dagger/testctx"
+	"github.com/dagger/testctx/otelmw"
```
cwlbraa, line 37: out of context, "mw" doesn't immediately read as "middleware"... it makes sense to alias as that but idk about having that as the package name

vito: How about `oteltest`? Looks like it's largely unclaimed - just a some unused packages, and an official but deprecated package in a module that you'd have to go out of your way to install: https://pkg.go.dev/search?q=oteltest

testctx/testctx.go
```diff
@@ -1,366 +0,0 @@
...
-		})
-}
-
-func N(b *TB[*testing.B]) int {
-	return b.RunnableTB.(*testing.B).N
-}
-
-const TestCtxTypeAttr = "dagger.io/testctx.type"
-const TestCtxNameAttr = "dagger.io/testctx.name"
-const TestCtxPrewarmAttr = "dagger.io/testctx.prewarm"
```
cwlbraa, line 159: am i missing it somewhere or did we lose this metadata?

(i did read testctx and i can tell that at least `type` and `suite` are settable)

cwlbraa (approved):
approving with a note to self to make an effort to kill some of those custom spans later, i think i probably went overboard.

PR #9529, opened 2025-02-07
Fix java module entrypoint Scalar typedef code generation

jcsirot:
This PR fixes a bug in the code generation of a Java module's entry point. When a function's parameter or return type is a scalar, the engine throws the error `failed to find mod type for function "xxx" return type/arg` because scalar types are defined using `dag.typeDef().withObject`.

The PR fixes the generated code for module functions registration by defining scalar types with `dag.typeDef().withScalar`.


eunomie (commented):
👍 sounds good to me

eunomie (approved):

Issue #9524, opened 2025-02-06
Cannot use context directories from git alongside PAT auth


jedevc:
Based on this thread with @nipuna-perera in discord: https://discord.com/channels/707636530424053791/1336068640168218756/1336093598994665493

Module setup:

```go
// testmod
package main

import (
    "dagger/testmod/internal/dagger"
)

type Testmod struct{}

func New() *Testmod {
    return &Testmod{}
}

// Returns a container that echoes whatever string argument is provided
func (m *Testmod) ContainerEcho(
    //+defaultPath=./certs/test.crt
    cert *dagger.File,
) *dagger.Container {
    return dag.Container().
        From("alpine:3.12").
        WithFile("/src/cert.crt", cert).
        WithWorkdir("/src").
        WithExec([]string{"cat", "/src/cert.crt"})
}
```

```go
// tests (has testmod as a dependency)
package main

import (
    "dagger/tests/internal/dagger"
)

type Tests struct{}

// Returns a container that echoes whatever string argument is provided
func (m *Tests) ContainerEcho() *dagger.Container {
    return dag.Testmod().ContainerEcho()
}
```

When running `dagger call container-echo` (in `tests`), you get:

```
failed to add client resources from ID: failed to add secret from source client 4sfuyti2ov0duk9p8xlz90b4d: secret xxh3:a1b22fd50fdb4b95 not found in other store
```

This is likely due to the fact that the `File` that the returned value from the first module has a secret associated with it (in it's ID). But, we never put that secret into the session, so we're never able to load it back out and transfer it.

If this is the case, the fix should be simple - we should just add the secret to the session.

jedevc:
Discovered we don't need to push both modules, you can have one local, and one remote. For testing, this is nice, we can just use this https://github.com/dagger/dagger-test-modules/blob/main/context-dir/src/main.go behind a PAT.

nipuna-perera:
This is how I test. One local, one remote.

PR #9530, opened 2025-02-07
fix: transfer secret during `LoadContext`

jedevc:
Fixes https://github.com/dagger/dagger/issues/9524.

`LoadContext` could easily produce a `Directory` that's dependent on a `Secret`. We have logic to ensure that `Secret`s are all correctly transferred between various clients.

The underlying bug was occurring because of this chain of events:
- Main client session loads all the dependencies, and we inject git credentials
- We enter the top-level module `A` client session, and call the target function `ContainerEcho`, which calls `B`'s `ContainerEcho`, which in turn calls `LoadContext` to load the contextual argument from the dependency git context
- `B` returns a value dependent on the loaded git context, which then triggers the secret transfer code to try and transfer the git credential secret from `B` back to `A`.
	- However! :warning: this fails, because the credential was never loaded into `B`.

The fix is pretty simple - when we call `LoadContext` we need to transfer the secret from the main session which loaded all the modules. This fixes the bug, because now `B` will have access to the secret, so it can be transferred out to `A`.

I've manually verified this fix, but to test this, I'll need a private copy of the module here: https://github.com/dagger/dagger-test-modules/blob/main/context-dir/src/main.go (not sure if we have a PAT setup for this purpose :thinking:)

### DIFF
--- a/core/modulesource.go
+++ b/core/modulesource.go
@@ -20,6 +20,7 @@ import (
 	"github.com/dagger/dagger/dagql/call"
 	"github.com/dagger/dagger/engine"
 	"github.com/dagger/dagger/engine/buildkit"
+	"github.com/dagger/dagger/engine/server/resource"
 	"github.com/dagger/dagger/engine/slog"
 )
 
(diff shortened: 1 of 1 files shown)


cwlbraa (approved):
fix makes sense. lowkey, though, i'm kinda surprised that this bug is triggerable with exactly 0 mentions of the word "secret" in the issue... is "cert" a keyword here or something? is defaultPath a secrets thing?

marcosnils:
IIRC we do have tests for the PAT support. I think it shouldn't be difficult to come up with a test case that triggers this bug

jedevc:
> fix makes sense. lowkey, though, i'm kinda surprised that this bug is triggerable with exactly 0 mentions of the word "secret" in the issue... is "cert" a keyword here or something? is defaultPath a secrets thing?

Ah, the secret is the secret for loading the module. The one set here: https://github.com/dagger/dagger/blob/4feedf64eeaf00b3d2f3807912544e00fa7e454a/core/schema/git.go#L215-L239

sipsma (approved):

PR #9531, opened 2025-02-07
feat: add client generation outside module

TomChv:
### Summary

This PR enables calling module dependencies from a regular client library. The setup is quite simple:
1. Initialize a project in your favorite language
2. Initialize a dagger module `dagger init`
3. Install the dagger library
4. Install any dependencies
5. Run client generator `dagger client add --generator=<generator>`
6. Import the generated client and use it like if the library :)

This introduce few changes in an engine:
- A module can be initialized even if it does not contains sources (only dependency).
- Client gen has a new flag to specify if the generation is for a standalone client or not.
- If the client is standalone, it will depends on the published on (I need to add a flag in client generator to handle local changes) Note that the setup might still requires few manual operation on the user codebase, like installing the library, execute go mod tidy or configure the tsconfig to override the import path. These can be automated.
- Extends the SDK interface with a new `ClientGenerator` method This starts the work to split the SDK interface into smaller component The engine could then adapt his behaviour depending on which function the SDK implement (codegen, runtime, modgen etc..)
- Change dagger run to load a module if one is defined and serving the module with its dependencies.

### Client Generator interface

```go
// ClientGenerator is an interface that a module can implements to give client generation capabilities.
//
// The generated client is standalone and can be used in any project, even if no source code module is
// available.
type ClientGenerator interface {
	// Generate client returns client binding for the module with the given dependencies at the given output directory.
	// The generated client will be placed in the same directory as the module source root dir
	// and contains bindings for all of the module's dependencies in addition to the
	// core API and the module itself if it got source code.
	//
	// It's up to that function to update the source root directory with additional
	// configurations if needed.
	// For example (executing go mod tidy, updating tsconfig.json etc...)
	//
	// The generated client should use the published library version of the SDK.
	// However, if the last parameter is set to true, a copy of the current SDK library
	// should be copied to test the generated client with latest changes.
	// NOTE: this should only be used for testing purposes.
	GenerateClient(context.Context, dagql.Instance[*ModuleSource], *ModDeps, string, bool) (dagql.Instance[*Directory], error)
}
```

### TODO
- [x] Enable module loading without sources
- [x] Add client generator to the engine
- [ ] Add client generator function to our SDKs ⌛ (Python needed)
- [x] Update code generation to generate bindings ⌛ (cleanup code)
- [x] Add an option to copy local required files if needed (for local dev/tests mostly)
- [x] Write integrations tests

Fixes DEV-4939, Fixes DEV-4976, Fixes DEV-3024

### Example usage

```
go mod init dagger.io/demo
dagger init
dagger install github.com/shykes/hello
dagger client add --generator go
```

```go
package main

import (
	"context"
	"fmt"

	"dagger.io/demo/dagger" // Use local lib
)

func main() {
	ctx := context.Background()

	dag := dagger.Connect()

    out, err := dag.Hello().Hello(ctx)
    if err != nil { panic (err) }
    fmt.Println(out)
}
```

```shell
dagger run go run main.go
```


sdk/java/dagger-codegen-maven-plugin/src/main/resources/schemas/schema-0.15.4-010101000000-dev-8b9c9566abfc.json
jedevc, line 1: :eyes: these shouldn't be here.

cmd/dagger/client_add.go
```diff
@@ -0,0 +1,85 @@
...
+	generator string
+	localSDK bool
+)
+
+func init() {
+	clientAddCmd.Flags().StringVar(&generator, "generator", "", "Generator to use to generate the client")
+	clientAddCmd.Flags().BoolVar(&localSDK, "local-sdk", false, "Use local SDK dependency")
+}
+
+var clientAddCmd = &cobra.Command{
```
jedevc, line 24: Hm, so if I want to regenerate this, what do I do? Do I run the add command again?

It *feels* like we might want to just be able to run `dagger develop` and it would re-generate any existing code. Do I need to specify the generator again, and set the right directory again?

*Ideally* it feels like `dagger develop` would just work here :thinking: Or at least a `dagger client generate` or `dagger client sync`.

Don't really know, but the generated clients *should* end up in the `dagger.json` right so we can easily regen them? cc @shykes I know you had ideas around the CLI here.

cmd/codegen/generator/generator.go
```diff
@@ -42,6 +42,10 @@ type Config struct {
 
 	// Merge indicates whether to merge the module deps with the existing project.
 	Merge *bool
+
+	ClientOnly bool
+
+	LocalSDK bool
```
jedevc, line 48: doc-comments :pray:

cmd/codegen/generator/go/generator.go
```diff
@@ -64,6 +64,48 @@ func (g *GoGenerator) Generate(ctx context.Context, schema *introspection.Schema
 
 	mfs := memfs.New()
 
+	if g.Config.ClientOnly {
```
jedevc, line 67: Can we split this into an entirely separate function? It doesn't really share *any* logic with the rest of this, we could have `GenerateClient` here, as another method in the interface. This then matches the `generateClient` in the engine SDK interface.

cmd/codegen/generator/go/generator.go
```diff
@@ -319,7 +361,7 @@ func generateCode(
 	fset *token.FileSet,
 	pass int,
 ) error {
-	funcs := templates.GoTemplateFuncs(ctx, schema, schemaVersion, cfg.ModuleName, cfg.ModuleParentPath, pkg, fset, pass)
+	funcs := templates.GoTemplateFuncs(ctx, schema, schemaVersion, cfg.ModuleName, cfg.ModuleParentPath, pkg, fset, pass, cfg.ClientOnly, cfg.LocalSDK)
```
jedevc, line 364: :thinking: this is getting a bit long now, probably worth just passing in `cfg` directly.

cmd/codegen/generator/go/generator.go
```diff
@@ -64,6 +64,48 @@ func (g *GoGenerator) Generate(ctx context.Context, schema *introspection.Schema
...
 	mfs := memfs.New()
 
+	if g.Config.ClientOnly {
+		layers := []fs.FS{mfs}
+
+		// Use the published package library for external dagger packages.
+		packageImport := "dagger.io/dagger"
+
+		// If localSDK is needed, we need to add local files to the overlay and change the package import
+		if g.Config.LocalSDK {
```
jedevc, line 74: Ahhh something I just realized. The client that we generate *isn't* the same as using the `dagger.io` one.

The main difference at the moment is the `client.go` - the `sdk/go/client.go` has automatic provisioning. While the one that's generated for modules is more lightweight, and expects to be used in the context of `dagger run`.

Since generally I (personally) would like to get rid of the automatic provisioning entirely, and make the two clients essentially the same, I'm kind of okay with treating this as *just* separate. But it does make it harder for someone to pick up if they're using our upstream library - because now they *need* to switch to using `dagger run`.

Instead of using the module client, could we use the `client.go` that we've already written? They *should* have identical methods/fields.

cmd/codegen/generator/go/templates/src/_dagger.gen.go/defs.go.tmpl
```diff
@@ -123,7 +126,7 @@ func (e *ExecError) Unwrap() error {
 {{ if eq .Kind "ENUM" }}{{ template "_types/enum.go.tmpl" . }}{{ end }}
 {{ end }}
 
-{{ if IsModuleCode }}
+{{ if or (IsModuleCode) (IsStandaloneClient) }}
 type Client struct {
 	query *querybuilder.Selection
 	client graphql.Client
```
jedevc, line 132: Ahhh something I just realized! :scream: The client that we generate *isn't* the same as using the `dagger.io` one.

The main difference at the moment is the `client.go` - the `sdk/go/client.go` has automatic provisioning. While the one that's generated for modules is more lightweight, and expects to be used in the context of `dagger run`.

This does make it harder for someone to pick up if they're using our upstream library - because now they *need* to switch to using `dagger run`, they can't use `dagger.Connect` as they were before (the one generated for modules doesn't have any parameters), they can't use the `/dag` subpackage for a module.

So instead of doing this change, we should instead copy in `sdk/go/client.go` (and everything else it needs, like the `engineconn`). Should *hopefully* be a pretty simple change.

core/schema/module.go
```diff
@@ -836,16 +841,48 @@ func (s *moduleSchema) moduleInitialize(
...
 	}
 	return mod, nil
 }
 
+func (s *moduleSchema) moduleGenerateClient(
+	ctx context.Context,
+	mod *core.Module,
+	args struct {
+		Generator dagql.String
+		LocalSdk dagql.Boolean
```
jedevc, line 861: This should be an optional?

jedevc (commented):
Skimmed through, left some high level comments - some structural, but also some around the design and feel of the feature which we should definitely resolve now!

We can always come back later and fix the backend, but the CLI design is a bit stickier, so worth making sure we like it!

cmd/dagger/client_add.go
charjr: I agree with this entirely. I'm a novice when it comes to Dagger CLI but that does mean I can offer a novice's opinion:

When my module is set up, anything at all seems out of date or missing, I always hit `dagger develop`. 

If that doesn't work, I have to start digging.

cmd/codegen/generator/go/templates/src/_dagger.gen.go/defs.go.tmpl
TomChv: > So instead of doing this change, we should instead copy in sdk/go/client.go (and everything else it needs, like the engineconn). Should hopefully be a pretty simple change.

At first, this is what I wanted to do but the problem is that engineconn is internal, so I would need to make it public so it's usable by the published client, is that okay?

core/schema/module.go
TomChv: You're right! It's a missing fix while I was debugging the schema generation that was broken

cmd/dagger/client_add.go
TomChv: > Don't really know, but the generated clients should end up in the dagger.json right so we can easily regen them? cc @shykes I know you had ideas around the CLI here.

You may have multiple clients at some points, but I agree, having a `client` section in the daggerJSON make sense

```
{
  "client": [
     {
         "generator": "go"
         "outputPath": "./dagger" 
     }
  ]
}
```

cmd/codegen/generator/go/templates/src/_dagger.gen.go/defs.go.tmpl
jedevc: Yeah I'm okay with that :tada:

sdk/java/dagger-codegen-maven-plugin/src/main/resources/schemas/schema-0.15.4-010101000000-dev-8b9c9566abfc.json
TomChv: So how the java SDK is generated?

cmd/codegen/generator/go/generator.go
TomChv: Thanks @helderco, I moved the loading logic inside the withEngine function so now it works for both dagger run and simple run. I added tests for that.

shykes:
@TomChv I am very excited for this :)

A few questions:

1. In the PR description, can you add some details about the client generator interface? What interface do I need to implement in a generator

2. I notice you don't pass a target path in `client add`. Should we make that mandatory? It would make it more clear what is going on, if the command always says "add a client HERE". Also it gives us a unique identifier for each generated client, which is useful for listing, removing, and updating. wdyt?

TomChv:
> I notice you don't pass a target path in client add. Should we make that mandatory? It would make it more clear what is going on, if the command always says "add a client HERE". Also it gives us a unique identifier for each generated client, which is useful for listing, removing, and updating. wdyt?

Yeah, with a default to ./dagger, that's doable :)
Will do that tomorrow.

> Will it be possible to add self-calls? Does it require engine plumbing to be possible?

This is already supported and tested ;) but it doesn't solve the case of https://github.com/dagger/dagger/issues/8030  we would need to keep the same module object

shykes:
@jedevc @TomChv @helderco I'm going to move the UX & API design discussion for this in an issue. From experience, discussing the spec in the PR is not ideal. Harder to find the history later, and hard to follow everything that's going on.

shykes:
@jedevc @helderco @TomChv first pass: https://github.com/dagger/dagger/issues/9582

Let's discuss the UI part there please 🙏 

cc @charjr would be interested in your thoughts also

shykes:
> > Will it be possible to add self-calls? Does it require engine plumbing to be possible?
> 
> This is already supported and tested ;) but it doesn't solve the case of #8030 we would need to keep the same module object for that

Those two sentences seem to contradict each other? #8030 *is* self-calls, so what is "supported and tested"? Or am I missing something?

TomChv:
> Those two sentences seem to contradict each other? https://github.com/dagger/dagger/issues/8030 is self-calls, so what is "supported and tested"? Or am I missing something?

The bindings are generated so you can self call your module, e.g, if you have source in your module, the client add will also generated bindings to call this module (so it's self call bindings)
But when it's called from outside the module context, it just acts as if it was a dependency, so the bindings works, but we need to do some work for it to be usable inside the module itself.

wingyplus:
👋 want to drop some output issue. I try implementing the Elixir generator (https://github.com/wingyplus/elixir-generator), the generator exports `dagger_sdk` directory to the host but the cli said that the client generated at `./dagger`.


```
$ dagger cli
ent add --generator github.com/wingyplus/elixir-generator@main
✔ connect 0.2s
│ ✔ starting engine 0.0s
│ ✔ connecting to engine 0.1s
│ ✔ starting session 0.1s

✔ load client generator module 1.0s
│ ✔ finding module configuration 0.2s
│ ✔ initializing client generator module 0.8s
│ 
│ ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.1s
│ ○ .asModule: Module! 0.6s
│ ✔ .initialize: Module! 0.0s
│ ✔ .serve: Void 0.0s
│ 
│ ✔ Module.dependencies: [Module!]! 0.0s
│ 
│ ✔ Module.name: String! 0.0s

✔ Module.generateClient(generator: "github.com/wingyplus/elixir-generator@main"): Directory! 20.7s
✔ .export(path: "."): String! 0.1s
│ ✔ export directory / to host . 0.1s

Generated client at ./dagger
$ ls
Permissions Size Blocksize User      Date Modified Name
drwxr-xr-x     -         - wingyplus 17 Feb 01:48  _build
.rw-r--r--   241      4.1k wingyplus 17 Feb 01:26  dagger.json
drwxr-xr-x     -         - wingyplus 17 Feb 01:56  dagger_sdk # SDK is here.
drwxr-xr-x     -         - wingyplus 17 Feb 01:48  deps
drwxr-xr-x     -         - wingyplus 17 Feb 01:48  lib
.rw-r--r--   479      4.1k wingyplus 17 Feb 01:44  mix.exs
.rw-r--r--   487      4.1k wingyplus 17 Feb 01:48  mix.lock
.rw-r--r--   461      4.1k wingyplus 17 Feb 01:23  README.md
drwxr-xr-x     -         - wingyplus 17 Feb 01:23  test
```

TomChv:
> 👋 want to drop some output issue (not a blocker, just a confusion issue). I try implementing the Elixir generator ([wingyplus/elixir-generator](https://github.com/wingyplus/elixir-generator)), the generator exports `dagger_sdk` directory to the host but the cli said that the client generated at `./dagger`.
> 
> ```
> $ dagger cli
> ent add --generator github.com/wingyplus/elixir-generator@main
> ✔ connect 0.2s
> │ ✔ starting engine 0.0s
> │ ✔ connecting to engine 0.1s
> │ ✔ starting session 0.1s
> 
> ✔ load client generator module 1.0s
> │ ✔ finding module configuration 0.2s
> │ ✔ initializing client generator module 0.8s
> │ 
> │ ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.1s
> │ ○ .asModule: Module! 0.6s
> │ ✔ .initialize: Module! 0.0s
> │ ✔ .serve: Void 0.0s
> │ 
> │ ✔ Module.dependencies: [Module!]! 0.0s
> │ 
> │ ✔ Module.name: String! 0.0s
> 
> ✔ Module.generateClient(generator: "github.com/wingyplus/elixir-generator@main"): Directory! 20.7s
> ✔ .export(path: "."): String! 0.1s
> │ ✔ export directory / to host . 0.1s
> 
> Generated client at ./dagger
> $ ls
> Permissions Size Blocksize User      Date Modified Name
> drwxr-xr-x     -         - wingyplus 17 Feb 01:48  _build
> .rw-r--r--   241      4.1k wingyplus 17 Feb 01:26  dagger.json
> drwxr-xr-x     -         - wingyplus 17 Feb 01:56  dagger_sdk # SDK is here.
> drwxr-xr-x     -         - wingyplus 17 Feb 01:48  deps
> drwxr-xr-x     -         - wingyplus 17 Feb 01:48  lib
> .rw-r--r--   479      4.1k wingyplus 17 Feb 01:44  mix.exs
> .rw-r--r--   487      4.1k wingyplus 17 Feb 01:48  mix.lock
> .rw-r--r--   461      4.1k wingyplus 17 Feb 01:23  README.md
> drwxr-xr-x     -         - wingyplus 17 Feb 01:23  test
> ```

I didn't finished the export path feature, that's why it's confusing for now, I'll add a parameter to the interface function to specify the output directory soon :)

TomChv:
I squash all my commit to simplify the rebase on #9505, now I'm going to start the rebase, wish me luck 🚀

TomChv:
Failure on test everything else isn't related to this PR: https://v3.dagger.cloud/dagger/traces/e3aec1efb6618d36d5a2fe2b6807c641?listen=bdc294a9d8aa54cc&span=5cb3827b70f01851

Same for scan engine that is a flake like we all know :)

sipsma:
> Failure on test everything else isn't related to this PR: https://v3.dagger.cloud/dagger/traces/e3aec1efb6618d36d5a2fe2b6807c641?listen=bdc294a9d8aa54cc&span=5cb3827b70f01851

The failure in there appears to be:
```
failed to get required client generation files: failed to get required client generation files: select: Select: Generator has no such field: "requiredClientGenerationFiles"
```

Which doesn't seem like a flake, just FYI

cmd/dagger/engine.go
```diff
@@ -60,10 +64,82 @@ func withEngine(
...
+		}
+
 		return fn(ctx, sess)
 	})
 }
 
+// doDaggerConfigExist searches for dagger.json recursively in parent directories until
+// it finds a dagger.json file, reaches the context directory or reaches the root directory.
+// This is very similar to the findUp logic of the engine but it only looks for the
+// file existence prior to loading the module to not create unnecessary operation on the
```
sipsma, line 107: I wouldn't worry about unnecessary operations, it's fast enough to not be a bottleneck. We don't want to duplicate logic like this because it makes this whole system much harder to understand and maintain.

We should use the engine API for this, if there's anything about it that needs to be adjusted let me know and we can figure it out.

core/schema/module.go
```diff
@@ -651,6 +657,67 @@ func (s *moduleSchema) currentModuleWorkdirFile(
...
+	// Remove the definitions from the module so they does not conflict with its self dependency
+	mod = mod.CloneWithoutDefs()
+
+	generatedClientDir, err := generator.GenerateClient(
+		ctx,
+		source,
+		mod.Deps,
+		args.OutputDir.String(),
+		args.LocalSdk.GetOr(false).Bool(),
+	)
```
sipsma, line 714: Is this all necessary? All you need is the schema json file in `generator.GenerateClient`, so why can't you just do something like
```suggestion
	generatedClientDir, err := generator.GenerateClient(
		ctx,
		source,
		mod.Deps.Append(mod),
		args.OutputDir.String(),
		args.LocalSdk.GetOr(false).Bool(),
	)
```

core/schema/modulesource.go
```diff
@@ -2004,6 +2004,18 @@ func (s *moduleSourceSchema) moduleSourceAsModule(
 	}
 	srcInstContentHashed := src.WithMetadata(digest.Digest(src.Self.Digest), true)
 
+	// If no SDK is set, then we don't need to get the module runtime nor codegen
+	if src.Self.SDK == nil {
```
sipsma, line 2008: What case results in us needing a `Module` that doesn't have an SDK?

sipsma (commented):
Just some high-level comments+questions on the implementation, I'll look more at the UX-y stuff next

TomChv:
> The failure in there appears to be:

Good catch, I pushed a fix

core/schema/modulesource.go
TomChv: In case your setup doesn't implement an actual module but just gather dependencies, for example:

```shell
dagger init

dagger install ./my-dep

dagger client add --generator=xxx ....
```

So you end up with a dagger module, that doesn't have any source or configured SDK but you want to generate a client to call its deps :P

core/schema/module.go
TomChv: > Is this all necessary? All you need is the schema json file in generator.GenerateClient, so why can't you just do something like

Because as explained in the comment, if you simply add the current module as a dep, the typedef registered by the module will conflict with the same typedef installed as dep.
So to avoid that, I clone the mod first to not impact the original state and remove its current loaded typedef so they can be load as dep.

cmd/dagger/engine.go
TomChv: Yeah there's an explanation why, in case the configuration does not exist, when used with Dagger session, we do not want to load the source root dir: https://github.com/dagger/dagger/blob/c1468fcd4e0a8d011696d29ddfe7b6437a120e24/core/schema/modulesource.go#L341

So I could add an extra argument in `moduleSource` like `skipLoadRootDirIfConfigMissing` but I thought thought it was a bit too big and specialized for that use case so I added this middleware func to avoid that.
Let me know what your thought

cmd/codegen/generator/generator.go
```diff
@@ -50,11 +50,23 @@ type Config struct {
 	// where a pre-existing go.mod file is checked during dagger init for whether its module
 	// name is the expected value.
 	IsInit bool
+
+	// ClientOnly indicates that the codegen should only generate the client code.
+	ClientOnly bool
+
+	// LocalSDK indicates that the codegen should use the local SDK instead of the published one.
```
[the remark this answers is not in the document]
TomChv: It's used exclusively for integration tests / local testing because the generated client will depends on the published library but if you're changing the library itself, you need a way to use the latest local changes during tests and that's exactly what localSDK enable.

For example in that PR, I now export `Context` in the TS SDK `index.ts` because it's used by the generated client.
But this change is not published in the library so to make sure it work, I bundle the local SDK to the generated client.

```ts
// from tsconfig updator
  if (localSDK) {
    defaultTsConfig.compilerOptions.paths[daggerPathAlias] = [daggerPath] // @dagger.io/dagger -> ./sdk instead of the published lib
  }
```
Same for Golang, if `localSDK` is set, it will use the local go library instead of the published one.

This feature is targeting our maintainers, since I expect users to always use a released version of Dagger, they should always use the published version and never bundle the local SDK :)

cmd/dagger/engine.go
sipsma: That behavior was fixed here: https://github.com/dagger/dagger/pull/9659 so shouldn't be an issue anymore

core/schema/module.go
sipsma: But the line `mod = mod.CloneWithoutDefs()` seems like it's a no-op. You are changing the `mod` var but then just passing `mod.Deps`. I don't see how `mod = mod.CloneWithoutDefs()` could possibly be doing anything

sipsma: I pulled this PR down and ran `TestClientGenerator` with the suggested diff above and they all passed

core/schema/modulesource.go
sipsma: I'd probably just make `moduleGenerateClient` an API on `ModuleSource` rather than `Module` then. My concern is that right now there's a simple rule that "`Module` is initialized when it has an SDK and runtime" but this change makes it much more complicated which makes it all much more confusing to understand and think through

sipsma:
Trying it out locally, super cool!

One thing I noticed so far

---
```
go mod init dagger.io/demo

dagger init

dagger install github.com/shykes/hello

cat >main.go <<EOF
package main
func main() {}
EOF

dagger client add --generator=go .
```

The above results in `dagger.gen.go` starting with `package dagger`, which doesn't work because I already put a `main.go` with `package main` in there.

I think we should ideally try to match the existing package name in `dagger.gen.go`.

I'm not 100% sure how easy/difficult that is. If it's extremely difficult I wouldn't consider it a hard blocker for merging this first PR.

---

cmd/dagger/engine.go
TomChv: True! Fixed in [6236eae](https://github.com/dagger/dagger/pull/9531/commits/6236eae1f741f803eb29548448d9444773fc7682)

TomChv:
> The above results in dagger.gen.go starting with package dagger, which doesn't work because I already put a main.go with package main in there.

> I think we should ideally try to match the existing package name in dagger.gen.go.

> I'm not 100% sure how easy/difficult that is. If it's extremely difficult I wouldn't consider it a hard blocker for merging this first PR.

I think that's doable for main but it will be hard to get a bullet proof resolution, because a user may be in a subdirectory when generating. I can add a special case for now that default to `main` if the outputDir is `.`, and then we assume that the user pwd is the root directory of the module.

core/integration/client_generator_test.go
```diff
@@ -0,0 +1,727 @@
...
+			},
+		}
+
+		for _, tc := range testCases {
+			tc := tc
+
+			t.Run(tc.generator, func(ctx context.Context, t *testctx.T) {
+				c := connect(ctx, t)
+
+				devEngine := devEngineContainerAsService(devEngineContainer(c))
```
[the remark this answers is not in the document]
sipsma: Okay yeah I can see why it's tricky; when you run a nested exec it the SDKs don't need the CLI at all, but your logic requires the CLI be involved. This isn't a real-world issue so not worth changing how the actual code works but I am trying something in the tests to see if we can avoid the slowness of the nested engines. I'll push a fix to this PR if it works out since this really has nothing to do with what you're actually working on.

sipsma:
@TomChv I pushed a fix for [tests needing a separate dev engine](https://github.com/dagger/dagger/pull/9531#discussion_r1970439838) in [this commit](https://github.com/dagger/dagger/pull/9531/commits/e5226cc360be39e98d123b344a4ef605123bf176).

There are some failing integ tests but it looks like they were already failing before that commit, `TestClientGenerator/TestOutputDir/nested_directory_(generated/nested/test)/go`

cmd/dagger/client_add.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+type clientAddHandler struct {
+	dag        *dagger.Client
+	cmd        *cobra.Command
+	outputPath string
+}
+
+func (c *clientAddHandler) Run(ctx context.Context) (rerr error) {
+	mod, _, err := initializeClientGeneratorModule(ctx, c.dag, ".")
```
sipsma, line 73: The fact that we only load `.` for the module means this code will only work if you run it with the source code, right? That might be okay for typescipt/python, but since Go is compiled that's sort of a big/surprising limitation.

I'm totally fine with merging this PR with that limitation but we should fix that in a follow-up I think. For git deps specifically this should be easy, just need to embed the refs of the defs to load in the compiled bin. It's more complicated for local deps, but could figure something out there too.

TomChv: I'm not sure I understand this comment, because there's a find up logic in the engine, I would assume that the whole ctx dir can be accessible?
Do you have an example to share so I can have a better understanding

core/schema/modulesource.go
TomChv: Make sense, but then how would it work for dependency serving when running dagger session?
I need to serve the dependencies and if it's available the current module.

```go
		if params.ServeModule {
			mod, exist, err := initializeClientGeneratorModule(ctx, sess.Dagger(), ".")
			if err != nil && !errors.Is(err, ErrConfigNotFound) {
				return fmt.Errorf("failed to initialize current module: %w", err)
			}

			if exist {
				for _, dep := range mod.Dependencies {
					err := dep.AsModule().Serve(ctx)
					if err != nil {
						return fmt.Errorf("failed to serve dependency %w", err)
					}
				}
			}
		}
```

I think I need to change some logic there to:
- Move `moduleGenerateClient` in `ModuleSource` like you mentionned
- Add a `dependencies` method in `ModuleSource` to returns all deps of the module
- Add `sdk` method in `ModuleSource` so if an SDK is set, we can call `AsModule` on the source and set it.

What do you think of that plan?

sipsma: > Move moduleGenerateClient in ModuleSource like you mentionned

👍

> Add a dependencies method in ModuleSource to returns all deps of the module so they can be served

That already exists and is the one you are using already: https://github.com/TomChv/dagger/blob/e45ebd5f4222708ca1db05ff34a68320e24b9ed4/cmd/dagger/module_inspect.go#L140

> Add sdk method in ModuleSource so if an SDK is set, we can call AsModule on the source and serve it.

I'm fine with that generally speaking but am missing where that would be needed? AFAICT you wouldn't need to change anything in the client-side code for this. If it is needed for whatever reason though then that sgtm

cmd/dagger/client_add.go
sipsma: In all your existing examples for `go`, you are running it with `go run main.go`. That will only work if `go run main.go` is run in the same directory that has the module source.

It should be possible to do `go build -o binary main.go`, move `binary` anywhere and then execute `./binary` and everything still works.

The compiled binary just needs to have the list of deps to load embedded in it. Then it should work seamlessly for git deps. Local deps are more complicated like I said but various options possible when we get there.

core/schema/modulesource.go
TomChv: Fixed in [e3d4f80](https://github.com/dagger/dagger/pull/9531/commits/e3d4f80722e913be991b928d61f6766a79c73d50), I didn't had to add a SDK method, so it's all good :D

cmd/dagger/client_add.go
TomChv: Ohhh I see, yeah you're totally right! the binary needs to have access to the dagger.json and local deps, I'm not sure how this would work

TomChv:
Current failing tests are non related to that PR, I'll add the changes around dagger.json in a follow up PR so we can iterate faster

sipsma:
> Current failing tests are non related to that PR, I'll add the changes around dagger.json in a follow up PR so we can iterate faster

@TomChv it looks like quite a few are failures in `TestClientGenerator` that also fail locally. A lot of errors along the lines of
```
# test.com/test/generated
generated/dagger.gen.go:21:2: querybuilder redeclared in this block
        generated/dagger.gen.go:19:2: other declaration of querybuilder
generated/dagger.gen.go:21:2: "dagger.io/dagger/querybuilder" imported and not used
```

TomChv:
> # test.com/test/generated
generated/dagger.gen.go:21:2: querybuilder redeclared in this block
        generated/dagger.gen.go:19:2: other declaration of querybuilder
generated/dagger.gen.go:21:2: "dagger.io/dagger/querybuilder" imported and not used

Oh this is a related to my latest change, I know why

cmd/codegen/codegen.go
```diff
@@ -102,5 +102,9 @@ func generate(ctx context.Context, introspectionSchema *introspection.Schema, in
 		return nil, fmt.Errorf("use target SDK language: %s: %w", sdks, generator.ErrUnknownSDKLang)
 	}
 
-	return gen.Generate(ctx, introspectionSchema, introspectionSchemaVersion)
+	if cfg.ClientOnly {
+		return gen.GenerateClient(ctx, introspectionSchema, introspectionSchemaVersion)
+	}
```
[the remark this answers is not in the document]
TomChv: Because a client generated with `clientOnly` is a bit different than just the regular core client and also different than a module client.
The logic before between module & client generation was mixed into the old `Generate` function.
Because I didn't wanted to risk to break the current behaviour, I just added that option to safely change the logic of the generateClient function.
The generation of the core client may depends on some logic of the GenerateModule, I'm pretty sure this could be simplified too but not worth it yet I think.

I plan to work on the codegen side later, splitting the client & module gen into 2 subcommands, so it will be a good opportunity to fix that logic.

cmd/codegen/generator/go/templates/src/_dagger.gen.go/defs.go.tmpl
```diff
@@ -123,7 +132,7 @@ func (e *ExecError) Unwrap() error {
 {{ if eq .Kind "ENUM" }}{{ template "_types/enum.go.tmpl" . }}{{ end }}
 {{ end }}
 
-{{ if IsModuleCode }}
+{{ if or (IsModuleCode) }}
```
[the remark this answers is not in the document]
TomChv: Oups, good catch

cmd/codegen/generator/go/templates/src/dag/dag.gen.go.tmpl
```diff
@@ -1,4 +1,4 @@
-{{ if not IsModuleCode }}
+{{ if and (not IsModuleCode) (not IsStandaloneClient) }}
```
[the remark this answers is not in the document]
TomChv: Because `dag` is not used anymore, so I'm not sure there's a reason to generate it?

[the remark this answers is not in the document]
jedevc: `dag` is the client that can be used without `Connect`. It definitely can still be used, it was added to make it easier to transfer code between modules and standalone.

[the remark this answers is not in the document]
TomChv: Ohh then yeah I'll add the generation for it

jedevc (approved):
After comments resolved, feel free to merge - we'll do any follow-ups later, the main things that would be good to do:
- `dagger.json` persistent configuration
- Cleanup of the commands + flags in `codegen`
- Loading deps improvement from https://github.com/dagger/dagger/pull/9531#discussion_r1970521611 as suggested by @sipsma

cmd/codegen/generator/generator.go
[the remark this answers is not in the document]
shykes: What is "the local SDK"?

PR #9532, opened 2025-02-07
sdk/java: @DefaultPath and @Ignore

eunomie:
Handle `@DefaultPath` and `@Ignore` annotations.

For instance the following module:

```java
@Object
public class Testjava extends AbstractModule {
  public Testjava() {
    super();
  }

  @Function
  public Directory dir(@DefaultPath(".") @Ignore({"**", "!*.json"}) Directory directoryArg) {
    return directoryArg;
  }
}
```

```console
$ dagger call dir -q

_type: Directory
digest: sha256:461847f561c209b33811c523376fcc839335c6a46ae1a5941781d9c33c824ef0
entries:
    - dagger.json
```

- Errors are raised if `@Default` and `@DefaultPath` are used together.
- `@DefaultPath` can only be used on `io.dagger.client.Directory` and `io.dagger.client.File`
- `@Ignore` can only be used on `io.dagger.client.Directory`


sdk/java/dagger-java-sdk/src/main/java/io/dagger/module/info/StringOptionInfo.java
```diff
@@ -0,0 +1,3 @@
+package io.dagger.module.info;
+
+public record StringOptionInfo(boolean isSet, String value) {}
```
jcsirot, line 3: why not using `Optional<String>` instead?

eunomie: Well... I don't know :-D
I'll give a try 👍

eunomie: Updated with `Optional` 👍

eunomie:
regarding `scan` failing check, see https://github.com/dagger/dagger/pull/9533

Otherwise, does that sounds good to you @jcsirot?

sdk/java/dagger-java-annotation-processor/src/test/resources/io/dagger/java/module/DaggerJava.java
```diff
@@ -37,7 +35,9 @@ public Container containerEcho(@Default("Hello Dagger") String stringArg) {
    * @return Standard output of the grep command
    */
   @Function
-  public String grepDir(Directory directoryArg, @Nullable String pattern)
+  public String grepDir(
+      @DefaultPath("sdk/java") @Ignore({"**", "!*.java"}) Directory directoryArg,
+      @Nullable String pattern)
```
helderco, line 40: Could `java.util.Optional` be used instead of `@Nullable`?

Like:
```diff
-public String grepDir(Directory directoryArg, @Nullable String pattern)
+public String grepDir(Directory directoryArg, Optional<String> pattern)
```
Or:
```diff
-public String nullableDefault(@Nullable @Default("Foo") String stringArg) {
+public String nullableDefault(@Default("Foo") Optional<String> stringArg) {
```

eunomie: That's a good question. But it makes things a little bit more complexe, especially as `Optional` is not a serializable type. This doesn't mean we can't do it, but it requires more type manipulation.

I'll investigate a bit more, but in a different PR so this doesn't block this one that is on the `@DefaultPath`/`@Ignore`.

jcsirot: @helderco agree, I think we should try to stick as closely as possible to the language's idioms.
@eunomie This will require some minor changes in how module calls are handled in the entry point code, but nothing too complex at first glance. That said, we can definitely address this in another PR.

eunomie:
I've publish a commit that contains tests here: https://github.com/lgtdio/dagger/commit/89398919e26ea5223dc9706154d9be4a25306545
I'll open a PR with that right after this one is merge.
For now I'm not able to have it work by referencing a local sdk. I might be wrong somewhere, but not sure why.
But once this PR is merged it will then not be a problem to have the tests as the sdk will be available.

I'll ensure the CI passes (as much as possible) but in the meantime if @jcsirot you can review the changes so it can be merged it would be super great 🙏

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -616,4 +673,15 @@ private String parseParameterDescription(Element element, String paramName) {
             .findFirst();
     return blockTag.map(tag -> tag.getContent().toText()).orElse("");
   }
+
+  private static CodeBlock stringArrayToList(String[] array) {
+    CodeBlock.Builder code = CodeBlock.builder().add("$T.of(", List.class);
+    for (int i = 0; i < array.length; i++) {
```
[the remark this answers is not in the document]
eunomie: 👍 

After so many years in Go, everything looks like a `for` loop 😅

sdk/java/dagger-java-annotation-processor/src/test/resources/io/dagger/java/module/DaggerJava.java
helderco: Of course, wasn't suggesting it would be in this PR 👍

helderco:
>  For now I'm not able to have it work by referencing a local sdk. I might be wrong somewhere, but not sure why. But once this PR is merged it will then not be a problem to have the tests as the sdk will be available.

It's worth investing on making that work instead of pushing that can for later. You can ask for help if you're having difficulty. It's much better to keep the test in the same PR to help validate the feature, rather than a follow up PR.

eunomie:
> It's worth investing on making that work instead of pushing that can for later. You can ask for help if you're having difficulty. It's much better to keep the test in the same PR to help validate the feature, rather than a follow up PR.

Done, I was finally able to fix the tests :-)
So the annotations are now covered 👍

jcsirot (commented):
LGTM

helderco:
I'm looking into it.

helderco (approved):
LGTM 👍

PR #9554, opened 2025-02-10
fix: avoid importing buildkit to the CLI

jedevc:
This was bad, since the CLI is built for Windows, but our buildkit package uses Unix constants.

We can just move the offending environment variable names into `engine`, which is safe to import from the CLI.


vito (approved):
Woops, thanks for the fix! 🙏

PR #9555, opened 2025-02-10
Cleanup mage and simplify building a dev engine

jedevc:
This PR reworks two main things about our dev setups:
- It now builds a CLI that is linked to the started engine - it should now be possible to run the built CLI standalone, *without* needing to wrap it in a `with-dev` call.
- It purges mage entirely from our build setup - which we've wanted to do for *ages*. We can take all the complexity from there, and move it into `./hack/build` which is a nice little dagger shell script :tada: (which should also help us test and work on shell!)


hack/build
```diff
@@ -0,0 +1,14 @@
...
+
+# HACK: strip "build" from the script path to get the parent module
+.use ${0%/build}/..
+
+PLATFORM=current
+CONTAINER=dagger-engine.dev
+
+cli | binary --runner-host=docker-container://$CONTAINER --platform=$PLATFORM | export ./bin/dagger &
+engine | container --platform=$PLATFORM | as-tarball --forced-compression=gzip | export ./bin/engine.tar &
+engine | load-to-docker /var/run/docker.sock --platform=$PLATFORM | start --name=$CONTAINER &
```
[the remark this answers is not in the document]
cwlbraa: also is `&` the same `&` in posix shell scripts? like we're backgrounding 3 "processes" here?

cmd/dagger/engine.go
```diff
@@ -2,19 +2,56 @@ package main
...
+
+var (
+	// RunnerHost holds the host to connect to.
+	//
+	// Note: this is filled at link-time.
+	RunnerHost string
+)
+
+func init() {
+	if RunnerHost == "" {
```
jedevc, line 33: Note to self: env var should have most precedence.

hack/build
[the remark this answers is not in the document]
jedevc: Ooh yup, we can do multi-line pipes here.

And yes, `&` is a background process - then `_wait` waits for them all to finish nicely.

hack/dev
```diff
@@ -1,12 +1,11 @@
 #!/usr/bin/env bash
 
-set -e -u
+# NOTE: this script is kept for legacy reasons - you should just be able to
+# directly run the binaries produced by hack/build.
```
cwlbraa, line 4: it's really not legacy, it lets you run other shit with all the appropriate envs set... running `go test` is a huge pain without this.

hack/with-dev
```diff
@@ -1,12 +1,19 @@
...
 export DAGGER_SRC_ROOT="$(cd $(dirname $(realpath "${BASH_SOURCE[0]}"))/.. && pwd)"
-export MAGEDIR="$DAGGER_SRC_ROOT/.dagger/mage"
+export DAGGER_BIN_ROOT="${DAGGER_SRC_ROOT}/bin"
 
-pushd $MAGEDIR > /dev/null
-eval $(go run main.go -w $DAGGER_SRC_ROOT engine:devenv)
-popd > /dev/null
+export _EXPERIMENTAL_DAGGER_CLI_BIN=$DAGGER_BIN_ROOT/dagger
+if [[ "$(uname -s)" != "Linux" ]]; then
+    export _EXPERIMENTAL_DAGGER_CLI_LINUX_BIN=$DAGGER_BIN_ROOT/dagger-linux
```
cwlbraa, line 13:
```suggestion
    export _TEST_DAGGER_CLI_LINUX_BIN=$DAGGER_BIN_ROOT/dagger-linux
```

ftfy :)

hack/build
```diff
@@ -0,0 +1,27 @@
...
+cli |\
+    dev-binaries --runner-host=docker-container://$CONTAINER --platform=current |\
+    export ./bin &
+
+engine |
+    container ${_EXPERIMENTAL_DAGGER_GPU_SUPPORT:+--gpu-support} |\
+    as-tarball --forced-compression=gzip |\
+    export ./bin/engine.tar &
+engine |\
+    load-to-docker ${DOCKER_HOST:-/var/run/docker.sock} \
```
cwlbraa, line 18:
```suggestion
    load-to-docker ${DOCKER_HOST:-$(docker context inspect | jq -r '.[0].Endpoints.docker.Host')} \
```

/var/run/docker.sock is fairly linux-specific i think? definitely doesn't work by default with colima.

cwlbraa: i think this works too and doesn't require jq:

```suggestion
    load-to-docker ${DOCKER_HOST:-$(docker context inspect -f '{{.Endpoints.docker.Host}}')} \
```

jedevc: Yeah, we can't put bash commands in dagger shell - intentionally, it's intended that there's no way to escape the sandbox :cry:

I can put this logic into `./hack/dev` though.

jedevc: This *feels* related to https://github.com/dagger/dagger/issues/8025

jedevc: huh @helderco this looks like a bug :eyes:

vito (approved):
lgtm! Nice trick having the CLI bake in the RunnerHost

PR #9558, opened 2025-02-11
feat: inject all sdk.envs to all sdks

jedevc:
Previously, we were filtering these, and limiting them to *just* the go sdk - we should inject them everywhere we can (for consistency).


jedevc:
Closing this, we're no longer taking this approach - we're going with the approach in #9323.

PR #9559, opened 2025-02-11
chore: revert sdk.envs field

jedevc:
Follow-up to, and partial reversion of https://github.com/dagger/dagger/pull/9454.

Reverting this temporarily, we still need more design discussion around this, there's still some things to work out. But ideally, we shouldn't block the release.

We keep the struct-ification of the `sdk` field though, that's useful, since we intend to extend that with `pin` and friend in the future: see https://github.com/dagger/dagger/issues/9156.


jedevc:
Shouldn't need to? No details of the `env` field were added to the API, so the only changes from https://github.com/dagger/dagger/pull/9454 were for the struct-ification.

marcosnils (approved):

PR #9569, opened 2025-02-12
Improve releasing during v0.15.4

jedevc:


gerhard:
There seem to be a bunch of linting issues that we should fix before merging. FWIW: https://github.com/dagger/dagger/pull/9474#pullrequestreview-2583908579

Not sure why Typescript failed twice in a row... We have similar issues with Python last time: https://github.com/dagger/dagger/pull/9474#issuecomment-2624812501

Running it locally to see if it's only an issue remotely.

gerhard:
`dagger call check --targets=sdk/typescript` passed locally the first time using a clean 0.15.4 Engine, this must be a remote flake.

gerhard (approved):
I am approving this optimistically since we suspect that switching to 0.15.4 will make our runs more reliable.

Before merging please:
- fix the failing lint check
- ensure that Typescript goes green

🚀

jedevc:
I'm so confused, I have no idea what's going on with the typescript job. It seems to do *exactly* the same thing as python last week. Just run and timeout, while still actually doing work.

It does seem to be taking a *really* long time to build modules, but... why specifically here?

jedevc:
Noticing that something is odd in our logs here, maybe related? See https://github.com/dagger/dagger/actions/runs/13307440639/job/37172643790, and follow operation 2320. 
image

This completes quickly, but it takes a moment for all the logs to eventually appear. For example:
image

This could just be a weird quirk of the plain progress output? But also, this is really odd, once a component can start printing logs, it'll print *all* available logs, so this implies they're being streamed in very small chunks to the client?

vito:
@jedevc A guess: it's technically correct that the `withExec` call itself is `DONE` in the sense that the `withExec` API call completed - however the actual `exec` hasn't started, so that's why the logs come in later.

Maybe the plain frontend needs to check for the new "pending" state and avoid printing `DONE` while that's the case.

jedevc:
Great, bump the runtime length, and it passes immediately in 4m. :stuck_out_tongue:

PR #9575, opened 2025-02-13
fix: fix return list type is not slice of pointer of struct

### LINKED ISSUE #9556: 🐞 Go codegen render return type as value struct instead of pointer struct when a function return a slice
### What is the issue?

If declare a function that returns a list of type (`[]*T`, `[]*dagger.Container` for example), the Go codegen will render as `[]T` instead of `[]*T` that look break the convention. 

### Dagger version

dagger v0.15.3 (registry.dagger.io/engine:v0.15.3) darwin/arm64

### Steps to reproduce

Create a module that has a function like this example below:

```go
func (m *KptBuilder) Build(source *dagger.Directory) (ctrs []*dagger.Container) {
	// ...
}
```

Then, create a new module and include module above as a dependency. We will get a function like:

```go
func (r *KptBuilder) Build(ctx context.Context, source *Directory) ([]Container, error) { // kpt-builder (../../../../modules/kpt-builder/main.go:14:1)
	assertNotNil("source", source)
	q := r.query.Select("build")
	q = q.Arg("source", source)

	q = q.Select("id")

	type build struct {
		Id ContainerID
	}

	convert := func(fields []build) []Container {
		out := []Container{}

		for i := range fields {
			val := Container{id: &fields[i].Id}
			val.query = q.Root().Select("loadContainerFromID").Arg("id", fields[i].Id)
			out = append(out, val)
		}

		return out
	}
	var response []build

	q = q.Bind(&response)

	err := q.Execute(ctx)
	if err != nil {
		return nil, err
	}

	return convert(response), nil
}
```

### Log output

_No response_

wingyplus:
Fixes #9556

Fix any methods that return list of object must return in form of `[]*T`.

The go template override `FormatReturnType` to make it works and expose `FormatType` from common functions type make it accessible from `FormatReturnType`.

### DIFF
--- a/cmd/codegen/generator/go/templates/functions.go
+++ b/cmd/codegen/generator/go/templates/functions.go
@@ -52,7 +52,6 @@ type goTemplateFuncs struct {
 func (funcs goTemplateFuncs) FuncMap() template.FuncMap {
 	return template.FuncMap{
 		// common
-		"FormatReturnType":          funcs.FormatReturnType,
 		"FormatInputType":           funcs.FormatInputType,
 		"FormatOutputType":          funcs.FormatOutputType,
 		"GetArrayField":             funcs.GetArrayField,
@@ -72,6 +71,7 @@ func (funcs goTemplateFuncs) FuncMap() template.FuncMap {
 		"FormatExperimental":      funcs.formatExperimental,
 		"FormatName":              formatName,
 		"FormatEnum":              funcs.formatEnum,
+		"FormatReturnType":        funcs.FormatReturnType,
 		"SortEnumFields":          funcs.sortEnumFields,
 		"FieldOptionsStructName":  funcs.fieldOptionsStructName,
 		"FieldFunction":           funcs.fieldFunction,
@@ -89,6 +89,29 @@ func (funcs goTemplateFuncs) FuncMap() template.FuncMap {
 		"ModuleRelPath":           funcs.moduleRelPath,
 		"GitDependencies":         funcs.gitDependencies,
 	}
+}
+
+func (funcs goTemplateFuncs) FormatReturnType(f introspection.Field, scopes ...string) (representation string, err error) {
+	scope := strings.Join(scopes, "")
+	ff := funcs.CommonFunctions.FormatTypeFuncs.WithScope(scope)
+
+	for ref := f.TypeRef; ref != nil; ref = ref.OfType {
+		switch ref.Kind {
+		case introspection.TypeKindNonNull:
+			continue
+		case introspection.TypeKindList:
+			defer func() {
+				representation = ff.FormatKindList(representation)
+			}()
+		case introspection.TypeKindObject:
+			representation = ff.FormatKindObject(representation, ref.Name, funcs.CommonFunctions.ConvertID(f))
+			representation = "*" + representation
+			return
+		default:
+			return funcs.CommonFunctions.FormatType(ref, scope, funcs.CommonFunctions.ConvertID(f))
+		}
+	}
+	return "", fmt.Errorf("unexpected type kind %s", f.TypeRef.Kind)
 }
 
 // comments out a string
@@ -293,8 +316,6 @@ func (funcs goTemplateFuncs) fieldFunction(f introspection.Field, topLevel bool,
 		retType = "error"
 	case f.TypeRef.IsScalar() || f.TypeRef.IsList():
 		retType = fmt.Sprintf("(%s, error)", retType)
-	default:
-		retType = "*" + retType
 	}
 	signature += " " + retType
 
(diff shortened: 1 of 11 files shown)


cmd/codegen/generator/go/templates/functions.go
```diff
@@ -89,6 +89,29 @@ func (funcs goTemplateFuncs) FuncMap() template.FuncMap {
...
+		case introspection.TypeKindObject:
+			representation = ff.FormatKindObject(representation, ref.Name, funcs.CommonFunctions.ConvertID(f))
+			representation = "*" + representation
+			return
+		default:
+			return funcs.CommonFunctions.FormatType(ref, scope, funcs.CommonFunctions.ConvertID(f))
+		}
+	}
+	return "", fmt.Errorf("unexpected type kind %s", f.TypeRef.Kind)
+}
```
wingyplus, line 113: This is quite a bit hack. Feel free to suggest if anyone has a better solution. 🙏

TomChv: I think it's fine that way.

A big refactor on how the codegen works will happen at some point and that will be a good time to fix all these small magic tricks.

TomChv (commented):
Looks fine IMO, just please add a changie notification since it may impacts users.

jedevc:
Ideally we should version-compat this change - it has the potential to break published modules, so should have a check here.

wingyplus:
> Ideally we should version-compat this change - it has the potential to break published modules, so should have a check here.

I’ve never dealing with a version compat before. How to do it?

TomChv:
@wingyplus Any news on this PR? I'm ready to approve whenever you address changes :D

wingyplus:
> @wingyplus see https://github.com/dagger/dagger/pull/9118/files#diff-8062ad596c2b352a00e64f30eb9fff7ade878660570bf2c3c5f5aa727dd10496R864 for an example of where I've done this recently.
> 
> Generally:
> 
> * If a core function signature changes, you use [dagql views](https://github.com/dagger/dagger/blob/e2be4eaedc36f1c65efcff760cae4d596e323b82/core/schema/container.go#L335-L336)
> * If you're just changing behavior somewhere, you can just directly call `CheckVersionCompatibility` with the `Query`.

May I know what scenario I should check here? My understanding is that the codegen binary is built into the dagger engine so it should fallback to previous behavior if we switch the binary back to previous version.

wingyplus:
> @wingyplus It seems there's not conflict, please regenerate the dagger.gen.go client and then I can review :D

Will fix in around 2-3 hours

wingyplus:
@TomChv sorry for the delay. I rebased and squashed all the commits. 🙏

sipsma (commented):
Thanks for tackling the fix for this, I hit it yesterday and was happily surprised to see it was already being worked on. The changes here lgtm.

I agree w/ @jedevc that trying to make this backwards-compatible is important though. The reason being that there's plenty of user module code that will start failing to compile after this is fixed, which is extra bad when it's some external dependency module you as a user don't directly control and thus can't fix yourself.

As painful as it would be, I'd ideally want us to maintain the old behavior for modules w/ older engine versions in their `dagger.json`. I don't recall if we currently pass that dagger.json engine version information to codegen anywhere, but if we don't I think we could plumb it through to the codegen config (i.e. [here](https://github.com/sipsma/dagger/blob/d427955ee1bb3ace8d88e981ede878c1cec75eaf/cmd/codegen/generator/generator.go#L27-L27)) and use that to only have this new behavior for engine versions above the next one.

The code to support both is going to be ugly, but can be cleaned up after there's been enough time for the ecosystem to upgrade. This seems worth it because this fix is for something extremely important that we don't want to be stuck w/ forever.

wingyplus:
> Thanks for tackling the fix for this, I hit it yesterday and was happily surprised to see it was already being worked on. The changes here lgtm.
> 
> I agree w/ @jedevc that trying to make this backwards-compatible is important though. The reason being that there's plenty of user module code that will start failing to compile after this is fixed, which is extra bad when it's some external dependency module you as a user don't directly control and thus can't fix yourself.
> 
> As painful as it would be, I'd ideally want us to maintain the old behavior for modules w/ older engine versions in their `dagger.json`. I don't recall if we currently pass that dagger.json engine version information to codegen anywhere, but if we don't I think we could plumb it through to the codegen config (i.e. [here](https://github.com/sipsma/dagger/blob/d427955ee1bb3ace8d88e981ede878c1cec75eaf/cmd/codegen/generator/generator.go#L27-L27)) and use that to only have this new behavior for engine versions above the next one.
> 
> The code to support both is going to be ugly, but can be cleaned up after there's been enough time for the ecosystem to upgrade. This seems worth it because this fix is for something extremely important that we don't want to be stuck w/ forever.

@sipsma Thank you for the feedback. Let me take a look at the details and will return in around 1-2 days. 🙏

jpadams:
Thanks for keeping this one alive @wingyplus !

Let us know what help you need to get this over the line 🙏

PR #9581, opened 2025-02-13
ci: simplify telemetry tests caller

jedevc:


vito (approved):
I'm OK with this now that I'm not running it all the time, but the original reason was to avoid some of the slow parts of `testCmd`, namely building and mounting an engine tarball, since the telemetry test suite doesn't need that. It adds up and gets pretty painful when this is part of the dev loop. I guess that could be an arg to `testCmd` or something.

Either way approving for now, agree the duplication is smelly, I can take another swing at it next time I'm in the area.

PR #9589, opened 2025-02-14
fix(php): bump graphql dependency

jedevc:
We started relying on functionality introduced in v1.14.1 of php-graphql-client. But we failed to bump to v0.14.0.

This wasn't caught during testing, but only when upgrading an older module to the latest version of dagger.


jedevc:
Since this affects upgrading *all* older PHP modules, we should definitely ship this fix in the next release.

However, a short term fix is to update the `composer.lock` of the PHP module, which will fetch the latest version of the graphql dep.

charjr (approved):
LGTM!

Thank you for bumping this.

PR #9604, opened 2025-02-17
sdk/java: check object definition

eunomie:
Ensure the java object is extending the right base class and check the default constructor exists.

### DIFF
--- a/sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
+++ b/sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
@@ -101,6 +101,21 @@ public class DaggerModuleAnnotationProcessor extends AbstractProcessor {
                 "The class %s must extend %s".formatted(qName, AbstractModule.class.getName()));
           }
 
+          boolean hasDefaultConstructor =
+              typeElement.getEnclosedElements().stream()
+                      .filter(elt -> elt.getKind() == ElementKind.CONSTRUCTOR)
+                      .map(ExecutableElement.class::cast)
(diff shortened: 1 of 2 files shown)


sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -88,6 +89,30 @@ ModuleInfo generateModuleInfo(Set<? extends TypeElement> annotations, RoundEnvir
...
+              typeElement.getEnclosedElements().stream()
+                  .filter(elt -> elt.getKind() == ElementKind.CONSTRUCTOR)
+                  .map(ExecutableElement.class::cast)
+                  .anyMatch(constructor -> constructor.getParameters().isEmpty());
+
+          if (!hasDefaultConstructor) {
+            throw new RuntimeException(
+                "The class %s must have a public no-argument constructor that calls super()"
+                    .formatted(qName));
+          }
```
helderco, line 114: Is this a language limitation?

eunomie: It's a bit tricky, and I just saw this can be improved.

If the class has one or more constructor, one **must** be a default one with no argument. That way we can instantiate the object. This has been raised while working on module constructor as we are defining a specific constructor. Basically in that case we need one that will be called with the different arguments, and one empty to instantiate the object.
And I replicated it here.

But in fact, this can be improve: if there's no constructor, then it's fine.

Basically:
- no constructor set: Java adds automatically a default no arg constructor that will call `super()` because parent class has a no arg constructor
- one or more constructor set: a default no arg constructor must exist

jcsirot (commented):
LGTM

helderco (approved):

PR #9605, opened 2025-02-17
sdk/java: replace `@Nullable` by `Optional<?>` arguments

eunomie:
This removes the home made `@Nullable` annotation and replace it by the built in `Optional`.

Before:

```java
@Function
public String echo(@Nullable String arg) {
    if (arg == null) {
      arg = "the arg was null";
    }
    return arg;
}
```

After:

```java
@Function
public String echo(Optional<String> arg) {
    return arg.orElse("the arg was null");
}
```

`Optional` also works with `@Default`:

```java
@Function
public String echo(@Default("Foo") Optional<String> arg) {
    return arg.get();
}
```

This works because we can't set a `null` value.
But when this will be possible, the Java code will work as expected. We can imagine to have `@Default("Foo") Optional<String> arg`, send a null value, and in the java code have a `arg.orElse("in case of null')`.
Values will be:
- the string set if it's the case
- `"Foo"` if no value is set
- `"in case of null"` if the value is explicitly set to `null`.


sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -239,6 +184,76 @@ ModuleInfo generateModuleInfo(Set<? extends TypeElement> annotations, RoundEnvir
...
   }
 
+  private List<ParameterInfo> parseParameters(ExecutableElement elt) {
+    return elt.getParameters().stream()
+        .map(
+            param -> {
+              TypeMirror tm = param.asType();
+              TypeKind tk = tm.getKind();
+
+              boolean isOptional = false;
```
jcsirot, line 194: I'm not a big fan of string manipulation when comparing types. Maybe this could be more robust:
```java
if (tm instanceof DeclaredType dt) {
  isOptional = dt.asElement().getQualifiedName().contentEquals(Optional.class.getCanonicalName());
  if (isOptional) {
    tm = dt.getTypeArguments().get(0);
  }
}
```

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -507,6 +435,117 @@ static JavaFile generate(ModuleInfo moduleInfo) {
...
+  private static CodeBlock functionInvoke(ObjectInfo objectInfo, FunctionInfo fnInfo) {
+    CodeBlock.Builder code = CodeBlock.builder();
+    CodeBlock fnReturnType = typeName(fnInfo.returnType());
+
+    for (var parameterInfo : fnInfo.parameters()) {
+      CodeBlock paramType = typeName(parameterInfo.type());
+
+      String defaultValue = "null";
+      TypeKind tk = getTypeKind(parameterInfo.type().kindName());
+      if (tk == TypeKind.INT
```
jcsirot, line 447: Maybe we should also add `TypeKind.SHORT` and `TypeKind.CHAR`

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
eunomie: This type system drives me crazy 🤪
But I finally have something working. Maybe not the most efficient, but working as expected.

```java
var optionalType =
    processingEnv.getElementUtils().getTypeElement(Optional.class.getName()).asType();
if (tm instanceof DeclaredType dt) {
  if (processingEnv.getTypeUtils().isSameType(dt.asElement().asType(), optionalType)) {
    isOptional = true;
    tm = dt.getTypeArguments().get(0);
    tk = tm.getKind();
  }
}
```

(will be push soon, once the base branch passes)

eunomie:
could someone from @dagger/sdk-java add a ✅ to the PR so that I can merge it? 🙏 
TIA

jedevc:
FYI @helderco, since you understand the details of optionals more than I possibly do :laugh

helderco (approved):

PR #9609, opened 2025-02-18
helm: Enable setting imagePullSecrets

vmaffet:
This PR enables setting `imagePullSecrets` making it easier to use private custom Dagger images.


vmaffet-fast:
I might have done the changie command in the wrong location 🤔

matipan (approved):
Definitely useful, thanks for sending the patch!

PR #9617, opened 2025-02-18
feat: add `name` in Directory API

### LINKED ISSUE #9522: Add a name method to Directory type

I am specifically in the go sdk, so my example is specific to that, but I assume this is not "just a change to the go sdk."  I am importing a directory into a function and for the construction of some custom types (and for logging), I'd like to use the base name of the directory that was passed in.  While I am specifically after the basename, I think the full path makes the most sense as I can do with that what I will.  That said, the File type has a name method and I'm merely looking for the same on the Directory type, so it pragmatically makes the most sense for this to behave the same for consistency's sake.

I need it for a map key and log output.  It also just seems like something that *should* be here given it exists on the File type

I'm not - I tried to export the directory to the angine and get the name from that directory (it exports the contents of the directory, not the directory itself).

TomChv:
Based on https://discord.com/channels/707636530424053791/1336739628350439454.

This PR adds a function `name` to the Directory API to retrieve the latest name of the branch.

### TODOS

- [x] API changes
- [x] Client generation
- [x] Integration tests

Fixes #9522

### DIFF
--- a/core/integration/directory_test.go
+++ b/core/integration/directory_test.go
@@ -1214,3 +1214,57 @@ func (DirectorySuite) TestDigest(ctx context.Context, t *testctx.T) {
 		require.Equal(t, "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855", digest)
 	})
 }
+
+func (DirectorySuite) TestDirectoryName(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	t.Run("empty directory name", func(ctx context.Context, t *testctx.T) {
+		name, err := c.Directory().Name(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "/", name)
+	})
+
+	t.Run("structured directory", func(ctx context.Context, t *testctx.T) {
+		dir := c.Directory().WithDirectory("nested", c.Directory()).WithDirectory("very/nested", c.Directory())
+
+		t.Run("root directory", func(ctx context.Context, t *testctx.T) {
+			rootName, err := dir.Name(ctx)
+			require.NoError(t, err)
+			require.Equal(t, "/", rootName)
+		})
+
+		t.Run("nested directory", func(ctx context.Context, t *testctx.T) {
+			nestedName, err := dir.Directory("nested").Name(ctx)
+			require.NoError(t, err)
+			require.Equal(t, "nested", nestedName)
+		})
+
+		t.Run("very nested directory", func(ctx context.Context, t *testctx.T) {
+			veryNestedName, err := dir.Directory("very/nested").Name(ctx)
+			require.NoError(t, err)
+			require.Equal(t, "nested", veryNestedName)
+		})
+	})
+
+	t.Run("git directory", func(ctx context.Context, t *testctx.T) {
+		dir := c.Git("https://github.com/dagger/dagger#ee32df913f57c876e067bd5ecc159561510b6f50").Head().Tree()
+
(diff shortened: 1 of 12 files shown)


marcosnils:
@TomChv  have a conceptual question for this PR. Shouldn't `dag.Directory().Directory("non-existent").Name(ctx)` raise an error if the directory doesn't exist? I'm raising this as a question given that calling `ID(ctx)` with the same  above directory will result in a "no such file or directory" error. IMO I think we should be consistent and apply the same existence checking with `Name`. 

cc @jedevc @smolinari

core/integration/directory_test.go
```diff
@@ -1214,3 +1214,57 @@ func (DirectorySuite) TestDigest(ctx context.Context, t *testctx.T) {
...
+
+		t.Run("very nested directory", func(ctx context.Context, t *testctx.T) {
+			veryNestedName, err := dir.Directory("very/nested").Name(ctx)
+			require.NoError(t, err)
+			require.Equal(t, "nested", veryNestedName)
+		})
+	})
+
+	t.Run("git directory", func(ctx context.Context, t *testctx.T) {
+		dir := c.Git("https://github.com/dagger/dagger#ee32df913f57c876e067bd5ecc159561510b6f50").Head().Tree()
```
marcosnils, line 1250: I wish this could return something relevant to the URL being used? @shykes  suggested returning the last  element of the path here I think?

jedevc: Kinda tricky without special casing this though, or adding extra `llb.Copy`s into the `dag.Git` implementation.

Not entirely sure what this would be as well, would it be `dagger` as the repo name? What if you've chosen to clone a subdir using `#<commit>:<dir>`?

Also not super sure what the use of this would be.

TomChv: Yeah I had the same thought in mind, it seems kind of out of scope to returns the full base url + path inside it.
It wouldn't be the name of the directory anymore but rather the full path and we decided to not support that for various reason: https://discord.com/channels/707636530424053791/1336739628350439454/1337383925655404667

TomChv:
> have a conceptual question for this PR. Shouldn't dag.Directory().Directory("non-existent").Name(ctx) raise an error if the directory doesn't exist? I'm raising this as a question given that calling ID(ctx) with the same above directory will result in a "no such file or directory" error. IMO I think we should be consistent and apply the same existence checking with Name.

I going to write a test for that :)

TomChv:
> Agreed that Name on a dir that doesn't exist should error out. Probably should also do the same of File.name if it doesn't already.

I already added a test for that, it errors :)

sipsma (approved):

core/integration/directory_test.go
```diff
@@ -1214,3 +1214,62 @@ func (DirectorySuite) TestDigest(ctx context.Context, t *testctx.T) {
...
+	t.Run("empty directory name", func(ctx context.Context, t *testctx.T) {
+		name, err := c.Directory().Name(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "/", name)
+	})
+
+	t.Run("not found directory", func(ctx context.Context, t *testctx.T) {
+		_, err := c.Directory().Directory("foo").Name(ctx)
+		requireErrOut(t, err, "no such file or directory")
+	})
```
jedevc, line 1230: Can you also a corresponding test in `file_test.go` for `File.name`? Would be good to ensure consistent behavior here.

jedevc (approved):
One test nit, otherwise LGTM :tada:

Issue #9281, opened 2024-12-23
shell: write files with special characters


shykes:
## Problem

I can't use the dagger shell to write files with special characters like newlines or tabs.

## Solution

This would require either 1) support for interpreting escape sequences like `\n` or `\t`, or 2) functional heredocs and/or 3) multi-line editing.

Issue #9202, opened 2024-12-13
shell: readline wipes entire line when deleting the first word


shykes:
## What I did

1. Open interactive dagger shell:

`dagger shell`

2. Type a long command without typing enter:

`container | from alpine | terminal`

3. Place the cursor at the end of the first word `container`

4. Type `alt+delete` or your equivalent key binding for deleting the word to the left

## What I expected

The first word `container` is deleted. The current line is now `⋈ | from alpine | terminal`

## What happened

The entire line is deleted. The current line is now `⋈ `

PR #9620, opened 2025-02-18
Bubbletea-ify shell

vito:
Fixes https://github.com/dagger/dagger/issues/8990
Fixes https://github.com/dagger/dagger/issues/9202
Fixes https://github.com/dagger/dagger/issues/9281

This is a somewhat dramatic change to the `dagger shell` UX and implementation.

Previously `shell` started the TUI and backgrounded it to show the prompt and accept user input. After accepting user input it would foreground the TUI to show progress, and then background it when its command completed. This had the downside of not being able to let progress "linger" in the scrollback.

Now `shell` is directly integrated into the TUI, where we show a prompt and history of commands (using the regular progress output, zoomed to the shell's span).

* To switch from the input prompt to navigating the progress, press <kbd>Esc</kbd>.
* To switch back to the input prompt, press <kbd>Tab</kbd> or <kbd>i</kbd>.
* The prompt should accept all the usual readline keybinds. Let me know if something's broken.
* Autocomplete should also work, but could use a bit more love.
  * [ ] Not sure if the keybinds are good here.
  * [ ] Would be nice to use the full completion features (categories + descriptions).

### Major change: don't print object leaf fields

Instead of auto-selecting an object's scalar/leaf fields and printing them in YAML format, this PR also changes objects to simply print as `TypeName@xxh3:...`.

A few motivations:

* Sometimes those fields are extremely verbose, which sometimes negatively influences API design
* Most of the time those fields aren't very illustrative (e.g. `Container` doesn't show its `imageRef` or the command that it runs)
* It's possible that selecting those fields is destructive

### Open question: scroll wheel/mouse input handling

Now the entire shell session is rendered in our TUI, rather than printed plainly to the terminal scrollback. The TUI isn't able to rendering anything offscreen, so if you scroll the terminal it'll just be cut off at the top.

To handle that, currently our TUI hijacks mouse events so that it can approximate[^1] scrolling. There's no way to _only_ intercept scroll events, and we don't do anything with click events, so if you try to select text nothing happens. Annoying!

The workaround for that is to just hold <kbd>Shift</kbd> while using the mouse, which is the universal[^2] technique to bypass sending mouse input to the program. But not everyone knows that.

Eating mouse input got in the way for the demo so I hastily removed it, which is a "breaking" change across the entire TUI.

**The question: (when) should we eat mouse input?**

There are a few options:

| | Mouse handling | Scroll prompt progress | Scroll outer terminal | Copy/paste |
|---------|-----------------|----------------------|---------------------|------------|
| **A** | Always consume | Scroll | <kbd>Shift</kbd> + scroll | <kbd>Shift</kbd> + select |
| **B** | Never consume | <kbd>Esc</kbd> + arrows/hjkl | Scroll | Select |
| **C** | Only consume in nav mode | <kbd>Esc</kbd> + arrows/hjkl/scroll | Scroll (prompt mode)<br><kbd>Shift</kbd> + scroll (nav mode) | Select (prompt mode)<br><kbd>Shift</kbd> + select (nav mode) |

Some notes:

* Option A is the current default on `main` and v0.15.x.
* Option B is implemented by this PR, but it was a hasty change, and is a change across-the-board (not just for `shell`).
* Option C is a compromise, but I'm not sure if it's intuitive.

[^1]: It switches focus between spans and renders outward from the focused span, it doesn't act like a pager.
[^2]: By my measurement, anyway; never ran into a terminal where this didn't work.


helderco:
Doing `Ctrl+D` exits with status 2.

helderco:
When I enter the shell all I see is the prompt. I need to scroll up to see the welcome message ("Dagger interactive shell. Type ".help" for more information. Press Ctrl+D to exit.").

shykes (approved):

vito:
For continuity: I went with a new option **D**:

* Never consume mouse input, but only in `shell` mode
* As top-level shell commands complete, print them to a scrollback, so you can just scroll up like normal
* In-progress commands are displayed with a highlighted background, if supported
* Switching to nav keeps same highlighted background to distinguish from completed items in the scrollback

This feels like a good balance to me: mouse actions "just work" as you would expect, scrollback behaves like a normal shell, and we don't break any behavior outside of `shell`.

Issue #9619, opened 2025-02-18
🐞 Memory consumption with large module trees is out of control, failing BenchmarkModule/BenchmarkLotsOfDeps


cwlbraa:
### What is the issue?

Shortly after we extracted BenchmarkModule to only run on chron-main and labels, BenchmarkModule started failing to cancellation. on further investigation, it turns out engines are getting oomkilled partway through each run. 

I hand-bisected and tracked this problem down to https://github.com/dagger/dagger/pull/9483. The tests pass prior to that commit and fail afterwards. 

We can scale down the levels of nesting in the test from 6 to 4 and then they can pass with 16G of engine memory, but even with that concession, this test takes ~3m40s to pass with the per-client caching and ~2m38s to pass without- a significant performance regression. 

### Dagger version

v0.15.4

### Steps to reproduce

dagger core engine local-cache prune && dagger call bench specific --run=BenchmarkModule/BenchmarkLotsOfDeps

### Log output

[GHA benchmark passes](https://github.com/dagger/dagger/actions/workflows/benchmark-engine.yml?query=is%3Asuccess) (very few, but all right prior to this change going in)
[1 failed trace example](https://v3.dagger.cloud/dagger/traces/57fb74496078f556877642e9dacc097e?listen=0feef42bb9966e53)

PR #9477, opened 2025-01-29
notify discord on benchmark failure

cwlbraa:
not yet mergeable, but need a little help configuring the discord webhook. I think I don't have enough powers on our discord server to generate a webhook, and I'd also kinda like to make a channel so these failures can't pollute #engine-dev

PR #9621, opened 2025-02-19
make function calls cached session-wide

sipsma:
Should fix https://github.com/dagger/dagger/issues/9619, can run `BenchmarkLotsOfDeps` locally now without excessive memory usage.

---

Recently, function calls were "upgraded" from being tainted (i.e. never
cached) to cached-per-client in terms of dagql caching. This helped
fixed a problem with duplicated telemetry.

However, we currently have two layers of caching between dagql and
buildkit and this cache-per-client logic inadvertently ended up also
applying to the buildkit cache key, which increased the amount of cache
invalidation there.

This surprisingly didn't have a super noticeable effect on performance
at least as far as our tests that always run go.

However, the benchmark tests Connor added did end up hitting this,
specifically in BenchmarkLotsOfDeps, which creates a complicated DAG of
function calls that, if not cached correctly, will have O(n^2)
performance (in memory and time).

The fix here changes function calls to be cached in dagql session-wide
and restores the previous buildkit cache key so that we get consistent
session-wide cache hits there again too.

There is some complication from secrets, sockets and other
client-specific resources though. They need to transfer from return
values of function calls to the calling client whether or not the client
hit the cache.

To handle that problem, this change adds support for a "post call"
callback that, if set, dagql will always execute on results whether or
not they were retrieved from cache or an actual execution.

The implementation is admittedly pretty kludgy and ugly right now, but I
think we should go with it for now to fix this problem. We are about to
embark on quite a bit of work around dagql caching so hopefully we can
find a cleaner way of implementing post calls (or some equivalent
replacement) at that time.

### DIFF
--- a/dagql/objects.go
+++ b/dagql/objects.go
@@ -521,7 +521,32 @@ func (r Instance[T]) call(
 		}
 	}
 
+	// field implementations can optionally return a wrapped Typed val that has
+	// a callback that should always run after the field is called
+	if postCallVal, ok := val.(*PostCallTyped); ok {
+		val = postCallVal.Typed
+		if postCallVal.PostCall != nil {
+			if err := postCallVal.PostCall(ctx); err != nil {
+				return nil, nil, fmt.Errorf("post-call error: %w", err)
+			}
+		}
+	}
+
 	return val, newID, nil
+}
+
+// PostCallTyped wraps a Typed value with an additional callback that
+// needs to be called after any value is returned, whether the value was from
+// cache or not
+type PostCallTyped struct {
+	Typed
+	PostCall func(context.Context) error
+}
+
+var _ Wrapper = PostCallTyped{}
+
+func (p PostCallTyped) Unwrap() Typed {
+	return p.Typed
 }
 
 type View interface {
--- a/core/interface.go
+++ b/core/interface.go
@@ -278,7 +278,7 @@ func (iface *InterfaceType) Install(ctx context.Context, dag *dagql.Server) erro
 					})
 				}
 
-				res, err := callable.Call(ctx, &CallOpts{
+				postCallRes, err := callable.Call(ctx, &CallOpts{
 					Inputs:       callInputs,
 					ParentTyped:  runtimeVal,
 					ParentFields: runtimeVal.Fields,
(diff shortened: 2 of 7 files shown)


sipsma:
The benchmark tests are not running due to unrelated issues (minimum client version bumped to v0.16.0), but local runs of LotsOfDeps look good.

I'll take this out of draft. If others review and feel okay with it we can include in v0.16.0 but it's not a big deal if it waits for v0.16.1

.github/workflows/benchmark-engine.yml
```diff
@@ -23,13 +23,21 @@ jobs:
...
         uses: actions/checkout@v4
       - name: Warm up Engine
-        run: dagger core version
+        run: |
+          # Detect if a dev engine is available, if so: use that
+          # We don't rely on PATH because the GHA runner messes with that
+          if [[ -n "$_EXPERIMENTAL_DAGGER_CLI_BIN" ]]; then
+            ls -lh $(dirname $_EXPERIMENTAL_DAGGER_CLI_BIN)
+            export PATH=$(dirname "$_EXPERIMENTAL_DAGGER_CLI_BIN"):$PATH
+          fi
```
cwlbraa, line 38: how would there be a dev engine here? there's nothing resembling a build ahead of this step?

jedevc (approved):

.github/workflows/benchmark-engine.yml
cwlbraa: the (initial?) intent here is not to have `dev-` equivalent tests, both on main and on branches the outside engine is always the released version

cwlbraa: from the #9477 where i'm trying to fix the notifications for the benchmark jobs...

https://github.com/dagger/dagger/actions/runs/13377712071/job/37361212417?pr=9477

reports v0.16.0-longdevhash in the warmup, but is running on a 15.4 runner, so clearly i'm confused about something here

sipsma: Ah yeah this was just a shot in the dark, I will undo it. I think we could most likely just wait this out for v0.16.0, once our CI runners are upgraded to it this should go away afaik

dagql/objects.go
```diff
@@ -521,9 +521,28 @@ func (r Instance[T]) call(
...
+	}
+
 	return val, newID, nil
 }
 
+// PostCallTyped wraps a Typed value with an additional callback that
+// needs to be called after any value is returned, whether the value was from
+// cache or not
+type PostCallTyped struct {
+	Typed
```
vito, line 542: We might need/want this to also implement `Unwrap() Typed` so that things like telemetry middleware (`core.AroundFunc`) can still get to the underlying value

sipsma: Ah sure, I guess telemetry gets to it before the unwrapping happens in the current code, will fix, thanks!

PR #9623, opened 2025-02-19
helm: Enable setting engine pod annotations

vmaffet:
This PR enables setting the engine pod annotations in the same way as the labels.


jedevc (approved):

matipan (approved):
We've seen a similar issue with Istio on some tests we did very recently. Thanks for the patch

PR #9624, opened 2025-02-19
docs: add PHP examples to cookbook

jasonmccallister:
Adds PHP snippet examples to the cookbook page.

### Adding new PHP snippets

Here are the steps I'm using to create the snippets - maybe verbose but would be worth looking at making this a Dagger function to automate in the future?

1. Run the following
```bash
cd docs/current_docs/cookbook/snippets/<example_name>
mkdir php
cd php
dagger init --sdk php --name my-module
```
2. Update the example to match the other languages
3. Remove the following files: `docs/current_docs/cookbook/snippets/<example_name>/php/README.md`

> Note: If you are running `yarn run start` you might get compilation errors (below) from the `vendor` dir README.md

<details>
Screenshot 2025-02-19 at 11 36 21 AM
</details>

### DIFF
--- a/docs/current_docs/cookbook/snippets/clone-git-repository/php/src/MyModule.php
+++ b/docs/current_docs/cookbook/snippets/clone-git-repository/php/src/MyModule.php
@@ -14,11 +14,22 @@ use function Dagger\dag;
 class MyModule
 {
     #[DaggerFunction]
-    public function clone(string $repository, string $branch): Container
+    public function clone(string $repository, string $locator, string $ref): Container
     {
         $r = dag()->git($repository);
         $d = dag()->directory();
-        $d = $r->branch($branch)->tree();
+
+        switch ($locator) {
+            case 'branch':
+                $d = $r->branch($ref)->tree();
+                break;
+            case 'tag':
+                $d = $r->tag($ref)->tree();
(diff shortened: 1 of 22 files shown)


docs/current_docs/cookbook/snippets/clone-git-repository/php/src/MyModule.php
```diff
@@ -0,0 +1,28 @@
...
+use Dagger\Attribute\DaggerObject;
+use Dagger\Container;
+
+use function Dagger\dag;
+
+#[DaggerObject]
+class MyModule
+{
+    #[DaggerFunction]
+    public function clone(string $repository, string $branch): Container
```
vikram-dagger, line 17: I know the PHP SDK doesn't support enums so we can't reproduce the other snippets _exactly_ but it may be worth trying to get as close as possible: maybe add a $locator argument and a switch-case for three possible values, or default to branch if none match. WDYT?

jasonmccallister: Yeah, we should definitely keep that inline. I'll update the example to align.

vikram-dagger (approved):

PR #9625, opened 2025-02-19
[PHP] Update composer call to update generated code dependencies

carnage:
Solve issue where we update a dependency in the generated code and the module composer.lock prevents that dependency from being updated. 

Thanks to @naderman for the assist.


jedevc (approved):
Oooh nice! This looks like a continuation of the fix that we discussed about in discord right?

charjr:
> Oooh nice! This looks like a continuation of the fix that we discussed about in discord right?

:+1: 

@carnage If I understand correctly; Since it's an `update`, does this ignore the lock file?

Does this put us in a better position than simply ignoring the lock?

charjr:
Also I know that we haven't personally used any other PHP packages in our modules. 

However, if someone uses another library like `nesbot/carbon`, will this update command still install it? It looks like we might still need the `withExec(['composer', 'install'])` to be called afterwards just in case.

jedevc:
Awesome, that's pretty much how it works in go, I think it's fair to have the same behavior here :tada:

charjr (approved):
I did a quick bit of testing locally and this is what I can sum up.

### Old Behaviour

- It can perform a fresh install of dependencies
- It can install new dependencies on the module.
- It **does not install** any new dependencies of the sdk.
- It **does not update** any existing dependencies of the sdk.

### New Behaviour

- It can perform a fresh install of dependencies
- It can install new dependencies on the module.
- It **will install** any new dependencies of the sdk.
- **It will perform** the bare minimum of **updates** to satisfy the sdk dependencies.

I think overall, this is a definite improvement. It should solve a lot of hiccups.

PR #9504, opened 2025-02-04
(deprecated) POC: native LLM support

shykes:
This PR is a prototype for native LLM support in Dagger.

* New core type: `Llm`. Represents the state of a LLM - which it turns out fits perfectly in a DAG of immutable states!
* Add any Dagger object to the Llm state: the LLM can now discover the object's functions, and call them - making it an agent! It's like magic

This paves the way to using Dagger as a backend for agentic workflows. Most AI agent platforms focus on 1) distributed infrastructure for event dispatching, job queues, state persistence; and 2) model-facing "frontend" code for prompt engineering, connecting tools etc. What's missing is 3) implementing the actual backend logic for the tools, and orchestrating them into a repeatable "micro-workflow".

This is incredibly similar to the problem we already solve in CI:

1) CI platforms provide the distributed infrastructure
2) CI scripts focus on domain-specific issues like what job status to present in the github pull request, which checks are required to merge the PR, etc
3) the actual implementation of builds and tests is best left to Dagger, for maximum repeatability and portability.

## Setup

1. Write a `.env` *in your current directory*, configured with your LLM key and other optional configuration. (Eg. your OpenAI API token).

There are several options for format:

- Env variable, example: `LLM_KEY=env://OPENAI_API_KEY`
- File, example: `LLM_KEY=file:///Users/bob/.openai-token.key`
- 1password secret, example: `LLM_KEY=op://Private/bsfoobarsh5bazhsp36yw524/credential`
- Hashicorp vaul, example: `LLM_KEY=vault://blablabla`
- Cleartext value (not recommended): `LLM_KEY=<actual token>`

2. Install latest (stable) dagger

3. Use stable dagger to build dev CLI & run dev engine

```console
dagger shell <<EOF
github.com/shykes/dagger@llm | cli | binary --platform=current | export "$HOME/bin/dagger-llm"
github.com/shykes/dagger@llm | engine | service llm | up
EOF
```

4. Use dev CLI to connect to dev engine.

```console
_EXPERIMENTAL_DAGGER_RUNNER_HOST=tcp://localhost:1234 ~/bin/dagger-llm shell
```

5. Test that it works

Run this inside your dev shell session:

```console
llm | with-directory $(directory | with-new-file name.txt "Your name is Thor") | with-prompt "read all .txt files available in the directory" | with-prompt "what is your name?" | history
```

## Known bugs

I am at a loss as to how to fix these bugs, humbly begging for your help @jedevc @vito @sipsma 😭 
[...]

PR #9626, opened 2025-02-19
ci: allow tagging arbitrary pre-releases

jedevc:
Based off of @gerhard's work in [`2273cd2` (#9504)](https://github.com/dagger/dagger/pull/9504/commits/2273cd29a5283f35a767f18a532a7b7c9ff0f3ff).

We should be able to merge this directly to `main`, and once `llm` is rebased, the tagging should work a bit easier.

### DIFF
--- a/.dagger/cli.go
+++ b/.dagger/cli.go
@@ -79,7 +79,7 @@ func (cli *CLI) Publish(
 	}
 
 	args := []string{"release", "--clean", "--skip=validate", "--verbose"}
-	if tag != "" {
+	if tag != "" && semver.Prerelease(tag) == "" {
 		args = append(args, "--release-notes", fmt.Sprintf(".changes/%s.md", tag))
 	} else {
 		args = append(args,
--- a/releaser/main.go
+++ b/releaser/main.go
@@ -158,6 +158,7 @@ func (r *Releaser) Publish(
 	if semver.IsValid(tag) {
 		version = tag
 	}
+
 	report := ReleaseReport{
 		Date:    time.Now().UTC().Format(time.RFC822),
 		Ref:     tag,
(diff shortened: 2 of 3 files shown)


.dagger/cli.go
```diff
@@ -79,7 +79,7 @@ func (cli *CLI) Publish(
 	}
 
 	args := []string{"release", "--clean", "--skip=validate", "--verbose"}
-	if tag != "" {
+	if tag != "" && semver.Prerelease(tag) == "" {
```
gerhard, line 82: `Prerelease` is a nice touch, I was going to check if this functionality exists - great to see you had the same idea & already implemented it!

releaser/main.go
```diff
@@ -170,7 +171,13 @@ func (r *Releaser) Publish(
 		Tag:    tag,
 		Notify: true,
 	}
-	err := r.Dagger.Engine().Publish(ctx, []string{tag, commit}, dagger.DaggerDevDaggerEnginePublishOpts{
+
+	tags := []string{tag, commit}
+	if semver.IsValid(version) && semver.Prerelease(version) == "" {
+		// this is a full semver release
+		tags = append(tags, "latest")
```
gerhard, line 178: I would call this a "public" instead of "full semver", but just a nit, LGTM!

releaser/main.go
```diff
@@ -203,7 +210,13 @@ func (r *Releaser) Publish(
 	report.Artifacts = append(report.Artifacts, artifact)
 
 	if report.hasErrors() {
-		// early-exit if engine or cli could not publish
+		// early-exit if engine / cli could not Publish
+		return &report, nil
+	}
+
+	isPrerelease := semver.IsValid(version) && semver.Prerelease(version) != ""
+	if isPrerelease {
```
gerhard, line 218: I would be tempted to wrap this in a `isPrerelease()` function, and just reuse it in both places, but just a nit, not worth blocking on it.

gerhard (approved):
LGTM 👍 🚀

jedevc:
Will merge tomorrow, so I can keep an eye on the engine/cli nightly things that are pushed (and to avoid breaking v0.16).

PR #9628, opened 2025-02-19
POC: native LLM support 

shykes:
This PR is a prototype for native LLM support in Dagger.

* New core type: `Llm`. Represents the state of a LLM - which it turns out fits perfectly in a DAG of immutable states!
* Add any Dagger object to the Llm state: the LLM can now discover the object's functions, and call them - making it an agent! It's like magic

This paves the way to using Dagger as a backend for agentic workflows. Most AI agent platforms focus on 1) distributed infrastructure for event dispatching, job queues, state persistence; and 2) model-facing "frontend" code for prompt engineering, connecting tools etc. What's missing is 3) implementing the actual backend logic for the tools, and orchestrating them into a repeatable "micro-workflow".

This is incredibly similar to the problem we already solve in CI:

1) CI platforms provide the distributed infrastructure
2) CI scripts focus on domain-specific issues like what job status to present in the github pull request, which checks are required to merge the PR, etc
3) the actual implementation of builds and tests is best left to Dagger, for maximum repeatability and portability.

## Setup

1. Write a `.env` *in your current directory*, configured with your LLM key and other optional configuration. (Eg. your OpenAI API token).

There are several options for format:

- Env variable, example: `LLM_KEY=env://OPENAI_API_KEY`
- File, example: `LLM_KEY=file:///Users/bob/.openai-token.key`
- 1password secret, example: `LLM_KEY=op://Private/bsfoobarsh5bazhsp36yw524/credential`
- Hashicorp vaul, example: `LLM_KEY=vault://blablabla`
- Cleartext value (not recommended): `LLM_KEY=<actual token>`

2. Install latest (stable) dagger

3. Use stable dagger to build dev CLI & run dev engine

```console
dagger shell <<EOF
github.com/shykes/dagger@llm | cli | binary --platform=current | export "$HOME/bin/dagger-llm"
github.com/shykes/dagger@llm | engine | service llm | up
EOF
```

4. Use dev CLI to connect to dev engine.

```console
_EXPERIMENTAL_DAGGER_RUNNER_HOST=tcp://localhost:1234 ~/bin/dagger-llm shell
```

5. Test that it works

Run this inside your dev shell session:

```console
llm | with-directory $(directory | with-new-file name.txt "Your name is Thor") | with-prompt "read all .txt files available in the directory" | with-prompt "what is your name?" | history
```

## Known bugs

I am at a loss as to how to fix these bugs, humbly begging for your help @jedevc @vito @sipsma 😭 
I added details about each bug below, including links to relevant files.

~~### Bug 1. "Can't instantiate..."~~ *fixed!*

~~### Bug 2. "FIeld \"withExec\" not found..."~~ *fixed!*

~~### Bug 3. Cannot create int from float64~~ *fixed!* thank you @aluzzardi 🙏 

~~### Bug 4. Cannot load modules~ *fixed!* thank you @vito 🙏 

~~### Bug 5. `Llm` can't use module types~~ *fixed!* thank you @sipsma 🙏

~~### Bug 6. broken auto-chaining~~ *fixed!* thank you @vito 🙏 

~~### Bug 7. modules can't access LLM config in `.env`~~ *fixed!*


core/schema/engine.go
```diff
@@ -114,3 +114,9 @@ func (s *engineSchema) cachePrune(ctx context.Context, parent *core.EngineCache,
 func (s *engineSchema) cacheEntrySetEntries(ctx context.Context, parent *core.EngineCacheEntrySet, args struct{}) ([]*core.EngineCacheEntry, error) {
 	return parent.EntriesList, nil
 }
+
+// A utility to retrieve the root query object from the underlyign dagql server
+// Note: this only works after the root query object has been installed!
+func (s *engineSchema) rootQuery() *core.Query {
```
sipsma, line 120: seems unused?

sipsma:
Been trying it out locally and reviewing the code.

One comment not attached to any single part of the code is that the LLMs seem to be unable to work with lists, specifically they can't figure out how to retrieve a single element from a list or even be aware that lists exist as a type.

For a super trivial example:

`workspace` module:
```go
type Workspace struct {
	Files []*dagger.File
}

func New(files ...*dagger.File) *Workspace {
	return &Workspace{
		Files:      files,
	}
}

// Return the first line of the content of the given file. Accepts only a single file, not a list.
func (m *Workspace) FirstLine(ctx context.Context, f *dagger.File) (string, error) {
	contents, err := f.Contents(ctx)
	if err != nil {
		return "", err
	}

	return contents[:strings.Index(contents, "\n")], nil
}
```

calling `test` module:
```go
type Test struct{}

func (m *Test) Fn(ctx context.Context, files ...*dagger.File) (string, error) {
	ws := dag.Workspace(files)
	return dag.Llm().
		WithWorkspace(ws).
		WithPrompt("You have access to a list of Files. Your job is to check the first line of each file, find the one with the first line being \"123\" and tell me the name of the file.").
		LastReply(ctx)
}
```

At least w/ Anthropic (default model, whatever that is), it consistently just tries to pass the full list of `[]*dagger.File` to the `FirstLine` tool, which errors out on `loadFileFromID`.
* Interestingly, when I ask it "what are the values in the Files list?", it says "there appears to be one file with the identifier "xxh3:d248a472c8789267" in the files list".
* I believe this is because it's just getting the ID *for the list as a whole*, so it has no insight into the fact that it's a list or there are elements

 I'm not 100% sure but it seems like we might need something like:
* Including the fact that a given type is a list in the tool description, so the LLM knows
* a "static" tool call that gives the LLM the ability to select elements from a list (i.e. equivalent of `nth` in dagql IDs)

core/bbi/flat/flat.go
```diff
@@ -0,0 +1,337 @@
...
+				// If target object has a sync() function, call it
+				if _, hasSync := obj.ObjectType().FieldSpec("sync"); hasSync {
+					// If target object has a sync() function, call it
+					if _, _, err := obj.Select(ctx, s.srv, dagql.Selector{
+						Field: "sync",
+					}); err != nil {
+						return "", err
+					}
+				}
+				s.self = obj
```
sipsma, line 220: These may be handled somehow, but I'm not following how setting `s.self` behaves in various scenarios where the LLM returns multiple Tool calls.

e.g. we *don't* reset it, even to `nil`, in the other cases below (case 2 + 3). In case 2, specifically, what if the LLM starts trying to chain off the returned object there? It seems like we still would continue using `s.self`.

Is it not possible for the LLM to return multiple tool calls right now? If it *is*, it seems like we might need some more robust handling around this.

sipsma:
Another general comment, right now it seems like it would be possible to give an LLM an object that contains fields of type `Secret`, at which point the LLM could arbitrarily decide to use those secrets. That feels like a massive footgun sitting around (i.e. secrets get dumped into LLM context)

Might be worth just preventing LLMs from calling plaintext on any secrets, even if they have access to them. Or any equivalent fix.

marcosnils:
> Might be worth just preventing LLMs from calling plaintext on any secrets, even if they have access to them. Or any equivalent fix.

Wouldn't this be limiting? i.e: If I want the LLM to perform certain action which requires a token, I could intentionally tell it to given "X" secret, just get the plaintext value out it and use it.  I think that could be valid use-case.

sipsma:
> > Might be worth just preventing LLMs from calling plaintext on any secrets, even if they have access to them. Or any equivalent fix.
> 
> 
> 
> Wouldn't this be limiting? i.e: If I want the LLM to perform certain action which requires a token, I could intentionally tell it to given "X" secret, just get the plaintext value out it and use it.  I think that could be valid use-case. 
> 
> 

I's be fine with the LLM using objects of type Secret (i.e. set it as an env var in an exec). I just think it would be much safer to prevent it from calling .plaintext and thus putting the secret plaintext in the context window.

I'm sure some people don't care if their secrets are uploaded to the LLM service but some will probably care a lot, so this seems like a better starting place.

We could equivalently do secret scrubbing on the context data sent to the llm service, etc.

dagql/demo.cast
jedevc, line 1: Not sure this is meant to be here?

dagql/server.go
```diff
@@ -230,47 +237,54 @@ type Loadable interface {
 
 // InstallObject installs the given Object type into the schema.
 func (s *Server) InstallObject(class ObjectType) {
-	s.installLock.Lock()
-	defer s.installLock.Unlock()
+	// FIXME: shortcut to get our "agent" middleware to work
+	// s.installLock.Lock()
+	// defer s.installLock.Unlock()
```
jedevc, line 242: :warning: Leaving a comment just to remind us that we need to fix this before merging.

docs/docs-graphql/schema.graphqls
```diff
@@ -2744,6 +3121,9 @@ type Query {
   """Load a ListTypeDef from its ID."""
   loadListTypeDefFromID(id: ListTypeDefID!): ListTypeDef!
 
+  """Load a Llm from its ID."""
+  loadLlmFromID(id: LlmID!): Llm!
```
jedevc, line 3125: Question for the room: should this be `Llm` or `LLM` in the API? I feel like it should *probably* be `LLM`, but also happy to shorten it.

dagql/idtui/vterm.go
```diff
@@ -190,7 +191,11 @@ func (term *Vterm) View() string {
 var style = styles.LightStyleConfig
 
 func init() {
-	if isDark {
+	if isDark &&
+		// termenv is unable to detect Windows Terminal (at least) background
+		// color, and light Markdown on light background is unreadable, so we
+		// use this env var to force the color.
+		os.Getenv("LIGHT") == "" {
```
jedevc, line 198: Is this a convention? (boo to windows terminal)

jedevc (approved):

PR #9643, opened 2025-02-20
fix: avoid prompting users to upgrade from pre-release versions

jedevc:
[As noticed by @jpadams](https://discord.com/channels/707636530424053791/1319379215849881631/1341950857033420830) during the release of v0.16:

> ```
> A new release of dagger is available: v0.16.0-llm.1 → v0.16.0
> To upgrade, see https://docs.dagger.io/install
> ```
> danger of leaving a bunch of daggers lying around

We really shouldn't be prompting users to upgrade from pre-releases (that have been cut directly off of `main`, or from another branch). We were doing this already for dev builds, but we should extend this to prereleases as well.

1. Users who do this should know what they're doing.
2. It's very confusing, since `llm.1` isn't actually included in `v0.16.0` so the ugprade prompt is actually wrong.

(fyi @aluzzardi, I *think* this new behavior is the expected one, but just wanted to check!)

### DIFF
--- a/engine/version.go
+++ b/engine/version.go
@@ -80,7 +80,7 @@ func cleanVersion(v string) string {
 }
 
 func CheckVersionCompatibility(version string, minVersion string) bool {
-	if IsDevVersion(version) && IsDevVersion(Version) {
+	if isDevVersion(version) && isDevVersion(Version) {
 		// Both our version and our target version are dev versions - in this
 		// case, strip pre-release info from our target, we should pretend it's
 		// just the real thing here.
@@ -90,7 +90,7 @@ func CheckVersionCompatibility(version string, minVersion string) bool {
 }
 
 func CheckMaxVersionCompatibility(version string, maxVersion string) bool {
-	if IsDevVersion(version) && IsDevVersion(Version) {
+	if isDevVersion(version) && isDevVersion(Version) {
 		// see CheckVersionCompatibility
 		version = BaseVersion(version)
 	}
@@ -121,7 +121,11 @@ func BaseVersion(version string) string {
 	return version
 }
 
-func IsDevVersion(version string) bool {
+func IsFinalRelease(version string) bool {
+	return semver.IsValid(version) && semver.Prerelease(version) == ""
+}
+
+func isDevVersion(version string) bool {
 	if version == "" {
 		return true
 	}
--- a/cmd/dagger/version.go
+++ b/cmd/dagger/version.go
@@ -64,7 +64,7 @@ func long() string {
 }
 
 func updateAvailable(ctx context.Context) (string, error) {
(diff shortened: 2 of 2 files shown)


engine/version.go
```diff
@@ -121,7 +121,11 @@ func BaseVersion(version string) string {
 	return version
 }
 
-func IsDevVersion(version string) bool {
+func IsPublicRelease(version string) bool {
```
gerhard, line 124: I would call this `IsFinalRelease()`. A pre-release MAY be a public release.

Also, anything which has any characters following `MAJOR.MINOR.PATCH` is considered a pre-release. Whether it starts with `-dev`, `-alpha`, `-rc` or something else (`-llm` in this case) is irrelevant: they are all pre-releases.

Source: https://semver.org/#spec-item-9

gerhard (commented):
I left a comment which might be worth unpacking.

I did not dig into the code, but how do we determine which is the latest release? This is the source of truth that I **thought** we were using: https://dl.dagger.io/dagger/latest_version (CDN). When I produced `0.16.0-llm.1` manually, I checked the origin file and that did not update.

jedevc:
> I did not dig into the code, but how do we determine which is the latest release?

We inspect `registry.dagger.io/engine:latest` and get the annotations from there. The logic is here: https://github.com/dagger/dagger/blob/4fcf5efdf0927883e5fd58d789b9a969471acff5/cmd/dagger/version.go#L85-L115

gerhard:
If we are getting the annotations from `registry.dagger.io/engine:latest` then that's fine.

So this message is correct:

```console
A new release of dagger is available: v0.16.0-llm.1 → v0.16.0
```

I read it as: _upgrade to the final `0.16.0` release from the current `0.16.0` pre-release_

If however this was `v0.17.0-llm.1`, which I am just about to tag and kick off, then this would **not** be correct:

```console
A new release of dagger is available: v0.17.0-llm.1 → v0.16.1
```

The initial thinking is "we don't want messages like the one above for pre-releases", but I think that the correct behaviour would be to have them **if** `latest` is higher than the current pre-release. A few examples when I think that we should show this message:
- `0.16.0-llm.1` -> `0.16.0`
- `0.16.2-dev.xxx` -> `0.16.2`
- `0.16.1` -> `0.16.2-rc.1` if `latest` points to this tag (we don't have this yet)

And here are a few examples when we shouldn't:
- `0.17.0-llm.1` x `0.16.2`
- `0.16.2-dev.xxx` x `0.16.1`

gerhard (approved):
OK to merge as is, we can follow-up with the above as/when needed.

jedevc:
Closed, since I don't 100% disagree with @gerhard's logic.

We *should* fix this by instead making pre-releases against v0.17.0 instead of anything in the v0.16.0 series.

PR #9645, opened 2025-02-20
feat: decouple SDK loading from module source

TomChv:
It's required by #9531 because it loads a standalone SDK not related to the current source.

I first wanted to create expose the SDK with API call but I figured out that we can keep it internal as long as it's not
bundle to a specific object so it can call from anywhere.


jedevc (approved):
SGTM, it's an internal only API! No risk of breakage :laughing:

PR #9646, opened 2025-02-20
ci: add goreleaser pre-release config

jedevc:
Follow-up to https://github.com/dagger/dagger/pull/9626.

When we do pre-releases, like for our LLM branch, we should avoid publishing to our homebrew tap, etc. To handle this, we create a new config and use that for our pre-releases.

(this also includes some nice changes to make it easier to `--dry-run` and debug the cli publish step, by returning the `dist` directory, and having optional secrets)


gerhard (approved):
👍 🚀

PR #9649, opened 2025-02-20
ci: when publishing, add prerelease variants

jedevc:
This will let us tag `v0.17.0-llm.2`, which would then create a "pointer" at `v0.17.0-llm` on the remote file server.

Then, we also update `install.sh` to allow handling these build pointers. In addition, this also lets us set `DAGGER_VERSION=0.16` (since we already have major.minor pointers) set.

I don't *think* we were using these pointers before? (cc @gerhard, maybe there was some use I wasn't familiar with). But we can pretty easily just integrate this in.


install.sh
jedevc, line 1: Note to self: `install.ps1` should also be updated.

gerhard:
I started looking into this, but it will take a bit since I need to go through all the different variants. As I do that, I am going to add some tests for it since this is becoming unwieldy.

jedevc:
@gerhard I've added some tests for some of the particularly tricky bits.

shykes:
Is this mergeable? Would be very helpful!

gerhard (approved):
LGTM 👍 🚀

PR #9650, opened 2025-02-20
Add arm64 coverage and module loading to provision tests

sipsma:
We need coverage of module loading on arm64 to catch problems where x86 artifacts (built by our x86 ci runners) accidentally end up in arm64 images.

This is a quick update to our provision tests to give us that additional coverage. Just runs `dagger call -m ... --help` to trigger a module load for the SDKs covered here.


sipsma:
idk why I can't trigger the publish the gha step here... the yaml seems like it should result in it running 😞

PR #9651, opened 2025-02-20
Fix provision tests

sipsma:
I could not for the life of me figure out how to trigger publish jobs on the PR in https://github.com/dagger/dagger/pull/9650 so I went ahead and just pushed to see what happened, didn't work.

Have fixes here, but will bang my head until I can convince gh to run the job here so we don't have to repeat this loop more


sipsma:
Ah, it seems that the `paths` setting under `pull_request` can't start with `./`... Sure, okay? Works as expected now either way

jedevc (approved):

PR #9652, opened 2025-02-20
reapply tsx bundling w/ fix for x-architecture builds

cwlbraa:
^ in the title, this reapplies #9526.


sipsma:
DCO was missing on the revert commit (git doesn't do that automatically for those for some reason...), pushed a fix for that

PR #9653, opened 2025-02-20
make function calls cached session-wide

sipsma:
Giving https://github.com/dagger/dagger/pull/9621 another try. I [ended up reverting it for v0.16.1](https://github.com/dagger/dagger/pull/9634#issuecomment-2670392727) because, for some reason, sdk dev tests on main started consistently failing due to a problem with it seemingly out of nowhere. The problem had to do with `.(Enumerable)` assertions not working, which should have been caught much earlier in the original PR itself, but wasn't somehow.

Either way, this makes some adjustments to the original PR to do the unwrapping a bit earlier and avoid all the type assertions.

@vito RE: conversation in discord about this, found how to get it work now. Basically just replaced the `GetOrInitializeOnHit` (which appears to have been unused atm) with a variant `GetOrInitializeWithPostCall`. That way we can move the unwrapping inside a bit but still ensure it always runs even on cache hits.

---

Original PR commit msg:

Recently, function calls were "upgraded" from being tainted (i.e. never cached) to cached-per-client in terms of dagql caching. This helped fixed a problem with duplicated telemetry.

However, we currently have two layers of caching between dagql and buildkit and this cache-per-client logic inadvertently ended up also applying to the buildkit cache key, which increased the amount of cache invalidation there.

This surprisingly didn't have a super noticeable effect on performance at least as far as our tests that always run go.

However, the benchmark tests Connor added did end up hitting this, specifically in BenchmarkLotsOfDeps, which creates a complicated DAG of function calls that, if not cached correctly, will have O(n^2) performance (in memory and time).

The fix here changes function calls to be cached in dagql session-wide and restores the previous buildkit cache key so that we get consistent session-wide cache hits there again too.

There is some complication from secrets, sockets and other client-specific resources though. They need to transfer from return values of function calls to the calling client whether or not the client hit the cache.

To handle that problem, this change adds support for a "post call" callback that, if set, dagql will always execute on results whether or not they were retrieved from cache or an actual execution.

The implementation is admittedly pretty kludgy and ugly right now, but I think we should go with it for now to fix this problem. We are about to embark on quite a bit of work around dagql caching so hopefully we can find a cleaner way of implementing post calls (or some equivalent replacement) at that time.

### DIFF
--- a/core/modfunc.go
+++ b/core/modfunc.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	bkgw "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/identity"
@@ -22,6 +23,7 @@ import (
 
 	"github.com/dagger/dagger/analytics"
 	"github.com/dagger/dagger/dagql"
+	"github.com/dagger/dagger/engine"
 	"github.com/dagger/dagger/engine/buildkit"
 	"github.com/dagger/dagger/engine/server/resource"
 	"github.com/dagger/dagger/engine/slog"
@@ -213,7 +215,7 @@ func (fn *ModuleFunction) setCallInputs(ctx context.Context, opts *CallOpts) ([]
 	return callInputs, nil
 }
 
-func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Typed, rerr error) { //nolint: gocyclo
+func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t *dagql.PostCallTyped, rerr error) { //nolint: gocyclo
 	mod := fn.mod
 
 	lg := bklog.G(ctx).WithField("module", mod.Name()).WithField("function", fn.metadata.Name)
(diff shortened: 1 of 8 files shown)


core/modfunc.go
```diff
@@ -389,18 +391,34 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Typ
...
+				return fmt.Errorf("failed to get client metadata: %w", err)
+			}
+			if _, alreadyRan := callerClientMemo.LoadOrStore(clientMetadata.ClientID, struct{}{}); alreadyRan {
+				return nil
+			}
+
+			for _, id := range returnedIDs {
+				if err := fn.root.AddClientResourcesFromID(ctx, id, clientID, false); err != nil {
+					return fmt.Errorf("failed to add client resources from ID: %w", err)
+				}
```
jedevc, line 417: question: does this also affect the changes in https://github.com/dagger/dagger/pull/9697?

For example, can `LoadContext` also end up being cached in some cases where we'd need to do a similar transfer?

jedevc (approved):
One question, but LGTM! :tada:

core/modfunc.go
sipsma: Good question but as far as I can think, it shouldn't matter. The changes in this PR ensure that the caller always loads secrets in *return* values whether or not the call itself was cached. So then the scenarios are:
1. Call is not cached: the secrets needed for the contextual args will be loaded since it's gonna execute
2. Call *is* cached: contextual args aren't needed at all since the call is cached. If any of the secrets referenced by the contextual args end up referenced in the return value, then the secrets will be transferred.

PR #9655, opened 2025-02-20
sdk/java: use cache to store maven dependencies

eunomie:
Before the built dependencies were copied, replace it by a cache and re-use the same container. This improve the performances when creating multiple modules for instance.


jpadams:
This makes sense to me from userland, but don't understand the implications for an SDK enough to approve. Added folks who can.

PR #9656, opened 2025-02-20
feat: introduce Anthropic support to `llm` type

grouville:
This PR introduces the anthropic support to the `llm` feature of Dagger.

It is based on @samalba's https://github.com/shykes/dagger/pull/297.

The implementation introduces a `SendQuery` interface that every provider needs to implement, as anthropic is not openai client retrocompatible.

This sets the foundation to a more granular, per provider implementation. 

The Anthropic implementation streams the output the same way the openAI client does, with the same span collection, for a stream as it goes output parsing

cc @kpenfound @vito feel free to try it out, and _ensure that I didn't break anything_

The tool calling capabilities of the model is not as smooth as the openAI one. I am digging that, but, the plumbing is there.


kpenfound (approved):
Tested and OpenAI/Ollama still works

PR #9657, opened 2025-02-20
[Typescript SDK] Fix type declaration path in package.json

jholm117:
I was having trouble loading the type declaration file in my project. I noticed that the "types" path in the package.json didn't match the actual location of the declaration file. I tried updating it with a `yarn patch` operation in my project and the issue was gone!

### DIFF
--- a/sdk/typescript/package.json
+++ b/sdk/typescript/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "author": "hello@dagger.io",
   "license": "Apache-2.0",
-  "types": "./dist/index.d.ts",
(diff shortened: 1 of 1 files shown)


TomChv (approved):
Good catch! I tried to build locally and I confirm that it's out in `src` base dir

```shell
head -n 5  dist/src/index.d.ts
export { gql } from "graphql-tag";
export { GraphQLClient } from "graphql-request";
export * from "./api/client.gen.js";
export * from "./common/errors/index.js";
export type { CallbackFct } from "./connect.js";
```

TomChv:
I'm trying to make the CI green and then I'll merge this PR

PR #9658, opened 2025-02-20
fix panic when dagger call/functions called in dir with no modules

sipsma:
Calling `dagger functions` or `dagger call` was panicking rather than erroring cleanly due to an oversight in `initializeModule`. Now it errors nicely.


helderco (approved):
LGTM 👍

PR #9659, opened 2025-02-21
fix load of full dir when no dagger.json 

sipsma:
There was some leftover logic that resulted in the full source root dir
being loaded when no dagger.json was found, which in the case of
`dagger` commands being run in a dir w/ no module present could result
in lots of files being loaded.

This just removes that logic. This is most important for the shell,
which is totally valid to use outside a module dir, but also allows the
CLI to error out faster for `call`, `functions`, etc.


jedevc (approved):
:tada: yup can confirm this fixes my original issue!

helderco (approved):
lgtm 👍

PR #9664, opened 2025-02-21
ci: fix goreleaser prereleases (again)

jedevc:
The Version correctly shouldn't have a `v` (according to goreleaser docs).

However, the `v` is present in the name template for the archive - so we need to make sure it's there.


gerhard (approved):
LGTM!

PR #9676, opened 2025-02-21
sdk/java: read description from annotation first

eunomie:
In case `@Module`, `@Object` or `@Function` annotation defines a `description` field, uses it instead of the JavaDoc comment.

> [!NOTE]
> to be merged after #9523 

Only the last commit is relevant, the rest is #9523


eunomie:
Integrated back in #9523

PR #9678, opened 2025-02-21
[Typescript SDK] Add @opentelemetry/core to package dependencies

jholm117:
This package was not declared in the `package.json` so I was seeing a corresponding runtime error. The package was already available via transitive dependencies so in less strict environments it probably hasn't been an issue.

The package is directly imported here: https://github.com/dagger/dagger/blob/main/sdk/typescript/src/telemetry/init.ts#L1


jholm117:
Hey @TomChv any chance I could get a review? Looks like you've been tagged

PR #9679, opened 2025-02-21
fix(helm): sanitize version label

b4nst:
The Helm `app.kubernetes.io/version` label is not sanitized. This can lead to incorrect template generation on some system updating the version.

For example with flux, we need to use an OCI registry as the source of the helm chart (because HEAD requests on `registry.dagger.io/dagger-helm/dagger:<tag>` fails with 403 unauth, preventing usage of HelmRepository resource). The default behavior of flux is [to append the digest](https://fluxcd.io/flux/components/helm/helmreleases/#chart-reference) to the version: `<version>+<digest[0:12]>`.

### DIFF
--- a/helm/dagger/templates/_helpers.tpl
+++ b/helm/dagger/templates/_helpers.tpl
@@ -36,7 +36,7 @@ Common labels
 {{- define "dagger.labels" -}}
 helm.sh/chart: {{ include "dagger.chart" . }}
 {{ include "dagger.selectorLabels" . }}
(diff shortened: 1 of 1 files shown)


shykes (approved):

gerhard (approved):
This is a sensible change, and all Helm tests are still green 🚀

Thank you @b4nst 👍

b4nst:
I've just confirmed it resolved the issue. I'm able to deploy with 0.16.2. Thanks!

PR #9681, opened 2025-02-21
error out in dagger query/listen if dagger.json totally invalid

sipsma:
The fallback behavior of these commands when failing to load a module meant that even if a dagger.json was found but completely invalid, we silently went ahead and skipped the module load. The intention was to only fallback when no module was found though.

Those commands now error out as expected in that case.

### DIFF
--- a/core/integration/module_cli_test.go
+++ b/core/integration/module_cli_test.go
@@ -1645,3 +1645,23 @@ func (CLISuite) TestDaggerUpdate(ctx context.Context, t *testctx.T) {
 		})
 	}
 }
+
+func (CLISuite) TestInvalidModule(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	t.Run("normal context dir", func(ctx context.Context, t *testctx.T) {
+		modGen := goGitBase(t, c).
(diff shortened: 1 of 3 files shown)


core/integration/module_cli_test.go
```diff
@@ -1645,3 +1645,23 @@ func (CLISuite) TestDaggerUpdate(ctx context.Context, t *testctx.T) {
...
+		requireErrOut(t, err, `failed to check if module exists`)
+	})
+
+	t.Run("fallback context dir", func(ctx context.Context, t *testctx.T) {
+		modGen := goGitBase(t, c).
+			WithNewFile("dagger.json", `{"name": "broke", "engineVersion": "v100.0.0", "sdk": 666}`)
+
+		_, err := modGen.With(daggerQuery(`{version}`)).Stdout(ctx)
+		requireErrOut(t, err, `failed to check if module exists`)
+	})
```
helderco, line 1666: These subtests have the same name 🙂

sipsma: Doh, thank you, fixed

jedevc (approved):

PR #9685, opened 2025-02-22
ci: Upgrade macOS runners to latest ARM ones

gerhard:
We expect better performance


sipsma:
I looked into this the other day but unfortunately the current macos arm runners don't support nested virt (need m3 or above), so I don't think this will work at this time https://github.com/douglascamata/setup-docker-macos-action

gerhard:
I was just digging through that history @sipsma, great timing!

Closing for now, leaving the context with @nabsul 💪

PR #9686, opened 2025-02-23
fix(sdk/java): dependencies to build should be outside of src

eunomie:
When a Java module is initialized with no subdirectory (for instance `dagger init --sdk=java my-module`) some temporary files will be copied to the user directory because they are under the `ModSourceDirPath` (`/src`).
Those files are a temporary copy of the `sdk/java`. From there the different files will be generated, built, packaged, but we don't want them to be exported to the user.

This change simply move the temporary copy of `sdk/java` out of `/src`, this ensures those files will be out of the exported ones.

---

Current (`main`) status:

```console
$ dagger init --sdk=java my-module
$ cd my-module
$ ls
LICENSE     dagger-io   dagger.json pom.xml     src         target
$ ls dagger-io
LICENSE                          dagger-codegen-maven-plugin      dagger-java-samples              pom.xml
README.md                        dagger-java-annotation-processor dagger-java-sdk
```

Expected:

```console
$ dagger init --sdk=java my-module
$ cd my-module
$ ls
LICENSE     dagger.json pom.xml     src         target
```

### DIFF
--- a/sdk/java/runtime/main.go
+++ b/sdk/java/runtime/main.go
@@ -21,7 +21,7 @@ const (
 
 	ModSourceDirPath = "/src"
 	ModDirPath       = "/opt/module"
-	GenPath          = "dagger-io"
+	GenPath          = "/dagger-io"
 )
 
 type JavaSdk struct {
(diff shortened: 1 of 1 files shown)


jpadams (approved):
LGTM

eunomie:
Failed jobs related to PHP, merging the fix.

PR #9693, opened 2025-02-24
chore(sdk/java): bump maven/java images

eunomie:
Align `.dagger/sdk_java` with `sdk/java`.
Fix disable SVE on arm64 machines so we can use java 23 everywhere (align maven and jre images).
Update maven and jre images, smaller, alpine based, less vulnerabilities.

### DIFF
--- a/.dagger/sdk_java.go
+++ b/.dagger/sdk_java.go
@@ -15,8 +15,8 @@ const (
 	javaSDKVersionPomPath   = javaSDKPath + "/pom.xml"
 	javaSchemasDirPath      = javaSDKPath + "/dagger-codegen-maven-plugin/src/main/resources/schemas"
 	javaGeneratedSchemaPath = "target/generated-schema/schema.json"
-	mavenImage              = "maven:3.9.9-eclipse-temurin-23-alpine"
-	mavenDigest             = "sha256:0e5e89100c3c1a0841ff67e0c1632b9b983e94ee5a9b1f758125d9e43c66856f"
+	mavenImage              = "maven:3.9.9-eclipse-temurin-21-alpine"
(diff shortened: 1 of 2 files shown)


.dagger/sdk_java.go
```diff
@@ -15,8 +15,8 @@ const (
 	javaSDKVersionPomPath   = javaSDKPath + "/pom.xml"
 	javaSchemasDirPath      = javaSDKPath + "/dagger-codegen-maven-plugin/src/main/resources/schemas"
 	javaGeneratedSchemaPath = "target/generated-schema/schema.json"
-	javaVersion             = "17"
-	mavenVersion            = "3.9"
+	mavenImage              = "maven:3.9.9-eclipse-temurin-23-alpine"
```
jcsirot, line 18: is there any reason to use java 23 rather than java 21 which is the latest LTS version?

jcsirot (commented):
LGTM

sdk/java/runtime/main.go
```diff
@@ -298,23 +298,28 @@ func (m *JavaSdk) finalJar(
...
-		ctr = ctr.WithEnvVariable("JAVA_OPTS", "-XX:UseSVE=0")
-	}
+		From(fmt.Sprintf("%s@%s", JavaImage, JavaDigest)).
+		With(disableSVEOnArm64(ctx))
 	return ctr
 }
 
+func disableSVEOnArm64(ctx context.Context) dagger.WithContainerFunc {
+	return func(ctr *dagger.Container) *dagger.Container {
+		if platform, err := ctr.Platform(ctx); err == nil && strings.Contains(string(platform), "arm64") {
```
jedevc, line 316: `err` could be a lot of things here, ideally we shouldn't silently fail.

eunomie:
Also moved to java 21 that is the current LTS

jedevc (approved):

PR #9697, opened 2025-02-24
fix: move fix for context transfer secret

jedevc:
Looks like a recent refactor moved this - it *should* be for local sources, as well as dir sources.

Re-applied from #9530.

> [!WARNING]
>
> This *absolutely* needs a test - sloppy work from me, since I didn't do that originally - would have avoided this lol.

### DIFF
--- a/core/modulesource.go
+++ b/core/modulesource.go
@@ -333,6 +333,14 @@ func (src *ModuleSource) LoadContext(
 			return inst, fmt.Errorf("failed to select directory: %w", err)
 		}
 
+		mainClientCallerID, err := src.Query.MainClientCallerID(ctx)
+		if err != nil {
+			return inst, fmt.Errorf("failed to retrieve mainClientCallerID: %w", err)
+		}
+		if err := src.Query.AddClientResourcesFromID(ctx, &resource.ID{ID: *inst.ID()}, mainClientCallerID, false); err != nil {
+			return inst, fmt.Errorf("failed to add client resources from ID: %w", err)
+		}
+
 		return inst, nil
 
 	case ModuleSourceKindGit:


core/contenthash.go
```diff
@@ -103,7 +103,7 @@ func GetContentHashFromDef(
 		)
 		defer telemetry.End(span, func() error { return rerr })
 
-		dgst, err := bkcontenthash.Checksum(ctx, ref, "/", bkcontenthash.ChecksumOpts{}, nil)
+		dgst, err := bkcontenthash.Checksum(ctx, ref, subdir, bkcontenthash.ChecksumOpts{}, nil)
```
jedevc, line 106: :eyes: @sipsma looks like this was wrong, only discovered this when writing tests that accessed different sub-directories from within the same client+session.

Kinda surprised we haven't hit this anywhere else :thinking: are we hitting this code path as much as we should be?

core/integration/module_test.go
```diff
@@ -4696,6 +4696,61 @@ func (m *Test) GetRelDepSource(ctx context.Context, src *dagger.Directory) (*dag
...
+				// c := connect(ctx, t)
+				devEngine := devEngineContainerAsService(devEngineContainer(c))
+
+				modRef := testGitModuleRef(tc, mod)
+				modGen := goGitBase(t, c).
+					WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+					WithServiceBinding("dev-engine", devEngine).
+					WithEnvVariable("_EXPERIMENTAL_DAGGER_RUNNER_HOST", "tcp://dev-engine:1234").
+					WithWorkdir("/work").
+					With(mountedSocket)
```
jedevc, line 4717: NOTE: this is *not* right. But for some reason, the git credentials created by `privateRepoSetup` aren't properly propagated in nested sessions.

core/contenthash.go
sipsma: Yeah that looks like a typo, but probably didn't come up because until very recently I think all the uses of it ended up providing subdir as `/`. I could be missing something but from a quick glance I think the only situation now where subdir isn't `/` is when using module sources of `dir` kind, which is sort of obscure too.

core/modulesource.go
```diff
@@ -417,19 +418,24 @@ func (src *ModuleSource) LoadContext(
...
-			return inst, fmt.Errorf("failed to add client resources from ID: %w", err)
-		}
-
-		return MakeDirectoryContentHashed(ctx, bk, ctxDir)
 
 	default:
 		return inst, fmt.Errorf("unsupported module src kind: %q", src.Kind)
 	}
+
+	mainClientCallerID, err := src.Query.MainClientCallerID(ctx)
```
sipsma, line 430: I didn't look deeply yet, so take with salt, but my first impression is that this should be using `NonModuleParentClientMetadata` rather than getting the `MainClientCallerID` (similar to local context loads above in this func). Wondering if this could explain the problem you mentioned here: https://github.com/dagger/dagger/pull/9697#discussion_r1970263023

jedevc: Ohhh. Yeah. That does sounds right, thank you!

core/integration/module_test.go
```diff
@@ -5396,19 +5431,37 @@ func daggerCallAt(modPath string, args ...string) dagger.WithContainerFunc {
...
+		require.NoError(t, err)
+		token = strings.TrimSpace(string(decoded))
+	}
 
 	return func(ctr *dagger.Container) *dagger.Container {
-		sock := c.Host().UnixSocket(sockPath)
-		if sock != nil {
-			// Ensure that HOME env var is set, to ensure homePath expension in test suite
-			homeDir, _ := os.UserHomeDir()
-			ctr = ctr.WithEnvVariable("HOME", homeDir)
```
jedevc, line 5407: I'm guessing from here https://github.com/dagger/dagger/pull/8365#discussion_r1762068678, that this isn't required? `SSH_AUTH_SOCK` doesn't contain `~`.

Additionally this seems to make these tests not work when running them as a normal user (e.g. using `./hack/with-dev go test ...`)

core/modulesource.go
jedevc: Okay, it definitely *should* be `NonModuleParentClientMetadata`.

But it probably should also be that here: https://github.com/dagger/dagger/blob/3b2687279b3e53798bbb6117815c40580c1b1ebc/engine/buildkit/client.go#L629-L639

Making that change, I can get the tests to pass a bit better, now instead of failing to find the credentials, I just fail to transfer the secrets between sessions! :tada:

---

This now presents a new problem :thinking: In the tests, we load the module multiple times by using `dagger call -m <modRef>` multiple times. Because it's all the same session, we only load it once - but the problem is, loading uses `SetSecret` as a side-effect, which is then required to call `LoadContext` later.

This is *reasonably* annoying. Not really sure what the right thing here is, but I've pushed a commit that disables some module loading caching, and it kind of "works". That said, there's definitely a smarter thing to do here, right?

My best idea is currently, attach the secret to the root session, then, as a "post call" (as in #9653) transfer if to the non-module parent. We can't just transfer between the non-module parents, since they're all sibling calls, no return values / args are passed between them.

@sipsma does that *sort* of make sense? It doesn't seem like the post call type slips in super neatly as a return value to a schema handler, so I'm a little skeptical.

sipsma: Yeah the problem makes sense. I had to pull and try locally to fully grok and in the process managed to fix it using PostCalls, so I pushed that as a commit here: https://github.com/dagger/dagger/pull/9697/commits/bdfcc2b4cfce7490bdc832b352368001629c31bd

It needs a little cleanup mainly because of the stuff required to convince the go compiler to accept the right type signatures and handle `PostCallTyped` return vals while retaining the right schema. I will push a cleanup for that too, just want to double check nothing else broke in CI (just checked `TestContextDirectoryGit` locally so far).

The fix pretty much just uses PostCall to run secret transfers for `git` and `moduleSource` too. I think we'll need to find some general way of dealing with this rather than whacking each mole individually like this, but this fix seems like it should hold us over for now if nothing else.

sipsma: Pushed a follow-up cleanup commit https://github.com/dagger/dagger/pull/9697/commits/133b1e24c4efb675c73e77606b27207e736bb99f, was able to avoid a lot of the nonsense copypasta from the first one

sipsma (approved):

core/modulesource.go
jedevc: Eyy thanks :tada: solution looks really good :pray: 

Would we want to maybe wrap the `PostCallTyped` with a generic so we could get some typing there? :thinking: not a blocker, so merged it!

PR #9700, opened 2025-02-24
mitigate upstream otel goroutine leak in telemetry flush

sipsma:
There's a problem upstream where a canceled context during `batchSpanProcessor.ForceFlush` leaks a goroutine. I saw this actually happen while testing stuff and it's extra bad because it leaks quite a substantial amount of memory that's indirectly referenced in our various span processors.

For now, this just mitigates it. I have [an upstream fix](https://github.com/open-telemetry/opentelemetry-go/pull/6363) that we will hopefully pick up someday after it's released and we've upgraded.

cc @vito I don't really love the mitigation of just giving this a long timeout, but not sure what's better, let me know if you can think of anything better.

### DIFF
--- a/engine/server/session.go
+++ b/engine/server/session.go
@@ -176,6 +176,13 @@ func (client *daggerClient) FlushTelemetry(ctx context.Context) error {
 	var errs error
 	if client.tracerProvider != nil {
 		slog.ExtraDebug("force flushing client traces")
+		// FIXME: mitigation for goroutine leak fixed upstream in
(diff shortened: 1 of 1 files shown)


vito (approved):
Nice catch, workaround seems reasonable 👍

sipsma:
GHA claims this is in progress but it succeeded a while ago... https://github.com/dagger/dagger/actions/runs/13507554558/job/37740539729?pr=9700 pressing merge

PR #9702, opened 2025-02-25
LLM: Gemini client

kpenfound:
Adds Gemini client


kpenfound:
Draft until I fix tool calling. Currently getting a 400 from the Google API

kpenfound:
Works!
image

Get a `GEMINI_API_KEY` and try it out. Prompts seem to need to really push it. I can test more prompting now

kpenfound:
Looking at that screenshot it might be sending the response back as another user prompt... double checking on that

kpenfound:
Mostly successfully writes curl clone (it messed up an import and did an exec instead of check)
Screenshot 2025-02-25 at 3 07 24 PM

core/llm_google.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+	}
+
+	return &LLMResponse{
+		Content:   content,
+		ToolCalls: toolCalls,
+	}, nil
+}
+
+// TODO: this definitely needs a unit test
```
kpenfound, line 175: This does need a test. Our bbi `map[string]interface{}` schema doesn't cast directly to genai's schema like it does for openai and anthropic

PR #9663, opened 2025-02-21
docs(sdk/java): ide integration and module structure

eunomie:
Adds IDE integration and module structure documentation for Java SDK

PR #9707, opened 2025-02-25
fix(sdk/java): code completion on VS Code

eunomie:
The VS Code java plugin is missing some features for years: the ability to code complete based on the target/generated-sources files. This allows to explicitly instruct maven/vs code by adding the generated source directories as source directories so that code completion can happen.

Also added some comments to the pom.xml file to help understand which are the critical parts of the file that shouldn't be touched by the user.

### DIFF
--- a/sdk/java/runtime/template/pom.xml
+++ b/sdk/java/runtime/template/pom.xml
@@ -63,6 +63,10 @@
                     <version>3.3.1</version>
                 </plugin>
                 <plugin>
+                    <!--
+                    This plugin configuration is required by Dagger to generate the module.
+                    Please do not remove or modify it.
+                    -->
                     <artifactId>maven-compiler-plugin</artifactId>
                     <version>3.13.0</version>
                     <configuration>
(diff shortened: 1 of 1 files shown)


eunomie:
cc @vikram-dagger the comments added to the `pom.xml` file are related to https://github.com/dagger/dagger/pull/9663#discussion_r1965513404 in order to make it clearer some parts should not be modified.

sdk/java/runtime/template/pom.xml
```diff
@@ -108,6 +112,36 @@
...
+                the generated code created by `dagger init` and `dagger develop` and
+                helps for code completion.
+                -->
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <version>3.3.0</version>
+                <executions>
+                    <execution>
+                        <id>add-generated-sources</id>
+                        <phase>generate-sources</phase>
```
jcsirot, line 126: this line is not required since `generate-sources` is the build-helper-maven-plugin default phase

eunomie: I removed it, thanks

jcsirot (commented):
LGTM

jedevc (approved):

PR #9711, opened 2025-02-25
fix(sdk/elixir): fix sdk crash due because of String.to_existing_atom

wingyplus:
Instead of convert input arguments from `FunctionCallArgValue` to atom.
Convert argument name in argument definition to string instead.


jedevc (approved):
LGTM :tada:

Updated to remove the skipped tests, since they'll start passing once this is merged on main!

PR #9719, opened 2025-02-26
replace impurity with mix of custom cache keys + field spec settings

sipsma:
This is a mix of smallish cleanups/preparation for more changes:

Impure is gone now and replaced with:
* `CachePerCall` - a custom cache key impl that lets the API always run when called but be cached after the result of that call is passed around (i.e. for snapshots of mutating state)
* `DoNotCache` - a setting that exists purely on the Field now rather than being tied into IDs. We still want to straight-up skip caching values for various reasons (correctness and memory usage), but there's no longer any need for that to be part of the ID. This replaces it.

Related to the above, we no longer cache the results of selecting "trivial fields" (i.e. selecting a field already set on a parent object, as opposed to actually executing a an API impl for the field). We skip it via `DoNotCache`.
* The reason for this is that it has O(n^2) memory usage for nested struct cases. Selecting a series of nested fields like `Obj.field1.field2.field3` results in progressively caching each object in the chain even though you already have everything cached you need in the "outer" one.
* This started mattering a lot more in my trials at cross-session caching, so want to get it fixed now in prep for that.

A few other related things are here too, i.e. removing code referenced to `Meta` as a field/id configuration, which hasn't been used for a while.


core/schema/module.go
```diff
@@ -37,26 +37,23 @@ func (s *moduleSchema) Install() {
...
 
-		dagql.FuncWithCacheKey("currentModule", s.currentModule, core.CachePerClient).
+		dagql.FuncWithCacheKey("currentModule", s.currentModule, dagql.CachePerClient).
 			Doc(`The module currently being served in the session, if any.`),
 
-		dagql.Func("currentTypeDefs", s.currentTypeDefs).
-			// Impure for now, could use a finer grain cache key if we had the ability to mix
-			// a digest of the dagql server schema into the cache key.
-			Impure("Can change when modules are loaded into the schema.").
+		dagql.FuncWithCacheKey("currentTypeDefs", s.currentTypeDefs, dagql.CachePerCall).
```
vito, line 43: neat - though, not sure about the name `CachePerCall` 🤔 - initially it sounded like the opposite of what it does (the call itself isn't cached since you get a different result each time, but each result is cached)

that said, don't know what else to suggest.

dagql/server.go
```diff
@@ -273,7 +272,9 @@ func (s *Server) installObject(class ObjectType) {
 				}
 				return res, nil
 			},
-			nil,
+			CacheSpec{
+				DoNotCache: "The given ID ultimately determines the purity of its result.",
```
vito, line 276: (copying here too)

```suggestion
				DoNotCache: "There's no point caching the loading call of an ID, vs. letting the ID's calls cache on their own.",
```

vito (approved):
LGTM, only minor suggestions

dagql/cachekey.go
```diff
@@ -0,0 +1,70 @@
...
+// For instance, the API may return a snapshot of some live mutating state; in that case the first call to get the snapshot
+// should always run but if the returned object is passed around it should continue to be that snapshot rather than the API
+// always re-running.
+func CachePerCall[P Typed, A any](
+	_ context.Context,
+	_ Instance[P],
+	_ A,
+	cacheCfg CacheConfig,
+) (*CacheConfig, error) {
+	randID := identity.NewID()
```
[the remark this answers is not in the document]
jedevc: Best to hold that now, think we need to downgrade due to #9759.

dagql/call/id.go
```diff
@@ -300,24 +279,18 @@ func (id *ID) Append(
 // given metadata changed.
 // customDigest, if not empty string, will become the ID's digest.
 // tainted sets the ID's tainted flag.
-func (id *ID) WithMetadata(customDigest digest.Digest, tainted bool) *ID {
+func (id *ID) WithMetadata(customDigest digest.Digest) *ID {
```
jedevc, line 282: suggestion: we can just rename this to `WithDigest` now (and remove `tainted` from the doc string)

dagql/objects.go
```diff
@@ -322,15 +318,14 @@ func (r Instance[T]) String() string {
...
 
 // WithMetadata returns an updated instance with the given metadata set.
-// isPure changes the purity of the instance.
 // customDigest overrides the default digest of the instance to the provided value.
 // NOTE: customDigest must be used with care as any instances with the same digest
 // will be considered equivalent and can thus replace each other in the cache.
 // Generally, customDigest should be used when there's a content-based digest available
 // that won't be caputured by the default, call-chain derived digest.
-func (r Instance[T]) WithMetadata(customDigest digest.Digest, isPure bool) Instance[T] {
+func (r Instance[T]) WithMetadata(customDigest digest.Digest) Instance[T] {
```
jedevc, line 326: suggestion: `WithDigest` maybe as well?

dagql/objects.go
```diff
@@ -963,11 +949,31 @@ func (fields Fields[T]) findOrInitializeType(server *Server, typeName string) Cl
...
+	// calls deduped. The string value is a reason why the field should not be cached.
+	DoNotCache string
+}
+
+type GenericGetCacheConfigFunc func(context.Context, Object, map[string]Input, CacheConfig) (*CacheConfig, error)
+
+type GetCacheConfigFunc[T Typed, A any] func(context.Context, Instance[T], A, CacheConfig) (*CacheConfig, error)
+
+// CacheConfig is the configuration for caching a field. Currently just custom digest
+// but intended to support more in time (TTL, etc).
```
jedevc, line 967: :eyes: oh this is *coooooool* :tada:

dagql/testdata/introspection.json
```diff
@@ -636,17 +636,7 @@
...
-              {
-                "name": "impure",
-                "args": [
-                  {
-                    "name": "reason",
-                    "value": "Because I said so."
-                  }
-                ]
-              }
-            ]
```
jedevc, line 649: I never saw these used in the first place (so seems good to remove them), but also, mildly curious as to why we had an `impure` directive at all :thinking:

core/schema/query.go
```diff
@@ -40,8 +40,7 @@ func (s *querySchema) Install() {
 		// JSON and written to a core.File. This is currently used internally for calling
 		// module SDKs and is thus hidden the same way the rest of introspection is hidden
 		// (via the magic __ prefix).
-		dagql.NodeFunc("__schemaJSONFile", s.schemaJSONFile).
-			Impure("Return value depends on what the caller has installed in the server.").
+		dagql.NodeFuncWithCacheKey("__schemaJSONFile", s.schemaJSONFile, dagql.CachePerCall).
```
jedevc, line 43: As an idea for how we could cache these in the future ... everytime we modify client state, we could increase a client counter, and then mixin that counter + the client id to the target.

Maybe you already have ideas for how to do this though!

dagql/objects.go
```diff
@@ -421,28 +416,33 @@ func (r Instance[T]) preselect(ctx context.Context, sel Selector) (map[string]In
...
-		tainted,
 		sel.Nth,
 		"",
 		idArgs...,
 	)
 
-	if field.CacheKeyFunc != nil {
-		customDgst, err := field.CacheKeyFunc(ctx, r, inputArgs, newID.Digest())
+	doNotCache := field.CacheSpec.DoNotCache != ""
+	if field.CacheSpec.GetCacheConfig != nil {
```
jedevc, line 430: Question - why is `DoNotCache` is called as a chain function, instead of using a function similar to `CachePerClient`. Since it's set in `CacheConfig` anyways.

It feels a bit confusing that there's a couple of ways to control caching now, I'm not sure if this was just to make porting easier, or if there's a fundamental reason why we can't?

jedevc (approved):

dagql/testdata/introspection.json
vito: No concrete reason, it was just interesting to surface at the schema layer because it could affect how things use those fields. For example, the CLI used to fetch object fields and print them; it could avoid fetching `@impure` fields as a safety precaution. But now that feature's gone, and it never used it to begin with.

dagql/objects.go
sipsma: Yeah the main issue right now is that we purposely *don't* re-call custom cache key funcs when loading an ID that was given to us (as opposed to making a new call). The idea there is that if we're given an ID, it's from a call already made and should just re-use the digest given to us. This is currently important for `CachePerCall` to work.

So that also means at the moment that `DoNotCache` has to be set on the field's definition rather than set by a custom cache key func.

I am not a huge fan of this behavior for the long term, but it works at the moment. We could change it by adding some ability to set whether a digest should be re-calculated on ID load too, but I couldn't find a clean way to do that so just wanted to go with this for now.

I'm figuring the best time to sort this out will be when moving to the sqlite db. That whole effort is gonna entail all sorts of other new features around mapping between (multiple) digests and results anyways.

core/schema/query.go
sipsma: Yeah that's not a bad idea. I was figuring when a schema is installed/served we could calculate a digest based on which modules are installed and use that as a content-based digest for this, which *technically* is slightly better in that we could re-use the same cache for this file when two clients end up with the exact same schema, but practically speaking either approach works and is mostly the same.

core/schema/module.go
sipsma: Yeah it's a very subtle thing that's hard to summarize in a short name... Nothing better in terms of names is coming to mind for me either, but if I think of anything before merge I'll update it.

PR #9739, opened 2025-02-28
tui: look up calls by output digest, too

vito:
Currently for uploaded local dirs the UI shows the raw digest or a placeholder instead of the Host.directory call that created it.

image

This happens because the server's `Host.directory` actually returns a call with a custom content digest of the blob, rather than the call digest of the `Host.directory` call.

Now when querying the DB for a call we'll also look for spans that had the digest as an output:

image


dagql/dagui/db.go
```diff
@@ -749,6 +758,15 @@ func (db *DB) Call(dig string) *callpbv1.Call {
 	}
 	callPayload, ok := db.CallPayloads[dig]
 	if !ok {
+		creators, ok := db.CreatorSpans[dig]
```
sipsma, line 761: very nit, just if you end up having to do more pushes anyways: this seems like a case where it'd be easier to follow the logic if you inverted to an early return case

```go
if ok {
   // return it
}

// fallback looking for creators logic
```

sipsma (approved):

dagql/dagui/db.go
vito: good point! just pushed a refactor

jedevc (approved):

PR #6262, opened 2023-12-13
docs: Restructures Zenith quickstart

vikram-dagger:
The current Zenith quickstart uses a pre-built module and focuses on teaching the user the various sub-commands/features of the Dagger CLI. While this is a good learning path for the CLI itself, it does not do Dagger justice as it fails to capture the power of creating reusable modules and leveraging the Daggerverse. It is also less impactful than the pre-Zenith quickstart, which provides a more hands-on learning experience by having the user build a new pipeline from scratch.

This commit restructures and rewrites the Zenith quickstart to attempt to resolve the above shortcomings. It provides a hands-on experience of creating a Daggerized pipeline using modules from the Daggerverse, whilst also walking the user through a few important CLI sub-commands.

PR #9743, opened 2025-03-01
docs: improve agent dev version install instructions

marcosnils:

### DIFF
--- a/docs/current_docs/agents.mdx
+++ b/docs/current_docs/agents.mdx
@@ -65,7 +65,7 @@ curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.17.0-llm.4 
 If you want to install globally, and sudo is available, you can specify an alternative install location by setting BIN_DIR and using sudo -E:
 
 ```console
-curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.17.0-llm.4 BIN_DIR=/usr/local/bin SUDO -E sh
+curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.17.0-llm.4 BIN_DIR=/usr/local/bin sudo -E sh
 ```
 
 You can adjust `BIN_DIR` to customize where the `dagger` CLI is installed.
--- a/docs/current_docs/agents.mdx
+++ b/docs/current_docs/agents.mdx
@@ -59,7 +59,13 @@ Once this feature is merged (current target is 0.17), a development build will n
 Install the development version of LLM-enabled Dagger:
 
 ```console
-curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.17.0-llm.4 BIN_DIR=/usr/local/bin sh
+curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.17.0-llm.4 BIN_DIR=$HOME/.local/bin sh
+```
+
+If you want to install globally, and sudo is available, you can specify an alternative install location by setting BIN_DIR and using sudo -E:
+
+```console
+curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.17.0-llm.4 BIN_DIR=/usr/local/bin SUDO -E sh
 ```
 
 You can adjust `BIN_DIR` to customize where the `dagger` CLI is installed.


jpadams:
https://specifications.freedesktop.org/basedir-spec/latest/#:~:text=User%2Dspecific%20executable%20files%20may%20be%20stored,environment%20variable%2C%20at%20an%20appropriate%20place.

docs/current_docs/agents.mdx
```diff
@@ -59,7 +59,13 @@ Once this feature is merged (current target is 0.17), a development build will n
...
 
 ```console
-curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.17.0-llm.4 BIN_DIR=/usr/local/bin sh
+curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.17.0-llm.4 BIN_DIR=$HOME/.local/bin sh
+```
+
+If you want to install globally, and sudo is available, you can specify an alternative install location by setting BIN_DIR and using sudo -E:
+
+```console
+curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.17.0-llm.4 BIN_DIR=/usr/local/bin SUDO -E sh
```
vito, line 68:
```suggestion
curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=0.17.0-llm.4 BIN_DIR=/usr/local/bin sudo -E sh
```

vito, line 68: though, imo this can be omitted, I think it's obvious enough from seeing `BIN_DIR` above, plus the sentence below ("You can adjust BIN_DIR ..."):

```suggestion
```

piping to `sudo bash` spooks some folks out so maybe it's best to leave that exercise to the reader 😆

vito (approved):
had a couple suggestions, but approving so it's unblocked

shykes:
I prefer default to global install like it is today. Otherwise it may not work because `dagger` won't be in your PATH.

marcosnils:
> I prefer default to global install like it is today. Otherwise it may not work because `dagger` won't be in your PATH.

 I also prefer the global approach but that means that you need to run the `curl | bash` command either with `root` or `sudo` which as you know, has been a topic for debate for a while. I've mostly used the same patterns we're currently highlighting in the non-agent installation page here. https://docs.dagger.io/install

marcosnils:
> https://xdgbasedirectoryspecification.com/
> 
> https://specifications.freedesktop.org/basedir-spec/latest/#:~:text=User%2Dspecific%20executable%20files%20may%20be%20stored,environment%20variable%2C%20at%20an%20appropriate%20place. image

 @jpadams I'm following the same pattern we already have in the current installation page here https://docs.dagger.io/install

jpadams:
> > https://xdgbasedirectoryspecification.com/
> > https://specifications.freedesktop.org/basedir-spec/latest/#:~:text=User%2Dspecific%20executable%20files%20may%20be%20stored,environment%20variable%2C%20at%20an%20appropriate%20place. image
> 
> @jpadams I'm following the same pattern we already have in the current installation page here https://docs.dagger.io/install

Yep. I've seen that change subtly over time. I'm going to look at why.

It used to just go in `/usr/local/bin` not that long ago I think, which was in more PATHs

vikram-dagger:
> > > https://xdgbasedirectoryspecification.com/
> > > https://specifications.freedesktop.org/basedir-spec/latest/#:~:text=User%2Dspecific%20executable%20files%20may%20be%20stored,environment%20variable%2C%20at%20an%20appropriate%20place. image
> > 
> > 
> > @jpadams I'm following the same pattern we already have in the current installation page here https://docs.dagger.io/install
> 
> Yep. I've seen that change subtly over time. I'm going to look at why.
> 
> It used to just go in `/usr/local/bin` not that long ago I think, which was in more PATHs

Adding @gerhard for opinion

gerhard (approved):
This is a good change.

+1 for `~/.local/bin` . This [commit message](https://github.com/dagger/dagger/pull/6262/commits/ab0a8019dc29958fa4063c264edeba1621ad7019) is still relevant today: 

image

shykes:
I didn't know `$HOME/.local/bin` was a documented standard, that's cool. But it doesn't work out of the box on Mac OS. So it's a bad out of the box install experience for Mac users

Let's follow these principles:

1. Consistency between AI and CI install instructions. Soon the llm branch will be merged, and it will be the same install instructions for everyone. So we should aim to converge them ASAP
2. NO FRICTION. Install instructions should work as smoothly as possible, for as many users as possible. Obeying abstract rules and standards is nice but strictly less important. For example "don't write on /usr/local/bin" -> ok but what's the alternative? if it's "write to a directory that may or may not be on the user PATH" then nope that's not a good alternative. Etc.

marcosnils:
> * Consistency between AI and CI install instructions. Soon the llm branch will be merged, and it will be the same install instructions for everyone. So we should aim to converge them ASAP

That was my intention with this PR. I understand that it doesn't meet point #2 so IIUC, the solution here would be to find a way where we can satisfy both points. IMO I think we should keep this PR for the `linux` installation and if possible, add the ability to publish the `llm` binary to `brew` also given that `brew install` would be the recommended "NO FRICTION" approach without having to go the `dmg` packaging route which is a nightmare. 

So to summarize, here's what I propose: 

-  Allow the `llm` branch binaries to be installed from brew
-  Slightly change this PR to modify the `agent` install instructions so they follow the same per-os pattern we have in the CI installation page. 

WDYT?

jedevc:
Closing this out! Since with the next v0.17.0 release, `llm` support will be part of our regular releases.

PR #9746, opened 2025-03-03
docs: minimal header

jpadams:
image

Just down to `Daggerverse | Docs`

Request from @mircubed


vikram-dagger (approved):

vikram-dagger:
We could also discuss if we still need "Docs" in the header, because "Introduction" in the sidebar also points to the docs index page and is always visible.

PR #9751, opened 2025-03-03
docs: Remove Java integrations page, update redirections

vikram-dagger:


docs/netlify.toml
```diff
@@ -416,12 +416,12 @@
 
 [[redirects]]
   from = "/manuals/administrator/ci"
-  to = "/integrations"
+  to = "/ci/integrations"
```
jpadams, line 419: The global /integrations -> /ci/integrations redirect covers this today, but makes sense to collapse that chain.

jpadams (approved):
👍

PR #9758, opened 2025-03-03
feat: add persistency for generated clients

TomChv:
Update `generateClient` to only set fields in the
configuration so now the generation
logic happens in `generatedContextDirectory`.
That way, it works with `dagger develop`.

Also add persistency in `dagger.json` for generated clients.

```json
{
  "clients": [
    {
       "generator": "go"
       "directory": "daggger"
       "dev": true # Omit if not set
    }
  ]
}
````

Correctly use contextDirectory in codegen instead
of only the source directory.

Next step after this PR is merged is changes around the `GenerateClient` interface to:
1. Remove `introspectionJSON` arg in favor of an API call
2. Rename `localSDK` into `dev` (might be able to work around this one with an API call too)

### CODE REVIEW

File: cmd/dagger/module.go [hunk 1]
```diff
@@ -539,11 +539,6 @@ This command is idempotent: you can run it at any time, any number of times. It
 				developSourcePath = filepath.Join(srcRootAbsPath, inferredSourcePath)
 			}
 
-			// if there's no SDK and the user isn't changing the source path, there's nothing to do.
-			// error out rather than silently doing nothing.
-			if modSDK == "" && developSourcePath == "" {
-				return fmt.Errorf("dagger develop on a module without an SDK requires either --sdk or --source")
-			}
```
jedevc, line 546: We should keep this error. If there are no clients *or* sdk then we still need it.

core/integration/client_generator_test.go
```diff
@@ -494,6 +496,267 @@ main()
 	})
 }
 
+func (ClientGeneratorTest) TestPersistency(ctx context.Context, t *testctx.T) {
```
jedevc, line 499: nit: 
```suggestion
func (ClientGeneratorTest) TestPersistence(ctx context.Context, t *testctx.T) {
```

core/modules/config.go
```diff
@@ -233,3 +236,14 @@ func (cfg ModuleCodegenConfig) Clone() *ModuleCodegenConfig {
...
+
+type ModuleConfigClient struct {
+	// The generator the client uses to be generated.
+	Generator string `json:"generator"`
+
+	// The directory the client is generated in.
+	Directory string `json:"directory"`
+
+	// Whether the client is generated using the local SDK.
+	LocalLibrary *bool `json:"localLibrary,omitempty"`
```
jedevc, line 248: Probably worth more comments on all these one *explaining* what they mean - since these turns up as user-facing docs.

jedevc: Also, are these final? These are incredibly fiddly to change later if we do backwards compat, etc. I'd like to avoid changing them too much, so let's check in with @shykes on the design here.

core/schema/modulesource.go
```diff
@@ -1936,6 +1938,95 @@ func (s *moduleSourceSchema) moduleSourceGeneratedContextDirectory(
 		}
 	}
 
+	// Generate clients
```
jedevc, line 1941: This entire function definitely needs to be split down if that's at all possible - it's getting *very* long.

cmd/codegen/generator/go/generator.go
```diff
@@ -158,7 +158,8 @@ func (g *GoGenerator) GenerateClient(ctx context.Context, schema *introspection.
 		)
 
 		// Get the go package from the module
-		pkg, _, err := loadPackage(ctx, "/module")
+		// We assume that we'll be located at the root source directory
+		pkg, _, err := loadPackage(ctx, ".")
```
jedevc, line 162: :eyes: do we not need to set a `WithWorkdir` correspondingly in `GenerateClient` as well then?

jedevc:
Looks mostly good, but given this is very user facing (since there are config `dagger.json` changes now too), we should confirm the design here before merging.

Also, for regenerating during `dagger develop`. *I'm really a fan of this* - I think having all the codegen done through that is neat. However, I know we've previously discussed with @shykes breaking it down a bit more - so worth making sure that this is *definitely* the right place to do it!

cmd/codegen/generator/go/generator.go
TomChv: We do :)

```
		dagql.Selector{
			Field: "withWorkdir",
			Args: []dagql.NamedInput{
				{
					Name:  "path",
					Value: dagql.NewString(filepath.Join(goSDKUserModContextDirPath, rootSourcePath)),
				},
			},
		},
```

core/schema/modulesource.go
```diff
@@ -1936,6 +1938,95 @@ func (s *moduleSourceSchema) moduleSourceGeneratedContextDirectory(
...
+			if err != nil {
+				return genDirInst, fmt.Errorf("failed to transform module source into module: %w", err)
+			}
+
+			deps = mod.Self.Deps.Append(mod.Self)
+		}
+
+		useLocalSDK := dagql.Boolean(false)
+		if client.LocalLibrary != nil {
+			useLocalSDK = dagql.Boolean(*client.LocalLibrary)
```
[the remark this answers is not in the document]
TomChv: I will call it dev and apply the dev of arg name for localSDK in this PR dev, since it seems that it makes things confusing

core/modules/config.go
TomChv: We agreed with Solomon to rename it `dev`, so I updated the PR with that changes :)

TomChv: Actually, using `dag.Version`, the SDK module itself could be able to know if that's a built version or not, that way we can get rid of this dev param and it's up to the SDK to adapt his behaviour

core/schema/sdk.go
```diff
@@ -276,8 +276,8 @@ func (sdk *moduleSDK) GenerateClient(
 				Value: dagql.String(outputDir),
 			},
 			{
-				Name:  "useLocalSdk",
-				Value: dagql.NewBoolean(useLocalSDK),
+				Name:  "dev",
+				Value: dagql.NewBoolean(dev),
```
jedevc, line 280: :eyes: Bit confused here, is this not going to break with SDKs that don't have a `dev` argument? Why does not work at all, I don't see any changes to any other runtimes for this, like Python.

Is the `dev` argument not getting sent if it's `false`? What happens if I initialize it with `--dev=true` for `--sdk=python`? We should only attach this if the SDK supports it - we can do introspection, similar to how @rajatjindal is doing in https://github.com/dagger/dagger/pull/9323/files#diff-0f56671aa816ba46aa7ca99e164f99cb39d656cea7c4293b9b475b30030ea8eeR247.

If it's not defined - we should either: 1. error out nicely, 2. ignore it, and discard the dev option. Definitely not throw an internal error out to the user.

jedevc (approved):

PR #9763, opened 2025-03-04
feat(sdk/java): public/private exported/non exported fields

eunomie:
Update the way fields are serialized and exposed.

Fields can be serialized whatever their visibility. It doesn't mean they are exposed, a function is created with the field name to get the value.

| | serialized | exposed |
|-|----------|----------|
| `transient` | ❌ | ❌ |
| `static` | ❌ | ❌ |
| `public` | ✔️  | ✔️ |
| (package) | ✔️ | ❌ |
| `@Function` (package) | ✔️  | ✔️ |
| `protected` | ✔️ | ❌ |
| `@Function protected` | ✔️  | ✔️ |
| `private` | ✔️ | ❌ |
| `@Function private` | ✔️  | ✔️ |

Integration tests have been updated to cover private fields exposed / non exposed via a function of the name of the field.


helderco:
It feels a little strange to me that private properties get exposed by default and you need the `@Internal`. Makes more sense if the property was public. If private fields in Go were serializable you wouldn't need `// +private`. That was added because those fields are forced to be public if you need serialization.

So I'd go the other way around, by exposing automatically `public` properties (even if not usual for people to mark properties as public in Java), and using `@Function` on a private property to expose it, like in [TypeScript](https://docs.dagger.io/api/state): 
```typescript
@object()
class MyModule {
  @func()
  greeting: string
}
```

However, I'm ok deferring this to the Java community on the DX, so not blocking. Wdyt @jcsirot?

helderco (approved):
Implementation lgtm, but I want to defer to @jcsirot and maybe others on the DX. If there's consensus, then 👍

helderco:
Should rebase btw.

eunomie:
I added one last commit with the docs (that part was missing)

As it's not the same code as when you approved, let me know if this version sounds good to you @helderco

Issue #9194, opened 2024-12-13
✨ Publish java SDK to mvn central


benkeil:
### What are you trying to do?

I want to use the Java SDK to build pipelines with Kotlin script.

### Why is this important to you?

Our whole Stack is Kotlin and we benefit fromm Extensions, so it's also our first choice for dagger.io.

### How are you currently working around this?

we don't

eunomie:
@benkeil Just to let you know we are working on publishing the SDK to maven central.

Regarding Kotlin, this is working well (not the modules for now, just integration in Kotlin code).
Once libraries will be published we will also add some docs to use them with kotlin.

For instance this very basic code:

```kotlin
package com.mycompany.app

import io.dagger.client.Container
import io.dagger.client.Dagger.dag
import io.dagger.client.Directory
import io.dagger.client.ReturnType
import java.util.concurrent.atomic.AtomicBoolean

object Test {
    @JvmStatic
    fun main(args: Array<String>) {
        try {
            // Get reference to the local project
            val src: Directory = dag().host().directory(".")

            val hasFailure = AtomicBoolean(false)

            listOf(17, 21, 23).forEach { javaVersion ->
                val gradle: Container = dag()
                    .container()
                    .from("gradle:jdk$javaVersion")
                    .withDirectory("/src", src)
                    .withExec(
                        listOf("gradle", "test"),
                        Container.WithExecArguments().withExpect(ReturnType.ANY) // Do not fail on errors
                    )

                val output = gradle.stdout()
                println("Output for JDK $javaVersion:\n$output")

                if (!gradle.exitCode().equals(0)) {
                    hasFailure.set(true)
                }
            }

            if (hasFailure.get()) {
                throw RuntimeException("Some tests failed")
            }
        } finally {
            dag().close()
        }
    }
}
```

rocketraman:
Any updates on this?

PR #9767, opened 2025-03-04
chore(sdk/java): improve versioning

eunomie:
Improve versioning to prepare the publication of the different components to maven central.
Related to #9194 

> [!WARNING]
> **Breaking change**
> This requires some changes in your module source code.
>
> In order to allow the build of the module to be successful, you need to define a property in the module's `pom.xml` to set the different versions:
> - set the property `dagger.module.deps` (the value can be anything at first, a `dagger develop` will update it)
>
>     ```xml
>     <dagger.module.deps>0.16.3-template-module</dagger.module.deps>
>     ```
>
> - set the version of `io.dagger:dagger-java-sdk` and `io.dagger:dagger-java-annotation-processor` to this property. Be careful, the `dagger-java-annotation-processor` is set two times, one time as a dependency, one time as the annotation processor in the `maven-compiler-plugin`
>
>     ```xml
>     <dependency>
>       <groupId>io.dagger</groupId>
>       <artifactId>dagger-java-sdk</artifactId>
>       <version>${dagger.module.deps}</version>
>     </dependency>
>     ```
>
> The full change might look like this:
>
> ```diff
> diff --git a/modules/construct/pom.xml b/modules/construct/pom.xml
> index fccd4c1da..df2fa6250 100644
> --- a/modules/construct/pom.xml
> +++ b/modules//construct/pom.xml
> @@ -12,18 +12,19 @@
>      <properties>
>          <maven.compiler.release>17</maven.compiler.release>
>          <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
> +        <dagger.module.deps>0.16.3-construct-module</dagger.module.deps>
>      </properties>
>
>      <dependencies>
>          <dependency>
>              <groupId>io.dagger</groupId>
>              <artifactId>dagger-java-sdk</artifactId>
> -            <version>1.0.0-SNAPSHOT</version>
> +            <version>${dagger.module.deps}</version>
>          </dependency>
>          <dependency>
>              <groupId>io.dagger</groupId>
>              <artifactId>dagger-java-annotation-processor</artifactId>
> -            <version>1.0.0-SNAPSHOT</version>
> +            <version>${dagger.module.deps}</version>
>              <scope>provided</scope>
>          </dependency>
>          <dependency>
> @@ -71,7 +76,7 @@
>                              <path>
>                                  <groupId>io.dagger</groupId>
>                                  <artifactId>dagger-java-annotation-processor</artifactId>
> -                                <version>1.0.0-SNAPSHOT</version>
> +                                <version>${dagger.module.deps}</version>
>                              </path>
>                          </annotationProcessorPaths>
>                          <annotationProcessors>
> ```

The PR is composed of two main aspects (split in two commits, the last one is a small fix regarding the "internal" deps):

1. update the different `pom.xml` to set the version based on the Dagger engine version.

    That way when `bump` is called, all the components of the Java SDK will get the right version. This helps to work locally with the different libraries and avoid conflicts (especially on IDE side) between SDK built locally (without modules) and the deps from the modules.

2. automatically set the different versions in user module:

    This will set the version for the dependencies built specifically for the module. The user `pom.xml` will be updated when we `dagger develop` to reflect the version used. As some of the dependencies are specific to a module (because it can have different dependencies so different generated code) the version includes the module name and an extra `-template` suffix to make it clear.

### DIFF
--- a/.dagger/sdk_java.go
+++ b/.dagger/sdk_java.go
@@ -128,6 +128,14 @@ func (t JavaSDK) Bump(ctx context.Context, version string) (*dagger.Directory, e
 			"-DgenerateBackupPoms=false",
 			"-Dproperty=daggerengine.version",
 			"-DnewVersion=" + version,
+		}).
+		WithWorkdir("runtime/template").
+		WithExec([]string{
+			"mvn",
+			"versions:set-property",
+			"-DgenerateBackupPoms=false",
+			"-Dproperty=dagger.module.deps",
+			"-DnewVersion=" + version + "-template-module",
 		})
 
 	poms := []string{
@@ -136,6 +144,7 @@ func (t JavaSDK) Bump(ctx context.Context, version string) (*dagger.Directory, e
 		"/sdk/java/dagger-java-sdk/pom.xml",
 		"/sdk/java/dagger-java-samples/pom.xml",
 		"/sdk/java/pom.xml",
+		"/sdk/java/runtime/template/pom.xml",
 	}
 
 	bumped := dag.Directory()
--- a/.dagger/sdk_java.go
+++ b/.dagger/sdk_java.go
@@ -106,20 +106,43 @@ func (t JavaSDK) Publish(
 	return err
 }
 
-var javaVersionRe = regexp.MustCompile(`<daggerengine\.version>([0-9\.\-a-zA-Z]+)<\/daggerengine\.version>`)
+var stableVersionRe = regexp.MustCompile(`^[0-9]+\.[0-9]+\.[0-9]+$`)
 
 // Bump the Java SDK's Engine dependency
 func (t JavaSDK) Bump(ctx context.Context, version string) (*dagger.Directory, error) {
-	contents, err := t.Dagger.Source().File(javaSDKVersionPomPath).Contents(ctx)
-	if err != nil {
-		return nil, err
+	version = strings.TrimPrefix(version, "v")
+	v := version
+	if !stableVersionRe.MatchString(v) {
+		v = fmt.Sprintf("%s-SNAPSHOT", v)
+	}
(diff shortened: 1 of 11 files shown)


helderco:
@eunomie, does this mean that a Java module fetches the library from maven rather than vendoring the SDK in the user's directory?

eunomie:
> @eunomie, does this mean that a Java module fetches the library from maven rather than vendoring the SDK in the user's directory?

No, a java module will not fetch the libs from maven.
The runtime module installs locally the libs by building them.

Right now, all modules are using the same version for their lib (the `1.0.0-SNAPSHOT`).
But if there's multiple java projects, each with different dependent modules, and as the runtime is using a cache for maven, we have a conflict. There's no immediate impact as we are rebuilding them, but it means the version correspond to multiple modules and that's bad. So one aspect of the PR is to define a specific version using the module name so that it will reduce/remove the conflict and improve caching and performances.

One of the next step will be to fetch some of the libs from maven. The Java SDK is 3 different components, but one of them could be extracted so we don't have to build it again and again (this is the only not tight to the introspection file) That's not for this PR, but that's a step in this direction.

The second aspect of the PR is just to remove that `1.0.0-SNAPSHOT` so we can build it locally using a mvn install and get versioned libraries. It helps for local development, it helps to avoid conflicts when editing a module as currently everything is mixed under a single version.

eunomie:
> @eunomie, does this mean that a Java module fetches the library from maven rather than vendoring the SDK in the user's directory?

Just to specify a point: there's no vendoring.
In order to help IDEs/users we are copying some of the generated files under the `target/generated-sources` folder, but those files should not be used to build the module.

helderco (approved):
LGTM 👍

PR #9769, opened 2025-03-04
shell bubbletea followups

vito:
* [x] Better detection for incomplete input
* [x] Print logs _after_ printing child rows so it appears at the bottom of the screen
* [x] Figure out how to deal with scrollback, current attempt leaves long output chopped off :(
  * brought back flush-on-complete for now, which seems less bad
* [x] Fix revealed status not being propagated in case where revealed span arrives in same batch as its parents but is processed before them


vito:
This was pulled in as part of #9628

PR #9773, opened 2025-03-05
add `dagger llm`

vito:
* like dagger shell, but accepts LLM prompts instead
* builds up a Llm.withPrompt chain
* use /with command to set the current Llm state + tools

[asciicast](https://asciinema.org/a/CPEvr96l5dYoAyKodHdUiaKYZ)


vito:
(merged this in to `llm` manually)

PR #9774, opened 2025-03-05
cache result wrapping and context cancel management in dagql cache

sipsma:
Splitting out another seperable chunk from https://github.com/dagger/dagger/pull/9682

This updates dagql's cache to use the same "singleflight" implementation added for the local sync rework.

The idea of deduping equivalent calls is the same as the previous cachemap implementation, but with two important additional features:
* Context cancellation is handled better.
  * Previously when calls were deduped, the "first" context was always used, which meant that if it was canceled (i.e. first client disconnected for any reason), all other clients waiting on the same call would get canceled too.
  * Now, the call is only canceled if *all* caller contexts are canceled.
* Results returned from the cache are wrapped in a type that are ref-counted and releasable. When all refs for the result are released, they are removed from the cache.
  * This is not used in dagql quite yet, but will be in follow-up PRs. It still makes sense to include in this PR since it comes along for the ride with re-using the singleflight implementation from the local sync code.

The main goals here are to:
* Test the context cancelation improvements in isolation; they are a subtle change so worth verifying by themselves
* Setup the follow-ups that actually use the ref counted results to support engine-wide dagql cache


engine/cache/cache.go
```diff
@@ -0,0 +1,192 @@
+package cache
+
+import (
+	"context"
+	"fmt"
+	"sync"
+)
+
+type Cache[K comparable, V any] interface {
+	GetOrInitializeValue(context.Context, K, V) (Result[K, V], error)
```
jedevc, line 10: nit: doc comments would be very helpful :pray:

jedevc (approved):

jedevc:
> all other clients waiting on the same call would get canceled too

Huh. I wonder if this will potentially relieve some of the weird everything-gets-context-cancelled at the same time issues we've seen.

sipsma:
> Huh. I wonder if this will potentially relieve some of the weird everything-gets-context-cancelled at the same time issues we've seen.

Yeah wondering that too. Or even if errors don't go away, I have been wondering if something about the behavior before this PR is resulting in the "real" error getting swallowed and just context canceled from another call showing up. We'll see 🤞

> I wonder, would it be worth adding a test for this behavior now we've fixed it? Make two long-running calls, cancel one of them, and check the other one runs successfully.

Yeah totally, I did add unit tests already but integ test should be trivial too and is valuable I agree

PR #9776, opened 2025-03-05
chore: remove cargo-culted json labels on core types

jedevc:
These don't *seem* to be used, and also... conceptually, they shouldn't be. For the most part, these structs shouldn't be getting directly serialized, but the labels imply that they are.

We're really inconsistent with this anyways (as noted multiple times previously), so let's just remove them.

### DIFF
--- a/core/cache.go
+++ b/core/cache.go
@@ -14,7 +14,7 @@ import (
 
 // CacheVolume is a persistent volume with a globally scoped identifier.
 type CacheVolume struct {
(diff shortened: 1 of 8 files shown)


vito (approved):
These are probably from pre-DagQL when IDs were literally Base64-JSON encoded objects. But even then the keys didn't really matter. I suppose they were just a nice reminder that we have to deal with JSON marshalability.

PR #9777, opened 2025-03-05
fix: don't set `SSH_ASKPASS` to `echo`

jedevc:
I don't really have context for why this was added (somewhere in https://github.com/dagger/dagger/pull/8805), but it's definitely not right, resulting in *always* setting a completely incorrect auth token.

See the following:

    $ GIT_TERMINAL_PROMPT=0 SSH_ASKPASS=echo git credential fill <<EOF
    protocol=https
    host=github.com
    EOF

    protocol=https
    host=github.com
    username=Username for 'https://github.com':
    password=Password for 'https://Username%20for%20%27https%3A%2F%2Fgithub.com%27%3A%20@github.com':

This outputs the prompts into the username/password fields, instead of *correctly* erroring out, and not attaching this as an authentication token.

I'm not actually sure if this was causing user-facing bugs, but I hit this while working on a total refactor of the git implementation.

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -5480,7 +5480,7 @@ func privateRepoSetup(c *dagger.Client, t *testctx.T, tc vcsTestCase) (dagger.Wi
 				WithExec([]string{
 					"git", "config", "--global",
 					"credential.https://" + tc.expectedHost + ".helper",
-					`!f() { test "$1" = get && echo "password=` + token + `"; }; f`,
+					`!f() { test "$1" = get && echo -e "password=` + token + `\nusername=git"; }; f`,
 				})
 		}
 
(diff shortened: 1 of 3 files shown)


sipsma:
@jedevc Funnily enough I just hit this too while being very confused why some auth tests I was adding for my other PR were successfully pulling git repos with no auth present. Other buildkit edge merging shenanigans were involved too, but part of the problems were:
* This SSH_ASKPASS issue
* Not checking for an empty string in the Password returned by the git credential attachable 
* The `privateRepoSetup` util in the integ tests only setting a password and not username

I had to make those fixes anyways so I'll push them here. `TestCLI/TestDaggerInstall` is passing for me locally now.

The fix to SSH_ASKPASS that seems to work is to only set that env var when the protocol isn't http/https. cc @grouville please double-check that makes sense to you too.

sipsma (approved):

jedevc:
Ah! Thanks!

How convenient you found the same issue :laughing:

PR #9778, opened 2025-03-05
feat: expose introspectionJSON in module source

TomChv:
Add `introspectionJSONFile` in `ModuleSource` so we can later remove the `introspectionJSONFIle` argument in the SDKs.

This changes happens because @shykes mention that it's possible to get the module TypeDef directly through the TypeDef API instead of using this introspectionJSON, so it's only an implementation details for our SDKs but not fore future standalone client generator.

**Note**: Since `SDKs` depends on releases dagger version, we will need to wait for a release in order to change the interface.
So I suggest that: 
- Next release we deprecate the `introspectionJSONFile` argument in SDK `Codegen` & `GenerateClient`.
- In May or June, we remove the use of that argument in all our SDKs and ask maintainers to update their SDKs too.


core/schema/modulesource.go
```diff
@@ -1305,6 +1309,40 @@ func (s *moduleSourceSchema) moduleSourceRepoRootPath(
 	return src.Git.RepoRootPath, nil
 }
 
+func (s *moduleSourceSchema) introspectionJSONFile(
+	ctx context.Context,
+	src dagql.Instance[*core.ModuleSource],
+	args struct {
+		IncludeSelf dagql.Optional[dagql.Boolean]
```
wingyplus, line 1316: The SDK (Elixir,Java,Php) should not set this pr set it to `false`?

TomChv: It depends, for client generation you will want to set it to true to generate self bindings.
For `Codegen` you should not for now since we do not support self call yet

core/schema/modulesource.go
```diff
@@ -149,6 +149,10 @@ func (s *moduleSourceSchema) Install() {
 			ArgDoc("generator", `The generator to use`).
 			ArgDoc("outputDir", `The output directory for the generated client.`).
 			ArgDoc("dev", `Generate in developer mode`),
+
+		dagql.NodeFunc("introspectionJSONFile", s.introspectionJSONFile).
```
helderco, line 153: It makes some sense to have this in `ModuleSource` but it would also be helpful to be able to get the core API bindings. Right now some SDKs are getting this from the `codegen` binary when updating the core bindings in the client library. This would make that much more accessible.

So what about a top-level field (or maybe an `introspection` field returning an `Introspection` type, to scope introspection related APIs in the future, the same way we have `engine`) that receives an optional ModuleSource instead? Then if not provided, you still get the core schema. If provided, you additionally get a module's.

core/schema/modulesource.go
```diff
@@ -1308,6 +1312,40 @@ func (s *moduleSourceSchema) moduleSourceRepoRootPath(
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load dependency modules: %w", err)
+	}
+
+	// If the current module source has sources, we can transform it into a module
+	// to generate self bindings.
+	if args.IncludeSelf.Valid && args.IncludeSelf.Value.Bool() && src.Self.SDK != nil {
+		var mod dagql.Instance[*core.Module]
+		err = s.dag.Select(ctx, src, &mod, dagql.Selector{
+			Field: "asModule",
```
helderco, line 1332: Hmm... what we do here may be key to solve self-bindings. This solution is simpler for now but creates a chicken or egg situation for an SDK module to get self-bindings in the `Codegen` function. So at that point I suppose we'll need to use something internal rather than through the API.

Just thinking ahead, it's not blocking for now.

helderco (commented):
I'm all for having an API call for this 🎉 

There's just a bit of bikeshedding on what we expose on the API. We want to be careful over what we expose to avoid changing/deprecating things later.

Can you add test coverage?

core/schema/modulesource.go
helderco: We keep pushing self-bindings but they're actually pretty important. They just need some investment to figure out the best solution. Some of it depends on the language, but the hope is that there's something that the engine can standardize on to make it easier for SDKs.

core/schema/modulesource.go
helderco: For example, the Python SDK generates core bindings using this function:

https://github.com/dagger/dagger/blob/c38d8440dc933483ca51ea7b8a54463d191844df/sdk/python/dev/src/python_sdk_dev/main.py#L132-L138

And it gets the introspection file through the root module:

https://github.com/dagger/dagger/blob/c38d8440dc933483ca51ea7b8a54463d191844df/.dagger/sdk_python.go#L125

Which in turn gets it from the `codegen` CLI:

https://github.com/dagger/dagger/blob/c38d8440dc933483ca51ea7b8a54463d191844df/.dagger/sdk.go#L98

But, this means that I can't use the dev module directly to generate those bindings. Getting the introspection from the API would free that restriction.

TomChv: Why don't we just send the module source to the codegen? Because we would need it as argument to `introspection` in the coreAPI binding so that's the same problem no?

helderco: No, because the core API doesn't have a ModuleSource, right? Unless you instantiate one with an invalid value on purpose and tell it not to fail, then you'd get the core functions by default?

TomChv: Ohhh I see! okay yeah

sdk/java/dagger-codegen-maven-plugin/src/main/resources/schemas/schema-0.16.4-010101000000-dev-576994640977.json
eunomie, line 1: I think this file should not exist. We removed all of them. What's exactly the purpose here?

TomChv: I think it's my fault, I forget to remove it when I pushed! I'll update the PR

core/schema/modulesource.go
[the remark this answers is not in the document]
jedevc: Agreed on `schemaIntrospection` - not sure about `File`, it feels a bit [redundant](https://en.wikipedia.org/wiki/Hungarian_notation#Disadvantages).

core/schema/modulesource.go
jedevc: To do this, I *think* we probably want to split `Codegen` into `Codegen`+`Typedefs` if we can.

So during `Codegen` you can call `schemaIntrospectionFile`, and *that* can call `Typedefs`, so you can get the introspection file.

docs/docs-graphql/schema.graphqls
```diff
@@ -2363,6 +2363,14 @@ type ModuleSource {
   """
   repoRootPath: String!
 
+  """
+  A JSON file with the GraphQL schema introspection, including every dependency installed in this module
+  """
+  schemaIntrospectionFile(
+    """Exclude the given module from the result"""
+    exclude: [String!] = []
+  ): File!
```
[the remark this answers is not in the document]
TomChv: Yeah, basically we would always exclude the module itself for now (see: https://github.com/dagger/dagger/pull/9778#discussion_r1995490726)

PR #9797, opened 2025-03-06
helm: Enable setting environment variables for engine

vmaffet:
This PR enables setting additional environment variables for the engine container.

### DIFF
--- a/helm/dagger/values.yaml
+++ b/helm/dagger/values.yaml
@@ -25,8 +25,10 @@ engine:
   kind: DaemonSet
 
   ### Customize Dagger Engine start args
-  #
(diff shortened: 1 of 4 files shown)


helm/dagger/values.yaml
```diff
@@ -25,9 +25,11 @@ engine:
   kind: DaemonSet
 
   ### Customize Dagger Engine start args
-  #
   args: []
 
+  ### Customize Dagger Engine environment variables
+  env: []
```
jedevc, line 31: Should this be an object instead of an array?

jedevc (approved):

PR #9798, opened 2025-03-06
docs(sdk/java) docs custom applications

eunomie:
Add documentation on how to integrate the Dagger Java SDK with custom applications.

It covers three cases:

- Java / maven
- Java / gradle
- Kotlin / gradle

> [!WIP]
> Output is not currently added, it failed because libraries are not yet published. I might revisit the code but I wanted to have it as close as possible to the other languages


jasonmccallister:
@eunomie any reason to have Kotlin on its own tab vs a call out under the Java tab?

eunomie:
> @eunomie any reason to have Kotlin on its own tab vs a call out under the Java tab?

No strong opinion here. I split it because we can see Kotlin as a different language than Java. That's the main reason.
If we put it under Java, we probably want to change the "Java" title of the tab, because it's not just Java anymore, and we will have to find a good way to present the different options, for now Java/Maven, Java/Gradle, Kotlin/Gradle. We might have the same question if for instance we add Clojure/Leiningen for instance.

As a user, I think to see `Go | Python | TypeScript | PHP | Java | Kotlin | Clojure` would be ok.

Maybe it's a question of what is represented by the tab, is it the language or the SDK. In this case we have a single SDK and multiple languages.

But I can change it to group everything under Java, at least for now as it might be the only place we might have it.

eunomie:
I pushed a new version, including Clojure example.

(still missing the outputs as the libs are not published yet)

I also tried to stick to a "Java" tab then to have sub tabs Maven/Gradle and by combination of language/build tool:
image

Let me know what you think, if it's better than the previous version.

vikram-dagger (changes requested):
This seems fine, but the output of `dagger run` is missing for each example.

docs/current_docs/api/clients-sdk.mdx
```diff
@@ -544,6 +544,203 @@ Completed tests for PHP 8.3
...
+mvn archetype:generate -DgroupId=com.mycompany.app -DartifactId=my-app -DarchetypeArtifactId=maven-archetype-quickstart -DarchetypeVersion=1.5 -DinteractiveMode=false
+```
+
+Add the `dagger-java-sdk` dependency to your project's `pom.xml` file:
+
+```xml
+<dependency>
+    <groupId>io.dagger</groupId>
+    <artifactId>dagger-java-sdk</artifactId>
+    <version>0.16.3</version>
```
vikram-dagger, line 590: This should be updated to the latest version

docs/current_docs/api/clients-sdk.mdx
```diff
@@ -544,6 +544,203 @@ Completed tests for PHP 8.3
...
+</dependency>
+```
+</TabItem>
+<TabItem value="Gradle">
+```groovy
+implementation 'io.dagger:dagger-java-sdk:0.16.3'
```
vikram-dagger, line 569: This should be updated to the latest version

docs/current_docs/api/clients-sdk.mdx
```diff
@@ -544,6 +544,203 @@ Completed tests for PHP 8.3
...
+implementation("io.dagger:dagger-java-sdk:0.16.3")
+```
+
+Create a new file named `Test.kt` in the `app/src/main/kotlin/com/mycompany/app` directory and add the following code to it:
+
+```java file=./snippets/clients-sdk/kotlin/Test.kt
+```
+
+This Kotlin program imports the Dagger SDK and defines an object named `Test` with a `main` function.
+This main function creates a Dagger client, which provides an interface to the Dagger API.
```
vikram-dagger, line 681:
```suggestion
This `main` function creates a Dagger client, which provides an interface to the Dagger API.
```

vikram-dagger:
> Maybe it's a question of what is represented by the tab, is it the language or the SDK. In this case we have a single SDK and multiple languages.

Late answer but it is the SDK.

> Let me know what you think, if it's better than the previous version.

Yes, this lgtm!

eunomie:
I'm putting this one on the side for now. I'll rework/update it once the library will be published (I hope soon)

PR #9800, opened 2025-03-06
quick fix to wolfi build errors to unblock failing engine builds

sipsma:
Wolfi made a change that causes glibc triggers to require busybox triggers having already run:
https://github.com/wolfi-dev/os/commit/8229c0379cada98fb9504dd19a068dbbe2bd0d98

This isn't obviously wrong, but our current alpine module likely runs triggers in a slightly different order and we thus end up with a glibc trigger running, trying to execute /usr/bin/sh and then failing because the busybox symlink (created in its trigger) hasn't been installed yet.

The fix here is a quick hack to unblock. Will look at something more robust to follow up with.

### DIFF
--- a/modules/alpine/main.go
+++ b/modules/alpine/main.go
@@ -259,8 +259,15 @@ func (m *Alpine) withPkgs(
 			Exclude: pkg.rmFileNames,
 		})
 		ctr = ctr.With(pkgscript("post-install", pkg.name, pkg.postInstall))
+		// HACK: quick fix for busybox trigger needing to be run before glibc trigger (which needs /usr/bin/sh symlink to busybox created)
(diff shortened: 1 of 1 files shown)


grouville (approved):

sipsma:
After looking around some more, it seems like these busybox symlinks might be special-cased in apko too: https://github.com/chainguard-dev/apko/blob/3e63d01bb69480954b6040f5470197aead3f81da/pkg/build/build_implementation.go#L187

So I think this hack may be necessary in one form or another

jpadams:
thread https://discord.com/channels/707636530424053791/1003718839739105300/1347276525057085460

PR #9804, opened 2025-03-07
feat: embed remote dependencies loading on connection for client gen

TomChv:
Embed remote dependencies serving in the `Connect` method.
This enables clients to not depends on local `dagger.json` to load these dependencies, it's only requires for local one.

### List of changes

- Remote dependencies  references are passed as argument to the codegen so they can be generated (we really need to split these 2 codegen commands..., will do when things around client gen are more stable)
- Update `Serve` in engine to not fail if a same dependency is served multiple times in the session, instead it will silently ignore it. This is require because an user may connect its client multiple time in a same execution, which would create errors if it's not ignored.
- Update client generator to load dependencies
- Update integration tests
- Update TS SDK runtime because it's required to access some methods (**WILL THIS IMPACT END USER?**)
- Add local dependencies loading inside the codegen
- Remove `serveModule` logic in session in profit to client loading


cmd/codegen/generator/generator.go
```diff
@@ -57,6 +57,11 @@ type Config struct {
 	// Dev indicates that the codegen should use the local SDK instead of the published one.
 	// This is only relevant when ClientOnly is true.
 	Dev bool
+
+	// DependenciesRef is the list of all dependencies used by the module.
+	// This is used for client generator to automatically serves the dependencies
+	// on connection.
+	DependenciesRef []string
```
jedevc, line 64:
```suggestion
	DependencyRefs []string
```

`DependenciesRef` is the wrong plural.

We could even just have `Dependencies` tbf.

jedevc, line 64: Hmmm. I also think we're missing something here.

What if the user has renamed the dependency? e.g. I've got a module `github.com/shykes/hello`, but set it's `name` field to `"foobar"`. It should be served as `"foobar"`, I think this information is getting lost.

I think we actually need the *entire* module config struct. It's not just the `name`, it's the `pin`, and any other config we might choose to add later on.

cmd/codegen/generator/go/templates/src/_dagger.gen.go/defs.go.tmpl
```diff
@@ -244,10 +244,18 @@ func Connect(ctx context.Context, opts ...dagger.ClientOpt) (*Client, error) {
...
 	}
 
-	return &Client{
+	c := &Client{
 		query:  dag.QueryBuilder(),
 		client: dag.GraphQLClient(),
-	}, nil
+	}
+
+	{{ range $i, $dep := DependenciesRef }}
```
jedevc, line 252: Bleh, it's really confusing we have these near identical `Client`s. 

We should try and unify them - in `client.go.tmpl`. The `IsDevMode` statement should be in `client.go.tmpl`, so all the client code can be in one place.

Also, maybe a comment explaining why which one is which?

cmd/codegen/generator/go/templates/src/dag/dag.gen.go.tmpl
```diff
@@ -28,11 +28,21 @@ func initClient() *dagger.Client {
 		opts = append(opts, dagger.WithServeCurrentModule(true))
 		{{ end }}
 
+    ctx := context.Background()
```
jedevc, line 31: nit: indent in this file (go uses tabs)

cmd/dagger/engine.go
```diff
@@ -106,7 +107,17 @@ func withEngine(
...
-					err := dep.AsModule().Serve(ctx)
+					depKind, err := dep.Kind(ctx)
+					if err != nil {
+						return fmt.Errorf("failed to get dependency kind: %w", err)
+					}
+
+					// Skip remote dependencies because they are already serves during the call to `Connect`
+					if depKind == dagger.ModuleSourceKindGitSource {
+						continue
+					}
```
jedevc, line 118: Hmm, this seems a little weird to me. I'm curious why `Connect` can't also load the local deps - it's still the current directory right?

Makes more sense to keep the logic in the same place right? Or is there a reason why it can't be unified?

engine/server/session.go
```diff
@@ -1252,6 +1252,13 @@ func (srv *Server) ServeModule(ctx context.Context, mod *core.Module) error {
 	client.stateMu.Lock()
 	defer client.stateMu.Unlock()
 
+	// don't add the same module twice
+	// This can happen with generated clients since all remote dependencies are added
+	// on each connection and this could happen multiple times.
+	if client.deps.LookupDep(mod.Name()) {
+		return nil
+	}
```
jedevc, line 1260: Ahh I see. That's kind of reasonable I guess :thinking: 

But, as mentioned above, this logic is probably a bit more complex, since it should handle trying to serve two different versions of the same module (this should error out). Or if there's a module with this name already, and *that* module doesn't have the same module source.

jedevc (commented):
Eyy okay, the general idea looks good :tada:

I think the main blocker is that a module dependency isn't *just* a ref - it's all the other things in the config struct as well https://github.com/dagger/dagger/blob/b4cc743f54a823765e791d23c898bd1bbb65f950/core/modules/config.go#L182-L192

All of that needs to make it into the `Module` that we're actually serving in the client, or we'll lose a lot of info.

cmd/dagger/engine.go
TomChv: It's because the local dependencies may not be available depending on where the code is run.
We would also need to deal with local path resolution while this is already handle by the engine.

So to avoid complexities, I just choose to embed remote dependency loading.

engine/server/session.go
TomChv: So what should I do? I'm not sure I understand when I should error or not.

cmd/codegen/generator/generator.go
TomChv: `asString` already uses the pin and the repo, but I need to change the logic to also include the name.
I might pass a flag as argument then instead of all references as flags

cmd/dagger/engine.go
jedevc: > because the local dependencies may not be available depending on where the code is run

Yeah, but this is the same even if we embed in `dagger run` where the logic is now right?

TomChv: Yeah but then it's going to fail if the local dependency is called, the server will throw an error to mention that the object doesn't exist, so it gracefully fail IMO

cmd/codegen/generator/go/templates/src/dag/dag.gen.go.tmpl
```diff
@@ -28,11 +28,21 @@ func initClient() *dagger.Client {
...
+		ctx := context.Background()
+
 		var err error
-		client, err = dagger.Connect(context.Background(), opts...)
+		client, err = dagger.Connect(ctx, opts...)
 		if err != nil {
 			panic(err)
 		}
+
+    {{ if IsStandaloneClient }}
```
jedevc, line 39: nit: indent :eyes:

engine/server/session.go
jedevc: No, this check won't error on the last case right? It would just ignore this, and continue without serving `module-y`.

We actually need to check that the looked-up dependency is the same as the `mod` that's being passed in.

cmd/dagger/engine.go
TomChv: Do you have a flexible solution to deal with path relative to the generated client? Because if the client is generated in `a/b/c` and the dagger.json is in `a` and a local deps is in `../d`, then if you move the client file, everything will break :/

jedevc: I think it's *really* confusing that different dependencies are loaded differently, where local ones are loaded in an entirely different part of the code base from the remote ones.

> Last wednesday we agreed to only embed remote dependency, so isn't it fine enough for now?

Yeah, we did, but we're still handling local dependencies here - it's not like we're just ignoring those, we're still trying to load them?

> Yes but we do not want to error in case of local dependency embedding because if you create a CLI or something, the user will not have access to the local dep

Right. Yes. But IMO, better an explicit error if we hit this case, rather than skipping the loading of a dependency that we've tried to install - which then fails when you try to use it, right?

FYI @sipsma @helderco

TomChv: > Right. Yes. But IMO, better an explicit error if we hit this case, rather than skipping the loading of a dependency that we've tried to install - which then fails when you try to use it, right?

Since local dependency are mostly for local development, this situation would happen if a local dep is used for a published library :/
I'll add that changes because it makes sense but I'm scared that it will be quite unstable regarding path resolutions

TomChv: Ohhhh I did a try to add local deps and that's way harder than expected because it depends where the `main`  is executed... so if your main command is in a sub directory, the relative path to that local module is different.
I'm not sure how to tackle that :/

cmd/codegen/generator/generator.go
```diff
@@ -23,6 +23,12 @@ const (
 	SDKLangTypeScript SDKLang = "typescript"
 )
 
+type DependencyConfig struct {
+	Name string
+	Pin  string
+	Ref  string
+}
```
jedevc, line 30: I don't think this needs re-declaration actually - you should just be able to use this directly: https://github.com/dagger/dagger/blob/84ee1197610edd94b8997323bc7c7e7f379e288a/core/modules/config.go#L183-L193

It's a separate package so we can easily automatically generate docs from it.

cmd/codegen/generator/go/templates/src/_dagger.gen.go/client.go.tmpl
```diff
@@ -177,3 +177,78 @@ func (c errorWrappedClient) MakeRequest(ctx context.Context, req *graphql.Reques
...
+	return c, nil
+}
+{{ end }}
+
+// serveModuleDependencies services all dependencies of the module.
+// Local dependencies are served by the dagger.json.
+// Remote dependencies are generated by the client generator.
+func serveModuleDepencencies(ctx context.Context, client *Client) error {
+	{{ range $i, $dep := Dependencies -}}
+	if err := client.ModuleSource("{{ $dep.Ref }}", ModuleSourceOpts{RefPin: "{{ $dep.Pin }}"}).WithName("{{ $dep.Name }}").AsModule().Serve(ctx); err != nil {
```
jedevc, line 207: super-nit: split to multiline chain for readability

cmd/codegen/generator/go/templates/src/_dagger.gen.go/client.go.tmpl
```diff
@@ -177,3 +177,78 @@ func (c errorWrappedClient) MakeRequest(ctx context.Context, req *graphql.Reques
...
+	if !configExist {
+		return nil
+	}
+
+	dependencies, err := modSrc.Dependencies(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, dep := range dependencies {
```
jedevc, line 227: This is a lot better than it was before :tada: But I think it's still a bit weird to have different code for loading local / git.

IMO the `Dependencies` function should include *all* dependencies - not just git ones. When it comes to loading a local one though, we can modify the path to be relative to the found `dagger.json`.

> [!NOTE]
>
> Aside: we currently can't tell if a dep is local/git based on the config. This is a bit weird actually - we might want to have a `kind` in the module dependency config that we use to load the module. Probably not for now though.

That said, it's all in one place now, which is better. As long as:
1. we add some comments explaining *why* this is separate
2. we rename `Dependencies` to be `GitDependencies` or similar, to clarify the naming.

engine/server/session.go
```diff
@@ -1265,6 +1265,19 @@ func (srv *Server) ServeModule(ctx context.Context, mod *core.Module) error {
...
 
+	// don't add the same module twice
+	// This can happen with generated clients since all remote dependencies are added
+	// on each connection and this could happen multiple times.
+	depMod, exist := client.deps.LookupDep(mod.Name())
+	if exist {
+		// Error if there's a conflict between dependencies
+		if depMod.GetSource().AsString() != "" && mod.Source.Self.AsString() != depMod.GetSource().AsString() {
+			return fmt.Errorf("module %s already exists with different source %s", mod.Name(), depMod.GetSource().AsString())
+		}
```
jedevc, line 1276: `AsString` doesn't include the `Pin`.

TomChv: > AsString doesn't include the Pin. 

I think it does, it include the commit reference so that includes the pin no?

jedevc: Where does it include the commit reference?

https://github.com/dagger/dagger/blob/84ee1197610edd94b8997323bc7c7e7f379e288a/core/modulesource.go#L216-L227

The `version` doesn't neccessarily have the commit info. It *is* included in the field called `Pin`.

https://github.com/dagger/dagger/blob/84ee1197610edd94b8997323bc7c7e7f379e288a/core/modulesource.go#L493-L498

cmd/codegen/generator/go/templates/modules.go
```diff
@@ -38,6 +39,11 @@ func (funcs goTemplateFuncs) isDevMode() bool {
 	return funcs.cfg.Dev
 }
 
+func (funcs goTemplateFuncs) gitDependencies() []modules.ModuleConfigDependency {
+	fmt.Printf("%#v\n", funcs.cfg.GitDependencies)
```
jedevc, line 43: :eyes:

cmd/codegen/generator/go/templates/src/_dagger.gen.go/client.go.tmpl
```diff
@@ -177,3 +177,84 @@ func (c errorWrappedClient) MakeRequest(ctx context.Context, req *graphql.Reques
...
+		if depKind == ModuleSourceKindGitSource {
+			continue
+		}
+
+		if err := dep.AsModule().Serve(ctx); err != nil {
+			return err
+		}
+	}
+
+		sdkSource, err := modSrc.SDK().Source(ctx)
```
jedevc, line 248: nit: indent!

jedevc (commented):
Make sure to resolve the comments :)

cmd/codegen/generator/go/templates/modules.go
TomChv: Oops, my bad

cmd/codegen/generator/go/templates/src/_dagger.gen.go/client.go.tmpl
TomChv: Good catch

TomChv: Fixed by [1e89777](https://github.com/dagger/dagger/pull/9804/commits/1e897770722e32d102fc73d2bc8c176f9d83f40e)

cmd/codegen/generator/go/templates/modules.go
TomChv: Fixed by [1e89777](https://github.com/dagger/dagger/pull/9804/commits/1e897770722e32d102fc73d2bc8c176f9d83f40e)

cmd/codegen/generator/go/templates/src/_dagger.gen.go/client.go.tmpl
```diff
@@ -177,3 +177,84 @@ func (c errorWrappedClient) MakeRequest(ctx context.Context, req *graphql.Reques
...
+	return c, nil
+}
+{{ end }}
+
+// serveModuleDependencies services all dependencies of the module.
+// Local dependencies are served by the dagger.json.
+// Remote dependencies are generated by the client generator.
+func serveModuleDepencencies(ctx context.Context, client *Client) error {
+	{{ range $i, $dep := GitDependencies -}}
+	if err := client.ModuleSource(
```
jedevc, line 207: Is this bit still necessary? Isn't it included in `Dependencies` below?

TomChv: No, I'm still making the difference between the 2, local depends on dagger.json, but git doesn't, they are standalone dependencies

jedevc:
(Otherwise LGTM once merge conflicts resolved)

TomChv:
> (Otherwise LGTM once merge conflicts resolved)

I'm fixing it, php reference is painful haha, I'm solving conflict with it at least 2 time a day

cmd/codegen/generator/go/templates/src/_dagger.gen.go/client.go.tmpl
jedevc: Okay, but the git dependencies will be tried to be `Serve`d twice right if there *is* a `dagger.json`? Which is where the de-duplication logic comes in.

jedevc: Oh I see, yes, you're manually skipping over it. If you do don't skip over it... the de-duplication would apply right?

jedevc (approved):
I'm still not 100% convinced on the design of how local and git are split out, I just really hope that we don't get any bugs around that!

But let's get this in, since it's blocking a whole stream of work.

cmd/codegen/generator/go/templates/src/_dagger.gen.go/client.go.tmpl
TomChv: That's why there's the check in the session, to verify that we never load 2 times the same module yes.

TomChv:
> I'm still not 100% convinced on the design of how local and git are split out, I just really hope that we don't get any bugs around that!

I'm writing down a note so we can easily move back to that when we'll discuss the design:

Standalone client should be usable inside software like CLI, API that runs in container or even as a library.
Which means that depending on the environment, the client may not have access to the `dagger.json` that references all the dependencies.

To workaround this problem, I came up with a simple solution: *split the way Git module and local modules are loaded*
- **git module** are statically generated and loaded when the client calls `Connect`, it works no matter the environment it runs on.
- **local modules** are dynamically loaded if a `dagger.json` is found, if there's no `dagger.json`, local modules will not be callable (*for now until we find a way to correctly embed the local modules*). *NOTE* when loading the dagger.json, git dependencies are skipped since they have already been loaded statically before. This is to avoid extra logic and keep the code simple.

Issue #9794, opened 2025-03-05
Engine Pprof dumps should be easier to generate when running tests


cwlbraa:
Currently, we're all using `./hack/dev` + `./hack/with-dev go test` to pull pprof dumps. It'd be better if we could orchestrate pulling dumps at the end of `dagger call test specific` invocations, or something UX-equivalent to "run these tests and give me a pprof dump from the nested engine"

cc @sipsma

PR #9809, opened 2025-03-07
engine dev tooling: test dumps

cwlbraa:
fixes https://github.com/dagger/dagger/issues/9794

code complete, but currently broken because directory.WithNewFile() takes a string for content and I think passing these profiles through UTF-8 encoding breaks them. Gonna switch to pulling them with curl and passing actual files around.

I'm also getting 500s for cpu profiles rn, not sure why. 

taking a default 30s cpu profile
```
dagger call test dump --run="TestCache/TestVolume" --pkg=./core/integration --route=pprof/profile --delay=1s export --path=/tmp/dump0
```

or taking a heap dump every 3s:
```
dagger call test dump --run="TestCache/TestVolume" --pkg=./core/integration --delay=1s --interval=3s export --path=/tmp/dump1
```

### DIFF
--- a/.dagger/test.go
+++ b/.dagger/test.go
@@ -192,23 +192,33 @@ func (t *Test) Dump(
 	// Enable verbose output
 	// +optional
 	testVerbose bool,
-	// wait this long before starting to take dumps
+	// debug subroute to dump, like pprof/profile, pprof/heap, or requests
 	// +optional
-	// +default="30s"
+	// +default="pprof/heap"
+	route string,
+	// when true, take a final dump after the tests have completed
+	// +optional'
+	// +default=true
+	final bool,
+	// wait this long before starting to take dumps. delay does not include engine startup.
+	// +optional
+	// +default="1s"
 	delay string,
-	// wait this long between dumps
+	// wait this long between dumps. negative values will fetch exactly 1 dump excluding the one controlled by "final"
 	// +optional
-	// +default="30s"
+	// +default="-1s"
 	interval string,
 ) (*dagger.Directory, error) {
 	d, err := time.ParseDuration(delay)
 	if err != nil {
 		return nil, err
 	}
+
 	i, err := time.ParseDuration(interval)
 	if err != nil {
 		return nil, err
 	}
+
 	return t.testAndDump(
 		ctx,
 		&testOpts{
@@ -222,7 +232,10 @@ func (t *Test) Dump(
 			count:         count,
 			testVerbose:   testVerbose,
 		},
-		d, i,
+		route,
+		final,
+		d,
+		i,
 	)
 }
 
@@ -268,6 +281,8 @@ func (t *Test) test(
 func (t *Test) testAndDump(
 	ctx context.Context,
 	opts *testOpts,
+	route string,
+	final bool,
 	delay time.Duration,
 	interval time.Duration,
 ) (*dagger.Directory, error) {
@@ -292,11 +307,11 @@ func (t *Test) testAndDump(
 			bench:         false,
 		},
 	)
+	baseFileName := strings.ReplaceAll(route, "/", "-")
 
 	dumps := dag.Directory()
 	errs := []error{}
 	doneCh := make(chan struct{})
-
 	dumpCount := 0
 	go func() {
 		time.Sleep(delay)
@@ -306,13 +321,16 @@ func (t *Test) testAndDump(
 			case <-doneCh:
 				return
 			default:
-				heapData, err := fetchHeapDump(debugEndpoint)
+				heapData, err := fetchDump(debugEndpoint, route)
 				if err != nil {
 					errs = append(errs, err)
 				} else {
-					fileName := fmt.Sprintf("heap-dump-%d.prof", dumpCount)
+					fileName := fmt.Sprintf("%s-%d.pprof", baseFileName, dumpCount)
 					dumps = dumps.WithNewFile(fileName, string(heapData))
 					dumpCount++
+				}
+				if interval < 0 {
+					return
 				}
 
 				select {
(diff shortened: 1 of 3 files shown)


.dagger/test.go
```diff
@@ -159,6 +163,82 @@ func (t *Test) Specific(
 	)
 }
 
+// Run specific tests
+func (t *Test) Dump(
```
sipsma, line 167: nit: do you think it would be possible to instead just continue to centralize around the `test specific` command and make the new args here optional args to that function? Just to limit our number of variations on different functions for running tests.

I guess the trickiness comes from the fact that `test specific` currently errors when the tests fail, but that would prevent us from returning the `Directory` of pprof stuff...

Maybe something like returning a new object like `TestReport` that includes fields for the directory of pprof dumps, whether the tests failed or not, etc.

None of this is a blocker, just throwing it out there as a change we'll likely want at one point or another.

.dagger/test.go
```diff
@@ -198,6 +278,126 @@ func (t *Test) test(
...
+			bench:         false,
+		},
+	)
+	baseFileName := strings.ReplaceAll(route, "/", "-")
+
+	dumps := dag.Directory()
+	errs := []error{}
+	doneCh := make(chan struct{})
+	dumpCount := 0
+	go func() {
```
sipsma, line 316: nit: I think there's still cases this could be running still even w/ the `doneCh` synchronization, which would be extra bad if we return and it's still running. I'd probably just use an `errgroup` and be sure everything in here is using `ctx` to extent possible

.dagger/test.go
```diff
@@ -159,6 +163,82 @@ func (t *Test) Specific(
...
+	// +default=1
+	// +optional
+	count int,
+	// Enable verbose output
+	// +optional
+	testVerbose bool,
+	// debug subroute to dump, like pprof/profile, pprof/heap, or requests
+	// +optional
+	// +default="pprof/heap"
+	route string,
```
sipsma, line 198: if possible we'll at some point want to be able to get e.g. both cpu profiles and heap profiles. Not sure how complicated that'd be right now

hack/build
```diff
@@ -25,6 +25,7 @@ engine |\
         ${_EXPERIMENTAL_DAGGER_GPU_SUPPORT:+--gpu-support} |\
     start \
         --name=$CONTAINER \
+        --debug \
```
sipsma, line 28: I think adding this by default is causing the sdk dev jobs to fail, e.g. https://github.com/dagger/dagger/actions/runs/13730718938/job/38407111073?pr=9809

pretty sure 6060 is just conflicting with something... I'm not 100% sure what it's conflict with off the top of my head, but maybe it's as easy as just changing the port to 6061.

Worst case we don't have to leave in this `--debug` but it would be nice since I personally also manually add this a lot

.dagger/test.go
```diff
@@ -198,6 +278,126 @@ func (t *Test) test(
...
+	}
+
+	if dumpCount == 0 {
+		return nil, fmt.Errorf("test passed, but no dumps collected: %w", errors.Join(errs...))
+	}
+	return dumps, nil
+}
+
+// fetchDump fetches from a debug HTTP endpoint and returns it as a byte array
+func fetchDump(debugEndpoint string, route string) ([]byte, error) {
```
sipsma, line 376: Another way to avoid the byte<->string conversion issue would be to change this from directly using go http client to running a lightweight container that has `curl` and a service dep on the engine. Then it can output right to its filesystem and we can grab it as a `File` or `Directory` without having to use `WithNewFile`

sipsma (commented):
Lookin good so far, would be nice to update our CI here too to call this new functionality and output the pprof profiles as GHA artifacts. I *think* that's relatively straightforward, but if not isn't a blocker since this is already valuable locally

IDK how but someday would be cool to upload things like this to dagger cloud, but we don't really have the infra *yet*. We might get it with @gerhard's project though, so just something to keep in mind

.dagger/test.go
cwlbraa: yep, i feel similarly, gonna errgroup it. the initial impl was some very overcomplicated LLM slop and I trimmed it down a lot just to get something working, but `errs` loudly implies an errgroup.

cwlbraa: (in the llms defense I didn't initially ask it to do the plural version, i pluralized myself from its shitty unary implementation)

hack/build
cwlbraa: noted, won't commit this blindly. tbh i would prefer if we could use the default port here if we do commit this.

.dagger/test.go
cwlbraa: cool, that was my plan too. i actually had to prompt the LLM initially to not do it that way, i wanted to do it in raw go and i've come to regret that lmao

cwlbraa:
> would be nice to update our CI here too to call this new functionality and output the pprof profiles as GHA artifacts. I _think_ that's relatively straightforward, but if not isn't a blocker since this is already valuable locally

@sipsma do you just want heap dumps? tbh i'm hesitant to put dumps of any kind on mainline CI or wire this into `test specific` because i'm shaky about the performance implications of performance monitoring, shaky about how predictable timing will be, and shaky about the [error-propogation element of this that you've pointed out](https://github.com/dagger/dagger/pull/9809#discussion_r1985901227)

sipsma:
> > would be nice to update our CI here too to call this new functionality and output the pprof profiles as GHA artifacts. I _think_ that's relatively straightforward, but if not isn't a blocker since this is already valuable locally
> 
> @sipsma do you just want heap dumps? tbh i'm hesitant to put dumps of any kind on mainline CI or wire this into `test specific` because i'm shaky about the performance implications of performance monitoring, shaky about how predictable timing will be, and shaky about the [error-propogation element of this that you've pointed out](https://github.com/dagger/dagger/pull/9809#discussion_r1985901227)

Yeah just heap dumps are plenty useful, and honestly they'd only be needed at the end after the tests have finished executing. I'm not concerned about the perf implications on CI then.

sipsma (approved):
LGTM, would be nice to have a heap dump written in CI just at the end after the tests are done (to help w/ checking for memory leaks), but can be a follow up if preferred

Issue #9579, opened 2025-02-13
🐞 [Dagger Shell] non-legit unsupported argument of kind map


TomChv:
### What is the issue?

Reported on Discord https://discord.com/channels/707636530424053791/1339586071557701744/1339586071557701744

> I was doing some tests of the dockerSDK on a project with dagger shell and I keep hitting the same  issue when I try to configure the ports:
> ```
> docker | build | up --ports 8000:8000 --args ./tls-exterminator,8000:github.com
> Error: unsupported argument of kind map
> ```
>
> This is something that isn't happening with the Dagger CLI
> ```
> dagger call docker build up --ports 8000:8000 --args ./tls-exterminator,8000:github.com
> ```

@helderco 

### Dagger version

0.15.4

PR #9824, opened 2025-03-10
shell: rewrite state management

helderco:
### Context

What is state? It’s essentially a struct that holds all the information needed for building an API query using the query builder. The current query builder doesn’t deserialize from a string so we pass around the raw data instead, and only build the query when we need to resolve it.

This fundamentally changes how state is passed around. When you use a pipe (`|`) between functions, each of them is executed in their own goroutine, connected via an `os.Pipe`. So one function’s stdout is connected to the next function’s stdin.

## 1. State no longer serialized

Rather than serializing the state and deserializing between functions, we now hold the instance in a map in memory and pass around the key instead.

This avoids issues where some deserialized argument values would cause errors when getting marshalled again by the query builder. One example of this is when you have a list of input object types like `[]PortForward`. For instance, in the `--ports` argument of `Container.up`:

Screenshot 2025-03-06 at 18 29 14

That argument is serialized as `[{"backend": 80, "frontend": 8000}]`

Fixes https://github.com/dagger/dagger/issues/9579

When I started working on this my approach was to either avoid a value being marshalled twice, or properly unmarshall it like in modules. However it started to get a bit involved and I had already had some discussions in the past that if the state stayed in memory it could make a few things easier.

Note that we could be building the query builder instance along the way now instead of building it in the end, but left that for another time. It doesn’t make a difference performance wise and this makes it easier to debug what’s the content of the state anyway.
## 2. State interpolation

This change also resolves multiple references to states in the same string, rather than only matching strings that contained a state and nothing else.

```
⋈ FOO=$(directory | with-new-file foo foo | file foo)
⋈ $FOO | export ./$($FOO | name).txt
/Users/helder/Projects/dagger/foo.txt
```

Previously you’d get an error:
```
Error: stat /Users/helder/Projects/dagger/DSH:eyJtb2RSZWYiOiIiLCJjbWQiOiIiLCJjYWxscyI6W3sib2JqZWN0IjoiUXVlcnkiLCJuYW1lIjoiZGlyZWN0b3J5IiwiYXJndW1lbnRzIjp7fSwicmV0dXJuT2JqZWN0IjoiRGlyZWN0b3J5In0seyJvYmplY3QiOiJEaXJlY3RvcnkiLCJuYW1lIjoid2l0aE5ld0ZpbGUiLCJhcmd1bWVudHMiOnsiY29udGVudHMiOiJmb28iLCJwYXRoIjoiZm9vIn0sInJldHVybk9iamVjdCI6IkRpcmVjdG9yeSJ9LHsib2JqZWN0IjoiRGlyZWN0b3J5IiwibmFtZSI6ImZpbGUiLCJhcmd1bWVudHMiOnsicGF0aCI6ImZvbyJ9LCJyZXR1cm5PYmplY3QiOiJGaWxlIn0seyJvYmplY3QiOiJGaWxlIiwibmFtZSI6Im5hbWUiLCJhcmd1bWVudHMiOnt9LCJyZXR1cm5PYmplY3QiOiIifV19Cg==.txt: file name too long
```

## One more thing

I added the `-n` shorthand for `--no-mod`:
```
dagger -n
```


jedevc:
Oh! I remember actually why I wanted this! https://discord.com/channels/707636530424053791/1307129843434918009/1332319693691818076

> looking into some stuff where it looks like dagger shell produces some odd graphql queries - e.g. int flags are getting queried as strings, enums are turning into strings, etc.
>
> the issue seems to be that we have all the correctly typed values right here https://github.com/dagger/dagger/blob/dd4d69d11fb39cf9247fd3a6add99231133e88b6/cmd/dagger/shell_exec.go#L236-L241
but when we build the QueryBuilder, it's been JSON serialized + deserialized, so all of our nice rich types are entirely gone, and enums are now just normal strings, and ints are json.Numbers

This should fix us producing invalid graphql queries, so we can make the graphql validation a bit stricter hopefully!

core/integration/module_shell_test.go
```diff
@@ -29,7 +29,7 @@ func daggerShell(script string) dagger.WithContainerFunc {
 
 func daggerShellNoMod(script string) dagger.WithContainerFunc {
 	return func(c *dagger.Container) *dagger.Container {
-		return c.WithExec([]string{"dagger", "--no-mod"}, dagger.ContainerWithExecOpts{
+		return c.WithExec([]string{"dagger", "-n"}, dagger.ContainerWithExecOpts{
```
jedevc, line 32: :tada: eyy this is great, I type this *so* much.

I like the ergonomics - `-n` and `-m` :smile:

cmd/dagger/rand/text.go
```diff
@@ -0,0 +1,30 @@
...
+import "crypto/rand"
+
+const base32alphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZ234567"
+
+// Text returns a cryptographically random string using the standard RFC 4648 base32 alphabet
+// for use when a secret string, token, password, or other text is needed.
+// The result contains at least 128 bits of randomness, enough to prevent brute force
+// guessing attacks and to make the likelihood of collisions vanishingly small.
+// A future version may return longer texts as needed to maintain those properties.
+func Text() string {
```
jedevc, line 22: :tada: any chance we could move this outside of `cmd/dagger`? I'd like to update a lot of our `identity.NewID` usage after this merges.

It could go in `core`? Or `engine`?

cmd/dagger/shell_commands.go
```diff
@@ -149,21 +149,10 @@ func (c *ShellCommand) Execute(ctx context.Context, h *shellCallHandler, args []
...
-			if err != nil {
-				return fmt.Errorf("cannot expand command argument at %d", i)
-			}
-			if v == nil {
-				return fmt.Errorf("unexpected nil value while expanding argument at %d", i)
-			}
-			arg = fmt.Sprintf("%v", v)
-		}
-		a = append(a, arg)
+	// sesolve state values in arguments
```
jedevc, line 152:
```suggestion
	// resolve state values in arguments
```

cmd/dagger/rand/text.go
helderco: Sure, I'll move it to `core` :)

PR #9825, opened 2025-03-10
docs: Improves metadata and tags for better social sharing and SEO

jasonmccallister:
All of the links from X look like the below image, this change uses the new dagger-factory.jpg icon for social sharing to make the links posted on X look more interesting. Hopefully this makes the link more interesting to capture the users attention.

1. Adds default descriptions and image urls for social sharing
2. Sets default metadata description - default is to use the second heading (e.g. the docs landing page description is "what is Dagger"). Individual markdown pages can still override the description using markdown front matter
3. Updates the default image URL to be complete as it's better for links (I believe its actually required)

Sharing docs links on X screenshot:
Screenshot 2025-03-10 at 10 48 45 AM

Updated links will look like the following:
Screenshot 2025-03-10 at 12 04 37 PM

### DIFF
--- a/docs/docusaurus.config.ts
+++ b/docs/docusaurus.config.ts
@@ -15,7 +15,7 @@ const config: Config = {
   favicon: "img/favicon.svg",
 
   // Set the production url of your site here
-  url: "https://docs.dagger.io",
+  url: url,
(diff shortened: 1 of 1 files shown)


jpadams (approved):
👍

PR #9829, opened 2025-03-10
ci: Run all Engine & CLI tests against a dev engine only

gerhard:
We spoke about this with Erik, Justin & Connor a few times in the last few weeks, and the consensus is that we want all Engine tests to run against **dev** Engines (the next version that ships), and there is marginal value in running them against an already released version.

The thinking goes that if all tests continue running the same (or better) in the next version of Dagger, then we are confident in the upcoming release.

While we might want to do the same for all other jobs in this workflow, we want to take a smaller step since we are planning to try out a different CI runner type.

### DIFF
--- a/.github/main.go
+++ b/.github/main.go
@@ -168,29 +168,20 @@ func (ci *CI) withTestWorkflows(runner *dagger.Gha, name string) *CI {
 		})).
 		WithJob(runner.Job("scripts", "check --targets=scripts")).
 		WithJob(runner.Job("cli-test-publish", "cli test-publish")).
-		WithJob(runner.Job("engine-test-publish", "engine publish --image=dagger-engine.dev --tag=main --dry-run")).
+		WithJob(runner.Job("cli-test-publish", "cli test-publish", dagger.GhaJobOpts{
+			Runner: []string{GoldRunner(false)},
+		})).
+		WithJob(runner.Job("engine-test-publish", "engine publish --image=dagger-engine.dev --tag=main --dry-run", dagger.GhaJobOpts{
+			Runner: []string{GoldRunner(false)},
+		})).
 		WithJob(runner.Job("scan-engine", "engine scan")).
-		With(splitTests(runner, "test-", false, []testSplit{
-			{"modules", []string{"TestModule"}, &dagger.GhaJobOpts{
-				Runner: []string{GoldRunner(false)},
-			}},
-			{"module-runtimes", []string{"TestGo", "TestPython", "TestTypescript", "TestElixir", "TestPHP", "TestJava"}, &dagger.GhaJobOpts{
-				Runner: []string{GoldRunner(false)},
-			}},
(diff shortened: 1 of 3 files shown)


cwlbraa (approved):
onboard with the intended change, as discussed, but tbh i'm a little surprised by the volume of deletion here, is all that code really only to support running against released versions? or is it  that we obviated a bunch of bash+yaml chaff by moving onto dagger shell invocations?

gerhard:
> i'm a little surprised by the volume of deletion here, is all that code really only to support running against released versions? or is it that we obviated a bunch of bash+yaml chaff by moving onto dagger shell invocations?

We are running 2 fewer jobs.

Jobs before this change:
1. test-cli-engine
2. test-everything-else
3. test-module-runtimes
4. test-modules
5. testdev-container
6. testdev-module-runtimes
7. testdev-modules


Jobs after this change:
1. testdev-cli-engine
2. testdev-container
3. testdev-everything-else
4. testdev-module-runtimes
5. testdev-modules

cwlbraa:
@gerhard AHHHH i missed the `.gen.` in the filename! makes sense!

gerhard:
@sipsma @jedevc Do we need a stable Engine to run `TestProvision`? FTR: https://v3.dagger.cloud/dagger/traces/b735d4252d1ea4afc084216617f3e5fb?span=779e4caf961aab8e

image

sipsma:
> @sipsma @jedevc Do we need a stable Engine to run `TestProvision`? FTR: https://v3.dagger.cloud/dagger/traces/b735d4252d1ea4afc084216617f3e5fb?span=779e4caf961aab8e
> 
> 
> 
> image

The errors look cgroup v2 related, did the runner host change from ubuntu?

.github/main.go
```diff
@@ -171,24 +171,12 @@ func (ci *CI) withTestWorkflows(runner *dagger.Gha, name string) *CI {
...
-			}},
-			{"cli-engine", []string{"TestCLI", "TestEngine"}, &dagger.GhaJobOpts{
-				Runner: []string{GoldRunner(false)},
-			}},
 			{"cgroupsv2", []string{"TestProvision", "TestTelemetry"}, &dagger.GhaJobOpts{
 				// HACK: our main runners don't support cgroupsv2
 				Runner: []string{"ubuntu-latest"},
 				// NOTE: needed to silence redis warning logs in tests
 				SetupCommands: []string{"sudo sysctl -w vm.overcommit_memory=1"},
 			}},
```
jedevc, line 179: We can move this into `testdev` as well.

It doesn't need a stable engine, but what's happening now is that `evenything-else` in `testdev` is picking up `TestProvision` (since it's a test suite that's remaining, see the logic in `splitTests`)

gerhard: I understand what is happening now, thank you @jedevc. Pushing the fix now.

jedevc (approved):

PR #9830, opened 2025-03-10
docs: Use property, not name and add images

jasonmccallister:
Follow up to #9825 

**TL;DR** its not _easily_ possible to override the default description for doc style pages, easy for blog pages. 

We have the following options for improving SEO on the docs:

**Change the structure of the actual content** 

The first `#` will become the title, the next block will become the description. We are using `##` a lot following the `#` so that defaults to short descriptions (e.g. "What is Dagger?")

**Set the description manually for all doc pages**

Probably the best choice in the long run as we want to properly summarize each page to improve search results

**Customize the doc component**

See this [thread](https://github.com/facebook/docusaurus/issues/2968) for more details. I'm not too familiar with this approach 

cc @vikram-dagger


jpadams (approved):
👍

PR #9838, opened 2025-03-11
Add Dagger Cloud GIF to intro for visualization

mircubed:
Adds cloud-trace.gif to the introduction to provide a visual representation of Dagger Cloud tracing.

### DIFF
--- a/docs/current_docs/index.mdx
+++ b/docs/current_docs/index.mdx
@@ -20,7 +20,7 @@ Dagger is an open-source runtime for composable workflows. It's perfect for syst
 
 - **Built-in Observability:** Full visibility into operations with tracing, logs, and metrics. Debug complex workflows and know exactly what's happening.
 
-<p align="center"><img src="img/cloud-trace.gif" width="60%"></img>
+![Dagger Trace](/img/cloud-trace.gif)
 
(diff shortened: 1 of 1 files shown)


jasonmccallister:
@mircubed made a path fix to the image, there should be a deploy preview available in a minute.

jasonmccallister (approved):
Waiting on the preview deploy, which might be broken atm.

docs/current_docs/index.mdx
```diff
@@ -20,6 +20,8 @@ Dagger is an open-source runtime for composable workflows. It's perfect for syst
 
 - **Built-in Observability:** Full visibility into operations with tracing, logs, and metrics. Debug complex workflows and know exactly what's happening.
 
+<p align="center"><img src="img/cloud-trace.gif" width="60%"></img>
```
vikram-dagger, line 23:
```suggestion
![Dagger Trace](/img/cloud-trace.gif)
```

I'm generally not in favour of mixing HTML with Markdown unless there's a good reason. In this case, I think the suggestion above should achieve the required result.

vikram-dagger: Side note: images which appear in the current docs content should be stored in static/img/current_docs (we can fix this separately, noting here for reference)

vikram-dagger: Going ahead and pushing this change to fix the broken build

PR #9845, opened 2025-03-11
extend dagger helm chart

jholm117:
I added a bunch of configuration options that my org is using and I'm hoping to get them into main so that we don't have to rely on my fork moving forward. 

I tried to make sure the values made sense for both statefulset and daemonset modes but I may have missed the mark there. Definitely open to feedback and let me know if I need to break up the PR I know there are a lot of changes.

Summary of changes:

- add support for configuring the Dagger engine via the `engine.json` file.
- Add container port and optionally bind to a host port.
- Add nodeSelector
- Add arbitrary volume mounts
- Add lifecycle preStop hook
- Add podManagementPolicy
- Add option to disable hostPath mount for run and data volumes.
- Add option to use pvc for data volume in StatefulSet

### DIFF
--- a/helm/dagger/templates/engine-daemonset.yaml
+++ b/helm/dagger/templates/engine-daemonset.yaml
@@ -64,7 +64,7 @@ spec:
             - "--addr"
             - "tcp://0.0.0.0:{{ .Values.engine.port }}"
             - "--addr"
-            - "unix:///var/run/dagger/engine.sock"
+            - "unix:///run/dagger/engine.sock"
             {{- end }}
             {{- if .Values.engine.args }}
             {{- toYaml .Values.engine.args | nindent 12 }}
@@ -128,7 +128,7 @@ spec:
             {{- end }}
             {{- if .Values.engine.hostPath.runVolume.enabled }}
             - name: varrundagger
-              mountPath: /var/run/dagger
+              mountPath: /run/dagger
             {{- end }}
             {{- if .Values.engine.config }}
             - name: dagger-engine-config
@@ -153,7 +153,7 @@ spec:
         {{- if .Values.engine.hostPath.runVolume.enabled }}
         - name: varrundagger
           hostPath:
-            path: /var/run/dagger-{{ include "dagger.fullname" . }}
+            path: /run/dagger-{{ include "dagger.fullname" . }}
         {{- end }}
         {{- if (or .Values.engine.config .Values.engine.configJson) }}
         - name: dagger-engine-config
--- a/helm/dagger/templates/engine-daemonset.yaml
+++ b/helm/dagger/templates/engine-daemonset.yaml
@@ -129,6 +129,7 @@ spec:
             {{- if .Values.engine.hostPath.runVolume.enabled }}
             - name: varrundagger
               mountPath: /var/run/dagger
+            {{- end }}
             {{- if .Values.engine.config }}
             - name: dagger-engine-config
               mountPath: /etc/dagger/engine.toml
--- a/helm/dagger/templates/engine-daemonset.yaml
+++ b/helm/dagger/templates/engine-daemonset.yaml
@@ -148,12 +148,12 @@ spec:
         {{- if .Values.engine.hostPath.dataVolume.enabled }}
         - name: varlibdagger
           hostPath:
-            path: /var/lib/dagger
+            path: /var/lib/dagger-{{ include "dagger.fullname" . }}
         {{- end }}
         {{- if .Values.engine.hostPath.runVolume.enabled }}
         - name: varrundagger
           hostPath:
-            path: /var/run/dagger
+            path: /var/run/dagger-{{ include "dagger.fullname" . }}
         {{- end }}
         {{- if (or .Values.engine.config .Values.engine.configJson) }}
         - name: dagger-engine-config
--- a/helm/dagger/templates/engine-daemonset.yaml
+++ b/helm/dagger/templates/engine-daemonset.yaml
@@ -13,9 +13,9 @@ spec:
       name: {{ include "dagger.fullname" . }}-engine
   template:
     metadata:
-      {{- if (or .Values.engine.config .Values.magicache.enabled .Values.engine.annotations) }}
+      {{- if (or .Values.engine.config .Values.engine.configJson .Values.magicache.enabled .Values.engine.annotations) }}
       annotations:
-        {{- if .Values.engine.config }}
+        {{- if (or .Values.engine.config .Values.engine.configJson) }}
         checksum/config: {{ include (print $.Template.BasePath "/engine-config.yaml") . | sha256sum }}
         {{- end }}
         {{- if .Values.magicache.enabled }}
@@ -29,6 +29,10 @@ spec:
         name: {{ include "dagger.fullname" . }}-engine
         {{- include "dagger.labels" . | nindent 8 }}
     spec:
+      {{- with .Values.engine.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- with .Values.engine.tolerations }}
       tolerations:
         {{- toYaml . | nindent 8 }}
@@ -54,9 +58,17 @@ spec:
         - name: dagger-engine
           image: {{ if .Values.engine.image.ref }}{{ .Values.engine.image.ref }}{{ else }}registry.dagger.io/engine:v{{ .Chart.Version }}{{ end }}
           imagePullPolicy: {{ .Values.engine.image.pullPolicy }}
-          {{- if .Values.engine.args }}
+          {{- if or .Values.engine.port .Values.engine.args }}
           args:
+            {{- if .Values.engine.port }}
+            - "--addr"
+            - "tcp://0.0.0.0:{{ .Values.engine.port }}"
+            - "--addr"
+            - "unix:///var/run/dagger/engine.sock"
+            {{- end }}
+            {{- if .Values.engine.args }}
             {{- toYaml .Values.engine.args | nindent 12 }}
+            {{- end }}
           {{- end }}
           {{- if (or .Values.engine.env .Values.magicache.enabled) }}
           env:
@@ -77,6 +89,15 @@ spec:
               name: {{ include "dagger.fullname" . }}-magicache-token
               {{- end }}
           {{- end }}
+          {{- if .Values.engine.port }}
+          ports:
+            - name: dagger
+              containerPort: {{ .Values.engine.port }}
+              protocol: TCP
+              {{- if .Values.engine.hostPort }}
+              hostPort: {{ .Values.engine.hostPort }}
+              {{- end }}
+          {{- end }}
           securityContext:
             privileged: true
             capabilities:
@@ -93,9 +114,19 @@ spec:
             {{- if .Values.engine.readinessProbeSettings }}
             {{- toYaml .Values.engine.readinessProbeSettings | nindent 12 }}
             {{- end }}
+          {{- if .Values.engine.lifecycle }}
+          lifecycle:
+            {{- if .Values.engine.lifecycle.preStop }}
+            preStop:
+              {{- toYaml .Values.engine.lifecycle.preStop | nindent 14 }}
+            {{- end }}
+          {{- end }}
           volumeMounts:
+            {{- if .Values.engine.hostPath.dataVolume.enabled }}
             - name: varlibdagger
               mountPath: /var/lib/dagger
+            {{- end }}
+            {{- if .Values.engine.hostPath.runVolume.enabled }}
             - name: varrundagger
               mountPath: /var/run/dagger
             {{- if .Values.engine.config }}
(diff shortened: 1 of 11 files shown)


helm/dagger/templates/engine-statefulset.yaml
```diff
@@ -11,14 +11,20 @@ spec:
   # DO NOT RUN MORE THAN 1 REPLICA.
   # Only one (1) Engine is able to hold the lock, other replicas will not be able to start.
   replicas: 1
+  podManagementPolicy: {{ .Values.engine.statefulSet.podManagementPolicy }}
```
gerhard, line 14: It's useful to know that this exists, but we do not expect to support more than `1` anytime soon.

helm/dagger/templates/engine-statefulset.yaml
```diff
@@ -11,14 +11,20 @@ spec:
...
   selector:
     matchLabels:
       name: {{ include "dagger.fullname" . }}-engine
+  {{- if and .Values.engine.storage.data.persistentVolumeClaim.enabled .Values.engine.storage.persistentVolumeClaimRetentionPolicy }}
+  persistentVolumeClaimRetentionPolicy:
+    whenDeleted: {{ .Values.engine.storage.persistentVolumeClaimRetentionPolicy.whenDeleted }}
+    whenScaled: {{ .Values.engine.storage.persistentVolumeClaimRetentionPolicy.whenScaled }}
+  {{- end }}
```
gerhard, line 22: Since these options are only used for the `statefulSet` I **think** that it would be best to namespace them, same as we did with `podManagementPolicy`. WDYT @jholm117 ?

helm/dagger/templates/engine-statefulset.yaml
```diff
@@ -80,12 +90,28 @@ spec:
...
           {{- end }}
+          {{- if .Values.engine.port }}
+          ports:
+            - name: dagger
+              containerPort: {{ .Values.engine.port }}
+              protocol: TCP
+              {{- if .Values.engine.hostPort }}
+              hostPort: {{ .Values.engine.hostPort }}
+              {{- end }}
+          {{- end }}
```
gerhard, line 101: I really like how you layered the `hostPort` inside the port declaration.

Do you think that we should add a custom `--addr` arguments when this is defined @jholm117? I'm thinking:

```yaml
          args:
            - "--addr"
            - "tcp://0.0.0.0:{{ .Values.engine.port }}"
            - "--addr"
            - "unix:///var/run/buildkit/buildkitd.sock"
```

WDYT @jedevc ?

helm/dagger/templates/engine-statefulset.yaml
```diff
@@ -80,12 +90,28 @@ spec:
...
               add:
                 - ALL
           resources: {{- toYaml .Values.engine.resources | nindent 12 }}
+          {{- if .Values.engine.lifecycle }}
+          lifecycle:
+            {{- if .Values.engine.lifecycle.preStop }}
+            preStop:
+              {{- toYaml .Values.engine.lifecycle.preStop | nindent 14 }}
+            {{- end }}
+          {{- end }}
```
gerhard, line 114: Nice addition! I had a use-case a while back where I was using `dagger core engine local-cache prune` to manually free up the disk space on shutdown. FTR: https://docs.dagger.io/configuration/cache/

helm/dagger/templates/engine-statefulset.yaml
```diff
@@ -99,27 +125,63 @@ spec:
...
+            {{- end }}
             {{- if .Values.engine.config }}
             - name: config
               mountPath: /etc/dagger/engine.toml
               subPath: engine.toml
             {{- end }}
+            {{- if .Values.engine.configJson }}
+            - name: config
+              mountPath: /etc/dagger/engine.json
+              subPath: engine.json
```
gerhard, line 140: What will happen if we have both of these configs @jedevc?

I am wondering whether we should only mount one of these. The fact that we are using the same name for both volume mounts - `config` - will result in an invalid manifest if both are defined. Not that likely to happen, but I can imagine the beginning of a face palm / table flip moment here.

helm/dagger/templates/engine-statefulset.yaml
jholm117: That's fair but FWIW I've been autoscaling the stateful set using EBS volumes instead of the hostpath mount to great success!

helm/dagger/templates/engine-statefulset.yaml
jholm117: Definitely, good call.

helm/dagger/templates/engine-statefulset.yaml
```diff
@@ -99,27 +125,63 @@ spec:
...
+            {{- end }}
+            {{- with .Values.engine.volumeMounts }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
       terminationGracePeriodSeconds: {{ .Values.engine.terminationGracePeriodSeconds }}
       volumes:
+        {{- if not .Values.engine.storage.data.persistentVolumeClaim.enabled }}
```
gerhard, line 147: Should this be a `.Values.engine.storage.run.hostPath.enabled` condition instead? We are using a `hostPath` value which will make the connection easier to make.

helm/dagger/templates/engine-statefulset.yaml
jholm117: I thought about that and decided to leave it out since I could already define it via `Values.args` and wanted to keep it simple. I think at the very least I can add a comment in the values file next to port that instructs the user to include the `--addr tcp://...` arg. Up to you!

helm/dagger/templates/engine-statefulset.yaml
```diff
@@ -99,27 +125,63 @@ spec:
...
+              {{- end }}
+              {{- if .Values.engine.configJson }}
+              - key: engine.json
+                path: engine.json
+              {{- end }}
         {{- end }}
+        {{- with .Values.engine.volumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+  {{- if .Values.engine.storage.data.persistentVolumeClaim.enabled }}
```
gerhard, line 174: https://github.com/dagger/dagger/pull/9845/files#r1995602905 applies here too.

helm/dagger/values.yaml
```diff
@@ -15,6 +15,13 @@ engine:
   #   [log]
   #     format = "json"
 
+  ### Alternative: Configure Dagger Engine using JSON format
```
gerhard, line 18: I think that one of these should have a higher priority. WDYT @jedevc ?

helm/dagger/values.yaml
```diff
@@ -88,9 +110,74 @@ engine:
...
 
+  ### Configure lifecycle hooks for the Dagger Engine
+  #
+  lifecycle:
+    # PreStop hook executed before container termination
+    preStop:
+      # exec:
+      #   command: ["/bin/sh", "-c", "sleep 10"]  # Example: graceful shutdown delay
+
+  ### Storage configuration for StatefulSet
```
gerhard, line 125: I like that you captured the `StatefulSet` applicability @jholm117 ! WDYT about nesting this entire config under `statefulSet`?

helm/dagger/templates/engine-statefulset.yaml
jholm117: While there is a uniqueness constraint on volume names I don't believe there is one on volume mounts since it's not actually a name for the mount, rather a reference to the volume.

helm/dagger/values.yaml
```diff
@@ -88,9 +110,74 @@ engine:
...
+    {}
     # name: "default"
 
+  ### Configure lifecycle hooks for the Dagger Engine
+  #
+  lifecycle:
+    # PreStop hook executed before container termination
+    preStop:
+      # exec:
+      #   command: ["/bin/sh", "-c", "sleep 10"]  # Example: graceful shutdown delay
```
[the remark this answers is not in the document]
jholm117: > The Pod's termination grace period countdown begins before the PreStop hook is executed, so regardless of the outcome of the handler, the container will eventually terminate within the Pod's termination grace period.

I think it's realistic, maybe I'd use a longer sleep in practice. In reality I used an adaptation of this script: https://github.com/seatgeek/buildkit-prestop-script

helm/dagger/templates/engine-statefulset.yaml
gerhard: That is an interesting use-case! If you were able to make it work, do you want to capture this use-case in the comment above? Something along the lines of:

```diff
- # DO NOT RUN MORE THAN 1 REPLICA
+ # DO NOT RUN MORE THAN 1 REPLICA IF USING HOSTPATH
# Only one (1) Engine is able to hold the lock, other replicas will not be able to start.
+ # If using persistent volumes, this is not an issue. Use more than 1 replica if you know that you need it.
```

helm/dagger/values.yaml
jholm117: I could be totally off-base but my understanding was they each provide unique configuration options and can coexist.

helm/dagger/templates/engine-statefulset.yaml
gerhard: I think that we should add these values explicitly when and the port value is defined, otherwise users get caught out when the unix port is not specified explicitly and the health checks start failing. At least that was my experience many months ago when I forgot about the need to add `--addr unix:///var/run/buildkit/buildkitd.sock` explicitly.

helm/dagger/templates/engine-statefulset.yaml
gerhard: Ah yes, you are right - this is the reference! 👍

gerhard (commented):
This looks almost there @jholm117 ! Keen to review it again once you go through the comments.

I'm happy to add the changelog fragment in case you don't get to it. [FTR](https://github.com/dagger/dagger/blob/c38d8440dc933483ca51ea7b8a54463d191844df/CONTRIBUTING.md#3-add-release-notes-fragment)

helm/dagger/templates/engine-statefulset.yaml
jholm117: Definitely, nice catch!

helm/dagger/values.yaml
jholm117: I moved the pvc template config under statefulSet, but I kept the hostpath config outside that scope since it applies to the daemonset as well (and renamed the keys). Let me know what you think!

helm/dagger/templates/engine-statefulset.yaml
jholm117: I had the same experience 😂 updated!

helm/dagger/templates/engine-statefulset.yaml
jholm117: Updated!

jholm117:
> This looks almost there @jholm117 ! Keen to review it again once you go through the comments.
> 
> I'm happy to add the changelog fragment in case you don't get to it. [FTR](https://github.com/dagger/dagger/blob/c38d8440dc933483ca51ea7b8a54463d191844df/CONTRIBUTING.md#3-add-release-notes-fragment)

@gerhard Thanks for the review! I think I addressed all the comments and added the changelog entry but feel free to modify it!

gerhard:
Thanks for all the updates @jholm117 💪

Today is busier than usual, and this might spill into tomorrow, but I am keen to go over this again ASAP with the intent of having it merged in time for the next release.

cc @jedevc @matipan in case you want to take a quick look at this too, before it gets merged.

helm/dagger/values.yaml
TomChv: Based on https://github.com/dagger/dagger/pull/8800, `engine.json` should be prioritise since it aims to fully replace `engine.toml`  but it doesn't support by it so depending on the case, you may need to also set some options in `engine.toml` 

Based on https://github.com/dagger/dagger/blob/ecf75160867d8355b51728a0052de8bc3a1c950d/.dagger/engine.go#L105-L109 and what I could find on https://docs.dagger.io/configuration/engine/#configuration, both can cohexist with a priority over `engine.json`!

> If an option is specified in both engine.json and in engine.toml, then the engine.json settings will take precedence. 

That said, I think it's fine to have both, just reverse the 2 description so json is shown first and as the default to go for, and explain that toml should be used in addition if a specific option isn't supported by the json format.

TomChv (commented):
Looks pretty good to me, left 1 important comment about `engine.json` / `engine.toml`.
I think it's mergeable after this comment is resolved.

I leave it to @gerhard for the final approval :D

helm/dagger/templates/engine-statefulset.yaml
TomChv: Please look at https://github.com/dagger/dagger/pull/9866, the path will be change to `unix:///run/dagger/engine.sock` (backward compatibility is handled but might be fine to update that PR if it lands after 9866 is merged.

helm/dagger/values.yaml
jholm117: Sounds good to me, updated!

jholm117:
> Thanks for all the updates @jholm117 💪
> 
> Today is busier than usual, and this might spill into tomorrow, but I am keen to go over this again ASAP with the intent of having it merged in time for the next release.
> 
> cc @jedevc @matipan in case you want to take a quick look at this too, before it gets merged.

Happy to! Do I need to update the chart version?

helm/dagger/values.yaml
TomChv: Thank looks good to me on this side 🚀

helm/dagger/templates/engine-statefulset.yaml
jholm117: Should it be `unix:///run/dagger/engine.sock` or `unix:///var/run/dagger/engine.sock`?

TomChv: Screenshot 2025-03-17 at 19 57 03

I think `/var/bin/dagger/engine.sock`

TomChv:
@jholm117 I see that the `helm` CI job is failing, this need to be fixed before merging ;)

helm/dagger/values.yaml
jedevc: Yup, both can co-exist - the `engine.json` takes priority at a per-setting level.

TomChv:
> Okay think I fixed the lint error. The check says timeout on this page, but on the Checks page it looks like it succeeded? 

Yeah looks good, I retriggered the other jobs, cgroupv2 is expected to fail until https://github.com/dagger/dagger/pull/9876 is merged, the rest should be green :D

helm/dagger/templates/engine-statefulset.yaml
jedevc: Sigh, actually, no, this can't be removed because of a bug in how that engine socket is mounted.

Will open a separate PR to fix that.

helm/dagger/templates/engine-daemonset.yaml
```diff
@@ -54,9 +58,17 @@ spec:
...
           image: {{ if .Values.engine.image.ref }}{{ .Values.engine.image.ref }}{{ else }}registry.dagger.io/engine:v{{ .Chart.Version }}{{ end }}
           imagePullPolicy: {{ .Values.engine.image.pullPolicy }}
-          {{- if .Values.engine.args }}
+          {{- if or .Values.engine.port .Values.engine.args }}
           args:
+            {{- if .Values.engine.port }}
+            - "--addr"
+            - "tcp://0.0.0.0:{{ .Values.engine.port }}"
+            - "--addr"
+            - "unix:///run/dagger/engine.sock"
```
gerhard, line 67: I just loaded the `/run` change which came in effect since 2015: https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard . Good change 👍

gerhard (approved):
Great job @jholm117 ! This is my favourite Helm chart external contribution to date 💪

Thank you @jedevc for shepherding this and adding it all the last touches 🙏

#shipit

jedevc (approved):

PR #9847, opened 2025-03-12
feat: add flag to limit LLM access

cwlbraa:
basic wiring, session attachable, and remote-module-name matching all working.

this works:
```
dev && with-dev dagger shell --allow-llm=all -m https://github.com/shykes/toy-programmer -c "go-program \"greet me $(date)\""
```

this fails:
```
dev && with-dev dagger shell -m https://github.com/shykes/toy-programmer -c "go-program \"greet me $(date)\""
```

TODO:
- [x] actual prompting via frontend cc @vito 
- [x] `dagger call` wiring
- [x] pluralize flag! it's clearer now it's gotta be multiple modules if you've got multiple remote dependencies using LLM functionality.
- [ ] figure out what happens when a git module source makes a module out of a subdirectory
- [x] prompt history


engine/session/prompt.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+
+	return &PromptResponse{
+		Response: userResponse,
+	}, nil
+}
+
+// not threadsafe, must be holding promptMutex
+func (a *LLMAllowedModuleHistory) load() error {
```
cwlbraa, line 91: these methods are AI slop. i cleaned them up a bit and they work fine afaict, but feel free to suggest a redesign, i'm not married to the load/persist design at all, but it feels good enough.

vito:
Gonna merge and keep working on any followups on `llm`

PR #9853, opened 2025-03-13
ensure dev GHA jobs actually use dev engine

sipsma:
Need to test in GHA, will update w/ details once problem+fix confirmed


jedevc (approved):
Yeah, this is probably the best way for now. We can probably do some simplification in the future, but there's such a tangle of logic here :cry:

PR #9856, opened 2025-03-13
feat(sdk/java) enums

eunomie:
Support Java enums! 🎉 

Enums defined in the module **must** have the `@Enum` annotation. This allows the annotation processor to find them and register them and their possible values to the Dagger engine.

For instance:

```java
@Enum
public enum Severity {
  LOW,
  MEDIUM,
  HIGH
}
```

Only a single enum can be defined with a specific name, whatever their fully qualified name. For instance it you have two enums `io.dagger.module.mymodule.Severity` and `io.dagger.module.foo.Severity` an error will be raised.

Enums can be defined as independent objects or can be nested inside the main module object.
For instance this is valid:

```java
@Object
public class MyModule {
  @Enum
  public enum Severity {
    UNKNOWN,
    LOW,
    MEDIUM,
    HIGH,
    CRITICAL
  }

  @Function
  public String echo(Severity severity) {
    return severity.name();
  }
}
```

Enums requires a specific change in the way we deserialize the JSON. Java enums requires to call the `valueOf(String)` method on the enum class. See https://jakarta.ee/specifications/jsonb/3.0/jakarta-jsonb-spec-3.0#enum for more details.

Integration tests and docs have been added.

---

**edit: new type management system**

Rework the way we deal with types. Extract the type management in dedicated classes to add specific behavior depending on the types.

This also fixes a lot of issues with List. In short, when you have a `List<Something>` and try to deserialize
some JSON to it, it's not straightforward: `Something` disappears, so you just have `List` and `List` is an interface:

- the fact `Something` disappear is an issue when the type is required to deserialize. For instance an enum is serialized as a string, but deserialized into a specific type. If we have `["low", "medium", "high"]` to deserialize and only the type `List` we don't know if it's a list of string or a list of the enum. And as the enums needs a special call to be set we can't do it.
- `List` is an interface. We can't set something as an interface, it has to be a concrete type. For instance we have to use an `ArrayList`. But `ArrayList` is not the defined type.

All in all, the idea is here, when we are deserializing, to not deserialize as `List` but always as arrays. Then to convert to a `List`.

So for instance if the parameter type is `List<MyEnum>` we will deserialize the value into a `MyEnum[]` then call `Arrays.toList` on it. The usage of the array fixes the issues mentionned above, and from a JSON
perspective list or arrays are identical.

The new types exposes several methods, to know if it's a list or not, and to expose the different code blocks we need for the entrypoint generation. In particular a new method has been added when we want the class of a type
instead of just adding `.class` in the code generation.

Most of the `List` related behavior (to return or get a list) is covered in the `enums` integration tests.


sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -520,10 +561,9 @@ private static CodeBlock functionInvoke(ObjectInfo objectInfo, FunctionInfo fnIn
           .beginControlFlow("if (inputArgs.get($S) != null)", parameterInfo.name())
           .addStatement(
               CodeBlock.builder()
-                  .add("$L = (", parameterInfo.name())
-                  .add(paramType)
                   .add(
-                      ") $T.fromJSON(inputArgs.get($S), ",
+                      "$L = $T.fromJSON(inputArgs.get($S), ",
+                      parameterInfo.name(),
```
eunomie, line 566: This cast was useless as the type is defined

core/integration/module_java_test.go
helderco, line 1: What about?
- Returning an enum
- Lists of enums (either as arg or return)
- Descriptions

eunomie: Thanks for the comment, this is way more interesting java question that I thought 😅 

So most of that was working... but not to lists in all cases.
Roughly the way Java types are working is part of the type disappear. So `List<MyEnum>` is `List`. But as `MyEnum` requires a specific case, because it's serialized as a string, but `List` doesn't know it's a string or an enum, that makes things... interesting.

But I'm fixing that, and that will improve the way we deal with types.

eunomie: I pushed a new commit, that should fix the types management, and by the way the management of lists of enums (input or output). This is bigger than just enums, but was necessary due to the specificity of enums that can't be deserialized right away.
I've added more details in the PR body.

sdk/java/dagger-java-annotation-processor/src/test/resources/io/dagger/gen/entrypoint/entrypoint.java
```diff
@@ -195,17 +208,17 @@ private JSON invoke(JSON parentJson, String parentName, String fnName,
...
-          stringArray = (String[]) JsonConverter.fromJSON(inputArgs.get("stringArray"), String[].class);
+          stringArray = JsonConverter.fromJSON(inputArgs.get("stringArray"), String[].class);
         }
         Objects.requireNonNull(stringArray, "stringArray must not be null");
-        List ints = null;
+        List<Integer> ints = null;
         if (inputArgs.get("ints") != null) {
-          ints = (List) JsonConverter.fromJSON(inputArgs.get("ints"), List.class);
+          ints = Arrays.asList(JsonConverter.fromJSON(inputArgs.get("ints"), Integer[].class));
         }
```
eunomie, line 217: A lot of great things here, and a good example of the improvements regarding types:

`List ints` replaced by `List<Integer>`: the inner type is now better handled. So we are dealing with typed list and not just generic `List`.

The input is a JSON array of integers, but the end variable is a `List<Integer>`. So we would like to deserialize the input into the list. But from the JSON perspective it doesn't matter, this is an array.
And on the Java side, `Integer[].class` is what is the closest to the JSON type. `List.class` is generic and it's hard to correctly deserialize the JSON array into this type.

So the JSON array is deserialized into an array of the expected type (`Integer` here) then converted into a `List` so that the `List<Integer>` type of the variable is respected.

core/integration/module_java_test.go
```diff
@@ -288,6 +288,84 @@ func (JavaSuite) TestConstructor(_ context.Context, t *testctx.T) {
...
+	})
+
+	t.Run("can not use a value not defined in the enum", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		_, err := javaModule(t, c, "enums").
+			With(daggerCall("print", "--severity=FOO")).
+			Stdout(ctx)
+
+		require.Error(t, err)
```
TomChv, line 310: Ideally, the error should include the list of possible values for the enums https://docs.dagger.io/api/enumerations/
Maybe you can test that behaviour here for consistency with other SDKs?

core/integration/module_java_test.go
```diff
@@ -288,6 +288,84 @@ func (JavaSuite) TestConstructor(_ context.Context, t *testctx.T) {
...
+		out, err := javaModule(t, c, "enums").
+			With(daggerCall("array-to-string", "--severities=HIGH,LOW")).
+			Stdout(ctx)
+
+		require.NoError(t, err)
+		require.Equal(t, "HIGH,LOW", out)
+	})
+}
+
 func javaModule(t *testctx.T, c *dagger.Client, moduleName string) *dagger.Container {
```
TomChv, line 369: Just a quick note but I really like this pattern for testing features of an SDK, instead of inlining the source code like we do in others.
@helderco do you think we could migrate other SDKs to that pattern when we have some time?
This could be an opportunity to refactor tests and simplify/optimize them

It's also a very nice way to play with a standalone test without having to run the whole dev engine when something is not working properly. It's something I do very often when I work on a SDK.
Good idea 🚀 ⭐

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -69,9 +71,27 @@ ModuleInfo generateModuleInfo(Set<? extends TypeElement> annotations, RoundEnvir
     Set<ObjectInfo> annotatedObjects = new HashSet<>();
     boolean hasModuleAnnotation = false;
 
+    Map<String, EnumInfo> enumInfos = new HashMap<>();
+
     for (TypeElement annotation : annotations) {
       for (Element element : roundEnv.getElementsAnnotatedWith(annotation)) {
-        if (element.getKind() == ElementKind.PACKAGE) {
+        if (element.getKind() == ElementKind.ENUM && element.getAnnotation(Enum.class) != null) {
```
TomChv, line 78: Nit: A switch case may more appropriate, so if at some point you add other annotations, you don't have to change the whole block, just add a new statement.

 *(It's an early "optimization" so feel free to ignore it if you do not find that useful)*

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -203,8 +223,22 @@ ModuleInfo generateModuleInfo(Set<? extends TypeElement> annotations, RoundEnvir
...
     }
 
+    // Ensure only one single enum is defined with a specific name
+    Set<String> enumSimpleNames = new HashSet<>();
+    for (var enumQualifiedName : enumInfos.keySet()) {
+      String simpleName = enumQualifiedName.substring(enumQualifiedName.lastIndexOf('.') + 1);
+      if (enumSimpleNames.contains(simpleName)) {
+        throw new RuntimeException(
+            "The enum %s has already been registered via %s"
+                .formatted(simpleName, enumQualifiedName));
```
TomChv, line 233: This should already be triggered by the engine when you return the module typedef no? Is trowing an error early really useful?

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerType.java
```diff
@@ -0,0 +1,293 @@
...
+      return of(new TypeInfo(name, TypeKind.FLOAT.name()));
+    } else if (name.equals("double")) {
+      return of(new TypeInfo(name, TypeKind.DOUBLE.name()));
+    } else if (name.equals("void")) {
+      return of(new TypeInfo(name, TypeKind.VOID.name()));
+    } else if (name.endsWith("[]")) {
+      return of(new TypeInfo(name, TypeKind.ARRAY.name()));
+    } else {
+      return of(new TypeInfo(name, ""));
+    }
```
TomChv, line 116: Nit: Could be a switch case too with an if on default to check if it endsWith `[]`?

TomChv (commented):
Looks Good! Very complete PR with doc addition to make it perfect 🚀

I left a couple of nit comment, feel free to ignore them if you think they are not relevant (I'm not a Java programmer and I'm not familiar with the standard rules)

core/integration/module_java_test.go
helderco: We started out like that in Go, but adding Python and then TypeScript it was way harder to have parametrized tests with per-sdk sources because you have to open different files to compare. Then there's the extra boilerplate and files if every module in tests come from the file system. Making them inline and generated on demand made it much easier to maintain. Still, we do sometimes commit a `dagger init`ed module under `core/integration/modules` when it makes sense or makes things easier.

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
helderco: Yeah, I've spoken with Tom before on letting the SDK lean on errors returned by the engine, but haven't pushed for it and it's up to the SDK to maybe provide a earlier exit with a better error message. But it does make an SDK leaner and simpler to maintain to reuse the validations that engine already makes, when possible.

core/integration/module_java_test.go
TomChv: I was more interested for specific SDK tests, not for common integrations tests that tests multiples SDK, I didn't remember seeing that pattern except for interfaces.
Anyway thanks for the explanations :)

helderco (approved):
lgtm

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerType.java
eunomie: Yeah, I thought about that several times but never decided which one to pick 😅 
I'll give a new try

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
eunomie: Well, the result looks weird.

So I wrote a module with two enum with the same name `Severities`. The resulting typedef generated contains both as I removed the check:

```java
        .withEnum(
            dag().typeDef().withEnum("Severities")
                .withEnumValue("LOW")
                .withEnumValue("MEDIUM")
                .withEnumValue("HIGH"))
        .withEnum(
            dag().typeDef().withEnum("Severities")
                .withEnumValue("LOW")
                .withEnumValue("MEDIUM")
                .withEnumValue("HIGH")
                .withEnumValue("CRITICAL"));
```

I defined two functions, each of them dealing with a different enum:

```java
  @Function
  public String sev(Severities severity) {
    return severity.name();
  }

  @Function
  public io.dagger.modules.testenums.Severities sev2(String severity) {
    return io.dagger.modules.testenums.Severities.valueOf(severity);
  }
```

The `register` contains:

```java
                .withFunction(
                    dag().function("sev",
                        dag().typeDef().withKind(TypeDefKind.STRING_KIND))
                        .withArg("severity", dag().typeDef().withEnum("Severities")))
                .withFunction(
                    dag().function("sev2",
                        dag().typeDef().withEnum("Severities"))
                        .withArg("severity", dag().typeDef().withKind(TypeDefKind.STRING_KIND))))
```

I can call `sev` without any problem.
And when I'm trying to call `sev2` I have `! unknown command "sev2" for "dagger call"`.

While I agree it fails, that doesn't sounds like the right kind of error.

So I'd be tempted to keep the check in the SDK for now.

But I do agree we should centralize the logic in the engine. At the same time, I think it's also good to ensure the SDK is respecting the contracts, like we can't have two enums with the same name (and I guess that's also the case for other types, but that's a different story)

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerType.java
eunomie: done 👍

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
eunomie: done 👍

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
TomChv: It's strange, the engine should have catch the duplication :O

sdk/java/dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/IDAbleVisitor.java
```diff
@@ -66,6 +67,16 @@ TypeSpec generateType(List<Type> types) {
...
+                    .beginControlFlow("if (clazz.isEnum())")
+                    .addStatement(
+                        "$T valueOf = clazz.getMethod($S, $T.class)",
+                        Method.class,
+                        "valueOf",
+                        String.class)
+                    .addStatement(
+                        "return clazz.cast(valueOf.invoke(null, jsonb.fromJson(json, $T.class)))",
+                        String.class)
+                    .endControlFlow()
```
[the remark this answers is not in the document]
eunomie: Interesting, I'll have a look at that.
For now it's using name only, but I'll see to have the smoothest transition.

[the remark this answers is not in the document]
eunomie: 🤔 
That will be fun I think... The thing is in Java an enum member only have a value, not a name (or the other way around, but that's called `value`). And there's no strict way to define a specific value, only conventions. And an enum value can have any number of fields.

For instance:

```java
  public enum Severity {
    INFO("info-severity", 1),
    WARN("warn-severity", 2),
    ERROR("error-severity", 3);

    private final String severity;
    private final int criticality;

    Severity(String s, int c) {
      this.severity = s;
      this.criticality = c;
    }

    public String getSeverity() {
      return severity;
    }
    
    public int getCriticality() {
      return criticality;
    }
  }
```

And of course, there's no standard way to get an enum value based on the field values.
In this case there's a `Severity.valueOf(String)` that will convert a name to an enum, but it only stick to the `INFO`, `WARN`, etc.

I'm not entirely sure how that will work, but I'll probably move with this first, as it just uses default, standard, java enums and that works great. But I'll keep an eye on the name/value question to see how we can support those changes too.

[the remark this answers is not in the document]
helderco: For Java modules defining enums, I think it's ok not to be able to have a name different than a value. In GraphQL that destinction doesn't exist so we'll need to implement a custom directive to preserve the value if it's different from the name.[^1] Thus, in this case, the transition should be simple. Just need to update the API call for registering the enum, in the SDK.

[^1]: For context, GraphQL enums have only values, but the values are "names", in the sense that they must be valid identifiers (as in variable name).

Ideally everyone would use the same name and value when defining enums, but most SDKs allow setting them differently. So, since a Java module can depend on a Go module that has different values and names, how does the Java codegen handle that?

core/integration/module_java_test.go
eunomie: Fixed, added the line after this one 👍

sdk/java/dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/IDAbleVisitor.java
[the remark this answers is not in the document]
eunomie: If I have that in Go:

```go
type Severity string

const (
	Unknown Severity = "UNKNOWN_SEVERITY"
	Low Severity = "low_severity"
	Medium Severity = "MEDIUM_SEVERITY"
	High Severity = "HIGH_SEVERITY"
	Critical Severity = "CRITICAL_SEVERITY"
)
```

It will creates this in Java:

```java
public enum MyEnumSeverity {
    UNKNOWN_SEVERITY,
    low_severity,
    MEDIUM_SEVERITY,
    HIGH_SEVERITY,
    CRITICAL_SEVERITY
}
```

So it uses the values. As the values are valid identifiers, like for GraphQL enums, that works well.

eunomie:
I'm going to merge it. We can continue the enum discussion when we'll add the name in addition to the value, but my understanding is Java is close to GraphQL so that shouldn't be a problem (and is not yet done).

Thanks for the reviews 👍

Issue #5691, opened 2023-08-24
Prevent a `.dockerignore` papercut when instructing users on re-using Dockerfiles


starthal:
### What is the issue?

Related: #3654

The current [docs](https://docs.dagger.io/quickstart/429462/build-dockerfile/#reuse-existing-dockerfiles) mention using an existing Dockerfile and refactoring toward Dagger API calls.

> The good news here is that Dagger can natively run Dockerfiles with full compatibility

I think this is technically true but it misled me to believe that the context behavior would match Docker as well.

I suggest those docs mention that not all Docker behaviors are automatically respected. It wasn't obvious why my build wasn't caching as expected until I searched for an issue here.

Or, the cookbook could include a snippet for explicitly iterating the lines of an existing `.dockerignore` and passing them to the exclude API -- assuming they'd be evaluated equivalently (Go's `filepath.Match` ?)

PR #9857, opened 2025-03-13
feat(engine): honor the dockerignore file if present during docker build

rajatjindal:
This is a POC for fixing #5691 

During buildDocker function, we check if a docker ignore file exists and use that to exclude those paths before building the container.

TODO's:

- [ ] Verify that the name of the file follow the same pattern as supported by Docker
- [ ] Ensure that we do not error out if dockerignore file does not exist
- [ ] Add test cases

Alternative considerations:

I read through on how buildkit do it. while it is possible to do the same in dagger [here](https://github.com/dagger/dagger/blob/main/core/schema/host.go#L246), it will be a breaking change, and I also think will surprise our existing users. Therefore keeping the change isolated in `buildDocker` function.

### DIFF
--- a/core/schema/directory.go
+++ b/core/schema/directory.go
@@ -1,6 +1,7 @@
 package schema
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io/fs"
@@ -8,6 +9,7 @@ import (
 
 	"github.com/dagger/dagger/core"
 	"github.com/dagger/dagger/dagql"
+	"github.com/moby/patternmatcher/ignorefile"
 )
 
 type directorySchema struct {
@@ -98,7 +100,7 @@ func (s *directorySchema) Install() {
 		dagql.Func("export", s.exportLegacy).
 			View(BeforeVersion("v0.12.0")).
 			Extend(),
-		dagql.Func("dockerBuild", s.dockerBuild).
+		dagql.NodeFunc("dockerBuild", s.dockerBuild).
 			Doc(`Builds a new Docker container from this directory.`).
 			ArgDoc("dockerfile", `Path to the Dockerfile to use (e.g., "frontend.Dockerfile").`).
 			ArgDoc("platform", `The platform to build.`).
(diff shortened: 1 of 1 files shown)


core/container.go
```diff
@@ -446,7 +447,7 @@ func (container *Container) Build(
 	}
 
 	inputs := map[string]*pb.Definition{
-		dockerui.DefaultLocalNameContext:    contextDir.LLB,
+		dockerui.DefaultLocalNameContext:    buildctxDir.LLB,
```
jedevc, line 450: naming nit: `contextDir` should be for the `Context`, `dockerfileDir` should be the name for the `Dockerfile`.

core/integration/directory_test.go
```diff
@@ -883,6 +884,114 @@ CMD cat /secret
...
+		WORKDIR /src
+		COPY . .
+		`)
+
+		output, err := src.DockerBuild().Directory("/src").File("foo.txt").Contents(ctx)
+		require.NoError(t, err)
+		require.Contains(t, output, "content in foo file")
+	})
+
+	t.Run("use .dockerignore file by default", func(ctx context.Context, t *testctx.T) {
```
jedevc, line 902: Fun little edge case lol. https://github.com/dagger/buildkit/blob/f7cf40f994b0ef850330f403e478258c5eab3766/frontend/dockerfile/dockerfile_test.go#L498-L498

`Dockerfile.dockerignore` is valid as well :tada:

core/schema/container.go
```diff
@@ -812,6 +812,7 @@ func (s *containerSchema) build(ctx context.Context, parent *core.Container, arg
 	return parent.Build(
 		ctx,
 		dir.Self,
+		dir.Self,
```
jedevc, line 815: Shouldn't we also handle it in this case?

core/schema/directory.go
```diff
@@ -343,26 +346,98 @@ type dirDockerBuildArgs struct {
...
 		return nil, err
 	}
+
+	var destdir dagql.Instance[*core.Directory]
+	if len(excludes) > 0 {
+		err = s.srv.Select(ctx, parent, &destdir,
+			dagql.Selector{
+				Field: "withDirectory",
+				Args: []dagql.NamedInput{
+					{Name: "path", Value: dagql.NewString("./filtered")},
```
jedevc, line 405: This can just be the empty string, which puts it at the root, so we don't need the second selector :tada:

engine/server/server.go
```diff
@@ -179,6 +180,10 @@ type NewServerOpts struct {
 	BuildkitConfig *bkconfig.Config
 }
 
+func MyBuild(ctx context.Context, c client.Client) (_ *client.Result, err error) {
+	return nil, fmt.Errorf("FROM INSIDE MY BUILD")
```
jedevc, line 184: :eyes: testing code?

rajatjindal: yep... this was to get deep into buildkit to understand the differences in behavior. I will cleanup the code.

core/schema/container.go
rajatjindal: I think we originally discussed to handle this only in docker build function to avoid any surprises in other scenarios.

core/container.go
rajatjindal: done. thank you for this great suggestion. 🎉

core/schema/directory.go
rajatjindal: hmm i did tried this but that returns the dir with the excluded files (I think because src and dest dir are same 🤔 from llb perspective?)

i renamed the dir from `filtered` to `buildctx` and also added a comment to make it clearer why we are doing this.

core/schema/container.go
rajatjindal: I had never used container.Build and they are indeed same. I updated the code to cover that as well. 👍

core/schema/directory.go
rajatjindal: Hi @jedevc, i think this is the only blocking comment from merging this code. please let me know your thoughts about this one. 🙏

marcosnils (approved):
LGTM! left a trivial nit comment.

core/integration/directory_test.go
```diff
@@ -883,6 +884,134 @@ CMD cat /secret
...
+			WithNewFile("Dockerfile",
+				`FROM golang:1.18.2-alpine
+		WORKDIR /src
+		COPY . .
+		`)
+
+		output, err := src.DockerBuild().Directory("/src").File("foo.txt").Contents(ctx)
+		require.NoError(t, err)
+		require.Contains(t, output, "content in foo file")
+	})
```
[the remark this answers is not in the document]
rajatjindal: I wanted to have a testcase so that we have explicit coverage for the scenario instead of relying on an implicit validation elsewhere. But also happy to remove it if you insist.

core/schema/directory.go
rajatjindal: Hi @jedevc, a kind reminder for this comment.

core/integration/directory_test.go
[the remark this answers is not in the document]
jedevc: @marcosnils yup, we should have both! `Container.build` and `Directory.dockerBuild` are different methods. If we don't have both tests, we run the risk that one of them drifts in behavior.

core/container.go
```diff
@@ -391,7 +393,7 @@ func (container *Container) Build(
 ) (*Container, error) {
 	container = container.Clone()
 
-	container.Services.Merge(contextDir.Services)
+	container.Services.Merge(dockerfileDir.Services)
```
jedevc, line 396: I think we probably need to include the `contextDir` services as well?

core/schema/directory.go
rajatjindal: I think I can use the new `dir.filter` api (https://github.com/dagger/dagger/pull/9976) to simplify this further. but do we want to hold off until that api is merged?

core/container.go
jedevc: The services map is used to spin up any dependent services right before we actually execute an operation that might require the use of them. We need to track them correctly, or we can fail to evaluate certain things.

It shouldn't affect things for this code here, but if we ever change the calling code where `contextDir` isn't a subset, then we'd hit this codepath.

core/schema/directory.go
```diff
@@ -352,26 +354,105 @@ type dirDockerBuildArgs struct {
...
 }
 
-func (s *directorySchema) dockerBuild(ctx context.Context, parent *core.Directory, args dirDockerBuildArgs) (*core.Container, error) {
-	platform := parent.Query.Platform()
+func getDockerIgnoreFileContent(ctx context.Context, parent dagql.Instance[*core.Directory], filename string) ([]byte, error) {
+	file, err := parent.Self.File(ctx, filename)
+	if err != nil {
+		if strings.Contains(err.Error(), "no such file or directory") {
+			return nil, nil
+		}
```
jedevc, line 362: `errors.Is()` please! Manual comparison of error strings is very fragile.

core/schema/directory.go
jedevc: Just merged it, feel free to use it :tada:

core/schema/directory.go
rajatjindal: I don't think we have that here. This error is a custom error coming from here: https://github.com/dagger/dagger/blob/main/core/directory.go#L238

rajatjindal:
Hi @jedevc, I think this PR is also ready for final review/merge.

core/schema/directory.go
jedevc: fair enough - we should update that error then.

rajatjindal: 👍, I had to return `syscall.ENOENT` to ensure the error returned is still "no such file or directory", otherwise this would become a backward incompatible change.

jedevc (approved):

PR #9860, opened 2025-03-13
feat: add dagger develop --recursive flag

jedevc:


TomChv (approved):

cmd/dagger/module.go
```diff
@@ -507,88 +512,157 @@ This command is idempotent: you can run it at any time, any number of times. It
...
-			}
-			// if SDK is set but source path isn't and the user didn't provide --source, we'll use the default source path
-			if modSDK != "" && modSourcePath == "" && developSourcePath == "" {
-				inferredSourcePath, err := inferSourcePathDir(srcRootAbsPath)
-				if err != nil {
-					return err
+			ctx, span := Tracer().Start(ctx, "develop")
+			defer telemetry.End(span, func() error { return err })
+
+			eg, ctx := errgroup.WithContext(ctx)
```
cwlbraa, line 532: hey look at that, it's the (elusive) correct use case for errgroup.WithContext: cancel the whole thing when the first one fails.

core/integration/module_cli_test.go
```diff
@@ -562,6 +562,27 @@ func (CLISuite) TestDaggerDevelop(ctx context.Context, t *testctx.T) {
...
+		c := connect(ctx, t)
+
+		base := c.Container().From(golangImage).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work/dep").
+			With(daggerExec("init", "--source=.", "--name=dep", "--sdk=go")).
+			WithExec([]string{"rm", "dagger.gen.go"}).
+			WithWorkdir("/work").
+			With(daggerExec("init", "--source=.", "--sdk=go")).
+			With(daggerExec("install", "./dep")).
```
cwlbraa, line 576: probably doesn't need a test, but do you know if there's any way to mess things up here with symlinks? my intuition is that they should just transparently work, but i don't have anything to back that up with.

cwlbraa (approved):
lgtm!

jedevc:
@nipuna-perera pointed out that this would be a bit less useful in a daggerverse-style repo setting. Since you'd need to create a dummy top-level module with only dependencies, so you could run this command.

Maybe that's acceptable?

nipuna-perera:
I haven't tested it yet, but I don't see it as a huge drawback. It may even be beneficial for testing if I can call functions from the top level without changing dir.

shykes (approved):
Love it!

PR #9863, opened 2025-03-14
tests: add llm integration tests

jedevc:


core/integration/llm_test.go
```diff
@@ -0,0 +1,127 @@
...
+	name, err := program.Name(ctx)
+	require.NoError(t, err)
+	out, err := goGitBase(t, c).
+		WithWorkdir("/src").
+		WithMountedFile(name, program).
+		WithExec([]string{"go", "run", name}).
+		Stdout(ctx)
+	require.NoError(t, err)
+	require.Regexp(t, re, out)
+}
```
cwlbraa, line 127: 🤨 this feels like the wrong thing to test, like doesn't all this output come from the replay file? maybe once we have code that regenerates the stubs with an actual LLM, this would make sense as a test-the-test assertion.

core/llm.go
```diff
@@ -284,6 +305,18 @@ func (r *LLMRouter) LoadConfig(ctx context.Context, getenv func(context.Context,
 	if err != nil {
 		return err
 	}
+
+	historyReplay, err := getenv(ctx, "LLM_HISTORY_REPLAY")
```
cwlbraa, line 309: 🤔 does this force us to only do dagger-in-dagger style integration tests instead of being able to call the API directly when appropriate? 

or is there some way to do `c := connect(ctx, t).WithEnv(LLM_HISTORY_REPLAY=%x)`?

core/integration/llm-hello-world.golden
```diff
@@ -0,0 +1,175 @@
+[
+  {
+    "role": "user",
+    "content": "You are an expert go programmer. You have access to a workspace.\nUse the read, write, build tools to complete the following assignment.\nDo not try to access the container directly.\nDon't stop until your code builds.\n\nAssignment: write a hello world program\n",
+    "token_usage": {
+      "InputTokens": 0,
+      "OutputTokens": 0,
+      "TotalTokens": 0
+    }
+  },
```
cwlbraa, line 10: did you write this by hand?

core/integration/llm_test.go
jedevc: Yeah, it *does* come from the output file, so true this does feel a *bit* weird.

But also - we need something here that tests the LLM, since once we plumb through the replay regeneration, then we need something to check that the output actually gets generated reasonably.

core/integration/llm-api-limit.golden
```diff
@@ -0,0 +1,52 @@
...
+  {
+    "role": "assistant",
+    "content": "",
+    "token_usage": {
+      "InputTokens": 411,
+      "OutputTokens": 48,
+      "TotalTokens": 459
+    }
+  }
+]
```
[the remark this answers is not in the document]
jedevc: Yeah, I didn't really intend this to be handrolled - I did consider it briefly, but I kinda think something a bit more automated will be a bit easier to manage? I don't want to have to rework all our tests by hand every time we change how the LLM functionality works.

core/integration/llm-hello-world.golden
jedevc: Ah, no. This is actually from the new `historyJSON` field, which dumps all of this, so we can replay it later.

My idea is that when you run the llm tests with a specific flag, it'll run using OpenAI or similar, and then the committed replay files will be the ones run on PRs. I'll work on plumbing this through today.

:thinking: we should also have some *super* dummy tests that can run on `main` that test that the client implementations keep working - since this coverage doesn't give us that.

core/llm.go
jedevc: Sigh, yeah. We can only call the API nested.

I like the `WithEnv` idea a lot, but it's not really plumbed in, and it's a bit tricky, since `connect` doesn't *always* spawn a new subprocess with an env, e.g. in our tests where we're doing nested dagger.

:thinking: maybe we could put it in the model name? Something like `replay/<base64-encoded-replay>`? Or another parameter there? I don't really mind, the logic is pretty similar either way.

cwlbraa: > Something like replay/<base64-encoded-replay>? Or another parameter there? 

i feel like base64ing these replays could get waaaay too large, especially with the automatic recording... any tool call that returns a big text blob is gonna blow that field up real real big. maybe size isn't a real problem though, idk. an additional parameter sounds nice, though.

core/integration/llm-api-limit.golden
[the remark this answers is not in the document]
cwlbraa: yep agreed. i was being too indirect, but trying to ask if it makes any sense to you to want to split the user & assistant prompts into separate files. that said, i was asking that before i knew that these were coming out of `historyJSON`... with the automation it prolly doesn't matter anyways.

core/integration/llm_test.go
```diff
@@ -22,96 +24,102 @@ func TestLLM(t *testing.T) {
...
-
-type Test struct{}
-
-func (m *Test) Run(
-	assignment string,
-	model string, // +optional
-) *dagger.Container {
-	return m.llm(assignment, model).ToyWorkspace().Container()
+type LLMTestCase struct {
+	Ref   string
```
cwlbraa, line 28: aw man i wish i had thought of popping the examples out of the file and onto disk, it reads soooooo much better now.

core/llm.go
jedevc: Agreed, spun off a discord convo, since this affects our public facing API: https://discord.com/channels/707636530424053791/1351531596288753716/1351531917807321149

PR #9866, opened 2025-03-14
update default engine unix sock to /run/dagger/engine.sock

sipsma:
For less confusion and more consistency in naming, this updates the engine listener to use `unix:///run/dagger/engine.sock`.

For now, I'm leaving around the old `unix:///run/buildkit/buildkitd.sock` too to avoid backwards incompatibility. The two socks are just different listeners for the same server so there's no harm in both existing for now.

This updates everywhere I could find `buildkitd.sock` across Helm charts, various docs on integrations, etc. So it's worth a double check those all look correct to others now.

### DIFF
--- a/.changes/unreleased/Changed-20250317-102936.yaml
+++ b/.changes/unreleased/Changed-20250317-102936.yaml
@@ -1,0 +1,9 @@
+kind: Changed
+body: |-
+  The default unix socket used to connect to the engine is now at /run/dagger/engine.sock
+
+  The previous socket path still exists for backwards compatibility but may be removed in a future version.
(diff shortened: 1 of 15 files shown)


jedevc:
I know there's also references in the `connhelpers` - through `dial-stdio`. We should also update all references to buildkit's `appdefaults.Address`, e.g.

https://github.com/dagger/dagger/blob/758604428f70ac78df9106017e2dfa2f62436ecf/cmd/dialstdio/main.go#L25-L25

jedevc (commented):
:tada: neat! Probably worth adding a release note, since the old endpoint is now deprecated.

jedevc (approved):

PR #9868, opened 2025-03-15
fix incorrect Go SDK version.gen.go path

sipsma:
[A recent change](https://github.com/dagger/dagger/pull/9531) modified the path of the version.gen.go file the Go SDK uses to know which version of the CLI to autodownload. However, the path that our release CI uses was not changed, so we ended up in a situation where the new path wasn't changed (stayed on v0.16.2) but the old (unused) one was.

This manually corrects the version.gen.go file and updates CI with the correct path.


jedevc (approved):
Oops :facepalm:

PR #9876, opened 2025-03-17
ci: run provision tests in serial

jedevc:
These provision tests are pretty disk hungry, since they all start their own docker daemons. So we should avoid spinning up them all in parallel, or we're gonna consume too much disk all at once.

This is an attempt to fix the constantly failing cgroupsv2 jobs on main!


TomChv:
Seems like it has a positive impact based on https://v3.dagger.cloud/dagger/traces/52c65005ebb57d3df9996327c375ea0c?listen=48e834582cbc0868&listen=02672aff0587a099, the job isn't failing anymore (except the flackes because of cache key but it's unrelated)

jedevc:
cc @gerhard @sipsma - as discussed we'll just increase the disk space.

PR #9877, opened 2025-03-17
ci: Add a simpler `call` action for Alternative CI Runners

gerhard:
After all Alternative CI jobs started failing since #9673, this introduces a separate `call` action which is only used by these alternative CI runners. We want something simpler that has a separate lifecyle from the main `call` action that we use in the primary CI runners.

The problem with the current `call` action logging config is that GitHub Actions UI breaks, as well as the Namespace UI, so troubleshooting the source of this issue is very difficult.

I even downloaded the logs locally, which did not make it easier to find the issue. FTR:

    ~/Downloads/logs_35440011783 on ☁️
    ❯ ls -lah
    total 264M
    drwx------ 11 gerhard staff        352 Mar 10 19:24  .
    drwxr-xr-x 59 gerhard _lpoperator 1.9K Mar 10 19:24  ..
    -rw-rw-r--  1 gerhard staff        28M Nov 30  1979 '0_docs-lint-on-namespace-remote-engine _ remote-dagger-engine.txt'
    -rw-rw-r--  1 gerhard staff        15M Nov 30  1979 '1_docs-lint-on-namespace-local-engine _ local-dagger-engine.txt'
    -rw-rw-r--  1 gerhard staff        86M Nov 30  1979 '2_test-cli-engine-on-namespace-remote-engine _ remote-dagger-engine.txt'
    -rw-rw-r--  1 gerhard staff        30M Nov 30  1979 '3_sdk-go-on-namespace-remote-engine _ remote-dagger-engine.txt'
    -rw-rw-r--  1 gerhard staff        30M Nov 30  1979 '4_sdk-python-on-namespace-remote-engine _ remote-dagger-engine.txt'
    -rw-rw-r--  1 gerhard staff        31M Nov 30  1979 '5_sdk-typescript-on-namespace-remote-engine _ remote-dagger-engine.txt'
    -rw-rw-r--  1 gerhard staff        15M Nov 30  1979 '6_sdk-go-dev-on-namespace-local-engine _ local-dagger-engine.txt'
    -rw-rw-r--  1 gerhard staff        16M Nov 30  1979 '7_sdk-typescript-dev-on-namespace-local-engine _ local-dagger-engine.txt'
    -rw-rw-r--  1 gerhard staff        16M Nov 30  1979 '8_sdk-python-dev-on-namespace-local-engine _ local-dagger-engine.txt'

What's most peculiar about this issue is that the trace in Dagger Cloud doesn't show any issues:
https://v3.dagger.cloud/dagger/traces/57d1391e843c1d8a7cfb473f60c034e2

### DIFF
--- a/.github/actions/call-ci-alt-runner/action.yml
+++ b/.github/actions/call-ci-alt-runner/action.yml
@@ -1,0 +1,97 @@
+name: "dagger call for Alternative CI Runners"
+description: ""
+
+inputs:
+  function:
+    description: "The Dagger function to call"
+    required: true
+
+  module:
+    description: "The Dagger module to call"
+    default: "."
+    required: false
+
+  version:
+    description: "Dagger version to run against"
+    default: "v0.16.3"
+    required: false
+
+  dev-engine:
+    description: "Whether to run against a dev Engine"
+    default: "false"
+    required: false
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup
+      shell: bash
+      run: |
+        mkdir -p ${{ runner.temp }}/actions/call
+        touch ${{ runner.temp }}/actions/call/local-envs
+
+    - name: Install dagger
+      shell: bash
+      env:
+        DAGGER_VERSION: "${{ inputs.version }}"
+      run: |
+        if [[ -x "$(command -v dagger)" ]]; then
+          echo "::group::Checking dagger"
+          version="$(dagger --silent version | cut --fields 2 --delimiter ' ')"
+          if [[ "$version" != "$DAGGER_VERSION" ]]; then
+            echo "dagger ${version} is installed, but needed ${DAGGER_VERSION}"
+            exit 1
+          fi
+          echo "::endgroup::"
+        else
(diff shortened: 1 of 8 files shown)


gerhard:
That's a great idea @TomChv! Changing it now 👍

gerhard:
> @jedevc: Honestly, I see no reason to not just replace the old `call` and remove it entirely.
> 
> The only place we use it now properly is in `publish.yml`, I think we can just use our dagger-for-github action for that as well.

That is a good idea, and I'm keen to leave it for a separate PR since `publish` has a `release-report` with a temporary workaround which will make this PR bigger than fixing the immediate problem.

This is a good step towards removing `call` entirely, since it decouples Alternative CI Runners from it, and it only leaves the following depending on it:
- `benchmark.yml`
- `publish.yml`

TomChv (approved):

jedevc (approved):

gerhard:
Will fix `testdev-cgroupsv2` separately, merging this now so that the `Alternative CI Runners` jobs go green ✅

PR #9883, opened 2025-03-18
docs: update social sharing image

jasonmccallister:
Updates the image shared via social links.


mircubed (approved):

docs/docusaurus.config.ts
```diff
@@ -131,15 +131,15 @@ const config: Config = {
...
-        content: `${url}/img/dagger-factory-share.jpg`,
+        content: `${url}/img/daggernaut-carpenter-robots-share.jpg`,
       },
       {
         name: "author",
         content: "Dagger",
       },
       {
         property: "twitter:image",
-        content: `${url}/img/dagger-factory-share.jpg`,
```
vikram-dagger, line 142: Are we using this older image? If not, we should delete it as well.

vikram-dagger (commented):
See comment

docs/docusaurus.config.ts
jasonmccallister: Probably not, but maybe worth keeping for a while for the cached links?

vikram-dagger (approved):

PR #9894, opened 2025-03-18
docs: Move content-related images and remove unused images

vikram-dagger:
Cleaning up the root img/ directory


vikram-dagger:
@mircubed @jasonmccallister please check that you are not relying on these image paths in other locations, like the dagger.io website.

jasonmccallister:
Not sure why we are moving these into index? I get cleaning up the old images. 

I believe some of those urls are used in the dagger.io site.

vikram-dagger:
> Not sure why we are moving these into index? I get cleaning up the old images.

Mainly because the `img/` root is for common images cutting across docs versions (like the Discord, GitHub icons etc), and `current_docs`/`versioned_docs` is for images belonging to a specific version of the docs. I know we only have one version `current_docs` atm but this separation was useful previously when we had two versions of the docs running in parallel, it made it easier to see which files belonged to which version.
 
> I believe some of those urls are used in the dagger.io site.

Yup, I thought so too, so this needs a corresponding update in the website.

jpadams:
> Not sure why we are moving these into index? I get cleaning up the old images.
> 
> I believe some of those urls are used in the dagger.io site.

We confirmed that nobody else is relying on the location of these images, so we're free to move them.

jpadams (approved):
👍

PR #9839, opened 2025-03-11
shell: make .help leaner

helderco:
⚠️ May be superceded by:
- https://github.com/dagger/dagger/pull/9895

Removed dependencies and stdlib. Moved builtins to the bottom.

## Before

```
BUILTIN COMMANDS
  .core         Load any core Dagger type
  .deps         Dependencies from the module loaded in the current context
  .help         Show documentation for a command, a module, or a function
  .install      Install a dependency
  .login        Log in to Dagger Cloud
  .logout       Log out from Dagger Cloud
  .stdlib       Standard library functions
  .uninstall    Uninstall a dependency
  .update       Update a dependency
  .use          Set a module as the default for the session

AVAILABLE MODULE FUNCTIONS
  version               -
  tag                   -
  mod-codegen           When set, module codegen is automatically applied when retrieving the Dagger source code
  mod-codegen-targets   -
  check                 Check that everything works. Use this as CI entrypoint.
  with-mod-codegen      Enable module auto-codegen when retrieving the dagger source code
  cli                   Develop the Dagger CLI
  source                Return the Dagger source code
  go                    Dagger's Go toolchain
  engine                Develop the Dagger engine container
  docs                  Develop the Dagger documentation
  scripts               Run Dagger scripts
  test                  Run all tests
  bench                 Run all benchmarks
  generate              Run all code generation - SDKs, docs, etc
  sdk                   Develop Dagger SDKs
  dev                   Creates a dev container that has a running CLI connected to a dagger engine

AVAILABLE MODULE DEPENDENCIES
  alpine           A Dagger Module to integrate with Alpine Linux
  dagger-cli       -
  dirdiff          -
  docusaurus       Docusaurus Dagger Module
  dotnet-sdk-dev   A dev module for dagger-dotnet-sdk.
  elixir-sdk-dev   A module for help developing Elixir SDK.
  go               -
  graphql          -
  helm             -
  notify           Send notifications
  php-sdk-dev      -
  ps-analyzer      The PowerShell Script Analyzer (PSScriptAnalyzer) for Dagger.
  python-sdk-dev   The Python SDK's development module.
  shellcheck       -
  version          Shared logic for managing Dagger versions
  wolfi            A Dagger Module to integrate with Wolfi Linux

STANDARD COMMANDS
  cache-volume   Constructs a cache volume for a given cache key.
  container      Creates a scratch container.
  directory      Creates an empty directory.
  engine         The Dagger engine container configuration and state
  git            Queries a Git repository.
[...]

PR #9895, opened 2025-03-18
shell: merge `.help` sections

helderco:
Supercedes https://github.com/dagger/dagger/pull/9839

More importantly it fixes an issue where shadowed functions were being shown.

## Bikeshedding

This puts module functions, dependencies, and stdlib commands under the same category. May want to break stdlib commands under their own "Core Functions" category for example.

See https://github.com/dagger/dagger/pull/9839#issuecomment-2734013528 for more context


cmd/dagger/shell_help.go
```diff
@@ -82,65 +83,150 @@ func (h *shellCallHandler) DependenciesList() string {
...
-				return f.CmdName(), f.Short()
-			}),
-		)
-		types = append(types, "<function>")
+	// NB: This grouping by Section may not seem useful right now but the way these
+	// groups are organized is very mercurial so leaving it in to make it easier
+	// to change again later.
+	groups := []shellDocUsagesSection{
+		{
+			"Available Functions",
```
shykes, line 92:
```suggestion
			"",
```

cmd/dagger/shell_help.go
```diff
@@ -82,65 +83,150 @@ func (h *shellCallHandler) DependenciesList() string {
...
-		types = append(types, "<dependency>")
+	types := "<command>"
+	if len(doc.Groups) > 2 {
+		types += " | <function>"
 	}
 
-	doc.Add("Standard Commands", nameShortWrapped(h.Stdlib(), func(c *ShellCommand) (string, string) {
-		return c.Name(), c.Short()
-	}))
+	doc.Add("Builtin Commands", nameShortWrappedIter(h.allBuiltinUsages()))
```
shykes, line 110:
```suggestion
	doc.Add("", nameShortWrappedIter(h.allBuiltinUsages()))
```

shykes (changes requested):
This is definitively an improvement, I think we should merge as-is, with just one tweak, see below.

## Requested change

Present the sections without an uppercase title. The grouping conveys all the information we need - the words in the titles get in the way. The `.help` loop is so tight, even small simplifications are worth it.

## Follow-up improvements

I agree splitting out the stdlib functions (after module functions, before builtins) would be worth it. But only if we preserve shadowing: each function in the search path should only appear once, and the command shown must match the command that will be called. Also, just like the other sections, this one should have no title. The hierarchy of the search path is implicit in the order of the sections - that's enough info.

helderco:
> [...] only if we preserve shadowing: each function in the search path should only appear once, and the command shown must match the command that will be called.

If you look at the options in https://github.com/dagger/dagger/pull/9839#issuecomment-2734013528, I used this PR to generate them. In this case it's **Option 1**, just without section titles. So it already takes into account shadowing, no matter if you group into sections or not.

This is how it looks:

Screenshot 2025-03-19 at 13 18 22

Note that I've split the current module into its own section so it's clearer it's not a function nor a dependency. Could group it with the dependencies (as in currently available modules by name).

Note also that if the module's constructor has required arguments the top level functions no longer show, but the module name still does, since you need its constructor to access those functions. Here's an example on the `modules/go` module:

Screenshot 2025-03-19 at 13 20 14

helderco:
> 1. Module constructor

Yeah, I wanted to organize it like that it just doesn't technically follow lookup logic. But I've argued that it's not an issue with builtins because of the `.` prefix and on second thought the module constructor will only show if there's no conflict with a function that's named the same. It's only an issue if someone expects the group order to reflect lookup order and get confused why the module gets shadowed. As long as we don't sell it like that I'm ok with it.

Will require a non-trivial change to have a different order in the groups than the lookup order here (because of how shadowing is applied in `.help`), but I already wanted to do that.

helderco:
Done, this is good to go.

jedevc (approved):

PR #9898, opened 2025-03-19
add TestAllowLLM, currently hanging mysteriously

cwlbraa:
putting this up as draft so maybe @jedevc can see something dumb i'm doing in here that's making this hang... i've tried invoking via shell and call plus various ways of passing the allowed module. still seems like regardless of what I do it hangs. 

the daggerCliBase technique + daggerForwardSecrets thing works for the TestAPILimit test right above this one, so if it's something in the test code, it's a typo I can't see even though i'm looking right at it... one thing that's not in this draft is that i'm actually calling this using the commented out impl for daggerForwardSecrets.

one other possibility is there is some code path that blocks on a tty read, and we don't have a tty here. tomorrow i'm gonna try to pivot to using `hostDaggerExec` and `newTUIConsole` so we can actually try to interact with the prompt that might or might not be displaying here.

### DIFF
--- a/core/integration/llm_test.go
+++ b/core/integration/llm_test.go
@@ -152,6 +152,32 @@ func (LLMSuite) TestAPILimit(ctx context.Context, t *testctx.T) {
 	}
 }
 
+func (LLMSuite) TestAllowLLM(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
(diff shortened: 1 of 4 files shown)


jedevc:
Yeahhhh I suspect this might be something with the prompt. I'd be pretty heavily inspired by the way that our `terminal` tests work, hopefully those should do it?

PR #9901, opened 2025-03-19
spin out docs into its own Dagger module

shykes:
- Shrink the monolith
- Move a shared dependency to sdk/php/dev. Bonus: remove duplicate code


shykes:
Keeping the dream alive :)

jedevc (approved):
Pushed a couple changes:
- Updated the `docs` CI job to use the check instead of manually calling `docs lint`.
- Updated the docs README, since the preview command has now changed.

LGTM, merging once green.

shykes:
Nice thank you!

PR #9803, opened 2025-03-07
docs: Update quickstart for Dagger shell

vikram-dagger:

Issue #8834, opened 2024-11-01
shell: make object construction less magic?


shykes:
## Problem

In `dagger call`, the user must express the value of each argument as single string value. In order to support object types (`Container`, `Directory`, `Service` etc), the CLI supports a "magic encoding" of these object types as a string. For example:

- A `Container` type can be encoded as a remote docker image address, for example `index.docker.io/alpine:latest`
- A `Directory` type can be encoded as either a remote git ref (`https://github.com/dagger/dagger#main:docs`) or a local directory (`./foo/bar`)
- etc.

Without this feature, `dagger call` would be simply useless. But, it is far from perfect. It doesn't support all possible values for objects; and the magic can be hard to explain and understand.

## Solution

With `dagger shell`, we have more options:

- *Pipelines as arguments* allow passing any object as argument. This works for core *and custom* objects
- *Builtins* allow exposing the full power of the core API as shell commands

In short: the dagger shell is expressive enough to no longer need magic encoding. So... maybe we should disable it?

| Before | After |
| -- | -- |
| `dagger call build --source ./frontend` | `build $(.context ./frontend)` |
| `dagger call build --source https://github.com/dagger/dagger` |  `build $(.git https://github.com/dagger/dagger)` |
| `dagger call test --db tcp://localhost:4242` | `test --db $(.tunnel tcp://localhost:4242)` |

etc

shykes:
wdyt @helderco ?

helderco:
I love the simplicity and explicitness 👍, I’m just concerned about the added verbosity. Both work interchangeably right now, so you can already avoid using the flags when working with `dagger shell` and showing it off in demos. Then see how it feels.

PR #9905, opened 2025-03-19
docs: Update cookbook and return values pages for Dagger Shell

vikram-dagger:


docs/current_docs/cookbook/cookbook.mdx
```diff
@@ -1505,9 +2409,23 @@ The following Dagger Function demonstrates how to use native-language concurrenc
 
 Execute a Dagger Function which performs different types of tests by executing other Dagger Functions concurrently.
 
-```shell
-dagger call --source=. run-all-tests
-```
+    <Tabs groupId="shell">
```
vikram-dagger, line 2412: Ping @helderco @jasonmccallister for double-check on this example

docs/current_docs/api/return-values.mdx
```diff
@@ -45,9 +45,23 @@ Here is an example of a Dagger Function that returns operating system informatio
...
+```shell
+dagger -c 'os-info ubuntu:latest'
+```
+</TabItem>
+<TabItem value="Dagger Shell">
+```shell title="Type 'dagger' to start Dagger Shell in interactive mode and enter the command below."
+os-info ubuntu:latest
+```
+</TabItem>
+<TabItem value="Dagger CLI">
```
helderco, line 59: Technically the shell is a part of the CLI. So to contrast with shell is `call` and `core`.

docs/current_docs/api/return-values.mdx
```diff
@@ -45,9 +45,23 @@ Here is an example of a Dagger Function that returns operating system informatio
...
+<TabItem value="System shell">
+```shell
+dagger -c 'os-info ubuntu:latest'
+```
+</TabItem>
+<TabItem value="Dagger Shell">
+```shell title="Type 'dagger' to start Dagger Shell in interactive mode and enter the command below."
+os-info ubuntu:latest
+```
+</TabItem>
```
helderco, line 58: I hope at some point this becomes the only option. Users should know (from basics, admonition, or notice) to either execute it in interactive mode or provide inline with `-c`.

docs/current_docs/cookbook/cookbook.mdx
helderco: You need `host` if you need to go outside the current module's context. If you start with `-n` it works like `dagger call` though. You could use some other path within the context but then there wouldn't be an example on reaching for a directory outside of the repo. 

So to avoid confusion I think it's good to keep `$(host)`, but should update the path in the `call` command as well.

docs/current_docs/cookbook/cookbook.mdx
```diff
@@ -447,15 +821,43 @@ The following Dagger module uses a constructor to set a module-wide `Directory`
...
+    <TabItem value="Dagger CLI">
     ```shell
     dagger call foo
     ```
+    </TabItem>
+    </Tabs>
 
 - Call the Dagger Function with a constructor argument. The specified directory (`/src/myapp`) is used:
 
+    <Tabs groupId="shell">
```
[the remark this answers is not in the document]
helderco: This works because the first `.` is interepreted as a command, pointing at a module in the current directory. The second `.` is an argument to the (first) `.` command. Being a `Directory` flag it resolves to the current directory.

You can also call the current module's constructor through it's name (show show before the `⋈` prompt symbol), so this can avoid confusion:
```shell
my-module . | run-all-tests
```

For curiosity, if you wanted to be even more specific:
```shell
my-module $(host | directory .) | run-all-tests
```
See issue:
- https://github.com/dagger/dagger/issues/8834

docs/current_docs/api/return-values.mdx
kpenfound: Yeah if you're referring to the tabbed `-c` vs in shell, I agree. The idea was that since this is all new, the default (first tab) is something a user can just copy/paste and run and hopefully eventually learn that going into shell is easier. I think once this isn't such a new feature we can drop the tabs

kpenfound (approved):
Looks good! +1 on switching `Dagger CLI` to`Dagger Call`

docs/current_docs/api/return-values.mdx
kpenfound: Yeah if you're referring to the tabbed -c vs in shell, I agree. The idea was that since this is all new, the default (first tab) is something a user can just copy/paste and run and hopefully eventually learn that going into shell is easier. I think once this isn't such a new feature we can drop the tabs

docs/current_docs/cookbook/cookbook.mdx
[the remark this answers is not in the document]
vikram-dagger: Done

docs/current_docs/cookbook/cookbook.mdx
vikram-dagger: @helderco I've updated this but I think the `call` command is good as is? I didn't understand why you suggested updating that path.

docs/current_docs/api/return-values.mdx
vikram-dagger: > Technically the shell is a part of the CLI. 
I know :) However I couldn't find a good heading for the tab here. I considered `dagger call/core` but when you see it on the page, it's odd that the other tab headings are non-commands and this one is a command. I'm open to suggestions that fit with the style/format of the other tabs, please add them.

docs/current_docs/api/return-values.mdx
vikram-dagger: > I hope at some point this becomes the only option. Users should know (from basics, admonition, or notice) to either execute it in interactive mode or provide inline with `-c`.

FTR, when we discussed this internally, the plan was to only display the above two options of interactive and inline. However we have seen in the past (from modules launch) that when we suddenly remove the "familiar way" from the docs, it leads to significant user confusion. So I am proposing a third tab, the "dagger call/core" option, to ease the migration friction. This also has the advantage of allowing users to see and map the meaning of required arguments from `dagger call` to `dagger`/`dagger -c`. Once users are more familiar with the new approach, this "dagger call/core" tab can be easily removed from all examples.

vikram-dagger: Also worth noting that in the sections which mostly see new users, the quickstart and the feature pages, the `dagger call` option is _not_ shown and screen recordings have been updated to primarily reflect the shell.
- https://github.com/dagger/dagger/pull/9803
- https://github.com/dagger/dagger/pull/9910

docs/current_docs/api/return-values.mdx
helderco: Ok, let's revisit afterwards then, when there's a better fit. Then we can change everywhere at once.

docs/current_docs/cookbook/cookbook.mdx
```diff
@@ -175,21 +293,75 @@ When working with private Git repositories, ensure that [SSH authentication is p
...
+    dagger <<EOF
+    clone https://github.com/dagger/dagger TAG v0.12.1 |
+      file /src/CHANGELOG.md |
+      contents
+    EOF
     ```
+    </TabItem>
+    <TabItem value="Dagger Shell">
+    ```shell title="Type 'dagger' to start Dagger Shell in interactive mode and enter the command below."
+    clone https://github.com/dagger/dagger TAG v0.12.1 | file /src/CHANGELOG.md | contents
```
helderco, line 328: General check for "Dagger Shell" tabs having the same multiple breaks as "System shell" EOF.

docs/current_docs/cookbook/cookbook.mdx
helderco: > I didn't understand why you suggested updating that path.

Oh, I think that was meant for something else, in "Set a module-wide default path", dagger call uses `/src/myapp` while shell uses `/home/admin/images`.

vikram-dagger: Fixed in https://github.com/dagger/dagger/pull/9936

PR #9915, opened 2025-03-20
add TestLLM/TestAllowLLM, TestLLM/TestAPILimit, and fix bug in nested client --allow-llm metadata propagation

cwlbraa:
this PR adds tests for the `--allow-llm` flag and its associated policies.

in the process of writing these tests I found and fixed a bug in nested client allow-llm propagation that made ExperimentalPrivilegedNesting clients always try to prompt the user  and did a little refactoring to make the code easier to navigate. 

As part of this PR, I intend to:
- [x] add an env var way of setting this flag (cc @samalba)
- [x] look for a way for the prompt functionality to fail fast in the cases where it knows its can't read from tty. currently it just hangs.
- [x] write tests that exercise the interactive prompt flow, emulating existing terminal tests. ~Currently I only exercise the happy-paths that don't prompt.~
- [x] requires https://github.com/dagger/dagger-test-modules/pull/8 is merged and references to those modules in this PR are fixed to point at dagger/dagger-test-modules instead of cwlbraa/dagger-test-modules.
- [ ] intentionally not doing: (automatically) testing the prompt-response-history functionality -- it's stateful enough to be annoying to test. happy to add something if somebody has an idea of how to work around the fact that all the hostDaggerCall tests share the same history file.

### DIFF
--- a/cmd/dagger/engine.go
+++ b/cmd/dagger/engine.go
@@ -91,7 +91,9 @@ func withEngine(
 		params.Interactive = interactive
 		params.InteractiveCommand = interactiveCommandParsed
 
-		params.PromptHandler = Frontend
+		if hasTTY {
+			params.PromptHandler = Frontend
+		}
 
 		// Connect to and run with the engine
 		sess, ctx, err := client.Connect(ctx, params)
--- a/engine/server/session.go
+++ b/engine/server/session.go
@@ -919,8 +919,10 @@ func (srv *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // size.
 func (srv *Server) ServeHTTPToNestedClient(w http.ResponseWriter, r *http.Request, execMD *buildkit.ExecutionMetadata) {
 	clientVersion := engine.Version
+	allowedLLMModules := execMD.AllowedLLMModules
 	if md, _ := engine.ClientMetadataFromHTTPHeaders(r.Header); md != nil {
 		clientVersion = md.ClientVersion
+		allowedLLMModules = md.AllowedLLMModules
 	}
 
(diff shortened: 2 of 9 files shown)


engine/server/session.go
```diff
@@ -933,7 +935,7 @@ func (srv *Server) ServeHTTPToNestedClient(w http.ResponseWriter, r *http.Reques
 			ClientStableID:    execMD.ClientStableID,
 			Labels:            map[string]string{},
 			SSHAuthSocketPath: execMD.SSHAuthSocketPath,
-			AllowedLLMModules: execMD.AllowedLLMModules,
+			AllowedLLMModules: allowedLLMModules,
```
cwlbraa, line 938: this is the bugfix

cmd/dagger/engine.go
```diff
@@ -91,7 +91,9 @@ func withEngine(
 		params.Interactive = interactive
 		params.InteractiveCommand = interactiveCommandParsed
 
-		params.PromptHandler = Frontend
+		if hasTTY {
+			params.PromptHandler = Frontend
+		}
```
cwlbraa, line 96: this isn't 100% correct, but an 80% solution. if you construct the process such that neither stdout or stderr are terminals, like `dagger ... 2>&1 | tee /tmp/output.txt`, this will not attempt to prompt even if stdin is a tty. detecting on stdin also has issues because you could pipe a response in or even a shell script. 

leaving PromptHandler as nil if there's no terminal to output to is a good 80% solution here, imo. the error that comes out of this is pretty explicit about what flag you need to pass, and doing this way will never leave the user in an infinite-hang-when-i-run-this-remotely situation.

vito (approved):

PR #9930, opened 2025-03-20
docs: adding database-agent example

jasonmccallister:
Adds an example agent

### DIFF
--- a/docs/current_docs/agents.mdx
+++ b/docs/current_docs/agents.mdx
@@ -51,6 +51,7 @@ Here are several example repositories containing examples of Dagger modules with
 - [test-debugger](https://github.com/kpenfound/greetings-api/blob/main/DEBUGGER_AGENT.md): an agent that automatically debugs failing tests in CI.
 - [technical-content-summarizer](https://github.com/jasonmccallister/technical-content-summarizer): an agent that accepts a URL, optional min and max character length and forbidden words and summarizes the content for a non-technical audience.
 - [swe-agent](https://github.com/kpenfound/greetings-api/blob/main/SWE_AGENT.md): an agent that gets assigned GitHub issues and solves them with pull requests.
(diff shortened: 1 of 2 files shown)


docs/current_docs/agents.mdx
```diff
@@ -51,6 +51,7 @@ Here are several example repositories containing examples of Dagger modules with
 - [test-debugger](https://github.com/kpenfound/greetings-api/blob/main/DEBUGGER_AGENT.md): an agent that automatically debugs failing tests in CI.
 - [technical-content-summarizer](https://github.com/jasonmccallister/technical-content-summarizer): an agent that accepts a URL, optional min and max character length and forbidden words and summarizes the content for a non-technical audience.
 - [swe-agent](https://github.com/kpenfound/greetings-api/blob/main/SWE_AGENT.md): an agent that gets assigned GitHub issues and solves them with pull requests.
+- [database-agent](https://github.com/jasonmccallister/database-agent): An agent that connects to an existing database and allows the user to ask plain language questions to explore and query a database.
```
vikram-dagger, line 54: We have two pages with examples, is that intentional? @kpenfound for feedback.

kpenfound: Ah that's my bad, I didn't catch that the PR creating the examples page didn't remove the original list. We should remove that list. Can be in a separate PR if you prefer @jasonmccallister

jasonmccallister: Was looking at that, maybe a partial/examples component would be good? With the ability to pass which section we want to show (e.g. show the AI examples)? This way we would have one complete list but be able to pull it into multiple pages as needed? Does that make sense? I'm happy to try and squeeze that into this PR.

vikram-dagger (approved):
Please fix the conflict before merging

PR #9935, opened 2025-03-21
Expose a Dagger module as an MCP server

tiborvass:
This introduces a new `dagger mcp` command that starts an MCP stdio server

## Usage
`dagger mcp` or `dagger mcp -m ref`, where the `ref` is a path to a local module or a [remote one](https://docs.dagger.io/api/remote-modules/)

## Implementation
In this version, each dagger function corresponds to an MCP tool.

When a function is called, the list of available tools is refreshed. This allows the LLM to handle objects it didn't know previously. Such is the case for a function returning a `dagger.Container`.

The MCP server lives on the engine side, and the `dagger mcp` command simply forwards the stdio pipes to it using a new Pipe session attachable.

### Integrating this command with your IDE / other MCP clients
`mcp_config.json`:
```json
{
    "mcpServers": {
        "dagger": {
            "command": "/path/to/dev/dagger",
            "env": {
                "_EXPERIMENTAL_DAGGER_RUNNER_HOST":"docker-container://dagger-engine.dev",
                "OPENAI_API_KEY":"toto"
            },
            "args": [
                "mcp",
                "-m",
                "ref/to/module"
            ]
        }
    }
}
```

### Analysis of Client support

Client | Resources | Prompts | Tools | Sampling | Roots | Notes
-- | -- | -- | -- | -- | -- | --
Claude Desktop App | ✅ | ✅ | ✅ | ❌ | ❌ | Dynamic reload of tools not supported -- we error on the prompts / resources as we don't support it
Goose | ❌ | ❌ | ✅ | ❌ | ❌ | Dynamic reload of tools is supported (via notification)
Cursor | ❌ | ❌ | ✅ | ❌ | ❌ | Supports tools. Dynamic reload not supported atm -- UX under experimentation
Zed | ❌ | ✅ | ❌ | ❌ | ❌ | Prompts appear as slash commands -- No current support for tools
Windsurf Editor | ❌ | ❌ | ✅ | ❌ | ❌ | Need pro sub to test
-- | -- | -- | -- | -- | -- | --

## Todo
- [ ] Currently, `OPENAI_API_KEY` or equivalent needs to be set (can be just `foo`) because MCP is a method on the LLM object even though MCP doesn't make API calls. This requirement will go away as we separate MCP from LLM.
- [ ] Currently `-s` or `--progress plain` needs to be specified so only stderr is used, as `dagger mcp` command uses both `stdin` and `stdout` for MCP communication. We could force `--progress plain`, or decide on a minimalist reporting when using `-s`.
- [ ] SSE forwarding coming soon.
- [ ] This doesn't yet use the multi-object API.
- [ ] Module constructors won't work if there are arguments to the constructor. Need to figure out the UX. Should it be like `dagger call` ? Or should we piggyback on `dagger shell` by instead doing `dagger -c "... | mcp"` ? (Would need to think how that could work on the client side...). Or `dagger mcp -c '...'` or `dagger -mcp -c '...'` ?
- [x] `*_id` functions should not be exposed to MCP. Apparently it's still exposed despite logic attempting to prevent it.
- [ ] Moar tests
- [ ] Nice IDE integration examples
- [x] Fix bug spewing logs upon closing connections

Co-authored by @grouville


cmd/dagger/mcp.go
```diff
@@ -0,0 +1,70 @@
...
+		cmd.SetContext(idtui.WithPrintTraceLink(cmd.Context(), true))
+		return withEngine(cmd.Context(), client.Params{}, mcpStart)
+	},
+	Hidden: true,
+	Annotations: map[string]string{
+		"experimental": "true",
+	},
+}
+
+// dagger -m github.com/org/repo mcp key1=val1 key2=val2
```
cwlbraa, line 38: what do you see these keys and values being passed to? are they module-level flags?

tiborvass:
@cwlbraa Just want to clarify that this `mcp` command is not really meant to be used by users, but by a program (generally an MCP client) that wants to start the MCP server. So it is up to that program what/they want to show for progress.

@grouville and I are finishing a barebones cli to be able to easily test this with an LLM. Will update the description.

Currently, the official mcp-go client doesn't forward os.Stderr and the upstream change may need some UX debate, but in my fork I set "cmd.Stderr = os.Stderr" in the mcp-go client so that our testing cli does show --progress plain with the dagger operations visible. It just gets in the way of the interactiveness (the prompt ">" is lost when there's stderr output). Anyway this is for the testing mcp client, but we need a general solution. `dagger mcp logs` ? Probably a bad idea, but open to anything that makes the UX better.

cmd/dagger/mcp.go
tiborvass: My bad, this is more of a TODO of how to handle module constructors. Forgot to mention it in the description.

cwlbraa:
> @cwlbraa Just want to clarify that this mcp command is not really meant to be used by users, but by a program (generally an MCP client) that wants to start the MCP server. So it is up to that program what/they want to show for progress.

it's client specific, but claude desktop for example has [pretty detailed debugging instructions](https://modelcontextprotocol.io/docs/tools/debugging#viewing-logs) and i'm curious if `~/Library/Logs/Claude/mcp*.log` includes these stderr plain logs.

> dagger mcp logs? Probably a bad idea

agreed this feels like kinda a bad idea, something we should leave to docker lol

cmd/dagger/mcp.go
cwlbraa: if it's possible they should retain the syntax of `call`, like --module-constructor=value

marcosnils:
@tiborvass  how does the dynamic tool support actually work?  IIUC I can only start `dagger mcp` with a single module so I don't seem to find the way to dynamically load modules in an already started dagger mcp server so they can be re-discovered by the client.

tiborvass:
@marcosnils My bad, the terminology is confusing. What we meant by dynamic is that prior to calling a function that returns a `dagger.Container` type, the container functions are not available, they become available or at least we try, but some clients don't handle the "tool list update" of MCP.

What you are referring to is also justified being called dynamic and I believe it would require a "load module" function exposed in the LLM Env API. But note that this also suffers from the same problem that apart from Goose, nothing handles tools list update.

grouville (approved):
In a good shape ✅ 

Some of the todos left, but good enough for beta testing. Everything is hidden, under the hidden `dagger mcp` command ; AND, everything is scoped to the `dagger mcp` subcommand -- no impact on main branch at all

How to test it (dev mode):

```shell
OPENAI_API_KEY=toto dagger_dev --progress plain mcp -m github.com/shykes/hello
```

#### Notes
- The options `--progress plain` or `-s` are mandatory atm to not have the `stdin` polluted. One of the todos is to override it when not set in the CLI
- One of the provider key is still mandatory as our `MCP` method is still part of `LLM`

Then you can paste this on stdio: 
```json
{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"capabilities":{},"clientInfo":{"name":"goose","version":"1.0.12"},"protocolVersion":"1.0.0"}}
{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}
{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"arguments":{"name":"Fred"},"name":"Hello_hello"}}
```

PR #9938, opened 2025-03-21
test: PHP Scalar, List and Void signatures

charjr:
Adds integration tests requiring that the PHP SDK can handle method signatures of:
- Scalar Kind
- List Kind
- Void Kind


TomChv (approved):

core/integration/testdata/modules/php/void-kind/dagger.json
jedevc, line 1: No objection to having it this way, but is there a reason to *not* have all this be one module, and have the module declare multiple types?

PR #9912, opened 2025-03-19
Agents quickstart

kpenfound:
This creates the first agent quickstart where you can create your own toy-programmer

PR #9941, opened 2025-03-21
redo agent quickstart

kpenfound:
messed up DCO on this one https://github.com/dagger/dagger/pull/9912


kpenfound:
Test with

```
dagger -m https://github.com/dagger/dagger/docs@pull/9941/head call server up
```

vikram-dagger (approved):

vikram-dagger:
I like this as a starting point and we should merge it, but there are some changes I could suggest for the next iteration:
- Explain concept "passing tools to the agent"
- Explain concept "agent loop"
- Add a capsule on how Dagger works (assuming these are users coming from the AI world so with no prior knowledge, we should explain very briefly the sandboxing, the SDKs, etc with links to more detail).
- Create the workspace module in the quickstart instead of using an external one (makes it feel less like a black box)
- TBD: build an agent that does tasks on an existing repo instead of a generic programmer bot (makes the "reason to do this" clearer)
- TBD: Demo call caching
- Expand it at the end to give the agent more tools/capabilities (possible example: install s3 module, prompt "when done, build the binary and upload to s3 using the s3 tool")
- Switch it to a multi-page format like the CI quickstart (imo makes it easier to read but I know there are differing opinions on this)

I also think the goal affects the structure: do we want the user to exit having a demo in hand, or do we want the user to exit knowing how to use this in an existing project? The current structure serves the first goal well, but I'm uncertain if a user would be comfortable creating their own agent for their own project.

PR #9967, opened 2025-03-25
test: PHP Object signatures for built-in types

charjr:
Adds 2 tests asserting that the PHP SDK can:
1. Receive an object (built-into-dagger) as an argument.
2. Modify the object.
3. Return the object.

First test asserts it receive and return a `File` with its contents capitalized.

Second test asserts it can receive and return a `Directory` with a new `File` added to it.


charjr:
@wingyplus Thank you! I have formatted it now. :slightly_smiling_face:

wingyplus (approved):
LGTM

Issue #7637, opened 2024-06-12
✨ Support `--depth` when using `dag.git`


Xiot:
### What are you trying to do?

Trying to pull down more history for a ref than just the single commit.

The request is to have a property on the `dag.git` to specify the `depth` (number of commits) to checkout

### Why is this important to you?

We currently have a couple different jobs within our workflow that rely on checking the differences between the given commit and its parent.

### How are you currently working around this?

Right now, we are using `dag.git` to do the initial checkout, then running a separate `git fetch --depth=2`, to pull the history.

For some reason, the separate fetch doesn't always pull down the history properly.

jedevc:
Yup, this seems very reasonable to me, in my head this is strongly related to all the API improvements in https://github.com/dagger/dagger/issues/8520#issue-2539515747.

As a note to anyone picking this up - the logic for this needs to end up in our custom buildkit git source here: https://github.com/dagger/dagger/blob/d4d759140e4964437f9468d6537ff7a9fd313837/engine/sources/gitdns/source.go#L516-L523

There are *two* git repositories involved in a clone - the "farm" which keeps track of most of the repository, and then the clone off of the farm to get the git dir that the user sees. When specifying a depth, we need to ensure that the "farm" has at least the required depth, and then we just need to set the depth correctly when cloning off of the farm. The rest of this, plumbing through from the dagger API should be pretty easy.

jedevc:
Fixed by https://github.com/dagger/dagger/pull/9980.

PR #9968, opened 2025-03-25
feat: add git diff functionality

jedevc:
Opening as a base to discuss!

x-linking https://github.com/dagger/dagger/issues/8520 for how this should interact with modules (tl;dr we can use contextual git).

## Usage

```
# get the diff of the working directory
$ dagger core host directory --path=. as-git working diff

# get the diff to a specified target
$ dagger core host directory --path=. as-git working diff --target=<commit-sha>

# determine if a git directory is dirty or not
$ dagger core host directory --path=. as-git working dirty
```

## Changes

- All `GitRef`s can now be `diff`ed against arbitrary commits.
- `GitRepository.working` is a variant of `Git.head` that gets the current state, but *includes* any changes to the working directory. This means that:
	- `GitRef.diff` with no args will return a non-empty diff
	- `GitRef.dirty` will return true
	- `GitRef.tree` will return a directory with the working directory changes
	- `GitRef.commit` returns a `<commit>-dirty`

### Caveats

- `GitRef.diff` for remote refs doesn't work very well because of shallow-cloning (see #7637, working on a fix for this as part of https://github.com/dagger/dagger/pull/9980)

## To debate

- Is this at all the right API?
	- We can always add *more*, but it's harder to take it away.
	- Can this be used as useful info for LLMs?
- Should `GitRef.diff` take a `String`? Or a `GitRef`?

Let the bikeshedding begin!

### DIFF
--- a/cmd/dagger/flags.go
+++ b/cmd/dagger/flags.go
@@ -803,7 +803,7 @@ func (v *gitRefValue) Get(ctx context.Context, dag *dagger.Client, _ *dagger.Mod
 		if ref := gitURL.Fragment.Ref; ref != "" {
 			return repo.Ref(ref), nil
 		}
-		return repo.Head(), nil
+		return repo.Working(), nil
 	}
 
 	// Otherwise it's a local dir path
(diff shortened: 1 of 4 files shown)


jedevc:
Leaving this as a note - but I don't think this is the right API.

Having dirty state encoded as refs is weird, and can result in inconsistencies. I was thinking about adding a new type, like `GitState`? That would at least keep operations type-safe.

But yeah, not particularly sure. Essentially, this just needs to be discoverable *somehow* so we could build it into agent tooling.

PR #9972, opened 2025-03-25
fix: docs recorder runtime rust panic

sagikazarmark:
This PR fixes the gif generator module for the documentation. Prior to this fix, the gifs showed a Rust error.

Initially, I thought it was due to a glibc version mismatch between the build and the runtime container, but after some brute force line-by-line debugging, I narrowed it down to the one in this PR.

I don't fully understand why that caused the problem, but it did (along with a number of other issues).

### DIFF
--- a/docs/recorder/main.go
+++ b/docs/recorder/main.go
@@ -30,7 +30,7 @@ func getTermcast(wdir *dagger.Directory) *dagger.Termcast {
 	return dag.Termcast(dagger.TermcastOpts{
 		Container: dag.Wolfi().
 			Container(dagger.WolfiContainerOpts{Packages: []string{"docker-cli", "curl"}}).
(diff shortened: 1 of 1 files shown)


vikram-dagger (approved):

vikram-dagger:
@sagikazarmark from this fix, it looks like there are maybe two `dagger` binaries installed? Would it make sense to remove one from the container, so as to avoid this type of error recurring?

sagikazarmark:
@vikram-dagger can you point me where you see the second binary installed? Because I can't find it.

I only changed the path where the Dagger binary is added.

vikram-dagger:
> @vikram-dagger can you point me where you see the second binary installed? Because I can't find it.
> 
> I only changed the path where the Dagger binary is added.

Ah sorry, I misunderstood the fix then. Ignore.

PR #9975, opened 2025-03-26
allow to disable streaming on OpenAI compatible engines

### LINKED ISSUE #9969: ✨ LLM: configurable streaming

Some LLM servers that serve OpenAI endpoints do not behave the same as OpenAI when it comes to streaming, such as llama.cpp or certain Azure setups.

We should have a configuration, like an environment variable `OPENAI_DISABLE_STREAMING`, to not use streaming for the OpenAI client requests.

_No response_

_No response_

eunomie:
Some engines are not able to managed tools and streaming at the same time, like llama.cpp.
This allows to define an environment variable to disable streaming:

    OPENAI_DISABLE_STREAMING

This uses strconv.ParseBool so 1, t, T, TRUE, true, True will set disable the streaming while 0, f, F, FALSE, false, False will keep it.

By default streaming is not disabled.

Streaming will be disabled only if tools are used. If a query doesn't define any tools, streaming will be used.

Fixes #9969

### DIFF
--- a/core/llm.go
+++ b/core/llm.go
@@ -8,19 +8,21 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 
 	"dagger.io/dagger/telemetry"
 	"github.com/anthropics/anthropic-sdk-go"
-	"github.com/dagger/dagger/dagql"
-	"github.com/dagger/dagger/engine"
-	"github.com/dagger/dagger/engine/client/secretprovider"
 	"github.com/iancoleman/strcase"
 	"github.com/joho/godotenv"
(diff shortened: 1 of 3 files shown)


TomChv (approved):

TomChv:
Please fix conflict and feel free to merge it :)

kpenfound:
Nice I haven't had a chance to dig in yet but had a question about this

> Streaming will be disabled only if tools are used. If a query doesn't define any tools, streaming will be used.

This is only while the option is set to true right? I know the original PR this case was always true

I ask because the current state of streaming + tool calling works for almost all OpenAI compat providers

Issue #9405, opened 2025-01-17
🐞 ModuleSource fails with `would clobber existing tag`


vmaffet-fast:
### What is the issue?

We have recently started seeing issues with private modules

When moving tags get updated on the repo containing the dagger modules, the first few dagger calls will error on loading the dependencies. 
The error looks like this:
```
From ssh://github.com/<owner>/<repo>
 * branch            bf8c67bef729d2754633367b112060ef39adfcea -> FETCH_HEAD
 ! [rejected]        actions-v1 -> actions-v1  (would clobber existing tag)
 ! [rejected]        modules-v2 -> modules-v2  (would clobber existing tag)
```

It seems like it fetches the ref successfully (`bf8c67be` here) but unrelated tags update fail the pipeline.

I will be looking into moving the tags to branches, however it seemed strange that the dagger call would fail when these tags are not used in my dagger.json. 

### Dagger version

dagger v0.15.1 (registry.dagger.io/engine:v0.15.1) darwin/arm64

### Steps to reproduce

Not sure on the exact sequence of events, but moving tags such as `actions-v1` get updated, then doing `dagger call` on a module with dependencies to that repository fails with the `would clobber existing tag` message.

Here is an example dagger.json of a module that was failing:
```json
{
  "name": "dagger",
  "engineVersion": "v0.15.1",
  "sdk": "go",
  "dependencies": [
    {
      "name": "conftest",
      "source": "ssh://github.com/<owner>/<repo>/modules/conftest@vincent/conftest",
      "pin": "bf8c67bef729d2754633367b112060ef39adfcea"
    },
    {
      "name": "github",
      "source": "ssh://github.com/<owner>/<repo>/modules/github@main",
      "pin": "c1bfdd036bc965688f02f6c0c6fa287be88dc79b"
    },
    {
      "name": "helm",
      "source": "ssh://github.com/<owner>/<repo>/modules/helm@vincent/helm",
      "pin": "d17aa3014e06587309852fac1f62ed717575707e"
    },
    {
      "name": "registry",
      "source": "ssh://github.com/<owner>/<repo>/modules/registry@main",
      "pin": "c1bfdd036bc965688f02f6c0c6fa287be88dc79b"
    }
  ],
  "source": "."
}
```

### Log output

```
1   : dagger -v call -m .dagger all --auth-token env:TOKEN
2   : │ connect
3   : │ │ starting engine
3   : │ │ starting engine DONE [0.0s]
4   : │ │ connecting to engine
[...]

Issue #9328, opened 2025-01-07
Cloned git branches are replaced with tags


jedevc:
When cloning a git repository, the branch is accidentally confused with a tag, and we get a false tag hovering around - we should not have this.

```
$ dagger -s core git --url https://github.com/dagger/dagger.git branch --name=main tree entries --path .git/refs/tags
main
```

See https://github.com/moby/buildkit/issues/4557 for the upstream buildkit issue.

jedevc:
Fixed by https://github.com/dagger/dagger/pull/9980.

PR #9980, opened 2025-03-26
feat: lift git logic out of buildkit into dagop

jedevc:
Fixes:
- https://github.com/dagger/dagger/issues/9328
- https://github.com/dagger/dagger/issues/9405
- https://github.com/dagger/dagger/issues/7637
- ...and a host of other inconsistencies.

This PR lifts the implementation for `dag.Git` *entirely* out of buildkit, and into dagq; (using DagOp #9395). While in the area, I also modernized and refactored a lot of the cruft out - there's a lot of logic that handles ancient backwards compatibility case that we really just don't need to worry about. Potentially I've missed something that is actually required, but I'd prefer to accidentally remove it (and add it back later), than to keep holding on to legacy code that we don't understand the purpose of.

Finally, we update the API with a couple small low-hanging features:
- `GitRef.ref` to get the *fully* resolved ref, like `refs/tags/xyz`. This is useful so you can actually see what you're cloning - and we needed to use it internally, so it was easy to expose.
- `GitRef.tree` now takes a `depth` arg, which is very easy to add since we now own the entire backend.

I've also forked `gitutil` out of buildkit, we needed *quite* a few changes for this to work nicely in dagger. Some of these maybe could still be upstreamed at some point.


core/git.go
```diff
@@ -141,111 +132,329 @@ func (repo *RemoteGitRepository) PBDefinitions(ctx context.Context) ([]*pb.Defin
...
+			cleanups.Run()
+		}
+	}()
+
+	if repo.AuthToken.Self != nil {
+		username := "x-access-token"
+		if repo.URL.Host == "bitbucket.org" {
+			// NOTE: bitbucket.org is picky, and needs *this* username
+			username = "x-token-auth"
+		}
```
jedevc, line 219: cc @grouville, this is a simplification of part of #8778 - it turns out, we don't need the whole credential helper here?

Anyways, just wanted to check that this bit seemed good to you, all our tests seem to pass correctly :tada:

core/git.go
```diff
@@ -254,92 +463,240 @@ func (ref *RemoteGitRef) PBDefinitions(ctx context.Context) ([]*pb.Definition, e
...
-		opts = append(opts, llb.AuthHeaderSecret(ref.Repo.AuthHeader.ID().Digest().String()))
+
+	if err := os.Remove(filepath.Join(checkoutDirGit, "FETCH_HEAD")); err != nil && !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("failed to remove FETCH_HEAD for remote %s: %w", remote, err)
 	}
 
-	clientMetadata, err := engine.ClientMetadataFromContext(ctx)
+	// TODO: this feels completely out-of-sync from how we do the rest
+	// of the clone - caching will not be as great here
+	_, err = checkoutGit.Run(ctx, "submodule", "update", "--init", "--recursive", "--depth=1")
```
jedevc, line 649: Note to the reviewer: the `TODO`s in this PR are mostly intentional, they don't need to be done here, since they're not a regression from the *existing* behavior.

However, just marking them for future improvement.

core/schema/git.go
```diff
@@ -383,48 +388,80 @@ func (s *gitSchema) gitLegacy(ctx context.Context, parent dagql.Instance[*core.Q
...
+			// and the top-level discard git dir setting
+			strconv.FormatBool(inst.Self.Repo.DiscardGitDir),
+
+			// also include what auth methods are used, currently we can't
+			// handle a cache hit where the result has a different auth
+			// method than the caller used (i.e. a git repo is pulled w/
+			// a token but hits cache for a dir where a ssh sock was used)
+			strconv.FormatBool(ref.Repo.AuthToken.Self != nil),
+			strconv.FormatBool(ref.Repo.AuthHeader.Self != nil),
+			strconv.FormatBool(ref.Repo.SSHAuthSocket != nil),
```
jedevc, line 421: @sipsma so I think this is right.

Instead of having this be on `tree`, I've moved it up to the `ref` selection - so it'll apply to every operation on a `GitRef`.

The reasoning being, we should have a content hash *here*, which means we should easily get the following behavior:
- Cache the results of `dag().git().ref()` per session (which effectively means we only resolve each remote + ref combo once per session, similar to buildkit today).
- And then when it's not cached between sessions, get a "content hashed" call, that includes the ref + commit + remote + etc. So we don't need to invoke the actual op multiple times when this is shared.

But would really love a bit of a closer inspection, just to make sure I'm definitely thinking about this right.

dagql/cachekey.go
```diff
@@ -46,6 +46,36 @@ func CachePerClientObject[A any](
 	return &cacheCfg, nil
 }
 
+// CachePerSession is a CacheKeyFunc that scopes the cache key to the session by mixing in the session ID to the original digest of the operation.
+// It should be used when the operation should be run for each session, but not more than once for a given session.
+func CachePerSession[P Typed, A any](
```
jedevc, line 51: Just thought this would be worth having? It wasn't useful until #9682, but with this, there should be a way to scope cache to just a session (so if two parallel sessions are running, that we won't share results).

Then this'll become more useful when we have *true* global function caching.

util/gitutil/cli.go
jedevc, line 1: Decided to start carving out a `util` directory - happy to reconsider and put it somewhere else, but IMO, it's just not clear what component should own this.

core/git.go
```diff
@@ -254,92 +463,240 @@ func (ref *RemoteGitRef) PBDefinitions(ctx context.Context) ([]*pb.Definition, e
...
+		if doFetch {
+			svcs, err := ref.Repo.Query.Services(ctx)
+			if err != nil {
+				return fmt.Errorf("failed to get services: %w", err)
+			}
+			detach, _, err := svcs.StartBindings(ctx, ref.Repo.Services)
+			if err != nil {
+				return err
+			}
+			defer detach()
```
jedevc, line 524: Heh, pretty happy with this - cc @vito 

This lets us *only* spin up services when we're doing the operations that need them - if we don't do a `git fetch`, then we never need to access them.

core/integration/module_call_test.go
```diff
@@ -390,14 +390,14 @@ func (m *Test) FnRef(ref *dagger.GitRef) *dagger.Directory {
 
 			out, err := modGen.With(daggerCall("fn-repo", "--repo", ".git", "file", "--path=.git/HEAD", "contents")).Stdout(ctx)
 			require.NoError(t, err)
-			require.Equal(t, mainSha, out)
```
vito, line 393: Does this mean the repo is no longer in a `detached HEAD` state?

If so, will that break anything? :thinking: 

`detached HEAD` was a slightly troublesome concept back in my Concourse days with the `git` resource - I felt it was more precise to leave it sitting on an exact SHA, but some folks (or maybe some tools (no disrespect intended)) found it confusing. So curious if there's a particular motivation here.

core/schema/git.go
```diff
@@ -476,87 +513,65 @@ func (s *gitSchema) withAuthHeader(ctx context.Context, parent *core.GitReposito
...
+	dir, err := parent.Self.Tree(ctx, s.srv, args.DiscardGitDir, args.Depth)
+	if err != nil {
+		return inst, err
+	}
+	return dagql.NewInstanceForCurrentID(ctx, s.srv, parent, dir)
 }
 
 type treeArgsLegacy struct {
 	DiscardGitDir bool `default:"false"`
+	Depth         int  `default:"1"`
```
vito, line 567: (I may be hallucinating, but...) I remember Buildkit did some fun internal deduping, where multiple Git refs backed by the same repo would actually re-use the same checkout. Do we still do that?

Semi confusingly attached this comment to a `Depth` line because it jogged the memory - if we always do `depth=1` it might end up less efficient than incrementally fetching a repo with a common history.

vito (commented):
:cool: - just some questions!

core/integration/module_call_test.go
jedevc: It now depends what you ask for.
- So if you ask for a branch, then it puts `HEAD` at that branch.
- Otherwise, `HEAD` is detached.
This actually matches the behavior of `git-checkout` - which is definitely what I expect (I'd actually considered the current behavior a bug). It's also what github actions does.

I do think it is confusing to have the detached head state, when that's not the experience matched by `git-checkout` - see a similar buildkit issue: https://github.com/moby/buildkit/issues/4446.

I've wanted to "fix" this for a while, for example, there's tooling (like goreleaser) that kind of expects that the right tag/branch is present when trying to determine versioning info.

> If so, will that break anything? 🤔

I'm reasonably sure not :eyes: But obviously, worth changelogging that some stuff around this is changing, if we end up doing that.

core/schema/git.go
jedevc: Yes! So, there are always two repos actually:
- The shared store (keyed with a snapshot label that equals the dagql ID)
- The specific checkout to this tree operation

We depth option only controls the depth of this tree operation - the shared store just clones everything out of the box, no shallow shenanigans.

> it might end up less efficient than incrementally fetching a repo with a common history

This is entirely true. Buildkit always clones with a `--depth=1` where possible, and this is now skipped - we just pull everything.

My main reason for doing this is that it's annoying confusing to juggle all of the "this repo might be shallow / it might not be", recovery, etc. But I think it should be easy to add back (without the weird `for` recovery loop that buildkit does for this).

jedevc: > where multiple Git refs backed by the same repo would actually re-use the same checkout

Just to be *super* clear - there *was* an incorrect behavior where asking for a remote with different refs that pointed at the same commit would get confused (see https://github.com/dagger/dagger/issues/9328).

This meant you could potentially sometimes end up in a really weird case where the ref you asked for wasn't actually anywhere in the `.git` directory at all. To resolve this, we don't reuse checkouts unless the fully-resolved ref is also the same.

core/integration/module_call_test.go
vito: > I do think it is confusing to have the detached head state, when that's not the experience matched by `git-checkout` - see a similar buildkit issue: [moby/buildkit#4446](https://github.com/moby/buildkit/issues/4446) (where it actually breaks a gradle build).

Well - if I `git checkout <sha>` it does, which is how I think of CI, since I don't want CI testing against (and caching results of) moving refs. But that's because I've internalized that "I give it a ref, it resolves that to a sha, and goes from there" which is different from just running `git checkout main` => run tests.

Either way I'm sympathetic to the idea of maintaining compatibility with tools out of the box, just as long as all these caching mechanisms behave appropriately.

core/schema/git.go
jedevc: Updated so that we're keeping the `--shallow` clone behavior from buildkit (but made it hopefully more readable :pray:)

dagql/cachekey.go
sipsma: Added this in https://github.com/dagger/dagger/pull/10184 too (just in case you're confused while rebasing)

core/schema/git.go
```diff
@@ -476,87 +502,96 @@ func (s *gitSchema) withAuthHeader(ctx context.Context, parent *core.GitReposito
...
+		// use the commit + fully-qualified ref for the base of the content hash
+		var dgstInputs []string
+		var commit string
+		err = s.srv.Select(ctx, parent, &commit, dagql.Selector{Field: "commit"})
+		if err != nil {
+			return inst, err
+		}
+		dgstInputs = append(dgstInputs, commit)
+		var fullref string
+		err = s.srv.Select(ctx, parent, &fullref, dagql.Selector{Field: "ref"})
```
sipsma, line 533: why do we need both the commit and the ref mixed in here? Would have expected the commit is all we want

core/git.go
```diff
@@ -409,15 +809,25 @@ func (ref *LocalGitRef) PBDefinitions(ctx context.Context) ([]*pb.Definition, er
...
+	commit, fullref, err := ref.Resolve(ctx)
+	if err != nil {
+		return nil, err
+	}
+	cacheKey := dagql.CurrentID(ctx).Digest().Encoded()
+
+	locker := ref.Query.Locker()
+	locker.Lock(gitSnapshotIndex + cacheKey)
+	defer locker.Unlock(gitSnapshotIndex + cacheKey)
+	bkref, err := op.Cache().New(ctx, nil, op.Group(),
```
sipsma, line 827: Why do we have this lock but then unconditionally always create+return a new ref? I would have expected the lock to be useful when it's "get-or-init"-type behavior

core/git.go
```diff
@@ -462,31 +881,253 @@ func (ref *LocalGitRef) Tree(ctx context.Context, srv *dagql.Server, discardGitD
...
+
+	return tempFile.Name(), nil
+}
+
+func runWithStandardUmaskAndNetOverride(ctx context.Context, cmd *exec.Cmd, hosts, resolv string) error {
+	errCh := make(chan error)
+
+	go func() {
+		defer close(errCh)
+		runtime.LockOSThread()
```
sipsma, line 970: I guess this is why our git logic was so good at triggering that go 1.24 bug...

Someday it would be nice to try restoring the mount namespace and unlock here since burning an OS thread every single time git is invoked is kinda crazy, but obviously pre-existing behavior so not high priority.

sipsma (commented):
Looks great overall! Couple questions but nothing major

core/schema/git.go
jedevc: Uh yeah, I moved this back :thinking: I'm now slightly struggling to remember *exactly* why.

I think it was related to accidentally edge merging *way* earlier than we should be :thinking: but yes, ended up deciding to lift it up for now, just to entirely avoid that problem.

core/schema/git.go
jedevc: Yeah, this is super intentional :tada:

Before, if I checked out a tag + branch that pointed to the exact same commit, I would get a racy git state - either the tag or the branch would be around. It was entirely possible to just be missing one of them, but not reliably (kind of a pain if you ran exec `git show <ref>`, which you'd expect to just work IMO).

Because this PR mimics the behavior of `git checkout <ref>`, we're actually on the checked-out branches now, so this behavior becomes even more important.

core/git.go
jedevc: Reasonable point. There's no need for a lock here :tada:

core/schema/git.go
```diff
@@ -55,24 +55,24 @@ func (s *gitSchema) Install() {
 	}.Install(s.srv)
 
 	dagql.Fields[*core.GitRepository]{
-		dagql.Func("head", s.head).
+		dagql.NodeFuncWithCacheKey("head", s.head, dagql.CachePerSession).
```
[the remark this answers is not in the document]
jedevc: Yes - the idea is that per-session, we should have a consistent view of each git repository, but outside of that session, we shouldn't be preserving anything.

Part of the reason for a consistent view is it's "just nice", but the main reason is that there's a cost to doing so - getting the fully-resolved ref and the commit for a repo requires querying the git remote, so we want to ideally do that only once per session.

core/git.go
jedevc: Mmmmm, I *seem* to remember that there's a reason for not doing this - something to do with it being very hard to *safely* call `UnlockOSThread`.

Maybe https://github.com/golang/go/issues/20458? Struggling to remember the exact reason, it could just be cargo-culted.

sipsma (approved):

core/git.go
sipsma: Yeah you have to be super careful because of issues like the one you linked, but afaict since this code only calls the stdlib `os/exec` stuff that's safe here. It is possible though, we do it for our netns stuff during execs for example.

Either way like I said not a big deal for now; the only world in which I can imagine this being meaningful is if the go runtime/scheduler doesn't optimally handle lots of threads getting killed off, which feels plausible but is pure speculation on my part.

PR #9992, opened 2025-03-27
serve dependences for shell prompt mode

cwlbraa:
this lets prompt users `dagger install` modules on the CLI and then manipulate them in prompt mode as though the LLM was acting "inside" the `dagger -m` toplevel module.

there's a alternate approach where we do a bit more and call
```go
	def, err := h.getOrInitDef(cfg.Digest, func() (*moduleDef, error) {
		return initializeModule(ctx, h.dag, cfg.Source)
	})
```
for each dep to initialize it, but at least in my naive implementation that didn't seem to add any additional functionality. @TomChv speculated that that route might allow us to wire in shell completions. Not sure which way is more correct, thoughts @helderco?

i want to automated-test this, but it's proving difficult because of conflicting requirements between needing a TTY to manipulate prompt mode, but also needing historyJSON from the prompt LLM to save a recording. this should be mergable as-is, @marcosnils and I have both manually tested it thoroughly.

### DIFF
--- a/cmd/dagger/shell_fs.go
+++ b/cmd/dagger/shell_fs.go
@@ -261,6 +261,7 @@ func (h *shellCallHandler) maybeLoadModule(ctx context.Context, path string) (*m
 	return def, cfg, err
 }
 
+// this exposes dependency modules to the LLM while loading the primary module
 func (h *shellCallHandler) maybeLoadModuleAndDeps(ctx context.Context, path string) (*moduleDef, *configuredModule, error) {
 	def, cfg, err := h.maybeLoadModule(ctx, path)
 	if err != nil {
@@ -268,13 +269,20 @@ func (h *shellCallHandler) maybeLoadModuleAndDeps(ctx context.Context, path stri
 	}
 
 	if def != nil {
+		depsCtx, depsSpan := Tracer().Start(ctx, "serving dependency modules", telemetry.Encapsulate())
+		var eg errgroup.Group
 		for _, dep := range def.Dependencies {
-			serveCtx, serveSpan := Tracer().Start(ctx, "initializing dependency module")
-			err := dep.Source.AsModule().Serve(serveCtx)
-			telemetry.End(serveSpan, func() error { return err })
-			if err != nil {
-				return nil, nil, fmt.Errorf("failed to serve dependency module: %w", err)
-			}
+			eg.Go(func() error {
+				serveCtx, serveSpan := Tracer().Start(depsCtx, fmt.Sprintf("serving dependency module %s", dep.Name))
+				err := dep.Source.AsModule().Serve(serveCtx)
+				telemetry.End(serveSpan, func() error { return err })
+				return err
+			})
+		}
+		err := eg.Wait()
+		telemetry.End(depsSpan, func() error { return err })
+		if err != nil {
+			return nil, nil, err
 		}
 	}
 
--- a/cmd/dagger/shell_fs.go
+++ b/cmd/dagger/shell_fs.go
@@ -59,7 +59,7 @@ func (h *shellCallHandler) ChangeDir(ctx context.Context, path string) error {
 
 	var subpath string
 
-	def, cfg, err := h.maybeLoadModule(ctx, path)
+	def, cfg, err := h.maybeLoadModuleAndDeps(ctx, path)
 	if err != nil {
 		return err
 	}
@@ -257,7 +257,28 @@ func (h *shellCallHandler) maybeLoadModule(ctx context.Context, path string) (*m
 	def, err := h.getOrInitDef(cfg.Digest, func() (*moduleDef, error) {
 		return initializeModule(ctx, h.dag, cfg.Source)
 	})
+
 	return def, cfg, err
+}
+
+func (h *shellCallHandler) maybeLoadModuleAndDeps(ctx context.Context, path string) (*moduleDef, *configuredModule, error) {
+	def, cfg, err := h.maybeLoadModule(ctx, path)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if def != nil {
+		for _, dep := range def.Dependencies {
+			serveCtx, serveSpan := Tracer().Start(ctx, "initializing dependency module")
+			err := dep.Source.AsModule().Serve(serveCtx)
+			telemetry.End(serveSpan, func() error { return err })
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to serve dependency module: %w", err)
+			}
+		}
+	}
+
+	return def, cfg, nil
 }
 
 // parseModRef transforms user input into a full module reference that can be used with
--- a/cmd/dagger/shell.go
+++ b/cmd/dagger/shell.go
@@ -190,7 +190,7 @@ func (h *shellCallHandler) Initialize(ctx context.Context) error {
 	var cfg *configuredModule
 
 	if !shellNoLoadModule {
-		def, cfg, err = h.maybeLoadModule(ctx, ref)
+		def, cfg, err = h.maybeLoadModuleAndDeps(ctx, ref)
 		if err != nil {
 			return err
 		}


cmd/dagger/shell_fs.go
```diff
@@ -257,9 +257,28 @@ func (h *shellCallHandler) maybeLoadModule(ctx context.Context, path string) (*m
...
+	}
+
+	for _, dep := range def.Dependencies {
+		serveCtx, serveSpan := Tracer().Start(ctx, "initializing dependency module", telemetry.Encapsulate())
+		err = dep.Source.AsModule().Serve(serveCtx)
+		telemetry.End(serveSpan, func() error { return err })
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to serve dependency module: %w", err)
+		}
+	}
```
helderco, line 277: I don't think this is necessary, at least not in shell mode. It'll make things slower. Not quite sure what you're trying to do.

In shell, you can reference a dependency by name and it will load on demand. Same way you can use a path or URL and it'll load automatically too.

cwlbraa: > In shell, you can reference a dependency by name and it will load on demand. 

> this lets prompt users dagger install modules and manipulate them in prompt mode as though they were "inside" the core module.

we don't want to reference every dependency by name for prompt mode, the idea is that we want folks to be able to install dependencies and have the prompt/shell act as though they were inside SDK code. calling "serve" here triggers the installation machinery engine-side that makes the dependency modules available to the LLM.

grouville: Hey 👋  I am a bit confused here too 👀 

In a sense, it seems similar / overlapping (in theory) to what we're doing in https://github.com/dagger/dagger/pull/9984 ; but, basically at `runtime` ?

From above comments, I see 2 things:
1. At runtime, I can install a module and its dependencies
2. These dependencies appear top-level ; they are not scoped anymore to the _dynamic_ loading of the module (done by name)

The current state of your PR seems to indicate this UX:
- I load a shell with `-M` (I `dagger -M`), so no module
- Inside prompt mode you install `github.com/shykes/hello`
- All the functions of hello appear in my current environment, no need to do `hello` to access `hello`'s functions

grouville: Btw, we got some integrations tests that might interest you: https://github.com/grouville/dagger/commit/0aaf23060984fb60d44eb32bb36cd3d7931d5d82

cwlbraa: > The current state of your PR seems to indicate this UX:
I load a shell with -M (I dagger -M), so no module
Inside prompt mode you install github.com/shykes/hello
All the functions of hello appear in my current environment as top-level functions, no need to do hello to access hello's functions

you can't see it in the diff, but that's the one shell-module-load case (maybeLoadModule callsite) where the modules deps _are not_ installed. the other 2 where the deps are installed are for `dagger -m github.com/shykes/hello` and the plain implicit `dagger -m .` -> in both these cases this PR makes the LLM inside the shell act as though it's inside of the module where the shell started up. 

0 effect on `dagger -M`.

marcosnils: > In shell, you can reference a dependency by name and it will load on demand. Same way you can use a path or URL and it'll load automatically too.

 @helderco in this particular case, the way that the `prompt` mode discovers the tools is by checking what's registered in the main GraphQL API. @cwlbraa 's intention with this PR is to make that registration eagerly so the LLM can interact with the dependecy modules. 

If you think there's any downsides in following this approach, let's chat to find the best way to implement it. Ater checking with Connor, we haven't observed any noticeable side-effects or performance  penalties while testing this :pray:

cwlbraa: > let's chat to find the best way to implement it

yeah @helderco  im super open to alternatives here, poked at several today but everything i tried ended up being complex enough to make me revert to the existing solution

shykes (approved):

shykes:
Thank you @cwlbraa !

I'm approving as-is, I understand there are unresolved discussions on the best way to implement this, but since 1) this feature is load-bearing for the UX and we need to get some testing ASAP, and 2) any improvements can be merged as follow-up without breaking the UX - it seems low-risk to merge.

cmd/dagger/shell_fs.go
helderco: Good call on the errgroup, I was going to mention that. 🙂

> the idea is that we want folks to be able to install dependencies and have the prompt/shell act as though it were inside SDK code

Ok, that clarifies it for me, thanks.

> we haven't observed any noticeable side-effects or performance penalties while testing this

Yeah, now that things are clearer it makes sense why. The biggest performance hit already happens on load. When you load a module, its dependencies are loaded first. Since you're just calling serve, the overhead is in extending the API schema, which should be mostly cached (I think). Again, the most expensive part is loading, which happens anyway, where the runtime for each dependency is built and called in order to register the module's types.

If I load our dagger-dev module which has many dependencies, that overhead can be ≈0.5s (out from a total of ≈15s). Even though it’s unnecessary for the UX of shell alone, I’m ok with that overhead in order to make llm seamless. But, something to consider is that since the overhead is small, it could be on llm side alone too. If people could notice a difference when switching to llm mode, than it should be a noticeable difference for all the users that don’t use llm and get the hit anyway.

> let's chat to find the best way to implement it

I think this should ideally be handled by an API call. Even if a module’s `Serve` had a bool to include dependencies or not. There’s also been a few (experimental) PRs related to codegen calling Serve on deps and it’s all over the place.

This also makes a fundamental shift in what it means to load a module, so I’m wondering about consistency. Where does it make sense to load the dependencies like this or not? And does the difference make sense to fresh eyes?

> the other 2 where the deps are installed are for `dagger -m github.com/shykes/hello` and the plain implicit `dagger -m .`

Not just. The `-m` flag sets an initial context. You can change that context with the (experimental) `.cd` command. When you navigate the filesystem with `.cd` the maybeLoadModule is triggered. This way, if you change to another module’s directory the shell loads it automatically. Thus the “maybe”.

It also means you could be navigating the same module’s subdirs and so this function is going to be called every time, even for the same module. So I would move this code inside the function passed to `getOrInitDef`, after `initializeModule`. This way the dependencies are only served once, after the module itself.

> 0 effect on `dagger -M`.

Not quite. `dagger -M` just makes sure the initial context is the core api, and not look for a module. But you can still do `.cd` in the session. That will load a module if there is one, and serve its dependencies. So it doesn’t prevent loading modules.

This means dependencies of a module can still get loaded and served if you start the session with  `-M`, but that may just be semantics. If the worry was that dependencies would get loaded even when just starting with `-M` it’s correct that they won’t. Not until you `.cd` into a module.

cwlbraa: i typed a response to this 2 weeks ago, didn't send it, and now it's gone :( 

> So I would move this code inside the function passed to getOrInitDef, after initializeModule. This way the dependencies are only served once, after the module itself.

gonna do this right now.

> that may just be semantics.

definitely semantics lol, but re-reading my old comment I'm also confused as to what I was talking about... i definitely meant to say the 2 callsites were `.cd` and `-m`, but now looking at [the third one](https://github.com/cwlbraa/dagger/blob/6231ef1f81b64d4dad5386362534c4f4fb6ae128/cmd/dagger/shell_exec.go?plain=1#L329) inside StateLookup, I'm realizing I don't actually know what it does and I was just assuming it was used for `-M`. I can see it used in `.help` impl and the core `Exec` implementation, but that doesn't make things entirely clear to me.

helderco: Having an initial context with `-c` or using `.cd` aren't the only ways to load a module. You can also use module refs as commands, it just won't become the default:
- local path
- git URL
- dependency name (if current context is a loaded module)

For example:

```
❯ dagger -M
Dagger interactive shell. Type ".help" for more information. Press Ctrl+D to exit.

✔ github.com/shykes/hello | hello
hello, world!

⋈ .help github.com/shykes/hello 
```

So in this case, due to what's in StateLookup, the above won't serve any dependencies.

That's what I meant by consistentency. But it's not just within shell. There's other commands to consider, like:
- run
- session
- listen
- query

Like I said before, rather than calling serve on deps within the CLI, I would rather this be done by the engine, either always on `ModuleSource.serve` or via an option (e.g., `all`, to include dependencies).

As for `getOrInitDef`, it's just for keeping the introspection done at module load in memory (cache), so it doesn't call `initializeModule` multiple times for the same module.

PR #9995, opened 2025-03-28
feat: add libsecret provider

jedevc:
This was scratching an itch :laughing: 

On my laptop, I use GNOME, which provides `gnome-keyring` for storing passwords/keys, etc - it's fairly neat, and integrates with the desktop environment to prompt for the unlock password, etc. I can use it to store encrypted passwords for things like API keys, etc.

To do this, we can integrate with `libsecret`, which allows accessing this info. This is pretty much as close to a standard as there is across Linux desktop environments, it's also supported in KDE by KWallet for example (though I haven't personally tried that).

### DIFF
--- a/engine/client/secretprovider/libsecret.go
+++ b/engine/client/secretprovider/libsecret.go
@@ -1,0 +1,134 @@
+package secretprovider
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"path"
+	"strings"
+
+	"github.com/jedevc/go-libsecret"
+)
+
+// libsecretProvider looks up secrets using libsecret, which connects to
+// gnome-keyring and other similar providers.
+// https://specifications.freedesktop.org/secret-service-spec/latest/ref-dbus-api.html.
+//
+// Format:
+// - libsecret://<collection>/<id>
+// - libsecret://<collection>/<label>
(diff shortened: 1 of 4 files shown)


nipuna-perera (commented):
This is great! Would it be possible to do the same for GNU Pass? I personally use it on my Mac and I know others in my company that do so too.  Would love to have support for that.

jedevc:
Sure! Would definitely love to support that as well, though it doesn't look like `pass` implements the freedesktop.org secret provider that would let it work with `libsecret`.

kpenfound (approved):
Awesome! I don't have an environment to test this. I guess there's no key lookup api?

jedevc:
> I guess there's no key lookup api?

`libsecret` is the key lookup API :cry: It communicates over D-Bus, so that's why those deps get pulled in.

Having looked through the ecosystem, it looks like there's several different ways of querying keys in the keystore:
- By `id` (which is the path to the object in the D-Bus API)
- By `label` (which is the human readable name - this is weird, but isn't hard to support)
- By `key`/`value` pairs (e.g. this is how the `gh` cli stores tokens on Linux)
So I just decided it would be reasonable to support all of these, since there's not really a consistent style (docs on this *suck*).

PR #9996, opened 2025-03-28
add a new recorder module using vhs

sagikazarmark:
Usage:

```shell
dagger -m docs/recorder2 call render export --path /tmp/out --wipe
```

## Todo

- [x] Finish setting up the base image
- [x] Improve caching of individual tapes
- [x] Create separate recording for quickstart and features
- [x] Skip shell initialization from recordings
- [x] Add secrets (once VHS supports them)
- [x] Fix quickstart recordings


jasonmccallister:
@sagikazarmark here is the `.tape` that I have been using will pretty good results:

```
Set Width 1500
Set Height 800
Set Padding 5
Set WindowBar Colorful

Hide
Type "dagger"
Enter
Sleep 1s
Ctrl+L
Show

Type "container | from alpine | with-exec apk add curl | with-exec -- curl -L https://dagger.io | stdout"
Enter
Sleep 1s
Wait+Screen     /div/
Sleep 2s
```

It outputs the following:

feature-interactive-shell

vikram-dagger:
This looks great, thank you @sagikazarmark !

To get this ready for merge, I suggest the following changes:
- Remove the old recorder and replace with this (rename this to `recorder`)
- Create two functions, one for quickstart recordings and one for feature page recordings (the current example for the shell page can go into the latter function).
- Add any one example for the quickstart recordings (since that runs in the context of existing source code)
- Update the STYLE_GUIDE.md file with instructions on how to run the new functions

Some questions:
- Would this work for the second services feature page recording which relies on a host db service?
- Is there a way to "split screen" a recording or show two terminals at once? Not a blocker for merge but for the first services feature page recording which instantiates a web service, it would be cool to show the service running in one terminal and a `curl localhost` in another. Understood in advance this may not be possible.
- Is there a way to inject an env variable (like a GitHub token - needed for the secrets feature page recording)?

In case you already have functions for any/all of the above, feel free to just point me to it :)

sagikazarmark:
@jasonmccallister thanks!

One thing I would add is I noticed that sleep isn't enough in many of the cases (even when just starting shell), so I used wait whenever possible.

@vikram-dagger

> To get this ready for merge, I suggest the following changes:

I mostly intended this to start the conversation, I don't think it's fully ready yet, but I'm glad you like it!

> Remove the old recorder and replace with this (rename this to `recorder`)

I would probably leave it as is for now. If it gets merged, we can convert the remaining "tapes" in subsequent PRs. I found using the Dagger shell in VHS to be a bit of a pain due to how the prompt is presented, so it takes some time to convert those to tapes.

 
>  Would this work for the second services feature page recording which relies on a host db service?

We probably have to do some hacking to put that service into the background, but we can probably make it work.

> Is there a way to "split screen" a recording or show two terminals at once? Not a blocker for merge but for the first services feature page recording which instantiates a web service, it would be cool to show the service running in one terminal and a `curl localhost` in another. Understood in advance this may not be possible.

I don't think there is a way to do split screen, but maybe we can do two recordings and put them side by side. We need to time it right, so it may not be worth the effort, but we can try.
 
> Is there a way to inject an env variable (like a GitHub token - needed for the secrets feature page recording)?

Yes, `vsh` forwards env vars to the shell, so anything set on the container should appear as env var in the scripts.

For now, it's possible to set those in the container, but I'll try to come up with a way to add them directly to the module API.

jasonmccallister:
@sagikazarmark yeah, the ending bit here:

```
Wait+Screen     /div/
Sleep 2s
```

Waits for the div output on the screen and then sleeps for 2s so the user can get a pause and see what happens - without it the recording feels too fast.

Also these options are up for discussion/removal/changing:

```
Set Padding 5
Set WindowBar Colorful
```

As well as the `Set Height 800` was specific to this example as we needed more content.

vikram-dagger (approved):

PR #9997, opened 2025-03-28
feat(engine/secret): add support for ttl for vault

rajatjindal:
This PR adds the support for ttl in secrets provider. Currently it is only supported for vault (as we are caching only vault secrets at the moment).

Also, I noticed we support only kv based secrets in vault, and we can possibly extend the secrets provider to also support dynamic cred engines such as `aws`, `postgres` etc, but that is out of scope for this work item, so just putting it out here for the record.

This PR still needs a design review/approval from the team. cc @marcosnils

### DIFF
--- a/core/integration/secret_test.go
+++ b/core/integration/secret_test.go
@@ -4,9 +4,11 @@ import (
 	"bytes"
 	"context"
 	_ "embed"
+	"fmt"
 	"io"
 	"testing"
 
+	"dagger.io/dagger"
 	"github.com/stretchr/testify/require"
 
 	"github.com/dagger/dagger/dagql/call"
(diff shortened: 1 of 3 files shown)


engine/client/secretprovider/secretprovider.go
```diff
@@ -4,34 +4,67 @@ import (
...
 	resolver, ok := resolvers[scheme]
 	if !ok {
-		return nil, "", fmt.Errorf("unsupported secret provider: %q", scheme)
+		return nil, fmt.Errorf("unsupported secret provider: %q", scheme)
+	}
+
+	return resolver, nil
+}
+
+func ResolverForID(id string) (SecretResolver, string, error) {
```
jedevc, line 45: Okay, I'd actually prefer a different approach here, lemme know what you think.

There's nothing specific about vault here, and I think we should refactor to lift the caching behavior *out* of vault, so we're not just special-casing it, which makes it harder to add more later.

I think that we should define a `cachedSecretProvider` - it takes a secret provider as an arg, and returns another secret provider. Then the logic for caching can be defined there *not* specific to vault, which will make it much easier to extend in the future. Then, we just have:

```go
var resolvers = map[string]SecretResolver{
	// ...
	"vault":   cachedSecretProvider(vaultProvider),
```


As part of that, I think we'd probably want to change the signature of the `secretprovider` to take a `*url.URL` instead of a `string` directly? That makes it easier to get the value of the `ttl` query param. I was thinking of doing this already in https://github.com/dagger/dagger/pull/9995, given we're parsing it as a URL anyways, might as well just pass it in as such to each secret provider.

rajatjindal: There is one subtle difference between the secret providers. e.g. for `file` (and other providers), we just read the file and returns its content as is, while for `vault`, we get the secret (which is an object), cache the whole object and return the specific key from that object as requested by the user (as part of the `uri`)

so if we cache it based on `uri`, we will end up making duplicate calls to vault for `vault://secret/some-object.foo` and `vault://secret/some-object.bar`. This will specifically be problematic if we want to get credentials for dynamic vault engine e.g. database or aws, as we will generate two set of creds in those two calls.

so I am thinking we will have to implement provider specific caching logic here anyways. what do you think?

what we can do is to define an interface, which has one `func GetSecret(context.Context, uri string)`, and for vault it returns result from cached data (while keeping a few parts of retrieving from cache, storing in cache generic). But that key has to calculated based on the provider.

what do you think?

rajatjindal: Hi @jedevc, I have done an alternate implementation where I am using a CacheProvider interface which currently supports two type of caching. one is object based (for vault and similar), and one where we return the complete value (not used currently, but we can configure file/env etc provider to use this cache provider):

https://github.com/dagger/dagger/compare/main...rajatjindal:dagger:allow-ttl-secret-2?expand=1

please let me know your thoughts.

rajatjindal:
Hi @jedevc a kind reminder for feedback on this PR.

engine/client/secretprovider/secretprovider.go
jedevc: I think that's probably a bit overly complex.

I think the approach in this PR is good enough as is. I hadn't realize the differences in caching of how the vault keys worked.

engine/client/secretprovider/secretprovider.go
```diff
@@ -4,34 +4,67 @@ import (
...
-	"vault": vaultProvider,
+	"vault": getVaultProvider(0), // no ttl for backward compatibility
 }
 
-func ResolverForID(id string) (SecretResolver, string, error) {
-	scheme, path, ok := strings.Cut(id, "://")
-	if !ok {
-		return nil, "", fmt.Errorf("parse %q: malformed id", id)
+func getSecretResolverWithTTL(scheme string, ttl time.Duration) (SecretResolver, error) {
+	if scheme == "vault" && ttl > 0 {
```
jedevc, line 33: We shouldn't have to special case `vault` in `secretprovider.go`. All the vault logic should go in `vault.go`.

engine/client/secretprovider/secretprovider.go
```diff
@@ -4,34 +4,67 @@ import (
...
+	path := parsed.Path
+
+	var ttl time.Duration
+	ttlStr := strings.TrimSpace(parsed.Query().Get("ttl"))
+	if ttlStr != "" {
+		ttl, err = time.ParseDuration(ttlStr)
+		if err != nil {
+			return nil, "", fmt.Errorf("invalid ttl %q provided for secret %q. err: %w", ttlStr, id, err)
+		}
 	}
```
jedevc, line 61: All this logic *should* be moved into `vault.go`, even though we can't make the caching concept generic.

core/integration/secret_test.go
```diff
@@ -186,3 +188,113 @@ func (SecretSuite) TestBigScrubbed(ctx context.Context, t *testctx.T) {
...
+			vault := c.Container().
+				From("hashicorp/vault").
+				WithEnvVariable("VAULT_DEV_ROOT_TOKEN_ID", "vault-root-token").
+				WithEnvVariable("VAULT_LOG_LEVEL", "debug").
+				WithExposedPort(8200).
+				AsService(dagger.ContainerAsServiceOpts{
+					UseEntrypoint:                 true,
+					ExperimentalPrivilegedNesting: true,
+					InsecureRootCapabilities:      true,
+				})
```
marcosnils, line 293: why do we create a vault service here  if we're not accessing it outside the module? wouldn't it be better if we just keep this withing the module test?

marcosnils (commented):
left a nit comment. besides that, LGTM based on the previous discussions in this thread :pray:

core/integration/secret_test.go
rajatjindal: to test the secret fetching functionality, we need to simulate a vault service running on host system.

jedevc:
@rajatjindal is this ready for review?

engine/client/secretprovider/vault.go
```diff
@@ -3,20 +3,44 @@ package secretprovider
...
+
+	// this is just path part without the query params such as ttl
+	key := parsed.Path
+
+	var ttl time.Duration
+	ttlStr := strings.TrimSpace(parsed.Query().Get("ttl"))
+	if ttlStr != "" {
+		ttl, err = time.ParseDuration(ttlStr)
+		if err != nil {
+			return nil, fmt.Errorf("invalid ttl %q provided for secret %q. err: %w", ttlStr, key, err)
```
jedevc, line 40: The standard way of doing this is:

```suggestion
			return nil, fmt.Errorf("invalid ttl %q provided for secret %q: %w", ttlStr, key, err)
```

engine/client/secretprovider/vault.go
```diff
@@ -31,26 +55,46 @@ func vaultProvider(ctx context.Context, key string) ([]byte, error) {
...
 		}
-	}
 
-	// check if path is in key cache
-	if _, ok := vaultCache[key]; !ok {
 		// read the secret
 		s, err := vaultClient.KVv2(mount).Get(ctx, secretPath)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("path: %q: %w", secretPath, err)
```
jedevc, line 70: With error strings, generally we only use `:` to separate each error from the next in the chain, so it's fine to do:
```suggestion
			return nil, fmt.Errorf("path %q: %w", secretPath, err)
```

engine/client/secretprovider/vault.go
```diff
@@ -3,20 +3,44 @@ package secretprovider
...
 
+type dataWithTTL struct {
+	expiresAt time.Time
+	data      map[string]any
+}
+
 var (
 	vaultClient *vault.Client
-	vaultCache  = make(map[string]map[string]any)
+	vaultCache  = make(map[string]dataWithTTL)
```
jedevc, line 22: :thinking: just realized, I think this probably needs a lock right?

I don't think there's anything to stop parallel secret access? So we should make sure to lock when accessing this cache.

core/integration/secret_test.go
```diff
@@ -186,3 +189,105 @@ func (SecretSuite) TestBigScrubbed(ctx context.Context, t *testctx.T) {
 	require.NoError(t, err)
 	require.Equal(t, "***", stdout)
 }
+
+func (SecretSuite) TestVaultSecretsProviderTTL(ctx context.Context, t *testctx.T) {
```
marcosnils, line 193: are there any chances of: 

- Make the test faster since it has to sleep at least 20s per case iteration
- Make the test more robust since it seems to be dependent on how much time things take to run?

engine/client/secretprovider/vault.go
marcosnils: Yep, seems to be the case since we're only synchronizing readers here: https://github.com/dagger/dagger/blob/31c82dcd8871ae0ced55f9e3b77eed207098c169/core/secret.go?plain=1#L215 which is where the provider gets called.

core/integration/secret_test.go
rajatjindal: refactored the test, and found the sweet spot where it works.

core/services.go
```diff
@@ -211,7 +211,7 @@ func (ss *Services) StartBindings(ctx context.Context, bindings ServiceBindings)
 	detach := func() {
 		detachOnce.Do(func() {
 			go func() {
-				<-time.After(DetachGracePeriod)
+				<-time.After(120 * time.Second)
```
marcosnils, line 214: @rajatjindal seems like a miss?

marcosnils, line 214: @rajatjindal seems like a miss?

core/integration/secret_test.go
```diff
@@ -183,3 +185,116 @@ func (SecretSuite) TestBigScrubbed(ctx context.Context, t *testctx.T) {
...
+	"context"
+	"dagger/foo/internal/dagger"
+	"fmt"
+	"time"
+)
+
+type Foo struct{}	
+
+// This function sets the value of secret in vault, gets its plaintext value. Then
+// this function updates the value of secret in vault (to simulate expired or changed secret), and sleeps for 20s to allow for ttl (if any)
```
marcosnils, line 215:
```suggestion
// this function updates the value of secret in vault (to simulate expired or changed secret), and sleeps for 5s to allow for ttl (if any) 
```

core/services.go
rajatjindal: yep, indeed. fixed 👍

marcosnils (approved):

marcosnils:
SGTM! would love one more :eyes:  from @jedevc  before merging

jedevc (approved):

jedevc:
If we could follow-up with some docs at https://docs.dagger.io/api/arguments#external-secret-providers, that would be great :tada:

marcosnils:
> If we could follow-up with some docs at https://docs.dagger.io/api/arguments#external-secret-providers, that would be great 🎉

 cc @vikram-dagger

PR #10000, opened 2025-03-28
ci: fix php docs generation and linting

jedevc:
At some point, this seems to have broken:
1. we weren't actually outputting the updated php docs to the right place
2. we weren't linting to see if those docs had actually been updated

And so, they were falling behind, this brings them up to date and fixes the issue.


gerhard (commented):
LGTM, but would like @charjr & @helderco to approve this before merging ⛙

helderco (approved):

PR #10007, opened 2025-03-29
Environment API: a new API for exposing a Dagger environment to human…

shykes:
- LLM is now a regular type. It can receive an environment
- Environment: an execution context within the Dagger Engine. Basically, a sandbox.
- Binding: a named mapping to a value, scoped to an Environment
- Environment and Binding are now dynamic, instead of LLM.
- Environment.with[TYPE]Binding: set a binding from from a value of type [TYPE]
- Binding.as[TYPE]: get the value of a binding, converted to type [TYPE]


jedevc:
cherry-picked https://github.com/dagger/dagger/pull/10000 over (needed for regen-ing docs correctly)

sipsma (approved):

PR #10008, opened 2025-03-29
infra: Add hostNetwork, dnsPolicy and dnsConfig to Helm chart

tonygilkerson:
Added hostNetwork, dnsPolicy and dnsConfig as optional fields to the Helm chart.  

This is needed in some cases when using `kube-pod` to access the dagger engine running in a Kubernetes cluster.  In my case I installed RKE2, deployed the dagger helm chart and got `failed to resolve image "docker.io/library/alpine:latest"` errors.  After digging into it a bit I found that if I set `hostNetwork: true` it started working

```
export _EXPERIMENTAL_DAGGER_RUNNER_HOST="kube-pod://...
...

$ dagger query <<EOF
> {
>   container {
>     from(address:"alpine") {
>     withExec(args: ["uname", "-a"]) { stdout }
>     }
>   }
> }
> EOF
...
Error: make request: input: container.from failed to resolve image "docker.io/library/alpine:latest" (platform: "linux/amd64"): failed to resolve source metadata for docker.io/library/alpine:latest: failed to do request: Head "https://registry-1.docker.io/v2/library/alpine/manifests/latest": dial tcp: lookup registry-1.docker.io on 10.87.0.1:53: read udp 10.87.0.1:47496->10.87.0.1:53: i/o timeout
```

### DIFF
--- a/helm/dagger/templates/engine-daemonset.yaml
+++ b/helm/dagger/templates/engine-daemonset.yaml
@@ -143,6 +143,16 @@ spec:
             {{- with .Values.engine.volumeMounts }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
+      {{- if .Values.engine.hostNetwork }}
+      hostNetwork: {{ .Values.engine.hostNetwork }}
+      {{- end }}
+      {{- if .Values.engine.dnsPolicy }}
+      dnsPolicy: {{ .Values.engine.dnsPolicy }}
+      {{- end }}
+      {{- with .Values.engine.dnsConfig }}
+      dnsConfig:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       terminationGracePeriodSeconds: {{ .Values.engine.terminationGracePeriodSeconds }}
       volumes:
         {{- if .Values.engine.hostPath.dataVolume.enabled }}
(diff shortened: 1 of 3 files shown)


marcosnils:
@tonygilkerson  thx for the patch. I'm wondering if by any chance we might want to provide this capability in a more a  generic way to handle these cases where  potentially any of the deployment properties might be potentially overridden by users? I'm mostly asking because adding `dnsPolicy`, `dnsConfig` and `hostNetwork` as top level attributes in the values.yaml file seems like a code smell to me.

tonygilkerson:
@marcosnils I hear you and I am willing to try and make it more generic but I might need you to give me a better idea of what you are thinking.  I don't see much difference in the values I added and the existing chart values such as `persistentVolumeClaim`.  Are you wanting something generic enough to apply to the other values in the chart?

For example, one way to do what you asked is to use a Helm post render hook to modify the rendered manifests before they are applied to the cluster. This is very generic and would allow one to update any value in the entire Chart but it is a different user experience.   It would require the user to understand how to write kustomization overlays and stuff like that.

marcosnils:
hey @tonygilkerson! sorry I missed this, somehow it got buried in one of my gmail Github labels. 

Back to the issue, I was thinking if we could maybe follow a similar pattern that we follow here: 
https://github.com/dagger/dagger/blob/main/helm/dagger/templates/engine-daemonset.yaml#L143-L145 but at the container level. This way, we could users arbitrarily set new fields in the engine container spec without the need to add a separate value in the template file. 

WDYT?

PR #10010, opened 2025-03-30
shell: `.llm` -> `$agent`, ui polish

vito:
* `.llm` => `$agent`
* get rid of llm variables => shell syncing now that it doesn't set vars
* get rid of obnoxious background highlight in nav mode
* chain calls by the same actor (i.e. tool calls, thinking spans)

### DIFF
--- a/cmd/dagger/llm.go
+++ b/cmd/dagger/llm.go
@@ -10,7 +10,6 @@ import (
 	"os"
 
 	"dagger.io/dagger"
-	"dagger.io/dagger/querybuilder"
 	"dagger.io/dagger/telemetry"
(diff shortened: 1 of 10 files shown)


cmd/dagger/llm.go
```diff
@@ -277,11 +277,11 @@ type dynamicObject struct {
 	typeName string
 }
 
-func (do *dynamicObject) XXX_GraphQLType() string {
+func (do *dynamicObject) XXX_GraphQLType() string { //nolint:stylecheck
```
jedevc, line 280: lol wut :eyes: feels like the linter should have been catching this before

jedevc (approved):

PR #10023, opened 2025-03-31
refactor(modules/golangci): use $HOME instead of path.Join

wingyplus:
Eliminate FIXME in the `golangci` by using `$HOME` together with `expand` option.


jedevc (approved):
Neat :tada:

PR #10025, opened 2025-03-31
docs: Update Shell feature page with variables example

vikram-dagger:


jpadams (approved):
nice!

tested example in system and dagger shell 👍

PR #10029, opened 2025-04-01
Improve releasing during v0.18.0

sipsma:

PR #10030, opened 2025-04-01
ci(alt runners): Bump Dagger Engine version

gerhard:
I thought this was already handled by the post-release process, specifically this command, but it doesn't seem to be the case 🤷

    find .github/ -type f -exec sed -i 's/0-17-1/0-17-2/g; s/0\.17\.1/0\.17\.2/g' {} +

Follow-up to https://github.com/dagger/dagger/pull/9994


gerhard:
Precedes:
- https://github.com/dagger/dagger/pull/10029

If the above gets merged first, then this should be closed.

If this gets merged first, then the above should be merged next.

PR #10031, opened 2025-04-01
Docs v0.18.0 Update llm snippets

kpenfound:


docs/current_docs/api/llm.mdx
```diff
@@ -134,7 +102,7 @@ Dagger [modules](../features/modules.mdx) are collections of Dagger Functions. W
 
 Environments configure a number of inputs and outputs for the LLM. For example, an environment might provide a `Directory`, a `Container`, and a `string` variable. The LLM can use the scalars and the functions of these objects to complete the assigned task.
 
-The documentation for the Objects are provided to the LLM, so make sure to provide helpful documentation in your Dagger Functions. The LLM should be able to figure out how to use the tools on its own. If you provide too much help, it will be redundant with the function descriptions.
+The documentation for the Modules are provided to the LLM, so make sure to provide helpful documentation in your Dagger Functions. The LLM should be able to figure out how to use the tools on its own. Don't worry about describing the objects too much in your prompts because it will be redundant with this automatic documentation.
```
vikram-dagger, line 105:
```suggestion
The documentation for the modules are provided to the LLM, so make sure to provide helpful documentation in your Dagger Functions. The LLM should be able to figure out how to use the tools on its own. Don't worry about describing the objects too much in your prompts because it will be redundant with this automatic documentation.
```

We stopped capitalizing modules

docs/current_docs/quickstart/agent/index.mdx
```diff
@@ -96,17 +96,17 @@ dagger functions
 
 You should see information about two auto-generated Dagger Functions: `container-echo` and `grep-dir`.
 
-## Install the agent environment module
+## Install a dependency with useful tools
 
-An agent needs an environment in which to operate. The agent uses this environment to solve the assignment given to it. This environment is a Dagger module, which you can either create yourself or use from elsewhere.
+An agent needs tools to complete it's tasks. The agent uses this module to solve the assignment given to it. Agents can use any Dagger module as tools, which you can either create yourself or use from elsewhere.
```
vikram-dagger, line 101:
```suggestion
An agent needs tools to complete its tasks. The agent uses this module to solve the assignment given to it. Agents can use any Dagger module as tools, which you can either create yourself or use from elsewhere.
```

docs/current_docs/api/llm.mdx
```diff
@@ -134,7 +102,7 @@ Dagger [modules](../features/modules.mdx) are collections of Dagger Functions. W
 
 Environments configure a number of inputs and outputs for the LLM. For example, an environment might provide a `Directory`, a `Container`, and a `string` variable. The LLM can use the scalars and the functions of these objects to complete the assigned task.
```
vikram-dagger, line 103: This needs an additional explanation: what is the connection between an environment and one or more modules

docs/current_docs/api/llm.mdx
```diff
@@ -163,4 +131,4 @@ Consider the following Dagger Function:
 
 Here, the `ToyWorkspace` is given as an input to the `Env`. It contains a number of Dagger Functions: `Read()`, `Write()`, and `Build()`. When this environment is attached to an `LLM`, the LLM can call any of these Dagger Functions to change the state of the `ToyWorkspace` and complete the assigned task.
```
vikram-dagger, line 132:
```suggestion
Here, the `ToyWorkspace` module is attached as an input to the `Env` environment. The module contains a number of Dagger Functions: `Read()`, `Write()`, and `Build()`. When the environment containing this module is attached to an LLM, the LLM can call any of the Dagger Functions to change the state of the `ToyWorkspace` and complete the assigned task.
```

docs/current_docs/api/llm.mdx
kpenfound: Any suggestions? I'm not sure what I would say beyond what's here. An environment defines inputs and outputs, that's all

docs/current_docs/api/llm.mdx
```diff
@@ -163,4 +131,4 @@ Consider the following Dagger Function:
...
 
-In the `Env`, a `ToyWorkspace` called `after` is specified as a desired output of the LLM. This means that the LLM should return the `ToyWorkspace` object after it has completed its task. The resulting `ToyWorkspace` object is then available for further processing or for use in other Dagger Functions.
+In the `Env`, a `ToyWorkspace` called `after` is specified as a desired output of the LLM. This means that the LLM should return the `ToyWorkspace` object as a result of completing its task. The resulting `ToyWorkspace` object is then available for further processing or for use in other Dagger Functions.
```
vikram-dagger, line 134:
```suggestion
In the `Env`, a `ToyWorkspace` instance called `after` is specified as a desired output of the LLM. This means that the LLM should return the `ToyWorkspace` module instance as a result of completing its task. The resulting `ToyWorkspace` object is then available for further processing or for use in other Dagger Functions.
```

vikram-dagger: I find the terminology a bit confusing - we have environments, modules, objects . I added module instance but I'm not sure this helps.

docs/current_docs/api/llm.mdx
vikram-dagger: I guess I'm confused because before this we mention modules, but in this paragraph we say ` For example, an environment might provide a `Directory`, a `Container`, and a `string` variable.` Should we instead say that `an environment might provide a module` to connect the two ideas?

docs/current_docs/quickstart/agent/snippets/go/main.go
```diff
@@ -15,7 +15,7 @@ func (m *CodingAgent) GoProgram(
 	environment := dag.Env().
 		WithToyWorkspaceInput("before", workspace, "these are the tools to complete the task").
```
vikram-dagger, line 16:
```suggestion
		WithToyWorkspaceInput("before", workspace, "these are the tools to complete the assignment").
```

We use `assignment` in all the other prompts, should this also use the same term?

docs/current_docs/quickstart/agent/snippets/go/main.go
```diff
@@ -15,7 +15,7 @@ func (m *CodingAgent) GoProgram(
...
 		WithStringInput("assignment", assignment, "this is the assignment, complete it").
-		WithToyWorkspaceOutput("after")
+		WithToyWorkspaceOutput("after", "the ToyWorkspace with the completed assignment")
```
vikram-dagger, line 18: We don't use the module name in the input prompt, but we use it in the output prompt. Should both prompts refer to the object in the same way?

docs/current_docs/api/llm.mdx
vikram-dagger: Suggesting this to see if it's clearer?

docs/current_docs/quickstart/agent/snippets/go/main.go
kpenfound: Not necessary IMO. In the input, it can see the type and it's functions. For the output, it could help to provide guidance

levlaz (commented):
Overall this is a really good quickstart. Adding some general feedback. 

1. Can we get rid of the fluff, I think our docs should be concise and helpful. A few things jumped out at me in this regard, but it exists in other places too. 

> seamless chaining

> Dagger Shell is the fastest way to interact with the Dagger API, allowing access to both core types and custom Dagger Functions using a familiar Bash-like syntax. Type dagger to launch Dagger Shell in interactive mode.

 Overall I feel the docs are not the place to "sell" dagger so using too many adjectives makes it feel off putting to me. 
 
 2. It would be better not to click out to other pages when trying to get through the guide. So we should either add some info on how to set up the simplest model in the "Configure the LLM endpoint" section, or make this a pre-requisite instead. 
 
 3. Should we encourage people to explore the ToyWorkspace API as well just to get a better sense of how all the bits fit together? Could do it in the same section where we get help for our new function. 
 
 4. Can we clean up some of the passive voice too?
 
 i.e **The container with the LLM's solution can then be obtained by chaining** should be **Obtain the container with the LLM's solution by chaining** ...
 
 5. My function returned a "hello world" instead of a curl clone. Is there any prompt that we can use that is more consistent? Feels like getting this to feel magical would be ideal.

jpadams (approved):
👍

PR #10033, opened 2025-04-01
fix nil dereference issue on LLM type

marcosnils:
Fixes 10032. Initializes `once` when LLM is created


vito (approved):
Ack, maybe a bad rebase - thanks! Strange that it passed evals, I guess they all ended up with `Clone()` being called somewhere.

PR #10038, opened 2025-04-01
env(privileged: Bool): create a privileged environment, with core API…

shykes:
`env(privileged: Bool)`: create a privileged environment, with core API…, host and current module access

Deprecates `LLM.withQuery()`


shykes:
Just realized this was dropped from the release

shykes:
> This is a lateral change, right? It doesn't address the inability for the model to switch back to `Query`? (Not a problem just making sure.)

Correct it only addresses the API, not the implementation issue. As discussed with @marcosnils

shykes:
> wait so privileged gives you all 3? core, host, and current module access? and privileged:false gives you none, then you construct accesses yourself?

Yes, same as `withQuery`

vito (approved):
Regen'd SDKs + docs, and fixed the CLI - all green now, so merging!

(Side note: the TestLLM golden example diffs are quite large, but seem OK considering what's being tested in each case.)

PR #10043, opened 2025-04-01
WIP dag.LLM MCP client support

cwlbraa:
this PR intends to allow users to run stdio MCPs within dagger and wire them up to their dagger LLMs.

it's extremely hacky atm, but works with this example:

```sh
with-dev dagger -M -c 'llm |\
  with-mcp $(container |\
             from "mcr.microsoft.com/playwright:v1.51.1-noble" |\
             with-entrypoint -- npx -y g-search-mcp) |\
  with-prompt "search google for hello world. use the us locale. show me the result urls." |\
  last-reply'
```

this is a fairly cool demonstration of the utility here. the g-search MCP uses playwright for browser emulation to search google. playwright's a pretty beefy dependency that can't run on any arbitrary OS... like it's not supported on alpine, but works on ubuntu. it happens to have a nice prepackaged container image that we can abuse.

other cool features of dagger as an MCP server runtime include:
- controlled access to secrets, like run the AWS MCP and only give it 1 specific AWS credential directly from your password manager.
- volume mounts for stitching together different host-filesystem-modifying MCPs 

things that are bad in the current state of this PR:
- the library used for MCP transport uses a bufio.NewScanner that chokes and blocks forever on responses >64kb in size. this is very easy to trigger with many MCP impls, like half of them never make it past tool registration
  - i'm gonna fork mark3labs/mcp-go to make things work for real
- extremely lazy handling of MCP output types. we let the tool return `[]Content{type, text, data, mimetype}` and don't really do anything to square the circle between dagger types and MCP/JSONRPC types.
  - need to see how this might or might not interact with Environment, too.
- Service lifecycle: we're running these services without any particular intent to stop them. uncertain if they hang around after the session ends. 
  - stdout is definitely still printing to OTEL logs even though in this case it's showing responses from the MCP servers.
  - the logs also aren't segregated from the rest of the lastReply sync call. the traces should be made to look more like normal service running.
  - there's a bunch of logspam that indicates we're failing to kill processes because they've already dead
- tcp/SSE MCP transport support. Should be fairly easy, just didn't start here. Likely requires args on WithMCP (stdio bool & port int).

### DIFF
--- a/core/llm.go
+++ b/core/llm.go
@@ -560,6 +560,23 @@ func (llm *LLM) WithSystemPrompt(prompt string) *LLM {
 	return llm
 }
 
+// Append a system prompt message to the history
+func (llm *LLM) WithMCP(ctx context.Context, ctr *Container, srv *dagql.Server) (*LLM, error) {
+	llm = llm.Clone()
+	bk, err := llm.Query.Buildkit(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get buildkit client: %w", err)
+	}
+	_, err = bk.MCPClient(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to instantiate MCP Client", err)
+	}
+
+	llm.env.mcpServers = append(llm.env.mcpServers)
+	llm.dirty = true
+	return llm, nil
+}
+
 // Return the last message sent by the agent
 func (llm *LLM) LastReply(ctx context.Context, dag *dagql.Server) (string, error) {
 	if err := llm.Sync(ctx, dag); err != nil {
--- a/core/llm.go
+++ b/core/llm.go
@@ -21,6 +21,7 @@ import (
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/dagger/dagger/dagql"
+	"github.com/dagger/dagger/dagql/call"
 	"github.com/dagger/dagger/engine"
 	"github.com/dagger/dagger/engine/client/secretprovider"
 )
@@ -476,7 +477,7 @@ func (llm *LLM) Clone() *LLM {
 
 // Generate a human-readable documentation of tools available to the model
 func (llm *LLM) ToolsDoc(ctx context.Context, srv *dagql.Server) (string, error) {
-	tools, err := llm.mcp.Tools(srv)
+	tools, err := llm.mcp.Tools(ctx, srv)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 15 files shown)


codefromthecrypt:
@cwlbraa I think this is closest to the scenario @shykes, many others and I were having [here](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/269#discussioncomment-12788891) about MCP and trace propagation in practice.

Just in the last hour I started digging around in dagger to figure out how the various otel bits work. There is definitely some interesting code both in presentation and propagation. The key part I am missing on the main branch is how MCP client -> 

MCP server internal to dagger works, and specifically how these traces marry up using only MCP transports (stdio or sse, and it seems maybe only stdio is supported).

Basically I would expect a need to inject (mcp client) and extract (mcp server) the trace context (I guess SpanID in dagger terms) on each function call. Otherwise you'd get the wrong parent (like a session not a function scoped), or a new trace.

Straight-forward ways could be to cheat a little and add a field to the JSON message of the function call, to stash the correct incoming context. That said, I've not found that going on here. It could also be addressed with a side-channel, or some post-export reconciliation (re-parenting), or even custom UI logic (though that would only apply to a UI we control).

Sorry about the prelude, but this context in mind, can you explain how this is resolved in dagger, possibly in this PR for the dagger managed MCP client with a stdio child process MCP server?

ezynda3:
Just wanted to mention that mcp-go no longer uses `NewScanner` in the client impl

cwlbraa:
> can you explain how this is resolved in dagger, possibly in this PR for the dagger managed MCP client with a stdio child process MCP server?

@codefromthecrypt trace parenting is not resolved here, but when executing the MCP server our existing OTEL implementation is configuring the standard OTEL env vars for collection- i think this is what @shykes means when he's talking about unified runtime.

per-request traceparent propogation as part of the MCP protocol is something we'd want to see happen.

cwlbraa:
@ezynda3 fwiw i'm referencing this other [mcp-go](https://github.com/riza-io/mcp-go/blob/main/stdio/stdio.go) using a bufio.NewScanner. 

when i get back to this branch, i intend to try again with a more up-to-date mark3labs/mcp-go, but i still think it will require major changes because [the transport implementations](https://github.com/mark3labs/mcp-go/blob/main/client/transport/stdio.go) assume that the client should be managing the server process rather than consuming it's stdin/stdout streams.

shykes:
@cwlbraa if you're interested I have a POC of a userland approach https://github.com/shykes/x/tree/main/mcp-sdk

It only extracts the tools, doesn't proxy the call yet... proxying to stdio without controlling the inside of the container will be tricky

codefromthecrypt:
@cwlbraa much belated thanks for clarifying the points!

cwlbraa:
https://cs.opensource.google/go/x/tools/+/master:internal/mcp/README.md 😈 the go maintainers are making progress

PR #10046, opened 2025-04-01
env: fix missing output binding key

tiborvass:
With this fix, `env | with-file-output foo description | output name` now returns `foo` instead returning an empty result.


vito (approved):
:man_facepalming: 

I suppose this never mattered because we ended up not pursuing outputs => shell env syncing, since it was unclear how to even set them up in the REPL.

PR #10057, opened 2025-04-02
docs: Improve Python constructor examples

vikram-dagger:


docs/current_docs/api/snippets/constructors/python/complex/main.py
```diff
@@ -0,0 +1,40 @@
...
+    dag,
+    DefaultPath,
+    Doc,
+    field,
+    object_type,
+)
+
+
+
+
```
TomChv, line 13: Why so much empty lines? Normally only 2 should be good no?

docs/current_docs/api/constructors.mdx
```diff
@@ -37,19 +37,6 @@ Here is an example module with a custom constructor:
...
-If a constructor argument needs an asynchronous call to set the default value, it's
-possible to replace the default constructor function from `__init__()` to
-a factory class method named `create`, as in the following code listing:
-
-:::warning
-This factory class method must be named `create`.
-:::
-
-```python file=./snippets/constructors/python/async/main.py
-```
```
TomChv, line 52: Why removing that?

TomChv (commented):
Left couple of comments :)

docs/current_docs/api/snippets/constructors/python/complex/main.py
vikram-dagger: Thanks for catching this!

docs/current_docs/api/constructors.mdx
helderco: It is not standard but it may be needed for more complex use cases or if you need any async call in the constructor.

So it think it should be somewhere, but not under "Simple Constructor". My concern is people using this even for simpler use cases or cases that are covered by the standard dataclass.

docs/current_docs/api/constructors.mdx
```diff
@@ -126,12 +117,26 @@ Here is an example of a Dagger module with a default constructor argument of typ
...
+You can use other [`@dataclasses.dataclass`](https://docs.python.org/3/library/dataclasses.html) decorators to generate more complex constructors. For example:
+
+```python file=./snippets/constructors/python/complex/main.py
+```
+
+A few important notes:
+
+- Class attributes become constructor arguments by default, unless specified as `field(init=False)`.
+- The [`__post_init__` function](https://docs.python.org/3/library/dataclasses.html#post-init-processing) is called at the end of the auto-generated constructor. So by setting `init=False` on the container argument, it changes from being a constructor argument to a regular argument that is still required.
+- Setting a default value like `None` in an attribute is the same as setting `@dataclasses.field(default=None)`. `@dataclasses.field()` is only required to specify other non-default values, like `init` or to set the default using a [factory function](https://docs.python.org/3/library/dataclasses.html#default-factory-functions).
```
helderco, line 130: Not decorators:
```suggestion
- Setting a default value like `None` in an attribute is the same as setting `dataclasses.field(default=None)`. `dataclasses.field()` is only required to specify other non-default values, like `init` or to set the default using a [factory function](https://docs.python.org/3/library/dataclasses.html#default-factory-functions).
```
This is standard dataclass. You could also need `dagger.field()` if you want it exposed as a function in the API, with a default value or not.

docs/current_docs/api/constructors.mdx
```diff
@@ -126,12 +117,26 @@ Here is an example of a Dagger module with a default constructor argument of typ
...
+:::info
+You can use other [`@dataclasses.dataclass`](https://docs.python.org/3/library/dataclasses.html) decorators to generate more complex constructors. For example:
+
+```python file=./snippets/constructors/python/complex/main.py
+```
+
+A few important notes:
+
+- Class attributes become constructor arguments by default, unless specified as `field(init=False)`.
+- The [`__post_init__` function](https://docs.python.org/3/library/dataclasses.html#post-init-processing) is called at the end of the auto-generated constructor. So by setting `init=False` on the container argument, it changes from being a constructor argument to a regular argument that is still required.
```
helderco, line 129: It's incorrect that it changes from being a constructor argument to a regular argument. It's both if `init=True`.
```suggestion
- Attributes are both fields and constructor arguments by default. So by setting `init=False` on the container argument, it's no longer added as a constructor argument but is still a required field that must be initialized if there's no default value.
- The [`__post_init__` function](https://docs.python.org/3/library/dataclasses.html#post-init-processing) is called at the end of the auto-generated constructor which can be used, for example, to initialize field values that depend on other field values.
```

docs/current_docs/api/constructors.mdx
```diff
@@ -126,12 +117,26 @@ Here is an example of a Dagger module with a default constructor argument of typ
...
+- Setting a default value like `None` in an attribute is the same as setting `@dataclasses.field(default=None)`. `@dataclasses.field()` is only required to specify other non-default values, like `init` or to set the default using a [factory function](https://docs.python.org/3/library/dataclasses.html#default-factory-functions).
+- The difference between `@dataclasses.field()` and `@dagger.field()` is that the latter exposes the attribute as a Dagger field, which automatically creates a getter function for it in the API.
+
 For default values that are more complex, dynamic or just [mutable](https://docs.python.org/3/library/dataclasses.html#mutable-default-values),
 use a [factory function](https://docs.python.org/3/library/dataclasses.html#default-factory-functions) without arguments in
 [dataclasses.field(default_factory=...)](https://docs.python.org/3/library/dataclasses.html#dataclasses.field):
```
helderco, line 135:
```suggestion
[`dataclasses.field(default_factory=...)`](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) or `dagger.field(default=...)`:
```

docs/current_docs/api/constructors.mdx
```diff
@@ -126,12 +117,26 @@ Here is an example of a Dagger module with a default constructor argument of typ
...
+
+```python file=./snippets/constructors/python/complex/main.py
+```
+
+A few important notes:
+
+- Class attributes become constructor arguments by default, unless specified as `field(init=False)`.
+- The [`__post_init__` function](https://docs.python.org/3/library/dataclasses.html#post-init-processing) is called at the end of the auto-generated constructor. So by setting `init=False` on the container argument, it changes from being a constructor argument to a regular argument that is still required.
+- Setting a default value like `None` in an attribute is the same as setting `@dataclasses.field(default=None)`. `@dataclasses.field()` is only required to specify other non-default values, like `init` or to set the default using a [factory function](https://docs.python.org/3/library/dataclasses.html#default-factory-functions).
+- The difference between `@dataclasses.field()` and `@dagger.field()` is that the latter exposes the attribute as a Dagger field, which automatically creates a getter function for it in the API.
```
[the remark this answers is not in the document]
vikram-dagger: I didn't entirely understand the first difference so it is likely that my edit is incorrect, please review it once.

docs/current_docs/api/constructors.mdx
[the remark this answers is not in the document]
vikram-dagger: I would leave it as is, it's better to have more examples than less.

docs/current_docs/api/constructors.mdx
vikram-dagger: Added this back

vikram-dagger:
Ping @helderco for re-review

docs/current_docs/api/constructors.mdx
```diff
@@ -112,6 +99,10 @@ If you plan to use constructor fields in other module functions, ensure that the
 
 Constructors can be passed both simple and complex types (such as `Container`, `Directory`, `Service` etc.) as arguments. Default values can be assigned in both cases.
 
+:::important
+Explicitly declare the type even when a default value is assigned, so that the Dagger SDK can extend the GraphQL schema correctly.
```
helderco, line 103: This may not make sense in all languages. In Go, for example, it won't compile if you don't and default values aren't even supported in the language, we do it via a specially formatted comment.

This is more general:

```suggestion
Explicitly declare the type even when it can be inferred, so that the Dagger SDK can extend the GraphQL schema correctly.
```

Even then it's possible for an SDK to actually support type inferrence when registering types to the API, depending on the language.

docs/current_docs/api/constructors.mdx
```diff
@@ -163,8 +182,6 @@ When assigning default values to complex types in TypeScript, it is necessary to
 </TabItem>
 </Tabs>
 
-It is necessary to explicitly declare the type even when a default value is assigned, so that the Dagger SDK can extend the GraphQL schema correctly.
-
 Here is an example call for this Dagger Function:
```
helderco, line 185: This part in the end is way too far from the code it applies to, which is the first code snippet in the "Default values for complex types" section.

I understand it's because of the general vs language specific parts but ends up being confusing. Maybe put the extra information inside a "tip" admonition so it's self contained?

docs/current_docs/api/constructors.mdx
```diff
@@ -126,13 +117,41 @@ Here is an example of a Dagger module with a default constructor argument of typ
...
+
+Here's an example:
+
+```python file=./snippets/constructors/python/complex/main.py
+```
+
+A few important notes:
+
+- Class attributes become constructor arguments by default, unless specified as `field(init=False)`.
+- Attributes are both fields and constructor arguments by default. So by setting `init=False` on the container argument, it's no longer added as a constructor argument but is still a required field that must be initialized if there's no default value.
```
helderco, line 132: These are basically saying the same thing.

```suggestion
- Data class fields become constructor arguments by default unless specified as `field(init=False)`. So, even though the `container` field is not added as a constructor argument it's still a required field that must be initialized  with either a default value or in post-init processing.
```

docs/current_docs/api/constructors.mdx
```diff
@@ -126,13 +117,41 @@ Here is an example of a Dagger module with a default constructor argument of typ
...
+- The [`__post_init__` function](https://docs.python.org/3/library/dataclasses.html#post-init-processing) is called at the end of the auto-generated constructor which can be used, for example, to initialize field values that depend on other field values.
+- Setting a default value like `None` in an attribute is the same as setting `dataclasses.field(default=None)`. `dataclasses.field()` is only required to specify other non-default values, like `init` or to set the default using a [factory function](https://docs.python.org/3/library/dataclasses.html#default-factory-functions).
+- There are two important differences between `dataclasses.field()` and `dagger.field()`
```
helderco, line 135:
```suggestion
- The `dagger.field` descriptor wraps `dataclasses.field`, adding two important differences:
```

docs/current_docs/api/constructors.mdx
```diff
@@ -126,13 +117,41 @@ Here is an example of a Dagger module with a default constructor argument of typ
...
+  - `dagger.field()` exposes the attribute as a Dagger Function in the API. However, `dagger.field()` doesn't expose everything in `dataclasses.field()`.
```
helderco, line 136:
```suggestion
  - `dagger.field()` is used to expose the attribute as a Dagger Function in the API (a simple getter), but it also becomes a part of the object's state.
```

docs/current_docs/api/constructors.mdx
```diff
@@ -126,13 +117,41 @@ Here is an example of a Dagger module with a default constructor argument of typ
...
+  - With `dataclass.field()`, you can specify separate `default` and `default_factory` arguments for either a simple value or a default using a callable respectively. However, with `dagger.field()`, the same `default` argument supports both cases.
```
helderco, line 137:
```suggestion
  - Not all arguments from `dataclasses.field()` can be set with `dagger.field()` which sets opinionated defaults, but you can specify both simple values and factory functions via the same `default` argument while with `dataclasses.field()` you need to use separate `default` and `default_factory` arguments.
```

docs/current_docs/api/constructors.mdx
```diff
@@ -126,13 +117,41 @@ Here is an example of a Dagger module with a default constructor argument of typ
...
+
+For default values that are more complex, dynamic or just [mutable](https://docs.python.org/3/library/dataclasses.html#mutable-default-values), use a [factory function](https://docs.python.org/3/library/dataclasses.html#default-factory-functions) without arguments in
+[`dataclasses.field(default_factory=...)`](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) or `dagger.field(default=...)`:
```
helderco, line 140:
```suggestion
[`dataclasses.field(default_factory=...)`](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) or [`dagger.field(default=...)`](https://dagger-io.readthedocs.io/en/latest/module.html#dagger.field):
```

docs/current_docs/api/constructors.mdx
```diff
@@ -126,13 +117,41 @@ Here is an example of a Dagger module with a default constructor argument of typ
 ```python file=./snippets/constructors/python/default-object/main.py
 ```
 
-For default values that are more complex, dynamic or just [mutable](https://docs.python.org/3/library/dataclasses.html#mutable-default-values),
-use a [factory function](https://docs.python.org/3/library/dataclasses.html#default-factory-functions) without arguments in
-[dataclasses.field(default_factory=...)](https://docs.python.org/3/library/dataclasses.html#dataclasses.field):
+:::important
+Dagger `@object_type` classes are Python [data classes](https://docs.python.org/3/library/dataclasses.html) with the Dagger API exposed. This means that you can adjust the generation of the constructor using standard data class features, such as attributes and post-init processing.
+:::
```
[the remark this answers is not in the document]
vikram-dagger: See if better?

docs/current_docs/api/constructors.mdx
vikram-dagger: Switched to a partial for the output, which makes it easier to handle.

docs/current_docs/api/constructors.mdx
```diff
@@ -120,19 +113,52 @@ Here is an example of a Dagger module with a default constructor argument of typ
...
 ```
 
-For default values that are more complex, dynamic or just [mutable](https://docs.python.org/3/library/dataclasses.html#mutable-default-values),
-use a [factory function](https://docs.python.org/3/library/dataclasses.html#default-factory-functions) without arguments in
-[dataclasses.field(default_factory=...)](https://docs.python.org/3/library/dataclasses.html#dataclasses.field):
+<PartialComplexConstructorOutput />
+
+#### Multiple complex types
+
+Since `@object_type` classes are Python data classes, it is possible to provide multiple complex types as constructor arguments:
```
helderco, line 128:
```suggestion
#### Dagger objects are Python data classes

Since `@object_type` classes are [Python data classes](https://docs.python.org/3/library/dataclasses.html), you can fine tune the generation of the constructor in different ways. 

Here is a more complex example:
```

helderco (commented):
I like new sub-sections. There's just one last detail.

helderco (approved):
Thank you! 🙏

PR #9123, opened 2024-12-06
feat: support Source Map in TypeScript SDK

TomChv:
Depends on #9031

Implement `SourceMap` to display the actual position of the definition when codegen.

It's very useful to navigate inside the codebase when using local dependencies since generated function to call
the dependency doesn't directly use the dep's code but instead call it through the Dagger API.

Same as go in #8498

### TODO
- [x] Registration by the TS SDK (tested manually for now)
- [x] Generation on the TS client (tested manually for now)
- [x] Integration tests
- [x] Add column support
- [ ] Add doc

### Example of generation

```ts
export type TestLocationContainerOpts = {
  name: string // test-location (../../../../../../../dagger/location/ts-to-go/test-location/src/index.ts:25)
}

export type TestLocationHumanOpts = {
  name: string // test-location (../../../../../../../dagger/location/ts-to-go/test-location/src/index.ts:20)
  age: number // test-location (../../../../../../../dagger/location/ts-to-go/test-location/src/index.ts:20)
}

export enum TestLocationFoo { // test-location (../../../../../../../dagger/location/ts-to-go/test-location/src/index.ts:6)
  Bar = "bar",
  Baz = "baz",
  Foo = "foo",
}

export class TestLocationPerson extends BaseClient { // test-location (../../../../../../../dagger/location/ts-to-go/test-location/src/index.ts:12)
  ...

  age = async (): Promise<number> => { // test-location (../../../../../../../dagger/location/ts-to-go/test-location/src/index.ts:12)
  ... 
  }
}
```

### Demo

https://github.com/user-attachments/assets/f64654b8-c3ff-4a6f-b127-29011b9993c8

PR #10058, opened 2025-04-02
fix: mark `llm`/`env` apis as experimental

jedevc:
This patch allows marking certain APIs as `experimental`, just like we allow marking certain APIs as `deprecated`.

To do this, it's mostly just plumbing in the graphql layer, however, a few more substantial changes are also included:
- In python, we add a new `insert_stubs` to dynamically add the directives from our custom introspection API (added for go in #8498, and for typescript in #9123). This is required so that we can read the new `experimental` annotation during docs generation.
	- I cherry-picked this (with some minor changes) across from #9518 where it was needed to add support to the upcoming `enumValue` directive.
- We switch from using https://github.com/CDThomas/graphql-json-to-sdl to our own implementation in Go. By taking control of this logic, we can more closely control the output: importantly outputting any custom directives. This doesn't seem to be supported by any generator I could find, but it was pretty simple to just port the upstream graphql version to go.

TODO:
- [x] add dagql plumbing
- [x] sdks
  - [x] go
  - [x] python
  - [x] typescript
- [x] `schema.graphqls`


jedevc:
(note: failures seem to ephemeral, and aren't related - see [discord](https://discord.com/channels/707636530424053791/1003718839739105300/1358837478672433332))

dagql/objects.go
```diff
@@ -1060,6 +1067,23 @@ func (field Field[T]) ArgDeprecated(name string, paras ...string) Field[T] {
 	panic(fmt.Sprintf("field %s has no such argument: %q", field.Spec.Name, name))
 }
 
+func (field Field[T]) ArgExperimental(name string, paras ...string) Field[T] {
```
vito, line 1070: nice - left a comment noting we'll want this too but ya beat me to it :)

sdk/go/dag/dag.gen.go
```diff
@@ -97,6 +97,8 @@ func Engine() *dagger.Engine {
 }
 
 // Initialize a new environment
+//
+// Experimental: "Environments are not yet stabilized"
```
vito, line 101: These quotes seem a bit off

dagql/testdata/introspection.json
```diff
@@ -88,7 +88,7 @@
                 "args": [
                   {
                     "name": "reason",
-                    "value": "use something else"
+                    "value": "\"use something else\""
```
vito, line 91: Guessing same root cause for quote weirdness?

dagql/introspection/types.go
```diff
@@ -263,7 +263,7 @@ func Install[T dagql.Typed](srv *dagql.Server) {
 			if self.Value == nil {
 				return dagql.Null[dagql.String](), nil
 			} else {
-				return dagql.NonNull(dagql.NewString(self.Value.Raw)), nil
+				return dagql.NonNull(dagql.NewString(self.Value.String())), nil
```
vito, line 266: Maybe `.String()` here is adding the quotes...?

dagql/testdata/introspection.json
jedevc: Ah, so this is actually intentional. I updated it to return `.String` instead of `.Raw` - that way it works similarly to `defaultValue`. Otherwise, `string`s just don't have quotes, and can't really be distinguished from `int`s, etc.

sdk/go/dag/dag.gen.go
jedevc: Oops, that was a go specific problem, have updated there :tada:

dagql/objects.go
jedevc: Ah, I think that while the annotation is correctly propagated, it's not actually propagating through to the codegen... yet.

dagql/introspection/types.go
jedevc: (see above)

docs/docs-graphql/schema.graphqls
vito, line 1: Nice, assuming these preserve definition order now?

vito (approved):

docs/docs-graphql/schema.graphqls
jedevc: Yup, the `arg`s do - the fields now follow a *different* alphanumeric ordering (that makes way more sense, it respects casing now - which it didn't used to).

I think this is probably still what we want? But it's easy to change it up again in the future if we want to.

PR #10075, opened 2025-04-06
Update module-dependencies.mdx

lamalex:
Removed incorrect note about subdirectory requirements

### DIFF
--- a/docs/current_docs/api/module-dependencies.mdx
+++ b/docs/current_docs/api/module-dependencies.mdx
@@ -143,7 +143,7 @@ dagger call example --build-src=https://github.com/golang/example#master:/hello 
 </TabItem>
 </Tabs>
 
-You can also use local modules as dependencies. For example:
+You can also use local modules as dependencies, as long as they are in the same git repository. For example:
 
 ```shell
 dagger install ./path/to/module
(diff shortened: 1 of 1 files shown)


docs/current_docs/api/module-dependencies.mdx
```diff
@@ -143,7 +143,7 @@ dagger call example --build-src=https://github.com/golang/example#master:/hello
 </TabItem>
 </Tabs>
 
-You can also use local modules as dependencies. However, they must be stored in a sub-directory of your module. For example:
+You can also use local modules as dependencies. For example:
```
shykes, line 146:
```suggestion
You can also use local modules as dependencies, as long as they are in the same git repository. For example:
```

shykes (approved):
thank you! I added a small suggestion to mention git scoping, otherwise looks good.

lamalex:
That note was actually already on the page right below this, which is why I hadn’t added it but i guess doubled is twice as nice

shykes:
> That note was actually already on the page right below this, which is why I hadn’t added it but i guess doubled is twice as nice

Oh sorry I missed that! (was reviewing on my phone).

In any case thank you for the fix 🙏

PR #10087, opened 2025-04-07
fix error when parsing multi-line git config files

marcosnils:

### DIFF
--- a/engine/session/git.go
+++ b/engine/session/git.go
@@ -195,7 +195,7 @@ func (s GitAttachable) GetConfig(ctx context.Context, req *GitConfigRequest) (*G
 	gitMutex.Lock()
 	defer gitMutex.Unlock()
 
-	cmd := exec.CommandContext(ctx, "git", "config", "-l")
+	cmd := exec.CommandContext(ctx, "git", "config", "-l", "-z")
 	var stdout, stderr bytes.Buffer
 	cmd.Stdout, cmd.Stderr = &stdout, &stderr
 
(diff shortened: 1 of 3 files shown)


grouville (approved):
LGTM ✅ Please make sure to rebase prior merging 🙏

engine/session/git.go
```diff
@@ -233,26 +233,69 @@ func parseGitConfigOutput(output []byte) (*GitConfig, error) {
...
+		if strings.HasSuffix(value, "\\") {
+			inMultiline = true
+			currentKey = key
+			currentValue = value[:len(value)-1] // emove trailing backslash
+		} else if isGitConfigKeyAllowed(strings.ToLower(key)) {
+			entries = append(entries, &GitConfigEntry{
+				Key:   key,
+				Value: value,
+			})
+		}
```
jedevc, line 284: This feels duplicated from above. Can we refactor so that this block only appears once?

engine/session/git.go
```diff
@@ -233,26 +233,69 @@ func parseGitConfigOutput(output []byte) (*GitConfig, error) {
...
+
 		parts := strings.SplitN(line, "=", 2)
 		if len(parts) != 2 {
 			return nil, fmt.Errorf("invalid format: line %q doesn't match key=value pattern", line)
 		}
 
-		if isGitConfigKeyAllowed(strings.ToLower(parts[0])) {
+		key := parts[0]
+		value := parts[1]
```
jedevc, line 272: FYI, this can be simplified to use `strings.Cut(line, "=")`.

engine/session/git.go
marcosnils: not really the same block though, this one checks the start of the multiline while the above has is only valid while you're parsing a multi-line block and need to switch some variables. I could refactor but it won't improve readability that much since I still need to manage the special logic on each block :pray:

marcosnils:
let's not merge this yet. Just found out it needs some changes :pray:

grouville:
> let's not merge this yet. Just found out it needs some changes 🙏

Oh, what was the edge case ? 👀

PR #10089, opened 2025-04-07
docs: llms.txt and llms-full.txt support

jpadams:
Autogenerated from docs.

Using code from:
https://github.com/prisma/docs/blob/main/docusaurus.config.ts#L95

from this issue:
https://github.com/facebook/docusaurus/issues/10899 

`dagger -c 'server | up'`

Sample output:
https://gist.github.com/jpadams/864dda6b3df0e0536497efd5be315405 `llms.txt`
https://gist.github.com/jpadams/9cbf07138b2e33baf57d9b8558f05265 `llms-full.txt`

closes ECO-430


docs/docusaurus.config.ts
```diff
@@ -82,6 +85,103 @@ const config: Config = {
       },
     ],
     */
+    async function pluginLlmsTxt(context) {
```
jedevc, line 88: suggestion: if this is from prisma, might be worth putting a link to the original here.

docs/docusaurus.config.ts
```diff
@@ -82,6 +85,103 @@ const config: Config = {
...
+      return {
+        name: "llms-txt-plugin",
+        loadContent: async () => {
```
jedevc, line 91: any chance we could split this plugin to `docs/plugins/llm/index.ts`?

jpadams: Can try. 👍

docs/docusaurus.config.ts
jpadams: done

docs/docusaurus.config.ts
jpadams: done

jedevc (approved):

docs/docusaurus.config.ts
jedevc: Nice! Thanks :pray: :yellow_heart:

PR #9770, opened 2025-03-04
docs: add security-by-design docs

grouville:
Closes #9723

Users have been confused by https://docs.dagger.io/cookbook/#clone-a-remote-git-repository-into-a-container-by-branch-tag-or-commit. The source of the confusion seems to be deeper than just the reproduction of the cookbook.

Basically, we **lack an understanding** of the security-by-design model in our docs.

This PR updates the cookbook example to differentiate the HTTP(S) and SSH based refs and also introduces a `security-model` page. This page is currently not linked to any other page in our docs as I don't really know where it would fit @vikram-dagger

PR #10092, opened 2025-04-08
docs: Make Git cloning recipes consistent in cookbook

vikram-dagger:
Follow up on #9770


grouville (approved):
Thank you, great addition

Issue #10091, opened 2025-04-08
🐞 typescript-sdk with bun times out when it takes longer than 5 minutes


MatthiasGrandl:
### What is the issue?

When a job takes longer than 5 minutes, it times out. This only affects the bun runtime. node works fine. This is due to: https://github.com/oven-sh/bun/pull/6217

[Link to Discord thread](https://discord.com/channels/707636530424053791/1347311312975302718)

### Dagger version

v0.18.2

### Steps to reproduce

```ts
  /**
   * Bun timeout
   */
  @func()
  async bunTimeoutRepro(): Promise<string> {
    return dag
      .container()
      .from("alpine:latest")
      .withExec(["sleep", "10m"])
      .stdout();
  }
```

### Log output

```
Error logs:

✘ .bunTimeoutRepro: String! 5m1s
AbortError: The operation timed out
Error: Encountered an unknown error while requesting data via graphql
! process "bun /src/.dagger/src/__dagger.entrypoint.ts" did not complete successfully: exit code: 1
Error: input: test.bunTimeoutRepro process "bun /src/.dagger/src/__dagger.entrypoint.ts" did not complete successfully: exit code: 1
```

PR #10094, opened 2025-04-08
feat: generate TS SDK as bundled library.

TomChv:
**Summary**

- Remove the generation of the full sdk directory but instead generate a smaller bundle (4.5M).
   This allows better maintenance and simpler dependency management for the user.
I- t also speed up the init/call time since the only required dependency is `typescript`.
- Update TS runtime to use the bundle SDK and remove previous lib generation.
- Update package.json / tsconfig.json / deno.json to use the bundled library.
- Update export on the Typescript SDK to make accessible the telemetry package in the bundle.
- Add rollup bundle configuration in  the Typescript SDK. Add bundling steps in the TS SDK build process.
- Update integrations tests and generator to support bundle mode.
- Add migration guide to bundle SDK (the sdk directory must be deleted manually on the host because the runtime cannot delete existing files from host, it will simply copy the bundle lib and never use the old SDK, however it will still be present on the host)

**EDIT**: Add support for both bundle and local mode.

**:bulb: Questions :bulb:**:
- Should we support both mode bundle / lib?

**EDIT**: it's done and tested :)

- Should I add a "remote mode" based on the `package.json` value 
  - if `@dagger.io/dagger="xxx.xx"` -> remote lib
  - if `@dagger.io/dagger=./sdk` -> convert to bundle and remove this line OR bundle the whole library?
  - if `@dagger.io/dagger` doesn't exist -> bundle mode

**EDIT**: Done, for local and bundle mode, remote lib will come in a follow up later.

 - Should I make this directory committable? The SDK will always try to regenerate it so it's not that useful IMO but I want your opinions

**EDIT**: I talked with @helderco and we could make the client binding commitable but not the entire directory. So this will happen later in a follow up.

:rocket: Side effect: this fixes https://github.com/dagger/dagger/issues/10091 because the bundle library is targeting node so fetch isn't limit to a 5min request in the node lib. 

**TODO**:
- Use bundle in generated client


TomChv:
**Benchmark**
:bulb: The lib benchmark is done with dagger `v0.18.2`
:bulb: I removed the time to load the TS SDK because this only happens on the 1st run after building the dev engine so it's not relevant to our benchmarks.
:bulb: I'm only measuring the `loadModule` step, not the actual execution since it depends on what the user is running.

**Steps**
- Create an empty directory
- `git init` to isolate the module context to avoid parent file to be copy by the engine
- Init the project (`init measure`)
- `dagger call container-echo --string-arg foo stdout` (`call no cache measure`)
- `dagger call container-echo --string-arg foo stdout` (`call cache measure`)
- Update the module file with a comment
- `dagger call container-echo --string-arg foo stdout` (`call after edition measure`)

**Bundle benchmark**

| Setup | Init | Call (no-cache) | Call (cache) | Call (after edition) | 
|-------|------|-----------------|--------------|----------------------|
| Node yarn | [11.5s](https://v3.dagger.cloud/Quartz/traces/e39a616cae4f5d46a28d75fe25108fc4)  | [14.3s](https://v3.dagger.cloud/Quartz/traces/a5c8b0d151326402515eb6b5098ece47) | [2.3s](https://v3.dagger.cloud/Quartz/traces/c8f91ef85e88943da90187eae625e026) | [6.3s](https://v3.dagger.cloud/Quartz/traces/475a6f81bc714d7844a8b2b716971fc7) |
| Node npm | [14.1s](https://v3.dagger.cloud/Quartz/traces/35f57e3743ccf2b5e59fa0288f54374a)  | [13.3s](https://dagger.cloud/Quartz/traces/c88815df1ed7938f5f59a1256ba254ba)  | [0.8s](https://dagger.cloud/Quartz/traces/e46b58d55d30f8ac937e27cb6fdbe76b) | [4.5s](https://dagger.cloud/Quartz/traces/36512b726140f8ac11fe038ac9395999) |
| Node pnpm | [9.7s](https://v3.dagger.cloud/Quartz/traces/120fe007dc40367f3507efb604d72bfb)  | [8.9s](https://dagger.cloud/Quartz/traces/b921f63791ad8cf0853ff2bdb6ddb79f)  | [0.8s](https://dagger.cloud/Quartz/traces/af05a22526ad2c880e0ba16761bb39c5) | [4.3s](https://dagger.cloud/Quartz/traces/9582b11f797854b411b46da697aab2f8) |
| Deno | [8.0s](https://v3.dagger.cloud/Quartz/traces/f625a64ba03b68bf6803d871863fb8ee)  | [8.5s](https://dagger.cloud/Quartz/traces/0dbeb262400c7a2e8ed55cdd919925cd)  | [0.8s](https://dagger.cloud/Quartz/traces/a00b5285b7125eb7d48a7cb945b484e9) | [3.8s](https://dagger.cloud/Quartz/traces/d0809e554399bb692b42ea1b2e129b02) |
| Bun  | [5.0s](https://v3.dagger.cloud/Quartz/traces/eb54f0b869701ee03d277055cc6676d4) | [7.1s](https://dagger.cloud/Quartz/traces/29946575e3c54e458472a8e551458e5d)  | [0.8s](https://dagger.cloud/Quartz/traces/21c7078b91b3504850da82995396aea8) | [3.3s](https://dagger.cloud/Quartz/traces/b003178e66257457c2fdbaebb6fd491d) |

**Lib benchmark**

| Setup | Init | Call (no-cache) | Call (cache) | Call (after edition) | 
|-------|------|-----------------|--------------|----------------------|
| Node yarn | [21.3s](https://v3.dagger.cloud/Quartz/traces/73550e4529a387df289a7bc5d6be1264)  | [18.8s](https://v3.dagger.cloud/Quartz/traces/bcbc8f1a933bfbb2eb02d0789090e226) | [2.2s](https://v3.dagger.cloud/Quartz/traces/bee9191d2cd0c10c37d0c70d2dc90b14) | [7.5s](https://dagger.cloud/Quartz/traces/eedf7b54b8ed8f4c9f2bcd3112837431) |
| Node npm | [26.5s](https://v3.dagger.cloud/Quartz/traces/997846a136a8bdb57ae40780a1953233)  | [17.7s](https://dagger.cloud/Quartz/traces/3a5237c256ba2bce7670caf21c410680) | [0.8s](https://dagger.cloud/Quartz/traces/a1beece57051bd86262c364a91e5e1fa) | [5.1s](https://dagger.cloud/Quartz/traces/26e8119a6fbfb117085f80e596786729) | 
| Node pnpm | [15.2s](https://v3.dagger.cloud/Quartz/traces/fbe2a7af26ce0c870a59eaee86ba367a)  | [16.4s](https://dagger.cloud/Quartz/traces/262620a97db459bcb8ea1a01a3f32e7c)  | [0.8s](https://dagger.cloud/Quartz/traces/3cdb920f6a444fe596f006a631c303e5) | [6.5s](https://dagger.cloud/Quartz/traces/54db7349630b838d175d49ac33d9f68f) |
| Bun | [11.0s](https://v3.dagger.cloud/Quartz/traces/c3b6fe12021f11f0372622b246915120) | [6.8s](https://dagger.cloud/Quartz/traces/7288bc6ff3e7f34fdb3c6f0076186870) | [0.9s](https://dagger.cloud/Quartz/traces/e87da52966f321552a103245c893918c) | [4.0s](https://dagger.cloud/Quartz/traces/c207e6b2e3c9dc4118178878ddfd34c7) | 
| Deno | [8.1s](https://v3.dagger.cloud/Quartz/traces/b1c97bb0fd2c8c1a5bfb6bca507fabe7)  | [25.1s](https://dagger.cloud/Quartz/traces/ecc17b3b09a949f0f0dffe576b4e3aed)  | [0.9s](https://dagger.cloud/Quartz/traces/cc27e8336dc2b846884cf968c9cf9384) | [6.6s](https://dagger.cloud/Quartz/traces/4935c71372f26be01f01a1e0640e4b17) |

**Performence diff result (in %)**
:bulb: Difference from lib to bundle (- means bundle is faster, + means lib is faster)

| Setup | Init | Call (no-cache) | Call (cache) | Call (after edition) | 
|-------|------|-----------------|--------------|----------------------|
| Node yarn | -50% | -24% | +4.4% | -17%  |
| Node npm | -47% | -28% | 0% | -12%  |
| Node pnpm | -35% | -48% | 0% | -30%  |
| Bun | -31%  | +22% | -11% | -1%  |
| Deno | -30%  | -200% | -11% | -50% |

As we can observe, in most case it's much faster to use the bundle library than the generated library.
We can reach an overall 30% performance improvement in cold call (no-cache) and an overall 15% improvement on call after editing module's file.

shykes:
Nicely done! On top of the performance improvement, this should make it easier to run Dagger behind corporate proxies

helderco (commented):
It's not entirely clear to me how this handles pinned or additional user dependencies. Can you clarify?

helderco:
> * Remove the generation of the full sdk directory but instead generate a smaller bundle (4.5M).

The `sdk` dir should only contain the library source files. It's `node_modules` that I expect to be the larger directory. Unless you mean it's smaller compared to previous `sdk` + `node_modules`. Does this PR remove the need for `node_modules`? What if a dependency has binary files and not just pure TypeScript code? If `node_modules` is still needed, does it copy the bundle or is it only additional dependencies?

> * Should we support both mode bundle / lib?

I think so. Default to bundle unless user has anything that contradicts it.

> * Should I add a "remote mode" based on the `package.json` value
>   
>   * if `@dagger.io/dagger="xxx.xx"` -> remote lib
>   * if `@dagger.io/dagger=./sdk` -> convert to bundle and remove this line OR bundle the whole library?
>   * if `@dagger.io/dagger` doesn't exist -> bundle mode
> * Should I make this directory committable? The SDK will always try to regenerate it so it's not that useful IMO but I want your opinions

My approach in https://github.com/dagger/dagger/pull/9538 is to only make the client bindings committable. That’s independently from either vendoring, pinning, or bundling the client library. Ideally I want to move from vendoring to pinning published package, then bundling is just providing packages locally instead of having to reach for the package registry. That could happen only within the module runtime though, not necessarily also when developing.

For supporting the offline use case, I think it will require adding a global `--offline` flag to the CLI, because we can’t go pulling container images too in that case. Then I’d enforce the bundle, even locally, if that flag is used.

TomChv:
> It's not entirely clear to me how this handles pinned or additional user dependencies. Can you clarify?

The user still has its package.json, with a `typescript` dependency when he inits, so he's free to add any dependencies and they will be downloaded by the runtime.
We still have a `install dependency` step in the runtime, for these user's specific dependency, the big difference is that the SDK directory doesn't requires any dependency to run (no `sdk/package.json` or sub local package anymore)

> EDIT: to be clear, I'm thinking it's valid for a user to want to pin a dependency of the dagger client library, for some reason. It's also possible the module also depends on a library dependency so there could be a version conflict if we're too strict on enforcing our own pin.

The user can't no more pin a dependency of the dagger client library because it's bundled.
So he doesn't have to, the library is already self contained except for `typescript` because the user actually needs it in his module.

> The sdk dir should only contain the library source files. It's node_modules that I expect to be the larger directory. Unless you mean it's smaller compared to previous sdk + node_modules. Does this PR remove the need for node_modules? What if a dependency has binary files and not just pure TypeScript code? If node_modules is still needed, does it copy the bundle or is it only additional dependencies?

The local SDK lib doesn't rely on the node_modules anymore except for `typescript`, that's the whole point of the bundling :)
If a user has specific dependency, binaries or anything, it will be downloaded in the runtime, but as I said above, this will only include user's specific dependency and `typescript`.

> I think so. Default to bundle unless user has anything that contradicts it.

Then current users will need to manually migrate to bundle mode, because to know if we should bundle or not, I'll need to check the `package.json` `@dagger.io/dagger` dependency.
Another way would be to add a field `dagger.bundle: false` in the `package.json`, so the user can manually specify the behaviour of the SDK. 
Is it a ok?

> For supporting the offline use case, I think it will require adding a global --offline flag to the CLI, because we can’t go pulling container images too in that case. Then I’d enforce the bundle, even locally, if that flag is used.

Yeah definitely, for now even with the bundling, `offline` isn't possible but it's a big step forward!

helderco:
For comparison: 

|                 | After `dagger init` | After `yarn install` |
| --------------- | ------------------- | -------------------- |
| Stable `dagger` | 664.0 KiB           | 95.1 MiB             |
| This PR         | 4.5 MiB             | 28.6 MiB             |

The reason is that `node_modules/typescript` alone is 24.0 MiB, but that isn’t actually needed for IDE completions locally.

So, for IDE completions, with this PR you only need `dagger develop`. It’s funny that I’m going on the opposite direction in Python in order to not require `dagger develop` but `uv sync` would still be required. 😄

sdk/typescript/runtime/main.go
```diff
@@ -488,53 +497,118 @@ func (t *TypescriptSdk) configureModule(ctr *dagger.Container) *dagger.Container
...
+			WithDirectory("/", dag.CurrentModule().Source().Directory("bundled_static_export/module"))
+	case Local:
+		return t.SDKSourceDir.
+			WithoutDirectory("codegen").
+			WithoutDirectory("runtime").
+			WithoutDirectory("tsx_module").
+			WithoutDirectory("bundled_Lib").
+			WithoutDirectory("src/provisioning")
+	default:
+		// This should never happens.
```
helderco, line 551:
```suggestion
		// This should never happen.
```

helderco (approved):

Issue #7358, opened 2024-05-11
✨ Make secret name optional


sagikazarmark:
### What are you trying to do?

I have a function that creates a secret: https://daggerverse.dev/mod/github.com/sagikazarmark/daggerverse/registry-config@b45dbd7448bb967aca4a538af9ce7f042abf0316#RegistryConfig.secret

The current secret API takes a name at the moment: `dag.SetSecret("name", "value")`.

I'm not sure, but my guess is that name is supposed to be unique (ie. to make sure there is no interference between calls to the same function).

### Why is this important to you?

Right now I'm working around the issue mentioned above by exposing an optional secret name argument in the API. I'd rather not do that.

### How are you currently working around this?

I'm planning to generate a secret name based on the content: https://github.com/sagikazarmark/daggerverse/pull/104

That will make sure it's unique enough and won't require me to expose options in the API for the secret name.

shykes:
FYI @jedevc @vito @helderco @aluzzardi I think we may need to revisit the design decisions that led to this point. Since then, we launched functions, modules, and didn't revisit the initial plan of adding "secrets providers". Is the current design still relevant? Do we still need the secret name?

helderco:
One use case we have today is when adding secrets to a docker build. The secret name will be turned into the secret's `id` in the [Dockerfile](https://github.com/MikaelElkiaer/dagger-examples/blob/ae8a49103ea086f09eb50a423d4bed4871cb0c12/Dockerfile#L2):
```
RUN --mount=type=secret,id=secret-file curl
```

One problem today is that if you're passing a secret via the CLI, the secret name is an hash. So you currently need to work around that by either creating a new secret inside your function, with the plaintext of the `Secret` argument and the name that you use in the `Dockerfile`, or get the secret's name and pass it to the build as a build argument.

<details><summary>Build arg example</summary>

Fix for https://github.com/MikaelElkiaer/dagger-examples/blob/ae8a49103ea086f09eb50a423d4bed4871cb0c12/dockerfile-secrets/dagger/main.go:

```go
func (m *DockerfileSecrets) Build(ctx context.Context, buildContext *Directory, secretFile *Secret) (*Container, error) {
	secretName, err := secretFile.Name(ctx)
	if err != nil {
		return nil, err
	}
	return buildContext.
		DockerBuild(dagger.DirectoryDockerBuildOpts{
			BuildArgs: []BuildArg{
				{Name: "SECRET_ID", Value: secretName},
			},
[...]

PR #10096, opened 2025-04-08
feat(engine): add secretArg support to docker build 

### LINKED ISSUE #9437: Update `Directory.DockerBuild` Secrets args types
Secrets args should be the same as BuildArg and be a list of object with a name and the secret value so we don't have to set the name with in  a [hacky way](https://docs.dagger.io/cookbook#use-secret-in-dockerfile-build).

Also explained by @helder at https://github.com/dagger/dagger/issues/7358#issuecomment-2321481240

```go
type SecretArg struct {
   Name string
   Value *Secret
}

// DirectoryDockerBuildOpts contains options for Directory.DockerBuild type DirectoryDockerBuildOpts struct {
 	// They will be mounted at /run/secrets/[secret-name]. 	
    Secrets []*SecretArg
}
```

/cc @shykes @vito @sipsma

rajatjindal:
This PR adds SecretArg to docker build api.  fixes #9437


core/container.go
```diff
@@ -1699,6 +1695,25 @@ func (container *Container) command(opts ContainerExecOpts) ([]string, error) {
 	return args, nil
 }
 
+// need bikeshedding for the name. this is basicaly SecretArg + actual secret
+type SecretArgInternal struct {
+	Name     string
+	SecretID SecretID
+	Secret   *Secret
```
jedevc, line 1702: I think we can remove the need for this by having a `dagql.Instance` here. e.g. https://github.com/dagger/dagger/blob/e40bb3b8f341ffe6ad70ad46d6912716cdf711c7/core/service.go#L50-L51

core/schema/container.go
```diff
@@ -821,14 +869,39 @@ func (s *containerSchema) build(ctx context.Context, parent *core.Container, arg
...
+	for i, arg := range args.SecretArgs {
+		secret, ok := secretStore.GetSecret(arg.Value.Value.ID().Digest())
+		if !ok {
+			return nil, fmt.Errorf("secret %q not found", arg.Value.Value.ID().Digest())
+		}
+
+		// if secret name is not explicitly provided, fallback to fetching
+		// from the secret store
+		secretName := arg.Value.Name
+		if arg.Value.Name == "" {
```
jedevc, line 882: :eyes: this should only be for the fallback case right? It shouldn't be possible to trigger this by passing an empty string to the new API.

core/schema/directory.go
```diff
@@ -120,6 +121,21 @@ func (s *directorySchema) Install() {
...
 				pid 1 process in the container. Otherwise it may result in unexpected behavior.`,
 			),
+		dagql.NodeFunc("dockerBuild", s.dockerBuild).
+			View(AfterVersion("v0.18.3")).
+			Doc(`Builds a new Docker container from this directory.`).
+			ArgDoc("dockerfile", `Path to the Dockerfile to use (e.g., "frontend.Dockerfile").`).
+			ArgDoc("platform", `The platform to build.`).
+			ArgDoc("buildArgs", `Build arguments to use in the build.`).
+			ArgDoc("target", `Target build stage to build.`).
+			ArgDoc("secretArgs", `Secrets to pass to the build.`,
```
marcosnils, line 131: nit: the `secretArgs` here is different (shorter) than the one in `schema/container.go`. Just checking if that's intentional

core/schema/directory.go
```diff
@@ -458,22 +474,44 @@ func (s *directorySchema) dockerBuild(ctx context.Context, parent dagql.Instance
...
+	for i, arg := range args.SecretArgs {
+		secret, ok := secretStore.GetSecret(arg.Value.Value.ID().Digest())
+		if !ok {
+			return nil, fmt.Errorf("secret %q not found", arg.Value.Value.ID().Digest())
+		}
+
+		// if secret name is not explicitly provided, fallback to fetching
+		// from the secret store
+		secretName := arg.Value.Name
+		if arg.Value.Name == "" {
```
marcosnils, line 492: same comment as `schema/container.go`

core/secret.go
```diff
@@ -195,7 +206,15 @@ func (store *SecretStore) GetSecretPlaintext(ctx context.Context, idDgst digest.
 	defer store.mu.RUnlock()
 	secret, ok := store.secrets[idDgst]
 	if !ok {
-		return nil, fmt.Errorf("secret %s: %w", idDgst, secrets.ErrNotFound)
+		// It seems like when setting secret using SetSecret, it is stored in the secretStore as Secret@xxh3:......,
+		// but with the new secrets api, the key does not have that prefix.
+		// TODO(rajatjindal): check with Justin/Andrea if we need to handle it differently
```
marcosnils, line 211: @rajatjindal just checking if this has been validated already :pray:

core/schema/container.go
rajatjindal: The comment here https://github.com/dagger/dagger/issues/7358#issuecomment-2321481240 talked about possibility of having a fallback. 

happy to go either route depending on how folks recommend.

core/secret.go
rajatjindal: no, this still need validation from @aluzzardi / @jedevc.

core/schema/container.go
jedevc: Yes there should be a fallback, but we shouldn't allow setting a secret with an empty string, where this would fallback to reading from the secret store.

PR #10100, opened 2025-04-08
feat: add default value comments for go

### LINKED ISSUE #10084: ✨ Show the default value of optional arguments in generated code

Default values can be specified for optional arguments, e.g. 
```golang
// Say hello to the world!
func (hello *Hello) Hello(ctx context.Context,
	// Change the greeting
	// +optional
	// +default="hello"
	greeting string,
	// Change the name
	// +optional
	// +default="world"
	name string,
```

This is visible in the help of the cli:
Image

However, the generated code only contains the description and not the default. See the Opts struct generated in Golang:
```golang
// HelloHelloOpts contains options for Hello.Hello
type HelloHelloOpts struct {
	//
	// Change the greeting
	//
	Greeting string
	//
	// Change the name
	//
	Name string
	//
	// Encode the message in giant multi-character letters
	//
	Giant bool
	//
	// Make the message uppercase, and add more exclamation points
	//
	Shout bool
	//
	// Custom container for running the figlet tool
	//
	FigletContainer *Container
}
```

Here is the module I used for the example: https://github.com/shykes/daggerverse/blob/main/hello/main.go

When consuming a module through code, it is still important to know the default value to decide whether to overwrite it.

This leads people to add the default value inside the description, which will eventually lead to mismatched values, e.g:
```golang
	// Target operating system defaults to "linux"
	// +optional
	// +default="linux"
	os string,
	// Target architecture, defaults to "amd64"
	// +optional
	// +default="amd64"
	arch string,
```

jedevc:
Fixes #10084.

Since we do special opts structs for go, this only applies for this special case - we don't need it in typescript or python.

### DIFF
--- a/sdk/go/dagger.gen.go
+++ b/sdk/go/dagger.gen.go
@@ -652,6 +652,8 @@ type ContainerAsTarballOpts struct {
 	// Use the specified media types for the image's layers.
 	//
 	// Defaults to OCI, which is largely compatible with most recent container runtimes, but Docker may be needed for older runtimes without OCI support.
+	//
+	// Default: OCIMediaTypes
 	MediaTypes ImageMediaTypes
 }
 
@@ -681,6 +683,8 @@ func (r *Container) AsTarball(opts ...ContainerAsTarballOpts) *File {
 // ContainerBuildOpts contains options for Container.Build
 type ContainerBuildOpts struct {
 	// Path to the Dockerfile to use.
+	//
+	// Default: "Dockerfile"
 	Dockerfile string
 	// Target build stage to build.
 	Target string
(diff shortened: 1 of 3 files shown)


sdk/go/dagger.gen.go
```diff
@@ -2859,7 +2859,7 @@ func (r *Directory) WithFiles(path string, sources []*File, opts ...DirectoryWit
 // DirectoryWithNewDirectoryOpts contains options for Directory.WithNewDirectory
 type DirectoryWithNewDirectoryOpts struct {
 	// Permission granted to the created directory (e.g., 0777).
-	Permissions int
+	Permissions int `default:"420"`
```
jedevc, line 2862: :sob: just realized the weirdness of this... not too much of an issue, since we have the same problem in our graphql schema, and in our other sdks.

But would be cool to maybe attach *more* directives that could act as hints to display this as octal/hex/etc.

vito: Hah yeah unfortunate. These octal numbers are annoying to deal with but felt like the most intuitive DX for these APIs. I remember we brielfy considered something like "rw-rw----" but that's just not what people usually interface with in APIs like this, and anything too declarative would probably stray too far from common convention/memory.

vito (approved):

PR #10106, opened 2025-04-08
MCP: mask some functions in core types, to avoid tool bloat

shykes:


grouville (approved):
LGTM -- will wait for other's perspective on that / impact on evals 🙏

core/mcp.go
```diff
@@ -194,6 +194,91 @@ func (m *MCP) tools(srv *dagql.Server, typeName string) ([]LLMTool, error) {
 			// never a reason to call "sync" since we call it automatically
 			continue
 		}
+		// Hide functions from the largest and most commonly used core types,
```
cwlbraa, line 197: ... so this is a allowlist? hard to tell from the diff what exactly is being hidden.

shykes: No this is a blocklist. The explicit list is of functions to hide

core/mcp.go
```diff
@@ -194,6 +194,109 @@ func (m *MCP) tools(srv *dagql.Server, typeName string) ([]LLMTool, error) {
...
+			case
+				"currentModule",
+				"currentTypeDefs",
+				"defaultPlatform",
+				"engine",
+				"env",
+				"error",
+				"function",
+				"generatedCode",
+				"llm",
```
vito, line 211: aww :P ...but probably fair

core/mcp.go
```diff
@@ -194,6 +194,109 @@ func (m *MCP) tools(srv *dagql.Server, typeName string) ([]LLMTool, error) {
...
+				"mounts",
+				"pipeline",
+				"platform",
+				"rootfs",
+				"terminal",
+				"up",
+				"user",
+				"withAnnotation",
+				"withDefaultTerminalCmd",
+				"withFile",
```
vito, line 247: nit: might as well keep this for the same reasoning as keeping `withDirectory` - it's common to encounter a `File`, and having to go through too many hops to turn it into `Directory` is less likely to work (let alone whether the model predicts that `withDirectory` copies additively)

core/mcp.go
```diff
@@ -203,8 +306,14 @@ func (m *MCP) tools(srv *dagql.Server, typeName string) ([]LLMTool, error) {
 		if err != nil {
 			return nil, fmt.Errorf("field %q: %w", field.Name, err)
 		}
+		var toolName string
+		if typeName == "Query" {
```
vito, line 310: :+1:

vito (approved):
lgtm, just one nitpick but not blocking

core/mcp.go
```diff
@@ -194,6 +194,91 @@ func (m *MCP) tools(srv *dagql.Server, typeName string) ([]LLMTool, error) {
...
+				"exposedPorts",
+				"file",
+				"imageRef",
+				"import",
+				"label",
+				"labels",
+				"mounts",
+				"pipeline",
+				"platform",
+				"terminal",
```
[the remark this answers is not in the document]
shykes: Yeah exactly

core/mcp.go
shykes: Yeah seems safer to remove it for now :)

core/mcp.go
shykes: Done 👍

core/mcp.go
jedevc: @shykes would you be opposed (either here or a follow-up) to having this as part of the schema declaration? So we'd do `dagql.Func(...).Doc(...).ExposeToLLM()` or something (name doesn't really matter to me).

Otherwise, it feels like it's easy for the schema to drift from this list over time, which would be nice to avoid.

core/schema/container.go
```diff
@@ -60,11 +58,11 @@ func (s *containerSchema) Install() {
...
 		dagql.NodeFunc("from", s.from).
-			Doc(`Initializes this container from a pulled base image.`).
+			Doc(`Download a container image, and apply it to the container state. All previous state will be lost.`).
 			ArgDoc("address",
-				`Image's address from its registry.`,
-				`Formatted as [host]/[user]/[repo]:[tag] (e.g., "docker.io/dagger/dagger:main").`),
-
+				`Address of the container image to download, in standard OCI ref format. Example:"registry.dagger.io/engine:latest"`,
+			),
+		// FIXME: deprecate
```
jedevc, line 65: :eyes: sorry, might have missed the context here. I do agree it's weird to have two ways to build a docker container, but I'm actually sure which of the two we'd deprecate.

core/schema/container.go
```diff
@@ -184,26 +180,26 @@ func (s *containerSchema) Install() {
...
 
 		dagql.Func("entrypoint", s.entrypoint).
-			Doc(`Retrieves entrypoint to be prepended to the arguments of all commands.`),
+			Doc(`Return the container's OCI entrypoint.`),
 
 		dagql.Func("withEntrypoint", s.withEntrypoint).
-			Doc(`Retrieves this container but with a different command entrypoint.`).
-			ArgDoc("args", `Entrypoint to use for future executions (e.g., ["go", "run"]).`).
-			ArgDoc("keepDefaultArgs", `Don't remove the default arguments when setting the entrypoint.`),
+			Doc(`Set an OCI-style entrypoint. It will be included in the container's OCI configuration. Note, \"withExec\" ignores the entrypoint by default.`).
```
jedevc, line 186:
```suggestion
			Doc(`Set an OCI-style entrypoint. It will be included in the container's OCI configuration. Note, "withExec" ignores the entrypoint by default.`).
```

The escapes aren't needed, this is a raw string (as is, this will actually insert the `\` into the generated docs).

core/schema/container.go
```diff
@@ -484,13 +470,13 @@ func (s *containerSchema) Install() {
...
-			Doc(`The output stream of the last executed command.`,
-				`Returns an error if no command was set.`),
+			Doc(`The buffered standard output stream of the last executed command`,
+				`Returns an error if no command was executed`),
 
 		dagql.Func("stderr", s.stderr).
 			View(AllVersion).
-			Doc(`The error stream of the last executed command.`,
-				`Returns an error if no command was set.`),
+			Doc(`The buffered standard error stream of the last executed command`,
```
jedevc, line 478: mini nit: this isn't really a stream, it's the entire output buffer.

shykes: Yeah that's why it says "buffered".

core/schema/container.go
shykes: There was a decision to deprecate this one a long time ago, I was surprised to find that it wasn't actually deprecated. Got lost in the lore I guess.

jpadams:
@shykes this exercise made me think about hiding `from` and having a `container` interface similar to `git` and `http` to make it easier on LLMs.

Though the interfaces are all a little different... :/
Some give you exactly what you ask for. Some return a different core type. Some give you a special complex type.

```
⋈  container alpine # would return Container, though we need --platform too

⋈  http https://dagger.io # returns File, not an Http or Uri object

⋈  git https://github.com/dagger/dagger #  returns GitRepository which is hardly ever used usually
⋈  git https://github.com/dagger/dagger | head | tree # returns Directory
```

core/schema/container.go
vito: Between the two I agree that this is the one to deprecate - motivated primarily by an LLM's tendency to see it as the primary way to follow a "build" instruction. With this gone it'll be more likely to go with `go build` or whatever is appropriate.

The `Directory` API is called `dockerBuild` which is much more explicit.

jedevc: One thing I *did* technically realize while doing https://github.com/dagger/dagger/pull/9980.

I guess it was possible for you to attach a service binding before calling `.Build`, and then access that service during the build? But also, we had no tests, I've never seen that pattern in the wild, so I actually just removed it there.

PR #10113, opened 2025-04-09
fix: correctly escape special characters for elixir docstrings

jedevc:
Discovered in https://github.com/dagger/dagger/pull/10106#discussion_r2035065993

It was possible to generate bad elixir code if a function had a doc string that included quotes.

### DIFF
--- a/sdk/elixir/dagger_codegen/lib/dagger/codegen/elixir_generator/renderer.ex
+++ b/sdk/elixir/dagger_codegen/lib/dagger/codegen/elixir_generator/renderer.ex
@@ -42,7 +42,13 @@ defmodule Dagger.Codegen.ElixirGenerator.Renderer do
         "\"\"\""
       ]
     else
-      [?", s, ?"]
+      [
+        ?",
+        s
+        |> String.replace("\\", "\\\\")
+        |> String.replace("\"", "\\\""),
+        ?",
+      ]
     end
   end
 


grouville (approved):
Thank you, I didn't know where to do it 🙏

sdk/elixir/dagger_codegen/lib/dagger/codegen/elixir_generator/renderer.ex
```diff
@@ -42,7 +42,13 @@ defmodule Dagger.Codegen.ElixirGenerator.Renderer do
...
       ]
     else
-      [?", s, ?"]
+      [
+        ?",
+        s
+        |> String.replace("\\", "\\\\")
+        |> String.replace("\"", "\\\""),
+        ?",
+      ]
```
wingyplus, line 51: We need to escape it on triple quote as well. Example:

Give the Elixir module:
```elixir
# lib/hello.ex
defmodule Hello do
  @doc """
  \abcd
  """
  def hello(), do: "world"
end
```

Then start a shell with `iex -S mix` and call `h Hello.hello` function, the shell will print `\a` as an unexpected character:

```
iex(1)> h Hello.hello

                                  def hello()                                   

^'bcd

wingyplus: I don't know how to push to this PR so I open a new [PR](https://github.com/dagger/dagger/pull/10136) that superseded this to fix triple quote docstrings. 🙏

PR #9953, opened 2025-03-24
ci: Make testdev-everything-else & testdev-module-runtimes smaller

gerhard:

PR #10117, opened 2025-04-09
ci: Split all module-runtimes

gerhard:
As this is our second slowest test suite - `13m 20s` P90 for the last 2 weeks - and also most likely to flake, we (@sipsma  @jedevc) were curious to see how this behaves when everything is split.

FWIW, this provided some surprising results, so we want to dig a little bit more into the behaviour of this particular split:
- https://github.com/dagger/dagger/pull/9953

Builds on top of:
- https://github.com/dagger/dagger/pull/10086

---
We **may** not want to merge this, the primary reason for this PR is to inform.

### DIFF
--- a/.github/main.go
+++ b/.github/main.go
@@ -182,25 +182,25 @@ func (ci *CI) withTestWorkflows(runner *dagger.Gha, name string) *CI {
 				Runner: []string{AltGoldRunner()},
 			}},
 			{"modules", []string{"TestModule"}, &dagger.GhaJobOpts{
-				Runner: []string{AltGoldRunner()},
+				Runner: []string{GoldRunner(true)},
 			}},
 			{"module-runtimes", []string{"TestGo", "TestPython", "TestTypescript", "TestElixir", "TestPHP", "TestJava"}, &dagger.GhaJobOpts{
-				Runner: []string{AltPlatinumRunner()},
+				Runner: []string{PlatinumRunner(true)},
 			}},
 			{"container", []string{"TestContainer"}, &dagger.GhaJobOpts{
-				Runner: []string{AltGoldRunner()},
+				Runner: []string{GoldRunner(true)},
 			}},
 			{"LLM", []string{"TestLLM"}, &dagger.GhaJobOpts{
-				Runner: []string{AltGoldRunner()},
+				Runner: []string{GoldRunner(true)},
 			}},
 			{"cli-engine", []string{"TestCLI", "TestEngine"}, &dagger.GhaJobOpts{
-				Runner: []string{AltGoldRunner()},
+				Runner: []string{GoldRunner(true)},
 			}},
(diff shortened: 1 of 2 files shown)


gerhard:
Before this change, `testdev-module-runtimes` would use 32 vCPUs take `14m 34s`: https://github.com/dagger/dagger/actions/runs/14386987464/job/40344749251 & https://v3.dagger.cloud/dagger/traces/b14386b31458d3b0f7247d481beb2573

image

After this change, even though we run 6 x 16 vCPUs, with each runtime getting its own isolated VM, we have the following:
| job | duration |
| :- | - |
| testdev-elixir-module-runtime | `6m 57s` |
| testdev-go-module-runtime | `7m 1s` |
| testdev-java-module-runtime | `7m 32s` |
| testdev-php-module-runtime | `6m 52s` |
| testdev-python-module-runtime | `8m 31s` |
| testdev-typescript-module-runtime | `9m 59s` |

While we know that building the dev Engine adds about `3m` of extra time to each job, it was surprising to see a slow-down across every single test suite when running them individually, in single tenant VMs, each with 16 vCPUs.

WDYT @sipsma @jedevc ?

sipsma:
> While we know that building the dev Engine adds about 3m of extra time to each job, it was surprising to see a slow-down across every single test suite when running them individually, in single tenant VMs, each with 16 vCPUs.

Looking at the runtimes of the spans for the tests themselves:

Recent run from main: https://v3.dagger.cloud/dagger/traces/b32ae8e35bded919ed8607cecdca211e
* `TestGo` - 4m24s
* `TestPython` - 6m3s
* `TestTypescript` - 4m22s

This PR:
* `TestGo` - 1m0s - https://v3.dagger.cloud/dagger/traces/489f10f25a2573c45e2b3c8e1ddc19a7
* `TestPython` - 2m25s - https://v3.dagger.cloud/dagger/traces/6580f7d646cc4dcc421314387089bdc3
* `TestTypescript` - 3m35s - https://v3.dagger.cloud/dagger/traces/f54c775adc8c2ac87d9752518167383b

So they all seem to be sped up by isolating them (as expected).

The reason the sum total of the split up tests is much worse is probably because not only is there the 3m of building the dev engine, there's also the time spent building the nested engine as a service that runs in the dev engine (which shows up in the spans as time before the actual `TestGo`/`TestPython`/`TestTypescript` spans).

---

In terms of other things to try:
1. Could just keep the tests combined but give even more CPU (i.e. 64). Trying that would at least help confirm whether or not the tests are CPU bound at this point.
2. The nested dev engine service theoretically isn't needed for these particular tests, they could instead just use a standard nested exec instead. The nested dev engine service is still needed in the general case (i.e. running tests locally against stable engine, other test suites that really require an independent engine, etc.), but *these* sdk runtime tests don't need that afaik. If we got rid of it for just these tests then we'd skip a big chunk of overhead.

Issue #10116, opened 2025-04-09
🐞 shell mode doesn't show env API's child module extensions


cwlbraa:
### What is the issue?

The Env api should show installed dependencies as part of its WithXInput/WithXOutput APIs.

Curiously, after loading a module, calling `.refresh` will cause the env API extensions for dependencies to appear. This bug also _doesn't_ occur when loading a module via cd, like `dagger -c ".cd ..; env | .help"` will show all the env extensions of ..'s installed modules.

### Dagger version

dagger v0.18.2

### Steps to reproduce

for example, DaggerDev, `dagger -c "env | .help"` should show `with-wolfi-input` and `with-wolfi-output` in the list of available functions of env. it does not.

however, `dagger -c ".refresh; env | .help"` shows the `with-wolfi-X` functions, as does `cd modules/wolfi && dagger -c ".cd ../.. env | .help"`

cwlbraa:
I've tried several approaches to fixing this, including explicitly loading typedefs and inspecting dependency modules and saving served dep modules via h.getOrInitDef. None of these seem to have any effect. 

The fact `.refresh` works implies to me there's some ordering issue here, like the Serve call we're making happens too late in the module load to propogate env API changes to the client, and then calling `.refresh` causes those changes to propogate after the fact.

cwlbraa:
> For example, this shouldn’t work:
>
> cd modules/alpine && dagger -c ".cd ../.. && env | .help"

AHHHH you're correct, this does in fact not work, and also dispels the mystery for me, thank you!

>  To solve this in the CLI you could add an option to initializeModule (or do it always) for serving dependencies and do it between the module’s Serve and inspectModule

ah, interesting, like you'd need to serve the dependencies and then inspectModule on their parent? i'd tried inspectModuling the deps and didn't get anywhere

PR #10118, opened 2025-04-09
move shell/prompt dependency serving server-side

cwlbraa:
fixes https://github.com/dagger/dagger/issues/10116, replacing https://github.com/dagger/dagger/pull/9992

the main meat of this PR adds 
```golang
module.Serve(ctx, dagger.ModuleServeOpts{IncludeDependencies: true})
```

to the API, enabling shell and other clients to serve the dependencies of modules to emulate "running inside SDK code". previously, on #9992, we were doing this client side, but that led to #10116 where Env wouldn't expose module input/outputs until after users called `.refresh`.

this PR also silently and automatically pre-populates `$agent` on shell startup. this is super useful for both automated testing and human inspection of the default LLM context.

this PR raised hard questions about [modules being able to replace core API functions](https://github.com/dagger/dagger/issues/10245). this PR does not resolve those issues, but instead updates the shell helpers to reflect reality - when "git" has been replaced, for example, `.stdlib | .help git` will throw a "does not exist" error.


helderco:
> I still don't understand exactly why the client-side implementation was broken, but with this change `cd ~/src/dagger; dagger -c ".refresh; env | .help"` shows the desired `with-wolfi-` functions.

I added an explanation in:
- https://github.com/dagger/dagger/issues/10116#issuecomment-2792654465

core/schema/module.go
```diff
@@ -85,7 +85,8 @@ func (s *moduleSchema) Install() {
 		dagql.NodeFunc("serve", s.moduleServe).
 			DoNotCache(`Mutates the calling session's global schema.`).
 			Doc(`Serve a module's API in the current session.`,
-				`Note: this can only be called once per session. In the future, it could return a stream or service to remove the side effect.`),
+				`Note: this can only be called once per session. In the future, it could return a stream or service to remove the side effect.`).
+			ArgDoc("serveDependencies", "expose the dependencies of this module to the client"),
```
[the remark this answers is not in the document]
cwlbraa: includeDependencies works for me, equally clear and less redundant.

cwlbraa:
added a couple tests and pre-populated $agent for ease-of-testing.

helderco:
Need to figure out what to do if a dependency has the same name as a core root function (e.g., git). Should probably return a naming conflict error. Not just for serving dependencies, should be the same within a module, otherwise you can't get core's `dag.Git()`.

That also means we can't have a github.com/dagger/dagger/version module anymore (have to rename it).

core/schema/module.go
```diff
@@ -86,7 +86,8 @@ func (s *moduleSchema) Install() {
 		dagql.NodeFunc("serve", s.moduleServe).
 			DoNotCache(`Mutates the calling session's global schema.`).
 			Doc(`Serve a module's API in the current session.`,
-				`Note: this can only be called once per session. In the future, it could return a stream or service to remove the side effect.`),
+				`Note: this can only be called once per session. In the future, it could return a stream or service to remove the side effect.`).
+			ArgDoc("includeDependencies", "expose the dependencies of this module to the client"),
```
helderco, line 90:
```suggestion
			ArgDoc("includeDependencies", "Expose the dependencies of this module to the client"),
```

cmd/dagger/shell.go
```diff
@@ -215,6 +215,10 @@ func (h *shellCallHandler) Initialize(ctx context.Context) error {
 		return fmt.Errorf("initial context: %w", err)
 	}
 
+	// silently attempt to initialize llm to populate $agent.
+	// swallow errors, we'll try again and error if the user attempts to use it.
+	_, _ = h.llm(ctx)
```
helderco, line 220: A "no valid LLM endpoint configuration" error will still show in cloud/TUI. Best hide it with a span.

cwlbraa: ah yeah, i had noticed this when i first made the change and never got around to trying to hide it. will fix!

helderco: It shows with `-vv`:

Screenshot 2025-04-27 at 15 42 45

Is it really needed? Is it only for interactive mode or non-interactive as well? What's the use case?

cmd/dagger/module_inspect.go
```diff
@@ -426,11 +417,6 @@ func (m *moduleDef) GetObjectFunction(objectName, functionName string) (*modFunc
 }
 
 func (m *moduleDef) GetFunction(fp functionProvider, functionName string) (*modFunction, error) {
-	// This avoids an issue with module constructors overriding core functions.
-	// See https://github.com/dagger/dagger/issues/9122
-	if m.HasModule() && fp.ProviderName() == "Query" && m.MainObject.AsObject.Constructor.CmdName() == functionName {
-		return m.MainObject.AsObject.Constructor, nil
-	}
```
cwlbraa, line 429: @helderco was this not working as described in the comment?

helderco: At the time the "Query" object didn't include the module's constructor, it was only core, before loading the module. I guess that was fixed when Erik refactored module loading. Now, I just call `loadTypeDefs` on every function under `Query` when the module is inspected, which includes the module's constructor so this is no longer needed.

cmd/dagger/shell.go
cwlbraa: it's for both - lets you do `dagger -c '$agent | tools'` under test or interactive mode to inspect your prompt-mode LLM's "environment" setup. 

i use it to test the actual banner feature of this PR: the prompt-mode llm can call dependency module constructors. 

the logging, however, is deeply unnecessary. maybe i should just completely black-hole it with `context.Background()` 🤔

cwlbraa: fixed, i think, by fully backgrounding.

cmd/dagger/module.go
```diff
@@ -877,7 +877,7 @@ func optionalModCmdWrapper(
 			switch {
 			case configExists:
 				mod := modSrc.AsModule()
-				err := mod.Serve(ctx)
+				err := mod.Serve(ctx, dagger.ModuleServeOpts{IncludeDependencies: true})
```
helderco, line 880: This makes `dagger listen` and `dagger query` behave the same by serving dependencies too, so it's more consistent. It's also helpful to debug what the shell sees.

helderco (approved):
LGTM, just needs another rebase and regen.

PR #10122, opened 2025-04-09
for prompt mode: selectroot and module scoping

cwlbraa:
this PR adds a selectRoot tool and the ability for Environments to specify an initial selection for an LLM. The idea here is that the default behavior for the dagger shell should be to drop the llm into the module, but also give the llm the ability to back out to the core API. 

this is currently broken because i'm trying to pass a module to dagql to construct an environment, but trying to get a module in the shell context via dag.CurrentModule() chokes- presumably that function is intended to run in an SDK module container?

### DIFF
--- a/cmd/dagger/llm.go
+++ b/cmd/dagger/llm.go
@@ -97,6 +97,7 @@ func (s *LLMSession) reset() {
 	s.llm = s.dag.LLM(dagger.LLMOpts{Model: s.model}).
 		WithEnv(s.dag.Env(dagger.EnvOpts{
 			Privileged: true,
+			Module:     s.dag.CurrentModule().Source().AsModule(),
 		}))
 }
 
--- a/cmd/dagger/shell_commands.go
+++ b/cmd/dagger/shell_commands.go
@@ -240,6 +240,7 @@ func (h *shellCallHandler) llmBuiltins() []*ShellCommand {
 				if h.llmSession == nil {
 					return fmt.Errorf("LLM not initialized")
 				}
+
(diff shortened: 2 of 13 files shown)


cmd/dagger/llm.go
```diff
@@ -97,6 +97,7 @@ func (s *LLMSession) reset() {
 	s.llm = s.dag.LLM(dagger.LLMOpts{Model: s.model}).
 		WithEnv(s.dag.Env(dagger.EnvOpts{
 			Privileged: true,
+			Module:     s.dag.CurrentModule().Source().AsModule(),
```
cwlbraa, line 100: this is clearly the wrong approach for passing the module as initialSelection, but i'm not sure what'd be better... CurrentModule seems to not be set in the shell case, and it's pretty weird going CurrentModule.Source.AsModule anyways.

can we figure out the client's -m loaded module from the server side and just pass a bool here? Maybe we should just be passing a generic Object?

cwlbraa:
this approach ends up being fraught for multiple reasons:
1. the api doesn't like generic objects or instantiated-module-objects being passed to the core API without types, so we can't put this on env() easily.
2. WithFooInput(select) sorta worked, but hit a bug deep in the codegen where it's default bool status broke old-module-new-engine backwards compatibility. it also exposed LLM selections to the API more than we'd like to - they're really an implementation detail that we want to concern the LLMs with, not the users.


most importantly, though, this PR is superseded by @vito's [selectTools tool calling scheme](https://github.com/dagger/dagger/pull/10134) that foregoes object selections entirely.

PR #10124, opened 2025-04-10
fix: handle empty argument case in shell completion

kasattejaswi:
This is my first PR, please let me know if I am missing something. 

This PR addresses bug where the shell was crashing due to a slice bounds out of range [1:0] error in `dispatchCall`, caused by accessing `args[0]` when `call.Args` is empty.


jedevc:
:wave: hey @kasattejaswi was there a reason you closed this? it looks like the PR did have some wrong commits in it, but you should be able to `git rebase -i`/etc to get the right ones, instead of needing to close + reopen.

kasattejaswi:
Hey @jedevc I actually messed up with the signing thing. I read that documentation after committing the code. So I thought, it would be cleaner to do things again. Will keep in mind in future.

jedevc:
For signing, you can actually just do `git commit --amend -s` to rewrite the last commit with the sign-off :tada:

Generally, it's fine to do lots of rewriting git history in a PR, as long as the final diff + commits look relatively clean! :smile:

Issue #10107, opened 2025-04-08
🐞 Panic bug in tab completion in `dagger shell` - at least with `with-new-directory`


jpadams:
### What is the issue?

Was using TAB completion in Dagger Shell and got panic when tabbing.

### Dagger version

0.18.2

### Steps to reproduce

`dagger -M`

`⋈  dir=$(directory | with-new-directory  `<hit `TAB`>

### Log output

```
Caught panic:

runtime error: slice bounds out of range [1:0]

Restoring terminal...

goroutine 1 [running]:
runtime/debug.Stack()
	/usr/lib/go/src/runtime/debug/stack.go:26 +0x64
runtime/debug.PrintStack()
	/usr/lib/go/src/runtime/debug/stack.go:18 +0x1c
github.com/charmbracelet/bubbletea.(*Program).recoverFromPanic(0x140004463c0)
	/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.3.4/tea.go:758 +0x84
panic({0x1064b01c0?, 0x14000eb4570?})
	/usr/lib/go/src/runtime/panic.go:785 +0x124
main.(*shellAutoComplete).dispatchCall(0x48?, 0x14000a23360, 0x14000a22f00, 0x25)
	/app/cmd/dagger/shell_completion.go:159 +0x1ac
main.(*shellAutoComplete).dispatch(0x1065bb760?, 0x14000359d40?, 0x140003072d8?, 0x0?)
	/app/cmd/dagger/shell_completion.go:142 +0x50
main.(*shellAutoComplete).Do(0x14000194b60, {0x140006c8a88?, 0x124?, 0x14000abf040?}, 0x14000e5bb18?, 0x25)
	/app/cmd/dagger/shell_completion.go:107 +0x2c0
main.(*shellCallHandler).AutoComplete(0x140002b0a80, {0x140006c8a88, 0x1, 0x63}, 0x0, 0x25)
	/app/cmd/dagger/shell.go:443 +0x150
github.com/vito/bubbline/editline.(*Model).autoComplete(0x140008fc000)
	/go/pkg/mod/github.com/vito/bubbline@v0.0.0-20250312195236-5f4f49d6ebcb/editline/editline.go:617 +0x9c
github.com/vito/bubbline/editline.(*Model).Update(0x140008fc000, {0x1064908a0, 0x140000b7290})
[...]

PR #10125, opened 2025-04-10
fix: handle empty argument case in shell completion

kasattejaswi:
Fixes https://github.com/dagger/dagger/issues/10107.

This is my first PR, please let me know if I am missing something. 

This PR addresses bug where the shell was crashing due to a slice bounds out of range [1:0] error in `dispatchCall`, caused by accessing `args[0]` when `call.Args` is empty.

### DIFF
--- a/cmd/dagger/shell_completion.go
+++ b/cmd/dagger/shell_completion.go
@@ -156,6 +156,9 @@ func (h *shellAutoComplete) dispatchCall(previous *CompletionContext, call *synt
 	for _, arg := range call.Args {
 		args = append(args, arg.Lit())
 	}
+	if len(args) == 0 {
+		return previous
+	}
 	return previous.lookupField(args[0], args[1:])
 }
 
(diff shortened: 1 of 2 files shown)


jedevc:
Hm, so this indeed fixes the panic, but I think that this should actually not return *any* completions.

image

None of the suggested completions is actually valid here.

But also, this happens here as well:

image

So I think all of the logic actually needs updating - but this is a good start to actually fix the panic.

I pushed a test for this (and the above behavior) so at least we'll avoid regressing on the panic - maybe a nice follow-up would be to fix the completion so that we don't get the misleading completions after `container <tab>`?

kasattejaswi:
Sure @jedevc, looking into it.

jedevc:
I've merged this as-is, since it's a good self-contained fix! Feel free to follow-up with another PR if you make any progress on the completion correctness bit :yellow_heart:

PR #10131, opened 2025-04-10
bump runners to v0.18.2

vito:
Looks like this was missed in the past couple of ships


jedevc (approved):

vito:
Noticed the TS/Python checks were legit failing due to formatting lints, so patched those up too.

Issue #9582, opened 2025-02-14
Generated clients


shykes:
## Overview

Dagger should make it very easy to generate clients for a Dagger module, and use these clients as part of an existing development workflow. The same generated clients should be usable whether developing Dagger Functions, or an external program.

Examples of generated clients:
- Bindings for Go, Python, Typescript... Same as those already generated by Dagger SDKs, but decoupled from server-side API extensions (custom types and functions)
- CI configurations (eg. reusable Github Actions, CircleCI orbs, Jenkins groovy scripts...)
- Framework-specific bindings (React component? LangGraph middleware? MCP server?)

#### Install a client

```console
dagger client install GENERATOR [PATH] [-n|--no-generate]
```

Install a generated client for the current module, using the specified GENERATOR, and write at the specified path.
The generator configuration is saved to `dagger.json`.

- If PATH is omitted, a generator-specific default is chosen
- PATH must be a directory. If the directory doesn't exist, it will be created.
- If files are already present at the path, and are not known to the generator, they must not be overwritten. Conflicts will result in an error without data loss.

By default, the newly added client is also generated. This can be disabled with `-n` or `--no-generate`.

#### Uninstall a client

```console
dagger client uninstall PATH
```

Uninstall the generated client at the given path. This will remove the client from `dagger.json`, and also remove the generated files, *without removing pre-existing files*.

#### List clients

```console
dagger client list [-q]
```

List all clients for the current module.

By default it will print the path and generator in two columns. With `-q` it only prints the paths.

#### Generate clients

```console
dagger client generate [PATH...]
```

Generate (or re-generate) the specified clients, or all clients by default.

#### Update a client generator

```console
dagger client update [PATH...]
```

Update the generators for the specified clients, or for all clients by default. This will update each generator to its latest available version, following the same dependency management conventions as `dagger install` for regular Dagger dependencies.

#### Configure a client

```console
dagger client config PATH [KEY=VALUE...]
```

Update the generator configuration for the specified client. Each generator has its own configuration schema.

If no configuration value is specified, the current configuration is printed.

#### Print available client configuration options

*FIXME*

TomChv:
> Install a generated client for the current module, using the specified GENERATOR, and write at the specified path.
> The generator configuration is saved to dagger.json.

I like this proposal, so just to confirm, in the `dagger.json`
We would have:

```json
{
  "clients": [
    {
[...]

PR #10133, opened 2025-04-10
feat: add client CLI capabilities

TomChv:
Related to https://github.com/dagger/dagger/issues/9582

**Summary of changes**
- Rename `add` to `install` for consistency.
- Add `list` with optional `--json` option
- Add uninstall
- New integration tests for CLI commands
- New `WithoutClient` method on `ModuleSource`

There are 2 other commands describe in the issue that I did not implemented:
- `update`: Not sure how to implement it properly yet, for now existing generators are bundled inside the engine as SDK methods so it's not a priority
- `config [set] [get]`: This is not something we actually support in `dagger config` so I think it's not a priority either, updating the file manually is fine enough for now.


cmd/dagger/client.go
```diff
@@ -14,3 +43,183 @@ var clientCmd = &cobra.Command{
 		"experimental": "true",
 	},
 }
+
+var clientInstallCmd = &cobra.Command{
+	Use:     "install [options] [path]",
```
jedevc, line 48: We might also want to alias to `use` as well, since that's aliased at the top-level as well.

cmd/dagger/client.go
```diff
@@ -14,3 +43,183 @@ var clientCmd = &cobra.Command{
...
+				Source struct {
+					ConfigClients []struct {
+						Generator string
+						Directory string
+					}
+				}
+			}
+
+			err = dag.Do(ctx, &dagger.Request{
+				Query: loadModClientConfQuery,
```
jedevc, line 194: Why not just use the code-gened API directly?

jedevc (approved):
:tada: yup this is neat!

Left a couple notes, but LGTM!

cmd/dagger/client.go
TomChv: Because it's complex with the codegened API to get nested fields in an array so I used that "hack" to make it cleaner on the code side :)

PR #10134, opened 2025-04-10
llm: switch to `select_tools` scheme

### LINKED ISSUE #10039: 🐞 llm.withModel doesn't update provider-specific MCP configuration
### What is the issue?

When changing model of an existing LLM instance, with `llm.withModel`, the underlying MCP implementation is not reconfigured to adapt to the provider. As of 0.18.0 that means the "needs system prompt" flag is not changed.

### Dagger version

dagger v0.18.0 (docker-image://registry.dagger.io/engine:v0.18.0) darwin/arm64

### Steps to reproduce

```console
dagger -c 'llm --model gemini | with-model gpt-4o'
```

### Log output

_No response_

vito:
## Evals comparison with `main`

Bottom line up front: with these changes, all evals now perform much more consistently across all major providers.

There is a higher baseline token cost now that we apply a system prompt across all providers (instead of just Gemini) - but scenarios that take multiple turns and chain tools, the token usage actually drops sharply, owing to the new `select_tools` mechanics.

| Model | Eval | Success Rate | Input / Output Tokens |
|-------|------|-------------|---------------------|
| `claude-3-5-sonnet-latest` | Basic | 100% <br />(30 attempts) | 23.0 → 2267.2 (+2244.2) <br />4.0 → 98.2 (+94.2) |
| `claude-3-5-sonnet-latest` | BuildMulti | 23% → 97% (+74%) <br />(30 attempts) | 9346.6 → 6101.0 (-3245.6) <br />648.0 → 863.9 (+215.9) |
| `claude-3-5-sonnet-latest` | BuildMultiNoVar | 90% → 100% (+10%) <br />(30 attempts) | 8685.7 → 5155.5 (-3530.2) <br />695.3 → 911.5 (+216.1) |
| `claude-3-5-sonnet-latest` | ReadImplicitVars | 90% → 100% (+10%) <br />(30 attempts) | 1696.9 → 5425.3 (+3728.4) <br />159.0 → 406.9 (+247.8) |
| `claude-3-5-sonnet-latest` | UndoChanges | 90% → 100% (+10%) <br />(30 attempts) | 8537.3 → 8916.9 (+379.6) <br />890.1 → 962.1 (+72.0) |
| `claude-3-5-sonnet-latest` | WorkspacePattern | 100% → 97% (-3%) <br />(30 attempts) | 13506.7 → 9053.6 (-4453.1) <br />702.9 → 693.8 (-9.1) |
| `gemini-2.0-flash` | Basic | 100% <br />(100 attempts) | 168.0 → 680.0 (+512.0) <br />2.0 |
| `gemini-2.0-flash` | BuildMulti | 10% → 98% (+88%) <br />(100 attempts) | 19106.8 → 18283.8 (-823.0) <br />81.1 → 378.2 (+297.1) |
| `gemini-2.0-flash` | BuildMultiNoVar | 55% → 93% (+38%) <br />(100 attempts) | 29572.5 → 19626.3 (-9946.2) <br />164.9 → 470.8 (+305.9) |
| `gemini-2.0-flash` | ReadImplicitVars | 10% → 99% (+89%) <br />(100 attempts) | 357.0 → 4196.9 (+3839.9) <br />28.2 → 180.8 (+152.6) |
| `gemini-2.0-flash` | UndoChanges | 12% → 98% (+86%) <br />(100 attempts) | 16236.2 → 10279.9 (-5956.3) <br />171.2 → 329.6 (+158.4) |
| `gemini-2.0-flash` | WorkspacePattern | 96% → 83% (-13%) <br />(100 attempts) | 5503.9 → 7869.0 (+2365.1) <br />275.7 → 308.3 (+32.6) |
| `gpt-4.1` | Basic | 100% <br />(50 attempts) | 22.0 → 1443.3 (+1421.3) <br />3.0 → 48.3 (+45.3) |
| `gpt-4.1` | BuildMulti | 98% → 96% (-2%) <br />(50 attempts) | 27892.8 → 20401.2 (-7491.6) <br />256.4 → 221.5 (-34.9) |
| `gpt-4.1` | BuildMultiNoVar | 100% → 92% (-8%) <br />(50 attempts) | 29974.8 → 22331.0 (-7643.8) <br />240.9 → 225.0 (-15.9) |
| `gpt-4.1` | ReadImplicitVars | 100% <br />(50 attempts) | 6982.0 → 3320.0 (-3662.0) <br />127.3 → 68.0 (-59.3) |
| `gpt-4.1` | UndoChanges | 84% → 100% (+16%) <br />(50 attempts) | 14502.7 → 8069.0 (-6433.8) <br />225.7 → 152.8 (-72.9) |
| `gpt-4.1` | WorkspacePattern | 98% → 96% (-2%) <br />(50 attempts) | 3390.4 → 8777.2 (+5386.8) <br />191.1 → 159.2 (-31.9) |
| `gpt-4o` | Basic | 100% <br />(50 attempts) | 22.0 → 695.0 (+673.0) <br />3.4 → 4.0 (+0.6) |
| `gpt-4o` | BuildMulti | 90% → 100% (+10%) <br />(50 attempts) | 29212.3 → 18757.7 (-10454.5) <br />220.6 → 213.8 (-6.8) |
| `gpt-4o` | BuildMultiNoVar | 94% → 98% (+4%) <br />(50 attempts) | 33805.1 → 20317.0 (-13488.1) <br />226.5 → 290.5 (+64.0) |
| `gpt-4o` | ReadImplicitVars | 100% <br />(50 attempts) | 7230.8 → 5992.5 (-1238.3) <br />112.5 → 134.8 (+22.3) |
| `gpt-4o` | UndoChanges | 78% → 96% (+18%) <br />(50 attempts) | 17552.7 → 9078.4 (-8474.3) <br />232.7 → 164.3 (-68.5) |
| `gpt-4o` | WorkspacePattern | 86% → 100% (+14%) <br />(50 attempts) | 3641.8 → 9222.3 (+5580.4) <br />170.1 → 166.6 (-3.5) |

## Tools are now `snake_case`

Some UIs like Goose assume `snake_case` and our `Container_withFooBar` scheme was confusing case conversion libraries. So now those are `Container_with_foo_bar` and all the builtin tools have been renamed accordingly.

## Evals now live in the repo

```sh
# run all evals, sane defaults for attempt count per provider
dagger -m modules/evaluator call evals-across-models

# run one eval, analyze its results, and suggest a change to the system prompt
dagger -m modules/evaluator call evaluate --model model --name BuildMulti
```

## Retry logic for API requests

Brought this in too since otherwise we can't really trust CI/eval results.

Also, Gemini will retry on `FinishReason(10)` - afaict this is an internal Gemini issue where it generates a malformed function call. There's no further information (I tried wiretapping). Maybe this is fixed in 2.5.

## No more "current selection" - select _tools_ instead

The model now "goes shopping" - it selects tools for operating on the available objects, and can make calls against any object of any type.

Each `Foo_bar` tool now takes a `Foo:` argument, which defaults to the newest (highest numbered) `Foo`.

The single `select_tools` tool lists all available tools and objects in its description, to continuously reinforce the context to the model.

```
Select tools for interacting with the available objects.

Available tools:
- Container_withExec (returns Container!)
- ...

Available objects:
- Container#1: The container to use to build Booklit.
- Container#2: Container#1.withMountedDirectory(path: "/src", source: Directory#1)
- Container#3: Container#2.withExec(args: ["go", "build", "./cmd/booklit"])
- Directory#1: The Booklit source code repository.
```

## New system prompt, applied for all models

There is a new hand-written system prompt, now applied to all models instead of just Gemini. For the most part I followed the [GPT 4.1 prompting guide](https://cookbook.openai.com/examples/gpt4-1_prompting_guide).

I spent a long time trying two things:

1. Pushing the system prompt into a tool instead (#10059)
  * This just doesn't work well enough. It works okayish as long as it's listed last, but I ended up with multiple tools competing for the last spot. In the end, tool descriptions are just not weighted strongly enough for this to be viable.
2. Having an AI agent continuously iterate towards a minimum viable system prompt
  * This worked decently, turns out Claude 3.7 is pretty good at it. But in the end a hand-written prompt following the GPT 4.1 guidance performed much better.

We strived to avoid a system prompt for two reasons, but I think it's time to reconsider:

1. It would be nice if we didn't need one because the tool scheme is obvious enough
  * In hindsight this feels like a pipe dream. Some models work well, most don't. Local ones especially won't. At this point I would rather just consistently apply the same prompt everywhere instead of relying on out-of-the-box behavior that can change with each new model.
2. We can't control the system prompt with MCP
  * MCP has a specific API exactly for this purpose: the `instructions` field, returned on MCP server initialization. Some clients, like Goose, already respect it. I think we should just argue that anyone that doesn't respect it is wrong. It's more likely to be an oversight, since these are very early days.

## Rename `return` to `save`

Reasoning:

* Gemini would sometimes try to call `return_` instead of `return`. Maybe it's trying to avoid a common language keyword? :man_shrugging:
* Models were pretty keen to call `saveAs` when it briefly existed, so that's why I went with `save`.
* `return` sounds too final - I think it's nice to let the model save new versions of the output as the requirements change see `UndoChanges` eval which previously had to rely on `BindResult`.

I also considered `yield` since that seems like the most precise word to use, but avoided it for the same reason to switch from `return` (it's a keyword in some languages).

## Agent can set vars in `dagger shell`

`env(privileged: true)` now makes env writable - meaning the model can declare outputs of its own.

You can now do a prompt like this in `dagger shell`:

> set $go to a container for go dev

This works by having `shell` append a system prompt to watch out for `$foo` values and declare them as outputs, which the model then fills in with `save`. The agent also sets a description which is pretty neat.

From then on you can keep using `$go` - except now it'll be expanded as a prompt var. (TODO: maybe that's a bad thing? Depends on if you would need to tell the agent to re-assign. Suggest we leave that for next iteration since this PR's already big and the choice there doesn't feel certain.)

Seems like this could replace `$_` and its kinda weird `bindResult` API but I've kept it for now.

## Exit the LLM loop when the model calls `save`

`qwen2.5-coder:14b` kept calling `return` nonstop, so now we exit the loop when it's called. This `returned?` state resets every time the LLM is given new state (like a new prompt).

## Remove object input getters

Reasoning:

* Inputs could easily be named something that confuses the model.  Even with innocuous names I was seeing a lot of pointless calls to them, especially with local models like `qwen`.
* Each input consumed a tool slot (128 limit). We have to compete for real estate with other tools (i.e. MCP servers), and our scheme already consumes a dynamic amount, so we should minimize our footprint.
* Requiring a full API/token roundtrip just to read a static value like `Container#1` is a bit wasteful.

Now we just list the available objects in the `select_tools` description, and don't even expose their name, to try to prevent models treating them like variables.

Now when you pass inputs to the model it can skip right to using them instead of having to get their IDs.

## Remove string input getters

String inputs are exposed as a `user_provided_values` tool which directly embeds each value (JSON encoded) in its description. This allows models to directly use them without extra roundtrips, which is especially helpful since these values tend to be small. In addition, we show each value's description, not its name, to avoid the model treating them like variables. (If no description is provided, it defaults to the name.)

It might sound a little wacky to embed this stuff directly in the description, but it actually should reduce token cost overall. Even with large values, the value _should_ be used eventually, so the only difference is that we put the value in the tool description instead of in the message history.

## meta

fixes #10039
supersedes #10059
supersedes #9986


core/mcp.go
```diff
@@ -168,22 +167,34 @@ func ToolFunc[T any](fn func(context.Context, T) (any, error)) func(context.Cont
...
 }
 
-func (m *MCP) tools(srv *dagql.Server, typeName string) ([]LLMTool, error) {
+func (m *MCP) tools(srv *dagql.Server) ([]LLMTool, error) {
 	schema := srv.Schema()
+	tools := []LLMTool{}
+	typeNames := m.env.Types()
+	if m.env.Root() != nil {
+		typeNames = append(typeNames, m.env.Root().Type().Name())
+	}
```
cwlbraa, line 176: i'm assuming this is the equivalent of 
```golang
        // if env.Root() != nil {
 	// 	m.Select(env.Root())
 	// }
```
? so we add just the one root type and then expect the LLM to figure out what tools it might want from there? is that Server? Query?

core/llm_docs.md
```diff
@@ -0,0 +1,227 @@
...
+      # - Plate#1: A plate for serving food.
+      #
+      # Available tools:
+      # - Oven_bake
+      # - Oven_broil
+      # - Potato_dice
+      # - Potato_peel
+      # - Potato_slice
+      # - Sink_rinse
+      # - Fries_plate
```
cwlbraa, line 47: is this commented section representative of a dynamically returned tool description? like on each turn we change how the selectTools tool describes itself?

core/mcp.go
```diff
@@ -314,55 +333,114 @@ func (m *MCP) tools(srv *dagql.Server, typeName string) ([]LLMTool, error) {
...
+
 	// 1. CONVERT CALL INPUTS (BRAIN -> BODY)
 	//
-	if m.Current() == nil {
-		return nil, fmt.Errorf("no current context")
-	}
-	target, ok := dagql.UnwrapAs[dagql.Object](m.Current())
-	if !ok {
-		return nil, fmt.Errorf("current context is not an object, got %T", m.Current())
-	}
```
cwlbraa, line 358: 💀 rest in pieces

core/mcp.go
```diff
@@ -632,103 +761,147 @@ Each parameter corresponds to a named result with a specific purpose. Do not cal
...
+				desc := `Select tools for interacting with the available objects.`
+				desc += "\n\nAvailable tools:"
 				for _, tool := range tools {
-					desc += fmt.Sprintf("\n- %s", tool.Name)
+					if m.selectedTools[tool.Name] {
+						// already have it
+						continue
+					}
+					desc += "\n- " + tool.Name + " (returns " + tool.Returns + ")"
+				}
```
cwlbraa, line 843: so with env(privileged:true) what tools are initially available in the tool description? literally all the core API, module and it's dependencies?

(i read some other code that made me think that this was dependent on input objects)

core/env.go
```diff
@@ -153,12 +154,84 @@ func (env *Env) Ingest(obj dagql.Object) string {
...
+		obj += "}"
+		return obj
+	default:
+		return lit.Display()
+	}
+}
+
 func (env *Env) Types() []string {
 	types := make([]string, 0, len(env.typeCount))
 	for typ := range env.typeCount {
```
cwlbraa, line 237: nit: i initially thought this was a `range int`, like for `typ := range 10` and typ was incrementing integers lmao
```suggestion
	for typ := range env.typeCounts {
```

core/mcp.go
```diff
@@ -68,10 +72,7 @@ func newMCP(env *Env, endpoint *LLMEndpoint) *MCP {
...
 
 func (m *MCP) DefaultSystemPrompt() string {
-	if m.needsSystemPrompt {
-		return defaultSystemPrompt
-	}
-	return ""
+	return defaultSystemPrompt
 }
 
 func (m *MCP) WithEnvironment(env *Env) *MCP {
```
cwlbraa, line 78: this was dead code before, and i think it's still dead now?

core/mcp.go
```diff
@@ -68,10 +72,7 @@ func newMCP(env *Env, endpoint *LLMEndpoint) *MCP {
 var defaultSystemPrompt string
 
 func (m *MCP) DefaultSystemPrompt() string {
-	if m.needsSystemPrompt {
-		return defaultSystemPrompt
-	}
-	return ""
+	return defaultSystemPrompt
```
cwlbraa, line 75: nit: clarity is lacking here now, prolly something you intended to clean up, but there's that spot above where needsSystemPrompt hinges on whether we're on gemini or not, but now we ignore that for the `dagger mcp` callpath. 

do we not use this for claude and others when we run in prompt mode? do they just figure it out?

cmd/codegen/generator/go/generator.go
```diff
@@ -427,11 +427,13 @@ func renderFile(
...
 	formatted, err := format.Source(source)
 	if err != nil {
+		os.Stderr.Write(source)
 		return nil, fmt.Errorf("error formatting generated code: %w", err)
 	}
 	formatted, err = imports.Process(filepath.Join(cfg.OutputDir, "dummy.go"), formatted, nil)
 	if err != nil {
-		return nil, fmt.Errorf("error formatting generated code: %w", err)
+		os.Stderr.Write(source)
+		return nil, fmt.Errorf("error processing imports in generated code: %w", err)
```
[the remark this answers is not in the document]
vito: nope, had my own reason for debugging there :P

core/mcp.go
vito: yep! it's Query - but this is untested at the moment

core/llm_docs.md
vito: Yep exactly.

core/mcp.go
vito: Now we set the system prompt for all the models, since it turned out Gemini isn't the only one that needs it especially once we added the `return` tool.

Here are some eval results before/after: https://v3.dagger.cloud/dagger/traces/cf99304783496c922d67d712cc8eb863

core/mcp.go
vito: Should be whatever's defined on `Query`

(pardon the terse replies, heading out)

core/mcp.go
cwlbraa: cool then i guess the cleanup is just that `if (google)` conditional

cmd/codegen/generator/go/generator.go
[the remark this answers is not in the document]
vito: Also I'm inclined to just leave these debug lines there, in the end they still helped more than `-i` and they're already in error paths

core/mcp.go
```diff
@@ -633,103 +747,202 @@ Each parameter corresponds to a named result with a specific purpose. Do not cal
...
 
-func (m *MCP) Builtins(srv *dagql.Server) ([]LLMTool, error) {
-	builtins := []LLMTool{}
-
-	if len(m.env.outputsByName) > 0 {
-		builtins = append(builtins, m.returnBuiltin())
+func (m *MCP) Builtins(srv *dagql.Server, tools []LLMTool) ([]LLMTool, error) {
+	builtins := []LLMTool{
+		{
+			Name:        "think",
```
vito, line 789: On the fence here - it's very interesting to see what it writes, and it's more likely to follow the plan that it writes. We can also give it custom UI. For now it just has a :thought_balloon: emoji but I'd ilke to dim the text.

But I haven't validated whether it actually helps the evals, in isolation from all the other changes.

core/mcp.go
```diff
@@ -633,103 +747,202 @@ Each parameter corresponds to a named result with a specific purpose. Do not cal
...
+				return t.Type().Name() == name
+			}) {
+				continue
+			}
+			allTypes[name] = objectType
 		}
 		builtins = append(builtins, LLMTool{
-			Name:    "select" + typeName,
-			Returns: typeName,
+			Name:        "declareOutput",
```
vito, line 841: Maybe `bindOutput`?

core/mcp.go
```diff
@@ -992,20 +1245,10 @@ func typeToJSONSchema(schema *ast.Schema, t *ast.Type) (map[string]any, error) {
 
 const jsonSchemaIDAttr = "x-id-type"
 
-func idPattern(typeName string) string {
```
vito, line 995: Now we use an `enum` of all possible IDs, which is also respected by Gemini.

core/mcpserver.go
```diff
@@ -243,7 +243,8 @@ func (llm *LLM) MCP(ctx context.Context, dag *dagql.Server) error {
 	}
 
 	s := mcpServer{
-		mcpserver.NewMCPServer("Dagger", "0.0.1"),
+		mcpserver.NewMCPServer("Dagger", "0.0.1",
+			mcpserver.WithInstructions(defaultSystemPrompt)),
```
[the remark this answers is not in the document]
grouville: It works well -- we've been playing with it all day long

cmd/codegen/generator/go/generator.go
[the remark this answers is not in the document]
sipsma: I funnily added the exact same debug lines in a different WIP PR, so +1 to keeping

shykes (approved):

cmd/codegen/generator/go/generator.go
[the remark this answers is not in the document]
cwlbraa: that makes 3 of us lmao

PR #10136, opened 2025-04-11
fix(sdk/elixir): escape special characters in docstrings

wingyplus:
This superseded #10113 with escape both single and triple quote docstrings.


sdk/elixir/lib/dagger/gen/container.ex
```diff
@@ -1213,7 +1213,7 @@ defmodule Dagger.Container do
   @doc """
   Retrieves this container with an unset working directory.
 
-  Should default to "/".
+  Should default to \"/\".
```
jedevc, line 1216: Do `"` need to be escaped in triple quotes? I get that `\` should be escaped.

.dagger/build/sdk.go
```diff
@@ -76,6 +76,7 @@ const TypescriptSDKTSXVersion = "4.15.6"
 func (build *Builder) typescriptSDKContent(ctx context.Context) (*sdkContent, error) {
 	tsxNodeModule := dag.Container(dagger.ContainerOpts{Platform: build.platform}).
 		From(tsdistconsts.DefaultNodeImageRef).
+		WithExec([]string{"npm", "install", "-g", "npm"}).
```
jedevc, line 79: :eyes: is this connected?

TomChv: That line is quite confusing yeah 👀

wingyplus: Oh sorry, I found the build stuck during install tsx on my machine and that issue is resolved when upgrading the npm.

But yeah, it is not related to this PR and should get remove.

sdk/elixir/lib/dagger/gen/container.ex
wingyplus: It is not necessary but need it can so I make consistent to both single and triple quote string.

.dagger/build/sdk.go
wingyplus:
```suggestion
```

sdk/elixir/lib/dagger/gen/container.ex
jedevc: :thinking: it depends on what's common in elixir, but IMO we shouldn't be unnecessarily adding quotes. Yes, there's a visual difference, but `\"` is a different value in a single quoted string vs a triple quoted string.

If we want to keep it consistent, we could always use a triply-quoted string if there's a `"` in the string.

wingyplus: Using triple quote for all of documentation would be a great idea. 

Fixing...

jedevc (approved):

wingyplus:
I see the telemetry test is flake around 2-3 times. 

image

PR #10138, opened 2025-04-11
helm: enable using persistent volume claims in DaemonSet

vmaffet:
This PR enables using PVCs as ephemeral storage for the cache in DaemonSets

⚠️  This moves the following parameters in the values file as pvc are no longer specific to statefulSets
From:
```yaml
engine:
  statefulSet:
    persistentVolumeClaim:
      enabled: false

      # PVC specifications
      storageClassName: ""
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 10Gi
```

To:
```yaml
engine:
  persistentVolumeClaim:
    enabled: false

    # PVC specifications
    volumeClaimTemplate:
      # storageClassName: ""
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 10Gi
```

### DIFF
--- a/helm/dagger/.changes/unreleased/Breaking-20250411-114148.yaml
+++ b/helm/dagger/.changes/unreleased/Breaking-20250411-114148.yaml
@@ -1,0 +1,6 @@
+kind: Breaking
+body: 'helm: Enable using persistent volume claims in DaemonSet'
+time: 2025-04-11T11:41:48.535164+01:00
+custom:
+    Author: vmaffet
+    PR: "10138"
--- a/helm/dagger/values.yaml
+++ b/helm/dagger/values.yaml
@@ -46,17 +46,6 @@ engine:
       # Options: Delete, Retain
       whenScaled: Retain
 
-    # Use PersistentVolumeClaim for data volume
-    persistentVolumeClaim:
-      enabled: false
-      # PVC specifications
-      storageClassName: ""
-      accessModes:
-        - ReadWriteOnce
-      resources:
-        requests:
-          storage: 10Gi
-
     # Controls how pods are created during initial scale up,
     # when replacing pods, and when scaling down.
     # Options: OrderedReady, Parallel (default: OrderedReady)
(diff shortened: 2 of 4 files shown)


helm/dagger/values.yaml
```diff
@@ -150,14 +139,29 @@ engine:
...
 
+  ### Persistent Volume Claim configuration
+  #
+  persistentVolumeClaim:
+    # Use PersistentVolumeClaim for data volume
+    enabled: false
+
+    # PVC specification
+    volumeClaimTemplate:
+      # storageClassName: ""
```
TomChv, line 158: Why commenting storage class name here?

TomChv (commented):
My knowledge on Kubernetes isn't up to date so I may be wrong but is it actually possible to use PVC in DaemonSet? 

Based on https://github.com/kubernetes/kubernetes/issues/78902 I'm not sure, is there any example of PVC with DaemonSet available to look at?

Otherwise the code looks ok, I will let @gerhard or @jedevc double check for approval.

vmaffet:
> My knowledge on Kubernetes isn't up to date so I may be wrong but is it actually possible to use PVC in DaemonSet?
> 
> Based on [kubernetes/kubernetes#78902](https://github.com/kubernetes/kubernetes/issues/78902) I'm not sure, is there any example of PVC with DaemonSet available to look at?
> 
> Otherwise the code looks ok, I will let @gerhard or @jedevc double check for approval.

The implementation here is to support PVCs through generic ephemeral volumes, it enables creating a volume claim of a specified storage class easily (https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes). 
It’s a more limited option than what is available with stateful sets indeed.

helm/dagger/values.yaml
vmaffet: It is better to not define the storage class than set it to `""`  to get the default storage class. 
https://kubernetes.io/docs/concepts/storage/storage-classes/#default-storageclass

This was previously done by checking if it is empty. 
https://github.com/dagger/dagger/blob/fb596490b86fdb64f5ce4102597c200872591083/helm/dagger/templates/engine-statefulset.yaml#L188-L190

However, my PR moves to a more generic expansion of the whole `volumeClaimTemplate` structure so we can set any property (but does not support finer checking like this).

TomChv: Make sense!

helm/dagger/values.yaml
```diff
@@ -150,14 +139,29 @@ engine:
...
       enabled: true
 
+  ### Persistent Volume Claim configuration
+  #
+  persistentVolumeClaim:
+    # Use PersistentVolumeClaim for data volume
+    enabled: false
+
+    # PVC specification
+    volumeClaimTemplate:
```
jedevc, line 157: Do we need this extra layer of indirection? We don't use this `template` kind of style anywhere else.

helm/dagger/.changes/unreleased/Breaking-20250411-114148.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Breaking
+body: 'helm: Enable using persistent volume claims in DaemonSet'
```
jedevc, line 2: Since this is a breaking change, can we get another line explaining how/where the config has moved?

jedevc (commented):
LGTM :tada:

FYI @gerhard @matipan - because this is a breaking change, do we need to hold this till `v0.19.0`?

PR #10163, opened 2025-04-15
cli: perfom file lock when writing config to avoid corruption

marcosnils:
this commit also introduces a check to see if the current
credentials.json file is corrupted and if that is the case, it triggers
the authentication flow again automatically


cmd/dagger/main.go
```diff
@@ -257,6 +262,15 @@ func checkForUpdates(ctx context.Context, w io.Writer) {
 	})
 }
 
+// checkCloudTokens triggers the Login flow if an invalid credentials file is detected.
+func checkCloudToken(ctx context.Context, w io.Writer) error {
+	_, err := auth.Token(ctx)
+	if v, ok := err.(*os.PathError); ok && v.Err == os.ErrNotExist {
```
vito, line 268: The comment says this will run for an invalid/malformed credentials file, but this seems to only check for the file not existing

Also I guess we need to be careful not to do anything if the user is intentionally not logged in (as in they don't use Cloud)

marcosnils: > The comment says this will run for an invalid/malformed credentials file, but this seems to only check for the file not existing

agh! you're correct, forgot to commit some changes after testing errors. Should be ok in the follow-up commit.

marcosnils: > Also I guess we need to be careful not to do anything if the user is intentionally not logged in (as in they don't use Cloud)

 yep, already checked this and we're not doing anything if the user has never logged in.

cmd/dagger/main.go
```diff
@@ -257,6 +263,15 @@ func checkForUpdates(ctx context.Context, w io.Writer) {
 	})
 }
 
+// checkCloudTokens triggers the Login flow if an invalid credentials file is detected.
+func checkCloudToken(ctx context.Context, w io.Writer) error {
+	_, err := auth.Token(ctx)
+	if _, ok := err.(*json.SyntaxError); ok {
```
vito, line 269: worth using `errors.Is` or `errors.As` here since it might be wrapped

vito (approved):

Issue #10129, opened 2025-04-10
🐞 "failed to forward exit code: failed to close terminal: EOF" when running Dagger with -vvvv --progress=plain


cavo789:
### What is the issue?

Hi all

I'm using the latest version of Dagger engine i.e. 0.18.2

By running my module using `dagger call [...] -vvvv --progress=plain --interactive tests` (where `[...]` represent my own flags), I got the error "failed to forward exit code: failed to close terminal: EOF"

```python
container = (
    await container.with_exec(["php", "artisan", "config:clear"])
        .with_exec(["php", "artisan", "key:generate"], expect=ReturnType.SUCCESS)
        .with_exec(["php", "artisan", "optimize"], expect=ReturnType.SUCCESS)
        .with_exec(["php", "artisan", "migrate"], expect=ReturnType.ANY)
)

# If non zero, most probably the database server wasn't configured
if not await ctr.exit_code() == 0:
    raise StepMigrationFailed(await ctr.stderr()) 
```

By updating my CLI to `dagger call [...] tests`, the exact same code is now working.

My `StepMigrationFailed` is just a custom Python exception class that accept an optional string parameter.

I suppose that output is too big due to the presence of the verbose mode / debug info.

Note: I've also enabled debug as explained on Discord (https://discord.com/channels/707636530424053791/1357325635160244374)

(Thank you for Dagger; amazing app)

### Dagger version

dagger v0.18.2 (docker-image://registry.dagger.io/engine:v0.18.2) linux/amd64

jshbrntt:
I get this issue using `--progress plain`.

jedevc:
`--interactive` is not supported with `--progress=plain`. Since you're explicitly opting to not run with the TTY output, then we can't take it over with a shell right now.

It's not impossible to support - but see here: https://github.com/dagger/dagger/blob/20bab132591af1b9401b9d84a0d5a8b231525ec9/cmd/dagger/terminal.go#L18-L20

This error *should* be being propagated, but isn't, which is a bug.

cavo789:
Oh, ok, I've added `--interactive` just because I thought it was nice to jump in a terminal if something was wrong. Didn't know it can have side effect. I can remove it, no prob.

I'm not sure about the rest of your answer. The error I got was *failed to forward exit code: failed to close terminal: EOF* (thus not *running shell without the TUI is not supported*).

By removing the verbose mode (`-vvvv`) I didn't get the error.

jedevc:
[...]

PR #10165, opened 2025-04-15
fix: graceful failure when opening terminals

jedevc:
Partial fix for https://github.com/dagger/dagger/issues/10129, by improving the error message - it still doesn't work, but actually informs the user that it isn't supported.

Previously, we were often returning invalid terminals from the `OpenTerminal` function. This could happen because gRPC does not actually wait for the call to `Session` to complete - instead, any errors returned (like the one that reports that the user can't open a terminal in a `--progress=plain` session) is actually returned on the first call to `Recv`.

To handle this, we now send a void `Ready` message, and wait to receive that - this allows us to correctly receive the error, which we can then correctly propagate upwards to inform the user what really went wrong.

### DIFF
--- a/engine/buildkit/client.go
+++ b/engine/buildkit/client.go
@@ -744,6 +744,9 @@ func (c *Client) OpenTerminal(
 
 	term, err := terminalClient.Session(ctx)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


vito (approved):
Nice!

PR #10168, opened 2025-04-15
connect nested session attachables through http API

sipsma:
Follow-up to https://github.com/dagger/dagger/pull/10162

Still saw an occurrence of the flake, but this time it's because the healthcheck failed 2 times in a row rather than closed early due to a bug: https://v3.dagger.cloud/dagger/traces/64ca948ee08ce288105b158ab8889321?span=8960846c5fe9627d&logs#8960846c5fe9627d:L245

At first I tried disabling the grpc healthchecks entirely, but that just peeled back another layer and revealed more problems where grpc was calling the dialer multiple times (not good when we are just operating off a single pre-made conn); I suspect that the grpc healthchecks were inadvertently acting as something like a keep-alive and/or otherwise tweaking something about the grpc connection state that resulting in it never being closed.

Either way, all that nonsense is not really worth our time right now, so the fix here just re-arranges how nested session attachables are setup:
1. Previously they worked by creating a `socketpair` and giving one end to the session.Manager and one end to the nested exec init, which was appealing in that it was pretty efficient and low-overhead
2. However it had a side effect of sometimes creating a long-ish period of time between when the session.Manager started trying to connect to the attachables in the exec and when the attachables were actually ready, which caused all the woes we were seeing
3. Now, we connect nested attachable the same way that non-nested clients connect through the HTTP API to /sessionAttachables
   * This improves the situation because it means that `session.Manager` only starts trying to connect to the attachables when they are for sure ready-to-go
5. This theoretically has a little bit more overhead, but doesn't appear to be noticeable at this time. It also is simpler in the code, so overall seems like a good change.

### DIFF
--- a/.changes/unreleased/Fixed-20250421-145347.yaml
+++ b/.changes/unreleased/Fixed-20250421-145347.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: |
+  Fix "session not found" errors when invoking modules.
+time: 2025-04-21T14:53:47.33409304-07:00
+custom:
+  Author: sipsma
+  PR: "10168"
--- a/cmd/init/main.go
+++ b/cmd/init/main.go
@@ -8,10 +8,12 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/http"
 	"os"
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"strconv"
 	"syscall"
 	"time"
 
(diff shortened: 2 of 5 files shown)


sipsma:
After a few re-runs I upgraded to the new buildkit error of `transport: Error while dialing: only one connection allowed`.

I'm highly tempted to just rip more of the session grpc code out and into dagger. May require updating our buildkit fork to change `SessionManager` to an interface so we can have our own implementation.

sipsma:
Okay I've re-run `TestContainer` w/out hitting the flake enough times in a row now that I feel good it's squashed, bringing out of draft

vito (approved):

PR #10175, opened 2025-04-16
Fix caching on `+private` state

sipsma:
Our publish job has been failing most of the time on main after upgrading to v0.18.3 (but usually passed on a re-run) Was able to repro locally and realized it has to do with getting cache hits on module objects that have private state. The engine doesn't know about private fields so it can't hold cache refs open to them, which opens the door for getting a cache hit on an object with a local dir stored in private state, the original cache ref for that local dir being released, resulting in it being evicted and then getting a cache miss on a dir the caller knows nothing about and fails to load.
* See the newly added integ test for this in action, which is probably easier to follow than my run-on sentence

Similar sort of cases involving private state have bitten us when it comes to caching in the past, so trying to end the nightmare by updating the engine + all SDKs to actually register private fields, so the engine knows about them, but then just hide them from the public api of the module.

Got this working on Go so far. Tricky part was that previously it was possible to use private state to hide types we don't support like `map`, so needed some careful handing of types like that to keep this backwards compatible.

TODO:
- [ ] Rest of SDKs

### DIFF
--- a/cmd/codegen/generator/go/generator.go
+++ b/cmd/codegen/generator/go/generator.go
@@ -427,6 +427,12 @@ func renderFile(
 
 	formatted, err := format.Source(source)
 	if err != nil {
+		// TODO:
+		// TODO:
+		// TODO:
+		// TODO:
+		// TODO:
+		os.WriteFile("/broken.gen.go", source, 0600)
 		return nil, fmt.Errorf("error formatting generated code: %w", err)
 	}
 	formatted, err = imports.Process(filepath.Join(cfg.OutputDir, "dummy.go"), formatted, nil)
--- a/cmd/codegen/generator/go/templates/module_funcs.go
+++ b/cmd/codegen/generator/go/templates/module_funcs.go
@@ -61,14 +61,14 @@ func (ps *parseState) parseGoFunc(parentType *types.Named, fn *types.Func) (*fun
 			spec.returnsError = true
 			break
 		}
-		spec.returnSpec, err = ps.parseGoTypeReference(result, nil, false)
+		spec.returnSpec, err = ps.parseGoTypeReference(result, nil, false, false)
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse return type: %w", err)
 		}
 	case 2:
 		spec.returnsError = true
 		result := results.At(0).Type()
-		spec.returnSpec, err = ps.parseGoTypeReference(result, nil, false)
+		spec.returnSpec, err = ps.parseGoTypeReference(result, nil, false, false)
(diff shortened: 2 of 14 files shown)


jedevc:
Agreed on wanting to get this once-and-for-all, but just going to share my comment from last time we talked about this: https://github.com/dagger/dagger/pull/8358#issue-2510229992

> [!NOTE]
>
> Using the `unknownCollectIDs` function that PR introduces might help solve this as a stop-gap?
>
> That said, I still think the idea of having a known `id:` prefix, or similar has reasonable potential. Though it still feels like kicking the can down the road.
>
> Or, somehow, having a custom JSON encoder per language that knows how to collect IDs from objects before encoding them, so we can "attach" them.

Handling maps was the main reason to *not* do what's in this PR, since there were cases in some places of doing stuff like `map[string]Container` - we still need to expose the `ContainerID` somehow to get the right behaviors, but it's behind a data type that is JSON-able, but doesn't have fields.

If it was *just* maps though, we could handle this pretty easily, we could add new map typedefs to the dagger registration API - even if we don't expose them in the graphql API (yet), it would be useful to be able to have them as values that could then be neatly marked as private.

jedevc:
:thinking: as another thought, actually. If we're around here, maybe we should start to pull apart the concept of "private"? It feels like there are a few different concepts:

- Completely opaque to the engine (e.g. what `private` is today)
	- It feels like this could still maybe be useful? As long as it doesn't contain dagger objects, which we could potentially actually check for :thinking: 
- Exposed to the engine as a typedef, but not served as an API (maybe "hidden")
- Exposed to the engine as a typedef, served in the API, but *not* exposed to an LLM (we've been calling it "masked" so far)

Definitely a tangent, but just brain dumping something that's been rolling round my head for a couple weeks now.

sipsma:
Well turns out the issue w/ contextual dirs actually gets a bit more fundamental and there's a case where you can hit cache on contextual dirs after contents were changed (i.e. `dagger-dev` module uses contextual args to load integ test code and you make a change to a test case).

So actually I think we're gonna have to solve the even more basic problem of contextual args not counting towards function call cache keys. I was originally planning on including that in the next batch of dagql cache updates but extracting it out to merge quickly.

This will also fix the problem with `+private`, in addition to the newly found problem, so will most likely close this out if it turns out to not be needed now. May want it someday but it's a lot of boat rocking so will avoid it until proven necessary again.

sipsma:
Closing in favor of https://github.com/dagger/dagger/pull/10187 for now. Will re-open if it turns out we need this in the short term too for whatever reason.

PR #10182, opened 2025-04-16
mcp: properly handle almost all types for the default value of a function argument

tiborvass:
Default values for arrays are still handled weirdly but pedantically by only allowing the empty value with the string "[]".


cwlbraa (approved):
don't do it right now but this is classic "damn i wish we wrote a unit test" code

PR #8097, opened 2024-08-03
python: add support for defining interfaces

helderco:
This is a long time coming!

Closes https://github.com/dagger/dagger/issues/6213

## What?

This adds support for defining Dagger interfaces in Dagger modules written in Python.

## How?

To define an interface, create a [Protocol](https://typing.python.org/en/latest/spec/protocol.html) class, decorated with `@dagger.interface`:

```python
import typing
import dagger

@dagger.interface
class Duck(typing.Protocol):
    @dagger.function
    async def quack(self) -> str: ...

    @dagger.function
    def with_color(self, color: str) -> typing.Self: ...
```

Note that:
- The functions in the interface need to be decorated with `@dagger.function` too.
- Unless the return type is an object type, the function should be async to reflect the fact that the automatically generated client bindings for it will need to make an API request.
- To refer to the current interface, use `typing.Self`.

## Example

Here’s the same toy example in the [original issue](https://github.com/dagger/dagger/issues/6213).

`Pond` module (no deps):

```python
import typing
import dagger


@dagger.interface
class Duck(typing.Protocol):
    @dagger.function
    async def quack(self) -> str: ...


@dagger.object_type
class Pond:
    ducks: list[Duck] = dagger.field(default=list)

    @dagger.function
    def with_duck(self, duck: Duck) -> typing.Self:
        self.ducks.append(duck)
        return self

    @dagger.function
    async def quack_all(self) -> str:
        return "\n".join([await duck.quack() for duck in self.ducks])
```

`Mallard` module (no deps):

```python
import dagger

@dagger.object_type
class Mallard:
    @dagger.function
    def quack(self) -> str:
        return "mallard quack"
```

`Crested` module (no deps):

```python
import dagger

@dagger.object_type
class Crested:
    @dagger.function
    def quack(self) -> str:
        return "crested quack"
```

`Caller` module (deps on `Mallard`, `Crested`, and `Pond`):

```python
import dagger
from dagger import dag

@dagger.object_type
class Caller:
    @dagger.function
    async def test(self) -> str:
        return await (
            dag.pond()
            .with_duck(dag.mallard().as_pond_duck())
            .with_duck(dag.crested().as_pond_duck())
            .quack_all()
        )
```

## Future improvements

Support passing an implementation directly to an argument that expects an interface:

```diff
-.with_duck(dag.mallard().as_pond_duck())
+.with_duck(dag.mallard())
```

PR #10188, opened 2025-04-17
docs: add Python interfaces

helderco:
Depends on:
- https://github.com/dagger/dagger/pull/8097

### DIFF
--- a/docs/current_docs/api/interfaces.mdx
+++ b/docs/current_docs/api/interfaces.mdx
@@ -4,12 +4,7 @@ slug: /api/interfaces
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-
 # Interfaces
-
-:::important
-The information on this page is only applicable to Go and TypeScript SDKs. Interfaces are not currently supported in the Python SDK.
-:::
 
 ## Declaration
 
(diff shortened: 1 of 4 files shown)


docs/current_docs/api/interfaces.mdx
```diff
@@ -4,13 +4,8 @@ slug: /api/interfaces
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-
 # Interfaces
 
-:::important
-The information on this page is only applicable to Go and TypeScript SDKs. Interfaces are not currently supported in the Python SDK.
```
jasonmccallister, line 11: Since we have SDKs for Java and PHP now, do we need to keep this and reference those SDKs as not implemented?

@eunomie I'm not sure the Java SDK has support for interfaces?

@dragoonis does the PHP SDK have support for interfaces?

jasonmccallister (changes requested):
Small comment on the supported SDKs section.

docs/current_docs/api/interfaces.mdx
helderco: No, they don't support it. Thanks for the heads up, I can bring it back.

helderco: Done

jasonmccallister (approved):

PR #10191, opened 2025-04-17
fix: raise up --no-mod flag to same level as --mod

jedevc:
It makes sense that the `--no-mod` flag should be allowed everywhere that the `--mod` flag is.

The main rationale for this was around `dagger query`, where it wasn't possible to run without attempting to load the current directory. But it also extends to other commands.

---

I originally raised this in discord [here](https://discord.com/channels/707636530424053791/1003718839739105300/1354148401280716821):

> @jedevc — 25/03/2025, 17:42
> 
> question around --no-mod - I assume it should be allowed wherever a --mod flag is? it feels weird that it currently isn't.
> for example, i can't attach it to dagger query, if there is a module in the current directory, that command will always attempt to load it 😢

> @shykes — 25/03/2025, 17:52
>
> makes sense to me yeah

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -32,7 +32,7 @@ var (
 	}
 
 	moduleURL         string
-	moduleFlags       = pflag.NewFlagSet("module", pflag.ContinueOnError)
+	moduleNoURL       bool
 	allowedLLMModules []string
 
 	sdk           string
@@ -94,30 +94,41 @@ func getCompatVersion() string {
 	return compatVersion
 }
 
-func init() {
-	moduleFlags.StringVarP(&moduleURL, "mod", "m", "", "Path to the module directory. Either local path or a remote git repo")
+func moduleAddFlags(cmd *cobra.Command, persistent bool) {
+	var flags *pflag.FlagSet
+	if persistent {
+		flags = cmd.PersistentFlags()
+	} else {
+		flags = cmd.Flags()
+	}
+	flags.StringVarP(&moduleURL, "mod", "m", "", "Module reference to load, either a local path or a remote git repo (defaults to current directory)")
+	flags.BoolVarP(&moduleNoURL, "no-mod", "M", false, "Don't automatically load a module (mutually exclusive with --mod)")
+	cmd.MarkFlagsMutuallyExclusive("mod", "no-mod")
+
 	var defaultAllowLLM []string
 	if allowLLMEnv := os.Getenv("DAGGER_ALLOW_LLM"); allowLLMEnv != "" {
 		defaultAllowLLM = strings.Split(allowLLMEnv, ",")
 	}
-	moduleFlags.StringSliceVar(&allowedLLMModules, "allow-llm", defaultAllowLLM, "List of URLs of remote modules allowed to access LLM APIs, or 'all' to bypass restrictions for the entire session")
+	flags.StringSliceVar(&allowedLLMModules, "allow-llm", defaultAllowLLM, "List of URLs of remote modules allowed to access LLM APIs, or 'all' to bypass restrictions for the entire session")
+}
 
+func init() {
 	for _, fc := range funcCmds {
 		if !fc.DisableModuleLoad {
-			fc.Command().PersistentFlags().AddFlagSet(moduleFlags)
+			moduleAddFlags(fc.Command(), true)
 		}
 	}
 
-	funcListCmd.PersistentFlags().AddFlagSet(moduleFlags)
-	listenCmd.PersistentFlags().AddFlagSet(moduleFlags)
-	queryCmd.PersistentFlags().AddFlagSet(moduleFlags)
-	configCmd.PersistentFlags().AddFlagSet(moduleFlags)
+	moduleAddFlags(funcListCmd, true)
+	moduleAddFlags(listenCmd, true)
+	moduleAddFlags(queryCmd, true)
+	moduleAddFlags(configCmd, true)
 
-	mcpCmd.PersistentFlags().AddFlagSet(moduleFlags)
+	moduleAddFlags(mcpCmd, true)
 
-	shellCmd.PersistentFlags().AddFlagSet(moduleFlags)
-	rootCmd.Flags().AddFlagSet(moduleFlags)
+	moduleAddFlags(shellCmd, true)
 	shellAddFlags(shellCmd)
+	moduleAddFlags(rootCmd, false)
 	shellAddFlags(rootCmd)
 
 	moduleInitCmd.Flags().StringVar(&sdk, "sdk", "", "Optionally install a Dagger SDK")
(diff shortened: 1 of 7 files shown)


helderco:
I've been meaning to do this for a while but never gotten around to it. 

I would exclude `call`, `core`, and `functions`. Unless `dagger call -M` basically becomes `dagger core`. If so, should we deprecate `dagger core`?

cmd/dagger/module.go
```diff
@@ -94,30 +94,41 @@ func getCompatVersion() string {
...
-	moduleFlags.StringVarP(&moduleURL, "mod", "m", "", "Path to the module directory. Either local path or a remote git repo")
+func moduleAddFlags(cmd *cobra.Command, persistent bool) {
+	var flags *pflag.FlagSet
+	if persistent {
+		flags = cmd.PersistentFlags()
+	} else {
+		flags = cmd.Flags()
+	}
+	flags.StringVarP(&moduleURL, "mod", "m", "", "Path to the module directory. Either local path or a remote git repo")
+	flags.BoolVarP(&moduleDisabled, "no-mod", "M", false, "Don't load module during shell startup (mutually exclusive with --mod)")
```
helderco, line 105: Need to generalize from "shell".

cmd/dagger/module.go
```diff
@@ -32,7 +32,7 @@ var (
 	}
 
 	moduleURL         string
-	moduleFlags       = pflag.NewFlagSet("module", pflag.ContinueOnError)
+	moduleDisabled    bool
```
helderco, line 35: "Disabled" can be misleading. For example in shell you can start with -M but should be able load on .cd

jedevc:
> I would exclude `call`, `core`, and `functions`. Unless `dagger call -M` basically becomes `dagger core`. If so, should we deprecate `dagger core`?

:thinking: yeah, not 100% how to handle these cases. ATM in this PR, they parse correctly, but error out because they require a module. Similar to `install`/`update`/etc. We could remove it for those, but then we still have a few places where you can't `-m`/`-M` interchangeably - I think it makes sense to be able to error at runtime? Don't really have strong opinions here.

For `call` specifically - I would *like* to have `dagger call -M`. But also, this is what I originally wanted here: https://github.com/dagger/dagger/issues/6947#issuecomment-2135596979. I think it's confusing that there's a `core` + `call` that are the same actions. I think we should support both, and deprecate `core`? But also very happy to keep both around.

cmd/dagger/module.go
jedevc: Renamed to `moduleNoURL`.

cmd/dagger/module.go
```diff
@@ -94,30 +94,41 @@ func getCompatVersion() string {
...
-			fc.Command().PersistentFlags().AddFlagSet(moduleFlags)
+			moduleAddFlags(fc.Command(), true)
 		}
 	}
 
-	funcListCmd.PersistentFlags().AddFlagSet(moduleFlags)
-	listenCmd.PersistentFlags().AddFlagSet(moduleFlags)
-	queryCmd.PersistentFlags().AddFlagSet(moduleFlags)
-	configCmd.PersistentFlags().AddFlagSet(moduleFlags)
+	moduleAddFlags(funcListCmd, true)
```
helderco, line 122: `dagger functions` could at some point return core functions if `-M` is used, but I don't see a point in allowing it if it's just going to return an error.

cmd/dagger/module.go
```diff
@@ -94,30 +94,41 @@ func getCompatVersion() string {
...
 	}
-	moduleFlags.StringSliceVar(&allowedLLMModules, "allow-llm", defaultAllowLLM, "List of URLs of remote modules allowed to access LLM APIs, or 'all' to bypass restrictions for the entire session")
+	flags.StringSliceVar(&allowedLLMModules, "allow-llm", defaultAllowLLM, "List of URLs of remote modules allowed to access LLM APIs, or 'all' to bypass restrictions for the entire session")
+}
 
+func init() {
 	for _, fc := range funcCmds {
 		if !fc.DisableModuleLoad {
-			fc.Command().PersistentFlags().AddFlagSet(moduleFlags)
+			moduleAddFlags(fc.Command(), true)
```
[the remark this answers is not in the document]
helderco: A bit torn here. `dagger call -M` does make some sense  as it's the same as `dagger core`, but I don't see a point in adding `-m` to `dagger core`.

I would just break `funcCmds`, keep `moduleAddFlags(callModCmd.Command(), false)` and turn `dagger core` into an alias for `call -M`. `funcCmds` was meant to allow for more call-like commands, but shell is the new approach so I don't see us adding more of these.

cmd/dagger/module.go
```diff
@@ -127,25 +138,25 @@ func init() {
...
 	moduleInstallCmd.Flags().StringVarP(&installName, "name", "n", "", "Name to use for the dependency in the module. Defaults to the name of the module being installed.")
 	moduleInstallCmd.Flags().StringVar(&compatVersion, "compat", modules.EngineVersionLatest, "Engine API version to target")
-	moduleInstallCmd.Flags().AddFlagSet(moduleFlags)
+	moduleAddFlags(moduleInstallCmd, true)
 
 	moduleUnInstallCmd.Flags().StringVar(&compatVersion, "compat", modules.EngineVersionLatest, "Engine API version to target")
+	moduleAddFlags(moduleUnInstallCmd, true)
 
 	moduleUpdateCmd.Flags().StringVar(&compatVersion, "compat", modules.EngineVersionLatest, "Engine API version to target")
+	moduleAddFlags(moduleUpdateCmd, true)
```
helderco, line 151: I'm torn here as well. All the module management commands only make sense in the context of a module, so it doesn't make sense to add `-M`.

helderco (commented):
Maybe split `moduleAddFlags` in two? One for commands where a module is required (no `-M`) and another where it's optional (`-M` is allowed).

For example, in `dagger query` it makes sense to have module loading be optional, but `dagger install` only makes sense with a module.

cmd/dagger/module.go
helderco: In other words, if a user calls `dagger functions --help`, I think it's reasonable to expect that what's listed as available options can in fact be used and not error without a solution.

cmd/dagger/module_inspect.go
```diff
@@ -37,6 +37,9 @@ func initializeCore(ctx context.Context, dag *dagger.Client) (rdef *moduleDef, r
 // By default, looks for a module in the current directory, or above.
 // Returns an error if the module is not found or invalid.
 func initializeDefaultModule(ctx context.Context, dag *dagger.Client) (*moduleDef, error) {
+	if moduleNoURL {
+		return nil, fmt.Errorf("cannot load module when --no-module is specified")
```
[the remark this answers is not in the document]
jedevc: Yeah, fair! Refactored a bit, so that now `-M` isn't valid for the cases where we *need* a module.

Leaving the error message in though, just in case it gets hit, better than panicing or similar down the line.

cmd/dagger/module.go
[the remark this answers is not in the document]
jedevc: Done!

jedevc:
Apparently I can't reopen this - moved to https://github.com/dagger/dagger/pull/10595.

PR #10194, opened 2025-04-17
docs: Add link to IDE integration page in quickstarts

vikram-dagger:


vikram-dagger:
> I feel like this should be a partial that we import so we can update easier in the past, but LGTM!

Good idea. Added this now.

jasonmccallister (approved):

PR #10195, opened 2025-04-17
Add contextual git directories

### LINKED ISSUE #8520: API to query current git context

jedevc:
Fixes #8520

> [!WARNING]
>
> This PR is under construction! :construction_worker_woman: 

This allows a module to get git information about itself - it can do this through the core `Git` type. Later on, we'll extend the API with more functionality, like in #9968 which will be able to track dirty state, and get diffs.

This was *previously* sort of possible, by using a context dir and calling `.asGit` on it, but:
1. it's not particularly ergonomic, and doesn't convey the user's intention particularly well
2. it has to upload all the files in the local directory to handle this - we should be able to apply optimizations in quite a few cases, which should make it faster.

---

TODO:
- [ ] Ensure we have design consensus
- [ ] Engine plumbing
	- [x] Initial implementation
	- [ ] Fix outstanding comments
- [ ] SDK support
	- [x] Go SDK
	- [ ] Python SDK
	- [ ] Typescript SDK
- [ ] Testing
	- [x] Local source tests
	- [ ] Git source tests through http://github.com/dagger/dagger-test-modules

### DIFF
--- a/cmd/codegen/generator/go/templates/module_funcs.go
+++ b/cmd/codegen/generator/go/templates/module_funcs.go
@@ -147,6 +147,9 @@ func (spec *funcTypeSpec) TypeDefCode() (*Statement, error) {
 		if argSpec.defaultPath != "" {
 			argOptsCode = append(argOptsCode, Id("DefaultPath").Op(":").Lit(argSpec.defaultPath))
 		}
+		if argSpec.defaultGit != "" {
+			argOptsCode = append(argOptsCode, Id("DefaultGit").Op(":").Lit(argSpec.defaultGit))
+		}
 
 		if len(argSpec.ignore) > 0 {
 			ignores := make([]Code, 0, len(argSpec.ignore))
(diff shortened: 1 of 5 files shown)


cmd/codegen/generator/go/templates/module_funcs.go
```diff
@@ -378,6 +391,9 @@ type paramSpec struct {
 	// Only applies to arguments of type File or Directory.
 	// If the argument is not set, load it from the given path in the context directory
 	defaultPath string
+	// Only applies to arguments of type GitRepository or GitRef.
+	// If the argument is not set, load it from the given path in the context git
+	defaultGit string
```
wingyplus, line 396: Note for me: it needs to add an option to the Elixir SDK to support this feature.

PR #10206, opened 2025-04-18
docs: improve features/shell

shykes:


nipuna-perera (commented):
I was looking for this today and couldn't find it in the docs. Please add an example/instructions on how to pass arguments to the constructor when the module is already loaded. They syntax for that is

`. --myArg value | moduleFunction args...`

It wasn't apparent to me and `.help` didn't make it clear.

docs/current_docs/features/shell.mdx
```diff
@@ -67,13 +71,30 @@ container |
...
 
-Dagger Shell supports multiple ways to run commands:
+- Shell variables: `container=$(container | from alpine)`
+- Shell functions: `container() { container | from alpine; }`
+- Job control: `frontend | test & backend | test & wait`
+- Quoting: single quotes and double quotes have the same meaning as in bash
+
+## Tracing
+
+Dagger supports opentelemetry tracing out of the box, and Dagger Shell takes advant
```
vikram-dagger, line 93: Quick note: incomplete sentence

shykes: Thank you! My kids startled me and I lost track when I resumed work.

nipuna-perera:
> > I was looking for this today and couldn't find it in the docs. Please add an example/instructions on how to pass arguments to the constructor when the module is already loaded. They syntax for that is
> > 
> > `. --myArg value | moduleFunction args...`
> > 
> > It wasn't apparent to me and `.help` didn't make it clear and I figured it out by trial and error.
> 
> Will do. By the way you can also use the module name instead of `.`.
> 
> For example:
> 
> ```console
> .cd github.com/dagger/dagger/docs
> docs --source=. | ...
> ```
> 

TIL

vikram-dagger:
> I was looking for this today and couldn't find it in the docs. Please add an example/instructions on how to pass arguments to the constructor when the module is already loaded. They syntax for that is
> 
> `. --myArg value | moduleFunction args...`
> 
> It wasn't apparent to me and `.help` didn't make it clear and I figured it out by trial and error.

JFYI there is an example of similar usage in https://docs.dagger.io/cookbook#set-a-module-wide-default-path

vikram-dagger:
- Some of the information in the "modules" section (constructors section for example) seems too detailed  for this high-level features page. Would it make sense to move some details elsewhere and link to it?
- It feels like there is tension between this page and the one at https://docs.dagger.io/api/cli. Both discuss shell but the second one also covers other CLI commands like `call`, `core`, etc. Should we update that other page as well? Maybe split into two sections, one only for shell and one for other commands?

shykes:
> * It feels like there is tension between this page and the one at https://docs.dagger.io/api/cli. Both discuss shell but the second one also covers other CLI commands like `call`, `core`, etc. Should we update that other page as well? Maybe split into two sections, one only for shell and one for other commands?

I guess that tension already exists in the current version. So can we address it separately?

vikram-dagger:
> I guess that tension already exists in the current version. So can we address it separately?

Yes, it exists now as well. Fine to address it separately.

vikram-dagger (approved):

vikram-dagger:
LGTM, approving! Rhere is an open question about the shell built-ins so I'm not sure whether to merge right away or wait for that discussion to resolve.

shykes:
Thanks @vikram-dagger . I pushed a few changes:

- Removed references to `.cd`
- Generally simplified the content. We lose a few details in the process, but it felt like a good balance. We definitely should have a meatier shell reference.

vikram-dagger:
> * Generally simplified the content. We lose a few details in the process, but it felt like a good balance. We definitely should have a meatier shell reference.

Agreed, and imho a reworked version of https://docs.dagger.io/api/cli would work for this.

PR #10210, opened 2025-04-18
Update helm chart to support setting runtimeClassName

bleggett:
Small change to make setting the `runtimeClassName` possible via values.yaml override, similar to how `priorityClassName` is already overridable.


helm/dagger/templates/engine-daemonset.yaml
```diff
@@ -29,6 +29,9 @@ spec:
         name: {{ include "dagger.fullname" . }}-engine
         {{- include "dagger.labels" . | nindent 8 }}
     spec:
+      {{- if .Values.engine.runtimeClassName }}
+      runtimeClassName: {{ .Values.engine.runtimeClassName }}
```
jedevc, line 33: nit: should this appear next to `priorityClassName` for readability?

jedevc (approved):

PR #10213, opened 2025-04-19
fix current module during initialization

sipsma:
Module initialization (when we ask the runtime for what type defs it has) previously got just enough data to support checking the current module name, but basically nothing else. This is because the module is not yet fully constructed during initialization.

This update results in as much of the module being available during init as is possible. This should cover various cases like:
1. access to dag.CurrentModule().Source()
2. creating execs during module initialization (which previously panicked because the instance ID was empty)
   * also check for that condition that created a panic and error instead, just in case that codepath can be triggered some other way


vito (approved):

sipsma:
`TestContainer` fail is the flake fixed here: https://github.com/dagger/dagger/pull/10168, merging

PR #10229, opened 2025-04-21
llm: chaining tool calls

vito:


gerhard:
Now is a good time to rebase:
- https://github.com/dagger/dagger/pull/10327

vito:
superseded by #10366 which now has chaining features too - in that PR it works more reliably than this iteration (maybe "method chaining" as a phrase gets the model in the right framing?)

Issue #10222, opened 2025-04-21
panic in buildkit solver, possibly related to edge merging


sipsma:
Saw this happen one time so far: https://v3.dagger.cloud/dagger/traces/cdca068a2bc3f4a24669294d6d55b6a2?span=9bb360c22dab5941&logs

```
time="2025-04-19T01:42:46Z" level=debug msg="merging edges" dest_index=2 dest_vertex_digest="sha256:25054e2b8aa9a0a72570be4927725d143b9efc775b097111ab630993fc648079" dest_vertex_name="npm pkg set type=module" source_edge_index=2 source_edge_vertex_digest="sha256:8e05840352c710528a008aaf070ad90ea1c78ac5b3ae04d2ae04736a0e93df6b" source_edge_vertex_name="npm pkg set type=module"
time="2025-04-19T01:42:46Z" level=debug msg="merging edges" dest_index=3 dest_vertex_digest="sha256:25054e2b8aa9a0a72570be4927725d143b9efc775b097111ab630993fc648079" dest_vertex_name="npm pkg set type=module" source_edge_index=3 source_edge_vertex_digest="sha256:8e05840352c710528a008aaf070ad90ea1c78ac5b3ae04d2ae04736a0e93df6b" source_edge_vertex_name="npm pkg set type=module"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x534697]

goroutine 4188718 [running]:
github.com/pkg/errors.(*withStack).Error(0x410cc5?)
        <autogenerated>:1 +0x17
github.com/moby/buildkit/solver.notifyStarted.func1({0x3829f80, 0x0}, 0x0)
        /go/pkg/mod/github.com/dagger/buildkit@v0.0.0-20250418225557-9389ea772a3a/solver/jobs.go:1236 +0x142
github.com/moby/buildkit/solver.(*sharedOp).CalcSlowCache(0xc006975c00, {0x385aef0, 0xc01131d950}, 0x2, 0x34802e8, 0x0, {0x385bc18, 0xc01cb40138})
        /go/pkg/mod/github.com/dagger/buildkit@v0.0.0-20250418225557-9389ea772a3a/solver/jobs.go:1006 +0x3d3
github.com/moby/buildkit/solver.(*edge).createInputRequests.(*edge).createInputRequests.func1.func2({0x385aef0?, 0xc01131d950?})
[...]

PR #10231, opened 2025-04-22
try disabling buildkit edge merging

sipsma:
I realized that the [engine-wide dagql cache](https://github.com/dagger/dagger/pull/9682) *theoretically* may have made buildkit's "edge merge" feature redundant.

This would be great because edge merging has proven to be an enormous source of weird ephemeral bugs that are almost impossible to understand or fix
* [latest example potentially](https://github.com/dagger/dagger/issues/10222)

Seeing what happens in CI if it's disabled ([commit in my buildkit fork](https://github.com/sipsma/buildkit/commit/fb00ec7eb5287314229b7ab2cf9204adbe106da5) that this PR depends on). Will need a lot of runs and checking of things like disk space usage before merge (if it pans out at all now).

cc @jedevc

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,12 @@ require (
 replace (
 	dagger.io/dagger => ./sdk/go
 	github.com/dagger/dagger/engine/distconsts => ./engine/distconsts
-	github.com/moby/buildkit => github.com/dagger/buildkit v0.0.0-20250418225557-9389ea772a3a
+	// TODO:
+	// TODO:
+	// TODO:
+	// TODO:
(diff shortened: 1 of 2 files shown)


jedevc:
Mmmm, this seems fairly reasonable to me.

:thinking: I can't shake the feeling that we'd be missing *some* important optimization - but also, I feel like merging at the dagql level gives us so much more control. We really just want to know when we're doing this, instead of what happens today.

LGTM (in spirit :tada:)

sipsma:
Looks good so far in terms of test runtime. Only thing is a single consistently failing test in `TestContainer/TestNestedExec/cached`, which is interesting.

Will put this on the backburner for a bit to work on other higher priority things, seems worth picking back up soon though.

PR #10236, opened 2025-04-22
chore: use new experimental directive in more places

jedevc:
Since we added this new directive which is the right way to convey this metadata, we should mark all the existing experimental methods like this as well.

At least that way we're consistent!


jedevc:
FYI @dagger/sdk-java @dagger/sdk-elixir @dagger/sdk-php 

It would be great if you were able to add support for this annotation - otherwise, we're not able to read this new annotations.

This annotation is currently used for rich metadata about methods that aren't directly part of the graphql spec, currently:
- Sourcemap information about where module functions are implemented
- Experimental annotations

In the future:
- Enum values (https://github.com/dagger/dagger/pull/9518)

It'll always be optional to do (the base introspection should always give reasonable results), but I'd like to move more metadata to these custom directives, since it gives us a neat out-of-band way to convey richer information.

eunomie:
@jedevc I'm looking at implementing them in the Java SDK. But I can't find any example with an experimental tag.
When I'm generating introspection file (using `dagger query --doc ./cmd/codegen/introspection/introspection.graphql` for instance) I have some `sourceMap`, some `trivialResolveField` but nothing like an experimental flag.

For instance the `llm` object looks like to have the experimental flag https://github.com/dagger/dagger/blob/92c686f3fe6a912586915ad3715de8cbd6c994dc/core/schema/llm.go#L17-L23
But the generated json has nothing experimental:

```json
                    {
                        "args": [
                            {
                                "defaultValue": null,
                                "deprecationReason": null,
                                "description": "Model to use",
                                "directives": [],
                                "isDeprecated": false,
                                "name": "model",
                                "type": {
                                    "kind": "SCALAR",
                                    "name": "String",
                                    "ofType": null
                                }
                            },
                            {
                                "defaultValue": null,
                                "deprecationReason": null,
                                "description": "Cap the number of API calls for this LLM",
                                "directives": [],
                                "isDeprecated": false,
                                "name": "maxAPICalls",
                                "type": {
                                    "kind": "SCALAR",
                                    "name": "Int",
                                    "ofType": null
                                }
                            }
                        ],
                        "deprecationReason": null,
                        "description": "Initialize a Large Language Model (LLM)",
                        "directives": [],
                        "isDeprecated": false,
                        "name": "llm",
                        "type": {
                            "kind": "NON_NULL",
                            "name": null,
                            "ofType": {
                                "kind": "OBJECT",
                                "name": "LLM",
                                "ofType": null
                            }
                        }
                    },
```

Any idea what I am missing here? Is the query I'm using to generate the introspection file wrong?

jedevc:
@eunomie it works for in v0.18.4 - the `experimental` directive wasn't actually in v0.18.3 - would that explain it?

Running your above command against an engine in v0.18.4 gives me this snippet:

```json
                        "description": "Initialize a Large Language Model (LLM)",
                        "directives": [
                            {
                                "args": [
                                    {
                                        "name": "reason",
                                        "value": "\"LLM support is not yet stabilized\""
                                    }
                                ],
                                "name": "experimental"
                            }
                        ],
                        "isDeprecated": false,
                        "name": "llm",
```

eunomie:
@jedevc well, this is working with `main`. Maybe I wasn't on the right commit.
But all good, I'll see to implement it for the Java SDK 👍

charjr:
@jedevc 

> It would be great if you were able to add support for the custom directives added in #8498 - otherwise, we're not able to read this new annotations. 

At a quick glance I can't immediately tell what I'd need to add, but when I get the time I will probably ping you for help :sweat_smile:

PR #10248, opened 2025-04-23
feat: move http auth methods to top-level git

jedevc:
...and deprecate the standalone `WithAuthToken` and `WithAuthHeader` methods.
    
Originally, this started as just an attempt to configure HTTP and SSH authentication similarly - we wanted to standardize on doing it all in the top-level constructor, or all using `With` helpers. I have a preference for keeping at the top-level, so that a user can't mistakenly do `Directory.AsGit().WithAuthToken()`, which wouldn't be used. Additionally, we're currently relying on the fact that `dag.Git` doesn't *actually* query anything in the git repository, all operations are deferred till later - but we very well may want to change that in the future (especially as we get more caching enhancements). However, as I refactored the `git` implementation, I realized there's a bigger inconsistency here.

Prior to this refactor, we were loading credentials entirely inconsistently. While for SSH, we would avoid a potentially expensive request to the client for credentials if we were already provided them, for HTTP, we were unconditionally just fetching them. This resulted in a ton of code sprawl, which, if we move everything to the constructor, is vastly simplified.

TODO:
- [x] Design consensus (see :thread: https://discord.com/channels/707636530424053791/1346856793074307202)
	> @shykes
	> I think constructor as you suggested is good
	> I just wanted to make sure you didn't take inspiration from `Container.withRegistryAuth` in your quest for consistency
	>
	> @jedevc
	> happy to rework that one as a separate pr? 
	>
	> @shykes
	> That would be amazing, it's been on my list for ages but I keep not getting around to it
- [x] Implementation
- [x] Tests


core/schema/git.go
```diff
@@ -123,191 +156,202 @@ func (s *gitSchema) git(ctx context.Context, parent dagql.Instance[*core.Query],
...
 		if remote.User == nil {
+			// default to git user for SSH, otherwise weird incorrect defaults
+			// like "root" can get applied in various places. This matches the
+			// git module source implementation.
 			remote.User = url.User("git")
 		}
-		pass, ok := remote.User.Password()
-		if ok {
-			remote.User = url.UserPassword(remote.User.Username(), pass)
-		}
```
jedevc, line 160: @sipsma so this was part of a git fix you did in [`68b2782` (#9682)](https://github.com/dagger/dagger/pull/9682/commits/68b2782cc3dd5ca2bf7a08a8ea34b01d22e3b35f)

But, I'll be honest, I can't quite tell what this bit is for? It seems to just add the password... which is already in there? Just wanted to check.

This whole URL manipulation actually gets so much neater with the forked `gitutil` package, since now there's `remote.String()`, which allows us to manipulate any of those URL fields inline.

jed:
what do i need to do not to get mentioned in this repo any more?

jedevc:
@jed apologies there's a confusion when copy-pasting from our discord, where I go by `jed`. I've updated my username there in the hope that should stop the tagging confusion. Genuinely really sorry for the inconvenience :cry:

jed:
i appreciate that. github clearly needs a little more friction for @ mentions...

core/schema/git.go
sipsma: Believe there was an intention here to handle a case of `username == "" && userpassword != ""`, but it's off obviously. I was in the midst of trying to make all the pre-existing git integ tests pass when I added this code so I probably just moved on after they passed and didn't notice this bit was no-op. So nothing significant, thanks for cleaning up

docs/docs-graphql/schema.graphqls
```diff
@@ -2454,13 +2454,13 @@ type GitRepository {
   withAuthHeader(
     """Secret used to populate the Authorization HTTP header"""
     header: SecretID!
-  ): GitRepository!
+  ): GitRepository! @deprecated(reason: "Use \"httpAuthHeader\" instead.")
```
helderco, line 2457: Maybe add "in the constructor" (or similar) to clarify?

helderco (approved):
lgtm

docs/docs-graphql/schema.graphqls
jedevc: :tada: yup agreed! Thanks :pray:

grouville:
Coming in a bit late, but it's very neat 😍

gerhard:
I didn't go through all the code, but this started failing after we merged this PR: https://v3.dagger.cloud/dagger/traces/2fd74b07d6984b17ffefd842765a87c0?span=5df1ac3b414b9694

image

Do you see the problem @grouville @sipsma ?

sipsma:
> I didn't go through all the code, but this started failing after we merged this PR: https://v3.dagger.cloud/dagger/traces/2fd74b07d6984b17ffefd842765a87c0?span=5df1ac3b414b9694
> 
> image
> 
> Do you see the problem @grouville @sipsma ?

@jedevc sent out a follow up to this PR w/ a fix for that: https://github.com/dagger/dagger/pull/10297

PR #10254, opened 2025-04-23
llm: fix handling of optional ID args

vito:
This is a bit of a grab-bag now fixing various issues @kpenfound ran into for https://github.com/kpenfound/hello-dagger/pull/2

### DIFF
--- a/core/integration/llmtest/allow-llm.golden
+++ b/core/integration/llmtest/allow-llm.golden
@@ -10,11 +10,42 @@
   },
   {
     "role": "assistant",
-    "content": "Hello! How can I help you today?",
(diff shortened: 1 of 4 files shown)


kpenfound:
This is working well with my agent that encountered the original bugs. Let me know if there's anything I can do to help get this merged!

vito:
@kpenfound Just needs approval I suppose but more testing couldn't hurt. The evals look good but I'm particularly curious about the simplified string responses. Your original use case covers that too though. CI is green and this is mostly obvious fixes so I'm comfortable with it as is.

cwlbraa (approved):

PR #10256, opened 2025-04-23
rein in evals cost

vito:
* only run when relevant files change (currently running on every merge :money_with_wings:)
* stop testing `gpt-4o` (4.1 effectively replaces it, not worth 2x cost)


samalba (approved):
🚀🚀🚀

PR #10259, opened 2025-04-24
docs: Replace some instances of "pipeline" with "workflow"

vikram-dagger:
The term "pipeline" is well understood in CI context. However Dagger is relevant to more than just CI use cases. There are now multiple AI use cases which are a good fit for Dagger and the term "pipeline" is confusing in those contexts and to those audiences. We are now using "composable workflows" on our main website to cover both CI and AI use cases and this change should also be reflected in our docs.

This commit
- replaces usage of "pipeline" with "workflow" in some pages
- - excludes CI-specific pages, which continue to use "pipeline" or "CI pipeline" as this is the correct and more precise terminology for the CI context
- updates the title of  https://docs.dagger.io/features/programmable-pipelines from "Programmable Pipelines" to "Programmable Workflows" and the URL to https://docs.dagger.io/features/programmability 
- - uses one word to match other feature page URLs and removes the noun in case we need further changes to the "pipeline/workflow" terminology


kpenfound:
I know we talked about it synchronously but can you add the reasoning for the change to the description @vikram-dagger

vikram-dagger:
> I know we talked about it synchronously but can you add the reasoning for the change to the description @vikram-dagger

Done

kpenfound (approved):
LGTM!

vikram-dagger:
Adding @shykes and @jpadams for review

PR #10264, opened 2025-04-24
skip deduping running nested execs

sipsma:
De-duping currently running nested execs, such as module function calls, can break when:
1. Client A starts the function call
2. Client B makes the same call, gets deduped with the existing one
3. Client A cancels the call (i.e. ctrl-C's, CI job shuts down for whatever reason, etc.)
4. The call continues running, but now is trying to talk back to a session that doesn't exist anymore when dagger APIs are invoked

Errors vary, but example from the new integ test that repros this:
```
    testctx.go:197: 29  : │ [15.0s] | failed to return error: input: get or init client: initialize client: failed to load module: load moduleSource(refString: ".").asModule: Module!: load base: load: failed to stat local path: failed to stat path: failed to get requester session: session for "b9y05fidme5ahpym64tr8ymq5" not found
    testctx.go:197: 29  : │ [15.0s] |  
    testctx.go:197: 29  : │ [15.0s] | original error: input: get or init client: initialize client: failed to load module: load moduleSource(refString: ".").asModule: Module!: load base: load: failed to stat local path: failed to stat path: failed to get requester session: session for "b9y05fidme5ahpym64tr8ymq5" not found
    testctx.go:197: 29  : │ [15.0s] | Stderr:
    testctx.go:197: 29  : │ [15.0s] | time=2025-04-24T20:39:19.608Z level=ERROR msg="failed to emit telemetry" error="failed to upload metrics: failed to send metrics to http://127.0.0.1:38113/v1/metrics: 400 Bad Request"
```

Fixing this by adding support for skipping the dedupe behavior for selected calls.

---

After fixing this in dagql I started hitting failures in the new test case due to buildkit's edge merging. I tried disabling it in buildkit entirely but it seems to have broken the cache in other ways (inconsistent cache keys that result in inconsistent cache hits, seemingly).
* There's a path to addressing this but it will require content caching every mount + copy operation in dagger to replace buildkit's, which I started down the path of but it seemed to have some noticeable performance impacts since it required calculating the content hashes "synchronously" rather than on-demand. This can be addressed with the other ongoing dagql cache work, but not doable in the time frame we want this fix in.

---

TODO:
- [x] depends on a change in `sipsma/buildkit`: https://github.com/sipsma/buildkit/commit/aaabd126316fde846dd8c127780e0873c8ddbb83, will push to `dagger/buildkit` after LGTM


core/container_exec.go
```diff
@@ -308,6 +308,9 @@ func (container *Container) WithExec(ctx context.Context, opts ContainerExecOpts
 		llb.Platform(platform.Spec()),
 		execMDOpt,
 	}
+	if opts.ExperimentalPrivilegedNesting {
+		marshalOpts = append(marshalOpts, llb.SkipEdgeMerge)
```
jedevc, line 312: neat :eyes:

dagql/objects.go
```diff
@@ -357,15 +357,15 @@ func NoopDone(res Typed, cached bool, rerr error) {}
 
 // Select calls the field on the instance specified by the selector
 func (r Instance[T]) Select(ctx context.Context, s *Server, sel Selector) (Typed, *call.ID, error) {
-	inputArgs, newID, doNotCache, err := r.preselect(ctx, s, sel)
+	inputArgs, newID, doNotCache, skipDedupe, err := r.preselect(ctx, s, sel)
```
jedevc, line 360: nit: lots of return args, maybe we should assemble them into a private struct?

engine/cache/cache.go
```diff
@@ -137,6 +139,7 @@ func (c *cache[K, V]) GetOrInitialize(
 func (c *cache[K, V]) GetOrInitializeWithCallbacks(
 	ctx context.Context,
 	key K,
+	skipDedupe bool,
```
jedevc, line 142: Would be nice to have a test for the new `skipDedupe` arg if feasible, not a blocker since we have the integration test.

jedevc:
LGTM, just wanted to check though - why specifically are only nested execs causing this? I get the explanation at the top of the PR, but not sure why it's specific to the nested exec.

jedevc (approved):

engine/cache/cache.go
sipsma: yep, added!

dagql/objects.go
sipsma: agreed, fixed that

sipsma:
> LGTM, just wanted to check though - why specifically are only nested execs causing this? I get the explanation at the top of the PR, but not sure why it's specific to the nested exec.

@jedevc Only nested execs (including function calls) talk back to the engine API, including their session/client IDs in the requests. So if their session goes away due to the scenario in the description the engine gets very confused; there's a client from a removed session coming in making requests "from the dead" 💀

PR #10267, opened 2025-04-25
docs: Add automatic labels for docs PRs

vikram-dagger:
I noticed @jedevc manually applying labels to docs PRs and was wondering if this could be a solution: https://github.com/actions/labeler

Note: I'm not quite sure how to test it without first merging it

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,8 +1,8 @@
 # Add 'area/documentation' label to any PR where
 # - any file in the `docs` directory is changed
-# - the head branch name starts with `docs` or has a `docs:` section in the name
+# - the head branch name starts with `docs` or has a `docs` section in the name
 area/documentation:
 - any:
   - changed-files:
     - any-glob-to-any-file: 'docs/*'
-  - head-branch: ['^docs', 'docs:']
+  - head-branch: ['^docs', 'docs']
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,8 @@
+# Add 'area/documentation' label to any PR where
+# - any file in the `docs` directory is changed
+# - the head branch name starts with `docs` or has a `docs` section in the name
+area/documentation:
+- any:
+  - changed-files:
+    - any-glob-to-any-file: 'docs/*'
```
jedevc, line 7: This would catch any automatic docs updates from API changes - ideally, we wouldn't include those with `area/documentation`?

I think using the PR title would probably be the easiest, but looks like this isn't a supported feature :cry: https://github.com/actions/labeler/issues/809

vikram-dagger: > This would catch any automatic docs updates from API changes - ideally, we wouldn't include those with `area/documentation`?

I see. Would it matter much though? Maybe a few incorrectly-labelled PRs might be OK if it saves us time on manual labelling?

We could also try to reduce the incidence of false positives, with exclusions for the API docs (`!` is supported).

> I think using the PR title would probably be the easiest, but looks like this isn't a supported feature 😢 [actions/labeler#809](https://github.com/actions/labeler/issues/809)

Yes, I checked on this as well and it doesn't seem to be supported.

vikram-dagger: > The problem is, I think this would probably be most engine PRs (we change the APIs subtly in probably most PRs?) 

I didn't understand this @jedevc - can you give me an example? Would it help if we restricted the path to `docs/current_docs` only?

Also btw, I saw that there is an open PR in this action's repo to support filtering on the PR title, so that will help!

PR #10271, opened 2025-04-25
fix: incorrect dagop caching of git ops

jedevc:
Woops, #9980 had some issues.

1. I have suddenly remembered why this was important: https://github.com/dagger/dagger/pull/9980#discussion_r2039628759. This *needs* to be lifted up, or the dagql cache key becomes permanently tainted with the client/session ids we used to get here. A ref is "frozen" in time, and so gets detached from these - this was causing us to not cache certain operations.
2. The custom op cache key calculation had a slight bug - we were accidentally including IDs *twice* in the cache key, once with itss ID digest, and again once with its encoding (which *wasn't always the same*) - so we weren't ever caching git checkouts.
3. #9682 added a sanity check condition that means that all SSH urls passed to `git` *must* have a socket passed with them. Effectively this means that we can't pass it directly through to `GitRef.tree` - because we can't get here without the socket. *technically* this is a regression - but #9980 started depending on that behavior, by resolving the ref earlier, before we get to `tree`. I think it makes sense to officially retire this behavior, it's been deprecated for almost a year, and wasn't a widely used feature anyways at the time.
	- I did spend some time attempting to keep it working - it's really not worth the effort IMO, since it prevents us from cleaning up the git code properly here (and would probably introduce weird caching inconsistencies).
	- We don't even have tests for this case, so :shrug:

Finally, I've added a telemetry test for a git clone, which should make sure that `tree` has it's little caching label. We weren't checking this before.

Putting into the v0.18.5 milestone, since by itself #9980 is a perf *regression*.

### DIFF
--- a/core/schema/git.go
+++ b/core/schema/git.go
@@ -462,10 +462,6 @@ type tagsArgs struct {
 }
 
 func (s *gitSchema) tags(ctx context.Context, parent dagql.Instance[*core.GitRepository], args tagsArgs) (dagql.Array[dagql.String], error) {
-	if !core.DagOpInContext[core.RawDagOp](ctx) {
-		return DagOp(ctx, s.srv, parent, args, s.tags)
-	}
-
 	var patterns []string
 	if args.Patterns.Valid {
 		for _, pattern := range args.Patterns.Value {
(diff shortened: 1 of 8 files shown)


core/schema/git.go
```diff
@@ -450,10 +462,6 @@ type tagsArgs struct {
 }
 
 func (s *gitSchema) tags(ctx context.Context, parent dagql.Instance[*core.GitRepository], args tagsArgs) (dagql.Array[dagql.String], error) {
-	if !core.DagOpInContext[core.RawDagOp](ctx) {
-		return DagOp(ctx, s.srv, parent, args, s.tags)
-	}
```
jedevc, line 455: There's no reason to dagop cache this. The ID here isn't stable (it'll be mixed in with client/session most of the time) - so this actually just wastes buildkit cache space.

dagql/idtui/viztest/main.go
```diff
@@ -431,3 +431,9 @@ func (*Viztest) List(ctx context.Context, dir *dagger.Directory) (string, error)
 	}
 	return strings.Join(ents, "\n"), nil
 }
+
+func (*Viztest) GitReadme(ctx context.Context, remote string) (string, error) {
+	result, err := dag.Git(remote).Head().Tree().File("README.md").Contents(ctx)
```
sipsma, line 436: not a blocker on merge here, but seems like this may get annoying to update whenever we change the README

sipsma (approved):

PR #10273, opened 2025-04-25
don't spam the output when sending telemetry upload fails

marcosnils:
handles better the case where the client gets an error when sending
telemetry. If the --debug flag is supplied, all errors are displayed as
they were before this change

before:

image

after:

image

### DIFF
--- a/dagql/idtui/frontend_pretty.go
+++ b/dagql/idtui/frontend_pretty.go
@@ -1945,7 +1945,7 @@ func (fe *frontendPretty) handlePromptString(ctx context.Context, message string
 
 func handleTelemetryErrorOutput(w io.Writer, to *termenv.Output, err error) {
 	if err != nil {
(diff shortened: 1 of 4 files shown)


dagql/idtui/frontend_pretty.go
```diff
@@ -1942,6 +1943,13 @@ func (fe *frontendPretty) handlePromptString(ctx context.Context, message string
 	}
 }
 
+func handleTelemetryErrorOutput(w io.Writer, to *termenv.Output, err error) {
+	if err != nil {
+		fmt.Fprintf(w, "%s - %s\n(%s)\n", to.String("WARN").Foreground(termenv.ANSIYellow), "failures detected while emitting telemetry. trace information inocomplete", err.Error())
```
sipsma, line 1948:
```suggestion
		fmt.Fprintf(w, "%s - %s\n(%s)\n", to.String("WARN").Foreground(termenv.ANSIYellow), "failures detected while emitting telemetry. trace information incomplete", err.Error())
```

PR #10286, opened 2025-04-28
fix: ensure local commit check actually works

jedevc:
This checks wasn't actually working as desired - we weren't actually checking that the commit was local (we need `^{commit}` and `--verify` for that), but the functionality wasn't entirely broken, because we were missing the trailing whitespace (so `skipFetch` was never true, so we lost a little optimization here).

While we're here, we need to make this command non-fatal - a failure here doesn't actually need to be marked as a failure in OTEL, so we can skip past it (as noted by @shykes in discord here: https://discord.com/channels/707636530424053791/1003718839739105300/1365847042353664020)

### DIFF
--- a/core/git.go
+++ b/core/git.go
@@ -508,7 +508,9 @@ func (ref *RemoteGitRef) Tree(ctx context.Context, srv *dagql.Server, discardGit
 
 		// skip fetch if commit already exists
 		doFetch := true
-		if res, err := git.Run(ctx, "rev-parse", ref.FullRef); err == nil && string(res) == ref.Commit {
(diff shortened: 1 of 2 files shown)


sipsma (approved):
If possible to add test would be good to have obviously, but if not feasible no blocker. LGTM

jedevc:
Mm, agreed tests would be nice, but struggling to think about how - will keep thinking, and if I have a good idea, will do a follow-up.

Wondering if *maybe* we should be instrumenting special "counters" or similar into our ops, so we can check how often specific branches of code are hit (I *think* buildkit does something like this?). But yeah, it's weird, because we're wanting to test edge cases in the caching behavior.

PR #10292, opened 2025-04-28
docs: Use autocapture for video player

vikram-dagger:
Instead of manually capturing events, this commit switches to use PostHog's autocapture to automatically track clicks on the play/pause button, with additional metadata for the video file and title. This also eliminates video click events appearing twice in the PostHog activity log.

https://posthog.com/docs/product-analytics/autocapture


jpadams (approved):
Deleting code is the best!

PR #10299, opened 2025-04-29
fix: typescript clien gen conflict with module

TomChv:
I noticed during a demo that if a module source is `.` and  implements functions and also generate a client, the `sdk/index.ts` file will not expose the generated client in the `sdk` folder because it doesn't expect it to be there.

I changes the logic in the runtime to handle this specific edge case and updated integrations tests.

This behaviour could be handled in the codegen binary ideally but it's a quick fixes for now.

### DIFF
--- a/core/integration/client_generator_test.go
+++ b/core/integration/client_generator_test.go
@@ -415,37 +415,39 @@ main()
 						With(daggerExec("init", "--name=test", "--sdk=go", "--source=.dagger")).
 						WithNewFile(".dagger/main.go", `package main
 		
-		type Test struct{}
+import "context"
+
+type Test struct{}
 		
-		func (t *Test) Hello() string {
-			return "hello"
-		}
+func (t *Test) Hello(ctx context.Context) (string, error) {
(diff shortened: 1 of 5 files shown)


core/integration/client_generator_test.go
```diff
@@ -476,11 +478,11 @@ export class Test {
...
 async function main() {
-    await connection(async () => {
-      const res = await dag.test().hello()
+  await connection(async () => {
+    const res = await dag.test().hello()
 
-      console.log("result:", res)
-    })
+    console.log("result:", res)
+  })
```
cwlbraa, line 485: nit: fix your editor's indent settings

cwlbraa (approved):

core/integration/client_generator_test.go
TomChv: The string snippet are not handled by my editor sadly haha so I have to do it manually every time! Good catch

TomChv: Here I'm actually fixing a indent issue from a previous PR, normally we use 2 space, not 4 :)

cwlbraa: nit: fix someone else's editor indent settings lol

Issue #10298, opened 2025-04-29
🐞 Arguments missing from `call <function> --help`


kpenfound:
### What is the issue?

I have arguments missing from the arguments list in `dagger call <function> --help`. The arguments show properly in dagger shell `.help <function>`.

### Dagger version

0.18.5

### Steps to reproduce

Module defined here: https://github.com/kpenfound/dagger-programmer/blob/main/src/dagger_programmer/main.py

Run:

```
dagger -m github.com/kpenfound/dagger-programmer call translate --help
```

The arguments list is missing required arg "mod":

```
ARGUMENTS
      --language string          The language to translate the module to [required]
      --dependencies Directory   Dependencies required for the module (default [])
```

In shell `dagger -m github.com/kpenfound/dagger-programmer`:

```
✔ .help translate 0.0s
Returns a dagger module in the specified language translated from the provided module

USAGE
  translate <mod> <language> [options]

REQUIRED ARGUMENTS
  language string   The language to translate the module to
  mod Module        The dagger module to translate

OPTIONAL ARGUMENTS
  --dependencies []Directory   Dependencies required for the module (default: [])

RETURNS
  Directory - A directory.

  Use "translate <mod> <language> | .help" for more details.
```

kpenfound:
Actually its worse. The required arg `--mod` does not work with `dagger call`. It only works in shell

```
> dagger -m github.com/kpenfound/dagger-programmer call translate --mod github.com/shykes/hello --language python
✔ connect 0.2s
✔ load module 0.9s
✔ parsing command line arguments 0.0s

Error: input: query:
query{daggerProgrammer{translate(mod:"github.com/shykes/hello", language:"python"){id:sync}}}

error: parse selections: parse field "translate": init arg "mod" value as core.DynamicID (ModuleID!) using core.DynamicID: decode "Module" ID: failed to decode base64: illegal base64 data at input byte 6


Full trace at https://dagger.cloud/kpenfound/traces/40291159cd25e98c53af2f1b3506e2af
```

kpenfound:
The argument is also missing from `--help` on 0.18.4

sipsma:
At first glance I wouldn't be surprised if the CLI's `--mod` (short form `-m`) flag conflicts and results in this. cc @helderco if that is indeed the problem, was there any recent change that would have broken this or long-standing?

kpenfound:
Its possible that this has always been broken since I've only used this particular module in shell

helderco:
[...]

Issue #7032, opened 2024-04-05
Function args conflict with CLI args


jedevc:
Take the following go module:

```go
package main

import (
	"context"
	"strings"
)

type Playground struct{}

func (m *Playground) Foo(ctx context.Context, progress string) (string, error) {
	return strings.ToUpper(progress), nil
}
```

On call:

```
$ dagger-dev call foo --progress=plain
Connected to engine ff7f91f8a66b (version 4b825dc642cb6eb9a060e54bf8d69288fbee4904)
✔ connect 0.2s
✘ initialize 3.5s
! flag already exists: progress
  ✔ ModuleSource.resolveFromCaller: ModuleSource! 0.3s
  ✔ ModuleSource.asModule: Module! 1.3s
  ✔ Module.initialize: Module! 1.6s
    ✔ exec go build -o /runtime . 1.1s

Error: flag already exists: progress                                                                                                                                                                                                                                                                        
Run 'dagger call foo --help' for usage.                                                                                                                                                                                                                                                                     
```

The `--progress` flag is already defined by dagger, so we can't create it again.

Proposal: dagger-specific flags should appear *before* the `call` arg, and shouldn't be allowed afterwards - this would be a breaking change, since dagger args after the `call` wouldn't be permitted.

cc @helderco

Related:
- https://github.com/sagikazarmark/daggerverse/issues/64

helderco:
Yeah, I've talked about this with @grouville in the context of (re: global flags):
- https://github.com/dagger/dagger/issues/6943

I've actually wanted to do what you suggest for a long time, and it's how I use it, but then see people using global flags at the end and wondered if not having that would be confusing. However, of course it can lead to conflicts, exactly what I talked with @grouville about.

helderco:
We could also possibly use a `--` to separate flags from functions which would be optional, in case there's a conflict.

helderco:
I think we should start with my suggestion in last comment and let users set `--` to split function args from the rest in case there's a conflict but not otherwise enforcing persistent flags before the module's constructor which is a breaking change and also doesn't solve the whole issue (`call` specific flags need to be set after `call` which could still conflict with constructor arguments like `--output` and `--json`).

PR #10305, opened 2025-04-30
cli: fix regression in dagger call when a function argument conflicts with an existing flag

helderco:
Closes https://github.com/dagger/dagger/issues/10298 [^1]

This fixes a regression introduced in https://github.com/dagger/dagger/pull/7417 where conflicting flags no longer throw an error in `dagger call`. 

Full explanation in https://github.com/dagger/dagger/issues/10298#issuecomment-2840466435.

[^1]: Fixes the missing error to clear confusion but the underlying problem of supporting a conflicting flag is in another duplicate [issue](https://github.com/dagger/dagger/issues/7032).

### DIFF
--- a/.changes/unreleased/Fixed-20250430-111645.yaml
+++ b/.changes/unreleased/Fixed-20250430-111645.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Return an error when a function argument conflicts with a persistent flag in `dagger call`
+time: 2025-04-30T11:16:45.32856Z
+custom:
+    Author: helderco
+    PR: "10305"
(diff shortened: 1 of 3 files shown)


jedevc:
@helderco I think this should also close https://github.com/dagger/dagger/issues/7032?

helderco:
> @helderco I think this should also close #7032?

No, that's still unresolved. It's just that right now the error is being ignored but it wasn't meant to. This PR just makes sure the error propagates.

I think the blocker for #7032 was just postponing the decision to make a breaking change but we can actually support `--` in a compatible way for when there is a conflict, i.e.., without always enforcing persistent flags to be before `call` (which wouldn't save you from a conflict between constructor args and `call` flags like `--output` and `--json` anyway).

Issue #10156, opened 2025-04-14
flake: `TestJava` fails with `in epilog non whitespace content is not allowed but got 0`


sipsma:
This has been happening a lot lately.

Example: https://v3.dagger.cloud/dagger/traces/e8c0fc90b5ecaa75eeeba4ce885ead5b?span=4c77921594061dd0&logs#4c77921594061dd0:L155

Primary error seems to be:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:3.1.2:install (default-install) on project dagger-sdk-parent: Failed to install metadata io.dagger:dagger-sdk-parent/maven-metadata.xml: Could not parse metadata /root/.m2/repository/io/dagger/dagger-sdk-parent/maven-metadata-local.xml: in epilog non whitespace content is not allowed but got 0 (position: END_TAG seen ...</metadata>\n0... @13:2)  -> [Help 1]
```

Which I don't know what to make of

sipsma:
Variation on error message:
```
61  : [6.0s] | [INFO] ------------------------------------------------------------------------
61  : [6.0s] | [INFO] BUILD FAILURE
61  : [6.0s] | [INFO] ------------------------------------------------------------------------
61  : [6.0s] | [INFO] Total time:  1.963 s
61  : [6.0s] | [INFO] Finished at: 2025-04-14T18:57:12Z
61  : [6.0s] | [INFO] ------------------------------------------------------------------------
61  : [6.0s] | [ERROR] Failed to execute goal on project construct: Could not resolve dependencies for project io.dagger.modules.construct:construct:jar:1.0-SNAPSHOT
61  : [6.0s] | [ERROR] dependency: io.dagger:dagger-java-sdk:jar:0.16.4-my-module-module (compile)
61  : [6.0s] | [ERROR]  Could not find artifact io.dagger:dagger-java-sdk:jar:0.16.4-my-module-module in central (https://repo.maven.apache.org/maven2)
61  : [6.0s] | [ERROR] dependency: io.dagger:dagger-java-annotation-processor:jar:0.16.4-my-module-module (provided)
61  : [6.0s] | [ERROR]  Could not find artifact io.dagger:dagger-java-annotation-processor:jar:0.16.4-my-module-module in central (https://repo.maven.apache.org/maven2)
```

[...]

PR #10307, opened 2025-04-30
sdk/java: lock concurrent access to maven cache

eunomie:
Set the sharing mode of the maven cache volume to lock.
This ensure there's no concurrent write to the cache that should limit
the risk of corruption of maven files, as concurrent access to the maven
local repository is not supported.

Error looks like:

    [ERROR] Failed to execute goal
    org.apache.maven.plugins:maven-install-plugin:3.1.2:install
    (default-install) on project dagger-sdk-parent: Failed to install
    metadata io.dagger:dagger-sdk-parent/maven-metadata.xml:
    Could not parse metadata
    /root/.m2/repository/io/dagger/dagger-sdk-parent/maven-metadata-local.xml:
    in epilog non whitespace content is not allowed but got 0 (position:
    END_TAG seen ...</metadata>\n0... @13:2)

To lock the cache might have an impact in term of performances in case
of concurrent access, but it will keep good performances on single
access and ensure there's no corruption anymore.

This should help to solve #10156

### DIFF
--- a/sdk/java/runtime/main.go
+++ b/sdk/java/runtime/main.go
@@ -33,6 +33,7 @@ type JavaSdk struct {
 type moduleConfig struct {
 	name    string
 	subPath string
+	dirPath string
 }
 
 func (c *moduleConfig) modulePath() string {
@@ -141,7 +142,7 @@ func (m *JavaSdk) buildJavaDependencies(
 	}
 	return ctr.
 		// Cache maven dependencies
-		WithMountedCache("/root/.m2", dag.CacheVolume("sdk-java-maven-m2")).
+		WithMountedCache("/root/.m2", dag.CacheVolume("sdk-java-maven-m2"), dagger.ContainerWithMountedCacheOpts{Sharing: dagger.CacheSharingModeLocked}).
 		// Mount the introspection JSON file used to generate the SDK
 		WithMountedFile("/schema.json", introspectionJSON).
 		// Copy the SDK source directory, so all the files needed to build the dependencies
@@ -386,9 +387,14 @@ func (m *JavaSdk) setModuleConfig(ctx context.Context, modSource *dagger.ModuleS
 	if err != nil {
 		return err
 	}
+	dirPath, err := modSource.LocalContextDirectoryPath(ctx)
+	if err != nil {
+		return err
+	}
 	m.moduleConfig = moduleConfig{
 		name:    modName,
 		subPath: subPath,
+		dirPath: dirPath,
 	}
 
 	return nil


TomChv (approved):

helderco:
Most of our tests use the same name for modules. You could use the module source digest instead.

helderco:
There's also options on cache volumes to make it private and not shared for example.

eunomie:
> Most of our tests use the same name for modules. You could use the module source digest instead.
> There's also options on cache volumes to make it private and not shared for example.

🤔 yes, if I just use the name it will not solve all cases.

Digest might be wrong as if the module content change the cache will be invalid and that's not what we want.

I'll have a look at the `namespace` and to find a more precise way to specify the volume.

eunomie:
I've updated it to use the module path/sub-path. That should cover most of the cases I guess.
That way the cache is only tight to a local directory, and not shared across multiple modules.

I wonder if I shouldn't put the path as the namespace, maybe that could be more elegant.

helderco:
> I've updated it to use the module path/sub-path. That should cover most of the cases I guess. That way the cache is only tight to a local directory, and not shared across multiple modules.
> 
> I wonder if I shouldn't put the path as the namespace, maybe that could be more elegant.

It's like the name. Most of our tests create a module at `/work/test`. Same path and same name. Even concurrently.

Have you tried the sharing mode on the monted cache like I said? For example:

https://github.com/dagger/dagger/blob/37955449d1b32196d5be0f98c412c7bb5d7ff772/core/integration/provision_test.go#L194-L196

eunomie:
> Have you tried the sharing mode on the monted cache like I said? For example:

Ho right, I only tried with the cache volume option, not the one on the mount cache function.

So yes, that looks better with lock mode. It will keep good performances when single access, and ensure we don't have corruption. It might slow down a bit when multiple concurrent access, but that should still be better than without cache so I think that's worth it.

sdk/java/runtime/main.go
```diff
@@ -141,7 +142,7 @@ func (m *JavaSdk) buildJavaDependencies(
 	}
 	return ctr.
 		// Cache maven dependencies
-		WithMountedCache("/root/.m2", dag.CacheVolume("sdk-java-maven-m2")).
+		WithMountedCache("/root/.m2", dag.CacheVolume("maven-m2", dagger.CacheVolumeOpts{Namespace: "sdk-java"}), dagger.ContainerWithMountedCacheOpts{Sharing: dagger.CacheSharingModeLocked}).
```
jedevc, line 145: I would probably not set the namespace - I actually hadn't realized it was possible to set manually :cry: 

It's intended as a unique value per module (so each calling module gets a unique value of it).

jedevc (approved):

PR #10308, opened 2025-04-30
Add secretprovider tests

jedevc:
We don't have any secretprovider tests - which feels scary :scream: 

So, this PR adds them:
- The basic providers, `env`/`file`/`cmd` are easy
- The `vault` test was originally written by @kpenfound in [`62fcd099093caf0645fe74f189b4bde6825281a5..8081eaf5f4e38f3fb6a2dcc2db24965d896fd658`#diff-ef6449f08e](https://github.com/dagger/dagger/compare/62fcd099093caf0645fe74f189b4bde6825281a5..8081eaf5f4e38f3fb6a2dcc2db24965d896fd658#diff-ef6449f08e3ecb813cac28681c34d89c5863e191d44a45d9502ea09653172d6a)
- The `libsecret` test is an absolute pain, but given it's more obscure, is definitely important to test.
- :warning: The `op` test doesn't exist yet, I'm not really sure we want the tests to be dependent on an external service.


kpenfound (approved):
Nice!

PR #10309, opened 2025-04-30
refactor: decouple sdk interface

TomChv:
Add function implementation checking in the engine so it is aware of which function the SDK implements.
Move sdk related code into its own package except the interface definition bececause it's used in `core`.
Improve sdk documentation and code readability.
Make `requiredClientGenerationFiles` optional.
Put the Golang sdk implementation into its own file. Put some resolving logic from schema/module_source.go into core/module_source.go to avoid circular dependency with the new `sdk` package.

Note: for now this does not make the implementation of SDK function optional but this should be done in a follow up since it needs changes in the CLI to set what is expected to be implemented depending on the command execution (`dagger develop` need Codegen | `dagger call` needs Runtime etc...)


core/sdk.go
```diff
@@ -0,0 +1,197 @@
...
+
+The generated client is standalone and can be used in any project,
+even if no source code module is available.
+
+This interface MUST be implemented to support standalone client oriented
+features (`dagger client install`).
+*/
+type ClientGenerator interface {
+	/*
+		RequiredClientGenerationFiles returns the list of files that are required
```
jedevc, line 21: nit: should just be indented once?

core/sdk/code_generator_module.go
jedevc, line 1: These files can be named `module_...`, then they'll appear next to each other.

core/sdk.go
jedevc, line 1: Can this be moved to the `sdk` package as well?

core/modulerefs.go
```diff
@@ -314,3 +314,31 @@ func (p *parsedGitRefString) getGitRefAndModVersion(
...
+			return v, nil
+		}
+	}
+	return "", fmt.Errorf("unable to find version %s", match)
+}
+
+func isSemver(ver string) bool {
+	re := regexp.MustCompile(`^v[0-9]+\.[0-9]+\.[0-9]+$`)
+	return re.MatchString(ver)
+}
```
jedevc, line 344: This function can be removed. `semver.IsValid` is the way to do this :grin:

core/sdk.go
TomChv: > Can this be moved to the sdk package as well?

Sadly no, because `core/modulesource` has `SDK` in its field and `sdk/` package depends on core :/

core/sdk.go
TomChv: The formatter is auto formatting with 2 :/

core/modulerefs.go
TomChv: I replaced isSemver with a call to `semver.IsValid` and removed the tests for isSemver + removed the function

core/sdk/code_generator_module.go
TomChv: done!

core/modulesource_test.go
```diff
@@ -47,3 +51,390 @@ func TestGitModuleSourceSymbolic(t *testing.T) {
 		})
 	}
 }
+
+func TestMatchVersion(t *testing.T) {
```
eunomie, line 55: Just wondering if this test shouldn't be in a `core/modulerefs_test.go`?
This is mostly about conventions and keeping the test as close as possible to the code itself, but also helps IDEs like Goland that groups test files with their source code
image

TomChv: Done in [b06969c](https://github.com/dagger/dagger/pull/10309/commits/b06969c102812b7feede5d7a4850acc20e368101)

jedevc (approved):
LGTM once tests pass :grin:

TomChv:
Failed tests isn't related to this Pr: https://v3.dagger.cloud/dagger/traces/3725485313b25b333d45a6176cc27f94?listen=0c0b6ad3e0fde1c8&listen=e630495b1ba20fae&listen=d8d9824faa6188da&listen=64ad47fb98519a8d&listen=bce999b99eef0926&listen=e002e856fab41994

PR #10311, opened 2025-04-30
cache secrets by plaintext; make cache key configurable

sipsma:
# Problem
Secrets have been cached by their "name", which for `setSecret` was the actual name and for URI-based secrets was the URI. This caching behavior was inherited from buildkit.

This impacts when the cache is busted for operations that include the secret.

For example w/ `SetSecret`
```go
dag := dagger.Connect(ctx)
secret1 := dag.SetSecret("FOO", "123")
ctr1 := dag.Container().
  From("alpine").
  WithSecretVariable("FOO", secret1).
  WithExec([]string{"true"}).
  Sync(ctx)

secret2 := dag.SetSecret("FOO", "abc")
// ctr2 is cached with ctr1 because secret1 and secret2 
// have the same name, even though they have different values
ctr2 := dag.Container().
  From("alpine").
  WithSecretVariable("FOO", secret2).
  WithExec([]string{"true"}).
  Sync(ctx)
```

For example w/ `Secret` (uri based)
```go
os.SetEnv("FOO", "123")
dag1 := dagger.Connect(ctx)
secret1 := dag1.Secret("env://FOO")
ctr1 := dag1.Container().
  From("alpine").
  WithSecretVariable("FOO", secret1).
  WithExec([]string{"true"}).
  Sync(ctx)

os.SetEnv("FOO", "abc")
dag2 := dagger.Connect(ctx)
secret2 := dag2.Secret("env://FOO")
// ctr2 is cached with ctr1 because secret1 and secret2 
// have the same name (URI in this case), even though they have different values
ctr2 := dag2.Container().
  From("alpine").
  WithSecretVariable("FOO", secret2).
  WithExec([]string{"true"}).
  Sync(ctx)
```

For many use cases, this behavior of caching by the name rather than plaintext value works out because users don't want the a change in plaintext to bust the cache, e.g.
* sets of tokens w/ different values that are all valid auth for the same resource (i.e. different users with their own tokens for the same resource)
* secrets that frequently rotate in value

However, there are also cases where this is quite surprising and users want the cache key to be derived from the plaintext value (i.e. a change in plaintext busts the cache and different secrets with the same plaintext are cached together).
* E.g. different clients source a token from the same env var (`env://AUTH_TOKEN`) but each client has different values that are not equivalent
* For `SetSecret` there was always a fallback of giving the secret a unique name (and thus cache key)
* **However, for URI-based secrets that's not always an option** unless you have users change environment variables, secret names in external providers, etc.
* The above problem was exacerbated by dagql engine-wide caching since it made it much easier for different clients to get cache hits between each other, including those using meaningfully different secrets that have the same URI

Basically, there's different use cases that call for drastically different behavior in terms of how/when secrets bust cache, specifically in terms of whether the secret cache key should be based on the plaintext or a different user-supplied string.

# Solution

The fix here:
1. Changes the default cache key for secrets to be derived from the plaintext. Different plaintexts bust cache for operations including the secrets
   * This is a breaking change
2. `Query.secret` (the one used for URI-based secrets) has a new optional argument `cacheKey`. If it's set, it will override the plaintext-based cache key and be used instead.
   * This enables users to selectively obtain the previous behavior

Overall, this behavior feels the most intuitive in the default case, especially for URI based secrets, which we consider the new-and-improved replacement for `setSecret`. The fact that URI-based secrets all share cache across clients just based on the URI string doesn't really make much sense outside of the "historical" reasons. Then for users who want to opt into cache control over secrets, the new `cacheKey` argument is available to them.

Going back to the previous `Secret` URI example, the new behavior is:

For example w/ `Secret` (uri based)
```go
os.SetEnv("FOO", "123")
dag1 := dagger.Connect(ctx)
secret1 := dag1.Secret("env://FOO")
ctr1 := dag1.Container().
  From("alpine").
  WithSecretVariable("FOO", secret1).
  WithExec([]string{"true"}).
  Sync(ctx)

os.SetEnv("FOO", "abc")
dag2 := dagger.Connect(ctx)
secret2 := dag2.Secret("env://FOO")
// ctr2 is NOT cached with ctr1 because secret1 and secret2 
// have different plaintexts
ctr2 := dag2.Container().
  From("alpine").
  WithSecretVariable("FOO", secret2).
  WithExec([]string{"true"}).
  Sync(ctx)
```

For use cases that call for different secrets having the same cache key (closer to the old behavior, but more configurable now), below are examples of usage of the new `cacheKey` arg.

In code:
```go
os.SetEnv("FOO", "123")
secret1 := dag.Secret("env://FOO", dagger.SecretOpts{CacheKey: "my-tokens"})

ctr1 := dag.Container().
  From("alpine").
  WithSecretVariable("FOO", secret1).
  WithExec([]string{"true"}).
  Sync(ctx)

os.SetEnv("BAR", "abc")
secret2 := dag.Secret("env://BAR", dagger.SecretOpts{CacheKey: "my-tokens"})
// ctr2 is cached with ctr1 because secret1 and secret2 
// have the same cacheKey, even though they have different values + URIs
ctr2 := dag.Container().
  From("alpine").
  WithSecretVariable("FOO", secret2).
  WithExec([]string{"true"}).
  Sync(ctx)
```

In dagger shell:
```
⋈ some-function --secret-arg $(secret env://FOO --cache-key my-tokens)
```

In dagger call:
```
dagger call some-function --secret-arg env://FOO?cacheKey=my-tokens
```

# Security

We haven't stored secret plaintexts on disk and still won't after this change.

However, we do need to be careful deriving cache keys from plaintexts since those cache keys are sent as IDs all over (OTEL progress primarily, but also logged various places) and the cache keys are also written to the local disk as part of the cache.

Specifically, we don't want to make it easy to perform offline brute forcing attempts of secret cache keys to find their plaintext. This is very similar to the problem web backends have around password hashing.

The implementation here uses `[argon2`](https://en.wikipedia.org/wiki/Argon2) (specifically the implementation in the go stdlib) to hash the secrets and turn them into cache keys.
* This allows us to tune the hashing to be expensive enough to make brute-forcing infeasible but also not so slow it noticeably impacts the engine.
* e.g. on my laptop the hash takes 10-30ms, so it's not a bottleneck, but that's still 10s of millions times slower than the time it takes to try 1 sha256 hash on a modern GPU (<1 ns)

The main difference from the more typical password-hashing use-case is that we *don't* want a unique salt per secret since we need deterministic cache keys. Instead, we use a salt unique to the engine instance as a whole (stored on the local disk along-side the cache).
* There will be some thinking required to get this working for remote cache transferred between engines (where the salt is different and thus secret cache keys), but secrets are already tricky in that situation anyways, so this doesn't make the situation much worse. These are all implementation details that can be tweaked or reassessed at that time if needed.

---

TODOs:
- [x] More new tests
- [x] Add support for `cacheKey` arg being included in secret URI (to support `dagger call` mainly)
- [x] Consider whether `SetSecret` should have the same change in default cache key (for consistency)
   * After thinking about this some more, I'm not going to change `SetSecret`. The name provided to `SetSecret` is essentially the same as the new `cacheKey` arg to `Secret`, so I think keeping the current behavior is overall the least confusing.
- [x] Fill in doc strings
- [x] Changelog for breaking change
   - additional changelog for vault secret provider panic fix found during this: https://github.com/dagger/dagger/pull/10311/commits/ba695546ed1d1b87079576ae38064307b9bd00cf

Follow-ups:
- [ ] Update docs


cmd/dagger/flags.go
```diff
@@ -448,6 +449,12 @@ func (v *secretValue) Set(s string) error {
 		s = secretSource + "://" + val
 	}
 
+	const cacheKeyPrefix = ",cacheKey="
```
jedevc, line 452: Agreed, I don't love this syntax either. Dunno, I feel like maybe it might be worth struggling through the pain of getting secret provider as *full URIs* so this could work?

But no real other suggestions :cry:

jedevc (approved):

cmd/dagger/flags.go
sipsma: It's mainly the `cmd://echo blah blah?cacheKey=my-tokens` situation that's hard to deal with, since that's not a valid URI (it complains the host cannot have spaces, which makes sense).

I *did* just realize that `cmd:///echo blah blah?cacheKey=my-tokens` (3 `/` instead of 2) is valid because the command is technically part of the path rather than host and spaces get url encoded)

I wouldn't want users to have to type 3 `/`s though, could special case it I guess to automatically put an extra `/`. Also some potential weirdness about disambiguating `/command` from `command`, but getting outrageously obscure at that point.

I'll try something out quick here and see if it feels any better/worse than the current solution

sipsma: Got this working w/ a slightly different approach that required an extra chunk of careful query param parsing in cmd/dagger/flags.go, but added integ+unit test coverage to cover it

PR #10313, opened 2025-05-01
docs(deno): deno is selected as runtime if deno.lock is present

rawkode:
This is implemented [here](https://github.com/dagger/dagger/blob/3af6558837398cd3377013d4a54712441eb080f5/sdk/typescript/runtime/config.go#L333).


rawkode:
While the code does check for `deno.lock`, in the absence of `deno.json`; the deno support doesn't actually work with a `package.json` and `deno.lock` combination (which deno itself supports).

This is due to `dagger develop` requiring the `deno.json` to configure the unstable flags and compiler options.

Perhaps we should update the code to only support `deno.json` and remove the `deno.lock` check?

cc @TomChv

TomChv (approved):
LGTM, you're right, we need the `deno.json` or we cannot configure the project to run with Dagger !

Issue #7721, opened 2024-06-22
✨ Build-Kit Build reproducibility in Dagger


danielwegener:
### What are you trying to do?

I'd love to have binary reproducible builds (motivation: https://reproducible-builds.org/) in dagger.  

### Why is this important to you?

Beside its nice security properties, I'd also love the fact, that I could use it for stable image pinning and actual change detection in k8s deployments (as image digests would only change if the underlying application does change).

### How are you currently working around this?

Directory.withTimestamp helps, but sometimes required to remount a whole container filesystem.

TomChv:
@sipsma @vito @jedevc I think this one is for one of you ;)

danielwegener:
Oh sorry just saw that this was kind of already envisioned in https://dagger.io/blog/reproducible-builds#:~:text=%2C%20fixedTimestamp)-,The%20Future,-Containerization%2C%20resource%20pinning
Is this feature somewhere on the roadmap or would contributions in this direction be welcomed?

hugochinchilla:
this would be nice to have, I'm having to patch my Dockerfiles to ensure each run step forces all mtimes to be a prefixed value and it's getting nasty

PR #10318, opened 2025-05-02
feat: support SOURCE_DATE_EPOCH and rewrite-timestamp

danielwegener:
As described on https://github.com/moby/buildkit/blob/master/docs/build-repro.md, buildkit supports the reproducible-builds variable SOURCE_DATE_EPOCH and a rewrite-timestamp build-arg that uses the epoch as timestamp for actions on the filesystem.

This PR follows the idea outlined in #7721 by:
- Reading SOURCE_DATE_EPOCH at the client and propagate it as ClientMetadata
- Propagate SOURCE_DATE_EPOCH from ClientMetadata to ExecutionMetadata
- Propagate the env SOURCE_DATE_EPOCH to nestedClients and Process.Env 
- Add rewriteTimestamp option to terminal publish, export and asTarball operations

According to the builkit-doc, this would affect:
- the created timestamp in the [OCI Image Config](https://github.com/opencontainers/image-spec/blob/main/config.md#properties)
- the created timestamp in the history objects in the [OCI Image Config](https://github.com/opencontainers/image-spec/blob/main/config.md#properties)
- the org.opencontainers.image.created annotation in the [OCI Image Index](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys)
- the timestamp of the files exported with the local exporter
- the timestamp of the files exported with the tar exporter

And also:
- The result of every nested build-tool that uses the SOURCE_DATE_EPOCH variable

Ending up in possibly deterministic, binary reproducible images.

As this is mostly compiler driven developed without deep understanding of daggers internals (just best guesses), It would be helpful to get some feedback on this.

closes: #7721

### DIFF
--- a/core/container.go
+++ b/core/container.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/dagger/dagger/engine"
 	"io/fs"
 	"os"
 	"path"
@@ -1214,6 +1215,7 @@ func (container *Container) Publish(
 	platformVariants []*Container,
 	forcedCompression ImageLayerCompression,
 	mediaTypes ImageMediaTypes,
+	rewriteTimestamp dagql.Boolean,
 ) (string, error) {
 	if mediaTypes == "" {
 		// Modern registry implementations support oci types and docker daemons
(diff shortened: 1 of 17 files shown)


core/container.go
```diff
@@ -1214,6 +1215,7 @@ func (container *Container) Publish(
 	platformVariants []*Container,
 	forcedCompression ImageLayerCompression,
 	mediaTypes ImageMediaTypes,
+	rewriteTimestamp dagql.Boolean,
```
jedevc, line 1218: I think the arg here should *ideally* be a timestamp? Then, if left empty, we would default to `clientMetadata.SourceDateEpoch`.

Just because currently, to use this feature, you have to update your module code *and* have the client pass `SOURCE_DATE_EPOCH` which feels tricky - is there ever a reason why we wouldn't want to respect `SOURCE_DATE_EPOCH` if it was set by the client?

We're not tied strictly to the buildkit semantics here, so we're free to build the API we want :tada:

cc @sipsma in case you have opinions about this

core/container.go
```diff
@@ -1233,6 +1235,19 @@ func (container *Container) Publish(
 		opts[string(exptypes.OptKeyForceCompression)] = strconv.FormatBool(true)
 	}
 
+	clientMetadata, err := engine.ClientMetadataFromContext(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	if clientMetadata.SourceDateEpoch != nil {
+		opts[string(exptypes.OptKeySourceDateEpoch)] = strconv.FormatInt(clientMetadata.SourceDateEpoch.Unix(), 10)
```
jedevc, line 1244: We definitely need tests for this - they'd belong in `core/integration/container_test.go`/etc (but yes, we should make sure we have a solid idea of the API first!)

The code around anything interacting with buildkit is going through a *lot* of churn right now, so we need to make sure we're not accidentally bricking anything in the future :yellow_heart:

PR #10334, opened 2025-05-05
fix `err` overwrite causing data race detector failures in integ tests

sipsma:
This was causing flakiness in tests when the detector went off (e.g. https://v3.dagger.cloud/dagger/traces/1df689be91b6f1d837af53ff3cd62b41?span=4acfd90e58c43f70&logs)


sipsma:
Pushed an additional commit w/ a separate fix for a consistent failure we're hitting in TestTelemetry caused by `redis` image using latest and thus changing output (just made it `:7.4.3` now)

vito (approved):
:pray:

PR #10327, opened 2025-05-05
ci: Move a bunch more jobs to alt 2 ci runners

gerhard:
These have proven themselves up to 30-40% faster. A few of these jobs can benefit from caching, so let's see how well they work out in practice.

---

| Job Name | Before (P90 2w) | After (single run) |
|:--------|-------------------|-------------------|
| X | 1016s (16m 56s) | 818s (13m 38s) |
| Y | 940s (15m 40s) | 588s (9m 48s) |
| Z | 768s (12m 48s) | 459s (7m 39s) |

PR #10340, opened 2025-05-06
make `TestCrossSessionSecretURICaching` resilient to cache prunes

sipsma:
This test started flaking in [this PR](https://github.com/dagger/dagger/pull/10327). Didn't repro but looking at it seems like it could be explained by the cache getting pruned while the test is running (which would cause the assertions of cache re-use to fail).

The update uses nested execs to run commands, which ensures that the session stays open for the duration of the test case assertions and thus should prevent the cache from getting pruned until the test is done.


jedevc (approved):

gerhard (approved):
Nice one!

Merging now so that I can rebase the other PR on top of this one, and make sure that everything still works as expected before merging that PR.

PR #10341, opened 2025-05-06
mcp: log std{in,out,err} to files specified by DAGGER_LOG_STD{IN,OUT,ERR} envvars

tiborvass:
This makes debugging MCP in certain environments easier by providing a way to log the dagger cli's stdin/stdout/stderr.

### DIFF
--- a/cmd/dagger/engine.go
+++ b/cmd/dagger/engine.go
@@ -97,9 +97,6 @@ func withEngine(
 
 		params.WithTerminal = withTerminal
 
-		params.Stdin = os.Stdin
-		params.Stdout = os.Stdout
-
 		params.Interactive = interactive
 		params.InteractiveCommand = interactiveCommandParsed
 
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -78,11 +78,47 @@ var (
 	Frontend idtui.Frontend
 )
 
+var (
+	stdin          io.Reader
+	stdout, stderr io.Writer
+)
+
+func logFileFromEnv(envName string) *os.File {
+	if logFileName := os.Getenv(envName); logFileName != "" {
+		f, err := os.Create(logFileName)
+		if err != nil {
+			panic(err)
+		}
+		return f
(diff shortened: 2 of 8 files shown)


cmd/dagger/main.go
```diff
@@ -78,12 +78,48 @@ var (
...
+
+func wrapReader(r io.Reader, envName string) io.Reader {
+	f := logFileFromEnv(envName)
+	if f == nil {
+		return r
+	}
+	return io.TeeReader(r, f)
+}
+
+func wrapWriter(w io.Writer, envName string) io.Writer {
```
cwlbraa, line 105: kinda nit, but iiuc these pass straight through if vars are unset, right?

```suggestion
func maybeWrapReader(r io.Reader, envName string) io.Reader {
	f := logFileFromEnv(envName)
	if f == nil {
		return r
	}
	return io.TeeReader(r, f)
}

func wrapWriter(w io.Writer, envName string) io.Writer {
```

cmd/dagger/engine.go
```diff
@@ -97,9 +97,6 @@ func withEngine(
 
 		params.WithTerminal = withTerminal
 
-		params.Stdin = os.Stdin
-		params.Stdout = os.Stdout
```
cwlbraa, line 101: were these 2 lines added when MCP was added? 

i'm not sure the downstream effect of this deletion given that other changes are tightly scoped to `dagger mcp` and the set env vars

cwlbraa (approved):
lgtm pending an explanation for the withEngine change

cmd/dagger/engine.go
tiborvass: yes

tiborvass: I could indeed put the withEngine change separately, but this was part of the refactor as only MCP uses it for now and it's a way to make sure only MCP has the session attachable.

tiborvass:
rebased

vito (approved):

Issue #8810, opened 2024-10-29
Allow passing "empty" values from Go SDK


jedevc:
Suppose the following scenario:

```go
// module foo
func (f *Foo) DoFoo(
	// +optional
	// +default=true
	arg bool,
	// +optional
	arg2 string,
} error {
	// ...
}
```

```go
// module bar
func (f *Bar) DoBar() error {
	return dag.Foo().DoFoo(dagger.FooDoFooOpts{Arg: false})
}
```

As described in https://github.com/dagger/dagger/issues/7777#issuecomment-2203387161, this `false` won't apply, and we'll get the default `true` value set for `arg`.

That's because of the following logic in the generated go code: https://github.com/dagger/dagger/blob/6873b3a3960f7927c6b0971e97141ad3bc9a3e55/sdk/go/dagger.gen.go#L1092-L1095

If a value is a zero-value (e.g. `false`, `0`, `""`), then we won't include in the values sent to the server. This is done because in go **there's no way to distinguish between a value that's not provided and a value that's set to the zero value**. This issue mostly appears with default booleans of `true`, but could also appear with empty strings.

This is a pretty big issue for the Go SDK - it means that users are incentivized to just avoid options that default to `true` if they want to be called from the Go SDK (other languages don't have this issue).

We should work out what to do, since this is a pretty big stumbling block. Some options off the top of my head:

- Document this limitation somewhere - all SDKs have their quirks, so we could have a docs page where we write them all down, and we just accept that this is one of Go's.

- Explicitly handle the boolean default `true` case - in this case, invert the option for go. So in the above example, the Go SDK gets an option `NoArg`:

  ```go
  // module bar
  func (f *Bar) DoBar() error {
    	return dag.Foo().DoFoo(dagger.FooDoFooOpts{NoArg: true})
  }
  ```
  
  However, this only handles for the boolean case, we still haven't handled the similar (and more complex) `""`/`0` cases.

- Give up entirely on using structs for opts - instead allow using a `With` pattern - this is a pretty standard pattern in the go ecosystem of doing options (which we use ourselves in the engine/cli).

  ```go
  // module bar
  func (f *Bar) DoBar() error {
    	return dag.Foo().DoFoo(dagger.FooDoFooWithArg(false))
  }
  ```

  - This is very verbose. Needing to explicitly prefix `FooDoFoo` for each distinct option is quite miserable.
[...]

PR #10342, opened 2025-05-06
"reload" host dirs and files

### LINKED ISSUE #9934: Reload host directories
## Problem

When calling `Host.directory()` or `Host.file()`, I get a one-time snapshot of the file or directory, *and it will never change for the whole session*.

This hasn't been too much of a problem in one-off function calls with `dagger call`. But it becomes a major obstacle to using the Dagger Shell interactively for local dev.

## Solution

Enable reloading of host files and directories in the same session.

## Design

API: TBD
Shell UX: TBD

cwlbraa:
fixes #9934

this PR seeks to provide users the means to auto-reload host directories.

this will be the new default behavior for +defaultPath. prior to this PR, it's very surprising in shell and prompt modes that input argument +defaultPaths are not reloaded - you can open a shell, be editing application code on your inner loop, but your dagger functions will never reload your source code unless you restart the shell. after this PR, it does what you'd expect, and does so fairly fast thanks to perf optimizations in localsource.

I feel like there are places I'm missing where noCache should be set to true. I covered the +defaultDir case, but maybe this should be the default for loading dirs from user-provided paths, too? 

maybe even for module loading in shell? even with this change, edits to module code aren't reflected in a persistent shell, prompt, or mcp session. 

this PR also deletes our copy of rand.Text now that we're on go 24.


core/integration/host_test.go
```diff
@@ -320,4 +321,47 @@ func (HostSuite) TestFile(ctx context.Context, t *testctx.T) {
...
+
+				file := c.Host().File(bPath, test.opts...)
+				content, err := file.Contents(ctx)
+				require.NoError(t, err)
+				require.Equal(t, "1", content)
+
+				require.NoError(t, os.WriteFile(bPath, []byte("12"), 0o600))
+
+				// file = c.Host().File(bPath, test.opts...) // this doesn't change behavior, but maybe worth testing?
+				content, err = file.Contents(ctx)
```
vito, line 361: Yeah that's an interesting property. I guess when you disable caching for these, it means any unlazying call that has it as an input will do a filesync?

Almost feels like something I'd want API control over - like to re-sync the directory, but have the returned object be stable from that point forward.

Or does that work like this?

```go
syncedFile, err := c.Host().File(bPath, dagger.HostFileOpts{NoCache: true}).Sync(ctx)


syncedFile.Contents(ctx) // "1"

// write "12" to bPath

syncedFile.Contents(ctx) // still "1"

reSyncedFile, err := c.Host().File(bPath, dagger.HostFileOpts{NoCache: true}).Sync(ctx)

reSyncedFile.Contents(ctx) // "2"
```

dagql/cachekey.go
```diff
@@ -76,6 +76,30 @@ func CachePerSessionObject[A any](
...
+	CacheType() CacheControlType
+}
+
+type CacheControlType int
+
+const (
+	CacheTypeUnset     CacheControlType = iota
+	CacheTypePerClient CacheControlType = 1
+	CacheTypePerCall   CacheControlType = 2
+)
```
jedevc, line 90: :eyes: I think we can drop `= 1` / `= 2` right? `iota` should auto-increment this.

core/integration/host_test.go
jedevc: Will let @sipsma confirm, but I think that once the local directory is loaded - it gets "frozen" at that point, and can't be modified.

Agreed that that example *should* be how it works, would be good to test it.

core/schema/host.go
```diff
@@ -259,6 +263,12 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.Instance[*core.Ho
 		buildkit.WithTracePropagation(ctx),
 	}
 
+	// this is slightl hacky - to bust the cache with noCache:true, we exclude a random text path.
+	// no real chance of excluding something real and this changes both the name and the cache key without changes to localsource.
+	if args.NoCache {
+		args.Exclude = append(args.Exclude, rand.Text())
```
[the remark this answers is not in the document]
jedevc: :cry: I don't love this, but yup, something like this is gonna be the way to hack it for now I think.

If you leave it in as a `HACK`, we can fix this up when we dagop-ify it, since then we can actually control the buildkit cache keys :smile:

core/integration/host_test.go
cwlbraa: will cover, i also don't know what the behavior is here XD

dagql/cachekey.go
cwlbraa: i think so? i don't recall if this was LLM or linter generated, i definitely wrote pseudocode and then asked claude to make it into a idiomatic go enum XD

core/schema/host.go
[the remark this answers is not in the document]
sipsma: There is a route to avoid this by customizing the LLB, which we've done in the past for other sources pre-dagop-ification.

e.g. `gitdns` used this (old commit, not present in current codebase): https://github.com/dagger/dagger/blob/83b7c60e49e860b1b9d5d7296461fc21edd33f9c/engine/sources/gitdns/state.go#L15

We'd be able to include an extra attr in the LLB, which would could then see in the local source cache key and automatically include a rand string in the cache key ([here](https://github.com/sipsma/dagger/blob/8ff111162057de9d4e699da3f30a0b806020d83e/engine/sources/local/source.go#L127-L127))

I would _slightly_ prefer that since it is simpler than always setting an extra exclude in these situations, but not a blocker for me since I can't *think* of why the random exclude hack would break or meaningfully slow anything down.

core/integration/host_test.go
sipsma: Yeah when `NoCache: true` is set then the local dir will still be "frozen" once it's been turned into an id or sync is called. At least, that's what I'd predict the behavior is based on the code (and what I'd want it to be), but indeed worth testing.

dagql/cachekey.go
cwlbraa: omg you know what actually happened? my lsp hints show the `= 1` as overlay text so i didn't think it was getting committed XD

core/schema/host.go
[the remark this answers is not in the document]
cwlbraa: > If you leave it in as a HACK, we can fix this up when we dagop-ify it, since then we can actually control the buildkit cache keys 😄

and i'm down to take a shot at this, I've been curious about dagopifying anyways. i just don't want to bring it in scope for this particular change.

core/integration/host_test.go
cwlbraa: > Yeah when NoCache: true is set then the local dir will still be "frozen" once it's been turned into an id or sync is called. At least, that's what I'd predict the behavior is based on the code (and what I'd want it to be), but indeed worth testing.

unfortunately that's not the current behavior. having revisited this to make the commented-out test case explicit, the behavior still matches the comment i've left in, which is to say that @vito's example above does not work, it returns 12 for the 2nd Contents call on the already-sync'd file w/NoCache:true - the exact same behavior as though you'd recreated the file object from scratch.

sipsma: How did you test that? I just pulled this PR down and modified the test case like this:
```diff
diff --git a/core/integration/host_test.go b/core/integration/host_test.go
index dbba43e2b..1b532d575 100644
--- a/core/integration/host_test.go
+++ b/core/integration/host_test.go
@@ -363,13 +363,18 @@ func (HostSuite) TestFile(ctx context.Context, t *testctx.T) {
                                bPath := filepath.Join(dir, rand.Text())
                                require.NoError(t, os.WriteFile(bPath, []byte("1"), 0o600))
 
-                               file := c.Host().File(bPath, test.opts...)
+                               file, err := c.Host().File(bPath, test.opts...).Sync(ctx)
+                               require.NoError(t, err)
                                content, err := file.Contents(ctx)
                                require.NoError(t, err)
                                require.Equal(t, "1", content)
 
                                require.NoError(t, os.WriteFile(bPath, []byte("12"), 0o600))
 
+                               content, err = file.Contents(ctx)
+                               require.NoError(t, err)
+                               require.Equal(t, "1", content)
+
                                file = c.Host().File(bPath, test.opts...)
                                content, err = file.Contents(ctx)
                                require.NoError(t, err)
```

And it passes, which would indicate that `.Sync` does indeed result in the load being frozen.

cwlbraa: ah, i assumed any unlazy call implied the underlying object was sync'd, but the assignment afterwards is what matters

core/integration/host_test.go
```diff
@@ -320,4 +320,61 @@ func (HostSuite) TestFile(ctx context.Context, t *testctx.T) {
 		require.NoError(t, err)
 		require.Equal(t, "hello world", content)
 	})
+	t.Run("cache behavior", func(ctx context.Context, t *testctx.T) {
```
sipsma, line 323: worth adding variations for `HostDirectoryOpts` too

core/modulesource.go
```diff
@@ -350,6 +350,7 @@ func (src *ModuleSource) LoadContext(
 				Args: []dagql.NamedInput{
 					{Name: "path", Value: dagql.String(path)},
 					{Name: "exclude", Value: dagql.ArrayInput[dagql.String](dagql.NewStringArray(ignore...))},
+					{Name: "noCache", Value: dagql.NewBoolean(true)},
```
sipsma, line 353: worth adding a test for this behavior existing now

cwlbraa: oh yeah, agreed, for some reason i thought i already had but it's not in the diff

core/integration/host_test.go
cwlbraa: yep agreed

core/integration/host_test.go
cwlbraa: tests improved to reflect this and nail it down fully

core/modulesource.go
cwlbraa: i did already have it, but committed it as part of a refactor that i abandoned and squashed out. recovered. it passes.

core/integration/host_test.go
cwlbraa: done

cwlbraa:
test coverage dramatically improved, and they all confirm the desired behavior (aka i didn't need production code changes to make them pass)

core/integration/module_test.go
```diff
@@ -5873,146 +6076,3 @@ func logGen(ctx context.Context, t *testctx.T, modSrc *dagger.Directory) {
 		}
 	})
 }
-
-func (ModuleSuite) TestSSHAgentConnection(ctx context.Context, t *testctx.T) {
```
cwlbraa, line 5877: it's been driving me nuts that these are below the test helpers so i moved em up

sipsma (approved):

cwlbraa:
@jedevc added changelog.


> Is the idea to also extend it to args passed through the shell like `host directory --path=...` as a follow-up?

there are multiple places to take this and shipping this small cut of functionality leaves a lot of doors open. When you say "args passed through the shell" if you mean module constructor arguments, then solidly yes, those should definitely have the same behavior as +defaultPath.

if we're talking literally in the shell `host | directory --path=...`, this functionality is already available as `host | directory --path=... --noCache`.

the other thing we should consider as a group is whether or when to make this the default behavior for all `host | directory`s. for now, I don't want to make such a subtle and sweeping breaking change. we can't actually change the default without either making a breaking change or finding a way to fix the [bool default issue](https://github.com/dagger/dagger/issues/8810).

PR #10343, opened 2025-05-06
add docs for secret caching behavior

sipsma:
Follow-up for https://github.com/dagger/dagger/pull/10311 to explain how secrets impact cache (now, as of v0.18.6) and how the cache behavior can be customized.

Let me know if it's comprehensible as is or if it would benefit from more examples/explanation.

### DIFF
--- a/docs/current_docs/api/arguments.mdx
+++ b/docs/current_docs/api/arguments.mdx
@@ -750,8 +750,6 @@ The result will be a JSON-formatted list of issues from Dagger's repository.
 
 When invoking the Dagger Function using the Dagger CLI, secrets can be sourced from multiple providers. Dagger can read secrets from the host environment, the host filesystem, and the result of host command execution, as well as from external secret managers [1Password](https://1password.com/) and [Vault](https://www.hashicorp.com/products/vault).
 
-When a `Secret` is included in other operations, the cache for those operations will be based on the plaintext value of the secret. If the same operation is run with a secret with the same plaintext value, that operation may be cached rather than re-executed. In the above example, the cache for the `withExec` will be based on the plaintext value of the secret. If two clients execute the container with a secret whose plaintext values are the same, the container execution may be cached. Otherwise, if the plaintext values of the secret are different, the container execution will not be cached.
-
 ### Host secret providers
 
 Here is an example call for this Dagger Function, with the secret sourced from a host environment variable named `GITHUB_API_TOKEN` via the `env` provider:
@@ -870,6 +868,10 @@ dagger call github-api --token=vault://infra/github.credential
 ```
 </TabItem>
 </Tabs>
+
+### Caching
+
+When a `Secret` is included in other operations, the layer cache entries for those operations will be based on the plaintext value of the secret. If the same operation is run with a secret with the same plaintext value, that operation may be cached rather than re-executed. In the above example, the cache for the `withExec` will be based on the plaintext value of the secret. If two clients execute the container with a secret whose plaintext values are the same, the container execution may be cached. Otherwise, if the plaintext values of the secret are different, the container execution will not be cached.
 
 ## Service arguments
 
--- a/docs/current_docs/api/arguments.mdx
+++ b/docs/current_docs/api/arguments.mdx
@@ -750,6 +750,8 @@ The result will be a JSON-formatted list of issues from Dagger's repository.
 
 When invoking the Dagger Function using the Dagger CLI, secrets can be sourced from multiple providers. Dagger can read secrets from the host environment, the host filesystem, and the result of host command execution, as well as from external secret managers [1Password](https://1password.com/) and [Vault](https://www.hashicorp.com/products/vault).
 
+When a `Secret` is included in other operations, the cache for those operations will be based on the plaintext value of the secret. If the same operation is run with a secret with the same plaintext value, that operation may be cached rather than re-executed. In the above example, the cache for the `withExec` will be based on the plaintext value of the secret. If two clients execute the container with a secret whose plaintext values are the same, the container execution may be cached. Otherwise, if the plaintext values of the secret are different, the container execution will not be cached.
+
 ### Host secret providers
 
 Here is an example call for this Dagger Function, with the secret sourced from a host environment variable named `GITHUB_API_TOKEN` via the `env` provider:
--- a/docs/current_docs/api/secrets.mdx
+++ b/docs/current_docs/api/secrets.mdx
@@ -62,7 +62,11 @@ dagger core container \
 
 ## Caching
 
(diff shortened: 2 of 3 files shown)


docs/current_docs/api/secrets.mdx
```diff
@@ -60,6 +60,8 @@ dagger core container \
 
 [Secret arguments](./arguments.mdx#secret-arguments) can be sourced from multiple providers: the host environment, the host filesystem, the result of host command execution, and external secret managers [1Password](https://1password.com/) and [Vault](https://www.hashicorp.com/products/vault).
 
+When a `Secret` is included in other operations, the cache for those operations will be based on the plaintext value of the secret. If the same operation is run with a secret with the same plaintext value, that operation may be cached rather than re-executed. In the above example, the secret is specified as `env://API_TOKEN`, so the cache for the `with-exec` will be based on the plaintext value of the secret. If two clients execute the container and have the same `API_TOKEN` environment variable value, they may share cache for the container. However, if two clients have different values for the `API_TOKEN` environment variable, the `with-exec` will not share cache, even though both clients specify the secret as `env://API_TOKEN`.
```
marcosnils, line 63: If this section of the docs is all about secrets, shouldn't we also mention the cacheKey implicancies  here as well  on top of the cookbook? Seems something I'd expect to find here

sipsma: Yeah I wasn't sure what level of details we generally include in these sections. I can link to the cookbook from here if that's alright?

marcosnils: sure, that also LGTM

docs/current_docs/cookbook/cookbook.mdx
```diff
@@ -1821,6 +1821,33 @@ Build from a Dockerfile with a mounted secret from the host environment:
...
 
+### Customize secret caching behavior
+
+By default, the cache for operations that include a secret will be based on the plaintext value of the secret. Operations that include secrets with the same plaintext value may share cache, but if the plaintext differs then the operation will not share cache.
+
+In some cases, users may desire that operations share cache even if the secret plaintext value is different. For example, a secret may often rotate in plaintext value but not be meaningfully different such that cache for operations including it should be busted.
+
+For these use cases, the optional `cacheKey` argument to `Secret` construction can be used to specify the "cache key" of the secret. Secrets that share the same `cacheKey` will be considered equivalent when checking cache of operations that include them, even if their plaintext value differs.
+
+Example usage:
```
vikram-dagger, line 1832:
```suggestion
#### Example

Use a secret with a specified cache key:

```

Minor suggestion for consistency with other examples

docs/current_docs/cookbook/cookbook.mdx
```diff
@@ -1821,6 +1821,33 @@ Build from a Dockerfile with a mounted secret from the host environment:
     </TabItem>
     </Tabs>
 
+### Customize secret caching behavior
+
+By default, the cache for operations that include a secret will be based on the plaintext value of the secret. Operations that include secrets with the same plaintext value may share cache, but if the plaintext differs then the operation will not share cache.
```
vikram-dagger, line 1826:
```suggestion
By default, the cache for operations that include a secret will be based on the plaintext value of the secret. Operations that include secrets with the same plaintext value may share the cache, but if the plaintext differs then the operation will not share the cache.
```

Question: this refers to "cache for operations" In our docs, we use the terms "layer cache" and "cache volumes": https://docs.dagger.io/features/caching. Which of these is the "cache for operations"? Can we replace "cache for operations" with one of the other two terms, so readers understand can grok it better?

docs/current_docs/cookbook/cookbook.mdx
```diff
@@ -1821,6 +1821,33 @@ Build from a Dockerfile with a mounted secret from the host environment:
...
+
+In some cases, users may desire that operations share cache even if the secret plaintext value is different. For example, a secret may often rotate in plaintext value but not be meaningfully different such that cache for operations including it should be busted.
```
vikram-dagger, line 1828:
```suggestion
In some cases, users may desire that operations share the cache even if the secret plaintext value is different. For example, a secret may often rotate in plaintext value but not be meaningfully different; in these cases, it should still be possible to reuse the cache for operations that include that secret.
```

docs/current_docs/api/secrets.mdx
vikram-dagger: I think it's useful to include the two paragraphs that you have in the cookbook about the cache key in these pages as well for information. And with that, include a link to the cookbook for the usage example.

vikram-dagger, line 63:
```suggestion
## Caching

When a `Secret` is included in other operations, the cache for those operations will be based on the plaintext value of the secret. If the same operation is run with a secret with the same plaintext value, that operation may be cached rather than re-executed. In the above example, the secret is specified as `env://API_TOKEN`, so the cache for the `with-exec` will be based on the plaintext value of the secret. If two clients execute the container and have the same `API_TOKEN` environment variable value, they may share cache for the container. However, if two clients have different values for the `API_TOKEN` environment variable, the `with-exec` will not share cache, even though both clients specify the secret as `env://API_TOKEN`.
```

docs/current_docs/api/secrets.mdx
```diff
@@ -60,6 +60,8 @@ dagger core container \
...
+
 ## Security considerations
```
vikram-dagger, line 65: Are there any security considerations to be aware of around secret caching? If yes, they could be added here.

docs/current_docs/api/arguments.mdx
```diff
@@ -750,6 +750,8 @@ The result will be a JSON-formatted list of issues from Dagger's repository.
 
 When invoking the Dagger Function using the Dagger CLI, secrets can be sourced from multiple providers. Dagger can read secrets from the host environment, the host filesystem, and the result of host command execution, as well as from external secret managers [1Password](https://1password.com/) and [Vault](https://www.hashicorp.com/products/vault).
 
+When a `Secret` is included in other operations, the cache for those operations will be based on the plaintext value of the secret. If the same operation is run with a secret with the same plaintext value, that operation may be cached rather than re-executed. In the above example, the cache for the `withExec` will be based on the plaintext value of the secret. If two clients execute the container with a secret whose plaintext values are the same, the container execution may be cached. Otherwise, if the plaintext values of the secret are different, the container execution will not be cached.
```
vikram-dagger, line 753:
```suggestion
### Caching

When a `Secret` is included in other operations, the cache for those operations will be based on the plaintext value of the secret. If the same operation is run with a secret with the same plaintext value, that operation may be cached rather than re-executed. In the above example, the cache for the `withExec` will be based on the plaintext value of the secret. If two clients execute the container with a secret whose plaintext values are the same, the container execution may be cached. Otherwise, if the plaintext values of the secret are different, the container execution will not be cached.
```

vikram-dagger: Would you consider moving this after the section "External secret providers"?

docs/current_docs/cookbook/cookbook.mdx
sipsma: I updated "share cache" to be "share layer cache entries" ("share the layer cache" didn't sound right to me)

docs/current_docs/api/secrets.mdx
sipsma: Not really anything the user needs to be aware of or take action from. There's some implementation details [mentioned under "Security" in this PR description](https://github.com/dagger/dagger/pull/10311) but based on what we have in the docs today I don't think it's anything we need to mention here.

vikram-dagger (approved):

PR #10344, opened 2025-05-07
fix(sdk/elixir): fix publish failure on 0.18.6

wingyplus:
The ElixirSdkDev set the working directory `/sdk` while the `t.Dagger.Source` is `/sdk/elixir` cause a publish function write to incorrect place.

Tested by add a terminal debugger before calling `hex.publish and call `dagger call sdk elixir publish --tag=0.18.6 --dry-run`, then cat the `mix.exs` file, the `@version` must be updated to 0.18.6.


wingyplus:
This bug affects only user who uses the SDK on the hex.pm because the SDK didn't get publish but it also fine when using with module because it fetch from git directly.

jedevc (approved):
Thanks :tada:

PR #10351, opened 2025-05-07
chore: run go modernize

jedevc:
The result of running [modernize](https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize):

```
$ go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -test -fix ./...
```

With a few manual tidy-ups :tada:

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"slices"
 	"sync"
 	"time"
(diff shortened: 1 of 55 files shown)


gerhard (approved):
The changes look good to me, this is a nice clean-up.

While `testdev-call-and-shell` is a known flake, I would double-check that the `elixir-dev` failure is also a transient issue before merging this.

Issue #10356, opened 2025-05-07
Dagger shell calls 1password on start


kpenfound:
Not sure how to phrase the title 😂 

Starting from 0.18.6, if you have LLM credentials in your environment pointing to 1password refs, like:

```
GEMINI_API_KEY=op://Employee/GEMINI_API_KEY/credential
```

Dagger will authenticate to pull those credentials as soon as shell is started. This is disruptive because I don't always need to authenticate the LLM client in every shell session.

PR #10357, opened 2025-05-07
lazy LLM api key loading/model selection

cwlbraa:
fixes https://github.com/dagger/dagger/issues/10356

~this also uncomments the evals so we can run `dagger call evals` now that the appropriate APIs are released. note these don't build a dagger-in-dagger dev engine so you usually wanna use em with `./hack/dev`.~ commented again because we keep changing the api :)

TODO:
- [x] run evals
- [x] manually test the secret prompting thing at the core of 10356
   - [x] fix: still prompting- client needs some changes too 
- [x] ~refactor llm.endpoint init to use sync.Once~ bad idea, not doing this, err handling gets weird.
- [ ] ??? very open to suggestions as to what else to test here, there shouldn't be behavior changes outside fixing the issue we're trying to fix, but i also don't entire trust our test coverage here (hence running evals)

### DIFF
--- a/.dagger/main.go
+++ b/.dagger/main.go
@@ -190,32 +190,32 @@ func (dev *DaggerDev) Test() *Test {
 	return &Test{Dagger: dev}
 }
 
-// TODO: these depend on unreleased APIs
-func (dev *DaggerDev) Evals(
-	ctx context.Context,
-	// +defaultPath=./core/llm_docs.md
-	docs *dagger.File,
-	// +defaultPath=./core/llm_dagger_prompt.md
-	systemPrompt *dagger.File,
-) error {
-	return dag.Evaluator(dagger.EvaluatorOpts{
-		Docs:          docs,
-		InitialPrompt: systemPrompt,
-	}).EvalsAcrossModels().Check(ctx)
-}
+// // TODO: these depend on unreleased APIs
+// func (dev *DaggerDev) Evals(
+// 	ctx context.Context,
+// 	// +defaultPath=./core/llm_docs.md
+// 	docs *dagger.File,
+// 	// +defaultPath=./core/llm_dagger_prompt.md
+// 	systemPrompt *dagger.File,
+// ) error {
+// 	return dag.Evaluator(dagger.EvaluatorOpts{
+// 		Docs:          docs,
+// 		InitialPrompt: systemPrompt,
+// 	}).EvalsAcrossModels().Check(ctx)
+// }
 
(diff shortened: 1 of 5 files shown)


core/llm.go
```diff
@@ -480,6 +467,26 @@ func (llm *LLM) Clone() *LLM {
 	return &cp
 }
 
+func (llm *LLM) Endpoint(ctx context.Context) (*LLMEndpoint, error) {
+	if llm.endpoint != nil {
```
sipsma, line 471: I saw what you mentioned about `sync.Once` in the description, but you can use [OnceValues](https://pkg.go.dev/sync#OnceValues) to help w/ errors. Or even just a mutex would be fine here

sipsma (commented):
Not familiar w/ the eval stuff, but otherwise LGTM

core/llm.go
cwlbraa: lmao i read through [the proposal to add that](https://github.com/golang/go/issues/56102) and several others and came to the conclusion that none of them ever shipped... clearly i misunderstood the final outcome XD 

will fix, oncevalues is exactly what i needed.

.dagger/main.go
```diff
@@ -156,31 +156,31 @@ func (dev *DaggerDev) Test() *Test {
...
-// 	docs *dagger.File,
-// 	// +defaultPath=./core/llm_dagger_prompt.md
-// 	systemPrompt *dagger.File,
-// ) (string, error) {
-// 	return dag.Evaluator(dagger.EvaluatorOpts{
-// 		Docs:          docs,
-// 		InitialPrompt: systemPrompt,
-// 	}).GenerateSystemPrompt(ctx)
-// }
+func (dev *DaggerDev) Evals(
```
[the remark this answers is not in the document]
cwlbraa: sgtm, will remove before merging

dagger.json
```diff
@@ -37,6 +37,10 @@
       "name": "elixir-sdk-dev",
       "source": "sdk/elixir/dev"
     },
+    {
+      "name": "evaluator",
+      "source": "modules/evaluator"
+    },
```
vito, line 43: woops gotta remove this part too

```suggestion
```

vito (approved):
(approving to unblock, can clean up that dependency later)

sipsma (approved):

dagger.json
cwlbraa: good catch, removing

Issue #10320, opened 2025-05-05
Errors are being cached


marcosnils:
# Problem

Given the following example: 

```go
package main

import (
	"context"
	"time"
)

type CacheErrorsTest struct{}

func (m *CacheErrorsTest) Test(ctx context.Context) {
	c := dag.Container().From("alpine").
		WithMountedCache("/lala", dag.CacheVolume("cache")).
		WithEnvVariable("BUST", time.Now().String()).
		WithExec([]string{"rm", "-rf", "/lala/foo.txt"})

	_, err := c.WithExec([]string{"ls", "/lala/foo.txt"}).Sync(ctx)

	if err == nil {
		panic("should error")
	}

	c.WithExec([]string{"touch", "/lala/foo.txt"}).Sync(ctx)

	_, err = c.
		//WithEnvVariable("BUST", time.Now().String()).  uncomment to make the test pass
		WithExec([]string{"ls", "/lala/foo.txt"}).Sync(ctx)

	if err != nil {
		panic("should not error")
	}

}
```
You'll notice that without a cache bust, the second `ls` exec will fail given that it will return the cached error response from the first one. 

Seem to me that this comes from the `dagQL` level somehow given that the trace (https://v3.dagger.cloud/marcos-test/traces/000a7698719eb365026ed03cc27ed7e1?span=06966b2aa4707511) doesn't show the second operation as being `cached`  from buildkit's side.

# Solution

The test should pass since we shouldn't be caching error responses.

sipsma:
I believe this is actually a buildkit behavior, not dagql: https://github.com/dagger/buildkit/blob/c2b4992517e8670bc446435227e3209a336abc1d/solver/jobs.go#L1098-L1098
* It basically means that while an LLB op has been solved in a session, the result of that op (whether error or not) is cached in memory (on top of the actual on-disk caching)

@marcosnils can you check if this repro's on an older version like v0.17.x? Those lines in buildkit have been in place for basically ever, so it's possible we just never noticed it. But want to check we didn't change something that previously had resulted in us avoiding it somehow.

marcosnils:
> [@marcosnils](https://github.com/marcosnils) can you check if this repro's on an older version like v0.17.x? T

 yep. Still happens in v0.17.x. According to Erik seems like this behavior has been like this forever.

PR #10377, opened 2025-05-12
tests: avoid flake from error caching

jedevc:
As we discovered recently, errors can cache between two sessions when they're active at the same time. This is a pain, and should probably be considered a bug - see #10320.

However, these tests shouldn't be failing - we've had this behavior *forever*, it's not a new failure. It's also not what we're testing, the flakes just add to our CI noise - we're aware it's an issue, and we should fix it (and add a test specifically for it).


gerhard (approved):
👏 🚀

PR #10379, opened 2025-05-12
llm: durable to Bedrock API quirks

vito:
Ran into these trying Dagger with behind-the-firewall OpenAI provider. Seems like it generates janky function calls for some reason. Adding these checks fixed everything. :shrug: 

This is two commits but I suspect they both fix one issue, I just happened to hit one error before the other. It's probably generating a fully empty function call struct.

### DIFF
--- a/core/llm_openai.go
+++ b/core/llm_openai.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"dagger.io/dagger/telemetry"
+	"github.com/dagger/dagger/engine/slog"
 	"github.com/openai/openai-go"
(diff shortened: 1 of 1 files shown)


kpenfound:
What OpenAI frontend is this with @vito ?

jedevc (approved):

vito:
@kpenfound I'm not sure where the quirks come from in the stack, but this is with:

* `OPENAI_API_URL=` pointing to https://github.com/wealthsimple/llm-gateway
* `OPENAI_API_MODEL=openai/bedrock-3.7-sonnet-latest`

PR #10366, opened 2025-05-09
llm: static tool scheme

vito:
## Switch _mostly_ from dynamic tools to static

Instead of listing available objects + methods in tool descriptions, there are new tools:

* `list_objects` - list all known objects and their descriptions
* `list_methods(type)` - list known methods, optionally for a specific type
  * shows method name (`http`, `Container.withExec`), required args, and return type
  * does NOT show description - required args is enough

Instead of representing APIs as tools that dynamically become available, there are tools for calling methods:

* `call_method` - call a single method against a `self`
* `chain_methods` - call a chain of methods starting from a `self`
  * I experimented with this separately in #10229 but this PR's implementation seems to work much more consistently (maybe "method chaining" is a strong metaphor)

So what does "mostly" mean?:

* The `save` tool still has a dynamic description. I experimented with making this static (by adding a `list_outputs` and having `save` save one arbitrary name+value output at a time) but it seemed to really degrade model behavior - it's tough to beat a single function with a schema for the required outputs.
* The `user_provided_values` tool still works the same as before (a description derived from the inputs).
* I think these both are fine for all intents and purposes. We don't need MCP clients to save values, and inputs are static anyway.

## Remove the `think` tool

It's nifty, but not fully proven - evals pass just fine without it. Some clients (like Zed) already have their own `thinking` tool, some models have an explicit "thinking" mode, so let's wait until we're sure we need it and add it more thoughtfully (opt-in?).

# Evals Report ([analysis](https://v3.dagger.cloud/dagger/traces/30e3f9b1cfb79eb1aa7625926b2e4260))

| Model | Eval | Success Rate | Input / Output Tokens | Traces |
|-------|------|-------------|---------------------|-------|
[...]

PR #10385, opened 2025-05-12
llm(google): use new (non-deprecated) genai client

vito:
Noticed this whole package is deprecated. Looks like this new package defines a few things that were missing, e.g. `FinishReason(10)` will say `MALFORMED_FUNCTION_CALL` or something now (if we see it again).


vito:
The eval failures seem unrelated - been seeing those on `main` too. #10366 seems to fix it.

kpenfound (approved):
Nice I missed that the other package has been deprecated!

Noting here that I'd also love to get rid of the `convertToolsToGenai` because it'll definitely bite us someday, but at the time of implementing there was no way to easily pass in a json tool schema like we do with openai. Maybe this is an improvement we can make to upstream

jedevc (approved):

PR #10393, opened 2025-05-13
refactor: client gen dependency loading

TomChv:
Pass the `moduleSourceID` instead of a JSON file so it simplifies SDK implementation and let it rely on the codegen.

If a moduleSourceID is passed to the codegen CLI,
it will fetch its dependencies and store it in the codegen configuration.

Improve Go & SDK client codegen to only generate what is needed:
- If there's only remote dep, do not load any dagger.json, just serve git deps
- If there's local dependency, load the dagger.json, (throw an error or emit a warning if missing) and serve them
- If the module itself needs to be serve, load the dagger;json (same error handling as local) and serve it.

That way, all dependencies are sent to the codegen and then it can adapt its behaviour.

:bulb: I also added a quick fix on the Go client gen, it seems there was an issue when it was using the remote library, the `Close` method wasn't exposed by the published dagger client.


cmd/codegen/main.go
```diff
@@ -112,18 +116,43 @@ func ClientGen(cmd *cobra.Command, args []string) error {
 		cfg.IntrospectionJSON = string(introspectionJSON)
 	}
 
-	if dependenciesJSONFilePath != "" {
-		dependenciesJSON, err := os.ReadFile(dependenciesJSONFilePath)
+	if moduleSourceID != "" {
+		dag, err := dagger.Connect(ctx)
```
jedevc, line 120: Is this two separate connections to the engine, one in `codegen.go` as well? That's a bit confusing if so, we should only create one connection.

cmd/codegen/generator/generator.go
```diff
@@ -12,7 +12,7 @@ import (
 
 	"dagger.io/dagger"
 	"github.com/dagger/dagger/cmd/codegen/introspection"
-	"github.com/dagger/dagger/core/modules"
+	"github.com/dagger/dagger/core"
```
jedevc, line 15: Huh we should avoid depending on this. `core` is a huge module, I think if we can remove the dependency we should hopefully make the binary a lot smaller (it's only for `core.ModuleSourceKind`)

cmd/codegen/generator/go/templates/src/_dagger.gen.go/client.go.tmpl
```diff
@@ -213,48 +227,41 @@ func serveModuleDepencencies(ctx context.Context, client *Client) error {
 		Serve(ctx); err != nil {
 		return err
 	}
-	{{ end }}
+		{{ end -}}
+	{{- end -}}
 
+	{{/* Serve the local module if there are any local dependencies */}}
+	{{- if or (HasLocalDependencies) (ImplementModule) }}
```
jedevc, line 234: Can we do this? https://github.com/dagger/dagger/issues/9582#issuecomment-2786975469

We should always attempt to load the config.

cmd/codegen/main.go
```diff
@@ -112,18 +116,43 @@ func ClientGen(cmd *cobra.Command, args []string) error {
...
+				Data: &res,
+			})
+		if err != nil {
+			return fmt.Errorf("failed to load module source dependencies: %w", err)
+		}
+
+		cfg.ModuleDependencies = res.Source.Dependencies
+
+		// If the module source is implementing functions, we need to load the module itself
+		src, err := dag.LoadModuleSourceFromID(dagger.ModuleSourceID(moduleSourceID)).SDK().Source(ctx)
```
jedevc, line 150: This is a new feature right? We weren't previously serving our own source back right?

I wonder if because we now have a ModuleSourceID, if we could now use the new `includeDependencies`, instead of doing the manual loading: https://github.com/dagger/dagger/blob/cc5e55e8bed125117b241d204714a4353810af6a/docs/docs-graphql/schema.graphqls#L2803-L2812

jedevc, line 150: Implementation wise, this gets the source of the SDK, that doesn't seem right to me.

cmd/codegen/generator/go/templates/src/_dagger.gen.go/client.go.tmpl
TomChv: Why would we load the config if it's not useful because it's only remote dependencies?

cmd/codegen/main.go
TomChv: Yeah I could pass the `dag` as parameter to the `generate` interface function, so the caller is reponsible for opening/closing the connection.

cmd/codegen/main.go
TomChv: I'm not sure I understood this comment.

The `Source` is a simple string field that point to where the SDK is.
It's only useful for the codegen to know if the module we generate for implements its own functions.

I'm not sure we can run `serve` on a module that has no source.

TomChv: How could I check if the module I generate for implements functions then? If it does, I need to serve it, otherwise I shouldn't

cmd/codegen/generator/go/templates/src/_dagger.gen.go/client.go.tmpl
jedevc: Because the `dagger.json` might have changed, and the code might not have been regenerated. Then the code will not match up with the version that's being served, and we can get very confusing errors.

We should validate against the `dagger.json` *if it's there*. If it's not, and we have only remote dependencies, then it's fine, we can just continue on. But if it's there, why not use it for validation?

cmd/codegen/main.go
jedevc: Oh, I see what you mean :thinking: the comment is confusing because a module can technically have an SDK and not implement any functions, this isn't what we're testing for.

Why couldn't we try and `Serve` an SDK-less module? It works :thinking: (and also means we need to handle fewer corner cases in each SDK)

jedevc: > I'm not sure I understood this comment.

The comment is probably not on the right line.

I have two points:
- Were we previously serving + generating the API of our own module back to ourselves? Or is this new?
- If we *are* serving our own API, can we use the method I link to? Does that make sense? It feels like we're implementing lots of the same logic as we've already implemented there.

TomChv: I'll check and see if it can work

TomChv: It's new, in module we never generated the API of our own module.
I'm gonna check if we can serve an SDK less dependencies or not, because it might simplify a lot the resolution if we can, I could just add "include-dependencies" and boom

TomChv: It works! That's so nice!

TomChv: It works! I'm using the include deps so I do not even need to serve all modules in a loop

cmd/codegen/main.go
TomChv: Fixed, I added a field `Dag` in the config because depending on the context we cannot connect to the engine.
So it checks which flag are set and will connect to the engine if it makes sense to

jedevc (approved):
Would be good to have a test case for the scenario when the code is generated, but then the `dagger.json` is updated (and the pin data doesn't match).

This should error out, can we have a check for that?

TomChv:
> Would be good to have a test case for the scenario when the code is generated, but then the dagger.json is updated (and the pin data doesn't match).

> This should error out, can we have a check for that?

I'll add that before merging!

PR #10396, opened 2025-05-13
Automatically upgrade dagger-cloud-api on engine releases

matipan:
As a part of the release process we automatically upgrade daggerverse using the `modules/daggerverse` module. Since the development of the new Module Catalog & Insights feature the dagger cloud API now has `daggerverse` itself as a dependency. Because of this, after we upgrade daggerverse we must also run `go mod tidy` on the API itself, else we leave it broken on our main production instance. This upgrade also needs to include Kubernetes deployments and github workflows.

### Why the new module?
I thought of a few alternatives, the two worth mentioning:
- Have a central dagger-io module with logic to bump dagger versions in Go applications, agnostic to the app itself. Then have a modules/daggerverse and modules/dagger-cloud-api that import dagger-io and use it bump it on their corresponding Go apps
- Have a central dagger-io module that takes care of everything on one go and has app-specific functions for app-specific operations.

I went with the second approach, replacing `modules/daggerverse` with `modules/dagger-io` and taking care of all dagger upgrades in a single dagger call. This implementation was much simpler, required less code and achieved the same task. It also allows us to treat the Dagger upgrade as a single deployable unit within dagger.io, which is definitely what we want.

> [!NOTE]
> We don't need to validate dagger-cloud-api on a preview environment. If daggerverse's crawling works then dagger-cloud-api will also work since it depends on the exact same underlying code.


jedevc:
Question - WDYT about maybe moving this module into `dagger/dagger.io`? It feels weird that the module requires so much knowledge of the internals of the repo, but that we run it from here (it also makes it harder to update the update logic, so it can sometimes drift).

We could call the remote module from inside our github action here.

matipan:
> Question - WDYT about maybe moving this module into `dagger/dagger.io`? It feels weird that the module requires so much knowledge of the internals of the repo, but that we run it from here (it also makes it harder to update the update logic, so it can sometimes drift).
> 
> We could call the remote module from inside our github action here.

@jedevc I'm all for that. I went with this approach to keep things consistent but I do agree its better to keep all that logic there and only do the triggering here! 👍🏼

marcosnils (approved):

Issue #9584, opened 2025-02-14
.env support & default secrets


shykes:
## Overview

Dagger should natively support `.env` files. It's a [de facto standard](https://www.dotenv.org/docs/security/env.html) for managing environment-specific configuration in a lightweight, portable way. It is widely supported, included by `docker` and `docker-compose`.

### Default behavior

If `.env` exists in the current directory, or any parent directory, `dagger` should load it by default.

### --env-file

An env file can be specified explicitly with `dagger --env-file PATH`.

### Secret references

Regular env file must store a secret value in plaintext. But Dagger allows you to keep only the *reference* of the secret in the file, minimizing the risk of secret leaks, and making it easier to rotate credentials.

Any variable value that matches the format of a Dagger secret reference, will automatically be parsed as a secret URI, and securely resolved by the engine.

For example:

```
USERNAME=admin
PASSWORD=op://Private/foo/credential
PROD_AWS_TOKEN=vault://foo/bar/prod/token
```

### CLI usage

To use the variables with the `dagger` CLI:

- In `dagger shell`, simply reference the variable anywhere in your commands with `$FOO` or `${FOO}`.
- In `dagger call`, this also works, but remember to use single-quotes around arguments containing a variable expansion, to prevent your system shell from doing the expansion with its own variables - unless that's what you want!

In both cases, variable expansion *only uses the contents of the env file*. The host environment variables are never used.

### Usage by Dagger Functions: default secrets

Env files are an acceptable sandbox for env variables, in the same way that the git context is an acceptable sandbox for local directories.

This means that, once we support Env files, and as long as we forbid escaping the sandbox, we can now allow functions to define a default env variable for their secrets.

For example:

```golang
// Login to Dagger Cloud
function (m *DaggerCloud) Login(
  // Dagger Cloud token for authentication  
  // +default="$DAGGER_CLOUD_TOKEN"
token *dagger.Secret,
  ) error {
    // ....
  }
```

jedevc:
Sounds good :tada:

1. My main question is whether transitive dependencies should be able to use default secrets from the original caller. E.g. CLI calls module A, module A calls module B, module B has a default secret - is it allowed to access it?

   It feels like the feature is a little lacking if it can't (though you can always just explicitly pass the secret between A and B), but if you can, it feels like a huge ecosystem vulnerability waiting to happen.

[...]

PR #10397, opened 2025-05-13
Add support for environment variables to function args

jasonmccallister:
This PR adds the ability to define `+defaultEnv="FOO"` to function arguments to reduce verbosity and improve developer experience.

There are a few things we need to discuss in terms of implementation but I wanted to get the ball rolling.

### DIFF
--- a/cmd/codegen/generator/go/templates/module_funcs.go
+++ b/cmd/codegen/generator/go/templates/module_funcs.go
@@ -6,7 +6,6 @@ import (
 	"go/ast"
 	"go/types"
 	"maps"
-	"os"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


jedevc:
x-linking https://github.com/dagger/dagger/issues/9584 (though I know we've drifted away a bit from working with `.env` out-of-the-box, cc @shykes) 

So the main reason we haven't done this before is that we need a way to resolve the security issue here: a user running `dagger call potentially-bad-module` shouldn't have to worry about the module hoovering up environment variables (that potentially have valuable secrets).

The other unresolved design question IMO is about *which* modules can access this - is it just the top-level one, or can any dependency module?

rawkode:
> The other unresolved design question IMO is about which modules can access this - is it just the top-level one, or can any dependency module?

In a mono repository, pretty much everything becomes a dependency module.

Could we provide the means within the `dagger.json`, where dependent modules are defined, of explicitly saying that we give a module access to a set of environment variables?

```json
 "dependencies": [
    {
      "name": "infisical",
      "source": "github.com/jpadams/daggerverse/infisical@f89b7240f7e711a323dae5a66df9b78dfe6a4ec2",
      "env": [ "INFISICAL_TOKEN" ],
    }
  ],
```

nipuna-perera:
Doesn't `.env` support solve both the issues (security and ergonomics). For example a `.env` with the following

```env
pipelineUser=env://PIPELINE_USR
pipelinePassword=op://PIPELINE_PSW
notACredential=test
```

No need to specify a pragma in code, if there is a parameter with the same name (eg: `pipelineUser`) AND if the `.env` value is present, use that. In my experience, if you are taking a credential as a parameter, you want that to be substituted for the whole module chain. 

If a pragma is required, then something like `+fromEnv` to indicate to dagger to use `.env` would work too.

PR #10400, opened 2025-05-14
workflows: Change Alt Runners 4

fidencio:
After a long week of experimentations, I think we were able to trim down a little bit the expected values, thus we're bringing here 3 different "types" that should demonstrate such potential improvement.

Together with this change, we're trying to stick to run what the Alt Runners 1 are running:
* doc-lint
* sdk-go
* run those in sequence (not parallel)

Ideally someone from the community side should be able to take over this PR, and test this against the GHA already existent before getting it pushed.


fidencio:
Considering the 0.18.8 release from Yesterday, let me know if you prefer that I update the runners to 0.18.8 before any testing on your side, @gerhard.

gerhard:
This is looking great @fidencio 💪

We should test with `0.18.7`, and upgrade to `0.18.8` as soon as https://github.com/dagger/dagger/pull/10409 gets merged.

I am going to push this on a branch so that I can test the workflow change before I get it merged.

gerhard:
I made on tiny change to fix this: https://github.com/dagger/dagger/actions/runs/15044070207/job/42282250611#step:3:33

The first job is now passing, and I'm now hitting this on `sdk-go-on-nf-type-01-0-18-7-remote-engine`: https://github.com/dagger/dagger/actions/runs/15044121177/job/42282605438#step:3:50

image

.github/workflows/alternative-ci-runners-4.yml
```diff
@@ -13,37 +13,41 @@ permissions:
   contents: read
 
 jobs:
-  docs-lint-on-northflank-kata-remote-engine:
+  docs-lint-on-nf-type-01-0-18-7-remote-engine:
```
gerhard, line 16: What are your thoughts @fidencio on removing `0-18-7` from the job name? Keeping this means that we will have a new variant in our metrics that will make it harder to visualise trends.

.github/workflows/alternative-ci-runners-4.yml
```diff
@@ -13,37 +13,41 @@ permissions:
...
+#    with:
+#      machine: nf-type-02-0.18.7
+#      function: check --targets=docs
+#
+#  sdk-go-on-nf-type-02-0-18-7-remote-engine:
+#    needs: docs-lint-on-nf-type-02-0-18-7-remote-engine
+#    uses: ./.github/workflows/_dagger_on_northflank_remote_engine.yml
+#    with:
+#      machine: nf-type-02-0.18.7
+#      function: check --targets=sdk/go
```
gerhard, line 40: I'm a fan of keeping things clean & removing this. We can always add it back when we are ready to re-enable it.

.github/workflows/alternative-ci-runners-4.yml
fidencio: That works for me, let me update the PR once I have this set up on my side (for all the other runners as well)

.github/workflows/alternative-ci-runners-4.yml
fidencio: Sure, i will drop those as well.

fidencio:
@gerhard, PR updated now that the test base was moved to 0.18.8 as well.
Thanks!

gerhard (approved):
👍 🚀

PR #10408, opened 2025-05-14
chore: deduplicate directory solve code

alexcb:
This unifies the codepath directory methods execute when performing a buildkit solve


alexcb:
This PR is causing failures :(

it seems to occur when I change the Directory.Entries call to use Evaluate. it _might_ be related to Evaluate has an early return:

```
if dir.LLB == nil {                                                              
    return nil, nil                                                              
}  
```

which then causes `res` to be nil when it reaches

```
	ref, err := res.SingleRef()
	if err != nil {
		return nil, err
	}
```

PR #10411, opened 2025-05-15
fix: ensure dagger develop/publish do not take args

jedevc:
Running `dagger develop path/to/module` is an easy typo - however, it's not valid, you need to use the `-m` flag. This didn't error out though, we just accepted any additional args and ignored them.

We should explicitly forbid args to these functions if we're not using them, which will help prevent typos.


jedevc:
cc @jpadams just wanted to check this is how `dagger publish` is intended to work?

helderco (approved):
lgtm

Issue #10398, opened 2025-05-14
✨ (sdk/java) better handle errors/exceptions


eunomie:
### What are you trying to do?

When an error is raised during execution (for instance raised by a `withExec`) the `DaggerQueryException` raised doesn't allow to get access to the underlying error easily.

The exception should have an easy way to access the underlying error without to deal with graphql errors. Default message should be improved and we should have access to the details of the error.

For instance, here is a Go code from the [cookbook](https://docs.dagger.io/cookbook/#terminate-gracefully) about error handling:

```go
func (m *MyModule) Test(ctx context.Context) (string, error) {
	out, err := dag.Container().From("alpine").
		// ERROR: cat: read error: Is a directory
		WithExec([]string{"cat", "/"}).
		Stdout(ctx)
	var e *dagger.ExecError
	if errors.As(err, &e) {
		return fmt.Sprintf("Test pipeline failure: %s", e.Stderr), nil
	} else if err != nil {
		return "", err
	}
	return out, nil
}
```

The error raised can be read _as_ a `ExecError` with access to `stderr` for instance. The Java SDK currently doesn't support something like that.

The Java equivalent is currently:

```java
  @Function
  public String test() {
    try {
      String out = dag().container().from("alpine")
              // ERROR: cat: read error: Is a directory
              .withExec(List.of("cat", "/"))
              .stdout();
      return out;
    } catch (DaggerQueryException e) {
      return "Test pipeline failure: %s".formatted(e.getMessage());
    } catch (InterruptedException | ExecutionException e) {
      throw new RuntimeException(e);
    }
  }
```

But the `getMessage` doesn't return the `stderr`.

That said, the detail exists and the underlying error raised by graphql is:

```
[Error{
  message=process "cat /" did not complete successfully: exit code: 1,
  locations=null,
  path=[container, from, withExec, stdout],
  extensions={
    stdout=, 
    _type=EXEC_ERROR,
    exitCode=1,
    cmd=["cat","/"],
    stderr=cat: read error: Is a directory
  }
}]
```

Currently only the `message` is available. We should alto expose the `path` and have a way to expose the `extensions` based on the `_type` so that it's easy to print nice default error message as well as a way to get access to the different fields easily.

My understanding is the only specific error is `EXEC_ERROR` but we should have a way to specify some more later.

[...]

PR #10414, opened 2025-05-15
[#10398] DQE - added methods to improve error messages

robertonav20:
Hi @eunomie 
i've added these methods

- default: i add also the `path` and `extension.type` field i think could be useful
- `toEnanchedMessage` which contains `message, path, type code, exit code and cmd`
- `toFullMessage` which contains `message, path, type code, exit code, cmd and STDERR`

Below the patterns of messages
```java
  private static final String SIMPLE_MESSAGE = "Message: [%s]\nPath: [%s]\nType Code: [%s]\n";
  private static final String ENANCHED_MESSAGE =
      "Message: [%s]\nPath: [%s]\nType Code: [%s]\nExit Code: [%s]\nCmd: [%s]\n";
  private static final String FULL_MESSAGE =
      "Message: [%s]\nPath: [%s]\nType Code: [%s]\nExit Code: [%s]\nCmd: [%s]\nSTDERR: [%s]\n";
```

Let me know if is ok or is should review something
Thanks in advance

### DIFF
--- a/sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/DaggerQueryException.java
+++ b/sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/DaggerQueryException.java
@@ -1,10 +1,23 @@
 package io.dagger.client;
 
-import io.smallrye.graphql.client.GraphQLError;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
+import io.smallrye.graphql.client.GraphQLError;
 
 public class DaggerQueryException extends Exception {
+
+  private static final String SIMPLE_MESSAGE = "Message: [%s]\nPath: [%s]\nType Code: [%s]\n";
+  private static final String ENANCHED_MESSAGE =
+      "Message: [%s]\nPath: [%s]\nType Code: [%s]\nExit Code: [%s]\nCmd: [%s]\n";
+  private static final String FULL_MESSAGE =
+      "Message: [%s]\nPath: [%s]\nType Code: [%s]\nExit Code: [%s]\nCmd: [%s]\nSTDERR: [%s]\n";
+
+  protected static final String TYPE_KEY = "_type";
+  protected static final String EXIT_CODE_KEY = "exitCode";
+  protected static final String CMD_KEY = "cmd";
+  protected static final String STDERR_KEY = "stderr";
 
   private GraphQLError[] errors;
 
@@ -14,11 +27,34 @@ public class DaggerQueryException extends Exception {
   }
 
   public DaggerQueryException(GraphQLError... errors) {
-    super(Arrays.stream(errors).map(GraphQLError::getMessage).collect(Collectors.joining("\n")));
+    super(Arrays
+        .stream(errors).map(e -> String.format(SIMPLE_MESSAGE, e.getMessage(),
+            StringUtils.join(e.getPath(), ","), e.getExtensions().getOrDefault(TYPE_KEY, null)))
+        .collect(Collectors.joining("\n")));
     this.errors = errors;
   }
 
   public GraphQLError[] getErrors() {
     return errors;
+  }
+
+  public String toEnanchedMessage() {
+    return Arrays.stream(errors).map(e -> String.format(ENANCHED_MESSAGE, e.getMessage(),
+        StringUtils.join(e.getPath(), ","), e.getExtensions().getOrDefault(TYPE_KEY, null),
+        e.getExtensions().getOrDefault(EXIT_CODE_KEY, null),
+        StringUtils.join(
+            ((Object[]) e.getExtensions().getOrDefault(CMD_KEY, Collections.emptyList())), ",")))
+        .collect(Collectors.joining("\n"));
+  }
+
+  public String toFullMessage() {
+    return Arrays.stream(errors)
+        .map(e -> String.format(FULL_MESSAGE, e.getMessage(), StringUtils.join(e.getPath(), ","),
+            e.getExtensions().getOrDefault(TYPE_KEY, null),
+            e.getExtensions().getOrDefault(EXIT_CODE_KEY, null),
+            StringUtils.join(
+                ((Object[]) e.getExtensions().getOrDefault(CMD_KEY, Collections.emptyList())), ","),
+            e.getExtensions().getOrDefault(STDERR_KEY, null)))
+        .collect(Collectors.joining("\n"));
   }
 }
--- a/sdk/java/dagger-java-sdk/src/test/java/io/dagger/client/DaggerQueryExceptionTest.java
+++ b/sdk/java/dagger-java-sdk/src/test/java/io/dagger/client/DaggerQueryExceptionTest.java
@@ -1,0 +1,92 @@
+package io.dagger.client;
+
+import static io.dagger.client.DaggerQueryException.CMD_KEY;
+import static io.dagger.client.DaggerQueryException.EXIT_CODE_KEY;
+import static io.dagger.client.DaggerQueryException.STDERR_KEY;
+import static io.dagger.client.DaggerQueryException.TYPE_KEY;
+import static org.assertj.core.api.Assertions.assertThat;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+import io.smallrye.graphql.client.GraphQLError;
+
+
+public class DaggerQueryExceptionTest {
+
+    @Test
+    void shouldReturnDefaultMessage() {
+        GraphQLError error =
+                buildError("ERROR", new Object[] {"container", "from", "withExec", "stdout"},
+                        Map.of(TYPE_KEY, "EXEC_ERROR"));
+        GraphQLError error2 = buildError("ERROR2",
+                new Object[] {"container", "from", "withExec", "withExec", "stdout"},
+                Map.of(TYPE_KEY, "EXEC_ERROR"));
+
+        String result = new DaggerQueryException(error, error2).getMessage();
+        String expected =
+                "Message: [ERROR]\nPath: [container,from,withExec,stdout]\nType Code: [EXEC_ERROR]\n\nMessage: [ERROR2]\nPath: [container,from,withExec,withExec,stdout]\nType Code: [EXEC_ERROR]\n";
+        assertThat(result).isEqualTo(expected);
+    }
+
+    @Test
+    void shouldReturnEnanchedMessage() {
+        GraphQLError error = buildError("ERROR",
+                new Object[] {"container", "from", "withExec", "stdout"}, Map.of(TYPE_KEY,
+                        "EXEC_ERROR", EXIT_CODE_KEY, "1", CMD_KEY, new Object[] {"cat", "WRONG"}));
+        GraphQLError error2 = buildError("ERROR2",
+                new Object[] {"container", "from", "withExec", "withExec", "stdout"},
+                Map.of(TYPE_KEY, "EXEC_ERROR", EXIT_CODE_KEY, "2", CMD_KEY,
+                        new Object[] {"cat", "WRONG2"}));
+
+        String result = new DaggerQueryException(error, error2).toEnanchedMessage();
+        String expected =
+                "Message: [ERROR]\nPath: [container,from,withExec,stdout]\nType Code: [EXEC_ERROR]\nExit Code: [1]\nCmd: [cat,WRONG]\n\nMessage: [ERROR2]\nPath: [container,from,withExec,withExec,stdout]\nType Code: [EXEC_ERROR]\nExit Code: [2]\nCmd: [cat,WRONG2]\n";
+        assertThat(result).isEqualTo(expected);
+    }
+
+    @Test
+    void shouldReturnFullMessage() {
+        GraphQLError error =
+                buildError("ERROR", new Object[] {"container", "from", "withExec", "stdout"},
+                        Map.of(TYPE_KEY, "EXEC_ERROR", EXIT_CODE_KEY, "1", CMD_KEY,
+                                new Object[] {"cat", "WRONG"}, STDERR_KEY, "DEEP ERROR DETAILS"));
+        GraphQLError error2 = buildError("ERROR2",
+                new Object[] {"container", "from", "withExec", "withExec", "stdout"},
+                Map.of(TYPE_KEY, "EXEC_ERROR", EXIT_CODE_KEY, "2", CMD_KEY,
+                        new Object[] {"cat", "WRONG2"}, STDERR_KEY, "DEEP ERROR DETAILS2"));
+
+        String result = new DaggerQueryException(error, error2).toFullMessage();
+        String expected =
+                "Message: [ERROR]\nPath: [container,from,withExec,stdout]\nType Code: [EXEC_ERROR]\nExit Code: [1]\nCmd: [cat,WRONG]\nSTDERR: [DEEP ERROR DETAILS]\n\nMessage: [ERROR2]\nPath: [container,from,withExec,withExec,stdout]\nType Code: [EXEC_ERROR]\nExit Code: [2]\nCmd: [cat,WRONG2]\nSTDERR: [DEEP ERROR DETAILS2]\n";
+        assertThat(result).isEqualTo(expected);
+    }
+
+    private GraphQLError buildError(String message, Object[] path, Map<String, Object> extensions) {
+        return new GraphQLError() {
+            @Override
+            public String getMessage() {
+                return message;
+            }
+
+            @Override
+            public List<Map<String, Integer>> getLocations() {
+                return null;
+            }
+
+            @Override
+            public Object[] getPath() {
+                return path;
+            }
+
+            @Override
+            public Map<String, Object> getExtensions() {
+                return extensions;
+            }
+
+            @Override
+            public Map<String, Object> getOtherFields() {
+                return null;
+            }
+        };
+    }
+}


robertonav20:
Hi @eunomie,
thanks for your response and review.
I'm glad to help dagger community.

i'm trying to understand your point of view:
- DaggerExecException should be a subtype of DaggerQueryException... you mean DaggerExecException should extends DaggerQueryException and raised up with withExec error as i made? right? In this way the default message of exception what's should be?
- DaggerExecException default message is fine or should we change it?
- DaggerExceptionConstants/DaggerExceptionUtils should are inside DaggerException or DaggerQueryException... my perspective is to split the code in 2 points to give more semantic and test them separately... i don't know give me your feedback
- about messages they are just placeholders the user could choose if use them or not from my pov (customer) they contains all basilar information to understand eventually issue
- Last point "the different fields under DaggerExecException than the concatenation string" what do you have in mind? an exception with prefilled fields like type, exitCode, path and so on then expose methods to concatenate them? could you show me a little example?

If you want, i'm available for a fast call

Thanks in advance

eunomie:
Sorry @robertonav20 I think my message was not clear enough 😅 
 
> * DaggerExecException should be a subtype of DaggerQueryException... you mean DaggerExecException should extends DaggerQueryException

Yes.  `withExec` is a query. So queries can raise a `DaggerQueryException`. In case of `withExec` (I think that's the only case right now) we want a more specific one because we want to surface certain data sent by the engine. So I would make more sense to have `DaggerExecException` extending `DaggerQueryException`.

This also means we can avoid all those changes:

```diff
-      throws ExecutionException, DaggerQueryException, InterruptedException {
+      throws ExecutionException, DaggerExecException, DaggerQueryException, InterruptedException {
```

In many cases we don't care that much about the difference between both, but we would like to specify only when required.

>  and raised up with withExec error as i made? right? In this way the default message of exception what's should be?
> * DaggerExecException default message is fine or should we change it?
> * about messages they are just placeholders the user could choose if use them or not from my pov (customer) they contains all basilar information to understand eventually issue
> * Last point "the different fields under DaggerExecException than the concatenation string" what do you have in mind? an exception with prefilled fields like type, exitCode, path and so on then expose methods to concatenate them? could you show me a little example?

In fact I'm not that interested in the messages :-)
What I found interesting is to have a direct access to the different fields from the original error (and the original error itself).

I have 2 cases in mind.

The first one is this example in Go from the [cookbook](https://docs.dagger.io/cookbook#terminate-gracefully)

```go
func (m *MyModule) Test(ctx context.Context) (string, error) {
	out, err := dag.
		Container().
		From("alpine").
		// ERROR: cat: read error: Is a directory
		WithExec([]string{"cat", "/"}).
		Stdout(ctx)
	var e *dagger.ExecError
	if errors.As(err, &e) {
		return fmt.Sprintf("Test pipeline failure: %s", e.Stderr), nil
	} else if err != nil {
		return "", err
	}
	return out, nil
}
```

In Java that could be something like:

```java
@Function
public string test() throws ExecutionException, InterruptedException, DaggerQueryException {
  try {
    return dag().container().from("alpine").withExec(List.of("cat", "/")).Stdout();
  } catch (DaggerExecException e) {
    throw new RuntimeException("Test pipeline failure: "+e.stderr());
  }
}
```

And behind that the `DaggerExecException` should have the following:

```java
public class DaggerExecException {
  public List<String> cmd() {}
  public String stdout() {}
  public String stderr() {}
}
```

(or with `get*` if that sounds more Java)

That's what I mean by to have access to the different fields.

This is basically a reimplementation of the equivalent Go type:

```go
// ExecError is an API error from an exec operation.
type ExecError struct {
	original *gqlerror.Error
	Cmd      []string
	ExitCode int
	Stdout   string
	Stderr   string
}
```

And if we want to be as close as possible, we should probably also just keep the message as received:

```go
func (e *ExecError) Extensions() map[string]any {
	return e.original.Extensions
}

func (e *ExecError) Message() string {
	return e.original.Error()
}
```

It's still possible to add other functions with predefined messages if that's useful.

To be close to the Go version is interesting because it means the visible behaviour is very close whatever the SDK used.

The other use case I'm seeing (and I think that will solve the initial issue that is to see the underlying error) is to correctly return the error in case of a `withExec` error (by return I mean serialise).

I'll take the generated Go code as an example to explain what should happen and where the Java SDK currently misses the point.

```go
	result, err := invoke(ctx, []byte(parentJson), parentName, fnName, inputArgs)
	if err != nil {
		var exec *dagger.ExecError
		if errors.As(err, &exec) {
			return exec.Unwrap()
		}
		return err
	}
```
`invoke` is the method that will run the user code. So if there's a `dagger.ExecError` we unwrap it, returning the underlying graphql error.

Then this is catch by this code:

```go
	defer func() {
		if rerr != nil {
			if err := fnCall.ReturnError(ctx, convertError(rerr)); err != nil {
				fmt.Println("failed to return error:", err, "\noriginal error:", rerr)
			}
		}
	}()
```

With the `convertError` function as follows:

```go
	var gqlErr *gqlerror.Error
	if errors.As(rerr, &gqlErr) {
		if gqlErr.Extensions != nil {
				keys = append(keys, k)
			}
			sort.Strings(keys)
			for _, k := range keys {
				if err != nil {
				}
			}
		}
		return dagErr
	}
}
```

This is the very interesting part.
So in case of a graphql error (and so in an error raised by `withExec` we are constructing a specific `dag.Error` with the message and the different fields. This will be interpreted by the engine and the client to display the correct error to the user (and solve the main issue).
Right now the generated Java code is the following:

```java
    } catch (InvocationTargetException e) {
      fnCall.returnError(dag().error(e.getTargetException().getMessage()));
      throw e;
    } catch (Exception e) {
      fnCall.returnError(dag().error(e.getMessage()));
      throw e;
    }
```

We should have something more like:

```java
} catch (DaggerQueryException e) {
  fnCall.returnError(dag().error(e.getTargetException().getMessage()).withValue(..., ...).withValue(..., ...))
}
```

With all the `extensions` values.


> * DaggerExceptionConstants/DaggerExceptionUtils should are inside DaggerException or DaggerQueryException... my perspective is to split the code in 2 points to give more semantic and test them separately... i don't know give me your feedback

No big deal here, just a feeling while reading it, but it's been a long time I haven't followed "classical java code style" and years of Go obviously changed my mind on how to structure some parts of the code. But that's definitely not a problem, if you think that helps to give more sense and test, this is totally fine.

---

I hope all that makes more sense, sorry I should have written more precisely what I had in mind.

robertonav20:
Hi @eunomie 
thanks for your explanation I will try to reach your request.... Don't worry, Is my fault if didn't understand at first time depends on my knowledge of dagger and dagger sdk

So just a recap
- constant class and utils  => OK
- placeholder messages are ok but they should not be default message of exceptions
- add builder pattern to exceptions to give more control on what should be displayed in cas of errors
- dagger exec exception should extends dagger query exception and remove dagger exception 

This lists should summarise the changes 😎

eunomie:
> * add builder pattern to exceptions to give more control on what should be displayed in cas of errors

No, not really. In fact we need to change the method that handles the exception and adds the graphql error fields to the `dag.Error` we are sending. This one already contains the `with*` methods.

This is part of the generated code, here: https://github.com/dagger/dagger/blob/307b910653f1eb0b2d1f9bdd0b0f7ccb0c1c5664/sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java#L523-L534

But maybe it's best to have it in a different PR to simplify things. Codegen is one extra layer of complexity 🫣

robertonav20:
Ok let's see step by step..
 Anyway I already change it to add the exec exception i think should not be too hard to do it

robertonav20:
Hi @eunomie,
i have made the changes you requested, I believe they are close to your goal.... let me know

thanks in advance

docs/current_docs/features/snippets/debugging-1/java/src/main/java/io/dagger/modules/mymodule/MyModule.java
```diff
@@ -11,7 +11,7 @@
 @Object
 public class MyModule {
   @Function
-  public String foo() throws ExecutionException, DaggerQueryException, InterruptedException {
+  public String foo() throws ExecutionException, DaggeQueryException, InterruptedException {
```
eunomie:
```suggestion
  public String foo() throws ExecutionException, DaggerQueryException, InterruptedException {
```

Just a small typo here

docs/current_docs/quickstart/agent/snippets/part2/workspace/java/src/main/java/io/dagger/modules/workspace/Workspace.java
```diff
@@ -52,27 +52,18 @@ public Workspace writeFile(String path, String contents) {
...
   public String listFiles() throws ExecutionException, DaggerQueryException, InterruptedException {
-    return dag()
-      .container()
-      .from("alpine:3")
-      .withDirectory("/src", source)
-      .withWorkdir("/src")
-      .withExec(List.of("tree", "/src"))
-      .stdout();
+    return dag().container().from("alpine:3").withDirectory("/src", source).withWorkdir("/src")
+        .withExec(List.of("tree", "/src")).stdout();
```
eunomie: Not necessarily everywhere, but I think if we can keep the multiline version it would be better.
This might be against the default formatter, but this is only for the docs and I think it's easier to read and to compare with other languages

docs/current_docs/quickstart/agent/snippets/part2/workspace/java/src/main/java/io/dagger/modules/workspace/Workspace.java
```diff
@@ -52,27 +52,18 @@ public Workspace writeFile(String path, String contents) {
...
-    return dag().container()
-        .from("node:21-slim")
-        .withDirectory("/src", source)
-        .withMountedCache("/root/.npm", nodeCache)
-        .withWorkdir("/src")
-        .withExec(List.of("npm", "install"))
-        .withExec(List.of("npm", "run", "test:unit", "run"))
+    return dag().container().from("node:21-slim").withDirectory("/src", source)
+        .withMountedCache("/root/.npm", nodeCache).withWorkdir("/src")
+        .withExec(List.of("npm", "install")).withExec(List.of("npm", "run", "test:unit", "run"))
```
eunomie: Same as above, I think better to keep the multiline version here.

sdk/java/dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/ObjectVisitor.java
```diff
@@ -225,30 +225,34 @@ private void buildFieldMethod(
       fieldMethodBuilder
           .addException(InterruptedException.class)
           .addException(ExecutionException.class)
-          .addException(ClassName.bestGuess("DaggerQueryException"));
+          .addException(ClassName.get("io.dagger.client.exception", "DaggerQueryException"))
+          .addException(ClassName.get("io.dagger.client.exception", "DaggerExecException"));
```
eunomie: Just wondering if we need both explicitly or the `DaggerQueryException` should cover them

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
```diff
@@ -526,6 +549,22 @@ static JavaFile generate(ModuleInfo moduleInfo) {
                                   "fnCall.returnError($T.dag().error(e.getTargetException().getMessage()))",
                                   Dagger.class)
                               .addStatement("throw e")
+                              .nextControlFlow("catch ($T e)", DaggerExecException.class)
+                              .addStatement(
+                                  "fnCall.returnError($T.dag().error(e.getMessage())"
+                                      + ".withValue(\"exitCode\", $T.from($T.join(e.getExitCode())))"
+                                      + ".withValue(\"path\", $T.from($T.join(e.getPath())))"
+                                      + ".withValue(\"cmd\", $T.from($T.join(e.getCmd())))"
+                                      + ".withValue(\"stderr\", $T.from(e.getStdErr())))",
```
eunomie: I think we are really close but not exactly right.

Here is the Go code handling the error:

```go
			for _, k := range keys {
				if err != nil {
				}
			}
```

What it'd like to point out here is we should not handle them as string but marshal them to their json representation. For instance an exit code of `1` should remain `1` with here it will be a list of string, so something like `["1"]`.

Also to notice that `JSON.from` doesn't convert a value to its json representation, it only casts a  java string into a `JSON` type (that is a string). So the conversion to json needs to happen before.

Also, I think the types are wrong. If `cmd` is a list of string, the others are either string or int.

For reference here is the Go code that converts a graphql error into an `ExecError`. That helps to understand what are the underlying types:

```go
func getCustomError(err error) error {
	var gqlErr *gqlerror.Error

	if !errors.As(err, &gqlErr) {
		return nil
	}

	ext := gqlErr.Extensions

	typ, ok := ext["_type"].(string)
	if !ok {
		return nil
	}

	if typ == "EXEC_ERROR" {
		e := &ExecError{
			original: err,
		}
		if code, ok := ext["exitCode"].(float64); ok {
			e.ExitCode = int(code)
		}
		if args, ok := ext["cmd"].([]interface{}); ok {
			cmd := make([]string, len(args))
			for i, v := range args {
				cmd[i] = v.(string)
			}
			e.Cmd = cmd
		}
		if stdout, ok := ext["stdout"].(string); ok {
			e.Stdout = stdout
		}
		if stderr, ok := ext["stderr"].(string); ok {
			e.Stderr = stderr
		}
		return e
	}

	return nil
}

// ExecError is an API error from an exec operation.
type ExecError struct {
	original error
	Cmd      []string
	ExitCode int
	Stdout   string
	Stderr   string
}
```

As an example here is an error raised during a withExec (this is the graphql error we are receiving):

```graphql
[Error{
  message=process "cat /" did not complete successfully: exit code: 1,
  locations=null,
  path=[container, from, withExec, stdout],
  extensions={
    stdout=, 
    _type=EXEC_ERROR,
    exitCode=1,
    cmd=["cat","/"],
    stderr=cat: read error: Is a directory
  }
}]
```

eunomie (commented):
Thanks @robertonav20 👍 
I think we are pretty close. The last things I caught is the way we are typing the values from the graphql error and the missing json conversion.

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
eunomie: To perform the conversion into json you should be able to use `JsonConverter.toJSON` that takes an object and return a `JSON` type directly (this is basically a `JSON.from(JsonBuilder.create().toJson(object))` with some extra configuration)

docs/current_docs/features/snippets/debugging-1/java/src/main/java/io/dagger/modules/mymodule/MyModule.java
robertonav20: yes sorry but the formatter is crazy not manage this situation :(

docs/current_docs/quickstart/agent/snippets/part2/workspace/java/src/main/java/io/dagger/modules/workspace/Workspace.java
robertonav20: formatter issue

docs/current_docs/quickstart/agent/snippets/part2/workspace/java/src/main/java/io/dagger/modules/workspace/Workspace.java
robertonav20: formatter issue

sdk/java/dagger-java-annotation-processor/src/main/java/io/dagger/annotation/processor/DaggerModuleAnnotationProcessor.java
robertonav20: i used JsonConverter.toJSON method for each field

robertonav20:
Hi @eunomie,
now should be ok, let me know if there is something to fix 😎 !

Thanks for your precious support!

sdk/java/dagger-codegen-maven-plugin/src/main/java/io/dagger/codegen/introspection/ObjectVisitor.java
robertonav20: yes it was enough! i had a doubt about that because i think there was a specific catch somewhere

sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/exception/DaggerExceptionUtils.java
```diff
@@ -0,0 +1,96 @@
...
+    return null;
+  }
+
+  public static String getType(GraphQLError error) {
+    return String.valueOf(getExtensionValueByKey(error, TYPE_KEY));
+  }
+
+  public static String getExitCode(GraphQLError error) {
+    return String.valueOf(getExtensionValueByKey(error, EXIT_CODE_KEY));
+  }
```
eunomie: This should return an integer, not a string

sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/exception/DaggerExceptionUtils.java
```diff
@@ -0,0 +1,96 @@
...
+    return StringUtils.join(error.getPath(), ".");
+  }
+
+  public static String getCmd(GraphQLError error) {
+    Object cmdList = getExtensionValueByKey(error, CMD_KEY);
+    if (cmdList != null && cmdList instanceof JsonArray array) {
+      return array.stream()
+          .map(JsonValue::toString)
+          .collect(Collectors.joining(" "))
+          .replace("\"", "");
```
eunomie: I think we should just return a `List<String>` and not try to join them. That way we are similar to what Go SDK is doing.

sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/exception/DaggerExecException.java
```diff
@@ -0,0 +1,53 @@
...
+        .collect(Collectors.joining(" "))
+        .replace("\"", "");
+  }
+
+  public String getStdErr() {
+    return Arrays.asList(getErrors()).stream()
+        .map(DaggerExceptionUtils::getStdErr)
+        .collect(Collectors.joining(" "))
+        .replace("\"", "");
+  }
```
eunomie: One extra change (this should be applied here, but also at the level of `DaggerQueryException` and it requires a little change in `QueryBuilder`): I don't think we need to handle more than one error at a time. At least that's what it looks like when I'm looking at the Go SDK code and that's also what I'm seeing on various tries.

So we can have something simpler like:

```java
  public DaggerExecException(GraphQLError error) {
    super(error);
  }

  public Integer getExitCode() {
    return DaggerExceptionUtils.getExitCode(getError());
  }

  public List<String> getPath() {
    return DaggerExceptionUtils.getPath(getError());
  }

  public List<String> getCmd() {
    return DaggerExceptionUtils.getCmd(getError());
  }

  public String getStdOut() {
    return DaggerExceptionUtils.getStdOut(getError());
  }

  public String getStdErr() {
    return DaggerExceptionUtils.getStdErr(getError());
  }
```

(with `getErrors` changed to `getError` in `DaggerQueryException`)

sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/exception/DaggerExceptionUtils.java
robertonav20: yes you are right, i I thought is really safe? the returned code could be not a integere value?

sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/exception/DaggerExceptionUtils.java
robertonav20: 👍

sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/exception/DaggerExecException.java
robertonav20: 👍

sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/exception/DaggerExceptionUtils.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  public static List<String> getPath(GraphQLError error) {
+    return Arrays.stream(error.getPath()).map(Object::toString).toList();
+  }
+
+  public static List<String> getCmd(GraphQLError error) {
+    Object cmdList = getExtensionValueByKey(error, CMD_KEY);
+    if (cmdList != null && cmdList instanceof JsonArray array) {
+      return array.stream().map(JsonValue::toString).toList();
```
eunomie: As `cmd` is a json array we have to unmarshall the elements of the array.

For instance, if the error contains, as `cmd`, the following value:
```graphql
cmd=["cat","/"]
```
what we expect if to have `getCmd` returning a list of string with the two values `cat` and `/foo`, and once printed to have `[cat, /foo]`. But currently what is returned is an array of the _json_ representation of the strings and what we are receiving is a list with two values `"cat"` and `"/foo"` and if we print them we have `["cat", "/foo"]`

Not saying this is the best solution, but it looks like to work with the following version:

(the `cmdList != null` is already covered by the `instanceof`)

```suggestion
    if (cmdList instanceof JsonArray array) {
      return array.getValuesAs(JsonString.class).stream().map(JsonString::getString).toList();
```

this change also requires this one to work:

```diff
-import jakarta.json.JsonValue;
+import jakarta.json.JsonString;
```

---

My understanding here is `JsonValue::toString` will convert the value (here a string) to a json string (so adding quotes, changing `\n` to `\\n`, etc) while we want to opposite that is to access the value, as a string.

eunomie (commented):
There is one last change required about `cmd` type, and then I'll hit the approve button :-)

sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/exception/DaggerExceptionUtils.java
robertonav20: ok understood the point, thanks for your suggestion!

robertonav20:
Hi @eunomie,
i think we made a good improvement, let me know if i should add something else
Maybe i could pick up another topic 😊

eunomie (approved):

PR #10420, opened 2025-05-16
feat: add experimental sweep option to garbage collection

jedevc:
The new option can be used to configure how much data to sweep when we perform a garbage collection pass. This requires a minor buildkit patch (we can maintain this in our fork for now): https://github.com/dagger/buildkit/tree/gc-target-space.


jedevc:
I do not like the option name `sweep`, happy to discuss what it should be :grin:

core/integration/localcache_test.go
```diff
@@ -175,11 +178,27 @@ func (EngineSuite) TestLocalCacheAutomaticGC(ctx context.Context, t *testctx.T)
...
+			blocks:       2,
+			maxUsedSpace: "3GB",
+			target:       fmt.Sprint(3 * 1024 * 1024 * 1024), // 3GB
+		},
+		{
+			// test creates 2x2gb, this means we have no free storage, so gc kicks in and clears 2x2gb
+			name:         "sweep",
+			blocks:       2,
+			maxUsedSpace: "3GB",
+			sweep:        "3GB",
```
sipsma, line 199: nit: test for a percentage value of sweep is probably worthwhile too

sipsma (approved):
LGTM overall, some nits but approving to unblock merge since I'm out next week.

As far as the name, I don't hate "sweep", but maybe "sweep size" is a little more specific/clear? Or even more verbose would be "single sweep size", since it's the size of one sweep through the gc.

PR #10425, opened 2025-05-19
Avoid conficts in generated PHP docs

jedevc:
This is a quick hack to try and avoid conflicts in these files.

Ideally these just shouldn't be committed to the repo at all, but we're a bit dependent on netlify for our docs builds right now.

(note this doesn't fix the issue in `renderer.index` - I have no idea what format that is, so I can't fix the issue).

### DIFF
--- a/docs/.dagger/main.go
+++ b/docs/.dagger/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"dagger/docs/internal/dagger"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


vikram-dagger:
Adding @jasonmccallister to review since afaik he worked on doctum

jasonmccallister (approved):

jasonmccallister:
> note this doesn't fix the issue in renderer.index - I have no idea what format that is, so I can't fix the issue

That format is the `serialize` function in the PHP stdlib. I'm going to take a similar approach you did and `unserialize` the file in the step where we are formatting JSON - just want to verify that its not a configuration option before doing the extra work.

Issue #8484, opened 2024-09-18
Directory.withSymlink


shykes:
## Problem

The only way to create a symlink in a directory is to execute `ln -s`, which is not ideal.

## Solution

Add `Directory.withSymlink` and `Container.withSymlink` to the core API.

jedevc:
#9395 should give us a neat way to implement this without needing layers of buildkit changes.

sipsma:
Fixed by https://github.com/dagger/dagger/pull/10435

PR #10435, opened 2025-05-19
feat: directory.WithSymlink

alexcb:
Adds a new WithSymlink method to the dagger API.

Note that this only implements the directory portion of the API, a follow-up commit will be required for container.WithSymlink once ContainerDagOp has been implemented.

See https://github.com/dagger/dagger/issues/8484 for details.


alexcb:
I'm a little confused about the failures, is there something auth-related that's causing the 500 errors? e.g.

```
10  : GitRepository.ref ERROR [0.5s]
10  : ! cannot resolve "https://gitlab.com/dagger-modules/private/test/more/dagger-test-modules-private.git": git error: exit status 128
stderr:
error: RPC failed; HTTP 500 curl 22 The requested URL returned error: 500
fatal: expected flush after ref listing
```

I'm not seeing any errors when I run `dagger call test specific --pkg="./core/integration" --run="^Test(Directory|Container)/" --test-verbose --failfast` locally.

core/container.go
```diff
@@ -940,7 +1023,7 @@ func locatePath[T *File | *Directory](
 
 	// NB(vito): iterate in reverse order so we'll find deeper mounts first
 	for i := len(container.Mounts) - 1; i >= 0; i-- {
-		mnt := container.Mounts[i]
+		mnt := &container.Mounts[i]
```
alexcb, line 1026: note that this change was needed to allow `mount.Result = newDir.Result` to be set in `container.WithSymlink`

core/schema/container.go
```diff
@@ -463,6 +463,13 @@ func (s *containerSchema) Install() {
 				),
 			),
 
+		dagql.NodeFunc("withSymlink", DagOpContainerWrapper(s.srv, s.withSymlink)).
+			Doc(`Return a snapshot with a symlink`).
+			Args(
+				dagql.Arg("target").Doc(`Location of the file or directory to link to (e.g., "/existing/file").`),
+				dagql.Arg("linkName").Doc(`Location where the symbolic link will be created (e.g., "/new-file-link").`),
+			),
```
jedevc, line 471: Probably makes sense to also take optional `expand`, `permissions` and `owner` args, similar to `withNewFile`: https://github.com/dagger/dagger/blob/54a7d54ad79bb38bb0111086510bb4844d9a0b86/core/schema/container.go#L381-L388

core/container.go
```diff
@@ -687,6 +688,58 @@ func (container *Container) WithNewFile(ctx context.Context, dest string, conten
...
+		// symlink will be added to a mounted directory
+		newDir, err := dir.WithSymlink(ctx, srv, target, strings.TrimPrefix(containerPath, mount.Target+"/"))
+		if err != nil {
+			return nil, err
+		}
+		mount.Result = newDir.Result
+		return container, nil
+	}
+
+	// otherwise symlink will be added to root fs
```
jedevc, line 716: :thinking: could we use the directory logic from above as well here? Maybe using `RootFS` + `WithRootFS` or similar?

Just a thought, but it would be *nice* if the logic here could be shared between here and `directory.go` - obviously if it's not particularly easy, not worth forcing it, but given we want to convert *every* operation to use dagop like this, then we'll see this pattern a lot :tada:

core/directory.go
```diff
@@ -816,6 +817,55 @@ func (dir *Directory) Root() (*Directory, error) {
...
+		return nil, fmt.Errorf("no dagop")
+	}
+
+	cache := dir.Query.BuildkitCache()
+	newRef, err := cache.New(ctx, immutableRef, op.Group(), bkcache.WithRecordType(bkclient.UsageRecordTypeRegular),
+		bkcache.WithDescription(fmt.Sprintf("symlink %s -> %s", linkName, target)))
+	if err != nil {
+		return nil, err
+	}
+	err = op.Mount(ctx, newRef, func(root string) error {
```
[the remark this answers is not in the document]
jedevc: Oops, just found this, answered in discord! https://discord.com/channels/707636530424053791/1374076794327535617/1376865769966665800

Quoting myself:

> best i can tell, this is related to buildkit's lazy pulling of container images where buildkit will try and avoid pulling an image until it's needed - but because it's been deferred, it means that you might need a session to pull it in, since the registry might require auth
> that's why it's ending up everywhere - and also why it's okay in local+git sources, because you don't need to pull anything.

core/container.go
sipsma: just curious if you saw the existing `writeToPath` method: https://github.com/sipsma/dagger/blob/84dc5b3ba37238d957db1025c70dca797d35a055/core/container.go#L1099-L1099

seems like exactly what we'd want here, though it does internally do a `WithRootFS` call

---

> which does show up, I'm left wondering if WithRootFS needs some extra attention to make it work with dagops?

Yeah, looks like it. Probably just needs to check if the dir provided as a rootfs has a `.Result` set and, if so, set that to the container's `FSResult`

core/schema/container.go
alexcb: Adding `expand` makes sense; however permissions seems like a no go, from `man chmod`:

```
       chmod never changes the permissions of symbolic links; the chmod system
       call cannot change their permissions.  This is not a problem since  the
       permissions  of  symbolic links are never used.  However, for each sym‐
       bolic link listed on the command line, chmod changes the permissions of
       the pointed-to file.
```

However, apparently BSD does support it:

```
       -h      If the file is a	symbolic link, change the mode of the link it-
	       self rather than	the file that the link points to.
```




`chown` on the other hand does support ownership via the `--no-dereference` flag, so there's probably an OS call we can make to support ownership.

core/container.go
alexcb: I added check for `.Result` under alexcb@cae09e8bedd07b9a8297d89af4e2db5908ab2309, and it seems to be working now :tada: 

I initially went looking into using `writeToPath` when I first started on this PR, but ran into difficulties using it -- possibly due to the missing `.Result` check? I tried using it again, but tests started to fail, so I'll need to look into this deeper.

core/container.go
```diff
@@ -1035,6 +1059,43 @@ func (container *Container) withMounted(
...
+
+	container.Services.Merge(svcs)
+
+	// set image ref to empty string
+	container.ImageRef = ""
+
+	return container, nil
+}
+
+func (container *Container) replaceMount(
```
alexcb, line 1080: the majority of this comes from `withMounted`, I wasn't sure if I should just add another flag such as `replaceMount bool`?

or alternatively have both `withMounted` and `replaceMount` call some other helper such as `modifyMounts` and we could then pass which function to use, i.e. `container.Mounts.With` or `container.Mounts.Replace`

core/container.go
alexcb: This change is no longer needed since we're going to add a `replaceMount` call.

I'm not sure if we really want it to return a pointer to the actual data, to allow modifications, especially if there's no current use-case.

core/container.go
alexcb: `writeToPath` has been changed to use a new function `ContainerMounts.Replace` which prevents the mount index ordering from being changed.

core/container.go
alexcb: Additionally, should this function return an error if 1) no mount was found? and 2) multiple identical mounts are found?

jedevc: Having a separate function SGTM :tada:

Agree it would be good to return an error if no mount was found - ideally this shouldn't happen right, since it's only called internally? But it should still error.

I don't think we should handle identical mounts here, if that's an error case, then we should error on adding it in the first place (seems out-of-scope for here).

core/integration/container_test.go
jedevc, line 1: One other test that would be useful is a check to make sure that `WithSymlink` is cached.

Probably not required for *every* future addition, but given this is the first use of `ContainerDagOp` it would be useful :tada:

Perhaps something like https://github.com/dagger/dagger/blob/54a7d54ad79bb38bb0111086510bb4844d9a0b86/core/integration/http_test.go#L198-L204

Essentially do the same operation twice (across two separate clients, ensuring the first client is closed first), and then make sure you get the same stdout each time.

(note - content based caching *doesn't* work for `DirectoryDagOp` yet - I need to fix this. so we can just test for *exactly* the same operation.)

core/container.go
```diff
@@ -314,6 +314,18 @@ func (mnts ContainerMounts) With(newMnt ContainerMount) ContainerMounts {
 	return mntsCp
 }
 
+func (mnts ContainerMounts) Replace(newMnt ContainerMount) ContainerMounts {
```
jedevc: Probably worth erroring if the `newMnt.Target` wasn't already there (or at least returning an `ok bool`).

jedevc (approved):
Left a couple nitpicky comments, but generally LGTM!

Will give you a chance to look over them, but won't block merging on them :grin:

core/integration/container_test.go
alexcb: Thanks for the `/dev/random` idea, I could swear my test failed (showing caching didn't work), however I've now run it 100+ times and couldn't replicate it. Maybe I'm imagining that first failure?

alexcb: ok, I'm not crazy, CI just had it fail on it after I pushed a completely unrelated (typo in a comment) fix.

Here's the failure:
```
Container.withExec(args: ["sh", "-c", "head -c 99 /dev/random | base64 -w0"]): Container!

1   : /.dagger-cli session --label dagger.io/sdk.name:go --label dagger.io/sdk.version:n/a DONE [0.5s]

12  : Container.withExec(args: ["sh", "-c", "head -c 99 /dev/random | base64 -w0"]): Container!
12  : [0.1s] | pwRv6zez2rbdrH0mb8mwkWUrGSOPSfpoZuiMpsVcAPB2QtVMYLXo0by+rPCy1m2FSyxTwrNYFHLsr98ULj//k0d+636yiIvcTcndMVrr4g3ZxptC7whMEelYnJjS8lJYgNz8
12  : Container.withExec DONE [0.3s]

13  : Container.stdout: String!
13  : Container.stdout DONE [0.3s]
13  : [0.3s] | pwRv6zez2rbdrH0mb8mwkWUrGSOPSfpoZuiMpsVcAPB2QtVMYLXo0by+rPCy1m2FSyxTwrNYFHLsr98ULj//k0d+636yiIvcTcndMVrr4g3ZxptC7whMEelYnJjS8lJYgNz8


        Error Trace:    /app/core/integration/container_test.go:5131
        Error:          Not equal: 
                        expected: "+aBMhWi0ha1VxY6I4h1KvaNWQMOv37nAmAxfvLy9C54gvMnywCNRSCT9KDTBwfK4SZcTJqoVFw7lD9ujqanXBIzX8wcjDop/oOGhtLegcwhc7DxNquAycF/8l0BQeeynF2Vq"
                        actual  : "pwRv6zez2rbdrH0mb8mwkWUrGSOPSfpoZuiMpsVcAPB2QtVMYLXo0by+rPCy1m2FSyxTwrNYFHLsr98ULj//k0d+636yiIvcTcndMVrr4g3ZxptC7whMEelYnJjS8lJYgNz8"
                        
                        Diff:
                        --- Expected
                        +++ Actual
                        @@ -1 +1 @@
                        -+aBMhWi0ha1VxY6I4h1KvaNWQMOv37nAmAxfvLy9C54gvMnywCNRSCT9KDTBwfK4SZcTJqoVFw7lD9ujqanXBIzX8wcjDop/oOGhtLegcwhc7DxNquAycF/8l0BQeeynF2Vq
                        +pwRv6zez2rbdrH0mb8mwkWUrGSOPSfpoZuiMpsVcAPB2QtVMYLXo0by+rPCy1m2FSyxTwrNYFHLsr98ULj//k0d+636yiIvcTcndMVrr4g3ZxptC7whMEelYnJjS8lJYgNz8
        Test:           TestContainer/TestSymlinkCaching
```

core/integration/container_test.go
```diff
@@ -4997,3 +5010,134 @@ func main() {
...
+}
+
+func (ContainerSuite) TestSymlinkCaching(ctx context.Context, t *testctx.T) {
+	c1 := connect(ctx, t)
+	out1, err := c1.Container().
+		From(alpineImage).
+		WithSymlink("bar", "foo").
+		WithExec([]string{"sh", "-c", "head -c 99 /dev/random | base64 -w0"}).
+		Stdout(ctx)
+	_ = c1.Close()
```
sipsma: no need to close any of these I think; closing also makes the cache ref prunable and thus might explain the flakes you mentioned

core/directory.go
```diff
@@ -817,6 +819,57 @@ func (dir *Directory) Root() (*Directory, error) {
...
 }
 
+func symlink(ctx context.Context, cache bkcache.Manager, parent bkcache.ImmutableRef, s session.Group, target, linkName string) (bkcache.ImmutableRef, error) {
+	newRef, err := cache.New(ctx, parent, s, bkcache.WithRecordType(bkclient.UsageRecordTypeRegular),
+		bkcache.WithDescription(fmt.Sprintf("symlink %s -> %s", linkName, target)))
+	if err != nil {
+		return nil, err
+	}
+	err = MountRef(ctx, newRef, nil, func(root string) error {
+		fullLinkName := path.Join(root, linkName)
```
sipsma: I think this might be vulnerable to a user-provided `linkName` like `../../../../foo`, in which case we could end up making the link outside `root` if there's enough `../`s?

Also could be a similar problem with escaping `filepath.Join(root, dir.Dir)` (i.e. escaping the subdir of `Directory` that has been selected).

Maybe we should just have a check of `strings.HasPrefix(fullLinkName, filepath.Clean(root)+"/")`, and error if not? `fs.RootPath` (from `"github.com/containerd/continuity/fs"`) would work too but it does way more work that isn't necessary here I think.

Either way, worth an integ test for this case. Slightly finicky since it depends on the filesystem layout of snapshots the engine stores (currently like `/var/lib/dagger/worker/snapshots/snapshots/<id>/fs/`), but maybe just include enough `../`s to handle that plus a good chunk extra in case we change that layout in the future.

alexcb: That's an interesting question, somehow I thought containers magic would protect us here.

I tried a very basic test so far:

```
	_, err := c.Container().
		From(alpineImage).
		WithSymlink("_", "../../../../../../../../../../../../../../../outside-hack").Sync(ctx)
	require.NoError(t, err)
```

Then I was able to locate where the `linkName` ended up:

```
$ docker exec -ti dagger-engine.dev sh -c 'find / | grep outside-hack'
/var/lib/dagger/worker/snapshots/snapshots/4937/fs/outside-hack
```

it looks like it's still within the snapshot fs:

```
$ docker exec -ti dagger-engine.dev sh -c 'ls -la /var/lib/dagger/worker/snapshots/snapshots/4937/fs/outside-hack'
lrwxrwxrwx    1 root     root             1 Jun  3 20:46 /var/lib/dagger/worker/snapshots/snapshots/4937/fs/outside-hack -> _
```

As for adding a test for this, should we try to do some similar call within the dagger engine container? or would it be sufficient to do something like:

```
	t.Run("symlink cant escape root fs", func(ctx context.Context, t *testctx.T) {
		ctr := c.Container().
			From(alpineImage).
			WithNewFile("some-file", "data").
			WithSymlink("some-file", "../../../../../../../../../../../../../../../this-should-be-in-the-root-fs")

		entries, err := ctr.Rootfs().Entries(ctx)
		require.NoError(t, err)
		require.Contains(t, entries, "this-should-be-in-the-root-fs")

		s, err := ctr.File("this-should-be-in-the-root-fs").Contents(ctx)
		require.NoError(t, err)
		require.Equal(t, "data", s)
	})
```

alexcb: I also tried the reverse by doing:

```
	_, err := c.Container().
		From(alpineImage).
		WithSymlink("../../../../../../../../../../../../../../..", "escape").WithNewFile("escape/here-i-am", "hackhackhack").Sync(ctx)
```

and it still seems to be in the correct location:

```
$ docker exec -ti dagger-engine.dev sh -c 'find / | grep here-i-am'
/var/lib/dagger/worker/snapshots/snapshots/4941/fs/here-i-am
```

sipsma: Actually, thought about this again and realized that we probably **do** want `fs.RootPath`, otherwise there's the other delightful case of a user making an absolute symlink in the snapshot like `/foo/bar -> /abc/123`, then specifying the `linkName` as `/foo/bar/newsymlink`, in which case `os.Symlink` will probably follow `/foo/bar` blindly and leave us at the engine's rootfs again 🫠 `fs.RootPath` should fully protect us though

Symlinks are always a delight 😅

EDIT: github didn't load your comments before I sent this one, so seems out place, catching up on what you left

sipsma: Okay yeah, this was even more complicated than I though at first because we do a `path.Join(dir.Dir, linkName)` earlier, which just barely avoided the simplest case of this being a problem, but you can still break out of the rootfs with a test like this:
```go
	t.Run("symlink cant escape root fs", func(ctx context.Context, t *testctx.T) {
		dir, err := c.Directory().
			WithNewFile("/etc/some-subdir/some-file", "data").
			Sync(ctx)
		require.NoError(t, err)
		dir, err = dir.
			WithSymlink("/etc", "/symlink-to-etc").
			Sync(ctx)
		require.NoError(t, err)
		dir, err = dir.
			WithSymlink("some-subdir/some-file", "/symlink-to-etc/symlink-to-some-file").
			Sync(ctx)
		require.NoError(t, err)

		entries, err := dir.Directory("/etc/some-subdir").Entries(ctx)
		require.NoError(t, err)
		require.Contains(t, entries, "some-file")
		require.Contains(t, entries, "symlink-to-some-file")

		s, err := dir.File("/symlink-to-etc/some-subdir/symlink-to-some-file").Contents(ctx)
		require.NoError(t, err)
		require.Equal(t, "data", s)
	})
```
(sorry for the mess, I did a bunch of Syncs to be sure I knew where it was failing exactly)

Interestingly, when running that locally I think I actually hit two separate issues:
1. At first it failed on the `os.MkdirAll` step that creates the parent dir. It seems like `os.MkdirAll` doesn't appreciate a path component being an existing symlink (even if it points to a directory)
2. Did a quick patch to workaround that locally and then managed to actually see the container breakout, after running the test:
```
sipsma@dagger_dev:~/repo/github.com/sipsma/dagger$ docker exec -it dagger-engine.dev sh -c 'ls -la /etc/symlink-to-some-file'
lrwxrwxrwx    1 root     root            21 Jun  3 22:24 /etc/symlink-to-some-file -> some-subdir/some-file
```

I think `fs.RootPath` will fix both those problems above, so probably do need it after-all 🤷‍♂️

alexcb: This has been fixed now, good catch.

sipsma (approved):
🚀

PR #10442, opened 2025-05-21
Platform modules

shykes:
This is a WIP PR for implementing "platform modules" [as discussed in this discord thread](https://discord.com/channels/707636530424053791/1374218324778422273)

NOTE: it doesn't fully work currently...

I think the problem is I hijack the context directory at the wrong time:

- I need the platform module's context directory when loading its own functions
- But then I need the *original* context when resolving default path etc.

### DIFF
--- a/core/modules/config.go
+++ b/core/modules/config.go
@@ -30,6 +30,7 @@ func ParseModuleConfig(src []byte) (*ModuleConfigWithUserFields, error) {
 		return nil, fmt.Errorf("module requires dagger %s, but you have %s", meta.EngineVersion, engine.Version)
 	}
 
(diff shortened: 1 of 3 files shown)


chrisjpalmer:
Hi Solomon. Very cool. Gave this a test on my local with the following command:

```sh
dagger call dev --target ../my-test-repo engine publish --image=experimental-dagger
```

with the following json:

```json
// ../my-test-repo/dagger.json
{
  "name": "main",
  "engineVersion": "v0.18.4",
  "platform": {
      "name": "golang-service",
      "source": "infrastructure/dagger-modules/golang-service"
  }
}
```

and it works!

Then I could run the module like so:

```sh
dagger functions
dagger call my-function
```

PR #10443, opened 2025-05-21
chore: cleanup core/util.go

jedevc:
We can use a lot of the functions from the new slices/maps packages, and a lot of these patterns have other ways of doing them in more modern go.

Some of this is just dead code as well.


TomChv (approved):
LGTM, to merge after my comment is addressed

PR #10444, opened 2025-05-21
ci: update docusaurus module

jedevc:
This includes fixes for various hacks, and drastically improves the caching - now it actually works, so it won't trigger a whole build from scratch every single time.

See:
- https://github.com/levlaz/daggerverse/pull/14 https://github.com/levlaz/daggerverse/pull/15
- https://github.com/levlaz/daggerverse/pull/13

### DIFF
--- a/docs/.dagger/main.go
+++ b/docs/.dagger/main.go
@@ -63,11 +63,8 @@ pagination_prev: null
 // Build the docs website
 func (d Docs) Site() *dagger.Directory {
 	opts := dagger.DocusaurusOpts{
-		Dir:  "/src/docs",
+		Dir:  "./docs",
 		Yarn: true,
-		// HACK: cache seems to cause weird ephemeral errors occasionally -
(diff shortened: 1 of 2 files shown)


marcosnils:
the main reason to disable the cache originally was that it was introducing some strange cache pollution inconsistencies when the docusaurus module across different dagger docs versions. Seems like this change might introduce that same behavior again? Not sure if the work we did with cache volume namespacing might probably improve this?

jedevc:
The reason the caching got incorrect was because we were caching `node_modules` and the `build` directory in shared caching mode. This broke parallel builds pretty easily.

Now, we cache `node_modules/.cache` only, but in private mode, so this should avoid breaking parallel builds.

marcosnils:
> Now, we cache `node_modules/.cache` only, but in private mode, so this should avoid breaking parallel builds.

 SGTM missed the changes in the docusaurus module :pray:

PR #10445, opened 2025-05-21
feat: add container dagop

jedevc:
This PR extends the dagop concept introduced in https://github.com/dagger/dagger/pull/9395 to a `ContainerDagOp` which allows using dagops on containers, mapping all mounts correctly :tada:

See example usage in https://github.com/jedevc/dagger/compare/container-dagop...jedevc:dagger:container-dagop-magic?expand=1.

Intended to be used in:
- https://github.com/dagger/dagger/pull/10435
- https://github.com/dagger/dagger/pull/10457

### DIFF
--- a/core/dagop.go
+++ b/core/dagop.go
@@ -14,9 +14,11 @@ import (
 	bkcache "github.com/moby/buildkit/cache"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/client/llb"
+	bkgw "github.com/moby/buildkit/frontend/gateway/client"
 	bksession "github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/snapshot"
 	"github.com/moby/buildkit/solver"
+	"github.com/moby/buildkit/solver/pb"
 	"github.com/moby/buildkit/worker"
 	"github.com/opencontainers/go-digest"
 )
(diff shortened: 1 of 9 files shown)


core/dagop.go
```diff
@@ -321,44 +339,384 @@ func (op RawDagOp) Exec(ctx context.Context, g bksession.Group, inputs []solver.
...
-	if val := ctx.Value(dagOpContextKey(t.Name())); val != nil {
-		t, ok = val.(T)
+type ContainerDagOp struct {
+	ID *call.ID
+
+	Mounts []*pb.Mount
+
+	// Data is any additional data that should be passed to the dagop. It does
+	// not contribute to the cache key.
+	Data any
```
sipsma, line 407: what's the use case for this? didn't see it actually used in other dependent PRs, but might have missed it while sifting through

core/dagop.go
```diff
@@ -321,44 +339,384 @@ func (op RawDagOp) Exec(ctx context.Context, g bksession.Group, inputs []solver.
...
+	// handle secret mounts
+	for _, secret := range container.Secrets {
+		if secret.MountPath == "" {
+			continue
+		}
+		uid, gid := 0, 0
+		if secret.Owner != nil {
+			uid, gid = secret.Owner.UID, secret.Owner.GID
+		}
+		mount := &pb.Mount{
```
sipsma, line 583: should this have `Output: pb.SkipOutput` for good measure if nothing else?

same comment for sockets below

core/dagop.go
```diff
@@ -321,44 +339,384 @@ func (op RawDagOp) Exec(ctx context.Context, g bksession.Group, inputs []solver.
...
+	}
+
+	return mounts, states, nil
+}
+
+// setAllContainerMounts is the reverse of getAllContainerMounts, and rewrites
+// the container mounts based on the mounts and states specified.
+func setAllContainerMounts(ctx context.Context, container *Container, mounts []*pb.Mount, states []llb.State) error {
+	for _, mount := range mounts {
+		if mount.Output == -1 {
```
sipsma, line 627: nit: `pb.SkipOutput` rather than `-1` for clarity

core/dagop.go
```diff
@@ -321,44 +339,384 @@ func (op RawDagOp) Exec(ctx context.Context, g bksession.Group, inputs []solver.
...
+
+		st := states[mount.Output]
+
+		def, err := st.Marshal(ctx, llb.Platform(container.Platform.Spec()))
+		if err != nil {
+			return err
+		}
+
+		switch mount.Output {
+		case -1:
```
sipsma, line 639: would this result in us skipping re-adding all the mounts like cache volume, tmpfs, secret, etc. ? or am I missing something where that's handled?

core/dagop.go
jedevc: Mmmm, this didn't *seem* to be breaking anything, but yes, worth fixing anyways :scream:

sipsma (approved):

jedevc:
The failing test is in gnome secret provider - which is failing on `main`, so merging this now, will investigate that failure separately.

PR #10449, opened 2025-05-21
feat: directory.exists(path, type)

### LINKED ISSUE #6713: Check if a file or directory exists
## Problem

The Dagger API does not support checking if a file exists. There is an SDK-specific workaround (maybe? not documented) which is to 1) call any function that requires the file to exist, like `Sync` or `Contents`, 2) check for error, and 3) check for the error message returned when the file doesn't exist. This possibly can be wrapped in a convenience type by the SDK (again hard to tell as it is not documented).

This makes for a bad DX.

## Solution

Several solutions have been proposed:

### Option 1: `File.exists()`, `Directory.exists()`

Add a new core function to the `File` and `Directory` types:

```graphql
type Directory {
 "Returns true if the directory exists"
 exists: Bool!
}
 
type File {
 "Returns true if the file exists"
 exists: Bool!
}
```

### Option 2: `Directory.exists(path)`

In this option, you query a parent directory for the existence of a given path:

```graphql
type Directory {
 "Returns true if a file or directory exists at the given path"
 exists(path: String!): Bool!
}
```

alexcb:
Adds a new method to directories to allow checking if a file (or directory) exists at a specific path, fixes #6713


core/schema/directory.go
```diff
@@ -96,6 +96,12 @@ func (s *directorySchema) Install() {
 			Args(
 				dagql.Arg("paths").Doc(`Paths of the files to remove (e.g., ["/file.txt"]).`),
 			),
+		dagql.NodeFunc("exists", s.exists).
```
alexcb, line 99: todo: look into changing this to:
```suggestion
		dagql.Node("exists", s.exists).
```

core/integration/directory_test.go
```diff
@@ -1453,3 +1453,69 @@ func (DirectorySuite) TestDirectoryName(ctx context.Context, t *testctx.T) {
 		})
 	})
 }
+
+func (DirectorySuite) TestExists(ctx context.Context, t *testctx.T) {
```
alexcb, line 1457: todo: add tests for `"symlink"` type

jedevc:
FYI @helderco @shykes

docs/docs-graphql/schema.graphqls
```diff
@@ -2174,6 +2196,18 @@ The `ErrorValueID` scalar type represents an identifier for an object of type Er
...
+"""File type."""
+enum ExistsType {
+  """Tests path is a regular file"""
+  REGULAR_TYPE
+
+  """Tests path is a directory"""
+  DIRECTORY_TYPE
+
+  """Tests path is a directory"""
+  SYMLINK_TYPE
```
sipsma: doc string is off here

core/directory.go
```diff
@@ -1088,3 +1146,67 @@ func validateFileName(file string) error {
...
+
+// CacheSharingMode marshals to its lowercased value.
+//
+// NB: as far as I can recall this is purely for ~*aesthetic*~. GraphQL consts
+// are so shouty!
+func (et ExistsType) MarshalJSON() ([]byte, error) {
+	return json.Marshal(strings.ToLower(string(et)))
+}
+
+// CacheSharingMode marshals to its lowercased value.
```
sipsma: nit: `CacheSharingMode` is copy-pasta'd

core/directory.go
```diff
@@ -935,6 +937,62 @@ func (dir *Directory) Without(ctx context.Context, srv *dagql.Server, paths ...s
...
+	immutableRef, err := ref.CacheRef(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	osStatFunc := os.Stat
+	if targetType == ExistsTypeSymlink {
+		// symlink testing requires the Lstat call, which does NOT follow symlinks
+		osStatFunc = os.Lstat
+	}
```
sipsma: So in the case where you have a symlink like `/link -> /some/file`, we will by default follow the symlink and test the target (only exception being if you actually set the expected type to Symlink)?

Very grey area, but IMO it would be much simpler cognitively to just always use Lstat; so always check the provided path without following symlinks. Then if desired we could add an extra arg that says "follow any symlinks". I feel like that's just easier to reason about.

Otherwise it just gets confusing. e.g. in the `/link -> /some/file` case, what if I wanted to assert that `/link` itself was a file? It would be possible by making a few calls but requires a lot of thinking since whether the symlink is followed or not depends on the expected type arg.

core/directory.go
alexcb: good catch.

Should I remove the marshalling to lowercased code all together? I don't know if there's a particular reason for it for this case.

core/directory.go
alexcb: > So in the case where you have a symlink like /link -> /some/file, we will by default follow the symlink and test the target (only exception being if you actually set the expected type to Symlink)?

That's correct. 

> Very grey area, but IMO it would be much simpler cognitively to just always use Lstat; so always check the provided path without following symlinks. Then if desired we could add an extra arg that says "follow any symlinks". I feel like that's just easier to reason about.

The current code tries to mimic how the `test` command works -- it follows the symlink in the majority of the cases It's interesting to note that `man test` describes the `-f` option as:

```
       -f FILE
              FILE exists and is a regular file
```

and even though the man pages uses the wording "regular file", it will automatically follow a symlink. My best guess is because the `open` syscall also does the same thing.

I agree it's a rather grey area. Ultimately it comes down to a product question -- i don't have very strong feelings about it.

> what if I wanted to assert that /link itself was a file?

I agree it's a bit confusing -- This is the downfall with trying to mimic the `test` shell command, I think it has to be done as two commands, which isn't great. This seems to be an issue in a few languages:

```
# using test command directly
test -f path && ! test -L path

# using bash
if [[ -f "$file" && ! -L "$file" ]]; then ...

# python
if os.path.isfile(file) and not os.path.islink(file):
   ...
```

Another idea, is to include a new type such as `ExistsTypeActualFile`, which would combine the above tests? It's slightly cheeky since the term regular is rather unclear :)

I wonder what the use-case would be for asserting it's an actual file vs a symlink to an existing file?

core/directory.go
alexcb: I ended up removing the lowercase code, since it's not a pattern all data types do.

core/directory.go
sipsma: I see, personally I'd rather go with something that feels the most simple to us rather than just do the same as `test` for the sake of consistency with it, but I get the argument the other way.

My personal order of preference would be:
1. (best) Never follow symlinks by default. Add an optional arg like `followSymlinks` that if set true enables following them.
2. Always follow symlinks by default. Add an optional arg like `doNotFollowSymlinks` that if set true disables following them.
3. (worst) the current behavior where they will always be followed unless the expected type arg is itself set to expect a symlink

But I could live with any of the above

cc @jedevc @shykes for other opinions

> I wonder what the use-case would be for asserting it's an actual file vs a symlink to an existing file?

Feels like as valid a use-case as any others for these new APIs. Maybe someone changed a filesystem layout where something used to be a symlink but is now a file and they want to test that the change worked, etc.

shykes: I don't have an opinion on this one, don't block on me 🙏

alexcb: I would be in favour of the second option:

> Always follow symlinks by default. Add an optional arg like doNotFollowSymlinks that if set true disables following them.

since this would maintain parity with test, bash, python, etc.

I can start hacking at this unless there's any objections.

Issue #8419, opened 2024-09-11
Global configuration for dependency modules


jedevc:
Let's pretend I've created a generic reusable module, `golang` - it can build, test, lint, generate docs, etc, everything I want a module to do.

I want to install this module as a dependency to my project. Because `golang` is a good module, it even supports setting the `version` in the constructor.

But there's an issue with this - if I want to call `golang` at multiple points in my module, and I *need* to pin the version everywhere, today I have a few options:
- Specify the `version` everywhere I construct `golang`. This is kinda sad, and ends up with lots of duplicates.
- Call the constructor *once*, and then store the result. This is also kinda sad, since now I need to pass the object *everywhere* in my own module (which gets worse the more such objects from other modules I might have)
- Write a wrapper function that wraps the constructor, and just call that everywhere (this feels like the best option with what we have right now)

However, some future options include:
- Have dependency "opts" in `dagger.json` - then automagically feed these defaults to the constructor (similar to https://github.com/dagger/dagger/issues/6323)

Not quite sure which is best, though I'm leaning towards adding more functionality to `dagger.json` for this. For one, it feels annoying to have to write my own wrappers everywhere, and makes it hard to update these values. Syntax like `dagger install github.com/.../golang --version=v1.23` would be pretty neat.

Whatever option we pick, we need to have this based on the *caller* - not the dependency module. I might/might not have control over the default in the dependency `golang` module, and so might not be able to easily bump the version.

cc @shykes (inspired by our conversation in https://discord.com/channels/707636530424053791/1003718839739105300/1282281815347826742)

Issue #6323, opened 2023-12-23
A config file for my modules


shykes:
Now that modules have constructors, we have a way to reduce the verbosity of calling them from the command-line: write module configuration to a config file.

shykes:
This is relevant to #6112

shykes:
Note, at least one user has developed a wrapper tool to do this.

jedevc:
Ooh this seems similar to https://github.com/dagger/dagger/issues/8419, but I remember when opening it I couldn't find this issue 😂

marcosnils:
Shall we close #8419 and consolidate here @jedevc ?

shykes:
A real-world example of a user implementing a stopgap: https://discord.com/channels/707636530424053791/1324306231875338300

PR #10453, opened 2025-05-22
Inline dependencies (formerly known as platform modules)

shykes:
*This is a second take on the "platform module" idea, first prototyped in #10442.*

Implement the inline dependency feature described in #10464 

## Usage

```console
$ dagger install --inline github.com/shykes/x/go
$ cat dagger.json
{
  ...
  "dependencies": [
    {
      "name": "go",
      "source": "github.com/shykes/x/go",
      "pin": "c71f1e5edcda371e19f754f0eab1faf31ba2d770",
      "inline": true
    }
  ]
}
```

```console
dagger -c 'go | env | terminal'
```

Note the new `inline` field.

Any dependency can be marked as inline, and will co-exist with other dependencies, and a SDK, normally.

A module can have multiple inline dependencies. Namespacing is unchanged from regular dependencies: each dependency is namespaced under its own name; `dagger call` cannot call dependencies, but dagger shell can. This could be changed, but I don't handle that in the PR yet.

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -43,7 +43,8 @@ var (
 	moduleSourcePath string
 	moduleIncludes   []string
 
-	installName string
+	installName     string
+	installPlatform bool
 
 	developSDK        string
 	developSourcePath string
@@ -135,6 +136,8 @@ func init() {
 	modulePublishCmd.Flags().StringVarP(&moduleURL, "mod", "m", "", "Module reference to publish, remote git repo (defaults to current directory)")
 
 	moduleInstallCmd.Flags().StringVarP(&installName, "name", "n", "", "Name to use for the dependency in the module. Defaults to the name of the module being installed.")
+	moduleInstallCmd.Flags().BoolVarP(&installPlatform, "platform", "p", false, "Install a platform dependency")
+
 	moduleInstallCmd.Flags().StringVar(&compatVersion, "compat", modules.EngineVersionLatest, "Engine API version to target")
 	moduleAddFlags(moduleInstallCmd, moduleInstallCmd.Flags(), false)
 
@@ -330,7 +333,9 @@ var moduleInstallCmd = &cobra.Command{
 				depSrc = depSrc.WithName(installName)
 			}
 
-			modSrc = modSrc.WithDependencies([]*dagger.ModuleSource{depSrc})
+			modSrc = modSrc.WithDependencies([]*dagger.ModuleSource{depSrc}, dagger.ModuleSourceWithDependenciesOpts{
+				Platform: installPlatform,
+			})
 			if engineVersion := getCompatVersion(); engineVersion != "" {
 				modSrc = modSrc.WithEngineVersion(engineVersion)
 			}
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -43,8 +43,8 @@ var (
 	moduleSourcePath string
 	moduleIncludes   []string
 
-	installName     string
-	installPlatform bool
+	installName   string
+	installInline bool
 
 	developSDK        string
 	developSourcePath string
@@ -136,7 +136,7 @@ func init() {
 	modulePublishCmd.Flags().StringVarP(&moduleURL, "mod", "m", "", "Module reference to publish, remote git repo (defaults to current directory)")
 
 	moduleInstallCmd.Flags().StringVarP(&installName, "name", "n", "", "Name to use for the dependency in the module. Defaults to the name of the module being installed.")
-	moduleInstallCmd.Flags().BoolVarP(&installPlatform, "platform", "p", false, "Install a platform dependency")
+	moduleInstallCmd.Flags().BoolVarP(&installInline, "inline", "", false, "Install inline (execute in the context of the caller instead of its own)")
 
 	moduleInstallCmd.Flags().StringVar(&compatVersion, "compat", modules.EngineVersionLatest, "Engine API version to target")
 	moduleAddFlags(moduleInstallCmd, moduleInstallCmd.Flags(), false)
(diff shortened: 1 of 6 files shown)


jedevc:
Chucking down a list of thoughts, and ideas :tada:

---

What's the imagined flow for setting up a `app` repo that uses a `platform` module? From the CLI setup?

This needs to be pretty smooth IMO, if we want to remove wrappers. Maybe we just need an optional `dagger install` `--platform` flag (or equivalent, dependent on naming), which would just set `platform: true` for that dependency.

---

As implemented I think this will still work when we have a `source` specified - it's just that that dependency will now use our root, right? That's kinda neat - your `dag.Go` calls won't need to be passed your source, they'll just use yours by default. 

This makes me think of #6323 and #8419 - where we're overriding default args for a dependency. I wonder if instead of the concept of a "platform", this is more like the concept of a set of generic overrides :thinking: 

e.g. for `github.com/shykes/x/go`, what if you could have:

```json
{
  "name": "demo",
  "engineVersion": "v0.18.9-010101000000-dev-6839e70e9de3",
  "dependencies": [
    {
      "name": "go",
      "source": "github.com/shykes/x/go",
      "pin": "b0966cb5dc12504fbf9060e26742e7c0e5f843c0",
      "overrides": {
        "source": ".",
        "version": "1.23.8"
      }
    }
  ]
}
```

So you wouldn't just be able to configure an overridden `source`, you'd be working with an overridden `version` (and any other constructor args).

Or maybe these are entirely separate concepts - I do think the need for overrides becomes super important though with this change, since consumers of a platform may need to configure various parameters of the platform (like version, or base image, etc). So, even if have separate `platform` and `overrides`, I think it's worth considering how these would behave together.

---

I think being able to call using `dagger call` is pretty important, since otherwise anyone wanting to use this needs to switch entirely to `dagger shell`. I'm biased, since I generally still use a lot of `dagger call`.

We could allow `-m` to use named dependencies in `dagger.json`. Then you can easily call those dependencies with `dagger call`. That feels generally pretty useful anyways, as a separate change, detached from this one.

---

I think we also need a way to make sure that platforms have a way to "re-root" themselves. For example, a `docusaurus` module might be written to expect the docs starting at the `.` root of the repo, but, in our context, they might be at `./docs`. So I think it would be nice to have a way to map `./docs` as the root.

I think this is super critical in the context of monorepos, where you might have a *lot* of different bits in weird locations.

shykes:
Note: there is ongoing bikeshedding over the word "platform". It seems that "inline dependency" might be a less confusing word. TBD.

> What's the imagined flow for setting up a `app` repo that uses a `platform` module? From the CLI setup?
> 
> This needs to be pretty smooth IMO, if we want to remove wrappers. Maybe we just need an optional `dagger install` `--platform` flag (or equivalent, dependent on naming), which would just set `platform: true` for that dependency.

Yes exactly, I have an unpushed commit implementing this, but it doesn't work yet :(

> As implemented I think this will still work when we have a `source` specified - it's just that that dependency will now use our root, right? That's kinda neat - your `dag.Go` calls won't need to be passed your source, they'll just use yours by default.

Yes exactly, implemented in this way (as opposed to #10442), a "platform"/inline dependency can co-exist with other dependencies and a regular source with SDK, and everything should work.

> This makes me think of #6323 and #8419 - where we're overriding default args for a dependency. I wonder if instead of the concept of a "platform", this is more like the concept of a set of generic overrides 🤔
> 
> e.g. for `github.com/shykes/x/go`, what if you could have:
> 
> ```json
> {
>   "name": "demo",
>   "engineVersion": "v0.18.9-010101000000-dev-6839e70e9de3",
>   "dependencies": [
>     {
>       "name": "go",
>       "source": "github.com/shykes/x/go",
>       "pin": "b0966cb5dc12504fbf9060e26742e7c0e5f843c0",
>       "overrides": {
>         "source": ".",
>         "version": "1.23.8"
>       }
>     }
>   ]
> }
> ```
> 
> So you wouldn't just be able to configure an overridden `source`, you'd be working with an overridden `version` (and any other constructor args).
> 
> Or maybe these are entirely separate concepts - I do think the need for overrides becomes super important though with this change, since consumers of a platform may need to configure various parameters of the platform (like version, or base image, etc). So, even if have separate `platform` and `overrides`, I think it's worth considering how these would behave together.

Yes I think overrides are 1) super important, and 2) orthogonal.

With a platform/inline dependency, I can access the caller's context any way I want, for example I can push a new version of my module which accesses more context directories, without changing the caller's code.

> I think being able to call using `dagger call` is pretty important, since otherwise anyone wanting to use this needs to switch entirely to `dagger shell`. I'm biased, since I generally still use a lot of `dagger call`.

Yes I agree.

> We could allow `-m` to use named dependencies in `dagger.json`. Then you can easily call those dependencies with `dagger call`. That feels generally pretty useful anyways, as a separate change, detached from this one.

This actually already works :)

But, it only works when calling from within the parent module. It doesn't solve the fact that `dagger -m HOST_MODULE call go build` won't work.

> I think we also need a way to make sure that platforms have a way to "re-root" themselves. For example, a `docusaurus` module might be written to expect the docs starting at the `.` root of the repo, but, in our context, they might be at `./docs`. So I think it would be nice to have a way to map `./docs` as the root.
> 
> I think this is super critical in the context of monorepos, where you might have a _lot_ of different bits in weird locations.

Isn't that handled by the module DX? Ie. you can 1) choose the location of your dagger.json, and 2) choose your `defaultPath`. That would apply equally to inline/platform or regular modules.

shykes:
Renamed "platform module" to "inline dependency" per bikeshedding discussion.

Also opened an issue: #10464

nipuna-perera:
Please stay alive!

shykes:
@nipuna-perera got distracted by keynote and conference madness 😅 Let's get this thing moving

shykes:
@jedevc you nerd-sniped me with this "configurable context" idea

PR #10454, opened 2025-05-22
refactor: client install CLI DX

TomChv:
@shykes [mentioned](https://discord.com/channels/707636530424053791/1120503349599543376/1367677268033409188) that the current `dagger client install` CLI doesn't completely follows the spec described in #9582.

This PR aims to fix that difference:
Before: `dagger client install --generator=<generator> <output-path>` 
Now: `dagger client install <generator> <output-path>`

### DIFF
--- a/cmd/dagger/client.go
+++ b/cmd/dagger/client.go
@@ -45,25 +45,27 @@ var clientCmd = &cobra.Command{
 }
 
 var clientInstallCmd = &cobra.Command{
-	Use:     "install [options] [path]",
+	Use:     "install [options] generator [path]",
 	Aliases: []string{"use"},
 	Short:   "Generate a new Dagger client from the Dagger module",
-	Example: "dagger client install --generator=go ./dagger",
+	Example: "dagger client install go ./dagger",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		return withEngine(cmd.Context(), client.Params{}, func(ctx context.Context, engineClient *client.Client) error {
-			if generator == "" {
-				return fmt.Errorf("generator must set (ts, go, python or custom generator)")
-			}
-
 			// default the output to the current working directory if it doesn't exist yet
 			cwd, err := pathutil.Getwd()
 			if err != nil {
 				return fmt.Errorf("failed to get current working directory: %w", err)
 			}
 
-			outputPath := filepath.Join(cwd, "dagger")
-			if len(args) > 0 {
-				outputPath = args[0]
+			switch len(args) {
+			case 0:
+				return fmt.Errorf("generator must set (ts, go, python or custom generator)")
+			case 1:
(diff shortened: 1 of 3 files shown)


jedevc:
:eyes: I actually prefer the current DX - it stylistically matches the `--sdk` flag to `init`/`develop`.

That said, if this is the route we go down, the implementation LGTM.

cc @shykes

TomChv:
> 👀 I actually prefer the current DX - it stylistically matches the `--sdk` flag to `init`/`develop`.
> 
> That said, if this is the route we go down, the implementation LGTM.
> 
> cc @shykes

Yeah agree with you, waiting for @shykes approval before merging

shykes:
The reason for the syntax I proposed, is to *make the path optional*. Most users (me included) don't know intuitively what path to provide. So each generator should pick its own default. I suspect this is not currently supported in the backend. So it's not just a matter of UI bikeshedding: the spec encapsulated a desired backend behavior.

shykes:
```
dagger client install GENERATOR [PATH] [-n|--no-generate]
```

So I want to be able to run:

```
dagger client install go
```

And have the go generator figure out where to install it (by default)

shykes:
(btw this kind of detail is why I insisted that we discuss the spec in the issue, before rushing an implementation)

TomChv:
> The reason for the syntax I proposed, is to make the path optional. 

The path is already optional, the client will be generated in `dagger` by default if nothing is set, if you have any suggestion for a specific default path on each SDK, I'm happy to implement that too.

shykes:
we discussed this live with Tom fyi

jedevc:
Sure, but having `--generator=go` isn't mutually exclusive with having `PATH` be optional. We could have `dagger client install --generator=go`? I agree that users may not know what path to put in, but I don't see why that's the same as having a generator flag (even if that flag is required).

Just thinking about ways to reduce friction for users who are already using `init`/`develop` today.

TomChv:
> Sure, but having --generator=go isn't mutually exclusive with having PATH be optional. We could have dagger client install --generator=go? I agree that users may not know what path to put in, but I don't see why that's the same discussion as having a generator flag (even if that flag is required).

> Just thinking about ways to reduce friction for users who are already using init/develop today.

Yeah both solutions works, if `--generator` looks important, we can close that PR, maybe we can ask some early tester of the client generator?

jedevc:
Don't want to block on having `--generator` as a flag, implementation looks good to me if @shykes agrees.

shykes:
Yeah you're right that the 2 topics are orthogonal @jedevc .

`dagger client install go` is simpler for users than `dagger client install --generator=go`. It also follows unix conventions (required arguments should not be flags, also known as "options"). And it's more consistent with `dagger install <thing to install>`

Issue #10367, opened 2025-05-09
Lift all buildkit operations into dagger


jedevc:
Using DagOp (from https://github.com/dagger/dagger/pull/9395), we can lift buildkit operations into dagger-native code, while *preserving* most of the caching semantics that we rely on today. This is part of our efforts to rely more on our dagql caching (see efforts like #9682) - once dagql caching becomes more sophisticated we'd like to rely on it more completely, but we can't do that if all of the logic is defined in buildkit. The aim is to migrate as much of our code into `github.com/dagger/dagger/core` as possible, so that when we come to switch over, we're not dependent on lots of buildkit behavior.

I've already done a couple of conversions, and hopefully now, there's some prior art to be inspired by! :tada: But to summarize:
- Today a lot of operations work by creating buildkit `llb` inside of dagger operations (dagql -> llb -> buildkit ops -> snapshots). This is complicated - we want to directly do these operations in dagger (`dagql` -> `snapshots`).
- DagOp works by taking a dagql schema handler, and calling *itself* - but from inside a buildkit operation, which allows it to cache neatly 😁 (and get written to disk, and be lazified, and all that good stuff). Because of this, there's no underlying llb (or, there is, but it's magically hidden away).
- Some conversions are very simple: you just use `DagOpFileWrapper`/`DagOpDirectoryWrapper` for handlers that return `File`s/`Directory`s respectively, the implementation of the function doesn't change. You can also easily add caching for any handler that returns an arbitrary dagql type using `DagOpWrapper`.
- However, in a lot of cases you need to more carefully split what behavior gets done inside/outside the dagop, so you can use `DagOpFile`/`DagOpDirectory`/`DagOp` directly.
- When inside a DagOp and returning a `File`/`Directory` you can directly return `ImmutableRef`s by attaching them as `Result`s - this lets you manually create snapshots, and return them. Be careful, and immediately return these - these don't have corresponding LLB, so they don't chain super well, they need to be immediately returned from the handler.

The following components need to be migrated:
- [ ] `SourceOp`
  - [x] `GitSource` #9980
  - [x] `HTTPSource` #10317
  - [ ] `LocalSource`
  - [ ] `ImageSource`/`OCISource`
  - [ ] `BlobSource` (this is custom)
- [ ] `FileOp`
  - [ ] `Mkdir` (becomes `withDirectory`)
  - [ ] `Mkfile` (becomes `withNewFile`)
[...]

PR #10457, opened 2025-05-22
feat: rewrite `Container.withExec` to use DagOp

jedevc:
The continuation of https://github.com/dagger/dagger/issues/10367, converts `withExec` to completely avoid generating LLB, and instead lifting *all* the relevant functionality into dagger.

TODO:
- [x] Something about the `RawDigest` functionality I've added so we can get a form of content-caching for operations doesn't make me super happy.

### DIFF
--- a/cmd/engine/.dagger/main.go
+++ b/cmd/engine/.dagger/main.go
@@ -121,6 +121,7 @@ func (e *DaggerEngine) Container(
 	ctr = ctr.
 		WithFile(cliPath, cli).
 		WithEnvVariable("_EXPERIMENTAL_DAGGER_RUNNER_HOST", distconsts.DefaultEngineSockAddr)
+	// ctr = ctr.WithEnvVariable("BUILDKIT_SCHEDULER_DEBUG", "1")
 
 	return ctr, nil
 }
--- a/core/container.go
+++ b/core/container.go
@@ -170,6 +170,24 @@ func (container *Container) Clone() *Container {
 	return &cp
 }
 
+func (container *Container) WithoutInputs() *Container {
+	container = container.Clone()
+
+	container.FS = nil
+	container.FSResult = nil
+
+	container.Meta = nil
+	container.MetaResult = nil
+
+	for i, mount := range container.Mounts {
+		mount.Source = nil
+		mount.Result = nil
+		container.Mounts[i] = mount
+	}
+
+	return container
+}
+
 var _ dagql.OnReleaser = (*Container)(nil)
 
 func (container *Container) OnRelease(ctx context.Context) error {
@@ -1767,21 +1785,17 @@ func (container *Container) AsService(ctx context.Context, args ContainerAsServi
 		}
 	}
 
-	container, err := container.WithExec(ctx, ContainerExecOpts{
-		Args:                          cmdargs,
-		UseEntrypoint:                 useEntrypoint,
-		ExperimentalPrivilegedNesting: args.ExperimentalPrivilegedNesting,
-		InsecureRootCapabilities:      args.InsecureRootCapabilities,
-		Expand:                        args.Expand,
-		NoInit:                        args.NoInit,
-	})
-	if err != nil {
-		return nil, err
+	if len(container.Config.Entrypoint) > 0 && useEntrypoint {
+		cmdargs = append(container.Config.Entrypoint, cmdargs...)
 	}
 
 	return &Service{
-		Creator:   trace.SpanContextFromContext(ctx),
-		Container: container,
+		Creator:                       trace.SpanContextFromContext(ctx),
+		Container:                     container,
+		Args:                          cmdargs,
+		ExperimentalPrivilegedNesting: args.ExperimentalPrivilegedNesting,
+		InsecureRootCapabilities:      args.InsecureRootCapabilities,
+		NoInit:                        args.NoInit,
 	}, nil
 }
 
--- a/core/container.go
+++ b/core/container.go
@@ -1747,22 +1747,6 @@ type ContainerAsServiceArgs struct {
 	NoInit bool `default:"false"`
 }
 
-func (container *Container) AsServiceLegacy(ctx context.Context) (*Service, error) {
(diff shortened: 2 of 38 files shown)


jedevc:
Hm, getting closer to stuff working, `TestContainer` is now consistently passing.

But also got this flake:

```
                        input: container.from.withExec.withMountedCache.withDefaultArgs.withExposedPort.asService.start solve mount /: could not find dagql server: client "va0fn9719fkjx5vnb5zcrjpb4" not found
                        solve mount /.dagger_meta_mount: could not find dagql server: client "va0fn9719fkjx5vnb5zcrjpb4" not found
```

I suspect this was always here, we just weren't edge-merging dagops before :stuck_out_tongue: Gotta think about how to resolve this - the timing required to trigger is fairly specific:
- 2 `WithExec`s come in with the same cache-key, so creates two dagops that edge merge
- The first client disconnects, and discards the dagql server.
- The second one starts trying to execute, but has a discarded dagql server :cry:

I hacked around this, but I think it needs a real fix.

jedevc:
Still got some little bits and pieces to do, but should be ready for a proper review :pray:

core/container.go
```diff
@@ -288,6 +379,21 @@ func (mnt ContainerMount) SourceState() (llb.State, error) {
 	return defToState(mnt.Source)
 }
 
+var _ HasRawDigest = (*ContainerMount)(nil)
+
+func (mnt ContainerMount) RawDigest() digest.Digest {
+	return dagql.HashFrom(
+		"containerMount",
+		mnt.SourcePath,
+		mnt.Target,
```
vito: nit: this seems like a sneaky source of collisions - say in one call I mount `/foo/bar` to `/baz`, and in another I mount `/foo` to `/bar/baz`. Those will both concatenate the same. A bit contrived, but still gives me the heebie-jeebies.

Maybe we could change `HashFrom` to write a null byte after every value as a delineator?

jedevc: Yeah definitely! I thought about doing the null byte thing, but I guess you could still pass null bytes into it?

Maybe we could escape null bytes before injecting them? Don't really know, will have a think!

cmd/engine/.dagger/main.go
```diff
@@ -121,6 +121,7 @@ func (e *DaggerEngine) Container(
 	ctr = ctr.
 		WithFile(cliPath, cli).
 		WithEnvVariable("_EXPERIMENTAL_DAGGER_RUNNER_HOST", distconsts.DefaultEngineSockAddr)
+	// ctr = ctr.WithEnvVariable("BUILDKIT_SCHEDULER_DEBUG", "1")
```
sipsma, line 124: Whatever the scheduler did to you that required enabling this, I'm sorry you had to experience it

core/container_exec.go
```diff
@@ -136,224 +149,262 @@ func (container *Container) WithExec(ctx context.Context, opts ContainerExecOpts
...
+				Origin:    op.opt.CauseCtx,
+				Mounts:    op.Mounts,
+				ExecMD:    execMD,
+				Meta:      metaSpec,
+				Secretenv: secretEnvs,
+			}
+		} else {
+			// Only release actives if err is nil.
+			for i := len(p.Actives) - 1; i >= 0; i-- { // call in LIFO order
+				p.Actives[i].Ref.Release(context.TODO())
```
sipsma: nit: probably can give `context.WithoutCancel(ctx)` here? I guess there could be things in the context we care about (either now or the future). Same for the other `context.TODO()`s elsewhere

core/container_emulator.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/docker/docker/pkg/idtools"
+	"github.com/moby/buildkit/snapshot"
+	"github.com/moby/buildkit/util/archutil"
+	"github.com/moby/buildkit/util/bklog"
+	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	copy "github.com/tonistiigi/fsutil/copy"
+)
+
+const qemuMountName = "/dev/.buildkit_qemu_emulator"
```
sipsma: nit: we have a const in `executor_spec` for this path (`buildkitQemuEmulatorMountPoint`), can de-dupe them now 🎉

core/container_exec.go
```diff
@@ -136,224 +149,262 @@ func (container *Container) WithExec(ctx context.Context, opts ContainerExecOpts
...
+			rerr = buildkit.RichError{
+				ExecError: rerr.(*errdefs.ExecError),
+				Origin:    op.opt.CauseCtx,
+				Mounts:    op.Mounts,
+				ExecMD:    execMD,
+				Meta:      metaSpec,
+				Secretenv: secretEnvs,
+			}
+		} else {
+			// Only release actives if err is nil.
```
sipsma, line 319: nit: since this is a success case, is there a reason to have it in this defer vs. just putting in the actual code below? Would make it easier to follow that way

jedevc:
Temporarily putting back into draft while I rework some bits, and handle the implications of all the rebasing :cold_sweat: 

TODO:
- [ ] Remove `RawDigest`
- [ ] Cleanup commented out code
- [ ] Tidy tidy tidy
- [ ] Cleanup git history (probably looking to rebase-merge this one)

core/container_exec.go
jedevc: No real reason, this is kind of just pulled in directly from buildkit.

But, also, I do *slightly* like it, it just keeps all the mount manager cleanup in one place?

core/container.go
jedevc: Two things:
- I've changed `HashFrom` to put in null bytes.
- I've removed the hellish `RawDigest` thing, and instead just made a generic `DigestOf` function (seems similar to what @sipsma is doing in #10497). And then have a special `Container.WithoutInputs` to "pureify" the container from the buildkit stuffs.

So these should both make sure we avoid nasty collisions.

sipsma (approved):

PR #10458, opened 2025-05-22
cli: Add --cloud flag (EXPERIMENTAL)

gerhard:
> [!WARNING]
> This is a highly **EXPERIMENTAL** feature. One day, all Dagger Cloud users **might** be able to use this. We expect this to change before it becomes available to everyone.

This flag requires that the CLI:
- is connected to Dagger Cloud
- has an org set (e.g. `dagger login <YOUR_CLOUD_ORG>`)
- that org must have the `CLOUD_ENGINES` feature enabled (or be a `Dagger-Internal` org)

This flag provisions a Cloud Engine just-in-time, connects to it and then runs all subsequent operations in it. **ALL** commands will work exactly the same, just that they will run against this Cloud Engine which is configured with:
- `16 vCPUs`
- `32GB RAM`
- `100GB NVMe storage`

> [!NOTE]
> If the CLI is a development version, the Engine provisioned will be `main` (a.k.a. latest development).

By default, the Engine shuts down after `30 minutes` of inactivity (a.k.a. TTL). As soon as another command runs, the TTL resets to `30 minutes`.

State persists across Engines. After 30+ minutes of inactivity, running another command will start a new Engine with the same state that the last running Engine had. While there are a few scenarios which may result in engines starting with no state, this should happen no more than 1 in 10 runs.

The other half of this: https://github.com/dagger/dagger.io/pull/4350 (private repo PR)

---

https://github.com/user-attachments/assets/1553939e-2935-44e5-abb7-84e3a4488886

---

If you are a `Dagger` Cloud Org user, you can test this using the following steps:

1. Build a development version of the `dagger` CLI
2. Point it to a preview Dagger Cloud API
`export DAGGER_CLOUD_URL=https://api-parc-create-engine-api.preview.dagger.cloud` 👈 this part is **essential**
3. Login to Dagger Cloud
4. Run any `dagger` CLI command in an on-demand Dagger Cloud Engine

### DIFF
--- a/engine/client/drivers/cloud.go
+++ b/engine/client/drivers/cloud.go
@@ -1,0 +1,99 @@
+package drivers
+
+import (
+	"context"
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"net"
+	"net/url"
+	"time"
+
+	"github.com/dagger/dagger/internal/cloud"
+	"github.com/dagger/dagger/internal/cloud/auth"
+)
+
+var (
+	TLSHandshakeTimeout = 15 * time.Second
+)
+
+func init() {
+	register("dagger-cloud", &daggerCloudDriver{})
+}
+
+// daggerCloudDriver creates and manages a Cloud Engine, then connects to it
+type daggerCloudDriver struct{}
+
+type daggerCloudConnector struct {
+	EngineSpec cloud.EngineSpec
+}
+
+func (dc *daggerCloudConnector) Connect(ctx context.Context) (net.Conn, error) {
+	serverAddr := dc.EngineSpec.URL
+
+	// Extract hostname for SNI
+	host, _, err := net.SplitHostPort(serverAddr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to split host and port from '%s': %w", serverAddr, err)
+	}
+
+	clientCert, err := dc.EngineSpec.TLSCertificate()
+	if err != nil {
+		return nil, fmt.Errorf("failed to unserialize Cloud Engine cert: %w", err)
+	}
+
+	tlsConfig := &tls.Config{
+		Certificates: []tls.Certificate{*clientCert},
+		ServerName:   host,             // Required for SNI and server certificate validation
+		MinVersion:   tls.VersionTLS12, // TLS 1.2 is the oldest version that is still maintained in 2025
+	}
+
+	dialer := net.Dialer{Timeout: 10 * time.Second}
+	rawConn, err := dialer.DialContext(ctx, "tcp", serverAddr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial TCP to '%s': %w", serverAddr, err)
+	}
+
+	// Create TLS Client Connection
+	// tls.Client wraps the rawConn. If Handshake fails, rawConn should be closed.
+	// If Handshake succeeds, closing tlsConn will close rawConn.
+	tlsConn := tls.Client(rawConn, tlsConfig)
+
+	// Use HandshakeContext for better cancellation and timeout control
+	handshakeCtx, cancelHandshake := context.WithTimeout(ctx, TLSHandshakeTimeout)
+	defer cancelHandshake()
+
+	if err := tlsConn.HandshakeContext(handshakeCtx); err != nil {
+		rawConn.Close() // Ensure raw connection is closed if handshake fails
+		return nil, fmt.Errorf("tls handshake failed with '%s': %w", serverAddr, err)
+	}
+
+	return tlsConn, nil
+}
+
+func (d *daggerCloudDriver) Provision(ctx context.Context, _ *url.URL, opts *DriverOpts) (Connector, error) {
+	client, err := cloud.NewClient(ctx)
+	if err != nil {
+		return nil, errors.New("please run `dagger login <org>` first")
+	}
+
+	_, err = client.User(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	org, _ := auth.CurrentOrg()
+	if org == nil {
+		return nil, errors.New("please associate this Engine with an org by running `dagger login <org>")
+	}
+
+	return d.create(ctx, client)
+}
+
+func (d *daggerCloudDriver) create(ctx context.Context, client *cloud.Client) (*daggerCloudConnector, error) {
+	engineSpec, err := client.Engine(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &daggerCloudConnector{EngineSpec: *engineSpec}, nil
+}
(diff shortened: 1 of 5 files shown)


fidencio:
Considering this will be targetting dagger cloud, I also think that `--cloud` is better / more clear.

gerhard:
My first action item is to rename to `--cloud`, thank you 🙏

> @shykes Am I understanding correctly that the specific provider and machine configuration is completely hidden from the client? If so, I like that too :) We can worry about customization as a follow-up

Yes. "Engine configs" will be a follow-up that will add the customisation. The most important elements are size (how much cpu / memory / disk) & location (US, EU, etc.).

Instead of exposing providers, we may want to expose clouds (e.g. AWS, GCP, Azure, etc.). We need to think about this some more, since that will make the options more difficult as not all choices are available across all providers. Maybe we can start with a subset of options - size, architecture, location, cache size? - and see how those land before adding providers. Future discuss, for sure.

> @shykes I like the default lifecycle: one machine per user, with auto-shutdown. Simple, efficient. We can expand from there in the future.

That's exactly what we were aiming for too in this first iteration. As a quick follow-up, we could expose explicit tags, so that we can create as many Engines as needed, with different cache namespaces.

> @shykes Looking at the underlying implementation, it looks like the concept of a separate engine address, although hidden by the CLI, is still exposed by the underlying Dagger Cloud API endpoint (ie. API clients must retrieve a hostname and cert, then connect separately). I have no issue with this, as long as we agree it's a stopgap and in the future we will move that routing logic behind the Dagger Cloud API. For the same reasons we moved it behind the CLI: the more implementation details we hide, the more flexibility we have

Routing traffic through the API has significant implications on (top 3 of my mind)
- latency
- costs (network egress)
- blast radius (single point of failure)

This is a big & complex problem space that I will happily defer until the stopgap no longer works.

> @shykes A corollary of the previous point, is that we must be ready to break this API in the future. I'm guessing that won't be a problem

Yes, definitely. This is highly experimental that will be shipped and not released until it's proven to work well in practice.

> @shykes One last point: since we hide machine lifecycle from the client (good), to allow more powerful lifecycles in the future, we will need to get useful information from the client. The more useful information, the more powerful our routing and provisioning layer can be. How are you thinking about this problem?

In the current state, the most important information that we are getting from the client is the version. This determines the Engine version. Other pieces of information that we **might** want to use from the client fairly soon:
- function & arguments (e.g. for automatic cache name spacing)
- location (so that by default, we create Engines closest to the client)
- IaaS (this is a refinement on location, where we **might** want to create Engines on e.g. Azure if the client is on Azure too)

shykes:
> > @shykes Looking at the underlying implementation, it looks like the concept of a separate engine address, although hidden by the CLI, is still exposed by the underlying Dagger Cloud API endpoint (ie. API clients must retrieve a hostname and cert, then connect separately). I have no issue with this, as long as we agree it's a stopgap and in the future we will move that routing logic behind the Dagger Cloud API. For the same reasons we moved it behind the CLI: the more implementation details we hide, the more flexibility we have
> 
> Routing traffic through the API has significant implications on (top 3 of my mind)
> 
> * latency
> * costs (network egress)
> * blast radius (single point of failure)
> 
> This is a big & complex problem space that I will happily defer until the stopgap no longer works.

I wouldn't call the current design a stopgap. It's adding more moving parts to preemptively address the issues you mention: latency, cost, blast radius. So it's an optimization. In my opinion, a premature one.

Here's what I fear will happen:

- In theory, we can break the current API at any time, but in practice, there is inertia
- Sunk cost fallacy: "well we already did the work to expose the concept of engine endpoint to the client, they know how to query for it, how to deal with certificates etc; it's not worth the effort to change it"
- Drift: instead of a single endoint (dagger cloud API) there are two (dagger cloud + engine endpoint). Now we must do everything twice: load balancing, auth, clients, multi-homing etc. It would be much better to do all these things only once.


I think you would be better off grouping everything behind dagger cloud API endpoint like we discussed. Then handling optimizations at the infra layer, rather than app layer.

Example: egress costs. Run another dagger cloud api endpoint at the edge, then use HTTP redirects to that endpoint as needed. No need to reinvent redirects and duplicate auth at the app layer.

> > @shykes A corollary of the previous point, is that we must be ready to break this API in the future. I'm guessing that won't be a problem
> 
> Yes, definitely. This is highly experimental that will be shipped and not released until it's proven to work well in practice.

To reiterate my previous point: I worry that we are setting ourselves up to accept a premature optimization via sunk cost fallacy. It's not just that we maybe might break this API later. We *will* break it later to correct this, and I'm making sure you're ok with that.


> In the current state, the most important information that we are getting from the client is the version. This determines the Engine version. Other pieces of information that we **might** want to use from the client fairly soon:
> 
> * location (so that by default, we create Engines closest to the client)
> * IaaS (this is a refinement on location, where we **might** want to create Engines on e.g. Azure if the client is on Azure too)

Sounds great. I think module will be great too.

> * function & arguments (e.g. for automatic cache name spacing)

That would imply the ability to dispatch different function calls to different engines *from the same session*. This is harder to do if we don't move engine routing behind the cloud api endpoint (because engine routing can only happen once, at the beginning of the session, and that's hardcoded in the client so hard to change)

gerhard:
I noted all concerns & suggestions. We can change anything, anytime. This is an experimental feature, implemented the simplest way while pushing as many concerns as possible to layers that are already built & optimised for those concerns.

Dispatching function calls to different Engines, in the same session, is something that we've already been thinking about with @matipan. Keeping the discussions at that level of detail is helpful.

I am keen to get this implementation of the ideal architecture shipped before we iterate some more on the design. This is the third iteration, and it seems good enough to ship to me. Going to make the last few changes now so that all checks pass, then keen on getting this merged so that we can get feedback from more internal users.

Thanks for all the feedback!

shykes (approved):
@gerhard sorry if I appear to be overreacting to the "redirect vs. proxy" design disagreement.

Zooming out, besides this design point, I love the direction and think we should merge this asap to get people testing it.

Assuming this is a hidden command, the sooner we merge the better! Easier to test with a stable release.

gerhard:
Thank you for the ✅

Merging as all checks are now green. Keen to see how this behaves in a stable CLI release.

engine/client/drivers/cloud.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/dagger/dagger/internal/cloud"
+	"github.com/dagger/dagger/internal/cloud/auth"
+)
+
+var (
+	TLSHandshakeTimeout = 15 * time.Second
+)
+
+func init() {
+	register("dagger-cloud", &daggerCloudDriver{})
```
jedevc, line 21: nit: could we just claim `cloud://` instead?

gerhard: Just followed existing conventions here:
- `docker-container`
- `docker-image`
- `kube-pod`
- `podman-container`

`tcp`, `unix` & `ssh` are the only exceptions which are standard protocols.

Don't you think that introducing `cloud://` here would be confusing?

PR #10459, opened 2025-05-22
feat(ts-sdk): correctly the error using returnError

TomChv:
When an error is thrown by a TypeScript module, call `ReturnError` instead of logging the error and exit 1.

It seems I never applied the change introduced b
https://github.com/dagger/dagger/pull/8442 so here it is.

### DIFF
--- a/sdk/typescript/src/module/entrypoint/entrypoint.ts
+++ b/sdk/typescript/src/module/entrypoint/entrypoint.ts
@@ -7,7 +7,7 @@ import { load } from "./load.js"
 import { register } from "./register.js"
 
 export async function entrypoint(files: string[]) {
-  // Start a Dagger session to get the call context
+  // Start a Dagger session to get the cƒall context
   await connection(
     async () => {
       const fnCall = dag.currentFunctionCall()
@@ -45,16 +45,9 @@ export async function entrypoint(files: string[]) {
             parentArgs,
             fnArgs: args,
           })
-        } catch (e) {
-          if (e instanceof Error) {
-            if (e.cause) {
-              console.error(`${e.cause}`)
-            }
-            console.error(`Error: ${e.message}`)
-          } else {
-            console.error(e)
-          }
-          process.exit(1)
+        } catch (e: unknown) {
+          await fnCall.returnError(dag.error(stringifyError(e)))
+          return
         }
       }
 
@@ -70,4 +63,33 @@ export async function entrypoint(files: string[]) {
     },
     { LogOutput: process.stdout },
   )
+}
+
+/**
+ * Take the error thrown by the user module and stringify it so it can
+ * be returned to Dagger.
+ *
+ * If the error is an instande of Error, we stringify the message and the cause.
+ * If the error is not an instance of Error (which is the case for unexpected errors happening
+ * inside the code), we try to stringify it, if it fails we convert it to a string.
+ *
+ * Hopefully, this will be enough to make the error readable by the user.
+ * If the stringify fails, it will fails when calling `returnError` and should
+ * still be displayed to the user and Cloud.
+ */
+function stringifyError(e: unknown): string {
+  if (e instanceof Error) {
+    let error = `${e.name}: ${e.message}`
+    if (e.cause) {
+      error += `: ${e.cause}`
+    }
+
+    return error
+  }
+
+  try {
+    return JSON.stringify(e)
+  } catch {
+    return String(e)
+  }
 }


sdk/typescript/src/module/entrypoint/entrypoint.ts
```diff
@@ -45,16 +45,9 @@ export async function entrypoint(files: string[]) {
...
-            if (e.cause) {
-              console.error(`${e.cause}`)
-            }
-            console.error(`Error: ${e.message}`)
-          } else {
-            console.error(e)
-          }
-          process.exit(1)
+        } catch (e: unknown) {
+          await fnCall.returnError(dag.error(stringifyError(e)))
```
vito, line 49: Nice, this is a good start - but it would be extra nice if we could preserve any GraphQL error extensions, like we do in the Go SDK:

```go
	var gqlErr *gqlerror.Error
	if errors.As(rerr, &gqlErr) {
		if gqlErr.Extensions != nil {
				keys = append(keys, k)
			}
			sort.Strings(keys)
			for _, k := range keys {
				if err != nil {
				}
			}
		}
		return dagErr
	}
}
```

Hopefully there's a particular error type we can check for for that? You can test it by calling a function that returns an exec error, like `echo foo >&2; exit 1`.

sdk/typescript/src/module/entrypoint/entrypoint.ts
```diff
@@ -71,3 +64,32 @@ export async function entrypoint(files: string[]) {
...
+ * inside the code), we try to stringify it, if it fails we convert it to a string.
+ *
+ * Hopefully, this will be enough to make the error readable by the user.
+ * If the stringify fails, it will fails when calling `returnError` and should
+ * still be displayed to the user and Cloud.
+ */
+function stringifyError(e: unknown): string {
+  if (e instanceof Error) {
+    let error = `${e.name}: ${e.message}`
+    if (e.cause) {
```
vito, line 83: FWIW you can also attach arbitrary named JSON values to the Error type, which will get bubbled up as GraphQL error extensions.

Similar to the other comment, which _preserves_ GraphQL error extensions, this would be a way for you to attach values of your own, which will get bubbled up to places like the LLM.

For example maybe a stack trace would be useful to attach, if that's readily available?

sdk/typescript/src/module/entrypoint/entrypoint.ts
```diff
@@ -7,7 +7,7 @@ import { load } from "./load.js"
 import { register } from "./register.js"
 
 export async function entrypoint(files: string[]) {
-  // Start a Dagger session to get the call context
+  // Start a Dagger session to get the cƒall context
```
vito, line 10: lil typo:
```suggestion
  // Start a Dagger session to get the call context
```

sdk/typescript/src/module/entrypoint/entrypoint.ts
```diff
@@ -71,3 +64,32 @@ export async function entrypoint(files: string[]) {
     { LogOutput: process.stdout },
   )
 }
+
+/**
+ * Take the error thrown by the user module and stringify it so it can
+ * be returned to Dagger.
+ *
+ * If the error is an instande of Error, we stringify the message and the cause.
```
vito, line 72:
```suggestion
 * If the error is an instance of Error, we stringify the message and the cause.
```

vito (commented):
Looking good, thanks for picking this up!

I think it's worth adding a golden test for this so that we can directly see how the DX improves. You can use the Go golden tests `fail-log`, `fail-effect`, and `fail-log-native` for reference.

sdk/typescript/src/module/entrypoint/entrypoint.ts
TomChv: Okay, with the logic I implemented, I'm translating any existing fields in the throw error.

The GraphQL error extension is [already parsed into an exec Error object with the field stderr and stdout](https://github.com/dagger/dagger/blob/main/sdk/typescript/src/common/graphql/compute_query.ts#L220), so I can needed to assign them to the `dagger.error` object.

So I guess it's good with this PR changes?

TomChv: Done

sdk/typescript/src/module/entrypoint/entrypoint.ts
```diff
@@ -71,3 +66,39 @@ export async function entrypoint(files: string[]) {
...
+function formatError(e: unknown): DaggerError {
+  if (e instanceof Error) {
+    let error = dag.error(e.message)
+
+    // If the error is an instance of ExecError, we can add the extensions fields
+    // in the error object.
+    if (e instanceof ExecError) {
+      Object.entries(e)
+        .filter(([key]) =>
+          ["cmd", "exitCode", "stdout", "stderr"].includes(key),
```
vito: Ah sorry, what I said was confusing, and there are new developments now with #10512 

What we want is...:

* When a GraphQL client error is turned into an `ExecError`, we need to keep all the extended error fields reachable somehow, and turn them back into fields on the `dag.error()`, exactly as they were prior to turning it into the nice `ExecError` type. With #10512 this will critically include the OTel trace context - the _origin_ of the error. I guess you could just hold a reference from the `ExecError` back to the original `ClientError`?
* When a GraphQL `ClientError` is returned with arbitrary extended fields, we _also_ need to preserve those and turn them back into fields on the `dag.error()`. I see that currently gets turned into a `GraphQLRequestError`, so I think we need to reach into the underlying `ClientError` and extract the extensions from there.

TomChv: Should be better that way then ([bedb110](https://github.com/dagger/dagger/pull/10459/commits/bedb1102a77f574c0a1c701602af735a21e084aa)):

```ts
    // If the error is an instance of GraphQLError or a inherit type of it,
    // we can add the extensions fields in the error object.
    if (e instanceof ExecError || e instanceof GraphQLRequestError) {
      Object.entries(e.extensions ?? []).forEach(([key, value]) => {
        if (value !== "" && value !== undefined && value !== null) {
          error = error.withValue(key, JSON.stringify(value) as JSON)
        }
      })
    }
```

For any ExecError or GQLError (ClientError), I keep the extensions in a field and set them in the dagger error.

sdk/typescript/src/module/entrypoint/entrypoint.ts
```diff
@@ -71,3 +67,35 @@ export async function entrypoint(files: string[]) {
...
+    }
+
+    return error
+  }
+
+  try {
+    return dag.error(JSON.stringify(e))
+  } catch {
+    return dag.error(String(e))
+  }
```
vito, line 100: What's this meant to handle? :thinking: Looks a little odd but maybe because I'm missing JS/TS knowledge.

I guess this is because non-`Error` values can be thrown? Should we even turn those into errors?

vito (approved):
lgtm (modulo one question)! I'll test it out with #10512 if this merges first - theoretically the errors will become even cleaner.

sdk/typescript/src/module/entrypoint/entrypoint.ts
vito: Looks perfect :tada:

sdk/typescript/src/module/entrypoint/entrypoint.ts
TomChv: Yeah it's possible in case of a "panic" in the code, for example if the user tries to access an undefined value in its module, in these case, I'm not sure the exception will be in an `Error` object so it's an extra safety measure to correctly return the error to the user instead of crashing and display that `tsx xxx` error message.

However we shouldn't hit that case in normal scenario

PR #10460, opened 2025-05-22
llm: bump Claude default to 4.0 Sonnet

vito:


shykes:
We didn't bump to 3.7 because it had a reputation of being over-eager and super expensive. Is 4 different in this regard?

shykes (approved):

Issue #10112, opened 2025-04-09
TUI: default verbosity that doesn't scare away normal people


shykes:
## Problem

The default verbosity of the TUI is awesome, but... it's a lot. People get overwhelmed - like running shell scripts with `set -x` by default. Or stracing your unix tools by default.

This is actually 2 problems in 1:

- Problem 1: the default verbosity is too high
- Problem 2: the lower verbosity, as implemented, is still overwhelming. We need a *different* kind of low-verbosity output

## Solution

Default output should be a new output, with a *fixed depth* of 2. Basically, only show top-level function calls, and their sub-calls - then stop there. No more detail.

nipuna-perera:
I want to request the same for non TUI output also if possible. The `--progress=plain` output right now is very verbose by default and really hard to find things if you don't know exactly what you are looking for. Even then I have to stumble a bunch before I get to the meaningful part.

shykes:
Update: today I have felt the need for this again.

vito:
What if we make the TUI match the web UI, so everything is collapsed by default, and instead of a global verbosity level controlling whether _everything_ is expanded, you individually expand/collapse spans?

I'm down to try the depth=2 trick too, just fishing for any "a-ha" alternatives first!

shykes:
Just checking on this. The TUI has been improved in other ways, but this remains a problem, at least to me.

vito:
@shykes Not a whole lot to go on there. 🙁 The 'what you saw' vs. 'what you wanted to see' diff helped a ton in the error output thread and led to it being immediately addressed. Want to try that here?

I can take a quick guess that there's an off-by-one error involved; right now it's effectively depth=3 vs. what you described in the OP as depth=2 (in code it's `Depth <= 2` 😇)

PR #10468, opened 2025-05-25
tui: reduce verbosity, per-span toggles, new DAG style

vito:
* [x] Only expand 2 levels by default (#10112)
* [x] Support individually toggling spans instead of (but not replacing) global verbosity model
* [x] Reduce the amount of lines all over the place by switching to DAG style
* [x] UI cue to draw the eye towards error logs (bold red bars) and in-flight or failed spans (red/yellow lines)
* [x] Tuck CLI flags (File, Directory etc.) underneath `parsing command line arguments` span so they don't pollute toplevel (and add regression golden test, this used to work)
* [x] Add `E` hotkey for `--no-exit` at runtime
  * split out to #10511 
* [x] Finally fix ordering of ID fields to match schema order
* [x] Use multiline rendering for calls that have excessively long arguments
* [ ] ~~Fix macOS terminal~~
  * Need to investigate this separately
* [ ] ~~Can we simplify away from verbosity levels?~~
  * There might be something to this, but didn't figure it out yet
* [x] Take a swing at #10461
  * split out to #10512 

[asciicast](https://asciinema.org/a/6q6Ka8eTthfmr5nkBKW3ZyUoI)


cmd/dagger/flags.go
```diff
@@ -281,11 +281,11 @@ func (v *containerValue) String() string {
 	return v.address
 }
 
-func (v *containerValue) Get(_ context.Context, c *dagger.Client, _ *dagger.ModuleSource, _ *modFunctionArg) (any, error) {
+func (v *containerValue) Get(ctx context.Context, c *dagger.Client, _ *dagger.ModuleSource, _ *modFunctionArg) (any, error) {
 	if v.address == "" {
 		return nil, fmt.Errorf("container address cannot be empty")
 	}
-	return c.Container().From(v.String()), nil
+	return c.Container().From(v.String()).Sync(ctx)
```
jedevc, line 288: I *seem* to remember that we weren't doing this for perf reasons before, but I think the telemetry really does justify changing it. cc @helderco, I have *vague* memories of discussing this, but can't quite remember if there was anything else.

One point though, it might be worth error-group-ing calls to `GetFlagValue`, so at least these could be loaded in parallel?

jedevc:
Looking now :tada:

One thing that seems off, is in a clean state, running `dagger-dev ./hack/build`, it seems to hang for a really long time with only `connect` output on the default verbosity? Not really sure *why* though.

vito:
> One thing that seems off, is in a clean state, running `dagger-dev ./hack/build`, it seems to hang for a really long time with only `connect` output on the default verbosity? Not really sure _why_ though.

Huh good catch, for some reason the `looking for module` span isn't showing up! Will look into it

edit: that span is marked `internal` for some reason :thinking: - I'm 99% sure this is a pre-existing issue, I've noticed the same hang before in regular usage

cmd/dagger/flags.go
vito: Yeah good point - addressed this in https://github.com/dagger/dagger/pull/10468/commits/e8f1c8127f623e592988635070418e1715cc9d6c using `conc`

vito:
Fixed up all the points raised - gonna go ahead and merge to unblock the errors PR! Can address any further feedback there.

PR #10476, opened 2025-05-26
Support remotes engines with all auth methods (EXPERIMENTAL)

matipan:
Follow up to #10458.

There are two main authentication mechanisms in the CLI/SDKs:
- DAGGER_CLOUD_TOKEN: used primarily in CI environments
- `dagger login`: used by individual users, likely in their machines

We currently only support using `dagger login` to provision remote engines. We should also support the use of engine tokens since what runs on CI and other remote environments. We faced this limitation when building an internal PoC that uses the Go SDK to `dagger.Connect` to an engine. At the moment we can target a remote runner by setting `dagger.WithRunnerHost("dagger-cloud://default-engine")`. This is definitely a point for debate, maybe we want this to be a separate config. But since we are in an experimental/prototyping phase I think its worth keeping it simple and using the existing drivers mechanism.

### DIFF
--- a/engine/client/drivers/cloud.go
+++ b/engine/client/drivers/cloud.go
@@ -10,7 +10,6 @@ import (
 	"time"
 
 	"github.com/dagger/dagger/internal/cloud"
-	"github.com/dagger/dagger/internal/cloud/auth"
 )
 
 var (
@@ -74,17 +73,7 @@ func (dc *daggerCloudConnector) Connect(ctx context.Context) (net.Conn, error) {
 func (d *daggerCloudDriver) Provision(ctx context.Context, _ *url.URL, opts *DriverOpts) (Connector, error) {
 	client, err := cloud.NewClient(ctx)
 	if err != nil {
-		return nil, errors.New("please run `dagger login <org>` first")
-	}
-
-	_, err = client.User(ctx)
-	if err != nil {
-		return nil, err
-	}
-
-	org, _ := auth.CurrentOrg()
-	if org == nil {
-		return nil, errors.New("please associate this Engine with an org by running `dagger login <org>")
+		return nil, errors.New("please run `dagger login <org>` first or configure a DAGGER_CLOUD_TOKEN")
 	}
 
 	return d.create(ctx, client)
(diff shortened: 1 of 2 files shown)


jedevc:
Not sure I understand the limitations of the driver mechanism here? I quite like that you can do `dagger.WithRunnerHost`.

shykes:
One issue with `dagger-cloud://` is that it assumes that multiple engine endpoints are exposed to the client, and it's up to the client to explicitly select this or that endpoint. We don't know for sure we want to do that, so locking in that design is a one-way door.

The alternative is that Dagger Cloud chooses the endpoint for you, based on the context: your CLI version, username, platform, current module, and in the future function call. You could say "that's just `dagger-cloud://default-engine`", but then you're making the main use case more complicated.

gerhard (approved):
This makes perfect sense!

I explicitly focused on a single auth mechanism, and deferred `DAGGER_CLOUD_TOKEN` to when it was needed. It seems that we've hit it within days, and I'm glad that we did it this way, as the changes are small, and they make perfect sense to me.

Good to merge as soon as all checks are green. Excited to see this land in the next release 🚀

> [!NOTE]
> As a reminder, as soon as this gets merged, we can test it out with:
> <pre>curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_COMMIT=head sh</pre>

gerhard:
Does a dedicated function to connect to remote Engines sound a sensible next step? Something that could be used from within sessions too, meaning that we could open new connections from **within** sessions? To me, this sounds like new core API functions. We could go down the module path too, but this functionality seems fundamental. `Cloud()` feels appropriate here.

I always thought of the current implementation of `dagger-cloud://` as a stopgap, until we figure out all the other pieces. We are pragmatic here and use what we have, but the end-goal is programatic dispatch across `n` Engines which behave like a worker pool and **maybe** pull jobs instead of clients pushing jobs, which is what is happening today.

In that world, we are simply enqueuing jobs, and waiting for the result to arrive asynchronously, which doesn't require the client to stick around beyond the initial job enqueue. The tricky part is clients which need to upload local context, which is why I am putting a pin in this - it's impractical to unpack with the building blocks that we have, which is why...

WDYT about a first-class `Cloud()` function that maybe chains with `Engine()` something along the lines of:

```
dag, err := Cloud().Engine()
```

With the above, all calls to `dag` would be sent to this Cloud Engine.

PR #10496, opened 2025-05-28
core: JSONValue

shykes:
Implement a new core object type: `JSONValue`. See #10370

### DIFF
--- a/core/jsonvalue.go
+++ b/core/jsonvalue.go
@@ -10,7 +10,7 @@ import (
 
 func (*JSONValue) Type() *ast.Type {
 	return &ast.Type{
-		NamedType: "JSONObject",
+		NamedType: "JSONValue",
 		NonNull:   true,
 	}
 }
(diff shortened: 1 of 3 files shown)


core/jsonvalue.go
```diff
@@ -0,0 +1,137 @@
...
+	b, err := json.Marshal(v)
+	if err != nil {
+		return &JSONValue{}, err
+	}
+	return &JSONValue{data: b}, nil
+}
+
+// Set sets value at path (dot-separated).
+// "" or "." replaces the whole value.
+func (o *JSONValue) Set(path string, value any) (*JSONValue, error) {
```
jedevc, line 33: WDYT about supporting arbitrary JSONPath? There's a go implementation we could use here: https://github.com/oliveagle/jsonpath

core/schema/jsonvalue.go
```diff
@@ -0,0 +1,161 @@
...
+func (s jsonvalueSchema) Install() {
+	// Top-level constructor: jsonvalue (creates new empty jsonvalue)
+	dagql.Fields[*core.Query]{
+		dagql.Func("json", s.newJSONValue).
+			Doc("Initialize an empty JSON value"),
+	}.Install(s.srv)
+
+	// Expose methods for JSONValue manipulation
+	dagql.Fields[*core.JSONValue]{
+		dagql.Func("get", s.unset).Doc("Return the JSON-encoded value, or a sub-value at the given path").Args(
```
jedevc, line 27: Instead of having dedicated `get<Type>`/`set<Type>` methods, what about having just `get`+`set` and then `as<Type>` methods? Then they chain as, `.Get("x").AsInt()`, which feels very much like interfaces...

For `set`, we'd want to have a scalar value that can accept *any* value (I think this is possible, happy to provide pointers, if this seems like a reasonable idea).

shykes:
Can't reopen... This PR bot is getting on my nerves

shykes:
test

core/schema/jsonvalue.go
shykes: How would the "universal scalar" type work?

PR #10497, opened 2025-05-28
[WIP] Support for persistent dagql cache + multiple cache keys per-object

sipsma:
This is deeply WIP still, sending out to track progress publicly and start getting some real CI runs soon.

This is the next big step forward in making our cache logic "dagger native" and consists of:
1. A persistent sqlite-based DB for dagql cache
2. Support for associating and matching cache based on multiple cache keys associated with a single object in dagql
   * This is very in the weeds but also pretty fundamental for supporting content-based caching and "per-client state" in dagql natively w/out reliance on buildkit

Plan is:
1. *This PR* - Enable the persistent cache just for selected APIs that are already "dagger-native" (i.e. APIs that have always been implemented without buildkit or one of the APIs that have been updated to use dag-op)
   * This will get the new implementation out in the real world without having a blast radius of "literally everything"
   * Goal is to just maintain current performance (i.e. swapping APIs to the new cache shouldn't be slower than existing buildkit cache)
2. *Follow-up to this PR* - enable persistent module function call caching using the new DB
   * Expected to be a **massive** user-facing performance improvement 🎉
   * Will entail lots of breaking-change handling that's best saved for a follow-up effort rather than entangling with the already-too-large changes here

---

TODOs
- [ ] Existing API support
   - [x] Enable *selective* use of persistent dagql cache
      - Currently anything not marked as `DoNotCache` is ending up in the DB, which isn't what we want yet (as mentioned above)
   - [ ] Update `Instance` to allow writing more digests for the object
      - The DB supports this, but still need to wire into API used by core schema
   - [ ] Due-diligence to ensure Secret plaintexts never get written to cache (not checking at all currently)
   - [ ] Git (currently commented out)
      - [ ] Use new APIs for associating multiple digests w/ an object to support content-caching (needed for git module support most crucially)
      - [ ] Manage dag-op-ified objects w/ cache refs using the persistent dagql cache rather than buildkit cache
   - [ ] Modules+function calls (currently commented out)
      - [ ] Incorporate module source digest into `metaDigest` column for calls in db
   - [ ] HTTP
      - [ ] Manage dag-op-ified objects w/ cache refs using the persistent dagql cache rather than buildkit cache
   - [ ] `Container.withExec`
      - [ ] Manage dag-op-ified objects w/ cache refs using the persistent dagql cache rather than buildkit cache
   - [ ] `Directory.withSymlink`
      - [ ] Manage dag-op-ified objects w/ cache refs using the persistent dagql cache rather than buildkit cache
- [ ] Telemetry
   - [x] Re-integrate; just commented out related lines while refactoring code. Not expected to be super impacted yet
   - [ ] Re-add support for `CACHED` status on dagql cache; db interface needs to return this from its various methods
- [ ] DB
   - [x] Wrap in existing "session cache" interface and enable releasing refs when session ends.
   - [ ] Pruning support 
      - Gonna be in an "awkward in-between state" since both the dagql cache and buildkit cache will exist for a bit still. Will likely push some changes to our buildkit fork to handle this for now.
   - [ ] Primitive engine version upgrade and/or schema versioning handling
     - Only goal for now is to not end up in situation where an engine tries to use a pre-existing db schema that's out of sync with what it expected; if it sees one it should rm it and make a new db.
     - "Proper" migration handling such that the cache isn't thrown out is out-of-scope of this PR
   - [ ] Low hanging performance improvements (will skip for now if any end up feeling like pre-mature optimization)
     - [ ] Pre-compile queries (one per-field since each field has a unique set of queries)
     - [ ] Handle scalars more sanely
       - The current db schema requires all call dependencies, including args, to be a "result" in the DB, which means every unique scalar used as an arg needs to be written as a row to a table. This is simple to start but silly and not performant, the schema should be updated (i.e. complicated slightly) to handle this better
     - [ ] Investigate other sqlite go interfaces a bit more to see performance impact
       - [useful external benchmarks here](https://github.com/cvilsmeier/go-sqlite-bench)
     - [ ] Try variations on sql queries (there's various approaches to the set-intersection problem we do here which *may* have subtle performance tradeoffs)
     - [ ] Optimize field selection with sqlite `json` handling (currently all cached individually, which is technically n^2 in space for nested objects)
     - [ ] Optimize list selection (currently modeled as a "`selectNth` pseudo-call" for simplicity)
- [ ] Testing
   - [ ] Determine what new integ tests are needed, add them
   - [ ] Do some manual testing against "real-world" user pipelines to catch any gaps in integ tests early
- [ ] Internal docs
   - [ ] Document db schema and queries; they are a bit subtle at first glance since they need to handle A) arbitrary numbers/types of args on dynamically loaded fields and B) checking cache when input objects can be associated w/ multiple digests

---

Related spun-out PRs:
- https://github.com/dagger/dagger/pull/10410
- https://github.com/dagger/dagger/pull/10540
- https://github.com/dagger/dagger/pull/10586
- https://github.com/dagger/dagger/pull/10620


sipsma:
Been working on the various spin-off PRs noted in the description here; this is still in progress (just across split-out PRs for the most part at the moment).

sipsma:
Trying to do this as one PR turned out to be silly in retrospect. Had already been splitting it out into huge separate PRs noted in the description. Most recent work continued in https://github.com/dagger/dagger/pull/10975, which added sqlite to the dagql cache for the purpose of tracking function call TTLs. Will continue the effort in separate PRs

PR #10501, opened 2025-05-29
shell: add .printenv builtin command for showing env variables

### LINKED ISSUE #10021: ✨ shell: Add .env command

There are often times when you need to see env vars that the shell has, it would be nice to have the `.env` command as a built-in. By default, with no arguments - entering `.env` would show all env vars as a list. Might be worthwhile to also support a specific var with `.env FOO_BAR` to eliminate calls like `env | grep FOO`.

_No response_

Currently you have to use `.echo $FOO_BAR` to check if the shell has an environment variable.

jasonmccallister:
Adds a new builtin command for `.printenv` that optional accepts an argument of the specific variable to show. If no argument is provided, it will print all available env vars in the shell using the interpreter.

Resolves #10021

### DIFF
--- a/cmd/dagger/shell_commands.go
+++ b/cmd/dagger/shell_commands.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 	"slices"
 	"strings"
@@ -438,6 +439,35 @@ func (h *shellCallHandler) registerCommands() { //nolint:gocyclo
 
 Writes any specified operands, separated by single blank (' ') characters and followed by a newline ('\n') character, to the standard output. If the -n option is specified, the trailing newline is suppressed.
 `,
+		},
+		&ShellCommand{
+			Use:  ".env [name]",
+			Args: MaximumArgs(1),
+			Description: `Show available environment variables or a specific variable
+
+If no name is provided, all environment variables are printed. If a name is provided, the value of that environment variable is printed.
+			`,
+			State: NoState,
+			Run: func(ctx context.Context, cmd *ShellCommand, args []string, _ *ShellState) error {
+				if len(args) == 0 {
+					// Print all environment variables
+					env := os.Environ()
+					if len(env) == 0 {
+						return h.Print(ctx, "No environment variables set")
+					}
+
+					return h.Print(ctx, strings.Join(env, "\n"))
+				}
+
+				// Print a specific environment variable
+				name := args[0]
+				value, ok := os.LookupEnv(name)
+				if !ok {
+					return fmt.Errorf("environment variable %q not set", name)
+				}
+
+				return h.Print(ctx, value)
+			},
 		},
 		&ShellCommand{
 			Use: ".wait",
--- a/cmd/dagger/shell_commands.go
+++ b/cmd/dagger/shell_commands.go
@@ -11,7 +11,6 @@ import (
 
 	"dagger.io/dagger/telemetry"
 	"github.com/spf13/cobra"
-	"mvdan.cc/sh/v3/expand"
 	"mvdan.cc/sh/v3/interp"
 )
 
@@ -453,10 +452,9 @@ If no name is provided, all environment variables are printed. If a name is prov
 
 				if len(args) == 0 {
 					// Print all environment variables
-					hc.Env.Each(func(name string, v expand.Variable) bool {
-						fmt.Fprintf(hc.Stdout, "%s=%s\n", name, v.String())
-						return true
-					})
+					for name, vr := range hc.Env.Each {
+						fmt.Fprintf(hc.Stdout, "%s=%s\n", name, vr)
+					}
 
 					return nil
 				}
--- a/cmd/dagger/shell_commands.go
+++ b/cmd/dagger/shell_commands.go
@@ -441,7 +441,7 @@ Writes any specified operands, separated by single blank (' ') characters and fo
 `,
 		},
 		&ShellCommand{
-			Use:  ".env [name]",
+			Use:  ".printenv [name]",
 			Args: MaximumArgs(1),
 			Description: `Show available environment variables or a specific variable
 
--- a/cmd/dagger/shell_commands.go
+++ b/cmd/dagger/shell_commands.go
@@ -462,10 +462,6 @@ If no name is provided, all environment variables are printed. If a name is prov
 				// Print a specific environment variable
 				name := args[0]
 
-				// if the user prefixed the name with a $ sign, remove it
-				if strings.HasPrefix(name, "$") {
-					name = strings.TrimPrefix(name, "$")
-				}
 
 				value, ok := os.LookupEnv(name)
 				if !ok {
--- a/cmd/dagger/shell_commands.go
+++ b/cmd/dagger/shell_commands.go
@@ -461,6 +461,12 @@ If no name is provided, all environment variables are printed. If a name is prov
 
 				// Print a specific environment variable
 				name := args[0]
+
+				// if the user prefixed the name with a $ sign, remove it
+				if strings.HasPrefix(name, "$") {
+					name = strings.TrimPrefix(name, "$")
+				}
+
 				value, ok := os.LookupEnv(name)
 				if !ok {
 					return fmt.Errorf("environment variable %q not set", name)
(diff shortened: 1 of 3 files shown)


cmd/dagger/shell_commands.go
```diff
@@ -439,6 +440,41 @@ func (h *shellCallHandler) registerCommands() { //nolint:gocyclo
...
+			Run: func(ctx context.Context, cmd *ShellCommand, args []string, _ *ShellState) error {
+				if len(args) == 0 {
+					// Print all environment variables
+					env := os.Environ()
+					if len(env) == 0 {
+						return h.Print(ctx, "No environment variables set")
+					}
+
+					return h.Print(ctx, strings.Join(env, "\n"))
+				}
```
helderco, line 460: Instead of `os.Environ()` I think it's better to use the interpreter's own env vars. It'll reflect more precisely what vars are available in the script, even user defined ones. The host env vars are just the default when the shell starts. 
```suggestion
				hc := interp.HandlerCtx(ctx)
				if len(args) == 0 {
					// Print all environment variables
					hc.Env.Each(func(name string, v expand.Variable) bool {
						fmt.Fprintf(hc.Stdout, "%s=%s\n", name, v.String())
						return true
					})
					return nil
				}

```

cmd/dagger/shell_commands.go
```diff
@@ -439,6 +440,41 @@ func (h *shellCallHandler) registerCommands() { //nolint:gocyclo
...
+				if strings.HasPrefix(name, "$") {
+					name = strings.TrimPrefix(name, "$")
+				}
+
+				value, ok := os.LookupEnv(name)
+				if !ok {
+					return fmt.Errorf("environment variable %q not set", name)
+				}
+
+				return h.Print(ctx, value)
```
helderco, line 475:
```suggestion
				v := hc.Env.Get(name)
				if !v.IsSet() {
					return fmt.Errorf("environment variable %q not set", name)
				}

				return h.Print(ctx, v.String())

```

cmd/dagger/shell_env.go
helderco, line 1: Missed cleanup?

cmd/dagger/shell_commands.go
```diff
@@ -439,6 +440,41 @@ func (h *shellCallHandler) registerCommands() { //nolint:gocyclo
 Writes any specified operands, separated by single blank (' ') characters and followed by a newline ('\n') character, to the standard output. If the -n option is specified, the trailing newline is suppressed.
 `,
 		},
+		&ShellCommand{
+			Use:  ".env [name]",
```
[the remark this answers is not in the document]
jasonmccallister: Yeah, I think that is a better name. I realized there is a `env` command as well which might get confusing.

jasonmccallister:
@helderco addressed/committed changes. Thank you for the quick review and also good to know about the interpreter!

cmd/dagger/shell_commands.go
```diff
@@ -439,6 +440,38 @@ func (h *shellCallHandler) registerCommands() { //nolint:gocyclo
...
+			State: NoState,
+			Run: func(ctx context.Context, cmd *ShellCommand, args []string, _ *ShellState) error {
+				hc := interp.HandlerCtx(ctx)
+
+				if len(args) == 0 {
+					// Print all environment variables
+					hc.Env.Each(func(name string, v expand.Variable) bool {
+						fmt.Fprintf(hc.Stdout, "%s=%s\n", name, v.String())
+						return true
+					})
```
mvdan: FYI: hc.Env.Each is an iterator, so you can do:

```
for name, vr := range hc.Env.Each {
    fmt.Fprintf(hc.Stdout, "%s=%s\n", name, vr)
}
```

(As a bonus, this skips the explicit call to String, which is implied by %s)

helderco:
@jasonmccallister, can you add a couple test cases?

cmd/dagger/shell_commands.go
jasonmccallister: Thank you @mvdan, making that change now 🙏

jasonmccallister:
> can you add a couple test cases?

Absolutely, took me a minute to track down the location of the tests but added cases for 

1. Showing all env vars
2. Showing only one env var
3. Showing the error for a non-existing env var
4. Setting a env var using the shell and ensuring is shows with `.printenv`

core/integration/module_shell_test.go
```diff
@@ -1025,3 +1026,55 @@ func (ShellSuite) TestInterpreterBuiltins(ctx context.Context, t *testctx.T) {
 		requireErrOut(t, err, "does not exist")
 	})
 }
+
+func (ShellSuite) TestPrintenvCommand(ctx context.Context, t *testctx.T) {
+	t.Run("printenv all", func(ctx context.Context, t *testctx.T) {
+		actual := os.Environ()
```
helderco: The shell will see the env vars from the container it runs in, not where these tests are run. Rather than getting all env vars from a with exec though, I think it's enough to pick 1 to 3 known vars from that container.

core/integration/module_shell_test.go
```diff
@@ -1025,3 +1026,55 @@ func (ShellSuite) TestInterpreterBuiltins(ctx context.Context, t *testctx.T) {
...
+		for _, v := range actual {
+			require.Contains(t, out, v)
+		}
+	})
+
+	t.Run("printenv specific", func(ctx context.Context, t *testctx.T) {
+		if err := os.Setenv("FOOBAR", "baz"); err != nil {
+			t.Fatalf("failed to set environment variable: %v", err)
+		}
+		defer os.Unsetenv("FOOBAR")
```
helderco: Same as before, but even if it wasn't there's `t.Setenv()` that abstracts this, however it'll throw an error because it can't be used with `t.Parallel()` (enabled for all tests in the suite automatically). Simpler to just pick a known env var.

core/integration/module_shell_test.go
```diff
@@ -1025,3 +1026,55 @@ func (ShellSuite) TestInterpreterBuiltins(ctx context.Context, t *testctx.T) {
...
+
+		c := connect(ctx, t)
+		out, err := daggerCliBase(t, c).
+			With(daggerShell(`.printenv FOOBAR`)).
+			Stdout(ctx)
+		require.NoError(t, err)
+		require.NotEmpty(t, out)
+		require.Equal(t, "baz\n", out)
```
helderco: Trivial, but the `Equal` assertion already excludes `NotEmpty`.

core/integration/module_shell_test.go
```diff
@@ -1025,3 +1026,55 @@ func (ShellSuite) TestInterpreterBuiltins(ctx context.Context, t *testctx.T) {
...
+	})
+
+	t.Run("printenv non-existing", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+		_, err := daggerCliBase(t, c).
+			With(daggerShell(`.printenv NON_EXISTING_VAR`)).
+			Sync(ctx)
+		requireErrOut(t, err, "environment variable 'NON_EXISTING_VAR' not found")
```
helderco: Doesn't match:
```suggestion
		requireErrOut(t, err, `environment variable "NON_EXISTING_VAR" not set`)
```

core/integration/module_shell_test.go
```diff
@@ -1025,3 +1026,55 @@ func (ShellSuite) TestInterpreterBuiltins(ctx context.Context, t *testctx.T) {
...
+	})
+
+	t.Run("printenv shows set envs", func(ctx context.Context, t *testctx.T) {
+		script := `
+		ctr=$(container))
+		.printenv ctr`
```
helderco: Typo:
```suggestion
		script := `
		ctr=$(container)
		.printenv ctr`
```

core/integration/module_shell_test.go
```diff
@@ -1025,3 +1026,55 @@ func (ShellSuite) TestInterpreterBuiltins(ctx context.Context, t *testctx.T) {
...
+
+		c := connect(ctx, t)
+		out, err := daggerCliBase(t, c).
+			With(daggerShell(script)).
+			Stdout(ctx)
+		require.NoError(t, err)
+		require.Contains(t, out, "ctr=")
```
helderco: `printenv` doesn't print the var's name, just its value:
```suggestion
```

helderco (approved):

PR #10502, opened 2025-05-29
docs: agent guide: use test func in workspace instead of privileged env

kpenfound:
here's the change i'll be pushing to the 'agents in an existing project' guide today https://github.com/kpenfound/hello-dagger-py/pull/3
context:
- we will not have self-calling shipped before the workshop and hack night next week
- agent reliability with privileged env is not great,
- copying the existing test function into the workspace and removing the privileged env greatly increases reliability even though it is not a very clean solution (the agent is no longer using "the same tools" that devs use)

### DIFF
--- a/docs/current_docs/quickstart/agent/inproject.mdx
+++ b/docs/current_docs/quickstart/agent/inproject.mdx
@@ -178,7 +178,6 @@ This code creates a Dagger Function called `develop` that takes an assignment an
 There are a few important points to note here:
 - The variable `environment` is the environment to define inputs and outputs for the agent. Each input and output provides a description which acts as a declarative form of prompting.
 - Other Dagger module dependencies are automatically detected and made available for use in this environment. In this example, Dagger finds the `workspace` sub-module installed and automatically generates bindings for it: `with-workspace-input` and `with-workspace-output`. These bindings can be used to provide (or retrieve) the workspace to the agent.
-- The `environment` is privileged, which means that the environment includes all the Dagger Functions in the existing Daggerized CI pipeline, plus the core Dagger API. This allows the agent to use the existing `test` function to run unit tests.
 - The LLM is supplied with a prompt file instead of an inline prompt. By separating the prompt from the Dagger Function code, it becomes easier to maintain and update the instructions given to the LLM.
 
 The final step is to add the prompt file. Create a file at `.dagger/develop_prompt.md` with the following content:
@@ -191,10 +190,6 @@ Your assignment is: $assignment
 ## Constraints
 - Before writing code, analyze the Workspace to understand the project.
 - Do not make unneccessary changes.
-- Run tests with the HelloDagger Test tool.
-- Read and write files with the Workspace_read_file, Workspace_list_files, and Workspace_write_file tools
-- You must pass the Directory from the Workspace_get_source to the HelloDagger Test tool.
-- Do not select Container or Directory tools. Only use Workspace and HelloDagger tools
 - Always run tests to validate your code changes.
 - Do not stop until you have completed the assignment and the tests pass.
 ```
--- a/docs/current_docs/quickstart/agent/inproject.mdx
+++ b/docs/current_docs/quickstart/agent/inproject.mdx
@@ -8,7 +8,7 @@ import VideoPlayer from '../../../src/components/VideoPlayer';
 
 # Add an AI agent to an Existing Project
 
-Now that you've learned the basics of creating an AI agent with Dagger in [Built an AI agent](./index.mdx), its time to apply those lessons to a more realistic use case. In this guide, you will build an AI agent that builds features for an existing project.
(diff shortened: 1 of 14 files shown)


jpadams:
A diagram with the relationships between the project, `hello-dagger` module, `workspace` submodule and such might be helpful

jpadams:
We should link back to first AI agent quickstart from here since we're asking folks to dive it after it
image

jpadams:
image

jpadams:
> Make sure your LLM provider has been properly configured:

we should link back to the LLM setup

jpadams:
A little confusing if folks copy paste this whole block, since the commands will run serially.
Would be better to break them up so each one can be experienced.

image

jpadams:
I didn't get a blue page
```diff
diff --git a/src/views/HomeView.vue b/src/views/HomeView.vue
index d5c0217..e62f130 100644
--- a/src/views/HomeView.vue
+++ b/src/views/HomeView.vue
@@ -3,7 +3,7 @@ import TheWelcome from '../components/TheWelcome.vue'
 </script>

 <template>
-  <main>
+  <main style="background: #2196f3; min-height: 100vh;">
     <TheWelcome />
   </main>
 </template>
 ```

vikram-dagger:
Can't comment on these items inline since they're not part of the diff so pasting here. I may have misunderstood the change we are making, apologies in this case

- `Now that you've learned the basics of creating an AI agent with Dagger in [Built an AI agent]` -> typo, `Build`
- At this point `Replace the .dagger/workspace/main.go`, we are adding the `test` function to the workspace module. But since we are starting with the daggerized CI quickstart, doesn't it also exist in the main project module? If yes, then should we include instructions to remove it from there?
- `In this Dagger module, each Dagger Function performs a different operation:` -> the new `test` function is missing in the explanatory bullet list
- `This agent will make changes to the application and use the existing test function from the Daggerized CI pipeline` -> similar to previous, if `test` is moved into the workspace module, should this wording change?
- `The get-source Dagger Function gets the source code directory from the workspace.` -> I couldn't find this function in the code snippet
- `See the functions of the new workspace module:` - maybe useful to show the output here as a checkpoint for the user

kpenfound:
Thanks for all the feedback! I've pushed changes to incorporate all of it

@jpadams 

> I don't see generated videos in the docs preview via dagger call server up

I've seen this sometimes with `dagger call server up` but I believe it works sometimes too, its just inconsistent. Maybe something to look at in the docusaurus module

> A diagram with the relationships between the project, hello-dagger module, workspace submodule and such might be helpful

What kind of diagram? Like showing the file structure? or that the hello-dagger module has a dependency on workspace?

> A little confusing if folks copy paste this whole block, since the commands will run serially.
Would be better to break them up so each one can be experienced.

Thats how it was originally :P made the change

> I didn't get a blue page.

Yeah, LLMs be LLMing 😂 FWIW I did get a blue page in my tests. We do have the tip below that command that says "If the agent misbehaves, try providing more guidance in the prompt file based on what went wrong."


@vikram-dagger 

The test function isn't being moved to the workspace, its just being duplicated there. Ideally we wont have to do this in the future, but until we have self-calls this is the best way for the agent to get a test function. Because of this change, `get-source` has been deleted since the LLM does not need a getter anymore. Except for PHP because automatic getters aren't a thing in the PHP sdk yet

jpadams (approved):
👍 Awesome

jpadams:
Noticed this in the python too
```diff
diff --git a/docs/current_docs/quickstart/agent/snippets/part3/agent/python/agent.py.snippet b/docs/current_docs/quickstart/agent/snippets/part3/agent/python/agent.py.snippet
index 1de9cc8e6..bef86242f 100644
--- a/docs/current_docs/quickstart/agent/snippets/part3/agent/python/agent.py.snippet
+++ b/docs/current_docs/quickstart/agent/snippets/part3/agent/python/agent.py.snippet
@@ -2,7 +2,7 @@
 async def develop_issue(
     self,
     github_token: Annotated[
-        Secret, Doc("Github Token with permissions to write issues and contents")
+        dagger.Secret, Doc("Github Token with permissions to write issues and contents")
     ],
     issue_id: Annotated[int, Doc("Github issue number")],
     repository: Annotated[str, Doc("Github repository url")],
  ```

vikram-dagger:
> The test function isn't being moved to the workspace, its just being duplicated there. Ideally we wont have to do this in the future, but until we have self-calls this is the best way for the agent to get a test function. 

Would it actually make sense to move it, instead of duplicating it? In general, I don't like the idea of duplicating a function between a module and a sub-module, it seems like a bad pattern for us to show in a quickstart.

> Because of this change, `get-source` has been deleted since the LLM does not need a getter anymore. 

In this case, I think the output below needs a fix, since as you mentioned the code does not include a `source` function.

image

PR #10503, opened 2025-05-29
ci: Migrate the remaining jobs to an alternative CI runner

gerhard:
We have been hitting networking issues in AWS, this should address them and also speed things up. Let's check this before merging.

Also, use the platform cache across the SDK jobs since it seems to speed **each** job up by a good few minutes.

---
Note: wait for these results before merging (approve is fine).

### DIFF
--- a/.github/main.go
+++ b/.github/main.go
@@ -23,9 +23,10 @@ type CI struct {
 	// +private
 	Workflows *dagger.Gha
 
-	GithubRunner *dagger.Gha // +private
-	DaggerRunner *dagger.Gha // +private
-	Alt2Runner   *dagger.Gha // +private
+	GithubRunner       *dagger.Gha // +private
+	DaggerRunner       *dagger.Gha // +private
+	AltRunner          *dagger.Gha // +private
+	AltRunnerWithCache *dagger.Gha // +private
 }
 
 func New() *CI {
@@ -61,9 +62,16 @@ func New() *CI {
 			}),
 			WorkflowDefaults: workflow,
 		}),
-		Alt2Runner: dag.Gha(dagger.GhaOpts{
+		AltRunner: dag.Gha(dagger.GhaOpts{
 			JobDefaults: dag.Gha().Job("", "", dagger.GhaJobOpts{
-				Runner:         AltBronzeRunnerWithCache(),
+				Runner:         AltGoldRunner(),
+				TimeoutMinutes: timeoutMinutes,
+			}),
+			WorkflowDefaults: workflow,
+		}),
+		AltRunnerWithCache: dag.Gha(dagger.GhaOpts{
+			JobDefaults: dag.Gha().Job("", "", dagger.GhaJobOpts{
+				Runner:         AltSilverRunnerWithCache(),
 				TimeoutMinutes: timeoutMinutes,
 			}),
 			WorkflowDefaults: workflow,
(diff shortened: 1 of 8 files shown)


.github/main.go
```diff
@@ -173,48 +180,36 @@ func (ci *CI) withTestWorkflows(runner *dagger.Gha, name string) *CI {
...
-			{"client-generator", []string{"TestClientGenerator"}, &dagger.GhaJobOpts{
-				Runner: AltGoldRunner(),
-			}},
-			{"interface", []string{"TestInterface"}, &dagger.GhaJobOpts{
-				Runner: AltGoldRunner(),
-			}},
-			{"call-and-shell", []string{"TestCall", "TestShell", "TestDaggerCMD"}, &dagger.GhaJobOpts{
-				Runner: AltGoldRunner(),
-			}},
+			{"container", []string{"TestContainer"}, &dagger.GhaJobOpts{}},
```
jedevc, line 207: nit: we should allow this to be `nil`, and handle that in `splitTests`.

jedevc (approved):

.github/main.go
gerhard: I tried this, and the explicit `&dagger.GhaJobOpts{}` felt more honest & easier to follow for me.

It also made it easier to know what type that `nil` was keeping the place of. Adding extra properties was also easier - the type was already there.

Adding the diff here for completeness, but the current implementation feels right to me:

```diff
diff --git a/.github/main.go b/.github/main.go
index f0e320861..7d6a9a908 100644
--- a/.github/main.go
+++ b/.github/main.go
@@ -197,19 +197,19 @@ func (ci *CI) withTestWorkflows(runner *dagger.Gha, name string) *CI {
 			Runner: AltBronzeRunnerWithCache(),
 		})).
 		With(splitTests(runner, "testdev-", true, []testSplit{
-			{"cgroupsv2", []string{"TestProvision", "TestTelemetry"}, &dagger.GhaJobOpts{}},
+			{"cgroupsv2", []string{"TestProvision", "TestTelemetry"}, nil},
 			{"modules", []string{"TestModule"}, &dagger.GhaJobOpts{
 				Runner: AltPlatinumRunner(),
 			}},
 			{"module-runtimes", []string{"TestGo", "TestPython", "TestTypescript", "TestElixir", "TestPHP", "TestJava"}, &dagger.GhaJobOpts{
 				Runner: AltPlatinumRunner(),
 			}},
-			{"container", []string{"TestContainer"}, &dagger.GhaJobOpts{}},
-			{"LLM", []string{"TestLLM"}, &dagger.GhaJobOpts{}},
-			{"cli-engine", []string{"TestCLI", "TestEngine"}, &dagger.GhaJobOpts{}},
-			{"client-generator", []string{"TestClientGenerator"}, &dagger.GhaJobOpts{}},
-			{"interface", []string{"TestInterface"}, &dagger.GhaJobOpts{}},
-			{"call-and-shell", []string{"TestCall", "TestShell", "TestDaggerCMD"}, &dagger.GhaJobOpts{}},
+			{"container", []string{"TestContainer"}, nil},
+			{"LLM", []string{"TestLLM"}, nil},
+			{"cli-engine", []string{"TestCLI", "TestEngine"}, nil},
+			{"client-generator", []string{"TestClientGenerator"}, nil},
+			{"interface", []string{"TestInterface"}, nil},
+			{"call-and-shell", []string{"TestCall", "TestShell", "TestDaggerCMD"}, nil},
 			{"everything-else", nil, &dagger.GhaJobOpts{
 				Runner: AltPlatinumRunner(),
 			}},
@@ -239,12 +239,16 @@ func splitTests(runner *dagger.Gha, name string, dev bool, splits []testSplit) d
 			}
 			doneTests = append(doneTests, split.tests...)
 
-			opts := *split.opts
+			opts := dagger.GhaJobOpts{}
+			if split.opts != nil {
+				opts = *split.opts
+			}
 			opts.TimeoutMinutes = 30
 			opts.UploadLogs = true
 			if dev {
 				opts.DaggerVersion = "."
 			}
+
 			w = w.WithJob(runner.Job(name+split.name, command, opts))
 		}
 		return w
```

PR #10505, opened 2025-05-29
add support for using default engine GC policy during manual prune

sipsma:
Previously the core API for manually pruning the engine local cache only supported pruning *all* entries. This adds an optional arg to enable using the engine-wide default GC policy during these prunes.

The use case we have in mind is one where the automatic GC is disabled and instead triggered manually by users (e.g. at a time of low-traffic such that GC overhead is has minimal interference).

Ideally, the core API for managing the local cache would support the full breadth of pruning options as is possible in the engine config, but that's a bigger effort; the option added here is still an improvement in the meantime.


.changes/unreleased/Added-20250529-130429.yaml
```diff
@@ -0,0 +1,12 @@
+kind: Added
+body: |
+  Add support for using engine-wide default GC policy when triggering manual local cache prunes.
+
+  The manual prune API now supports an optional arg to enable honoring the engine-wide default automatic GC configuration. e.g.
```
vikram-dagger, line 5: Should we add this to https://docs.dagger.io/configuration/cache?

vikram-dagger:
Looks like a CI error related to the PHP reference:

`process "diff -r a/docs/static/reference/php b/docs/static/reference/php" did not complete successfully: exit code: 1`

.changes/unreleased/Added-20250529-130429.yaml
sipsma: yes good idea, added!

PR #10512, opened 2025-05-30
tui: reduce error verbosity, support go-to-error

vito:
(TODO: based on #10468 - needs rebase after merge)

Context: https://github.com/dagger/dagger/discussions/10461

(TODO: better description, prioritizing getting #10468 mergeable first)

When returning errors, capture the current trace context and store it in error attributes


dagql/idtui/testdata/TestTelemetry/TestGolden/fail-effect
```diff
@@ -14,17 +14,17 @@ Expected stderr:
 ● parsing command line arguments X.Xs
 
 ● viztest: Viztest! X.Xs
-▼ .failEffect: Container! X.Xs
+▼ .failEffect: Container! X.Xs ERROR
 ! process "sh -c echo this is a failing effect; exit 1" did not complete successfully: exit code: 1
```
vito, line 18: Ideally this would be deduped too, but it's pretty tricky. Going to leave it for another iteration.

The trickiness is in the fact that the "error" status here is actually coming from an effect. The `failEffect` function actually returns successfully, but the TUI bubbles up an error in its returned effects to any call sites that returned the failed effect. (`s/effect/LLB vertex/` for those deep in Buildkit)

The hard part is deduping this without _also_ deduping it from the `Container.withExec` call where we _want_ to see the error, since that works via the same logic.

sipsma (approved):
👨‍🎨

PR #10523, opened 2025-06-03
feat: support Claude oauth authentication

grouville:
A lot of users have a Claude Max subscription. It would be nice if they could use that plan when using Dagger + Anthropic


grouville:
closing as it is unclear if it is actually what Anthropic allows

Issue #10515, opened 2025-06-01
🐞 elixir: issue with the ex_doc


wingyplus:
### What is the issue?

1. Link to is incorrect. For example, the experimental_with_gpu is pointed to link https://github.com/dagger/dagger/blob/v0.18.9/lib/dagger/gen/container.ex#L223 which should pointed to `sdk/elixir/lib/dagger/gen/container.ex`.
2. The experimental doesn’t highlight as warning . For example, https://hexdocs.pm/dagger/Dagger.Container.html#experimental_with_gpu/2

### Dagger version

dagger v0.18.9

Nero-F:
I think I can tackle this one

wingyplus:
The 2 is correct because the experimental directive doesn’t merge into main yet.

Fixing only source url for this issue would be enough.

wingyplus:
@Nero-F thank you. Assigned to you. :)

PR #10524, opened 2025-06-03
fix(sdk/elixir): exdoc source_url now pointing to correct link

Nero-F:
ex_doc uses `:source_url` and `:source_ref` for deducing source links but it was not relative to the `sdk/elixir` folder
using `:source_url_pattern` seems to solve this issue #10515


wingyplus (approved):
Thank you for fixing this @Nero-F ❤️

Issue #7707, opened 2024-06-19
Unbundle the SDK: client without runtime, runtime without client


shykes:
## Problem

A Dagger SDK can do 2 things:

1. Generate client bindings to consume functions and types from other modules
2. Build a runtime to provide new functions and types

These two features are tightly coupled: an SDK must implement both or it is considered incomplete. A module is expected to use both, or it is probably "doing it wrong".

But there are cases where I don't actually want both:

- Example 1: I want to run Dagger pipelines from my software, without running my software in a Dagger runtime. For this, I need generated client bindings, but no runtime. This is not possible.

- Example 2: I want to expose my Dockerfile and docker-compose.yaml as a Dagger module. I could manually rewrite my docker-compose and Dockerfiles to code, but I would lose compatibility with my team's workflow. I could use a generic "Docker" module with functions like `service(name: String!): DockerComposeService` and `type DockerComposeService { up: Service! }`. But that's a generic schema, it doesn't express the specifics of my compose file with custom types. Instead I want `backend(db: ServiceID!, source: Directory!, stripeAPIToken Secret!)`, infered from the contents of my file. **This requires a custom runtime**, but it does not require generating a client. This is not possible today.

## Solution

Decouple the two big features of a Dagger SDK - client and runtime - so that one can be provided, and used, without the other.

## Design

Consider this a request for bikeshed. Some considerations:

- `dagger develop` combines many SDK features in one command. We have already discussed the idea of unbundling it. Maybe that would help.
- The concept of a "runtime" is mostly hidden from the user. Should we embrace it, and show it to the user as a first-class concept?
- Perhaps each feature of an SDK can be explained as a pipeline? One pipeline to generate client bindings; another pipeline to build a runtime; and SDKs may or may not implement all pipelines?
- In the second example above ("generate functions from my docker project"), would we talk about a "Docker SDK"? If not, then what terminology would be used?

helderco:
Just to be clear on "it's not possible today", every SDK's codegen is already decoupled from the runtime module actually. You just need to get the corresponding executable outside of the dagger cli (basically need a clone of dagger's repo and possibly build the codegen binary, but it can be comoditized in a module, even using an SDK's runtime module as a dependency for most of the work).

However the runtime module does provide a sort of unified and simple way to access that feature, but returns more than what you want if just need the client bindings.

SDK runtime modules currently need to implement two functions only (by convention):
- `Codegen()`
- `ModuleRuntime()`

[...]

PR #10525, opened 2025-06-03
feat: unbundle sdk method call to allow partial implementation

TomChv:
Fixes https://github.com/dagger/dagger/issues/7707

This PR applies https://github.com/dagger/dagger/issues/7707#issuecomment-2912150141 with few changes to make it simpler:
- Rely on returned error type to ignore if a SDK doesn't implement a function
- Add tests for SDK with only runtime or only codegen

:bulb: **Quick note on my implementation**

Instead of defining a new field in `asModule` like I mentioned in https://github.com/dagger/dagger/issues/7707#issuecomment-2912150141, I thought it was much simpler to just check the error type to know what happens and ignore it if that make sense.
That way we do not have API changes, it's all inside the engine logic that become "*smarter*" on its interaction with the SDK.

I also improved the error messages displayed to the user in case of missing implementation so it's simpler both for us and the user to understand what's happening on failure.


core/schema/modulesource.go
```diff
@@ -1808,6 +1810,8 @@ func (s *moduleSourceSchema) runCodegen(
 	return genDirInst, nil
 }
 
+var ErrSDKClientGeneratorNotImplemented = errors.New("the SDK does not implement the `ClientGenerator` interface")
```
jedevc: nit: group all the error types at the top of the file.

core/schema/modulesource.go
```diff
@@ -1977,7 +1982,7 @@ func (s *moduleSourceSchema) moduleSourceGeneratedContextDirectory(
 func (s *moduleSourceSchema) runModuleDefInSDK(ctx context.Context, src, srcInstContentHashed dagql.Instance[*core.ModuleSource], mod *core.Module) (*core.Module, error) {
 	runtimeImpl, ok := src.Self.SDKImpl.AsRuntime()
 	if !ok {
-		return nil, fmt.Errorf("sdk implementation does not implement Runtime interface")
+		return nil, fmt.Errorf("the SDK does not implement the `Runtime` interface")
```
jedevc: We can create an error type for this as well, even though we don't handle it today.

It may be useful to handle it in the future.

.changes/unreleased/Added-20250603-162638.yaml
```diff
@@ -0,0 +1,42 @@
+kind: Added
+body: |-
+  Unbundle the SDK interface to support partial implementation.
+
+  When you implement a custom SDK, you can now choose to just implement part of the `SDK` interface depending on your needs.
+  That means:
+  - `dagger develop` no longer fails if your SDK just implement the `Runtime` interface, it will simply not call `Codegen` if not defined.
+  - `dagger call` and `dagger functions` give a clear error message if it's not supported by the SDK.
```
jedevc, line 8: I don't think the changelog needs to be this long - we can just say that partial SDKs are allowed, and explain what that means. 

For a full example, users can look at this PR - it's not a normal user feature, it's definitely for power-users.

core/integration/module_test.go
```diff
@@ -2286,6 +2286,192 @@ type Test struct {}
...
+			generatedFile, err := ctr.With(daggerExec("develop")).File("/work/hello.txt").Contents(ctx)
+
+			require.NoError(t, err)
+			require.Equal(t, "Hello, world!", generatedFile)
+		})
+
+		t.Run("explicit error on dagger call", func(ctx context.Context, t *testctx.T) {
+			_, err := ctr.With(daggerExec("call", "foo")).Sync(ctx)
+
+			requireErrOut(t, err, "the SDK does not implement the `Runtime` interface")
```
jedevc: I think the error message might be better as `"only-codegen" sdk does not support a runtime`.

core/integration/module_test.go
```diff
@@ -2286,6 +2286,192 @@ type Test struct {}
...
+    WithDirectory("/src", m.Src).
+    WithWorkdir("/src").
+    WithExec([]string{"go", "build", "-o", "/bin/sdk", "."}).
+    WithEntrypoint([]string{"/bin/sdk"}), nil
+}
+`, golangImage)).
+			WithWorkdir("/work/sdk/src").
+			WithExec([]string{"go", "mod", "init", "dagger.test/sdk"}).
+			WithExec([]string{"go", "get", "dagger.io/dagger@v0.18.9"}).
+			WithNewFile("/work/sdk/src/main.go", `package main
```
jedevc: I think it makes sense to move these little custom SDKs out into a dedicated go file, and `embed` it.

core/integration/module_test.go
TomChv: Yeah but how can I compare the error if I make it dynamic?

```go
		if err != nil && !errors.Is(err, ErrSDKCodegenNotImplemented) {
			return genDirInst, fmt.Errorf("failed to run codegen: %w", err)
		}
```

core/integration/module_test.go
TomChv: I'm not sure I can, it will fails on the lint because the module imports dagger types that needs to be generated? 
Is there a way to embed and copy the whole directory inside the engine?

TomChv: nvm, I could use `c.Host()`

jedevc: `testdata` isn't linted.

core/integration/module_test.go
jedevc: You can implement a custom error type.

```go
type ErrSDKCodegenNotImplemented struct {
	SDK string
}

func (err ErrSDKCodegenNotImplemented) Error() string {
	// ...
}
```

core/integration/module_test.go
TomChv: I see, I added them as whole module

core/integration/module_test.go
TomChv: Done

.changes/unreleased/Added-20250603-162638.yaml
TomChv: Done

core/schema/modulesource.go
TomChv: Done

core/schema/modulesource.go
TomChv: Done

TomChv:
Some tests are failing but I'm not able to repro the failure locally:

```
dagger --cloud -m .dagger call test specific --race=true --parallel=16 --run="TestTelemetry|TestProvision"
✔ connect 4.6s
✔ load module 10.8s

✔ daggerDev(
│ │ source: Host.directory(exclude: ["bin", ".git", "**/node_modules", "**/.venv", "**/__pycache__"], noCache: true, path: "/Users/tomchauveau/Documents/github.com/dagger/dagger"): Directory!
│ ): DaggerDev! 9.6s
✔ .test: DaggerDevTest! 0.4s
∅ .specific(parallel: 16, race: true, run: "TestTelemetry|TestProvision"): Void 3m22s
│ ✔ TestTelemetry 4.8s
│ ∅ TestProvision 32.2s
│ ∅ TestTelemetry 2m56s
```

core/integration/module_test.go
```diff
@@ -2286,6 +2286,74 @@ type Test struct {}
...
+// cc: https://github.com/dagger/dagger/issues/7707
+func (ModuleSuite) TestUnbundleSDK(ctx context.Context, t *testctx.T) {
+	t.Run("only codegen", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		ctr := c.Container().From(golangImage).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work/sdk").
+			With(daggerExec("init", "--name=only-codegen", "--sdk=go", "--source=.")).
+			WithDirectory("/work/sdk", c.Host().Directory("./testdata/sdks/only-codegen")).
```
helderco: Why both `dagger init` and copying the already inited module with `c.Host`?

core/integration/module_test.go
```diff
@@ -2286,6 +2286,74 @@ type Test struct {}
...
+			generatedFile, err := ctr.With(daggerExec("develop")).File("/work/hello.txt").Contents(ctx)
+
+			require.NoError(t, err)
+			require.Equal(t, "Hello, world!", generatedFile)
+		})
+
+		t.Run("explicit error on dagger call", func(ctx context.Context, t *testctx.T) {
+			_, err := ctr.With(daggerExec("call", "foo")).Sync(ctx)
+
+			requireErrOut(t, err, "'./sdk' SDK does not support a runtime")
```
helderco: Maybe this would be clearer:
```suggestion
			requireErrOut(t, err, "'./sdk' SDK does not support defining and executing functions")
```
Reasoning is because the word "runtime" is known by SDK maintainers when they start learning how to add modules support, but not necessarily easily understood by others. The SDK could be used just for client generation for example.

core/integration/testdata/sdks/only-runtime/src/main.go
```diff
@@ -0,0 +1,68 @@
...
+	"dagger.io/dagger/dag"
+)
+
+func main() {
+	ctx := context.Background()
+	defer dag.Close()
+
+	name, err := dag.CurrentModule().Name(ctx)
+	if err != nil {
+		fmt.Println(fmt.Errorf("failed to module name: %w", err))
```
helderco:
```suggestion
		fmt.Println(fmt.Errorf("failed to get module name: %w", err))
```

core/schema/modulesource.go
```diff
@@ -29,6 +29,30 @@ import (
...
+func (err ErrSDKRuntimeNotImplemented) Error() string {
+	return fmt.Sprintf("'%s' SDK does not support a runtime", err.SDK)
+}
+
+type ErrSDKCodegenNotImplemented struct {
+	SDK string
+}
+
+func (err ErrSDKCodegenNotImplemented) Error() string {
+	return fmt.Sprintf("'%s' SDK does not support code generation", err.SDK)
```
helderco: Generating client bindings has always been the true "codegen" in my book. Modules could generate files that aren't strictly code.
```suggestion
	return fmt.Sprintf("%q SDK does not support module generation", err.SDK)
```

core/schema/modulesource.go
```diff
@@ -29,6 +29,30 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+type ErrSDKRuntimeNotImplemented struct {
+	SDK string
+}
+
+func (err ErrSDKRuntimeNotImplemented) Error() string {
+	return fmt.Sprintf("'%s' SDK does not support a runtime", err.SDK)
```
helderco:
```suggestion
	return fmt.Sprintf("%q SDK does not support a runtime", err.SDK)
```

core/schema/modulesource.go
```diff
@@ -1858,18 +1882,22 @@ func (s *moduleSourceSchema) runClientGenerator(
 		return genDirInst, fmt.Errorf("failed to load dependencies of this modules: %w", err)
 	}
 
-	// If the current module source has sources, we can transform it into a module
-	// to generate self bindings.
+	// If the current module source has sources and its SDK implements the `Runtime` interface,
+	// we can transform it into a moduleto generate self bindings.
```
helderco:
```suggestion
	// we can transform it into a module to generate self bindings.
```

core/schema/modulesource.go
```diff
@@ -1858,18 +1882,22 @@ func (s *moduleSourceSchema) runClientGenerator(
...
 	if srcInst.Self.SDK != nil {
-		var mod dagql.Instance[*core.Module]
-		err = s.dag.Select(ctx, srcInst, &mod, dagql.Selector{
-			Field: "asModule",
-		})
-		if err != nil {
-			return genDirInst, fmt.Errorf("failed to transform module source into module: %w", err)
-		}
+		// We must make that if inside the condition to avoid checking a nil pointer on `SDKImpl`
+		// and make the engine crash.
```
helderco:
```suggestion
		// We must make sure that if inside the condition to avoid checking a nil pointer on `SDKImpl`
		// and make the engine crash.
```
The general wording of this could be clearer though.

TomChv:
@helderco All comments are addressed, waiting for an approval to merge :)

core/schema/modulesource.go
TomChv: Done in [`8578e58` (#10525)](https://github.com/dagger/dagger/pull/10525/commits/8578e58c014beaba9e01f6ddade8c8fa8e18c9b2)

core/schema/modulesource.go
TomChv: Done https://github.com/dagger/dagger/pull/10525/commits/8578e58c014beaba9e01f6ddade8c8fa8e18c9b2

core/schema/modulesource.go
TomChv: Done https://github.com/dagger/dagger/pull/10525/commits/8578e58c014beaba9e01f6ddade8c8fa8e18c9b2

core/schema/modulesource.go
TomChv: Done https://github.com/dagger/dagger/pull/10525/commits/8578e58c014beaba9e01f6ddade8c8fa8e18c9b2

core/integration/testdata/sdks/only-runtime/src/main.go
TomChv: Done https://github.com/dagger/dagger/pull/10525/commits/8578e58c014beaba9e01f6ddade8c8fa8e18c9b2

core/integration/module_test.go
TomChv: Done https://github.com/dagger/dagger/pull/10525/commits/8578e58c014beaba9e01f6ddade8c8fa8e18c9b2

core/integration/module_test.go
TomChv: Done https://github.com/dagger/dagger/pull/10525/commits/8578e58c014beaba9e01f6ddade8c8fa8e18c9b2

PR #10537, opened 2025-06-05
fix: ts telemetry tests

TomChv:


TomChv:
Closed since it's a duplicate of #10535

PR #10540, opened 2025-06-05
make current Query object available via context

sipsma:
Spinning out from https://github.com/dagger/dagger/pull/10497 to double check it doesn't (somehow) break anything in isolation and to rip the bandaid on a huge source of rebase conflicts earlier.

---

Putting `Query` as a field in structs is a surprisingly significant source of pain when it comes to serializing/deserializing objects into/from the DB, mainly because the standard json marshal/unmarshal interfaces don't accept a context and thus don't accept smuggling.

I dealt with it for a while but hitting more annoying cases (specifically when it's hidden behind objects wrapped in interfaces, i.e. `GitRepository`), so just want to squash it.

This whole problem is also unnecessary since now-a-days everything in query is based on the context anyways, so we can just plumb `Query` itself around through context for now.

There's probably even more detailed refactors possible here, but went with the approach here since it seemed to be easier to have AI do (though this still involved a somewhat painful amount of human hand-editing afterwards).

---

Draft while I check a full integ suite test run in CI, subset passed locally


vito (approved):
Makes sense! Always seemed like the winds were blowing this way 💨

Issue #10390, opened 2025-05-13
`failed to stat` local path, `no such file or directory`


jedevc:
I've seen an increase in the number of flakes that look something like (both in CI, locally, and in the middle of test runs):

```
Error: failed to get configured module: input: moduleSource failed to resolve dep to source: failed to load local dep: select: failed to load local module source context: select: failed to get content hash: failed to get content hash: failed to get snapshot: failed to sync: failed to copy "/home/runner/work/dagger/dagger": failed to stat /var/lib/dagger/worker/snapshots/snapshots/4/fs/home/runner/work/dagger/dagger/sdk/typescript/runtime/bundled_static_export/module/telemetry.ts: lstat /var/lib/dagger/worker/snapshots/snapshots/4/fs/home/runner/work/dagger/dagger/sdk/typescript/runtime/bundled_static_export/module/telemetry.ts: no such file or directory
```
([source](https://v3.dagger.cloud/dagger/traces/aa005d2ecc7b4acfe9af3d2e2dca2042#fccc414c2c83710f:L1))

This file *should* exist, so something has likely gone wrong.

I suspect this is something to do with [our implementation of the `Local` source](https://github.com/dagger/dagger/blob/main/engine/sources/local/source.go), but no real way to know for sure right now.

jedevc:
I just saw this locally 👀 https://v3.dagger.cloud/dagger/traces/2bc329b88cc59afdda328d47963b830a

jedevc:
It seems like I can hit this *pretty* reliably using `dagger develop --recursive .` in the base of the dagger repo.

sipsma:
I didn't hit it in a few attempts of `dagger develop --recursive .`, but did randomly hit it with just `./hack/dev` locally: https://v3.dagger.cloud/dagger/traces/2c794daf24b3895d92566cf4367860b3?listen=c17083e438361936&listen=4ef4ecfb9b50b273&listen=9c50b6beb1f8a3d3&listen=c6d545d4ad5c67c5&listen=8866d4774238e821&listen=297dddc8f9305f87

We haven't changed anything meaningful in the local source implementation in a while, but we have changed the `cache` code which is also used by the local source impl to manage concurrent syncs. Possible there's an issue in that cache code or maybe timing there has just been tweaked in such a way that a problem w/ the local source code is becoming more obvious

die-tech:
Just curious 🤔  how to re-produce this ? 
[...]

PR #10541, opened 2025-06-05
revert local source synchronization handling back to simpler singleflight implementation

sipsma:
We unified the singleflight implementation in local source w/ the spiritually-similar cache used by dagql a while back. This wasn't strictly necessary other than attempting to re-use code where it felt reasonable to, reduce maintainance burden, etc.

[Recently, we started frequently seeing strange flaky errors during filesyncs about missing files](https://github.com/dagger/dagger/issues/10390), which have proved hard to repro w/ any consistency locally, which makes them hard to debug.

The local filesync implementation didn't change much during this time, but the cache implementation it shares with dagql _did_.

This change reverts the local source sync back to its earlier/simpler singleflight implementation in order to see if the errors go away.

If they do, that's great and we should probably just stick with this simpler, more-specialized implementation. But that would also mean we should look to see if there's any synchronization problems that might be subtly affecting the dagql cache implementation.

And even if the errors are still happening after this change is released, that's still a useful data point if nothing else.


sipsma:
Ah well, hit it first try here in the dev-engine specifically (not stable released one), so I guess this wasn't it. I'll move to draft and see if I can do any debugging in CI since it seems relatively speaking very easy to hit here (vs locally, where I've seen it like twice ever)

sipsma:
Might have figured out the problem, described in the related issue: https://github.com/dagger/dagger/issues/10390#issuecomment-2950884152

Will re-purpose this PR w/ a fix

sipsma:
Failure here after the fix: https://github.com/dagger/dagger/actions/runs/15501092730/job/43648752253?pr=10541

*But* that one is in the outer stable engine, so expected to still be possible here (until this PR is merged and released and our CI is upgraded). Will re-run some more to ensure it doesn't pop up in the inner dev engine ever.

sipsma:
> I'm actually reasonably sure this shouldn't be the case 🤔

@jedevc I believe you that it *shouldn't* do that, it's definitely wasteful, but I added very simple debug logs locally to filesync that printed each time `copier.copy` was called ([this line](https://github.com/sipsma/dagger/blob/8f09292ec963060a28570ce7fbaefb392feea8af/engine/sources/local/copy/copy.go#L301-L301)) and could see for a sync made during module loads of the dagger repo like `include: ["version/dagger.json", "version/**/*"]` that it was hitting unrelated files in dirs like `sdk/typescript/`.

So it's definitely doing it for some reason. In fairness it would still skip the actual copy of the file, the problem was just that it assumed the file still existed and errored out if it disappeared, which could happen in race condition scenarios. Wonder if the `**/*` glob style is confusing it somehow?

PR #10552, opened 2025-06-09
ci: bump spectaql

jedevc:
Fixes the currently failing docs job on `main`. And also, just good to keep this up-to-date.


jpadams (approved):
Thank you! Hopefully eveything else can handle Node 22 as well, but seems likely from the tests passing.
As an additional smoke test, I ran `dagger -c 'server | up'` from the `docs` directory as well, since that's an important dev flow for us. All looks good! 👍

PR #10555, opened 2025-06-09
add support for prometheus metrics listener in engine

sipsma:
Starting out with just a very simple Prometheus gauge for the number of connected clients. This is meant to support some cloud-related use cases. 

Current usage:
* ATM you have to set an env var on the engine container to configure the listener
   * Example: `DAGGER_METRICS_ADDR: 0.0.0.0:9090`
   * Will add config file + command line support before merging once we confirm we want to go with this approach
* After that prometheus metrics are available over HTTP at `/metrics`
   * Example: `curl http://<engine-endpoint>:9090/metrics`
* The metric added here is named `dagger_connected_clients`
   * Example line from output: `dagger_connected_clients 1`

---

@gerhard One thing that occurred to me while writing this is that we could try to instead use OTEL metrics, which is appealing in that it could unify more with the rest of our telemetry stack
* We do already support OTEL metrics and exporting those to cloud, but they are "per-client" and integrated with the rest of our per-client OTEL spans/logs.
   * For this new use case, we'd want to export metrics for the engine "as a whole"
* This would be more of a "push" model, whereas exposing a Prometheus endpoint on the engine is more "pull"
* Potential pros of this approach:
   1. Integrates more with the rest of our existing cloud OTEL stack (persistence of data for each engine in the backend db, etc.)
   2. We could plausibly use the engine pushing these metrics as a healthcheck (if it's not pushing engine-wide metrics, something probably went wrong)
   3. The "push" model may be simpler in terms of networking; i.e. it's often easier to have the engine reach out to the cloud than it is for us to reach the engine in arbitrary networking environments (which isn't a problem today but may become more of a headache in the longer term)
* Potential cons
   1. More complicated to get going

Given it's extreme simplicity, I'm also totally fine starting w/ this Prometheus listener approach to get us going and then iterate from there, possibly moving to OTEL metrics down the line. Just wanted to raise this now.
* [There's also integration between Prometheus and OTEL](https://opentelemetry.io/blog/2024/prom-and-otel/), though I haven't taken the time to read through the details there yet so no opinion on whether it's a path worth considering

---

Draft TODOs:
- [ ] Figure out above question of OTEL vs. Prometheus (will discuss with Gerhard)
- [ ] Support configuration of metrics endpoint as engine config file setting + command line param (rather than just an env var)
- [ ] Clarify if the `/healthz` endpoint is needed for Prometheus
   - Windsurf added this for me, have not yet looked into whether it's really needed


gerhard:
So good to see this come together @sipsma!

> One thing that occurred to me while writing this is that we could try to instead use OTEL metrics, which is appealing in that it could unify more with the rest of our telemetry stack.

Using OTel metrics is a great idea. Yes, we should absolutely do that **in addition to** exposing this Prometheus `/metrics` endpoint.

There is complexity associated with depending on the `/metrics` interface even if we build the Grafana dashboard or share some Prometheus alerts. Wiring everything together, keeping up with the updates, combining with logs, mapping it to operations that run within the Engine, etc. is just too much work. Enthusiasts are free to do it, but anyone that wants the polished end-result which is being improved on constantly, will opt for Dagger Cloud. Another option is to [plug the OTel data into an LLM](https://www.honeycomb.io/blog/tales-from-the-trench-building-with-llms-honeycomb), but that brings its own set of complexities.

On top of the above, we both know that there is a lot more to this than what can be exposed via  `/metrics`. OTel allows us to attach a lot more contextual meaning, which makes the data a lot more insightful. In conclusion: OTel as a more detailed & insightful addition **on top of** this new & experimental `/metrics` endpoint makes the most sense to me.

Since I started typing this, the PR moved along a bit, so I'm going to do one final review before we merge & ship it, preferably in the next release. cc @jedevc

PR #10562, opened 2025-06-10
evals: refactor into a repeatable pattern

vito:
The goal: establish a repeatable pattern for evals, decoupling them from Dagger's own evals.

Evals are now implemented as separate types that implement this interface:

```go
type Eval interface {
  Name(context.Context) (string, error)
  Prompt(base *dagger.LLM) *dagger.LLM
  Check(ctx context.Context, prompt *dagger.LLM) error

  DaggerObject
}
```

Other things done along the way:

* [x] Use extracted https://github.com/vito/runt helper, to replace the `evalT` boilerplate. (This isn't essential to the pattern, I just like to be able to use assertion packages.)
* [x] Fix returning an LLM from a module that depends on types from that module.
  * Previously all of the LLM unlaziers (`LastReply`, `Sync`, `Loop`, `ToolsDoc`, `History` etc. etc.) would evaluate the prompt in the caller's context. This stopped working because now evals return an `LLM` from `Prompt`, and that `LLM` was being evaluated in the Evaluator module context.
  * Thankfully we already had an eval covering this scenario (`ModuleDependencies`) - but it had been passing up until now because the evals were still ultimately being evaluated within the same module. Now we test that not only does the LLM see the module dependencies, it sees them after the `LLM` type crosses module boundaries. :tada:
* [x] Emit user prompts when evaluating the LLM, rather than when the prompt is installed (`LLM.withPrompt`).
* [x] Fix `LLM.attempt` so that it also works when appended to a prompt, not prepended.
* [x] Ensure we show telemetry for tool/API calls even if they have already occurred within the session


cmd/dagger/module_inspect.go
```diff
@@ -68,7 +68,7 @@ func initializeModule(
 	}
 
 	serveCtx, serveSpan := Tracer().Start(ctx, "initializing module", telemetry.Encapsulate())
-	err = modSrc.AsModule().Serve(serveCtx, dagger.ModuleServeOpts{IncludeDependencies: true})
```
vito: @cwlbraa @helderco Looking to double check this - without this fix, I ran into errors while trying to run this in shell, since there are two `Workspace` types:

* https://dagger.cloud/dagger/traces/71e3bd754197c0d17bc26a092e688a63?listen=ae52ee3b8632482b&listen=e9d1a95c779e20a7&span=9bab1abb165eac4a

With this fix applied it works fine:

* https://dagger.cloud/dagger/traces/09970837d24c661f4a23adeec263298c?span=a2257556816ecafd

This one _seems_ redundant - we still pass `IncludeDependencies: true` for the primary module (that's how I'm able to call `evaluator` above), just not here.

vito: splitting this out into https://github.com/dagger/dagger/pull/10581

core/llm.go
```diff
@@ -73,13 +73,17 @@ type LLM struct {
 	err  error
 
 	// History of messages
-	messages []ModelMessage
+	messages []*ModelMessage
```
vito, line 76: Context: this was a way-off shot in the dark at the OOMing issue encountered in #10562, but I think it makes sense to keep anyway. This slice gets copied here and there, so might as well avoiding copying the full message history.

core/llm.go
```diff
@@ -766,6 +766,29 @@ func (llm *LLM) loop(ctx context.Context, dag *dagql.Server) error {
...
+					attribute.String(telemetry.UIMessageAttr, "sent"),
+				))
+				defer span.End()
+				stdio := telemetry.SpanStdio(ctx, InstrumentationLibrary,
+					log.String(telemetry.ContentTypeAttr, "text/markdown"))
+				defer stdio.Close()
+				fmt.Fprint(stdio.Stdout, msg.Content)
+			}()
+		}
+
```
vito, line 778: This is a bit more work but I think makes more sense than showing the user prompt span at `WithPrompt` time, where it's not actually being sent to the LLM yet.

core/schema/llm.go
```diff
@@ -173,8 +173,9 @@ func (s *llmSchema) loop(ctx context.Context, llm *core.LLM, args struct{}) (*co
 func (s *llmSchema) attempt(_ context.Context, llm *core.LLM, _ struct {
 	Number int
 }) (*core.LLM, error) {
-	// nothing to do; we've "forked" it by nature of changing the query
-	return llm, nil
+	// clone the LLM object, since it updates in-place when Sync is called, and we
+	// want to branch off from here
+	return llm.Clone(), nil
```
vito, line 178: This is a subtle fix to make this consistent with the recent refactor to how `sync` works: it updates in-place to ensure all unlaziers (`tokenUsage`, `lastReply`, etc) are called against the same result.

As-is, if you append `Attempt(1..10)` and then call `LastReply` it'll actually only evaluate once. So the fix is to just clone it.

Seems like kind of a pain to test. :thinking: Open to ideas. I noticed this by looking through the evals traces. It doesn't affect evals on `main` because they call `Attempt` before appending their prompt.

modules/evaluator/main.go
```diff
@@ -33,28 +36,45 @@ func New(
...
 	}
-	if m.InitialPrompt != nil {
-		env = env.WithFileInput("initialSystemPrompt", m.InitialPrompt,
-			"An initial system prompt to evaluate and improve.")
+	// FIXME: it would be nice to not have to do this workaround. it's hard
+	// because we want to accept Eval, but then the type has no AsWorkspaceEval().
+	//
+	// fortunately the IDs are the same nonetheless, so we can just convert it
+	// with the available plumbing
+	m.Evals = append(m.Evals, dag.LoadWorkspaceEvalFromID(dagger.WorkspaceEvalID(id)))
```
vito, line 100: Ran into some friction with interfaces here, but the workaround ain't too bad. cc @sipsma 

The problem is kind of tricky: there's an `Eval` interface in the `evaluator/workspace/` module, and it's the same interface defined locally here, but I have to convert it from my interface to the `WorkspaceEval` version.

It would be nice if this arg took a Go interface. :thinking:

PR #10566, opened 2025-06-11
fix: use the current id view for version compat decisions

jedevc:
Importantly! It's *never* correct to use this `View` to determine how to do compat. The view there is used as the default for graphql queries, but everything else uses the `ID.View` or the `Selector.View`.

To understand why, imagine the following scenario:
- In server A (using version A), we call a field and get back an ID which has `ID.View = A`
- Later on, in server B (using version B), we attempt to load this ID.

While we make sure that we select the field+inputs correctly according to view `A` in this case, *if the result is not cached and we need to recompute it*, the actual implementation will see a server that is using version `B`. Previously, it seems like we may have been getting away with this, just because of caching.

However, when I started adding versioning to the typedefs API in #9518, it became quite easy to hit this scenario.

I've just gone through and fixed all the references to the server view that I can find, and have updated the comment there to indicate it shouldn't be used for this (it still needs to be public so it can be manipulated as part of coremod shenanigans).

### DIFF
--- a/core/directory.go
+++ b/core/directory.go
@@ -25,7 +25,6 @@ import (
 
 	"dagger.io/dagger/telemetry"
 	"github.com/dagger/dagger/dagql"
-	"github.com/dagger/dagger/engine"
 	"github.com/dagger/dagger/engine/buildkit"
 )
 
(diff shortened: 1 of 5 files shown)


jedevc:
Putting this into a milestone since I guess technically this is a bug fix of sorts? Really have no idea how to trigger this effectively in a test, since it requires *bizarre* cache shenanigans.

PR #10567, opened 2025-06-11
add caller and callee module metadata to dag.call spans

marcosnils:
this commit enriches the current dag.call spans with the metadata from
both the caller and the callee modules. It enables tracing function
calls across modules.


core/telemetry.go
```diff
@@ -67,6 +67,52 @@ func AroundFunc(
...
+				attribute.String(telemetry.ModuleSubpathAttr, m.GetSource().SourceRootSubpath),
+				attribute.String(telemetry.ModuleNameAttr, m.Name()),
+			)
+
+			if m.GetSource().Local != nil {
+				attrs = append(attrs,
+					attribute.String(telemetry.ModuleContextDirAttr, m.GetSource().Local.ContextDirectoryPath),
+				)
+			}
+		}
```
vito: Bikeshed: I wonder if we should consolidate this into just one attribute for the module ref. :thinking: Or is the granularity particularly helpful?

Right now the consumer has to put all this information together if it wants to include both local dir and git runs of a given module. I think we could turn those back into the same module ref, using either:

* `<GitRemote>[/<Subpath>]<@<GitRef>` for git source
* `<SimplifiedHTMLRepoURL>[/<Subpath>]@<Commit>` for local source

My guess is that we'd always want to treat these the same, because whether you ran from a source checkout or a Git ref doesn't mean much - you'd want to see `github.com/dagger/dagger` as a caller whether it's from our CI (which uses a local checkout) or from someone doing `dagger -m github.com/dagger/dagger shell`.

sdk/go/telemetry/attrs.go
```diff
@@ -7,6 +7,10 @@ const (
 	// represents.
 	DagCallAttr = "dagger.io/dag.call"
 
+	// The base64-encoded, protobuf-marshalled receiver callpbv1.Call that this span
+	// represents.
+	DagCallReceiverAttr = "dagger.io/dag.receiver"
+
```
vito:
```suggestion
```
this was from an earlier iteration i think - dead code?

vito (commented):
Sweet, glad this worked - the code is pretty straightforward, just had one bikeshed

core/telemetry.go
marcosnils: I mostly left them as separate attributes because some of them (like subpath, git.ref, etc) were already there. On top of that, I also thought that for consolidation, we'll have to move some of the logic that we do in our backend to put all the things in the right place which if needed to change, will take more time given the engine release cycles. 

That's mostly the reason I've preferred to send the data as-is from here and then handle the stitching in the BE :pray:

vito: > On top of that, I also thought that for consolidation, we'll have to move some of the logic that we do in our backend to put all the things in the right place which if needed to change, will take more time given the engine release cycles.

Sorry I don't follow this part.

My thinking behind the single module ref attribute is:

* Keeping knowledge in one place: if we expose module refs + metadata in this decomposed form, then both the engine and Cloud have to be aware of module ref syntax. If we just send a ref, only the engine needs to know this, and we can change the ref syntax more freely.
* Generally we've been trying to push logic into the engine. It even already kind of knows how to construct refs with [`ModuleSource.AsString`](https://github.com/dagger/dagger/blob/996d92a59922b2a75c74039b0779da46a154c51c/core/modulesource.go#L217), so you could _almost_ just call `ms.AsString()` and call it a day, except right now local sources just show their path. (I wonder if it'd be valueable for a `Directory` to be annotated with its Git metadata? :thinking:)
* It seems easier to aggregate in ClickHouse if there's just one attribute, no? (Assuming we want to treat from-local-dir and from-git-ref usage of a module as the same.)

To me that seems like more upside and more independence from the engine release cycles, but maybe I'm missing something with how y'all plan to use it.

vito (approved):

core/telemetry.go
```diff
@@ -93,6 +117,92 @@ func AroundFunc(
...
+	ref          string
+	version      string
+	functionName string
+}
+
+func parseCallerCalleeRefs(ctx context.Context, q *Query, call *callpbv1.Call) (*moduleCallRef, *moduleCallRef) {
+	cm, _ := q.MainClientCallerMetadata(ctx)
+	fc, _ := q.CurrentFunctionCall(ctx)
+	m, _ := q.CurrentModule(ctx)
+	sd, _ := q.CurrentServedDeps(ctx)
```
vito, line 139: Makes sense but it's unfortunate that we have to go down the `srv.clientFromContext` and `srv.clientFromIDs` path so many times, since each time involves holding a couple of locks open briefly. They're `RLocks`, so they won't compete with each other in concurrent API calls, but spamming `RLocks` can lead to starving the `Lock` and slow other sessions down. This seems like a somewhat hot path (every API call made from a module IIUC), so I'm being a little paranoid about it.

cc @sipsma in case he swoops in and says this is nothing to worry about :P

If we agree this is a problem, maybe the fix is to push this deeper into `Server` somehow, like a one-stop 'give me telemetry attributes' call.

dagql/idtui/golden_test.go
```diff
@@ -159,6 +167,29 @@ func (s TelemetrySuite) TestGolden(ctx context.Context, t *testctx.T) {
...
+						depCalled = true
+					} else if s.Name == "Viztest.traceFunctionCalls" {
+						require.Equal(t, "", s.ExtraAttributes[telemetry.ModuleCallerFunctionCallNameAttr])
+						require.Contains(t, s.ExtraAttributes[telemetry.ModuleCallerRefAttr], "")
+						require.Equal(t, "traceFunctionCalls", s.ExtraAttributes[telemetry.ModuleFunctionCallNameAttr])
+						require.Contains(t, s.ExtraAttributes[telemetry.ModuleRefAttr], "github.com/dagger/dagger/viztest")
+						rootCalled = true
+					}
+				}
+				require.True(t, rootCalled && depCalled)
```
vito: nit: might as well split these so we can tell which one was missing

```suggestion
				require.True(t, rootCalled)
				require.True(t, depCalled)
```

dagql/idtui/golden_test.go
```diff
@@ -64,6 +64,14 @@ func TestTelemetry(t *testing.T) {
...
 
 func (s TelemetrySuite) TestGolden(ctx context.Context, t *testctx.T) {
+	// setup a git repo so function call tests can pick up the right metadata
+	cmd := exec.Command("sh", "-c", "git init && git remote add origin git@github.com:dagger/dagger")
+	if co, err := cmd.CombinedOutput(); err != nil {
+		t.Fatalf("failed to initialize viztest git repo: %v: (%s)", err, co)
+	}
+	t.Cleanup(func() {
+		exec.Command("rm", "-rf", ".git").Run()
+	})
```
vito, line 74: Hmm, yeah I guess we can't really avoid this if we want to make sure the labels are set up properly. It would be nice to localize this to the test that needs it somehow, but not a blocker, maybe some other tests will want it in the future.

core/telemetry.go
```diff
@@ -93,6 +117,92 @@ func AroundFunc(
...
+
+	calleeRef.functionName = call.Field
+	calleeRef.version = call.Module.Pin
+	calleeRef.ref, _, _ = strings.Cut(call.Module.Ref, "@")
+
+	// if it doesn't have a pin, it's a local callee module
+	if calleeRef.version == "" {
+		if ms.Local != nil {
+			calleeRef.ref = strings.ReplaceAll(calleeRef.ref, ms.Local.ContextDirectoryPath, "")
+			if ms.Git != nil {
```
vito: Can these both be set? I thought it was one or the other. :thinking: 

If you're thinking of this as the case where you use a Git module that then depends on an in-repo ("local") module like `./foo`, I _think_ that gets normalized for us already into a Git ref made relative to the original ref, but I could be wrong. (I know we did that somewhere but I don't recall if that was in Daggerverse or in the engine.)

dagql/idtui/golden_test.go
```diff
@@ -159,6 +167,29 @@ func (s TelemetrySuite) TestGolden(ctx context.Context, t *testctx.T) {
...
+			Function: "trace-function-calls",
+			DBTest: func(t *testctx.T, db *dagui.DB) {
+				require.NotEmpty(t, db.Spans.Order)
+				var depCalled, rootCalled bool
+				for _, s := range db.Spans.Order {
+					if s.Name == "Dep.getFiles" {
+						require.Equal(t, "TraceFunctionCalls", s.ExtraAttributes[telemetry.ModuleCallerFunctionCallNameAttr])
+						require.Contains(t, s.ExtraAttributes[telemetry.ModuleCallerRefAttr], "github.com/dagger/dagger/viztest")
+						require.Equal(t, "getFiles", s.ExtraAttributes[telemetry.ModuleFunctionCallNameAttr])
+						require.Contains(t, s.ExtraAttributes[telemetry.ModuleRefAttr], "github.com/dagger/dagger/viztest")
```
vito: These are both the same value so the test isn't covering that we have these the right way around. One of these should probably be `.../dep` right? Would be nice to avoid `require.Contains` so we can do a more precise match.

dagql/idtui/golden_test.go
```diff
@@ -159,6 +167,29 @@ func (s TelemetrySuite) TestGolden(ctx context.Context, t *testctx.T) {
 		{Module: "./viztest/typescript", Function: "fail-log", Fail: true},
 		{Module: "./viztest/typescript", Function: "fail-effect", Fail: true},
 		{Module: "./viztest/typescript", Function: "fail-log-native", Fail: true},
+		{
+			Function: "trace-function-calls",
```
vito, line 172: Kind of a pain but we might need more test cases to cover all the branches above. Right now we're only testing a local module with a local dependency.

We could add a dependency on https://github.com/dagger/dagger-test-modules for that - looks like it already even has a function that calls a local dependency:

https://github.com/dagger/dagger-test-modules/blob/73670b0338c02cdd190f56b34c6e25066c7c8875/dagger/main.go

core/telemetry.go
```diff
@@ -93,6 +117,92 @@ func AroundFunc(
...
+		ms = m.GetSource()
+	}
+
+	if ms == nil {
+		return nil, nil
+	}
+
+	if fc != nil {
+		callerRef.functionName = fc.Name
+		if ms.Git != nil {
```
vito, line 161: Could maybe build on `ModuleSource.AsString` here or use its helper, `GitRefString`

core/telemetry.go
marcosnils: Take into account that this is called **only** for the spans that represent a module call as defined here https://github.com/dagger/dagger/pull/10567/files#diff-b93188b9d79aa9868ea185ca205ea31f7619506c4c764b9a5f49182aa9b48f3eR79. In the preliminary tests that I've done, they usually represent a very small % across all the spans in the trace.

vito: Ah true, yeah this might be fine then. I misread that as 'any API call coming from a module', but it's actually 'any API call going _to_ a module' so it should be much less frequent indeed.

dagql/idtui/golden_test.go
```diff
@@ -159,6 +167,82 @@ func (s TelemetrySuite) TestGolden(ctx context.Context, t *testctx.T) {
...
+						require.Equal(t, "TraceFunctionCalls", s.ExtraAttributes[telemetry.ModuleCallerFunctionCallNameAttr])
+						require.Equal(t, "github.com/dagger/dagger/viztest", strings.Split(s.ExtraAttributes[telemetry.ModuleCallerRefAttr].(string), "@")[0])
+						require.Equal(t, "getFiles", s.ExtraAttributes[telemetry.ModuleFunctionCallNameAttr])
+						require.Equal(t, "github.com/dagger/dagger/viztest/dep", strings.Split(s.ExtraAttributes[telemetry.ModuleRefAttr].(string), "@")[0])
+						depCalled = true
+					} else if s.Name == "Viztest.traceFunctionCalls" {
+						require.Equal(t, "", s.ExtraAttributes[telemetry.ModuleCallerFunctionCallNameAttr])
+						require.Contains(t, s.ExtraAttributes[telemetry.ModuleCallerRefAttr], "")
+						require.Equal(t, "traceFunctionCalls", s.ExtraAttributes[telemetry.ModuleFunctionCallNameAttr])
+						require.Contains(t, s.ExtraAttributes[telemetry.ModuleRefAttr], "github.com/dagger/dagger/viztest")
```
vito: nit: just missed a spot?

```suggestion
						require.Equal(t, s.ExtraAttributes[telemetry.ModuleCallerRefAttr], "")
						require.Equal(t, "traceFunctionCalls", s.ExtraAttributes[telemetry.ModuleFunctionCallNameAttr])
						require.Equal(t, s.ExtraAttributes[telemetry.ModuleRefAttr], "github.com/dagger/dagger/viztest")
```

vito (approved):
:tada:

PR #10568, opened 2025-06-11
Chore: make pathfunc optional for dagop wrappers

alexcb:
Refactors the dagop wrappers to make the path function optional. When ommitted, it will now default to the parent's `dir` or `file` value.

### DIFF
--- a/core/schema/container.go
+++ b/core/schema/container.go
@@ -567,7 +567,7 @@ func (s *containerSchema) Install() {
 			View(BeforeVersion("v0.12.0")).
 			Extend(),
 
-		dagql.NodeFunc("asTarball", DagOpFileWrapper(s.srv, s.asTarball, s.asTarballPath)).
+		dagql.NodeFunc("asTarball", DagOpFileWrapper(s.srv, s.asTarball, WithPathFn(s.asTarballPath))).
 			Doc(`Package the container state as an OCI image, and return it as a tar archive`).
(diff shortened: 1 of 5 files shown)


alexcb:
The one thing I am a bit unsure of, is previously the default `Path` value was hard-coded to either `/` or `file`, depending on the type that's being _returned_; however, this changes it to be dependant on the _parent_ type, which seems a bit odd.

e.g. `c.Directory("some-dir").File("a")` will have the default path set to `some-dir` rather than `file`; both of these seem wrong since we would want `some-dir/a`.

Then there's also the case of `c.Directory("foo").Directory("bar")`, which would set the default to `foo` rather than `bar`.

Maybe instead we should require a pathfunc be explicitly set in all cases? That might prevent future bugs since we'll have to be explicit about the behaviour we want.

jedevc:
Oh I see, yeah, it's a bit weird. Since the DagOpDirectory/DagOpFile are about the *return type*, then using the parent's arg is a bit weird :thinking: I see what you mean.

That said, I really like the transformation into an opts struct, if you wanna clean up this PR to just have that change, we should take that :grin:

alexcb:
> That said, I really like the transformation into an opts struct

thanks :)

I removed the default path func code I added, so we'll go back to the defaults of `"file"` or `"/"`, so this PR should simply be the opts struct change now.

PR #10569, opened 2025-06-11
chore: convert withFile to dagop

alexcb:


alexcb:
~~Note that if we merge #10568 first, we'll need to rebase this afterwards (or vice-versa) in order to prevent compilation errors.~~ done!

core/directory.go
```diff
@@ -530,46 +532,170 @@ func (dir *Directory) WithDirectory(ctx context.Context, destDir string, src *Di
...
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return false, nil
+		}
+		return false, err
+	}
+	return fi.Mode().IsDir(), nil
+}
+
+type hack interface {
```
jedevc: :eyes: we can move `schema.Evaluatable` into `core.Evaluatable`, then we don't need this hack.

jedevc (commented):
Looks good! Thanks!

One tiny comment, otherwise, good to merge :tada:

core/directory.go
alexcb: How embarrassing! I actually had this change in a previous draft, which must have been lost will all the rebasing. Good catch.

PR #10574, opened 2025-06-12
fix: remove `dev` flag in client gen API

TomChv:
Fixes Go SDK to rely on `replace` directive to detect if it should copy the Go SDK locally (for testing purpose).
By doing so, we can remove the `dev` argument from the API. Remove `dev` flag from `dagger client install`.

To use the local copy of the SDK on Go:

```
go mod edit -replace dagger.io/dagger=./<output-dir>/<sdk-path>
```

Example:
```
go mod edit -replace dagger.io/dagger=./dagger/sdk
```

This will add the Go SDK loaded in the engine:
```
> dagger client install go
# Generate client in `dagger`

> ls ./dagger/sdk
client.go     dagger.gen.go engineconn    go.mod        go.sum        querybuilder  telemetry
```


cmd/codegen/generator/go/generator.go
```diff
@@ -138,6 +138,48 @@ func (g *GoGenerator) GenerateModule(ctx context.Context, schema *introspection.
 	return genSt, nil
 }
 
+func (g *GoGenerator) isDaggerPackageReplaced() (string, bool, error) {
+	goModFile, err := os.ReadFile("go.mod")
+	if err != nil {
+		return "", false, fmt.Errorf("failed to read go.mod: %w", err)
+	}
+
+	goMod, err := modfile.Parse("go.mod", goModFile, nil)
```
jedevc, line 147:
```suggestion
	goMod, err := modfile.ParseLax("go.mod", goModFile, nil)
```

When go make breaking changes to the file format, it's useful to not *immediately* error out because of that :facepalm:

cmd/codegen/generator/go/generator.go
```diff
@@ -138,6 +138,48 @@ func (g *GoGenerator) GenerateModule(ctx context.Context, schema *introspection.
...
+			// Remove the output dir from the replace path and trim the leading slash to obtain
+			// a local path usable on overlay
+			// Example:
+			// - ./dagger/sdk generated in dagger -> sdk)
+			// - ./sdk generated in . -> sdk)
+			// - ./dagger/foo/bar generated in dagger -> foo/bar)
+			replacedPath = strings.TrimPrefix(
+				strings.TrimPrefix(filepath.Clean(replacedPath), filepath.Clean(g.Config.OutputDir)),
+				"/",
+			)
```
jedevc: This logic should become a bit simpler with `strings.CutPrefix`.

cmd/codegen/generator/go/generator.go
```diff
@@ -138,6 +138,48 @@ func (g *GoGenerator) GenerateModule(ctx context.Context, schema *introspection.
 	return genSt, nil
 }
 
+func (g *GoGenerator) isDaggerPackageReplaced() (string, bool, error) {
```
jedevc:
```suggestion
func (g *GoGenerator) daggerPackageReplacement() (string, bool, error) {
```

cmd/codegen/generator/go/templates/src/_dagger.gen.go/client.go.tmpl
```diff
@@ -1,187 +1,11 @@
...
 	query  *querybuilder.Selection
 	client graphql.Client
 }
 
-{{/* If dev mode is set, we recreate an implementation of the client */}}
-{{ if IsDevMode }}
-// ClientOpt holds a client option
-type ClientOpt interface {
-	setClientOpt(cfg *engineconn.Config)
-}
```
jedevc, line 18: Very very very neat :tada:

Since we just pull in `client.go` directly :grin:

cmd/codegen/generator/go/generator.go
```diff
@@ -138,6 +138,48 @@ func (g *GoGenerator) GenerateModule(ctx context.Context, schema *introspection.
...
+		if replace.Old.Path == "dagger.io/dagger" {
+			// We need to exclude the first parent directory of the replaced path since it's the
+			// root of the generated directory (c.Config.OutputDir) and the overlays root is that
+			// path.
+			// FIXME(TomChv): This will disapear once I fix the overlays root to the module root instead
+			// of the client output directory.
+			replacedPath := replace.New.Path
+
+			if strings.HasPrefix(replacedPath, "/") {
+				return "", false, fmt.Errorf("invalid go replace path %s ; path must have as parent the output directory %s", replacedPath, g.Config.OutputDir)
```
helderco:
```suggestion
				return "", false, fmt.Errorf("go replace path %q not under %q", replacedPath, g.Config.OutputDir)
```

cmd/codegen/generator/go/generator.go
```diff
@@ -146,12 +188,16 @@ func (g *GoGenerator) GenerateClient(ctx context.Context, schema *introspection.
 
 	layers := []fs.FS{mfs}
 
-	// If dev is set, we need to add local files to the overlay and change the package import
-	if g.Config.Dev {
+	replacedPath, replaced, err := g.isDaggerPackageReplaced()
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if dagger package is replaced: %w", err)
```
helderco:
```suggestion
		return nil, fmt.Errorf("failed to check if dagger.io/dagger package is replaced: %w", err)
```

docs/current_docs/api/snippets/services/use-endpoint/typescript/index.ts
```diff
@@ -9,7 +9,7 @@ export class MyModule {
     service = await service.start()
 
     // wait for service to be ready
-    let endpoint = await service.endpoint({"port": 80, "scheme": "http"})
+    const endpoint = await service.endpoint({ port: 80, scheme: "http" })
```
helderco: Did you intend to include this change? 🙂

helderco (commented):
lgtm, will wait until feedback is addressed before accepting.

docs/current_docs/api/snippets/services/use-endpoint/typescript/index.ts
TomChv: Yeah, the linter failed because of that, but it may be already fixed on main so it will be disappear when I rebase :)

cmd/codegen/generator/go/generator.go
TomChv: As discussed on Discord, this may not be possible with CutPrefix since the order of removal matters.

1. We check if the path is abs, if true -> raise error
2. We remove the output dir from the path
3. We remove the possible "/" that may rest if the output dir has no "/" (dagger -> dagger/ for ./dagger/sdk it will rest "/sdk" and we need "sdk" only)

cmd/codegen/generator/go/generator.go
TomChv: Done

cmd/codegen/generator/go/generator.go
TomChv: done

cmd/codegen/generator/go/generator.go
TomChv: Done

cmd/codegen/generator/go/generator.go
TomChv: Done

TomChv: It's sadly not working with `ParseLax` because the replace directive for the SDK directory will be dropped by Parse because the directory doesn't exist at the time we parse the mod (since we plan to create it during the generation).

Explanation from GPT:
```
If ../local/module does not exist or cannot be resolved in a real module context (e.g., you're parsing a raw go.mod file outside a real Go workspace), ParseLax may drop this entry.

But Parse includes it because it's structurally valid.
```

The workaround would be to directly use the format syntax and manually get the replace line to get the replace dir but it may lead to other bug.

PR #10577, opened 2025-06-12
Fix ListOfType::getSubType can accept ReturnsListOfTypes

Neirda24:
This is a very small fix that allows

```php
#[ReturnsListOfTypes(new ListOfType('string'))]
public function myCustomFunction(): array { }
```

### DIFF
--- a/sdk/php/src/ValueObject/ListOfType.php
+++ b/sdk/php/src/ValueObject/ListOfType.php
@@ -52,7 +52,7 @@ final readonly class ListOfType
     }
 
     private static function getSubtype(
-        Attribute\ListOfType $attribute,
+        Attribute\ListOfType|Attribute\ReturnsListOfType $attribute,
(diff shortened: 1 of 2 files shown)


charjr:
Ideally, I'd like to add some integration tests for this change too. 

This is where I'd want to add some tests: [dagger/core/integration/module_php_test.go](https://github.com/dagger/dagger/blob/21bfce95d2260c342d4f40870cd96a72031d2912/core/integration/module_php_test.go#L146)

It would look similar to each of those tests which call functions from here: [dagger/core/integration/testdata/modules/php/list-kind/src/ListKind.php](https://github.com/dagger/dagger/blob/main/core/integration/testdata/modules/php/list-kind/src/ListKind.php)

I would want to add something like capitilizeNestedStrings. 

If you'd like to write it, I'll be happy to help.
If not, I can add it separately if you'd prefer?

PR #10579, opened 2025-06-12
core: cache schema introspection

vito:
Following the evals refactor (#10562) the CI job started OOMing. A huge portion of this was logs, which is fixed by #10578, but there was still an alarming increase in heap usage (1.9GB -> 7.48GB) after extracting the evals out into a module and switching to interfaces:

image

This change addresses a longstanding TODO in `(*dagql.Server).Schema` to only re-generate the schema when it actually changes. That change on its own only addresses part of above, but then we also use that logic as a cache key for the `__schemaJSONFile` API, which altogether brings us from 1.9GB (pre-refactor) to 1.38GB (post-refactor with both fixes):

image

image

### DIFF
--- a/.changes/unreleased/Fixed-20250612-142053.yaml
+++ b/.changes/unreleased/Fixed-20250612-142053.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Fixed high engine memory usage exacerbated by high log volume and composing multiple modules. In one scenario this reduced heap usage from 14GB to 7GB.
+time: 2025-06-12T14:20:53.284900233-04:00
+custom:
+    Author: vito
+    PR: "10578"
--- a/.changes/unreleased/Fixed-20250612-142232.yaml
+++ b/.changes/unreleased/Fixed-20250612-142232.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Schema introspection is now cached, as long as the schema doesn't change. In the same scenario as the high log volume fix, this further reduced heap usage from 7GB to 1.4GB.
+time: 2025-06-12T14:22:32.678239656-04:00
+custom:
+    Author: vito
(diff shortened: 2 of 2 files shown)


sipsma:
Didn't review yet, but before I forget it'd be worth adding changelogs for this and the other OTEL log buffer fix. Seems worth highlighting to users

dagql/objects.go
```diff
@@ -27,6 +27,7 @@ type Class[T Typed] struct {
 	inner   T
 	idable  bool
 	fields  map[string][]*Field[T]
+	dirtier func()
```
sipsma: nit: worth a comment, would have no clue what this is if I was coming in with no context 😄

dagql/server.go
```diff
@@ -339,38 +364,58 @@ func (s *Server) Query(ctx context.Context, query string, vars map[string]any) (
...
+			schema.AddTypes(def)
+			schema.AddPossibleType(def.Name, def)
+		}
+		schema.Directives = map[string]*ast.DirectiveDefinition{}
+		for n, d := range s.directives {
+			schema.Directives[n] = d.DirectiveDefinition(view)
+		}
+		h := xxh3.New()
+		json.NewEncoder(h).Encode(s.schemas[view])
+		s.schemas[view] = schema
```
sipsma: probably missing something, but wouldn't we want these lines to flip in order? set `s.schemas[view] = schema` and then json encode`s.schemas[view]` to the hash? so we get the current one we just constructed hashed

vito: woops, yep - bad refactor

dagql/objects.go
vito: renamed this (and `bumpRevision`) to `invalidateSchemaCache`

PR #10580, opened 2025-06-12
add new "cloud" top level query to interact with Dagger Cloud

### LINKED ISSUE #10417: ✨ Way to ask Dagger API for your Dagger Cloud URL

For introspection/automation/CI/LLM etc

> I would have hoped for a more direct access to the full URL (including the org value which is retrieved from the token), maybe something like a getter on the engine object or something like this 🙂 
> I agree that the full trace URL is printed at the end of the pipeline, my concern is that we haven't found the right log level to set on our CI, and as I said before this URL can be buried under tons of logs. I wished I could find a way to easily print the cloud URL in GitHub actions job summary for instance.

https://discord.com/channels/707636530424053791/1347167341636747296/1349069168984719390

_No response_

_No response_

marcosnils:
introduces a new "cloud" top level query and type to interact with
Dagger Cloud. This was introduced a requirement from users to be
able to retrieve the Cloud URL from their pipelines.

Fixes #10417


marcosnils:
I'm also removing the CloudToken from the client metadata since the `analytics` endpoint doesn't require it

marcosnils:
CI seems sad due to  GCP outage? Getting 500 errors from chainguard.

core/schema/cloud.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (s *cloudSchema) Install() {
+	dagql.Fields[*core.Query]{
+		dagql.Func("cloud", s.cloud).
+			Doc("Dagger Cloud configuration and state"),
+	}.Install(s.srv)
+
+	dagql.Fields[*core.Cloud]{
+		dagql.Func("traceURL", s.traceURL).
+			DoNotCache("").
```
jedevc:
```suggestion
			DoNotCache("This value changes every single run").
```

core/schema/cloud.go
```diff
@@ -0,0 +1,49 @@
...
+	}.Install(s.srv)
+}
+
+func (s *cloudSchema) cloud(ctx context.Context, parent *core.Query, args struct{}) (*core.Cloud, error) {
+	return &core.Cloud{}, nil
+}
+
+func (s *cloudSchema) traceURL(ctx context.Context, parent *core.Cloud, args struct{}) (string, error) {
+	m, err := engine.ClientMetadataFromContext(ctx)
+
```
jedevc: nit: spacing

jedevc (commented):
Code seems good to me - fyi @shykes, public api change.

docs/docs-graphql/schema.graphqls
```diff
@@ -110,6 +113,22 @@ The `CacheVolumeID` scalar type represents an identifier for an object of type C
 """
 scalar CacheVolumeID
 
+"""Dagger Cloud configuration and state"""
+type Cloud {
+  """A unique identifier for this Cloud."""
+  id: CloudID!
+
+  """
+  The URL for the Dagger Cloud instance associated with the current trace
```
shykes:
```suggestion
  The trace URL for the current session
```

I didn't understand the comment (what's a "dagger cloud instance?") so proposing what I understood.

core/schema/cloud.go
```diff
@@ -0,0 +1,48 @@
...
+func (s *cloudSchema) Install() {
+	dagql.Fields[*core.Query]{
+		dagql.Func("cloud", s.cloud).
+			Doc("Dagger Cloud configuration and state"),
+	}.Install(s.srv)
+
+	dagql.Fields[*core.Cloud]{
+		dagql.Func("traceURL", s.traceURL).
+			DoNotCache("This value changes every single run").
+			Doc("The URL for the Dagger Cloud instance associated with the current trace"),
```
shykes:
```suggestion
			Doc("The trace URL for the current session"),
```

PR #10581, opened 2025-06-12
cli: don't serve transitive dependencies

vito:
when loading a module by ref, only serve that module, otherwise it's easy to hit name conflicts (e.g. `evaluator/workspace`, `evals/workspace`)

### DIFF
--- a/cmd/dagger/module_inspect.go
+++ b/cmd/dagger/module_inspect.go
@@ -68,7 +68,7 @@ func initializeModule(
 	}
 
 	serveCtx, serveSpan := Tracer().Start(ctx, "initializing module", telemetry.Encapsulate())
-	err = modSrc.AsModule().Serve(serveCtx, dagger.ModuleServeOpts{IncludeDependencies: true})
+	err = modSrc.AsModule().Serve(serveCtx)
 	telemetry.End(serveSpan, func() error { return err })
 	if err != nil {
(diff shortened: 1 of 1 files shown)


helderco:
Reference to original comment:
- https://github.com/dagger/dagger/pull/10562#discussion_r2139015116

> @cwlbraa @helderco Looking to double check this - without this fix, I ran into errors while trying to run this in shell, since there are two `Workspace` types:
> 
> * https://dagger.cloud/dagger/traces/71e3bd754197c0d17bc26a092e688a63?listen=ae52ee3b8632482b&listen=e9d1a95c779e20a7&span=9bab1abb165eac4a
> 
> With this fix applied it works fine:
> 
> * https://dagger.cloud/dagger/traces/09970837d24c661f4a23adeec263298c?span=a2257556816ecafd
> 
> This one _seems_ redundant - we still pass `IncludeDependencies: true` for the primary module (that's how I'm able to call `evaluator` above), just not here.

I'll defer to @cwlbraa since he's the one that needed this for LLM prompts in:
- https://github.com/dagger/dagger/pull/9992
- https://github.com/dagger/dagger/pull/10118

Note that serving dependencies was also added to `dagger query` and  `dagger listen` for consistency and removing this makes it inconsistent again, but we could add it as a CLI flag.

PR #10583, opened 2025-06-12
LLM: minor tweaks

vito:
(see individual commits, will self-review)

### DIFF
--- a/core/llm.go
+++ b/core/llm.go
@@ -80,6 +80,10 @@ type LLM struct {
 
 	// The environment accessible to the LLM, exposed over MCP
 	mcp *MCP
+
+	// Keep track of the last-used server state, so we know where to introspect
+	// types from in case the LLM gets passed around.
+	lastSrv *dagql.Server
 }
 
 type LLMEndpoint struct {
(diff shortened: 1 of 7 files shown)


core/llm.go
```diff
@@ -73,13 +73,17 @@ type LLM struct {
 	err  error
 
 	// History of messages
-	messages []ModelMessage
+	messages []*ModelMessage
```
vito, line 76: Context: this was a way-off shot in the dark at the OOMing issue encountered in #10562, but I think it makes sense to keep anyway. This slice gets copied here and there, so might as well avoiding copying the full message history.

core/llm.go
```diff
@@ -766,6 +766,29 @@ func (llm *LLM) loop(ctx context.Context, dag *dagql.Server) error {
...
+					attribute.String(telemetry.UIMessageAttr, "sent"),
+				))
+				defer span.End()
+				stdio := telemetry.SpanStdio(ctx, InstrumentationLibrary,
+					log.String(telemetry.ContentTypeAttr, "text/markdown"))
+				defer stdio.Close()
+				fmt.Fprint(stdio.Stdout, msg.Content)
+			}()
+		}
+
```
vito, line 791: This is a bit more work but I think makes more sense than showing the user prompt span at `WithPrompt` time, where it's not actually being sent to the LLM yet.

vito:
Eh, actually I don't think this is worth standing on its own; it's all in service of the evals so I'll let #10562 cover it.

Issue #10336, opened 2025-05-06
✨ Self-referencing modules for llm/env


eunomie:
### What are you trying to do?

[Environments](https://docs.dagger.io/api/llm#environments) are a way to define a specific set of functions an LLM will be able to see as tools. This allows to reduce the number of tools the LLM will see or to provide specific tools, both in order to improve the ability to complete a task.

> [!NOTE]
> This is extracted from #8030 but focusing only the LLM/Env/type awareness and not on the self calling aspects.

It's currently possible to use a primitive type as an environment input (like `Container` or `Directory` for instance). But to use a specific type, a workspace for instance, that embeds primitives and expose only certain functions, this needs to be declared in a different module, and used as a dependency.

This makes the usage of environments more complex where it could be interesting to have the code of the type we want to expose to the LLM at the same place than the rest of the code calling the LLM.

The current limitation is the generated code on the `Env` type (like `withXXXInput`, for instance `withContainerInput`) is not aware of the actual types defined by the module. The generated code is only aware of the types that are exposed to the engine, including dependencies of the current module, but not the module itself.

---

**Use Case**

Context:
- a type `Workspace` containing several functions to be seen as tools for the LLM

Currently:
- `dag.LLM().WithEnv(dag.Env().WithWorkspaceInput("input", dag.Workspace(src), "..."))....`
- `dag.Workspace` refers to a `workspace` module, as a dependency of the current module

Goal:
- `dag.LLM().WithEnv(dag.Env().WithWorkspaceInput("input", Workspace{src}, "..."))....`
- `Workspace` here is a local type

---

This means the generated code on `Env` is aware of the local types. If not possible then it means we have to generate types as we do usually for the dependencies and have something more like `dag.Env().WithWorkspaceInput("input", dag.Self().Workspace(src), "...")`but that would be more confusing for the users.

eunomie:
Let's say I have a module `my-agent` that contains a type `workspace` (a `struct` in Go, a `class` in Java, etc).

What I want is:

```go
dag.Env().WithWorkspaceInput("input", &Workspace{}, "the workspace")
```

**Type Exposure**

By default the `Workspace` type is not exposed. It needs to be exposed by a function of the module to be exposed. So it needs somethings like that:

```go
func (m *MyAgent) Workspace() *Workspace { return &Workspace{} }
```

Then we can imagine to keep the above call or to replace it by

```go
dag.Env().WithWorkspaceInput("input", m.Workspace(), "the workspace")
```

[...]

PR #10584, opened 2025-06-13
SDK: expose module type definitions

eunomie:
> [!NOTE]
> This has been extracted and cleaned from #10475 
> This is a first step to #10336 and #8030

The main purpose of this PR is to allow SDKs to expose the module type definitions on a specific function.

This is a mandatory step to unlock the ability for modules to self invoke themselves. This pull request contains the ability to self call themselves, but it's **experimental** as it doesn't support all the cases.

## Self Call Example

```go
func (m *MyModule) ContainerEcho(
	// +optional
	// +default="Hello Self Calls"
	stringArg string,
) *dagger.Container {
}

func (m *MyModule) Print(ctx context.Context, stringArg string) (string, error) {
	return dag.MyModule().ContainerEcho(dagger.MyModuleContainerEchoOpts{
		StringArg: stringArg,
	}).Stdout(ctx)
}

func (m *MyModule) PrintDefault(ctx context.Context) (string, error) {
	return dag.MyModule().ContainerEcho().Stdout(ctx)
}
```

In this example `print` function will call `container-echo` through the engine instead of directly.
This also allows to use directives like `optional` and `default` as you can see in `print-default`. This is something not doable by calling the function directly.

## `moduleDefs`

The PR introduces a new `moduleDefs` method SDK modules can implement.

Without this PR, the container returned by `moduleRuntime` is called in two different ways:

- with an empty string for parent name: the engine is asking for the type definition of the module
- with a parent name: the engine is invoking a function defined in the module

In a module using the Go SDK for instance, this is done in the generated `invoke` function you can find in `dagger.gen.go` that is used at the _entrypoint_ of the module container runtime. The `invoke` code looks like:

```go
switch parentName {
        case "MyModule":
                // invoking module's functions
        case "":
                // code that generates the type definition
                return dag.Module()....
}
```

The idea of `moduleDefs` is to split those two responsibilities.:

- when the engine used to call the runtime container with an empty parent name, the engine will call `moduleDefs` that ultimately returns a `dagger.Module` object (similar to what is performed inside the container when called with an empty function name)
- when the engine used to call the runtime container with a parent name, in order to invoke a module's functions, the container from `moduleRuntime` will be executed

## Implementation

The `moduleDefs` function, exposed by the SDK, needs to be executed on behalf of the module we are loading. This allows to use the right `engineVersion` for instance.
To do that, the `moduleDefs` is constructed the same way than the `moduleRuntime` function:

```graphql
moduleDefs(
  modSource: ModuleSource!
  introspectionJSON: File!
  outputFilePath: String!
): Container!
```

This function takes the module source, the introspection JSON as well as a file path to write the output.
And the function returns a Container with a defined entrypoint.
The entrypoint will be called by the engine, with a specific set of parameters to run it on behalf of the module (as the module is not yet loaded those internal parameters are required).

When the `moduleDefs` is called, it must write, in JSON, the ID of the module representing the different types exposed by the module.

A pseudo code version could be like:

```go
exposedTypes := dag.Module().WithObject(dag.TypeDef().WithObject()...)
json.WriteToFile(outputFilePath, exposedTypes.ToJSON())
```

The Engine will then read this `outputFilePath` file to load the module with the exposed types.

## Introspection JSON

Some SDKs might require to generate Dagger types before to be able to generate the types exposed by a module.
So it requires an introspection JSON to be available.
But then the module runtime needs a new introspection JSON extended with the exposed types.

The flow is the following when loading a module:
- get the introspection JSON based on the current types
- call `moduleDefs` with this introspection JSON
- append returned types to the module (to reference itself) -> only if self calls are enabled
- call `moduleRuntime` with a new introspection JSON based on the extended types
    - the container generated by `moduleRuntime` will be `sync`ed so the cache is resolved, even if the container will be called only later when invoking a function
    - this is a need for Go modules that requires a specific socket that is closed between the typedefs call and the invocation

If a SDK has already access to the default types (the module and typedefs types) because of the generated files are already part of the SDK or because they are published in a library that can be used (that's not yet the case for Java) then the introspection JSON on `moduleTypeDefs` can be safely ignored.

## Feature Flag / Compatibility

Self Calls are not enable by default. This means the SDKs must still work with the current way of doing, through runtime container. And in addition the new way with `moduleTypeDefs` for self calls.

> [!NOTE]
> For compatibility reasons, if the `moduleDefs` function is not exposed by the module SDK, then the
> `moduleRuntime` will be run and the container executed as before, and self calls will not be enabled.

To enable self calls, you can edit manually the `dagger.json` to add the experimental flag as follows:

```json
{
  "sdk": {
    "experimental": {
      "SELF_CALLS_FEATURE": true
    }
  }
}
```

And it's also possible to use `dagger init` and `dagger develop` commands:

- `dagger init --with-self-calls` to init a new module with self calls enabled
- `dagger develop --with-self-calls` to enable it on an existing module
- `dagger develop --without-self-calls` to disable it

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -6082,52 +6082,52 @@ func (m *Test) PrintDefault(ctx context.Context) (string, error) {
 }
 `,
 		},
-		{
-			sdk: "typescript",
-			source: `import { dag, Container, object, func } from "@dagger.io/dagger"
-
-@object()
-export class Test {
-  /**
-   * Returns a container that echoes whatever string argument is provided
-   */
-  @func()
-  containerEcho(stringArg: string = "Hello Self Calls"): Container {
-    return dag.container().from("alpine:latest").withExec(["echo", stringArg])
-  }
-
-  @func()
-  async print(stringArg: string): Promise<string> {
-    return dag.test().containerEcho({stringArg}).stdout()
-  }
-
-  @func()
-  async printDefault(): Promise<string> {
-    return dag.test().containerEcho().stdout()
-  }
-}
-`,
-		},
-		{
-			sdk: "python",
-			source: `import dagger
-from dagger import dag, function, object_type
-
-@object_type
-class Test:
-    @function
-    def container_echo(self, string_arg: str = "Hello Self Calls") -> dagger.Container:
-        return dag.container().from_("alpine:latest").with_exec(["echo", string_arg])
-
-    @function
-    async def print(self, string_arg: str) -> str:
-        return await dag.test().container_echo(string_arg=string_arg).stdout()
-
-    @function
-    async def print_default(self) -> str:
-        return await dag.test().container_echo().stdout()
-`,
-		},
+		//		{
+		//			sdk: "typescript",
+		//			source: `import { dag, Container, object, func } from "@dagger.io/dagger"
+		//
+		// @object()
+		// export class Test {
+		//   /**
+		//    * Returns a container that echoes whatever string argument is provided
+		//    */
+		//   @func()
+		//   containerEcho(stringArg: string = "Hello Self Calls"): Container {
+		//     return dag.container().from("alpine:latest").withExec(["echo", stringArg])
+		//   }
+		//
+		//   @func()
+		//   async print(stringArg: string): Promise<string> {
+		//     return dag.test().containerEcho({stringArg}).stdout()
+		//   }
+		//
+		//   @func()
+		//   async printDefault(): Promise<string> {
+		//     return dag.test().containerEcho().stdout()
+		//   }
+		// }
+		// `,
+		//		},
+		//		{
+		//			sdk: "python",
+		//			source: `import dagger
+		// from dagger import dag, function, object_type
+		//
+		// @object_type
+		// class Test:
+		//     @function
+		//     def container_echo(self, string_arg: str = "Hello Self Calls") -> dagger.Container:
+		//         return dag.container().from_("alpine:latest").with_exec(["echo", string_arg])
+		//
+		//     @function
+		//     async def print(self, string_arg: str) -> str:
+		//         return await dag.test().container_echo(string_arg=string_arg).stdout()
+		//
+		//     @function
+		//     async def print_default(self) -> str:
+		//         return await dag.test().container_echo().stdout()
+		// `,
+		//		},
 	}
 
 	for _, tc := range tcs {
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -61,7 +61,7 @@ func (t *TypescriptSdk) ModuleRuntime(
 		Container(), nil
 }
 
-func (t *TypescriptSdk) ModuleTypes(
+func (t *TypescriptSdk) ModuleTypesExp(
 	ctx context.Context,
 	modSource *dagger.ModuleSource,
 	introspectionJSON *dagger.File,
--- a/sdk/python/runtime/main.go
+++ b/sdk/python/runtime/main.go
@@ -181,7 +181,7 @@ func (m *PythonSdk) ModuleRuntime(
 }
 
 // Container for executing the Python module runtime
-func (m *PythonSdk) ModuleTypes(
+func (m *PythonSdk) ModuleTypesExp(
 	ctx context.Context,
 	modSource *dagger.ModuleSource,
 	introspectionJSON *dagger.File,


core/sdk.go
```diff
@@ -174,6 +174,40 @@ type Runtime interface {
...
+	*/
+	TypeDefs(
+		context.Context,
+
+		// Current module dependencies.
+		*ModDeps,
+
+		// Current instance of the module source.
+		dagql.Instance[*ModuleSource],
+	) (*Container, error)
```
jedevc: Should we not attempt to try and have this directly return a list of `TypeDef`s, or even a `Module`?

Even if the SDK runtime chooses to use a container behind the scenes, I think it makes sense to allow an SDK to directly return them if possible.

eunomie: This was with the idea to keep the changes as small as possible, but yes it might make more sense to do that. I'll have a look.

eunomie: I'm trying to see what would be the best option to do that, looking at both the Go SDK and the Java as they are quite different.

So I wonder if it wouldn't be a good idea to add something like a `fromJSON` or `load` function on the `dagger.Module` type that will take a json representation of those types.
So we can `dag.Module().FromJSON(jsonTypesFile)` and the runtime code is very small and almost shareable between SDKs, no function execution required as currently.
In Java SDK that could be something like:

```go
func (m *JavaSdk) ModuleTypeDefs(
	ctx context.Context,
	modSource *dagger.ModuleSource,
	introspectionJSON *dagger.File,
) (*dagger.Module, error) {
	mvnCtr, err := m.codegenBase(ctx, modSource, introspectionJSON)
	if err != nil {
		return nil, err
	}
	
	// returns a json file with the type details
	typesJson, err := m.generateTypes(ctx, mvnCtr, introspectionJSON)
	
	return dag.Module().
		FromJSON(typesJson), nil
}
```

The main change is the SDK needs to generate this json file instead of the type registration code.
Basically this is just this code but in JSON:
```go
	return dag.Module().
		WithObject(
			dag.TypeDef().WithObject("MyModule", dagger.TypeDefWithObjectOpts{SourceMap: dag.SourceMap("main.go", 8, 6)}).
				WithFunction(
					dag.Function("ContainerEcho",
						dag.TypeDef().WithObject("Container")).
						WithSourceMap(dag.SourceMap("main.go", 10, 1)).
						WithArg("stringArg", dag.TypeDef().WithKind(dagger.TypeDefKindStringKind), dagger.FunctionWithArgOpts{SourceMap: dag.SourceMap("main.go", 10, 34)})).
				WithFunction(
					dag.Function("Hello",
						dag.TypeDef().WithKind(dagger.TypeDefKindStringKind)).
						WithSourceMap(dag.SourceMap("main.go", 14, 1))))
```

@jedevc is that the kind of thing you had in mind?

jedevc: Yes! This is pretty much **exactly** what I had in mind as the end state. To simplify getting to that state, we could even have `dag.Module().AsJSON()`, so we keep similar typedef code for now and transfer it later? Just thinking out loud.

But yeah, in the spirit of keeping changes minimal for now, we only need to change the `TypeDefs` interface function to return a `Module` - under the hood, we can still do the changes in this patch, where we're dispatching that to a `Container`.

We can come and do the big JSON changes for each SDK one-by-one later, as a performance optimization.

sipsma:
@eunomie Looked at the `socket not found` error getting hit. It has to do with the fact that the new TypeDefs container and Runtime container are different (due to the new `--typedefs-only arg`) and thus end up as different cache entries. The Runtime container now doesn't get evaluated until the actual function invocation occurs, at which time it's running codegen (again) and looking for the SSH socket that's not available at that time.

I think the fundamental problem is that we end up running codegen twice now in this new scheme. That's not ideal for performance besides the socket issue it introduces.

I think splitting out the new `TypeDefs` method in the SDK interface makes perfect sense. But at least for this first step maybe it would make more sense to just implement `TypeDefs` for the Go SDK using the existing `Runtime` implementation? So `TypeDefs` for the go sdk just calls to `Runtime` and uses that container rather than trying to pass that extra `--typedefs-only` arg and bust cache between the two.

I haven't parsed through the exact behavior changes `--typedefs-only` introduces, so let me know if the suggestion above isn't actually possible. I'm just going off the idea the previously we used the `Runtime` container for getting typedefs, so it should theoretically be possible to continue to use it for now to implement `TypeDefs`.

eunomie:
Thanks @sipsma for the explanation.

> I think the fundamental problem is that we end up running codegen twice now in this new scheme. That's not ideal for performance besides the socket issue it introduces.
> 
> I think splitting out the new `TypeDefs` method in the SDK interface makes perfect sense. But at least for this first step maybe it would make more sense to just implement `TypeDefs` for the Go SDK using the existing `Runtime` implementation? So `TypeDefs` for the go sdk just calls to `Runtime` and uses that container rather than trying to pass that extra `--typedefs-only` arg and bust cache between the two.

The goal is to run codegen twice, and to decouple typedefs from runtime.
It's not this PR specifically, but the final goal is to allow self invocation. So this means to have the typedefs independently from the runtime, as the runtime might refer to the typedefs.

A very simple example would be the following:

```go
func (m *MyModule) ContainerEcho(stringArg string) *dagger.Container {
}

func (m *MyModule) Echo(stringArg string) *dagger.Container {
	return dag.MyModule().ContainerEcho(stringArg)
}
```

The `moduleRuntime` codegen will never succeed because of the dependency to `MyModule`. In fact we have sort of a circular dependency here. Currently typedefs depends on runtime, that requires codegen, that requires typedefs.
So my solution was to split it in two:
- the typedefs only codegen, that only looks at the types exposed and not at the code called
- the runtime that doesn't care about types but only about function calling

All that means 2 codegen phases, one for typedefs, one for function calls. Right now it's the same with just a flag, that could be completely different, or use something like @jedevc is mentioning above, but my understanding is it will not really change a lot to this problem, no?

So yes, for this very first phase it would be possible to avoid the different codegen and only use the existing `moduleRuntime`, but it means the very next phase is to split it. So not sure that would have a lot of benefit to push back this split. Or maybe there's a benefit to have the `moduleTypeDefs` relying on `moduleRuntime` that I don't see/understand.

eunomie:
Regarding performances, yes there's a cost here. But I don't see a lot of different ways to avoid it.
Maybe the above suggestion to completely abstract the types and avoid to return a container to run can help, at least for the Java SDK I can see that as a good solution. But there will be a cost.

Thinking again, maybe the idea of an intermediate json might help has there's not anymore this typedefs container. Maybe.

eunomie:
> The Runtime container now doesn't get evaluated until the actual function invocation occurs, at which time it's running codegen (again) and looking for the SSH socket that's not available at that time.

Could you detail a bit more? I'm not sure to understand the link between the container cached and not evaluated and the SSH socket.

jedevc:
> I haven't parsed through the exact behavior changes `--typedefs-only` introduces, so let me know if the suggestion above isn't actually possible. I'm just going off the idea the previously we used the `Runtime` container for getting typedefs, so it should theoretically be possible to continue to use it for now to implement `TypeDefs`.

Hah, okay, I chatted with @eunomie, lemme try and summarize and explain.

So, this is part of the work to do self-calls. We *need* the separation into two phases for this. For the go sdk (as currently structured in `main`) - with self calls, we won't actually be able do the typedef analysis until after we've done codegen.

Here's why:
1. The engine will get all the dependencies and try to do codegen, by generating a client and types and methods, etc.
2. Now, if the module code is actually valid, we can build it using `go build` into the `/runtime`.
3. We can then call `/runtime` with the empty string to get all the typedefs.

But. In the scenario of self-calls *the code won't be valid after generating all the types and methods* - because we'll be missing the self-bindings (we only know what those types should be after step 3). So we can't build it, so we can't get the typedefs. We have this weird circular dependency.

Thankfully, we can fix this. `TypeDefs` doesn't *need* to be dependent on codegen, it's kinda artificial - we can do all the go introspection of function signatures, and generate the typedefs without knowing every single type in the signature. One option (I think currently implemented in this PR) is to generate a single standalone binary (that doesn't even include our own source code) that **only** generates the typedefs, then we can build this, and actually get all the types at the end.

So, we need these two phases. To make sure that we fix the caching issue that @sipsma points out, we *need* to eagerly evaluate the `Runtime` container when we load the module. Atm in `main`, this is implied - we could just make that explicit (e.g. in `ModuleSource.AsModule`, just call `.Runtime` and throw away the result, so that it populates the cache) and that should solve the bug.

However, as @sipsma points out, there's a little bit of a perf issue here (curious to see how this pans out though, it might not be terrible?). For every module load, we're now having to generate, build, and execute two binaries. Ideally, the solution to this is to improve `TypeDefs` so that we don't need to build a binary at all - instead of generating code that builds typedefs, we could just port that code (might be automatable using an LLM?) to just directly generate the typedef structure (e.g. as JSON, see the suggestion [above](https://github.com/dagger/dagger/pull/10584#discussion_r2154098016)). Functionally, this would mean changing the method `TypeDefCode` to `TypeDefJSON` or similar: https://github.com/dagger/dagger/blob/aaa13befbe949553e6f0f3deb1b03ff5be2f34ec/cmd/codegen/generator/go/templates/module_objects.go#L171-L171

eunomie:
This is not a solution at all, but I just confirmed that with this change, that works:

```diff
diff --git a/core/sdk/go_sdk.go b/core/sdk/go_sdk.go
index 3d6fc8390..664aff2e9 100644
--- a/core/sdk/go_sdk.go
+++ b/core/sdk/go_sdk.go
@@ -523,13 +523,13 @@ func (sdk *goSDK) baseWithCodegen(
 	// cause failures if we also mount this in Runtime.
 	// Issue we run into is that when we try to run sdk checks
 	// using .dagger, it fails trying to find the socket
-	sshAuthSelectors, err := sdk.getUnixSocketSelector(ctx)
-	if err != nil {
-		return ctr, err
-	}
-	if len(sshAuthSelectors) > 0 {
-		selectors = append(selectors, sshAuthSelectors...)
-	}
+	//sshAuthSelectors, err := sdk.getUnixSocketSelector(ctx)
+	//if err != nil {
+	//	return ctr, err
+	//}
+	//if len(sshAuthSelectors) > 0 {
+	//	selectors = append(selectors, sshAuthSelectors...)
+	//}
 
 	// now that we are done with gitconfig and injecting env
 	// variables, we can run the codegen command.
```

(if there's no need for a private repo)

My understanding is the first codegen part (the one generating the typedefs) shouldn't require the ssh socket, as it shouldn't need the dependencies if I'm right (functions can return a local type or a core type, but not a dependent type, right?)

core/sdk.go
helderco: I understand the appeal at first glance but having the SDK generate a JSON document means it's extra work for SDKs to keep it in sync with changes to the typedef models. To have a nicely typed API I'd need to generate a client from a JSON schema, otherwise adding string based keys are easier to mistype and you'll only get feedback at runtime, instead of while developing.

On the other hand, the current way to extend the schema is completely lazy and already type safe, without having to generate a client specifically for this so I don't see the benefit. Since they're not mutually exclusive, I'm not opposed to having that option though.

helderco:
There's a chicken and egg situation with the Java SDK because of the new `Module.fromJSON` method. The SDK modules can't use a dev build of the engine. Need to wait for a release, otherwise running tasks from the main `.dagger` module with a stable `dagger` binary will fail. So changes need to be compatible between dev and stable.

TomChv:
I rebased the PR since there were some conflict introduced by #10683.

eunomie:
> There's a chicken and egg situation with the Java SDK because of the new `Module.fromJSON` method. The SDK modules can't use a dev build of the engine. Need to wait for a release, otherwise running tasks from the main `.dagger` module with a stable `dagger` binary will fail. So changes need to be compatible between dev and stable.

Right. So if I move the Java SDK implementation on a new PR that should solve it, right? In that case we can keep this PR with the main behaviour and the Go SDK implementation, then have different PRs for Java, Python and Typescript.

helderco:
As long as the Go module isn't using a new or incompatible API you can include in this PR. For example, if instead of using `fromJSON` you return a graphql query from Java, the new `moduleTypeDefs` function can resolve it using current features so it won't break. My Python implementation of `moduleTypeDefs` doesn't use `fromJSON`, for example, and it's not an issue.

eunomie:
I did some tests related to performances, to have a first idea of the current cost:

`dagger init --sdk=go`

| total time | time inside generateContextDirectory | PR total time | PR time generateContextDirectory |
|----------|------------|----|----|
| 5.6 | 4.9 | 8.3 (+2.7) | 7.5 (+2.6) |
| 1.6 | 0.9 | 3.6 (+2) | 2.6 (+1.7) |
| 1.7 | 1.1 | 3.7 (+2) | 2.6 (+1.5) |
| 1.8 | 1.0 | 3.7  (+1.9) | 2.6 (+1.6) |

`dagger call container-echo --string-arg Hello stdout`

| total time | time to load module | PR total time | PR time to load module |
|----------|------------|----|----|
| 3.3 | 2.3 | 4.9 (+1.6) | 2.2 (=) |
| 1.6 | 0.4 | 1.7 (=) | 0.5 (=) |
| 1.6 | 0.5 | 1.5 (=) | 0.5 (=) |
| 1.7 | 0.5 | 1.5 (=) | 0.5 (=) |

Same call after an edit in the module's code

| total time | time to load module | PR total time | PR time to load module |
|----------|------------|----|----|
| 3.2 | 2.0 | 3.9 | 2.4 |
| 1.6 | 0.4 | 1.6 | 0.5 |

---

- the init time of a go module is slower
- the first call to a module (when it's about to re-generate the code) it's slower
- when it's not the first call, performances are similar

The init duration is something that can be improved. Right now, the initialisation is done... two times. The template code is generated once to create the types the first time, then it's discarded and it's generated again with the codegen part.
This is sub-optimal, but this only affects the `dagger init` command.

First call of a module (or after a change) is also slower. Less in absolute but it's still ~50%.

jedevc:
@eunomie that seems like a pretty hefty perf hit. 

That said, might be worth iterating on that in `main`? Not to extend this PR out way further, if there's no super low hanging fruit.

But, if we're taking such a performance hit, IMO, we shouldn't be enabling it by default - is there a way that we could allow modules to opt-in, and then they can perform self calls?

eunomie:
> we shouldn't be enabling it by default - is there a way that we could allow modules to opt-in, and then they can perform self calls?

I haven't done it yet, but I think this is doable to not enable it by default.
I was thinking about a global flag instead of per module (also because I think this might be easier to do, globally in the engine instead of per SDK)

But I agree we should push it to main and iterate from there. It will be easier with smaller PRs.

eunomie:
A quite important change to notice: the global flag has been removed in favour of a flag per module (inside `dagger.json` to enable self calls.

```json
{
  "sdk": {
    "experimental": {
      "self-calls": true
    }
  }
}
```

This also means (in fact that was also the case previously) that all SDK that wants to enable self calls must be backward compatible.

I've fixed the Python code that way, at least as a temporary solution.

If/when performances issues will be fixed, then we can imagine to always use the new way.

eunomie:
```console
$ dagger init --sdk=go --with-self-calls my-module
```

eunomie:
I've pushed an important commit: the module/typedefs JSON serialisation has been removed 🎉 
This reduces the size of the PR by ~46% so that's good on this side.
This also makes things simpler in Go SDK.

Now, Go SDK has two aspects, one to generate the registration code string (used in main `dagger.gen.go`, if self calls are not enabled) and one to generate almost the same but by executing the code and not just the string (used to create the module typedefs and send the id back).

There's still room for improvement here, especially when we will enable self calls by default, but that's a great step for the PR.

eunomie:
Two new flags to `develop`:
- `dagger develop --with-self-calls` will enable self calls on an existing module
- `dagger develop --without-self-calls` will disable them

eunomie:
I have something a bit weird, not sure if I did something wrong.

If I init a module, by default without self calls, then enable it, everything is working fine.

```console
$ dagger init --sdk=typescript my-module
$ cd my-module
$ dagger develop --with-self-calls
```

If I try to init the module with self calls, it doesn't work.

```console
$ dagger init --sdk=typescript --with-self-calls my-module
▶ connect 0.5s

● moduleSource(refString: "my-module", disableFindUp: true, allowNotExists: true, requireKind: LOCAL_SOURCE): ModuleSource! 0.0s
▶ .configExists: Boolean! 0.0s

▶ ModuleSource.localContextDirectoryPath: String! 0.0s

▶ ModuleSource.sourceRootSubpath: String! 0.0s

● ModuleSource.withName(name: "my-module"): ModuleSource! 0.0s
● .withSDK(source: "typescript"): ModuleSource! 6.6s
▶ .withSourceSubpath(path: "."): ModuleSource! 0.0s
● .withEngineVersion(version: "latest"): ModuleSource! 0.0s
● .withExperimentalFeatures(features: [SELF_CALLS_FEATURE]): ModuleSource! 0.0s
▼ .generatedContextDirectory: Directory! 5.6s ERROR
╰─▼ Container.withExec(
    ┆ args: ["tsx", "--no-deprecation", "--tsconfig", "/src/tsconfig.json", "/src/src/__dagger.entrypoint.ts"]
    ┆ experimentalPrivilegedNesting: true
    ): Container! 1.5s ERROR
  ┃ could not find module entrypoint: class MyModule from import. Class should be exported to benefit from all features.
  ┃ /src/sdk/core.js:91601
  ┃       throw new IntrospectionError(`could not find main object ${this.name} in module ${JSON.stringify(mainModule, null, 2)
  ┃ ""} located at ${this.ast.files}`);
  ┃             ^
  ┃
  ┃
  ┃ IntrospectionError: could not find main object MyModule in module  located at /src/src/__dagger.entrypoint.ts,/src/src/../s
  ┃ lient.gen.ts
  ┃     at new DaggerModule (/src/sdk/core.js:91601:13)
  ┃     at scan (/src/sdk/core.js:91951:19)
  ┃     at async connection.LogOutput (/src/sdk/core.js:92143:27)
  ┃     at async <anonymous> (/src/sdk/core.js:90285:11)
  ┃     at async withGQLClient (/src/sdk/core.js:90267:12)
  ┃     at async <anonymous> (/src/sdk/core.js:90283:9)
  ┃     at async connection (/src/sdk/core.js:90281:5)
  ┃     at async entrypoint (/src/sdk/core.js:92141:3) {
  ┃   cause: undefined,
  ┃   code: 'D110'
  ┃ }
  ┃
  ┃ Node.js v22.11.0
  ! process "tsx --no-deprecation --tsconfig /src/tsconfig.json /src/src/__dagger.entrypoint.ts" did not complete successfully:
    exit code: 1
```

eunomie:
Typescript init is fixed. It's just in case of a `dagger init --with-self-calls` the `ModuleTypeDefs` function needs user module source code to generate the types. But it doesn't exist _yet_, as it's created during typedef.
So I just did what is already done in Go SDK, the template is created inside `ModuleTypeDefs` if no user files are present. It will be re-created during codegen, that's something to optimise later, but it shouldn't be a big issue, even on performances side.

core/sdk.go
eunomie: done differently, without JSON anymore

eunomie:
I just pushed an updated version that contains tests for Go, Python and Typescript (instead of Go only)

core/modulesource.go
```diff
@@ -799,3 +807,33 @@ func (fs ModuleSourceStatFS) Stat(ctx context.Context, path string) (*fsutiltype
...
+type ModuleSourceExperimentalFeature string
+
+func (f ModuleSourceExperimentalFeature) String() string { return string(f) }
+
+var ModuleSourceExperimentalFeatures = dagql.NewEnum[ModuleSourceExperimentalFeature]()
+
+var (
+	ModuleSourceExperimentalFeatureSelfCalls = ModuleSourceExperimentalFeatures.Register("SELF_CALLS_FEATURE", "Self calls")
+	_                                        = ModuleSourceExperimentalFeatures.AliasView("SELF_CALLS", "SELF_CALLS_FEATURE", enumView)
+)
```
jedevc, line 1232: Aliases aren't needed here. The only reason we need them for other cases is for backwards compat.

Because this is a new property, we can just use the ideal name, `SELF_CALLS`.

cmd/codegen/generate_typedef.go
```diff
@@ -0,0 +1,71 @@
...
+	defer cfg.Close()
+
+	// ensure we have a dagger connection, this will be required to create type defs
+	if cfg.Dag == nil {
+		dag, err := dagger.Connect(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to connect to dagger daemon: %w", err)
+		}
+		cfg.Dag = dag
+	}
```
jedevc: nit: this logic should be encapsulated into `getGlobalConfig` - it could take a flag, `connect bool`, and then the caller can determine whether we need to create a connection.

cmd/codegen/typedefs.go
```diff
@@ -0,0 +1,60 @@
...
+		introspectionSchema = resp.Schema
+		introspectionSchemaVersion = resp.SchemaVersion
+
+		// Set the parent schema
+		generator.SetSchemaParents(introspectionSchema)
+	}
+
+	logsW := os.Stdout
+
+	_, _ = fmt.Fprintf(logsW, "generating %s typedefs\n", cfg.Lang)
```
jedevc: nit: `slog`

cmd/codegen/generator/go/templates/module_types.go
```diff
@@ -20,6 +21,14 @@ type ParsedType interface {
 
 	// Go types referred to by this type
 	GoSubTypes() []types.Type
+
+	// TypeDef representation of the type
+	TypeDefObject(*dagger.Client) (*dagger.TypeDef, error)
```
jedevc: nit: just call this `TypeDef`? Since this is the "primary" way to do this now?

And maybe group next to `TypeDefCode`?

cmd/codegen/generator/go/templates/module_objects.go
```diff
@@ -231,6 +232,57 @@ func (spec *parsedObjectType) TypeDefCode() (*Statement, error) {
...
+	withObjectOpts := dagger.TypeDefWithObjectOpts{}
+	if spec.doc != "" {
+		withObjectOpts.Description = strings.TrimSpace(spec.doc)
+	}
+	if spec.sourceMap != nil {
+		withObjectOpts.SourceMap = spec.sourceMap.TypeDefObject(dag)
+	}
+	if spec.name == "" {
+		panic(spec)
+	}
```
jedevc, line 181: nit: return an error instead of `panic`ing

cmd/codegen/generator/go/templates/typedefs.go
```diff
@@ -0,0 +1,259 @@
...
+	errEmptyCode = fmt.Errorf("no code yet")
+)
+
+func (funcs goTemplateFuncs) visitTypes(
+	strict bool,
+	rootVisitor rootVisitor,
+	structVisitor structVisitor,
+	ifaceVisitor ifaceVisitor,
+	enumVisitor enumVisitor,
+) error {
```
jedevc: nice! :tada:

litte suggestion: modify this to take an `opts` struct so we can name the visitors when we pass them in.

core/json.go
```diff
@@ -64,6 +64,9 @@ func (p *JSON) UnmarshalJSON(bs []byte) error {
 	if p == nil {
 		return fmt.Errorf("cannot unmarshal into nil JSON")
 	}
+	if string(bs) == "null" {
+		return nil
+	}
```
jedevc: Why is this needed? This means we actually skip writing to `p`, which doesn't seem right.

core/module.go
```diff
@@ -143,9 +143,11 @@ func (mod *Module) Install(ctx context.Context, dag *dagql.Server) error {
 		if ok {
 			// NB: this is defense-in-depth to prevent SDKs or some other future
 			// component from allowing modules to extend external objects.
-			return fmt.Errorf("type %q is already defined by module %q",
-				objDef.Name,
-				modType.SourceMod().Name())
+			// TODO: Ensure module is defined only once
```
jedevc: :eyes:

core/modules/config.go
```diff
@@ -90,6 +90,8 @@ type ModuleConfigUserFields struct {
 type SDK struct {
 	Source string         `json:"source"`
 	Config map[string]any `json:"config,omitempty"`
+	// The experimental features enabled for this module.
+	Experimental map[string]bool `json:"experimental,omitempty"`
```
jedevc, line 94: LGTM! (we've previously discussed separately)

Wondering if this should just be `[]string` though?

cc @shykes just checking this looks like the right place in `dagger.json` to have this setting?

core/schema/modulesource.go
```diff
@@ -1504,6 +1519,53 @@ func (s *moduleSourceSchema) moduleSourceWithUpdateBlueprint(
...
+	if len(args.Features) > 0 {
+		if tmpSrc.SDK.Experimental == nil {
+			tmpSrc.SDK.Experimental = make(map[string]bool)
+		}
+		for _, feature := range args.Features {
+			tmpSrc.SDK.Experimental[feature.String()] = false
+		}
+	} else {
+		tmpSrc.SDK.Experimental = make(map[string]bool)
+	}
```
jedevc, line 1601: suggestion: `withoutExperimentalFeatures` should probably remove the properties.

The entire `experimental` field should be removed if there are no enabled properties.

cmd/codegen/typedefs.go
```diff
@@ -0,0 +1,60 @@
...
+
+		// Ignoring generated.PostCommands:
+		// PostCommands are used to perform Go tasks like go mod tidy.
+		// Some commands are needed to ensure types are correctly read, but the final ones are not
+		// as we don't care about the runnable code, only about types and function signatures.
+
+		if generated.NeedRegenerate {
+			slog.Info("needs another pass...")
+			continue
+		}
```
jedevc, line 50: Is this even possible?

core/sdk.go
```diff
@@ -168,6 +168,25 @@ type Runtime interface {
...
+
+	TypeDefs(
+		context.Context,
+
+		// Current module dependencies.
+		*ModDeps,
+
+		// Current instance of the module source.
+		dagql.ObjectResult[*ModuleSource],
+	) (dagql.ObjectResult[*Module], error)
```
jedevc, line 206: I would suggest splitting this into a new interface. Then we can simplify `HasModuleTypeDefs` into just checking `typedefs, hasTypeDefs := sdk.(Typedefs)`.

core/sdk/go_sdk.go
```diff
@@ -216,6 +221,111 @@ func (sdk *goSDK) Codegen(
...
+		},
+		dagql.Selector{
+			Field: "file",
+			Args: []dagql.NamedInput{
+				{
+					Name:  "path",
+					Value: dagql.NewString("typedefs.json"),
+				},
+			},
+		},
```
jedevc, line 352: nit: allow specifying `typedefs.json` as a flag to `generate-typedefs`

core/sdk/go_sdk.go
```diff
@@ -226,6 +336,8 @@ func (sdk *goSDK) Runtime(
 		return inst, fmt.Errorf("failed to get dag for go module sdk runtime: %w", err)
 	}
 
+	ctx, span := core.Tracer(ctx).Start(ctx, "go SDK: load runtime")
+	defer telemetry.End(span, func() error { return rerr })
```
jedevc:
```suggestion
```

This was removed in #10468

core/typedef.go
```diff
@@ -526,6 +526,11 @@ func (typeDef *TypeDef) WithObjectConstructor(fn *Function) (*TypeDef, error) {
 	typeDef = typeDef.Clone()
 	fn = fn.Clone()
 	fn.ParentOriginalName = typeDef.AsObject.Value.OriginalName
+	// Constructors are invoked by setting the ObjectName to the name of the object its constructing and the
+	// FunctionName to "", so ignore the name of the function.
+	// This is to be aligned with moduleSchema.typeDefWithObjectConstructor
```
jedevc, line 591: Makes sense! We can remove the logic from `moduleSchema.typeDefWithObjectConstructor` then right?

sdk/go/fs.go
```diff
@@ -23,3 +23,6 @@ var GoSum []byte
 
 //go:embed engineconn/*.go querybuilder/marshal.go querybuilder/querybuilder.go go.mod go.sum client.go dagger.gen.go telemetry/*.go
 var GoSDK embed.FS
+
+//go:embed dagger.gen.go
+var GoDagGen []byte
```
jedevc, line 28: Struggling to understand why we need this - can we not just generate this in the binary? 

I guess it is a speed-up (?), but it does feel like it just *expands* the number of possible paths through the `codegen` binaryh.

Also just feels a little strange that the binary that generates this file can embed itself. Maybe we could just put this file *into* the go runtime container?

sdk/typescript/runtime/main.go
```diff
@@ -60,6 +62,44 @@ func (t *TypescriptSdk) ModuleRuntime(
...
+
+	modID, err := runtime.
+		Container().
+		WithMountedFile(runtime.cfg.entrypointPath(), entrypointFile()).
+		WithEnvVariable("REGISTER_TYPEDEF", "true").
+		WithEnvVariable("MODULE_NAME", runtime.cfg.name).
+		WithExec(runtime.runtimeCmd(), dagger.ContainerWithExecOpts{
+			ExperimentalPrivilegedNesting: true,
+		}).
+		File(moduleIDPath).
```
jedevc: `moduleIDPath` should be somehow parameterized.

jedevc (commented):
Looks reasonable! Lots of minor comments, but the general shape looks right to me!

I think the big next step would be to see how we can optimize - one idea would be to have the go `codegen` produce one query instead of an ID? That'll reduce the need to actually communicate with the engine, so there's one big query instead of lots of little queries.

However, as we previously discussed, that would be a big change to the actual schema.

I think let's get this in soon, and we can benchmark what needs doing!

jedevc:
Pinging @helderco and @TomChv for review as well.

cmd/codegen/generator/go/templates/module_types.go
jedevc: Also, curious under what conditions we now actually use the generated `TypeDefCode` - could we remove that?

.changes/unreleased/Added-20250820-165515.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Added
+body: allow modules to call their own functions through the engine (self calls)
```
TomChv, line 2: A bit more explanation on what exactly you can do now may be useful, or maybe we can save that for a doc page as a follow up

cmd/codegen/generate_typedef.go
TomChv: +1 so we can also remove the `cfg.Close` and just check if the `connect` flag is set to `true`

cmd/codegen/generator/go/generate_typedefs.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	pkgInfo, partial, err := g.bootstrapMod(mfs, res)
+	if err != nil {
+		return nil, fmt.Errorf("bootstrap package: %w", err)
+	}
+	if outDir != "." {
+		if err = mfs.MkdirAll(outDir, 0700); err != nil {
+			return nil, err
+		}
```
TomChv, line 50: Nit: All this code up to the call to `generateTypedef` is very similar to generateModule, I wonder if we can simplify it with a common function.
It's not a blocker at all, just an idea for a follow up

cmd/dagger/module.go
```diff
@@ -51,6 +51,9 @@ var (
 	developSourcePath string
 	developRecursive  bool
 
+	selfCalls   bool
+	noSelfCalls bool
```
TomChv, line 55: Why both? if one is true the other's false and vice-versa specially since you're only using `selfCalls` in the main logic except for a check

cmd/codegen/generator/go/templates/module_types.go
eunomie: The `TypeDefCode` is used to generate the code, for `dagger.gen.go` file.
Both are really close, but it's more a question of one is executed right away where the other is written to a file.

cmd/codegen/generator/go/templates/module_objects.go
eunomie: I think that was only used to debug 🫣 But I'm replacing it with an error, yes.

core/json.go
eunomie: 🤔 
I don't remember exactly, but as I removed the json export, I'll remove this also, forgot to do it...

cmd/codegen/typedefs.go
eunomie: yes, for instance on `dagger init`

core/integration/module_go_test.go
```diff
@@ -596,6 +596,62 @@ func main() {
...
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work").
+			With(daggerExec("init", "--name=bare", "--sdk=go", "--with-self-calls"))
+
+		daggerjson, err := modGen.File("dagger.json").
+			Contents(ctx)
+		require.NoError(t, err)
+		require.Contains(t, daggerjson, "\"SELF_CALLS_FEATURE\": true")
+
+		modGen = modGen.With(daggerExec("develop", "--without-self-calls"))
```
TomChv, line 648: So that means we can "remove" self-calls generation? is there another way to do that instead of an extra flag?

core/integration/module_test.go
```diff
@@ -5587,6 +5587,114 @@ export class Dep {
 	}
 }
 
+func (ModuleSuite) TestSelfCalls(ctx context.Context, t *testctx.T) {
```
TomChv, line 6049: Can we also test support for constructor and object return type? 

Example in go:

```go
type Test struct {
  A string
}

func New(a string) *Test {
  return &Test{A: a}
}

func (t *Test) Self() *Test {
  return t
}

// To see if we can self call the constructor
func (t *Test) Build(a string) *Test {
  return dag.Test({A: a}) 
}

// Just to check if we can self call a function that returned the self object
func (t *Test) Clone() *Test {
  return dag.Test({A: t.A}).Self()
}
```

```
dagger call --a "foo" build a // Should output "foo"
```

core/sdk.go
TomChv: I agree, I'm not a fan of this `HasModuleTypeDefs` function since we already have a generic function to check/convert to the right interface with the `AsXXX`.

core/sdk/module_runtime.go
```diff
@@ -14,12 +14,21 @@ type runtimeModule struct {
 	mod *module
 }
 
-func (sdk *runtimeModule) Runtime(
+func (sdk *runtimeModule) HasModuleTypeDefs() bool {
```
TomChv: Can we remove this part in favor of something in the `SDK` interface?

Typedef is part of runtime but there may be another way?

Not a blocker though, it might be simpler to fetch all the SDK capabilities before calling anything so that way we don't have this check here.

```go
type Capability string

var (
  ImplementsRuntime Capability = "ImplementRuntime"
  ImplementsTypedef =  "ImplementTypedef"
  ImplementCodegen = // ...
)

type SDK interface {
  Capabilities() []Capability

// OR

  IsImplementing(Capability) bool
}
```

Then we can simply change the As to directly convert to the right interface.

core/sdk/module_runtime.go
```diff
@@ -14,12 +14,21 @@ type runtimeModule struct {
...
+func (sdk *runtimeModule) TypeDefs(
 	ctx context.Context,
 	deps *core.ModDeps,
 	source dagql.ObjectResult[*core.ModuleSource],
-) (inst dagql.ObjectResult[*core.Container], rerr error) {
-	ctx, span := core.Tracer(ctx).Start(ctx, "module SDK: load runtime")
+) (inst dagql.ObjectResult[*core.Module], rerr error) {
+	if !sdk.HasModuleTypeDefs() {
+		return inst, fmt.Errorf("failed to get typedefs object: module %s does not implement moduleTypeDefs", sdk.mod.mod.Self().Name())
+	}
```
TomChv: I like the extra safety but this is already handled before no?

sdk/typescript/runtime/main.go
TomChv: Why? It's inside the typedef logic and the ts entrypoint isn't something we can or should export, I could do it from an env variable but I don't understand what would be the addition?

sdk/typescript/runtime/main.go
```diff
@@ -81,15 +121,8 @@ func (t *TypescriptSdk) Codegen(
...
-	// Check if there's any user source files, if not, add the template file.
-	// NOTE: This should be moved in a `Init` function once we improve the SDK interface.
-	sourcesFiles, err := runtimeBaseCtr.Container().Directory(".").Glob(ctx, "src/**/*.ts")
-	if err != nil {
-		return nil, fmt.Errorf("failed to list user source files: %w", err)
-	}
-
-	if len(sourcesFiles) == 0 {
-		runtimeBaseCtr = runtimeBaseCtr.withInitTemplate()
+	if runtimeBaseCtr, err = runtimeBaseCtr.addInitTemplateIfNoUserFile(ctx); err != nil {
```
TomChv, line 117: Why this change?

core/schema/modulesource.go
```diff
@@ -1893,6 +1956,13 @@ func (s *moduleSourceSchema) loadModuleSourceConfig(
 	return modCfg, nil
 }
 
+func canSelfCalls(src dagql.ObjectResult[*core.ModuleSource]) bool {
+	sdkImpl, ok := src.Self().SDKImpl.AsRuntime()
+	return ok &&
+		sdkImpl.HasModuleTypeDefs() &&
+		src.Self().SDK.ExperimentalFeatureEnabled(core.ModuleSourceExperimentalFeatureSelfCalls)
+}
```
TomChv: I'm not a fan of that, just `src.Self().IsImplementing(XXXX) && src.Self().SDK.Experimental...` should be enough

core/schema/modulesource.go
```diff
@@ -1922,6 +1992,26 @@ func (s *moduleSourceSchema) runCodegen(
...
 	}
 
+	// If possible, add the types defined by the module itself to the "deps" so that they can be
+	// part of the code generation.
+	// This is not really a dependency as it's the module itself, but that will allow to generate
+	// the types.
+	if srcInst.Self().SDK != nil {
+		// Only if the SDK implements a specific `moduleTypeDefs` function.
+		// If not, we will have circular dependency issues.
+		if canSelfCalls(srcInst) {
```
TomChv: Nit: not a blocker but we should move a big part of that `runCodegen` function logic into smaller function, it's becoming way to big.
I'll try to follow up with that whenever I have time

sdk/typescript/runtime/main.go
eunomie: Without this, a `dagger init --sdk=typescript --with-self-calls` is failing.
So the init template is also added to the typedefs phase.

This change particularly is because I extracted this logic in a dedicated function as it's used at two different places.

<details><summary>FTR the change (diff pasted, to avoid link issues as I'll force push...)</summary>

```diff
commit 4cc7ead800f2fffe1eaacd21a5224cb985695d8a
Author: Yves Brissaud <yves@dagger.io>
Date:   Tue Aug 19 16:33:02 2025 +0200

    fix typescript init --with-self-calls
    
    When running
    
        dagger init --sdk=typescript --with-self-calls my-module
    
    there is no user code. So the template needs to be added.
    
    This was donc in codegen phase, but now this is required for moduleTypeDefs
    phase.
    
    Extract the logic to add the init template and call it also inside ModuleTypeDefs
    function.
    
    Signed-off-by: Yves Brissaud <yves@dagger.io>

diff --git a/sdk/typescript/runtime/main.go b/sdk/typescript/runtime/main.go
index 5f7649339..33f787040 100644
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+
 	"typescript-sdk/internal/dagger"
 )
 
@@ -78,6 +79,10 @@ func (t *TypescriptSdk) ModuleTypeDefs(
 		withInstalledDependencies().
 		withUserSourceCode()
 
+	if runtime, err = runtime.addInitTemplateIfNoUserFile(ctx); err != nil {
+		return nil, err
+	}
+
 	modID, err := runtime.
 		Container().
 		WithMountedFile(runtime.cfg.entrypointPath(), entrypointFile()).
@@ -116,15 +121,8 @@ func (t *TypescriptSdk) Codegen(
 		withGeneratedLockFile().
 		withUserSourceCode()
 
-	// Check if there's any user source files, if not, add the template file.
-	// NOTE: This should be moved in a `Init` function once we improve the SDK interface.
-	sourcesFiles, err := runtimeBaseCtr.Container().Directory(".").Glob(ctx, "src/**/*.ts")
-	if err != nil {
-		return nil, fmt.Errorf("failed to list user source files: %w", err)
-	}
-
-	if len(sourcesFiles) == 0 {
-		runtimeBaseCtr = runtimeBaseCtr.withInitTemplate()
+	if runtimeBaseCtr, err = runtimeBaseCtr.addInitTemplateIfNoUserFile(ctx); err != nil {
+		return nil, err
 	}
 
 	// Extract codegen directory
diff --git a/sdk/typescript/runtime/module.go b/sdk/typescript/runtime/module.go
index 9fe0a7585..824eada25 100644
--- a/sdk/typescript/runtime/module.go
+++ b/sdk/typescript/runtime/module.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"path/filepath"
+
 	"typescript-sdk/internal/dagger"
 	"typescript-sdk/tsdistconsts"
 
@@ -344,6 +346,27 @@ func (m *moduleRuntimeContainer) withInitTemplate() *moduleRuntimeContainer {
 	return m
 }
 
+// Check if there's any user source files
+func (m *moduleRuntimeContainer) hasUserSourceFiles(ctx context.Context) (bool, error) {
+	sourcesFiles, err := m.ctr.Directory(".").Glob(ctx, "src/**/*.ts")
+	if err != nil {
+		return false, fmt.Errorf("failed to list user source files: %w", err)
+	}
+
+	return len(sourcesFiles) > 0, nil
+}
+
+func (m *moduleRuntimeContainer) addInitTemplateIfNoUserFile(ctx context.Context) (*moduleRuntimeContainer, error) {
+	// Check if there's any user source files, if not, add the template file.
+	// NOTE: This should be moved in a `Init` function once we improve the SDK interface.
+	if hasSourceFiles, err := m.hasUserSourceFiles(ctx); err != nil {
+		return nil, err
+	} else if !hasSourceFiles {
+		return m.withInitTemplate(), nil
+	}
+	return m, nil
+}
+
 // Add the entrypoint to the container runtime so it can be called by the engine.
 func (m *moduleRuntimeContainer) withEntrypoint() *moduleRuntimeContainer {
 	m.ctr = m.ctr.
```
</details>

eunomie:
🤔 I don't know why I can't add a reply to some comments...

@TomChv https://github.com/dagger/dagger/pull/10584#discussion_r2301561778

Well, this is the code you added :-)

cmd/dagger/module.go
eunomie: It's to allow to have two different flags on `dagger develop`: `--with-self-calls` and `--without-self-calls` to let the user enable/disable the feature.

core/integration/module_go_test.go
eunomie: Yes, we can disable them on a module, for instance if we see bad performances.
It's also doable by just editing `dagger.json`

core/modules/config.go
eunomie: Yeah why not for `[]string`. The map has the advantage to directly query with the key instead of looping over the array, but as we don't expect a lot of keys, that might work too.

core/modulesource.go
```diff
@@ -811,3 +819,32 @@ func (fs ModuleSourceStatFS) Stat(ctx context.Context, path string) (*fsutiltype
...
 }
+
+type ModuleSourceExperimentalFeature string
+
+func (f ModuleSourceExperimentalFeature) String() string { return string(f) }
+
+var ModuleSourceExperimentalFeatures = dagql.NewEnum[ModuleSourceExperimentalFeature]()
+
+var (
+	ModuleSourceExperimentalFeatureSelfCalls = ModuleSourceExperimentalFeatures.Register("SELF_CALLS_FEATURE", "Self calls")
```
jedevc:
```suggestion
	ModuleSourceExperimentalFeatureSelfCalls = ModuleSourceExperimentalFeatures.Register("SELF_CALLS", "Self calls")
```

Avoids the codegen stutter of `ExperimentalFeatures.SELF_CALLS_FEATURE`

cmd/codegen/generator/go/templates/module_types.go
jedevc: Mm, it's for going in the `dagger.gen.go` - but does that `""` empty method ever get called now? Since the introspection is being done through the new `TypeDef`s method?

sdk/typescript/runtime/main.go
jedevc: Just because now the definition of `moduleIDPath` now needs to go both *here* and in the binary. It's fiddly to change - it's a magic string, we can just have it one place (similar to how it's done in python).

sdk/typescript/runtime/main.go
TomChv: Make sense

core/sdk/go_sdk.go
```diff
@@ -36,6 +37,10 @@ type goSDK struct {
 	rawConfig map[string]any
 }
 
+func (sdk *goSDK) HasModuleTypeDefs() bool {
```
TomChv: This is not used anymore now?

eunomie: Nope, not needed anymore with the `AsTypeDefs`. I forgot to remove this one, will do it.
The check (for all SDK except go) on the presence of the SDK function `sdk.funcs["moduleTypeDefs"]` has moved inside `sdk.AsTypeDefs`. That makes things simpler, that was good suggestion 👍

cmd/codegen/generator/go/templates/module_types.go
eunomie: The `""` empty method is still used today, in case self calls are not enabled. So we have to keep it for now. It's a bit of a duplication, but the goal is to fix performance issues then only go through `TypeDef` so we can remove the previous version.

eunomie: I renamed `TypeDefObject` to `TypeDef`

core/sdk.go
eunomie: Done 👍

sdk/go/fs.go
eunomie: Yes, the idea is typedefs should only require the core types + module source code to be generated. They do not need dependencies for instance. By using the already generated code, we avoid to generate the sdk just for the typedefs.

core/sdk/module_runtime.go
eunomie: Done 👍

core/sdk/module_runtime.go
eunomie: Fixed 👍

core/schema/modulesource.go
eunomie: Improved with the `AsXXX` function 👍

core/modulesource.go
eunomie: Done 👍

core/sdk/go_sdk.go
eunomie: re-removed 🫣

shykes:
Follow-up from live design discussion.

--> What's the status of performance for this PR? It's been discussed earlier in the thread, but the PR has evolved a lot since then...

eunomie:
I'll need more data points, but here are the first results I have: with this PR, on Go modules, **without** self calls, this PR is faster on a `dagger call` than the `v0.19.0`.
On the reasons why:
- the changes mostly split the two responsibilities of the `containerRuntime` into two different functions
    - it doesn't add something new, in the end the same tasks are performed, to get the types exposes, and to invoke functions
    - to have similar performance sounds reasonable
- the new `moduleDefs` function, corresponding to one branch of the previous `containerRuntime` as you can see in the generated `dagger.gen.go` is using less code generation (this branch is not generated anymore) and calling dagger client directly. This could have a positive impact on performances
- the cost of running the `moduleDefs` on its own compared to running it as part of `containerRuntime` looks like to be compensated by the warming of some cache, like the go dependencies one. That way when it's time to run the `containerRuntime` to generate the runnable code things are quicker

That's a first take on performances, more detailed date soon.

TomChv:
Hey, that's the benchmark I have so far for Go, TypeScript and Python

*NOTE: I had a little issue with trace marker, sometime the output didn't contains the finished trace, specially for Go-dev so I also included the total time the command took to run in these case*)

**Go** -> Dev looks faster 

```
                               │    go.txt     │                go-dev.txt                │
                               │     s/op      │      s/op       vs base                  │
BenchSdk/functions/loadModule      4.400 ±  2%    2.800 ±   ∞ ¹        ~ (p=0.182 n=10+1)
BenchSdk/functions/cmdDuration     5.300 ±  4%    3.400 ±  3%    -35.85% (p=0.000 n=10)
BenchSdk/call/loadModule           4.400 ±  2%    2.750 ±   ∞ ¹  -37.50% (p=0.030 n=10+2)
BenchSdk/call/containerEcho       500.0m ± 20%   600.0m ± 17%    +20.00% (p=0.007 n=10)
BenchSdk/call/cmdDuration          5.800 ±  2%    4.000 ±  5%    -31.03% (p=0.000 n=10)
BenchSdk/develop/develop         3000.0m ±  0%   900.0m ± 11%    -70.00% (p=0.000 n=10)
BenchSdk/develop/cmdDuration       4.250 ±  4%    2.600 ±  4%    -38.82% (p=0.000 n=10)
geomean                            3.246          2.040          -37.15%
```

**TypeScript** -> Dev is slower

```
                              │ typescript.txt │          typescript-dev.txt          │
                              │      s/op      │    s/op     vs base                  │
BenchSdk/call/loadModule            13.45 ± 3%   23.30 ± 2%  +73.23% (p=0.000 n=10+9)
BenchSdk/call/containerEcho         2.400 ± 4%   2.700 ± 4%  +12.50% (p=0.001 n=10)
BenchSdk/develop/develop            6.200 ± 2%   8.100 ± 7%  +30.65% (p=0.000 n=10)
BenchSdk/functions/loadModule       13.00 ± 1%   23.20 ± 4%  +78.46% (p=0.000 n=10+9)
geomean                             7.142        10.43       +46.00%
```

**Python** -> Dev is slower

```
                              │ python.txt │           python-dev.txt            │
                              │    s/op    │    s/op      vs base                │
BenchSdk/develop/develop        3.900 ± 8%   4.650 ± 14%  +19.23% (p=0.000 n=10)
BenchSdk/functions/loadModule   14.05 ± 3%   22.55 ±  4%  +60.50% (p=0.000 n=10)
BenchSdk/call/loadModule        13.95 ± 3%   22.30 ±  3%  +59.86% (p=0.000 n=10)
BenchSdk/call/containerEcho     1.500 ± 7%   1.600 ±  6%   +6.67% (p=0.038 n=10)
geomean                         5.819        7.821        +34.40%
```

The result on TypeScript & Python make sense because we didn't implement any optimisation that were unlocked by that PR, we simply implemented the interface.
These will be done in follow ups.

eunomie:
Asking for comments and opinions.
With the current implementation, the `moduleDefs` function implemented by a SDK returns a container with an entrypoint. Then the engine execute this entrypoint and reads the produced module ID representing the types exposed.

This is done by passing in input a file path where to write the module ID, then the engine will read the file content and load the module.

A simplified version of the code looks like this:

```go
dag.Select(ignoreCtx, ctr, &modDefsID,
	dagql.Selector{
		Field: "withExec",
		Args: []dagql.NamedInput{
			{Name: "args", Value: dagql.ArrayInput[dagql.String]{}},
			{Name: "useEntrypoint", Value: dagql.NewBoolean(true)},
			{Name: "experimentalPrivilegedNesting", Value: dagql.NewBoolean(true)},
			{Name: "execMD", Value: dagql.NewSerializedString(&execMD)},
		},
	},
	dagql.Selector{
		Field: "file",
		Args: []dagql.NamedInput{{Name:  "path", Value: dagql.NewString(moduleIDPath)}},
	},
	dagql.Selector{Field: "contents"},
)

var modID core.ModuleID
json.Unmarshal([]byte(modDefsID), &modID)

dag.Select(ctx, dag.Root(), &inst,
	dagql.Selector{
		Field: "loadModuleFromID",
		Args: []dagql.NamedInput{{Name:  "id", Value: dagql.NewID[*core.Module](modID.ID())}},
	})
```

But while looking one more time to this, it really looks like I'm reinventing the function call and the `returnValue`, `returnError` in some way.

Like this part of the generated `dagger.gen.go`:

```go
fnCall := dag.CurrentFunctionCall()

result, err := invoke(ctx, []byte(parentJson), parentName, fnName, inputArgs)
resultBytes, err := json.Marshal(result)

if err := fnCall.ReturnValue(ctx, dagger.JSON(resultBytes)); err != nil {
	return fmt.Errorf("store return value: %w", err)
}
```

One of the benefit of the current implementation of `moduleDefs` is the calls to dagger API are made directly by the SDK/codegen. It doesn't require to generate some code then to build it before to run it.

But in some way I have the feeling to have reinvented a poor solution of what's already existing with function call. Even if that's not really a function call as one of the benefit of the PR is to clearly separate between getting the types and calling functions.
Does that make sense to keep it that way?

eunomie:
For performances reason and as a way to unblock this PR, the new `moduleTypes` behavior has been disabled for typescript and python SDKs. They will come back as dedicated follow up PR that will be easier to write and review.
That way, the new beahavior can still be included, go SDK performances are good, and we can iterate on it for the other SDKs.

grouville (approved):
LGTM ✅ Let's follow-up with all the neats + python and ts / sdk port to that new implementation

PR #10586, opened 2025-06-13
add support for internal-only args; use for dag-op

sipsma:
Another spin-off from the [persistent dagql cache work](https://github.com/dagger/dagger/pull/10497)

---

This adds support for marking arguments to fields as internal, in which case they can only be set by internal server-side self calls and never by external clients in graphql requests. They won't show up in introspection schema requests either, so they never make it to SDKs or docs.

We have wanted this in a variety of situations for a while, but the immediate motivation here is to update DagOp to use it to distinguish the "outer" lazy implementation and the "inner" call that actually triggers the execution of the operation.

Previously we have accomplished that by plumbing things through context, but this no longer worked with the persistent dagql cache. That new caching implementation has logic such that even when digests of a call don't match, you can still get cache hits based on the inputs all being equivalent. Since inner/outer DagOp calls had the exact same inputs to the exact same field, they started incorrectly hitting cache for each other.

Now, DagOps have internal args that indicate whether the real/inner operation should be executed unlazily and other metadata previously plumbed just through the LLB op. This fixes the persistent cache use case since now there are actual inputs to the call that distinguish the outer+inner calls.

Besides the need from the persistent cache, this approach feels like a good step in that it moves just a bit more logic/data away from LLB and into "dagger-native" code (i.e. into our actual API schema).

This should be generally useful too for other situations where we have been forced to add args that really shouldn't be public in the interest of supporting internal server self calls, though this commit only updates DagOp-ified APIs for now.

---

TODO:
- [ ] I'll give a quick shot at updating the `namespace` arg to `CacheVolume` to use this too since it's one of those APIs we didn't actually want to be public

### DIFF
--- a/core/dagop.go
+++ b/core/dagop.go
@@ -29,24 +29,6 @@ func init() {
 	buildkit.RegisterCustomOp(ContainerDagOp{})
 }
 
-type dagOpContextKey string
-
-func withDagOpContext(ctx context.Context, op buildkit.CustomOp) context.Context {
-	return context.WithValue(ctx, dagOpContextKey(op.Name()), op)
-}
-
-func DagOpFromContext[T buildkit.CustomOp](ctx context.Context) (t T, ok bool) {
-	if val := ctx.Value(dagOpContextKey(t.Name())); val != nil {
-		t, ok = val.(T)
-	}
-	return t, ok
-}
-
-func DagOpInContext[T buildkit.CustomOp](ctx context.Context) bool {
-	_, ok := DagOpFromContext[T](ctx)
-	return ok
-}
-
 // NewDirectoryDagOp takes a target ID for a Directory, and returns a Directory
 // for it, computing the actual dagql query inside a buildkit operation, which
 // allows for efficiently caching the result.
@@ -113,13 +95,8 @@ type FSDagOp struct {
 	// dagql running inside a dagop to determine where it should write data.
 	Path string
 
-	// Data is any additional data that should be passed to the dagop. It does
-	// not contribute to the cache key.
-	Data any
-
-	// utility values set in the context of an Exec
-	g   bksession.Group
-	opt buildkit.OpOpts
+	// Data is any additional data that should be passed to the dagop.
+	Data string
 }
 
 func (op FSDagOp) Name() string {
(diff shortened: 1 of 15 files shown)


core/dagop.go
```diff
@@ -718,28 +689,3 @@ func newDagOpLLB(ctx context.Context, dagOp buildkit.CustomOp, id *call.ID, inpu
 		buildkit.WithPassthrough(),
 	)
 }
-
-// loadDagOpID loads the target ID from the server.
-//
-// It takes two contexts, the first to load the receiver with, the second to
-// load the current call with. This *ensures* that we aren't accidentally
-// applying the dagop to multiple calls - only the most recent.
-func loadDagOpID(ctx context.Context, loadCtx context.Context, srv *dagql.Server, id *call.ID) (dagql.Typed, error) {
```
sipsma, line 727: @jedevc this was a surprise nice side effect; we can get rid of the little "dance of two contexts" here since now whether something should run in dag-op context is determined by args in the call rather than the actual `context.Context`

core/schema/cache.go
```diff
@@ -37,7 +36,7 @@ func (s *cacheSchema) Dependencies() []SchemaResolvers {
 
 type cacheArgs struct {
 	Key       string
-	Namespace string `default:""`
+	Namespace string `internal:"true" default:""`
```
jedevc, line 39: This is *probably* worth adding a changelog for :cry: 

Yeah, it was always designed as an internal arg, but I think I have seen a couple cases of folks attempting to use it. No need to do any backwards compat, but should probably be in the changelog at least!

core/dagop.go
jedevc: Nice! Yes, agreed, this is a much better way of doing it :grin:

sipsma:
> I'll be honest, I don't fully understand how/why this fixes the inputs thing from the pure dagql cache, but I'll take your word for it 😁

@jedevc It's basically that before this change:
1. The "outer" dagop call would look like `foo(bar: "abc")`, with returned result digest `xxh3:123123`
2. The "inner" dagop cal would look like `foo(bar: "abc")`, with returned result digest `xxh3:456456`

The new cache logic is fancy enough to see "oh look, two calls to the same field with the exact same inputs, that's a cache hit", even when the exact return result digest doesn't match (since we handle multiple digests for a given result more robustly now).

But after this change:
1. The "outer" dagop call would look like `foo(bar: "abc", isDagOp: false)`, with returned result digest `xxh3:123123`
2. The "inner" dagop cal would look like `foo(bar: "abc", isDagOp: true)`, with returned result digest `xxh3:456456`

So it can actually see that the calls are distinct now and doesn't try to share cache for them.

jedevc:
Right! Thanks, that clears it up a lot :tada:

core/schema/cache.go
sipsma: You're right, added!

PR #10593, opened 2025-06-16
ci: ensure cni binaries are built with latest go version

jedevc:
Fixes the failing `scan` check on main, which is detecting that we're building our CNI binaries with an old go version.

To do this, we need to update alpine, since the older version we're on doesn't have a supported version of go.

### DIFF
--- a/cmd/engine/.dagger/build/builder.go
+++ b/cmd/engine/.dagger/build/builder.go
@@ -129,6 +129,7 @@ func (build *Builder) Engine(ctx context.Context) (*dagger.Container, error) {
 			Alpine(dagger.AlpineOpts{
 				Branch: consts.AlpineVersion,
 				Packages: []string{
+					"ca-certificates",
 					// for Buildkit
 					"git", "openssh-client", "pigz", "xz",
 					// for CNI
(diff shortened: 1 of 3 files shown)


cmd/engine/.dagger/build/builder.go
```diff
@@ -139,15 +139,15 @@ func (build *Builder) Engine(ctx context.Context) (*dagger.Container, error) {
...
-				ln -s /sbin/iptables-legacy /usr/sbin/iptables
-				ln -s /sbin/iptables-legacy-save /usr/sbin/iptables-save
-				ln -s /sbin/iptables-legacy-restore /usr/sbin/iptables-restore
-				ln -s /sbin/ip6tables-legacy /usr/sbin/ip6tables
-				ln -s /sbin/ip6tables-legacy-save /usr/sbin/ip6tables-save
-				ln -s /sbin/ip6tables-legacy-restore /usr/sbin/ip6tables-restore
-			`})
+			WithExec([]string{"sh", "-c", strings.Join([]string{
+				"mkdir -p /usr/local/sbin",
+				"ln -s /usr/sbin/iptables-legacy /usr/local/sbin/iptables",
```
jedevc, line 145: Alpine 3.22 moves a lot of binaries into `/usr/sbin`, so we make our links in `/usr/local/bin` (which has more path priority, and apk should never touch).

See https://github.com/dagger/dagger/pull/7670 for context, fyi @marcosnils

sipsma:
Merging now to green-ify CI

Issue #8383, opened 2024-09-09
✨ Support multi line `+ignore` directive


wingyplus:
### What are you trying to do?

Currently, `+ignore` doesn't support multi-line declaration so I need to declare it on a single line. What I want to purpose is make it support multi-line. For example:

```go
// +ignore=[
//      "**/*",
//      "!generated/", 
//      "!src/", 
//      "!scripts/", 
//      "!composer.json", 
//      "!composer.lock", 
//      "!LICENSE", 
//      "!README.md"
//  ]
```

### Why is this important to you?

Single-line ignore is hard to read when they're grow larger. :(

helderco:
I asked for this here:
- https://github.com/dagger/dagger/pull/7744#issuecomment-2274122714

Feedback:

> > > I need to check how I can parse that but that must be possible.
> 
> Hm, you want `parsePragmaComment`. This is tricky, currently it's all assumed in one line with a regexp.
> 
> Being able to do opening/closing braces means the language is no longer regular, so we won't be able to use regexp for it. I'm not confident there's a simple way to do it (but I'm happy to do it in a follow-up) - I don't think we should delay this PR for that though.

wingyplus:
@helderco Yeah, I open it after discussed with Justin at https://discord.com/channels/707636530424053791/708371226174685314/1282767180311695382.

PR #10594, opened 2025-06-16
feat(go): allow multiline pragmas

jedevc:
Fix for https://github.com/dagger/dagger/issues/8383.

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -3842,7 +3842,10 @@ import (
 type Test struct {}
 
 func (t *Test) Call(
-  //+ignore=["foo.txt", "bar"]
+  // +ignore=[
+  //   "foo.txt",
+  //   "bar"
+  // ]
   dir *dagger.Directory,
 ) *dagger.Directory {
  return dir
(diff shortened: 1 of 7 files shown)


nipuna-perera:
Would it be possible to support backticks instead of `"` for multi line? It's a bit awkward to use `"` for multiline because that's not exactly valid or idiomatic go.

jedevc:
The parser for the pragma values is using JSON (just like it does today) - so you can't have multi-line strings.

Allowing multi-line strings and *everything* like that isn't quite in scope right now, I just wanted to make sure we could support multi-line arrays (since those are really horrible).

nipuna-perera:
Fair enough. Would documentation be done as a separate PR?

nipuna-perera (commented):
lgtm

core/integration/module_test.go
```diff
@@ -3841,7 +3841,10 @@ import (
 type Test struct {}
 
 func (t *Test) Call(
-  //+ignore=["foo.txt", "bar"]
+  // +ignore=[
+  //   "foo.txt",
+  //  "bar"
```
helderco: Not important, just stands out to me 🙂 
```suggestion
  //   "bar"
```

helderco (approved):
👍

PR #10595, opened 2025-06-16
fix: raise up --no-mod flag to same level as --mod

jedevc:
Revival of https://github.com/dagger/dagger/pull/10191 - apparently, once you force push to a closed PR, you *can't* re-open it. How wonderful.

It makes sense that the `--no-mod` flag should be allowed everywhere that the `--mod` flag is.

The main rationale for this was around `dagger query`, where it wasn't possible to run without attempting to load the current directory. But it also extends to other commands.


jedevc:
Hm @helderco looks like implementing your suggestion https://github.com/dagger/dagger/pull/10191#discussion_r2069104587 for changing `PersistentFlags` to `Flags` causes this test to fail: https://dagger.cloud/dagger/traces/b0774de1bcbbeca0154b2d57487449b2?span=cf99e357ac7762f0

Would we want to allow this case now?

helderco:
> Hm @helderco looks like implementing your suggestion [#10191 (comment)](https://github.com/dagger/dagger/pull/10191#discussion_r2069104587) for changing `PersistentFlags` to `Flags` causes this test to fail: https://dagger.cloud/dagger/traces/b0774de1bcbbeca0154b2d57487449b2?span=cf99e357ac7762f0
> 
> Would we want to allow this case now?

No worries, that'll be fixed/handled in https://github.com/dagger/dagger/pull/10477 so you can keep it persistent for now.

helderco (approved):
lgtm, thanks! 👍

PR #10599, opened 2025-06-17
fix: ensure we compute the same cache key for different git auth

jedevc:
Follow up to https://github.com/dagger/dagger/pull/10271.

See @sipsma's comment here: https://discord.com/channels/707636530424053791/1382699979805556798/1384363853764104233


marcosnils (approved):
:ship:

PR #10600, opened 2025-06-17
python: support defining enum member descriptions using docstrings in modules

helderco:
Been meaning to do this for a while. Rather than requiring our custom class for descriptions, users can use docstrings now:

## Before
```python
import dagger

@dagger.enum_type
class Options(dagger.Enum):
    ONE = "ONE", "The first value"
    TWO = "TWO", "The second value"
```

## After
```python
import enum
import dagger

@dagger.enum_type
class Options(enum.Enum):
    ONE = "ONE"
    """The first value"""
    
    TWO = "TWO"
    """The second value"""
```

### DIFF
--- a/sdk/python/dev/src/python_sdk_dev/docs.py
+++ b/sdk/python/dev/src/python_sdk_dev/docs.py
@@ -32,7 +32,6 @@ class Docs:
         """Build and preview the documentation in the browser."""
         return (
             self.container.with_(mounted_workdir(self.build()))
-            .with_exec(["uv", "run", "python", "-m", "http.server", str(bind)])
             .with_exposed_port(bind)
-            .as_service()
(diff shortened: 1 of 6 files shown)


sdk/python/dev/src/python_sdk_dev/docs.py
```diff
@@ -32,7 +32,6 @@ def preview(
         """Build and preview the documentation in the browser."""
         return (
             self.container.with_(mounted_workdir(self.build()))
-            .with_exec(["uv", "run", "python", "-m", "http.server", str(bind)])
             .with_exposed_port(bind)
-            .as_service()
+            .as_service(args=["uv", "run", "python", "-m", "http.server", str(bind)])
```
jedevc, line 36: BEGONE :tada:

Thought we got all of these ages ago, I'm so glad we deprecated this pattern :grin:

PR #10606, opened 2025-06-18
Avoid telemetry errors on expected failures

jedevc:
Mostly in the `drivers` package, but also a small cleanup in the git backend.

### DIFF
--- a/engine/client/drivers/docker.go
+++ b/engine/client/drivers/docker.go
@@ -3,6 +3,7 @@ package drivers
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -20,7 +21,6 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	connh "github.com/moby/buildkit/client/connhelper"
 	connhDocker "github.com/moby/buildkit/client/connhelper/dockercontainer"
-	"github.com/pkg/errors"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/trace"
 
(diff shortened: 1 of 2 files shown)


engine/client/drivers/docker.go
```diff
@@ -120,12 +120,12 @@ func (d *dockerDriver) create(ctx context.Context, imageRef string, containerNam
 	}
 
 	// ensure the image is pulled
-	if _, err := traceExec(ctx, exec.CommandContext(ctx, "docker", "inspect", "--type=image", imageRef), telemetry.Encapsulated()); err != nil {
+	if _, err := traceExec(ctx, exec.CommandContext(ctx, "docker", "inspect", "--type=image", imageRef), true, telemetry.Encapsulated()); err != nil {
```
marcosnils, line 123: this SGTM initially. The only concern that I have is that if this `docker inspect` fails for some other reason besides not finding the images (docker engine connection error), then the trace will be marked as non-errored incorrectly. 

Having said that, I understand it should be rare given that `collectLeftOverEngines` is called above which should error first

marcosnils (approved):

vito (commented):
IMO, if a span looks like it's representing 1:1 a command running, its status should just reflect the exit status of that command, whether it's expected or not. If I look at this in the trace I'm going to see that it ran `docker inspect`, which succeeded, but then see it go on to `pull` anyhow, and think there's a logic error.

The relevant span even marks itself `encapsulated` so it seems like a bug that a user saw it in the first place.

https://github.com/dagger/dagger/pull/10608 fixes what I think is the root cause - the `connect` span and other enclosing spans should have been setting their status to `OK` instead of `UNSET`

PR #10607, opened 2025-06-18
fix: avoid mutating global slice TypesHiddenFromModuleSDKs

jedevc:
Follow-up from https://github.com/dagger/dagger/pull/10326.

Not sure if this was ever practically getting hit, but was potentially a bug waiting to happen :thinking:


TomChv (approved):
MB, thanks for the fix

PR #10605, opened 2025-06-18
Introduce httpAuthUsername

skycaptain:
Continuation of #10427.

PR #10613, opened 2025-06-19
Pass username from credential helper 

### LINKED ISSUE #10612: Passing username from credential helper to Git.httpAuthUsername does not work
> As far as I can tell, this only adds the internal capability of the `httpAuthUsername` to the general code, but not the general capability of reading it from git, right? 
> 
> I am trying to clone private repositories in GitLab with the CI_JOB_TOKEN (I don't want to create a long-lived SSH key), with the following .gitlab-ci.yaml example
> ```yaml
> .dagger:
>   extends: [.docker]
>   before_script:
>     - apk add curl
>     - curl -fsSL https://dl.dagger.io/dagger/install.sh | DAGGER_COMMIT=head BIN_DIR=/usr/local/bin sh
>     - export MODULE_URL="..."
>     - echo -e "https://gitlab-ci-token:$CI_JOB_TOKEN@some.host" > ~/.git-credentials
>     - git config --global credential.helper store
>     - git clone $MODULE_URL # Works flawlessly
>     - dagger call ...
> ```

_Originally posted by @daniel-naegele in https://github.com/dagger/dagger/issues/10605#issuecomment-2985933172_

skycaptain:
See https://github.com/dagger/dagger/pull/10605#issuecomment-2985933172 where this was reported first.

With #10605 we introduced `Git.httpAuthUsername`; however, during refactoring, the part where the username was passed on from the credential helper was accidentally removed. This PR adds it back.

During testing, I encountered the following error:

```
failed to resolve git src: failed to resolve git src: select: select: failed to compute cache key for Query.git: assign input "httpAuthUsername" (string) as {Value:x-token-auth Valid:true} (dagql.Optional
com/dagger/dagger/dagql.String]): assign: cannot assign dagql.Optional[github.com/dagger/dagger/dagql.String] to string
```

I couldn't find any solution other than using `dagql.String` again. This approach was specifically requested not to be used by @jedevc during review of [#10605](https://github.com/dagger/dagger/pull/10605#discussion_r2154294672). Perhaps this was why @grouville used `dagql.String` in the first place? Thus, I reverted that bit from the previous PR and added the fix on top.

Closes #10612.


skycaptain:
Let me know if there is a proper solution for the `string`/`daqgl.String` situation.

jedevc:
Pushed some commits, with some minor fixups (some related, some not, see commit messages for more detail).

The main bit is untangling the `string`/`dagql.String` stuff (looks like the `sshKnownHosts` part that this was similar to was just wrong in the first place, so fixed that too).

I also made sure we had tests for this, making sure that credentials are properly loaded from the client - we didn't have tests for this really at all (see my comment in https://github.com/dagger/dagger/pull/10427#issuecomment-2894426502), so I've added these.

skycaptain:
> Pushed some commits, with some minor fixups (some related, some not, see commit messages for more detail).

Many thanks. Looks great! Seems to work from what I can tell, testing locally.

> The main bit is untangling the `string`/`dagql.String` stuff (looks like the `sshKnownHosts` part that this was similar to was just wrong in the first place, so fixed that too).

Yes, that `sshKnownHosts` part really tricked me. Why should that part work and `httpAuthUsername` following the exact same pattern would not. Good, this issue is now alos fixed

> I also made sure we had tests for this, making sure that credentials are properly loaded from the client - we didn't have tests for this really at all (see my comment in [#10427 (comment)](https://github.com/dagger/dagger/pull/10427#issuecomment-2894426502)), so I've added these.

I was just about to push the tests I wrote this morning when I saw your pushes. Yours are much more comprehensive. So thanks for that as well.

daniel-naegele:
Hey folks I know you are fast and I don't want nag/annoy, but is there an ETA on when this PR is merged? Or is there a way to specify in the install script to use another branch/repo and build from source? This is kinda urgent for me as I am currently writing my bachelors thesis around Dagger as its main tool and need a proper solution to pull other repos in a GitLab environment, SSH Keys and PATs are for me and our GitLab setup not a really viable workaround...

jedevc:
@daniel-naegele the last activity here was less than 24 hours ago :grin: it's moving, don't worry! It's also in the next release milestone, so we'll it'll be part of our next release sometime next week (exact datetime tbd).

I'm just waiting for @grouville to have a chance to review :grin:

grouville (approved):
Thank you ! I'm not familiar with these views, learning along the way 🙏 

I also finally understand the comment you made about testing the loading of credentials for git in a self contained test 🙏

PR #10617, opened 2025-06-20
cli: new `dots` frontend, support `$DAGGER_PROGRESS`

vito:
[asciicast](https://asciinema.org/a/723921)


jedevc (approved):
Code LGTM :grin: I'll be honest, I don't *personally* love dots-style output, but always nice to support formats that everyone's familiar with :tada:

(also neaaaat, the actual frontend implementation is so much shorter than I thought it would have to be)

Issue #7789, opened 2024-06-29
engine: Impure APIs don't work with lists


sipsma:
Another tangential issue I noticed during adventures with https://github.com/dagger/dagger/pull/7767

An impure dagql API that has a list return value in the chain does not appear to work correctly.

Because it's impure, the query will be re-evaluated if submitted to the server again. However, if there were selections on the list, then the ID for the query will still include `nth` fields in that ID, *but using the nth values from the first call to the API*.

That seems like a bug because when you re-run the impure API:
1. there's no guarantee the order of the results are the same, so you could be selecting nth on different results every time
2. there's no guarantee the size of the list of the results are the same, so you could be trying to select an nth value beyond the size of the list

I don't know if there's any core APIs today that have this problem, but I did hit it in #7767. While I had a workaround there that I ended up wanting to follow for additional reasons, this seems like sort of a ticking time bomb for when we want an API with this behavior.

cc @vito

sipsma:
Believe this was fixed by https://github.com/dagger/dagger/pull/10410 and https://github.com/dagger/dagger/pull/10620. Now dagql lists are evaluated and then elements selected from that list. Also, since this issue was created `Impure` was broken down into finer grain cache keys; the closest equivalent is `DoNotCache` but that should only evaluate the list as a whole once per-query-submission now.

PR #10620, opened 2025-06-21
dagql: generalize and unify around `Result` (formerly known as `Instance`)

sipsma:
Another big carve-out from [persistent cache work](https://github.com/dagger/dagger/pull/10497), this change:
1. Generalizes (and renames) `dagql.Instance` to be able to wrap any type (whereas previously it could only wrap objects)
   * `Instance` is renamed to -> `Result`, which is a bit more clear IMO especially once it is used in the persistent cache PR for also wrapping "cache results"
   * Additionally, there are now two types: `dagql.Result` and `dagql.ObjectResult` (which is a `dagql.Result` plus some more object-specific methods)
   * I *wanted* to make these interfaces, but Go reflection that we use for figuring out API types is just not really usable with interfaces that have a type param, despite many attempts to find a path.
2. Does some refactoring so that internally all field implementations return an `Result`, and various impacted parts of the codebase

The motivation is that the persistent cache was currently introducing even more wrapper types to support carrying around cache-related metadata (i.e. ID in db, etc.), which was rapidly devolving into insanity mixed in with all our existing types, wrappers, etc.

I decided to centralize everything there around the existing `Instance` types, which makes sense since `Instance` is already a wrapper for ID-metadata, which is very related to the cache metadata.

However, doing that required very substantial updates, most important support for wrapping any type, not just objects.
* i.e. if a field returns a `String`, we still want to be able to cache it and reference it, which in this new world means it needs to be wrapped as an `Instance`.

@jedevc mentioned that he had wanted `Instance` to be able to wrap any type for telemetry-related purposes too, so hopefully this should support that too.

### DIFF
--- a/core/interface.go
+++ b/core/interface.go
@@ -188,7 +188,7 @@ func (iface *InterfaceType) Install(ctx context.Context, dag *dagql.Server) erro
 	ctx = bklog.WithLogger(ctx, bklog.G(ctx).WithField("interface", iface.typeDef.Name))
 	slog.ExtraDebug("installing interface")
 
-	if iface.mod.InstanceID == nil {
+	if iface.mod.ResultID == nil {
 		return fmt.Errorf("installing interface %q too early", iface.typeDef.Name)
 	}
 	class := dagql.NewClass(dag, dagql.ClassOpts[*InterfaceAnnotatedValue]{
--- a/core/module.go
+++ b/core/module.go
@@ -52,8 +52,8 @@ type Module struct {
 	// The module's enumerations
 	EnumDefs []*TypeDef `field:"true" name:"enums" doc:"Enumerations served by this module."`
 
-	// InstanceID is the ID of the initialized module.
-	InstanceID *call.ID
+	// ResultID is the ID of the initialized module.
+	ResultID *call.ID
 }
 
 func (*Module) Type() *ast.Type {
@@ -106,7 +106,7 @@ func (mod *Module) IDModule() *call.Module {
 		panic(fmt.Sprintf("unexpected module source kind %q", mod.Source.Self().Kind))
 	}
 
-	return call.NewModule(mod.InstanceID, mod.Name(), ref, pin)
+	return call.NewModule(mod.ResultID, mod.Name(), ref, pin)
 }
 
 func (mod *Module) Evaluate(context.Context) (*buildkit.Result, error) {
--- a/core/schema/service.go
+++ b/core/schema/service.go
@@ -215,7 +215,7 @@ func (s *serviceSchema) containerAsService(ctx context.Context, parent *core.Con
 func (s *serviceSchema) containerUp(ctx context.Context, ctr dagql.ObjectResult[*core.Container], args struct {
 	UpArgs
 	core.ContainerAsServiceArgs
-}) (res dagql.Result[dagql.Nullable[core.Void]], _ error) {
+}) (res dagql.Nullable[core.Void], _ error) {
 	var inputs []dagql.NamedInput
 	if args.Args != nil {
 		inputs = append(inputs, dagql.NamedInput{
@@ -269,7 +269,7 @@ func (s *serviceSchema) containerUp(ctx context.Context, ctr dagql.ObjectResult[
 	return s.up(ctx, svc, args.UpArgs)
 }
 
-func (s *serviceSchema) containerUpLegacy(ctx context.Context, ctr dagql.ObjectResult[*core.Container], args UpArgs) (res dagql.Result[dagql.Nullable[core.Void]], _ error) {
+func (s *serviceSchema) containerUpLegacy(ctx context.Context, ctr dagql.ObjectResult[*core.Container], args UpArgs) (res dagql.Nullable[core.Void], _ error) {
 	var svc dagql.ObjectResult[*core.Service]
 	err := s.srv.Select(ctx, ctr, &svc,
 		dagql.Selector{
@@ -354,7 +354,7 @@ type UpArgs struct {
 
 const InstrumentationLibrary = "dagger.io/engine.schema"
 
-func (s *serviceSchema) up(ctx context.Context, svc dagql.ObjectResult[*core.Service], args UpArgs) (res dagql.Result[dagql.Nullable[core.Void]], _ error) {
+func (s *serviceSchema) up(ctx context.Context, svc dagql.ObjectResult[*core.Service], args UpArgs) (res dagql.Nullable[core.Void], _ error) {
 	void := dagql.Null[core.Void]()
 
 	useNative := !args.Random && len(args.Ports) == 0
@@ -405,5 +405,5 @@ func (s *serviceSchema) up(ctx context.Context, svc dagql.ObjectResult[*core.Ser
 	// wait for the request to be canceled
 	<-ctx.Done()
 
-	return dagql.NewResultForCurrentID(ctx, void)
+	return void, nil
 }
--- a/dagql/dagql_test.go
+++ b/dagql/dagql_test.go
@@ -161,7 +161,7 @@ func TestSelectArray(t *testing.T) {
 	}.Install(srv)
 
 	dagql.Fields[*points.Point]{
-		dagql.Func("instanceNeighbors", func(ctx context.Context, self *points.Point, _ struct{}) (dagql.InstanceArray[*points.Point], error) {
+		dagql.Func("instanceNeighbors", func(ctx context.Context, self *points.Point, _ struct{}) (dagql.ResultArray[*points.Point], error) {
 			var pt0 dagql.Result[*points.Point]
 			err := srv.Select(ctx, srv.Root(), &pt0,
 				dagql.Selector{
@@ -273,7 +273,7 @@ func TestSelectArray(t *testing.T) {
 	})
 
 	t.Run("select all as instance array", func(t *testing.T) {
-		var points dagql.InstanceArray[*points.Point]
+		var points dagql.ResultArray[*points.Point]
 		assert.NilError(t, srv.Select(ctx, srv.Root(), &points,
 			pointSel,
 			dagql.Selector{
@@ -300,7 +300,7 @@ func TestSelectArray(t *testing.T) {
 	})
 
 	t.Run("select all individual instances", func(t *testing.T) {
-		var points dagql.InstanceArray[*points.Point]
+		var points dagql.ResultArray[*points.Point]
 		assert.NilError(t, srv.Select(ctx, srv.Root(), &points,
 			pointSel,
 			dagql.Selector{
@@ -342,7 +342,7 @@ func TestSelectArray(t *testing.T) {
 	})
 
 	t.Run("select all children", func(t *testing.T) {
-		var points dagql.InstanceArray[*points.Point]
+		var points dagql.ResultArray[*points.Point]
 		assert.ErrorContains(t, srv.Select(ctx, srv.Root(), &points,
 			pointSel,
 			dagql.Selector{
@@ -2527,18 +2527,18 @@ func TestServerSelect(t *testing.T) {
 		require.True(t, ok, "TestObject class not found")
 
 		// Create an instance
-		objInstance, err := testObjClass.New(testObj)
+		objResult, err := testObjClass.New(testObj)
 		require.NoError(t, err)
 
 		// Test selecting a simple field
 		var result int
-		err = srv.Select(ctx, objInstance, &result, dagql.Selector{Field: "value"})
+		err = srv.Select(ctx, objResult, &result, dagql.Selector{Field: "value"})
 		require.NoError(t, err)
 		assert.Equal(t, 42, result)
 
 		// Test selecting a string field
 		var textResult string
-		err = srv.Select(ctx, objInstance, &textResult, dagql.Selector{Field: "text"})
+		err = srv.Select(ctx, objResult, &textResult, dagql.Selector{Field: "text"})
 		require.NoError(t, err)
 		assert.Equal(t, "hello", textResult)
 	})
@@ -2557,12 +2557,12 @@ func TestServerSelect(t *testing.T) {
 		require.True(t, ok, "NestedObject class not found")
 
 		// Create an instance
-		objInstance, err := nestedObjClass.New(nestedObj)
+		objResult, err := nestedObjClass.New(nestedObj)
 		require.NoError(t, err)
 
 		// Test selecting through a chain of objects
 		var result int
-		err = srv.Select(ctx, objInstance, &result,
+		err = srv.Select(ctx, objResult, &result,
 			dagql.Selector{Field: "inner"},
 			dagql.Selector{Field: "value"})
 		require.NoError(t, err)
@@ -2581,12 +2581,12 @@ func TestServerSelect(t *testing.T) {
 		require.True(t, ok, "TestObject class not found")
 
 		// Create an instance
-		objInstance, err := testObjClass.New(testObj)
+		objResult, err := testObjClass.New(testObj)
 		require.NoError(t, err)
 
 		// Test selecting a null field
 		var result *string
-		err = srv.Select(ctx, objInstance, &result, dagql.Selector{Field: "nullableField"})
+		err = srv.Select(ctx, objResult, &result, dagql.Selector{Field: "nullableField"})
 		require.NoError(t, err)
 		assert.Assert(t, result == nil)
 	})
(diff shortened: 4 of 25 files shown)


dagql/builtins.go
```diff
@@ -123,6 +134,62 @@ func (d DynamicArrayOutput) SetField(val reflect.Value) error {
 	return nil
 }
 
+type DynamicInstanceArrayOutput struct {
```
jedevc:
```suggestion
type DynamicResultArrayOutput struct {
```

Maybe? There's a few other `Instance` types/functions around, does it make sense to rename all of them to `Result`?

`NewInstanceForCurrentID` springs to mind.

dagql/call/id.go
```diff
@@ -193,6 +193,9 @@ func (id *ID) DisplaySelf() string {
 }
 
 func (id *ID) Display() string {
+	if id == nil {
```
jedevc, line 196: Noticed a lot of these `nil` checks (seem good, very nice to not panic everywhere), just kinda curious what's changed to make them all suddenly necessary.

dagql/dagql_test.go
```diff
@@ -234,16 +353,37 @@ func TestSelectArray(t *testing.T) {
...
 				Nth:   1,
 			},
 		))
-		assert.Equal(t, point.Self.X, 5)
-		assert.Equal(t, point.Self.Y, 7)
+		assert.Equal(t, point.Self().X, 5)
+		assert.Equal(t, point.Self().Y, 7)
+	})
+
+	t.Run("select nth caching", func(t *testing.T) {
```
jedevc, line 370: Just to check my understanding - is this also a fix for https://github.com/dagger/dagger/issues/7789? Seems somewhat similar.

I'm guessing this comes out of where we want to persist these to disk, and we need proper IDs for each element?

modules/gha/steps.go
```diff
@@ -81,6 +81,14 @@ func (j *Job) uploadEngineLogsStep() []api.JobStep {
 				"overwrite": "true",
 			},
 		},
+		/* BREAK GLASS: print kernel logs too
```
jedevc, line 84: My extended sympathies for you here :heart:

dagql/objects.go
```diff
@@ -310,44 +325,66 @@ func (class Class[T]) Call(
...
+	return o.self.Type()
 }
 
-var _ Object = Instance[Typed]{}
-
-// ObjectType returns the ObjectType of the instance.
-func (r Instance[T]) ObjectType() ObjectType {
-	return r.Class
+/*
+func (o instance[T]) Type() *ast.Type {
```
jedevc: :eyes:

dagql/types.go
```diff
@@ -63,19 +63,44 @@ type IDType interface {
...
+	// NthValue returns the Nth value of the wrapped value when the wrapped value
+	// is an Enumerable. If the wrapped value is not Enumerable, it returns an error.
+	NthValue(int) (AnyResult, error)
+
+	// WithPostCall returns a new AnyResult with the given post-call function attached to it.
+	WithPostCall(fn cache.PostCallFunc) AnyResult
+}
+
+// AnyObjectResult is an AnyResult that wraps a selectable value (i.e. a graph object)
+type AnyObjectResult interface {
```
jedevc, line 97: I'm assuming this interface is because later on, we want to introduce a `CachedObjectResult` or similar? And so we need an interface for it?

Asking because, it feels a bit confusing that:
1. There's only one implementer of this interface
2. Code in `core` refers to a mix of `ObjectResult` and `AnyObjectResult` - I guess I'm confused why we export `ObjectResult` at all? Is there a reason why all code outside of `dagql` would need to know if it has a `ObjectResult` or not?

dagql/objects.go
```diff
@@ -310,44 +325,66 @@ func (class Class[T]) Call(
...
+}
+
+func (r Result[T]) SetField(field reflect.Value) error {
+	return assign(field, r.self)
+}
 
 // Unwrap returns the inner value of the instance.
-func (r Instance[T]) Unwrap() Typed {
-	return r.Self
+func (r Result[T]) Unwrap() Typed {
```
jedevc, line 354: Is there a use for `Unwrap`? It seems like it's now the same behavior as `Self`.

core/codegen.go
```diff
@@ -35,9 +36,6 @@ func (*GeneratedCode) TypeDescription() string {
 
 func (code GeneratedCode) Clone() *GeneratedCode {
 	cp := code
-	if cp.Code.Self != nil {
```
vito, line 38: What's this about? :thinking:

core/interface.go
```diff
@@ -389,49 +398,63 @@ func (iface *InterfaceType) Install(ctx context.Context, dag *dagql.Server) erro
...
 }
 
-func wrapIface(ifaceType *InterfaceType, underlyingType ModType, res dagql.Typed) (dagql.Typed, error) {
+func wrapIface(curID *call.ID, ifaceType *InterfaceType, underlyingType ModType, res dagql.AnyResult, srv *dagql.Server) (dagql.AnyResult, error) {
 	switch underlyingType := underlyingType.(type) {
 	case *InterfaceType, *ModuleObjectType:
 		switch res := res.(type) {
-		case *ModuleObject:
-			return &InterfaceAnnotatedValue{
+		case dagql.ObjectResult[*ModuleObject]:
```
jedevc: This switch case feels odd - I guess we can't match on interfaces?

core/schema/env.go
```diff
@@ -118,7 +118,11 @@ func (s environmentSchema) withStringInput(ctx context.Context, env *core.Env, a
 	Value       string
 	Description string
 }) (*core.Env, error) {
-	return env.WithInput(args.Name, dagql.NewString(args.Value), args.Description), nil
+	str, err := dagql.NewInstanceForCurrentID(ctx, dagql.NewString(args.Value))
```
jedevc: Hm, wait, this seems odd? If I evaluated the ID here, I get a `core.Env` not a `dagql.String`.

core/schema/service.go
```diff
@@ -351,12 +354,12 @@ type UpArgs struct {
 
 const InstrumentationLibrary = "dagger.io/engine.schema"
 
-func (s *serviceSchema) up(ctx context.Context, svc dagql.Instance[*core.Service], args UpArgs) (dagql.Nullable[core.Void], error) {
+func (s *serviceSchema) up(ctx context.Context, svc dagql.ObjectResult[*core.Service], args UpArgs) (res dagql.Result[dagql.Nullable[core.Void]], _ error) {
```
jedevc: :thinking: what is this return type :thinking: 

Why is it a nullable void? `None` vs `Some(Void)` feel the same :thinking:

jedevc:
Maybe the PR should be renamed?

dagql/objects.go
sipsma: Yeah, `Self` returns a `T` and is thus only available on the struct `Result[T]`, whereas `Unwrap` returns `Typed` and is thus available on the interface `AnyResult`. We need the `AnyResult` interface for when we pass these around to functions that aren't type-param'd, so 🤷‍♂️

dagql/builtins.go
sipsma: Ah right I forgot to rename those, will fix

dagql/types.go
sipsma: Yeah it makes a difference because there's APIs available to callers outside of `dagql` that specifically want an `ObjectResult` rather than just any `Result`. E.g. `dagql.Select` should only be called on an `ObjectResult`, which is selectable, rather than a `Result` which may or may not be selectable (i.e. if it's a string).

The problem then is that we can't make those sorts of APIs accept a type param, so then we have the interface wrappers of `AnyObjectResult` around `ObjectResult[T]`. That's the reason there's only one implementer of the interface.
* This is actually the same as before this change, where there was `type Instance[T] struct` and `type Object interface`, where `Instance[T]` is the only implementer of `Object`.

EDIT: also, the reason we export both `ObjectResult` and `AnyObjectResult` is that `ObjectResult` is more type-safe because of its type-param, which makes it much easier to use e.g. as the parent field in `NodeFunc`.

dagql/dagql_test.go
sipsma: It's actually the same fix as was made in https://github.com/dagger/dagger/pull/10410, except there I had only fixed it in `dagql.Load`, but realized as part of this PR the same fix needed to be made for `dagql.Select` too.

Good question whether #7789 is fixed as part of this, I had forgotten about that issue. I'll need to check, but maybe?

core/container.go
```diff
@@ -1271,7 +1277,10 @@ func (container *Container) WithGPU(ctx context.Context, gpuOpts ContainerGPUOpt
 	return container, nil
 }
 
-func (container Container) Evaluate(ctx context.Context) (*buildkit.Result, error) {
+func (container *Container) Evaluate(ctx context.Context) (*buildkit.Result, error) {
+	if container == nil {
```
vito, line 1291: Curious about this nil check

core/interface.go
```diff
@@ -120,20 +122,30 @@ func (iface *InterfaceType) loadImpl(ctx context.Context, id *call.ID) (*loadedI
 	return loadedImpl, nil
 }
 
-func (iface *InterfaceType) CollectCoreIDs(ctx context.Context, value dagql.Typed, ids map[digest.Digest]*resource.ID) error {
-	switch value := value.(type) {
-	case dagql.Instance[*InterfaceAnnotatedValue]:
-		mod, ok := value.Self.UnderlyingType.SourceMod().(*Module)
+func (iface *InterfaceType) CollectCoreIDs(ctx context.Context, value dagql.AnyResult, ids map[digest.Digest]*resource.ID) error {
+	if value == nil {
```
vito, line 126: Curious about this too

modules/gha/steps.go
sipsma: Oh yeah, was checking OOMs... Fun story: the problem chain started with "The dagger CLI is getting OOM'd when running `TestContainer`" and ended up root caused at "The Proxy/CA tests were recompiling code too much across many processes, which seems to have become slightly more memory hungry perhaps due to more use of generics in packages imported by the integ test package, which was just enough to start triggering the OOM killer, which chose the CLI for some reason" (which is why https://github.com/dagger/dagger/pull/10646 appeared randomly last week)

core/mcp.go
```diff
@@ -722,15 +722,19 @@ func (m *MCP) returnBuiltin() (LLMTool, bool) {
 					return nil, fmt.Errorf("invalid type for argument %s: %T", name, arg)
 				}
 				if output.ExpectedType == "String" {
-					output.Value = dagql.String(argStr)
+					var err error
+					output.Value, err = dagql.NewInstanceForCurrentID(ctx, dagql.String(argStr))
```
vito: Seems a little odd - I think the current ID will be an LLM unlazying call, the string shouldn't be the result of that I'd guess

core/interface.go
sipsma: Well you can match on interfaces, but the interfaces purposely aren't type-param'd (for reasons mentioned in some of my other comment replies), so then `.Self()` isn't available.

But, now that you mention it, I could just have a `case dagql.AnyResult` and then do a type assertion on `.Unwrap()` to get the `*ModuleObject`, which would be cleaner/less dupe-y. I'll try that.

core/module.go
```diff
@@ -723,10 +723,6 @@ func (mod *Module) PBDefinitions(ctx context.Context) ([]*pb.Definition, error)
 func (mod Module) Clone() *Module {
 	cp := mod
 
-	if mod.Source.Self != nil {
```
vito, line 726: Is this the same reason as elsewhere? I guess the idea is we shouldn't need to clone things that are intentionally held by reference, on principle

dagql/objects.go
sipsma: Oops, forgot to mark this one with my traditional
```go
// TODO: FIX
// TODO: FIX
// TODO: FIX
```
😄 Will fix!

core/schema/env.go
sipsma: Good catch! Will fix

core/schema/coremod.go
```diff
@@ -342,12 +342,17 @@ type CoreModScalar struct {
...
 	if !ok {
 		return nil, fmt.Errorf("CoreModScalar.ConvertFromSDKResult: found no scalar type")
 	}
-	return s.DecodeInput(value)
+
+	input, err := s.DecodeInput(value)
+	if err != nil {
+		return nil, fmt.Errorf("CoreModScalar.ConvertFromSDKResult: failed to decode input: %w", err)
+	}
+	return dagql.NewInstanceForCurrentID(ctx, input)
```
vito: Should this be renamed to `NewResultForCurrentID`?

core/schema/service.go
sipsma: Yeah this was necessary in an earlier iteration for reasons that are not worth detailing, but can be simplified again, will do so

dagql/call/id.go
sipsma: Mostly because as I was iterating on the code I would have bugs that accidentally triggered panics, so even though those are gone it seemed nice to leave in these guards for the long term anyways

core/telemetry.go
```diff
@@ -153,7 +154,11 @@ func logResult(ctx context.Context, res dagql.Typed, self dagql.Object, id *call
 //
 // Effects will become complete as spans appear from Buildkit with a
 // corresponding effect ID.
-func collectEffects(ctx context.Context, res dagql.Typed, span trace.Span, self dagql.Object) {
+func collectEffects(ctx context.Context, res dagql.AnyResult, span trace.Span, self dagql.AnyObjectResult) {
+	if res == nil {
```
vito: Curious about this one too

dagql/types.go
```diff
@@ -86,32 +111,28 @@ type Object interface {
...
-	//
-	// The returned value is the raw Typed value returned from the field; it must
-	// be instantiated with a class for further selection.
-	//
-	// Any Nullable values are automatically unwrapped.
-	ReturnType(context.Context, *Server, Selector) (Typed, *call.ID, error)
+// InterfaceValue is a value that wraps some underlying object with a interface to that object's API. This type exists to support unwrapping it and getting the underlying object.
+type InterfaceValue interface {
+	// UnderlyingObject returns the underlying object of the InterfaceValue
+	UnderlyingObject() (Typed, error)
```
vito, line 123: Interesting, is this new? Always wondered how interfaces would/should/could affect DagQL, and hoped we could find a way to map them to real GraphQL interfaces

core/codegen.go
sipsma: `Self` isn't a public field anymore, it's just a method, so at baseline would've needed to make a bunch of changes to make this possible again. But also even before this change anyone who tries to use `Self` and then make changes to it needs to call `Clone` on it anyways, so then at that point these extra `Clone`s weren't serving any purpose.

core/container.go
sipsma: Triggered a panic at one point due to a bug I wrote while iterating, I believe in  `Container.terminalLegacy`, added this at that time just to avoid panics. Not strictly necessary anymore, but feels like "correct" behavior that's reasonable to leave in.

core/telemetry.go
sipsma: I'll double check this one. Probably the same story as others where I had some bug at some point where `res` was incorrectly nil and just wanted to avoid a panic, but unlike the other cases where I added checks like this the `res == nil` case doesn't feel valid for the long-term... Should be an error if anything.

I'll rm it and if I see any panics will check out what's going wrong.

core/schema/service.go
sipsma: Oh I didn't read your question closely enough, I was just commenting on the wrapping of it in `Result` 😄

I forget exactly why it was `Nullable[Void]` in the first place. I guess it makes sense to just return a true `null` always rather than "instantiate nothing" by returning a concrete `Void`. Either way, not gonna play with changing that as part of this PR; will just remove the extra `Result` wrapping.

core/schema/coremod.go
sipsma: Yep forgot a few `Instance` names, fixed now!

dagql/types.go
sipsma: Yeah this was a new need. Before adding this and the code in `toSelectable` that uses it, just the `dynamicOtherIfaceByIfaceList` [test case](https://github.com/sipsma/dagger/blob/2ec34a0eef9a7c52265a910e557c0af7c6b568f6/core/integration/testdata/modules/go/ifaces/main_test.go#L245-L245) was failing. 

That test makes a tangled web of calls between modules where interfaces from one implements interfaces from another. Before the changes here we were always ending up with a `Instance[*InterfaceAnnotatedValue]` which fast-path'd `toSelectable`. But now that there's a difference between `Result` and `ObjectResult` we only fast-path the case where it's already an `ObjectResult` and we sometimes end up with just a `Result[*InterfaceAnnotatedValue]` that needs to be "upgraded" to `ObjectResult`. It doesn't quite work though because only the underlying object type may be known to the dagql server this is ending up on.
* 😵‍💫

Either way, yeah augmenting dagql with this dash of interface knowledge was the best fix I could think of for now. Was also hoping this could be a starting point for moving more of the interface logic over to dagql, but obviously not gonna go more down that path than needed in this particular PR in the interest of reining in scope.

core/telemetry.go
sipsma: Rm'd it and everything is happy, so will leave it out

jedevc (approved):

core/mcp.go
sipsma: Yep good catch, fixed this and the similar one in `core/env.go`

PR #10632, opened 2025-06-24
feat: add enum member value TypeScript support

TomChv:
Update TS SDK to support enum/value feature introduced by #9518.
Handle serialization through metadata with an external function generated in client.
Handle deserialization inside client gen with an function generated in client.
Add tests for enum generation

Small side changes:
- Update some tests modules version & configuration using the bundle SDK
- Update TS codebase to use new enum names (`TypeDefKind.StringKind` -> `TypeDefKind.String`)
- Update linter to ignore duplication on enum value/unused method in generated client (required tu support new feature)
- Update tests to match new spec

Signed-off-by: Tom Chauveau <tom@dagger.io>

### DIFF
--- a/sdk/python/codegen/src/codegen/generator.py
+++ b/sdk/python/codegen/src/codegen/generator.py
@@ -5,6 +5,7 @@ import logging
 import re
 import textwrap
 from abc import ABC, abstractmethod
+from collections import OrderedDict
 from collections.abc import Callable, Container, Iterable, Iterator
 from dataclasses import dataclass, field
 from datetime import date, datetime, time
@@ -735,25 +736,33 @@ class Enum(Handler[GraphQLEnumType]):
         if body := super().render_body(t):
             yield body
 
+        by_value = OrderedDict()
         for name, value in sorted(t.values.items()):
+            val = self._get_value(value)
+            if val not in by_value:
+                by_value[val] = []
+            by_value[val].append(name)
+
+        for val, names in by_value.items():
             yield ""
 
-            val = None
-            if value.ast_node and (
-                directive := self.ctx.schema.get_directive("enumValue")
-            ):
-                args = graphql.get_directive_values(directive, value.ast_node)
-                if args:
-                    val = args["value"]
-            if not val:
-                val = value.value
-
             # repr uses single quotes for strings, contrary to black
-            val = repr(val).replace("'", '"')
-            yield f"{name} = {val}"
+            valrepr = repr(val).replace("'", '"')
+            yield f"{names[0]} = {valrepr}"
+            if desc := t.values[names[0]].description:
+                yield doc(desc)
 
-            if value.description:
-                yield doc(value.description)
+            for i, name in enumerate(names[1:]):
+                yield f"{name} = {names[0]}"
+                if desc := t.values[names[1 + i]].description:
+                    yield doc(desc)
+
+    def _get_value(self, value) -> str:
+        if value.ast_node and (directive := self.ctx.schema.get_directive("enumValue")):
+            args = graphql.get_directive_values(directive, value.ast_node)
+            if args:
+                return args["value"]
+        return value.value
 
 
 class Field(Protocol):
--- a/sdk/python/src/dagger/client/gen.py
+++ b/sdk/python/src/dagger/client/gen.py
@@ -268,48 +268,39 @@ class ImageLayerCompression(Enum):
     """Compression algorithm to use for image layers."""
 
     ESTARGZ = "EStarGZ"
-
-    EStarGZ = "EStarGZ"
+    EStarGZ = ESTARGZ
 
     GZIP = "Gzip"
-
-    Gzip = "Gzip"
+    Gzip = GZIP
 
     UNCOMPRESSED = "Uncompressed"
-
-    Uncompressed = "Uncompressed"
+    Uncompressed = UNCOMPRESSED
 
     ZSTD = "Zstd"
-
-    Zstd = "Zstd"
+    Zstd = ZSTD
 
 
 class ImageMediaTypes(Enum):
     """Mediatypes to use in published or exported image metadata."""
 
     DOCKER = "DockerMediaTypes"
-
-    DockerMediaTypes = "DockerMediaTypes"
+    DockerMediaTypes = DOCKER
 
     OCI = "OCIMediaTypes"
-
-    OCIMediaTypes = "OCIMediaTypes"
+    OCIMediaTypes = OCI
 
 
 class ModuleSourceKind(Enum):
     """The kind of module source."""
 
     DIR = "DIR_SOURCE"
-
-    DIR_SOURCE = "DIR_SOURCE"
+    DIR_SOURCE = DIR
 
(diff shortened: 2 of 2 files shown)


cmd/codegen/generator/typescript/templates/functions.go
```diff
@@ -301,14 +312,51 @@ func (funcs typescriptTemplateFuncs) sortInputFields(s []introspection.InputValu
...
+		return cmp.Compare(strcase.ToCamel(x.Name), strcase.ToCamel(y.Name))
 	})
 	s = slices.CompactFunc(s, func(x, y introspection.EnumValue) bool {
-		return funcs.formatEnum(x.Name) == funcs.formatEnum(y.Name)
+		return strcase.ToCamel(x.Name) == strcase.ToCamel(y.Name)
 	})
+
+	// Remove enums without values that may exist
+	s = slices.DeleteFunc(s, func(x introspection.EnumValue) bool {
+		return x.Name == "" && x.Directives.EnumValue() == ""
```
jedevc: :thinking: how does this happen?

This feels like a bug, I have seen this happen in some edge cases, but some of them should be fixed by https://github.com/dagger/dagger/pull/10647.

The engine should never send values that look like this.

cmd/codegen/generator/typescript/templates/functions.go
```diff
@@ -301,14 +312,51 @@ func (funcs typescriptTemplateFuncs) sortInputFields(s []introspection.InputValu
 
 func (funcs typescriptTemplateFuncs) sortEnumFields(s []introspection.EnumValue) []introspection.EnumValue {
 	slices.SortStableFunc(s, func(x, y introspection.EnumValue) int {
-		return cmp.Compare(funcs.formatEnum(x.Name), funcs.formatEnum(y.Name))
+		return cmp.Compare(strcase.ToCamel(x.Name), strcase.ToCamel(y.Name))
```
jedevc, line 317: Important! Make sure to call `slices.Clone` before doing any of this.

I had to add this for go, should have added it here too - but without this, we're actually mutating the original (which led to some surprising results when I did this)

cmd/codegen/generator/typescript/templates/src/testdata/type_test_enum_with_directive_want.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+/**
+ * Utility function to convert a ImageLayerCompression name to its value so
+ * it can be properly used inside the module runtime.
+ */
+function ImageLayerCompressionNameToValue(name: string): ImageLayerCompression {
+  switch (name) {
+    case "EStarGZ":
```
jedevc, line 38: Are we missing `Estargz` as a case both here and both? Seems like it *should* be here - we should also have a test that makes sure that *both* values work in queries.

Might be worth extending https://github.com/dagger/dagger/blob/94e41c7388b953df3cea17b231e4ea45dc544b7d/core/integration/module_type_test.go#L1023-L1051 ?

cmd/codegen/generator/typescript/templates/src/testdata/type_test_enum_with_directive_want.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Compression algorithm to use for image layers.
+ */
+export enum ImageLayerCompression {
+  EstarGz = "EStarGZ",
+  Estargz = "EStarGZ",
+  Gzip = "Gzip",
+  Uncompressed = "Uncompressed",
+  Zstd = "Zstd",
+}
```
jedevc: I think this is supported in typescript?

```suggestion
export enum ImageLayerCompression {
  EstarGz = "EStarGZ",
  Estargz = EstarGz,
  Gzip = "Gzip",
  Uncompressed = "Uncompressed",
  Zstd = "Zstd",
}
```

That way, `ImageLayerCompression.EstarGz == ImageLayerCompression.Estargz` which is a pretty handy property (we do this in go).

sdk/typescript/src/module/introspector/typescript_module/typedef_utils.ts
```diff
@@ -3,17 +3,17 @@ import { IntrospectionError } from "../../../common/errors/index.js"
 import { TypeDef } from "../typedef.js"
 
 export function isTypeDefResolved(typeDef: TypeDef<TypeDefKind>): boolean {
-  if (typeDef.kind !== TypeDefKind.ListKind) {
+  if (typeDef.kind !== TypeDefKind.List) {
```
jedevc: Confused why this doesn't break :thinking: 

The `_KIND` values shouldn't be present before v0.18.11 - so how does this work for version compatibility on older versions? Why is this accessible?

TomChv: Ohhh it's because I updated some modules and regenerated clients! I'll revert that back to kind! Good catch

jedevc: weiiird that it wouldn't have caught by the legacy tests though

cmd/codegen/generator/typescript/templates/src/testdata/type_test_enum_with_directive_want.ts
TomChv: Since EstarGZ == estarGZ, if you sent the value: `ImageLayerCompression.EstarGz`, it will match the 1 case:

```
    case ImageLayerCompression.EstarGz: // == ImageLayerCompression.Estargz
      return "EStarGZ"
```

on the otherside if we retrieve the name, that means it come from an external client so it will be `EstarGZ`.

I'll see if I can stil add that extra case but I'm not sure it's an issue

cmd/codegen/generator/typescript/templates/functions.go
TomChv: It actually fixed the empty value in array but using slices.Clone, very good catch 🚀

cmd/codegen/generator/typescript/templates/functions.go
TomChv: Fixed by using `slices.Clone` during array manipulation ;)

cmd/codegen/generator/typescript/templates/src/testdata/type_test_enum_with_directive_want.ts
jedevc: Beauuuuutiful :tada:

cmd/codegen/generator/typescript/templates/src/testdata/type_test_enum_with_directive_want.ts
jedevc: Can you add this as a test as I mentioned above? If we can do `dagger query` with both estargz values, then, SGTM as is, with no changes.

TomChv: Tests added, it works fine on Go and Typescript but fails on Python, I added a condition so @helderco can fix it whenever he has time :)

jedevc (approved):

PR #10636, opened 2025-06-24
fix plain output being way too verbose

vito:
[Comparison](https://gist.github.com/vito/00947008407bebd6c61c8d23eedb77f3) (213 => 53 lines)

This change simply brings the code in line with the pretty frontend. Not sure if this regressed or was always out of sync. :thinking:


gerhard (approved):
👍

wingyplus:
Nice! Is the log being verbose based on the verbose flag?

jedevc (approved):

vito:
@wingyplus afaik it already respects the verbosity flag, this fix was for an orthogonal issue that made it too noisy

PR #10643, opened 2025-06-25
dagql: fix Server.Schema <-> Class.Extend deadlock

vito:
The fact that these are happening concurrently is a little smelly, but we should at least avoid the deadlock.


vito:
Ran `dagger-dev` 10 times from the repo root and none of them deadlocked, vs. the second one deadlocking with v0.18.11, so it seems like this fixes it. Merging!

PR #10662, opened 2025-06-30
Allow loading `Container`s to host

jedevc:
Fixes https://github.com/dagger/dagger/issues/8025.

This adds a new `Container.load` API, which accepts a `name` and exports the image to the host's container store. It's similar in style to how `Export` works, which exports a tarball to the host. Bikeshedding welcome!

:warning: Depends on https://github.com/dagger/dagger/pull/10671.

TODO:
- [x] `Container.load` API
- [x] Manual store override with `_EXPERIMENTAL_DAGGER_RUNNER_IMAGESTORE`
- [x] Containerd store loading
  - [x] Containerd tests
- [x] Cleanup


go.mod
```diff
@@ -11,7 +11,9 @@ require (
 replace (
 	dagger.io/dagger => ./sdk/go
 	github.com/dagger/dagger/engine/distconsts => ./engine/distconsts
-	github.com/moby/buildkit => github.com/dagger/buildkit v0.0.0-20250519134855-c1734e1f6155
+	// XXX: replace before merge
+	// github.com/moby/buildkit => github.com/dagger/buildkit v0.0.0-20250519134855-c1734e1f6155
+	github.com/moby/buildkit => github.com/jedevc/buildkit v0.8.2-0.20250702153345-4eaf5b2e5121
```
jedevc: Depends on the following change in dagger/buildkit: https://github.com/dagger/buildkit/compare/master...jedevc:buildkit:oci-custom-store

core/schema/container.go
```diff
@@ -569,6 +574,28 @@ func (s *containerSchema) Install() {
 			View(BeforeVersion("v0.12.0")).
 			Extend(),
 
+		dagql.NodeFunc("load", s.load).
```
jedevc: @shykes suggests `exportImage`, which, agreed does make it clearer.

core/container.go
```diff
@@ -1489,9 +1492,14 @@ func (container *Container) Export(
 	}
 
 	opts := map[string]string{
-		"tar":                           strconv.FormatBool(true),
+		"tar":                           strconv.FormatBool(tar),
 		string(exptypes.OptKeyOCITypes): strconv.FormatBool(mediaTypes == OCIMediaTypes),
 	}
+
+	// HACK: these properties are implemented in our buildkit fork
```
sipsma: nit: if it's just a matter of `cp -r` and updating a few import statements, I'd almost rather do that at this point rather than maintain the code across two repos. Not a big deal at all, just a thought

core/schema/container.go
```diff
@@ -2034,6 +2063,141 @@ func (s *containerSchema) asTarball(
...
+	bk, err := query.Buildkit(ctx)
+	if err != nil {
+		return core.Void{}, fmt.Errorf("failed to get buildkit client: %w", err)
+	}
+
+	variants, err := dagql.LoadIDs(ctx, s.srv, args.PlatformVariants)
+	if err != nil {
+		return core.Void{}, err
+	}
+
```
sipsma: nit: `variants` looks like it's only used in the `if loader.ContentStore != nil {`, not the tarball codepath, so may as well only call it in the branch that needs it

core/schema/container.go
```diff
@@ -2034,6 +2063,141 @@ func (s *containerSchema) asTarball(
...
+			if err != nil {
+				return err
+			}
+			defer f.Close()
+
+			// stream in chunks *definitely* smaller than the max gRPC message size
+			buf := make([]byte, 3*1024*1024)
+			_, err = io.CopyBuffer(loader.Tarball, f, buf)
+			if err != nil {
+				loader.Tarball.Close()
```
sipsma: would we be potentially leaking something if we don't `.Close()` this in all earlier codepaths that return early? maybe this should just be done in a defer?

core/container.go
```diff
@@ -1466,18 +1467,20 @@ func (container *Container) Export(
 	platformVariants []*Container,
 	forcedCompression ImageLayerCompression,
 	mediaTypes ImageMediaTypes,
-) error {
+	tar bool,
+	leaseID string,
```
sipsma: nit: at this point seems worth making an args struct given the number of params

core/container.go
jedevc: Yeah, good plan.

Issue #10546, opened 2025-06-07
🐞 UnknownDaggerError using Dagger with Deno


michaelmass:
### What is the issue?

There a strange interaction when adding to many instructions on the same container when I am using `Dagger` with `Deno`. If I call `.withEnvVariable` more than 20 times the programmable workflow throws an `UnknownDaggerError`. I've made a simple script below to replicate this error.

Thanks for supporting `Deno`. This is the feature I was expecting for a long time. Keep up the good work! 🦾

### Dagger version

dagger v0.18.9 (docker-image://registry.dagger.io/engine:v0.18.9) darwin/arm64

### Steps to reproduce

You only have to call `dagger run -i deno run  --no-lock --node-modules-dir=false -A test.ts` on the following code. 

```ts
import {connect} from "npm:@dagger.io/dagger@0.18.9";

await connect(async client => {
  await client.container()
    .from("node:22-slim")
    .withEnvVariable('1', 'test')
    .withEnvVariable('2', 'test')
    .withEnvVariable('3', 'test')
    .withEnvVariable('4', 'test')
    .withEnvVariable('5', 'test')
    .withEnvVariable('6', 'test')
    .withEnvVariable('7', 'test')
    .withEnvVariable('8', 'test')
    .withEnvVariable('9', 'test')
    .withEnvVariable('10', 'test')
    .withEnvVariable('11', 'test')
    .withEnvVariable('12', 'test')
    .withEnvVariable('13', 'test')
    .withEnvVariable('14', 'test')
    .withEnvVariable('15', 'test')
    .withEnvVariable('16', 'test')
    .withEnvVariable('17', 'test')
    .withEnvVariable('18', 'test')
    .withEnvVariable('19', 'test')
    .sync();
});
```

Here's my deno version
```
deno 2.3.3 (stable, release, aarch64-apple-darwin)
v8 13.7.152.6-rusty
typescript 5.8.3
```

### Log output

```
dagger run -i deno run  --no-lock --node-modules-dir=false -A ci/test.ts
✔ connect 0.2s

✔ version: String! 0.0s

✔ container: Container! 0.0s
$ .from(address: "node:22-slim"): Container! 0.3s CACHED
✔ .withEnvVariable(name: "1", value: "test"): Container! 0.0s
✔ .withEnvVariable(name: "2", value: "test"): Container! 0.0s
✔ .withEnvVariable(name: "3", value: "test"): Container! 0.0s
[...]

PR #10673, opened 2025-07-02
fix: deno fetch failure 

TomChv:
Fixes: https://github.com/dagger/dagger/issues/10546

- Use native Deno fetch instead of NodeFetch when creating a GraphQL client
- Upgrade Deno runtime to latest 2.4.0

See https://github.com/dagger/dagger/issues/10546#issuecomment-3027288130 for explanations of that changes

### DIFF
--- a/sdk/typescript/src/common/graphql/client.ts
+++ b/sdk/typescript/src/common/graphql/client.ts
@@ -1,6 +1,6 @@
 import * as opentelemetry from "@opentelemetry/api"
 import { GraphQLClient } from "graphql-request"
-import fetch from "node-fetch"
+import nodeFetch from "node-fetch"
 import {
   RequestInfo as NodeFetchRequestInfo,
   RequestInit as NodeFetchRequestInit,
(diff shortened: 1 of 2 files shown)


sdk/typescript/src/common/graphql/client.ts
```diff
@@ -22,7 +22,18 @@ const createFetchWithTimeout =
     }, timeout)
 
     try {
-      return (await fetch(input as NodeFetchRequestInfo, {
+      // Edge case for Deno that doesn't work as expected with node-fetch and would
+      // rather rely on its native fetch implementation
+      // See: https://github.com/dagger/dagger/issues/10546
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      if (typeof (globalThis as any).Deno !== "undefined") {
```
jedevc: Maybe split this condition out to something like `isDeno` or something? Just for readability?

jedevc (approved):

sdk/typescript/src/common/graphql/client.ts
TomChv: Oh yeah good idea!

PR #10679, opened 2025-07-03
feat: add field to get latest git ref tag

jedevc:


core/schema/git.go
```diff
@@ -71,6 +73,8 @@ func (s *gitSchema) Install() {
 	dagql.Fields[*core.GitRepository]{
 		dagql.NodeFuncWithCacheKey("head", s.head, dagql.CachePerSession).
 			Doc(`Returns details for HEAD.`),
+		dagql.NodeFuncWithCacheKey("latest", s.latest, dagql.CachePerSession).
```
vito: nit: `latest` doesn't convey semver to me (could also be latest commit). wdyt of `latestVersion`? (Or similar)

jedevc: SGTM :tada:

vito (approved):
noice

PR #10680, opened 2025-07-03
Jsonvalue

shykes:
An API for manipulating JSON values as Dagger objects.

## Examples

```console
⧖ json | with-contents '{"hello": "world"}' | field hello | as-string
world
```

```console
⧖ json | with-field hello $(json | new-string world) | contents --pretty
{
 "hello": "world"
}
```

## Float support

Float support is missing, because [it breaks in the Rust SDK](https://github.com/dagger/dagger/pull/10680#issuecomment-3105241956)

## See also

- This is a replacement of *#10496*
- This is a dependency of #10744


shykes:
Cross-posting this comment by @jedevc, from #10496 

> Instead of having dedicated get<Type>/set<Type> methods, what about having just get+set and then as<Type> methods? Then they chain as, .Get("x").AsInt(), which feels very much like interfaces...

> For set, we'd want to have a scalar value that can accept any value (I think this is possible, happy to provide pointers, if this seems like a reasonable idea).

@jedevc could you tell me more about how we might achieve this?

core/integration/jsonvalue_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// Test creating a JSON integer and retrieving its value
+	jsonInt := c.JSON().NewInteger(42)
+
+	// Test AsInteger method
+	value, err := jsonInt.AsInteger(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if value != 42 {
```
alexcb: nit: use `require.Equal(t, 42, value)` instead

core/integration/jsonvalue_test.go
```diff
@@ -0,0 +1,50 @@
...
+		t.Fatalf("expected 42, got %d", value)
+	}
+
+	// Test with negative integer
+	jsonNegInt := c.JSON().NewInteger(-123)
+	negValue, err := jsonNegInt.AsInteger(ctx)
+	if err != nil {
```
alexcb: similar nit: use `require.NoError(t, err)` instead, as it's more concise.

core/schema/jsonvalue.go
```diff
@@ -0,0 +1,294 @@
...
+		),
+		dagql.Func("asString", s.asString).Doc("Decode a string from json"),
+		dagql.Func("newBoolean", s.newBoolean).Doc("Encode a boolean to json").Args(
+			dagql.Arg("value"),
+		),
+		dagql.Func("asBoolean", s.asBoolean).Doc("Decode a boolean from json"),
+
+		dagql.Func("asArray", s.asArray).Doc("Decode an array from json"),
+		dagql.Func("fields", s.fields).Doc("List fields of the encoded object"),
+		dagql.Func("field", s.field).Doc("Lookup the field at the given path, and return its value. The path ").Args(
```
alexcb: there's some extra text at the end of the line here.
```suggestion
		dagql.Func("field", s.field).Doc("Lookup the field at the given path, and return its value.").Args(
```

alexcb (commented):
Looking good!

Should there also be `asFloat` / `newFloat` methods?

core/integration/jsonvalue_test.go
shykes: good catch thanks! Vibe-coded the first pass of tests, then ran into insane codegen issues, then forgot to go back and clean up...

shykes:
> Looking good!
> 
> Should there also be `asFloat` / `newFloat` methods?

Yes good point! Will add

Thanks for the review!

shykes:
@alexcb float added. But now I have a weird error when generating Rust sdk? Possibly related? 🤷 https://dagger.cloud/dagger/traces/b1e0feb783637b6a57f80809ed0dfebf#f8b38419903644f4

EDIT: yup, definitely related... 😨 

```
error[E0412]: cannot find type `float` in this scope
    --> crates/dagger-sdk/src/gen.rs:3183:44
     |
3183 |  pub async fn as_float( &self, ) -> Result<float, DaggerError> { let mut query = self.selection.select("asFloat"); query.execute(self.gra...
     |                                            ^^^^^ not found in this scope
     |
help: perhaps you intended to use this type
     |
3183 |  pub async fn as_float( &self, ) -> Result<f32, DaggerError> { let mut query = self.selection.select("asFloat"); query.execute(self.graphql_client.cl
one()).await }
     |                                            ~~~
help: you might be missing a type parameter
     |
3178 |  #[builder(setter(into, strip_option), default)] pub pretty: Option<bool>, } impl<float> Value { /// Decode an array from json
     |                                                                                  +++++++

error[E0412]: cannot find type `float` in this scope
    --> crates/dagger-sdk/src/gen.rs:3209:34
```

shykes:
I'm keeping the code for `asFloat` and `newFloat`, but disabling it for now until the issue is resolved with the Rust SDK.

alexcb:
> I'm keeping the code for `asFloat` and `newFloat`, but disabling it for now until the issue is resolved with the Rust SDK.

Bummer. Given it's non-trivial, I guess we'll have to wait for real users to request this.

shykes:
The last remaining test failures are flakes...

core/schema/jsonvalue.go
```diff
@@ -0,0 +1,294 @@
...
+	var v any
+	if err := json.Unmarshal(obj.Data, &v); err != nil {
+		return nil, err
+	}
+
+	current := v
+	for _, pathSegment := range args.Path {
+		m, ok := current.(map[string]any)
+		if !ok {
+			return nil, fmt.Errorf("path not found: not an object")
```
helderco: Maybe include the path so far in the error?

core/schema/jsonvalue.go
```diff
@@ -0,0 +1,294 @@
...
+
+	// JSONValue methods
+	dagql.Fields[*core.JSONValue]{
+		dagql.Func("bytes", s.bytes).Doc("Return the value encoded as json").Args(
+			dagql.Arg("pretty").Doc("Pretty-print"),
+			dagql.Arg("indent").Doc("Indent each line by the given number of whitespaces"),
+		),
+		dagql.Func("withBytes", s.withBytes).Doc("Return a new json value, decoded from the given bytes").Args(
+			dagql.Arg("bytes"),
+		),
```
helderco: I think the term "bytes" can be a bit confusing because it has a specific meaning in a lot of languages and this actually represents a `dagger.JSON` which is a scalar value, i.e., a string.

The [GJSON library](https://github.com/tidwall/gjson?tab=readme-ov-file#result-type) uses the field name "Raw" for this. If `withRaw` looks weird, maybe use `doc` aka "document"? Or even just `JSON` and `withJSON`.

shykes: I agree `bytes` is not ideal, but IMO `raw, `doc` or `json` all have issues that make it worse.

What about `contents`?

shykes:
> Neat!

Haha thanks ❤️

It's a building block for #10744 . It will allow removing `withStringBinding`, `withIntegerBinding`, `withBooleanBinding`, and any other special case for binding non-objects to an env. Instead, you can just build a `JSONValue` of your choice, and pass *that* as an object binding.

core/schema/jsonvalue.go
helderco: > What about `contents`?

Can't think of anything better but it's better than "bytes".

core/integration/jsonvalue_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type JSONSuite struct{}
+
+func TestJSON(t *testing.T) {
+	testctx.New(t, Middleware()...).RunTests(JSONSuite{})
+}
+
+func (JSONSuite) TestInteger(ctx context.Context, t *testctx.T) {
```
jedevc, line 18: suggestion: a couple more tests? We don't need *loads*, just checking that:
- all the `As` methods work
- we can access nested paths
- trying to lookup an empty path gives an error

core/schema/jsonvalue.go
```diff
@@ -0,0 +1,294 @@
...
+
+type jsonvalueSchema struct{}
+
+var _ SchemaResolvers = &jsonvalueSchema{}
+
+func (s jsonvalueSchema) Install(srv *dagql.Server) {
+	// Top-level constructor: json (creates new empty JSONValue)
+	dagql.Fields[*core.Query]{
+		dagql.Func("json", s.newJSONValue).
+			Doc("Initialize a mutable JSON value"),
```
jedevc: nit: it's not mutable, like all dagger types, it's immutable.

core/schema/jsonvalue.go
vito: Maybe `encode` and `decode`?

shykes: I changed it to `contents` and `withContents`. It looks good I think.

docs/docs-graphql/schema.graphqls
```diff
@@ -2830,6 +2854,58 @@ scalar InterfaceTypeDefID
...
+  field(path: [String!]!): JSONValue!
+
+  """List fields of the encoded object"""
+  fields: [String!]!
+
+  """A unique identifier for this JSONValue."""
+  id: JSONValueID!
+
+  """Encode a boolean to json"""
+  newBoolean(value: Boolean!): JSONValue!
```
jedevc: Hehe, I get to cosplay as @helderco here 😁 (who continually reminds me ❤️)

All arguments should have doc-strings :tada:

core/schema/jsonvalue.go
```diff
@@ -0,0 +1,294 @@
...
+	data, err := json.Marshal(nil)
+	if err != nil {
+		return nil, err
+	}
+	return &core.JSONValue{Data: data}, nil
+}
+
+func (s jsonvalueSchema) contents(ctx context.Context, obj *core.JSONValue, args struct {
+	Pretty dagql.Optional[dagql.Boolean]
+	Indent dagql.Optional[dagql.Int] `default:"1"`
```
jedevc: Can we make this a string arg? Having tab-indented as an option would be nice.

(also, 1 as a default is *strange*, 1-space indented JSON is... unique)

core/schema/jsonvalue.go
```diff
@@ -0,0 +1,294 @@
...
+	return &core.JSONValue{Data: data}, nil
+}
+
+func (s jsonvalueSchema) asInteger(ctx context.Context, obj *core.JSONValue, args struct{}) (dagql.Int, error) {
+	var v any
+	if err := json.Unmarshal(obj.Data, &v); err != nil {
+		return 0, err
+	}
+	switch val := v.(type) {
+	case float64:
```
jedevc: `var v int` should simplify this, instead of needing the `switch` case.

core/schema/jsonvalue.go
```diff
@@ -0,0 +1,294 @@
...
+func (s jsonvalueSchema) asString(ctx context.Context, obj *core.JSONValue, args struct{}) (dagql.String, error) {
+	var v any
+	if err := json.Unmarshal(obj.Data, &v); err != nil {
+		return "", err
+	}
+	str, ok := v.(string)
+	if !ok {
+		return "", fmt.Errorf("value is not a string")
+	}
+	return dagql.String(str), nil
```
jedevc: As above, we can simplify a lot of these by directly unmarshalling to the desired `string` type.

core/schema/jsonvalue.go
jedevc: Agreed on this, otherwise the error is gonna be very hard to read.

shykes:
@jedevc tests finally pass 🙏

docs/docs-graphql/schema.graphqls
```diff
@@ -2830,6 +2854,58 @@ scalar InterfaceTypeDefID
...
+  """Decode an integer from json"""
+  asInteger: Int!
+
+  """Decode a string from json"""
+  asString: String!
+
+  """Return the value encoded as json"""
+  contents(
+    """Pretty-print"""
+    pretty: Boolean
```
jedevc: Should be:

```suggestion
    pretty: Boolean = false
```

To be consistent with the rest of the schema.

docs/docs-graphql/schema.graphqls
```diff
@@ -2830,6 +2854,58 @@ scalar InterfaceTypeDefID
...
+  id: JSONValueID!
+
+  """Encode a boolean to json"""
+  newBoolean(value: Boolean!): JSONValue!
+
+  """Encode an integer to json"""
+  newInteger(value: Int!): JSONValue!
+
+  """Encode a string to json"""
+  newString(value: String!): JSONValue!
```
jedevc: 🤔 No `newFloat`?

As a hater of floats, I think this is reasonable, but users may ask for it.

shykes: See https://github.com/dagger/dagger/pull/10680#issuecomment-3105241956 ... I had to remove it 😭

docs/docs-graphql/schema.graphqls
shykes: Isn't this a generated file?

jedevc: Yeah, something in the original code is wrong, there will be a missing `default:` field or something.

docs/docs-graphql/schema.graphqls
jedevc: Uh, do you have the code around still? I can take a look, I'm familiar with rust.

core/schema/jsonvalue.go
shykes: 👍✅

core/schema/jsonvalue.go
shykes: I believe this is necessary because `json.Unmarshal` unmarshals json numbers into go float64 values?

shykes: Edit: I just tried it, and it appears I was wrong (and so were both chatgpt and claude??)

docs/docs-graphql/schema.graphqls
shykes: OK, as a general note it's part of the spec that `bool` has an implicit default of `false`, but I don't mind adding it if it's consistent with the rest of the schema.

shykes:
@jedevc assuming checks pass, would love a LGTM 🙏

jedevc:
LGTM :tada:

(if you could hold merging until the release has gone out, that would be wonderful)

PR #10683, opened 2025-07-07
refactor: split codegen CLI

TomChv:
Various work around the codegen binary with as main purpose to split codegen CLI into 2 subcommand: `generate-module` & `generate-client`

This PR also include the following changes:
- Clean up overall codegen binary architecture (will improve it more with follow ups PRs)
- Split up arguments to match the commands
- Move some functions to different location to improve maintainability


sdk/go/generate.go
```diff
@@ -1,3 +1,3 @@
 package dagger
 
-//go:generate go -C ../../ run ./cmd/codegen --output ./sdk/go --merge
```
jedevc, line 3: IMO, I think this justifies another subcommand:
- `generate-library` generates this?
- `generate-module` generates *just* modules
- `generate-client` generates the standalone generated client

Not sure on naming, but we shouldn't have this be `generate-module`, it's not a module.

cmd/codegen/codegen.go
```diff
@@ -97,7 +97,7 @@ func generate(ctx context.Context, introspectionSchema *introspection.Schema, in
 		return nil, fmt.Errorf("use target SDK language: %s: %w", sdks, generator.ErrUnknownSDKLang)
 	}
 
-	if cfg.ClientOnly {
+	if cfg.ClientConfig != nil {
```
jedevc: So, I'd be curious if we could push the refactor a bit further here: atm, we have `GenerateModule` -> `Generate` -> `gen.GenerateModule` - it goes "specific" -> "generic" -> "specific" again.

Is there a way that we can have `GenerateModule` call `gen.GenerateModule` itself, extracting all of the setup somehow? Not quite sure which direction to take that, but it feels like that would be the cleaner code structure.

cmd/codegen/generator/config.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Close existing dagger client if it exists.
+// This is a convenience method to be used in the main codegen command using a defer.
+func (c *Config) Close() error {
+	if c.Dag != nil {
+		return c.Dag.Close()
+	}
+
+	return nil
+}
```
jedevc, line 41: nit: this is a bit weird.

Actually, is there a reason we can't use `dagger.io/dagger/dag` for the default global client? Then we can just call `defer dag.Close()` in each function, which will close it if it was setup. https://github.com/dagger/dagger/blob/80660ea3951e2a9a39e2697641d242afba7341dd/sdk/go/dag/dag.gen.go#L36-L47

cmd/codegen/generator/config.go
```diff
@@ -0,0 +1,77 @@
...
+	// Whether we are initializing a new module.
+	// Currently, this is only used in go codegen to enforce backwards-compatible behavior
+	// where a pre-existing go.mod file is checked during dagger init for whether its module
+	// name is the expected value.
+	IsInit bool
+}
+
+type ModuleSourceDependencies struct {
+	Kind   string
+	Name   string `json:"moduleOriginalName"`
```
jedevc, line 63: :thinking: Remind me, why is this not `moduleName`?

cmd/codegen/generator/config.go
```diff
@@ -0,0 +1,77 @@
...
+	Merge bool
+
+	// Whether we are initializing a new module.
+	// Currently, this is only used in go codegen to enforce backwards-compatible behavior
+	// where a pre-existing go.mod file is checked during dagger init for whether its module
+	// name is the expected value.
+	IsInit bool
+}
+
+type ModuleSourceDependencies struct {
```
jedevc:
```suggestion
type ModuleSourceDependency struct {
```

cmd/codegen/introspection/introspect.go
jedevc, line 1: Perhaps a fun follow-up - we have several `introspection` packages:
- `dagql/introspection`
- `cmd/codegen/introspection`
- And *many* copies of `introspection.graphql` around.

Might be good to *unify* all these together.

cmd/codegen/introspect.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/dagger/dagger/cmd/codegen/introspection"
+	"github.com/spf13/cobra"
+)
+
+var outputSchema string
+
+var introspectCmd = &cobra.Command{
+	Use:  "introspect",
+	RunE: Introspect,
+}
```
jedevc, line 18: Is the only place we use this here? https://github.com/dagger/dagger/blob/80788af41531cbde52d122927f2fcefc256a1249/.dagger/sdk.go#L95-L96

I think we should switch everywhere that's getting introspection like this to use `./cmd/introspect` instead, and just remove this subcommand.

cmd/codegen/generator/overlay.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+)
+
+func Overlay(ctx context.Context, logsW io.Writer, overlay fs.FS, outputDir string) (rerr error) {
```
jedevc: nit: I think we can use https://pkg.go.dev/log/slog for all of our logging. No neat to pass around the writer `os.Stdout` everywhere.

cmd/codegen/generator/client.go
```diff
@@ -0,0 +1 @@
+package generator
```
jedevc: :eyes: is this file needed?

sdk/go/generate.go
```diff
@@ -1,3 +1,3 @@
...
+//go:generate go -C ../../ run ./cmd/codegen generate-module --output ./sdk/go --merge
```
jedevc: I don't think this needs `--merge` actually :thinking: Right? There's a `go.mod` *in* `sdk/go`.

If that's the case, there's no more uses of `--merge`, so we should remove the behavior.

cmd/codegen/generator/go/load_package.go
jedevc, line 1: `loader.go` is fine here. We should avoid naming files after specific functions they contain, just make it about the "general" behavior.

cmd/codegen/generator/module.go
jedevc, line 1: nit: another file we don't need.

cmd/codegen/generator/config.go
TomChv: Yeah, sadly `dagger.io/dagger/dag` doesn't expose `Do`, and not `GraphQLClient` either, I could change that if it make sense though

cmd/codegen/generator/config.go
TomChv: Because it may be different from the true module name that we want to load, I do not remember the exact reason but it wasn't working.
We're using the originalName when we're loading module so why not there?

sdk/go/generate.go
TomChv: Are you sure? I can see it on main: https://github.com/dagger/dagger/blob/main/sdk/go/go.mod

jedevc: Isn't the behavior of `--merge` to use the *same* as the top-level `go.mod` though? See https://github.com/dagger/dagger/pull/8095.

We always want to use `sdk/go/go.mod`, not `go.mod`.

TomChv: Ohhh so yeah I think we can remove that flag, will do some update

cmd/codegen/introspect.go
TomChv: That requires a little bit of update of that introspect command to save the result in a file, should be very easy but I would rather do it in a follow up since it touch the actual dagger CLI

cmd/codegen/codegen.go
TomChv: I did some changes to remove as much condition as possible, so now we have `Generate(Library|Module|Client)` in the interface.
`Generate` still exist but only do the generic logic to fetch the schema and then call the given callback that generate the function. 
We could remove it too and do that logic in the CLI func, this would lead to a little bit of duplication but would be fine.

sdk/go/generate.go
TomChv: done :)

cmd/codegen/generator/config.go
TomChv: This would require some changes in the generated binding for the global client, I think it should be done in a follow up same as the introspect comments https://github.com/dagger/dagger/pull/10683#discussion_r2209972117 and https://github.com/dagger/dagger/pull/10683#discussion_r2209976363

cmd/codegen/common.go
```diff
@@ -0,0 +1,59 @@
...
+	if introspectionJSONPath != "" {
+		introspectionJSON, err := os.ReadFile(introspectionJSONPath)
+		if err != nil {
+			return generator.Config{}, fmt.Errorf("read introspection json: %w", err)
+		}
+		cfg.IntrospectionJSON = string(introspectionJSON)
+	}
+
+	return cfg, nil
+}
```
TomChv, line 59: I think I can refactor that part to be command specific, correct me if I'm wrong there:
- generate-module always have a introspectionJSONPath set
- generate-client always have a introspectionJSONPath set and moduleID
- generate-library never has introspectionJSONPath

So technically, I could do that work per command and remove that `cfg.Close` function since I could close the client earlier (even prior to call `Generate`)

cmd/codegen/codegen.go
jedevc: Yeah! Nice nice nice! :tada:

Like it a lot, it's much neater now :yellow_heart:

cmd/codegen/generator/config.go
jedevc: Hm, yeah, that's an oversight that it doesn't. `dag` from `dagger.io/dagger/dag` should ideally support all the same methods as available on `dag` available in modules.

So yeah, would be good to have there. I don't think the changes should be too hard, happy to do them here, or a follow-up.

cmd/codegen/introspect.go
jedevc: :thinking: why does it touch the actual dagger CLI?

We don't actually need to update the introspection command, we could just use `RedirectStdout`.

cmd/codegen/common.go
jedevc: Hm, I like that currently `generate-*` doesn't *assume* which one is set. I don't think we need it to be command specific (it's actually nice it's generic).

In this case, just because those are always true, we might change those in the future.

cmd/codegen/generate_module.go
```diff
@@ -0,0 +1,83 @@
...
+	moduleConfig.ModuleParentPath = moduleParentPath
+
+	cfg.ModuleConfig = moduleConfig
+
+	generator, err := getGenerator(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to get generator: %w", err)
+	}
+
+	slog.Info("generating %s module: %s\n", "language", cfg.Lang, "module-name", cfg.ModuleConfig.ModuleName)
```
jedevc: nit: this string shouldn't have `%s` in it, the formatted fields should be good.

Worth fixing this, and just check the other instances to make sure they look right too.

cmd/codegen/generate_module.go
```diff
@@ -0,0 +1,83 @@
...
+
+	return Generate(ctx, cfg, generator.GenerateModule)
+}
+
+func init() {
+	// Generation flags
+	generateModuleCmd.Flags().StringVar(&lang, "lang", "go", "language to generate")
+	generateModuleCmd.Flags().StringVarP(&outputDir, "output", "o", ".", "output directory")
+	generateModuleCmd.Flags().StringVar(&introspectionJSONPath, "introspection-json-path", "", "optional path to file containing pre-computed graphql introspection JSON")
+	generateModuleCmd.Flags().BoolVar(&bundle, "bundle", false, "generate the client in bundle mode")
```
jedevc: nit: these can be added to `rootCmd.PersistentFlags`.

cmd/codegen/introspect.go
TomChv: Oh yeah that's right!

cmd/codegen/generator/config.go
TomChv: I'm wondering if that's something we need for other SDKs too or we can keep it a Go specific feature?

jedevc: Yeah, they should - I think it's just that I did a bad job on the standalone go package originally, and missed this out.

I think the others are already structured in a way where they don't have this problem?

cmd/codegen/generate_module.go
TomChv: Done in [862e5d4](https://github.com/dagger/dagger/pull/10683/commits/862e5d4f7d10035bf0678cd6ea8520cc586123c5)

cmd/codegen/generate_module.go
TomChv: Done in [862e5d4](https://github.com/dagger/dagger/pull/10683/commits/862e5d4f7d10035bf0678cd6ea8520cc586123c5)

cmd/codegen/introspect.go
TomChv: Done in [862e5d4](https://github.com/dagger/dagger/pull/10683/commits/862e5d4f7d10035bf0678cd6ea8520cc586123c5)

cmd/codegen/generate_client.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	cfg.ClientConfig = clientConfig
+
+	generator, err := getGenerator(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to get generator: %w", err)
+	}
+
+	slog.Info("generating SDK client\n", "language", cfg.Lang)
```
jedevc: nit: we don't need `\n` at the end, these can just be plain messages.

cmd/codegen/codegen.go
```diff
@@ -33,73 +49,47 @@ func Generate(ctx context.Context, cfg generator.Config) (err error) {
...
+				cmd.Dir = filepath.Join(cfg.OutputDir, cfg.ModuleConfig.ModuleSourcePath)
 			}
 			cmd.Stdout = os.Stdout
 			cmd.Stderr = os.Stderr
-			fmt.Fprintln(logsW, "running post-command:", strings.Join(cmd.Args, " "))
+			slog.Info("running post-command:", "args", strings.Join(cmd.Args, " "))
 			err := cmd.Run()
 			if err != nil {
-				fmt.Fprintln(logsW, "post-command failed:", err)
+				slog.Error("post-command failed:", "error", err)
```
jedevc: same here, no need for `:`

jedevc:
I think failures are because those args need to be declared under `PersistentFlags`, not just `Flags`.

cmd/codegen/codegen.go
TomChv: Done

cmd/codegen/generate_client.go
TomChv: Done

TomChv:
> I think failures are because those args need to be declared under `PersistentFlags`, not just `Flags`.

Fixed :)

jedevc:
Woops, accidentally clicked the wrong button there!

Woooonderful work :tada: Let's get this in! Thanks for spending some time on just cleaning up and making this so much neater :tada:

PR #10697, opened 2025-07-08
POC: inject default secret in current module constructor

shykes:
This is a first POC for designing the much-requested "default secrets" feature, as part of #10370 .

In this POC, a default secret will be injected if the following conditions are true:

- A top-level module is loaded (with `dagger -m` or implicitly from the current directory)
- That module has a constructor
- That constructor has an argument called `token`
- That argument is of type `Secret`

If all the above is true, *then* that argument is modified to have a new default value: the secret at URI `env://TOKEN`

## Example

```console
export TOKEN=<a valid github token>
dagger -m github.com/sagikazarmark/daggerverse/gh@v0.11.0 -c 'clone --repo=github.com/dagger/dagger'
```

This is equivalent to calling `dagger -m github.com/sagikazarmark/daggerverse/gh@v0.11.0 -c '. --token=env://TOKEN | clone --repo=github.com/dagger/dagger'

FYI @jedevc @eunomie @rawkode


rawkode:
So happy to see this moving forward. 

Would it be possible to expand the scope from top level module to any local module within the Git repository?

Meaning the security boundary is our Git repo and any external modules would lose access?

nipuna-perera:
Glad to see this moving also!

One drawback I see with this method is that it's tied to a constructor. We have modules that we chain and sometimes a downstream module needs a secret. For example, say I have this

`Java` -> `Maven` -> `Pipeline`

I have a token `SONAR_TOKEN` that isn't relevant to `Java` or `Maven`. I won't be able to default it with this proposed method. 

Another drawback would be the host ENV var becomes tied to the _name_ of the module param if the consumer is using "inline/platform" type of module. Module may be created and maintained by different teams, so one could have a `PASSWORD` and the other a `PIPELINE_PSW`. The consumer then has to match the env var which has the same value. 

I like a more explicit approach. Secrets are name-spaced to the module and the user can have a `.env` (or some other file) like the following

```
MODULE1_PIPELINE_PSW=env://PIPELINE_PSW
MODULE2_PIPELINE_PSW=env://PIPELINE_PSW
MODULE3_PASSWORD=env://PIPELINE_PSW
```

This supports the sandbox better because user still has to tell dagger to use their env var and it's not pulled in under the hood. The drawback is, if you are using a whole bunch of modules in tandem, this env file could get messy. 

Ideally, you'd have a top level module that collects all secrets and distributes.

shykes:
@nipuna-perera yes the goal is to have a `.env`-style file where you explicitly set key-value pairs. That will be the next POC 😁

shykes:
Looking for my next incremental step. I see at least 3 well scoped tasks:

1) I need a new primitive in the API to parse objects from a string. This logic today exists in the CLI, for parsing command-line arguments (when passing directories, containers, secrets etc from the CLI). That logic needs to be moved to the engine, so that I can also use it to parse the contents of the env config file.

2) I need to refine the exact configuration mechanism. Is it `.env.dagger.json`? Or do we want to try and be compatible with actual `.env` files? Do we want to support findup, or only the current directory?

3) So far this is not tied to the `Env` type. But that's the goal. Do we try to make progress towards that in this PR? Or do we leave that for a follow-up?

FYI @jedevc @eunomie @TomChv @sipsma in case you're interested

shykes:
> 1. I need a new primitive in the API to parse objects from a string. This logic today exists in the CLI, for parsing command-line arguments (when passing directories, containers, secrets etc from the CLI). That logic needs to be moved to the engine, so that I can also use it to parse the contents of the env config file.

Work in progress: ~~https://github.com/dagger/dagger/pull/10753~~ #10770

shykes:
Update:

- #10770 is solid. I started building on top of it here
- More object types supported beyond `Secret`: `Directory`, `File`, `Service`, `GitRepository`, `GitRef`, `Container`...
- `.env.dagger.json` is optional
- ~~**BUG**: loading a module with a default object works, but *calling* it doesn't. Something wrong with the generated ID... [Discord discussion here](https://discord.com/channels/707636530424053791/1398005425353392208/1398006099747278849)~~ bug fixed thank you @vito 🙏

nipuna-perera:
> In this POC, you can pre-configure arguments to your module constructor

Is the plan to support non-constructor function args too? If not, is there a limitation there?

nadirollo:
I've just came to revisit this issue, as I was thinking of the design of our blueprints:

In our org, we are developing blueprints quite generic, so the consumers can tweak them to their needs (to certain extent). This turns out in blueprints with PLENTY of arguments (more than a normal module, where you will have everything defined in code inside the module definition.

If we were able to define defaults outside of the module in a `.env.dagger.json` the DevEx of blueprints would increase a lot. e.g.
`dagger init --blueprint=my-go-pipeline
setup `env.dagger.json` with the parameters relevant to my usage of the blueprint
Then, I can "just" call any of the functions that the blueprint exposes, without having to pass in all the params:
```
dagger test
vs
dagger --package-path path/to/go/package --token env:MY_TOKEN_FOR_PRIVATE_DEPS test
```
or
```
dagger deploy
vs
dagger --package-path path/to/go/package --token env:MY_TOKEN_FOR_PRIVATE_DEPS deploy --image-name my-team/my-image --clusters dev,stg
```

Without having to remember the many parameters required, and having them coded somewhere so I can call dagger locally.
At the moment, we find our engineers adding another abstraction (makefiles / justfiles) so they can put the parameters and can forget about all the arguments required to call their pipelines locally

Quite excited about this feature and if the possibility of unlocking defaults for functions, not just constructor... 👍

vmaffet-fast:
⬆️ Adding on Nadir's comments, I think this would help ensure dagger functions are called the same way locally and in CI. 

We currently see engineers use tools like `make` or `just` so that they do not need to rewrite a long dagger call. However, in CI the call goes through dagger directly, thus introducing a difference that can easily go unnoticed.

Having configurable defaults would go a long way in reducing the need to use these wrappers.

shykes:
Back from vacation, picking this back up

shykes:
> If we were able to define defaults outside of the module in a `.env.dagger.json` the DevEx of blueprints would increase a lot. e.g. `dagger init --blueprint=my-go-pipeline setup `env.dagger.json` with the parameters relevant to my usage of the blueprint

@nadirollo it seems there was a typo in your example snippet, could you elaborate on exactly what commands you would imagine being typed? Thank you!

vmaffet-fast:
> @nadirollo it seems there was a typo in your example snippet, could you elaborate on exactly what commands you would imagine being typed? Thank you!

From my last discussion with him, we were envisioning two types of commands.

**1.** A "setup" command that will save your default parameters into a `.env.dagger.json`. This could be done during the `dagger init` call maybe? e.g. 
```
dagger init --blueprint=<my-blueprint-ref> --defaults src=./infra/platform/app,image-name=example/platform-app
```
Which would output the following `.env.dagger.json`:
```json
{
  "src": "./infra/platform/app",
  "image-name": "example/platform-app"
}
```

**2.** The `call` commands then just works with no parameters provided (alla makefile)
e.g. Now we can do
```
dagger call test
```
instead of
```
dagger call --src ./infra/platform/app --image-name example/platform-app test
```


(Adding on why this is more important for blueprints is that normally we'd setup these defaults in the `// +default` section of an argument, but for blueprints, you will need to have these defaults for the specific project you want to build/test, so you'd alway need to pass in the source path etc. without a .env like file)

shykes:
@vmaffet-fast thanks for the details! That makes sense.

> (Adding on why this is more important for blueprints is that normally we'd setup these defaults in the // +default section of an argument, but for blueprints, you will need to have these defaults for the specific project you want to build/test, so you'd alway need to pass in the source path etc. without a .env like file)

For the specific case of source directory, you should still use `+default` since that will be interpreted in the context of each module using the blueprint. So using a `.env` would be redundant.

That said, you would still need a `.env` file for secrets and other values that can't be set in a default value.

nipuna-perera:
> That said, you would still need a `.env` file for secrets and other values that can't be set in a default value.

In my experience, most Dagger calls primarily require secrets. In an E2E pipeline, these secrets can include:

- Git access tokens
- Package manager credentials (tokens or username/password)
- Enterprise OAuth2 credentials (username/password)
- Tokens for static analysis tools
- Authentication tokens for deployment tools

Many tools are moving away from traditional username/password authentication and now require token-based authentication. This means we can no longer rely on LDAP credentials alone, which adds complexity and increases the number of flags we need to pass.

shykes:
@nipuna-perera agreed! Which is why this feature will be so helpful. I was just pointing out that it won't be as needed for blueprints + source directories.

shykes:
Hi all, I made a second POC which deprecates this one: #11000

shykes:
Continuing in #11034

PR #10708, opened 2025-07-09
dagql: fix case where session cache refs are never closed

sipsma:
If the session cache Release method was called while some calls were still running, it was possible for them to only be appended to the list of results after Release was already done, in which case they'd stay in the underlying cache and leak forever.

This adds handling to track when the session cache is closed and release any results that show up after that.

---

This seemed to be resulting in memory leaks. It wasn't perfectly repro-able, but was able to see leaks fairly often just after running tests and canceling them in the middle. I suspect that dag-op refs were more prone to it, likely just due to the timing of their execution in the bk solver, which is extra bad since that also results in leaks of on-disk cache refs.

After this change I can't repro that anymore. Also added some unit test coverage.

---

This change expanded in size quite a bit since adding error checks for use of session caches after they were closed resulted in error popping up all over the place in integ tests.

I ended up having to fix this by removing quite a few of our uses of `dagql.Server` stored on various structs.
* I took this quite a bit further but started getting mystery new errors. Wouldn't be surprised if I just made a typo somewhere in the very large diff, but I couldn't find it and want to get this merged ASAP, so there's files in `schema/` that probably should have similar updates but don't. The current ones are the ones that were strictly necessary for now.

### DIFF
--- a/.changes/unreleased/Fixed-20250709-184653.yaml
+++ b/.changes/unreleased/Fixed-20250709-184653.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Fix memory leak in engine
+time: 2025-07-09T18:46:53.359054779-07:00
+custom:
+  Author: sipsma
+  PR: "10708"
--- a/core/sdk/module_client_generator.go
+++ b/core/sdk/module_client_generator.go
@@ -18,15 +18,10 @@ type clientGeneratorModule struct {
 func (sdk *clientGeneratorModule) RequiredClientGenerationFiles(
 	ctx context.Context,
 ) (res dagql.Array[dagql.String], err error) {
-	query, err := core.CurrentQuery(ctx)
+	dag, err := sdk.mod.dag(ctx)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to get dag for sdk module %s: %w", sdk.mod.mod.Self().Name(), err)
 	}
-	dagqlCache, err := query.Cache(ctx)
-	if err != nil {
-		return nil, err
-	}
-	dag := sdk.mod.dag.WithCache(dagqlCache)
 
 	// Return an empty array if the SDK doesn't implement the
 	// `requiredClientGenerationFiles` function.
@@ -51,15 +46,10 @@ func (sdk *clientGeneratorModule) GenerateClient(
 	deps *core.ModDeps,
 	outputDir string,
 ) (inst dagql.ObjectResult[*core.Directory], err error) {
-	query, err := core.CurrentQuery(ctx)
+	dag, err := sdk.mod.dag(ctx)
 	if err != nil {
-		return inst, err
+		return inst, fmt.Errorf("failed to get dag for sdk module %s: %w", sdk.mod.mod.Self().Name(), err)
 	}
-	dagqlCache, err := query.Cache(ctx)
-	if err != nil {
-		return inst, err
-	}
-	dag := sdk.mod.dag.WithCache(dagqlCache)
 
 	schemaJSONFile, err := deps.SchemaIntrospectionJSONFile(ctx, []string{})
 	if err != nil {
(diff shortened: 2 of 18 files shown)


marcosnils (approved):

sipsma:
Looking into the test failures, seems like sometimes some APIs are using a cache that has already been closed, which turned into an error case as part of this change and is actually caught now...

sipsma:
This is turning into a pretty large scale update needed; we are using `dagql.Server`s stored all over the place which are from previous sessions and thus triggering the new error. Working through it all, just going to be a pretty huge diff.

sipsma:
This should be good to go now

.changes/unreleased/Fixed-20250709-184653.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixed
+body: Fix memory leak in engine
```
jedevc, line 2: :tada:

jedevc:
> This is turning into a pretty large scale update needed; we are using `dagql.Server`s stored all over the place which are from previous sessions and thus triggering the new error. Working through it all, just going to be a pretty huge diff.

Curious how this is possible? Why are we getting servers from previous sessions?

core/sdk/module_client_generator.go
```diff
@@ -41,6 +51,16 @@ func (sdk *clientGeneratorModule) GenerateClient(
...
 ) (inst dagql.ObjectResult[*core.Directory], err error) {
+	query, err := core.CurrentQuery(ctx)
+	if err != nil {
+		return inst, err
+	}
+	dagqlCache, err := query.Cache(ctx)
+	if err != nil {
+		return inst, err
+	}
+	dag := sdk.mod.dag.WithCache(dagqlCache)
```
jedevc: Maybe struggling to understand what this is doing - it's putting the right session cache in, but how can it end up being the wrong one?

Might be worth splitting it out to a `sdk.mod.Dag(ctx)` or something?

sipsma: Yeah the SDKs are a little special because of a combination of
1. they *need* a `dagql.Server` where the SDK API has been `.Install()`'d
2. the calls are cached persistently (to make caching effective for module loading), which includes across sessions in the dagql cache
3. `dagql.SessionCache` associated with a given `dagq.Server` is released at the end of a session

So what can go wrong is
1. we make dagql calls that return objects which reference `sdk`, no cache hit
2. the calls are made, session closes and the `SessionCache` is released
3. new calls come in that get cache hits and the same objects from before that reference `sdk`
4. the `SessionCache` being used has already been released, isn't the one from the current session

So the idea here is to use the *schema* of the `dagql.Server` where the SDK API was installed, but with the current session's cache.

`sdk.mod.Dag(ctx)` sounds good to me, will update to that.

dagql/session_cache.go
```diff
@@ -59,6 +64,14 @@ func WithTelemetry(telemetry TelemetryFunc) CacheCallOpt {
 	})
 }
 
+// TODO:
+// TODO:
+// TODO:
+// TODO:
```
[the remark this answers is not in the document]
sipsma: I swear I always ctrl-F `TODO:` and always miss one somehow....

sipsma:
> Curious how this is possible? Why are we getting servers from previous sessions?

@jedevc What I described [here](https://github.com/dagger/dagger/pull/10708#discussion_r2208241317) is one example, but there's basically just a ton of other variations on that. A lot of it is related to module loading since that's where we construct different `dagql.Server`s with different APIs installed. I pretty much just kept running integ tests and replaced usage of `dagql.Server` from structs w/ grabbing it from `context` until the errors went away.

jedevc (approved):

jedevc:
Makes sense, related to module loading seems reasonable, I was just worried about it being *every* stored dagql server on a `.*Schema` somehow ending up being wrong (which would be quite wrong, if that was the case).

PR #10709, opened 2025-07-10
chore: tidy up engine startup and connection telemetry

jedevc:
This patch improves the telemetry at client startup in a few ways:
- Earlier notification of cloud telemetry info - this ensures we can open the web view *before* the engine is provisioned.
- Adds logging info as a __real__ OTEL span, instead of a virtual log. This means that we can get metadata about the versions we're actually connecting to in cloud (previously missing).
- Correctly nests the image store configuration under the "connecting to engine" span (which requires defering the load operation a little bit).

### DIFF
--- a/dagql/idtui/frontend_pretty.go
+++ b/dagql/idtui/frontend_pretty.go
@@ -158,10 +158,6 @@ func (fe *frontendPretty) Shell(ctx context.Context, handler ShellHandler) {
 	<-ctx.Done()
 }
 
-func (fe *frontendPretty) ConnectedToEngine(ctx context.Context, name string, version string, clientID string) {
-	// noisy, so suppress this for now
-}
-
 func (fe *frontendPretty) SetCloudURL(ctx context.Context, url string, msg string, logged bool) {
 	if fe.OpenWeb {
 		if err := browser.OpenURL(url); err != nil {
--- a/engine/client/drivers/docker.go
+++ b/engine/client/drivers/docker.go
@@ -86,7 +86,7 @@ const InstrumentationLibrary = "dagger.io/client.drivers"
 // are identified by looking for containers with the prefix
 // "dagger-engine-").
 func (d *dockerDriver) create(ctx context.Context, imageRef string, containerName string, volumeName string, cleanup bool, port int, opts *DriverOpts) (helper *connh.ConnectionHelper, rerr error) {
-	ctx, span := otel.Tracer("").Start(ctx, "create")
+	ctx, span := otel.Tracer("").Start(ctx, "create container")
 	defer telemetry.End(span, func() error { return rerr })
 	slog := slog.SpanLogger(ctx, InstrumentationLibrary)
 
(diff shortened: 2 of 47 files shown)


dagql/idtui/frontend_pretty.go
```diff
@@ -158,10 +158,6 @@ func (fe *frontendPretty) Shell(ctx context.Context, handler ShellHandler) {
 	<-ctx.Done()
 }
 
-func (fe *frontendPretty) ConnectedToEngine(ctx context.Context, name string, version string, clientID string) {
-	// noisy, so suppress this for now
-}
```
jedevc, line 163: This seems to be a relic of before we really had `Enscapsulated` spans. We can how have our cake (have an OTEL span) and eat it too (not show it to the user by default) :grin:

jedevc:
Also pushed a commit to avoid printing the full path to the dagger binary, especially for `dagger session` calls, this is very annoying (and leaks info about the layout of the user's system).

engine/client/drivers/docker.go
```diff
@@ -124,7 +124,7 @@ func (d *dockerDriver) create(ctx context.Context, imageRef string, containerNam
 	}
 
 	// ensure the image is pulled
-	if _, err := traceexec.Exec(ctx, exec.CommandContext(ctx, "docker", "inspect", "--type=image", imageRef), telemetry.Encapsulated()); err != nil {
+	if _, err := traceexec.Exec(ctx, exec.CommandContext(ctx, "docker", "inspect", "--type=image", imageRef, "--format", "{{ .ID }}"), telemetry.Encapsulated()); err != nil {
```
vito, line 127: cc @jpadams - iirc `--format` (in)compatibility came up when investigating Apple `container` compatibility, not sure what state we're in

engine/client/imageload/docker.go
```diff
@@ -22,9 +22,12 @@ func (Docker) ID() string {
...
 }
 
-func (loader Docker) Loader(ctx context.Context) (*Loader, error) {
+func (loader Docker) Loader(ctx context.Context) (_ *Loader, rerr error) {
+	ctx, span := otel.Tracer("").Start(ctx, "check for docker daemon")
+	defer telemetry.End(span, func() error { return rerr })
+
 	// check docker is running
-	cmd := exec.CommandContext(ctx, "docker", "info")
+	cmd := exec.CommandContext(ctx, "docker", "info", "--format", "{{.ServerVersion}}")
```
vito: cc @jpadams here too

vito (commented):
Looks like the Telemetry tests need to be updated, with some new scrubbers to clean up the random engine names. Plus some questions for @jpadams but I suspect those are OK since nothing's reading the output in a structured way anyhow.

engine/client/drivers/docker.go
jedevc: Atm, for apple `container` functionality, we require using `docker-container` directly, there's no automatic lifecycle management (due to other compat).

I'm working on proper lifecycle management in https://github.com/dagger/dagger/pull/10714 with @kpenfound :tada:

engine/client/imageload/docker.go
jedevc: Mm, this one does affect `docker-container`, will fix this one.

Edit: will make sure we get it for https://github.com/dagger/dagger/pull/10714, that way we're not duplicating logic.

jedevc:
Going to rebase this onto and merge this after https://github.com/dagger/dagger/pull/10714, since there's a *lot* of conflicty things.

engine/client/imageload/docker.go
jedevc: Removed this. It's going away in #10714, so no need for it here.

PR #10712, opened 2025-07-10
POC: blueprint modules

shykes:
- **A module may reference another as its blueprint.**
- **fix install**


chrisjpalmer:
Thanks @shykes this looks perfect. Gave it a test and very much like it.

shykes:
I added the ability to update the blueprint with `dagger update`

core/schema/modulesource.go
```diff
@@ -2199,6 +2363,12 @@ func (s *moduleSourceSchema) moduleSourceAsModule(
 		}
 	}
 
+	// FIXME: If using a blueprint, restore target context and name for file operations
```
jedevc: Do you mean `dag.CurrentModule().Source()`?

Not sure what's currently not working.

cmd/dagger/module.go
```diff
@@ -277,6 +282,30 @@ If --sdk is specified, the given SDK is installed in the module. You can do this
...
+				// Install the blueprint
+				modSrc = modSrc.WithBlueprint(blueprintSrc)
+				_, err = modSrc.
+					WithEngineVersion(modules.EngineVersionLatest).
+					GeneratedContextDirectory().
+					Export(ctx, contextDirPath)
+				if err != nil {
+					return fmt.Errorf("failed to install blueprint: %w", err)
+				}
+				fmt.Fprintln(cmd.OutOrStdout(), "Installed blueprint", initBlueprint)
```
jedevc: suggestion: merge this output with the previous? Then we get a nice one-liner:

```
Initialized module example in /tmp/foo with blueprint github.com/shykes/x/go
```

core/module.go
```diff
@@ -22,6 +22,9 @@ type Module struct {
 	// The source of the module
 	Source dagql.ObjectResult[*ModuleSource] `field:"true" name:"source" doc:"The source for the module."`
 
+	// The source to load contextual dirs/files from, which may be different than Source for blueprints
+	ContextSource dagql.ObjectResult[*ModuleSource]
```
jedevc, line 26: There's a few places where I'm wondering if we should use this instead:
- Analytics: should analytics show the blueprint (`Source`), or the module (`ContextSource`)?
- LLM prompts: same question.

Leaning towards using `ContextSource` in both cases.

core/schema/modulesource.go
```diff
@@ -783,12 +820,23 @@ func (s *moduleSourceSchema) initFromModConfig(configBytes []byte, src *core.Mod
...
 
+	// blueprint is incompatible with some dagger.json fields
+	if src.Blueprint.Self() != nil {
+		if src.SDK != nil {
+			return fmt.Errorf("blueprint and sdk can't both be set")
+		}
+		if src.Dependencies.Len() != 0 {
+			return fmt.Errorf("blueprint and dependencies can't both be set")
+		}
+	}
```
jedevc: SGTM :tada:

Should the `source` field also be incompatible here?

cmd/dagger/module.go
```diff
@@ -277,6 +282,30 @@ If --sdk is specified, the given SDK is installed in the module. You can do this
...
+
+				// Install the blueprint
+				modSrc = modSrc.WithBlueprint(blueprintSrc)
+				_, err = modSrc.
+					WithEngineVersion(modules.EngineVersionLatest).
+					GeneratedContextDirectory().
+					Export(ctx, contextDirPath)
+				if err != nil {
+					return fmt.Errorf("failed to install blueprint: %w", err)
+				}
```
jedevc: suggestion: modify `modSrc` above, so we're not doing two different calls to `Export`.

core/schema/modulesource.go
```diff
@@ -444,6 +456,10 @@ func (s *moduleSourceSchema) localModuleSource(
 			return inst, err
 		}
 
+		if err := s.loadBlueprintModule(ctx, bk, localSrc); err != nil {
+			return inst, err
+		}
```
jedevc: nit: put this into the errorgroup below? to match the other loading code.

core/integration/blueprint_test.go
```diff
@@ -0,0 +1,270 @@
...
+		out, err = modGen.
+			With(daggerExec("call", "get-actual-module-name")).
+			Stdout(ctx)
+		require.NoError(t, err)
+		require.Contains(t, out, "target-context-test")
+		require.NotContains(t, out, "blueprint-context-test")
+	})
+}
+
+func (BlueprintSuite) TestBlueprintConfigurationPersistence(ctx context.Context, t *testctx.T) {
```
jedevc: Tests LGTM (haven't looked too closely), but could you add a test for a blueprint with a dependency? Just feels like an edge case that might *potentially* break in the future if we don't have a test :pray:

chrisjpalmer:
Hey there guys. I have been doing some testing on this branch to verify it for the use case we have at nine. You may have seen me spamming the discord earlier with some stuff about "dependency injection" and interfaces... At the moment we are experimenting with dagger shell and injecting types in to certain modules ( which we call "workflow modules" ) to allow us to seperate the concerns of the dependencies out. 

I set up a[ demo of what we are trying to achieve](https://github.com/chrisjpalmer/dagger-deps-example).

When testing this out on 0.18.12, the demo was working fine. You can run `dagger -c 'workflow-pretend $(secrets-manager) | doit'` and it works without issues. Although this demo doesn't contain anything related to blueprint modules, I decided to test it against this branch of dagger anyway:

```sh

dagger -m github.com/shykes/dagger@module-blueprint <<EOF
  dev |
  with-directory example \$(git https://github.com/chrisjpalmer/dagger-deps-example | head | tree) |
  with-workdir example |
  with-exec -- dagger -c 'workflow-pretend \$(secrets-manager) | doit'
EOF

```

In this case I am seeing a weird error:

```
  ╰─✘ workflowPretend(
      ┆ sm: secretsManager: SecretsManager!
      ): WorkflowPretend! 0.2s ERROR
    ! instantiate: toSelectable: unknown type "SecretsManager"
```

Am I doing something wrong haha?

Thank you!

core/module.go
shykes: I don't know how either of those work, so don't have an informed opinion.

core/schema/modulesource.go
shykes: That's an old FIXME, will remove

core/integration/blueprint_test.go
shykes: @jedevc is it enough for the blueprint to have a dependency without calling it (to make sure loading all works), or do you want to test the ability to call the dependency also?

cmd/dagger/module.go
shykes: Done ✅

core/schema/modulesource.go
shykes: Done ✅

core/schema/modulesource.go
shykes: Done ✅

core/integration/blueprint_test.go
shykes: Done ✅

cmd/dagger/module.go
shykes: done ✅

shykes:
Thank you for the review @jedevc 🙏 

I applied all your suggestions, except for the analytics annotation, because I don't know how to do it. If that's OK, let's keep that for a follow-up?

chrisjpalmer:
Did some more testing and confirmed that my example works on v0.18.12 but not on the latest main branch:

### dagger/dagger@v0.18.12

Trace: https://dagger.cloud/nine/traces/59c4e0afbc3271db9f6eb3a5baf855cd

```
dagger -m github.com/dagger/dagger@v0.18.12 <<EOF
  dev |
  with-directory example \$(git https://github.com/chrisjpalmer/dagger-deps-example | head | tree) |
  with-workdir example |
  with-exec -- dagger -c 'workflow-pretend \$(secrets-manager) | doit'
EOF
```

### dagger/dagger@main

Trace: https://dagger.cloud/nine/traces/35e514dd9cad792f2b3277754e86c908
```
dagger -m github.com/dagger/dagger@main <<EOF
  dev |
  with-directory example \$(git https://github.com/chrisjpalmer/dagger-deps-example | head | tree) |
  with-workdir example |
  with-exec -- dagger -c 'workflow-pretend \$(secrets-manager) | doit'
EOF
```

... will raise this in the discord given that this isn't really related to this branch

cmd/dagger/module.go
```diff
@@ -161,8 +165,9 @@ var moduleInitCmd = &cobra.Command{
 This creates a dagger.json file at the specified directory, making it the root of the new module.
 
 If --sdk is specified, the given SDK is installed in the module. You can do this later with "dagger develop".
+If --blueprint is specified, the given blueprint is installed in the module.
 `,
-	Example: "dagger init --sdk=python",
```
sipsma, line 165: nit: not sure if it makes sense to remove `--sdk` from the example help text entirely vs. leaving it and adding `--blueprint`

core/module.go
sipsma: > Analytics: should analytics show the blueprint (Source), or the module (ContextSource)?

I feel like analytics should show the actual module wrapping the blueprint as opposed to the "underlying" blueprint, right? Otherwise every module using a `Go` blueprint would be the same. So then wouldn't we want to use `Source` for that case at least?

sipsma (approved):
nits but LGTM

cmd/dagger/module.go
shykes: I agree, I just wasn't sure how to do that... Is it OK to make the example string multi-line?

shykes:
> Did some more testing and confirmed that my example works on v0.18.12 but not on the latest main branch:

Thanks for checking on `main`... I was pulling my hair out trying to understand how this PR could cause your issue 😅

cmd/dagger/module.go
shykes: Done ✅

core/module.go
shykes: How do I do that?

jedevc: Pushed it in 306fef8b36eba3d7212ef44f8705476f71283c2d.

jedevc (approved):

Issue #10676, opened 2025-07-02
Proposal: better support for other local container runtimes


jedevc:
Today, we only really officially support the docker runtime for auto-starting the dagger engine.

Ideally though, [we should support more](https://github.com/dagger/dagger/pull/10657/files#r2174855222), to name a few:
- [ ] Containerd (for `nerdctl`, `finch`, etc).
- [ ] Apple's `container`.
- [ ] `podman`

These should all support `{driver}-container` and `{driver}-image` to support all the scenarios we support with docker today.

> [!IMPORTANT]
>
> All of the problems with garbage collection, etc, that we have today with the `docker` driver will be inherited - fixing those would be out-of-scope (for now).

We should also look into supporting a non-experimental environment variable to select the default local driver: something like `DAGGER_DRIVER`, which would be set to `docker` (the default, today's behavior), or do any of the above (e.g. `containerd`, `apple`, `podman`, etc). `_EXPERIMENTAL_DAGGER_RUNNER_HOST` would still take precedent if it was set.

jedevc:
Slightly reconsidered while working on autoselection in https://github.com/dagger/dagger/pull/10714.

We want to *support* all of these still, but I think it makes more sense to rename: `{driver}-image://registry.dagger.io/engine` should instead be `image+{driver}://registry.dagger.io/engine`.

That way, a generic `image://registry.dagger.io/engine` is free to be an auto-selection from all the available image drivers.

PR #10714, opened 2025-07-11
Support for more driver types

jedevc:
Fixes:
- https://github.com/dagger/dagger/issues/10676

TODO:
- [x] Nerdctl / Finch
  - [x] Integration tests
  - [ ] Manual sanity checks
- [x] Podman
  - [x] Integration tests
  - [x] Manual sanity checks
- [ ] Apple `container`

### DIFF
--- a/core/integration/container_test.go
+++ b/core/integration/container_test.go
@@ -3719,7 +3719,7 @@ func (ContainerSuite) TestInsecureRootCapabilitesWithService(ctx context.Context
 	// testing it can startup, create containers and bind mount from its filesystem to
 	// them.
 	randID := identity.NewID()
-	dockerc := dockerSetup(ctx, t, t.Name(), c, "23.0.1", middleware)
+	dockerc := dockerSetup(ctx, t, c, containerSetupOpts{name: "provisioner", version: "23.0.1", middleware: middleware})
 	out, err := dockerc.
 		WithExec([]string{"sh", "-e", "-c", strings.Join([]string{
 			fmt.Sprintf("echo %s-from-outside > /tmp/from-outside", randID),
@@ -4295,8 +4295,7 @@ func (ContainerSuite) TestImageLoadCompatibility(ctx context.Context, t *testctx
 	c := connect(ctx, t)
 
 	for _, dockerVersion := range []string{"20.10", "23.0", "24.0"} {
-		dockerc := dockerSetup(ctx, t, t.Name(), c, dockerVersion, nil)
-
+		dockerc := dockerSetup(ctx, t, c, containerSetupOpts{name: t.Name(), version: dockerVersion})
 		for _, mediaType := range []dagger.ImageMediaTypes{dagger.ImageMediaTypesOcimediaTypes, dagger.ImageMediaTypesDockerMediaTypes} {
 			mediaType := mediaType
 			for _, compression := range []dagger.ImageLayerCompression{dagger.ImageLayerCompressionGzip, dagger.ImageLayerCompressionZstd, dagger.ImageLayerCompressionUncompressed} {
@@ -5217,7 +5216,7 @@ func (ContainerSuite) TestSymlinkCaching(ctx context.Context, t *testctx.T) {
 
 func (ContainerSuite) TestLoadDocker(ctx context.Context, t *testctx.T) {
 	c := connect(ctx, t)
-	dockerc := dockerSetup(ctx, t, "provisioner", c, "", nil)
+	dockerc := dockerSetup(ctx, t, c, containerSetupOpts{name: "provisioner"})
 	dockerc, err := dockerLoadEngine(ctx, c, dockerc, "registry.dagger.io/engine:dev")
 	require.NoError(t, err)
 
@@ -5277,7 +5276,7 @@ func (ContainerSuite) TestLoadDocker(ctx context.Context, t *testctx.T) {
 
 func (ContainerSuite) TestLoadContainerd(ctx context.Context, t *testctx.T) {
 	c := connect(ctx, t)
-	nerdctl := nerdctlSetup(ctx, t, "provisioner", c, "v2.1.2")
+	nerdctl := nerdctlSetup(ctx, t, c, containerSetupOpts{name: "provisioner", version: "v2.1.2"})
 	nerdctl, err := nerdctlLoadEngine(ctx, c, nerdctl, "registry.dagger.io/engine:dev")
 	require.NoError(t, err)
 
@@ -5313,7 +5312,7 @@ func (ContainerSuite) TestLoadContainerd(ctx context.Context, t *testctx.T) {
 
 func (ContainerSuite) TestLoadNone(ctx context.Context, t *testctx.T) {
 	c := connect(ctx, t)
-	dockerc := dockerSetup(ctx, t, "provisioner", c, "", nil)
+	dockerc := dockerSetup(ctx, t, c, containerSetupOpts{name: "provisioner"})
 	dockerc, err := dockerLoadEngine(ctx, c, dockerc, "registry.dagger.io/engine:dev")
 	require.NoError(t, err)
 
@@ -5342,7 +5341,7 @@ func (ContainerSuite) TestLoadNone(ctx context.Context, t *testctx.T) {
 func (ContainerSuite) TestLoadInNested(ctx context.Context, t *testctx.T) {
 	// this shouldn't be possible! we shouldn't allow access to the external client.
 	c := connect(ctx, t)
-	dockerc := dockerSetup(ctx, t, "provisioner", c, "", nil)
+	dockerc := dockerSetup(ctx, t, c, containerSetupOpts{name: "provisioner"})
 	dockerc, err := dockerLoadEngine(ctx, c, dockerc, "registry.dagger.io/engine:dev")
 	require.NoError(t, err)
 
--- a/core/integration/provision_test.go
+++ b/core/integration/provision_test.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"cmp"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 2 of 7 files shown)


shykes:
My main objection is that all these drivers could be shell scripts. And they would be easier to read...

So shouldn't we just implement a simple shell hook system, convert these to builtin scripts, and allow users to configure their own script for customization?

shykes:
Also I think tight coupling of "which image to run" with "how to run it" into a single URI, is an anti-pattern and we should unwind it, not double down on it.

jpadams:
From `gh pr checkout 10714`, used `./hack/dev` and then `dagger -c 'cli | binary --platform current | export ./bin/dagger'`

to get an engine image and CLI binary

Tried it with `docker-image://` and worked fine.

Then I did a `docker save localhost/dagger-engine.dev > lde.tar`
and `container image load --input lde.tar` to get the image into Apple Container image store (after starting Apple Container with `container system start`, of course).

I moved my `docker` binary out of `PATH` and changed to generic `image://`
```
_EXPERIMENTAL_DAGGER_RUNNER_HOST=image://localhost/dagger-engine.dev
```
and `dagger` found `container` and used it instead of `docker`.

https://dagger.cloud/dagger/traces/befd72d07de264b809e1054b8dacc3ed?listen=9649ddef87d339c8&listen=4b1aa3a83f946106&listen=384104c699d5393c&listen=fc6c55f22bd33231&listen=b5d16a643ac81ba3&listen=bafe98580449d3e1

I did notice that if I stopped things with
```
container system stop
```
then we don't try to `container system start` which is consistent with us not starting the Docker system either.

jpadams:
> My main objection is that all these drivers could be shell scripts. And they would be easier to read...
> So shouldn't we just implement a simple shell hook system, convert these to builtin scripts, and allow users to configure their own script for customization?

I'm thinking that the Go interface and Go implementation might be easier to write and test, though.
https://github.com/dagger/dagger/pull/10714/files#diff-44e3a80cf8531b5f0b54d30da7246ede52a041879cf098dbe4be3c7cbd8b29b7R86
Also, there won't likely be a ton of these.

@shykes if we merged this, and we really wanted to get off Go implementations of drivers in `dagger/dagger` and move to an open shell-script approach, looks like it would be pretty contained surgery. So, I'd like to see us merge this, try it, get feedback and revisit later. Seems like most folks will use built-in defaults. More on that below.

> Also I think tight coupling of "which image to run" with "how to run it" into a single URI, is an anti-pattern and we should unwind it, not double down on it.

I was using the `container://` (or `image://`) default that would come with the `dagger` CLI, but I see it as nice to be able to specify `container+apple://` if I'm running multiple container runtimes. In this test, I had `docker` binary back in place in `PATH` and had Docker Desktop running when I used
```
export _EXPERIMENTAL_DAGGER_RUNNER_HOST=image+apple://localhost/dagger-engine.dev
```
and successfully landed on a Dagger Engine in Apple Container instead of Docker taking precedence (or me not knowing the precedence order).
https://dagger.cloud/dagger/traces/510d83d5e82e3516bf42ff60000859f2?listen=11cf77d3250198cd&listen=3cfa9345bddbf4e4&listen=2905deb4217a82a1&listen=acb9696477787755&listen=10b587e2e6021a8d

engine/client/drivers/apple.go
```diff
@@ -0,0 +1,153 @@
...
+			envs = append(envs, env)
+		} else {
+			args = append(args, "-e", env)
+		}
+	}
+	for _, port := range opts.ports {
+		return fmt.Errorf("unsupported port argument %q", port)
+	}
+
+	if opts.cpus != "" {
```
jpadams, line 79: @jedevc how do I set these `opts` like `cpus` and `memory`?

jpadams: Oh,
Flow for Memory and CPU Options

  1. URL Query Parameters: Options like cpus and memory are passed as URL query parameters
  when provisioning a driver (container.go:144-145)
  2. containerCreateOpts Structure: These are extracted into a containerCreateOpts struct
  (container.go:218-219)
  3. runOpts Structure: The options are then passed to a runOpts struct which contains:
    - cpus string (container.go:112)
    - memory string (container.go:113)
  4. Apple Driver Implementation: In apple.go:79-90, the Apple driver handles these options:
  
  so like this works:
  `?cpus=8&memory=16G`
  
  ```
  export _EXPERIMENTAL_DAGGER_RUNNER_HOST='image+apple://localhost/dagger-engine.dev?cpus=8&memory=16G'
  ```

```
"resources": {
        "memoryInBytes": 17179869184,
        "cpus": 8
      },
```
but how to set `cpus=8&memory=16G` as defaults without using the query param approach in the special env var? @jedevc 🤔

jpadams: Also, subsequent runs with `image://` will use existing engine, even if you fall back to the defaults for memory and cpu or set them explicitly to different values. Need to kill the existing engine to get a new one with `image://`. If using `container://` then you'd be bound to a particular instance with a particular mem and cpu already chosen.

shykes:
@jpadams but how do I configure dagger to use `container` *without* forcing a particular image to be provisioned? In your case `dagger-engine.dev`. I still want the CLI to handle choosing the image for me.

jpadams:
> @jpadams but how do I configure dagger to use `container` _without_ forcing a particular image to be provisioned? In your case `dagger-engine.dev`. I still want the CLI to handle choosing the image for me.

I was doing this for testing to get a dev engine and CLI, but once merged/released, I believe you just use `dagger` and then we loop through the list of runtimes in order and use the first one found with an engine from our registry*:
https://github.com/dagger/dagger/pull/10714/files#diff-e79ed504b420e6690e2d5752b329ef0bae57eb8c1b0b27cc237f0ef9f3f7aaecR55-R61

The order being the order in which they were registered:
https://github.com/dagger/dagger/pull/10714/files#diff-44e3a80cf8531b5f0b54d30da7246ede52a041879cf098dbe4be3c7cbd8b29b7R26-R59

https://github.com/dagger/dagger/pull/10714/files#diff-e79ed504b420e6690e2d5752b329ef0bae57eb8c1b0b27cc237f0ef9f3f7aaecR44-R48

1. `docker`
2. Apple `container`
3. `podman`
4. `finch`
5. `nerdctl`

*with the added detail of accepting the overrides of the `_EXPERIMENTAL_DAGGER_RUNNER_HOST`

sipsma (approved):
In terms of code this LGTM, so approving to unblock if the other discussions resolve w/out any major changes needed.

> if we merged this, and we really wanted to get off Go implementations of drivers in dagger/dagger and move to an open shell-script approach, looks like it would be pretty contained surgery. So, I'd like to see us merge this, try it, get feedback and revisit later. Seems like most folks will use built-in defaults. More on that below.

I agree with this. I definitely don't mind the idea of shelling out to a binary that accepts commands/args implementing the interface we currently have in go (so that would include shell scripts and any other executable).

But that would be a bigger independent lift, so I'd rather merge this now for its improvements and switch to a different approach in a follow-up.

> but how do I configure dagger to use container without forcing a particular image to be provisioned? In your case dagger-engine.dev. I still want the CLI to handle choosing the image for me.

To just use an existing container you'd do `container://<container name>` (or `container+apple://<container name>`, `container+docker://<container name>`, etc.)

When you specify `image://<container name>` or one of the runtime-specific variants, you are telling the CLI to do the whole provisioning of that image and then basically the same thing as `container://` to actually connect.

I think this works pretty well. `image://` is just a layer on top of `container://` that does the extra provisioning too, but if you just want to connect to a `container://` that's just as easy too.

shykes:
@sipsma I'm fine with merging as-is, but IMO there's another set of changes we need to make before we can stabilize (eg. make the variable not experimental).

This is because IMO, everything covered by these drivers is for the special case: engine developers and power users who don't mind "voiding the warranty" by provisioning the engine themselves.

What's missing is a way for regular users, who want to *keep auto-provisoning* (ie. who *don't* want to specify either an image or a contianer), to configure whether to use podman, docker, container etc.


TLDR: as long as everyone is OK with this being potentially broken one more time before we stabilize, I'm OK with merging.

I wanted to take on that extra layer of change this week, but didn't have the time - sorry about that, would have preferred to bring a solution along with my complaining that we aren't done :)

jpadams:
> What's missing is a way for regular users, who want to keep auto-provisoning (ie. who don't want to specify either an image or a contianer), to configure whether to use podman, docker, container etc.

@shykes agree, that is required if the regular users have multiple container runtimes.
If they have a single supported one installed, up and running, it should be used automatically.

 In addition there needs to be a way for auto-provisioning users to specify certain settings like how much memory and cpu to give to the Apple Container micro-vm.

Today you can do that in the experimental RUNNER var's query params, but not otherwise.

I can imagine these all being env vars, or there being some global Dagger config file that is consulted, e.g. something in `~/.config/dagger/engine.json`. 

Hmmm...have to see if that would work.
https://docs.dagger.io/configuration/engine/#configuration

@jedevc 
I'd think it would be something like this in `engine.json`:
```
{
  "logLevel": "info",
  "container": {
    "runtime": "apple",
    "cpus": "6",
    "memory": "12G"
  }
}
```

Ideally with `dagger config` or similar commands to get/set `engine.json`.

shykes:
My general feeling is that we are allowing our engine driver system to become a hugely complicated beast. There's layers and layers of indirections here... All this to wrap a couple execs with a half dozen flags. Seems *way* overengineered to me (not just this PR, the whole sequence of gradual expansion in scope and complexity, culminating in this PR)

shykes:
FYI @jedevc @jpadams @sipsma I have submitted a POC of a config system I would be happy to stabilize. Note that *it is built on top of this PR* to maximize continuity :) Would love to get your thoughts.

https://github.com/dagger/dagger/pull/10790

jedevc:
> > if we merged this, and we really wanted to get off Go implementations of drivers in dagger/dagger and move to an open shell-script approach, looks like it would be pretty contained surgery. So, I'd like to see us merge this, try it, get feedback and revisit later. Seems like most folks will use built-in defaults. More on that below.
> 
> I agree with this. I definitely don't mind the idea of shelling out to a binary that accepts commands/args implementing the interface we currently have in go (so that would include shell scripts and any other executable).

I disagree. I think provisioning is complicated - the shell script isn't just going to be a single `docker run` command, it needs to have different args provided, it needs multiple entrypoints for `run`ing, `connect`ing, etc - essentially the interface here: https://github.com/dagger/dagger/blob/50aad4d4f2762ada01d414dfa0332d8c151072b5/engine/client/drivers/container.go#L82-L101

There's a lot of effort done to try and handle *slight* incompatibilities between docker/podman/nerdctl and other docker-compatible CLIs. We need to parse stdout output, etc.

There's also engine deletion - but the future of that should be up for debate as well, but that logic is also quite complex.

I'm worried all we'll do is just take the complexity of drivers and put them into bash scripts, instead of go.

> TLDR: as long as everyone is OK with this being potentially broken one more time before we stabilize, I'm OK with merging.

Not really super hyped about doing this, even though it's "`EXPERIMENTAL`" - as I've pointed out before, even though it's "`EXPERIMENTAL`", users rely on it everywhere, and it's central to our integration docs. Let's work out what we're actually doing before merging here, so we're not going to break things for users - it would have been good to have this discussion before starting on adding support for `container`/etc, but the second best time is now.

jedevc:
> I can imagine these all being env vars, or there being some global Dagger config file that is consulted, e.g. something in ~/.config/dagger/engine.json.

:100: agree, I think getting these into `engine.json` is the best place for these.

jpadams:
From @shykes:
> What's missing is a way for regular users, who want to keep auto-provisoning (ie. who don't want to specify either an image or a contianer), to configure whether to use podman, docker, container etc.
> 
> TLDR: as long as everyone is OK with this being potentially broken one more time before we stabilize, I'm OK with merging.

From @jedevc:
> Not really super hyped about doing this, even though it's "EXPERIMENTAL" - as I've pointed out before, even though it's "EXPERIMENTAL", users rely on it everywhere, and it's central to our integration docs. Let's work out what we're actually doing before merging here, so we're not going to break things for users

If we agree @jedevc that a good way forward here is `engine.json` config for the few values we need to be able to set, then shall we implement that?

It seems like with this PR and the engine.json config (possibly with env var overrides) and keeping (though deprecated) the existing `_EXPERIMENTAL_DAGGER_RUNNER_HOST` and `docker-container://` etc , then we're not leaving anyone in the lurch, right?

> some global Dagger config file that is consulted, e.g. something in ~/.config/dagger/engine.json.

> 💯 agree, I think getting these into engine.json is the best place for these.

jedevc:
Reworked the changelog, and removing the `needs/docs` label.

It might break again :cry: but deciding to bias towards adding support for the other runtimes asap. At the very least, we'll preserve the "automatic" selection, even if we change the behavior of `_EXPERIMENTAL_DAGGER_RUNNER_HOST`.

Let's continue the discussion of the future of the drivers in #10790.

(there's other work to do on top/around this, e.g. https://github.com/dagger/dagger/pull/10810, and I want to avoid blocking that)

engine/client/drivers/apple.go
jedevc: Yup!

There is no way except to modify the query string - I think we do need an easier way, but this should be in the context of #10790? I don't want to invent more env vars/configs, until we decide what to do there.

> Also, subsequent runs with `image://` will use existing engine

Mmm, conflicted on desirable behavior here. It's not perfect atm, but agreed it's a bit weird that you can get a holdover engine that's not quite right.

I guess to fix this, when we search for leftover engines we should match by cpu/memory as well. I might tackle this as a followup?

shykes: I will propose a solution in #10790 . Spoiler alert it will involve taking less responsibility for platform-specific customizations.

Issue #6990, opened 2024-04-02
✨ Add support for live development with running Dagger services 


levlaz:
### What are you trying to do?

I want to be able to make changes to code locally and see them reflected in my running Dagger services similar to how docker-compose and docker run `--v` works. 

Discussion about this can be found here: https://discord.com/channels/707636530424053791/1224583500569514024/1224583500569514024

### Why is this important to you?

The introduction of `*Service` has made it really easy to get a one liner dev environment up and running on my local laptop using the dagger cli without having to install any local dependencies. Supporting the ability to hot reload would unlock a whole new category of powerful local development use cases for all dagger users. 

### How are you currently working around this?

There is no workaround right now. :(

sheldonhull:
Is there any alternative to this right now other than a live file watcher rerunning dagger commands? 
Being able to replace my wrapped up docker calls for live reloading of site and other tooling is critical for my further adoption. Without this most of the local service stuff that docker compose, or my wrapped up magefile style docker commands can't be replaced without losing out on the development experience.

levlaz:
Unfortunately there is no other workaround right now. 

> Is there any alternative to this right now other than a live file watcher rerunning dagger commands? 

I had not thought of this approach, we are planning to make some
optimizations to dagger cli startup time which may make this a viable
short term option. 

> Being able to replace my wrapped up docker calls for live reloading of site and other tooling is critical for my further adoption. Without this most of the local service stuff that docker compose, or my wrapped up magefile style docker commands can't be replaced without losing out on the development experience.

I completely agree, I think this would unlock some really great local
development use cases.

helderco:
I would try mutagen first.

mvgijssel:
I would also love this feature directly integrated into Dagger. For now I’m going to see if I can integrate Dagger with Tilt https://tilt.dev/

levlaz:
> I would also love this feature directly integrated into Dagger. For now I’m going to see if I can integrate Dagger with Tilt https://tilt.dev/

tilt seems interesting, I have not heard of it, please do share how it
goes if you give it a try!

sheldonhull:
I've used tilt but it has its own DSL (Starlark) and while having it invoke Dagger automatically could solve it short term, ideally I'd have a way to do a file watcher locally for live reloading. I think this is one of the core areas for me that I need to solve if I want to replace more of the mage/local automation workflows.

wmcnamee-coreweave:
Just wondering if there's been any progress/movement on this front?

[...]

PR #10723, opened 2025-07-14
chore: lift services into dagger

jedevc:
Instead of going through the buildkit gateway, directly invoke and call the mount manager from inside of dagger. This is a follow-up to #10457, it's essentially the exact same transform, but for services.

This turns out to be a dependency for changes we'd like to do around #6990 - if we have direct access to the mount manager, then we open the way to propagate synced directories in.

### DIFF
--- a/core/terminal.go
+++ b/core/terminal.go
@@ -4,9 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 
-	bkgw "github.com/moby/buildkit/frontend/gateway/client"
 	bkgwpb "github.com/moby/buildkit/frontend/gateway/pb"
 	"github.com/muesli/termenv"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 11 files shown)


shykes:
Anything that helps accelerate #6990 sounds good to me! Beyond that I can't really give a helpful review on the implementation

core/terminal.go
```diff
@@ -62,82 +78,59 @@ func (container *Container) Terminal(
...
-					}
-					return fmt.Errorf("error forwarding container stderr to terminal: %w", err)
-				}
-				return nil
-			})
+		&ServiceIO{
+			Stdin:    term.Stdin,
+			Stdout:   term.Stdout,
+			Stderr:   term.Stderr,
+			ResizeCh: term.ResizeCh,
```
vito, line 133: Very nice - this is the Nth time that I've seen inverting the flow of Reader/Writer has made for cleaner code, even aside from grouping them into one arg

PR #10730, opened 2025-07-15
chore: convert file.WithTimestamps to dagop

alexcb:


core/file.go
```diff
@@ -309,22 +311,46 @@ func (file *File) WithName(ctx context.Context, filename string) (*File, error)
...
+	newRef, err := query.BuildkitCache().New(ctx, fileCacheRef, bkSessionGroup, bkcache.WithRecordType(bkclient.UsageRecordTypeRegular),
+		bkcache.WithDescription(fmt.Sprintf("WithTimestamps %d", unix)))
 	if err != nil {
 		return nil, err
 	}
-	file.LLB = def.ToPB()
-	file.File = path.Base(file.File)
-
+	err = MountRef(ctx, newRef, bkSessionGroup, func(root string) error {
+		fullPath, err := RootPathWithoutFinalSymlink(root, file.File)
```
jedevc, line 335: Can you add a comment to the definition of this function? Have vaguely forgotten why it's needed in this case :sweat_smile:

jedevc (approved):

core/file.go
alexcb: good call, I lifted part of the comment out of `containerdfs.RootPath`. Let me know if it makes sense or needs an example in the comment.

core/util.go
```diff
@@ -396,6 +396,12 @@ var (
 	enumView = AfterVersion("v0.18.11")
 )
 
+// RootPathWithoutFinalSymlink joins a path with a root, evaluating and bounding all
```
alexcb, line 399: @jedevc here's the new comment

Issue #8548, opened 2024-09-23
Change order of include/exclude patterns to allow overrides on module load


helderco:
## Context

When loading a module, the engine compiles a list of include/exclude patterns in order to filter the files that need to be uploaded from the host to the engine.

This is how the engine joins **include** patterns:
- `<module root>/<dagger.json include...>`
- `<module root>/dagger.json`
- `<module root>/**/*`
- `<sdk required paths>`

## Problem

Notice that a module’s `dagger.json` patterns are added before anything else so it **doesn’t have a chance to override** anything that’s more global like the entire root directory, or the SDK’s implicit includes (required paths).

The exclude patterns are simpler and already correctly ordered:
- `.git/**`
- `<module root>/<dagger.json exclude>...`

## Solution

Reorder the patterns in the following way:
- `<sdk required paths>`
- `<module root>/**/*`
- `<module root>/<dagger.json include...>`
- `<module root>/dagger.json`

This allows a `dagger.json` file to **override** the implicit SDK includes, limit which files are included from the root directory without resorting to a more difficult to manage list of excludes, but also not negate the inclusion of `dagger.json` itself.

## Example

Consider a `foo/bar/dagger.json` module with:

```json
{
    "name": "foobar",
    "sdk": "go",
    "source": ".dagger",
    "include": [
        "!../../**/vendor/",
        "!**/*",
        "src/"
    ]
}
```

The following is a diff between how the final **include** patterns are ordered now, and how I suggest we reorder them to:

```diff
-!**/vendor/
-!foo/bar/**/*
-foo/bar/src/
-foo/bar/dagger.json
-foo/bar/**/*
**/go.mod
**/go.sum
**/go.work
**/go.work.sum
**/vendor/
**/*.go
+foo/bar/**/*
+!**/vendor/
+!foo/bar/**/*
+foo/bar/src/
+foo/bar/dagger.json
```

I’m aware that you can exclude something that was included instead of negating the include, I’m just trying to prove a point on the ability to override.

PR #10736, opened 2025-07-16
feat: apply git ignore pattern on contextual directory argument

TomChv:
Update local module loading to lookup for all .gitignore files inside the context directory
Add `IgnorePattenrs` in `LocalModuleSource`.
Apply `IgnorePatterns` when loading contextual directory argument.

### DIFF
--- a/core/modulesource.go
+++ b/core/modulesource.go
@@ -333,6 +333,32 @@ func (src *ModuleSource) LoadContext(
 			return inst, fmt.Errorf("path %q is outside of context directory %q, path should be relative to the context directory", path, ctxPath)
 		}
 
+		// Rebased the ignore patterns to be relative to the mounted path from the context directory.
+		// - If the contextual path is /foo/bar, all /foo/bar/.gitignore patterns must become the root or they'll not be applied correctly
+		// - Recursive patterns are not affected
+		// - Patterns that are outside the contextual path are ignored.
+		// - If relativePathToCtx is "." (meaning the contextual path is the same as the mounted path), then all ignore
+		// patterns are applied.
+		for _, pattern := range src.Local.IgnorePatterns {
+			if !strings.HasPrefix(pattern, "**") &&
+				!strings.HasPrefix(pattern, relativePathToCtx+"/") &&
+				!strings.HasPrefix(pattern, "!"+relativePathToCtx+"/") &&
+				relativePathToCtx != "." {
+				continue
+			}
+
+			// If the pattern is negative, we need to trim the leading "!" before removing the relative path
+			// and add it again after.
+			// !foo/bar/baz -> baz -> !baz
+			if strings.HasPrefix(pattern, "!") {
+				pattern = strings.TrimPrefix(pattern, "!")
+				ignore = append(ignore, "!"+strings.TrimPrefix(pattern, relativePathToCtx+"/"))
+				continue
+			}
+
+			ignore = append(ignore, strings.TrimPrefix(pattern, relativePathToCtx+"/"))
+		}
+
 		err = dag.Select(localSourceCtx, dag.Root(), &inst,
 			dagql.Selector{
 				Field: "host",
@@ -454,6 +480,7 @@ func (src *ModuleSource) LoadContext(
 
 type LocalModuleSource struct {
 	ContextDirectoryPath string
+	IgnorePatterns       []string
 }
 
 func (src LocalModuleSource) Clone() *LocalModuleSource {
--- a/core/schema/modulesource.go
+++ b/core/schema/modulesource.go
@@ -22,6 +22,7 @@ import (
 	"github.com/dagger/dagger/engine/buildkit"
 	"github.com/dagger/dagger/engine/client/pathutil"
 	"github.com/dagger/dagger/engine/server/resource"
+	"github.com/moby/buildkit/util/bklog"
 	"github.com/opencontainers/go-digest"
 	fsutiltypes "github.com/tonistiigi/fsutil/types"
 	"golang.org/x/sync/errgroup"
@@ -410,6 +411,7 @@ func (s *moduleSourceSchema) localModuleSource(
 		Kind:              core.ModuleSourceKindLocal,
 		Local: &core.LocalModuleSource{
 			ContextDirectoryPath: contextDirPath,
+			IgnorePatterns:       nil,
 		},
 	}
 
@@ -444,7 +446,7 @@ func (s *moduleSourceSchema) localModuleSource(
 			return inst, err
 		}
 
-		// load this module source's context directory, sdk and deps in parallel
+		// load this module source's context directory, ignore patterns, sdk and deps in parallel
 		var eg errgroup.Group
 		eg.Go(func() error {
 			if err := s.loadModuleSourceContext(ctx, localSrc); err != nil {
@@ -472,6 +474,10 @@ func (s *moduleSourceSchema) localModuleSource(
 				return nil
 			})
 		}
+
+		eg.Go(func() error {
+			return s.loadLocalModuleIgnorePatterns(ctx, bk, localSrc)
+		})
 		if err := eg.Wait(); err != nil {
 			return inst, err
 		}
@@ -842,6 +848,50 @@ func (s *moduleSourceSchema) initFromModConfig(configBytes []byte, src *core.Mod
 	src.RebasedIncludePaths = append(src.RebasedIncludePaths, rebasedIncludes...)
 
 	return nil
+}
+
+// loadModuleIgnorePatterns loads the .gitignore files inside the context directory
+// and returns the list of patterns that should be ignored by the module
+func (s *moduleSourceSchema) loadLocalModuleIgnorePatterns(
+	ctx context.Context,
+	bk *buildkit.Client,
+	src *core.ModuleSource,
+) error {
+	return (func() (rerr error) {
+		ctx, span := core.Tracer(ctx).Start(ctx, fmt.Sprintf("load .gitignore patterns up to %q", src.Local.ContextDirectoryPath), telemetry.Internal())
+		defer telemetry.End(span, func() error { return rerr })
+
+		lg := bklog.G(ctx).WithField("module", src.ModuleName)
+		ctx = bklog.WithLogger(ctx, lg)
+
+		var contextDirectory dagql.ObjectResult[*core.Directory]
+
+		err := s.dag.Select(ctx, s.dag.Root(), &contextDirectory,
+			dagql.Selector{Field: "host"},
+			dagql.Selector{
+				Field: "directory",
+				Args: []dagql.NamedInput{
+					{Name: "path", Value: dagql.String(src.Local.ContextDirectoryPath)},
+					{Name: "include", Value: dagql.ArrayInput[dagql.String](dagql.NewStringArray("**.gitignore"))},
+				},
+			},
+		)
+
+		if err != nil {
+			return fmt.Errorf("failed to load context directory with only .gitignore: %w", err)
+		}
+
+		gitIgnorePatterns, err := loadGitIgnoreInDirectory(ctx, contextDirectory.Result.Self(), ".")
+		if err != nil {
+			return fmt.Errorf("failed to load .gitignore in context directory %q: %w", src.Local.ContextDirectoryPath, err)
+		}
+
+		lg.Debugf("pattern ignored by .gitignore %q", strings.Join(gitIgnorePatterns, ", "))
+
+		src.Local.IgnorePatterns = gitIgnorePatterns
+
+		return nil
+	})()
 }
 
 // load (or re-load) the context directory for the given module source
--- a/core/schema/util.go
+++ b/core/schema/util.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"path/filepath"
+	"strings"
 
 	"github.com/dagger/dagger/core"
 	"github.com/dagger/dagger/dagql"
@@ -73,3 +75,121 @@ var AllVersion = core.AllVersion
 
 type BeforeVersion = core.BeforeVersion
 type AfterVersion = core.AfterVersion
+
+// Extract the gitignore patterns from its file content and return them as relative paths
+// based on the given parent directory.
+// The parent directory is required to match the gitignore patterns relative to
+// the context directory since the given content may be a .gitignore file from a subdirectory
+// of that context.
+//
+// Example:
+//   - contextDir = "/my-project/foo"
+//   - .gitignoreDir = "/my-project/foo/bar/.gitignore"
+//   - parentDir = "/my-project/foo/bar" -> so **/node_modules becomes my-project/foo/bar/**/node_modules
+//
+// Pattern additional formatting is based on https://git-scm.com/docs/gitignore so it can
+// be correctly applied with further dagger include/exclude filter patterns.
+//
+// Here are the rules of that filter/formatting:
+//
+//   - We ignore empty lines and comments (starting with `#`) (`\#` isn't ignored).
+//
+//   - If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative.
+//     Otherwise, the pattern is recursive.
+//     If pattern is already starting with **, not change needed.
+//     If the pattern starts with *, it's not recursive but only matches the directory itself.
+//
+//   - If a pattern is negative exclusion (starts with `!`) or targets directory only
+//     (ends with `/`), we treat is as a regular path then readd the exclusion to make
+//     sure the recusive pattern is applied if needed.
+func extractGitIgnorePatterns(gitIgnoreContent string, parentDir string) []string {
+	ignorePatterns := []string{}
+
+	// Split gitignore files by line
+	ignorePatternsLines := strings.Split(string(gitIgnoreContent), "\n")
+
+	for _, linePattern := range ignorePatternsLines {
+		// ignore comments, negatives and empty lines
+		if strings.HasPrefix(linePattern, "#") || linePattern == "" {
+			continue
+		}
+
+		// Save if the pattern is a directory only or negate so we can work with the path
+		// only and reconstruct it later
+		isDirOnly := strings.HasSuffix(linePattern, "/")
+		isNegate := strings.HasPrefix(linePattern, "!")
+		pattern := strings.TrimPrefix(strings.TrimSuffix(linePattern, "/"), "!")
+
+		// Based on https://git-scm.com/docs/gitignore
+		// If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative.
+		// Otherwise, the pattern is recursive.
+		// If pattern is already starting with **, not change needed
+		// If the pattern starts with *, it's not recursive but only matches the directory itself.
+		if !strings.Contains(pattern, "/") &&
+			!strings.HasPrefix(pattern, "**") &&
+			!strings.HasPrefix(pattern, "*") {
+			pattern = "**/" + pattern
+		}
+
+		// Rebase the pattern based on the relative path from the context.
+		relativePattern := filepath.Join(parentDir, pattern)
+
+		// Reconstruct the pattern with negative or directory only pattern
+		if isNegate {
+			relativePattern = "!" + relativePattern
+		}
+		if isDirOnly {
+			relativePattern = relativePattern + "/"
+		}
+
+		ignorePatterns = append(ignorePatterns, relativePattern)
+	}
+
+	return ignorePatterns
+}
+
+// Load git ignore patterns in the current directory and all its children
+// It assumes that the given `dir` only contains `.gitignore` files and directories that may
+// contain `.gitignore` files.
+func loadGitIgnoreInDirectory(ctx context.Context, dir *core.Directory, parentDir string) ([]string, error) {
+	result := []string{}
+	name := dir.Dir
+
+	entries, err := dir.Entries(ctx, ".")
+	if err != nil {
+		return nil, fmt.Errorf("failed to list files in directory %s: %w", dir.Dir, err)
+	}
+
+	for _, entry := range entries {
+		if entry == ".gitignore" {
+			file, err := dir.File(ctx, entry)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get file %s: %w", entry, err)
+			}
+
+			content, err := file.Contents(ctx)
+			if err != nil {
+				return nil, fmt.Errorf("failed read file git ignore %s: %w", entry, err)
+			}
+
+			parentPath := filepath.Join(parentDir, name)
+
+			result = append(result, extractGitIgnorePatterns(string(content), parentPath)...)
+			continue
+		}
+
+		subDir, err := dir.Directory(ctx, entry)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get directory %s: %w", entry, err)
+		}
+
+		subDirResult, err := loadGitIgnoreInDirectory(ctx, subDir, filepath.Join(parentDir, name))
+		if err != nil {
+			return nil, fmt.Errorf("failed to get directory git ignore %s: %w", entry, err)
+		}
+
+		result = append(result, subDirResult...)
+	}
+
+	return result, nil
+}
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -5082,6 +5082,234 @@ func (t *Test) IgnoreDirButKeepFileInSubdir(
 	})
 }
 
+func (ModuleSuite) TestDefaultGitIgnore(ctx context.Context, t *testctx.T) {
+	type gitIgnoreFile struct {
+		path    string
+		content string
+	}
+
+	type expectedDir struct {
+		path string
+		want string
+	}
+
+	t.Run("correctly use .gitignore files", func(ctx context.Context, t *testctx.T) {
+		sourceFile := `
+package main
+
+import (
+	"dagger/test/internal/dagger"
+)
+
+type Test struct {
+	Repo      *dagger.Directory
+	ModuleDir *dagger.Directory
+}
+
+func New(
+	//+defaultPath="/"
+	repo *dagger.Directory,
+
+	//+defaultPath="."
+	moduleDir *dagger.Directory,
+) *Test {
+	return &Test{
+		Repo:      repo,
+		ModuleDir: moduleDir,
+	}
+}`
+
+		gitIgnore := []gitIgnoreFile{
+			{
+				path: "/work",
+				content: `**.txt
+bar/*.go
+!c.txt
+*.py
+/.git
+**.gitignore
+**.gitattributes
+`,
+			},
+			{
+				path:    "/work/yamlFiles",
+				content: `foo.yaml`,
+			},
+			{
+				path: "/work/mod",
+				content: `/internal
+/dagger.gen.go
+/dagger.json
+/go.mod
+/go.sum
+**.go
+!main.go
+`,
+			},
+		}
+
+		expected := []expectedDir{
+			{
+				path: ".",
+				want: "a.json\nbar/\nc.txt\nfoo.xml\nmod/\nyamlFiles/\n",
+			},
+			{
+				path: "bar",
+				want: "sub.py\n",
+			},
+			{
+				path: "yamlFiles",
+				want: "bar.yaml\n",
+			},
+		}
+
+		expectedModuleDir := "LICENSE\nmain.go\n"
+		c := connect(ctx, t)
+
+		repo := c.Directory().
+			WithNewFile("a.json", "{}").
+			WithNewFile("b.txt", "b").
+			WithNewFile("c.txt", "c").
+			WithNewFile("foo.xml", "<>").
+			WithNewFile("script.py", "print('hello')").
+			WithDirectory("bar", c.Directory().WithNewFile("baz.txt", "baz").WithNewFile("d.go", "package main").WithNewFile("sub.py", "")).
+			WithDirectory("yamlFiles", c.Directory().WithNewFile("foo.yaml", "foo").WithNewFile("bar.yaml", "bar"))
+
+		modGen := c.Container().From(golangImage).
+			WithExec([]string{"apk", "add", "git"}).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithDirectory("/work", repo).
+			WithWorkdir("/work").
+			// Setup context directory
+			WithExec([]string{"git", "init"}).
+			WithWorkdir("/work/mod").
+			With(daggerExec("init", "--source=.", "--name=test", "--sdk=go")).
+			WithNewFile("main.go", sourceFile)
+
+		for _, ignore := range gitIgnore {
+			modGen = modGen.WithNewFile(filepath.Join(ignore.path, ".gitignore"), ignore.content)
+		}
+
+		for _, expected := range expected {
+			expectedResult := expected.want
+
+			dir, err := modGen.With(daggerCall("repo", "directory", "--path", expected.path, "entries")).Stdout(ctx)
+			require.NoError(t, err)
+			require.Equal(t, expectedResult, dir)
+		}
+
+		moduleDir, err := modGen.With(daggerCall("module-dir", "entries")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, expectedModuleDir, moduleDir)
+	})
+
+	t.Run("correctly rebase path based on defaultPath", func(ctx context.Context, t *testctx.T) {
+		sourceFile := `
+package main
+
+import (
+	"dagger/test/internal/dagger"
+)
+
+type Test struct {
+	Repo      *dagger.Directory
+	ModuleDir *dagger.Directory
+}
+
+func New(
+	//+defaultPath="/bar"
+	repo *dagger.Directory,
+
+	//+defaultPath="./dagger"
+	moduleDir *dagger.Directory,
+) *Test {
+	return &Test{
+		Repo:      repo,
+		ModuleDir: moduleDir,
+	}
+}`
+
+		gitIgnore := []gitIgnoreFile{
+			{
+				path: "/work",
+				content: `**.txt
+bar/*.go
+!c.txt
+*.py
+/.git
+**.gitignore
+**.gitattributes
+`,
+			},
+			{
+				path:    "/work/bar",
+				content: `yamlFiles/foo.yaml`,
+			},
+			{
+				path: "/work/mod",
+				content: `/dagger/internal
+dagger.gen.go
+dagger.json
+/go.mod
+/go.sum
+**.go
+!main.go
+`,
+			},
+		}
+
+		expected := []expectedDir{
+			{
+				path: ".",
+				want: "sub.py\nyamlFiles/\n",
+			},
+			{
+				path: "yamlFiles",
+				want: "bar.yaml\n",
+			},
+		}
+
+		expectedModuleDir := "dagger.gen.go\ngo.mod\ngo.sum\nmain.go\n"
+		c := connect(ctx, t)
+
+		repo := c.Directory().
+			WithNewFile("a.json", "{}").
+			WithNewFile("b.txt", "b").
+			WithNewFile("c.txt", "c").
+			WithNewFile("foo.xml", "<>").
+			WithNewFile("script.py", "print('hello')").
+			WithDirectory("bar", c.Directory().WithNewFile("baz.txt", "baz").WithNewFile("d.go", "package main").WithNewFile("sub.py", "").
+				WithDirectory("yamlFiles", c.Directory().WithNewFile("foo.yaml", "foo").WithNewFile("bar.yaml", "bar")))
+
+		modGen := c.Container().From(golangImage).
+			WithExec([]string{"apk", "add", "git"}).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithDirectory("/work", repo).
+			WithWorkdir("/work").
+			// Setup context directory
+			WithExec([]string{"git", "init"}).
+			WithWorkdir("/work/mod").
+			With(daggerExec("init", "--source=dagger", "--name=test", "--sdk=go")).
+			WithNewFile("dagger/main.go", sourceFile)
+
+		for _, ignore := range gitIgnore {
+			modGen = modGen.WithNewFile(filepath.Join(ignore.path, ".gitignore"), ignore.content)
+		}
+
+		for _, expected := range expected {
+			expectedResult := expected.want
+
+			dir, err := modGen.With(daggerCall("repo", "directory", "--path", expected.path, "entries")).Stdout(ctx)
+			require.NoError(t, err)
+			require.Equal(t, expectedResult, dir)
+		}
+
+		moduleDir, err := modGen.With(daggerCall("module-dir", "entries")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, expectedModuleDir, moduleDir)
+	})
+}
+
 func (ModuleSuite) TestFloat(ctx context.Context, t *testctx.T) {
 	depSrc := `package main
 


core/modulesource.go
```diff
@@ -333,6 +333,32 @@ func (src *ModuleSource) LoadContext(
 			return inst, fmt.Errorf("path %q is outside of context directory %q, path should be relative to the context directory", path, ctxPath)
 		}
 
+		// Rebased the ignore patterns to be relative to the mounted path from the context directory.
+		// - If the contextual path is /foo/bar, all /foo/bar/.gitignore patterns must become the root or they'll not be applied correctly
+		// - Recursive patterns are not affected
+		// - Patterns that are outside the contextual path are ignored.
+		// - If relativePathToCtx is "." (meaning the contextual path is the same as the mounted path), then all ignore
+		// patterns are applied.
+		for _, pattern := range src.Local.IgnorePatterns {
```
jedevc: As we discussed before, *if* we can, we should try and get this logic into `Host.directory` (probably through an opt-in optional arg?), even if it requires a bit more effort.

1. Then it's reusable, external clients can use it.
2. We can use it for all CLI directory args (not just for the default context args). Without that, we'd have to implement it twice, or we have different directories included for whether you pass it in manually, or rely on the default.

core/schema/util.go
```diff
@@ -73,3 +75,121 @@ var AllVersion = core.AllVersion
...
+//
+//   - If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative.
+//     Otherwise, the pattern is recursive.
+//     If pattern is already starting with **, not change needed.
+//     If the pattern starts with *, it's not recursive but only matches the directory itself.
+//
+//   - If a pattern is negative exclusion (starts with `!`) or targets directory only
+//     (ends with `/`), we treat is as a regular path then readd the exclusion to make
+//     sure the recusive pattern is applied if needed.
+func extractGitIgnorePatterns(gitIgnoreContent string, parentDir string) []string {
```
jedevc: If we can get some tests for this, that would be pretty helpful - something that compares the actual behavior of the two.

(we could also put this in `/utils/ignoreconvert` or something like that, it's not schema-specific)

core/schema/util.go
```diff
@@ -73,3 +75,121 @@ var AllVersion = core.AllVersion
...
+//   - We ignore empty lines and comments (starting with `#`) (`\#` isn't ignored).
+//
+//   - If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative.
+//     Otherwise, the pattern is recursive.
+//     If pattern is already starting with **, not change needed.
+//     If the pattern starts with *, it's not recursive but only matches the directory itself.
+//
+//   - If a pattern is negative exclusion (starts with `!`) or targets directory only
+//     (ends with `/`), we treat is as a regular path then readd the exclusion to make
+//     sure the recusive pattern is applied if needed.
```
jedevc: Can you give examples of these?

core/schema/util.go
```diff
@@ -73,3 +75,121 @@ var AllVersion = core.AllVersion
...
+}
+
+// Load git ignore patterns in the current directory and all its children
+// It assumes that the given `dir` only contains `.gitignore` files and directories that may
+// contain `.gitignore` files.
+func loadGitIgnoreInDirectory(ctx context.Context, dir *core.Directory, parentDir string) ([]string, error) {
+	result := []string{}
+	name := dir.Dir
+
+	entries, err := dir.Entries(ctx, ".")
```
jedevc: We should use `dir.Glob` instead, it handles the recursion for us.

core/modulesource.go
TomChv: I agree with you but there's still that problem of context directory that we'll need to work around. I'm quite not sure yet to how solve it.

To load an argument we call `LoadContext` with the default path set in the argument.
Like `+defaultPath="/foo/bar"`, but in the logic your pointing at, I'm looking for .gitignore on all context directory, not just the given path and its children.

```
// - Recursive patterns are not affected <- That means they must be applied
```

In `Host.Directory`, we could have a new argument like `useGitIgnore` that will apply all .gitignore existing in the given directory, that's totally possible.
But for contextual argument, it's more tricky, applying `.gitignore` existing in `/foo/bar` isn't enough, we need to get .gitignore from the context dir and anything under it.

So I think the logic will be duplicated in order to make that possible, what do you think?

up! I have an idea! I could simply just keep the recursive pattern and implement the git ignore research in Host.Directory too, we'll have a little bit of duplication but that's the best of both world imo

jedevc: So, is the problem that if we load `foo/bar`, that we won't load `foo/other/.gitignore`? If my understanding is right, we can handle that - we can have the gitignore loading logic in `Host.directory` right? We're not restricted there, we can read any files on the host if we want.

If that's not the problem, can you share an example?

TomChv: We can have the git ignore logic in host.directory yeah, but that means we mean to dedup the findup logic to find the parent directory (either a .git repo or the argument dir is considered the root if there's no .git?)

TomChv: My only problem is where do we stop looking for a parent `.gitignore`, for context dir it's easy, we already have the ContextDirectoryPath when looking for these gitignore, but if we include that in host.Directory, we don't have it so we need to figure it out inside the func, which lead to either deduplication or no-complete ignore (arg dir is considered the root dir).

jedevc: Ah, I see, yeah, we'll have a copy of the findup logic. However, it shouldn't be a copy, we should extract that logic to a specific function and reuse it there.

TomChv: Make sense! alright! I'll push my changes so far, and continue on this

core/schema/util.go
TomChv: Done, I added some tests in [365929e](https://github.com/dagger/dagger/pull/10736/commits/365929ec62617b929b10a62820e26ec77e990f05)

core/schema/util.go
TomChv: Done in [365929e](https://github.com/dagger/dagger/pull/10736/commits/365929ec62617b929b10a62820e26ec77e990f05)

core/schema/util.go
TomChv: Done in [fc5426a](https://github.com/dagger/dagger/pull/10736/commits/fc5426a5563598ca3bbea4945da0bb3b55a2c97f)

core/modulesource.go
TomChv: Note from a 1:1 with Justin

1. We look for .git
2. If not .git -> skip
3. If there's -> grab all .gitignore of that repo

Add a hidden argument so we don't duplicate findup if it's called from `LoadContext`, if contextDirectoryPath is set, skip find up

TomChv:
EDIT: moving the logic inside host.Directory fixed that problem since it's now only part of the logs instead of arguments

@vito 

The Cloud integration is kind of bad with this new .gitignore feature.
We should have something to collapse arguments

Screenshot 2025-07-22 at 15 48 53

https://dagger.cloud/dagger/traces/bb359ddf51bf6b1472f7417a29ce6f57

core/modulesource.go
TomChv: Done in [ec87d7f](https://github.com/dagger/dagger/pull/10736/commits/ec87d7fd1e4d75f359d1446720ccf23536ab71e8)

core/util.go
```diff
@@ -513,3 +513,135 @@ func getRefOrEvaluate[T fileOrDirectory](ctx context.Context, t T) (bkcache.Immu
...
+//   - We ignore empty lines and comments (starting with `#`) (`\#` isn't ignored).
+//
+//   - If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative.
+//     Otherwise, the pattern is recursive.
+//     If pattern is already starting with **, not change needed.
+//     If the pattern starts with *, it's not recursive but only matches the directory itself.
+//     Example: foo/bar stays foo/bar but foo becomes **/foo
+//
+//   - If a pattern is negative exclusion (starts with `!`) or targets directory only
+//     (ends with `/`), we treat is as a regular path then readd the exclusion to make
```
helderco:
```suggestion
//     (ends with `/`), we treat is as a regular path then read the exclusion to make
```

core/util.go
```diff
@@ -513,3 +513,135 @@ func getRefOrEvaluate[T fileOrDirectory](ctx context.Context, t T) (bkcache.Immu
...
+			dagql.Selector{
+				Field: "file",
+				Args: []dagql.NamedInput{
+					{Name: "path", Value: entryValue},
+				},
+			},
+			dagql.Selector{
+				Field: "contents",
+			},
+		)
```
helderco: This could potentially be optimized by reading the files concurrently rather than sequentially, but should measure with a heavy directory with lots of .gitignore to see if it's worth it. Not a blocker though, but why not anyway?

core/util.go
```diff
@@ -513,3 +513,135 @@ func getRefOrEvaluate[T fileOrDirectory](ctx context.Context, t T) (bkcache.Immu
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get dag server: %w", err)
+	}
+
+	var entries dagql.Array[dagql.String]
+	err = dag.Select(ctx, dir, &entries,
+		dagql.Selector{
+			Field: "glob",
+			Args: []dagql.NamedInput{
+				{Name: "pattern", Value: dagql.String("**.gitignore")},
```
helderco: This may match files like `_.gitignore`[^1] or `example.gitignore`. The pattern `**/.gitignore` should match both `.gitignore` and `subdir/.gitignore`.

[^1]: May seem strange but people may rename the file temporarily in order to disable it when debugging, for example.

core/schema/host.go
```diff
@@ -283,10 +287,63 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
...
+			if err != nil {
+				return i, fmt.Errorf("failed to find up .git: %w", err)
+			}
+
+			if found {
+				parentPath = dotGitPath
+			}
+		}
+
+		dotGitIgnoreParentPath = parentPath
```
helderco: `parentPath` seems unnecessary, could assign directly to `dotGitIgnoreParentPath`.

core/schema/host.go
```diff
@@ -316,6 +373,80 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
 	return core.MakeDirectoryContentHashed(ctx, bk, dir)
 }
 
+func loadDirectoryGitIgnorePatterns(ctx context.Context, parentPath string) (patterns []string, rerr error) {
+	fmt.Println("loadDirectoryGitIgnorePatterns", parentPath)
```
helderco: Leftover from debugging? There's more like this.

core/schema/host.go
```diff
@@ -316,6 +373,80 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
...
+	}
+
+	var directory dagql.ObjectResult[*core.Directory]
+	err = srv.Select(ctx, srv.Root(), &directory,
+		dagql.Selector{Field: "host"},
+		dagql.Selector{
+			Field: "directory",
+			Args: []dagql.NamedInput{
+				{Name: "path", Value: dagql.String(parentPath)},
+				{Name: "include", Value: dagql.ArrayInput[dagql.String](dagql.NewStringArray("**.gitignore"))},
```
helderco:
```suggestion
				{Name: "include", Value: dagql.ArrayInput[dagql.String](dagql.NewStringArray("**/.gitignore"))},
```

core/schema/host.go
```diff
@@ -283,10 +287,63 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
...
+		if err != nil {
+			return i, fmt.Errorf("failed to load .gitignore patterns: %w", err)
+		}
+
+		rebasedIgnorePatterns, err := rebaseGitIgnorePatterns(dotGitIgnoreParentPath, args.Path, ignorePatterns)
+		if err != nil {
+			return i, fmt.Errorf("failed to rebase .gitignore patterns: %w", err)
+		}
+
+		excludesPatterns = append(excludesPatterns, rebasedIgnorePatterns...)
```
helderco, line 332: I think .gitignore patterns should come before user defined ones so they can be overridden. See:
- https://github.com/dagger/dagger/issues/8548

core/schema/host.go
```diff
@@ -316,6 +373,80 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
...
+	relativePathToParent, err := filepath.Rel(parentPath, hostPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get relative path from %q: %w", parentPath, err)
+	}
+
+	result := []string{}
+	// Rebased the ignore patterns to be relative to the mounted path from the context directory.
+	// - If the contextual path is /foo/bar, all /foo/bar/.gitignore patterns must become the root or they'll not be applied correctly
+	// - Recursive patterns are not affected
+	// - Patterns that are outside the contextual path are ignored.
```
helderco, line 439: Can you come up with an example where this could happen (as a comment)? It's good to have that guard just want to understand that use case.

core/schema/host.go
```diff
@@ -150,6 +150,7 @@ func (s *hostSchema) Install(srv *dagql.Server) {
 				dagql.Arg("exclude").Doc(`Exclude artifacts that match the given pattern (e.g., ["node_modules/", ".git*"]).`),
 				dagql.Arg("include").Doc(`Include only artifacts that match the given pattern (e.g., ["app/", "package.*"]).`),
 				dagql.Arg("noCache").Doc(`If true, the directory will always be reloaded from the host.`),
+				dagql.Arg("applyGitIgnore").Doc(`If true, gitignore patterns will be applied to the directory.`),
```
helderco: On the fence about the name but no strong opinion. Just have a preference on following `rg`, which is simpler, familiar, and allows supporting other VCSs later:
```
       --no-ignore-vcs
           When given, filter rules from source control ignore files (e.g., .gitignore) are not respected. By default, ripgrep respects git's ignore rules for
           automatic filtering. In some cases, it may not be desirable to respect the source control's ignore rules and instead only respect rules in  .ignore
           or .rgignore.

           This flag implies --no-ignore-parent for source control ignore files as well.

           This flag can be disabled with --ignore-vcs.
```

So it would be the opposite of that:
```suggestion
				dagql.Arg("ignoreVcs").Doc(`Respect filter rules from source control ignore files when inside a repo (only .gitignore is supported).`),
```

This would make this diff in shell:
```diff
- host | directory . --apply-git-ignore
+ host | directory . --ignore-vcs
```

core/schema/host.go
```diff
@@ -283,10 +287,63 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
...
+			if found {
+				parentPath = dotGitPath
+			}
+		}
+
+		dotGitIgnoreParentPath = parentPath
 	}
+
+	if dotGitIgnoreParentPath != "" {
+		ignorePatterns, err := loadDirectoryGitIgnorePatterns(ctx, dotGitIgnoreParentPath)
```
helderco: This can be optimized a bit more. If you're uploading a subdir of the context directory you'll be fetching all the `.gitignore` files in the repo, even sibling directories. What I suggest doing instead is to also send `args.Path`, split the relative path into parts, and just add include rules for each part, like I did here: 

https://github.com/dagger/dagger/pull/8501/files#diff-35a72c75f37c6dcddd3e1df88cccf5a8b22ac1f6304301c66c3a1c7394ff0eafR1290-R1294

For example:
- Target path: `/mnt/dagger/my/subdir`
- Context dir: `/mnt/dagger`

Use following includes (in this order), targeting the context dir:
- `.gitignore`
- `my/.gitignore`
- `my/subdir/**/.gitignore` (this will include `my/subdir/.gitignore`)

core/util.go
```diff
@@ -513,3 +513,135 @@ func getRefOrEvaluate[T fileOrDirectory](ctx context.Context, t T) (bkcache.Immu
...
+//     Otherwise, the pattern is recursive.
+//     If pattern is already starting with **, not change needed.
+//     If the pattern starts with *, it's not recursive but only matches the directory itself.
+//     Example: foo/bar stays foo/bar but foo becomes **/foo
+//
+//   - If a pattern is negative exclusion (starts with `!`) or targets directory only
+//     (ends with `/`), we treat is as a regular path then readd the exclusion to make
+//     sure the recusive pattern is applied if needed.
+//     Example: !foo becomes foo then **/foo then !**/foo
+func extractGitIgnorePatterns(gitIgnoreContent string, parentDir string) []string {
```
helderco: Just a thought, but there's another potential optimization that could be made to avoid loading patterns from `.gitignore` files inside directories that match already discovered ignore patterns. For example, in dagger's root `.gitignore` there's `.venv` and `.idea`, but these files will get included anyway:
```
▶ host | directory ~/Projects/dagger --include 'sdk/python/**/.gitignore' | glob '**/.gitignore'
sdk/python/.gitignore
sdk/python/.idea/.gitignore
sdk/python/.pytest_cache/.gitignore
sdk/python/.ruff_cache/.gitignore
sdk/python/.venv/.gitignore
sdk/python/dev/.gitignore
sdk/python/dev/.idea/.gitignore
sdk/python/dev/sdk/runtime/.gitignore
sdk/python/runtime/.gitignore
```

However, I believe the impact from that is trivial so I'm not suggesting to actually do it, unless there's a measurable performance benefit. And would rather leave that logic to buildkit/filesync anyway since it's still going to be just two filesync calls (barring the cost of opening each `.gitignore` file but in my testing I found it to be trivial, although didn't measure on a slow computer).

helderco (commented):
Thanks for moving this forward! 🙏

core/schema/host.go
TomChv: Done in [36bfc50](https://github.com/dagger/dagger/pull/10736/commits/36bfc504d118ee019f9e935bb11ccd04d4fd0036)

core/schema/host.go
TomChv: Done in [36bfc50](https://github.com/dagger/dagger/pull/10736/commits/36bfc504d118ee019f9e935bb11ccd04d4fd0036)

core/schema/host.go
TomChv: Done in [36bfc50](https://github.com/dagger/dagger/pull/10736/commits/36bfc504d118ee019f9e935bb11ccd04d4fd0036)

core/schema/host.go
TomChv: Done in [36bfc50](https://github.com/dagger/dagger/pull/10736/commits/36bfc504d118ee019f9e935bb11ccd04d4fd0036)

core/schema/host.go
TomChv: Done in [36bfc50](https://github.com/dagger/dagger/pull/10736/commits/36bfc504d118ee019f9e935bb11ccd04d4fd0036)

core/schema/host.go
TomChv: Done in [36bfc50](https://github.com/dagger/dagger/pull/10736/commits/36bfc504d118ee019f9e935bb11ccd04d4fd0036)

core/util.go
TomChv: Done in [36bfc50](https://github.com/dagger/dagger/pull/10736/commits/36bfc504d118ee019f9e935bb11ccd04d4fd0036)

core/util.go
TomChv: Done in [36bfc50](https://github.com/dagger/dagger/pull/10736/commits/36bfc504d118ee019f9e935bb11ccd04d4fd0036)

core/util.go
TomChv: I'll keep that for a follow up, I plan to also support .dockerignore so it will be the occasion, I would love to merge that PR asap since it's much needed for performances optimization

core/util.go
TomChv: I'll do that in a follow up since it's related to optimization

core/schema/host.go
TomChv: Done in [e9a293e](https://github.com/dagger/dagger/pull/10736/commits/e9a293e2daf0c446bc63b1393da5c62e042095f3)

TomChv: Fixed on [43f00ba](https://github.com/dagger/dagger/pull/10736/commits/43f00bae925ba10619646f8848eeb5f5513a6f3c), I forgot to rebase the gitignore include path on the parentPath, it's done now :)

core/schema/host.go
shykes: What's `rg`?

shykes: Ah I see, ripgrep

shykes:
@TomChv @helderco shouldn't this be enabled by default? Are you worried about breaking changes? Pesonally I'm more worried about nobody ever taking advantage of this obscure flag

core/schema/host.go
shykes: Here's what I propose:

```suggestion
				dagql.Arg("no-git-auto-ignore").Doc(`Don't apply .gitignore`),
				dagql.Arg("no-docker-auto-ignore").Doc(`Don't apply .dockerignore`),
				dagql.Arg("no-hg-auto-ignore").Doc(`Don't apply .hgignore`),
				dagql.Arg("no-auto-ignore").Doc(`Disable the auto-ignore feature, which looks for and applies standard ignore files such as .gitignore, .dockerignore, .hgignore, etc.`),
```

Obviously `no-docker-auto-ignore` and `no-hg-auto-ignore` are not yet supported, but this shows how they (or any other external ignore file) will be incorporated in the future. Note that not all of them are VCS-specific.

TomChv:
> @TomChv @helderco shouldn't this be enabled by default? Are you worried about breaking changes? Pesonally I'm more worried about nobody ever taking advantage of this obscure flag

I talked about it we Justin and we wanted to take it slow in case some people were sending files that are ignored by their .gitignore.
But I enabled ignoreVCS on directory sent from the CLI, we may switch it back to false if it creates any problem for our users

core/schema/host.go
helderco: Yeah, I'm good with that. Also implies flipping it to ignore by default which I think is best.

helderco:
> @TomChv @helderco shouldn't this be enabled by default? Are you worried about breaking changes? Pesonally I'm more worried about nobody ever taking advantage of this obscure flag

Yeah, I agree.

> we may switch it back to false if it creates any problem for our users

@TomChv, changing a default is not an easy task. It's possible we break a few users that rely on getting ignored files somehow, but that's speculative and the benefit for most users outweighs that in my mind.

TomChv:
> Yeah, I agree.

Alright, I changed it to true by default

EDIT: Actually it seems that it's breaking the SDK loading because it detects a recursive call.

I'll switch it back to false for now

> failed to get configured module: failed to load local module source context: select: failed to load .gitignore patterns: failed to load directory ignore patterns: select: failed to load .gitignore patterns: failed to load directory ignore patterns: select: recursive call detected
https://dagger.cloud/dagger/traces/b69b5ea8798b89848c919d069bee82ed

core/modulesource.go
helderco: If this feature needs to be explicitly enabled I can see possible confusion if user puts the flag but then nothing happens. If not an error, could be just a warning. If feature is enabled by default than I think it's ok.

core/schema/host.go
helderco: My comment was specifically on line 422:
> Patterns that are outside the contextual path are ignored.

And I meant for you to just comment here, not to code.

core/schema/util.go
```diff
@@ -73,3 +74,47 @@ var AllVersion = core.AllVersion
...
+//   - hostPath = "/foo/bar"
+//   - .gitignore files will be loaded from the following paths: [.gitignore, foo/.gitignore, foo/bar/**.gitignore]
+//
+// We assume the hostPath is always a child of the parentPath.
+func getGitIgnoreIncludePaths(parentPath string, hostPath string) ([]string, error) {
+	var paths []string
+
+	current := hostPath
+	for {
+		if current == parentPath {
```
helderco: This looks brittle if they're formatted slightly differently like if `parentPath` has a slash in the end this wouldn't be hit. I suggest working from the relative path between them instead.

.changes/unreleased/Added-20250723-184041.yaml
```diff
@@ -0,0 +1,9 @@
+kind: Added
+body: |-
+  Automatically apply .gitignore patterns when loading contextual directory argument.
```
helderco: The main motivation for this feature was actually for module loading (to avoid the need to add explicit patterns to `dagger.json`) but this title makes it easy to miss. I suggest generalizing and maybe add a list of where this is applied in the longer description.

helderco:
> Alright, I changed it to true by default

Remember that you can't just make the default value for the option `true` in the API, it needs to be `false` because of the Go SDK, so you need to rename the option as well in order for that default to be `false`:
```go
dagql.Arg("noGitAutoIgnore")
```

core/schema/util.go
TomChv: I can run `filepath.Clean` on the comparison so we don't have that problem no?

.changes/unreleased/Added-20250723-184041.yaml
TomChv: The module loading doesn't add the context directory anymore since https://github.com/dagger/dagger/pull/9505 so I'm not sure it make sense anymore there?

helderco: Oh, I missed that. I was under the impression that this would solve https://github.com/dagger/dagger/issues/6627. You just need to add the gitignore patterns before the include patterns from `dagger.json`. The default include is everything inside the source directory, even `node_modules`, etc.... So it definitely makes sense to do this.

core/schema/util.go
helderco: In my mind there's more room for something unexpected like a difference in casing or slashes, depending on OS, especially something I'm not aware. The alternative that I suggested is easy and clean.

.changes/unreleased/Added-20250723-184041.yaml
TomChv: I changed the description, let me know if it's what you had in mind :)

core/schema/util.go
TomChv: Done in [5955331](https://github.com/dagger/dagger/pull/10736/commits/5955331f945363a279125a5531bd42a63900bf10)

core/schema/host.go
TomChv: Done in [5955331](https://github.com/dagger/dagger/pull/10736/commits/5955331f945363a279125a5531bd42a63900bf10)

core/schema/util.go
```diff
@@ -73,3 +76,47 @@ var AllVersion = core.AllVersion
...
+	if hostRelPathFromParent == "." {
+		return []string{"**/.gitignore"}, nil
+	}
+
+	pathParts := strings.Split(hostRelPathFromParent, string(os.PathSeparator))
+
+	var paths []string
+	for i := range pathParts {
+		relativePath := filepath.Join(pathParts[:i]...)
+		paths = append([]string{filepath.Join(relativePath, ".gitignore")}, paths...)
```
helderco: This needs to be an append rather than prepend now (from general to more specific).

TomChv: Done in [cb540db](https://github.com/dagger/dagger/pull/10736/commits/cb540db7ceb347f7c16e3d7150f2be7584ba1168)

TomChv:
I did some tests and benchmarks on that feature, the big difference happens on repository that has a lot of generated artifacts, builds and files.

For example, here's a benchmark on a simple Next application:

|                | 0.18.4 Cold cache | 10736 Cold cache  | 0.18.4 Cached | 10737 Cached  |
|----------------|-------------------|-------------------|---------------|---------------|
| Module loading | [0.6s](https://dagger.cloud/Quartz/traces/0ce123cec282d67b6e89b9f8b93ef171?listen=1e83080b2c8669ba&listen=0bb02f9499435ebf&listen=b4a72a2b5643f3b3&listen=326518a7e4e30895#1e83080b2c8669ba)              | [0.1](https://dagger.cloud/Quartz/traces/7ede0de230ad9a03f541297b63dcc383?listen=c8823b9ddca0cac1&listen=21ffe749d7099a20&listen=b59d5396e7b3068d&listen=0c216fb67096ccfd&listen=0a28446efda5873d&listen=8a4db347325554a2&listen=b8921650eb674953&listen=7dc51188eef4d7d4)               | [0.1s](https://dagger.cloud/Quartz/traces/44048584222047e91e0ad0ad7b915d0b?listen=5416571d2e8470c2&listen=9cfac27711e409a4&listen=686e3e71cc8219df&listen=ec73b67ea79429d7&listen=2e78da14941433e4)          | [0.1s](https://dagger.cloud/Quartz/traces/814cab1e6ffe3fa4930af0c703b09f20?listen=631d3a820e779d1c&listen=689586aeb9b15bcb&listen=0303707ae76ff7aa&listen=539c7026aa9fe07d&listen=39b7df9b27669aeb)          |
| Host directory | [28.1s](https://dagger.cloud/Quartz/traces/0ce123cec282d67b6e89b9f8b93ef171?listen=1e83080b2c8669ba&listen=0bb02f9499435ebf&listen=b4a72a2b5643f3b3&listen=326518a7e4e30895#1e83080b2c8669ba)             | [0.1s](https://dagger.cloud/Quartz/traces/7ede0de230ad9a03f541297b63dcc383?listen=b8921650eb674953&listen=7dc51188eef4d7d4&listen=c8823b9ddca0cac1&listen=21ffe749d7099a20&listen=b59d5396e7b3068d&listen=0c216fb67096ccfd&listen=0a28446efda5873d)              | [0.5s](https://dagger.cloud/Quartz/traces/44048584222047e91e0ad0ad7b915d0b?listen=ec73b67ea79429d7)          | [0.0s](https://dagger.cloud/Quartz/traces/814cab1e6ffe3fa4930af0c703b09f20)          |

We can observe that module loading is also improved a bit thanks to these changes

I also did some tests on our dagger repo, on [greetings-api](https://github.com/kpenfound/greetings-api/tree/main) and [openmeter](https://github.com/openmeterio/openmeter) but since I just cloned these repository, there's not much difference in speed because everything has already been ignored during the clone.

helderco:
> [...] the big difference happens on repository that has a lot of generated artifacts, builds and files.

I think that's pretty obvious or easy to understand. I'm more interested in the **overhead** of the implementation.

You should also measure in ms at least, and isolate the `.gitignore` specific part. Baseline is it takes 0 ns without it, of course.

> I also did some tests on our dagger repo, on [greetings-api](https://github.com/kpenfound/greetings-api/tree/main) and [openmeter](https://github.com/openmeterio/openmeter) but since I just cloned these repository, there's not much difference in speed because everything has already been ignored during the clone.

Again, this is the metric that I'm more interested in, i.e., a clean clone of a repo with lots of `.gitignore` files. This tells us the overhead of reading all those files. Basically looking for what's the impact in cases where it's not needed. But the openmeter repo only has 7 `.gitignore` files while our own dagger/dagger repo has 243!
```
❯ dagger -M -c 'git https://github.com/dagger/dagger | head | tree | glob "**/.gitignore"'
```
So measure against our repo, or another one that has even more `.gitignore` files.

For example, if processing each `.gitignore` file takes 5ms, then reading 200 of those in sequence can take a full 1s. That's not negligible, although we can decide that it's still worth it (even on clean clones since it's not easy to figure out in advance if you'll benefit or not). If you then optimize into reading in parallel instead of in sequence and it maybe drops to 100ms instead of 1s, then that's much easier to accept.

Another interesting metric is comparing the effect on cache before and after a `npm install` (for example) on a clean clone. I suspect that without this it invalidates cache, while with it it doesn't.

helderco:
FTR, this change LGTM, I just want to make sure the impact of this change is negligible before approving.

core/schema/host.go
```diff
@@ -316,6 +367,90 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
...
+	rebasedIgnorePatterns, err := rebaseGitIgnorePatterns(parentPath, hostPath, gitIgnorePatterns)
+	if err != nil {
+		return nil, fmt.Errorf("failed to rebase .gitignore patterns: %w", err)
+	}
+
+	return rebasedIgnorePatterns, nil
+}
+
+func rebaseGitIgnorePatterns(parentPath string, hostPath string, patterns []string) ([]string, error) {
+	relativePathToParent, err := filepath.Rel(parentPath, hostPath)
```
helderco, line 430: Need to ensure both paths are absolute before getting the relative path:
```
▼ host | directory . | glob '**' 1.9s ERROR
├─✘ glob 1.9s ERROR
│ ! failed to load .gitignore patterns: failed to rebase .gitignore patterns: failed to get relative path from "/Users/helder/Projects/dagger": Rel: can't make . relative to /Users/helder/Projects/dagger
```

TomChv: Fixed by [c8718d0](https://github.com/dagger/dagger/pull/10736/commits/c8718d071a0012d64f6279b90d7afd743ab20bac)

TomChv:
Screenshot 2025-08-01 at 16 34 29

Still working on the benchmarks, I'm measuring the module loading time for each of these case and we can see that it's slower when there's a certain number of `.gitignore` files, I'll implement the parallel reading to see if we can improve that.
It's hard to exactly tell how long reading a file is since metrics are displayed in second on cloud, not ms. I don't know if there's a way to show that without updating the engine code directly?

TomChv:
Screenshot 2025-08-01 at 18 57 06

I worked on some concurrency implementation, without a semaphore this completly overload the engine but with a semaphore set to `runtime.NumCPU()`, it works and seem the reduce the time a bit.

I added a test with 500 and 1000 .gitignore files, we can see a clear time increase and how the concurrency help but it's still always slower than not loading .gitignore during modules.
The only case where it really improve performances is when directories like `node_modules` or `.venv` are ignored.

I'm not sure what's the right path to take here, loading all .gitignore in files will decrease performances on larger repo during module loading but in exchange that will increase host directory copy speed.

This PR also has hard time loading the dagger repository, I know that it also struggles locally with my local engine so results may be different on cloud or another one (but I cannot test my PR with a cloud engine).
Result for concurrency loading on dagger repo leads to timeout in Cloud (even if I get the result locally: https://github.com/dagger/dagger/issues/10809)

The trouble may come from the number of local dependencies, which lead to a large number of coroutine / .gitignore files read and also maybe some duplication between the root module and its dep.
Maybe updating the module git loading to only source dir + its parent would improve the result. 

EDIT: [d3856ce](https://github.com/dagger/dagger/pull/10736/commits/d3856cefb3e791bf462420e9f0be4c3e98358ebc) solves the dagger repo issue, and significantly improve the speed while keeping the right ignore behaviour for modules.

Screenshot 2025-08-01 at 19 50 24

My module to generate .gitignore files:

```go
package main

import (
	"context"
	"dagger/test/internal/dagger"
	"fmt"
)

type Test struct {
	Repo *dagger.Directory
}

func New(
	//+defaultPath="/"
	repo *dagger.Directory,
) *Test {
	return &Test{
		Repo: repo,
	}
}

func (m *Test) Populate(ctx context.Context, n int) *dagger.Directory {
	mainDir := dag.Directory()

	for i := 0; i < n; i++ {
		mainDir = mainDir.WithDirectory(fmt.Sprintf("dir%d", i), dag.Directory().WithNewFile(".gitignore", "/"))
	}

	return mainDir
}
```

TomChv:
Following https://github.com/dagger/dagger/pull/10736#issuecomment-3145206630, I made a quick graph to compare the performance differences between all my changes

Screenshot 2025-08-01 at 20 41 35

Wee see that the latest, with better precision on .gitignore filter during module source loading is the one that compares to `0.18.14` on a clean repository.
So it can only get better as more files and artifacts will be ignored later.

TomChv:
Tests failing are CI are working fine locally: https://dagger.cloud/Quartz/traces/a5a24659712f77fd0237b25f9566416a?listen=06229ee34984e9d6&listen=25a9c1ddda80397b&listen=64841aa90e2d8fd1&listen=bb8c4ed72c442323

I'll rerun these tests after the CI engine is shutdown, it may be a cache issue

EDIT: CI is green 🚀

.changes/unreleased/Added-20250723-184041.yaml
```diff
@@ -0,0 +1,13 @@
+kind: Added
+body: |-
+  Automatically apply .gitignore patterns when directory on module call
```
helderco: This sentence looks incomplete

.changes/unreleased/Added-20250723-184041.yaml
```diff
@@ -0,0 +1,13 @@
+kind: Added
+body: |-
+  Automatically apply .gitignore patterns when directory on module call 
+  
+  - Apply .gitignore patterns when loading contextual directory argument.
+  - Apply .gitignore pattenrs when a directory is sent as an argument to a module function.
+  - Apply .gitignore patterns when a local module is loaded.
+
+  Add `NoGitAutoIgnore` argument to `Host.Directory` disable this behavior.
```
helderco:
```suggestion
  Add `NoGitAutoIgnore` argument to `Host.Directory` to disable this behavior.
```

core/util.go
```diff
@@ -513,3 +517,161 @@ func getRefOrEvaluate[T fileOrDirectory](ctx context.Context, t T) (bkcache.Immu
...
+					Field: "contents",
+				},
+			)
+			if err != nil {
+				return fmt.Errorf("failed to get file %s: %w", entryValue, err)
+			}
+
+			resultMu.Lock()
+			defer resultMu.Unlock()
+			result = append(result, extractGitIgnorePatterns(string(content), filepath.Dir(string(entryValue)))...)
```
helderco: Should do as minimally as possible in a lock.
```suggestion
			patterns := extractGitIgnorePatterns(string(content), filepath.Dir(string(entryValue)))
			resultMu.Lock()
			defer resultMu.Unlock()
			result = append(result, patterns...)
```

core/schema/modulesource.go
```diff
@@ -946,17 +947,44 @@ func (s *moduleSourceSchema) loadModuleSourceContext(
...
-		err := dag.Select(ctx, dag.Root(), &src.ContextDirectory,
+		moduleSourcePath := src.SourceSubpath
+		if moduleSourcePath == "" {
+			moduleSourcePath = src.SourceRootSubpath
+		}
+
+		// Load .gitignore patterns before loading the module so they can be used in the `Include` argument
+		// instead of Exclude so we get the right behaviour.
+		// We only include the .gitignore from the module source path and above up to the context directory.
+		ignorePatterns, err := loadDirectoryGitIgnorePatterns(ctx, src.Local.ContextDirectoryPath, filepath.Join(src.Local.ContextDirectoryPath, moduleSourcePath))
```
helderco, line 960: Yeah, I can see how this change improves things. @jedevc actually had a good idea to further optimize, which can be done in a follow-up:
- Select all the contents with `host.directory(include=["**/.gitignore"])` and get a `dagql.ObjectResult[*Directory]`
- Use a `Directory.walk(callback)` (doesn't exist yet) to avoid mount/unmounting the directory many times
- Create an internal `Directory._gitignoresFor("path/to/dir")` wrapped in a `DagOpWrapper`

That way if the passed directory is the same the output is cached and nothing will be mounted.

helderco (commented):
LGTM, just one minor thing left. Could use a review from @sipsma or @jedevc to see if all good on the engine/cache side.

.changes/unreleased/Added-20250723-184041.yaml
```diff
@@ -0,0 +1,13 @@
+kind: Added
+body: |-
+  Automatically apply .gitignore patterns on directory loading fo module call
```
helderco: Typo
```suggestion
  Automatically apply .gitignore patterns on directory loading for module call 
```

helderco (approved):
There's several ways to further optimize this, please follow-up when you can.

For now it seems to me the overhead is not a deal breaker, but would still like another look by @sipsma or @jedevc on the engine/cache side in case there's any red flags.

.changes/unreleased/Added-20250723-184041.yaml
TomChv: Fixed by [f150df4](https://github.com/dagger/dagger/pull/10736/commits/f150df4ca097760047ad0d1246621487521a63cb)

core/util.go
TomChv: Done!

.changes/unreleased/Added-20250723-184041.yaml
```diff
@@ -0,0 +1,13 @@
+kind: Added
+body: |-
+  Automatically apply .gitignore patterns on directory loading for module call 
+  
+  - Apply .gitignore patterns when loading contextual directory argument.
+  - Apply .gitignore pattenrs when a directory is sent as an argument to a module function.
+  - Apply .gitignore patterns when a local module is loaded.
+
+  Add `NoGitAutoIgnore` argument to `Host.Directory` to disable this behavior.
```
jedevc, line 9: This is a breaking change - calling `Host.directory` will now give different results. 

On the fence as to whether we want to do this or not. I think we definitely can break for context dirs and module loading, but we might want to preserve the behavior with `Host.directory`.

We'd do this by having `GitAutoIgnore` instead, and just setting it when loading.

core/schema/host.go
```diff
@@ -241,6 +242,9 @@ type hostDirectoryArgs struct {
 
 	core.CopyFilter
 	HostDirCacheConfig
+
+	ContextDirectoryPath string `internal:"true" default:""`
```
jedevc: nit: call this `GitIgnoreRoot` or something. It's not *directly* tied to the idea of context directories.

core/schema/host.go
```diff
@@ -316,6 +372,102 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
...
+			Args: []dagql.NamedInput{
+				{Name: "path", Value: dagql.String(parentPath)},
+				{Name: "include", Value: dagql.ArrayInput[dagql.String](
+					dagql.NewStringArray(gitIgnoreIncludePath...),
+				)},
+				{Name: "noGitAutoIgnore", Value: dagql.Boolean(true)},
+			},
+		},
+	)
+
```
jedevc: nit: blank line

core/schema/modulesource.go
jedevc: The bit we'd wrap in `Directory.__gitignoresFor` is just `core.LoadGitIgnoreInDirectory` :tada:

The logic already looks good, it's just a matter of making it an internal schema method.

core/schema/modulesource.go
```diff
@@ -946,17 +947,44 @@ func (s *moduleSourceSchema) loadModuleSourceContext(
...
+			return fmt.Errorf("failed to load .gitignore patterns: %w", err)
+		}
+
+		rebasedIgnorePatterns, err := rebaseGitIgnorePatterns(src.Local.ContextDirectoryPath, src.Local.ContextDirectoryPath, ignorePatterns)
+		if err != nil {
+			return fmt.Errorf("failed to rebase .gitignore patterns: %w", err)
+		}
+
+		for _, pattern := range rebasedIgnorePatterns {
+			if strings.HasPrefix(pattern, "!") {
```
jedevc: `strings.CutPrefix`

core/schema/modulesource.go
```diff
@@ -946,17 +947,44 @@ func (s *moduleSourceSchema) loadModuleSourceContext(
...
+		fullIncludePaths = append(fullIncludePaths, src.RebasedIncludePaths...)
+
+		err = dag.Select(ctx, dag.Root(), &src.ContextDirectory,
 			dagql.Selector{Field: "host"},
 			dagql.Selector{
 				Field: "directory",
 				Args: []dagql.NamedInput{
 					{Name: "path", Value: dagql.String(src.Local.ContextDirectoryPath)},
 					{Name: "include", Value: dagql.ArrayInput[dagql.String](dagql.NewStringArray(fullIncludePaths...))},
+					{Name: "noGitAutoIgnore", Value: dagql.NewBoolean(true)}, // Disable .gitignore applying since it's done above.
```
jedevc, line 988: Mild confusion here. I was thinking that we wouldn't need *any* logic in `loadSourceModuleContext`, it could all be done in `host.Directory`. What's the reason it can't be done there?

core/util.go
```diff
@@ -513,3 +517,163 @@ func getRefOrEvaluate[T fileOrDirectory](ctx context.Context, t T) (bkcache.Immu
...
+//     Otherwise, the pattern is recursive.
+//     If pattern is already starting with **, not change needed.
+//     If the pattern starts with *, it's not recursive but only matches the directory itself.
+//     Example: foo/bar stays foo/bar but foo becomes **/foo
+//
+//   - If a pattern is negative exclusion (starts with `!`) or targets directory only
+//     (ends with `/`), we treat is as a regular path then read the exclusion to make
+//     sure the recusive pattern is applied if needed.
+//     Example: !foo becomes foo then **/foo then !**/foo
+func extractGitIgnorePatterns(gitIgnoreContent string, parentDir string) []string {
```
jedevc: `parseGitIgnore`? This is a parsing function.

dagql/idtui/testdata/TestTelemetry/TestGolden/path-args
```diff
@@ -19,7 +19,7 @@ Expected stderr:
   ┆ file: Host.file(path: "/app/dagql/idtui/golden_test.go"): File!
   ┆ dir: Host.directory(path: "/app/dagql/idtui"): Directory!
   ┆ contextFile: Directory.file(path: "main.go"): File!
-  ┆ contextDir: Host.directory(path: "/app/dagql/idtui/viztest", exclude: [], noCache: true): Directory!
+  ┆ contextDir: Host.directory(path: "/app/dagql/idtui/viztest", exclude: [], noCache: true, contextDirectoryPath: "/app/dagql/idtui"): Directory!
```
jedevc: Hm. I don't think the `internal` args should be appearing in telemetry.

.changes/unreleased/Added-20250723-184041.yaml
jedevc: Ah, I see above this was discussed and agreed on.

That's fine, but since this is a breaking change, we need to make it as such in the changelog.

No compat layer required IMO, because modules can't call the `Host` api anyways.

core/schema/modulesource.go
jedevc: Yeah, we absolutely *need* something like this.

Currently, the performance is gonna be crappy because for every gitignore we're going to mount+read+unmount an entire directory using the containerd snapshotter. I imagine this is where the slowdown comes from.

I don't *think* we should block - but this is a pretty urgent follow-up, we shouldn't delay it.

core/schema/modulesource.go
TomChv: Because when we load a module, we extract the pattern from the `exclude` and `include` field of the dagger.json then we rebase and reformat them so they are all in the `include`.
But in `Host.Directory`, we only use the exclude field so having both led to unexpected behaviour and ignore.
So by getting the .gitignore and negate them so they are in the include, we get the correct behaviour.

jedevc: What do you mean by unexpected behavior? Why can't we just pass the `include` + `exclude` directly in, and also have the gitignore behavior, so the merging gets done there?

dagql/idtui/testdata/TestTelemetry/TestGolden/path-args
TomChv: I regenerate the tests and it's still displayed
Not a blocker but I'll ping @vito

TomChv:
@jedevc Comment applied, I wait for CI and I merge this PR

core/schema/modulesource.go
helderco: It's because of https://github.com/dagger/dagger/pull/10736#discussion_r2228174485

Module loading starts with including `dagger.json` and `<source_dir>/**` patterns. Then every pattern from `dagger.json` is added. If `dagger.json` has any excludes, they'll be negated and moved to the includes.

If module loading added `.gitignore` patterns to `Host.directory`'s excludes they would always be applied after the patterns in `dagger.json`. This way `.gitignore` patterns are applied after `<source_dir>/**` but before user added patterns in `dagger.json`, allowing users to override something specifically, if needed.

jedevc: Thinking about this again, in the context of https://github.com/dagger/dagger/pull/10883.

I don't think it should be possible to alter the `gitignore` patterns in `dagger.json`. I'm struggling to imagine a legitimate use case.

One issue is how this would interact with local vs git sources.
- If I have a `.gitignore`d file `foo.txt`, but I override it in `dagger.json` to `"include": ["foo.txt"]`, then it could be loaded.
- But once I push this to a remote module, and load it from a git source, then `foo.txt` is no longer there.

I think it makes more sense to have a consistent behavior.

(I'm a bit biased, because the implementation in #10883 doesn't do *any* pattern conversion - it directly adds gitignore support to the file transfer protocol. There's no extra ordering to do, it has to be applied separately. If this is a *required* feature, I can think about other ways to do it, but I think it's just potentially confusing to be able to override the patterns in `.gitignore`)

PR #10740, opened 2025-07-16
chore: refactor dagop mounting code

alexcb:
This introduces a new execInMount function, which deduplicates a lot of the boiler plate code needed to perform dagops.


core/exec_in_mount_util.go
```diff
@@ -0,0 +1,105 @@
...
+	return func() (string, bool) {
+		return fmt.Sprintf(format, a...), true
+	}
+}
+
+func ImmutableOp() (string, bool) {
+	return "", false
+}
+
+type FileOrDirectory interface {
```
alexcb: I had a real hard time naming this. I'm completely open to suggestions.

Bad suggestions:
- `FileOrDirectoryWithResultsAccessors`
- `ThingsThatAreTraditionallyReferencedByINodes`

jedevc: Happy if it's just called `FileOrDirectory`, best to probably keep it private though, it's only used here right?

core/exec_in_mount_util.go
jedevc, line 1: Happy to just have this in `util.go` as well.

core/exec_in_mount_util.go
```diff
@@ -0,0 +1,105 @@
...
+
+type ExecInMountSaveOpt func() (string, bool)
+
+func SaveResultsWithDesc(format string, a ...any) ExecInMountSaveOpt {
+	return func() (string, bool) {
+		return fmt.Sprintf(format, a...), true
+	}
+}
+
+func ImmutableOp() (string, bool) {
```
jedevc: :thinking: confused what this does!

core/exec_in_mount_util.go
```diff
@@ -0,0 +1,105 @@
...
+	"fmt"
+
+	"github.com/dagger/dagger/engine/buildkit"
+	bkcache "github.com/moby/buildkit/cache"
+	bkclient "github.com/moby/buildkit/client"
+)
+
+var errEmptyResultRef = fmt.Errorf("empty result reference")
+
+type ExecInMountSaveOpt func() (string, bool)
```
jedevc: I like this! WDYT about just having this return a struct though?

core/exec_in_mount_util.go
alexcb: making it private is a good idea, naming probably isn't as important that way.

core/exec_in_mount_util.go
alexcb: good call, that'll make `ImmutableOp()` slightly more clear :)

alexcb: I ended up changing all of this under ce05d821452e58685ba7c74cc7b90a670854659f to have it follow the `WithXXXOpt(.....)` pattern, which then configures the fields under the `execInMountOpt` struct.

If no options are specified then it'll default to not saving the results, and ultimately `ImmutableOp()` is no longer used.

core/exec_in_mount_util.go
alexcb: I ended up removing this entirely, nothing to see here :sweat_smile:

core/exec_in_mount_util.go
alexcb: `cat core/exec_in_mount_util.go >> core/util.go && rm core/exec_in_mount_util.go` done :)

PR #10741, opened 2025-07-16
Dagop glob

alexcb:


alexcb:
blocked by https://github.com/dagger/dagger/pull/10740

core/directory.go
```diff
@@ -337,12 +305,17 @@ func (dir *Directory) Glob(ctx context.Context, pattern string) ([]string, error
...
-		Path:           dir.Dir,
-		Callback: func(path string, info os.FileInfo) error {
+	_, err = execInMount(ctx, dir, func(root string) error {
+		resolvedDir, err := containerdfs.RootPath(root, dir.Dir)
+		if err != nil {
+			return err
+		}
+		return fsutil.Walk(ctx, resolvedDir, &fsutil.FilterOpt{
+			IncludePatterns: []string{pattern},
+		}, func(path string, info fs.FileInfo, err error) error {
```
jedevc: HAHA :tada: Been wanting to do this for ages, and *not* have the weird `WalkDirRequest`.

Looks good to me, and will approve, but there may be a quick easy way to avoid the two pattern matchers? Since now we control the call to `fsutil.Walk`.

Not entirely sure off the top-of-my-head, but if there is, fixing that while in the area would be lovely :grin:

alexcb: and here I was thinking surely the fact that there's two pattern matches must be done for a very particular reason and I would open a can of worms if I tried combining them :laughing: 

I'll take a closer look to see if it can be done.

alexcb: I was able to lift the fsutils code into dagger under 7ed3457a44f30efa3a802de3853cf99dccf79083 as a PoC.

Based on CI exploding, I suspect it needs some work. Interestingly the `TestGlob` integration tests passed for me locally, so I'll have to look a bit closer to ensure we have code coverage for the tree pruning code that's used.

Any thoughts on proceeding down this path?

jedevc: Yeah, there's probably something missing in our test suite :eyes:

Honestly, the lifting looks *significantly* simpler than I was imagining, yeah, this looks like a good path! The less dependency we have on `fsutil`, the better IMO (ideally we'd just isolate it to filesync eventually...)

We don't need to block on this though, up to you - if you wanna spin out a second PR with that, we can merge this one first.

alexcb: I had to pull in the patternmatcher lib, then made a few changes to it to allow us to interact with a single patter matcher. Best to look at the individual commits to see what I changed.

alexcb: > Interestingly the TestGlob integration tests passed for me locally

I suspect I fat-fingered something, because I went back in git history and the test failed, so the coverage was there all along.

util/patternmatcher/patternmatcher.go
```diff
@@ -0,0 +1,501 @@
+package patternmatcher
```
jedevc, line 2: Maybe include a package comment of what this is a fork/vendor of?

jedevc (approved):

Issue #7202, opened 2024-04-26
Git push


sagikazarmark:
### What are you trying to do?

First class or at least decent support for pushing to Git repositories.

### Why is this important to you?

Dagger already has first-class support for pulling from Git repositories (even if private repo support isn't trivial).

It's often important to push changes back to a repository. A couple use cases in mind:

- Regenerate SDKs from API descriptors upon change
- Update GitOps manifests when a new version is out

Unfortunately, feeding credentials into Git (whether it's SSH or HTTPS) is not trivial.

### How are you currently working around this?

There is already an existing example using [go-git](https://github.com/go-git/go-git): https://daggerverse.dev/mod/github.com/matipan/daggerverse/image-updater@7dafaf057c4524654903414fc3a7b5e2374f5695

Ideally, there should be a generic "git push" solution: whether it's built into Dagger or available as a separate module.

shykes:
In case it's useful, I recently started a new git module. It doesn't support push, but has a clean and simple foundation to add it.

https://daggerverse.dev/mod/github.com/shykes/git

sagikazarmark:
@shykes Sounds good!

What do you think would be the best way to implement authentication?

GitHub supports http+token auth, but SSH is probably still the most common way to authenticate with Git. CI systems (like GHA) probably work better with API tokens.

shykes:
I would start simple, and add a `[]SSHKey` field to the `Repo` state, with assorted `WithSSHKey` and `WithSSHKeys` to add them. then a `Push()` function that makes sure to write the ssh keys to the right place in the container, and calls `git push`. Start with passing the argument as-is. Then, iterate from there.

sagikazarmark:
Would it make sense to add `WithSSHKey` to the `Git` object instead (or as well) so they can be used to clone private repos as well?

shykes:
> Would it make sense to add `WithSSHKey` to the `Git` object instead (or as well) so they can be used to clone private repos as well?

Yes that makes perfect sense.

habib-source:
Any updates on this?

jedevc:
@cwlbraa also mentioned he has a pretty strong use case for this in https://github.com/dagger/container-use :tada:

IMO, it feels like we need a handful of new APIs:
- Authoring new commits.
  
[...]

PR #10742, opened 2025-07-17
add `Directory.withPatch`

vito:
Takes a Git-compatible patch string and applies it to the directory, using `git apply` under the hood.

This way you can make incremental changes to a directory without having to add entire new file contents at a time. Also supports binary diffs, so this is a bit beefier than `patch -p1`.

### DIFF
--- a/core/directory.go
+++ b/core/directory.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"slices"
@@ -25,6 +26,7 @@ import (
 	"github.com/pkg/errors"
 	fstypes "github.com/tonistiigi/fsutil/types"
 	"github.com/vektah/gqlparser/v2/ast"
+	"go.opentelemetry.io/otel/trace"
 
 	"dagger.io/dagger/telemetry"
 	"github.com/dagger/dagger/dagql"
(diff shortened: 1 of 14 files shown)


core/directory.go
```diff
@@ -496,6 +498,68 @@ func (dir *Directory) WithNewFileDagOp(ctx context.Context, dest string, content
...
+	defer stdio.Close()
+
+	newRef, err := query.BuildkitCache().New(ctx, parentRef, bkSessionGroup, bkcache.WithRecordType(bkclient.UsageRecordTypeRegular),
+		bkcache.WithDescription("patch"))
+	if err != nil {
+		return nil, err
+	}
+	err = MountRef(ctx, newRef, bkSessionGroup, func(root string) (rerr error) {
+		var stdout, stderr strings.Builder
+		apply := exec.Command("git", "apply", "-")
```
shykes, line 537: Just double-checking, this executes the `git` binary on the host? Is this a new runtime dependency, or did it already exist?

Not a dealbreaker, just curious as it has implications for packaging.

jedevc: This is the `git` binary builtin to the engine - which we already use as the backend for all the `dag.Git` operations, so no packaging changes.

core/directory.go
```diff
@@ -496,6 +498,68 @@ func (dir *Directory) WithNewFileDagOp(ctx context.Context, dest string, content
 	return dir, nil
 }
 
+func (dir *Directory) WithPatch(ctx context.Context, patch string) (*Directory, error) {
```
jedevc, line 501: I think this is pretty reasonable, but it does feel a bit weird that it's a git operation not associated with the `Git` type (see soooomewhat similar discussion of patches in the context of https://github.com/dagger/dagger/issues/7202#issuecomment-2966176223).

Something like `dir.AsGit().WithPatch().Tree()` feels like the thing I'd want there? But then there's not really a concept of a staging area there, so bleh, don't really know.

---

I'm happy to introduce this method now, and keep working on how this integrates with the Git API down the road (with the caveat we might have multiple ways of doing it in the future?).

docs/docs-graphql/schema.graphqls
```diff
@@ -1579,6 +1579,16 @@ type Directory {
     permissions: Int = 420
   ): Directory!
 
+  """Retrieves this directory with the given Git-compatible patch applied."""
+  withPatch(
+    """
+    Patch to apply (e.g., "diff --git a/file.txt b/file.txt\nindex
+    1234567..abcdef8 100644\n--- a/file.txt\n+++ b/file.txt\n@@ -1,1 +1,1
+    @@\n-Hello\n+World\n").
```
jedevc, line 1587: Beautiful.

core/directory.go
shykes: I don't think this is a git-specific operation. At least in the context of coding agents, it's pretty standard for regular file operations to include either a search-replace, or a patch operation. I don't know if that's the use case @vito had in mind, but it's definitely a valid one. And it's unrelated to git.

core/directory.go
shykes: Ok good to know. The dependency introduced by `dag.Git` is an important one to know about.

core/directory.go
vito: Yeah I had to look this up - this is unrelated to Git, just happens to use `git apply` under the hood, for maximal compatibility with fun things like binary diffs. Despite using the `git` CLI it doesn't need to even run against a Git repo - any directory will do.

core/directory.go
```diff
@@ -496,6 +498,62 @@ func (dir *Directory) WithNewFileDagOp(ctx context.Context, dest string, content
...
+	defer stdio.Close()
+
+	newRef, err := query.BuildkitCache().New(ctx, parentRef, bkSessionGroup, bkcache.WithRecordType(bkclient.UsageRecordTypeRegular),
+		bkcache.WithDescription("patch"))
+	if err != nil {
+		return nil, err
+	}
+	err = MountRef(ctx, newRef, bkSessionGroup, func(root string) (rerr error) {
+		apply := exec.Command("git", "apply", "-")
+		apply.Dir = root
```
vito: oops, I think this needs to respect `dir.Dir`? will test

sipsma (approved):

sipsma:
CI errors are https://github.com/dagger/dagger/issues/10750, not related, merging

PR #10744, opened 2025-07-17
Generic env bindings

shykes:
Simplify `Env` with generic bindings. This is part of #10370 

- [ ] New core type: `Object`. A generic type that matches any Dagger object. Must be an interface, supported by all SDKs.
- [x] New core function: `Env.withBinding`. Create a new binding with the given `ID` as value.
- [x] New core function: `Env.binding`. Retrieve a binding.
- [x] New core function: `Env.bindings`. List all bindings.
- [ ] Deprecate `Env.with[XXX]Binding`
- [ ] Deprecate `Env.inputs` and `Env.outputs`
- [ ] Alias `LLM.withBinding`, `LLM.bindings`, `LLM.binding` for convenience

## Generic Object

The current implementation uses a temporary stopgap: a scalar `ID` type. This works, but SDKs won't generate useful bindings for it.

Instead, @vito proposes:

> We could make it take an interface that all objects satisfy (e.g. in Go every object implements XXX_GraphQLID and XXX_GraphQLType
> [...]
> In Go it would probably look like this:
> ```
> type Object interface {
>     XXX_GraphQLType() string
>     XXX_GraphQLID(context.Context) (string, error)
> }
> ```

> The downside to this approach is that it's another box for every SDK maintainer to check. They'll have to special-case an ID argument to take an Object interface (analogous to how ContainerID actually takes a Container), or come up with some other language-specific tactic. Or just leave users to do the boilerplate themselves, but that's no fun.

### DIFF
--- a/cmd/codegen/generator/go/templates/src/internal/dagger/dagger.gen.go.tmpl
+++ b/cmd/codegen/generator/go/templates/src/internal/dagger/dagger.gen.go.tmpl
@@ -5,3 +5,9 @@ package dagger
 
 {{ template "_dagger.gen.go/defs.go.tmpl" . }}
 {{ end }}
+
+// Object is an interface that matches all Dagger object types
+type Object interface {
+    XXX_GraphQLType() string
+    XXX_GraphQLID(context.Context) (string, error)
+}
--- a/core/ids.go
+++ b/core/ids.go
@@ -1,10 +1,69 @@
 package core
 
 import (
+	"context"
+	"fmt"
+
 	"github.com/dagger/dagger/dagql"
+	"github.com/dagger/dagger/dagql/call"
+	"github.com/vektah/gqlparser/v2/ast"
 )
 
 type JSONValueID = dagql.ID[*JSONValue]
+
+type ID string
+
+func (id ID) Load(ctx context.Context, srv *dagql.Server) (dagql.AnyResult, error) {
+	var callID = new(call.ID)
+	err := callID.Decode(string(id))
+	if err != nil {
+		return nil, err
+	}
+	return srv.Load(ctx, callID)
+}
+
+func (ID) Type() *ast.Type {
+	return &ast.Type{
+		NamedType: "ID",
+		NonNull:   true,
+	}
+}
+
+var _ dagql.Typed = ID("")
+
+func (ID) TypeName() string {
+	return "ID"
+}
+
+func (ID) TypeDescription() string {
+	return "A generic object ID"
+}
+
+var _ dagql.ScalarType = ID("")
+
+var _ dagql.Input = ID("")
+
+func (id ID) Decoder() dagql.InputDecoder {
+	return id
+}
+
+func (id ID) ToLiteral() call.Literal {
+	return call.NewLiteralString(string(id))
+}
+
+func (ID) DecodeInput(val any) (res dagql.Input, err error) {
+	switch x := val.(type) {
+	case string:
+		if x == "" {
+			return nil, nil
+		}
+		return ID(x), nil
+	case *call.ID:
+		return ID(x.Display()), nil
+	default:
+		return nil, fmt.Errorf("cannot convert %T to ID", val)
+	}
+}
 
 type ContainerID = dagql.ID[*Container]
 
--- a/cmd/codegen/generator/go/templates/format.go
+++ b/cmd/codegen/generator/go/templates/format.go
@@ -47,6 +47,9 @@ func (f *FormatTypeFunc) FormatKindScalarBoolean(representation string) string {
 }
 
 func (f *FormatTypeFunc) FormatKindScalarDefault(representation string, refName string, input bool) string {
+	if refName == "ID" {
+		return "Object"
+	}
 	if obj, ok := strings.CutSuffix(refName, "ID"); input && ok {
 		representation += "*" + f.scope + formatName(obj)
 	} else {
--- a/core/env.go
+++ b/core/env.go
@@ -111,6 +111,16 @@ func (env *Env) WithInput(key string, val dagql.Typed, description string) *Env 
 	return env
 }
 
+// Add an input (read-only) binding to the environment
+func (env *Env) WithBinding(name string, object dagql.AnyResult, description string) *Env {
+	env = env.Clone()
+	binding := &Binding{Key: name, Value: object, Description: description, env: env}
+	_ = binding.ID() // If val is an object, force its ingestion
+	// FIXME: rename 'inputs' to 'bindings' internally
+	env.inputsByName[name] = binding
+	return env
+}
+
 // Register the desire for a binding in the environment
 func (env *Env) WithOutput(key string, expectedType dagql.Type, description string) *Env {
 	env = env.Clone()
@@ -143,6 +153,11 @@ func (env *Env) Outputs() []*Binding {
 
 // List all inputs in the environment
 func (env *Env) Inputs() []*Binding {
+	return env.Bindings()
+}
+
+// List all bindings
+func (env *Env) Bindings() []*Binding {
 	res := make([]*Binding, 0, len(env.inputsByName))
 	for _, v := range env.inputsByName {
 		res = append(res, v)
@@ -152,22 +167,32 @@ func (env *Env) Inputs() []*Binding {
 
 // Lookup an input binding
 func (env *Env) Input(key string) (*Binding, bool) {
+	return env.Binding(key)
+}
+
+// Lookup a binding
+func (env *Env) Binding(name string) (*Binding, bool) {
 	// next check for values by ID
-	if val, exists := env.objsByID[key]; exists {
+	if val, exists := env.objsByID[name]; exists {
 		return val, true
 	}
 	// next check for values by name
-	if val, exists := env.inputsByName[key]; exists {
+	if val, exists := env.inputsByName[name]; exists {
 		return val, true
 	}
 	return nil, false
 }
 
 // Remove an input
-func (env *Env) WithoutInput(key string) *Env {
+func (env *Env) WithoutBinding(name string) *Env {
 	env = env.Clone()
-	delete(env.inputsByName, key)
+	delete(env.inputsByName, name)
 	return env
+}
+
+// Remove an input
+func (env *Env) WithoutInput(key string) *Env {
+	return env.WithoutBinding(key)
 }
 
 func (env *Env) Ingest(obj dagql.AnyResult, desc string) string {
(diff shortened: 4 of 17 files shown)


vito:
@shykes I like the API simplification for sure, but have some questions:

---

How are you feeling about descriptions being optional?

They're still extremely load-bearing for the LLM use case, and we actually don't use names at all to prevent agents from treating like them variables in arguments to places.

So maybe `LLM.withBinding` should _just_ take an object and a description? :thinking: 

---

I think we'll want a bit of sugar for adding object bindings, so you don't have to do this:

```go
ctrID, err := ctr.ID(ctx)
if err != nil {
  return err
}

dirID, err := dir.ID(ctx)
if err != nil {
  return err
}

env.
  WithBinding("ctr", dagger.ID(ctrID), "ctr description").
  WithBinding("dir", dagger.ID(dirID), "dir description")
```

---

Have you considered adding a single `asBinding` function to every object type?

```go
env.
  WithBinding(ctr.AsBinding("ctr")).
  WithBinding(dir.AsBinding("dir")).
  WithBinding(dag.StringBinding("foo", "bar")) // winging it here
```

shykes:
> How are you feeling about descriptions being optional?

It's part of my proposal :) See #10370 , section "Descriptions become optional". (Unfortunetely github discussions don't support anchor links... 😭 )

> They're still extremely load-bearing for the LLM use case, and we actually don't use names at all to prevent agents from treating like them variables in arguments to places.
> 
> So maybe `LLM.withBinding` should _just_ take an object and a description? 🤔

But that would cause Env to drift even further towards being LLM-only, which is the opposite of what I'm trying to do..

How would we map bindings to shell variables if we removed the name?

> I think we'll want a bit of sugar for adding object bindings, so you don't have to do this:
> 
> ```go
> ctrID, err := ctr.ID(ctx)
> if err != nil {
>   return err
> }
> 
> dirID, err := dir.ID(ctx)
> if err != nil {
>   return err
> }
> 
> env.
>   WithBinding("ctr", dagger.ID(ctrID), "ctr description").
>   WithBinding("dir", dagger.ID(dirID), "dir description")
> ```

I was kind of hoping we could do this:

```go
env.
  WithBinding("ctr", ctr, "ctr description").
  WithBinding("dir", dir, "dir description")
```

I thought that would just work out of the box, but maybe that's naive of me?

vito:
@shykes 

> But that would cause Env to drift even further towards being LLM-only, which is the opposite of what I'm trying to do..

I'm suggesting this for `LLM.withBinding`, not `Env.withBinding` :smile:  Seems like a good opportunity to make it more natural for that domain without compromising in the underlying `Env` implementation details.

To keep it consistent with `Env` we could keep the required `name` argument that we have today. My only ask is for the `description` to be required on the `LLM` aliases, because it's definitely not optional there.

---

> I was kind of hoping we could do this:

Definitely not out of the box - you can see the codegen just has a `string` argument for `value`, so that'll be a compiler error.

We could make it take an interface that all objects satisfy (e.g. in Go every object implements `XXX_GraphQLID` and `XXX_GraphQLType`), but then it can't accept arbitrary strings either, so that would give rise to `WithObjectBinding` and `WithStringBinding` I suppose.

The last option I suggested would work around all that, at the expense of one automagic field added to every object type.

For LLM, it could be like this - we could move the descriptions entirely out of `Env`:

```go
dag.LLM().WithBinding(ctr.AsBinding("ctr"), "Some container")
```

shykes:
> > But that would cause Env to drift even further towards being LLM-only, which is the opposite of what I'm trying to do..
> 
> I'm suggesting this for `LLM.withBinding`, not `Env.withBinding` 😄 Seems like a good opportunity to make it more natural for that domain without compromising in the underlying `Env` implementation details.
> 
> To keep it consistent with `Env` we could keep the required `name` argument that we have today. My only ask is for the `description` to be required on the `LLM` aliases, because it's definitely not optional there.

Oooh I see! Yeah that makes sense. I see the temptation to go all the way and make `name` optional... But how would we square that with the underlying `Env.withBinding` still requiring `name`? Maybe we could have the llm itself generate the name? :-P But on expensive LLMs it would be slow and expensive... Now if we had a reliable way to lookup any LLM's "cheaper sibling"... that would be a cool pattern to use, and I bet the community would find other uses for it :)

> Definitely not out of the box - you can see the codegen just has a `string` argument for `value`, so that'll be a compiler error.

Now that you say it, it makes sense 😭.

> We could make it take an interface that all objects satisfy (e.g. in Go every object implements `XXX_GraphQLID` and `XXX_GraphQLType`), but then it can't accept arbitrary strings either, so that would give rise to `WithObjectBinding` and `WithStringBinding` I suppose.

I already have a plan for scalars: enter #10680  😁

So if I understood correctly, your suggestion of making `ID` an interface that all objects satisfy, could work?

vito:
> But how would we square that with the underlying Env.withBinding still requiring name? Maybe we could have the llm itself generate the name? :-P But on expensive LLMs it would be slow and expensive... Now if we had a reliable way to lookup any LLM's "cheaper sibling"... that would be a cool pattern to use, and I bet the community would find other uses for it :)

lol, yeah. To be honest the easiest thing is probably to just keep it, since we already require it. Today it does nothing, but maybe it will in the future; better to have everyone passing it now than figure out how to add it later.

> So if I understood correctly, your suggestion of making `ID` an interface that all objects satisfy, could work?

More like `Object`, since `ID` already exists as a scalar value, but yeah.

In Go it would probably look like this:

```go
type Object interface {
	XXX_GraphQLType() string
	XXX_GraphQLID(context.Context) (string, error)
}
```

In fact, we've dealt with this sort of thing already in our very own `dagger` prompt mode implementation:

https://github.com/dagger/dagger/blob/e71fe8e0e6912a74b47639576cdc5c3570b1a83a/cmd/dagger/llm.go#L320-L337

The downside to this approach is that it's another box for every SDK maintainer to check. They'll have to special-case an `ID` argument to take an `Object` interface (analogous to how `ContainerID` actually takes a `Container`), or come up with some other language-specific tactic. Or just leave users to do the boilerplate themselves, but that's no fun.

shykes:
> More like `Object`, since `ID` already exists as a scalar value, but yeah.

It might not change anything, but I believe `ID` does *not* exist as a scalar type in `main`? Since I added its definition in this PR (`type ID dagql.String`) and it didn't seen to conflict with anything?

> The downside to this approach is that it's another box for every SDK maintainer to check. They'll have to special-case an `ID` argument to take an `Object` interface (analogous to how `ContainerID` actually takes a `Container`), or come up with some other language-specific tactic. Or just leave users to do the boilerplate themselves, but that's no fun.

OK I don't fully understand this... Will let it sit for a while.

Thanks!

vito:
> It might not change anything, but I believe `ID` does _not_ exist as a scalar type in `main`? Since I added its definition in this PR (`type ID dagql.String`) and it didn't seen to conflict with anything?

Oh yeah I meant assuming the rest of the context of this PR. Also I didn't notice it was an alias defined only in `core/` and not exposed as `ID` through the schema yet, which explains why it's just a `string` here. Not sure if that's intentional or just missing a `dagql` tweak to expose it. If it's intentional, though, that would make the `value` argument be the more strict `ID!` type rather than `String!` which seems like it shuts the door on passing arbitrary JSON values in. Another manifestation of the same problem I mentioned here:

> We could make it take an interface that all objects satisfy (e.g. in Go every object implements `XXX_GraphQLID` and `XXX_GraphQLType`), but then it can't accept arbitrary strings either, so that would give rise to `WithObjectBinding` and `WithStringBinding` I suppose.

shykes:
@vito I specifically *don't* want to support passing arbitrary values in. That's why I have this other PR #10680 to make it very easy to represent any JSON value (*including scalars*) as an object. That way, 100% of bindings are objects.

vito:
My bad - makes sense! Went to edit but ya caught me first. :P

shykes:
@jedevc @eunomie @vito @sipsma quick recap, I am stuck no the best way to implement a generic ID, that matches all object IDs, and doesn't add too much work to SDK devs (ideally zero work).

I'm trying to list options from @vito 's comments above, but am having trouble doing so.

shykes:
Thanks to @vito 's guidance, I got a generic `Object` type to work in the Go SDK :)

TODO: do the same in other SDKs

cmd/codegen/generator/go/templates/src/internal/dagger/dagger.gen.go.tmpl
```diff
@@ -4,4 +4,10 @@
...
 {{ template "_dagger.gen.go/defs.go.tmpl" . }}
-{{ end }}
\ No newline at end of file
+{{ end }}
+
+// Object is an interface that matches all Dagger object types
+type Object interface {
+    XXX_GraphQLType() string
+    XXX_GraphQLID(context.Context) (string, error)
+}
```
helderco, line 13: Interesting. Should `ID()` be added, or some objects don't have an `ID()` function?

Re: Python impl., all object types extend from a base class in the library. That could be used directly as the type, unless we add the `id()` function, but even that "interface" already exists in the Python client library too.

helderco:
> TODO: do the same in other SDKs

Let me know when it's ready for me to help out with the Python implementation!

cmd/codegen/generator/go/templates/src/internal/dagger/dagger.gen.go.tmpl
vito: `ID()` can't be added unfortunately because they all have a different return type (`ContainerID`, `DirectoryID`, etc)

core/ids.go
```diff
@@ -1,11 +1,70 @@
...
+type ID string
+
+func (id ID) Load(ctx context.Context, srv *dagql.Server) (dagql.AnyResult, error) {
+	var callID = new(call.ID)
+	err := callID.Decode(string(id))
+	if err != nil {
+		return nil, err
+	}
+	return srv.Load(ctx, callID)
+}
```
jedevc, line 23: Idea: we should support a generic `Query.loadID` function.

I have this idea that we should be able to copy-paste IDs out of cloud, and paste them into the shell to replay them and debug them :thinking:

That function would let you do that.

jedevc: Obviously, the big problem is around not having access to the same local files... but it would be a start.

helderco: What would be the return type of `Query.loadID`?

jedevc: I guess I would like it to somehow be `Object` :thinking: 

Essentially the empty interface, like `interface{}` in go. All objects implement it, anything can be cast into it.

Then you could do `loadID().AsContainer()` and similarly.

PR #10747, opened 2025-07-17
add `File.search`, `Directory.search`

vito:
* based on #10742 
* uses `ripgrep --json` under the hood (added to engine image)
* returns a structured `[SearchResult!]!` output


docs/docs-graphql/schema.graphqls
```diff
@@ -3742,6 +3818,22 @@ The `ScalarTypeDefID` scalar type represents an identifier for an object of type
 """
 scalar ScalarTypeDefID
 
+type SearchResult {
+  filePath: String!
```
jedevc, line 3891: This feels a little odd to me - definitely makes sense in the context of `Directory.search`, but `File.search` doesn't really have a `path`, it only really has a `name`.

No real other suggestions, so SGTM :tada:

docs/docs-graphql/schema.graphqls
```diff
@@ -3742,6 +3818,22 @@ The `ScalarTypeDefID` scalar type represents an identifier for an object of type
...
+
+  """A unique identifier for this SearchResult."""
+  id: SearchResultID!
+
+  lineNumber: Int!
```
jedevc, line 3896: Any chance we could instead get:
- `start` / `end` (the raw position in the file, useful for byte slicing, etc)
- `startLine` / `endLine`
- `startColumn` / `endColumn`

Naming hard obviously, but feels like we should definitely have a way to get the start+end, and being able to get the columns feels useful too.

Maybe even a new `FilePosition` type might be useful here? For example, we already having something kinda similar in the `SourceMap` type (if we like the idea, I can update the `SourceMap` to have that shared type). That type could then be used for some cool things - e.g. you could have `File.read(start, end FilePosition)`.

core/schema/file.go
```diff
@@ -50,6 +50,25 @@ func (s *fileSchema) Install() {
...
+			Doc(`Searches for content matching the given pattern, which may be a regular expression or a literal string.`).
+			Args(
+				dagql.Arg("pattern").Doc(`The text to match.`),
+				dagql.Arg("literal").Doc(`Interpret the pattern as a literal string instead of a regular expression.`),
+				dagql.Arg("multiline").Doc(`Enable searching across multiple lines.`),
+			),
+		dagql.NodeFunc("withReplaced",
+			DagOpFileWrapper(s.srv, s.withReplaced,
+						WithPathFn(keepParentFile[fileReplaceArgs]))).
+			View(AllVersion). // entries returns different results in different versions
```
jedevc: Don't need this here, this is just because `entries` is a weird API that uses the `View` - if we don't use that, we don't need it.

core/schema/file.go
```diff
@@ -50,6 +50,25 @@ func (s *fileSchema) Install() {
...
 			),
+		dagql.NodeFunc("search", DagOpWrapper(s.srv, s.search)).
+			View(AllVersion). // entries returns different results in different versions
+			Doc(`Searches for content matching the given pattern, which may be a regular expression or a literal string.`).
+			Args(
+				dagql.Arg("pattern").Doc(`The text to match.`),
+				dagql.Arg("literal").Doc(`Interpret the pattern as a literal string instead of a regular expression.`),
+				dagql.Arg("multiline").Doc(`Enable searching across multiple lines.`),
+			),
+		dagql.NodeFunc("withReplaced",
```
jedevc: A bit less convinced on this particular API, but no real objections, it feels useful :tada:

But I think it should support the same args as `search` (e.g. `literal`, `multiline`)?

Kinda wish we supported multiple return values :cry: It would be cool to support returning what actually got replaced.

vito: Ah yeah this is me doing more exploratory stuff on top of the same branch to see what it takes to build a decent coding agent. Really it's all still WIP, finding a lot of stuff to do.

> But I think it should support the same args as `search` (e.g. `literal`, `multiline`)?

I started down this path but backed out due to complexity. For the use case driving this (LLM coding agents) regexes aren't needed and the larger API surface area seems more likely to cause a mess. For example, `Directory.search` makes more sense with regexp as its default, but that's NOT what I want by default for `File.withReplaced` because its core use case is passing in large chunks of code. Requiring the LLM to escape all that, or remember to pass `literal: true`, doesn't seem worth the trouble.

docs/docs-graphql/schema.graphqls
vito: Had similar temptations but didn't feel happy about adding more complexity to it yet. Hadn't thought about the parallels to `SourceMap` - if there's at least a reusable type we can get out of it, that feels a lot better! Will look into it.

docs/docs-graphql/schema.graphqls
vito: Well - to me the `name` is still a path, it's just `foo.txt` in an otherwise empty directory. I don't think it's _wrong_ for this to be here, or for it to be populated in file search results.

I did deliberate for a bit on adding a `File.path` attribute that would be e.g. `core/file.go` in `coreDir.File("file.go")` but backed out as it seemed like an information leak.

docs/docs-graphql/schema.graphqls
vito: Having looked into it again, I'm still inclined to just wait until someone needs these more detailed APIs. The complexity arises in the fact that a line can actually have _many_ submatches, which means we would need a whole new type like `[]SearchMatches` or something to carry pairs of start/end offsets within the `MatchedLines`. `EndLine` can be deduced from `LineNumber` + counting linebreaks in `MatchedLines`. Searching for multiline strings is a bit of a fringe case so not sure I'd want to make the API favor it too much.

Issue #10738, opened 2025-07-16
`unknown type` errors when using interfaces


sipsma:
(This is currently only a known issue on main, not in a release yet, just creating to track fixing it before next release)

The evals workflow has been failing with errors along the lines of `instantiate: toSelectable: unknown type EvaluatorEval`. Similar errors reported by users testing off main w/ interfaces.

Caused by https://github.com/dagger/dagger/pull/10620 

Have a basic idea of what's wrong here, working on a fix

chrisjpalmer:
Hi there. I was able to reproduce this issue with [my demo project](https://github.com/chrisjpalmer/dagger-deps-example) on the current main branch. Here are the details:

## No issue on 0.18.12

When running my demo project on 0.18.12, there is no issue, all works as expected.

Trace: https://dagger.cloud/nine/traces/59c4e0afbc3271db9f6eb3a5baf855cd

```sh
dagger -m github.com/dagger/dagger@v0.18.12 <<EOF
  dev |
  with-directory example \$(git https://github.com/chrisjpalmer/dagger-deps-example | head | tree) |
  with-workdir example |
  with-exec -- dagger -c 'workflow-pretend \$(secrets-manager) | doit'
EOF
```

## Issue on current main

When running on the current main branch, this following error occurs

Trace: https://dagger.cloud/nine/traces/35e514dd9cad792f2b3277754e86c908

```sh
dagger -m github.com/dagger/dagger@main <<EOF
  dev |
  with-directory example \$(git https://github.com/chrisjpalmer/dagger-deps-example | head | tree) |
  with-workdir example |
  with-exec -- dagger -c 'workflow-pretend \$(secrets-manager) | doit'
EOF

Error: instantiate: toSelectable: unknown type "SecretsManager"                                                                                                                                                   
! process "dagger -c workflow-pretend $(secrets-manager) | doit" did not complete successfully: exit code: 1
```

PR #10749, opened 2025-07-17
fix interface object loading

sipsma:
Fixes https://github.com/dagger/dagger/issues/10738

There was a very subtle issue where `ObjectResult`s were accidentally being "downgraded" to `Result` when dagql call internals were attempting to unwrap them. This was due to unintentionally calling DerefValue on the embedded `Result` of `ObjectResult` and returning the received, which was a `Result` rather than full `ObjectResult`.

This broke interface loading specifically because we currently rely on some objects already being selectable rather than needing to load their type information; if the object is an interface value from some other module unknown to the current one we can't load the type and error out.

This technically restores us back to the previous way this all worked, but does still feel like very thin ice. Need this regression fixed before the release though, so kicking the can on reforming the whole interface system here.

---

draft TODOs:
- [x] Add integ test coverage; this was only caught by the `evals` CI on main (which work again with this change in place), need to be able to catch earlier in integ tests though


dagql/objects.go
```diff
@@ -418,6 +418,14 @@ func (r ObjectResult[T]) MarshalJSON() ([]byte, error) {
 	return r.Result.MarshalJSON()
 }
 
+func (r ObjectResult[T]) DerefValue() (AnyResult, bool) {
+	derefableSelf, ok := any(r.self).(DerefableResult)
+	if !ok {
+		return r, true
```
sipsma, line 424: To expand on the PR description, this `return r, true` is the crucial line. Previously calling `DerefValue` on an `ObjectResult` went to `Result.DerefValue` because `Result` is embedded in `ObjectResult`. That resulted in us returning `r` as in the embedded `Result` rather than the `r` here, which is `ObjectResult`.

marcosnils:
> * Add integ test coverage; this was only caught by the `evals` CI on main (which work again with this change in place), need to be able to catch earlier in integ tests though

curious why current interfaces integ tests didn't catch this one :thinking:

sipsma:
> > * Add integ test coverage; this was only caught by the `evals` CI on main (which work again with this change in place), need to be able to catch earlier in integ tests though
> 
> curious why current interfaces integ tests didn't catch this one 🤔

@marcosnils The conditions that trigger this bug are highly specific. The existing integ tests didn't quite hit the right ones in the right order to find it.

sipsma:
Added integ test coverage now too. Had to add another module to the integ test setup because this requires you get an object from one module, use it to implement an interface in another dependency module and then *that* dep module passes the object again to its own dep to implement another interface (😵‍💫)

---

@chrisjpalmer also verified that this fixes the problem you hit on main, the following works for me:
```
dagger -m github.com/sipsma/dagger@fix-unknown-type <<EOF
  dev |
  with-directory example \$(git https://github.com/chrisjpalmer/dagger-deps-example | head | tree) |
  with-workdir example |
  with-exec -- dagger -c 'workflow-pretend \$(secrets-manager) | doit'
```

chrisjpalmer:
Thank you so much @sipsma ! Really appreciate this :)

Issue #10750, opened 2025-07-17
`session cache is closed` errors on main


sipsma:
Seeing some flakes on main+PRs with `session cache is closed`, e.g. https://dagger.cloud/dagger/traces/2c21ffb6385408be6b9b89cf9e7bec4d?listen=ac11982282afb038&listen=87dded5a7770eee9&listen=52f3eb5dc55d05b4&listen=b4c672b2d466aa9f#165234f9b49aa894:L1735

https://github.com/dagger/dagger/pull/10708 introduced that error case; it's *correctly* error'ing because we shouldn't be trying to use session caches that are closed (and doing so can result in memory leaks), so need to find where we are still trying to do that and fix it.

PR #10751, opened 2025-07-17
Fix more uses of closed dagql servers

sipsma:
Will fix https://github.com/dagger/dagger/issues/10750

Draft, haven't had success repro'ing locally yet so gonna parallelize in CI


vito:
This trace hit an error but I _think_ it's just a network flake bubbling up in a way I'm not used to seeing - retriggered:

* https://dagger.cloud/dagger/traces/d9a1a48eab60ba434d00cfd3bf088e2a

PR #10752, opened 2025-07-18
exp: allow to run dagger 'tasks' from a dagger.yaml file

eunomie:
This is experimental, just to see if that could be a good idea or not.

All start with a `dagger.yaml` file you can put in your project. The `dagger.yaml` file is composed of two sections: `vars` and `tasks`.

`vars` allow to define some variable (could be basic variable like strings, or more complex using any dagger primitives).

For instance this is declaring a Go version and defining a `ctr` variable representing a basic container to work on a go project:

```yaml
vars:
  go_version: 1.24.4
  ctr: >-
    $(container |
    from golang:$go_version-alpine |
    with-exec go install golang.org/x/tools/cmd/goimports@latest |
    with-exec go install mvdan.cc/gofumpt@latest |
    with-workdir /go/src |
    with-mounted-file /go/src/go.mod go.mod |
    with-mounted-file /go/src/go.sum go.sum |
    with-mounted-cache /root/.cache $(cache-volume go-cache) |
    with-mounted-cache /go/pkg/mod $(cache-volume go-pkg-mod) |
    with-exec go mod download |
    with-mounted-directory /go/src .)
```

Those variables will be translated as dagger shell variables, so available as `$go_version` and `$ctr`.

`tasks` allow to define some tasks to run. They have a name and a body that is a dagger shell command. This command can use previously defined variables.

For instance:

```yaml
tasks:
  terminal: >
    $ctr | terminal

  go:test: >
    $ctr |
    with-exec -- go test -coverpkg=./... -coverprofile=.coverage -shuffle=on ./... |
    stdout
```

This will define two tasks, one `terminal` that will open a terminal on the dev environment defined by `$ctr` and the second will run tests inside this container.

Usage:

- dagger task: list all available tasks
- dagger task TASK: will run the corresponding task by concatenating all the vars and the task body and run it as a dagger shell command

To test it, you can use `daggerfile` branch of `eunomie/docker-runx` project that contains a quick example: https://github.com/eunomie/docker-runx/tree/daggerfile

Clone the project, switch to daggerfile branch and run `dagger task` then `dagger task go:test`

### DIFF
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -151,6 +151,7 @@ func init() {
 		shellCmd,
 		clientCmd,
 		mcpCmd,
+		taskCmd,
 	)
 
 	rootCmd.AddGroup(moduleGroup)
--- a/cmd/dagger/task.go
+++ b/cmd/dagger/task.go
@@ -1,0 +1,93 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"syscall"
+
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	DaggerFileName = "dagger.yaml"
+)
+
+var taskCmd = &cobra.Command{
+	Use:   "task [TASK]",
+	Short: "Run a task defined in a dagger.yaml",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return runTask(cmd.Context(), args)
+	},
+	Annotations: map[string]string{
+		"experimental": "true",
+	},
+}
+
+func runTask(ctx context.Context, args []string) error {
+	daggerfile, err := readDaggerFile()
+	if err != nil {
+		return err
+	}
+
+	if len(args) == 0 {
+		fmt.Println("Available tasks:")
+		for taskName := range daggerfile.Tasks {
+			fmt.Println("- ", taskName)
+		}
+		return nil
+	}
+
+	if len(args) > 1 {
+		return fmt.Errorf("too many arguments")
+	}
+
+	taskName := args[0]
+
+	taskDef, err := daggerfile.FullDefinition(taskName)
+	if err != nil {
+		return err
+	}
+
+	binary, err := os.Executable()
+	if err != nil {
+		return err
+	}
+	return syscall.Exec(binary, []string{binary, "-c", taskDef}, os.Environ())
+}
+
+type DaggerFile struct {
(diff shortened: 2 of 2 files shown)


jedevc:
I quite like this :grin: Especially as we've been looking for a way to reduce/remove wrappers around dagger. This does like like the approach of just "bring the wrapper into dagger".

One issue I'm a little worried about is how this would interact with the work around environments, etc, that @shykes is currently looking at.

Would be curious to hear other's thoughts for sure, but it looks neat and I feel like I'd use it :tada:

eunomie:
Yes, I need to dig more into the environments to see how that fits.

One thing we also discussed on [Discord](https://discord.com/channels/707636530424053791/708371226174685314/1395792028423684257) is shell is already capable to define functions for instance.
So we could have something similar using only the shell syntax instead of adding a new yaml file (or any other)

Copy-pasting the discord example:

```shell
#!/usr/bin/env dagger

go_version=1.24.4

ctr() {
  container |
  from golang:$go_version-alpine |
  with-exec go install golang.org/x/tools/cmd/goimports@latest |
  with-exec go install mvdan.cc/gofumpt@latest |
  with-workdir /go/src |
  with-mounted-file /go/src/go.mod go.mod |
  with-mounted-file /go/src/go.sum go.sum |
  with-mounted-cache /root/.cache $(cache-volume go-cache) |
  with-mounted-cache /go/pkg/mod $(cache-volume go-pkg-mod) |
  with-exec go mod download |
  with-mounted-directory /go/src .
}

terminal() {
  ctr | terminal
}

go_test() {
  ctr |
  with-exec -- go test -coverpkg=./... -coverprofile=.coverage -shuffle=on ./... |
  stdout
}
```

It still requires some changes/tooling to expose the right functions and make them easy to use, but the idea is really close.

One thing I still like with the yaml version compared to the shell one is it's just a bit more declarative. It's not completely true as it only declares some function bodies in the end, but visually it's different.

There was also discussion (easier to keep track here) regarding include of files:

```yaml
includes:
  - ./vars.yaml # could contain both vars and tasks sections

tasks:
  terminal: "$ctr | terminal"

  go:test:
    shell: ./go-test # meaning the go:test task is defined by the file 'go-test' that is a standalone dagger shell file
```

I guess in shell we could have something similar:

```shell
#!/usr/bin/env dagger

.include "vars" # vars is a dagger shell file

terminal() {
  ctr | terminal
}

go_test() {
  .include "go-test" # go-test is a dagger shell file that will be run if go_test is called
}
```

Issue #10737, opened 2025-07-16
✨ Import image from local container runtime


jedevc:
Essentially, the opposite of https://github.com/dagger/dagger/issues/8025.

We should be able to import images from the local container runtime. Probably under the `Host` API, so something like `dag.Host().Container("<imagename>")`. That matches up with how we load files/directories already.

This way, it would be possible to do something like:

```
$ docker image ls
REPOSITORY                    TAG        IMAGE ID       CREATED          SIZE
my-img                        latest     d4a4cfe3b1dd   2 minutes ago    23.4MB
$ dagger shell
> host | container "my-img" | with-exec echo,"executed in the context of this host image!"
```

This currently isn't possible (without weird `docker save` + `Container.import` hackery)

Related:
- https://github.com/dagger/dagger/issues/8677
- https://github.com/dagger/dagger/issues/5235

jedevc:
Also realized from https://discord.com/channels/707636530424053791/1397155347164041276/1397155652895379516, that we don't support loading from a tarball 🤔 

We should support this for sure - something like `dag.Container().fromTarball()` feels reasonable.

cschar:
this would be great.. i'm googling this and finding its being worked on... right now!

Issue #8483, opened 2024-09-17
Default value for `Container` arguments


shykes:
## Problem

When my function receives a container as argument, I can't set a default value for it. The way the CLI accepts a string value for container arguments, and interprets it as a remote `Container.from()` pipeline, could be used in the same way here. But it's not.

I would expect this to work:

```golang
func Foo(
  // +optional
  // +default="alpine:latest"
  c *dagger.Container,
)
```

But it doesn't.

## Solution

Make the above example work :)

Transmitt0r:
Same for `Directory` in 0.13.0 Go SDK

marcosnils:
> Same for `Directory` in 0.13.0 Go SDK

 You can do this with directories by following using another `defaultPath` argument documented here: https://docs.dagger.io/manuals/developer/functions/#directories-and-files

tiborvass:
👋 I can tackle this one.

Transmitt0r:
> > Same for `Directory` in 0.13.0 Go SDK
> 
> 
> 
>  You can do this with directories by following using another `defaultPath` argument documented here: https://docs.dagger.io/manuals/developer/functions/#directories-and-files

Thanks for the link, was confused by https://docs.dagger.io/manuals/developer/entrypoint-function#default-values-for-complex-types

shykes:
@tiborvass do you still want to do this one? No problem if not, I'll just unassign and queue it up

jasonmccallister:
Would be nice to also set a default value for a `*dagger.Service` 🙏

jedevc:
x-linking #10770, which should be the way to unify all of this.

PR #10753, opened 2025-07-18
Host resources

shykes:
A new feature in the engine API: load resources from the host using a single standardized address.

This is a feature that already exists in the CLI, but we're moving in the engine so that it can be used from other places. Specifically a persisted config file to pre-configure default values. Most especially default secrets.


shykes:
Open question: are these host resources, or just... *resources*?

vito:
> Open question: are these host resources, or just... _resources_?

'Host resources' makes sense to me since they're always resolved relative to the host - e.g. when you pass `localhost:1234` as a service arg, `./foo` as a directory/file, etc.

It also makes sense to gatekeep them a bit, e.g. modules currently can't access the `Host` API, and that feels like what we'd want here?

sipsma:
> Open question: are these host resources, or just... resources?

`host` doesn't make sense to me in this example: 
```
host | resource https://github.com/dagger/dagger | as-git-repository
```

Since it's a remote git repo.

I'd personally vote for just dropping the leading `host` and adding a new top-level, e.g. 
```
resource https://github.com/dagger/dagger | as-git-repository
```

But not highly opinionated there.

---

Also seems like we'll *might* want to add an enum alongside each of the `as*` fields that indicates which type it actually is. That way clients can do something like:
```go
resource := dag.Resource("env://FOO")
kind, err := resource.Kind(ctx) // returns an enum indicating which kind of resource it is
if err != nil { // ... }
switch kind {
case dagger.SecretResourceKind:
   secret := resource.AsSecret()
   // do stuff with it ...
case dagger.Container:
   ctr := resource.AsContainer()
   // do stuff with it ...
}
```

That's what we've done for other similar parts of the API (like ModuleSource). Otherwise you are forced to try each `as*` field and see if it returns an error; that'd be weird, would create confusing telemetry, not be especially performant, etc. etc.

However, that does bring up the related question of whether it's always going to be a one-to-one mapping of a given URI to its valid "kind" (container, secret, etc.) or if it would every make sense for multiple to be possible.

e.g. I suppose since we support turning a local dir into a `GitRepository` both of these might be allowed:
1. `resource ./some/local/git/dir | as-directory`
2. `resource ./some/local/git/dir | as-git-repository`

In which case the enum idea might need some adjustment 🤔 Just food for thought at this point

shykes:
@sipsma yeah in this case it's specifically *not* going to be a one-to-one mapping IMO. This design is an attempt at making the API more readable and concise - but conceptually it could also be done like this:

```
loadDirectory(address: String!): Directory!
loadSecret(address: String!): Secret!
loadFile(address: String!): File!
loadGitRepository(address: String!): GitRepository!
```

And so on.

shykes:
@sipsma re: dropping the "host": yeah I was wondering the same thing.

- On the one hand: as you said, calling it "host" is pushing it in some cases...
- But on the other hand: if we remove "host", there are some cases where the address is interpreted in a very host-specific way (`env://FOO`, `op://FOO`, `~/foo`, `cmd:foo` etc) so it would feel like an exception to `host()`...

jedevc:
Just to throw another design point into the mix - if we put it behind `host` it definitely won't be accessible to modules. Do we want it exposed to modules? If so, we probably need to put it somewhere else. I'm voting no though, but just raising depending on the vision.

core/schema/host.go
```diff
@@ -243,6 +272,499 @@ type hostDirectoryArgs struct {
...
+			{
+				Name:  "cacheKey",
+				Value: dagql.Opt(dagql.String(cacheKey)),
+			},
+		},
+	})
+	return inst, err
+}
+
+func (s *hostSchema) resourceAsService(
```
jedevc, line 687: Thoughts on a future `resourceAsContainer`? That would be used for parsing the CLI args for a remote image. Would also be used to use it as default values for a `Container` arg, etc, see https://github.com/dagger/dagger/issues/8483.

Reason I'm raising, is because in the context of https://github.com/dagger/dagger/issues/10737, there's two potential places to load from: the host, or a remote registry. Just wondering what the API would look like here :thinking:

shykes: This PR already defines `resourceAsContainer`, is that what you mean? https://github.com/dagger/dagger/pull/10753/files/2cc8e15e087cb659a3f919892587a15206927976#diff-09f798f10567cb9017b829bfa481df0a90e444368d36bd8f7025a69d1d2521c9R486

shykes:
> Just to throw another design point into the mix - if we put it behind `host` it definitely won't be accessible to modules. Do we want it exposed to modules? If so, we probably need to put it somewhere else. I'm voting no though, but just raising depending on the vision.

Good point. No strong opinion about hiding it from modules. I think, given that runtime access is correctly sandboxed, there's no strong requirement to hide this API from modules. It boils down to making sure devs are not confused by the API.

core/schema/host.go
jedevc: Ah, it's at the top :woman_facepalming: Yeah that!

Just kinda wondering how this would work if a host container store was connected, it should probably be able to look there as well?

jedevc:
Forgot to comment this: very nice, I like it :tada:

shykes:
Closing in favor of #10770

PR #10761, opened 2025-07-18
disable dag-op edge merges

sipsma:
Edge merging is a constant thorn and may explain some remaining weirdness we are seeing with sessions disappearing and related errors.

Disabling it is easy, but we have a few gaps to fill to get our own content-based cache logic fleshed out enough to replace what we previously got via edge merges. Thankfully this is mostly a matter of filling in more dagql cache key funcs so far, so working through that.


sipsma:
cc @jedevc got this working now w/out any performance regressions, at least based on test runtime. Needs cleanup but curious if this looks roughly along the same lines as what you were imagining in terms of fixing this.

core/schema/directory.go
```diff
@@ -216,8 +216,43 @@ type subdirectoryArgs struct {
...
+	if err != nil {
+		return res, err
+	}
+
+	res, err = dagql.NewObjectResultForCurrentID(ctx, srv, dir)
+	if err != nil {
+		return res, err
+	}
+	if parent.Self().IsContentHashed {
+		res, err = core.MakeDirectoryContentHashed(ctx, bk, res)
```
jedevc: Mildly confused by this logic - the idea is that we keep a directory content-hashed upon selections? Not sure I understand what this is doing in terms of the behavior -- is it making a certain test pass?

Also would be nice if `MakeDirectoryContentHashed` set `IsContentHashed`.

core/integration/container_test.go
```diff
@@ -4404,9 +4404,11 @@ func (ContainerSuite) TestNestedExec(ctx context.Context, t *testctx.T) {
 		output1a := runCtrs(c1, hostDir1, subdirA)
 		// run an exec that has /tmpdir/b/f included
 		output1b := runCtrs(c1, hostDir1, subdirB)
+
+		// TODO: fix comment
 		// sanity check: those should be different execs, *not* cached
 		// (this happens because content-hashing doesn't occur on the root dir)
-		require.NotEqual(t, output1a, output1b)
+		require.Equal(t, output1a, output1b)
```
jedevc, line 4015: I'm guessing these are getting cached at the dagql layer now? Not at the buildkit layer?

But yes, this seems very reasonable :tada:

core/schema/directory.go
sipsma: Yes, when disabling edge merges only with no other changes, `TestContainer/TestNestedExec` fails because the execs aren't cached when expected. But with this change, it passes again.

`IsContentHashed` may or may not stick around, it was just the quick way to try this out. But agree if that or similar sticks I'll try to make it more seamless like that.

jedevc: Right, yeahhh. Makes sense :tada:

I'm also chill with totally changing the behavior of the test if we want the caching to be different here. Potentially also relevant is this logic here: https://github.com/dagger/dagger/blob/e035126279e30913831033156f0b9d544873f896/core/dagop.go#L476-L479

sipsma: > I'm also chill with totally changing the behavior of the test if we want the caching to be different here.

The places the test expects caching to occur seem reasonable to me. Actually the only changes to the test I'd want are places where it currently doesn't expect cache hits even though it would make sense (comments about caching not working because its on the root mount). So I'd really want to keep the current level of caching (and possibly improve it further) rather than regress anything there.

> Potentially also relevant is this logic here: 

I played around with this first actually, but it required enabling buildkit content hashing on the root mount and that very very noticeably destroyed performance in our integ tests. The approach in this PR is the second attempt that seemed to work w/out performance being hit.

core/integration/container_test.go
sipsma: Yes exactly, the caching actually improves in cases like this now 🚀

core/schema/directory.go
jedevc: Wonderful :tada: Yeah, the approach all seems good then.

Happy to approve once out of draft :grin:

sipsma:
Quick update, I have been working on this and slightly expanding the scope to include making all mounts+rootfs of containers to be persisted in the `Container` struct as `dagql.ObjectResult[*Directory]`. This will clean-up the approach quite a bit since we can have IDs and thus cache keys associated with each mount+rootfs. It's also a pretty nice change to have generally since it connects our API objects and de-associates Container with most LLB-specific logic that still exists in it.

sipsma:
Hitting some long-standing bugs as I churn through all this, noting so I remember to add backfill test coverage

---

without-file on mounted file doesn't work
```
f=$(file "f" "foo")
ctr=$(container | from alpine:3.20 | with-mounted-file /f $f)
▶ $ctr | without-file /f | file /f | contents 0.0s
foo
```

---

with-file on mount point doesn't work (same for with-files, etc.)
```
● d=$(directory | with-new-file f1 foo1 | with-new-file f2 foo2) 0.0s

● fmnt=$(file f3 foo3) 0.0s

● ctr=$(container | from alpine:3.20 | with-mounted-directory /mnt $d | with-mounted-file /mnt/f2 $fmnt) 0.1s

▶ $ctr | file /mnt/f1 | contents 0.7s
foo1                                                                                                                                                                                                                                           

▶ $ctr | file /mnt/f2 | contents 0.0s
foo3                                                                                                                                                                                                                                           

● fmnt2=$(file f4 foo4) 0.0s

▶ $ctr | with-file /mnt/f2 $fmnt2 | file /mnt/f2 | contents 0.1s
foo3
```

core/dagop.go
```diff
@@ -344,29 +343,23 @@ func NewContainerDagOp(
...
 		return nil, err
 	}
 	inputs = append(inputs, extraInputs...)
 
-	// NB: strip out the buildkit inputs! this is so that we rely entirely on
-	// the *buildkit* inputs.
-	ctrDigest, err := DigestOf(ctr.WithoutInputs())
-	if err != nil {
-		return nil, err
-	}
```
jedevc, line 358: So I remember discussing this - the idea here is that the dagql ID becomes the ultimate authoritative source right?

I guess a couple questions:
- What happens to buildkit slow content-caching? e.g. if I mount a directory with a *completely* different dagql ID, but with exactly the same contents, buildkit will no longer cache this - right?
  - I think that's what we want, right? If we truly truly want the content caching, that's where we add in `MakeDirectoryContentHashed` and support for multiple cache keys, etc. It gives us significantly more control over what gets content hashed, so should actually be a decent perf improvement... (at least theoretically)
- Is there a reason that we shouldn't *entirely* remove this technique of `DigestOf(x.WithoutInputs)` and all the messing around with `extractLLBDependencies`?
  - Ideally we would consolidate `FSDagOp` and `RawDagOp` around the ID as well.
  - Is the approach we went for in https://github.com/dagger/dagger/pull/10880 kind of odds with this idea?

sipsma: > * What happens to buildkit slow content-caching? e.g. if I mount a directory with a _completely_ different dagql ID, but with exactly the same contents, buildkit will no longer cache this - right?
>       
>    * I think that's what we want, right? If we truly truly want the content caching, that's where we add in `MakeDirectoryContentHashed` and support for multiple cache keys, etc. It gives us significantly more control over what gets content hashed, so should actually be a decent perf improvement... (at least theoretically)

Yep exactly

> * Is there a reason that we shouldn't _entirely_ remove this technique of `DigestOf(x.WithoutInputs)` and all the messing around with `extractLLBDependencies`?
>       
>    * Ideally we would consolidate `FSDagOp` and `RawDagOp` around the ID as well.
>    * Is the approach we went for in [chore: CacheMap for directory dagops #10880](https://github.com/dagger/dagger/pull/10880) kind of odds with this idea?

I think we should move away from that, but I had reached the point of just wanting to get this first batch of updates out rather than keep rebasing. But I do agree we should simplify all that stuff further on top of the changes here, just would prefer it's in a follow-up so we can get the flakes this PR fixes out of the way

core/dagop.go
```diff
@@ -825,5 +889,6 @@ func newDagOpLLB(ctx context.Context, dagOp buildkit.CustomOp, id *call.ID, inpu
 		llb.WithCustomNamef("%s %s", dagOp.Name(), id.Name()),
 		buildkit.WithTracePropagation(ctx),
 		buildkit.WithPassthrough(),
+		llb.SkipEdgeMerge,
```
jedevc, line 892: :tada: begone foul demon :tada:

engine/buildkit/linux_namespace.go
```diff
@@ -19,7 +19,7 @@ import (
 
 const (
 	idleTimeout    = 1 * time.Second
-	workerPoolSize = 5
+	workerPoolSize = 2
```
jedevc: :eyes: what's the logic here?

engine/sources/local/source.go
```diff
@@ -173,7 +173,7 @@ func (ls *localSourceHandler) CacheKey(ctx context.Context, g session.Group, ind
 		return "", "", nil, false, err
 	}
 	digestString := digest.FromBytes(dt).String()
-	return "session:" + ls.src.Name + ":" + digestString, digestString, nil, true, nil
+	return ls.src.Name + ":" + digestString, digestString, nil, true, nil
```
jedevc, line 176: :thinking: hm, because we're including `SessionID` here, we really haven't needed this for a while... right?

Just checking my understanding!

engine/buildkit/linux_namespace.go
sipsma: Oh this was part of a side-quest looking into perf bottlenecks when it seemed like this change might have been slowing things down. Was trying to reduce the number of goroutines but turned out that the main issue there was [introspection queries](https://github.com/dagger/dagger/pull/10967), so will rm this change.

engine/sources/local/source.go
sipsma: Yeah I changed this while trying to figure out some unexpected cache misses. Didn't end up being load bearing but will leave the change in since like you said it's not needed anymore.

PR #10765, opened 2025-07-19
prompt mode: fix history being dropped on interrupt

vito:
when an agent starts going off the rails it's natural to interrupt it and course-correct. but right now that actually completely undoes the prompt and whatever progress it made.

this mistake was being made at multiple layers:

* we were Fork()ing at each prompt, even though we don't even support rewinding in the UI yet
* the $agent var handling wasn't being respected everywhere, so if WithPrompt didn't succeed it would end up resetting back to the old value in $agent

---

* before: https://dagger.cloud/vito/traces/e94f1c558ef794bc1f03d776b52eb56a
* after: https://dagger.cloud/vito/traces/953db6a3d3719748be4b73851d337293

### DIFF
--- a/.changes/unreleased/Fixed-20250721-111016.yaml
+++ b/.changes/unreleased/Fixed-20250721-111016.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Prompt mode now handles interrupts more gracefully, continuing from where the agent left off rather than completely forgetting the prompt.
+time: 2025-07-21T11:10:16.055120258-04:00
+custom:
+    Author: vito
+    PR: "10765"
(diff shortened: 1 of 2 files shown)


shykes (approved):

cmd/dagger/llm.go
```diff
@@ -81,27 +81,27 @@ func NewLLMSession(ctx context.Context, dag *dagger.Client, llmModel string, she
...
-		WithSystemPrompt(`When the user's query contains a variable like $foo, determine if the request is asking you to save a value. If so, declare the output binding.`)
+		WithSystemPrompt(`When the user's query contains a variable like $foo, determine if the request is asking you to save a value. If so, declare the output binding.`))
 }
 
 func (s *LLMSession) Fork() *LLMSession {
-	cp := *s
-	cp.undo = s
-	return &cp
+	// FIXME: this was a half-baked feature, currently does more harm than good
+	// because we lose partial progress on interrupt
```
jedevc, line 108: may be worth linking to this pr?

vito:
Pushed an additional fix for LLM initialization. For a while now swapping over to prompt mode would have an empty prompt, instead of properly initializing and show you what model it's using. Now that's fixed, and we log an error if it fails, instead of cramming the error into the prompt.

PR #10766, opened 2025-07-19
fix a couple of `TestTelemetry` flakes

vito:
* error origins could sometimes not be wired up depending on span receiving order
* engine version scrubber choked on the `v` in `-dev`


sipsma (approved):

vito:
Hmm, still need a bit work on the first fix - now a bunch of other tests are doing the error revealing logic, and I don't want to just `--update` them because we'd be losing coverage for e.g. pending states.

Probably need to add a flag of some sort to toggle revealing on a test-by-test basis.

PR #10770, opened 2025-07-21
address(): a unified address to load containers, directories, secrets…

shykes:
*This is an alternative to #10753*

Extract CLI logic for loading different object types from a standard address; and expose it in the engine API. This is a requirement for #10697, which is part of #10370.

## Schema

```graphql
extend type Query {
  """
  initialize an address to load directories, containers, secrets or other object types.
  """
  address(value: String!): Address!
}

"""
A standardized address to load containers, directories, secrets, and other
object types. Address format depends on the type, and is validated at type
selection.
"""
type Address {
  """Address value"""
  value: String!

  """Load a container from the address."""
  container: Container!

  """Load a directory from the address."""
  directory(exclude: [String!] = [], include: [String!] = [], noCache: Boolean = false): Directory!

  """Load a file from the address."""
  file(exclude: [String!] = [], include: [String!] = [], noCache: Boolean = false): File!

  """Load a git ref (branch, tag or commit) from the address."""
  gitRef: GitRef!

  """Load a git repository from the address."""
  gitRepository: GitRepository!

  """A unique identifier for this Address."""
  id: AddressID!

  """Load a secret from the address."""
  secret: Secret!

  """Load a service from the address."""
  service: Service!
}

"""
The `AddressID` scalar type represents an identifier for an object of type Address.
"""
scalar AddressID

extend type Container {
  originalAddress: Address!
}

extend type Directory {
  originalAddress: Address!
}

extend type File {
  originalAddress: Address!
}

extend type Service {
  originalAddress: Address!
}

extend type Secret {
  originalAddress: Address!
}

extend type GitRef {
  originalAddress: Address!
}

extend type GitRepository {
  originalAddress: Address!
}
```

## Examples

- `address https://github.com/dagger/dagger | git-repository`
- `address env://MYTOKEN | secret`
- `address ~/.zshrc | file`
- `address tcp://localhost:4242 | service`
- `address https://github.com/dagger/dagger#main | git-ref`
- `address https://github.com/dagger/dagger#main:docs | directory`


vito:
A bit surprised at dropping the `Host` here; that made a lot of sense to me before (https://github.com/dagger/dagger/pull/10753#issuecomment-3090214212) from a design and security standpoint.

What's to prevent arbitrary modules from being able to do e.g. `address("env://GITHUB_TOKEN").plaintext`?

And from an API standpoint, how does that differ from `secret("env://GITHUB_TOKEN").plaintext`?

Even the case of `host.resource("https://example.com/git/repo")` _could_ make sense to me if you think of it like routing the network _through_ the host, i.e. if it's pointing to a private network that the host can reach but the engine can't. (It's a bit of a stretch, though, sure.)

shykes:
> A bit surprised at dropping the `Host` here; that made a lot of sense to me before ([#10753 (comment)](https://github.com/dagger/dagger/pull/10753#issuecomment-3090214212)) from a design and security standpoint.
> 
> What's to prevent arbitrary modules from being able to do e.g. `address("env://GITHUB_TOKEN").plaintext`?

From a security POV, it's all sandboxed anyway. So `address("env://GITHUB_TOKEN")` has the same result as `secret("env://GITHUB_TOKEN")`. So it's purely a cosmetic choice whether to place it under `Host`.

`Host` is just a rough bucket of API calls that are *obviously* and *exclusively* executed in the context of the caller host. But the bucket is already leaky.

- `Directory.export`, `Container.export`, `File.export`: write to the caller/host filesystem, but not under `Host`
- `secret()`: load secrets from the caller/host, but not under `Host`
- `moduleSource()`: when given a local path, load the module from the caller/host 's filesystem, but not under `Host`
- `git()`: load auth configuration from the caller/host's system
- `container.From`, `container.dockerBuild`: load docker auth configuration from the caller/host's system

IMO there are so many exceptions already, that there's no point in hiding from module devs that some of their API calls reach into their own context, and never into the top-level caller's context (except for LLMs). That ship has sailed. As a corollary, I think the only reason to keep `Host` hidden, is that it has a confusing name. As soon as we agree on a less confusing name, we should un-hide it.

> And from an API standpoint, how does that differ from `secret("env://GITHUB_TOKEN").plaintext`?

It's a convenience wrapper. The logic for the wrapper already existed: it's currently in `cmd/dagger/flags.go`, so that's where you'll find the answer to your question.

For some types, the wrapper is very thin. For others, it's much thicker.

In the specific case of secrets:

- The wrapper handles backwards compatibility with older secret formats not supported by `secret()`
- The wrapper parses an optional `#cacheKey=xxx` from the address, and passes it as a separate argument to `secret()`

The general idea is: `address()` guarantees a single string address, with no need to pass other arguments.

> Even the case of `host.resource("https://example.com/git/repo")` _could_ make sense to me if you think of it like routing the network _through_ the host, i.e. if it's pointing to a private network that the host can reach but the engine can't. (It's a bit of a stretch, though, sure.)

To reiterate and summarize my earlier point: I think it's hopeless to split the Dagger API between "parts that always interact with the caller/host" and "parts that never do". It's hopelessly entangled and I think we should embrace that.

vito:
@shykes Ahh, I see now - it's just _always_ sandboxed (if there is one), so the CLI would use these on the outside to gain access to values from the host, and modules can also use it, but they will just be accessing their own sandbox.

From how we got here, I assumed the intent was still for these to go through the host.

shykes:
Could anyone help me figure out why CI is still red? Is it a flake?

shykes:
Update: not a flake. I definitely broke some tests.. Fixing.

shykes:
This test fail on Elixir SDK is mysterious: https://dagger.cloud/dagger/traces/6ff475644f5c28f16642b8697a060cd0?span=fb5064e84b407ea4

Do we have a known flake there?

shykes:
Update on failing tests:

- Security scan issue was fixed in a separate PR, thank you @jedevc 🙏 
- Updated TUI viz test data (thank you @jedevc for the tip: `test | telemtry --update | export .`). It looks less good as before, but we can address that in a follow-up if @vito deems it worth pursuing.
- Elixir SDK test: that's from main. Will need to be addressed separately. Ignoring.

vito:
> Updated TUI viz test data (thank you @jedevc for the tip: `test | telemtry --update | export .`). It looks less good as before, but we can address that in a follow-up if @vito deems it worth pursuing.

Yeah seems worth it. I can try to sneak an improvement in this PR if we can agree on how it should look. We can just have the UI special-case IDs that match that structure:

```graphql
address(value: "alpine").container: Container!
address(value: "./foo/bar").directory: Directory!
```

Some options:

1. `Container(alpine)`, `Directory(./foo/bar)`
2. `Container("alpine")`, `Directory("./foo/bar")`
3. `address(value: "alpine").container: Container!`
    * Just unpack one extra layer, no new syntax.
    * This could MAYBE be a general heuristic? For any call whose field has no arguments but receiver does, reveal the receiver, since that's what distinguishes from other values. This also came up before with e.g. `Binding.asContainer` - _what binding_?
4. Other options?

We'll want to do the same in Cloud for consistency - shouldn't be hard.

shykes:
I am stuck on this test failure: https://dagger.cloud/dagger/traces/9b496c12795ef4a512ba1d4f73d4e823?span=cade61194ee7bac3

```
40  : [2.9s] | Error: failed to serve module: input: moduleSource.asModule failed to call module "test" to get functions: process "/runtime" did not complete successfully: exit code: 2
40  : [2.9s] | Stdout:
40  : [2.9s] | failed to return error: get or init client: initialize client: failed to load module: load moduleSource(refString: ".").asModule: Module!: load: load base: failed to stat local path: failed to stat path: failed to get requester session: session for "vqnqq9d0veiihso5fdtukuytn" not found 
40  : [2.9s] | original error: get parent name: get or init client: initialize client: failed to load module: load moduleSource(refString: ".").asModule: Module!: load: load base: failed to stat local path: failed to stat path: failed to get requester session: session for "vqnqq9d0veiihso5fdtukuytn" not found
40  : [2.9s] | Stderr:
40  : [2.9s] | unexpected status 200: get or init client: initialize client: failed to load module: load moduleSource(refString: ".").asModule: Module!: load: load base: failed to stat local path: failed to stat path: failed to get requester session: session for "vqnqq9d0veiihso5fdtukuytn" not found
```

shykes:
> I can try to sneak an improvement in this PR if we can agree on how it should look. We can just have the UI special-case IDs that match that structure:

@vito I forgot an important point: there's a planned part 2 for this PR, which is specifically for display. My idea:

- If an object was created from an `Address`, that information should be carried in a field, perhaps called `originalAddress`.
- Then I can change the CLI introspection logic, to print `originalAddress` instead of the raw ID in default values.
- Bonus: add a field to *every* object which converts their ID to a human-readable shell one-liner :) But that's maybe a part 3.

shykes:
> > I can try to sneak an improvement in this PR if we can agree on how it should look. We can just have the UI special-case IDs that match that structure:
> 
> @vito I forgot an important point: there's a planned part 2 for this PR, which is specifically for display. My idea:
> 
> * If an object was created from an `Address`, that information should be carried in a field, perhaps called `originalAddress`.
> * Then I can change the CLI introspection logic, to print `originalAddress` instead of the raw ID in default values.
> * Bonus: add a field to _every_ object which converts their ID to a human-readable shell one-liner :) But that's maybe a part 3.

Done. See `extend type XXX { originalAddress }`

shykes:
Picking this back up.

shykes:
Still stuck on this test: https://dagger.cloud/dagger/traces/0d066611ff8a319ebb878441d56502cb?span=970b5d6849d26a79

shykes:
@sipsma just in case you have an idea what this could be... 🙏  ☝️ 

The failing test seems to be [this one](https://github.com/shykes/dagger/blob/address-api/core/integration/cross_session_test.go#L1376-L1407).

sipsma:
> @sipsma just in case you have an idea what this could be... 🙏 ☝️
> 
> The failing test seems to be [this one](https://github.com/shykes/dagger/blob/address-api/core/integration/cross_session_test.go#L1376-L1407).

@shykes pushed a quick fix, the `Address.secret` API just needed to honor the cacheKey value when set: https://github.com/dagger/dagger/pull/10770/commits/bd5304aa7cba99863f09777c14a067800be3eb37

shykes:
> > @sipsma just in case you have an idea what this could be... 🙏 ☝️
> 
> > 
> 
> > The failing test seems to be [this one](https://github.com/shykes/dagger/blob/address-api/core/integration/cross_session_test.go#L1376-L1407).
> 
> 
> 
> @shykes pushed a quick fix, the `Address.secret` API just needed to honor the cacheKey value when set: https://github.com/dagger/dagger/pull/10770/commits/bd5304aa7cba99863f09777c14a067800be3eb37

Thank you!

shykes:
@sipsma sorry to bother you again, I'm still seeing a very similar set of failing tests. Not sure if they were failing all along, or just popped up now?

https://dagger.cloud/dagger/traces/0d066611ff8a319ebb878441d56502cb?span=970b5d6849d26a79

core/schema/container.go
```diff
@@ -648,6 +648,9 @@ func (s *containerSchema) Install(srv *dagql.Server) {
 		dagql.Func("imageRef", s.imageRef).
 			Doc(`The unique image reference which can only be retrieved immediately after the 'Container.From' call.`),
 
+		dagql.Func("originalAddress", s.originalAddress).
+			Doc(`The address this container was originally loaded from, if any`),
```
jedevc: A bit curious on this API - what's the use case?

I'm a little bit worried about it being public and accessible to modules, e.g. this would let a module see where on the host a Directory was loaded from for example.

core/schema/address.go
```diff
@@ -0,0 +1,590 @@
...
+	srv, err := core.CurrentDagqlServer(ctx)
+	if err != nil {
+		return inst, err
+	}
+	if err := srv.Select(ctx, srv.Root(), &inst, q...); err != nil {
+		return inst, err
+	}
+	f := inst.Self().Clone()
+	f.OriginalAddress = r.Self()
+	return dagql.NewObjectResultForCurrentID(ctx, srv, f)
```
jedevc: I *think* we should just be directly returning `inst` here.

Otherwise, the `Directory` loaded using `address` is going to have a different ID digest from the one returned from `Host.directory` - and we suddenly won't be caching them interchangeably (since they'll have different IDs).

If we just return `inst` directly, then that object has the right ID, and so will cache correctly. cc @sipsma who's been doing more ID shenanigans.

> [!NOTE]
>
> This means we wouldn't be able to save `OriginalAddress` on the container. That's kind of the main reason I was questioning the API and if it's necessary, since it pokes a hole in caching.

This does feel *very* similar to the problem I'm trying to solve in #10895, of "origin tracing" - essentially, I want to solve this by analyzing the IDs themselves, instead of needing to save anything directly on the object.

Just kind of spitballing here, I'm not actually sure what kind of impact this would have on the cache off the top of my head, but it does feel like something worth getting right.

core/schema/container.go
shykes: My use case is #10697. I need a way to introspect default argument values and print them to the user in a human-readable way, even when they're objects loaded from `.env.dagger.json`. That use case is why I developed an address API in the first place.

This pattern seemed natural to me, since it generalizes `Container.imageRef`.

Your concern makes sense, I didn't think of the risk of leaking addresses to third-party modules.

Maybe instead of exposing the address in the object itself, I can expose it in its typedef? That way, introspecting clients get a human-readable default; but modules receiving the object can't get to its address.

core/schema/address.go
shykes: If I return `inst` I can't set `originalAddress`

core/schema/container.go
jedevc: Hm, if the default args are *going* to be on the typedef, then yeah, that feels like the better place.

I remember we discussed this before, and I didn't *love* the idea of modifying the typedefs from user-space, but I think we decided to go for it (?), so that we avoided having multiple different layers of introspection API.

I think that's how the default args set in `.env.dagger.json` would work? So, if we can, keeping the info on the typedef feels much cleaner to me.

---

If we weren't to do that, how would we actually get the value of the container? Would we have to load an ID and evaluate it? Ideally, we'd avoid that, since that would actually require downloading + evaluating the resource.

e.g. if the git repository we're trying to load doesn't exist (or doesn't have authentication), `originalAddress` would fail - which does feel confusing, especially, when trying to present that info to users in help.

Not sure how to solve that with the current approach.

jedevc: My potential suggestion for resolving this is to:
- Remove `originalAddress` (in this PR)
- When object ID are loaded and set as defaults (and attached to typedefs, in a future PR):
  - Traverse the ID, ascending the `Receiver` property until we reach a top-level query, that's called `address`
  - Extract the arg from that `address`, and attach it to the typedef

Then, we keep all of the logic on the typedefs, which feels reasonable to me.

shykes: > I remember we discussed this before, and I didn't *love* the idea of modifying the typedefs from user-space, but I think we decided to go for it (?), so that we avoided having multiple different layers of introspection API.

Yes that's right.

> I think that's how the default args set in `.env.dagger.json` would work? So, if we can, keeping the info on the typedef feels much cleaner to me.

I wonder what's the best field in the typedef API to attach it to?

Since it's for human-friendly rendering of a value, maybe it's a new field?

> If we weren't to do that, how would we actually get the value of the container? Would we have to load an ID and evaluate it? 

Not sure what you mean. The way the "default secrets" POC works today, it reads values from the env file, then converts them to objects with eg `address().Container`

Then the default arg value is changed to be the object ID, converted to a string.

That 👆 all works, I only need to fix human-friendly display of the object ID.

> Ideally, we'd avoid that, since that would actually require downloading + evaluating the resource.

Ah i think i get it. But wouldn't that "load by ID" be lazy?

> e.g. if the git repository we're trying to load doesn't exist (or doesn't have authentication), `originalAddress` would fail - which does feel confusing, especially, when trying to present that info to users in help.

Yes indeed. Maybe ID loading is not lazy enough to avoid 100% if these situations?

> 
> 
> 
> Not sure how to solve that with the current approach.

jedevc: > Ah i think i get it. But wouldn't that "load by ID" be lazy?

*loading* the object by ID is suitably lazy, but the main issue comes when you try and evaluate `originalAddress`, you then need to call the entire history.

So if the chain is:
- `address()`
- `.container()`
- `.withExec()`
- `originalAddress()`

Then to evaluate `originalAddress`, we need to evaluate `withExec`, which needs to evaluate `container`, which needs to evaluate `address`.

There's currently no way to evaluate a field without actually calling all the parent fields (wondering if we'd maybe want that, it could be useful). But even if we had it, I'd argue it's still a bit of a weird usage - evaluating anything in the *middle* of the chain could be an error, but evaluating the end succeeds? Straaaange.

jedevc: > Since it's for human-friendly rendering of a value, maybe it's a new field?

Yeah! Something like that, maybe `DefaultAddress`?

core/schema/address.go
shykes: Removed `originalAddress` as discussed, problem solved

core/schema/container.go
shykes: I removed `orignalAddress`. I will figure out in a separate PR how to connect this to typedefs.

core/integration/address_test.go
```diff
@@ -1,13 +1,21 @@
...
 
-func TestSecretParse(t *testing.T) {
-	t.Parallel()
+type AddressSuite struct{}
+
+func TestAddress(t *testing.T) {
+	testctx.New(t, Middleware()...).RunTests(AddressSuite{})
+}
+
+func (AddressSuite) TestSecretParse(ctx context.Context, t *testctx.T) {
```
jedevc: Can we have tests for all the types that this can produce? Just don't want these to unexpectedly break later.

core/schema/address.go
```diff
@@ -0,0 +1,595 @@
...
+				Name:  "exclude",
+				Value: dagql.ArrayInput[dagql.String](dagql.NewStringArray(args.Exclude...)),
+			})
+		}
+		if len(args.Include) > 0 {
+			q[1].Args = append(q[1].Args, dagql.NamedInput{
+				Name:  "include",
+				Value: dagql.ArrayInput[dagql.String](dagql.NewStringArray(args.Include...)),
+			})
+		}
```
jedevc: This is quite hard to read. This form is a bit easier: https://github.com/dagger/dagger/blob/a86457436a24edf8d2f9de0e30bc3b5a91af28ee/core/modfunc.go#L686-L693

core/schema/address.go
```diff
@@ -0,0 +1,595 @@
...
+}
+
+// Convert an address to an absolute local path:
+// - file:// is stripped if needed
+func getLocalPath(path string) string {
+	// allow `file://` scheme or no scheme
+	path = strings.TrimPrefix(path, "file://")
+	// make windows paths usable in the Linux engine container
+	// FIXME: still needed engine-side?
+	path = filepath.ToSlash(path)
```
jedevc: Probably worth checking this, does anyone have a windows machine to test on?

core/schema/address.go
```diff
@@ -0,0 +1,595 @@
...
+	ctx context.Context,
+	r dagql.ObjectResult[*core.Address],
+	args struct{},
+) (
+	inst dagql.ObjectResult[*core.Container],
+	err error,
+) {
+	var (
+		addr = r.Self().Value
+	)
```
jedevc: This is a bit longer than it needs to be. `addr := r.Self().Value` is a bit neater.

(also applies in a few other places in this patch)

core/schema/address.go
```diff
@@ -0,0 +1,595 @@
...
+	if err != nil {
+		return inst, err
+	}
+	err = srv.Select(ctx, srv.Root(), &inst, q...)
+	if err != nil {
+		return inst, err
+	}
+	// FIXME: do we still need to add this result in the secret store,
+	// because it doesn't have the same ID?
+	return inst, nil
```
jedevc, line 438: Resolved! It now has the same ID.

core/schema/address.go
```diff
@@ -0,0 +1,595 @@
...
+			return inst, err
+		}
+		host = h
+		ports = append(ports, dagql.InputObject[core.PortForward]{
+			Value: core.PortForward{
+				Backend:  nPort,
+				Frontend: &nPort,
+				Protocol: core.NetworkProtocolUDP,
+			},
+		})
```
jedevc: Almost all this logic is written out twice.

core/schema/secret.go
```diff
@@ -91,7 +91,7 @@ func (s *secretSchema) secret(
 		return i, fmt.Errorf("failed to create instance: %w", err)
 	}
 
-	if args.CacheKey.Valid {
+	if args.CacheKey.Valid && args.CacheKey.Value != "" {
```
jedevc: :eyes: this shouldn't be necessary. If there is no cache key, the caller shouldn't be providing it.

core/schema/address.go
shykes: Good catch thanks. Fixed.

core/integration/address_test.go
shykes: Will do

shykes: @jedevc I added a first batch of tests, for git addresses. I added it directly to `git_test.go`, which required cleaning it up to avoid duplicated code. Bonus: removed lots of pre-existing duplicated code too :)

core/integration/git_test.go
```diff
@@ -29,172 +29,228 @@ func TestGit(t *testing.T) {
...
+}
+
+// verify ref is the sample hidden commit (from pull request)
+func requireSampleGitHiddenCommit(ctx context.Context, t *testctx.T, ref *dagger.GitRef, c *dagger.Client) {
+	requireSampleGitCheckout(ctx, t, ref.Tree(), c)
+	requireGitRefIsCommit(ctx, t, "318970484f692d7a76cfa533c5d47458631c9654", ref, c)
+}
+
+// verify repository has expected branches, tags, and commits
+func requireSampleGitRepo(ctx context.Context, t *testctx.T, repo *dagger.GitRepository, c *dagger.Client) {
```
jedevc: The `requireSample` functions probably shouldn't be in the global scope? They should probably be directly under `TestGitRefs`.

I also don't find that the newer style is really much more readable (requiring more jumping around to understand what's happening), but :shrug: up to you.

I'd have been happy with a simpler approach where we leave all the original tests the same, and simply add new tests to check that the `address` functions return the expected results - and just trust that the rest of the tests chain properly. But whatever works.

core/integration/git_test.go
```diff
@@ -29,172 +29,228 @@ func TestGit(t *testing.T) {
...
+	headBranch := repo.Head()
+	mainBranch := repo.Branch("main")
+	mainBranchLong := repo.Branch("refs/heads/main")
+	requireSampleGitBranches(ctx, t, c, headBranch, mainBranch, mainBranchLong)
+	latestTag := repo.LatestVersion()
+	requireSampleGitLatestTag(ctx, t, latestTag, c)
+	requirSampleGitTag(ctx, t, repo.Tag("v0.9.5"), c)
```
jedevc:
```suggestion
	requireSampleGitTag(ctx, t, repo.Tag("v0.9.5"), c)
```

core/schema/address.go
jedevc: I actually suspect this can be removed - as I remember, all paths on the engine should already be `ToSlash`ed.

core/schema/address.go
jedevc: The comment can be removed.

core/integration/git_test.go
shykes: The problem is that the existing git tests all start from a `GitRepository` in a monolithic way. But I `Address` can also produce `GitRef`, `Directory` and `File` (speaking only of the git-relevant parts), and that would have required a *lot* of copy-pasting.

To be fair I didn't fully take advantage of the refactor by adding those entrypoints. Let me add them, and I think the reason for doing it this way will become more clear.

shykes: > I also don't find that the newer style is really much more readable (requiring more jumping around to understand what's happening), but 🤷 up to you.

I'm pretty sure that's because you wrote it :) At least in part. It took a while to parse the intent of your tests.

core/integration/git_test.go
shykes: How did this even build? 😭 I actually ran these tests... Weird

core/integration/git_test.go
shykes: Update: I'm simplifying + expanding the tests

core/integration/address_test.go
shykes: Tests added for all types

core/integration/host_test.go
```diff
@@ -23,6 +23,26 @@ func TestHost(t *testing.T) {
...
 
+func (HostSuite) withHostDir(_ context.Context, _ *testctx.T, c *dagger.Client, path string, fn func(wd *dagger.Directory)) {
+	for _, workdir := range []*dagger.Directory{
+		c.Host().Directory(path),
+		c.Address(path).Directory(),
+		c.Address("file://" + path).Directory(),
+	} {
+		fn(workdir)
+	}
+}
```
jedevc: I like the naming here: https://github.com/dagger/dagger/blob/8ff111162057de9d4e699da3f30a0b806020d83e/core/integration/module_config_test.go#L1258-L1265

Something like `testOnHostDir`? It's not really a `with` function in the same kind of way we use `With`.

I also think we should avoid attaching things to the `HostSuite` object. It's weird to have to instantiate it.

Also, make sure to wrap using `t.Run`, so each of these is a separate subtest.

core/integration/services_test.go
```diff
@@ -2117,9 +2116,18 @@ func (ServiceSuite) TestContainerToHost(ctx context.Context, t *testctx.T) {
...
+		for _, svc := range []*dagger.Service{
+			c.Host().Service(nil),
+			c.Address("").Service(),
+			c.Address("localhost:80").Service(),
+			c.Address("tcp://").Service(),
+			c.Address("http://localhost").Service(),
+			c.Address("foo://bar").Service(),
+		} {
+			_, err := svc.ID(ctx)
+			require.Error(t, err)
```
jedevc: I know this is copied from the previous test, but we should use `requireErrContains` to also match on the error contents.

core/integration/address_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"testing"
+
+	"github.com/dagger/dagger/engine/distconsts"
+	"github.com/dagger/testctx"
+	"github.com/stretchr/testify/require"
+)
+
+type AddressSuite struct{}
+
+func TestAddress(t *testing.T) {
```
jedevc, line 19: Also make sure to add a test for `Address.value`. Don't see that tested anywhere.

It does seem strange that the tests for only some `Address` things are here - would be either worth adding some very minimal tests for those here (even if they are slight duplicates of other ones), *or* just adding a comment as to why that is.

shykes: I tried, but for example `Secret` and `Service` cannot really be tested without *actually* setting up services or secrets on the host. And `host_test.go` already has all that scaffolding...

I actually wrote out a whole test for services in `address_test.go`... Only to remove it when I realized, oh, to test anything useful I need to actually run those services, and I already did that...

core/integration/services_test.go
shykes: I think that's too specific. The important thing is that it returns an error. What the error says might be changed over time for better UX, not worth over-fitting tests for that.

core/integration/host_test.go
shykes: I tried wrapping them in `t.Run` (and wanted to call my function `runWithHostDir()`) but it messed up telemetry in ways I was unable to fix.

core/integration/address_test.go
shykes: > Also make sure to add a test for Address.value. Don't see that tested anywhere.

Done

shykes: >  would be either worth adding some very minimal tests for those here (even if they are slight duplicates of other ones), or just adding a comment as to why that is.

Done

core/integration/host_test.go
shykes: Done

shykes:
@jedevc you were right about the tests FYI.

shykes:
I've been re-running this failed tests several times, assuming it was a flake (error message seems unrelated). But it won't go away...

FYI @sipsma @alexcb @vito @jedevc @kpenfound in case you recognize the error...

shykes:
Just barely missed the release train 😭

PR #10772, opened 2025-07-21
add `Env.withCaller`

vito:
A sneaky workaround for the lack of self-referential inputs: when called from a module function call, this will pass the calling object in to the environment.

Worth some API bikeshedding, but seems like a pretty significant win, since it unblocks us from the hairy problem of self-calls. Now, at least, you can "agentify" any object by attaching an `Agent()` method to it.

### DIFF
--- a/docs/docs-graphql/schema.graphqls
+++ b/docs/docs-graphql/schema.graphqls
@@ -1829,6 +1829,15 @@ type Env {
     description: String!
   ): Env!
 
+  """Provide the calling object as an input to the environment"""
+  withCaller(
+    """The name of the binding"""
+    name: String!
+
+    """The description of the input"""
+    description: String!
+  ): Env!
+
   """Create or update a binding of type Cloud in the environment"""
   withCloudInput(
     """The name of the binding"""
--- a/docs/static/api/reference/index.html
+++ b/docs/static/api/reference/index.html
@@ -6930,6 +6930,27 @@
                         </td>
                       </tr>
                       <tr class="row-has-field-arguments">
+                        <td data-property-name=""><a class="property-name" id="Env-withCaller" href="#Env-withCaller"><code>withCaller</code></a> - <span class="property-type"><a href="#definition-Env"><code>Env!</code></a></span> </td>
+                        <td> Provide the calling object as an input to the environment </td>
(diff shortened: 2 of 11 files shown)


shykes:
cc @eunomie since it mentions self-calls.

shykes:
@vito how would this merge with #10370 ?

vito:
@shykes A few options:

---

At an API level, we could have a `currentObjectID`:

```graphql
extend type Query {
  currentObjectID: ID!
}
```

And then you just pass that ID to `Env.withBinding`.

But, at an SDK level, I think we'd run into a problem where perhaps we made `Env.withBinding` take a catch-all object type so you can pass a `*dagger.Container` instead of `dagger.ContainerID`. But in this case, you have no such object to refer to, so it seems like we start slipping down another tricky self-call-related slope.

To get around that, we could add a `currentObject` with a generic `Object` type:

```graphql
extend type Query {
  currentObject: Object!
}

type Object {
  id: ID!
  typeName: String!
  # ???
}
```

(Very similar to the traditional GraphQL [Node type](https://graphql.org/learn/global-object-identification/), but this `Object` above is concrete, not an interface.)

And then you can do:

```go
return dag.LLM().WithBinding("foo", dag.CurrentObject(), "blah blah")
```

I like that more, since you can avoid the extra sync call to get the ID:

```go
selfID, err := dag.CurrentObjectID(ctx)
if err != nil {
  return nil, err
}
return dag.LLM().WithBinding("foo", selfID, "blah blah"), nil
```

---

The third option is to just keep it as syntactic sugar since it's a very common case:

```go
return dag.LLM.WithCaller("foo", "blah blah")

return dag.LLM.WithSelf("foo", "blah blah") // or maybe this is less confusing?
```

Also worth noting: this works for an env _input_ but not an env _output_, since getting the value out would require self referential types again (`env.Output("foo").As???`), unless you pop open the hood and grab it with a raw GraphQL query.

But maybe we should call it `withCallerInput` for consistency anyway, and just not have a `withCallerOutput`?

Or maybe that's another reason to have an `Object` type, and you just do `env.Output("foo").AsObject()`? Not sure it's worth the trouble tbh.

PR #10773, opened 2025-07-21
feat: stabilize `DAGGER_HOST` env var

jedevc:
This environment variable is used by *many* users - changing it's format would be a major breaking change, that we'd need to release note anyways.

I think we should bite the bullet, and consider it stable. Especially with the changes around #10714, it would be useful to have a "proper" way to do this.

### DIFF
--- a/cmd/dagger/engine.go
+++ b/cmd/dagger/engine.go
@@ -20,8 +20,11 @@ import (
 )
 
 const (
-	GPUSupportEnv        = "_EXPERIMENTAL_DAGGER_GPU_SUPPORT"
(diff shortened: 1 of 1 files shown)


marcosnils:
SGTM! mostly validating that by removing this from experimental, it's aligned with a long-ish term plan of supporting this type of remote engine configuration.

shykes:
I would prefer to wait before we stabilize. I promise we won't let this rot. But we need another round of changes beyond #10714. I will work on a detailed proposal so that we don't lose momentum. I'm also OK with merging #10714, as long as there's no sadness when we break it.

jedevc:
Closing in preferenec of #10790.

PR #10775, opened 2025-07-21
Dagop file contents

alexcb:


core/file.go
```diff
@@ -229,6 +229,31 @@ func (file *File) Contents(ctx context.Context) ([]byte, error) {
...
+			return err
+		}
+
+		info, err := os.Stat(fullPath)
+		if err != nil {
+			return err
+		}
+		fileSize := info.Size()
+		if fileSize > buildkit.MaxFileContentsSize {
+			// TODO: move to proper error structure
```
alexcb: This comes from the non-dagop code above. I'm unsure if this means changing `fmt.Errorf` to something like:

```
type ErrTooBig struct {
   size int
}

func (err ErrTooBig) Error() string {
    return fmt.Sprintf("it's too big, it was %d bytes", err.size)
}
```

core/container_exec.go
```diff
@@ -355,6 +355,17 @@ func (container *Container) WithExec(ctx context.Context, opts ContainerExecOpts
 	worker := opt.Worker.(*buildkit.Worker)
 	worker = worker.ExecWorker(opt.CauseCtx, *execMD)
 	exec := worker.Executor()
+
+	svcs, err := query.Services(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get services: %w", err)
+	}
+	detach, _, err := svcs.StartBindings(ctx, container.Services)
```
alexcb: I'm kind of surprised this was somehow working without this? but when I switched contents() over to be a dagop, the exec below started failing when trying to do a lookup on the service hostname alias.

sipsma:
The integ tests _seem_ to be meaningfully slower in this PR than main:
* [this PR](https://github.com/dagger/dagger/actions/runs/16484436979?pr=10775) (re-ran 3 times)
* [comparison w/ recent main commit](https://github.com/dagger/dagger/actions/runs/16506627883)

There is often a decent amount of noise in the test runtimes, but this seems significant. It'd be worth another pair of eyes to check I'm not hallucinating and/or getting unlucky rolls that happen to all be on the extra slow end.

Assuming the above is correct, I'm not immediately sure what the cause would be, but we should figure it out. The first guess that comes to mind is that we are "double-caching" by reading the file contents (which are in buildkit cache) and then writing them again because `contents` is wrapped in dag-op? Just an initial guess though.

core/container_exec.go
jedevc: Mmmmm. This is my bad :stuck_out_tongue_closed_eyes: 

So there are two places we start bindings:
- Outside dagop - genuinely, we didn't used to be able to do better. Everything gets done in buildkit, we have no control. This means we will sometimes start services for an op that never actually runs because it'll be cached.
- Inside dagop. We can do all of that now. This means we only start services when we need them.

This was working before, because we were starting bindings *everywhere* we evaluated refs. But, now because of the split, I think potentially we might be calling this twice.

My *suggestion* here, would be to split this change out of this PR. And in that, remove *all* of the outside dagop service starts (like [this one](https://github.com/dagger/dagger/blob/3e890bc1ad98bfa919e2d5a0e45ba3017433f7b2/core/file.go#L258-L286)).

We only need to start the bindings when we use the services - inside the dagops. I think with `withExec`, `http` and `git` merged, we have all the buildkit ops that use service converted - so we can push the service starts down into all of them.

This *should* actually be a performance boost - we won't start unnecessary services, and stops service code from infiltrating *every* method, so should be a generally nice cleanup.

WDYT?

jedevc:
Yeah, does seem to be slower :cry: 

The double caching thing does sound likely - based on the impl of `contents`, I would *probably* not wrap it in dagop? It's already just directly reading from a cache ref.

core/container_exec.go
alexcb: Ah ha! this makes a lot of sense. I'll give it a try.

alexcb: I was able to remove a lot under https://github.com/dagger/dagger/pull/10803

There's still some calls left that are needed for exporting or publishing.

alexcb:
Good catch @sipsma, definitely don't want this slow-down.

> I would probably not wrap it in dagop? It's already just directly reading from a cache ref.

@jedevc so far my initial tests seems to be working. I'll work on tidying up the changes before force pushing a nearly-rewritten PR :)

In the meanwhile I moved the `StartBindings` related changes into #10803

alexcb:
#10803 has now been merged.

I backed out the dagop wrapping for the cache ref reads, and the tests run times are now comparable with main, e.g.
attempts [2](https://github.com/dagger/dagger/actions/runs/16971087622/attempts/2) and [3](https://github.com/dagger/dagger/actions/runs/16971087622/attempts/3) took 10m46s and 11m53s respectively

Note that the scan test is currently failing (waiting on #10878)

alexcb:
#10937 introduced a non-trivial conflict that needs to be dealt with now.

core/file.go
```diff
@@ -168,49 +168,59 @@ func (file *File) Contents(ctx context.Context, offset, limit *int) ([]byte, err
 		return nil, nil
 	}
 
-	r, err := file.Open(ctx)
-	if err != nil {
-		return nil, err
-	}
-	defer r.Close()
```
jedevc, line 175: As a follow-up, everywhere that uses this should probably be refactored to use `execInMount`.

It would be nice to remove `file.Open` entirely?

But yeah, this is a nice change, as I read it, the previous implementation actually does a mount/unmount for *every* call to `Read`? Oof.

jedevc (approved):

PR #10779, opened 2025-07-22
cli: add support for a []BuildArg flag

helderco:
This allows passing build args to a docker build via the CLI.

Example
```
host | directory . | docker-build --platform darwin/arm64 --build-args BIN_NAME=docker-run,NO_ARCHIVE=true --target export-bin | export .
```

\cc @eunomie

### DIFF
--- a/cmd/dagger/flags.go
+++ b/cmd/dagger/flags.go
@@ -58,6 +58,8 @@ func GetCustomFlagValue(name string) DaggerValue {
 		return &moduleValue{}
 	case Platform:
 		return &platformValue{}
+	case BuildArg:
+		return &buildArgValue{}
 	case Socket:
 		return &socketValue{}
 	case GitRepository:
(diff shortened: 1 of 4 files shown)


cmd/dagger/flags.go
```diff
@@ -741,6 +746,43 @@ func (v *platformValue) Get(ctx context.Context, dag *dagger.Client, _ *dagger.M
...
+	pair := strings.Trim(s, `"`)
+	kv := strings.SplitN(pair, "=", 2)
+	if len(kv) != 2 {
+		return fmt.Errorf("%s must be formatted as name=value", pair)
+	}
+	if kv[0] == "" {
+		return fmt.Errorf("%s cannot have an empty name", pair)
+	}
+	v.name = kv[0]
+	v.value = kv[1]
```
jedevc: suggestion: `strings.Cut` is a neater way of doing this.

helderco: Indeed! Just copied from pflag's [stringToString](https://github.com/spf13/pflag/blob/c78f730fb023e4012c4097b24408867cd5c5bdde/string_to_string.go#L91-L97) flag but that code's 7 years old 😅

I thought about reusing that flag, but it turns out to be more code on our side since we already have the slice abstraction/generic hooked up to everything needed.

PR #10790, opened 2025-07-25
DAGGER_HOST, DAGGER_IMAGE

shykes:
## Overview

Stable configuration for provisioning and connecting to the dagger engine from the dagger CLI.

| Variable  | Description |
| -- |  -- |
| `DAGGER_CLOUD_ENGINE` | Connect to Dagger Cloud experimental engine |
| `DAGGER_HOST` | Connect directly to a Dagger Engine. Supports `tcp://`, `ssh://`, `unix://`, `container://`, `kube-pod://`. You are responsible for engine provisioning, and version management. Warranty is void. |
| `DAGGER_IMAGE` | Provision a Dagger Engine from the given container image. Tag is optional: if omitted, the current CLI version is used |
| `DAGGER_IMAGE_VERSION` | Provision a Dagger Engine of the given version. Can be used in combination with `DAGGER_IMAGE`, or standalone. |
| `DAGGER_IMAGE_RUNNER` | Provision a Dagger Engine with the given container runner. If omitted, a runner is auto-discovered. Supports `docker`, `podman`, `containerd`, `apple` |
| `DAGGER_IMAGE_MEMORY` | Provision a Dagger Engine with custom memory configuration |
| `DAGGER_IMAGE_CPUS` | Provision a Dagger Engine with custom CPU allocation |
| `DAGGER_IMAGE_PORT` | Provision a Dagger Engine with a custom port number |
| `DAGGER_IMAGE_CONTAINER_NAME` | Provision a Dagger Engine with a custom container name |
| `DAGGER_IMAGE_VOLUME_NAME` | Provision a Dagger Engine with a custom volume name |
| `DAGGER_IMAGE_VOLUME_NAME` | Provision a Dagger Engine with a custom volume name |

## Compatibility

`_EXPERIMENTAL_DAGGER_RUNNER_HOST` is deprecated, but we maintain backwards compatibility, with a warning.

## Hooks

~~An experimental hook is included, to override how to provision and connect to the engine: `.config/dagger/hooks/engine-run`.~~

I removed the hook implementation, for simplicity. We may not need it after all?

## Implementation

This is built on top of #10714 by @jedevc . It acts as a facade to the driver system. Drivers are still used, but they are no longer presented directly to the user via a URL scheme system. This implies a follow-up, which can collapse much of the driver system to simplify the code. For now, I am prioritizing continuity with Justin's work at the expense of overall simplicity, seems worth it!

### DIFF
--- a/engine/client/drivers/container.go
+++ b/engine/client/drivers/container.go
@@ -9,7 +9,6 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
-	"strconv"
 	"strings"
 
 	"dagger.io/dagger/telemetry"
@@ -128,23 +127,14 @@ func (d *imageDriver) Available(ctx context.Context) (bool, error) {
 }
 
 func (d *imageDriver) Provision(ctx context.Context, target *url.URL, opts *DriverOpts) (Connector, error) {
-	cleanup := true
-	if val, ok := os.LookupEnv("DAGGER_LEAVE_OLD_ENGINE"); ok {
-		b, _ := strconv.ParseBool(val)
-		cleanup = !b
-	} else if val := target.Query().Get("cleanup"); val != "" {
-		cleanup, _ = strconv.ParseBool(val)
-	}
-
-	port, _ := strconv.Atoi(target.Query().Get("port"))
 	target, err := d.create(ctx, containerCreateOpts{
+		containerName: opts.ContainerName,
+		volumeName:    opts.VolumeName,
+		cleanup:       opts.Cleanup,
+		port:          opts.Port,
+		cpus:          opts.CPUs,
+		memory:        opts.Memory,
 		imageRef:      target.Host + target.Path,
-		containerName: target.Query().Get("container"),
-		volumeName:    target.Query().Get("volume"),
-		cleanup:       cleanup,
-		port:          port,
-		cpus:          target.Query().Get("cpus"),
-		memory:        target.Query().Get("memory"),
 	}, opts)
 	if err != nil {
 		return nil, err
@@ -152,7 +142,6 @@ func (d *imageDriver) Provision(ctx context.Context, target *url.URL, opts *Driv
 	return containerConnector{
 		backend: d.backend,
 		host:    target.Host,
-		values:  target.Query(),
 	}, nil
 }
 
@@ -162,7 +151,6 @@ func (d *imageDriver) ImageLoader(ctx context.Context) imageload.Backend {
 
 type containerConnector struct {
 	host    string
-	values  url.Values
 	backend containerBackend
 }
 
@@ -179,7 +167,6 @@ func (d *containerDriver) Provision(ctx context.Context, target *url.URL, opts *
 	return containerConnector{
 		backend: d.backend,
 		host:    target.Host,
-		values:  target.Query(),
 	}, nil
 }
 
@@ -189,10 +176,12 @@ func (d *containerDriver) ImageLoader(ctx context.Context) imageload.Backend {
 
 func (d containerConnector) Connect(ctx context.Context) (net.Conn, error) {
 	args := []string{}
-	if context := d.values.Get("context"); context != "" {
-		args = append(args, "--context="+context)
+	// FIXME: what did '?context=..' do, and do we need it back?
+	if cmd, ok := os.LookupEnv("DAGGER_IMAGE_COMMAND"); ok {
+		args = append(args, "sh", "-c", cmd)
+	} else {
+		args = append(args, "buildctl", "dial-stdio")
 	}
-	args = append(args, "buildctl", "dial-stdio")
 
 	// using uncancelled context because context remains active for the
 	// duration of the process, after dial has completed
@@ -294,11 +283,11 @@ func (d *imageDriver) create(ctx context.Context, opts containerCreateOpts, dopt
 	}
 
 	if dopts.DaggerCloudToken != "" {
-		runOptions.env = append(runOptions.env, fmt.Sprintf("%s=%s", EnvDaggerCloudToken, dopts.DaggerCloudToken))
+		runOptions.env = append(runOptions.env, "DAGGER_CLOUD_TOKEN="+dopts.DaggerCloudToken)
 	}
-	if dopts.GPUSupport != "" {
+	if dopts.GPUSupport {
 		runOptions.gpus = true
-		runOptions.env = append(runOptions.env, fmt.Sprintf("%s=%s", EnvGPUSupport, dopts.GPUSupport))
+		runOptions.env = append(runOptions.env, "_EXPERIMENTAL_DAGGER_GPU_SUPPORT=1")
 	}
 	if opts.port != 0 {
 		runOptions.ports = append(runOptions.ports, fmt.Sprintf("%d:%d", opts.port, opts.port))
--- a/cmd/dagger/engine.go
+++ b/cmd/dagger/engine.go
@@ -2,15 +2,12 @@ package main
 
 import (
 	"context"
-	"fmt"
 	"os"
 
 	"dagger.io/dagger/telemetry"
 	"github.com/dagger/dagger/dagql/dagui"
-	"github.com/dagger/dagger/engine"
 	"github.com/dagger/dagger/engine/client"
 	"github.com/dagger/dagger/engine/client/imageload"
-	"github.com/dagger/dagger/engine/distconsts"
 	"github.com/dagger/dagger/engine/slog"
 	enginetel "github.com/dagger/dagger/engine/telemetry"
 	"github.com/dagger/dagger/util/cleanups"
@@ -20,45 +17,10 @@ import (
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 )
 
-const (
-	GPUSupportEnv        = "_EXPERIMENTAL_DAGGER_GPU_SUPPORT"
-	RunnerHostEnv        = "_EXPERIMENTAL_DAGGER_RUNNER_HOST"
-	RunnerImageLoaderEnv = "_EXPERIMENTAL_DAGGER_RUNNER_IMAGESTORE"
-)
-
 var (
-	// RunnerHost holds the host to connect to.
-	//
-	// Note: this is filled at link-time.
-	RunnerHost string
-
 	// RunnerImageLoader holds the image store for the client.
 	RunnerImageLoader string
 )
-
-func init() {
-	if v, ok := os.LookupEnv(RunnerHostEnv); ok {
-		RunnerHost = v
-	}
-	if RunnerHost == "" {
-		RunnerHost = defaultRunnerHost()
-	}
-
-	RunnerImageLoader = os.Getenv(RunnerImageLoaderEnv)
-}
-
-func defaultRunnerHost() string {
-	tag := engine.Tag
-	if tag == "" {
-		// can happen during naive dev builds (so just fallback to something
-		// semi-reasonable)
-		return "container://" + distconsts.EngineContainerName
-	}
-	if os.Getenv(GPUSupportEnv) != "" {
-		tag += "-gpu"
-	}
-	return fmt.Sprintf("image://%s:%s", engine.EngineImageRepo, tag)
-}
 
 type runClientCallback func(context.Context, *client.Client) error
 
@@ -88,15 +50,11 @@ func withEngine(
 		if debug {
 			params.LogLevel = slog.LevelDebug
 		}
+		// Apply global user configuration (env & CLI) as an override
+		params = params.Override(engineConfig)
 
-		if useCloudEngine {
-			params.RunnerHost = "dagger-cloud://default-engine-config.dagger.cloud"
-		} else if params.RunnerHost == "" {
-			params.RunnerHost = RunnerHost
-		}
-
-		if RunnerImageLoader != "" {
-			backend, err := imageload.GetBackend(RunnerImageLoader)
+		if imageLoader != "" {
+			backend, err := imageload.GetBackend(imageLoader)
 			if err != nil {
 				return cleanup.Run, err
 			}
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -39,6 +39,7 @@ import (
 	"github.com/dagger/dagger/dagql/dagui"
 	"github.com/dagger/dagger/dagql/idtui"
 	"github.com/dagger/dagger/engine"
+	"github.com/dagger/dagger/engine/client"
 	"github.com/dagger/dagger/engine/client/pathutil"
 	"github.com/dagger/dagger/engine/slog"
 	enginetel "github.com/dagger/dagger/engine/telemetry"
@@ -67,7 +68,9 @@ var (
 	interactiveCommandParsed []string
 	web                      bool
 	noExit                   bool
-	_, useCloudEngine        = os.LookupEnv("DAGGER_CLOUD_ENGINE")
+	engineConfig             = client.DefaultParams()
+
+	imageLoader = os.Getenv("_EXPERIMENTAL_DAGGER_RUNNER_IMAGESTORE")
 
 	dotOutputFilePath string
 	dotFocusField     string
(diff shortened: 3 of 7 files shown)


jedevc:
Hm, if we're doing this, I do wonder if we should just go all out and refactor the entire driver interface. Maintaining both is gonna be a pain.

- `DAGGER_ENGINE_IMAGE`

  One thing to keep in mind is that we *still* need to have a way of providing "provisioning parameters". Things like `volume`, or `cpu`s and `memory`. I think probably most of these should be moved to the global config `engine.json` (we could rename that now as well?), instead of going through env vars.

- `DAGGER_ENGINE_IMAGE_COMMAND`

  Pretty heavily disagree on exposing this. Changing this will probably just result in clients failing to connect for weird reasons, all of this should be internal. Is there a specific use case that this accomplishes?

- `DAGGER_ENGINE_VERSION`

  I like that this is split from `DAGGER_ENGINE_IMAGE` (so that we'd default to the CLI version), but I don't think that we should allow overriding it. Setting this to different than the CLI is not that useful outside of working on dagger itself. Handling different versions of clis and engines is best effort support - if we actually intend to support this properly, we need better test matrices, etc.

I also propose `DAGGER_ENGINE_IMAGE_PROVIDER` (or a similar config in the `engine.json`). If a user has *both* podman + docker installed, there should be a way to choose which one, while still having the sensible default.

shykes:
> * `DAGGER_ENGINE_IMAGE`
>   One thing to keep in mind is that we _still_ need to have a way of providing "provisioning parameters". Things like `volume`, or `cpu`s and `memory`. I think probably most of these should be moved to the global config `engine.json` (we could rename that now as well?), instead of going through env vars.

IMO this can be done with hooks. It's not worth the considerable effort to support all this extra configuration in a way that is clean, well maintained, and supports all edge cases. Power users will prefer hooks anyway, and we will be grateful not to be on the hook for maintaining all that extra code.

> * `DAGGER_ENGINE_IMAGE_COMMAND`
>   Pretty heavily disagree on exposing this. Changing this will probably just result in clients failing to connect for weird reasons, all of this should be internal. Is there a specific use case that this accomplishes?

Meh, it's just an escape hatch. Only power users will even think to use this. And they will now that it voids the warranty.

> * `DAGGER_ENGINE_VERSION`
>   I like that this is split from `DAGGER_ENGINE_IMAGE` (so that we'd default to the CLI version), but I don't think that we should allow overriding it. Setting this to different than the CLI is not that useful outside of working on dagger itself. Handling different versions of clis and engines is best effort support - if we actually intend to support this properly, we need better test matrices, etc.

Again - meh. Yeah it voids the warranty, but so does `DAGGER_ENGINE_HOST`... It's simpler to keep it than to remove it, and it's very obvious and natural what it does. I'd rather leave it for the benefit of power users, than go out of my way to nerf it.

> I also propose `DAGGER_ENGINE_IMAGE_PROVIDER` (or a similar config in the `engine.json`). If a user has _both_ podman + docker installed, there should be a way to choose which one, while still having the sensible default.

Great idea, I added that 👍

shykes:
I added a basic hook implementation. Example, bootstrapped from stable dagger:

```console
#!/usr/bin/env dagger

ctr() {
  container |
  from alpine |
  with-exec apk add socat |
  with-file /bin/dagger $(./cmd/dagger | binary) |
  with-service-binding dagger-engine $(./cmd/engine | service dagger-dev) |
  with-new-file --permissions=711 --expand '$HOME/.config/dagger/hooks/engine-run' '#!/bin/sh\nsocat - tcp://dagger-engine:1234' |
  without-env-variable _EXPERIMENTAL_DAGGER_RUNNER_HOST
}

ctr | terminal
```

Then: `dagger`

shykes:
@jedevc @jpadams @marcosnils @kpenfound @gerhard @matipan updated description. Please let me know your thoughts.

Implementation works, just neefs tests & cleanup.

- no more hooks
- no escape hatch
- simpler var names
- extra vars for memory, cpu, container name etc

jedevc:
Got some comments :eyes: It's better than what we have now!

I'm not *entirely* convinced that splitting out all the config settings into lots of different env vars is easier than what we have now - some of the options don't work with each other (e.g. I can set `DAGGER_HOST=tcp://...` and `DAGGER_CLOUD_ENGINE` together, but it's not entirely apparent what that *should* do).

> `DAGGER_HOST`

What's the priority on different vars? If I have all `DAGGER_HOST`, `DAGGER_IMAGE` and `DAGGER_CLOUD_ENGINE` set, what happens?

This was one of the main reasons I liked the idea of one unified env var, but as long as the behavior is clear and well-defined, I'm fine.

> `DAGGER_IMAGE_RUNNER`

I think this should be `DAGGER_DRIVER` (maybe `DAGGER_RUNNER` is *okay*). This isn't just for provisioning an image, it also affects the `container://` value in `DAGGER_HOST`.

> `DAGGER_IMAGE_MEMORY`, `DAGGER_IMAGE_CPUS`

I'm wondering if these should even be *available* as environment variables. These essentially are equivalent to controlling the settings in Docker Desktop to assign the CPU and memory - we want them to be persistent. So maybe we should *only* allow configuration in `~/.config/dagger/engine.json`?

Something like:

```json
{
  "drivers": {
    "apple": {
      "cpus": 8,
      "memory": "16G"
    }
  }
}
```

> `DAGGER_IMAGE_PORT`, `DAGGER_IMAGE_CONTAINER_NAME`, `DAGGER_IMAGE_VOLUME_NAME`, `DAGGER_IMAGE_VOLUME_NAME`

Thinking back, the only real reason these exist is to enable the `./hack/dev`
use case - they're not really important for end users.

See:

```
$ dagger-dev version
dagger v0.18.15-010101000000-dev-0420629f94c7 (docker-image://localhost/dagger-engine.dev?container=dagger-engine.dev&volume=dagger-engine.dev&cleanup=false) linux/amd64
```

I'd suggest removing these entirely. Once they're "official" environment variables, we need to support them indefinitely. Until we have an actual end-user case, I think we should either have them only configurable as part of the legacy `_EXPERIMENTAL_DAGGER_RUNNER_HOST`, or set using `_EXPERIMENTAL_DAGGER_IMAGE_PORT` or similar.

> `DAGGER_CLOUD_ENGINE`

Discussed this a bit with @matipan:

- Does configuring `DAGGER_IMAGE` affect the cloud engine?
- How would we configure the "size" of cloud engine instances?
    - We can't reuse the `CPU` or `MEMORY` env vars, because cloud only supports discrete instance sizes (does feel like a bit of a footgun though - I can definitely see a user configuring `DAGGER_IMAGE_CPUS` expecting to configure the number of cloud CPUs).

engine/client/drivers/container.go
```diff
@@ -126,31 +125,21 @@ func (d *imageDriver) Available(ctx context.Context) (bool, error) {
 }
 
 func (d *imageDriver) Provision(ctx context.Context, target *url.URL, opts *DriverOpts) (Connector, error) {
-	cleanup := true
-	if val, ok := os.LookupEnv("DAGGER_LEAVE_OLD_ENGINE"); ok {
-		b, _ := strconv.ParseBool(val)
-		cleanup = !b
-	} else if val := target.Query().Get("cleanup"); val != "" {
-		cleanup, _ = strconv.ParseBool(val)
-	}
```
jedevc, line 135: :eyes: we're removing this functionality?

jedevc:
I realize it's frustrating to suggest another approach with an almost complete implementation - I would have liked to discuss this more before an implementation appeared.

But I wonder if *instead*, we should only expose *two* environment variables:
- `DAGGER_HOST` - exactly as proposed in this PR, just the "primitives": `tcp`/`ssh`/`unix`/`container`/`kube-pod`
- `DAGGER_DRIVER` - a single string, `cloud`/`image`/`image+docker`/etc. Essentially just the protocol in the env var we support today, not a full URL. 
	- This is essentially anything that does "provisioning"

*All* configuration details for provisioning **must** be configured in the `~/.config/dagger/engine.json` file. There would be blocks *per* driver, so you could configure `drivers.image-docker` separately from `drivers.image-apple`. You'd also be able to configure `drivers.image` to configure which backend to use by default.

For more fine-grained configuration, we could potentially support reading a `dagger.engine.json` file or similar when loading a module, that would override the global defaults. This would allow you to configure a specific configuration for different modules.

I feel like we should lean into the config file approach - I feel like is what we'll end up with anyways, there's too many environment variables to carry around, so users will just put them into a `.profile` file or similar anyways.

matipan:
> All configuration details for these must be configured in the ~/.config/dagger/engine.json file. There would be blocks per driver, so you could configure drivers.image-docker separately from drivers.image-apple. You'd also be able to configure drivers.image to configure which backend to use by default.

The argument against a config file for me is what happens in a CI environment. Having to manually write a config file to then run the dagger command feels like a paper cut. However, I don't think that will be a problem. The majority of the configs that you mention should not be applicable to a CI environment, so users shouldn't have the need to set up the config file. And if we find a case where it's needed, then that's a great excuse to improve our CI integrations (e.g., our Dagger GitHub action) and let it take care of creating the file with the right properties.

> - How would we configure the "size" of cloud engine instances?
>   - We can't reuse the CPU or MEMORY env vars, because cloud only supports discrete instance sizes (does feel like a bit of a footgun though - I can definitely see a user configuring DAGGER_IMAGE_CPUS expecting to configure the number of cloud CPUs).

I would prefer to stay away from letting users define custom resources in our cloud environment. It complicates things quite a bit, and based on our experience, I think it's best to stick to instance "classes" and let users choose based on that. That said, I do wonder how we might want to let other properties be configurable. The most relevant one that comes to mind is whether a user needs a dedicated or shared engine. There's the option of hiding this property away in the engine config and not letting the CLI specify it. But it's worth considering when thinking about how to configure these options.

shykes:
> I'm not _entirely_ convinced that splitting out all the config settings into lots of different env vars is easier than what we have now - some of the options don't work with each other (e.g. I can set `DAGGER_HOST=tcp://...` and `DAGGER_CLOUD_ENGINE` together, but it's not entirely apparent what that _should_ do).

Right. In this PR there are 3 "tiers" of configuration: cloud, host, image.

Note that in main there are already 2 tiers: cloud, and everything else.

> > `DAGGER_HOST`
> 
> What's the priority on different vars? If I have all `DAGGER_HOST`, `DAGGER_IMAGE` and `DAGGER_CLOUD_ENGINE` set, what happens?

At the moment the lookup order is: cloud, host, image. We could possibly switch it to: host, cloud, image.

> This was one of the main reasons I liked the idea of one unified env var, but as long as the behavior is clear and well-defined, I'm fine.

I agree that was an advantage of the unified URL (with the caveat that we already had to support `DAGGER_CLOUD_ENGINE` so it wasn't actually fully unified). However I think the tradeoff is worth it.

> > `DAGGER_IMAGE_RUNNER`
> 
> I think this should be `DAGGER_DRIVER` (maybe `DAGGER_RUNNER` is _okay_). This isn't just for provisioning an image, it also affects the `container://` value in `DAGGER_HOST`.

Ah. Sorry, the complexity of the code made me miss that important detail 😛

Indeed the `_IMAGE_` prefix doesn't match what you explained. Let me noodle on that for a minute.

> > `DAGGER_IMAGE_MEMORY`, `DAGGER_IMAGE_CPUS`
> 
> I'm wondering if these should even be _available_ as environment variables. These essentially are equivalent to controlling the settings in Docker Desktop to assign the CPU and memory - we want them to be persistent. So maybe we should _only_ allow configuration in `~/.config/dagger/engine.json`?

I don't mind, but we should have a consistent rule for which configuration goes in env variable, CLI flags, and config file. Lacking that, we should just default to supporting all of them in all three places, which is easier to reason about.

>  > `DAGGER_IMAGE_PORT`, `DAGGER_IMAGE_CONTAINER_NAME`, `DAGGER_IMAGE_VOLUME_NAME`, `DAGGER_IMAGE_VOLUME_NAME`

> I'd suggest removing these entirely. Once they're "official" environment variables, we need to support them indefinitely. Until we have an actual end-user case, I think we should either have them only configurable as part of the legacy `_EXPERIMENTAL_DAGGER_RUNNER_HOST`, or set using `_EXPERIMENTAL_DAGGER_IMAGE_PORT` or similar.

Wouldn't this mean continuing to maintain the code for our own use case, *and* being forced to keep `_EXPERIMENTAL_DAGGER_RUNNER_HOST` in our own code, instead of leading by example and using the stable variables?

I would be all for a clean cut. But not a messy partial cut.

jedevc:
> > > `DAGGER_IMAGE_MEMORY`, `DAGGER_IMAGE_CPUS`
> > 
> > I'm wondering if these should even be _available_ as environment variables. These essentially are equivalent to controlling the settings in Docker Desktop to assign the CPU and memory - we want them to be persistent. So maybe we should _only_ allow configuration in `~/.config/dagger/engine.json`?
> 
> I don't mind, but we should have a consistent rule for which configuration goes in env variable, CLI flags, and config file. Lacking that, we should just default to supporting all of them in all three places, which is easier to reason about.

I think the rule should be that the configuration of *any* specific driver should be in the config file. The only env vars we should support should be the ones that select which driver to use, i.e. the proposal here: https://github.com/dagger/dagger/pull/10790#issuecomment-3179560145

> > > `DAGGER_IMAGE_PORT`, `DAGGER_IMAGE_CONTAINER_NAME`, `DAGGER_IMAGE_VOLUME_NAME`, `DAGGER_IMAGE_VOLUME_NAME`
> 
> > I'd suggest removing these entirely. Once they're "official" environment variables, we need to support them indefinitely. Until we have an actual end-user case, I think we should either have them only configurable as part of the legacy `_EXPERIMENTAL_DAGGER_RUNNER_HOST`, or set using `_EXPERIMENTAL_DAGGER_IMAGE_PORT` or similar.
> 
> Wouldn't this mean continuing to maintain the code for our own use case, _and_ being forced to keep `_EXPERIMENTAL_DAGGER_RUNNER_HOST` in our own code, instead of leading by example and using the stable variables?
> 
> I would be all for a clean cut. But not a messy partial cut.

The main reason to have `volume`, `port`, `container-name` as URL query parameters was so that we could build a `dagger` binary that would automatically use the image that we just built (instead of *needing* to set the `_EXPERIMENTAL_DAGGER_RUNNER_HOST` in the env, which I know was miserable to do). They weren't really intended to be overridden by the user.

https://github.com/dagger/dagger/blob/5fd1f23dd48337ee02b4917116d1e236fdaed0d7/hack/build#L13-L14

If we're just building them in, we could also just add args for those above there, so that they can be set at build time, but never at runtime.

That's still a *little* bit of special casing? But not exposing any configuration surface, if we do this, there's no reason to have those.

shykes:
> > * How would we configure the "size" of cloud engine instances?
> >   
> >   * We can't reuse the CPU or MEMORY env vars, because cloud only supports discrete instance sizes (does feel like a bit of a footgun though - I can definitely see a user configuring DAGGER_IMAGE_CPUS expecting to configure the number of cloud CPUs).
> 
> I would prefer to stay away from letting users define custom resources in our cloud environment. It complicates things quite a bit, and based on our experience, I think it's best to stick to instance "classes" and let users choose based on that. That said, I do wonder how we might want to let other properties be configurable. The most relevant one that comes to mind is whether a user needs a dedicated or shared engine. There's the option of hiding this property away in the engine config and not letting the CLI specify it. But it's worth considering when thinking about how to configure these options.

My view on this is that when using Dagger Cloud, the client should have *no input at all* on the infrastructure configuration. Those should all be configured out-of-band by administrators of the Dagger Cloud account, and mapped at request time by user, module, function name, CLI version, API token... *maybe* by freeform tags, and even that I'm worried will be abused.

shykes:
@jedevc the split between `DAGGER_HOST` and `DAGGER_IMAGE` is the best UX so far. Anything that removes `DAGGER_IMAGE` is a UX regression in my view.

jedevc:
We can keep `DAGGER_IMAGE`, though I'd still prefer that it be a setting in a config file. I don't really care though about it though, I just don't see where I'd use it practically.

I guess it's useful if I'm setting up a mirror? But in that case, I'd want something where I can configure it in a more persistent way, not just as a temporary environment variable.

shykes:
I think `DAGGER_IMAGE` is very useful for power users and administrators who need to point to a local registry mirror, or to a custom built engine image. I like that it is narrowly scoped to a container image ref. so it can drop the url scheme and other unnecessary details.

PR #10793, opened 2025-07-25
Fix: pass cloudOrg through to nested clients

kpenfound:
Follow up fix for https://github.com/dagger/dagger/pull/10580 to make the trace-url available to the SDKs. Looks like the cloudOrg wasn't being passed through to nested sessions, so it would work with `dagger core cloud trace-url` but not via SDK

### DIFF
--- a/engine/server/session.go
+++ b/engine/server/session.go
@@ -895,18 +895,6 @@ func (srv *Server) ServeHTTPToNestedClient(w http.ResponseWriter, r *http.Reques
 		allowedLLMModules = md.AllowedLLMModules
 	}
 
-	// Inherit CloudOrg from the main client caller
-	var cloudOrg string
-	srv.daggerSessionsMu.RLock()
-	if sess, ok := srv.daggerSessions[execMD.SessionID]; ok {
-		if mainClient, err := srv.clientFromIDs(execMD.SessionID, sess.mainClientCallerID); err == nil {
-			if mainClient.clientMetadata != nil {
-				cloudOrg = mainClient.clientMetadata.CloudOrg
-			}
-		}
-	}
-	srv.daggerSessionsMu.RUnlock()
-
 	httpHandlerFunc(srv.serveHTTPToClient, &ClientInitOpts{
 		ClientMetadata: &engine.ClientMetadata{
 			ClientID:          execMD.ClientID,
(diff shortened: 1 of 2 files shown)


engine/server/session.go
```diff
@@ -911,6 +911,18 @@ func (srv *Server) ServeHTTPToNestedClient(w http.ResponseWriter, r *http.Reques
...
+	var cloudOrg string
+	srv.daggerSessionsMu.RLock()
+	if sess, ok := srv.daggerSessions[execMD.SessionID]; ok {
+		if mainClient, err := srv.clientFromIDs(execMD.SessionID, sess.mainClientCallerID); err == nil {
+			if mainClient.clientMetadata != nil {
+				cloudOrg = mainClient.clientMetadata.CloudOrg
+			}
+		}
+	}
+	srv.daggerSessionsMu.RUnlock()
```
marcosnils: I'm not super familiarized with this code but I skimmed through the server struct looking to re-use some existing functions and works with this
```patch
diff --git a/engine/server/session.go b/engine/server/session.go
index fd4034d53..f8a32c9b1 100644
--- a/engine/server/session.go
+++ b/engine/server/session.go
@@ -906,15 +906,21 @@ func (srv *Server) ServeHTTPToNestedClient(w http.ResponseWriter, r *http.Reques
 		clientVersion = engine.Version
 	}
 	allowedLLMModules := execMD.AllowedLLMModules
+	var cloudOrg string
+	if c, _ := srv.clientFromIDs(execMD.SessionID, execMD.CallerClientID); c != nil && c.clientMetadata != nil {
+		cloudOrg = c.clientMetadata.CloudOrg
+	}
 	if md, _ := engine.ClientMetadataFromHTTPHeaders(r.Header); md != nil {
 		clientVersion = md.ClientVersion
 		allowedLLMModules = md.AllowedLLMModules
 	}
 
 	httpHandlerFunc(srv.serveHTTPToClient, &ClientInitOpts{
 		ClientMetadata: &engine.ClientMetadata{
 			ClientID:          execMD.ClientID,
 			ClientVersion:     clientVersion,
+			CloudOrg:          cloudOrg,
 			ClientSecretToken: execMD.SecretToken,
 			SessionID:         execMD.SessionID,
 			ClientHostname:    execMD.Hostname,
```

 probably @sipsma  knows what's the "proper" way here

jedevc: So, I *think* we want the non-module client parent - not the main client.

In practice, this really only matters for our integration tests. But without that, we'd instead be trying to get the info from the top-level client, *not* the `dagger call` in the integration tests (which makes things very annoying to actually test and instrument).

jedevc: Mm, as we discussed online - it's *still* one single trace. So, it really should be the main caller.

We don't support a trace-inside-a-trace. So this is a bit of a weird case, but the main client is the right one here.

kpenfound:
thanks for moving this forward @jedevc 🙏

marcosnils:
LGTM to merge once tests pas

Issue #10781, opened 2025-07-22
✨ WithExec option for secret stdin


daniel-naegele:
### What are you trying to do?

A very neat feature would be the ability to specify stdin for a command in WithExec which is a secret variable so that it doesn't get exposed in the logs.

### Why is this important to you?

When trying to use the tool skopeo, I encountered its limitation that it does not expand environment variable in the login credentials. Directly supplying the credentials is a no-go, as they would be exposed then in Dagger's logs.  The other option to supply a password to skopeo is to use the stdin, just like Docker does. However, the official skopeo container does not allow for tools like sh/bash to pipe the secret into the skopeo command.

The following example does not work
```go
dag.Container().From("quay.io/skopeo/stable").WithSecretVariable("PASSWORD", password).
  WithExec([]string{"skopeo", "login", "--username", username, "--password", "$PASSWORD", "some.registry.host"})
```

One could use the already existing Stdin option in WithExec, however the credential then gets exposed in the logs.

### How are you currently working around this?

The current workaround uses the third login option for skopeo, using an auth file (same format as Dockers auth file). Thus, I am first creating the auth file using a Docker container and then mounting the authfile inside the skopeo container.
```go
	authFile := dag.Container().
		From("docker").
		WithSecretVariable("PASSWORD", password).
		WithExec([]string{"sh", "-c", "echo $PASSWORD | docker login -u " + user + " --password-stdin " + registry}).
		File("/root/.docker/config.json")

skopeoArgs := []string{...}

imageDigest, err := dag.Container().
                From("quay.io/skopeo/stable").
		WithFile("config.json", authFile).
		WithExec(skopeoArgs).
```

skycaptain:
Have you considered using a mounted secret and `--password-stdin`? e.g. roughly

```go
WithMountedSecret("/run/secrets/MY_TOKEN", authToken).
WithExec(
	[]string{"sh"},
	dagger.ContainerWithExecOpts{
		Stdin: fmt.Sprintf(`
skopeo login '%s' \
--username='%' \
--password-stdin \
'%s' < /run/secrets/MY_TOKEN
`, url, authUsername),
	},
).
WithoutMount("/run/secrets/MY_TOKEN")
```

daniel-naegele:
The problem is, that the default container image for skopeo does not have a shell included. Otherwise I could have used a similar approach like the auxiliary Docker container in my second listing.

jedevc:
[...]

PR #10800, opened 2025-07-28
feat: add RedirectStdin argument to WithExec

jedevc:
Fixes https://github.com/dagger/dagger/issues/10781.

This allows reading stdin from a file in the container - we name this option `RedirectStdin` to be inline with `RedirectStdout`/`RedirectStderr`.


core/integration/container_test.go
```diff
@@ -579,44 +579,40 @@ func (ContainerSuite) TestExecStdin(ctx context.Context, t *testctx.T) {
 	require.Equal(t, res.Container.From.WithExec.Stdout, "hello")
 }
 
-func (ContainerSuite) TestExecRedirectStdoutStderr(ctx context.Context, t *testctx.T) {
```
marcosnils, line 582: Is the `TestExecRedirectStdoutStderr` removal intentional?

jedevc: Hm, yeah, it appears further down in the diff which is confusing.

I rewrote some of the tests here to avoid the weird `testutil.Query`. We should just be testing using the SDK.

marcosnils: weeeeeeeeeird. I wasn't seeing it down the diff before.

marcosnils (approved):

docs/docs-graphql/schema.graphqls
```diff
@@ -774,6 +774,8 @@ type Container {
     sure, you don't need this.
     """
     noInit: Boolean = false
+
+    redirectStdin: String = ""
```
helderco, line 778: There's no description.

PR #5158, opened 2023-05-20
Merge exposed ports on From and Build

helderco:
Follow-up to https://github.com/dagger/dagger/issues/5052 to fix `ExposedPorts` support.

## What?

In https://github.com/dagger/dagger/issues/4588 we want to override an existing `Container`’s `Env`, `Labels` and `ExposedPorts` config when using `From` or `Build`. Everything else is completely replaced. The reasoning is that those fields have these corresponding CRUD methods:
| OCI                | Create/Update     | Delete               | List           | 
| ------------------ | ----------------- | -------------------- | -------------- |
| **`Env`**          | `WithEnvVariable` | `WithoutEnvVariable` | `EnvVariables` |
| **`Labels`**       | `WithLabel`       | `WithoutLabel`       | `Labels`       |
| **`ExposedPorts`** | `WithExposedPort` | `WithoutExposedPort` | `ExposedPorts` |

However, merging exposed ports wasn’t added in https://github.com/dagger/dagger/issues/5052:

https://github.com/dagger/dagger/blob/f525a276ff9ded404a630ba66c6b3eefa4b6c87d/core/util.go#L300-L301

This change fixes that.

## How?

It’s now possible to expose a port before `.From` or `.Build`:

```go
httpSrv := client.Container().
    WithExposedPort(8080).
    Build(src).
    WithExec([]string{"python", "-m", "http.server", "8080"})
```

The port will be preserved, just like environment variables and labels.

## Backwards compatibility

Listing with `exposedPorts` now includes ports that may not be used that were already exposed in the image config. This is to be consistent with other fields and reflects what will happen on `.Publish`.

PR #10803, opened 2025-07-28
chore: move service StartBindings to container.WithExec

alexcb:


core/container.go
```diff
@@ -495,21 +495,11 @@ func (container *Container) Build(
 	if err != nil {
 		return nil, err
 	}
-	svcs, err := query.Services(ctx)
```
jedevc, line 501: Hmmmmmm, now getting very suspicious about the `container.Services.Merge` code above. 

Originally, this was because all the results are evaluated lazily - we don't know where `ExecOp`s might be running - so we have to be able to spin all those up.

But now, we know that when `dockerfileDir`/etc are evaluated, they're "self-contained" - they'll start their own services. I wonder if we can drop the service merging logic entirely??

cc @vito on this one, as the author of service logic, just wanna check my own understanding :sweat_smile:

alexcb: Interesting point. I removed a number of them and tests seem to be happy. I left some merges under the `Publish` and `Export` calls, which are still done via buildkit.

jedevc: I suspect we could *even* remove those. I guess the only reason we would need those is if it was possible to `Publish`/`Export` to a remote `Service`, but there's no dagger-api for that right now.

Even when those `Container`s are built under buildkit, they're still evaluating dagops, which should now be starting their own services.

vito: Lovely! Yeah this all sounds about right. It was a huge pain to thread these service dependencies through everywhere, nice to see it isolated.

alexcb: @jedevc Here's an example of a test failing when I removed the code from the `Publish` and `Export` methods: https://dagger.cloud/dagger/traces/4aea1494b656243f631143b8e30122af#b27ba1b808322dde:EL238

Here's the relevant error message:

```

        testctx.go:197: 44  : [1.4s] | Note: Unnecessary use of -X or --request, GET is already inferred.
        testctx.go:197: 44  : [1.4s] | * Host myapp:8080 was resolved.
        testctx.go:197: 44  : [1.4s] | * IPv6: (none)
        testctx.go:197: 44  : [1.4s] | * IPv4: 10.88.1.79
        testctx.go:197: 44  : [1.4s] |   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
        testctx.go:197: 44  : [1.4s] |                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 10.88.1.79:8080...
        testctx.go:197: 44  : [1.4s] | * connect to 10.88.1.79 port 8080 from 10.88.1.84 port 38734 failed: Connection refused
        testctx.go:197: 44  : [1.4s] | * Failed to connect to myapp port 8080 after 4 ms: Could not connect to server
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
        testctx.go:197: 44  : [1.4s] | * closing connection #0
        testctx.go:197: 44  : [1.4s] | curl: (7) Failed to connect to myapp port 8080 after 4 ms: Could not connect to server
        --- FAIL: TestContainer/TestContainerAsService/use_default_args_and_entrypoint_by_default (23.18s)
            testctx.go:173: 
                        Error Trace:    /app/core/integration/container_test.go:4558
                                                                /go/pkg/mod/github.com/dagger/testctx/oteltest@v0.0.3/log.go:37
                                                                /go/pkg/mod/github.com/dagger/testctx/oteltest@v0.0.3/trace.go:83
                        Error:          Received unexpected error:
                                        input: container.from.withExec.withServiceBinding.withExec.stdout process "sh -c curl -vXGET 'http://myapp:8080/hello'" did not complete successfully: exit code: 7
                        Test:           TestContainer/TestContainerAsService/use_default_args_and_entrypoint_by_default
```

It's interesting to note that `Host myapp:8080 was resolved.` succeeded -- so maybe there's some healthcheck related code that's not being used?

To make matters worse, this test passes locally when I run `./hack/dev && ./hack/with-dev go test -count=1 -v -run 'TestContainer/TestContainerAsService' ./core/integration/`

alexcb: > To make matters worse, this test passes locally

There was a race condition -- it turned out that calling `c.Container().Import(binctr.AsTarball())` caused the `container.Ports` to get lost, which prevented a `portHealthChecker` from being created.

core/container.go
```diff
@@ -1732,6 +1672,14 @@ func (container *Container) Import(
...
 
 	container.Config = imgSpec.Config
+	container.Ports = nil
+	for portOCISpec := range container.Config.ExposedPorts {
+		p, err := NewPortFromOCI(portOCISpec)
+		if err != nil {
+			return nil, err
+		}
+		container.Ports = append(container.Ports, p)
+	}
```
jedevc: So actually... I think this is intentional. Doing `Container.from` doesn't import the port config, so we shouldn't do it here.

Looking at the racy test that uses `import`, I actually suspect that we should be calling `WithExposedPort` *after* import.

jedevc:
LGTM, one comment, I actually think the test is broken :laughing: Don't think we need the engine change.

core/container.go
alexcb: That's an interesting point -- I backed out my changes, and figured I would add a test for this:

```
		// create new container with default values
		defaultBin := c.Container().Import(binctr.AsTarball())

		// Exposed ports are intentionally dropped when calling Import (similar to what Container.from does)
		exposedPorts, err := defaultBin.ExposedPorts(ctx)
		require.NoError(t, err)
		require.Empty(t, exposedPorts)
```

and the `require.Empty` test is now failing (with it returning an entry for port 8080).

This is because container.Import sets `container.Config = imgSpec.Config`, then when we call `ExposedPorts(ctx)` it does a range over both `parent.Ports` and `parent.Config.ExposedPorts`.

So what should we do here?

1) set `imgSpec.Config.ExposedPorts` to nil when a call to `Import` is done? This could be done immediately after: https://github.com/dagger/dagger/blob/0026f7da8e5fea15daf626f08f6fce4a607b46d5/core/container.go#L1734

2) update the the core/service.go code that calls `newHealth(bk, gc, fullHost, ctr.Ports).Check(ctx)` to use the union of `ctr.Ports` and `ctr.Config.ExposedPorts`?
here's the current code that only uses `ctr.Ports` https://github.com/dagger/dagger/blob/0026f7da8e5fea15daf626f08f6fce4a607b46d5/core/service.go#L430

3) change `ExposedPorts(ctx)` to only return `ctr.Ports`? currently it does a union under: https://github.com/dagger/dagger/blob/0026f7da8e5fea15daf626f08f6fce4a607b46d5/core/schema/container.go#L2536-L2565

4) remove the `require.Empty` check I attempted to add?

5) something else?

alexcb: > Doing Container.from doesn't import the port config

I gave this a test, and it seems like it does import some of it:

```
		defaultBin := c.Container().From("nginx:1.29.0")

		exposedPorts, err := defaultBin.ExposedPorts(ctx)
		require.NoError(t, err)
		require.Len(t, exposedPorts, 1)

		port, err := exposedPorts[0].Port(ctx)
		require.NoError(t, err)
		require.Equal(t, port, 80)

		output, err := curlctr.
			WithServiceBinding("myapp", defaultBin.AsService()).
			WithExec([]string{"sh", "-c", "curl -v 'http://myapp:80'"}).
			Stdout(ctx)
		require.NoError(t, err)
		require.Contains(t, output, "Welcome to nginx")
```

I added a printf, and confirmed that `container.Ports` is never set, but `container.Config.ExposedPorts` gets set -- which explains why the test above passes. Ultimately does this make sense to have this test pass?

sipsma: I feel like a mismatch between `container.Ports` and `container.Config.ExposedPorts` is *probably* accidental? But in the interest of back-compat it's probably easier to just leave it as it was for now and fix that separately, IMO

alexcb: sounds good -- I'll remove my `require.Empty(t, exposedPorts)` check, and open a separate issue for it.

vito: @sipsma @alexcb Context on `container.Ports` vs. `container.Config.ExposedPorts` is here: https://github.com/dagger/dagger/pull/5158

It's to deal with the unfortunate reality that many `Dockerfile`s have `EXPOSE` lines for ports that are actually _suggestions_ and aren't actually bound by the service all the time. :man_facepalming: 

This is the second time coming back to this and going "wtf?" though so I'm trending towards "people who do that are wrong and anyway who cares about Dockerfiles" - but last time we made a decision like that (not respecting `ENTRYPOINT`) it didn't go well. :sob:

alexcb: Thanks for those details @vito , it clears things up, somewhat :laughing:

I added a big comment in the test explaining the difference with a potential TODO in case we ever re-evaluate this decision, but for now it's the behaviour we want and I'm more than happy to leave it as-is :)

jedevc (approved):
Refactoring go brrrrr 🧹🧹

PR #10810, opened 2025-07-29
feat: add host.container to import containers from the host

### LINKED ISSUE #10737: ✨ Import image from local container runtime

jedevc:
Fixes #10737 as a follow-up to #10662.

TODO:
- [ ] Tests passing
- [ ] Cleanup
  - Move more implementation to `core/`


guilhermedelyra:
hi! what's the status on this? 😁
looking forward on using this feature

core/schema/host.go
```diff
@@ -196,6 +197,12 @@ func (s *hostSchema) Install(srv *dagql.Server) {
 				dagql.Arg("host").Doc(`Upstream host to forward traffic to.`),
 			),
 
+		dagql.NodeFuncWithCacheKey("container", s.container, dagql.CachePerClient).
+			Doc(`Accesses a container on the host.`).
+			Args(
+				dagql.Arg("name").Doc(`Name of the container to access.`),
```
helderco: Shouldn't this be name of the image instead?

shykes:
@jedevc how do you differentiate between "container instance" and "container image" on the host system?

jedevc:
This is probably a naming issue, this PR only lets you import *images*, not running containers.

I could call it `host.Image`? That might be clearer tbf, and matches `exportImage`.

shykes:
> This is probably a naming issue, this PR only lets you import _images_, not running containers.
> 
> I could call it `host.Image`? That might be clearer tbf, and matches `exportImage`.

IMO `Host.ContainerImage` would be more explicit. Also leaves the door open for `Host.ContainerInstance` later? 😁

shykes:
Alternatively, you can keep `Host.Container()` (it's nice and short), and in the future we could add a fallback to lookup a container instance by name, if no image exists. The risk of conflict is pretty low between docker container instances & images... Maybe not worth polluting our API with the distinction.

jedevc:
I don't *love* `ContainerImage` - explicit yes, but we don't really have that same language anywhere else.

WDYT about having `Host.image: Container` and maybe one day later `Host.container: Container` (for the instance).

shykes:
> I don't _love_ `ContainerImage` - explicit yes, but we don't really have that same language anywhere else.

It's normal that we don't have that language anywhere else: `Host.<foo>` uses the host's vocabulary. So we should go for what is the most familiar to the user, *in the context of the host system*. In that context, the fact that Dagger doesn't strongly differentiate between container instance and image is irrelevant: it's the host's house so it's their rules, so to speak.

> WDYT about having `Host.image: Container` and maybe one day later `Host.container: Container` (for the instance).

I think `Image` is too vague in the context of the host system. For example, is it a disk image of the whole host system? Etc.

jedevc:
Renamed from `host.container` to `host.containerImage`.

jedevc:
If anyone has spare review time, don't think this one should be too hard!

engine/client/imageload/apple.go
```diff
@@ -48,3 +49,17 @@ func (loader Apple) loadTarball(ctx context.Context, name string, tarball io.Rea
 
 	return nil
 }
+
+func (loader Apple) saveTarball(ctx context.Context, name string, tarball io.Writer) (rerr error) {
+	ctx, span := otel.Tracer("").Start(ctx, "save "+name)
+	defer telemetry.End(span, func() error { return rerr })
+
+	cmd := exec.CommandContext(ctx, "container", "image", "save", name)
```
vito, line 57: oh nice, was wondering what things this would be compatible with, glad it's extensible

jedevc:
Thanks for the approve :pray: 

I'll merge for now, I'm not super in love with `containerImage` either - happy to continue the discussion and update it if we come up with a better option before the next release v0.18.17 next week.

I don't really like `Container.fromHostImage`, keeping it on host feels important. `Host.containerFromImage` is quite wordy, so I don't love that either :laughing:

PR #10819, opened 2025-07-30
Dagger Shell: Allow using flags for required args

### LINKED ISSUE #9701: Shell: Allow using flags for required args
Using positional args in shell is great!

However:
- It is harder to demo because people don't know what the positional args are for
- For the same reason, Dagger executions in scripts (like in a CI yaml) will prefer to use `dagger call`
- This makes it less clear when we show "see the same thing we ran locally runs in CI" if that then has to be translated from shell to call

I'm proposing the option to use flags for required args in shell to avoid the split between call vs shell. This will make it reasonable to use `shell -c` in scripted environments like CI.

Example:
Dagger Call: `- dagger call build --aaa 111 --bbb 222 --ccc 333 test out`
Current option for shell: `- dagger shell -c "build 111 222 333 | test | out"`
Proposed option: `- dagger shell -c "build --aaa 111 --bbb 222 --ccc 333 | test | out"`

jpadams:
fixes #9701 

With this PR you can use a  flag for any arg (e.g. `--argname`) including required args where previously Dagger Shell would strictly require a positional arg WITHOUT a flag. Use all the flags or just the ones you want for clarity.

I was inspired to do this because today users typically use `dagger call` for CI since the flags make the call more self-documenting. In the case that someone is a Dagger Shell maximalist who wants to use Shell from the command line to CI, the ability to use flags means their `dagger -c` one-liners in CI will be just as clear as a `dagger call`.

Any args without flags should act as positional args to fill in the remaining ordered positional slots.

- [x] more manual testing

- docs after PR merged because we want to ship this, get feedback and incorporate into new docs
- perhaps make error messages better regarding required/positional in a followup
- perhaps more tests in a followup if needed
- After merging this PR, will look at idea of autocompletion of flags

### DIFF
--- a/core/integration/module_shell_test.go
+++ b/core/integration/module_shell_test.go
@@ -1285,7 +1285,7 @@ func (ShellSuite) TestNamedArguments(ctx context.Context, t *testctx.T) {
 	t.Run("mixed positional and named arguments", func(ctx context.Context, t *testctx.T) {
 		// Test mixing positional and named arguments (positional first)
 		out, err := daggerCliBase(t, c).
-			With(daggerShell(`directory | with-file /test --contents="hello world" | file /test | contents`)).
+			With(daggerShell(`directory | with-new-file /test --contents="hello world" | file /test | contents`)).
 			Stdout(ctx)
 		require.NoError(t, err)
 		require.Equal(t, "hello world", out)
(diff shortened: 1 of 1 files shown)


TomChv (commented):
Code looks okay to me, I reran the CI to make sure tests pass.

If it's all good, I let @helderco or @shykes give a final approval to merge it or close the PR

jedevc:
@shykes before this auto-closes after going stale, are we interested in this feature?

Or happy to just let it drop?

shykes:
People have definitely asked for it... I'm just concerned about breaking changes. Can both styles coexist? Or are there weird edge cases?

jedevc:
It shouldn't be a breaking change - we'd just be allowing a new pattern, not removing an old one.

Both styles should be able to coexist I think. Potentially this PR just works? But probably needs a bit more investigation.

jpadams:
In my testing, it "just works" and is pleasant and predictable to use.

shykes (approved):
OK let's do it! I agree consistency would be good. My only concern was breaking changes.

jedevc:
Don't merge as-is! The added tests are failing, asked in #engine-dev if anyone can take this :tada:

TomChv:
Hey,

I'm taking care of that, it's mostly just fixing typos

EDIT: test fixed and PR rebased, I'm waiting for a green CI to merge :)

PR #10822, opened 2025-07-31
add service

fixeasy:
# Add Kubernetes `Service` for Dagger Engine

## Context

We operate a large-scale software factory supporting key business units (DSM/DSI).  
To enhance our CI capabilities and accelerate delivery, we are deploying **Dagger** as our new CI solution.

Our setup:  
- Dagger runs as a **DaemonSet** in Kubernetes  
- Shared Dagger modules live in GitLab  
- Projects use these modules via GitLab CI pipelines

This aims to provide a scalable and reliable CI platform aligned with our business objectives.

## Current Limitation

The current Helm chart exposes Dagger **only via a NodePort configured directly on the pods** (`DaemonSet` or `StatefulSet`).  
As a result, each CI pipeline must explicitly target a specific node IP, for example:

```yaml
_EXPERIMENTAL_DAGGER_RUNNER_HOST: "tcp://<node_IP>:<dagger-port>"
```
This approach:
- Strongly couples the CI to the cluster's topology
- Prevents load distribution across nodes
- Reduces resilience, e.g., if the targeted node is restarted

## ✅ Solution

This PR introduces:
- A **Kubernetes Service** to expose the `dagger-engine` pod :
```yaml
_EXPERIMENTAL_DAGGER_RUNNER_HOST: "tcp://dagger-engine-service.dagger.svc.cluster.local:<dagger-port>"
```
- **Session affinity (`sessionAffinity: ClientIP`)** to ensure that all requests from a pipeline are routed to the **same pod**, maintaining execution consistency.

## Changes
- Added `engine.service.*` configuration in `values.yaml`
- Created new Helm template: `engine-service.yaml`
- Added `app:` label to pods so the Service selector can match them
- Modified pod templates to skip `hostPort` configuration if `service.enabled` is `true`

## Goal

This change enables a more scalable, resilient, and Kubernetes-native integration of Dagger, and removes the need to target specific nodes from the CI.

### DIFF
--- a/helm/dagger/templates/engine-daemonset.yaml
+++ b/helm/dagger/templates/engine-daemonset.yaml
@@ -27,6 +27,9 @@ spec:
       {{- end }}
       labels:
         name: {{ include "dagger.fullname" . }}-engine
+        {{- if .Values.engine.service.enabled }}
+        app: {{ include "dagger.fullname" . }}-engine
+        {{- end }}
         {{- include "dagger.labels" . | nindent 8 }}
     spec:
       {{- if .Values.engine.runtimeClassName }}
@@ -100,7 +103,7 @@ spec:
             - name: dagger
               containerPort: {{ .Values.engine.port }}
               protocol: TCP
-              {{- if .Values.engine.hostPort }}
+              {{- if and .Values.engine.hostPort (not .Values.engine.service.enabled) }}
               hostPort: {{ .Values.engine.hostPort }}
               {{- end }}
           {{- end }}
--- a/helm/dagger/templates/engine-service.yaml
+++ b/helm/dagger/templates/engine-service.yaml
@@ -1,0 +1,26 @@
+{{- if .Values.engine.service.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "dagger.fullname" . }}-engine-service
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "dagger.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.engine.service.type }}
+  selector:
+    app: {{ include "dagger.fullname" . }}-engine
+  ports:
+    - protocol: TCP
+      port: {{ .Values.engine.port }}
+      targetPort: {{ .Values.engine.port }}
+      {{- if and (eq .Values.engine.service.type "NodePort") .Values.engine.service.nodePort }}
+      nodePort: {{ .Values.engine.service.nodePort }}
+      {{- end }}
+  {{- if .Values.engine.service.sessionAffinity.enabled }}
+  sessionAffinity: ClientIP
+  sessionAffinityConfig:
+    clientIP:
+      timeoutSeconds: {{ .Values.engine.service.sessionAffinity.sessionAffinitytimeoutSeconds }}
+  {{- end }}
+{{- end }}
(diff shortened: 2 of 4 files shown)


fixeasy:
Just a small bump to keep this active. This PR is still relevant.

TomChv (commented):
Changes on the chart looks good to me, however I'm not sure about the effectiveness of that change.

@matipan will confirm if I say a mistake but afaik each pod has its own cache volume so it's not distributed for now.
That means to benefit from cache on every run (not just inside the same pipeline), you'll want to call the same node.

The sessionAffinity you added will not work if your clientIP change (e.g., if you're using ephemeral GH runner to run your CI).

Right now you have to manually select the engine you want to call from your pool, but with that new service it's bit different since on each new clientIP, it will distribute to a random engine that may not have any cache loaded for your pipeline.

Is there a solution for that? Can we still explicitely call a specific engine? 

If the cache volume is shared across nodes, then you can ignore my comment; if not, I think we should solve that problem before merging.

fixeasy:
@TomChv 

Thanks for your feedback! 

We already identified this cache sharing issue.
We tried to use a shared volume between nodes, but it failed. This is technically not possible today because **Dagger puts a lock on `/var/lib/dagger`**. Only one instance can use this volume, all others fail to start.

So we are forced to use **one volume per Dagger instance**, and we decided to stay on **`hostPath`**.

The main problem we want to solve with this Service is **node scaling**.
When the Dagger engine is under high load, we add new nodes to absorb and spread the traffic.
If we configure in GitLab CI:

```yaml
_EXPERIMENTAL_DAGGER_RUNNER_HOST: "tcp://<node_IP>:<dagger-port>"
```
- node IP may change
- scaling becomes useless, because we always target the same node.

This is why we need a Kubernetes Service to expose:
```yaml
_EXPERIMENTAL_DAGGER_RUNNER_HOST: "tcp://dagger-engine-service.dagger.svc.cluster.local:<dagger-port>"
```

This way, all Dagger instances on all nodes can be used.

We know that cache efficiency will be lower, but for now we don’t have another option.

Finally, about the `sessionAffinity`:
In our case, each gitlab-ci pipeline creates a runner through the runner manager inside Kubernetes.
This runner keeps the **same IP during the whole pipeline lifecycle**, so it always talks to the same Dagger engine.
Without this configuration, requests from the same runner would be sent to different engines, and since the execution context is not shared, the pipeline would fail.

That’s why it is necessary to configure the Service with this field.

TomChv:
@fixeasy Thanks for your explanations! 

I asked the team for a 2nd look just to be sure and we'll merge if it's all good 🚀

matipan:
Hey @fixeasy. Thanks for sending this patch over!

`sessionAffinity: ClientIP` can be a brittle mechanism. It relies heavily on the client's source IP being preserved all the way to the Kubernetes service. In many common production and even local setups (like Docker for Mac, some service meshes, or environments with L7 load balancers), the source IP is lost or masqueraded.

When that happens, the session affinity breaks down, and requests get load-balanced across different engine pods. Since Dagger sessions are stateful, this would cause builds to fail intermittently and would be very difficult for a user to debug. Our philosophy for the official chart is to avoid configuration options that can fail silently based on external infrastructure.

The good news is that the current chart is flexible enough to support this. You can currently configure both custom `labels` as well as which `port` dagger listens on. You can use those two properties and creating your own `Service` that points to the pods using the previously defined `label`.

We'll be closing this PR, but we really appreciate you thinking about how to improve the chart. Thanks again!

PR #10825, opened 2025-07-31
Update python packages

helderco:
* Updated default uv version to 0.8
* Changed default build backend to uv_build instead of hatchling now that it's stable
* Changed default Python version to 3.13


jedevc (approved):
LGTM :tada: Do users need to do anything specific to their modules to upgrade? Or should it just all be automatic?

helderco:
> Do users need to do anything specific to their modules to upgrade? Or should it just all be automatic?

Only modules created before [v0.13.0](https://github.com/dagger/dagger/releases/tag/sdk/python/v0.13.0), if they need to stick to a lower version. After that release, new modules got the `requires-python = ">=3.12"` option which means they'll stay on that version even after this change. Only new modules will get the new default.

I'll add a note on the change log. 👍

PR #10828, opened 2025-08-01
Feat/rust/correct error handling

kjuulh:
A small feature, this basically unwraps the error exec to unwrap the error such that we show more clearly what fails and whatnot.

### DIFF
--- a/sdk/rust/crates/dagger-sdk/src/core/gql_client.rs
+++ b/sdk/rust/crates/dagger-sdk/src/core/gql_client.rs
@@ -16,9 +16,22 @@ pub struct GraphQLError {
 #[allow(dead_code)]
 pub struct GraphQLErrorMessage {
     pub message: String,
-    locations: Option<Vec<GraphQLErrorLocation>>,
-    extensions: Option<HashMap<String, String>>,
-    path: Option<Vec<GraphQLErrorPathParam>>,
+    pub locations: Option<Vec<GraphQLErrorLocation>>,
+    pub extensions: Option<GraphQLExtensions>,
+    pub path: Option<Vec<GraphQLErrorPathParam>>,
+}
+
+#[derive(Deserialize, Debug, Clone)]
+#[serde(tag = "_type")]
+pub enum GraphQLExtensions {
+    #[serde(rename = "EXEC_ERROR")]
+    ExecError {
+        cmd: Vec<String>,
+        #[serde(rename = "exitCode")]
+        exit_code: Option<i16>,
+        stderr: Option<String>,
+        stdout: Option<String>,
+    },
 }
 
 #[derive(Deserialize, Debug, Clone)]
@@ -51,7 +64,7 @@ impl GraphQLError {
     }
 
     pub fn with_json(json: Vec<GraphQLErrorMessage>) -> Self {
-        Self::with_message_and_json("Look at json field for more details", json)
+        Self::with_message_and_json("request to dagger failed", json)
     }
 
     pub fn message(&self) -> &str {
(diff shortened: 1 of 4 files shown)


kjuulh:
> Code SGTM, but also pinging @vito for a review, since I can't remember what subtlety we try for to make errors readable in traces (and where/when we print them).

Sounds good. Yes, I actually think currently the Rust sdk is an outlier as it produces horrible errors on failure. Because i doesn't understand how to parse the embedded EXEC_ERROR in the graphql response.

Let me know what you usually do and I'll bring it in line with the other sdks.

vito:
(edit: sorry, misclicked the close button)

Hey, thanks for looking into this!

The state of the art here is to turn GraphQL extended errors into a proper `dag.CurrentFunctionCall().ReturnError(...)` call, passing a `dagger.Error` that has all extended attributes preserved.

Here's a recent example of adding this to the TypeScript SDK: https://github.com/dagger/dagger/pull/10459

```typescript
function formatError(e: unknown): DaggerError {
  if (e instanceof Error) {
    let error = dag.error(e.message)
    // If the error is an instance of GraphQLError or a inherit type of it,
    // we can add the extensions fields in the error object.
    if (e instanceof ExecError || e instanceof GraphQLRequestError) {
      Object.entries(e.extensions ?? []).forEach(([key, value]) => {
        if (value !== "" && value !== undefined && value !== null) {
          error = error.withValue(key, JSON.stringify(value) as JSON)
        }
      })
    }
    return error
  }
```

Or, in Go:

```go
	var gqlErr *gqlerror.Error
	if errors.As(rerr, &gqlErr) {
		if gqlErr.Extensions != nil {
				keys = append(keys, k)
			}
			sort.Strings(keys)
			for _, k := range keys {
				if err != nil {
				}
			}
		}
		return dagErr
	}
}
```

Doing it this way will not only preserve the `ExecError` data for the caller but also propagate the error origin (OTel trace context) and any other metadata we might add in the future.

kjuulh:
> (edit: sorry, misclicked the close button)
> 
> Hey, thanks for looking into this!
> 
> The state of the art here is to turn GraphQL extended errors into a proper `dag.CurrentFunctionCall().ReturnError(...)` call, passing a `dagger.Error` that has all extended attributes preserved.
> 
> Here's a recent example of adding this to the TypeScript SDK: #10459
> 
> ```ts
> function formatError(e: unknown): DaggerError {
>   if (e instanceof Error) {
>     let error = dag.error(e.message)
>     // If the error is an instance of GraphQLError or a inherit type of it,
>     // we can add the extensions fields in the error object.
>     if (e instanceof ExecError || e instanceof GraphQLRequestError) {
>       Object.entries(e.extensions ?? []).forEach(([key, value]) => {
>         if (value !== "" && value !== undefined && value !== null) {
>           error = error.withValue(key, JSON.stringify(value) as JSON)
>         }
>       })
>     }
>     return error
>   }
> ```
> 
> Or, in Go:
> 
> ```go
> func convertError(rerr error) *dagger.Error {
> 	var gqlErr *gqlerror.Error
> 	if errors.As(rerr, &gqlErr) {
> 		dagErr := dag.Error(gqlErr.Message)
> 		if gqlErr.Extensions != nil {
> 			keys := make([]string, 0, len(gqlErr.Extensions))
> 			for k := range gqlErr.Extensions {
> 				keys = append(keys, k)
> 			}
> 			sort.Strings(keys)
> 			for _, k := range keys {
> 				val, err := json.Marshal(gqlErr.Extensions[k])
> 				if err != nil {
> 					fmt.Println("failed to marshal error value:", err)
> 				}
> 				dagErr = dagErr.WithValue(k, dagger.JSON(val))
> 			}
> 		}
> 		return dagErr
> 	}
> 	return dag.Error(rerr.Error())
> }
> ```
> 
> Doing it this way will not only preserve the `ExecError` data for the caller but also propagate the error origin (OTel trace context) and any other metadata we might add in the future.

Awesome Vito, thanks for the detailed reply. I'll take a look at it hopefully in the coming days. I don't think I need to change much, except maybe some error printing. But thankfully Rust will actually preserve the full context, using the :? option, rather than just normally printing the error which will give the nicer message.

TomChv:
I'm closing this PR since there's no activity since August 4.
Feel free to reopen it whenever you have time to apply: https://github.com/dagger/dagger/pull/10828#issuecomment-3151061728 :)

TomChv:
I'm closing this PR since there's no activity since August 4.
Feel free to reopen it whenever you have time to apply: https://github.com/dagger/dagger/pull/10828#issuecomment-3151061728 :)

vito:
Won't be needed anymore with https://github.com/dagger/dagger/pull/11434

Issue #8034, opened 2024-07-25
✨ Optional stripping of trailing newline when passing secrets


yann-soubeyrand:
### What are you trying to do?

I try to pass a GitHub token as a secret to a Dagger function which uses it as an environment variable in a container. For this, I use the `cmd:gh auth token` syntax. The problem is that the `gh auth token` command issues a trailing newline which is not stripped and appears in the environment variable value. The command using this environment variable doesn’t work then. I’d like to have a way to tell Dagger to strip the trailing newline.

### Why is this important to you?

Errors which occur due to the trailing newline may not be obvious to debug and having a documented syntax which works in all the situations to strip this newline could improve the situation.

### How are you currently working around this?

Currently, I pass the token like this `cmd:echo -n $(gh auth token)` to strip the trailing newline.

yann-soubeyrand:
On Discord, there have been some proposals:

- We could add an argument when passing a secret, like `cmd:command-to-run:s`.
- We could use a different scheme, like `cmd-s:command-to-run`.
- We could add an annotation on the function parameter, like in Go
```
func (*Module) Function(
// +strip
secret *dagger.Secret
)
```

yann-soubeyrand:
If feasible, I personally prefer the last one.

helderco:
Related:

> I'm good with reverting the change to strip newlines. I think we could consider the inverse of my suggestion on that PR and instead support a CLI syntax that _does_ strip newlines for use cases that call for that. But I think this situation proves it's probably better to default to the exact contents of the file/command/env var.

\- _Originally posted by @sipsma in https://github.com/dagger/dagger/issues/7220#issuecomment-2094405767_

helderco:
> If feasible, I personally prefer the last one.

I don't think that's a good option. The function shouldn't need to know if the secret needs stripping or not. That responsibility should fall to the one loading the secret. See:

> I don't think we should do this with options at the point of use. Secrets should be marked to do this at point of load.

I also prefer @sipsma's suggestion here:

> Hm yeah I agree this is likely what users would want 99.99% of the time and fairly annoying to require them to deal with themselves in userspace (i.e. get secret plaintext, strip newlines, create new modified secret).
> 
> I feel like there _must_ be some obscure corner case where this isn't desired, but I can't think of any concretely either. And push comes to shove if we need to support the 0.01% of cases we can always add some kludge like `--token=file-raw:/some/token/file/where/trailing/newlines/matter` or similar.

[...]

PR #10836, opened 2025-08-04
feat: add stripped variants for some secret providers

jedevc:
Fixes https://github.com/dagger/dagger/issues/8034.

Hit this, decided to do `cmd+strip`, which is a fairly standard way of doing "protocol variations".

In the future, we could switch the default, but decided to pass on that change for now.

### DIFF
--- a/.changes/unreleased/Added-20250804-141150.yaml
+++ b/.changes/unreleased/Added-20250804-141150.yaml
@@ -1,0 +1,10 @@
+kind: Added
+body: |-
+    New `file+strip`, `env+strip` and `cmd+strip` secret providers
+
+    These new secret providers are like the base providers, but additionally strip
+    any whitespace from the output.
+time: 2025-08-04T14:11:50.660055633+01:00
+custom:
+    Author: jedevc
+    PR: "10836"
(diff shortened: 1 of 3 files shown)


marcosnils (approved):
SGTM. We should not forget to mention this in our docs under the `Secrets` section. cc @vikram-dagger

jedevc:
cc @shykes does the API look reasonable?

shykes:
This seems more complicated than just stripping by default. What's the benefit of the extra complexity?

jedevc:
There's a good summary of why it should be opt in at https://github.com/dagger/dagger/issues/8034#issuecomment-2249980056.

I seem to remember we stripped secrets by default - and got issues like https://github.com/dagger/dagger/issues/7220, for SSH keys in particular, the trailing newline at the end of the file *is* significant :shrug:

shykes:
Yes to be clear, I don't think we should strip all secrets by default. I think we should specifically strip the output of commands (`cmd://`) by default.

- Files and env variables: controlled directly by the user, so reflects their exact intent. Easy for them to change if needed. Err on the side of verbatim.
- Commands: not controlled directly by the user, may not reflect their exact intent. Harder for them to change. Err on the side of default convenience.

jedevc:
Fair, I get that. I do think it's a bit confusing that we'd get different behavior if you did `file:///path/to/file` and `cmd://cat /path/to/file` though. It's also a breaking change, so we'd probably want to wait for v0.19.0 if we do that.

Also regardless of what the defaults are, there should be a way to indicate to strip/not-strip.

PR #10839, opened 2025-08-04
chore: update releasing instructions to call out external contributors

jedevc:
This is just a nice thing to do :tada:

We work out which contributors are external by looking at whether they're in a specific github team - any members of the `maintainers` tag or `sdk-` tag are treated as external.


helm/dagger/.changie.yaml
```diff
@@ -33,6 +33,34 @@ kinds:
...
+    {{- range $author := $authorList }}
+      {{- if not (has $author $maintainers) }}
+        {{- $contributorSet = set $contributorSet $author true }}
+      {{- end }}
+    {{- end }}
+  {{- end }}
+  {{- if $contributorSet -}}
+  ### Contributors
+  Special thanks to our external contributors this release!
+  {{- $contributors := list }}
```
marcosnils: nit: keys $contributorSet  | sortAlpha ?

marcosnils: thinking it twice, why is `contributorSet` a dict in the first place? IIUC a list should be good here?

jedevc: We need to remove all of the duplicates :eyes:

marcosnils: woot! 🙌. Gotcha, my initial nit comment is still relevant, isn't it?

jedevc: Yeah, I updated to take your suggestion :tada:

Issue #10841, opened 2025-08-04
Shell mode cannot pass Service objects to functions expecting ServiceID


grouville:
# Shell Mode Service Object Handling Issue

## Problem Statement

The Dagger shell mode cannot pass Service objects to functions that accept `ServiceID` parameters, specifically affecting `git --experimental-service-host` and `http --experimental-service-host`.

### API Definition

The git function expects a Service object ID as defined in:

**`core/schema/git.go:148`**
```go
type gitArgs struct {
    URL                     string
    KeepGitDir              dagql.Optional[dagql.Boolean] `default:"false"`
    ExperimentalServiceHost dagql.Optional[core.ServiceID]  // <-- Expects ServiceID
    // ...
}
```

### Shell Mode Flag Parsing Issue

When parsing flags in shell mode, the system uses a `serviceValue` type that expects URL strings:

**`cmd/dagger/flags.go:506-513`**
```go
func (v *serviceValue) Set(s string) error {
    if s == "" {
        return fmt.Errorf("service address cannot be empty")
    }
    u, err := url.Parse(s)  // <-- Expects URL format
    if err != nil {
        return err
    }
    // Creates NEW service from URL...
}
```

This creates a fundamental mismatch: the API expects an existing Service object ID, but the flag parser only knows how to create new services from URLs.

### Object ID Bypass Logic

The shell has logic to handle object IDs in `cmd/dagger/shell_exec.go:774-818`, specifically:

```go
func (h *shellCallHandler) parseFlagValue(ctx context.Context, value string, arg *modFunctionArg) (rval string, bypass bool, rerr error) {
    // ... 
    if r.IsObject() {  // Line 809
        id, ok := r.Value.(string)
        if !ok {
            return id, true, fmt.Errorf("expected an object ID string, got %T", r.Value)
        }
        return id, true, nil  // <-- Should bypass flag parsing
    }
    // ...
}
```

However, this bypass only works if the flag type supports it. The current implementation at `cmd/dagger/flags.go:663-674` adds a hidden string flag for unsupported types but doesn't properly handle the Service case.

## Reproduction

```bash
# FAILS: Service ID parsed as URL
$ export GIT_TOKEN="testtoken"
$ dagger -c 'jo=$(host | service "80:8080") ; git "http://gitserver/parent.git" --experimental-service-host $jo --http-auth-username oauth2 --http-auth-token env:GIT_TOKEN | head | tree'
[...]

PR #10845, opened 2025-08-05
shell: Fix stdlib commands incorrectly pre-resolving state in arguments

helderco:
Fixes https://github.com/dagger/dagger/issues/10841

## Affected commands

Affected functions should only be `git` and `http` if using an object passed as a variable or direct command substitution (`$(...)`) because from the list of commands in `.stdlib` these are the only ones that accept object arguments:

  - cache-volume
  - container
  - directory
  - engine
  - env
  - file
  - **git**
  - host
  - **http**
  - llm
  - secret
  - set-secret
  - version

Chained functions after `git` or `http` should not be affected.

## Workaround

Until this is released, can use `.core | git` instead of just `git`.

### DIFF
--- a/.changes/unreleased/Fixed-20250807-180016.yaml
+++ b/.changes/unreleased/Fixed-20250807-180016.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: 'shell: Fixed passing objects on `http` and `git` stdlib commands'
+time: 2025-08-07T18:00:16.277293Z
+custom:
+    Author: helderco
+    PR: "10845"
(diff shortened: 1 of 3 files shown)


grouville:
Works, thank you ! ~~There is indeed a gap regarding the hostname from git's `--experimental-service-host` API, but unrelated to this issue~~ By relying on the `endpoint`, I can now access:

```shell
dagger_dev -c 'jo=$(host | service "80:8080") ; to=$($jo | endpoint);  git "http://$to/parent.git" --experimental-service-host $jo --http-auth-username oauth2 --http-auth-token env:GIT_TOKEN | head | tree'
```

grouville (approved):
🙏

PR #10846, opened 2025-08-05
feat: typescript sdk remote lib in module

TomChv:
Add the capability to use the remote dagger library on the Typescript SDK.

This make possible to commit the generated client bindings and simplify overall module setup.
For backward compatibility and consistency across the various setup (local, bundle, remote), I used the import map to rename the remote dependency `@dagger.io/core` so the user still import the client binding with `@dagger.io/dagger`.

### Setup remote module on Node/Bun

```shell
yarn init // bun init
yarn add @dagger.io/dagger // or bun install @dagger.io/dagger
dagger init --sdk=typescript --name=example --source=.
```

This will produce the following import aliases:

```json
{
    "paths": {
      "@dagger.io/dagger": [
        "./sdk/index.ts"
      ],
      "@dagger.io/dagger/telemetry": [
        "./sdk/telemetry.ts"
      ],
      "@dagger.io/core": [
        "./node_modules/@dagger.io/dagger/dist/src/index"
      ]
    }
}
```

So the exported `sdk/index.ts` will rely on `@dagger.io/core` and the generated bindinds:

```ts
export { ... } from "@dagger.io/core"

export type { ConnectOpts, CallbackFct } from "@dagger.io/core"

export * from "./client.gen.js"
```

### Setup remote module on Deno

```shell
deno init
deno install npm:@dagger.io/dagger
dagger init --sdk=typescript --name=example --source=.
```

This will produce the following imports on the `deno.json`

```json
  "imports": {
    "@dagger.io/dagger": "./sdk/index.ts",
    "typescript": "npm:typescript@^5.8.2",
    "@dagger.io/dagger/telemetry": "./sdk/telemetry.ts",
    "@dagger.io/core": "npm:@dagger.io/dagger"
  },
```

Since import alias and actual import maps are merged in Deno, we have the alias aside of the actual package, it's a bit tricky but it works fine ;)

### Side notes

- This also fixes the import inconsistency on generated client, I added the same alias system so now the user can import the client using `@dagger.io/dagger` instead of `@dagger.io/client`
- Remove unused `__tsclientconfig.updator.ts` 
- Add `remote` option to codegen binary, I'll unify bundle/remote in a single flag to avoid duplication after I finish the benchmarks.

### Benchmarks

Screenshot 2025-08-28 at 14 03 00

We see that it can speed up a bit the module loading and overall it's similar to our current performances so no lose on this side.

I also wanted to test the code execution, to see if there's a difference between executing the library from bundled code or from the node_modules directory

Screenshot 2025-08-28 at 14 06 10

We see that there's not much difference except for the bun no cache case but this may be caused by my internet connection being sometime unstable. When looking at the trace, I noticed that the image downloading took 2.8s where normally it takes 0.9 to 1.2s.

### Conclusion

Overall, using the remote SDK library doesn't impact performances and add the convenience to publish the generated client, which may lead to some optimization like not regenerating the client if the call is trigger from a CI, since it can be commited.
It may be interesting as a config in the `dagger.json` or a flag from the CLI.

### DIFF
--- a/cmd/codegen/main.go
+++ b/cmd/codegen/main.go
@@ -26,7 +26,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&outputDir, "output", "o", ".", "output directory")
 	rootCmd.PersistentFlags().StringVar(&introspectionJSONPath, "introspection-json-path", "", "optional path to file containing pre-computed graphql introspection JSON")
 	rootCmd.PersistentFlags().BoolVar(&bundle, "bundle", false, "generate the client in bundle mode")
-	rootCmd.PersistentFlags().BoolVar(&remote, "remote", false, "generate the client using the remote sdk")
+	rootCmd.PersistentFlags().BoolVar(&remote, "remote", false, "generate the client using the published client library")
 }
 
 func main() {


TomChv:
PR green, I'm not sure what's the strategy to test that since it relies on the published library, but maybe I can pin a specific version.
This would also help to ensure backward compatibility.

I plan to do some benchmarks first then, will do that tomorrow.

TomChv:
One thing before we merge that: how do we properly test that remote library?

cmd/codegen/main.go
```diff
@@ -26,6 +26,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&outputDir, "output", "o", ".", "output directory")
 	rootCmd.PersistentFlags().StringVar(&introspectionJSONPath, "introspection-json-path", "", "optional path to file containing pre-computed graphql introspection JSON")
 	rootCmd.PersistentFlags().BoolVar(&bundle, "bundle", false, "generate the client in bundle mode")
+	rootCmd.PersistentFlags().BoolVar(&remote, "remote", false, "generate the client using the remote sdk")
```
helderco:
```suggestion
	rootCmd.PersistentFlags().BoolVar(&remote, "remote", false, "generate the client using the published client library")
```

helderco (commented):
Need to figure out testing. You can prepare the library and get it released first, then make the necessary changes in the sdk module after. We also have ways to mock a specific dagger version in our test suite, although I'm not sure if we have any test case that mocks the version in the way you need to here.

I've done this in Python already, and used `dag.Version()` too to start, but when I started thinking more seriously about testing use cases I found cracks where the version handling needs to be more robust, so I reverted the change to get ready for release.

TomChv:
> Need to figure out testing. You can prepare the library and get it released first, then make the necessary changes in the sdk module after. We also have ways to mock a specific dagger version in our test suite, although I'm not sure if we have any test case that mocks the version in the way you need to here.

In theory, the bundle or local library is exactly the same as the published one so the only thing to test is that we correctly pick up the library version based on the engine.

The important thing to test is that:
- `tsconfig.json` is properly configured to handle the remote library
- generated client correctly uses the published version
- static files (`index.ts` etc..) are correctly exporting the generated client and library function.

So we could lock a specific engine version and do the test on it, this is not likely to changes with time.

> I've done this in Python already, and used dag.Version() too to start, but when I started thinking more seriously about testing use cases I found cracks where the version handling needs to be more robust, so I reverted the change to get ready for release.

So what are our solutions? Ideally we could create in the tests a custom registry, publish library and use that in the integration tests so we have the right behaviour.

https://verdaccio.org/ may be a solution for that?

marcosnils:
We spoke with Tom and we agreed on closing this PR since it's only a typescript specific thing that it's not addressing any user specific request. It's also bringing fragmentation to the SDK DX given that this SDK will be the only one supporting this if the PR gets merged. 

Tom will eventually come back to this after making some interfaces changes in the core SDK  codegen flow which will facilitate this implementation to all main SDKs. 

> So what are our solutions? Ideally we could create in the tests a custom registry, publish library and use that in the integration tests so we have the right behavior.

after syncing with @TomChv  we  agreed that this could be tested by installing the typescript sdk locally via `yarn install ./ts_sdk` and then  verifying that all the files get generated accordingly. This will require a small change in the typescript SDK  to upload relative path directories into the SDK init codegen phase but we agreed it is the correct thing to do.

PR #10847, opened 2025-08-05
feat: add contextual git

### LINKED ISSUE #8520: API to query current git context

jedevc:
Fixes #8520. Follow-up to https://github.com/dagger/dagger/pull/10195.

This allows a module to get git information about itself - it can do this through the core `Git` type.

The `defaultPath` annotation is extended to allow setting a default value for a `GitRepository`/`GitRef` parameter. If the caller does not provide a value, the method uses the specified default - either a local path (e.g., `"./.git"`) or a remote repository URL (e.g., `"https://github.com/dagger/dagger.git"`).

### DIFF
--- a/core/modfunc.go
+++ b/core/modfunc.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 
 	"dagger.io/dagger/telemetry"
+	"github.com/dagger/dagger/util/gitutil"
 	bkgw "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/identity"
 	bksession "github.com/moby/buildkit/session"
@@ -160,7 +161,7 @@ func (fn *ModuleFunction) setCallInputs(ctx context.Context, opts *CallOpts) ([]
 			return nil, fmt.Errorf("failed to convert arg %q: %w", input.Name, err)
 		}
 
-		if len(arg.metadata.Ignore) > 0 && arg.metadata.DefaultPath == "" { // DefaultPath (aka contextual) args already have ignore applied
+		if len(arg.metadata.Ignore) > 0 && !arg.metadata.isContextual() { // contextual args already have ignore applied
 			converted, err = fn.applyIgnoreOnDir(ctx, opts.Server, arg.metadata, converted)
 			if err != nil {
 				return nil, fmt.Errorf("failed to apply ignore pattern on arg %q: %w", input.Name, err)
@@ -213,7 +214,7 @@ func (fn *ModuleFunction) CacheConfigForCall(
 
 	var ctxArgs []*FunctionArg
 	for _, argMetadata := range fn.metadata.Args {
-		if argMetadata.DefaultPath == "" {
+		if !argMetadata.isContextual() {
 			// not a contextual argument
 			continue
 		}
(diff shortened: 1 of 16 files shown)


shykes:
Love the feature but a little worried about a proliferation of `+defaultXYZ`, it's not exactly intuitive and will get worse the more we add...

What would happen if we just made this `+default` ?

docs/docs-graphql/schema.graphqls
```diff
@@ -2404,6 +2404,8 @@ type Function {
 
     """The source map for the argument definition."""
     sourceMap: SourceMapID
+
+    defaultGit: String = ""
```
helderco: This argument doesn't have a description.

helderco:
Python and TypeScript support marked as done but I don't see those changes.

docs/docs-graphql/schema.graphqls
jedevc: Hm - clearly I did not have the foresight to actually push my changes. 🤔 

will push you when I get back to my keyboard 😅

jedevc:
> Love the feature but a little worried about a proliferation of `+defaultXYZ`, it's not exactly intuitive and will get worse the more we add...
> 
> What would happen if we just made this `+default` ?

I'd be a big fan. But for consistency, we'd probably want to make `+default` work with `File`s/`Directory`s right? Then the question is what we do with `defaultPath` (deprecate it?)

jedevc:
Hm, one reason to have a separate `defaultGit` is that in Python/TS/maybe others, the `// +default` is just part of the syntax: `param: Directory = defaultValue`.

But you can't do that with `Directory` as easily `param: Directory = "./path/to/directory"` because `"./path/to/directory"` isn't a `Directory`.

Maybe there's a clever solution here?

shykes:
> Hm, one reason to have a separate `defaultGit` is that in Python/TS/maybe others, the `// +default` is just part of the syntax: `param: Directory = defaultValue`.
> 
> But you can't do that with `Directory` as easily `param: Directory = "./path/to/directory"` because `"./path/to/directory"` isn't a `Directory`.
> 
> Maybe there's a clever solution here?

I was wondering if the work on standardized "object address"  #10770 could help with that? Since it will give us a standardized way to load various objects from a string representation...

Maybe that's a motivation to revisit the language discrepancy problem?

shykes:
I found the [original discussion](https://github.com/dagger/dagger/issues/7647#issuecomment-2201921231) that led to `+defaultPath` instead of `+default`. The decision was made because some languages (Python, Typescript) have their own syntax for default values, and we didn't want to perpetually confuse them with two generic ways to set defaults. By narrowing `+default` to `+defaultPath`, the idea was to avoid the ambiguity.

jedevc:
@shykes, as discussed ([in discord](https://discord.com/channels/707636530424053791/1407291515671285810)), updated to re-use `defaultPath`.

jedevc:
Also updated to include `dag.CurrentModule().Git()` to correspond with `dag.CurrentModule().Source()`.

This can be different from the contextual arg in the case of blueprints. The upstream module will *always* be at `CurrentModule`, while the contextual arg will also be the downstream module.

sipsma:
Was chatting live with @jedevc about this; the relationship between `dag.CurrentModule().Git()` and caching is tricky and will have some caveats. Module function calls are obviously not cached yet, but they do have cache keys and they will be cached in the near future. Those cache keys are based on the content of the module source itself and any contextual args provided to the function call (or its parents).

This means that even if you specify a branch like `main` for the ref of the module being invoked, where you'd expect to use the latest commit off `main`, you can easily get cache hits from old commits *if it just so happens that all the file contents for the module source + contextual args are the same*.

Example scenario:
1. You execute `dagger -m github.com/foo/bar@main ...`, which resolves the module source to `github.com/foo/bar@abc123`, make some function calls
2. You push a bunch of commits to `main` that don't change the module source or any relevant contextual dir args at all. New head is at `def456`
3. You execute `dagger -m github.com/foo/bar@main ...` again

In that scenario, at step 3 the function calls will not actually execute and instead be cached. So if you were expecting the function call to run again and the `dag.CurrentModule().Git()` call to return the latest commit (`def456`), you'd be wrong.

---

I don't think this is the end of the world now that I type it out, but it probably will result in some confusion. If there's no immediate need for `dag.CurrentModule().Git()` I'd probably air on the side of skipping it for now and sorting more of this out later.

jedevc:
Mm, good point. I'm gonna hold off on adding that method just for now - I think just having contextual git *should* be fine.

I've dropped [`f1c6d8e` (#10847)](https://github.com/dagger/dagger/pull/10847/commits/f1c6d8eda7185fa347edd85436b56c412e747b4d) now, but can easily bring it back if neccessary.

core/modfunc.go
```diff
@@ -624,40 +625,36 @@ func (fn *ModuleFunction) loadContextualArg(
...
-
-		dir, err := fn.mod.ContextSource.Value.Self().LoadContext(ctx, dag, arg.DefaultPath, nil, arg.Ignore)
+		dir, err := fn.mod.ContextSource.Value.Self().LoadContextDir(ctx, dag, arg.DefaultPath, nil, arg.Ignore)
 		if err != nil {
 			return nil, fmt.Errorf("failed to load contextual directory %q: %w", arg.DefaultPath, err)
 		}
 		return dagql.NewID[*Directory](dir.ID()), nil
 
 	case "File":
-		slog.Debug("moduleFunction.loadContextualArg: loading contextual file", "fn", arg.Name, "file", arg.DefaultPath)
```
TomChv, line 646: Why removing these debug logs? There have not been moved to `LoadContextDir` either

sdk/typescript/src/module/entrypoint/register.ts
```diff
@@ -146,10 +146,8 @@ export class Register {
         }
 
         // Check if both values are used, return an error if so.
-        if (arg.defaultValue && arg.defaultPath) {
-          throw new Error(
-            "cannot set both default value and default path from context",
-          )
+        if ([arg.defaultValue, arg.defaultPath].filter((v) => v).length > 1) {
+          throw new Error("cannot set both multiple defaults")
```
TomChv: typo? Only `cannot set multiple defaults` is enough?

core/modfunc.go
jedevc: Do you really want these logs?

Personally, I didn't find them to be very useful, as we're making a lot of queries here which help document the flow in the engine logs.

The reason it's removed is I was faced with either:
- Adding more for `LoadContextGit` for consistency (which I don't think are useful)
- Removing the existing ones for consistency
- Just having it be entirely inconsistent.

If they're really useful, I'll add them back in, but given the engine runs on `--debug` logging by default, I don't really want to clutter it up any more than we have to.

TomChv: No that's fine I think, it was really useful when I made that feature but now that it's stable that okay.
I we have an unexpected an issue I can still put them back and remove them after I found a fix

TomChv:
I retriggered the failed job, if they green after I'll merge that PR :)

PR #10801, opened 2025-07-28
ci: improve version module performance

jedevc:
Before:

```
$ time dagger call -m github.com/dagger/dagger/version@main version
v0.18.15-250728111339-1a8c04db77d6

Full trace at https://dagger.cloud/jedevc/traces/fa964659c3e20b2df74c3b43590d70ab
dagger-dev call -m github.com/dagger/dagger/version@main version  3.44s user 0.88s system 4% cpu 1:37.37 total

$ time dagger call -m github.com/dagger/dagger/version@main version
v0.18.15-250728111339-1a8c04db77d6

Full trace at https://dagger.cloud/jedevc/traces/8276c6e06d137a4fd0b665f431100f4e
dagger-dev call -m github.com/dagger/dagger/version@main version  0.75s user 0.21s system 9% cpu 10.175 total
```

After:

```
$ time dagger call -m github.com/jedevc/dagger/version@improve-version-perf version
v0.18.15-250728150307-c367a1eb271b

Full trace at https://dagger.cloud/jedevc/traces/45d5694213e5a322564c55867b766afd
dagger-dev call -m github.com/jedevc/dagger/version@improve-version-perf   3.42s user 0.79s system 4% cpu 1:36.15 total

$ time dagger call -m github.com/jedevc/dagger/version@improve-version-perf version
v0.18.15-250728150307-c367a1eb271b

Full trace at https://dagger.cloud/jedevc/traces/daa306ed18f07de4e1f3504f0273a439
dagger-dev call -m github.com/jedevc/dagger/version@improve-version-perf   0.44s user 0.12s system 10% cpu 5.276 total
```

The uncached performance is pretty much identical (we're still having to do all the same work), but the cache performance is twice as fast (from ~10s to ~5s) - we no longer need to pull the whole repo everytime.

|                | Uncached Total Time | Cached Total Time |
|----------------|---------------------|-------------------|
| [**Before**](https://dagger.cloud/jedevc/traces/8276c6e06d137a4fd0b665f431100f4e?span=394fb270e360d8c4)     | 97.37s              | 10.175s           |
| [**After**](https://dagger.cloud/jedevc/traces/daa306ed18f07de4e1f3504f0273a439?span=77a3d59c1f51993d)      | 96.15               | 5.276s            |

PR #10849, opened 2025-08-06
feat: add `git.commonAncestor`

jedevc:
This is pretty much the main reason we need our own `Git` implementation in `version/git.go` (which is slow).

See https://github.com/dagger/dagger/pull/10801 for more context.

We can implement this ourselves (with a reasonable amount of reshuffling and refactoring).

### DIFF
--- a/core/schema/git.go
+++ b/core/schema/git.go
@@ -139,6 +139,11 @@ func (s *gitSchema) Install(srv *dagql.Server) {
 			Doc(`The resolved commit id at this ref.`),
 		dagql.NodeFunc("ref", DagOpWrapper(srv, s.fetchRef)).
 			Doc(`The resolved ref name at this ref.`),
+		dagql.NodeFunc("commonAncestor", s.commonAncestor).
+			Doc(`Find the best common ancestor between this ref and another ref.`).
+			Args(
+				dagql.Arg("other").Doc(`The other ref to compare against.`),
+			),
 	}.Install(srv)
 }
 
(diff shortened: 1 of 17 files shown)


core/schema/git.go
```diff
@@ -139,6 +139,11 @@ func (s *gitSchema) Install(srv *dagql.Server) {
 			Doc(`The resolved commit id at this ref.`),
 		dagql.NodeFunc("ref", DagOpWrapper(srv, s.fetchRef)).
 			Doc(`The resolved ref name at this ref.`),
+		dagql.NodeFunc("mergeBase", s.mergeBase).
+			Doc(`Find the merge-base (best common ancestor) between this ref and another ref.`).
```
vito: Bikeshedding mostly for the sake of it, from a quick read of https://git-scm.com/docs/git-merge-base:

* Would it be clearer to just call it `commonAncestor`? I'm not sure how well known `git merge-base` command is for folks who would be reaching for this. (Possibly very well known - if you're in this area of the API, maybe you've dealt with `git` release engineering before?) There doesn't appear to be a precedent yet for choosing abstract names over `git` plumbing names (which can be famously obtuse), so not really sure.
* Should the API support returning more than one ancestor? Sounds like the command does, but maybe that's a fringe case that would be fatal to this use case anyway. Or should we just pick the first and simplify things for the caller?
* Should the API accept multiple other refs instead of just one? e.g. for a n-way merge

jedevc: > * Would it be clearer to just call it `commonAncestor`?

Reasonable, happy with that!

> * Should the API support returning more than one ancestor?

Mm, not convinced about this - struggling to imagine where it would be useful, and generally I guess users would probably only want the first? We can easily add another method later that's `commonAncestors` (plural) if it's super required.

> * Should the API accept multiple other refs instead of just one? e.g. for a n-way merge

Mm. Not sure actually? No strong opinions, but I do like the current ergonomics of `ref.commonAncestor(otherRef)`.

vito: Yeah agree overall.

I wonder if, for the last one, you could do `CommonAncestor(CommonAncestor(refA, refB), refC)` in the meantime

jedevc: Yeah, doing the the n-way merge as you mention should definitely work.

PR #10850, opened 2025-08-06
add `CacheVolume.snapshot: Directory!`

vito:


vito:
Not really viable to do efficiently and soundly; went with a different approach that keeps the implementation as internal details instead in #10876

PR #10853, opened 2025-08-07
feat: move gitIgnore loading into dagOp

TomChv:
Optimize performances by moving the gitIgnore loading into a dagOpWrapper internal operation so it's cached
more efficiently.
Also removed the dag.Select to read file in favour of a mount + fs.Walkdir.

This PR follows up #10736 and apply this comment: https://github.com/dagger/dagger/pull/10736#discussion_r2254524913

## Benchmarks

Screenshot 2025-08-07 at 17 46 27

Specific load git ignore benchmark: https://github.com/dagger/dagger/pull/10853#issuecomment-3164939667
GitIgnoreFor specific benchmarks: https://github.com/dagger/dagger/pull/10853#issuecomment-3168369309

### DIFF
--- a/core/directory.go
+++ b/core/directory.go
@@ -1150,7 +1150,7 @@ func (dir *Directory) GitIgnorePatterns(ctx context.Context) ([]string, error) {
 				return prevErr
 			}
 
-			if !d.IsDir() && d.Name() != ".gitignore" {
+			if d.IsDir() || d.Name() != ".gitignore" {
 				return nil
 			}
 


TomChv:
As asked by @helderco, I also did some benchmark specially on how long it takes to load the modules git ignore:

<details>
<summary>Code snippet</summary>

```go
		start := time.Now()

		// Load .gitignore patterns before loading the module so they can be used in the `Include` argument
		// instead of Exclude so we get the right behaviour.
		// We only include the .gitignore from the module source path and above up to the context directory.
		ignorePatterns, err := loadDirectoryGitIgnorePatterns(ctx, src.Local.ContextDirectoryPath, filepath.Join(src.Local.ContextDirectoryPath, moduleSourcePath))
		if err != nil {
			return fmt.Errorf("failed to load .gitignore patterns: %w", err)
		}

		rebasedIgnorePatterns, err := rebaseGitIgnorePatterns(src.Local.ContextDirectoryPath, src.Local.ContextDirectoryPath, ignorePatterns)
		if err != nil {
			return fmt.Errorf("failed to rebase .gitignore patterns: %w", err)
		}

		end := time.Since(start)

		slog.Debug("moduleSource .gitignore patterns", "duration", end.String(), "duration_ms", end.Milliseconds(), "duration_seconds", end.Seconds(), "duration_ns", end.Nanoseconds())

```

</details>

Screenshot 2025-08-07 at 18 31 51

helderco:
Hmmm... how is 1000 .gitignore better than our dagger/dagger? I suggest you run something like `dagger -M -c 'host | directory . | glob "**"` to rule out other complications with module loading. If you isolate the measurement to just the part that's figuring out all the patterns to add to the exclude, it should be faster on dagger/dagger than on the repo with 500 .gitignore files even.

TomChv:
> Hmmm... how is 1000 .gitignore better than our dagger/dagger? I suggest you run something like dagger -M -c 'host | directory . | glob "**" to rule out other complications with module loading. If you isolate the measurement to just the part that's figuring out all the patterns to add to the exclude, it should be faster on dagger/dagger than on the repo with 500 .gitignore files even.

My theory is that the time it takes to get the cache is slower on dagger than for 1000 .gitignore that are not in nested directories.
What I'm testing there is how long it takes to loadGitIgnore during moduleLoading, not inside `host.Directory.__gitIgnoreFor`

I'll do some benchmark specially inside gitIgnoreFor to see what we got

TomChv:
So I did another benchmark where I isolated only the `gitIgnoreFor` operation and compute 2 things:
- The time it takes to read/process a file (excluding the append to the list)
- The total time of the function

See snippets below for the code

Screenshot 2025-08-08 at 17 45 18

We observe that the dagger repo has a slightly longer overhead but it may be cause by the large number of patterns in all these gitignore.
The 1000 and 500 test case only have 1 line in the .gitignore while average on dagger repo is 4.

The result also may vary depending on my computer, I had my fan running and noticed it takes slightly longer than when they are turned off, but overall it's quite fast, even for a large number of files.

<details>
<summary>Code implementation</summary>

```go

func (dir *Directory) GitIgnoreFor(ctx context.Context) ([]string, error) {
	result := []string{}
	var timeByReadFile []time.Duration

	start := time.Now()

	_, err := execInMount(ctx, dir, func(root string) error {
		resolvedDir, err := containerdfs.RootPath(root, dir.Dir)
		if err != nil {
			return err
		}

		return filepath.WalkDir(resolvedDir, func(path string, d fs.DirEntry, prevErr error) error {
			if prevErr != nil {
				return prevErr
			}

			if d.Name() != ".gitignore" {
				return nil
			}

			startFile := time.Now()
			content, err := os.ReadFile(path)
			if err != nil {
				return err
			}

			relPath, err := filepath.Rel(resolvedDir, path)
			if err != nil {
				return err
			}

			patterns := parseGitIgnore(string(content), filepath.Dir(relPath))
			endFile := time.Since(startFile)
			timeByReadFile = append(timeByReadFile, endFile)

			result = append(result, patterns...)

			return nil
		})
	})
	if err != nil {
		if errors.Is(err, errEmptyResultRef) {
			// empty directory, i.e. llb.Scratch()
			return []string{}, nil
		}
		return nil, err
	}

	end := time.Since(start)

	var total time.Duration
	for _, t := range timeByReadFile {
		total += t
	}

	slog.Info("GitIgnoreFor",
		"dir_name", dir.Dir,
		"duration", end,
		"duration_ms", end.Milliseconds(),
		"avg_by_file_ns", total.Nanoseconds()/int64(len(timeByReadFile)),
		"avg_by_file_ms", total.Milliseconds()/int64(len(timeByReadFile)))

	return result, nil
}

```

</details>

core/schema/host.go
```diff
@@ -413,17 +414,29 @@ func loadDirectoryGitIgnorePatterns(ctx context.Context, parentPath string, host
...
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load directory ignore patterns: %w", err)
 	}
 
-	gitIgnorePatterns, err := core.LoadGitIgnoreInDirectory(ctx, directory)
-	if err != nil {
-		return nil, fmt.Errorf("failed to load .gitignore in context directory %q: %w", parentPath, err)
+	for i := 1; i <= result.Len(); i++ {
+		entry, err := result.Nth(i)
```
jedevc: `dagql.ResultArray` is just an array type, so you can do `for _, entry := range result`

core/directory.go
```diff
@@ -1136,6 +1136,52 @@ func (dir *Directory) Mount(ctx context.Context, f func(string) error) error {
...
+		if err != nil {
+			return err
+		}
+
+		return filepath.WalkDir(resolvedDir, func(path string, d fs.DirEntry, prevErr error) error {
+			if prevErr != nil {
+				return prevErr
+			}
+
+			if d.Name() != ".gitignore" {
```
jedevc: Make sure `d` is a regular file. If a directory is called `.gitignore` (for some reason), we shouldn't error out.

core/schema/host.go
```diff
@@ -413,17 +414,29 @@ func loadDirectoryGitIgnorePatterns(ctx context.Context, parentPath string, host
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to get entry %d in directory ignore patterns: %w", i, err)
+		}
+
+		entryValue, ok := dagql.UnwrapAs[dagql.String](entry)
```
jedevc: We don't need `UnwrapAs`, you can just call `string(entry)`, since it's just a string type.

core/schema/directory.go
```diff
@@ -202,6 +202,8 @@ func (s *directorySchema) Install(srv *dagql.Server) {
 				dagql.Arg("target").Doc(`Location of the file or directory to link to (e.g., "/existing/file").`),
 				dagql.Arg("linkName").Doc(`Location where the symbolic link will be created (e.g., "/new-file-link").`),
 			),
+		dagql.NodeFunc("__gitIgnoreFor", DagOpWrapper(srv, s.gitIgnoreFor)).
```
jedevc: Maybe just call this `__gitignorePatterns`?

Just to confirm - this returns *docker*-style patterns right? After converting them from gitignore style patterns? If so, worth clarifying in `Doc`.

core/schema/host.go
TomChv: Screenshot 2025-08-11 at 13 30 17

I think I still need to unwrap it no?

jedevc: Oh. Just call `Self`.

core/schema/directory.go
TomChv: Yes, applied in [7ff0141](https://github.com/dagger/dagger/pull/10853/commits/7ff0141df57caa39e12fadd9c25b19c5062431bb)

core/schema/host.go
TomChv: Oh right! Applied in [7ff0141](https://github.com/dagger/dagger/pull/10853/commits/7ff0141df57caa39e12fadd9c25b19c5062431bb)

core/directory.go
TomChv: Applied in [7ff0141](https://github.com/dagger/dagger/pull/10853/commits/7ff0141df57caa39e12fadd9c25b19c5062431bb)

core/schema/host.go
TomChv: Applied in [7ff0141](https://github.com/dagger/dagger/pull/10853/commits/7ff0141df57caa39e12fadd9c25b19c5062431bb)

core/directory.go
```diff
@@ -1136,6 +1136,52 @@ func (dir *Directory) Mount(ctx context.Context, f func(string) error) error {
...
+		if err != nil {
+			return err
+		}
+
+		return filepath.WalkDir(resolvedDir, func(path string, d fs.DirEntry, prevErr error) error {
+			if prevErr != nil {
+				return prevErr
+			}
+
+			if !d.IsDir() && d.Name() != ".gitignore" {
```
jedevc:
```suggestion
			if d.IsDir() || d.Name() != ".gitignore" {
```

🤔

jedevc: Looks like this is making the tests fail.

TomChv: Yeah I fixed it, my bad

jedevc (approved):

PR #10855, opened 2025-08-08
fix(core/git): ensure HTTP auth is propagated to submodules

grouville:
Fixes https://discord.com/channels/707636530424053791/1397694731496067194/1397694731496067194

Previously, HTTP(S) authentication worked only for the parent repository clone. Submodule clones failed because the auth header (http.<scheme>://<host>/.extraheader) passed via CLI args (-c) wasn’t inherited by nested Git processes (git submodule update). By injecting the header through the GIT_CONFIG_* environment variables instead, the auth configuration propagates to submodules automatically, fixing the inheritance issue and ensuring consistent authentication.

### Update 2025-08-12

Since opening this PR, I’ve made a few adjustments:
- Implemented shallow fetch fallback:
> Now we always attempt git submodule update with --depth=1 first. If the server doesn’t support shallow clones (fatal: dumb http transport does not support shallow capabilities), we retry without the --depth argument. This preserves performance benefits where supported while ensuring compatibility.
- Added smart-HTTP test helper:
> Extracted the previously inline smart-HTTP setup into a reusable helper (gitSmartHTTPServiceDirAuth). This simplifies writing tests involving authenticated smart-HTTP git clones, which are necessary to properly support shallow fetches.
- Extended submodule auth tests:
> Expanded TestSubmoduleAuth to verify behavior for both smart and dumb HTTP protocols using the same repository setup. This clearly demonstrates the new fallback mechanism works correctly in both cases.
- Reduced PR scope:
> Removed unrelated refactors and kept the changes tightly focused on submodule auth and shallow-fetch behavior.

### DIFF
--- a/core/git.go
+++ b/core/git.go
@@ -799,9 +799,17 @@ func doGitCheckout(
 
 	// TODO: this feels completely out-of-sync from how we do the rest
 	// of the clone - caching will not be as great here
-	_, err = checkoutGit.Run(ctx, "submodule", "update", "--init", "--recursive", "--depth=1")
-	if err != nil {
-		return fmt.Errorf("failed to update submodules for %s: %w", remote, err)
+	subArgs := []string{"submodule", "update", "--init", "--recursive", "--depth=1"}
+	if _, err := checkoutGit.Run(ctx, subArgs...); err != nil {
+		if strings.Contains(err.Error(), "does not support shallow") {
+			subArgs = slices.DeleteFunc(subArgs, func(s string) bool {
+				return strings.HasPrefix(s, "--depth")
+			})
+			_, err = checkoutGit.Run(ctx, subArgs...)
+		}
+		if err != nil {
+			return fmt.Errorf("failed to update submodules for %s: %w", remote, err)
+		}
 	}
 
 	if discardGitDir {
--- a/core/git.go
+++ b/core/git.go
@@ -256,7 +256,7 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI, 
 		if err != nil {
 			return nil, nil, err
 		}
-		authHeader = "basic " + base64.StdEncoding.EncodeToString(
+		authHeader = "Basic " + base64.StdEncoding.EncodeToString(
 			fmt.Appendf(nil, "%s:%s", username, password),
 		)
 	} else if repo.AuthHeader.Self() != nil {
--- a/core/git.go
+++ b/core/git.go
@@ -234,6 +234,7 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI, 
 		}
 	}()
 
+	var authHeader string
 	if repo.AuthToken.Self() != nil {
 		// caller-supplied username takes priority; otherwise pick a host-specific default
 		username := repo.AuthUsername
@@ -255,24 +256,19 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI, 
 		if err != nil {
 			return nil, nil, err
 		}
-		authHeader := "basic " + base64.StdEncoding.EncodeToString(
+		authHeader = "basic " + base64.StdEncoding.EncodeToString(
 			fmt.Appendf(nil, "%s:%s", username, password),
 		)
-		opts = append(opts, gitutil.WithArgs(
-			"-c", "http."+repo.URL.Remote()+".extraheader=Authorization: "+authHeader,
-		))
 	} else if repo.AuthHeader.Self() != nil {
 		secretStore, err := query.Secrets(ctx)
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to get secret store: %w", err)
 		}
-		authHeader, err := secretStore.GetSecretPlaintext(ctx, repo.AuthHeader.ID().Digest())
+		byteAuthHeader, err := secretStore.GetSecretPlaintext(ctx, repo.AuthHeader.ID().Digest())
 		if err != nil {
 			return nil, nil, err
 		}
-		opts = append(opts, gitutil.WithArgs(
-			"-c", "http."+repo.URL.Remote()+".extraheader=Authorization: "+string(authHeader),
-		))
+		authHeader = string(byteAuthHeader)
 	}
 
 	if repo.SSHAuthSocket.Self() != nil {
@@ -304,8 +300,11 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI, 
 	if err != nil {
 		return nil, nil, err
 	}
+
+	var resolvPath string
 	if netConf != nil {
-		resolvPath, err := mountResolv(netConf)
+		var err error
+		resolvPath, err = mountResolv(netConf)
 		if err != nil {
 			return nil, nil, err
 		}
@@ -316,6 +315,18 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI, 
 			return os.Remove(resolvPath)
 		})
 	}
+
+	opts = append(opts, gitutil.WithExec(func(ctx context.Context, cmd *exec.Cmd) error {
+		if authHeader != "" {
+			// Scope auth to this host and make it inherit to sub-commands:
+			// use GIT_CONFIG_* so submodule clones see http.extraheader (unlike -c).
+			base := repo.URL.Scheme + "://" + repo.URL.Host
+			cmd.Env = gitutil.MergeGitConfigEnv(cmd.Env, map[string]string{
+				"http." + base + "/.extraheader": "Authorization: " + authHeader,
+			})
+		}
+		return runWithStandardUmaskAndNetOverride(ctx, cmd, "", resolvPath)
+	}))
 
 	return gitutil.NewGitCLI(opts...), cleanups.Run, nil
 }
--- a/core/git.go
+++ b/core/git.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"bytes"
 	"context"
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
@@ -234,43 +233,6 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI, 
 		}
 	}()
 
-	var authHeader string
-	if repo.AuthToken.Self() != nil {
-		// caller-supplied username takes priority; otherwise pick a host-specific default
-		username := repo.AuthUsername
-		switch {
-		case username != "":
-			// explicit override – use as-is
-		case repo.URL.Host == "bitbucket.org":
-			// NOTE: bitbucket.org is picky, and needs *this* username
-			username = "x-token-auth"
-		default:
-			username = "x-access-token"
-		}
-
-		secretStore, err := query.Secrets(ctx)
-		if err != nil {
-			return nil, nil, fmt.Errorf("failed to get secret store: %w", err)
-		}
-		password, err := secretStore.GetSecretPlaintext(ctx, repo.AuthToken.ID().Digest())
-		if err != nil {
-			return nil, nil, err
-		}
-		authHeader = "Basic " + base64.StdEncoding.EncodeToString(
-			fmt.Appendf(nil, "%s:%s", username, password),
-		)
-	} else if repo.AuthHeader.Self() != nil {
-		secretStore, err := query.Secrets(ctx)
-		if err != nil {
-			return nil, nil, fmt.Errorf("failed to get secret store: %w", err)
-		}
-		byteAuthHeader, err := secretStore.GetSecretPlaintext(ctx, repo.AuthHeader.ID().Digest())
-		if err != nil {
-			return nil, nil, err
-		}
-		authHeader = string(byteAuthHeader)
-	}
-
 	if repo.SSHAuthSocket.Self() != nil {
 		socketStore, err := query.Sockets(ctx)
 		if err == nil {
@@ -308,23 +270,34 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI, 
 		if err != nil {
 			return nil, nil, err
 		}
-		opts = append(opts, gitutil.WithExec(func(ctx context.Context, cmd *exec.Cmd) error {
-			return runWithStandardUmaskAndNetOverride(ctx, cmd, "", resolvPath)
-		}))
 		cleanups.Add("remove updated /etc/resolv", func() error {
 			return os.Remove(resolvPath)
 		})
 	}
 
-	opts = append(opts, gitutil.WithExec(func(ctx context.Context, cmd *exec.Cmd) error {
-		if authHeader != "" {
-			// Scope auth to this host and make it inherit to sub-commands:
-			// use GIT_CONFIG_* so submodule clones see http.extraheader (unlike -c).
-			base := repo.URL.Scheme + "://" + repo.URL.Host
-			cmd.Env = gitutil.MergeGitConfigEnv(cmd.Env, map[string]string{
-				"http." + base + "/.extraheader": "Authorization: " + authHeader,
-			})
+	if repo.AuthToken.Self() != nil {
+		secretStore, err := query.Secrets(ctx)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to get secret store: %w", err)
 		}
+		password, err := secretStore.GetSecretPlaintext(ctx, repo.AuthToken.ID().Digest())
+		if err != nil {
+			return nil, nil, err
+		}
+		opts = append(opts, gitutil.WithHTTPTokenAuth(repo.URL, string(password), repo.AuthUsername))
+	} else if repo.AuthHeader.Self() != nil {
+		secretStore, err := query.Secrets(ctx)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to get secret store: %w", err)
+		}
+		byteAuthHeader, err := secretStore.GetSecretPlaintext(ctx, repo.AuthHeader.ID().Digest())
+		if err != nil {
+			return nil, nil, err
+		}
+		opts = append(opts, gitutil.WithHTTPAuthorizationHeader(repo.URL, string(byteAuthHeader)))
+	}
+
+	opts = append(opts, gitutil.WithExec(func(ctx context.Context, cmd *exec.Cmd) error {
 		return runWithStandardUmaskAndNetOverride(ctx, cmd, "", resolvPath)
 	}))
 
--- a/core/integration/git_test.go
+++ b/core/integration/git_test.go
@@ -811,33 +811,65 @@ git add .gitmodules
 git commit -m 'Make submodule URL relative'
 git remote add origin file:///srv/parent.git
 git push -u origin main
+# dumb HTTP needs server-info
+git --git-dir=/srv/submodule.git update-server-info
+git --git-dir=/srv/parent.git    update-server-info
 `})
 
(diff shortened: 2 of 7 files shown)


core/integration/git_test.go
```diff
@@ -768,6 +770,127 @@ func (GitSuite) TestWithAuth(ctx context.Context, t *testctx.T) {
...
+		WithExec([]string{"git", "config", "--global", "user.email", "test@dagger.io"}).
+		WithExec([]string{"git", "config", "--global", "user.name", "Test User"}).
+		WithExec([]string{"git", "config", "--global", "init.defaultBranch", "main"}).
+		// submodule repo
+		WithDirectory("/tmp/sub", submoduleContent).
+		WithWorkdir("/tmp/sub").
+		WithExec([]string{"git", "init"}).
+		WithExec([]string{"git", "add", "."}).
+		WithExec([]string{"git", "commit", "-m", "Initial submodule commit"}).
+		WithExec([]string{"git", "clone", "--bare", "--no-hardlinks", ".", "/srv/submodule.git"}).
```
marcosnils: mostly curious:  why are we doing all this `/tmp` and then `git clone --bare` dancing?

core/git.go
```diff
@@ -266,39 +257,118 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI,
...
+		if strings.HasPrefix(e, "GIT_CONFIG_COUNT=") {
+			if n, err := strconv.Atoi(strings.TrimPrefix(e, "GIT_CONFIG_COUNT=")); err == nil && n > count {
+				count = n
+			}
+			continue
+		}
+		out = append(out, e)
+	}
+
+	// Append in sorted key order for reproducibility
```
marcosnils: nit: there's `keys := slices.Sorted(maps.Keys(entries))` for this

core/git.go
```diff
@@ -705,43 +773,47 @@ func doGitCheckout(
...
-	if err != nil {
+	if _, err = checkoutGit.Run(ctx, "reflog", "expire", "--all", "--expire=now"); err != nil {
 		return fmt.Errorf("failed to expire reflog for remote %s: %w", remote, err)
 	}
-
 	if err := os.Remove(filepath.Join(checkoutDirGit, "FETCH_HEAD")); err != nil && !errors.Is(err, os.ErrNotExist) {
 		return fmt.Errorf("failed to remove FETCH_HEAD for remote %s: %w", remote, err)
 	}
 
+	// Sync submodule URLs with .gitmodules so the next update pulls from the right location.
```
marcosnils: not sure I follow why this is needed. What does the "so the next update" means? Isn't this `doGitCheckout` function only gets called for new checkouts each time? How would we need to do a `submodule sync`? 🤔

core/integration/git_test.go
grouville: Hey Marcos,

After your comment, I reworked the setup to be closer to a real Git flow and avoid the old `/tmp → clone --bare → sed` path.

We need the parent repo’s `.gitmodules` in history to contain a relative submodule URL because that’s the exact case that triggers the bug we’re fixing: when you clone the parent over HTTP, Git resolves `../submodule.git` against the parent’s remote (e.g. `http://host/parent.git` → `http://host/submodule.git`) and does a second HTTP fetch. Without our fix, that second fetch wasn’t inheriting the auth header.

The tricky bit is that you can’t just run `git submodule add ../submodule.git` during setup: at author time Git treats it as a local path, so it fails unless that path exists.

**Here’s a minimal repro of that failure**

```shell
# fresh
rm -rf /tmp/repro && mkdir -p /tmp/repro && cd /tmp/repro
sudo rm -rf /srv/submodule.git ; sudo mkdir -p /srv && sudo git init --bare /srv/submodule.git

# parent worktree (no sibling submodule.git here)
mkdir parent && cd parent
git init
git config user.email test@example.com
git config user.name "Test User"
echo parent > parent.txt
git add . && git commit -m "init parent"

# This fails: ../submodule.git is a *local* path and doesn't exist here
git -c protocol.file.allow=always submodule add ../submodule.git sub
```

Which gives this error:

```shell
fatal: repository '/tmp/repro/submodule.git' does not exist
fatal: clone of '/tmp/repro/submodule.git' into submodule path '/tmp/repro/parent/sub' failed
```

Git is looking for `../submodule.git` on disk relative to the parent’s worktree, and it doesn’t exist there, so the command fails.

To get both a working fetch and a committed relative URL, we:

- Add the submodule from its actual bare location (`file:///srv/submodule.git`) so `git submodule add` can fetch right away.

- Immediately rewrite `.gitmodules` to `../submodule.git` with `git config -f` and commit that change.

That leaves the history exactly like you’d see in a real repo that uses relative submodules on a server, so when the test later clones the parent over HTTP, it reproduces the real-world resolution and fetch, and exercises the auth propagation path we just fixed.

The rest of the changes (init bare repos in /srv, commit in temp worktrees, push to the bares) are just to mirror normal Git usage and keep the test hermetic and reproducible with Dagger’s Directory() API.

core/git.go
grouville: Updated, thank you 🙏 I also improved the function to make it easier to read

core/git.go
grouville: Yeah, you’re right: this was a leftover from my earlier testing that I thought was needed.

In our flow, `doGitCheckout` is always working on a brand-new worktree, and the submodules are initialized for the first time via `git submodule update --init --recursive`. That means `.gitmodules` and `.git/config` are already in sync at that point, so running `git submodule sync` was redundant 👌

core/git.go
```diff
@@ -283,22 +273,88 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI,
...
+	if user == "" {
+		switch host {
+		case "bitbucket.org":
+			// NOTE: bitbucket.org is picky, and needs *this* username
+			user = "x-token-auth"
+		default:
+			user = "x-access-token"
+		}
+	}
+	pw := strings.TrimSpace(string(password))
```
jedevc: Why space trimming here? If the user's specified a token that contains whitespace, we should use it.

core/git.go
```diff
@@ -283,22 +273,88 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI,
...
+		)
+	}
+
+	// Final COUNT == prior COUNT + number of new pairs.
+	out = append(out, countPrefix+strconv.Itoa(next+len(keys)))
+	return out
+}
+
+// buildBasicAuthHeader creates 'Authorization: Basic <base64(user:pass)>'.
+func buildBasicAuthHeader(host, explicitUser string, password []byte) string {
```
jedevc:
```suggestion
func basicAuthHeader(host, explicitUser string, password string) string {
```

core/git.go
```diff
@@ -249,11 +236,14 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI,
 		if err != nil {
 			return nil, nil, err
 		}
-		opts = append(opts, gitutil.WithArgs(
-			"-c", "http."+repo.URL.Remote()+".extraheader=Authorization: "+string(authHeader),
-		))
+		hl := strings.TrimSpace(string(authHeader))
+		if !strings.HasPrefix(strings.ToLower(hl), "authorization:") {
```
jedevc: 🤔 why?

It's definitely an edge case, but there should be one behavior - we shouldn't attempt to read this as just a header line.

core/git.go
```diff
@@ -283,22 +273,88 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI,
...
-		})
+		cleanups.Add("remove updated /etc/resolv", func() error { return os.Remove(resolvPath) })
 	}
 
+	// Inject auth headers and DNS config into git commands.
+	// Auth: Uses GIT_CONFIG_* env vars to set http.extraheader for submodules.
+	// DNS: Binds custom resolv.conf in mount namespace for session consistency.
+	opts = append(opts, gitutil.WithExec(func(ctx context.Context, cmd *exec.Cmd) error {
+		if headerLine != "" {
+			base := repo.URL.Scheme + "://" + repo.URL.Host
```
jedevc: Yeah, this is the main part of the patch, I kinda missed on the first pass that this is where the magic happens.

I'm a bit... on the fence. It's kinda crappy, but it's also definitely better than what was there before.

For example, if I have a gitlab submodule in my github repo, then there's actually no way for me to specify how to clone that submodule. Which is annoying. We need better submodule APIs, but this SGTM in the meantime.

core/integration/git_test.go
```diff
@@ -768,6 +770,144 @@ func (GitSuite) TestWithAuth(ctx context.Context, t *testctx.T) {
...
+    git add .gitmodules
+    git commit -m 'Make submodule URL relative'
+    git remote add origin file:///srv/parent.git
+    git push -u origin main
+  `})
+
+	gitReposDir := gitReposCtr.Directory("/srv") // serve this via git-http-backend
+
+	// --- lighttpd smart-http with basic auth ---
+	gitService := c.Container().
```
jedevc: Question! Why aren't we using `gitServiceHTTP`?

If there's a good reason for not, then these should be split out to helpers.

core/git.go
```diff
@@ -283,22 +273,88 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI,
...
+		return runWithStandardUmaskAndNetOverride(ctx, cmd, "", resolvPath)
+	}))
+
 	return gitutil.NewGitCLI(opts...), cleanups.Run, nil
 }
 
+// mergeGitConfigEnv merges key/value config pairs into the environment
+// using GIT_CONFIG_COUNT/KEY/VALUE. Preserves existing entries.
+// Keys are sorted for deterministic builds.
+func mergeGitConfigEnv(env []string, entries map[string]string) []string {
```
jedevc: These helpers should be split into the `gitutil` package.

core/git.go
jedevc: See https://github.com/dagger/dagger/pull/10836 - we should be doing secret whitespace trimming at source.

core/integration/git_test.go
```diff
@@ -768,6 +770,144 @@ func (GitSuite) TestWithAuth(ctx context.Context, t *testctx.T) {
...
 }
 
+func (GitSuite) TestSubmoduleAuth(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+	t.Cleanup(func() { _ = c.Close() })
+
+	// Unique token per run
+	authToken := c.SetSecret("submodule-test-token", "test-token-"+uuid.NewString())
+
+	// Build content declaratively (keeps your favorite part)
```
jedevc: `keeps your favorite part`?

core/git.go
grouville: Good point. I have no idea on how to fix it entirely though ... Would love to have your insight if you prefer to have a fool-proof solution here

jedevc: https://github.com/dagger/dagger/issues/10863 - wrote some thoughts there.

But yeah, this PR is significantly better than the current behavior in `main` so, we should definitely take this :eyes: Just didn't really think about this case before :cry: API design hard.

core/git.go
grouville: I was not aware, the intent was the same, but implementation wrong, thanks 🙏

core/integration/git_test.go
grouville: The main reason was that, as we update the submodule with a _depth of 1_, only smartHTTP is supported.

We get this error: `fatal: dumb http transport does not support shallow capabilities`. Now that I've dug a bit more: 1) we should either implement a fallback mechanism, as you did for the parent 2) I should move / extract the inline helper as a smartHTTP helper for the tests 3) maybe both ?


_Tested with:_

```go
func (GitSuite) TestSubmoduleAuth(ctx context.Context, t *testctx.T) {
	c := connect(ctx, t)
	t.Cleanup(func() { _ = c.Close() })

	authToken := c.SetSecret("submodule-test-token", "test-token-"+uuid.NewString())
	submoduleContent := c.Directory().WithNewFile("submodule.txt", "This is the submodule content")
	parentContent := c.Directory().WithNewFile("parent.txt", "This is the parent content")

	gitReposCtr := c.Container().
		From(alpineImage).
		WithExec([]string{"apk", "add", "git"}).
		WithExec([]string{"git", "config", "--global", "user.email", "test@dagger.io"}).
		WithExec([]string{"git", "config", "--global", "user.name", "Test User"}).
		WithExec([]string{"git", "config", "--global", "init.defaultBranch", "main"}).
		WithExec([]string{"sh", "-lc", "mkdir -p /srv && git init --bare /srv/submodule.git && git init --bare /srv/parent.git"}).
		// submodule: commit content and push to bare
		WithDirectory("/tmp/sub", submoduleContent).
		WithExec([]string{"sh", "-lc", `
set -eux
cd /tmp/sub
git init
git add .
git commit -m 'Initial submodule commit'
git remote add origin file:///srv/submodule.git
git push -u origin main
`}).
		// parent: add submodule via absolute URL, then rewrite .gitmodules to relative and commit
		WithDirectory("/tmp/parent", parentContent).
		WithExec([]string{"sh", "-lc", `
set -eux
cd /tmp/parent
git init
git add .
git commit -m 'Initial parent commit'
git -c protocol.file.allow=always submodule add file:///srv/submodule.git sub
git commit -m 'Add submodule (absolute url to fetch)'
git config -f .gitmodules submodule.sub.url ../submodule.git
git add .gitmodules
git commit -m 'Make submodule URL relative'
git remote add origin file:///srv/parent.git
git push -u origin main
# dumb HTTP needs server-info
git --git-dir=/srv/submodule.git update-server-info
git --git-dir=/srv/parent.git    update-server-info
`})

	gitReposDir := gitReposCtr.Directory("/srv")

	svc, baseURL := httpServiceDirAuth(ctx, t, c, "", gitReposDir, "x-access-token", authToken)
	parentURL := baseURL + "/parent.git"

	t.Run("clone with HTTPAuthToken", func(ctx context.Context, t *testctx.T) {
		tree := c.Git(parentURL, dagger.GitOpts{
			ExperimentalServiceHost: svc,
			HTTPAuthUsername:        "x-access-token",
			HTTPAuthToken:           authToken,
		}).Branch("main").Tree()

		txt, err := tree.File("parent.txt").Contents(ctx)
		require.NoError(t, err)
		require.Equal(t, "This is the parent content", txt)

		sub, err := tree.File("sub/submodule.txt").Contents(ctx)
		require.NoError(t, err)
		require.Equal(t, "This is the submodule content", sub)
	})

	t.Run("clone without auth fails", func(ctx context.Context, t *testctx.T) {
		_, err := c.Git(parentURL, dagger.GitOpts{
			ExperimentalServiceHost: svc,
		}).Branch("main").Tree().File("parent.txt").Contents(ctx)
		require.Error(t, err)
	})
}
```

jedevc: Aha that's an excellent catch. Yeaaahh, having smart http would be good for tests.

But also, yup, good shout, let's do the same fallback: https://github.com/dagger/dagger/blob/40dba8d312c76caed679a9ba952048c52c96a259/core/git.go#L514-L526

core/integration/git_test.go
```diff
@@ -768,6 +769,110 @@ func (GitSuite) TestWithAuth(ctx context.Context, t *testctx.T) {
 	})
 }
 
+func (GitSuite) TestSubmoduleAuth(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+	t.Cleanup(func() { _ = c.Close() })
+
+	authToken := c.SetSecret("submodule-test-token", "test-token-"+uuid.NewString())
```
jedevc: We use `identity.NewID()` for generating random IDs.

core/integration/services_test.go
```diff
@@ -2257,6 +2257,95 @@ server {
...
+auth.require = ( "/" => ( "method" => "basic", "realm" => "Git", "require" => "valid-user" ) )
+`
+	}
+
+	ctr := c.Container().
+		From("debian:bookworm-slim").
+		WithExec([]string{"bash", "-lc", `
+set -eux
+apt-get update
+apt-get install -y --no-install-recommends git lighttpd ca-certificates
```
jedevc: Is there a reason we're using lightttpd here over nginx? A lot of our other tests use nginx, would be good to be consistent.

core/integration/git_test.go
```diff
@@ -768,6 +769,110 @@ func (GitSuite) TestWithAuth(ctx context.Context, t *testctx.T) {
...
+			sub, err := tree.File("sub/submodule.txt").Contents(ctx)
+			require.NoError(t, err)
+			require.Equal(t, "This is the submodule content", sub)
+		})
+
+		t.Run("without auth fails", func(ctx context.Context, t *testctx.T) {
+			_, err := c.Git(parentURL, dagger.GitOpts{
+				ExperimentalServiceHost: gitSrv,
+			}).Branch("main").Tree().File("parent.txt").Contents(ctx)
+			require.Error(t, err)
```
jedevc, line 843: Check the error contents - this makes sure that we're actually checking that *auth* is failing, and not some other thing.

`requireErrOut` is a good helper for this.

util/gitutil/config_env_test.go
```diff
@@ -0,0 +1,64 @@
...
+			n, _ := strconv.Atoi(strings.TrimPrefix(parts[0], "GIT_CONFIG_KEY_"))
+			idx[parts[1]] = n
+		}
+	}
+
+	require.Contains(t, idx, "http.https://host/.extraheader")
+	require.Contains(t, idx, "core.autocrlf")
+
+	gotIdx := []int{idx["http.https://host/.extraheader"], idx["core.autocrlf"]}
+	sort.Ints(gotIdx)
```
jedevc: `slices.Sort`

util/gitutil/config_env_test.go
```diff
@@ -0,0 +1,64 @@
...
+	add := map[string]string{
+		"http.https://host/.extraheader": "Authorization: basic abc",
+		"core.autocrlf":                  "false",
+	}
+	got := MergeGitConfigEnv(env, add)
+
+	idx := map[string]int{}
+	for _, e := range got {
+		if strings.HasPrefix(e, "GIT_CONFIG_KEY_") {
+			parts := strings.SplitN(e, "=", 2)
```
jedevc: `strings.Cut`

util/gitutil/config_env_test.go
```diff
@@ -0,0 +1,64 @@
...
+	add := map[string]string{
+		"b.key": "2",
+		"a.key": "1",
+	}
+	got := MergeGitConfigEnv(env, add)
+	idxA, idxB := -1, -1
+	for _, e := range got {
+		if strings.HasPrefix(e, "GIT_CONFIG_KEY_") && strings.HasSuffix(e, "=a.key") {
+			idxA, _ = strconv.Atoi(strings.TrimPrefix(strings.SplitN(e, "=", 2)[0], "GIT_CONFIG_KEY_"))
+		}
```
jedevc: `strings.Cut`, split before this, we don't need to keep slicing and dicing the string over and over.

grouville:
Failing because of: https://github.com/dagger/dagger/pull/10878

core/git.go
```diff
@@ -317,6 +316,18 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI,
 		})
 	}
 
+	opts = append(opts, gitutil.WithExec(func(ctx context.Context, cmd *exec.Cmd) error {
```
jedevc: We're already adding `gitutil.WithExec` above, we should only add it once.

jedevc: Actually, we could also just update `gitutil` itself to take a `WithConfig` option, and hide all that logic in there.

That would be the neatest way to do it! 🧹

jedevc:
@grouville just needs a brief oneline changelog, then feel free to merge!

PR #10865, opened 2025-08-12
create dagger cloud trace url predicate type

kpenfound:
With the ability to get a Dagger Cloud trace URL from the Dagger client, its now possible to attach the current session as in-toto attestation evidence on an artifact produced from your Dagger function.

This is created from the [in-toto predicate template](https://github.com/in-toto/attestation/blob/main/spec/predicates/template/template.md)

This markdown file once on main will be pointed to by the URL `https://dagger.io/evidence/trace-url/v1`

An example use case for this predicate type is with Artifactory and this [artifactory dagger module](https://daggerverse.dev/mod/github.com/kpenfound/dag/artifactory)


jedevc:
Seems reasonable, I did some of the attestation stuff in buildkit once-upon-a-time :grin: 

Curious if we need any changes to the API or whatever? How would one create an attestation like this, does it need to be manually created?

kpenfound:
@jedevc right now it'll be manually created, but its made possible by the addition of `Client.Cloud.TraceURL` in the API. In my use case, I'll be creating them in artifactory with this module https://daggerverse.dev/mod/github.com/kpenfound/dag/artifactory

Open to anything for the location of this file. The current path follows standards but it really doesn't matter as we'll be redirecting from a dagger.io URL to the markdown file

jedevc:
Yeah SGTM, I don't know about the `./attestation` path, but :shrug: 

Maybe `./docs` may be better? No strong opinions.

PR #10867, opened 2025-08-12
Improved TUI debugging

jedevc:
This PR adds improved TUI debugging, by allowing a new `T` hotkey to open a terminal on a span.

[asciicast](https://asciinema.org/a/jhVkugQYzGpLM6rzjmoMvPLTM)

This is supported for existing terminal functions on `Container` and `Directory`, as well as on `Service`, where `Service.terminal` has now been added as well!

Related:
- https://github.com/dagger/dagger/issues/10923

### DIFF
--- a/dagql/idtui/frontend_pretty.go
+++ b/dagql/idtui/frontend_pretty.go
@@ -31,6 +31,7 @@ import (
 	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/term"
 
+	"dagger.io/dagger"
 	"dagger.io/dagger/telemetry"
 	"github.com/dagger/dagger/dagql/dagui"
 	"github.com/dagger/dagger/engine/slog"
@@ -44,6 +45,8 @@ var ErrInterrupted = errors.New("interrupted")
 
 type frontendPretty struct {
 	dagui.FrontendOpts
+
+	dag *dagger.Client
 
 	// don't show live progress; just print a full report at the end
 	reportOnly bool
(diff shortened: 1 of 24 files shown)


dagql/idtui/frontend_pretty.go
```diff
@@ -623,6 +632,8 @@ func (fe *frontendPretty) keys(out *termenv.Output) []key.Binding {
 		key.NewBinding(key.WithKeys("r"),
 			key.WithHelp("r", "go to error"),
 			KeyEnabled(focused != nil && focused.ErrorOrigin != nil)),
+		key.NewBinding(key.WithKeys("t"),
```
vito, line 635: This should probably be conditional on whether the focused span is terminal-able, like above

dagql/idtui/frontend_pretty.go
```diff
@@ -1300,6 +1369,7 @@ func (fe *frontendPretty) handleEditlineKey(msg tea.KeyMsg) (cmd tea.Cmd) {
 	return cmd
 }
 
+//nolint:gocyclo // splitting this up doesn't feel more readable
```
vito, line 1401: 👍

docs/docs-graphql/schema.graphqls
```diff
@@ -3968,6 +3968,11 @@ type Service {
     kill: Boolean = false
   ): ServiceID!
 
+  """Forces evaluation of the pipeline in the engine."""
+  sync: ServiceID!
```
vito, line 3972: Why was this added anyway? Seems like it's a no-op? (`Evaluate` just returns `nil, nil`)

jedevc: This just "forces" it to be sent to the engine.

```go
// this never gets sent to the engine, since there's never a query
LoadServiceFromID(id).Terminal()

// i need some method to actually "force" it to be sent
LoadServiceFromID(id).Terminal().Sync(ctx)
```

*Maybe* `Service.terminal` could always take a `ctx`, similar to `Start`? But then it has a different signature than `Container.terminal`/`Directory.terminal`.

Issue #10868, opened 2025-08-13
🐞 Dagger's gitignore parser does not respect tailing slash


skycaptain:
### What is the issue?

A trailing slash is commonly used in .gitignore files to ignore complete folders. In one of my projects, I have the following lines to ignore all build directories:

```
# Ignore build directories
build*/
```

In the same project, I have a file as part of my Dagger module under `.dagger/build.go`. Recent versions of Dagger appear to have an issue where this file is ignored when running the module as shown below. The problem can be mitigated by renaming the file, but it seems Dagger's gitignore parser misinterprets that line from my gitignore file and also ignores files instead of just directories. This worked fine with v0.18.14, as also shown below.

### Dagger version

dagger v0.18.15-250813115555-11d52c2b8d11 (docker-image://localhost/dagger-engine.dev?container=dagger-engine.dev&volume=dagger-engine.dev&cleanup=false) darwin/arm64/v8

### Steps to reproduce

MRE: https://github.com/skycaptain/dagger-issue-build-ignore

Run with v0.18.4:

```
$ dagger functions
▶ connect 0.2s
▶ load module: . 1.8s

Name    Description
build   -
test    -
```

Run with latest HEAD:

```
$ dagger functions
▶ connect 0.5s
▼ load module: . 1.2s ERROR
╰─▼ Container.withExec(args: ["go", "build", "-ldflags", "-s -w", "-o", "/runtime", "."]): Container! 0.2s ERROR
  ┃ # dagger/my-project-dev                                                                                                                                                                                                                
  ┃ ./main.go:16:33: undefined: Build                                                                                                                                                                                                      
  ┃ ./main.go:17:10: undefined: Build                                                                                                                                                                                                      
  ! process "go build -ldflags -s -w -o /runtime ." did not complete successfully: exit code: 1

Error logs:

▼ Container.withExec(args: ["go", "build", "-ldflags", "-s -w", "-o", "/runtime", "."]): Container! 0.2s ERROR
# dagger/my-project-dev                                                                                                                                                                                                                    
./main.go:16:33: undefined: Build                                                                                                                                                                                                          
./main.go:17:10: undefined: Build                                                                                                                                                                                                          
! process "go build -ldflags -s -w -o /runtime ." did not complete successfully: exit code: 1
Error: failed to serve module: input: moduleSource.asModule failed to call module "my-project-dev" to get functions: call constructor: process "go build -ldflags -s -w -o /runtime ." did not complete successfully: exit code: 1
```

[...]

PR #10870, opened 2025-08-13
chore: fix incorrect behaviour for dir only pattern on .gitignore

TomChv:
Fixes https://github.com/dagger/dagger/issues/10868


core/util.go
```diff
@@ -539,9 +539,14 @@ func getRefOrEvaluate[T fileOrDirectory](ctx context.Context, t T) (bkcache.Immu
...
 //   - If a pattern is negative exclusion (starts with `!`) or targets directory only
-//     (ends with `/`), we treat is as a regular path then read the exclusion to make
+//     (ends with `/`), we treat it as a regular path then read the exclusion to make
 //     sure the recusive pattern is applied if needed.
+//     For directory only exclusion, we need to add a `/**` suffix to the path to make sure
+//     the exclusion is only applied to directory and not files that may match the pattern
+//     (see https://github.com/dagger/dagger/issues/10868).
 //     Example: !foo becomes foo then **/foo then !**/foo
+//     Example: foo/ becomes **/foo/
+//     Example: foo*/ becomes **/foo*/**
```
jedevc, line 549: What about `!foo*/`?

core/util_test.go
```diff
@@ -53,6 +53,19 @@ foo/bar
...
+		gitIgnoreContent := `foo/
+!build*/
+./node_modules/
+`
+
+		patterns := parseGitIgnore(
+			gitIgnoreContent, ".",
+		)
+
+		require.Equal(t, []string{"**/foo/", "!**/build*/**", "node_modules/"}, patterns)
```
jedevc, line 66: Can we actually evaluate these patterns against an example directory in the tests?

Maybe by using `mobypatternmatcher` directly?

jedevc:
This isn't quite right.

A gitignore that has `foo*/` and a dockerignore that has `foo*/**` isn't equivalent.

For the `git` case, the `foo/` directory itself is excluded, while the dockerignore pattern includes it (and it's just empty).

jedevc:
Tom's on holiday, so gonna pick this up, since we need to resolve this before he gets back.

TomChv:
Closing in favor of #10883

PR #10873, opened 2025-08-14
Propose new capitalization rule

vikram-dagger:
- Propose rule in style guide for capitalization
- Update text to match proposed style rule
- Add back missing link to engine config schema
- Update reference overview page to be in sync with latest nav changes


vikram-dagger:
We earlier decided on this rule in Discord: 

`unless we're referencing a feature/product, keep in lowercase, since we're usually talking about instances of a thing: a module, a function, a workflow. You run "an instance of a  Dagger engine" (90% of our usage in docs) but  "Dagger Engine provides an API for ..." (10% of our usage in docs)`

However when implementing this, I found two issues:
- I found myself second-guessing which was the correct style of capitalization to use for a given case - in most cases the capitalized version seemed more correct
- From our public blog posts, it looks like we capitalize Dagger Functions, Dagger Shell and Dagger Engine quite consistently
- - https://dagger.io/blog/a-shell-for-the-container-age-introducing-dagger-shell
- - https://dagger.io/blog/run-dagger-on-kubernetes
- - https://dagger.io/blog/dagger-0-12
- - https://dagger.io/blog/argo-cd-kubernetes

So I'm proposing a new and simpler rule in this PR, which is to _always_ capitalize these product/feature names.

Note: I'm not wedded to this, very happy to change it further if we can identify a more specific bright-line rule about capitalization of these terms.

PR #10875, opened 2025-08-14
chore: defer session cleanups while in no-exit state

jedevc:
By using the `-E` flag, or the `E` hotkey, we can enter a special TUI no-exit state.
    
While in this state, even after the dagger operation is completed, we keep the TUI open and running. However, previously, we were completely shutting down the session and the telemetry.
    
For some interesting future operations (e.g. #10867), we want to defer these callbacks until we're quitting the TUI completely - we want to keep an active client as long as the TUI is running in some form :smile:

### DIFF
--- a/dagql/idtui/frontend_plain.go
+++ b/dagql/idtui/frontend_plain.go
@@ -2,6 +2,7 @@ package idtui
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"slices"
@@ -14,6 +15,7 @@ import (
 	"github.com/dagger/dagger/dagql/call/callpbv1"
 	"github.com/dagger/dagger/dagql/dagui"
 	"github.com/dagger/dagger/engine/slog"
+	"github.com/dagger/dagger/util/cleanups"
 	"github.com/muesli/termenv"
 	"github.com/pkg/browser"
 	"github.com/vito/go-interact/interact"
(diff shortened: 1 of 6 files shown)


dagql/idtui/frontend_plain.go
```diff
@@ -192,7 +192,13 @@ func (fe *frontendPlain) Run(ctx context.Context, opts dagui.FrontendOpts, run f
 		}()
 	}
 
-	runErr := run(ctx)
+	cleanup, runErr := run(ctx)
+	if cleanup != nil {
+		if err := cleanup(); runErr == nil {
+			runErr = err
+		}
```
vito: This took a lot of squinting over here, and I can't tell if there's a typo, because I think technically the logic works out, but maybe this would be less squinty?:

```suggestion
		runErr = errors.Join(runErr, cleanup())
```

dagql/idtui/frontend_pretty.go
```diff
@@ -978,7 +984,16 @@ type doneMsg struct {
...
-	return doneMsg{fe.run(fe.runCtx)}
+	cleanup, err := fe.run(fe.runCtx)
+	return cleanupMsg{
+		cleanup: cleanup,
+		msg:     doneMsg{err},
+	}
+}
+
+type cleanupMsg struct {
+	cleanup func() error
```
vito: Since the `error` return value gets dropped everywhere, maybe we should just turn it into a `func()` that calls + logs if there was an error?

jedevc:
Spend quite a bit of time working out an issue where using `-E` wouldn't actually wait for the result of `cleanup`, and so would leave traces in incomplete states on dagger cloud.

PR #10876, opened 2025-08-14
Coding agents

vito:
This is a HUGE PR resulting from a bunch of experiments that sort of blended from one to another. (Sorry... again. I'll split this up after my wedding. :sweat_smile:)

The big things:

* `Env` now has a filesystem, which automatically propagates between tool calls and gets re-hydrated as contextual args to modules
* `LLM` now supports MCP servers (in other words, `LLM` is an MCP client), which _also_ run with the same filesystem, and even propagate changes out to subsequent tool calls
* `LLM` now captures and displays logs in each tool call response (supersedes #10768)
* `LLM` now exposes its tools in "dynamic mode" by default, like it used to. The static tool calling scheme (`call_method`, `chain_methods`) are now opt-in, and `dagger mcp` now opts in.
* `LLM` now supports "blocking" APIs, so they're not exposed to the model
* `LLM.withModule` installs a module into the LLM's `Env`, exposing its functions as tools, with automatic Env filesystem propagation
* The TUI now displays total cost of current session in a sidebar.
* A few new APIs to faciliate implementing coding agents efficiently in Dagger.

With this PR, you can implement a CLI agent as a function that returns an `LLM`, and start it via `dagger call`, which now goes into prompt mode for an `LLM` return type.

Here's Doug, a custom coding agent inspired by Claude and the rest, but implemented fully natively in Dagger:

[asciicast](https://asciinema.org/a/Ke04q94Pus0Ffpib45MZWv6hh)

Here's Claude Code, running as an MCP server, used alongside an LSP MCP server:

[asciicast](https://asciinema.org/a/733583)

### DIFF
--- a/core/client_resource.go
+++ b/core/client_resource.go
@@ -16,8 +16,8 @@ func GetClientResourceAccessor(ctx context.Context, parent *Query, externalName 
 	}
 
 	var scopeDigest digest.Digest
-	if m.Self() != nil {
-		scopeDigest = digest.Digest(m.Self().Source.Value.Self().Digest)
+	if m != nil {
+		scopeDigest = digest.Digest(m.Source.Value.Self().Digest)
 	}
 
 	// Use an HMAC, which allows us to keep the externalName un-inferrable.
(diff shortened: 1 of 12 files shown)


vito:
split into:

* #10907 
* #10946 
* #10937

PR #10880, opened 2025-08-14
chore: CacheMap for directory dagops

alexcb:


core/schema/wrapper.go
```diff
@@ -150,6 +152,18 @@ func getOpts[T dagql.Typed, A any](opts ...DagOpOptsFn[T, A]) *DagOpOpts[T, A] {
 	return &o
 }
 
+func getSelfDigest(a any) (digest.Digest, error) {
+	switch x := a.(type) {
```
alexcb, line 156: sadly I couldn't come up with a cleaner way using generics :(

TomChv: Do `getSelfDigest` will eventually extend to other object types? If so, we could create an interface for these specific object and then you can call the methods there?

alexcb: It's a good idea, but unfortunately I can't do that without some other casting, since this is called from the `DagOpDirectory[T dagql.Typed, A any](..., self T, ...)` method.

I'm not sure I really want to open up the can of worms if I try changing `dagql.Typed` to something more constrained.

> ... will [it] eventually extend to other object types

There might be other types, if `DagOpDirectory` gets called on other types -- possibly `core.Container`, but I'm not entirely sure.

PR #10883, opened 2025-08-15
feat: add native gitignore in fsutil

jedevc:
Implements my suggestion in https://github.com/dagger/dagger/issues/10868#issuecomment-3183991570.

We remove the original approach from https://github.com/dagger/dagger/pull/10736, which attempts to convert `gitignore`-style patterns into `dockerignore`-style patterns (supported by buildkit, which uses mobypatternmatcher). Instead, we add a `gitignoreFS`, which creates a filtered FS in the style of `filterFS`, but that automatically extracts gitignore filters from `.gitignore` files in the tree, and applies them using go-git.

This has a couple of advantages:
- Instead of converting, we directly process gitignore patterns. No weird issues like https://github.com/dagger/dagger/pull/10870 can ever occur again. Using a mature implementation like `go-git` ensures we're as close to `git` as possible.
- Fewer file transfers to-and-from the host filesystem - we don't need to load gitignores to process patterns, the file-transfer protocol simply understands them natively.
- Simpler implementation - all the gitignore code is completely isolated to *one* component in the file transfer stack, instead of being spread out through gitignores and the like.

### DIFF
--- a/core/integration/host_test.go
+++ b/core/integration/host_test.go
@@ -299,6 +299,133 @@ func (HostSuite) TestDirectoryExcludeInclude(ctx context.Context, t *testctx.T) 
 	})
 }
 
+func (HostSuite) TestDirectoryGitIgnore(ctx context.Context, t *testctx.T) {
+	dir := t.TempDir()
+	gitignore := strings.Join([]string{
+		"**.txt",
+		".git/",
+		"!subdir/e.txt",
+		"*.md",
+		"internal/foo/",
+	}, "\n")
+
+	require.NoError(t, os.MkdirAll(filepath.Join(dir, ".git"), 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, ".gitignore"), []byte(gitignore), 0o600))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "b.md"), []byte("2"), 0o600))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "c.txt.rar"), []byte("3"), 0o600))
+	require.NoError(t, os.MkdirAll(filepath.Join(dir, "subdir"), 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "subdir", "b.md"), []byte("1"), 0o600))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "subdir", "e.txt"), []byte("2"), 0o600))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "subdir", "g.txt"), []byte("6"), 0o600))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "subdir", "h.yaml"), []byte("6"), 0o600))
+	require.NoError(t, os.MkdirAll(filepath.Join(dir, "subdir2"), 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "subdir2", ".gitignore"), []byte(`*.yaml
+.gitignore
+`), 0o600))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "subdir2", "foo.go"), []byte("1"), 0o600))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "subdir2", "bar.txt"), []byte("1"), 0o600))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "subdir2", "baz.md"), []byte("1"), 0o600))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "subdir2", "bool.yaml"), []byte("1"), 0o600))
+
+	c := connect(ctx, t)
+
+	t.Run("apply auto git ignore by default", func(ctx context.Context, t *testctx.T) {
+		hostDir := c.Host().Directory(dir)
+
+		rootHostDir, err := hostDir.Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{".gitignore", "c.txt.rar", "subdir/", "subdir2/"}, rootHostDir)
+
+		subDirEntries, err := hostDir.Directory("subdir").Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{"e.txt", "h.yaml"}, subDirEntries)
+
+		subDir2Entries, err := hostDir.Directory("subdir2").Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{"foo.go"}, subDir2Entries)
+	})
+
+	t.Run("correctly apply parent .gitignore when children path is given", func(ctx context.Context, t *testctx.T) {
+		subDirEntries, err := c.Host().Directory(filepath.Join(dir, "subdir")).Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{"e.txt", "h.yaml"}, subDirEntries)
+	})
+
+	t.Run("don't apply .gitignore if no .git found", func(ctx context.Context, t *testctx.T) {
+		dir := t.TempDir()
+
+		require.NoError(t, os.WriteFile(filepath.Join(dir, ".gitignore"), []byte(`**/**`), 0o600))
+		require.NoError(t, os.WriteFile(filepath.Join(dir, "bar.txt"), []byte("1"), 0o600))
+		require.NoError(t, os.WriteFile(filepath.Join(dir, "foo.go"), []byte("1"), 0o600))
+
+		entries, err := c.Host().Directory(dir).Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{".gitignore", "bar.txt", "foo.go"}, entries)
+	})
+
+	t.Run("don't load files ignored by .gitignore", func(ctx context.Context, t *testctx.T) {
+		dir := t.TempDir()
+
+		require.NoError(t, os.MkdirAll(filepath.Join(dir, ".git"), 0o755))
+		require.NoError(t, os.WriteFile(filepath.Join(dir, ".gitignore"), []byte("bar/\nbaz.txt\n"), 0o600)) // ignore everything
+		require.NoError(t, os.MkdirAll(filepath.Join(dir, "foo/"), 0o755))
+		require.NoError(t, os.WriteFile(filepath.Join(dir, "foo/foo.txt"), []byte("1"), 0o600))
+		require.NoError(t, os.MkdirAll(filepath.Join(dir, "bar/"), 0o755))
+		require.NoError(t, os.WriteFile(filepath.Join(dir, "bar/bar.txt"), []byte("1"), 0o600))
+		require.NoError(t, os.WriteFile(filepath.Join(dir, "baz.txt"), []byte("1"), 0o600))
+
+		// sanity check!
+		entries, err := c.Host().Directory(filepath.Join(dir, "foo/")).Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{"foo.txt"}, entries)
+
+		_, err = c.Host().Directory(filepath.Join(dir, "bar/")).Entries(ctx)
+		require.Error(t, err)
+		requireErrOut(t, err, "no such file or directory")
+
+		_, err = c.Host().File(filepath.Join(dir, "baz.txt")).Contents(ctx)
+		require.Error(t, err)
+		requireErrOut(t, err, "no such file or directory")
+	})
+
+	t.Run("disable git auto ignore", func(ctx context.Context, t *testctx.T) {
+		entries, err := c.Host().Directory(dir, dagger.HostDirectoryOpts{NoGitAutoIgnore: true}).Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{".git/", ".gitignore", "b.md", "c.txt.rar", "subdir/", "subdir2/"}, entries)
+
+		subDirEntries, err := c.Host().Directory(filepath.Join(dir, "subdir"),
+			dagger.HostDirectoryOpts{NoGitAutoIgnore: true},
+		).Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{"b.md", "e.txt", "g.txt", "h.yaml"}, subDirEntries)
+
+		subDir2Entries, err := c.Host().Directory(filepath.Join(dir, "subdir2"),
+			dagger.HostDirectoryOpts{NoGitAutoIgnore: true},
+		).Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{".gitignore", "bar.txt", "baz.md", "bool.yaml", "foo.go"}, subDir2Entries)
+	})
+
+	t.Run("use gitignores after no git ignores", func(ctx context.Context, t *testctx.T) {
+		// tests a weird edge case, where we populate the local dir, but then
+		// copy doesn't actually respect gitignores
+
+		dir := t.TempDir()
+		require.NoError(t, os.MkdirAll(filepath.Join(dir, ".git"), 0o755))
+		require.NoError(t, os.WriteFile(filepath.Join(dir, ".gitignore"), []byte("a.txt\n"), 0o600))
+		require.NoError(t, os.WriteFile(filepath.Join(dir, "a.txt"), []byte("1"), 0o600))
+		require.NoError(t, os.WriteFile(filepath.Join(dir, "b.txt"), []byte("2"), 0o600))
+
+		entries, err := c.Host().Directory(dir, dagger.HostDirectoryOpts{NoGitAutoIgnore: true}).Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{".git/", ".gitignore", "a.txt", "b.txt"}, entries)
+
+		entries, err = c.Host().Directory(dir).Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{".git/", ".gitignore", "b.txt"}, entries)
+	})
+}
+
 func (HostSuite) TestDirectoryCacheBehavior(ctx context.Context, t *testctx.T) {
 	baseDir := t.TempDir()
 	c := connect(ctx, t)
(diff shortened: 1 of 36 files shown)


core/integration/host_test.go
```diff
@@ -299,6 +299,89 @@ func (HostSuite) TestDirectoryExcludeInclude(ctx context.Context, t *testctx.T)
...
+		hostDir := c.Host().Directory(dir)
+
+		rootHostDir, err := hostDir.Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{".gitignore", "c.txt.rar", "subdir/", "subdir2/"}, rootHostDir)
+
+		subDirEntries, err := hostDir.Directory("subdir").Entries(ctx)
+		require.NoError(t, err)
+		// require.Equal(t, []string{"b.md", "e.txt", "h.yaml"}, subDirEntries) // XXX: wrong?
+		require.Equal(t, []string{"e.txt", "h.yaml"}, subDirEntries)
```
jedevc: These tests are directly imported from #10736. But I think they're wrong, at least from my testing using `git` directly?

We previously ignore `*.md` - so `b.md` shouldn't be found.

*Unlike* in `dockerignore`, `*.md` in `gitignore` should ignore all markdown files, no matter where they are:

```
❯ cd `mktemp -d`
❯ git init
Initialized empty Git repository in /tmp/tmp.JGS60Al5B1/.git/
❯ echo '*.md' >> .gitignore
❯ touch a.md
❯ mkdir b
❯ touch b/c.md
❯ git add .
❯ git status
On branch main

No commits yet

Changes to be committed:
  (use "git rm --cached <file>..." to unstage)
	new file:   .gitignore
```

jedevc:
There's *one* failing check, that I'm unsure how to resolve.

https://github.com/dagger/dagger/blob/88f94234bb4843aee9d804f0b2b510c0be95a578/core/integration/module_test.go#L5094-L5098

`internal/telemetry` is ignored - if I pass this as an argument... what should happen?
- Should it error out? (currently implemented)
- Should it error out with a specific "this file/directory is ignored, and cannot be loaded"?
- Should it be loaded, but with none of it's contents, since they're ignored?
- Should we discard the gitignore pattern that would *directly* ignore `internal/telemetry`, but continue to apply all the other patterns inside it?
- Should we just load `internal/telemetry` as-is, no changes? (but then, what's the point of the gitignore at all?)

TomChv:
> `internal/telemetry` is ignored - if I pass this as an argument... what should happen?
> 
> * Should it error out? (currently implemented)
> * Should it error out with a specific "this file/directory is ignored, and cannot be loaded"?
> * Should it be loaded, but with none of it's contents, since they're ignored?
> * Should we discard the gitignore pattern that would _directly_ ignore `internal/telemetry`, but continue to apply all the other patterns inside it?
> * Should we just load `internal/telemetry` as-is, no changes? (but then, what's the point of the gitignore at all?)

If you try to access a directory that doesn't exist with our current API, it will error out, I think we should keep that DX to avoid inconsistency but maybe as you mention in option 2, with a specific error so the user can have a better understanding of why this dir/file is missing.

core/schema/host.go
```diff
@@ -150,6 +150,7 @@ func (s *hostSchema) Install(srv *dagql.Server) {
 				dagql.Arg("exclude").Doc(`Exclude artifacts that match the given pattern (e.g., ["node_modules/", ".git*"]).`),
 				dagql.Arg("include").Doc(`Include only artifacts that match the given pattern (e.g., ["app/", "package.*"]).`),
 				dagql.Arg("noCache").Doc(`If true, the directory will always be reloaded from the host.`),
+				dagql.Arg("noGitAutoIgnore").Doc(`Don't apply .gitignore filter rules inside the directory`),
```
jedevc, line 154: Wondering about this.

I think this makes sense as the default, but it is a bit weird to me that users can't disable it when passing directories from the CLI.

Do we need a way to override it?

TomChv: I think we should have a way but how?

It may be simpler as a pragma so you can target a specific argument, because from the CLI it would be global to the dagger call cmd

```go
func (e *Example) A(
  //+noGitAutoIgnore
  dir *dagger.Directory
) { ... } 

```

core/integration/host_test.go
TomChv: I think this come from a bad interpretation of the official git ignore rules: https://git-scm.com/docs/gitignore#_pattern_format

I assumed that `*.go` means every go files in the current directory but not the children, actually it also apply to any children directory.

core/schema/host.go
jedevc: Mm, yeah, but what about if it's not a contextual arg?

jedevc:
Hm, mildly tempted to just have the default for `Host.directory` ... not have gitignore support by default.

Just do it for modules? That would resolve both the above issues for now.

core/schema/host.go
TomChv: Well that option can apply to any directory, even if it's not a contextual directory.

Because when we do `dagger call a --dir /foo/bar`, we look for the closest .git to see if we can apply git ignore, it's totally decouple from contextual argument in this case.

You may have `//+noGitAutoIgnore` but no `//+defaultPath`

jedevc: Sorry, my bad, I should have been clearer - what if it's not an arg at all - what if it's not a user function?

Like, what if it's a core function, or a function in a module you don't control?

We'd have to have weird flag syntax, something like `--dir file:///foo/bar?noGitAutoIgnore=true`.

I don't really love it, kind of why I'm suggesting https://github.com/dagger/dagger/pull/10883#issuecomment-3229077512 in the meantime.

TomChv: Yeah I guess then no support for `Host.directory` except for module argument may be the solution.

I wonder what's @helderco opinion

helderco: That's the problem with  `dagger call`, but with shell there's no issue:
```
dagger -c 'a $(host | directory --no-git-auto-ignore /foo/bar)'
```
We could disable just for call, but then how would you enable it?

If it were up to me we'd deprecate `dagger call` in favor of shell instead of having this split.

helderco: To clarify I think only in rare use cases people will want it off so my preference is to have it enabled by default as much as possible.

Also, `dagger call` is already more limited than shell in a lot of ways so it's not a deal breaker this is yet another thing, although I do understand if someone that needs it off already uses `dagger call` a lot it can be frustrating being forced to switch to shell syntax. That's why I'd rather not have this split in general. It creates friction.

jedevc: Yeah, okay, I'm happy with that answer :tada:

I'd consider this resolved!

helderco:
> There's _one_ failing check, that I'm unsure how to resolve.

I think this should be split into two tests.

In order to match `.gitignore`, since the intent of the test is to negate the ignore on a specific file, the first pattern should be `internal/telemetry/**`:
```diff
func (t *Test) IgnoreDirButKeepFileInSubdir(
-  // +ignore=["internal/telemetry", "!internal/telemetry/proxy.go"]
+  // +ignore=["internal/telemetry/**", "!internal/telemetry/proxy.go"]
  // +defaultPath="./dagger"
  dir *dagger.Directory,
) *dagger.Directory {
  return dir
}
```

But you raise an interesting question. What happens when you do have these patterns?
```
internal/telemetry
!internal/telemetry/proxy.go
```
I think like git, the second pattern should have no effect (basically `internal/telemetry` shouldn't exist) and if you try to select an ignored subdir from it it should be like any other directory that doesn't have the target path:
```
❯ dagger -M -c 'directory | directory internal/telemetry | entries'
! /internal/telemetry: no such file or directory
```

jedevc:
Still struggling a bit here. I think I have two questions:

1. If I do `dagger -c 'host | directory internal/telemetry/ | entries'` where `internal/telemetry/` is in the `.gitignore`, what should happen? Currently this errors out in `host.directory`, because `internal/telemetry/` gets filtered out by `gitignoreFS`.

   Is that reasonable? Or does one of the alternative bullet points from https://github.com/dagger/dagger/pull/10883#issuecomment-3213816607 sound better?

2. Does that failing test even make sense now? `internal/telemetry` is part of the gititgnore, so you can't unignore it using an include pattern - they're two separate layers. I think this makes sense, if it's in the gitignore it's not gonna be in a modulesource that's from git (and not local) - so you get a more consistent behavior.

   This is a change from the original impl, I explained some of the logic in https://github.com/dagger/dagger/pull/10736#discussion_r2279078577.

   It wouldn't be *impossible* to have that behavior (so that you could do `!internal/telemetry` if you really wanted to override the git ignore) - but it would definitely require a big fork/change of fsutil, so wanna confirm, since I actually think we shouldn't allow it.

helderco:
> 1. If I do `dagger -c 'host | directory internal/telemetry/ | entries'` where `internal/telemetry/` is in the `.gitignore`, what should happen? Currently this errors out in `host.directory`, because `internal/telemetry/` gets filtered out by `gitignoreFS`.
>    Is that reasonable? Or does one of the alternative bullet points from [feat: add native gitignore in fsutil #10883 (comment)](https://github.com/dagger/dagger/pull/10883#issuecomment-3213816607) sound better?

I think it’s reasonable. If it’s trivial it can be interesting to give a more specific error message but could be confusing if that doesn’t happen everywhere there’s exclude patterns filtering a directory (otherwise users may think that’s not the problem if the error doesn’t specify any exclude patterns).

> 2. Does that failing test even make sense now? `internal/telemetry` is part of the gititgnore, so you can't unignore it using an include pattern - they're two separate layers. I think this makes sense, if it's in the gitignore it's not gonna be in a modulesource that's from git (and not local) - so you get a more consistent behavior. 

Oh yeah, that’s a good point i.e., getting consistent behavior with git. How do you disallow overrides in `Host.directory`, for example? Do you have to apply all the `.gitignore` patterns first and only filter the user’s patterns on the resulting directory? Does it take a non-trivial performance hit to do that?

> […] wanna confirm, since I actually think we shouldn't allow it.

Agreed. I didn’t previously consider consistency with git here as far as the overrides are concerned, was just concerned about an escape hatch out of extreme caution. In that case, if an escape hatch ends up being necessary, I can support doing that through simply disabling the auto gitignore feature, as we already have (except on the CLI’s directory flag).

core/integration/host_test.go
TomChv: You can change the test behaviour here

core/integration/host_test.go
```diff
@@ -299,6 +299,92 @@ func (HostSuite) TestDirectoryExcludeInclude(ctx context.Context, t *testctx.T)
...
+
+		subDir2Entries, err := hostDir.Directory("subdir2").Entries(ctx)
+		require.NoError(t, err)
+		// require.Equal(t, []string{"baz.md", "foo.go"}, subDir2Entries) // XXX: wrong?
```
TomChv: Same here, change the test behaviour based on https://github.com/dagger/dagger/pull/10883/files#r2303934302

core/integration/host_test.go
```diff
@@ -299,6 +299,92 @@ func (HostSuite) TestDirectoryExcludeInclude(ctx context.Context, t *testctx.T)
...
+		require.Equal(t, []string{"foo.go"}, subDir2Entries)
+	})
+
+	t.Run("correctly apply parent .gitignore when children path is given", func(ctx context.Context, t *testctx.T) {
+		subDirEntries, err := c.Host().Directory(filepath.Join(dir, "subdir")).Entries(ctx)
+		require.NoError(t, err)
+		// require.Equal(t, []string{"b.md", "e.txt", "h.yaml"}, subDirEntries) // XXX: wrong?
```
TomChv: Same here, change the test behaviour based on https://github.com/dagger/dagger/pull/10883/files#r2303934302

core/integration/module_test.go
```diff
@@ -5119,6 +5115,236 @@ func (t *Test) IgnoreDirButKeepFileInSubdir(
...
+!main.go
+`,
+			},
+		}
+
+		expected := []expectedDir{
+			{
+				path: ".",
+				want: "yamlFiles/\n",
+				// want: "sub.py\nyamlFiles/\n", // XXX: this seems wrong!
```
TomChv: Same here, change the test behaviour based on https://github.com/dagger/dagger/pull/10883/files#r2303934302

core/schema/host.go
```diff
@@ -278,49 +324,72 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
...
+		if !filepath.IsLocal(exclude) {
+			continue
+		}
+		exclude = filepath.Join(relPath, exclude)
+		if negative {
+			exclude = "!" + exclude
+		}
+		excludePatterns = append(excludePatterns, exclude)
+	}
+	localOpts = append(localOpts, llb.ExcludePatterns(excludePatterns))
```
TomChv, line 371: Can't we have `llb.WithGitIgnore` or something like that instead of that custom `[dagger.gitignore]` in the exclude patterns?

engine/sources/local/source.go
```diff
@@ -111,8 +111,24 @@ func (ls *localSource) Resolve(ctx context.Context, id source.Identifier, sm *se
 		return nil, fmt.Errorf("invalid local identifier %v", id)
 	}
 
+	gitIgnore := false
+	cachebuster := ""
+
+	excludes := make([]string, 0, len(localIdentifier.ExcludePatterns))
+	for _, pattern := range localIdentifier.ExcludePatterns {
+		if pattern == "[dagger.gitignore]" {
```
TomChv, line 121: Do this `[dagger.gitignore]` will be displayed on cloud UI or it will be ignored in the spans log?

I know this will not be part of the `exclude` parameters, but it may appears when you click on the  `Host.Directory` details.

Is there another way? Like by extending the llb localOpts in
module source? (See my comment above in `modulesource.go`)

core/schema/host.go
TomChv: Discussed on discord, we can remove that logic once we move host.Directory to DagOp :)

jedevc: As discussed, nope, since `llb` is in buildkit.

However, as we refactor `localSource` into a dagop, this'll get much neater.

engine/sources/local/source.go
jedevc: See below where we construct the `llb.WithCustomName` - we create what appears there, and deliberately *only* include the `args.Include`/`args.Exclude` there.

So the weird `[dagger.*]` won't be there.

jedevc:
Working on the failures, realized I had a misunderstanding of how the local/remote file interactions work

jedevc:
There we go :sweat_smile: It needed the corresponding changes to `Copy`.

Otherwise, we were syncing the remote -> local (using gitignore), but then filling up the snapshot using the *entire* contents of the local snapshot (not using the gitignore).

Just required a little bit of hackery, but it works now!

core/schema/host.go
```diff
@@ -278,49 +326,76 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
...
-	// To be removed via dagopification.
-	if args.NoCache {
-		args.Exclude = append(args.Exclude, rand.Text())
-	}
-
 	localName := fmt.Sprintf("upload %s from %s (client id: %s, session id: %s)", args.Path, stableID, clientMetadata.ClientID, clientMetadata.SessionID)
+
+	includePatterns := make([]string, 0, 2+len(args.Include))
+	if relPath != "." {
+		includePatterns = append(includePatterns, "!*", relPath)
```
TomChv, line 333: It took me a while to understand how you could give the host path (context dir) but still make it work if we're important a children dir of that ctx dir but I get it with that line.

So just to be sure I understand, let's say we have a repo at `/work` and you import `/work/bar/foo`.

hostPath will be: `/work`
relpath will be : `bar/foo`

Include: `["!*", "bar/foo"]` -> Everything is ignored except bar/foo

Then you do the rest of the logic with ignores etc...
Am I right or I misunderstood something?

jedevc: Yup, completely correct.

We also move the `include`/`exclude` paths to be "rebased".

I'll add a test for this, since it's a bit of a complex interaction. (edit: actually not needed, since all the existing include/exclude tests should cover this)

babo:
This change just broke my module constructor which refers to a file with a default name `.args` which is in the `.gitignore` to prevent it to accidentally checked in to the repository. How to resolve this?

`"failed to get value for argument "args": select: failed to stat file'

        // Constructor
        func New(ctx context.Context,
	        // environment variables file
	        // +defaultPath=".args"
	        // +optional
	        args *dagger.File) {...}

danielgafni:
https://github.com/dagger/dagger/issues/10994

nathan-joslin:
@jedevc Does this support git config's global `core.excludesfile` and `.git/info/exclude`? I am not seeing this mentioned in threads nor handled in the changes.

nathan-joslin:
@jedevc Does this handle `.git/info/exclude`? I do not see this mentioned in threads or the changes.

As a side note, git config's `core.excludesfile` is a third gitignore location. There likely is an argument that a global gitignore file is outside of dagger's sandbox and as such shouldn't be handled implicitly/by default (but potentially explicitly through flags, env vars, etc.)

jedevc:
Adding another note (for me, most likely) - I don't think this works as expected if we have `.gitignore` in the exclude patterns. Since it never makes it to the local clone, then the copy to the snapshot won't be done correctly.

staticaland:
Pardon my ignorance, but how do I go from:

```go
// Sync files to S3 bucket
func (m *Aws) SyncToBucket(
	ctx context.Context,
	bucket string,
	src *dagger.Directory,
	
...

	ctr = ctr.WithMountedDirectory("/work", src)
	
...
```

To this:

> Add NoGitAutoIgnore argument to Host.Directory to disable this behavior.

I don't know enough about Dagger to understand that yet 😅

## Context

MkDocs static site, where the output dir is in `.gitignore`. Pretty common pattern, that is for example included [in GitHub's Python `.gitignore` file](https://raw.githubusercontent.com/github/gitignore/master/Python.gitignore).

I've worked around this by using a separate directory when running in Dagger locally vs fully automated, where I don't expect any accidental commits of the built files to happen.

nathan-joslin:
@staticaland You may be interested in #10994.

You may be able to use the host functions directly to ensure the gitignore'd dir is available when needed. See [`func (*Host) Directory`](https://pkg.go.dev/dagger.io/dagger#Host.Directory).

PR #10886, opened 2025-08-17
feat: support OIDC auth with the Dagger Cloud API

matipan:
The problem: public repositories must expose their credentials if they wish to send telemetry Dagger Cloud. We expose our DAGGER_CLOUD_TOKEN in this repository. Since its only telemetry we don't worry much. But as we introduce compute-related features its critical that we have a secure mechanism to authenticate against api.dagger.cloud without exposing sensitive credentials.

To fix this issue we implement an OIDC authentication flow against our cloud API. To use this mechanism users need to specify `DAGGER_CLOUD_TOKEN=oidc` and be in a context where environment variables are correctly set for the supported providers. At the moment we only support GitHub as a provider. On top of specifying `DAGGER_CLOUD_TOKEN=oidc` users need to add the `id-token: write` permission to their workflow so that the dagger CLI can request an OIDC token from GitHub. 

Overall, the authentication flow is as follows:

image

- User sets `DAGGER_CLOUD_TOKEN="oidc"`
- Dagger CLI looks for environment variables of an OIDC provider. Right now only GitHub is supported through `id-token: write` workflow permissions and the `ACTIONS_ID_TOKEN_REQUEST_URL` env variable
- It queries the OIDC provider to get an OIDC JWT
- It sends that JWT to `api.dagger.cloud/v1/oidc`
- The API checks the token claims and looks for the `repository_owner` claim. It uses it to find an organization that has a source mapped to the `repository_owner`
- It creates a new JWT that can be used for all subsequent API calls
- It responds with the token plus the organization information so that the caller can know who is logged in.

In https://github.com/dagger/dagger.io/pull/4573 you'll find detailed information on the implementation as well as a demo showing how all works together in the context of a GitHub workflow.

### DIFF
--- a/engine/telemetry/url.go
+++ b/engine/telemetry/url.go
@@ -3,8 +3,6 @@ package telemetry
 import (
 	"context"
 	"fmt"
-	"os"
-	"strings"
 
 	"github.com/dagger/dagger/internal/cloud/auth"
 	"go.opentelemetry.io/otel/trace"
@@ -15,28 +13,9 @@ func URLForTrace(ctx context.Context) (url string, msg string, ok bool) {
 		return "", "", false
 	}
 
-	var orgName string
-	if cloudToken := os.Getenv("DAGGER_CLOUD_TOKEN"); cloudToken != "" {
-		// Try token auth first
-		token, ok := ParseDaggerToken(cloudToken)
-		if ok {
-			orgName = token.orgName
-		} else {
-			// DAGGER_CLOUD_TOKEN might be OIDC so try fetching
-			// the org from the state
-			org, err := auth.CurrentOrg()
-			if err != nil {
-				return "", "", false
-			}
-			orgName = org.Name
-		}
-	} else {
-		// Try OAuth next
-		org, err := auth.CurrentOrg()
-		if err != nil {
-			return "", "", false
-		}
-		orgName = org.Name
+	orgName, err := auth.CurrentOrgName()
+	if err != nil {
+		return "", "", false
 	}
 
 	if orgName == "" {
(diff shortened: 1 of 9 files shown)


engine/telemetry/url.go
```diff
@@ -21,6 +21,14 @@ func URLForTrace(ctx context.Context) (url string, msg string, ok bool) {
...
 		if ok {
 			orgName = token.orgName
+		} else {
+			// DAGGER_CLOUD_TOKEN might be OIDC so try fetching
+			// the org from the state
+			org, err := auth.CurrentOrg()
+			if err != nil {
+				return "", "", false
+			}
+			orgName = org.Name
```
jedevc: This logic is duplicated in a *lot* of places. Any chance we could write a helper? (or rework `auth.CurrentOrg` to inspect the contents of `DAGGER_CLOUD_TOKEN` if it's set)

Noticed a couple places where we're not consistently getting the org name:
- For attaching the current org to the client metadata, we're not actually using the explicitly set `DAGGER_CLOUD_TOKEN` at all (cc @marcosnils) https://github.com/dagger/dagger/blob/28043bb7c21924d5e0663abf6ef3a2b8e8591f6c/engine/client/client.go#L1167-L1170
- Similar logic here: https://github.com/dagger/dagger/blob/05b5eeda9473dac0bd68db2189cc2d3d48bccf50/engine/telemetry/cloud.go#L40-L57

internal/cloud/auth/auth.go
```diff
@@ -176,3 +187,114 @@ func SetCurrentOrg(org *Org) error {
 
 	return writeFile(orgFile, data, 0o600)
 }
+
+var (
+	once      sync.Once
```
jedevc:
```suggestion
	oidcOnce      sync.Once
```

engine/telemetry/url.go
marcosnils: yep. Makes sense. Let me send a quick refactor here 🙏

internal/cloud/auth/auth.go
```diff
@@ -164,6 +175,24 @@ func CurrentOrg() (*Org, error) {
...
+			orgName = token.orgName
+		} else {
+			org, err := CurrentOrg()
+			if err != nil {
+				return "", err
+			}
+			orgName = org.Name
+		}
+	} else {
+	}
```
jedevc: :eyes: Should we fill this with `CurrentOrg()` again?

jedevc (approved):
LGTM, with one comment.

jedevc:
DCO is failing :thinking:

matipan:
> DCO is failing 🤔

My bad. Started using jujutsu recently and haven't configured it to correctly sign off all commits

PR #10902, opened 2025-08-18
hack/build: always start with debug

vito:
Don't think there's any reason NOT to do this, helps to have the debug listeners set up before you know you need'em (e.g. a deadlock).

### DIFF
--- a/hack/build
+++ b/hack/build
@@ -28,7 +28,7 @@ engine_pid=$!
         ${_EXPERIMENTAL_DAGGER_GPU_SUPPORT:+--gpu-support} |\
     start \
         --name=$CONTAINER \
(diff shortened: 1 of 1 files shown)


vito:
@cwlbraa Nah - it literally just adds the `-p 6060:6060` flag to `docker start`:

https://github.com/dagger/dagger/blob/29f117fc4577b252ca89ae04d34259a68b7ad342/cmd/engine/.dagger/docker.go#L129-L131

I guess there's a risk here for port conflicts if you've already got something bound on 6060 locally. Not sure what happens then.

vito:
(feel free to revert if this causes any issues obviously)

PR #10905, opened 2025-08-18
chore: bump internal tooling to v0.18.15

sipsma:


sipsma:
CI failures are due to the engine failing to start w/ `sed: write error`, fix with more details here: https://github.com/dagger/dagger/pull/10906

Unfortunately we may need to do another release before we can upgrade CI since it's causing tests to fail to start the dev engine >50% of the time currently

jedevc:
Succeeded by https://github.com/dagger/dagger/pull/10912.

Issue #10817, opened 2025-07-30
Calling constructor with `ignore`+`defaultPath` directives are not applied


jedevc:
## Setup

Assume the following module:

```go
package main

import (
	"context"
	"dagger/foo/internal/dagger"
)

type Foo struct {
	Entries []string
}

func New(
	ctx context.Context,
	// description goes here
	// +defaultPath="./arg"
	// +ignore=["foo.txt"]
	myArg *dagger.Directory,
) (*Foo, error) {
	entries, err := myArg.Entries(ctx)
	if err != nil {
		return nil, err
	}
	return &Foo{entries}, nil
}
```

With the following testdata:

```
$ mkdir arg
$ touch arg/{foo,bar}.txt
```

Get the typedefs for them:

```
$ dagger query </path/to/dagger/cmd/dagger/typedefs.graphql > out.json
```

Get the typedef for `Foo`'s constructor:

```
$ jq '.typeDefs[] | select(.asObject.name == "Foo") | .asObject.constructor' < out.json
```
```json
{
  "args": [
    {
      "defaultPath": "",
      "defaultValue": null,
      "description": "description goes here",
      "ignore": [
        "foo.txt"
      ],
      "name": "myArg",
      "typeDef": {
        "asEnum": null,
        "asInput": null,
        "asInterface": null,
        "asList": null,
        "asObject": {
          "name": "Directory"
        },
        "asScalar": null,
        "kind": "OBJECT_KIND",
        "optional": false
      }
    }
  ],
  "description": "",
  "name": "",
  "returnType": {
    "asEnum": null,
    "asInput": null,
    "asInterface": null,
    "asList": null,
    "asObject": {
      "name": "Foo"
    },
    "asScalar": null,
    "kind": "OBJECT_KIND",
    "optional": false
  }
}
```

Get the typedef for `Query.foo`:

```
$ jq '.typeDefs[] | select(.asObject.name == "Query") | .asObject.functions[] | select(.name == "foo")' < out.json
```
```json
{
  "args": [
    {
      "defaultPath": "",
      "defaultValue": null,
      "description": "\ndescription goes here\n",
      "ignore": [],
      "name": "myArg",
      "typeDef": {
        "asEnum": null,
        "asInput": null,
        "asInterface": null,
        "asList": null,
        "asObject": {
          "name": "Directory"
        },
        "asScalar": null,
        "kind": "OBJECT_KIND",
        "optional": false
      }
    }
  ],
  "description": "",
[...]

PR #10907, opened 2025-08-19
LLM: MCP servers, contextual Env filesystem, better TUI

vito:
* [ ] Bikeshed [API changes](https://github.com/dagger/dagger/pull/10907/files#diff-6d45938d3a99ad730e8d639cd2218f57c90828f1deeaad17c073cad22bfd0b50)

This PR adds everything you need to implement coding agents to a level of sophistication much closer to Claude Code. I'm maintaining a Dagger-native (sandboxed!) coding agent called [Doug](https://github.com/vito/daggerverse/blob/5896f2a5500e16ca57e5d1f1ae79324a1bf691b5/doug/main.go).

You can use Doug in two ways - the first way is to invoke it directly by passing it your workspace and optionally a module to act as the build/test toolchain:

```sh
git clone https://github.com/vito/dang
cd dang
dagger -m github.com/vito/daggerverse/doug call dev --source . --module .
```

The second way is to add a function to your module and build-your-own coding agent, using Doug as a library. This is where you can additionally configure MCP servers, or whatever other prompting and agent setup you might want. For Dang's coding agent, I have it pre-configured with an LSP MCP server:

> I have two apologies to give: the first is for this example being in my bespoke language sideproject, and the second is for their names being so similar. Dang = the language, Doug = the agent.

```graphql
pub coder: LLM! {
  doug.agent(
    llm.withEnv(env.withCurrentModule.withWorkspace(source)).withMCPServer(
      "lsp",
      container.
        from("golang").
        withExec(["go", "install", "golang.org/x/tools/gopls@latest"]).
        withExec(["go", "install", "github.com/isaacphi/mcp-language-server@latest"]).
        withWorkdir("/src").
        withDefaultArgs(["mcp-language-server", "--workspace", "/src", "--lsp", "gopls"]).
        asService
    )
  )
}
```

With that, you can just call `coder`:

```
dagger call coder
```

Calling a function that returns an LLM now pops you into interactive prompt mode, where you can try something like this:

```
> Implement support for the && operator. Skip short-circuiting for now to keep it simple.
```

From here you should see the agent draft a TODO list, do some research to understand the project (perhaps in a sub-Task, just like Claude Code), start writing code, adding tests, running tests, and finally stop once the tests are passing. Now you can press <kbd>Ctrl+S</kbd> to save the changes to your workdir.

## Summary of Changes

* `Env` now has a filesystem, which automatically propagates between tool calls and gets re-hydrated as contextual args to modules
* `LLM` now supports MCP servers (in other words, `LLM` is an MCP client), which _also_ run with the same filesystem, and even propagate changes out to subsequent tool calls
* `LLM` now captures and displays logs in each tool call response (supersedes #10768)
* `LLM` now exposes its tools in "dynamic mode" by default, like it used to. The static tool calling scheme (`call_method`, `chain_methods`) are now opt-in, and `dagger mcp` now opts in.
* `LLM` now supports "blocking" APIs, so they're not exposed to the model
* `Env.withModule` installs a module into the env, exposing its functions as tools, with automatic hostfs propagation
* The TUI now displays total cost of current session in a sidebar.
* A few new APIs to faciliate implementing coding agents efficiently in Dagger.

## Testing strategy

* There are a ton of internal plumbing changes to how tool calls are represented to the agent. These are covered by the existing evals suite.
* There's a new eval specifically for MCP servers, which uses Claude Code as the MCP server, since that's a well-known stable one.
* Doug has its own evals, which I'm relying on for covering a lot of the new workspace propagation functionality. It provides great coverage but ideally we'd have coverage within our own suite. I'm punting on that for now, pending a decision whether we should just pull Doug into our repo.

You can run Doug's evals like so:

```sh
dagger -m github.com/vito/daggerverse/doug/evals call check
```

For now these only run against Claude Sonnet 4.0 - the other models just don't seem good enough.

## Demos

Here's Claude Code, running as an MCP server, used alongside an LSP MCP server:

[asciicast](https://asciinema.org/a/733583)

Here's Doug, a custom coding agent inspired by Claude and the rest, but implemented fully natively in Dagger:

[asciicast](https://asciinema.org/a/Ke04q94Pus0Ffpib45MZWv6hh)

## Future plans

* Auto-sync changes (Jesus take the wheel mode)
* Viewing a diff of changes, instead of just the summary (same for `Changeset` syncing in CLI)
* Syncing changes in the other direction (what hotkey?)
* Zed ACP integration? Need to research it.


docs/docs-graphql/schema.graphqls
```diff
@@ -1,3 +1,6 @@
+"""Indicates that the argument defaults to a contextual path."""
+directive @defaultPath(path: String!, ignore: [String!]!) on ARGUMENT_DEFINITION
```
vito: This is exposed and annotated on contextual args so that the inner MCP implementation can determine which arguments are contextual and satisfy them using the Env hostfs.

docs/docs-graphql/schema.graphqls
```diff
@@ -40,9 +43,6 @@ type Binding {
   """Retrieve the binding value, as type JSONValue"""
   asJSONValue: JSONValue!
 
-  """Retrieve the binding value, as type LLM"""
-  asLLM: LLM!
```
vito, line 95: LLM is now a banned type for LLMs. This prevents them from calling methods that start agents. Giving them that power unchecked (i.e. not wrapped in a simpler API that uses the LLM internally) always causes them to go off the rails.

docs/docs-graphql/schema.graphqls
```diff
@@ -1317,6 +1317,9 @@ type CurrentModule {
   """A unique identifier for this CurrentModule."""
   id: CurrentModuleID!
 
+  """The fully instantiated module implementing the current function call."""
+  meta: Module!
```
vito: Might need a better name here. 😅

This is exposed so that modules can inject themselves into an Env, e.g. `dag.Env().WithModule(dag.CurrentModule().Meta())`.

We could _almost_ avoid this by having `dag.CurrentEnv()` automatically have the current module injected, except I immediately had a case where I wanted to pass that Env to _another_ module and let it inject itself. (Use case: inject build/test module into Env, pass it to Doug module to gain its coding agent tools.)

Another tactic could be adding `Env.withCurrentModule` instead.

But, there may be other reasons to want access to the current `Module`, so I'm on the fence: do we just bite the bullet and find a good name for it?

docs/docs-graphql/schema.graphqls
```diff
@@ -1831,6 +1834,8 @@ The `EnumValueTypeDefID` scalar type represents an identifier for an object of t
 scalar EnumValueTypeDefID
 
 type Env {
+  hostfs: Directory!
```
vito: I kept this name from the original #10370 proposal, but I'm not sure how I feel about it, vs. something like `workdir` or `workspace`. The "host" bit is what seems weird to me. I don't feel strongly though.

docs/docs-graphql/schema.graphqls
```diff
@@ -2037,26 +2048,8 @@ type Env {
...
-  """Declare a desired LLM output to be assigned in the environment"""
-  withLLMOutput(
-    """The name of the binding"""
-    name: String!
-
-    """A description of the desired value of the binding"""
-    description: String!
-  ): Env!
+  """load a module and expose its functions to the model"""
+  withModule(module: ModuleID!): Env!
```
vito, line 2340: This is analogous to passing the `-m` flag on the CLI, multiple times if it supported that.

Instead of passing in a pre-constructed `DaggerDev` object, you install the `DaggerDev` module. Doing this makes its functions available as tools. Calling those functions implicitly constructs the module with the current Env hostfs used to satisfy any contextual args. Strong parallels to the shell here.

docs/docs-graphql/schema.graphqls
```diff
@@ -2210,6 +2203,9 @@ type Env {
     """The description of the output"""
     description: String!
   ): Env!
+
+  """Return a new environment without any outputs"""
+  withoutOutputs: Env!
```
vito, line 2548: This is used to implement _sub-tasks_ in a coding agent.

Claude and other coding agents have a `Task` tool which takes a prompt, runs an LLM, and returns a string reply. Critically, that LLM runs in the same Env, so that it sees the same filesystem and tools, _but_ we don't want that sub-Env to have the same "desired outputs" as the outer one, because its purpose is to just return a text reply.

Not sure how to square this with the deprecation of inputs/outputs in #10744

docs/docs-graphql/schema.graphqls
```diff
@@ -2958,6 +2957,9 @@ type LLM {
   """return the provider used by the llm"""
   provider: String!
 
+  """Returns an LLM that will only sync one step instead of looping"""
+  step: LLM!
```
vito: Currently, `loop` doesn't give the client any opportunity to monitor progress; it blocks until the agent ends its turn.

So, this `step` API allows the client to perform just one iteration of the loop, show the user the changes (e.g. in the TUI sidebar), and call `hasPrompt` to determine whether it should keep looping.

docs/docs-graphql/schema.graphqls
```diff
@@ -2967,9 +2969,38 @@ type LLM {
...
   tools: String!
 
+  """Return a new LLM with the specified tool disabled"""
+  withBlockedFunction(
+    """The type name whose field will be disabled"""
+    typeName: String!
+
+    """The field name to disable"""
+    fieldName: String!
+  ): LLM!
```
vito, line 3491: This API allows a module to pass itself in, but disable a few of its functions that might make the agent go haywire.

This is also used to implement sub-tasks, to allow sub-tasks to inherit the same Env but prevent them from spawning more sub-tasks (`withBlockedFunction("Doug", "task")`) because they tend to just keep going deeper and deeper and deeper.

I would have loved to make this more strongly typed somehow, but it at least results in a runtime error if you reference an unknown function.

docs/docs-graphql/schema.graphqls
```diff
@@ -2967,9 +2969,38 @@ type LLM {
...
+
+  """Provide the calling object as an input to the LLM environment"""
+  withCaller(
+    """The name of the binding"""
+    name: String!
+
+    """The description of the input"""
+    description: String!
+  ): LLM!
```
vito: This was superseded by `withModule`, should probably just remove it. 🤔

docs/docs-graphql/schema.graphqls
```diff
@@ -2967,9 +2969,38 @@ type LLM {
...
   """allow the LLM to interact with an environment via MCP"""
   withEnv(env: EnvID!): LLM!
 
+  """Add an external MCP server to the LLM"""
+  withMCPServer(
+    """The name of the MCP server"""
+    name: String!
+
+    """
+    The MCP service to run. If the service exposes a port, HTTP+SSE will be used to communicate.
```
vito: (TODO: that second part isn't implemented yet.)

docs/docs-graphql/schema.graphqls
```diff
@@ -2988,6 +3019,11 @@ type LLM {
     file: FileID!
   ): LLM!
 
+  """
+  Use a static set of tools for method calls, e.g. for MCP clients that do not support dynamic tool registration
+  """
+  withStaticTools: LLM!
```
vito, line 3526: In my testing it felt a lot more natural to just expose functions directly as tools to agents. Very vibes-based, but the thought is that _our_ agent implementation fully supports dynamically changing tools, and it's the simplest way to expose tools to agents, so that should _probably_ be the default.

The only issue is when we're exposed over something that doesn't support dynamic tool registration, so that's what this API is for, and now it lives on `LLM` instead of `Env`, to decouple it from MCP/LLM concerns.

docs/docs-graphql/schema.graphqls
```diff
@@ -3488,6 +3524,8 @@ type Query {
     platform: Platform
   ): Container!
 
+  currentEnv: Env!
```
vito: `dag.CurrentEnv()` is a new API to return the current evaluation environment. In a plain module function call, `dag.CurrentEnv().Hostfs()` will return the contextual directory. When called from an LLM, `dag.CurrentEnv()` directly exposes the MCP's `Env`.

vito (commented):
API self-review to kick off the bikeshedding.

cc @shykes

docs/docs-graphql/schema.graphqls
jedevc: Heh nice.

I think I thought about doing this as part of https://github.com/dagger/dagger/issues/10817 - currently, we lose `defaultPath` and `ignore` in the constructor typedefs - because there's no way to channel those through graphql.

cmd/container-mount/main.c
```diff
@@ -0,0 +1,112 @@
...
+    if (container_pid <= 0) {
+        fprintf(stderr, "Error: Invalid container PID: %s\n", argv[1]);
+        return 1;
+    }
+
+    // Step 1: Create a detached clone of the source mount
+#ifdef DEBUG
+    printf("Creating detached mount of %s...\n", source_path);
+#endif
+    int fd_mnt = open_tree(AT_FDCWD, source_path,
```
sipsma: I didn't look at the history of each of the syscalls here yet, but I know some of them are relatively recent additions, which would likely bump our minimum required kernel version quite a bit.

I'm not inherently opposed to that but it's probably worth doing consciously and checking if there's any significant LTS releases of the kernel and/or major distros that we'd be ruling out now.

vito: I think we're OK - these were added in Linux 5.2, which came out in 2019. All major distros have an LTS with at least that version, and the oldest Linux kernel version still in support is 5.4:

* https://claude.ai/share/0f79e5b8-4053-4c33-8dc1-27463abd8cda
* https://endoflife.date/linux

shykes: Is there an explainer on this feature somewhere? @vito I think you said it's tightly coupled to MCP? That kind of worries me, but getting a better overall understanding of the API changes would help

vito: @shykes There are no API changes for this specific part; it's all internal to the LLM/MCP plumbing.

There's an explainer of the mechanics here:

https://github.com/dagger/dagger/blob/ed7eaa31c3e1b39db25d4818884ef4c003edb27f/core/service.go#L941-L950

If I said "tightly coupled" before, I probably just meant that instead of adding APIs for this, as I initially planned, I opted to leave it as an internal implementation detail of MCP for now, that way we can bikeshed a proper API later if/when we want it.

The specific need for MCP is pretty nuanced:

1. I have a running service,
2. I need to remount a possibly updated directory into the service,
3. then make a call out-of-band to the service,
4. then snapshot the changes the service made to the directory (if any),
5. then remount a read-write copy of that snapshot so the service can't keep changing the immutable snapshot

Other use cases aren't so nuanced. For updating services with live directories for example, we only really need the first two steps. MCP servers are a strange case where we need _outputs_ from the service, too.

In either case, there's a chance we might need to use something dumber like `rsync` instead to trigger fsnotify changes for maximal compatiblity. But I haven't been able to prove that yet. It'd be nice if we can just remount all the time.

sipsma: > I think we're OK - these were added in Linux 5.2, which came out in 2019. All major distros have an LTS with at least that version, and the oldest Linux kernel version still in support is 5.4:

Nice, SGTM to then

core/schema/env.go
```diff
@@ -32,8 +33,17 @@ func (s environmentSchema) Install(srv *dagql.Server) {
...
+			Doc("Return a new environment without any outputs"),
 		dagql.Func("output", s.output).
 			Doc("retrieve an output value by name"),
+		dagql.Func("withWorkspace", s.withWorkspace).
+			Doc("Return a new environment with a new host filesystem").
+			Args(
+				dagql.Arg("workspace").Doc("The directory to set as the host filesystem"),
+			),
+		dagql.Func("withModule", s.withModule).
+			Doc("load a module and expose its functions to the model"),
```
jedevc: nit: casing (not just this one, some others in this file as well)

docs/docs-graphql/schema.graphqls
jedevc: So I think there are potentially some weird caching things if we have this - currently, there's an attempt to avoid loading the same "module" twice. So if you load a module from two different commits, but it's not changed, they should resolve to the same module.

But by exposing this here, you can access `dag.CurrentModule().Meta().Source().Commit()` - which gets to the same problem as @sipsma describes in https://github.com/dagger/dagger/pull/10847#issuecomment-3229134157.

dagql/idtui/frontend_pretty.go
```diff
@@ -1921,82 +2123,154 @@ func (fe *frontendPretty) renderStep(out TermOutput, r *renderer, row *dagui.Tra
...
-	} else if span.IsFailedOrCausedFailure() {
-		toggler = out.String(IconFailure)
-	} else if span.IsPending() {
-		toggler = out.String(DotEmpty)
-	} else {
-		toggler = out.String(DotFilled)
-	}
-	if span.ActorEmoji != "" && span.Message != "" {
-		toggler = out.String(span.ActorEmoji)
-	}
```
jedevc, line 1952: Should we stop sending `ActorEmoji`?

dagql/idtui/testdata/TestTelemetry/TestGolden/typescript/pending
```diff
@@ -1,26 +1,26 @@
...
-├─● .withExec(args: ["sleep", "1"]): Container! X.Xs
-├─✘ .withExec(args: ["false"]): Container! X.Xs ERROR
-│ ! process "false" did not complete successfully: exit code: 1
-╰─○ .withExec(args: ["sleep", "1"]): Container! X.Xs
+  $ Container.from(address: "alpine:latest"): Container! X.Xs CACHED
+  ✔ .withEnvVariable(name: "NOW", value: "XXXX-XX-XXTXX:XX:XX.XXXZ"): Container! X.Xs
+  ✔ .withExec(args: ["sleep", "1"]): Container! X.Xs
+  ✘ .withExec(args: ["false"]): Container! X.Xs ERROR
+  ! process "false" did not complete successfully: exit code: 1
+  ○ .withExec(args: ["sleep", "1"]): Container! X.Xs
```
jedevc, line 24: nit: with the completed state now being a tick, the pending state being an empty circle (the empty version of the completed state) makes less sense IMO. No good suggestions though.

dagql/server.go
```diff
@@ -268,6 +268,23 @@ var coreDirectives = []DirectiveSpec{
...
+		Description: FormatDescription(`Indicates that the argument defaults to a contextual path.`),
+		Args: NewInputSpecs(
+			InputSpec{
+				Name: "path",
+				Type: String(""),
+			},
+			InputSpec{
+				Name: "ignore",
+				Type: ArrayInput[String](nil),
+			},
```
jedevc, line 297: I think this should be a a separate directive.

We can have `ignore` without a `defaultPath`, which indicates to filter it out on function call.

modules/claude/main.go
```diff
@@ -0,0 +1,53 @@
+//
```
jedevc: :eyes:

dagql/idtui/testdata/TestTelemetry/TestGolden/typescript/pending
vito: Yeah true. Hmmm... Which do you prefer? I found myself missing the checkmark, feels a little more like progress is being made instead of just a bunch of dots

I don't think 'empty circle' => 'half circle' => progress bar gets replaced with a checkmark is TOO big of a stretch, though

dagql/idtui/frontend_pretty.go
vito: I don't think we need to. Some UIs might find it useful, just the current TUI iteration doesn't leverage it. The web UI still uses it - at least, for now

docs/docs-graphql/schema.graphqls
vito: I think I can circumvent all this weirdness by just adding an `Env.withCurrentModule` - gonna do that now

jedevc (approved):

docs/docs-graphql/schema.graphqls
shykes: @vito quick questions:

- Is it correct that there are no conflicts between modules, because each module is exposed as a single top-level tool, which calls the module constructor?
- Environment workspace: this is a one-to-many situation? The same env workspace is passed to all modules installed in the env, and all module's contextual args will loaded from there? Won't that break if several modules receive contextual dirs, since those assume a single module?

I'm still stuck on the fear of taking a complex mental model for filesystems, and making it even more intractable. I know we already discussed this... But we hadn't discussed the "one env workspace, multiple module contexts" edge case. It's kind of breaking my brain.

docs/docs-graphql/schema.graphqls
shykes: I'm going to pick my battles and leave the issue of input/output deprecation for a later PR...

docs/docs-graphql/schema.graphqls
```diff
@@ -3426,6 +3433,9 @@ type LLM {
   """return the LLM's current environment"""
   env: Env!
 
+  """Indicates that the LLM can be synced or stepped"""
+  hasPrompt: Boolean!
```
shykes, line 3439: Walking in without context: I don't understand what `hasPrompt` does exactly, why it's called that, or the relationship with the description.

docs/docs-graphql/schema.graphqls
```diff
@@ -3447,6 +3457,9 @@ type LLM {
   """return the provider used by the llm"""
   provider: String!
 
+  """Returns an LLM that will only sync one step instead of looping"""
+  step: LLM!
```
shykes: Does `step` actually perform the step? Or does it change the LLM configuration, so that it will step once? If so, what will trigger the step?

IMO either the description is misleading, or the name is wrong.

docs/docs-graphql/schema.graphqls
shykes: - I'm guessing you already thought of it, but: could this be an argument to `WithModule`?
- The terminology is inconsistent: are we disabling a tool, blocking a function, or disabling a field?

Re: making it more strongly typed. You gotta do what you gotta do 😁

docs/docs-graphql/schema.graphqls
```diff
@@ -3456,9 +3469,27 @@ type LLM {
...
+
   """allow the LLM to interact with an environment via MCP"""
   withEnv(env: EnvID!): LLM!
 
+  """Add an external MCP server to the LLM"""
+  withMCPServer(
+    """The name of the MCP server"""
+    name: String!
+
+    """The MCP service to run and communicate with over stdio"""
```
shykes: How tied to stdio? Could this support http servers in the future, and if so, would the API be the same?

docs/docs-graphql/schema.graphqls
shykes: If this is for graceful handling of certain MCP clients, doesn't that mean it's tied to our MCP server implementation? And if so, isn't LLM supposed to be detached from those MCP concerns?

docs/docs-graphql/schema.graphqls
shykes: Is this the one we talked about disabling for now, to minimize entanglement with other ongoing architecture discussions?

shykes (approved):
Optimistically approving, I left a few comments regarding:

1. Function naming nits
2. The concept of env workspace and how it relates to module context, in particular with the edge case of multiple modules per env. Are "module context" and "env workspace" really the same thing?
3. `LLM.currentEnv` which breaks my brain, and I think you suggested we disable for now, to cross as few one-way doors as possible?

docs/docs-graphql/schema.graphqls
vito: I looked into this and had something that might have worked, but I backed out after I looked around and it seemed like the standard is moving away from SSE and towards some other HTTP protocol, so now I don't even know how to test it. I figure we can add support for this once things are a bit more stable.

docs/docs-graphql/schema.graphqls
vito: Updated and revised docs - now it returns an `LLMID`, so behaves like `sync`. Much less awkward. (It works by using an internal `__step` API under the hood.)

docs/docs-graphql/schema.graphqls
vito: Revised

docs/docs-graphql/schema.graphqls
vito: I've kept it for now since it's actively used by Doug for storing a TODO list in the environment as an input. As a compromise I've marked it experimental with a spooky warning.

I doubt this is the final form - there were a few times I wanted a module to have a space for storing arbitrary fields, and logically what would probably make the most sense is storing it on the `Doug` object itself, but that's not currently possible because the module gets re-constructed fresh with only the workspace on every tool call. Saving this for a future iteration.

docs/docs-graphql/schema.graphqls
vito: > I'm guessing you already thought of it, but: could this be an argument to WithModule?

`WithModule` and `WithCurrentModule` (which is probably where this would be used most of the time) live on `Env`, so controlling it there felt like coupling `Env` too tightly to LLM/MCP. I don't think we'd need this elsewhere, like in shell. :thinking: 

> * The terminology is inconsistent: are we disabling a tool, blocking a function, or disabling a field?

I went with "function" over "tool" because that's the mode of referencing the thing to be blocked. We don't need the ability to block "fields" as the user knows them since they don't manifest as tools anyway.

For blocking an MCP tool for example we might want `withBlockedMCPTool(server: String!, toolName: String!)` later.

`fieldName` should be `function` and I should choose one of `block` or `disable` though, will fix those.

docs/docs-graphql/schema.graphqls
vito: For this PR, I'm decoupling *Env* from LLM/MCP, not LLM from MCP. I don't think this is inherently coupled to MCP, either; LLM in general is concerned with what tools are available and how they're exposed, this is just another way to tweak it that might be beneficial to some models in theory, or required for some constraints (like MCP clients that do not support dynamic tools).

vito:
I think I've addressed all bikeshedding feedback so once this PR is green I'm gonna bias towards merging, we can tweak anything else in follow-ups :pray:

docs/docs-graphql/schema.graphqls
vito: Oops forgot to reply to this comment

> * Is it correct that there are no conflicts between modules, because each module is exposed as a single top-level tool, which calls the module constructor?

All of the module's top-level functions are exposed, and each of them auto-call the module constructor with the current context and then call the function against that instance. Any objects returned by those functions, like `DaggerDevTest#1`, are also magically dynamic/contextual, so you don't get stuck calling `Specific` against a stale `DaggerDevTest` instance after changing the workspace. That part is necessary to support chained APIs like ours.

The module tools are optimistically exposed with simple names like e.g. `build` for best model intuition, but if there are two `Build` functions the second one will be qualified as `TypeName.build`.

> Environment workspace: this is a one-to-many situation? The same env workspace is passed to all modules installed in the env, and all module's contextual args will loaded from there? Won't that break if several modules receive contextual dirs, since those assume a single module?

It will break if you install a module that doesn't know what to do with that contextual dir. The solution is to not do that. :upside_down_face: You should only use this for modules that can work with the same workspace.

---

One env workspace, multiple module contexts is pretty fundamental to the coding agent use case; it's how you're able to install a generic coding agent module (Doug) and give it a toolchain curated for your project.

I wouldn't bet that this is all perfect, my goal is to get _something_ useful out there and keep iterating based on how it's used and what challenges we run into. To be honest I still consider _all_ of the LLM APIs to be pretty experimental.

PR #10916, opened 2025-08-20
ci: migrate php docs gen to sdk

jedevc:
This moves php docs generation out of the docs generator and into the sdk generation. It could kind of have gone both places - but there's a significant advantage to having it in the SDK.

We need to use the generated php code as the input to the doctum generation step - if we don't, we actually don't end up updating the docs in one pass, it requires *two* generation passes, which is super annoying (and means you end up running `generate` steps twice).

This fixes that issue!


vikram-dagger:
Looks good. I see some CI failures, unsure if those are related. Approving in advance.

PR #10924, opened 2025-08-21
feat: add ability to combine outputs in WithExec

### LINKED ISSUE #3496: Support for combined stdout and stderr output
There's many times you want to get both the stdout and stderr. It's slightly tedious to do this with the Go SDK since you have to grab them both separately (and check err, etc.). Maybe there could just be a field `Output` that grabs both for you at the same time?

eunomie:
Allow to get a combined output of both standard output and standard error at once.

A new function `combinedOutput` is available, in addition to `stdout` and `stderr`. It means you can pick the one you want depending on the context. For instance to pick the combined output if the command failed.

```console
$ dagger -q <<EOF
container | from alpine |
with-new-file /test.sh "echo out; echo err >&2" |
with-exec -- sh /test.sh |
combined-output
EOF

out
err
```

> [!WARNING]
> As always while working with streams, order in the output is not guaranteed.

fixes #3496

### DIFF
--- a/core/container_exec.go
+++ b/core/container_exec.go
@@ -450,6 +450,10 @@ func (container *Container) Stderr(ctx context.Context) (string, error) {
 	return container.metaFileContents(ctx, buildkit.MetaMountStderrPath)
 }
 
+func (container *Container) CombinedOutput(ctx context.Context) (string, error) {
+	return container.metaFileContents(ctx, buildkit.MetaMountCombinedOutputPath)
+}
+
 func (container *Container) ExitCode(ctx context.Context) (int, error) {
 	contents, err := container.metaFileContents(ctx, buildkit.MetaMountExitCodePath)
 	if err != nil {
--- a/.changes/unreleased/Added-20250822-112220.yaml
+++ b/.changes/unreleased/Added-20250822-112220.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: Allow to get a combined output of both standard output and standard error after a withExec on a container
+time: 2025-08-22T11:22:20.827952+02:00
+custom:
+    Author: eunomie
+    PR: "10924"
--- a/core/integration/container_test.go
+++ b/core/integration/container_test.go
@@ -212,6 +212,30 @@ func (ContainerSuite) TestExecStdoutStderr(ctx context.Context, t *testctx.T) {
 	})
 }
 
+func (ContainerSuite) TestExecCombinedOutput(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	ctr := c.Container().
+		From(alpineImage).
+		WithNewFile("/test.sh", `echo "out"
+echo "err" >&2
+`).
+		WithExec([]string{"sh", "/test.sh"})
+	out, err := ctr.Stdout(ctx)
+	require.NoError(t, err)
+	require.Equal(t, out, "out\n")
+
+	out, err = ctr.Stderr(ctx)
+	require.NoError(t, err)
+	require.Contains(t, out, "err\n")
+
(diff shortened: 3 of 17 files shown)


core/container_exec.go
```diff
@@ -57,6 +57,9 @@ type ContainerExecOpts struct {
 	// Redirect the command's standard error to a file in the container
 	RedirectStderr string `default:""`
 
+	// Combine the command's standard output and standard error to the standard output
+	CombinedOutput bool `default:"false"`
```
eunomie: suggestion: if we keep this behaviour, indicates it's `stderr` redirected to `stdout` and so combined output is available under `stdout` but not under `stderr`


```suggestion
	RedirectStderrToStdout bool `default:"false"`
```

eunomie:
The PR's behaviour is to redirect standard error to standard output so combined output can be retrieved using `stdout`.
Currently standard error is still available independently by calling `stderr`.

❓ Should `stderr` be empty if it's redirected to standard output?
Other option would be to have `stderr` returning the exact same content than `stdout` (and so keep the `CombinedOutput` flag name)

vito:
Nice, wanted this for a while, thanks for picking it up! Related issue: #3496 

API wise it feels confusing to call `Container.stdout` and get both (even though it's opt-in).

I think we could change the engine to just always write a combined output file in addition to the separate ones, and just add a getter for it:

https://github.com/dagger/dagger/blob/d6553a4da0e87a81c3dbcbddc4a94a1ab2743d90/engine/buildkit/executor_spec.go#L572-L584

Then you can just call `Container.output` without having to opt-in ahead of time.

We'll be storing extra data for each command but it seems a small price to pay for the simplicity imo.

eunomie:
> We'll be storing extra data for each command but it seems a small price to pay for the simplicity imo.

That's why I haven't done it at start, but if it sounds like a better approach, I'll do that 👍 

> Then you can just call `Container.output` without having to opt-in ahead of time.

I think I would prefer something like `Container.combinedOutput`. It's more verbose for sure, but it explains itself better.

Thinking a bit more about that, this also means we can retrieve one or the other after the `withExec`, and that's probably way better. I can imagine some cases where I run a `withExec` to get the output, in a normal case. But if something goes wrong, for instance if the exit code is not the one I expected, I'd like to get the combined output to print it to the user.
By having a flag in the `withExec` this means I have to choose up front. By having a `combinedOutput` function, it means I can pick the one I need when I want.

Thanks for the suggestion, I'll update that 😃

core/integration/module_test.go
```diff
@@ -5119,6 +5119,93 @@ func (t *Test) IgnoreDirButKeepFileInSubdir(
 	})
 }
 
+func (ModuleSuite) TestCombinedOutput(ctx context.Context, t *testctx.T) {
+	tcs := []struct {
+		sdk    string
+		source string
```
jedevc: IMO, this API should be tested as part of `ContainerSuite`.

We also don't need to test each SDK separately, since it's an API entirely in the engine, we can just use the client available in `connect` directly.

jedevc:
Hey, looks nice! :eyes:

core/integration/module_test.go
eunomie: Sure, I'll update that 👍

eunomie: I moved the test on `ContainerSuite`, it's way better there, thanks :-)

eunomie:
I pushed a new version:
- with a `combinedOutput` func (like `stdout` and `stderr`) instead of a configuration on `withExec`
- always write both stdout and stderr to a new file to be able to retrieve the combined output
- move test to `ContainerSuite`

jedevc (approved):

PR #10930, opened 2025-08-22
feat: extract git info from github event

jedevc:


engine/telemetry/labels.go
```diff
@@ -259,8 +266,43 @@ func (labels Labels) WithGitHubLabels() Labels {
 		labels["dagger.io/vcs.change.branch"] = pr.GetHead().GetRef()
 		labels["dagger.io/vcs.change.head_sha"] = pr.GetHead().GetSHA()
 		labels["dagger.io/vcs.change.label"] = pr.GetHead().GetLabel()
+
+		gitLabels["dagger.io/git.ref"] = pr.GetHead().GetSHA()
```
jedevc, line 271: Note: there's no way of getting the author/committer/commit message information here. Best we can do is extract the user name+email of the *person* who pushed, but not the commit info.

Maybe we could put that in `vcs`?

marcosnils: SGTM I guess. The same thing could happen with regular pushes, correct? i.e: the user pushing not being the same as the one authoring the commits in the push. I guess eventually we need to revisit this since we're taking the HEAD commit author rn but ideally it might be worth to track all the involved contributors in a pr / push.

jedevc:
Added some new assertions! Thanks for the advice, didn't realize we had those.

Issue #10913, opened 2025-08-19
🐞 DefaultPath for Files loads the file's whole directory


kpenfound:
### What is the issue?

Currently DefaultPath for Files will load the entire directory containing the file and then select the file from it https://github.com/dagger/dagger/blob/v0.18.16/core/modfunc.go#L644
So `Host().Directory(dir).File(file)`

This is surprising when that directory is large and takes a long time when we only want one file

The desired behavior is
`Host().File(file)`
or if we must load the directory for some reason,
`Host.Directory(dir, include:file).File(file)`

### Dagger version

v0.18.16

### Steps to reproduce

Repro here https://github.com/esafak/dagger-defaultpath

PR #10931, opened 2025-08-22
fix: speed up loading contextual files

jedevc:
Fixes https://github.com/dagger/dagger/issues/10913

Ideally, we'd just call `host.file` here, but then we'd have to reproduce all of the other logic in `LoadContext`, which doesn't sound incredibly fun.

So this is a quick fix!


eunomie (commented):
Golden test is not passing because of the new `include`

```diff
                -  ┆ contextDir: Host.directory(path: "/app/dagql/idtui/viztest", exclude: [], noCache: true): Directory!
                +  ┆ contextDir: Host.directory(path: "/app/dagql/idtui/viztest", exclude: [], include: [], noCache: true): Directory!
```

But other than that, this sounds good to my current knowledge :-)

PR #10937, opened 2025-08-25
Filesystem APIs: search/replace, reading file regions

vito:
[Schema diff](https://github.com/dagger/dagger/pull/10937/files#diff-6d45938d3a99ad730e8d639cd2218f57c90828f1deeaad17c073cad22bfd0b50)

replaces #10747

* Add `Directory.search` and `File.search`, powered by `ripgrep`
* Add `File.withReplaced` for applying basic edits to a file
* Add `File.contents(offsetLines, limitLines)` for reading line regions

delta from #10747:

* Now includes `SearchSubmatches` (requested by @jedevc)
* Test coverage for `File.withReplaced`


core/search.go
```diff
@@ -0,0 +1,224 @@
...
+		dagql.Arg("literal").Doc(`Interpret the pattern as a literal string instead of a regular expression.`),
+		dagql.Arg("multiline").Doc(`Enable searching across multiple lines.`),
+		dagql.Arg("dotall").Doc(`Allow the . pattern to match newlines in multiline mode.`),
+		dagql.Arg("ignoreCase").Doc(`Enable case-insensitive matching.`),
+		dagql.Arg("filesOnly").Doc(`Only return matching files, not lines and content`),
+		dagql.Arg("limit").Doc(`Limit the number of results to return`),
+	}
+}
+
+func (opts SearchOpts) RipgrepArgs() []string {
```
sipsma, line 75: IMO we should add ` --no-ignore` so it doesn't honor `.gitignore` by default and `--hidden` so it doesn't skip things that start with `.`. I can definitely see those being useful, but they feel better as options in our API (probably defaulting off).

core/directory.go
```diff
@@ -523,6 +523,52 @@ func (dir *Directory) WithPatch(ctx context.Context, patch string) (*Directory,
...
+		if err != nil {
+			return err
+		}
+		rgArgs := opts.RipgrepArgs()
+		for _, glob := range globs {
+			rgArgs = append(rgArgs, "--glob="+glob)
+		}
+		if len(paths) > 0 {
+			rgArgs = append(rgArgs, "--")
+			rgArgs = append(rgArgs, paths...)
```
sipsma: Paths could be set to absolute paths, right? In which case we'd escape out of the `resolvedDir`?

Also, while I saw in the docs that `rg` won't follow symlinks when recursing, it didn't explicitly say what it does if a path itself is a symlink. We definitely don't want to follow those in that case either

core/schema/file.go
```diff
@@ -29,7 +29,11 @@ func (s *fileSchema) Install(srv *dagql.Server) {
 		Syncer[*core.File]().
 			Doc(`Force evaluation in the engine.`),
 		dagql.Func("contents", s.contents).
-			Doc(`Retrieves the contents of the file.`),
+			Doc(`Retrieves the contents of the file.`).
+			Args(
+				dagql.Arg("offset").Doc(`Start reading after this line`),
+				dagql.Arg("limit").Doc(`Maximum number of lines to read`),
```
sipsma: I could easily see us adding a limit in terms of bytes in the future, maybe we should name this existing arg `lineLimit` (or `maxLines`, etc.) just to be future proof?

Same idea for `offset`.

core/file.go
```diff
@@ -156,57 +164,296 @@ func (file *File) Evaluate(ctx context.Context) (*buildkit.Result, error) {
...
+		return nil, err
+	}
+
+	mountable, err := ref.Mount(ctx, true, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get mountable reference: %w", err)
+	}
+
+	lm := snapshot.LocalMounter(mountable)
+	defer lm.Unmount()
```
sipsma: We do this unmount unconditionally but also call Unmount in fileReader.Close, which feels off

core/file.go
```diff
@@ -156,57 +164,296 @@ func (file *File) Evaluate(ctx context.Context) (*buildkit.Result, error) {
...
+	}
+
+	results := []*SearchResult{}
+	err = MountRef(ctx, ref, bkSessionGroup, func(root string) error {
+		resolvedDir, err := containerdfs.RootPath(root, filepath.Dir(file.File))
+		if err != nil {
+			return err
+		}
+		rgArgs := opts.RipgrepArgs()
+		rgArgs = append(rgArgs, filepath.Base(file.File))
```
sipsma: We probably want a `--` before the path here, right?

Also, a check that file.File isn't absolute is probably worth it; there's been some confusion in the past around `Directory.Dir` and `File.File` being absolute vs. relative, using an abs path here accidentally would be not-good.

core/file.go
```diff
@@ -156,57 +164,296 @@ func (file *File) Evaluate(ctx context.Context) (*buildkit.Result, error) {
...
+				matchesFrom = append(matchesFrom, match)
+			}
+		}
+		matches = matchesFrom
+	}
+
+	// Check for matches
+	if len(matches) == 0 {
+		return nil, fmt.Errorf("search string not found")
+	}
```
sipsma: nit: this being an unconditional error case is probably not what I'd personally want usually, I think I'd prefer a no-op. Could be an option to our API I suppose. Grey area though, not a blocker

core/search.go
vito: Done! Added `skipIgnored` and `skipHidden` and renamed `ignoreCase` to `insensitive` so it's not confused with ignore-files

core/file.go
vito: Woops yea that's janky. I guess it's been unmounting immediately, which works anyway since the file is already opened.

Just noticed this is completely redundant with `File.Open` though. That one only supports `LLB` but #10907 will make it work for `Result` too, so I'll just remove this.

core/file.go
vito: Worth a bikeshed!

The use case driving this at the moment is LLMs making precise edits to files, which most coding agents do via find + replace with a unique substring. If the substring is not unique, it adds more context until it is, hence the defensiveness there.

* `old: "})\n\nvar a = 1"`
* `new: "})\n\nvar a = 1\n\nvar b = 2"`

Sometimes the LLM gets the `old` part wrong. If that's a no-op the LLM won't notice and recover.

What's the use case for a no-op? :thinking: Maybe `sed`-like batch editing a bunch of `File`s without knowing what's in them first?

I guess you'd be using `all` in that case - maybe `all` could allow 0? :thinking:

vito: Made it so `all: true` does a no-op when there are no matches, since I think that makes sense no matter what, but I can be swayed into a separate option as well.

core/directory.go
vito: > Paths could be set to absolute paths, right? In which case we'd escape out of the `resolvedDir`?

Yep - adding safeguards now.

> Also, while I saw in the docs that `rg` won't follow symlinks when recursing, it didn't explicitly say what it does if a path itself is a symlink. We definitely don't want to follow those in that case either

Good call-out - when you pass a symlink as an arg, it happily follows it:

```
dagger on  filesystem-apis [$!?] via 🐹 v1.25.0 took 17s
❯ ln -sf ~/junk/

dagger on  filesystem-apis [$!?] via 🐹 v1.25.0
❯ rg -l 'Welcome to'
sdk/elixir/test/dagger/client_test.exs
docs/current_docs/getting-started/quickstarts/basics/index.mdx
docs/current_docs/getting-started/api/clients-cli.mdx

dagger on  filesystem-apis [$!?] via 🐹 v1.25.0
❯ rg --no-follow -l 'Welcome to' junk
junk/bind-rename-with-replace.txt
junk/and-and/cmd/dang/main.go
junk/doc-strings/cmd/dang/main.go
junk/dang-quot3/cmd/dang/main.go
junk/dang-with-and/cmd/dang/main.go
junk/half-baked-bbt-v2/bbt-v2-guide.md
junk/and-ops/cmd/dang/main.go
junk/half-baked-bbt-v2/docs/current_docs/quickstart/basics/index.mdx
junk/half-baked-bbt-v2/docs/current_docs/api/clients-cli.mdx
junk/half-baked-bbt-v2/sdk/elixir/test/dagger/client_test.exs
```

core/file.go
vito: It's already wrapped in a `filepath.Base` (with `filepath.Dir` passed to `containerdfs.RootPath` above), so I _think_ it should be fine when it comes to absolute vs. relative, but I've added a `--` to handle any funky filenames.

core/schema/file.go
vito: Done! (`offsetLines`, `limitLines`)

core/directory.go
vito: Ok this was kind of a pain in the butt to fix, but here it is: https://github.com/dagger/dagger/pull/10937/commits/dd661fb163c7403299603d17f62b04ef67190bed - not sure if there's a cleaner approach. :thinking:

cwlbraa:
gimme :)

core/file.go
sipsma: SGTM!

core/directory.go
```diff
@@ -523,6 +523,79 @@ func (dir *Directory) WithPatch(ctx context.Context, patch string) (*Directory,
...
+			for _, p := range paths {
+				resolved, err := containerdfs.RootPath(root, p)
+				if err != nil {
+					return err
+				}
+				// make it relative, now that it's safe, just for less obtuse errors
+				resolved, err = filepath.Rel(root, resolved)
+				if err != nil {
+					return err
+				}
```
sipsma:
```suggestion
				resolved, err := containerdfs.RootPath(resolvedDir, p)
				if err != nil {
					return err
				}
				// make it relative, now that it's safe, just for less obtuse errors
				resolved, err = filepath.Rel(resolvedDir, resolved)
				if err != nil {
					return err
				}
```
I think? Just so that we make path relative to `dir.Dir` and not the snapshot's root, since `dir.Dir` is what the caller sees as dir's "root"

vito: image

Yeah probably - let's see if I can add a test for that...

vito: My brain hurts now, but managed to test it + apply the fix - great catch!

Issue #8235, opened 2024-08-26
Dagger for codegen


shykes:
## Problem

In theory Dagger is a perfect fit for code generation (and docs generation) pipelines. But in practice the experience is a little rough:

1. Caller must explicitly export the generated directory back to the client's filesystem, which is cumbersome. For example `dagger call codegen -o .`
2. The target path is relative to the working directory, which means the command must be changed to match the current workdir - this makes the use of Dagger brittle. For example: `cd foo; dagger call codegen -o ..; cd bar; dagger call codegen -o ../..`
3. The export functionality is a blunt instrument: it doesn't support syncing, so for example it can't remove out of date previously generated files.

## Solution

Implement first-class support for codegen pipelines, in a way that doesn't break the fundamentals of the Dagger API.

## Design

TBD. Please make suggestions below to start the conversation!

shykes:
FYI @aluzzardi @sipsma @vito as discussed previously

shykes:
Relevant to #9079

shykes:
@marcosnils 👋 I consolidated #9079 into this one, which predates it.

sagikazarmark:
I don't know if it changes the scope, but there is another similar use case: code formatting.

The primary difference is that generated code is usually not modified by hand (or other tools, like code formatters) whereas handwritten code may be.

Another potential difference is generated code is often in a single or a few places within a codebase where nothing else lives which could simplify exporting content (eg. overwrite everything in a folder). That's not the case with "other" files that formatters may touch.

Personally, I've been _mostly_ satisfied with how Dagger currently works for code generation use cases (I usually just overwrite an entire generated folder). But I haven't really used Dagger for code formatting (yet).

sipsma:
The hardest problem here lies around caching; specifically the in magically tracking the source of a directory to support a default export target. These directories synced from the host are cached based on their content, completely disconnected from the paths/clients they were sourced from. This is important to preserve to keep our cache reusability maximal.

If we made the synced in Directory include state on where it was sourced from we'd either:
1. Have to include that path in the cache key, which breaks the cache reusability mentioned above
2. Not include that path in the cache key, which creates extremely confusing situations where you end up getting cached return values that don't have the same "client source directory" embedded in them. This would just be inconsistent+broken, not a viable option
3. Something very convoluted where we unset the "client source directory" before passing as an arg to a function, then attempt to find any references to it in the return value and reset it
[...]

PR #10946, opened 2025-08-25
add Changeset APIs + CLI exporting

vito:
Introduces `Changeset`, a new API representing a diff between two directories - similar to `Directory.diff`, except it can represent _removals_ as well.

[Schema diff](https://github.com/dagger/dagger/pull/10946/files#diff-6d45938d3a99ad730e8d639cd2218f57c90828f1deeaad17c073cad22bfd0b50)

## Better export mechanics

A goal of this PR is to address #8235. The plan is:

* [x] When the CLI calls a function that returns `Changes`, it prompts to automatically apply them to your working directory, after showing you a diff.
* Following #10907 - when `Changes` are returned, they are automatically applied to the `Env` hostfs.

Pros:

* It's very clear now when an API returns a full `Directory` or a delta to be applied.
* We don't need to wipe the entire destination anymore, now that we know exactly what paths were removed.
* A single API returning `Changeset` can be used both as a CLI entrypoint and as part of the internal build pipeline via `Directory.withChanges`.

## Future possibilities

An API like this would be neat, and dovetail nicely into the new `Directory.withPatch` API:

```graphql
extend type Changeset {
  "Returns a Git-compatible patch file."
  asPatch: File!
}
```

Building on this, we could potentially have APIs for `git` committing changes etc., possibly using the patch format as a generic medium.

### DIFF
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -712,6 +712,7 @@ func handleChangesetResponse(ctx context.Context, dag *dagger.Client, response a
 	changeset := dag.LoadChangesetFromID(dagger.ChangesetID(changesetID))
 
 	var summary strings.Builder
+	var noChanges bool
 	if err := (func() (rerr error) {
 		ctx, span := Tracer().Start(ctx, "analyzing changes")
 		defer telemetry.End(span, func() error { return rerr })
@@ -721,9 +722,19 @@ func handleChangesetResponse(ctx context.Context, dag *dagger.Client, response a
 			return fmt.Errorf("get patch contents: %w", err)
 		}
 
+		noChanges = patch == ""
+		if noChanges {
+			slog.Info("no changes to apply")
+			return nil
+		}
+
 		return idtui.SummarizePatch(idtui.NewOutput(&summary), patch, -1)
 	})(); err != nil {
 		return err
+	}
+
+	if noChanges {
+		return nil
 	}
 
 	exportDest := outputPath
--- a/core/integration/changeset_test.go
+++ b/core/integration/changeset_test.go
@@ -489,6 +489,10 @@ func (t *Test) Update() *dagger.Changeset {
 		WithNewFile("foo.txt", "foo\nbaz").
 		Changes(t.Dir)
 }
+
+func (t *Test) NoChanges() *dagger.Changeset {
+	return t.Dir.Changes(t.Dir)
+}
 `,
 		)
 
@@ -513,6 +517,10 @@ func (t *Test) Update() *dagger.Changeset {
 	contents, err = modGen.File("./outdir/baz.txt").Contents(ctx)
 	require.NoError(t, err)
 	require.Equal(t, "im new here", contents)
+
+	out, err := modGen.With(daggerCall("no-changes", "-o", "./outdir")).Stderr(ctx)
+	require.NoError(t, err)
+	require.Contains(t, out, "no changes to apply")
 }
 
 func (s ChangesetSuite) TestWithChanges(ctx context.Context, t *testctx.T) {


core/schema/directory.go
```diff
@@ -138,6 +138,19 @@ func (s *directorySchema) Install(srv *dagql.Server) {
 			Args(
 				dagql.Arg("other").Doc(`The directory to compare against`),
 			),
+		dagql.NodeFunc("changes", s.changes).
+			Doc(
+				`Return a virtual comparison between this directory and an older snapshot that can be applied to another filesystem.`,
+				`Returns an error if the other directory is not an ancestor of this directory.`,
```
vito: TODO: remove, this is no longer the case, it actually behaves reasonably (`removedPaths` includes all the irrelevant paths)

shykes:
Nit: I would call it `Change`, singular. There's a pretty common convention that a type name that is plural designates an array or some sort of collection. Better to avoid that confusion IMO.

shykes:
> ```graphql
> extend type Directory {
>   """
>   Return a virtual comparison between this directory and an older snapshot that can be applied to another filesystem.
>   """
>   changes(
>     """The older directory snapshot to compare against"""
>     older: DirectoryID!
>   ): Changes!

This deprecates `diff()`, right? I would either A) straight up break `diff` (assuming we can use compat mode + make migration easy), or B) call it something obviously similar, like `difference` or `compare`. I think `changes()` is too vague IMO - also not really symmetrical to `withChanges()`, so the `X() / withX()` pair doesn't really work IMO.

>   """Return a directory with changes from another directory applied to it."""
>   withChanges(
>     """Changes to apply to the directory"""
>     changes: ChangesID!
>   ): Directory!
> }

Per my previous comment on Change singular, I'd do:

```graphql
  withChange(change: ChangeID)
```

vito:
@shykes 

> Nit: I would call it `Change`, singular. There's a pretty common convention that a type name that is plural designates an array or some sort of collection. Better to avoid that confusion IMO.

Hmm, yeah that's true at least in some languages (like Go), and we kinda have to avoid friction across all of them.

`Change` feels weird to me though. To me it _does_ feel like it should be plural, considering it represents a set of changed files, not just one change. Depends on where you draw the lines in your head - is a Git commit one change, or many changes?

* Ironically we also have `File.contents`, which I always thought should be `File.content` - maybe that's why I went for `Changes` :smile:  

Maybe we can leapfrog this whole issue by calling it `Diff` instead, or maybe even `DirectoryDiff` to be more conservative.

Or `Patch`?

---

> This deprecates `diff()`, right? I would either A) straight up break `diff` (assuming we can use compat mode + make migration easy), or B) call it something obviously similar, like `difference` or `compare`. I think `changes()` is too vague IMO

The `diff` migration is a little fraught, since I also changed the order of the arguments, because I got it wrong every single time I used it. I always wanted to keep the chain going, like `foo.bar.baz.diff(foo)`, but actually you have to pre-medidate the diff, like `foo.diff(foo.bar.baz)`.

I went with "changes" because it sounded a little more directional -- you're asking the directory that _has_ the changes, what they are relative to some other directory. Not married to it though. Words are hard.

`changesFrom`? `diffFrom`?

>  - also not really symmetrical to `withChanges()`, so the `X() / withX()` pair doesn't really work IMO.

FWIW the symmetry seems the same to me:

```
c = dir.changes(olderDir)
dir.withChanges(c) will always == dir

d = dir.directory("foo")
dir.withDirectory("foo", d) will always == dir
```

shykes:
How about `ChangeSet`? 😁

core/dagop.go
```diff
@@ -128,7 +128,8 @@ func (op FSDagOp) Exec(ctx context.Context, g bksession.Group, inputs []solver.R
 	if !ok {
 		return nil, fmt.Errorf("server root was %T", opt.Server.Root())
 	}
-	obj, err := opt.Server.LoadType(ContextWithQuery(ctx, query), op.ID)
+	ctx = ContextWithQuery(ctx, query)
+	obj, err := opt.Server.LoadType(ctx, op.ID)
```
jedevc: Wild that we never noticed that before :thinking:

core/directory.go
```diff
@@ -930,6 +924,189 @@ func (dir *Directory) Diff(ctx context.Context, other *Directory) (*Directory, e
...
+	// Compute all the changes once
+	if err := changes.computeChanges(ctx); err != nil {
+		return nil, err
+	}
+
+	return changes, nil
+}
+
+// computeChanges calculates added, changed, and removed files/paths
+func (ch *Changes) computeChanges(ctx context.Context) error {
```
jedevc: suggestion: split related changes into `core/changes.go`?

core/directory.go
```diff
@@ -930,6 +924,189 @@ func (dir *Directory) Diff(ctx context.Context, other *Directory) (*Directory, e
...
+		}
+		ch.RemovedPaths = append(ch.RemovedPaths, fp)
+	}
+
+	return nil
+}
+
+type Changes struct {
+	Before dagql.ObjectResult[*Directory] `field:"true" doc:"The older/lower snapshot to compare against."`
+	After  dagql.ObjectResult[*Directory] `field:"true" doc:"The newer/upper snapshot."`
```
jedevc: WDYT about also tracking a `Diff dagql.ObjectResult[*Directory]` field? The user can call it themselves, but might be nice as a shorthand (we also compute it ourselves a few times).

core/directory.go
```diff
@@ -930,6 +924,189 @@ func (dir *Directory) Diff(ctx context.Context, other *Directory) (*Directory, e
...
+		Args:  []dagql.NamedInput{{Name: "pattern", Value: dagql.String("**")}},
+	}); err != nil {
+		return fmt.Errorf("failed to get paths from after directory: %w", err)
+	}
+	// Get diff paths (changed + added files)
+	if err := srv.Select(ctx, ch.Before, &diffPaths, dagql.Selector{
+		Field: "diff",
+		Args: []dagql.NamedInput{
+			{Name: "other", Value: dagql.NewID[*Directory](ch.After.ID())},
+		},
```
jedevc: I have a funny feeling we might run into this: https://github.com/dagger/dagger/blob/634e07d70eaca9e8545e4c92517bb5c7dddf6605/core/directory.go#L910-L913

Might be worth removing that limitation? I think the suggested shenanigans are reasonable!

dagql/objects.go
```diff
@@ -714,6 +714,9 @@ func (r ObjectResult[T]) call(
 	// values that have a pure content-based cache key different from the call-chain ID digest.
 	if idable, ok := val.(IDable); ok && idable != nil && !doNotCache {
 		valID := idable.ID()
+		if valID == nil {
+			return nil, fmt.Errorf("impossible: nil ID returned for value: %+v (%T)", val, val)
```
jedevc, line 718: "impossible", I feel like I have definitely hit this before, my sympathies :pray:

docs/docs-graphql/schema.graphqls
```diff
@@ -119,6 +122,39 @@ The `CacheVolumeID` scalar type represents an identifier for an object of type C
 """
 scalar CacheVolumeID
 
+"""
+A comparison between two directories representing changes that can be applied.
+"""
+type Changes {
```
jedevc: WDYT about `Changes.export` to match `Directory.export`? Not sure if you've mentioned how we'd do the export.

docs/docs-graphql/schema.graphqls
```diff
@@ -119,6 +122,39 @@ The `CacheVolumeID` scalar type represents an identifier for an object of type C
...
+  """The newer/upper snapshot."""
+  after: Directory!
+
+  """The older/lower snapshot to compare against."""
+  before: Directory!
+
+  """
+  Files and directories that existed before and were updated in the newer directory.
+  """
+  changedPaths: [String!]!
```
jedevc: Maybe `modifiedPaths` to avoid stuttering like `Changes.changedPaths`?

jedevc:
Continuing bikeshedding...

I'm not a huge `Changes` fan either, `ChangeSet` feels a bit *strange* to me though coming from a git-world, where it's usually a *series* of changes (i.e. multiple commits) - this is just one `Change` (which if we're going for `Change*` is probably my preference).

`Diff` would be nice, and probably my favorite! That also feels nice going into #7202, you could just do `GitRef.newCommit(diff: Diff, ...)` or whatever. Also, `Diff.asPatch` feels cool!

Frustrating that `Directory.diff` then returns a `Directory`. I'd happily see that functionality updated, so that it returns a `Diff` - we'd need version compat for it, and probably would still need to keep an internal `__diff` or something so that we can keep the functionality for the `Select` call.

shykes:
> `ChangeSet` feels a bit _strange_ to me though coming from a git-world, where it's usually a _series_ of changes (i.e. multiple commits) - this is just one `Change` (which if we're going for `Change*` is probably my preference).

In git terminology "changeset" usually refers to the set of changes in a single commit.

> `Diff` would be nice, and probably my favorite! That also feels nice going into #7202, you could just do `GitRef.newCommit(diff: Diff, ...)` or whatever. Also, `Diff.asPatch` feels cool!

Yes `Diff` also works, but it's weird that `Directory.diff()` doesn't return a `Diff`...

> Frustrating that `Directory.diff` then returns a `Directory`.

Haha exactly :)

> I'd happily see that functionality updated, so that it returns a `Diff` - we'd need version compat for it, and probably would still need to keep an internal `__diff` or something so that we can keep the functionality for the `Select` call.

Yeah breaking `Directory.diff()` seems like a can of worms, per @vito.

vito:
I'm good with `ChangeSet`, with a slight preference for `Changeset` (per [Wikipedia](https://en.wikipedia.org/wiki/Changeset)) to make it clear that it's a discrete type and not just a `Set` of `Change`s (which doesn't have its own type).

Tangent: to keep pulling on this Git/VCS analogy: what if you could have an agent look at PR reviews, start resolving all feedback in parallel, and then attempt to merge all of those `Changeset`s together? That would mean we actually have to do file content-wise merge, not just "last file layer wins" which is what you'd get without any special effort.

And then how do we handle conflicts? Maybe you could give that to another agent to resolve, or bubble it up to the user? (The latter is probably safer, last time I tried to tell an agent to resolve merge conflicts it just blindly picked one side, but maybe one day... or maybe with enough context. It would be neat if the changes were paired with an LLM-authored description as a hint for resolving.) In that case we'd probably want a `conflictedPaths` getter.

That can all be future stuff though if we want it. Just thinking out loud.

shykes:
> I'm good with `ChangeSet`, with a slight preference for `Changeset` (per [Wikipedia](https://en.wikipedia.org/wiki/Changeset)) to make it clear that it's a discrete type and not just a `Set` of `Change`s (which doesn't have its own type).

Ha ha works for me :)

> Tangent: to keep pulling on this Git/VCS analogy: what if you could have an agent look at PR reviews, start resolving all feedback in parallel, and then attempt to merge all of those `Changeset`s together? That would mean we actually have to do file content-wise merge, not just "last file layer wins" which is what you'd get without any special effort.

> And then how do we handle conflicts? Maybe you could give that to another agent to resolve, or bubble it up to the user? (The latter is probably safer, last time I tried to tell an agent to resolve merge conflicts it just blindly picked one side, but maybe one day... or maybe with enough context. It would be neat if the changes were paired with an LLM-authored description as a hint for resolving.) In that case we'd probably want a `conflictedPaths` getter.
> 
> That can all be future stuff though if we want it. Just thinking out loud.

I didn't completely understand all this but it sounds very cool :)

core/directory.go
vito: Yeah sounds good to me - especially if we plan to deprecate `diff` we'll need a way to actually, you know, get the diff. :smile: 

I'm not sure what to call it though. It's not really `diff` because it's a `Directory` and the problem this new type is solving is that a `Directory` can't represent removals.

Maybe `writes`? (Or is a removal still considered a "write"...)

Request For Dictionary @shykes 

For additional context, the current strategy (in absense of `Changeset.export)`) is that you `Directory.export` the writes and then remove all the `removedPaths`.

docs/docs-graphql/schema.graphqls
vito: So, I want that, but didn't see a quick path to implementing it, since we need to also remove paths. Open to hints there and then I can try to put it in this PR - are removals expressable over the filesync protocol?

For now I was planning to just do the removals manually client-side.

docs/docs-graphql/schema.graphqls
shykes: ~~Just call it `paths`, it's self-evident from context. `Changeset.paths`: the paths affected by this changeset.~~

Nevermind, I hadn't seen the other xxPaths fields, forget I said anything

core/directory.go
shykes: I would call it `layer` or `asLayer`.

docs/docs-graphql/schema.graphqls
shykes:
```suggestion
type Delta {
```

docs/docs-graphql/schema.graphqls
```diff
@@ -1387,6 +1423,16 @@ type Directory {
...
 
+  """
+  Return a virtual comparison between this directory and an older snapshot that can be applied to another filesystem.
+
+  Returns an error if the other directory is not an ancestor of this directory.
+  """
+  changes(
+    """The older directory snapshot to compare against"""
+    older: DirectoryID!
+  ): Changes!
```
shykes:
```suggestion
  changes(
    """The base directory snapshot to compare against"""
    from: DirectoryID!
  ): Delta!
```

shykes (approved):
Small API nits, otherwise looks great!

I found a type name I like better than `Changeset`: `Delta` :)

docs/docs-graphql/schema.graphqls
vito: (Chatted in Discord but leaving here for continuity:) I think I like `Changeset` more - it'll lend itself well to having a description in the future, and being used as a Git commit, maybe even having that description interpreted by an LLM to figure out how to resolve a merge conflict while being faithful to both changes. `Delta` feels kinda like we painted ourselves into a naming corner and reached for a thesaurus to escape :smile:

vito:
Addressed the outstanding TODO for having the CLI apply changes when a function returns a `Changeset`:

[asciicast](https://asciinema.org/a/737614)

Which involved adding some more APIs:

* `Changeset.layer`: as requested
* `Changeset.asPatch: File!`: returns the changes in a Git-compatible patch format, used for displaying the change summary
  * shares test coverage between `Directory.withChanges` and `Directory.withPatchFile(Changeset.asPatch)`
* `Changeset.sync`: in dogfooding I noticed a few places that call `.Sync` to pre-compute things in goroutines, so I just added it
  * ideally this sort of thing would become unnecessary by evaluating chained API call args in parallel in `dagql`
* `Directory.withPatchFile`: a `File`-accepting form of `withPatch`
* `Directory.withPatch` and `withPatchFile` now no-op with an empty patch, rather than failing

Along the way I switched to `charmbracelet/huh` for our TUI prompts which should be much nicer going forward as it's about a billion times more flexible and powerful than our current prompt implementation.

@shykes @jedevc Sorry for expanding scope post-approval, but that last TODO felt important; if these changes are too egregious I can split them out to another PR, but my understanding is nothing is blocked on `Changeset` so might as well get it as good as possible before merging. If we're not good-to-go prior to vacation I'm happy to hand the PR off for final touches etc.

core/schema/wrapper.go
```diff
@@ -159,6 +159,9 @@ func getSelfDigest(a any) (digest.Digest, error) {
 	case *core.GitRef:
 		// FIXME can core.DigestOf(x) be used instead? When set, the TestGit/TestAuthClient test failed intermittently
 		return "", nil // fallback to using dagop ID
+	case *core.Changeset:
+		// FIXME this seems weird?
+		return "", nil
```
jedevc, line 175: Indeed :sweat_smile: 

Some discussion in private discord here: https://discord.com/channels/707636530424053791/1347633524093616190/1416099558315786451

I wonder how much of this might be removable at some point :thinking:

cmd/dagger/functions.go
```diff
@@ -694,6 +702,207 @@ func handleResponse(returnType *modTypeDef, response any, o, e io.Writer) error
...
+	slog := slog.SpanLogger(ctx, InstrumentationLibrary)
+
+	if _, err := changeset.Layer().Export(ctx, exportDest); err != nil {
+		return err
+	}
+
+	removed, err := changeset.RemovedPaths(ctx)
+	if err != nil {
+		return err
+	}
```
jedevc, line 774: Question? If I added `Host.removeFiles` or similar, would this be able to be moved to the engine?

Dunno if it's a huuuge blocker, but I can't see if it's possible to export a changeset from `dagger shell`. I *think* `--output` is only for `dagger call`/`core`.

cmd/dagger/functions.go
```diff
@@ -694,6 +702,207 @@ func handleResponse(returnType *modTypeDef, response any, o, e io.Writer) error
...
+		slog.Info("removing file", "path", filePath)
+		if err := os.Remove(filePath); err != nil {
+			slog.Error("failed to remove file; ignoring", "path", filePath, "error", err)
+		}
+	}
+
+	return nil
+}
+
+func summarizePatch(out *termenv.Output, patch string, maxWidth int) error {
```
jedevc: Move to `idtui`?

core/integration/directory_test.go
```diff
@@ -755,6 +755,863 @@ func (DirectorySuite) TestDiff(ctx context.Context, t *testctx.T) {
 	*/
 }
 
+func (DirectorySuite) TestChanges(ctx context.Context, t *testctx.T) {
```
jedevc: nit: maybe a separate file for these tests?

core/integration/directory_test.go
jedevc, line 1: Can't see tests for the output flag.

cmd/dagger/functions.go
```diff
@@ -694,6 +702,207 @@ func handleResponse(returnType *modTypeDef, response any, o, e io.Writer) error
...
+	})(); err != nil {
+		return err
+	}
+
+	exportDest := outputPath
+	if exportDest == "" {
+		exportDest = "."
+
+		var confirm bool
+		form := idtui.NewForm(
```
jedevc, line 745: Should we still prompt even if there are no changes?

image

This feels a bit weird, since it won't do anything if I hit `y`.

jedevc:
Neat! I like it, code looks good!

`huh` is a very neat little library :grin:

cmd/dagger/functions.go
vito: Yeah that seems like it would unblock a `Changeset.export` indeed :+1: - maybe a quick followup PR? (Previously: https://github.com/dagger/dagger/pull/10946#discussion_r2305227165)

cmd/dagger/functions.go
vito: Good catch - added a short-circuit, will try it out with some tests tomorrow

edit: done

core/integration/directory_test.go
vito: Added!

jedevc (approved):
Approving, LGTM :tada: :tada:

Ping @shykes on the API, but IMO, they're good!

shykes:
@jedevc yup looks good, we already did some bikeshedding on the name, time to ship :)

(thanks for checking)

shykes:
🥳

PR #10949, opened 2025-08-26
EnvFile: native support for standard .env files

shykes:
Native support for env files in the `.env` format.

This is a building block towards native support of `.env` files for configuring a module (see #10697).

[Schema diff](https://github.com/dagger/dagger/pull/10949/files#diff-6d45938d3a99ad730e8d639cd2218f57c90828f1deeaad17c073cad22bfd0b50)


alexcb (commented):
Looks neat!

I think it could use some integration tests before merging it.

core/envvars.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// Clone creates a deep copy of the EnvFile
+func (ef *EnvFile) Clone() *EnvFile {
+	newEnviron := make([]string, len(ef.Environ))
+	copy(newEnviron, ef.Environ)
+	return &EnvFile{
+		Environ: newEnviron,
+	}
+}
```
jedevc: nit: `slices.Clone`

core/schema/envfile.go
```diff
@@ -0,0 +1,90 @@
...
+		dagql.Func("variables", s.variables).
+			Doc("Return all variables"),
+		dagql.Func("withVariable", s.withVariable).
+			Doc("Add a variable").
+			Args(
+				dagql.Arg("name").
+					Doc("Variable name"),
+				dagql.Arg("value").
+					Doc("Variable value"),
+			),
```
jedevc: tiny nit: any reason to not put `Doc` on the same line as `Arg`?

core/schema/envfile.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/dagger/dagger/dagql"
+)
+
+type envfileSchema struct{}
+
+var _ SchemaResolvers = &envfileSchema{}
+
+func (s envfileSchema) Install(srv *dagql.Server) {
+	dagql.Fields[*core.Query]{
+		dagql.Func("envFile", s.newEnvFile).
```
jedevc, line 16: Don't want to have a super extended bikeshed, but what do you think about instead calling this `EnvVariables` / `EnvVariableSet` / etc.

IMO, I like the idea of having a function that can take a collection of various env vars, potentially being to bulk-inject them into a function (using a future `Container.withEnvVariables`.

That way, it's detached from the concept of an underlying file, while still being able to be parsed using `File.asEnvFile`, or converted to one using `EnvVariables.file` (or maybe even `EnvVariables.asFile`).

(Also the stutter of `EnvFile.file` feels kind bleh, another motivator for renaming the type/constructor)

jedevc (commented):
Agreed tests are needed, so that we don't break things!

core/envvars.go
shykes: 🫡

core/schema/envfile.go
shykes: I find it more readable but obviously it's a highly subjective opinion 😛 Happy to conform to whatever the convention is.

core/schema/envfile.go
shykes: I think `EnvVariables` has the same problem as `Changes` (ambiguous meaning of plural), and `EnvVariableSet` just feels too abstract and verbose. I agree `EnvFile.file` kind of stutters, but `EnvFile`is still the best option overall: it's familiar and shorter. Maybe the stutter goes away once `File` becomes an interface along with other core types? 😇

TomChv: Since we already have `File.AsEnvFile`, we could have `EnvFile.AsFile` so we keep consistency.
It's a bit repetitive too but I agree that `EnvVariables` may be ambiguous and `Env` is already taken for MCP thing.

`Env.AsFile` would have been perfect but we can't have that I guess.

core/schema/envfile.go
TomChv: Our convention is the following (extracted from `schema/directory.go`):

```go
		dagql.Func("file", s.file).
			Doc(`Retrieve a file at the given path.`).
			Args(
				dagql.Arg("path").Doc(`Location of the file to retrieve (e.g., "README.md").`),
			),
			
		dagql.NodeFunc("withFile", DagOpDirectoryWrapper(srv, s.withFile, WithPathFn(keepParentDir[WithFileArgs]))).
			Doc(`Retrieves this directory plus the contents of the given file copied to the given path.`).
			Args(
				dagql.Arg("path").Doc(`Location of the copied file (e.g., "/file.txt").`),
				dagql.Arg("source").Doc(`Identifier of the file to copy.`),
				dagql.Arg("permissions").Doc(`Permission given to the copied file (e.g., 0600).`),
			),
```

So as a general notation:

```
dagql.<Func|NodeFunc|...>("<field name>", <resolved>).
  Doc(`<doc string>`).
  Args(
    dagql.Arg("<arg name>").Doc(`<doc string>`),
    ...
  ),
```

core/schema/envfile.go
jedevc: @shykes fair enough.

One note though, sort of naming related, is it feels weird for a function to take a `EnvFile` (e.g. if a caller is just manually creating a group of envs, there's no backing file) - would we instead suggest a user write the function to take a `[]EnvVariable`?

And if so, what if the user passes an `EnvFile` on the CLI? Ideally that should automatically "cast" into an `[]EnvVariable`?

jedevc: I actually think the plural would be fine for this case - it is kind of a collection.

(it would actually be super cool if we could extend SDKs to support iterators over types like this one day, e.g. if they had a `asArray` function)

core/schema/envfile.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (s envfileSchema) variables(ctx context.Context, obj *core.EnvFile, args struct{}) (dagql.Array[core.EnvVariable], error) {
+	return obj.Variables(), nil
+}
+
+func (s envfileSchema) withVariable(ctx context.Context, obj *core.EnvFile, args struct {
+	Name  dagql.String
+	Value dagql.String
+}) (*core.EnvFile, error) {
+	return obj.WithVariable(args.Name.String(), args.Value.String(), false), nil
```
TomChv: Why expand is set to `false` and isn't a parameter? Is that on purpose?

TomChv: Also, `asEnvFile` automatically expand variable so there's an inconsistency there:

Screenshot 2025-09-04 at 12 43 52

TomChv: It's a bit tricky though, I'm not sure we want to `expand` the variable when it got registered, but instead only when it's read.

If you convert `EnvFile` to a file, we should keep the original syntax like `BAR=${FOO}` stay that way but if you do `EnvFile.EnvVariable("BAR")` or `EnvFile.EnvVariables` then we should interpolate.

Wdyt @shykes?

TomChv:
Hey @shykes 

I pushed a commit to add tests and improve the code consistency (copy of my commit messages):

```
refactor: cleanup code & add tests:
- Add EnvFileSuite test
- Cleanup schema doc indentation
- Cleanup schema function signature argument (`obj` -> `parent`)
- Improve `core/envvars.go` error messages
- Update `core/envvars.go` code style for consistency

NOTE: I commented some tests since there's a little point to resolve
around variable expansion (see: https://github.com/dagger/dagger/pull/10949#discussion_r2321595484)
```

The next things to do on this PR:
- Resolve `EnvFile.File` API naming: https://github.com/dagger/dagger/pull/10949#discussion_r2318420506
- Resolve `expand` issue: https://github.com/dagger/dagger/pull/10949#discussion_r2321595484

core/schema/envfile.go
TomChv: Fixed by [d07e4e9](https://github.com/dagger/dagger/pull/10949/commits/d07e4e9523dba2d5bf7af42da9c17466466a415f)

core/envvars.go
TomChv: Fixed by [d07e4e9](https://github.com/dagger/dagger/pull/10949/commits/d07e4e9523dba2d5bf7af42da9c17466466a415f)

core/schema/envfile.go
shykes: I refactored expansion, now it's configurable at construction.

core/envvars.go
```diff
@@ -0,0 +1,179 @@
...
+	if ef.Expand {
+		expanded, err := sh.Expand(value, ef.get)
+		if err != nil {
+			// Shell expansion errors -> return raw value
+			return value, true
+		}
+		return expanded, true
+	}
+	return value, true
+}
```
jedevc: Note: this doesn't handle recursive shell expansion - is that intentional.

e.g.

```
A=1
B=$A
C=$B
```

I'd expect `ef.Variable("C")` to return 1.

If we want to do this, we need some tracker to make sure we don't infinitely recurse.

core/schema/envfile.go
```diff
@@ -0,0 +1,111 @@
...
+			Doc("Remove all occurrences of the named variable").
+			Args(
+				dagql.Arg("name").Doc("Variable name to remove"),
+			),
+		dagql.Func("variable", s.variable).
+			Doc("Lookup a variable by name (last occurrence wins)").
+			Args(
+				dagql.Arg("name").Doc("Variable name to lookup"),
+			),
+		dagql.Func("file", s.file).
```
jedevc: WDYT about calling this `asFile` for consistency across the API?

core/envvars.go
```diff
@@ -0,0 +1,179 @@
...
+	return &cp
+}
+
+// WithoutVariable removes all occurrences of the named variable
+func (ef *EnvFile) WithoutVariable(name string) *EnvFile {
+	ef = ef.Clone()
+	ef.Environ = nil
+	for _, kv := range ef.Environ {
+		k, _, _ := strings.Cut(kv, "=")
+		if !shell.EqualEnvKeys(k, name) {
```
jedevc:
```suggestion
		if k != name {
```

We don't need the dockerfile shell package here, it's def is:

```go
// EqualEnvKeys compare two strings and returns true if they are equal.
// On Unix this comparison is case-sensitive.
// On Windows this comparison is case-insensitive.
func EqualEnvKeys(from, to string) bool {
	return from == to
}
```

This only runs on linux.

shykes: Ok, note that I copy-pasted it from code that is already in the engine - maybe we should delete it from there too?

core/schema/envfile.go
shykes: I like things shorter, but you're right it would be more consistent. I'll change it.

core/envvars.go
shykes: Not intentional, but known 😅

I wanted to unblock @chrisp asap in the default secrets POC, so punted on a more sophisticated implementation.

I can do it later today.

shykes: Fixed! It also supports double quotes, single quotes, and escaping (this is the norm for most standard .env parsers apparently)

shykes:
Almost there... I'm struggling with a test failure in expansion

core/integration/envfile_test.go
```diff
@@ -0,0 +1,220 @@
...
+DOG=${FOO}-${HELLO}
+`)
+
+	variables, err := file.AsEnvFile().Variables(ctx)
+	require.NoError(t, err)
+
+	compareEnv(ctx, t, map[string]string{
+		"FOO":   "bar",
+		"HELLO": "world",
+		//	"DOG":   "bar-world", // TODO: This is automatically expand here
```
TomChv: You can uncomment that I think now?

core/integration/envfile_test.go
```diff
@@ -0,0 +1,220 @@
...
+		WithVariable("HELLO", "world").
+		WithVariable("DOG", "${FOO}-${HELLO}")
+
+	variables, err := envFile.Variables(ctx)
+	require.NoError(t, err)
+
+	compareEnv(ctx, t, map[string]string{
+		"FOO":   "bar",
+		"HELLO": "world",
+		//	"DOG":   "bar-world", // TODO: this is not expanded
```
TomChv: You can uncomment that I think now?

core/schema/envfile.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (s envfileSchema) get(ctx context.Context, parent *core.EnvFile, args struct {
+	Name dagql.String
+}) (dagql.String, error) {
+	if val, ok := parent.Lookup(args.Name.String()); ok {
+		return dagql.NewString(val), nil
+	}
+	// We could return null instead of "", but clients don't handle it well:
+	//  - Go SDK ignores it (returns a string)
+	//  - Shell prints "<nil>"
```
jedevc: IMO, that shouldn't stop us from doing it? For example, that's what https://github.com/dagger/dagger/pull/10956 is doing, it feels weird to have pretty much the same use case in both, introduce them both at the same time, and have different behaviors.

shykes: Yeah that's fair. I just don't know what the status of nullable results is: do we plan on improving support for it, is it something stable and definitive, or something that might go away?

shykes:
Found the bug! Pushed a fix, and cleaned up the tests

core/schema/envfile.go
shykes: actually @jedevc that comment was out of date. I'm intentionally mimicking the `getenv` behavior, which always returns a string value.

I'm also reserving a space for `variable()` that will return a variable object, and that will be nullable.

PR #10950, opened 2025-08-27
docs: security feature to sandboxed runtime

jpadams:
image

### DIFF
--- a/docs/netlify.toml
+++ b/docs/netlify.toml
@@ -186,6 +186,11 @@
   status = 301
 
 [[redirects]]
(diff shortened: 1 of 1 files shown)


jasonmccallister (approved):
We should probably add a redirect for security to this new uri - but otherwise good to go

jpadams:
> We should probably add a redirect for security to this new uri - but otherwise good to go

added commit for that, tested, working! Thank you!


https://github.com/user-attachments/assets/a42b0ee3-75ae-4191-abec-e320323f89f3

PR #10955, opened 2025-08-27
DNM: see how json/v2 does

sipsma:
Curious to see how [jsonv2](https://tip.golang.org/doc/go1.25#json_v2) performs in our CI since there's purported decoding perf benefits and we do our fair share of json decoding.


sipsma:
eh, this is a PITA until we do a release w/ go 1.25, will revisit then

PR #10956, opened 2025-08-28
findup API: find a file or directory by walking up the directory tree

shykes:
[Schema diff](https://github.com/dagger/dagger/pull/10956/files#diff-6d45938d3a99ad730e8d639cd2218f57c90828f1deeaad17c073cad22bfd0b50)

This adds support for executing a "findup" search, like for example looking for the nearest `.git`, `dagger.json` etc.

It is available in `Directory` and `Host`

### DIFF
--- a/core/schema/host.go
+++ b/core/schema/host.go
@@ -162,6 +162,12 @@ func (s *hostSchema) Install(srv *dagql.Server) {
 				dagql.Arg("noCache").Doc(`If true, the file will always be reloaded from the host.`),
 			),
 
+		dagql.NodeFuncWithCacheKey("findUp", s.findUp, dagql.CacheAsRequested).
+			Doc(`Search for a file or directory by walking up the tree from system workdir. Return its relative path. If no match, return null`).
+			Args(
+				dagql.Arg("name").Doc(`name of the file or directory to search for`),
+			),
+
 		dagql.NodeFuncWithCacheKey("unixSocket", s.socket, s.socketCacheKey).
 			Doc(`Accesses a Unix socket on the host.`).
 			Args(
@@ -292,7 +298,7 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
 				return inst, fmt.Errorf("failed to get absolute path from git ignore root %s: %w", hostPath, err)
 			}
 		} else {
-			dotGitPath, found, err := findUp(ctx, core.NewCallerStatFS(bk), hostPath, ".git")
+			dotGitPath, found, err := host.Self().FindUp(ctx, core.NewCallerStatFS(bk), hostPath, ".git")
 			if err != nil {
 				return inst, fmt.Errorf("failed to find up .git: %w", err)
 			}
@@ -501,6 +507,39 @@ func (s *hostSchema) file(ctx context.Context, host dagql.ObjectResult[*core.Hos
 		return i, err
 	}
 	return i, nil
+}
+
+type hostFindUpArgs struct {
+	Name string
+	HostDirCacheConfig
+}
+
+func (s *hostSchema) findUp(ctx context.Context, host dagql.ObjectResult[*core.Host], args hostFindUpArgs) (i dagql.Nullable[dagql.String], err error) {
+	query, err := core.CurrentQuery(ctx)
+	if err != nil {
+		return i, err
+	}
+	bk, err := query.Buildkit(ctx)
+	if err != nil {
+		return i, fmt.Errorf("failed to get buildkit client: %w", err)
+	}
+	cwd, err := bk.AbsPath(ctx, ".")
+	if err != nil {
+		return i, fmt.Errorf("failed to get cwd: %w", err)
+	}
+	foundPath, found, err := host.Self().FindUp(ctx, core.NewCallerStatFS(bk), cwd, args.Name)
+	if err != nil {
+		return i, fmt.Errorf("failed to find %s: %w", args.Name, err)
+	}
+	if !found {
+		return dagql.Null[dagql.String](), nil
+	}
+	foundPath = path.Join(foundPath, args.Name)
+	relPath, err := filepath.Rel(cwd, foundPath)
+	if err != nil {
+		return i, fmt.Errorf("failed to make path relative to cwd: %w", err)
+	}
+	return dagql.NonNull(dagql.NewString(relPath)), nil
 }
 
 type hostTunnelArgs struct {
--- a/docs/docs-graphql/schema.graphqls
+++ b/docs/docs-graphql/schema.graphqls
@@ -1534,6 +1534,17 @@ type Directory {
     include: [String!] = []
   ): Directory!
 
+  """
+  Search up the directory tree for a file or directory, and return its path. If no match, return null
+  """
+  findUp(
+    """The name of the file or directory to search for"""
+    name: String!
+
+    """The path to start the search from"""
+    start: String!
+  ): String
+
   """Returns a list of files and directories that matche the given pattern."""
   glob(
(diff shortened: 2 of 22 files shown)


jedevc:
Can you share a use case?

shykes:
> Can you share a use case?

Loading `.env` files 😁 in combination with #10949

docs/docs-graphql/schema.graphqls
```diff
@@ -1516,6 +1516,35 @@ type Directory {
     include: [String!] = []
   ): Directory!
 
+  """
+  Search up the directory tree for a file or directory, and return its path
+  """
+  findup(
```
vito: Bikeshed: seems like the `U` should be capitalized, I've never seen `findup` as one word so don't think it's earned special capitalization treatment (by my arbitrary judgment)

Maybe this?:

* `findUp`
* `findDirectoryUp`
* `findFileUp`

Or could we even do without the `findup` API and just have the `Directory` / `File` versions? Doesn't seem very useful to know a path exists but then still not know whether it's a file or directory.

docs/docs-graphql/schema.graphqls
```diff
@@ -1516,6 +1516,35 @@ type Directory {
...
+  ): String!
+
+  """Search up the directory tree for a directory, and return it"""
+  findupDirectory(
+    """The name of the directory to search for"""
+    name: String!
+
+    """The path to start the search from"""
+    start: String!
+  ): Directory!
```
vito: nit: Should this be nullable, since we might not find it, and that feels like it's usually handled, rather than treated as an exceptional case? As in, for languages with exceptions, that feels more natural as control flow rather than having to `try ... catch`. (This is mostly vibes tbh.)

Also what happens if you do `findDirectoryUp` but then it finds on a `File`? Maybe that should be an error, rather than `null`, if we make it nullable?

Also, noting a specific weird case: `.git` is sometimes a directory and sometimes a file, thanks to submodules. 😵‍💫 Don't know what to suggest there, or if it's even something we're worried about for any planned usage.

core/schema/host.go
```diff
@@ -160,6 +160,24 @@ func (s *hostSchema) Install(srv *dagql.Server) {
...
+		dagql.NodeFuncWithCacheKey("findup", s.findup, dagql.CacheAsRequested).
+			Doc(`Search for a file or directory by walking up the tree from system workdir. Return its relative path`).
+			Args(
+				dagql.Arg("name").Doc(`name of the file or directory to search for`),
+			),
+
+		dagql.NodeFuncWithCacheKey("findupFile", s.findupFile, dagql.CacheAsRequested).
+			Doc(`Search for a file by walking up the tree from system workdir`).
+			Args(
+				dagql.Arg("name").Doc(`name of the file or directory to search for`),
```
vito, line 168:
```suggestion
				dagql.Arg("name").Doc(`name of the file to search for`),
```

docs/docs-graphql/schema.graphqls
jedevc: Could we instead have just `findUp`? And have it take an optional arg of `ExistsType`?

https://github.com/dagger/dagger/blob/547d08ee68bee3ea633ce0f23d3cd7e02d2834cf/core/directory.go#L1174-L1194

Then there's only one function (and also it extend naturally as we maybe add more file types to that, e.g. unix socket files)

docs/docs-graphql/schema.graphqls
jedevc: > Also, noting a specific weird case: `.git` is sometimes a directory and sometimes a file, thanks to submodules. 😵‍💫 Don't know what to suggest there, or if it's even something we're worried about for any planned usage.

Should hopefully be resolved by https://github.com/dagger/dagger/pull/10956#discussion_r2313560960

docs/docs-graphql/schema.graphqls
shykes: The DX is much nicer if I can just get the object I'm looking for (file or directory) and chain from there.

ie. I want this:

```golang
dotenv := host.FindupFile(".env").AsEnvFile()
```

And not this:

```golang
dotenvPath, err := host.Findup(ctx, ".env")
if err != nil {
  return err
}
dotenv := host.File(dotenvPath).AsEnvFile()
```

docs/docs-graphql/schema.graphqls
shykes: @vito I like nullable, but noticed clients don't handle it super well.. Or at least they don't handle nullable *strings* well (Go SDK seems to ignore it; dagger shell prints `<nil>`). But yeah I agree it would make more sense for this to be nullable.

docs/docs-graphql/schema.graphqls
shykes: OK you convinced me @jedevc , less is more... I removed `FindupDirectory()` and `FindupFile()` completely. Didn't add a `ExistType` argument, but can do that as a follow-up.

docs/docs-graphql/schema.graphqls
shykes: @vito update:

- I made everything nullable
- Then realized it didn't work (returning null object value from the engine shows up as non-null object in the Go SDK client)
- Gave up, removed `FindupDirectory()` and `FindupFile()` for this and other reasons, one less problem to solve

shykes:
@grouville @TomChv @jedevc @alexcb would love a second look 🙏

core/schema/host.go
```diff
@@ -161,6 +161,12 @@ func (s *hostSchema) Install(srv *dagql.Server) {
 				dagql.Arg("noCache").Doc(`If true, the file will always be reloaded from the host.`),
 			),
 
+		dagql.NodeFuncWithCacheKey("findUp", s.findUp, dagql.CacheAsRequested).
+			Doc(`Search for a file or directory by walking up the tree from system workdir. Return its relative path. If no match, return null`).
+			Args(
+				dagql.Arg("name").Doc(`name of the file or directory to search for`),
+			),
```
jedevc, line 169: Should this also take a `start` path? The caller should be able to use `.` to get the current dir, but just gets a more consistent API (and enables future use cases).

core/schema/host.go
```diff
@@ -502,6 +508,39 @@ func (s *hostSchema) file(ctx context.Context, host dagql.ObjectResult[*core.Hos
...
+	}
+	bk, err := query.Buildkit(ctx)
+	if err != nil {
+		return i, fmt.Errorf("failed to get buildkit client: %w", err)
+	}
+	cwd, err := bk.AbsPath(ctx, ".")
+	if err != nil {
+		return i, fmt.Errorf("failed to get cwd: %w", err)
+	}
+	foundPath, found, err := findUp(ctx, core.NewCallerStatFS(bk), cwd, args.Name)
```
TomChv: Since findUp will be part of the API, do we still need our internal [`moduleSource.findUp(s)`](https://github.com/dagger/dagger/blob/a86457436a24edf8d2f9de0e30bc3b5a91af28ee/core/schema/modulesource.go#L2544)?

Shouldn't it be inside `core.Host` in that case or directly in `schema/host.go`? It feels a bit strange that `schema/host.go` call  a method defined in `schema/modulesource.go`

core/schema/host.go
shykes: Good idea, do you mind if we make it a follow-up?

core/schema/host.go
shykes: I moved it to `core/host.go`.

I left it unchanged, to avoid unnecessary refactoring.

shykes:
> Please check my comment on `host.findUp` and add tests for `Directory.FindUp` in `core/integration` before merging.

@TomChv done and done.

Issue #10200, opened 2025-04-17
✨ Add remote fetching (tags, branches) to `Git` functionality


nipuna-perera:
### What are you trying to do?

I'm running `source.AsGit().WithAuthToken().Tags(ctx)` and realized it won't show the remote tags because it doesn't do a fetch. Would it be possible to add a `AsGit().Fetch(branch or all)`?

### Why is this important to you?

The `AsGit` is very useful but I can't use it because I need to fetch before doing much with the repo. The fetch itself requires a PAT so I have to set up all of that whereas `AsGit().WithAuthToken()` handles it.

### How are you currently working around this?

Setting up all the plumbing necessary inside the container like this #9858 

I also think some of this is related to the work @jedevc is doing in #10195, #7637, #9980 and #9968 

@shykes suggested `source.AsGit().Remote("origin").Tags()` as that preents side-effects.

jedevc:
Mm, very reasonable, I've also wanted this.

I like `source.AsGit().Remote("origin").Tags()`, which would then look up `"origin"` in the target repository, which would then be the remote "origin", and return it as another `GitRepository`.

A couple little notes:
- Where do the auths go? I'm kind of leaning towards putting them as args to `Remote` which mirrors how the args in the `git` top-level constructor work. I think being able to do `WithAuthToken` on a result from `AsGit` is a bit of an anti-pattern, since an auth token isn't used for a directory here (so I'd like to deprecated those).
- Remote naming becomes pretty important. e.g. `origin`/`upstream`/etc - users might name these differently - is that an issue?

shykes:
@jedevc would the host's auth be loaded implicitly by default? PAT etc?

jedevc:
Hm, just realized that it's really not possible to implement this for remote git providers - it's only valid for local git: `Directory.asGit.remote` is valid, `git.remote` isn't.

It would be *really* cool if we could somehow have it follow fork/upstream relationships, but there's no way to get this generically from the Git API - we'd have to implement provider-specific integrations (which I was hoping to avoid, since it's a pain).

jedevc:
Authentication aside, I ran into a super concrete thing we need in our own CI.

If I specify a high `--depth`, and the local directory for `dir.asGit` is a shallow clone, we should fallback to fetching any history from the remotes.

PR #10959, opened 2025-08-28
feat: add GitRepository.url api

jedevc:
There's currently no easy way to get this without checking out the whole repo using `tree`.

A few use cases: 
- Determining the fully-resolved ref (if an implied URL is provided, like `github.com/dagger/dagger`, without the leading `https://`, see #10960 for where this now becomes ambiguous)
- Getting the URL of a remote, as in the case of #10200

Note, this can *only* be implemented for remote repos, local repos have no externally accessible url, so we should return an error.

### DIFF
--- a/core/git.go
+++ b/core/git.go
@@ -40,6 +40,7 @@ import (
 )
 
 type GitRepository struct {
+	URL     dagql.Nullable[dagql.String] `field:"true" doc:"The URL of the git repository."`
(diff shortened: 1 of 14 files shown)


jedevc:
Ahh, yeah, good point :) I'll do that instead.

shykes:
Love it!

PR #10960, opened 2025-08-28
feat: determine git proto when not specified

jedevc:
Previously, we were always assuming that a git url without a protocol (like `github.com/dagger/dagger.io`) could always be interpreted as an HTTPS URL.

However, this is a bit weird, we should also consider that it could be interpreted as an SSH URL.

This *also* means that a private module URL such as `github.com/dagger/dagger.io` can now be accessed using SSH as well as HTTPS, depending on the available auth available at runtime!


core/schema/git.go
```diff
@@ -441,6 +477,10 @@ func (s *gitSchema) git(ctx context.Context, parent dagql.ObjectResult[*core.Que
 		inst = inst.ResultWithPostCall(postCall)
 	}
 
+	// TODO: it would be really nice if we could check backend.Accessible here
+	// (and take a snapshot of all the remote refs), but withAuthToken and
+	// withAuthHeader might be used to add auth later, so we *currently* can't :(
```
jedevc, line 485: I am wildly looking forward to removing `withAuthToken` (and kicking it out of backwards compat)

util/gitutil/error.go
jedevc, line 1: Note: this is *technically* a breaking change, we change the format of git errors to be more stongly typed, and to not include the stderr of the message directly (which was sometimes large, and made for repetitive log output).

However, I have decided that version compat is not going to stretch as far as the exact bytes of an error message, because otherwise we will never be able to change anything :stuck_out_tongue_closed_eyes: 

At some point, I want to have git return a `RichError`/`ExecError`, but that's a can of worms for another time.

sipsma (approved):

grouville (approved):
Pretty cool !

jedevc:
Pushed one more commit based on #10959 to properly test that the auto-determined URLs are actually correct.

Issue #9362, opened 2025-01-11
🐞 Dagger Fails to Honor GOPROXY Environment Variable


crrow:
### What is the issue?

Dagger is not correctly recognizing or using the GOPROXY environment variable to configure the cmd environment when executing go cmd.

### Dagger version

dagger v0.15.1 (registry.dagger.io/engine:v0.15.1) darwin/arm64

### Log output

<img width="1144" alt="image" src="https://github.com/user-attachments/assets/89f5feea-d253-4b20-8cbd-9bf1eddf4894" />

gtshepard:
Any plans to support ```GOPROXY```? even if it's in this style via [dagger.json](https://docs.dagger.io/extending/packages/) config

Currently quite cumbersome to get a dagger module to pull private golang package when it's installed as a dependency in a dagger go project

Ideally would like to do anonymous pulls from an internal private registry url I set in GOPROXY.

If must use ```GOPRIVATE``` it means only option is authenticating via ssh. This has led to more complicated local development experience and infra setup (infra to use dagger in Github Actions workflows).

Essentially have to setup ssh agent forwarding for dagger engine.


## Locally (mac)
 
```bash
eval "$(ssh-agent -s)"
```

```bash
SSH_AUTH_SOCK=`launchctl getenv SSH_AUTH_SOCK` ssh-add
```

```bash
docker run --rm \
  --mount type=bind,src=/run/host-services/ssh-auth.sock,target=/run/host-services/ssh-auth.sock \
  -e SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock \
  -v /var/lib/dagger \
  -v ./engine.json:/etc/dagger/engine.json \
  --name dagger-engine-custom-5555 \
  --privileged \
  registry.dagger.io/engine:v0.18.16
```

in separate shell

```bash
export _EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-container://dagger-engine-custom-5555
```

then call the dagger func that relies on the go private package

```bash
dagger --progress=plain call \
       greet \
       --name jim
```

## GHA Workflows

Deploying dagger helm chart for engine requires similar idea with ssh agent side car. 

Adds complexity to the k8s deployment and adds additional operational overhead managing a key/account

gtshepard:
After digging deeper found GOPROXY can be set https://github.com/dagger/dagger/issues/6599

For the dagger engine that you spin up and and connect to, the this environment var

```bash
[...]

PR #10962, opened 2025-08-29
docs: mention GOPROXY configuration

jedevc:
See https://github.com/dagger/dagger/issues/9362.

We've supported this functionality since #7255, but we never added it in our docs (see #7814). I think it's worth including, otherwise there's really no way to know it exists.

It's hacky, but worth including :tada: We can easily our docs once we have a better method.


jpadams (approved):
Thank you!

PR #10965, opened 2025-08-29
Grab bag of engine optimizations

sipsma:
Needed to debug some perf regressions in https://github.com/dagger/dagger/pull/10761; while doing so I found a lot of extremely low hanging fruit. Just gonna group them together rather than send out 5+ separate PRs.

Draft because I'm pushing more commits here still. Pushing them one-by-one so I can verify they independently fix individual bottlenecks (i.e. don't make each other redundant) and don't have any dependency on #10761 

---

[dagql: optimize InputSpecs.Inputs](https://github.com/dagger/dagger/pull/10965/commits/d68a30b39ac99a33890e0d87f12e65f400f25e13)

Before this, a pprof cpu profile of the engine while running TestTelemetry showed that 7-10% of our cpu time was spent in dagql.InputSpecs.Inputs. Most of this was due to the `seen` map creation and read/writes.

Given the small size of `n` for the function, using an O(n^2) scan of the slice to check for whether an arg was already seen is actually much more performant than the map.

After that fix, Inputs went down to only consuming 0.1% of the cpu time. It could probably be optimized further by e.g. caching results but given its no longer a bottleneck that extra complication doesn't seem worth it at the moment.

---

[optimize ID calcDigest](https://github.com/dagger/dagger/pull/10965/commits/1acfc183633f2ab35dcdfbe62e8f6712287334b6)

After the previous optimizations in this commit series, ID.calcDigest was using ~13% of cpu time during runs of TestTelemetry.

The biggest bottleneck in it was (surprisingly) that it grew the stack enough to trigger a resize, which the go runtime spent a lot of work doing. Since `calcDigest` is called *a lot*, that resizing added up quite a bit and was much more expensive than the actual marshaling and hashing.

To fix the usage of stack size:
1. The xxh3 hash object is stored in a sync.Pool. That hasher includes a >1kb array (not slice) and it appears the call to xxh3.New() was always being inlined, so it didn't end up in the heap. Using a sync.Pool ensures it is heap-stored while also saving work allocating/GC'ing the object.
2. After the above, the protobuf marshal code started being the one triggering stack size growth. To fix that, we now roll our own by just writing each relevant pb field to the hash. This requires some tedious manual effort + maintaince, but given the perf benefits seems worth it.

After all the above, `calcDigest` went from 13% of cpu usage down to 0.9%.

---

[fix local source stat hash byte append](https://github.com/dagger/dagger/pull/10965/commits/c89216e84cdd867ba74a240033adb4480e546105)

We were appending to a []byte but not using the newly updated slice, so the data written was just being overwritten each time. Surprisingly, this didn't *noticeably* break anything yet? I just noticed it while working on something else in the ID code that also used the same Append methods.

---

TODO:
- [ ] ~~Disable accidentally enabled content-hashing in dag-op for non-container ops, which results in 30GB+ of (very short lived) memory allocations over the course of just 1m of integ test execution~~
   * Nevermind, seems to be reliant on #10761, will include there instead
- [ ] ~~Reduce goroutine creation in various places; 200k+ are spawned in a single minute of integ test execution, which is possibly a genuine burden on the go runtime scheduler~~
   * Didn't seem to make much of a difference with the other stuff right now
- [x] optimize id `calcDigest` (avoid fmt.Sprintf, etc.)
- [ ] ~~avoid frequent running of extremely complicated queries for introspection+typeDefs; dagql is not optimized for it (it's better at things that are "core API"-shaped) and it results in massive amounts of work~~
   * This is VERY much worth doing (the amount of calls this triggers is mind boggling) but I'm running into weird issues with it right now where CI seems to sometimes be faster but then randomly take 20+ minutes? So I'll skip this for now
   * Spun out to a separate pr here https://github.com/dagger/dagger/pull/10967

### DIFF
--- a/dagql/call/id.go
+++ b/dagql/call/id.go
@@ -3,7 +3,9 @@ package call
 import (
 	"encoding/base64"
 	"encoding/binary"
+	"encoding/hex"
 	"fmt"
+	"math"
 	"strings"
 	"sync"
 
@@ -19,6 +21,10 @@ import (
 var marshalBufPool = &sync.Pool{New: func() any {
 	b := make([]byte, 0, 1024)
 	return &b
+}}
+
+var hasherPool = &sync.Pool{New: func() any {
+	return xxh3.New()
 }}
 
 func New() *ID {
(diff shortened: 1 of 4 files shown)


dagql/objects.go
```diff
@@ -1093,19 +1093,47 @@ func (specs InputSpecs) Input(name string, view View) (InputSpec, bool) {
 }
 
 func (specs InputSpecs) Inputs(view View) (args []InputSpec) {
-	seen := make(map[string]bool, len(specs.raw))
+	// This function is currently in the hot path, so we optimize duplicate checks by only using
+	// a map when the number of args is above a certain threshold, using slice iteration otherwise.
+	// The previous implementation that only used a map was a genuine bottleneck since most of the
+	// time the number of args is small.
+
+	const useMapThreshold = 15 // based on some benchmarks on an m4 laptop, fairly approximate though
```
jedevc, line 1099: Are we ever actually ending up in this path? More than 15 inputs?

dagql/call/id.go
```diff
@@ -485,17 +491,169 @@ func (id *ID) calcDigest() (string, error) {
...
+	buf, err = appendLiteralBytes(arg.Value, buf)
+	if err != nil {
+		return nil, fmt.Errorf("failed to append argument %q: %w", arg.Name, err)
+	}
+
+	buf = append(buf, 0)
+	return buf, nil
+}
+
+func appendLiteralBytes(lit *callpbv1.Literal, buf []byte) ([]byte, error) {
```
jedevc, line 595: nit: tiny comment explaining this?

dagql/objects.go
sipsma: Yeah I saw 17 at some point

PR #10967, opened 2025-08-31
optimize out complex introspection queries

sipsma:
Introspection queries go through dagql and trigger a pretty insane
amount of calls to dagql.Server.Resolve each time a client connects
or a new server is made.

This isn't a currently bottleneck on machines with enough CPU, but it
makes pprof traces almost unusable at times due to the sheer number of
goroutines. I would not be shocked if there are scenarios where it is a
genuine bottleneck too (especially in cases of lots of concurrent
clients).

This change updates some of our internal calls to introspection queries
to be implemented directly rather than as an actual dagql query. It also
tweaks dagql.Server.Resolve a bit to only use separate goroutines when
there are multiple selections, otherwise just calling directly.

To test the difference, the number of times Resolve gets called when
running `dagger -m ./modules/alpine call --help`:
* Before: 23712
* After: 3548

There are also pprof traces that previously were not even loadable in my
browser that I can now open successfully.

This isn't a complete fix; the CLI typedef queries appear to be another
significant source that could be optimized out. And like I said there
isn't a immediately obvious bottleneck getting cleared here. But still
seems worth it overall.

### DIFF
--- a/core/schema/coremod.go
+++ b/core/schema/coremod.go
@@ -2,7 +2,6 @@ package schema
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 	"sync"
@@ -11,6 +10,7 @@ import (
 	"github.com/dagger/dagger/core"
 	"github.com/dagger/dagger/dagql"
 	"github.com/dagger/dagger/dagql/call"
+	dagqlintrospection "github.com/dagger/dagger/dagql/introspection"
 	"github.com/dagger/dagger/engine/server/resource"
 	"github.com/dagger/dagger/engine/slog"
 	"github.com/opencontainers/go-digest"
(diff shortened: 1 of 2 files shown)


helderco:
I really want to move codegen to depend on an API call which results in a JSON structure based on TypeDefs. In that case there's no longer a need to generate the schema just to get a (lossy) GraphQL introspection result just for codegen/introspection. Only needed while serving the schema.

Would you say that would bring a significant performance improvement if we did that (even after your optimization), or not much of a difference?

sipsma:
> I really want to move codegen to depend on an API call which results in a JSON structure based on TypeDefs. In that case there's no longer a need to generate the schema just to get a (lossy) GraphQL introspection result just for codegen/introspection. Only needed while serving the schema.
> 
> Would you say that would bring a significant performance improvement if we did that (even after your optimization), or not much of a difference?

@helderco Yeah, actually while this PR gets rid of some fully internal calls to introspection queries, it currently still leaves in the CLI's big+complicated typedefs query. If we replaced that with a single API call that just returned the whole set of typedefs as a json string (I *think* that's what you're suggesting, lemme know if misunderstanding), that would also help a lot with reducing overhead latency of basically any CLI command I think.

helderco:
> [...] If we replaced that with a single API call that just returned the whole set of typedefs as a json string (I _think_ that's what you're suggesting, lemme know if misunderstanding), that would also help a lot with reducing overhead latency of basically any CLI command I think.

Yes, I think that's correct, but it's not just the CLI, it's about generating SDK client bindings from it too.

Basically centralizing introspection (primed for codegen) on the same data structure, and making codegen templates much simpler and easier to maintain. Not to mention much faster to generate client bindings, e.g., in the Python SDK Go module rather than executing a Python script in a container. Or not requiring building, mounting, and executing an external `cmd/codegen` in TS (which has an issue when using the full git ref as the sdk), keeping those (much simpler) templates in `sdk/typescript`, etc.

Don't mean to go off topic on that, was just wondering if there would be a measurable performance gain in the engine as well, compared to having to extend and serve the GraphQL schema for the GraphQL introspection. I don't know what's the overhead of those two things. While building the schema introspection JSON would no longer be needed, extending the schema itself is probably cached and reused when actually calling functions? Does it take a bit to start and stop serving every time, or does it stay served once, even from building the `introspectionJSON` file?[^1]

[^1]: Just in case you know or have a good idea. Otherwise disregard 🙂

PR #10969, opened 2025-09-01
Improve docs on private dependencies

jedevc:
https://github.com/dagger/dagger/pull/10960 should vastly simplify installation of private dependencies.

It's now easier to just skip the explicit `https`/`ssh` proto, since it can be automagically determined in more cases. I've updated the docs to make this clearer.

I've also moved the info on private repos directly next to the info on installing other dependencies, instead of living in remote repositories, where I don't really expect it to be.

### DIFF
--- a/docs/current_docs/extending/modules/module-dependencies.mdx
+++ b/docs/current_docs/extending/modules/module-dependencies.mdx
@@ -5,6 +5,21 @@ description: "Learn how to manage module dependencies in Dagger"
 
 # Module Dependencies
 
+A module reference follows the format: `[proto://]host/repo[/subpath][@version]`, where:
+- `host` is the domain name of the Git hosting service, e.g. `github.com`.
+- `repo` is the repository path, typically in the format `owner/name`.
+- `proto://` is optional and can be either `ssh://` or `https://` if you want to be explicit about the protocol. If omitted, Dagger will automatically choose the protocol based on available authentication methods.
+- `/subpath` is optional and specifies a subdirectory within the repository, useful for monorepos.
+- `@version` can be a tag, branch, or commit. If omitted, the default branch is used.
+
+For example, in the reference `github.com/shykes/daggerverse/hello@v0.3.0`:
+- `github.com` is the host
+- `shykes/daggerverse` is the repository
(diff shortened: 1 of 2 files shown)


docs/current_docs/extending/modules/module-dependencies.mdx
```diff
@@ -152,6 +167,18 @@ dagger install ./path/to/module
 Installing a module using a local path (relative or absolute) is only possible if your module is within the repository root (for Git repositories) or the directory containing the `dagger.json` file (for all other cases).
 :::
 
+### Private modules
+
+You can also install private modules from remote repositories, as long as you have access to them. Dagger supports authentication via both HTTPS (using Git credential managers) and SSH (using `SSH_AUTH_SOCK`). For more information, see [Remote Repositories](/extending/remote-repositories).
```
vikram-dagger:
```suggestion
You can also install private modules from remote repositories, as long as you have access to them. Dagger supports authentication via both HTTPS (using Git credential managers) and SSH (using `SSH_AUTH_SOCK`). For more information, see the [documentation on remote repositories](../extending/modules/remote-repositories.mdx).
```

We use filesystem paths in links because Docusaurus can then detect breaks when building the docs site. It is unable to detect breaks when using URL paths.

docs/current_docs/extending/modules/module-dependencies.mdx
```diff
@@ -152,6 +167,18 @@ dagger install ./path/to/module
...
+
+When installing a private module, you can use either the normal ref style (preferred if possible, since it allows any available auth method to be used) or be explicit, and specify the protocol. For example, the following commands install the same private module:
```
vikram-dagger:
```suggestion
When installing a private module, you can use either the normal ref style (preferred if possible, since it allows any available authentication method to be used), or explicitly specify the protocol. For example, the following commands all install the same private module:
```

vikram-dagger (approved):
Thank you @jedevc - minor fixes needed, approving in advance so you can merge

jedevc:
Holding merge until just before/after v0.18.17, since the PR that improves this isn't released yet!

vito (approved):

PR #10973, opened 2025-09-02
feat: add dagger client update command

TomChv:
Add `dagger client update` command so you can update external client generator.

Add `WithUpdatedClients` to `ModuleSource` to update one or more clients.
Add integration tests.

Additional changes:
- Small fix on client install to clean the path in the generator is a local module

### DIFF
--- a/core/integration/client_generator_test.go
+++ b/core/integration/client_generator_test.go
@@ -1300,7 +1300,7 @@ func main() {
 
 func (ClientGeneratorTest) TestClientUpdate(ctx context.Context, t *testctx.T) {
 	const (
-		// pins from github.com/TomChv/dagger-client-generator module
+		// pins from github.com/dagger/client-generator-test module
 		// to test the update command
 		latestPin = "d76af75d104a2777bcd7dc6d240966760801b81f"
 		v01Pin    = "446f2691deba58f99b55b86430fade1c773e486f"
@@ -1316,7 +1316,7 @@ func (ClientGeneratorTest) TestClientUpdate(ctx context.Context, t *testctx.T) {
 	"name": "test",
 	"clients": [
 		{
-			"generator": "github.com/TomChv/dagger-client-generator@` + v01Pin + `",
+			"generator": "github.com/dagger/client-generator-test@` + v01Pin + `",
 			"directory": "dagger"
 		}
 	]
@@ -1326,7 +1326,7 @@ func (ClientGeneratorTest) TestClientUpdate(ctx context.Context, t *testctx.T) {
 	"name": "test",
 	"clients": [
 		{
-			"generator": "github.com/TomChv/dagger-client-generator@main",
+			"generator": "github.com/dagger/client-generator-test@main",
 			"directory": "dagger"
 		}
 	]
@@ -1336,7 +1336,7 @@ func (ClientGeneratorTest) TestClientUpdate(ctx context.Context, t *testctx.T) {
 	"name": "test",
 	"clients": [
 		{
-			"generator": "github.com/TomChv/dagger-client-generator",
+			"generator": "github.com/dagger/client-generator-test",
 			"directory": "dagger"
 		}
 	]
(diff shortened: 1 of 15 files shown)


core/schema/modulesource.go
```diff
@@ -2501,13 +2507,159 @@ func (s *moduleSourceSchema) moduleSourceWithClient(
...
+
+	src = src.Clone()
+	newClientConfig := make([]*modules.ModuleConfigClient, len(src.ConfigClients))
+	for i, client := range src.ConfigClients {
+		clientGeneratorSource, _, _ := strings.Cut(client.Generator, "@")
+
+		// If the client is a builtin SDK, the version is tied to the engine so we skip it.
+		if sdk.IsModuleSDKBuiltin(client.Generator) {
+			newClientConfig[i] = client.Clone()
+
```
marcosnils:
```suggestion
```

core/schema/modulesource.go
```diff
@@ -2501,13 +2507,159 @@ func (s *moduleSourceSchema) moduleSourceWithClient(
...
+			continue
+		}
+
+		// If there is an update request but the client is not in the input list, skip it
+		if _, ok := updateReqs[clientGeneratorSource]; !ok && len(updateReqs) > 0 {
+			newClientConfig[i] = client.Clone()
+
```
marcosnils:
```suggestion
```

core/schema/modulesource.go
```diff
@@ -2501,13 +2507,159 @@ func (s *moduleSourceSchema) moduleSourceWithClient(
...
+		// If the client is a builtin SDK, the version is tied to the engine so we skip it.
+		if sdk.IsModuleSDKBuiltin(client.Generator) {
+			newClientConfig[i] = client.Clone()
+
+			continue
+		}
+
+		// If there is an update request but the client is not in the input list, skip it
+		if _, ok := updateReqs[clientGeneratorSource]; !ok && len(updateReqs) > 0 {
```
marcosnils, line 2587: checking for len(updateReqs) here seems unnecessary?  the if condition is already checking for that.

core/schema/modulesource.go
```diff
@@ -2501,13 +2507,159 @@ func (s *moduleSourceSchema) moduleSourceWithClient(
...
+		// At that point, we know that the client must be updated either with a given version
+		// or for a global update.
+		var clientModule dagql.ObjectResult[*core.ModuleSource]
+		clientModule, err = core.ResolveDepToSource(ctx, bk, dag, src, client.Generator, "", "")
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve client module source: %w", err)
+		}
+
+		// Ignore local dependency except if it's in the list of update then we throw an error
+		if clientModule.Self().Kind == core.ModuleSourceKindLocal {
```
marcosnils, line 2601: isn't there a way to check if a client is local without calling the `resolveDepToSource`? I'm not super familiarized  with the code here but if there was a way to know this before resolving to source I'd assume this could speed things up considerably.

core/schema/modulesource.go
```diff
@@ -2501,13 +2507,159 @@ func (s *moduleSourceSchema) moduleSourceWithClient(
...
+		delete(updateReqs, clientGeneratorSource)
+	}
+
+	// Verify that all updateReq has been processed, otherwise there's a
+	// invalid update request.
+	if len(updateReqs) > 0 {
+		deps := make([]string, 0, len(updateReqs))
+		for updateReq := range updateReqs {
+			deps = append(deps, updateReq)
+		}
```
marcosnils: nit: can be simplified with `maps.Keys(updateReqs)

core/schema/modulesource.go
```diff
@@ -2501,13 +2507,159 @@ func (s *moduleSourceSchema) moduleSourceWithClient(
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to get buildkit client: %w", err)
+		}
+
+		clientModule, err := core.ResolveDepToSource(ctx, bk, dag, src, moduleConfigClient.Generator, "", "")
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve client module source: %w", err)
+		}
+
+		if clientModule.Self().Kind == core.ModuleSourceKindLocal {
```
marcosnils, line 2535: same as https://github.com/dagger/dagger/pull/10973/files#r2349983365.

core/integration/client_generator_test.go
```diff
@@ -1298,6 +1298,175 @@ func main() {
 	})
 }
 
+func (ClientGeneratorTest) TestClientUpdate(ctx context.Context, t *testctx.T) {
+	const (
+		// pins from github.com/TomChv/dagger-client-generator module
```
marcosnils: let's put this module somewhere else? IIUC if you make a single commit to that repo to fix something, some of the below tests will break. Maybe move it to? https://github.com/dagger/dagger-test-modules

core/integration/client_generator_test.go
```diff
@@ -1298,6 +1298,175 @@ func main() {
...
+			name:       "existing client has branch, update cmd has no version",
+			daggerjson: clientwithBranch,
+			updateCmd:  []string{"client", "update", "github.com/TomChv/dagger-client-generator"},
+			contains:   []string{fmt.Sprintf("github.com/TomChv/dagger-client-generator@%s", latestPin)},
+			notContains: []string{
+				"github.com/TomChv/dagger-client-generator@main",
+			},
+		},
+		{
+			name:       "existing client has version, update cmd has branch",
```
marcosnils, line 1405: client generators are not pinned? seems like we should follow the same pattern we did for modules and pin them as well since this  has a high chance of breaking things. Shall we open a follow-up issue/PR?

core/schema/modulesource.go
TomChv: No it's actually important as explained in the comment, if the `updateReqs` list is empty, that means the CLI command is `dagger client update` so all dependencies must be updated.  However, if you execute `dagger client update github.com/gen/my-gen`, you ask for a specific client gen to be updated.

The core loop go through all clients in the `dagger.json` so if you have 3 clients in your dagger.json, like `go`, `typescript` and `github.com/gen/my-gen`, the core loop will always go through them.
That means when the loop is on `go`: `updateReqs[clientGeneratorSource]` will return empty but if we see that `len(updateReqs)` is not empty, that means the user wanted to update a specific dependency so we skip the current one.

core/schema/modulesource.go
TomChv: That's the only way I found, `resolveDepToSource` is a convenient function where you can pass a git URL or a path and it will convert it to a module.
Technically we could naively parse the `client.Generator` string and check if it's a path or an URL but that would be flacky because your dependency may be in a `github.com/foo/bar` repository (not likely to happen but could)

We could reduce the function to just call `core.ParseRefString` but that means we cannot verify that the module is accessible by the engine etc... 

@jedevc Maybe you have an idea on that?

core/integration/client_generator_test.go
TomChv: We can move it yes, but it's not a blocker since I do not plan to commit on that repo, it was only for tests like we do for `dagger update` integration tests
 https://github.com/dagger/dagger/blob/91f59ea253ca099dcd5137e6489c5d6671efe64b/core/integration/module_cli_test.go#L1467

core/integration/client_generator_test.go
TomChv: I could add that field in a follow up yes

core/schema/modulesource.go
TomChv: I'm collecting the map to print it, with `map.Keys` I get an iterator that I cannot print directly so it's simpler that way.

Maybe I'm missing something around iterator but I'm not sure you can print them directly (I tried but I'm just getting the address)

core/schema/modulesource.go
marcosnils: > We could reduce the function to just call `core.ParseRefString` but that means we cannot verify that the module is accessible by the engine etc...

 I was thinking about doing this at this stage actually. Should we care about validating this at this stage or shall we let it fail afterwards since it'll have to be resolved again later when we call the generation part? In any case, I don't think this is going to be that big of a deal since I'd assume non-bultin generators will take some time to become popular. I'm good with leaving as-is for now. We could revisit if this becomes a bottleneck in the future.

core/schema/modulesource.go
marcosnils: yeah.. I meant `slices.Collect(maps.Keys(updateReqs))`

core/integration/client_generator_test.go
marcosnils: > it was only for tests like we do for `dagger update` integration tests

Well.. we should eventually move that repo as well :grimacing:.

I also don't think it's the same since the `shykes/daggerverse` repo had multiple commits after those tests were added and they didn't break. Seems to me they're careful crafted so the tests don't break if the repo has more commits. I don't think it's the same for the tests in this PR. 

I can quickly create the `dagger-client-generator` repo in the dagger org and share that :pray: . We can also mark it as read-only once tests have moved there. Let's add that to the readme :pray:

core/schema/modulesource.go
TomChv: Oh thanks, I have no experience with iterators in Go, I didn't know that exist! Thanks

TomChv: Done

core/integration/client_generator_test.go
TomChv: Done

PR #10975, opened 2025-09-02
wip: generalized function caching

sipsma:
Trying this out to see how much breaks immediately. The final change is going to have to deal with lots of back-compat issues and most likely a million corner cases around secrets. For now ignoring those to see how much bubbles up.

Originally the plan was to build this all on top of dagql persistent caching, but given recent conversations it probably makes sense to see if we can fast-path it and then rebase it on top of the new persistent cache once that's ready.

### DIFF
--- a/core/modfunc.go
+++ b/core/modfunc.go
@@ -292,12 +292,14 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Any
 	}
 
 	var cacheMixins []string
-	if !opts.Cache {
-		// Scope the exec cache key to the current session ID. It will be
-		// cached in the context of the session but invalidated across
-		// different sessions.
-		cacheMixins = append(cacheMixins, clientMetadata.SessionID)
-	}
+	/*
+		if !opts.Cache {
+			// Scope the exec cache key to the current session ID. It will be
+			// cached in the context of the session but invalidated across
+			// different sessions.
+			cacheMixins = append(cacheMixins, clientMetadata.SessionID)
+		}
+	*/
 	if !opts.SkipCallDigestCacheKey {
 		// If true, scope the exec cache key to the current dagql call digest. This is needed currently
 		// for module function calls specifically so that their cache key is based on their arguments and


shykes:
Dropping my thoughts on DX, since we just discussed it live...

## Option 1: two pragmas

One pragma to disable, another to configure.

| Feature | Go pragma |
| -- | -- |
| Disable caching | `+nocache` |
| Configure caching | `+ttl=session\|DURATION` |

| Desired behavior | How to configure it |
| -- | -- |
| Maximum caching | - (default) |
| No caching | `+nocache` |
| Cache for the current session | `+ttl=session` |
| Cache for 1 hour | `+ttl=1h` |

## Option 2: one pragma

One pragma to rule them all.

| Feature | Go pragma |
| -- | -- |
| Configure caching | `+cache=never\|session\|DURATION` |

| Desired behavior | How to configure it |
| -- | -- |
| Maximum caching | - (default) |
| No caching | `+cache=never` |
| Cache for the current session | `+cache=session` |
| Cache for 1 hour | `+cache=1h` |

I like option 2 best, because you can't beat the simplicity of a single pragma called `cache` :) I'm slightly sad that the word `ttl` doesn't show up anywhere, but that might be for the best.

cc @sipsma @vito @jedevc @marcosnils

sipsma:
@shykes Yeah I agree having a single pragma is appealing and simple. Even for hypothetical cases like "cache per session and also with a ttl" you could still easily support something like `+cache=session,1m`

EDIT: or I suppose just multiple `//+cache` pragmas like
```go
// +cache=session
// +cache=1m
```

shykes:
@sipsma if we go with option 2 (one pragma), I'm OK with *not* allowing multiple `+cache` directives, at least initially. This means you won't be able to express "cache this function up to 1mn or until the end of the session, whichever comes first". But a) I can live with that, and b) we can always add it later if people ask for it.

(yes I know I was asking for the opposite just an hour ago :) But I didn't expect a single pragma to work so well)

engine/cache/db/db.go
sipsma, line 1: As is probably evident, I originally used sqlc to generate the code in this package. Unfortunately while using sqlc I hit multiple parser bugs in it (known upstream, can dig up the issues if interested) and ended up just ejecting to manage by hand for now. It's small+simple enough right now that it seems reasonable.

util/hashutil/hash.go
sipsma, line 1: There was a point where the refactor to use this package was more directly related to this PR because the cache logic was mixing expirations into digests and thus needed to do it quickly. After some fixes I ended up having to mix in the session ID string instead, which severed some of the relationship to this PR.

I'll default to just leaving this in since it's a good change to have, but if there's any preference I can try to spin it out to a separate PR.

cmd/engine/.dagger/build/builder.go
```diff
@@ -148,6 +148,8 @@ func (build *Builder) Engine(ctx context.Context) (*dagger.Container, error) {
 					"e2fsprogs",
 					// for Directory.search
 					"ripgrep",
+					// for dbs
+					"sqlite",
```
vito, line 152: Is this mostly for troubleshooting, to have `sqlite3` around when we exec into the engine? (The engine already uses SQLite without this so assuming so)

core/integration/module_test.go
```diff
@@ -6452,6 +6453,281 @@ func (m *Test) Nothing() (*dagger.Directory, error) {
...
+
+		out1a, err := modGen1.
+			WithEnvVariable("CACHE_BUST", rand.Text()).
+			With(daggerCall("test-set-secret")).Stdout(ctx)
+		require.NoError(t, err)
+		out1b, err := modGen1.
+			WithEnvVariable("CACHE_BUST", rand.Text()).
+			With(daggerCall("test-set-secret")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, out1a, out1b)
```
vito, line 6711: What is being compared here? The `Container@xxh3:abcdef` hashes that `dagger call` prints?

core/integration/module_test.go
```diff
@@ -6452,6 +6453,281 @@ func (m *Test) Nothing() (*dagger.Directory, error) {
...
+			WithEnvVariable("CACHE_BUST", rand.Text()).
+			With(daggerCall("test-set-secret")).Stdout(ctx)
+		require.NoError(t, err)
+		out2b, err := modGen2.
+			WithEnvVariable("CACHE_BUST", rand.Text()).
+			With(daggerCall("test-set-secret")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, out2a, out2b)
+
+		require.NotEqual(t, out1a, out2a)
```
vito, line 6727: Pretty interesting - how does this work? (Edit: a ha, found it: https://github.com/dagger/dagger/pull/10975/files#diff-6e7d0c02c8bd065fb312e4109724ee002cb041cd44c4c6e97abe267ce31cedf4R845)

If `SetSecret` wasn't called by the function, and instead the `rand.Text()` was set as a plain env var, I suppose it would have been cached across both sessions?

core/integration/module_typescript_test.go
```diff
@@ -1227,23 +1227,23 @@ import { object, func } from "@dagger.io/dagger"
 
 @object()
 export class SubSub {
-  @func("zoo")
+	@func({ alias: "zoo" })
```
vito, line 1230: Backwards compatible I assume?

(edit: yep - https://github.com/dagger/dagger/pull/10975/files#diff-1b38594299b58a410cef08f33c6fe98ba83c019c1dc69b648858eb29b10644bcR44)

core/modules/config.go
```diff
@@ -77,6 +77,10 @@ type ModuleConfig struct {
 
 	// The clients generated for this module.
 	Clients []*ModuleConfigClient `json:"clients,omitempty"`
+
+	// If true, disable the new default function caching behavior for this module. Functions will
+	// instead default to the old behavior of per-session caching.
+	DisableDefaultFunctionCaching *bool `json:"disableDefaultFunctionCaching,omitempty"`
```
vito, line 83: An escape hatch for modules that bump their engine version but haven't made necessary adjustments, right? (Nice)

Or, does this get set automatically on upgrade, and module authors have to remove it? Seems safest, but maybe _too_ safe?

core/schema/module.go
```diff
@@ -156,6 +157,13 @@ func (s *moduleSchema) Install(dag *dagql.Server) {
 				dagql.Arg("ignore").Doc(`Patterns to ignore when loading the contextual argument value.`),
 				dagql.Arg("sourceMap").Doc(`The source map for the argument definition.`),
 			),
+
+		dagql.Func("withCachePolicy", s.functionWithCachePolicy).
+			Doc(`Returns the function updated to use the provided cache policy.`).
+			Args(
+				dagql.Arg("policy").Doc(`The cache policy to use.`),
+				dagql.Arg("timeToLive").Doc(`The TTL for the cache policy, if applicable. Provided as a duration string, e.g. "5m", "1h30s".`),
```
vito, line 165: nit: Should this just accept a value in seconds? Not sure if we've wedded ourselves to Go's duration format elsewhere. SDKs touch this so I think it'd be better to be as dry and universal as possible. Different languages will have their own native duration syntax, like `1.day` or something in Ruby, so having to convert that to Go's particular format seems awkward.

util/hashutil/hash.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (h *Hasher) WithFloat64(f float64) *Hasher {
+	*h.bufPtr = binary.BigEndian.AppendUint64(*h.bufPtr, math.Float64bits(f))
+	*h.bufPtr = append(*h.bufPtr, 0)
+	return h
+}
+
+func (h *Hasher) WithDelim() *Hasher {
+	*h.bufPtr = append(*h.bufPtr, 0)
```
vito, line 89: neat

core/object.go
```diff
@@ -498,12 +491,8 @@ func objFun(ctx context.Context, mod *Module, objDef *ObjectTypeDef, fun *Functi
 		Spec: &spec,
 		Func: func(ctx context.Context, obj dagql.ObjectResult[*ModuleObject], args map[string]dagql.Input, view call.View) (dagql.AnyResult, error) {
 			opts := &CallOpts{
-				ParentTyped:  obj,
-				ParentFields: obj.Self().Fields,
-				// TODO: there may be a more elegant way to do this, but the desired
-				// effect is to cache SDK module calls, which we used to do pre-DagQL.
-				// We should figure out how user modules can opt in to caching, too.
```
vito, line 505: Nice seeing this comment go away :P

dagql/types.go
```diff
@@ -49,8 +49,8 @@ type ObjectType interface {
 	//
 	// Unlike natively added fields, the extended func is limited to the external
 	// Object interface.
-	// cacheKeyFun is optional, if not set the default dagql ID cache key will be used.
-	Extend(spec FieldSpec, fun FieldFunc, cacheSpec CacheSpec)
+	// cacheConfigFunc is optional, if not set the default dagql ID cache key will be used.
+	Extend(spec FieldSpec, fun FieldFunc, cacheConfigFunc GenericGetCacheConfigFunc)
```
vito: Just curious: is there a reason `cacheConfigFunc` couldn't be folded into FieldSpec?

I moved `CacheSpec` into `FieldSpec` as a side refactor momentarily on a PR because I wanted to be able to access it generically through `ObjectType.FieldSpec` - right now it only exists on `Field[T]` so you can't grab it from an `ObjectType`, only a `Class[T]`, presuming you already know the `T`.

Ended up not needing that so I rolled back the refactor, but still wondering if there's a good reason for the split.

engine/cache/db/schema.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS calls (
+    call_key TEXT PRIMARY KEY,
+    storage_key TEXT NOT NULL,
```
vito, line 3: Cool cool, so SQLite is just storing pointers to arbitrarily large objects in the existing cache storage

engine/version.go
```diff
@@ -44,6 +44,16 @@ var (
...
+
+	// TODO: ADJUST VERSION BASED ON WHEN RELEASE HAPPENS
+	// TODO: ADJUST VERSION BASED ON WHEN RELEASE HAPPENS
+	// TODO: ADJUST VERSION BASED ON WHEN RELEASE HAPPENS
+	// TODO: ADJUST VERSION BASED ON WHEN RELEASE HAPPENS
+	// TODO: ADJUST VERSION BASED ON WHEN RELEASE HAPPENS
+	// TODO: ADJUST VERSION BASED ON WHEN RELEASE HAPPENS
+	// MinimumDefaultFunctionCachingModuleVersion is the minimum module version at which
+	// we will enable default function caching behavior.
+	MinimumDefaultFunctionCachingModuleVersion = "v0.19.2"
```
vito: (flagging since v0.19.3 will be out soon)

vito (commented):
Just one nit re: the SDK API, the rest are mostly questions I answered by reviewing more, otherwise lgtm!

engine/cache/db/schema.sql
sipsma: Yeah, for now this is essentially a key to the large object storage in buildkit cache (speaking loosely), but once that part is also brought into the dagql fold storage_key will likely be just a foreign key to another table in this same sqlite db

cmd/engine/.dagger/build/builder.go
sipsma: Yeah I needed it for debugging at first but decided we may as well leave it in.

Also flirted with trying https://github.com/cvilsmeier/sqinn-go (which would require its installation) but decided to try that later

core/integration/module_test.go
sipsma: Ah yes, which I suppose technically works but actually wasn't the intention and is confusing to think through in terms of whether it's actually testing what it wants. Updating to have it do "with-exec", "--args", `sh,-c,echo $TOP_SECRET | rev`

core/integration/module_test.go
sipsma: > If SetSecret wasn't called by the function, and instead the rand.Text() was set as a plain env var, I suppose it would have been cached across both sessions?

Yep exactly. We don't actually persist anything to the db until it's been proven safe, so nothing will hit cache until that time. In the case of a SetSecret, that just never happens.

core/modules/config.go
sipsma: > Or, does this get set automatically on upgrade, and module authors have to remove it? Seems safest, but maybe too safe?

That's what I went with yes. Agree there's a big tradeoff here since now users have to take action to get faster, but I really don't want them to be greeted with broken modules after an upgrade, especially a situation like this where it's not e.e.g a simple compilation error but instead a confusing issue involving caching.

To be clear, brand new init'd modules won't get the function cache disabled; it's only module from older versions that get it after a dagger develop

dagql/types.go
sipsma: No pretty sure it can, I'll do that here too

core/schema/module.go
sipsma: Hm yeah good point. In the go sdk we'd still want to support string duration since e.g. `//+cache=45m` is much clearer than `//+cache=<whatever 45 * 60 is>`, but the Go SDK can just parse the duration string into seconds. In other languages with actual decorators being able to use something more native would be very nice, I agree.

I'll try it out

sipsma: Think I'm gonna back out of this for now actually. I'm not super convinced it's actually that much better for users, e.g. in python it goes from
```python
@dagger.function(cache="40s")
```
to
```python
from datetime import timedelta
@dagger.function(cache=timedelta(seconds=40))
```

which has pros and cons, sort of a wash.

And for the ts SDK there isn't really a standard time duration type that's also available across all of node, deno + bun. So it ends up being the same sort of UX just with a lot more work for the SDK to implement.

I think what'd make more sense is to:
1. keep the current arg as a duration string
   * this makes SDK support trivial since they can just passthrough a string from the user
2. down the line, if an SDK wants to add some native support, they easily can and all they have to do is convert the native duration type to seconds and then the equivalent of `fmt.Sprintf("%ds", durationSeconds)`, which isn't too much overhead for them.

vito: Yeah that's fine.

FWIW my only reservation was the fact that this is in the _core API_, which users don't touch, SDKs implementors do. How the value manifests to users is still at the discretion of each SDK. If it's that painful to use a proper duration type in Python, I imagine the Python SDK author would simply not make users do that. :smile:

What I'm being cautious about are little Go-specific details leaking into the cross-language substrate. Things we take for granted from using Go all the time, but are unusual to folks coming from different ecosystems.

But, durations are probably fine; we can always claim it's `sleep(1)`-format. :innocent: It's trivial to just do `"%ds"` from seconds anyway like you said. And the Ruby SDK could still accept `1.week` and turn that into seconds and then into `"%ds"`.

engine/version.go
sipsma: bumped it, should go into v0.19.4!

dagql/types.go
sipsma: Done

dagql/objects.go
```diff
@@ -382,6 +386,15 @@ func (r Result[T]) ResultWithPostCall(fn cache.PostCallFunc) Result[T] {
 	return r
 }
 
+func (r Result[T]) WithSafeToPersistCache(safe bool) AnyResult {
```
vito, line 389: Neat how dynamic this is 👍

vito (approved):
:shipit: :shipit: :shipit: :shipit:

shykes:
!!!!!!!!!!!!!! 🥳

PR #10976, opened 2025-09-03
Python package updates

helderco:
General package updates, and fix for gql v4 (follow-up to https://github.com/dagger/dagger/pull/10888).

Main gql related changes:
- httpx errors are now captured under `gql.TransportConnectionFailed`
- Removed `dagger.ExecuteTimeoutError` because I don't see the need for its own exception
- `dsql_gql` now returns a `gql.GraphQLRequest` instead of `graphql.DocumentNode` but affected code is considered private/internal

Other notes:
- I wanted to rename `dagger.TransportError` to `dagger.ClientRequestError` but was concerned about a possible breaking change. Not sure if people catch it. I could deprecate the old name through an alias but decided not to bother for now.
- This supercedes https://github.com/dagger/dagger/pull/10942
-  It's best practice for a library to constrain on the minimum version and not upper. To avoid the breaking change from gql 4.0 in the future I'd rather do something else, like copy `uv.lock` from template rather than re-compute on every `dagger init`.


TomChv (approved):
Can this be a breaking change for users that may have installed the gql[httpx] in their module dep?

Otherwise changes LGTM, please create a changelog if this change affect users

jedevc (approved):

PR #10982, opened 2025-09-03
infra: setup alternative workflows to test cloud engines

matipan:
We want to test the performance and reliability of the new cloud engines. In the past we've always introduced changes to our CI using "alternative-*" workflows. This is so that we can easily compare the performance of the new CI changes against our current baseline. We are initially using 4 jobs: `testdev-everything-else`, `testdev-module-runtimes`, `go` and `php`. This were picked because, in the case of Engine & CLI, they are the slowest and least reliable jobs as can be seen [here](https://ui.honeycomb.io/dagger/environments/experimental/board/tXGtVxx7uVP/Is-Dagger-Dagger-production-ready):

image

We don't expect any performance improvements on the individual jobs themselves since its the same underlying infrastructure. The only difference is caching, we did not have any before and we do now. However, these jobs are not particularly good at caching so its hard to tell whether we will see any performance improvements.

For now we will keep these jobs running twice per day as we monitor and move forward on other relevant changes to our CI.

This PR needs a [dagger.io](https://github.com/dagger/dagger.io/pull/4603) change to go out before a workflow can run.

### DIFF
--- a/.github/workflows/alternative-ci-engines-1.yml
+++ b/.github/workflows/alternative-ci-engines-1.yml
@@ -1,0 +1,128 @@
+# NOTE: this is a temporary workflow created to test side-by-side the performance
+# of running our CI using our cloud engines. This is not a production workflow.
+# It will not be run on PRs and won't block releases in any way.
+# There's a few steps being done here that will not be taken over to the main CI
+# when/if we migrate to cloud engines.
+name: Alternative CI Engines 1
+
+on:
+  # Run the workflow every day TWICE:
+  # 1. 9:06AM UTC (low activity)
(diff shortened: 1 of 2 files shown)


TomChv (commented):
Can you explain in the PR description how this works and why that change?
Otherwise LGTM, happy to give an approval whenever you update the PR description

jedevc (approved):

jedevc:
LGTM once lint is passing!

matipan:
Small update: this is not yet ready for review. After today's discussion there's some changes I'm doing to the workflows. I'll update the title once the PR is ready.

PR #10983, opened 2025-09-03
chore: refactor directory.WithDirectory to dagop

alexcb:


core/integration/directory_test.go
```diff
@@ -288,6 +289,41 @@ func (DirectorySuite) TestWithDirectory(ctx context.Context, t *testctx.T) {
 	})
 }
 
+func (DirectorySuite) TestWithDirectoryUnion(ctx context.Context, t *testctx.T) {
```
sipsma, line 296: Is this covering stuff not covered in `func (DirectorySuite) TestDirectMerge`? Just checking if there's any consolidation possible, not a blocker at all if there's any trouble in consolidating

core/integration/directory_test.go
```diff
@@ -288,6 +289,41 @@ func (DirectorySuite) TestWithDirectory(ctx context.Context, t *testctx.T) {
...
+	c := connect(ctx, t)
+
+	dir1 := c.Container().From(alpineImage).
+		WithMountedDirectory("/working", dag.Directory()).
```
sipsma: nit: here and elsewhere should use `c.Directory()` instead of `dag.Directory()` so it's all going through the same client (even though it funnily happens to work in this case 😄)

core/schema/container.go
```diff
@@ -404,7 +404,7 @@ func (s *containerSchema) Install(srv *dagql.Server) {
 					`If the group is omitted, it defaults to the same as the user.`),
 			),
 
-		dagql.Func("withDirectory", s.withDirectory).
+		dagql.NodeFunc("withDirectory", DagOpContainerWrapper(srv, s.withDirectory)).
```
sipsma: `Container.WithDirectory` is implemented by just `dagql.Server.Select`'ing calls to `Directory.withDirectory`, so I don't actually think we need to dag-op-ify `Container.withDirectory`?

If there's any compelling reason to then it's totally fine, doesn't really hurt that much either to have it wrapped for now.

core/directory.go
```diff
@@ -704,36 +707,150 @@ func (dir *Directory) WithDirectory(
...
+		// copy but preserves the other caching benefits of MergeOp. This is the same
+		// behavior as "COPY --link" in Dockerfiles.
+		copyingRef, err := query.BuildkitCache().New(ctx, nil, nil,
+			bkcache.CachePolicyRetain,
+			bkcache.WithRecordType(bkclient.UsageRecordTypeRegular),
+			bkcache.WithDescription("Directory.withDirectory source"))
+		if err != nil {
+			return nil, fmt.Errorf("failed to create new cache for source directory: %w", err)
+		}
+		err = MountRef(ctx, copyingRef, nil, func(copyDest string) error {
```
sipsma: This is a bit subtle but important: the benefit of this approach relies on the fact that the copy happens as its own independent + cache-able step. When we just inline the copying here we actually lose that.

What we should ideally do for this `!canDoDirectMerge` branch is actually make a `dagql.Server.Select` call to create this "rebased" ref and then plug the result of that into the Merge. Specifically something along the lines of:
```go
srv.Select(ctx, srv.Root(), &rebasedDir, 
  dagql.Selector{Field: "directory"}, // scratch
  dagql.Selector{Field: "withDirectory", args: []dagql.NamedInput{
    {Name: "path", Value: dagql.String(destDir)},
    {Name: "directory", Value: dagql.NewID[*Directory](src.ID())},
  }},
)
```
* will need to watch out for infinite recursion of course, but as long as the case of `dir` being scratch is handled separately, should work out.

That way the "rebasing" is it's own cached operation and can be re-used independently of just this particular `Directory.WithDirectory` call.

Lemme know if that's making sense, might be a very subtle distinction. If this ends up not being plausible for whatever reason, it's also not a hard blocker.

core/integration/directory_test.go
alexcb: The particular bug I ran into required a directory containing a subdir, where as `TestDirectMerge` appears to just be merging a directory of files at the top level.

e.g. The bug I initially introduced caused `data/a` and `data/b` to incorrectly use the last version of the `data` directory, which caused `data/a` to be lost.

I initially didn't hit the bug until the go sdk generation side of the build, which was a rather painful experience :sweat_smile: 

This test has the added bonus of only using Directory and no Containers -- at the time it helped me chasing the bug.

core/integration/directory_test.go
alexcb: good catch!

core/schema/container.go
alexcb: I suspect I was on auto-pilot and replaced this, I backed it out and it seems to be working.

alexcb:
running `./hack/with-dev /bin/sh -c ' cd /home/alex/dagger/dagger/dagql/idtui/viztest && dagger call -vvv use-cached-exec-service'`

shows:

```
✔ viztest: Viztest! 0.0s
▼ .cachedExecService: Service! 0.3s ◆ CPU Pressure (some): 215µs ◆ CPU Pressure (full): 198µs ◆ Memory Bytes (current): 655 kB ◆ Memory Bytes (peak): 17 MB ◆ Network Rx: 90 B ◆ Network Tx: 176 B
├╴✔ container: Container! 0.0s
├╴▼ .from(address: "busybox"): Container! 0.1s SKIPPED
│ ├╴▼ resolving docker.io/library/busybox:latest 0.1s
│ │ ╰╴▼ remotes.docker.resolver.HTTPRequest 0.1s
│ │   ╰╴✔ HTTP HEAD 0.1s
│ │     
│ ╰╴▼ Container.from(address: "docker.io/library/busybox:latest@sha256:d82f458899c9696cb26a7c02d5568f81c8c8223f8661bb2a7988b269c8b9051e"): Container! 0.0s SKIPPED
│   ╰╴✔ resolving docker.io/library/busybox:latest@sha256:d82f458899c9696cb26a7c02d5568f81c8c8223f8661bb2a7988b269c8b9051e 0.0s
├╴∅ .withExec(args: ["echo", "exec-service cached for good"]): Container! 0.0s SKIPPED
├╴∅ .withExec(args: ["echo", "im also cached for good"]): Container! 0.0s SKIPPED
├╴∅ .withExec(args: ["echo", "im a buster", "2025-10-02 00:58:09.330096568 +0000 UTC m=+0.012770428"]): Container! 0.0s SKIPPED
├╴∅ .withExec(args: ["sleep", "1"]): Container! 0.0s SKIPPED
├╴∅ .withExec(args: ["echo", "im busted by that buster"]): Container! 0.0s SKIPPED
├╴▼ .withNewFile(path: "/srv/index.html", contents: "<h1>hello, world!</h1>"): Container! 0.0s SKIPPED
│ ├╴$ file(name: "index.html", contents: "<h1>hello, world!</h1>", permissions: 420): File! 0.0s CACHED
│ │ 
│ ╰╴∅ Missing.withFile(
│     ┆ path: "/srv/index.html"
│     ┆ source: file(name: "index.html", contents: "<h1>hello, world!</h1>", permissions: 420): File!
│     ): Directory! 0.0s SKIPPED
├╴✔ .withExposedPort(port: 80): Container! 0.0s
╰╴✔ .asService(args: ["httpd", "-v", "-h", "/srv", "-f"]): Service! 0.0s
```

in particular this line:

```
├╴∅ .withExec(args: ["echo", "im a buster", "2025-10-02 00:58:09.330096568 +0000 UTC m=+0.012770428"]): Container! 0.0s SKIPPED
```

Which seems a bit confusing since one would expect the `"echo", "im a buster", ....` to always be run.

So maybe we need to re-evaluate if introducing this "SKIPPED" state is a good idea or not.

-----

update: definitely a bad idea -- ultimately there was a bug found where the directory arg inputs were not correctly being fed into the `core.NewDirectoryDagOp(..., deps)` call.

core/directory.go
alexcb: I think it makes sense. To double check is it related to forcing the call to go through `GetOrInitializeWithCallbacks`? via https://github.com/alexcb/dagger/blob/eaec2ec943fa9162c19269f176cc23e2e1330256/dagql/objects.go#L684

Either way, I re-worked it as you suggested and I think it's working -- any thoughts on if we want a test for this (and how we could do it)?

core/schema/wrapper.go
```diff
@@ -237,11 +244,15 @@ func DagOpDirectory[T dagql.Typed, A any](
 	if err != nil {
 		return nil, err
 	}
-
 	argDigest, err := core.DigestOf(args)
 	if err != nil {
 		return nil, err
 	}
+	argDeps, err := core.InputsOf(ctx, args)
```
sipsma, line 256: nit: to save some potential future confusion, can we plop this into the other dag-ops (container + raw)? Even if the arg types don't have `Inputs` yet, if we do add those in the future would be nice for it to just work right away

core/directory.go
sipsma, line 1: We can rm `mergeStates` now right? seems like it's unused after this

core/directory.go
```diff
@@ -737,6 +745,184 @@ func (dir *Directory) WithDirectory(
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get directory ref: %w", err)
+	}
+	srcRef, err := getRefOrEvaluate(ctx, src)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get source directory ref: %w", err)
+	}
+
+	canDoDirectMerge :=
+		filter.IsEmpty() &&
```
sipsma, line 785: think this is missing the previous logic that also discluded copies that change the mode or owner (https://github.com/sipsma/dagger/blob/aaef136ca05fdb441c5624cdc6ceb1ba182b2d0c/core/directory.go#L949-L950)

core/directory.go
```diff
@@ -737,6 +745,184 @@ func (dir *Directory) WithDirectory(
...
+			destDir == "/" &&
+			src.Dir == "/"
+
+	cache := query.BuildkitCache()
+
+	if dirRef == nil {
```
sipsma, line 792: nit: comment explaining what this branch is ("destination is empty dir, so just do a plain copy", right?) would be helpful

sipsma (approved):
couple comments but LGTM after that!

core/directory.go
sipsma: Honestly testing here is extremely tricky... The normal tricks of reading /dev/random from execs won't really suffice. I think I'm ok with leaving it as is for now rather than dragging out further, we do at least have tests for the correctness even if not for the optimization

core/directory.go
alexcb: unfortunately not yet -- I had to create a legacy `WithDirectoryLLB` func, which is used by `WithChanges`. I recall attempting to convert `WithChanges` at the same time, but was getting a lot of errors and figured I could tackle converting it in a follow up PR.

core/directory.go
alexcb: good catch on the owner.

It doesn't look like there's any `mode` option though: https://github.com/sipsma/dagger/blob/aaef136ca05fdb441c5624cdc6ceb1ba182b2d0c/core/schema/directory.go#L136-L145

PR #10987, opened 2025-09-04
chore: remove explicit loop captures

jedevc:
The need for this was removed by go 1.22: https://go.dev/doc/go1.22#language


jedevc:
Hmmm, wondering if we need to do something that ensures that we no longer support modules that have a `go.mod` with a go version <1.22 (since that doesn't have this fix).

I think we can do that somewhere.

jedevc:
Backed out of the changes in `sdk/go`, I don't want to make this a breaking change.

jedevc:
Got a prior approval, and have restored to that state *minus* any potentially breaking changes in `sdk/go`.

PR #10988, opened 2025-09-04
chore: make provision tests use latest uv

jedevc:


helderco:
Still an issue with the Go SDK in macOS? But ok to solve in a diff PR.

helderco:
Looks like a flake (timeout).

PR #10991, opened 2025-09-05
feat: Add pie and docker-php-ext-install config hooks on dagger.json for PHP sdk

Neirda24:
Hi.

In this PR I add :
* `sdk.config.phpExtInstall` to install additional extension for the module (such as pcntl for example)
* `sdk.config.usePie` to enable or not `php/pie`
* `sdk.config.phpVersion` to use a different php version per module
* Use the official composer image (official not in the docker registry meaning) which is smaller and built for COPYing the binary.
* fixed an issue with `func(string $var = null)` where nullability of `$var` was implicit and this is deprecated (was causing issues when trying with php 8.5)
* update php template to update php versions in composer.json

I have a few TODOs left because I need your help on some subjects.

I know this may be a too much for a single PR so let me know if and how we can split things up but it seemed easier to put it all in for discussion.

### DIFF
--- a/sdk/php/runtime/main.go
+++ b/sdk/php/runtime/main.go
@@ -12,22 +12,32 @@ import (
 )
 
 const (
-	PhpImage      = "php:8.3-cli-alpine"
-	PhpDigest     = "sha256:e4ffe0a17a6814009b5f0713a5444634a9c5b688ee34b8399e7d4f2db312c3b4"
-	ComposerImage = "composer:2@sha256:6d2b5386580c3ba67399c6ccfb50873146d68fcd7c31549f8802781559bed709"
-	ModSourcePath = "/src"
-	GenPath       = "sdk"
+	PhpImage       = "php:8.4-cli-alpine" // TODO: because of how defaults works, this is not used. Should I switch to `nil` by default and fallback to using those constants ?
+	PhpDigest      = "sha256:fada271bbcae269b0b5f93212432a70ffb0aa51de0fa9c925455e8a1afae65ca"
+	CurlImage      = "alpine/curl:8.14.1"
+	CurlDigest     = "sha256:4007cdf991c197c3412b5af737a916a894809273570b0c2bb93d295342fc23a2"
+	ComposerImage  = "composer/composer:2-bin@sha256:73bf0499280eef8014f37daf5c4fb503f7964d9e5d53d447ff26d01d8a7e5d23"
+	PieReleasePhar = "https://github.com/php/pie/releases/download/1.2.0/pie.phar"
+	PieDigest      = "sha256:5ea836df7244a05d62b300a2294b5b6ae10c951f4f6a5e0d2ae2de84541142f0" // TODO: not sure how to check this after curl
+	ModSourcePath  = "/src"
+	GenPath        = "sdk"
 )
 
 type PhpSdk struct {
 	SourceDir *dagger.Directory
+	// Set custom php version to use with this module
+	PhpVersion string
+	// If true will use pie to install extra php modules from composer.json
+	UsePie bool
+	// List of extension to install through `docker-php-ext-install`
+	PhpExtInstall []string
 }
 
 func New(
-	// Directory with the PHP SDK source code.
-	// +optional
-	// +defaultPath="/sdk/php"
-	// +ignore=["**", "!generated/", "!src/", "!scripts/", "!composer.json", "!composer.lock", "!LICENSE", "!README.md"]
+// Directory with the PHP SDK source code.
+// +optional
+// +defaultPath="/sdk/php"
+// +ignore=["**", "!generated/", "!src/", "!scripts/", "!composer.json", "!composer.lock", "!LICENSE", "!README.md"]
 	sdkSourceDir *dagger.Directory,
 ) (*PhpSdk, error) {
 	if sdkSourceDir == nil {
@@ -36,6 +46,20 @@ func New(
 	return &PhpSdk{
 		SourceDir: sdkSourceDir,
 	}, nil
+}
+
+func (m *PhpSdk) WithConfig(
+// +default="8.4-cli-alpine@sha256:fada271bbcae269b0b5f93212432a70ffb0aa51de0fa9c925455e8a1afae65ca"
+	phpVersion string,
+// +default=false
+	usePie bool,
+// +default=[]string{}
+	phpExtInstall []string,
+) *PhpSdk {
+	m.PhpVersion = phpVersion
+	m.UsePie = usePie
+	m.PhpExtInstall = phpExtInstall
+	return m
 }
 
 func (m *PhpSdk) Codegen(
@@ -72,13 +96,30 @@ func (m *PhpSdk) CodegenBase(
 	}
 
 	base := dag.Container().
-		From(fmt.Sprintf("%s@%s", PhpImage, PhpDigest)).
-		WithExec([]string{"apk", "add", "git", "openssh", "curl"}).
-		WithFile("/usr/bin/composer", dag.Container().From(ComposerImage).File("/usr/bin/composer")).
-		WithMountedCache("/root/.composer", dag.CacheVolume(fmt.Sprintf("composer-%s", PhpImage))).
+		From(fmt.Sprintf("php:%s", m.PhpVersion)).
+		WithExec([]string{"apk", "add", "git", "openssh", "curl", "jq"}).
+		WithFile("/usr/bin/composer", dag.Container().From(ComposerImage).File("/composer")).
+		WithMountedCache("/root/.composer", dag.CacheVolume(fmt.Sprintf("composer-%s", m.PhpVersion))).
 		WithEnvVariable("COMPOSER_HOME", "/root/.composer").
 		WithEnvVariable("COMPOSER_NO_INTERACTION", "1").
 		WithEnvVariable("COMPOSER_ALLOW_SUPERUSER", "1")
+
+	if len(m.PhpExtInstall) > 0 {
+		for _, ext := range m.PhpExtInstall {
+			base = base.
+				WithExec([]string{"docker-php-ext-install", ext})
+		}
+	}
+
+	if m.UsePie {
+		base = base.
+			WithExec([]string{"apk", "add", "gcc", "make", "autoconf", "libtool", "bison", "re2c", "pkgconf"}). // For php/pie
+			WithFile("/usr/bin/pie",
+				dag.Container().From(fmt.Sprintf("%s@%s", CurlImage, CurlDigest)).
+					WithExec([]string{"curl", "-o", "/usr/bin/pie", "-sLO", PieReleasePhar}).
+					WithExec([]string{"chmod", "+x", "/usr/bin/pie"}).
+					File("/usr/bin/pie"))
+	}
 
 	/**
 	 * Mounts PHP SDK code and installs it
@@ -143,6 +184,13 @@ func (m *PhpSdk) CodegenBase(
 		ctr = ctr.WithExec([]string{
 			"composer",
 			"install"})
+	}
+
+	if m.UsePie {
+		ctr = ctr.WithExec([]string{
+			"pie",
+			"install",
+			"--no-interaction"})
 	}
 
 	return ctr, nil
--- a/sdk/php/runtime/scripts/init-template.sh
+++ b/sdk/php/runtime/scripts/init-template.sh
@@ -9,7 +9,20 @@ if ! [ -f composer.json ]; then
     sed -i "s/Example/$CLASS/g" ./src/Example.php
     mv ./src/Example.php ./src/$CLASS.php
 
+    PHP_VERSION=${PHP_VERSION:?}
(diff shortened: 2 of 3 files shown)


sdk/php/runtime/main.go
```diff
@@ -72,14 +96,31 @@ func (m *PhpSdk) CodegenBase(
...
+		WithFile("/usr/bin/composer", dag.Container().From(ComposerImage).File("/composer")).
+		WithMountedCache("/root/.composer", dag.CacheVolume(fmt.Sprintf("composer-%s", m.PhpVersion))).
 		WithEnvVariable("COMPOSER_HOME", "/root/.composer").
 		WithEnvVariable("COMPOSER_NO_INTERACTION", "1").
 		WithEnvVariable("COMPOSER_ALLOW_SUPERUSER", "1")
 
+	if len(m.PhpExtInstall) > 0 {
+		for _, ext := range m.PhpExtInstall {
+			base = base.
+				WithExec([]string{"docker-php-ext-install", ext})
```
charjr, line 110: The `docker-php-ext-install` script cannot be automated easily:

> Remember, you must install dependencies for your extensions manually. If an extension needs custom configure arguments, you can use the docker-php-ext-configure script

https://hub.docker.com/_/php#php-core-extensions

Neirda24: yes but then I tought we would need to have full sh script as extension point ?

charjr:
>     * `sdk.config.phpVersion` to use a different php version per module 
>     * update php template to update php versions in composer.json

This might be nice _one day_, but currently [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). Reason being, the PHP SDK uses attributes, which were [released in version 8.0](https://php.watch/versions/8.0/attributes). PHP is still in 8.X. Making it impossible to change major versions, without a massive overhaul of the entire SDK. 

Minor versions do not break backwards compatibility, so simply updating the image to use the latest (_stable_) version, (as you have done in this PR) supports every version we possibly can and is a far simpler change to make.

sdk/php/runtime/main.go
```diff
@@ -12,22 +12,32 @@ import (
 )
 
 const (
-	PhpImage      = "php:8.3-cli-alpine"
-	PhpDigest     = "sha256:e4ffe0a17a6814009b5f0713a5444634a9c5b688ee34b8399e7d4f2db312c3b4"
-	ComposerImage = "composer:2@sha256:6d2b5386580c3ba67399c6ccfb50873146d68fcd7c31549f8802781559bed709"
-	ModSourcePath = "/src"
-	GenPath       = "sdk"
+	PhpImage       = "php:8.4-cli-alpine" // TODO: because of how defaults works, this is not used. Should I switch to `nil` by default and fallback to using those constants ?
```
charjr, line 15: Simply updating from 8.3 to 8.4 would be a nice change, I think this could be a separate, small PR

Neirda24:
> > ```
> > * `sdk.config.phpVersion` to use a different php version per module 
> > * update php template to update php versions in composer.json
> > ```
> 
> This might be nice _one day_, but currently [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). Reason being, the PHP SDK uses attributes, which were [released in version 8.0](https://php.watch/versions/8.0/attributes). PHP is still in 8.X. Making it impossible to change major versions, without a massive overhaul of the entire SDK.
> 
> Minor versions do not break backwards compatibility, so simply updating the image to use the latest (_stable_) version, (as you have done in this PR) supports every version we possibly can and is a far simpler change to make.

sometimes there are breaking changes and even without, it brings deprecations which can cause issues in performance or when parsing some output (see pie).

If we don't allow this what would be the timeline to update to newer php versions to benefit from new features / bugfix ?

charjr:
>     * `sdk.config.usePie` to enable or not `php/pie`

If we abstract the process of installing extensions, automate it away from our users... will they notice if it's PIE or not?

If there isn't a difference, and it's all _under the hood_ then a single, robust method is less work, less maintenance, and easier to debug when users face problems.

charjr:
>     * Use the official composer image (official not in the docker registry meaning) which is smaller and built for COPYing the binary.

This shaves 75.6MB off of the downloads! Since the _official_ and _community_ versions are [both maintained by the same person](https://github.com/composer/docker), I'm happy for us to make the switch. :smile: 

If we can get a separate PR for this, that would be brilliant.

charjr:
>     * fixed an issue with `func(string $var = null)` where nullability of `$var` was implicit and this is deprecated (was causing issues when trying with php 8.5)

This is a great spot! In my opinion that's an outright bug. Please make a separate PR for this change, that can go in instantly.

charjr:
> sometimes there are breaking changes and even without, it brings deprecations which can cause issues in performance or when parsing some output (see pie).

The issue is, allowing multiple PHP versions is a [maintenance burden](https://discord.com/channels/707636530424053791/1162025276872609832/1404742042801930342). Suddenly every contribution the PHP SDK must work on the earliest supported version AND not be deprecated or removed in the latest supported version.

For instance, I can simplify the SDK code with some of 8.4's new features (array_find, array_first, etc.). But if we allow multiple PHP versions, I cannot. That is a problem for the maintainers.

The tradeoff needs to be worth it, It needs to solve a real problem that users are facing. What problem does this solve for the user? 

The PHP SDK itself doesn't read your PHP code. The PHP SDK reads your Dagger module, that is all. If the Dagger module used PHP 8.5, as long as your module is written in PHP 8.5, it doesn't matter what your code is.

You can take a legacy project, written in PHP 5.0, have your module spin up a php:5.0 container and run your code without issue. 

> If we don't allow this what would be the timeline to update to newer php versions to benefit from new features / bugfix ?

For minor versions we should be fine as soon as they're stable. :+1: 
I will admit, we could have updated to 8.4 sooner, my bad, but thank you for spotting it.

Neirda24:
> > sometimes there are breaking changes and even without, it brings deprecations which can cause issues in performance or when parsing some output (see pie).
> 
> The issue is, allowing multiple PHP versions is a [maintenance burden](https://discord.com/channels/707636530424053791/1162025276872609832/1404742042801930342). Suddenly every contribution the PHP SDK must work on the earliest supported version AND not be deprecated or removed in the latest supported version.
> 
> For instance, I can simplify the SDK code with some of 8.4's new features (array_find, array_first, etc.). But if we allow multiple PHP versions, I cannot. That is a problem for the maintainers.
> 
> The tradeoff needs to be worth it, It needs to solve a real problem that users are facing. What problem does this solve for the user?
> 
> The PHP SDK itself doesn't read your PHP code. The PHP SDK reads your Dagger module, that is all. If the Dagger module used PHP 8.5, as long as your module is written in PHP 8.5, it doesn't matter what your code is.
> 
> You can take a legacy project, written in PHP 5.0, have your module spin up a php:5.0 container and run your code without issue.
> 
> > If we don't allow this what would be the timeline to update to newer php versions to benefit from new features / bugfix ?
> 
> For minor versions we should be fine as soon as they're stable. 👍 I will admit, we could have updated to 8.4 sooner, my bad, but thank you for spotting it.

True... So both modules and the SDK are stuck on the same version... What I am still annoyed by is the fact that module contributors won't get notified of changes / breaking changes. So at some point the module will stop working. Also some extensions will be working on some specific php version (issues pie is trying to address) so even if we allow module author to provide shell script to be ran where they can install / configure whatever they want, it means that we are stuck with major AND minor and we still don't have a way to provide notifications / deprecations and/or allow multiple version for backward compatibility...

The codegen itself can easily be ran in its own container. The only compatibility burden I see are the files generated which are quite simple because they are a simple mapping to the graphql query. am I right ?

Neirda24:
> > ```
> > * fixed an issue with `func(string $var = null)` where nullability of `$var` was implicit and this is deprecated (was causing issues when trying with php 8.5)
> > ```
> 
> This is a great spot! In my opinion that's an outright bug. Please make a separate PR for this change, that can go in instantly.

on this note I spotted something else. in `generated/Directory::search()` method. `$pattern` is a required parameter but it is declared after two optional ones `$paths` and `$globs` which is an issue. But I didn't get yet how to change this. any idea ?

Neirda24:
> > ```
> > * `sdk.config.usePie` to enable or not `php/pie`
> > ```
> 
> If we abstract the process of installing extensions, automate it away from our users... will they notice if it's PIE or not?
> 
> If there isn't a difference, and it's all _under the hood_ then a single, robust method is less work, less maintenance, and easier to debug when users face problems.

Yes they will notice. because if you target pie. then you add extensions almost like package `<company>/<package>`. IMO we can support it out of the box it is a no brainer.

charjr:
> True... So both modules and the SDK are stuck on the same version... 

I think this is best for maintainability

> What I am still annoyed by is the fact that module contributors won't get notified of changes / breaking changes. 

Changes especially breaking ones, will be in the release notes, e.g. [0.18.7](https://github.com/dagger/dagger/releases/tag/v0.18.17) or [0.18.7 Typescript/SDK](https://github.com/dagger/dagger/releases/tag/sdk%2Ftypescript%2Fv0.18.17)

The reason the PHP SDK release notes never have any breaking changes listed is that so far there have not been any breaking changes.

> So at some point the module will stop working. 

That point is in the user's control, it is when the user decides to upgrade their module to a later version. So this is exactly the same as using any other library following [semver](https://semver.org/). 

Also some extensions will be working on some specific php version (issues pie is trying to address) 

> so even if we allow module author to provide shell script to be ran where they can install / configure whatever they want

The module itself does not need customizable extensions, it shouldn't be doing the heavy lifting. 

For example, if you need a database: _your module is not the database_, your module sets up a container with the database inside. Your module does not need any extensions installed, your module needs to install extensions in the container.

```php
        return dag()
            ->container()
            ->from('php:8.4-cli-alpine')
            ->withFile('/usr/bin/install-php-extensions', dag()
                ->container()
                ->from('mlocati/php-extension-installer')
                ->file('/usr/bin/install-php-extensions'))
            ->withExec(['install-php-extensions', $extension]);
```

That would install a PHP extension for the user, in a container they can use, without writing anything but PHP.

As it stands, I haven't encountered any situation that needed additional extensions installed in the module. If there are real use cases for flexible installation of extensions into the module itself, then we should absolutely discuss them and see if we're missing a critical feature. 


> it means that we are stuck with major AND minor and we still don't have a way to provide notifications / deprecations and/or allow multiple version for backward compatibility...

In terms of notifications, see above.

We will upgrade the SDK to use 8.4, when we do we will need to include a release fragment using [changie](https://changie.dev/) which will automatically become part of the release notes on the next tagged release.

When this happens, it is up to the user if they upgrade. It ends up being no different to wanting to upgrade to [PHPUnit 12.X](https://github.com/sebastianbergmann/phpunit/releases/tag/12.0.0) which dropped support for PHP 8.2

That said, the Dagger PHP SDK is not as strict as PHPUnit it can handle 8.0, 8.1, 8.2 and 8.3. Any change that breaks that should always come with a note for the changelog. 

> The codegen itself can easily be ran in its own container. The only compatibility burden I see are the files generated which are quite simple because they are a simple mapping to the graphql query. am I right ?

Correct.

The module runs in its own container, specifically, this one:
https://github.com/dagger/dagger/blob/b9a64f6bf2dff6e9c851b9b5a73d569c163f8fa2/sdk/php/runtime/main.go#L15

The container generates its own version compatible `codegen` directory. Your local version of `codegen` _never_ goes into the container, it is there for your own use, so you can get autocomplete, static analysis, etc.

charjr:
> on this note I spotted something else. in `generated/Directory::search()` method. `$pattern` is a required parameter but it is declared after two optional ones `$paths` and `$globs` which is an issue. But I didn't get yet how to change this. any idea ?

[This is worth a separate issue](https://github.com/dagger/dagger/issues/11009), I'm not entirely sure myself but my guess is it's an issue in the underlying library.

charjr:
> Yes they will notice. because if you target pie. then you add extensions almost like package `<company>/<package>`. IMO we can support it out of the box it is a no brainer.

Looking at the docs, it doesn't look like this is automatable:
> If the extension you would like to install needs additional libraries or other
dependencies, then these must be installed beforehand too.

https://github.com/php/pie/blob/4f5fa2e3535bc3ee584df2446c9da3d6b76aff98/docs/usage.md?plain=1#L70-L71

I'm also still not sure what problem this is solving:
> The module itself does not need customizable extensions, it shouldn't be doing the heavy lifting.
>
> For example, if you need a database: your module is not the database, your module sets up a container with the database inside. Your module does not need any extensions installed, your module needs to install extensions in the container.

charjr:
I'm going to close this for now as we have merged:
- upgrading the image to PHP 8.4
- replacing the composer image with a smaller binary only image
- Fixing the deprecation in existing code

If we want support for custom images, I would opt for doing it similarly to Python:
https://github.com/dagger/dagger/blob/15de1b6b40e4c84393ffc3f16e91fc1b3d4a3408/sdk/python/runtime/discovery.go#L442-L444

In this way we would allow customization, but maintainers can focus on high quality for a single _standard_ image.

Lastly adding PIE to the SDK wont give people access to PIE inside their modules, so it would have limited use. We would need to see real unresolvable issues before it became worth it IMHO.

For anyone wanting PIE in their modules, getting the phar is a one liner:

```php
$pie = dag()->container()->from('ghcr.io/php/pie:bin')->file('/pie');
```

PR #10995, opened 2025-09-05
feat: LocalSource dagOp migration (WIP)

TomChv:


core/schema/host.go
```diff
@@ -370,26 +343,26 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
...
+			Exclude: args.Exclude,
+		}, args.Gitignore, args.NoCache)
 
-	localOpts = append(localOpts, llb.WithCustomName(localName))
+		if err != nil {
+			return inst, fmt.Errorf("failed to get directory: %w", err)
+		}
 
-	localLLB := llb.Local(hostPath, localOpts...)
+		dir, err = dir.Directory(ctx, relPath)
```
TomChv: This part is still broken because `.Directory` uses `bk.Solves` to get the children path.
But in our case we only have an immutable ref.
How can I solve that problem?

TomChv:
⚠️ Current progress on `TestHost/TestDirectoryGitIgnore/don't_load_files_ignored_by_.gitignore`

## The bug

The test `TestHost/TestDirectoryGitIgnore/don't_load_files_ignored_by_.gitignore` make sure that
if you try to do `Host.Directory` on an ignored file or dir, you get the error: `no such file or directory` instead of
an empty result.

That test is failing in the CI and can also be reproduced locally with a specific setup:

I noticed that if I run `dagger --cloud -c 'test | specific --race --parallel=16 --run=TestHost'`
I'm expecting an error on the test but I got no error, instead its empty dir: https://dagger.cloud/Quartz/traces/d5c16a2ab29185945f58a4a5e13fd02a?listen=45217c5ec267e79f&listen=687183f977a844da&listen=246cfc105899c4d7

```
        Test:           TestHost/TestDirectoryGitIgnore/don't_load_files_ignored_by_.gitignore                                                       
        Messages:       expected error, got: []string{} (root entries: []string{".git/", ".gitignore", "foo/"})  
```

Note: we see in the root entries that `bar/` doesn't exist and is never copied so **why we are not throwing an error??**

However if If I run `dagger --cloud -c 'test | specific --race --parallel=16 --run=TestHost/TestDirectoryGitIgnore'`
Then the test succeed, I get the expected behaviour: https://dagger.cloud/Quartz/traces/f6342f7347de9ee0769787f2936e7e7e

**So why the test would succeed if I just run part of the suite but fail when I run the full suite?**

## Deep dive summary

I tried to check the content local and remote to see if somehow the directory exist inside the diff [but it doesn't]( https://dagger.cloud/Quartz/traces/ebf10f08f194a0efe575a2eacfc07ea5?listen=80bbafabc9964e28&listen=bc2c82afc76405b9&listen=781325b684cc944c&listen=d9f6a783e2e38dc5&listen=57f3b24010b0bc8d):

```
copyPath= only=map[string]struct {}{".git":struct {}{}, ".gitignore":struct {}{}, "foo":struct {}{}, "foo/foo.txt":struct {}{}}
copyPath=bar only=map[string]struct {}{}
```

We see that the root map `Host.Directory(".")` doesn't contain bar, so it shouldn't exist anywhere.
To confirm it, we can also see that `Host.Directory("bar")` returns an empty `map`.

The error is normally thrown by: https://github.com/TomChv/dagger/blob/aac284ad23e2867743c19f146cfad0761ff806e4/engine/sources/local/copy/copy.go#L128

But for some reason, that operation succeed if I run the full test suite... 

My guess is that think it's a consequence of the changes on Snapshot, now that I return the expected directory directly, we don't have that `.Directory` that triggers the error.
For example, if snapshot returns an empty dir but then I do `.Directory("./bar")` this would trigger the error, but we don't do that anymore so we need to catch that error in another way...

I'm not sure what would be the fix or the next step to understand what's wrong there.

TomChv:
Ok after more and more tests I finally understood the issue and how to fix it.

In our tests, we first load the whole directory so we got a cache ref on it, in that case the reference to `bar/` doesn't exist, so it's an empty dir.
Then we try to load `/bar`, but when we check the changes, there's nothing so we simply end up return an empty directory ref... If it's not cache, things will fail because thanks to the later `copy` but if things are somehow cached then it's empty.

To fix that, I created a variable to keep track of either we found or not the subdirectory when we explored the changes.
By doing so, we can early return an error even before checking the cache.
I originally wanted to rely on `only` and throw an error if only is empty but if the non-existing directory that you're trying to load has an existing parent, the only map will not be empty and that will lead to an empty directory return.


**Example for `Host.Directory("dagger/dag")`**

.gitignore
```
dagger/dag
```

result:
```
[] # It should throw an error instead because `dagger` is copied but `dagger/dag` doesn't exist in it
```

logs:
```
[SYNC] only=map[string]struct {}{"dagger":struct {}{}} relPathFound=false
time="2025-09-29T15:16:58Z" level=debug msg="GetContentHashKey reusing ref 1t9srefzmr4onr0t9j9sd5g88 with digest sha256:0e5db88383bce812f795689f5318a2b2f4fde740ef31c2a2365a46368aafddd2" client_hostname=Toms-MacBook-Pro.local client_id=k8xngomnmlhwz9leywun03oq6 session_id=p30umh6d8d3cp9xlgsxuj9cep spanID=e484abe8de3eebbb traceID=0218d1f90b2aad7548cd09fb18c3e9a0
```

As we see `relPathFound` is false so we should throw an error (disabled here to show the example) but we  see that `only` has `dagger`. (normal since it exist but here we copy `dagger/dag`).

We see that the cache hit on that and that leads to an empty directory instead of an error.

engine/sources/local/source.go
```diff
@@ -33,200 +27,75 @@ import (
 	"github.com/dagger/dagger/engine/client/pathutil"
 )
 
-type Opt struct {
-	CacheAccessor bkcache.Accessor
+type LocalSource struct {
```
jedevc: Ideally we should lift this out of `sources/`, since those are our special buildkit sources - since this isn't a proper buildkit source anymore, we can move it. A new package in `engine/buildkit` might work?

engine/sources/local/source.go
```diff
@@ -33,200 +27,75 @@ import (
...
-			id.SharedKeyHint = v
-		case pb.AttrLocalDiffer:
-			switch v {
-			case pb.AttrLocalDifferMetadata, "":
-				id.Differ = fsutil.DiffMetadata
-			case pb.AttrLocalDifferNone:
-				id.Differ = fsutil.DiffNone
-			}
-		}
+// TODO: Write a dummy local source to verify that we never use llb.Local anymore
```
jedevc: :eyes:

engine/sources/local/source.go
```diff
@@ -33,200 +27,75 @@ import (
...
-		ExcludePatterns: ls.src.ExcludePatterns,
-		Gitignore:       ls.gitIgnore,
-		FollowPaths:     ls.src.FollowPaths,
-		Cachebuster:     ls.cachebuster,
-	})
+func (ls *LocalSource) Snapshot(ctx context.Context, session session.Group, sm *session.Manager, clientPath string, opts SnapshotSyncOpts) (bkcache.ImmutableRef, error) {
+	clientMetadata, err := engine.ClientMetadataFromContext(ctx)
 	if err != nil {
-		return "", "", nil, false, err
+		return nil, fmt.Errorf("failed to get requester session ID: %w", err)
```
jedevc: `return nil, err` is enough in scenarios like this, we don't need to keep wrapping the error if it doesn't add more info.

engine/sources/local/source.go
```diff
@@ -33,200 +27,75 @@ import (
...
+	caller, err := sm.Get(timeoutCtx, clientMetadata.ClientID, false)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get session: %w", err)
 	}
 
-	ref, err := ls.snapshot(ctx, g, caller)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get snapshot: %w", err)
+	// If relPath is ".", we want to use the root path so we can unset it
+	// because `"."` leads to unnecessary complications inside snapshot.
```
jedevc: Does it? That's strange.

The only type of path cleaning I'd expect to see here would be with `filepath.Clean`.

engine/sources/local/source.go
```diff
@@ -456,6 +339,15 @@ func (ls *localSourceHandler) getRef(
 	}, nil
 }
 
+func curTracer(ctx context.Context) trace.Tracer {
+	return trace.SpanFromContext(ctx).TracerProvider().Tracer("dagger.io/filesync")
+}
```
jedevc: Probably make a separate `tracing.go`, and follow the same pattern as here: https://github.com/dagger/dagger/blob/718b5d7d7d6a4c41284b51d67e7f87940bee3fb1/core/tracing.go#L11-L13

engine/sources/local/source.go
```diff
@@ -456,6 +339,15 @@ func (ls *localSourceHandler) getRef(
...
+
+func newSpan(ctx context.Context, name string) (context.Context, trace.Span) {
+	tr := curTracer(ctx)
+	return tr.Start(ctx, name)
+}
```
jedevc: This is probably worth clearing up, I don't think this helper looks that useful.

core/schema/host.go
```diff
@@ -245,124 +242,107 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
...
-	hostPath, err = bk.AbsPath(ctx, hostPath)
-	if err != nil {
-		return inst, fmt.Errorf("failed to get absolute path from git ignore root %s: %w", hostPath, err)
-	}
-
-	relPath := "."
+	// The operation is not wrapped in dagOp at the schema level, so we manually handle
+	// it inside the resolver.
+	// That's because it's more convenient to call `MakeDirectoryContentHashed` here
+	// that having an option in the `dagOpDirectoryWrapper`
```
jedevc: As we discussed previously - we should add an arg to the wrapper to call `MakeDirectoryContentHashed` on the result - that should simplify the need to mess around with `args.InDagOp` here.

engine/sources/local/source.go
TomChv: `filepath.Clean(".")` returns `"."` and that makes `local.Sync` quite complex to do.

For example `strings.CutPrefix("bar", ".")` will return `"bar", false` while technically which will skip the cache change even though we should do it.
Same in `copier.copy`, if `srcComponents = "."`, that will change the logic etc...

So I have 2 choices:
- Set `relPath` to `""` if it's `"."`
- Do a bunch of `if relpath == "."` to adapt the logic in localfs & copy.

I took the choice of simplicity there, but if you see a reason to go the other way or have another idea, I'm happy to discuss about it :D

jedevc: Nope, that's a good explanation.

So either:
- update the comment to replace "unneccessary complications" with that detail
- or just remove "unneccessary complications" entirely, imo, that's fairly clear

core/schema/host.go
TomChv: There's no other function that `host.Directory` that actually calls `MakeDirectoryContentHashed` so I thought it would not be useful to add an arg in the wrapper for a single case.

Do you think it's worth going the general way for a single function?

engine/sources/local/source.go
TomChv: Happy to do that & refactor few things

core/schema/host.go
jedevc: It'll hugely simplify this function :) so worth it IMO.

Just because we're only using it one place now, doesn't mean we won't want to use it in more places in the future.

engine/filesync/copy/mkdir_windows.go
jedevc, line 1: We could simplify by removing the `windows` files and just consolidate entirely under one file.

This is all engine-side, we only support Linux. If we ever extend to support Windows, IMO, there's a lot of bigger issues.

@sipsma any reason to keep these around?

engine/filesync/localfs.go
```diff
@@ -320,9 +346,14 @@ func (local *localFS) Sync( //nolint:gocyclo
...
 	}
 
-	ctx, copySpan := newSpan(ctx, "copy")
+	ctx, copySpan := tracer(ctx).Start(ctx, "copy")
 	defer telemetry.End(copySpan, func() error { return rerr })
 
+	// If we didn't find any files/dir in the given relative path, we can early return an error.
+	if local.copyPath != "" && !relPathFound {
+		return nil, fmt.Errorf("%s: no such file or directory", local.copyPath)
+	}
```
jedevc, line 355: Is that because if `copyPath` was `foo/`, we'd expect to see at *least* `foo/`?

engine/server/server.go
```diff
@@ -443,13 +449,9 @@ func NewServer(ctx context.Context, opts *NewServerOpts) (*Server, error) {
...
 	archutil.WarnIfUnsupported(srv.enabledPlatforms)
 
-	ls, err := local.NewSource(local.Opt{
+	srv.workerFileSyncer = filesync.NewFileSyncer(filesync.FileSyncerOpt{
 		CacheAccessor: srv.workerCache,
 	})
-	if err != nil {
-		return nil, err
-	}
-	srv.workerSourceManager.Register(ls)
```
jedevc, line 452: Did you manage to look into a dummy local source here? So that we can error out if something tries to load an `llb.Local`?

This would go in `engine/sources/local`, just as a protection mechanism to effectively "disable" local sources, requiring all loading to be done through the `Host` API.

engine/filesync/tracing.go
```diff
@@ -0,0 +1,11 @@
+package filesync
+
+import (
+	"context"
+
+	"go.opentelemetry.io/otel/trace"
+)
+
+func tracer(ctx context.Context) trace.Tracer {
```
jedevc:
```suggestion
func Tracer(ctx context.Context) trace.Tracer {
```

For consistency.

engine/filesync/localfs.go
TomChv: Yeah! And in that case that means the directory exists but is empty :)

engine/server/server.go
TomChv: When I pushed the changes and triggerred the CI, nothing called it so I assumed it was same to simply remove it.
We got tests for `dockerbuild` etc but the test passes.

I can add it again to error out like you mentioned but is that really necessary?

engine/filesync/tracing.go
TomChv: Why making that method public if we never expect to use it?
I do not see a scenario where we call `filesync.Tracer`

engine/server/server.go
jedevc: So, I *think* that a custom frontend might be able to trigger it. Buildkit frontends (from using the `#syntax` directive) can execute arbitrary LLB. I'd rather be safe than sorry, and add a check.

engine/filesync/tracing.go
jedevc: Fair, all the other files use `Tracer` though.

We should be consistent whichever we pick, no strong preferences.

TomChv: Done in [`7e19608` (#10995)](https://github.com/dagger/dagger/pull/10995/commits/7e1960851e4ac370f59f079687b7fb1e16c4f45e)

engine/server/server.go
TomChv: Done in [`7e19608` (#10995)](https://github.com/dagger/dagger/pull/10995/commits/7e1960851e4ac370f59f079687b7fb1e16c4f45e)

core/schema/host.go
```diff
@@ -145,7 +143,11 @@ func (s *hostSchema) Install(srv *dagql.Server) {
 	}.Install(srv)
 
 	dagql.Fields[*core.Host]{
-		dagql.NodeFuncWithCacheKey("directory", s.directory, dagql.CacheAsRequested).
+		dagql.NodeFuncWithCacheKey("directory",
+			DagOpDirectoryWrapper(
+				srv, s.directory,
+				WithHashContentDir[*core.Host, hostDirectoryArgs](),
```
jedevc, line 149: I *think* you can omit the generic args here, they should be implied from the context?

core/schema/host.go
jedevc: This comment is not true anymore.

core/schema/host.go
TomChv: Screenshot 2025-10-20 at 15 04 22

PR #10999, opened 2025-09-06
modules/go: fix protoc version mismatch

shykes:
This fixes an error in the ./cmd/engine module, where generate() manually installs pinned go packages with a dependency on protoc, which conflict with the system version of protoc installed in the base image


TomChv:
I fixed the linter error, I'll merge that PR as soon as the CI is green

PR #11000, opened 2025-09-07
POC: load defaults from .env

shykes:
This is a POC for loading default configuration from `.env`, and injecting it into the dagger schema.

## How it works

- Dagger automatically searches for `.env` by findup.
- Each time a module is loaded, the `.env` is searched for variables that match the module name. For example, a module named `app` will match all variables starting with `app_` or `APP_`
- Matching variables are matched to corresponding constructor argument names

## Example

```console
cat > .env <<.
DISCORD_TOKEN=op://myvault/mydiscordtoken/credential
.
```

```console
dagger -m github.com/shykes/x/discord -c servers
```

--> This will list your discord servers, using the token in the `.env` file.

## See also

- Earlier POC with a different approach: https://github.com/dagger/dagger/pull/10697


shykes:
FYI @nipuna-perera @rawkode @nadirollo @vmaffet @ChrisChinchilla

This is closer to what @nipuna-perera suggested:

- Use standard `.env`
- Load it from the client's workdir, not the module
- Applies to multiple modules: the module name must be used as a prefix to the variable

It still works with all object types, not just secrets. (thanks to #10770).

nipuna-perera:
> FYI @nipuna-perera @rawkode @nadirollo @vmaffet @ChrisChinchilla
> 
> This is closer to what @nipuna-perera suggested:
> 
> * Use standard `.env`
> * Load it from the client's workdir, not the module
> * Applies to multiple modules: the module name must be used as a prefix to the variable
> 
> It still works with all object types, not just secrets. (thanks to #10770).

Love where this is headed. I _think_ I can live with just the constructor args being handled but it would be really amazing if it was any function arg. For example, I may have a `Java` module that has a `Maven` struct that may take a user/pass that wouldn't necessarily be needed for the `Java` module. If only constructor args were supported, I'd potentially have to split that into multiple modules. Is there a technical limitation for this being only for constructor args?

shykes:
No technical limitation, mostly UX: what would be the naming convention for all those different module and function arguments

nipuna-perera:
> No technical limitation, mostly UX: what would be the naming convention for all those different module and function arguments 

Typically the arg names don't collide. Even if they do, dagger can warn right? I'd say keep it simple and substitute any matching args.

shykes:
Final PR: #11034

PR #11010, opened 2025-09-08
make the snapshot context timeout configurable with an env var

abloch:


jedevc:
@abloch can you share context on what this is for? I'm not sure I understand.

abloch:
@jedevc I got a few timeouts when dagger was trying to access the registry.
this is the error I saw:
```
error:select: failed to get content hash: failed to get session: no active session for rug913txjapapu9y0g1h3vk9h: timeout: context deadline exceeded
```

I was assuming that increasing the timeout may solve this

jedevc:
This is actually an unrelated issue - the issue is not to do with snapshot timeouts, but to do with a session not existing when it should: see `failed to get session: no active session for`.

This *should* likely be resolved by https://github.com/dagger/dagger/pull/10761.

abloch:
thanks - that's a good point. will try.
nevertheless - I think hard-coding 5 seconds timeout is inferior to having this timeout configurable.

this number is arbitrary - it can sometimes make sense, on other (loaded or slow machines) it might not be enough

.github/workflows/publish.yml
```diff
@@ -18,7 +18,7 @@ on:
 
 jobs:
   publish:
-    if: ${{ github.repository == 'dagger/dagger' && github.event_name == 'push' }}
+    if: ${{ github.event_name == 'push' }}
```
TomChv, line 21: Why this change?

engine/sources/local/source.go
```diff
@@ -184,7 +186,18 @@ func (ls *localSourceHandler) Snapshot(ctx context.Context, g session.Group) (bk
 	}
 
 	timeoutCtx, cancel := context.WithCancelCause(ctx)
-	timeoutCtx, _ = context.WithTimeoutCause(timeoutCtx, 5*time.Second, fmt.Errorf("timeout: %w", context.DeadlineExceeded))
+	timeout := 5
+	envTimeoutStr := os.Getenv("SNAPSHOT_CONTEXT_TIMEOUT")
```
TomChv, line 190: I think if we want to configure the snapshot context timeout  at some point, this should be part of the [engine configuration](https://docs.dagger.io/reference/configuration/engine/) through `engine.[json|toml]` instead of being fetch from an environment variable.

However, I wouldn't recommend working on that changes until we know if that something we want to do or not.

TomChv (commented):
Left a quick review, I let @jedevc and @sipsma decide if that's a change we want to actually do or not.

jedevc:
Have you actually seen this particular thing time out before for a different reason? Extending the timeout is not actually likely to fix the session not found issue you're seeing.

I don't really want to add additional configuration dials if they don't actually solve the issue.

.github/workflows/publish.yml
abloch: you're right. I mistakenly forgot to delete the change I made for local builds.
will remove asap

abloch:
@jedevc @TomChv - I agree with your points. closing this PR until I find evidence that the timeout was indeed the issue.

thanks for taking the time to review it

PR #11011, opened 2025-09-08
fix: git ignore backward compatibility

TomChv:
As discussed with @jedevc, this PR ensure that module with version < `v0.18.18` are not automatically loading .gitignore
so that we don't break compatibility.

I'm waiting for a clear response on https://discord.com/channels/707636530424053791/1413511775546118215/1414520241912021062 before applying any fixes on disabling gitignore for specific arguments

### DIFF
--- a/core/modulesource.go
+++ b/core/modulesource.go
@@ -353,10 +353,7 @@ func (src *ModuleSource) LoadContextDir(
 		}
 
 		// Ensure backward compatibility by not applying .gitignore rules if the engine version is less than v0.17.17
-		noGitIgnore := false
-		if semver.Compare(src.EngineVersion, "v0.18.17") < 0 {
-			noGitIgnore = true
-		}
+		noGitIgnore := src.compareEngineVersion("v0.18.17") < 0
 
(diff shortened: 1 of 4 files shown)


core/modulesource.go
```diff
@@ -351,6 +352,12 @@ func (src *ModuleSource) LoadContextDir(
 			return inst, fmt.Errorf("path %q is outside of context directory %q, path should be relative to the context directory", path, ctxPath)
 		}
 
+		// Ensure backward compatibility by not applying .gitignore rules if the engine version is less than v0.17.17
+		noGitIgnore := false
+		if semver.Compare(src.EngineVersion, "v0.18.17") < 0 {
```
jedevc: nit: make a little helper function?

TomChv: Done in [33173b4](https://github.com/dagger/dagger/pull/11011/commits/33173b44c2fbefc2ba15ba73d512188eaf856b8a)

jedevc: sorry, my bad, I meant something like `supportsGitignore(semver string)`. Then the code can become a bit more readable.

Issue #10994, opened 2025-09-05
🐞 defaultPath cannot use files ignored by .gitignore on `dagger call`


TomChv:
### What is the issue?

With the new dagger version v0.18.17, we are automatically applying .gitignore when calling host.Directory.
But that means you can't use an ignored file as a argument anymore:

```go
func (e *Example) Example(
  ctx context.Context

  //+defaultPath=./foo.txt
  foo *dagger.File,
  ...
) (...)
```

```
# .gitignore
foo.txt
```

See for the full context: https://discord.com/channels/707636530424053791/1413511775546118215

We need a way to bypass the `.gitignore` for specific argument or add the defaultPath to the list of `include`.
That may be the right quick fix

### Dagger version

dagger v0.18.17

danielgafni:
Lol mine was 4 minutes later #10996. I guess I'll close it. Here is my original issue: 

This erroneously affects files that are already added staged (for ex with `git add --force`, or added prior to `.gitignore` modifications), but would have fallen under .gitignore if they wouldn't.

I have a few small .mp4 files checked into the repo for testing purposes. My .gitignore also ignores **/.mp4. My tests in Dagger now started failing after an upgrade to `0.8.17`.

**Files explicitly added to git should not be excluded from COPY operations - that's valid source code!**

TomChv:
I opened this PR to ensure we won't break old modules with the gitignore changes: https://github.com/dagger/dagger/pull/11011, I should have been done before the release but better later than never.

I also did some tests and it seems it's absolutely impossible to bypass the auto ignore from a `dagger call`.
But it's totally possible from dagger shell, here's an example:

```go
package main

import (
	"dagger/test/internal/dagger"
)

type Test struct{}

func (m *Test) Repo(
	//+defaultPath="/LICENSE"
	file *dagger.File,
) *dagger.File {
	return file
}
```

```
#.gitignore
/dagger.gen.go
/internal/dagger
/internal/querybuilder
/internal/telemetry
/.env
LICENSE
```

🔴 FAIL:

```shell
dagger call repo --file=./LICENSE contents 
# failed to stat file /LICENSE: lstat /LICENSE: no such file or directory
```

🟢 Work:

```shell
dagger -c 'repo --file $(host | directory . --no-git-auto-ignore | file /LICENSE) | contents'
#  Apache License
# Version 2.0, January 2004
# http://www.apache.org/licenses/
# ...
```

[...]

PR #11017, opened 2025-09-09
Various gitignore fixes

jedevc:
Fixes https://github.com/dagger/dagger/issues/10994 and other related issues.

See individual commits for detailed explanation.


core/schema/host.go
```diff
@@ -151,7 +151,7 @@ func (s *hostSchema) Install(srv *dagql.Server) {
 				dagql.Arg("exclude").Doc(`Exclude artifacts that match the given pattern (e.g., ["node_modules/", ".git*"]).`),
 				dagql.Arg("include").Doc(`Include only artifacts that match the given pattern (e.g., ["app/", "package.*"]).`),
 				dagql.Arg("noCache").Doc(`If true, the directory will always be reloaded from the host.`),
-				dagql.Arg("noGitAutoIgnore").Doc(`Don't apply .gitignore filter rules inside the directory`),
+				dagql.Arg("gitignore").Doc(`Apply .gitignore filter rules inside the directory`),
```
jedevc, line 155: The big change here is this.

I want to make this opt-in - we made a breaking change accidentally, and I think it's worth backing out of that. Since this API cannot be called by modules, it's outside module-compat so we should be fine for that.

This makes more sense with the approach we're going for now; we only need to add it for module loading, instead of needing to opt out for flag loading, and for context directories, and tests, etc.

jedevc: @shykes can I get a check here?

Since we all agreed on the API before, wanna make sure I get everyone involved with that original discussion on board here.

engine/sources/local/copy/copy.go
```diff
@@ -254,7 +257,7 @@ type parentDir struct {
 	copied  bool
 }
 
-func newCopier(destRoot string, chown Chowner, tm *time.Time, mode *int, xeh XAttrErrorHandler, includePatterns, excludePatterns []string, useGitignore bool, alwaysReplaceExistingDestPaths bool, changeFunc fsutil.ChangeFunc) (*copier, error) {
+func newCopier(destRoot string, chown Chowner, tm *time.Time, mode *int, xeh XAttrErrorHandler, only []string, includePatterns, excludePatterns []string, useGitignore bool, alwaysReplaceExistingDestPaths bool, changeFunc fsutil.ChangeFunc) (*copier, error) {
```
sipsma: nit:

```suggestion
func newCopier(destRoot string, chown Chowner, tm *time.Time, mode *int, xeh XAttrErrorHandler, only map[string]struct{}, includePatterns, excludePatterns []string, useGitignore bool, alwaysReplaceExistingDestPaths bool, changeFunc fsutil.ChangeFunc) (*copier, error) {
```

Seems like we may as well just have the caller provide it as a set already, since right now they accumulate it as a list and then it always gets converted into a set. I could see this plausibly mattering for particularly gnarly cases of like 10k-100k file syncs (i.e. node_modules)

core/schema/host.go
shykes: Yes I'm on board. Technically this should be `gitAutoIgnore` for consistency with past bikeshedding, but I'm going to assume this is your personal preference, and drop that particular nit 😁

jedevc: Yup, I don't think we need the `auto` in there - it's not automatic anymore, you need to enable it, so I think we should remove that, and just keep it to `gitignore`.

engine/sources/local/copy/copy.go
jedevc: Swapped to accumulate directly into a set, and avoid the list conversion!

PR #11018, opened 2025-09-09
Fix typescript introspection for blueprints

kpenfound:
Blueprint modules in typescript were failing to load during introspection because it was looking for the wrong main object name between the main module and blueprint module.

I added tests to confirm the fix and also confirm that its not broken in python.

The main fix is here:
- sdk/typescript/src/module/introspector/dagger_module/module.ts
- sdk/typescript/src/module/introspector/typescript_module/ast.ts

One possible implication of this change is a case where I have one normal typescript module and its missing the expected main object but has some other object and that gets picked up somehow. I'll make sure that case is covered in the tests somewhere

Original implementation of blueprints for reference: https://github.com/dagger/dagger/pull/10712

Here's the test failure without the fix applied ("main" module is called "App", blueprint module is called "MyblueprintTs"):
```
▼ MyblueprintTs.hello: String! 1.8s ERROR
could not find module entrypoint: class App from import. Class should be exported to benefit from all features.                                                                                                             
could not find module entrypoint: class App from import. Class should be exported to benefit from all features.                                                                                                             
/src/myblueprint-ts/sdk/core.js:135361                                                                                                                                                                                      
      throw new IntrospectionError(`could not find main object ${this.name} in module ${JSON.stringify(mainModule, null, 2) ?? ""} located at ${this.ast.files}`); 
```

### DIFF
--- a/core/integration/blueprint_test.go
+++ b/core/integration/blueprint_test.go
@@ -53,49 +53,39 @@ func (BlueprintSuite) TestBlueprintUseLocal(ctx context.Context, t *testctx.T) {
 	})
 }
 
-func (BlueprintSuite) TestBlueprintWithDependency(ctx context.Context, t *testctx.T) {
-	c := connect(ctx, t)
-	t.Run("use a blueprint which has a dependency", func(ctx context.Context, t *testctx.T) {
-		modGen := blueprintTestEnv(t, c).
-			WithWorkdir("app").
-			With(daggerExec("init", "--blueprint=../myblueprint-with-dep"))
-		// Verify blueprint was installed by calling function
-		out, err := modGen.
-			With(daggerExec("call", "hello")).
-			Stdout(ctx)
-		require.NoError(t, err)
-		require.Contains(t, out, "hello from blueprint")
-	})
-}
+func (BlueprintSuite) TestBlueprintInit(ctx context.Context, t *testctx.T) {
+	type testCase struct {
+		name          string
+		blueprintPath string
+	}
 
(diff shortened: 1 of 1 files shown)


kpenfound:
Going to change tactics a bit here after discussion. The ts sdk should loop through all objects to make sure we get them all rather than just choosing the next object as the main object

kpenfound:
ready for review @TomChv !

core/integration/blueprint_test.go
```diff
@@ -68,6 +68,36 @@ func (BlueprintSuite) TestBlueprintWithDependency(ctx context.Context, t *testct
...
+			With(daggerExec("init", "--blueprint=../myblueprint-py"))
+		// Verify blueprint was installed by calling function
+		out, err := modGen.
+			With(daggerExec("call", "hello")).
+			Stdout(ctx)
+		require.NoError(t, err)
+		require.Contains(t, out, "hello from blueprint")
+	})
+}
+
```
TomChv: Could be simplified with an array

```go
func (BlueprintSuite) TestBlueprint(ctx context.Context, t *testctx.T) {
  type testCase struct {
    name string
    blueprintPath string
  }

  for _, tc := range []testCase{
    {
      name: "typescript"
      blueprintPath: "../myblueprint-ts"
    }
    
    // Also add a test for go maybe or all other tests are already with Go?
  }
} {
     t.Run(t.name,  ... ) {}
   }
```

TomChv (approved):
Left a little comment to improve the test, otherwise LGTM.

I'll give my approval since it's a minor change that doesn't impact the implementation, you can merge as soon as you finished the tests changes :)

PR #11023, opened 2025-09-10
Spelling

jsoref:
Roughly this PR would address most of the items in https://github.com/jsoref/dagger/actions/runs/17617300634/attempts/1#summary-50053683369 (resulting in https://github.com/jsoref/dagger/actions/runs/17617302128/attempts/1#summary-50053688729)

I'm marking this as a draft because I fully expect a lot of pushback and I couldn't get CI to work in my fork. I'm doing this in spare time (I've signed into Discord so feel free to tag me there, but do note that I don't regularly **use** discord, so I may not respond for a few months, whereas I am likely to respond w/in a few weeks to GitHub comments/tags).

I'm aware that a nontrivial portion of these files are generated, my tooling generally corrects both sources and generated files, if it's preferred that I drop things, I just need a list of files to exclude, that isn't a problem -- I'm used to dropping files.

I can do splits, but it's much easier for me to split by commit or possibly directory/file type than change type ("comment" / "docs" / "code") as that requires considerably more effort and is fairly error-prone beyond the fact that such distinctions can be quite murky and changes often span multiple types.


jedevc:
Hey @jsoref, it looks like this PR is going to need a lot of input from maintainers to get it into a state where it's ready to merge.

There are a lot of changes to generated files, changelogs, etc - which we wouldn't want. There's also just changes that are incorrect/highly opinionated, a brief scan reveals quite a few just plainly wrong suggestions. It's not just a quick, LGTM, merge, move on :sweat_smile: 

Since this is gonna require more maintainer time than we really have (we're more focused on user-fixes + feature work), I think it's unlikely we'll be able to merge this.

If there are *user-facing* changes (to api docs, etc) that are purely spelling/grammar changes (not opionated), then I'd happy to review those :tada: However, fixing everything all at once (including internal comments seen only by developers) just isn't worth it IMO.

PR #11024, opened 2025-09-10
Ignore dependabot events on evals workflow

kpenfound:
Update the evals workflow to run on PR events with path filtering instead of push events with path filtering.

Push events with path filtering was causing this workflow to get executed when tags are pushed and on dependabot PRs.

For the tags, it was running because [path filters are not evaluated for pushes of tags](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore).
For dependabot, I'm not actually sure, but probably something similar.

Update: the `on: push` is important because evals are only run on branches of dagger/dagger. I've changed the approach to ignore the dependabot events instead


kpenfound:
The docs check is failing because of the markdown link in contributing.md. I'll fix the docs module to not worry about markdown outside of `./docs/`

kpenfound:
~~Hold up on this. It needs to remain a `push` event, so we'll have to avoid the unnecessary runs another way~~

vito (approved):

PR #11026, opened 2025-09-11
chore: change azure devops regex to allow "%" for url-encoded spaces in project and repository names

Haavasma:
Azure devops repositories allow creating organizations, projects and repositories with spaces in it's names.
This results in git repository URLs to possibly look like this:
`https://dev.azure.com/My%20Organization/My%20Project/_git/My%20Repo`
or with ssh:
`ssh://git@dev.azure.com/v3/My%20Org/My%20Project/My%20Repo`

When creating remote dagger modules on such a repository, these modules cannot be installed or used, as dagger only allows alphanumerical + "_.-". when validating the URL.

### DIFF
--- a/core/integration/http_test.go
+++ b/core/integration/http_test.go
@@ -29,10 +29,10 @@ func (HTTPSuite) TestHTTP(ctx context.Context, t *testctx.T) {
 	c := connect(ctx, t)
 
 	// do two in a row to ensure each gets downloaded correctly
-	url := "https://raw.githubusercontent.com/dagger/dagger/main/CONTRIBUTING.md"
+	url := "https://raw.githubusercontent.com/dagger/dagger/main/LICENSE"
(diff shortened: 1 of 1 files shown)


engine/vcs/vcs.go
```diff
@@ -303,8 +303,16 @@ func FromDir(dir, srcRoot string) (vcs *Cmd, root string, err error) {
...
+				return nil, "", fmt.Errorf(
+					"directory %q uses %s, but parent %q uses %s",
+					filepath.Join(
+						srcRoot,
+						rootRet,
+					),
+					vcsRet.Cmd,
+					filepath.Join(srcRoot, root),
+					vcs.Cmd,
+				)
```
jedevc: These lines shouldn't change? It appears to be an auto-formatter, please make sure your changes are just the ones relevant to the change you want to make.

Haavasma: reverted

PR #11030, opened 2025-09-11
Improve performance of version module

jedevc:
This currently works - but is actually slower than the current implementation - so depends on #11032 for speedier + better cached git checkouts.

I'll annotate the PR with an explanation of the changes.

Prior attempts here:
- https://github.com/dagger/dagger/pull/10801
- https://github.com/dagger/dagger/pull/10952

### DIFF
--- a/.github/workflows/engine-cli.gen.yml
+++ b/.github/workflows/engine-cli.gen.yml
@@ -840,7 +840,7 @@ jobs:
                 #!/bin/bash --noprofile --norc -e -o pipefail
 
                 cd $(mktemp -d)
-                dagger shell -M -c "git https://github.com/dagger/dagger.git | ref $DAGGER_REF | tree | export ."
+                git clone https://github.com/dagger/dagger.git . --revision=$DAGGER_REF
 
                 ./hack/build
 
@@ -1032,7 +1032,7 @@ jobs:
                 #!/bin/bash --noprofile --norc -e -o pipefail
 
                 cd $(mktemp -d)
-                dagger shell -M -c "git https://github.com/dagger/dagger.git | ref $DAGGER_REF | tree | export ."
+                git clone https://github.com/dagger/dagger.git . --revision=$DAGGER_REF
 
                 ./hack/build
 
@@ -1224,7 +1224,7 @@ jobs:
                 #!/bin/bash --noprofile --norc -e -o pipefail
 
                 cd $(mktemp -d)
-                dagger shell -M -c "git https://github.com/dagger/dagger.git | ref $DAGGER_REF | tree | export ."
+                git clone https://github.com/dagger/dagger.git . --revision=$DAGGER_REF
 
                 ./hack/build
 
@@ -1416,7 +1416,7 @@ jobs:
                 #!/bin/bash --noprofile --norc -e -o pipefail
 
                 cd $(mktemp -d)
-                dagger shell -M -c "git https://github.com/dagger/dagger.git | ref $DAGGER_REF | tree | export ."
+                git clone https://github.com/dagger/dagger.git . --revision=$DAGGER_REF
 
                 ./hack/build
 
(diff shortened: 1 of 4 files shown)


jedevc:
So, this depends on https://github.com/dagger/dagger/pull/11032.

The reason this is somehow *slower* is because we are *always* performing a full checkout - even though a checkout already exists. This happens because the cache key for `tree` in `Directory.asGit.ref.tree` depends on the cache key for `ref` - which has `CachePerSession` set.

This was originally set so that we would ensure that the results didn't get stale. However - this isn't required after #11032, because `git ls-remote` *informs* us whether the ref is actually stale or not.

version/git.go
```diff
@@ -13,69 +12,30 @@ import (
...
 
-func git(ctx context.Context, gitDir *dagger.Directory, dir *dagger.Directory) (*Git, error) {
-	ctr := dag.Wolfi().
-		Container(dagger.WolfiContainerOpts{Packages: []string{"git"}}).
-		WithWorkdir("/src")
+func git(ctx context.Context, repo *dagger.GitRepository, dir *dagger.Directory) (*Git, error) {
+	checkout := repo.Head().Tree(dagger.GitRefTreeOpts{Depth: -1})
 
+	ctr := dag.Container().From("alpine/git:latest")
+	// ctr := dag.Wolfi().Container(dagger.WolfiContainerOpts{Packages: []string{"git"}})
```
jedevc, line 26: Even with https://github.com/dagger/dagger/pull/11057, this still seems to be slow. I *think* generalized function caching can make this much faster, but in the meantime, this alternative is much faster.

version/git.go
```diff
@@ -13,69 +12,30 @@ import (
...
-		// enter detached head state, then we can rewrite all our refs however we like later
-		ctr = ctr.WithExec([]string{"sh", "-c", "git checkout -q $(git rev-parse HEAD)"})
-
-		// manually add a remote (since .git/config was removed earlier)
-		ctr = ctr.WithExec([]string{"git", "remote", "add", "origin", "https://github.com/dagger/dagger.git"})
-
-		// do various unshallowing operations (only the bare minimum is
-		// provided by the core git functions which are used by our remote git
-		// module sources)
-		maxDepth := "2147483647" // see https://git-scm.com/docs/shallow
```
jedevc, line 56: This is all an elaborate dance to handle shallow clones - for something that is actually quite an edge case, it adds a lot of performance overhead. *Any* change to the git directory was invalidating all this.

So, we simply now require that the version module be built from a full-clone. Garbage in, garbage out.

**If** we want to do this, we should probably do it using something like https://github.com/dagger/dagger/issues/10200#issuecomment-3140047394.

version/git.go
```diff
@@ -125,10 +85,6 @@ func (git *Git) VersionTags(
 	// Optional commit sha to get tags for
 	commit string, // +optional
 ) ([]VersionTag, error) {
-	if !git.Valid {
-		return nil, nil
-	}
```
jedevc, line 130: I think this was from before we swapped the default to preserve the git dir - so this *could* happen.

However, this was ages ago, so it can go.

version/git.go
```diff
@@ -274,24 +222,8 @@ func (git *Git) Commit(ctx context.Context, ref string) (*Commit, error) {
...
-		}).
-		Stdout(ctx)
-	if err != nil {
-		return nil, err
-	}
-	raw = strings.TrimSpace(raw)
-
-	return git.Commit(ctx, raw)
+func (git *Git) MergeBase(ctx context.Context, ref string, ref2 string) (string, error) {
+	return git.Repo.Ref(ref).CommonAncestor(git.Repo.Ref(ref2)).Commit(ctx)
```
jedevc, line 226: This has native support in the dagger API now: #10849.

version/main.go
```diff
@@ -31,8 +31,7 @@ func New(
 	inputs *dagger.Directory,
 	// +optional
 	// +defaultPath="/"
-	// +ignore=["*", "!.git", "!**/.gitignore", ".git/config"]
-	gitDir *dagger.Directory,
+	gitDir *dagger.GitRepository,
```
jedevc, line 32: Contextual git #10847 :tada:

This means we can checkout any ref, at any depth which is what we need for all the operations.

PR #11031, opened 2025-09-11
fix: don't set origin on local git

jedevc:
We were setting it to a filepath that's not persistent or fetchable.


jedevc:
Tests are legitimately failing, don't merge.

PR #11032, opened 2025-09-11
Resolve `git ls-remote` once

jedevc:
This is a combined cleanup + improvement of the git backends.

To avoid a bunch of problems downstream, we eagerly evaluate `git ls-remote` on the target as soon as we get it - instead of when we resolve a ref.

This allows us to:
- Avoid extraneous calls to `ls-remote` - we make one giant call, and then have all the tags, branches, etc.
- This means we hugely reduce the amount of chit-chat with the server.
- Maybe set better cache keys for improved performance? Especially for local sources.

### DIFF
--- a/.changes/unreleased/Breaking-20250923-113245.yaml
+++ b/.changes/unreleased/Breaking-20250923-113245.yaml
@@ -1,0 +1,9 @@
+kind: Breaking
+body: |-
+    `GitRepository`'s `withAuthToken` and `withAuthHeader` are now no-ops.
+
+    Instead authentication should *only* be provided at the top-level `git` query, since it's now queried immediately.
+time: 2025-09-23T11:32:45.600115242+01:00
+custom:
+    Author: jedevc
+    PR: "11032"


shykes:
I'm sure this is the right thing to do, but will it make some end-to-end calls slower because of the eagerness?

jedevc:
@shykes in most cases we were gonna have to perform this operation anyways - sometimes, it was possible to perform it multiple times (e.g. you could call `.tags` and `.ref` on a single `GitRepository`, and we'd do `ls-remote` twice). However, there is one case this is *technically* a slow down, which is when you were checking out a single commit - then we wouldn't have done an `ls-remote`.

For the huge improvement in cache keys as well as correctness, though, this is an easy win.

jedevc:
Realized this is kind of bound up in #11073 - the *idea* of this PR is that we "freeze" the remote in place at the point of construction, so you get a nice consistent view of the repo.

This isn't true for the tags that are pulled as part of the history - however, this *should* be solvable with some clever logic (or removing that functionality, but then #11030 isn't really possible in the same way).

sipsma (approved):

Issue #6723, opened 2024-02-23
`dagger call` is tedious with too many flags+args


sipsma:
Somewhat related to https://github.com/dagger/dagger/issues/6112 but a bit more general

Passing a large number of flags and args when using `dagger call` can become way too tedious to type out by hand. One example would be support for filtering dirs as mentioned [here](https://github.com/dagger/dagger/issues/6155#issuecomment-1961485732). But really any time a function accepts more than a few flags and/or the arguments to the flag end up being verbose are a problem.  

This leaves users reaching for wrappers like shell scripts/makefiles/etc. While that's not the end of the world (i.e. we aren't going to try to stop that), it's not great that users have to layer more abstractions on top of dagger to use it in relatively simple situations.

A few ideas have come up to potentially solve this:
1. `defaultArgs` specified in `dagger.json` as described [here](https://github.com/dagger/dagger/issues/6291#issuecomment-1907368992)
   * Basically, module constructors can have default args encoded in `dagger.json` and those args have the same features as the syntax in `dagger call` (e.g. local files/dirs specified by path, etc.)
2. Support for "recording+saving" `dagger call` invocations and then being able to run them by name. As a strawman for the sake of a concrete example:
   * `dagger call --save-as=my-build -m mydep build --dir '/some/path,**/*.go,!testdata' --some-arg false --token env:MY_TOKEN`
   * After the above, you can now run `dagger call my-build`
3. Support for `dagger script` (https://github.com/dagger/dagger/issues/5774), possibly with an addition to support running scripts written to a file rather than inline.

These aren't mutually exclusive either. I think `2.` might be my preferred starting place at the moment for the sake of being both general+relatively straightforward to implement, but all are pretty tempting and viable.

shykes:
@sipsma FYI I am reviving and refreshing #5774. It is my favorite candidate for solving this issue, I will flesh out my argument and try to build momentum. I think you're right that it's related to #5774, maybe we can kill 2 birds with one stone.

I think we should merge this issue and #6112. The conversation is getting pretty fragmented. WDYT? I'm also going to close #5728.

sipsma:
> @sipsma FYI I am reviving and refreshing #5774. It is my favorite candidate for solving this issue, I will flesh out my argument and try to build momentum. I think you're right that it's related to #5774, maybe we can kill 2 birds with one stone.
> 
> I think we should merge this issue and #6112. The conversation is getting pretty fragmented. WDYT? I'm also going to close #5728.

[...]

PR #11034, opened 2025-09-12
User defaults: persist any dagger function argument to .env

shykes:
## Overview

*Local defaults*: user can add default arguments to their modules, using a local .env file.

This avoids [typing too many CLI arguments on every call](https://github.com/dagger/dagger/issues/6723).

## How it works

1. Choose a module argument you want to persist locally. For example, the argument `token` of the module `github.com/shykes/x/discord`
2. Pick a location for your `.env` file. This could be in the module directory, or in a parent directory on your filesystem, depending on when you want it to be loaded. See standard `.env` conventions.
3. Write the argument as a variable in `.env`. The variable name must be `<module name>_<argument name>` (case-insensitive). If your `.env` file is *within your module directory*, you can also use the variable name `<argument name>` (the module name is implicit). You can combine both notations. Variables that don't match this convention will be ignored. The variable value should be the same value you passed explicitly on the CLI. For example if you called `dagger -m github.com/shykes/x/discord --token=op://foo/bar`, your variable should be `DISCORD_TOKEN=op://foo/bar`.
4. Call your module the usual way, but omit the argument. For example `github.com/shykes/x/discord -c servers`.

## Supported values

Variable values support the same format as command-line arguments:

- Literal scalar values
- JSON-encoded arrays
- `Directory`: local path or git address
- `File`: local path or git address
- `Socket`: local path
- `Container`: container image address
- `Secret`: url with schemes `op://` (1password), `vault://` (hashicorp vault), `env://` (environment variable), `cmd://` (command execution), `file://` (local file path), `libsecret://` (libsecret provider)
- `GitRepository`: git address
- `GitRef`: git address
- `Service` local `tcp://` url

## Example

```console
cat > .env <<.
DISCORD_TOKEN=op://myvault/mydiscordtoken/credential
.
```

```console
dagger -m github.com/shykes/x/discord -c servers
```

--> This will list your discord servers, using the token in the `.env` file.

## TODO

- [x] `dagger --env-file` to select an alternative filename
- [ ] `dagger --no-env-file` to disable local defaults entirely
- [x] Clean up implementation
- [x] Local defaults can be cleanly introspected by the CLI and SDKs
- [x] Support arguments to functions other than the constructor (as requested by @nipuna-perera and @nadirollo )
- [x] Clearly tell the user which default was applied where
- [x] Evaluate quotes (with escaping supported) in env file
- [x] Tests
- [ ] Don't inject defaults in dependencies by default. Make it a configurable opt-in

## See also

- [POC 1](https://github.com/dagger/dagger/pull/10697) (deprecated)
- [POC 2](#11000) (deprecated)


nipuna-perera:
Can't wait to try this! Does this still only support constructor args? Or is it expanded to support any arg?

shykes:
@nipuna-perera I'm making my way to "any args" :)

shykes:
I added suppport for implicit module name.

nadirollo:
This is going to be a game changer for us! With this we can finally replace many makefiles/justfiles that wrap around dagger!
If it ends up supporting function args, that would be amazing :) 

Can't wait to try this out 👍

shykes:
> This is going to be a game changer for us! With this we can finally replace many makefiles/justfiles that wrap around dagger!

Awesome! Yes, removing the need for wrappers is the goal :)

>  If it ends up supporting function args, that would be amazing :)

Do you mean arguments to functions other than the constructor? Yeah I'm working on that right now :) @nipuna-perera also asked for that.

May I ask, are you thinking of functions attached to the top-level object? Or even further, functions attached to potentially any object type in the module?

nadirollo:
I think args to the top-level object would be great. Our blueprints are "simple" enough that don't require many chainable objects (yet...?)

nipuna-perera:
> May I ask, are you thinking of functions attached to the top-level object? Or even further, functions attached to potentially any object type in the module?

I replied here too - https://discord.com/channels/707636530424053791/1413254056188579901/1417690771175243828

I think it should apply uniformly to any dagger addressable object functions. Otherwise it adds a layer of complexity. One may expect a chained function to have the same behavior of the main dagger object. It's probably a niche use case but for the sake of UX I don't see a drawback of it applying to everything.

babo:
Is this supports dagger.Secret as a string value or as a file?

shykes:
> Is this supports dagger.Secret as a string value or as a file?

It supports all the same secret addresses as for CLI arguments:

- file: `file://`
- env variable: `env://`
- 1password: `op://`
- hashicorp vault: `vault://`

I don't believe literal string values are supported

See https://docs.dagger.io/features/secrets

core/integration/suite_test.go
```diff
@@ -64,7 +64,8 @@ func BenchMiddleware() []testctx.Middleware[*testing.B] {
 
 func connect(ctx context.Context, t testing.TB, opts ...dagger.ClientOpt) *dagger.Client {
 	opts = append([]dagger.ClientOpt{
-		dagger.WithLogOutput(testutil.NewTWriter(t)),
+		// FIXME: test spans are easier to read in the TUI when this is silenced
+		dagger.WithLogOutput(io.Discard),
```
vito, line 68: I could believe this is better, I find myself reading the plain output a lot but it might just be because it's there and I see it first. Commenting anyway in case you intended to undo this and forgot

core/schema/coremod.go
```diff
@@ -248,6 +248,7 @@ func (m *CoreMod) TypeDefs(ctx context.Context, dag *dagql.Server) ([]*core.Type
 						Description: introspectionArg.Description,
 					}
 
+					// FIXME: allow user-defined local defaults on core functions?
```
vito: just flagging a FIXME, fine if out of scope, it does seem a little weird to reason about imo, maybe a step too far?

core/schema/modulesource.go
```diff
@@ -728,7 +734,7 @@ func (s *moduleSourceSchema) directoryAsModule(
 			},
 		},
 		dagql.Selector{
-			Field: "asModule",
+			Field: "(asModule)",
```
vito: wat

core/schema/module.go
```diff
@@ -78,6 +78,9 @@ func (s *moduleSchema) Install(dag *dagql.Server) {
 		dagql.NodeFunc("generatedContextDirectory", s.moduleGeneratedContextDirectory).
 			Doc(`The generated files and directories made on top of the module source's context directory.`),
 
+		dagql.Func("localDefaults", s.moduleLocalDefaults).
+			Doc(`User-defined defaults values, loaded from local .env files.`),
```
vito:
```suggestion
			Doc(`User-defined default values, loaded from local .env files.`),
```

re: trailing `.` - initially removed it, but we're super inconsistent there atm, to the point where I don't know what the pattern/guiding principle is anymore. For some reason I remember "first line has no `.`, subsequent ones do" but I don't remember if that came from solid ground.

core/host.go
```diff
@@ -24,6 +25,33 @@ func (*Host) TypeDescription() string {
...
+// Lookup an environment variable in the host system from the current context
+func (Host) GetEnv(ctx context.Context, name string) string {
+	dag, err := CurrentDagqlServer(ctx)
+	if err != nil {
+		return ""
+	}
+	var plaintext string
+	err = dag.Select(ctx, dag.Root(), &plaintext,
+		dagql.Selector{
+			Field: "secret",
```
vito: lol sneaky

This works obviously but it might be wise to cut out the DagQL overhead and cut straight to the Buildkit plumbing:

https://github.com/dagger/dagger/blob/7c41c975fc9ee293510f3871fd9cfb8ae3c45a4e/core/secret.go#L164-L187

core/modulesource.go
```diff
@@ -250,6 +251,153 @@ func (src *ModuleSource) Pin() string {
...
+	if src.Kind != ModuleSourceKindLocal {
+		return nil, "", nil
+	}
+
+	dag, err := CurrentDagqlServer(ctx)
+	if err != nil {
+		return nil, "", err
+	}
+
+	// FIXME: .env must be at the root of the module directory
```
vito, line 266: (flagging FIXME, same as before, fine if saving for future iteration)

core/modulesource.go
```diff
@@ -250,6 +251,153 @@ func (src *ModuleSource) Pin() string {
...
+	envFilePath := path.Join(moduleDirPath, ".env")
+	var envFile *EnvFile
+	if err := dag.Select(ctx, dag.Root(), &envFile,
+		dagql.Selector{Field: "host"},
+		dagql.Selector{
+			Field: "file",
+			Args: []dagql.NamedInput{
+				{Name: "path", Value: dagql.String(envFilePath)},
+			},
+		},
```
vito, line 312: could maybe just try accessing it instead of checking if it exists? or is the idea to avoid scary looking failed telemetry?

core/modulesource.go
```diff
@@ -250,6 +251,153 @@ func (src *ModuleSource) Pin() string {
...
+// loadocalDefaults loads and merges environment files for local module defaults.
+// It combines the inner .env file (from the module's source root) with relevant
+// entries from the outer .env file (found via find-up from the host), filtered
+// by the module name and original module name as prefixes.
+//
+// Example:
+// Inner .env (mymodule/.env): FOO=bar
+// Outer .env (found via find-up): MYMODULE_BAZ=qux, OTHER_VAR=ignored
+// Result: FOO=bar, BAZ=qux (prefix "MYMODULE_" removed from outer entries)
+func (src *ModuleSource) LoadLocalDefaults(ctx context.Context) error {
```
vito: nit: this could maybe be refactored into a `sync.Once`'d lazy-initialized method, but I see it's really only used in various bootstraping constructors, of which there will probably only be a few, so no big deal tbh.

vito (commented):
Somewhat rushed initial review, mostly nits! Looking good

core/schema/modulesource.go
shykes: What the... No idea how this happened 🤷‍♂️

core/schema/module.go
shykes: I have no clue either... I saw one with a dot, so did the same thing. No opinion either way.

core/modulesource.go
shykes: Thanks. Yeah this is fine for a future iteration.

core/schema/coremod.go
shykes: Yes maybe too far. I could see some applications in dependency injection, but at the very least it would require some more thinking. I'll remove the FIXME.

core/modulesource.go
shykes: If I try to access it directly, then I have to process "errors that aren't really errors". I know there's a SDK facility for that, but I prefer to use pure engine API out of principle. The principle being: "make the engine smart and the SDKs dumb, it scales better" :)

core/host.go
shykes: How bad is the overhead? I kind of like that it shows up in traces..

core/modulesource.go
```diff
@@ -265,6 +413,14 @@ func (src *ModuleSource) CalcDigest() digest.Digest {
 		src.ContextDirectory.ID().Digest().String(),
 	}
 
+	// Include local defaults in digest so changes to env files invalidate cache
+	if src.LocalDefaults != nil {
+		vars, _ := src.LocalDefaults.Variables(context.Background(), false)
```
sipsma: should have `CalcDigest` take a `ctx` arg rather than use background here

core/schema/modulesource.go
```diff
@@ -2374,7 +2394,9 @@ func (s *moduleSourceSchema) moduleSourceAsModule(
 
 	// Create module with blueprint source for SDK operations
 	mod := &core.Module{
-		Source:        dagql.NonNull(src),
+		Source:         dagql.NonNull(src),
+		OriginalSource: dagql.NonNull(originalSrc),
+
 		ContextSource: dagql.NonNull(originalSrc),
```
sipsma: Do we need both `OriginalSource` and `ContextSource` when they are the same value, or can we consolidate them?

core/modulesource.go
```diff
@@ -265,6 +413,14 @@ func (src *ModuleSource) CalcDigest() digest.Digest {
...
+		for _, v := range vars {
```
sipsma: nit: maybe worth sorting these by `v.Name` first, then appending for better digest stability

core/modulesource.go
sipsma: You could still save the engine a little effort by changing this to
```go
if err := dag.Select(ctx, dag.Root(), &envFile,
		dagql.Selector{Field: "host"},
        dagql.Selector{
			Field: "directory",
			Args: []dagql.NamedInput{
				{Name: "path", Value: dagql.String(moduleDirPath)},
				{Name: "include", Value: dagql.ArrayInput[dagql.String]{".env"}},
			},
		},
		dagql.Selector{
			Field: "file",
			Args: []dagql.NamedInput{
				{Name: "path", Value: dagql.String(".env")},
			},
		},
        dagql.Selector{
			Field: "asEnvFile",
			Args: []dagql.NamedInput{
				{Name: "expand", Value: dagql.Opt(dagql.NewBoolean(true))},
			},
		},
	); err != nil {
		return nil, "", fmt.Errorf("failed to load inner env file in %s: %w", moduleDirPath, err)
	}
```

which helps because it chains off the same `host.Directory` call as above, which lets the engine instantly know it can re-use the cache there. The current code has the same end effect but the engine has to do more work to realize that it's reloading the same file as before and thus re-use cache for it.

Doubtful this is a *huge* bottleneck or something but IMO we should be willing to save 100s of milliseconds where we can in terms of module loading (provided it's a simple adjustment like this).

core/modfunc.go
```diff
@@ -213,32 +468,56 @@ func (fn *ModuleFunction) CacheConfigForCall(
...
+			// Only default objects need processing at this time.
+			// Primitive default values were already processes earlier
+			//  in the flow.
+			// This applies to both types of object defaults:
+			//  1) "contextual args" from `defaultPath` annotations
+			//  2) "local defaults" from user-defined .env
 			continue
 		}
-		ctxArgs = append(ctxArgs, argMetadata)
+		localDefault, hasLocalDefault, err := fn.LocalDefault(ctx, argMetadata.Name)
```
sipsma: `fn.LocalDefault` seems like it's doing a bunch of duplicated work for each value of `argMetadata` (e.g. tracing through the code I think it triggers a call to `EnvFile.variables`, which in turn can call `host.EnvVar` multiple times). The dagql cache helps quite a bit, but there's still enough overhead in even getting to the point of hitting cache that this `O(n^2)`-ish behavior feels like it could become noticeable with enough variables.

Is there a straightforward way to re-write this such that we load the local defaults once outside the loop and then just do lookups here? Could also be done in different ways like changing `EnvFile` to load variables when initialized, etc.

shykes: Yeah I should probably optimize it... The system env var lookup is something I just added yesterday, so it's changing the picture in terms of cost...

The reason I implemented it this way in the first place, is because of variable expansion *within the env file*. Each entry in the env file can reference *earlier* variables in the file, so in order to evaluate any variable, at the very least I have to evaluate every value that came before. I could pre-evaluate, but then I need to re-evaluate on every new variable.

shykes:
@sipsma I am confronted with a mystery... I pushed an extra commit to handle `dagger --env-file`. I expected this one to be trivial... But it's not 😭 

The feature works fine when testing manually in `dagger call dev terminal`.

But, for some reason, the integration test I wrote *that does the exact same thing* fails... (`TestLocalDefaults/TestCustomEnvFile`). 😭 😭 😭 

I suspect it has something to do with nested clients, but I did already plumb that through, and after all, my `dagger call dev terminal' manual testing is also a nested client... so it's a complete mystery to me why one works but the other doesn't.

Here's the commit: https://github.com/dagger/dagger/pull/11034/commits/d12f07ab4e676755a4eb9b7340fccf866defb535

sipsma:
> I suspect it has something to do with nested clients, but I did already plumb that through, and after all, my `dagger call dev terminal' manual testing is also a nested client... so it's a complete mystery to me why one works but the other doesn't.

@shykes In that commit it looks like you are forwarding the client's env file setting into nested exec containers, so I suspect:
1. the outer client defaults to using `.env`
2. it creates a nested exec, in which the `DAGGER_ENV_FILE` env var is set to the outer client's value (overriding what was set in the nested container definition)

I can't currently see why we would want to set the `DAGGER_ENV_FILE` env var to what the outer client has set (the outer client is on the host, the inner one is in a container), so can we just remove that plumbing?

shykes:
Ah that makes sense. It's just that I saw a comment in `clientMetadata` advising to also apply any changes in the logic to nested clients. And my test was failing so I assumed that was the reason. But maybe i jumped to conclusions... brain is fried, I'll take a look later 😅 thank you

core/host.go
vito: Ah I assumed you wouldn't want it to show up in traces since it seems like a kind of confusing implementation detail (the name on the tin at this level is 'read host env var', nothing to do with secrets, that just happens to be the already-established path for the mechanism).

The overhead shouldn't be too bad either way, in principle.

shykes:
> > I suspect it has something to do with nested clients, but I did already plumb that through, and after all, my `dagger call dev terminal' manual testing is also a nested client... so it's a complete mystery to me why one works but the other doesn't.
> 
> @shykes In that commit it looks like you are forwarding the client's env file setting into nested exec containers, so I suspect:
> 
> 1. the outer client defaults to using `.env`
> 2. it creates a nested exec, in which the `DAGGER_ENV_FILE` env var is set to the outer client's value (overriding what was set in the nested container definition)
> 
> I can't currently see why we would want to set the `DAGGER_ENV_FILE` env var to what the outer client has set (the outer client is on the host, the inner one is in a container), so can we just remove that plumbing?

Update: I tried removing that plumbing... Then worried that I overshot, and re-added just *a little bit of plumbing*... No luck.

No matter what I do, `dagger --env-file` works fine in non-nested clients, but fails with nesting. I don't understand why...

shykes:
cc @alexcb @jedevc @TomChv in case I'm doing something obviously stupid...

core/modulesource.go
shykes: I'm happy to do it, OK if I make it a follow-up PR?

core/modulesource.go
shykes: @sipsma the issue is that I need to differentiate between 1) no `.env` in the module (not an error) and 2) an error occurred while processing this query.

I was trying to avoid having to differentiate 1 and 2 in a SDK-specific way (ie. by checking the type of `error`), but I guess performance is a good reason to do it the SDK-specific way.

core/host.go
shykes: You're right that the reference to "secret" is a little confusing. Ideally we would have a distinct session attachable just for reading env variables :) But then would have to make sure they don't leak in the traces...

core/integration/suite_test.go
shykes: Yeah it's specific to browsing a big tree of test results in the TUI. Each test starts with the same 20 lines of logs (initialize dagger session etc), which makes it hard to navigate. If there was a hotkey to hide/show logs in spans, I would just use that :)

core/modfunc.go
shykes: Done! Should be much better now.

core/modulesource.go
vito: Yeah no big deal at all

core/schema/modulesource.go
shykes: I will try consolidating them.

shykes: Looks like consolidating didn't break any tests :)

core/host.go
shykes: @vito done

cmd/dagger/main.go
```diff
@@ -89,6 +90,13 @@ func init() {
 	}
 }
 
+func init() {
+	envFileName = os.Getenv("DAGGER_ENV_FILE")
+	if envFileName == "" {
+		envFileName = ".env"
+	}
+}
```
jedevc: In more recent go versions you can do `cmp.Or(os.Getenv("DAGGER_ENV_FILE"), ".env")` if you like.

Just something I've found myself using more and more when I want a one-liner.

core/schema/module.go
jedevc: Suggestion is still relevant.

core/modulesource.go
```diff
@@ -250,13 +252,172 @@ func (src *ModuleSource) Pin() string {
...
+				{Name: "expand", Value: dagql.Opt(dagql.NewBoolean(true))},
+			},
+		},
+	); err != nil {
+		return nil, envFilePath.String(), fmt.Errorf("failed to load outer env file from %q: %s", envFilePath.String(), err.Error())
+	}
+	return envFile, envFilePath.String(), nil
+}
+
+// loadocalDefaults loads and merges environment files for local module defaults.
```
jedevc:
```suggestion
// LoadLocalDefaults loads and merges environment files for local module defaults.
```

core/modulesource.go
```diff
@@ -265,6 +426,18 @@ func (src *ModuleSource) CalcDigest() digest.Digest {
 		src.ContextDirectory.ID().Digest().String(),
 	}
 
+	// Include local defaults in digest so changes to env files invalidate cache
+	if src.LocalDefaults != nil {
+		vars, _ := src.LocalDefaults.Variables(ctx, false)
```
jedevc: Doesn't seem like we should be discarding this error?

core/object.go
```diff
@@ -369,10 +369,10 @@ func (obj *ModuleObject) installConstructor(ctx context.Context, dag *dagql.Serv
 	if err != nil {
 		return fmt.Errorf("failed to create function: %w", err)
 	}
-
+	fn.mergeLocalDefaultsTypeDefs(ctx)
```
jedevc: Error handling?

core/object.go
```diff
@@ -471,6 +486,11 @@ func objFun(ctx context.Context, mod *Module, objDef *ObjectTypeDef, fun *Functi
 	if err != nil {
 		return f, fmt.Errorf("failed to create function %q: %w", fun.Name, err)
 	}
+	// Apply local user defaults to the function's arguments, so that they show
+	// up in installed typedefs (for introspection)
```
jedevc, line 492: nit: the rest of this code generates a dagql field, not a typedef.

core/query.go
```diff
@@ -49,11 +49,22 @@ type Server interface {
...
 	// chains of dependency modules.
 	NonModuleParentClientMetadata(context.Context) (*engine.ClientMetadata, error)
 
+	// Modules served to the main client, aka "non-module parent client", aka "NMPC".
+	//
+	// The NMPC is the nearest ancestor client that is not a module.
+	// It is either a caller from the host like the CLI or a nested exec.
+	// Useful for figuring out where local sources should be resolved from through
+	// chains of dependency modules.
+	NonModuleParentClientModules(context.Context) ([]Mod, error)
```
jedevc: This doesn't seem used.

core/modfunc.go
```diff
@@ -184,20 +185,278 @@ func (fn *ModuleFunction) setCallInputs(ctx context.Context, opts *CallOpts) ([]
...
 		hasArg[name] = true
 	}
-
 	return callInputs, nil
 }
 
+// Load the local defaults for this function, and apply them to the function's
+// typedefs. This makes the local defaults visible in typedef introspection.
+// It does not affect applying local defaults *at function call*
+func (fn *ModuleFunction) mergeLocalDefaultsTypeDefs(ctx context.Context) error {
```
jedevc: Just a general note on how this is implemented - because this goes into the typedefs, etc, then everytime an inner/outer env is changed, we get a new module id. If we want to have the behavior of modifying the typedefs (yup), then we need a new id, otherwise we'll have worse caching issues.

This affects function call caching - e.g. I expect that with https://github.com/dagger/dagger/pull/10975, we'd invalidate cache anytime that any default is changed through this mechanism.

I *think* this may be fixable though, we should be able to strip defaults out of the module before computing the cache key.

Just a note, no action required.

core/modulesource.go
```diff
@@ -120,6 +121,7 @@ type ModuleSource struct {
 	Dependencies    dagql.ObjectResultArray[*ModuleSource] `field:"true" name:"dependencies" doc:"The dependencies of the module source."`
 	ConfigBlueprint *modules.ModuleConfigDependency
 	Blueprint       dagql.ObjectResult[*ModuleSource] `field:"true" name:"blueprint" doc:"The blueprint referenced by the module source."`
+	LocalDefaults   *EnvFile                          `field:"true" name:"localDefaults" doc:"User-defined defaults read from local .env files"`
```
jedevc: Mildly worried about this being on `ModuleSource` instead of just being on `Module` (especially since it can contain client-info).

Mayyybe not a huge issue, but given that this was previously "pure", it seems a bit weird. Moving it to `Module` would mean that `ModuleSource.asModule` has to mix the client cache key in, which is also weird.

But with it being on `ModuleSource`, I *think* that means that when the either the inner or outer env is modified, we'd attempt to reload the module? None of the optimizations here would be applying anymore: 

https://github.com/dagger/dagger/blob/f25119042ef13246c26ee2d979fc6845d7e3ce05/core/schema/modulesource.go#L2416-L2421

I wanna tread carefully here, and make sure we don't accidentally cause a perf regression for module loading.

I don't really have good suggestions here, beyond splitting out the local defaults handling for inner + outer (so only inner would apply for `ModuleSource.LocalDefaults), while outer would get loaded later. Which would only really fix this for modifications to the outer, the inner could still retrigger module loading.

Maybe you've already considered this, and I'm missing the reason this is solved - but as a suggestion, we could remove `localDefaults` from the API schema. This wouldn't solve any of the perf issues, but at least would leave us free to change what's loaded where if we want to later without needing to make breaking changes.

core/envfile.go
```diff
@@ -0,0 +1,259 @@
...
+//
+//	result: `
+//	  TOKEN=topsecret
+//	  DEBUG=true
+//	`
+//
+// Note: case-insensitive search will be needed to match the resulting variables
+// against variable names
+// Note: all variables are expanded.
+func (ef *EnvFile) LookupPrefix(ctx context.Context, prefix string) (*EnvFile, error) {
```
jedevc: suggestion: `FilterPrefix`

core/envfile.go
```diff
@@ -0,0 +1,259 @@
...
+// WithVariables adds multiple environment variables to the EnvFile
+func (ef *EnvFile) WithEnvFiles(others ...*EnvFile) *EnvFile {
+	ef = ef.Clone()
+	for _, other := range others {
+		if other == nil {
+			continue
+		}
+		// Last variable assignment wins: other file's variables win over
+		// our own.
+		ef.Environ = append(ef.Environ, other.Environ...)
```
jedevc, line 67: Question - what if I have two files?

```env
A=1
```
```env
# should this file be able to access A?
B=$A
```

Or even,

```env
A=1
# should B=1? or should it be overridden when A is updated later?
B=$A
```
```env
A=2
```

IMO, variables in different files should exist in different "namespaces" - I don't see any handling for that case though.

Regardless, we should add a test for whichever behavior we decide on.

core/envfile.go
```diff
@@ -0,0 +1,259 @@
...
+	for name, value := range dotenv.AllRaw(ef.Environ) {
+		vars = append(vars, EnvVariable{Name: name, Value: value})
+	}
+	return vars, nil
+}
+
+func (ef *EnvFile) variables(ctx context.Context) (vars []EnvVariable, _ error) {
+	hostGetEnv := func(name string) string {
+		return Host{}.GetEnv(ctx, name)
+	}
```
jedevc, line 116: Is this really useful? It feels weird that the host env can interact in this way. Personally, I'd rather have users use secrets explicitly to interact with the host env if they want to do it. Curious about a usage scenario.

But if we keep it, *all dagql functions that somehow call this* need to be updated with a `dagql.CachePerClient` cache key. Otherwise, we get collisions where the client envs are different but the env files are the same, and we'd get incorrect cache collisions (which could be at some point be used by a malicious module to exfiltrate host envs).

That kinda sucks to have to do, it also means that depending on where you evaluate a given `EnvFile`, you'll get different results (e.g. passing it to a module will give different results).

core/modulesource.go
```diff
@@ -250,13 +252,172 @@ func (src *ModuleSource) Pin() string {
 	}
 }
 
+func (src *ModuleSource) innerEnvFile(ctx context.Context) (*EnvFile, string, error) {
+	// We only allow loading an env file from local modules, for safety
+	if src.Kind != ModuleSourceKindLocal {
+		return nil, "", nil
+	}
```
jedevc, line 259: Hm, this can lead to a weird scenario where a module can behave differently between local/remote.

If I have a `.env` locally that makes a required arg optional, then the API to that function will look different to the caller, depending on whether it's local or remote.

This is kinda weird. It means that if I have `./foo` that depends on `./bar`, and `bar`'s constructor takes a `greeting string`, and `./bar/.env` contains `GREETING`. If I call `foo` locally, then `dag.Foo()` takes an optional arg, but if I call it remotely, then `dag.Foo()` takes a required arg (*even if I push `.env`*).

This is a pretty big inconsistency, I'd suggest we only allow setting defaults from inner envs if they already are optional? Otherwise, you end up with modules that can be called by cloning them (or pushing them).

Don't really know - I'm a bit doubtful about the value of inner env files, but clearly they're here for a reason :tada:

jedevc (dismissed):
Nice :tada:

I'm a bit confused on the value of inner envs, but I do like them - I think there's a consistency issue there that needs addressing first.

Otherwise, it's mostly notes on consistency and caching.

core/envfile.go
shykes: I didn't initially implement it, but it was a strong user request. If we don't allow expanding system variables from `.env`, then they can't fully ditch their dagger wrapper scripts in favor of `.env`, which would be a shame since we're 99% of the way there...

shykes: Happy to change *how* we implement it if needed!

core/schema/module.go
shykes: Thanks I had missed the suggestion :)

shykes:
> Nice 🎉

Thanks! I'm actually impatient to use it for myself 😁

> I'm a bit confused on the value of inner envs, but I do like them - I think there's a consistency issue there that needs addressing first.

It's purely UX. With inner env, you can omit the module name as prefix.

So, in the case of our own `dagger/dagger` module. Let's say I want to persist a default value for `dagger call test specific --run=...`, so I can just do `dagger call test specific` and my selection of tests is applied.

- If my `.env` is *outside* the repo checkout ("outer .env") I have to set `DAGGERDEV_TEST_SPECIFIC_RUN=TestWhatever`.
- If my `.env` is *inside* the repo checkout, I can set `TEST_SPECIFIC_RUN=TestWhatever` and it's just nicer :)

The difference is even starker with simple constructor arguments, then you can even forget the env var is for Dagger at all, eg. `TOKEN=op://foo/bar`.

Since Dagger generally has a "user friction" problem, every little bit of removed friction does add up.

core/modulesource.go
shykes: The issue in your example is mostly that `.env` is applied on dependencies, but it shouldn't.

Once that's fixed (I could use some help on that btw), your scenario would unfold like this:

- You have `./foo` that depends on `./bar`
- `bars`'s constructor takes a `greeting` string
- `./bar/.env` contains `GREETING` -> but when using `bar` as a dependency, that `.env` will be ignored
- If you call `foo` locally, it will work the same as if you call it remotely, *unless you have relevant `.env` files on your local filesystem*

With that important caveat, hopefully the purpose and behavior become more clear: `.env` are for local development. As a convenience, you can have `.env` inside our outside the git checkout, but fundamentally it all works the same.

core/envfile.go
shykes: Yeah makes sense. Thanks.

core/modulesource.go
shykes: Can't comment on the performance aspect. But the reason I attached it to `ModuleSource`, is that from a client's point of view, it makes sense to load a module source, and be able to inspect the user defaults that currently apply to it. Then when loading, you can disable applying user defaults.

core/modulesource.go
jedevc: Riiiiiight! That's what you mean when you say doesn't apply to dependencies. Cool, that does resolve my point, thanks for the worked example.

I'd thought that you'd meant a top-level env would be able to modify default args deep in the dependency tree, so hadn't considered this. 

Yeah that all makes sense to me, if we have that protection then.

core/query.go
shykes: Ah yes, this is a leftover of my attempt to disable this feature for dependencies. Spoiler alert: I failed 😭

core/modfunc.go
shykes: Should I take a note for a follow-up PR to improve caching?

core/envfile.go
shykes: I had a lengthy discussion about that with ChatGPT 😅 The issue with `Filter` is that it doesn't convey that the prefix is removed from the keys, like a scoping or namespacing operation. I agree `LookupPrefix` is not great either. I was thinking of changing it to `Namespace()`

core/modulesource.go
shykes: > Riiiiiight! That's what you mean when you say doesn't apply to dependencies. Cool, that does resolve my point, thanks for the worked example.
> 
> I'd thought that you'd meant a top-level env would be able to modify default args deep in the dependency tree, so hadn't considered this.

That is indeed what I meant (and it does do that). But I think it does both?

> Yeah that all makes sense to me, if we have that protection then.

Just to confirm: I *want* that protection, but at the moment I am unable to implement it 😭

moltar:
It would be awesome if we could load not only from `.env`, but from any valid code file (specific to the used SDK), e.g. if I am using a TypeScript SDK, then load from an `env.ts` file.

shykes:
@rawkode it's finally here... you've been asking about this for so long :)

shykes:
Update: I couldn't get `--env-file` to work in nested clients, so I spun it out of this PR, since it's not a deal-breaker. Will make it a follow-up PR

rawkode:
> @rawkode it's finally here... you've been asking about this for so long :) 

I am so happy!

core/envfile.go
shykes: Renamed to `Namespace()`.

For the record, other contenders discussed with ChatGPT:

•	`FilterPrefix` – too low-level / util-like
•	`ExtractPrefixed` – wordy, feels like library helper
•	`TrimPrefixEnv` – clunky, Go-stdlib-ish not system API
•	`SelectPrefix` – precise but verbose and bland
•	`View` – generic, unclear without docs
•	`Subset` / Sub – data-structure feel, not env-specific
•	`Match` – ambiguous, sounds like boolean/regex
•	`Enter` – metaphorical, may confuse non-syseng
•	`Namespace` – ✅ chosen: domain-accurate, clear, balanced
•	`Scope` – overloaded term, compiler-centric
•	`Chroot` – clever UNIX metaphor, too insider-y
•	`Collect` – vague, implies aggregation not filtering
•	`Select` – overly generic, context missing
•	`ByPrefix` – clear but awkward/unnatural name
•	`Extract` – too broad, loses prefix meaning

core/object.go
shykes: Fixed 🙏

core/modulesource.go
shykes: `CalcDigest` doesn't return an error... I didn't want to mess with that plumbing. Is it OK to do?

shykes: For now I'm adding error checking, and if user defaults fail to load, 1) I skip them from digest calculation, 2) I log the error.

Added a FIXME to fully escalate the error. OK if I make that a follow-up?

cmd/dagger/main.go
shykes: I moved this feature to a follow-up PR... Will apply your suggestion there

shykes: Applied the suggestion in the follow-up PR: https://github.com/dagger/dagger/pull/11167

core/modfunc.go
jedevc: Yeah, let's chat about this one later, ideally once @sipsma is back, I suspect there's subtlety I'm not grasping here either.

core/modulesource.go
jedevc: Argh, the problem is, `Variables` really shouldn't be making host calls. It's not just the error, it's the fact that that's now client-dependent.

Moving this discussion to this thread https://github.com/dagger/dagger/pull/11034#discussion_r2394149518, since it's kinda linked.

core/envfile.go
jedevc: ARGH, this is really frustrating.

My initial take is that we should *temporarily* make envfiles not able to access host variables. I'm happy to work on adding it back, but my worry is that we spend ages talking about how to fix it and make it play happy in cache now - when honestly, pretty much everything else is ready to merge.

Here's my reasoning, if we keep it as is, *everywhere* we call `Variables` needs to have `dagql.CachePerClient` - which (finger-in-the-air guess) is gonna actively hurt performance. The weird error handling is kind of a symptom of this.

If this is absolutely absolutely required to merged in this initial version (and iterating quickly in follow-ups is not an option), then I can spend some time on this, but I'd prefer to ship something useful, and improve it over time.

---

My suggestion for how to handle this properly:
- When calling `asEnvFile`, specify a `withHost` property. If set to `true`, then envs in the current host would be allowed during interpretation. *However* this is different than before because now only the cache key of `asEnvFile` needs to be special, we would then load *all* the host env vars at that point of call into the engine. Is that a security issue? Dunno. It actually has cache coherency though.

core/modulesource.go
jedevc: Mm yeah, fair.

I *think* that the `UserDefaults` on `ModuleSource` should only include the inner env. The outer env should ideally be loaded and handled later.

Or, if we can't do that now, then every schema handler that calls `LoadUserDefaults` is gonna need to be `dagql.CachePerClient` (since it accesses a client).

core/envfile.go
shykes: > When calling asEnvFile, specify a withHost property. If set to true, then envs in the current host would be allowed during interpretation. However this is different than before because now only the cache key of asEnvFile needs to be special, we would then load all the host env vars at that point of call into the engine. Is that a security issue? Dunno. It actually has cache coherency though.

This seems easy to do, you're saying if I do this, the caching issue related to host env goes away?

jedevc: Yeah. We'll still need to set a custom cache key - if `withHost` is set, then we mixin the Client ID.

So, something like how `CacheAsRequested` works: https://github.com/dagger/dagger/blob/8c6516e3257650b055360a7a19672a674fa34714/dagql/cachekey.go#L92-L101

shykes: @jedevc actually I just realized what you meant - getting *all* system env variables upfront. This makes me laugh because we were just talking about your recent change to getting *all* git refs from a remote server upfront 😁 Hopefully we don't move on to uploading the whole host filesystem upfront 😛 (jk)

I do think uploading all env variables upfront is weird - has a smell to it. Would trigger unexpected performance or security issues in some edge cases.

I'll spin it out into a follow-up... But I do have questions :)

shykes:
This looks like a legitimate test failure... https://dagger.cloud/dagger/traces/3a0f35ffcb5847a40c5bae36330d3eda?span=a2ef985d225c652e

`TestCall/TestByName/local_ref_with_@`

To repro locally: `dagger call -m github.com/shykes/dagger@local-defaults test specific --run=TestCall/TestByName/local_ref_with_@`

core/envfile.go
shykes: @jedevc I disabled the feature for now

marcofranssen:
I'm mostly using .envrc files used by direnv. Maybe just relying on exported envars is sufficient and leaving the rest to tools like direnv.

Or at least support both .env and .envrc

shykes:
> I'm mostly using .envrc files used by direnv. Maybe just relying on exported envars is sufficient and leaving the rest to tools like direnv.
> 
> Or at least support both .env and .envrc

Supporting `.envrc` seems like a can of worms, and feature creep on Dagger's part. Our starting point was a Dagger-specific config file, something like `dagger.env.json` or something like that. But it seemed like a nice convenience to piggyback on a pre-existing standard instead. The extreme simplicity of `.env` was well-suited.

Expanding into `.envrc` and other more complex systems, with dependencies and edge cases, seems like a recipe for disaster.

I'll think about the idea of also supporting straight system env variables. With the scaffolding in place, this could be done safely I think. I just need to think through the UX implications - for example there's an elevated risk of accidental false positives. (eg. your dagger module accepts an argument `token` and you happen to have a variable `TOKEN` in your system environment, so they match, but you did not expect or want them to.

In the short term, we're going to merge `.env` support only, see how people use and how many problems it can solve. Then discuss our next move.

I understand how you would prefer to do everything with your favorite environment config (in your case `.envrc`), but realistically you're always going to deal with tool-specific config files. Think of this `.env` as a Dagger-specific config file.

sipsma:
> This looks like a legitimate test failure... https://dagger.cloud/dagger/traces/3a0f35ffcb5847a40c5bae36330d3eda?span=a2ef985d225c652e
> 
> `TestCall/TestByName/local_ref_with_@`
> 
> To repro locally: `dagger call -m github.com/shykes/dagger@local-defaults test specific --run=TestCall/TestByName/local_ref_with_@`

It seems like quite a few tests besides that one are failing. From ones I looked at it seems like they are failing during `dagger init` because the code is trying to read the inner env file and failing (due to the fact that the directory wasn't created yet):
```
load user defaults: failed to check for inner env file in /work/test@test: select: failed to get snapshot: failed to receive stat message: rpc error: code = NotFound desc = get full root path: rpc error: code = NotFound desc = eval symlinks: lstat /work/test@test: no such file or directory
```

This chunk of code looks wrong: https://github.com/dagger/dagger/blob/e577d90024dfea531db2ca3cf0b2ce499ae349eb/core/modulesource.go#L291-L301

It unconditionally returns an error when `err != nil` but then attempts to handle certain errors.

marcofranssen:
> > I'm mostly using .envrc files used by direnv. Maybe just relying on exported envars is sufficient and leaving the rest to tools like direnv.
> > Or at least support both .env and .envrc
> 
> Supporting `.envrc` seems like a can of worms, and feature creep on Dagger's part. Our starting point was a Dagger-specific config file, something like `dagger.env.json` or something like that. But it seemed like a nice convenience to piggyback on a pre-existing standard instead. The extreme simplicity of `.env` was well-suited.
> 
> Expanding into `.envrc` and other more complex systems, with dependencies and edge cases, seems like a recipe for disaster.
> 
> I'll think about the idea of also supporting straight system env variables. With the scaffolding in place, this could be done safely I think. I just need to think through the UX implications - for example there's an elevated risk of accidental false positives. (eg. your dagger module accepts an argument `token` and you happen to have a variable `TOKEN` in your system environment, so they match, but you did not expect or want them to.
> 
> In the short term, we're going to merge `.env` support only, see how people use and how many problems it can solve. Then discuss our next move.
> 
> I understand how you would prefer to do everything with your favorite environment config (in your case `.envrc`), but realistically you're always going to deal with tool-specific config files. Think of this `.env` as a Dagger-specific config file.

Honestly I think `.env` vs `.envrc` there is not much difference.

`.env` is usually used with the tool `dotenv` while `.envrc` is usually used with the tool `direnv`. Both tools add the variables to your shell exports. From a format perspective in .envrc I have to explicitly add the `exports MY_VAR` while in `.env` I can just define it as `MY_VAR`.

Working with dagger and other tools in our projects we solely rely on the shell exports for things to work the way we want. (direnv helps us loading and unloading the variables when switching directories) to not have side effects as you described above.

I think for a better integration with other tooling and projects it would be a huge benefit if this was less opiniated and just treat these things as env vars.

At the moment I therefore do Makefiles to dynamically compose the dagger commands based on shell exports being available or not. Which makes things very verbose and from a UX perspective massively long commands to pass in all the secret variables and such.

Why not pickup all env vars with a convention like `DAGGER_MYMODULE_VAR`

shykes:
> > This looks like a legitimate test failure... https://dagger.cloud/dagger/traces/3a0f35ffcb5847a40c5bae36330d3eda?span=a2ef985d225c652e
> > `TestCall/TestByName/local_ref_with_@`
> > To repro locally: `dagger call -m github.com/shykes/dagger@local-defaults test specific --run=TestCall/TestByName/local_ref_with_@`
> 
> It seems like quite a few tests besides that one are failing. From ones I looked at it seems like they are failing during `dagger init` because the code is trying to read the inner env file and failing (due to the fact that the directory wasn't created yet):
> 
> ```
> load user defaults: failed to check for inner env file in /work/test@test: select: failed to get snapshot: failed to receive stat message: rpc error: code = NotFound desc = get full root path: rpc error: code = NotFound desc = eval symlinks: lstat /work/test@test: no such file or directory
> ```
> 
> This chunk of code looks wrong:
> 
> https://github.com/dagger/dagger/blob/e577d90024dfea531db2ca3cf0b2ce499ae349eb/core/modulesource.go#L291-L301
> 
> It unconditionally returns an error when `err != nil` but then attempts to handle certain errors.

Sorry about that @sipsma ... This looks like residue from my trying to find an answer to this: https://github.com/dagger/dagger/pull/11034#discussion_r2389481025.

shykes:
> Working with dagger and other tools in our projects we solely rely on the shell exports for things to work the way we want. (direnv helps us loading and unloading the variables when switching directories) to not have side effects as you described above.
> 
> I think for a better integration with other tooling and projects it would be a huge benefit if this was less opiniated and just treat these things as env vars.

The problem is there no such thing as "less opinionated". Your approach of relying on system variables + a tool to manage them is by no means a universally accepted method. The ecosystem is highly fragmented. For example if we *only* supported system env variables, someone would make the exact same argument as you, in reverse: "why not support `.env`"?

> At the moment I therefore do Makefiles to dynamically compose the dagger commands based on shell exports being available or not. Which makes things very verbose and from a UX perspective massively long commands to pass in all the secret variables and such.

Once this PR is merged, you have another option, which is to write a `.env`. I understand it's not your favorite option, you would prefer to manage everything with direnv. But just wanted to clarify that you don't *have* to write a Makefile wrapper or run super-long commands every time. You just prefer the inconvenience of a Makefile wrapper to the inconvenience of writing a `.env`.

> Why not pickup all env vars with a convention like `DAGGER_MYMODULE_VAR`

Yes we might do that, as I said before. I need to think about the implications. But, it won't be a universal solution either, because not everyone uses system env vars to configure their tools.

shykes:
> This chunk of code looks wrong:
> 
> https://github.com/dagger/dagger/blob/e577d90024dfea531db2ca3cf0b2ce499ae349eb/core/modulesource.go#L291-L301

I removed the offending error checking code... But still getting failed tests, investigating further.

core/modfunc.go
sipsma: Invalidating cache for all function calls in the module when an unrelated default is changed would indeed be an issue, but agree that shouldn't be a blocker here. Can easily follow up with more refined cache key logic later

PR #11019, opened 2025-09-10
feat: mount client ~/.config/dagger/ca-certificates to engine

jedevc:
Simply allows configuration via the same mechanism that we propagate `~/.config/dagger/engine.json` by.

Issue #9007, opened 2024-11-20
Session config file


jedevc:
This is something that has come up quite a bit recently, mostly in the context of https://github.com/dagger/dagger/pull/8800#issuecomment-2479499749, but also around https://github.com/dagger/dagger/issues/8766#issuecomment-2474813161.

Even when we have a way to configure *engine* settings, there are a collection of user configurations that would make sense to apply per client-session, without requiring a whole engine reload. For example:
- Registry settings - today these are all only in the engine config, but it makes it very difficult to make changes (requiring an engine restart), and also you might want different settings per project you're on. It also means if you're using a shared engine, the shared engine has to be configured with potentially project-specific registries.
- Build resources/limits (from https://github.com/dagger/dagger/issues/4196) - for the same reasons.
- Network configurations (including DNS, CAs, private git clones, etc).

I think we should have a different config file for these - it's annoying to have a different file, but the other alternatives don't feel right:
- `engine.json`/`engine.toml` isn't the right place, since it's global to the engine.
- `dagger.json` isn't the right place, since those are committed to git, and these settings are potentially machine specific - they *might* be committed to git, but they shouldn't *need* to be.
- Config environment variables in the client session could work, but these are trickier to document and use (no `.schema.json` for example). Also more difficult to copy around and manage.
- Explicit flags to every dagger command will end up *very* repetitive (and potentially even have confidential data then in the case of network configs).

So I'm proposing a `dagger.settings.json` (let the bikeshedding begin, no idea what the actual file should be). There are two potential places to load this from - one in `~/.config/dagger/dagger.settings.json`, next to `engine.json` from https://github.com/dagger/dagger/pull/8800, and another next to `dagger.json` in your dagger module. We would try and load from both. This allows you to have machine specific, and also project specific session configs.

From the core-api POV, we could introduce a `Engine.configure` API, which would take optional args that would configure the entire session it's been made from (similar in style to `Module.serve`, it modifies session state).

grouville:
I suppose that this is where we would implement the mounting logic of the `.gitconfig`, for session-scoped git rewrite rules: https://github.com/dagger/dagger/issues/7787. Basically, the clean implementation to solve all those issues

shykes:
1. Agree we need a config file for Dagger that is environment specific, in other words: not `dagger.json`, and not checked into the app repo. Somewhere in the home directory makes sense.

[...]

PR #11040, opened 2025-09-12
engine: support registry mirror configs on engine.json

### LINKED ISSUE #4217: Support client-side registry configuration of CA+Mirrors
# Background

Users often need CA certs or mirror configuration in order to push/pull container images.

Right now, the only way to accomplish that is to customize the engine image and install those certs in the right directory + setup a configuration file for mirrors. This is often extremely inconvenient.

In many cases the user has the required CA certs locally or has access to the correct mirror configuration. They just need a way to tell our engine to use them.

It thus makes sense to allow this to be configured by clients and then passed to the engine server rather than forcing it to be a part of engine setup prereqs. 

Related: 
* https://github.com/dagger/dagger/issues/2142
* https://github.com/dagger/dagger/issues/4105
* https://github.com/dagger/dagger/issues/2016

# Options
## A) Part of Core API
We can put this explicitly in the core API.

Pros:
1. Each SDK picks it up for free via codegen. Only the CLI has to have any actual implementation. It's thus also available to raw graphql clients too.
2. Centralizing configuration in our schema as much as possible keeps things more consistent for users
3. Gives lots of flexibility to users (settings can be per request) but there's also opportunities to keep boilerplate low (see proposal below)
4. Interesting possible integration points with extensions (see proposal below)

Cons:
1. Doesn't inherently encode a "standard way" of picking up config from the environment, which may be what users actually want most often.
   * You could implement this (i.e. you could read serialized settings from a file on the host an then include that in your queries), but it would not be "standardized"
   * Queries thus become more specific to individual clients (already true due to the existence of the host api, but this takes us further down that road)

### Strawman schema proposal

Fine grained API:
```graphql
input RegistryConfig {
   // each string is the actual contents of the rootCA cert, not a path to it
   RootCAs: [String!]
   // a list of mirrors to use for the provided address
   Mirrors: [String!]
}

type Container {
 from(address: String!, registryConfig RegistryConfig): Container!
 ...
 publish(address: String!, platformVariants: [ContainerID!], registryConfig RegistryConfig): String!
}
```

The fine grained approach would give the ultimate flexibility at the cost of boilerplate.

Easier to use might be a "wrapper" API (though not mutually exclusive with above):
```graphql
extend type Query {
  withDefaultRegistryConfig(registryHost: String!, registryConfig: RegistryConfig!) Query!
}
```

`withDefaultRegistryConfig` could be very interesting in combination with extensions; you could imagine a company that wants to enforce use of certain ca+mirror configuration vends out an extension that users call to pick up all the necessary configs.

## B) Part of CLI Configuration
We could make these settings something you configure via the CLI. They would then get persisted to a file under `~/.config/dagger/` or something. When you use `dagger run` or an SDK uses `dagger session` they would get picked up automatically

Pros:
1. The code written by users in SDKs is detached from these settings. You can run the same SDK code with different CA+Mirror settings without changing the code necessarily

Cons:
1. Less flexible, essentially operates as unchangeable global state.
2. The CA and mirror config must exist as strings persisted in the local filesystem
   * Misses out on possible compelling use cases where, e.g., the settings get configured by an extension

## Preferred Option
After writing this out I'm realizing these options are actually not contradictory at all. In fact if they both existed they might complement one another pretty nicely. Many possibilities but you could imagine that:
1. The CLI allows you configure defaults for registry config, which would actually map in the implementation to all queries just being wrapped in `withDefaultRegistryConfig`
   * I think there's a ton of possibilities for re-using this pattern elsewhere outside these registry settings (cache config, networking, etc.)
3. Additionally, users who prefer being explicit in their code (or really need per-query settings rather than global ones) can also use the explicit API in their queries

So, I actually think we should consider doing both options in the fullness of time. Proposal for an order of operations:
1. Add support for configuring these settings in the CLI only
2. In the future when more compelling use cases like the one mentioned involving extensions arise, also add `withDefaultRegistryConfig` to the core api schema that exposes the same config settings
3. If ever needed, can also add support for individual registry config settings in `Container.from` and `Container.publish` (though I think use cases that call for that might be obscure)

# Implementation

Implementation-wise, I think we should strive for something that is not specific to any one provisioner target. That means it would be best avoid reliance on e.g. docker bind mounts since that requires dagger to be running in docker (and the daemon to be local).

I think the best approach would be the upstream change suggested here: https://github.com/moby/buildkit/issues/3366

That would just let us re-use the existing interface between CLI<->Runner.

A fallback would be creating our own separate channel/API between CLI<->Runner (free to do now since we control the engine image). But that's probably not worth it unless we run into problems upstreaming it.

matipan:
Currently registry mirrors can only be configured on `engine.toml`. This requires users to have to manually provision dagger engines with a buildkit-specific config. In this PR we add support to configuring registries in `engine.json` directly. With this change users can add their configs on `~/.config/dagger/engine.json` and the dagger cli (e.g `dagger call`) will automatically pick it up when provisioning the engine.

This PR implements only the first step that was described in #4217: CLI only config. There's an active discussion on whether this configuration should be engine wide or scoped to the individual session. The work for that will come in a separate PR. This is the first step that allows us to at least be able to configure it.

Fixes #4217.

### DIFF
--- a/.changes/unreleased/Added-20250912-163622.yaml
+++ b/.changes/unreleased/Added-20250912-163622.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: 'engine: support registry mirror configs on `engine.json`'
+time: 2025-09-12T16:36:22.510006635-03:00
+custom:
+    Author: matipan
+    PR: "11040"
--- a/core/integration/registry_mirrors_test.go
+++ b/core/integration/registry_mirrors_test.go
@@ -35,7 +35,7 @@ func (EngineSuite) TestRegistryMirrorsCustomCA(ctx context.Context, t *testctx.T
 		engineWithConfig(ctx, t, func(ctx context.Context, t *testctx.T, cfg config.Config) config.Config {
 			return config.Config{
 				Registries: map[string]config.RegistryConfig{
-					"ghcr.io": {
+					"docker.io": {
 						Mirrors: []string{"testreg:5000"},
 					},
 					"testreg:5000": {
@@ -62,7 +62,7 @@ func (EngineSuite) TestRegistryMirrorsHTTP(ctx context.Context, t *testctx.T) {
 	engine := devEngineContainer(c, engineWithConfig(ctx, t, func(ctx context.Context, t *testctx.T, cfg config.Config) config.Config {
 		return config.Config{
 			Registries: map[string]config.RegistryConfig{
-				"ghcr.io": {
+				"docker.io": {
 					Mirrors: []string{"testreg:5000"},
 				},
 				"testreg:5000": {
@@ -86,7 +86,7 @@ func testImagePull(ctx context.Context, t *testctx.T, c *dagger.Client, devEngin
 	require.NoError(t, err)
 	t.Cleanup(func() { c2.Close() })
 
-	out, err := c2.Container().From("ghcr.io/dagger/engine").WithExec([]string{"echo", "hello"}).Stdout(ctx)
+	out, err := c2.Container().From("alpine:3.22.1@sha256:4bcff63911fcb4448bd4fdacec207030997caf25e9bea4045fa6c8c44de311d1").WithExec([]string{"echo", "hello"}).Stdout(ctx)
 	require.NoError(t, err)
 	require.Equal(t, "hello", strings.TrimSpace(out))
 
@@ -96,5 +96,5 @@ func testImagePull(ctx context.Context, t *testctx.T, c *dagger.Client, devEngin
 		WithExec([]string{"cat", "/cache/logs/registry.log"}).
 		Stdout(ctx)
 	require.NoError(t, err)
-	require.True(t, strings.Contains(out, "GET /v2/dagger/engine"))
+	require.True(t, strings.Contains(out, "GET /v2/library/alpine"))
 }
--- a/docs/current_docs/reference/configuration/engine.mdx
+++ b/docs/current_docs/reference/configuration/engine.mdx
@@ -345,9 +345,38 @@ redirected to a different one.
 <Tabs groupId="config">
 <TabItem value="engine.json">
 
-:::warning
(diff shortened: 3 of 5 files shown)


shykes:
@matipan to parahphrase [my question in the issue](https://github.com/dagger/dagger/pull/11019#issuecomment-3286559345): since we agree that in the future this should be moved from an engine-wide config to a session config, is there a config file other than *engine.json* that we could use now, to avoid a breaking config change later?

jedevc:
x-linking https://github.com/dagger/dagger/issues/9007.

Are we renaming `engine.json` entirely? Or are we splitting the functionality across multiple files (which you seemed to dislike back in https://github.com/dagger/dagger/issues/9007#issuecomment-2524329261).

If we're renaming the file, my suggestion is to pick a new name (throwing out `config.json` as a potential option, or even just have a full-on config directory, load all of the JSON files in there) - attempt to read both, but update our docs to point to the new one. Some settings are read per-session, some are per-engine lifetime, but over time, we try and move the implementation to be more per-session.

matipan:
@shykes as we discussed in yesterday's meeting, there is no other config file we could use. That said, we wouldn't be introducing breaking changes in the future.

Even if we introduce a new session based `config.json` I still think we should keep `engine.json` and allow users to configure things at an engine level. For developers on their machines having a way to configure engines without needing to reprovision it is great. But I'd argue its also great for engine operators to provide an already configured engine that does not need their users providing configs that are applicable to the entire organization. If we provide this configs as part of `engine.json` as is in this PR, then when we introduce a new session-level config we wouldn't break any API. Users would still be able to provide configuration on engine provisioning, but they would also be able to provide session-based config for things applicable to the session.

shykes:
@matipan I don't think that's how it will go, but we can discuss that separately. Since there is only one config file for now, let's go ahead and merge as-is 👍

As for future evolution of config files, we have some debating to do..

PR #11041, opened 2025-09-12
avoids starting service twice when used by different clients

### LINKED ISSUE #11014: 🐞 Services are not de-duplicated correctly
### What is the issue?

Mounting a service into another container before returning it, creates 2 instances of the service in the engine, and the exposed tunnel moves intermittently between instances of the same service.

For more context:
https://discord.com/channels/707636530424053791/1414607480117530656

### Dagger version

0.18.3

### Steps to reproduce

```golang
func (m *Tests) ServiceTest(ctx context.Context) (*dagger.Service, error) {
        // Define a database service
    db := dag.
        Container().
        From("postgres:14").
        WithExposedPort(5432).
        WithEnvVariable("POSTGRES_PASSWORD", "postgres").
        AsService(dagger.ContainerAsServiceOpts{
            UseEntrypoint: true,
        })

        // Initialize some schema in the database
    init, err := dag.Container().From("postgres:14").
        WithEnvVariable("CACHE_BUSTER", time.Now().String()).
        WithServiceBinding("db", db).
        WithEnvVariable("PGPASSWORD", "postgres").
        WithEnvVariable("PG_HOST", "db").
        WithExec([]string{"sh", "-c", "psql -h db -U postgres -c 'CREATE TABLE test (id SERIAL PRIMARY KEY, name TEXT)'"}).
        WithExec([]string{"echo", "Database is ready: %s", time.Now().String()}).
        Stdout(ctx)
    if err != nil {
        return nil, err
    }
    fmt.Println("Initialized: " + init)

        // Return the service
    return db, nil
}
```
Then, expose the service with the CLI:
```
dagger call service-test up
```
And try to hit the database querying for that table just created:
```
$> psql -h 127.0.0.1 -U postgres -d postgres -c "SELECT * FROM test;"
 id | name
----+------
(0 rows)

$> psql -h 127.0.0.1 -U postgres -d postgres -c "SELECT * FROM test;"
ERROR:  relation "test" does not exist
LINE 1: SELECT * FROM test;
```

### Log output

_No response_

marcosnils:
this commit fixes the issue where running `dagger call svc up` only runs
the service once even if used by different clients.

Fixes #11014

### DIFF
--- a/core/integration/services_test.go
+++ b/core/integration/services_test.go
@@ -695,6 +695,8 @@ func (m *Caller) Run(ctx context.Context) (string, error) {
 		WithDefaultArgs([]string{"sh", "-c", "echo started >> /cache/svc.txt && while true; do wc -l /cache/svc.txt && sleep 5; done"}).
 		AsService()
 
+	// we're passing the service to a different module that calls "up" on it so it's a different client
+	// which is what triggers the behavior this test is testing.
 	dag.Starter().Start(ctx, svc)
 
 	dag.Container().From("alpine").WithServiceBinding("db", svc).
--- a/core/integration/services_test.go
+++ b/core/integration/services_test.go
@@ -644,6 +644,76 @@ func (m *Caller) Run(ctx context.Context) error {
 	require.NoError(t, err)
 }
 
+func (ServiceSuite) TestServiceTunnelStartsOnceForDifferentClients(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	out, err := goGitBase(t, c).
+		WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+		WithWorkdir("/work/starter").
+		With(daggerExec("init", "--source=.", "--name=starter", "--sdk=go")).
+		WithNewFile("main.go", `package main
+
+import (
+	_ "embed"
+	"context"
+
+	"dagger/starter/internal/dagger"
+)
+
+type Starter struct{}
+
+
+func (m *Starter) Start(ctx context.Context, s *dagger.Service) {
+	go func() {s.Up(ctx)}()
+}
+`,
+		).
+		WithWorkdir("/work/caller").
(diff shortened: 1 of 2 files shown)


core/integration/services_test.go
```diff
@@ -644,6 +644,76 @@ func (m *Caller) Run(ctx context.Context) error {
...
+		WithEnvVariable("CACHE", time.Now().String()).
+		WithExec([]string{"truncate", "-s", "0", "/cache/svc.txt"}).Sync(ctx)
+
+	svc := dag.Container().From("alpine").
+		WithMountedCache("/cache", svcCache).
+		WithExposedPort(8080, dagger.ContainerWithExposedPortOpts{ExperimentalSkipHealthcheck: true}).
+		WithDefaultArgs([]string{"sh", "-c", "echo started >> /cache/svc.txt && while true; do wc -l /cache/svc.txt && sleep 5; done"}).
+		AsService()
+
+	dag.Starter().Start(ctx, svc)
```
jedevc, line 700: I'm guessing we need to do this to simulate starting it from another client (in this case a module)?

Because this is a bit of a tricky test case, I'd make sure it's annotated, while I can remember how it works *now*, that probably won't stay the case :sweat_smile:

core/service.go
```diff
@@ -777,7 +777,7 @@ func (svc *Service) startTunnel(ctx context.Context) (running *RunningService, r
 		return nil, fmt.Errorf("failed to get buildkit client: %w", err)
 	}
 
-	upstream, err := svcs.Start(svcCtx, svc.TunnelUpstream.ID(), svc.TunnelUpstream.Self(), true)
+	upstream, err := svcs.Start(svcCtx, svc.TunnelUpstream.ID(), svc.TunnelUpstream.Self(), svc.TunnelUpstream.Self().TunnelUpstream.Self() != nil)
```
jedevc, line 780: we love a one-line fix.

Issue #11055, opened 2025-09-15
`Wolfi.container()` is very slow in our CI


shykes:
Since last week (approximately), `github.com/dagger/dagger/modules/wolfi | container`  takes a *long* time in the context of our CI: at least 1mn.

To reproduce, run any function in our CI that requires building a wolfi base image. For example:

```console
dagger -m github.com/dagger/dagger/cmd/dagger | cli
```

Specifically the custom span `GetPackageWithDependencies` is where all the time goes. Note: this custom span is not in our source code, it's in the underlying wolfi tool.

[Sample trace](https://dagger.cloud/dagger/traces/985abe511efe59e5f1828b57728a38a4)

PR #11057, opened 2025-09-15
modules/alpine: upgrade apko dependency

sipsma:
Fixes https://github.com/dagger/dagger/issues/11055

We were on a pretty old version. We were getting what appeared to be slow/cpu-intensive apk solver steps (GetPackagesWithDependencies), which smelled an awful lot like an inefficient algorithm that might have been made noticeable by recent changes to our pinning.

Looks like a simple bump of apko fixes it. Most likely one of these two changes we pick up fix it:
* https://github.com/chainguard-dev/apko/pull/1406
* https://github.com/chainguard-dev/apko/pull/1317

There appears to have been at least one somewhat odd backwards incompatible change (maybe regression?) where key file names returned by the new dep have `%40` instead of `@`, but it's a trivial thing to workaround on our end.


sipsma:
Looks like the time spent on `GetPackagesWithDependencies` is consistently 0.0s again ([example](https://dagger.cloud/dagger/traces/5d8c7c66e887977e5da0c630f1e33d93?span=10e910ad98de7b7b)), so this should be good

shykes (approved):
Yup, fixed it for me!

Repro:

```console
dagger -m github.com/sipsma/dagger/cmd/dagger -c binary
```

--> [Resulting trace](https://dagger.cloud/dagger/traces/af22107ab93e2b014dd3a1df327c118f?span=7aba01c1da92123a) (zoomed to the offending span)

Screenshot 2025-09-15 at 13 14 33

PR #11058, opened 2025-09-15
don't use gitAutoIgnore on Host.File

sipsma:
It doesn't make sense to use gitAutoIgnore when an exact file path is being requested to be loaded. It also broke the
`./hack/with-dev go test ...` workflow since the `bin/` dir in the repo is .gitignored, meaning the dev dagger bin couldn't be loaded.


shykes:
Makes complete sense @sipsma.

Just double-checking, do we still have consensus that *even for directories* we shouldn't auto-ignore either? Basically we should only apply gitignore specifically when loading a module, and not when loading *arguments* to a module's functions. Everyone still on board with that?

sipsma:
> Just double-checking, do we still have consensus that even for directories we shouldn't auto-ignore either? Basically we should only apply gitignore specifically when loading a module, and not when loading arguments to a module's functions. Everyone still on board with that?

I'm not personally up to date on that thread after being out last week, so not sure where everyone is at. I guess I can see that being a bit more of a grey area vs. this situation where it seems pretty clear cut that if you specify an exact file path to load it doesn't make sense to skip via .gitignore.

The implementation of `Host.Directory` currently has a `noGitAutoIgnore` arg (default false), so changing the behavior should just be a matter of flipping that to be `gitAutoIgnore` (default false) if we wanna change it before the release.

cc @TomChv @jedevc

jedevc:
See https://github.com/dagger/dagger/pull/11017. 

Would like a review from one more person before merging, but this would supersede this PR, by disabling it by default everywhere.

sipsma:
> See #11017.
> 
> Would like a review from one more person before merging, but this would supersede this PR, by disabling it by default everywhere.

Ah I hadn't seen that yet! Just approved it, I'll close this one out in favor of that 👍

PR #11052, opened 2025-09-15
chore: update engine dependencies 

TomChv:
Replace https://github.com/dagger/dagger/pull/11046.

This PR updates engine dependencies and resolves dependency breaking changes such as https://github.com/dagger/dagger/pull/10659#issuecomment-3018788339

**TODO**
- [x] Updated all non-AI dependencies in `go.mod` and `sdk/go/go.mod`
- [x] Update gqlgen
- [x] Update otel
- [x] Update AI dependencies
- [x] Update remaining deps

**Notes**

- To update containerd we will need to update our fork first.
- There's some otel dependencies (like `go.opentelemetry.io/otel/log`) pinned to `v0.12.2`, I'm not sure if I should update them or not -> Updated to `v0.14.0`

PR #11059, opened 2025-09-16
chore: update go to 1.25.1

jedevc:


TomChv:
Quick note, this PR isn't bumping our own `go.mod`, this will be done in a follow up and also update our `golangci` at the same time otherwise it will trigger failures.

See: https://github.com/dagger/dagger/pull/11052#discussion_r2355205722

```
level=info msg="golangci-lint has version 1.64.5 built with go1.24.0 from 0a603e49 on 2025-02-13T21:19:55Z"
level=info msg="[config_reader] Used config file ../.golangci.yml"
Error: can't load config: the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.0)
Failed executing command with error: can't load config: the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.0)
```

The migration requires is a bit of work since rules are applied differently, I already started but this will come after this PR and #11052

PR #11060, opened 2025-09-16
docs: Add PHP cookbook recipe for SSH cloning

vikram-dagger:


docs/current_docs/cookbook/snippets/clone-git-repository-ssh/php/src/MyModule.php
```diff
@@ -0,0 +1,33 @@
...
+use Dagger\Socket;
+
+use function Dagger\dag;
+
+#[DaggerObject]
+class MyModule
+{
+    // Demonstrates an SSH-based clone requiring a user-supplied sshAuthSocket.
+    #[DaggerFunction]
+    public function cloneWithSsh(string $repository, string $ref, Socket $sock): Container
```
charjr, line 20:
```suggestion
use Dagger\Attribute\DaggerObject;
use Dagger\Attribute\Doc;
use Dagger\Container;
use Dagger\Socket;

use function Dagger\dag;

#[DaggerObject]
class MyModule
{
    #[DaggerFunction]
    #[Doc('Demonstrates an SSH-based clone requiring a user-supplied sshAuthSocket.')]
    public function cloneWithSsh(string $repository, string $ref, Socket $sock): Container
```

The PHP SDK doesn't parse doc block comments currently.
It only adds docs to dagger with the Doc attribute.

PR #11064, opened 2025-09-16
feat: add changeset native export

jedevc:
Follow up to @vito's #10946.

This adds a new `Changeset.export` field that exports, handling the deleted files - this removes the need for the specialized client handling.

### DIFF
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -38,6 +38,7 @@ var (
 
 const (
 	Directory     string = "Directory"
+	Changeset     string = "Changeset"
 	Container     string = "Container"
 	File          string = "File"
 	Secret        string = "Secret"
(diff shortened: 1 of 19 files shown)


cmd/dagger/functions.go
```diff
@@ -651,13 +652,13 @@ func handleResponse(ctx context.Context, dag *dagger.Client, returnType *modType
...
-		return handleChangesetResponse(ctx, dag, response)
-	}
-
 	// Handle the `export` convenience, i.e, -o,--output flag.
 	switch returnType.Name() {
+	case Changeset:
+		if outputPath == "" {
+			return handleChangesetResponse(ctx, dag, response)
+		}
+		fallthrough
```
vito, line 661: 🍂

engine/client/filesync.go
```diff
@@ -145,6 +145,18 @@ func (t FilesyncTarget) DiffCopy(stream filesync.FileSend_DiffCopyServer) (rerr
...
 	}
 
+	for _, removePath := range opts.RemovePaths {
+		if !filepath.IsAbs(removePath) {
+			removePath = filepath.Join(opts.Path, removePath)
+		}
+		if err != nil {
+			return fmt.Errorf("get full remove path: %w", err)
+		}
+		if err := os.RemoveAll(removePath); err != nil {
```
vito: Can this just be an `os.Remove`? These should always be file paths, I'm a little afraid of somehow ending up with `"/"` or `"."` or `""` here.

Maybe if the path ends with `/` we could do `RemoveAll`, since that's at least more explicit? (Except if it's just `"/"` haha.)

docs/docs-graphql/schema.graphqls
```diff
@@ -192,6 +192,12 @@ type Changeset {
   """The older/lower snapshot to compare against."""
   before: Directory!
 
+  """Applies the diff represented by this changeset to a path on the host."""
```
vito, line 195: 👍

vito (approved):
Thanks!

PR #11070, opened 2025-09-16
dedupe concurrent dagql calls within same session

sipsma:
Previously on this journey, we disabled deduping of concurrent dagql calls and the related "edge-merging" feature of buildkit. Motivation being that we did not have any handling of a number of corner cases like a client canceling their request that got deduped with other clients.

We are still leaving buildkit edge merging off forever, but this change reintroduces de-duping to the dagql cache with the new twist that it only occurs within calls that belong to the same session.

Now if a single session has two concurrent calls that are the same and still running, they will be deduped. If two separate sessions are making the same concurrent call, they will still run separately (and one will "win" in terms of what's written to cache).

The isolation to a single session avoids the previous problems with client cancelation.

**The benefit is more predictable and improved performance.** In particular, I noticed while working on something else that the way we load module deps in parallel and recursively was resulting in a lot of overlapping concurrent calls in the same session (using the dagger repo as an example, which has a complicated module dependency dag).

The change here gets rid of that duplication. This practically speaking only appears to shave ~2s off module loading time for me. However, it makes it much easier to reason about how often various pieces of that code will run (at most once, per-session), which in turn makes the whole thing easier to think about and less delicate in terms of future changes impacting performance.

The module loading example is just one, I'm sure there's other real-world use cases that result in concurrent overlapping dags being solved by a single client; any code that does solves in parallel has a decent chance of benefitting.


dagql/objects.go
```diff
@@ -655,18 +656,32 @@ func (r ObjectResult[T]) call(
...
+	cacheKey := cache.CacheKey[CacheKeyType]{
+		ResultKey: string(newID.Digest()),
+	}
+	if doNotCache {
+		cacheKey.ResultKey = ""
+	}
+	// dedupe concurrent calls only if the ID digest is the same and if the two calls are from the same session
+	// we don't want to dedupe across sessions since:
+	// 1. it creates problems when one session closes and others were waiting on the result
+	// 2. it makes it easy to accidentally leak session specific information that isn't yet precisely scoped in the ID
```
jedevc: Is client id instead of session id preferable?

For example, I could imagine a case where a nested client (like a function call) could close early, etc.

engine/sources/local/localfs.go
```diff
@@ -463,8 +464,11 @@ func (local *localFS) toRootPath(path string) string {
 //
 // Unlike other methods below, we don't need to verifyExpectedChange since there was no change applied to the path.
 func (local *localFS) GetPreviousChange(ctx context.Context, path string, stat *types.Stat) (CachedChange, error) {
-	rootPath := local.toRootPath(path)
-	return local.changeCache.GetOrInitialize(ctx, rootPath, func(_ context.Context) (*ChangeWithStat, error) {
+	cacheKey := cache.CacheKey[string]{
+		ResultKey:      local.toRootPath(path),
+		ConcurrencyKey: cacheConcurrencyKey,
+	}
+	return local.changeCache.GetOrInitialize(ctx, cacheKey, func(_ context.Context) (*ChangeWithStat, error) {
```
jedevc: Could maybe do this with a helper, like `cache.ConcurrentCacheKey(local.toRootPath)`.

dagql/objects.go
sipsma: True, will swap that out

Issue #10818, opened 2025-07-30
Add `deprecated` directive for fields/functions/args


jedevc:
This will allow module authors to mark parts of their API as deprecated (currently not possible).

For example, in go:

```go
package main

import (
	"context"
	"dagger/foo/internal/dagger"
)

type Foo struct {
    MyDeprecatedField string // +deprecated
}

// +deprecated
type MyAdditionalType struct {
    X string
}

// +deprecated
func (foo *Foo) MyDeprecatedFunc() {
}

func (foo *Foo) MyOtherFunc(]
    myDeprecatedArg string // +deprecated
) {
}
```

> [!NOTE]
>
> This requires that we allow "annotating" things that we've not previously been annotating - currently our directives/decorators are really only supported on arguments + fields.

Then, this `Deprecated` value needs to appear on every relevant type def: arguments, types, enums, etc.

This should then be propagated to the graphql schema - with the `@deprecated` directive.

jedevc:
@TomChv @helderco does this make sense in python/ts? Happy to tackle the go+engine parts, but will probably need a hand for the others.

helderco:
Yes, it's doable. We can have a `@dagger.deprecated("reason")` decorator for that. Or just `@dagger.function(deprecated="reason")`.

PR #11071, opened 2025-09-16
sdks: Allow module level `deprecation` directive

grouville:
Fixes https://github.com/dagger/dagger/issues/10818

This allows module creators to deprecate fields, args, enum values, functions and types (module types). The codegen and the engine will catch-up with those.

TODO:
- [x] Engine binding(s)
- [x] Go SDK
- [x] Integration test suite
- [ ] Typescript SDK (in progress)
- [ ] Python SDK


grouville:
cc @TomChv a first pass on the ts sdk could be done. It's ready (i think)

cmd/codegen/generator/go/templates/module_enums.go
```diff
@@ -135,9 +139,31 @@ func (ps *parseState) parseGoEnum(t *types.Basic, named *types.Named) (*parsedEn
...
+			comment = strings.TrimSpace(docComment)
+			maps.Copy(pragmas, docPragmas)
+		}
+		if val, ok := astSpec.(*ast.ValueSpec); ok && val.Comment != nil {
+			linePragmas, lineComment := parsePragmaComment(val.Comment.Text())
+			if comment == "" {
+				comment = strings.TrimSpace(lineComment)
+			}
+			maps.Copy(pragmas, linePragmas)
+		}
```
TomChv, line 157: I don't understand this code block, you get the pragma from the `docForAstSpec` no?

What this code block for? What happens if it override the previous map value?

core/integration/module_test.go
```diff
@@ -5964,6 +5964,289 @@ export class Dep {
...
+)
+
+// Reference the enum so it appears in the schema.
+func (m *Test) UseMode(mode Mode) Mode {
+	return mode
+}
+`
+	const tsSrc = `import { field, func, object } from "@dagger.io/dagger"
+
+  /** @deprecated This module is deprecated and will be removed in future versions. */
```
TomChv, line 6265: Does it also support multi line like?

```
// My cool doc
// +deprecated="of course its deprecated"
```

```
/*
* My normal doc
* @deprecated hello that's deprecated
*/
````

File: sdk/typescript/src/module/introspector/typescript_module/ast.ts [hunk 3]
```diff
@@ -20,6 +20,11 @@ export type ResolvedNodeWithSymbol<T extends keyof DeclarationsMap> = {
   file: ts.SourceFile
 }
 
+export type SymbolDoc = {
+  description: string
+  deprecated: string
```
TomChv: I think `deprecated` should be an optional value instead of always be set.

```
export type SymbolDoc = {
  description: ""
  deprecated?: string
}
```

So that if that flag doesn't exist, it's clearly represented in the code.
Otherwise function like:

```
/*
* @deprecated 
*/
class Deprecated


class NotDeprecated
```

Will bot have their field `deprecated` set to `""` so you have a false positive.

(That's also simpler for tests because you wouldn't need to update every fields with an empty `deprecated: ""`)

I'll add comments to where you should propagate the optional.

sdk/typescript/src/module/introspector/test/testdata/deprecatedFunction/expected.json
```diff
@@ -0,0 +1,42 @@
...
+          "deprecated": "",
+          "arguments": {},
+          "returnType": {
+            "kind": "STRING_KIND"
+          }
+        },
+        "current": {
+          "name": "current",
+          "description": "",
+          "deprecated": "",
```
TomChv: That deprecated shouldn't exist here for example (see my comments in `ast.ts`)

sdk/typescript/src/module/introspector/test/testdata/deprecatedField/expected.json
```diff
@@ -0,0 +1,69 @@
...
+          "deprecated": "Inline note that this field is going away.",
+          "type": {
+            "kind": "STRING_KIND"
+          },
+          "isExposed": true
+        },
+        "latestField": {
+          "name": "latestField",
+          "description": "",
+          "deprecated": "",
```
TomChv: Shouldn't exist (see comments in `ast.ts`)

sdk/typescript/src/module/introspector/dagger_module/typeObjectProperty.ts
```diff
@@ -22,6 +22,7 @@ export class DaggerObjectTypeProperty
 {
   public name: string
   public description: string
+  public deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/introspector/dagger_module/typeObjectProperty.ts
```diff
@@ -36,7 +37,9 @@ export class DaggerObjectTypeProperty
     super(node)
 
     this.name = symbol.name
-    this.description = this.ast.getDocFromSymbol(this.symbol)
+    const { description, deprecated } = this.ast.getSymbolDoc(this.symbol)
```
TomChv, line 40: Should return (see my comments in `ast.ts`)

```
{
  description: string
  deprecated?: string
}
```

sdk/typescript/src/module/introspector/dagger_module/typeObject.ts
```diff
@@ -33,6 +33,7 @@ import {
 export class DaggerTypeObject extends Locatable implements DaggerObjectBase {
   public name: string
   public description: string
+  public deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/introspector/dagger_module/property.ts
```diff
@@ -21,6 +21,7 @@ export class DaggerProperty
 {
   public name: string
   public description: string
+  public deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/introspector/dagger_module/objectBase.ts
```diff
@@ -8,6 +8,7 @@ import { References } from "./reference.js"
 export interface DaggerObjectPropertyBase extends Locatable {
   name: string
   description: string
+  deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/introspector/dagger_module/objectBase.ts
```diff
@@ -22,6 +23,7 @@ export type DaggerObjectPropertiesBase = {
 export interface DaggerObjectBase extends Locatable {
   name: string
   description: string
+  deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/introspector/dagger_module/object.ts
```diff
@@ -26,6 +26,7 @@ import { References } from "./reference.js"
 export class DaggerObject extends Locatable implements DaggerObjectBase {
   public name: string
   public description: string
+  public deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/introspector/dagger_module/interfaceFunction.ts
```diff
@@ -19,6 +19,7 @@ export type DaggerInterfaceFunctions = {
 export class DaggerInterfaceFunction extends Locatable {
   public name: string
   public description: string
+  public deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/introspector/dagger_module/function.ts
```diff
@@ -18,6 +18,7 @@ export type DaggerFunctions = { [name: string]: DaggerFunction }
 export class DaggerFunction extends Locatable {
   public name: string
   public description: string
+  public deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/introspector/dagger_module/enumClass.ts
```diff
@@ -16,6 +16,7 @@ export class DaggerEnumClassValue
   public name: string
   public value: string
   public description: string
+  public deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/introspector/dagger_module/enumBase.ts
```diff
@@ -4,6 +4,7 @@ export interface DaggerEnumBaseValue extends Locatable {
   name: string
   value: string
   description: string
+  deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/introspector/dagger_module/enum.ts
```diff
@@ -12,6 +12,7 @@ export class DaggerEnumValue extends Locatable implements DaggerEnumBaseValue {
   public name: string
   public value: string
   public description: string
+  public deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/introspector/dagger_module/argument.ts
```diff
@@ -19,6 +19,7 @@ export type DaggerArguments = { [name: string]: DaggerArgument }
 export class DaggerArgument extends Locatable {
   public name: string
   public description: string
+  public deprecated: string
```
TomChv: Should be optional `deprecated?: string`

sdk/typescript/src/module/entrypoint/register.ts
```diff
@@ -122,11 +143,17 @@ export class Register {
...
    */
   addFunction(fct: Method | DaggerInterfaceFunction): Function_ {
-    return dag
+    let fn = dag
       .function_(fct.alias ?? fct.name, addTypeDef(fct.returnType!))
       .withDescription(fct.description)
       .withSourceMap(addSourceMap(fct))
       .with(this.addArg(fct.arguments))
+
+    if (fct.deprecated) {
```
TomChv: Should be `if (fct.deprecated !== undefinied)` so we can also register function with empty deprecation reason like `/* @deprecated */`
(See my comments on `ast.ts`)

sdk/typescript/src/module/entrypoint/register.ts
```diff
@@ -140,6 +167,10 @@ export class Register {
           sourceMap: addSourceMap(arg),
         }
 
+        if (arg.deprecated) {
+          opts.deprecated = arg.deprecated
+        }
```
TomChv: Just `opts.deprecated = args.deprecated` if `args.deprecated` can be undefined

sdk/typescript/src/module/entrypoint/register.ts
```diff
@@ -85,11 +100,17 @@ export class Register {
...
       Object.values(enum_.values).forEach((value) => {
-        typeDef = typeDef.withEnumMember(value.name, {
+        const memberOpts: TypeDefWithEnumMemberOpts = {
           value: value.value,
           description: value.description,
           sourceMap: addSourceMap(value),
-        })
+        }
+
+        if (value.deprecated) {
```
TomChv: Same here, if `value.deprecated` is optional, you could directly assign it since `memberOpts.deprecated` type is `string | undefined`

sdk/typescript/src/module/entrypoint/register.ts
```diff
@@ -56,13 +65,19 @@ export class Register {
       // Register all fields that belong to this object
       Object.values(object.properties).forEach((field) => {
         if (field.isExposed) {
+          const fieldOpts = {
+            description: field.description,
+            sourceMap: addSourceMap(field),
+          } as TypeDefWithFieldOpts
+
+          if (field.deprecated) {
+            fieldOpts.deprecated = field.deprecated
```
TomChv: Same here

sdk/typescript/src/module/entrypoint/register.ts
```diff
@@ -42,11 +45,17 @@ export class Register {
...
-      // Register the class Typedef object in Dagger
-      let typeDef = dag.typeDef().withObject(object.name, {
+      const objectOpts: TypeDefWithObjectOpts = {
         description: object.description,
         sourceMap: addSourceMap(object),
-      })
+      }
+
+      if (object.deprecated) {
+        objectOpts.deprecated = object.deprecated
```
TomChv: Same here

core/schema/module.go
```diff
@@ -282,6 +292,7 @@ func (s *moduleSchema) typeDefWithObject(ctx context.Context, def *core.TypeDef,
 	Name        string
 	Description string `default:""`
 	SourceMap   dagql.Optional[core.SourceMapID]
+	Deprecated  string `default:""`
```
TomChv: Should be optional too

TomChv (commented):
Thanks for that PR! This will be a great addition 🚀 

One important thing to fix on your implementation is the difference between having an empty deprecation reason and no deprecation at all.

I went heavy on the TS comment so you can see exactly what's wrong but it also applies for the engine.
You do it correctly on Go though 😃 

I think the new `Deprecated` argument in the schema should be optional so:
- `""` means it's deprecated without reason (so it should be marked as deprecated in the generated binding without additional doc
- `"xxxx"` `means it's deprecated` with a reason
- `nil` means no deprecatation

Is the current codegen implementation ignores deprecated if it's `""`? If so it seems like an incorrect behaviour and should be fixed.

core/integration/module_test.go
grouville: Yes, I updated the test `context` to prove it

core/schema/module.go
grouville: In order to do that, I also had to change the dagql representation of the DeprecatedReason. (see associated commit, starting with `dagql(*)`. It is actually more compliant with the graphQL spec

core/schema/module.go
```diff
@@ -131,6 +131,12 @@ func (s *moduleSchema) Install(dag *dagql.Server) {
 				dagql.Arg("description").Doc(`The doc string to set.`),
 			),
 
+		dagql.Func("withDeprecated", s.functionWithDeprecated).
+			Doc(`Returns the function with the given deprecated string.`).
+			Args(
+				dagql.Arg("deprecated").Doc(`If deprecated, the reason or migration path.`),
```
TomChv: I do not understand the argument name, should it be `reason` for clarity?
Why `If deprecated`? You should call `withDeprecated` only if you want to deprecate a function no?

core/schema/module.go
```diff
@@ -282,6 +292,7 @@ func (s *moduleSchema) typeDefWithObject(ctx context.Context, def *core.TypeDef,
 	Name        string
 	Description string `default:""`
 	SourceMap   dagql.Optional[core.SourceMapID]
+	Deprecated  *string
```
TomChv, line 315: Optional argument doesn't need pointer, they use dagql wrapper: `dagql.Optional[dagql.String]`

Example: https://github.com/dagger/dagger/blob/f46ef8e7bb683ea17154e332b99897dff28efe62/core/schema/directory.go#L437

TomChv, line 315:
```suggestion
	Deprecated  dagql.Optional[dagql.String]
```

core/schema/module.go
```diff
@@ -313,6 +324,7 @@ func (s *moduleSchema) typeDefWithObjectField(ctx context.Context, def *core.Typ
 	TypeDef     core.TypeDefID
 	Description string `default:""`
 	SourceMap   dagql.Optional[core.SourceMapID]
+	Deprecated  *string
```
TomChv, line 347:
```suggestion
	Deprecated  dagql.Optional[dagql.String]
```

core/schema/module.go
```diff
@@ -384,29 +396,31 @@ func (s *moduleSchema) typeDefWithEnumValue(ctx context.Context, def *core.TypeD
 	Value       string
 	Description string `default:""`
 	SourceMap   dagql.Optional[core.SourceMapID]
+	Deprecated  *string
```
TomChv, line 419:
```suggestion
	Deprecated  dagql.Optional[dagql.String]
```

core/schema/module.go
```diff
@@ -384,29 +396,31 @@ func (s *moduleSchema) typeDefWithEnumValue(ctx context.Context, def *core.TypeD
...
-	return def.WithEnumValue(args.Value, args.Value, args.Description, sourceMap)
+	return def.WithEnumValue(args.Value, args.Value, args.Description, args.Deprecated, sourceMap)
 }
 
 func (s *moduleSchema) typeDefWithEnumMember(ctx context.Context, def *core.TypeDef, args struct {
 	Name        string
 	Value       string `default:""`
 	Description string `default:""`
 	SourceMap   dagql.Optional[core.SourceMapID]
+	Deprecated  *string
```
TomChv, line 433:
```suggestion
	Deprecated  dagql.Optional[dagql.String]
```

core/schema/module.go
```diff
@@ -466,6 +480,12 @@ func (s *moduleSchema) functionWithDescription(ctx context.Context, fn *core.Fun
 	return fn.WithDescription(args.Description), nil
 }
 
+func (s *moduleSchema) functionWithDeprecated(ctx context.Context, fn *core.Function, args struct {
+	Deprecated *string
```
TomChv:
```suggestion
	Deprecated dagql.Optional[dagql.String]
```

core/schema/module.go
```diff
@@ -474,6 +494,7 @@ func (s *moduleSchema) functionWithArg(ctx context.Context, fn *core.Function, a
 	DefaultPath  string    `default:""`
 	Ignore       []string  `default:"[]"`
 	SourceMap    dagql.Optional[core.SourceMapID]
+	Deprecated   *string
```
TomChv, line 517:
```suggestion
	Deprecated   dagql.Optional[dagql.String]
```

core/schema/module.go
grouville: What's the difference ? https://discord.com/channels/707636530424053791/1003718839739105300/1423084110494306314

TomChv: Fair enough, it just felt more consistent to have `dagql.Optional` everywhere than a mix.

TomChv (commented):
LGTM on Typescript!

cmd/codegen/generator/go/templates/module_enums.go
grouville: yeah, we already get the doc comment via `docForAstSpec`, but that only covers the block comment that’s shared across every const in the `const (...)` group.

The `inline // +dagger:`  stuff you put at the end of a single enum member lives in `ValueSpec.Comment`, so without that block we’d never see per-value pragmas.

 _Example:_
```go
  // +deprecated="use Zeta" // (docForAstSpec)
  const (
      ModeAlpha Mode = "alpha" // +deprecated // (ValueSpec.Comment)
      ModeZeta  Mode = "zeta"
  )
```

The doc comment hits both members, but that trailing `// +deprecated` applies only to `ModeAlpha`.
 
If someone were to write such enum implementation, as we only deprecate enum members, the inline enum (without the text) wins. We could make `use Zeta` the global deprecation text if set, but I personally find it hard to unify across SDKs.

I'd prefer a UX where: you don't deprecate the enum itself, you just deprecate its members (as that's what we do under the hood). AND, if you don't specify a text, then it's the default graphQL enum error that gets triggered

core/schema/module.go
grouville: Agreed, thank you 🙏 Updated

grouville:
> LGTM on Typescript & Go implementation! There's still my little comment about the argument of `WithDeprecated`.
> 
> Feel free to ping me when you're done with Python so I can also review it properly :)

Python has been updated and all above comments have been addressed / replied, I'll try to do the interfaces tonight

grouville:
The `Core tests / test-cli-engine (pull_request)` failure seems consistent, but I am unable to repro it locally:

```shell
./hack/with-dev go test -v -run "TestEngine/TestLocalCacheGC/sweep_automaticGC=false" ./core/integration -timeout=180s
```

EDIT: it was a flake

cc @TomChv @eunomie It is ready to be reviewed

core/moddeps.go
```diff
@@ -175,7 +175,9 @@ func (d *ModDeps) lazilyLoadSchema(ctx context.Context, hiddenTypes []string) (
 	}
 
 	// add any extensions to objects for the interfaces they implement (if any)
+	// fmt.Fprintf(os.Stderr, "🧪 objects: %d, interfaces: %d\n", len(objects), len(ifaces))
```
eunomie, line 178: nit: debug log

core/moddeps.go
```diff
@@ -175,7 +175,9 @@ func (d *ModDeps) lazilyLoadSchema(ctx context.Context, hiddenTypes []string) (
...
 	for _, objType := range objects {
+		// fmt.Fprintf(os.Stderr, "🔍 object=%s\n", objType.typeDef.Name)
```
eunomie, line 180: nit: debug log

core/typedef.go
```diff
@@ -602,7 +613,7 @@ func (typeDef *TypeDef) WithOptional(optional bool) *TypeDef {
 	return typeDef
 }
 
-func (typeDef *TypeDef) WithObjectField(name string, fieldType *TypeDef, desc string, sourceMap *SourceMap) (*TypeDef, error) {
+func (typeDef *TypeDef) WithObjectField(name string, fieldType *TypeDef, desc string, sourceMap *SourceMap, deprecated *string) (*TypeDef, error) {
```
eunomie, line 616: nit: should we keep `sourceMap` as the last argument to align it with all others?


```suggestion
func (typeDef *TypeDef) WithObjectField(name string, fieldType *TypeDef, desc string, deprecated *string, sourceMap *SourceMap) (*TypeDef, error) {
```

core/schema/module.go
```diff
@@ -347,7 +359,7 @@ func (s *moduleSchema) typeDefWithObjectField(ctx context.Context, def *core.Typ
 	if err != nil {
 		return nil, err
 	}
-	return def.WithObjectField(args.Name, fieldType.Self(), args.Description, sourceMap)
+	return def.WithObjectField(args.Name, fieldType.Self(), args.Description, sourceMap, args.Deprecated)
```
eunomie, line 362: nit: (see comment in `object.go`) should we keep the `sourceMap` arg as the last one to be aligned with other functions?

core/schema/module.go
```diff
@@ -554,7 +575,7 @@ func (s *moduleSchema) functionWithArg(ctx context.Context, fn *core.Function, a
 		td = td.WithOptional(true)
 	}
 
-	return fn.WithArg(args.Name, td, args.Description, args.DefaultValue, args.DefaultPath, args.Ignore, sourceMap), nil
+	return fn.WithArg(args.Name, td, args.Description, args.DefaultValue, args.DefaultPath, args.Ignore, sourceMap, args.Deprecated), nil
```
eunomie, line 578: nit: same, should we keep `sourceMap` as the last arg?

cmd/codegen/generator/go/templates/src/_types/object.go.tmpl
```diff
@@ -46,6 +46,9 @@ type {{ $field | FieldOptionsStructName }} struct {
 	//
 	{{- end }}
 	// Default: {{ $arg.DefaultValue }}
+	{{- end }}
+	{{- if $arg.IsDeprecated }}
+{{ $arg.DeprecationReason | FormatDeprecation }}
```
eunomie, line 51: nit: indent?

core/integration/module_test.go
```diff
@@ -6155,6 +6156,716 @@ func (m *Test) PrintDefault(ctx context.Context) (string, error) {
...
+)
+class LegacyRecord:
+    note: str = dagger.field(
+        deprecated="This field is deprecated and will be removed in future versions."
+    )
+
+
+@dagger.enum_type
+class Mode(enum.Enum):
+    """Mode is deprecated; use zeta instead."""
```
eunomie, line 6369: Am I right thinking this line has no impact? We can only deprecate enum values, not enum, right?

grouville: Yes, we do register it, but i'll remove it 🙏 It's a leftover from my registration test (we do parse it atm), we don't do anything with it

grouville:
@eunomie doing the nits as a follow-up 🙏

PR #11073, opened 2025-09-17
feat: ensure commit checkouts have tag history

jedevc:
:cry: sigh.

We were never testing this behavior, but as of https://github.com/dagger/dagger/pull/11030, I wanted to start relying on it.

We technically had subtly different behavior for checking out refs and checking out SHAs directly; the former included tags in the history, the latter didn't.

> [!WARNING]
>
> Technically this is mergeable, but bleh, *using the tags like this is vulernable to wild consistency issues*. Need to think through the edge cases, probably using https://github.com/dagger/dagger/pull/11032 as a helper.
> - `doFetch` needs to get reset if a tag is updated remotely
> - If a `ls-remote` is captured, a tag is added, and then it's added/removed as part of the tag history for when we fetch it... we should be able to respect that? Or detect that? Don't really know.
>   - Simple tags could just be rewritten!
>   - Annotated tags... harder.

### DIFF
--- a/core/git.go
+++ b/core/git.go
@@ -23,6 +23,7 @@ import (
 	bkcache "github.com/dagger/dagger/internal/buildkit/cache"
 	bkclient "github.com/dagger/dagger/internal/buildkit/client"
 	"github.com/dagger/dagger/internal/buildkit/executor/oci"
+	"github.com/dagger/dagger/internal/buildkit/identity"
 	bksession "github.com/dagger/dagger/internal/buildkit/session"
 	"github.com/dagger/dagger/internal/buildkit/snapshot"
 	"github.com/dagger/dagger/internal/buildkit/solver/pb"
@@ -462,6 +463,8 @@ func (repo *RemoteGitRepository) mount(ctx context.Context, depth int, refs []Gi
 				}
 			}
 
+			// TODO: should set doFetch if a tag in ls-remote has been updated?
+
 			if doFetch {
 				fetchRefs = append(fetchRefs, ref)
 			}
@@ -708,11 +711,6 @@ func (ref *RemoteGitRef) Tree(ctx context.Context, srv *dagql.Server, discardGit
 			}
 			checkoutGit := git.New(gitutil.WithWorkTree(checkoutDir), gitutil.WithGitDir(checkoutDirGit))
 
-			_, err = checkoutGit.Run(ctx, "-c", "init.defaultBranch=main", "init")
-			if err != nil {
-				return err
-			}
-
 			err := doGitCheckout(ctx, checkoutGit, ref.repo.URL.Remote(), gitURL, ref.FullRef, ref.Commit, depth, discardGitDir)
 			if err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


core/integration/git_test.go
```diff
@@ -592,6 +593,71 @@ func (GitSuite) TestGitTags(ctx context.Context, t *testctx.T) {
...
+		})
+
+		t.Run(desc+"/tag", func(ctx context.Context, t *testctx.T) {
+			requireGitTagsExist(t, git.Tag("v0.12.0"))
+		})
+
+		t.Run(desc+"/commit", func(ctx context.Context, t *testctx.T) {
+			// v0.12.0 => 133917c6f9ce36d8cfdc595d9b7bd2c14cbc2c20
+			requireGitTagsExist(t, git.Commit("133917c6f9ce36d8cfdc595d9b7bd2c14cbc2c20"))
+		})
```
jedevc, line 631: This is the case that was previously failing.

jedevc:
Think we can "safely" take this, will check my thoughts tomorrow.

This just makes the behavior consistent. Happy to punt on a decision if this behavior is *desirable*.

core/git.go
```diff
@@ -796,6 +802,12 @@ func doGitCheckout(
 			return fmt.Errorf("failed to set remote origin to %s: %w", remoteURL, err)
 		}
 	}
+	if strings.HasPrefix(destref, "refs/dagger.tmp/") {
+		_, err = checkoutGit.Run(ctx, "update-ref", "-d", destref)
```
sipsma, line 809: very nit, please ignore if non-issue: we exit earlier than this in various points after making the temp ref, in which case it will still exist. I'm *guessing* the cost of doing so is extremely minimal and thus that doesn't matter? But double checking in case it's worth trying to always call `update-ref -d` so it's never left around

core/integration/git_test.go
```diff
@@ -592,6 +593,71 @@ func (GitSuite) TestGitTags(ctx context.Context, t *testctx.T) {
 	})
 }
 
+func (GitSuite) TestGitCheckedTags(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	requireGitTagsExist := func(t *testctx.T, git *dagger.GitRef) {
+		ctr := dag.Container().
```
sipsma: nit: using `dag` instead of `c` here, so missing out on the opts set by `connect` (might not matter)

jedevc: Oops, typo.

core/git.go
jedevc: Updated the comments to be more clear here - we're always checking out into an empty new git dir in a new snapshot.

The only case we exit early here is in an error case, in which case the snapshot is never actually returned. So no need to delete it.

jedevc:
Yeah, okay, let's take this - it just makes the behavior consistent, which is *definitely* better than before.

Not gonna release note it or anything though, I'm still actually undecided as to whether we want this functionality. At least this stops it from *accidentally* breaking in the future.

PR #11076, opened 2025-09-17
sdk(typescript): deprecate `@enumType()`

grouville:
Since https://github.com/dagger/dagger/pull/10632 and following our [recent docs changes](https://docs.dagger.io/extending/enumerations), the preferred way to use enums in Typescript is through the native `enum` type.

However, the legacy `@enumType` was never marked as deprecated.

This removes all occurences to the previous pattern and introduces a `legacyEnym` test for retrocompatibility


TomChv (approved):
Changes LGTM, could you also add a test in `core/integration/legacy_test.go` just to verify the end-to-end execution?

I give an early approval since the CI is green, make sure you add the tests before merging :)

core/integration/legacy_test.go
```diff
@@ -1172,6 +1172,38 @@ export class Test {
 	}
 }
 
+func (LegacySuite) TestLegacyTypescriptEnumDecorator(ctx context.Context, t *testctx.T) {
```
jedevc, line 1175: Add a little note here - something like "Deprecated in dagger/dagger#11076".

jedevc (approved):

jedevc:
If you can hold merge until CI is passing (think it's an infra issue we were discussing earlier)

Issue #10618, opened 2025-06-20
sdk/java: implement telemetry


eunomie:
Telemetry allows to communicate all information about the execution of a workflow. This is the way to communicate calls to other modules for instance.
Those information will be rendered in both the CLI and [Dagger cloud](https://dagger.cloud).

The telemetry layer has not been yet implemented in the Java SDK.

## Example

To better understand the issue, let's take an example using two modules, one in Go (with the telemetry integration) and one in Java (without telemetry)

Both are using a [dependent module](https://github.com/eunomie/reddit-summarizer/tree/main/glow).

Here is the Go code:

```go
func (m *MyModule) ContainerEcho(stringArg string) *dagger.Container {
	return m.Base().WithExec([]string{"echo", stringArg})
}

func (m *MyModule) Base() *dagger.Container {
	return dag.Container().From("alpine:latest")
}

func (m *MyModule) Print(ctx context.Context, stringArg string) (string, error) {
	str, err := m.ContainerEcho(stringArg).Stdout(ctx)
	if err != nil {
		return "", err
	}
	return dag.Glow().DisplayMarkdown(ctx, str)
}
```

And if we run `dagger -c 'print "Hello **dagger**"'` this is what we can see on Dagger cloud:

<img width="845" alt="Image" src="https://github.com/user-attachments/assets/6f6625c6-1e50-49bb-a52d-143d06aa10a9" />

We can see that `.print` will contain more spans, with the container call and `glow` (the dependent module) call.

Now this time in Java. Code is similar:

```java
  @Function
  public Container containerEcho(String stringArg) {
    return base().withExec(List.of("echo", stringArg));
  }

  @Function
  public Container base() {
    return dag().container().from("alpine:latest");
  }

  @Function
  public String print(String stringArg) throws ExecutionException, DaggerQueryException, InterruptedException {
    String str = containerEcho(stringArg).stdout();
    return dag().glow().displayMarkdown(str);
  }
```

But the result is way different:

<img width="1162" alt="Image" src="https://github.com/user-attachments/assets/ced5ddc5-f8c0-4e35-bf17-1f37c55424cd" />

We can't see any detail. We only see that the Java code has been running and the output, but not how this has been done.

The telemetry should solve this missing part.

## Existing implementations

Telemetry is using [Open Telemetry](https://opentelemetry.io/).
As references here are the implementation:
- in TypeScript SDK: https://github.com/dagger/dagger/tree/main/sdk/typescript/src/telemetry
[...]

PR #11083, opened 2025-09-18
[#10618] DT - added telemetry to java sdk

robertonav20:
Hi
here the suggestion to implement Telemetry to Java SDK
let me know if we should review something.

I would test it directly with dagger cloud, is possible during execution of github actions?

Thanks in advance

Main PR https://github.com/dagger/dagger/pull/10831


robertonav20:
Hi @eunomie 
this is the new PR cleaned

eunomie:
Hi @robertonav20 
Sorry, I'm really late :-(

Yes you can test it with dagger cloud, I think that's the easiest way. It can be seen through GitHub Actions, for instance during tests related to the java SDK, but it might be harder to deal with them.

How I'm testing it:
- generate a module and use the generated template
    - inside my `dagger/dagger` working copy
    - `dagger init --sdk=../../sdk/java modules/my-java-mod`
- inside `modules/my-java-mod`:
    - `dagger call container-echo --string-arg Hello stdout -w`

This will create a module using the local version of the Java SDK (only works if there's no modification outside of `sdk/java`, if not you have to build a new engine), call a function and open the traces on cloud.

This is what the traces looks like if you expand the `.containerEcho` call:
image

In comparison, the same call but on a module generated using `--sdk=go`:
image

You can see that we are seeing the details of the operation, with the `container.from.withExec`. That what should be visible once the telemetry is sending all the traces.

robertonav20:
Hi @eunomie 
ok understood, so probably on my tests I initialized module with a wrong sdk.
As I see in you screenshot it seems is working correctly

eunomie:
> As I see in you screenshot it seems is working correctly

No, it's not working.
In the first screenshot, with Java, we only see the log output.
In the second screenshot, with Go SDK, you can see the `container`, `.from` and `.withExec` blocks. They are the ones we'd like to also see from Java SDK.

robertonav20:
Sorry I'm focusing on wrong part of screenshot 😅, so at this point have you an idea of what is missing or what is wrong? Because I followed the typescript implementation and I think I covered all parts starting from telemetry.ts

Could depends on attributes not attached to span?

robertonav20:
Hi @eunomie,
i found the issues

- VertxGraphQLClient didn't inherit trace context
- Input function call didn't pass to active span below the result

Here the link to to public trace https://dagger.cloud/dagger-demo/traces/b956c81b3321e105982b36ac4841f690?listen=ffc5313a999bd5da&listen=ac3189d859e5cf7f if you would analyze it deeply

{B619D269-500E-438A-8E46-0138AFC16247}

robertonav20:
Hi,
i realized which custom trace creation was useless because span is sent with `VertxGrapQLClient`

Below the new trace and the link https://dagger.cloud/dagger-demo/traces/f5e7ba337f9ed909cf21076bc047a200?listen=98dbd3186fcbd91e&listen=74d50c54abe89029
{5BF70436-7C94-41CA-B73B-111E5052349C}

robertonav20:
Hi @eunomie,
i fixed broken tests and changed telemetry logs level.

eunomie:
Hi @robertonav20 
Sorry I've been a bit busy on other topics.
I just tried it right now, and it just works fine, that's really nice!

Right side of the screenshot is `v0.19.3`, left side is this branch where we can see the extra entries (the details of what is done behind `container-echo` in this case)
image

I'm reviewing the code so we can merge it asap.

sdk/java/dagger-java-sdk/src/main/java/io/dagger/client/engineconn/Connection.java
```diff
@@ -76,14 +78,24 @@ public static Connection get(String workingDir) throws IOException {
 
   private static Connection getConnection(int port, String token, Optional<CLIRunner> runner) {
     Vertx vertx = Vertx.vertx();
+    // Inject Dagger Cloud token
     String encodedToken =
         Base64.getEncoder().encodeToString((token + ":").getBytes(StandardCharsets.UTF_8));
-    DynamicGraphQLClient dynamicGraphQLClient =
+
+    // LOG.info("graphql headers {}", headers);
```
eunomie: nit: debug code

robertonav20:
Hi @eunomie ,
thanks for the review, so what should I do
- remove java test module, I added because I didn't see any module java based for this reason I added too, but is fine
- remove commented debug
- open telemetry initialization logs removal

These changes are correct?

eunomie (approved):

eunomie:
> remove java test module, I added because I didn't see any module java based

Yes that's true, we don't have java module there. The `modules/` ones are modules we are using, not test modules.
There's some test modules in integration tests, under `core/integration/testdata`.
Maybe that could be worth to add one test there (in `module_java_test.go`) to ensure that works and we will not have regression, but that could be a follow up for sure.

Thanks a lot @robertonav20 for the feature 👍

robertonav20:
Ok thanks for your explanation!

PR #11092, opened 2025-09-18
Improved pinning for git modules

jedevc:
:warning: currently broken!

1. When a module is pinned, we lose track of what the original ref is. This means dependencies that use contextual git will get the wrong ref.
2. When calling a module in CI (like [here](https://github.com/dagger/dagger/blob/09b87500ac9883eda96cc38470ce22a084e235ab/.github/main.go#L20-L27)), you can choose between either addressing it via ref (and risk it getting updated in between), or addressing it via sha (and losing the ref info).

Both of these need fixing!


core/integration/git_test.go
```diff
@@ -1139,6 +1139,56 @@ func (GitSuite) TestRemoteUpdates(ctx context.Context, t *testctx.T) {
 	require.Contains(t, entries, "abc")
 }
 
+func (GitSuite) TestRemoteUpdatesFrozenTag(ctx context.Context, t *testctx.T) {
+	// similar to above, upstream changes between fetches, but we check that
+	// the tag gets *resolved* and doesn't get updated
```
jedevc, line 1144: This is the reason I'd like to get this fix in for the release:
1. It's a longstanding issue, we've never really handled it
2. It's a *potentially* breaking change, since the fix is to *always* fetch by commit instead of by ref name (since that ref is already resolved). Some git servers may not support this - I think these are very very few and far between though.

sipsma (approved):

PR #11099, opened 2025-09-19
Use released engine in CI

kpenfound:
As discussed, this removes the dev engine from our CI steps and adds one small new step that uses the dev engine to make sure it can load our CI module.

TODO
- [x] Merge https://github.com/dagger/dagger-for-github/pull/187
- [x] Tag dagger-for-github v8.2.0
- [ ] Refactor the `dagger call` steps to use dagger-for-github instead of the script. Update `callDaggerStep` in `modules/gha/steps.go`. This could also be done in a follow-up PR to keep this small


.github/main.go
```diff
@@ -107,6 +107,12 @@ func New() *CI {
 			"Helm",
 			"check --targets=helm",
 		).
+		withWorkflow(
+			ci.AltRunner,
+			true,
+			"Dev Engine",
+			"check --targets=SDKs",
```
jedevc:
```suggestion
			"check --targets=sdk",
```

This currently isn't running any checks (see https://dagger.cloud/dagger/traces/dc58508c934fd43873a2a4ad91403989).

.github/main.go
```diff
@@ -107,6 +107,12 @@ func New() *CI {
 			"Helm",
 			"check --targets=helm",
 		).
+		withWorkflow(
```
jedevc, line 110: If we change this so that it's *instead* doing something like `dev | with-exec ...`, then we can remove all the dev install steps as well!

Food for thought, we could do this as a follow-up as well.

modules/gha/steps.go
```diff
@@ -21,7 +21,13 @@ func (j *Job) checkoutStep() api.JobStep {
 
 func (j *Job) installDaggerSteps() []api.JobStep {
 	steps := []api.JobStep{
-		j.bashStep("install-dagger", map[string]string{"DAGGER_VERSION": j.DaggerVersion}),
```
jedevc, line 24: nit: can we remove `install-dagger.sh`?

.github/main.go
matipan: That makes sense. I started testing it and there's a few things not working. I'm happy to do a follow up with it so that we can get this merged this week!

modules/gha/steps.go
matipan: Done!

.github/main.go
matipan: Fixed!

jedevc:
Hm, noticing [in the trace](https://dagger.cloud/dagger/traces/14ad0b5f31a1c324349718da5eab2140?listen=92ba2d4a11d016ce&listen=ef86199a1dadf06f&listen=5fadad13eb146119&listen=f02e3139b0f2dca1) that we're trying to start the engine through docker:

image

I can't quite remember the setup, but shouldn't it be connecting to a kubernetes sidecar? Instead of starting the docker container itself? :thinking:

matipan:
> Hm, noticing [in the trace](https://dagger.cloud/dagger/traces/14ad0b5f31a1c324349718da5eab2140?listen=92ba2d4a11d016ce&listen=ef86199a1dadf06f&listen=5fadad13eb146119&listen=f02e3139b0f2dca1) that we're trying to start the engine through docker:
> image
> 
> I can't quite remember the setup, but shouldn't it be connecting to a kubernetes sidecar? Instead of starting the docker container itself? 🤔

In the old setup yes. But now we are using namespace ubuntu runners so that engine gets executed there.

jedevc (approved):

PR #11106, opened 2025-09-22
chore: ensure commit checkouts have tag history (take 2)

jedevc:
Follow-up to https://github.com/dagger/dagger/pull/11073.

This logic *has* to be duplicated for the remote fetching.

This was likely only previously working because dagger/dagger gets cloned in other places, so we already had all those tags. But the logic needs to be in both places.


jedevc:
Wait. This doesn't matter, we always provide `--tags`.

PR #11111, opened 2025-09-22
EnvFile: better dotenv compatibility, more tests

shykes:
- Evaluate quotes by default: `foo="a b c"`
- Support default values in variables: `foo=${bar:-baz}`
- Cleaner implementation, with a low-level go package
- Lots of tests
- Optional "raw" mode

### DIFF
--- a/core/dotenv/dotenv.go
+++ b/core/dotenv/dotenv.go
@@ -1,0 +1,201 @@
+package dotenv
+
+import (
+	"fmt"
+	"strings"
+
+	"mvdan.cc/sh/v3/expand"
+	"mvdan.cc/sh/v3/syntax"
+)
+
+// Evaluate an array of key=value strings in the dotenv syntax,
+// and return a map of evaluated variables
+func All(environ []string) (map[string]string, error) {
+	vars := make(map[string]string, len(environ))
+	for _, line := range environ {
+		line = strings.TrimSpace(line)
+		if line == "" {
+			continue // skip empty lines
+		}
+		name, value, err := parseEnvLine(line, vars)
+		if err != nil {
+			return vars, err
+		}
+		if name != "" {
+			vars[name] = value
+		}
+	}
+	return vars, nil
+}
+
+// Evaluate an array of key=value strings in the dotenv syntax,
+// in raw mode: values are left unprocessed.
+func AllRaw(environ []string) map[string]string {
+	vars := make(map[string]string, len(environ))
+	for _, kv := range environ {
+		kv = strings.TrimSpace(kv)
+		if kv == "" {
+			continue // skip empty lines
+		}
+		name, value, _ := strings.Cut(kv, "=")
+		vars[name] = value
+	}
+	return vars
+}
+
+// Evaluate an array of key=value strings in the dotenv syntax,
+// and return the value of the specified variable
+func Lookup(environ []string, name string) (string, bool, error) {
+	vars, err := All(environ)
+	if err != nil {
+		return "", false, err
+	}
+	value, ok := vars[name]
+	return value, ok, nil
+}
+
+// Evaluate an array of key=value strings in the dotenv syntax,
+// and return the value of the specified variable in raw mode
+func LookupRaw(environ []string, name string) (string, bool) {
+	vars := AllRaw(environ)
+	value, ok := vars[name]
+	return value, ok
+}
+
+func Exists(environ []string, name string) bool {
+	for _, kv := range environ {
+		kv = strings.TrimSpace(kv)
+		if kv == "" {
+			continue // skip empty lines
+		}
+		k, _, _ := strings.Cut(kv, "=")
+		if k == name {
+			return true
+		}
+	}
+	return false
+}
+
+// parseEnvLine parses one line of a dotenv file into (key, value).
+func parseEnvLine(line string, lookup map[string]string) (string, string, error) {
+	// For simple KEY=VALUE assignments without quotes or $ expansion,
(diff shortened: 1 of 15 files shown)


shykes:
This is a requirement for  #11034

core/dotenv/dotenv.go
```diff
@@ -0,0 +1,208 @@
...
+	fields, err := expand.Fields(cfg, w)
+	if err != nil {
+		return "", err
+		// FIXME:
+		// If expansion fails, fall back to a literal concatenation
+		s, _ := expand.Literal(cfg, w)
+		return s, nil
+	}
+	if len(fields) == 0 {
+		return "", fmt.Errorf("empty expansion")
```
chrisjpalmer, line 202: Can see you got two return statements :)

core/envvars.go
```diff
@@ -78,61 +78,43 @@ func (ef *EnvFile) WithoutVariable(name string) *EnvFile {
...
 	}
-	return vars
+	return ef.variables(ctx)
 }
 
-// Return true if the variable exists
-func (ef *EnvFile) Exists(name string) bool {
-	_, found := ef.lookup(name)
-	return found
+func (ef *EnvFile) variablesRaw(_ context.Context) (vars []EnvVariable, _ error) {
```
grouville, line 88: This will be non-deterministic. If this is expected then all good and resolve that point🙏 

> I'm not sure if this is expected to be consumed by the user and don't really know how you'll use it in https://github.com/dagger/dagger/pull/11034
>
> _What I worry is that this could break caching for any operation that relies on that API, but I agree that it's very nit ahah 👼_ 

For `Container().EnvVariables()`, we seem to ensure determinism with [walkenv](https://github.com/dagger/dagger/blob/main/core/schema/container.go#L1338-L1340): 

```go
core.WalkEnv(cfg.Env, func(k, v, _ string) {
		vars = append(vars, core.EnvVariable{Name: k, Value: v})
})
```

shykes: Ah you're right. Technically it's not a bug, since the map guarantees there are no duplicates. But might as well preserve the order, why not.

core/dotenv/dotenv.go
```diff
@@ -0,0 +1,201 @@
...
+	parser := syntax.NewParser()
+	file, err := parser.Parse(strings.NewReader(line), "")
+	if err != nil {
+		return "", "", fmt.Errorf("parse error: %q: %w", line, err)
+	}
+	if len(file.Stmts) == 0 {
+		return "", "", nil // blank line
+	}
+	stmt := file.Stmts[0]
+	call, ok := stmt.Cmd.(*syntax.CallExpr)
```
mvdan, line 102: you having to do this is a bit unfortunate - just like we have methods on the parser for specific types of nodes, like Arithmetic or Document, we could have one for assignments.

core/dotenv/dotenv.go
```diff
@@ -0,0 +1,201 @@
...
+	if len(assigns) != 1 {
+		return "", "", fmt.Errorf("can't assign multiple variables: %q", line)
+	}
+	assign := assigns[0]
+	if assign.Name == nil {
+		return "", "", fmt.Errorf("missing name in variable assignment: %q", line)
+	}
+	key := assign.Name.Value
+
+	// dotenv twist: handle 'foo=a b c'
```
mvdan, line 116: I'm curious about this - where is their syntax defined for spaces in values?

because if they were trying to be shell-like, such lines would have to be something like `foo="a b c"`. if they're not, then I wonder where they define their syntax, as I'd be curious to read about it.

I also ask because this causes you to do a `strings.Join` on spaces below, which is lossy - `foo=a  b    c` (two spaces and then three) should not be the same as `foo=a b c`, I imagine. or likewise if there are any tabs, or any of the expansions result in newlines.

core/dotenv/dotenv.go
```diff
@@ -0,0 +1,201 @@
...
+			value, ok := lookup[name]
+			if !ok {
+				return ""
+			}
+			return value
+		}),
+		NoUnset: true,
+	}
+	// Perform shell-like expansion: escapes, quotes, parameters
+	fields, err := expand.Fields(cfg, w)
```
mvdan, line 193: FWIW you most likely want `expand.Literal`, because it will not perform word splitting. If you split words, and then re-join with one space, you're basically replacing any amount of contiguous whitespace with one space, which may cause issues for some users.

shykes:
> drive-by review as I saw this on discord :)

Nice, thank you!

core/dotenv/dotenv.go
shykes: Yes, that's the tradeoff with this approach... The shell parser works great for 99% of cases 😅

I actually handle simple assignments (eg. `foo=a b c`) upstream, and bypass the shell parser entirely. This allows me to avoid normalizing non-space whitespaces, so that for example CSV values can be pasted in, and just work. But, it turns out, the `.env` convention is to normalize spaces. So, the "lossy" result is apparently expected.

> I'm curious about this - where is their syntax defined for spaces in values?

LOL 😭 There is no spec... I've been relying on what *seems* to be the most commonly accepted convention, which makes this exercise... shall we say, extra fun.

core/dotenv/dotenv.go
shykes: I seem to remember I had no choice but to do it this way, because of other constraints related to the envfile format... But maybe I got it wrong?

I need to merge this to unblock follow-up features, but eager to improve how I use `sh` if there's a way. The good news is, I have a good set of tests.

PR #11112, opened 2025-09-22
core: treat Azure pkt-len auth probe as a private repo

grouville:
Azure Repos serves the org login page (HTML) when an unauthenticated HTTPS ref lacks the `{org}@` information:
- `curl -i https://dev.azure.com/daggere2e/private/_git/dagger-test-modules.git` returns HTML as a 200
- `curl -i https://daggere2e@dev.azure.com/daggere2e/private/_git/dagger-test-modules.git` returns a 401

When authenticating for a private azure module, we inadventendly stop earlier (before the git credential's session attachable logic) because go-git’s pktline parser raises `ErrInvalidPktLen`.

We now treat that error like `ErrAuthenticationRequired` as this only happens in subtle edge cases where go-git's parser has to parse a login webpage


jedevc:
Lol. Looks like azure gives a 2XX response, and also puts `www-authenticate` in the headers. This is how `git clone` seems to give us the correct behavior. Don't think they should be doing this though https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/WWW-Authenticate.

Anyways. To me that looks slightly like `go-git` is doing the check for authentication failed wrong.

Happy with this quick hack though - another option would to just *try* and do this from `gitutil` instead, and use the authoritative git cli, then we would be able to drop go-git from this bit entirely.

grouville:
> Happy with this quick hack though - another option would to just _try_ and do this from `gitutil` instead, and use the authoritative git cli, then we would be able to drop go-git from this bit entirely.

Given how scoped the fix is, and the fact that it immediately helps users, I'd personally prefer to merge this small stopgap, so that it can be part of `v0.19`. Happy to follow-up later

sheldonhull:
thanks @grouville for looking at this with me, following for any future updates, and we can jump on a liveshare to test it out whenever you have the other improvements in. 👋

sheldonhull:
Update, we did some testing and confirmed with these changes we are good to!
It solved it for the dagger.Directory itself, so not only did it work with module source, but also works when I mapped other flags of `*dagger.Directory` type.

@grouville  _Merci pour le coup de main, t’assures._ 🇫🇷

core/schema/git.go
```diff
@@ -518,14 +519,17 @@ func (s *gitSchema) git(ctx context.Context, parent dagql.ObjectResult[*core.Que
...
 	// check if repo is public
 	repo := git.NewRemote(memory.NewStorage(), &config.RemoteConfig{
 		Name: "origin",
 		URLs: []string{remote.Remote()},
 	})
 	_, err := repo.ListContext(ctx, &git.ListOptions{Auth: nil})
 	if err != nil {
+		if errors.Is(err, pktline.ErrInvalidPktLen) {
+			return false, nil
+		}
```
jedevc, line 534: Worth adding an explanation as to why this is the case inline.

grouville:
Failing on CI's `OpenSSL version mismatch. Built against 3050003f, you have 30500000` error which is unrelated. Merging

Issue #10812, opened 2025-07-29
Remove deprecated fields for v0.19.0


jedevc:
Making a tracking issue so we don't forget this :eyes:

We've got some deprecated fields which can be removed for v0.19.0:
- [x] `withAuthToken`/`withAuthHeader`, deprecated since #10248
  - Done in https://github.com/dagger/dagger/pull/11032 (hard removal due to internal refactoring)
- [x] `keepBytes`, deprecated since #8725
  - Done in https://github.com/dagger/dagger/pull/11115
- [x] `setSecretFile`, deprecated since #9682 
  - Done in https://github.com/dagger/dagger/pull/11115
- [x] `Container.build`, deprecated since #10811.
  - Done in https://github.com/dagger/dagger/pull/11118

jedevc:
All done!

PR #11118, opened 2025-09-23
chore: remove `Container.build` API

jedevc:
See https://github.com/dagger/dagger/issues/10812 - missed this in https://github.com/dagger/dagger/pull/11115.

### DIFF
--- a/.changes/unreleased/Breaking-20250923-171403.yaml
+++ b/.changes/unreleased/Breaking-20250923-171403.yaml
@@ -1,0 +1,9 @@
+kind: Breaking
+body: |-
+    Removed deprecated `Container.build` API
+
+    Use `Directory.dockerBuild` instead.
+time: 2025-09-23T17:14:03.935709067+01:00
+custom:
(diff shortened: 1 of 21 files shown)


jedevc:
Failures are legit but not related to this change :cry: just another one of mine.

jedevc:
Merging this, will fix the other issues in a follow-up.

rafaelpirolla:
I think the python example isn't working anymore: https://docs.dagger.io/cookbook/builds/#build-image-from-dockerfile

jedevc:
@rafaelpirolla the python example uses `Directory.docker_build` which still exists.

```python
@object_type
class MyModule:
    @function
    async def build(
        self,
        src: Annotated[
            dagger.Directory,
            Doc("location of directory containing Dockerfile"),
        ],
    ) -> str:
        """Build and publish image from existing Dockerfile"""
        ref = src.docker_build().publish("ttl.sh/hello-dagger")  # build from Dockerfile
        return await ref

PR #11126, opened 2025-09-25
core: enable nesting by default

shykes:
Nesting is now considered stable and safe. Therefore we can enable it by default.

This allows any containerized tool or service executed by Dagger to open a client session to the same Dagger engine,
without accessing parent resources. This is similar to a unix process being able to make system calls, without
accessing the context of its parent process.

- Deprecate `experimentalPrivilegedNesting` argument. It can safely be omitted
- Keep the plumbing for simplicity. It's simply enabled at the schema level

### DIFF
--- a/core/integration/container_test.go
+++ b/core/integration/container_test.go
@@ -4284,9 +4284,7 @@ func (ContainerSuite) TestNestedExec(ctx context.Context, t *testctx.T) {
 		_, err := c.Container().From(alpineImage).
 			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
 			WithNewFile("/query.graphql", `{ defaultPlatform }`). // arbitrary valid query
-			WithExec([]string{"dagger", "query", "--doc", "/query.graphql"}, dagger.ContainerWithExecOpts{
-				ExperimentalPrivilegedNesting: true,
-			}).
+			WithExec([]string{"dagger", "query", "--doc", "/query.graphql"}).
 			Sync(ctx)
 		require.NoError(t, err)
 	})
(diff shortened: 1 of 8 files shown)


jedevc:
Just a note on this - we used to do some additional cache mixins when running with nesting. I was going to suggest that this was potentially a blocker... but it seems like during theseus we've removed those, and moved them into dagop (which is the right place for em).

Since that's not an issue anymore, this sounds totally reasonable to me, but there may be something subtle here I'm missing (cc @sipsma @vito).

shykes:
Note: this patch is broken (I was targeting an old main...) I will push a cleaner version soon

sipsma (commented):
Generally LGTM, I know there's some places in integ tests where we specifically didn't want nesting for various reasons (`daggerNonNestedExec`), but I think that should be fixable by allowing overwriting `_EXPERIMENTAL_DAGGER_RUNNER_HOST`. I suspect the engine is a bet aggressive about setting that when nested execs are started, so wonder if we'll need something like:
```go
func daggerNonNestedExec(args ...string) dagger.WithContainerFunc {
	return func(c *dagger.Container) *dagger.Container {
		return c.
			// Don't persist stable client id between runs. this matches the behavior
			// of actual nested execs. Stable client IDs on the filesystem don't work
			// when run inside layered containers that can branch off and run in parallel.
			WithEnvVariable("XDG_STATE_HOME", "/tmp").
			WithMountedTemp("/tmp").
			WithExec([]string{"sh", "-c", "export dagger _EXPERIMENTAL_DAGGER_RUNNER_HOST="+thePathToTheCLIWeWantToUse+" "+strings.Join(args, " "))
	}
}
```

Also will want to double check the tests runtime aren't impacted too much via overhead of setting nesting on execs that previously skipped it, but can't see CI until PR is rebased. Not expecting that to be a problem either, just worth a double check.

vito:
Maybe it would make sense to have a `disableNesting` arg?

Issue #10882, opened 2025-08-15
PHP SDK: Implement Implicit Getters


charjr:
### Feature 

SDK's can now implicitly expose a "getter" function on properties: https://docs.dagger.io/api/state/

The PHP SDK will want to implement this for feature parity and reduce boilerplate code for getters.

### Proposal
Each field that you want to expose:
    - MUST have the `DaggerFunction` attribute, consistent with Java and TS SDK getters.
    - MUST be a public field, consistent with Go SDK getters.

The combination of these two requirements is consistent with PHP SDK functions.

```php
// not exposed
public Foo $foo;

// not exposed
#[DaggerFunction] private Bar $bar;

// exposed
#[DaggerFunction] public Baz $baz;

// exposed
#[DaggerFunction] public function getFoo(): Foo
{
    return $this->foo;
}
```
```
FUNCTIONS
  baz
  get-foo

```

- [ ] Add implicit getters
- [ ] Make cookbook snippet visible #11129

PR #11129, opened 2025-09-25
docs: Add PHP snippets to extending section

vikram-dagger:


docs/current_docs/extending/modules/state.mdx
```diff
@@ -44,6 +44,15 @@ Here's an example where one field is exposed as a Dagger Function, while the oth
 ```typescript file=./snippets/state-functions/typescript/index.ts
 ```
 
+</TabItem>
+<TabItem value="php" label="PHP">
+PHP already offers `private`, `protected` and `public` keywords to handle member visibility in a class. However, Dagger will only expose those members of a Dagger module that are explicitly decorated with the `#[DaggerFunction]` decorator. Others will remain private.
```
vikram-dagger: This doesn't seem to behave the same way as in TypeScript. Even though the field is marked as public and decorated with `#[DaggerFunction]`, it is not exposed in `dagger call --help` as shown in the example output.

Ping @charjr for review/advice.

charjr: Blocked by https://github.com/dagger/dagger/issues/10882

Exposing fields has not yet been implemented. The plan was as follows:

Each field that you want to expose:
- MUST have the DaggerFunction attribute, consistent with Java and TS SDK getters.
- MUST be a public field, consistent with Go SDK getters.

That combination is consistent with PHP and Java SDK functions

So this documentation would be something like this:

> Dagger will only expose `public` properties that are explicitly decorated with the `#[DaggerFunction]` decorator. Others will remain private.

marcosnils:
@charjr  just pushed a change here to temporarily hide the state/getters page. Just wondering if you had the time to check the rest of the guides :pray: .

docs/current_docs/extending/modules/snippets/cache-volumes/php/src/MyModule.php
```diff
@@ -0,0 +1,33 @@
...
+#[DaggerObject]
+class MyModule
+{
+    #[DaggerFunction]
+    #[Doc('Build an application using cached dependencies')]
+    public function build(
+      #[Doc('Source code location')]
+      Directory $source,
+    ): Container
+    {
```
charjr:
```suggestion
    ): Container {
```

nitpick: [code style](https://www.php-fig.org/per/coding-style/#45-method-and-function-parameters)

docs/current_docs/extending/modules/snippets/state-functions/php/src/MyModule.php
```diff
@@ -0,0 +1,36 @@
...
+    #[DaggerFunction]
+    public function __construct(
+        #[Doc('The greeting to use')]
+        string $greeting = 'Hello',
+        #[Doc('Who to greet')]
+        string $name = 'World',
+    ) {
+        $this->greeting = $greeting;
+        $this->name = $name;
+    }
```
charjr:
```suggestion
    #[DaggerFunction]
    public function __construct(
        #[DaggerFunction]
        #[Doc('The greeting to use')]
        public string $greeting = 'Hello',
        #[Doc('Who to greet')]
        private string $name = 'World',
    ) {}
```

PHP has what it calls [constructor promotion](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion). By putting visibility on the arguments they're interpreted as properties as well.

TLDR: This suggested change is functionally identical but smaller.

PR #11131, opened 2025-09-26
PHP: (fix) Optional argument before required ones

Neirda24:
This fixes some deprecations showing up about optional arguments being put before required ones.

BEFORE:
```php
    public function search(
        ?array $paths = null,
        ?array $globs = null,
        string $pattern,
        ?bool $literal = false,
        ?bool $multiline = false,
        ?bool $dotall = false,
        ?bool $insensitive = false,
        ?bool $skipIgnored = false,
        ?bool $skipHidden = false,
        ?bool $filesOnly = false,
        ?int $limit = null,
    ): array {
```

AFTER:
```diff
    public function search(
+      string $pattern,
        ?array $paths = null,
        ?array $globs = null,
-      string $pattern,
        ?bool $literal = false,
        ?bool $multiline = false,
        ?bool $dotall = false,
        ?bool $insensitive = false,
        ?bool $skipIgnored = false,
        ?bool $skipHidden = false,
        ?bool $filesOnly = false,
        ?int $limit = null,
    ): array {
```


sdk/php/src/Codegen/Introspection/ObjectVisitor.php
charjr, line 1: LGTM, I don't see anywhere else within `Introspection` that requires arguments sorted.

charjr (commented):
The changes LGTM (at the time of review), I'm only holding off on approval to confirm if it's okay to go in the 0.19.0 release? cc @jedevc

PR #11133, opened 2025-09-26
ci: use dagger-for-github on exec step

matipan:


kpenfound:
@matipan looks great so far, questions I have
1. was the "warm engine" step ever actually needed, or was that only for the local dev engine? Before it was run in between the install and whatever the first call was -- edit: I think this question is answered by the change to modules/gha/workflow.go :)
2. Is the `oidc` / parc engine basically only waiting on the release, or is there more involved? Basically wondering if its worth splitting out the dagger-for-github part from the parc part, but if this is basically it then ship it

matipan:
> was the "warm engine" step ever actually needed, or was that only for the local dev engine? Before it was run in between the install and whatever the first call was -- edit: I think this question is answered by the change to modules/gha/workflow.go :)

Yeah, I think it was mostly for dev engines. It was only doing `dagger core version`. We could bring it back but I honestly don't know the value of it.

> Is the oidc / parc engine basically only waiting on the release, or is there more involved? Basically wondering if its worth splitting out the dagger-for-github part from the parc part, but if this is basically it then ship it

Only waiting for the release. The changes I made in this PR work as is if you use a CLI built from latest dagger. I tested that in a separate branch by creating a new workflow and doing the full end-to-end process. That said, if tomorrow after the release we realize its more involved then we should definitely split the PR in two 👍🏼

matipan:
Closing this PR in favor of #11151. Since we need to make changes to the `permissions` object of existing workflows we can't test that through a PR coming from a fork, it has to be a branch of the repo.

PR #11139, opened 2025-09-29
v0.19.0 breaking changes

jedevc:
This PR includes all v0.19.0 changes that were merged into the v0.19.0 branch instead of main because they were breaking.

All of these PRs have already been approved, but since v0.19.0 is tomorrow, we should bring them into main.


vito (approved):

jedevc:
Manually setting DCO to pass - those failing are previously merged commits, so do not need a sign off.

Issue #6076, opened 2023-11-06
Go SDK: support go mod vendor in modules


sipsma:
Right now vendor directories in a go module don't really work as expected (or at all depending on the situation).

They should work seamlessly and enable the removal off all deps on the internet in terms of go modules and use of private packages the engine may not be able to access.

PR #11143, opened 2025-09-29
go: allow vendor/* files through

tiborvass:
This allows to build Go apps that use vendoring, using Dagger's Go module. Useful when making local modifications to vendored libraries.


grouville (approved):
✅

tiborvass:
Related to #6076

PR #11144, opened 2025-09-30
feat: capture contextual information for provisioning PARC engines

matipan:
We send the module, function name and command being executed to the API and let it make the decision for how to provision an engine. The function name parsing function added only supports commands when flags are set using the `--cloud=true` format. If flags do not use `=` to set values we do not set the function name. This was discussed on Monday's meeting and agreed as a good enough stop gap given current constraints. See the `spanName` function for a deeper explanation on why this is a stop gap.

### DIFF
--- a/cmd/dagger/function_name_test.go
+++ b/cmd/dagger/function_name_test.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFunctionName(t *testing.T) {
+	type example struct {
+		name string
+		args []string
+		want string
+	}
+	for _, test := range []example{
+		{
+			name: "invalid flag without equals",
+			args: []string{"--cloud", "test", "specific"},
+			want: "",
+		},
+		{
+			name: "valid flag with equals before function",
+			args: []string{"--cloud=true", "test", "specific"},
+			want: "test",
+		},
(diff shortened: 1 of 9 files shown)


cmd/dagger/function_name_test.go
```diff
@@ -0,0 +1,68 @@
...
+			want: "",
+		},
+		{
+			name: "single dash flags before function",
+			args: []string{"-m", "dev", "test", "specific"},
+			want: "",
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			if test.name == "" {
```
marcosnils: is this `if` ok here? should it be outside the above `t.Run`?

engine/client/drivers/cloud.go
```diff
@@ -81,20 +81,28 @@ func (d *daggerCloudDriver) Provision(ctx context.Context, _ *url.URL, opts *Dri
...
-func (d *daggerCloudDriver) ImageLoader(ctx context.Context) imageload.Backend {
-	return nil
-}
+	var (
+		module, function, tag string
+		execCmd               []string
+	)
+	if opts != nil {
+		module = opts.Module
+		tag = opts.GPUSupport
```
marcosnils: woot! fix this

marcosnils (approved):

PR #11148, opened 2025-09-30
llm: bump Anthropic default

vito:
*sigh*

In #11145 I bumped the version that we test, but not the actual default version that we ship. :upside_down_face:


jedevc (approved):
🤦 

good timing 😁

jedevc:
`go mod tidy` needed running, have updated that.

PR #11151, opened 2025-09-30
ci: use dagger-for-github instead of exec and migrate SDKs to PARC

matipan:
There's three main changes in this PR:
1. Migrating to `dagger-for-github` on the `exec` step instead of the scripts. This means we can remove the `install` as well as the `warm engine` steps. For cases with `dagger-dev` enabled we still use `exec` and a dev engine but its just one.
2. Setup PARC engines for all SDKs job using the temporary scale out workaround with OIDC
3. Switch to using `OIDC` for all telemetry instead of exposing our public token

In the case of 1. we are loosing GitHub job summaries simply because we haven't released the latest merged changes to dagger-for-github. After we do that we can come back here, bump up the version and set the right flags. I don't want to wait for that though, I would rather push this through.


.github/main.go
```diff
@@ -13,7 +13,7 @@ const (
 	upstreamRepository = "dagger/dagger"
 	ubuntuVersion      = "24.04"
 	defaultRunner      = "ubuntu-" + ubuntuVersion
-	publicToken        = "dag_dagger_sBIv6DsjNerWvTqt2bSFeigBUqWxp9bhh3ONSSgeFnw"
```
shykes, line 16: 🥳

modules/gha/steps.go
```diff
@@ -46,41 +46,62 @@ func (j *Job) installDaggerSteps() []api.JobStep {
...
 }
 
-func (j *Job) uploadEngineLogsStep() []api.JobStep {
-	if j.DaggerDev == "" {
-		return nil
+func (j *Job) callDaggerStep() api.JobStep {
+	// for dagger dev engines continue to use the exec step so that we don't
+	// override the dagger CLI
+	if j.DaggerDev != "" {
+		return j.callDaggerWithExecStep()
```
jedevc, line 53: Think I mentioned this before, but would be nice to drop this bit entirely - all we need to do is to do start the dev engine inside of dagger, instead of launching it in docker.

That should simplify a few of these code paths.

modules/gha/steps.go
```diff
@@ -46,41 +46,62 @@ func (j *Job) installDaggerSteps() []api.JobStep {
...
-			If:    "always()",
-			Shell: "bash",
-			Run:   `sudo dmesg || true`,
-		},
-		*/
+	env := map[string]string{}
+	// Debug mode
+	if j.Debug {
+		env["DEBUG"] = "1"
+	}
```
jedevc: This is used by the `exec.sh` script. But `dagger-for-github` won't use it right?

Ditto for `DOTENV`.

modules/gha/steps.go
matipan: Definitely! We discussed a bit via discord. Sharing some of that context here.

I've been trying to run the check inside a dev engine by using the `dev` function that returns a dev engine container ready with the dagger cli connected to it. Running it all locally works with a small hack: you need to send over the `.git` folder, otherwise module loading of `/app` inside the container will not work. The following command works:
```sh
 dagger call --cloud dev --target .git with-workdir --path /app with-exec --args "cp","-r","/mnt","/app/.git" with-exec --args "sh","-c","dagger call check --targets=sdk"
```

Alternatively, we could do a checkout, mount the entire directory and run the command within `/mnt`:
```sh
 dagger call --cloud dev --target . with-workdir --path /mnt with-exec --args "sh","-c","dagger call check --targets=sdk"
```

Moving this logic over to a module function means that we will have a special case where the function ignores the directory in `New` and instead specifies its own. Initially I wanted this to be a part of the checks interface (i.e `dagger call check --targets=dev-engine`). But given this constraint of needing a special directory we can't make it happen without changing the arguments of the `Check` function, which I would rather avoid. 

Happy to open a follow up PR with this!

PR #11152, opened 2025-09-30
fix: avoid recomputing content hashes for git tree hashes

jedevc:
When creating a clone of a private repo we do a fun little dance to mixin the content hash of the clone so that we avoid mixing up two repos with the same content, but different URLs. This scenario appears frequently with out module tests (with mirrored versions of the same repo in multiple places).

When we introduced `maintainContentHashing` in https://github.com/dagger/dagger/pull/10761, the intention was to ensure that a `Directory` that was previously content-hashed (e.g. on load from a client using `Host.directory`) *stayed* content content-hashed when we select one of it's subdirectories. This is a relatively cheap operation, because we should have already created a cache context.

However - these two ideas conflict. To see this:

    $ dagger shell

    ▶ git "git@github.com:dagger/dagger.io" | head | tree
    Directory@xxh3:85f08c1a0f41f188

    ▶ git "git@github.com:dagger/dagger.io.git" | head | tree
    Directory@xxh3:6cf30aa8e3e2274e

    ▶ git "git@github.com:dagger/dagger.io" | head | tree | directory api
    Directory@sha256:f3e0e7d9fec74300ec961a442ad139b0d3b1b7262b3c217a47af4d1a8b472c97

    ▶ git "git@github.com:dagger/dagger.io.git" | head | tree | directory api
    Directory@sha256:f3e0e7d9fec74300ec961a442ad139b0d3b1b7262b3c217a47af4d1a8b472c97

Note how the top-level directory uses a custom dagql hash, while the selection uses a content hash; now these two *different* URLs which used to result in two different IDs now result in a collision. This causes issues downstream (not sure why, struggling to reproduce - I'm not sure why this would end up being computed again, since this collision should only be reachable after we've actually cloned the repo, however, this could happen if the buildkit cache is tidied at the right point *during* a test run).

The fix here is to narrow our definition of a content hash - the buildkit `contenthash` package always creates SHA256 hashes, so we should *only* maintain content hashing if we know that a SHA256 hash was set (if it's creating using dagql.From, it'll be an XXH3 hash, and we've deliberately overwritten it).

This feels hacky - but I don't have a super clever suggestion here, it feels like the correct solution is to work out *how* to allow these content hash collisions (so it's not an issue in the first place), but as of the current state of dagql<->buildkit cache, I'm not sure how feasible that is.

### DIFF
--- a/core/schema/directory.go
+++ b/core/schema/directory.go
@@ -14,6 +14,7 @@ import (
 	"github.com/dagger/dagger/core"
 	"github.com/dagger/dagger/dagql"
 	"github.com/moby/patternmatcher/ignorefile"
+	"github.com/opencontainers/go-digest"
 )
 
 type directorySchema struct{}
@@ -1106,7 +1107,12 @@ func maintainContentHashing[A any](
 		if err != nil {
 			return res, err
 		}
-		if parent.ID().HasCustomDigest() { // in practice right now, a custom digest is always a content hash
+
+		// in practice right now, a custom digest is always a content hash
+		// *unless* it's been manually rewritten using dagql.HashFrom (e.g. in
+		// the case of GitRef.tree - that case is manually rewritten to avoid
+		// accidental collisions later)
+		if parent.ID().HasCustomDigest() && parent.ID().Digest().Algorithm() == digest.SHA256 {
 			query, err := core.CurrentQuery(ctx)
 			if err != nil {
 				return res, err


jedevc:
See https://dagger.cloud/dagger/traces/1d638a87553a4e5daf3666ffd14cced3 for an example of the failure this hopes to fix. This *appears* unrelated to the `ls-remote` and similar changes in #11032, since it doesn't appear in https://github.com/dagger/dagger/commits/36d05a12 (where the above failure came from).

Instead, I *suspect* this started happening more after merging https://github.com/dagger/dagger/pull/11070, which allows more dagql cache sharing. However, still struggling to create a reproducer for this scenario, so can't confirm.

TomChv (approved):

Issue #11160, opened 2025-10-01
CLI exports Changeset to the wrong directory


shykes:
## Problem

When `dagger call` calls a function that returns a `Changeset`, the CLI prompts the user to apply the changes to the local filesystem.

Unfortunately, the changes are applied *relative to the CLI's working directory*, which is brittle and often surprising.
This means that the same command will fail or succeed (or worse, silently dump data in the wrong place) depending on the workdir.
This is surprising because the rest of `dagger call` doesn't behave differently based on workdir, but rather based on selected module.

Examples:

Assuming a `Generate` function in `./docs` that returns changes *relative to the root* (this is our current pattern in dagger/dagger:

| Command | Workdir | Changeset root | `docs` written to.... | Success? |
| -- | -- | -- | -- | -- |
| `dagger -m ./docs call generate` | `.` | `.` | `./docs` | ✅ |
| `cd ./docs; dagger call generate` | `./docs` | `.` | `./docs/docs` | ☠️|
| `cd ./docs/static; dagger call generate` | `./docs/static` | `.` | `./docs/static/docs` | ☠️|

If I change my `Generate` function to return changes *relative to the `docs` module*, then I move the problem:

| Command | Workdir | Changeset root | `docs` written to.... | Success? |
| -- | -- | -- | -- | -- |
| `dagger -m ./docs call generate` | `.` | `./docs` | `./` | ☠️ |
| `cd ./docs; dagger call generate` | `./docs` | `./docs` | `./docs` | ✅|
| `cd ./docs/static; dagger call generate` | `./docs/static` | `.` | `./docs/static` | ☠️|

## Solution

Trace back the origin of the Changeset to a Host directory. When that information is available, apply the changeset there.

Also: when the information is *not* available, possibly the user does not expect to be prompted to apply the change, at least not without an explicit flag perhaps.

shykes:
FYI @vito @jedevc 

This is one aspect of #8235 that isn't yet tackled by `Changeset`. I think it's the last remaining issue though! Everything else is *much* better now :)

Context: I'm porting our CI to use `Changeset`

jedevc:
Hard agree with this.

We can sort of track the origin of what a `Changeset` applies to by tracking the ancestor of `Before`, and work out way back to a `Host.directory` (similar to the approach in #11158).

[...]

PR #11161, opened 2025-10-01
Simplify our CI

shykes:
This cleans up and simplifies our CI module in several ways:

- Use `Changeset` in all generation functions
- Expand top-level `generate()` to also generate Github Actions config, and Changelog
- Decouple format checks (`lint()`) from generated file checks (`checkGenerated()`). Make that a new top-level CI job.
- Remove the `dirdiff` module which was used to check generated files. We use the native `Changeset` type now. Maybe this will speed up module loading @matipan @kpenfound @TomChv ?
- New helpers `Parallel` and `Job` to simplify the "errgroup + custom spans" pattern
- Remove the heavyweight `sdk all` abstraction
- Simplify structure of generated GHA config
- Remove the heavyweight "checks router" abstraction
- Remove custom `SourceDeveloped` logic, making it generic in `modules/go` instead.


.dagger/checks.go
```diff
@@ -2,166 +2,124 @@ package main
...
-		{
-			Name: name + "/test-publish",
-			Check: func(ctx context.Context) error {
-				branches, err := dev.Git.Branches(ctx, dagger.VersionGitBranchesOpts{
-					Commit: "HEAD",
-				})
-				if err != nil {
-					return err
+// Lint the Go codebase
+func (dev *DaggerDev) CheckLintGo(
```
vito: In a world where we auto-call entrypoints with void/scalar return values and a `Check` prefix, won't this be redundant with `CheckLint` since that also calls it?

Maybe it should just be called `LintGo` to avoid that?

```suggestion
func (dev *DaggerDev) LintGo(
```

.dagger/main.go
```diff
@@ -243,48 +250,122 @@ func (dev *DaggerDev) Bench() *Bench {
...
+				})
+			}
+			return jobs.Run(ctx)
+		}).
+		Run(ctx)
+	if err != nil {
+		return nil, err
+	}
+	var result *dagger.Changeset
+	// FIXME: this is a workaround to TUI being too noisy
```
vito, line 267: For now, you could have job-created spans set `telemetry.Reveal()`. and we can have the Statuses API just do that by default (#9327)

.dagger/scan.go
```diff
@@ -4,10 +4,11 @@ import (
 	"context"
 
 	"github.com/dagger/dagger/.dagger/internal/dagger"
-	"golang.org/x/sync/errgroup"
+	"github.com/dagger/dagger/util/parallel"
 )
 
-func (dev *DaggerDev) Scan(ctx context.Context) error {
+// Scan source code and artifacts for security vulnerabilities
+func (dev *DaggerDev) CheckScan(ctx context.Context) error {
```
vito, line 11: nit: I'm not a huge fan of the `Check` prefix everywhere, since the names end up a bit clunky ("verb verb" instead of the usual "verb noun"). I get the idea - we need some way to signal which things should be auto-called. But the names just sound a bit unnatural, like it's unclear why they have a different pattern from their siblings.

WDYT of moving them to a separate type returned by a single function, `checks` (or `check` - feels like both work)? Then it's `dagger call check scan` which at least feels like a hierarchy. Same number of characters to type (`check-scan` vs. `check scan`) and sounds less weird imo.

Later, we could grow that into Concourse style pipelines, where checks can return values that automatically propagate into other checks that take them as arguments, evaluating the pipeline in DAG order. The separate outermost type acts as a scope where everything descending from it is OK to auto-run. In the Concourse PoC I went with `pipeline` as the entrypoint name, which could work here too (`dagger call pipeline scan`), but it might feel weird right now since there aren't actual pipeline semantics.

vito: Ah, I see there are `CheckFoo`s on other objects, like `ElixirSdk.CheckLint` - is the idea to call them throughout the API so long as they're reachable without any required args?

If so I do see the appeal over having to make a separate type that mirrors a similar hierarchy. Hmm... (Still don't love the naming, but hmm...)

vito: `// +check`?

This was raised last year, it's yet another pragma but it feels better to me; it's more explicit, and it doesn't affect how users name their functions.

(We don't necessarily need to solve this for this PR btw, still a nitpick, just getting some bikeshedding in while the paint is wet.)

sdk/python/src/dagger/client/gen.py
```diff
@@ -11147,7 +11207,8 @@ async def text(self) -> str:
 @typecheck
 class Secret(Type):
     """A reference to a secret value, which can be handled more safely
-    than the value itself."""
+    than the value itself.
```
vito: Are these trailing linebreaks intentional?

docs/.dagger/main.go
```diff
@@ -289,8 +203,8 @@ func (d Docs) Bump(version string) (*dagger.Directory, error) {
 export const daggerVersion = "%s";
 `, version)
 
-	dir := dag.Directory().WithNewFile("docs/current_docs/partials/version.js", versionFile)
-	return dir, nil
+	layer := dag.Directory().WithNewFile("docs/current_docs/partials/version.js", versionFile)
```
vito: Any reason not to use `d.Source` here? Would help with #11160 once that's fixed

.dagger/sdk_typescript.go
```diff
@@ -215,14 +195,32 @@ func (t TypescriptSDK) Bump(ctx context.Context, version string) (*dagger.Direct
 
 	// NOTE: if you change this path, be sure to update .github/workflows/publish.yml so that
 	// provision tests run whenever this file changes.
-	return dag.Directory().WithNewFile("sdk/typescript/src/provisioning/default.ts", engineReference), nil
+	layer := dag.Directory().WithNewFile("sdk/typescript/src/provisioning/default.ts", engineReference)
```
vito: could use `t.Source` here (I assume)

.dagger/sdk_rust.go
```diff
@@ -202,50 +209,58 @@ func (r RustSDK) Bump(ctx context.Context, version string) (*dagger.Directory, e
...
+	base := r.DevContainer().
 		WithExec([]string{
 			"cargo", "install", "cargo-edit@" + cargoEditVersion, "--locked",
 		}).
 		WithExec([]string{
 			"cargo", "set-version", "-p", crate, version,
 		})
 
-	return dag.Directory().WithNewFile(rustVersionFilePath, versionBumpedContents).
+	layer := dag.Directory().WithNewFile(rustVersionFilePath, versionBumpedContents).
```
vito: (here too)

.dagger/sdk_rust.go
```diff
@@ -7,89 +7,96 @@ import (
...
-		File(strings.TrimPrefix(rustGeneratedAPIPath, "sdk/rust/"))
-
-	return dag.Directory().
-		WithDirectory("sdk/rust", r.Dagger.Source.Directory("sdk/rust")).
-		WithFile(rustGeneratedAPIPath, generated), nil
+		Directory(".").
+		Filter(relGeneratorNoiseFilter) // exclude garbage from generation
+	absLayer := dag.Directory().
+		WithDirectory("sdk/rust", relLayer)
+	return absLayer.Changes(r.Source()), nil
```
vito: Should maybe be this?


```suggestion
	absLayer := r.Source().
	  WithoutDirectory("sdk/rust").
		WithDirectory("sdk/rust", relLayer)
	return absLayer.Changes(r.Source()), nil
```

.dagger/sdk_python.go
```diff
@@ -163,12 +175,13 @@ func (t PythonSDK) Publish(
...
-func (t PythonSDK) Bump(ctx context.Context, version string) (*dagger.Directory, error) {
+func (t PythonSDK) Bump(ctx context.Context, version string) (*dagger.Changeset, error) {
 	// trim leading v from version
 	version = strings.TrimPrefix(version, "v")
 	engineReference := fmt.Sprintf("# Code generated by dagger. DO NOT EDIT.\n\nCLI_VERSION = %q\n", version)
 
 	// NOTE: if you change this path, be sure to update .github/workflows/publish.yml so that
 	// provision tests run whenever this file changes.
-	return dag.Directory().WithNewFile("sdk/python/src/dagger/_engine/_version.py", engineReference), nil
+	layer := dag.Directory().WithNewFile("sdk/python/src/dagger/_engine/_version.py", engineReference)
```
vito: use t.Source?

.dagger/sdk_php.go
```diff
@@ -185,13 +171,13 @@ func (t PHPSDK) Publish(
...
+func (t PHPSDK) Bump(ctx context.Context, version string) (*dagger.Changeset, error) {
 	version = strings.TrimPrefix(version, "v")
 	content := fmt.Sprintf(
 		"<?php\n\n/* Code generated by dagger. DO NOT EDIT. */\nreturn '%s';\n",
 		version,
 	)
 
-	dir := dag.Directory().WithNewFile(filepath.Join(phpSDKPath, phpSDKVersionFile), content)
-	return dir, nil
+	layer := dag.Directory().WithNewFile(filepath.Join("sdk/php", phpSDKVersionFile), content)
```
vito, line 184: could use `t.Source` here (I assume)

.dagger/sdk_php.go
```diff
@@ -6,138 +6,124 @@ import (
...
-		Directory(".")
-	return dag.Directory().WithDirectory(phpSDKPath, generated)
+		Directory(".").
+		Filter(dagger.DirectoryFilterOpts{
+			Exclude: []string{
+				"vendor",
+			},
+		})
+	// Make the change relative to the repo root
+	absLayer := dag.Directory().WithDirectory("sdk/php", relLayer)
```
vito: could use same trick as rust SDK (withoutDirectory => withDirectory)

.dagger/sdk_go.go
```diff
@@ -146,6 +125,6 @@ package engineconn
 const CLIVersion = %q
 `, version)
 
-	dir := dag.Directory().WithNewFile("sdk/go/engineconn/version.gen.go", versionFile)
-	return dir, nil
+	layer := dag.Directory().WithNewFile("sdk/go/engineconn/version.gen.go", versionFile)
```
vito: could use `t.Source` here (I assume)

.dagger/checks.go
shykes: > In a world where we auto-call entrypoints with void/scalar return values and a `Check` prefix, won't this be redundant with `CheckLint` since that also calls it?

Yes, but it's `CheckLint` that we'll remove. It's an "aggregator" that will no longer be required once the core can aggregate natively, and better (for example with built-in filtering etc).

.dagger/scan.go
shykes: As discussed live - `+check` is definitely interesting. For this PR I only rely on features available in main, but we should definitely bikeshed this in #11211

sdk/python/src/dagger/client/gen.py
shykes: It's a regression in the python codegen... I can't get it to produce the exact same output 😭

docs/.dagger/main.go
shykes: I was probably just transposing what was there before (or blindly changing things to debug an issue).

I'll fix it.

sdk/python/src/dagger/client/gen.py
shykes: Is the new output acceptable? If it's important to keep the exact same output as before, I'm going to need some help getting it... I have been down the rabbit hole and can't figure it out.

.dagger/main.go
shykes: Will tackle in a follow-up

sdk/python/src/dagger/client/gen.py
shykes: cc @grouville @tiborvass

grouville: I think it's fine. If you want the same output, you need to change the path of the ignore rule [here](https://github.com/dagger/dagger/blob/main/sdk/python/ruff.toml#L64) to match the new workdir (codegen) that you set

shykes:
All green!

For the record, how I fixed my nightmarish combination of failures today:

- Reverted use of `Changeset` in new "generate dagger runtimes" logic in `modules/go` (30+ changesets to merge, caused random "file not found" errors) fyi @vito 
- Limited parallelism in 1) dagger runtime generations prior to go linting, and 2) go linting. Thank you @sipsma for the tip. This solved the OOM/timeout issue in CI
- Lint error explosion. This was in fact caused by *fixing* our linting logic, which failed to properly escalate some checks as errors. This is now fixed, all newly discovered lint issues have been addressed. Thank you @marcosnils @jedevc @sipsma @vito @kpenfound @matipan for the bug hunt!

.changes/.dagger/main.go
jedevc, line 1: Nice :tada: I like this.

cmd/engine/.dagger/main.go
```diff
@@ -277,86 +279,78 @@ func (e *DaggerEngine) Publish(
...
-		}(); err != nil {
-			return err
-		}
+				return nil
+			})
 	}
-
-	return nil
+	// FIXME: is this intentionally serialized?
+	return jobs.RunSerial(ctx)
```
jedevc: Nope :tada: feel free to parallelize.

.dagger/checks.go
```diff
@@ -2,151 +2,105 @@ package main
...
-		{
-			Name: name + "/test-publish",
-			Check: func(ctx context.Context) error {
-				return sdk.TestPublish(ctx, "HEAD")
-			},
-		},
-	}
+func (dev *DaggerDev) CheckLint(ctx context.Context) error {
+	return parallel.New().
+		WithJob("lint go packages", dev.CheckLintGo).
```
jedevc: nit: should `lint` be prefixed to all of this, given the function name already has `CheckLint`.

`CheckReleaseDryRun` doesn't have the same prefixing, just wandering if it's intentionally inconsistent here.

.dagger/checks.go
```diff
@@ -2,151 +2,105 @@ package main
...
 
-func (r *checkRouter) Get(targets ...string) []Check {
-	var checks []Check
-	for _, target := range targets {
-		checks = append(checks, r.get(target)...)
-	}
-	return checks
+// Lint the helm chart
+func (dev *DaggerDev) CheckLintHelm(ctx context.Context) error {
+	// FIXME: temporary wrapper
```
jedevc: What's the temporary wrapper here?

IMO, if we're committing `TODO`s/`FIXME`s into `main`, there should be enough *inline* context for what the desired end state is, so that anyone can come and pick it up. Or if there's no work left to do, then remove it.

.dagger/checks.go
```diff
@@ -2,151 +2,105 @@ package main
...
-			},
-		},
-	}
+func (dev *DaggerDev) CheckLint(ctx context.Context) error {
+	return parallel.New().
+		WithJob("lint go packages", dev.CheckLintGo).
+		WithJob("lint docs", dev.CheckLintDocs).
+		WithJob("lint helm chart", dev.CheckLintHelm).
+		WithJob("lint install scripts", dev.CheckLintScripts).
+		WithJob("lint SDKs", dev.CheckLintSDKs).
```
jedevc: I guess this more of a question for #11211 - but are both `CheckLint` and `CheckLintSDKs` registered as checks?

If so, we're going to run the contents of that check twice? Caching/dagql *sort* of should avoid us running work twice? But ideally we'd avoid it.

(note: there's a few other places that also do this)

.dagger/go.mod
```diff
@@ -13,10 +13,11 @@ require (
 	github.com/99designs/gqlgen v0.17.80
 	github.com/BurntSushi/toml v0.3.1
 	github.com/Khan/genqlient v0.8.1
-	github.com/containerd/platforms v0.2.1
+	github.com/containerd/platforms v1.0.0-rc.1
+	github.com/dagger/dagger v0.0.0-00010101000000-000000000000
```
jedevc: Just for neatness, can we put this and the corresponding replace at the top of the file? With the existing `github.com/dagger/dagger` imports.

.dagger/go.mod
```diff
@@ -63,3 +63,5 @@ replace go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp => go.opente
 replace go.opentelemetry.io/otel/log => go.opentelemetry.io/otel/log v0.14.0
 
 replace go.opentelemetry.io/otel/sdk/log => go.opentelemetry.io/otel/sdk/log v0.14.0
+
+replace github.com/dagger/dagger => ..
```
jedevc: I'm sure there was a reason for not doing this previously - there's a reason that `distconsts` has a `go.mod` to make it a package. But it slips my mind, I suppose it's fine if it works now.

Maybe there was something about all the dependencies from the parent go.mod getting pulled in? But that doesn't seem to be happening tbf, so not sure. We should [keep an eye on it though](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence).

.dagger/main.go
```diff
@@ -220,48 +163,125 @@ func (dev *DaggerDev) Bench() *Bench {
...
+	// +optional
+	check bool,
+) (*dagger.Changeset, error) {
+	var (
+		genDocs, genEngine, genChangelog, genGHA *dagger.Changeset
+		genSDKs                                  []*dagger.Changeset
+	)
+	maybeCheck := func(ctx context.Context, cs *dagger.Changeset) (*dagger.Changeset, error) {
+		//		cs = dev.Source.WithChanges(cs). // apply to source
+		//							Changes(dev.Source) // check for *actual* changes
```
jedevc: nit: commented code

.dagger/main.go
```diff
@@ -220,48 +163,125 @@ func (dev *DaggerDev) Bench() *Bench {
...
+
+		if !check {
+			// Always use the context, for correct span attribution
+			return cs.Sync(ctx)
+		}
+		diffSize, err := cs.AsPatch().Size(ctx)
 		if err != nil {
-			return err
+			return cs, err
```
jedevc:
```suggestion
			return nil, err
```

.dagger/sdk_dotnet.go
```diff
@@ -11,36 +11,49 @@ type DotnetSDK struct {
...
-	introspection := t.Dagger.introspection(installer)
 	src := t.Dagger.Source.Directory("sdk/dotnet")
-	return dag.DotnetSDKDev(dagger.DotnetSDKDevOpts{Source: src}).Test(ctx, introspection)
+	return dag.
+		DotnetSDKDev(dagger.DotnetSDKDevOpts{Source: src}).
+		Test(ctx, t.Dagger.introspectionJSON())
 }
 
-func (t DotnetSDK) TestPublish(ctx context.Context, tag string) error {
+func (t DotnetSDK) CheckReleaseDryRun(ctx context.Context) error {
```
jedevc: Can we just remove this function now (and similar non implemented checks), since we have the dynamic check? I think as implemented, it'll show up as a job, but it's actually not doing anything.

.dagger/sdk_java.go
```diff
@@ -30,21 +34,23 @@ func (t JavaSDK) Lint(ctx context.Context) error {
...
 }
 
 // Regenerate the Java SDK API
-func (t JavaSDK) Generate(ctx context.Context) (*dagger.Directory, error) {
-	return dag.Directory(), nil
+func (t JavaSDK) Generate(ctx context.Context) (*dagger.Changeset, error) {
+	// FIXME: why is this a no-op? 🤔
+	return dag.Directory().
+		Changes(dag.Directory()).
+		Sync(ctx)
```
jedevc, line 48: @eunomie?

.dagger/sdk_php.go
```diff
@@ -6,138 +6,123 @@ import (
...
+	src := t.Source()
+	relLayer := dag.PhpSDKDev(dagger.PhpSDKDevOpts{Source: src}).
 		Base().
-		With(installer).
-		WithoutDirectory(phpSDKGeneratedDir).
-		WithDirectory(phpSDKGeneratedDir, dag.Directory()).
+		With(t.Dagger.devEngineSidecar()).
+		WithoutDirectory("generated").
+		WithDirectory("generated", dag.Directory()).
+		// FIXME: why not inject the right dagger binary, instead of leaking this env var?
```
jedevc, line 74: Mmmmm. There's some magic in our `core/integration` tests that use this, but for our use here, we don't need to be setting it.

The original source code for `devEngineSidecar` was pretty much copy-pasted from `core/integration` so that's why it would have it in the first place.

.dagger/sdk_python.go
```diff
@@ -5,126 +5,136 @@ import (
...
-	return dag.Directory().WithDirectory(
-		pythonSubdir,
-		dag.PythonSDKDev().Generate(introspection),
+func (t PythonSDK) Generate(_ context.Context) (*dagger.Changeset, error) {
+	devContainer := dag.PythonSDKDev().Container()
+
+	// We don't control the input source, it's defined in wrapped native module
+	srcMountPath := "/src"
+	src := devContainer.Directory(srcMountPath)
+	// FIXME: workaround for Directory.changes() bug
```
jedevc, line 96: Curious about the rationale behind this change - I know you mentioned wanting to improve performance by merging more back into the monolith, but... we're still importing it.

IMO, we should either leave the `Generate` functionality in the python SDK until we lift *all* of it up here.

.github/workflows/check-go-tidy.gen.yml
jedevc, line 1: This is a weird file name.

docs/.dagger/main.go
```diff
@@ -88,114 +77,88 @@ func (d Docs) Server() *dagger.Container {
...
+	// 2. Generate the API reference docs
+	//withApiReference := dag.Container().
+	//	From("node:22").
+	//	WithMountedDirectory("/src", withGqlSchema).
+	//	WithMountedDirectory("/mnt/spectaql", spectaql()).
+	//	WithWorkdir("/src/docs").
+	//	WithExec([]string{"yarn", "add", "file:/mnt/spectaql"}).
+	//	// -t specifies the target directory where spectaql will write the generated output
+	//	WithExec([]string{"yarn", "run", "spectaql", "./docs-graphql/config.yml", "-t", "./static/api/reference/"}).
+	//	Directory("/src")
```
jedevc: Are we intentionally not doing this anymore?

modules/go/main.go
```diff
@@ -416,32 +502,146 @@ func goCommand(
...
+	}
+	jobs := parallel.New().WithLimit(3)
+	for _, mod := range modules {
+		jobs = jobs.WithJob(mod, func(ctx context.Context) error {
+			_, err := dag.Container().
+				From(lintImage).
+				WithFile(configPath, dag.CurrentModule().Source().File("lint-config.yml")).
+				WithMountedCache("/go/pkg/mod", p.ModuleCache).
+				WithMountedCache("/root/.cache/go-build", p.BuildCache).
+				// FIXME: concurrency ok?
```
jedevc: Seems to be! :eyes:

modules/go/main.go
```diff
@@ -416,32 +502,146 @@ func goCommand(
...
+	lintImageRepo := "docker.io/golangci/golangci-lint"
+	lintImageTag := "v2.5.0-alpine"
+	lintImageDigest := "sha256:ac072ef3a8a6aa52c04630c68a7514e06be6f634d09d5975be60f2d53b484106"
+	lintImage := lintImageRepo + ":" + lintImageTag + "@" + lintImageDigest
+	configPath := "/etc/golangci.yml"
+	modules, err := p.Modules(ctx)
+	if err != nil {
+		return err
+	}
+	jobs := parallel.New().WithLimit(3)
```
jedevc, line 627: Worth adding an inline note about this IMO, since it took a while of debugging.

sdk/python/src/dagger/client/gen.py
jedevc, line 1: Hm, where do these changes come from? This is the client library changing.

.dagger/checks.go
shykes: `CheckLint` is a temporary shim until native checks are merged.

I am going to push a separate branch that expects native checks - which we can use for native CI testing

.dagger/checks.go
shykes: It's temporary in the sense that once we have native support for checks + multiple blueprints, it won't be needed anymore - it will be picked up automatically in the upstream module. This wrapper replicates that behavior as closely as possible, so that the diff is as small as possible to add support for the new native features

sdk/python/src/dagger/client/gen.py
shykes: See https://github.com/dagger/dagger/pull/11161#discussion_r2431194600

docs/.dagger/main.go
shykes: Not intentional... 😬 Fixing

sdk/python/src/dagger/client/gen.py
jedevc: @grouville I'm not sure I understand. Why would the linting change that?

I think it's worth spending a moment to try and *avoid* a regression if at all possible, it's weird that the formatting changes.

shykes: It's an issue with how the ruff formatter is configured. The configuration was tightly coupled to the path of the file, workdir, or both. I had to refactor the dagger functions orchestrating this, and in the process caused a different config to be applied. Tracking down what exactly changed is painful. That's the tldr.

grouville: Added a commit with the fix

shykes: Thank you!

.dagger/checks.go
```diff
@@ -2,151 +2,102 @@ package main
...
 }
 
-type Check struct {
-	Name  string
-	Check func(context.Context) error
+func (dev *DaggerDev) CheckReleaseDryRun(ctx context.Context) error {
+	return parallel.New().
+		WithJob("Helm chart", dag.Helm().CheckReleaseDryRun).
+		WithJob("CLI", dag.DaggerCli().CheckReleaseDryRun).
+		WithJob("Engine", dag.DaggerCli().CheckReleaseDryRun).
```
kpenfound:
```go
WithJob("CLI", dag.DaggerCli().CheckReleaseDryRun).
WithJob("Engine", dag.DaggerCli().CheckReleaseDryRun).
```
are these two the same, or typo?

util/parallel/parallel.go
```diff
@@ -0,0 +1,98 @@
...
+	return otel.
+		Tracer("dagger.io/util/parallel").
+		Start(ctx, name, trace.WithAttributes(attr...))
+}
+
+func (job Job) Runner(ctx context.Context) func() error {
+	ctx, span := startSpan(ctx, job.Name)
+	return func() error {
+		defer span.End()
+		if job.Func == nil {
```
sipsma, line 64:
```suggestion
	return func() error {
		ctx, span := startSpan(ctx, job.Name)
		defer span.End()
		if job.Func == nil {
```

Without that change, the span can start but then not actually be running because the `errgroup` is waiting for a parallelism limit to unblock, which would be confusing and make the span look like it was taking much longer than it actually was.

.dagger/sdk_dotnet.go
kpenfound: I think each of the SDKs need to have the same interface for the top level loop to operate

.dagger/sdk_go.go
```diff
@@ -5,55 +5,34 @@ import (
...
+}
+
+func (t GoSDK) Source() *dagger.Directory {
+	// FIXME: use pre-call filtering when this module is spun out
+	// (or when we get self-calls)
+	return t.Dagger.Source.Filter(dagger.DirectoryFilterOpts{
+		Include: []string{
+			"sdk/go",
+			"**/go.mod",
+			"**/go.sum",
```
sipsma, line 27: nit: seems like this would invalidate cache based on quite a few go.mods in the repo that we don't care about here, could be more precise (but not a huge deal)

.dagger/sdk_java.go
shykes: Removing the FIXME. I asked Yves out-of-band, it's not blocking to this PR.

.dagger/sdk_php.go
shykes: Leaving this FIXME to not forget following up.

.dagger/sdk_python.go
shykes: It was nightmarish to debug the generate loop across 2 languages and modules. I would have preferred to leave it alone and just change the wrapper, like I did for the other languages, but I just couldn't get it to work any other way.

Lifting even *one* function was quite painful, I'm happy to continue lifting the rest, but that definitely should be a follow-up - otherwise it will lead to the scope of this PR to balloon.

I agree it's not ideal to make a special case here, but given the benefits I think it's fine.

.dagger/go.mod
shykes: @sipsma ? Does this ring a bell?

modules/go/lint-config.yml
```diff
@@ -33,6 +36,15 @@ linters:
...
         - ifElseChain
+    gosec:
+      excludes:
+        - G115 # stupid
+        - G114 # dumb
+        - G112 # bullshit
+        - G204 # dumb
+        - G306 # dumb
+        - G107 # ridiculous
+
```
sipsma: 😂 this may be worth clarifying

sipsma (approved):
Couple more comments but nothing super blocking, stuff left seems addressable in follow-ups if trying to avoid further rebase hell

util/parallel/parallel.go
shykes: Ah! This is what I did initially, but it leads to the jobs being displayed in a random order every run, which I got comments about. Is there a way to get my cake and eat it too?

modules/go/lint-config.yml
shykes: Hey don't shoot the messenger 😛

.dagger/sdk_go.go
shykes: You're right. Adding a FIXE so we can follow-up

modules/go/main.go
```diff
@@ -142,6 +152,82 @@ type Go struct {
...
+
+	Exclude []string
+}
+
+func (p Go) GenerateDaggerRuntimes(ctx context.Context) (Go, error) {
+	mods, err := p.Modules(ctx)
+	if err != nil {
+		return p, err
+	}
+	daggerMods := map[string]*dagger.Directory{}
```
grouville, line 167: It seems that the goroutines spawned by `parallel.New().WithLimit(3)` concurrently assign into `daggerMods`, which is not fully concurrent safe, example: https://go.dev/play/p/JX-jo2sPsQI

It's only 3 goroutines, on a few different paths here, but, unless we can assure that `daggerModPath` is a different key in each goroutine's assigned map, there's a risk (non blocking as it's very very unlikely)

util/parallel/parallel.go
sipsma: Ah okay, well probably not since I guess the problem is likely that `errgroup` likely ends up choosing a random waiter when multiple waiters are waiting. We could implement our own equivalent of `errgroup.Group` that has the desired FIFO behavior (it's actually a very simple util) but not a blocker. I'm ambivalent as to whether consistent order vs. possible misleading span time is better/worse, so up to you.

shykes: Yeah not sure which is worse either 😭 

For now I'll prioritize getting this merged... Will add a FIXME.

modules/go/main.go
shykes: Concurrency is not an issue in this case, because all the goroutines write the same thing: `nil`. So even if several goroutines write to the same key, it doesn't matter which writer wins: they're all writing the same value.

grouville: It just raises a `fatal error: concurrent map writes` that completely stops the program execution when a race is encountered -> https://go.dev/play/p/ycOqiNu0yf5

shykes: Oh!!! Well we don't want that! OK I'll add a mutex. Thanks!

shykes: I added a mutex ✅

.dagger/go.mod
shykes: For the record: I'm going to merge as-is, but will take responsibility for fixing any issues that arise.

.dagger/sdk_python.go
shykes: Taking responsibility for moving forward as-is, in spite of non-ideal inconsistency, knowing that it will be fixed in a follow-up.

.dagger/go.mod
shykes: It appears that applying this broke the "dev-engine" check in a way that I don't understand. https://github.com/dagger/dagger/actions/runs/18668252943/job/53223921125?pr=11161

shykes:
@jedevc don't ask me why, but when I rebased on #11252, it caused the "dev engine" check to fail 100% of the time. Reverting the commit makes the issue go away. I honestly don't know why...

I initially thought I had screwed up the go.mods when applying your cosmetic suggestion (move the replace to top of file etc), but reverting that didn't fix the issue.

For now I'm leaving the revert commit on top.

shykes:
Back to green :)

jedevc:
Closing and re-opening, something *bizarre* is happening with scheduling in GHA, hoping this restores it to a good state.

Edit: seems to have done it.

jedevc:
I removed both the revert commits and added my own little tidy up commit for the go.mods:
- I ran `dagger develop` and `go mod tidy`
- I did some regrouping of OTEL deps
- I removed the `util/parallel` dep from releaser's `dagger.json`.

There's still an issue, but it's now something different, to do with `git filter-branch`. Investigating. (edit: resolved)

jedevc:
Okay, I'm not 100% sure what I've done, but it's green now :shrug:

shykes:
Thank you everyone!

PR #11162, opened 2025-10-02
llm: fix static tool calling scheme, add eval

vito:
* Fix static method calling tools being discarded
* Add eval coverage, by just re-using the BuildMulti eval + assertion by configuring it for static calling (this worked out nicely!)

### DIFF
--- a/core/mcp.go
+++ b/core/mcp.go
@@ -1594,7 +1594,7 @@ func (m *MCP) loadBuiltins(srv *dagql.Server, allTools, objectMethods *LLMToolSe
 	}
 
 	if m.staticTools {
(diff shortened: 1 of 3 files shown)


shykes (approved):
Thank you!

vito:
note: evals are still unable to run because they just run against the outer engine, whereas previously they would run against a dev engine (hence `testdev`), cc @marcosnils @matipan 

ran locally; merging

marcosnils:
> note: evals are still unable to run because they just run against the outer engine, whereas previously they would run against a dev engine (hence `testdev`), cc @marcosnils @matipan

 ack @vito. We'll change the evals CI tests so they run against the dev engine again :pray:

PR #11163, opened 2025-10-02
tui: limit change summary width

vito:
Helps avoid this scenario:

image

### DIFF
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -733,7 +733,7 @@ func handleChangesetResponse(ctx context.Context, dag *dagger.Client, response a
 			return nil
 		}
 
-		return idtui.SummarizePatch(idtui.NewOutput(&summary), patch, -1)
+		return idtui.SummarizePatch(idtui.NewOutput(&summary), patch, 50)
 	})(); err != nil {
 		return err
 	}


cmd/dagger/functions.go
```diff
@@ -733,7 +733,7 @@ func handleChangesetResponse(ctx context.Context, dag *dagger.Client, response a
 			return nil
 		}
 
-		return idtui.SummarizePatch(idtui.NewOutput(&summary), patch, -1)
+		return idtui.SummarizePatch(idtui.NewOutput(&summary), patch, 50)
```
jedevc, line 736: Any reason for 50? 80 is standard terminal width.

(or tbf, we could just use the current terminal width from the frontend?)

jedevc (approved):

cmd/dagger/functions.go
vito: Not particularly. This appears within a TUI bubble element that has its own padding etc. so the technically correct number would be (terminal width - padding), but it doesn't seem worth the effort to chase that down in `huh` at the moment. FWIW, our sidebar does the right thing (the summary gets called with the available width of the sidebar which can change dynamically).

Since we know there's padding, 50 feels just as wrong as 80, and I'd rather err on the smaller side than exceed the width of an actually-80 width terminal. :P

PR #11166, opened 2025-10-02
✨ `dagger init --template` and `dagger develop --eject`

kpenfound:
This adds functionality to init a module from a template, or to "eject" from a blueprint.

Today if you use a blueprint, you might eventually run into limitations where you need to write customize your module. There's currently no easy path to go from a blueprint to a custom module, you have to start from scratch or copy code.

To solve this, "ejecting" from a blueprint means you will switch your module to an SDK module that implements the original blueprint as a dependency, allowing you to start customizing the functions.

This is the same idea for `dagger init --template <module> --sdk <sdk>`, but you know you want to start customizing on day 1 rather than using the module as a blueprint first.

## Example

Using the [blueprint guide](https://docs.dagger.io/getting-started/quickstarts/blueprint), I start with `hello-dagger` installed as a blueprint in my project, `hello-dagger-template`:

```bash
dagger init --blueprint=github.com/kpenfound/blueprints/hello-dagger
```

```json
{
  "name": "hello-dagger-template",
  "engineVersion": "v0.18.17",
  "blueprint": {
    "name": "hello-dagger",
    "source": "github.com/kpenfound/blueprints/hello-dagger",
    "pin": "2bb1d301e87cbe218b65c99f4e74618cda252638"
  }
}
```

```
$ dagger functions

Name        Description
build       Build the application container
build-env   Build a ready-to-use development environment
publish     Publish the application container after building and testing it on-the-fly
test        Return the result of running unit tests
```

Now I want to start customizing my module, so I need to eject from the blueprint into a custom module:

```bash
dagger develop --eject --sdk go
```

```json
{
  "name": "hello-dagger-template",
  "engineVersion": "v0.19.1",
  "sdk": {
    "source": "go"
  },
  "dependencies": [
    {
      "name": "hello-dagger",
      "source": "github.com/kpenfound/blueprints/hello-dagger@main",
      "pin": "2bb1d301e87cbe218b65c99f4e74618cda252638"
    }
  ],
  "source": ".dagger"
}
```

The `hello-dagger` module is now a dependency. Better yet, my generated `.dagger/main.go` doesn't have `ContainerEcho` or `GrepDir`, it has the same interface my blueprint module had, so existing commands should continue working as I begin customizing my module:

```go
// A generated module for HelloDaggerTemplate functions
//
// This module has been generated from a blueprint and provides passthrough
// functions to the original module.

package main

import (
        "context"
        "dagger/hello-dagger-template/internal/dagger"
)

type HelloDaggerTemplate struct{}

// Build the application container
func (m *HelloDaggerTemplate) Build() dagger.Container {
        return dag.HelloDagger().Build()
}

// Build a ready-to-use development environment
func (m *HelloDaggerTemplate) BuildEnv() dagger.Container {
        return dag.HelloDagger().BuildEnv()
}

// Publish the application container after building and testing it on-the-fly
func (m *HelloDaggerTemplate) Publish(ctx context.Context) (string, error) {
        return dag.HelloDagger().Publish()
}

// Return the result of running unit tests
func (m *HelloDaggerTemplate) Test(ctx context.Context) (string, error) {
        return dag.HelloDagger().Test()
}
```

TODO

- [ ] Python generation
- [ ] Typescript generation

### DIFF
--- a/cmd/codegen/generator/go/generate_module.go
+++ b/cmd/codegen/generator/go/generate_module.go
@@ -84,7 +84,9 @@ func (g *GoGenerator) GenerateModule(ctx context.Context, schema *introspection.
 
 	if len(initialGoFiles) == 0 {
 		// write an initial main.go if no main pkg exists yet
-		if err := mfs.WriteFile(StarterTemplateFile, []byte(baseModuleSource(pkgInfo, moduleConfig.ModuleName)), 0600); err != nil {
+		dependencies := extractDependenciesFromSchema(schema)
+		moduleSource := generateModuleSource(schema, pkgInfo, moduleConfig.ModuleName, dependencies)
+		if err := mfs.WriteFile(StarterTemplateFile, []byte(moduleSource), 0600); err != nil {
 			return nil, err
 		}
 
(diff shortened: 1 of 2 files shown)


kpenfound:
@shykes @eunomie @grouville @TomChv 

Starting a discussion here about the SDK side of this.

The way this works basically is piggy backing on the dagger init flow of the module codegen. So the module is being initialized and there is no existing module code.

- For the purposes of "ejecting" from a blueprint, we switch the blueprint module to a dependency at this point in time, so a *new* module is being generated and it has a dependency.
- When we generate the module code, if there are 0 dependencies, we return the static template with ContainerEcho and GrepDir (same behavior as before)
- If there is a dependency (we are ejecting from a blueprint), we look at the schema of that dependency from the existing introspection, and generate code that produces passthrough functions to the dependency's top level functions
- The cool thing is, this passthrough generation could potentially be used for a lot more than this eject feature

To kick off the discussion, some initial questions from talking with @shykes :
- This currently piggy backs on the existing GenerateModule interface, but should it be a new interface?
- What implications does this kind of code have on SDK maintainers?
- How does this affect / is this affected by ongoing SDK changes @TomChv @eunomie ?

eunomie:
@kpenfound This looks really cool!

I'll have a deeper look about how it's done first, but yes one of the functions we wanted to extract (in some way) from the existing codegen is the creation of the template. Right now it's simply to create the `container-echo/grep-dir` but I think it deserve it's own function. That way we could also imagine to have different templates. But it also means we have a clear common place for every SDK to provide their template instead of having that deep inside the SDKs implementations.

While writing it also makes me think: it could be also really good to have a way to call the `eject` from start. In this case it's a bit like a new template, but it's way more powerful due to the dependency + codegen as you can have it on the language of your choice.
But I'm pretty sure we could imagine cases where you want to have from start the ejected version. That way it serves as a baseline and you know you will have extra configuration/targets for instance.

But maybe that could be in a follow up.

eunomie:
My first test didn't work

```
! failed to serve module: error serving dependency hello-dagger: module hello-dagger (source:
  github.com/kpenfound/blueprints/hello-dagger@main | pin: 2bb1d301e87cbe218b65c99f4e74618cda252638) already exists with
  different source /Users/yves/dev/src/github.com/dagger/hello-dagger (pin: )
```

This is because I'm doing that in a clone of `hello-dagger` project and the dependency (the blueprint) is also `hello-dagger`. It's not really because of this code, and the solution is to either rename the module or the dependency, and that's easy to do.

But maybe there's a way to check that when we are performing the `--eject` and rename the dependency? Just to ensure the first `dagger call` or `dagger functions` after the `--eject` will work.

eunomie:
OK, so if I understand correctly the actual implementation, the `develop --eject` will just transform the blueprint ref in a dependency, and then the codegen will "just" create passthrough functions if there's dependencies. right?
So regarding what I wrote earlier, we could imagine to have a combo `init --blueprint` + `develop --eject` by just adding the module as a dependency before the codegen and that should work automatically, right? That's cool.

Also, I wonder, especially if we extract the template function on SDKs, if we shouldn't improve the way we are passing the schema to the SDK modules. Because we are duplicating some logic everywhere (the stuff on introspectionJSON file) and we are depending on `cmd/codegen` that doesn't looks great to me. All those types should be on their own package, or even provided in some way by the engine, and not from a `cmd` package.
But that's great, this is going in the direction we want to go with the SDKs.

kpenfound:
@eunomie thanks for the replies!

> Right now it's simply to create the container-echo/grep-dir but I think it deserve it's own function. That way we could also imagine to have different templates. But it also means we have a clear common place for every SDK to provide their template instead of having that deep inside the SDKs implementations.

Thats exactly the kind of thing I'm hoping to discuss! Maybe there's even a path to deleting the templates and generating the basic module from a schema definition if the module generation stuff is generalized a bit more

> While writing it also makes me think: it could be also really good to have a way to call the eject from start. In this case it's a bit like a new template, but it's way more powerful due to the dependency + codegen as you can have it on the language of your choice.

So that actually works already if you do `dagger init --blueprint <module> --eject --sdk <sdk>`. Originally I had an alias where you could do `dagger init --template <module> --sdk <sdk>` but the coexisting blueprint + template could be very confusing

> Also, I wonder, especially if we extract the template function on SDKs, if we shouldn't improve the way we are passing the schema to the SDK modules. Because we are duplicating some logic everywhere (the stuff on introspectionJSON file) and we are depending on cmd/codegen that doesn't looks great to me. All those types should be on their own package, or even provided in some way by the engine, and not from a cmd package.

I was wondering about that as well. Right now the duplicated code is very _slightly_ different because it converts the graphql types to the SDK specific type name, however the code is similar enough we could simplify it

kpenfound:
Design is still TBD, especially with toolchains in the picture https://github.com/dagger/dagger/pull/11238

PR #11170, opened 2025-10-03
PHP: Avoid installing dev dependencies for end-users

charjr:
- Upgraded dependencies
- Moved graphql and code generation from "require-dev" to "require" since they are required by all users.
- Use `composer install --no-dev` so that end-users don't download dev-dependencies in their SDK runtime; reducing the memory size of the SDK for end-users.

PS: The `sdk/php/dev` module takes the SDK directory and installs dev dependencies, so it is unaffected.


sdk/php/runtime/main.go
```diff
@@ -89,7 +89,7 @@ func (m *PhpSdk) CodegenBase(
 		WithDirectory("/sdk", m.SourceDir).
 		WithWorkdir("/sdk").
 		// Needed to run codegen
-		WithExec([]string{"composer", "install"})
+		WithExec([]string{"composer", "install", "--no-dev"})
```
jedevc, line 92: Funnily enough we also discussed about this for typescript in discord here: https://discord.com/channels/707636530424053791/1157515934597136404/1423607325607858227

@TomChv seem reasonable to you to do this across the board? Feels like a good consistency thing.

TomChv: Done at https://github.com/dagger/dagger/pull/11182

PR #11171, opened 2025-10-03
ci: update all github runners to 0.19.0

jedevc:
Looks like this was missed during https://github.com/dagger/dagger/pull/11150.

There are still a few versions that need manually bumping and aren't automated.

This was causing publish to fail on main.


matipan (approved):

matipan:
Merging. Failures are 429 errors from docker.io

PR #11172, opened 2025-10-03
ingore unnecessary files in docs context

marcosnils:
this speeds up module loading considerably when you have dirty files
in your working directory


shykes (approved):
Thanks for catching this!

Note for a follow-up: we can probably include explicitly files we *do* need, and omit the rest (it's specifically for generating API reference etc)

PR #11128, opened 2025-09-25
docs: update container runtime docs

jedevc:
As discussed in a team call earlier this week, the docs for the other container runtimes aren't right.

Since https://github.com/dagger/dagger/pull/10714, we actually do not want to be symlinking things, pretending that they are docker - because they aren't. We actually natively support these.

This PR updates the docs for how to setup all of these.

Issue #9475, opened 2025-01-29
Netlify docs deployment sometimes requires manual rebuild to update examples


jedevc:
Sometimes, the code snippets aren't updated from the netlify build, and manual intervention is neccessary:

![Image](https://github.com/user-attachments/assets/de77820b-82ad-4ed5-a949-0945bc7fe885)

This hasn't really gone away, and is a bit annoying - we should either:
- Disable the cache entirely, and suffer through the slower deployment times
- Fix the underlying issue (ideally)

Just opening this to track, not 100% sure what's actually going on.

FYI @vikram-dagger

vikram-dagger:
I've noticed this as well and still haven't found an identifiable pattern for when the cache is refreshed vs when not. As best I can see, if there's a change in an `include`-d file (as opposed to an actual output page), it's not picked up as a change and the cache gets used. However, even this doesn't happen every time.

PR #11180, opened 2025-10-06
docs: move kubernetes and openshift to deployment url

jedevc:
During https://github.com/dagger/dagger/pull/11128, we moved the Kubernetes and OpenShift pages to a new Deployment section, which fits them better. See the justification here: https://github.com/dagger/dagger/pull/11128#discussion_r2379448890.

However, I didn't update the slugs, so the pages were still at the same place.

This patch moves them to the new locations, and updates the redirects.


jedevc:
Netlify check is failing, but retriggered without cache and it works: https://deploy-preview-11180--devel-docs-dagger-io.netlify.app/

See https://github.com/dagger/dagger/issues/9475.

marcosnils (approved):

PR #11182, opened 2025-10-06
chore(typescript): only install prod dependencies in module runtime

TomChv:
Update module dependency installation to only install production dependencies when its available.

*Note*: This flag doesn't exist on deno since the runtime has no `dev/prod` separation.


jedevc (approved):

TomChv:
> SGTM, but this is a potentially breaking change?
> 
> Should probably at least changelog it with `Changed`.

Only for users who installed dev dependencies, I'll add a note

PR #11183, opened 2025-10-06
Fixes for changeset previews

### LINKED ISSUE #11168: 🐞 `Changes()` does not handle directory deletion
### What is the issue?

`Changes()` handle directory deletion in unexpected ways.

Given the following Dagger module:

```go
func (l *Mod) One(
	// +defaultPath="."
	source *dagger.Directory,
) *dagger.Changeset {
	return source.WithoutDirectory("bar").Changes(source)
}

func (l *Mod) Two(
	// +defaultPath="."
	source *dagger.Directory,
) *dagger.Changeset {
	return source.WithoutFile("foo").Changes(source)
}

func (l *Mod) Three(
	// +defaultPath="."
	source *dagger.Directory,
) *dagger.Changeset {
	return source.WithoutDirectory("bar").WithoutFile("foo").Changes(source)
}
```

In a situation where both `./foo` and `./bar/baz` files exist:

```sh
dagger call one    # no changes to apply
dagger call two    # prompts to remove foo
dagger call three  # only prompts to remove foo but removes both foo and bar when accepted
```

I expect both `one` and `three` to prompt for `bar` removal.

### Dagger version

dagger v0.19.0 (image://registry.dagger.io/engine:v0.19.0) darwin/arm64/v8

### Steps to reproduce

_No response_

### Log output

_No response_

jedevc:
Fixes #11168 (and other similar bugs).

See individual commits for details, there are two changes here:
- Include removed nested files in `Changes.asPatch` (this ensures that the changeset preview includes line counts of those files)
- Update handling of changeset previews with https://github.com/jedevc/diffparser, which ensures proper mode detection, so that we correctly distinguish between new/deleted files.

Here's a little example. Before:
```
$ touch foo && mkdir foo.d
$ dagger core host directory --path=. without-directory --path foo.d/ with-new-directory --path=bar.d without-file --path foo with-new-file --path=bar --contents="" changes --from=.
```
image

After:
```
$ touch foo && mkdir foo.d
$ dagger-dev core host directory --path=. without-directory --path foo.d/ with-new-directory --path=bar.d without-file --path foo with-new-file --path=bar --contents="" changes --from=.
```
image


vito (commented):
Nice, thanks for fixing the parser!

> Include removed nested files in Changes.asPatch (this ensures that the changeset preview includes line counts of those files)

I think we should include the line counts, but they should be rolled up into their parent directory in the summary. The idea is to prevent removal of a directory from dominating the UI:

Image

Repro:

```sh
dagger-dev core host directory --path=. without-directory --path=./core changes --from=./
```

jedevc:
Made it so that directory removals absorb their children's line counts. There's no corresponding case for additions, but I *feel* that's kind of expected? It's easy to add if we want that as well.

Your repro now gives:

image

My example from above now gives (after populating the files with a bit of content, and putting some stuff in `foo.d`):

image

marcosnils (approved):
:thumbsup:

PR #11184, opened 2025-10-06
Fix light mode in shell

jedevc:
Reapplies https://github.com/dagger/dagger/pull/10266.

This seems to have been removed in #10907, which sadly makes light mode unusable again.


vito (approved):
🙏 Sorry about that! Looks like this got lost when I swapped from Monokai to just respecting the terminal colors.

PR #11187, opened 2025-10-06
fix(python_sdk): register all enum members and add introspection test

grouville:
Fix `Module._register` so we add the enum once after building all members. Before this, only the first value (GO) ever showed up. Dropped in a Python integration test that introspects the module and expects all Language members, so we’ll catch it next time.


grouville:
My bad, it was a mistake on my end. This actually properly works already

PR #11199, opened 2025-10-08
.dagger: add coder

vito:
Adds a coding agent for developing Dagger.

### DIFF
--- a/.dagger/main.go
+++ b/.dagger/main.go
@@ -112,6 +112,32 @@ func (dev *DaggerDev) SourceDeveloped(targets ...string) *dagger.Directory {
 	return src
 }
 
+// Start a coding agent for the Dagger project.
+func (dev *DaggerDev) Coder(ctx context.Context) (*dagger.LLM, error) {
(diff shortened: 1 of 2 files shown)


vito:
@eunomie 

> That looks really nice :-)

Thanks! Next step is to be able to point it at a GitHub issue and let it research and start working on its own. Working on that now. :smile: 

> Should `Doug` be part of dagger API at some point?

If you mean 'part of the Dagger repo' as in `modules/doug/`, maybe someday, not sure when. :thinking: (As in I don't know if there's an appropriate time or if I should just do it now.)

If you mean adopting its functionality natively in `core/mcp.go` as built-in tools for all agents - I don't think so, it's pretty nice DX + dogfooding to implement Doug as a module.

eunomie:
I was mostly thinking about how to offer this kind of capabilities (to create custom agents) to users the easiest way, without for them to rebuild the equivalent. Like could we imagine at some point to write an agent with `dag.Agent(dag.LLM()....)`

Issue #11189, opened 2025-10-07
API: Merge Changesets


shykes:
## Problem

The new `Changeset` type is awesome, but lacks the ability to merge Changesets.

A workaround is possible, but requires applying all changesets to a common ancestor directory, then re-computing changes. This is unnecessarily verbose, and possibly less inefficient. It also requires access to a common ancestor, which is not always handy. I've had this problem in our own CI for SDK generation.

## Solution

Implement `Changeset.merge` (name TBD).

Possible names:

- `Changeset.merge`
- `Changeset.withChangeset`

shykes:
FYI @vito

vito:
Quick note: for this to work I think we'll want to think of it like a 3-way merge of patches (or applying patches repeatedly, depending on the API shape). That is, if Changeset A modifies lines 3-5 of `foo.go` and Changeset B modifies lines `80-89` of the same file, the merged Changeset should reflect both changes, not just whichever one ended up being the upper layer.

I mention this because I'm pretty sure you _won't_ get that behavior by repeatedly applying changesets to the common ancestor.

eunomie:
While working on `generate` feature, I'm having a look at this.

I think I understand the question of multiple changesets modifying the same files and merging them, but I guess that only applies to text files. If the file is a binary for instance that can't work.

But I guess we don't have an easy way to know the type of the file and if we want them to be merged. An automated merged might be completely wrong and might require a manual decision.

So maybe one way is first to detect if there's any kind of conflict (same file modified multiple times) and if not just apply everything. And if there's conflicts 🤔 fail with a message describing the files in conflict until we have a tool to help the user resolve them?

At least this could be a plan to start moving on this topic and have a single place where we are merging changesets, and a place to solve that issue later. WDYT?

shykes:
yes that's fine. We could have an optional strategy argument, and add more strategies to the enum over time.

Right now in our own toolchains we implement our own version, with a "last write wins" strategy

idlsoft:
I just used `Changeset` in my build, it's great, but I wish it had something like `WithBase` or `WithRoot`.
Something that  would do this:
```go
func ChangesetWithBase(dag *dagger.Client, oldDir *dagger.Directory, newDir *dagger.Directory, base string) *dagger.Changeset {
	return dag.Directory().WithDirectory(base, newDir).Changes(dag.Directory().WithDirectory(base, oldDir))
}
```

vito:
[...]

PR #11200, opened 2025-10-08
llm: don't parallelize `Env`/`Changeset` APIs

vito:
Instead of assuming all non-MCP server tools are ReadOnly, only consider them ReadOnly if they return something other than Env/Changeset. Otherwise, when calling a tool such as EditFile, only one of their results will "win."

## Alternatives considered

Theoretically we could merge concurrently-created `Changeset`s with #11189 - but it seems safer and simpler to just apply them one-by-one since these calls shouldn't take very long.

### DIFF
--- a/core/mcp.go
+++ b/core/mcp.go
@@ -522,13 +522,24 @@ func (m *MCP) typeTools(allTools *LLMToolSet, srv *dagql.Server, schema *ast.Sch
 		} else {
 			toolName = typeDef.Name + "_" + field.Name
 		}
+
 		allTools.Add(LLMTool{
 			Name:        toolName,
 			Field:       field,
 			Description: strings.TrimSpace(field.Description),
 			Schema:      toolSchema,
-			Strict:      false, // unused
-			HideSelf:    autoConstruct == nil,
+
+			// TODO: would be nice, but have to 'or-null' all args and list everything
+			// in required, which is annoying.
+			Strict: false,
+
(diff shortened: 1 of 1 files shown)


core/mcp.go
```diff
@@ -1156,7 +1173,18 @@ func (m *MCP) CallBatch(ctx context.Context, tools []LLMTool, toolCalls []LLMToo
 		allResults = append(allResults, m.callBatchRegular(ctx, tools, readOnlyToolCalls)...)
 	}
 
-	// 3. Execute destructive MCP calls one server at a time to avoid workspace conflicts
+	// 3. Execute destructive non-MCP calls sequentially (they modify Env/Changeset state)
+	for _, call := range destructiveCalls {
+		result, isError := m.Call(ctx, tools, call)
```
eunomie, line 1178: Dumb question, but what happen if the call raises an error? The next call are still performed?

vito: Yeah - at this layer the job is to just pass along the result of all tool calls, including failures, since LLM APIs strictly expect every tool call to be paired with a response. In this case, the LLM technically requested a batch of tool calls all at once, which normally just runs them all in parallel, but our implementation has to be slightly smart about managing them. Traditionally these calls would all be running in the same shared mutable environment, but ours are actually sandboxed, so to avoid "last-write wins" shenanigans we serialize ones that are destructive and only run "safe" tools in parallel. From the LLM's perspective these all happened independent of one another so there's no need to stop early.

PR #11158, opened 2025-10-01
Live development

jedevc:
Implements https://github.com/dagger/dagger/issues/10895.

> [!WARNING]
>
> This is not ready, opening a PR to track work.

TODO:
- [ ] `Container.wait`/`step` (kind of like sync, but guaranteed to be different ID)
- [x] `Container.watch`
  - [ ] `Directory.watch` / `File.watch`
- [x] `Container.asService(watch: bool)`
	- [ ] Dynamic service bindings

PR #11204, opened 2025-10-08
Little refactors during work on livedev

jedevc:
Extracting refactoring work from #11158, since some of these are useful (and I've wanted separately).

Also, it seems likely that the entire structure of that PR will change, so wanting to extract as much utility from that as possible.

Mostly changes to IDs and Services, so cc @vito

### DIFF
--- a/core/service.go
+++ b/core/service.go
@@ -676,7 +676,6 @@ func (svc *Service) startContainer(
 		}
 
 		return &RunningService{
-			Service:     svc,
 			Host:        fullHost,
 			Ports:       ctr.Ports,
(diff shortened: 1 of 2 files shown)


jedevc:
Also pulled in my little ID visitor that allows mutating the graph, and some tests I wrote for it. Then, rewrote `WalkIDs` to use it.

The mutation part isn't actually used anywhere *yet*, but it's come up a few times as potentially useful, so I didn't want it to get completely lost.

jedevc:
Hm I've broken secret transfer - dropped the walk ids stuff,

PR #11211, opened 2025-10-09
dagger checks

shykes:
Checks are a thin opinionated layer on top of dagger functions. Run `dagger checks` to find and run all check functions in your project - locally or in CI.

## Testing

You can test this with a special branch of our CI which takes advantage of it:

Starting from a stable release of dagger:

```console
#!/usr/bin/env/dagger -M

github.com/shykes/dagger@checks |
 dev |
 with-workdir /src/dagger |
 with-directory . https://github.com/shykes/dagger#ci-checks: |
 terminal
```

Then once in an interactive terminal:

- List all checks: `dagger checks -l`
- List only lint-related checks: `dagger checks '**/lint**' -l`
- List all SDK checks: `dagger checks 'sdk/**' -l`
- Run a given selection `dagger checks go/lint`

## TODO

- [x] Tests
- [x] Repro & fix infinite recursion issue reported by @tiborvass 
- [x] Fix disappearing emoji/messages
- [x] Fix "telemetry black box" issue (individual run streams shows no telemetry)
- [x] Fix issue with `dagger checks '**/test'` (leading `**`)
- [x] Get more useful errors / logs from failed checks
- [x] Remove useless `message` field
- [x] Pretty-print logs of failed checks
- [ ] Integrate with #11238 
- [ ] Exit 1 on failed checks
- [ ] Switch to pragmas/decorators for exporting checks
- [ ] Clean up / refactor CLI frontend
- [ ] Add `Check.Logs: File!` so other clients can get useful info without complex telemetry rendering
- [ ] Resolve weird otel attribute decoding error
- [ ] Tweak ergonomics of filtering argument (regexp or glob? delimited matching?)

## Merge plan

1. Merge backend as-is @shykes 
2. Merge frontend as-is with light cleanup @shykes 
3. Follow-up: fix DX (pragmas etc) @vito 
4. Follow-up: refactor frontend @vito


core/schema/checks.go
```diff
@@ -0,0 +1,68 @@
...
+	dagql.Fields[*core.Check]{
+		dagql.Func("name", s.name).
+			Doc("Return the fully qualified name of the check"),
+		dagql.Func("resultEmoji", s.resultEmoji).
+			Doc("An emoji representing the result of the check"),
+	}.Install(srv)
+}
+
+func (s checksSchema) checks(ctx context.Context, q *core.Query, args struct {
+	Include dagql.Optional[dagql.ArrayInput[dagql.String]]
```
vito: nit: I think this can be simplified to `*[]string`, or `[]string` with a default:


```suggestion
	Include *[]string // same API
```

```suggestion
	Include []string `default"[]"` // "required with default" - if you want to avoid pointer
```

I guess the ideal would be to leave the arg optional with a `null` default and without a pointer on the server side, but don't think it exists:

```suggestion
	Include []string `optional:"true"`
```

If you want to add that, wouldn't be too hard - [here's the area](https://github.com/dagger/dagger/blob/03d3cf86b02f690b6cf5571b2dd2db48be1483d9/dagql/objects.go#L1458)

But, super nitpick, doesn't really matter

core/checks.go
```diff
@@ -0,0 +1,283 @@
...
+	return CheckStatuses
+}
+
+func (r CheckStatus) ToLiteral() call.Literal {
+	return CheckStatuses.Literal(r)
+}
+
+// Check represents a validation check with its result
+type Check struct {
+	Path        []string `field:"true" doc:"The path of the check within its module"`
```
vito, line 49: Check hierarchy path, not file path, right? Wonder if there's a terminology tweak that would clarify it. What does Go call it?

```
        -run regexp
            Run only those tests, examples, and fuzz tests matching the regular
            expression. For tests, the regular expression is split by unbracketed
            slash (/) characters into a sequence of regular expressions, and each
            part of a test's identifier must match the corresponding element in
            the sequence, if any. Note that possible parents of matches are
            run too, so that -run=X/Y matches and runs and reports the result
            of all tests matching X, even those without sub-tests matching Y,
            because it must run them to look for those sub-tests.
            See also -skip.
```

...ah, it dances around it and avoids calling it a path.

core/checks.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func (r *CheckGroup) List(ctx context.Context) ([]*Check, error) {
+	return r.Checks, nil
+}
+
+// Run all the checks in the group
+func (r *CheckGroup) Run(ctx context.Context) (*CheckGroup, error) {
+	attr := []attribute.KeyValue{
+		attribute.Bool("dagger.io/ui.reveal", true),
```
vito:
```suggestion
		attribute.Bool(telemetry.UIRevealAttr, true),
```

core/checks.go
```diff
@@ -0,0 +1,283 @@
...
+			defer func() {
+				if rerr != nil {
+					span.SetStatus(codes.Error, rerr.Error())
+				}
+				span.End()
+			}()
+			// Reset output fields, in case we're re-running
+			check.Completed = false
+			check.Passed = false
+			check.Message = ""
```
vito: So the idea is mutation is OK within this object (normally unusual in DagQL but has precedence with `LLM` - for now), because it's already descending from a pinned module ref, and in a user's mind there's just a 'set of checks' for that module ref, so it's OK to maintain a single state?

What about keeping knowledge re-runs? Usually you see a UI cue like `main@deadbeef > integration #3` for the third run.

(Non-blocking)

core/module.go
```diff
@@ -81,6 +81,146 @@ func (mod *Module) Name() string {
...
+		func() {
+			_, span := Tracer(ctx).Start(ctx, fmt.Sprintf("%s() returnType=%s returnsObject=%v", fn.Name, fn.ReturnType.ToType().Name(), fn.ReturnType.AsObject.Valid))
+			defer span.End()
+
+			if functionRequiresArgs(fn) {
+				return
+			}
+			// 1. Scan object for checks
+			returnType := fn.ReturnType.ToType().Name()
+			if strings.HasSuffix(returnType, checkReturnTypeSuffix) {
```
vito, line 170: What's the idea behind a suffix check vs. checking for the exact `CheckStatus` type?

Also if there was any IRL bikeshedding over the return type cue vs. a pragma I missed it - curious to hear the pros/cons. I have some initial hesitance around requiring the function signature to change, but I do like that it's something we determine from types now instead of naming conventions, and it's only replacing `Void` so not much lost there. (I'm talking myself into it at this point but still feel like I missed something, so just curious!)

core/checks.go
shykes: I could call it a "function path"?

core/checks.go
shykes: I didn't understand this :( Can we discuss it live when you have time?

core/module.go
shykes: > What's the idea behind a suffix check vs. checking for the exact CheckStatus type?

It's an escape hatch to facilitate testing in real-world modules before the feature is merged and released. It allows modules to define their own `CheckStatus` type as a stopgap, and this build will pick it up. For example: https://github.com/dagger/dagger/pull/11262

shykes: > Also if there was any IRL bikeshedding over the return type cue vs. a pragma I missed it - curious to hear the pros/cons. 

You didn't miss anything, this is it 😁

My issue with pragmas is that it's work per-SDK, and compared to how easy this was, it seems like a waste of effort.

Also, we already look at the signature anyway (no required arguments, return value must be X), so it seemed clean to do *everything* with the signature.

core/schema/checks.go
shykes: Thanks! I'll try `[]string` with default. Much simpler!

core/module.go
```diff
@@ -82,6 +82,146 @@ func (mod *Module) Name() string {
...
+				}
+
+			}
+			return
+		}()
+	}
+	// Also walk fields for sub-checks
+	for _, field := range obj.Fields {
+		func() {
+			_, span := Tracer(ctx).Start(ctx, fmt.Sprintf("%s() returnsObject=%v", field.TypeDef.AsObject.Valid))
```
tiborvass: Extra %s, forgot to debug field name ?
```suggestion
			_, span := Tracer(ctx).Start(ctx, fmt.Sprintf("%s() returnsObject=%v", field.Name, field.TypeDef.AsObject.Valid))
```

core/module.go
```diff
@@ -82,6 +82,146 @@ func (mod *Module) Name() string {
...
+			}
+			// 2. Recursively scan reachable children objects
+			if returnsObject := fn.ReturnType.AsObject.Valid; returnsObject {
+				subObj, ok := mod.ObjectByName(fn.ReturnType.ToType().Name())
+				if ok {
+					subObjects[fn.Name] = subObj
+				}
+
+			}
+			return
```
tiborvass: unnecessary

core/module.go
```diff
@@ -82,6 +82,146 @@ func (mod *Module) Name() string {
...
+			returnType := fn.ReturnType.ToType().Name()
+			if strings.HasSuffix(returnType, checkReturnTypeSuffix) {
+				checks = append(checks, &Check{
+					Path:        []string{gqlFieldName(fn.Name)},
+					Description: fn.Description,
+				})
+				return
+			}
+			// 2. Recursively scan reachable children objects
+			if returnsObject := fn.ReturnType.AsObject.Valid; returnsObject {
```
tiborvass: @shykes This fixes my infinite recursion issue.

```suggestion
			if returnsObject := fn.ReturnType.AsObject.Valid; returnsObject && returnType != obj.Name {
```

shykes: oooh, thank you :)

tiborvass:
I find myself needing a Sync(ctx) method on CheckGroup, since I need to detect errors early when a module fails to build.

shykes:
Notes from discussion with @vito today:

- The `message` field is not useful, because it shows the top-level call's error message, which itself is not useful.
- Instead `dagger checks` should show red/green, and render the check's trace in a human-readable way. As a starting point, the multiplexed logs with a simple bracket-prefix.
- Filtering: would be cool to either make it easy to pipe the list through `grep` for custom filtering, or allow a grep-like filter that is simpler than globs. (easier to remember eg. `lint` than `**/lint*`).
- Virtual checks? What if checks could report additional sub-checks, wich are not actual dagger functions, but are reported as such in the telemetry? This is a crossover with #9327 . @vito mentioned: perhaps we can implement an equivalent of `testctx.Run()`: run sub-functions in a user-visible way, with magical support for filtering. --> @vito this is growing on me by the minute! Would solve a real need... which is more fine-grained filtering. We have a way to say "only run the go linter" but not "run the go linter on only this go module". Maybe this is the solution! So we could solve fine-grained reporting + fine-grained filtering in one fell swoop 😍 

FYI @dagger/maintainers

shykes:
@kpenfound I tried transposing your "checks + toolchains" fix, and I successfully installed the go toolchain + listed checks, **BUT** when running them, I get this dagql error:

```
Name            Result   Message
go/check-tidy   🔴        cannot sub-select Go!
go/lint         🔴        cannot sub-select Go!
```

https://dagger.cloud/dagger/traces/e37bf3f49de18d9f90d58d462cd7dfcd

kpenfound:
@shykes I suspect I was hitting that on my branch too, but just wasn't seeing the error because `Message` wasn't working on the commit I was on. I had the same sus 0.7s for each check:
```
▼ .run: CheckGroup! 0.7s
  ✔ go/golangCiLint 0.7s
  ✔ go/test 0.7s
```

I'll work on this branch now!

core/checks.go
shykes: @vito forgot to ask you about this live yesterday...

cmd/dagger/check_logs.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// Render logs for each check span
+	for _, span := range checkSpans {
+		if err := clr.renderSpanLogs(db, w, span); err != nil {
+			fmt.Fprintf(w, "Error rendering logs for span %s: %v\n", span.Name, err)
+			continue
+		}
+	}
+
+	return nil
```
shykes: 1. Searches telemetry DB for spans with check attribute
2. For each check span, searches for log entries and prints them

This part is vibe coded... @vito does the telemetry-specific part seem right to you? So far I'm getting matches on the attributes, but not on the logs..

shykes:
@vito re: your preference for "grep-style" filtering. I was thinking of just making that the default - basically regexp matching instead of custom glob.

So this: `dagger checks lint` would be like `| grep lint`.

wdyt?

vito:
@shykes That sounds good to me, the only thing I'd worry about is false positives, like `test` matching `latest` or something. But maybe that's a concern for both cases?

Maybe we could do a word-anchored grep? ...with special handling for camel-casing?

cmd/dagger/checks_frontend.go
```diff
@@ -0,0 +1,601 @@
...
+		if cleanup != nil {
+			defer cleanup()
+		}
+		return err
+	}
+
+	// Start the TUI with alternate screen (clears on exit)
+	fe.program = tea.NewProgram(fe, tea.WithAltScreen())
+
+	if _, err := fe.program.Run(); err != nil {
```
grouville, line 133: Here, it seems that, similarly to above:

```go
cleanup, err := f(ctx)
if cleanup != nil {
	defer cleanup()
}
```

we do need `defer` the execution of the `.program.cleanup` which was set as part of the `.program.Run()`'s `eventloop()`, otherwise we might skip the telemetry flush and leak an engine session ?

Non blocking as everything will be changing in this file + it's a hidden / unused feature atm, BUT, might be worth keeping / double checking ?

PR #11221, opened 2025-10-11
feat: allow pins in module ref

### LINKED ISSUE #11117: Support for pinned module refs
When calling a module in CI (like [here](https://github.com/dagger/dagger/blob/09b87500ac9883eda96cc38470ce22a084e235ab/.github/main.go#L20-L27)), you can choose between either:

- Addressing it via ref
  ```
  $ dagger -m github.com/dagger/dagger@main
  ```
  This is bad for handling git events because it's unpinned; between receiving the event and dagger processing the event, `main` might get updated causing CI to run on the wrong commit.
  
- Addressing it via commit sha
  ```
  $ dagger -m github.com/dagger/dagger@3dd66bb0c182fbf0b72ca7b986985ec506027af4
  ```
  This is bad for features like contextual git - the "original" ref, and the metadata like, if this is a tag/branch/pr etc gets lost. Which would be super useful (e.g. being able to run a different path for a tag release vs a commit release)

We should support a syntax for doing *both* at the same time - so that you can get the ref, but also pin.

## Options

- Don't change the module ref syntax, but allow an additional `pin` option on the CLI.
  ```
  $ dagger -m github.com/dagger/dagger@main --pin 3dd66bb0c182fbf0b72ca7b986985ec506027af4
  ```
- Split the version field with a separator character:
  ```
  $ dagger -m github.com/dagger/dagger@main:3dd66bb0c182fbf0b72ca7b986985ec506027af4
- Allow query parameters
  ```
  $ dagger -m github.com/dagger/dagger@main?pin=3dd66bb0c182fbf0b72ca7b986985ec506027af4
  ```

## Related

- https://github.com/dagger/dagger/pull/11092 for other pinning problems

jedevc:
Fixes #11117.

Now, a module ref can include a pin. For example:
    
	github.com/dagger/dagger@main:210bdce1068db4170b8cc5120d11d8ef8c81f755
    
This creates a contextual git ref where the ref=main and the commit=210bdce1068db4170b8cc5120d11d8ef8c81f755.
    
Also, fixes a bug where installing github.com/dagger/dagger@210bdce1068db4170b8cc5120d11d8ef8c81f755 would set ref=main, even though main wasn't specified anywhere. This was incorrect.

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -2,7 +2,6 @@ package core
 
 import (
 	"bytes"
-	"cmp"
 	"context"
 	_ "embed"
 	"encoding/base64"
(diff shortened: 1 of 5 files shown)


jedevc:
@shykes does this syntax seem reasonable to you? It *does* feel a little weird to me, but also we need a way of doing this (see context in #11117)

shykes:
syntax seems fine to me, wdyt @vito @sipsma @marcosnils ?

vito:
A bit unfortunate that it's the opposite of OCI refs: `url@<tag>:<pin>` vs `url:<tag>@<pin>`.

But I guess we've already crossed that bridge by aligning with Go module refs instead. Somehow `:` just doesn't feel like a 'more narrowly specifying' symbol than `@` but I can't tell if that's just because the OCI syntax is so ingrained in my psyche.

Would it be possible to just flip it? Feels like it's mostly a matter of retconning it alongside the original `foo@bar` syntax. Either way, when both `@` and `:` are present we're no longer playing by Go module ref rules, so we should be able to do whatever we want, right?

In the end, I feel like we'd want both `dagger install github.com/foo/bar:main` and `dagger install github.com/foo/bar@main` to put the equivalent of `github.com/foo/bar:main@abcdef` in `dagger.json`. You can draw parallels to how Go accepts logical refs like `github.com/foo/bar@main` and fully expands it to a version pin. The logic would be 'specify however you want' (just like Go) and then we'll figure out whether it's precise or needs resolution, and produce a normalized ref.

Or maybe I'm missing some lurking complexity :innocent:

tiborvass:
@vito Are users installing modules with commit without branch/tag often enough that we can't ignore it? Do we need a `probablyGitSha(ref)` logic ? Is there a minimum git sha length support ?

`dagger install github.com/foo/bar:deadbeef` vs `dagger install github.com/foo/bar@deadbeef`

tiborvass:
I'd say by assuming the git sha is minimum 10 digit hex (as shown in `git log --oneline`), we can differentiate between `github.com/foo/bar@<branch-or-tag>` and `github.com/foo/bar@<git-sha>` which would allow us to handle both `@` and `:` interchangeably. If both are present, it's always branch first, then commit. Does that sound good ?

vito:
@tiborvass

> Are users installing modules with commit without branch/tag often enough that we can't ignore it?

Yeah I don't think we can ignore it. I make Go point directly to a commit all the time, in the early lifetime of a module that I don't want to version yet. Sometimes like `go get github.com/vito/foo@$(git -C ~/src/foo rev-parse HEAD)` after pushing `foo`. I'd expect the same to work with Dagger.

The commit-without-branch case is interesting though because it can't reasonably be expanded to include a ref (like `main`) - since they exist on multiple branches/tags/etc. I think it should just normalize to `<url>@<commit>`.

---

I don't think we even need to have a static heuristic based on length or anything - [`GitRepository.ref` already accepts commit, branch, or tag](https://docs.dagger.io/api/reference/#GitRepository-ref) and will give you whatever it resolved to. We can just grab the `commit` off the result value to know whether it was specified as a commit or a ref. To handle short commit hashes we could do `expanded.HasPrefix(original)` instead of strict equality so we're not fooled into thinking it's a ref (like `github.com/foo/bar:deadbe@deadbeefdeadbeef1234567890` should just be `github.com/foo/bar@deadbeefdeadbeef1234567890`).

tiborvass:
@vito good point! So would you agree the logic is:

1. split what's after `url`, into two components with `:` and `@` as separators.
2. if there are two components, normalize to gitref by making the first component the branch/tag and the second component the commit.
3. if there is one component call GitRepository.ref on it, get the commit and check if common prefix in which case component is a git commit, otherwise it's a branch or tag.

I'd also add for consideration:
4. If both branch/tag and commit are specified, and upon git fetching the branch/tag, the commit is not found, FALLBACK to downloading the whole repo and then lookup the commit (current behavior) ?

Also:
5. Can we do `--depth 1` for even greater performance ? Or is that already the case ? To be fair, `--depth 1` is mostly useful for tags, but I believe not for moving branches. Without `--depth 1` it would still be better than fetching all branches, but it'd be so good to benefit from that performance boost.

vito:
@tiborvass yep, agree with 1-3

> 4. If both branch/tag and commit are specified, and upon git fetching the branch/tag, the commit is not found, FALLBACK to downloading the whole repo and then lookup the commit (current behavior) ?

As long as we fail loudly if the commit can't be found, :+1: 

> 5. Can we do `--depth 1` for even greater performance ? Or is that already the case ? To be fair, `--depth 1` is mostly useful for tags, but I believe not for moving branches. Without `--depth 1` it would still be better than fetching all branches, but it'd be so good to benefit from that performance boost.

My understanding is stuff like this and above are handled magically behind-the-scenes for us in Buildkit, where Git repos are aggressively deduped and fetched incrementally or some sort of magic like that.

shykes:
@vito didn't we reimplement core git operations in "pure dagger" and stop using buildkit's nativr implementations a few months back?

tiborvass:
After looking at the code, it seems like `depth 1` is default when checking out a git tree https://github.com/dagger/dagger/blob/b22f7708d4e0115dda35152ed6a0ba4faa760cd8/core/schema/git.go#L816 as well as in buildkit codepath (dockerfile and dockerui): https://github.com/dagger/dagger/blob/b22f7708d4e0115dda35152ed6a0ba4faa760cd8/internal/buildkit/source/git/source.go#L471

jedevc:
Yup, all the logic around cloning with `--depth` is all in https://github.com/dagger/dagger/blob/9f722e07b7f8dcd75e4c3370620bd96d421f8aae/core/git.go#L116-L124

We don't rely on any of the buildkit codepath anymore (except for dockerfile's that reference git sources).

tiborvass:
Soooo, it turns out, github supports fetching by commit now... and statistically it doesn't seem slower.

So I don't know how much that changes the desired logic. Or if it's only for git servers that don't support fetching by commit. cc @jedevc @vito

Issue #8402, opened 2024-09-10
Container.withError


shykes:
## Problem

When you have a beautiful chain of 20 function calls, no error checking end-to-end... except the one in the middle, then you're forced to break your chain, plumb error checking all the way up... So much beauty ruined.

## Solution

Allow me to attach an error to the container, so that it is lazily returned to the caller, and I don't have to deal with error passing in my code.  Example:

```golang
func Beautiful(ctx context.Context) *dagger.Container {
  return dag.Container().
    Foo().
    Bar().
    Baz().
    With(func(c *dagger.Container) *dagger.Container {
      contents, err := c.
        WithExec([]string{"check", "the", "thing"}).
        File("the-thing-output.txt").
        Contents(ctx)
      if err != nil {
        return c.WithError(err)
      }
      if contents != "expected-thing" {
        return c.WithErrorf(fmt.Errorf("The thing wasn't the expected shape! Failed to check the thing"))
      }
      return c
    }).
    Bla().
    Bla().
    ChefsKiss()
}
```

jedevc:
Agreed, I previously considered whether we could *instead* have `WithErr` (naming tbd), which takes a `func (*dagger.Container) (*dagger.Container, error)`. I'd rather we don't *explicitly* embed errors into containers, but I'm struggling to explain why I think that pattern is less desirable.

Regardless of the interface, the error would then get surfaced in the next function in the chain which returns an `error`, and if one is not present, it would be exposed at the top level of the module.

I like the idea :tada:

---

After an error is returned, do all the later steps get skipped - including future `With`s? I would assume so. Kinda neatly, we're just building monadic do-notation in go with this :laughing:

PR #11222, opened 2025-10-11
Allow to raise errors with withErrorMessage

eunomie:
Rel #8402 

This PR adds a `withError` to `dagger.Container` and `dagger.Directory` objects.
If the argument sent to `withError` is not empty, the error will be raised.

This is particularly useful to raise errors in a self chainable way, like when using `with`.

```go
ctr = ctr.
	With(func(r *dagger.Container) *dagger.Container {
		ctr, ok = doSomething(r)
		if !ok {
			return r.WithError("not ok")
		}
		return ctr
	})
```

### DIFF
--- a/.changes/unreleased/Added-20251015-075538.yaml
+++ b/.changes/unreleased/Added-20251015-075538.yaml
@@ -1,0 +1,9 @@
+kind: Added
+body: |-
+    Allow to raise errors using withError.
+
+    This is for instance really useful from inside self-chainable blocks (using with) where an error can't be returned.
+time: 2025-10-15T07:55:38.141332-07:00
+custom:
+    Author: eunomie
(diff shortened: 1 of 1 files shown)


eunomie:
> TLDR: I think we should call it `WithError` and `Error`, instead of `WithErrorMessage` and `ErrorMessage`

OK, I can rename it and remove the go speficic implementation. That was my first version.
But what is `ErrorMessage` you mentioned that should be renamed to `Error`? Is that a field with the error you had in mind? Because for now the error is not stored anywhere, it's just raised immediately.

shykes:
Ah my bad. I thought it was lazy with a field. Then ignore my comment about a field :)

shykes (approved):

PR #11223, opened 2025-10-12
track and export to host directory origin

### LINKED ISSUE #11160: CLI exports Changeset to the wrong directory

vito:
fixes #11160

* [ ] Bikeshed API
  * Is "origin" too vague?
* [ ] Does this affect LLM's interpretation of a `Changeset` return value?

### DIFF
--- a/cmd/dagger/functions.go
+++ b/cmd/dagger/functions.go
@@ -718,9 +718,15 @@ func handleChangesetResponse(ctx context.Context, dag *dagger.Client, response a
 
 	var summary strings.Builder
 	var noChanges bool
+	var exportDest string = "."
 	if err := (func() (rerr error) {
 		ctx, span := Tracer().Start(ctx, "analyzing changes")
 		defer telemetry.End(span, func() error { return rerr })
+
+		// Try to get the origin path from the Before directory
+		if origin, err := changeset.Before().Origin(ctx); err == nil && origin != "" {
+			exportDest = origin
+		}
 
 		preview, err := idtui.PreviewPatch(ctx, changeset)
 		if err != nil {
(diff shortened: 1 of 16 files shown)


shykes:
Keeping this one on life support :)

PR #11229, opened 2025-10-13
Export introspection schema JSON for a module

eunomie:
Allow to export the introspection schema JSON for a module. This can be useful to debug engine/sdk features.

If you have a module initialized at `my/module` you can use it like:

```
host | directory 'my/module' | as-module | schema-json
```

For instance to get it on your host:

```
host | directory 'my/module' | as-module | schema-json | export schema.json
```

### DIFF
--- a/core/schema/module.go
+++ b/core/schema/module.go
@@ -643,7 +643,7 @@ func (s *moduleSchema) schemaJSON(
 	mod *core.Module,
 	args struct{},
 ) (dagql.Result[*core.File], error) {
-	return mod.Deps.SchemaIntrospectionJSONFile(ctx, []string{"Host"})
+	return mod.Deps.SchemaIntrospectionJSONFileForModule(ctx)
 }
 
 func (s *moduleSchema) moduleDependencies(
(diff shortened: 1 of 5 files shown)


eunomie:
cc @grouville

core/schema/module.go
```diff
@@ -635,6 +637,14 @@ func (s *moduleSchema) moduleUserDefaults(ctx context.Context, mod *core.Module,
 	return mod.UserDefaults(ctx)
 }
 
+func (s *moduleSchema) schemaJSON(
+	ctx context.Context,
+	mod *core.Module,
+	args struct{},
+) (dagql.Result[*core.File], error) {
+	return mod.Deps.SchemaIntrospectionJSONFile(ctx, []string{"Host"})
```
jedevc: We should move this magic string to a separate function, it feels a little fragile.

eunomie: The array of hidden types for modules has been extracted.
A new function specific for modules automatically hide those types.

PR #11234, opened 2025-10-14
chore: refactor file.Open to use execInMount

alexcb:
Rather than calling ref.ReadFile, which causes buildkit to do a mount/unmount each time Read() is called, instead mount the file and keep it open until the file is closed.
    
This change required refactoring execInMount into a new execInMountCloser which allows the mount to persist over multiple calls to Read().

### DIFF
--- a/core/file.go
+++ b/core/file.go
@@ -532,13 +532,19 @@ func (file *File) Open(ctx context.Context) (io.ReadCloser, error) {
 
 	filePath, err := containerdfs.RootPath(root, file.File)
 	if err != nil {
-		_, _ = closer(true)
+		_, closeErr := closer(true)
+		if closeErr != nil {
+			err = errors.Join(err, closeErr)
+		}
 		return nil, err
 	}
 
 	r, err := os.Open(filePath)
 	if err != nil {
-		_, _ = closer(true)
+		_, closeErr := closer(true)
+		if closeErr != nil {
+			err = errors.Join(err, closeErr)
+		}
 		return nil, err
 	}
 
--- a/core/util.go
+++ b/core/util.go
@@ -3,6 +3,7 @@ package core
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/fs"
 	"maps"
@@ -335,7 +336,10 @@ func MountRef(ctx context.Context, ref bkcache.Ref, g bksession.Group, f func(st
 	}
 	err = f(dir)
 	if err != nil {
-		_ = closer()
+		closeErr := closer()
+		if closeErr != nil {
+			err = errors.Join(err, closeErr)
+		}
 		return err
 	}
 	return closer()
@@ -494,7 +498,10 @@ func execInMount[T fileOrDirectory](ctx context.Context, obj T, f func(string) e
 	}
 	err = f(root)
 	if err != nil {
-		_, _ = closer(true)
+		_, closeErr := closer(true)
+		if closeErr != nil {
+			err = errors.Join(err, closeErr)
+		}
 		return nil, err
 	}
 	return closer(false)


alexcb:
This work relates to https://github.com/dagger/dagger/pull/10775#pullrequestreview-3180063804 ; however due to the container schema's `import` [function](https://github.com/alexcb/dagger/blob/09f6d9751deb2e6fa14db12902cac1b0df039ca2/core/schema/container.go#L2395-L2401), I had to maintain a reader interface which prevented me from deleting the `Open` method entirely.

core/util.go
```diff
@@ -319,53 +319,55 @@ func ptr[T any](v T) *T {
...
 }
 
-// ReadonlyMountRef is a utility for easily mounting a ref read-only.
+// MountRefCloser is a utility for mounting a ref.
 //
-// To simplify external logic, when the ref is nil, i.e. scratch, the callback
-// just receives a tmpdir that gets deleted when the function completes.
-func ReadonlyMountRef(ctx context.Context, ref bkcache.Ref, g bksession.Group, f func(string) error) error {
+// To simplify external logic, when the ref is nil, i.e. scratch, a tmpdir is created (and deleted when the closer func is called).
+
```
jedevc: mega nit:
```suggestion
//
```

so this shows up as one comment

core/util.go
```diff
@@ -319,53 +319,55 @@ func ptr[T any](v T) *T {
...
-		return err
+		return "", func() error { return nil }, err
 	}
 	lm := snapshot.LocalMounter(mount)
-	defer lm.Unmount()
-
 	dir, err := lm.Mount()
 	if err != nil {
-		return err
+		return "", func() error { return nil }, err
```
jedevc:
```suggestion
		return "", nil, err
```

core/util.go
```diff
@@ -481,61 +483,96 @@ type fileOrDirectory interface {
...
+		return nil, err
+	}
+	return closer(false)
+}
+
+// execInMountCloser evaluates an object and mounts the root fs and returns the mounted path and a closer, which will unmount
+// the file or directory object's root filesystem, and potentially return a modified object, if both the withSavedSnapshot option is specified and the abort flag was not set.
+// The abort flag is only used when the withSavedSnapshot option is specified.
+// NOTE: prefer execInMount where possible, unless finer-grained control of the filesystem mount is required.
+func execInMountCloser[T fileOrDirectory](ctx context.Context, obj T, optFns ...execInMountOptFn) (string, func(abort bool) (T, error), error) {
```
jedevc: Could readonly be a `execInMountOptFn`? To avoid it needing to propagate everywhere.

core/file.go
```diff
@@ -509,22 +509,54 @@ func (file *File) WithTimestamps(ctx context.Context, unix int) (*File, error) {
...
-	bk, err := query.Buildkit(ctx)
+
+	filePath, err := containerdfs.RootPath(root, file.File)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get buildkit client: %w", err)
+		_, _ = closer(true)
+		return nil, err
 	}
 
-	fs, err := reffs.OpenDef(ctx, bk, file.LLB)
```
jedevc, line 522: Oof, this is such a cursed function lol.

Very tempted to suggest to keep pulling on this thread, and rip it out entirely - it also seems to be used in `resolveUIDGID` which seems like it could use the mount ref helpers.

core/file.go
```diff
@@ -509,22 +509,54 @@ func (file *File) WithTimestamps(ctx context.Context, unix int) (*File, error) {
...
+
+func (frc fileReadCloser) Close() error {
+	return frc.close()
+}
+
+var _ io.ReadCloser = fileReadCloser{}
+
 func (file *File) Open(ctx context.Context) (io.ReadCloser, error) {
-	query, err := CurrentQuery(ctx)
+	root, closer, err := execInMountCloser(ctx, file, allowNilBuildkitSession)
```
jedevc: nit: don't call this `execInMount` since it doesn't really exec anything itself. Maybe just `mountRef`? Or similar?

core/util.go
alexcb: good point -- I can also simplify other bits:

```
-               return "", noOpCloser, err
+               return "", nil, err
```

core/file.go
alexcb: good point. I renamed it to `mountObj`, since `MountRef` already exists (and is actually used by `mountObj`)

core/file.go
alexcb: noted! I'll add it to my TODO list, and tackle it in a separate PR.

core/util.go
alexcb: I removed the `ReadonlyMountRef` function, and added a `mountRefAsReadOnly` option.

The `mountObj` function (fka `execInMountCloser`), is different enough from `MountRef` (e.g. it also performs the snapshot call), so I decided to keep the opts structs separate. I did however plumb through the case where `!opt.commitSnapshot` will result in the readonly flag being set.

core/util.go
```diff
@@ -314,58 +314,63 @@ func ptr[T any](v T) *T {
...
 	if err != nil {
 		return err
 	}
-	lm := snapshot.LocalMounter(mount)
-	defer lm.Unmount()
-
-	dir, err := lm.Mount()
+	err = f(dir)
 	if err != nil {
+		_ = closer()
```
sipsma: nit: `errors.Join` instead? just in case it actually happens somehow

alexcb: good point -- I actually did that in one place, and oddly didn't think to do that everywhere :see_no_evil:

PR #11237, opened 2025-10-15
feat: add `Changeset.isEmpty`

jedevc:
Adds a shorthand `Changeset.empty` which can now be used to check if the Changeset has any patches to apply.

TODO:
- [x] Tests
- [x] Implementation should check the patch (the paths have false positives because of overlay semantics)


jedevc:
Based on discussion with @vito, we switched `empty` to `isEmpty`, which seems to be more of a graphql "standard" in the ecosystem.

PR #11238, opened 2025-10-15
✨ Toolchains

kpenfound:
This is a new feature call Toolchains.

A toolchain is a way of calling a module meant to be used as a standalone tool that has the same source context as your main module. For example, a linter module that should work out of the box with your project can now be used without integrating it into your SDK code.

Toolchains can coexist with your main module being a Blueprint, SDK, or neither.

Toolchains will likely replace the need to `dagger -m <dependency>`, since the usage was similar except more limited.

For example, if I have the following modules:
- Name: Go, Functions: [Build, Lint]
- Name: Npm, Functions: [Build, Lint]

and I install them as toolcahins:
- `dagger toolchain install ./go`
- `dagger toolchain install github.com/foo/npm`

I can:
- `dagger call go build`
- `dagger call npm lint`
- `dagger call <toolchain> <function> ...`
- ...

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -746,6 +746,9 @@ This command is idempotent: you can run it at any time, any number of times. It 
 var toolchainCmd = &cobra.Command{
 	Use:   "toolchain",
 	Short: "Manage toolchains",
+	Annotations: map[string]string{
+		"experimental": "true",
+	},
 }
 
 var toolchainInstallCmd = &cobra.Command{


core/schema/modulesource.go
```diff
@@ -2637,8 +2731,9 @@ func (s *moduleSourceSchema) loadDependencyModules(ctx context.Context, src *cor
 	}
 
 	var eg errgroup.Group
-	depMods := make([]dagql.Result[*core.Module], len(src.Dependencies))
-	for i, depSrc := range src.Dependencies {
+	depMods := make([]dagql.Result[*core.Module], len(src.Self().Dependencies))
+	for i, depSrc := range src.Self().Dependencies {
+		i := i // capture loop variable
```
grouville: I _think_ go fixed it

nipuna-perera:
I didn't realize this wasn't already the case until I ran into a use case where I need to support both `install` and `install --blueprint`. One comment I have on consistency

> If I install both Go and Npm, I can:
\
`dagger call go build`
`dagger call npm lint`

This isn't consistent with how it works when I do a `dagger install` (non blueprint). If I do a `dagger install` of these, it would be accessed like

```
dagger call -m go build
dagger call -m npm lint
```

I think the blueprint install should behave the same. Or the regular install should change.

kpenfound:
You're right @nipuna-perera , I'm changing this up to be a new feature called toolchains. Toolchains are independent from blueprints, SDKs, and dependencies, and are namespaced by their name.

So your main module can be either:
- a blueprint
- a module built with an SDK (+ dependencies)

And then on top of that, you can add toolchains, accessed directly by their name (no -m). Ideally this could replace most or all use cases of `dagger call -m <dependency` because a) it has your main module's context, and b) it can be called on a remote module

shykes:
Would be cool to have `dagger toolchain list`, that just lists installed toolchains by name (and maybe description if available?)

kpenfound:
@shykes done

```
> dagger functions

Name        Description
build       Build the application container
build-env   Build a ready-to-use development environment
hello       toolchain 'hello': A Dagger module to say hello to the world
publish     Publish the application container after building and testing it on-the-fly
test        Return the result of running unit tests
yarn        toolchain 'yarn': A module for Example functions

> dagger toolchain list

Name    Description
hello   A Dagger module to say hello to the world
yarn    A module for Example functions

> cat dagger.json
{
  "name": "hello-dagger-template",
  "engineVersion": "v0.19.3-010101000000-dev-dea201bb73a8",
  "blueprint": {
    "name": "hello-dagger",
    "source": "https://github.com/kpenfound/blueprints/hello-dagger@main",
    "pin": "2bb1d301e87cbe218b65c99f4e74618cda252638"
  },
  "toolchains": [
    {
      "name": "hello",
      "source": "github.com/shykes/hello@main",
      "pin": "2d789671a44c4d559be506a9bc4b71b0ba6e23c9"
    },
    {
      "name": "yarn",
      "source": "github.com/kpenfound/blueprints/yarn@main",
      "pin": "2bb1d301e87cbe218b65c99f4e74618cda252638"
    }
  ]
}
```

kpenfound:
@jedevc @sipsma @vito looking for feedback on the implementation and if its maintainable and not too hacky that it'll cause headaches down the road. Specifically the proxying in core/object.go

core/schema/modulesource.go
```diff
@@ -1491,9 +1523,306 @@ func (s *moduleSourceSchema) moduleSourceWithBlueprint(
...
+	// Build a new module source with updated toolchains
+	tmpSrc := parentSrc.Self().Clone()
+	tmpSrc.Toolchains = nil
+	tmpSrc.ConfigToolchains = nil
+
+	for _, toolchainID := range newUpdatedToolchainArgs {
+		var toolchainSrc dagql.ObjectResult[*core.ModuleSource]
+		err := dag.Select(ctx, dag.Root(), &toolchainSrc,
+			dagql.Selector{
+				Field: "loadID",
```
vito: Don't think that's a thing 🤔 (though I wish it was)

```suggestion
				Field: "loadModuleSourceFromID",
```

Now that I look deeper though, is there a reason we need to re-load all these? Seems like we have a `dagql.ObjectResult[*core.ModuleSource]` in each case already before adding IDs to an array, only to re-load them - couldn't we just store the objects directly?

edit: ah, in `moduleSourceWithUpdateDependencies` this is done because the sources ultimately get passed to a `withDependencies` dagql.Select where they're all passed as args, whereas this one updates in-place. If it helps with deduping it might be worth adding a `withToolchains`, or even replacing `withToolchain`

test_output.txt
vito, line 1: rm?

core/schema/modulesource.go
```diff
@@ -1491,9 +1523,306 @@ func (s *moduleSourceSchema) moduleSourceWithBlueprint(
...
+
+	// Add to toolchains array
+	parentSrc.ConfigToolchains = append(parentSrc.ConfigToolchains, tmpConfig.Dependencies[0])
+	parentSrc.Toolchains = append(parentSrc.Toolchains, tmpSrc.Dependencies[0])
+
+	parentSrc.Digest = parentSrc.CalcDigest(ctx).String()
+	return parentSrc, nil
+}
+
+func (s *moduleSourceSchema) moduleSourceWithUpdateToolchains(
```
vito: Feels like this could be deduped somehow with `moduleSourceWithUpdateDependencies`?

core/schema/modulesource.go
```diff
@@ -1491,9 +1523,306 @@ func (s *moduleSourceSchema) moduleSourceWithBlueprint(
...
+
+	resultSrc := parentSrc.Self().Clone()
+	resultSrc.ConfigToolchains = tmpConfig.Toolchains
+	resultSrc.Toolchains = tmpSrc.Toolchains
+	resultSrc.Digest = resultSrc.CalcDigest(ctx).String()
+
+	return dagql.NewResultForCurrentID(ctx, resultSrc)
+}
+
+func (s *moduleSourceSchema) moduleSourceWithoutToolchains(
```
vito: Feels like this could be deduped with `moduleSourceWithoutDependencies` too

core/gqlformat.go
```diff
@@ -26,6 +26,11 @@ func gqlObjectName(name string) string {
 	return strcase.ToCamel(name)
 }
 
+// GqlObjectName is the exported version of gqlObjectName
+func GqlObjectName(name string) string {
```
vito: Still needed? (Not seeing it w/ <kbd>Ctrl+F</kbd>)

core/schema/modulesource.go
```diff
@@ -2503,6 +2835,265 @@ func (s *moduleSourceSchema) moduleSourceIntrospectionSchemaJSON(
...
+	hasSDK := len(mod.ObjectDefs) > 0
+
+	var err error
+	if hasSDK {
+		// Merge toolchain fields into SDK's main object
+		err = mergeToolchainsWithSDK(mod, toolchainMods)
+	} else {
+		// No SDK - create shadow module to hold toolchain fields
+		mod, err = createShadowModuleForToolchains(ctx, mod, toolchainMods, dag)
+	}
```
vito, line 2998: 🧑‍🍳🤌💋

core/schema/modulesource.go
```diff
@@ -2525,105 +3117,61 @@ func (s *moduleSourceSchema) moduleSourceAsModule(
...
-	if err != nil {
-		return inst, fmt.Errorf("failed to load dependencies as modules: %w", err)
+	// In blueprint mode, use the blueprint as the main source
+	// This must happen before creating the module so the SDK loads from blueprint source
+	if isBlueprintMode {
+		src = src.Self().Blueprint
+		tcCtx.src = src
+		// Copy toolchains from original source to blueprint source
+		// A blueprint having toolchains is currently not supported
+		tcCtx.src.Self().Toolchains = tcCtx.originalSrc.Self().Toolchains
```
vito: We should do this without mutating the object directly, since that really voids the warranty with DagQL; every object is supposed to be immutable, driven by the ID that queried it. Mutating in this way risks spooky action at a distance.

Instead you could do something like this, if we add the `withToolchains` API mentioned earlier:

```go
var sourceIDs []core.ModuleSourceID
for _, src := range tcCtx.originalSrc.Self().Toolchains {
  sourceIDs = append(sourceIDs, dagql.NewID[*core.ModuleSource](src.ID()))
}
err := dag.Select(ctx, tcCtx.src, &tcCtx.src, dagql.Selector{
  Field: "withToolchains",
  Args: []dagql.NamedInput{
    {
      Name: "toolchains",
      Value: dagql.ArrayInput[core.ModuleSourceID](sourceIDs),
    },
  },
})
```

And hopefully everything continues to work because the mutation isn't load-bearing. 🙃

core/schema/modulesource.go
```diff
@@ -2657,13 +3222,19 @@ func (s *moduleSourceSchema) loadDependencyModules(ctx context.Context, src *cor
 	for _, depMod := range depMods {
 		deps = deps.Append(depMod.Self())
 	}
+	for _, tcMod := range tcMods {
+		clone := tcMod.Self().Clone()
+		clone.IsToolchain = true
+		clone.ContextSource = dagql.NonNull(src)
```
vito, line 3167: This particular mutation is fine since we call `Clone()` first and only care about the underlying object 👍

core/schema/modulesource.go
kpenfound: Fixed, that must have been something missed in a refactor. I dont know what "loadID" was supposed to be lol

kpenfound:
Thanks for the review @vito , i've made the updated changes. Need to do another self review pass now

cmd/dagger/module.go
```diff
@@ -725,6 +743,235 @@ This command is idempotent: you can run it at any time, any number of times. It
...
+var toolchainCmd = &cobra.Command{
+	Use:   "toolchain",
+	Short: "Manage toolchains",
+}
+
+var toolchainInstallCmd = &cobra.Command{
+	Use:     "install [options] <module>",
+	Short:   "Install a toolchain to the current module",
+	Long:    "Install another module as a toolchain to the current module.",
+	Example: "dagger toolchain install github.com/example/toolchain",
```
sipsma: extreme nit: I guess `example` is a real org? https://github.com/example/

wonder if we should avoid putting it in example commands

cmd/dagger/module.go
```diff
@@ -725,6 +743,235 @@ This command is idempotent: you can run it at any time, any number of times. It
...
+			if !alreadyExists {
+				return fmt.Errorf("module must be fully initialized")
+			}
+
+			contextDirPath, err := modSrc.LocalContextDirectoryPath(ctx)
+			if err != nil {
+				return fmt.Errorf("failed to get local context directory path: %w", err)
+			}
+
+			modSrc = modSrc.WithoutBlueprint()
```
sipsma: is this supposed to be `WithoutToolchains(extraArgs)`?

core/schema/modulesource.go
```diff
@@ -717,16 +736,28 @@ func (s *moduleSourceSchema) loadBlueprintModule(
...
+	// Load toolchains array
+	if len(src.ConfigToolchains) > 0 {
+		src.Toolchains = make([]dagql.ObjectResult[*core.ModuleSource], len(src.ConfigToolchains))
+		for i, pcfg := range src.ConfigToolchains {
+			toolchain, err := core.ResolveDepToSource(ctx, bk, dag, src, pcfg.Source, pcfg.Pin, pcfg.Name)
+			if err != nil {
+				return fmt.Errorf("failed to resolve toolchain to source: %w", err)
+			}
+			src.Toolchains[i] = toolchain
+		}
```
sipsma, line 758: nit: if straightforward this seems like something worth parallelizing (probably with `util/parallel`, though I think you'll want [the same fix](https://github.com/dagger/dagger/pull/11211/commits/068e00b0338974091b591dfa5b367f72e4b62fea) that's in the checks PR)

core/schema/modulesource.go
```diff
@@ -1392,323 +1424,221 @@ func (s *moduleSourceSchema) generatedCodeWithVCSIgnoredPaths(ctx context.Contex
...
-func (s *moduleSourceSchema) moduleSourceWithDependencies(
-	ctx context.Context,
-	parentSrc *core.ModuleSource,
-	args struct {
-		Dependencies []core.ModuleSourceID
-	},
-) (*core.ModuleSource, error) {
-	parentSrc = parentSrc.Clone()
+// itemType distinguishes between dependencies and toolchains in error messages and field access
+type itemType int
```
sipsma: nit: `item` is very nondescript, maybe `moduleRelationType` (and `relatedModule` when `item` is used as an arg name in other functions below)?

not a blocker, naming is hard

core/schema/modulesource.go
```diff
@@ -1392,323 +1424,221 @@ func (s *moduleSourceSchema) generatedCodeWithVCSIgnoredPaths(ctx context.Contex
...
+		return "unknowns"
+	}
+}
+
+// itemAccessor provides unified access to dependencies or toolchains in a ModuleSource
+type itemAccessor struct {
+	typ itemType
+}
+
+func (a itemAccessor) getItems(src *core.ModuleSource) []dagql.ObjectResult[*core.ModuleSource] {
```
sipsma: nit: feel like it would be much more straightforward to just have `getItems` and `setItems` be methods on `ModuleSource` like `func (src *ModuleSource) getItems(typ itemType)`

core/schema/modulesource.go
```diff
@@ -1392,323 +1424,221 @@ func (s *moduleSourceSchema) generatedCodeWithVCSIgnoredPaths(ctx context.Contex
...
+) ([]core.ModuleSourceID, error) {
 	dag, err := core.CurrentDagqlServer(ctx)
 	if err != nil {
-		return inst, fmt.Errorf("failed to get dag server: %w", err)
+		return nil, fmt.Errorf("failed to get dag server: %w", err)
 	}
 
 	type updateReq struct {
-		symbolic string // either 1) a name of a dep or 2) the source minus any @version
-		version  string // the version to update to, if any specified
```
sipsma, line 1669: can we leave these comments?

core/schema/modulesource.go
```diff
@@ -1392,323 +1424,221 @@ func (s *moduleSourceSchema) generatedCodeWithVCSIgnoredPaths(ctx context.Contex
...
-	updateReqs := make(map[updateReq]struct{}, len(args.Dependencies))
-	for _, updateArg := range args.Dependencies {
+	updateReqs := make(map[updateReq]struct{}, len(updateArgs))
+	for _, updateArg := range updateArgs {
 		req := updateReq{}
 		req.symbolic, req.version, _ = strings.Cut(updateArg, "@")
 		updateReqs[req] = struct{}{}
 	}
 
-	// loop over the existing deps, checking each one for whether they should be updated based on the args
```
sipsma, line 1678: this comment seemed useful too since this code is a bit dense

core/modulesource.go
sipsma, line 1: I think we need to incorporate Toolchains (and blueprint?) into `ModuleSource.CalcDigest()`?

The digest we use there determines the cache key for various SDK calls on the module, so if there's a change to the toolchain/blueprint list, we should have a new digest.

sipsma (commented):
still reviewing, just dropping what I've seen so far

core/schema/modulesource.go
```diff
@@ -2691,6 +3172,23 @@ func (s *moduleSourceSchema) loadDependencyModules(ctx context.Context, src *cor
 		return nil, fmt.Errorf("failed to load module dependencies: %w", err)
 	}
 
+	// Load all toolchains as dependencies
+	tcMods := make([]dagql.Result[*core.Module], len(src.Self().Toolchains))
+	if len(src.Self().Toolchains) > 0 {
+		var tceg errgroup.Group
```
sipsma: seems like we may as well use the same `eg` we use for deps and load toolchains in parallel with normal deps too?

cmd/dagger/module.go
kpenfound: Ugh yes, thanks. Originally this was all blueprints and detangling them has been tough

core/schema/modulesource.go
kpenfound: Got it, after this is in we're going to work on rebasing the checks PR on this so we could resolve this then

core/modulesource.go
kpenfound: Yeah that makes sense! Merging in the function caching stuff I basically just made sure the tests still passed 😅

kpenfound:
@sipsma 

> The approach generally makes sense to me, but I'm confused how/whether the case of a toolchain that has multiple objects defined would work? Do we prevent that somewhere?

If I understand your question, this is prevented here, where we only proxy to the main object:

https://github.com/dagger/dagger/pull/11238/files#diff-f811ad106ed3533deeb7e92d73931fd62118c28a230b57d5cf6d30e1bd278113R477-R481

One thing I haven't tested (will do this now) is how the proxying looks in shell, since shell basically already does this for dependencies.

sipsma:
> If I understand your question, this is prevented here, where we only proxy to the main object:

I guess what I'm wondering is if my module has a toolchain like this:
```go
type Tool struct {}

func (Tool) Foo() *OtherObj {
  // ...
}

type OtherObj struct{}

func (OtherObj) Bar() string {
   // ....
}
```

Then what happens if I call `dagger -c 'foo | bar'` on my module? Maybe it works already some way or another, I just couldn't piece together in my head how that would work in this current implementation.

kpenfound:
@sipsma added a (passing) test that covers that :) Basically once you have `foo`, you're in `foo`'s server, so its all the same from there

sipsma:
> @sipsma added a (passing) test that covers that :) Basically once you have `foo`, you're in `foo`'s server, so its all the same from there

That's the thing, I couldn't find anywhere that we proxy to a totally separate dagql server; it seemed like we'd only be hitting one server still in a query like that.

Looked again, think I figured it out (medium confidence, correct me if I'm missing something).

We indeed are **not** actually hitting a separate server for that `foo | bar` type of query. It's all one dagql server:
1. `func (s *moduleSourceSchema) loadDependencyModules` returns both normal deps and toolchains
2. We then set the return value of that as `mod.Deps` in `ModuleSource.asModule`
3. When the CLI calls `Module.serve` (which is what installs the module into the current server), it's setting `IncludeDependencies: true`, which means that both the module and its deps are installed ("deps" in this case being both normal deps and toolchains)

So what's actually happening is the toolchain modules plus the module itself all get installed into the same dagql server, which is then served to the caller.

This probably works I suppose (thanks for adding the test, those pay off in the long run!), but doesn't seem to match the original understanding, so just FYI in case this raises any issues.

The only thing I can imagine that *might* not work is if you created a module A that has toolchains, but then created module B with a normal dependency on module A.
* In that case, `IncludeDependencies: true` isn't gonna be set (and can't be) when creating the dagql server presented to module B, so if you tried to do `dag.Foo().Bar()` (using example from before) from your module B code, then it might actually break.

I have no clue if that use case matters at all, so not a blocker if we don't care. If we do care it's probably a simple-ish fix too.

kpenfound:
@sipsma thanks for digging deeper! What confuses me about that is how `IncludeDependencies: true` is being set? As for the last question, that would be fine because toolchains are not meant to be accessible as dependencies in the client. You'd have to install the module as a dependency instead.

sipsma:
> What confuses me about that is how `IncludeDependencies: true` is being set?

That arg is just hardcoded true in the CLI: https://github.com/sipsma/dagger/blob/07db9ab39ee5ebe4722f77b56c3d9f61a33c5288/cmd/dagger/module_inspect.go#L74-L74

(tbh I had either never noticed or totally forgotten this too, you learn something new every day 😅)

> As for the last question, that would be fine because toolchains are not meant to be accessible as dependencies in the client. You'd have to install the module as a dependency instead.

Cool, SGTM then! I'll do another review pass today

cmd/dagger/module.go
sipsma, line 1: Do we want to mark these commands as experimental/hidden at first? Or are we good to release right away?

kpenfound: I think its good to go. We have a few follow ups like configuration https://github.com/dagger/dagger/issues/11316 and integration with checks https://github.com/dagger/dagger/pull/11211 before we formally launch it but IMO there's no reason to hide it. Wdyt @shykes ?

Issue #11107, opened 2025-09-22
🐞 `Directory.Diff` fails when input directories point to the same folder


matipan:
### What is the issue?

Running `Diff` on two `Directory` types that point to the same underlying folder throws an error indicating that the directory was not found. Changing the contents of the underlying directory between calls fixes this issue.

### Dagger version

dagger v0.18.18 (image://registry.dagger.io/engine:v0.18.18) linux/amd64

### Steps to reproduce

Calling `dagger call foo entries` throws this error:
```go
func (m *Testing) Foo(ctx context.Context) *dagger.Directory {
	ctr := dag.Container().From("alpine").WithExec([]string{"sh", "-c", "echo 'hello' > /tmp/hello"})

	old := ctr.Directory("/tmp")
	n := ctr.Directory("/tmp")
	return n.Diff(old)
}```

If we modify the underlying `/tmp` directory in between calls the error goes away:
```go
func (m *Testing) Foo(ctx context.Context) *dagger.Directory {
	ctr := dag.Container().From("alpine").WithExec([]string{"sh", "-c", "echo 'hello' > /tmp/hello"})

	old := ctr.Directory("/tmp")
	n := ctr.WithExec([]string{"sh", "-c", "echo 'goodbye' > /tmp/goodbye"}).Directory("/tmp")
	return n.Diff(old)
}
```

### Log output

```sh
➜ dagger call foo entries
▶ connect 0.4s
▶ load module: . 2.2s
● parsing command line arguments 0.0s

● testing: Testing! 0.0s
▶ .foo: Directory! 2.1s CACHED
✘ .entries: [String!]! 0.0s ERROR
! /tmp: no such file or directory
```

Example trace: https://dagger.cloud/dagger/traces/7a75c44b2d045dad92c6401f8b621b54

jedevc:
Hah. Technically I guess this is expected lol - since technically there *is* not diff between those, and the subdir `tmp` in the snapshot doesn't exist.

I would suggest fixing this by making sure that both inputs to `Diff` are both rooted at `/` before doing the snapshot - this also fixes the weird issue where it doesn't let you diff if the inputs don't have the same directory paths.

Then we always get an output of `/` - and this will go away as a nice little side-effect.

PR #11241, opened 2025-10-16
feat: add git repo dirty detection

jedevc:
Adds a new `GitRepository.uncommited: Changeset` API that returns a changeset representing all the uncommitted changes in the repo.

This will only yield a non-empty changeset for repositories created using `Directory.asGit` which contains a worktree.

### DIFF
--- a/.changes/unreleased/Added-20251022-145836.yaml
+++ b/.changes/unreleased/Added-20251022-145836.yaml
@@ -1,0 +1,7 @@
+kind: Added
+body: |
+    New `GitRepository.uncommited` API to get changeset of dirty changes in the working directory
+time: 2025-10-22T14:58:36.682074344+01:00
+custom:
+    Author: jedevc
+    PR: "11241"


jedevc:
Ended up running into https://github.com/dagger/dagger/issues/11107 for the empty diff :facepalm: 

So through a basic fix for that in as well.

Edit: split out to https://github.com/dagger/dagger/pull/11268

shykes:
@jedevc is this intended to combine staged and unstaged changes? How does it interact with worktrees?

By default, as a developer I would assume:

- Combines staged and unstaged
- Always uses the primary worktree
- In the future we could make these configurable in a non-breaking way, with optional args.

Is my assumption correct?

Assuming yes 👆 , I propose one of:

- `GitRepository.UncommittedChanges`
- `GitRepository.Uncommitted`

I think `dirty` makes more sense as a boolean ("are there uncommitted changes?"). But if we're getting the actual changes, the above are more descriptive IMO.

jedevc:
> By default, as a developer I would assume:
> * Combines staged and unstaged
> * Always uses the primary worktree
> * In the future we could make these configurable in a non-breaking way, with optional args.
> 
> Is my assumption correct?

Yup to all of them.

One thing to note about combining staged + unstaged changes: if you apply the changeset you get from this new API, then it *doesn't* restage those previously staged changes. There's no way to do that, without modifying the `.git/index` file (which is probably not a good idea, since it's got lots of other stuff in it).

> - `GitRepository.Uncommitted`

I like this one.

core/schema/git.go
```diff
@@ -113,6 +113,11 @@ func (s *gitSchema) Install(srv *dagql.Server) {
 				dagql.Arg("patterns").Doc(`Glob patterns (e.g., "refs/tags/v*").`),
 			),
 
+		dagql.NodeFunc("__cleaned", DagOpDirectoryWrapper(srv, s.cleaned, WithPathFn(keepParentGitDir[cleanedArgs]))).
```
vito, line 116: (Squinting) maybe this would be useful to expose? Non-blocking, one-way doors and all that. Might as well wait til someone asks.

I could imagine using this to optimize uploads, by getting the clean base ref of their modifications, uploading only the modifications, and basing them on a warm Git cache. (@tiborvass mentioned this recently)

In my mind this would just look for whatever `HEAD` is and return a fresh Git ref pointing to it, but looking at the implementation it looks like it does what it says on the tin (cleans up the repo). I guess the difference is this approach will work even for un-pushed commits, which is an important distinction for that optimization, too.

core/schema/git.go
```diff
@@ -690,6 +695,76 @@ func (s *gitSchema) branches(ctx context.Context, parent *core.GitRepository, ar
...
+			dagql.Selector{
+				Field: "tree",
+			},
+		); err != nil {
+			return inst, fmt.Errorf("failed to select head tree for clean repo: %w", err)
+		}
+		cleaned = dirty
+	} else {
+		// wrapped in an internal field to get good caching behavior
+		if err := dag.Select(ctx, parent, &cleaned, dagql.Selector{
```
vito, line 745: Ah ha, got it, we use it to compute the changes

core/schema/git.go
jedevc: So there's a reason that this isn't just `repo.Head().Tree()`. If I have a `vendor` directory that's gitignored, then it shouldn't actually be part of the uncommitted state. `__cleaned` includes this directory, but `repo.Head().Tree()` wouldn't (we run `git clean -fd`, not `git clean -fdx`).

(Note, it's also just cheaper to compute! We only have to touch a few files to reset the state, compared to creating a whole checkout.)

This means that the diff is harder to read, it's less relevant, and if you apply it on top of a git repo, then it'll try removing files that don't exist (which is fine, it's just weird).

Issue #11242, opened 2025-10-16
🖐️ Apple containers


nadirollo:
### What happened? What did you expect to happen?

What am I doing wrong? 😅 

For "reasons", some of our engineers have started looking at apple containers to replace their colima/podman (90% of our team uses apple laptops, so naturally they will eventually look for the "native" container runtime).

I just tried to run dagger with apple containers under the hood, and I'm seeing some issues (which were reported from a couple of our engineers, hence I'm looking at.. ).  Maybe is my setup or any missing configs, so any pointers on where to go would be grand.

With a fresh install of apple containers (v0.5.0) and removed all traces of docker-desktop:
```
$> container --version
container CLI version 0.5.0 (build: release, commit: 48230f3)
$> container ls
ID  IMAGE  OS  ARCH  STATE  ADDR  CPUS  MEMORY
```
Then:
```
$> dagger core version
├╴▼ connecting to engine 2.2s
│ ┃ 09:04:52 INF connected name=dagger-engine-v0.19.2 client-version=v0.19.2 server-version=v0.19.2
│ ├╴▼ creating client 2.2s
│ │ ├╴✘ moby.buildkit.v1.Control/Info 0.1s ERROR
│ │ │ ! connection error: desc = "error reading server preface: command [container exec -i dagger-engine-v0.19.2 buildctl dial-stdio] has exited with exit status 1,
│ │ │   make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=Error: dial unix /run/dagger/engine.sock: connect: connection
│ │ │   refused\nUsage:\n  dial-stdio [flags]\n\nFlags:\n      --addr string   The address serving the grpc api (default \"unix:///run/dagger/engine.sock\")\n  -h, --
│ │ │   help          help for dial-stdio\n      --timeout int   The timeout in seconds for connecting to the grpc api (default 5)\n\n"
│ │ ├╴✘ moby.buildkit.v1.Control/Info 0.0s ERROR
│ │ │ ! connection error: desc = "error reading server preface: command [container exec -i dagger-engine-v0.19.2 buildctl dial-stdio] has exited with exit status 1,
│ │ │   make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=Error: dial unix /run/dagger/engine.sock: connect: connection
[...]

PR #11243, opened 2025-10-16
fix: apple container engine start

eunomie:
When we provision engine, we try to always 'start' existing container before to check the state of the container.
This works well with Docker as a 'docker start' on an already existing and started container has no impact.

But with apple containers, to 'container start' an already existing container will stop and remove the container.

Instead, check the running status of the container, and if the container is already running, to not attempt to start it.

That way, from one call to the other, the container is kept running.

Fixes: #11242

### DIFF
--- a/engine/client/drivers/apple.go
+++ b/engine/client/drivers/apple.go
@@ -94,7 +94,14 @@ func (apple) ContainerRun(ctx context.Context, name string, opts runOpts) error 
 
 	cmd := exec.CommandContext(ctx, "container", args...)
 	cmd.Env = envs
-	return traceexec.Exec(ctx, cmd)
+	_, stderr, err := traceexec.ExecOutput(ctx, cmd)
+	if err != nil {
+		if isAppleContainerAlreadyInUseOutput(stderr) {
+			return errContainerAlreadyExists
+		}
+		return err
+	}
+	return nil
 }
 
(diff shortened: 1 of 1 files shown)


engine/client/drivers/apple.go
```diff
@@ -111,15 +118,11 @@ func (apple) ContainerRemove(ctx context.Context, name string) error {
 	return traceexec.Exec(ctx, exec.CommandContext(ctx, "container", "rm", "-f", name))
 }
 
-func (apple) ContainerStart(ctx context.Context, name string) error {
-	_, stderr, err := traceexec.ExecOutput(ctx, exec.CommandContext(ctx, "container", "start", name), telemetry.Encapsulated())
-	if err == nil {
+func (a apple) ContainerStart(ctx context.Context, name string) error {
+	if running, err := a.ContainerIsRunning(ctx, name); err == nil && running {
```
jedevc, line 127: LGTM as a workaround, but FYI the reason we don't currently rely on this is because there's a race condition where the container may be running and then stop.

Not a huge issue, but that's why the current behavior is as it is.

jedevc (approved):

engine/client/drivers/apple.go
eunomie: Hum, ok I see, thanks for the details, I haven't thought about that.

I'll just add a TODO so we remember why we did this fix and to help remember to check with newer version of apple container that maybe at some point will have the same behavior than the others.

PR #11246, opened 2025-10-17
feat: auto apply changeset with -y

eunomie:
On dagger call, with `-y` or `--auto-apply`, returned `changeset` will be automatically applied without to ask the user.

For instance

    dagger -y call build

### DIFF
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -332,6 +332,7 @@ func installGlobalFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&interactiveCommand, "interactive-command", "/bin/sh", "Change the default command for interactive mode")
 	flags.BoolVarP(&web, "web", "w", false, "Open trace URL in a web browser")
 	flags.BoolVarP(&noExit, "no-exit", "E", false, "Leave the TUI running after completion")
+	flags.BoolVarP(&autoApply, "auto-apply", "y", false, "Automatically apply changes when a changeset is returned")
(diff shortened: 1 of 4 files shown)


cmd/dagger/main.go
```diff
@@ -332,6 +332,7 @@ func installGlobalFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&interactiveCommand, "interactive-command", "/bin/sh", "Change the default command for interactive mode")
 	flags.BoolVarP(&web, "web", "w", false, "Open trace URL in a web browser")
 	flags.BoolVarP(&noExit, "no-exit", "E", false, "Leave the TUI running after completion")
+	flags.BoolVarP(&autoApply, "auto-apply", "y", false, "auto apply changes when a changeset is returned")
```
shykes:
```suggestion
	flags.BoolVarP(&autoApply, "auto-apply", "y", false, "Automatically apply changes when a changeset is returned")
```

shykes (approved):
Small UI nit.

shykes:
love it

cmd/dagger/main.go
eunomie: updated, thanks

eunomie:
I've updated the doc of the flag, as well at the PR body as the flag moved from call `dagger call -y build` to the root `dagger -y call build` that I think makes more sense, especially it's less confusing between this flag and the flags of the object we are calling.

PR #11253, opened 2025-10-20
chore: remove XXX comments

jedevc:
Both of these aren't really relevant.

### DIFF
--- a/core/integration/legacy_test.go
+++ b/core/integration/legacy_test.go
@@ -1406,7 +1406,6 @@ func (m *Test) Branch(ctx context.Context, name string) (string, error) {
 
 func (LegacySuite) TestLegacyContainerBuild(ctx context.Context, t *testctx.T) {
 	// Deprecated in dagger/dagger#10811
-	// XXX: update ^^^
 	//
 	// Ensure that the old schemas still have Container.Build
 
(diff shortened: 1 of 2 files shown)


core/object.go
```diff
@@ -330,8 +330,7 @@ func (obj *ModuleObject) installConstructor(ctx context.Context, dag *dagql.Serv
 	// if no constructor defined, install a basic one that initializes an empty object
 	if !objDef.Constructor.Valid {
 		spec := dagql.FieldSpec{
-			Name: gqlFieldName(mod.Name()),
-			// Description: "TODO", // XXX(vito)
```
jedevc, line 334: From my memory, this has been here since constructors were originally added. I'm not sure what we'd even put here - no one's complained anyways.

core/integration/legacy_test.go
```diff
@@ -1406,7 +1406,6 @@ func (m *Test) Branch(ctx context.Context, name string) (string, error) {
 
 func (LegacySuite) TestLegacyContainerBuild(ctx context.Context, t *testctx.T) {
 	// Deprecated in dagger/dagger#10811
-	// XXX: update ^^^
```
jedevc, line 1409: :facepalm: oops, that's a my bad.

core/object.go
vito: I figured we should preserve whatever comment is above `func New()` - but yeah seems like no one's cared yet, fine to remove

jedevc: Mm, but this is in the case that there is no set constructor, and we're doing an auto-generated one? So there is no `func New()`.

PR #11254, opened 2025-10-20
PHP: Add static analysis to CI pipeline

charjr:
Use PHPStan in the PHP SDK Dev module to perform static analysis on the SDK. 

Helps fix potential errors before the code runs.


sdk/php/psalm.xml
charjr, line 1: Psalm is another static analysis tool but it is not used by the SDK at all.

Presumably this config file was leftover by a previous contributer and they were relying on a locally installed copy of Psalm.

sdk/php/phpstan-baseline.neon
charjr, line 1: The baseline is a file generated to contain a list of _existing warnings_ that should be ignored, _for now_.

This way it allows us to start checking for potential issues without having to make too many changes at once.

sdk/php/phpstan.dist.neon
charjr, line 1: PHPStan config, currently just checks src/

sdk/php/dev/src/PhpSdkDev.php
charjr, line 1: The coding style of this file had gone awry.

A lot of this is just styling fixes (except for adding the `analyze` function)

.dagger/sdk_php.go
```diff
@@ -80,6 +80,19 @@ func (t PHPSDK) Test(ctx context.Context) error {
 	return err
 }
 
+// Run Static Analysis on the PHP SDK
```
jedevc: Can we maybe run this as part of lint? Not the ideal place for it long term, but would be nice to get it running in CI as well 🎉

charjr: I've done so, but when rebasing I noticed [this change 3 days ago](https://github.com/dagger/dagger/pull/11161/files#diff-c8eca79572e490257ae231f110e321e794ffc0943fa417463b3e0fc1ec0c0cfb).

@shykes Am I clobbering your work by adding a static analysis tool to the PHP Lint function?

shykes: @charjr it's fine, just add it in the simplest way that makes sense to you. I'm working on part 2 of our CI refactor, which will probably move it around a little bit. But easier for you to merge first, and I'll deal with potential merge conflicts.

shykes:
Thank you for the contribution! @jedevc are you carrying this one or should I take it?

PR #11256, opened 2025-10-20
CI "brick check": make sure the engine can still run its own CI

shykes:
NOTE: this PR depends on #11161

## Problem

Our CI runs on Dagger. So, we run a stable release of Dagger, to test a dev version of Dagger.

But, what if a dev version of Dagger passes all checks, but then fails to run *its own CI*? If this defect goes undetected until the next release, our own CI could be "bricked": fails to run, but can't be fixed without pushing a change, which can't be tested because it fails to run, etc.

So, our CI includes a special check that runs a small subset of our CI on a dev engine. This way, we can catch any issue that would risk "bricking" our CI.

But: the purpose of that check is not obvious from the code, configuration or trace output. This results in confusion - "why is this check running, and why is it slightly different from the others? Is it a duplicate?"

See screenshot for an example. 
image

## Solution

Make the purpose of the "no brick" check more explicit.

We don't add or remove any CI checks. We merely rename one, to make its purpose more clear.

### DIFF
--- a/.dagger/checks.go
+++ b/.dagger/checks.go
@@ -89,6 +89,17 @@ func (dev *DaggerDev) CheckTestHelm(ctx context.Context) error {
 	return dag.Helm().Test(ctx)
 }
 
+// "No-brick" check: verify that the version of dagger being checked by this module, can itself run this module.
+// In other words, we use Dagger's own CI as a test workload in Dagger's CI.
+// We only run a small subset of the module, selected to be as representative as possible, without wasting compute.
+// NOTE: the caller is responsible for calling this with the right engine.
+// FIXME: use dagger-in-dagger to no longer require on the caller.
+func (dev *DaggerDev) CheckNoBrick(ctx context.Context) error {
+	// Testing SDKs is not as compute-intensive as the core tests,
+	// and exercises more engine features, including loading modules across SDKs.
+	return dev.CheckTestSDKs(ctx)
+}
+
 // Run all checks for all SDKs
 func (dev *DaggerDev) CheckTestSDKs(ctx context.Context) error {
 	type tester interface {
--- a/.github/main.go
+++ b/.github/main.go
@@ -183,9 +183,9 @@ func New() *CI {
 		).
 		withWorkflow(
 			ci.AltRunner,
-			true,
-			"Dev Engine",
-			"check-test-sdks",
+			true,       // run this on a dev engine
+			"No brick", // Check that the engine being tested, can itself run the testing module (to avoid being "bricked")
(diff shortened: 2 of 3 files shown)


.dagger/checks.go
```diff
@@ -89,6 +89,17 @@ func (dev *DaggerDev) CheckTestHelm(ctx context.Context) error {
 	return dag.Helm().Test(ctx)
 }
 
+// "No-brick" check: verify that the version of dagger being checked by this module, can itself run this module.
+// In other words, we use Dagger's own CI as a test workload in Dagger's CI.
+// We only run a small subset of the module, selected to be as representative as possible, without wasting compute.
+// NOTE: the caller is responsible for calling this with the right engine.
+// FIXME: use dagger-in-dagger to no longer require on the caller.
+func (dev *DaggerDev) CheckNoBrick(ctx context.Context) error {
```
jedevc, line 97: I don't really think `NoBrick` is much clearer tbf - it might not fail out-right, we could always have some "other problem".

As a suggestion we could just call it `CheckDaggerCI`.

.dagger/checks.go
```diff
@@ -89,6 +89,17 @@ func (dev *DaggerDev) CheckTestHelm(ctx context.Context) error {
 	return dag.Helm().Test(ctx)
 }
 
+// "No-brick" check: verify that the version of dagger being checked by this module, can itself run this module.
+// In other words, we use Dagger's own CI as a test workload in Dagger's CI.
+// We only run a small subset of the module, selected to be as representative as possible, without wasting compute.
+// NOTE: the caller is responsible for calling this with the right engine.
+// FIXME: use dagger-in-dagger to no longer require on the caller.
```
jedevc, line 96: Have a rough implementation:

```go
func (dev *DaggerDev) CheckNoBrick(ctx context.Context) error {
	ctr, err := dev.Dev(ctx, dev.Source, DistroAlpine, false, false)
	if err != nil {
		return err
	}
	ctr = ctr.WithMountedDirectory(".git/", dev.Git.Head().Tree().Directory(".git/"))

	_, err = ctr.WithExec([]string{"sh", "-c", "dagger call check-test-sdks"}).Sync(ctx)
	return err
}
```

Then we can *entirely* remove the dev engine dance from our CI.

Happy to PR this, or add it on top, or for you to just `Co-authored-by` me (whatever works).

shykes: Nice. I'm going to close this, and will add a `co-authored-by` commit on a separate PR with other related CI follow-up improvements.

shykes:
Closing in favor of #11262

Issue #7432, opened 2024-05-21
Toolchains


shykes:
*This proposal is an alternative to #7199. The problem is the same, but the proposed solution is different.*

## Problem

There are two types of Dagger modules: those that are standalone software projects, and those that exist in the context of another software project. Let's call those *standalone modules* and *contextual modules*, respectively.

Dagger is designed to support both, which is good, but causes friction for users in some areas.

- Contextual modules almost always need access to their context directory. Today this requires an explicit argument, (which usually looks like `dagger call MYFUNC --source=.`), which is verbose.
- Contextual modules often need to filter their context directory, usually to avoid the prohibitive cost of uploading a huge monorepo on every task.
- The best practice for contextual modules is to make the context directory itself a module, and store the module source code in a configurable subdirectory. This additional concept of "module source" being different from "module root", adds complexity for the user.
- For standalone modules, the recommended value for `source` is `.`. For contextual modules, it is `./dagger`. The current default value is `./dagger`, which hurts the experience of creating a standalone module. If we change it to `.`, it will hurt the experience of creating contextual modules. Either way, the user experience suffers.

## Solution

I propose introducing the concept of *toolchain* as a first-class concept in Dagger.

## What is a toolchain?

According to gpt4o:

> A toolchain is a set of programming tools used to perform complex software development tasks, typically including:
>
> 1. Compiler: Translates source code into executable code.
> 2. Linker: Combines object files into a single executable or library.
> 3. Debugger: Helps identify and fix bugs.
> 4. Version Control: Manages changes to source code over time.
> 5. Build Automation: Automates the build process, often using tools like Make or CI/CD systems.
>
> Each tool in the chain works together to streamline development and deployment.

Every software project has a toolchain. And every daggerized project has a toolchain implemented as Dagger module.

## Daggerizing a toolchain

When a project adopts Dagger, it designates a Dagger module as that project's toolchain. The toolchain could be purpose-built for the project (eg the`./ci` module in `github.com/dagger/dagger`) or more generic (for example, a Docker toolchain that understands `Dockerfile` and `docker-compose.yaml`; or a Ruby-on-Rails toolchain for developing RoR apps). Most projects start using an off-the-shelf toolchain, then eventually create their own as the project becomes more complex. Dagger can support both equally well.

### How it works

A Dagger module can designate another module as its toolchain. This is done with the `toolchain` field in `dagger.json`. Any module can be used as a toolchain for another module.

The rules are for the same as for dependencies:

[...]

PR #11262, opened 2025-10-21
CI: follow-up improvements

shykes:
Continuing the streak of improvements started in #11161.

## TODO

- [x] identify checks by their return type, not a `Check` prefix in their name. This is in preparation for, and to help the design of, [an upstream "checks" feature](#11211).
- [x] update contributing guide to match new function names
- [x] make sure dagger can still run its own CI (deprecates [11256](#11256))
- [x] fix typo in GHA config
- [x] simplify GHA config by eliminating unused features: 'dev engine', customizable runner in happy path, unused runner configs
- [x] Rename `dagger call dev` to `dagger call playground` to better reflect functionality
- [x] Consolidate Go operations (lint, test) in a centralized go "toolchain", in preparation for #7432 

## For follow-up

- [ ] remove dependency on native SDK dev modules, when possible (for faster loading)
- [ ] FIXME in `util/parallel` (don't choose between accurate span duration, and repeatable span order)


.dagger/stopgap_check_status.go
```diff
@@ -0,0 +1,9 @@
+package main
+
+// FIXME: stopgap until core API defines CheckStatus
+type CheckStatus string
+
+const (
+	CheckCompleted CheckStatus = "COMPLETED"
+	CheckSkipped   CheckStatus = "SKIPPED"
+)
```
jedevc: Maybe not the place to discuss, but just to check - these can't be compounded with an error.

So I can't have an errored+skipped check - there's really 3 states here right?
- Completed
- Skipped
- Errored (the function returned an error)

.dagger/checks.go
```diff
@@ -7,97 +7,115 @@ import (
...
+		}).
+		WithJob("Engine", func(ctx context.Context) error {
+			_, err := dag.DaggerEngine().ReleaseDryRun(ctx)
+			return err
+		}).
+		WithJob("SDKs", func(ctx context.Context) error {
 			jobs := parallel.New()
-			for _, sdk := range allSDKs[dryRunner](dev) {
-				jobs = jobs.WithJob(sdk.Name(), sdk.CheckReleaseDryRun)
+			for name, fn := range map[string]func(context.Context) (CheckStatus, error){
```
jedevc: We should use a slice here, map iteration is random :cry: 

Also, the `allSDKs` helper isn't used here? I liked that!

util/parallel/parallel.go
```diff
@@ -37,6 +37,15 @@ func (p parallelJobs) WithJob(name string, fn JobFunc) parallelJobs {
 	return p
 }
 
+// NewJobFunc converts a function that returns a value and an error into a JobFunc
+// by discarding the returned value and only propagating the error.
+func NewJobFunc[ReturnType any](fn func(context.Context) (ReturnType, error)) JobFunc {
```
jedevc: Neat. But unused?

.github/main.go
```diff
@@ -86,107 +68,18 @@ func New() *CI {
...
-		}),
 	}
-
 	return ci.
-		withWorkflow(
-			ci.AltRunnerWithCache,
-			false,
-			"Check generated files",
-			"check-generated",
-		).
```
jedevc, line 113: Because this is now gone, we can remove the `start-dev-dagger.sh` script in `modules/gha` and everything that needed it.

Should be a nice simplification, the more we can strip out of that module, the closer we can see we're getting to native CI.

.dagger/stopgap_check_status.go
shykes: `Status` only indicates whether the check completed or not. It doesn't say if the checked passed or failed. That information is sent separately, by returning an error (or not).

I can't combine these two properties (status and success) in a single type, because errors are a special case in our API, and developers would have to manually convert their errors to a special type, which would be a PITA.

We could even remove `CheckSkipped`, it's a nice-to-have. The real purpose of `CheckStatus` is simply to exist as a return value, so that checks can be detected. The actual value is unimportant (except to distinguish skipped checks)

.dagger/checks.go
shykes: Re `allSDKs`: I liked it too... Long story short, it's a massive pain to use with the new signature of checks... At least for now, because not all the checks return the same type (each module has to re-define their own `CheckStatus` type as a stopgap until engine has a native type for it). As a result, there is no interface that will match all checks functions... I implemented a workaround with `reflect` but it just was getting too complicated. So, since `dagger checks` will deprecated this aggregation feature anyway... I'm hardcoding the list for now.

util/parallel/parallel.go
shykes: Leftover from my attempt at circumventing the pain of using `allSDKs` with new check signature + fragmented types...

.github/main.go
shykes: Done :) That felt good

.dagger/checks.go
shykes: Mmm, actually in this case I can use `allSDKs` since all checks are from the monolith module. Let me bring that back.

shykes: It's back!

shykes:
@jedevc I think `ci-in-ci` is failing because we don't mount our docker config, and as a result we get rate-limited... https://dagger.cloud/dagger/traces/a171641758d116ed0e68a8ac4a1d3235

shykes:
> @jedevc I think `ci-in-ci` is failing because we don't mount our docker config, and as a result we get rate-limited... https://dagger.cloud/dagger/traces/a171641758d116ed0e68a8ac4a1d3235

FIxed

PR #11263, opened 2025-10-21
h2c: fix ineffective conn assignment

vito:
Haven't sunk any time figuring out a repro for this, but seems obviously wrong from a look at the code.

### DIFF
--- a/engine/buildkit/client.go
+++ b/engine/buildkit/client.go
@@ -467,7 +467,7 @@ func (c *Client) ListenHostToContainer(
 			connsL.Unlock()
 
 			if !found {
-				conn, err := c.Dialer.Dial(proto, upstream)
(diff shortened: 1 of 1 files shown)


shykes (approved):

vito:
Papertrail: `test-everything-else` failed on the [first run](https://dagger.cloud/dagger/traces/557019ff228a0c864febe64785a135b7), in a suspiciously related looking test, but it passed in the next two reruns and counting.

vito:
Hasn't failed in a few runs, and judging by the original failure, which was a timeout, even [`echo hello` was taking 51s](https://dagger.cloud/dagger/traces/557019ff228a0c864febe64785a135b7?span=cb4c358a5a5f5260) so I'm gonna chalk that up to infra overload.

PR #11264, opened 2025-10-21
llm: support saving/loading sessions

vito:
## Goals

* [ ] Uploading changes _from_ local state
* [ ] Saving and loading LLM sessions
* [ ] (Auto-)compacting message history

### WARNING: Breaking changes

`LLM.loop` is now required to actually initiate the LLM loop. Previously it
was "lazy" and happened when you requested the last reply, but that pattern
required the LLM object to update in-place, whereas now we want an ID returned
that includes the message history.

### DIFF
--- a/core/telemetry_test.go
+++ b/core/telemetry_test.go
@@ -71,7 +71,8 @@ func (ms *mockServer) Auth(context.Context) (*auth.RegistryAuthProvider, error) 
 
 func (ms *mockServer) Buildkit(context.Context) (*buildkit.Client, error) { return nil, nil }
 
-func (ms *mockServer) Services(context.Context) (*Services, error) { return nil, nil }
+func (ms *mockServer) Services(context.Context) (*Services, error)     { return nil, nil }
+func (ms *mockServer) MCPClients(context.Context) (*MCPClients, error) { return nil, nil }
 
 func (ms *mockServer) Platform() Platform               { return Platform{} }
 func (ms *mockServer) OCIStore() content.Store          { return nil }


dagql/call/callpbv1/call.proto
```diff
@@ -75,6 +75,14 @@ message Call {
...
   // "recipe-based" digest
   bool isCustomDigest = 11;
+
+  // Whether to re-fetch the value and update its cache key, rather than seek a
+  // cache hit.
+  //
+  // This flag allows IDs to be re-written in "live" form, allowing the ID to
+  // encode how the data is fetched, what the current cache key is, and whether
+  // to fetch a new version.
+  bool reload = 12;
```
vito, line 85: @jedevc Worth bikeshedding this re: #11158 

This commit has the relevant changes together, provided the link still works: https://github.com/dagger/dagger/pull/11264/commits/e80d4c1838df775bd1394083fb4cc47b96525fcd

**Context**
This is to handle the case where, with an LLM, you've installed a module like `dagger-dev`, where to run tests you call `test specific` - i.e. there's an intermediate object, `DaggerDevTest`. For an LLM, that's two discrete calls, with `test` returning something like `DaggerDevTest#1`, and then `DaggerDevTest_specific` being called against that. The trouble is, unless we do something about it, that `DaggerDevTest#1` will contain whatever original snapshot of the workspace existed at the time it was first called. When an LLM does an `EditFile`, and then calls `DaggerDevTest_specific`, it won't pick up those changes.

The solution on `main` is to make `DaggerDevTest#1` actually be a 'contextual object' that gets re-hydrated on every use. But, that works by embedding a Go callback in `*core.MCP` that re-constructs it fresh every time. Right now I'm trying to make LLMs serializable for save/reload, so a callback is a no-go; we need a way to serialize a live-updating ID.

So, the idea in this PR is to store an un-pinned version of the object's ID, where contextual args have been stripped out, and the ID is marked in a way that tells DagQL to reload it - hence this field.

I know in your live development PR you're also doing ID walking to rewrite `Host.directory` calls to force parts of it to reload, so it would be neat if the two techniques could converge. Maybe it would be a small tweak, where instead of setting a random digest, you just set `reload: true`? 🤔 

cc @sipsma for general bikeshedding since we're in the guts of a very central component

vito: Quick note: I think we'll have to do a bit more than the code I have here so far, to properly support this.

For example, say you rewrote an ID to contain a `reload: true` object in an _argument_. Technically that reloading should leak outward, right? Otherwise we risk a cache hit on the outer ID, whose digest is based on this value that needs reloading. What we really want is to reload the arg and THEN seek a cache hit for the resulting now-pinned ID.

Feels kinda like the `taint: true` flag we had before, except we don't necessarily want to consider the outer ID tainted, we just need to refresh everything deeply before considering the outer cache keys.

Something like that anyway.

(Maybe related to https://github.com/dagger/dagger/pull/11277 `SelectID`?)

PR #11267, opened 2025-10-22
feat(typescriptSDK): decouple module introspection from runtime using `ModuleTypes`

TomChv:
Additional changes:
- Pin Typescript dep to `v5.9.3` so we always download the same version
- Remove register typedef from ts runtime entrypoint
- Remove file loading when introspecting
- Remove unused functions

Before and after comparison

Screenshot 2025-10-21 at 17 16 39

### DIFF
--- a/sdk/typescript/runtime/introspector.go
+++ b/sdk/typescript/runtime/introspector.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"path/filepath"
 	"typescript-sdk/internal/dagger"
 	"typescript-sdk/tsdistconsts"
 )
@@ -19,7 +18,8 @@ type Introspector struct {
 func NewIntrospector(sdkSourceDir *dagger.Directory) *Introspector {
 	ctr := dag.
 		Container().
-		From(tsdistconsts.DefaultBunImageRef).
+		From(tsdistconsts.DefaultAlpineImageRef).
+		WithExec([]string{"apk", "add", "libgcc", "libstdc++"}).
 		WithMountedFile(introspectorBinPath, sdkSourceDir.File(introspectorBinPath)).
 		WithMountedDirectory(typescriptLibPath, sdkSourceDir.Directory("typescript-library")).
 		WithWorkdir(ModSourceDirPath)
@@ -39,10 +39,8 @@ func (i *Introspector) AsEntrypoint(
 	clientBindings *dagger.File,
 ) *dagger.Container {
 	return i.Ctr.
-		WithDirectory(ModSourceDirPath, sourceCode, dagger.ContainerWithDirectoryOpts{
-			Include: []string{"src"},
-		}).
-		WithFile(filepath.Join(ModSourceDirPath, "sdk/client.gen.ts"), clientBindings).
+		WithMountedDirectory("src", sourceCode).
+		WithMountedFile("sdk/client.gen.ts", clientBindings).
 		WithEnvVariable("TYPEDEF_OUTPUT_FILE", outputFilePath).
 		WithEntrypoint([]string{introspectorBinPath, moduleName, "src", "sdk/client.gen.ts"})
 }
--- a/sdk/typescript/runtime/main.go
+++ b/sdk/typescript/runtime/main.go
@@ -79,7 +79,7 @@ func (t *TypescriptSdk) ModuleTypes(
 	return NewIntrospector(t.SDKSourceDir).
 		AsEntrypoint(outputFilePath,
 			cfg.name,
-			modSource.ContextDirectory().Directory(cfg.subPath),
+			modSource.ContextDirectory().Directory(cfg.subPath).Directory("src"),
 			clientBindings,
 		), nil
 }


sdk/typescript/runtime/module.go
```diff
@@ -149,7 +149,7 @@ func (m *moduleRuntimeContainer) configurePackageJSON(file *dagger.File) *dagger
 	if m.cfg.packageJSONConfig != nil {
 		_, ok := m.cfg.packageJSONConfig.Dependencies["typescript"]
 		if !ok {
-			ctr = ctr.WithExec([]string{"npm", "pkg", "set", "dependencies.typescript=^5.5.4"})
+			ctr = ctr.WithExec([]string{"npm", "pkg", "set", "dependencies.typescript=5.9.3"})
```
jedevc: Should be a constant that's not buried here.

Any suggestions for how to make sure we keep this up-to-date? Dependabot won't catch it.

TomChv: Ideally we should keep the same version that dependabot set in the `typescript/sdk` package.json

grouville (approved):
LGTM ✅

#### Non blocking

Moving the resolver fully onto the checker makes sense (no more executing the user module), but now the [PropertyAccessExpression](https://github.com/dagger/dagger/pull/11267/files#diff-94d962fd2257b8d4b9f1f2a241b0943e168d9cfd7c2b719379d9f5e536d9b25cR497) branch only returns enum members.

_Example:_ 

```typescript
import { object, func } from "@dagger.io/dagger"

export const defaults = {
  greetings: { english: "hello dagger" },
} as const

@object()
export class Demo {
  @func()
  async say(message: string = defaults.greetings.english): Promise<string> {
    return message
  }
}
  ```
 
 now returns: 
 
 ```shell
 $ dag call say --help
▶ connect 0.5s
▶ load module: . 3.0s

-

USAGE
  dagger call say [arguments]

ARGUMENTS
      --message string
 ```
 
 whereas on main, we would still see `(default "hello dagger")`
 
That means defaults like `message = defaults.greetings.english` no longer surface in the typedef. At runtime, everything still works and message stays optional, so the impact is only "cosmetic": dagger call --help/generated clients stop displaying the real default (since TypeScript treats these parameters as optional)

Simpler defaults still work:

```ts
@object()
export class Demo {
  @func()
  async say(message: string = "hello dagger"): Promise<string> {
    return message
  }
}
```
 
This was previously handled by the old `getValueByExportedName` traversal, but we never had a fixture covering it, so it wasn’t an officially supported scenario.
 
 I’m fine letting it as a follow-up IF a user requires for it

sdk/typescript/runtime/introspector.go
```diff
@@ -0,0 +1,48 @@
...
+
+type Introspector struct {
+	//+private
+	Ctr *dagger.Container
+}
+
+func NewIntrospector(sdkSourceDir *dagger.Directory) *Introspector {
+	ctr := dag.
+		Container().
+		From(tsdistconsts.DefaultBunImageRef).
```
sipsma: we always use bun because it's the fastest? or something else? Just curious, not a problem at all

sdk/typescript/runtime/introspector.go
```diff
@@ -0,0 +1,48 @@
...
+	outputFilePath string,
+
+	moduleName string,
+
+	sourceCode *dagger.Directory,
+
+	clientBindings *dagger.File,
+) *dagger.Container {
+	return i.Ctr.
+		WithDirectory(ModSourceDirPath, sourceCode, dagger.ContainerWithDirectoryOpts{
```
sipsma: nit: if this can be `WithMountedDirectory` that would be a bit more efficient by avoiding copying

sdk/typescript/runtime/module.go
```diff
@@ -149,7 +160,7 @@ func (m *moduleRuntimeContainer) configurePackageJSON(file *dagger.File) *dagger
 	if m.cfg.packageJSONConfig != nil {
 		_, ok := m.cfg.packageJSONConfig.Dependencies["typescript"]
 		if !ok {
-			ctr = ctr.WithExec([]string{"npm", "pkg", "set", "dependencies.typescript=^5.5.4"})
+			ctr = ctr.WithExec([]string{"npm", "pkg", "set", fmt.Sprintf("dependencies.typescript=%s", tsdistconsts.DefaultTypeScriptVersion)})
```
sipsma, line 163: just double checking that dropping the `^` was intentional here (not personally sure whether it matters or not)

TomChv: Yeah it's totally on purpose, so we always load by default the same `typescript` version instead of letting the package manager download the latest available version that matches the pattern.

sdk/typescript/runtime/introspector.go
TomChv: I wonder if I could just use the alpine base image, I'll give it a try

sdk/typescript/runtime/introspector.go
TomChv: I'll give it a try

sdk/typescript/runtime/introspector.go
TomChv: Done [f0b76ec](https://github.com/dagger/dagger/pull/11267/commits/f0b76ec99c361a230a3e04d47aef1a9d74ac04f9)

sdk/typescript/runtime/introspector.go
TomChv: Done [f0b76ec](https://github.com/dagger/dagger/pull/11267/commits/f0b76ec99c361a230a3e04d47aef1a9d74ac04f9)

PR #11271, opened 2025-10-22
feat: add Module.call

jedevc:
This PR adds a new `Module.call` API, which allows invoking a function call in a module, *without* going through the GraphQL API.

For example, take the following module:

```go
package main

type Test struct {
	X int
}

func (m *Test) Run(Y int) int {
	return m.X + Y
}
```

The following query is now possible:

```graphql
query {
  moduleSource(refString: ".") {
    asModule {
      call(
        object: "Test"
        function: "run"
        parent: "{\"X\": 101}"
        inputs: "{\"Y\": 202}"
      )
    }
  }
}
```
```json
{
    "moduleSource": {
        "asModule": {
            "call": "303"
        }
    }
}
```

This is intended for use in our scale-out functionality - an engine can remotely call another engine using this.

### DIFF
--- a/core/modfunc.go
+++ b/core/modfunc.go
@@ -93,9 +93,14 @@ func NewModFunction(
 }
 
 type CallOpts struct {
-	Inputs         []CallInput
-	ParentTyped    dagql.AnyResult
-	ParentFields   map[string]any
+	CallID *call.ID
+
+	Inputs []CallInput
+
+	ParentTyped   dagql.AnyResult
+	ParentFields  map[string]any
+	ParentModType *ModuleObjectType
+
 	SkipSelfSchema bool
 	Server         *dagql.Server
 
@@ -578,6 +583,8 @@ func (fn *ModuleFunction) CacheConfigForCall(
 			return nil, err
 		}
 		dgstInputs = append(dgstInputs, clientMetadata.SessionID)
+	} else if cachePolicy == FunctionCachePolicyNever {
+		cacheCfgResp.CacheKey.DoNotCache = true
 	}
 
 	cacheCfgResp.CacheKey.CallKey = hashutil.HashStrings(dgstInputs...).String()
@@ -605,7 +612,7 @@ func (fn *ModuleFunction) loadFunctionRuntime(ctx context.Context) (runtime dagq
 	return runtime, nil
 }
 
-func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.AnyResult, rerr error) { //nolint: gocyclo
+func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (_ dagql.Typed, _ dagql.AnyResult, rerr error) { //nolint: gocyclo
 	mod := fn.mod
 
 	lg := bklog.G(ctx).WithField("module", mod.Name()).WithField("function", fn.metadata.Name)
@@ -620,13 +627,12 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Any
 
 	clientMetadata, err := engine.ClientMetadataFromContext(ctx)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	callID := dagql.CurrentID(ctx)
 	execMD := buildkit.ExecutionMetadata{
 		ClientID:          identity.NewID(),
-		CallID:            callID,
+		CallID:            opts.CallID,
 		ExecID:            identity.NewID(),
 		Internal:          true,
 		ParentIDs:         map[digest.Digest]*resource.ID{},
@@ -639,12 +645,11 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Any
 	} else {
 		cacheMixins = append(cacheMixins, cache.CurrentStorageKey(ctx))
 	}
-
 	execMD.CacheMixin = hashutil.HashStrings(cacheMixins...)
 
 	callInputs, err := fn.setCallInputs(ctx, opts)
 	if err != nil {
-		return nil, fmt.Errorf("set call inputs: %w", err)
+		return nil, nil, fmt.Errorf("set call inputs: %w", err)
 	}
 
 	bklog.G(ctx).Debug("function call")
@@ -657,7 +662,7 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Any
 
 	parentJSON, err := json.Marshal(opts.ParentFields)
 	if err != nil {
-		return nil, fmt.Errorf("marshal parent value: %w", err)
+		return nil, nil, fmt.Errorf("marshal parent value: %w", err)
 	}
 
 	if opts.ParentTyped != nil {
@@ -668,27 +673,27 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Any
 			AsObject: dagql.NonNull(fn.objDef),
 		}, true)
 		if err != nil {
-			return nil, fmt.Errorf("get mod type for parent: %w", err)
+			return nil, nil, fmt.Errorf("get mod type for parent: %w", err)
 		}
 		if !ok {
-			return nil, fmt.Errorf("find mod type for parent %q", fn.objDef.Name)
+			return nil, nil, fmt.Errorf("find mod type for parent %q", fn.objDef.Name)
 		}
 		if err := parentModType.CollectCoreIDs(ctx, opts.ParentTyped, execMD.ParentIDs); err != nil {
-			return nil, fmt.Errorf("collect IDs from parent fields: %w", err)
+			return nil, nil, fmt.Errorf("collect IDs from parent fields: %w", err)
 		}
 	}
 
 	if mod.ResultID != nil {
 		execMD.EncodedModuleID, err = mod.ResultID.Encode()
 		if err != nil {
-			return nil, fmt.Errorf("encode module ID: %w", err)
+			return nil, nil, fmt.Errorf("encode module ID: %w", err)
 		}
 	}
 
 	fnCall := &FunctionCall{
 		Name:      fn.metadata.OriginalName,
 		Parent:    parentJSON,
-		ParentID:  callID.Receiver(),
+		ParentID:  opts.CallID.Receiver(),
 		InputArgs: callInputs,
 	}
 	if envID, ok := EnvIDFromContext(ctx); ok {
@@ -699,14 +704,14 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Any
 	}
 	execMD.EncodedFunctionCall, err = json.Marshal(fnCall)
 	if err != nil {
-		return nil, fmt.Errorf("marshal function call: %w", err)
+		return nil, nil, fmt.Errorf("marshal function call: %w", err)
 	}
 
 	srv := dagql.CurrentDagqlServer(ctx)
 
 	runtime, err := fn.loadFunctionRuntime(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("failed to load runtime: %w", err)
+		return nil, nil, fmt.Errorf("failed to load runtime: %w", err)
 	}
 
 	var metaDir dagql.ObjectResult[*Directory]
@@ -716,7 +721,7 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Any
 		},
 	)
 	if err != nil {
-		return nil, fmt.Errorf("create mod metadata directory: %w", err)
+		return nil, nil, fmt.Errorf("create mod metadata directory: %w", err)
 	}
 
 	var ctr dagql.ObjectResult[*Container]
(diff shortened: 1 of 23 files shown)


core/schema/module.go
```diff
@@ -590,6 +595,111 @@ func (s *moduleSchema) moduleServe(ctx context.Context, modMeta *core.Module, ar
...
+	})
+
+	result, err := callable.Call(ctx, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	// XXX: this marshals the whole *object*, but we really just want the fields!
+	// this is tricky, what should this even look like.
+	dt, err := json.Marshal(result.Unwrap())
```
jedevc: Okay, so the version I hacked together at the scale-out hackathon really didn't work :thinking:

It worked for primitive types, but wouldn't work properly if you tried taking/returning your own object (which was fine, because we didn't allow scaling those out in the first place). But we should probably have the `call` API support it anyways.

@sipsma I'm thinking that we should probably have all objects marshaled into a wire-format for the query that's similar to how they're passed to SDKs? Thinking about *just* using `ConvertToSDKInput` and such, and working around the ID problems I had with that. Or is it worth splitting it to separate methods on the `ModType`, and having a separate `query` interface here.

sipsma: First impression is that using `ConvertToSDKInput` feels correct given it's how parent objects are provided to SDKs over-the-wire already anyways. I don't remember what problems you had with IDs there off the top of my head though, was it just the wrong type in the method signatures?

jedevc: `ConvertToSDKInput` feels kinda right, but in this case, what I want is to take the `dagql.AnyResult` I get from calling the `Callable`, and instead turn it into some representation to return as a JSON object from `Module.call`.

Definitely should *work*, but given that the graphql server here *isn't* an SDK, does is still make sense? (at least for now)

> I don't remember what problems you had with IDs there off the top of my head though, was it just the wrong type in the method signatures?

This was more in determining which functions were eligible for scale-out, I was having issues where a `dagql.Typed` `DynamicID` was provided as an input, and I wanted to determine whether it had any core IDs in the object fields. (but that's a separate kind of problem, I think I have an idea for that one too)

jedevc: Cool, have just used the SDK wire format then.

core/modtypes.go
```diff
@@ -5,26 +5,30 @@ import (
...
-	// objects
-	ConvertFromSDKResult(ctx context.Context, value any) (dagql.AnyResult, error)
+	// ConvertFromSDKResult converts a value returned from an SDK into a value
+	// expected by the server including conversion of IDs to their "unpacked"
+	// objects.
+	//
+	// It returns two values, the first as a conversion to a standard
+	// dagql.Typed, with no additional wrapping, and a dagql.AnyResult with all
+	// the appropriate result construction.
+	ConvertFromSDKResult(ctx context.Context, id *call.ID, value any) (dagql.Typed, dagql.AnyResult, error)
```
jedevc, line 23: This is the main "plumbing" needed to make this change happen: essentially, allow converting from an SDK result even if we aren't associated with a call ID.

We're always able to convert to some `dagql.Typed` by just decoding and loading the value into the type - but if there's no associated call ID, we can't actually create a result (most of the time). e.g. if I return a `[]*dagger.File`, then while each element has an ID, the array itself doesn't - there's no possible result in this case.

I chose to merge this functionality into `ConvertFromSDKResult` - this was already doing the decoding, and it wasn't too tricky to extend it (duplicating all the code felt messy).

Additionally, I made it a bit clearer by removing the use of `dagql.CurrentID` in here, and instead just use the explicitly passed ID. This was kind of confusing but also needed to clarify the caller's intention (do they have an associated call id or not).

core/schema/module.go
```diff
@@ -658,6 +672,205 @@ func (s *moduleSchema) moduleServe(ctx context.Context, modMeta *core.Module, ar
...
+		argType, err := callable.ArgType(name)
+		if err != nil {
+			return inst, fmt.Errorf("failed to get type for argument %q: %w", name, err)
+		}
+		// NOTE: we use the SDK wire format here, even though this isn't really a SDK
+		// Mostly, this is just because it's easily available.
+		decodedValue, _, err := argType.ConvertFromSDKResult(ctx, nil, value)
+		if err != nil {
+			return inst, fmt.Errorf("failed to decode input %q: %w", name, err)
+		}
```
jedevc, line 753: This is where we use the new plumbing - it's not correct to use the `AnyResult` that we would have returned, because this is a manually created argument, there's no actual associated ID.

Side-note: why not actually use IDs? Well, that only works if it's all definitely in one engine, as soon as we start doing cross-engine calls, we need to send all the fields as part of the request - if we just send IDs, we'd have to replay the entire query, which is 1. wasted work, and 2. maybe incorrect if it's not deterministic.

core/schema/module.go
```diff
@@ -658,6 +672,205 @@ func (s *moduleSchema) moduleServe(ctx context.Context, modMeta *core.Module, ar
...
+	}
+
+	inst, err = dagql.NewResultForCurrentID(ctx, core.JSON(dt))
+	if err != nil {
+		return inst, err
+	}
+
+	// HACK: this isn't quite right, but there's no way to actually transfer
+	// secrets between calls of this form
+	inst = inst.ResultWithSafeToPersistCache(true)
```
jedevc, line 797: 🤷 

This seems right? I think? At least, the basic caching test I added works: JSON should always be safe to persist!

One thing though - I have no idea how resource transfer could even work here, so I deliberately left it out, it seems out of scope for now. How can resources be transferred over multiple engines? Maybe we could actually return them as part of our response?

core/schema/module.go
```diff
@@ -658,6 +672,205 @@ func (s *moduleSchema) moduleServe(ctx context.Context, modMeta *core.Module, ar
...
+	}
+
+	resp, err := callable.CacheConfigForCall(ctx, mod, inputs, dagql.CurrentID(ctx).View(), req)
+	if err != nil {
+		return nil, err
+	}
+	// Prefix the call key with a special string to avoid potential collisions
+	// with the "actual" query (this function returns JSON, not the actual type)
+	resp.CacheKey.CallKey = "moduleCall:" + resp.CacheKey.CallKey
+	resp.UpdatedArgs = nil
```
jedevc, line 851: Here, we're just hijacking the module function's cache settings and using them ourselves.

But we need to inject some magic string so that we don't mixup the results of this with the results of the actual function call. Maybe there's a super clever way to avoid this? I'm not sure how though, *unless* we had `objFun` call `Module.call` under the hood.

core/modfunc.go
```diff
@@ -578,6 +583,8 @@ func (fn *ModuleFunction) CacheConfigForCall(
 			return nil, err
 		}
 		dgstInputs = append(dgstInputs, clientMetadata.SessionID)
+	} else if cachePolicy == FunctionCachePolicyNever {
+		cacheCfgResp.CacheKey.DoNotCache = true
```
jedevc, line 587: @sipsma this seemed to be missing? Couldn't tell if it was deliberate...

Essentially, I wanted to be able to have the handler's cache key function able to set the cache policy (so I could get a different cache policy per function I called), but we weren't actually propagating it here.

I haven't handled the TTL case, but that should be trivial if we want that to work.

core/modtypes.go
sipsma: > Additionally, I made it a bit clearer by removing the use of dagql.CurrentID in here, and instead just use the explicitly passed ID. 

LGTM, feels much better to be explicit I agree (independent of everything else)

---

> This is the main "plumbing" needed to make this change happen: essentially, allow converting from an SDK result even if we aren't associated with a call ID.
>
> We're always able to convert to some dagql.Typed by just decoding and loading the value into the type - but if there's no associated call ID, we can't actually create a result (most of the time). e.g. if I return a []*dagger.File, then while each element has an ID, the array itself doesn't - there's no possible result in this case.

I almost wonder if we should just be sending the ID as part of `Module.call` and then using that here? I understand the distinction between the `Typed` and `AnyResult` return values but it's extremely confusing at first glance. I feel like I'd almost rather just send over that ID too and keep `ConvertFromSDKResult` (and `Callable.Call`, etc.) simpler rather than maintain the two return values forever. I feel like there's a chance we'll end up wanting that "original call ID" someday for other reasons too 🤷‍♂️

core/schema/module.go
sipsma: > This seems right? I think? At least, the basic caching test I added works: JSON should always be safe to persist!

Every function call (even the local ones) return JSON already, we then look through the deser'd object for `SetSecret` calls and disable cache persistence if found.

AFAICT all of that is equally as valid to `Module.call` as it is to the normal path of function calls.

But the thing is `Callable.Call` still returns an `AnyResult` where we've already figured all that out, so why not just pick up the value from there?

i.e.
```go
typed, res, err := callable.Call(ctx, opts)
if err != nil {
    return inst, err
}

// ...

inst = inst.ResultWithSafeToPersistCache(res.IsSafeToPersistCache())
```

> One thing though - I have no idea how resource transfer could even work here, so I deliberately left it out, it seems out of scope for now. How can resources be transferred over multiple engines? Maybe we could actually return them as part of our response?

Yeah we could return them in the response provided they aren't SetSecret by just sending the secret provider value over...

If we want to handle SetSecret though we may require some separate channel I think; either a new API endpoint or somehow piggy backing on session attachables.

Doesn't matter until we support returning non-scalars though 🙂

core/schema/module.go
sipsma: > But we need to inject some magic string so that we don't mixup the results of this with the results of the actual function call. Maybe there's a super clever way to avoid this? I'm not sure how though, _unless_ we had `objFun` call `Module.call` under the hood.

Makes sense to start here for sure, but yeah the only reasonable solution in time is to get the two paths of calling a module function to share cache, at least around the part that's most expensive (running the container and reading the result).
* If we don't then the cache for a "local" function call and "remote" one would be entirely separate, which won't work long run

Don't think that's worth blocking on here though necessarily.

core/modfunc.go
sipsma: Wasn't missing, it's initialized from the "base cache spec" here: https://github.com/sipsma/dagger/blob/3e867fcf3b5f7d02f029e523b6e019605a3bd8aa/dagql/objects.go#L617-L617

The idea being that there's the base cache spec (which has the default recipe-based digest and other stuff set), then `GetCacheConfig` is called in case any dynamic updates are needed.

There is some spaghetti here though because we handle the `per-session` case only in `GetCacheConfig` , which I'm guessing is what threw you off. It might make sense to just put `PerSession` as an option in the base cache spec rather than require it use a custom cache config func.

core/telemetry.go
```diff
@@ -71,6 +71,11 @@ func AroundFunc(
 		attribute.String(telemetry.DagCallAttr, callAttr),
 	}
 
+	// HACK: probably some more elegant way to do make Module.call passthrough
+	// if base == "Module" && id.Field() == "call" {
+	// 	attrs = append(attrs, attribute.Bool(telemetry.UIPassthroughAttr, true))
+	// }
```
sipsma, line 77: noting to cleanup before merge (not sure if it should be left in and cleaned up or just rm'd)

core/modtypes.go
jedevc: > I almost wonder if we should just be sending the ID as part of `Module.call` and then using that here? I understand the distinction between the `Typed` and `AnyResult` return values but it's extremely confusing at first glance.

1. Yeah, agreed, there's another reason here - I also wanted to avoid needing to pass `dagql.Result`s to `CollectCoreIDs` here: https://github.com/dagger/dagger/blob/0e2d278f5c6ef7e3232d03fbc373dd3902165044/core/hook.go#L223-L249
	
	For all the input values we get, I need a way to get all the core IDs as part of them - there's no associated result for those, they're just inputs. So I started by adjusting `CollectCoreIDs` to just take a `dagql.Typed`.
	
	But because `ModuleObject.CollectCoreIDs` calls `ConvertFromSDKResult`, we now need an associated ID for them. Which I don't think we're guaranteed to always have? I'll be honest, a lot of this is slipping, I can undo as much of this as I can.

2. To pass the ID along, this makes this API almost impossible to call outside of the scale-out scenario. I was kind of hoping to have this be a useful low-level plumbing call as well, @TomChv wanted something that looked like this to be able to call functions from a dynamically loaded module for benchmarking.

   Testing this also just is very fiddly and fragile, since you need to construct the IDs to send, but I can brainstorm a bit.

core/schema/module.go
jedevc: > But the thing is `Callable.Call` still returns an `AnyResult` where we've already figured all that out, so why not just pick up the value from there?

Mm, yeah, I guess this is where the changes I made to `ConvertFromSDKResult` are less good. Because we pass a `nil` ID in, we don't get a `Result`, it's `nil`.

core/modtypes.go
TomChv: Yeah during the offsite I asked Justin if it was possible to dynamically load and call a module from another module.
Because for benchmark, we dynamically get a dev container of the dagger engine given one or multiple version.

Right now I'm using shell as a workaround to inject a container as argument, for example: `benchmark $(github.com/dagger/dagger@v0.19.0 | dev)` but that's not ideal at all, because I cannot send multiple containers easily nor express that with code inside the module.

Ideally I would like to do: `dag.ModuleSource("github.com/dagger/dagger@v0.19.0").Call("xx")`

sipsma:
After more testing we ended up deciding to scale-out based on checks rather than individual calls (for now), which means we won't need `Module.call` yet (just using `Module.check` instead essentially), so closing out for now in lieu of https://github.com/dagger/dagger/pull/11415

PR #11272, opened 2025-10-22
refactor: break up ID.Append

vito:
This was already getting a bit unwieldy as we've been gradually adding
bits and bobs to the protocol, and it was also used to represent all
forms of modification.

Refactored into an `IDOpts` pattern, and added `ID.With(...IDOpts)` for
applying selective modifications to an ID, rather than having to
re-`Append` from its receiver.

Signed-off-by: Alex Suraci <suraci.alex@gmail.com>

### DIFF
--- a/core/schema/secret.go
+++ b/core/schema/secret.go
@@ -166,18 +166,14 @@ func (s *secretSchema) setSecret(
 		accessor,
 	)
 
-	currentID := dagql.CurrentID(ctx)
-	callID := currentID.Receiver().Append(
-		currentID.Type().ToAST(),
-		currentID.Field(),
-		currentID.View(),
-		currentID.Module(),
-		0,
-		dgst,
-		call.NewArgument("name", call.NewLiteralString(args.Name), false),
-		// hide plaintext in the returned ID, we instead rely on the
-		// digest of the ID for uniqueness+identity
-		call.NewArgument("plaintext", call.NewLiteralString("***"), false),
+	callID := dagql.CurrentID(ctx).With(
+		call.WithArgs(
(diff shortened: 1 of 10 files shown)


core/schema/secret.go
```diff
@@ -166,18 +166,14 @@ func (s *secretSchema) setSecret(
...
-		call.NewArgument("name", call.NewLiteralString(args.Name), false),
-		// hide plaintext in the returned ID, we instead rely on the
-		// digest of the ID for uniqueness+identity
-		call.NewArgument("plaintext", call.NewLiteralString("***"), false),
+	callID := dagql.CurrentID(ctx).With(
+		call.WithArgs(
+			call.NewArgument("name", call.NewLiteralString(args.Name), false),
+			// hide plaintext in the returned ID, we instead rely on the
+			// digest of the ID for uniqueness+identity
+			call.NewArgument("plaintext", call.NewLiteralString("***"), false),
```
jedevc, line 174: Was thinking of doing something similar for the `sensitive` value here. Since we only set it to `true` in a couple of places.

vito: Yeah was wondering the same thing, left it as-is for now but happy to tweak.

Alternatively, seems like we could just not add this arg? 🤔

sipsma: > Alternatively, seems like we could just not add this arg? 🤔

That would be ideal but I recall something weird happening when it wasn't there because it's considered a required arg, might be misremembering or out of date though.

vito: OK I tried removing it and it did indeed lead to 'missing required arg' failures - same if I marked it sensitive.

Slowly backing away. :slightly_smiling_face:

PR #11276, opened 2025-10-22
llm: add ctrl+u for uploading changes to llm

vito:
The inverse of <kbd>Ctrl+S</kbd>: this allows the user to make changes out-of-band and then sync them efficiently into the env workspace.

To do this efficiently, the CLI tracks an `mtime` representing the last time that data was synced (in either direction). Then, when the user presses <kbd>Ctrl-U</kbd>, we walk the directory and only upload paths with an `mtime` past the last sync.

## Testing strategy

I added a new test, so now we have some coverage of the LLM CLI interactive flow, which is nice. The not-so-nice part is it is closer to an eval in that it relies on an LLM to run (and implicitly relies on the LLM to behave as expected).

So, I've quarantined these tests out into a separate CI job that runs it along with the evals. Unfortunately I think we have to merge this PR to see it actually run, since it's a new workflow. But it passes on my machine. :innocent: 

To verify yourself:

```sh
dagger call test specific --env-file file://.env --pkg ./cmd/dagger --run CMD/LLM
```

### DIFF
--- a/cmd/dagger/llm.go
+++ b/cmd/dagger/llm.go
@@ -675,6 +675,10 @@ func (s *LLMSession) SyncFromLocal(ctx context.Context) (rerr error) {
 			return err
 		}
 		if d.IsDir() {
+			if d.Name() == ".git" {
+				// don't recurse into .git
+				return filepath.SkipDir
+			}
 			// nothing to do for directories
 			return nil
 		}


cmd/dagger/llm.go
```diff
@@ -620,6 +651,143 @@ func (s *LLMSession) Model(model string) (*LLMSession, error) {
...
+
+	ctx, span := Tracer().Start(ctx, "syncing local changes",
+		telemetry.Reveal())
+	defer telemetry.End(span, func() error { return rerr })
+	slog := slog.SpanLogger(ctx, InstrumentationLibrary)
+	stdio := telemetry.SpanStdio(ctx, InstrumentationLibrary)
+
+	var pathsToUpload []string
+
+	// Look for paths modified since the last sync, and only upload those.
```
sipsma, line 671: I haven't parsed every detail here yet, but at first glance I'm confused why the CLI would need to do this since the local dir sync already handles updating files when their mtime is different (and in fact will repeat that logic).

(have to go to meeting so just leaving as driveby comment)

vito: I'm not familiar with the filesync optimizations and when/how they know to kick in; that's always been deep magic to me, and I don't know how that magic becomes anchored to our API.

The challenge is the CLI isn't the one that initiates the host dir syncing in the first place - that happens on behalf of the module's context dir setup (`//+defaultPath`) where ignore rules are applied, but the CLI doesn't know about those. So AFAICT the only option I have is to do a full `host.directory(".")` sync without any include/exclude, which feels like it'd be unbearably slow in a large enough repo. Does that get handled magically by the filesync `mtime` checks? But even if so, the CLI won't know about ignore paths, so I assume it'd end up uploading a bunch of junk the first time.

(Would totally want a better approach here though, lmk if you have any ideas given that context)

sipsma: Thanks for the context, makes a lot more sense now.

One thing about the existing logic, since the CLI doesn't know about the included paths of the contextual dir, when it does WalkDir("."), it'll end up hitting files that weren't included right? And then if those files happened to have some out-of-band update, they'll now get included and make their way into the LLM's dir in the engine, right?
* e.g. you have two LLM sessions open working on two different subtrees of some directory, but then when changes made to one subtree get written the other LLM now accidentally picks up all those changes

> So AFAICT the only option I have is to do a full host.directory(".") sync without any include/exclude, which feels like it'd be unbearably slow in a large enough repo. Does that get handled magically by the filesync mtime checks?

It could be unbearably slow if the files haven't been synced to the engine previously, yeah.

Very quick overview of the existing filesync mechanics:
1. The engine maintains a single mutable snapshot for each "client machine"
   * there's a `~/.local/state/dagger/stable_client_id` file written by the CLI with an id persisted across sessions and shared by all clients, the mutable snapshot is keyed by that stable id
1. That mutable snapshot is essentially a lazily filled in representation of the entire client filesystem, all the way up to `/` in theory
1. When you filesync paths into the engine, the client sends the engine file+dir metadata for all the paths. The engine compares that metadata against what it currently has in its mutable snapshot. If it sees a file it doesn't already have or that a file has changed contents (based on mtime+size as a heuristic), it asks the client for the contents of the file and writes it to the mutable snapshot
   * there's a bunch of careful handling of concurrent syncs to ensure consistency without globally locking the whole snapshot, etc.
1. To create an immutable snapshot from a single sync, the engine just has to do a copy of the included paths from the mutable snapshot to a separate immutable one. The engine also keeps track of all the content digest stuff along the way of this whole process which allows that immutable snapshot to have that content digested cache key.
   * The copy is a bit inefficient since it dupes data. There's ideas on improving it with either hardlinks and/or overlay but both have complications that have made it not worth it yet.

(I guess that wasn't "very quick" but too late since I typed it all on accident 😅)

---

Overall it doesn't seem worth blocking merge on any of this, probably better to improve iteratively as needed. But I would say that if there was somehow a way for the code here in the CLI to know the include/exclude paths, that would be very helpful regardless of the approach here.
* At minimum that could help avoid the weird case I mentioned at the beginning (if that's real)
* Probably could also enable a lot of this code to go away and just rely on the engine filesync mechanics

vito: Thanks for the details, sounds neat!

> One thing about the existing logic, since the CLI doesn't know about the included paths of the contextual dir, when it does WalkDir("."), it'll end up hitting files that weren't included right? And then if those files happened to have some out-of-band update, they'll now get included and make their way into the LLM's dir in the engine, right?

Yeah that's true. I think that's OK for a first pass, since without this change it's pretty painful to get changes in to the LLM for out-of-band fixes, and if some extra junk comes along for the ride that's not a huge deal. I'll probably run into that now and then and be reminded to fix it, and maybe that'll become easier as more of our roadmap lands (like maybe #11158 affects this in some way).

PR #11277, opened 2025-10-23
Lazy Runtime loading

TomChv:
**Summary**

Update moduleSource loading to no longer force load the runtime container if it's not required.
Instead, we load the module runtime just before actually calling it so it's much more efficient.

:warning: **ONLY sdks that implements`ModuleTypes` interface from #10584 benefits from this optimization**

Effects:
- If a module A with a dependency on B & C only use the module B in a function, we will load the module B runtime just in time when the query calls B.
- We no longer load the runtime when executing `dagger functions` or `--help`
- The `moduleRuntime` traces are no longer part of `initializing module`, they are now wrap inside the function call (this can be improve later).

**Trace for dagger functions (using the typescript SDK)**
Screenshot 2025-10-23 at 15 30 10

**Trace for `dagger call` (using the TypeScript SDK)**
Screenshot 2025-10-23 at 15 39 02

**Benchmark**

**NOTE**: It's normal that `call/XXX` takes more times now since the module loading happens at this stage but it's expect, the important things to look at is the overall cmd duration.

For Typescript with default setup

Screenshot 2025-10-23 at 16 52 28

We can see a 50% improvement for functions which is what we expect since we no longer build the runtime when simply exploring functions.

For Golang with default setup

Screenshot 2025-10-23 at 16 59 04

We also see around 50% improvement for dagger functions.

### DIFF
--- a/core/integration/module_go_test.go
+++ b/core/integration/module_go_test.go
@@ -1482,7 +1482,7 @@ func New() *Minimal {
 	require.JSONEq(t, `{"minimal":{"config":"{\"a\":1}"}}`, out)
 }
 
-// this is no longer allowed, but verify the SDK errors out
+// this is no longer allowed, but verify the Engine errors out
 func (GoSuite) TestExtendCore(ctx context.Context, t *testctx.T) {
 	moreContents := `package dagger
 
@@ -1509,7 +1509,10 @@ func (c *Container) Echo(ctx context.Context, msg string) (string, error) {
 		require.Error(t, err)
 		require.NoError(t, c.Close())
 		t.Log(logs.String())
-		require.Regexp(t, "cannot define methods on objects from outside this module", logs.String())
+
+		// With lazy module loading, the error is no longer thrown by the SDK but directly by the engine
+		// when evaluating the query against the engine GQL schema.
+		require.Contains(t, logs.String(), `Cannot query field \"echo\" on type \"Container\"`)
 	})
 
 	t.Run("in same mod name", func(ctx context.Context, t *testctx.T) {
@@ -1526,7 +1529,9 @@ func (c *Container) Echo(ctx context.Context, msg string) (string, error) {
 		require.Error(t, err)
 		require.NoError(t, c.Close())
 		t.Log(logs.String())
-		require.Regexp(t, "cannot define methods on objects from outside this module", logs.String())
+		// With lazy module loading, the error is no longer thrown by the SDK but directly by the engine
+		// when evaluating the query against the engine GQL schema.
+		require.Contains(t, logs.String(), `type "Container" is already defined by module "daggercore"`)
 	})
 }
 
--- a/core/modfunc.go
+++ b/core/modfunc.go
@@ -39,9 +39,8 @@ const MaxFunctionCacheTTLSeconds = 7 * 24 * 60 * 60 // 1 week
 const MinFunctionCacheTTLSeconds = 1
 
 type ModuleFunction struct {
-	mod     *Module
-	objDef  *ObjectTypeDef // may be nil for special functions like the module definition function call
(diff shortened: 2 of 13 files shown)


TomChv:
It seems this PR may introduce a flack, I can see that python has some issue sometime loading the runtime: https://dagger.cloud/dagger/traces/eb30b8548deddef5d13713a5770eaaee?span=ae93500b9f9fab9a

But it doesn't happen every time, I'm not sure what's the root cause :/

TomChv:
> Also, do we have integ tests involving modules that require an SSH auth sock in order to build the runtime container? If not, that's something to cover here for sure. That could go wrong in the scenario of a module not getting loaded until it's called by another one and then the SSH sock being hard to fine since it's not getting loaded directly by the main caller anymore.

I think we already have these tests in [`TestCustomSDK`](https://github.com/dagger/dagger/blob/44c9087a050e3fe157e8e63f6ac6b5191de10e8a/core/integration/module_test.go#L2145), the SDK ref is a ssh socket with auth I think.

Example of trace: https://dagger.cloud/dagger/traces/62676db598ce20aee4def3a279d3af98?listen=c985e71641da923f&listen=65b56c0229742827&listen=41db3604a93b284c&listen=93bc7db4d1c07b2b&listen=7fa216c090d97759&listen=da10c32dd6a3ce72&listen=99015f526cf8322b

I think because we load the `moduleSource` to get the contextDirectory of that module source, we don't need the socket when loading the runtime because we already have the source we need.
For the dependencies it the same, we still load all the module source when (without the runtime) when initializing dependencies.

> The flake you're seeing is a bit concerning, probably worth making sure it's new and not pre-existing.

I never saw these tests fail before, I agree that it's quite concerning and it only happens in Python.

EDIT: I tried multiple local test run and also reproing the test cases that fails to directly run the binary but I couldn't find a way to repro it.
Do you have a tip on how I could track that? The issue barely make sense honestly

I don't think it's worth blocking the PR for that, specially since we will update the Python SDK to also support lazy loading which should fix that flack

sipsma:
> I never saw these tests fail before, I agree that it's quite concerning and it only happens in Python.
> 
> EDIT: I tried multiple local test run and also reproing the test cases that fails to directly run the binary but I couldn't find a way to repro it. Do you have a tip on how I could track that? The issue barely make sense honestly
> 
> I don't think it's worth blocking the PR for that, specially since we will update the Python SDK to also support lazy loading which should fix that flack

I just saw [a very similar error in a totally different PR](https://dagger.cloud/dagger/traces/15d3285a163e03f4bf91e67d1c2d4db7?listen=13511cae4b6fcd8a&listen=e49b285a5dd5fb60&listen=9ce5aa6e5507f4c6), so yeah don't think it's related to your change here most likely

sipsma (approved):

vito:
Not sure if it helps, but last time I ran into that error (`Field "withDirectory" argument "directory" of type "DirectoryID!" is required, but it was not provided`) it was from a code path in my PR doing a `dagql.Select` without setting a `View`. That arg was renamed from `directory` to `source` in v0.19.0, so something is making that call with an outdated (or unset?) engine version as its View.

Maybe this could be something like a Python module having a v0.18.x `dagger.json` and that causing the lazy runtime loading to see an old version of the schema?

In other words, we need to make sure the lazy runtime loading always internally uses the SDK runtime module's engine version View, and not the View of the function being invoked?

I _think_ @eunomie faced something like this in the `moduleTypes` PR #10584 

Or it's a total red herring; if it's a flake and not consistently failing that's REALLY weird.

TomChv:
I rebased my PR on main (including function caching).
I'll merge when the CI is green :D (except maybe the generate check if it's a flacky fail)

shykes:
🥳🥳🥳🥳🥳

PR #11294, opened 2025-10-24
fix: rotate gitlab readonly token

grouville:
On the gitlab free tier, they only last 1 year. As before, it's a read-only. CI started breaking starting today


sipsma (approved):

sipsma:
@grouville there is something wrong w/ the check-release-dry-run step right now on all PRs, not this PR's fault. I'll go ahead and merge this. Thanks for the fix!

PR #11295, opened 2025-10-24
dagop: never use potentially stale client metadata

sipsma:
We see quite a few flakes around missing sessions in the midst of dagop. It's always when integ tests are running in parallel and a lot that are sharing a step will hit the same error.

I think it may be a timing thing where the client metadata stored in an op is cached in dagql, then the unlazy happens later with timing such that the stored client metadata is okay at first but becomes invalid later due to the session it was from disconnecting.

Instead, I think we should just always use the client metadata of whoever is invoking the operation.

This is much simpler as easy to think through now-a-days since we deleted edge merging, no possible concerns with multiple clients mixing data anymore.

---

After further investigation, looking into the buildkit solver code it seems like even though edge merging is gone, there's still "merging" of states based on the vtx digest, so the session iterator can still return data from a client that goes away.

Pushed an additional commit that handles that scenario too.

Also added there an integ test that consistently repro's the `failed to update rootfs` error, which no longer happens after the fixes in this commit.

### DIFF
--- a/.changes/unreleased/Fixed-20251105-152032.yaml
+++ b/.changes/unreleased/Fixed-20251105-152032.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: 'Fix "failed to update rootfs: session not found" errors caused by incorrect
+  use of stale client metadata.'
+time: 2025-11-05T15:20:32.548933582-08:00
+custom:
+  Author: sipsma
+  PR: "11295"


sipsma:
Hm, the error did happen once again after a while: https://dagger.cloud/dagger/traces/e27ec6d8141c1fcda38dfbf9fd0c9ca5?listen=4695200ea8decd83&listen=e33051783f271d68&showHidden=e33051783f271d68&showHidden=b9d946cb9613a850

Looking into the buildkit solver code it seems like even though edge merging is gone, there's still "merging" of states based on the vtx digest, so the session iterator can still return data from a client that goes away.

Think this is resolvable but will take some more effort and increase the blast radius, so will save this for the next release.

cc @andrestone, sorry this will be delayed a bit 😕

EDIT: pushed an additional commit to handle that now!

internal/buildkit/solver/jobs.go
sipsma, line 1: I removed a lot of unused code from the solver (related to edge merging, which we removed previously), just to help me read through the code as I was working on all this.

engine/server/session.go
```diff
@@ -714,7 +715,12 @@ func (srv *Server) clientFromIDs(sessID, clientID string) (*daggerClient, error)
 	defer srv.daggerSessionsMu.RUnlock()
 	sess, ok := srv.daggerSessions[sessID]
 	if !ok {
-		return nil, fmt.Errorf("session %q not found", sessID)
+		// This error can happen due to per-LLB-vertex deduplication in the buildkit solver,
+		// where for instance the first client cancels and closes its session while others
+		// are waiting on the result. In this case its safe to retry the operation again with
+		// the still connected client metadata.
+		err := flightcontrol.RetryableError{Err: fmt.Errorf("session %q not found", sessID)}
```
alexcb, line 722: nice find/change!

PR #11297, opened 2025-10-24
LLM CLI: add auto-compacting

vito:
Adds two pretty low-level APIs to `LLM` for clearing either system prompts or user messages. The compaction dance uses both: first we swap out the system prompt, since the compacting agent is single-purpose, and then we swap out the user messages with the prompt returned by the compacting agent.

The compacting agent is implemented in the CLI at the moment. An earlier iteration had this directly implemented by the API (`LLM.compact`) but this setup gives us a bit more flexibility without requiring us to maintain a compacting prompt in the core API.

### DIFF
--- a/cmd/dagger/llm.go
+++ b/cmd/dagger/llm.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/charmbracelet/bubbles/key"
@@ -77,6 +78,9 @@ type LLMSession struct {
 
 	plumbingCtx  context.Context
 	plumbingSpan trace.Span
+
+	autoCompact  bool
+	autoCompactL *sync.Mutex
 }
 
 func NewLLMSession(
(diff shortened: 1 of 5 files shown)


cmd/dagger/llm.go
```diff
@@ -197,6 +205,12 @@ func (s *LLMSession) WithPrompt(ctx context.Context, input string) (*LLMSession,
 		if !hasMore {
 			break
 		}
+
+		// Check if we need to compact in-between steps
```
grouville, line 227: It seems, from reading the code, that we call `autoCompact` while `HasPrompt` could still true, so `Compact + WithoutMessageHistory()` would wipe the tool-output user message that the next `Step` is supposed to send no ?

Basically, we seem to have chosen to allow compaction in-between steps, which could be a destroying operation ?

What was the impact on the agent’s ability from your experience?

To test, I dropped the compaction threshold to 0.1 (even 0.0 but this led to an infinite loop ahah): it seemed fine-ish on small vibe feeling tasks

vito: Well - you have to interrupt to auto-compact at some point, otherwise the agent will just keep going and going until it thinks it's done, and the point of auto-compacting is to keep it from blowing out the context window while it's working away.

> Basically, we seem to have chosen to allow compaction in-between steps, which could be a destroying operation ?

The tool results are fed to the compacting agent, since that works by just appending the compacting prompt to the message history so far. So they're not really dropped; they're fed in with the rest of the history, and then the entire history is reset to the summarized prompt.

> What was the impact on the agent’s ability from your experience?

It's been decent; compacting will always be lossy, and there have been moments where I wanted to disable it, but overall it's helped a lot for long-running sessions, and acts as a nice defensive mechanism for when the agent gets caught down a bad path and needs to take a step back.

PR #11298, opened 2025-10-25
llm: ReadLogs queries by span ID

vito:
* avoids the need to track state in MCP
* allows LLM to query by span ID after-the-fact
* support an LLM running N tools in parallel and querying the logs for any of them, not just the last one

A small change splitting out from #11264

### DIFF
--- a/core/mcp.go
+++ b/core/mcp.go
@@ -86,8 +86,6 @@ type MCP struct {
 	lastResult dagql.Typed
 	// Indicates that the model has returned
 	returned bool
-	// History of spans that have logs that we can read
-	loggedSpans []trace.SpanID
 	// Saved objects by ID (Foo#123)
 	objsByID map[string]contextualBinding
 	// Auto incrementing number per-type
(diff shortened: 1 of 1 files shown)


grouville (approved):
LGTM ✅ 

#### Non blocking question

When testing this with the **prompt mode** (https://dagger.cloud/dagger/traces/7041785a17e13814f0e4e246c578cfa3), it looks like _ReadLogs_ no longer falls back to the last recorded span.

From what I can tell, the schema now makes `span` [required](https://github.com/dagger/dagger/blob/c3f66babde18b4802e843ce99c884d409b8330ec/core/mcp.go#L1566), and `readLogsTool` doesn’t use the most recent one anymore.

In my trace above, when the log output isn’t truncated (≤ llmLogsLastLines), limitLines seems to never print the [helper banner that carries the spanID](https://github.com/dagger/dagger/blob/c3f66babde18b4802e843ce99c884d409b8330ec/core/mcp.go#L2345), thus, for the short output of **echo out**, there’s no span value available for the LLM to use, leading to the failure when the LLM tries to use the tool

I lack the context on whether it is expected / an acceptable tradeoff ? (seems to be, just curious for that context)

vito:
@grouville Yeah I think that's OK - the intent of `ReadLogs` is just to support viewing logs that were truncated. There's no point supporting it for things that are already in the context window (like the `ReadLogs(<SpanID>)` header itself).

Thanks for reviewing! :bow:

PR #11306, opened 2025-10-27
test: dir.changes should not include common files

alexcb:


alexcb:
quick reproduction: `./hack/build && ./hack/with-dev go test -count=1 -v -run TestChangeset/TestChangeset/test_common_files_dont_get_added ./core/integration/`

The 

```
		require.Equal(t, []string{"file2.txt"}, entries)
```

test fails with:

```
        	Error:      	Not equal: 
        	            	expected: []string{"file2.txt"}
        	            	actual  : []string{"file2.txt", "file3.txt"}
```

it seems wrong that "file3.txt" is also being added -- especially when it doesn't show up under the `changes.AddedPaths(ctx)` results.

PR #11308, opened 2025-10-28
engine: add new `engine clients` API

matipan:
This PR adds a new engine API to fetch the currently active client IDs on the engine:
```sh
$ dagger core engine clients
▶ connect 2.8s
✔ loading type definitions 0.9s

✔ engine: Engine! 0.0s
✔ .clients: [String!]! 0.0s

mhh6v25ipu1z2zd62vql98uan
```

We also report the Client ID to Dagger Cloud when provisioning remote engines.

### DIFF
--- a/docs/docs-graphql/schema.graphqls
+++ b/docs/docs-graphql/schema.graphqls
@@ -1881,6 +1881,9 @@ scalar DirectoryID
 
 """The Dagger engine configuration and state"""
 type Engine {
+  """The list of connected client IDs"""
+  clients: [String!]!
+
   """A unique identifier for this Engine."""
   id: EngineID!
 
--- a/docs/static/api/reference/index.html
+++ b/docs/static/api/reference/index.html
@@ -7418,6 +7418,10 @@
                     </thead>
                     <tbody>
                       <tr>
+                        <td data-property-name=""><a class="property-name" id="Engine-clients" href="#Engine-clients"><code>clients</code></a> - <span class="property-type"><a href="#definition-String"><code>[String!]!</code></a></span> </td>
+                        <td> The list of connected client IDs </td>
+                      </tr>
+                      <tr>
                         <td data-property-name=""><a class="property-name" id="Engine-id" href="#Engine-id"><code>id</code></a> - <span class="property-type"><a href="#definition-EngineID"><code>EngineID!</code></a></span> </td>
                         <td> A unique identifier for this Engine. </td>
                       </tr>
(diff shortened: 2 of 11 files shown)


internal/cloud/client.go
```diff
@@ -99,27 +98,22 @@ type SerializableCertificate struct {
...
 	OrgID          string                   `json:"org_id,omitempty"`
 	UserID         string                   `json:"user_id,omitempty"`
-	Size           string                   `json:"size,omitempty"`
-	TTL            string                   `json:"ttl,omitempty"`
-	TTLDuration    time.Duration            `json:"ttl_duration,omitempty"`
 	URL            string                   `json:"url,omitempty"`
 	CertSerialized *SerializableCertificate `json:"cert,omitempty"`
-	Module         string                   `json:"module,omitempty"`
-	Function       string                   `json:"function,omitempty"`
-	ExecCmd        []string                 `json:"exec_cmd,omitempty"`
```
eunomie, line 122: I'm not really familiar with this code, but does that mean those fields (`Module`, `Function`, `ExecCmd`) were never read? As well as `EngineRequest`?

core/schema/engine.go
```diff
@@ -19,6 +19,11 @@ func (s *engineSchema) Install(srv *dagql.Server) {
 			Doc("The Dagger engine container configuration and state"),
 	}.Install(srv)
 
+	dagql.Fields[*core.Engine]{
+		dagql.Func("clients", s.clients).
+			Doc("The list of connected client IDs"),
+	}.Install(srv)
```
jedevc, line 25: WDYT about instead returning a list of sessions? And then allowing getting clients from that?

Then they're grouped :tada: (and we can easily allow getting the main client as well)

sipsma: I agree that session ID feels a little bit more straightforward for some reason but since there's a one-to-one mapping of sessionID->main client caller ID, it's ultimately arbitrary

sipsma (approved):

internal/cloud/client.go
matipan: They were read. In fact those are the fields that matter the most. I moved them to a separate EngineRequest struct and EngineSpec composes on top of that. The rest of the fileds I deleted them because they were never used and will likely not be used in the near future.

matipan:
TestCLI|TestEngine seems to be a flake. I just ran it locally on this branch using a PARC engine and it worked:

https://dagger.cloud/dagger/traces/cfb42d853b29f81b74ebe92776564ab3?listen=e1da096044cd7e91&listen=cdb2b5bdadcb5561&listen=c62d64f60f9ed3c6#37e9463e413d7469

PR #11311, opened 2025-10-28
PHP: Fix simple errors

charjr:
This only touches 5 php files, 1 config file, and the rest are just regenerated docs.

Fixes 6 issues detected during static analysis.

Each issue fixed as a separate commit, but since they're fairly small changes anyway you can probably just review it as a whole.

### DIFF
--- a/docs/static/reference/php/Dagger/Client/AbstractScalar.html
+++ b/docs/static/reference/php/Dagger/Client/AbstractScalar.html
@@ -152,8 +152,8 @@
             <div id="method-details">
                     <div class="method-item">
                     <h3 id="method___construct">
-        <div class="location">at line 11</div>
-        <code>                    
+        <div class="location">at line 9</div>
+        <code>final                    
     <strong>__construct</strong>(string $value)
         </code>
     </h3>
@@ -186,7 +186,7 @@
             </div>
                     <div class="method-item">
                     <h3 id="method_getValue">
-        <div class="location">at line 15</div>
+        <div class="location">at line 14</div>
         <code>                    string
     <strong>getValue</strong>()
         </code>
@@ -219,7 +219,7 @@
             </div>
                     <div class="method-item">
                     <h3 id="method___toString">
-        <div class="location">at line 20</div>
+        <div class="location">at line 19</div>
         <code>                    string
     <strong>__toString</strong>()
         </code>
@@ -252,8 +252,8 @@
             </div>
                     <div class="method-item">
                     <h3 id="method_from">
-        <div class="location">at line 25</div>
-        <code>        static            <a href="../../Dagger/Client/AbstractScalar.html"><abbr title="Dagger\Client\AbstractScalar">AbstractScalar</abbr></a>
+        <div class="location">at line 24</div>
+        <code>final        static            <a href="../../Dagger/Client/AbstractScalar.html"><abbr title="Dagger\Client\AbstractScalar">AbstractScalar</abbr></a>
     <strong>from</strong>(string $value)
         </code>
     </h3>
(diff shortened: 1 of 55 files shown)


sdk/php/src/Client/AbstractScalar.php
```diff
@@ -4,12 +4,11 @@
...
 namespace Dagger\Client;
 
-use Stringable;
-
-abstract readonly class AbstractScalar implements Stringable
+abstract readonly class AbstractScalar implements \Stringable
 {
-    public function __construct(private string $value)
-    {
+    final public function __construct(
```
charjr, line 9: In this instance, `final` prevents any inheriting class from overriding the method.

This is desirable, if AbstractScalars could override this method we would not be able to depend on it to reconstruct them later.

sdk/php/src/Command/SchemaGeneratorCommand.php
```diff
@@ -13,14 +13,6 @@
 #[AsCommand('dagger:schema')]
 class SchemaGeneratorCommand extends Command
 {
-    private const WRITE_DIR =
```
charjr, line 16: This constant is completely unused here, it looks like the class was perhaps created by duplicating `CodeGenCommand` and they forgot to remove it.

sdk/php/src/ValueObject/Argument.php
```diff
@@ -92,13 +92,7 @@ private static function getType(ReflectionParameter $parameter): ListOfType | Ty
...
     {
         if ($parameter->isDefaultValueAvailable()) {
-            $betterReflection = BetterReflectionParameter
-                ::createFromClassNameAndMethod(
-                    $parameter->getDeclaringClass()->getName(),
-                    $parameter->getDeclaringFunction()->getName(),
-                    $parameter->getName(),
-                );
-            $default = $betterReflection->getDefaultValue();
+            $default = $parameter->getDefaultValue();
```
charjr, line 95: BetterReflection does not [and will not ](https://github.com/Roave/BetterReflection/issues/1504#issuecomment-2840392615) support default values of objects.

So we must rely on PHP's built-in reflection which actually works out simpler anyway. 🎉

sdk/php/src/ValueObject/DaggerObject.php
```diff
@@ -41,7 +41,7 @@ public static function fromReflection(\ReflectionClass $class): self
         $description = (current($class
             ->getAttributes(Attribute\Doc::class)) ?: null)
             ?->newInstance()
-            ?->description
+            ->description
```
charjr, line 44: `?->` is a null safe operator. What I didn't realise is that it _short circuits_ (similar to conditional `&&`) if the left hand side is null, so the second null safe operator is unnecessary.

docs/static/reference/php/Dagger/Client/AbstractScalar.html
```diff
@@ -152,8 +152,8 @@ <h2>Details</h2>
             <div id="method-details">
                     <div class="method-item">
                     <h3 id="method___construct">
-        <div class="location">at line 11</div>
-        <code>                    
+        <div class="location">at line 9</div>
```
vito, line 155: Sorry if this has been asked already, but is there a flag to disable line numbers in the generated docs? Doesn't feel very useful to me, vs. the insane amount of churn it introduces

charjr: Not to worry, it's never been asked of me at least.

I checked the code for Doctum (the library generating the code). [There is no config for line numbers](https://github.com/code-lts/doctum/blob/40df854fd30212c7dc8cbeb070fdf0679d400732/src/Doctum.php#L156-L199).

There's nothing I can do to fix it quickly in this PR.

In a separate PR: Swapping libraries may be the best solution, I'd be quite tempted to try [phpDocumentor](https://docs.phpdoc.org/) which seems to be a more mature library.

PR #11315, opened 2025-10-28
add filesync telemetry for root paths

marcosnils:
adds OTel spans for filesync uploads. This providers better visibility
into what's effectively being uploaded to the context. It only reports
root paths given that in most cases this will otherwise produce a very
large amount of spans which will put a lot of pressure on the tracing
backend.

We're also collecting the amount of written bytes as well as a the total
duration for each root path

screenshot on how this looks in the TUI: 

image

 
### DIFF
--- a/engine/filesync/localfs.go
+++ b/engine/filesync/localfs.go
@@ -26,6 +26,7 @@ import (
 	"github.com/tonistiigi/fsutil/types"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sys/unix"
@@ -219,11 +220,13 @@ func (local *localFS) Sync( //nolint:gocyclo
 			mu:      sync.Mutex{},
 		}
 
-		rps, loaded := rootPathSpans.LoadOrStore(rootPath, dirSpan)
-		if !loaded {
+		rps, _ := rootPathSpans.LoadOrStore(rootPath, dirSpan)
+		rps.(*rootPathSpan).mu.Lock()
+		if rps.(*rootPathSpan).span == nil {
 			_, span := Tracer(egCtx).Start(egCtx, rootPath)
 			rps.(*rootPathSpan).span = span
 		}
+		rps.(*rootPathSpan).mu.Unlock()
 
 		return rps.(*rootPathSpan)
 	}
@@ -231,7 +234,6 @@ func (local *localFS) Sync( //nolint:gocyclo
 	defer func() {
 		rootPathSpans.Range(func(_, value any) bool {
 			dirSpan := value.(*rootPathSpan)
-			dirSpan.span.SetAttributes(attribute.Int64(telemetry.FilesyncWrittenBytes, dirSpan.writtenBytes))
 			dirSpan.span.End(trace.WithTimestamp(dirSpan.maxStop))
 			return true
 		})
@@ -306,21 +308,32 @@ func (local *localFS) Sync( //nolint:gocyclo
 					// we see and then change the span end time before returning.
 					rootPathSpan := getRootPath(path)
 
-					var err error
-					appliedChange, written, err := local.WriteFile(ctx, kind, path, upperStat, remote)
+					appliedChange, written, werr := local.WriteFile(ctx, kind, path, upperStat, remote)
+
 					writeEnd := time.Now()
+					m := telemetry.Meter(ctx, "dagger.io/filesync")
+					fsMetric, err := m.Int64Gauge(telemetry.FilesyncWrittenBytes)
+					if err != nil {
+						return err
+					}
+					attrs := []attribute.KeyValue{
+						attribute.String(telemetry.MetricsTraceIDAttr, rootPathSpan.span.SpanContext().TraceID().String()),
+						attribute.String(telemetry.MetricsSpanIDAttr, rootPathSpan.span.SpanContext().SpanID().String()),
+					}
 					rootPathSpan.mu.Lock()
 					rootPathSpan.writtenBytes += written
+					fsMetric.Record(ctx, written, metric.WithAttributes(attrs...))
 					// only track the max(end) of all the files within a root path
 					if writeEnd.After(rootPathSpan.maxStop) {
 						rootPathSpan.maxStop = writeEnd
 					}
-					if err != nil {
+					if werr != nil {
 						rootPathSpan.span.SetStatus(codes.Error, err.Error())
 					}
 					rootPathSpan.mu.Unlock()
-					if err != nil {
-						return err
+
+					if werr != nil {
+						return werr
 					}
 
 					cachedResultsMu.Lock()
--- a/dagql/dagui/spans.go
+++ b/dagql/dagui/spans.go
@@ -191,8 +191,6 @@ type SpanSnapshot struct {
 	ChildCount int  `json:",omitempty"`
 	HasLogs    bool `json:",omitempty"`
 
-	FilesyncWrittenBytes int64 `json:",omitempty"`
-
 	ExtraAttributes map[string]json.RawMessage `json:",omitempty"`
 }
 
@@ -280,9 +278,6 @@ func (snapshot *SpanSnapshot) ProcessAttribute(name string, val any) {
 
 	case telemetry.ContentTypeAttr:
 		snapshot.ContentType = val.(string)
-
-	case telemetry.FilesyncWrittenBytes:
-		snapshot.FilesyncWrittenBytes = val.(int64)
 
 	case "rpc.service":
 		// encapsulate these by default; we only maybe want to see these if their
--- a/dagql/idtui/frontend.go
+++ b/dagql/idtui/frontend.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"maps"
 	"os"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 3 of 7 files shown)


engine/filesync/localfs.go
```diff
@@ -260,10 +302,29 @@ func (local *localFS) Sync( //nolint:gocyclo
...
+					// since we can't really know when a root path is done as
+					// all files are synced in parallel, we track everything
+					// we see and then change the span end time before returning.
+					rootPathSpan := getRootPath(path)
+
+					var err error
+					writeStart := time.Now()
+					appliedChange, written, err := local.WriteFile(ctx, kind, path, upperStat, remote)
+					writeEnd := time.Since(writeStart)
+					rootPathSpansMu.Lock()
```
sipsma: nit: feel like it'd be worth adding a separate mutex to `rootPathSpan` so we can lock just that individual object here rather than creating lock contention across all the paths (which should only be needed for the shared map of spans afaict). Might matter in particularly gnarly cases with lots of files being synced.

engine/filesync/localfs.go
```diff
@@ -260,10 +302,29 @@ func (local *localFS) Sync( //nolint:gocyclo
...
+					rootPathSpan.writtenBytes += written
+					// keep the max since multiple files are uploaded in parallel
+					if writeEnd > rootPathSpan.maxWriteDuration {
+						rootPathSpan.maxWriteDuration += writeEnd
```
sipsma: kinda confused, if the duration is longer than the previous max we set the new max to the old max plus this duration? is it supposed to be `maxWriteDuration = writeEnd`?

also, if we're using this to set the trace's end timestamp, shouldn't we just be saving the max `time.Now()` value as called right after `local.WriteFile` returns?

engine/filesync/localfs.go
```diff
@@ -196,6 +199,45 @@ func (local *localFS) Sync( //nolint:gocyclo
...
+		span             trace.Span
+		start            time.Time
+		writtenBytes     int64
+		maxWriteDuration time.Duration
+	}
+	rootPathSpans := make(map[string]*rootPathSpan)
+	var rootPathSpansMu sync.Mutex
+
+	getRootPath := func(path string) *rootPathSpan {
+		rootPath := strings.Split(path, string(os.PathSeparator))[0]
```
sipsma, line 214: I don't think `os.PathSeparator` will work as intended here cause it will always be linux-based since this code runs in the engine.

engine/filesync/localfs.go
```diff
@@ -196,6 +199,45 @@ func (local *localFS) Sync( //nolint:gocyclo
...
 
+	// Track spans per top-level directory
+	type rootPathSpan struct {
+		span             trace.Span
+		start            time.Time
+		writtenBytes     int64
+		maxWriteDuration time.Duration
+	}
+	rootPathSpans := make(map[string]*rootPathSpan)
+	var rootPathSpansMu sync.Mutex
```
sipsma: nit: the use case here seems to match exactly what the go docs say about when a `sync.Map` is more performant than a map+mutex:
> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate [Mutex](https://pkg.go.dev/sync#Mutex) or [RWMutex](https://pkg.go.dev/sync#RWMutex).

so probably worth switching to that

sipsma:
> It only reports root paths given that in most cases this will otherwise produce a very large amount of spans which will put a lot of pressure on the tracing backend.

Makes total sense, but wonder if we should also impose a limit on the max number of root paths to avoid problems if a user is trying to sync dirs w/ `MAX_NUM_DIRENTS_IN_A_SINGLE_DIR` files

engine/filesync/localfs.go
marcosnils: > kinda confused, if the duration is longer than the previous max we set the new max to the old max plus this duration? is it supposed to be `maxWriteDuration = writeEnd`?

correct, that was the intention here. Strange since I typoed the assignment operator but still the TUI was showing the correct results :crazy:

> also, if we're using this to set the trace's end timestamp, shouldn't we just be saving the max `time.Now()` value as called right after `local.WriteFile` returns?

just realized this wasn't entirely right,  what I should track instead is the startTime and max(stopTime) across all the files within the same root path so I can accurately report the span duration. I'll make the change to reflect that.

engine/filesync/localfs.go
marcosnils: > nit: feel like it'd be worth adding a separate mutex to `rootPathSpan` so we can lock just that individual object here rather than creating lock contention across all the paths (which should only be needed for the shared map of spans afaict). Might matter in particularly gnarly cases with lots of files being synced.

SGTM.  Initially I thought about putting  this logic within the `cachedResultsMu` lock below which already holds a lock for all paths but it felt a bit dirty to be using some other lock for this. I guess one lock per root path seems like the most sensible thing to do here.

engine/filesync/localfs.go
marcosnils: right.. I assume for windows users this path might arrive as `\\` or `/`, correct? https://cs.opensource.google/go/go/+/master:src/internal/filepathlite/path_windows.go;l=19-20

seems like the  best way to proceed here is to always do the `\\` | `/` check? Mostly checking if by any chance you know we're doing a similar handling somwhere else.

engine/filesync/localfs.go
marcosnils: > nit: the use case here seems to match exactly what the go docs say about when a `sync.Map` is more performant than a map+mutex:

 lol.. was about to use it but decided to go with map + mutex because generics.. Let me change that :muscle:

engine/filesync/localfs.go
```diff
@@ -196,6 +199,45 @@ func (local *localFS) Sync( //nolint:gocyclo
...
+		}
+		rootPathSpans[rootPath] = dirSpan
+		return dirSpan
+	}
+
+	defer func() {
+		// ensure all spans are ended when done
+		rootPathSpansMu.Lock()
+		for _, dirSpan := range rootPathSpans {
+			dirSpan.span.SetAttributes(attribute.Int64(telemetry.FilesyncWrittenBytes, dirSpan.writtenBytes))
```
vito: Any reason this couldn't be a metric? Feels more natural since we already have CPU usage, network, memory, etc. - and a well-beaten path for tying metrics to specific spans. Would be nice to be able to see it live-update, too, for large transfers, instead of only appearing at the end.

e.g.:

https://github.com/dagger/dagger/blob/b22f7708d4e0115dda35152ed6a0ba4faa760cd8/core/llm_anthropic.go#L77-L104

vito (commented):
Excited for this!

engine/filesync/localfs.go
sipsma: I'm not sure there's something that does exactly what you need yet, but there are some utils that'd help out in this file: https://github.com/sipsma/dagger/blob/916c67f121f7ce52da2c5c24d2bb928380ff92e9/engine/client/pathutil/pathutil.go#L39-L39 (normalizePath and GetDrive specifically)

engine/filesync/localfs.go
sipsma: Yeah hopefully the stdlib adds a generics-based sync.Map someday

engine/filesync/localfs.go
marcosnils: @sipsma  justin mentioned about the client already normalizing all the paths to linux/style once they're received by the engine. Is it possible that at this stage all the paths we see from the  `lower` and `upper` FSs are already normalized at this stange and it's safe to use `os.Fileseparator` here?

sipsma: Ah yeah I guess the client does `filepath.ToSlash` before sending the paths, so that might not need extra handling. I'm not 100% sure but I think windows clients may still be including the drive prefix (e.g. `C://`), so you may need to strip that out if present still

marcosnils: > I think windows clients may still be including the drive prefix (e.g. `C://`), so you may need to strip that out if present still

 gotcha, thx. I'll try to make a quick test in windows to see how that looks like :pray:

marcosnils: image

 works in windows.  After checking with Erik really quick we found out that we're already stripping the drive here: https://github.com/dagger/dagger/blob/485c6da454c862b874393d18b3c1d371d1130108/engine/filesync/filesyncer.go?plain=1#L120-L122

engine/filesync/localfs.go
marcosnils: image

:heavy_check_mark: . 

This will  colorize "B" and "kB" in green, "MB" in yellow and anything greater than that in red.

engine/filesync/localfs.go
```diff
@@ -260,10 +302,39 @@ func (local *localFS) Sync( //nolint:gocyclo
...
+					appliedChange, written, werr := local.WriteFile(ctx, kind, path, upperStat, remote)
+
+					writeEnd := time.Now()
+					m := telemetry.Meter(ctx, "dagger.io/filesync")
+					fsMetric, err := m.Int64Gauge(telemetry.FilesyncWrittenBytes)
 					if err != nil {
 						return err
 					}
+					attrs := []attribute.KeyValue{
+						attribute.String(telemetry.MetricsTraceIDAttr, rootPathSpan.span.SpanContext().TraceID().String()),
```
sipsma, line 320: tests showing a panic on this line: https://dagger.cloud/dagger/traces/9f78297444137d035bbc5a52bf06493b?span=10f0f3a825d52517

(just in case you didn't see it already since finding the panic in the engine requires some spelunking around the telemetry)

shykes:
🍿🍿🍿🍿🍿🍿🍿

PR #11319, opened 2025-10-28
fix(ts): recover property access defaults for core enums

grouville:
#### Context

In #11267, Tom changed `resolveParameterDefaultValue` to lean entirely on the TypeScript `TypeChecker`. That dropped the old object-based lookup we had for `ModuleTypes`. Due to that, some types cannot be resolved anymore (imported libs) and are marked as `unknown`
  
**_The unintended consequence_**

This is generally not a problem, as, at runtime, everything is properly resolved.

However, my deprecation PR needs to have a correct static resolving of all enum defaults (to distinguish _omittable_ mandatory args [args with defaults values, variadics], with the others) to early return when users put pragmas on unwanted fields

####  What this change does

Change `resolveParameterDefaultValue` to properly unwrap _property-access defaults_:
- ask the checker for the constant
- fallback to the enum member declaration
- fallback again to the literal type the checker inferred

#### How to manually test:

```shell
$ cd sdk/typescript
$ bun run --bun mocha src/module/introspector/test/scan.spec.ts --grep coreEnumDefault
```

**_Update 10/28/2025_**

It's not a complete feature as it requires also knowing the types of the SDK lib. I tried implementing how to resolve the lib and add it to the compiledprogram, which is seen as `any` atm, but failed.

At least, we do not drop the default value anymore now. 

**_Update 11/12/2025_**

I managed to resolve the SDK library inside the introspector by mounting a minimal synthetic `@dagger.io/dagger` package (containing `index.ts` and `client.gen.ts`) under `node_modules` for both `ModuleTypes` and the TS runtimes.

With this in place, the _TypeChecker_ no longer sees SDK imports as `any`. The existing resolution pipeline in `resolveParameterDefaultValue`:
- first asks the checker for a constant value,
- then walks enum member declarations,
- and finally falls back to literal types inferred by the checker,

now runs against real types instead of _any_. As a result, enum defaults such as `NetworkProtocol.Udp` are correctly recovered as `"UDP"` in docs, and the unresolved warnings that were emitted during introspection disappear.

_Important_

This change is intentionally scoped: it doesn’t alter how the SDK itself is generated or how modules behave at runtime. It only ensures that the TypeScript Program used for introspection has access to the same type information that user code imports.

The integration remains hermetic, cache‑friendly, and free of external dependencies, whilst making sure that defaults that reference SDK types are no longer dropped

### DIFF
--- a/sdk/typescript/runtime/tsutils/module/core.d.ts
+++ b/sdk/typescript/runtime/tsutils/module/core.d.ts
@@ -1,16 +1,28 @@
-export type ConnectOpts = Record<string, unknown>
-export type CallbackFct = (...args: any[]) => any
+export type ConnectOpts = {
+  Workdir?: string
+  LogOutput?: unknown
+}
 
-export function connection(cb: CallbackFct, opts?: ConnectOpts): any
-export function connect(opts?: ConnectOpts): any
+export type CallbackFct = (client: unknown) => Promise<void>
+
+export function connection(
+  cb: () => Promise<void>,
+  opts?: ConnectOpts,
+): Promise<void>
+export function connect(cb: CallbackFct, opts?: ConnectOpts): Promise<void>
 export class Context {}
 
-export function func(): MethodDecorator
-export function check(): MethodDecorator
-export function argument(): ParameterDecorator
-export function object(): ClassDecorator
-export function field(): PropertyDecorator
-export function enumType(): ClassDecorator
-export function entrypoint(): MethodDecorator
+export type ArgumentOptions = {
+  defaultPath?: string
+  ignore?: string[]
+}
 
-export function getTracer(name?: string): any
+export function func(alias?: string): MethodDecorator
+export function check(): MethodDecorator
+export function argument(opts?: ArgumentOptions): ParameterDecorator
+export function object(): ClassDecorator
+export function field(alias?: string): PropertyDecorator
+export function enumType(): ClassDecorator
+export function entrypoint(files: string[]): Promise<void>
+
+export function getTracer(name?: string): unknown


sdk/typescript/src/module/introspector/typescript_module/ast.ts
```diff
@@ -407,6 +407,65 @@ export class AST {
...
+      ) {
+        const literal = t as ts.LiteralType
+        if (literal.value !== undefined) {
+          return literal.value as string | number | bigint
+        }
+
+        const intrinsic = literal as unknown as { intrinsicName?: string }
+        if (intrinsic.intrinsicName !== undefined) {
+          return intrinsic.intrinsicName as string
+        }
```
TomChv, line 487: Can you explain this part? Is it because it's the enum name?

sdk/typescript/src/module/introspector/typescript_module/ast.ts
```diff
@@ -407,6 +407,65 @@ export class AST {
...
+      t: ts.Type,
+    ): string | number | boolean | bigint | undefined => {
+      if (t.flags & ts.TypeFlags.BooleanLiteral) {
+        const intrinsic = t as unknown as { intrinsicName?: string }
+        if (intrinsic.intrinsicName === "true") {
+          return true
+        }
+        if (intrinsic.intrinsicName === "false") {
+          return false
+        }
```
TomChv:
```suggestion
        // Handle boolean value
        if (intrinsic.intrinsicName === "true" || intrinsinc.intrinsincName === "false") {
          return intrinsic.intrinsicName === "true" ? true : false
        }
```

sdk/typescript/src/module/introspector/typescript_module/ast.ts
```diff
@@ -407,6 +407,65 @@ export class AST {
...
+      return undefined
+    }
+
+    if (type.isUnion()) {
+      for (const subtype of type.types) {
+        const literal = resolveLiteral(subtype)
+        if (literal !== undefined) {
+          return literal
+        }
+      }
```
TomChv, line 506: I'm not sure about that part, does this mean we can handle union like `"foo" | "bar" ????`
I'm not sure the typedef introspection handle that

sdk/typescript/src/module/introspector/typescript_module/ast.ts
grouville: TS sometimes gives us literals without a value, only through _intrinsicName_. That happens for booleans (and other cases when the literal is carried through a symbol), not because of the enum member name. You can see it by running:

```shell
  bun --eval 'import ts from "./sdk/typescript/node_modules/typescript";
  // source code
  const source = `const yes = true;`;
  const fileName = "inline.ts";
  
  // parse it
  const sourceFile = ts.createSourceFile(fileName, source, ts.ScriptTarget.ES2022, true);
  const program = ts.createProgram({ rootNames: [fileName], options: { target: ts.ScriptTarget.ES2022 } });
 const checker = program.getTypeChecker()

// extract type and log it
  const stmt = sourceFile.statements[0]          // const yes = true;
  const decl = stmt.declarationList.declarations[0]
  const literal = checker.getTypeAtLocation(decl.initializer!)
  console.log({ value: literal.value ?? null, intrinsicName: literal.intrinsicName });
  '
```

Output:
>   { value: null, intrinsicName: 'true' }

So, when `literal.value` is `undefined`, we fall back to `intrinsicName` to keep the actual literal (here `"true"`). Otherwise we’d drop the default when it’s referenced through another symbol.

sdk/typescript/src/module/introspector/typescript_module/ast.ts
grouville: Let’s take a concrete example:
```ts
const defaults = { proto: NetworkProtocol.Udp }
function use(proto: NetworkProtocol = defaults.proto) {}
```

When you write the enum default inside an object like that, the checker tells us `defaults.proto` has the type `NetworkProtocol.Tcp | NetworkProtocol.Udp`.

Our `type.isUnion()` branch simply walks those union members so we can recover the literal default ("UDP"). 

Without it we’d get `undefined` and drop the default.

You can see exactly what the checker returns by running (from the repo root):
  
  ```shell
    bun --eval '
  const ts=require("./sdk/typescript/node_modules/typescript");
  const source=`import { NetworkProtocol } from "./dummy";
  const defaults = { proto: NetworkProtocol.Udp };
  function use(proto: NetworkProtocol = defaults.proto) {}
  `;
  const enumSource=`export enum NetworkProtocol {Tcp = "TCP", Udp = "UDP"}`;
  const opts={target:ts.ScriptTarget.ES2022,module:ts.ModuleKind.CommonJS};
  const file="file.ts";
  const host=ts.createCompilerHost(opts,true);
  const getSource=host.getSourceFile.bind(host);
  host.getSourceFile=(name,lang)=>{
    if(name===file) return ts.createSourceFile(name,source,lang,true);
    if(name.endsWith("dummy.ts")) return ts.createSourceFile(name,enumSource,lang,true);
    return getSource(name,lang);
  };
  const fileExists=host.fileExists.bind(host);
  host.fileExists=name=>name===file||name.endsWith("dummy.ts")||fileExists(name);
  const readFile=host.readFile.bind(host);
  host.readFile=name=>{
    if(name===file) return source;
    if(name.endsWith("dummy.ts")) return enumSource;
    return readFile(name);
  };
  const program=ts.createProgram([file],opts,host);
  const checker=program.getTypeChecker();
  const init=program.getSourceFile(file).statements.find(ts.isFunctionDeclaration).parameters[0].initializer;
  const type=checker.getTypeAtLocation(init);
  console.log("typeToString",checker.typeToString(type));
  console.log("isUnion",type.isUnion());
  type.isUnion() && type.types.forEach(sub =>
    console.log("member", checker.typeToString(sub), "value", sub.value, "intrinsic", sub.intrinsicName)
  );
  console.log("value",type.value,"intrinsic",type.intrinsicName);
  '
  ```
  
The output:
```shell
typeToString NetworkProtocol
isUnion true
# our loop finds the correct value
member NetworkProtocol.Tcp value tcp intrinsic undefined
member NetworkProtocol.Udp value udp intrinsic undefined
# what we would otherwise get --> undefined, that we would drop
value undefined intrinsic undefined
```

The union wrapper has no literal information (value undefined), but each member carries the actual literal (tcp/udp). This is what makes the `coreEnumDefault` test pass

sdk/typescript/src/module/introspector/typescript_module/ast.ts
grouville: I updated with a switch case, more readable, with the requested grouping and an added comment 🙏

sdk/typescript/src/module/introspector/typescript_module/ast.ts
TomChv: Make sense!

sdk/typescript/src/module/introspector/typescript_module/ast.ts
TomChv: Okay nice!

sdk/typescript/src/module/introspector/typescript_module/ast.ts
TomChv: Nice!

sdk/typescript/runtime/introspector.go
```diff
@@ -37,9 +38,17 @@ func (i *Introspector) AsEntrypoint(
 
 	clientBindings *dagger.File,
 ) *dagger.Container {
+	// Synthesize a minimal @dagger.io/dagger package so TS can resolve the bare import.
+	sdkPkg := dag.Directory().
+		WithFile("client.gen.ts", clientBindings).
+		WithNewFile("index.ts", tsutils.StaticBundleIndexTS)
```
TomChv: The static bundle also imports from `./telemetry.ts`, but it should never be introspected so can you double check if that work if you also do `import { getTracer } from "@dagger.io/dagger/telemetry"`

sdk/typescript/runtime/runtime_bun.go
```diff
@@ -88,6 +88,8 @@ func (b *BunRuntime) SetupContainer(ctx context.Context) (*dagger.Container, err
 	// Merge all the generated files together and setup an entrypoint command.
 	return runtimeWithDep.ctr.
 		WithMountedDirectory(GenDir, sdkLibrary).
+		// Make @dagger.io/dagger resolvable for ts-introspector (it doesn't read tsconfig paths).
```
TomChv, line 91: That's curious we need that but since that the same client, I guess it works.

I'm wondering (for a follow up) if we could remove the `sdk` dir locally and just put the bundle version in the `node_modules` so we don't duplicate that logic here.

Normally your addition shouldn't impact performances since its mounted.

TomChv (approved):
LGTM, I left 2 questions then you should merge that PR :)

TomChv:
Lint CI is failing, otherwise LGTM: https://dagger.cloud/dagger/traces/c2677f9376ff384d3bfef4308cb8f090

PR #11320, opened 2025-10-29
disable unneeded and extremely expensive content hashing

sipsma:
dag-op is using FileOp as a base, which enables content hashing of all inputs by default.

This is extremely expensive in terms of CPU time, disk reads, memory, etc. It also serves no purpose anymore (that I can see) because we now ensure all inputs to dag-ops have their dagql ID digest mixed into the op's cache key.

This had been getting MUCH worse recently because we've been porting more and more ops to dag-op. I think in particular `withDirectory` was especially painful because it resulted in every input arg getting content hashed over and over.

Specifically, the previous behavior was actually `O(n^2)` when `withDirectory` was chained together. This is because each `withDirectory` step would get unlazied (requiring creation of the merged filesystem on disk) and content hashed in its entirety. It ends up O(n^2) because in each step you do the same work as previous step plus some new work for the newly added filesystem (so it's a triangle-number sort of pattern).

---

As one data point, comparing a fully uncached (fresh engine) build of the dagger engine:
- [Before](https://dagger.cloud/dagger/traces/afa69a53192cb13876979262ccbe0f0e?listen=0af5d48ade8f84b9&listen=4775d348767b8e71) - 7m33s
- [After](https://dagger.cloud/dagger/traces/852ca7cf7163add6ec6c43d35fee546b) - 4m51s

### DIFF
--- a/.changes/unreleased/Fixed-20251029-010919.yaml
+++ b/.changes/unreleased/Fixed-20251029-010919.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Optimize numerous core APIs to be less CPU+memory intensive.
+time: 2025-10-29T01:09:19.498795983-07:00
+custom:
+  Author: sipsma
+  PR: "11320"


sipsma:
Tests seems much faster too:
1. [Here](https://github.com/dagger/dagger/actions/runs/18895352099?pr=11320)
2. [main](https://github.com/dagger/dagger/actions/runs/18892242125)

There is one legit looking failure in `TestDirectory/TestDirCaching`. Will figure out how to fix.

EDIT: pushed a fix as a separate commit. Seems like it allows us to retain the caching optimizations tested in `TestDirCaching` while retaining the perf benefits.

vito (approved):
must-go-faster

PR #11326, opened 2025-10-29
feat: allow contextual git to access uncommited changes

jedevc:
Previously, a contextual git directory was loading a *bare* repository (essentially only including the `.git` path). However, we now want access to `GitRepository.uncommitted`, so that we can get the changes that haven't yet been committed to the repo.

To do this, *technically* all we need to do is remove the `Include` filter. However, the performance of this essentially means we'd upload everything, which is hugely expensive. Anything in the gitignore isn't actually going to be part of the changeset, so we can skip uploading it.

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -5278,6 +5278,45 @@ func (ModuleSuite) TestContextGitRemoteDep(ctx context.Context, t *testctx.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


shykes:
Perfect timing! 🥳

shykes:
@jedevc OK to merge?

jedevc:
Yup good to go!

PR #11327, opened 2025-10-29
reduce mutex contention during disk usage calculation 

sipsma:
https://github.com/dagger/dagger/pull/11336 reduced mutex contention quite a bit around various boltdb operations, but after that the next biggest source of contention as seen in pprof profiles is around locks used to acquire refs during disk pruning.

This change results in reusing already known cache records rather than trying to reacquire them, saving us from a lot of extra locks/unlocks. This matters during disk pruning since there can be an enormous number of refs to work through.

### DIFF
--- a/cmd/engine/debug.go
+++ b/cmd/engine/debug.go
@@ -35,10 +35,16 @@ func setupDebugHandlers(addr string) error {
 	m.Handle("/debug/pprof/symbol", http.HandlerFunc(pprof.Symbol))
 	m.Handle("/debug/pprof/trace", http.HandlerFunc(pprof.Trace))
 	m.Handle("/debug/pprof/heap", pprof.Handler("heap"))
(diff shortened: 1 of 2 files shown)


sipsma:
Interestingly, while this makes a huge difference for me locally (22m to run `TestModule` uncached vs. 33m before), CI doesn't seem to be impacted. It's highly possible this has to do with disk bandwidth since an enormous chunk of the lock contention was around bbolt write locks held during syncs.

So I think this is promising for engines running on disks that aren't top-of-the-line, but needs a little more investigation.

EDIT: I scoped this down quite a bit to a fairly safe isolated change that reduces the locking during disk usage calculation. The benefit is less but probably worth approaching the improvements in this area iteratively.

sipsma:
Moving this one to draft for a little bit. CI has been having some hangs pretty consistently that are worth investigating before merge.

EDIT: I think the hangs were unrelated issues that have since been addressed on main. This should be good to go

PR #11329, opened 2025-10-29
fix: host().directory().File is no longer caching

### LINKED ISSUE #10992: `Directory.File()` no longer cached based on file contents
### What is the issue?

Whenever a `File` is derived from a `Directory`, _any_ directory changes taint all caches using the `File` - even if the `File` has the exact same content and metadata.

~~When using `_EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-container://...`, caches are invalidated more compared to the variable not being set.~~

#### Example project

Take a directory with this layout:

```
.
├── dagger.json
└── somefile
```

where `somefile` is empty and the contents of `dagger.json` are:

```json
{"name": "dagger-foo"}
```

#### Dagger invocation

We create a container, add `somefile` to it and execute a dummy command:

```bash
$ dagger <<'EOF'
container | from alpine:latest | with-file /work/somefile $(host | directory . | file somefile) | with-exec sleep 5
EOF
```

#### Expectation

After the first time, everything in the function chain can be cached, as long as the contents of `somefile` stay the same.

#### Actual behavior

The cache for `with-exec sleep 5` is invalidated every time _any_ file in the directory is modified ~~when `_EXPERIMENTAL_DAGGER_RUNNER_HOST` is set to a `docker-container://` type~~.

### Dagger version

dagger v0.18.17 (docker-container://dagger-engine-custom) linux/amd64

### Steps to reproduce

1. Run dagger ([cloud trace](https://dagger.cloud/sleiner-v/traces/8bc8a68dcb4d2d6fb6c22f3a6f5ed5aa)):
    ```bash
    dagger <<'EOF'
    container | from alpine:latest | with-file /work/somefile $(host | directory . | file somefile) | with-exec sleep 5
    EOF
    ```
    As expected, the execution of `sleep 5` is not cached.
2. Run the exact same command again ([cloud trace](https://dagger.cloud/sleiner-v/traces/fafd2c48d55531a0b19003b829a0211a))
    Also, as expected, everything is cached now.
3. Change something unrelated in the working directory:
    ```bash
    touch anotherfile
    ```
4. Run the exact same dagger command again ([cloud trace](https://dagger.cloud/sleiner-v/traces/7f28e2ea45b67f23af54170ea4c341dd#efdfb4753eb9df53)). As opposed to expectations, the `sleep 5` execution is **not cached**.

### Log output

Cloud traces are linked for each reproduction step.

alexcb:
This restores the caching behaviour of directory.file.
Fixes #10992

### DIFF
--- a/core/contenthash.go
+++ b/core/contenthash.go
@@ -55,7 +55,37 @@ func GetContentHashFromDirectory(
 	if err != nil {
 		return "", fmt.Errorf("failed to marshal state: %w", err)
 	}
-	dgst, err := GetContentHashFromDef(ctx, bk, def.ToPB(), dirInst.Self().Dir)
+	dirPath := dirInst.Self().Dir
+	if !strings.HasSuffix(dirPath, "/") {
+		// omit directory name from the hash
+		dirPath += "/"
+	}
+	dgst, err := GetContentHashFromDef(ctx, bk, def.ToPB(), dirPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to get content hash: %w", err)
+	}
+
+	return dgst, nil
+}
+
+func GetContentHashFromFile(
+	ctx context.Context,
+	bk *buildkit.Client,
+	fileInst dagql.ObjectResult[*File],
+) (digest.Digest, error) {
+	if fileInst.Self() == nil {
+		return "", fmt.Errorf("file instance is nil")
+	}
+
+	st, err := fileInst.Self().State()
+	if err != nil {
+		return "", fmt.Errorf("failed to get state: %w", err)
+	}
+	def, err := st.Marshal(ctx, llb.Platform(fileInst.Self().Platform.Spec()))
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal state: %w", err)
+	}
+	dgst, err := GetContentHashFromDef(ctx, bk, def.ToPB(), fileInst.Self().File)
 	if err != nil {
 		return "", fmt.Errorf("failed to get content hash: %w", err)
 	}
@@ -118,7 +148,9 @@ func GetContentHashFromDef(
 		)
 		defer telemetry.End(span, func() error { return rerr })
 
-		dgst, err := bkcontenthash.Checksum(ctx, ref, subdir, bkcontenthash.ChecksumOpts{}, nil)
+		dgst, err := bkcontenthash.Checksum(ctx, ref, subdir, bkcontenthash.ChecksumOpts{
+			FollowLinks: true,
+		}, nil)
 		if err != nil {
 			return "", fmt.Errorf("failed to checksum ref: %w", err)
 		}
--- a/core/schema/directory.go
+++ b/core/schema/directory.go
@@ -14,6 +14,7 @@ import (
 	"github.com/dagger/dagger/core"
 	"github.com/dagger/dagger/dagql"
 	"github.com/dagger/dagger/internal/buildkit/client/llb"
+	"github.com/dagger/dagger/util/hashutil"
 	"github.com/moby/patternmatcher/ignorefile"
 	"github.com/opencontainers/go-digest"
 )
@@ -80,7 +81,7 @@ func (s *directorySchema) Install(srv *dagql.Server) {
 				The format of the digest is not guaranteed to be stable between releases of Dagger.
 				It is guaranteed to be stable between invocations of the same Dagger engine.`,
 			),
-		dagql.Func("file", s.file).
+		dagql.NodeFunc("file", s.file).
 			Doc(`Retrieve a file at the given path.`).
 			Args(
 				dagql.Arg("path").Doc(`Location of the file to retrieve (e.g., "README.md").`),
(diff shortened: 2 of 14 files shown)


core/schema/directory.go
```diff
@@ -580,10 +581,56 @@ func (s *directorySchema) digest(ctx context.Context, parent *core.Directory, ar
...
+	if err != nil {
+		return inst, err
+	}
+
+	dgst = hashutil.HashStrings(
+		filename,
+		string(dgst),
+	)
+
+	return fileResult.WithObjectDigest(dgst), nil
```
alexcb, line 633: Note that I tried doing something similar to `maintainContentHashing` -- where we only hash when the parent has a overridden content hash; however that causes the `directory.filter()` caching to break.

https://github.com/alexcb/dagger/commit/8885c07ad446c681f5e906f7fe9e7b48de0fa868 has a start to this patch -- but it will need further work in order to prevent breaking `directory.filter()`.

core/contenthash.go
```diff
@@ -118,7 +148,9 @@ func GetContentHashFromDef(
 		)
 		defer telemetry.End(span, func() error { return rerr })
 
-		dgst, err := bkcontenthash.Checksum(ctx, ref, subdir, bkcontenthash.ChecksumOpts{}, nil)
+		dgst, err := bkcontenthash.Checksum(ctx, ref, subdir, bkcontenthash.ChecksumOpts{
+			FollowLinks: true,
```
sipsma, line 152: What's the reasoning for adding this? Symlinks just inherently spook me a bit, so only want to change this setting if it's truly necessary

core/schema/directory.go
```diff
@@ -580,10 +581,56 @@ func (s *directorySchema) digest(ctx context.Context, parent *core.Directory, ar
...
+	}
+	dgst, err := core.GetContentHashFromFile(ctx, bk, fileResult)
+	if err != nil {
+		return inst, err
+	}
+
+	dgst = hashutil.HashStrings(
+		filename,
+		string(dgst),
+	)
```
sipsma, line 631: Suspect there's a good reason, but it's not immediately clear to me why we need to include the filename here too rather than just the `dgst` from the content hash

core/schema/modulesource.go
```diff
@@ -559,6 +559,13 @@ func (s *moduleSourceSchema) localModuleSource(
 	return dagql.NewResultForCurrentID(ctx, localSrc)
 }
 
+// isErrNotExist returns true if the error related to a file not existing
+// During the great dagopification, directory.File() changed from returning ErrNotExist to
+// a generic "no such file or directory" error -- so we need to handle both cases
+func isErrNotExist(err error) bool {
+	return errors.Is(err, os.ErrNotExist) || strings.HasSuffix(err.Error(), "no such file or directory")
+}
```
sipsma: nit: I'd almost rather handle this in `directory.File()` such that it returns an `os.ErrNotExist` in that case. Otherwise we're gonna need this sort of handling of both error types potentially all over the codebase in the long run

sipsma (approved):
Overall LGTM, a couple questions but approving opportunistically in case they all become no-ops

core/contenthash.go
alexcb: It's needed for https://github.com/alexcb/dagger/blob/3cae2f3d93ff2220e88683eddbb2a493d726c9e6/core/integration/directory_test.go#L2381 to pass.

e.g. consider:

```
ln -s target linkname
echo some-data > target
```

The content hash of `linkname` must contain the actual data that it's referencing, e.g. `some-data`.

Without including the contents of the target, a user could then change the data, e.g. to `something-else`, and the content hash would remain the same, and the file would incorrectly return the original `some-data` contents.

core/schema/directory.go
alexcb: Consider two directories:

```
echo data > foo
```

and

```
echo data > bar
```

If you have two directories, and didn't include the filename in the hash, you would get the exact same hash.

You can then end up with two different structs:

```
core.File{
    LLB: same-deduplicated-directory
    File: "foo",
}
```

and


```
core.File{
    LLB: same-deduplicated-directory
    File: "bar",
}
```

Without including the filename in the hash, you end up with a race where sometimes the deduplicated directory contains `foo` and other times it contains `bar`.

If we wanted to have this get deduplicated, we would then have to consistently store the file in the same filename no matter what the `core.File` struct has for the `File` string value. However, if we chose to do this, we wouldn't be able to do any merge ops (since the generic file would have to be renamed to the original value, e.g. foo or bar).

core/schema/modulesource.go
alexcb: I had a try at switching this around, but when I do that the dockerfile tests start failing, specifically the docker compatibility check: https://github.com/alexcb/dagger/blob/3cae2f3d93ff2220e88683eddbb2a493d726c9e6/core/integration/dockerfile_test.go#L299

```
Error "failed to get content hash: /src/.dockerignore: file does not exist" does not contain "/src/.dockerignore: no such file or directory"
```

Do we want to go through the tests and update them to test for the new error string? i.e. does it matter to our end user if the error message changes from `no such file or directory` to `file does not exist`?

alexcb: also sub question: should we be removing the `failed to get content hash: ` prefix when it's an os error?

sipsma: > Do we want to go through the tests and update them to test for the new error string? i.e. does it matter to our end user if the error message changes from no such file or directory to file does not exist?

I'd be fine with that. This is a nit though to be clear, so if it take things way outta scope or otherwise becomes a rabbit hole, feel free to merge it all as it was before.

> also sub question: should we be removing the failed to get content hash: prefix when it's an os error?

I guess I'd lean towards leaving that in and just rely on the error wrapping to check if it's `os.ErrNotExist`, simply because including prefixes like `failed to get content hash` can help track down the exact code path where an error occurred (poor man's stack trace I suppose). But no strong feelings either way.

alexcb: so it turns out there's `os.IsNotExist` which is functionally equivalent to my original hacky function. 

and all I had to do was add in:

```
		if os.IsNotExist(err) {
			return nil, fmt.Errorf("%s: No such file or directory", targetPath)
		}
```

under the new `directory.File` method.

As I was going down the road of converting more calls to use the new `directory.File` implementation, I was finding myself hitting wack-a-mole with updating all these tests. I'm leaning towards undoing my test changes in favour of just maintaining the original error message (but doing it in the file method instead).

PR #11334, opened 2025-10-30
Improve releasing during v0.19.4

sipsma:
Also includes bumping all the dagger.json's to not disable function caching and annotate some functions here and there:
* publish-related functions are cache per session
* I also annotated some of the test functions as cache per session because we often want to re-run the same test repeatedly (i.e. for a flake)
   * this works for now, but wondering if there's a better way of approaching this such as allowing a function with a cache annotation to have that behavior overridden at runtime 🤔

### DIFF
--- a/modules/daggerverse/main.go
+++ b/modules/daggerverse/main.go
@@ -21,6 +21,7 @@ type Daggerverse struct {
 	Repo string
 }
 
+// +cache="session"
 func New(
 	ctx context.Context,
 	// GitHub Personal Access Token which access to dagger/dagger.io repo
@@ -57,6 +58,7 @@ func New(
 }
 
 // Deploy preview environment running Dagger main: dagger call --github-token=env:GITHUB_PAT deploy-preview-with-dagger-main
+// +cache="session"
 func (h *Daggerverse) DeployPreviewWithDaggerMain(
 	ctx context.Context,
 	target string,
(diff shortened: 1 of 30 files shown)


modules/daggerverse/main.go
```diff
@@ -21,6 +21,7 @@ type Daggerverse struct {
 	Repo string
 }
 
+// +cache="session"
```
vito, line 24: Is this one needed? 🤔

sdk/php/dev/dagger.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "php-sdk-dev",
-  "engineVersion": "v0.19.3",
+  "engineVersion": "v0.19.4",
   "sdk": {
     "source": "php"
-  }
+  },
+  "disableDefaultFunctionCaching": true
```
vito: Missed a spot - or is this because the PHP SDK doesn't support the pragmas yet?

sipsma: nope just missed it, will fix!

modules/daggerverse/main.go
sipsma: I thought so because it's making API calls to github, so we are touching the external mutable world.

Possible that a TTL could be preferable but not entirely sure, so we can refine as we go

vito: Ah sorry didn't look close enough and assumed it was just a dumb constructor. LGTM 👍

PR #11336, opened 2025-10-31
disable sync in every boltdb in bk+containerd

sipsma:
Depends on https://github.com/dagger/dagger/pull/11337

Disabling sync in both bk+containerd boltdbs makes an absurd difference in execution time for me locally. The improvement is actually palpable, everything feels much more snappy.

These numbers are all from my laptop
* `TestModule`
   * Before: >30m
   * After: 9m
* Cached `load module` of the CI monolith
   * Before: 9s
   * After: 2s
* Fully uncached engine build
   * Before: >4m
   * After: 2m27s
* Partially cached engine build (i.e. rebuild after code change)
   * Before: ~50s
   * After: 30s

It seems to noticeably help in CI too (some very fast runs [here](https://github.com/dagger/dagger/actions/runs/18962321021?pr=11336)), but perhaps not quite as dramatically (guessing the disks are good there).

Generally speaking, disabling sync should be totally fine since it's not like dagger is supposed to be an ACID-compliant database for your production data. The engine can do a sync when it shuts down, but if it crashes then it's okay if we just have to reset all data on restart.

The reason that sync was so incredibly detrimental to performance is that bbolt holds a lock on every write, which includes the sync. So simple operations like updating a piece of metadata for a cache ref could easily get blocked for multiple seconds given all the contention around that write lock.
* Getting rid of the sync means holding the lock much less and reducing that contention

---

Something relevant I just realized: one (out the four) of the boltdbs actually already was running w/ `NoSync: true`, and if its corrupted on load it's mv'd to a tmp file (as a debug backup) and a new one gets initialized. The end result would be a reset empty cache, even though quite a bit of other state would still be sitting around and inaccessible (at least for a little while til some async gc happens).

So actually this isn't a new problem, we've pretty much already been in this situation and not realized it

---

TODO:
- [x] Handle best-effort sync on graceful engine shutdown (close all DBs, run a sync syscall)
- [x] Decide on erroring out when loading a corrupt db vs. automatically resetting state
   - We will automatically reset the local state in this case. From the perspective that the local cache is not guaranteed to last any amount of time (it's just a cache), automatically resetting state seems reasonable and more convenient for users.
   - We will include this in the published metrics (see cmd/engine/metrics.go) so it's visible other than just the engine logs

### DIFF
--- a/engine/server/server.go
+++ b/engine/server/server.go
@@ -701,7 +701,8 @@ func (srv *Server) Clients() []string {
 	return slices.Collect(maps.Keys(clients))
 }
 
-// TODO: doc, explain difference from close below
+// GracefulStop attempts to close all boltdbs and do a final syncfs since all the DBs
+// run with NoSync=true for performance reasons.
 func (srv *Server) GracefulStop(ctx context.Context) error {
 	var eg errgroup.Group
 	eg.Go(func() error {


sipsma:
CI was definitely faster too, but not as dramatic: https://dagger.cloud/dagger/ci?branch=less-sync&minTimestamp=1761873900516&ref=b74e33fea84f7f1fa831d1d68d29e97bd43f2a11&repo=github.com%2Fdagger%2Fdagger

Part of that is because the previous engine is still being used for the setup phase. But looking at just the `test specific` spans in isolation while they are faster without a doubt, it's not the 3x improvement I saw locally. Guessing our CI infra just has really fast disks that could stand the weight of incessant bbolt syncing better than my poor laptop

engine/server/server.go
```diff
@@ -255,11 +261,53 @@ func NewServer(ctx context.Context, opts *NewServerOpts) (*Server, error) {
...
 	}
+
+	//
+	// setup various buildkit/containerd entities and DBs
+	//
+
+	if err := srv.initBoltDBs(); err != nil {
+		// it's possible for DBs to get corrupted because we run them w/ Sync: false (for performance)
+		// for now, just error out and let users manually reset engine state
+		return nil, fmt.Errorf("failed to initialize databases: %w", err)
```
TomChv: Maybe we can give a hint to the user if this ever happened on how to reset the engine state?
Or maybe a doc page somewhere?[](url)

TomChv:
It definitely speed things up locally (right is main and left is your PR), is still faster to run using `--cloud` but as you say the compute there is just better than our little machines 🤣 

Screenshot 2025-10-31 at 10 46 01

engine/server/server.go
sipsma: Yeah I'm trying to decide right now whether it's better to keep this behavior vs. just automatically resetting state. But agree if I leave this as is will need some docs

tiborvass:
My main concern with this, is disaster recovery logic: what happens when you kill -9 the engine, does/can it always restart cleanly ?

sipsma:
> My main concern with this, is disaster recovery logic: what happens when you kill -9 the engine, does/can it always restart cleanly ?

Yep mention that in the description:
> Something relevant I just realized: one (out the four) of the boltdbs actually already was running w/ NoSync: true, and if its corrupted on load it's mv'd to a tmp file (as a debug backup) and a new one gets initialized. The end result would be a reset empty cache, even though quite a bit of other state would still be sitting around and inaccessible (at least for a little while til some async gc happens).
> 
> So actually this isn't a new problem, we've pretty much already been in this situation and not realized it

Currently trying to decide on whether to keep that sort of logic vs. just erroring out and letting users reset state manually. Talking to infra people in discord

tiborvass:
~@sipsma was `checks.go` inadvertently removed ?~ nvm I thought the `force-pushed` link worked differently. Super surprised by the diff of `force-pushed`.

sipsma:
> @sipsma was `checks.go` inadvertently removed ?

@tiborvass no, I don't see that in the diff here?

tiborvass:
@sipsma My bad, I clicked on the `force-pushed` link

engine/server/server.go
```diff
@@ -592,6 +596,63 @@ func NewServer(ctx context.Context, opts *NewServerOpts) (*Server, error) {
 	return srv, nil
 }
 
+func (srv *Server) initBoltDBs() (err error) {
+	defer func() {
+		if panicErr := recover(); panicErr != nil {
+			err = fmt.Errorf("panic while initializing boltdbs: %v", panicErr)
```
tiborvass, line 634: %w ?

sipsma: `recover()` returns, an `any`, not an `error`, so gotta use %v

tiborvass (approved):

PR #11346, opened 2025-11-03
change deprecated container.build to directory.dockerbuild in docs

marcosnils:
updates docs so that they use directory.dockerbuild instead of the
deprecated container.build method.


docs/current_docs/cookbook/snippets/builds/dockerfile-context/typescript/index.ts
```diff
@@ -1,4 +1,4 @@
-import { dag, Directory, File, object, func } from "@dagger.io/dagger"
+import { dag, Directory, File, object, func } from "@dagger.io/dagger";
```
eunomie: nit/question(non blocking): do we care about `;` at the end of typescript lines?

marcosnils: good catch. Seems like the linter/formatter in my IDE added those. As we're not using them anywhere else I'll just go ahead and remove them.

PR #11349, opened 2025-11-03
core/sdks: enable runtime debug mode

grouville:
#### Context
This adds support for jumping into SDK runtime containers when `sdk.debug` is set in `dagger.json`.

It’s a simple DX loop to debug SDK internals. When enabled, Dagger drops you into the SDK’s container (Go, Python, or TypeScript) before running the user's function.
    
Example: 

```json
{
  "name": "dagger-dev",
  "engineVersion": "v0.19.4",
  "sdk": {
    "source": "go",
    "debug": true
  }
}
```
    
#### Part 2 of 2
- [x]  [#11353](https://github.com/dagger/dagger/pull/11353)￼: adds the debug field + cache-busting
- This PR: uses it to enter SDK runtimes

## How to test

```shell
$ dagger init --sdk=python/go/ts
$ vim dagger.json # add the `debug: true` in the "sdk" key
$ dagger call container-echo ... # any of the functions, you will jump into the runtime
```

### DIFF
--- a/core/sdk/go_sdk.go
+++ b/core/sdk/go_sdk.go
@@ -456,6 +456,12 @@ func (sdk *goSDK) Runtime(
 		return inst, fmt.Errorf("failed to build go runtime binary: %w", err)
 	}
 
+	if cfg := source.Self().SDK; cfg != nil && cfg.Debug {
+		if err := dag.Select(ctx, ctr, &ctr, dagql.Selector{Field: "terminal"}); err != nil {
+			return inst, fmt.Errorf("failed to enable go sdk runtime terminal: %w", err)
+		}
+	}
+
 	return ctr, nil
 }
 
@@ -619,12 +625,6 @@ func (sdk *goSDK) baseWithCodegen(
 
 	if err := dag.Select(ctx, ctr, &ctr, selectors...); err != nil {
 		return ctr, fmt.Errorf("failed to mount introspection json file into go module sdk container codegen: %w", err)
-	}
-
-	if cfg := src.Self().SDK; cfg != nil && cfg.Debug {
-		if err := dag.Select(ctx, ctr, &ctr, dagql.Selector{Field: "terminal"}); err != nil {
-			return ctr, fmt.Errorf("failed to enable go sdk runtime terminal: %w", err)
-		}
 	}
 
 	return ctr, nil
--- a/sdk/typescript/runtime/runtime_bun.go
+++ b/sdk/typescript/runtime/runtime_bun.go
@@ -32,10 +32,6 @@ func NewBunRuntime(
 		}).
 		WithWorkdir(cfg.modulePath())
 
-	if cfg.debug {
-		ctr = ctr.Terminal()
-	}
-
 	return &BunRuntime{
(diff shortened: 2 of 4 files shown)


grouville:
the failing check is a flake

sdk/typescript/runtime/runtime_deno.go
```diff
@@ -32,6 +32,10 @@ func NewDenoRuntime(
 		}).
 		WithWorkdir(cfg.modulePath())
 
+	if cfg.debug {
```
TomChv: I think you want to open a terminal after the container is fully setup no?
So it should be at the end of `SetupContainer` instead?

sdk/typescript/runtime/runtime_bun.go
```diff
@@ -32,6 +32,10 @@ func NewBunRuntime(
 		}).
 		WithWorkdir(cfg.modulePath())
 
+	if cfg.debug {
```
TomChv: I think you want to open a terminal after the container is fully setup no?
So it should be at the end of `SetupContainer` instead?

sdk/typescript/runtime/runtime_node.go
```diff
@@ -35,6 +35,10 @@ func NewNodeRuntime(
 		WithMountedDirectory("/usr/local/lib/node_modules/tsx", sdkSourceDir.Directory("/tsx_module")).
 		WithExec([]string{"ln", "-s", "/usr/local/lib/node_modules/tsx/dist/cli.mjs", "/usr/local/bin/tsx"})
 
+	if cfg.debug {
```
TomChv: I think you want to open a terminal after the container is fully setup no?
So it should be at the end of `SetupContainer` instead?

core/sdk/go_sdk.go
```diff
@@ -621,6 +621,12 @@ func (sdk *goSDK) baseWithCodegen(
 		return ctr, fmt.Errorf("failed to mount introspection json file into go module sdk container codegen: %w", err)
 	}
 
+	if cfg := src.Self().SDK; cfg != nil && cfg.Debug {
+		if err := dag.Select(ctx, ctr, &ctr, dagql.Selector{Field: "terminal"}); err != nil {
+			return ctr, fmt.Errorf("failed to enable go sdk runtime terminal: %w", err)
+		}
+	}
+
```
TomChv: In Python you open the terminal right before returning the terminal, in Typescript you do it before setup the container and in Go you do it right in the middle.

I think it should be consistent, maybe at the end to get the full context?

sdk/typescript/runtime/runtime_deno.go
grouville: Good catch, thank you 🙏

sdk/typescript/runtime/runtime_node.go
grouville: agreed, done 🙏

core/sdk/go_sdk.go
grouville: agreed, good catch 🙏

PR #11350, opened 2025-11-03
fix loading cached contextual dirs

sipsma:
There's a bug w/ function caching in the following scenario:

Say there's 3 functions like A calls B calls C. C uses a context dir arg. Then on a subsequent run A is cache invalidated but B + C are cached. The problem is the contextual dir arg to C was never loaded and A sees it from the cached result and tries to load it but does so from its context, which is different than the original host.

So you get various errors along the lines of not being able to load a path because it's missing.

---

Fix is not incredibly straightforward. Went through a few failed attempts before settling on the following:
1. Add new internal-only `_contextualDirectory` + `_contextualFile` APIs that are used to load contextual dir + file args. These show up in the IDs for such argument values rather than direct `Host.Directory/File`
2. The implementation of those internal APIs accepts a content-based digest for the module being loaded from, which can then be loaded direct from the dagql session cache and used to perform the load with the right context every time

The pro of this approach is that it fixes the problem w/out breaking everything else and/or requiring further large-scale improvements to the dagql caching system that will take a while.

The downsides are:
1. It's pretty ugly+convoluted; tried to make up for this with comments but that can only help so much
2. The telemetry looks a little weird (e.g. you can see `_contextualDirectory` show up)
   * I made up for this by including the human-readable module name as an arg, so users will understand what's going on and where the dir/file are coming from at least.
3. There are different cache keys for a dir/file loaded from the local host as a contextual arg vs. just normally loading the dir/file via the `Host` API
   * In practice the cases where users would benefit from this overlap is quite small (would have to regularly have cases where a contextual arg and Host loaded arg end up being the exact same and plugged into the same operation), so this isn't the end of the world

---

Conceptually, the "real" fix should be that dependencies between results (such as between a module object return and a contextual dir it references) should be tracked and handled such that they form a single "unit" rather than a tangle of IDs that are all independent. We will absolutely need that as we continue down theseus, but that's a big chunk of work, not viable for a quick fix.

I also tried an approach of adding `ModuleSource.contextDirectory` as an API, which ends up looking a little bit nicer. Unfortunately it introduces a new set of conundrums around the fact that `moduleSource` is cached per-client (and has to be for now). I tried dealing with those issues but it proved to be too complex and not worth it for a "quick fix".

### DIFF
--- a/core/integration/module_config_test.go
+++ b/core/integration/module_config_test.go
@@ -1335,10 +1335,6 @@ func getVCSTestCase(t *testctx.T, url string) vcsTestCase {
 }
 
 func testGitModuleRef(tc vcsTestCase, subpath string) string {
-	return testGitModuleRefAtCommit(tc, subpath, tc.gitTestRepoCommit)
-}
-
-func testGitModuleRefAtCommit(tc vcsTestCase, subpath string, commit string) string {
 	url := tc.gitTestRepoRef
 	if subpath != "" {
 		if !strings.HasPrefix(subpath, "/") {
@@ -1346,7 +1342,7 @@ func testGitModuleRefAtCommit(tc vcsTestCase, subpath string, commit string) str
 		}
 		url += subpath
 	}
-	return fmt.Sprintf("%s@%s", url, commit)
+	return fmt.Sprintf("%s@%s", url, tc.gitTestRepoCommit)
 }
 
 func (ConfigSuite) TestDaggerGitRefs(ctx context.Context, t *testctx.T) {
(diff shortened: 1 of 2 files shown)


sipsma:
Also snuck in [Revert "core: re-add content hashing of git and dir module sources"](https://github.com/dagger/dagger/pull/11350/commits/3a7bdd8d355524899d7354fdf674e072101df42d) since it turns out to cause problems in the case of private git repos. I think this was a flake on main already but seems to have been made worse by this PR, so including it here.

PR #11357, opened 2025-11-05
make dev releases more like stable releases

### LINKED ISSUE #11325: Make dev releases more like stable releases
## Problem

We continuously publish dev builds from the `main` branch of Dagger, as usable releases. But the publishing process is slightly different, which requires special cases in the build scripts, and even in the CLI code.

## Solution

Change how we publish nightly builds, to be more like stable releases. This will allow simplifying the release toolchain, *and* the CLI code. It will also remove complicated coupling between the two, thus making the whole system simpler.

Specifically:

- Instead of publishing the engine image at the tag <git_commit>, publish it at the tag <version_string>.
- Remove the concept of 'tag' in the CLI build, and associated tooling. For example `github.com/dagger/dagger/version.tag()` can be removed. And the 'tag' value in the go build as well.
- In the CLI code for auto-downloading the engine, use the version string instead of the 'tag' string.

eunomie:
Fixes #11325 

At the level of the engine, only version is used, no different version + tag anymore.
A specific SDKVersion field so that SDKs keep the existing behaviour.

During publishing, in addition to the actual `main` + git commit (to remove once we ensure everything is ok?) also publish it under the version as defined by the engine.

---
From first commit:

We are publishing dev versions based on `main`.
To align them with the tagged versions, keep only one single way
to identify the engine: Version as comming from the version module.

The version is defined as following:
- if this is a dev version: `v[major].[minor].[patch]-[timestamp]-dev-[digest]`
- if this is an untagged version: `v[major].[minor].[patch]-[timestamp]-[commit]`
- if this is a tagged version: `v[major].[minor].[patch]`

The `engine.Tag` (comming from Version.Tag) was also used to find
the right reference to pull non bundled SDKs like PHP, Java, Elixir.
The rule is:
- if this is a tagged version: use the tag
- otherwise, use the most recent tag existing on `main`

This is now available under `Version.TagOrCommit` and set to
`engine.SDKVersion`.

### DIFF
--- a/releaser/dagger.json
+++ b/releaser/dagger.json
@@ -42,6 +42,10 @@
       "name": "notify",
       "source": "github.com/gerhard/daggerverse/notify@2024-02-13",
       "pin": "d83fad32dc187838c849d4ea8152cc97a46de00a"
+    },
+    {
+      "name": "version",
+      "source": "../version"
     }
   ]
 }
--- a/releaser/main.go
+++ b/releaser/main.go
@@ -183,6 +183,14 @@ func (r *Releaser) Publish(
 	if isPublicRelease {
 		tags = append(tags, "latest")
 	}
+	if !isPublicRelease {
+		engineVersion, err := dag.Version().Version(ctx)
+		if err != nil {
+			artifact.Errors = append(artifact.Errors, dag.Error(err.Error()))
+		} else {
+			tags = append(tags, engineVersion)
+		}
+	}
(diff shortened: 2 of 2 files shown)


shykes:
Thanks for taking this! Sorry I misunderstood when we were discussing it earlier today.

grouville:
If this conflicts with Solomon's PR, would it be useful to rebase on top of it too ?

shykes:
Actually let's pause it until #11398 is merged. i found a workaround so it's not as urgent now. It's mire urgent to finish merging the sdk toolchains for the split 🙏

PR #11358, opened 2025-11-05
add more logging when git operations fail in the engine

marcosnils:

### DIFF
--- a/util/gitutil/cli.go
+++ b/util/gitutil/cli.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"dagger.io/dagger/telemetry"
+	"github.com/dagger/dagger/engine/slog"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 )
@@ -312,6 +313,7 @@ func (cli *GitCLI) Run(ctx context.Context, args ...string) (_ []byte, rerr erro
 			}
 		default:
 		}
+		slog.Error("git command failed", "args", cmd.Args, "error", err, "stderr", errbuf.String())
 		return buf.Bytes(), fmt.Errorf("git error: %w", translateError(err, errbuf.String()))
 	}
 	return buf.Bytes(), nil


util/gitutil/cli.go
```diff
@@ -312,6 +313,7 @@ func (cli *GitCLI) Run(ctx context.Context, args ...string) (_ []byte, rerr erro
 			}
 		default:
 		}
+		slog.Error("git command failed", "args", cmd.Args, "error", err, "stderr", errbuf.String())
```
tiborvass, line 316: Is it unsafe to add it to the `git error` output ?

marcosnils: I don't think any particular sensitive information will be logged there. If you consider the repo URL as somthing "safe", it's already being logged in the output in some other places.

tiborvass: So then wouldn't it be more visible to the user (and thus easier to report to us) to see it in the actual error that's returned, and not in the log files ? I'm saying why not modify the `fmt.Errorf("git error: %w"` line ?

util/gitutil/cli.go
```diff
@@ -312,6 +313,7 @@ func (cli *GitCLI) Run(ctx context.Context, args ...string) (_ []byte, rerr erro
...
 		return buf.Bytes(), fmt.Errorf("git error: %w", translateError(err, errbuf.String()))
```
tiborvass, line 317: Something like this or maybe just cmd.Args

```suggestion
		return buf.Bytes(), fmt.Errorf("git error: %w: args=%v env=%v", translateError(err, errbuf.String()), cmd.Args, cmd.Env)
```

util/gitutil/cli.go
sipsma, line 1: @marcosnils the current code does `stdio := telemetry.SpanStdio(ctx, InstrumentationLibrary)` and sends the git cli output to otel, shouldn't that cover seeing the output already?

marcosnils: good catch Erik. Didn't realize that `stdio` was being created from the span logs. In that case yes, this should have been displayed in the span logs. I'll go ahead and close this PR

PR #11366, opened 2025-11-05
core/git: cache `git.Remote()` metadata in session cache

grouville:
Every `NewGitRepository` (e.g., each `dag.Git(...)` or `GitRepository` call) ran `git ls-remote` with no caching. On my setup this averaged to ~0.7s per call, which can quickly add up to a noticeable latency on big dagger pipelines

This PR adds a **session-level cache** for `ls-remote`: multiple `.Remote()` calls that resolve the same repository/ref now hit the cache within the session.

image

This is the first step towards optimizing our git operations, next steps (in follow-up PRs):
- Re-introduce lazy execution for Git remote discovery (by introducing a `git.__lsRemote` internal method) so we only perform the check when needed along the call path
- Re-evaluate the necessity and placement of `IsRemotePrivate()` to avoid redundant checks

### DIFF
--- a/core/git_remote.go
+++ b/core/git_remote.go
@@ -99,7 +99,7 @@ func (repo *RemoteGitRepository) Remote(ctx context.Context) (result *gitutil.Re
 		// Serialize to JSON so the cache can persist a plain string payload without
 		// requiring gitutil.Remote to satisfy dagql's typing interfaces. The decode
 		// step also hands each repo its own copy, so later tweaks (e.g. setting
-		// repo.Remote.Head) stay scoped to that caller.
+		// repo.Remote.Head) stay scoped to that caller
 		payload, err := json.Marshal(remote)
 		if err != nil {
 			return nil, err
@@ -112,13 +112,6 @@ func (repo *RemoteGitRepository) Remote(ctx context.Context) (result *gitutil.Re
 	}
 
 	slog.Info("loaded git remote metadata", "cache_hit", cacheRes.HitCache(), "cache_key", cacheKey)
-
-	// Auth-related tests expect ls-remote to re-run so that auth failures
-	// surface even if the metadata is cached from a prior success.
-	if cacheRes.HitCache() && (repo.AuthToken.Self() != nil || repo.AuthHeader.Self() != nil || repo.SSHAuthSocket.Self() != nil) {
-		slog.Info("forcing git ls-remote despite cache hit due to HTTP auth token", "cache_key", cacheKey)
-		return repo.runLsRemote(ctx)
-	}
 
 	val := cacheRes.Result()
 	strRes, ok := dagql.UnwrapAs[dagql.Result[dagql.String]](val)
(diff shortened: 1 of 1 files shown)


core/integration/git_test.go
```diff
@@ -1367,6 +1367,30 @@ func (GitSuite) TestIsRemotePublic(ctx context.Context, t *testctx.T) {
 	}
 }
 
+func (GitSuite) TestGitLsRemoteSessionCache(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+	defer func() {
+		require.NoError(t, c.Close())
+	}()
+
+	repo := c.Git("https://github.com/dagger/dagger")
```
tiborvass: I would use a smaller repo. https://github.com/dagger/dagger-test-modules ?

grouville:
`ci_in_ci` failure seems unrelated

core/git_remote.go
```diff
@@ -62,7 +67,132 @@ func (repo *RemoteGitRepository) PBDefinitions(ctx context.Context) ([]*pb.Defin
...
+		repo: repo,
+		Ref:  target,
+	}, nil
+}
+
+// remoteCacheKey mixes only the inputs that influence ls-remote so we avoid hashing
+// the entire repo object (no raw secret payloads or dagql wrapper state). We fold in
+// the remote URL, auth settings (usernames plus hashed secret/known-host metadata),
+// and service bindings (service digest + hostname + sorted aliases).
+func (repo *RemoteGitRepository) remoteCacheKey() string {
```
sipsma: I *think* this can be simplified to just include:
* The `SessionID` (as read from client metadata)
* the URL of the repo

At that point it will be scoped per-session (and have a lifetime limited to the session) and scoped per-URL. AFAIK the various auth-related things don't change the answer you get from `ls-remote` (just whether you get an error or not).

The current implementation actually has a corner case issue in that since it's not scoped to a session, concurrent clients will pick up this cache hit, which then creates a weird situation where we'll never re-run `ls-remote` again until no open client has a reference to it. So we definitely want to mix in `SessionID`, but at that point all the auth stuff should be unneeded too.

core/git_remote.go
```diff
@@ -62,7 +66,83 @@ func (repo *RemoteGitRepository) PBDefinitions(ctx context.Context) ([]*pb.Defin
...
+		repo: repo,
+		Ref:  target,
+	}, nil
+}
+
+// remoteCacheKey scopes cached git remote metadata to the current session and remote URL.
+func (repo *RemoteGitRepository) remoteCacheKey(ctx context.Context) (string, error) {
+	clientMetadata, err := engine.ClientMetadataFromContext(ctx)
+	if err != nil {
+		return "", fmt.Errorf("client metadata: %w", err)
```
tiborvass: I would just return the error as-is here (it's not adding much info compared to https://github.com/dagger/dagger/blob/f76b1d49fbc0fbdfaf9e95b5fe744cebec1e1ea9/engine/opts.go#L101) and add more info (RemoteGitRepository + url) at the caller of remoteCacheKey

tiborvass:
The failures are due to the tests relying on `ls-remote` to check for auth failures, but because the same git server sometimes is used to also test successes, the cache hit fails the test that expects an auth failure.

It is incredibly unlikely that anyone would be doing something similar to our tests in the real world, but an easy fix is to force an ls-remote in the unlikely case where we have a cache hit AND `httpAuthToken` is set. I believe that should fix the relevant failures in the CI.

core/git_remote.go
grouville: After iterations and experiments with Tibor [see comment below](https://github.com/dagger/dagger/pull/11366#issuecomment-3503641972), our test suite (especially `TestGit`) was failing, as it was testing, in a single session, different patterns of how git operations fail after a successful run.

As the `git ls-remote` operation is being used to assess if a repo is public or private, it seems that we need to include the auth sequence as part of the cache key ?

Some real-use pipelines do exist and could be expressed as part of a CI / platform team pipelines. Example: Token rotation -> A job may fetch metadata with an expiring token, revoke it (or switch users), and immediately re-query to confirm the revocation took effect ? Unlikely but possible

grouville:
> The failures are due to the tests relying on `ls-remote` to check for auth failures, but because the same git server sometimes is used to also test successes, the cache hit fails the test that expects an auth failure.
> 
> It is incredibly unlikely that anyone would be doing something similar to our tests in the real world, but an easy fix is to force an ls-remote in the unlikely case where we have a cache hit AND `httpAuthToken` is set. I believe that should fix the relevant failures in the CI.

Answered here: https://github.com/dagger/dagger/pull/11366#discussion_r2506360959

core/git_remote.go
sipsma: Okay that sounds good then, but we still need to mix in the Session ID to the cache key to avoid the problem I mentioned above where the cached result sticks around across sessions and thus is never invalidated until no one references it.

sipsma: Sorry, just saw that the code does that (was only looking at `remoteCacheScope`, saw the session ID is mixed in elsewhere)

PR #11371, opened 2025-11-07
fix nil deference in filesync telemetry

marcosnils:
we're also enabling the nilness linter to catch these issues earlier


matipan (approved):

marcosnils:
merging this since the ci-in-ci tests and the scan is something that seems to be failing often.

PR #11373, opened 2025-11-07
Kill the monolith

shykes:
The time has finally come to break up our CI monolith into cleanly separated [toolchains](#11238). This will make our CI simpler, smaller, and hopefully faster! It will also *finally* allow us to show our own CI as a "golden example" of how to use Dagger. It has been a long road getting to this moment... 😭

## Goals:

1. Faster CI
2. Simpler CI
3. A golden example of Dagger best practices

## LATER

- [ ] Open issue in dagger/dagger fix toolchain config (default values can only be strings)
- [ ] Open issue to allow shadowing core types with a toolchain (for `dagger call engine` instead of `dagger call engine-dev`)
- [ ] Convert `docs/recorder` to a toolchain (wip @eunomie)
- [ ] Translate more modules to `dang` for load speed
- [ ] Remove `.gitignore` from our modules; add a `generate` function; stop generating on-the-fly on go lint (faster!)
- [ ] Ditch `golangci-lint`, we don't need a dumb meta-linter, we have dagger to glue stuff together already. Use `staticcheck`
- [ ] [core] Rename "toolchain configurations" to "toolchain customizations"

## DONE

- [x] Configure docker auth in CI environment
- [x] Check that all modules actually work (harder because of lazy loading) (should be fixed by https://github.com/dagger/dagger/pull/11413)
- [x] Re-enable `bump` in `elixir-sdk` and `typescript-sdk`
- [x] Fix timeouts in engine tests. [example](https://dagger.cloud/dagger/traces/913f83436870e1660b90990e15a58527)
- [x] Harmonize dependency pins for load speed
- [x] Make sure all checks are exposed as such (wip @shykes )
- [x] Fx error in [`typescript-sdk:typecheck`](https://dagger.cloud/dagger/traces/e9a09f6948e82008d02b9c0e569a6f12?listen=dbb7c42d95e69ebd&showHidden=ebf16d60392d6f93)
- [x] Vendor all external module dependencies
- [x] Fix display of toolchain descriptions (in `dagger functions` and `dagger toolchain list`) (done @eunomie)
- [x] Fix error in [`typescript-sdk:test`](https://dagger.cloud/dagger/traces/9a61dd49fb060f47b0460c37b98a1b93?listen=d5f0c7c15a30af2a&listen=b3db866085d7081f&showHidden=c8aa8dbe98adc20f)
- [x] Spin out `elixir-sdk-dev` and absorb native toolchain (wip @kpenfound)
- [x] #11433 
- [x] Fix [error in `php-sdk-dev:generate`](https://dagger.cloud/dagger/traces/198b121a6a916fa1f95bec44a17deb10?span=5a4dc519e3105578) (done @TomChv)
- [x] `all-sdk` toolchain: actually include all SDKs (done @TomChv)
- [x] Fix commented-out code in typescript SDK
- [x] Fix module loading error in `.github/workflows/publish.yml`
- [x] Change CI configuration to call `dagger checks`
- [x] Test with new `dagger checks` #11374 
- [x] Fix missing files error in `dagger call php-sdk generate` (wip @shykes)
- [x] Fix broken json-schema generation in `docs-dev` (move it to `engine-dev`, for correct source access)
- [x] Fix toolchain configuration for `go lint` and `go check-tidy`
- [x] Fix performance bottleneck when engine loads toolchains
- [x] Fix display of `toolchains list`, too verbose
- [x] Spin out `engine-dev`
- [x] Spin out `cli-dev`
- [x] Spin out `docs-dev`
- [x] Spin out `ci`
- [x] Spin out Go toolchain
- [x] Spin out `go-sdk-dev`
- [x] Spin out `python-sdk-dev` and absorb native toolchain
- [x] Spin out `typescript-sdk-dev` and absorb native toolchain
- [x] Spin out `php-sdk-dev` and absorb native toolchain
- [x] Spin out `rust-sdk-dev` and absorb native toolchain
- [x] Spin out `java-sdk-dev` and absorb native toolchain
- [x] Spin out `dotnet-sdk-dev` and absorb native toolchain
- [x] Spin out `helm-dev`
- [x] Spin out `scripts` to `installers`
- [x] Convert  `.changes` to a toolchain (`changelog`)
- [x] Convert `releaser` to a toolchain (`release`)
- [x] Convert `scan.go` to a toolchain  (`security`)
- [x] Remove top-level constructors
- [x] Remove all non-builtin SDKs, for load time

### DIFF
--- a/toolchains/php-sdk-dev/main.go
+++ b/toolchains/php-sdk-dev/main.go
@@ -26,7 +26,6 @@ type PhpSdkDev struct {
 	Workspace         *dagger.Directory // +private
 	DoctumConfigPath  string            // +private
 	SourcePath        string            // +private
-	BaseContainer     *dagger.Container
 }
 
 // Develop the Dagger PHP SDK (experimental)
@@ -42,14 +41,22 @@ func New(
 	// +default="docs/doctum-config.php"
 	doctumConfigPath string,
 ) *PhpSdkDev {
-	composerBinary := dag.Container().
-		From(phpSDKComposerImage).
-		File("/usr/bin/composer")
+	return &PhpSdkDev{
+		Workspace:         workspace,
+		OriginalWorkspace: workspace,
+		SourcePath:        sourcePath,
+		DoctumConfigPath:  doctumConfigPath,
+	}
+}
 
+func (t PhpSdkDev) BaseContainer() *dagger.Container {
 	// Extract the PHP base container from the native SDK dev module
 	// - We build the base container eagerly, to avoid keeping a reference to DaggerDev
 	// - But we build the full dev container *lazily*, because we may have mutated our workspace with generated files
-	baseContainer := dag.Container().
+	composerBinary := dag.Container().
+		From(phpSDKComposerImage).
+		File("/usr/bin/composer")
+	return dag.Container().
 		From(phpSDKImage+"@"+phpSDKDigest).
 		WithMountedFile("/usr/bin/composer", composerBinary).
 		WithMountedCache(
@@ -63,14 +70,6 @@ func New(
 		With(func(c *dagger.Container) *dagger.Container {
 			return dag.DaggerEngine().InstallClient(c)
 		})
-
-	return &PhpSdkDev{
-		Workspace:         workspace,
-		OriginalWorkspace: workspace,
-		SourcePath:        sourcePath,
-		DoctumConfigPath:  doctumConfigPath,
-		BaseContainer:     baseContainer,
-	}
 }
 
 // Returns the PHP SDK workspace mounted in a dev container,
@@ -80,7 +79,7 @@ func (t PhpSdkDev) DevContainer(
 	//+default="false"
 	runInstall bool,
 ) *dagger.Container {
-	ctr := t.BaseContainer.
+	ctr := t.BaseContainer().
 		WithMountedDirectory(".", t.Workspace).
 		WithWorkdir(t.SourcePath)
 


TomChv:
Should we remove sdk dev that has been converted to toolchain or we will remove them in a follow up PR?

shykes:
> Should we remove sdk dev that has been converted to toolchain or we will remove them in a follow up PR?

~~Not 100% sure I understand the question.~~ (after reading your commit, it makes sense)

The answer is yes :)

Here's what we want to do in this PR, ideally:

1. Translate `sdk/typescript/dev` to Go ✅
2. Absorb `sdk/typescript/dev` into `toolchains/typescript-sdk-dev` ✅
3. Remove now useless `sdk/typescript/dev`
4. Bonus: translate `toolchains/typescript-sdk-dev` to dang :-P

wingyplus:
May I know the reason why convert the sdk-dev to Go? I remember that it’s use as a dogfood the sdk itself. 

However, I can help converting Elixir to Go if you preffered.

shykes:
> May I know the reason why convert the sdk-dev to Go? I remember that it’s use as a dogfood the sdk itself.

It's a tradeoff... The dogfooding is definitely useful. But we have other problems that are more important:

1. Awkward wrapping. The current situation is not a clean handoff to a native module for each SDK. Instead, it's a "worst of both worlds" where the monolith implements an outer module, which wraps the inner native module. The division between outer and inner module is different in each SDK, so everything is very inconsistent. And debugging anything takes a long time because you have to jump back and forth between the two modules and languages.

2. Big changes in the monolith. For a long time we just left the monolith alone, and only made small sporadic edits to keep it working. But now, we are refactoring the monolith in a major way - in fact we are splitting it up, refactoring, simplifying... It's a *lot* of work and rapid iteration is crucial. Having everything in one language really helps. We can revisit after the big split up, once the architecture is clean.

3. Speed. At the moment there is a performance penalty to using multiple SDKs in the same module. We are working on fixing that in the engine (and also there is SDK work required, like supporting the new `moduleTypes` entrypoint to support lazy loading). But in the meantime we really need our own module to be less slow. Switching to Go helps. And even further: we might port it to [dang](https://github.com/vito/dang), a new language by @vito that is Dagger-native and *requires no codegen* - leading to unbeatable performance.

> However, I can help converting Elixir to Go if you preffered.

Thank you :) If you start doing it, please edit the top-level message in #11398 to put your name next to the corresponding task. So that @TomChv and @eunomie can avoid conflict (they are working on the SDK-related tasks in that issue)

TomChv:
I added the dotnet-sdk-dev toolchain, originally in Go then I converted it to dang.
Issue is updated.

------

:bulb: I didn't add the dotnet-sdk-dev in for now, because it doesn't really implement a `Generate` function so should we simply create a placeholder for that function and add it to the list?

-------

I also noticed in `AllSdks` that you assume each SDK toolchain implements `Lint` & `Test` 
```go
	type tester interface {
		Test(context.Context) error
	}

	type linter interface {
		Lint(context.Context) error
	}
```

But for example in PHP, we directly name the tool that we use to lint
```go
// Lint the PHP code with PHP CodeSniffer (https://github.com/squizlabs/PHP_CodeSniffer)
// +check
func (t PhpSdkDev) PhpCodeSniffer(ctx context.Context) error {
```

```go
// Analyze the PHP code with PHPStan (https://phpstan.org)
// +check
func (t PhpSdkDev) PhpStan(ctx context.Context) error {
	_, err := t.
```

If we make a `Lint` method in `php` then we will have a duplication but if we don't we are not compliant with the `all-sdk` interface.
So how are we going to consolidate that?

TomChv:
Rust toolchain added to this PR, I cannot convert it to dang because we are using some external packages like semver, toml parser etc...
Maybe these could be supported in a potential dang stdlib ?? @vito

kpenfound:
> If we make a Lint method in php then we will have a duplication but if we don't we are not compliant with the all-sdk interface.
So how are we going to consolidate that?

My vote would be that we still conform to the interface, and the module can still have a private method that has a named linter tool if thats desired, and the lint function can call that. Interfaces are super valuable, and especially when they're in front of an implementation detail like which tool is being run behind the scenes, because then that can change in the future but the interface remains usable

shykes:
> I added the dotnet-sdk-dev toolchain, originally in Go then I converted it to dang. Issue is updated.
> 
> 💡 I didn't add the dotnet-sdk-dev in for now

I am confused, aren't your 2 comments contradictory? You added it or you didn't add it?

>  ...because it doesn't really implement a `Generate` function so should we simply create a placeholder for that function and add it to the list?

What list? Sorry I'm confused.

shykes:
> I also noticed in `AllSdks` that you assume each SDK toolchain implements `Lint` & `Test`
> 
> ```go
> 	type tester interface {
> 		Test(context.Context) error
> 	}
> 
> 	type linter interface {
> 		Lint(context.Context) error
> 	}
> ```

> If we make a `Lint` method in `php` then we will have a duplication but if we don't we are not compliant with the `all-sdk` interface. So how are we going to consolidate that?

You missed the fact that these interfaces are defined in a file called `stopgap_checks.go`, with with these comments inside:

```golang
// TODO: remove after merging https://github.com/dagger/dagger/pull/11211
```

Basically this interface is made obsolete by `dagger checks`. The name `Lint` should have not special meaning, and we shouldn't have to write our own check aggregation functions - now `dagger checks` can do the aggregation for us.

Also... there *is* a `Lint` function in the PHP SDK toolchain... in `toolchains/php-sdk-dev/stopgap_checks.go` (also to be removed)

TomChv:
> I am confused, aren't your 2 comments contradictory? You added it or you didn't add it?

Sorry I forgot a word, I meant I didn't add the dotnet-sdk-dev in the `all_sdk` module

> What list? Sorry I'm confused.

The list of SDK is in `all_sdk`

kpenfound:
@shykes I pushed up a WIP elixir dev module in dang

It has the same FIXMEs as the typescript one as well as this `syncImage` function that it seems like only elixir has. All of them need String.replace @vito

toolchains/php-sdk-dev/main.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+	return t.Changes(), nil
+}
+
+func (t *PhpSdkDev) Changes() *dagger.Changeset {
+	return t.Workspace.Changes(t.OriginalWorkspace)
+}
+
+func (t *PhpSdkDev) WithGeneratedClient() *PhpSdkDev {
+	relLayer := t.DevContainer(false).
```
charjr:
```suggestion
	relLayer := t.DevContainer(true).
```
Codegen can't run without the dependencies installed, aka. `composer install`.
it's hitting this:

https://github.com/dagger/dagger/blob/3807ba08e53f9506a02bbfc564225ff16a571b63/sdk/php/scripts/codegen.php#L9-L15

The autoload file is created by composer when installing dependencies.

TomChv: Done!

TomChv:
So I fixed an error on the `php-sdk-dev` and it works fine when running `dagger -m toolchains/php-sdk-dev -c 'generate | export .'` but `dagger checks check-generated` is failing.

I'm really confused about the Changeset... Running `dagger -m toolchains/php-sdk-dev -c 'generate | export .'` doesn't change the repo but `dagger -m toolchains/php-sdk-dev -c 'generate | modified-paths'` returns a long list of modification.

eunomie:
I just pushed a commit that aligns `dagger toolchain list` with `dagger functions` and fixes all the descriptions (description for toolchain list + descriptions in dang that are using a `pub description = "..."` line.

image

TomChv:
Check generated & `go/**` are timing out in the CI but locally they are working fine, I'm trying to figure out what's the origin

Screenshot 2025-11-19 at 16 23 53

eunomie:
> Check generated & `go/**` are timing out in the CI but locally they are working fine, I'm trying to figure out what's the origin

yes, it's the case for most of the `checks test-split` also 🤔

TomChv:
> > Check generated & `go/**` are timing out in the CI but locally they are working fine, I'm trying to figure out what's the origin
> 
> yes, it's the case for most of the `checks test-split` also 🤔

I noticed locally that sometime the command is finish but the telemetry is still uploading.
For example: https://dagger.cloud/Quartz/traces/e83fb7de1144d846239b88d851e71936


https://github.com/user-attachments/assets/b291e848-be91-4ee7-8c1c-0e0d5ba15a4e


We see that `go/checkTidy` completed in 2m12s but the traces is still running on cloud.
Maybe that's what create the timeout in CI?

kpenfound:
@TomChv @eunomie yeah if you see the timed out checks are all from dagger cloud, but the corresponding github actions execution is a failed check. So basically dagger cloud never reported it as finished/failed I guess

TomChv:
> @TomChv @eunomie yeah if you see the timed out checks are all from dagger cloud, but the corresponding github actions execution is a failed check. So basically dagger cloud never reported it as finished/failed I guess

Yeah it seems there's an issue with telemetry and checks, I think the engine may be killed before sending the last span or finish signal?

TomChv:
There's also an issue with go mod tidy, I don't really understand the root cause but I also saw it on our CI.
I tried to run `go mod tidy` on every go.mod of the repo but it didn't fixed the problem

Screenshot 2025-11-19 at 17 20 48

kpenfound:
@TomChv if its not happening locally and only in CI, we probably need to pull in main

TomChv:
> @TomChv if its not happening locally and only in CI, we probably need to pull in main

That go mod tidy error is happening in both

eunomie:
> I tried to run `go mod tidy` on every go.mod of the repo but it didn't fixed the problem

`dagger call go tidy` is what I run on the previous commit to fix it.
The thing is it worked to fix `go/check-tidy` but now it's failing when running tests.
Before it was possible to run the tests but `go/check-tidy` was failing.
🙃 

And I ran the `go tidy` from a playground, that way it was working. From my own machine, locally, it's going back to the previous state. Not sure what really is the issue and how to fix it.

kpenfound:
I think I found the issue @eunomie @TomChv 

The main dagger.json ignores `core/integration/**` in the go toolchain, however those files have imports relevant to the main go.mod. So when you ran go tidy with the toolchain it incorrectly removed things from the go.mod

shykes:
This one looks like a flake: https://dagger.cloud/dagger/traces/93301f52a469301d583232c8a4a110fe?span=a988b24b1568c38a

cc @kpenfound @vito @alexcb

shykes:
The remaining issue is that `.docker/config.json` seems to *NOT* be available in the CI runner's home directory at the moment.

cc @marcosnils @matipan @kpenfound 

See [any failed check in this group](https://dagger.cloud/dagger/ci?branch=ci-faster-load&minTimestamp=1763686132555&ref=da347876d7e2706b95080207f7a345e5b3d550aa&repo=github.com%2Fdagger%2Fdagger) for an example of what I mean.

sipsma:
> The remaining issue is that `.docker/config.json` seems to _NOT_ be available in the CI runner's home directory at the moment.
> 
> cc @marcosnils @matipan @kpenfound
> 
> See [any failed check in this group](https://dagger.cloud/dagger/ci?branch=ci-faster-load&minTimestamp=1763686132555&ref=da347876d7e2706b95080207f7a345e5b3d550aa&repo=github.com%2Fdagger%2Fdagger) for an example of what I mean.

@shykes took a look, I think this may be hitting a bug in env loading that's existed since it was added.

env values aren't actually loaded until this line: https://github.com/sipsma/dagger/blob/81fe990dd3cd47ae45dd91ba680f8854aa56aa66/core/modfunc.go#L543-L543

but that line doesn't specify to load from the context of the main client. So in your case in CI since the env filling in an argument of a function called by another function, it's trying to load the secret file from the module, not the main caller. So then it doesn't find it.

I can try to push a fix here if that's okay, but would require bumping the outer engine to a non-released commit again

EDIT: trying a fix here: https://github.com/dagger/dagger/pull/11459

shykes:
@sipsma aside from a few flakes, this seems to be the only remaining failed test: https://dagger.cloud/dagger/traces/197bd4d1bda1b21093f00e54c3e15d95?span=8117a85975265746

```

        Error Trace:    /app/core/integration/module_test.go:2999
                                                /go/pkg/mod/github.com/dagger/testctx/oteltest@v0.0.3/log.go:37
                                                /go/pkg/mod/github.com/dagger/testctx/oteltest@v0.0.3/trace.go:83
        Error:          Received unexpected error:
                        exit code: 2 [traceparent:197bd4d1bda1b21093f00e54c3e15d95-00751355a07ac6bf]
        Test:           TestModule/TestSecretNested/pass_embedded_secrets_between_modules/double_nested_and_called_repeatedly
```

shykes:
All green, let's do this!

PR #11374, opened 2025-11-07
Switch checks to `@check` pragma, improve UI

vito:
This PR switches checks to use the `@check` pragma/decorator, improving the developer experience.

Also adds a new UI:

[asciicast](https://asciinema.org/a/754999)

### Go

Add a `+check` pragma comment above functions that should be checks:

```go
package main

import "context"

type MyModule struct{}

// Returns a passing check
// +check
func (m *MyModule) PassingCheck(ctx context.Context) error {
    _, err := dag.Container().From("alpine:3").WithExec([]string{"sh", "-c", "exit 0"}).Sync(ctx)
    return err
}

// Returns a failing check
// +check
func (m *MyModule) FailingCheck(ctx context.Context) error {
    _, err := dag.Container().From("alpine:3").WithExec([]string{"sh", "-c", "exit 1"}).Sync(ctx)
    return err
}
```

### Python

Use the `@check` decorator on functions:

```python
import dagger
from dagger import dag, check, function, object_type

@object_type
class MyModule:
    @function
    @check
    async def passing_check(self) -> None:
        """Returns a passing check"""
        await dag.container().from_("alpine:3").with_exec(["sh", "-c", "exit 0"]).sync()

    @function
    @check
    async def failing_check(self) -> None:
        """Returns a failing check"""
        await dag.container().from_("alpine:3").with_exec(["sh", "-c", "exit 1"]).sync()
```

### TypeScript

Use the `@check()` decorator on methods:

```typescript
import { object, func, check } from "@dagger.io/dagger";
import { dag } from "./dagger/index.js";

@object()
class MyModule {
  /**
   * Returns a passing check
   */
  @func()
  @check()
  async passingCheck(): Promise<void> {
    await dag
      .container()
      .from("alpine:3")
      .withExec(["sh", "-c", "exit 0"])
      .sync();
  }

  /**
   * Returns a failing check
   */
  @func()
  @check()
  async failingCheck(): Promise<void> {
    await dag
      .container()
      .from("alpine:3")
      .withExec(["sh", "-c", "exit 1"])
      .sync();
  }
}
```

## Running Checks

Once you've marked functions as checks, run them with:

```bash
# List all checks
dagger checks -l

# Run all checks
dagger checks

# Run specific checks
dagger checks my-module/passing-check
```

## LLM `.contributing` docs

Adding the pragma was heavily LLM assisted. Adding pragmas is in the category of 'known but still annoying and tedious amount of work'. I started by having it research previous commits, and then write a doc, and then follow the doc for this PR. The docs aren't perfect, since some of them focused on _arg_ directives whereas this is a _field_ directive, but they still help you get 95% there; in my case the LLM figured out the right way, and I just needed to fix a couple things in post.

### DIFF
--- a/dagql/dagui/db.go
+++ b/dagql/dagui/db.go
@@ -3,7 +3,6 @@ package dagui
 import (
 	"context"
 	"iter"
-	"maps"
 	"slices"
 	"sort"
 	"time"
@@ -922,61 +921,6 @@ func WalkTree(tree []*TraceTree, f func(*TraceTree, int) WalkDecision) {
 	walk(tree, 0)
 }
 
-func (db *DB) CollectErrors(rows *RowsView) []*TraceTree {
-	reveal := make(map[*TraceTree]struct{})
-
-	rootCauses := NewSpanSet()
-
-	var collect func(row *TraceTree, ancestorFailed bool)
-	collect = func(tree *TraceTree, ancestorFailed bool) {
-		failed := tree.Span.IsFailedOrCausedFailure()
-
-		if failed {
-			if !tree.Span.RollUpLogs {
-				reveal[tree] = struct{}{}
-			}
-			if tree.Span.ErrorOrigin != nil {
-				rootCauses.Add(tree.Span.ErrorOrigin)
-			}
-		}
-
-		unset := tree.Span.IsUnset()
-
-		if failed ||
-			// continue through UNSET spans beneath failed parents; these correspond
-			// to basic span.End() calls which intend to just add measurement without
-			// bubbling up or masking failure
-			(unset && ancestorFailed) {
-			for _, child := range tree.Children {
-				collect(child, true)
-			}
-		}
-	}
-
-	for _, row := range rows.Body {
-		collect(row, false)
-	}
-
-	// Prioritize root cause errors if any are found.
-	if len(rootCauses.Order) > 0 {
-		orig := maps.Clone(reveal)
-		clear(reveal)
-		for _, span := range rootCauses.Order {
-			if tree, ok := rows.BySpan[span.ID]; ok {
-				reveal[tree] = struct{}{}
-			} else {
-				slog.Warn("tree not found for error origin span:", "id", span.ID, "name", span.Name)
-			}
-		}
-		if len(reveal) == 0 {
-			// fallback to original set
-			reveal = orig
-		}
-	}
-
-	return collectParents(rows.Body, reveal)
-}
-
 func (db *DB) FindResource(filter attribute.KeyValue) *resource.Resource {
 	for _, res := range db.Resources {
 		for _, kv := range res.Attributes() {
@@ -986,38 +930,4 @@ func (db *DB) FindResource(filter attribute.KeyValue) *resource.Resource {
 		}
 	}
 	return nil
-}
-
-func collectParents(rows []*TraceTree, targets map[*TraceTree]struct{}) []*TraceTree {
-	var result []*TraceTree
-
-	for _, row := range rows {
-		contains := false
-		for target := range targets {
-			if isOrContains(row, target) {
-				contains = true
-				break
-			}
-		}
-		if !contains {
-			continue
-		}
-		rowCopy := *row
-		rowCopy.Children = collectParents(row.Children, targets)
-		result = append(result, &rowCopy)
-	}
-
-	return result
-}
-
-func isOrContains(row, target *TraceTree) bool {
-	if row == target {
-		return true
-	}
-	for _, child := range row.Children {
-		if isOrContains(child, target) {
-			return true
-		}
-	}
-	return false
 }


charjr:
@vito What are "checks" in this instance?

vito:
@charjr More info: #11211 - checks are essentially a way to make all your CI "is this commit good?" functions discoverable and runnable in a more convenient way, with the eventual goal being to eliminate a ton of CI boilerplate.

charjr:
> @charjr More info: #11211 - checks are essentially a way to make all your CI "is this commit good?" functions discoverable and runnable in a more convenient way, with the eventual goal being to eliminate a ton of CI boilerplate.

Oh, that is good. :drooling_face: 

@vito I assume we'll want a `#[Check]` decorator for PHP? Are there any other prerequisites? (I haven't sorted custom spans yet)

vito:
@charjr Yeah, that would be great!

> Are there any other prerequisites? (I haven't sorted custom spans yet)

No prerequisites, the pragma is pretty simple - it just chains a `Function.WithCheck()` call onto the function builder. If you find the time, feel free to submit a PR targeting this one and I can merge it in, or it can just be a follow-up. (Or if you have permission feel free to push to this PR directly.)

.contributing/adding-pragmas.md
sipsma, line 1: noice

.dagger/stopgap_checks.go
```diff
@@ -6,59 +6,37 @@ import (
...
 		WithJob("SDKs", dev.dryRunSDKs).
 		Run(ctx)
 }
 
 func (dev *DaggerDev) dryRunSDKs(ctx context.Context) error {
 	type releaseDryRunner interface {
 		Name() string
-		ReleaseDryRun(context.Context) (MyCheckStatus, error)
+		// +check
+		ReleaseDryRun(context.Context) error
```
sipsma, line 34: does this work/do anything on inline interfaces like this?

.dagger/test.go
```diff
@@ -40,10 +40,11 @@ func (t *Test) All(
 	envFile *dagger.Secret,
 	// +optional
 	testVerbose bool,
-) (MyCheckStatus, error) {
+	// +check
```
sipsma: is this in the right place? same comment for line 165

.contributing/python-decorator-guide.md
```diff
@@ -0,0 +1,524 @@
...
+## Common Gotchas
+
+### 1. Forgetting to Regenerate the SDK
+
+If you add a new API method to `core/schema/module.go`, you must regenerate the Python SDK:
+
+```bash
+dagger develop --sdk=python
+# or
+make sdk-generate
```
sipsma, line 316: seems hallucinated

.contributing/python-decorator-guide.md
```diff
@@ -0,0 +1,524 @@
...
+# Run unit tests
+cd sdk/python
+pytest tests/
+
+# Test a sample module
+cd /tmp
+dagger init --sdk=python my-test
+# Edit dagger.json module file with @function(check=True)
+dagger functions  # Should show the check function
+dagger call lint  # Should execute successfully
```
sipsma, line 477: this doesn't seem correct (running `pytest` directly rather than with dagger and editing dagger.json to add a lint func)

cmd/dagger/.dagger/publish.go
```diff
@@ -196,8 +196,8 @@ func s3Path(bucket string, path string, args ...any) string {
 }
 
 // Verify that the CLI builds without actually publishing anything
-func (cli *DaggerCli) ReleaseDryRun(ctx context.Context) (CheckStatus, error) {
-	return CheckCompleted, parallel.New().
+func (cli *DaggerCli) ReleaseDryRun(ctx context.Context) error {
```
sipsma, line 200: do we want `//+check` here?

core/checks.go
```diff
@@ -134,47 +106,85 @@ func (r *CheckGroup) Run(ctx context.Context) (*CheckGroup, error) {
...
+			// Call the check
+			var status dagql.AnyResult
+			if err := dag.Select(dagql.WithRepeatedTelemetry(ctx), checkParent, &status, selectPath[len(selectPath)-1]); err != nil {
+				return err
 			}
-			if checkErr != nil {
-				return checkErr
+
+			if obj, ok := dagql.UnwrapAs[dagql.AnyObjectResult](status); ok {
+				// If the check returns a syncable type, sync it
```
sipsma, line 168: nit: I guess if a user puts a `sync` method on their own custom objects, we'll call it too? Is that intentional? Just checking because it gives special meaning to any method named sync, which I haven't seen before.

.dagger/test.go
vito: Noticed this during self-review, will push soon. (We also don't actually want `// +check` for `test specific` since that's the one that takes a filter arg.)

.dagger/stopgap_checks.go
vito: Nah, this is just a multi-cursor edit "hallucination"

core/checks.go
vito: Yeah - we already do this for `dagger call` (and maybe shell):

https://github.com/dagger/dagger/blob/8984278890d3158e3164f4e4090aba94773fb023/cmd/dagger/functions.go#L602-L631

I think at least defensible to just claim the `sync` name to always mean the same thing tbh, but it also hasn't come up in the wild yet to my knowledge.

sipsma (approved):

PR #11377, opened 2025-11-07
bump containerd deps for cves

sipsma:
Also reducing the verbosity since there were complaints and making the cache volume locked since locally I hit an error due to it being used concurrently


.dagger/scan.go
```diff
@@ -25,14 +25,15 @@ func (dev *DaggerDev) Scan(ctx context.Context) (MyCheckStatus, error) {
...
 		From("aquasec/trivy:0.67.2@sha256:e2b22eac59c02003d8749f5b8d9bd073b62e30fefaef5b7c8371204e0a4b0c08").
 		WithMountedDirectory("/mnt/ignores", ignoreFiles).
-		WithMountedCache("/root/.cache/", dag.CacheVolume("trivy-cache")).
+		WithMountedCache("/root/.cache/", dag.CacheVolume("trivy-cache"), dagger.ContainerWithMountedCacheOpts{
+			Sharing: dagger.CacheSharingModeLocked,
+		}).
 		With(dev.withDockerCfg)
 
 	commonArgs := []string{
 		"--format=json",
```
vito: While we're in the area, should we remove this? Doesn't look like we're parsing the output, and the human readable output seems much better.

(But, maybe there's a reason...)

vito (approved):

PR #11380, opened 2025-11-08
fix python sdk tests to actually use sidecar dev engine

sipsma:
The tests were being given a container with a dev engine attached via service dep, but then were being run as nested execs, which overrode the settings and made them not actually hit the dev engine they were supposed to hit at all.

This adds and option to the tests to disable nested execs (which is particular to the nested dev engine sidecar setup in our top-level module) and uses that to run tests.

---

This fixes the errors entirely for me locally. It also is likely the explanation for them since the python tests were being executed with a *very* confused setup where they had access to two separate engines and primarily were using the one they weren't intended to. I have not connected all the dots on how that created the flakes+errors we were seeing, but may just chock it up to "undefined behavior". I will poke around a little bit more to make sure there's not any deeper issues though.

### DIFF
--- a/sdk/python/dev/src/python_sdk_dev/main.py
+++ b/sdk/python/dev/src/python_sdk_dev/main.py
@@ -220,7 +220,11 @@ class PythonSdkDev:
         ] = False,
     ) -> TestSuite:
         """Run the test suite."""
-        return TestSuite(container=self.container, version=version, disable_nested_exec=disable_nested_exec)
+        return TestSuite(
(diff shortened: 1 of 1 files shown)


grouville:
Thank you, I don't know how you debugged that 🙏

sipsma:
> Thank you, I don't know how you debugged that 🙏

@grouville I looked at the code and the traces and saw that the dev engine the tests were supposed to be hitting had logs showing nothing connected to it.

PR #11392, opened 2025-11-10
core: fix engine panic in Container.withExec

vito:
(Needs a test, submitting early for visibility)

Noticed this panic locally and chucked it into an LLM to avoid a rabbit hole, here's its generated description:

---

Fixes a panic that occurs during container execution error handling when there's a mismatch between mount indices and output indices.

## The Bug

```
panic: runtime error: index out of range [2] with length 2
  at /app/core/container_exec.go:302
```

The error handling code in `WithExec`'s deferred function was incorrectly assuming that iteration indices through the `results` array correspond to indices in `p.OutputRefs`. This breaks when there are mounts with no outputs (readonly mounts, cache mounts, tmpfs, secrets, sockets).

## Root Cause

There are two distinct index spaces for mounts:

1. **Mount index**: Position in the `mounts.Mounts` array (always sequential: 0, 1, 2, 3, ...)
2. **Output index**: The `mount.Output` field value (can have gaps: 0, 1, 2, -1, -1, 3, -1, ...)

The `results` array from `extractContainerBkOutputs` is indexed by **output index**, not by mount position or iteration order. The buggy code was iterating `results` and assuming indices aligned with `p.OutputRefs`, which they don't.

Example scenario that triggers the panic:
- Mount 0 (root): Output = 0
- Mount 1 (meta): Output = 1
- Mount 2 (user bind): Output = 2
- Mount 3 (cache): Output = -1 (SkipOutput)
- Mount 4 (user bind): Output = 3

Result: `results` has length 4, but when iterating we'd try to access `p.OutputRefs[3]` when it might only have 3 entries (or be ordered differently).

## The Fix

Changed both loops to follow the same pattern as `setAllContainerMounts` in `dagop.go`:

1. **First loop** (lines 301-306): Iterate through `p.OutputRefs`, look up each mount using `outputRef.MountIndex`, then use `mount.Output` to index into `results`.

2. **Second loop** (lines 320-399): Iterate through `mounts.Mounts` by `mountIdx`, skip mounts with `Output == pb.SkipOutput`, and use `mount.Output` to index into `results`.

Both fixes properly map between mount indices and output indices, preventing the out-of-bounds access.

### DIFF
--- a/core/container_exec.go
+++ b/core/container_exec.go
@@ -298,8 +298,11 @@ func (container *Container) WithExec(
 			if err != nil {
 				return
 			}
-			for i, res := range results {
-				execMounts[p.OutputRefs[i].MountIndex] = res
+			for _, outputRef := range p.OutputRefs {
+				mount := mounts.Mounts[outputRef.MountIndex]
+				if mount.Output != pb.SkipOutput && results[mount.Output] != nil {
+					execMounts[outputRef.MountIndex] = results[mount.Output]
+				}
 			}
 			for _, active := range p.Actives {
 				if active.NoCommit {
@@ -314,9 +317,16 @@ func (container *Container) WithExec(
 				}
 			}
 
-			for i, res := range results {
+			for mountIdx, mount := range mounts.Mounts {
+				if mount.Output == pb.SkipOutput {
+					continue
+				}
+				res := results[mount.Output]
+				if res == nil {
+					continue
+				}
 				iref := res.Sys().(*worker.WorkerRef).ImmutableRef
-				switch i {
+				switch mountIdx {
 				case 0:
 					rootfsDir := &Directory{
 						Result: iref,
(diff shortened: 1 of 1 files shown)


tiborvass:
I agree that the code is convoluted, but this is my finding:

getAllContainerMounts() sets the mount.Output field by advancing an outputIdx only if Output is NOT set to SkipOutput. The resulting count is stored in OutputCount.
IOW we have this invariant: #mounts == OutputCount + #(Output == SkipOutput)

This PR doesn't want to assume that and redoes the != SkipOutput check everywhere.
If it's for greater safety so be it, but what bothers me is that it just seems to cover up the actual bug that is breaking assumptions in the code (like OutputCount is the number of OutputRefs).

> The results array from extractContainerBkOutputs is indexed by output index, not by mount position or iteration order. The buggy code was iterating results and assuming indices aligned with p.OutputRefs, which they don't.

results indices go from 0 to OutputCount, for sure.
What is less clear is whether OutputCount == len(OutputRefs). (Still trying to see when that's the case)

PR #11393, opened 2025-11-10
ref: rewrite python-sdk-dev into Go

eunomie:
for performances

This should be a drop-in replacement, API should be the same.


TomChv:
@eunomie Hey! I think you should put that work in https://github.com/dagger/dagger/pull/11373 directly in `toolchains/python-sdk-dev` :)

eunomie:
> @eunomie Hey! I think you should put that work in https://github.com/dagger/dagger/pull/11373 directly in `toolchains/python-sdk-dev` :)

Yes, working on it ;-)

PR #11397, opened 2025-11-11
ci: switch to runners without cache

matipan:
We've seeing a few issues on our CI that are related to the state of the underlying engine provisioned and managed by namespace. A few examples here:
- https://dagger.cloud/dagger/traces/7148ab0874f4f47c06458481f8b2c949
- https://dagger.cloud/dagger/traces/425ccc4b797aec8a6f42cd2b9a996cac

This PR moves away of the namespace managed dagger engine to an engine that is started on demand within the runner instance.

### DIFF
--- a/.github/main.go
+++ b/.github/main.go
@@ -32,6 +32,7 @@ type CI struct {
 	Workflows *dagger.Gha
 
 	AltRunner          *dagger.Gha // +private
+	AltSilverRunner    *dagger.Gha // +private
 	AltRunnerWithCache *dagger.Gha // +private
 }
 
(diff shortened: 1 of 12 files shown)


.github/main.go
```diff
@@ -340,6 +348,11 @@ func AltSilverRunnerWithCache() []string {
 	return Alt2Runner(8, true)
 }
 
+// Alternative Silver runner without caching: Single-tenant, 8 cpu
```
sipsma, line 351: I had the memory that we were getting 16 cpus before, am I making that up? The test runtimes in the PR don't look too shabby so not a blocker, just curious. If it is possible to get 16 that might be nice if it has a meaningful impact on test runtimes though.

sipsma (approved):

.github/main.go
matipan: Using the `.github` module lingo: all workflows except those grouped under "Core tests" were using "Silver" runners with caching enabled. Core tests were using "Gold" runners without cache. Silver == 8 cores and Gold == 16 cores. You can see the function above this one: `AltSilverRunnerWithCache`. That runner was being used inside "withSimpleDaggerCheck" which is used for most workflows.

PR #11400, opened 2025-11-12
Update dotnet sdk to latest .NET 10 LTS release

pjmagee:
- the go dev module also needs dotnet base image updating from 8 to 10


wingyplus:
@pjmagee Thanks for helping update! I will take a look at it tomorrow.

pjmagee:
> @pjmagee Thanks for helping update! I will take a look at it tomorrow.

no problem, i had some issues trying to figure out how to get introspection.json as it changed a lot but i found a way, and then i manage also get tests passing locally with package changes / minor changes based on linting rules / analyzers

frustratinly, i am still angry at source generator and tooling/IDE intellisense issues, Query appear is 'not valid' but when running/building/compiling code from command line, and running test from command line everything is passing, but vscode/Rider think query didnt exist, but i validated source generator work with update because i saw obj/net10/generated/Dagger.sdk.g.cs file was correct

PR #11401, opened 2025-11-12
telemetry: various low-hanging fruit optimizations

sipsma:
While working on checks/scale-out I managed to randomly hit locally the same hang we see in our CI occasionally where loading modules hangs forever and lots of spans never finish. In CI that seemed to be accompanied by `SQLITE BUSY` errors from telemetry, turns out locally it was too.

I only managed to repro it one more time despite many attempts in a loop. Engine SIGQUIT stack traces from both those occurrences (gist was giving me 500s, so uploading files) :
* [engine-goroutines-1.txt](https://github.com/user-attachments/files/23506111/engine-goroutines-1.txt)
* [engine-goroutines-2.txt](https://github.com/user-attachments/files/23506112/engine-goroutines-2.txt)

Main interesting thing is sessions are held open (perhaps expectedly if they can't shutdown for another reason) and lots of telemetry related stacks, with a single one stuck in a wait for the sqlite db to be unlocked.

I was hoping to track down the root cause of the seemingly permanent hang, but I couldn't repro enough to confirm whether any changes here are a fix. I suspect there's some actual bug somewhere that causes things to hang indefinitely, but didn't identify anything definitive. I even tried lowering the write lock timeout and triggered the SQLITE BUSY error very consistently but couldn't repro the hang consistently.

Instead I just accumulated a bunch of low-risk performance improvements, which are probably nice to have generally and hopefully at least prevent us from hitting the underlying bug as often.

Split up into commits each with more detailed messages.

### DIFF
--- a/core/telemetry.go
+++ b/core/telemetry.go
@@ -47,7 +47,7 @@ func AroundFunc(
 	context.Context,
 	func(res dagql.AnyResult, cached bool, rerr error),
 ) {
-	if dagql.IsSkipped(ctx) || isIntrospection(id) || isMeta(id) || isDagOp(id) {
+	if dagql.IsSkipped(ctx) || isIntrospection(ctx, id) || isMeta(id) || isDagOp(id) {
 		// introspection+meta are very uninteresting spans
 		// dagops are all self calls, no need to emit additional spans here
 		return ctx, dagql.NoopDone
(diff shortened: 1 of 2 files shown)


core/telemetry.go
```diff
@@ -347,14 +347,28 @@ func isIntrospection(id *call.ID) bool {
...
+	switch id.Receiver().Type().NamedType() {
+	case "Function":
+		switch id.Field() {
+		case "withCachePolicy",
+			"withArg",
+			"withSourceMap",
+			"withDescription":
+			return true
+		}
+	}
```
vito, line 372: I'm pretty torn on removing these since they're immensely useful when developing an SDK.

Can we send them with `--debug` is passed at least? We put it in OTel baggage already, just need to add a getter for it (grep for `debugBaggageKey`).

sipsma: Sure SGTM, fixed

PR #11403, opened 2025-11-12
feat: directory.Stat and container.Stat calls

alexcb:
This introduces `directory.Stat` and `container.Stat` calls which will return the file system status (e.g. name, type, size, permissions) of a file or directory.

This additionally switches some calls from `FileLLB()` to dagop `File()` calls.

### DIFF
--- a/core/integration/llmtest/hello-world.golden
+++ b/core/integration/llmtest/hello-world.golden
@@ -28,7 +28,7 @@
   },
   {
     "role": "user",
-    "content": "/app/main.go: stat /tmp/buildkit-mount4036873152/app/main.go: no such file or directory [traceparent:dfd6c4edd9550ded889de3c4cfec4646-fd9e714e00b16a68]",
+    "content": "stat /app/main.go: no such file or directory [traceparent:dfd6c4edd9550ded889de3c4cfec4646-fd9e714e00b16a68]",
     "tool_call_id": "toolu_011jkXpEMCTWmzgedkZGkMv3",
     "tool_errored": true
   },
--- a/core/container.go
+++ b/core/container.go
@@ -1511,6 +1511,7 @@ func (container *Container) File(ctx context.Context, filePath string) (*File, e
 		f, err = container.FS.Self().File(ctx, subpath)
 	case mnt.DirectorySource != nil: // mounted directory
 		f, err = mnt.DirectorySource.Self().File(ctx, subpath)
+		err = RestoreErrPath(err, filePath) // preserve the full filePath, rather than subpath
 	case mnt.FileSource != nil: // mounted file
 		return mnt.FileSource.Self(), nil
 	default:
--- a/core/directory.go
+++ b/core/directory.go
@@ -610,28 +610,47 @@ func (dir *Directory) Search(ctx context.Context, opts SearchOpts, verbose bool,
 	return results, nil
 }
 
+// cleanDotsAndSlashes is similar to path.Clean; however it does not remove any directory names, e.g. "keep/../this//.//" will return "keep/../this".
+// This is needed for cases where a referenced directory is a symlink, e.g. consider keep linking to some/other/directory, then keep/../this,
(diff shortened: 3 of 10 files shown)


core/integration/llmtest/hello-world.golden
```diff
@@ -28,7 +28,7 @@
   },
   {
     "role": "user",
-    "content": "lstat /app/main.go: no such file or directory [traceparent:dfd6c4edd9550ded889de3c4cfec4646-fd9e714e00b16a68]",
+    "content": "/app/main.go: stat /tmp/buildkit-mount4036873152/app/main.go: no such file or directory [traceparent:dfd6c4edd9550ded889de3c4cfec4646-fd9e714e00b16a68]",
```
sipsma: I think it'd be better to keep the error messages free of the internal engine paths. Mostly just for aesthetics and not confusing users. Is that a huge lift or pretty easy to tack on here?

core/container_exec.go
```diff
@@ -287,6 +287,14 @@ func (container *Container) WithExec(
 			if err != nil {
 				return
 			}
+
+			// FIXME there was an out of range panic that occurred when accessing execMounts below; however it's not clear why
```
sipsma, line 291: Do you have a repro? Sounds suspiciously like something @vito hit

core/file.go
```diff
@@ -480,6 +480,70 @@ func (file *File) Stat(ctx context.Context) (*fstypes.Stat, error) {
...
+	}
+
+	immutableRef, err := ref.CacheRef(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	osStatFunc := os.Stat
+	rootPathFunc := containerdfs.RootPath
+	// TODO Could there be a case where a File() is a symlink?
```
sipsma, line 505: Good question 😄 Hopefully not, but maybe it's worth a check in the code that it isn't? Just to be future proof?

core/container_exec.go
alexcb: unfortunately no repro -- i spent a couple hours trying to get one, but I only ever hit the panic once.

core/integration/llmtest/hello-world.golden
alexcb: yeah I was wondering if we should bite the bullet and spend some time catching and modifying errors.

I figure we could have some sort of common `TrimBuildkitPathFromErr()` helper, and use it whenever we are executing in a mount.

I'll give it a try and undo this and my change to `util/scrub/scrub.go`.

core/file.go
alexcb: so I went down a wee rabbit hole on this one, and wrote this test:

```
func (DirectorySuite) TestDirectoryRespectsSymlinks(ctx context.Context, t *testctx.T) {
	c := connect(ctx, t)
	d := c.Container().
		From(alpineImage).
		WithExec([]string{"sh", "-c", "mkdir -p /testdir/sub/submarine/subway && cd /testdir && echo 'important' > sub/submarine/data && ln -s sub/submarine/subway d && test -f d/../data"}).
		Directory("/testdir/d/..")

	entries, err := d.Entries(ctx)
	require.NoError(t, err)
	require.Equal(t, []string{"data", "subway/"}, ents)
}
```

it fails, because under the hood there's a `path.Clean("/testdir/d/..")` that's called, which returns `"testdir/"` rather than `"/testdir/sub/submarine"`.

if we want to respect symlinks, typically one would need to call `filepath.EvalSymlinks(...)`; however, in our case, we would need to mount the filesystem and use something like `containerdfs.RootPath`.

Maybe it could be done without introducing additional mount calls, since we already perform a mount to call `Stat()`, to make sure the directory exists.

But maybe this is best left as a separate (lower priority) task?

sipsma: > But maybe this is best left as a separate (lower priority) task?

Yeah no problem, not a blocker at all

sipsma (approved):

PR #11408, opened 2025-11-13
Adds File.AsJSON() to simplify handling JSON file contents

### LINKED ISSUE #11313: File.asJSON
Dagger as a native File API, and a native JSON API, but it can't connect them directly. Clients have to 1) retrieve the contents of the file, 2) send the contents back to a JSON object. It would be more efficient (and less verbose) to allow `File.asJSON`.

suprjinx:
This PR includes implementation (in core/file.go and core/json.go), an integration test, and the result of code generation.

Fixes #11313 

Assisted by Claude Code

### DIFF
--- a/core/integration/file_test.go
+++ b/core/integration/file_test.go
@@ -1068,23 +1068,25 @@ func (FileSuite) TestFileAsJSON(ctx context.Context, t *testctx.T) {
 	c := connect(ctx, t)
 
 	t.Run("it converts json file contents to JSON", func(ctx context.Context, t *testctx.T) {
-		jsonValue := c.Directory().
+		jsonValue, err := c.Directory().
 			WithNewFile("test.json", `{ "somekey": "somevalue" }`).
 			File("test.json").
-			AsJSON()
+			AsJSON().
+			Field([]string{"somekey"}).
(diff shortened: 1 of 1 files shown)


sipsma (approved):
This LGTM, cc @dagger/team since it's an API addition

shykes (changes requested):
Thanks for the contribution!

Instead of returning a `JSON` type (which is a scalar), `File.asJSON` should return a `JSONValue` type, which is an object type with various in-engine JSON operations available. This way I can do eg:

```golang
sdk, err := dag.
  Address("https://github.com/dagger/dagger#main:dagger.json").
  File().
  AsJSON().
  Field([]string{"sdk", "source"}).
  AsString(ctx)
```

Sorry about the confusion, it's not ideal that we have a `JSON` and `JSONValue` type... The `JSON` type came first, and now we have to preserve compatibility...

suprjinx:
@shykes thanks for the guidance -- updated to use JSONValue in the api

suprjinx:
@shykes Let me know if the current changes look correct. Thanks!

shykes:
Thanks, will check it out today

PR #11410, opened 2025-11-13
add `telemetry.EndWithCause`, clean up more errors

vito:
* replaces the clunky telemetry.End fn pattern with a simple error pointer
* error origins are now tracked for internal calls, too, massively reducing duplicate errors
* fix `ExecError` wrapping to not print the full `input: foo.bar.baz: ...` error
* tracks the span as the origin of the error by re-assigning the pointer, if the error does not already have an origin
* improves error display for loading an unknown SDK
  * happened to use this to test it but then didn't like the original output, decided to clean up along the way
* get rid of truncation to fix TestTelemetry flake

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -57,7 +57,7 @@ func (ModuleSuite) TestInvalidSDK(ctx context.Context, t *testctx.T) {
 			With(daggerQuery(`{bare{containerEcho(stringArg:"hello"){stdout}}}`)).
 			Stdout(ctx)
 		require.Error(t, err)
-		requireErrOut(t, err, `The "foo-bar" SDK does not exist.`)
+		requireErrOut(t, err, `invalid SDK: "foo-bar"`)
 	})
 
 	t.Run("specifying version with either of go/python/typescript sdk returns error", func(ctx context.Context, t *testctx.T) {


cmd/dagger/llm.go
```diff
@@ -412,7 +412,7 @@ func (s *LLMSession) maybeAutoCompact(ctx context.Context) (*dagger.LLM, error)
 	// If we're over 80% context usage, automatically compact
 	if contextUsage > 0.80 {
 		ctx, span := Tracer().Start(ctx, "auto-compacting LLM history", telemetry.Reveal())
-		defer telemetry.End(span, func() error { return nil })
+		defer telemetry.EndWithCause(span, &rerr)
```
vito, line 415: differs, but intentional

core/telemetry.go
```diff
@@ -45,7 +45,7 @@ func AroundFunc(
 	id *call.ID,
 ) (
 	context.Context,
-	func(res dagql.AnyResult, cached bool, rerr error),
+	func(res dagql.AnyResult, cached bool, rerr *error),
```
vito, line 48: This tweak is necessary so that `AroundFunc` can reassign the error

shykes (approved):
There's only one way and that's forward :)

PR #11413, opened 2025-11-14
feat: add eager-mod-loading flag

TomChv:
As discussed with @shykes, with the lazy runtime loading addition, it became hard  to verify that your module and its dependencies compiles just by running dagger functions.

This PR adds the flag `eager-runtime` to all modules commands so it disables lazy loading on module runtime. That way, running `dagger functions --eager-runtime` also let you verify that your module compiles properly.

Add `EagerModuleLoading` in engine client metadata.

**Difference between [`dagger functions --eager-runtime`](https://dagger.cloud/Quartz/traces/d9456aa3b8559a4163258804a454f079?listen=01a416d941de42de&listen=95d02eab609ac765&listen=a762178577185ff8) & [`dagger functions`](https://dagger.cloud/Quartz/traces/7e831d54c9f9aa3b8dcc4ac0f0cdd39c?listen=ac3ac0ec32ba7b05&listen=73b735068de22090&listen=e06cf8adf8612a21&listen=987a1df74aafef41)**

Screenshot 2025-11-14 at 15 28 43

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -58,8 +58,8 @@ var (
 
 	force bool
 
-	autoApply          bool
-	eagerModuleLoading bool
+	autoApply    bool
+	eagerRuntime bool
 )
 
 const (
@@ -120,7 +120,7 @@ func moduleAddFlags(cmd *cobra.Command, flags *pflag.FlagSet, optional bool) {
 	flags.StringSliceVar(&allowedLLMModules, "allow-llm", defaultAllowLLM, "List of URLs of remote modules allowed to access LLM APIs, or 'all' to bypass restrictions for the entire session")
 
 	// Add the eager module loading flag to disable lazy load on runtime.
-	flags.BoolVar(&eagerModuleLoading, "eager-mod-loading", false, "load module runtime eagerly")
+	flags.BoolVar(&eagerRuntime, "eager-runtime", false, "load module runtime eagerly")
 }
 
 func init() {
--- a/cmd/dagger/function_name.go
+++ b/cmd/dagger/function_name.go
@@ -8,9 +8,9 @@ import (
 
 func initModuleParams(a []string) client.Params {
 	params := client.Params{
-		ExecCmd:            a,
-		Function:           functionName(a),
-		EagerModuleLoading: eagerModuleLoading,
+		ExecCmd:      a,
+		Function:     functionName(a),
+		EagerRuntime: eagerRuntime,
 	}
 
 	if !moduleNoURL {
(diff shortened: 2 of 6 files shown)


cmd/dagger/module.go
```diff
@@ -117,6 +118,9 @@ func moduleAddFlags(cmd *cobra.Command, flags *pflag.FlagSet, optional bool) {
 		defaultAllowLLM = strings.Split(allowLLMEnv, ",")
 	}
 	flags.StringSliceVar(&allowedLLMModules, "allow-llm", defaultAllowLLM, "List of URLs of remote modules allowed to access LLM APIs, or 'all' to bypass restrictions for the entire session")
+
+	// Add the eager module loading flag to disable lazy load on runtime.
+	flags.BoolVar(&eagerModuleLoading, "eager-mod-loading", false, "load module runtime eagerly")
```
eunomie: Why not something a bit simpler, like `--load-runtime`? Especially if the goal is to load the runtime.

eunomie (commented):
A part of me still think the purpose of `dagger functions` is not to ensure the runtime is loading, but to return the list of available functions. But let's add it if it solves the current issues.

TomChv:
> A part of me still think the purpose of `dagger functions` is not to ensure the runtime is loading, but to return the list of available functions. But let's add it if it solves the current issues.

The `eager-mod-loading` flag can be added to `call`, `install` or any module command.
I agree that maybe a specific command sounds like a better idea but it also feels weird to create a command simply to by-pass lazy loading.

> Why not something a bit simpler, like --load-runtime? Especially if the goal is to load the runtime.

@shykes suggested to do `eager-mod-loading` so it's more explicit but I'm open to any suggestion.

shykes:
- I think the flag should have the word "eager" or "lazy" so that its function is clear
- Generally, the shorter the better
- Beyond that, I don't know for sure the right answer :)

shykes:
Also it should be a global flag, not command-specific, eg. should work with all module commands. I'm guessing that's already the case?

TomChv:
> Also it should be a global flag, not command-specific, eg. should work with all module commands. I'm guessing that's already the case?

Yeah it's already applied for all module commands

> * I think the flag should have the word "eager" or "lazy" so that its function is clear
> * Generally, the shorter the better
> * Beyond that, I don't know for sure the right answer :)

Right now it's `eager-mod-loading`, I could also name it `no-lazy-runtime`, what do you prefer?

eunomie:
`--eager-runtime`? `--with-eager-runtime` to keep the common `With*` construction?

TomChv:
> `--eager-runtime`? `--with-eager-runtime` to keep the common `With*` construction?

PR updated in favor of `eager-runtime`

PR #11414, opened 2025-11-14
go: tidy, show diff in check-tidy

vito:


vito:
> If you have suggestions for addressing that, we could apply your pattern on `check-generated` also.

In various places (TUI, LLM) I've had to implement a 'patch summary' output that just lists the files and how many lines were added/removed. Maybe that should be promoted into the API?

shykes:
> > If you have suggestions for addressing that, we could apply your pattern on `check-generated` also.
> 
> In various places (TUI, LLM) I've had to implement a 'patch summary' output that just lists the files and how many lines were added/removed. Maybe that should be promoted into the API?

LOL that's exactly what I implemented for `check-generated` :) I think so yes

PR #11415, opened 2025-11-14
Scale-out support

sipsma:
Supports horizontal scaling of checks to cloud engines.
* After earlier experiments, decided to do scale-outs based on checks rather than any generic function call. This makes it easier to support checks that have dependencies on containers/dirs (e.g. our `go/checkTidy` check has a dep on a `Go` container from a parent object).
* The underlying mechanism and code is very similar to the function based approach (just simpler), so we aren't locking ourselves out of re-arranging or support generic function calls in the future

To use it:
* You need a special enabled `DAGGER_CLOUD_TOKEN` set
* You need either to wait for this PR to be released or to set the cloud engine image to a custom value (see notes below)
* Run checks with `dagger checks --scale-out` (new hidden flag)

So the scale-out is hidden and not really external user facing quite yet. Mostly want to get this merged for a release to ease our own further development and dog-fooding.

---

Technically adds a new API `Module.check` which allows running a single check with a name given by an arg. This is used in the implementation to scale out the check to another engine, but also felt reasonable enough to expose to users. If there's concerns I can just make it a hidden api for now cc @shykes @vito @kpenfound 

---

Bare minimum TODOs before merge:
- [x] Don't hardcode my custom engine image (being built off this branch)
   - just rm'd this for now. Not currently going to try making the engine image configurable since it raises questions about whether users should be able to do that (even in a hidden config way). If you need to customize the image for a dev engine build, you just have to locally change it on line 184 of internal/cloud/client.go
   - longer term fix might be for cloud to only allow that from allow-listed users/tokens? then we can safely make the image configurable from the client 
- [x] Figure out how to enable/disable this ability. Right now it just always scales out each check.
   - Added a hidden flag `dagger checks --scale-out` and an env var to support cloud usage of this `_EXPERIMENTAL_DAGGER_CHECKS_SCALE_OUT`
- [x] Rm lots of code leftover from previous experiment

### DIFF
--- a/core/checks.go
+++ b/core/checks.go
@@ -212,7 +212,7 @@ func (c *Check) Run(ctx context.Context) (*Check, error) {
 	}
 
 	var span trace.Span
-	if !clientMD.IsCloudScaleOutClient { // don't dupe telemetry if the client is an engine scaling out to us
+	if clientMD.CloudScaleOutEngineID != "" { // don't dupe telemetry if the client is an engine scaling out to us
 		ctx, span = Tracer(ctx).Start(ctx, c.Name(),
 			telemetry.Reveal(),
 			trace.WithAttributes(
--- a/engine/client/client.go
+++ b/engine/client/client.go
@@ -1353,6 +1353,11 @@ func (c *Client) clientMetadata() engine.ClientMetadata {
 		basicToken = os.Getenv("DAGGER_CLOUD_TOKEN")
 	}
 
+	var remoteEngineID string
+	if c.connector != nil {
+		remoteEngineID = c.connector.EngineID()
+	}
+
 	return engine.ClientMetadata{
 		ClientID:                  c.ID,
 		ClientVersion:             clientVersion,
@@ -1372,7 +1377,7 @@ func (c *Client) clientMetadata() engine.ClientMetadata {
 		EagerRuntime:              c.EagerRuntime,
 		CloudBasicAuthToken:       basicToken,
 		EnableCloudScaleOut:       c.EnableCloudScaleOut,
-		IsCloudScaleOutClient:     c.isCloudScaleOutClient,
+		CloudScaleOutEngineID:     remoteEngineID,
 	}
 }
 
--- a/engine/client/drivers/cloud.go
+++ b/engine/client/drivers/cloud.go
@@ -71,6 +71,10 @@ func (dc *DaggerCloudConnector) Connect(ctx context.Context) (net.Conn, error) {
 	return tlsConn, nil
 }
 
+func (dc *DaggerCloudConnector) EngineID() string {
+	return dc.EngineSpec.InstanceID
+}
+
 func (d *daggerCloudDriver) Available(ctx context.Context) (bool, error) {
 	return true, nil // assume always available
 }
--- a/engine/client/drivers/container.go
+++ b/engine/client/drivers/container.go
@@ -200,6 +200,11 @@ func (d containerConnector) Connect(ctx context.Context) (net.Conn, error) {
 	return d.backend.ContainerDial(context.WithoutCancel(ctx), d.host, args)
 }
 
+func (d containerConnector) EngineID() string {
+	// not supported yet
+	return ""
+}
+
 const (
 	// trim image digests to 16 characters to makeoutput more readable
 	hashLen             = 16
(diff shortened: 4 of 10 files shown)


cmd/engine/.dagger/main.go
```diff
@@ -232,11 +232,13 @@ var targets = []struct {
...
 }{
 	{
-		Name:      "alpine (default)",
-		Tag:       "%s",
-		Image:     DistroAlpine,
-		Platforms: []dagger.Platform{"linux/amd64", "linux/arm64"},
+		Name:  "alpine (default)",
+		Tag:   "%s",
+		Image: DistroAlpine,
+		// Platforms: []dagger.Platform{"linux/amd64", "linux/arm64"},
```
sipsma: note to self to rm before merge (makes publishing custom engine image tolerable)

core/checks.go
```diff
@@ -253,3 +195,210 @@ func (c *Check) Query() []dagql.Selector {
...
+			Arg("refString", filepath.Join(
+				c.Module.Source.Value.Self().Local.ContextDirectoryPath,
+				c.Module.Source.Value.Self().SourceRootSubpath,
+			))
+	case ModuleSourceKindGit:
+		query = query.Select("moduleSource").
+			Arg("refString", c.Module.Source.Value.Self().AsString()).
+			Arg("refPin", c.Module.Source.Value.Self().Git.Commit).
+			Arg("requireKind", c.Module.Source.Value.Self().Kind)
+	case ModuleSourceKindDir:
```
TomChv, line 347: In which case this may happen? If I remember well the `ModuleSourceKindDir` is when we do `Directory.AsModuleSource` so it should never lead to an actual execution no?

core/checks.go
```diff
@@ -253,3 +195,210 @@ func (c *Check) Query() []dagql.Selector {
...
+			Arg("sourceRootPath", c.Module.Source.Value.Self().DirSrc.OriginalSourceRootSubpath)
+	}
+	query = query.Select("asModule")
+
+	query = query.Select("check").
+		Arg("name", c.Name())
+
+	query = query.Select("run")
+
+	query = query.SelectMultiple("completed", "passed")
```
TomChv, line 369: So essentially we create a new client on a remote engine and call the module on that engine?
~~Can this work recursively for sub checks?~~ EDIT: I saw the todo in the engine/client.

core/checks.go
sipsma: You can run checks from a module created via `Directory.AsModuleSource`, e.g. `dagger core git --url https://github.com/dagger/dagger head tree as-module checks --include sdk/python/lint run list`

core/checks.go
sipsma: Yeah recursive calls actually do work, but I don't want to enable it until we have a need for it and we add detection that prevents infinite recursion, which would be a potentially expensive mistake 💸

sipsma:
> Can you explain why we need proxies on the sessions attachable for engine to engine client?
Are you using the grpcutil.ProxyStream to filesync, use git etc from the caller's host and send these to the remote engine session? Is it before the caller's engine/client only act as a proxy?

Yeah when engine A scales out to engine B, engine B may need things from the original client. Engine B asks its "client" (engine A) for a e.g. local dir, and engine A proxies back to the original client.

This won't scale forever with recursive checks since it'd require many hops to get back to the original client, but is a easy starting place for now.

> The code make sense but is quite complex to understand, if you can add a little bit more doc that may be helpful for maintainability.

👍

I have a local commit with more comments but github is 500'ing so I can't push yet. EDIT: pushed

core/checks.go
TomChv: But that's the same as doing `dagger -m github.com/dagger/dagger checks --include sdk/python/lint run list`?

sipsma: That example is but the point is it's valid to run checks from `Directory.AsModuleSource`. Our cloud needs to do this for example

sipsma:
> Code LGTM, I'm having issue trying to test the PR though.
> 
> What is the tag to put on internal/engine/client l176? I tried `registry.dagger.io/engine:main` and also a custom version pushed to tll from a dev engine but it does work.
> 
> I can see the remote engine being created on Cloud Engine but it doesn't success (timing out for TTL) and see screenshot for `main` test:
> Image
> 
> .

The engine needs to be built off this PR otherwise it won't have the features needed for scale out. I've just been publishing them off my branch. So you could try `eriksipsma/dagger-test:22`, but you'll also need a special `DAGGER_CLOUD_TOKEN`.

Here's an example run, you can see the scale-out step from `connect to cloud engine`: https://dagger.cloud/dagger/traces/a79ddc097f949005d0d47ed3c11920a0?listen=1e5d269162d513c9&listen=bf58379317a18d15

PR #11424, opened 2025-11-15
telemetry: use at most one open db per client

sipsma:
Previously, each client db could be opened multiple times, with one shared one for writing during the lifetime of the client and multiple other ones for reading (over the sse endpoints).

We also have been seeing occasional mysterious hangs where it becomes impossible to acquire a write lock on the DB, with `SQLITE_BUSY` errors returned and the engine hanging indefinitely trying to write.

There's a few potential interrelated explanations:
1. While working on this PR I found out that `clientLogs.OnEmit` sometimes gets called after the session has shutdown and the DB was closed. This seems to be possible when the main client does a sudden cancel while there's still some containers running; those containers try to flush after the DB was already closed. I'm not totally sure what was happening before this change in that scenario, but may easily have entered "undefined behavior" territory since db Close may have been called in midst of write attempts.
2. `SQLITE_BUSY` can be triggered for a number of reasons, including attempts to use the DB while in the midst of shutdown, during a recovery attempt, etc. https://www.sqlite.org/wal.html#sometimes_queries_return_sqlite_busy_in_wal_mode

The change here wraps the client db with ref counting and synchronization such that there is at most one db conn open at a time and used by both readers+writers. It's closed when no-one has an open ref to it and re-opened if accessed after that.

To help performance, a ref is consistently held open for the lifetime of the client (to prevent frequent close/opens).

### DIFF
--- a/engine/clientdb/dbs.go
+++ b/engine/clientdb/dbs.go
@@ -72,6 +72,11 @@ func (dbs *DBs) Open(ctx context.Context, clientID string) (_ *DB, rerr error) {
 		if err := os.MkdirAll(filepath.Dir(dbPath), 0700); err != nil {
 			return nil, fmt.Errorf("mkdir %s: %w", filepath.Dir(dbPath), err)
 		}
+
+		// check whether the file exists already
+		_, statErr := os.Lstat(dbPath)
+		alreadyExists := statErr == nil
+
 		connURL := &url.URL{
 			Scheme: "file",
 			Host:   "",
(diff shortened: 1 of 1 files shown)


TomChv (approved):

engine/clientdb/dbs.go
```diff
@@ -17,61 +19,130 @@ import (
...
+		sqlDB, err := sql.Open("sqlite", connURL.String())
+		if err != nil {
+			return nil, fmt.Errorf("open %s: %w", connURL, err)
+		}
+		if err := sqlDB.Ping(); err != nil {
+			return nil, fmt.Errorf("ping %s: %w", connURL, err)
+		}
+
+		db.inner = sqlDB
+		if _, err := db.inner.Exec(Schema); err != nil {
```
vito: nit: Seems like this will re-run the migrations when reopening an existing DB too, which is fine at the moment because they're idempotent, but may be a risk in the future. Could be worth a quick `os.Stat` just to be extra clear about the intention, but not blocking

sipsma: fixed!

PR #11429, opened 2025-11-17
fix: pin dang sdk in `cmd/codegen` to a pre-check commit.

TomChv:
The dang sdk is breaking the engine dev build because it is using unreleased API, we pin the dang SDK to a commit that work to let the engine build

Full incident at: https://discord.com/channels/707636530424053791/1003718839739105300/1439929385103654965

/cc @eunomie


eunomie (approved):
LGTM

Issue #10481, opened 2025-05-27
Split out SDK `Init`


jedevc:
We should revive the work in https://github.com/dagger/dagger/pull/8201 (was closed due to lack of time, but it's quite useful in our new SDK interface world that's being worked on by @TomChv as part of https://github.com/dagger/dagger/issues/9582).

This would help fix:
- #6621
- #7834
- #7707

It also feels heavily linked to #10480, and would help decouple `init` and `develop` more.

This new `Init` method would be in a new `Init` interface in `core/sdk.go`, something like:

```go
type Init interface {
	Init(
		context.Context,
		*ModDeps,
		dagql.Instance[*ModuleSource],
	) (*GeneratedCode, error)
}
```

Then, this would only be called during `dagger init` - this should allow simplifying the SDKs a bit, since they now don't need to work out whether to create template files themselves. Also, we could imagine an SDK like `dagger init --sdk=magic`, which would generate a README as to how to use it - there's no codegen, but it could still be `init`ed.

PR #11430, opened 2025-11-17
feat: add `Init` in SDK interface.

TomChv:
Expand the `SDK.CodeGenerator` interface to implement Init.
ModuleSource.GeneratedContextDirectory call `init` only when a new module is created or a new SDK is added: `dagger init --sdk=xx` or `dagger develop --sdk=xx`.

Implement `ModuleInit` in the TypeScript SDK. (Can be improved in the follow up)
Handle backward compatibility to fallback to `Codegen` if `moduleInit` isn't implemented by the SDK.

### :bulb: Implementation details

For now it's part of the CodeGenerator interface because it made sense to me but @jedevc mentioned in #10481 that
it could be part of its own interface to custom SDK could generate README on `init` even if they do not generate code.
The change to put `Init` in its own interface makes the code slightly more complex so I'm not sure if the use case is worth it.

Fixes: https://github.com/dagger/dagger/issues/10481

### DIFF
--- a/core/modulesource.go
+++ b/core/modulesource.go
@@ -211,7 +211,6 @@ func (src ModuleSource) Clone() *ModuleSource {
 
 	if src.SDK != nil {
 		src.SDK = src.SDK.Clone()
-		src.IsInit = src.IsInit
 	}
 
 	origIncludePaths := src.IncludePaths


eunomie:
I think I would have done it differently. In this version init sort of replaces codegen, but only for the init. Init is like a codegen-like dedicated for commands with --sdk if I understand it well.

I would have completely split both responsibilities:
- the moduleInit that creates files based on the template as defined in the SDK
- the codegen that generates the required code

If you have `moduleInit` or `codegen` depending on the cases (in init or not) it means you have to put that logic inside the SDK. If you split both, this is up to the engine to deal with them. On the SDK side it means each function has a very clear goal and less side effects.

For instance you can have a flow where:
- `moduleInit` will return a Changeset or a Directory
- the result will be injected in the `moduleSource`
- `codegen` is called based on the modified `moduleSource`

In this case the init phase is really small and clear. And the codegen phase always assume the required files exist (the ones created either by the init or by the user).

PR #11434, opened 2025-11-18
Even better errors

vito:
(Building on #8442 ...)

Rips out the elegant but brittle + high-maintenance error-origin-via-extended-errors approach, in favor of appending metadata to error strings and regexing it back out on the frontend.

It's bit hokey, but it will work everywhere. The previous approach required coordination in each SDK and at every layer (Go SDK, TypeScript SDK, engine calling functions, functions calling API, parallelism, [...]) to support tracking error origins as extended fields and propagating them across boundaries.

Supporting multiple origins is now trivial. Just concatenate all the error strings, like you might already be doing (`errors.Join`).

### Along the way...:

* Clean up the super verbose error message preceding `exit code: 1`.
* Remove a few pointless + verbose error wrapping hotspots.

### DIFF
--- a/dagql/idtui/frontend_pretty.go
+++ b/dagql/idtui/frontend_pretty.go
@@ -747,7 +747,7 @@ func (fe *frontendPretty) Render(out TermOutput) error {
 	}
 
 	below := new(strings.Builder)
-	if logs := fe.logs.Logs[fe.ZoomedSpan]; logs != nil && logs.UsedHeight() > 0 {
+	if logs := fe.logs.Logs[fe.ZoomedSpan]; logs != nil && logs.UsedHeight() > 0 && !fe.hasShownRootError() {
 		logs.SetHeight(fe.window.Height / 3)
 		logs.SetPrefix(progPrefix)
 		fmt.Fprint(below, logs.View())
@@ -1976,10 +1976,11 @@ func (fe *frontendPretty) renderRow(out TermOutput, r *renderer, row *dagui.Trac
 			}
 		}
 	}
-	if row.ShouldShowCause() {
+	if len(row.Span.ErrorOrigins.Order) > 0 && (!row.Expanded || !row.HasChildren) {
 		multi := len(row.Span.ErrorOrigins.Order) > 1
 		for _, cause := range row.Span.ErrorOrigins.Order {
 			if multi {
+				r.fancyIndent(out, row, false, false)
 				fmt.Fprintln(out, prefix)
 			}
 			fe.renderErrorCause(out, r, row, prefix, cause)
@@ -2114,15 +2115,14 @@ func (fe *frontendPretty) renderErrorCause(out TermOutput, r *renderer, row *dag
 	}
 
 	if len(parents) > 0 {
-		// fmt.Fprint(out, prefix, indent)
 		r.fancyIndent(out, row, false, false)
-		fmt.Fprint(out, "  ")
 		if !fe.finalRender {
 			fmt.Fprint(out, "  ")
 		}
 		slices.Reverse(parents)
 		context := new(strings.Builder)
 		noColorOut := termenv.NewOutput(context, termenv.WithProfile(termenv.Ascii))
+		fmt.Fprint(noColorOut, VertBoldDash3+" ")
 		for _, p := range parents {
 			fe.renderStepTitle(noColorOut, r, p, prefix+indent, true)
 			fmt.Fprintf(noColorOut, " › ")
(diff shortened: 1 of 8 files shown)


dagql/idtui/frontend_pretty.go
```diff
@@ -1874,10 +1874,10 @@ func (fe *frontendPretty) goErrorOrigin() {
 	if focused == nil {
 		return
 	}
-	if focused.ErrorOrigin == nil {
+	if len(focused.ErrorOrigins.Order) == 0 {
 		return
 	}
-	fe.FocusedSpan = focused.ErrorOrigin.ID
+	fe.FocusedSpan = focused.ErrorOrigins.Order[0].ID // TODO which?
```
sipsma, line 1880: is the problem just choosing which error to focus on when there's multiple? if so, the first one seems like as good an option as any 🙂

vito: Yeah that's the only issue. May be a sign that there's more work to be done for the UI to catch up, but figure we can save that for another iteration

vito:
@TomChv Ah yeah those aren't meant to be visible. Do you have repro for that? I can fix it up.

That's the main downside of the new approach: because it's string munging, those tracking tags can end up in surprising places. But, the upside is also they end up everywhere, without any effort from the intermediate parties (SDKs etc)! So we just need to hide them from the UI, and use them for origin tracking, and the net result _should_ be much quieter and more polished error output with 'go-to-error' etc.

The first screenshot is a spot I already know about, but the second screenshot is a little surprising since those should be showing up as `ERROR r jump -v` instead of duplicating the error message. Was that ran with the CLI from this PR?

TomChv:
> @TomChv Ah yeah those aren't meant to be visible. Do you have repro for that? I can fix it up.
> 
> That's the main downside of the new approach: because it's string munging, those tracking tags can end up in surprising places. But, the upside is also they end up everywhere, without any effort from the intermediate parties (SDKs etc)! So we just need to hide them from the UI, and use them for origin tracking, and the net result _should_ be much quieter and more polished error output with 'go-to-error' etc.
> 
> The first screenshot is a spot I already know about, but the second screenshot is a little surprising since those should be showing up as `ERROR r jump -v` instead of duplicating the error message. Was that ran with the CLI from this PR?

Yeah if was running it using a dev engine of this PR. My repro code is the following:

```go
package main

import (
	"context"
	"errors"

	"go.opentelemetry.io/otel/codes"
	"golang.org/x/sync/errgroup"
)

type Test struct{}

func (m *Test) ErrSpan() error {
	return errors.New("hello world")
}

func (m *Test) MultiErrSpan(ctx context.Context) error {
	eg, gctx := errgroup.WithContext(ctx)
	
	eg.Go(func () error {
		rctx, span := Tracer().Start(gctx, "hello 1")
		defer span.End()
		
		_, err := dag.Container().From("alpine").WithExec([]string{"echo", "hello world"}).WithExec([]string{"exit", "2"}).Sync(rctx)
		if err != nil {
			span.RecordError(err)
			span.SetStatus(codes.Error, err.Error())
		}
		return err
	})
	
	eg.Go(func () error {
		rctx, span := Tracer().Start(gctx, "hello 2")
		defer span.End()
		
		_, err := dag.Container().From("alpine").WithExec([]string{"echo", "hello world 2"}).WithExec([]string{"exit", "1"}).Sync(rctx)
		if err != nil {
			span.RecordError(err)
			span.SetStatus(codes.Error, err.Error())
		}
		return err
	})
	
	return eg.Wait()
}
```

vito:
@TomChv Thanks for the repro.

Here's what I see:

image

If I run it with `-E` and expand `multiErrSpan`, I see this:

image

Which has some things to tune up, but still looks different from your screenshot. :thinking:

TomChv:
Screenshot 2025-11-19 at 16 52 31

I think it may be flacky because here I don't see any trace ID with a dev engine

vito:
@TomChv Hmm something's wonky with your setup I think - that's definitely not hitting a dev engine from this PR at least, since it's still showing the `process "exit 2" did not complete successfully: ` prefix that this PR removes.

TomChv:
Yeah I noticed that, I think having 2 dev engine running in parallel is causing some issue, I'll clean this up and try again

TomChv:
Screenshot 2025-11-19 at 16 59 43

Ok I rebuild a correct dev engine this time and yeah I can still see these parent trace in logs.

I can also see it on cloud: https://dagger.cloud/Quartz/traces/9365282b6a3e7b17cb03c7802ca977a5?showHidden=9709600a007bea28

vito:
Cloud hasn't been updated yet so that makes sense.

Still don't trust your CLI though. :skull: Does it repro if you `go build -o ./bin ./cmd/dagger` and run `./bin/dagger`?

TomChv:
> Cloud hasn't been updated yet so that makes sense.
> 
> Still don't trust your CLI though. 💀 Does it repro if you `go build -o ./bin ./cmd/dagger` and run `./bin/dagger`?

Ok I think there was something weird happening with the symlink of my dev CLI.
I had different result between the direct binary and my symlink, I updated it and now I got the same behaviour as you got.

So yeah, parentTrace still appears when running -E but that's all.

vito:
Pushed some more cleanup and tests, the `traceparent:...` no longer shows up in `-E` either. Merging!

PR #11437, opened 2025-11-18
Do not use external programs to ask for Git credentials

sleiner-v:
This patch disables the triggering of (unwanted AFAICT) credential prompts via the Dagger CLI.

One notable case where this happens is Visual Studio Code, which sets `GIT_ASKPASS` such that the user is prompted for credentials via the VS Code GUI. In any case where dagger was started from a VS Code shell and needed to access Git repositories (e.g. because of module dependencies), the user would be prompted via the GUI. This would happen even if credentials were otherwise available (e.g. via `.netrc`).

### DIFF
--- a/.changes/unreleased/Fixed-20251203-125612.yaml
+++ b/.changes/unreleased/Fixed-20251203-125612.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Disable Git askpass prompts to prevent unexpected IDE credential dialogs (e.g., VS Code popups that appear even when credentials are already configured). Credentials still work via helpers or .netrc. For the rare case where no credentials are configured, Dagger now fails early with a clear error instead of triggering a subtle prompt that users sometimes miss.
+time: 2025-12-03T12:56:12.87879947-08:00
+custom:
+    Author: sleiner-v
+    PR: "11437"


engine/session/git/git_credential.go
```diff
@@ -78,12 +78,11 @@ func (s GitAttachable) getCredentialFromHelper(ctx context.Context, req *GitCred
...
 	cmd.Stdin = strings.NewReader(input)
 
+	// Never prompt the user for credentials
 	cmd.Env = append(os.Environ(),
-		"GIT_TERMINAL_PROMPT=0",
+		"GIT_ASKPASS=",          // Do not use external programs to ask for credentials
+		"GIT_TERMINAL_PROMPT=0", // Disable Git's builtin prompting
 	)
-	if req.Protocol != "http" && req.Protocol != "https" {
-		cmd.Env = append(cmd.Env, "SSH_ASKPASS=echo")
```
sleiner-v, line 85: I have removed the `SSH_ASKPASS` override, since `SSH_ASKPASS` is not used by Git if `GIT_ASKPASS` is already set.

sleiner-v:
Hi @grouville,
just to make sure that we don't both wait on each other: Should I currently do anything for this MR?
Thanks in advance! :-)

grouville:
Hey @sleiner-v, I have 2 more commits to add on your PR (adding tests and improving some comments + adding a changelog), can you please allow push from maintainers on your fork ? 

```shell
gh pr view 11437 --json maintainerCanModify)
  ⎿  {                             
       "maintainerCanModify": false
     }
```

Overall I like it 🙏 ! But as it changes the behavior to the end-user (the fallback mechanism _could_ be useful to some users, it's worth documenting it well)

grouville (changes requested):
please allow me to push on your branch, to add the corresponding changelog + a test and some comments explaining the expected behavior

sleiner-v:
Hi @grouville,

thank you very much! :-) I have now given you the push privileges that you asked for.

PR #11448, opened 2025-11-19
engine: handle stale sessions in dag-op filesync too

sipsma:
[Previously](https://github.com/dagger/dagger/pull/11295), we added a fix for stale sessions showing up in dag-op due to per-LLB vertex deduplication in `Server.clientFromIDs`.

There's another codepath possibly impacted by the same problem when trying to get a session from the buildkit client. So we add the same fix there too.

This theoretically may be the cause of some flakes we occasionally hit with errors like:
```
failed to stat path: failed to get requester session: session for "XYZ" not found
```
* [example trace](https://dagger.cloud/dagger/traces/93301f52a469301d583232c8a4a110fe?span=5f287e2294793162)

---

Will give this a bunch of re-runs to see if that flake pops up again

### DIFF
--- a/engine/buildkit/client.go
+++ b/engine/buildkit/client.go
@@ -30,6 +30,7 @@ import (
 	solverresult "github.com/dagger/dagger/internal/buildkit/solver/result"
 	"github.com/dagger/dagger/internal/buildkit/util/bklog"
 	"github.com/dagger/dagger/internal/buildkit/util/entitlements"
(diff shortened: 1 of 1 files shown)


sipsma:
Have not seen the flake pop up, but have seen other flakes pop up quite a bit (seemingly more than before). I'll put this in draft for now

EDIT: these were other issues that have since been resolved

vito (approved):

PR #11453, opened 2025-11-20
refactor labels to provide an env getter

marcosnils:
this enables retrieving envs from different sources and not only
process env variables. This is particularly useful so we can use this
package from other parts which can inject labels from other sources


marcosnils:
nvm, we don't need this for now

tiborvass (approved):

PR #11456, opened 2025-11-20
core: fix error from cache hit on same git repo with different auth

sipsma:
See the added integ test for the case being fixed here.

In retrospect erroring out when transfering secrets/sockets from one client to another was unnecessarily zealous since in the worst case an attempt to use those by the client would just result in an error eventually.

In the current dagql caching system, it's possible for content-digested resources like git repo trees to end up in a client's call ID with auth sourced from an earlier client making the same call. In this case, even though the specific auth used to pull the git repo may differ between the two, it doesn't matter because both clients have already "proven" access to the repo by having constructed the tree already.

It's also not possible for the specific token value to leak between the two because the token's secret name is also content hashed, so to retrieve its plaintext you'd need to know its plaintext already.

### DIFF
--- a/.changes/unreleased/Fixed-20251120-122452.yaml
+++ b/.changes/unreleased/Fixed-20251120-122452.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: Fix "failed to add secret" errors when concurrently using a private git repo
+  from multiple clients that each have a different auth token value.
+time: 2025-11-20T12:24:52.320500703-08:00
+custom:
+  Author: sipsma
+  PR: "11456"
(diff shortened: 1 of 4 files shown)


grouville (approved):
🙏  [Context, to keep track](https://discord.com/channels/707636530424053791/1426220130102411384)

Just needs a rebase ?

grouville:
[Flake](https://dagger.cloud/dagger/traces/ff333f569962d036e45d232a25d97d8f) 👀 ? 
image

tiborvass:
LGTM, the test is properly failing without the fix.

PR #11458, opened 2025-11-20
engine: fix nested execs created by modules to not be module clients

sipsma:
There was some vestigial looking code when creating execs that resulted in nested execs created by modules to be marked as being clients for that module. This is subtly incorrect since they are not actually module clients (i.e. clients for function calls of the module), they are just execs created by the module.

This broke some cases of trying to load further modules from these nested execs because filesync tries to find the "non-module parent client", and would thus skip over the nested exec client up to the main caller.

---

reported by a user here: https://discord.com/channels/707636530424053791/1440182254126108763/1440182254126108763

### DIFF
--- a/engine/server/session.go
+++ b/engine/server/session.go
@@ -1348,6 +1348,24 @@ func (srv *Server) CurrentModule(ctx context.Context) (*core.Module, error) {
 	return nil, core.ErrNoCurrentModule
 }
 
+// If the current client is a module client or a client created by a module function, returns that module.
+func (srv *Server) ModuleParent(ctx context.Context) (*core.Module, error) {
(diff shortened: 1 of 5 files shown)


engine/server/session.go
```diff
@@ -1348,6 +1348,24 @@ func (srv *Server) CurrentModule(ctx context.Context) (*core.Module, error) {
...
+	}
+	if client.mod != nil {
+		return client.mod, nil
+	}
+	for i := len(client.parents) - 1; i >= 0; i-- {
+		parent := client.parents[i]
+		if parent.mod != nil {
+			return parent.mod, nil
+		}
+	}
```
vito, line 1365: neat

PR #11464, opened 2025-11-21
fix another set of old 0.19.6 strings

sipsma:
In my last attempt I grep'd for 0.19.6 and fixed those, but that missed places where the string in question is 0-19-6....

Didn't notice in the PR because publish only runs on main. So one more bump.


tiborvass (approved):
this is so brittle :( thanks!

PR #11466, opened 2025-11-21
[DNM] attempt to repro hang during lint/tidy

sipsma:
Seeing if I can repro the occasional hang we see (usually accompanied by SQLITE BUSY timeout errors) in CI to help get to the bottom of it. Can't repro locally...


sipsma:
Successfully figured out how to repro here, seems to be a modernc bug when context is canceled in the middle of a write... will keep pursuing in separate PRs

PR #11471, opened 2025-11-23
Vault: add non-standard approle auth method

### LINKED ISSUE #11468: Authentication to Vault via a non-standard approle auth method

I’m trying to authenticate my Dagger pipeline to HashiCorp Vault using an AppRole-based workflow, but not via the standard `approle` auth method path. Instead, I need to authenticate through a custom or non-default auth mount (for example, a renamed or namespaced AppRole mount such as `auth/custom-approle/login`).

This matters because my Vault environment does not expose the default `approle` mount.

Right now, I’m manually constructing API calls to Vault inside the Dagger pipeline, explicitly pointing to the custom AppRole mount (e.g., `auth/custom-approle/login`).

nilune:
Added `VAULT_APPROLE_MOUNT_PATH` env variable for authenticating to non-standard approle auth method in Vault

Resolves #11468

### DIFF
--- a/docs/current_docs/extending/modules/arguments.mdx
+++ b/docs/current_docs/extending/modules/arguments.mdx
@@ -844,7 +844,7 @@ dagger call github-api --token=op://infra/github/credential
 Vault can be authenticated with either token or AppRole methods.
 The Vault host can be specified by setting the environment variable `VAULT_ADDR`.
 For token authentication, set the environment variable `VAULT_TOKEN`.
-For AppRole authentication, set the environment variables `VAULT_APPROLE_ROLE_ID` and `VAULT_APPROLE_SECRET_ID` (and optionally the variable `VAULT_APPROLE_AUTH_METHOD`, otherwise the default value will be used - `approle`).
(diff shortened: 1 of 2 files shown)


marcosnils:
WDYT about renaming the variable from `VAULT_APPROLE_AUTH_METHOD` to `VAULT_APPROLE_MOUNT_PATH`. Given that we're ultimately calling the `WithMountPath` config in the API, it seems more consistent to use the same naming for the env variable.

nilune:
> WDYT about renaming the variable from `VAULT_APPROLE_AUTH_METHOD` to `VAULT_APPROLE_MOUNT_PATH`. Given that we're ultimately calling the `WithMountPath` config in the API, it seems more consistent to use the same naming for the env variable.

Given my lack of open-source experience and limited Go experience, I’m inclined to trust you :)
And yes, it does sound logical overall, so I renamed it.
Indeed, in both the Go and Python clients it’s called mountPath.

Updated PR.

marcosnils (approved):

PR #11480, opened 2025-11-24
configure toolchains with toolchain customization

kpenfound:
Follow up to https://github.com/dagger/dagger/pull/11373

New schema:

```json
{
  "name": "app",
  "engineVersion": "v0.19.4",
  "toolchains": [
    {
      "name": "hello",
      "source": "../hello",
      "customization": [
        {
          "function": ["configurableMessage"],
          "argument": "message",
          "default": "hola"
        }
      ]
    }
  ]
}
```

Old schema:

```json
{
  "name": "app",
  "engineVersion": "v0.19.4",
  "toolchains": [
    {
      "name": "hello",
      "source": "../hello",
      "arguments": [
        {
          "function": ["configurableMessage"],
          "name": "message",
          "default": "hola"
        }
      ]
    }
  ]
}
```

### DIFF
--- a/core/schema/modulesource.go
+++ b/core/schema/modulesource.go
@@ -2295,6 +2295,7 @@ func (s *moduleSourceSchema) loadModuleSourceConfig(
 
 		// TODO: this is for backwards compatibility until the configuration change is released
 		if i < len(src.ConfigDependencies) && src.ConfigDependencies[i] != nil {
+			//nolint:staticcheck
 			depCfg.Arguments = src.ConfigDependencies[i].Arguments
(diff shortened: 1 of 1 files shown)


shykes:
@kpenfound looks good, I would just call the field `customizations` plural, no?

Also, thinking ahead, will this allow us to also disable checks like we were discussing separately? The only tricky part I see is that `argument` and eg. `noCheck` would be mutually exclusive.

kpenfound:
@shykes yeah i was considering `customizations.ignoreChecks: []` or something but I'll keep that separate so we can bikeshed and get this small change in

PR #11482, opened 2025-11-24
ci: centralize published images around wolfi

### LINKED ISSUE #5668: Standardize on a single base OS for engine release

sipsma:
[We were publishing engine images for](https://github.com/dagger/dagger/pkgs/container/engine):
1. (default) alpine
2. (`-gpu`) ubuntu
3. (`-wolfi`) wolfi w/out gpu
4. (`-wolfi-gpu`) wolfi w/ gpu

This change simplifies everything to be just:
1. (default) wolfi w/out gpu
2. (`-gpu`) wolfi w/ gpu

AFAIK all the images had accumulated over time as we wanted to start trying out support for different base images but didn't want to change defaults yet. It seems like its now worth consolidating.

This will also greatly simplify adding support for CGo builds of the engine since there's only one libc variant/version to deal with rather than 3.

---

Fixes #5668

### DIFF
--- a/core/integration/proxy_test.go
+++ b/core/integration/proxy_test.go
@@ -202,8 +202,7 @@ redirect ^(https?://)(.*).example(/.*)$		$1$2$3
 
 		devEngine := devEngineContainer(c, func(ctr *dagger.Container) *dagger.Container {
 			return ctr.
-				// go right to /etc/ssl/certs to avoid testing the custom CA cert support (covered elsewhere)
-				WithMountedFile("/etc/ssl/certs/myCA.pem", certGen.caRootCert).
+				WithMountedFile("/usr/local/share/ca-certificates/myCA.pem", certGen.caRootCert).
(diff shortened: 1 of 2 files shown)


shykes:
Let's do it!

shykes:
Relevant: #5668

PR #11484, opened 2025-11-25
switch to mattn sqlite lib

sipsma:
modernc seems to have a bug where a context cancellation in the middle of a DB write can cause all future write attempts on the DB to hang for `busy_timeout` and then fail. Did not fully diagnose but feels like it may end up leaving a transaction open in that case (despite our careful handling of `defer` rollback/commit).

Switching to `mattn` fixes it without any other changes. The main downside is that it requires we build the engine w/ CGo. This change does C cross-compilation w/ zig since it's relatively easy to configure different targets, has built-in caching, etc.
* [I did see a warning](https://news.ycombinator.com/item?id=43505646) that someone doing exactly this (compile sqlite cgo w/ zig) got poor performance, but that was a few years back and I have not noticed any obvious impact in terms of CI runtimes. Worth keeping an eye on though.

This does make engine builds slower, but based on earlier testing in the "rebuild engine after code change" case it was only ~4s worse, which is not too bad.


sipsma:
@tiborvass pointed out that there was just in the last week a fix pushed to modernc for this: https://gitlab.com/cznic/sqlite/-/merge_requests/81

Sadly it was a day or two after I had previously checked and upgraded modernc 😢

I'll just upgrade us to that in a separate PR for now. I'm gonna check and see if there was any perceivable benefit from switching to `mattn` before closing this though; I did see a run of `TestTelemetry` where the test itself was >1m faster than the most recent one on `main`, but our CI is extremely noisy so unclear if that's real or not.

Moving back to draft for now

sipsma:
Trying stuff locally, I don't see any perf difference from mattn, so I'll close this out for now.

PR #11485, opened 2025-11-25
Extract java and python images in Dockerfiles

eunomie:
This allow to remove duplication (between the toolchains dev module and the sdk) and let dependabot to deal with the updates as it's a Dockerfile.
One Dockerfile per image. It's still possible to parse the Dockerfile, but I'd like to remove it at some point.
But in the meantime we can build from it, without to know and deal with the docker ref, so that's easier.

Done for python and java

### DIFF
--- a/toolchains/python-sdk-dev/dagger.json
+++ b/toolchains/python-sdk-dev/dagger.json
@@ -9,7 +9,7 @@
     "../../go.mod",
     "../../util/parallel",
     "../../util/dockerfileextractor",
-    "../../sdk/python/runtime/Dockerfile"
+    "../../sdk/python/runtime/images"
   ],
   "dependencies": [
     {
--- a/toolchains/python-sdk-dev/main.go
+++ b/toolchains/python-sdk-dev/main.go
@@ -35,7 +35,7 @@ func New(
 	//   "!sdk/python/dev/src/**/*.py",
 	//   "!sdk/python/docs/**/*.py",
 	//   "!sdk/python/docs/**/*.rst",
-	//   "!sdk/python/runtime/Dockerfile",
+	//   "!sdk/python/runtime/images",
 	//   "!sdk/python/src/**/*.py",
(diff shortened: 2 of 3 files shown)


sdk/java/runtime/main.go
```diff
@@ -464,3 +455,17 @@ func (m *JavaSdk) mavenCommand(args ...string) []string {
 	args = append(args, "--no-transfer-progress")
 	return args
 }
+
+//go:embed images/maven/Dockerfile
+var mavenImage string
+
+func (m *JavaSdk) MavenImage() *dagger.Container {
```
TomChv, line 462: Good, normally it shouldn't impact performances too much since you are embedding the file instead of accessing from `CurrentModule().Source()`

sdk/python/runtime/image.go
```diff
@@ -104,26 +107,28 @@ func NewImage(ref string) (Image, error) {
...
-	lines := strings.Split(dockerfile, "\n")
 	images := make(map[string]Image)
+	for _, dockerfile := range []string{baseDockerfile, uvDockerfile} {
+		lines := strings.Split(dockerfile, "\n")
 
-	for _, line := range lines {
-		if matches := fromLineRegex.FindStringSubmatch(strings.TrimSpace(line)); matches != nil {
-			ref := matches[1]
-			name := matches[2]
+		for _, line := range lines {
```
TomChv, line 114: Is this still useful? You only have 1 line in each `Dockerfile`

eunomie: yeah, I just wanted to keep it in case something changed. Just to be flexible enough. For instance if, as before, there was a comment in the file.

eunomie:
> Changes looks good, I'm happy to give a ✅ once the CI is green

There was one bug, I've fixed it and CI is happier now.
There's still some timeouts, but they are not related to those changes.
Let me know if that's good for you.

PR #11486, opened 2025-11-25
feat: typescript SDK add live telemetry

TomChv:
Use `LiveSpanProcessor` if live trace is enabled in ts sdk telemetry.

### DIFF
--- a/sdk/typescript/src/telemetry/live_processor.ts
+++ b/sdk/typescript/src/telemetry/live_processor.ts
@@ -1,0 +1,15 @@
+import type { Context } from "@opentelemetry/api"
+import { BatchSpanProcessor, type Span } from "@opentelemetry/sdk-trace-base"
+
+/**
+ * Live span processor implementation.
+ *
+ * It's a BatchSpanProcessor whose on_start calls on_end on the underlying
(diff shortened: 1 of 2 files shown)


sdk/typescript/src/telemetry/live_processor.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Live span processor implementation.
+ *
+ * It's a BatchSpanProcessor whose on_start calls on_end on the underlying
+ * SpanProcessor in order to send live telemetry.
+ */
+export class LiveProcessor extends BatchSpanProcessor {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  override onStart(_span: Span, _parentContext: Context): void {
+    this.onEnd(_span)
```
eunomie, line 13: Just to be sure to understand, the idea is the `onEnd` is sending the trace, right? So by calling `onEnd` during `onStart` it forces to send them right away?

This reminds me `mvn clean` to build something in the java world 🫣

TomChv: Exactly!

PR #11488, opened 2025-11-25
fix(cache): do not cache function errors in a single dagger session

### LINKED ISSUE #11465: 🐞 Dagger shell caches @functions even when exceptions are raised
### What is the issue?

Given the following module and function

```python
import dagger
import random
from dagger import dag, function, object_type


@object_type
class Test:
    @function
    async def test_cache(self, low: int, high: float, limit: float) -> float:

        data =  random.uniform( low,high)
        if data < limit: 
            # if i hit this i dont want caching to occu
            raise ValueError( f"{data} < {limit}")
        return data
```

**The behavior i expect to see:**

1. If a random number is drawn that is less than the limit an exception is thrown, no caching occurs. 
2. If a random number above the limit is drawn then the function returns and the return value is cached 

The result is I can call this function multiple times and it will rerun until it finds a value greater than the limit - then it will just return the cached value.


**Here are the results using the dagger cli - which works as expected:**


```shell
dagger call test-cache --low 0 --high 10 --limit 6
▶ connect 0.4s
▶ load module: . 1.0s

✔ test: Test! 0.0s
▼ .testCache(low: 0, high: 10.000000, limit: 6.000000): Float! 1.8s ERROR
┇ ...1 lines hidden...
┃ Traceback (most recent call last):                                                                                                                                                                                                                                                         
┃   File "/src/xxh3:6dc10608b1b5157d/test/sdk/src/dagger/mod/_module.py", line 430, in call                                                                                                                                                                                                  
┃     result = await cast(typing.Awaitable[R], result)                                                                                                                                                                                                                                       
┃              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                       
┃   File "/src/xxh3:6dc10608b1b5157d/test/src/test/main.py", line 14, in test_cache                                                                                                                                                                                                          
┃     raise ValueError( f"{data} < {limit}")                                                                                                                                                                                                                                                 
┃ ValueError: 5.153373527786046 < 6.0                                                                                                                                                                                                                                                        
! FunctionError: 5.153373527786046 < 6.0

Error logs:

▼ .testCache(low: 0, high: 10.000000, limit: 6.000000): Float! 1.8s ERROR
ERROR [dagger.mod]: Unhandled exception while executing function                                                                                                                                                                                                                             
Traceback (most recent call last):                                                                                                                                                                                                                                                           
  File "/src/xxh3:6dc10608b1b5157d/test/sdk/src/dagger/mod/_module.py", line 430, in call                                                                                                                                                                                                    
    result = await cast(typing.Awaitable[R], result)                                                                                                                                                                                                                                         
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                         
  File "/src/xxh3:6dc10608b1b5157d/test/src/test/main.py", line 14, in test_cache                                                                                                                                                                                                            
    raise ValueError( f"{data} < {limit}")                                                                                                                                                                                                                                                   
ValueError: 5.153373527786046 < 6.0                                                                                                                                                                                                                                                          
! FunctionError: 5.153373527786046 < 6.0

A new release of dagger is available: v0.19.6 → v0.19.7
https://github.com/dagger/dagger/releases/tag/v0.19.7
(base) mike@mike-dev:~/AutomationExample/test$ dagger call test-cache --low 0 --high 10 --limit 6
▶ connect 0.4s
▶ load module: . 0.8s

✔ test: Test! 0.0s
▶ .testCache(low: 0, high: 10.000000, limit: 6.000000): Float! 1.8s

7.555781637769714

A new release of dagger is available: v0.19.6 → v0.19.7
https://github.com/dagger/dagger/releases/tag/v0.19.7
(base) mike@mike-dev:~/AutomationExample/test$ 
(base) mike@mike-dev:~/AutomationExample/test$ dagger call test-cache --low 0 --high 10 --limit 6
▶ connect 0.4s
▶ load module: . 0.9s

✔ test: Test! 0.0s
▶ .testCache(low: 0, high: 10.000000, limit: 6.000000): Float! 0.0s

7.555781637769714

A new release of dagger is available: v0.19.6 → v0.19.7
https://github.com/dagger/dagger/releases/tag/v0.19.7

```



**Here are the results using the dagger shell - which caches failures:**

```shell
 dagger
Dagger interactive shell. Type ".help" for more information. Press Ctrl+D to exit.

▼ test | test-cache 0 10 9 1.9s ERROR r jump ↴
  ▼ Test.testCache(low: 0, high: 10.000000, limit: 9.000000): Float! 1.9s ERROR
  ┇ ...1 lines hidden...
  ┃ Traceback (most recent call last):                                                                                                                                                                                                                                                       
  ┃   File "/src/xxh3:6dc10608b1b5157d/test/sdk/src/dagger/mod/_module.py", line 430, in call                                                                                                                                                                                                
  ┃     result = await cast(typing.Awaitable[R], result)                                                                                                                                                                                                                                     
  ┃              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                     
  ┃   File "/src/xxh3:6dc10608b1b5157d/test/src/test/main.py", line 14, in test_cache                                                                                                                                                                                                        
  ┃     raise ValueError( f"{data} < {limit}")                                                                                                                                                                                                                                               
  ┃ ValueError: 2.7604536189960305 < 9.0                                                                                                                                                                                                                                                     
  ! FunctionError: 2.7604536189960305 < 9.0
▼ test | test-cache 0 10 9 0.0s ERROR r jump ↴
  ▼ Test.testCache(low: 0, high: 10.000000, limit: 9.000000): Float! 0.0s ERROR
  ! FunctionError: 2.7604536189960305 < 9.0
  ├╴$ Module.runtime: Container! 0.0s CACHED
  ├╴$ directory: Directory! 0.0s CACHED
  ╰╴$ Container.withMountedDirectory(
      ┆ path: "/.daggermod"
      ┆ source: directory: Directory!
      ): Container! 0.0s CACHED

test ⋈                                                       
```

grouville:
Fixes #11465

### Problem

In a long-lived engine session (interactive shell), function calls that previously failed were incorrectly replayed when invoked again with the same inputs.

The correct behavior should be:
	•	Failed calls should never be cached.
	•	Subsequent identical calls should rerun and, upon success, then be cached.

This issue did not affect the Dagger CLI (dagger call) because each invocation runs with a fresh session.

###  Solution

To resolve this, after a forced `DoNotCache=true` call successfully completes, we explicitly re-insert the successful result into the cache using the original key. This ensures subsequent calls correctly retrieve the successful outcome from the cache.

A boolean flag, `forcedDoNotCache`, was introduced to clearly distinguish cases:
- Set internally due to previous errors (eligible for recaching).
- Explicitly set by callers for intentionally non-cacheable operations (ineligible for recaching).

This fix effectively addresses the caching issue without needing the more complex approach of directly invalidating BuildKit’s cache entries.

## Manual repro

Given this test module: 

```shell
import dagger
import random
from dagger import dag, function, object_type


@object_type
class Test:
    @function(cache="session") 
    async def test_cache(self, low: int, high: float, limit: float) -> float:

        data =  random.uniform( low,high)
        if data < limit: 
            # if i hit this i dont want caching to occu
            raise ValueError( f"{data} < {limit}")
        return data
```

You can run, in a single session, these commands: `test | test-cache 0 10 4` (you can increase the `4` to `8` or `9`, so that the errors are more common) all the failed runs shall rerun and once a successful run is hit, it should be cached for th e next runs

### DIFF
--- a/dagql/session_cache.go
+++ b/dagql/session_cache.go
@@ -28,6 +28,9 @@ type SessionCache struct {
 	isClosed bool
 
 	seenKeys sync.Map
+
+	// noCacheNext keeps track of keys for which the next cache attempt should bypass the cache.
+	noCacheNext sync.Map
 }
 
 func NewSessionCache(
@@ -162,9 +165,35 @@ func (c *SessionCache) GetOrInitializeWithCallbacks(
 		ctx = telemetryCtx
 	}
 
+	// If this callKey previously failed, force the next attempt to be DoNotCache
+	// to bypass any potentially stale cached error.
+	forcedDoNotCache := false
+	if _, ok := c.noCacheNext.Load(key.CallKey); ok && !key.DoNotCache {
+		key.DoNotCache = true
+		forcedDoNotCache = true
+	}
+
 	res, err = c.cache.GetOrInitializeWithCallbacks(ctx, key, fn)
 	if err != nil {
+		// mark that the next attempt should run with DoNotCache
+		c.noCacheNext.Store(key.CallKey, struct{}{})
 		return nil, err
+	}
+
+	// success: we're in a good state now, allow normal caching again
+	c.noCacheNext.Delete(key.CallKey)
+
+	// If we forced DoNotCache due to a prior failure, we need to re-insert the successful
+	// result into the underlying cache under the original key so subsequent calls find it.
+	// The call above used a random storage key (due to DoNotCache=true), so the result
+	// won't be found by future lookups using the original key.
+	if forcedDoNotCache {
+		key.DoNotCache = false
+		cachedRes, cacheErr := c.cache.GetOrInitializeValue(ctx, key, res.Result())
+		if cacheErr == nil {
+			res = cachedRes
+		}
+		// If caching fails, we still return the successful result, just won't be cached
 	}
 
 	c.mu.Lock()
@@ -177,7 +206,7 @@ func (c *SessionCache) GetOrInitializeWithCallbacks(
 		return nil, err
 	}
 
-	if !key.DoNotCache {
+	if !key.DoNotCache || forcedDoNotCache {
 		c.results = append(c.results, res)
 	}
 
--- a/dagql/session_cache_test.go
+++ b/dagql/session_cache_test.go
@@ -2,6 +2,7 @@ package dagql
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


vito:
Love this fix - will be important for LLM sessions, since right now if you hit an error at the API level (like `dag.LLM()` refreshing its auth and failing) your session is essentially doomed.

sipsma (commented):
I'd be okay with the approach this fix takes, but it does have a bunch of race conditions in weird cases, e.g. two calls to the same function close to each where one gets an error and one succeeds. You could end up in a situation where even though the function call succeeded once, a subsequent attempt still re-runs w/ DoNotCache, depending on the exact timing of everything.

Those are gonna be pretty rare but it feels like shaky ground to be sitting on.

---

The crux of the issue is actually just that buildkit caches errors for the duration of a session. I think it's worth a try to just disable that entirely.

The idea would be to get rid of the `s.execDone` and `s.execErr` fields as used [here](https://github.com/sipsma/dagger/blob/875f8465253296ffb120f0dd6346415130e79d2d/internal/buildkit/solver/jobs.go#L959-L1023). The logic should essentially just be that `s.execRes` is set when there's no error and it should only skip re-running if `s.execRes` is not nil.

If that ends up blowing up and not being as simple as it sounds, I'm good with taking the fix here for now.

sipsma (approved):
LGTM for now with follow-ups to patch up the race conditions (https://discord.com/channels/707636530424053791/1441464505665323099/1445956297215315968)

PR #11489, opened 2025-11-25
configuration to ignore checks in a toolchain

kpenfound:
Follow up to https://github.com/dagger/dagger/pull/11480

Ignore checks within a toolchain

Note the check names are **not** prefixed with the toolchain name because that would be redundant.


kpenfound:
The engine test `TestEngine/TestLocalCacheGC` is flaky and I will rerun when checks are complete

TomChv:
#11501 Should fix that, I'll review the PR in the current state and you can rebase when possible :)

PR #11490, opened 2025-11-25
C# SDK - module support with interfaces, docs, attributes, roslyn analyzers

pjmagee:
Hi all,

This is a 2nd jab at doing a dotnet (C#) SDK. I didn't want to disrupt the existing experimental dotnet SDK. A few reasons:

- other sdks are language specific, so 'csharp' sdk makes sense and then we might make fsharp later
- quite different from the existing sdk in that it does have module support
- I didn't want to cause conflicts with the existing one and cause confusion so I kept it entirely isolated
- roslyn analyzers for intellisense hints/warnings on using the Dagger SDK properly that is picked up by vscode/Visual Studio/rider C# IDEs

requesting some feedback and to peek community interest for a C# SDK

I would like help, feedback etc - community driven :-)

```dagger init --sdk=github.com/pjmagee/dagger/sdk/csharp@dotnet-module-support --name=my-dagger-module --source=.dagger```

### DIFF
--- a/sdk/csharp/examples/standalone-client/Program.cs
+++ b/sdk/csharp/examples/standalone-client/Program.cs
@@ -14,7 +14,7 @@ Console.WriteLine(output);
 Console.WriteLine("\nExample 2: Running tests in a container");
 var exitCode = await Dag
     .Container()
-    .From("mcr.microsoft.com/dotnet/sdk:8.0")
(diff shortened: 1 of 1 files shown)


pjmagee:
Cleaning up and renaming branches, work continues :-)

https://github.com/pjmagee/dagger/tree/csharp/experimental

PR #11491, opened 2025-11-25
fix: load module using content digest ID

grouville:
Fixes https://discord.com/channels/707636530424053791/1440084860273426494/1440084860273426494

### Issue

When a module function returns an interface value, the returned result is cached using an identifier that includes the module's source path (e.g., `moduleSource(refString:"/abs/path/to/module")`).

On subsequent sessions, when this cached result is retrieved, the `IDDeps` method walks the ID to reload module dependencies. This action triggers calls to `moduleSource`, which leverages `CachePerClient`.

Consequently, different clients experience cache misses and are forced to re-resolve the local path.

### Root Cause

The `moduleSource` method previously used the context of the currently executing client for filesystem operations. Within a module function context (executing in an isolated container), the filesystem access (e.g., stat) would fail since the container lacks visibility into the host's filesystem where the module source actually resides.

### Solution

The solution is to retrieve the nearest non-module ancestor client's context via `NonModuleParentClientMetadata` before performing filesystem operations.

This approach ensures that filesystem operations correctly reference the host's context, consistent with established patterns already employed elsewhere in `modulesource.go`.

cc @chrisjpalmer 

## Update 11/25/2025

It turns out that this wasn't the proper solution as this introduced a breach to the isolation of modules. The correct one was to:  load the module using the content digest ID (constant across sessions) rather than the per-client ID digest.

### DIFF
--- a/core/schema/modulesource.go
+++ b/core/schema/modulesource.go
@@ -349,18 +349,6 @@ func (s *moduleSourceSchema) localModuleSource(
 		localPath = "."
 	}
 
-	// For local module sources, ensure we have the right client context for filesystem access.
-	// Modules run as their own clients, but need access to the original caller's filesystem
-	// (e.g., the CLI that started the session). Without this, when a module function returns
-	// an interface and that result is loaded from cache in a subsequent session, the module
-	// dependency loading will fail because it tries to stat the path from the module's
-	// container context instead of the host.
-	localSourceClientMetadata, err := query.Self().NonModuleParentClientMetadata(ctx)
-	if err != nil {
-		return inst, fmt.Errorf("failed to get local source client metadata: %w", err)
-	}
-	ctx = engine.ContextWithClientMetadata(ctx, localSourceClientMetadata)
-
 	// figure out the absolute path to the local module source
 	var localAbsPath string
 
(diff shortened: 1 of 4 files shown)


core/schema/modulesource.go
```diff
@@ -349,6 +349,18 @@ func (s *moduleSourceSchema) localModuleSource(
 		localPath = "."
 	}
 
+	// For local module sources, ensure we have the right client context for filesystem access.
+	// Modules run as their own clients, but need access to the original caller's filesystem
+	// (e.g., the CLI that started the session). Without this, when a module function returns
+	// an interface and that result is loaded from cache in a subsequent session, the module
+	// dependency loading will fail because it tries to stat the path from the module's
+	// container context instead of the host.
+	localSourceClientMetadata, err := query.Self().NonModuleParentClientMetadata(ctx)
```
sipsma: I haven't looked deeply enough yet to think through a fix, but just flagging right away that we can't make this change because it would allow function code in a module to load arbitrary host files/dirs by just making a `dag.ModuleSource("/arbitrary/path/on/callers/host")` call. This is due to the fact that any such queries will now get re-routed to point to the caller's host.

sipsma:
@grouville to get a better handle on the problem I had to try some stuff locally and in the process got the test you added passing w/ a different approach that doesn't have the issue mentioned here: https://github.com/dagger/dagger/pull/11491#discussion_r2562109172

So I just went ahead and pushed that to this PR, sorry to overwrite some of this, appreciate the effort and the test repro!

---

To explain the fix I pushed a bit, the problem is very similar to what I fixed in https://github.com/dagger/dagger/pull/11350. In that case, there were context directories that referred to modules loaded in earlier sessions. In this case, same problem but in interface value IDs. The fix is essentially the same though; load the module using the content digest ID (constant across sessions) rather than the per-client ID digest.

core/schema/modulesource.go
grouville: Aaah thanks. I suspected this was intentional but wasn't sure 🙏

sipsma (approved):

PR #11497, opened 2025-11-27
engine: optimize copies to use hardlinks when possible

sipsma:
Realized while talking to a user about disk usage problems that thanks to various forward progress on the Theseus front this has become some very low-hanging fruit for optimization.

Before this a `WithDirectory` operation could *under very specific circumstances* use `MergeOp`, which attempts to use hardlinks in its implementation in order to skip copies of files and thus save disk space and time.

This change greatly expands the circumstances in which that hardlink optimization is possible. It will essentially always be used now except for the cases where an owner/permission or other metadata is being modified during the copy.
* Theoretically possible to support hardlinking even in those cases via overlay metacopy, but that's a big separate can of worms for the future.

---

Results so far:
* Fresh (no existing cache) engine build (`engine-dev container sync`) consumes 1.7GB less space
* `TestModules` takes 2.1GB less space
* Some of the test suites seem to be quite a bit faster (tentatively, hopefully not just CI noise), by up to a few minutes. They all take no longer than 9 minutes to run now (vs. like 12m on main)
   * EDIT: I might have just been getting lucky last week with CI rolls, some of those jobs are taking over 11m again. The disk space benefits still stand though.

### DIFF
--- a/.changes/unreleased/Added-20251202-125310.yaml
+++ b/.changes/unreleased/Added-20251202-125310.yaml
@@ -1,0 +1,9 @@
+kind: Added
+body: |-
+  Optimize WithDirectory+WithFile to avoid copying file data on disk in more cases.
+
+  These APIs now avoid copying data on disk in all cases except when an owner, permission or timestamp is modified as part of the operation.
+time: 2025-12-02T12:53:10.6952586-08:00
+custom:
+  Author: sipsma
+  PR: "11497"


engine/buildkit/executor_spec.go
```diff
@@ -414,16 +420,6 @@ func (w *Worker) setupRootfs(ctx context.Context, state *execState) error {
 	if releaseRootMount != nil {
 		state.cleanups.Add("release rootfs mount", releaseRootMount)
 	}
-	// TODO: is this robust? the one for submounts is very complicated
-	if state.rootMount.Selector != "" {
-		for i, mnt := range rootMnts {
```
tiborvass, line 419: this for loop disappeared in the new version. Not exactly sure what it was trying to do and if you removed it deliberately.

sipsma: Yeah the loop is gone because it was unnecessary. We can just mount the rootfs and then tell runc to use a subdir of that mount rather than reset the source of each mount (which is now handled above on line 405)

Issue #11469, opened 2025-11-23
Exposing VAULT_TOKEN directly through dagger after any autentication


nilune:
### What are you trying to do?

I’m trying to make Dagger expose the actual VAULT_TOKEN value after it authenticates to Vault via AppRole. Many tools used inside CI steps—like Terraform, consul-template, and various plugins—do not need the secrets themselves but instead require the environment variable VAULT_TOKEN so they can fetch secrets autonomously.
Ideally, after Dagger performs the AppRole login (or any other), I want to retrieve that token programmatically through the Dagger API (e.g., as a field in a structure or a helper method) so it can be passed into the tools that need it.

### Why is this important to you?

A large part of my CI tooling expects Vault authentication to be provided via the VAULT_TOKEN environment variable. Without access to the actual token, I can’t rely on these tools’ built-in secret resolution logic. For example, Terraform and consul-template can securely fetch their own secrets if VAULT_TOKEN is present—avoiding any need to manually inject secrets into the pipeline.
Since Dagger already supports AppRole authentication up front, being able to retrieve the resulting token would let me fully integrate Vault-native workflows instead of writing additional wrappers or re-implementing Vault login inside each tool.

### How are you currently working around this?

Right now, I have to duplicate the Vault authentication step inside the tools or containers themselves rather than relying on Dagger’s authentication. This means manually running vault write auth/…/login in the CI steps or bundling custom scripts to fetch the token and export it as an environment variable.
It works, but it’s not clean: the logic is duplicated, harder to maintain, and defeats the purpose of having Dagger already handle Vault authentication. A built-in way to extract the resulting VAULT_TOKEN from Dagger would remove all this boilerplate.

marcosnils:
Making such change is on trivial given that it implies re-thinking our secret provider architecture. One thing I'd suggest as a  workaround is instead of trying to exact the AppRole authenticated token from Dagger, is to perform the AppRole auth outside Dagger and then set the `VAULT_TOKEN` to that. This way, you can pass that token to any other systems you like without depending in Dagger to expose that information. 

I'm inclined to close this issues given that it seems to me that's outside Dagger's scope to perfrom these kind of tasks. 

 cc @kpenfound

kpenfound:
[...]

PR #11500, opened 2025-11-28
nitpick: git for on-prem Azure DevOps

nilune:
Add module cloning for on-prem Azure DevOps for versions:
1. [Team Foundation Server 2018 U2](https://learn.microsoft.com/en-us/rest/api/azure/devops/git/repositories/get?view=vsts-rest-tfs-4.1&tabs=HTTP) - **with** the postfix `tfs`
2. [Azure DevOps Server 2019](https://learn.microsoft.com/en-us/rest/api/azure/devops/git/repositories/get-repository?view=azure-devops-server-rest-5.0&tabs=HTTP) - **with** the postfix `tfs`
3. [Azure DevOps Server 2020](https://learn.microsoft.com/en-us/rest/api/azure/devops/git/repositories/get-repository?view=azure-devops-server-rest-6.0&tabs=HTTP) - **with** the postfix `tfs`
4. [Azure DevOps Server 2022](https://learn.microsoft.com/en-us/rest/api/azure/devops/git/repositories/get-repository?view=azure-devops-server-rest-7.0&tabs=HTTP) - **without** the postfix `tfs`
5. [Azure DevOps Server 2022.1](https://learn.microsoft.com/en-us/rest/api/azure/devops/git/repositories/get-repository?view=azure-devops-server-rest-7.1&tabs=HTTP) - **without** the postfix `tfs`

Resolves problem in https://discord.com/channels/707636530424053791/1442826165231550585 and affects #11469

### DIFF
--- a/engine/vcs/vcs.go
+++ b/engine/vcs/vcs.go
@@ -343,7 +343,6 @@ func RepoRootForImportPath(importPath string, verbose bool) (*RepoRoot, error) {
 	rr, err := RepoRootForImportPathStatic(importPath, "")
 	if err == errUnknownSite {
 		rr, err = RepoRootForImportDynamic(importPath, verbose)
-
 		// RepoRootForImportDynamic returns error detail
 		// that is irrelevant if the user didn't intend to use a
 		// dynamic import in the first place.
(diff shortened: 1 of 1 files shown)


nilune:
@marcosnils I need your help :) If I understood correctly, there are large-scale changes happening to the repository settings and the overall project build process. It looks like the `playground terminal` no longer works for me (I think it was removed entirely), and the contributing docs also seem to be outdated.

As a result, I can’t build or test the new regex version at all. Moreover, writing proper tests is almost impossible, since this kind of Azure DevOps setup is on-prem and won’t be accessible to everyone.

So here’s the bottom line: I can’t figure out how to manually test pulling a repository from the current master branch for my on-prem Azure DevOps, and I’m not sure how to write the tests for it correctly.

nilune:
I was able to run native Golang tests using the command: `go test -v ./engine/vcs`. The old tests passed. I also added one new test for the new name.

nilune:
@marcosnils 
It works!!! I completely rebuilt the engine; added our private CA certificates and mirrors; and everything worked, now I can use git modules)))

The old tests work too; I double checked them manually.

Please merge!!!)))

marcosnils:
thx for the contribution @nilune 

Is there a chance you could do a quick check if the git@$host.com (ssh auth) format also works for your use-case?
I'd like to see if we can also make sure that SSH cloning also works so we don't find any surprise in the future 🙏

PR #11502, opened 2025-11-30
feat(sdk): Add C# SDK with module support

pjmagee:
experimental C# SDK with module support

- Attributes for Function with cache, Check, DefaultValue, Ignore, Field, Object, Name for alias support
- Defaults & optional values supported by C# syntax (string? value) or (string value = "hello")
- Constructor support with custom factory [Constructor] like python SDK `create()` for async
- Makes use of Roslyn analyzers for ensuring users of SDK use the API correctly, with IDE/Compiler hints,warnings, e.g. C# class not matching Dagger.json module name, or incorrectly using Attributes
- Makes use of .editorconfig to enable or override compiler/IDE C# analyzers
- .NET 10 long term support - officially released on November 11, 2025, and is supported for three years until November 14, 2028

- preview nuget package published at https://www.nuget.org/packages/DaggerIO/ will require Dagger team to create nuget.org account with API Key to publish update/packages but I followed existing sdks on version/bumping - happy for changes/reviews of any gaps (nuget package is NOT required for dagger modules, as the sdk is vendored with the generated code)


sdk/csharp/src/Dagger.SDK.CodeGen/Code/CodeRenderer.cs
```diff
@@ -0,0 +1,739 @@
...
+            var methodName = Formatter.FormatMethod(field.Name);
+
+            if (type.Name.Equals(field.Name, StringComparison.CurrentCultureIgnoreCase))
+            {
+                methodName = $"{methodName}_";
+            }
+
+            if (isAsync)
+            {
+                methodName = $"{methodName}Async";
```
TomChv: So that means the API is different on `csharp`?

Is there a way to keep the API names consistent with the `GraphQL` names?

For example, in the hello world, we should have:

```cs
        return await Dag.Container()
            .From("alpine:latest")
            .WithMountedDirectory("/mnt", directoryArg)
            .WithWorkdir("/mnt")
            .WithExec(new[] { "grep", "-R", pattern, "." })
            .Stdout(); // Instead of `StdoutAsync()`
```

TomChv: I can see in the generated client than the function is already declared `async` so why do we have the suffix `Async`?

```cs
public async Task<string> StdoutAsync(CancellationToken cancellationToken = default)
    {
        var queryBuilder = QueryBuilder.Select("stdout");
        return await QueryExecutor.ExecuteAsync<string>(GraphQLClient, queryBuilder, cancellationToken);
    }
```

sdk/csharp/src/Dagger.SDK.CodeFixes/AddDefaultPathAttributeCodeFixProvider.cs
```diff
@@ -0,0 +1,91 @@
...
+                title: "Add [DefaultPath(\".\")] attribute",
+                createChangedDocument: c =>
+                    AddDefaultPathAttributeAsync(context.Document, parameter, c),
+                equivalenceKey: nameof(AddDefaultPathAttributeCodeFixProvider)
+            ),
+            diagnostic
+        );
+    }
+
+    private async Task<Document> AddDefaultPathAttributeAsync(
```
TomChv, line 57: I'm curious about that one: I'm not sure I understand the purpose of this CodeFix, why adding a default path on directory argument?

sdk/csharp/src/Dagger.SDK.CodeFixes/AddFunctionAttributeCodeFixProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+                title: "Add [Function] attribute",
+                createChangedDocument: c =>
+                    AddFunctionAttributeAsync(context.Document, methodDeclaration, c),
+                equivalenceKey: nameof(AddFunctionAttributeCodeFixProvider)
+            ),
+            diagnostic
+        );
+    }
+
+    private async Task<Document> AddFunctionAttributeAsync(
```
TomChv, line 53: Do this mean any public function will be exposed by the dagger API?
What if you just want a public function so you can call it from an external package but you don't want to expose it?

sdk/csharp/src/Dagger.SDK/Attributes/FieldAttribute.cs
```diff
@@ -0,0 +1,27 @@
...
+
+/// <summary>
+/// Marks a property or field as exposed in a Dagger object.
+/// </summary>
+[AttributeUsage(
+    AttributeTargets.Property | AttributeTargets.Field,
+    AllowMultiple = false,
+    Inherited = false
+)]
+public class FieldAttribute : Attribute
```
TomChv: `Field` is not really a thing on dagger because it doesn't really exist in `GraphQL`, everything is a function.

As decided and applied in https://github.com/dagger/dagger/pull/7701, `field` should be  decorated as `func` since these are GQL accesor.

So is it possible to use the `Function` attribute for fields?

TomChv (commented):
:warning: I'm far from being a C# expert so some of my questions may look out of context due to my lack of knowledge.

That review is a first look, but there will be more coming next week.

The code looks very clean overall, thank you for making it easy to review!
Please rebase whenever you can, the repo is moving fast :)

Here's a summary of some tests I did:
- [x] Initialize a csharp project
- [x] Initialize a csharp project in a sub directory
- [x] Create and call check
- [x] Create and call function

About the performances, it's decently fast and usable, I'll not worry to much about optimization for now, we can follow up later. 

The goal is to merge that one during January, so it gives us a bit of time to prepare the ground and sets the current dotnet SDK as deprecated.

sdk/csharp/src/Dagger.SDK/Attributes/FieldAttribute.cs
pjmagee: Ah, I was following ts sdk. Its possible to use the existing `Function` attribute. I will this as a todo to fix for this PR

sdk/csharp/src/Dagger.SDK.CodeFixes/AddFunctionAttributeCodeFixProvider.cs
pjmagee: Code fixes are only applied manually, and can be configured. It appears as a small lightbulb. So if you dont want to apply [Function] then you don't need to apply it. Maybe it should only be a code analyzer rule as a hint?

sdk/csharp/src/Dagger.SDK.CodeFixes/AddDefaultPathAttributeCodeFixProvider.cs
pjmagee: e.g +defaultPath only works on `dagger.Directory` so this is to show the user they can add a default Path. It could also only be a hint. 

This is an Info-level suggestion with a code fix. It's meant as a convenience for the common case where Directory parameters represent source code (e.g., build/lint/test functions). Developers can ignore it when inappropriate (deployment targets, uploads, etc.). The severity is Info (not Warning) because it's context-dependent - we trust developers to evaluate whether it makes sense for their function's semantics. 

It enables IDE to quickly add [DefaultPath(".")] without them having to type it if they can use hotkeys to apply the suggestion.

sdk/csharp/src/Dagger.SDK.CodeGen/Code/CodeRenderer.cs
pjmagee: It's more of a code convention / practice that all async functions should be suffixed with `Async`, as microsoft does this in their own public libraries. You'll notice a large portion of most C# written code would follow this.

This section is what I was following: 

https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/task-asynchronous-programming-model#naming-convention

This is defined as a _should_ so we can decide collectively to ignore this and keep it matching the API 1:1. I would prefer keeping async, but as Dagger as only few leaf nodes that would have Async, I would also be happy matching the API with GraphQL names - I was just thinking of the C# mindset foremost. 

Also I think another interesting point would be, we wouldn't really want users to create their own async dagger functions and add the `Async` suffix to their methods. So in some ways, we're half abiding by this naming convention by only enforcing/generating async methods that return `async Task<T>`. It would be _silly_ to make dagger functions suffixed since these would likely be called from the command line. 

One part of me considers removing the 'Async' suffix because both we wouldn't want users defining their functions with the suffix. 

Would be nice if we could get more thoughts from others on this specific point. I certainly want us to be 100% confident with which approach before we merge this in.

pjmagee:
@TomChv really appreciate your feedback!! Especially with not being C# developer. Hopefully we can get some feedback from others. 

Regarding the Analyzers/Codefixes - They are also entirely optional and can be completely disabled, my idea behind adding them was to add some 'rails' or quick fixes. Though they could be considered maybe opinionated, or not 'complete'.

sdk/csharp/src/Dagger.SDK.CodeGen/Code/CodeRenderer.cs
TomChv: I think it's very important to be consistent with the GraphQL naming because dagger is cross language, so if I install your module and see `Async` everywhere, it might feels a bit strange.
That's also the same when you are installing dependencies, each call will be async.

> Would be nice if we could get more thoughts from others on this specific point. I certainly want us to be 100% confident with which approach before we merge this in.

/cc @shykes @marcosnils

sdk/csharp/src/Dagger.SDK.CodeFixes/AddDefaultPathAttributeCodeFixProvider.cs
TomChv: Oh okay, that's a great suggestion then!

sdk/csharp/src/Dagger.SDK.CodeFixes/AddFunctionAttributeCodeFixProvider.cs
TomChv: I think it's okay as a lightbulb, so it's easy to expose a function :)

TomChv:
> Regarding the Analyzers/Codefixes - They are also entirely optional and can be completely disabled, my idea behind adding them was to add some 'rails' or quick fixes. Though they could be considered maybe opinionated, or not 'complete'.

I think the analyzers/codefixes are a great idea! I wish I could add that to other SDKs too!

sdk/csharp/src/Dagger.SDK/Attributes/FieldAttribute.cs
pjmagee: this part is now done

sdk/csharp/src/Dagger.SDK.CodeGen/Code/CodeRenderer.cs
pjmagee: Yes, I also did something a bit funky with interfaces. I am tempted to remove the async suffix AND remove the implicit interace trimming of `I` that usually is seen across 99.9% of C# Code. Maybe I can add something like `[Interface(Name = "CustomName")]` and then that custom name is used to register the interface. Might be a bit cleaner.... and also removing hiding implicit logic of stripping any 'I' which now looking at it, i don't like.

TomChv: > Yes, I also did something a bit funky with interfaces. I am tempted to remove the async suffix AND remove the implicit interace trimming of I that usually is seen across 99.9% of C# Code. Maybe I can add something like [Interface(Name = "CustomName")] and then that custom name is used to register the interface. Might be a bit cleaner.... and also removing hiding implicit logic of stripping any 'I' which now looking at it, i don't like.

Yeah I think it's better to have that custom decorator to set the name to expose in the API.

shykes:
@dagger/maintainers so should we merge this?

eunomie:
> @dagger/maintainers so should we merge this?

IIRC one of the question was (maybe on discord?) how this could replace the already existing dotnet sdk. Or maybe we want to keep both?
I don't know if @TomChv you discussed a more detailed plan on that aspect?

core/sdk/loader.go
```diff
@@ -138,6 +138,8 @@ func (l *Loader) namedSDK(
 		return l.SDKForModule(ctx, root, &core.SDKConfig{Source: "github.com/dagger/dagger/sdk/php" + sdkSuffix, Config: sdk.Config, Experimental: sdk.Experimental}, nil)
 	case sdkElixir:
 		return l.SDKForModule(ctx, root, &core.SDKConfig{Source: "github.com/dagger/dagger/sdk/elixir" + sdkSuffix, Config: sdk.Config, Experimental: sdk.Experimental}, nil)
+	case sdkCSharp:
+		return l.SDKForModule(ctx, root, &core.SDKConfig{Source: "github.com/pjmagee/dagger/sdk/csharp@csharp/experimental" + sdkSuffix, Config: sdk.Config, Experimental: sdk.Experimental}, nil)
```
pjmagee, line 144: This would need updating actually once merged

shykes:
If this is the only .NET SDK with module support, then that's the one we should use, unless there's a major red flag. Module support is a priority for a SDK to be viable.

TomChv:
If you can make the PR green, I would be happy to give it another review so we can merge that

pjmagee:
> If you can make the PR green, I would be happy to give it another review so we can merge that

We're getting there! Will take a look some time this week

pjmagee:
I have to say.... windows, linux, git and LF/CRLF has caused me too much trouble :-)

pjmagee:
> If you can make the PR green, I would be happy to give it another review so we can merge that

green!

image

TomChv:
Cool! Please rebase your PR, and fix the go lint issues: https://dagger.cloud/dagger/checks/github.com/dagger/dagger@7c1c76a571c8578777eab6d0319fba08f79edc0c?check=go%3Alint#2f123919694a1896

Screenshot 2026-03-09 at 11 40 11

I'll start the review today :D

core/sdk/loader.go
TomChv: You will also need to follow up with an update here: https://github.com/dagger/dagger/blob/bb16585b1b432e05e11fc00f388a21d4dbb3d1c1/core/sdk/loader.go#L234-L235

PR #11508, opened 2025-12-01
feat: trace TS SDK test using @otr/mocha-test lib

TomChv:
Install `@otel-test-runner/mocha-test` as dev dependency. 
Update tests suite to use `@otel-test-runner/mocha-test`.

~~Note: My lib doesn't currently support calling `this.timeout(xxxx)` so I needed to update the tests suites a bit~~ -> UPDATED ✅  now we have almost 0 code changes 🚀 

**Demo with a dev engine from #11486**

https://github.com/user-attachments/assets/c05c45e4-6b92-4806-8ac5-6a3d8465e524

### DIFF
--- a/sdk/typescript/package.json
+++ b/sdk/typescript/package.json
@@ -52,7 +52,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^10.0.0",
-    "@otel-test-runner/mocha-test": "^0.0.10",
+    "@otel-test-runner/mocha-test": "0.0.12",
     "@trivago/prettier-plugin-sort-imports": "^6.0.0",
     "@types/adm-zip": "^0.5.7",
     "@types/mocha": "^10.0.10",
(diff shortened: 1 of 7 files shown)


sdk/typescript/src/module/introspector/test/files.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { describe, it } from "@otel-test-runner/mocha-test"
```
vito, line 1: Just curious: why didn't these files need an import before, like `invoke.spec.ts` had?

vito (approved):
Pretty sweet! Not sure what's up with the checks but approving in principle, seems like the ideal amount of code to change for what you get out of it. The dependencies are super lightweight too. 🧑‍🍳 🤌 💋

sdk/typescript/src/module/introspector/test/files.spec.ts
TomChv: Because `mocha` functions are available in the `global` namespace so it's like calling a standard native function when you create a test with mocha. These functions are only callable when running through the `mocha` CLI and that way you don't need to import them.

But:
- Linter sometime complain that the function doesn't exist (because `mocha` is the only one who knows 🧙‍♀️)
- I don't have a way through my lib to directly override these global functions (and I don't want to because it could lead to unexpected behaviour)

PR #11515, opened 2025-12-02
commit generated files

kpenfound:
This removes the dagger module codegen from the go toolchain's lint and tidy functions by committing the generated files. This means that lint and IDEs will work with a fresh checkout.

We will be one step closer to a generic go toolchain.

The `generate` and `checkGenerated` are now responsible for making sure the generated files are up to date.

### DIFF
--- a/.dagger/internal/dagger/dagger.gen.go
+++ b/.dagger/internal/dagger/dagger.gen.go
@@ -165,9 +165,6 @@ type CheckGroupID string
 // The `CheckID` scalar type represents an identifier for an object of type Check.
 type CheckID string
 
-// The `CiID` scalar type represents an identifier for an object of type Ci.
-type CiID string
-
 // The `CliID` scalar type represents an identifier for an object of type Cli.
 type CliID string
 
@@ -185,9 +182,6 @@ type DirectoryID string
 
 // The `DocsID` scalar type represents an identifier for an object of type Docs.
 type DocsID string
-
-// The `ElixirSdkID` scalar type represents an identifier for an object of type ElixirSdk.
-type ElixirSDKID string
 
 // The `EngineDevID` scalar type represents an identifier for an object of type EngineDev.
 type EngineDevID string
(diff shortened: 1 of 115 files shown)


kpenfound:
So this PR is obviously huge. One reason for that it includes `internal/telemetry`, `internal/querybuilder` for every single go module. I'm wondering if we can reference a single version of these libraries somewhere in the repo rather than duplicating it every time. Remember, this is just to get functional linting and IDE completion on a fresh clone, so its fine that we still regenerate these files at runtime. Possibly for `internal/dagger` too, I'm not sure if that includes any module specific code or if thats in the higher level `dagger.gen.go`

kpenfound:
Stuck at the moment getting generate to be deterministic. May need additional fixes in a released client

sipsma:
> So this PR is obviously huge. One reason for that it includes `internal/telemetry`, `internal/querybuilder` for every single go module. I'm wondering if we can reference a single version of these libraries somewhere in the repo rather than duplicating it every time. Remember, this is just to get functional linting and IDE completion on a fresh clone, so its fine that we still regenerate these files at runtime. Possibly for `internal/dagger` too, I'm not sure if that includes any module specific code or if thats in the higher level `dagger.gen.go`

@vito what's the reason again that telemetry is included with the generated module code as opposed to just being imported? Now that I think of it besides being noisy here it probably slows down modules loads a meaningful amount since each package is technically different and thus doesn't hit the go build cache.

Same issue applies to querybuilder, but that I really feel like can/should just be an import. It wouldn't even introduce a network dep since everything like this can get pre-built and cached in the engine alongside the other stuff we already pre-cache there.

kpenfound:
@sipsma I think @TomChv had some context too

vito:
> @vito what's the reason again that telemetry is included with the generated module code as opposed to just being imported? Now that I think of it besides being noisy here it probably slows down modules loads a meaningful amount since each package is technically different and thus doesn't hit the go build cache.

OTel dependency hell.

It might be easier now - when we first switched to OTel there was more code living there that was a moving target (we had a brittle `SpanProcessor`/`Span` interface wrapper and essentially a vendored OTel logs SDK package). The OTel log SDK is still a moving target to some extent with the Go OTel logs SDK [still being pre-1.0](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/log?tab=versions), meaning it's prone to bumping ahead of the desired pinned version, but maybe the API isn't so delicate nowadays.

There are _some_ extra things in there that are nice to align with your current SDK version (like our UI knob attributes) - but those should be phased out if/when we ever merge #9327.

What we _kind of_ want is for the modules to depend on the `dagger.io/dagger` that matches their engine version but I'm pretty sure that's a circular dependency.

We could try getting around it by extracting a `telemetry` into a repo of our own, and have the engine, `dagger.io/dagger`, and modules all depend on it.

(Also note that the Go SDK _still_ does a tricky `go mod replace` dance to make sure its OTel dependencies match exactly what `dagger.io/dagger` has, which it has to do currently because they're direct dependencies that `go mod tidy` is keen to bump - maybe we won't need that anymore once it's pointing to a concrete external dependency?)

re: `querybuilder` - yeah that one is more trivial to extract out iirc.

Side note: it would be _really_ cool if the modules could depend on the centrally codegen'd core API `dagger.gen.go` somehow - that would save ~13k lines of code per module vs. 785 (querybuilder) and 2.8k (telemetry).

PR #11517, opened 2025-12-03
util/parallel: disable 'reveal' by default

shykes:
This should fix the issue in web trace view, where too many spans are hidden by default with "hiding noisy spans"


shykes:
Finally got around to it @vito

PR #11520, opened 2025-12-04
refactor cloud auth to handle both CLI and token flows

marcosnils:
Moving the cloud auth initialization earlier in the engine connection
phase so we can pass all the possible auth methods through the client
metadata which can later be used for both PARC and scale-out use cases

### DIFF
--- a/internal/cloud/client.go
+++ b/internal/cloud/client.go
@@ -42,18 +42,19 @@ func NewClient(
 
 	var ts oauth2.TokenSource
 	if cloudAuth != nil {
+		t := *cloudAuth.Token
 		if cloudAuth.Token.Type() == "Basic" {
-			cloudAuth.Token.AccessToken = base64.StdEncoding.EncodeToString([]byte(cloudAuth.Token.AccessToken + ":"))
-			ts = oauth2.StaticTokenSource(cloudAuth.Token)
+			t.AccessToken = base64.StdEncoding.EncodeToString([]byte(t.AccessToken + ":"))
+			ts = oauth2.StaticTokenSource(&t)
 		} else if cloudAuth.Token.Type() == "OIDC" {
 			// translate OIDC token to Bearer but using a static source since
 			// we don't have a refresh token to get new OIDC tokens
 			ts = oauth2.StaticTokenSource(&oauth2.Token{
-				AccessToken: cloudAuth.Token.AccessToken,
+				AccessToken: t.AccessToken,
 				TokenType:   "Bearer",
 			})
 		} else {
-			ats, err := auth.TokenSource(ctx, cloudAuth.Token)
+			ats, err := auth.TokenSource(ctx, &t)
 			if err != nil {
 				return nil, err
 			}


internal/cloud/client.go
```diff
@@ -24,54 +25,53 @@ import (
...
-	if basicAuthToken == "" {
-		basicAuthToken = os.Getenv("DAGGER_CLOUD_TOKEN")
-	}
-	if basicAuthToken != "" {
-		httpClient.Transport, err = auth.DaggerCloudTransport(ctx, basicAuthToken)
-		if err != nil {
-			return nil, err
+	var ts oauth2.TokenSource
+	if cloudAuth != nil {
+		if cloudAuth.Token.Type() == "Basic" {
```
grouville, line 46: This seems to overwrite `AccessToken` but leaves `TokenType` as `"Basic"`, so the next `NewClient` on the same `*auth.Cloud` will double encode, no?

As the struct is shared (`GetCloudAuth` → `ClientMetadata` → `Provision` → `NewClient`), should we encode a _**copy**_ instead?

Right now it's probably fine since recursive scale out is disabled, but if a single parent ever triggers multiple scale-outs (or when EnableCloudScaleOut gets enabled),  this would break, or am i missing something ?

> quick repro:
  
  ```go
  
type captureRoundTripper struct {
	headers chan<- string
}

func (rt captureRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
	rt.headers <- r.Header.Get("Authorization")
	return &http.Response{
		StatusCode: http.StatusOK,
		Body:       http.NoBody,
		Header:     make(http.Header),
		Request:    r,
	}, nil
}

func TestNewClientDoesNotMutateBasicToken(t *testing.T) {
	ctx := context.Background()

	authHeaders := make(chan string, 2)
	rt := captureRoundTripper{headers: authHeaders}
	t.Setenv("DAGGER_CLOUD_URL", "http://example.com")

	cloudAuth := &auth.Cloud{
		Token: &oauth2.Token{
			AccessToken: "secret",
			TokenType:   "Basic",
		},
	}

	client1, err := NewClient(ctx, cloudAuth)
	require.NoError(t, err)
	client1.h.Transport = rt

	_, err = client1.h.Get("http://example.com")
	require.NoError(t, err)

	require.Equal(t, "secret", cloudAuth.Token.AccessToken)

	client2, err := NewClient(ctx, cloudAuth)
	require.NoError(t, err)
	client2.h.Transport = rt

	_, err = client2.h.Get("http://example.com")
	require.NoError(t, err)

	first := <-authHeaders
	second := <-authHeaders

	expectedHeader := "Basic " + base64.StdEncoding.EncodeToString([]byte("secret:"))
	require.Equal(t, expectedHeader, first)
	require.Equal(t, first, second)
	require.Equal(t, "secret", cloudAuth.Token.AccessToken)
}
```

marcosnils: > As the struct is shared (`GetCloudAuth` → `ClientMetadata` → `Provision` → `NewClient`), should we encode a copy instead?

 good catch and it makes sense.  Passing a copy  of the `auth.Cloud` to `NewClient` seems the right thing to do. I'll change that.

PR #11530, opened 2025-12-06
chore: post-release during v0.19.8

grouville:
- [x] The same files from CI as https://github.com/dagger/dagger/pull/11464 and https://github.com/dagger/dagger/pull/11463 have been updated
- [x] For the security check failing, a PR is [already on the way](https://github.com/dagger/dagger/pull/11531) 

A few follow-up commits have been added:
1. fix the `default.ts` generation
2. Bump all the trailing modules / toolchains / examples
3. Fixing some of the checks that started failing after the big bump + make the lint / ci / generate check happy
3. Improve the `RELEASING.md` with newly updated commands


grouville:
cc @marcosnils, all the checks seem to fail due to the LLM checks failing

PR #11544, opened 2025-12-08
Enable SDKs for usage in corporate proxy environments

sleiner-v:
Similar to #11407, this patch adds support for using the system-managed CA certificates in more places. The dagger engine then takes care of adding user-supplied CA certs to the container's system CA stores.


sleiner-v:
Turns out I overlooked parts of the Rust SDK. Will work on that and convert the PR to a draft in the mean time...

grouville:
Hi @sleiner-v, 

Thanks for the contribution. Will dig deeper next week 🙏

PR #11545, opened 2025-12-09
fix invalid overlay mounts

### LINKED ISSUE #8265: kernel warnings about invalid overlay mounts
When running our integ tests, checking the kernel logs shows periodic warnings about invalid overlay mounts:
```
overlayfs: lowerdir is in-use as upperdir/workdir of another mount, accessing files from both mounts will result in undefined behavior.
```

The kernel logs can usually be check by running `sudo dmesg`.

I see this locally from time to time when running the full integ tests with `dagger call --source=.:default --docker-cfg=file:/home/sipsma/.docker/config.json test all --parallel=8 --timeout=40m`
* Not every time, ~ every couple of runs
* local uname -a output: `Linux dagger_dev 6.9.7-arm64 #1 SMP Debian 6.9.7-1 (2024-06-27) aarch64 GNU/Linux`

If you open a shell in our CI infra to a dagger engine pod and run `dmesg` you will typically see a lot of these messages (seemingly many more than I get locally, at least by eye).

This is _possibly_ related to various flakes (e.g. https://github.com/dagger/dagger/issues/8031), but that's purely hypothetical and based on the idea that "undefined filesystem behavior" could result in files getting mixed up between different mounts and/or having unexpected contents. 

However, the fact that these messages show up means there is without a doubt a bug somewhere in dagger/buildkit/containerd/etc. and that bug *may* be resulting in files/dirs getting corrupted in unpredictable ways. So opening this up as a separate issue.

---

This is exceptionally tricky to track down because the kernel does not give you any details on the actual invalid mount, just that *some* process made *some* mount that had this improper configuration.

`strace` is not really tenable here because even with the `--seccomp-bpf` option the overhead is way too high.

The best option I've found so far is this ebpf based tool that traces every mount syscall made on the machine: https://github.com/iovisor/bcc/blob/master/tools/mountsnoop.py

There's some limitations though:
* The mount options get cut off after 412 characters due to some ebpf stack size limitation
* You need to have the kernel headers for the machine installed where you run `mountsnoop`, which is extremely difficult in our CI since it would require getting the amazon linux kernel headers inside the alpine-based pod container
   * This is certainly doable one way or another with enough effort if it comes down to it though

Given the above, I have so far only managed to catch this one time when running on my local machine:
* I start `mountsnoop` in one terminal as root
* In another I run as root `dmesg -C; dmesg -w | grep -q overlay && kill <mountsnoop pid>`

The last chunk of mountsnoop logs I got are here: https://gist.github.com/sipsma/c0b92759b52f6fec0532ca122e7d40f8
* These aren't logs from the entire run since I didn't redirect them to a file 😞, so I just got a chunk from the terminal output... Trying to repro again where all the logs are saved.

Unfortunately, I did not find any obvious instances of a dir being used as a lowerdir while still mounted as an upperdir elsewhere. However:
* As mentioned above, these are not the entire logs from the whole test run yet
* This is error prone to check manually, may be worth a script
* There's various complicated scenarios that are very hard to track down just by manual inspection of the logs like
   * bind mounting an overlay mount (or subtree of one) elsewhere, which means it needs to be unmounted from multiple places before being "gone"
   * use of `MNT_DETACH` resulting in the unmount not synchronously happening and thus allowing the mount to continue to exist for some undefined time period after the unmount syscall occurs

---

Another option for debugging would be to just change the code to log every mount (which would avoid some of the limitations of mountsnoop), though there are many places this happens in buildkit/containerd/runc and quite a bit of layers of abstractions over the actual mount syscall, so this could easily miss places if not extremely thorough.
* I'm sure there's plenty of crazy options in terms of forking go itself to log these and then rebuilding everything with that. But very much a last resort.

---

cc @jedevc @gerhard

sipsma:
Been working on fixing https://github.com/dagger/dagger/issues/8265, which has been an issue since forever but seems to have become much easier to trigger recently (I'd highly suspect due to the boltdb fix making race conditions requiring tight time windows much easier to hit now).
* I'm *hoping* the problems w/ overlapping overlays is actually impacting us and the cause of some of the strange missing file errors we flake out on in CI sometimes. But it's also impossible to know whether it's the root cause until we've fixed this problem, so gotta fix it either way 🤷‍♂️

Through eBPF tracing I've eventually figured out that at minimum the primary cause (*hopefully* the only one, but not confirmed) is a situation like:
1. Engine creates any overlay mount (for anything) with an upperdir
2. Completely unrelated runc child process starts initialization (has no relation to the overlay from step 1). It gets to the point of being in its own mount namespace but hasn't yet pivot_root'd
   * Because mount namespaces are forks of the parent, the mount ns of the child process will contain the unrelated overlay mount in step 1 during this time!
3. Engine unmounts the overlay from step 1.
   * It's gone from the engine's mount ns, but not the mount ns of the child process of step 2!
4. Engine creates a new overlay mount where the upper dir from step 1 is now a lower dir
   * Because that mount still exists on the system in the mount ns of the process from step 2, it's invalid and, at least according to the kernel warning, can result in undefined behavior (whether it truly does is an open question)
5. The runc process finishes init, does the pivot_root and releases its hold on that overlay mount that leaked into it

AFAICT this is not a unique problem to dagger. There's a long and storied history of github issues about these invalid overlays and it seems to have become mostly ignored in the container ecosystem because everyone just forces `index=off` now-a-days so the errors "go away" and it just becomes a kernel warning message.
* Example from containerd https://github.com/containerd/containerd/pull/4311#issuecomment-641342365

The idea of the fix is to always spawn runc processes from a "clean" mount ns that was made during engine init and has no leaked unrelated mounts. Some `open_tree`/`move_mount` magic allows the mounts to selectively move into that namespace.

---

Some other commits are for random problems I found along the way, e.g.
* Some old buildkit code leaks tmpdirs for mounts in many case. Do a `docker exec dagger-engine-v0.19.8 ls /tmp` after running some tests to see 😄
* Incorrect ref handling logic in http code

---

TODOs (some will be spin-off PRs)
- [x] Get actually working fully
   * ~~While I don't hit the invalid mount in the scenario I was hitting it before, it is still happening. I suspect it's just due to a flaw in my fix, but tbd.~~
   * Figured it out, there was indeed still a tiny race possible due to the way I was unsharing. Fixed now and can't trigger the invalid overlay anymore (at least in `TestModule`)
- [x] Separate PR for the "load bpftrace script at engine start" feature
   * [bpfscript used (among some earlier iterations in prior commits on that branch) to debug this](https://github.com/sipsma/dagger/blob/bed2387e8a1e3f810ff5c38fc6ffdfd4bed8cbce/cmd/engine/main.go#L242)
   * eBPF is insanely useful debugging issues like this, but honestly could be used to enable all sorts of monitoring/metrics as an actual feature in the long run too
   * unfortunately `bpftrace` is not a wolfi package so there's quite a bit of shenanigans required to build it (especially on arm64 for obscure reasons)
   * Sent that out here: https://github.com/dagger/dagger/pull/11548

---

Fixes #8265

### DIFF
--- a/core/git_remote.go
+++ b/core/git_remote.go
@@ -280,7 +280,7 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI, 
 	}
 
 	opts = append(opts, gitutil.WithExec(func(ctx context.Context, cmd *exec.Cmd) error {
-		return runWithStandardUmaskAndNetOverride(ctx, cmd, "", resolvPath)
+		return runWithStandardUmaskAndNetOverride(ctx, cmd, "", resolvPath, query.CleanMountNS())
 	}))
 
 	return gitutil.NewGitCLI(opts...), cleanups.Run, nil
@@ -681,14 +681,14 @@ func mergeResolv(dst *os.File, src io.Reader, dns *oci.DNSConfig) error {
 	return nil
 }
 
-func runWithStandardUmaskAndNetOverride(ctx context.Context, cmd *exec.Cmd, hosts, resolv string) error {
+func runWithStandardUmaskAndNetOverride(ctx context.Context, cmd *exec.Cmd, hosts, resolv string, cleanMntNS *os.File) error {
 	errCh := make(chan error)
 
 	go func() {
 		defer close(errCh)
 		runtime.LockOSThread()
 
-		if err := unshareAndRun(ctx, cmd, hosts, resolv); err != nil {
+		if err := unshareAndRun(ctx, cmd, hosts, resolv, cleanMntNS); err != nil {
 			errCh <- err
 		}
 	}()
@@ -697,10 +697,19 @@ func runWithStandardUmaskAndNetOverride(ctx context.Context, cmd *exec.Cmd, host
 }
 
 // unshareAndRun needs to be called in a locked thread.
-func unshareAndRun(ctx context.Context, cmd *exec.Cmd, hosts, resolv string) error {
-	if err := syscall.Unshare(syscall.CLONE_FS | syscall.CLONE_NEWNS); err != nil {
-		return err
+func unshareAndRun(ctx context.Context, cmd *exec.Cmd, hosts, resolv string, cleanMntNS *os.File) error {
+	// avoid leaking mounts from the engine by using an isolated clean mount namespace (see container start code,
+	// currently in engine/buildkit/executor_spec.go, for more details)
+	if err := unix.Unshare(unix.CLONE_FS); err != nil {
+		return fmt.Errorf("unshare fs attrs: %w", err)
 	}
+	if err := unix.Setns(int(cleanMntNS.Fd()), unix.CLONE_NEWNS); err != nil {
+		return fmt.Errorf("setns clean mount namespace: %w", err)
+	}
+	if err := unix.Unshare(unix.CLONE_NEWNS); err != nil {
+		return fmt.Errorf("unshare new mount namespace: %w", err)
+	}
+
 	syscall.Umask(0022)
 	if err := overrideNetworkConfig(hosts, resolv); err != nil {
 		return fmt.Errorf("failed to override network config: %w", err)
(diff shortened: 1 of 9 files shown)


vito (approved):
Very nice - explanation makes total sense. These new Linux mount syscalls are so handy, too

Issue #11417, opened 2025-11-14
Can the Go SDK just do `replace dagger.io/dagger => ./internal/dagger`?


vito:
Currently a Go SDK module gets its own personal Dagger Go SDK client generated to `./internal/dagger` which the module code then imports.

It also has a locally defined `dag` variable that is the instantiated Dagger client.

This has a few downsides:

* Can't use a package for module and non-module code, since it's awkward for a module to depend on the external SDK
* `goimports` sees undeclared `dag` and auto-imports `dagger.io/dagger/dag` which breaks everything - very annoying
* Not always clear what the import path is

This also applies to the `telemetry` import - would be nice if `dagger.io/dagger/telemetry` worked and resolved to the module's local package.

idlsoft:
I've started [experimenting](11547) with this.
Have yet to do an end-to-end test.

idlsoft:
This works now.

PR #11547, opened 2025-12-09
feat: codegen with portable APIs

idlsoft:
Go modules are generating dagger APIs under an internal package.

This means that
- code written in a given module cannot be used directly anywhere else
- you can't use  `go` libraries with `dagger.io/dagger` imports.

This PR adds an option to import APIs as `dagger.io/dagger` instead of `module/internal/dagger` (see #11417)

There is a simple unittest, but it can also be tested locally using `dagger -i call engine-dev playground terminal`:
```bash
mkdir mod1; cd mod1
dagger init --sdk=go --portable-api
dagger call container-echo --string-arg "Hello, world\!" stdout
```

Note the import in `main.go`, the replace in `go.mod` and an extra `go.mod` under `internal/dagger`.

Changes are grouped by commit to make it easier to follow.

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -6442,6 +6442,46 @@ func (m *Test) PrintDefault(ctx context.Context) (string, error) {
 	}
 }
 
+func (ModuleSuite) TestPortableAPI(ctx context.Context, t *testctx.T) {
+	t.Run("checkPortableCode", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+		modGen := modInit(t, c, "go", `// portable code
+package main
+
+import (
+	"dagger.io/dagger"
+)
+
+type Test struct{}
+
+// Returns a container that echoes whatever string argument is provided
+func (m *Test) ContainerEcho(stringArg string) *dagger.Container {
+	return dag.Container().From("alpine:latest").WithExec([]string{"echo", stringArg})
+}
+`, "--portable-api")
+
+		t.Run("has an API go.mod in internal/dagger", func(ctx context.Context, w *testctx.W[*testing.T]) {
+			_, err := modGen.
+				WithExec([]string{"grep", `module\s*dagger\.io/dagger`, "internal/dagger/go.mod"}).
+				Sync(ctx)
+			require.NoError(t, err)
+		})
+		t.Run("has a dagger.io/dagger replace in go.mod", func(ctx context.Context, w *testctx.W[*testing.T]) {
+			_, err := modGen.
+				WithExec([]string{"grep", `dagger\.io/dagger\s*=>\s*\./internal/dagger`, "go.mod"}).
+				Sync(ctx)
+			require.NoError(t, err)
+		})
+		t.Run("can call a function", func(ctx context.Context, t *testctx.T) {
+			out, err := modGen.
+				With(daggerCall("container-echo", "--string-arg=hello", "stdout")).
+				Stdout(ctx)
+			require.NoError(t, err)
+			require.Contains(t, out, "hello")
+		})
+	})
+}
+
 func (ModuleSuite) TestModuleDeprecationIntrospection(ctx context.Context, t *testctx.T) {
 	type sdkCase struct {
 		sdk        string


TomChv:
@idlsoft Hey!

What is the difference between your proposal and the generated client feature? #9582 

```
go mod init example.com
dagger init # No SDK, this can be used outside module
dagger install xxx # Install any module you want to call directly from your Golang project
dagger client install go
```

This will generate a client that you can use to call your module from an outside.

idlsoft:
> What is the difference between your proposal and the generated client feature? #9582 

This serves a different purpose, and gives extra flexibility on how you write your module code.
In your module you get to import `dagger.io/dagger` instead of `module/internal/dagger`, which then allows you to have a go library outside your modules with the same `dagger.io/dagger` imports and use it as a regular go dependency in all your modules.

Unlike a dagger client, which only lets you call dagger functions, this will allow you to reuse any code, with any types not serializable by the dagger engine. It's not cross-language but that's not its purpose.
And it's not a replacement, it doesn't affect how modules talk to each other. 

Btw this started with me asking a question [here](https://discord.com/channels/707636530424053791/1446060054896967730).

TomChv:
> This serves a different purpose, and gives extra flexibility on how you write your module code.
In your module you get to import dagger.io/dagger instead of module/internal/dagger, which then allows you to have a go library outside your modules with the same dagger.io/dagger imports and use it as a regular go dependency in all your modules.

So do that means you can technically `go install` your module if it's only using the core bindings?

> Btw this started with me asking a question [here](https://discord.com/channels/707636530424053791/1446060054896967730).

Okay yeah I understand the use case

core/modulesource.go
```diff
@@ -185,10 +185,11 @@ type ModuleSource struct {
...
 
-	Kind   ModuleSourceKind `field:"true" name:"kind" doc:"The kind of module source (currently local, git or dir)."`
-	Local  *LocalModuleSource
-	Git    *GitModuleSource
-	DirSrc *DirModuleSource
+	Kind                ModuleSourceKind `field:"true" name:"kind" doc:"The kind of module source (currently local, git or dir)."`
+	Local               *LocalModuleSource
+	Git                 *GitModuleSource
+	DirSrc              *DirModuleSource
+	PendingExperimental []string
```
TomChv, line 190: Why do you need this `pendingExperimental`?

cmd/codegen/generator/go/generate_module.go
```diff
@@ -117,83 +118,113 @@ func (g *GoGenerator) bootstrapMod(mfs *memfs.FS, genSt *generator.GeneratedStat
...
+	type goModGen struct {
+		modname     string
+		relpath     string
+		packageInfo *PackageInfo
+	}
+	var goModGens []*goModGen
+	if g.Config.IsPortableDaggerAPI() {
+		goModGens = append(goModGens, &goModGen{
+			modname: "dagger.io/dagger",
+			relpath: "internal/dagger",
```
TomChv, line 137: If I'm reusing your example

> dag-common has a dagger.io/dagger dependency.
mod1 is a dagger module and has a dag-common dependency.
mod1 declares a function called apt-install which uses code from dag-common

So here you are tricking the compiler into thinking that `dagger.io/dagger ` in `internal/dagger` and the `dagger.io/dagger` installed as a dependency for your `dag-common` lib are the same?

idlsoft: > So here you are tricking the compiler into thinking that `dagger.io/dagger ` in `internal/dagger` and the `dagger.io/dagger` installed as a dependency for your `dag-common` lib are the same?

That's exactly right.

core/modulesource.go
idlsoft: > Why do you need this `pendingExperimental`?

I found that the invocation order is important here. [This](https://github.com/dagger/dagger/pull/11547/changes#diff-eb32a3800e114b745331565e7c23812e70007de8638c547b05be86374fb33cbeL129):
```go
	case sdkGo:
		return &goSDK{root: root, rawConfig: sdk.Config, experimental: sdk.Experimental}, nil
```
executes when the client calls
```go
modSrc = modSrc.WithSDK(sdk)
```
So the client (dagger cli in this case) needs to call `modSrc.WithExperimentalFeatures([]dagger.ModuleSourceExperimentalFeature{dagger.ModuleSourceExperimentalFeaturePortableApi})` before `modSrc.WithSDK(sdk)`, but it can't currently because of [this](https://github.com/dagger/dagger/pull/11547/changes#diff-35a72c75f37c6dcddd3e1df88cccf5a8b22ac1f6304301c66c3a1c7394ff0eafL2113):
```go
func (s *moduleSourceSchema) moduleSourceWithExperimentalFeatures(
	_ context.Context,
	parentSrc *core.ModuleSource,
	args struct {
		Features []core.ModuleSourceExperimentalFeature
	},
) (*core.ModuleSource, error) {
...
	if parentSrc.SDK == nil {
		return nil, fmt.Errorf("module source has no SDK")
```
So adding `PendingExperimental` makes either order work.

Btw there is currently a problem because in dagger cli. If you do:
```bash
dagger init
dagger develop --sdk go --with-self-calls
```
you'll get an error. `PendingExperimental` will fix that.

idlsoft:
> Code looks ok, I left 2 questions. I need to test that PR.

FWIW I wrote a basic unittest, which was surprisingly easy to do.

> I'm also curious, what would be the diff between your PR and if we supported using the library `dagger.io/dagger` in the module code instead of redeclaring all the core bindings.

That would be ideal, but I assumed there was a good reason for that redeclaration. Maybe because it's guaranteed to be tied to the engine? But yeah, that would have been neat.

TomChv:
> That would be ideal, but I assumed there was a good reason for that redeclaration. Maybe because it's guaranteed to be tied to the engine? But yeah, that would have been neat.

There are reasons but that's a topic I plan to work on in early January because I might have an idea on how to workaround it.
If that's ok for you, I'll put this PR on hold, try to find a way to use the library inside module and if I can't make it work I'll merge your PR.

idlsoft:
> There are reasons but that's a topic I plan to work on in early January because I might have an idea on how to workaround it. If that's ok for you, I'll put this PR on hold, try to find a way to use the library inside module and if I can't make it work I'll merge your PR.

Sounds good, thank you.
The 1st commit with `PendingExperimental` is independent though, and will solve the `dagger develop --sdk=go --with-self-calls` issue, so could be useful.

TomChv:
> The 1st commit with PendingExperimental is independent though, and will solve the dagger develop --sdk=go --with-self-calls issue, so could be useful.

/cc @eunomie on that one

idlsoft:
> There are reasons but that's a topic I plan to work on in early January because I might have an idea on how to workaround it. 

I see discussions about committing generated code to the repo, wonder if what you're doing may help make it cleaner.

TomChv:
> I see discussions about committing generated code to the repo, wonder if what you're doing may help make it cleaner.

Yeah it should make the generated code much lighter, ideally we should only commit the generated bindings and that's it.
I'm planning to work on that this week, I'll give update once I have done some progress;

TomChv:
I investigated different solution to make the API portable without tricking the compiler and the best solution I found is to change the generated binding so they use interface.

For example here's an example of what `Container` would you like (simplified for the example):

```go
// In `dagger/daggen.gen.go`
type Container interface {
  From(string) Container
  With(WithContainerFunc) Container
}

// In dagger/impl.gen.go
type container struct {
	query *querybuilder.Selection

	// ...
}

type WithContainerFunc func(r Container) Container

// With calls the provided function with current Container.
//
// This is useful for reusability and readability by not breaking the calling chain.
func (r *container) With(f WithContainerFunc) Container {
	return f(r)
}

// Download a container image, and apply it to the container state. All previous state will be lost.
func (r *container) From(address string) Container {
	q := r.query.Select("from")
	q = q.Arg("address", address)

	return &container{
		query: q,
	}
}
```

So essentially we would generate in `dagger/dagger.gen.go` a bunch of interface to define all callable function etc and then `dagger/impl.gen.go` that implement the interface, finally `dagger/client.gen.go` that make the union between the interface and the implementation so `c.Container()` or `dag.Container()` still work.

The changes would indeed be a breaking change since all signature would need to move from `*dagger.Container` to `dagger.Container` but that change may be doable inside a `PostCommand` codegen

https://github.com/dagger/dagger/blob/b6993a36d01bc56ccb9a277924187c118682f4f8/cmd/codegen/generator/generator.go#L46

If implemented, that proposal is quite a big change but I believe that's the cleanest way to support portable API because now any core bindings generated by a module will implement the right interface.

I think this also requires https://github.com/dagger/dagger/pull/11708 so generated core bindings can use the published interface to keep interoperability.

What do you think of this proposal?

/cc @sipsma @eunomie

idlsoft:
> I investigated different solution to make the API portable without tricking the compiler and the best solution I found is to change the generated binding so they use interface.
> What do you think of this proposal?

I thought about that initially, but there is a problem with passing `Opts ...` arguments.
Those are structs, which makes them easy to initialize partially, but difficult to abstract away.

TomChv:
> I thought about that initially, but there is a problem with passing Opts ... arguments.
Those are structs, which makes them easy to initialize partially, but difficult to abstract away.

Hmm that's right but is that something we need to abstract to get a working implementation?

idlsoft:
> Hmm that's right but is that something we need to abstract to get a working implementation?

They're defined in `module/internal/dagger` so any code using them won't be portable.
fwiw I use opts all the time.

TomChv:
> They're defined in module/internal/dagger so any code using them won't be portable.
fwiw I use opts all the time.

And if we're also making them interface too? that would require a bit of plumbing to put the argument in the query but in theory that could solve our problem?

idlsoft:
> And if we're also making them interface too? that would require a bit of plumbing to put the argument in the query but in theory that could solve our problem?

Aside from the amount of work, initializing an `Opt` partially will be a problem in `go`. You'll need to call a function to create an opt, and you'll need to pass all parameters.

TomChv:
*Reporting our live discussion so it doesn't get lost*

So we talk about a potential other solution to directly use the core bindings when generating module bindings.
If we find a way to remove the need to generate the `UnmarshalJSON` method for each object in https://github.com/dagger/dagger/blob/f6d55efdf724db7e3a85b662d5f57c0210f790b2/cmd/codegen/generator/go/templates/src/_types/object.go.tmpl#L216-L226 then we could reuse these types and have interoperability.

@idlsoft suggested that instead of generating that method only for method, we could instead change the constructor of the object so it can take as argument a `Marshaller` and use it.

```go
func (r *Client) Container(opts ...ContainerOpts) *Container {
	q := r.query.Select("container")
	for i := len(opts) - 1; i >= 0; i-- {
		// `platform` optional argument
		if !querybuilder.IsZeroValue(opts[i].Platform) {
			q = q.Arg("platform", opts[i].Platform)
		}
	}

	return &Container{
		query: q,
	    unmarshal: func (r *Container, bs []byte) error {
			var id string
			err := json.Unmarshal(bs, &id)
				if err != nil {
					return err
				}
				*r = *dag.LoadContainerFromID(ContainerID(id))
				return nil
		},
	}
}
```

This could happen here when calling `dag.XXX`, we could pass the marshaller that use `dag.LoadXXXFromID`: https://github.com/dagger/dagger/blob/f6d55efdf724db7e3a85b662d5f57c0210f790b2/cmd/codegen/generator/go/templates/src/dag/dag.gen.go.tmpl#L78-L94

:bulb: This is possible because we control the way the object are created :)

For the published library, we won't create a marshaller so it would simply return an error if it doesn't exist

```go
func (c *Container) UnmarshalJSON(data []byte) error {
    if c.unmarshal != nil {
        return c.unmarshal(c, data)
    }

	return errors.New("unmarshal isn't supported for type Container")
}	
````

This also need https://github.com/dagger/dagger/pull/11708 and the usage of the published core types when defining dependencies function.
To be backward compatible, the generated package could also expose type alias for these core bindings.

For example:

```go
type Container = dagger.Container
type ContainerOpts = dagger.ContainerOpts
// ...
```

Eventually, the constructor `&XX{query: q}` would not be callable anymore since the type would come from `dagger.io/dagger` but we could easily create a builder package to return `*XXX`

Example:

```go
// in dagger.io/dagger/object

func NewContainer(q *querbuilder.XX, unmarshaller (c *Container, bs []byte) error) *Container {
  return &Container{
    query: q
    unmarshaller: unmarshaller
  }
}
```

idlsoft:
I've looked a bit through the generated code, and I see a couple of references to global vars:
- `UnmarshalJSON(bs []byte) error` methods use `dag`
- `MarshalJSON() ([]byte, error)` methods use `marshalCtx`

Idk if we can avoid both, but at the very least we should try to avoid `dag`, because it's the one that's module-dependent.
We can store it in `ctx` using `context.WithValue`, and then extract it using `ctx.Value`.

idlsoft:
I updated the PR, just to keep it current, but I think there is better way to solve this.

The main obstacle seems to be that core dagger types implement standard `[Un]MarshalJSON`, and therefore can only access `dag` and `marshalCtx` via static variables.

That in turn is driven by the fact that everything is initiated within `dispatch` and `invoke` by a simple `json.[Un]Marshal` call.

So... What if core types defined a methods that accept `dag` and `marshalCtx` accordingly, and the initiating code would just pass the values?

Something like
```golang
type DaggerUnmarshaller {
    DaggerUnmarshal(data []byte, dag *dagger.Client) error
}
type DaggerMarshaller {
    DaggerMarshal(marshalCtx context.Context) ([]byte, error)
}

func DaggerUnmarshal(data []byte, dag *dagger.Client, v any) error {
    if dv, ok := v.(DaggerUnmarshaller); ok {
        return dv.DaggerUnmarshal(data, dag)
    } else {
        return json.Unmarshal(data)
    }
}
func DaggerMarshal(v any, marshalCtx context.Context) ([]byte, error) {
...
}
```
and then `dispatch` and `invoke` can replace their `json.[Un]Marshal` with`dagger.Dagger[Un]marshal`

Does that make sense?

idlsoft:
Looks like something like this would be possible with `encoding.json.v2`:

```golang
type CtxUnmarshaler interface {
    UnmarshalWithCtx(data []byte, ctx context.Context) error
}

func UnmarshalWithContext(ctx context.Context, data []byte, v any) error {
    return jsonv2.Unmarshal(data, v,
        jsonv2.UnmarshalFromFunc(func(dec *jsontext.Decoder, val any) error {
            if u, ok := val.(CtxUnmarshaler); ok {
                // Read raw JSON for this value (or decode into a buffer)
                b, err := dec.ReadValue()
                if err != nil { return err }
                return u.UnmarshalWithCtx(b, ctx)
            }
            return jsonv2.ErrSkip // let default handling proceed
        }),
    )
}
```

idlsoft:
After spending more time looking at the code, I'm not sure how critical it is not to have `marshalCtx` and `dag` as global variables.
Perhaps it's ok to leave them as is until `encoding.json.v2` is official.
My original concern was with `dagger.Client` being a custom type for every module, but  `UnmarshalJSON` code only ever calls `dag.LoadXXX` methods, which can easily be extracted into an interface.

PR #11548, opened 2025-12-10
engine: support running ebpf programs at engine start

sipsma:
eBPF was crucial in debugging https://github.com/dagger/dagger/pull/11545 and https://github.com/dagger/dagger/issues/11556. It also has quite a bit of potential for future debugging, monitoring, performance analysis, and possibly even user facing features in terms of tracing (?)

Running it as a sidecar on my machine was very tedious, it's much nicer to have it in the engine itself and part of its logs. For the previous debugging I was shelling out to `bpftrace`, but getting that to build for wolfi was a nightmare and extremely slow, so trying out `cilium/ebpf` as a library directly.
* Downside is that rather than the higher-level bpftrace script, we gotta write some C code (that compiles to ebpf, no CGo or anything needed)
* Potential upsides in the long term are more flexibility and fine-grain control

LLMs (claude code + codex) seem to be consistently very good at converting "here's what I want to trace" into working ebpf programs. The two programs checked in here (`ovl_inuse` for debugging overlay in-use warnings + `filetracer` for tracing all file syscalls) were entirely written by them.
* Included a `engine/ebpf/AGENTS.md` to help out more here
* The idea generally speaking is that when debugging we can ask an LLM to update one of the existing progs or add a new one. On a case-by-case basis we can decide to check one in if it seems re-usable; but if the prog is one-off that's fine to just never merge.

We obviously don't want to always run these programs, currently they can be configured by:
1. explicitly setting an env var on the engine `DAGGER_EBPF_PROG_<prog name>=y` (e.g. `DAGGER_EBPF_PROG_FILETRACER=y`)
   * names are defined and mapped to impl at top of `cmd/engine/main.go`
2. If `ExtraDebug` is on, progs can be enabled by default by uncommenting entries in the `defaultEbpfPrograms` map at the top of `cmd/engine/main.go`
   * this is a little more convenient for local debugging since you're often changing code around anyways

Also added the ability to selectively enable progs per test workflow (add name of progs to new arg added in `toolchains/test-split/engine-tests.gen.dang`), which is useful when trying to debug CI-only flakes.

---

Overall the overhead of tracing seems to be extremely low. The `filetracer` prog that logs every file create/rename/delete/etc. doesn't seem to have much noticeable impact on test runtime (though the engine logs do have over 1 million lines)


modules/alpine/main.go
```diff
@@ -120,15 +120,7 @@ func (m *Alpine) Container(ctx context.Context) (*dagger.Container, error) {
...
-			// HACK: Wolfi will occasionally publish APKINDEX files that are missing required
-			// versions of glibc and related packages. This is a problem when one of them, like
-			// ld-linux, specifies a dependency on an exact version that is not present.
-			// This resolves itself usually within an hour, but to avoid fundamental breakage of
-			// builds we are currently pinning these packages for now.
-			"busybox=1.37.0-r49",
-			"glibc=2.42-r0",
-			"ld-linux=2.42-r0",
-			"libcrypt1=2.42-r0",
+			"busybox",
```
sipsma, line 135: This is technically tangential, but I needed it at various times to install some packages successfully. The workaround was added prior to when we bumped our go-apk dependency which fixed a ton of bugs. Won't be sure until we try but I'm hoping we may now longer need this workaround 🤞

alexcb (approved):
looks pretty useful! I'm not that familar with eBPFs, so take my +1 with a grain of salt :)

PR #11550, opened 2025-12-10
chore: convert directory.Export to dagop

alexcb:


core/directory.go
```diff
@@ -1657,29 +1657,21 @@ func (dir *Directory) Export(ctx context.Context, destPath string, merge bool) (
 		return fmt.Errorf("failed to get buildkit client: %w", err)
 	}
 
-	var defPB *pb.Definition
-	if dir.Dir != "" && dir.Dir != "/" {
-		src, err := dir.State()
-		if err != nil {
-			return err
-		}
-		src = llb.Scratch().File(llb.Copy(src, dir.Dir, ".", &llb.CopyInfo{
```
sipsma, line 1666: nice we can skip this now! Probably meaningfully faster for huge exports

engine/buildkit/filesync.go
```diff
@@ -110,54 +107,42 @@ func (c *Client) LocalDirExport(
...
 	}.AppendToOutgoingContext(ctx)
 
-	_, descRef, err := expResult.Export(ctx, cacheRes, nil, clientMetadata.ClientID)
-	if err != nil {
-		return fmt.Errorf("failed to export: %w", err)
-	}
-	if descRef != nil {
-		descRef.Release()
+	lbl := "copying files"
+	progress := local.NewProgressHandler(ctx, lbl)
```
sipsma: nit: do we need this? I've been seeing buildkit progress stuff pop in pprof cpu profiles occasionally and was thinking about tearing it out sometime since we just use our own telemetry.

not a blocker if it's somehow load bearing still for now

sipsma (approved):

engine/buildkit/filesync.go
alexcb: good catch -- fsutils will avoid calling it if it's set to nil, it seems safe to remove.

PR #11557, opened 2025-12-11
Fix `failed to get content hash` flakes

sipsma:
This is the root cause of the occasional "failed to get content hash ...
not found" flakes we get in CI occasionally.

See the included integ test for the simplified scenario where this can
happen. Essentially when selecting a file from a directory that's
content hashed, we can get cache hits from dirs who are located at
differing subpaths of the snapshot root (i.e. root dirs are different,
but subdirs have equal content). In this case we were trying to read
files at the wrong subpath.

Fix is to just simplify Directory.file a bit. DagOp is not really
necessary for such a cheap operation (just selecting a subpath) and
didn't provide even any hypothetical benefit from keeping evaluation
lazy because the content hashing happened outside the dagop context.

Now we just consistently use the right parent directory throughout the
whole operation (by not going through dagop paths) and thus consistently
use the right subpaths.


alexcb:
Interesting find!

I thought we were going to eventually remove the `FileLLB()` function though. Currently `FileLLB()` calls `dir.StatLLB()`; where as the dagop version of `dir.File` calls `dir.Stat`, which then mounts the ref.

Maybe instead the dagop version of `File()` should make an API call to `dir.Stat()` via dagql (i.e. using `srv.Select(....)` instead), then there's no need to have a dagop version of `File()` at all?

sipsma:
> I thought we were going to eventually remove the FileLLB() function though. Currently FileLLB() calls dir.StatLLB(); where as the dagop version of dir.File calls dir.Stat, which then mounts the ref.

@alexcb I switched to FileLLB because when I used `File` I got extremely confusing panics in ContainerDagOp CacheMap in typescript tests that mount the same ref multiple places, I suspect related to it ending up with a mix of mounts that have LLB and Results. I looked for a bit but became it's own separate rabbit hole so want to separate that out and probably just go with FileLLB usage for the moment. Lemme know if I was misunderstanding your suggestion though.

alexcb:
>  I looked for a bit but became it's own separate rabbit hole so want to separate that out and probably just go with FileLLB usage for the moment

I'm happy to approve it as is, and then I could give my idea a try in a follow up PR -- which should be a bit easier since you have a test for this issue now :)

PR #11558, opened 2025-12-11
feat: speed up directory and file chown implementations

### LINKED ISSUE #11556: Performance regression in `WithMountedDirectory` with `owner` parameter
### Problem 

`Container.WithMountedDirectory` with the `owner` parameter seems to face a  performance regression since v0.19.6 (time of its dagop-ification).

A user reported mounting a directory with 100.000 files going from [~4 seconds to ~3+ minutes](https://discord.com/channels/707636530424053791/1448382606063173643)

### Root cause

My current debugging seems to isolate one culprit: 
- Commit `27af6d5f4` changed from LLB-based chown (single operation) to `Directory.Chown` which uses `filepath.WalkDir + os.Lchown` per file (`O(n)` syscalls):

```go
  // directory.go:1757
  filepath.WalkDir(chownPath, func(path string, d fs.DirEntry, err error) error {
      os.Lchown(path, ownership.UID, ownership.GID)  // One syscall per file!
  })
```

### How to repro

```shell
dagger init --sdk=go # in both 0.18.5 vs 0.19.8
```

Copy this repro:

```go
package main

import (
	"context"
	"fmt"
	"time"

	"dagger/chown-perf-repro/internal/dagger"
)

type ChownPerfRepro struct{}

func (m *ChownPerfRepro) Run(ctx context.Context, fileCount int) (string, error) {
	if fileCount == 0 {
		fileCount = 1000
	}

	dir := dag.Container().From("alpine").
		WithExec([]string{"sh", "-c", fmt.Sprintf(
			"mkdir -p /src && for i in $(seq 1 %d); do echo $i > /src/file$i.txt; done", fileCount)}).
		Directory("/src")

	start := time.Now()
	_, err := dag.Container().From("alpine").
		WithMountedDirectory("/mnt", dir, dagger.ContainerWithMountedDirectoryOpts{
			Owner: "1000:1000",
		}).
		WithExec([]string{"true"}).
		Sync(ctx)
	if err != nil {
		return "", err
	}
	withOwner := time.Since(start)

	start = time.Now()
	_, err = dag.Container().From("alpine").
		WithMountedDirectory("/mnt", dir).
		WithExec([]string{"true"}).
		Sync(ctx)
	if err != nil {
		return "", err
	}
	withoutOwner := time.Since(start)

	ratio := float64(withOwner) / float64(withoutOwner)
	return fmt.Sprintf("files=%d with_owner=%s without_owner=%s ratio=%.1fx",
		fileCount, withOwner, withoutOwner, ratio), nil
}
```

#### Output

_0.18.5_
```shell
$ chown-perf-repro2 cd /tmp/chown-perf-repro2 && /tmp/dagger_0_18_5 call run --file-count=100000
✔ connect 0.5s
✔ load module 0.8s

✔ chownPerfRepro: ChownPerfRepro! 0.0s
✔ .run(fileCount: 100000): String! 26.0s

files=100000 with_owner=24.69771825s without_owner=784.128631ms ratio=31.5x

A new release of dagger is available: v0.18.5 → v0.19.8
https://github.com/dagger/dagger/releases/tag/v0.19.8
➜  chown-perf-repro2 dagger call call run --file-count=100000
✔ connect 10.1s
✔ load module: . 11.8s
! unknown command "call" for "dagger call"

```

_0.19.8_
```
$  chown-perf-repro2 dagger call run --file-count=100000
✔ connect 0.5s
✔ load module: . 0.7s

✔ chownPerfRepro: ChownPerfRepro! 0.0s
✔ .run(fileCount: 100000): String! 3m38s

files=100000 with_owner=3m30.378930021s without_owner=970.462954ms ratio=216.8x

```

### Explored solution

Instead of chowning at mount definition time, `chown` at exec time using `fscopy.Copy`

grouville:
Fixes #11556

This is an attempt to speed up the _directory_ and _file_ `chown` implementation

#### Problem

The dagopified Chown walked and lchowned every entry: on overlayfs each lchown copies file data into the upper layer before changing ownership, so for, 100k files, some users went from ~25s (v0.18.5) to ~3m30s (v0.19.x).

#### Proposed solution

In the new flow, we create a child snapshot parented on the source, mount src (ro) and dst (rw), and copy the target with WithChown + hardlink optimization.

As entries are created, ownership is applied.  The data is also hardlinked instead of copied, and, as we never mutate the lower layer, **_overlay shouldn’t do a per-file copy-ups_** (current understanding, if not, then I don't understand why this leads to some perf improvements).

Upon my manual tests, this seems to bring the performance back to the pre-regression

_EDIT 12/11_: following Erik's comment, this doesn't apply the hardlink optimization. Given this + the current tests failing, it's highly possible that it doesn't work at all, leading to a fake optimization. Forgot to run the other tests 🤖

### DIFF
--- a/core/container.go
+++ b/core/container.go
@@ -1568,10 +1568,19 @@ func (container *Container) replaceMount(
 ) (*Container, error) {
 	target = absPath(container.Config.WorkingDir, target)
 
+	var prev ContainerMount
+	for _, mnt := range container.Mounts {
+		if mnt.Target == target {
+			prev = mnt
+			break
+		}
+	}
+
 	var err error
 	container.Mounts, err = container.Mounts.Replace(ContainerMount{
 		DirectorySource: &dir,
 		Target:          target,
+		Readonly:        prev.Readonly,
 	})
 	if err != nil {
 		return nil, err


core/directory.go
```diff
@@ -1747,27 +1747,81 @@ func (dir *Directory) Chown(ctx context.Context, chownPath string, owner string)
...
 			if err != nil {
 				return err
 			}
-			if err := os.Lchown(path, ownership.UID, ownership.GID); err != nil {
-				return fmt.Errorf("failed to set chown %s: %w", path, err)
+
+			opts := []fscopy.Opt{
+				fscopy.WithCopyInfo(fscopy.CopyInfo{
+					AlwaysReplaceExistingDestPaths: true,
+					EnableHardlinkOptimization:     true,
```
sipsma, line 1803: The hardlink optimization won't actually be enabled here because it can't be safely used when changing owner: https://github.com/sipsma/dagger/blob/ddbfb553a7b618bbea5fea04e88d3c3ea74335bb/internal/fsutil/copy/copy.go#L76-L76

grouville:
Is superseded by https://github.com/dagger/dagger/pull/11559

PR #11559, opened 2025-12-12
less fsyncs via overlay volatile

sipsma:
Looked into the [chown perf regression](https://github.com/dagger/dagger/issues/11556) with ebpf and found that each chown was causing overlay to call fsync on every file (within its copy-up internal functions), each of which took ~8ms on my laptop. 8ms isn't a lot to a human but when you are serially chowning 100k files it adds up quickly. Syscalls like this shouldn't be that slow generally, especially for the test case where each file is only 1 byte.

It's not entirely clear to me why overlay makes this so slow. I can repro the slowness outside of dagger with hand-crafted overlay mounts, but when I remove overlay from the equation and use the same underlying filesystem it's much much faster.
* This might be worth looking into separately. It may easily just be expected behavior I haven't encountered before, but smells slightly off...

This all reminded me of @marcosnils's effort a long time ago to upstream support for the overlay `volatile` option, which met a bunch of friction and didn't get merged. Thankfully, we no longer have those problems!

`volatile` skips the fsyncs overlay does and fixes the perf problem for me.

cc @grouville 

---

First run in CI is definitely quite fast, a couple of checks are several minutes faster than they were most recently on main:
1. [This PR
](https://dagger.cloud/dagger/ci?branch=chown-slow&minTimestamp=1765512047557&ref=b9a8e35c3d32b0583f5d69d327529535d2fadca7&repo=github.com%2Fdagger%2Fdagger)
2. [Main](https://dagger.cloud/dagger/ci?branch=main&minTimestamp=1765395967216&ref=fd2afa31d6b18742557d2248bbcb469ff7767e14&repo=github.com%2Fdagger%2Fdagger)

### DIFF
--- a/engine/buildkit/executor_spec.go
+++ b/engine/buildkit/executor_spec.go
@@ -33,6 +33,7 @@ import (
 	"github.com/dagger/dagger/internal/buildkit/solver/pb"
 	"github.com/dagger/dagger/internal/buildkit/util/bklog"
 	bknetwork "github.com/dagger/dagger/internal/buildkit/util/network"
+	"github.com/dagger/dagger/internal/buildkit/util/overlay"
 	"github.com/dagger/dagger/util/cleanups"
 	"github.com/docker/docker/pkg/idtools"
 	"github.com/google/uuid"
@@ -396,6 +397,7 @@ func (w *Worker) filterEnvs(_ context.Context, state *execState) error {
 	return nil
 }
 
+//nolint:gocyclo
 func (w *Worker) setupRootfs(ctx context.Context, state *execState) error {
 	var err error
 	state.rootfsPath, err = os.MkdirTemp("", "rootfs")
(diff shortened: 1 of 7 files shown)


engine/buildkit/executor_spec.go
```diff
@@ -427,8 +429,19 @@ func (w *Worker) setupRootfs(ctx context.Context, state *execState) error {
 	if err := mount.All(rootMnts, state.rootfsPath); err != nil {
 		return fmt.Errorf("mount rootfs: %w", err)
 	}
+
+	overlayIncompatDirs := overlay.VolatileIncompatDirs(rootMnts)
```
vito, line 433: For the curious (as I was) what an 'incompat' dir is: https://docs.kernel.org/filesystems/overlayfs.html#volatile-mount

Seems like a way to determine whether a mount should be discarded in the event of a rough shutdown. I guess it shouldn't matter much in our case, but that the right thing to do is to clean it up after use in the 'happy path'? The docs also say the "user" should notice its presence and discard the dirs, so I guess the OS doesn't do that itself?

sipsma: Yeah exactly, sorry I explained all this in more depth in the updated commit message but that doesn't autopropagate to the PR description. I'll put a nice comment on the code before merging too

PR #11560, opened 2025-12-12
core(git): re-lazify git API

grouville:
## Overview

This PR _re-lazifies_ the **Git** API while keeping auth and cache semantics explicit.

`GitRepository` and `GitRef` are now lazy objects: creating them does not require remote access.
 
Remote state is now materialized when leafs call the `__resolve` API, which is now the single boundary where protocol selection, auth handling, and ref canonicalization happen.

The goal is to make behavior easier to reason about and safer to evolve:
- handle construction is pure/lazy
- resolve is where access is validated and remote state is read
- downstream operations run on resolved identity.

###  Why?

The eager approach has potential room for improvement (performance wise):
1. Performance: Every `git()` call runs `ls-remote` even if the repo was never used _(also making **opportunistic head resolution impossible**, which **is going to be a follow-up of that PR**)_
2. Separation of concerns: auth injection lived inside `git()`, which already had a lot on its plate

### Design (current flow)

1. Resolve the receiver through `__resolve`.
2. `repo.__resolve`:
     - canonicalizes SSH URL user (`ssh://...` without user -> `git@...`);
     - injects parent-client auth when no explicit auth is set on the object;
     - reads remote snapshot (`ls-remote`) and marks repo resolved;
3. `ref.__resolve`:
     - canonicalizes symbolic ref to concrete name/SHA;
     - rewrites ref digest from canonical target + repo transport semantics.
4. `__tree` computes checkout and keys tree identity from checkout shape.

_Implementation note_: When selecting a different protocol / auth coupling, we re-enter in `git(...).__resolve` via `dagql.Server.Select` with rebuilt args, rather than hand-crafted ID replay helpers.

### Cache and security contract

- _in one-shot CLI usage_, auth is effectively checked once per command;
- _in long-lived clients_, the resolved value is reused within that client.

#### Here's the flow

- **Before:** `git()` eagerly detected protocol and injected auth
- **After:** `git()` just stores the URL

### introducing `__tree`

In this PR, we split `tree` and `__tree` on purpose.

1. `tree` is the public API entrypoint. it is `CachePerClient`, and it always resolves through `__resolve` first: that's where protocol choice, auth evaluation, and remote snapshotting happen. As this is the security/correctness boundary, and we do not want to bypass it.

2. `__tree` is the internal execution op (not user-facing). it assumes the ref is already resolved (commit SHA is known), then does checkout/fetch and computes identity from checkout semantics (`commit`, `depth`, `discardGitDir`).

The reason for introducing `__tree` is to get both properties at once: resolve-time checks stay per-client, while checkout results can still be cache-hit from semantic identity once resolved: instead of being tied to per-client call envelope details.
  
This is exactly what the newly introduced `TestGitTreeCacheAcrossProtocols` test checks: once resolved, content-equivalent trees can share cache across transports when `.git` is discarded, while `.git`-preserving trees remain transport-sensitive

#####   Digest semantics:
- `repo.__resolve`: protocol/auth inputs + `ls-remote` snapshot (+ pinned head when explicitly requested via `git(ref:, commit:)`)
- `ref.__resolve`: canonical ref target (`name` + `commit`) + `discardGitDir` + repo transport identity (`remote URL`); auth is intentionally excluded from the ref object digest
- `__tree`: checkout semantics only: `commit + depth + discardGitDir`

### DIFF
--- a/core/schema/git_resolve.go
+++ b/core/schema/git_resolve.go
@@ -85,7 +85,7 @@ func (s *gitSchema) resolveRemoteRepository(
 		remote.URL.User = url.User("git")
 	}
 
-	if err := s.scopeExplicitSSHAuthSocket(ctx, srv, remote); err != nil {
+	if err := s.scopeProvidedSSHAuthSocket(ctx, srv, remote); err != nil {
 		return zero, err
 	}
 
@@ -95,7 +95,7 @@ func (s *gitSchema) resolveRemoteRepository(
 	}
 
 	// 3) If auth wasn't provided explicitly, try caller-scoped auth.
-	if !hasExplicitRemoteAuth(remote) {
+	if !remote.SSHAuthSocket.Valid && !remote.AuthToken.Valid && !remote.AuthHeader.Valid {
 		resolved, ok, err := s.resolveWithClientAuth(ctx, srv, parent, remote)
 		if err != nil {
 			return zero, err
@@ -106,14 +106,31 @@ func (s *gitSchema) resolveRemoteRepository(
 	}
 
 	// 4) Resolve ls-remote and rewrite identity to semantic repo inputs.
-	return s.resolveResolvedRemoteRepository(ctx, srv, repo, remote)
+	resolved := repo.Clone()
+	resolved.Backend = remote
+
+	lsRemote, err := resolved.Backend.Remote(ctx)
+	if err != nil {
+		return zero, err
+	}
+	resolved.Remote = lsRemote
+	resolved.Resolved = true
+
+	result, err := dagql.NewObjectResultForCurrentID(ctx, srv, resolved)
+	if err != nil {
+		return zero, err
+	}
+
+	// Keep auth in repo identity; keep ls-remote in identity to invalidate on pushes.
+	if resolved.Remote != nil {
+		dgstInputs := remoteRepoResolveDigestInputs(remote, resolved)
+		result = result.WithObjectDigest(hashutil.HashStrings(dgstInputs...))
+	}
+
+	return result, nil
 }
 
-func hasExplicitRemoteAuth(remote *core.RemoteGitRepository) bool {
-	return remote.SSHAuthSocket.Valid || remote.AuthToken.Valid || remote.AuthHeader.Valid
-}
-
-func (s *gitSchema) scopeExplicitSSHAuthSocket(
+func (s *gitSchema) scopeProvidedSSHAuthSocket(
 	ctx context.Context,
 	srv *dagql.Server,
 	remote *core.RemoteGitRepository,
@@ -191,39 +208,7 @@ func (s *gitSchema) resolveWithClientAuth(
 	}
 }
 
-func (s *gitSchema) resolveResolvedRemoteRepository(
-	ctx context.Context,
-	srv *dagql.Server,
-	repo *core.GitRepository,
-	remote *core.RemoteGitRepository,
-) (dagql.ObjectResult[*core.GitRepository], error) {
-	var zero dagql.ObjectResult[*core.GitRepository]
-
-	resolved := repo.Clone()
-	resolved.Backend = remote
-
-	lsRemote, err := resolved.Backend.Remote(ctx)
-	if err != nil {
-		return zero, err
-	}
-	resolved.Remote = lsRemote
-	resolved.Resolved = true
-
-	result, err := dagql.NewObjectResultForCurrentID(ctx, srv, resolved)
-	if err != nil {
-		return zero, err
-	}
-
-	// Keep auth in repo identity; keep ls-remote in identity to invalidate on pushes.
-	if resolved.Remote != nil {
-		dgstInputs := remoteRepoResolveDigestInputs(remote, resolved)
-		result = result.WithObjectDigest(hashutil.HashStrings(dgstInputs...))
-	}
-
-	return result, nil
-}
-
-// Keep cache-identity inputs centralized for repo/ref resolution paths.
+// Keep pinned head in digest so pinned refs don't alias moving branch snapshots.
 func appendPinnedHeadDigestInputs(inputs []string, pinnedHead *gitutil.Ref) []string {
 	if pinnedHead == nil {
 		return inputs
@@ -253,6 +238,13 @@ func remoteRepoResolveDigestInputs(remote *core.RemoteGitRepository, resolved *c
 		remote.URL.String(),
 		strconv.FormatBool(resolved.DiscardGitDir),
 	}
+	inputs = appendRemoteAuthDigestInputs(inputs, remote)
+	inputs = append(inputs, "remote", resolved.Remote.Digest().String())
+	return appendPinnedHeadDigestInputs(inputs, resolved.PinnedHead)
+}
+
+// Auth material is part of repo identity so different auth contexts cannot collide.
+func appendRemoteAuthDigestInputs(inputs []string, remote *core.RemoteGitRepository) []string {
 	if remote.SSHAuthSocket.Valid {
 		if dgst := core.SocketIDDigest(remote.SSHAuthSocket.Value.ID()); dgst != "" {
 			inputs = append(inputs, "sshAuthSock", dgst.String())
@@ -268,8 +260,7 @@ func remoteRepoResolveDigestInputs(remote *core.RemoteGitRepository, resolved *c
 			inputs = append(inputs, "authHeader", dgst.String())
 		}
 	}
-	inputs = append(inputs, "remote", resolved.Remote.Digest().String())
-	return appendPinnedHeadDigestInputs(inputs, resolved.PinnedHead)
+	return inputs
 }
 
 // resolveAmbiguousURL chooses protocol from auth hints, then probes https/ssh.
@@ -324,6 +315,7 @@ func (s *gitSchema) resolveViaGitCall(
 		args = appendInheritedGitAuthArgs(args, remote)
 	}
 
+	// Preserve non-auth git() args when replaying through a canonical URL/auth form.
 	args = appendInheritedGitCallArgs(args, parent, remote)
 
 	var result dagql.ObjectResult[*core.GitRepository]
@@ -338,25 +330,17 @@ func (s *gitSchema) resolveViaGitCall(
 }
 
 func appendInheritedGitAuthArgs(args []dagql.NamedInput, remote *core.RemoteGitRepository) []dagql.NamedInput {
-	if remote.AuthToken.Valid {
-		args = append(args, dagql.NamedInput{
-			Name: "httpAuthToken", Value: dagql.Opt(dagql.NewID[*core.Secret](remote.AuthToken.Value.ID())),
-		})
+	if token := remote.AuthToken; token.Valid {
+		args = append(args, dagql.NamedInput{Name: "httpAuthToken", Value: dagql.Opt(dagql.NewID[*core.Secret](token.Value.ID()))})
 	}
-	if remote.AuthHeader.Valid {
-		args = append(args, dagql.NamedInput{
-			Name: "httpAuthHeader", Value: dagql.Opt(dagql.NewID[*core.Secret](remote.AuthHeader.Value.ID())),
-		})
+	if header := remote.AuthHeader; header.Valid {
+		args = append(args, dagql.NamedInput{Name: "httpAuthHeader", Value: dagql.Opt(dagql.NewID[*core.Secret](header.Value.ID()))})
 	}
 	if remote.AuthUsername != "" {
-		args = append(args, dagql.NamedInput{
-			Name: "httpAuthUsername", Value: dagql.NewString(remote.AuthUsername),
-		})
+		args = append(args, dagql.NamedInput{Name: "httpAuthUsername", Value: dagql.NewString(remote.AuthUsername)})
 	}
-	if remote.SSHAuthSocket.Valid {
-		args = append(args, dagql.NamedInput{
-			Name: "sshAuthSocket", Value: dagql.Opt(dagql.NewID[*core.Socket](remote.SSHAuthSocket.Value.ID())),
-		})
+	if sock := remote.SSHAuthSocket; sock.Valid {
+		args = append(args, dagql.NamedInput{Name: "sshAuthSocket", Value: dagql.Opt(dagql.NewID[*core.Socket](sock.Value.ID()))})
 	}
 	return args
 }
@@ -366,44 +350,36 @@ func appendInheritedGitCallArgs(
 	parent dagql.ObjectResult[*core.GitRepository],
 	remote *core.RemoteGitRepository,
 ) []dagql.NamedInput {
-	if keepGitDir := parent.ID().Arg("keepGitDir"); keepGitDir != nil {
-		if v, ok := keepGitDir.Value().ToInput().(bool); ok && v {
-			args = append(args, dagql.NamedInput{
-				Name: "keepGitDir", Value: dagql.Opt(dagql.Boolean(true)),
-			})
+	appendIfTrue := func(name string) {
+		arg := parent.ID().Arg(name)
+		if arg == nil {
+			return
+		}
+		v, ok := arg.Value().ToInput().(bool)
+		if ok && v {
+			args = append(args, dagql.NamedInput{Name: name, Value: dagql.Opt(dagql.Boolean(true))})
+		}
+	}
+	appendIfNonEmpty := func(name string) {
+		arg := parent.ID().Arg(name)
+		if arg == nil {
+			return
+		}
+		v, ok := arg.Value().ToInput().(string)
+		if ok && v != "" {
+			args = append(args, dagql.NamedInput{Name: name, Value: dagql.NewString(v)})
 		}
 	}
 
-	if sshKnownHosts := parent.ID().Arg("sshKnownHosts"); sshKnownHosts != nil {
-		if v, ok := sshKnownHosts.Value().ToInput().(string); ok && v != "" {
-			args = append(args, dagql.NamedInput{
-				Name: "sshKnownHosts", Value: dagql.NewString(v),
-			})
-		}
-	}
-
-	if commit := parent.ID().Arg("commit"); commit != nil {
-		if v, ok := commit.Value().ToInput().(string); ok && v != "" {
-			args = append(args, dagql.NamedInput{
-				Name: "commit", Value: dagql.NewString(v),
-			})
-		}
-	}
-
-	if ref := parent.ID().Arg("ref"); ref != nil {
-		if v, ok := ref.Value().ToInput().(string); ok && v != "" {
-			args = append(args, dagql.NamedInput{
-				Name: "ref", Value: dagql.NewString(v),
-			})
-		}
-	}
+	appendIfTrue("keepGitDir")
+	appendIfNonEmpty("sshKnownHosts")
+	appendIfNonEmpty("commit")
+	appendIfNonEmpty("ref")
 
 	if remote.Services != nil {
 		if svcArg := parent.ID().Arg("experimentalServiceHost"); svcArg != nil {
 			if litID, ok := svcArg.Value().(*call.LiteralID); ok {
-				args = append(args, dagql.NamedInput{
-					Name: "experimentalServiceHost", Value: dagql.Opt(dagql.NewID[*core.Service](litID.Value())),
-				})
+				args = append(args, dagql.NamedInput{Name: "experimentalServiceHost", Value: dagql.Opt(dagql.NewID[*core.Service](litID.Value()))})
 			}
 		}
 	}
@@ -519,7 +495,19 @@ func (s *gitSchema) resolveRef(
 	}
 	ref.Resolved = true
 
-	result, err := dagql.NewObjectResultForID(ref, srv, resolvedRefID(parent, resolvedRepo, ref))
+	// Keep caller intent (requested name) while pinning to the resolved commit.
+	refArgs := []*call.Argument{
+		call.NewArgument("name", call.NewLiteralString(parent.Self().Ref.Name), false),
+	}
+	if ref.Ref != nil && ref.Ref.SHA != "" {
+		refArgs = append(refArgs, call.NewArgument("commit", call.NewLiteralString(ref.Ref.SHA), false))
+	}
+
+	resolvedRefID := resolvedRepo.ID().Receiver().
+		Append(ref.Type(), "ref", call.WithArgs(refArgs...)).
+		Append(ref.Type(), "__resolve")
+
+	result, err := dagql.NewObjectResultForID(ref, srv, resolvedRefID)
 	if err != nil {
 		return zero, err
 	}
@@ -543,28 +531,4 @@ func refResolveDigestInputs(resolvedRepo *core.GitRepository, ref *gitutil.Ref) 
 		inputs = append([]string{remote.URL.String()}, inputs...)
 	}
 	return inputs
-}
-
-func resolvedRefID(
-	parent dagql.ObjectResult[*core.GitRef],
-	resolvedRepo dagql.ObjectResult[*core.GitRepository],
-	ref *core.GitRef,
-) *call.ID {
-	repoID := resolvedRepo.ID().Receiver()
-	return repoID.
-		Append(ref.Type(), "ref", call.WithArgs(resolvedRefCallArgs(parent, ref)...)).
-		Append(ref.Type(), "__resolve")
-}
-
-func resolvedRefCallArgs(
-	parent dagql.ObjectResult[*core.GitRef],
-	ref *core.GitRef,
-) []*call.Argument {
-	args := []*call.Argument{
-		call.NewArgument("name", call.NewLiteralString(parent.Self().Ref.Name), false),
-	}
-	if ref.Ref != nil && ref.Ref.SHA != "" {
-		args = append(args, call.NewArgument("commit", call.NewLiteralString(ref.Ref.SHA), false))
-	}
-	return args
 }


core/integration/git_test.go
```diff
@@ -1341,34 +1340,6 @@ func decodeAndTrimPAT(encoded string) (string, error) {
 	return strings.TrimSpace(string(decodedPAT)), nil
 }
 
-// Ensure IsRemotePublic correctly detects repo visibility (see dagger/dagger#11112)
-func (GitSuite) TestIsRemotePublic(ctx context.Context, t *testctx.T) {
```
grouville, line 1345: We removed the eager `ls-remote` (via `schema.IsRemotePublic`) that preemptively checked repo visibility.

Instead, the DAG recurses during resolution, trying protocol/auth variants lazily. This is more efficient for the common case (public repos over HTTPS) and defers work until actually needed.

_Azure edge case_: Azure Repos return 200 with HTML login page for unauthenticated requests lacking {org}@. The old code handled this; the new code relies on the auth retry loop. We can't easily integration-test this without org-scoped tokens, so removing this deprecated test. If Azure breaks, we'll catch it in production and fix then.

core/integration/git_test.go
```diff
@@ -1394,6 +1365,65 @@ func (GitSuite) TestGitLsRemoteSessionCache(ctx context.Context, t *testctx.T) {
 	require.Equal(t, commit, commit2)
 }
 
+func (GitSuite) TestGitTreeDigestTracksContent(ctx context.Context, t *testctx.T) {
```
grouville, line 1385: This test is a regression test for `contenthash` stability, which we had to tweak as part of the PR.

The lazy resolution canonicalizes `repo/ref` IDs, which exposed a bug where the contenthash singleflight key didn't include checksum options (depth, excludes). This caused incorrect cache hits when the same tree was hashed with different options.

As part of the changes, this adds a safeguard for future tweaks around the contenthash, with these assertion:
- Same commit → same digest (stability)
- .git changes → no effect (correctness)
- Content changes → different digest (correctness)
- Different Depth options → different digest (the bug I fixed in this commit `fix(contenthash): include checksum options in singleflight key`)

core/schema/git.go
```diff
@@ -74,6 +67,8 @@ func (s *gitSchema) Install(srv *dagql.Server) {
 	}.Install(srv)
 
 	dagql.Fields[*core.GitRepository]{
+		dagql.NodeFunc("url", s.url).
```
grouville: `url` became a lazy resolver because schemeless URLs ("github.com/dagger/dagger") don't know their protocol until resolution runs.

The old struct field was populated at construction (eager implementation), but that's too early for this new  lazy resolution pattern. Instead, we keep exposing the url field as a resolver for retro-compatibility of the API

#### The flow is:
1. `url()` calls `__resolve` on the repo
2. `__resolve` does protocol detection only if `remote.URL == nil` (schemeless)
3. After resolution, returns `remoteGitRepo.URL.String()` from the resolved repo

core/schema/git.go
```diff
@@ -133,10 +128,13 @@ func (s *gitSchema) Install(srv *dagql.Server) {
 			Args(
 				dagql.Arg("header").Doc(`Secret used to populate the Authorization HTTP header`),
 			),
+
+		dagql.NodeFuncWithCacheKey("__resolve", s.repoResolve, dagql.CachePerClient).
```
grouville: `CachePerClient` is critical here: it mixes client ID into the cache key so different clients can't see each other's auth-injected repos.

_Consistency question_: `tree/commit/ref` have `CachePerClient`, but `tags/branches` don't. Since `__resolve` runs first and has `CachePerClient`, the leaf ops might not need it (the resolved receiver already incorporates client ID). However, **defense-in-depth argues for consistency.**

**Current decision**: Add` CachePerClient` to `tags/branches` for consistency. Follow-up: add test coverage for auth isolation on these methods.

grouville: The question I'm asking myself atm is : I have an inconsistency between the `GitRef` methods and the `GitRepository methods`, regarding the scoping with `CachePerClient`.

I think I added the `dagql.CachePerClient` on `tree/commit/ref` before introducing the recursive nodes in the DAG. Now that it includes the client as a cachekey, as part of the path in the DAG, then it shouldn't be necessary.

##### 2 options
1) Either I remove it, as they're redundant
2) Or, add it to `tags/branches` (for consistency)

No strong opinion here. But i'm leaning towards some defensive programming, as a mistake could be made inadvertendly

core/schema/git.go
```diff
@@ -181,47 +182,15 @@ type gitArgs struct {
...
 
-//nolint:gocyclo
 func (s *gitSchema) git(ctx context.Context, parent dagql.ObjectResult[*core.Query], args gitArgs) (inst dagql.Result[*core.GitRepository], _ error) {
 	srv, err := core.CurrentDagqlServer(ctx)
 	if err != nil {
 		return inst, fmt.Errorf("failed to get current dagql server: %w", err)
 	}
 
 	remote, err := gitutil.ParseURL(args.URL)
-	if errors.Is(err, gitutil.ErrUnknownProtocol) {
```
grouville, line 198: That logic is now inside [repoResolve](https://github.com/dagger/dagger/pull/11560/changes#diff-bb8fba2a10b96c16befe2b145e4e580f38839f20e0518cfd8c678d2296c27cb1R789)

core/schema/git.go
```diff
@@ -240,239 +209,23 @@ func (s *gitSchema) git(ctx context.Context, parent dagql.ObjectResult[*core.Que
...
 	}
 
-	var (
-		sshAuthSock    dagql.ObjectResult[*core.Socket]
-		httpAuthToken  dagql.ObjectResult[*core.Secret]
-		httpAuthHeader dagql.ObjectResult[*core.Secret]
-	)
-
-	switch remote.Scheme {
-	case gitutil.SSHProtocol:
```
grouville, line 338: Some of the auth logic is now inside [injectAuth()](https://github.com/dagger/dagger/pull/11560/changes#diff-bb8fba2a10b96c16befe2b145e4e580f38839f20e0518cfd8c678d2296c27cb1R920)

core/dagop.go
```diff
@@ -131,6 +131,8 @@ func (op FSDagOp) Backend() buildkit.CustomOpBackend {
 }
 
 func (op FSDagOp) Digest() (digest.Digest, error) {
+	// TODO(check): consider whether CacheKey should be included here to avoid
```
grouville: todo: remove

core/git_remote.go
```diff
@@ -41,21 +41,38 @@ import (
 )
 
 type RemoteGitRepository struct {
+	// when URL is nil, it indicates an ambiguous git URL that needs to be resolved
 	URL *gitutil.GitURL
 
 	SSHKnownHosts string
-	SSHAuthSocket dagql.ObjectResult[*Socket]
+	SSHAuthSocket dagql.Optional[SocketID]
```
grouville, line 46: We store IDs instead of loaded objects. This enables lazy loading: secrets aren't fetched during DAG construction. The tradeoff: callers must explicitly `.Load()` when they need the value, but this is intentional as it defers work until resolution time.

core/git_remote.go
```diff
@@ -41,21 +41,38 @@ import (
...
-	AuthHeader   dagql.ObjectResult[*Secret]
+	AuthToken    dagql.Optional[SecretID]
+	AuthHeader   dagql.Optional[SecretID]
 }
 
 var _ GitRepositoryBackend = (*RemoteGitRepository)(nil)
 
+// NeedsAuthResolution returns true if the repository needs auth to be resolved
+// (protocol detection, SSH socket, or HTTP credentials).
+func (repo *RemoteGitRepository) NeedsAuthResolution() bool {
```
grouville: This predicate gates whether we attempt auth injection.

> Note the subtle design:
>  The URL == nil check is defensive: it's unreachable from the call site in `repoResolve`, which handles protocol detection separately. The real work is the SSH/HTTP checks.

An alternative design could eliminate this predicate and let `injectAuth` return early when there's nothing to inject. Kept as a minor optimization to skip setup when we know upfront there's nothing to do.

Also, the name is slightly misleading, it's really `"MightInjectAuth"`, as it returns true when auth could be injected, not when it's required.

core/git.go
```diff
@@ -20,11 +20,13 @@ import (
 )
 
 type GitRepository struct {
-	URL     dagql.Nullable[dagql.String] `field:"true" doc:"The URL of the git repository."`
 	Backend GitRepositoryBackend
 	Remote  *gitutil.Remote
 
 	DiscardGitDir bool
+
+	// Internal per-repo pinned HEAD (ref/commit)
```
grouville: Since `git()` is now lazy, we no longer fetch the remote at construction time. Without that fetch, there's no `repo.Remote` to derive `HEAD` from.

We store the caller's requested ref/commit as `PinnedHead` and include it in the digest so that `git("foo", ref: "main")` and `git("foo", ref: "dev")` don't collide in cache, even though neither has resolved yet (and we also sometimes need to keep the user's intent in the DAG)

core/schema/git_resolve.go
```diff
@@ -0,0 +1,152 @@
+package schema
+
+// Git Lazy Resolution
```
grouville: This file contains the lazy resolution helpers. The mental model is HTTP redirects: when resolution canonicalizes a receiver (e.g., `"github.com/foo"` → `"https://github.com/foo"`), we "redirect" by rebuilding the call chain on the new base.

Key helpers:
-`redirect` --> replays current op on resolved receiver (depth 1)
-`redirectThroughRef` --> replays through GitRef layer (depth 2, Git-specific)
-`alreadyResolving` --> prevents infinite recursion when redirect triggers `__resolve` again

Design note: `redirectThroughRef` hardcodes the `GitRef` type. This is intentional: it's `Git-specific` and not meant to generalize. If we need similar patterns elsewhere, i'd be happy to deisgn / tweak a proper abstraction then.

core/git_remote.go
grouville: I'm more tempted to implement the alternative design as a follow-up commit

core/git.go
```diff
@@ -103,13 +106,49 @@ func (*GitRef) TypeDescription() string {
 }
 
 func (ref *GitRef) PBDefinitions(ctx context.Context) ([]*pb.Definition, error) {
+	if ref.Backend == nil {
```
grouville:
```go
if ref.Backend == nil {
    return nil, nil
}
```

Guards against the lazy state. `PBDefinitions` is called during DAG traversal to collect `LLB` definitions.

In the lazy model, `refs` exist before resolution: `Backend` is nil until `__resolve` runs. Returning `nil` is safe: no definitions to contribute yet.

core/git.go
```diff
@@ -103,13 +106,49 @@ func (*GitRef) TypeDescription() string {
...
 	return ref.Backend.PBDefinitions(ctx)
 }
 
 func (ref *GitRef) Tree(ctx context.Context, srv *dagql.Server, discardGitDir bool, depth int) (*Directory, error) {
 	return ref.Backend.Tree(ctx, srv, ref.Repo.Self().DiscardGitDir || discardGitDir, depth)
 }
 
+// Resolve resolves the ref's SHA and name to their canonical forms, and initializes
+// the backend. This chains to GitRepository.Resolve() first.
+func (ref *GitRef) Resolve(ctx context.Context) error {
```
grouville: Design note on mutation: This method mutates `ref.Ref.SHA`, `ref.Ref.Name`, and `ref.Backend` in place. 

This is safe because:
  1. `__resolve` is the only entry point, and it's cached per-client
  2. The same object won't be resolved twice in the same session

However, this is implicit: if someone calls `Resolve()` directly outside of `__resolve`, they'd hit race conditions with cached objects.

> A cleaner design would return a new resolved ref, but that would require restructuring the dagql object model

core/schema/git_resolve.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"github.com/dagger/dagger/core"
+	"github.com/dagger/dagger/dagql"
+	"github.com/dagger/dagger/dagql/call"
+)
+
+// alreadyResolving returns true if we're inside a redirected __resolve call.
+// When redirect() loads the new ID, it triggers __resolve again. This check
+// prevents infinite recursion by detecting that re-entry.
+func alreadyResolving(id *call.ID) bool {
```
grouville: Fragility note: This uses a magic string comparison. If someone renames `__resolve`, this breaks silently.

A safer approach would be a constant (`const resolveField = "__resolve"`) or a _context marker_. Kept simple for now since `__resolve` is a dagql convention unlikely to change, but worth noting

core/schema/git_resolve.go
```diff
@@ -0,0 +1,152 @@
...
+		opToReplay.Field(),
+		call.WithArgs(opToReplay.Args()...),
+		call.WithView(opToReplay.View()),
+	)
+
+	result, err := srv.Load(ctx, redirectedID)
+	if err != nil {
+		return zero, err
+	}
+	return result.(dagql.ObjectResult[T]), nil
```
grouville: _Panic risk_: This type assertion will panic if the loaded type doesn't match. Same for `redirectScalar`'s `result.Unwrap().(T)`.

In practice, this is safe because we're replaying the exact same operation on a receiver of the same type: the result type is guaranteed. But, a defensive _**ok check**_ would be cleaner:

```go
typed, ok := result.(dagql.ObjectResult[T])
if !ok {
    return zero, fmt.Errorf("unexpected type %T", result)
}
```

core/schema/git_resolve.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// redirectScalar is like redirect but for scalar return types (e.g., String, Array).
+func redirectScalar[T any](
+	ctx context.Context,
+	resolvedReceiverID *call.ID,
+) (T, error) {
+	var zero T
+
+	srv, err := core.CurrentDagqlServer(ctx)
```
grouville: _Code duplication_: `redirect` and `redirectScalar` share ~90% of their logic. The only difference is `srv.Load` vs `srv.LoadType` and return type handling.

We could extract the _ID-building_ into a shared helper. Kept it separate as of now for clarity since there are only two variants and the duplication is small. _If we add more redirect variants, consolidate then?_

core/schema/git_resolve.go
grouville: I've updated with a const being reference everywhere

core/git.go
grouville: Added a comment for now; I'm ok to enforce it, but it requires a bit more work --> pending PRs would fail nonetheless

core/schema/git.go
```diff
@@ -103,12 +98,12 @@ func (s *gitSchema) Install(srv *dagql.Server) {
 		dagql.NodeFunc("latestVersion", s.latestVersion).
 			Doc(`Returns details for the latest semver tag.`),
 
-		dagql.Func("tags", s.tags).
+		dagql.NodeFuncWithCacheKey("tags", s.tags, dagql.CachePerClient).
```
sipsma, line 99: Is there a reason to use `CachePerClient` vs. `CachePerSession` here? Just curious, per-session would have been my prior guess.

Also curious why we're making `tags` and `branches` more scoped cache keys but not `latestVersion`

grouville:
Update: the diff on the `goldenTest` is suspicious as it should be cached? Upon manual tests: 

```shell
▶ ✔ git github.com/dagger/dagger | head | tree 0.0s
Directory@xxh3:6eed722d7a126c3f                                                                   

▶ ✔ git ssh://github.com/dagger/dagger | head | tree 2m19s
Directory@xxh3:3aa4f390aad24cfc   

▶ ✔ git git@github.com:dagger/dagger.git | head | tree 0.0s
Directory@xxh3:3aa4f390aad24cfc                        
```

~All those three should resolve to the same ID, let's dig a bit more 👀~ After talking with @sipsma, as we include the auth key, this is expected

core/schema/git.go
grouville: This has been done ✅

core/schema/git.go
```diff
@@ -481,31 +234,23 @@ func (s *gitSchema) git(ctx context.Context, parent dagql.ObjectResult[*core.Que
...
-	repo.Remote.Head = head
+	repo.PinnedHead = head
 	repo.DiscardGitDir = discardGitDir
 
 	inst, err = dagql.NewResultForCurrentID(ctx, repo)
 	if err != nil {
 		return inst, fmt.Errorf("failed to create GitRepository instance: %w", err)
 	}
 
+	// todo(question): do we really care about caching git() now ?
```
sipsma: it seems pretty lightweight now afaict, so probably shouldn't make a big difference one way or the other I'd guess

core/schema/git.go
```diff
@@ -515,49 +260,56 @@ func (s *gitSchema) git(ctx context.Context, parent dagql.ObjectResult[*core.Que
...
+		return dagql.Null[dagql.String](), err
+	}
+
+	if receiverChanged(resolved, parent) {
+		str, err := redirectScalar[dagql.String](ctx, resolved.ID().Receiver())
+		if err != nil {
+			return dagql.Null[dagql.String](), err
 		}
-		return false, err
+		return dagql.NonNull(str), nil
```
sipsma: kind of confused what exactly is going on here, comment will help

core/schema/git.go
```diff
@@ -911,14 +740,293 @@ func (s *gitSchema) commonAncestor(
...
+// repoResolve implements __resolve for GitRepository.
+// It detects the protocol (https/ssh) and injects auth if needed.
+func (s *gitSchema) repoResolve(
+	ctx context.Context,
+	parent dagql.ObjectResult[*core.GitRepository],
+	_ struct{},
+) (dagql.ObjectResult[*core.GitRepository], error) {
+	var zero dagql.ObjectResult[*core.GitRepository]
+
+	if alreadyResolving(parent.ID()) {
```
tiborvass: I would suggest renaming this to `alreadyResolved()` or even shorter `didResolve()` because `alreadyResolving` is confusing: it feels like there's a goroutine actively resolving but that's not the intention here AFAIU.

core/schema/git.go
grouville: > Also curious why we're making tags and branches more scoped cache keys but not `latestVersion`

I forgot about `latestVersion`, will change that 🙏 

> Is there a reason to use CachePerClient vs. CachePerSession here? 

The issue is that auth isn't in the call ID when tree() is invoked, it gets injected later inside `__resolve`:

```markdown
  Original ID (cache lookup):     `git("url").branch("main").tree()`
  After `__resolve` redirect:       `git("url", sshAuthSocket: X).branch("main").tree()`
```

The cache lookup uses the original ID (without auth). So with `CachePerSession`:
- Module A (with SSH auth) calls git("url").tree() → fetches private data → cached
- Module B (no auth, same session) calls git("url").tree() → cache HIT → gets Module A's data without `__resolve` ever running

**Question**: Is intra-session cache sharing between modules acceptable? Or should modules be isolated even within the same session? _[tests seem to expect the second point]_

_I don't think we consider module dependencies as "safe" in our security mode. If that's correct, `CachePerClient` seems to be needed on leaf ops (tree, commit, etc.) to prevent cross-module data access via cache._

core/integration/git_test.go
grouville: cc @alexcb, as you've touched `contenthash`

core/contenthash.go
```diff
@@ -98,6 +116,19 @@ func GetContentHashFromDef(
...
+
+// GetContentHashFromDefWithOpts computes a content hash with custom checksum options.
+// Set storeMetadata=false when options are context-specific (e.g., excludes).
+func GetContentHashFromDefWithOpts(
+	ctx context.Context,
+	bk *buildkit.Client,
+	def *pb.Definition,
+	subdir string,
+	opts bkcontenthash.ChecksumOpts,
+	storeMetadata bool,
```
sipsma: I think there's a problem right now in that we should refuse to store the metadata when either include or exclude is set.

Probably the best fix is to just remove this `storeMetadata` as an arg and instead determine entirely in this function whether it's safe to store the metadata based on `subdir` and the options provided.

core/schema/git.go
sipsma: > The issue is that auth isn't in the call ID when tree() is invoked, it gets injected later inside __resolve:

Ah ok, that makes sense at first glance. Let's go with `PerClient` for now, if I change my mind after reading more I'll let you know 🙂

core/contenthash.go
grouville: Good catch, thanks 🙏

core/git_remote.go
```diff
@@ -215,10 +216,15 @@ func (repo *RemoteGitRepository) setup(ctx context.Context) (_ *gitutil.GitCLI,
 		}
 	}()
 
-	if repo.SSHAuthSocket.Self() != nil {
+	if repo.SSHAuthSocket.Valid && repo.SSHAuthSocket.Value.ID() != nil {
 		socketStore, err := query.Sockets(ctx)
 		if err == nil {
-			sockpath, cleanup, err := socketStore.MountSocket(ctx, repo.SSHAuthSocket.Self().IDDigest)
+			srv, _ := CurrentDagqlServer(ctx)
```
alexcb: can we add err handling here?

core/integration/git_test.go
alexcb: ah interesting bug! `checksumG.Do(ctx, key`..... sneaky that the key didn't contain those options. Your fix seems good.

core/contenthash.go
grouville: Added a commit implementing your suggestion, feel free to confirm I understood properly 🙏

core/git_remote.go
grouville: fixed 🙏

core/schema/git_resolve.go
```diff
@@ -0,0 +1,177 @@
...
+//	git("github.com/foo").branch("main").__resolve
+//	         ↑ repo              ↑ ref       ↑ leaf (current op)
+//
+// If just the repo changed, redirect() would try: newRepo.__resolve
+// But we need: newRepo.branch("main").__resolve
+//
+// So redirectThroughRef rebuilds both the ref AND the leaf on the canonical repo:
+//
+//	BEFORE: git("github.com/foo").branch("main").__resolve
+//	AFTER:  git("https://github.com/foo").branch("main").__resolve
```
sipsma: Of all the comments on these utility helpers, this part of this one is the most clear by far. The others I couldn't really understand just from reading the comment

core/schema/git_resolve.go
sipsma, line 1: For all these helpers, I have a feeling it may be quite a bit simpler to use `dagql.Server.Select(...)` and instead have these helpers constructor `Selector` args for that call, rather than hand crafting IDs which is much more complicated.

Did you attempt that at any point? If not, I really think it's worth a shot (seems like something an LLM could do pretty well). It just seems like it will be much easier to follow what's going on.

There's an example of what I'm talking about in `container.from`: https://github.com/sipsma/dagger/blob/412064595f695bed3d5b7e327f8b9c3167717e7d/core/schema/container.go#L909-L909
* if we realize the image ref was not pinned, we figure out the pinned digest and then reinvoke it using `dagql.Server.Select`.

core/schema/git.go
```diff
@@ -910,14 +744,293 @@ func (s *gitSchema) commonAncestor(
...
+	case gitutil.SSHProtocol:
+		if repo.SSHAuthSocket.Valid {
+			return nil, false, nil
+		}
+
+		clientMD, err := engine.ClientMetadataFromContext(ctx)
+		if err != nil {
+			return nil, false, fmt.Errorf("client metadata: %w", err)
+		}
+		if clientMD.SSHAuthSocketPath == "" {
```
sipsma: I'm not convinced this will work. If you e.g. called `git(...)` on the host and passed the `GitRepository` value as an arg to a module function, then that function did something that triggered this codepath, the function would be looking for the SSH sock at that time and thus not be querying the caller.

We *don't* want to use the caller client metadata either since that would make it trivial for any function to grab a caller's SSH sock.

I suspect we want to update `git` to handle grabbing the caller's SSH auth sock just in case it's needed later, and use the PostCall client resource stuff to ensure access to it is granted if/when needed.

Lemme know if I'm missing something though. If we don't have test coverage of this already, seems worth adding.

core/schema/git.go
```diff
@@ -906,18 +788,539 @@ func (s *gitSchema) commonAncestor(
...
+	if id := parent.ID(); id != nil && id.Field() == "__resolve" && parent.Self().Resolved {
+		return parent, nil
+	}
+
+	srv, err := core.CurrentDagqlServer(ctx)
+	if err != nil {
+		return zero, fmt.Errorf("failed to get current dagql server: %w", err)
+	}
+
+	repo := parent.Self()
```
sipsma: Probably want a clone here, anything wrapped in `dagql.Result` or `dagql.ObjectResult` should be treated as immutable.

core/schema/git.go
```diff
@@ -906,18 +788,539 @@ func (s *gitSchema) commonAncestor(
...
+	fmt.Fprintf(os.Stderr, "🔑 git.repoResolve canonical=%q redirect=%v authSSH=%v authToken=%v\n", canonicalURL, needsRedirect, authInjection != nil && authInjection.sshAuthSocket.ID() != nil, authInjection != nil && authInjection.authToken.ID() != nil)
+
+	if needsRedirect {
+		// Redirect to canonical git() and mark as resolved
+		canonicalRepo, err := s.redirectToCanonicalGit(ctx, srv, parent, remote, repo, authInjection)
+		if err != nil {
+			return zero, err
+		}
+		// Mark as resolved and fetch remote metadata
+		canonicalRepo.Self().Resolved = true
```
sipsma: same comment as above about needing a clone

core/schema/git.go
```diff
@@ -906,18 +788,539 @@ func (s *gitSchema) commonAncestor(
...
+		}
+		return canonicalRepo, nil
+	}
+
+	// Already canonical - mark as resolved and fetch remote metadata
+	repo.Resolved = true
+	if _, err := repo.Backend.Remote(ctx); err != nil {
+		return zero, err
+	}
+	return parent, nil
```
sipsma: This should be using `dagql.NewObjectResultForCurrentID`

grouville:
It feels like it's now in better shape. Still another pass could be done, afterwards, to cleanup / remove some helpers, but the caching behavior, at the moment, is correct and I think has the right security boundaries.

Still a small failure on `TestGit/TestGitTreeDigestTracksContent`, but it's a test I added along this PR in which I expect children of matching content to benefit from cache hits. It's not the case on main atm

grouville:
**02/10/2026**
Full refacto done. I'm quite happy with the current implementation: both the code, the tests and the cache story is easier, and feels more natural

Last flake was a session flake (unrelated to the PR IMO). Last commit also solved the rebase error I made, when rebasing on the better SSH's cache (Erik's PR). I'll have to rebase once again on top of https://github.com/dagger/dagger/pull/11838

Next: I am now focusing on the opportunistic git resolution, which was the aim of that PR --> unlocking a more performant git API utilization

grouville:
02/11/2026

Rebased on top of the latest [cache ID changes](https://github.com/dagger/dagger/pull/11838)

Issue #11552, opened 2025-12-11
🐞 Race condition / caching bug affecting file content (0.19.7, 0.19.8)


chrisjpalmer:
### What is the issue?

The issue is that multiple files content are being swapped when files are passed to a module function in parallel. Although the trace indicates that the files are correct, the contents of the files is incorrect.

### Dagger version

dagger v0.19.7 (image://registry.dagger.io/engine:v0.19.7) darwin/arm64/v8

### Steps to reproduce

This issue can be [reproduced my demo repo](https://github.com/chrisjpalmer/dagger-file-caching-bug/tree/main).

The `my-module-tests` module has 6 files (`a`-`f`) each containing one character which is the same as the file name.

`my-module-tests` calls `my-module`'s `ExpectContents` function, passing in each file and expecting the contents to match the file's letter. Sometimes this function passes, other times it does not.

The issue appears to occur on engine versions:

- 0.19.7
- 0.19.8

It doesn't occur on these engine versions:

- 0.19.3
- 0.19.4
- 0.19.5
- 0.19.6

Changing the code from this...

```
// fixtures is a directory fetched from the host.
return dag.MyModule().ExpectFileContents(gctx, variant, fixtures.File(variant))
```
to this,

```
return dag.MyModule().ExpectFileContents(gctx, variant, dag.File(variant, variant))
```

seems to make the issue go away.

### Log output

```
❯ dagger call -m my-module-tests test-my-module

✔ connect 0.3s
✔ load module: my-module-tests 0.3s

✔ myModuleTests: MyModuleTests! 0.0s
✘ .testMyModule(
  ┆ fixtures: context /Users/christopher.palmer/workspace/dagger-caching-bug/my-module-tests/my-module-tests/fixtures (exclude: [])
  ): Void 1.3s ERROR
✘ MyModule.expectFileContents(
  ┆ expect: "d"
  ┆ file: Directory.file(path: "d"): File!
  ): Void 1.1s ERROR
! expected value d doesn't match actual contents c
```

sipsma:
Thanks for the report, can repro, looking into it!

alexcb:
fixed in #11575

PR #11575, opened 2025-12-15
bug: fixes incorrect file caching

alexcb:
This fixes a file caching bug, reported in #11552, which resulted from an incorrect Digest() override, which ignored the file Source (contents).

Removing the override will cause the digest to use the entire args struct, which includes the Source field, which contains a digest of the file being copied in, e.g. `Source:File@xxh3:97e0aa643fe2496b`

### DIFF
--- a/.changes/unreleased/Fixed-20251215-145337.yaml
+++ b/.changes/unreleased/Fixed-20251215-145337.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Fix incorrect file contents caching race condition bug
+time: 2025-12-15T14:53:37.378560538-08:00
+custom:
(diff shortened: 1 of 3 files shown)


alexcb:
I'm not sure why this `Digest()` override was here in the first place -- we have a test to check that caching works under: https://github.com/alexcb/dagger/blob/7418495c3e5e023f331f69eef5720cdd473b3b0b/core/integration/container_test.go#L5267

and in particular, the `use file via directory` subtest selects the file via a `c.Host().Directory().File()` call -- which was required to reproduce the incorrect cached contents bug report.

sipsma (approved):

PR #11232, opened 2025-10-14
ci: vastly simplified version module

jedevc:
Dev + commit versions now are all v0.0.0 based. Less git manipulation = faster version computation.

#### TODO:
- [x] Ensure v0.0.0 is not rejected by engines, and detected as a dev version.
- [x] Make sure we have a better way of detecting dirty commits (or decide to not distinguish)
  - [x] https://github.com/dagger/dagger/pull/11241
  - [x] https://github.com/dagger/dagger/pull/11237
  - [x] https://github.com/dagger/dagger/pull/11326
- [x] Ensure dagger cloud avoids indexing modules with dev engine versions

#### UPDATE - 12/10/2025

The v0.0.0 base version broke the API View system. When a dev client connects with `v0.0.0`, checks like `AfterVersion("v0.15.0").Contains("v0.0.0")` return false because `v0.0.0 < v0.15.0`, effectively showing all the oldest versions of each API. This caused SDK codegen to produce broken code with duplicate declarations.

The fix implemented in the follow-up commit is:
- Read next release version from .changes/.next (e.g., `v0.19.9`) instead of hardcoding `v0.0.0`
- Dev versions now use format `v0.19.9-YYMMDD-dev-<digest|commit>`
- Updated `IsDevVersion()` to detect `-dev-` marker in prerelease instead of checking for `v0.0.0` base
- Fixed `ImageTag()` to handle CI detached HEAD by trying `origin/main` with fallback to `HEAD` commit

#### UPDATE - 12/15/2025
  Aligned pseudoversion timestamp format with Go's module spec:
  - 4-digit year (YYYY) instead of 2-digit (YY)
  - UTC timezone for consistency
  - Use current time for dirty builds instead of zero time
> Reference: https://cs.opensource.google/go/x/mod/+/refs/tags/v0.22.0:module/pseudo.go

PR #11576, opened 2025-12-16
perf(version): skip transferring pack files for faster version computation

grouville:
Follow-up from https://github.com/dagger/dagger/pull/11232

> Tentative to implement Erik's idea. It's WIP / an exploration

For PARC users (remote engines), computing the version was slow because we were transferring the entire `.git` directory, including `objects/pack/` which can be 100s of MBs.

The fix is to exclude pack files and adapt git operations to work without them. Since `git.Head().Tree()` needs pack files to do a checkout, we can't use it anymore. Instead, we mount the lightweight `.git` metadata and let git read from it directly.

For tag lookup, we parse packed-refs ourselves rather than shelling out.

Result: ~ -1s locally, bigger wins for remote engines.

### DIFF
--- a/version/main.go
+++ b/version/main.go
@@ -18,6 +18,7 @@ func New(
 	// A git repository containing the source code of the artifact to be versioned.
 	// +optional
 	// +defaultPath="/.git"
+	// +ignore=["objects/pack/", "rr-cache"]
 	git *dagger.Directory,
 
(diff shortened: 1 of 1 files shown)


grouville:
Temporarily closing in favor of #11580 + git infos via session-attachable

PR #11577, opened 2025-12-16
chore: extend Changeset API with merge

eunomie:
```go
changes := firstChanges.WithChanges(extraChanges)
```

In case of a conflict (file modified in both, or modified and deleted, etc) an error is raised.
This behavior can be controlled by `onConflict` with those possible values:

- `fail`: the default behavior, fail when a conflict exists
- `skip`: ignore the conflict but merge the rest of the changes
- `preferSelf`: use the changes from the parent changeset to resolve
  the conflict
- `preferOther`: use the changes from the extra changeset to resolve
  the conflict

Fixes: #11189

### DIFF
--- a/.changes/unreleased/Added-20260126-162107.yaml
+++ b/.changes/unreleased/Added-20260126-162107.yaml
@@ -1,0 +1,8 @@
+kind: Added
+body: |-
+    Merge multiple changesets using changeset.withChangest(other) or changeset.withChangesets(others...)
+    Multiple strategies are available in case of conflict, and uses git under the hood to merge the changesets.
+time: 2026-01-26T16:21:07.146023+01:00
+custom:
+    Author: eunomie
+    PR: "11577"


shykes:
> In case of a conflict (file modified in both, or modified and deleted, etc) an error is raised. It's still possible to force it to continue with the following flag:
> 
> ```
> continueOnConflicts: true
> ```
> 
> If the flag is applied, it's a last write wins strategy.

API nit: I think `continueOnConflicts` is too ambiguous, here's a more explicit proposal:

```graphql
type ChangeSet {
  withChanges(other: ChangeSet!, onConflict: OnMergeConflict=FAIL, granularity: MergeGranularity=FILE): ChangeSet
}

"What to do on a merge conflict"
enum OnMergeConflict {
  "A conflict causes the merge operation to fail"
  FAIL
  "A conflict is skipped, the merge operation continues"
  SKIP
  "The conflict is resolved by applying the version of the calling changeset"
  PREFER_SELF
  "The conflict is resolved by applying the version of the other changeset"
  PREFER_OTHER
}

enum MergeGranularity {
  "Each file path is treated as an atomic change unit"
  FILE
  "Attempt intra-file merging of non-overlapping edits"
  HUNK
}
```

I believe this implementation only supports file granularity, and the default conflict resolution strategy is fail.

eunomie:
> ```graphql
>   "A conflict is skipped, the merge operation continues"
>   SKIP
> ```

What means `skip` here? To ignore the conflict and keep the file as in the `before` directory?

shykes:
> > ```graphql
> >   "A conflict is skipped, the merge operation continues"
> >   SKIP
> > ```
> 
> What means `skip` here? To ignore the conflict and keep the file as in the `before` directory?

Yes I guess so. I guess it depends if the `ChangeSet` is lazy or eager??

eunomie:
I tried a new version based on the above API proposal.

For now I only have a `changeset` granularity. As a first step it's easier than dealing with files but it allows to start crafting the API.

In this situation I have the following merge strategies in case of a conflict:
- `fail`
- `prefer self`, so completely ignore the other changeset
- `prefer other`, so completely ignore the first changeset

And I added a `overwrite`: write the other changeset on top of the first, without to care about conflicts.

I'm still exploring the different possibilities, I'm not sure this `changeset` granularity is the right one.
`changeset` granularity + `overwrite` should in fact be equivalent to a `file` granularity and a `prefer other` merge strategy: last (other) write wins.

And maybe if we want to completely discard self or other in case of a conflict we want to error instead, because it means we are not adding the changesets anymore.

Just thinking out loud.

eunomie:
So, I have a version that looks like to work.
I haven't implemented the `hunk` granularity, but `file` granularity is working with the 4 different strategies:
- `fail`
- `skip` (revert the file to the initial state but apply other changes)
- `prefer self`
- `prefer other`

This is "hand made" by manipulating after directories of changesets before to apply them both, but that looks like to work and they are covered by integration tests.

Maybe a better solution could be to also use `git` and direct it to fix the conflicts the way we want. That's probably the solution for `hunk` level conflict resolution, but maybe as a follow up.

eunomie:
I have issues with the file granularity. Something weird.
But the granularity is not used for now, only `file` is defined. So it's better to just remove the flag for now, and re-introduce it when we'll have at least two values.

With that, the merge feature looks good to me know.
The two test failures I've seen looks like to be unrelated to those changes as they also occurs in other branches (I've seen them in dependabot generated branches).

And PR is now green 🎉

eunomie:
Failing test looks unrelated

docs/docs-graphql/schema.graphqls
```diff
@@ -235,13 +235,47 @@ type Changeset {
...
+
+  """
+  The conflict is resolved by applying the version of the calling changeset
+  """
+  PREFER_SELF
+
+  """
+  The conflict is resolved by applying the version of the other changeset
+  """
+  PREFER_OTHER
```
vito: WDYT of aligning this with Git's `ours` vs. `theirs`?

Ignoring for a moment that sometimes it's unintuitive... (like when rebasing)

Mostly just poking to see if you thought about it already 😄

eunomie: I haven't thought about that before.
I don't have strong opinions, I think I prefer `self`/`other` and I never use `ours`/`theirs` in git.
But especially if then we implement hunk level merge, it will be even closer to git, so maybe it's best to align now on that. I'll have a look.

core/changeset.go
```diff
@@ -125,11 +126,106 @@ type Changeset struct {
...
+		}
+		after, err := id.Load(ctx, srv)
+		if err != nil {
+			return fmt.Errorf("load after: %w", err)
+		}
+		ch.After = after
+		ch.afterIDEncoded = ""
+	}
+	return nil
+}
```
eunomie: This has been generated by Claude code.
I'm not sure to understand if that's good or not. This feels a bit weird to me, but it looks like it's working.

This is used in conjunction with the new `DagOpChangesetWrapper`. As soon as I wrapped my functions with `DagOpWrapper` it started to panic. And this fixes it.

But I'd be happy to have more eyes on it to see if that's a good thing or not.

cc @grouville

vito: Strange, it feels like this shouldn't be needed, since `dagql.ObjectResult[T]` already implements a `MarshalJSON` that delegates to `Result[T]` which delegates to `ID[T]` which already does the same encoding/decoding you have here.

Maybe the issue is just with having to initialize `pathsOnce`?

Trying it out locally now.

vito: Ah - `ObjectResult[T]` implements `MarshalJSON`, but not `UnmarshalJSON`, because that needs a `ctx` and a server to load the ID from.

vito: Pushed a slight simplification, so we can lean on the existing ID serialization code at least.

core/schema/directory.go
```diff
@@ -308,7 +310,25 @@ func (s *directorySchema) Install(srv *dagql.Server) {
 			Doc(`Files and directories that existed before and were updated in the newer directory.`),
 		dagql.NodeFunc("removedPaths", DagOpWrapper(srv, s.changesetRemovedPaths)).
 			Doc(`Files and directories that were removed. Directories are indicated by a trailing slash, and their child paths are not included.`),
+		dagql.NodeFunc("withChangeset", DagOpChangesetWrapper(srv, s.changesetWithChangeset)).
+			Doc(`Add changes to an existing changeset`,
+				`By default the opperation will fail in case of conflicts, for instance a file modified in both changesets. The behavior can be adjusted
+				using onConflict argument`).
```
vito:
```suggestion
				`By default the operation will fail in case of conflicts, for instance when a file is modified in both changesets. The behavior can be adjusted
				using the onConflict argument.`).
```

docs/docs-graphql/schema.graphqls
```diff
@@ -235,13 +235,64 @@ type Changeset {
...
 """
 scalar ChangesetID
 
+"""Mediatypes to use in published or exported image metadata."""
+enum ChangesetMergeConflict {
+  """A conflict causes the merge operation to fail"""
+  FAIL
+
+  """A conflict is skipped, the merge operation continues"""
+  SKIP
```
vito: IMO we should just drop support for this. I don't know when someone would ever want that. A saner option would be to leave merge conflicts in-place, which we can't easily do currently, but we'd get that for free if we used `git` to apply changes.

shykes: Works for me 👍

eunomie:
Thanks a lot for the review @vito 

I'm giving a try with Claude to see if it can replace the current behaviour with git. It would be better if we can have it right before to merge. If it's really too complexe maybe we can merge it as it, but let's see.

eunomie:
I pushed a new commit. This is the implementation of the merge using `git` under the hood.
Disclaimer: it has been generated by Claude with some guidance, changeset integration tests and go linter are green on my machine, but I haven't yet made a human code review and I haven't tested it extensively. But that could be a base of work.
I'll soon refresh my generator branch on top, to see it working on real use cases.

vito:
Nice - seems like a good start!

Ended up more complicated than I thought, which could be legit since my suggestion was naive. AFAIK `git apply` works without needing it to actually be a `git` repo, so I figured (hoped) it would just translate 1:1 to a `git apply` call with either `--theirs` or `--ours` based on the argument, and `-3` to support a 3-way merge with `Changeset.withChangesets`. But it looks like it set up temporary repos, and nuked `withChangesets` instead? Bad bot!

I see some fallback related code in there too which I know Claude is keen to do, those might make more sense to just error (having not looked too deeply, saying this based on my own Claude memory)

Anyway thanks for looking into it! And feel free to back out if it turns into a quagmire. :pray:

eunomie:
I'm having a new round.
I'll remove the leaveconflicts option for now. The granularity is currently set to file, so we can fail, ours, theirs, and that should be good enough for now.
By removing leaveconflicts option I think it will simplify how it works.
And we can go back with it when we will add the hunk level granularity, that this time might require a more advance way to work.
That's just brain dump, I'll see how it goes to find the right compromise between complexity and features. I think to focus on file level might be good enough for now if it simplifies the work.

eunomie:
Once merged, the `changesetMerge` function in `.dagger/util.go` could be replaced by:

```go
// Merge Changesets together
func changesetMerge(changesets ...*dagger.Changeset) *dagger.Changeset {
	if len(changesets) == 0 {
		return nil
	}
	return changesets[0].WithChangesets(
		changesets[1:],
		dagger.ChangesetWithChangesetsOpts{
			OnConflict: dagger.ChangesetsMergeConflictFailEarly,
		})
}
```

Will be in a follow up PR.

(until we got generators, then the function will be removed)

core/schema/directory.go
```diff
@@ -1094,6 +1114,149 @@ func (s *directorySchema) exportLegacy(ctx context.Context, parent dagql.ObjectR
...
+// ChangesetsMergeConflict only supports MULTI_FAIL_EARLY and MULTI_FAIL (no -X ours/theirs with octopus merge).
+type ChangesetsMergeConflict string
+
+var ChangesetsMergeConflictEnum = dagql.NewEnum[ChangesetsMergeConflict]()
+
+var (
+	MultiFailEarlyOnMergeConflicts = ChangesetsMergeConflictEnum.Register("MULTI_FAIL_EARLY",
+		`Fail before attempting merge if file-level conflicts are detected between any changesets`)
+	MultiFailOnMergeConflicts = ChangesetsMergeConflictEnum.Register("MULTI_FAIL",
+		`Attempt the octopus merge and fail if git merge fails due to conflicts`)
```
eunomie: `MULTI_` prefix is here to avoid conflict with the other `FAIL` and `FAIL_EARLY` values, when used through codegen

core/changeset.go
```diff
@@ -254,7 +327,7 @@ func (ch *Changeset) AsPatch(ctx context.Context) (*File, error) {
 				}
 				defer patchFile.Close()
 
-				cmd := exec.Command("git", "diff", "--no-prefix", "--no-index", "a", "b")
+				cmd := exec.CommandContext(ctx, "git", "diff", "--binary", "--no-prefix", "--no-index", "a", "b")
```
eunomie, line 330: `CommandContext` ensures the command is terminated if the context is cancelled.

`--binary` as the patches will be used to merge conflicts, so we want to ensure binary files are correctly handled.

core/changeset.go
```diff
@@ -322,3 +395,634 @@ func (ch *Changeset) Export(ctx context.Context, destPath string) (rerr error) {
...
+	for _, c := range conflicts {
+		if errors.Is(c.Err, ErrModifiedRemoved) {
+			paths = append(paths, c.Path)
+		}
+	}
+	return paths
+}
+
+// CheckConflicts detects conflicts using pre-computed path sets for O(1) lookups
+func (ch *ChangesetPaths) CheckConflicts(other *ChangesetPaths) Conflicts {
```
eunomie, line 444: This allows to catch file level conflicts and fail early is that's what is wanted by the caller.
For instance we could imagine that we don't want a same file to be modified by two changesets, even if the merge would be possible.

But as git doesn't handle that level of granularity, and as we already have the tools to know what files are modified, removed, added, this solves the problem quite easily.

core/schema/directory.go
vito: Do you remember which codegens failed on that? Enums are _supposed_ to be namespaced to their type (that's how it works in GraphQL), maybe some SDKs need to be fixed :thinking: (...in a magical backwards compatible way)

eunomie: 🤔 I can't find a way to reproduce it easily. I'll try again.
In my memories it was about Go. But I can't find the right way.

grouville: > Do you remember which codegens failed on that? Enums are _supposed_ to be namespaced to their type (that's how it works in GraphQL), maybe some SDKs need to be fixed 🤔 (...in a magical backwards compatible way)

I agree with Alex on that point ➕

eunomie: All right, I'm making progress here. I'm able to reproduce the issue on the CI:

```
Error: call function "Fn": exit code: 1 [traceparent:f276c641c8a06f1385f906ce07ea0c40-e1b6c6b7003d3b04]
Stderr:
# dagger/test/internal/dagger
internal/dagger/dagger.gen.go:13572:2: FailEarly redeclared in this block
        internal/dagger/dagger.gen.go:13491:2: other declaration of FailEarly
internal/dagger/dagger.gen.go:13578:2: Fail redeclared in this block
        internal/dagger/dagger.gen.go:13497:2: other declaration of Fail
```

This is without the rename with `MULTI_`. For instance https://dagger.cloud/dagger/traces/f276c641c8a06f1385f906ce07ea0c40?listen=6f4a3405ccf938d0&listen=6e35e97fdb942bb9&listen=0deb8a687923c508&listen=1fc7d0995c6b85ce&listen=1bd46d6c25d665bf&listen=66086b004ce9df52&span=7b1c144dde3dbb50#0deb8a687923c508:L32

What's weird is the same test is working on my machine. But the version is not exactly the same, so I wonder if there isn't a different issue there. I'll investigate more to understand the issue and be sure we have the right fix before to merge.

eunomie: I finally found the issue!!
It's again, about enums...

In short, it's failing in some tests. And I can reproduce it... by removing the `engineVersion` of `dagger.json`.
If I put a value, that works, even an old value.
But as soon as I remove the `engineVersion` (and that's what some tests are doing) then it breaks.

I'm not sure about the exact root cause, and I'll see how to fix that, but I'll keep the same value for both enums for sure.

Better than dealing with `MULTI_` without to understand why.

eunomie: All right, got it.
Scoped enum values (to allow to have two enums with identical values) has been introduced in v0.15.0 in the Go codegen.
So before that, it's impossible and there's a conflict during codegen.
And that's why some tests are not working anymore.

I've gated `withChangesets` and the corresponding enum values to `v0.15.0` and that should solve the issue.

vito: Amazing, thanks for tracking that down!

eunomie:
If no specific concern, I'm going to merge it today, so it unblocks the work on generators

PR #11580, opened 2025-12-16
core(changeset): lazify the `changeset` API + rely on `git` for efficient diffs

grouville:
## Summary

This PR is follow-up (_n°1/N_) of #11232.  [Skipping the git pack files](https://github.com/dagger/dagger/pull/11576) didn't appear to be fruitful in its current state. 

The alternative, is to go [all-in with the Dagger Git API and the changeset implementation](https://discord.com/channels/707636530424053791/1450576067935993928/1450602599446347867).

**Problem**: the current implementation is slow and could benefit some perf tweaks / improvement.

This is one of such optimization:
- Replace glob-all-files approach with targeted bidirectional diffs
- Make Changeset lazy (compute paths on-demand, not upfront)
- Use git as single source of truth for diffs (paths AND patches)

## Context

The Changeset API previously enumerated ALL files in both directories using glob("**"), then computed set differences. For large directories with few actual changes, this was _O(n) where n = total files._

## Approach

Lazy computation: `NewChangeset` no longer computes anything. Path computation (`addedPaths`, `modifiedPaths`, `removedPaths`) happens on-demand via dagql resolvers.

Git-based diffing: Uses `git diff --no-index` instead of buildkit operations:
- `git diff --name-status` for path enumeration
- `git diff --quiet` for fast `isEmpty` check (exits on first difference)
- Single `git diff` for patch generation (was _N per-file diffs before_)

Why git CLI? It's fast. See https://github.com/runxiyu/furgit --> git's diff is highly optimized C code, faster than any Go library alternative.

### ModuleVersion API changes

As part of the PR, I added the moduleVersion changes, as it's the main reason I did the changes.

With my empiricial non-precise tests, `moduleVersion` now takes:
- ~4.2s on [this branch](https://dagger.cloud/dagger/traces/ac2fcd247e6a898b1fab73143f58d01a?span=f36cc2e00a0ef1e2)
- ~6s on main (without relying on Dagger)
- ~16s with the [changeset API as on main](https://dagger.cloud/dagger/traces/b94bb3215128baf8bd27a7de908130f5?span=ab64e5ea94fed8ae)

To test it locally: 
1) checkout this branch
2) `./hack/dev `
3) prune the cache (make sure it doesn't pollute)
4) `./hack/with-dev dagger call engine-dev container terminal`

> ⚠️ Make sure that it's the dev engine loading the engine, I got bitten by it. e.g. `dagger call engine-dev container terminal` uses the main implementation to load it

####  Trade-offs && follow-ups

Requires mounting directories to run git, and git diff execution is not cached at the dagql level yet.

_Follow-ups_
- [x] It might be worth caching in-session the `ComputePaths`

### DIFF
--- a/core/changeset.go
+++ b/core/changeset.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"slices"
+	"sync"
 	"syscall"
 
 	"dagger.io/dagger/telemetry"
@@ -26,8 +27,9 @@ import (
 
 func NewChangeset(ctx context.Context, before, after dagql.ObjectResult[*Directory]) (*Changeset, error) {
 	return &Changeset{
-		Before: before,
-		After:  after,
+		Before:    before,
+		After:     after,
+		pathsOnce: &sync.Once{},
 	}, nil
 }
 
@@ -41,6 +43,13 @@ type ChangesetPaths struct {
 // ComputePaths computes the added, modified, and removed paths using git diff.
 // This must be called from a dagql resolver context where buildkit session is available.
 func (ch *Changeset) ComputePaths(ctx context.Context) (*ChangesetPaths, error) {
+	ch.pathsOnce.Do(func() {
+		ch.cachedPaths, ch.pathsErr = ch.computePathsOnce(ctx)
+	})
+	return ch.cachedPaths, ch.pathsErr
+}
+
+func (ch *Changeset) computePathsOnce(ctx context.Context) (*ChangesetPaths, error) {
 	if ch.Before.ID().Digest() == ch.After.ID().Digest() {
 		return &ChangesetPaths{}, nil
 	}
@@ -115,6 +124,10 @@ func (ch *Changeset) withMountedDirs(ctx context.Context, fn func(beforeDir, aft
 type Changeset struct {
 	Before dagql.ObjectResult[*Directory] `field:"true" doc:"The older/lower snapshot to compare against."`
 	After  dagql.ObjectResult[*Directory] `field:"true" doc:"The newer/upper snapshot."`
+
+	pathsOnce   *sync.Once
+	cachedPaths *ChangesetPaths
+	pathsErr    error
 }
 
 func (*Changeset) Type() *ast.Type {


grouville:
Failure on `test-split:test-modules` seems to be a flake, the test passes locally

core/changeset_git.go
```diff
@@ -0,0 +1,189 @@
...
+	out, err := runGitDiff(ctx, oldDir, newDir)
+	if err != nil {
+		return fileChanges{}, err
+	}
+	return parseGitOutput(out, oldDir, newDir), nil
+}
+
+// directoriesAreIdentical returns true if both directories have identical content.
+func directoriesAreIdentical(ctx context.Context, dir1, dir2 string) (bool, error) {
+	cmd := exec.CommandContext(ctx, "git", "diff", "--no-index", "--quiet", dir1, dir2)
```
grouville, line 32: ~~it should be `-z` too to properly escape the spaces too here~~

> Update: after re-reading, the `-z` flag is not needed for `directoriesAreIdentical` because `--quiet` suppresses all output and the `-z` flag formats output with `NUL` delimiters. With the `--quiet` option, there's nothing to format 😇

core/changeset.go
```diff
@@ -24,140 +24,97 @@ import (
...
+	Modified   []string
+	Removed    []string
+	AllRemoved []string
 }
 
-// computeChanges calculates added, changed, and removed files/paths
-func (ch *Changeset) computeChanges(ctx context.Context) error {
+// ComputePaths computes the added, modified, and removed paths using git diff.
+// This must be called from a dagql resolver context where buildkit session is available.
+func (ch *Changeset) ComputePaths(ctx context.Context) (*ChangesetPaths, error) {
```
vito, line 45: Agree that this should be cached somehow since clients will likely be requesting all three (or four), possibly in parallel. Maybe a simple `sync.Once` internal to the `Changeset` would do the trick, similar to `LLM`?

grouville: Added the sync 🙏

grouville:
Failures seem to be unrelated / flakes

grouville:
cc @eunomie i'll let you do the merge as it might conflict with your PR

eunomie:
Merged, I'll rebase mine on top of main now :-)
And it will improve my tests, so all good there 👍

PR #11586, opened 2025-12-19
fix(shell): preserve shell exit codes in TUI mode

grouville:
#### Problem
`dagger -M -c '.exit 20'` exits with code 1 instead of 20 when using the TUI, but works correctly with `--progress=plain`.

#### Why this happened
The shell wraps exit codes in `ExitError` to tell the TUI what code to use. But this wrapping happened *after* the TUI already stored the raw error. When the TUI later checked for an `ExitError`, it didn't find one, so it defaulted to exit code 1.

#### The fix
Move the `ExitError` wrapping to happen *before* the TUI stores the error, so the exit code is preserved through the entire chain.

- shell.go: Wrap the error inside the callback, not after it returns
- frontend_pretty.go: Check for existing `ExitError` before defaulting to 1
- Added host-based test (TTY mode needs a real terminal)

### DIFF
--- a/cmd/dagger/shell.go
+++ b/cmd/dagger/shell.go
@@ -51,14 +51,11 @@ var shellCmd = &cobra.Command{
 			handler := newShellCallHandler(dag, Frontend)
 
 			err := handler.RunAll(ctx, args)
-			if err == nil {
-				return nil
-			}
 
 			// Wrap exit status in ExitError so the TUI preserves the exit code
 			// and doesn't print a redundant error message.
 			var es interp.ExitStatus
-			if errors.As(err, &es) {
+			if handler.tty && errors.As(err, &es) {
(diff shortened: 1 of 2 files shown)


marcosnils:
let's not merge this yet.  There's something weird since I just ran the  test in `main` and it passed. Let me look a bit more into it.

marcosnils:
@grouville added some commits to fix the test and re-introduced some code which seems to have been removed/added unintentionally? LMK what you think.

grouville:
> @grouville added some commits to fix the test and re-introduced some code which seems to have been removed/added unintentionally? LMK what you think.

Thanks for the TUI test, I was having a hard time reproducing in the integration test. Didn't know about this netflix lib 🙏 
As part of the unrelated code, it was intentional but I should have made it a proposal as part of another PR / commit 👍

PR #11610, opened 2025-12-30
fix: handle deeply nested directories in cache size calculation

### LINKED ISSUE #11609: cache size calculation fails with ENAMETOOLONG on deeply nested directories
## Problem

> [!NOTE]
> AI wrote this issue. I have verified correctness to the best of my ability.

cache manager can't calculate disk usage for snapshots with deeply nested directories. hits ENAMETOOLONG when walking filesystem.

```
level=ERROR msg="failed to calculate size" error="lstat .../node_modules/.old-57CC/.../node_modules/.old-5AB6/.../astro-opengraph-images/...: file name too long"
```

seen with npm packages like astro-opengraph-images that create circular/deeply nested node_modules structures.

## Impact

- size set to 0 instead of actual usage
- might break LRU eviction (thinks snapshot is empty)
- log spam

## Root Cause

`cr.cm.Snapshotter.Usage()` at `refs.go:497` calls containerd snapshotter which walks filesystem without:
- depth limits
- circular reference detection  
- ENAMETOOLONG handling

## Proposed Fix

add safe filesystem walker with:
- max depth limit (1024 levels)
- inode tracking for circular reference detection
- graceful ENAMETOOLONG handling (skip problematic paths)

fallback to safe walker when regular Usage() fails with ENAMETOOLONG.

**files:** `internal/buildkit/cache/refs.go`

implementation uses `filepath.WalkDir` with depth checking. only activates on error so zero perf impact for normal snapshots.

shepherdjerred:
Fixes #11609

## Changes

Adds safe filesystem walker with protections against deeply nested/circular directory structures that cause ENAMETOOLONG errors during cache size calculation.

**New functions:**
- `safeDiskUsage()` - walks filesystem with max depth (1024), inode tracking for circular refs, ENAMETOOLONG handling
- `safeSnapshotUsage()` - calculates usage from snapshot mounts using safe walker
- `isFileNameTooLongError()` - detects ENAMETOOLONG errors

**Modified:**
- `size()` method in `refs.go` - falls back to safe calculation when snapshotter.Usage() fails with ENAMETOOLONG

## Behavior

- Normal snapshots: no change, uses regular snapshotter.Usage()
- Problematic snapshots: detects ENAMETOOLONG, automatically uses safe walker
- Returns actual size instead of 0
- Logs warnings when circular refs or max depth hit

## Testing

- Code builds successfully
- Needs runtime testing with actual deeply nested snapshots

**Files changed:** `internal/buildkit/cache/refs.go`

### DIFF
--- a/internal/buildkit/cache/refs.go
+++ b/internal/buildkit/cache/refs.go
@@ -78,10 +78,11 @@ func safeDiskUsage(ctx context.Context, root string) (int64, error) {
 			return filepath.SkipDir
 		}
 
-		// Get file info to check for circular references
-		info, err := d.Info()
-		if err != nil {
-			return nil // Skip entries we can't stat
+		// Get file info to check for circular references and size.
+		// Skip entries we can't stat rather than failing the entire walk.
+		info, _ := d.Info()
+		if info == nil {
+			return nil
 		}
 
 		// Check for circular references using inode number


sipsma:
continuing here: https://github.com/dagger/dagger/issues/11609#issuecomment-3774225399, the approach here has some issues but more fundamentally we need to nail down the root cause first

PR #11611, opened 2025-12-30
fix: Dagger logs alarming 'HTTP HEAD ERROR' for routine registry cache misses

### LINKED ISSUE #11604: ✨ [UX/DX] Dagger logs alarming "HTTP HEAD ERROR" for routine registry cache misses

I am running Dagger in a centralized "Build Service" where a CLI triggers remote builds. I connect to the Dagger Engine via the Go SDK and pipe the raw log stream directly back to the user's terminal using dagger.Connect(ctx, dagger.WithLogOutput(http.ResponseWriter), ...).

I want to present this log stream to developers without them seeing prominent red ERROR logs when Dagger essentially hits a "cache miss" (404) while checking for existing image layers in the registry.

Example of the current noisy output during a successful build:
```
13  : remotes.docker.resolver.HTTPRequest
14  : ┆ HTTP HEAD
14  : ┆ HTTP HEAD ERROR [0.3s]
13  : remotes.docker.resolver.HTTPRequest ERROR [0.3s]
```

We prioritize a "No WTF moments" Developer Experience. Currently, when a developer sees a wall of red errors, their immediate emotional response is "The build failed." It causes confusion and unnecessary panic, making the logs look broken even when the operation is 100% successful.

More importantly, this behavior trains users to ignore ERROR logs, which is a dangerous habit. We want ERROR to mean "Action Required" or "Process Failed," not "Routine Check Returned False."

Here is an image of what it looks like for a dev:

Image

We are currently forcing developers to tolerate the noise and explaining that "it's normal."

The only technical workaround would be to implement a brittle "Filtering Writer" in our application that intercepts the raw byte stream, parses for these specific error strings, and discards them before writing to the HTTP response. We would prefer not to maintain this kind of log-scraping logic downstream.

majiayu000:
## Summary
This PR fixes #11604

## Changes
- `dagql/dagui/spans.go`

### DIFF
--- a/dagql/dagui/spans.go
+++ b/dagql/dagui/spans.go
@@ -386,6 +386,12 @@ func (snapshot *SpanSnapshot) ProcessAttribute(name string, val any) { //nolint:
 		// parent failed, since some happy paths might involve _expected_ failures
 		snapshot.Encapsulated = true
 
+	case "http.method":
+		// encapsulate HTTP client spans by default; routine operations like
+		// registry cache checks may return 404s which are expected failures
+		// and should not be shown as alarming errors to users
+		snapshot.Encapsulated = true
+
 	default:
 		if snapshot.ExtraAttributes == nil {
 			snapshot.ExtraAttributes = make(map[string]json.RawMessage)


marcosnils:
I'm closing this since it's not addressing the underlying issue. Looks to me like an AI generated PR.

PR #11612, opened 2025-12-30
fix plain output to filter unwanted spans in default verbosity

### LINKED ISSUE #11604: ✨ [UX/DX] Dagger logs alarming "HTTP HEAD ERROR" for routine registry cache misses

marcosnils:
plain output was showing more information that it should when using the
default verbosity level.

Fixes #11604

### DIFF
--- a/docs/current_docs/reference/configuration/custom-runner.mdx
+++ b/docs/current_docs/reference/configuration/custom-runner.mdx
@@ -61,10 +61,10 @@ then the CLI will instead connect to the endpoint specified there. This
 environment variable currently accepts values in the following format:
 
 1. `container://<container name>` - Connect to the runner inside the given host container.
-    - Requires a local [container runtime](../../container-runtimes/).
+    - Requires a local [container runtime](/reference/container-runtimes/).
     - Can force a specific container runtime using `container+<runtime>://<container name>`, e.g. `container+podman://dagger-engine`.
 1. `image://<container image reference>` - Start the runner in Docker using the provided container image, pulling it locally if needed
-    - Requires a local [container runtime](../../container-runtimes/).
+    - Requires a local [container runtime](/reference/container-runtimes/).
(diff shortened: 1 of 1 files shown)


eunomie (approved):

vito:
oop, I think this broke a test: https://dagger.cloud/dagger/traces/f5a8ab7949cf0eb527bd0e39f8fa5662?span=0820bcd428347e20

gonna just put up a revert PR to unblock

PR #11615, opened 2025-12-30
Debug filesync flake

sipsma:
Debugging the `file not found` flake seen in module runtime tests (based on https://github.com/dagger/dagger/pull/11548)


sipsma:
managed to repro locally after a while, will continue there

PR #11616, opened 2026-01-02
test(python): add test for DefaultPath on constructor field without dagger.field()

### LINKED ISSUE #11554: 🐞 `DefaultPath` doesn't work in Python
### What is the issue?

Hello! 

I've been having issues with `DefaultPath` since forever (see my comment [here](https://github.com/dagger/dagger/issues/10994#issuecomment-3270660324)). 

Basically, it doesn't work neither in `@function` or as an `@object_type` attribute. 

I've tried: 

```py
@object_type
class ResearchDagger:
    root_dir: Annotated[Directory, DefaultPath(".")]
```

```py
@object_type
class ResearchDagger:
    @function
    def build(self, root_dir:  Annotated[Directory, DefaultPath(".")]): ...
```

I've tried providing default values such as `None` or `...` (just for the record, I know I'm not supposed to). 

I am always getting: 

```
! required flag(s) "root-dir" not set
```

### Dagger version

dagger 0.19.8

### Steps to reproduce

1. Call `dagger init --sdk python`
2. Edit `main.py`: 

```py
import dagger
from dagger import dag, function, object_type, Directory, DefaultPath
from typing import Annotated

@object_type
class Tmp:
    path: Annotated[Directory, DefaultPath(".")]

    @function
        """Returns a container that echoes whatever string argument is provided"""

    @function
    async def grep_dir(self, directory_arg: dagger.Directory, pattern: str) -> str:
        """Returns lines that match a pattern in the files of the provided Directory"""
        return await (
            dag.container()
            .from_("alpine:latest")
            .with_mounted_directory("/mnt", directory_arg)
            .with_workdir("/mnt")
            .with_exec(["grep", "-R", pattern, "."])
            .stdout()
        )
```

3. Observe the following: 

```
> dagger call container-echo
✔ connect 0.2s
✔ load module: . 4.1s
! required flag(s) "string-arg" not set
```

### Log output

```
! required flag(s) "string-arg" not set
```

majiayu000:
Fixes #11554

## Changes
- Add integration test to verify DefaultPath works on constructor fields without dagger.field()
- Fix trailing whitespace in existing test (minor style fix)

The test verifies that when a constructor field uses `Annotated[dagger.Directory, DefaultPath(".")]` without `dagger.field()`, the DefaultPath is properly registered with the engine and the parameter becomes optional with the default path from context.

### DIFF
--- a/core/integration/module_python_test.go
+++ b/core/integration/module_python_test.go
@@ -1768,12 +1768,38 @@ func (PythonSuite) TestIgnoreConstructorArg(ctx context.Context, t *testctx.T) {
         @dagger.object_type
         class Test:
             source: Annotated[
-                dagger.Directory, 
+                dagger.Directory,
                 dagger.DefaultPath("/"),
                 dagger.Ignore([".venv"]),
             ] = dagger.field()
         `).
 		With(daggerCall("source", "entries", "--json")).
+		Stdout(ctx)
+
+	require.NoError(t, err)
+	require.Contains(t, gjson.Parse(out).Value(), "dagger.json")
+}
+
+func (PythonSuite) TestDefaultPathConstructorWithoutField(ctx context.Context, t *testctx.T) {
+	// This test verifies that DefaultPath works on constructor fields
+	// even when dagger.field() is not used (issue #11554)
+	c := connect(ctx, t)
+
+	out, err := pythonModInit(t, c, `
+        from typing import Annotated
+        import dagger
+        from dagger import DefaultPath
+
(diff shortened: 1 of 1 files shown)


marcosnils:
this PR is incorrect since it's already testing something that's tested  here https://github.com/dagger/dagger/blob/9419454effd67b2d0abff2a18f46490c041823bb/core/integration/module_python_test.go?plain=1#L1761-L1781

closing since it seems to be AI related.

PR #11621, opened 2026-01-04
feat: engine client docker driver proxy env handling

### LINKED ISSUE #11553: ✨ Add custom-proxy engine setting

TL/DR:
Please add support of custom-proxy variables in the engine.json setup file:
https://docs.dagger.io/reference/configuration/engine#custom-proxy

Without internet connection the dagger engine fails to run.

To access the internet, corporations usually have some kind of proxy. Usually this is set using the environment variables (http_proxy, https_proxy, no_proxy, so on and so forth). This setting can vary between clients – some people run some kind of cntlm, or px proxy to avoid sending passwords in the requests.

Since the dagger cli is greatly maintained and new versions come often, the rebuild of the engine for each upgrade is proving to be hard to handle in a big department.

Every user needs to rebuild the engine image locally overriding the http proxy variables for each release.

frantjc:
allows Dagger to "just work" with its default configuration on machines that need http(s) proxies e.g.

```sh
HTTPS_PROXY=http://localhost:8080 dagger
```

will spawn an engine container with `HTTPS_PROXY=http://host.docker.internal:8080` when using the `docker` backend

on my work machine, for instance, this keeps me from needing to manage my own engine container with these env vars

for `nerdctl` we could use [`host.lima.internal`](https://github.com/containerd/nerdctl/issues/747#issuecomment-1065705918) if we are comfortable assuming that most `nerdctl` users are using it via Lima. it also looks like we could use [`host.finch.internal`](https://github.com/runfinch/finch/pull/1085) for `finch` if we are comfortable assuming that most `finch` users are on macOS

fixes #11553 (in a different way than requested)

EDIT: I noticed that CI doesn't seem to run the tests for `github.com/dagger/dagger/engine/client/drivers`, so here's the output of me running it for `podman` and `docker` on my machine:

```
$ DRIVER_TEST=1 go test ./engine/client/drivers/...
ok  	github.com/dagger/dagger/engine/client/drivers	(cached)
```

EDIT: implementation changed to pass config to engine via `~/.config/dagger/engine.json`

### DIFF
--- a/cmd/engine/main.go
+++ b/cmd/engine/main.go
@@ -292,6 +292,10 @@ func main() { //nolint:gocyclo
 			return err
 		}
 
+		if err := configureProxy(cfg); err != nil {
+			return err
+		}
+
 		bklog.G(ctx).Debug("setting up engine networking")
 		networkContext, cancelNetworking := context.WithCancelCause(context.Background())
 		defer cancelNetworking(errors.New("main done"))
@@ -839,4 +843,26 @@ func setupNetwork(ctx context.Context, netName, netCIDR string) (*networkConfig,
 		Bridge:        bridge,
 		CNIConfigPath: cniConfigPath,
 	}, nil
+}
+
+func configureProxy(cfg config.Config) error {
+	if cfg.Proxy == nil {
+		return nil
+	}
+
+	for key, value := range map[string]string{
+		"HTTP_PROXY": cfg.Proxy.HTTPProxy,
+		"HTTPS_PROXY": cfg.Proxy.HTTPSProxy,
+		"NO_PROXY": strings.Join(cfg.Proxy.NoProxy, ","),
+		"ALL_PROXY": cfg.Proxy.AllProxy,
+		"FTP_PROXY": cfg.Proxy.FTPProxy,
+	} {
+		if value != "" {
+			if err := os.Setenv(key, value); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
 }
--- a/engine/client/drivers/docker.go
+++ b/engine/client/drivers/docker.go
@@ -71,7 +71,7 @@ func (d docker) ContainerRun(ctx context.Context, name string, opts runOpts) err
 		args = append(args, "-v", volume)
 	}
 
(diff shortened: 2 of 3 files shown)


marcosnils (commented):
Skimming through this PR there are some things which are caughting my attention that I think we should discuss more.

1. If an engine is already provisioned and you run `HTTP_PROXY=foo dagger` this won't work and will very likely confuse users.  In the same way as any container runtime, there are two levels where you might want to configure a proxy even if it'll almost always be the same one. 1, the Dagger engine itself and 2, the containers that Dagger spans.

I'm thinking that a simple and probably "easier" and more correct alternative here would be to allow setting the proxy values in the `dagger.json` file so when the engine starts can use them without users having to start the engine manually to set those up. @sipsma WDYT?

2.  The `host.docker.internal` thing doesn't look ok to me since it's doing some custom setting which doesn't always work. If you need your engine to use some proxy that's running in your machine why not just setting the HTTP_PROXY to `host.docker.internal` or whatever endpoint you need?  adding this specific `host.docker.internal` special case doesn't seem right since it's not universal, it can be different depending how users start their engine and it's not enabled in all platforms by default (linux)

frantjc:
@marcosnils excellent point about `host.*.internal` not being supported by all `docker`s or `podman`s--I did not know that was the case (not to mention, what if `DOCKER_HOST` refers to a remote machine?) I've swapped the implementation to passing proxy configuration to the engine thru `~/.config/dagger/engine.json`, which I think is what you were suggesting. an example config file might look like:

```json
{
  "proxy": {
    "httpProxy": "http://host.docker.internal:8080",
    "httpsProxy": "http://host.docker.internal:8080",
    "noProxy": ["dagger.io"]
  }
}
```

I don't like it as much because now users have to separately maintain e.g. `HTTP_PROXY=http://localhost:8080` in their environment and `proxy.httpProxy: http://host.docker.internal:8080` in their `~/.config/dagger/engine.json`, but I don't see any better options.

frantjc:
superceded by https://github.com/dagger/dagger/pull/11773

PR #11624, opened 2026-01-05
modules/{alpine,wolfi}: support additional repos

moosingin3space:
This patch adds two new options to the Alpine and Wolfi modules:

- `ExtraRepositories`, which allows the user to configure additional repositories for their Wolfi/Alpine container, and
- `ExtraKeyURLs`, which allows the user to add the required public keys to validate the provided repositories.

This allows for easy integration of extra repositories.


shykes:
Thank you for the contribution!

PR #11627, opened 2026-01-06
README: refocus on software delivery

shykes:
We got the feedback that our current positioning ("workflows" that are "great for CI/CD and for AI agents") is confusing and dilutes the original value.

This new version refocuses clearly on software delivery: building, testing and shipping code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,42 +1,50 @@
-## What is Dagger?
+## Dagger: a better way to ship
 
-Dagger is an open-source runtime for composable workflows. It's perfect for systems with many moving parts and a strong need for **repeatability**, **modularity**, **observability** and **cross-platform support**. This makes it a great choice for AI agents and CI/CD workflows.
+Dagger is a platform for automating software delivery. It can build, test and ship any codebase, reliably and at scale.
 
-<p align="center"><img src="docs/static/img/readme/dagger-factory.jpg" width="75%"></p>
+Dagger runs locally, in your CI server, or directly in the cloud. 
 
-## Key Features
+```
+brew install dagger/tap/dagger
+```
 
-- **Containerized Workflow Execution:** Transform code into containerized, composable operations. Build reproducible workflows in any language with custom environments, parallel processing, and seamless chaining.
+## Why Dagger?
 
-- **Universal Type System:** Mix and match components from any language with type-safe connections. Use the best tools from each ecosystem without translation headaches.
+Dagger makes your software delivery *programmable*, *local-first*, *repeatable* and *observable*.
 
-- **Automatic Artifact Caching:** Operations produce cacheable, immutable artifacts — even for LLMs and API calls. Your workflows run faster and cost less.
+**Programmable**. Shell scripts and proprietary YAML are no longer acceptable for automating software delivery. Dagger provides: a complete execution engine and system API; SDKs for 8 languages; an interactive REPL; a rich ecosystem of reusable modules; and more.
 
-- **Built-in Observability:** Full visibility into operations with tracing, logs, and metrics. Debug complex workflows and know exactly what's happening.
+**Local-first**. Once you automate a task with Dagger, it will reliably run on any supported system: your laptop, AI sandbox, CI server, or dedicated cloud infrastructure. The only dependency is a container runtime like Docker.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,42 +1,66 @@
 ## What is Dagger?
 
-Dagger is an open-source runtime for composable workflows. It's perfect for systems with many moving parts and a strong need for **repeatability**, **modularity**, **observability** and **cross-platform support**. This makes it a great choice for AI agents and CI/CD workflows.
+Dagger is an open-source, agent-ready platform for end-to-end testing.
 
-<p align="center"><img src="docs/static/img/readme/dagger-factory.jpg" width="75%"></p>
+AI coding agents are fast but unreliable — they need an external system for trusted feedback on every code change. Dagger provides that feedback: repeatable test execution that agents can call as they develop, and CI can verify during review. It runs locally, in CI, or directly in the cloud.
```
TomChv: > and CI can verify during review

It feels like some words are missing?

marcosnils:
overall changes LGTM! the only thing caught my attention is the emphasis on coding agents from the README PR. Particularly this: 

> AI coding agents are fast but unreliable — they need an external system for trusted feedback on every code change. Dagger provides that feedback: repeatable test execution that agents can call as they develop, and CI can verify during review. It runs locally, in CI, or directly in the cloud.

The first impression after reading that sentence was: "are we going back to highlighting the AI use case as our main strength"?  Just being honest here with my reaction about the paragraph above not surfacing all the "test platform" goodies like checks, scale-out, splitting, observability, etc which elevate your application stack as a whole. Since we're not mentioning that, I am thinking if some users might think that Dagger is only valuable when using agents mostly/only.

README.md
```diff
@@ -1,42 +1,66 @@
...
+With LLMs, code changes are nearly instant. The bottleneck is now trusted feedback: you can only ship as fast as you can get repeatable test results.
 
-<p align="center"><img src="docs/static/img/readme/cloud-trace.gif" width="60%"></a>
+Dagger is designed for repeatability: tests run in containers; orchestration logic runs in sandboxed functions; host dependencies are explicit and strictly typed; intermediate artifacts are built just-in-time; everything is cached by default with fine-grained control. Same inputs, same outputs.
 
-- **Open Platform:** Works with any compute platform and tech stack — today and tomorrow. Ship faster, experiment freely, and don’t get locked into someone else's choices.
+### Local-first
 
-- **LLM Augmentation:** Native integration of any LLM that automatically discovers and uses available functions in your workflow. Ship mind-blowing agents in just a few dozen lines of code.
+Local execution is a core feature. Once configured, Dagger runs your tests reliably on any supported system. The only dependency is a recent Linux kernel. On non-Linux systems, Docker Desktop and similar products are supported out of the box.
```
rrati: Nit, but I would take a step up the chain and say a container engine is needed instead of a linux kernel.  While it's true non-linux container systems tend to use linux under the hood somehow (ie docker desktop using a vm on mac), that feels like a detail that isn't really relevant to dagger.  Unless I'm mistaken, dagger really just needs a container runtime to run the engine.

shykes: @rrati it's true that the dagger CLI only requires a container runtime, to start the engine. But the engine itself does *not* require a container runtime: it bundles its own. It's hard to explain the first part without causing confusion about the second part... So we only have 2 bad options:

- Option 1: "linux kernel". Less accurate on CLI dependency, more accurate on engine dependency
- Option 2: "container runtime". More accurate on CLI dep, less accurate on engine dep

🤷

rrati: I generally read project README's as explaining things at a high/top level.  So, if I'm new to the project, what do I need to know to understand how it works, it's value, and to get started using it.  The details of how it works under the hood I learn as I start using the tool, see its utility, and then once I'm invested I look under the hook because either I need to know how it works under the hood or am just curious.  From that perspective, option 2 seems more favorable to me since it is describing what a new person really needs to get started (I feel this way only because the engine is itself a container and bundles that linux kernel dep with it).  That makes it more welcoming to non-linux users imo.  Certainly somewhere in the "how it works docs" should be the details about the linux kernel dep.

shykes:
Made changes. Taking one last round of reviews, then will merge. Thanks all!

Issue #8638, opened 2024-10-07
🐞 <title>Fetching remote repository with SSH non-default port


pizza-prosciutto:
### What is the issue?

Fetching modules from a private repository using SSH fails when the repository uses a port other than 22.
Our privately hosted Bitbucket exposes port 7999 instead of the default 22.
Seems like the port in the URL is not taken into account.

### Dagger version

dagger v0.13.1 (registry.dagger.io/engine:v0.13.1) linux/amd64

### Steps to reproduce

Create a new module with `dagger init --sdk=go modules/hello`
Save it in a private repo which is accessible via SSH on a port other 7999
Try executing `dagger -m ssh://git@<private-bitbucket-server>:7999/<project>/dagger-demo.git/modules/hello functions`

### Log output

➜  /tmp dagger -m ssh://git@<private-bitbucket-server>:7999/<project>/dagger-demo.git/modules/hello functions
✔ connect 0.2s
✘ initialize 1m24.4s
! failed to get configured module: failed to get module ref kind: input: moduleSource resolve: failed to resolve git src to commit: failed to load cache key: error fetching default branch for repository ssh://xxxxx@<private-bitbucket-server>/<project>/dagger-demo.git: git error: exit status 128
  ✘ resolving module ref 1m24.4s
  ! failed to get configured module: failed to get module ref kind: input: moduleSource resolve: failed to resolve git src to commit: failed to load cache key: error fetching default branch for repository ssh://xxxxx@<private-bitbucket-server>/<project>/dagger-demo.git: git error: exit status 128
    ✘ moduleSource(refString: "ssh://git@<private-bitbucket-server>:7999/<project>/dagger-demo.git/modules/hello"): ModuleSource! 1m24.3s
    ! failed to resolve git src to commit: failed to load cache key: error fetching default branch for repository ssh://xxxxx@<private-bitbucket-server>/<project>/dagger-demo.git: git error: exit status 128
      ✘ cache request: git://<private-bitbucket-server>/<project>/dagger-demo.git 42.1s
      ! error fetching default branch for repository ssh://xxxxx@<private-bitbucket-server>/<project>/dagger-demo.git: git error: exit status 128

A new release of dagger is available: v0.13.1 → v0.13.3
To upgrade, see https://docs.dagger.io/install
https://github.com/dagger/dagger/releases/tag/v0.13.3
[...]

PR #11628, opened 2026-01-06
Allow dagger modules on non-default SSH port

wvell:
Fixes: #8638

I've added a test that passes the `RepoRootForImportDynamic` for  `vcs.RepoRootForImportPath`. I don't know if this is acceptable in this way but this was the only way for me to make this testable.

### DIFF
--- a/core/modulerefs.go
+++ b/core/modulerefs.go
@@ -214,8 +214,16 @@ func ParseGitRefString(ctx context.Context, refString string) (_ ParsedGitRefStr
 		cloneUser += "@"
 	}
 
-	gitParsed.SourceCloneRef = gitParsed.scheme.Prefix() + sourceUser + gitParsed.RepoRoot.Root
-	gitParsed.cloneRef = gitParsed.scheme.Prefix() + cloneUser + gitParsed.RepoRoot.Root
+	// For SSH URLs, inject port after host if it is defined: ssh://user@host:port/path
+	repoRootWithPort := gitParsed.RepoRoot.Root
+	if gitParsed.scheme == SchemeSSH && endpoint.Port > 0 {
+		if host, rest, ok := strings.Cut(repoRootWithPort, "/"); ok {
+			repoRootWithPort = fmt.Sprintf("%s:%d/%s", host, endpoint.Port, rest)
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


core/modulerefs_test.go
```diff
@@ -359,6 +359,20 @@ func TestParseRefString(t *testing.T) {
 				},
 			},
 		},
+		// Gerrit codereview on custom port
+		{
+			urlStr: "ssh://someuser@golang.org:29418/x/review/git-codereview",
+			want: &ParsedRefString{
```
marcosnils, line 364: you're not asserting here that the `SourceCloneRef` and the `cloneRef` effectively contains the port that the `urlStr` has 🤔

wvell: You are right, I tried not to change to much. I've added an optional check.

marcosnils:
SGTM. adding @grouville as a  second reviewer just to make sure I'm not missing anything :pray:

core/modulerefs.go
```diff
@@ -214,8 +214,16 @@ func ParseGitRefString(ctx context.Context, refString string) (_ ParsedGitRefStr
 		cloneUser += "@"
 	}
 
-	gitParsed.SourceCloneRef = gitParsed.scheme.Prefix() + sourceUser + gitParsed.RepoRoot.Root
-	gitParsed.cloneRef = gitParsed.scheme.Prefix() + cloneUser + gitParsed.RepoRoot.Root
+	// For SSH URLs, inject port after host if it is defined: ssh://user@host:port/path
+	repoRootWithPort := gitParsed.RepoRoot.Root
+	if endpoint.Port > 0 && gitParsed.scheme == SchemeSSH {
```
grouville: nit: I personally find `strings.Cut` a bit more readable instead of relying on indexes: 

```go
if gitParsed.scheme == SchemeSSH && endpoint.Port > 0 {
		if host, rest, ok := strings.Cut(repoRootWithPort, "/"); ok {
			repoRootWithPort = fmt.Sprintf("%s:%d/%s", host, endpoint.Port, rest)
		}
	}
```

cc @wvell

wvell: Yeah true, more readable. I've made the change.

PR #11631, opened 2026-01-06
[chore] remove StatLLB

alexcb:


core/directory.go
```diff
@@ -705,29 +668,36 @@ func (dir *Directory) File(ctx context.Context, file string) (*File, error) {
 	}, nil
 }
 
-func (dir *Directory) FileLLB(ctx context.Context, file string) (*File, error) {
-	query, err := CurrentQuery(ctx)
+func (dir *Directory) FileLLB(ctx context.Context, parent dagql.ObjectResult[*Directory], file string) (*File, error) {
```
alexcb, line 671: is there a way we can avoid having to pass in `parent dagql.ObjectResult[*Directory]`?

It's needed to do the `srv.Select(..., parent, ....)` call below -- is there a way we can extract it from the context instead?

alexcb:
This PR stems from the discussion under https://github.com/dagger/dagger/pull/11557#issuecomment-3662992983

core/directory.go
sipsma: For now I'd prefer to keep it explicit, even if the verbosity is annoying. The tradeoff is sneaking parent objects around might get confusing. In the future when this is all more stabilized and there's less to think about from buildkit, it's probably worth re-evaluating.

core/modulesource.go
```diff
@@ -1254,45 +1254,51 @@ func NewCallerStatFS(bk *buildkit.Client) *CallerStatFS {
...
 	if err != nil {
 		if status.Code(err) == codes.NotFound {
 			return nil, os.ErrNotExist
 		}
 		return nil, err
 	}
-	return stat, nil
+	fileMode := fs.FileMode(bkStat.Mode)
+	return &Stat{
+		Name:        bkStat.Path,
```
sipsma: nit: not a blocker but I do see the potential seeds of confusion being sown here in that in some existing parts of the code `Name` is just the name of the final element of the path, but here it's the full path. I *might* prefer just leaving this as only being the final name or there being a separate `Path` field? It's not a huge deal either way though, just mentioning in case there's an obvious way to disambiguate this.

alexcb: I got nerd-sniped on this and split out this function to return the resolved directory separately, which is only needed by the `FindUpAll` call, which uses it to track the found directory.

Any thoughts on the second commit in this PR or is it good to merge?

sipsma: LGTM!

sipsma (approved):

PR #11640, opened 2026-01-07
fsutil/copy: fix non-linux builds

tiborvass:
Commit 52cd3ce3231 from PR #11497 regressed cross compilation.
This commit fixes it.

### DIFF
--- a/internal/fsutil/copy/copy_unix.go
+++ b/internal/fsutil/copy/copy_unix.go
@@ -16,7 +16,7 @@ func getUIDGID(fi os.FileInfo) (uid, gid int) {
 	return int(st.Uid), int(st.Gid)
 }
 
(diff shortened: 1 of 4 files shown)


internal/fsutil/copy/copy_unix.go
```diff
@@ -16,7 +16,7 @@ func getUIDGID(fi os.FileInfo) (uid, gid int) {
 	return int(st.Uid), int(st.Gid)
 }
 
-func (c *copier) copyFileInfo(fi os.FileInfo, src, name string) error {
+func (c *copier) copyFileInfo(fi os.FileInfo, _, name string) error {
```
sipsma, line 19: i'd rather just remove this arg on every platform if it's not needed anywhere

PR #11642, opened 2026-01-07
Extending Modules

kpenfound:
This is an alternative to https://github.com/dagger/dagger/pull/11561

This introduces the concept of extending a module. If you extend a module, you will inherit all of its functions, but you can override any of them. Here's an example:

We have a HelloSimple toolchain here:

```go
package main

type HelloSimple struct{}

func (m *HelloSimple) Message() string {
	return "hello from simple"
}

func (m *HelloSimple) Goodbye() string {
	return "goodbye from simple"
}
```

This toolchain works great out of the box, however we need `Message()` to work differently. So we will create a new toolchain that extends HelloSimple

```go
package main

type HelloSimpleExtends struct{}

// Override Message from base
func (m *HelloSimpleExtends) Message() string {
	return "extended: hello from simple"
}
```

Now when we use the HelloSimpleExtends:

```
> dagger call hello-simple-extends message
extended: hello from simple

> dagger call hello-simple-extends goodbye
goodbye from simple
```

The example is a bit dumb but if you expand that to think about wrapping a toolchain to use specialized base images, composition between toolchains, adding services to base images, in the same way that Overlays was trying to solve, this is a cleaner implementation.

### DIFF
--- a/docs/docs-graphql/schema.graphqls
+++ b/docs/docs-graphql/schema.graphqls
@@ -3991,6 +3991,9 @@ type ModuleSource {
   """The engine version of the module."""
   engineVersion: String!
 
+  """The name of the module that this module extends, if any."""
+  extends: String!
+
   """
   The generated files and directories made on top of the module source's context directory.
   """
@@ -4125,15 +4128,6 @@ type ModuleSource {
   withName(
     """The name to set."""
     name: String!
-  ): ModuleSource!
-
-  """Set a toolchain as an overlay for another toolchain."""
-  withOverlayFor(
-    """The name of the toolchain to make an overlay."""
-    name: String!
-
-    """The name of the base toolchain to overlay."""
-    overlayFor: String!
   ): ModuleSource!
 
   """Update the module source with a new SDK."""
--- a/docs/static/api/reference/index.html
+++ b/docs/static/api/reference/index.html
@@ -12541,6 +12541,10 @@
                         <td> The engine version of the module. </td>
                       </tr>
                       <tr>
+                        <td data-property-name=""><a class="property-name" id="ModuleSource-extends" href="#ModuleSource-extends"><code>extends</code></a> - <span class="property-type"><a href="#definition-String"><code>String!</code></a></span> </td>
+                        <td> The name of the module that this module extends, if any. </td>
+                      </tr>
+                      <tr>
                         <td data-property-name=""><a class="property-name" id="ModuleSource-generatedContextDirectory" href="#ModuleSource-generatedContextDirectory"><code>generatedContextDirectory</code></a> - <span class="property-type"><a href="#definition-Directory"><code>Directory!</code></a></span> </td>
                         <td> The generated files and directories made on top of the module source's context directory. </td>
                       </tr>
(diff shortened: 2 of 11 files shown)


shykes:
Nice, yeah this seems better 👍

Questions:

- How does type merging work? For example if both modules define a type `Foo`, what happens?
- Dependencies work the regular way? For example if the extending module wants to call the "extendee" module, it needs to import it as a dependency explicitly correct?
- In this design, can we add convenience for handling arguments? One major pain of extending or wrapping a module today, is that all arguments have to be individually mapped - there's no way to say "pass through arguments except for this one". Could this feature have a facility for that?

kpenfound:
> How does type merging work? For example if both modules define a type Foo, what happens?

If module A has A.Foo, B extends A, B defines B.Foo, when we use B then B.Foo will take precedence, which is kind of the point of extending. You can add as many other functions as you want in B as well, but mainly you'll probably want to override functions from A.

> Dependencies work the regular way?

Yup! B probably wants to depend on A if its going to pass through to A.Foo in any way. Also if you're composing multiple toolchains, depend on those too

> In this design, can we add convenience for handling arguments?

For example, if we have A.Foo(Container, string, Directory), we could have B.Foo(string) overriding A.Foo that knows how to call A.Foo exactly how we want. Is that what you mean?

shykes:
> > How does type merging work? For example if both modules define a type Foo, what happens?
> 
> If module A has A.Foo, B extends A, B defines B.Foo, when we use B then B.Foo will take precedence, which is kind of the point of extending. You can add as many other functions as you want in B as well, but mainly you'll probably want to override functions from A.

But what about *other* types defined in A and B?

- Can a function in `B` return a type defined in `A` for example?
- If `B` defines a new type, will it be exposed?

> > In this design, can we add convenience for handling arguments?
> 
> For example, if we have A.Foo(Container, string, Directory), we could have B.Foo(string) overriding A.Foo that knows how to call A.Foo exactly how we want. Is that what you mean?

Yes, something like that 👍

vito:
This is definitely better than overlays, but I still worry a bit because it feels like a halfway step towards subclassing - which itself has somewhat fallen out of style, but our halfway step would be missing a key part that makes it work.

Take this scenario in Ruby:

```ruby
class Animal
  def speak
    puts make_sound
  end
  
  def make_sound
    "generic sound"
  end
end

class Dog < Animal
  def make_sound
    "woof!"
  end
end

dog = Dog.new
dog.speak  # Output: "woof!"
```

The key thing is when the original code calls a sibling method, it calls the _overridden_ method, which won't happen in our equivalent afaict - it'll just call the original (non-overridden) sibling. So if you're using a toolchain that has `Build` which calls `Generate`, and you override `Generate`, the original module will still call its own `Generate` and ignore your overridden one. _Maybe_ if we had self-calls we could do some fancy rerouting to the override, but we don't have self-calls yet.

I dunno if that's a dealbreaker (maybe I'm drawing parallels that aren't realistic) but wanted to raise it for the sake of the discussion.

kpenfound:
@vito yeah thats totally true. Do you think this would be good if we went all the way on subclassing, or does this framing give you any ideas for a better approach?

vito:
No immediate alternative ideas, but if we do land on subclassing as the solution, I can at least see pretty significant upside.

For example, now you can 'fork' a dependency without having to fork its code and know its language. You just publish a module that extends it under the hood and applies your tweak. And if your extension can be written in Dang, the overhead will be minimal. That's pretty neat, and most of the time I'd prefer to do that than fork the repo and roll the dice on the maintainer's responsiveness.

The downside I suppose is that it could inhibit improvements to the upstream module, since everyone just [flex tapes](https://github.com/user-attachments/assets/dfb75383-8701-4b97-8c67-cc883fa8eb50) over their problems. But we could mitigate that by tracking extensions of modules, so module authors can see what changes people have had to make, and integrate them gradually if they want. Most of the time we're not talking about substantially challenging things to implement, just more knobs.

Side question: can you _add_ functions? Or does it only shadow existing functions? And I suppose you can't remove them - not sure how you'd even express that, besides something in `dagger.json`.

kpenfound:
Nice @vito that all sounds good. I think a proper subclassing implementation would look quite a bit different than what I have since this is really just hijacking routing of functions in front of another module. Tracking which modules are extending yours sounds like it would fit nicely in the modules feature in cloud.

For your question, yeah you currently can add whatever you want (99% sure at least, memory is fuzzy now)

PR #11650, opened 2026-01-08
Expose Checks on current environment

eunomie:
Expose checks on current environment.
Add a `current-env` function for dagger shell

Checks can be retrieved from the current environment. So from a module function:

```go
dag.CurrentEnv().Checks()
```

Or using `dag.CurrentEnv().Check(checkName)` to find one.

Then checks can be listed or run.

cc @shykes

### DIFF
--- a/dagger.json
+++ b/dagger.json
@@ -51,7 +51,8 @@
             "docs/**",
             "dagql/idtui/viztest/broken/**",
             "modules/evals/**",
-            "**/broken*/**"
+            "**/broken*/**",
+            "core/integration/testdata/checks/hello-with-checks/"
           ]
         }
       ]


eunomie:
For now this only works for the main module: from the main module we can see all checks, defined by the module or by the toolchains.

eunomie:
The `go:lint` error is because the test module is using a new function defined by the PR. I don't know if there's a simple way to fix it, but it should auto-resolve once merged.

shykes:
> The `go:lint` error is because the test module is using a new function defined by the PR. I don't know if there's a simple way to fix it, but it should auto-resolve once merged.

Probably fixable either by 1) changing the `.golangci.yml` config, or 2) customizing the go toolchain in `dagger.json` [to ignore that directory](https://github.com/dagger/dagger/blob/main/dagger.json#L38)

PR #11654, opened 2026-01-09
feat: fix conflict on tsconfig.json when generating multiple ts client.

TomChv:
*Before*, when generating a client, we were also configuring the `tsconfig.json` with a path alias `@dagger.io/client` for that client so it's easy to import for the user.
But if you try to generate 2 clients, then a conflict will appear.
This PR fixes that issue by no longer configuring a path alias when generating a client. 
Instead the user can either:
- Import the client using relative path: `./<output-dir>/client.gen`
- Configure his own path alias: `@my-app/dagger` (or whatever alias he want)

We still update the tsconfig.json to correctly root the sdk library but that's all. 

**Changes**
- Optimize `GenerateClient` to not use any container when generating files.
- Update tsconfig updater to not create aliases for generated client, but instead respect existing aliases.
- Update packageJSON updater to not remove @dagger.io/dagger as dependency
- No longer use `local` lib location when generating client.
- Simplify Lib generator functions to take more context inside the constructor instead of arguments
- Update integrations tests and add a test for multiple generated client.

**Note**

This is not a breaking change since existing generated client already have `@dagger.io/client` configured in the module's tsconfig.json

### DIFF
--- a/sdk/typescript/runtime/lib_generator.go
+++ b/sdk/typescript/runtime/lib_generator.go
@@ -122,8 +122,6 @@ func (l *LibGenerator) GenerateBundleLibrary(
 
 	// If we're generating a standalone client, we export a lighter
 	// version of the library.
-	// Otherwise, we export the full library because if a client and a
-	// module coexist, they both depend on the same `sdk` directory.
 	if l.Opts.genClient && !l.Opts.coexistWithModule {
 		return result.
 			WithNewFile("index.ts", tsutils.StaticBundleClientIndexTS)
--- a/sdk/typescript/runtime/runtime_node.go
+++ b/sdk/typescript/runtime/runtime_node.go
@@ -62,11 +62,7 @@ func (n *NodeRuntime) SetupContainer(ctx context.Context) (*dagger.Container, er
 		ctx, span := Tracer().Start(gctx, "generate SDK library")
 		defer span.End()
 
-		sdkLibrary, err = NewLibGenerator(n.sdkSourceDir,
-			&LibGeneratorOpts{
-				moduleName: n.cfg.name,
-				modulePath: n.cfg.modulePath(),
-			}).
+		sdkLibrary, err = NewLibGenerator(n.sdkSourceDir, n.cfg.libGeneratorOpts()).
 			GenerateBundleLibrary(n.introspectionJSON, ModSourceDirPath).
 			Sync(ctx)
 		return err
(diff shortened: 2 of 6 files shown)


TomChv:
I'm waiting for the CI to be green so I can delete all the dead code left by this change

EDIT: done

sdk/typescript/runtime/lib_generator.go
```diff
@@ -39,74 +50,115 @@ func NewLibGenerator(sdkSourceDir *dagger.Directory) *LibGenerator {
...
-				modulePath,
 				Bundle,
+				outputDir,
 			),
 		)
+
+	// If we're generating a standalone client, we export a lighter
+	// version of the library.
+	// Otherwise, we export the full library because if a client and a
+	// module coexist, they both depend on the same `sdk` directory.
```
eunomie:
```suggestion
```

sdk/typescript/runtime/runtime_node.go
```diff
@@ -138,8 +142,12 @@ func (n *NodeRuntime) GenerateDir(ctx context.Context) (*dagger.Directory, error
...
 		defer span.End()
 
-		sdkLibrary, err = NewLibGenerator(n.sdkSourceDir).
-			GenerateBundleLibrary(n.introspectionJSON, n.cfg.name, n.cfg.modulePath()).
+		sdkLibrary, err = NewLibGenerator(n.sdkSourceDir,
+			&LibGeneratorOpts{
+				moduleName: n.cfg.name,
+				modulePath: n.cfg.modulePath(),
+			}).
+			GenerateBundleLibrary(n.introspectionJSON, ModSourceDirPath).
```
eunomie: **nit (non blocking)**: I think this exact piece of code is duplicated 4 or 5 times, I would have extracted that in a helper function

eunomie (approved):
Sounds good to my knowledge of the ts SDK :-)

Just added two quick nits, one we discussed live regarding a comment that I feel confusing, and a nit but that might be personal preference, so feel free to not follow it :-D

sdk/typescript/runtime/runtime_node.go
TomChv: Done in [0a5a3bf](https://github.com/dagger/dagger/pull/11654/commits/0a5a3bfed47db1c7bbe8c68ecd8be92e3e1e5162)

sdk/typescript/runtime/lib_generator.go
TomChv: Done in [0a5a3bf](https://github.com/dagger/dagger/pull/11654/commits/0a5a3bfed47db1c7bbe8c68ecd8be92e3e1e5162)

PR #11657, opened 2026-01-09
Remove deprecated "arguments" field from our dagger.json

shykes:


shykes:
> Is the `sdk-bundler` supposed to be included here? Otherwise LGTM

Nope, I don't even know what that is 😆

PR #11659, opened 2026-01-10
fix(cli): limit `develop --recursive` parallelism to the engine's NumCPU

grouville:
### Problem

As shown in this [discord thread](https://discord.com/channels/707636530424053791/1458961181900537988) and tracked by @sipsma, `dagger develop --recursive` was spawning unbounded goroutines for all module dependencies simultaneously.

For repos with many modules (20+), this caused memory to spike faster than work could complete, leading to OOM kills on smaller laptops.

##### Add a semaphore to limit concurrent module development to `runtime.NumCPU()`:

```go
sem := semaphore.NewWeighted(int64(runtime.NumCPU()))
eg.Go(func() (err error) {
    if err := sem.Acquire(ctx, 1); err != nil {
        return err
    }
    defer sem.Release(1)
    // ... work
})
```
This bounds peak memory while maintaining throughput: CPU is typically the bottleneck anyway.

##### Engine CPU count via provisioner

Per @sipsma's feedback, using client-side `runtime.NumCPU()` would be incorrect when client and engine run on different hosts (e.g., laptop with 4 CPUs connecting to a 32-CPU cloud engine).

We extended the buildkit Info endpoint to include SystemInfo:
```go
message InfoResponse {
    BuildkitVersion buildkitVersion = 1;
    SystemInfo systemInfo = 2;
}

message SystemInfo {
    int32 numCPU = 1;
}
```

This is fetched during the connection handshake alongside version info: no extra round-trip. The SystemInfo structure leaves room for future engine capabilities (memory, disk, etc.).

For nested sessions (modules running inside the engine), we pass `DAGGER_ENGINE_NUM_CPU` via environment variable, consistent with `DAGGER_SESSION_PORT` and `DAGGER_SESSION_TOKEN`.

### Alternatives Considered

I initially tried `conc/pool` with `WithMaxGoroutines()`, but it required creating trace spans inside the goroutine. This caused spans to [appear staggered in Dagger Cloud](https://dagger.cloud/dagger/traces/6bb4597cf54062cfb4bad577398eca8c?listen=248c880cf747b643) instead of [grouped together](https://dagger.cloud/dagger/traces/6a217736dabb354120a20fac7e5fce48?listen=ee527801cbf5cdbe), degrading the UI experience.

  The semaphore approach preserves the original trace grouping while still limiting parallelism.

### Testing

```shell
./hack/with-dev go test -v -count=1 ./core/integration -run "TestCLI/TestDaggerDevelop/recursive"
# PASS (10.39s)
```

cc @chrisjpalmer

### DIFF
--- a/engine/buildkit/executor_spec.go
+++ b/engine/buildkit/executor_spec.go
@@ -66,9 +66,9 @@ const (
 	DaggerHostnameAliasesEnv = "_DAGGER_HOSTNAME_ALIASES"
 	DaggerNoInitEnv          = "_DAGGER_NOINIT"
 
-	DaggerSessionPortEnv   = "DAGGER_SESSION_PORT"
-	DaggerSessionTokenEnv  = "DAGGER_SESSION_TOKEN"
-	DaggerEngineNumCPUEnv  = "DAGGER_ENGINE_NUM_CPU"
+	DaggerSessionPortEnv  = "DAGGER_SESSION_PORT"
+	DaggerSessionTokenEnv = "DAGGER_SESSION_TOKEN"
+	DaggerEngineNumCPUEnv = "DAGGER_ENGINE_NUM_CPU"
 
 	// this is set by buildkit, we cannot change
 	BuildkitSessionIDHeader = "x-docker-expose-session-uuid"


cmd/dagger/module.go
```diff
@@ -658,6 +660,7 @@ This command is idempotent: you can run it at any time, any number of times. It
 			defer telemetry.EndWithCause(span, &err)
 
 			eg, ctx := errgroup.WithContext(ctx)
+			sem := semaphore.NewWeighted(int64(runtime.NumCPU()))
```
sipsma: This won't technically be the correct number when the client+engine are on different hosts. If the client has way less CPUs than the remote engine this would be unnecessarily limiting I think.

There's definitely *some* way to enable the client to know the number of CPUs on the engine host; i.e. if absolutely nothing else we can add an endpoint like `/cpucount` that returns it. I'm sure there's something a little more elegant (maybe integrated with the provisioner code?), but I'd honestly be fine with something like that.

This isn't necessarily a blocker, just very nice to have now but can easily be improved later.

grouville: You're right ahah, didn't think about that use-case, let me see 🙏

chrisjpalmer: Thanks guys the approach seems great. I've been thinking we needed something like this for a while and glad to have something coming in to limit the parallelism :)

grouville: Ok, so I went with _integrating it into the provisioner_, as you suggested. 

The pattern: extend the buildkit `Info` endpoint to include a `SystemInfo` struct:

```go
message InfoResponse {
    BuildkitVersion buildkitVersion = 1;
    SystemInfo systemInfo = 2;
}
  
message SystemInfo {
    int32 numCPU = 1;
}
```

This is fetched during the connection handshake (alongside version info), so no extra round-trip. The `SystemInfo` structure leaves room for future engine capabilities (memory, disk, etc.) without protocol changes

PR #11678, opened 2026-01-13
top level ignoreChecks

kpenfound:
We have `ignoreChecks` available within the toolchains configuration in dagger.json, however @TomChv and I discovered the need for a top level one: when you're developing a toolchain that provides checks, but those checks dont make sense to run in the toolchain's own CI. I think thats basically the only use case for this but it is important


kpenfound:
Closing this as we are discussing alternatives to distinguish toolchains from non-toolchains and only running checks in toolchains

PR #11682, opened 2026-01-14
Python library and toolchain to send telemetry when using pytest

eunomie:
Creates a `pytest_otel` python library that exposes tests run to otel. This library is using `pytest` as the test runner.
The library is not (yet?) published, it's embedded in the container running pytest.

A `python` toolchain is created, containing a `test` function that automatically runs tests using `pytest` and this `pytest_otel` library. As the library is injected in the container, no configuration is required to the developer. The toolchain can be directly used on python project.

```console
$ dagger install github.com/dagger/dagger/toolchains/python

$ dagger check
```

Should be enough to run python tests using the otel library, allowing to see traces for individual tests.

The toolchain module can of course also be used as a dependency, for deeper integration. You can see for instance the way it's been integrated in the python SDK module.

### DIFF
--- a/toolchains/python-sdk-dev/dagger.json
+++ b/toolchains/python-sdk-dev/dagger.json
@@ -23,6 +23,10 @@
     {
       "name": "wolfi",
       "source": "../../modules/wolfi"
+    },
+    {
+      "name": "python",
+      "source": "../python"
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


eunomie:
Replaced by https://github.com/dagger/pytest

PR #11684, opened 2026-01-14
feat: split telemetry from @dagger.io/dagger

TomChv:
### Summary

Move telemetry logic into `@dagger.io/telemetry` and import it in `@dagger.io/dagger`

The release process is decouple from `@dagger.io/dagger`, there's a new check in the ts SDK that verify that we correctly bump the package version when we update that package so later we can automatically release it.

When we build a dagger engine, we use the local telemetry package so we are always up to date and it makes tests always using the latest changes.
 
### List of changes

- Create `@dagger.io/telemetry` package.
- Add check in `sdk-typescript-dev` to verify we correctly bump the version.
- Use `@dagger.io/telemetry` in `@dagger.io/dagger`.
- Update builder to use the local telemetry package.

### DIFF
--- a/sdk/typescript/yarn.lock
+++ b/sdk/typescript/yarn.lock
@@ -230,14 +230,14 @@
   resolved "https://registry.yarnpkg.com/@eslint-community/regexpp/-/regexpp-4.12.2.tgz#bccdf615bcf7b6e8db830ec0b8d21c9a25de597b"
   integrity sha512-EriSTlt5OC9/7SXkRSCAhfSxxoSUgBm33OH+IkwbdpgoqsSsUg7y3uh+IICI/Qg4BBWr3U2i39RpmycbxMq4ew==
 
-"@eslint/config-array@^0.23.2":
-  version "0.23.2"
-  resolved "https://registry.yarnpkg.com/@eslint/config-array/-/config-array-0.23.2.tgz#db85beeff7facc685a5775caacb1c845669b9470"
-  integrity sha512-YF+fE6LV4v5MGWRGj7G404/OZzGNepVF8fxk7jqmqo3lrza7a0uUcDnROGRBG1WFC1omYUS/Wp1f42i0M+3Q3A==
+"@eslint/config-array@^0.23.3":
+  version "0.23.3"
+  resolved "https://registry.yarnpkg.com/@eslint/config-array/-/config-array-0.23.3.tgz#3f4a93dd546169c09130cbd10f2415b13a20a219"
+  integrity sha512-j+eEWmB6YYLwcNOdlwQ6L2OsptI/LO6lNBuLIqe5R7RetD658HLoF+Mn7LzYmAWWNNzdC6cqP+L6r8ujeYXWLw==
   dependencies:
-    "@eslint/object-schema" "^3.0.2"
+    "@eslint/object-schema" "^3.0.3"
     debug "^4.3.1"
-    minimatch "^10.2.1"
+    minimatch "^10.2.4"
 
 "@eslint/config-helpers@^0.5.2":
   version "0.5.2"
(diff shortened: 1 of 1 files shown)


toolchains/typescript-sdk-dev/verify-pkg-version-bump/main.go
```diff
@@ -0,0 +1,52 @@
...
+			File("package.json").
+			AsJSON().
+			Field([]string{"version"}).
+			AsString(ctx)
+		if err != nil {
+			fmt.Printf("failed to load package.json: %s\n", err.Error())
+			os.Exit(1)
+		}
+
+		if semver.Compare(repoPackageVersion, currentPackageVersion) > 0 {
```
grouville, line 50: I think you also want it to fail when there's been modifications to the sdk's code and you didn't bump the package version. Shouldn't it be `>= 0` ?

toolchains/typescript-sdk-dev/ts-sdk.dang
```diff
@@ -74,6 +74,32 @@ type TypescriptSdkDev {
...
+		  let repoTelemetryDir = repoDir.directory("sdk/typescript/telemetry")
+
+			let binary=go(source:currentModule.source.directory("verify-pkg-version-bump")).binary("verify-pkg-version-bump")
+
+      let exitCode=container.
+        from("alpine").
+        withFile("/bin/verify-pkg-version-bump", binary).
+        withMountedDirectory("/work/repo", repoTelemetryDir).
+        withMountedDirectory("/work/current", currentTelemetryDir).
+        withExec(["verify-pkg-version-bump", version], expect: ReturnType.ANY, experimentalPrivilegedNesting: true).
```
grouville: I'm surprised that it's being executed, as `version` doesn't seem to be declared anywhere here in `["verify-pkg-version-bump", version]`

toolchains/typescript-sdk-dev/verify-pkg-version-bump/main.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func main() {
+	ctx := context.Background()
+
+	repoDir := dag.Host().Directory("/work/repo")
+	currentDir := dag.Host().Directory("/work/current")
+
+	changes := repoDir.Changes(currentDir)
+	isDifferent, err := changes.IsEmpty(ctx)
```
grouville: i think it's inverted: `IsEmpty` returns true if it's empty.

So, below, in `if isDifferent {`, you would always enter and compare versions when there's no difference

sdk/typescript/telemetry/src/sdk/index.ts
```diff
@@ -0,0 +1,113 @@
...
+      this._otelSDK.start()
+    }
+  }
+
+  /**
+   * Start the otel SDK.
+   *
+   * @deprecated please use `start`
+   */
+  public initiliaze() {
```
grouville: nit: it's been deprecated but there's a typo

sdk/typescript/telemetry/src/sdk/index.ts
```diff
@@ -0,0 +1,113 @@
...
+  BatchSpanProcessor,
+  type SpanProcessor,
+} from "@opentelemetry/sdk-trace-base"
+
+import { DaggerContextManager } from "./context_manager"
+import { LiveSpanProcessor, NEARLY_IMMEDIATE } from "./span_processor"
+import { isOtelEnabled } from "./util"
+
+/**
+ * OtelSDK is a wrapper around the NodeSDK that simplifies the instanciation
```
grouville: `instantiation`

sdk/typescript/telemetry/src/tracer/index.ts
```diff
@@ -0,0 +1,17 @@
+import { Tracer } from "./tracer"
+
+const DAGGER_TRACER_NAME = "dagger.io/sdk.typescript"
+
+/**
+ * Return an OpenTelemetry tracer to use with Dagger.
+ *
+ * As a conveniance function, you can use `startActiveSpan` that automatically close
```
grouville: `convenience`

sdk/typescript/src/module/entrypoint/entrypoint.ts
```diff
@@ -54,7 +54,7 @@ export async function entrypoint(files: string[]) {
           // Closing telemetry here since the process is shutdown after
           // so any parent's finally will not be executed.
           // That way, we guarantee that the spans are flushed.
-          await telemetry.close()
+          await telemetry.shutdown()
```
grouville, line 57: I like the rename from `close()` to `shutdown()` as it aligns with the official [otel's sdk terminology](https://www.npmjs.com/package/@opentelemetry/sdk-node)

toolchains/typescript-sdk-dev/verify-pkg-version-bump/main.go
grouville, line 50: I don't think it works either for another reason: 

```go
package main

import (
    "fmt"
    "golang.org/x/mod/semver"
)

func main() {
    pairs := [][2]string{
        {"0.0.6", "0.0.5"},
        {"v0.0.6", "v0.0.5"},
        {"0.0.5", "v0.0.5"},
        {"v0.0.5", "0.0.5"},
    }
    for _, p := range pairs {
        fmt.Printf("Compare(%q, %q) = %d\n", p[0], p[1], semver.Compare(p[0], p[1]))
    }
}
```

if you run that, you get: 

```shell
go run .                      
Compare("0.0.6", "0.0.5") = 0
Compare("v0.0.6", "v0.0.5") = 1
Compare("0.0.5", "v0.0.5") = -1
Compare("v0.0.5", "0.0.5") = 1
```

this `semver.Compare()` library expects semver with a leading`v`. As the `package.json`'s _**version**_ don't have one, then it's always going to return `0`, as showed above, leading to an incorrect detection

grouville (commented):
Amazing, thanks 🙏 The telemetry extraction looks super clean, and the API seems to align well with OpenTelemetry conventions

_> Happy to see `shutdown()` matching the OTel spec and the Python SDK 🙏 (only Go will differ now, in terms of terminology)_

Left a few typos / nits, but the main thing, IMO, is the version bump checke:r in `ts-sdk.dang` and `verify-pkg-version-bump/main.go`. I don't think it actually enforces a proper bump check in this form:
- the Dang code seems to reference an _undefined_ `version` variable
- the `IsEmpty()` check is inverted (runs when there are no changes, skips when there are)
- `semver.Compare` returns 0 for npm-style versions without the `v` prefix.

> Net effect: I might be wrong, but the telemetry changes will apparently never trigger a version bump failure?

For the _telemetry_ package itself, I wonder if adding a small non-exhaustive test that just catches obvious breakages when dependencies update occur would be hard ? If it's too complex, should be ok skipping it (it's the same for our overall CI atm ahah)

TomChv:
> For the telemetry package itself, I wonder if adding a small non-exhaustive test that just catches obvious breakages when dependencies update occur would be hard ? If it's too complex, should be ok skipping it (it's the same for our overall CI atm ahah)

I think I can figure out something! Good call

grouville:
Taking a 👀 on the bun crash, seems to occur prior my first commit

PR #11685, opened 2026-01-14
core: ModTree infrastructure for dynamic checks

shykes:
*This PR depends on #11726, please merge that first*

## Summary
- Replace walkObjectChecks with ModTree abstraction for traversing module trees
- Rename FnTreeNode to ModTreeNode with simplified structure
- Add support for walking arrays of objects with +check methods
- Add --all/-a flag to `dagger check` to enumerate dynamic checks
- Include dynamic checks in aggregate listing even without --all
- Fix TUI display for aggregate check execution
- Add modtree path matching tests

## Context
This PR adds the core infrastructure needed to support dynamic checks from toolchains. The ModTree provides a unified way to traverse module object graphs, finding and executing checks defined via the `+check` decorator.

Dynamic checks (from functions returning `[]*Object` where Object has `+check` methods) are supported with lazy enumeration controlled by the `--all` flag.

## Task list

- [ ] [Re-enable scale-out](https://github.com/dagger/dagger/pull/11685#issuecomment-3801747862)

### DIFF
--- a/core/modtree.go
+++ b/core/modtree.go
@@ -370,17 +370,16 @@ func (p ModTreePath) Equals(ctx context.Context, other ModTreePath) (result bool
 }
 
 func (p ModTreePath) Glob(ctx context.Context, pattern string) (bool, error) {
-	slashPattern := strings.ReplaceAll(pattern, ":", "/")
-	for _, pathVariant := range [][]string{p.APICase(), p.CliCase()} {
-		slashPath := strings.Join(pathVariant, "/")
-		if match, err := doublestar.PathMatch(slashPattern, slashPath); err != nil {
-			return false, err
-		} else if match {
-			debugTrace(ctx, "%q.Glob(%q) -> MATCH", slashPath, slashPattern)
-			return true, nil
-		}
-		debugTrace(ctx, "%q.Glob(%q) -> no match", slashPath, slashPattern)
+	// Normalize both pattern and path to CLI case (kebab-case) for consistent matching
+	slashPattern := strings.Join(NewModTreePath(pattern).CliCase(), "/")
+	slashPath := strings.Join(p.CliCase(), "/")
+	if match, err := doublestar.PathMatch(slashPattern, slashPath); err != nil {
+		return false, err
+	} else if match {
+		debugTrace(ctx, "%q.Glob(%q) -> MATCH", slashPath, slashPattern)
+		return true, nil
 	}
+	debugTrace(ctx, "%q.Glob(%q) -> no match", slashPath, slashPattern)
 	return false, nil
 }
 


shykes:
@eunomie it seems that your PR #11650 overlaps with my implementation here in a massive way. I will try to reconcile so I can merge this, before it goes to the trash... Can you help me please?

shykes:
Update, I opened #11726 to bridge the gap, I understand why you had to do it this way @eunomie, you had to bridge the CheckGroup model (a single root module) with the Env model (any number of modules). After discussing with @vito we agreed that it's a good opportunity to converge Env with the toolchain model. Hence #11726, then this.

shykes:
FYI @eunomie, as discussed we're missing scale-out, otherwise this should be mergeable.

I asked Claude Code to summarize the plan for re-enabling scale-out, see below.

Note @sipsma, it recommends *not* keeping the exact trick you used, using the distinction between `CheckGroup.run` and `Check.run` to avoid recursion. Could you just check that you are OK with the proposed solution? I think it's basically what you had already explained to me: use the info already available in clientMetadata.

👇 

> # Plan: Bringing Back Scale-Out Capability to ModTree Architecture
> 
> ## Context
> 
> The `check-plumbing` branch introduces the **ModTree** architecture - a hierarchical tree-based representation of module structure that replaces the previous flat approach to check discovery and execution. This is a significant architecture cleanup that:
> 
> 1. Provides a unified tree abstraction (`ModTreeNode`) for traversing module structure
> 2. Enables flexible pattern matching (glob patterns, path containment)
> 3. Simplifies check execution by recursively walking the tree
> 4. Sets the foundation for future extensibility (dynamic objects, new leaf types)
> 
> However, the scale-out capability (running checks on remote cloud engines) was intentionally removed to validate the tree-based approach in isolation. This document outlines how to restore scale-out without losing the ModTree architecture benefits.
> 
> ---
> 
> ## Understanding the Dual Codepath Design on `main`
> 
> On the `main` branch, there are **two distinct codepaths** for running checks:
> 
> ### 1. `CheckGroup.Run()` - The "Originator" Path
> 
> ```go
> func (r *CheckGroup) Run(ctx context.Context) (*CheckGroup, error) {
>     clientMD, err := engine.ClientMetadataFromContext(ctx)
>     // ...
>     for _, check := range r.Checks {
>         // Always creates telemetry span
>         ctx, span := Tracer(ctx).Start(ctx, check.Name(), ...)
>         eg.Go(func() error {
>             return check.run(ctx, dag, clientMD.EnableCloudScaleOut)  // <-- CAN scale out
>         })
>     }
> }
> ```
> 
> - Called by CLI when running `dagger check`
> - Caches DAG servers across checks (optimization)
> - **Always creates telemetry spans**
> - Passes `clientMD.EnableCloudScaleOut` to `check.run()` → **CAN initiate scale-out**
> 
> ### 2. `Check.Run()` - The "Target" Path
> 
> ```go
> func (c *Check) Run(ctx context.Context) (*Check, error) {
>     // Telemetry span is CONDITIONAL
>     if clientMD.CloudScaleOutEngineID != "" {
>         ctx, span = Tracer(ctx).Start(ctx, c.Name(), ...)
>     }
>     checkErr = c.run(ctx, dag, false)  // <-- CANNOT scale out (hardcoded false)
> }
> ```
> 
> - Called when a remote engine executes a scaled-out check
> - Remote query: `.moduleSource(...).asModule().check(name: "...").run()`
> - **Conditional telemetry** (only if we're a scale-out target, to avoid duplication)
> - Passes `false` to `check.run()` → **CANNOT initiate scale-out** (prevents infinite recursion)
> 
> ### Why Two Paths?
> 
> The recursion prevention is **structural, not runtime**:
> - Originator (`CheckGroup.Run`) → can scale out
> - Target (`Check.Run`) → cannot scale out (hardcoded)
> 
> When a check scales out, the remote engine receives a `Check.Run()` call, which by design never tries to scale out again. The `CloudScaleOutEngineID` check is only for **telemetry deduplication**, not recursion prevention.
> 
> ---
> 
> ## Current State on `check-plumbing`
> 
> The branch has **collapsed both paths into one**:
> 
> ```go
> // CheckGroup.Run()
> func (r *CheckGroup) Run(ctx context.Context) (*CheckGroup, error) {
>     for _, check := range r.Checks {
>         jobs = jobs.WithJob(check.Name(), func(ctx context.Context) error {
>             return check.Node.RunCheck(ctx, nil, nil)  // <-- unified path
>         })
>     }
> }
> 
> // Check.Run()
> func (c *Check) Run(ctx context.Context) (*Check, error) {
>     err := c.Node.RunCheck(ctx, nil, nil)  // <-- same unified path
> }
> ```
> 
> Both go through `ModTreeNode.RunCheck()` → `runLeafCheck()`.
> 
> ---
> 
> ## Design Decision: Keep Single Path (Recommended)
> 
> **Recommendation: Keep the unified single-path design. Do NOT restore the dual codepath.**
> 
> ### Rationale
> 
> 1. **The dual codepath is no longer necessary.** The recursion prevention on `main` relies on hardcoding `tryScaleOut=false` in `Check.Run()`. But there's a simpler mechanism already in place: `session.go:1617` sets `EnableCloudScaleOut: false` when creating the scale-out client:
> 
>    ```go
>    // engine/server/session.go:1614-1617
>    clientMetadata: &engine.ClientMetadata{
>        // FIXME: for now, disable recursive scale out to prevent any
>        // surprise "fork-bomb" scenarios.
>        EnableCloudScaleOut: false,
>    },
>    ```
> 
>    This means scaled-out sessions **already have scale-out disabled at the session level**. We can simply check `clientMD.EnableCloudScaleOut` in `runLeafCheck()` and get recursion prevention "for free".
> 
> 2. **Smallest diff.** Keeping the single path means:
>    - No changes to method signatures
>    - No changes to `CheckGroup.Run()` or `Check.Run()` structure
>    - Just add scale-out logic inside `runLeafCheck()`
> 
> 3. **Simpler mental model.** One execution path is easier to understand and maintain.
> 
> ### The Tradeoff
> 
> The `main` branch dual codepath has one advantage: the telemetry span creation is conditional in `Check.Run()` to avoid duplication when we're a scale-out target.
> 
> With the single path approach, we need to handle this inside `RunCheck()`. Looking at the current code, `RunCheck()` already creates a telemetry span unconditionally. We'll need to add the same conditional logic:
> 
> ```go
> func (node *ModTreeNode) RunCheck(ctx context.Context, include, exclude []string) (rerr error) {
>     clientMD, _ := engine.ClientMetadataFromContext(ctx)
> 
>     // Only create span if we're NOT a scale-out target (to avoid telemetry duplication)
>     if clientMD == nil || clientMD.CloudScaleOutEngineID == "" {
>         ctx, span := Tracer(ctx).Start(ctx, node.PathString(), ...)
>         defer func() { ... }()
>     }
>     // ... rest of execution
> }
> ```
> 
> This is a small addition compared to maintaining two separate execution paths.
> 
> ---
> 
> ## Implementation Plan
> 
> ### Step 1: Add scale-out logic to `runLeafCheck()` (core change)
> 
> ```go
> func (node *ModTreeNode) runLeafCheck(ctx context.Context) error {
>     clientMD, err := engine.ClientMetadataFromContext(ctx)
>     if err != nil {
>         return err
>     }
> 
>     // Try scale-out if enabled (will be false for scaled-out sessions)
>     if clientMD.EnableCloudScaleOut {
>         if ok, err := node.tryScaleOut(ctx); ok {
>             return err
>         }
>     }
> 
>     // Local execution (existing code)
>     return node.runLeafCheckLocally(ctx)
> }
> ```
> 
> ### Step 2: Add `tryScaleOut()` method to `ModTreeNode`
> 
> Port the logic from `main:core/checks.go:330-400`:
> 
> ```go
> func (node *ModTreeNode) tryScaleOut(ctx context.Context) (bool, error) {
>     q, err := CurrentQuery(ctx)
>     if err != nil {
>         return true, err
>     }
> 
>     cloudClient, useCloud, err := q.CloudEngineClient(ctx,
>         node.RootAddress(),   // module ref
>         node.PathString(),    // check path
>         nil,                  // no exec cmd
>     )
>     if err != nil {
>         return true, fmt.Errorf("engine-to-engine connect: %w", err)
>     }
>     if !useCloud {
>         return false, nil  // run locally
>     }
>     defer cloudClient.Close()
> 
>     query := cloudClient.Dagger().QueryBuilder()
> 
>     // Build module source query based on kind (see "Query Building" section below)
>     // ...
> 
>     query = query.Select("asModule")
>     query = query.Select("check").Arg("name", node.PathString())
>     query = query.Select("run")
>     query = query.SelectMultiple("completed", "passed")
> 
>     var res struct{ Completed, Passed bool }
>     return true, query.Bind(&res).Execute(ctx)
> }
> ```
> 
> ### Step 3: Conditional telemetry in `RunCheck()`
> 
> Modify `RunCheck()` to skip span creation when we're a scale-out target:
> 
> ```go
> func (node *ModTreeNode) RunCheck(ctx context.Context, include, exclude []string) (rerr error) {
>     clientMD, _ := engine.ClientMetadataFromContext(ctx)
> 
>     // Only create telemetry span if we're NOT a scale-out target
>     var span trace.Span
>     if clientMD == nil || clientMD.CloudScaleOutEngineID == "" {
>         ctx, span = Tracer(ctx).Start(ctx, node.PathString(),
>             telemetry.Reveal(),
>             trace.WithAttributes(...),
>         )
>         defer func() {
>             span.SetAttributes(attribute.Bool(telemetry.CheckPassedAttr, rerr == nil))
>             telemetry.EndWithCause(span, &rerr)
>         }()
>     }
> 
>     // ... rest unchanged
> }
> ```
> 
> ### Step 4: Extract local execution
> 
> Rename current `runLeafCheck()` body to `runLeafCheckLocally()` for clarity:
> 
> ```go
> func (node *ModTreeNode) runLeafCheckLocally(ctx context.Context) error {
>     var status dagql.AnyResult
>     if err := node.DagqlValue(ctx, &status); err != nil {
>         return err
>     }
>     // ... sync logic unchanged
> }
> ```
> 
> ---
> 
> ## Query Building for Remote Execution
> 
> The remote query must load the module based on its source kind. Port from `main:core/checks.go`:
> 
> ```go
> modSrc := node.Module.Source.Value.Self()
> 
> switch modSrc.Kind {
> case ModuleSourceKindLocal:
>     query = query.Select("moduleSource").
>         Arg("refString", filepath.Join(
>             modSrc.Local.ContextDirectoryPath,
>             modSrc.SourceRootSubpath,
>         ))
> 
> case ModuleSourceKindGit:
>     query = query.Select("moduleSource").
>         Arg("refString", modSrc.AsString()).
>         Arg("refPin", modSrc.Git.Commit).
>         Arg("requireKind", modSrc.Kind)
> 
> case ModuleSourceKindDir:
>     dirIDEnc, err := modSrc.DirSrc.OriginalContextDir.ID().Encode()
>     if err != nil {
>         return true, fmt.Errorf("encode dir ID: %w", err)
>     }
>     query = query.Select("loadDirectoryFromID").Arg("id", dirIDEnc)
>     query = query.Select("asModuleSource").
>         Arg("sourceRootPath", modSrc.DirSrc.OriginalSourceRootSubpath)
> }
> ```
> 
> ---
> 
> ## Files to Modify
> 
> | File | Changes |
> |------|---------|
> | `core/modtree.go` | Add `tryScaleOut()`, split `runLeafCheck()` → `runLeafCheckLocally()`, conditional telemetry in `RunCheck()` |
> 
> That's it. No changes needed to `checks.go` structure.
> 
> ---
> 
> ## New Imports Needed in `modtree.go`
> 
> ```go
> import (
>     "errors"
>     "path/filepath"
> 
>     "github.com/dagger/dagger/engine"
>     // ... existing imports
> )
> ```
> 
> ---
> 
> ## Summary
> 
> | Aspect | `main` branch | Recommended approach |
> |--------|--------------|---------------------|
> | Execution paths | Two (`CheckGroup.Run`, `Check.Run`) | One (`RunCheck` → `runLeafCheck`) |
> | Recursion prevention | Structural (hardcoded `false`) | Context-based (`EnableCloudScaleOut` flag) |
> | Telemetry dedup | Conditional span in `Check.Run()` | Conditional span in `RunCheck()` |
> | Diff size | Would require restoring dual paths | Minimal changes to `modtree.go` only |
> 
> **The single-path approach is the smallest, simplest diff** because it leverages the existing session-level `EnableCloudScaleOut: false` for recursion prevention, rather than requiring structural separation of originator vs target codepaths.
> 
> ---
> 
> ## Testing
> 
> 1. `_EXPERIMENTAL_DAGGER_CHECKS_SCALE_OUT=1 dagger check` - verify scale-out is attempted
> 2. Verify recursion doesn't happen (scaled-out engine doesn't try to scale out again)
> 3. Verify telemetry spans are not duplicated
> 4. Test all three module source kinds (Local, Git, Dir)
> 
> ---
> 
> ## Commits on This Branch
> 
> ```
> 903d85a52 fix: address lint issues in ModTree/checks
> 038dafd80 chore: regenerate SDKs
> b753155e2 core: ModTree infrastructure for checks
> ```
> 
> The main architectural commit is `b753155e2`. Review this to understand the full ModTree implementation.

core/modtree.go
```diff
@@ -0,0 +1,413 @@
...
+				continue
+			}
+		}
+		if len(exclude) > 0 {
+			if match, err := node.Match(ctx, exclude); err != nil {
+				return err
+			} else if match {
+				continue
+			}
+		}
```
eunomie, line 103: Include/exclude filtering uses the wrong variable - `node.Match()` is called instead of `child.Match()`. This means all children will be filtered based on the parent node's path rather than their own path, causing incorrect filtering behavior.

The FIXME comment on line 80 acknowledges this bug.

```suggestion
	for _, child := range children {
		// FIXME: filtering uses `node` instead of `child` - should match against the child being iterated
		if len(include) > 0 {
			if match, err := child.Match(ctx, include); err \!= nil {
				return err
			} else if \!match {
				continue
			}
		}
		if len(exclude) > 0 {
			if match, err := child.Match(ctx, exclude); err \!= nil {
				return err
			} else if match {
				continue
			}
		}
```

core/checks.go
```diff
@@ -230,252 +114,53 @@ func markdownTable(headers []string, rows ...[]string) string {
...
 	}
-	if r.Module != nil {
-		cp.Module = r.Module.Clone()
-	}
 	return &cp
 }
 
-func (c *Check) Name() string {
-	return strings.Join(c.Path, ":")
+func (c *Check) Path() []string {
```
eunomie, line 127: `Clone()` does a shallow copy with `cp := *r`, then modifies `cp.Checks[i]` in the loop. Since both slices share the same backing array after a shallow copy, writing to `cp.Checks[i]` also modifies the original `r.Checks[i]`.

This violates the expected semantics of Clone() - the caller expects an independent copy.

```suggestion
func (r *CheckGroup) Clone() *CheckGroup {
	cp := *r
	cp.Node = cp.Node.Clone()
	cp.Checks = make([]*Check, len(r.Checks))
	for i := range r.Checks {
		cp.Checks[i] = r.Checks[i].Clone()
	}
	return &cp
}
```

core/modtree.go
eunomie: 🫣 oops, that's from Claude...

core/checks.go
eunomie: 🫣 oops, that's from Claude too...

eunomie:
@shykes quick update on the PR:

- I rebased it, and changed to use the new toolchain registry (@kpenfound this removed the `WalkChecks` on `ToolchainRegistry` to use the new `ModTree.RollupChecks`. I hope that was the right change to do)
- I added back scale out, based on the plan you shared above: from my tests that looks good, but I'm not a huge user of it so happy for any feedback on it

CI's green, so if there's no concern I think we can imagine to merge it, and I'll refresh the generator branch on it.

shykes:
Nice thank you @eunomie ! @sipsma for scale-out, can we trust integration test, or is some manual testing required also?

sipsma:
> for scale-out, can we trust integration test, or is some manual testing required also?

No integ tests yet. Gonna be a bit of a project to figure those out since there's a lot of moving parts.

The main annoyance of testing manually is that you need the cloud infra to be using your dev engine too, so that you scale out to dev engine and not the old one:
1. Choose some public image registry you can publish to (I just use my dockerhub acct)
2. Hardcode the image name and tag you want to use here, to force cloud infra to use it (not exposed as a config, have to hardcode) https://github.com/sipsma/dagger/blob/2408141c2463e4c8e0685cd40e457c5132fcc109/internal/cloud/client.go#L178-L178
3. Publish the dev engine to that registry: `dagger call -m ./toolchains/engine-dev publish --image eriksipsma/dagger-test --tag test` (replace `image` and `tag` args as needed obviously`
4. Run against a dev engine w/ `dagger check --scale-out ...`

eunomie:
Thanks @sipsma for the test plan :-) I had never tried to run remotely with my own image.
And everything works fine!

So I guess we are good to go?

shykes:
Thanks @eunomie ! Can you take care of rebase & merge?

eunomie:
Yes, doing that now 👍

PR #11687, opened 2026-01-14
TUI: b to open services in browser

kpenfound:
https://github.com/user-attachments/assets/f3c41d78-373e-4774-bbaf-ec5654220d3a

### DIFF
--- a/dagql/dagui/db.go
+++ b/dagql/dagui/db.go
@@ -248,10 +248,8 @@ func (db DBLogExporter) Export(ctx context.Context, logs []sdklog.Record) error 
 			continue
 		}
 		spanID := SpanID{log.SpanID()}
-		if spanID == db.PrimarySpan {
-			// buffer raw logs so we can replay them later
-			db.PrimaryLogs[spanID] = append(db.PrimaryLogs[spanID], log)
-		}
+		// buffer raw logs so we can replay them later or extract structured data
+		db.PrimaryLogs[spanID] = append(db.PrimaryLogs[spanID], log)
 		// flag that the span has received logs
 		db.initSpan(spanID).HasLogs = true
 	}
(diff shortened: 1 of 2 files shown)


dagql/dagui/db.go
```diff
@@ -248,10 +248,8 @@ func (db DBLogExporter) Export(ctx context.Context, logs []sdklog.Record) error
 			continue
 		}
 		spanID := SpanID{log.SpanID()}
-		if spanID == db.PrimarySpan {
-			// buffer raw logs so we can replay them later
-			db.PrimaryLogs[spanID] = append(db.PrimaryLogs[spanID], log)
-		}
+		// buffer raw logs so we can replay them later or extract structured data
```
TomChv, line 251: Can this change have any side effect on the current TUI?

dagql/idtui/frontend_pretty.go
```diff
@@ -737,6 +745,67 @@ func KeyEnabled(enabled bool) key.BindingOpt {
...
+			if !strings.Contains(body, "tunnel started") {
+				continue
+			}
+
+			// Extract URL from the body text, skipping ANSI codes
+			if matches := serviceURLRegex.FindStringSubmatch(body); len(matches) > 1 {
+				return matches[1]
+			}
+		}
+		return ""
```
TomChv, line 782:
```suggestion

		return ""
```

For consistency with you implementation further down

TomChv (approved):
Left small question but overall LGTM, I give an early approval so this can be merged whenever comments are resolved.

@vito would be cool if you can also double check just in case.

dagql/dagui/db.go
vito: Yeah - this will collect all log records for all spans into memory which is quite expensive; that's why we only did it for the primary span before.

I think we should instead change the engine to emit telemetry in a way that we can detect at ingestion time in the frontend (`prettyLogs.Export`), by looking for specific log attributes or something.

PR #11699, opened 2026-01-15
chore: convert container.From to dagop

alexcb:
This converts container.From into a dagop, it includes code from `internal/buildkit/source/containerimage/` which has been copied under `core/containersource/`, then has a follow up commit which removes some of the unneeded code.

### DIFF
--- a/core/containersource/pull.go
+++ b/core/containersource/pull.go
@@ -22,8 +22,6 @@ import (
 	"github.com/dagger/dagger/internal/buildkit/util/estargz"
 	"github.com/dagger/dagger/internal/buildkit/util/imageutil"
 	"github.com/dagger/dagger/internal/buildkit/util/leaseutil"
-	"github.com/dagger/dagger/internal/buildkit/util/progress"
-	"github.com/dagger/dagger/internal/buildkit/util/progress/controller"
 	"github.com/dagger/dagger/internal/buildkit/util/pull"
 	"github.com/dagger/dagger/internal/buildkit/util/resolver"
 	digest "github.com/opencontainers/go-digest"
@@ -92,10 +90,6 @@ func (p *puller) Snapshot(ctx context.Context, g session.Group) (ir cache.Immuta
 	default:
 	}
 
-	// progressFactory needs the outer context, the context in `p.g.Do` will
-	// be canceled before the progress output is complete
-	progressFactory := progress.FromContext(ctx)
-
 	if p.cacheKeyErr != nil || p.cacheKeyDone {
 		return nil, p.cacheKeyErr
 	}
@@ -111,21 +105,12 @@ func (p *puller) Snapshot(ctx context.Context, g session.Group) (ir cache.Immuta
 	p.releaseTmpLeases = done
 	defer imageutil.AddLease(done)
 
-	resolveProgressDone := progress.OneOff(ctx, "resolve "+p.Src.String())
-	defer func() {
-		resolveProgressDone(err)
-	}()
-
 	p.manifest, err = p.PullManifests(ctx, getResolver)
 	if err != nil {
 		return nil, err
 	}
 
 	if len(p.manifest.Descriptors) > 0 {
-		progressController := &controller.Controller{
-			WriterFactory: progressFactory,
-		}
-
 		p.descHandlers = cache.DescHandlers(make(map[digest.Digest]*cache.DescHandler))
 		for i, desc := range p.manifest.Descriptors {
 			labels := snapshots.FilterInheritedLabels(desc.Annotations)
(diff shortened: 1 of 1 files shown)


alexcb:
@sipsma any thoughts on how much of the buildkit `containerimage` we want to cut out? Should I keep chipping away at it under this PR, or possibly as a follow up PR?

sipsma:
> @sipsma any thoughts on how much of the buildkit `containerimage` we want to cut out? Should I keep chipping away at it under this PR, or possibly as a follow up PR?

@alexcb Generally speaking let's air on the side of merging it sooner than later. We unlock a ton of important possibilities once we've removed all LLB (other than DagOp), most importantly removing the buildkit solver and associated complication. So I think we should be biased towards reaching that goal ASAP and then doing more cleanup after.

Taking a look now

core/containersource/pull.go
```diff
@@ -0,0 +1,266 @@
...
+		resolver := getOCILayoutResolver(p.store, p.SessionManager, g)
+		p.Puller.Resolver = resolver
+		// OCILayout has no need for session
+		getResolver = func(g session.Group) remotes.Resolver { return resolver }
+	default:
+	}
+
+	// progressFactory needs the outer context, the context in `p.g.Do` will
+	// be canceled before the progress output is complete
+	progressFactory := progress.FromContext(ctx)
```
sipsma: does everything work if we strip out all progress-related stuff? Here and elsewhere. Just curious because afaict we shouldn't rely on any of it anymore, but I could be forgetting somewhere

core/containersource/pull.go
```diff
@@ -0,0 +1,266 @@
...
+	defer func() {
+		resolveProgressDone(err)
+	}()
+
+	p.manifest, err = p.PullManifests(ctx, getResolver)
+	if err != nil {
+		return nil, err
+	}
+
+	if ll := p.layerLimit; ll != nil {
```
sipsma: we don't support the layer limit setting in our API (and don't think we ever really should), so this can probably be deleted

core/containersource/pull.go
```diff
@@ -0,0 +1,266 @@
...
+	}()
+
+	var parent cache.ImmutableRef
+	setWindowsLayerType := p.Platform.OS == "windows" && runtime.GOOS != "windows"
+	for _, layerDesc := range p.manifest.Descriptors {
+		parent = current
+		current, err = p.CacheAccessor.GetByBlob(ctx, layerDesc, parent,
+			p.descHandlers, cache.WithImageRef(p.manifest.Ref))
+		if parent != nil {
+			parent.Release(context.TODO())
```
sipsma: nit: `context.WithoutCancel(ctx)` instead of context.TODO

core/schema/container.go
```diff
@@ -820,12 +853,35 @@ func (s *containerSchema) from(ctx context.Context, parent dagql.ObjectResult[*c
...
 
-		return dagql.NewResultForCurrentID(ctx, ctr)
+		// Note that this must be done outside the dagop context, otherwise the image config data will be lost
+		ctr.FromCanonicalRefUpdateConfig(ctx, refName, nil)
+
+		if ctr.FS.Self().Dir == "" {
+			// FIXME The dir is set under FromCanonicalRef; however it's done inside a dagop, which doesn't propigate back here
+			// As a result, if FromCanonicalRef sets it to anything besides a "/", it will preemptively return an error.
+			// Ideally, we should simply return an error here when this is empty (indicating it failed to propigate)
+			// however, for the time being we will just set it to the root dir.
```
sipsma, line 876: I can't really think of any situation where it could ever make sense for Dir to be anything but `/` after a pull, so I don't think this is a problem

core/container.go
```diff
@@ -502,53 +503,144 @@ func (mnts ContainerMounts) Replace(newMnt ContainerMount) (ContainerMounts, err
...
-func (container *Container) FromCanonicalRef(
+// FromCanonicalRefUpdateConfig is must be called outside of a dagop context, and is responsible for fetching the image config
+// and applying it to the container's metadata
+func (container *Container) FromCanonicalRefUpdateConfig(
 	ctx context.Context,
 	refName reference.Canonical,
 	// cfgBytes is optional, will be retrieved if not provided
 	cfgBytes []byte,
 ) (*Container, error) {
-	container = container.Clone()
```
sipsma, line 550: double checking rm'ing the clone here was intentional

alexcb: it was -- It's called directly on the `DagOpContainer` that's created which triggers the call to `FromCanonicalRef` (which sets fetches the image and sets the ref to the snapshot's immutableRef)

core/containersource/pull.go
alexcb: good catch!

core/containersource/pull.go
alexcb: it does -- we can remove it :)

core/containersource/pull.go
alexcb: :axe:

PR #11700, opened 2026-01-15
feat: init on toolchain install

eunomie:
When a user runs dagger toolchain install without a dagger module, automatically init the dagger module in the current directory.

If -m/--mod or DAGGER_MODULE is set, do not init but keep the previous behavior.

### DIFF
--- a/cmd/dagger/module.go
+++ b/cmd/dagger/module.go
@@ -239,7 +239,8 @@ dagger init --sdk=go
 				}
 			}
 
-			modSrc := dag.ModuleSource(srcRootArg, dagger.ModuleSourceOpts{
+			// Check if module already exists before preparing
+			checkModSrc := dag.ModuleSource(srcRootArg, dagger.ModuleSourceOpts{
 				// Tell the engine to use the provided arg as the source root, don't
 				// try to find-up a dagger.json in a parent directory and use that as
 				// the source root.
@@ -252,8 +253,7 @@ dagger init --sdk=go
 				// We can only init local modules
 				RequireKind: dagger.ModuleSourceKindLocalSource,
 			})
-
-			alreadyExists, err := modSrc.ConfigExists(ctx)
+			alreadyExists, err := checkModSrc.ConfigExists(ctx)
 			if err != nil {
 				return fmt.Errorf("failed to check if module already exists: %w", err)
 			}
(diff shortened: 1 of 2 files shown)


cmd/dagger/module.go
```diff
@@ -783,6 +771,8 @@ var toolchainInstallCmd = &cobra.Command{
 			if err != nil {
 				return err
 			}
+			_, explicitModRef := getExplicitModuleSourceRef()
+
 			modSrc := dag.ModuleSource(modRef, dagger.ModuleSourceOpts{
```
grouville: Shouldn't we also have the same constraints as https://github.com/dagger/dagger/pull/11700/changes#diff-5776ff0964ca98cf11dfeea0d49469d367ae59dad47427190f6dbdf53004c20aR1247 ?

Between the check of the existence of the toolchain and the initialization, would it make sense to merge the checks to avoid any transient bug ?

core/integration/toolchain_test.go
```diff
@@ -320,6 +320,84 @@ func (ToolchainSuite) TestToolchainsWithConfiguration(ctx context.Context, t *te
...
+		// Note: NOT calling "dagger init" - the toolchain install should auto-init
+
+		// Install toolchain - should auto-initialize the module first
+		_, err := modGen.
+			With(daggerExec("toolchain", "install", "../hello")).
+			Stdout(ctx)
+		require.NoError(t, err)
+
+		// Verify the module was initialized by checking dagger.json exists
+		modGen = modGen.With(daggerExec("toolchain", "install", "../hello"))
```
grouville: nit: I don't think it's necessary to re-execute again as it's been done on line 334 ?

Maybe something like (pattern already being used on line 52 and 416 by Kyle)

```go
modGen = modGen.With(daggerExec("toolchain", "install", "../hello"))
_, err := modGen.Stdout(ctx)
require.NoError(t, err) 

// Verify the module was initialized by checking dagger.json exists
configContent, err := modGen.File("dagger.json").Contents(ctx)
```

core/integration/toolchain_test.go
```diff
@@ -320,6 +320,84 @@ func (ToolchainSuite) TestToolchainsWithConfiguration(ctx context.Context, t *te
...
+		// Install toolchain - should auto-initialize with directory name
+		out, err := modGen.
+			With(daggerExec("toolchain", "install", "../hello")).
+			Stdout(ctx)
+		require.NoError(t, err)
+		require.Contains(t, out, "Initialized module my-custom-app")
+
+		// Verify the module name in dagger.json
+		configContent, err := modGen.
+			With(daggerExec("toolchain", "install", "../hello")).
```
grouville: same here

grouville (approved):
Nice !

As it touches the UX, I'll let @shykes or @vito confirm the overview.

Running `dagger toolchain install X` without an existing module is equivalent to `dagger init && dagger toolchain install X`, so LGTM on my end ✅ 

> PS: I left 2 non-blocking nits to just unify a small testing pattern on the file (feel free to discard)

eunomie:
Thanks @grouville, I updated the tests, better now :-)

core/integration/toolchain_test.go
```diff
@@ -320,6 +320,70 @@ func (ToolchainSuite) TestToolchainsWithConfiguration(ctx context.Context, t *te
...
+		_, err := modGen.
+			With(daggerExec("toolchain", "install", "--mod=.", "../hello")).
+			Stdout(ctx)
+		require.Error(t, err)
+	})
+
+	t.Run("no auto-init when module already exists", func(ctx context.Context, t *testctx.T) {
+		modGen := toolchainTestEnv(t, c).
+			WithWorkdir("app").
+			With(daggerExec("init")) // Initialize first
```
grouville: Second pass: comment unnecessary

PR #11704, opened 2026-01-15
feat(sdk/java): add checks support to Java SDK

henkhofs:
## Summary

This PR adds support for the **checks** feature to the Java SDK, bringing it to feature parity with Python, TypeScript, and Go SDKs.

### Changes

- **New `@Check` annotation**: Marks functions as checks that can be executed for validation
- **Updated `FunctionInfo` record**: Added `isCheck` field to track check status
- **Updated annotation processor**: Detects `@Check` annotation and generates `withCheck()` call
- **Test module added**: `hello-with-checks-java` demonstrates usage with passing and failing checks

### Example Usage

```java
@Object
public class MyModule {
    @Function
    @Check
    public void validateBuild() throws Exception {
        Container container = Dagger.dag()
            .container()
            .from("alpine:3")
            .withExec(new String[]{"sh", "-c", "exit 0"});
        container.sync();
    }
}
```

### Implementation Details

1. **Check.java**: New annotation similar to `@Function` that can be combined with it
2. **FunctionInfo.java**: Added `isCheck` boolean field to the record
3. **DaggerModuleAnnotationProcessor.java**: 
   - Added `@Check` to supported annotations
   - Detects `@Check` on methods and sets `isCheck = true` in FunctionInfo
   - Generates `.withCheck()` call in the function definition code
4. **hello-with-checks-java**: Test module with passing and failing check examples

### Testing

The test module can be tested with:
```bash
cd core/integration/testdata/checks/hello-with-checks-java
dagger call passing-check  # Should succeed
dagger call failing-check  # Should fail
```

### DIFF
--- a/core/integration/testdata/checks/hello-with-checks/main.go
+++ b/core/integration/testdata/checks/hello-with-checks/main.go
@@ -3,6 +3,7 @@ package main
 
 import (
 	"context"
+	"sort"
 
 	"dagger/hello-with-checks/internal/dagger"
 )
@@ -46,7 +47,7 @@ func (m *HelloWithChecks) FailingContainer() *dagger.Container {
 	return dag.Container().From(m.BaseImage).WithExec([]string{"sh", "-c", "exit 1"})
 }
 
-Returns the names of all checks visible from the current environment.
+// Returns the names of all checks visible from the current environment.
 func (m *HelloWithChecks) CurrentEnvChecks(ctx context.Context) ([]string, error) {
 	checks, err := dag.CurrentEnv().Checks().List(ctx)
 	if err != nil {


sdk/java/dagger-java-sdk/src/main/java/io/dagger/module/annotation/Check.java
```diff
@@ -0,0 +1,34 @@
...
+ *   @Check
+ *   public void validateBuild() {
+ *     // Validation logic here
+ *     // Throw exception on failure
+ *   }
+ * }
+ * }</pre>
+ *
+ * <p>This annotation can be combined with {@link Function} to create functions that are both
+ * callable and marked as checks.
```
eunomie: The check must be on a callable function. When running dagger check we are looping through the exported, callable functions. If we can't call it, we can't call it as a check.

core/integration/testdata/checks/hello-with-checks-java/src/main/java/io/dagger/modules/hellowithchecksjava/HelloWithChecksJava.java
```diff
@@ -0,0 +1,34 @@
+import io.dagger.client.Container;
+import io.dagger.client.Dagger;
+import io.dagger.module.annotation.Check;
+import io.dagger.module.annotation.Function;
+import io.dagger.module.annotation.Object;
+
+/** A module for HelloWithChecksJava functions */
+@Object
+public class HelloWithChecksJava {
```
eunomie, line 12: Good to have this test module :+1:

Could you plug in it the `TestChecks` suite? (`core/integration/checks_test.go`)
:thinking: maybe have a look at the `TestJava` suite in case we need some magic (to allow this test module to use the SDK defined in the PR, so local, instead of the default that will point to git and not available until the PR is merged)

eunomie (commented):
Thanks for the PR!

Just a few comments.

Also it could be great to extend the existing test in `dagger-java-annotation-processor`:
- add a `@check` function (or add the annotation to one existing) to the test module in `src/java/dagger-java-annotation-processor/src/test/java/resources/io/dagger/java/module/DaggerJava.java`
- update `src/java/dagger-java-annotation-processor/src/test/java/resources/io/dagger/gen/entrypoint/Entrypoint.java` to show the `withCheck`

henkhofs:
Thanks for the feedback @eunomie! I've addressed all three points:

1. **@Check annotation documentation** - Updated to clarify that `@Check` must be combined with `@Function`. Checks must be callable functions.

2. **Unit tests** - Extended the annotation processor test suite:
   - Added `@Check` annotation to `validate()` function in test module
   - Updated expected `Entrypoint.java` to include `.withCheck()` call
   - Added invoke case for the validate function

3. **Integration tests** - Added Java checks test to `checks_test.go`:
   - New test case: `run checks from a blueprint (Java)`
   - Tests listing, running passing checks, running failing checks, and running all checks
   - Follows the same pattern as Go, TypeScript, and Python tests

All changes are in commit afbc99e62.

eunomie:
This can't really work that way.
You can have a look at how the java modules are tested in the `JavaSuite`, especially how they are using the `javaModule` function: https://github.com/dagger/dagger/blob/41d478768fd1237faed705f4137d3016ab074b88/core/integration/module_java_test.go#L379C1-L391C2

You can run the tests using: `dagger call engine-dev test --run=TestChecks --pkg=./core/integration` to be sure it's working before to push.

henkhofs:
thx, will work on it some more and will not push until those tests work/pass

henkhofs:
Quick update: the integration test was failing because Java blueprints need a `pom.xml` at the root level (similar to how Python needs `pyproject.toml`, TypeScript needs `package.json`, etc.). Added the missing `pom.xml` to the test blueprint and now all checks pass! ✅

henkhofs:
I don't think the ci failures have to do with this change, but could use a double-check :)

eunomie (approved):
I improved the check test suite, otherwise that sounds good to me.
Just waiting for a final green CI on tests (the red one of security check is something that should be fixed in a different PR, and not related to this one) and I'll merge it.
Thanks @henkhofs for the contribution

PR #11708, opened 2026-01-16
feat: use the remote library on versioned engine instead the local one

TomChv:
This removes the generation of internal/telemetry and internal/querybuilder that can be installed from the local library.

Add a new flag to --generate-module to pull the libraty that matches the  engine version's.

Add a new `_EXPERIMENTAL_DAGGER_DEV_ENGINE` to force the engine to act as a dev engine even if its tagged so we don't break engine_test files if we do some breaking changes in telemetry or querybuilder package.

Related to #11515 and #11547

### DIFF
--- a/core/integration/engine_test.go
+++ b/core/integration/engine_test.go
@@ -520,6 +520,9 @@ func (EngineSuite) TestModuleVersionCompat(ctx context.Context, t *testctx.T) {
 					return c.
 						WithEnvVariable("_EXPERIMENTAL_DAGGER_VERSION", tc.engineVersion).
 						WithEnvVariable("_EXPERIMENTAL_DAGGER_MIN_VERSION", tc.moduleMinVersion).
+						// Required to use the bundled go SDK instead of the remote one
+						// This avoid breaking that test because of a breaking change on
+						// the generated files.
 						WithEnvVariable("_EXPERIMENTAL_DAGGER_DEV_ENGINE", "1")
 				})
 				devEngineSvc = devEngineContainerAsService(devEngine)
--- a/toolchains/engine-dev/main.go
+++ b/toolchains/engine-dev/main.go
@@ -193,9 +193,14 @@ func (dev *EngineDev) Container(
 		WithFile(cliPath, cli).
 		WithEnvVariable("_EXPERIMENTAL_DAGGER_RUNNER_HOST", distconsts.DefaultEngineSockAddr)
 
-	// ctr = ctr.WithEnvVariable("_EXPERIMENTAL_DAGGER_VERSION", "v0.19.10")
-	// ctr = ctr.WithEnvVariable("_EXPERIMENTAL_DAGGER_DEV_ENGINE", "1")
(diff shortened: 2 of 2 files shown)


cmd/codegen/generator/go/generate_module.go
```diff
@@ -240,12 +251,17 @@ func (g *GoGenerator) syncModReplaceAndTidy(mod *modfile.File, genSt *generator.
 		}
 	}
 
+	if g.Config.ModuleConfig.LibVersion != "" {
+		genSt.PostCommands = append(genSt.PostCommands, exec.Command("go", "get", fmt.Sprintf("dagger.io/dagger@%s", g.Config.ModuleConfig.LibVersion)))
```
sipsma: Can we edit the go.mod we write to have this dep rather than call go get as a subprocess? Some of these go subprocesses are on the slow side, and we really need to fight for perf gains in the go codegen

toolchains/engine-dev/main.go
```diff
@@ -192,6 +192,9 @@ func (dev *EngineDev) Container(
 	ctr = ctr.
 		WithFile(cliPath, cli).
 		WithEnvVariable("_EXPERIMENTAL_DAGGER_RUNNER_HOST", distconsts.DefaultEngineSockAddr)
+
+	// ctr = ctr.WithEnvVariable("_EXPERIMENTAL_DAGGER_VERSION", "v0.19.10")
+	// ctr = ctr.WithEnvVariable("_EXPERIMENTAL_DAGGER_DEV_ENGINE", "1")
```
sipsma, line 203: nit: cleanup before merge

core/integration/engine_test.go
```diff
@@ -520,7 +519,8 @@ func (EngineSuite) TestModuleVersionCompat(ctx context.Context, t *testctx.T) {
 				devEngine := devEngineContainer(c, func(c *dagger.Container) *dagger.Container {
 					return c.
 						WithEnvVariable("_EXPERIMENTAL_DAGGER_VERSION", tc.engineVersion).
-						WithEnvVariable("_EXPERIMENTAL_DAGGER_MIN_VERSION", tc.moduleMinVersion)
+						WithEnvVariable("_EXPERIMENTAL_DAGGER_MIN_VERSION", tc.moduleMinVersion).
+						WithEnvVariable("_EXPERIMENTAL_DAGGER_DEV_ENGINE", "1")
```
sipsma, line 526: nit: comment on why this is needed will probably save someone time down the line

sipsma (approved):
LGTM, couple of nits before merge

toolchains/engine-dev/main.go
TomChv: I think we should keep these, so if someone wants to compile a dev engine, I added doc to explain why :D

core/integration/engine_test.go
TomChv: Done in [d54602a](https://github.com/dagger/dagger/pull/11708/commits/d54602a5293b274502b85e547dff9c50bdf8845a)

PR #11714, opened 2026-01-16
allow container address in default pragrams and customizations

kpenfound:
This lets you do

```go
// +defaultAddress="alpine:latest"
ctr *dagger.Container,
```

and

```json
{
  "function": ["testWithDefaultContainer"],
  "argument": "ctr",
  "defaultAddress": "alpine:3.18"
}
```

in json.

Added lots of tests so hopefully this isn't a terrible idea.

In making this I discovered that our documentation for python constructors with complex types has been wrong all along, so I've corrected that too.

Fixes https://github.com/dagger/dagger/issues/8483


kpenfound:
Question to follow up on from @shykes : Why does this work with `+default` when we had to use `+defaultPath` for context directories? Need to dig up the context on that. Maybe this should be `+defaultAddress` instead. Or if the original restriction is no longer valid, should `+defaultPath` become `+default`?

kpenfound:
Also need to update the typescript test to match the others

eunomie:
I don't remember why `defaultPath`, IIRC there's a good reason we can't get ride of it.
So in that spirit it can make sense to have something like `defaultAddress`.
But `default` is good too, and I like it, it can help to make some function smoother.

kpenfound:
Changed to `+defaultAddress` to separate the discussion on just using `+default` for everything

kpenfound:
Created this issue for discussion: https://github.com/dagger/dagger/issues/11745

core/integration/module_go_test.go
```diff
@@ -1886,3 +1886,41 @@ func (p *Playground) SayHello() string {
...
+		)
+
+	out, err := modGen.With(daggerCall("test-with-default-container")).Stdout(ctx)
+	require.NoError(t, err)
+	require.Contains(t, out, ".") // Alpine version contains a dot like "3.19.0"
+
+	// Test that we can override the default
+	out2, err := modGen.With(daggerCall("test-with-default-container", "--ctr=alpine:3.18")).Stdout(ctx)
+	require.NoError(t, err)
+	require.Contains(t, out2, ".")
```
eunomie: nit: should we check it contains 3.18?

core/integration/module_python_test.go
```diff
@@ -1921,3 +1921,34 @@ func pipLockMod(t *testctx.T, c *dagger.Client, inc []string) dagger.WithContain
...
+    ctr: dagger.Container = dataclasses.field(
+        default_factory=lambda: dag.container().from_("alpine:3.19")
+    )
+
+    @function
+    async def version(self) -> str:
+        return await self.ctr.with_exec(
+            ["/bin/sh", "-c", "cat /etc/alpine-release"]
+        ).stdout()
+`)
```
eunomie, line 1942: nit: should we instead test a version that is using `DefaultAdress` and not the constructor way? To be more aligned with the Go version of the test.

```suggestion
@object_type
class Test:
    @function
    async def version(
        self,
        ctr: Annotated[dagger.Container, DefaultAddress("alpine:3.19")],
    ) -> str:
        return await ctr.with_exec(["cat", "/etc/alpine-release"]).stdout()
`)
```

PR #11720, opened 2026-01-20
internal dev `cache-expert` agent skill

sipsma:
Had some promising but not quite fruitful trials w/ latest LLM coding models (codex 5.2-xhigh + opus 4.5) when it comes to debugging super gnarly engine issues around caching.

Seemed like one bottleneck was that it had to learn (and re-learn after context compaction) the whole codebase every time (which is often necessary for particularly tricky problems). So giving it detailed developer docs that focus it in on the parts that truly matter when debugging cache stuff may save a lot of file read/reasoning tokens.

Another issue was that it needed to be told all the time (or discover on its own) generally how to debug. Not just what commands to run but the techniques (i.e. println debugging of cache keys will get you very far).

Trying to package this all up in a skill. ATM the main utility of skill vs other packaging is just that LLMs have baked in knowledge on progressive disclosure of various skill docs, which is highly important here so we don't spend *all* our tokens just reading these docs.

---

This is still a WIP, it's basically become "the ultimate brain dump". I've been chipping away in the evenings, doing a mix of hand human writing and having LLMs interview me about a topic and then write everything it learned from me and from the code.

There's still a lot left to cover but probably reaching the point of utility. May give it a try on some caching related flakes and see how it does.


dagql/README.md
sipsma, line 1: This doc is quite out of date and seemed to confuse the LLM more than help, so I just rm'd it here. It's largely covered in new docs added to the skill here.

sipsma: cc @vito since you wrote it way back when

PR #11725, opened 2026-01-21
sdk(python): fix annotations with `from __future__ import annotations`

### LINKED ISSUE #11554: 🐞 `DefaultPath` doesn't work in Python

grouville:
Fixes #11554

## Context
When `from __future__ import annotations` (PEP 563) is enabled, `param.annotation` becomes a string instead of an actual type object. This causes all metadata extraction (`DefaultPath`, `Doc`, `Name`, `Ignore`, `Deprecated`) to fail silently: the annotations are simply ignored.

### Fix
The fix uses `get_type_hints(..., include_extras=True)` to resolve string annotations back to actual types while preserving `Annotated` metadata.

### To test

```sh
dagger call python-sdk test-suite run --args="tests/mod/test_future_annotations.py" --args="-v"
```

### DIFF
--- a/sdk/python/src/dagger/mod/_resolver.py
+++ b/sdk/python/src/dagger/mod/_resolver.py
@@ -111,6 +111,15 @@ class Function(Generic[P, R]):
         return get_type_hints(self.wrapped)
 
     @cached_property
+    def type_hints_with_extras(self):
+        """Type hints with Annotated metadata preserved.
+
+        Used for extracting metadata like DefaultPath, Doc, Name, etc.
+        from parameters when `from __future__ import annotations` is used.
+        """
+        return get_type_hints(self.wrapped, include_extras=True)
+
+    @cached_property
     def signature(self):
         return inspect.signature(self.wrapped, follow_wrapped=True)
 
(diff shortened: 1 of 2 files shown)


grouville:
failure seems to be a flake, test passes locally

PR #11726, opened 2026-01-21
core: add MainModule to Env

shykes:
Change how modules are attached to `Env`, to better match how dagger clients load modules.

- `Env.mainModule`: the main module for this environment. This represents the project being worked on.
- Deprecate `Env.withModule`: to merge functions from several modules into the env, use toolchains on the main module instead.

This unblocks simplifying `Check` (by reverting to the simpler "one checkgroup, one root module logic") and clears the way to merging #11685 .

### DIFF
--- a/docs/docs-graphql/schema.graphqls
+++ b/docs/docs-graphql/schema.graphqls
@@ -2480,11 +2480,18 @@ type Env {
   ): Env!
 
   """
+  Sets the main module for this environment (the project being worked on)
+
+  Contextual path arguments will be populated using the environment's workspace.
+  """
+  withMainModule(module: ModuleID!): Env!
+
+  """
   Installs a module into the environment, exposing its functions to the model
 
   Contextual path arguments will be populated using the environment's workspace.
   """
-  withModule(module: ModuleID!): Env!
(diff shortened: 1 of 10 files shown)


shykes:
@eunomie my understanding:

- Currently in main, `currentEnv().checks()` does *not* work as intended when called from a toolchain. This is a known issue that you were planning to fix
- This PR doesn't fix that issue (`currentEnv().checks()` still won't work from a toolchain), but we should still merge it, because it provides plumbing for 1) simplifying checks implementation, 2) unblocking [planned checks improvements](#11685), and 3) harmonizing how `Env` deals with modules
- To fix the unresolved issue, we need a way to access the "main module" from the context of a toolchain. You are looking for solutions, and will either push a fix in this pr, or in a follow-up pr - up to you.


cc @vito @kpenfound for awareness ☝️ 

[Relevant discord thread](https://discord.com/channels/707636530424053791/1463332258453651569)

eunomie:
@shykes yes, that's it 👌 
I'm currently looking at fixing the access from toolchains. If I can have a fix today, I'll push it to the branch and we can merge. And if I can't have a fix today, I suggest we still merge it: it fixes your issues with #11685 and as the toolchain issue is already in main it doesn't make things worse.

PR #11729, opened 2026-01-22
finish separating content digests out; refactor ids/cache keys

sipsma:
Couple interrelated goals here:
1. Fix the root cause of the `TestConstructor` flakes by more fundamentally squashing the possibility of content digest overlap resulting in cross-client cache confusion
2. Setup some cache key + ID changes that will be needed for the imminent removal of the buildkit solver
3. Finish what I originally started in https://github.com/dagger/dagger/pull/11625 but had to partially back out of due to needing a fix ASAP

Main thing is to manage "content digests" separately from "recipe digests", which allows us to have content based cache hits without losing the original recipe (which can create confusion across clients that happen to be getting cache hits based on content digests).

This is done by tracking them as separate fields in IDs and in the various engine cache structs.

In parts of the code that care, content digests are preferred over recipe digests when the content digest is available. The primary motivation for that is it's how everything intrinsically worked before, it's just much more explicit now. We also have the freedom to modify that in the future now if desired.

---

Another associated change that came along for the ride:

Previously, the `engine/cache/cache.go` package only knew about the call keys at the start of the call, with handling of results that have different keys associated handled outside the package.
   * e.g. `address("./foo).directory` returns a result for the call `host.directory("./foo")`, which is also content hashed so there's 3 keys involved:
     1. the cache key for `address("./foo).directory`
     2. the cache key for `host.directory("./foo")`
     3. the content digest cache key for `host.directory("./foo")` (since we always content has local dir loads)
   * previously ii) and iii) above were managed outside of the `cache` package in confusing ways. Now it's all managed "in-house" in the `cache` package. It's the same level of complexity in a sense, but at least more centralized.

As a next step along these lines I'm gonna update the `engine/cache` package from using its own struct full of various string keys to just be away of IDs and accept those. This should cleanup+centralize things even further while also adding important abilities like being able to track dependencies between results based on the receivers/args of IDs
* Was tempted to include in this PR but would rather get this one merged earlier since it fixes one of the more annoying flakes.

---

TODO:
- [x] Fix `TestConstructor` flakes
   - Could repro consistently before, can't anymore
- [x] fix remaining test failures
- [x] doc more, cleanup printlns

### DIFF
--- a/skills/cache-expert/SKILL.md
+++ b/skills/cache-expert/SKILL.md
@@ -13,42 +13,7 @@ Each operation takes immutable objects/scalar values as inputs and produces an i
 
 This enables caching: since inputs are immutable and operations are deterministic, cache keys can be derived from the operation and its inputs.
 
-**Key concepts:**
-- **ID**: A scalar value that encapsulates the operation that created an object. Enables non-scalar values to be used as inputs to other operations, forming a DAG.
-- **Digest**: A hash derived from an operation and its inputs, used for cache key computation.
-- **Call Cache Key**: Determines whether an operation's result can be retrieved from cache.
-
-## Core API Call Anatomy
-
-Every API call has three typed components:
-
-1. **Parent** - The object the operation is called on (e.g., `Container`)
-2. **Arguments** - The operation's input arguments
-3. **Return value** - The result (scalar or object with an ID)
-
-### Cache Key Computation
-
-By default, a call's cache key is a hash of:
-- Operation name (e.g., `Container.withExec`)
-- Parent's digest
-- Arguments' digests
-
-This matches the default ID digest. See [ids.md](references/ids.md) for details.
-
-### Cache Key Customization
-
-Cache keys can be scoped differently:
-- **Per-client** - Cached per connected client
-- **Per-session** - Cached for the duration of a session
-- **Per-call** - Never cached (unique each invocation)
-- **Custom** - Arbitrary cache key logic
-
-### Object IDs vs Cache Keys
-
-These are related but distinct:
-- **Call cache key** - Used to look up cached results
-- **Returned object ID** - May equal the cache key, or may be a separate operation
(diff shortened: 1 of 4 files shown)


core/integration/suite_test.go
```diff
@@ -65,7 +65,7 @@ func BenchMiddleware() []testctx.Middleware[*testing.B] {
 func connect(ctx context.Context, t testing.TB, opts ...dagger.ClientOpt) *dagger.Client {
 	opts = append([]dagger.ClientOpt{
 		// FIXME: test spans are easier to read in the TUI when this is silenced
-		dagger.WithLogOutput(io.Discard),
+		dagger.WithLogOutput(testutil.NewTWriter(t)),
```
vito: nit: comment above is no longer accurate, unsure if meant to commit

dagql/call/callpbv1/call.proto
```diff
@@ -75,6 +75,11 @@ message Call {
   // Whether the digest was set to a custom value, as opposed to the default
   // "recipe-based" digest
   bool isCustomDigest = 11;
```
vito, line 77: Is this still needed? 🤔 I feel like it is, but it's just slightly confusing because previously custom digests _were_ often content digests. Is there a third "class" of digests we still have which uses this, and can't use a content digest?

core/integration/suite_test.go
sipsma: yep was just a tmp change, removed

dagql/call/callpbv1/call.proto
sipsma: No it's definitely not needed anymore. I had it removed at one point as part of this but then I backed out after it didn't work first try; would rather get this PR out asap to fix the annoying flakes than keep piling more on.

There's more PRs coming soon in this area anyways, so just gonna revisit it then!

PR #11730, opened 2026-01-22
initial draft of agent skill for daggerizing a repo

sipsma:
This is a rough first draft based purely on a single instance where I daggerized a particular repo. I daggerized it via prompting and then asked the LLM to create a skill out of what it learned. Likely useful enough to merge as is and iterate on.

I don't think the opinions it expresses are likely what everyone shares, but I also don't have time to get bogged down in refining details right now since this was just a quick extracurricular thing. If there's preference to not merge I'll just close this for now and continue iterating on it on my own as time permits.


.skills/daggerize/SKILL.md
```diff
@@ -0,0 +1,49 @@
+---
+name: daggerize
+description: Add or improve Dagger automation in a codebase. Use when asked to create Dagger modules or toolchains, wire up dagger.json, or define build/test/run/service commands callable via `dagger call` (including `up --ports` for services). Also use to diagnose Dagger execution errors and adjust containers, caches, or inputs.
```
shykes: I would add language to match requests that don't specifically mention Dagger, but involve solving a problem that Dagger can help solve.

For example: "use when asked to make builds or integration tests more repeatable, more portable across environments, or easier to run locally. Also a good option for accelerating builds with caching, incremental execution. Do not use for native Windows or Mac execution"

Note: this can be a follow-up.

kpenfound:
Looks pretty good! I'm curious how it would discover toolchains that apply to the project. Maybe we need a static list for official ones before we undertake a daggerverse redesign. Here's the markdowns I used for daggerizing before claude renamed this pattern to skills https://github.com/kpenfound/dagger-programmer/tree/main/moduleWorkspace/knowledge

PR #11608, opened 2025-12-29
fix: support kernels without legacy xtables (6.17+)

shepherdjerred:
## Summary
Fix Dagger Engine failing on Linux kernel 6.17+ where `CONFIG_NETFILTER_XTABLES_LEGACY` is disabled.

Image available here: https://github.com/users/shepherdjerred/packages/container/package/dagger-engine

```
2025-12-29T23:34:37.727Z | dnsmasq[739]: read /etc/hosts - 18 names
2025-12-29T23:34:37.727Z | dnsmasq[739]: read /var/run/containers/cni/dnsname/dagger/addnhosts - 0 names
2025-12-29T23:34:37.732Z | time="2025-12-29T23:34:37Z" level=info msg="auto snapshotter: using overlayfs"
2025-12-29T23:34:37.735Z | time="2025-12-29T23:34:37Z" level=warning msg="failed to release network namespace \"kxhj354vbn6ent9a2x1tp6s9d\" left over from previous run: plugin type=\"loopback\" failed (delete): unknown FS magic on \"/var/lib/dagger/net/cni/kxhj354vbn6ent9a2x1tp6s9d\": 2fc12fc1"
2025-12-29T23:34:37.767Z | dnsmasq[739]: read /etc/hosts - 18 names
2025-12-29T23:34:37.767Z | dnsmasq[739]: read /var/run/containers/cni/dnsname/dagger/addnhosts - 0 names
2025-12-29T23:34:37.794Z | dnsmasq[739]: read /etc/hosts - 18 names
2025-12-29T23:34:37.794Z | dnsmasq[739]: read /var/run/containers/cni/dnsname/dagger/addnhosts - 0 names
2025-12-29T23:34:37.824Z | dnsmasq[739]: read /etc/hosts - 18 names
2025-12-29T23:34:37.824Z | dnsmasq[739]: read /var/run/containers/cni/dnsname/dagger/addnhosts - 0 names
2025-12-29T23:34:37.852Z | dnsmasq[739]: read /etc/hosts - 18 names
2025-12-29T23:34:37.852Z | dnsmasq[739]: read /var/run/containers/cni/dnsname/dagger/addnhosts - 0 names
2025-12-29T23:34:37.878Z | dnsmasq[739]: read /etc/hosts - 18 names
2025-12-29T23:34:37.878Z | dnsmasq[739]: read /var/run/containers/cni/dnsname/dagger/addnhosts - 0 names
[...]

PR #11731, opened 2026-01-22
(updated+fixed) support kernels without legacy xtables (6.17+) 

### LINKED ISSUE #11694: Add support for kernels without legacy xtables
https://github.com/dagger/dagger/pull/11608 attempted this but ended up causing perf regressions for unclear reasons. It may have resulted in CNI creation/teardown overhead that caused each container to take longer? Needs more investigation

sipsma:
This is an update to the reverted https://github.com/dagger/dagger/pull/11608 which fixes the performance regression that introduced.

After investigation, the previous PR was fine except that it removed the `ip6tables` package.

When I run `TestModule` (pretty heavy integ test suite):
* With the previous PR: 7m39s
* With the previous PR ***but leave in ip6tables***: 5m56s

These numbers are pretty consistent on my laptop, not noise.

It's not 100% clear to me yet why that is, but may be related to the fact that even when we don't use ipv6, CNI binaries attempt to do some ipv6 related setup that ends up going through a slower path when ip6tables isn't around.

I haven't yet tested this on a 6.17+ kernel or one with legacy iptables disabled, so I can't yet be sure leaving in ip6tables works on those machines. cc @shepherdjerred @Kernald @ReillyBrogan @ezynda3 in case any of you have time to test it quick. Otherwise we'll try to do a test before the next release.

---

Left your commit in place @shepherdjerred with an extra on top, so you'll still get credit, appreciate the original PR!

---

Fixes #11694

### DIFF
--- a/.changes/unreleased/Fixed-20260121-214039.yaml
+++ b/.changes/unreleased/Fixed-20260121-214039.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: |
+  support kernels without legacy xtables (Linux 6.17+ and older ones with iptables disabled)
+time: 2026-01-21T21:40:39.097979296-08:00
+custom:
+  Author: shepherdjerred + sipsma
+  PR: "11731"
--- a/cmd/engine/main.go
+++ b/cmd/engine/main.go
@@ -901,6 +901,10 @@ func setupNetwork(ctx context.Context, netName, netCIDR string) (*networkConfig,
 		return nil, fmt.Errorf("bridge from cidr: %w", err)
 	}
 
+	if err := netinst.EnsureIptablesSymlinks(ctx); err != nil {
+		return nil, fmt.Errorf("ensure iptables symlinks: %w", err)
+	}
+
 	// NB: this is needed for the Dagger shim worker at the moment for host alias
 	// resolution
 	err = netinst.InstallResolvconf(netName, bridge.String())
@@ -913,7 +917,7 @@ func setupNetwork(ctx context.Context, netName, netCIDR string) (*networkConfig,
 		return nil, fmt.Errorf("install dnsmasq: %w", err)
 	}
 
-	cniConfigPath, err := netinst.InstallCNIConfig(netName, netCIDR)
+	cniConfigPath, err := netinst.InstallCNIConfig(ctx, netName, netCIDR)
 	if err != nil {
 		return nil, fmt.Errorf("install cni: %w", err)
 	}
--- a/internal/buildkit/util/network/cniprovider/bridge.go
+++ b/internal/buildkit/util/network/cniprovider/bridge.go
@@ -9,7 +9,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
 
 	cni "github.com/containerd/go-cni"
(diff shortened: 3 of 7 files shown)


sipsma:
Okay did some more digging, another even more fundamental problem is that the code was modifying `internal/buildkit/util/network/cniprovider/bridge.go`, but we don't use that bridge setup.
* It's leftover from buildkit, but should probably consolidate this type of thing to reduce this type of confusion

The CNI config we actually use is in `network/netinst`.

Trying some modifications to that CNI config to see if the perf impact is any different (though I'm gonna guess no).

---

EDIT:

Yeah the problems go a bit further:
1. `iptables-nft` does not install the `nft` command, which means that setting the `ipMasqBackend` to `nftables` will never work anyways. It was always falling back to `iptables` ultimately, but it fixed the problem on hosts w/out legacy xtables because `iptables-nft` binaries translate legacy->nft.
   * However, it seems like that translation layer must be slow, which explains `TestModule` taking an extra 1m30s
2. Then I tried instead installing both `nftables` and `iptables-legacy` packages (they don't overlap or conflict in terms of symlinks)
   * When hardcoding `network/netinst/cni.go` to force `"ipMasqBackend": "nftables"`, `TestModule` took **6m42s**
      * so forcing `nftables` on my host (6.12 Linux) still introduces a perf regression, but not as bad
   * When hardcoding `network/netinst/cni.go` to force `"ipMasqBackend": "iptables"`, `TestModule` took **5m50s**
      * so forcing `iptables` on my host has no perf regression

So I think what we should do is:
1. Install both `nftables` and `iptables-legacy`
2. Use the same idea as is already present where `xtables` is preferred if available, only using `nftables` if that's the only option
   * except applied to `network/netinst/cni.go` this time

I am a bit concerned about the future where everyone has updated to `nftables` and whether the perf hit I'm seeing is still going to be present. But that's down the line and the perf hit seems to only be noticeable in "thousands of containers" type use cases.

Overall it's probably worth an audit of our whole CNI setup sometime. It hasn't been touched in quite a while so it's possible there's room for perf improvement generally speaking anyways.

sipsma:
@marcosnils tested again and there's a new even more fun problem: separate from the `bridge` plugin, `firewall` is hardcoded to use `iptables` subcommands (it also has a `firewalld` backend, but that's not in the cards for our use case).

So *that* requires we have `iptables-nft` (compat layer from `iptables`->`nftables`) installed, but that will bring us back to the perf regression on hosts that are not using `nft` 😵‍💫

I think our only option is to detect our state at engine startup at then symlink the commands correctly based on what's available: use legacy `iptables` if available since it's still the fastest, use the compat layer binaries only when `nftables` is the only one available.

Will revisit again tomorrow or next week, hit timebox on this for today.

ReillyBrogan:
This is probably why the kubernetes project installs a helper binary to call out to the right one

sipsma:
> This is probably why the kubernetes project installs a helper binary to call out to the right one

Correct, I saw the link you left earlier, appreciate it. I think we can get away with something a bit simpler because we don't have to handle as many cases, but spiritually it'll be the same idea.

marcosnils:
image


:works:

PR #11735, opened 2026-01-23
ci: make test binaries faster to build

sipsma:
We were running all tests with -race, which means that we don't hit earlier go build cache and is generally quite slow to build. It also is not incredibly valuable in that it applies to the client side of integ tests, which isn't the main point of the integ tests.

You could see this in telemetry because there was often a large gap between the start of the withExec for `go test ...` and the actual start of spans for tests.
* Some of them are as much as ~1m30s ([example](https://dagger.cloud/dagger/checks/github.com/dagger/dagger@996e343d1e1967042449d53b6b260170e9a59218?check=testSplit%3AtestContainer&listen=748b71a8d1d3ea2d&showHidden=548e5c4a839f73a4&showHidden=491fa7adcbbca294#7d71c75a53033197)), seems in particular when we hit an engine with fresh cache

This was exacerbated by the fact that we defaulted to running go test against `./...`, which re-built and re-ran a number of test packages that never actually got executed.

This change updates us to only run against `./...` and with `-race` in `testBase`. Every other suite doesn't use `-race` and is package specific.

Also snuck in a fix to break `TestCgroups` (which is archaic) to TestProvision and TestTelemetry as separate checks.


toolchains/test-split/engine-tests.gen.dang
```diff
@@ -1,20 +1,25 @@
 type TestSplit {
-		let testSpecific(run: [String!]!, ebpfProgs: [String!]): Void {
-				engineDev().test(race: true, run: run.join("|"), ebpfProgs: ebpfProgs)
+		let testSpecific(run: [String!]!, ebpfProgs: [String!], pkg: String, race: Boolean): Void {
+				engineDev().test(race: race, run: run.join("|"), ebpfProgs: ebpfProgs, pkg: pkg ?? "./core/integration")
```
vito: nit: You can set default values in args, and make them required on the receiving side which is handy for cases like this:

```suggestion
		let testSpecific(run: [String!]!, ebpfProgs: [String!], pkg: String! = "./core/integration", race: Boolean): Void {
				engineDev().test(race: race, run: run.join("|"), ebpfProgs: ebpfProgs, pkg: pkg)
```

On the calling side, the type for the `pkg` arg is optional (`String`).

toolchains/test-split/engine-tests.gen.dang
```diff
@@ -1,20 +1,25 @@
...
 		}
 
-		let testSkip(skip: [String!]!, ebpfProgs: [String!]): Void {
-				engineDev().test(race: true, skip: skip.join("|"), ebpfProgs: ebpfProgs)
+		let testSkip(skip: [String!]!, ebpfProgs: [String!], pkg: String, race: Boolean): Void {
+				engineDev().test(race: race, skip: skip.join("|"), ebpfProgs: ebpfProgs, pkg: pkg ?? "./core/integration")
```
vito:
```suggestion
		let testSkip(skip: [String!]!, ebpfProgs: [String!], pkg: String! = "./core/integration", race: Boolean): Void {
				engineDev().test(race: race, skip: skip.join("|"), ebpfProgs: ebpfProgs, pkg: pkg)
```

PR #11736, opened 2026-01-23
chore: bump python sdk dep for protobuf cve

sipsma:
Fixes scan failures https://dagger.cloud/dagger/checks/github.com/dagger/dagger@48433f58938d323ff6b72a2ea3e02abd562d4293?check=security%3AscanSource&listen=c5e4b1ad40f3aa3c&listen=3f34620b160e3548#3f34620b160e3548:L180

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -1,0 +1,1 @@
+CVE-2026-0994 # not relevant to us and not fixed yet
--- a/toolchains/security/main.dang
+++ b/toolchains/security/main.dang
@@ -10,6 +10,13 @@ type Security {
     withWorkdir("/home/trivy")
 
   """
+  The base trivy container
+  """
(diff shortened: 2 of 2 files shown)


sipsma:
scanner is still failing even though I bumped the dep to the latest version... looking deeper

EDIT: actually, it seems like there might not be a fix to this package yet... I don't think we're probably impacted by the vuln so I'm tempted to add to an ignore list, but will wait a lil bit to see if they release a fix soon.
* https://pypi.org/project/protobuf/#history
* https://github.com/protocolbuffers/protobuf/issues/25070

sipsma:
I'm just ignoring this CVE, no obvious progress upstream and it doesn't appear to be relevant to us as we don't directly use protobuf in the python SDK.

PR #11737, opened 2026-01-23
core/schema: add private _rebasedIncludePaths to moduleSource

tiborvass:
New private moduleSource._rebasedIncludePaths returns the list of include paths from dagger.json, relative to the context directory.

### DIFF
--- a/core/schema/modulesource.go
+++ b/core/schema/modulesource.go
@@ -114,6 +114,12 @@ func (s *moduleSourceSchema) Install(dag *dagql.Server) {
 		dagql.Func("originalSubpath", s.moduleSourceOriginalSubpath).
 			Doc(`The original subpath used when instantiating this module source, relative to the context directory.`),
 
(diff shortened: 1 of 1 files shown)


tiborvass:
I made it public, but renamed it to `rebasedIncludePaths` (like the struct field). Let me know if you preferred `rebasedIncludes` or something else.

grouville:
> rebasedIncludePaths

Personally, I don't understand this name (non blocking)

tiborvass:
@grouville that's why i want to make it private :D I didn't want to spend time debating on the name, although I agree it's not super clear.

An alternative is to choose between `includePaths`, `includePatterns`, `includes` and have this one be relative to the context directory. Then if we ever need the include patterns as they were in the dagger.json we could always add an `original`-prefix.

tiborvass:
I reverted the PR to a private API for easier merge, as I really need this.

tiborvass:
Failure seems unrelated

PR #11748, opened 2026-01-27
filesync: fix .gitignore handling

sipsma:
Mark gitignored entries in stats instead of excluding them from the walk, and have the server skip deletes/updates under those paths.

This avoids cross-client conflicts caused by confusing overlapping requirements of having `.gitignore` in the synced fs tree but also needing to exclude it sometimes.

Fixes CI flake about `change kind changed from "add" to "delete" during sync` and variations
* repro'd with `dagger call engine-dev test --parallel=12 --pkg ./core/integration --run='TestHost/TestDirectoryGitIgnore' --count=10`

---

Also added a reference doc about filesync to the cache-expert skill.

The code here was typed by codex. Required a lot of socratic-type guiding to get it to the right solution but still saved quite a bit time in debugging and then typing out the fix 🦾


sipsma:
This seems to change the timing of things (if I had to guess, actually a noticeable improvement in the module related tests that do tons of loading of dirs full of .gitignore'd things), which is triggering many occurrences of issues fixed by https://github.com/dagger/dagger/pull/11729, so I currently have commits from that PR here. We should merge that other one first in order to avoid a big wave of flakes.

PR #11753, opened 2026-01-28
🐞 only load toolchains in parent module

kpenfound:
This was causing an engine panic when you install a module which has toolchains.

The change is to only load toolchains on the parent module.

### DIFF
--- a/core/schema/modulesource.go
+++ b/core/schema/modulesource.go
@@ -3705,6 +3705,7 @@ func (s *moduleSourceSchema) loadDependencyModules(ctx context.Context, src dagq
 	if err != nil {
 		return nil, fmt.Errorf("failed to get default dependencies: %w", err)
 	}
+
 	deps := core.NewModDeps(query, defaultDeps.Mods)
 	for _, depMod := range depMods {
 		deps = deps.Append(depMod.Self())
@@ -3727,6 +3728,7 @@ func (s *moduleSourceSchema) loadDependencyModules(ctx context.Context, src dagq
 
 		deps = deps.Append(clone)
 	}
+
 	for i, depMod := range deps.Mods {
 		if coreMod, ok := depMod.(*CoreMod); ok {
 			// this is needed so that a module's dependency on the core


marcosnils (dismissed):

core/schema/modulesource.go
```diff
@@ -3552,6 +3555,12 @@ func (s *moduleSourceSchema) moduleSourceAsModule(
 		return inst, fmt.Errorf("module name must be set")
 	}
 
+	// Mark this as the parent module if not already set (nil means not yet set)
+	// This ensures only the top-level module (not dependencies) loads toolchains
+	if ctx.Value(isParentModuleKey{}) == nil {
```
marcosnils, line 3560: I approved this but I'm wondering if this addition might be redundant with some information in the ClientMetadata which we might be able to use to check this instead of adding a new context value. Erik surely knows 👐

core/schema/modulesource.go
```diff
@@ -3655,22 +3664,32 @@ func (s *moduleSourceSchema) loadDependencyModules(ctx context.Context, src dagq
 		return nil, fmt.Errorf("failed to get dag server: %w", err)
 	}
 
+	// Check if this is the parent module
+	isParentModule := ctx.Value(isParentModuleKey{}) == true
+
+	// Create a context without the parent module flag for loading dependencies and toolchains
+	// This ensures nested modules don't load their own toolchains
+	depCtx := context.WithValue(ctx, isParentModuleKey{}, false)
```
sipsma, line 3672: Does this mean that you can get different results in the cache depending on whether a module was loaded as the top-level one or not?

If so I don't think it works. Once one version of a module is loaded it's what will be cached and others will get that regardless of the ctx key.

If something impacts the cached result of a call it needs to be mixed in the cache key for the call.

More broadly, I don't understand why this restriction would be in place. Why can't you have a dep on a module that uses toolchains? Can we make that work?

core/schema/modulesource.go
kpenfound: Yeah if there's an existing way to determine this let me know

core/schema/modulesource.go
kpenfound: > Why can't you have a dep on a module that uses toolchains? Can we make that work?

This is trying to make it work! But still keeping the distinction between dependencies and toolchains. So we have module A with a dependency on module B. Module B has toolchains. Right now module A will panic trying to load its dependencies. Since module A's dependency should only be able to access the API of module B itself, there's no reason to load B's toolchains

As far as the cache key is concerned, that sounds a bit tricker. Its overkill to add something like "isParent" to the cache key just for this edge case.

I guess I can revisit the original panic and see if there's another way to attack it. Preventing the toolchains from being unnecessarily loaded seemed like the most direct approach

sipsma: > Since module A's dependency should only be able to access the API of module B itself

That's what's tripping me up, if A has a dep on B and B has toolchains, why wouldn't A be able to call the toolchain functions in B? I don't understand on an intuitive level.

In my mind toolchains extend the API of a module, so B's API includes its own functions plus those of its toolchains. Therefore, anyone with a dep on B is gonna see B's functions and B's toolchain's functions. That just seems like the most obvious behavior and it's also what would result in coherent cache behavior (which is always a good sign something is actually coherent).

By "can we make it work" I basically meant make it follow that model above ^

kpenfound: this fundamentally changes with https://gist.github.com/shykes/e4778dc5ec17c9a8bbd3120f5c21ce73 so we can probably punt on this.

> In my mind toolchains extend the API of a module, so B's API includes its own functions plus those of its toolchains

I don't think of it this way today so that's probably where the disconnect is. Initially we decided we might need a way to opt in to including toolchains in your API when someone else installs your module, but that wouldn't be the behavior by default. That way if I had a module like the dang-sdk I could have toolchains that help me develop it but those would not be included in my API. The fact that toolchains currently extend the API is really just to make them more easily usable by the CLI

grouville:
@kpenfound what's the status on this one ?

kpenfound:
Closing this as it'll be resolved by modules v2

PR #11760, opened 2026-01-29
Update Helm chart testing link in Kubernetes docs

AsocPro:
I was reading the docs and noticed this link was getting a 404 so I tracked down the updated location.


shykes (approved):
Nice catch, thank you!

PR #11767, opened 2026-01-29
continue untethering from LLB by removing PBDefinitions

sipsma:
We are rapidly approaching zero dependencies on non-dag-op LLB, which is the point where we'll finally be able to delete the buildkit solver and simplify caching enormously.

One of the biggest remaining uses of LLB left was telemetry, which needed it to track lazy operations through the cause/effect system.

This PR updates the cause/effect system to instead rely on tracking effects through `call.ID`. This is sufficient to delete all the PBDefinitions methods.
* There's one slight exception to this, which is that we still rely on the `load cache:` span from the buildkit solver to determine if a lazy operation was cached or not. This will go away once the solver is actually removed, but in the meantime we're still making progress in untethering from LLB in the codebase.

---

One nice side effect: we're seeing `CACHED` status in more spans than before. The spans in question were cached previously too, but due to the extreme complication of relying on the buildkit solver they weren't getting marked as `CACHED`. The simplifications here seem to have unintentionally helped there.

---

There's one failure left in `TestGolden`, but I think the new output actually makes more sense than the old output. Will get @vito's opinion before adjusting the code though.


dagql/idtui/viztest/main.go
```diff
@@ -631,7 +631,17 @@ func (*Viztest) CallBubblingDep(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+	//unlazy one of them to verify it shows up as cached
+	f := files[0]
+	_, err = f.Sync(ctx)
+	if err != nil {
+		return err
+	}
+
```
vito: How did it look without this change? Were they all 'pending'? Did that make sense before?

This particular test is hardly using the golden example TBH, it has a custom assertion where it's mainly inspecting specific span attributes that get added:

https://github.com/dagger/dagger/blob/327e647482837b875cf1b466ad8b9e1f1683b0e2/dagql/idtui/golden_test.go#L182-L204

Only mentioning because it might not be worth changing this, whatever golden file came out is probably fine for the purposes of the original test.

sipsma: Yeah it had them all as pending. TBH I mostly changed this because I wanted to confirm my understanding of what was happening. But then it felt like a reasonable thing to test anyways since it's covering what was covered before but also that if something is unlazied it will be cached as expected.

core/file.go
```diff
@@ -124,6 +103,29 @@ func NewFileWithContents(
 	return dir.File(ctx, name)
 }
 
+func NewFileWithContentsDagOp(
```
alexcb, line 106: I've seen this before 😅

alexcb: #11733 can be merged before, or after, doesn't really matter.

sipsma: Ah yeah I needed to dag-op-ify the file constructor to get progress looking fully correct, went with this for now to not expand the scope of this PR too much. The other usages of `NewFileWithContents` are one remaining llb thing to cleanup

PR #11768, opened 2026-01-29
go generated clients get their own go.mod

kpenfound:
This gives generated clients in go their own go.mod.

Doing this means we don't compete with the project's own go.mod at all

It includes backwards compat for existing generated clients so that we don't break anyone's working setup.

We are also directly handling the integration test's need to replace the local SDK (read: library) rather than making that part of client generation. In the integration tests we handle the replace for the local library. Because the generated clients have their own go.mod, the replace happens in the generated client instead of the parent go.mod.

Additionally, when we create a new generated client, we will require+replace it in the parent go.mod so that its ready to use. We don't run `go mod tidy` here though because if they're not using the library yet (basically the only possible case) it will get auto removed.

### DIFF
--- a/cmd/codegen/generator/go/generate_client.go
+++ b/cmd/codegen/generator/go/generate_client.go
@@ -29,24 +29,28 @@ func (g *GoGenerator) GenerateClient(ctx context.Context, schema *introspection.
 	}
 
 	if !exist {
-		// If no go.mod is found, we will generate a go.mod
-		goMod := new(modfile.File)
-		goMod.AddModuleStmt(strings.ToLower(g.Config.ClientConfig.ModuleName))
-		goMod.AddGoStmt(goVersion)
+		// Only create a root go.mod for legacy/standalone client mode (ClientDir == ".")
+		// For separate client modules, we don't create a parent go.mod - just the client's
+		if g.Config.ClientConfig.ClientDir == "." {
+			goMod := new(modfile.File)
+			goMod.AddModuleStmt(strings.ToLower(g.Config.ClientConfig.ModuleName))
+			goMod.AddGoStmt(goVersion)
 
-		modBody, err := goMod.Format()
-		if err != nil {
-			return nil, fmt.Errorf("failed to format go.mod: %w", err)
+			modBody, err := goMod.Format()
+			if err != nil {
+				return nil, fmt.Errorf("failed to format go.mod: %w", err)
+			}
+
+			if err := mfs.WriteFile("go.mod", modBody, 0600); err != nil {
+				return nil, fmt.Errorf("failed to write go.mod: %w", err)
+			}
+
+			return &generator.GeneratedState{
+				Overlay:        layerfs.New(mfs),
+				NeedRegenerate: true,
+			}, nil
 		}
-
-		if err := mfs.WriteFile("go.mod", modBody, 0600); err != nil {
-			return nil, fmt.Errorf("failed to write go.mod: %w", err)
-		}
-
-		return &generator.GeneratedState{
-			Overlay:        layerfs.New(mfs),
-			NeedRegenerate: true,
-		}, nil
+		// For separate client modules without a parent go.mod, continue with nil goModFile
 	}
 
 	packageImport := filepath.Join(
@@ -54,8 +58,8 @@ func (g *GoGenerator) GenerateClient(ctx context.Context, schema *introspection.
 		g.Config.ClientConfig.ClientDir,
 	)
 
-	// respect existing package import path if a package is set
-	if goModFile.Module != nil {
+	// respect existing package import path if a parent go.mod exists
+	if goModFile != nil && goModFile.Module != nil {
 		// Find parent go.mod directory and relative path to client
 		_, clientRelPath, err := g.findParentGoModDir()
 		if err != nil {
@@ -136,6 +140,7 @@ func (g *GoGenerator) GenerateClient(ctx context.Context, schema *introspection.
 
 	// Post commands
 	var postCmds []*exec.Cmd
+	clientAbsPath := filepath.Join(g.Config.OutputDir, g.Config.ClientConfig.ClientDir)
 
 	// On install (first time), update parent's go.mod with require + replace
 	if isInstall && goModFile != nil {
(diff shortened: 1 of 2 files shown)


cmd/codegen/generator/go/generate_client.go
```diff
@@ -51,99 +49,193 @@ func (g *GoGenerator) GenerateClient(ctx context.Context, schema *introspection.
...
-			return replacedPath, true, nil
+		clientRelPath, err := filepath.Rel(parentGoModDir, clientAbsPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get relative path: %w", err)
 		}
+
+		// Add require and replace to parent's go.mod for the client module
+		// Use shell command with cd because codegen.go will overwrite cmd.Dir
+		parentEditCmd := exec.Command("sh", "-c",
+			fmt.Sprintf("cd %s && go mod edit -require=%s@v0.0.0 && go mod edit -replace=%s=./%s",
```
TomChv: You could simply update the parent go.mod by calling the gomod method and then rewrite it, so you don't have to bother with that shell cmd:

```shell
mod.AddRequire("<pkg name>", "v0.0.0")
mod.AddReplace("<pkg name>", "v0.0.0", "<new path>", "v0.0.0")
```

cmd/codegen/generator/go/generate_client.go
```diff
@@ -51,99 +49,193 @@ func (g *GoGenerator) GenerateClient(ctx context.Context, schema *introspection.
...
 		g.Config.ClientConfig.ClientDir,
 	)
 
 	// respect existing package import path if a package is set
 	if goModFile.Module != nil {
-		packageImport = filepath.Join(goModFile.Module.Mod.Path, g.Config.ClientConfig.ClientDir)
+		// Calculate the client directory relative to the parent module root
+		clientAbsPath := filepath.Join(g.Config.OutputDir, g.Config.ClientConfig.ClientDir)
+
+		// Find the parent module directory
```
TomChv: Normally the output dir is always the root of the current dagger module, which should be at the same level of the go project go.mod.

With your current logic, you'll lookup until you find a go.mod in the context dir, is that the expected behaviour?

cmd/codegen/generator/go/generate_client.go
```diff
@@ -51,99 +49,193 @@ func (g *GoGenerator) GenerateClient(ctx context.Context, schema *introspection.
...
+			for {
+				if _, err := os.Stat(filepath.Join(dir, "go.mod")); err == nil {
+					parentGoModDir = dir
+					break
+				}
+				parentDir := filepath.Dir(dir)
+				if parentDir == dir {
+					break
+				}
+				dir = parentDir
```
TomChv: You already looked for the parent go.mod before, can't you just get save the parent dir during the previous stage?

cmd/codegen/generator/go/generate_client.go
TomChv, line 1: I see that you look multiple time for the parent go.mod and its path, maybe put that logic in a utility function?

core/integration/client_generator_test.go
```diff
@@ -60,10 +61,15 @@ func main() {
...
   fmt.Println("result:", res)
-}`, defaultGenDir))
+}`))
 				},
 				postSetup: func(ctr *dagger.Container) *dagger.Container {
-					return ctr
+					// Add SDK replace to client's go.mod for testing
+					// This is test-specific - production code doesn't need this
+					return ctr.
+						WithExec([]string{"sh", "-c", fmt.Sprintf("cd %s && go mod edit -replace dagger.io/dagger=./sdk", defaultGenDir)}).
```
TomChv: This is duplicated over all integration tests, you could put that in the goSetup?

TomChv (commented):
Great work! I think this PR just need some code cleanup but the logic is great 🚀

core/integration/client_generator_test.go
```diff
@@ -102,8 +108,14 @@ main()`, defaultGenDir))
 					With(nonNestedDevEngine(c)).
 					With(daggerNonNestedExec("init")).
 					With(tc.setup).
-					With(daggerClientInstall(tc.generator)).
-					With(tc.postSetup)
+					With(daggerClientInstall(tc.generator))
+
+				// Mount SDK for go generator tests so SDK replace directives work
```
TomChv, line 106: So that means the dev engine should still pull the latest dagger.io/dagger version except if there's a replace?

cmd/codegen/generator/go/generate_client.go
kpenfound: > the current dagger module, which should be at the same level of the go project go.mod.

Not necessarily. In fact one of the test cases tests that we have a dagger module in a subdirectory of a project which has a go.mod at the root. But I think this code doesn't actually find-up outside of the dagger.json anyway even though it looks like it should

core/integration/client_generator_test.go
kpenfound: This is trying to accomplish what we talked about the other day where dagger/dagger's "special" case of needing to use the local library for development/testing is codified in the integration tests rather than in the generator itself. So yes, if somewhere else in the dev engine creates a generated client and doesn't add a replace, it will be latest dagger.io/dagger.

cmd/codegen/generator/go/generate_client.go
kpenfound: Maybe, 2 questions about that
1. Is that actually better? We'd have to parse, require, replace, and write
2. The benefit of using exec is we can pass it back to cmd/codegen as a PostCommand to execute after the client is overlayed, which we wouldn't be able to do with modfile

TomChv: > Is that actually better? We'd have to parse, require, replace, and write

We are already parsing the parent go mod so it shouldn't impact performances, the only extra step would be to write the parent go.mod which was already happening before no?

 > The benefit of using exec is we can pass it back to cmd/codegen as a PostCommand to execute after the client is overlayed, which we wouldn't be able to do with modfile

Is there a benefit to do it as a PostCommand after the overlay?

cmd/codegen/generator/go/generate_client.go
TomChv: Yeah I think we only copy all `go.mod` in the module's context directory.
The codegen binary is then run from the module root dir so it can access upper dir *but* the current logic use the `outputDir` as root path and this one is relative (usually `.`).

cmd/codegen/generator/go/generate_client.go
kpenfound: > We are already parsing the parent go mod

I don't think we are?

Oh, I see, `goModFile, exist, err := g.readGoMod()` is the parent go.mod, I missed that. I think that means there's some logic here I don't actually want. Revisiting this part

kpenfound:
ready for another look @TomChv . In addition to the feedback you had, we now also pin the SDK to the engine version in the dagger.json (unless it calls for a dev version, then its up to the user).

cmd/codegen/generator/go/generate_client.go
```diff
@@ -31,119 +29,228 @@ func (g *GoGenerator) GenerateClient(ctx context.Context, schema *introspection.
...
+			updatedParentGoMod, err := goModFile.Format()
+			if err != nil {
+				return nil, fmt.Errorf("failed to format parent go.mod: %w", err)
+			}
+
+			if err := mfs.WriteFile("go.mod", updatedParentGoMod, 0600); err != nil {
+				return nil, fmt.Errorf("failed to write updated parent go.mod: %w", err)
+			}
+		}
+		// If parent go.mod is outside OutputDir, we can't update it via overlay
```
TomChv, line 170: Good point, ideally we can detect it from the codegen bin and display a warning in the TUI but I'm not sure it's doable for now.

kpenfound: Yeah agreed, it would be great to have some way to bubble up warnings (cc future @vito )

PR #11769, opened 2026-01-29
Make dagger connect faster

tiborvass:
Skip Available() call for drivers, getting rid of `docker version` altogether.
If Provision() or Connect() fails, try the next driver.
Reuse Info() call from Wait() loop and rename Wait() to WaitInfo()
Delay setting up buildkit session to before Run.

This approximately halves the connect time for me on a docker+ssh context.

### DIFF
--- a/engine/client/drivers/driver.go
+++ b/engine/client/drivers/driver.go
@@ -12,6 +12,7 @@ import (
 
 type Driver interface {
 	// Available returns true if the driver backend is running and available for use.
+	//
 	// Deprecated: only used by deprecated GetDriver()
 	Available(ctx context.Context) (bool, error)
 


sipsma:
The failures in test-provision, test-container and elsewhere all seem legit, like the cli is unable to connect to the engine

internal/buildkit/client/info.go
```diff
@@ -23,15 +26,33 @@ type SystemInfo struct {
...
+				BuildkitVersion: fromAPIBuildkitVersion(res.BuildkitVersion),
+				SystemInfo:      fromAPISystemInfo(res.SystemInfo),
+			}, nil
+		}
+
+		switch code := grpcerrors.Code(err); code {
+		case codes.Unavailable:
+		case codes.Unimplemented:
+			// only buildkit v0.11+ supports the info api, but an unimplemented
+			// response error is still a response so we can ignore it
```
grouville: Curious how we avoid the panic that `engine/client/newBuildkitClient` would hit here: it still seems to assume `WaitInfo` returns a real struct and immediately reads `info.BuildkitVersion.Package`

If the CLI connects to a runner whose BuildKit predates 0.11 (we didn’t ship ≥0.11 until
`298f21afb` right before `v0.3.8`), this now returns `(nil, nil)` and _we’d dereference it before we can tell the user their engine is too old_ ?

Am I missing some other guard? I agree it's very unlikely as it would require a user with `_EXPERIMENTAL_DAGGER_RUNNER_HOST` / `WithRunnerHost` with this client and pointing at such version ?

tiborvass: ~~If there's an oversight in this PR that introduces a panic compared to main, happy to fix it, but I'm not sure I understood where that is. If you're talking about a preexisting oversight, I'd rather fix it in a followup.~~ Nvm, I understood, indeed it is this PR introducing a bug. I'll fix it.

PR #11770, opened 2026-01-30
engine-dev: replace alpine with wolfi

tiborvass:
I've had trouble running `dagger call generate` and `dagger call engine-dev introspection-json as-json contents` with an amd64 engine due to possible Alpine issues.

```
load http(url: "https://dl-cdn.alpinelinux.org/alpine/edge/main/x86_64/libcrypto3-3.5.4-r0.apk", refID: "qbyaq3p936rlfmt7shw4nbi3z"): invalid response status 404 Not Found
```

I'm not exactly sure what happened: https://dagger.cloud/tiborvass/traces/22fa61bd3737d64e844dad167bd733c5?listen=5f98acc260801256&listen=0a627df8013f68cd&listen=f86b388a509b75ae&listen=8ec8afe6b07dcb10#4cb90657d7dd6fad

but using wolfi solves the problem.

### DIFF
--- a/toolchains/engine-dev/build/builder.go
+++ b/toolchains/engine-dev/build/builder.go
@@ -339,11 +339,10 @@ func (build *Builder) verifyPlatform(ctx context.Context, bin *dagger.File) erro
 	}
 	mntPath := filepath.Join("/mnt", name)
 	out, err := dag.
-		Alpine(dagger.AlpineOpts{
-			Branch:   consts.AlpineVersion,
+		Wolfi().
+		Container(dagger.WolfiContainerOpts{
 			Packages: []string{"file"},
 		}).
-		Container().
 		WithMountedFile(mntPath, bin).
 		WithExec([]string{"file", mntPath}).
 		Stdout(ctx)
--- a/toolchains/engine-dev/build/sdk.go
+++ b/toolchains/engine-dev/build/sdk.go
@@ -252,9 +252,7 @@ func (build *Builder) goSDKContent(ctx context.Context) (*sdkContent, error) {
 
 func unpackTar(tarball *dagger.File) *dagger.Directory {
 	return dag.
-		Alpine(dagger.AlpineOpts{
-			Branch: consts.AlpineVersion,
-		}).
+		Wolfi().
 		Container().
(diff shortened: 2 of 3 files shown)


modules/wolfi/main.go
```diff
@@ -35,7 +35,7 @@ func (w *Wolfi) Container(
 		ExtraRepositories: extraRepositories,
 		ExtraKeyUrls:      extraKeyURLs,
 	})
-	ctr := config.Container()
+	ctr := config.Container().WithDirectory("/usr/local/bin", dag.Directory())
```
sipsma: Why is this necessary?

tiborvass: I guess i could add it in a comment, but it's necessary because `/usr/local/bin` does not exist in the wolfi container and we were assuming it exists and it was failing on `ln -s $daggerCli /usr/local/bin/dagger`

sipsma: I think that means you should just create /usr/local/bin in the returned container where you want it to exist (i.e. in a caller module). I don't see why we should make it exist for every wolfi container if it's not part of what wolfi installs.

tiborvass: @sipsma the only reason i put it in Wolfi() is so that it's easier to simply change from Alpine() to Wolfi() since Alpine() does have an empty `/usr/local/bin` but Wolfi doesn't. Having said that, if you still maintain that it should be the caller, I'm happy to change it.

sipsma: Yeah I think it should be the caller. Alpine != Wolfi, so it is perfectly fine for /usr/local/bin to exist in one but not the other.

The reason is that assembling rootfs'es scares me 😆 It's so insanely delicate that adding an extra dir in the base rootfs before other things strikes me as the type of thing that'll be fine for a while but then 2y later cause some insanely subtle+obscure issue that will result in hours of debugging until we find this line hidden here, if that makes sense.

tiborvass:
@sipsma i updated the PR to remove the `/usr/local/bin` directory creation in the Wolfi module, and instead replaced the WithExec(ln -s) in the caller with WithSymlink() which has the added benefit of automatically creating the intermediate directories.

PR #11771, opened 2026-01-30
core: add check.Source() back

tiborvass:
It was inadvertently removed in the refactor in #11685

Original PR: https://github.com/dagger/dagger/pull/11630

Please first merge https://github.com/dagger/dagger/pull/11770


core/checks.go
```diff
@@ -11,9 +11,10 @@ import (
 
 // Check represents a validation check with its result
 type Check struct {
-	Node      *ModTreeNode `json:"node"`
-	Completed bool         `field:"true" doc:"Whether the check completed"`
-	Passed    bool         `field:"true" doc:"Whether the check passed"`
+	Node      *ModTreeNode  `json:"node"`
+	Completed bool          `field:"true" doc:"Whether the check completed"`
+	Passed    bool          `field:"true" doc:"Whether the check passed"`
+	Source    *ModuleSource `field:"true" doc:"The module source where the check is defined (i.e., toolchains)"`
```
eunomie: Should it be defined inside `Node *ModTreeNode` instead?
And exposed using a method, like we do for `name` for instance, that will return the value from `Node`.
If I'm right, `check.Node.Module.GetSource()` should give the module source, right?

eunomie: I'm thinking about that also because I'm implementing generators on top of ModTreeNode, so trying to see what should be in the tree node and what should be in the final type.

tiborvass: The PR is still not working as intended, and it's possible I need to do exactly that.

PR #11774, opened 2026-01-31
ui: add boundary span wrapping user telemetry

vito:
Trying a theory: we can identify which spans are 'user spans' by wrapping all points where we exec something, since that will cover both module function calls _and_ any telemetry produced by a `withExec` (e.g. test spans).


vito:
Closing for now, will resurrect when it's relevant again

PR #11779, opened 2026-02-02
✨ dagger generate

eunomie:
Add `dagger generate` feature.

Functions returning a `Changeset` can be decorated and will be listed as `generator` function and executed via  `dagger generate` command.

Generators can be used to construct assets, generate code that should be versioned, etc.

## Define generator function

Annotate a function with the `generator` pragma:

- `+generator` in go
- `@generator()` in typescript
- `@generator` in python

The decorated functions must return a `Changeset`.

## Apply changes

All the resulting changesets will be merged automatically before to be applied at once. If conflicts occur between the changesets, the command will fail. (this behavior might be configurable in the future)

As for any function returning a `Changeset`, pass `-y`/`--auto-apply` to automatically apply the changes without user confirmation.

### List available generators

```console
$ dagger generate --list
✔ connect 0.2s
✔ load . 31.0s
✔ fetch generator information 0.4s

Name                      Description
changelog:generate        Generate the changelog with 'changie merge'. Only run this manually, at release time.
docs:generate             Regenerate the API schema and CLI reference docs
engine-dev:generate       Generate any engine-related files
go-sdk:generate           Regenerate the Go SDK API
php-sdk:generate          Regenerate the PHP SDK API
python-sdk:generate       Regenerate the core Python client library
rust-sdk:generate         Regenerate the Rust SDK API client.
typescript-sdk:generate   Generate the Typescript client library
```

### Generate everything

```console
$ dagger generate
```

### Generate by name, toolchain name, etc

```console
$ dagger generate python-sdk:generate
✔ pythonSdk:generate 1m36s 5×⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡇⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡄⡇⣿⣿⡇
```

### DIFF
--- a/cmd/dagger/generators.go
+++ b/cmd/dagger/generators.go
@@ -17,15 +17,11 @@ import (
 )
 
 var (
-	generateListMode       bool
-	enableGenerateScaleOut bool
+	generateListMode bool
 )
 
 func init() {
 	generateCmd.Flags().BoolVarP(&generateListMode, "list", "l", false, "List available generators")
-
-	generateCmd.Flags().BoolVar(&enableGenerateScaleOut, "scale-out", false, "Enable scale-out to cloud engines for each generate executed")
-	generateCmd.Flags().MarkHidden("scale-out")
 }
 
 var generateCmd = &cobra.Command{
@@ -44,7 +40,7 @@ Examples:
 		return withEngine(
 			cmd.Context(),
 			client.Params{
-				EnableCloudScaleOut: enableGenerateScaleOut,
+				EnableCloudScaleOut: enableScaleOut,
 			},
 			func(ctx context.Context, engineClient *client.Client) error {
 				dag := engineClient.Dagger()
--- a/cmd/dagger/checks.go
+++ b/cmd/dagger/checks.go
@@ -19,15 +19,11 @@ import (
 )
 
 var (
-	checksListMode       bool
-	enableChecksScaleOut bool
+	checksListMode bool
 )
 
 func init() {
 	checksCmd.Flags().BoolVarP(&checksListMode, "list", "l", false, "List available checks")
-
-	checksCmd.Flags().BoolVar(&enableChecksScaleOut, "scale-out", false, "Enable scale-out to cloud engines for each check executed")
-	checksCmd.Flags().MarkHidden("scale-out")
 }
 
 var checksCmd = &cobra.Command{
@@ -47,7 +43,7 @@ Examples:
 		return withEngine(
 			cmd.Context(),
 			client.Params{
-				EnableCloudScaleOut: enableChecksScaleOut,
+				EnableCloudScaleOut: enableScaleOut,
 			},
 			func(ctx context.Context, engineClient *client.Client) error {
 				dag := engineClient.Dagger()
--- a/cmd/dagger/main.go
+++ b/cmd/dagger/main.go
@@ -69,6 +69,7 @@ var (
 	web                      bool
 	noExit                   bool
 	_, useCloudEngine        = os.LookupEnv("DAGGER_CLOUD_ENGINE")
+	enableScaleOut           bool
 
 	dotOutputFilePath string
 	dotFocusField     string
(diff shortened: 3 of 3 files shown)


eunomie:
> I would prefer the pragma/decorator to be `+generate` and `@generate`

@shykes I changed the pragma/decorator to `generate`. I kept `generators` internally in various places and API, like `generators` on a module to retrieve all the functions marked as `generate`. Let me know if that's good or if I should also change them.

All tests are finally green, so this is good for me on this side :-)

cmd/dagger/generators.go
```diff
@@ -0,0 +1,152 @@
...
+
+var (
+	generateListMode       bool
+	enableGenerateScaleOut bool
+)
+
+func init() {
+	generateCmd.Flags().BoolVarP(&generateListMode, "list", "l", false, "List available generators")
+
+	generateCmd.Flags().BoolVar(&enableGenerateScaleOut, "scale-out", false, "Enable scale-out to cloud engines for each check executed")
```
vito: A quick one, submitting early in case it ends up being the only nit!

```suggestion
	generateCmd.Flags().BoolVar(&enableGenerateScaleOut, "scale-out", false, "Enable scale-out to cloud engines for each generate executed")
```

core/schema/generators.go
```diff
@@ -0,0 +1,128 @@
...
+func (s generatorsSchema) Install(srv *dagql.Server) {
+	dagql.Fields[*core.GeneratorGroup]{
+		dagql.Func("list", s.list).
+			Doc("Return a list of individual generators and their details"),
+
+		dagql.Func("run", s.run).
+			Doc("Execute all selected generators"),
+
+		dagql.NodeFunc("isEmpty", s.groupIsEmpty).
+			Doc("Wether changeset from the generator execution is empty or not").
```
vito:
```suggestion
			Doc("Whether the generated changeset is empty or not").
```

core/schema/generators.go
```diff
@@ -0,0 +1,128 @@
...
+			Args(
+				dagql.Arg("onConflict").Doc(`Strategy to apply on conflicts between generators`),
+			),
```
vito: copy-pasta?

edit: oh, I see this is a convenience API that passes it through. hmm, maybe we should just drop it? seems easy/obvious enough to just do `.changes.isEmpty`

cmd/dagger/generators.go
shykes: This can be done as a followup, but: shouldn't we make `--scale-out` a global flag? Eventually it will be removed I think, and inferred automatically I believe?

core/schema/generators.go
eunomie: Changing it, I'll just check all changes are empty or not.
Thanks for the suggestion

eunomie: done, I changed it, so it now checking all changesets one by one (and return false as soon as one is not empty). That way there's no merge of the changesets and no need for the arg

cmd/dagger/generators.go
eunomie: Done, I moved it as a global flag. Only used by check and generate for now, but it's there.
It's still hidden, like it was and like the cloud one.

shykes:
> @shykes I changed the pragma/decorator to `generate`. I kept `generators` internally in various places and API, like `generators` on a module to retrieve all the functions marked as `generate`. Let me know if that's good or if I should also change them.

Yeah it's fine. Let's talk about it again before shipping the 3d kind of verb ("ship"?). I don't think `Module.ships()` makes sense so we'll need to harmonize. But I don't want to hold this up.

PR #11791, opened 2026-02-03
Better defaults TUI

vito:
Before:

image

After:

image

Along the way: don't print `user default:` when `-s` is passed.

### DIFF
--- a/core/modfunc.go
+++ b/core/modfunc.go
@@ -245,7 +245,7 @@ func (fn *ModuleFunction) mergeUserDefaultsTypeDefs(ctx context.Context) error {
 				log.String(telemetry.DefaultArgAttr, argName),
 				log.String(telemetry.DefaultValueAttr, argDefault.UserInput),
 			),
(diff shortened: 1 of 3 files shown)


nipuna-perera:
Love it! Thank you! If it goes off screen vertically, can you focus/scroll it? For example, if there are a ton of user defaults? 

Oh and almost forgot. Please have a look at how it appears in `dots` and plain. Right now, it's printed a bunch.

vito:
@nipuna-perera Done:

image

image


> If it goes off screen vertically, can you focus/scroll it?

Not at the moment - let's cross that bridge when we get to it, since this was meant to be a quick side venture. :slightly_smiling_face:

nipuna-perera:
Aww! This didn't make it to 0.20.0? :( or is it not relevant after workspaces?

vito:
@nipuna-perera Not sure yet - the usage patterns will probably change but I think we'll still need some sort of UI work like this. Reopening for now

PR #11795, opened 2026-02-04
fix case where cache mounts become unmountable after restart

sipsma:
When the engine started, it was checking mutable mounts for any leftover volatile dirs. The way it did this resulted in the "shared" mount pool needed by overlay cache mounts being skipped. This also ended up being saved as the mounts to use for that ref going forward, which meant any concurrent use of the overlay cache mount would be invalid due to duplicate work/upper dirs and also fail due to volatile existing.

Fix is to just ensure the shared mount pool is correctly used in all cases.

### DIFF
--- a/engine/server/server.go
+++ b/engine/server/server.go
@@ -805,7 +805,6 @@ func (srv *Server) Close() error {
 	// the server should be shutdown first
 	srv.daggerSessionsMu.Lock()
 	daggerSessions := srv.daggerSessions
-	srv.daggerSessions = nil
 	srv.daggerSessionsMu.Unlock()
 
 	for _, s := range daggerSessions {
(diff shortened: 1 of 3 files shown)


engine/server/server.go
```diff
@@ -805,7 +805,6 @@ func (srv *Server) Close() error {
 	// the server should be shutdown first
 	srv.daggerSessionsMu.Lock()
 	daggerSessions := srv.daggerSessions
-	srv.daggerSessions = nil
```
sipsma, line 808: technically unrelated but while repro'ing the problem I saw a panic during engine shutdown due to the fact that we set this map nil.

vito (approved):

PR #11799, opened 2026-02-04
speed up module loading in large monorepos

sipsma:
Module loading was specifying an include that had a `**/*` appended, which appears to create a huge slowdown in monorepos with large directory trees in terms of number of files/dirs.

This specifically applies to a case like:
1. Big monorepo
2. A dagger module subdir that has relatively few files

Before this change, the `**/*` pattern seemed to be possibly be resulting in walks/checks of the full tree. I didn't take time yet to diagnose why exactly that happens and if it's expected or not.

Dropping the `**/*` fixes this entirely and brings module loading time down to a few seconds on empty cache and less than 1 second afterwards (for a simple go sdk module).

It's not entirely clear why the `**/*` was there; I have a vague memory of it being strictly necessary in ancient iterations of module loading but was probably cargo-culted since then even when not needed anymore.

### DIFF
--- a/core/schema/host.go
+++ b/core/schema/host.go
@@ -300,6 +300,12 @@ func (s *hostSchema) directory(ctx context.Context, host dagql.ObjectResult[*cor
 			continue
 		}
 		include = filepath.Join(relPathFromRoot, include)
+		if include == "." {
+			// we were told to include the directory itself, but include
+			// filters seem to ignore ".", so we replace it with the intention
+			// of "*"
+			include = "*"
+		}
 		if negative {
 			include = "!" + include
 		}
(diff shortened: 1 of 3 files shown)


sipsma:
CI seems to disagree that `**/*` was unnecessary, will move to draft to figure out what's different since it worked well in manual tests.

EDIT: the change in pattern is correct but it's triggering a corner case in the host.Directory implementation when gitignore=true and you are loading from the git root ([this condition](https://github.com/sipsma/dagger/blob/b3e240e4ba9fb7616379e7f50223157b00fdeff8/engine/filesync/filesyncer.go#L81-L81)). Will fix

PR #11803, opened 2026-02-05
feat(sdk/python): AST based analysis to introspect

eunomie:
This PR replaces the runtime analysis for the Python SDK to an AST based one.

Context: when the Python SDK wants to run a Python module, it starts by analysing the user module source code. This allows to find the types and functions exposed by the module, that is used in two different phases:
- to register those types to the engine (so the schema is extended by them, allowing them to be called)
- when a function is called so the python SDK knows the types to instanciate, functions to call, values to send back, etc

Right now, this happen at runtime, by loading the module code. The module code is loaded, and it will be analysed.

The problem is this analysis must happen _without_ to load the code. This is required to allow self calls. Self calls means the code must be able to call itself. There's a chicken and egg issue there, where the code must used generated code based... on the code.

To make it clearer, look at this self call code in Go (as it doesn't yet exist in python):

```go
type Test struct{}

func (m *Test) ContainerEcho(stringArg string) *dagger.Container {
}

func (m *Test) Print(ctx context.Context, stringArg string) (string, error) {
	return dag.Test().ContainerEcho(stringArg).Stdout(ctx)
}
```

In the `Print` function it calls `dag.Test()` (that is "self") and then the `ContainerEcho()` function.
To generate this binding, the SDK must know `ContainerEcho` is a thing. This means this code is valid only if it knows the function exposed.

In python, the equivalent code should be something like that:

```python
@object_type
class Test:
    @function

    @function
    def print(self, string_arg: str) -> str:
        return dag.test().container_echo(string_arg).stdout()
```

With the current python implementation, this will break as the code is loaded. But at the time the code is loaded, it doesn't know `dag.test()` is a thing.
We need to do:
- first analyse the code to find `test` object with `container_echo` and `print` functions
- generate the code exposed by the engine based on those types
- load the code when needed to call the functions

All that to say we _must_ analyse the code before it's loaded.

And to do that, we need to completely change the way the code is analysed, moving to an AST based analysis instead of loading the code and looking at it. That's what this PR is doing.

This is a big one. And one way to ensure this is good is to rely on the extensive test suite we have. As everything is green on this side, we could be quite confident on the result.

> [!NOTE]
> Next step will be to plug that into the `moduleTypes` of the SDK, preparing the work for self calls

### DIFF
--- a/sdk/python/src/dagger/mod/_analyzer/__init__.py
+++ b/sdk/python/src/dagger/mod/_analyzer/__init__.py
@@ -1,0 +1,40 @@
+"""AST-based type analysis for Dagger modules.
+
+This module provides static analysis of Python source code to extract
+type information without importing the module. This is critical when
+dependencies may not be available at analysis time.
+
+Usage:
+    from dagger.mod._analyzer import analyze_module
+
+    metadata = analyze_module(
+        source_files=["src/mymodule/__init__.py"],
+        main_object_name="MyModule",
+    )
+"""
+
+from dagger.mod._analyzer.analyze import analyze_module
+from dagger.mod._analyzer.metadata import (
+    EnumMemberMetadata,
+    EnumTypeMetadata,
+    FieldMetadata,
+    FunctionMetadata,
+    LocationMetadata,
+    ModuleMetadata,
+    ObjectTypeMetadata,
+    ParameterMetadata,
+    ResolvedType,
+)
+
+__all__ = [
+    "EnumMemberMetadata",
+    "EnumTypeMetadata",
+    "FieldMetadata",
+    "FunctionMetadata",
+    "LocationMetadata",
+    "ModuleMetadata",
+    "ObjectTypeMetadata",
+    "ParameterMetadata",
+    "ResolvedType",
+    "analyze_module",
+]
--- a/sdk/python/src/dagger/mod/_analyzer/analyze.py
+++ b/sdk/python/src/dagger/mod/_analyzer/analyze.py
@@ -1,0 +1,143 @@
+"""Main entry point for AST-based module analysis.
+
+This module provides the analyze_module() function that performs
+static analysis of Python source files to extract type information
+without importing the module.
+"""
+
+from __future__ import annotations
+
+import ast
+import logging
+from pathlib import Path
+
+from dagger.mod._analyzer.errors import AnalysisError, ValidationError
+from dagger.mod._analyzer.metadata import ModuleMetadata
+from dagger.mod._analyzer.parser import ModuleParser
+
+logger = logging.getLogger(__name__)
+
+
+def analyze_module(
+    source_files: list[str | Path],
+    main_object_name: str,
+    *,
+    module_name: str | None = None,
+) -> ModuleMetadata:
+    """Analyze Python source files and extract module metadata.
+
+    This is the main entry point for AST-based type analysis. It parses
+    the source files, extracts decorated classes and functions, resolves
+    type annotations, and returns structured metadata.
+    """
+    if not source_files:
+        msg = "No source files provided"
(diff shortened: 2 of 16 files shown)


TomChv (approved):

grouville (approved):
Hey @eunomie @TomChv 👋

While digging through the AST analyzer I ran into ~~two~~ three regressions that were easy to overlook but bite pretty hard once you see them:

1. Keyword-only defaults (the args that live after `*` and must be passed by name, like `def build(*, enabled: bool = False)`) were filtered with `{... if default}`, so any literal that evaluates falsey (`False, 0, "", etc.`) vanished and the metadata flipped those params to **required**, even though they had a default

You can see the test [here](https://github.com/dagger/dagger/pull/11803/changes/16e46a635cf05e3f17ea8c0885c72c16faf65d28)

2. Module discovery only looked at `.py` files sitting in the package root or one folder below it.

> That happened to match today’s layout, but the moment someone drops `pkg/features/nested/foo.py`, those objects disappear and the module doc can even come from that nested package because its `__init__.py` is the first one we see

The new `tests/mod/test_discovery.py` in [this commit](https://github.com/dagger/dagger/pull/11803/changes/531f445d0e87c9677ee50e4f790e312a60637b49) sets up exactly that nesting so you can watch it fail on the old code

3. The AST path also never looked for `@dagger.up`, so service functions lost their flag and registration never called `with_up()`. That’s why [hello-with-services-py](https://dagger.cloud/dagger/checks/github.com/dagger/dagger@cc9f55ec11174e8345235a6121b6ff63df8a90c8?check=test-split%3Atest-base&listen=5ae04265a53202da&listen=1901aee091191eb1&showHidden=0911348d1ce0c74f&showHidden=568e20d1c513ec34&showHidden=1bb9452e565ac094&showHidden=0911348d1ce0c74f&showHidden=568e20d1c513ec34&showHidden=1bb9452e565ac094#3c995cb759f17fb6) stopped showing any services in `dagger up --list`


In this current state, this LGTM ✅

grouville:
I'd wait for theseus to be merged though 🙏

tiborvass:
@eunomie  Just to understand better: this both should improve performance (how much?) and enable self-calls in python? (or that was already supported?)

eunomie:
> @eunomie Just to understand better: this both should improve performance (how much?) and enable self-calls in python? (or that was already supported?)

Yes. Self calls can't easily work without something like that. Regarding performances, I'll get some numbers (now that it's on top of egraph) but it's also just one part of more changes to come regarding how SDKs are analyzing the module code, generating types, etc.

eunomie:
> @eunomie Just to understand better: this both should improve performance (how much?)

I have numbers. And basically there's no different in performances. Right now.
But it will be the case after, when combined with other changes related to codegen.

tiborvass (approved):

tiborvass:
I found no conflicts with neither Workspaces nor Lockfile PR

vaz-ar:
Isn't this a breaking change ?

eunomie:
> Isn't this a breaking change ?

@vaz-ar I don't think. It should be fully transparent. The implementation changed, but the behaviour should be identical. That's why for instance none of the integration tests have been impacted.

Or did you spot some use cases where it broke?

vaz-ar:
Yes I had some issues yesterday when upgrading, someone had made a module that used inheritance to split the code across multiple files, and most of the functions where not accessible anymore.
But that might be that this was not a good pattern to use and it was working because of this "bug" 😄

eunomie:
@vaz-ar Could you please open a new discussion on the topic https://github.com/dagger/dagger/discussions/categories/help ?
That way it will be easier to investigate and keep track of it. If you have an example or available traces that would be really helpful.

PR #11826, opened 2026-02-09
feat: correctly pin the go pkg in module generation

TomChv:
Update the  `go_sdk` to send a pinned lib version to a commit from `dagger-go-sdk`.

⚠️ This change the release process, now we will need to update the const in `go_sdk` to the latest commit of `dagger-go-sdk`, this commit will not be the one tagged because it's done before the release.
At some point, it would be much better to release the SDKs before the engine so we can set the commit of the released instead of the one before.

### DIFF
--- a/core/sdk/go_sdk.go
+++ b/core/sdk/go_sdk.go
@@ -28,8 +28,8 @@ const (
 	// Set to a commit on https://github.com/dagger/dagger-go-sdk if an unreleased
 	// change is needed in the generated library.
 	// Otherwise, update it to the latest known commit during release.
-	// CURRENT commit: v0.20.1
-	goSDKLibVersion = "8ac3501b479993c6e8289fa9022d42edd50bbf0e"
+	// CURRENT commit: v0.20.3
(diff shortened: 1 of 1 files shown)


sipsma:
@TomChv the errors in ci:bootstrap seem related to your change

TomChv:
> the errors in ci:bootstrap seem related to your change

Yeah definitely, I found a fix, still digging into the other errors

TomChv:
> To double check everything, we should do a test "local-only" release and verify it works. So use the release module/toolchains to build a CLI + Engine that think they are actually v0.19.12 (or whatever next version is) and then do tests to make sure we can use existing modules, init new ones, call everything etc. Just to avoid a situation like last release

Done in a separated PR: https://github.com/dagger/dagger/pull/11941

TomChv:
@sipsma PR rebased on top of #11941 and ready for another review :D

TomChv:
@sipsma Ready for review :D

TomChv:
I rebased the PR and updated the lib version to the latest commit on dagger-go-sdk.

I'll merge once the CI is green

PR #11834, opened 2026-02-10
feat: expose generators to env / current-env

eunomie:
Expose `generators` and `generator` functions to environments, allowing to access all the generate functions.

### DIFF
--- a/docs/docs-graphql/schema.graphqls
+++ b/docs/docs-graphql/schema.graphqls
@@ -2273,6 +2273,20 @@ type Env {
     include: [String!]
   ): CheckGroup! @experimental(reason: "Checks API is highly experimental and may be removed or replaced entirely.")
 
+  """
+  Return the generator with the given name from the installed modules. Must match exactly one generator.
+  """
+  generator(
+    """The name of the generator to retrieve"""
+    name: String!
+  ): Generator! @experimental(reason: "Generators API is highly experimental and may be removed or replaced entirely.")
+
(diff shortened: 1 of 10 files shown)


core/schema/env.go
```diff
@@ -88,6 +88,18 @@ func (s environmentSchema) Install(srv *dagql.Server) {
...
+			Doc("Return all generators defined by the installed modules").
+			Args(
+				dagql.Arg("include").Doc("Only include generators matching the specified patterns"),
+			),
+		dagql.Func("generator", s.envGenerator).
+			Experimental("Generators API is highly experimental and may be removed or replaced entirely.").
+			Doc("Return the generator with the given name from the installed modules. Must match exactly one generator.").
+			Args(
+				dagql.Arg("name").Doc("The name of the generator to retrieve"),
+			),
```
eunomie, line 102: @shykes: I'm thinking about your remark regarding `generator` and `generate`.
And I wonder what would be best. I think it still time to change them if we want.

My first thinking was to have `dagger generate` as the command, then `generator` and `generators` the names for the components that `dagger generate` will call.

And it lets the ability (that we don't have with checks) to also have a `generate` function exposed there that will retrieve _and_ execute the generators. Like the `dagger generate` but from the API.

eunomie:
I'm closing this PR, to avoid duplication of effort with the new workspace API that will handle access to checks, generate and other kind of artifacts instead of current-env.

PR #11835, opened 2026-02-10
engine: fix +ignore pragma for constructors

### LINKED ISSUE #11750: 🐞 The `+ignore` pragma is not always respected during module construction
### What is the issue?

Follow-up of Discord discussion: https://discord.com/channels/707636530424053791/1465717553530671389

I have a context directory, which looks like (simplified)

```go
    // +defaultPath="/"
    // +ignore=["**","!dagger.json"]
    source *dagger.Directory,
```

When calling a module without any arguments, the `source` is resolved as a "context" directory and the correct filters (`exclude`) are applied:

```
$ dagger call source-contents

dagger.json
```

However, if a `--source` is provided, it is no longer recognized as a "context" directory and the entire directory is uploaded during filesync:

```
$ dagger call --source /tmp/dagger-filesync source-contents

.gitattributes
.gitignore
dagger.gen.go
dagger.json
go.mod
go.sum
internal/
main.go
```

Notes
- This only occurs when the context directory is in the `New` constructor
- It is strange that the explicit `--source` is not recognized as a context dir

### Dagger version

dagger v0.19.10 (image://registry.dagger.io/engine:v0.19.10) darwin/arm64/v8

### Steps to reproduce

```go
package main

import (
    "context"
    "dagger/filesync/internal/dagger"
)

type Filesync struct {
    Source *dagger.Directory // +private
}

func New(
    // +defaultPath="/"
    // +ignore=["**","!dagger.json"]
    source *dagger.Directory,
) *Filesync {
    return &Filesync{Source: source}
}

func (m *Filesync) SourceContents(ctx context.Context) ([]string, error) {
    return m.Source.Entries(ctx)
}
```

Called with

```
dagger call --source /tmp/dagger-filesync source-contents

.gitattributes
.gitignore
dagger.gen.go
dagger.json
go.mod
go.sum
internal/
main.go
```

### Log output

_No response_

matipan:
Fixes #11750.

---

The issue is caused by a check on `setCallInputs`:
```go
if len(arg.metadata.Ignore) > 0 && !arg.metadata.isContextual() { // contextual args already have ignore applied
```

In this context an argument is considered contextual if either the `DefaultPath` or `DefaultAddress` is set. If it is then `setCallInputs` will not go through the ignore process of directories. The problem: arguments that are contextual (they have a default path) but a user specified a value will **not** be loaded as part of the contextual loading and will also be ignored in this check. This means `applyIgnoreOnDir` will never run for them.

Looking into the code I found two potential fixes:
1. Remove the `isContextual` check and apply ignore on all arguments. Great for simplicity, but it means we are going to run the `applyIgnoreOnDir` functions on truly contextual arguments twice.
2. Mark contextual arguments only when they are 100% coming from the context directory and not from user inputs.

I went with the second option even though it required more code changes. I'm not 100% sure on the performance impact of running `applyIgnoreOnDir` twice for contextual args, but it felt wasteful.

As for the test: I created a separate test even though we already have a `TestIgnore`. This was mainly to keep the existing test clean given that it tests a wide range of cases but for function arguments only. Happy to merge them and look into potential improvements for them.

Tested the change with a dev engine and a module with the following code:
```go
// ...
func New(
	// +defaultPath="."
	// +ignore=["**", "!dagger.json"]
	source *dagger.Directory,
) *NewMod {
	return &NewMod{
		Source: source,
	}
}

func (m *NewMod) TestSource(
	ctx context.Context,
) (string, error) {
	return dag.Container().
		From("alpine:latest").
		WithDirectory("/source", m.Source).
		WithExec([]string{"ls", "-lh", "/source"}).
		Stdout(ctx)
}

```

Specifying the arg using a dev engine vs v0.19.11:

Not specifying the arg (using the context):

---

### AI disclosure

We don't really do or require this in the project, but given the buzz in social media I thought it would be worth disclosing the use of AI for this PR.

I used it as a guide to look into the code and it helped me understand how arguments are considered contextual, at which point they are loaded and ignored and what is the difference between constructor and function args. Once we found the bug, the code changes were also AI assisted but not vibe coded.

### DIFF
--- a/core/integration/module_test.go
+++ b/core/integration/module_test.go
@@ -5619,6 +5619,51 @@ func (t *Test) IgnoreDirButKeepFileInSubdir(
 	})
 }
 
+func (ModuleSuite) TestIgnoreConstructor(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	modGen := goGitBase(t, c).
+		WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+		WithWorkdir("/work").
+		With(daggerExec("init", "--name=test", "--sdk=go", "--source=.")).
+		With(sdkSource("go", `
+package main
+
+import (
+  "context"
+  "dagger/test/internal/dagger"
+)
+
+type Test struct {
+  Source *dagger.Directory
+}
+
+func New(
+  // +defaultPath="."
+  // +ignore=["**","!dagger.json"]
+  source *dagger.Directory,
+) *Test {
+  return &Test{Source: source}
+}
+
+func (t *Test) SourceDir(ctx context.Context) *dagger.Directory {
+  return t.Source
+}
+`))
+
+	t.Run("default (no explicit arg)", func(ctx context.Context, t *testctx.T) {
+		out, err := modGen.With(daggerCall("source-dir", "entries")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "dagger.json", strings.TrimSpace(out))
+	})
+
+	t.Run("explicit --source", func(ctx context.Context, t *testctx.T) {
+		out, err := modGen.With(daggerCall("--source", ".", "source-dir", "entries")).Stdout(ctx)
+		require.NoError(t, err)
+		require.Equal(t, "dagger.json", strings.TrimSpace(out))
+	})
+}
+
 func (ModuleSuite) TestGitignore(ctx context.Context, t *testctx.T) {
 	c := connect(ctx, t)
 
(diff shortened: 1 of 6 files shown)


sipsma (commented):
Are we 100% sure that this was specific to constructors? I can't see anything in the diff that is specific to constructors, but might have missed something.

Asking because if the behavior of "explicit arg doesn't use ignore" was *true* for normal functions but *false* for constructors, then I totally agree we should make them consistent.

However, if the behavior today is that "explicit arg doesn't use ignore" for all functions, then I actually think it's pretty ambiguous whether that's correct or not. I can see cases where passing an explicit argument in place of your context dir:
1. should NOT use the default ignores, i.e. I really want you to use the directory I give you
2. *should* use the default ignores, i.e. don't make me manually type all that out again just because I'm overriding a path

It almost seems like something that should be possible to specify somehow when you pass the arg?

cc @shykes @kpenfound DX related

PR #11837, opened 2026-02-10
Avoid WithExec panics and surface underlying exec errors

sipsma:
Before this change, Container.WithExec could panic with "index out of range" while handling exec failures. The failure-handling path assumed mount and output slices were always aligned, and a deferred block could run even when mount preparation had already failed.

e.g.
 ```
  panic: runtime error: index out of range [0] with length 0

  goroutine 4176537 [running]:
  github.com/dagger/dagger/core.(*Container).WithExec.func2()
          /app/core/container_exec.go:299 +0x1198
  github.com/dagger/dagger/core.(*Container).WithExec(0x2af28c0?, {0x3830a70, 0x40a809b810}, {{0x611c800, 0x0, 0x0}, 0x1, {0x0, 0x0},
  {0x0, ...}, ...}, ...)
          /app/core/container_exec.go:564 +0xa94
  github.com/dagger/dagger/core/schema.(*containerSchema).withExec(0x611c800, {0x3830a70, 0x40a809b810}, {{0x40fdb9d0a0, 0x0, 0x1, 0x0}, {0x0,
  0x1, 0x4129b168a0, ...}}, ...)
          /app/core/schema/container.go:1109 +0x568
  github.com/dagger/dagger/dagql.NodeFuncWithCacheKey[...].func1({{0x40fdb9d0a0, 0x0, 0x1, 0x0}, {0x0, 0x1, 0x4129b168a0, 0x415301c108,
  0x4152f8f560}}, 0x4074545f50, ...)
          /app/dagql/objects.go:702 +0x128
  github.com/dagger/dagger/dagql.ObjectResult[...].call.func2()
          /app/dagql/objects.go:549 +0xc0
  github.com/dagger/dagger/dagql.(*cache).GetOrInitialize.func3()
          /app/dagql/cache.go:767 +0x58
  created by github.com/dagger/dagger/dagql.(*cache).GetOrInitialize in goroutine 4176534
          /app/dagql/cache.go:765 +0x12f0

  ```

After this change, WithExec returns joined errors for index mismatches instead of panicking, and the defer only runs after mount preparation succeeds. This keeps the engine alive and preserves richer error context when exec handling goes off the expected path.

The root cause of these mismatches is still unclear, but this change avoids process panics and lets us see the underlying error messages.

### DIFF
--- a/core/container_exec.go
+++ b/core/container_exec.go
@@ -272,11 +272,18 @@ func (container *Container) WithExec(
 		desc := fmt.Sprintf("mount %s from exec %s", m.Dest, strings.Join(metaSpec.Args, " "))
 		return cache.New(ctx, ref, bkSessionGroup, bkcache.WithDescription(desc))
 	}, runtime.GOOS)
+	if err != nil {
+		return nil, err
+	}
 	defer func() {
 		if rerr != nil {
 			execInputs := make([]bksolver.Result, len(mounts.Mounts))
 			for i, m := range mounts.Mounts {
 				if m.Input == pb.Empty {
+					continue
+				}
+				if m.Input < 0 || int(m.Input) >= len(mounts.Inputs) {
+					rerr = errors.Join(rerr, fmt.Errorf("mount index %d references input %d outside available inputs (%d)", i, m.Input, len(mounts.Inputs)))
 					continue
 				}
 				execInputs[i] = mounts.Inputs[m.Input].Clone()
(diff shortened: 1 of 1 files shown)


grouville:
Hey @sipsma, I've been looking at this change and noticed a potential resource leak ?

Before this PR, when `PrepareMounts` failed, the defer was already installed, so partial refs accumulated in `p.Actives` and `p.OutputRefs` would still get released via the deferred cleanup (since `rerr` gets set on return).

After this PR, the error check moved above the defer:
```go
p, err := bkcontainer.PrepareMounts(...)
if err != nil {
    return nil, err  // defer not installed yet
}
defer func() { ... }()
```

So if `PrepareMounts` fails partway through its loop (after appending refs but before completing), those partial refs should leak, no ?

This matches BuildKit's own pattern in `NewContainer` (`container.go:96-104`) which explicitly releases partial refs on `PrepareMounts` error:
```go
if err != nil {
    for i := len(p.Actives) - 1; i >= 0; i-- {
        p.Actives[i].Ref.Release(context.TODO())
    }
    for _, o := range p.OutputRefs {
        o.Ref.Release(context.TODO())
    }
    return nil, err
}
```

I have a fix that wraps `PrepareMounts` to explicitly release partial refs on error. Did I miss something ?

sipsma:
@grouville Yeah we should fix that, but I'd prefer to fix it by making `PrepareMounts` handle that cleanup itself. Generally speaking a function should handle its own cleanup, the fact that it doesn't is what creates extremely confusing situations like this.

PR #11838, opened 2026-02-10
make dagql cache ID and Result aware

sipsma:
~~This is unrefined slop atm, but directionally correct. Just want to kick off some full CI checks for errors and/or perf impact. Will do human cleanup before bringing out of draft.~~
* EDIT: successfully humanized 💪

---

Goal is to make the base cache aware of more dagql primitives, specifically of IDs and Results (whereas previously cache was decoupled and generic).

This is another crucial step for replacing the buildkit solver and enables a lot of simplifications + new features.

This is one of those situations where more tightly *coupling* two components actually ends up making everything better, despite that usually being the inverse:
1. ID awareness means the cache can track the relationship between results, i.e. the actual DAG formed by inputs+outputs.
   * Before this that sort of tracking happened outside the cache, which basically scattered the complexity around the whole codebase.
   * This is required for robust cache pruning logic
2. Result awareness enables persistence and pruning of said Results, which is needed once everything that's a DagOp today becomes purely native dagql operations that are persisted
3. All the PostCall and client resource insanity can finally go away 🙌. All the information needed for that can be gleaned from the cache DAG after this.
4. Unit tests are gonna get WAY more useful, we can write those instead of integ tests in more situations
5. Cache observability, debugging and answering questions like "why didn't I get a cache hit" become much more obvious to implement
6. Remote caching becomes much more obvious too

To be clear, this PR is not all the above; it's purely just refactoring the cache interface to accept IDs as the cache key and return results. The diff here is big enough already, so wanna merge this first quick.

---

High-level implementation details for reviews:
1. filesync doesn't use shared cache package anymore, this diverges way too much from its use of the cache package. It has its own highly simplified "singleflight plus ref counting" implementation now
2. Cache interface is
   * `GetOrInitCall` - same idea as previous methods but accepts ID as input (part of cache key) and returns AnyResult
   * `GetOrInitArbitrary` - in some places we cache more arbitrary stuff not attached to a single call (e.g. `git ls-remote` results), you use this one for that now; saves those users from having to create a "fake id" as input
3. `dagql.Result` is used directly by the cache now and has methods like `HitCache()` and `Release()`. It is created "per-call" but has a field storing shared metadata between all Results for the same data. Basically, `Result` is a "ref" that points to the sharedResult metadata.
4. There's some references in comments to a few ugly looking things that will go away with the "equivalence set" approach, read below in the last section of this description for info on that

---

To give a general orientation of where this PR leads in terms of finishing the buildkit solver replacement, here's what's left after this PR:
1. Update the cache to use smarter data structures (namely a disjoint-set union-find type of thing, wrapped up as an [e-graph](https://en.wikipedia.org/wiki/E-graph)). This:
   * Maximizes our ability to find cache hits, which is trickier than it seems when there's multiple digests associated with an operation/result
   * Gives us good theoretical grounding, which makes correctness easier
   * In my experience so far, makes LLMs ***way*** less confused, probably since they have been trained on this sort of stuff
   * Simplifies other parts of the codebase since the API is just "associate extra digest with operation/result at any time"
5. Generalize the current function caching sqlite db to handle more general persistent caching (including pruning)
6. Rewrite dag-op wrappers to just be encoded IDs that are loaded with the dagql cache rather than "IDs encoded inside LLB that goes through the buildkit solver and then the dagql cache"

I've prototyped the majority of that already to get a lay of the land, so main work left is converting to mergeable PRs. We're very close 🎉

cc @alexcb @vito in case you're curious on the plans in this area

### DIFF
--- a/cmd/engine/metrics.go
+++ b/cmd/engine/metrics.go
@@ -24,31 +24,6 @@ var (
 		Help: "Number of entries in the dagql cache",
 	})
 
-	dagqlCacheOngoingCallsEntriesGauge = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "dagger_dagql_cache_ongoing_calls_entries",
-		Help: "Number of in-progress dagql call entries",
-	})
-
-	dagqlCacheCompletedCallsEntriesGauge = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "dagger_dagql_cache_completed_calls_entries",
-		Help: "Number of completed dagql call entries indexed by call key",
-	})
-
-	dagqlCacheCompletedCallsByContentEntriesGauge = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "dagger_dagql_cache_completed_calls_by_content_entries",
-		Help: "Number of completed dagql call entries indexed by content digest",
-	})
-
-	dagqlCacheOngoingArbitraryEntriesGauge = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "dagger_dagql_cache_ongoing_arbitrary_entries",
-		Help: "Number of in-progress arbitrary dagql cache entries",
-	})
-
-	dagqlCacheCompletedArbitraryEntriesGauge = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "dagger_dagql_cache_completed_arbitrary_entries",
-		Help: "Number of completed arbitrary dagql cache entries",
-	})
-
 	localCacheTotalDiskSizeGauge = prometheus.NewGauge(prometheus.GaugeOpts{
 		Name: "dagger_local_cache_total_disk_size_bytes",
 		Help: "Total disk space consumed by the local cache in bytes",
@@ -71,21 +46,6 @@ func setupMetricsServer(ctx context.Context, srv *server.Server, addr string) er
 		return err
 	}
 	if err := prometheus.Register(dagqlCacheEntriesGauge); err != nil {
-		return err
-	}
-	if err := prometheus.Register(dagqlCacheOngoingCallsEntriesGauge); err != nil {
-		return err
-	}
-	if err := prometheus.Register(dagqlCacheCompletedCallsEntriesGauge); err != nil {
-		return err
-	}
-	if err := prometheus.Register(dagqlCacheCompletedCallsByContentEntriesGauge); err != nil {
-		return err
-	}
-	if err := prometheus.Register(dagqlCacheOngoingArbitraryEntriesGauge); err != nil {
-		return err
-	}
-	if err := prometheus.Register(dagqlCacheCompletedArbitraryEntriesGauge); err != nil {
 		return err
 	}
 	if err := prometheus.Register(localCacheTotalDiskSizeGauge); err != nil {
@@ -140,12 +100,6 @@ func setupMetricsServer(ctx context.Context, srv *server.Server, addr string) er
 	http.HandleFunc("/metrics", func(w http.ResponseWriter, r *http.Request) {
 		connectedClientsGauge.Set(float64(srv.ConnectedClients()))
 		dagqlCacheEntriesGauge.Set(float64(srv.DagqlCacheEntries()))
-		cacheStats := srv.DagqlCacheEntryStats()
-		dagqlCacheOngoingCallsEntriesGauge.Set(float64(cacheStats.OngoingCalls))
-		dagqlCacheCompletedCallsEntriesGauge.Set(float64(cacheStats.CompletedCalls))
-		dagqlCacheCompletedCallsByContentEntriesGauge.Set(float64(cacheStats.CompletedCallsByContent))
-		dagqlCacheOngoingArbitraryEntriesGauge.Set(float64(cacheStats.OngoingArbitrary))
-		dagqlCacheCompletedArbitraryEntriesGauge.Set(float64(cacheStats.CompletedArbitrary))
 
 		var dbReset float64
 		if srv.CorruptDBReset() {
--- a/core/integration/engine_test.go
+++ b/core/integration/engine_test.go
@@ -730,15 +730,10 @@ func (EngineSuite) TestPrometheusMetrics(ctx context.Context, t *testctx.T) {
 
 		// find the lines with metrics we care about testing
 		soughtMetrics := map[string]struct{}{
-			"dagger_connected_clients":                              {},
-			"dagger_dagql_cache_entries":                            {},
-			"dagger_dagql_cache_ongoing_calls_entries":              {},
-			"dagger_dagql_cache_completed_calls_entries":            {},
-			"dagger_dagql_cache_completed_calls_by_content_entries": {},
-			"dagger_dagql_cache_ongoing_arbitrary_entries":          {},
-			"dagger_dagql_cache_completed_arbitrary_entries":        {},
-			"dagger_local_cache_total_disk_size_bytes":              {},
-			"dagger_local_cache_entries":                            {},
+			"dagger_connected_clients":                 {},
+			"dagger_dagql_cache_entries":               {},
+			"dagger_local_cache_total_disk_size_bytes": {},
+			"dagger_local_cache_entries":               {},
 		}
 		foundMetrics := map[string]int{}
 		for _, line := range strings.Split(out, "\n") {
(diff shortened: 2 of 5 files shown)


dagql/session_cache.go
```diff
@@ -146,50 +110,58 @@ func (c *SessionCache) GetOrInitializeWithCallbacks(
...
 	// The call above used a random storage key (due to DoNotCache=true), so the result
 	// won't be found by future lookups using the original key.
+	// NOTE: this complication can be removed once we are off the buildkit solver as errors
+	// won't be cached for the duration of a session.
 	if forcedDoNotCache {
 		key.DoNotCache = false
-		cachedRes, cacheErr := c.cache.GetOrInitializeValue(ctx, key, res.Result())
+		cachedRes, cacheErr := c.cache.GetOrInitCall(ctx, key, ValueFunc(res))
 		if cacheErr == nil {
 			res = cachedRes
```
grouville, line 164: Do we agree that in the _forced retry + reinsert path_, `SessionCache` should still return real `nil` for `null` results, even if cache can return a non-nil `Result` object with `hasValue=false` ?

_Example_
1. First execution of operation K returns an error
2. Second execution of operation K returns nil, nil (valid nullable result)
3. `SessionCache` then reinserts that result into normal cache under K
4. If `SessionCache` does not re-check null after reinsert, it can return that wrapper instead of literal nil ?

I wonder if this could lead to unexpected panic afterward ?

grouville: A small repro that seems to panic: 

```go
func TestSessionCacheErrorThenNilReturnsNilAfterForcedRetry(t *testing.T) {
	t.Parallel()
	ctx := t.Context()
	cacheIface, err := NewCache(ctx, "")
	assert.NilError(t, err)
	sc := NewSessionCache(cacheIface.(*cache))
	srv := NewServer(cacheTestQuery{}, sc)

	var callCount atomic.Int32
	Fields[cacheTestQuery]{
		Func("flakyNull", func(context.Context, cacheTestQuery, struct{}) (Nullable[String], error) {
			switch callCount.Add(1) {
			case 1:
				return Nullable[String]{}, fmt.Errorf("boom")
			default:
				return Null[String](), nil
			}
		}),
	}.Install(srv)

	var out *string
	err = srv.Select(ctx, srv.Root(), &out, Selector{Field: "flakyNull"})
	assert.ErrorContains(t, err, "boom")
	assert.Equal(t, int32(1), callCount.Load())

	out = nil
	err = srv.Select(ctx, srv.Root(), &out, Selector{Field: "flakyNull"})
	assert.NilError(t, err)
	assert.Assert(t, out == nil)
	assert.Equal(t, int32(2), callCount.Load())

	out = nil
	err = srv.Select(ctx, srv.Root(), &out, Selector{Field: "flakyNull"})
	assert.NilError(t, err)
	assert.Assert(t, out == nil)
	assert.Equal(t, int32(2), callCount.Load())
}
```

output: 
```shell
--- FAIL: TestSessionCacheErrorThenNilReturnsNilAfterForcedRetry (0.00s)
  panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
  goroutine 7 [running]:
  testing.tRunner.func1.2({0x10cbee0, 0x1d1a3c0})
  testing.tRunner.func1()
        /home/guillaume/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/testing/testing.go:1737 +0x35e
  panic({0x10cbee0?, 0x1d1a3c0?})
        /home/guillaume/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/runtime/panic.go:792 +0x132
  github.com/dagger/dagger/dagql.Result[...].Type(0xc00007c558?)
        /home/guillaume/dagger/dagql/cache.go:377 +0x42
        /home/guillaume/dagger/dagql/server.go:759 +0x308
  github.com/dagger/dagger/dagql.TestSessionCacheErrorThenNilReturnsNilAfterForcedRetry(0xc0001c2700)
        /home/guillaume/dagger/dagql/session_cache_test.go:176 +0x337
  testing.tRunner(0xc0001c2700, 0x12bd5a0)
        /home/guillaume/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/testing/testing.go:1792 +0xf4
  created by testing.(*T).Run in goroutine 1
        /home/guillaume/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.4.linux-amd64/src/testing/testing.go:1851 +0x413
  FAIL  github.com/dagger/dagger/dagql  0.010s
  FAIL
```

cmd/engine/metrics.go
```diff
@@ -19,6 +19,36 @@ var (
...
+	})
+
+	dagqlCacheOngoingCallsEntriesGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "dagger_dagql_cache_ongoing_calls_entries",
+		Help: "Number of in-progress dagql call entries",
+	})
+
+	dagqlCacheCompletedCallsEntriesGauge = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "dagger_dagql_cache_completed_calls_entries",
+		Help: "Number of completed dagql call entries indexed by call key",
```
vito: What is a "call key" - like ID digest? Will this be super high cardinality and maybe blow up downstream metrics consumers?

Same q for `dagger_dagql_cache_completed_calls_by_content_entries` below

core/integration/engine_test.go
```diff
@@ -799,6 +835,167 @@ func (EngineSuite) TestPrometheusMetrics(ctx context.Context, t *testctx.T) {
...
+
+# Load module + dependency schema a few times to exercise cache lifecycle.
+for i in $(seq 1 4); do
+  dagger functions >/dev/null
+done
+			`}).Sync(ctx)
+	require.NoError(t, err)
+
+	// Give any async session/cache-release paths a moment to run before polling.
+	time.Sleep(2 * time.Second)
```
vito: Hoping this won't be flaky - could consider adding an endpoint to force GC but that may go against the spirit of the test (it'd pass even if we somehow accidentally disable auto GC)

core/schema/container.go
```diff
@@ -823,10 +823,13 @@ func (s *containerSchema) fromCacheKey(
 	}
 
 	resp := &dagql.GetCacheConfigResponse{CacheKey: req.CacheKey}
-	resp.CacheKey.CallKey = hashutil.HashStrings(
+	if resp.CacheKey.ID == nil {
+		return nil, errors.New("cache key ID is nil")
```
vito, line 827: Will this happen in practice? Or is it (maybe LLM-added) defensive?

core/modfunc.go
```diff
@@ -540,19 +544,18 @@ func (fn *ModuleFunction) CacheConfigForCall(
 		// Process "contextual arguments", aka objects with a `defaulPath`
 		ctxArgVals := make([]*argInput, len(ctxArgs))
 		for i, arg := range ctxArgs {
+			i, arg := i, arg
```
vito: nit: no longer needed right?

dagql/cache.go
```diff
@@ -0,0 +1,1221 @@
...
+}
+
+type cache struct {
+	mu sync.Mutex
+
+	// calls that are in progress, keyed by a combination of the call key and the concurrency key
+	// two calls with the same call+concurrency key will be "single-flighted" (only one will actually run)
+	ongoingCalls map[callConcurrencyKeys]*sharedResult
+
+	// calls that have completed successfully and are cached, keyed by the storage key
```
vito, line 177: What is a 'storage key'?

dagql/cache.go
```diff
@@ -0,0 +1,1221 @@
...
+}
+
+func (r arbitraryResult) Release(ctx context.Context) error {
+	if r.shared == nil {
+		return nil
+	}
+	return r.shared.release(ctx)
+}
+
+// TODO: Drop detached-result cloning once the cache uses an equivalence graph.
```
vito, line 299: Looking forward to the egraph stuff, sounds neat

dagql/cache.go
```diff
@@ -0,0 +1,1221 @@
...
+	if key.ID == nil {
+		return nil, fmt.Errorf("cache key ID is nil")
+	}
+
+	if key.DoNotCache {
+		// don't cache, don't dedupe calls, just call it
+
+		// we currently still have to appease the buildkit cache key machinery underlying function calls,
+		// so make sure it gets a random storage key
+		ctx = ctxWithStorageKey(ctx, rand.Text())
```
vito, line 785: Is this still true? (Just saw it carried over from the move)

dagql/session_cache.go
sipsma: Will fix!

cmd/engine/metrics.go
sipsma: Yeah, this was added for an integ test that makes an isolated engine, does a bunch of work (function calls), and then closes the client and verifies there's no cache leaks. On the first try there actually was a bug causing a leak, this metric was useful for figuring out what it was. But I don't think it's needed now for the integ test. I'll rm it or maybe leave it commented out for break-glass

core/integration/engine_test.go
sipsma: This is leftover and not needed anymore since we do polling after it anyways, will rm

core/schema/container.go
sipsma: No, codex just loves to be defensive about this sort of stuff. It's not *technically* wrong to check it though, so I just let it add these without fussing about it.

dagql/cache.go
sipsma: It's not a new concept, has existed since function caching was added, there's explanatory comments in a few places including here: https://github.com/sipsma/dagger/blob/b83bcf240d80697bfab9c2a5d6022b0086863e08/dagql/cache.go#L831-L834

Also doc'd in the `cache-expert` skill so LLMs should be able to explain it too.

It's basically a weird quirk we have to deal with because buildkit cache still exists and is separate from dagql cache. It should go away once we only have one caching system in a few PRs from now.

dagql/cache.go
sipsma: Yep, in the plan I put in the last section of the PR description, this will go away in the last step (when the buildkit solver+cache is gone)

alexcb (approved):
very cool to see the `string` to `call.ID` change, that's a lot of glue code changes!

There's a lot that changed here, but I didn't see anything obvious jump out -- best to wait for a few more +1 on the existing comments.

PR #11842, opened 2026-02-10
fix(container): make `PrepareMounts` clean up its own refs on error

grouville:
Follow-up of https://github.com/dagger/dagger/pull/11837

`PrepareMounts` allocates refs in a loop and can fail partway through. Before, callers were responsible for cleaning up partial refs on error. This was fragile: a caller refactor could silently introduce leaks.

Now `PrepareMounts` cleans up after itself:
- on success, caller owns the refs
- on failure, `PrepareMounts` releases any refs it created

Includes a regression test for the partial-failure path.


sipsma (approved):

sipsma:
Thanks for noticing and for the fix!

PR #11843, opened 2026-02-10
 chore: bump to go 1.26 

sipsma:
https://go.dev/doc/go1.26

Interested to see if there's anything noticeable in terms of performance.

Also added a `dagger-chores` skill to make this one-click (or prompt I suppose) in the future

### DIFF
--- a/toolchains/cli-dev/publish.go
+++ b/toolchains/cli-dev/publish.go
@@ -229,7 +229,7 @@ func (cli *CliDev) goreleaserBinaries() *dagger.Directory {
 	dir := dag.Directory()
 	for _, os := range oses {
 		for _, arch := range arches {
-			if arch == "arm" && os == "darwin" {
+			if arch == "arm" && (os == "darwin" || os == "windows") {
 				continue
 			}
 


sipsma:
Looks like we need to wait for wolfi to pick up the new version, moving to draft for now. Shouldn't be too long

EDIT: Still waiting, I'm testing periodically with `dagger call -m ./modules/wolfi container --packages go~1.26 terminal`

tiborvass:
I subscribed to https://github.com/wolfi-dev/os/pull/78619

xnox:
go-1.26 is available; but i'm not sure if anything else is needed on your side to switch to it.

tiborvass:
Opened https://github.com/wolfi-dev/os/issues/78640

xnox:
please use go-1.26, not go~1.26.

tiborvass:
@xnox Please correct me if i'm wrong, but this does not seem to be ideal for us to have to special case `go-1.26` when we use the fuzzy tilde `~` everywhere else for version matching. Why do you think `go~1.25` should resolve but not `go~1.26` ?

tiborvass:
I had to change `~` to `-` in the apk selector for the go toolchain. I also had to remove the windows/arm target that was removed in go1.26.

PR #11844, opened 2026-02-11
add `dagger trace` for viewing a trace

vito:
* retire `dagger watch` - stopped working ages ago I think?
* add `dagger trace <trace-id> [--org <org-name>]`

Currently, `trace` streams payloads via the existing Cloud GraphQL APIs, converts them back into OTLP data, and exports them back into the local frontend. Alternatively we could add OTLP fetching endpoints to Cloud, and I had a separate PR for that, the main advantage being that you don't need to specify an org, but this PR immediately works without having to bikeshed that and coordinate merging, so putting this up ASAP and we can swap later if/when we want by going back to the first commit.

### DIFF
--- a/cmd/dagger/trace.go
+++ b/cmd/dagger/trace.go
@@ -24,6 +24,7 @@ var traceCmd = &cobra.Command{
 	},
 	Aliases: []string{"t"},
 	Short:   "View a Dagger trace from Dagger Cloud.",
+	GroupID: cloudGroup.ID,
 	Example: `dagger trace 2f123ba77bf7bd2d4db2f70ed20613e8`,
 	RunE:    Trace,
 }


sipsma (approved):

cmd/dagger/trace.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/dagger/dagger/engine/slog"
+	cloud "github.com/dagger/dagger/internal/cloud"
+	"github.com/dagger/dagger/internal/cloud/auth"
+	"github.com/dagger/dagger/util/cleanups"
+	"github.com/spf13/cobra"
+)
+
+var traceOrgFlag string
+
+var traceCmd = &cobra.Command{
```
marcosnils, line 18: nit: should we add this under the cloudGroup.ID so it appears under "DAGGER CLOUD COMMANDS" in `--help`?

vito: Good call - the command is hidden at the moment, but added it to the group anyway so we don't forget in the future :+1:

PR #11847, opened 2026-02-11
cli: add logs frontend

vito:
even simpler than dots - just print logs, no dots or report


vito:
@grouville The context is LLMs - giving them a quieter output so they don't get distracted by the final report that `dots` prints. Since it's for an LLM, we don't need any of that other stuff and don't expect any interaction.

This is essentially a copy-paste of the `dots` frontend but trimmed down even more - the things you mentioned are also ignored by `dots`.

PR #11849, opened 2026-02-11
sdk(typescript): switch from `/* */` to `//` for comments

vito:
When we use a glob like `**/*.go` in doc strings or descriptions, that was causing these comments to become invalid syntax.

### DIFF
--- a/cmd/codegen/generator/typescript/templates/src/type_test.go
+++ b/cmd/codegen/generator/typescript/templates/src/type_test.go
@@ -30,7 +30,7 @@ func TestType(t *testing.T) {
 		require.NoError(t, err)
 		require.Equal(t, want, b.String())
 	})
-	
+
 	t.Run("scalar with glob comment", func(t *testing.T) {
 		wantFile := "testdata/type_test_scalar_comment_glob_want.ts"
 


vito:
> Also, with that new changes, the comments are no longer displayed when overing the class, functions or arguments. But maybe that's an IDE issue.

Ah that's a bummer - they rely on comment style? :sob: 

We can close this PR if that's the common case, didn't realize they were load-bearing.

One option is to only swap the style if it contains `*/`, but... yeesh.

TomChv:
> Ah that's a bummer - they rely on comment style? 😭

I mean, on Zed I can see that it's not displayed, I'll check tomorrow if maybe there's an extension to fix that.
I would expect the IDE to support both since the language allows it.

TomChv:
> Ah that's a bummer - they rely on comment style? 😭

Yeah sadly that's the convention :/

In TypeScript:
- `//` is for code comments
- `/* */` is for documentation, they'll be displayed by the IDE when hovered. 

I think the better option is to add a `\` to escape for example, this will allow the doc to be displayed correctly:

```
/**
 * Give all the path for example **\/*.go
 */
```

Screenshot 2026-02-16 at 18 50 07

It's a bit harder to implement but I think that's better than losing the documentation 😃

PR #11850, opened 2026-02-11
engine: harden BoltDB settings to avoid page-freed panics under NoSync

sipsma:
Before this change, several engine BoltDBs ran with `NoSync` only, and crashy/shutdown-edge cases could still surface bbolt corruption panics like `page ... already freed`.

After this change, every BoltDB in this tree that already uses `NoSync` also sets `NoFreelistSync` and `NoGrowSync`, aligning the option set across all DB open paths and matching the mitigation pattern used in containerd.

This updates all four relevant init points: snapshotter metadata store options, containerd metadata DB open, worker cache metadata store open, and solver cache store open. It also updates the `GracefulStop` comment to reflect the full option set.

---

Gonna give CI a ton of re-rolls to see if that panic pops up again (it was quite rare though, so won't be definitive).

### DIFF
--- a/.changes/unreleased/Fixed-20260211-182802.yaml
+++ b/.changes/unreleased/Fixed-20260211-182802.yaml
@@ -1,0 +1,7 @@
+kind: Fixed
+body: Update internal bbolt db configuration to do less syncing and possibly avoid
+  an upstream bug causing crashes
+time: 2026-02-11T18:28:02.858189627-08:00
+custom:
+  Author: sipsma
+  PR: "11850"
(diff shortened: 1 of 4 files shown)


sipsma:
Re-ran full CI 10x, no boltdb crashes https://dagger.cloud/dagger/checks/github.com/dagger/dagger@a4138a6589b6e293015c8f4dfd6efe2b9487dd23

Good initial sign, but won't know for sure until we don't see it for a while. Will go ahead and merge

Issue #8955, opened 2024-11-14
Second to last step is not cached


marcosnils:
# Problem

I have the following pipeline: 

```go
func (m *Test) Test() *dagger.Container {
	return dag.Container().From("alpine").
		WithExec([]string{"sleep", "5"})
}
```

If I run that multiple times, I'll get a cached  output as expected. Now, the moment that I add another `exec` **after** the last one, for some reason, the second to last exec gets re-executed and not computed from the cache. 

i.e: 

```go
// Returns a container that echoes whatever string argument is provided
func (m *Test) Test() *dagger.Container {
	return dag.Container().From("alpine").
		WithExec([]string{"sleep", "5"}).
		WithExec([]string{"sleep", "2"})
}
```
if I run this, the `sleep 5` exec will get re-executed. 

cc @jedevc @sipsma

marcosnils:
👋  I can't seem to repro the above issue anymore. However, there's a new variation of this issue which manifested after a user reported a cache invalidation confusion here: https://discord.com/channels/707636530424053791/1120503349599543376/1333424111799763039

 Here's the new code to reproduce: 

```golang
// Returns a container that echoes whatever string argument is provided
func (m *Second) Foo(context *dagger.Directory) *dagger.Container {
	return dag.Container().From("alpine:latest").
		WithMountedDirectory("/src", context).
		WithExec([]string{"sleep", "5"})
}

// Returns lines that match a pattern in the files of the provided Directory
func (m *Second) Bar(context *dagger.Directory) *dagger.Container {
	return m.Foo(context).WithExec([]string{"sleep", "2"})
}
```

If you run `dagger call  foo --context . stdout` and then `dagger call bar --context .  stdout`, you'll notice that the `sleep 5` will get re-executed. There's two catches which are also very odd: 

1. If you remove the `withMountedDirectory`, caching works as expected
2. If you add another step below `sleep 5` and run the same test as above, you'll notice that when calling `bar`, `sleep 5` will be cached now, but the last step of the `foo` function won't. 

Something 🐟  is happening here.

helderco:
> If you remove the `withMountedDirectory`, caching works as expected

This seems related to:
- https://github.com/dagger/dagger/issues/6421

jedevc:
Good call, that's definitely worth making sure of.

[...]

PR #11856, opened 2026-02-12
migrate all caching to dagql, remove buildkit solver

sipsma:
Status:

At first this PR was going to just be updating the cache to use a new e-graph implementation, which was setup for the final change that removed the buildkit solver and migrated all caching to dagql.

Then, during manual testing I hit a deadlock in the buildkit solver. I had codex spin on it overnight for 3 hours, it found part of the cause but not the full root cause.

It would be foolish to waste time on this, so I'm just expanding the scope of this PR to be the entire removal of the buildkit solver and replacement with dagql. The diff is gonna be gnarly, but I don't think anything else makes sense at this point.

TODOs:
- [x] e-graph
- [x] remove dag-op entirely (it was the bridge, but the bridge is no longer needed)
- [x] add generalized persistence to dagql cache
- [x] add pruning of persistent dagql cache entries
- [ ] delete buildkit solver related everything (lots of stuff that's unused is still technically initialized, need cleanup all over)
   * There's a couple of util functions left that we use which unfortunately end up entangling a massive amount of transitive buildkit packages. Once those last threads are cut it should all fall out
- [x] all integ tests passing
- [ ] update internal tech docs and this PR description with more explanation on how everything works

Overall, extremely happy with it. As things progressed, stuff fell into place super nicely and what used to be horrible conundrums that required hacks began having clear and obvious solutions
   - Lots of old cruft is gone already in this PR, with more to come!

### e-graph primer

It updates the cache implementation to use smarter data structures and algorithms, namely a union-find structure wrapped up as an [e-graph](https://en.wikipedia.org/wiki/E-graph). This:
* Maximizes our ability to find cache hits, which is trickier than it seems when there's multiple digests associated with an operation/result
* Gives us good theoretical grounding, which makes correctness+performance easier. This applies not only now but also down the line as we explore more optimizations and features
* In my experience so far, makes LLMs ***way*** less confused, probably since they have been trained on this sort of stuff
* Simplifies other parts of the codebase since the API will end up being "associate extra digest with operation/result at any time"

A lot of the material on e-graphs is very academic and dense, but like most things of that nature it's actually 100x less complicated than it sounds once you get passed the lingo. Part of the work here is to ensure the implementation here is as clear as possible to humans too.

In the meantime, a rough sketch of why e-graphs make sense to use here:
* In an e-graph you build up "terms" by defining operations where the result of one operation can be used as the input for another (typical DAG stuff)
* You also keep track of operations that are "equivalent" (the `e` in `e-graph`), where "equivalent" means "this result is interchangeable for another". The union-find data structure helps you do that efficiently

Outside of dagger, an example might be that "1+3" is equivalent to "2+2". Same output "content" but different "recipes" to get there.

In dagger, operations aren't arithmetic, they are our API (core + module functions). "Content" has varying meanings depending on the operation, but a common one is a digest of a resulting filesystem. There's many recipes that result in the same end content, which makes them equivalent.

There's more to it than just that, but that's the gist of why e-graphs match dagger's use case. It frames our problem well and gives us well established algorithms to do things like check for cache hits efficiently.

### other significant changes

* `CachePerClient`, `CachePerSession`, etc. used to be modeled by creating custom recipe digests where client-id/session-id/etc. was mixed in the digest.
   * Now those have been remodeled as "implicit inputs" to the call instead. This creates the same end effect of influencing the digest, but with the subtle-yet-crucial difference that these inputs show up in the DAG rather than opaquely influence it by changing a digest in an unobvious way
   * This makes a lot of things in the cache simpler (namely, recalculating digests after changes to the DAG) and should also be extremely helpful for understanding the cache too ("why is this not cached?" type questions)

---

Fixes https://github.com/dagger/dagger/issues/8955
* `TestContainer/TestWithMountedDirectoryCaching` started "failing" with `Congrats you fixed the bug; please enable the above test and delete this line` 🎉
* Didn't intend to fix this actually, just happened at some point after I stripped buildkit away


core/schema/module.go
```diff
@@ -71,6 +76,7 @@ func (s *moduleSchema) Install(dag *dagql.Server) {
 	dagql.Fields[*core.Module]{
 		// sync is used by external dependencies like daggerverse
 		Syncer[*core.Module]().
+			WithInput(dagql.CachePerClient).
```
sipsma: Why was this added? Is it load-bearing? If so, how?

core/schema/modulesource.go
```diff
@@ -105,6 +110,7 @@ func (s *moduleSourceSchema) Install(dag *dagql.Server) {
 	dagql.Fields[*core.ModuleSource]{
 		// sync is used by external dependencies like daggerverse
 		Syncer[*core.ModuleSource]().
+			WithInput(dagql.CachePerClient).
```
sipsma: Why was this added? Is it load-bearing? If so, how?

core/schema/modulesource.go
```diff
@@ -1050,7 +1059,7 @@ func (s *moduleSourceSchema) contextGitRepository(
 	if fDgst == "" {
 		fDgst = f.ID().Digest()
 	}
-	dgst := hashutil.HashStrings(fDgst.String(), "contextualGitRepository")
+	dgst := hashutil.HashStrings(fDgst.String(), args.Digest, "contextualGitRepository")
```
sipsma: including module digest here seems wrong, needs re-eval

core/schema/container.go
```diff
@@ -1022,11 +1049,40 @@ type containerExecArgs struct {
 	// calling it with args
 	SkipEntrypoint *bool `default:"false"`
 
-	ExecMD dagql.SerializedString[*buildkit.ExecutionMetadata] `name:"execMD" internal:"true" default:"null"`
+	// ExecMD carries runtime-only execution metadata; it is excluded from ID
+	// digests so cache identity only depends on explicit withExec args plus
+	// execCacheMixin implicit input.
+	ExecMD dagql.SerializedString[*buildkit.ExecutionMetadata] `name:"execMD" internal:"true" sensitive:"true" default:"null"`
```
sipsma: ExecMD needing to be a sensitive argument is weird

core/schema/container.go
```diff
@@ -1022,11 +1049,40 @@ type containerExecArgs struct {
...
+
+	// TODO: ??
+	OverrideCacheKey string `name:"overrideCacheKey" internal:"true" default:""`
```
sipsma: This is a totally reasonable approach, but need real comment now

core/schema/container.go
```diff
@@ -1022,11 +1049,40 @@ type containerExecArgs struct {
...
+		rawExecMD, ok := args["execMD"]
+		if !ok || rawExecMD == nil {
+			return dagql.NewString(""), nil
+		}
+
+		execMD, ok := rawExecMD.(dagql.SerializedString[*buildkit.ExecutionMetadata])
+		if !ok {
+			return nil, fmt.Errorf("unexpected execMD input type %T", rawExecMD)
+		}
+			if execMD.Self == nil || execMD.Self.CacheMixin == "" {
```
sipsma: Pretty sure modfunc is the only place we set this anymore and we essentially just rely on OverrideCacheKey now, is this really needed anymore?

core/schema/module.go
```diff
@@ -140,6 +145,21 @@ func (s *moduleSchema) Install(dag *dagql.Server) {
 			Args(
 				dagql.Arg("includeDependencies").Doc("Expose the dependencies of this module to the client"),
 			),
+
+		dagql.NodeFunc("_sourceContentScoped", s.moduleSourceContentScoped).
```
sipsma: Worth another pass to see if we can avoid this now. If not, there's a very clear path after the next PR, so not a huge deal.

core/schema/module.go
```diff
@@ -710,16 +732,59 @@ func (s *moduleSchema) functionWithCachePolicy(
...
+
+	mod, err := parent.Self().CurrentModule(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get current module: %w", err)
+	}
+	contentScopedID, err := mod.SourceContentScopedID(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get source content scoped ID for current module: %w", err)
+	}
+	resp.CacheKey.ID = resp.CacheKey.ID.WithArgument(call.NewArgument(
```
sipsma: If all the "CustomCacheConfigKey" funcs just overwrite args now, maybe we should just change the name of it and limit its scope to just that

core/schema/modulesource.go
```diff
@@ -222,7 +217,14 @@ func (s *moduleSourceSchema) Install(dag *dagql.Server) {
 			Doc(`The full absolute path to the context directory on the caller's host filesystem that this module source is loaded from. Only valid for local module sources.`),
 
 		dagql.NodeFunc("asModule", s.moduleSourceAsModule).
+			WithInput(dagql.CachePerClient).
 			Doc(`Load the source as a module. If this is a local source, the parent directory must have been provided during module source creation`),
+		dagql.NodeFunc("_asToolchain", s.moduleSourceAsToolchain).
```
sipsma: Add comment that while this is ugly it's not worth more effort since toolchains are going bye-bye in their current incarnation soon anyways

core/schema/wrapper.go
```diff
@@ -55,7 +55,7 @@ func DagOp[T dagql.Typed, A any, R dagql.Typed](
 
 	resultID := dagql.CurrentID(ctx)
 	if resultID != nil {
-		resultID = resultID.AppendEffectIDs(curIDForRawDagOp.Digest().String())
+		resultID = resultID.AppendEffectIDs(curIDForRawDagOp.DagOpDigest().String())
```
sipsma: Worth a check if this and other usages of DagOpDigest directly could instead just be replaced by OutputEquivalentDigest. If so, can consolidate and simplify

core/schema/wrapper.go
```diff
@@ -440,11 +441,19 @@ func DagOpContainer[A any](
...
-	ctrRes, err := core.NewContainerDagOp(ctx, curIDForContainerDagOp, argDigest, deps, ctr)
+
+	var execMD *buildkit.ExecutionMetadata
+	if withExecMD, ok := any(args).(interface {
+		DagOpExecutionMetadata() *buildkit.ExecutionMetadata
+	}); ok {
+		execMD = withExecMD.DagOpExecutionMetadata()
+	}
+
+	ctrRes, err := core.NewContainerDagOp(ctx, curIDForContainerDagOp, argDigest, deps, ctr, execMD, overrideBKCacheKey)
```
sipsma: This is pure slop that got missed until now... cleanup

core/schema/wrapper.go
```diff
@@ -474,16 +483,22 @@ type RawDagOpInternalArgs struct {
 	DagOpFilename string `internal:"true" default:"" name:"dagOpFilename"`
 }
 
+// dagOpIDWithInternalArgs returns a copy of id with dag-op internal args applied.
+func dagOpIDWithInternalArgs(id *call.ID, args ...*call.Argument) (*call.ID, error) {
```
sipsma: this seems like useless indirection, de-slop

core/sdk/module.go
```diff
@@ -65,7 +65,19 @@ func newModuleSDK(
...
+		}
+		return nil, fmt.Errorf(
+			"failed to get sdk object for sdk module %s (objects=%d interfaces=%d enums=%d resultDigest=%v): %w",
+			sdkModMeta.Self().Name(),
+			len(sdkModMeta.Self().ObjectDefs),
+			len(sdkModMeta.Self().InterfaceDefs),
+			len(sdkModMeta.Self().EnumDefs),
+			resultDigest,
+			err,
+		)
```
sipsma: leftover debugging, cleanup

core/sdk/utils.go
```diff
@@ -1,8 +1,88 @@
...
+		ID: scopedSrc.ID(),
+	}, dagql.ValueFunc(scopedSrc))
+	if err != nil {
+		return inst, err
+	}
+
+	return scopedSrc, nil
+}
+
+func ScopeModuleForSDKOperation(
```
sipsma, line 28: This and the related helper above could use another quick pass to see if it can be simplified further. Not totally convinced the complication of the call.WithScopeToDigest thing is actually needed

core/client_resource.go
```diff
@@ -27,3 +33,32 @@ func GetClientResourceAccessor(ctx context.Context, parent *Query, externalName
 	dt := h.Sum([]byte(externalName))
 	return hex.EncodeToString(dt), nil
 }
+
+// CachePerCallerModule scopes a call ID per caller module (using the module's source content digest). If the caller is not in a module, the input is just an empty string.
```
sipsma, line 40: Worth a pass to consider whether we should use the client ID if it's the main client... though seems to work totally fine in testing so far as is

core/container.go
```diff
@@ -2732,7 +2731,16 @@ func UpdatedRootFS(
 	rootfsID := curID.Append(
 		astType, "rootfs",
 		call.WithView(view),
-		call.WithModule(fieldSpec.Module))
+	)
+	inputArgs, err := dagql.ExtractIDArgs(fieldSpec.Args, rootfsID)
```
sipsma: this is so duped it seems like it should be a helper or method or something

core/dagop.go
```diff
@@ -35,6 +35,38 @@ func init() {
...
+}
+
+func dagOpContentOrStructuralDigest(id *call.ID) digest.Digest {
+	if id == nil {
+		return ""
+	}
+	if content := id.ContentDigest(); content != "" {
+		return content
+	}
+	return dagOpStructuralDigest(id)
```
sipsma: this looks like indirection slop, cleanup

core/dagop.go
```diff
@@ -483,6 +532,10 @@ func newContainerDagOp(
 type ContainerDagOp struct {
 	ID       *call.ID
 	CacheKey digest.Digest
+	// Optional runtime metadata used when evaluating the in-dagop call.
+	// This is serialized in the custom op payload but intentionally excluded
+	// from CacheMap/Digest identity derivation.
+	ExecMD *buildkit.ExecutionMetadata `json:",omitempty"`
```
sipsma: Is this still needed?

core/modfunc.go
```diff
@@ -673,7 +644,15 @@ func (fn *ModuleFunction) loadFunctionRuntime(ctx context.Context) (runtime dagq
 	mod := fn.mod
 	srv := dagql.CurrentDagqlServer(ctx)
 
-	modObj, err := dagql.NewObjectResultForID(mod, srv, mod.ResultID)
+	/* TODO: ???
+	runtimeModID, err := mod.SourceContentScopedID(ctx)
+	if err != nil {
+		return runtime, fmt.Errorf("failed to get function content scoped module id: %w", err)
+	}
+	*/
```
sipsma: rm, the uncommented code is proven correct now

core/modfunc.go
```diff
@@ -719,12 +698,13 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Any
...
-	if opts.OverrideStorageKey != "" {
-		cacheMixins = append(cacheMixins, opts.OverrideStorageKey)
-	} else {
-		cacheMixins = append(cacheMixins, dagql.CurrentStorageKey(ctx))
+	storageKey := opts.OverrideStorageKey
+	if storageKey == "" {
+		storageKey = dagql.CurrentStorageKey(ctx)
+	}
+	if storageKey != "" {
+		cacheMixins = append(cacheMixins, storageKey)
```
sipsma: don't think mixins is needed anymore

core/modfunc.go
```diff
@@ -925,9 +914,24 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Any
 			return nil, fmt.Errorf("create secret transfer post call: %w", err)
 		}
 		if hasNamedSecrets {
-			// Named secrets indicate a direct SetSecret result in the returned value.
-			// Those cannot be persisted safely across sessions.
-			safeToPersistCache = false
+			// Named secrets are only safe to persist across sessions when the return value is
+			// itself secret/socket data. If they're embedded in other objects (e.g. container
+			// state), persisting can incorrectly reuse secret-dependent results across sessions.
+			for _, id := range returnedIDsList {
```
sipsma: this comment is confusing now, also confusing that this is wrapped in a bool named `hasNamedSecrets` but then checks again and is related to sockets

core/module.go
```diff
@@ -261,7 +214,15 @@ func (mod *Module) ObjectUserDefaults(ctx context.Context, objName string) (*Env
...
-func (mod *Module) IDModule() *call.Module {
+// IDModule defines the call.Module field that will be set on call.IDs that represent calls on fields
+// from this module. The module field on call.IDs contribute to the recipe digest and also serve as
+// an input to calls when checking cache (so the digests included on the return call.Module.ID here
+// impact function call caching)
+func (mod *Module) IDModule(ctx context.Context) *call.Module {
+	// TODO: just make this return an error rather than panic
+	if mod.ResultID == nil {
+		panic("module ID is not set")
+	}
```
sipsma: worth addressing

core/module.go
```diff
@@ -283,20 +244,25 @@ func (mod *Module) IDModule() *call.Module {
...
 	default:
 		panic(fmt.Sprintf("unexpected module source kind %q", src.Kind))
 	}
 
-	contentCacheKey := mod.ContentDigestCacheKey()
-	return call.NewModule(mod.ResultID.WithDigest(digest.Digest(contentCacheKey)), mod.Name(), ref, pin)
+	contentScopedID, err := mod.SourceContentScopedID(ctx)
+	if err != nil {
+		// TODO: just make IDModule return an error rather than panic
+		panic(fmt.Sprintf("failed to get function content scoped ID for module: %v", err))
```
sipsma: fix

core/object.go
```diff
@@ -136,8 +140,26 @@ func (t *ModuleObjectType) CollectContent(ctx context.Context, value dagql.AnyRe
...
+				}
+				fieldID = fieldID.With(identityOpt)
+			} else {
+				// Best-effort fallback for field specs unavailable from runtime object type.
+				fieldID = fieldID.With(call.WithModule(curID.Module()))
+			}
+		} else {
+			// Best-effort fallback for non-object runtime values.
+			fieldID = fieldID.With(call.WithModule(curID.Module()))
+		}
```
sipsma: double check this, seems overly complicated

core/query.go
```diff
@@ -184,6 +184,13 @@ func CurrentQuery(ctx context.Context) (*Query, error) {
 }
 
 func CurrentDagqlServer(ctx context.Context) (*dagql.Server, error) {
+	// Prefer the dagql server explicitly attached to this resolver context.
+	// This is required for dynamic schemas (e.g. SDKs implemented as modules)
+	// that run selections against a server different from the session's default.
+	if srv := dagql.CurrentDagqlServer(ctx); srv != nil {
+		return srv, nil
+	}
```
sipsma: this might have been needed at some point, but worth checking if it really is anymore (and if it is, can these just become one?)

core/service.go
```diff
@@ -135,7 +135,20 @@ func (svc *Service) Hostname(ctx context.Context, id *call.ID) (string, error) {
...
+		if id == nil {
+			return "", errors.New("service ID is nil")
+		}
+		// Hostname identity should follow output-equivalence identity
+		// (e.g. content digests) rather than strict recipe identity so
+		// equivalent services converge across sessions.
+		// TODO: Carry a stable service-identity digest on Service (set at
+		// construction time from canonical output/structural identity) and
+		// use it here instead of deriving host identity ad hoc from call.ID.
+		hostDigest := id.OutputEquivalentDigest()
```
sipsma: There is a TODO here, but this isn't it

core/telemetry.go
```diff
@@ -279,15 +278,33 @@ func logResult(ctx context.Context, res dagql.AnyResult, self dagql.AnyObjectRes
 	}
 }
 
+func fieldSpecForCall(ctx context.Context, id *call.ID) (dagql.FieldSpec, bool) {
+	srv := dagql.CurrentDagqlServer(ctx)
+	if srv == nil || id == nil {
+		return dagql.FieldSpec{}, false
+	}
```
sipsma: kinda ugly and a lot of work just to check if a field is "sensitive" (of which there is only one in our API, secret.plaintext)

dagql/call/callpbv1/call.proto
```diff
@@ -72,17 +72,29 @@ message Call {
...
+
+  // Extra digests attached to this call.
+  //
+  // The call's recipe digest is authoritative call identity.
+  // Extra digests are annotations that may be used by cache and equivalence
+  // systems, but do not affect the recipe digest.
+  repeated ExtraDigest extraDigests = 14;
+
+  // Implicit inputs are engine-computed call inputs that are not explicit
+  // GraphQL field arguments, but still contribute to recipe identity.
```
sipsma, line 104: worth expanding comment; they also play into equivalence match during cache checks

dagql/call/id.go
```diff
@@ -138,7 +155,59 @@ func (id *ID) ContentDigest() digest.Digest {
...
+func (id *ID) moduleInputDigest() digest.Digest {
+	if id == nil || id.pb == nil || id.pb.Module == nil {
+		return ""
+	}
+	if id.module != nil {
+		if modID := id.module.ID(); modID != nil {
+			return modID.Digest()
+		}
+	}
+	return digest.Digest(id.pb.Module.CallDigest)
```
sipsma: indirection slop, probably fix

dagql/call/id.go
```diff
@@ -231,9 +306,106 @@ func (id *ID) Inputs() ([]digest.Digest, error) {
...
+	}
 	return inputs, nil
 }
 
+// SelfDigestAndInputs returns a digest of the call's "self" (excluding any ID
+// input digest values) and a list of input ID digests (receiver + ID literals).
+//
+// ID literals contribute only their type marker to the self digest; their digest
+// values are returned via the inputs slice.
+func (id *ID) SelfDigestAndInputs() (digest.Digest, []digest.Digest, error) {
```
sipsma: move next to `calcDigest` for easier comparison and contrast; they often need updates together

dagql/call/id.go
```diff
@@ -231,9 +306,106 @@ func (id *ID) Inputs() ([]digest.Digest, error) {
...
+		WithDelim()
+
+	// View
+	h = h.WithString(id.pb.View).
+		WithDelim()
+
+	return digest.Digest(h.DigestAndClose()), inputs, nil
+}
+
+// FIXME:!!!!!! This drops args and other inputs which may reference modules too
```
sipsma: this is not true anymore, rm comment

dagql/call/id.go
```diff
@@ -354,25 +527,79 @@ func WithView(view View) IDOpt {
...
+			extra.Label,
+		)
+	}
+}
+
+// Update the ID to have existing recipe digest and existing extra digests
+// mixed with the given string. Also update the ID to have an extra digest
+// set to the given string.
+// This is useful for making IDs shared cache based purely on the given string.
+func WithScopeToDigest(label string, scope digest.Digest) IDOpt {
```
sipsma, line 590: not convinced this ended up being really needed or worth the complication, re-assess if something simpler works

dagql/call/id.go
```diff
@@ -396,14 +623,54 @@ func WithArgs(args ...*Argument) IDOpt {
...
+		id.implicitInputs = append(id.implicitInputs, inputs...)
+		id.pb.ImplicitInputs = make([]*callpbv1.Argument, 0, len(id.implicitInputs))
+		for _, input := range id.implicitInputs {
+			if redactedInput := redactedArgForID(input); redactedInput != nil {
+				id.pb.ImplicitInputs = append(id.pb.ImplicitInputs, redactedInput.pb)
+			}
+		}
+	}
+}
+
```
sipsma, line 671: I can live with this since `plaintext` is the corner-iest of corner cases, but worth a comment if nothing else

dagql/call/id.go
```diff
@@ -665,38 +914,39 @@ func (id *ID) decode(
 		}
 		id.args = append(id.args, decodedArg)
 	}
+	for _, input := range id.pb.ImplicitInputs {
```
sipsma, line 1008: Do we need ExtraDigests here in FromAnyPB too?

dagql/call/id.go
```diff
@@ -857,3 +1196,144 @@ func mergeEffectIDs(existing []string, extra []string) []string {
...
+		out = append(out, &callpbv1.ExtraDigest{
+			Digest: extra.Digest,
+			Label:  extra.Label,
+		})
+	}
+	return out
+}
+
+func mergeExtraDigests(existing []*callpbv1.ExtraDigest, extra []*callpbv1.ExtraDigest) []*callpbv1.ExtraDigest {
+	// NOTE: this is currently O(n^2) in the number of combined extras because
```
sipsma, line 1508: worth a check if true

dagql/call/id.go
```diff
@@ -857,3 +1196,144 @@ func mergeEffectIDs(existing []string, extra []string) []string {
 	}
 	return merged
 }
+
+func cloneExtraDigests(in []*callpbv1.ExtraDigest) []*callpbv1.ExtraDigest {
```
sipsma, line 1452: all of these utils feel like they are dupes of something else and/or not used, double check

dagql/call/id_dagop.go
```diff
@@ -0,0 +1,204 @@
...
+	return id.DagOpDigest()
+}
+
+// DagOpDigest returns the digest used by dag-op-backed identity/effect
+// propagation.
+//
+// It is hashed in one pass like recipe digesting, except all ID inputs
+// (receiver, literal IDs, module identity) use output-equivalent digests
+// (content digest when available, otherwise dag-op digest).
+func (id *ID) DagOpDigest() digest.Digest {
```
sipsma: if we can live with just OutputEquivalentDigest, that would be preferable

dagql/cache.go
sipsma, line 1: whole file needs more commenting

dagql/cache_egraph.go
sipsma, line 1: whole file needs more comments and explanations

dagql/cache_egraph.go
```diff
@@ -0,0 +1,722 @@
...
+type eqMergePair struct {
+	a eqClassID
+	b eqClassID
+}
+
+func calcEgraphTermDigest(selfDigest digest.Digest, inputEqIDs []eqClassID) string {
+	h := hashutil.NewHasher().WithString(selfDigest.String())
+	for _, in := range inputEqIDs {
+		h = h.WithDelim().
+			WithString(strconv.FormatUint(uint64(in), 10))
```
sipsma: dumb, use WithUint

dagql/cachekey.go
```diff
@@ -5,188 +5,102 @@ import (
...
-func CachePerSession[P Typed, A any](
-	ctx context.Context,
-	inst ObjectResult[P],
-	args A,
-	req GetCacheConfigRequest,
-) (*GetCacheConfigResponse, error) {
-	return CachePerSessionObject(ctx, inst, args, req)
+// CachePerClient scopes a call ID per client by mixing in the client ID as
+// an implicit call input.
+var CachePerClient = ImplicitInput{
```
sipsma: probably re-name this file now. Also could rename things like this to e.g. `PerClientInput`?

dagql/cache_inputs.go
```diff
@@ -5,188 +5,102 @@ import (
...
-		schemaDgst := srv.SchemaDigest()
-		if resp.CacheKey.ID == nil {
-			return nil, fmt.Errorf("cache key ID is nil")
-		}
-		resp.CacheKey.ID = resp.CacheKey.ID.WithDigest(hashutil.HashStrings(
-			resp.CacheKey.ID.Digest().String(),
-			schemaDgst.String(),
-		))
-		return resp, nil
+func inputBoolArg(args map[string]Input, argName string) (bool, error) {
```
sipsma, line 94: seems like slop

dagql/identity_opt_test.go
sipsma, line 1: these tests need comments on what they are doing

dagql/server.go
```diff
@@ -673,6 +672,26 @@ func (s *Server) Load(ctx context.Context, id *call.ID) (AnyObjectResult, error)
 }
 
 func (s *Server) LoadType(ctx context.Context, id *call.ID) (AnyResult, error) {
+	ctx = idToContext(ctx, id)
+	ctx = srvToContext(ctx, s)
+
+	// Before recursing, check if we already have a cached result for this ID.
```
sipsma: setting ctx here and then using it later feels fragile/sus, make more robust and isolated

engine/server/client_resources.go
```diff
@@ -66,6 +66,11 @@ func (srv *Server) addClientResourcesFromID(ctx context.Context, destClient *dag
 	// Load IDs in the source client's metadata context so any cache-miss re-evaluation
 	// (e.g. host.unixSocket post-call side effects) targets the correct source client.
 	srcClientCtx := engine.ContextWithClientMetadata(ctx, srcClient.clientMetadata)
+	if srcClient.dagqlRoot != nil {
+		// This path can run during nested client initialization where the incoming ctx
+		// has no query set yet; ensure re-evaluated source IDs still have query access.
+		srcClientCtx = core.ContextWithQuery(srcClientCtx, srcClient.dagqlRoot)
+	}
```
sipsma: sus, but not a big deal

.gitignore
```diff
@@ -45,3 +45,5 @@ go.work.sum
 
 # merged from dagger/examples repository
 **/node_modules
+
+/scratch
```
sipsma: specific to my machine, rm from here

dagql/call/id.go
sipsma: Also, I just saw id.decode might be missing ExtraDigests too?

core/telemetry.go
sipsma: not an obvious perf bottleneck yet so gonna leave this alone for now

dagql/call/id.go
sipsma: nevermind, I missed it's not necessary since we only need this for pb fields that get wrapped

dagql/call/id.go
sipsma: true but probably better to just have this warning comment until the day we have many digests, at which time pprof will show this become a bottleneck

dagql/call/id.go
sipsma: After another pass it's actually as simple as can be. Added a more detailed comment explaining what/why it is.

core/schema/module.go
sipsma: Made an attempt but think it caused a perf regression. Not worth poking more right now

marcosnils:
:open_hands: ! I'm proposing the following image as a release logo once we merge this:

image

PR #11857, opened 2026-02-12
ci: gracefully handle git worktrees

kpenfound:
This is so that we can develop dagger/dagger with git worktrees. This should greatly simplify agent setups. I tested locally and it looks good, I can `dagger-dev -m version call version` from a regular branch or a worktree and it works correctly. Using main dagger the same command does not work in a worktree.

Summary:
In git worktrees, .git is a file (not a directory) containing a gitdir reference. This caused Dagger to crash when trying to sync .git as a directory. This change adds worktree support at multiple layers:

- version module: detect inaccessible git and fall back to digest-based versioning (v0.0.0-dev-<digest>)
- gitutil: use --git-common-dir for relative paths in worktrees
- client filesync: send empty directory when .git is a file
- server filesync: return empty temp dir when .git is not a directory

### DIFF
--- a/version/main.go
+++ b/version/main.go
@@ -15,10 +15,13 @@ import (
 )
 
 func New(
+	ctx context.Context,
+
 	// A git repository containing the source code of the artifact to be versioned.
 	// +optional
-	// +defaultPath="/.git"
-	git *dagger.Directory,
+	// +defaultPath="/"
+	// +ignore=["*", "!.git"]
+	gitParent *dagger.Directory,
 
 	// A directory containing all the inputs of the artifact to be versioned.
 	// An input is any file that changes the artifact if it changes.
@@ -35,12 +38,17 @@ func New(
 	// +defaultPath="/.changes/.next"
 	nextVersionFile *dagger.File,
 ) *Version {
-	return &Version{
-		Git:             git.AsGit(),
-		GitDir:          git,
+	v := &Version{
 		Inputs:          inputs.Filter(dagger.DirectoryFilterOpts{Gitignore: true}),
 		NextVersionFile: nextVersionFile,
 	}
+
+	if git, err := gitParent.Directory(".git").Sync(ctx); err == nil {
+		v.Git = git.AsGit()
+		v.GitDir = git
+	}
+
+	return v
 }
 
 type Version struct {
@@ -57,6 +65,10 @@ type Version struct {
 
 // Generate a version string from the current context
 func (v Version) Version(ctx context.Context) (string, error) {
+	if v.Git == nil {
+		return v.fallbackVersion(ctx)
+	}
+
 	dirty, err := v.Dirty(ctx)
 	if err != nil {
 		return "", err
@@ -106,8 +118,40 @@ func (v Version) Version(ctx context.Context) (string, error) {
 	return fmt.Sprintf("%s-%s-dev-%s", next, pseudoversionTimestamp(commitDate), commit[:12]), nil
 }
 
+func (v Version) fallbackVersion(ctx context.Context) (string, error) {
+	// Try to get next release version from .changes/.next
+	next := "v0.0.0"
+	if v.NextVersionFile != nil {
+		if content, err := v.NextVersionFile.Contents(ctx); err == nil {
+			for _, line := range strings.Split(content, "\n") {
+				line = strings.TrimSpace(line)
+				if strings.HasPrefix(line, "v") && semver.IsValid(line) {
+					next = line
+					break
+				}
+			}
+		}
+	}
+
+	// Generate a version based on inputs digest
+	rawDigest, err := v.Inputs.Digest(ctx)
+	if err != nil {
+		return "", err
+	}
+	_, digest, ok := strings.Cut(rawDigest, ":")
+	if !ok {
+		return "", fmt.Errorf("invalid digest: %s", rawDigest)
+	}
+
+	return fmt.Sprintf("%s-%s-dev-%s", next, pseudoversionTimestamp(time.Now()), digest[:12]), nil
+}
+
 // Return the tag to use when auto-downloading the engine image from the CLI
 func (v Version) ImageTag(ctx context.Context) (string, error) {
+	if v.Git == nil {
+		return v.fallbackVersion(ctx)
+	}
+
 	if tag, err := v.CurrentTag(ctx); err != nil {
 		return "", err
 	} else if tag != "" {
@@ -130,6 +174,10 @@ func (v Version) ImageTag(ctx context.Context) (string, error) {
 }
 
 func (v Version) Dirty(ctx context.Context) (bool, error) {
+	if v.Git == nil {
+		return true, nil
+	}
+
 	committed := v.Git.Head().Tree()
 
 	// Overlay local inputs onto committed tree, then diff.
@@ -147,6 +195,10 @@ func (v Version) Dirty(ctx context.Context) (bool, error) {
 }
 
 func (v Version) CurrentTag(ctx context.Context) (string, error) {
+	if v.Git == nil {
+		return "", nil
+	}
+
 	commit, err := v.Git.Head().Commit(ctx)
 	if err != nil {
 		return "", err


kpenfound:
@sipsma without those changes, we get this when we're in a worktree:

```
failed to sync: failed to create local fs: failed to create base fs: stat /var/lib/dagger/worker/snapshots/snapshots/3/fs/private/tmp/dagger-worktree-
    test/.git: not a directory
```

This is because (what I learned while putting this together) is that a git worktree's .git is actually a file, not a directory, that references the _actual_ .git directory

edit: the alternative would be to update the code where we load the `.git` to make sure its a directory, otherwise fall back to empty .git dir. Right now the .git path is ambiguous

sipsma:
@kpenfound right but I think we can just adjust the version module to not assume it's a directory. This diff worked for me, it's just a slight update to the changes to the version module in this PR, nothing else:
```diff
diff --git a/version/main.go b/version/main.go
index bd59ed390..b17235074 100644
--- a/version/main.go
+++ b/version/main.go
@@ -15,10 +15,13 @@ import (
 )
 
 func New(
+       ctx context.Context,
+
        // A git repository containing the source code of the artifact to be versioned.
        // +optional
-       // +defaultPath="/.git"
-       git *dagger.Directory,
+       // +defaultPath="/"
+       // +ignore=["*", "!.git"]
+       gitParent *dagger.Directory,
 
        // A directory containing all the inputs of the artifact to be versioned.
        // An input is any file that changes the artifact if it changes.
@@ -35,12 +38,17 @@ func New(
        // +defaultPath="/.changes/.next"
        nextVersionFile *dagger.File,
 ) *Version {
-       return &Version{
-               Git:             git.AsGit(),
-               GitDir:          git,
+       v := &Version{
                Inputs:          inputs.Filter(dagger.DirectoryFilterOpts{Gitignore: true}),
                NextVersionFile: nextVersionFile,
        }
+
+       if git, err := gitParent.Directory(".git").Sync(ctx); err == nil {
+               v.Git = git.AsGit()
+               v.GitDir = git
+       }
+
+       return v
 }
 
 type Version struct {
@@ -57,6 +65,10 @@ type Version struct {
 
 // Generate a version string from the current context
 func (v Version) Version(ctx context.Context) (string, error) {
+       if v.Git == nil {
+               return v.fallbackVersion(ctx)
+       }
+
        dirty, err := v.Dirty(ctx)
        if err != nil {
                return "", err
@@ -106,8 +118,40 @@ func (v Version) Version(ctx context.Context) (string, error) {
        return fmt.Sprintf("%s-%s-dev-%s", next, pseudoversionTimestamp(commitDate), commit[:12]), nil
sipsma@dagger_dev:~/repo/github.com/sipsma/dagger2/wts/test-wt$ git diff version/main.go | cat
diff --git a/version/main.go b/version/main.go
index bd59ed390..b17235074 100644
--- a/version/main.go
+++ b/version/main.go
@@ -15,10 +15,13 @@ import (
 )
 
 func New(
+       ctx context.Context,
+
        // A git repository containing the source code of the artifact to be versioned.
        // +optional
-       // +defaultPath="/.git"
-       git *dagger.Directory,
+       // +defaultPath="/"
+       // +ignore=["*", "!.git"]
+       gitParent *dagger.Directory,
 
        // A directory containing all the inputs of the artifact to be versioned.
        // An input is any file that changes the artifact if it changes.
@@ -35,12 +38,17 @@ func New(
        // +defaultPath="/.changes/.next"
        nextVersionFile *dagger.File,
 ) *Version {
-       return &Version{
-               Git:             git.AsGit(),
-               GitDir:          git,
+       v := &Version{
                Inputs:          inputs.Filter(dagger.DirectoryFilterOpts{Gitignore: true}),
                NextVersionFile: nextVersionFile,
        }
+
+       if git, err := gitParent.Directory(".git").Sync(ctx); err == nil {
+               v.Git = git.AsGit()
+               v.GitDir = git
+       }
+
+       return v
 }
 
 type Version struct {
@@ -57,6 +65,10 @@ type Version struct {
 
 // Generate a version string from the current context
 func (v Version) Version(ctx context.Context) (string, error) {
+       if v.Git == nil {
+               return v.fallbackVersion(ctx)
+       }
+
        dirty, err := v.Dirty(ctx)
        if err != nil {
                return "", err
@@ -106,8 +118,40 @@ func (v Version) Version(ctx context.Context) (string, error) {
        return fmt.Sprintf("%s-%s-dev-%s", next, pseudoversionTimestamp(commitDate), commit[:12]), nil
 }
 
+func (v Version) fallbackVersion(ctx context.Context) (string, error) {
+       // Try to get next release version from .changes/.next
+       next := "v0.0.0"
+       if v.NextVersionFile != nil {
+               if content, err := v.NextVersionFile.Contents(ctx); err == nil {
+                       for _, line := range strings.Split(content, "\n") {
+                               line = strings.TrimSpace(line)
+                               if strings.HasPrefix(line, "v") && semver.IsValid(line) {
+                                       next = line
+                                       break
+                               }
+                       }
+               }
+       }
+
+       // Generate a version based on inputs digest
+       rawDigest, err := v.Inputs.Digest(ctx)
+       if err != nil {
+               return "", err
+       }
+       _, digest, ok := strings.Cut(rawDigest, ":")
+       if !ok {
+               return "", fmt.Errorf("invalid digest: %s", rawDigest)
+       }
+
+       return fmt.Sprintf("%s-%s-dev-%s", next, pseudoversionTimestamp(time.Now()), digest[:12]), nil
+}
+
 // Return the tag to use when auto-downloading the engine image from the CLI
 func (v Version) ImageTag(ctx context.Context) (string, error) {
+       if v.Git == nil {
+               return v.fallbackVersion(ctx)
+       }
+
        if tag, err := v.CurrentTag(ctx); err != nil {
                return "", err
        } else if tag != "" {
@@ -130,6 +174,10 @@ func (v Version) ImageTag(ctx context.Context) (string, error) {
 }
 
 func (v Version) Dirty(ctx context.Context) (bool, error) {
+       if v.Git == nil {
+               return true, nil
+       }
+
        committed := v.Git.Head().Tree()
 
        // Overlay local inputs onto committed tree, then diff.
@@ -147,6 +195,10 @@ func (v Version) Dirty(ctx context.Context) (bool, error) {
 }
 
 func (v Version) CurrentTag(ctx context.Context) (string, error) {
+       if v.Git == nil {
+               return "", nil
+       }
+
        commit, err := v.Git.Head().Commit(ctx)
        if err != nil {
                return "", err
```

That was all I needed, no changes to filesync.

I made the worktree a subdir of my dagger repo checkout at `./wts/test-wt` (and put a .gitignore entry for `./wts`). I'm not 100% sure if that was load-bearing or not, just mentioning in case it was.

Does that work for you too? If so I'd prefer that.

grouville:
> @sipsma without those changes, we get this when we're in a worktree:
> 
> ```
> failed to sync: failed to create local fs: failed to create base fs: stat /var/lib/dagger/worker/snapshots/snapshots/3/fs/private/tmp/dagger-worktree-
>     test/.git: not a directory
> ```
> 
> This is because (what I learned while putting this together) is that a git worktree's .git is actually a file, not a directory, that references the _actual_ .git directory

Thanks both for digging into this, this was on my todo for so long 🙏 

My take is similar to Erik: the filesync part in this PR feels like a hack, and it directly opens bigger design discussions about worktree handling (`dir.AsGit()`, session-attachable/client-side ownership, `GitRepository` of type worktree passed as function arguments etc.).

I understand why we'll push for this short-term fix/hack of changing just the module Version (for our own needs), but IMO there's still a proper design fix to find around _git worktrees in Dagger_ (totally fine for those changes, to unlock you)

> There was also a concrete bug in the current Git change (not Erik's version):
> `util/gitutil/cli_helpers.go` can return `--git-common-dir` directly, and in regular repos that value is `.git` (relative). That breaks the existing `GitDir()` absolute-path assumption: `URL()` can become`file://.git`, and joins like `filepath.Join(gitDir, "index")` resolve relative to process CWD instead of repo-root context.

Given that, +1 for Erik’s patch : it fixes the immediate worktree/version issue with much smaller surface area, then that leaves us time to do a focused follow-up for proper worktree support, integrated in the git API itself

kpenfound:
Nice thanks @sipsma that worked! Just force pushed

PR #11812, opened 2026-02-06
Workspaces

shykes:
*Introduce workspaces: project-level configuration separate from module definitions.*

Dagger currently conflates project configuration and module definitions in a single `dagger.json`. This PR separates them: `.dagger/config.toml` for project configuration (which modules to use, how to configure them), `dagger.json` for module definitions (functions, types, dependencies). The result is a clearer model for both module users and module developers.

---

Configure your project with `.dagger/config.toml`:

```toml
[modules.ci]
source = "modules/ci"
entrypoint = true

[modules.go]
source = "github.com/dagger/go-toolchain@v1.0"
config.goVersion = "1.22"
```

```bash
dagger call build              # entrypoint module — functions at the top level
dagger call go lint            # namespaced module
dagger -W github.com/acme/project@main call build   # remote workspace
```

Modules interact with the project through a first-class API:

```go
func New(ws dagger.Workspace) *Ci {
    goModFiles := ws.Directory(".").Glob("**/go.mod")
    // dynamically discover and build Go modules in the workspace
}
```

Existing projects work without changes — the engine infers a compatibility workspace from legacy `dagger.json` and nudges users to migrate:

```
WARNING: this project uses a legacy dagger.json format.
Run 'dagger migrate' to update your project.
```

When ready, migration is automated and previewed before applying:

```
$ dagger migrate
MIGRATION legacy-dagger-json
Convert legacy dagger.json to .dagger/config.toml

[changeset preview]

Apply changes? [y/N]
```

---

**[Full design doc](https://github.com/dagger/dagger/blob/workspace/hack/designs/workspace.md)**

## Task list

- [ ] Simplify workspace location model: git root is workspace root; config directory is selected from workspace cwd
- [ ] Review docs site redirects (we're breaking a lot of links)
- [ ] Migration warning should be more visible
- [x] Migrated workspace config should have main module at the top
- [ ] In migration doc, explain clearly what is `legacy-default-path` and what to do with it
- [ ] Migration no longer has auto-population of module settings in file -> bring that back (checked)
- [x] CWD module autoload is only a fallback when no `.dagger/config.toml` exists; configured workspaces own module loading
- [ ] `github.com/dagger/dagger` fails to load after migration
- [ ] `dagger uninstall` doesn't work (`dagger uninstall dagger-dev` in dagger/dagger@v0.20.4 checkout)
- [ ] Settings: ignore `Workspace` arguments (special case)
- [x] Some workspace config edits don't preserve byte order (checked)
[...]

PR #11858, opened 2026-02-12
core: support content-hashable function results

vito:
Splitting this out of #11812 so I can get a clean test of it in isolation to make sure the accompanied refactor didn't break anything.

### DIFF
--- a/core/enum.go
+++ b/core/enum.go
@@ -7,9 +7,7 @@ import (
 
 	"github.com/dagger/dagger/dagql"
 	"github.com/dagger/dagger/dagql/call"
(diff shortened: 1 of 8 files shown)


sipsma:
The test that failed in test-base + test-modules is unrelated I believe. I started seeing that too today on other branches. `/shutdown` just seems to be taking a long time for some reason.

Did a quick pass and the approach looks good so far 👍

vito:
CI all green, closing as we hope to just stabilize + merge Workspaces, but can reconsider merging separately if it drags on more

PR #11862, opened 2026-02-14
core(changeset): fix subdirectory scoping in changeset layers

### LINKED ISSUE #11656: 🐞 Scoped changeset `Layer` leaks surrounding filesystem
### What is the issue?

Follow-up of Discord discussion: https://discord.com/channels/707636530424053791/1030538312508776540/threads/1459242431596990709

When using the `.Changes(...).Layer()` on a directory, for example

```go
layer := ctr.
    //
    // ...add/modify/delete some files
    //
    Directory("/usr/local").
    Changes(ctr.Directory("/usr/local").
    Layer()
```

You would expect the `layer` to container entries from just `/usr/local`. However, it seems to take a diff of the entire filesystem (from `/`), not just the scoped directory. 

Note: it looks like `AsPatch` does not have this issue though.

### Dagger version

dagger v0.19.9 (image://registry.dagger.io/engine:v0.19.9) darwin/arm64/v8

### Steps to reproduce

```go
func (m *Example) Repro() *dagger.Directory {
	base := dag.Container().
		WithDirectory("public", dag.Directory())

	layer := base.
		WithNewFile("Gemfile", "").
		WithDirectory(
			"public",
			dag.Directory().
				WithNewFile("asset_foo", "").
				WithNewFile("asset_bar", ""),
		).
		Directory("public").
		Changes(base.Directory("public")).
		Layer()

	return layer
}
```

Calling the repro

```
dagger call repro entries

Gemfile
public/
```

Expected: only the `public/` directory is present
Found: the `Gemfile` appears in the output layer

### Log output

_No response_

grouville:
Fixes #11656

Users expect`changeset.layer` to follow 3 simple rules for scoped changesets:
1) diff only what is inside the selected scope
2) keep the resulting layer shape at that same scope
3) if there are no scoped changes, the layer must be empty

Before this change, `changeset.layer` called `before.Diff(after)` directly. For scoped inputs, that bypassed the resolver diff path that normalizes both sides before diffing, which could lead to inconsistent scoped output, breaking 1 and 2 (issue above).

This fix routes `changeset.layer` through resolver `diff`, then re-anchors the result to the original scope path.

Performance-wise, this remains on the BuildKit path for `layer` (no `ComputePaths`, `AsPatch`, or git merge path). Extra work is limited to one `Entries` check and a re-anchor only when there is actual content.

### DIFF
--- a/core/integration/changeset_test.go
+++ b/core/integration/changeset_test.go
@@ -456,6 +456,92 @@ func (ChangesetSuite) TestChangeset(ctx context.Context, t *testctx.T) {
 		require.NotContains(t, entries, "root.txt")
 	})
 
+	t.Run("layer is scoped to subdirectory changes", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		oldDir := c.Directory().
+			WithDirectory("public", c.Directory())
+
+		newDir := oldDir.
+			WithNewFile("Gemfile", "source \"https://rubygems.org\"").
+			WithDirectory("public", c.Directory().
+				WithNewFile("asset_foo", "foo").
+				WithNewFile("asset_bar", "bar"))
+
+		changes := newDir.Directory("/public").Changes(oldDir.Directory("/public"))
+		layer := changes.Layer()
+
+		entries, err := layer.Entries(ctx)
+		require.NoError(t, err)
+		require.ElementsMatch(t, []string{"asset_foo", "asset_bar"}, entries)
+		require.NotContains(t, entries, "Gemfile")
+		require.NotContains(t, entries, "public/")
+	})
+
+	t.Run("layer ignores out-of-scope changes when scoped to subdirectory", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		oldDir := c.Directory().
+			WithDirectory("public", c.Directory())
+
+		newDir := oldDir.
+			WithNewFile("Gemfile", "source \"https://rubygems.org\"")
+
+		changes := newDir.Directory("/public").Changes(oldDir.Directory("/public"))
+		layer := changes.Layer()
+
+		entries, err := layer.Entries(ctx)
+		require.NoError(t, err)
+		require.Empty(t, entries)
+	})
+
+	t.Run("layer is relative to nested scoped path", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		oldDir := c.Directory().
+			WithDirectory("assets", c.Directory().
+				WithDirectory("public", c.Directory()))
+
+		newDir := oldDir.
+			WithNewFile("Gemfile", "source \"https://rubygems.org\"").
+			WithDirectory("assets", c.Directory().
+				WithDirectory("public", c.Directory().
+					WithNewFile("asset_foo", "foo").
+					WithNewFile("asset_bar", "bar")))
+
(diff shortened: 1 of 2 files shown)


core/integration/changeset_test.go
```diff
@@ -456,6 +456,79 @@ func (ChangesetSuite) TestChangeset(ctx context.Context, t *testctx.T) {
...
+			WithDirectory("public", c.Directory().
+				WithNewFile("asset_foo", "foo").
+				WithNewFile("asset_bar", "bar"))
+
+		changes := newDir.Directory("/public").Changes(oldDir.Directory("/public"))
+		layer := changes.Layer()
+
+		entries, err := layer.Entries(ctx)
+		require.NoError(t, err)
+		require.Equal(t, []string{"public/"}, entries)
```
alexcb: Are we able to change the current functionality of `.Layer()` to have it simply return the directory already set to `public/`?

I read up on the discord chat, which @cbochs mentioned:

> This feels like it's less useful and quite the footgun to allow this operation (a changeset layer) on filtered directories (i.e. .Directory("/usr/local")) to leak other fs info into that changeset. Especially if the recommendation is to use the layer as an overlay from root (i.e. ctr.WithDirectory("/", layer))

Is there a specific use-case for why we need to maintain the original file system path. If not, I don't see why it can't be changed.

grouville: I agree, updated! The main reason was to avoid a transient breaking change on our API. We usually merge the breaking changes prior a major version bump, no strong opinion here

grouville:
As i touched the graphQL docs, I need to regenerate (doing it rn)

Asking @sipsma and @vito for further review ; mainly: is it ok to merge this breaking change in a non-major version bump ?

core/schema/directory.go
```diff
@@ -1007,11 +1008,19 @@ func (s *directorySchema) changesetLayer(ctx context.Context, parent dagql.Objec
...
+		dagql.Selector{Field: "diff", Args: []dagql.NamedInput{
+			{Name: "other", Value: dagql.NewID[*core.Directory](changes.After.ID())},
+		}},
+	)
 	if err != nil {
 		return inst, err
 	}
-	return dagql.NewObjectResultForCurrentID(ctx, srv, dir)
+
+	return dagql.NewObjectResultForCurrentID(ctx, srv, scopedDiff.Self())
```
vito: Should we just return `scopedDiff` directly here?

core/schema/directory.go
```diff
@@ -295,7 +295,8 @@ func (s *directorySchema) Install(srv *dagql.Server) {
 		Syncer[*core.Changeset]().
 			Doc(`Force evaluation in the engine.`),
 		dagql.NodeFunc("layer", DagOpDirectoryWrapper(srv, s.changesetLayer)).
-			Doc(`Return a snapshot containing only the created and modified files`),
+			Doc(`Return a snapshot containing only the created and modified files`,
+				`For scoped changesets, returned paths are relative to the selected scope.`),
```
vito: TBH I don't think we even need this docs change, it's more confusing to mention it than to have it "just work". For example 'scoped changeset' isn't worth defining because a Changeset is just what you get from comparing two Directories, there's no 'scoped' vs. 'unscoped', conceptually you just passed different directories in.

core/schema/directory.go
grouville: Yes you're right, updated ✅

core/schema/directory.go
grouville: Thanks, I'm glad I asked 🙏

PR #11863, opened 2026-02-14
Add a local-defaults documentation for .env files

AsocPro:
I was helped on discord learn about the .env file features for defaulting of arguments and since there was no documentation covering the feature I offered to add a quick page and was told that would be great.

I verified that the docs build and that they look correct and all links are functional when running the doc server. I also validated that all of the cases I documented here function the way they are documented.


AsocPro:
I was trying to  look into the failing tests and it appears that after a rebase I'm getting even more. I'm not sure if the tests failing are related to my change as its just a doc change and it doesn't appear to do anything that would affect them. especially the trivy tests. But if my assumptions are  wrong I'm happy to dig in and try to figure out why my changes are causing failures.

PR #11871, opened 2026-02-17
fix(sdk/go): make Go codegen ordering deterministic

### LINKED ISSUE #11802: 🐞 Go codegen is non deterministic sometimes
### What is the issue?

When running `dagger develop` multiple times on the same module, ocassionally the output is different. Specifically the ordering of a particular module type's `.MarshalJSON`, `.UnmarshalJSON` functions is changing in the `/dagger.gen.go`.

Image
Image

I have an example repo with a module from our mono-repo which is showing this behaviour: https://github.com/chrisjpalmer/dagger-codegen-bug/tree/main. I also thought I found the cause... although I'll let you guys be the decider on that.

In the `examples/` directory you can see the before/after snapshot of `dagger.gen.go`:

[chrisjpalmer/dagger-codegen-bug@main/example-diff](https://github.com/chrisjpalmer/dagger-codegen-bug/tree/main/example-diff)

One thing to note is that the type (`Remote`), whose functions are being reordered, is not the main module type. It is a secondary type returned by the module. Its also worth noting that `Remote`'s functions seems to consistently fluctuate before and after the functions for an interface type we have called `RegistryAuth`. 

### Dagger version

0.19.10

### Steps to reproduce

_No response_

### Log output

_No response_

grouville:
Fixes #11802

Go codegen could reorder generated output between identical runs (e.g. `MarshalJSON` / `UnmarshalJSON` blocks), creating noisy diffs.

### Root cause

`token.Pos()` is not stable across runs for multi-file packages: it depends on `FileSet` insertion order, and `go/packages` does not guarantee file load order ([source](https://github.com/golang/tools/blob/master/go/packages/packages.go#L1392)).

### What this PR does

Following Tibor’s feedback, I removed `Pos()`-based ordering instead of trying to stabilize it.

- removed `Pos()` sorting from codegen traversal
- rely on stable name ordering from `pkgScope.Names()`
- keep explicit alphabetical sorts only where output is user-facing (`invoke` object cases, method emission)

**Why this is safe**: the previous `Pos()` sort was intended to preserve definition order, but in multi-file packages it was already loader-dependent and non-deterministic. `pkgScope.Names()` gives deterministic traversal, and user-facing ordering remains explicit: it mostly touches generated files for linter hints + CLI error messages.

I do understand the (small) appeal in UX, but it's a small gain vs a lot of CPU sorting

### Regression coverage

Added a focused deterministic unit test in `cmd/codegen/generator/go/templates/visit_determinism_test.go`

It builds the same package with different file insertion orders and asserts stable: struct traversal order and `GitRepo` method traversal order

Result:
- fails on `main`
- passes with this PR

Local repro:
Take Chris' go test and tweak his repro with the same structure as my module in `TestVisitTypesDeterministicAcrossFiles`, then run it like this:
```bash
./hack/dev
./hack/with-dev go test ./...`
```

### Side effect handled here

After removing `Pos()` ordering, enum value order became visible in CLI/help/error output and was inconsistent in some integration tests. This PR also normalizes enum value order (canonical sorted order) and updates expectations.

### Thanks

Thanks to @chrisjpalmer for the original repro and diagnosis:
https://github.com/chrisjpalmer/dagger-codegen-bug

### DIFF
--- a/core/integration/module_cli_test.go
+++ b/core/integration/module_cli_test.go
@@ -592,6 +592,48 @@ func (CLISuite) TestDevelopDeterministicCodegen(ctx context.Context, t *testctx.
 	// Test that running codegen multiple times produces identical output.
 	// This is critical for version control - we want to be able to commit
 	// generated files and know they won't change unless the API changes.
+	t.Run("go split methods across files", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		modGen := goGitBase(t, c).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work").
+			With(daggerExec("init", "--name=test", "--sdk=go", "--source=.")).
+			WithNewFile("/work/main.go", `package main
+
+type GitRepo struct{}
+
+func (m *GitRepo) RemoteA() *RemoteA {
+	return &RemoteA{}
+}
+
+func (m *GitRepo) RemoteB() *RemoteB {
+	return &RemoteB{}
+}
+`,
+			).
+			WithNewFile("/work/remote_a.go", `package main
+
+type RemoteA struct{}
+`,
+			).
+			WithNewFile("/work/remote_b.go", `package main
+
+type RemoteB struct{}
+`,
+			)
+
+		modGen = modGen.With(daggerExec("develop"))
+		firstGen, err := modGen.File("/work/dagger.gen.go").Contents(ctx)
+		require.NoError(t, err)
+
+		modGen = modGen.With(daggerExec("develop"))
+		secondGen, err := modGen.File("/work/dagger.gen.go").Contents(ctx)
+		require.NoError(t, err)
+
+		require.Equal(t, firstGen, secondGen, "Generated code should be deterministic across multiple runs")
+	})
+
 	t.Run("go with dependencies", func(ctx context.Context, t *testctx.T) {
 		c := connect(ctx, t)
 


tiborvass:
I am unable to reproduce the original bug with `dagger develop`, even the module test in the PR passes for me without the proposed fix.

However, I am able to reproduce the non-determinism when isolating the package loading code as outlined in https://github.com/chrisjpalmer/dagger-codegen-bug/tree/main but I find the proposed fix too complex for what's needed.

Essentially there are two solutions for determinism:
1. Sorted by object name. This is the default for the golang.org/x/tools/go/package. If we go this route we should just remove the sort.Slice lines, it would be the simplest.
2. Sort based on filename first, then offset within filename. This is the solution suggested in this PR, but I think the following would be simpler and still relatively efficient (not calling `token.Fset.Position` for every object).

```Go
    sort.Slice(ps.objs, func(i, j int) bool {
        pi, pj := ps.objs[i].Pos(), ps.objs[j].Pos()
        fi, fj := ps.fset.File(pi), ps.fset.File(pj)
        if fi == fj {
            return fi.Offset(pi) < fj.Offset(pj)
        }
        return fi.Name() < fj.Name()
    })
```

tiborvass:
Since the `sort.Slice` would be called many places you could create a generic `sortObjects`:

```Go
func sortObjects[T any](fset *token.Fset, objs []*T) {
    sort.Slice(objs, func(i, j int) bool {
        pi, pj := objs[i].Pos(), objs[j].Pos()
        fi, fj := fset.File(pi), fset.File(pj)
        if fi == fj {
            return fi.Offset(pi) < fj.Offset(pj)
        }
        return fi.Name() < fj.Name()
    })
}
```

cmd/codegen/generator/go/templates/sort_helpers.go
```diff
@@ -0,0 +1,89 @@
...
+	filename string
+	offset   int
+	name     string
+}
+
+func (k sortKey) cmp(other sortKey) int {
+	return cmp.Or(
+		cmp.Compare(k.filename, other.filename),
+		cmp.Compare(k.offset, other.offset),
+		cmp.Compare(k.name, other.name),
```
tiborvass: `name` field is irrelevant. If `filename` fields are identical, and so are `offset` fields, we're talking about the same object.

chrisjpalmer:
Thank you Guillaume!!

tiborvass:
@grouville On a second thought, I think we should go with option 1 and remove the sort altogether.


The `sort.Slice` was justified by the comment: https://github.com/dagger/dagger/blob/40a44e8cf3814b9a3b7ac06c67a728346d3954cb/cmd/codegen/generator/go/templates/modules.go#L86-L87

Despite not realizing that Pos() ordering is non-deterministic: https://github.com/golang/tools/blob/43ba12690711e1edbceb6801c757d574854b21b4/go/packages/packages.go#L1391-L1393

it also does not seem to take into account packages with more than one file, in which case, the most logical deterministic file ordering would be alphabetical, which would already not honor the original intent of the comment above sort.Slice.

So just get rid of it. The deterministic output will be ensured by the call to https://github.com/dagger/dagger/blob/40a44e8cf3814b9a3b7ac06c67a728346d3954cb/cmd/codegen/generator/go/templates/modules.go#L78

which sorts objects by name: https://pkg.go.dev/go/types#Scope.Names
```
Names returns the scope's element names in sorted order. 
```

grouville:
Would that create a big diff on next's version bump though ?

> @grouville On a second thought, I think we should go with option 1 and remove the sort altogether.
> 
> The `sort.Slice` was justified by the comment:
> 
> https://github.com/dagger/dagger/blob/40a44e8cf3814b9a3b7ac06c67a728346d3954cb/cmd/codegen/generator/go/templates/modules.go#L86-L87
> 
> Despite not realizing that Pos() ordering is non-deterministic: https://github.com/golang/tools/blob/43ba12690711e1edbceb6801c757d574854b21b4/go/packages/packages.go#L1391-L1393
> 
> it also does not seem to take into account packages with more than one file, in which case, the most logical deterministic file ordering would be alphabetical, which would already not honor the original intent of the comment above sort.Slice.
> 
> So just get rid of it. The deterministic output will be ensured by the call to
> 
> https://github.com/dagger/dagger/blob/40a44e8cf3814b9a3b7ac06c67a728346d3954cb/cmd/codegen/generator/go/templates/modules.go#L78
> 
> which sorts objects by name: https://pkg.go.dev/go/types#Scope.Names
> 
> ```
> Names returns the scope's element names in sorted order. 
> ```

tiborvass:
The `dagger.gen.go` file is usually in `.gitignore`, but even if it's not, the impact to users is simply a bit of a surprise for which we absolutely should mention this in the release notes.

core/integration/module_cli_test.go
```diff
@@ -647,6 +647,61 @@ func (m *Test) UsesDep(ctx context.Context) (string, error) {
 		// The generated code should be byte-for-byte identical
 		require.Equal(t, firstGen, secondGen, "Generated code should be deterministic across multiple runs")
 	})
+
+	t.Run("go secondary type and interface", func(ctx context.Context, t *testctx.T) {
```
eunomie: 🤔 I added this test to `main` (just this test, without the fix) and the test is passing on my machine.
Not saying the test itself is wrong, it's probably good to have it in the suite, but I wonder if there isn't a missing piece somewhere to force it to fail without the fix.

grouville: Good call, thanks for checking 🙏 

I also had trouble reproducing this reliably via `dagger develop` in integration (heisenberg/cache-sensitive)

So, in the end, I added a deterministic unit repro in `visit_determinism_test.go` that changes file insertion order for the same package and asserts stable traversal/method order. That test fails on `main` and passes with this PR, so we now have a reliable guard to avoid future regressions 🙏

grouville:
@tiborvass You were right on the approach 🙏 

I reworked the PR accordingly: removed the `Pos()`-based ordering path and kept deterministic name-based ordering. I also added a focused regression unit test for this (`visit_determinism_test.go`).

  Feel free to take another look when you have a minute 👼

PR #11874, opened 2026-02-17
Add `Workspace` API and caching support

vito:
Extracting a backwards-compatible portion of #11812 that we can ship first and begin adopting in the repo.

### DIFF
--- a/core/integration/workspace_test.go
+++ b/core/integration/workspace_test.go
@@ -48,6 +48,49 @@ func initDangModule(name, source string) dagger.WithContainerFunc {
 	}
 }
 
+// initDangBlueprint creates a Dang blueprint module and an app module that
+// uses it. The blueprint source is written to blueprints/<name>/ and the app
+// module is initialized at the workspace root with --blueprint pointing to it.
+func initDangBlueprint(name, source string) dagger.WithContainerFunc {
+	return func(ctr *dagger.Container) *dagger.Container {
+		return ctr.
+			// Create the blueprint module
+			WithWorkdir("blueprints/"+name).
+			With(daggerExec("init", "--sdk="+dangSDK, "--name="+name)).
+			WithNewFile("main.dang", source).
+			WithWorkdir("../../").
+			// Init the workspace root module using the blueprint
+			With(daggerExec("init", "--blueprint=./blueprints/"+name))
+	}
+}
+
+// TestWorkspaceBlueprint verifies that a blueprint module accepting a Workspace
+// argument can access the host filesystem, just like a toolchain module.
+func (WorkspaceSuite) TestBlueprint(ctx context.Context, t *testctx.T) {
+	c := connect(ctx, t)
+
+	ctr := workspaceBase(t, c).
+		WithNewFile("hello.txt", "hello from workspace").
+		With(initDangBlueprint("greeter", `
+type Greeter {
+  pub source: Directory!
+
+  new(source: Workspace!) {
+    self.source = source.directory(".")
+    self
+  }
+
+  pub read: String! {
+    source.file("hello.txt").contents
+  }
+}
+`))
+
+	out, err := ctr.With(daggerCall("read")).Stdout(ctx)
+	require.NoError(t, err)
+	require.Equal(t, "hello from workspace", strings.TrimSpace(out))
+}
+
 // TestWorkspaceFindUp verifies that Workspace.findUp searches up from the
 // start path and stops at the workspace root.
 func (WorkspaceSuite) TestFindUp(ctx context.Context, t *testctx.T) {


shykes (commented):
API nits:

- `install` and `moduleInit` are named inconsistently. Maybe `moduleInstall` and `moduleInit`, or `installModule` and `installModule`?

- Missing `Workspace.search`  and `Workspace.findUp` (so I can port all our toolchains :) But can be a follow-up

core/workspace.go
```diff
@@ -0,0 +1,31 @@
+package core
+
+import "github.com/vektah/gqlparser/v2/ast"
+
+// Workspace represents a detected workspace in the dagql schema.
+type Workspace struct {
+	Root       string `field:"true" doc:"Absolute path to the workspace root directory."`
+	ConfigPath string `field:"true" doc:"Path to config.toml (empty string if no config exists)."`
```
sipsma: Don't care if this is in the doc string vs. just a separate comment for devs above the fields, but I'd suggest being precise about whether ConfigPath is absolute vs. relative (and if it's relative, what is it relative *to*). I always end up wondering that about path values if there's no explicit doc on it.

dagql/cache.go
```diff
@@ -509,6 +509,12 @@ func (r Result[T]) WithContentDigest(customDigest digest.Digest) Result[T] {
 	return r
 }
 
+// WithContentDigestAny is WithContentDigest but returns an AnyResult, required
+// for polymorphic code paths like module function call plumbing.
+func (r Result[T]) WithContentDigestAny(customDigest digest.Digest) AnyResult {
```
sipsma, line 514: I'd probably add a `(r ObjectResult[T]) WithContentDigestAny(customDigest digest.Digest) AnyResult` too just out of caution. Otherwise you can get deranged situations where you call WithContentDigestAny on something that is underlying a ObjectResult but then you invisibly "downgrade" it to just a Result. Probably doesn't matter yet but could save someone/something hours of headaches someday...

core/modfunc.go
```diff
@@ -889,6 +931,15 @@ func (fn *ModuleFunction) Call(ctx context.Context, opts *CallOpts) (t dagql.Any
...
 		returnValue = returnValue.WithPostCall(resourceTransferPostCall)
+
+		// If this function accepts Workspace args, set a content digest on the
+		// result derived from all content it returned — both core object IDs
+		// (Directory, File, etc.) and primitive scalar values (String, Int, etc.).
+		// This ensures downstream calls that reference this result get a different
+		// cache key when the underlying content changes.
+		if fn.hasWorkspaceArgs() {
+			returnValue = returnValue.WithContentDigestAny(returnedContent.Digest())
+		}
```
sipsma, line 942: Do things break if we make this unconditional? I.e. drop the `if fn.hasWorkspaceArgs() {`?

If not, I'd say do that. The less branches in this logic the better. If it does break something that's okay for now though, can leave it as is and revisit

core/schema/workspace.go
```diff
@@ -0,0 +1,592 @@
...
+				dagql.Arg("skipMigrationCheck").Doc("If true, skip legacy dagger.json migration checks."),
+			),
+	}.Install(srv)
+
+	dagql.Fields[*core.Workspace]{
+		dagql.NodeFuncWithCacheKey("directory",
+			DagOpDirectoryWrapper(
+				srv, s.directory,
+				WithHashContentDir[*core.Workspace, workspaceDirectoryArgs](),
+			), dagql.CacheAsRequested).
```
sipsma: all of these APIs are`CacheAsRequested` but then hard coded to be CachePerClient (just indirectly), can just use CachePerClient directly here and for the others rather than CacheAsRequested

core/schema/workspace.go
```diff
@@ -0,0 +1,592 @@
...
+	}
+	cwd, err := bk.AbsPath(ctx, ".")
+	if err != nil {
+		return nil, fmt.Errorf("cwd: %w", err)
+	}
+
+	statFS := core.NewCallerStatFS(bk)
+	ws, err := workspace.Detect(ctx, statFS, bk.ReadCallerHostFile, cwd)
+	if err != nil {
+		if args.SkipMigrationCheck && strings.Contains(err.Error(), "migration") {
```
sipsma: If there's any way to use `errors.Is`/`As` rather than checking the error string, that'd be greatly preferable

core/schema/workspace.go
```diff
@@ -0,0 +1,592 @@
...
+	if err != nil {
+		return ctx, fmt.Errorf("get client metadata: %w", err)
+	}
+	if clientMetadata.ClientID == ws.ClientID {
+		return ctx, nil // already in the right context
+	}
+	// Clone metadata and override the client ID to the workspace owner's
+	override := *clientMetadata
+	override.ClientID = ws.ClientID
+	return engine.ContextWithClientMetadata(ctx, &override), nil
```
sipsma: I'd suggest adding a method to Server (sibling to [this one maybe](https://github.com/sipsma/dagger/blob/a668ebdd74e702205881f6e90dd33cb5df3263fc/core/query.go#L59-L59)) that gives you the whole clientMetadata for the client and use that.

Otherwise in the current change we'll create a weird frankenstein by overriding just the client ID and not everything else, which is gonna lead to confusion someday.

core/schema/workspace.go
```diff
@@ -0,0 +1,592 @@
...
+		return inst, fmt.Errorf("workspace file %q: %w", args.Path, err)
+	}
+
+	return inst, nil
+}
+
+// resolveWorkspacePath resolves a path relative to the workspace root.
+// Absolute paths are treated as relative to the root (leading "/" is stripped).
+// Returns an error if the resolved path escapes the workspace root via "..".
+func resolveWorkspacePath(path, root string) (string, error) {
```
sipsma: Not sure if this will work with Windows paths. There's a bunch of utils in the pathutil package that help with making sure the engine (which is *always* Linux) handles windows clients paths.

core/schema/workspace.go
```diff
@@ -0,0 +1,592 @@
...
+				{Name: "disableFindUp", Value: dagql.Boolean(true)},
+			},
+		},
+		dagql.Selector{Field: "kind"},
+	)
+	if err != nil {
+		return "", fmt.Errorf("resolve module kind: %w", err)
+	}
+
+	if kind == core.ModuleSourceKindLocal {
```
sipsma: Should this be an error when kind != local? I would have thought so

core/schema/workspace.go
```diff
@@ -0,0 +1,592 @@
...
+
+	if _, err := tmpFile.Write(configBytes); err != nil {
+		tmpFile.Close()
+		return fmt.Errorf("write temp file: %w", err)
+	}
+	tmpFile.Close()
+
+	if err := bk.LocalFileExport(ctx, tmpFile.Name(), workspace.ConfigFileName, configHostPath, true); err != nil {
+		return fmt.Errorf("export config: %w", err)
+	}
```
sipsma: Just use `bk.IOReaderExport` rather than creating a tmpfile https://github.com/sipsma/dagger/blob/a668ebdd74e702205881f6e90dd33cb5df3263fc/engine/buildkit/filesync.go#L230-L230

core/schema/workspace.go
```diff
@@ -0,0 +1,592 @@
...
+	if err != nil {
+		return "", err
+	}
+	bk, err := query.Buildkit(ctx)
+	if err != nil {
+		return "", fmt.Errorf("buildkit: %w", err)
+	}
+
+	// Module lives at .dagger/modules/<name>/ relative to workspace root
+	modulePath := filepath.Join(parent.Root, workspace.WorkspaceDirName, "modules", args.Name)
```
sipsma: Should sanitize `args.Name` (e.g. `../../../../../../`) or otherwise check that modulePath doesn't escape to some parent dirs its not supposed to be in.

core/workspace/config.go
sipsma, line 1: I didn't actually read this file 😎

At least yet. Are we doing a bunch of toml serializing ourselves vs. using a lib just to get some special stuff in like comments?

Totally fine with that if so (good example of something that's more maintainable in the age of LLMs than in the past). But I'd feel better if an LLM generated a shit-ton of unit tests for it too. I'd almost rather review unit tests for something like this than pretend I can really review the implementation in a robust way.

core/workspace/detect.go
```diff
@@ -0,0 +1,151 @@
...
+// Uses a 3-step fallback chain:
+//  1. FindUp .dagger/ → stat .dagger/config.toml → parse config, workspace root = parent of .dagger/
+//  2. FindUp .git → workspace root = directory containing .git (empty workspace)
+//  3. No .git → cwd is workspace root (empty workspace)
+func Detect(
+	ctx context.Context,
+	statFS core.StatFS,
+	readFile func(ctx context.Context, path string) ([]byte, error),
+	cwd string,
+) (*Workspace, error) {
```
sipsma: This is another tricky thing where it might be nice to have an LLM write unit test coverage. Seems like everything is mock-able enough already potentially? (not a blocker if big headache)

core/schema/workspace.go
sipsma: (out-of-date, but applies for PR that revives this)

core/schema/workspace.go
sipsma: (out-of-date, but applies for PR that revives this)

core/schema/workspace.go
sipsma: (out-of-date, but applies for PR that revives this)

core/workspace/config.go
shykes: @sipsma all good questions. I think @vito will prefer to remove these files from this PR altogether - they're probably not needed, and can stay in the slower train #11812 , which will be rebased on this.

shykes: (short answer is: we're using a lib to preserve comments and formatting)

shykes (commented):
I just realized we need this for blueprints too... Better to do in this PR, or as a follow-up?

core/modfunc.go
vito: I was really tempted to do this, will try it out! There's definitely something cool about module function calls being content-addressed instead of recipe-addressed like most other places: we already require the response to be serializable, and even extract proper content digests from returned object IDs, so might as well leverage it even more.

core/schema/workspace.go
vito: Had an LLM take a swing at this - extracted it out into a `pathutil.SandboxedRelativePath` helper with unit tests, but could use some human eyes on it to be 100% sure.

core/workspace/config.go
vito: Yeah I'll just remove these, not needed for this PR.

core/workspace/detect.go
vito: Removed - saving for #11812 where it's relevant

core/schema/workspace.go
vito: Ah right - LLM vomited this out, meant to get back to it and clean it up, definitely looked fishy. Added `Server.SpecificClientMetadata`

core/modfunc.go
vito: Done - seems like it didn't break anything!

(Update: it did. reverted. :D)

PR #11881, opened 2026-02-19
feat(sdk/elixir): check support

wingyplus:
Support `@check true` before a defn function to register it as a Dagger
check via Dagger.Function.with_check/1. Follows the same pattern as the
existing `@cache` attribute.

### DIFF
--- a/core/integration/module_elixir_test.go
+++ b/core/integration/module_elixir_test.go
@@ -333,6 +333,46 @@ func (ElixirSuite) TestReqAdapter(ctx context.Context, t *testctx.T) {
 	require.Equal(t, "hello-from-req-adapter\n", out)
 }
 
+func (ElixirSuite) TestCheck(ctx context.Context, t *testctx.T) {
+	t.Run("list checks", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		out, err := elixirModule(t, c, "hello-with-checks").
+			With(daggerExec("check", "-l")).
+			CombinedOutput(ctx)
+
+		require.NoError(t, err)
+		require.Contains(t, out, "passing-check")
+		require.Contains(t, out, "failing-check")
+		require.Contains(t, out, "passing-container")
+		require.Contains(t, out, "failing-container")
+	})
+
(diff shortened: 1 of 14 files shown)


tiborvass:
https://dagger.cloud/dagger/checks/github.com/dagger/dagger@642d313b85af15d64ae88fce2246a100c2a9bdbf?check=test-split:test-module-runtimes seems like a valid failure

core/integration/module_elixir_test.go
```diff
@@ -311,6 +311,46 @@ func (ElixirSuite) TestReqAdapter(ctx context.Context, t *testctx.T) {
...
+			CombinedOutput(ctx)
+
+		require.NoError(t, err)
+		require.Contains(t, out, "passing-check")
+		require.Contains(t, out, "failing-check")
+		require.Contains(t, out, "passing-container")
+		require.Contains(t, out, "failing-container")
+	})
+
+	t.Run("run passing checks", func(ctx context.Context, t *testctx.T) {
```
grouville, line 351: The CLI prints check names in hyphen-case (see the "list checks" subtest above where we assert passing-check, failing-check, etc.)

I think below regexes will never match, and is actually the failure that Tibor talked about in his comment above

Could you update them to `passing-check`/`passing-container` and `failing-check`/`failing-container`?

wingyplus: Thanks @grouville @tiborvass. I've fixed this, and the CI already passed except the test workspace, which doesn't seem related to this PR.

sdk/elixir/test/dagger/client_test.exs
```diff
@@ -48,6 +48,8 @@ defmodule Dagger.ClientTest do
     assert ["## What is Dagger?" | _] = String.split(readme, "\n")
   end
 
+  # 2 minutes
+  @tag timeout: 180_000
```
wingyplus, line 52: I found the timeout from this test in CI. So I added a timeout from 60 seconds to 180 seconds to make it more reliable.

shykes (approved):
thank you!

Issue #11287, opened 2025-10-23
Dynamic filtering of context directory


shykes:
## Problem

There is a buildkit feature that Dagger can't currently replicate: the ability to dynamically change the upload filter on a local directory.
There's a good reason for that: it breaks our caching model. If dagger can't tell in advance which parts of a local directory your function will need to read, it has to invalidate your function cache if *any* file in that directory changes. This has implications for caching that are difficult to even reason about, and so far it hasn't been worth exploring that.

But, with new upcoming features like toolchains, user defaults, function caching, blueprints, live directories... it's becoming more realistic to expose Dagger modules directly to end users, as "competition" for running the native tools directly.
In this competition, Dagger's main advantage is repeatability. Its main disadvantage is slowness + complexity. We should explore every avenue for overcoming our disadvantage.

Disadvantage of the current situation:

- Complexity: user has to filter directories manually in a project-specific way (and keep the filters up-to-date), because the toolchain can't know in advance every file it will need
- Performance: by default, unless the user manually fine-tunes the filters, too many files will get uploaded. In some cases (very large repos + tools that need broad default filters), this can make the module catastrophically slow out-of-the-box, because it has to upload the entire monorepo every time.

### Initial solution proposed by Solomon (bad 😛)

With the other optimizations coming, in many cases the bottleneck will be file upload. So, if we could make file upload *lazy*, to give toolchains a chance to receive a huge, unfiltered directory as argument, but only actually use parts of it...
The function would still be invalidated based on the digest of the entire directory, but the files wouldn't have to be transferred... This relies on the assumption that 1) computing changes to a big local directory is faster than uploading them, and 2) the time saved not uploading unneeded files is greater than the time wasted re-executing the function when they change.

Note:

- This should not break the caching model
- I'm not sure how the laziness implementation would work. Presumably we would allow downstream `include/exclude` and other filters, to apply to the upload?

My apologies if this is a dumb line of reasoning :-P

### New solution proposed by @sipsma 

[comment here](https://github.com/dagger/dagger/issues/11287#issuecomment-3440807340)

sipsma:
Even if we had this ability, wouldn't toolchains still need to somehow know what additional filters to apply based on the user's settings? If so, at that point we still need the user to be able to customize the filters and I feel like at that point we may as well just allow those filters to apply to the directory argument directly (vs. doing this via dynamic filtering).

> 1) computing changes to a big local directory is faster than uploading them

[...]

PR #11891, opened 2026-02-21
feat(cmd/dagger): support path filters (include, exclude, gitignore) for dagger.Directory arguments

idlsoft:
This is a simple client-side feature to complement the more comprehensive [server-side dynamic filtering](https://github.com/dagger/dagger/issues/11287).

It will allow specifying include/exclude options for arguments of type `dagger.Directory` using a url-like format.
For example: 
```/path/to/monorepo?include=apps/app1,libs/lib1,libs/lib2&exclude=libs/lib1/testdata```

Can be useful for builds that need a subset of directories.
For example a `golang` build, with `replace` support.

Works for command line args, as well as user defaults.

Unlike the dynamic filtering proposal this would most likely require a script to compute the includes/excludes prior to invoking dagger.
I have at least one use-case where this is actually preferable (includes are computed by an native executable).

### DIFF
--- a/core/schema/address.go
+++ b/core/schema/address.go
@@ -182,7 +182,11 @@ func (s *addressSchema) directory(
 			})
 		}
 	} else {
-		q = queryLocalDirectory(addr, args.CopyFilter)
+		if path, err := parsePathWithFilter(addr, &args.CopyFilter); err != nil {
+			return inst, err
+		} else {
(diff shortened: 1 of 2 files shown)


TomChv:
/cc @shykes I remember sometime ago we talked about that idea, I think we conclude that we should look for another solution.

Do you have a new opinion on that matter?

shykes:
wdyt @dagger/maintainers ?

PR #11892, opened 2026-02-21
sdk/typescript: surface detailed error context in UnknownDaggerError

shepherdjerred:
## Summary

When a non-`ClientError` exception occurs during a GraphQL request (e.g. transport failures with `kube-pod://`, Docker socket issues, or container exec errors that don't come back as proper `ClientError`), the TypeScript SDK throws a generic:

```
Encountered an unknown error while requesting data via graphql
```

This makes debugging CI failures extremely difficult. Real build failures (like `exited with code 1` or `binary already exists in destination`) get hidden behind this generic message, leading users to chase phantom transport issues when the root cause is a simple build error.

## Changes

In `sdk/typescript/src/common/graphql/compute_query.ts`, the catch-all error handler now extracts detailed context from the underlying error:

- The original error message
- The error type/class name (if not generic `Error`)  
- The full cause chain (up to 5 levels deep)

## Before

```
Encountered an unknown error while requesting data via graphql
```

## After

```
Encountered an error while requesting data via graphql
Error: process exited with code 1
Type: ExecError
Caused by: build script failed
```

## Context

I hit this repeatedly while running Dagger CI via `kube-pod://` transport. Container exec failures would consistently surface as the generic GraphQL error, making it impossible to tell whether the issue was a transport problem or a real build failure without digging through verbose engine logs.

### DIFF
--- a/sdk/typescript/src/common/graphql/compute_query.ts
+++ b/sdk/typescript/src/common/graphql/compute_query.ts
@@ -245,11 +245,34 @@ export async function compute<T>(
       )
     }
 
-    // Just throw the unknown error
+    // Build a detailed error message from the underlying error so that
+    // transport failures (kube-pod://, Docker, etc.) surface actionable
+    // context instead of the previous generic one-liner.
+    const parts: string[] = ["Encountered an error while requesting data via graphql"]
+
+    if (e instanceof Error) {
+      if (e.message) {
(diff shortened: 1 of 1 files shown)


shepherdjerred:
sorry claude got a bit too excited here

PR #11896, opened 2026-02-23
feat: add sourceMap to ID API

TomChv:
Add the sourcemap directives for new solvers created by the engine when an external typedef is registered in the engine.
This include:
- `loadXXXFromID`
- `XXXID`
- `env` bindings


TomChv:
I discussed with @vito and he suggested that we stick to sourcemap instead of having another directive.

The only change we need to do then is to add sourcemap to method added by the engine such as `LoadXXXFromID` and `XXXID`

PR #11903, opened 2026-02-24
core(modtree): fix `--scale-out` error propagation + telemetry

### LINKED ISSUE #11887: 🐞 `check --scale-out` never fails
### What is the issue?

`dagger check --scale-out` currently shows a passing check and exit status 0 even if checks failed.

There's a different in the local (to the engine) vs. scale-out paths.

The local path directly calls the underlying function, which will bubble up an error if the call returns it (or returns a Syncable object that fails on `sync`):

https://github.com/dagger/dagger/blob/42bb3b568c0a38671561a1f40e7a40cba49cde60/core/modtree.go#L151-L171

The scale-out path however just queries the `passed` state, which will just return `false` if the check fails (which we just ignore anyway):

https://github.com/dagger/dagger/blob/42bb3b568c0a38671561a1f40e7a40cba49cde60/core/modtree.go#L200-L206

Instead it should probably bubble up the underlying error somehow, to maintain symmetry. We at least want the error to bubble up through the telemetry span so it looks the same in both cases. (Or maybe the telemetry span should just be pushed down?)

### Dagger version

v0.19.11

### Steps to reproduce

* run `dagger check --scale-out` with a failing check, observe OK
* run `dagger check`, observe ERROR

### Log output

_No response_

vito:
fixes #11887

Also fixes telemetry for both, not sure if an issue existed for that yet.

* Check span was missing in native CI, making all the traces ugly and hiding their main output
* Generate output was placing everything beneath an extra span
* Generate output was appearing lazily
* Show `foo-bar:baz` instead of `fooBar:baz`

### DIFF
--- a/cmd/dagger/generators.go
+++ b/cmd/dagger/generators.go
@@ -134,15 +134,15 @@ func runGenerators(ctx context.Context, dag *dagger.Client, generatorGroup *dagg
 	// We don't actually use the API for rendering results
 	// Instead, we rely on telemetry
 	// FIXME: this feels a little weird. Can we move the relevant telemetry collection in the API?
-	cs := generatorGroup.
+	cs, err := generatorGroup.
 		Run().
 		Changes(
 			dagger.GeneratorGroupChangesOpts{
 				OnConflict: dagger.ChangesetsMergeConflictFailEarly,
 			},
-		)
-
-	ctx, span := Tracer().Start(ctx, "applying changes")
-	defer telemetry.EndWithCause(span, &rerr)
+		).Sync(ctx)
+	if err != nil {
+		return err
+	}
 	return handleChangesetResponse(ctx, dag, cs, autoApply)
 }


eunomie:
@vito should we add the same logic for generate functions?

vito:
@eunomie I'll look into it, though I'm not sure that we need to. Checks are a bit special because failing is part of their happy path, and we turn that into data like `passed bool`. For generators I'd expect errors to just propagate naturally, since we always expect those to work.

vito:
Tested error propagation + telemetry fixes with scale-out using the process described here - https://github.com/dagger/dagger/pull/11685#issuecomment-3812916604 - should be good to go.

core/modtree.go
```diff
@@ -526,7 +545,7 @@ func (node *ModTreeNode) Match(ctx context.Context, patterns []string) (bool, er
 }
 
 func (node *ModTreeNode) PathString() string {
-	return strings.Join(node.Path(), ":")
+	return strings.Join(node.Path().CliCase(), ":")
```
eunomie, line 548: 👍

PR #11905, opened 2026-02-24
fix(typescript): avoid ts-introspector crash on empty nested src dirs

### LINKED ISSUE #11904: 🐞 failed to get type defs json during module sdk codegen
### What is the issue?

I try to run tests for a pipeline written with the Typescript SDK following the documentation (https://docs.dagger.io/reference/best-practices/modules). When I try to launch the test, it fails with a `failed to get type defs json during module sdk codegen` error

### Dagger version

dagger v0.19.11 (image://registry.dagger.io/engine:v0.19.11) darwin/arm64/v8

### Steps to reproduce

I followed the steps described:

```
mkdir tests
cd tests
dagger init --name=tests --sdk=typescript --source=.
dagger install ..
```

I changed the package manager to use pnpm in my package.json and call the test using `dagger call -m tests all`

I also after debugging and seeing `bunfs` in the stack, I guessed it was using bun even if it is not the expected behavior in the documentation in tried to force using node by adding this in the package.json: 
```
"dagger": {
    "runtime": "node@24.13.1"
  }
```
But it hasn't changed anything.


Here is the index.ts for the test class: 
```
import {func, object} from "@dagger.io/dagger";
import {lintFails} from "./lint/fails/lint-fails";

@object()
export class Tests {
    @func()
    async all(): Promise<void> {
        await lintFails();
    }
}
```

### Log output

```
✔ connect 0.3s
✘ load module: tests 1.6s ERROR
┇ initializing module › ModuleSource.asModule › 
✘ module SDK: load typedefs object 0.7s ERROR
1 | (function (callback )                                                                                                                                                                                                                   
              ^                                                                                                                                                                                                                             
TypeError: reduce of empty array with no initial value                                                                                                                                                                                      
      at reduce (1:11)                                                                                                                                                                                                                      
      at map (1:11)                                                                                                                                                                                                                         
      at getTsSourceCodeFiles (/$bunfs/root/ts-introspector:280392:35)                                                                                                                                                                      
      at map (1:11)                                                                                                                                                                                                                         
      at getTsSourceCodeFiles (/$bunfs/root/ts-introspector:280392:35)                                                                                                                                                                      
      at main (/$bunfs/root/ts-introspector:280414:51)                                                                                                                                                                                      
      at main (/$bunfs/root/ts-introspector:280405:21)                                                                                                                                                                                      
      at /$bunfs/root/ts-introspector:280426:5                                                                                                                                                                                              
                                                                                                                                                                                                                                            
Bun v1.3.0 (Linux arm64)                                                                                                                                                                                                                    
! failed to get type defs json during module sdk codegen: exit code: 1
```

The full logs are available here : https://dagger.cloud/predell/traces/efeeb77c6bcee26726fc93b5c4c7180f

grouville:
Fixes #11904 

### Problem 
A user reporting getting this module init failure:

```markdown
failed to get type defs json during module sdk codegen
TypeError: reduce of empty array with no initial value
```

#### Root cause

`getTsSourceCodeFiles` recursively scans `src/`, and each level reduces `readdirSync(dir).map(...)` without an initial accumulator.

To trigger the error, the module does not need to be empty: any empty nested directory under `src/` can trigger this panic.

#### Solution

This failure occurs during module typedef introspection (`ts-introspector`, Bun path), so forcing `"dagger.runtime": "node@..."` does not change it.

The fix is to seed the reduce with `[]` + I added a regression coverage in `TestTypescript/TestInit/handles_empty_directory_in_src`.

### DIFF
--- a/sdk/typescript/src/module/entrypoint/introspection_entrypoint.ts
+++ b/sdk/typescript/src/module/entrypoint/introspection_entrypoint.ts
@@ -30,7 +30,7 @@ function getTsSourceCodeFiles(dir: string): string[] {
 
       return []
     })
-    .reduce((p, c) => [...c, ...p])
+    .reduce((p, c) => [...c, ...p], [])
 }
 
 async function main() {
--- a/core/integration/module_typescript_test.go
+++ b/core/integration/module_typescript_test.go
@@ -159,6 +159,22 @@ func (TypescriptSuite) TestInit(ctx context.Context, t *testctx.T) {
 		require.JSONEq(t, `{"existingSource":{"helloWorld":{"stdout":"hello\n"}}}`, out)
 	})
 
+	t.Run("handles empty directory in src", func(ctx context.Context, t *testctx.T) {
+		c := connect(ctx, t)
+
+		modGen := c.Container().From(golangImage).
+			WithMountedFile(testCLIBinPath, daggerCliFile(t, c)).
+			WithWorkdir("/work").
+			With(daggerExec("init", "--source=.", "--name=emptyDir", "--sdk=typescript")).
+			WithDirectory("/work/src/empty-dir", c.Directory())
+
+		out, err := modGen.
+			With(daggerQuery(`{emptyDir{containerEcho(stringArg:"hello"){stdout}}}`)).
+			Stdout(ctx)
+		require.NoError(t, err)
+		require.JSONEq(t, `{"emptyDir":{"containerEcho":{"stdout":"hello\n"}}}`, out)
+	})
+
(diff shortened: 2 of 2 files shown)


sdk/typescript/src/module/entrypoint/introspection_entrypoint.ts
```diff
@@ -30,7 +30,7 @@ function getTsSourceCodeFiles(dir: string): string[] {
 
       return []
     })
-    .reduce((p, c) => [...c, ...p])
+    .reduce((p, c) => [...c, ...p], [])
```
eunomie, line 33: it looks like black magic, or perl oneliner :-D

PR #11906, opened 2026-02-24
Toolchain config support for glob patterns for function and argument names

### LINKED ISSUE #11884: ✨ Toolchain-wide settings

We are trying to build toolchains that offer a very standardized out-of-the-box experience with our templates. While doing this we noticed that it's not easy to design these without some arguments being repeated between the functions we expose.

For example, for our dotnet toolchain, we want to offer an override for the SDK image to be used. This is needed for pretty much all the features we expose and we end up with a large amount of repeated defaults in customizations. It's also not great as it requires listing the functions available in the toolchain - if we add one more later, it will require adding a customization too.

Would there be an interest for a solution to this that is configurable in dagger.json? It could take the form of a customization without function name that would apply to arguments across all functions in the toolchain (this feels a bit wrong to me as it's very implicit) or, my preference, offer a toolchain configuration section that toolchains can always read

_No response_

The main workaround we have thought of so far is to use a separate configuration file so that we can configure this in one place. Every function would read the override from that file.

### LINKED ISSUE #12858: Toolchain customizations use surprising function/argument name casing
Toolchain customizations in dagger.json require the source code casing for function and argument names (e.g. `ContainerEcho` in Go), rather than the CLI-style names (e.g. `container-echo`). This is surprising since all other user-facing views use normalized names.

Fix in progress: PR #11906.

Originally reported in #11883 (converted to discussion #12806 during triage).

suprjinx:
This PR adds glob pattern support in the `function` and `argument` fields of toolchain customizations, allowing a single customization to apply to multiple functions. This allows setting a default for an argument across all functions without repeating the customization:

```json
{
  "name": "app",
  "engineVersion": "v0.19.4",
  "toolchains": [
    {
      "name": "hello",
      "source": "../hello",
      "customizations": [
        {
          "function": ["*"],
          "argument": "message",
          "default": "hola"
        },
        {
          "function": ["configurableMessage"],
          "argument": "message",
          "default": "bonjour"
        }
      ]
    }
  ]
}
```

Last match wins, so specific function patterns can override wildcards.

Additionally, function and argument names are cleansed of underscore and hyphens, so they can be supplied in customization as they are on command line (was already ignoring case). Eg, "kebab-argument" will match "kebabArgument" in function def.

Fixes #11884
Fixes #12858

Assisted by Claude Code


shykes:
@grouville @tiborvass want to take it?

grouville (dismissed):
One edge case was missing: matching the glob of chained function. I directly rebased on top of main + added a commit introducing a regression test and its fix.

Overall, in this current state, LGTM ✅

grouville:
Ok, the failure seems legit, digging 👀 Yes, it seems related to the merge of https://github.com/dagger/dagger/pull/11995.


I have a fix but we'll implement it at the end of the week

suprjinx:
@grouville can we reopen this and review?

suprjinx:
@grouville can we reopen this and review? it will simplify a lot of error-prone toolchain configuration we make currently to set identical defaults on a large number of functions.

PR #11908, opened 2026-02-24
install: honor --eager-runtime

tiborvass:
This pull request aims at allowing `dagger install` to honor `--eager-runtime` flag.

I'm not sure what is the correct design here.
Option 1 (this PR): `install` defaults to not load the runtime, but honors `--eager-runtime` flag.
Option 2: always make it eager + possibly have a `--lazy-runtime` ?


grouville (approved):
Thanks @tiborvass 🙏 

Following a discussion with @shykes, this is the correct fix: honor the flag to at least have an escape hatch when people want to verify that an installed module is properly loading.

The naming of the flag could be improved later on

PR #11909, opened 2026-02-25
core/git: default remote fetches to `--no-tags` (~80% faster fetches)

grouville:
### Problem
The remote fetch hot path currently uses `git fetch --tags` unconditionally. That is expensive on tag-heavy repos and makes branch/commit/tree flows pay global tag transfer cost even though tag APIs and ref resolution already come from `ls-remote`.

There is a real drawback to removing `--tags`: some Git servers/proxies reject direct SHA fetches (`not our ref`, `unadvertised object`, `couldn't find remote ref`). BuildKit historically handled this with conditional behavior (`--no-tags` for normal ref fetches, `--tags` for commit-oriented fetches as a compatibility hedge). After Git handling moved into dagop operations, that compatibility path was effectively lost while global `--tags` remained in the hot path.

### Solution
This change keeps SHA-first pinning and switches the default transport to `--no-tags`, then restores compatibility with a bounded retry: when SHA fetch is classified as unsupported, retry once with named refspecs into deterministic scratch refs. In other words, we do not keep conditional `--tags` as part of normal API behavior; we honor the same compatibility requirement only for remotes that actually need fallback.

### Benchmark
Local benchmarks using the git CLI show why it's worth implementing: cold fetches on tag-heavy repos improve materially (`git/git` 54.64s -> 7.00s, `dagger/dagger` 73.79s -> 20.51s). In engine-like fetch shape (`<sha>:refs/dagger.fetch/...`), warm runs are also fast once the scratch ref exists (0.44s, 0.40s, 0.41s).

Add `ErrSHAFetchUnsupported` classification in `gitutil` and focused tests for retry refspec determinism + error classification priority. Cache key shape, auth/session metadata cache scoping, and `ls-remote`-backed tag/version semantics are unchanged.

### DIFF
--- a/core/git_remote.go
+++ b/core/git_remote.go
@@ -357,7 +357,7 @@ func (repo *RemoteGitRepository) fetch(ctx context.Context, git *gitutil.GitCLI,
 		return err
 	}
 
-	if len(refs) == 0 {
+	if len(refs) == 0 && !includeTags {
 		// Nothing requested: avoid an implicit broad fetch from origin.
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


core/git_remote.go
```diff
@@ -357,33 +357,56 @@ func (repo *RemoteGitRepository) fetch(ctx context.Context, git *gitutil.GitCLI,
...
+	// Fetch by object SHA in the hot path (`--no-tags`), and only retry by named refs for SHA-incompatible remotes.
+	logger := slog.SpanLogger(ctx, InstrumentationLibrary)
+
 	gitDir, err := git.GitDir(ctx)
 	if err != nil {
 		return err
 	}
 
-	var refSpecs []string
+	shaRefSpecs := make([]string, 0, len(refs))
```
tiborvass: Nit
```suggestion
	shaRefSpecs := make([]string, len(refs))
```

core/git_remote.go
```diff
@@ -357,33 +357,56 @@ func (repo *RemoteGitRepository) fetch(ctx context.Context, git *gitutil.GitCLI,
...
 	for _, ref := range refs {
```
tiborvass: Nit
```suggestion
	for i, ref := range refs {
```

core/git_remote.go
```diff
@@ -357,33 +357,56 @@ func (repo *RemoteGitRepository) fetch(ctx context.Context, git *gitutil.GitCLI,
...
-		// fetch by sha, since we've already done tag resolution
-		// TODO: may need fallback if git remote doesn't support fetching by commit
-		refSpecs = append(refSpecs, ref.SHA)
+		// Default hot path: fetch exact objects by SHA; ref names are already resolved via ls-remote.
+		shaRefSpecs = append(shaRefSpecs, ref.SHA)
```
tiborvass: Nit
```suggestion
		shaRefSpecs[i] = ref.SHA
```

core/git_remote.go
```diff
@@ -395,23 +418,43 @@ func (repo *RemoteGitRepository) fetch(ctx context.Context, git *gitutil.GitCLI,
...
-	if _, err := git.Run(ctx, args...); err != nil {
-		if errors.Is(err, gitutil.ErrShallowNotSupported) {
-			// fallback to full fetch
-			args = slices.DeleteFunc(args, func(s string) bool {
-				return strings.HasPrefix(s, "--depth")
-			})
-			_, err = git.Run(ctx, args...)
+	if err := runFetch(shaRefSpecs); err == nil {
+		return nil
+	} else {
```
tiborvass: early return need not be followed by else clause

core/git_remote.go
```diff
@@ -395,23 +418,43 @@ func (repo *RemoteGitRepository) fetch(ctx context.Context, git *gitutil.GitCLI,
...
 		}
 
-		if err != nil {
+		namedSpecs := namedFetchRefSpecs(refs)
+		if len(namedSpecs) == 0 {
 			return fmt.Errorf("failed to fetch remote %s: %w", repo.URL.Remote(), err)
 		}
+
+		logger.Info("git fetch by sha failed; retrying with named refs", "remote", repo.URL.Remote(), "refspec_count", len(namedSpecs))
+		if retryErr := runFetch(namedSpecs); retryErr != nil {
```
tiborvass: Since you implement the fallback to fetching by named refs, you should still verify that what was pulled matches the expected SHAs at least for the refs that have an expected SHA.

grouville:
~~I added a fallback (temporary that fetches in the bare repo the tags when depths > 0, in a followup commit). Feels weird but allows to pass the tests (to check the perf improvement on the entire test suite). It's totally a design decision: we might want to explicit it as part of the API _maybe_  OR accept the consequence of the `--no-tags`~~

#### Update

It does seem to bring performance improvements. I'll add an integration test / a setup for others to benchmark. 

@sipsma interested in your benchmark feedback / feeling on whether this actually improves performance on fetches 🙏

core/git_remote.go
```diff
@@ -395,23 +423,48 @@ func (repo *RemoteGitRepository) fetch(ctx context.Context, git *gitutil.GitCLI,
...
+			return fmt.Errorf("failed to fetch remote %s: %w", repo.URL.Remote(), err)
 		}
 
-		if err != nil {
+		namedSpecs := namedFetchRefSpecs(refs)
+		if len(namedSpecs) == 0 {
 			return fmt.Errorf("failed to fetch remote %s: %w", repo.URL.Remote(), err)
 		}
+
+		logger.Info("git fetch by sha failed; retrying with named refs", "remote", repo.URL.Remote(), "refspec_count", len(namedSpecs))
```
sipsma: nit: lower to Debug, same for logger.Info below (been trying to keep our logs more useful and less noisy on Info level)

PR #11910, opened 2026-02-25
python-sdk-dev: split tests by py version and type

eunomie:
Introduce a specific object type exposing `unit` and `slow` check function executing tests.
Instead of relying on the python version matrix and handling the parallelism when calling the tests, use the built-in `check` behaviour. All the check functions will be run in parallel.

This change gives more visibility and more control to the developer to run python sdk tests.

Previously only one single check was available regarding those tests:

    $ dagger check -l
    ...
    python-sdk:test
    ...

By using `dagger check` it's only possible to run all the tests against all the different python versions. But when iterating on a python feature it could be better to simply run against one specific version, and only the unit tests for instance.

This is doable by switching to `dagger call`:

    $ dagger call python-sdk test-suite --version 3.14 unit

But it's sad we have to use two different types of calling dagger functions to do that.

A new function, named `python-PYTHON_VERSION` returns a new instance of the `TestForPythonVersion` object with the PYTHON_VERSION. This object contains the two checks.
One function is created for each python version of the test matrix. The check list become:

    $ dagger check -l
    ...
    python-sdk:python-310:slow            Run python slow tests
    python-sdk:python-310:unit            Run python unit tests
    python-sdk:python-311:slow            Run python slow tests
    python-sdk:python-311:unit            Run python unit tests
    python-sdk:python-312:slow            Run python slow tests
    python-sdk:python-312:unit            Run python unit tests
    python-sdk:python-313:slow            Run python slow tests
    python-sdk:python-313:unit            Run python unit tests
    python-sdk:python-314:slow            Run python slow tests
    python-sdk:python-314:unit            Run python unit tests
    ...

When `dagger check` will be run, all the tests will run, as expected. But we can now benefit from the path matching to do more things, and keeping a call to check:

For instance, to run unit tests against python 3.14 just use the check path

    $ dagger check python-sdk:python-314:unit

To run unit tests against all python versions

    $ dagger check python-sdk:**:unit

To run all tests for a specific python version

    $ dagger check python-sdk:python-314

To run all tests against all python versions

    $ dagger check python-sdk:python-*

This gives more control, more granularity and a more explicit way to run python tests.

### DIFF
--- a/toolchains/python-sdk-dev/main.go
+++ b/toolchains/python-sdk-dev/main.go
@@ -5,11 +5,10 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"runtime"
 	"strings"
 
 	"dagger/python-sdk-dev/internal/dagger"
-
-	"github.com/dagger/dagger/util/parallel"
 )
 
 // A toolchain to develop the Dagger Python SDK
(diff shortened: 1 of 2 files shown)


TomChv (commented):
LGTM, looks nice to more checks in the CI.

Note that this should be refactor once we support dynamic checks

TomChv (approved):

